PR #13565, opened 2018-09-05
test: add test for fragement props

gebilaoxiong:
This test case was added for this part of the code：

https://github.com/facebook/react/blob/master/packages/react/src/ReactElementValidator.js#L243-L248

Although the relevant test case is already included in `ReactJSXElementValidator-test.js`

I think we need to add a test case, which created by `React.createElement`


gaearon:
Sounds good, thanks

PR #13566, opened 2018-09-05
refactor: remove unnecessary judgment when check fragment props

gebilaoxiong:
There is a small problem with the judgment here.

**When we define `key` in props we use the descriptor like this**
```javascript
  Object.defineProperty(props, 'key', {
    get: warnAboutAccessingKey,
    configurable: true,
  });
```
When `enumerable` is not defined, the default value is false.
doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty

So when we use `Object.keys` to get the fragement props, we will never get the `key`.
code: https://github.com/facebook/react/blob/master/packages/react/src/ReactElementValidator.js#L239

### DIFF
--- a/packages/react/src/ReactElementValidator.js
+++ b/packages/react/src/ReactElementValidator.js
@@ -239,7 +239,7 @@ function validateFragmentProps(fragment) {
   const keys = Object.keys(fragment.props);
   for (let i = 0; i < keys.length; i++) {
     const key = keys[i];
-    if (key !== 'children' && key !== 'key') {
+    if (key !== 'children') {
       warning(
         false,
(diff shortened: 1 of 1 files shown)


react-linter:
<details>

  <p>Comparing: 9a110ebd8ce9f808a246a3c7f7b0cc13f78f13a6...7aef6317d77f3820fa22884cd5fed439912a1055</p>


  
## schedule

  </details>
  
<p align="right">
</p>

gebilaoxiong:
@gaearon 

If we continue to use `Object.keys` later
We can intuitively find there is no need to judge `key` and `ref`..

PR #13582, opened 2018-09-06
[schedule] Refactor Schedule, remove React-isms

acdlite:
Once the API stabilizes, we will move Schedule into a separate repo. To promote adoption, especially by projects outside the React ecosystem, we'll remove all React-isms from the source and keep it as simple as possible:

- No build step.
- No static types.
- Everything is in a single file.

If we end up needing to support multiple targets, like CommonJS and ESM, we can still avoid a build step by maintaining two copies of the same file, but with different exports.

This commit also refactors the implementation to split out the DOM- specific parts (essentially a requestIdleCallback polyfill). Aside from the architectural benefits, this also makes it possible to write host-agnostic tests. If/when we publish a version of Schedule that targets other environments, like React Native, we can run these same tests across all implementations.

### DIFF
--- a/packages/schedule/src/Schedule.js
+++ b/packages/schedule/src/Schedule.js
@@ -3,88 +3,82 @@
  *
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
+ *
  */
 
-// TODO: Currently there's only a single priority level, Deferred. Need to add
-// additional priorities (serial, offscreen).
-const Deferred = 0;
-const DEFERRED_TIMEOUT = 5000;
+/* eslint-disable no-var */
+
+// TODO: Currently there's only a single priority level, Deferred. Will add
+// additional priorities.
+var DEFERRED_TIMEOUT = 5000;
 
 // Callbacks are stored as a circular, doubly linked list.
-let firstCallbackNode = null;
+var firstCallbackNode = null;
 
-let priorityContext = Deferred;
-let isPerformingWork = false;
+var isPerformingWork = false;
 
-let isHostCallbackScheduled = false;
+var isHostCallbackScheduled = false;
 
-const hasNativePerformanceNow =
+var hasNativePerformanceNow =
   typeof performance === 'object' && typeof performance.now === 'function';
 
-let timeRemaining;
+var timeRemaining;
 if (hasNativePerformanceNow) {
-  timeRemaining = () => {
+  timeRemaining = function() {
     // We assume that if we have a performance timer that the rAF callback
     // gets a performance timer value. Not sure if this is always true.
-    const remaining = getTimeRemaining() - performance.now();
+    var remaining = getFrameDeadline() - performance.now();
     return remaining > 0 ? remaining : 0;
   };
 } else {
-  timeRemaining = () => {
+  timeRemaining = function() {
     // Fallback to Date.now()
-    const remaining = getTimeRemaining() - Date.now();
+    var remaining = getFrameDeadline() - Date.now();
     return remaining > 0 ? remaining : 0;
   };
 }
 
-const deadlineObject = {
+var deadlineObject = {
   timeRemaining,
   didTimeout: false,
 };
 
-function ensureHostCallbackIsScheduled(highestPriorityNode) {
+function ensureHostCallbackIsScheduled() {
   if (isPerformingWork) {
     // Don't schedule work yet; wait until the next time we yield.
     return;
   }
-  const timesOutAt = highestPriorityNode.timesOutAt;
+  // Schedule the host callback using the earliest timeout in the list.
+  var timesOutAt = firstCallbackNode.timesOutAt;
   if (!isHostCallbackScheduled) {
     isHostCallbackScheduled = true;
   } else {
-    // Cancel the existing work.
+    // Cancel the existing host callback.
     cancelCallback();
   }
-  // Schedule work using the highest priority callback's timeout.
   requestCallback(flushWork, timesOutAt);
 }
 
-function computeAbsoluteTimeoutForPriority(currentTime, priority) {
-  if (priority === Deferred) {
-    return currentTime + DEFERRED_TIMEOUT;
-  }
-  throw new Error('Not yet implemented.');
-}
-
-function flushCallback(node) {
-  // This is already true; only assigning to appease Flow.
-  firstCallbackNode = node;
+function flushFirstCallback(node) {
+  var flushedNode = firstCallbackNode;
 
   // Remove the node from the list before calling the callback. That way the
   // list is in a consistent state even if the callback throws.
-  const next = firstCallbackNode.next;
+  var next = firstCallbackNode.next;
   if (firstCallbackNode === next) {
     // This is the last callback in the list.
     firstCallbackNode = null;
+    next = null;
   } else {
-    const previous = firstCallbackNode.previous;
+    var previous = firstCallbackNode.previous;
     firstCallbackNode = previous.next = next;
     next.previous = previous;
   }
 
-  node.next = node.previous = null;
+  flushedNode.next = flushedNode.previous = null;
 
   // Now it's safe to call the callback.
-  const callback = node.callback;
+  var callback = flushedNode.callback;
   callback(deadlineObject);
 }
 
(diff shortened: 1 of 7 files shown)


packages/schedule/src/__tests__/ScheduleDOM-test.js
```diff
@@ -309,10 +309,10 @@ describe('ScheduleDOM', () => {
           const callbackC = jest.fn(() => callbackLog.push('C'));
           const callbackD = jest.fn(() => callbackLog.push('D'));
 
-          scheduleWork(callbackA); // won't time out
+          scheduleWork(callbackA, {timeout: 100}); // won't time out
```
acdlite: Needed to make a small tweak to an existing test: Callbacks are now executed in timeout-order. This is the only semantic change.

gaearon:
Do you expect to fix the timeout issue I encountered as part of this, or is it completely orthogonal?

react-linter:
<details>

  <p>Comparing: d66505dbc7172ec8648bb0d2891bf823c5da7e5e...79cae075bf68a7ed98801d2c6758edecc1f8463a</p>


  
## schedule

  </details>
  
<p align="right">
</p>

acdlite:
@gaearon In a follow-up. Though I might have accidentally fixed it. This PR is just shuffling things around in preparation for bigger changes.

gaearon:
Oh wow. I don't know what exactly changed but the time slicing fixture is now super smooth and doesn't lock up every few seconds. It's actually worrying because I don't like that the experience can change so much without tests also changing.

Interestingly I still don't get into the `.didTimeout` case you added earlier. But maybe some other change improved scheduling so that we don't actually timeout so often?

acdlite:
Yeah we definitely need more tests. I added some in this PR, planning to write many more.

sebmarkbage:
I think that we should drop all bundling (rollup), all minification, all build scripts for this package. It should just be a single file in its final location. `npm publish` is literally just the source of the package. It needs to be super clear exactly what you get (the content of a single file) and super unopinionated.

acdlite:
Ok I'll move everything back into one file and just put a big line in between the two halves, like we do with ReactFiberScheduler :D

sebmarkbage:
We should probably drop types and put them in separate libdefs. If we need ES module exports maybe we just have two copies of the file and keep them in sync.

plievone:
I read the tests in this and the continuations pull request briefly through. It seems that most of the tests consume the frame budget exactly, which is nice from the edge case standpoint, but the most common case might be where the last task goes a bit over the frame budget, and it would be good to document this and its variations and implications in the tests.

TrySound:
What's the point of not having a build step? Simple babel+commonjs plugin in `prepare` hook may work fine without effort in supporting two versions.

gaearon:
Hmm. Not sure what's going on here, but if you put

```js
<html>
  <body>
    <script src="../../../build/dist/react.development.js"></script>
    <script src="../../../build/dist/react-dom.development.js"></script>
    <script src="https://unpkg.com/babel-standalone@6/babel.js"></script>
    <div id="container"></div>
    <script type="text/babel">
      ReactDOM.unstable_createRoot(
      ).render(<h1>Hello World!</h1>);
    </script>
  </body>
</html>
```

into `packaging/fixtures/babel-standalone/dev.html`, and run `yarn build core,dom-client,schedule --type=UMD_DEV`, you'll see it not commit anything for five seconds until it expires. Seems like a bug?

gaearon:
Specifically it seems like in async mode we don't get deeper than `renderRoot` until we expire.

screen shot 2018-09-09 at 18 35 56

packages/schedule/src/Schedule.js
```diff
@@ -3,174 +3,327 @@
...
-// By separating the idle call into a separate event tick we ensure that
-// layout, paint and other browser work is counted against the available time.
-// The frame rate is dynamically adjusted.
+let priorityContext = Deferred;
+let isPerformingWork = false;
+
+let isHostCallbackScheduled = false;
+
+let timeRemaining;
+if (hasNativePerformanceNow) {
```
gaearon: This is used before declaration. Hence the bugs.

gaearon: https://github.com/facebook/react/pull/13606

gaearon:
Can also confirm time slicing fixture still works as intended (incl. the fix to its behavior on timeout) 👍

packages/schedule/src/Schedule.js
```diff
@@ -3,174 +3,327 @@
...
+  let timeoutID = -1;
+  requestCallback = (callback, absoluteTimeout) => {
+    timeoutID = setTimeout(callback, 0, true);
   };
-  cancelScheduledWork = function(callbackId: CallbackIdType) {
-    const callback = callbackId.scheduledCallback;
-    const timeoutId = timeoutIds.get(callback);
-    timeoutIds.delete(callbackId);
-    localClearTimeout(timeoutId);
+  cancelCallback = () => {
```
sebmarkbage: Why arrow functions? I figured this would just become an ES file eventually?

sebmarkbage: I mean ES5 file so that it can be published to npm directly.

NE-SmallTown: Ahh, when you use 'ES file', seems we all think it refer to 'ES6+ file'

acdlite: Should I get rid of const as well then?

packages/schedule/src/Schedule.js
```diff
@@ -3,174 +3,327 @@
...
+  deadlineObject.didTimeout = didTimeout;
+  try {
+    if (firstCallbackNode !== null) {
+      if (didTimeout) {
+        // Flush all the timed out callbacks without yielding.
+        do {
+          flushCallback(firstCallbackNode);
+        } while (
+          firstCallbackNode !== null &&
+          firstCallbackNode.timesOutAt <= getCurrentTime()
```
gaearon: Since `performance.now()` calls have a cost, I wonder if we could avoid recalculating it while we have expired work (since if 100 things expired checking the time is unnecessary — they can't "unexpire"). For example by having a nested loop. You read the time in the outer loop, and compare to it in the inner loop. Once you exit the inner loop, you read the time again. If there's nothing new that expired then we exit.

packages/schedule/src/Schedule.js
```diff
@@ -3,174 +3,327 @@
...
+        );
+      } else {
+        // Keep flushing callbacks until we run out of time in the frame.
+        while (
+          firstCallbackNode !== null &&
+          getTimeRemaining() - getCurrentTime() > 0
```
gaearon: I wonder if this could leave a small buffer as a heuristic. So we don't go over by a few ms from last iteration every time.

packages/schedule/src/Schedule.js
```diff
@@ -3,174 +3,327 @@
...
+  if (firstCallbackNode === null) {
+    // This is the first callback in the list.
+    firstCallbackNode = newNode.next = newNode.previous = newNode;
+    ensureHostCallbackIsScheduled(firstCallbackNode);
+  } else {
+    let next = null;
+    let node = firstCallbackNode;
+    do {
+      if (node.timesOutAt > timesOutAt) {
+        // This callback is lower priority than the new one.
```
gaearon: I'd like to avoid using the word "priority" in this file if we can since mixing that with timeouts changes the meaning of "higher" and "lower".

e.g.

```js
// The new callback times out before this one
```

packages/schedule/src/Schedule.js
```diff
@@ -3,174 +3,327 @@
...
+        next = node;
+        break;
+      }
+      node = node.next;
+    } while (node !== firstCallbackNode);
+
+    if (next === null) {
+      // No lower priority callback was found, which means the new callback is
```
gaearon: Same re:wording

packages/schedule/src/Schedule.js
```diff
@@ -3,174 +3,327 @@
...
-  cancelScheduledWork = function(callbackId: CallbackIdType) {
-    const callback = callbackId.scheduledCallback;
-    const timeoutId = timeoutIds.get(callback);
-    timeoutIds.delete(callbackId);
-    localClearTimeout(timeoutId);
+  cancelCallback = () => {
+    clearTimeout(timeoutID);
   };
+  getTimeRemaining = () => 0;
+} else if (window._sched) {
```
gaearon: Maybe `window.__scheduleMock` or something more obscure?

packages/schedule/src/Schedule.js
```diff
@@ -4,323 +4,390 @@
...
+          firstCallbackNode !== null &&
+          getFrameDeadline() - getCurrentTime() > 0
         );
       }
     }
+  } finally {
+    isPerformingWork = false;
+    if (firstCallbackNode !== null) {
+      // There's still work remaining. Request another callback.
+      ensureHostCallbackIsScheduled(firstCallbackNode);
```
NE-SmallTown: @acdlite Seems `ensureHostCallbackIsScheduled` doesn't use any argument in the function body, can I ask that why we pass the `firstCallbackNode` here?

gaearon: For Flow I think

gaearon: hmm wait this doesn't make sense. I dunno.

bvaughn: I think in the initial commit (7c4ff13) `ensureHostCallbackIsScheduled ` expected a param, but the second commit (5f74f04) changed it to just read from the package-level `firstCallbackNode` variable since that's always the value that was passed in anyway. Seems like the call sites were just overlooked as part of this change.

NE-SmallTown: I think the lint should warn on this but don't know why not

PR #13583, opened 2018-09-07
Fixed small CHANGELOG error

bvaughn:
Noticed this while scanning the changelog earlier. Looks like we accidentally linked to the issue rather than the PR that fixed it.


gaearon:
let's also fix https://github.com/facebook/react/releases/tag/v16.5.0?

bvaughn:
Oh whoops. Sure.

PR #12063, opened 2018-01-21
DOM diff and component stack in SSR hydrate mismatch warnings (#10085)

sompylasar:
Renders DOM attributes in the tags mentioned in the warnings. Borrows the idea and partially implementation of `getNodeSignature` from @giles-v https://github.com/facebook/react/pull/12115

Renders DOM diff and component stack showing visually the location where the hydration failed. Example warning with a diff (see [the `fixtures/ssr`](https://github.com/facebook/react/pull/12063/files#diff-5ec2d35d1f0d9ccc2f3a4a53d0615ba4) and [the tests](https://github.com/facebook/react/pull/12063/files#diff-1ee0c467f9deb87ae13e737c58c4ac34) for more warning examples):
```
Warning: Expected server HTML to contain a matching <div> in <div>.

  <div>
    {'nested'}
    <!-- -->
    {' '}
    <p>children <b>text</b></p>
+   <div>{['children ', …]}</div>
  </div>

    in div (at SSRMismatchTest.js:280)
    in div (at SSRMismatchTest.js:275)
    in div (at SSRMismatchTest.js:308)
    in SSRMismatchTest (at App.js:14)
    in div (at App.js:11)
    in body (at Chrome.js:17)
    in html (at Chrome.js:9)
    in Chrome (at App.js:10)
    in App (at index.js:8)
```
react-fixtures-ssr-example

Requires changes to ReactFiberReconciler interface functions that handle hydration errors to distinguish insertion from replacement and show insertion as one added line in the diff; show replacement as one removed, one added line, at correct position among the parentInstance's DOM children:
- add `index` (use `fiber.index`) to point at which child node the insertion or replacement occurs;
- add `isReplaced` to distinguish insertion from replacement.

<details>
<summary>
The latest screen recording from <code>fixtures/ssr</code> (7.8MB – click to expand)
</summary>

react-fixtures-ssr-10mb

</details>

------

<details><summary>Previous revisions</summary>

```
Warning: Expected server HTML to contain a matching <div>{['children ', …]}</div> in <div>nested<!-- --> <p>children <b>text</b></p></div>.
  <div>
    {'nested'}
    {' '}
    <p>children <b>text</b></p>
+   <div>{['children ', …]}</div>
  </div>
    in div (at SSRMismatchTest.js:280)
    in div (at SSRMismatchTest.js:275)
    in div (at SSRMismatchTest.js:308)
    in SSRMismatchTest (at App.js:14)
    in div (at App.js:11)
[...]

Issue #10085, opened 2017-07-01
Nicer Formatting of SSR Validation


sebmarkbage:
The new validation in #10026 only issues a warn for the first difference found in a HTML hydration scenario. Ideally it should instead queue up all the differences and then at the end (commit) issue a single warning with a nicely formatted diff.

1) Instead of warning add [these warn calls](https://github.com/facebook/react/blob/8d61138186e79e1e719786c8c76186e64b603bd5/src/renderers/dom/fiber/ReactDOMFiberComponent.js#L74-L115) to a global buffer (array, map, set, whatever).

2) Inside [prepareForCommit](https://github.com/facebook/react/blob/8d61138186e79e1e719786c8c76186e64b603bd5/src/renderers/dom/fiber/ReactDOMFiberEntry.js#L190), issue all the currently batched up warnings as a single message. 

3) Format that message in terms of a JSX diff in a nicely formatted way. With only the relevant nodes (parent and child with changes). Irrelevant child content can be replaced with ellipsis. E.g.

```
...
<div className="unchanged">
- <div className="foo" />
+ <div className="bar">…</div>
+ <span />
</div>
...
<div className="another_unchanged">
- <span />
</div>
...
```

This strategy won't yield perfect results because if we're asynchronously hydrating, and it gets interrupted by another tree, we'll flush a warning before the actual hydrating particular tree is flushed. So we might show a partial diff in that case. This is probably. It's just a warning.

akshaynaik404:
Hi @sebmarkbage, Hi wanted to work on this issue, but I couldn't understand the info you gave above, also I am a beginner in React.

Is this issue suitable for me to is there some docs that I should read to get up to speed with this codebase ...

sebmarkbage:
A general guide is here: https://facebook.github.io/react/contributing/how-to-contribute.html

For this particular issue you'll probably want to most test it in the SSR fixture. It's a bit messy because you have to build React itself in the root before testing in fixture.

https://github.com/facebook/react/tree/master/fixtures/ssr

This issue is certainly not easy to get right but also doesn't that much existing knowledge of the code base. Just how React works with SSR.

stepnivlk:
Hey, @sebmarkbage I'm trying to work on it. I'll probably have some questions, but overall it looks manageable.

probablyup:
[...]

PR #13602, opened 2018-09-09
Hydration warnings with DOM diff in react-reconciler (#10085)

sompylasar:
An attempt to implement the hydration warning and DOM diff in react-reconciler package as requested by @gaearon here: https://github.com/facebook/react/pull/12063#pullrequestreview-142612751

> What would diffing look like if it was implemented in the reconciler? Feel free to change host config API if necessary — the goal is that all hydratable renderers would get nice warnings, not just DOM.

The tests and fixtures were copied from the original PR: https://github.com/facebook/react/pull/12063 and slightly modified (see my review below).

### DIFF
--- a/packages/react-dom/src/__tests__/ReactMount-test.js
+++ b/packages/react-dom/src/__tests__/ReactMount-test.js
@@ -123,20 +123,29 @@ describe('ReactMount', () => {
 
   it('should warn if mounting into left padded rendered markup', () => {
     const container = document.createElement('container');
-    container.innerHTML = ReactDOMServer.renderToString(<div />) + ' ';
+    container.innerHTML = ' ' + ReactDOMServer.renderToString(<div />);
 
     expect(() => ReactDOM.hydrate(<div />, container)).toWarnDev(
-      'Did not expect server HTML to contain the text node " " in <container>.',
-      {withoutStack: true},
+      "Warning: Did not expect server HTML to contain the text node {' '} in <container>.\n\n" +
+        '  <container>\n' +
+        "-   {' '}\n" +
+        '    <div data-reactroot=""></div>\n' +
+        '  </container>\n\n' +
+        '    in div (at **)',
     );
   });
 
   it('should warn if mounting into right padded rendered markup', () => {
     const container = document.createElement('container');
-    container.innerHTML = ' ' + ReactDOMServer.renderToString(<div />);
+    container.innerHTML = ReactDOMServer.renderToString(<div />) + ' ';
 
     expect(() => ReactDOM.hydrate(<div />, container)).toWarnDev(
-      'Did not expect server HTML to contain the text node " " in <container>.',
+      "Warning: Did not expect server HTML to contain the text node {' '} in <container>.\n\n" +
+        '  <container>\n' +
+        '    <div data-reactroot=""></div>\n' +
+        "-   {' '}\n" +
+        '  </container>\n',
+      // Without the component stack here because it's empty: found an unexpected text node directly in the root node.
       {withoutStack: true},
     );
   });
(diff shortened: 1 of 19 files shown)


sompylasar:
Hi @gaearon , quick question: do you guys override the `didNotFindHydratable*` and similar functions in `HostConfig` by something else in non-public codebases? I'm about to pull them out of `ReactDOMHostConfig` into `react-reconciler` to have them available to other potential hydratable renderers as per your statement:

> Feel free to change host config API if necessary — the goal is that all hydratable renderers would get nice warnings, not just DOM.

What do you think? 
The full list is:
```
  didNotMatchHydratedContainerTextInstance,
  didNotMatchHydratedTextInstance,
  didNotHydrateContainerInstance,
  didNotHydrateInstance,
  didNotFindHydratableContainerInstance,
  didNotFindHydratableContainerTextInstance,
  didNotFindHydratableInstance,
  didNotFindHydratableTextInstance,
```
Thanks.

gaearon:
>do you guys override the didNotFindHydratable* and similar functions in HostConfig by something else in non-public codebases? 

No.

> I'm about to pull them out of ReactDOMHostConfig into react-reconciler to have them available to other potential hydratable renderers as per your statement:

Sounds good

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -160,6 +169,306 @@ describe('ReactMount', () => {
...
+      ),
+    ).toWarnDev(
+      'Warning: Expected server HTML to contain a matching <h2> in <div>.\n\n' +
+        '  <div>\n' +
+        "    {'nested'}\n" +
+        '    <!-- -->\n' +
+        "    {'   '}\n" +
+        '-   <h1>children <b>text</b></h1>\n' +
+        '+   <h2>children <b>text</b></h2>\n' +
+        '    <div data-ssr-mismatch-padding-after="1" />\n' +
```
sompylasar: Changes to the diff printer compared to PR #12063:
1) always prints elements with no children as self-closing, assuming the diff output is renderer-agnostic and is presented in JSX-like syntax, not in the underlying HTML syntax; previously the diff printer had access to `omittedCloseTags` from `react-dom`, now the diff printer is renderer-agnostic so cannot rely on that.
2) applies some logic to determine if plain strings should be printed in curly braces and quotes, or as plain text (see `printChildrenValue` and `shouldPrintStringAsRawUnescapedText` in `ReactFiberHydrationWarning.js`) to produce more natural, less cluttered output.

No change:
- the HTML comment nodes and other non-hydratable host nodes are still printed out to add more context to the diff (`react-reconciler` asks `react-dom` to print them via a `ReactDOMHostConfig` function); this is debatable, you may want to remove them to make the output JSX-only, and simplify some code.

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -173,12 +482,632 @@ describe('ReactMount', () => {
...
+      ReactDOM.hydrate(
+        <div data-ssr-extra-prop={true} data-ssr-extra-prop-2={true}>
+          <em>SSRMismatchTest default text</em>
+        </div>,
+        div,
+      ),
+    ).toWarnDev(
+      'Warning: Prop `data-ssr-extra-prop` did not match. ' +
+        'Server: null ' +
+        'Client: true\n' +
```
sompylasar: Change to the warning message compared to PR #12063:
- the values are not converted to strings before being printed with `JSON.stringify`, so they no longer appear in quotes if they are not strings

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -173,12 +482,632 @@ describe('ReactMount', () => {
...
+
+    expect(() =>
+      ReactDOM.hydrate(
+        <div>
+          <em>SSRMismatchTest default text</em>
+        </div>,
+        div,
+      ),
+    ).toWarnDev(
+      'Warning: Extra attributes from the server: data-ssr-extra-prop,data-ssr-extra-prop-2\n' +
```
sompylasar: TODO: Print the attribute names array nicer, with a space after comma. Now the default array toString is used implicitly.

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -173,12 +482,632 @@ describe('ReactMount', () => {
...
+    div.innerHTML = markup;
+
+    expect(() =>
+      ReactDOM.hydrate('SSRMismatchTest default text', div),
+    ).toWarnDev(
+      'Warning: Expected server HTML to contain a matching text node' +
+        " for {'SSRMismatchTest default text'} in <div>.\n\n" +
+        '  <div>\n' +
+        '-   <span data-reactroot="">SSRMismatchTest default text</span>\n' +
+        "+   {'SSRMismatchTest default text'}\n" +
```
sompylasar: I decided to always print the top-level plain text children as strings in curly braces, but this can be changed with a single boolean flip, see `printElementOrTextForHydrationDiff`.

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -173,12 +482,632 @@ describe('ReactMount', () => {
...
+        <div>
+          <span />
+          <div />
+        </div>,
+        div,
+      ),
+    ).toWarnDev(
+      'Warning: Expected server HTML to contain a matching <span> in <div>.\n\n' +
+        '  <div>\n' +
+        '    <!-- a comment <-- --&gt; > -->\n' +
```
sompylasar: Here I test that the HTML comment is printed the same way it would've been written in the original markup, `&gt;` ensures the HTML comment is not ended prematurely.

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -173,12 +482,632 @@ describe('ReactMount', () => {
...
+      document.createProcessingInstruction(
+        'dom-processing-instruction',
+        'content > ',
+      ),
+    );
+
+    expect(() => ReactDOM.hydrate(<div />, xml)).toWarnDev(
+      'Warning: Expected server HTML to contain a matching <div> in <xml>.\n\n' +
+        '  <xml>\n' +
+        '    <?dom-processing-instruction content &gt; ?>\n' +
```
sompylasar: Same as for HTML comments, but for other non-hydratable nodes, escape the `>` with `&gt;` in the diff.

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -173,12 +482,632 @@ describe('ReactMount', () => {
...
+          <span />
+          {'SSRMismatchTest special characters: \'"\t\n'}
+        </div>,
+        div,
+      ),
+    ).toWarnDev(
+      // TODO: Currently, the special characters in the JSX string output are not escaped, the output looks invalid.
+      'Warning: Expected server HTML to contain a matching <span> in <div>.\n\n' +
+        '  <div data-reactroot="">\n' +
+        "-   {'SSRMismatchTest special characters: '\"\t\n'}\n" +
```
sompylasar: TODO: Currently, the special characters in the JSX string output are not escaped, the output looks invalid.

Despite there's code that should escape them (`escapeNonPrintableCharacters`). Need to look into it once again.

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -173,12 +482,632 @@ describe('ReactMount', () => {
...
+          <div>{'SSRMismatchTest text'}</div>
+        </div>,
+        div,
+      ),
+    ).toWarnDev(
+      // TODO: Currently, the special characters in the HTML string output are not escaped, the output looks invalid.
+      'Warning: Expected server HTML to contain a matching <span> in <div>.\n\n' +
+        `  <div data-ssr-mismatch-attribute-with-special-characters={'"'} data-reactroot="">\n` +
+        '-   <div>SSRMismatchTest text</div>\n' +
+        "+   <span>{' — <div>'}</span>\n" +
```
sompylasar: The special logic for the plain versus quoted strings triggered here: the `<span>` contains text that contains `<` and `>`, the characters that are special in the JSX markup, so they need to be escaped (quoted in a JavaScript string, wrapped in curly braces).

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -173,12 +482,632 @@ describe('ReactMount', () => {
...
+        <div data-ssr-mismatch-attribute-with-special-characters={'"'}>
+          <span>&nbsp;&mdash; &lt;div&gt;</span>
+          <div>{'SSRMismatchTest text'}</div>
+        </div>,
+        div,
+      ),
+    ).toWarnDev(
+      // TODO: Currently, the special characters in the HTML string output are not escaped, the output looks invalid.
+      'Warning: Expected server HTML to contain a matching <span> in <div>.\n\n' +
+        `  <div data-ssr-mismatch-attribute-with-special-characters={'"'} data-reactroot="">\n` +
```
sompylasar: The special logic for the plain versus quoted strings triggered here: the `data-ssr-mismatch-attribute-with-special-characters` attribute contains the double quote `"` which is special in the JSX attribute markup, so it needs to be escaped (quoted in a JavaScript string, wrapped in curly braces).

packages/react-dom/src/__tests__/ReactRenderDocument-test.js
```diff
@@ -203,9 +203,9 @@ describe('rendering React components at document', () => {
             'with ReactDOM.hydrate() if you want React to attach to the server HTML.',
           {withoutStack: true},
         );
-      }).toWarnDev('Warning: Text content did not match.', {
-        withoutStack: true,
-      });
+      }).toWarnDev(
+        'Warning: Text content did not match. Server: "Goodbye world" Client: "Hello world"',
```
sompylasar: Here both texts in the warning are added to the test. They are printed as quoted strings, the result of `JSON.stringify`.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -9,10 +9,15 @@
 
 // TODO: direct imports like some-package/src/* are bad. Fix me.
 import {getCurrentFiberOwnerNameInDevOrNull} from 'react-reconciler/src/ReactCurrentFiber';
+import {
+  didNotMatchHydratedChildrenPropValue,
+  didNotMatchHydratedPropValue,
+  didNotMatchHydratedPropsHostInstanceHasExtraAttributes,
+  SUPPRESS_HYDRATION_WARNING,
+} from 'react-reconciler/src/ReactFiberHydrationWarning';
```
sompylasar: Here `react-dom` `ReactDOMComponent` imports from `react-reconciler` to issue specific warnings not covered within the reconciler itself.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -955,20 +895,32 @@ export function diffHydratedProperties(
       // satisfies our requirement. Our requirement is not to produce perfect
       // HTML and attributes. Ideally we should preserve structure but it's
       // ok not to if the visible content is still enough to indicate what
-      // even listeners these nodes might be wired up to.
+      // event listeners these nodes might be wired up to.
```
sompylasar: Looks like there was a typo here.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -842,7 +783,6 @@ export function diffHydratedProperties(
   let extraAttributeNames: Set<string>;
 
   if (__DEV__) {
-    suppressHydrationWarning = rawProps[SUPPRESS_HYDRATION_WARNING] === true;
```
sompylasar: The `suppressHydrationWarning` variable was file-level for some reason, not function-level, so it may have affected functions unrelated to this function call. Now its expression `rawProps[SUPPRESS_HYDRATION_WARNING] === true` is only used in one place within this function, so the variable was removed.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -955,20 +895,32 @@ export function diffHydratedProperties(
...
       // ok not to if the visible content is still enough to indicate what
-      // even listeners these nodes might be wired up to.
+      // event listeners these nodes might be wired up to.
       // TODO: Warn if there is more than a single textNode as a child.
       // TODO: Should we use domElement.firstChild.nodeValue to compare?
       if (typeof nextProp === 'string') {
         if (domElement.textContent !== nextProp) {
-          if (__DEV__ && !suppressHydrationWarning) {
-            warnForTextDifference(domElement.textContent, nextProp);
+          if (__DEV__) {
```
sompylasar: Now the check for `rawProps[SUPPRESS_HYDRATION_WARNING] === true` is performed within `didNotMatchHydratedChildrenPropValue` and similar functions provided by `react-reconciler` `ReactFiberHydrationWarning`, so no need to check it here. Still checking `__DEV__` to ensure this call is not bundled into production build.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -988,7 +940,7 @@ export function diffHydratedProperties(
       // Validate that the properties correspond to their expected values.
       let serverValue;
       const propertyInfo = getPropertyInfo(propKey);
-      if (suppressHydrationWarning) {
+      if (rawProps[SUPPRESS_HYDRATION_WARNING] === true) {
         // Don't bother comparing. We're ignoring all these warnings.
```
sompylasar: We still need to check `rawProps[SUPPRESS_HYDRATION_WARNING] === true` here as per the comment, to skip the heavy comparison code below if the warning was suppressed.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -955,20 +895,32 @@ export function diffHydratedProperties(
...
       // TODO: Should we use domElement.firstChild.nodeValue to compare?
       if (typeof nextProp === 'string') {
         if (domElement.textContent !== nextProp) {
-          if (__DEV__ && !suppressHydrationWarning) {
-            warnForTextDifference(domElement.textContent, nextProp);
+          if (__DEV__) {
+            didNotMatchHydratedChildrenPropValue(
+              tag,
+              rawProps,
+              (domElement: any),
```
sompylasar: This `(domElement: any)` typecasting is ugly, but it was required to convince Flow that `Element` is compatible with `Instance` when typechecking with renderers other than `dom`, such as `yarn flow native`, where `react-dom` should not be used at all, but alas it's still typechecked.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1239,3 +1138,5 @@ export function restoreControlledState(
       return;
   }
 }
+
+export {normalizeHTMLTextOrAttributeValue};
```
sompylasar: This `normalizeHTMLTextOrAttributeValue` function is used in `ReactDOMHostConfig`, so exporting it here. It's not a `function` declaration because the function implementation gets substituted in `__DEV__`.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -47,7 +44,8 @@ export type Props = {
 export type Container = Element | Document;
 export type Instance = Element;
 export type TextInstance = Text;
-export type HydratableInstance = Element | Text;
+export type HydratableInstance = Instance | TextInstance;
+export type HostInstance = HydratableInstance | Container | Node;
```
sompylasar: The new type `HostInstance` means all possible host node types (like comment nodes and other non-hydratable nodes) that the hydration process can potentially meet during its DOM traversal, not just `Element` (`Instance`) and `Text` (`TextInstance`).

------

❓I still haven't figured out how to make `yarn flow custom` typecheck succeed. The assumption is that `HostInstance` is at least a union of `HydratableInstance | Container`, but there is no strict guarantee that it is defined like that. Please advice, this is the only remaining thing that fails in CI now. ❓

packages/react-reconciler/src/ReactFiberHydrationWarning.js
```diff
@@ -0,0 +1,854 @@
...
+  }
+}
+
+function printElementOrText(
+  instanceDisplayName: string | null,
+  instanceProps: Object | null,
+  instanceTextOrHTML: string | DangerouslySetInnerHTMLString | null,
+  openingTagOnly?: boolean,
+  withinParent?: boolean,
+): string {
```
sompylasar: This instance printing code happened to be more complex than I expected, but this all for the sake of beautiful and natural-looking diffs.

sompylasar:
@gaearon Please take a look.

packages/react-dom/src/__tests__/ReactMount-test.js
sompylasar: > TODO: Currently, the special characters in the JSX string output are not escaped, the output looks invalid.
>
> Despite there's code that should escape them (`escapeNonPrintableCharacters`). Need to look into it once again.

Fixed in https://github.com/facebook/react/pull/13602/commits/c0b778097f0af2c950e8de5bac3abb244e195a08

packages/react-dom/src/__tests__/ReactMount-test.js
sompylasar: > TODO: Print the attribute names array nicer, with a space after comma. Now the default array toString is used implicitly.

Fixed in https://github.com/facebook/react/pull/13602/commits/9c8093f3516b695239620d128420e9870ec96787

packages/react-dom/src/__tests__/ReactMount-test.js
sompylasar: Refactored this flag to be an enum value, `PRINT_CHILDREN_VALUE_MODE_DEFAULT` or `PRINT_CHILDREN_VALUE_MODE_RAW` (see `getHydrationDiff`).

packages/react-dom/src/client/ReactDOMHostConfig.js
sompylasar: > The new type `HostInstance` means all possible host node types (like comment nodes and other non-hydratable nodes) that the hydration process can potentially meet during its DOM traversal, not just `Element` (`Instance`) and `Text` (`TextInstance`).
> 
> ------
> 
> ❓I still haven't figured out how to make `yarn flow custom` typecheck succeed. The assumption is that `HostInstance` is at least a union of `HydratableInstance | Container`, but there is no strict guarantee that it is defined like that. Please advice, this is the only remaining thing that fails in CI now. ❓

I decided to simplify the code and remove printing of non-hydratable nodes. This refactor resulted in removing `HostInstance` type as unnecessary, and removing `HydratableInstance` type (inlining the alias). See https://github.com/facebook/react/pull/13602/commits/3ad9c5ff4fcc385ba62689bab9d0c423b944185f description for details.

sompylasar:
@gaearon I resolved the outstanding issues by removing the requirement to print in the diff the non-hydratable nodes like HTML comments and other weird nodes that should not affect the hydration. This resulted in simpler `ReactFiberHydrationWarning` and `ReactDOMHostConfig` code, and also simplified the `ReactFiberHostConfig` API (it still needs to expose a couple more functions to pull out tag names and attributes from host instances). See https://github.com/facebook/react/pull/13602/commits/3ad9c5ff4fcc385ba62689bab9d0c423b944185f

packages/react-dom/src/__tests__/ReactMount-test.js
```diff
@@ -160,6 +169,303 @@ describe('ReactMount', () => {
...
+    const div = document.createElement('div');
+    const markup = ReactDOMServer.renderToString('server text');
+    div.innerHTML = markup;
+
+    expect(() => ReactDOM.hydrate('client text', div)).toWarnDev(
+      'Text content did not match. ' +
+        'Server: "server text" ' +
+        'Client: "client text"',
+      // Without the component stack here because it's empty: rendering a text node directly into the root node.
+      {withoutStack: true},
```
sompylasar: > Without the component stack here because it's empty: rendering a text node directly into the root node

A thought came to me here: why doesn't React show in the stack the place where the `render` or `hydrate` was called? This would be 1) consistent, there'd always be a stack; 2) useful for apps with multiple render entry points.

sompylasar:
High-level overview (asked by @gaearon via Twitter DM).

## BEFORE

hydration warnings with dom diff in react-reconciler BEFORE

## AFTER

hydration warnings with dom diff in react-reconciler AFTER

sompylasar:
A visualization of the **`findHydrationWarningHostInstanceIndex`** algorithm using the render structure from the test in `packages/react-dom/src/__tests__/ReactMount-test.js`
```js
it('should warn when hydrate replaces an element within server-rendered nested components (replacement diff)')
``` 

hydration warnings with dom diff in react-reconciler tree

sompylasar:
👆Rebased onto current `facebook/master` https://github.com/facebook/react/commit/275e76e83bc2be5dc0be9185ff747f383969289a

oyeanuj:
@gaearon @sompylasar Is this ready to merge? This would be a huge boon in debugging SSR errors!

sebmarkbage:
How is this approach affected by Partial Hydration? https://github.com/facebook/react/pull/14717 and https://github.com/facebook/react/pull/14884

It seems like this approach just logs the deltas as they are encountered (as opposed to putting together a log of all the differences at the etc) so maybe that's fine?

It would probably need to consider that the HTML might not be the final HTML yet.

sompylasar:
@sebmarkbage Thanks for looking into this and linking to your Partial Hydration work!

I noticed some of it before and how it slightly conflicts with my changes related to Flow typings, but I'll need to get a closer look and get some integrated tests together with Partial Hydration to answer your question.

This approach builds a diff between the Fiber tree (leaf host nodes) and the DOM tree synchronously at the time the hydration mismatch is encountered. In my current understanding, with Partial Hydration some Fibers won't have leaf host nodes where they might be present in the DOM, so the sequential diff may even not be possible to produce as the sizes of the holes under suspended hydration that we'd need to skip in the actual DOM would be unknown until it's rendered.

I'll try to find some time within the next few weeks after I'm back from 🏖, but no promises as I'll be busy with learning new things on the new job (haven't yet announced publicly 🙊).

mxstbr:
Is there any way I can test this PR on our production codebase?

gaearon:
You can build it and try it? 🙂 Not sure it there's anything specific that you are asking.

sompylasar:
👆Rebased onto current `facebook/master` https://github.com/facebook/react/commit/81a61b1d1ac4c3fa974a822382b2343a96f7cc31

sompylasar:
Cannot run `fixtures/ssr`, getting `Invariant Violation: ReactDOMServer does not yet support Suspense.` because `enableSuspenseServerRenderer` feature flag is not enabled on the React build that is used by the fixture.

sompylasar:
👆Rebased onto current `facebook/master` https://github.com/facebook/react/commit/d584fcdc6e10b9f4d4e9eec775731e7ffc0fb501

gaearon:
The screenshot above seems very verbose. Seems like we should somehow clip the length in the middle? I'm not sure all this info is valuable.

sompylasar:
It says "did not expect `<main>`" so this means server rendered full, client rendered empty.

I agree that it's too much and can be clipped. As you may notice, the attribute values are clipped, but not elements themselves. Usually the difference is not so dramatic.

sompylasar:
Also the first line phrasing "dis not expect" was kept from the original implementation. We can think of a better phrasing to make it clearer.

In the code there's also a TODO to think about phrasing in terms of universal hydration, not just server/client.

sompylasar:
Hey stale[bot], I still have hopes to ship this some day!

gaearon:
Thanks a lot for exploring this. The scope ended up being larger than we expected, and we ended up with a simpler (although perhaps less precise) solution: https://github.com/facebook/react/issues/10085#issuecomment-591650861. But this was still a very useful exploration.

gaearon:
We're adding diffs in https://github.com/facebook/react/pull/28512.

PR #13605, opened 2018-09-09
schedule/tracking (production bundle) should not cause react-dom (profiling bundle) to error

bvaughn:
This was an oversight on my part. `schedule/tracking` production bundle didn't initialize its refs (to be slightly smaller and faster). I hadn't really considered the use case a `react-dom` profiling bundle importing a `schedule/tracking` production (non-profiling) bundle.

This fixes it by changing `schedule/tracking` production bundle to also initialize the refs. This should not impact `react-dom` production bundle because it doesn't even require `schedule/tracking`.

We could also add a _warning_ about this mismatch to the `react-dom` profiling bundle, but...we don't normally have warnings in production bundles.


gaearon:
Hmm. What is the recommendation for using profile mode? Is it to only use profiling bundle of ReactDOM? Or also to use profiling bundle of `schedule`?

If profiling needs profiling bundle for `schedule`, then it seems like we need a "blessed" entry point for it too. In other words we should have a profiling entry point for any package that needs it in profiling mode.

react-linter:
<details>

  <p>Comparing: 7d1169b2d7ad8d2345c28a2063bc101a4cc2d570...ecc6fea5614a01b272498fb30077208faac2540f</p>


  
## schedule
schedule-tracking.development.js | -1.9% | -0.4% | 10.64 KB | 10.43 KB | 2.51 KB | 2.5 KB | NODE_DEV
**schedule-tracking.production.min.js** | **:small_red_triangle:+3.8%** | **:small_red_triangle:+5.6%** | **719 B** | **746 B** | **374 B** | **395 B** | **NODE_PROD**
schedule-tracking.profiling.min.js | -13.6% | -1.8% | 3.26 KB | 2.82 KB | 991 B | 973 B | NODE_PROFILING
ScheduleTracking-dev.js | -2.0% | -0.5% | 10.29 KB | 10.08 KB | 2.25 KB | 2.24 KB | FB_WWW_DEV
**ScheduleTracking-prod.js** | **:small_red_triangle:+7.6%** | **:small_red_triangle:+4.9%** | **899 B** | **967 B** | **425 B** | **446 B** | **FB_WWW_PROD**
ScheduleTracking-profiling.js | -2.6% | -0.7% | 6.82 KB | 6.64 KB | 1.23 KB | 1.22 KB | FB_WWW_PROFILING

  </details>
  
<p align="right">
</p>

bvaughn:
Yeah, I agree. First commit was just focused on fixing the runtime error. I'm adding another entry point now to the `schedule` package.

gaearon:
I was thinking that maybe in this case the error is justified? While the error itself is confusing, you can google it and discover it comes from mixing profiling and non-profiling bundles. Whereas if we don't error, you would just be unaware they're inconsistent?

bvaughn:
I've added a blessed entry-point for production+profiling `schedule/tracking` – although the name kind of sucks 😄 

Verified it works by ejecting a CRA app and modifying the Webpack config to add the following aliases:
```js
resolve: {
  alias: {
    'react-dom': 'react-dom/profiling',
    'schedule/tracking': 'schedule/tracking-profiling',
  },
}
```

bvaughn:
> I was thinking that maybe in this case the error is justified? While the error itself is confusing, you can google it and discover it comes from mixing profiling and non-profiling bundles. Whereas if we don't error, you would just be unaware they're inconsistent?

That's fair, unless we added a warning– you would only know something was wrong because of the fact that none of your interactions would be tracked.

bvaughn:
I don't feel strongly about this. If you have a strong intuition that one route is better, I'll defer.

Reverting my initial commit would certainly be less subtle, and so maybe also less confusing.

gaearon:
I think my preferred approach would be:

* No code in production schedule that's unnecessary in production (like those Set allocations)
* We have an explicit entry point for profiling version of schedule, the name you picked sounds fine
* We add an invariant with a nice message to the *profiling* build of ReactDOM for cases if it's used with non-profiling schedule, and ask to alias them both

Seems like it would handle all cases and avoid extra code.

bvaughn:
> Seems like it would handle all cases and avoid extra code

Except for the `invariant` itself, but if we're willing to add that– I agree that's probably the nicest experience.

bvaughn:
True!

bvaughn:
Okay. I've added an invariant and confirmed that it works with an ejected CRA. Would love input on the error message though. I'm not very good at writing these.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -168,6 +168,16 @@ let didWarnSetStateChildContext;
 let warnAboutUpdateOnUnmounted;
 let warnAboutInvalidUpdates;
 
+if (enableSchedulerTracking) {
+  // Provide explicit error message when production+profiling bundle of e.g. react-dom
+  // Is used with production (non-profiling) bundle of schedule/tracking
+  invariant(
+    __interactionsRef != null && __interactionsRef.current != null,
+    "Profiling renderer (e.g. 'react-dom/profiling') cannot be used with non-profiling 'schedule/tracking'. " +
+      "To fix, alias 'schedule/tracking' to 'schedule/tracking-profiling'.",
```
gaearon: This would be bundler specific and I’m not sure they all use the “alias” wording.

Another thing to consider is that the person seeing this might be unaware that schedule package exists or what it is.

How about 

```
It is not supported to run the profiling version of a renderer (for example, `react-dom/profiling`) without also replacing the `schedule/tracking` module with `schedule/tracking-profiling`. Your bundler might have a setting for aliasing both modules. Learn more at http://fb.me/react-profiling
```

The link points to a gist explaining how to do it in webpack. Later we change it to point to a blog post.

bvaughn: I dig it. 👍 Thanks for the ideas Dan.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -168,6 +168,18 @@ let didWarnSetStateChildContext;
 let warnAboutUpdateOnUnmounted;
 let warnAboutInvalidUpdates;
 
+if (enableSchedulerTracking) {
+  // Provide explicit error message when production+profiling bundle of e.g. react-dom
+  // Is used with production (non-profiling) bundle of schedule/tracking
```
gaearon: Nit: does this line need to start with capital letter? Made reading the sentence a bit confusing.

bvaughn: No, I can change the comment 😆

PR #13606, opened 2018-09-09
Enable no-use-before-define rule

gaearon:
This would have caught https://github.com/facebook/react/pull/13582#discussion_r216163970.

For now, I had to add suppressions for cyclical Flow references but we can remove this after https://github.com/babel/babel-eslint/pull/584 is merged and released.

I also enabled it for classes. Maybe this is excessive but I figured it doesn't hurt since classes aren't hoisted. I had to reorder two places because of that but it's just copy paste.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -37,6 +37,7 @@ module.exports = {
     'no-shadow': ERROR,
     'no-unused-expressions': ERROR,
     'no-unused-vars': [ERROR, {args: 'none'}],
+    'no-use-before-define': [ERROR, {functions: false, variables: false}],
     'no-useless-concat': OFF,
     'quotes': [ERROR, 'single', {avoidEscape: true, allowTemplateLiterals: true }],
     'space-before-blocks': ERROR,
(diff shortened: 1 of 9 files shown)


.eslintrc.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
     'no-shadow': ERROR,
     'no-unused-expressions': ERROR,
     'no-unused-vars': [ERROR, {args: 'none'}],
+    'no-use-before-define': [ERROR, {functions: false, variables: false}],
```
gaearon: Naming of these options is confusing. It basically means "still allow to rely on function hoisting, and don't complain about variables *unless* they are in the same scope".

react-linter:
<details>

  <p>Comparing: 8a8d973d3cc5623676a84f87af66ef9259c3937c...7e52e2f3a8e64daf49d933d0f731cba209fa47df</p>


  
## schedule

  </details>
  
<p align="right">
</p>

.eslintrc.js
sompylasar: > Naming of these options is confusing. It basically means "still allow to rely on function hoisting, and don't complain about variables unless they are in the same scope".

If `.eslintrc.js` is a `.js` would it make sense to put this explanation in a comment next to the added option?

Issue #12615, opened 2018-04-15
Unexpected error when hydrating with portal and SSR


majelbstoat:
**Do you want to request a *feature* or report a *bug*?**

*bug*

**What is the current behavior?**

Given the following (simplified) snippet:

```jsx
class HoverMenu extends React.Component {
  render() {
    if (typeof document === 'undefined') return null
    const root = document.getElementById('root')
    return ReactDOM.createPortal(<div>Hello World</div>, root)
  }
}

class Para extends React.Component {
  render() {
    return (
      <span>
        Some Text
        <HoverMenu />
      </span>
    )
  }
} 
```

where `div#root` is a valid `div` that exists, the following error is shown when hydrating after SSR:

`Warning: Expected server HTML to contain a matching <div> in <span>`

The warning goes away if I update the definition of `HoverMenu` to:

```jsx
class HoverMenu extends React.Component {
  componentDidMount() {
    this.setState({ isActive: true })
  }
  render() {
    const { isActive} = this.state
    if (!isActive) return null
    const root = document.getElementById('root')
    return ReactDOM.createPortal(<div>Hello World</div>, root)
  }
}
```

I'd prefer not to do that because of the double rendering caused by `setState` in `componentDidMount`.

I don't quite understand what that error is telling me. No `<div />` is rendered server-side in either case. The error is particularly confusing, as the `HoverMenu` DOM `div` is not even rendered inside a DOM `span`. (I wonder if this is happening because `HoverMenu` is nested inside a React `span`.)

**What is the expected behavior?**

No error is thrown. Or, at least that the error message is clearer.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

Chrome 65
React 16.2
(SSR through Next 5.1)

interphx:
I have a similar issue, which can also be solved by re-rendering on the client via setState.
In my case, I try to render a modal inside a portal. The `Modal` component returns `null` when rendered on the server and creates a portal on the client. However, the DOM gets messed up after hydrating. 

E.g. if I use it like this inside my main component:

```
<Modal>
This is a test
</Modal>

<div className="some-div-after-the-modal">
</div>
```

Instead of this:

```
<!-- Inside the portal container -->
<div class="modal-wrapper">
    <div class="modal-content">This is a test</div>
</div>

<!-- In the main component -->
<div className="some-div-after-the-modal">
</div>
```


I get this:


```
<!-- Inside the portal container -->
<div class="some-div-after-the-modal">
[...]

PR #13614, opened 2018-09-11
Bug/invariant when hydrate portal #12615

brandonvilla21:
I was unable to run `yarn flow` locally. I'll check for error/warnings displayed when CI executes type checking. This pull request is related to #12615 issue

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -45,6 +45,7 @@ import {
   DOCUMENT_FRAGMENT_NODE,
 } from '../shared/HTMLNodeType';
 import {ROOT_ATTRIBUTE_NAME} from '../shared/DOMProperty';
+import {REACT_PORTAL_TYPE} from 'shared/ReactSymbols';
 
 const ReactCurrentOwner = ReactSharedInternals.ReactCurrentOwner;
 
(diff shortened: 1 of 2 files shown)


react-linter:
<details>

  <p>Comparing: f6fb03edffcc5918ad2f37ed45bdc7c0c3b3d199...1bace8a173aabf522cb9e6d2f30bac01d2a970bd</p>


  
## schedule

  </details>
  
<p align="right">
</p>

brandonvilla21:
@gaearon  Hi Dan!,

I'm getting a Type Check error
Look like I can't use `$$typeof` property from `element` object at `hydrate()` method due to this:

Sorry, I guess I'm not good enough with Flow yet, is this the right approach to compare both elements?

```javascript
Cannot get `element.$$typeof` because:
 - property `$$typeof` is missing in `$Iterable` [1].
 - property `$$typeof` is missing in `Boolean` [1].
 - property `$$typeof` is missing in `Number` [1].
 - property `$$typeof` is missing in `React.Element` [1].
 - property `$$typeof` is missing in `React.Portal` [1].
 - property `$$typeof` is missing in `String` [1].

   ../../../packages/react-dom/src/client/ReactDOM.js:622:17
   622|       element ? element.$$typeof !== Symbol.for('react.portal') : true,
                        ^^^^^^^^^^^^^^^^

References:
   ../../../packages/react-dom/src/client/ReactDOM.js:619:20
   619|   hydrate(element: React$Node, container: DOMContainer, callback: ?Function) {
                           ^^^^^^^^^^ [1]
```

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -45,6 +45,7 @@ import {
   DOCUMENT_FRAGMENT_NODE,
 } from '../shared/HTMLNodeType';
 import {ROOT_ATTRIBUTE_NAME} from '../shared/DOMProperty';
+import {REACT_PORTAL_TYPE} from 'shared/ReactSymbols';
```
PatrykDeja: @brandonvilla21
Isn't that supposed to be like that:  ` '../shared/ReactSymbols';`

brandonvilla21: @PatrykDeja 
I'm accessing to a more global `shared` folder in that way. I suppose the project is using a config to set an alias in order to access to that global shared folder located in `packages/shared`. Otherwise, if I use `'../shared'` I'll be accessing to the parent shared folder where is not the file that I want to reach

fluffels:
This check seems to pass for me locally now, perhaps this can be re-run to check it?

PR #13615, opened 2018-09-11
 honor displayName set on ForwardRef if available

probablyup:
Since React.forwardRef returns a component object, some users
(including styled-components and react-native) are starting to
decorate it with various statics including displayName.

This adjusts React's various name-getters to honor this if set and
surface the name in warnings and hopefully DevTools.

### DIFF
--- a/packages/react/src/ReactElementValidator.js
+++ b/packages/react/src/ReactElementValidator.js
@@ -197,7 +197,9 @@ function validatePropTypes(element) {
   ) {
     // ForwardRef
     const functionName = type.render.displayName || type.render.name || '';
-    name = functionName !== '' ? `ForwardRef(${functionName})` : 'ForwardRef';
+    name =
+      type.displayName ||
+      (functionName !== '' ? `ForwardRef(${functionName})` : 'ForwardRef');
     propTypes = type.propTypes;
   } else {
     return;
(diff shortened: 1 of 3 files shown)


react-linter:
<details>

  <p>Comparing: f6fb03edffcc5918ad2f37ed45bdc7c0c3b3d199...807aee50d80921d1ae84365d73a1aedcc599808f</p>


  
## schedule

  </details>
  
<p align="right">
</p>

gaearon:
@probablyup Can you describe your use case in more detail? What do you see in warnings/DevTools today, and what do you want to see instead?

probablyup:
@gaearon Sure.

So one of the currently nuisances around the forwardRef API is it creates an interim component that is shown in the DevTools, component trees, etc. For many libraries, it makes shipping a component that is readily identifiable in tools like enzyme impossible because the top level component is no longer ours, but ForwardRef. Thus shallow rendering becomes relatively useless and targeting quite difficult.

To get around this, various libraries ([styled-components](https://github.com/styled-components/styled-components/blob/30c372cf635e0db77f74c0aaacf60deda729581c/src/models/StyledComponent.js#L249-L311) and [react-native](https://github.com/facebook/react-native/commit/e708010d18f938e2d6b6424cfc9485d8e5dd2800#diff-e250436371d8e58356be701ad28b37caR268) that I've seen so far) have begun decorating the component returned by the forwardRef API with displayName, etc. This should allow a forwardRef-wrapped component to appear the same in tooling, while layering in that useful ref functionality.

This is all a workaround for the additional component layer. If forwardRef was available as a decorator or class extend target, I don't think sort of thing would be needed.

gaearon:
Sounds good. Thanks.

PR #12629, opened 2018-04-17
Use browser event names for top-level event types in React DOM

philipp-spiess:
This PR builds on top of the great work of @gaearon in #11894. The main idea is to get rid of all `top*` strings used to identify internal events and replace them with numerical values which the compiler can inline.

Before merging this, I have two other questions which I've also asked [here](https://github.com/facebook/react/pull/11894#issuecomment-381887113):

1. Does it really make sense to prefix all event constants with `TOP_`? From the comment in https://github.com/facebook/react/blob/master/packages/react-dom/src/events/BrowserEventConstants.js#L10-L16 it's clear that only _some_ of the events are actually listened to at the top level. However the list contains more than that subset (incl. `input`, `invalid`, `reset`, `submit`, and the media events). With that in mind, I think renaming the module to `EventTypes.js` or `InternalEventTypes.js` would be more fitting. What do you think? I know that currently, the `top` prefix is used everywhere - Is there a reason for that? 
2. I had to make some worrisome changes to modules that contain `Native` in their name. Specifically when working on the Flow types [here](https://github.com/facebook/react/blob/master/packages/react-native-renderer/src/ReactNativeBridgeEventPlugin.js#L32) and [here](https://github.com/facebook/react/blob/master/packages/react-native-renderer/src/ReactNativeEventEmitter.js#L91). Would it make sense to type the internal event type as `number | string` so we don't need to touch anything on the react-native side? And what part of the changed modules will also be used on the react-native side? Everything inside `packages/events`?

## State of this PR

Besides the open questions, this PR works fine for `react-dom`. The difference in the bundle size is significant (well, at least enough to land #12507 🙈):

- `react-dom.production.min.js  (UMD_PROD)` `-1.5%` pre, `-0.9%` post gzip
- `react-dom-test-utils.production.min.js (UMD_PROD)` `-13.2%` pre, `-10.2%` post gzip
- `react-dom-unstable-native-dependencies.production.min.js (UMD_PROD)` `-1.2%` pre, `-0.6%` post gzip

All dom related unit tests are green and the fixtures behave as described.

There's currently one failing unit test suite (`ReactNativeEvents-test.internal.js`). We'll probably need to clearly look at the react-native boundaries here to see how/if we can keep this change to the DOM side.

[...]

PR #9333, opened 2017-04-05
Fix wheel/touch browser locking in IE and Safari

nhunzaker:
In order for Firefox, Edge, and Safari to optimize scroll rendering, scroll events must be attached locally.

The React synthetic event system isn't really built to handle this, attaching events locally causes a dispatch to the React Synthetic event system, which is problematic for nested scroll event handlers. To work around that, this PR adds a deduping mechanism that track what native events have already dispatched, bailing early if they have already been processed

### Testing

I have forked @nolanlawson's wheel jank example and built it with this branch's version of React. You can view the difference by doing the following:

1. Visit http://nh-react-scroll-fix.surge.sh in IE Edge or Safari
2. Using a wheel, or two-finger scroll, scroll outer container of the page for ~10 seconds.
3. Your scrolling should never be interrupted (though in Safari you _will_ see the scroll bar lock up, the page still scrolls)

---

Related issues: https://github.com/facebook/react/issues/1254#issuecomment-289296440

PR #13618, opened 2018-09-11
Do not bind topLevelType to dispatch

nhunzaker:
This PR is exploratory, but it's been on my mind a long time and I wanted to get others' thoughts.

A previous change (https://github.com/facebook/react/pull/12629) made it such that all top level event types correspond to their associated native event string values. That means we don't need to bind that event name into `dispatchEvent`, letting all event listener code use either the standard dispatch function, or the variant for interactive updates.

### Why

My primary motivation with this PR isn't to necessarily merge this, but to talk about a few event system things.

#### Why drop bind()

Dropping this `.bind()` is attractive to me because I believe we could eventually remove the logic in`ReactDOMBrowserEventEmitter` to track what events have already attached:

https://github.com/facebook/react/blob/f6fb03edffcc5918ad2f37ed45bdc7c0c3b3d199/packages/react-dom/src/events/ReactBrowserEventEmitter.js#L95-L103

https://github.com/facebook/react/blob/f6fb03edffcc5918ad2f37ed45bdc7c0c3b3d199/packages/react-dom/src/events/ReactBrowserEventEmitter.js#L130-L135

When working on the scroll-jank PR (https://github.com/facebook/react/pull/9333), tracking needs to happen per-element for touch, wheel, and scroll events. That means saving state on each DOM node, and doing a lot of property access.

#### Could it be faster?

I'm curious about the performance characteristics of naively attaching the same event. It could avoid a lot of DOM storage and property access (event listener attachment tracking) if we start to attach listeners locally, since `addEventListener` is idempotent:

```javascript
let button = document.createElement('button')
let callback = () => console.log("click!")

button.addEventListener('click', callback)
button.addEventListener('click', callback)

button.click()
```

### Challenges

We can't remove event listener tracking because SelectEventPlugin depends on this tracking to see if it should fire:

https://github.com/facebook/react/blob/f6fb03edffcc5918ad2f37ed45bdc7c0c3b3d199/packages/react-dom/src/events/SelectEventPlugin.js#L173-L177

But we might be able to come up with another way to determine when SelectEventPlugin should fire. Maybe, **instead of enumerating all plugins at dispatch time, they could be eagerly applied on mount**. Then we also wouldn't have to run through all event plugins on every dispatch because the event plugins would have the context to know that their behavior was specifically applied.

---

Still, it's an interesting series of changes to me, and I wanted to get the thoughts of others.

### DIFF
--- a/packages/events/PluginModuleType.js
+++ b/packages/events/PluginModuleType.js
@@ -16,7 +16,7 @@ import type {TopLevelType} from './TopLevelEventTypes';
 
 export type EventTypes = {[key: string]: DispatchConfig};
 
-export type AnyNativeEvent = Event | KeyboardEvent | MouseEvent | Touch;
+export type AnyNativeEvent = Event | KeyboardEvent | MouseEvent | TouchEvent;
 
 export type PluginName = string;
 
--- a/packages/events/ReactGenericBatching.js
+++ b/packages/events/ReactGenericBatching.js
@@ -20,8 +20,8 @@ import {
 let _batchedUpdatesImpl = function(fn, bookkeeping) {
   return fn(bookkeeping);
 };
-let _interactiveUpdatesImpl = function(fn, a, b) {
-  return fn(a, b);
+let _interactiveUpdatesImpl = function(fn, a) {
+  return fn(a);
 };
 let _flushInteractiveUpdatesImpl = function() {};
 
@@ -52,8 +52,8 @@ export function batchedUpdates(fn, bookkeeping) {
   }
 }
 
-export function interactiveUpdates(fn, a, b) {
-  return _interactiveUpdatesImpl(fn, a, b);
+export function interactiveUpdates(fn, a) {
+  return _interactiveUpdatesImpl(fn, a);
 }
 
 export function flushInteractiveUpdates() {
--- a/packages/react-dom/src/events/ReactDOMEventListener.js
+++ b/packages/react-dom/src/events/ReactDOMEventListener.js
@@ -57,6 +57,7 @@ function getTopLevelCallbackBookKeeping(
   targetInst: Fiber | null,
   ancestors: Array<Fiber>,
 } {
+  // This is safe because DOMTopLevelTypes are always native event type strings
   const topLevelType = unsafeCastStringToDOMTopLevelType(nativeEvent.type);
 
   if (callbackBookkeepingPool.length) {
--- a/packages/react-dom/src/events/ReactDOMEventListener.js
+++ b/packages/react-dom/src/events/ReactDOMEventListener.js
@@ -144,16 +144,13 @@ export function trapBubbledEvent(
   if (!element) {
     return null;
   }
+
+  // Check if interactive and wrap in interactiveUpdates
   const dispatch = isInteractiveTopLevelEventType(topLevelType)
     ? dispatchInteractiveEvent
     : dispatchEvent;
 
-  addEventBubbleListener(
-    element,
-    getRawEventName(topLevelType),
-    // Check if interactive and wrap in interactiveUpdates
-    dispatch,
-  );
+  addEventBubbleListener(element, getRawEventName(topLevelType), dispatch);
 }
 
 /**
@@ -172,16 +169,13 @@ export function trapCapturedEvent(
   if (!element) {
     return null;
   }
+
+  // Check if interactive and wrap in interactiveUpdates
   const dispatch = isInteractiveTopLevelEventType(topLevelType)
     ? dispatchInteractiveEvent
     : dispatchEvent;
 
-  addEventCaptureListener(
-    element,
-    getRawEventName(topLevelType),
-    // Check if interactive and wrap in interactiveUpdates
-    dispatch,
-  );
+  addEventCaptureListener(element, getRawEventName(topLevelType), dispatch);
 }
 
 function dispatchInteractiveEvent(nativeEvent) {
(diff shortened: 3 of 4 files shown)


packages/events/PluginModuleType.js
```diff
@@ -16,7 +16,7 @@ import type {TopLevelType} from './TopLevelEventTypes';
 
 export type EventTypes = {[key: string]: DispatchConfig};
 
-export type AnyNativeEvent = Event | KeyboardEvent | MouseEvent | Touch;
+export type AnyNativeEvent = Event | KeyboardEvent | MouseEvent | TouchEvent;
```
nhunzaker: This should probably go straight to master. As far as I can tell, this is a bug.

packages/events/ReactGenericBatching.js
```diff
@@ -52,8 +52,8 @@ export function batchedUpdates(fn, bookkeeping) {
   }
 }
 
-export function interactiveUpdates(fn, a, b) {
-  return _interactiveUpdatesImpl(fn, a, b);
+export function interactiveUpdates(fn, a) {
+  return _interactiveUpdatesImpl(fn, a);
 }
```
nhunzaker: This possibly affects other renders 😨

react-linter:
<details>

  <p>Comparing: f6fb03edffcc5918ad2f37ed45bdc7c0c3b3d199...b31923544883c444328f03568fc040e72244e310</p>


  
## schedule

  </details>
  
<p align="right">
</p>

gaearon:
This by itself seems fine to me and is a bit simpler. What's the concern?

nhunzaker:
@gaearon No concerns on my end. I just didn't know if I missed something and wasn't going to push the change too hard.

packages/react-dom/src/test-utils/ReactTestUtils.js
```diff
@@ -42,7 +42,9 @@ const [
   runEventsInBatch,
 ] = ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Events;
 
-function Event(suffix) {}
+function Event(type) {
+  this.type = type;
```
philipp-spiess: I‘m a bit confused about this - where do we use this event constructor?

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -57,6 +57,8 @@ function getTopLevelCallbackBookKeeping(
   targetInst: Fiber | null,
   ancestors: Array<Fiber>,
 } {
+  const topLevelType = unsafeCastStringToDOMTopLevelType(nativeEvent.type);
```
gaearon: What's the guideline for when it's "safe" to call it? How do we know it's okay here?

packages/react-dom/src/test-utils/ReactTestUtils.js
nhunzaker: SimulateNativeEvent:

https://github.com/facebook/react/blob/f6fb03edffcc5918ad2f37ed45bdc7c0c3b3d199/packages/react-dom/src/test-utils/ReactTestUtils.js#L479-L498

packages/react-dom/src/events/ReactDOMEventListener.js
nhunzaker: I think it's safe here because we know the nativeEvent type is always a DOMTopLevelType. 

This is otherwise only used to cast the top level types themselves when they are created, like:

https://github.com/facebook/react/blob/f6fb03edffcc5918ad2f37ed45bdc7c0c3b3d199/packages/react-dom/src/events/DOMTopLevelEventTypes.js#L25-L31

I wonder if there is an existing concept of a DOM event's type string. We don't really have top level types in anymore: just the browser standard event names.

gaearon: Should have a comment explaining why then.

nhunzaker: Commented! Part of me wonders if these could be annotated using the actual event type declarations here:

https://github.com/facebook/flow/blob/08f854876cc72165f1f7a2e023bd8e79e8a7fa02/lib/dom.js#L144-L153

If that were the case, we could remove the concept of a top level type and just make it an DOM event type

packages/events/ReactGenericBatching.js
gaearon: Have you checked whether it does? `ReactGenericBatching` is only accessible in this repository.

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -149,7 +152,7 @@ export function trapBubbledEvent(
     element,
     getRawEventName(topLevelType),
     // Check if interactive and wrap in interactiveUpdates
```
gaearon: This comment should move above btw.

nhunzaker: Done in 60de292f6

packages/events/ReactGenericBatching.js
nhunzaker: This should not affect other renders. ReactGenericBatching.interactiveUpdates is only used inside of ReactDOMEventListener.

gaearon:
(For context: since @acdlite merged the Scheduler change we'd need to sync on Monday anyway. So this doesn't have to delay now.)

nhunzaker:
Haha. Thanks. I wonder if anyone thought I was going rogue 😅

PR #13619, opened 2018-09-11
Fix typo

NE-SmallTown:
> If we get lower than that,

IMO, the `that` refer to the '120hz' rather than the `8`, so I think use 'higher' here would be better.

### DIFF
--- a/scripts/rollup/results.json
+++ b/scripts/rollup/results.json
@@ -785,48 +785,6 @@
       "gzip": 36760
     },
     {
-      "filename": "scheduler-tracking.development.js",
-      "bundleType": "NODE_DEV",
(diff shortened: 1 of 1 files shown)


NE-SmallTown:
If I miss something, please let me know, thanks!

NE-SmallTown:
If I miss something, please let me know, thanks!

NE-SmallTown:
If I miss something, please let me know, thanks!

NE-SmallTown:
@gaearon Any update on this or I miss something?

NE-SmallTown:
I don't know why this be closed when I process the conflict :( I will send another pr and delete this branch

Issue #13610, opened 2018-09-10
IE9: Unable to call apply on console.error when encountering an unknown tag


nhunzaker:
This was introduced in React 16.5.0.

If IE9 encounters an unknown html element in development, React crashes as it attempts to call console.error.apply here:

https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/shared/warningWithoutStack.js#L30

This is reproduceable with the following CodePen:

**Code**
https://codepen.io/nhunzaker/pen/YOeZXv

**Demo**
https://s.codepen.io/nhunzaker/debug/YOeZXv/mWMoNxzPjbZk

gaearon:
What's the error?

nhunzaker:
The spread operator turns into apply like so:

```javascript
// Before: console.error('Warning: ' + format, ...stringArgs);
console.error.apply(console, ['Warning' + format].concat(stringArgs))
```

IE9 doesn't support `apply` on `console.error`:

<img width="666" alt="screen shot 2018-09-10 at 1 50 50 pm" src="https://user-images.githubusercontent.com/590904/45323618-8f15d400-b500-11e8-83b6-2adb1b104fef.png">

link-alex:
I can try to check this one as "good first issue" if you don't mind :)

gaearon:
Should be fixed in 16.5.1

PR #13620, opened 2018-09-11
Fix warning without stack for ie9

link-alex:
Where console methods like log, error etc. don't have 'apply' method.
Because of the lot of tests already expect that exactly console['method']
will be called - had to reapply references for console.error method

https://github.com/facebook/react/issues/13610

### DIFF
--- a/packages/shared/warningWithoutStack.js
+++ b/packages/shared/warningWithoutStack.js
@@ -22,65 +22,53 @@ if (__DEV__) {
           'message argument',
       );
     }
+    if (args.length > 8) {
+      // Check before the condition to catch violations early.
+      throw new Error(
+        'warningWithoutStack() currently supports at most 8 arguments.',
+      );
+    }
     if (condition) {
       return;
     }
     if (typeof console !== 'undefined') {
-      const arr = args.map(item => '' + item);
-      const arrLength = arr.length;
+      const [a, b, c, d, e, f, g, h] = args.map(item => '' + item);
+      const message = 'Warning: ' + format;
 
-      if (arrLength === 0) {
-        console.error('Warning: ' + format);
-      } else if (arrLength === 1) {
-        console.error('Warning: ' + format, arr[0]);
-      } else if (arrLength === 2) {
-        console.error('Warning: ' + format, arr[0], arr[1]);
-      } else if (arrLength === 3) {
-        console.error('Warning: ' + format, arr[0], arr[1], arr[2]);
-      } else if (arrLength === 4) {
-        console.error('Warning: ' + format, arr[0], arr[1], arr[2], arr[3]);
-      } else if (arrLength === 5) {
-        console.error(
-          'Warning: ' + format,
-          arr[0],
-          arr[1],
-          arr[2],
-          arr[3],
-          arr[4],
-        );
-      } else if (arrLength === 6) {
-        console.error(
-          'Warning: ' + format,
-          arr[0],
-          arr[1],
-          arr[2],
-          arr[3],
-          arr[4],
-          arr[5],
-        );
-      } else if (arrLength === 7) {
-        console.error(
-          'Warning: ' + format,
-          arr[0],
-          arr[1],
-          arr[2],
-          arr[3],
-          arr[4],
-          arr[5],
-          arr[6],
-        );
-      } else if (arrLength === 8) {
-        console.error(
-          'Warning: ' + format,
-          arr[0],
-          arr[1],
-          arr[2],
-          arr[3],
-          arr[4],
-          arr[5],
-          arr[6],
-          arr[7],
-        );
+      // We intentionally don't use spread (or .apply) because it breaks IE11:
+      // https://github.com/facebook/react/issues/13610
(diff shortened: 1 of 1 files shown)


react-linter:
<details>

  <p>Comparing: f6fb03edffcc5918ad2f37ed45bdc7c0c3b3d199...f785a3dbcbf0ceb8112eda9532017b385638784e</p>


  
## schedule

  </details>
  
<p align="right">
</p>

packages/shared/warningWithoutStack.js
```diff
@@ -27,7 +29,14 @@ if (__DEV__) {
     }
     if (typeof console !== 'undefined') {
       const stringArgs = args.map(item => '' + item);
+      const originalConsoleError = console.error;
+
+      if (typeof console.error.apply === 'undefined') {
+        console.error = consoleError;
```
gaearon: There are more moving parts here than I'd like.

I'd like to use the same solution for all cases. Can we do that somehow?

The easiest one might just be to switch on `args` length and explicitly pass them all down.

link-alex: That was my first intention :)
But then I thought that maybe it’s not desired to switch to fixed amount of args.
If it’s ok for you - I can update.

gaearon: Fixed amount of args is okay. We have that in `invariant`.

link-alex:
All the parameters are passed explicitly now, but result looks a bit frustrating for me.. So please have a look, thanks!

packages/shared/warningWithoutStack.js
```diff
@@ -15,8 +15,6 @@
 let warningWithoutStack = () => {};
 
 if (__DEV__) {
-  const consoleError = Function.prototype.bind.call(console.error, console);
-
   warningWithoutStack = function(condition, format, ...args) {
```
gaearon: Let’s assert the number of arguments isn’t higher than supported here?

link-alex: As far as I can see the maximum of 8 arguments is used at the moment for this warning. Do you want to add extra check? But how many arguments do we need to check? That's exactly the reason I didn't use this approach initially..

btw, the tests marked as failed, but as I can see on circleci - tests itself passed, but later this happened. Is it related to the change? Or verification job could be just restarted?

Request failed [401]: https://api.github.com/repos/facebook/react/pulls/13620/requested_reviewers
Response: {
  "message": "Bad credentials",
  "documentation_url": "https://developer.github.com/v3"
}
Error:  TypeError: Cannot read property 'repo' of undefined
    at GitHub.APIMetadataForPR (/home/circleci/project/node_modules/danger/distribution/platforms/GitHub.js:274:27)
    at GitHub.<anonymous> (/home/circleci/project/node_modules/danger/distribution/platforms/GitHub.js:164:39)
    at step (/home/circleci/project/node_modules/danger/distribution/platforms/GitHub.js:40:23)
    at Object.next (/home/circleci/project/node_modules/danger/distribution/platforms/GitHub.js:21:53)
    at fulfilled (/home/circleci/project/node_modules/danger/distribution/platforms/GitHub.js:12:58)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
Danger failed

Kovensky:
This sounds a bit crazy, but I wonder if `Function.prototype.apply.apply(console.error, console, args)` would work 🤔

link-alex:
@Kovensky not really as far as I can see.. but interesting idea.

packages/shared/warningWithoutStack.js
gaearon: If 8 is current maximum the let’s check we have at most 8.

gaearon: Try rebasing to solve the CI issue

gaearon:
Thanks!

packages/shared/warningWithoutStack.js
```diff
@@ -22,12 +22,54 @@ if (__DEV__) {
...
+          break;
+        case 7:
+          console.error(message, a, b, c, d, e, f, g);
+          break;
+        case 8:
+          console.error(message, a, b, c, d, e, f, g, h);
+          break;
+        default:
+          throw new Error(
+            'warningWithoutStack() currently supports at most 8 arguments.',
```
NE-SmallTown: Why we need to do this again? I mean, we have done the `args.length > 8` check, we won't reach this branch

gaearon: Not really necessary. Just left here for exhaustiveness of switch. It should never happen.

lustoykov:
@Kovensky, @link-alex 

> This sounds a bit crazy, but I wonder if `Function.prototype.apply.apply(console.error, console, args)` would work 🤔

I think the correct syntax would be 

```js
const arrayArgs = ['message', 'a', 'b', 'c'];
const thisArg = null;

Function.prototype.apply.apply(console.error, [thisArg, arrayArgs]);
```

which would call `console.error` with scope `thisArg` and arguments `arrayArgs`. Unfortunately too unreadable.

PR #13621, opened 2018-09-11
Fix deadline judgement bug

NE-SmallTown:
When `deadline === null`, the expression `deadline.timeRemaining() > timeHeuristicForUnitOfWork` will still will be executed, it will throw error.

The reason it doesn't throw currently is `deadline` always won't be `null` in the `shouldYield` function.

If this needs to add a test, I will try(though I think it doesn't need and I don't know how to test just for this expression for now).

Of course we can just remove the `deadline === null` check, but I think it is useful in the future?

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -2371,7 +2371,7 @@ function shouldYield() {
   }
   if (
     deadline === null ||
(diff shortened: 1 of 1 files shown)


react-linter:
<details>

  <p>Comparing: b87aabdfe1b7461e7331abb3601d9e6bb27544bc...290e6d7e5793a87647d8ccdeef2221091fe87059</p>


  
## schedule

  </details>
  
<p align="right">
</p>

PR #13639, opened 2018-09-13
Remove function hoisting dependency

dutiyesh:
Move `return` statement below all function declaration


gaearon:
Hi, thanks for PR but we generally try to avoid purely stylistic changes.

PR #13640, opened 2018-09-13
Refactoring. Remove unnecessary else with return

janczer:
We can delete last `else` because this no necessary.

For example, if we have code like this:

```javascript
if (typeof child === 'number') {
  return child
} else {
  return this.nodes.indexOf(child)
}
```

And variable `child` has type `number` then the function always return `child` and never goes on the lines after if-statment. So we can delete `else`-branch:

```javascript
if (typeof child === 'number') {
  return child
}
return this.nodes.indexOf(child)
``` 

I think this part of code more readable then first one.


janczer:
I understand. Thanks

PR #13641, opened 2018-09-13
Rename "tracking" API to "tracing"

bvaughn:
* 9721ecb: Rename all files
* 69659e6: Rename all strings

Testing:
* Lint and Flow
* `yarn test` and `yarn test-build`
* Tested `tracing` and `packaging` fixtures against new built bundle
* Tested the "suspense" fixture and DevTools profiler against the built bundle

PR #13642, opened 2018-09-13
Add UMD production+profiling entry points

### LINKED ISSUE #13634: Missing UMD variant of react.profiling.min.js
The 16.5.0 package released to unpkg currently has the production/profiling variant present for cjs builds but there is no equivalent for umd builds. 

i.e. There is the file https://unpkg.com/react@16.5.0/cjs/react.profiling.min.js but there is no https://unpkg.com/react@16.5.0/umd/react.profiling.min.js

It would be great if a future release had the profiling variant packaged as umd as well.

Also thanks for the awesome framework!

bvaughn:
Resolves #13634 by enabling apps that use the UMD build of React (and co) to use the new profiling and tracing APIs in production mode.

Note that this PR builds on top of PR #13641. The only commit this branch contains is 18b4a27.

### Changes
* Removed unnecessary CJS profiling bundle for `react` package (since it's not necessary).
* Added new `UMD_PROFILING` bundle type and plugged into `react` and `react-dom`.
* Added new `umd/schedule.profiling.min.js` shim to `schedule` package. (This is just a copy of `umd/schedule.production.min.js`.)


bvaughn:
Whoops, yeah. Thanks for catching that.

scripts/rollup/wrappers.js
```diff
@@ -49,6 +50,16 @@ ${license}
 ${source}`;
   },
 
+  /***************** UMD_PROFILING *****************/
```
NE-SmallTown: Seems more '*' than others 😄

PR #13647, opened 2018-09-14
Remove duplicate code in EventPluginRegistry

rodrigooler:
I was reading the React core to further study the ecosystem, found some code duplicates, and did a refactor, I would like to mature the code. If PR is useful.

- [x] run yarn;
- [x] remove duplicate codes;
- [x] create testEventPlugin;
- [x] run `yarn test`;
- [x] run `yarn test --watch EventPluginRegistry`;
- [x] run `yarn test-prod`;
- [x] run `yarn prettier`; 
- [x] run `yarn lint`;

## Run tests

```
./node_modules/.bin/cross-env NODE_ENV=development jest --config ./scripts/jest/config.source.js /Users/rodrigooler/projects/react/packages/events/__tests__/EventPluginRegistry-test.internal.js&& yarn prettier && yarn lint
```

captura de tela 2018-09-14 as 04 52 25

### DIFF
--- a/packages/events/__tests__/EventPluginRegistry-test.internal.js
+++ b/packages/events/__tests__/EventPluginRegistry-test.internal.js
@@ -1,7 +1,7 @@
 /**
- * Copyright (c) Facebook, Inc. and its affiliates.
+ * Copyright (c) Facebook, Inc. and xits affiliates.
  *
- * This source code is licensed under the MIT license found in the
+ * This source code is licensed under the Mxit license found in the
  * LICENSE file in the root directory of this source tree.
  *
  * @emails react-core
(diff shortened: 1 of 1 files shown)


philipp-spiess:
You added up adding more lines than you removed. I don't think this is making the tests easier to understand. Now in order to understand one test example, you'll need more mental overhead. The arguments to your helper are also not easy to understand without looking at the implementation.

A better way to refactor this test would be to move away from using internal API. Not sure if this is possible though. I suggest to not invest too much time into event system related components as we might start to rewrite the event system soon.

gaearon:
:+1: Thanks for PR but we also prefer to have less abstraction in tests where possible so that it's easier to tweak an individual test when necessary without having to unwind the abstraction. And to make it clearer what tests are actually testing.

PR #12037, opened 2018-01-18
Minimally support iframes (nested browsing contexts) in selection event handling

acusti:
As discussed in https://github.com/facebook/react/pull/9184#issuecomment-355800102, this PR takes a minimal approach to making React work when rendering selectable elements across nested browsing contexts (iframes or browser windows opened via `window.open`). That means changing `ReactDOMSelection` and `SelectEventPlugin` to use the `window` and `document` objects relative to the DOM node being handled where possible, and updating `ReactInputSelection` to find and restore selection to the single active element on the page, even when that element is inside a nested browsing context. I did a quick test using the code in this PR of rendering a Draft editor into an iframe and it seems sufficient to make it all work without issue.

To try this out in your own project, update your `react-dom` dependency string to `github:brandcast/react-dom-built#e673e48`. Also make sure you are using latest React (16.2.0) to avoid a version mismatch.

I made additional fixes to ReactInputSelection’s `hasSelectionCapabilities` function in https://github.com/facebook/react/pull/9184/commits/82443f7738205ec6ee29bc1f5701089a9c4fadf2 that I have not replicated here to avoid bloating the PR. If I should add that change to this PR, let me know and I’d be happy to do so.

@gaearon Here is the full bundle size reporting table (note that the handling of the markdown table headers seems somewhat borked in `master` so the table is a bit unwieldy):

```
┌─────────────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────┐
[...]

PR #13650, opened 2018-09-14
Fix some iframe edge cases

### LINKED ISSUE #13648: 16.5 with better support of iframe has some side effects

bug

When transitioning from a state having an iframe inside a component to a state without that said iframe, we have an error at `setOffsets(node, offsets)` because `node.ownerDocument.defaultView` being `null`

Working on it

Should not break

Affected since this 16.5.0, coming from this PR exactly https://github.com/facebook/react/pull/12037/files#diff-26d90cba6ee597ef475fa80dcf76ae1d

`getOffsets` as a check on `ownerDocument.default` fallbacks to `window`, but `setOffsets` does not.

Is there a reason why?

JSteunou:
Should fix #13648 by being as protective in `setOffsets` than in `getOffsets`

### DIFF
--- a/packages/react-dom/src/client/ReactDOMSelection.js
+++ b/packages/react-dom/src/client/ReactDOMSelection.js
@@ -152,7 +152,7 @@ export function getModernOffsetsFromPoints(
  */
 export function setOffsets(node, offsets) {
   const doc = node.ownerDocument || document;
-  const win = doc ? doc.defaultView : window;
+  const win = (doc && doc.defaultView) || window;
   const selection = win.getSelection();
   const length = node.textContent.length;
   let start = Math.min(offsets.start, length);


gaearon:
Can you verify this fixes it? Also it would really be helpful to add a case to this fixture:

https://github.com/facebook/react/tree/master/fixtures/dom/src/components/fixtures/selection-events

packages/react-dom/src/client/ReactDOMSelection.js
```diff
@@ -151,9 +151,12 @@ export function getModernOffsetsFromPoints(
  * @param {object} offsets
  */
 export function setOffsets(node, offsets) {
-  const doc = node.ownerDocument || document;
-  const win = doc ? doc.defaultView : window;
-  const selection = win.getSelection();
+  const {ownerDocument} = node;
```
acusti: Thanks for tracking this down! This is definitely an oversight of #12037 that we didn’t address.

For context, the reason that `getOffsets` uses a different method to get a relative `window` and `document` object than `setOffsets` is that `getOffsets` only needs to use the relative `window` (which has `getSelections`), whereas `setOffsets` needs to use both the `window` object (for `getSelection`, again), and the `document` object (for `createRange`). So this should really be a hybrid of the two approaches, i.e.:
```js
  const doc = node.ownerDocument || document;
  const win = (doc && doc.defaultView) || window;
  ...
```
Where it still guards against a missing `ownerDocument` by falling back to the global and also guards against the case  of having an `ownerDocument` that’s  missing a `defaultView`.

JSteunou: I understand, I changed the logic a bit by over-protecting this path

Update incoming

JSteunou:
and yes the actual PR fix the issue we have

gaearon:
@acusti Can you sign off on the final version please? If it looks good to you I'll get it in.

gaearon:
Thanks.

JSteunou:
Thank you both!

einarq:
Was this tested in Edge? I saw a bug related to pasting stuff in a TinyMCE editor when using 16.5.1, upgrading to 16.5.2 solved it. But it is still failing in Edge. The issue might be related to something entirely different of course, but it goes away if I downgrade to 16.4.2.
The error coming from TinyMCE is "Unable to get property 'moveToBookmark' of undefined or null reference", not sure if that helps at all.

acusti:
@einarq Can you post a link to a repo or even just a code sample that makes it possible to reproduce the error on Edge? I’m most interested in how you are including and rendering the TinyMCE editor, but anything that will help me to reproduce it. If the error is coming from TinyMCE itself, I’m not sure how it’s related to the changes between 16.4.2 and 16.5.x, but I would be happy to take a look to try to figure it out.

einarq:
I’ll see if I can create a repro and provide some more information next week, but I was just mainly wondering if the fix for the issue was tested in Edge (assuming there was a repro for the bug that was fixed).

einarq:
I have traced the problem to the change in setOffsets between 16.4.2 and
16.5.0.

As previously mentioned we have a TinyMCE editor in every row of a list,
which supports paste to add more rows. When pasting just 5 rows, setOffsets
does not get called at all, but when pasting for instance 100 rows,
setOffsets gets called and fails with "Object expected". Perhaps some type
of timing issue?

A can try to create an example but I suspect it would be difficult due to
the complexity of the scenario with TinyMCE being used etc.

Could we first check if this code even works in Edge? Do you have a repro
case for this bug that was fixed, and if so perhaps that could be tested
with Edge?

Thanks,

Einar

einarq:
It fails due to win.getSelection being undefined.
I see that in 16.4.2 this function had an early return statement for when
getSelection is undefined, could we add that back in?

Changing setOffsets to this makes it work again for me:

  const doc = node.ownerDocument || document;
  const win = doc ? doc.defaultView : window;

  if (!win.getSelection) {
    return;
  }

acusti:
@einarq Thanks for tracking that down! If you want to make a PR that adds an early return `if (!win.getSelection)` to the top of `setOffsets` with the explanation of what it addresses, I imagine that would be easy to get merged. I’m puzzled as to why it’s still necessary, considering that support for `window.getSelection` was added in IE9, and:
```js
const win = (doc && doc.defaultView) || window;
```
will ensure that even if `doc.defaultView` is not a proper `Window` object, we’ll still get the `window` global, which must have the `getSelection` method on it. Only thing I could imagine is that `doc.defaultView` is present but is some kind of limited version of a `Window` object that doesn’t have the full APIs, maybe for cross-domain security reasons?

But again, having an early return there is harmless, so I’m fine with it, even though I don’t understand why it’s needed.

einarq:
Done. Uncharted territory for me (until now), so hope I did it correctly.
https://github.com/facebook/react/pull/14095

Thanks in advance for any help in getting it merged :)

Einar

PR #13662, opened 2018-09-15
Remove Flow typings from Schedule.js

alexeyraspopov:
As described in https://github.com/facebook/react/pull/13582, this file needs to be as close to ES5 as possible. Flow pragma (`@flow`) was deleted but there was still a single function signature. Removing it in this PR.

Not sure about what should be done to the `export` statement in the end though.

### DIFF
--- a/packages/schedule/src/Schedule.js
+++ b/packages/schedule/src/Schedule.js
@@ -186,7 +186,7 @@ function unstable_scheduleWork(callback, options) {
   return newNode;
 }
 
-function unstable_cancelScheduledWork(callbackNode: CallbackNode): void {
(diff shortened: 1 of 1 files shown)


gaearon:
We want to keep `export` / `import` since that part would be necessary for ESM. We'll just compile it alone for UMD later.

gaearon:
We should probably tweak this to do the checking for us.
https://github.com/facebook/react/blob/master/scripts/shared/pathsByLanguageVersion.js

NE-SmallTown:
> We'll just compile it alone for UMD later.

But the `main` field of the package.json of schedule refer to './src/index' rather than './npm/index', I'm a little confused here

PR #13665, opened 2018-09-16
Warn when `static propTypes/static defaultProps` is a function fix #9620

### LINKED ISSUE #9620: Warn when `static propTypes/static defaultProps` in ES6 class is a function
Bug

```
class TestWrongPropTypes extends Component {
    static propTypes() {
        return {
            children: PropTypes.string,
            missing: PropTypes.string.isRequired
        };
    }

    static defaultProps() {
        return { children: 'Default props via static function' };
    }

    render() {
        return <p>{this.props.children}</p>;
    }
}
```

In this example React will silently skip `propTypes` checking and default props setting for `TestWrongPropTypes` component.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/84v837e9/).**
https://jsbin.com/jidupehebu/edit?js,console
https://jsfiddle.net/84v837e9/30/

I know that in order to work, propTypes definition should be `static get propTypes = {...}` or `TestWrongPropTypes.propTypes = {...}`. But I'd like to have warning like "propTypes/defaultProps is function but should be either property or getter" to prevent such errors.

Discovered in React 15.X, but probably the same behavior will be also in earlier versions.

Zzzen:
This check might be helpful for newcomers since default props can only be returned from factory functions in vue.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberClassComponent.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.js
@@ -312,6 +312,20 @@ function checkClassInstance(workInProgress: Fiber, ctor: any, newProps: any) {
         'Use a static property to define defaultProps instead.',
       name,
     );
+    const noDefaultPropsFunction = typeof ctor.defaultProps !== 'function';
+    warningWithoutStack(
+      noDefaultPropsFunction,
+      'defaultProps was defined as a static function on %s. ' +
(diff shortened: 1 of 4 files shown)


Zzzen:
ping @gaearon

PR #13669, opened 2018-09-16
Add flow types in ReactControlledComponent

armujahid:
1. `setRestoreImplementation` is only used in `ReactDOM.js` and takes `restoreControlledState` as input which is of type `(Element, string, Object) => void` (see `ReactDOMComponent.js`)

2. `enqueueStateRestore` takes `DOMEventTarget` as input that's why type `EventTarget` is used

### DIFF
--- a/packages/events/ReactControlledComponent.js
+++ b/packages/events/ReactControlledComponent.js
@@ -3,6 +3,8 @@
  *
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
  */
 
 import invariant from 'shared/invariant';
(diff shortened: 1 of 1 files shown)


packages/events/ReactControlledComponent.js
```diff
@@ -35,11 +37,13 @@ function restoreStateOfTarget(target) {
   restoreImpl(internalInstance.stateNode, internalInstance.type, props);
 }
 
-export function setRestoreImplementation(impl) {
+export function setRestoreImplementation(
+  impl: (Element, string, Object) => void,
```
gaearon: maybe let's name the arguments?

armujahid: Do tell me if anything else is required.

armujahid:
@gaearon Let me know if anything else is required.

armujahid:
@nhunzaker Can you also review and merge this? I have done the requested change by @gaearon

nhunzaker:
Cool. Looks good to me. Thanks for matching up the names to other usage too:

https://github.com/facebook/react/blob/8a8d973d3cc5623676a84f87af66ef9259c3937c/packages/react-dom/src/client/ReactDOMComponent.js#L1225-L1229

💯

PR #13674, opened 2018-09-17
Revert "Do not bind topLevelType to dispatch"

gaearon:
Reverts facebook/react#13618

This breaks some tests using SimulateNative :-(.
Don't know what's the cause but if it breaks us, it can break other people too.

We could investigate this more but I think it's not high priority for now so I'll just get this in to unblock syncing (and 16.5.2)


gaearon:
GitHub is so weird. I don't know why it shows those separate "revert" commits even though I revert just one. It should be fine after squash though.

nhunzaker:
Bummer. Sorry to just catch wind of this :(.

PR #12064, opened 2018-01-21
Allow getDefaultProps in dev and production.

sportnak:
This should resolve [#9999](https://github.com/facebook/react/issues/9999)

Issue #9999, opened 2017-06-19
Question on react's warning getDefaultProps


mondwan:
As migrating from 15.4 to 15.5, I am rewrite original react components with `create-react-class`. 

However, there is a warning.

```
react-15.5.4.js?bust=1497843639843:3287 Warning: getDefaultProps is only used on classic React.createClass definitions. Use a static property named `defaultProps` instead.
```

 As I am really transferring from a **classic** `React.createClass` definition, I would like to ask is this warning appropriate? Or, this is a depreciation sames as the level of migrating to `create-react-class`?

Ooops, I get ticket #9999  :D

aweary:
@mondwan can you share an example reproducing the issue? I'm not seeing the warning when using the latest `create-react-class` version, or the last 15.5 release. [Example](https://jsfiddle.net/sch60ss0/). You can use [this JSFiddle](https://jsfiddle.net/84v837e9/) as a starting point for reproducing.

mondwan:
https://jsfiddle.net/84v837e9/105/

For [non minified version][1], it works as expected. However, it does not if using [minified version][2]

[1]: https://unpkg.com/create-react-class@15.6.0/create-react-class
[2]: https://unpkg.com/create-react-class@15.6.0/create-react-class.min

aweary:
@mondwan the issue is that you're using the minified `create-react-class` build, but the unminified `react` build. This means that the internal flag that `create-react-class` uses to suppress this warning is not applied, so React warns.

Make sure you use the correct build for your environment (minified for production, unminified for development).

mondwan:
Umm, but why this is a warning at the first place?

2017年6月20日 18:03，"Brandon Dail" <notifications@github.com>寫道：

aweary:
Because `getDefaultProps` is only used for components created with `createReactClass`. For example, the warning will catch cases where someone is moving from `createClass` to ES6 classes and they forget to remove `getDefaultProps` in favor of `defaultProps`.

You'll never get the warning when using the correct build of `create-react-class`. It's also important to use minified bundles for `react` and `react-dom` in production, as they're much more performant. On the same token, it's useful to use the unminified build in development as it contains a sizeable set of useful checks and warnings.

mondwan:
That's the point. I am still using react.createClass instead of es6 class
even though this has been replaced by create-react-class. It should not get
a warning at the first place.

Or, create-react-class is made of es6 class. So, rules like that are
effective?

[...]

PR #13675, opened 2018-09-18
Ensure there are no warnings for getDefaultProps when using createReactClass

CharlieTruong:
This aims to address the issue where there is a false positive warning for getDefaultProps when using the prod version of `createReactClass` with the dev version of React.
Issue: https://github.com/facebook/react/issues/9999

There was a previous PR that attempted to address this by setting flags even in prod: https://github.com/facebook/react/pull/12064 .  However, @gaearon raised a concern about setting those flags unnecessarily in prod.  

This PR resolves that by checking if `getDefaultProps` is on the prototype created by `createReactClass` instead of any of the dev flags.  This seems to work fine to distinguish components from `createReactClass` vs extending the class `Component`.
https://github.com/facebook/react/blob/15-stable/addons/create-react-class/factory.js#L857

### DIFF
--- a/packages/react/src/__tests__/ReactJSXElementValidator-test.js
+++ b/packages/react/src/__tests__/ReactJSXElementValidator-test.js
@@ -341,32 +341,6 @@ describe('ReactJSXElementValidator', () => {
     );
   });
 
-  it('does not warn when using getDefaultProps with createReactClass', () => {
-    // Set NODE_ENV to 'production' to verify that dev specific flags
-    // are not necessary
-    const ORIGINAL_ENV = process.env;
-    process.env = {
-      ...ORIGINAL_ENV,
-      NODE_ENV: 'production',
-    };
-    const createReactClass = require('create-react-class');
-    const ValidGetDefaultPropsComponent = createReactClass({
-      getDefaultProps: function() {
-        return {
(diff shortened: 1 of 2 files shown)


packages/react/src/__tests__/ReactJSXElementValidator-test.js
```diff
@@ -341,6 +341,32 @@ describe('ReactJSXElementValidator', () => {
...
+      },
+      render: function() {
+        return <span>{this.props.prop}</span>;
+      },
+    });
+
+    spyOnDevAndProd(console, 'error');
+    ReactTestUtils.renderIntoDocument(<ValidGetDefaultPropsComponent />);
+    expect(console.error).not.toHaveBeenCalled();
+    process.env = ORIGINAL_ENV;
```
mycatnamedweb: Isn't it a problem that this will not be reverted to its original value if the test fails?

CharlieTruong: Good call.  I pushed a commit to move this test to be under a `describe` block and ensured the `process.env` is reset in a `afterAll` block.  Thanks for the feedback!

temsa:
@gaearon Any reason why it has not been merged ?

Issue #13677, opened 2018-09-18
Children.map & Co do not traverse keyed Fragments


arty-name:
**Do you want to request a *feature* or report a *bug*?**

Report a bug

**What is the current behavior?**

<a href="https://reactjs.org/docs/react-api.html#reactchildren">The documentation on React.Children.map</a> tells us that "If children is a keyed fragment or array it will be traversed: the function will never be passed the container objects." [This demo](https://jsfiddle.net/cxq9dzet/) shows that arrays are indeed traversed, but the <a href="https://reactjs.org/docs/fragments.html#keyed-fragments">keyed fragments</a> are not traversed. Open the developer console and click "Run" to see the output.

**What is the expected behavior?**

Children.map, forEach, toArray, count should traverse keyed Fragments as per documentation.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

The example runs on v16.5.1, but I have also seen that fail locally on v16.4.2. I never tried this before so I don’t know if this ever worked. I see it fail in latest Firefox and Chrome on Ubuntu Linux.

arty-name:
I think that the test I have added in  #13679 also demonstrates the bug

jquense:
I don't think this is a bug actually, it's been discussed a few times in other issues and I believe the consensus was that it's intended that fragments be distinct elements vs an array of elements

arty-name:
@jquense so is this a bug in the documentation?

jquense:
Sounds like it but let's wait for a core team member to confirm

arty-name:
Some way to automatically “flatten” the `Fragments` among the children will be a much desired feature anyway. I would be happy to see the code to work as currently specified in the documentation.

bvaughn:
> I don't think this is a bug actually, it's been discussed a few times in other issues and I believe the consensus was that it's intended that fragments be distinct elements vs an array of elements

Yes, it is intentional the `.map` does not traverse into fragments. I'll update the docs. Apologies for the confusion!

arty-name:
After more searching I have found [this comment](https://github.com/facebook/react/issues/11859#issuecomment-355656427) by Dan Abramov: 

> I think changing behavior for both might be reasonable.

Which I interpret as this implementation not being particularly intentional. He then proposes to submit an RFC for these changes, but I couldn’t find any at the moment. I will try to create one now.

arty-name:
[...]

PR #13679, opened 2018-09-18
ReactChildren test 'should traverse children of a keyed fragment'

arty-name:
This is a unit test to reproduce #13677


bvaughn:
Thanks for the repro case! The failing behavior it captures is intentional though 😄 Sorry for the confusion!

PR #13683, opened 2018-09-19
Rename schedule to scheduler

gaearon:
Ran the fixtures and I think everything works.


packages/react/src/ReactSharedInternals.js
```diff
@@ -38,12 +38,12 @@ if (__UMD__) {
   // This re-export is only required for UMD bundles;
   // CJS bundles use the shared NPM package.
   Object.assign(ReactSharedInternals, {
-    Schedule: {
+    Scheduler: {
```
gaearon: I think this means a newer React wouldn't work with an older ReactDOM, or vice versa. Since it's scoped to UMDs only I think that's fine and I wouldn't worry about going to extra lengths to make this work for mismatching versions.

NE-SmallTown:
Can I ask why we use 'schedule' rather than 'scheduler'? IIRC, we use 'react-scheduler' before but rename it to 'schedule', sorry about don‘t come up with this confusion in that PR

bvaughn:
We didn't own the `scheduler` NPM package so we couldn't use it. It was the preferred name originally but we settled for `schedule` because that was all we had at the time.

bvaughn:
We have talked a little about potentially using `@react/<package>` for some future packages– but schedule/scheduler isn't inherently specific to React.

monkindey:
I think it will break the webpack config when we use the `react-dom/profiling`.

schedule ==> scheduler

### Before

```js
module.exports = {
  //...
  resolve: {
    alias: {
      'react-dom': 'react-dom/profiling',
      'schedule/tracing': 'schedule/tracing-profiling',
    }
  }
};
```

### After

```js
module.exports = {
  //...
  resolve: {
    alias: {
      'react-dom': 'react-dom/profiling',
      'scheduler/tracing': 'scheduler/tracing-profiling',
    }
  }
};
```

gaearon:
Yeah, we need to change these instructions.

monkindey:
[Here](https://gist.github.com/bvaughn/25e6233aeb1b4f0cdb8d8366e54a3977) ?

bvaughn:
Both [fb.me/react-profiling](https://fb.me/react-profiling) and [fb.me/react-interaction-tracing](https://fb.me/react-interaction-tracing) have been udpated!

PR #13687, opened 2018-09-19
[scheduler] add a test documenting current behavior

plievone:
Please feel free to use/adapt this test that I made to understand the current scheduler better. For me this behavior is reasonable, however I don't necessarily understand the reason for reversed ordering after timeout.


gaearon:
Is it the reverse ordering specifically that's confusing? Could it be because they're getting exactly the same clock time, so they time out "simultaneously", their order depending on whether there's a `<` or `<=` check somewhere? I wonder if it would be different if you let a tick pass between each schedule call (which is how it would likely work outside of tests).

Or maybe I misunderstand.

acdlite:
@plievone You're correct, the flushing the callbacks in reverse order was a bug. I've updated your test and confirmed it works on master. Thank you!

packages/scheduler/src/__tests__/Scheduler-test.internal.js
```diff
@@ -153,6 +156,55 @@ describe('Scheduler', () => {
...
+    // No time left after A1_100 and A2_200 are run
+    expect(flushWork(300)).toEqual(['A1_100', 'A2_200']);
+    expect(Date.now()).toEqual(300);
+    // B2_200 is started as there is still time left after B1_100
+    expect(flushWork(101)).toEqual(['B1_100', 'B2_200']);
+    expect(Date.now()).toEqual(600);
+    // C1_300 is started as there is even a little frame time
+    expect(flushWork(1)).toEqual(['C1_300']);
+    expect(Date.now()).toEqual(900);
+    // C2_300 is started even though there is no frame time
```
NE-SmallTown: Could we give more explanation about why "even though there is no frame time"

PR #13689, opened 2018-09-19
Fix circular dependency in TracingSubscriptions

mmarkelov:
Fixed Circular dependency warning in new version of Rollup.
It will cause build failure


gaearon:
Seems legit. Thanks.

Issue #13645, opened 2018-09-14
Element attributes be removed silently in frameset tag


LeonDWong:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
In some historical reason, i am still using obsolete tag `frameset` in my project. And when i render my component which includes `frameset` tag in react, some attributes are ignored, like `rows`, `cols`.
Could anyone tell me why and how to solve it by no hacking way? Thx.

**input**
```javascript
export default class obsoleteElment extends React.Component {

  render() {
    return (
      <html>
      <head>
        <meta httpEquiv="content-type" content="text/html; charset=UTF-8" />
          <title>old page</title>
      </head>
      <frameset id="frame-container" rows='60,*' frameBorder="0" border="0"  cols="180,11,*">
      </frameset>
      </html>
    )
  }
}
```

**output**
```javascript
      <html>
      <head>
        <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
          <title>old page</title>
      </head>
      <frameset id="frame-container" frameBorder="0" border="0">
      </frameset>
      </html>
```

**expect**
```javascript
      <html>
      <head>
        <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
          <title>old page</title>
      </head>
      <frameset id="frame-container" rows='60,*' frameBorder="0" border="0"  cols="180,11,*">
      </frameset>
      </html>
```

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

**react version**
![image](https://user-images.githubusercontent.com/17465046/45526873-9b08db00-b80b-11e8-980b-86407da1934c.png)

**node environment**
![image](https://user-images.githubusercontent.com/17465046/45526918-ca1f4c80-b80b-11e8-911d-cf0bb3ab9b32.png)
![image](https://user-images.githubusercontent.com/17465046/45526934-d905ff00-b80b-11e8-87d4-306ca3a998c3.png)

**OS**
![image](https://user-images.githubusercontent.com/17465046/45526979-18cce680-b80c-11e8-8928-ee784a5f0a7d.png)

Simek:
According to HTML specification `frameset` tag is obsolete in HTML5, but `cols` and `rows` are valid optional attributes anyway: https://html.spec.whatwg.org/multipage/obsolete.html#frameset

[...]

PR #13692, opened 2018-09-19
allow non-numeric values of 'cols' and 'rows' attributes (#13645)

Simek:
According to HTML spec not only positive numbers are valid values of  'cols' and 'rows' attributes. More information could be found in corresponding issue #13645.

I'm not sure if `DOMPropertyOperations` is the best fitting file to put that test in. Maybe it should be `ReactDOMIframe` since `frameset` tag is fully related to `iframe`?

### DIFF
--- a/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
+++ b/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
@@ -199,5 +199,13 @@ describe('DOMPropertyOperations', () => {
       ReactDOM.render(<my-icon size="5px" />, container);
       expect(container.firstChild.getAttribute('size')).toBe('5px');
     });
+
+    it('should not remove non-numeric cols and rows attributes from frameset', () => {
+      const container = document.createElement('div');
+      const frameset = <frameset rows="25%,*,25%" cols="180,11,*" />;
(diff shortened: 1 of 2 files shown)


gaearon:
Is this a breaking change for any meaningful scenario?

Simek:
I'm not sure but as far as I know, no.

Simek:
Maybe additional warning in DEV mode should pop up for non numeric values passed to `textarea` and any other DOM component that uses only numeric values?

Simek:
@gaearon Something more I have to do here?

gaearon:
I think this is would be a change in behavior for some cases in which we didn't warn. So I'm not sure it's something we can sneak into a patch or a minor. More investigation as to in which cases the behavior changes would be helpful.

Simek:
@gaearon Is it planned to be merged in 16.7 or in the next major?

Simek:
@gaearon Is it still valid?

gaearon:
Here's the original PR that tagged it as a number: https://github.com/facebook/react/pull/500. You can see it's not super trivial and there are specific cases we care about handling (https://gist.github.com/zpao/227a395cb5821a0fc91b). If you want to get this in, I think you'll need to verify none of the existing cases are broken by this.

PR #11696, opened 2017-11-28
Fix window.event in development (#11687)

ConradIrwin:
Fixes #11687

The code change is relatively simple, but I had to add a fixture to test
this as window.event is not supported by JSDom.

I included a commented-out test copying the pattern of another test in
that file, but I'm not sure how much value that adds so would happily
remove it.

PR #13697, opened 2018-09-19
Restore global window.event after event dispatching (#13688)

### LINKED ISSUE #13688: Global `window.event` is overwritten in React 16.5+ in development mode.

Report a bug.

Global `window.event` is overwritten in React 16.5+ in development mode. Here're minimal repro steps:

- React **16.5.2 in dev mode**: https://jsfiddle.net/sergei_startsev/ecz103vL/2/

If you click the button, you see `DOMContentLoaded` event type.

The current behavior contradicts with specified behavior [window property event returns the Event which is **currently being handled** by the site's code. Outside the context of an event handler, the value is **always** `undefined`.](https://developer.mozilla.org/en-US/docs/Web/API/Window/event). Moreover it works properly in production mode:

- React **16.5.2 in prod mode**: https://jsfiddle.net/sergei_startsev/st5L8e02/2/

It returns expected `click` event type.

It works properly in React 16.4.2 and prod mode:

- React **16.4.2** in dev mode: https://jsfiddle.net/sergei_startsev/o26L08qw/

The issue is reproduced in Chrome 69. It works properly in FF 62 (`window.event` isn't support in 62, however it should be reproduced in FF 63 - it was recently added, see [details](https://developer.mozilla.org/en-US/docs/Web/API/Window/event#Browser_Compatibility)).

It seems that the issue was introduced by @ConradIrwin in #11696.

sergei-startsev:
Used @ConradIrwin [suggestion](https://github.com/facebook/react/issues/13688#issuecomment-422939757) to fix #13688.

### DIFF
--- a/packages/shared/invokeGuardedCallbackImpl.js
+++ b/packages/shared/invokeGuardedCallbackImpl.js
@@ -98,6 +98,13 @@ if (__DEV__) {
       // browsers that support it.
       let windowEvent = window.event;
 
+      // Keeps track of the descriptor of window.event to restore it after event
+      // dispatching: https://github.com/facebook/react/issues/13688
+      const windowEventDescriptor = Object.getOwnPropertyDescriptor(
+        window,
+        'event',
+      );
+
       // Create an event handler for our fake event. We will synchronously
       // dispatch our fake event using `dispatchEvent`. Inside the handler, we
       // call the user-provided callback.
@@ -171,6 +178,10 @@ if (__DEV__) {
       // errors, it will trigger our global error handler.
       evt.initEvent(evtType, false, false);
       fakeNode.dispatchEvent(evt);
+
+      if (windowEventDescriptor) {
+        Object.defineProperty(window, 'event', windowEventDescriptor);
+      }
 
       if (didError) {
         if (!didSetError) {


viniciusll:
replace the code

Em qui, 20 de set de 2018 às 09:30, Dan Abramov <notifications@github.com>
escreveu:

sergei-startsev:
@Dan Could you clarify the idea of removing `window.event` assignment? It was initially added in #11696 to fix #13688, I'm not quite sure we can just remove it...

gaearon:
I mean, if we're restoring the descriptor, wouldn't that be sufficient to cover that use case too? I thought that would capture its current value too.

asapach:
@gaearon, the descriptor doesn't have a value, see https://github.com/facebook/react/issues/13688#issuecomment-422927292. The getter will instead return the current event emitted by `dispatchEvent()`. Which is what #11696 was trying to address.

sergei-startsev:
@gaearon `callCallback` is called when a fake `react-invokeguardedcallback` event is dispatched, `window.event` is overridden by original event in `callCallback`, so you cannot replace it by using the descriptor, it will restore a fake event, that's what you would like to avoid. The following code doesn't work properly:

```diff
- window.event = windowEvent;
+ Object.defineProperty(window, 'event', windowEventDescriptor);
```

After all required manipulations with fake events we have to restore the default behavior for `window.event`.

```js
Object.defineProperty(window, 'event', windowEventDescriptor);
```
Was this helpful?

gaearon:
Can we do

```js
Object.defineProperty(window, 'event', {
  ...windowEventDescriptor,
  get() { return windowEvent }
});
```

?

sergei-startsev:
Will it be actually beneficial? We still have to keep both `windowEvent` and `windowEventDescriptor`...

sergei-startsev:
I'm not quite sure regarding browser consistency, e.g. the issue isn't reproduced in FF 62 (the latest release version) -- `window.event` isn't supported here, but it looks like [they implemented it in FF 63](https://bugzilla.mozilla.org/show_bug.cgi?id=218415). The issue isn't also reproduced in IE 11.

sergei-startsev:
E.g. `Object.getOwnPropertyDescriptor(window, 'event')` returns `undefined` in IE 11. It also returns `false` if you try to get own property `window.hasOwnProperty('event')`...

gaearon:
What a mess.

gaearon:
OK. I can merge this if you can help test in all environments: IE9+ (would require some polyfills), including IE11, then different versions of FF and Chrome. And some mobile (e.g. iOS and Chrome on Android).

sergei-startsev:
OK, I'll try to cover all of them except iOS/Safari -- it can be a bit problematic.

sergei-startsev:
@gaearon I've used the following code to check that the fix works properly:

```js
console.log('Global `window.event` descriptor before/after dispatching a fake event:');

console.log(Object.getOwnPropertyDescriptor(window, 'event'));
  ReactDOM.render(
    <h1>React Component</h1>,
  );
console.log(Object.getOwnPropertyDescriptor(window, 'event'));
```

There're different results of `window.event` descriptor before and after dispatching `react-invokeguardedcallback` fake event **without** applying the fix, e.g. in Chrome 69:

```
{"enumerable":false,"configurable":true}
{"value":{"isTrusted":true},"writable":true,"enumerable":true,"configurable":true
```
Here's the [proof](https://jsfiddle.net/sergei_startsev/ze7fbmsa/).

**With** the applied fix [descriptors are the same](https://jsfiddle.net/sergei_startsev/u7jrqgaw/):

```
{"enumerable":false,"configurable":true}
{"enumerable":false,"configurable":true}
```

**Note:** Since FF 62 and IE9-11 returns `false` for `window.hasOwnProperty('event')` check, `window.event` will never be overridden:

https://github.com/facebook/react/blob/e1a067dea0ffcacd1f664f30cd14463b00f52fa7/packages/shared/invokeGuardedCallbackImpl.js#L118

And the issue isn't reproduced here.

---

I've checked the fix for the following environments:

- Windows 10
    - Chrome 69 +
    - Chrome 71 canary +
    - FF 62 (window.event isn't supported) +
    - FF 63 Beta (window.event is supported) +
    - Edge 42 +
    - IE 11 +
    - IE 10 +
    - IE 9 +

- Android 8
    - Chrome 69 +

I would appreciate if somebody could [check it](https://jsfiddle.net/sergei_startsev/u7jrqgaw/) in iOS/Safari. Anyway I don't see why it shouldn't work here.

philipp-spiess:
@sergei-startsev Looking good on iOS 12, 11, and 10 here. Unfortunately JSFiddle doesn't work on older iOS versions - but running the following script in the console works on iOS 8 and 9 as well:

```js
var descriptor = Object.getOwnPropertyDescriptor(window, 'event');
Object.defineProperty(window, 'event', descriptor)
```

sergei-startsev:
@philipp-spiess thanks for checking!

sergei-startsev:
@gaearon Is there anything else we would like to check here?

negatron99:
Hi, I am getting IE11 hanging and the repeated error "Assignment to read-only properties is not allowed in strict mode", on the line 

`window.event = windowEvent;`

in the cjs\react-dom.development.js

This is in version 16.7.0 

Commenting out the line allows the site to continue working, or commenting out the `'use strict'` lines in the js  file also allows it to work.

Has the build process added the 'use strict' by accident?

mouday:
reactv18.2.0 work in chrome v122, also has this error.

eps1lon:
With Chrome Version 122.0.6261.128 (Official Build) (64-bit)
```js
'use strict';
window.event = 1
```
produces no error. 

@mouday Do you have a minimal reproduction I could look at?

mouday:
i use react with monkey script, https://github.com/lisonge/vite-plugin-monkey
got this error.

mouday:
window.event = windowEvent;

TypeError: Cannot set property event of #<Object> which has only a getter

mouday:
already resolved, thank you.

PR #13705, opened 2018-09-21
Use Math.max instead of conditional operator

rainbowcatcher:
There is a unnecessary conditional operator. I think it is better to use `Math.max`.


gaearon:
Thanks, but we generally try to keep hot code as simple as possible, and that includes avoiding function calls.

PR #13709, opened 2018-09-22
Remove unreachable code in ReactFiberCommitWork

mmarkelov:
`while(true)` will run forever and so the code after it will never be executed.


Wazner:
The `while(true)` won't run forever, there is a break inside the loop.

mmarkelov:
@Wazner, oops my mistake. Thanks!

PR #13718, opened 2018-09-24
Simplify imports in react reconciler

mmarkelov:
Refactoring and implementing one code style for imports in react-reconciler package


packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -9,8 +9,8 @@
 
 import type {ReactElement} from 'shared/ReactElementType';
 import type {ReactPortal} from 'shared/ReactTypes';
-import type {Fiber} from 'react-reconciler/src/ReactFiber';
-import type {ExpirationTime} from 'react-reconciler/src/ReactFiberExpirationTime';
+import type {Fiber} from './ReactFiber';
+import type {ExpirationTime} from './ReactFiberExpirationTime';
```
nhunzaker: I'm curious if this was here for a reason. @gaearon do you know of any reason why this can't be relative?

mmarkelov:
@nhunzaker Merge master here. If there are something wrong with it, let me know

nhunzaker:
Sorry! 

Ran the build, no change in build size. This looks good to go 👍. 

Thanks!

gaearon:
The type imports were intentionally "absolute" to avoid symlink resolution issues on Windows.

This also broke CI after merge. (Maybe because some other change happened on master?)

Let's avoid stylistic changes like this indeed if we're not pursuing some larger goal.

packages/react-reconciler/src/ReactChildFiber.js
gaearon: Yes, it was because Flow has issues with symlinks on Windows.

gaearon: That said we didn't follow it consistently either.

nhunzaker: Okay. Reason enough to revert this? There's still the de-duplication stuff in here. We could revise this PR to remove the relative import changes.

mmarkelov:
@gaeron, I can do pr to get it back. And sorry, but it was absolutely unclear that it will get issue in Windows

gaearon:
Let's keep it this way I guess since so far it doesn't fail. Yeah I know it's not clear. No worries.

PR #13720, opened 2018-09-25
[scheduler] Priority levels, continuations, and wrapped callbacks

acdlite:
All of these features are based on features of React's internal scheduler. The eventual goal is to lift as much as possible out of the React internals into the Scheduler package.

Includes some renaming of existing methods.

- `scheduleWork` is now `scheduleCallback`
- `cancelScheduledWork` is now `cancelCallback`

Priority levels
---------------

Adds the ability to schedule callbacks at different priority levels. The current levels are (final names TBD):

- Immediate priority. Fires at the end of the outermost currently executing (similar to a microtask).
- Interactive priority. Fires within a few hundred milliseconds. This should only be used to provide quick feedback to the user as a result of an interaction.
- Normal priority. This is the default. Fires within several seconds.
- "Maybe" priority. Only fires if there's nothing else to do. Used for prerendering or warming a cache.

The priority is changed using `runWithPriority`:

```js
runWithPriority(InteractivePriority, () => {
  scheduleCallback(callback);
});
```

Continuations
--------------

Adds the ability for a callback to yield without losing its place in the queue, by returning a continuation. The continuation will have the same expiration as the callback that yielded.

Wrapped callbacks
-----------------

Adds the ability to wrap a callback so that, when it is called, it receives the priority of the current execution context.

@n8schloss

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -8,14 +8,34 @@
 
 /* eslint-disable no-var */
 
-// TODO: Currently there's only a single priority level, Deferred. Will add
-// additional priorities.
-var DEFERRED_TIMEOUT = 5000;
+// TODO: Use symbols?
+var ImmediatePriority = 1;
+var InteractivePriority = 2;
+var NormalPriority = 3;
+var WheneverPriority = 4;
+
+// Max 31 bit integer. The max integer size in V8 for 32-bit systems.
+// Math.pow(2, 30) - 1
+// 0b111111111111111111111111111111
+var maxSigned31BitInt = 1073741823;
+
+// Times out immediately
+var IMMEDIATE_PRIORITY_TIMEOUT = -1;
+// Eventually times out
+var INTERACTIVE_PRIORITY_TIMEOUT = 250;
+var NORMAL_PRIORITY_TIMEOUT = 5000;
+// Never times out
+var WHENEVER_PRIORITY_TIMEOUT = maxSigned31BitInt;
 
 // Callbacks are stored as a circular, doubly linked list.
 var firstCallbackNode = null;
 
-var isPerformingWork = false;
+var currentPriorityLevel = NormalPriority;
+var currentEventStartTime = -1;
+var currentExpirationTime = -1;
+
+// This is set when a callback is being executed, to prevent re-entrancy.
+var isExecutingCallback = false;
 
 var isHostCallbackScheduled = false;
 
@@ -25,6 +45,14 @@ var hasNativePerformanceNow =
 var timeRemaining;
 if (hasNativePerformanceNow) {
   timeRemaining = function() {
+    if (
+      firstCallbackNode !== null &&
+      firstCallbackNode.expirationTime < currentExpirationTime
+    ) {
+      // A higher priority callback was scheduled. Yield so we can switch to
+      // working on that.
+      return 0;
+    }
     // We assume that if we have a performance timer that the rAF callback
     // gets a performance timer value. Not sure if this is always true.
     var remaining = getFrameDeadline() - performance.now();
@@ -33,6 +61,12 @@ if (hasNativePerformanceNow) {
 } else {
   timeRemaining = function() {
     // Fallback to Date.now()
+    if (
+      firstCallbackNode !== null &&
+      firstCallbackNode.expirationTime < currentExpirationTime
+    ) {
+      return 0;
+    }
     var remaining = getFrameDeadline() - Date.now();
     return remaining > 0 ? remaining : 0;
   };
@@ -44,22 +78,22 @@ var deadlineObject = {
 };
 
 function ensureHostCallbackIsScheduled() {
-  if (isPerformingWork) {
+  if (isExecutingCallback) {
     // Don't schedule work yet; wait until the next time we yield.
     return;
   }
-  // Schedule the host callback using the earliest timeout in the list.
-  var timesOutAt = firstCallbackNode.timesOutAt;
+  // Schedule the host callback using the earliest expiration in the list.
+  var expirationTime = firstCallbackNode.expirationTime;
   if (!isHostCallbackScheduled) {
     isHostCallbackScheduled = true;
   } else {
     // Cancel the existing host callback.
-    cancelCallback();
+    cancelHostCallback();
   }
-  requestCallback(flushWork, timesOutAt);
+  requestHostCallback(flushWork, expirationTime);
 }
 
-function flushFirstCallback(node) {
+function flushFirstCallback() {
   var flushedNode = firstCallbackNode;
 
   // Remove the node from the list before calling the callback. That way the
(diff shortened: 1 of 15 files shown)


packages/scheduler/src/__tests__/SchedulerUMDBundle-test.internal.js
```diff
@@ -17,7 +17,8 @@ describe('Scheduling UMD bundle', () => {
   });
 
   function filterPrivateKeys(name) {
-    return !name.startsWith('_');
+    // TODO: Figure out how to forward priority levels.
```
acdlite: Probably should inline them? Also should use Symbols, with a fallback to magic numbers.

packages/scheduler/src/Scheduler.js
```diff
@@ -8,14 +8,34 @@
...
+var ImmediatePriority = 1;
+var InteractivePriority = 2;
+var NormalPriority = 3;
+var MaybePriority = 4;
+
+// Max 31 bit integer. The max integer size in V8 for 32-bit systems.
+// Math.pow(2, 30) - 1
+// 0b111111111111111111111111111111
+var maxSigned31BitInt = 1073741823;
+
```
NE-SmallTown: Why not just import `maxSigned31BitInt.js`

acdlite: Because eventually this will live in a separate repo

packages/scheduler/src/Scheduler.js
```diff
@@ -33,6 +61,12 @@ if (hasNativePerformanceNow) {
 } else {
   timeRemaining = function() {
     // Fallback to Date.now()
+    if (
+      firstCallbackNode !== null &&
+      firstCallbackNode.expirationTime < currentExpirationTime
+    ) {
+      return 0;
+    }
```
NE-SmallTown: I'm a little confused here, it seems we can just do
```
const now = hasNativePerformanceNow ? performance.now() : Date.now();
// or for tree-shaking? we can use if (hasNativePerformanceNow) { now = performance.now()} else { now = Date.now() }

timeRemaining = function() { ... } // we just need do this once rather than twice currently
```

packages/scheduler/src/Scheduler.js
```diff
@@ -70,20 +104,101 @@ function flushFirstCallback(node) {
     firstCallbackNode = null;
     next = null;
   } else {
-    var previous = firstCallbackNode.previous;
-    firstCallbackNode = previous.next = next;
-    next.previous = previous;
+    var lastCallbackNode = firstCallbackNode.previous;
+    firstCallbackNode = lastCallbackNode.next = next;
+    next.previous = lastCallbackNode;
```
NE-SmallTown: nice rename ~

packages/scheduler/src/Scheduler.js
```diff
@@ -70,20 +104,101 @@ function flushFirstCallback(node) {
...
+    // Insert the new callback into the list, sorted by its timeout.
+    if (firstCallbackNode === null) {
+      // This is the first callback in the list.
+      firstCallbackNode = continuationNode.next = continuationNode.previous = continuationNode;
+    } else {
+      var nextAfterContinuation = null;
+      var node = firstCallbackNode;
+      do {
+        if (node.expirationTime >= expirationTime) {
+          // This callback is equal or lower priority than the new one.
```
NE-SmallTown: Maybe "This callback priority is equal or lower than the new one" is better just IMO

packages/scheduler/src/Scheduler.js
```diff
@@ -70,20 +104,101 @@ function flushFirstCallback(node) {
...
+  if (typeof continuationCallback === 'function') {
+    var continuationNode: CallbackNode = {
+      callback: continuationCallback,
+      priorityLevel,
+      expirationTime,
+      next: null,
+      previous: null,
+    };
+
+    // Insert the new callback into the list, sorted by its timeout.
```
NE-SmallTown: timeout -> expirationTime?

NE-SmallTown: And seems it doesn't do sort, it just do a find & insert operation

gaearon:
So for now this just adds things? How much do you anticipate being able to remove (to balance out the size increase)?

acdlite:
@gaearon This PR adds roughly 200 lines of code to the Scheduler package. I expect this will be offset when we reimplement React's root scheduling and expiration time system on top of the new Scheduler primitives.

gaearon:
Right. Now that I re-read it, it only increased UMD and 10% is actually pretty little compared to overall size of React UMD.

packages/scheduler/src/Scheduler.js
```diff
@@ -70,20 +104,101 @@ function flushFirstCallback(node) {
...
+      previous.next = nextAfterContinuation.previous = continuationNode;
+      continuationNode.next = nextAfterContinuation;
+      continuationNode.previous = previous;
+    }
+  }
+}
+
+function flushImmediateWork() {
+  if (
+    currentEventStartTime === -1 &&
```
bvaughn: This condition looks subtle and could use a comment.

I _think_ it's meant to guard against reentrancy, e.g. if you called `wrap` and then immediately executed the wrapped function _within_ another scheduled callback– because I _think_ the only time `currentEventStartTime === -1` is essentially when it's called from a `finally` block (else it would have the current time).

packages/scheduler/src/Scheduler.js
```diff
@@ -116,36 +231,93 @@ function flushWork(didTimeout) {
...
+  var previousEventStartTime = currentEventStartTime;
+  currentPriorityLevel = priorityLevel;
+  currentEventStartTime = getCurrentTime();
+
+  try {
+    return eventHandler();
+  } finally {
+    currentPriorityLevel = previousPriorityLevel;
+    currentEventStartTime = previousEventStartTime;
+    flushImmediateWork();
```
bvaughn: Is it intentional that we still flush immediate in the event of an error?

packages/scheduler/src/Scheduler.js
```diff
@@ -116,36 +231,93 @@ function flushWork(didTimeout) {
...
+    var previousEventStartTime = currentEventStartTime;
+    currentPriorityLevel = parentPriorityLevel;
+    currentEventStartTime = getCurrentTime();
+
+    try {
+      return callback.apply(this, arguments);
+    } finally {
+      currentPriorityLevel = previousPriorityLevel;
+      currentEventStartTime = previousEventStartTime;
+      flushImmediateWork();
```
bvaughn: (Same question about flushing after an error)

packages/scheduler/src/Scheduler.js
```diff
@@ -70,20 +104,101 @@ function flushFirstCallback(node) {
...
+
+    // Insert the new callback into the list, sorted by its timeout.
+    if (firstCallbackNode === null) {
+      // This is the first callback in the list.
+      firstCallbackNode = continuationNode.next = continuationNode.previous = continuationNode;
+    } else {
+      var nextAfterContinuation = null;
+      var node = firstCallbackNode;
+      do {
+        if (node.expirationTime >= expirationTime) {
```
bvaughn: Might be worth a comment here that we check ">=" (instead of ">" like in `unstable_scheduleCallback`) intentionally, because we want the continuation to be the _first_ callback with this priority. (It's probably not that subtle but still may be worth mentioning explicitly...)

packages/scheduler/src/__tests__/Scheduler-test.internal.js
```diff
@@ -124,51 +167,311 @@ describe('Scheduler', () => {
...
+          }
+        }
+      };
+      scheduleCallback(work);
+      expect(flushWork()).toEqual([
+        'A',
+        'B',
+        'Schedule high pri',
+        // Even though there's time left in the frame, the low pri callback
+        // should yield to the high pri callback
```
bvaughn: Nice! Glad to see this explicitly tested

packages/scheduler/src/__tests__/Scheduler-test.internal.js
```diff
@@ -124,51 +167,311 @@ describe('Scheduler', () => {
...
     expect(clearYieldedValues()).toEqual([]);
+    advanceTime(1);
+    expect(clearYieldedValues()).toEqual(['Interactive']);
+
+    advanceTime(10000);
+    expect(clearYieldedValues()).toEqual(['Normal']);
+  });
 
-    // Now advance by just a bit more
+  it('wrapped callbacks inherit the current even when nested', () => {
```
bvaughn: nit "wrapped callbacks inherit the current **priority**" ?

packages/scheduler/src/Scheduler.js
```diff
@@ -70,20 +104,101 @@ function flushFirstCallback(node) {
...
+    };
+
+    // Insert the new callback into the list, sorted by its timeout.
+    if (firstCallbackNode === null) {
+      // This is the first callback in the list.
+      firstCallbackNode = continuationNode.next = continuationNode.previous = continuationNode;
+    } else {
+      var nextAfterContinuation = null;
+      var node = firstCallbackNode;
+      do {
```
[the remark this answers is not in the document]
acdlite: Yeah I'll add a comment. It's mostly just a fork of `scheduleWork` but it inserts the continuation before the first callback with equal expiration instead of after the last callback with equal expiration time.

packages/scheduler/src/Scheduler.js
acdlite: Yeah it's like `try/finally`. I'll add a test.

packages/scheduler/src/Scheduler.js
```diff
@@ -8,14 +8,34 @@
...
+// Max 31 bit integer. The max integer size in V8 for 32-bit systems.
+// Math.pow(2, 30) - 1
+// 0b111111111111111111111111111111
+var maxSigned31BitInt = 1073741823;
+
+// Times out immediately
+var IMMEDIATE_PRIORITY_TIMEOUT = -1;
+// Eventually times out
+var INTERACTIVE_PRIORITY_TIMEOUT = 250;
+var DEFAULT_PRIORITY_TIMEOUT = 5000;
```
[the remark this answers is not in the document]
acdlite: D'oh. Good catch. I keep going back and forth on which one I prefer, but usually in conversation I end up saying "normal" so that's what I went with here. These aren't final though.

packages/scheduler/src/__tests__/Scheduler-test.internal.js
```diff
@@ -124,51 +167,311 @@ describe('Scheduler', () => {
...
+    scheduleCallback(() => doWork('B', 100));
+
+    const tasks = [['C1', 100], ['C2', 100], ['C3', 100]];
+    const C = deadline => {
+      while (tasks.length > 0) {
+        doWork(...tasks.shift());
+        if (
+          tasks.length > 0 &&
+          !deadline.didTimeout &&
+          deadline.timeRemaining() <= 0
```
plievone: Drive-by-comment: This seems to be the suggested scheduler usage code, so could you clarify these terms more? 1) Perhaps deadline.didTimeout is rather didExpire (as you have called these elsewhere), but is it a deadline that expires or is there a way to express this without negation. 2) Also what happens if one forgets to check expiration, I would guess the scheduler calls us immediately again so it works, just slower, or is the problem more drastic. 3) Also timeRemaining is about the current frame (or how do you call it) which might confuse the user as it reads now, just after the didTimeout.

acdlite: Thanks for the feedback. Note that the actual names used here are not final yet.

> Perhaps deadline.didTimeout is rather didExpire (as you have called these elsewhere), but is it a deadline that expires or is there a way to express this without negation

The `deadline` naming is inherited from `requestIdleCallback`. You're right that the naming is confusing because it's about the frame deadline, which is different from the expiration. We haven't figured out the best terms to use yet but we'll take all this into consideration before we reach stable.

> Also what happens if one forgets to check expiration, I would guess the scheduler calls us immediately again so it works, just slower, or is the problem more drastic.

If you forget to check if the work is expired, but you do check `timeRemaining()`, then it will still  work because `timeRemaining()` will be 0 (though I guess I'm missing a test for this). Maybe this implies that we should unify the two APIs into one (these are also inherited from `requestIdleCallback`). I think it's likely we'll replace both with a single `shouldYield()` method. But I do think it's useful to provide a estimate for how much time before the next frame. But that could be a separate API.

packages/scheduler/src/Scheduler.js
NE-SmallTown: @bvaughn I don't understand. The code I show above **don't** "do a conditional check inside of a very "hot" function", it just do the check **once** when the `index.js` which be bundled **first run**. And it reduce duplicate code. So it's a win-win thing

NE-SmallTown: @bvaughn Sorry for the typo, I mean:
`const now = hasNativePerformanceNow ? performance.now : Date.now`, later we can just use `now()`

PR #13725, opened 2018-09-25
[ReactUpdateQueue.js] Remove unreachable code

NE-SmallTown:
1. Because the `updateExpirationTime > renderExpirationTime` will always be `false`, so I remove the branch. If this is wrong, maybe we need add some tests. But IMO it really should always be `false` because the `update` is created by _Captured_(i.e. render throw), so the `updateExpirationTime` should always equals `renderExpirationTime` when we iterate CapturedUpdate

2. Because we won't enter the branch, the `newFirstCapturedUpdate` will always be `null`, so I just remove it.

### DIFF
--- a/packages/react-reconciler/src/ReactUpdateQueue.js
+++ b/packages/react-reconciler/src/ReactUpdateQueue.js
@@ -469,49 +469,28 @@ export function processUpdateQueue<State>(
   }
 
   // Separately, iterate though the list of captured updates.
-  let newFirstCapturedUpdate = null;
   update = queue.firstCapturedUpdate;
   while (update !== null) {
-    const updateExpirationTime = update.expirationTime;
-    if (updateExpirationTime < renderExpirationTime) {
-      // This update does not have sufficient priority. Skip it.
-      if (newFirstCapturedUpdate === null) {
-        // This is the first skipped captured update. It will be the first
-        // update in the new list.
-        newFirstCapturedUpdate = update;
-        // If this is the first update that was skipped, the current result is
-        // the new base state.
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactUpdateQueue.js
```diff
@@ -526,20 +502,16 @@ export function processUpdateQueue<State>(
...
+  if (newFirstUpdate === null) {
     // We processed every update, without skipping. That means the new base
     // state is the same as the result state.
     newBaseState = resultState;
   }
 
   queue.baseState = newBaseState;
   queue.firstUpdate = newFirstUpdate;
-  queue.firstCapturedUpdate = newFirstCapturedUpdate;
+  queue.firstCapturedUpdate = null;
```
NE-SmallTown: I don't know why even I remove this line i.e. `queue.firstCapturedUpdate = newFirstCapturedUpdate`, it still works, maybe the queue has been set to `null` after we executed this function? I don't know, don't dig into it. So I still remain this line

acdlite:
I will review this PR if you write a unit test that demonstrates the bug

NE-SmallTown:
Thanks for you feedback, I'm not sure I understand, the reason I don't add a unit test(I know the PR rules) is that this is a dead code like `if (false) { ... }`, I don't know how to test this with a unit test, so IMO I think all tests passed can show that theses changes of this PR are ok as I don't add anything or modify the logic. If you could give a direction about the unit test, I will be appreciated.

NE-SmallTown:
Hi @acdlite , do I need provide something else?

PR #13517, opened 2018-08-30
[ReactFiberScheduler.js] Remove redundant if else branch

NE-SmallTown:
The codes of [this if statement](https://github.com/facebook/react/blob/4bcee56210bd3ab3f6440ff7313a255148e22107/packages/react-reconciler/src/ReactFiberScheduler.js#L1661-L1667) is same with the [else if statement](https://github.com/facebook/react/blob/4bcee56210bd3ab3f6440ff7313a255148e22107/packages/react-reconciler/src/ReactFiberScheduler.js#L1668-L1674), so we merge them as one branch

PR #13726, opened 2018-09-25
refactor-scheduler

yongningfu:
just make the code more consise. merge the logic to the outside if.


NE-SmallTown:
There is already a PR #13517

PR #13728, opened 2018-09-25
Support class component static contextType attribute

### LINKED ISSUE #13336: Imperative wrappers can't access current context value in commit phase
Sometimes we have an imperative wrapper like this:

```js
componentDidMount() {
  renderSomethingImperatively(this.props)
}

componentDidUpdate() {
  renderSomethingImperatively(this.props)
}

render() {
  return null
}
```

Portals eliminated the need for this for regular DOM jumps. But we still need this for embedding renderers (e.g. `react-art` [does this](https://github.com/facebook/react/blob/3b3b7fcbbdc734d7f2a2e509a0b07c0eb053f779/packages/react-art/src/ReactART.js#L59-L84)) and use cases like "Vue inside React".

For cross-renderer embedding, maybe we could extend portals to do that (https://github.com/facebook/react/issues/13332). There are still imperative use cases for cross-library rendering though.

One thing that becomes annoying is that new context won't propagate down through this imperative boundary. This is because we don't maintain a stack in the commit phase. We're traversing a flat linked list of effects. So we don't actually know what context value is current by the time `componentDidMount` or `componentDidUpdate` fires.

For `react-art` and friends, this means context from a host app is not accessible. This is quite annoying. You could hack around it with something like

```js
<MyConsumer>
  {value =>
    <ReactART.Surface>
      <MyContext.Provider value={value}>
        <Stuff />
      </MyContext.Provider>
    </ReactART.Surface>
  }
</MyConsumer>
```

But this is neither obvious nor convenient. You have to anticipate all contexts that can get used below.

This seems even less convenient for imperative cases like "Vue inside React". 

```js
componentDidMount() {
  renderSomethingImperatively(this.props) // ???
}

componentDidUpdate() {
  renderSomethingImperatively(this.props) // ???
}

render() {
  // <MyConsumer>{value => ???}</MyConsumer>
  return <div />
}
```

Seems like you could use `unstable_read()` in `getDerivedStateFromProps` and that would put it into state so you can use it in lifecycles. So maybe that's sufficient. It still means you need to be explicit about which contexts you want to remember though.

I wonder if we can find a better solution to these use cases.

bvaughn:
**Note**: This is an experimental proposal and we haven't published an RFC for it yet. The main motivation is that continuing to support legacy context makes React slower and larger, and we'd like to be able to help everyone migrate to the new context API as soon as possible. Since the new context API is currently too low-level and/or verbose when used in classes, we are considering adding this higher-level API that feels more familiar and will simplify the migration.

-----

New static `contextType` property is supported for class components. This property can be set to the value returned by `React.createContext()` in order to consume the context value via `this.context`, e.g.:
```js
const LocaleContext = React.createContext('blue');

class LocalizedComponent extends React.Component {
  static contextType = LocaleContext;

  render() {
    return <div>The locale is: {this.context}</div>
  }
}
```

Resolves #13336

DEV warnings have been added for the following cases:
* Class component has `contextType` instance property.
* Class component has both `contextType` and `contextTypes` static properties.
* Class component has an invalid `contextType` static property (e.g. mistaken `contextType = Context.Provider`).
* Stateless functional component has an (unsupported) `contextType` property.

New tests have been added to verify the following:
* Class lifecycle `nextContext` params.
* Class lifecycle `this.context` values.
* Class is forcefully updated (without `shouldComponentUpdate` being called) when context provider updates.
* All of the above DEV warnings.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberClassComponent.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.js
@@ -516,7 +516,7 @@ function constructClassInstance(
   props: any,
   renderExpirationTime: ExpirationTime,
 ): any {
-  let isContextConsumer = false;
+  let isLegacyContextConsumer = false;
   let unmaskedContext = emptyContextObject;
   let context = null;
   const contextType = ctor.contextType;
@@ -529,8 +529,9 @@ function constructClassInstance(
   } else {
     unmaskedContext = getUnmaskedContext(workInProgress, ctor, true);
     const contextTypes = ctor.contextTypes;
-    isContextConsumer = contextTypes !== null && contextTypes !== undefined;
-    context = isContextConsumer
+    isLegacyContextConsumer =
+      contextTypes !== null && contextTypes !== undefined;
+    context = isLegacyContextConsumer
       ? getMaskedContext(workInProgress, unmaskedContext)
       : emptyContextObject;
   }
@@ -640,7 +641,7 @@ function constructClassInstance(
 
   // Cache unmasked context so we can avoid recreating masked context unless necessary.
   // ReactFiberContext usually updates this cache but can't for newly-created instances.
-  if (isContextConsumer) {
+  if (isLegacyContextConsumer) {
     cacheContext(workInProgress, unmaskedContext, context);
   }
 
--- a/packages/react-reconciler/src/ReactFiberClassComponent.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.js
@@ -65,6 +65,8 @@ let didWarnAboutUndefinedDerivedState;
 let warnOnUndefinedDerivedState;
 let warnOnInvalidCallback;
 let didWarnAboutDirectlyAssigningPropsToState;
+let didWarnAboutContextTypeAndContextTypes;
+let didWarnAboutInvalidateContextType;
 
 if (__DEV__) {
   didWarnAboutStateAssignmentForComponent = new Set();
@@ -73,6 +75,8 @@ if (__DEV__) {
   didWarnAboutLegacyLifecyclesAndDerivedState = new Set();
   didWarnAboutDirectlyAssigningPropsToState = new Set();
   didWarnAboutUndefinedDerivedState = new Set();
+  didWarnAboutContextTypeAndContextTypes = new Set();
+  didWarnAboutInvalidateContextType = new Set();
 
   const didWarnOnInvalidCallback = new Set();
 
@@ -234,7 +238,7 @@ function checkShouldComponentUpdate(
   newProps,
   oldState,
   newState,
-  nextLegacyContext,
+  nextContext,
 ) {
   const instance = workInProgress.stateNode;
   if (typeof instance.shouldComponentUpdate === 'function') {
@@ -242,7 +246,7 @@ function checkShouldComponentUpdate(
     const shouldUpdate = instance.shouldComponentUpdate(
       newProps,
       newState,
-      nextLegacyContext,
+      nextContext,
     );
     stopPhaseTimer();
 
@@ -319,6 +323,13 @@ function checkClassInstance(workInProgress: Fiber, ctor: any, newProps: any) {
         'property to define propTypes instead.',
       name,
     );
+    const noInstanceContextType = !instance.contextType;
+    warningWithoutStack(
+      noInstanceContextType,
+      'contextType was defined as an instance property on %s. Use a static ' +
+        'property to define contextType instead.',
+      name,
+    );
     const noInstanceContextTypes = !instance.contextTypes;
     warningWithoutStack(
       noInstanceContextTypes,
@@ -326,6 +337,36 @@ function checkClassInstance(workInProgress: Fiber, ctor: any, newProps: any) {
         'property to define contextTypes instead.',
       name,
     );
+
+    if (
+      ctor.contextType &&
+      ctor.contextTypes &&
+      !didWarnAboutContextTypeAndContextTypes.has(ctor)
+    ) {
+      didWarnAboutContextTypeAndContextTypes.add(ctor);
+      warningWithoutStack(
+        false,
+        '%s declares both contextTypes and contextType static properties. ' +
+          'The legacy contextTypes property will be ignored.',
+        name,
+      );
+    }
+
+    if (
+      ctor.contextType &&
+      typeof ctor.contextType.unstable_read !== 'function' &&
+      !didWarnAboutInvalidateContextType.has(ctor)
+    ) {
+      didWarnAboutInvalidateContextType.add(ctor);
+      warningWithoutStack(
+        false,
+        '%s defines an invalid contextType. ' +
+          'contextType should point to the Context object returned by React.createContext(). ' +
+          'Did you accidentally pass the Context.Provider instead?',
+        name,
+      );
+    }
+
     const noComponentShouldUpdate =
       typeof instance.componentShouldUpdate !== 'function';
     warningWithoutStack(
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -475,12 +516,24 @@ function constructClassInstance(
   props: any,
   renderExpirationTime: ExpirationTime,
 ): any {
-  const unmaskedContext = getUnmaskedContext(workInProgress, ctor, true);
-  const contextTypes = ctor.contextTypes;
-  const isContextConsumer = contextTypes !== null && contextTypes !== undefined;
-  const context = isContextConsumer
-    ? getMaskedContext(workInProgress, unmaskedContext)
-    : emptyContextObject;
+  let isContextConsumer = false;
```
acdlite: Can we name this `isLegacyContextConsumer` to be more specific?

gaearon:
Should we warn if you put `contextType` on a function? Also have a test that verifies it doesn't accidentally work?

acdlite:
> (I don't think there's a way to actually call sCU with the current implementation.)

Yeah but the point is to prevent the implementation from regressing :D

bvaughn:
That's fair! #13729

packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -668,12 +721,21 @@ function mountClassInstance(
...
   instance.props = newProps;
   instance.state = workInProgress.memoizedState;
   instance.refs = emptyRefsObject;
-  instance.context = getMaskedContext(workInProgress, unmaskedContext);
+
+  const contextType = ctor.contextType;
+  if (
+    typeof contextType === 'object' &&
+    contextType !== null &&
+    typeof contextType.unstable_read === 'function'
```
sebmarkbage: Do we need this extra check? Seems like we can just throw.

bvaughn: I guess we don't, since we also [provide a nice DEV mode warning if it's not a function](https://github.com/facebook/react/blob/0dc0ddc1ef5f90fe48b58f1a1ba753757961fc74/packages/react-reconciler/src/ReactFiberClassComponent.js#L355-L368). I'll remove this type check.

bvaughn: Actually, if we don't check in at least the constructor path, we'll throw _before_ our DEV warning. So I think it's best to leave these checks in place after all. If you feel strongly about it, let's talk and come up with another plan. I think the DEV warnings are useful though.

benwiley4000:
Was this API proposed only for migration purposes? I've been using new Context for awhile now, and it works great in most cases, but starts to feel boilerplately when I need to access context from inside a class method. The above API seems to capture that use case perfectly. I'd love to start using it.

gaearon:
Sure you can start using it. Yeah it’s mostly for easier migration but you can use it in either way if you like.

satya164:
Was it considered to support consuming multiple context providers with the API?

```js
static contextTypes = {
  foo: FooContext,
  bar: BarContext,
};

this.context.foo
this.context.bar
```

lushc:
Hopefully multiple contexts will be reconsidered in the future. I'm not sure if I've been misusing them but in larger apps I tend to have a `ThemeContext`, `UserContext`, and some sort of `FooContext` born out of a separation of concerns. In some components I need access to both `user` and `foo` in the lifecycle methods. This is solved by composing HOCs but it'd be nice to forego that in favour of this API 👍

brunolemos:
Yes please consider supporting multiple contexts. It's not so edge. I just made the exact same suggestion here before reading the comments above: https://github.com/reactjs/reactjs.org/pull/1265#issuecomment-431006796

bvaughn:
I'm curious why you'd use a class component for the decorator rather than the render prop API @arackaf?

arackaf:
Same reason you'd put a `contextType` static property on a class component, instead of using render props: you want or need to access the context value outside the render method, inside of component lifecycle methods.

bvaughn:
It's the same reason though in the case of an HOC like you mentioned above:
```js
const addContext = (name, context) => Orig => class extends Component {
  static contextType = context;
  render(){
    return <Orig {...this.props} name={this.context} />
  }
}
```
Because the `Component` component you create has no lifecycles. I can't see a reason not to just use the render props version in this case?

arackaf:
Ah - oops. Edited :)

bvaughn:
I think we're still misunderstanding each other. 😄

I'm asking why you would use _this_ form:
```js
const addContext = (name, context) => Orig => class extends Component {
  static contextType = context;
  render(){
    return <Orig {...this.props} name={this.context} />
  }
}
```

Instead of something based on the render prop API:
```js
const addContext = (propName, Context) => Orig => props => (
  <Context.Consumer>
    {value => <Orig {...props} {...{ [propName]: value }} />}
  </Context.Consumer>
);
```

arackaf:
Oh sorry - yeah, that's much nicer. Completely.  Thanks for clarifying!

bvaughn:
Cool, no problem! I was just wondering if I was missing something 😄

NE-SmallTown:
I think maybe one problem is HOC is not intrusive but render prop is

kana-sama:
@bvaughn, it is doesn't matter, why contextType instead of render prop, because all of them are dirty workarounds of this "edge case" restriction. It is a fighting with react API, I could say.

the-spyke:
To me it looks rather strange, when you pass an object which is not a `context` or `contextType`. Why not:
```js
import { DataConsumer } from "../data_context.js";

class MyComp extends React.Component {
  static contextConsumer = DataConsumer;
...
```

gaearon:
@kana-sama @the-spyke If you want to provide feedback on proposals before they're implemented, I suggest you to watch the RFC repo: https://github.com/reactjs/rfcs. This change has already been released. We're not monitoring this thread.

the-spyke:
> We're not monitoring this thread.

@gaearon How did you find out about new messages then? Just kidding. I'm sorry, I know about RFCs, but it's too late too.

bvaughn:
> @bvaughn, it is doesn't matter, why contextType instead of render prop, because all of them are dirty workarounds of this "edge case" restriction. It is a fighting with react API, I could say.

I don't really know what you're saying here, to be honest. Did I say it doesn't matter somewhere? My "edge case" comment was referring to a single component consuming multiple contexts, not that API in general.

Dan's right that feedback like this is more appropriate for the RFC, although this change in particular has already been released so it's a bit late for us to incorporate most feedback 😄

That being said, to clarify, the new `contextType` approach is mostly to ease adoption for people migrating away from the old `contextTypes` + `prop-types` approach. It's simpler for them to change `contextTypes` -> `contextType` (assuming only a single context) than it is to add a wrapper with a render prop and the `forwardRef` API.

This also avoids the overhead of an additional fiber allocation for the wrapper, but that's a very trivial concern _in most cases_.

Mati365:
@gaearon `contextType` will can be accessed in constructor of component?

bvaughn:
@Mati365 `contextType` is a static property of class components. I assume you meant to ask if the `context` param will be passed to the constructor for components that use `contextType`– in which case, the answer is yes. We have a unit test for that [here](https://github.com/facebook/react/blob/b5539ad6280f45d4eb025dbe0fc1bd2c00b06531/packages/react/src/__tests__/ReactContextValidator-test.js#L367-L439) if you're curious.

shijistar:
How to let the component be re-rendered in case `static contextType` changes? I don't see any sample like render prop API [dynamic-context](https://reactjs.org/docs/context.html#dynamic-context)

bvaughn:
The component will automatically re-render if its context value changes. 👍

dschinkel:
I do not want my tests or my architecture relying on context. This is outrageous.  My tests have been simple by passing in the store and to prevent that is ridiculous.  Unreal.

PR #13731, opened 2018-09-26
Ensure "addEventListener" exists on "window" for "scheduler" package

### LINKED ISSUE #13694: Schedule, SSR, window.addEventListener is not a function
Bug

window.addEventListener is not a function in node_modules/schedule/cjs/schedule.development.js:366:10

- Build server.bundle.js for ssr (node) as usual
- Try to run it with "node server.bundle.js"
- See "window.addEventListener is not a function in node_modules/schedule/cjs/schedule.development.js:366:10"
- console.log(window) 1 line before 366 returns "{ navigator: {}, server: true }"

working as usual

"react": "^16.3.2"
Well, it worked just fine before I did yarn upgrade, first since March, I guess.

trueadm:
This fixes #13694.

When the scheduler package initializes, it tries to attach an event listener to `window` via `addEventListener`. In some environments, the `window` object might be mocked and thus `addEventListener` might be missing. This PR adds a check to ensure that this function does exist before trying to attach the event.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -470,7 +470,7 @@ var getFrameDeadline;
 
 // "addEventListener" might not be available on the window object
 // if this is a mocked "window" object. So we need to validate that too.
-if (typeof window === 'undefined' || typeof window.addEventListener === 'function') {
+if (typeof window === 'undefined' || typeof window.addEventListener !== 'function') {
   // If this accidentally gets imported in a non-browser environment, fallback
   // to a naive implementation.
   var timeoutID = -1;
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -470,7 +470,10 @@ var getFrameDeadline;
 
 // "addEventListener" might not be available on the window object
 // if this is a mocked "window" object. So we need to validate that too.
-if (typeof window === 'undefined' || typeof window.addEventListener !== 'function') {
+if (
+  typeof window === 'undefined' ||
+  typeof window.addEventListener !== 'function'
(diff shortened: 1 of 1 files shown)


gaearon:
It would also be nice to get an understanding of what changed that made this a problem for RN.

trueadm:
@gaearon Why would you do it in the `typeof window === 'undefined'` condition, shouldn't it be done in the `else` of that condition (which is where it is)? If `window` is undefined then we never even touch the logic to add the event listener. My understanding is that it's because this package was pulled in to RN environments where they are using a mocked `window` that doesn't have `addEventListener` on it.

gaearon:
I mean that for our purposes “window without addEventListener” is just as bad as “undefined window”. So we should treat them the same way — by falling back.

gaearon:
>My understanding is that it's because this package was pulled in to RN environments 

This is what I’d like to clarify. It’s not clear to me how it got pulled in. Do we know where it’s imported?

trueadm:
@gaearon I understand what you mean now, thanks for clarifying.

So my primary fix for this was to remove the `require` to `react-dom` in RN etc, but there isn't one that I could see anywhere. The most likely case for `react-dom` being pulled in (which pulls in `scheduler`) is that people are using third-party React packages that pull it in (which we can't manage) so I applied the fix as a conditional (I've applied changes as per your feedback).

gaearon:
>The most likely case for react-dom being pulled in, which pulls in scheduler, is that people are using third-party React packages that pull it in.

Ah. This makes sense.

lumcory:
I think this issue has resurface itself in react-dom@16.13.1 which pulls scheduler@0.19.1

PR #13732, opened 2018-09-26
Rename AsyncMode -> ConcurrentMode

trueadm:
This PR renames `AsyncMode` to `ConcurrentMode`. Specifically:

- `React` exports `unstable_ConcurrentMode` now rather than `unstable_AsyncMode`.
- `ReactTestRender` now takes `unstable_isConcurrent` as an option rather than `unstable_isAsync`.
- Many instances of `isAsync` have been renamed to `isConcurrent` to reflect the naming changes.
- `REACT_ASYNC_MODE_TYPE` has been renamed to `REACT_CONCURRENT_MODE_TYPE`.
- `react-is` now exports `isConcurrentMode`, renamed from `isAsyncMode`.
- Tests have been updated to use the new naming scheme.

### DIFF
--- a/packages/react-is/src/ReactIs.js
+++ b/packages/react-is/src/ReactIs.js
@@ -77,7 +77,7 @@ export function isAsyncMode(object: any) {
     false,
     'The ReactIs.isAsyncMode() alias has been deprecated, ' +
       'and will be removed in React 17+. Update your code to use ' +
-      'ReactIs.isConcurrentMode() instead. It has the exact same API.'
+      'ReactIs.isConcurrentMode() instead. It has the exact same API.',
   );
   return isConcurrentMode(object);
 }
--- a/packages/react-is/src/ReactIs.js
+++ b/packages/react-is/src/ReactIs.js
@@ -75,14 +75,16 @@ let hasWarnedAboutDeprecatedIsAsyncMode = false;
 
 // AsyncMode should be deprecated
 export function isAsyncMode(object: any) {
-  if (!hasWarnedAboutDeprecatedIsAsyncMode) {
-    hasWarnedAboutDeprecatedIsAsyncMode = true;
-    lowPriorityWarning(
-      false,
-      'The ReactIs.isAsyncMode() alias has been deprecated, ' +
-        'and will be removed in React 17+. Update your code to use ' +
-        'ReactIs.isConcurrentMode() instead. It has the exact same API.',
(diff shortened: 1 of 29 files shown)


packages/react-is/src/ReactIs.js
```diff
@@ -68,8 +68,8 @@ export const StrictMode = REACT_STRICT_MODE_TYPE;
 
 export {isValidElementType};
 
-export function isAsyncMode(object: any) {
-  return typeOf(object) === REACT_ASYNC_MODE_TYPE;
+export function isConcurrentMode(object: any) {
```
gaearon: Both this and symbol export rename would be a breaking change technically. Unfortunately we forgot the unstable prefix here. Let’s keep both and add a deprecation warning for isAsyncMode.

trueadm: I've made the change. Did I do it correctly?

packages/react-is/src/ReactIs.js
```diff
@@ -71,14 +71,19 @@ export const StrictMode = REACT_STRICT_MODE_TYPE;
...
+
 // AsyncMode should be deprecated
 export function isAsyncMode(object: any) {
-  lowPriorityWarning(
-    false,
-    'The ReactIs.isAsyncMode() alias has been deprecated, ' +
-      'and will be removed in React 17+. Update your code to use ' +
-      'ReactIs.isConcurrentMode() instead. It has the exact same API.',
-  );
+  if (!hasWarnedAboutDeprecatedIsAsyncMode) {
```
gaearon: This should be inside `if (__DEV__)`

bvaughn:
Hey @trueadm. This change will require a corresponding change [in react-devtools](https://github.com/facebook/react-devtools/blob/c2db8dd5c13edd29444c72714b49cdb1073a1a44/backend/attachRendererFiber.js#L74-L75). Are you also going to own that?

trueadm:
@bvaughn Just found the comment. Sure!

bvaughn:
No. Both async and concurrent modes are "unstable_" (e.g. `React.unstable_AsyncMode`) so they're expected to potentially change between minor releases.

wheelo:
Are there any doc about the usage of this **ConcurrentMode** Mode?

gaearon:
No — it's unstable. We only document stable features.

PR #13740, opened 2018-09-27
[scheduler] Improve naive fallback version used in non-DOM environments

acdlite:
Added some tests for the non-DOM version of Scheduler that is used as a fallback, e.g. Jest. The tests use Jest's fake timers API:

- `jest.runAllTimers()` flushes all scheduled work.
- `jest.advanceTimersByTime(ms)` flushes only callbacks that expire within the given milliseconds.

These capabilities should be sufficient for most product tests. Because jest's fake timers do not override `performance.now` or `Date.now`, we assume time is constant. This means Scheduler's internal time will not be aligned with other code that reads from `performance.now`. For finer control, the user can override `window._sched` like we do in our tests. We will likely publish a Jest package that has this built in.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -468,30 +468,53 @@ var requestHostCallback;
 var cancelHostCallback;
 var getFrameDeadline;
 
-// "addEventListener" might not be available on the window object
-// if this is a mocked "window" object. So we need to validate that too.
-if (
-  typeof window === 'undefined' ||
-  typeof window.addEventListener !== 'function'
-) {
-  // If this accidentally gets imported in a non-browser environment, fallback
-  // to a naive implementation.
-  var timeoutID = -1;
-  requestHostCallback = function(callback, absoluteTimeout) {
-    timeoutID = setTimeout(callback, 0, true);
-  };
-  cancelHostCallback = function() {
-    clearTimeout(timeoutID);
-  };
-  getFrameDeadline = function() {
-    return 0;
-  };
-} else if (window._schedMock) {
+if (typeof window !== 'undefined' && window._schedMock) {
   // Dynamic injection, only for testing purposes.
   var impl = window._schedMock;
   requestHostCallback = impl[0];
   cancelHostCallback = impl[1];
   getFrameDeadline = impl[2];
+} else if (
+  // If Scheduler runs in a non-DOM environment, it falls back to a naive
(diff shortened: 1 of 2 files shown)


packages/scheduler/src/Scheduler.js
```diff
@@ -474,17 +474,42 @@ if (
...
+        _currentTime = ms;
+        cb(didTimeout);
+      } finally {
+        _currentTime = -1;
+      }
+    }
+  }
+  requestHostCallback = function(cb, ms) {
+    if (_currentTime !== -1) {
+      // Protect against re-entrancy. Jest's fake timer queue flushes timer
```
gaearon: Maybe remove mentions of Jest? If it’s meant to be generic.

packages/scheduler/src/__tests__/SchedulerNoDOM-test.internal.js
```diff
@@ -0,0 +1,145 @@
...
+    // Delete addEventListener to force us into the fallback mode.
+    window.addEventListener = undefined;
+    const Scheduler = require('scheduler');
+    scheduleCallback = Scheduler.unstable_scheduleCallback;
+    runWithPriority = Scheduler.unstable_runWithPriority;
+    ImmediatePriority = Scheduler.unstable_ImmediatePriority;
+    InteractivePriority = Scheduler.unstable_InteractivePriority;
+  });
+
+  it('runAllTimers only flushes some callbacks', () => {
```
bvaughn: nit: Ambiguous wording

packages/scheduler/src/Scheduler.js
```diff
@@ -474,17 +474,42 @@ if (
...
-  // to a naive implementation.
-  var timeoutID = -1;
-  requestHostCallback = function(callback, absoluteTimeout) {
-    timeoutID = setTimeout(callback, 0, true);
+  // If Scheduler runs in a non-DOM environment, it falls back to a naive
+  // implementation using setTimeout. This only meant to be used for testing
+  // purposes, like with jest's fake timer API.
+  var _callback = null;
+  var _currentTime = -1;
+  function flushCallback(didTimeout, ms) {
```
bvaughn: This function declaration fails the `no-inner-declarations` lint rule.

packages/scheduler/src/Scheduler.js
acdlite: Good point

Issue #13690, opened 2018-09-19
unmountComponentAtNode works async ?


qhxin:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

i tried to use `ReactDOM.unmountComponentAtNode` and `ReactDOM.render` to manage my multi roots. i found that if i use them in sync code, i get an error just like this when i change route pages:
```
Warning: unmountComponentAtNode(): The node you're attempting to unmount was rendered by React and is not a top-level container. Instead, have the parent component update its state and rerender in order to remove this component.
```

```
Warning: render(...): Replacing React-rendered children with a new root component. If you intended to update the children of this node, you should instead have the existing children update their state and render the new components instead of calling ReactDOM.render.
```

i use browserHistory to change route.
my code is written in componentDidMount method of a root React component:

Page1.js
```
  componentDidMount() {
    const a = document.getElementById('root2');
    ReactDOM.unmountComponentAtNode(a);
    ReactDOM.render(
      <A />,
      a
    );
    ReactDOM.render(
      <B />,
      document.getElementById('root3')
    );
  }
```

Page2.js
```
  componentDidMount() {
    const a = document.getElementById('root2');
    ReactDOM.unmountComponentAtNode(a);
    ReactDOM.render(
      <C />,
      a
    );
    ReactDOM.render(
      <B />,
      document.getElementById('root3')
    );
  }
```

And then, if i use `setTimeout` to wrap `ReactDOM.render` after `ReactDOM.unmountComponentAtNode(a);`, the result can be success.

**What is the expected behavior?**

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

react 16.5.2
react-dom 16.5.2

gaearon:
Can you please provide a single runnable complete example that reproduces the issue. For example as a fiddle.

qhxin:
@gaearon  this is my demo. thanks.

[![Edit o2p74974z](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/o2p74974z)

qhxin:
@gaearon  if i don't use unmountComponentAtNode, React Root can clear the old Component's events/doms/state and sub components by its self? or the old component will unmount automatically ？

gaearon:
If you can send a fix I’d be happy to verify it.

qhxin:
[...]

PR #13744, opened 2018-09-27
ReactRootContainer should not be set to null if root has been rerender

AyWa:
Follow up to https://github.com/facebook/react/issues/13690

- [x] add unit test to reproduce the issues
- [x] implement quick/dirty fix in order to have a first iteration
- [ ] implement proper fix

Detail about the issue:

- When we call `render/unmountComponentAtNode` inside an other life cycle that accept side effect like `ComponentDidMount`, it is queue in order to prevent reentrancy from https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberScheduler.js#L1982
- if for the same `DOMContainer`, we have, an `unmountComponentAtNode` follow by a `render`, they will be queue correctly (as explain just before)
- when we process the queue, it will be done in correct order and so `render` will be correctly executed after the `unmount`. However because **callback** are executed at the end of the `commitLifeCycles` we will execute the `unmountComponentAtNode` even if a render has been executed, and so we will set the `_reactRootContainer` to null. Obviously there is multiple side effect, like next `unmountComponentAtNode` might fail or we will not execute correctly all the lifecycle etc.

Proposal of fix:

As far as I know, the life cycle seems to be fine, the only problem is because callback are call in the end, we need a way to "invalidate" the `unmountComponentAtNode` callback

- solution 1 https://github.com/facebook/react/commit/2ee1335f02469a16166e27e89f09efb9e17ded5b -> just a big hack -> because we are just checking if the container has a child element -> just made for fix purpose and can be easily broken (if component return null)

- solution 2 current commit -> one way to do could be to mutate `_internalRoot` with a boolean like `rootShouldBeUnmount`(need bettername) and then if a render happen after the unmount call we set the boolean to false again -> still feel hacky

- solution 3 could be just better to remove the callback if unmount root ? but for now we don't track if the function is the unMount one or not. (we can add property isUnmountRoot etc but doesn't know if it is better)

- any better idea to experiment ?

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOM-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOM-test.js
@@ -525,4 +525,33 @@ describe('ReactDOM', () => {
         '    in App (at **)',
     ]);
   });
+
+  it('should not remove _reactRootContainer from a root if unmountComponentAtNode is follow by a render', () => {
+    const containerParent = document.createElement('div');
+    const containerApp = document.createElement('div');
+
+    function App1() {
+      return (
+        <div id="app1" />
+      );
+    }
+    function App2() {
+      return (
+        <div id="app2" />
+      );
+    }
+
+    function Parent() {
+      ReactDOM.render(<App1 />, containerApp);
+      ReactDOM.unmountComponentAtNode(containerApp);
+      ReactDOM.render(<App2 />, containerApp);
+      return (
+        <div />
+      );
+    }
+
+    ReactDOM.render(<Parent />, containerParent);
+    expect(containerApp._reactRootContainer).not.toBeNull();
+    expect(containerApp.innerHTML).toEqual('<div id=\"app2\"></div>');
+  });
 });
--- a/packages/react-dom/src/__tests__/ReactDOM-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOM-test.js
@@ -531,27 +531,23 @@ describe('ReactDOM', () => {
     const containerApp = document.createElement('div');
 
     function App1() {
-      return (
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/client/ReactDOM.js
```diff
@@ -564,6 +564,9 @@ function legacyRenderSubtreeIntoContainer(
         callback,
       );
     } else {
+      if (root.shouldBeUnMount && children != null) {
```
AyWa: Flow will fail there, but waiting to get `Proposal of fix:`  opinion

packages/react-dom/src/__tests__/ReactDOM-test.js
```diff
@@ -525,4 +525,29 @@ describe('ReactDOM', () => {
...
+      return <div id="app1" />;
+    }
+    function App2() {
+      return <div id="app2" />;
+    }
+
+    function Parent() {
+      ReactDOM.render(<App1 />, containerApp);
+      ReactDOM.unmountComponentAtNode(containerApp);
+      ReactDOM.render(<App2 />, containerApp);
```
gaearon: I wouldn't say this is a supported way to use React. There shouldn't be side effects in the rendering phase.

Can you come up with a repro case that is more valid? e.g. maybe you can reproduce this with a class component when calling `render` and `unmountComponentAtNode` inside `componentDidMount` or another lifecycle?

AyWa: Hi @gaearon,
Didn't have time to investigate more, but does the test is okay now ? or do I miss something ?

PR #13745, opened 2018-09-27
Remove useless code

NE-SmallTown:
Because only when `currentFirstChild !== null` we will do the `deleteRemainingChildren(returnFiber, currentFirstChild);` line which I delete in this PR. But delete a `null` child for a fiber is a no-op operation, so we can just remove this.

### DIFF
--- a/packages/react-reconciler/src/ReactChildFiber.js
+++ b/packages/react-reconciler/src/ReactChildFiber.js
@@ -1097,7 +1097,6 @@ function ChildReconciler(shouldTrackSideEffects) {
     }
     // The existing first child is not a text node so we need to create one
     // and delete the existing ones.
-    deleteRemainingChildren(returnFiber, currentFirstChild);
     const created = createFiberFromText(
       textContent,
       returnFiber.mode,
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -1095,9 +1095,7 @@ function ChildReconciler(shouldTrackSideEffects) {
       existing.return = returnFiber;
       return existing;
     }
-    // The existing first child is not a text node so we need to create one
-    // and delete the existing ones.
-    deleteRemainingChildren(returnFiber, currentFirstChild);
+    // The existing first child is not a text node so we need to create one.
```
acdlite: If `currentFirstChild !== null && currentFirstChild.tag !== HostTest`, then yes we do need to delete the remaining children.

NE-SmallTown: Yes, I don't delete that line(which in [line 1093](https://github.com/facebook/react/pull/13745/files#diff-f8ecb7d2b88a9e6edcab34023a47fe47R1093)). I delete the line below.
Or you just mean I show adjust the comment?

jcs090218: Look carefully to the code. I think @acdlite is saying we need to keep `deleteRemainingChildren(returnFiber, currentFirstChild);` in both condition? They are returning the `existing;` on line 1096, which the code would not reach.

PR #13746, opened 2018-09-27
Enable getDerivedStateFromError

bvaughn:
Commits are split up for easier review:
- [x] 4ed84b0: Remove the `enableGetDerivedStateFromCatch` feature flag.
- [x] cd49672: Copy-paste `ReactErrorBoundaries` to `ReactLegacyErrorBoundaries` to preserve tests for legacy `componentDidCatch` recovery behavior. (We will eventually remove most of these tests when `componentDidCatch` no longer supports state updates.)
- [x] 56bbf1c: Update previous `ReactErrorBoundaries` tests to cover new `getDerivedStateFromCatch` render-phase behavior and fixed a minor reconciliation bug that testing uncovered.
- [x] f6fb111: Rename `getDerivedStateFromCatch` to `getDerivedStateFromError`.
- [x] [`3df17ba`](https://github.com/facebook/react/pull/13746/commits/3df17ba43eaa1d77e1f4ffa14f27331c9683d059): Add DEV warning if an error boundary specifies _only_ the `componentDidCatch` lifecycle and does not trigger a state update (because this would swallow the error).
- [ ] Add a test for re-throwing an error from `getDerivedStateFromError` to ensure that error bubbling works as expected.

(The only commits that aren't just rote find-and-replace are 56bbf1c and [`3df17ba`](https://github.com/facebook/react/pull/13746/commits/3df17ba43eaa1d77e1f4ffa14f27331c9683d059).)

### DIFF
--- a/packages/react-dom/src/__tests__/ReactLegacyErrorBoundaries-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactLegacyErrorBoundaries-test.internal.js
@@ -2004,14 +2004,19 @@ describe('ReactLegacyErrorBoundaries', () => {
 
   it('renders empty output if error boundary does not handle the error', () => {
     const container = document.createElement('div');
-    ReactDOM.render(
-      <div>
-        Sibling
-        <NoopErrorBoundary>
-          <BrokenRender />
-        </NoopErrorBoundary>
-      </div>,
-      container,
+    expect(() => {
+      ReactDOM.render(
+        <div>
+          Sibling
+          <NoopErrorBoundary>
+            <BrokenRender />
+          </NoopErrorBoundary>
+        </div>,
+        container,
+      );
+    }).toWarnDev(
+      'ErrorBoundary: Error boundaries should implement getDerivedStateFromError()',
+      {withoutStack: true},
     );
     expect(container.firstChild.textContent).toBe('Sibling');
     expect(log).toEqual([
--- a/packages/react-dom/src/__tests__/ReactLegacyErrorBoundaries-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactLegacyErrorBoundaries-test.internal.js
@@ -1,0 +1,2125 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails react-core
+ */
+
+'use strict';
+
+let PropTypes;
+let React;
+let ReactDOM;
+let ReactFeatureFlags;
+
+// TODO: Refactor this test once componentDidCatch setState is deprecated.
+describe('ReactLegacyErrorBoundaries', () => {
+  let log;
+
+  let BrokenConstructor;
+  let BrokenComponentWillMount;
+  let BrokenComponentDidMount;
+  let BrokenComponentWillReceiveProps;
+  let BrokenComponentWillUpdate;
+  let BrokenComponentDidUpdate;
+  let BrokenComponentWillUnmount;
+  let BrokenRenderErrorBoundary;
+  let BrokenComponentWillMountErrorBoundary;
+  let BrokenComponentDidMountErrorBoundary;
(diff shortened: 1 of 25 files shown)


packages/react-reconciler/src/__tests__/ErrorBoundaryReconciliation-test.internal.js
```diff
@@ -0,0 +1,111 @@
...
+          </ErrorBoundary>,
+        );
+        if (isConcurrent) {
+          renderer.unstable_flushAll();
+        }
+      }).toWarnDev(isConcurrent ? ['invalid', 'invalid'] : ['invalid']);
+      expect(renderer).toHaveRenderedChildren(
+        React.createElement(fallbackTagName, {prop: 'ErrorBoundary'}),
+      );
+    }
```
bvaughn: This test was useful for me since my initial "fix" to sync mode broke concurrent mode. I'm not sure about this amount of abstraction though. On the one hand, it makes it very easy to spot what's different between these otherwise very similar tests. On the other hand, it may be harder to read.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -425,7 +423,15 @@ function finishClassComponent(
...
-    workInProgress.child = null;
+    // Forcefully reset children so that a subsequent reconciliation will always re-add.
+    // This is important if e.g. an error boundary renders an element of the same type.
+    // Concurrent renderer mode will always retry an extra time on failure,
+    // so the fiber we need to reset varies.
+    if (workInProgress.mode & ConcurrentMode) {
+      workInProgress.child = null;
+    } else {
+      current.child = null;
+    }
```
bvaughn: I'm not super confident about this fix ^

packages/react-dom/src/__tests__/ReactLegacyErrorBoundaries-test.internal.js
```diff
@@ -0,0 +1,2125 @@
+/**
```
bvaughn: This file was literally copied from `ReactErrorBoundaries-test.internal.js` so it doesn't really need to be reviewed. Normally I would have `git mv`ed it, but in this case– I wanted the reviewer to be able to easily spot the places the new render-phase error boundary behavior _differed_ from the old behavior.

bvaughn:
Rebased and added the DEV warning we talked about.

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -135,6 +127,21 @@ function createClassErrorUpdate(
...
         componentStack: stack !== null ? stack : '',
       });
+      if (__DEV__) {
+        if (typeof getDerivedStateFromError !== 'function') {
+          // If componentDidCatch is the only error boundary method defined,
+          // then it needs to call setState to recover from errors.
+          // If no state update is scheduled then the boundary will swallow the error.
+          const updateQueue = fiber.updateQueue;
+          warningWithoutStack(
+            updateQueue !== null && updateQueue.firstUpdate !== null,
```
acdlite: This check isn't sufficient because `firstUpdate` might be a low priority update. It needs to be sync. You can check `fiber.expirationTime === Sync` instead.

bisubus:
Greetings, just noticed that this PR has been blocking mine, facebook/react#13986 . Incidentally, second errorInfo param still isn't in use, https://github.com/facebook/react/pull/13746/files#diff-1996f2b11f9c68c0a81652e32be88ddbR109 . I could fix this in my PR and make `getDerivedStateFromError(error, errorInfo, prevState)` compatible with componentDidCatch(error, errorInfo)` signature. @bvaughn @acdlite What do you think?

bvaughn:
@bisubus How is this PR blocking anything? 😄 It was merged four months ago.

bisubus:
@bvaughn Bad wording, it's the merge that was blocked. And it was a year and a few months, to be precise, I apologize for a blast from the past. Any way, I think I addressed the problem with errorInfo from this PR.

PR #13747, opened 2018-09-27
Refactor Scheduler.js

alexnoz:
es6+ refactoring of `Scheduler.js`


aweary:
Thanks for taking the time to do this @alexnoz! This file was intentionally written to avoid using ES2015+. See https://github.com/facebook/react/pull/13582 for some context.

PR #13748, opened 2018-09-27
pure

acdlite:
A higher-order component version of the `React.PureComponent` class. During an update, the previous props are compared to the new props. If they are the same, React will skip rendering the component and its children.

Unlike userspace implementations, `pure` will not add an additional fiber to the tree.

The first argument must be a functional component; it does not work with classes. (TODO: Add a warning.)

`pure` uses shallow comparison by default, like `React.PureComponent`. A custom comparison can be passed as the second argument.


packages/react-reconciler/src/ReactFiber.js
```diff
@@ -363,15 +368,8 @@ export function createWorkInProgress(
     }
   }
 
-  // Don't touching the subtree's expiration time, which has not changed.
   workInProgress.childExpirationTime = current.childExpirationTime;
-  if (pendingProps !== current.pendingProps) {
```
sophiebits: What's the significance of this change?

sophiebits: What's the significance of this change?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -124,7 +127,7 @@ export function reconcileChildren(
   current: Fiber | null,
   workInProgress: Fiber,
   nextChildren: any,
-  renderExpirationTime: ExpirationTime,
```
sophiebits: Why is this here? Did prettier change?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1103,62 +1177,68 @@ function memoizeState(workInProgress: Fiber, nextState: any) {
...
-        break;
-      case HostComponent:
-        pushHostContext(workInProgress);
-        break;
-      case ClassComponent: {
-        const Component = workInProgress.type;
-        if (isLegacyContextProvider(Component)) {
-          pushLegacyContextProvider(workInProgress);
+
+  if (current !== null) {
```
sophiebits: I don't understand the significance of these changes either.

packages/react/src/pure.js
```diff
@@ -0,0 +1,40 @@
...
+      warningWithoutStack(
+        false,
+        'pure: The first argument must be a functional component. Instead ' +
+          'received: %s',
+        render === null ? 'null' : typeof render,
+      );
+    } else {
+      const prototype = render.prototype;
+      if (!!(prototype && prototype.isReactComponent)) {
+        warningWithoutStack(
```
sophiebits: Unit tests for the warnings pls?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1275,19 +1357,45 @@ function beginWork(
...
+      return updatePureComponent(
+        current,
+        workInProgress,
+        type,
+        workInProgress.pendingProps,
+        updateExpirationTime,
+        renderExpirationTime
       );
+    }
+    case PureComponentLazy: {
```
sophiebits: this duplication kills me a little bit :)

packages/react-reconciler/src/ReactFiber.js
acdlite: Previously the `expirationTime` field represented the pending priority of props, state, and context. In the begin phase, we check if this is lower priority than the render priority. If so, we can bail out.

After this change, it represents only the pending priority of state and context. The props check is moved to the begin phase. That way, in the begin phase, we can distinguish between an update caused by props versus an update caused by state or context. `pure` will never bail out if context or props has changed.

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: Corresponds to the changes in https://github.com/facebook/react/pull/13748/files#r221093037

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: Me too :D

packages/react/src/pure.js
sophiebits: sorry I could have also done this

dimitropoulos:
you missed a golden opportunity to call this `WithPurity`, my friend.  thanks for the great stuff @acdlite

alexeyraspopov:
It seems to be the first time React includes HOC in its public API. Do we need to expect some sort of new guides on when and how to use it?

NE-SmallTown:
Agree with @alexbaumgertner , they will do that but just seems don't have much time currently, before they release they there will be blog post

bvaughn:
@acdlite Will DevTools need to be updated to recognize the new `REACT_PURE_TYPE`?

gaearon:
@alexeyraspopov We always include guides/docs/etc with releases. This is not a release yet.

bvaughn:
@acdlite Follow up https://github.com/facebook/react-devtools/pull/1173

probablyup:
Maybe I'm misremembering, but wasn't one of the design goals of SFC to be pure by default?

gaearon:
The goal was to make components easier to optimize. There are many ways to optimize, memoization based on shallow equality is just one possible optimization you can make.

Functions already have less overhead since 16 (less property checks and memory usage), and they're also much more optimizable by compilers. Unfortunately our compilation experiments are still very early.

probablyup:
Thanks for the explainer @gaearon 😄

PR #13756, opened 2018-10-01
--save is no longer needed

JoeCortopassi:
`--save` is on by default as of [npm 5](https://blog.npmjs.org/post/161081169345/v500), and `npm install aphrodite` is functionally equivalent to `npm install --save aphrodite` now


gaearon:
Sure, thanks

PR #13758, opened 2018-10-02
Simplify ReactFiberClassComponent logic

mmarkelov:
Add `enqueue` function helper to remove duplicates.


NE-SmallTown:
IIRC, they don't do the change which just about code style. On the other hand, this will increase the call stack and make the optimizing become more difficult(because the hot path of them is different)

PR #13759, opened 2018-10-02
Add jest-scheduler package for testing the scheduler package with Jest

bvaughn:
I've noticed that FB infra teams need to write the same kinds of tests internally as I've written within React. I expect this will also be true for OSS lib maintainers. So here's a stab at some shareable Jest matchers in the form of a new package.

**The interesting commit in this diff is 1a489e5** (where the new matchers were added). The second commit (1a489e5) was just a rote find-and-replace of our old matchers with the new ones for pre-existing tests and doesn't really need to be reviewed.

**Note that I will be on PTO from Oct 4th - Oct 18th and will be unresponsive to PR feedback. All React core team members have been added as collaborators to my fork. Please feel free to make changes directly to this branch before merging.**


packages/jest-scheduler/src/__tests__/JestScheduler-test.js
```diff
@@ -0,0 +1,164 @@
...
+    it('should work with an empty set of interactions', () => {
+      // Sanity test.
+      expect(getCurrent()).toHaveBeenTracedWith([]);
+    });
+
+    it('should trace interactions through callbacks', done => {
+      function onFinished(value) {
+        expect(value).toBe('something');
+
+        // Verify that the callback has been called with the expected interactions.
```
bvaughn: I added inline comments initially, as I wrote these tests, thinking I might point people to the test as a reference in lieu of docs– but maybe I should remove them?

bvaughn:
I guess this wasn't compelling enough to warrant a review.

Issue #13753, opened 2018-09-30
Provide UMD build of simple-cache-provider


kentcdodds:
I expect this is planned. Just wanted to make sure to file this to make sure it's not missed.

**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

There's only a cjs build of simple-cache-provider: https://unpkg.com/simple-cache-provider@0.10.0/

**What is the expected behavior?**

react and react-dom have a UMD build which makes it easy to test things out or teach people React in a simple `index.html` file without any tooling to complicate things. Having a UMD build of simple-cache-provider would help when teaching suspense.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

N/A

PR #13761, opened 2018-10-03
UMD react-cache build

mmarkelov:
Add UMD build for `react-cache`
#13753


gaearon:
Thanks. Have you checked it works?

mmarkelov:
@gaearon yes. It was just fine

PR #13766, opened 2018-10-04
Add envify transform to scheduler package

mridgway:
This package uses `process.env.NODE_ENV` but does not transform its usage during bundling like the rest of the React libraries do. This causes issues when `process` is not defined globally.


acdlite:
Thanks!

PR #13120, opened 2018-06-28
Preserve state of placeholders by hiding timed-out children instead of deleting them

acdlite:
When a placeholder times out, instead of unmounting the current tree, hide it and render the fallback view as a sibling. When the data resolves and the tree resumes rendering, the children are made visible again. Because they remain mounted throughout, the state is preserved.

Adds additional host config methods. For mutation mode:

- hideInstance
- hideTextInstance
- unhideInstance
- unhideTextInstance

For persistent mode:

- cloneHiddenInstance
- cloneUnhiddenInstance
- createHiddenTextInstance

### TODO

- [x] Need more comments.
- [ ] Implement new host config methods for each renderer:
  - [x] Noop
  - [x] DOM
  - [x] ART
  - [ ] Native
  - [ ] Fabric
  - [x] Test
    - Idk how this should be exposed publicly yet so all I did was add an `isHidden` property to the instance types.

PR #13767, opened 2018-10-04
Suspense settles even if timeout and ping coincide

sophiebits:
Instead of using effect tags, we use `.updateQueue !== null` to indicate the fallback should be rendered for strict mode (in addition to non-strict mode which already did).

(i have no idea what I'm doing)

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -949,27 +949,19 @@ function updatePlaceholderComponent(
 
     // Check if we already attempted to render the normal state. If we did,
     // and we timed out, render the placeholder state.
-    const alreadyCaptured =
-      (workInProgress.effectTag & DidCapture) === NoEffect;
+    const nextDidTimeout = workInProgress.updateQueue !== null;
 
-    let nextDidTimeout;
-    if (current !== null && workInProgress.updateQueue !== null) {
-      if (enableSchedulerTracing) {
+    if ((workInProgress.mode & StrictMode) === NoEffect) {
+      if (enableSchedulerTracing && nextDidTimeout) {
         // Handle special case of rendering a Placeholder for a sync, suspended tree.
         // We flag this to properly trace and count interactions.
         // Otherwise interaction pending count will be decremented too many times.
         captureWillSyncRenderPlaceholder();
       }
-
-      // We're outside strict mode. Something inside this Placeholder boundary
-      // suspended during the last commit. Switch to the placholder.
+      // The next time we render, we try the primary children again even if no
+      // promise has timed out.
       workInProgress.updateQueue = null;
-      nextDidTimeout = true;
     } else {
-      nextDidTimeout = !alreadyCaptured;
-    }
-
-    if ((workInProgress.mode & StrictMode) !== NoEffect) {
       if (nextDidTimeout) {
         // If the timed-out view commits, schedule an update effect to record
         // the committed time.
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -356,11 +356,7 @@ function commitLifeCycles(
       if (enableSuspense) {
         if ((finishedWork.mode & StrictMode) === NoEffect) {
           // In loose mode, a placeholder times out by scheduling a synchronous
-          // update in the commit phase. Use `updateQueue` field to signal that
-          // the Timeout needs to switch to the placeholder. We don't need an
-          // entire queue. Any non-null value works.
-          // $FlowFixMe - Intentionally using a value other than an UpdateQueue.
-          finishedWork.updateQueue = emptyObject;
+          // update in the commit phase.
           scheduleWork(finishedWork, Sync);
         } else {
           // In strict mode, the Update effect is used to record the time at
--- a/packages/react-reconciler/src/ReactFiberUnwindWork.js
+++ b/packages/react-reconciler/src/ReactFiberUnwindWork.js
@@ -72,6 +72,8 @@ import {
 import {findEarliestOutstandingPriorityLevel} from './ReactFiberPendingPriority';
 import {reconcileChildren} from './ReactFiberBeginWork';
 
+const emptyObject = {};
+
 function NoopComponent() {
   return null;
 }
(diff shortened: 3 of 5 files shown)


sebmarkbage:
This doesn’t seem right. The strict mode and non-strict mode have different implementations.

sophiebits:
There is a branch in commitWork where we schedule a new update upon commit for non-strict mode. But in both cases our next render after catching should be the fallback, hence the unified logic in throwException. In non-strict mode we do want to render the primary content again after the fallback if a new render is triggered, which is why we (still) mutate updateQueue in updatePlaceholderComponent.

 I can make the logic in retrySuspendedRoot conditional on mode but this already passes all the tests.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -356,11 +356,7 @@ function commitLifeCycles(
       if (enableSuspense) {
         if ((finishedWork.mode & StrictMode) === NoEffect) {
           // In loose mode, a placeholder times out by scheduling a synchronous
-          // update in the commit phase. Use `updateQueue` field to signal that
-          // the Timeout needs to switch to the placeholder. We don't need an
-          // entire queue. Any non-null value works.
-          // $FlowFixMe - Intentionally using a value other than an UpdateQueue.
-          finishedWork.updateQueue = emptyObject;
```
sebmarkbage: This should never have been landed.

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -232,6 +234,12 @@ function throwException(
           );
           thenable.then(onResolveOrReject, onResolveOrReject);
 
+          // Use `updateQueue` field to signal that the Timeout needs to switch
+          // to the placeholder. We don't need an entire queue. Any non-null
+          // value works.
+          // $FlowFixMe - Intentionally using a value other than an UpdateQueue.
+          workInProgress.updateQueue = emptyObject;
```
sebmarkbage: So this was bad enough when it was lose mode only but we shouldn't promote this hack to strict mode.

sebmarkbage:
So the idea is that updateQueue should never be used in StrictMode and therefore when we remove non-StrictMode we can also remove the hack.

I don't understand what bug we're fixing. Can you restate the bug? Is it happening in StrictMode or non-StrictMode?

sophiebits:
The bug is that in strict mode, it is possible for the fallback to render and stay onscreen forever despite the promise resolving.

This seems to happen when the promise resolves just as the fallback is expiring to come onscreen. I don’t have a fuller understanding of the bug.

sophiebits:
I guess I don’t fully understand what is *supposed* to happen with ShouldCapture, DidCapture, and expiration times. How are we meant to know which to render?

sebmarkbage:
When the reconciliation steps into this node (beginWork, the first time) this should always be empty or reset. As it traverses inside this node it might set ShouldCapture. When it completes, it will process the capture. It should set the DidCapture flag. Then it will step into itself again a second (beginWork, the second time). Since the flag is set, this second pass will be different.

It's kind of like call return used to work. We step into this subtree several times.

sebmarkbage:
If a new update gets scheduled while we're inside of it, it should be either lower priority, or we should restart the whole tree.

acdlite:
I think I understand what the root problem is. I’ll write a fuller comment when I get into the office. Short version: I think we should go back to using an update queue to model pings.

But before we decide on a fix, I want to make sure we have tests that cover all the cases that are broken. There are some subtle ones that I hadn’t previously considered. I’ll work on that.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -949,27 +949,19 @@ function updatePlaceholderComponent(
...
-      // suspended during the last commit. Switch to the placholder.
+      // The next time we render, we try the primary children again even if no
+      // promise has timed out.
       workInProgress.updateQueue = null;
-      nextDidTimeout = true;
     } else {
-      nextDidTimeout = !alreadyCaptured;
-    }
-
-    if ((workInProgress.mode & StrictMode) !== NoEffect) {
```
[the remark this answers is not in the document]
acdlite: Flow would have caught this if it were an opaque type, but since it's a bitmask we can't do that.

packages/react-reconciler/src/ReactFiberCommitWork.js
acdlite: An explanation for why you don't like this hack leads to better software and smarter teammates!

acdlite:
When a ping happens ("ping" means the promise resolves, so we should attempt to render the placeholder again)), we mark the fiber as "dirty" by increasing the priority of its `expirationTime` field, and traverse back through to the root. We also set the pending priority on all the ancestors (using `childExpirationTime`).  This part is the same as a typical update.

Unlike a typical update, we don't add anything to the update queue. We used to, but both for performance and modeling purposes, we moved away from that model. The idea is that a ping doesn't have payload; it doesn't represent or contain any data, it's only a signal to re-render. In that way, it's kind of like `forceUpdate`: it means some other mutable source, in this case a cache, has been updated.

I'm not ready to give up on this model, but I don't think we fully thought through the implications of concurrent pings and updates. Some cases to consider:

- What if a promise resolves before the end of the current render phase, while React has yielded?
- What if a thenable resolves synchronously inside the current render phase? (An edge case, but still important. Even just for modeling purposes.)
- What if a thenable resolves synchronously when calling `then`? (We should special case this one.)
- What if a placeholder is pinged, but then the placeholder receives a higher priority update? Does the update priority get dropped when we complete the higher priority?
- What if a placeholder times out, and the original update that suspended the render was part of the primary children, which are then unmounted? Is it possible this priority level gets lost, maybe if combined with the previous case? (If so, storing the two sets of children separately as in https://github.com/facebook/react/pull/13120 would fix it)

acdlite:
@sebmarkbage 

> If a new update gets scheduled while we're inside of it, it should be either lower priority, or we should restart the whole tree.

My instinct is that we should restart, but not sure. If it happens too often, we could end up restarting many times, whereas if we ping a lower priority, the restarts are batched. Also, if we restart, render phases of subsequent components are interrupted, so if they contain network fetches, those aren't initiated until later, reducing parallelization.

PR #13774, opened 2018-10-04
Restart from root if promise pings before end of render phase

acdlite:
Possible alternative to https://github.com/facebook/react/pull/13767

See my comment here: https://github.com/facebook/react/pull/13767#issuecomment-427094770:

> When a ping happens ("ping" means the promise resolves, so we should attempt to render the placeholder again), we mark the fiber as "dirty" by increasing the priority of its `expirationTime` field, and traverse back through to the root. We also set the pending priority on all the ancestors (using `childExpirationTime`).  This part is the same as a typical update.
> 
> Unlike a typical update, we don't add anything to the update queue. We used to, but both for performance and modeling purposes, we moved away from that model. The idea is that a ping doesn't have a payload; it doesn't represent or contain any data, it's only a signal to re-render. In that way, it's kind of like `forceUpdate`: it means some other mutable source, in this case a cache, has been updated.
> 
> I'm not ready to give up on this model, but I don't think we fully thought through the implications of concurrent pings and updates. Some cases to consider:
> 
>  - What if a promise resolves before the end of the current render phase, while React has yielded?
>  - What if a thenable resolves synchronously inside the current render phase? (An edge case, but still important. Even just for modeling purposes.)
>  - What if a thenable resolves synchronously when calling `then`? (We should special case this one.)
>  - What if a placeholder is pinged, but subsequently the placeholder re-renders at a higher priority? Does the priority of the ping get dropped when we complete the placeholder fiber at the higher priority? For a typical update, the lower priority would not get dropped because we read the remaining priority from the update queue, but we don't have an update queue in this case.
>  - What if a placeholder times out, and the original update that suspended the render was part of the primary children, which are then unmounted? Is it possible this priority level gets lost, maybe if combined with the previous case? (If so, storing the two sets of children separately as in https://github.com/facebook/react/pull/13120 would fix it)

Will add more tests

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
@@ -649,6 +649,22 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     expect(ReactNoop.getChildren()).toEqual([span('Async'), span('Sync')]);
   });
 
+  it('resolves successfully even if fallback render is pending', async () => {
+    ReactNoop.render(
+      <Placeholder delayMs={1000} fallback={<Text text="Loading..." />}>
+        <AsyncText text="Async" ms={3000} />
+      </Placeholder>,
+    );
+    expect(ReactNoop.flushNextYield()).toEqual(['Suspend! [Async]']);
+    await advanceTimers(1500);
+    expect(ReactNoop.expire(1500)).toEqual([]);
+    // Before we have a chance to flush, the promise resolves.
+    await advanceTimers(2000);
+    expect(ReactNoop.clearYields()).toEqual(['Promise resolved [Async]']);
+    expect(ReactNoop.flush()).toEqual(['Async']);
+    expect(ReactNoop.getChildren()).toEqual([span('Async')]);
+  });
+
   it('throws a helpful error when an update is suspends without a placeholder', () => {
     expect(() => {
       ReactNoop.flushSync(() =>
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -111,7 +111,7 @@ import {
   computeAsyncExpiration,
   computeInteractiveExpiration,
 } from './ReactFiberExpirationTime';
-import {ConcurrentMode, ProfileMode} from './ReactTypeOfMode';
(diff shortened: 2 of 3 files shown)


sophiebits:
Does it pass the test I had too?

acdlite:
@sophiebits Yes, just pushed a commit that includes your test

sophiebits:
Answer: it does fix my test.

I don't know how to port over the .expire() call to the test renderer though…

```js
  it('resolves successfully even if fallback render is pending', async () => {
    const root = ReactTestRenderer.create(
      <Placeholder delayMs={1000} fallback={<Text text="Loading..." />}>
        <AsyncText text="Async" ms={3000} />
      </Placeholder>,
      {unstable_isConcurrent: true}
    );
    expect(root).toFlushAndYieldThrough(['Suspend! [Async]']);
    jest.advanceTimersByTime(1500);
    //expect(ReactNoop.expire(1500)).toEqual([]);
     // Before we have a chance to flush, the promise resolves.
    jest.advanceTimersByTime(2000);
    expect(ReactTestRenderer).toHaveYielded(['Promise resolved [Async]']);
    expect(root).toFlushAndYield(['Async']);
    expect(root).toMatchRenderedOutput('Async');
  });
```

sophiebits:
no you didn't

acdlite:
Lol, ok now I did :D

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -649,6 +649,22 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     expect(ReactNoop.getChildren()).toEqual([span('Async'), span('Sync')]);
   });
 
+  fit('resolves successfully even if fallback render is pending', async () => {
```
sophiebits: fit

acdlite:
> I don't know how to port over the .expire() call to the test renderer though

Yeah it doesn't support that yet. Haven't decided if or how it should support it, since expiration isn't something a typical test should worry about unless you're testing expiration itself

sebmarkbage:
The patch makes sense to me.

acdlite:
Ok I'll go ahead and merge this. Some follow ups:

- Thenables that resolve synchronously during the render phase (as opposed to in a microtask). This is a special case so I'll do this in a separate PR.
- Updates that get dropped when the placeholder times out and deletes the fiber that contains update. Will follow-up on this when I pick up the task in https://github.com/facebook/react/pull/13120.

PR #13775, opened 2018-10-04
Terminology: Functional -> Function Component

gaearon:
Same as https://github.com/reactjs/reactjs.org/pull/863.
Sema doesn't mind.


gaearon:
Maybe also drop "stateless" in warnings? It's clunky, doesn't make the warning clearer, and promotes the SFC abbreviation which sounds super complicated as official. Everybody else calls this a "function".

acdlite:
Yes plz

PR #13776, opened 2018-10-04
Wrap retrySuspendedRoot using SchedulerTracing

sophiebits:
Previously, we were emptying root.pendingInteractionMap and permanently losing those interactions when applying an unrelated update to a tree that has no scheduled work that is waiting on promise resolution. (That is, one that is showing a fallback and waiting for the suspended content to resolve.)

The logic I'm leaving untouched with `nextRenderIncludesTimedOutPlaceholder` is *not* correct -- what we want is instead to know if *any* placeholder anywhere in the tree is showing its fallback -- but we don't currently have a better replacement, and this should unblock tracing with suspense again.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -257,11 +257,6 @@ let legacyErrorBoundariesThatAlreadyFailed: Set<mixed> | null = null;
 // Used for performance tracking.
 let interruptedBy: Fiber | null = null;
 
-// Do not decrement interaction counts in the event of suspense timeouts.
-// This would lead to prematurely calling the interaction-complete hook.
-// Instead we wait until the suspended promise has resolved.
-let nextRenderIncludesTimedOutPlaceholder: boolean = false;
-
 let stashedWorkInProgressProperties;
 let replayUnitOfWork;
 let isReplayingFailedUnitOfWork;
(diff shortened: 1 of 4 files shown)


packages/react/src/__tests__/ReactProfiler-test.internal.js
```diff
@@ -2631,6 +2646,11 @@ describe('Profiler', () => {
             },
           );
         });
+        expect(ReactTestRenderer).toHaveYielded([
```
sophiebits: The timing of these yields is unchanged; the old test was misleading.

packages/react/src/__tests__/ReactProfiler-test.internal.js
```diff
@@ -2639,32 +2659,28 @@ describe('Profiler', () => {
...
-          'Promise resolved [loaded]',
-        ]);
+        expect(ReactTestRenderer).toHaveYielded(['Promise resolved [loaded]']);
         expect(renderer).toFlushAndYield(['AsyncText [loaded]']);
         expect(renderer.toJSON()).toEqual(['loaded', 'updated']);
 
         expect(onRender).toHaveBeenCalledTimes(1);
         expect(onRender.mock.calls[0][6]).toMatchInteractions([
           initialRenderInteraction,
+          highPriUpdateInteraction,
```
sophiebits: The way this test changed is a little interesting. Our "hi-pri update" causes us to re-call AsyncText which means that in a sense, the update now *is* blocked on the suspended component before it fully resolves. I can see the argument either way for how to account for this. The good news is that if we only scheduled an update on the sibling then they would stay independent.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -766,7 +761,7 @@ function commitRoot(root: FiberRoot, finishedWork: Fiber): void {
         unhandledError = error;
       }
     } finally {
-      if (!nextRenderIncludesTimedOutPlaceholder) {
+      if (true) {
```
acdlite: oops

sophiebits: ok whatever, ANdrew

Issue #12989, opened 2018-06-07
Consider removing Mobile Safari empty onclick hack


gaearon:
See https://github.com/facebook/react/issues/238 and https://github.com/facebook/react/pull/1536 for historical context.

Is that still relevant? The code is here:

https://github.com/facebook/react/blob/52fbe7612e0527b8c86decac519c344626f6bd72/packages/react-dom/src/client/ReactDOMFiberComponent.js#L244-L245

Even if it's relevant, can we just feature test it, and not do this hack on other browsers? Seems like a waste of memory for event handlers (even though the function is the same every time).

VideoCasino:
Needs investigation

nhunzaker:
I'd like to limit this to Safari. Looks like we still need this for non-interactive elements. 

I did a few quick checks using:
https://codepen.io/nhunzaker/pen/yEMpJO?editors=1010

- This definitely is still an issue on iOS 11
- It looks like this isn't a problem in desktop Safari (7.1, 8, 11.1)

I wonder what is the best way to detect Safari (or generally when to apply this fix)

gaearon:
Can we literally feature-test it? Like dispatching an event and seeing if it bubbles, or something?

gaearon:
The problem with relying on useragent etc is that those can get messed up in webviews.

nhunzaker:
Ah totally. I agree that we should feature detect this. I can check later today, but I wonder if it could be something like:

```
let noBubble = true

// where should document come from?

let parent = document.createElement('div')
let child = document.createElement('div')
let event = document.createEvent('Event')

event.initEvent('click', true, false)

parent.addEventListener('click', () => {
  noBubble = false
})

child.dispatchEvent(event);
```

philipp-spiess:
The feature test using manual event dispatching does not seem to work ([here's what I used to test it](http://jsfiddle.net/96uy7rkh/14/)). I was thinking that this bail out maybe happens in Safari's hit-detection logic and we could feature detect using [`elementFromPoint()`](http://jsfiddle.net/9box1t8y/9/) but that also behaved correctly. 🤔

hectorcoronado:
Hello! Would I be correct in surmising this may plausibly be a good first contribution? Haven't really dug into it sufficiently, but it looks doable. If so, and if one of you isn't already working on it, any chance you could assign to me?

gaearon:
@hectorcoronado Feel free to dig in. This will need some research — it's great if you can do it!

[...]

PR #11927, opened 2017-12-27
Trap click events for portal root

aweary:
Should resolve #11918 

Still need to run the build in iOS Safari to verify, but I'm 90% sure it will since an empty click handler fixed it in the JSFiddle provided for the issue.

Issue #12717, opened 2018-04-29
onMouseDown causes splash on real ios browsers


huhaixiao:
**Do you want to request a *feature* or report a *bug*?**

*bug*

**What is the current behavior?**

unexpected **grey splash** caused by `onMouseDown` on real ios device's browsers, both safari and chrome.
[Please checkout this **grey splash** issue first](https://github.com/mui-org/material-ui/issues/11154)
I did some trial later, and find out that vanilla html `<button onmousedown="">btn</button>` doesn't cause **grey splash**, but with React's jsx `<button onMouseDown={()=>{}}></button>`, there is a noticeable **grey splash**

**What is the expected behavior?**

who wants a splash?

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

"react": "^16.3.2"

gaearon:
For future reference, the reproducing example is this:

```js
import React from "react";
import ReactDOM from "react-dom";

function Index() {
  return <button onMouseDown={() => {}}>Wtf</button>;
}

ReactDOM.render(<Index />, document.querySelector("#root"));
```

https://codesandbox.io/s/p3npm3o6j7

The splashes the author is referring to occur not when you click on the button, but when you click *on the body itself* several times in a quick succession.

gaearon:
The problem is not specific to React; this also reproduces it:

```js
document.body.addEventListener(
  "mousedown",
  function() {}
);
```

gaearon:
Here is what Polymer does:

https://github.com/Polymer/polymer/blob/59405b57b3f7f3732b7f1db07ef1a76b1871814e/lib/utils/gestures.js#L92-L93

https://github.com/Polymer/polymer/blob/59405b57b3f7f3732b7f1db07ef1a76b1871814e/lib/utils/gestures.js#L210

https://github.com/Polymer/polymer/blob/59405b57b3f7f3732b7f1db07ef1a76b1871814e/lib/utils/gestures.js#L545-L547

gaearon:
Hmm.. there was clearly some issue with how I was testing this. It appears that CodeSandbox preview mode is buggy. I'll need to look at it again — for now previous conclusions might be wrong.

gaearon:
I can't seem to reproduce this in a vanilla (non-CRA) template: https://codesandbox.io/s/n50krjw5v4

Any chance this is caused by some DEV-only code that CRA pulls in?

gaearon:
Apparently removing 

```html
[...]

PR #13778, opened 2018-10-04
Don’t add onclick listener to React root

### LINKED ISSUE #13777: The gray overlay when tap the react root container
*bug*

Since React@16.5 seems that `root` element has `onclick` event handler with noop function.
In ios Safari/Chrome browsers html element has gray overlay when tap on it.

The current behavior is that the whole `root` is blinking when we tap on it.

The simple example to reproduce:
https://codesandbox.io/s/wyo8rlj1zk
The issue is reproducing only in fullscreen mode https://wyo8rlj1zk.codesandbox.io/
(root container is steel blue)

I suppose changes came from here:
https://github.com/facebook/react/pull/11927
https://github.com/facebook/react/pull/11927/files

and the issue can be related with:
https://github.com/mui-org/material-ui/issues/11154
https://github.com/facebook/react/issues/12717

React 16.5
ios 11.3/12 Safari/Chrome

philipp-spiess:
Fixes #13777

As part of #11927 we introduced a regression by adding onclick handler to the React root. This causes the whole React tree to flash when tapped on iOS devices (for reasons I outlined in https://github.com/facebook/react/issues/12989#issuecomment-414266839).

To fix this, we should only apply onclick listeners to portal roots. I verified that my proposed fix indeed works by checking out our DOM fixtures and adding regression tests.

Strangely, I had to make changes to the DOM fixtures to see the behavior in the first place. This seems to be caused by our normal sites (and 
thus their React root) being bigger than the viewport:

An alternative approach to finding out if we're appending to a React root would be to add a third parameter to `appendChildToContainer` based on the tag of the parent fiber.


gaearon:
>This seems to be caused by our normal sites (and thus their React root) being bigger than the viewport:

So the issue only appears when React root is *bigger*? Or when it is smaller?

I didn't see this in testing. But I don't remember what I used for testing.

philipp-spiess:
> So the issue only appears when React root is bigger? Or when it is smaller?

Only when the React root is _smaller_ than the viewport. In the gif you see the red flashing as long as the text does not cause the viewport to overflow.

This explains my strange observations here as well: https://github.com/facebook/react/issues/12717#issuecomment-410874871

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -341,8 +341,16 @@ export function appendChild(
   parentInstance.appendChild(child);
 }
 
+type DOMContainer =
```
philipp-spiess: I can also export this type from `ReactDOM` if that makes more sense.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -360,7 +368,7 @@ export function appendChildToContainer(
   // event exists. So we wouldn't see it and dispatch it.
   // This is why we ensure that containers have inline onclick defined.
   // https://github.com/facebook/react/issues/11918
-  if (parentNode.onclick === null) {
+  if (!container._reactRootContainer && parentNode.onclick === null) {
```
acdlite: Compare to null and undefined explicitly, plz

NE-SmallTown:
Should we modify the comments together?

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -360,7 +368,7 @@ export function appendChildToContainer(
   // event exists. So we wouldn't see it and dispatch it.
   // This is why we ensure that containers have inline onclick defined.
```
NE-SmallTown: Should we modify the comments together?

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -366,9 +366,14 @@ export function appendChildToContainer(
   // through the React tree. However, on Mobile Safari the click would
   // never bubble through the *DOM* tree unless an ancestor with onclick
   // event exists. So we wouldn't see it and dispatch it.
-  // This is why we ensure that containers have inline onclick defined.
+  // This is why we ensure that non React root containers have inline onclick
+  // defined.
   // https://github.com/facebook/react/issues/11918
-  if (!container._reactRootContainer && parentNode.onclick === null) {
+  const {_reactRootContainer} = container;
```
acdlite: Another style nit: we don't use destructuring, with some exceptions like module-style objects or function args (component props).

packages/react-dom/src/__tests__/ReactDOMComponent-test.js
```diff
@@ -2659,4 +2659,44 @@ describe('ReactDOMComponent', () => {
...
+
+      function Component() {
+        return ReactDOM.createPortal(
+          <div onClick={() => {}} />,
+          portalContainer,
+        );
+      }
+
+      ReactDOM.render(<Component />, container);
+      expect(typeof portalContainer.onclick === 'function').toBeTruthy();
```
gaearon: You can just do `expect(typeof ...).toBe('function')`. Not that it matters.

packages/react-dom/src/client/ReactDOMHostConfig.js
gaearon: Would make more sense than duplicating to me.

gaearon:
(see nit in https://github.com/facebook/react/pull/13778#discussion_r223534374)

Issue #11799, opened 2017-12-07
Consider removing XML compatibility from SSR or hiding it behind an option


gaearon:
See https://github.com/facebook/react/pull/11708#issuecomment-349953542.
Not sure if it's important but seems suboptimal to send extra markup if most people don't need XML.

syranide:
Perhaps there's far too much overlap, but the logical solution otherwise should be to separate ReadDOM into ReactHTMLDOM and ReactXHTMLDOM/ReactXMLDOM right? Considering XHTML is considered dead at this point IIRC, perhaps it wouldn't be so bad as it's unlikely that someone would ever try to use both at the same time, so the overlap is irrelevant as long as it's shared.

gaearon:
This only concerns the SSR implementation, doesn’t it? In that case passing an option seems easier than creating another one from scratch. I don’t think we’d want a separate top level namespace since the use case is so uncommon.

antonbasic:
Probably still not very common to need XHTML with React but would just like to make a case for keeping it. The European Securities and Markets Authority requires issuers on EU regulated markets to send financial reports in XHTML. I'm working on projects that use react to support this and would be nice to be able to do that.

Issue #2278, opened 2014-09-30
Generate XML compatible markup


zpao:
Take the alternative approach discussed in #2161. I think we can take advantage of the self-closing information we already have in ReactDOM (and make use of in ReactDOMComponent).

cc @audreyt in case you're interested in taking a stab at this.

Acubed:
I made a pull request to fix the self-closing tags and boolean attributes. This still leaves some features to be implemented, including:

* XML Namespaces (required for stand-alone documents and valid Polyglot HTML)
* the "raw text" elements: `script` and `style` tags
* xml:lang attributes
* Comments containing the sequence `--`
* Extra whitespace added to some HTML elements (pre, textarea, etc)

The definitive reference on how to generate markup that parses the same in both of HTML5's parsing modes is [Polyglot Markup: A robust profile of the HTML5 vocabulary](http://www.w3.org/TR/html-polyglot/). In the long run, being able to generate non-HTML vocabularies would be nice: stand-alone SVG documents, DocBook, Atom, etc.

dcousens:
What was the status of this as of React 0.14?

jimfb:
My understanding is that this issue was fixed in https://github.com/facebook/react/pull/2842 and this issue can be closed.

@dcousens My understanding is that this works in 0.14; let us know if you run into any issues.

Acubed:
That patch mostly makes the page _not break_ when parsing HTML in XML mode. It's still not suitable for arbitrary XML, like SVG, Atom, DocBook, OpenDocument, or the huge variety of domain-specific vocabularies out there.

React is just describing a DOM document or fragment; and text/html, XML, EXI, and even a certain subset of HAML and Jade, are all serializations of a DOM document. Right now React is limited to describing the subset that can also be rendered in text/html, but there's no design limitation on supporting others.

All DOM elements have an xmlns, e.g. in my console right now, I get:

    > document.firstElementChild.namespaceURI
    < "http://www.w3.org/1999/xhtml"

Adding support for elements with arbitrary namespaces is all that you really need. e.g.:

    var svgns = 'http://www.w3.org/2000/svg';
    React.createElement(svgns, 'svg', {}, [])

If rendered in XML mode, this would output:

    <svg xmlns="http://www.w3.org/2000/svg"/>

If adding `xmlns=` to every element every time the namespace changes becomes too repetitive, there's standard ways to declare or calculate a namespace prefix.

kevinSuttle:
Any updates on this?

[...]

PR #13783, opened 2018-10-05
Omit optional close tags in ReactDOMServerRenderer

### LINKED ISSUE #13342: RFC: Omit end tags when possible in ReactDOMServerRenderer
# Background

There are a lot of instances defined in the HTML spec where end tags can be omitted while still maintaining valid markup.

Some examples:

> An li element’s end tag may be omitted if the li element is immediately followed by another li element or if there is no more content in the parent element.

> A p element’s end tag may be omitted if the p element is immediately followed by an address, article, aside, blockquote, details, div, dl, fieldset, figcaption, figure, footer, form, h1, h2, h3, h4, h5, h6, header, hr, main, nav, ol, p, pre, section, table, or ul element, or if there is no more content in the parent element and the parent element is an HTML element that is not an a, audio, del, ins, map, noscript, or video element, or an autonomous custom element.

You can see all of them over at https://www.w3.org/TR/html5/syntax.html#optional-tags

Utilizing these rules, I think it may be possible to reduce the amount of html needed to be initially downloaded by the user for serverside rendering.

Depending on what's being rendered, there could be a significant decrease in html size. Here's an example I pulled from the W3C of a table element utilizing these rules. The difference is 464 bytes compared to 629 – about a **26% decrease** in size.

```html
<table>
<caption>37547 TEE Electric Powered Rail Car Train Functions (Abbreviated)
<colgroup><col><col><col>
<thead>
<tr>
  <th>Function
  <th>Control Unit
  <th>Central Station
<tbody>
<tr>
  <td>Headlights
  <td>✔
  <td>✔
<tr>
  <td>Interior Lights
  <td>✔
  <td>✔
<tr>
  <td>Electric locomotive operating sounds
  <td>✔
  <td>✔
<tr>
  <td>Engineer’s cab lighting
  <td>
  <td>✔
<tr>
  <td>Station Announcements - Swiss
  <td>
  <td>✔
</table>
```

# Caveats

- Even though these rules are in the spec, they still feel like quirks to me. There probably are some browser-to-browser differences in how these are handled.
- Should this even be handled by React? I could see a babel plugin or something like that being a good fit for an optimization like this too.

# Possible roadmap

- Research what end tags would be stable to omit (https://www.w3.org/TR/html5/syntax.html#optional-tags)
- Roll out this optimization for a subset of tags
- Gauge real-world load time improvements and continue adding omission cases if everything is looking good

tvler:
Closes #13342

# Changes

## High level

SSR rendered html goes from this

```html
<div>
  <p>one</p>
  <p>two</p>
</div>
```

to this

```html
<div>
  <p>one
  <p>two
</div>
```

which complies to the Optional tags html spec https://www.w3.org/TR/html5/syntax.html#optional-tags

## ReactDOMServerRenderer

- Removes `footer` property from `Frame`
- Creates `createCloseTagMarkup`, which computes the footer value
- Creates `parentTagStack`, which tracks the current frame's html parent
- Creates `previousSiblingTag`, which tracks the current frame's previous html sibling
- Omits close tags for html elements that don't require them following the W3C Optional tags spec https://www.w3.org/TR/html5/syntax.html#optional-tags
- Moves the close tag rendering logic so that we have access to the next html sibling value before rendering (we need to know the next sibling value to implement the optional tags spec)

## canOmitCloseTag

- Function that implements the W3C Optional tags spec https://www.w3.org/TR/html5/syntax.html#optional-tags
- Returns either true or false if a given html element can omit its close tag, given the parent and next sibling values
- If an html element doesn't have a parent, that means it's at the root where the actual html parent and sibling values are unknown. In that case, always render close tags to be safe.

## REACT_TEXT_TYPE

- Symbol that denotes a text node

## isAutonomousCustomComponent

Returns true if an html element is an autonomous custom element https://html.spec.whatwg.org/multipage/custom-elements.html#autonomous-custom-element, false otherwise

# Review

- [ ] canOmitCloseTag correctly implements the Optional tags spec
- [ ] Tests cover all instances of optional close tags occurring

# Related issues about XML compatability

- #11799 Consider removing XML compatibility from SSR or hiding it behind an option
- #2278 Generate XML compatible markup


tvler:
Any ideas why my imports in tests are working locally but not on circleci? Should I downgrade to a certain version of node?

PR #13785, opened 2018-10-05
[scheduler] Eagerly schedule rAF at beginning of frame

acdlite:
Eagerly schedule the next animation callback at the beginning of the frame. If the scheduler queue is not empty at the end of the frame, it will continue flushing inside that callback. If the queue *is* empty, then it will exit immediately. Posting the callback at the start of the frame ensures it's fired within the earliest possible frame. If we waited until the end of the frame to post the callback, we risk the browser skipping a frame and not firing the callback until the frame after that.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -534,13 +534,13 @@ if (typeof window !== 'undefined' && window._schedMock) {
     }
   }
 
-  var scheduledCallback = null;
-  var isIdleScheduled = false;
+  var scheduledHostCallback = null;
+  var isMessageEventScheduled = false;
   var timeoutTime = -1;
 
   var isAnimationFrameScheduled = false;
 
-  var isPerformingIdleWork = false;
+  var isFlushingHostCallback = false;
 
   var frameDeadline = 0;
   // We start out assuming that we run at 30fps but then the heuristic tracking
@@ -564,7 +564,12 @@ if (typeof window !== 'undefined' && window._schedMock) {
       return;
     }
 
-    isIdleScheduled = false;
+    isMessageEventScheduled = false;
+
+    var prevScheduledCallback = scheduledHostCallback;
+    var prevTimeoutTime = timeoutTime;
+    scheduledHostCallback = null;
+    timeoutTime = -1;
 
     var currentTime = getCurrentTime();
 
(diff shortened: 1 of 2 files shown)


gaearon:
Should we add a more aggressive manual testing fixture?

sophiebits:
I don't understand what this changes.

acdlite:
It's to defend against the browser pushing our animation frame into the frame after the next one, by scheduling it as early as possible. Eagerly, even before we know whether we need another frame to do work in or not.

Can't think of a good fixture to simulate getting pushed into the next frame because it's so timing dependent.

acdlite:
I suppose

```js
requestAnimationFrame(() => {
  sleep(12); // or some number that will push us into next frame if scheduled too late
});
```

and then mounting a bunch of slow components and measuring the overall time might work. I'll try it.

packages/scheduler/src/__tests__/SchedulerDOM-test.js
```diff
@@ -50,17 +50,18 @@ describe('SchedulerDOM', () => {
...
   function advanceOneFrame(config: FrameTimeoutConfigType = {}) {
     runRAFCallbacks();
     runPostMessageCallbacks(config);
   }
 
   let frameSize = 33;
-  let startOfLatestFrame = Date.now();
-  let currentTime = Date.now();
+  let startOfLatestFrame = 0;
+  let currentTime = 0;
```
acdlite: This is to make debugging easier. These tests use overridden `Date.now()` so the start time is unimportant.

NE-SmallTown:
> I suppose
> 
> ```js
> function animate() {
>   sleep(12); // or some number that will push us into next frame if scheduled too late
>   requestAnimationFrame(animate);
> }
> requestAnimationFrame(animate);
> ```
> 
> and then mounting a bunch of slow components and measuring the overall time might work. I'll try it.

I'm a little confused here to use `sleep`, nested rAF will be scheduled in the next frame, this is not related to `sleep`(i.e. even there is no `sleep` the nested rAF will be scheduled in the next frame)

packages/scheduler/src/Scheduler.js
```diff
@@ -605,7 +605,22 @@ if (typeof window !== 'undefined' && window._schedMock) {
...
+      // will continue flushing inside that callback. If the queue *is* empty,
+      // then it will exit immediately. Posting the callback at the start of the
+      // frame ensures it's fired within the earliest possible frame. If we
+      // waited until the end of the frame to post the callback, we risk the
+      // browser skipping a frame and not firing the callback until the frame
+      // after that.
+      requestAnimationFrameWithTimeout(animationTick);
+    } else {
+      // No pending work. Exit.
+      isAnimationFrameScheduled = false;
```
sebmarkbage: We can now unify these checks so we can get rid of this flag and just use `scheduledCallback` as the flag.

acdlite: I thought this would work but now I don't think it does. There are three callbacks: the one scheduled with `requestAnimationFrame` (represented by `isAnimationFrameScheduled`), the `message` event handler (represented by `isIdleScheduled`), and the one that actually contains the work (`scheduledCallback`). Need all three concepts. For example, as added by this PR, sometimes an animation frame is scheduled even if there's no actual work scheduled. So I think all three are needed.

I'll rename these a bit so it's less confusing.

PR #13789, opened 2018-10-06
Minor simplify childrenAsString

Connormiha:
Removed check for array length, because `join('')` with emty array returns emtpy string anyway.


aweary:
It may be semantically correct, but calling `join` on an array is likely a more expensive operation, so it's probably best to just have an explicit case handling empty children so we can avoid the function call.

Thanks though!

Issue #12548, opened 2018-04-04
Upgrade to Babel 7


acdlite:
Babel 7 is still in beta, but presumably it will be released soon. We should get a head start on upgrading. Perhaps it's stable enough for us to upgrade even before the final release.

One benefit is we could use JSX fragment syntax in tests. Yay!

@reznord has volunteered to pick this up.

reznord:
babel-jest latest version still doesn't support babel v7. The tests might face an issue after I update

acdlite:
Did you try the instructions here on how to use Jest with Babel 7?

https://facebook.github.io/jest/docs/en/getting-started.html#using-babel

> Note: If you are using a babel version 7 then you need to install babel-jest with the following command:
>
> ```
> yarn add --dev babel-jest 'babel-core@^7.0.0-0' @babel/core regenerator-runtime
> ```

gaearon:
Babel Jest definitely supports Babel 7, we're using them together in CRA

TrySound:
Not sure about `7.0.0-0`, but there is a special version for such cases
```js
yarn add babel-core@bridge --dev
```

SimenB:
`bridge` is a tag for `7.0.0-0`, so it's the same

reznord:
Hi @acdlite @gaearon sorry for the delay on the update for this PR. Wasn't doing well recently. I will start working on this PR now. 

Once again, sorry for the delay

asiniy:
@reznord may I take this?

dantesolis:
Is anyone working on this still ? If not I would give it a try. Thanks :)

armujahid:
I also want to help. @dantesolis can we both work on it together? If no one else is working on it.

dantesolis:
@armujahid yes sure, we can work on this together. :)

armujahid:
@dantesolis I have started working on it. https://github.com/armujahid/react/tree/babel7 but its not complete yet.

dantesolis:
@armujahid, wow you move fast! I will start working on it today or tomorrow, how should we tackle it so we don’t work on the same thing at the same time?

armujahid:
@dantesolis I have added you as collaborator to my repository. You can see my changes in [babel7 branch](https://github.com/armujahid/react/tree/babel7)
I have done the following:

1. Use [babel-upgrade](https://github.com/babel/babel-upgrade) tool to upgrade package.jsons
2. Update references of babel packages in code.
[...]

PR #13793, opened 2018-10-08
Babel 7

mmarkelov:
I just a bit nervous about this PR. So it will be fine if someone can review it, and maybe help me with it. 
It should pass all checks. But it also maybe something wrong with it. 
#12548


packages/react/src/__tests__/createReactClassIntegration-test.js
```diff
@@ -453,7 +453,7 @@ describe('create-react-class-integration', () => {
     expect(() =>
       ReactDOM.render(<Foo foo="foo" />, document.createElement('div')),
     ).toWarnDev(
-      'Component: getDerivedStateFromProps() is defined as an instance method ' +
+      'Warning: Foo: getDerivedStateFromProps() is defined as an instance method ' +
```
acdlite: Why did these change?

packages/react-dom/src/__tests__/ReactDOMFiber-test.js
```diff
@@ -180,9 +180,9 @@ describe('ReactDOMFiber', () => {
     const Div = () => <div />;
     const EmptyFragment = () => <React.Fragment />;
     const NonEmptyFragment = () => (
-      <React.Fragment>
+      <>
```
vikr01: Why were all `React.Fragment`'s removed?

babel.config.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+module.exports = {
+  presets: ['@babel/react', '@babel/preset-flow'],
+  ignore: ['third_party'],
+  plugins: [
+    '@babel/proposal-class-properties',
+    'syntax-trailing-function-commas',
```
vikr01: From @loganfsmyth:

>[This can be removed, since it has landed in the language official and is thus always enabled.](https://github.com/airbnb/babel-preset-airbnb/pull/37#discussion_r216164876)

scripts/print-warnings/print-warnings.js
```diff
@@ -6,20 +6,16 @@
...
 const fs = require('fs');
 const through = require('through2');
 const traverse = require('babel-traverse').default;
 const gs = require('glob-stream');
 
 const evalToString = require('../shared/evalToString');
 
-const babylonOptions = {
+const parserOptions = {
   sourceType: 'module',
```
vikr01: `trailingFunctionCommas` should be removed from here, see [babel's changelog](https://github.com/babel/babel/blob/eac4c5bc17133c2857f2c94c1a6a8643e3b547a7/packages/babel-parser/CHANGELOG.md#691-2016-08-23).

packages/react-dom/src/__tests__/ReactDOMComponent-test.js
```diff
@@ -2427,7 +2427,7 @@ describe('ReactDOMComponent', () => {
       const container = document.createElement('div');
 
       ReactDOM.render(<img src={obj} />, container);
-      expect(container.firstChild.src).toBe('hello');
+      expect(container.firstChild.src).toMatch('hello');
```
1pete: `container.firstChild.src` on master is `hello` but in this pr is `http://localhost/hello`
so you change test case just to make test passes.
you should change settings instead of changing this test case, shouldn't you?

package.json
```diff
@@ -5,63 +5,64 @@
     "packages/*"
   ],
   "devDependencies": {
+    "@babel/cli": "^7.0.0",
+    "@babel/code-frame": "^7.0.0",
+    "@babel/core": "^7.0.0",
+    "@babel/helper-module-imports": "^7.0.0",
+    "@babel/parser": "^7.1.2",
+    "@babel/plugin-proposal-class-properties": "^7.0.0",
```
milesj: You should just use preset-env instead of doing this all manually.

mmarkelov:
I will close it for a while to resolve problems, that you mentioned. Thank you, gays. I'll try to do my best!

scripts/print-warnings/print-warnings.js
```diff
@@ -6,25 +6,20 @@
...
-  // As a parser, babylon has its own options and we can't directly
-  // import/require a babel preset. It should be kept **the same** as
-  // the `babel-plugin-syntax-*` ones specified in
-  // https://github.com/facebook/fbjs/blob/master/babel-preset/configure.js
   plugins: [
     'classProperties',
     'flow',
     'jsx',
-    'trailingFunctionCommas',
     'objectRestSpread',
```
TrySound: This is also part of the language now. Try to remove

scripts/error-codes/extract-errors.js
```diff
@@ -6,24 +6,19 @@
...
-  // As a parser, babylon has its own options and we can't directly
-  // import/require a babel preset. It should be kept **the same** as
-  // the `babel-plugin-syntax-*` ones specified in
-  // https://github.com/facebook/fbjs/blob/master/babel-preset/configure.js
   plugins: [
     'classProperties',
     'flow',
     'jsx',
-    'trailingFunctionCommas',
     'objectRestSpread',
```
TrySound: And here too

scripts/error-codes/__tests__/replace-invariant-error-codes-test.js
```diff
@@ -10,8 +10,11 @@
 let babel = require('babel-core');
```
vikr01: This should be updated.

scripts/babel/__tests__/wrap-warning-with-env-check-test.js
```diff
@@ -10,8 +10,11 @@
 let babel = require('babel-core');
```
vikr01: This needs to be updated.

vikr01:
There are some custom local babel plugins defined in `./scripts/babel`. Those should be upgraded to babel 7 as well.

- [`declare` helper](https://github.com/babel/babel/blob/master/packages/babel-preset-flow/src/index.js#L4), see @loganfsmyth's [comment](https://github.com/airbnb/babel-preset-airbnb/pull/37#discussion_r221809334)
- [`api.assertVersion(7)`](https://babeljs.io/docs/en/v7-migration-api#asserting-babel-version-7450-https-githubcom-babel-babel-pull-7450)
- [Import certain things from `@babel/core`](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-modules-systemjs/src/index.js#L3) instead of [taking them from `api`](https://github.com/facebook/react/blob/0af81997095dfa3db67a0c5d287b144797742e99/scripts/babel/transform-prevent-infinite-loops.js#L21)

scripts/print-warnings/print-warnings.js
```diff
@@ -6,27 +6,17 @@
  */
 'use strict';
 
-const babylon = require('babylon');
+const parser = require('@babel/parser');
 const fs = require('fs');
 const through = require('through2');
 const traverse = require('babel-traverse').default;
```
vikr01: You missed this one.

loganfsmyth:
@vikr01 

> Import certain things from @babel/core instead of taking them from api

FWIW, I kind of regret that we ended up doing this in Babel's core plugins, so I don't know that I'd follow that example. I think in the end, reading them off `api` is a safer choice.

scripts/babel/transform-object-assign-require.js
```diff
@@ -7,12 +7,17 @@
...
-  function getAssignIdent(path, file, state) {
+const declare = require('@babel/helper-plugin-utils').declare;
+const addDefault = require('@babel/helper-module-imports').addDefault;
+
+module.exports = declare(api => {
+  api.assertVersion(7);
+
+  function getAssignIdent(path, state) {
     if (state.id) {
       return state.id;
```
NMinhNguyen: For what it's worth, I had to update my local copy of this transform as follows: https://gist.github.com/NMinhNguyen/4c3f8fdbd99730df6e00e5c33a23751c. This is against beta 42 so things might have changed since then, but I definitely encountered problems without doing `babel.types.cloneNode(state.id)`

lunaruan:
@mmarkelov Thanks so much for your work on this! I added you as a co-author on the [PR](https://github.com/facebook/react/commit/b12a982062ff3be7fdb95a104b800607f1d25976).

PR #13797, opened 2018-10-10
WIP: Don't update state if setState updater returns the state itself

Anber:
In the current implementation, a state is kept if `updater` returns `null` or `undefined` but it will be updated if `updater` returns the same state. It's inconvenient if a state is updated by functions like `set` or `setIn` from lodash or immutable because we need to compare a new state with a previous one, but nobody cares about it even in such popular libraries like `formik` which causes performance issues.
So, with this small change, it will be possible to use state updaters like `prevState => _.set(prevState, 'touched.field_name', true)` and prevent unnecessary state updates if nothing is changed.

### DIFF
--- a/packages/react-reconciler/src/ReactUpdateQueue.js
+++ b/packages/react-reconciler/src/ReactUpdateQueue.js
@@ -385,7 +385,11 @@ function getStateFromUpdate<State>(
         // Partial state object
         partialState = payload;
       }
-      if (partialState === null || partialState === undefined) {
+      if (
+        partialState === prevState ||
+        partialState === null ||
+        partialState === undefined
+      ) {
         // Null and undefined are treated as no-ops.
         return prevState;
       }


NE-SmallTown:
I agree with this but probably this will break some codes like `this.state.foo = 'bar'; this.setState(this.state)` (I don't agree to this usage and hate this but it is been used in the real world)

markerikson:
Yeah, that's a very interesting point.  I've previously noted that `setState()` doesn't actually care about mutation - you can mutate all you want and then call `setState()`, even to the point of doing:

```js
this.state.someValue = 123;
this.setState({});
```

and it will update "correctly" in the sense that it re-renders.

Anber:
Direct modification of state is wrong even according to the [documentation](https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly) but
@markerikson’s example will still work.
The example which @NE-SmallTown mentioned is more serious because some people can use this trick for force reload.
It will definitely be a breaking change for some wrong cases but on the other hand, it will help to keep state clear. Maybe it will be the good change for future major release?

PR #13822, opened 2018-10-11
React.pure automatically forwards ref

sophiebits:
We're not planning to encourage legacy context, and without this change, it's difficult to use pure+forwardRef together. We could special-case `pure(forwardRef(...))` but this is hopefully simpler.

```js
React.pure(function(props, ref) {
  // ...
});
```

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -236,6 +236,7 @@ function updatePureComponent(
   renderExpirationTime: ExpirationTime,
 ) {
   const render = Component.render;
+  const ref = workInProgress.ref;
 
   if (
     current !== null &&
@@ -246,7 +247,7 @@ function updatePureComponent(
     // Default to shallow comparison
     let compare = Component.compare;
     compare = compare !== null ? compare : shallowEqual;
-    if (compare(prevProps, nextProps)) {
+    if (workInProgress.ref === current.ref && compare(prevProps, nextProps)) {
       return bailoutOnAlreadyFinishedWork(
         current,
         workInProgress,
@@ -261,10 +262,10 @@ function updatePureComponent(
   if (__DEV__) {
     ReactCurrentOwner.current = workInProgress;
     ReactCurrentFiber.setCurrentPhase('render');
-    nextChildren = render(nextProps);
+    nextChildren = render(nextProps, ref);
     ReactCurrentFiber.setCurrentPhase(null);
   } else {
-    nextChildren = render(nextProps);
+    nextChildren = render(nextProps, ref);
   }
 
   // React DevTools reads this flag.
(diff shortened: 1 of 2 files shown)


gaearon:
Neat. I like this.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -246,7 +247,7 @@ function updatePureComponent(
     // Default to shallow comparison
     let compare = Component.compare;
     compare = compare !== null ? compare : shallowEqual;
-    if (compare(prevProps, nextProps)) {
+    if (workInProgress.ref === current.ref && compare(prevProps, nextProps)) {
```
acdlite: Thanks! Need to fix this for `forwardRef`, too, but we can do that in a separate PR.

sophiebits: Looks like we do already:

https://github.com/facebook/react/blob/0af81997095dfa3db67a0c5d287b144797742e99/packages/react-reconciler/src/ReactFiberBeginWork.js#L187-L208

though I'm not sure we need to check legacy context there?

packages/react-reconciler/src/__tests__/ReactPure-test.internal.js
```diff
@@ -35,14 +35,14 @@ describe('pure', () => {
...
   // a lazy function component.
   sharedTests('normal', (...args) => {
     const Pure = React.pure(...args);
-    function Indirection(props) {
-      return <Pure {...props} />;
+    function Indirection(props, ref) {
+      return <Pure {...props} ref={ref} />;
     }
-    return Promise.resolve(Indirection);
+    return Promise.resolve(React.forwardRef(Indirection));
```
sophiebits: Ironically had to figure out how to compose these for the sole purpose of testing this change.

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: Problem is we never hit this code path because of: https://github.com/facebook/react/blob/0af81997095dfa3db67a0c5d287b144797742e99/packages/react-reconciler/src/ReactFiberBeginWork.js#L1209-L1217

sebmarkbage:
I don't think we should take this lightly.

It's true that one proposal was to always pass the ref as the second argument for all function components.

However, another proposal was to instead not special case it and move it to props.

The forwardRefs mechanism has the benefit is that we can treat *it* as a legacy API that unwraps props. For example, extract it from props and instead pass it as a second arg. This would be a slow operation since it would loop over and clone (just like createElement does today).

This PR makes it so that we'd have to keep doing that, not just on the "legacy" non-idiomatic `forwardRef` API but also for the best-practice `pure` API too.

This PR is effectively saying that we've decided to always pass the second argument ref and keep special casing ref.

sophiebits:
Fair enough. I was under the impression that `ref` as the second arg being the best practice had been decided when I was out.

PR #13823, opened 2018-10-11
Hide timed-out children instead of deleting them so their state is preserved

acdlite:
If the children timeout, we switch to showing the fallback children in place of the "primary" children. However, we don't want to delete the primary children because then their state will be lost (both the React state and the host state, e.g. uncontrolled form inputs). Instead we keep them mounted and hide them. Both the fallback children AND the primary children are rendered at the same time. Once the primary children are un-suspended, we can delete the fallback children — don't need to preserve their state.

The two sets of children are siblings in the host environment, but semantically, for purposes of reconciliation, they are two separate sets. So we store them using two fragment fibers.

However, we want to avoid allocating two extra fibers for every Suspense fiber. They're only necessary when the children time out, because that's the only time when both sets are mounted.

So, the extra fragment fibers are only used if the children time out. Otherwise, we render the primary children directly. This requires some custom reconciliation logic to preserve the state of the primary children. It's essentially a very basic form of re-parenting.

- [x] Reconcile fallback and primary children separately.
- [x] Consolidate various pieces of Suspense state into a single type stored on `memoizedState`.
- [x] Hide timed-out children by toggling their top-most host nodes.
- [ ] React Native?

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
@@ -1,0 +1,120 @@
+/**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails react-core
+ */
+
+'use strict';
+
+let ReactFeatureFlags;
+let React;
+let ReactDOM;
+let Suspense;
+let ReactCache;
+let cache;
+let TextResource;
+
+describe('ReactDOMSuspensePlaceholder', () => {
+  let container;
+
+  beforeEach(() => {
+    jest.resetModules();
+    ReactFeatureFlags = require('shared/ReactFeatureFlags');
+    ReactFeatureFlags.enableSuspense = true;
+    React = require('react');
+    ReactDOM = require('react-dom');
+    ReactCache = require('react-cache');
+    Suspense = React.unstable_Suspense;
+    container = document.createElement('div');
+
+    function invalidateCache() {
+      cache = ReactCache.createCache(invalidateCache);
+    }
+    invalidateCache();
+    TextResource = ReactCache.createResource(([text, ms = 0]) => {
+      return new Promise((resolve, reject) =>
+        setTimeout(() => {
+          resolve(text);
+        }, ms),
+      );
+    }, ([text, ms]) => text);
+  });
+
+  function advanceTimers(ms) {
+    // Note: This advances Jest's virtual time but not React's. Use
+    // ReactNoop.expire for that.
+    if (typeof ms !== 'number') {
+      throw new Error('Must specify ms');
+    }
+    jest.advanceTimersByTime(ms);
+    // Wait until the end of the current tick
+    return new Promise(resolve => {
+      setImmediate(resolve);
+    });
+  }
+
+  function Text(props) {
+    return props.text;
+  }
+
+  function AsyncText(props) {
+    const text = props.text;
+    TextResource.read(cache, [props.text, props.ms]);
+    return text;
+  }
+
+  it('hides and unhides timed out DOM elements', async () => {
+    let divs = [
+      React.createRef(null),
+      React.createRef(null),
+      React.createRef(null),
+    ];
+    function App() {
+      return (
+        <Suspense maxDuration={500} fallback={<Text text="Loading..." />}>
+          <div ref={divs[0]}>
+            <Text text="A" />
+          </div>
+          <div ref={divs[1]}>
+            <AsyncText ms={1000} text="B" />
+          </div>
+          <div style={{display: 'block'}} ref={divs[2]}>
+            <Text text="C" />
+          </div>
+        </Suspense>
+      );
+    }
+    ReactDOM.render(<App />, container);
+    expect(divs[0].current.style.display).toEqual('none');
+    expect(divs[1].current.style.display).toEqual('none');
+    expect(divs[2].current.style.display).toEqual('none');
+
+    await advanceTimers(1000);
+
+    expect(divs[0].current.style.display).toEqual('');
+    expect(divs[1].current.style.display).toEqual('');
+    // This div's display was set with a prop.
+    expect(divs[2].current.style.display).toEqual('block');
+  });
+
+  it('hides and unhides timed out text nodes', async () => {
+    function App() {
+      return (
+        <Suspense maxDuration={500} fallback={<Text text="Loading..." />}>
+          <Text text="A" />
+          <AsyncText ms={1000} text="B" />
+          <Text text="C" />
+        </Suspense>
+      );
+    }
+    ReactDOM.render(<App />, container);
+    expect(container.textContent).toEqual('Loading...');
+
+    await advanceTimers(1000);
+
+    expect(container.textContent).toEqual('ABC');
+  });
+});
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -411,6 +411,22 @@ export function removeChildFromContainer(
   }
 }
 
+export function hideInstance(instance: Instance): void {
+  throw new Error('TODO');
+}
+
+export function hideTextInstance(textInstance: TextInstance): void {
+  throw new Error('TODO');
+}
+
+export function unhideInstance(instance: Instance, props: Props): void {
+  throw new Error('TODO');
+}
+
+export function unhideTextInstance(textInstance: TextInstance): void {
+  throw new Error('TODO');
+}
+
 // -------------------
 //     Hydration
 // -------------------
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -420,9 +420,7 @@ export function hideInstance(instance: Instance): void {
   // TODO: Does this work for all element types? What about MathML? Should we
   // pass host context to this method?
   instance = ((instance: any): HTMLElement);
-  if (instance.style !== undefined && instance.style !== null) {
-    instance.style.display = 'none';
-  }
+  instance.style.display = 'none';
 }
 
 export function hideTextInstance(textInstance: TextInstance): void {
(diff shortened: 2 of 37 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -352,19 +351,19 @@ function commitLifeCycles(
...
         finishedWork.updateQueue = emptyObject;
         scheduleWork(finishedWork, Sync);
       } else {
         // In strict mode, the Update effect is used to record the time at
-        // which the placeholder timed out.
-        const currentTime = requestCurrentTime();
-        finishedWork.stateNode = {timedOutAt: currentTime};
+        // which the children timed out.
+        // $FlowFixMe - Intentionally using a value other than an UpdateQueue.
+        finishedWork.updateQueue = {timedOutAt: requestCurrentTime()};
```
sebmarkbage: If we use this field for something other than the UpdateQueue, we must never use this field in any generic branches. I.e. we can't have any code that checks `.updateQueue` for a bunch of different or all tag types.

What's your motivation though? This doesn't make sense to me. It is a signal that stores a stateful value over time and never is allowed to be reset when fully processed. Sounds like state to me. Either stateNode or memoizedState.

acdlite:
Ok this is ready for review. I don't know how to implement the hide/unhide methods for React Native so I've left those empty for now.

trueadm:
So this approach uses `display: none` and `display: block` I see. Could we not just detach the DOM node from the document?

acdlite:
> So this approach uses display: none and display: block

When unhiding, it will either apply the user-provided `display` style prop, if it exists, or unset the `display` property using `el.style.display = null`:

https://github.com/facebook/react/blob/fe89105e92fd09287b4458cbe5e466da0e7758f1/packages/react-dom/src/client/ReactDOMHostConfig.js#L432-L445

We considered detaching the DOM nodes but you lose more state with that approach. Like uncontrolled form inputs.

trueadm:
@acdlite I'm curious to what state we lose. I thought form elements (inputs, selects etc) keep their state when in a detached state.

acdlite:
@trueadm Oh maybe I'm wrong about that one. We lose iframe state, and I believe we lose both form inputs and video playback. I'll check again.

trueadm:
@acdlite I expect iframes to lose state, there was an earlier proposal to allow them to not do so but I think it was scrapped. Video elements shouldn't lose state anymore, they should pause if I remember from memory (apart from on iOS, which used to cause them to reload even if you apply `display: none` on them, you had to use `visibility` and change their position to `absolute` and off-screen). I guess no matter what approach, we'll have edge cases. This is probably one the community has researched in depth already and maybe we could reach out to see if anyone has some recent findings with examples.

acdlite:
@trueadm Sounds good. I'll defer to you and @sebmarkbage's judgment.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -371,6 +521,7 @@ function completeWork(
       break;
     }
     case HostRoot: {
+      updateHostContainer(workInProgress);
```
gaearon: Is host environment throws, does it mean we don't clean up the stack due to `pop`s being moved after?

gaearon:
Do we know that all code paths in new traversal loops are covered by test? Last time we added traversal (context) it took a few releases to flush the bugs out. Maybe there's some fuzz that can help?

packages/react-reconciler/src/ReactFiberCompleteWork.js
acdlite: I'll revert this, mistake when resolving conflicts

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -431,6 +428,32 @@ function unwindWork(
       const effectTag = workInProgress.effectTag;
       if (effectTag & ShouldCapture) {
         workInProgress.effectTag = (effectTag & ~ShouldCapture) | DidCapture;
```
sebmarkbage: What's the difference between DidCapture effect and alreadyCaptured?

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -84,22 +88,94 @@ function markRef(workInProgress: Fiber) {
   workInProgress.effectTag |= Ref;
 }
 
-function appendAllChildren(parent: Instance, workInProgress: Fiber) {
+function appendAllChildren(
+  parent: Instance,
+  workInProgress: Fiber,
+  needsVisibilityToggle: boolean,
+  isHidden: boolean,
```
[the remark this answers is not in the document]
acdlite: I did it this way to avoid additional code size, but I guess this is a really hot path

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -412,19 +417,38 @@ export function removeChildFromContainer(
 }
 
 export function hideInstance(instance: Instance): void {
-  throw new Error('TODO');
+  // TODO: Does this work for all element types? What about MathML? Should we
+  // pass host context to this method?
+  instance = ((instance: any): HTMLElement);
+  if (instance.style !== undefined && instance.style !== null) {
```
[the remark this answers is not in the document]
acdlite: My mistake, forgot it's always non-null. I no know DOM.

packages/react-reconciler/src/ReactFiberCompleteWork.js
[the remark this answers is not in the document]
gaearon: Code size between modes shouldn't be a problem. One function will be DCE'd in persistent, another in mutative.

packages/react-reconciler/src/ReactFiberUnwindWork.js
acdlite: `DidCapture` effect isn't being used for SuspenseComponent anymore, just forgot to remove it. (It's still used by error boundaries).

The motivation for storing this on the state object is to unify the concurrent and non-concurrent modes as much as possible. The begin phase no longer distinguishes between them at all, unlike before where a non-null `updateQueue` was a hack to indicate non-current mode.

packages/react-reconciler/src/ReactFiberCompleteWork.js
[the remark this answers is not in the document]
acdlite: It’s not a static feature flag like persistent vs mutation. Happens at runtime.

sebmarkbage:
It's interesting that this is a feature that adds significant complexity to support sync suspense mode.

acdlite:
I would argue it's still useful even outside of sync mode, though I agree it's an edge case regardless

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1604,13 +1604,19 @@ function retrySuspendedRoot(
...
 
   const rootExpirationTime = root.expirationTime;
   if (rootExpirationTime !== NoWork) {
-    requestWork(root, rootExpirationTime);
+    try {
+      requestWork(root, rootExpirationTime);
+    } catch (e) {
+      console.log(e);
+      throw e;
+    }
```
acdlite: Oops, accidentally committed this. Will revert.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -714,6 +714,20 @@ function updateIndeterminateComponent(
...
+    // concurrent tree, in an inconsistent state. We want to tree it like
+    // a new mount, even though an empty version of it already committed.
+    // Disconnect the alternate pointers.
+    current.alternate = null;
+    workInProgress.alternate = null;
+    // Don't assign the `current` variable to null, though, since reconcileChildren uses the
+    // existence of current as a heuristic for whether to schedule placement
+    // effects on the children. This is super weird but reconcileChildren is
+    // deeply nested and it doesn't seem worth it to refactor for this edge
+    // case. (There are unit tests to prevent regressions.)
```
acdlite: Better fix: schedule a placement effect on `workInProgress`

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -431,17 +429,15 @@ export function hideTextInstance(textInstance: TextInstance): void {
...
-    if (props[STYLE] !== undefined && props[STYLE] !== null) {
-      const styleProp = props[STYLE];
-      if (styleProp.hasOwnProperty('display')) {
-        display = styleProp.display;
-      }
-    }
-    // $FlowFixMe Setting a style property to null is the valid way to reset it.
-    instance.style.display = display;
-  }
+  const styleProp = props.style;
```
acdlite: Let's revert this commit, except for the redundant existence check for `instance.style`

packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
```diff
@@ -0,0 +1,120 @@
...
+          <Text text="C" />
+        </Suspense>
+      );
+    }
+    ReactDOM.render(<App />, container);
+    expect(container.textContent).toEqual('Loading...');
+
+    await advanceTimers(1000);
+
+    expect(container.textContent).toEqual('ABC');
```
NE-SmallTown: Should we add `expect` for `0 < time < 500` and `500 < time < 1000` just like Suspense-test does?

NE-SmallTown: I'm a little confused here, if you could tell me I would be very happy!

Why `textContent` is `Loading...` after execute `ReactDOM.render` immediately? IMO, it should shows `Loading...` only after time >= 500 just like how Suspense works

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -702,17 +708,22 @@ function resolveDefaultProps(Component, baseProps) {
...
   renderExpirationTime,
 ) {
-  invariant(
-    current === null,
-    'An indeterminate component should never have mounted. This error is ' +
-      'likely caused by a bug in React. Please file an issue.',
-  );
+  if (_current !== null) {
+    // An indeterminate component only mounts if it suspended inside a non-
+    // concurrent tree, in an inconsistent state. We want to tree it like
```
aweary: > We want to tree it

🌲

packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
bvaughn: This test is not running in concurrent mode, so placeholders are shown immediately (sync).

NE-SmallTown: @bvaughn Thanks, I don't notice that. But if so, seems we don't need to set the `maxDuration` prop?

PR #13829, opened 2018-10-11
Deprecate context object as a consumer and add a warning message

trueadm:
This PR deprecates usage of React context as a consumer in DEV mode, along with a warning message informing the user of this change. Under the hood, we create a new object for `context.consumer` rather than having a cyclic reference to `context`. This object proxies values from the original context so it remains backwards compatible in React 16.

## dev when using consumer: 

- Before: When using `Context.Consumer`, it's the same as `Context` so nothing special is handled.

- After: When `Context.Consumer` is used the at the beginning of reconciliation, the fiber is given the `_context` field from the `Consumer`, that references to `Context`.

## dev when using context:

- Before: When using `Context`, it's the same as the previous case (but the wrong behaviour).

- After: When `Context` we check if `_context` field exists on it (which we added to the new proxy object) and because it won't (only Context.Consumer does) we trigger a new warning.

## prod when using consumer:

- Before & After: When using `Context.Consumer`, it's the same as `Context` so nothing special is handled.

## prod when using context:

- Before & After: When using `Context`, it's the same as the previous case so nothing special is handled.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -1107,7 +1107,7 @@ function updateContextConsumer(
   let context: ReactContext<any> = workInProgress.type;
   // The logic below for Context differs depending on PROD or DEV mode. In
   // DEV mode, we create a separate object for Context.Consumer that acts
-  // like a proxy to Context. This proxy object adds unecessary code in PROD
+  // like a proxy to Context. This proxy object adds unnecessary code in PROD
   // so we use the old behaviour (Context.Consumer references Context) to
   // reduce size and overhead. The separate object references context via
   // a property called "_context", which also gives us the ability to check
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -1105,6 +1105,13 @@ function updateContextConsumer(
   renderExpirationTime: ExpirationTime,
 ) {
   let context: ReactContext<any> = workInProgress.type;
+  // The logic below for Context differs depending on PROD or DEV mode. In
+  // DEV mode, we create a separate object for Context.Consumer that acts
+  // like a proxy to Context. This proxy object adds unecessary code in PROD
+  // so we use the old behaviour (Context.Consumer references Context) to
+  // reduce size and overhead. The separate object references context via
+  // a property called "_context", which also gives us the ability to check
+  // in DEV mode if this property exists or not and warn if it does not.
   if (__DEV__) {
     if (workInProgress.type._context === undefined) {
       if (!hasWarnedAboutUsingContextAsConsumer) {
@@ -1113,12 +1120,10 @@ function updateContextConsumer(
           false,
           'You are using the Context from React.createContext() as a consumer.' +
             'The correct way is to use Context.Consumer as the consumer instead. ' +
-            'This usage is deprecated and will be removed in the next major version.',
+            'This usage is deprecated and will be removed in a future major release.',
         );
       }
     } else {
-      // Context.Consumer is a separate object in DEV, where
-      // _context points back to the original context
       context = workInProgress.type._context;
     }
   }
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -8,7 +8,6 @@
  */
 
 import type {ReactProviderType, ReactContext} from 'shared/ReactTypes';
-import lowPriorityWarning from 'shared/lowPriorityWarning';
 import type {Fiber} from 'react-reconciler/src/ReactFiber';
 import type {FiberRoot} from './ReactFiberRoot';
 import type {ExpirationTime} from './ReactFiberExpirationTime';
@@ -1113,18 +1112,17 @@ function updateContextConsumer(
   // a property called "_context", which also gives us the ability to check
   // in DEV mode if this property exists or not and warn if it does not.
   if (__DEV__) {
-    if (workInProgress.type._context === undefined) {
+    if (context._context === undefined) {
       if (!hasWarnedAboutUsingContextAsConsumer) {
         hasWarnedAboutUsingContextAsConsumer = true;
-        lowPriorityWarning(
+        warning(
           false,
-          'You are using the Context from React.createContext() as a consumer.' +
-            'The correct way is to use Context.Consumer as the consumer instead. ' +
-            'This usage is deprecated and will be removed in a future major release.',
+          'Rendering <Context> directly is not supported and will be removed in ' +
+          'a future major release. Did you mean to render <Context.Consumer> instead?',
         );
       }
     } else {
-      context = workInProgress.type._context;
+      context = context._context;
     }
   }
   const newProps = workInProgress.pendingProps;
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -1118,7 +1118,7 @@ function updateContextConsumer(
         warning(
           false,
           'Rendering <Context> directly is not supported and will be removed in ' +
-          'a future major release. Did you mean to render <Context.Consumer> instead?',
+            'a future major release. Did you mean to render <Context.Consumer> instead?',
         );
       }
     } else {
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -8,6 +8,7 @@
  */
 
 import type {ReactProviderType, ReactContext} from 'shared/ReactTypes';
+import lowPriorityWarning from 'shared/lowPriorityWarning';
 import type {Fiber} from 'react-reconciler/src/ReactFiber';
 import type {FiberRoot} from './ReactFiberRoot';
 import type {ExpirationTime} from './ReactFiberExpirationTime';
@@ -1096,12 +1097,31 @@ function updateContextProvider(
   return workInProgress.child;
 }
 
+let hasWarnedAboutUsingContextAsConsumer = false;
+
 function updateContextConsumer(
   current: Fiber | null,
   workInProgress: Fiber,
   renderExpirationTime: ExpirationTime,
 ) {
-  const context: ReactContext<any> = workInProgress.type;
+  let context: ReactContext<any> = workInProgress.type;
+  if (__DEV__) {
+    if (workInProgress.type._context === undefined) {
+      if (!hasWarnedAboutUsingContextAsConsumer) {
+        hasWarnedAboutUsingContextAsConsumer = true;
+        lowPriorityWarning(
+          false,
+          'You are using the Context from React.createContext() as a consumer.' +
+            'The correct way is to use Context.Consumer as the consumer instead. ' +
+            'This usage is deprecated and will be removed in the next major version.',
+        );
+      }
+    } else {
+      // Context.Consumer is a separate object in DEV, where
+      // _context points back to the original context
+      context = workInProgress.type._context;
+    }
+  }
   const newProps = workInProgress.pendingProps;
   const render = newProps.children;
 
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -1112,7 +1112,7 @@ function updateContextConsumer(
   // a property called "_context", which also gives us the ability to check
   // in DEV mode if this property exists or not and warn if it does not.
   if (__DEV__) {
-    if (context._context === undefined) {
+    if ((context: any)._context === undefined) {
       if (!hasWarnedAboutUsingContextAsConsumer) {
         hasWarnedAboutUsingContextAsConsumer = true;
         warning(
@@ -1122,7 +1122,7 @@ function updateContextConsumer(
         );
       }
     } else {
-      context = context._context;
+      context = (context: any)._context;
     }
   }
   const newProps = workInProgress.pendingProps;
--- a/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
@@ -1578,11 +1578,9 @@ Context fuzz tester error! Copy and paste the following line into the test suite
     expect(() => {
       ReactNoop.render(<Component />);
       ReactNoop.flush();
-    }).toLowPriorityWarnDev(
-      'You are using the Context from React.createContext() as a consumer.' +
-        'The correct way is to use Context.Consumer as the consumer instead. ' +
-        'This usage is deprecated and will be removed in a future major release.',
-      {withoutStack: true},
+    }).toWarnDev(
+      'Rendering <Context> directly is not supported and will be removed in ' +
+      'a future major release. Did you mean to render <Context.Consumer> instead?',
     );
   });
 });
--- a/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
@@ -1190,12 +1190,12 @@ describe('ReactNewContext', () => {
 
       function FooAndBar() {
         return (
-          <FooContext>
+          <FooContext.Consumer>
             {foo => {
               const bar = BarContext.unstable_read();
               return <Text text={`Foo: ${foo}, Bar: ${bar}`} />;
             }}
-          </FooContext>
+          </FooContext.Consumer>
         );
       }
 
@@ -1557,5 +1557,32 @@ Context fuzz tester error! Copy and paste the following line into the test suite
         }
       }
     });
+  });
+
+  it('should warn with an error message when using context as a consumer in DEV', () => {
+    const BarContext = React.createContext({value: 'bar-initial'});
+    const BarConsumer = BarContext;
+
+    function Component() {
+      return (
+        <React.Fragment>
+          <BarContext.Provider value={{value: 'bar-updated'}}>
+            <BarConsumer>
+              {({value}) => <div actual={value} expected="bar-updated" />}
+            </BarConsumer>
+          </BarContext.Provider>
+        </React.Fragment>
+      );
+    }
+
+    expect(() => {
+      ReactNoop.render(<Component />);
+      ReactNoop.flush();
+    }).toLowPriorityWarnDev(
+      'You are using the Context from React.createContext() as a consumer.' +
+        'The correct way is to use Context.Consumer as the consumer instead. ' +
+        'This usage is deprecated and will be removed in the next major version.',
+      {withoutStack: true},
+    );
   });
 });
--- a/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
@@ -1581,7 +1581,7 @@ Context fuzz tester error! Copy and paste the following line into the test suite
     }).toLowPriorityWarnDev(
       'You are using the Context from React.createContext() as a consumer.' +
         'The correct way is to use Context.Consumer as the consumer instead. ' +
-        'This usage is deprecated and will be removed in the next major version.',
+        'This usage is deprecated and will be removed in a future major release.',
       {withoutStack: true},
     );
   });
--- a/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
@@ -1656,7 +1656,7 @@ Context fuzz tester error! Copy and paste the following line into the test suite
       ReactNoop.flush();
     }).toWarnDev(
       'Rendering <Context.Consumer.Provider> is not supported and will be removed in ' +
-      'a future major release. Did you mean to render <Context.Provider> instead?',
+        'a future major release. Did you mean to render <Context.Provider> instead?',
     );
-  });  
+  });
 });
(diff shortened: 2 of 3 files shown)


packages/react/src/ReactContext.js
```diff
@@ -67,7 +67,46 @@ export function createContext<T>(
...
-  context.Consumer = context;
+
+  if (__DEV__) {
+    // A separate object, but proxies back to the original context object for
+    // backwards compatibility. It has a different $$typeof, so we can properly
+    // warn for the incorrect usage of Context as a Consumer.
+    context.Consumer = {
+      $$typeof: REACT_CONTEXT_TYPE,
+      _context: context,
+      get _calculateChangedBits() {
```
gaearon: I think getters don't work on www, can you rewrite this as `Object.defineProperty` calls?
Worth checking tho

trueadm: Really? They work in IE9 the last time I checked.

trueadm: I changed over to `Object.defineProperties`. It's a shame Flow has such bad support for them though :(

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1096,12 +1097,31 @@ function updateContextProvider(
...
+        lowPriorityWarning(
+          false,
+          'You are using the Context from React.createContext() as a consumer.' +
+            'The correct way is to use Context.Consumer as the consumer instead. ' +
+            'This usage is deprecated and will be removed in the next major version.',
+        );
+      }
+    } else {
+      // Context.Consumer is a separate object in DEV, where
+      // _context points back to the original context
```
gaearon: This comment confuses me. It talks about DEV but is in PROD code path?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1096,12 +1097,31 @@ function updateContextProvider(
...
+  let context: ReactContext<any> = workInProgress.type;
+  if (__DEV__) {
+    if (workInProgress.type._context === undefined) {
+      if (!hasWarnedAboutUsingContextAsConsumer) {
+        hasWarnedAboutUsingContextAsConsumer = true;
+        lowPriorityWarning(
+          false,
+          'You are using the Context from React.createContext() as a consumer.' +
+            'The correct way is to use Context.Consumer as the consumer instead. ' +
+            'This usage is deprecated and will be removed in the next major version.',
```
gaearon: "in a future major release"

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1096,12 +1097,31 @@ function updateContextProvider(
...
+        );
+      }
+    } else {
+      // Context.Consumer is a separate object in DEV, where
+      // _context points back to the original context
+      context = workInProgress.type._context;
```
gaearon: Can you explain why we read it differently in DEV and PROD? It would help to have a written explanation of DEV/PROD matrix for Consumer/Context object and how behavior changes for each.

packages/react/src/ReactContext.js
```diff
@@ -67,7 +67,61 @@ export function createContext<T>(
...
+    // A separate object, but proxies back to the original context object for
+    // backwards compatibility. It has a different $$typeof, so we can properly
+    // warn for the incorrect usage of Context as a Consumer.
+    const consumer = {
+      $$typeof: REACT_CONTEXT_TYPE,
+      _context: context,
+      Provider: context.Provider,
+    };
+    // $FlowFixMe: Flow complains about not setting a value, which is intentional here
+    Object.defineProperties(consumer, {
```
gaearon: Is it better to have consumer proxy to context, or the other way around? Why?

trueadm: I checked and the context gets written to the most in our tests.

packages/react-reconciler/src/ReactFiberBeginWork.js
trueadm: Can you reference me to another matrix? I updated the comment

packages/react-reconciler/src/ReactFiberBeginWork.js
trueadm: It's a DEV path.

packages/react-reconciler/src/ReactFiberBeginWork.js
gaearon: There's none. I just meant if you could write up of what's supposed to change.

* dev when using consumer: (before) and (after)
* dev when using context: (before) and (after)
* prod when using consumer: (before) and (after)
* prod when using context: (before) and (after)

That would be easier for me to review since I can focus on intent separately from code.

trueadm: I've put a big nice comment in there that explains things better now. I'll aadd the matrix to this PR.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1096,12 +1097,36 @@ function updateContextProvider(
...
 function updateContextConsumer(
   current: Fiber | null,
   workInProgress: Fiber,
   renderExpirationTime: ExpirationTime,
 ) {
-  const context: ReactContext<any> = workInProgress.type;
+  let context: ReactContext<any> = workInProgress.type;
+  // The logic below for Context differs depending on PROD or DEV mode. In
+  // DEV mode, we create a separate object for Context.Consumer that acts
+  // like a proxy to Context. This proxy object adds unecessary code in PROD
```
gaearon: unnecessary

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1096,12 +1097,36 @@ function updateContextProvider(
...
+  let context: ReactContext<any> = workInProgress.type;
+  // The logic below for Context differs depending on PROD or DEV mode. In
+  // DEV mode, we create a separate object for Context.Consumer that acts
+  // like a proxy to Context. This proxy object adds unnecessary code in PROD
+  // so we use the old behaviour (Context.Consumer references Context) to
+  // reduce size and overhead. The separate object references context via
+  // a property called "_context", which also gives us the ability to check
+  // in DEV mode if this property exists or not and warn if it does not.
+  if (__DEV__) {
+    if (workInProgress.type._context === undefined) {
```
gaearon: Can this just be `context._context`? To avoid extra reads (even in DEV).

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1096,12 +1097,36 @@ function updateContextProvider(
...
+        hasWarnedAboutUsingContextAsConsumer = true;
+        lowPriorityWarning(
+          false,
+          'You are using the Context from React.createContext() as a consumer.' +
+            'The correct way is to use Context.Consumer as the consumer instead. ' +
+            'This usage is deprecated and will be removed in a future major release.',
+        );
+      }
+    } else {
+      context = workInProgress.type._context;
```
gaearon: This can also be just `context._context`.

packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
```diff
@@ -1558,4 +1558,31 @@ Context fuzz tester error! Copy and paste the following line into the test suite
...
+        </React.Fragment>
+      );
+    }
+
+    expect(() => {
+      ReactNoop.render(<Component />);
+      ReactNoop.flush();
+    }).toLowPriorityWarnDev(
+      'You are using the Context from React.createContext() as a consumer.' +
+        'The correct way is to use Context.Consumer as the consumer instead. ' +
```
gaearon: This message might confuse people a little bit. Maybe make it more visual?

```
Rendering <Context> directly is not supported and will be removed in a future major release. Did you mean to render <Context.Consumer> instead?
```

packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
```diff
@@ -1558,4 +1558,31 @@ Context fuzz tester error! Copy and paste the following line into the test suite
...
+        'This usage is deprecated and will be removed in a future major release.',
+      {withoutStack: true},
```
gaearon: Why no stack? This is exactly the kind of warning where I think the stack would be highly valuable.

packages/react/src/ReactContext.js
```diff
@@ -67,7 +67,61 @@ export function createContext<T>(
...
+        get() {
+          return context._currentValue2;
+        },
+        set(_currentValue2) {
+          context._currentValue2 = _currentValue2;
+        },
+      },
+      Consumer: {
+        get() {
+          return context.Consumer;
```
gaearon: What if you render `<Context.Consumer.Consumer>`? Seems like that should also warn because it also relies on objects being shared.

Which makes me think: should the warning move into these getters instead? Fire for first getter accessed, ignore the rest. This could also nicely let us warn once per context type.

trueadm: Moving the warning into the getters will stop the error from coming up for just using `<Context />`. I'll address the nested `Consumer.Consumer` issue.

packages/react/src/ReactContext.js
```diff
@@ -67,7 +68,71 @@ export function createContext<T>(
...
+    // backwards compatibility. It has a different $$typeof, so we can properly
+    // warn for the incorrect usage of Context as a Consumer.
+    const consumer = {
+      $$typeof: REACT_CONTEXT_TYPE,
+      _context: context,
+      Provider: context.Provider,
+    };
+    // $FlowFixMe: Flow complains about not setting a value, which is intentional here
+    Object.defineProperties(consumer, {
+      _calculateChangedBits: {
```
gaearon: `_calculateChangedBits` and `unstable_read` never change, can we directly point to the original without getters for them?

packages/react/src/ReactContext.js
```diff
@@ -78,18 +78,12 @@ export function createContext<T>(
     const consumer = {
       $$typeof: REACT_CONTEXT_TYPE,
       _context: context,
+      _calculateChangedBits: context._calculateChangedBits,
       Provider: context.Provider,
+      unstable_read: context.unstable_read,
```
gaearon: It's assigned right after, no? This is undefined.

trueadm: Moved the line above, my bad.

packages/react/src/ReactContext.js
```diff
@@ -67,9 +68,60 @@ export function createContext<T>(
...
+  if (__DEV__) {
+    // A separate object, but proxies back to the original context object for
+    // backwards compatibility. It has a different $$typeof, so we can properly
+    // warn for the incorrect usage of Context as a Consumer.
+    const consumer = {
+      $$typeof: REACT_CONTEXT_TYPE,
+      _context: context,
+      _calculateChangedBits: context._calculateChangedBits,
+      Provider: context.Provider,
+      unstable_read: context.unstable_read,
```
gaearon: Now that I think of it `Context.Consumer.unstable_read()` should also warn.

packages/react/src/ReactContext.js
```diff
@@ -67,9 +68,71 @@ export function createContext<T>(
...
+
+  if (__DEV__) {
+    // A separate object, but proxies back to the original context object for
+    // backwards compatibility. It has a different $$typeof, so we can properly
+    // warn for the incorrect usage of Context as a Consumer.
+    const consumer = {
+      $$typeof: REACT_CONTEXT_TYPE,
+      _context: context,
+      _calculateChangedBits: context._calculateChangedBits,
+      Provider: context.Provider,
```
gaearon: Let's make `<Context.Consumer.Provider>` warn as well.

packages/react/src/ReactContext.js
```diff
@@ -67,9 +68,87 @@ export function createContext<T>(
...
 
+  let hasWarnedAboutUsingNestedContextConsumers = false;
+  let hasWarnedAboutUsingConsumerUnstableRead = false;
+  let hasWarnedAboutUsingConsumerProvider = false;
+
+  if (__DEV__) {
+    // A separate object, but proxies back to the original context object for
+    // backwards compatibility. It has a different $$typeof, so we can properly
+    // warn for the incorrect usage of Context as a Consumer.
+    const consumer = {
```
gaearon: Nit: I'd probably call this `Consumer`

gaearon:
For testing SSR, adding a warning and then cloning [this test](https://github.com/facebook/react/blob/8ca8a594e6fc7dfcba73b8595344e7522d40171d/packages/react-dom/src/__tests__/ReactDOMServerIntegrationNewContext-test.js#L195-L250) to use `<Context>` and expect warnings by passing `1` as second argument to `await render(...)` should be good.

trueadm:
I'll do that change in a follow up on Monday. :)

gaearon:
Cool. Don't forget a client-side test though. It's easy to change a field name, forget to update the proxy, and break it.

jaydenseric:
React v16.6 had some breaking changes to the way context works for SSR libraries. I'm trying to fix https://github.com/jaydenseric/graphql-react/issues/11 but finding it pretty hard to work out whats going on… any tips?

Here are some relevant locations:

- Where context consumers are rendered: https://github.com/jaydenseric/graphql-react/blob/v3.0.0/src/preload.mjs#L73
- Where context value is established: https://github.com/jaydenseric/graphql-react/blob/v3.0.0/src/preload.mjs#L135

jaydenseric:
I thought I'd worked it out, but discovered (after a release and deployment 😞) that context consumer behavior is pretty different when `NODE_ENV=production`:

```
{ '$$typeof': Symbol(react.element),
  type:
   { '$$typeof': Symbol(react.context),
     _context:
      { '$$typeof': Symbol(react.context),
        _calculateChangedBits: null,
        _currentValue: undefined,
        _currentValue2: undefined,
        Provider: [Object],
        Consumer: [Circular],
        _currentRenderer: null,
        _currentRenderer2: null,
        displayName: 'GraphQLContext',
        currentValue: [GraphQL] },
     _calculateChangedBits: null },
  key: null,
  ref: null,
  props: { children: [Function] },
  _owner: null,
  _store: {} }
```

vs

```
{ '$$typeof': Symbol(react.element),
  type:
   { '$$typeof': Symbol(react.context),
     _calculateChangedBits: null,
     _currentValue: undefined,
     _currentValue2: undefined,
     Provider: { '$$typeof': Symbol(react.provider), _context: [Circular] },
     Consumer: [Circular],
     displayName: 'GraphQLContext',
     currentValue:
      GraphQL {
        reset: [Function],
        request: [Function],
        query: [Function],
        cache: {},
        requests: {},
        on: [Function: on],
        off: [Function: off],
        emit: [Function: emit] } },
  key: null,
  ref: null,
  props: { children: [Function] },
  _owner: null }
```

https://github.com/jaydenseric/graphql-react/issues/12

It's midnight now, I give up.

packages/react/src/ReactContext.js
FezVrasta: Hi I'm getting the `Context.Consumer.Consumer` warning/error but I can't find where I am actually calling it in my code.

I just have 3 contexts and none of them are being used wrongly. May it be a bug or in the code that throws the warning or am I doing something extremely stupid?

I just wanted to let you know because I'm quite sure something is off here.

trueadm:
@jaydenseric Did you have any luck fixing your issue? I wonder if the follow up PR fixes the issues your experienced: https://github.com/facebook/react/pull/14033.

jaydenseric:
@trueadm I did eventually get it working, there were a few things that had to change; [here is a diff](https://github.com/jaydenseric/graphql-react/compare/v3.0.0...v4.0.1#diff-3e6d5558503509cddfcf7ef1bc7ee575).

Looking over it a now, a few days later it's hard to remember _exactly_ the issues…

1. You can’t do [`if (element.type.Consumer)`](https://github.com/jaydenseric/graphql-react/compare/v3.0.0...v4.0.1#diff-3e6d5558503509cddfcf7ef1bc7ee575L72) to check if an element is a context consumer anymore. You have to check the `element.type.$$typeof` Symbol. The bundle impact is way too big to import it the "proper" way from [`react-is`](https://npm.im/react-is), so [I defined it manually](https://github.com/jaydenseric/graphql-react/compare/v3.0.0...v4.0.1#diff-3e6d5558503509cddfcf7ef1bc7ee575R23). Also [babel/babel#7998](https://github.com/babel/babel/issues/7998) would cause a CJS runtime error.
2. You can’t do [`element.type.currentValue`](https://github.com/jaydenseric/graphql-react/compare/v3.0.0...v4.0.1#diff-3e6d5558503509cddfcf7ef1bc7ee575L75) to get the current value anymore. You have to do [`element.type._currentValue`](https://github.com/jaydenseric/graphql-react/compare/v3.0.0...v4.0.1#diff-3e6d5558503509cddfcf7ef1bc7ee575R103).
3. For some reason, context value could not be retrieved anymore for a context consumer that is a descendant of another context consumer. For a while I have been putting of fixing an edge case bug in `graphql-react` [that was recently fixed in `react-apollo`](https://github.com/apollographql/react-apollo/pull/2304), where you have to use maps to create scopes for context under providers. Actioning that fix also worked around the new limitation.
4. I got really stuck until figuring out that there are 2 ways you need to retrieve a context consumers' provider; [1 each for `NODE_ENV` undefined  (`element.type._context.Provider`) and `production` (`element.type.Provider`)](https://github.com/jaydenseric/graphql-react/compare/v3.0.0...v4.0.1#diff-3e6d5558503509cddfcf7ef1bc7ee575R104). I now run all tests 4 times; CJS dev/prod, ESM dev/prod.

I live in fear non semver major React releases will break this logic again; It's happened twice now. Do you think the changes coming in https://github.com/facebook/react/pull/14033 will cause this implementation to break again? It's kinda hard for me to tell looking at it.

`react-apollo` is looking at give up on efficiently walking the tree in favor of multiple renders to string (https://github.com/apollographql/react-apollo/pull/2533) until there is an elegant way to do things the Suspense way. Although that's one of the ways I first considered doing SSR in `graphql-react` and have recent promising experiments, it might be easier to skip that trouble and avoid React 16.7+ hooks, etc. if it breaks SSR and wait for Suspense. Everyone would like to know, but is it far off?

packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
laurosilvacom: This message is still not clear @trueadm . I'm already calling Context.Consumer as a consumer? But I'm still getting the error message. 

```jsx
import React from "react"

const defaultState = {
  dark: false,
  toggleDark: () => {},
}

const ThemeContext = Context.Consumer(defaultState)

// Getting dark mode information from OS!
// You need macOS Mojave + Safari Technology Preview Release 68 to test this currently.
const supportsDarkMode = () =>
  window.matchMedia("(prefers-color-scheme: dark)").matches === true

class ThemeProvider extends React.Component {
  state = {
    dark: false,
  }

  toggleDark = () => {
    let dark = !this.state.dark
    localStorage.setItem("dark", JSON.stringify(dark))
    this.setState({ dark })
  }

  componentDidMount() {
    // Getting dark mode value from localStorage!
    const lsDark = JSON.parse(localStorage.getItem("dark"))
    if (lsDark) {
      this.setState({ dark: lsDark })
    } else if (supportsDarkMode()) {
      this.setState({ dark: true })
    }
  }

  render() {
    const { children } = this.props
    const { dark } = this.state
    return (
      <ThemeContext.Provider
        value={{
          dark,
          toggleDark: this.toggleDark,
        }}
      >
        {children}
      </ThemeContext.Provider>
    )
  }
}

export default ThemeContext

export { ThemeProvider }
```

trueadm: @laurosilvacom you should be using ThemeContext = createContext(...) and then, doing ThemeContext.Consumer and ThemeContext.Provider

laurosilvacom: I'm not calling `const ThemeContext = React.createContext(defaultState)` but still getting the same error. Am I missing something about how I'm suppose to call `<Context.Consumer>` instead?

trueadm: I was suggesting you should call it and use it that way. I can help more when I’m back in the office in a few days.

laurosilvacom: Sounds good. I appreciate the help.

gaearon: @laurosilvacom 

I don't understand what you're trying to do.

You should create context like this:

```js
const ThemeContext = React.createContext(defaultState) // NOT "Context.Consumer(defaultState)"
```

and then subscribe to it either like this:

```js
class Button extends Component {
  static contextType = ThemeContext; // NOT ThemeContext.Consumer

  render() {
    // ...
  }
}
```

or like this:

```js
class Button extends Component {
  render() {
    return (
      // NOT <ThemeContext>
      <ThemeContext.Consumer>
        {theme => ...}
      </ThemeContext.Consumer>
    );
  }
}
```

Does that help?

laurosilvacom: Yes!

I did not realize I needed to subscribe to it by calling ` <ThemeContext.Consumer>`.  

I had:

 <ThemeContext>
        {theme => ...}
 </ThemeContext>

Thanks @gaearon!

gaearon: Right, which is why the warning says:

>Rendering `<Context>` directly is not supported and will be removed in a future major release. Did you mean to render `<Context.Consumer>` instead?

Do you still feel it was unclear? Which part could be improved? Thanks!

laurosilvacom: An error message that's direct would help. 

> Rendering <Context> directly is not supported and will be removed in a future major release. You want to use `Context.Consumer>`.

PR #13832, opened 2018-10-11
Add failing test for Suspense bug

gaearon:
I expect that `maxDuration` that is larger than both the actual waiting time *and* the current expiration context timeout should be equivalent to not specifying `maxDuration` at all. But that doesn't seem like what happens. 

A more concrete example:

```js
import React, { lazy, unstable_Suspense as Suspense } from "react";
import { unstable_createRoot as createRoot } from "react-dom";

let done = false;
let promise;
function Delay({ ms, children }) {
  if (done) {
    return children;
  }
  if (!promise) {
    promise = new Promise(resolve => {
      setTimeout(() => {
        done = true;
        resolve();
      }, ms);
    });
  }
  throw promise;
}

class Spinner extends React.Component {
  componentDidMount() {
    alert("spin");
  }
  render() {
    console.log("spinner");
    return <h1>🌀 'Loading....'</h1>;
  }
}

class App extends React.Component {
  state = {};
  render() {
    return (
      <>
        <h1 onClick={() => this.setState({ showStuff: true })}>Suspense</h1>
        {this.state.showStuff && (
          <Suspense maxDuration={2000} fallback={<Spinner />}>
            <Delay ms={10}>Ready</Delay>
          </Suspense>
        )}
      </>
    );
  }
}

createRoot(document.getElementById("root")).render(<App />);
```

I'd expect `Spinner` to never be shown — or at least the behavior to be consistent between `maxDuration={2000}` and missing `maxDuration`.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
@@ -241,4 +241,63 @@ describe('ReactSuspense', () => {
     ]);
     expect(root).toMatchRenderedOutput('AsyncSibling');
   });
+
+  // This is a regression test for a Suspense bug.
+  it('specifying larger maxDuration is equivalent to omitting it', () => {
+    [100, 500, 1000, 2000, 5000, 10000, undefined].forEach(maxDuration => {
+      const Suspense = React.unstable_Suspense;
+
+      class UnexpectedFallback extends React.Component {
+        componentDidMount() {
+          throw new Error('Did not expect this fallback to be shown.');
+        }
+        render() {
+          return <Text text="Loading..." />;
+        }
+      }
+
+      let inst;
+      class App extends React.Component {
+        state = {
+          show: false,
+        };
+        render() {
+          inst = this;
+          return (
+            this.state.show && (
+              // The bug was triggered by the presence of maxDuration.
+              <Suspense
+                maxDuration={maxDuration}
+                fallback={<UnexpectedFallback />}>
+                <AsyncText text="A" ms={1} />
+              </Suspense>
+            )
+          );
+        }
+      }
+
+      const root = ReactTestRenderer.create(<App />, {
+        unstable_isConcurrent: true,
+      });
+      expect(root).toFlushWithoutYielding();
+      expect(root).toMatchRenderedOutput(null);
+      ReactTestRenderer.unstable_interactiveUpdates(() => {
+        inst.setState({
+          show: true,
+        });
+      });
+      expect(root).toMatchRenderedOutput(null);
+
+      // Suspend (but it hasn't expired yet so we don't show a spinner)
+      jest.advanceTimersByTime(10);
+      expect(root).toFlushAndYield(['Suspend! [A]', 'Loading...']);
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
```diff
@@ -241,4 +241,63 @@ describe('ReactSuspense', () => {
     ]);
     expect(root).toMatchRenderedOutput('AsyncSibling');
   });
+
+  // This is a regression test for a Suspense bug.
+  it('specifying larger maxDuration is equivalent to omitting it', () => {
+    [100, 500, 1000, 2000, 5000, 10000, undefined].forEach(maxDuration => {
```
gaearon: I don't know if all of these are a bug. At the very least `[2000, undefined]` seems to demonstrate a bug to me. Others might fail for another (maybe legit?) reason although I don't see it immediately.

acdlite:
This is a known behavior for updates at user-blocking (nee interactive) priority, the consequence of an intentional performance/modeling trade-off.

They time out effectively immediately, because our heuristic for computing the elapsed time is to subtract the normal expiration time offset from the render expiration time. Because the user-blocking expiration offset is ~1, and normal expiration time offset is ~5, the computed elapsed time of a user-blocking update is always ~4 seconds higher than the real one. Code is here:

https://github.com/facebook/react/blob/a68ca9a5b54394bb61131f73f38329bfaeae331c/packages/react-reconciler/src/ReactFiberUnwindWork.js#L297-L313

The reason we made this trade-off was so we could avoid storing the start time of every pending update. It seemed reasonable because if you want to suspend, you probably should be deferring the update to normal priority anyway.

We might want to consider revisiting that trade-off. I think it's probably fine for now. There are other compelling reasons to go back to tracking all the pending levels, so if we do, we should definitely pick this task back up at that point.

acdlite:
Follow-up: I think we can fix this without needing to track every expiration time, once we start tracking the current priority level with Scheduler, we can use that to subtract the correct number. The Scheduler priority integration (which we were already planning to do for other reasons) replaces the need to maintain our own list of tasks per root — Scheduler is already doing that.

acdlite:
Oh right, past Andrew also left this comment:

> This will cause  high priority (interactive) work to expire earlier than necessary,  but we can account for this by adjusting for the Just Noticeable Difference. 

IOW, if we're in Concurrent Mode, and we get a timeout of 0ms, we can round that up to the nearest noticable threshold. Say, 50ms. That should avoid the pathological case where an already resolved promise pings in a microtask.

That'll be an easier fix for now, since Scheduler integration is a larger effort.

koba04:
I've also struggled this.
I've got the reason but I think it seems that the behavior is confusing for many developers.
Because when they specify `maxDuration`, they would expect that `maxDuration` will be treated as the timeout.

I think it would be nice to print a warning in development mode if the timeout has come earlier than `maxDuration`.

gaearon:
@koba04 I don't think what you're describing is necessarily related to this PR btw. It _is_ going to be necessary to address this in the docs though.

gaearon:
I don't know if this is still relevant. `maxDuration` is dead but maybe this heuristic can still bite us in some way?

PR #13833, opened 2018-10-12
Fix User Timing oddities with Suspense, pure, and lazy

gaearon:
This is for User Timings perf interation. We've been neglecting testing it when adding new features — let's try to keep it working in the future.

* `pure` components now show their name instead of `Unknown`
* There was a naming mismatch because `lazy` components began work as `Unknown` but completed with real name. I fix this by canceling immediately and restarting right after they get the right tag. Seems a bit odd but probably okay for this corner case?
* Suspense printed error boundary message which seemed misleading.

### DIFF
--- a/packages/react-reconciler/src/ReactDebugFiberPerf.js
+++ b/packages/react-reconciler/src/ReactDebugFiberPerf.js
@@ -20,6 +20,7 @@ import {
   ContextProvider,
   ContextConsumer,
   Mode,
+  SuspenseComponent,
 } from 'shared/ReactWorkTags';
 
 type MeasurementPhase =
@@ -315,7 +316,10 @@ export function stopFailedWorkTimer(fiber: Fiber): void {
       return;
     }
     fiber._debugIsCurrentlyTiming = false;
-    const warning = 'An error was thrown inside this error boundary';
+    const warning =
+      fiber.tag === SuspenseComponent
+        ? 'Rendering has been suspended'
+        : 'An error was thrown inside this error boundary';
     endFiberMark(fiber, null, warning);
   }
 }
--- a/packages/react-reconciler/src/ReactDebugFiberPerf.js
+++ b/packages/react-reconciler/src/ReactDebugFiberPerf.js
@@ -318,7 +318,7 @@ export function stopFailedWorkTimer(fiber: Fiber): void {
     fiber._debugIsCurrentlyTiming = false;
     const warning =
       fiber.tag === SuspenseComponent
-        ? 'Rendering has been suspended'
+        ? 'Rendering was suspended'
         : 'An error was thrown inside this error boundary';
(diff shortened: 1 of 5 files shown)


packages/shared/getComponentName.js
```diff
@@ -63,13 +64,23 @@ function getComponentName(type: mixed): string | null {
         return 'Context.Consumer';
       case REACT_PROVIDER_TYPE:
         return 'Context.Provider';
-      case REACT_FORWARD_REF_TYPE:
+      case REACT_FORWARD_REF_TYPE: {
         const renderFn = (type.render: any);
         const functionName = renderFn.displayName || renderFn.name || '';
         return (
           (type: any).displayName ||
           (functionName !== '' ? `ForwardRef(${functionName})` : 'ForwardRef')
```
sophiebits: should we be consistent?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -719,11 +719,15 @@ function mountIndeterminateComponent(
     Component !== null &&
     typeof Component.then === 'function'
   ) {
+    // We can't start a User Timing measurement with correct label yet.
+    // Cancel and resume right after we know the tag.
+    cancelWorkTimer(workInProgress);
```
sophiebits: does this make two segments in the flame chart?

packages/react-reconciler/src/ReactDebugFiberPerf.js
```diff
@@ -315,7 +316,10 @@ export function stopFailedWorkTimer(fiber: Fiber): void {
       return;
     }
     fiber._debugIsCurrentlyTiming = false;
-    const warning = 'An error was thrown inside this error boundary';
+    const warning =
+      fiber.tag === SuspenseComponent
+        ? 'Rendering has been suspended'
```
sophiebits: Rendering was suspended?

packages/shared/getComponentName.js
gaearon: Maybe.. I figured this one would be common and might be annoying to see it everywhere. Or maybe not?

packages/react-reconciler/src/ReactFiberBeginWork.js
gaearon: No, it just discards the in-progress segment.

packages/react-reconciler/src/ReactDebugFiberPerf.js
gaearon: sure

packages/shared/getComponentName.js
gaearon: I guess it could be helpful to see that a pure component re-rendered

gaearon:
Addressed review

Issue #13029, opened 2018-06-12
Share ref with multiple ref handlers


joaovieira:
Before React 16.3 we were able to proxy the same element ref to multiple listeners, for example, to grab hold of the element for internal purposes and also expose it publicly, like so:

```jsx
class MyComponent extends Component {
  attachRef = el => {
    this.buttonEl = el;
    this.props.buttonRef(el);
  }

  // Do something with `this.buttonEl`

  render () {
    return <button ref={this.attachRef}>Button</button>;
  }
}
```

After React 16.3 this is more complicated as the ref prop can be a function or an object:

```jsx
class MyComponent extends Component {
  buttonRef = React.createRef();

  attachRef = el => {
    this.buttonRef.current = el;
    
    if (typeof this.props.inputRef === 'function') {
      this.props.inputRef(el);
    } else {
      this.props.inputRef.current = el;
    }
  }

  // Do something with `this.buttonRef.current`

  render () {
    return <button ref={this.attachRef}>Button</button>;
  }
}
```

**First, is this the right approach?** 

And if so, [Typescript types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L53) say:

```ts
interface RefObject<T> {
  readonly current: T | null;
}
```

Which prevents us from assigning to `current`. **Shouldn't it not be readonly?**

gaearon:
The approach you're suggesting should work. Yeah, it's more complicated. But we also don't encourage often doing this as it breaks encapsulation.

The property is usually read-only in the way it's currently used, but in your case it's fair to want to write to it. It's possible that we'll ask to change that type to be writeable in the future. (We don't maintain the typings ourselves.)

Luixo:
@gaearon 
Is there any update on writeability of the `current`?
The problem of composing the refs is still relevant.
If the property should be writeable - I need you to confirm that - I can open a PR in typings repo.
But it looks for me as it should not be.
Maybe we should allow to pass an array of `RefObject`s to ref property of the element?

joaovieira:
Just linked to a real use case:

https://github.com/researchgate/react-intersection-observer/issues/46
https://github.com/researchgate/react-intersection-observer/blob/master/src/IntersectionObserver.js#L107

gaearon:
If someone can't find a workaround for their `findDOMNode` use case please file a new issue and succinctly describe your use case. Preferably with an example. Thanks!

luddzo:
[...]

PR #13841, opened 2018-10-12
Deprecate findDOMNode in StrictMode

sebmarkbage:
# Motivation

The main motivation is that findDOMNode is breaking abstraction levels but allowing a parent to reason about what kind of children a component might render. It creates a refactoring hazard where you can't change the implementation details of a component because a parent might be reaching into its DOM node.

Attaching explicit refs is the alternative. The main thing that used to be difficult is creating a higher order component that seamlessly work with its children. This can now be achieved using forwardRefs.

Up until this point it was a loose recommendation that you don't use this functions but there are certain technical details that are twisting our hands to want to deprecate it.

- In the Fiber architecture it is not possible to cache the "current" node at a higher level since at any given point there are several possible future trees. This causes the findDOMNode algorithm to turn into a very complicated and possibly slow search algorithm in bad cases.
- In the compiler project, it becomes difficult to know if any given DOM node might need to be found. That forces us to replicate the tree structure just in case someone calls it. We'd rather not support this. Instead only explicit refs will be the ones that gets materialized.

We'll only deprecate this in StrictMode for the foreseeable future. It won't get deleted for a long time because lots of existing code depends on it. This is more about preventing new code being written using this outdated technique.

# PR

There are two scenarios. One is that we pass a component instance that is
already in strict mode or the node that we find is in strict mode if
an outer component renders into strict mode.

I use a separate method findHostInstanceWithWarning for this so that
a) I can pass the method name (findDOMNode/findNodeHandle).
b) Can ignore this warning in React Native mixins/NativeComponent that use this helper.

I don't want to expose the fiber to the renderers themselves.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberReconciler.js
+++ b/packages/react-reconciler/src/ReactFiberReconciler.js
@@ -250,7 +250,7 @@ function findHostInstanceWithWarning(
             methodName,
             methodName,
             componentName,
-            getStackByFiberInDevAndProd(fiber),
+            getStackByFiberInDevAndProd(hostFiber),
           );
         } else {
           warningWithoutStack(
--- a/packages/react-reconciler/src/ReactFiberReconciler.js
+++ b/packages/react-reconciler/src/ReactFiberReconciler.js
@@ -242,8 +242,8 @@ function findHostInstanceWithWarning(
           warningWithoutStack(
             false,
             '%s is deprecated in StrictMode. ' +
-              '%s was passed an instance of %s which is in a StrictMode subtree. ' +
-              'Use an explicit ref directly on the element you want to get a handle on.' +
+              '%s was passed an instance of %s which is inside StrictMode. ' +
+              'Instead, add a ref directly to the element you want to reference.' +
               '\n%s' +
               '\n\nLearn more about using refs safely here:' +
               '\nhttps://fb.me/react-strict-mode-find-node',
@@ -256,9 +256,8 @@ function findHostInstanceWithWarning(
           warningWithoutStack(
             false,
             '%s is deprecated in StrictMode. ' +
-              '%s was passed an instance of %s which renders a StrictMode subtree. ' +
-              'The nearest child is in StrictMode. ' +
-              'Use an explicit ref directly on the element you want to get a handle on.' +
+              '%s was passed an instance of %s which renders a StrictMode children. ' +
+              'Instead, add a ref directly to the element you want to reference.' +
               '\n%s' +
               '\n\nLearn more about using refs safely here:' +
               '\nhttps://fb.me/react-strict-mode-find-node',
--- a/packages/react-reconciler/src/ReactFiberReconciler.js
+++ b/packages/react-reconciler/src/ReactFiberReconciler.js
@@ -256,7 +256,7 @@ function findHostInstanceWithWarning(
           warningWithoutStack(
             false,
             '%s is deprecated in StrictMode. ' +
-              '%s was passed an instance of %s which renders a StrictMode children. ' +
+              '%s was passed an instance of %s which renders StrictMode children. ' +
               'Instead, add a ref directly to the element you want to reference.' +
               '\n%s' +
               '\n\nLearn more about using refs safely here:' +
--- a/packages/react-reconciler/src/ReactFiberReconciler.js
+++ b/packages/react-reconciler/src/ReactFiberReconciler.js
@@ -61,6 +61,8 @@ import {
 import {createUpdate, enqueueUpdate} from './ReactUpdateQueue';
 import ReactFiberInstrumentation from './ReactFiberInstrumentation';
 import * as ReactCurrentFiber from './ReactCurrentFiber';
+import {getStackByFiberInDevAndProd} from './ReactCurrentFiber';
+import {StrictMode} from './ReactTypeOfMode';
 
 type OpaqueRoot = FiberRoot;
 
@@ -82,9 +84,11 @@ type DevToolsConfig = {|
 |};
 
 let didWarnAboutNestedUpdates;
+let didWarnAboutFindNodeInStrictMode;
 
 if (__DEV__) {
   didWarnAboutNestedUpdates = false;
+  didWarnAboutFindNodeInStrictMode = {};
 }
 
 function getContextForSubtree(
@@ -209,6 +213,68 @@ function findHostInstance(component: Object): PublicInstance | null {
   return hostFiber.stateNode;
 }
 
+function findHostInstanceWithWarning(
+  component: Object,
+  methodName: string,
+): PublicInstance | null {
+  if (__DEV__) {
+    const fiber = ReactInstanceMap.get(component);
+    if (fiber === undefined) {
+      if (typeof component.render === 'function') {
+        invariant(false, 'Unable to find node on an unmounted component.');
+      } else {
+        invariant(
+          false,
+          'Argument appears to not be a ReactComponent. Keys: %s',
+          Object.keys(component),
+        );
+      }
+    }
+    const hostFiber = findCurrentHostFiber(fiber);
+    if (hostFiber === null) {
+      return null;
+    }
+    if (hostFiber.mode & StrictMode) {
+      const componentName = getComponentName(fiber.type) || 'Component';
+      if (!didWarnAboutFindNodeInStrictMode[componentName]) {
+        didWarnAboutFindNodeInStrictMode[componentName] = true;
+        if (fiber.mode & StrictMode) {
+          warningWithoutStack(
+            false,
+            '%s is deprecated in StrictMode. ' +
+              '%s was passed an instance of %s which is in a StrictMode subtree. ' +
+              'Use an explicit ref directly on the element you want to get a handle on.' +
+              '\n%s' +
+              '\n\nLearn more about using refs safely here:' +
+              '\nhttps://fb.me/react-strict-mode-find-node',
+            methodName,
+            methodName,
+            componentName,
+            getStackByFiberInDevAndProd(fiber),
+          );
+        } else {
+          warningWithoutStack(
+            false,
+            '%s is deprecated in StrictMode. ' +
+              '%s was passed an instance of %s which renders a StrictMode subtree. ' +
+              'The nearest child is in StrictMode. ' +
+              'Use an explicit ref directly on the element you want to get a handle on.' +
+              '\n%s' +
+              '\n\nLearn more about using refs safely here:' +
+              '\nhttps://fb.me/react-strict-mode-find-node',
+            methodName,
+            methodName,
+            componentName,
+            getStackByFiberInDevAndProd(hostFiber),
+          );
+        }
+      }
+    }
+    return hostFiber.stateNode;
+  }
+  return findHostInstance(component);
+}
+
 export function createContainer(
   containerInfo: Container,
   isConcurrent: boolean,
@@ -265,6 +331,8 @@ export function getPublicRootInstance(
 }
 
 export {findHostInstance};
+
+export {findHostInstanceWithWarning};
 
 export function findHostInstanceWithNoPortals(
   fiber: Fiber,
--- a/packages/react-dom/src/__tests__/findDOMNode-test.js
+++ b/packages/react-dom/src/__tests__/findDOMNode-test.js
@@ -154,6 +154,7 @@ describe('findDOMNode', () => {
         'findDOMNode was passed an instance of IsInStrictMode which is inside StrictMode. ' +
         'Instead, add a ref directly to the element you want to reference.' +
         '\n' +
+        '\n    in div (at **)' +
         '\n    in IsInStrictMode (at **)' +
         '\n    in StrictMode (at **)' +
         '\n' +
--- a/packages/react-dom/src/__tests__/findDOMNode-test.js
+++ b/packages/react-dom/src/__tests__/findDOMNode-test.js
@@ -12,6 +12,7 @@
 const React = require('react');
 const ReactDOM = require('react-dom');
 const ReactTestUtils = require('react-dom/test-utils');
+const StrictMode = React.StrictMode;
 
 describe('findDOMNode', () => {
   it('findDOMNode should return null if passed null', () => {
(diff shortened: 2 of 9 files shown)


TrySound:
Does it mean all possible usages of `findDOMNode` are deprecated or just some cases?

gaearon:
Only in strict mode — yes.

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -209,6 +213,68 @@ function findHostInstance(component: Object): PublicInstance | null {
...
+    if (hostFiber.mode & StrictMode) {
+      const componentName = getComponentName(fiber.type) || 'Component';
+      if (!didWarnAboutFindNodeInStrictMode[componentName]) {
+        didWarnAboutFindNodeInStrictMode[componentName] = true;
+        if (fiber.mode & StrictMode) {
+          warningWithoutStack(
+            false,
+            '%s is deprecated in StrictMode. ' +
+              '%s was passed an instance of %s which is in a StrictMode subtree. ' +
+              'Use an explicit ref directly on the element you want to get a handle on.' +
```
sophiebits: How about this:

> %s was passed an instance of %s which is inside StrictMode. Instead, add a ref directly to the element you want to reference.

Also, do we even need "which is inside StrictMode"? It or ConcurrentMode should always appear in the stack unless they're using createRoot.

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -209,6 +213,68 @@ function findHostInstance(component: Object): PublicInstance | null {
...
+            methodName,
+            componentName,
+            getStackByFiberInDevAndProd(fiber),
+          );
+        } else {
+          warningWithoutStack(
+            false,
+            '%s is deprecated in StrictMode. ' +
+              '%s was passed an instance of %s which renders a StrictMode subtree. ' +
+              'The nearest child is in StrictMode. ' +
```
sophiebits: "The nearest child is in StrictMode." feels redundant

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -209,6 +213,68 @@ function findHostInstance(component: Object): PublicInstance | null {
...
+            '%s is deprecated in StrictMode. ' +
+              '%s was passed an instance of %s which is in a StrictMode subtree. ' +
+              'Use an explicit ref directly on the element you want to get a handle on.' +
+              '\n%s' +
+              '\n\nLearn more about using refs safely here:' +
+              '\nhttps://fb.me/react-strict-mode-find-node',
+            methodName,
+            methodName,
+            componentName,
+            getStackByFiberInDevAndProd(fiber),
```
sophiebits: Should we pass hostFiber here too? Since you'll need to find it in order to change your code.

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -209,6 +213,68 @@ function findHostInstance(component: Object): PublicInstance | null {
...
 
+function findHostInstanceWithWarning(
+  component: Object,
+  methodName: string,
+): PublicInstance | null {
+  if (__DEV__) {
+    const fiber = ReactInstanceMap.get(component);
+    if (fiber === undefined) {
+      if (typeof component.render === 'function') {
+        invariant(false, 'Unable to find node on an unmounted component.');
```
sophiebits: I know these are the same, but dev-only invariants feel risky.

packages/react-reconciler/src/ReactFiberReconciler.js
sebmarkbage: It should always be one the same stack since fiber is a child of hostFiber.

sebmarkbage:
Notably this doesn't warn if no node is found. E.g. called on a strict mode component that renders null. I think this is fine because I think the next path would be that we make this always return null for strict mode.

packages/react-reconciler/src/ReactFiberReconciler.js
sebmarkbage: Yea, it is but I'm not confident people understand what that means so clarifying the mechanism might help.

packages/react-reconciler/src/ReactFiberReconciler.js
sophiebits: Other way around, right? They have the same ancestors, but the hostFiber stack is longer/deeper.

packages/react-reconciler/src/ReactFiberReconciler.js
sophiebits: I don't know that this is any clearer than the previous line though.

sebmarkbage: I changed to "StrictMode children". I think that's clearer and fits in one line.

packages/react-reconciler/src/ReactFiberReconciler.js
sebmarkbage: Ah. You mean for the instance case. I figured that you'd want to look at refactoring the top component and then work downwards. But I guess we could use the same stack for both.

packages/react-reconciler/src/ReactFiberReconciler.js
sebmarkbage: yea see what you're saying but refactoring this function is a pita and makes it worse.

Once we fix RN to not use the special NativeMixin/NativeComponent, we can remove this special case and always warn with the same function.

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -209,6 +213,67 @@ function findHostInstance(component: Object): PublicInstance | null {
...
+            methodName,
+            methodName,
+            componentName,
+            getStackByFiberInDevAndProd(hostFiber),
+          );
+        } else {
+          warningWithoutStack(
+            false,
+            '%s is deprecated in StrictMode. ' +
+              '%s was passed an instance of %s which renders a StrictMode children. ' +
```
gaearon: "a StrictMode children" sounds odd, maybe drop "a"?

sag1v:
Hi, Is there any discussion or RFC about deprecating `findDOMNode`? I'm not sure ref forwarding can solve all cases.

oriSomething:
@sag1v I event wrote about it:
https://github.com/reactjs/rfcs/issues/56

It'll just make things even more complicated

sebmarkbage:
I might open an RFC for `<Fragment ref={arrayOfChildNodes => ...}>{children}</Fragment>` which would solve some outstanding issues:

1) It would have a callback that fires when any of them changes so you would have a way to detect when the child components changes.
2) It only needs to know what the "current" children are during the commit phase. Not at arbitrary random times which makes the implementation much faster and more predictable.
3) It's opt-in so React doesn't need to store or emit trees to support these for everything in the whole tree.
4) It works with function components.

What we do know is that the *current* API is broken and needs to be deprecated but we could potentially add a *different* API to support the one remaining use case.

bvaughn:
I haven't thought much about the Fragment-ref proposal, but initial reaction is that I like it, and can see myself using it.

sag1v:
@sebmarkbage This looks solid and totally address the use case i mentioned [here](https://github.com/reactjs/rfcs/issues/56#issuecomment-429610085).

migueloller:
@sebmarkbage, `<Fragment ref={...}>{children}</Fragment>` sounds like a powerful enough replacement for `findDOMNode`. We use `findDOMNode` to portal components to an `<iframe>` and want to keep the frame document clean but need to measure DOM nodes. Instead of wrapping components in a `<div ref={...}>{children}</div>` we use `findDOMNode`. `<Fragment ref={...}>{children}</Fragment>` would be perfect!

eps1lon:
> Only in strict mode — yes, all usage.

Seems like this is not currently the case (will explain later why I think this is good): https://codesandbox.io/s/qko6n6vzjj

It only warns if `findDOMNode` is called with a component. If one calls it with an element node (or nully) then it doesn't warn (because it simply returns the argument) because of an early return in https://github.com/facebook/react/blob/275e76e83bc2be5dc0be9185ff747f383969289a/packages/react-dom/src/client/ReactDOM.js#L615-L617

This is however very valuable for library maintainers that allow users to pass custom components e.g. 
```js
// library.js
// imagine it's a class component
function ({Component}) {
  return <Component ref={ref => doSomething(React.findDOMNode(ref))} />
}
// userland.js
<LibraryComponent Component={UserLandComponent} />
```
This way we can still test the library codebase for bad components that don't properly forward their refs and keep supporting users both in and out of `React.StrictMode`.

I know this is most likely a case of "Shouldn't have done that in the first place" but it would be really nice if we could support this use case.

If this was your intention I'll gladly add a test to avoid regression. Might be also good idea to amend the blog post.

oliviertassinari:
@sebmarkbage We are looking into migrating away from the `findDOMNode()` method on Material-UI. @eps1lon has started experimenting. Now, I'm wondering. `forwardedRef` is a great, it's definitely solving the higher-order component issue.

### But how are we supposed to solve the composition problem?

Let's say we have the following React tree `A > B > C`:
- `A` a class component.
- `B` a class component.
- `C` a simple `div` primitive.

How does `A` can get a ref on `C` without agreeing on some sort of "standard" property for cascading the reference? [Like in the documentation using `forwardedRef`](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-in-higher-order-components). So far, our answer on Material-UI is the [`RootRef`](https://material-ui.com/api/root-ref/) component.
```jsx
import React from 'react';
import RootRef from '@material-ui/core/RootRef';

class MyComponent extends React.Component {
  constructor() {
    super();
    this.domRef = React.createRef();
  }

  componentDidMount() {
    console.log(this.domRef.current); // DOM node
  }

  render() {
    return (
      <RootRef rootRef={this.domRef}>
        <SomeChildComponent />
      </RootRef>
    );
  }
}
```

It's a component we use internally for the [Tootlip](https://material-ui.com/demos/tooltips/) or that we can use for the [ClickAwayListener](https://material-ui.com/utils/click-away-listener/). 

`<Fragment ref={arrayOfChildNodes => ...}>{children}</Fragment>` would be awesome!

oliviertassinari:
@gaearon Sure https://codesandbox.io/s/105v4q6n0l

sag1v:
@oliviertassinari I have the exact same use case and now the API is forced to expose a `ref`. I've posted an example in [this comment](https://github.com/reactjs/rfcs/issues/56#issuecomment-429610085).  
I wish we could have another solution which won't suffer from the downsides `findDOMNode` has but will be able to provide the same capabilities. :crossed_fingers: 
`findDOMNode` allowed us to provide a much simpler API for consumers and without exposing implementation details like refs.

oliviertassinari:
@sag1v It's the best solution I can think of too. We will have to release a major on Material-UI side if it's the go to approach, and have people change their implcation. But I think that it's laborious and error prone (hard to debug), they might be a better way 🤔.
```jsx
<TrapWithRef event="click">
  {(trapped, ref) => <Box innerRef={ref} isFocused={trapped} />}
</TrapWithRef>
```

gaearon:
I'm still confused why you have `<Box innerRef>` in examples rather than `<Box ref>`.

sag1v:
@gaearon `<Box innerRef>` is the consumer, it's just an example of the usage.
If it's clearer then it can be something like this:  

```
<TrapWithRef event="click">
  {
    (trapped, ref) => (
      <div>
        <div ref={ref}>{trapped ? 'open' : 'close' }</div>
      </div>
    )
  }
</TrapWithRef>
```

The only advantage for this approach is that the consumer can decide where to attach the `ref`.

gaearon:
I mean that you can pass `<Box ref>`, and `Box` can forward it anywhere. Just making sure we're on the same page that you don't need to introduce ref-like props to your API. Specifically, I still don't understand why

>agreeing on some sort of "standard" property for cascading the reference

is necessary, and how that's relevant to the example above.

Macil:
I recently updated [ReactFloatAnchor](https://github.com/StreakYC/react-float-anchor) and [ReactMenuList](https://github.com/StreakYC/react-menu-list) to no longer use findDOMNode, and I had to make changes similarly to @sag1v. ReactFloatAnchor takes [`ref => ReactNode` instead of just `ReactNode` now](https://github.com/StreakYC/react-float-anchor/blob/master/CHANGELOG.md#300-2018-10-30). In ReactMenuList's [SubMenuItem](https://github.com/StreakYC/react-menu-list/blob/master/src/SubMenuItem.js#L269), I needed to use ReactFloatAnchor and pass the ref onto MenuItem's dom element, so I added a "domRef" prop to MenuItem which MenuItem passes as a ref to its div. (This was also pretty awkward, because [MenuItem](https://github.com/StreakYC/react-menu-list/blob/master/src/MenuItem.js) itself needs a ref to its own div, so I had to pass a function as the ref to the div which called my own setRef function with props.domRef if present.)

@gaearon In my case, I couldn't use ref forwarding because MenuItem has many imperative methods and users need to be able to get a ref to the MenuItem itself too.

oliviertassinari:
> I'm still confused why you have <Box innerRef> in examples rather than <Box ref>.

@gaearon The fact that we don't understand each other is a good sign, something is off. As a library author, we don't control what the user's `Box` component will provide. It can be a functional component, that don't accept reference or it can be a class component, that will return a reference to the class component instance. Either way, we don't have access to the underlying DOM node.

sag1v:
@gaearon I made a [small and concise codesandbox example](https://codesandbox.io/s/30z5p94yqq) (a very naive implementation ) with just the parts that are relevant for this issue (from my angle at least, maybe @oliviertassinari has some more use-cases ).

As you can see, both `Trap` and `TrapWithRef` are almost identical, 
The only difference is that `Trap` uses `findDOMNode` and doesn't care what and how the children are rendered. its just finds the first DOM node and use it for its internal logic.

As for `TrapWithRef`, its exposing a `ref` **callback** and will only work if and when the child (consumer of the library) will use it and pass it to some DOM element.  

It may seem like a small change to the API but this is quite significant IMHO. it feels awkward and confusing (for the consumers).

I'll be honest, there is an upside for this pattern though, The consumer gets to decide where or what exactly to "Trap". but this is a very small upside, IMO.

sag1v:
Another concern i have for exposing an explicit `ref` instead of using `findDOMNode` is that it may conflict with other components.  
For example, if someone is using 2 parent components that needs a ref to the element (for a DOM logic) they can't assign it to the same DOM element (as far as  i know).  

Given this example, how can we use both `ref` and `ref2`?
```
<ElementResize>
  {(rect, ref) => (
    <Trap event="click">
      {(trapped, ref2) => (
        <div ref={???}>
          <div> {trapped ? 'focused' : 'blur'}</div>
          <div>{rect.width}</div>
        </div>
      )}
    </Trap>
  )}
</ElementResize>
```

oliviertassinari:
> Given this example, how can we use both ref and ref2?

@sag1v We have an helper for that on Material-UI. But yes, it's a bit verbose otherwise.

TrySound:
@sag1v With hooks ref composition may not be required
```js
const Component = () => {
  const ref = React.useRef()
  const rect = useElementResize(ref)
  const trapped = useTrap(ref)

  return (
    <div ref={ref}>
      <div> {trapped ? 'focused' : 'blur'}</div>
      <div>{rect.width}</div>
    </div>
  )
}
```

sag1v:
@TrySound Yeah i know, i already played with it 😄 

image

Macil:
@sebmarkbage Would the Fragment ref feature give refs to the direct child DOM nodes specifically, or just refs to whatever dom nodes or component instances are below? Example:

```js
class Foo extends React.Component {
  render() {
    return <div>aaaa <span>b</span></div>;
  }
}

class Bar extends React.Component {
  render() {
    return <Fragment ref={arrayOfChildNodes => ...}> <Foo /> </Fragment>
  }
}
```

Would arrayOfChildNodes contain a Foo instance, or an HTMLDivElement instance?

The DOM node behavior would be a useful findDOMNode replacement (https://github.com/styled-components/styled-components/issues/2154), though the component behavior would massively simplify the case where you want refs on a dynamic list of elements (See this [SO thread](https://stackoverflow.com/questions/47055464/how-to-assign-refs-to-multiple-components) about the situation. The accepted answer has a few minor issues. The other answer doesn't directly answer the question. I previously created an [npm module](https://github.com/Macil/react-multi-ref) specifically to solve this scenario.) and also the ref-sharing case (https://github.com/facebook/react/issues/13029). (Maybe fragment could support both as different props like `domRef` and `ref`.)

sebmarkbage:
Wrote an RFC for it. https://github.com/reactjs/rfcs/blob/fragmentrefs-rfc/text/0000-fragment-refs.md

PR #13842, opened 2018-10-12
[scheduler] Rename priority levels

acdlite:
- "Interactive" -> "User-blocking"
- "Whenever" -> "Idle"

These are the terms used by @spanicker in their main-thread scheduling proposal: https://github.com/spanicker/main-thread-scheduling#api-sketch

That proposal also uses "microtask" instead of "immediate" and "default" instead of "normal." Not sure about "microtask" because I don't think most people know what that is. And our implementation isn't a proper microtask, though you could use it to implement microtasks if you made sure to wrap every entry point. I don't really have a preference between "default" and "normal."

These aren't necessarily the final names. Still prefixed by `unstable_`.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -10,9 +10,9 @@
 
 // TODO: Use symbols?
 var ImmediatePriority = 1;
-var InteractivePriority = 2;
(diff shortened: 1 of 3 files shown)


bvaughn:
me too

PR #13503, opened 2018-08-28
When a root expires, flush all expired work in a single batch

acdlite:
Still need to add some Suspense tests, and some comments. Pushing so Dan can test if it fixes the issue in the fixture.

PR #13847, opened 2018-10-13
Add failing test for expiration

gaearon:
I don't know if this is a bug but this behavior seems weird.

The sequence is:

* Perform update 1
* Schedule update 2
* Start rendering update 2
* Advance time so that update 2 is expired
* Schedule update 3
* Flush expired work

My expectation is this should only flush update 2 (it's the one that expired). Update 3 hasn't expired yet.

However, this flushes update 3 instead. If I revert https://github.com/facebook/react/pull/13503, this flushes both 2 and 3. I can't get it to only flush 2, but that seems like the desired behavior.

By itself flushing 3 isn't that bad. The real situation I was trying to reproduce is when we expire another update *while* rendering a just-expired update, and that kicks off another rendering cycle without yielding. But I couldn't get to reproducing that because even a more basic scenario above didn't work as I expected.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactIncrementalUpdates-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactIncrementalUpdates-test.internal.js
@@ -602,4 +602,66 @@ describe('ReactIncrementalUpdates', () => {
     expect(ReactNoop.getChildren('a')).toEqual([span('goodbye')]);
     expect(ReactNoop.getChildren('b')).toEqual([span('goodbye')]);
   });
+
+  fit('just gettin a feel for it', () => {
+    class Foo extends React.Component {
+      componentDidUpdate() {
+        ReactNoop.yield('Commit: ' + this.props.prop);
+      }
+      componentDidMount() {
+        ReactNoop.yield('Commit: ' + this.props.prop);
+      }
+      render() {
+        ReactNoop.yield('Render: ' + this.props.prop);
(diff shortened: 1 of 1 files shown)


gaearon:
ok, this particular behavior is because we don't recalculate time:

https://github.com/facebook/react/blob/4773fdf7cdf5d6d775ad0960f23ee1a830e7b82b/packages/react-reconciler/src/ReactFiberScheduler.js#L1947-L1952

gaearon:
Even if we recalculated time in the test (which I'm not sure how to do), the actual issue I'm trying to debug is here:

https://github.com/facebook/react/blob/4773fdf7cdf5d6d775ad0960f23ee1a830e7b82b/packages/react-reconciler/src/ReactFiberScheduler.js#L2126

When something expires, we keep doing all expired work. But this we don't show it to the user so from their perspective we're just freezing the browser. Not very useful. Seems like we should yield sometimes (always?) if we just flushed something that expired.

I tried just adding a `break` into this loop when `currentRendererTime >= nextFlushedExpirationTime` (which means we just rendered something expired). But this wasn't sufficient because some other callback from the scheduler got called in the same tick, pulling us into this loop again.

Remembering `currentRendererTime` during `break` and exiting further `performWork` calls with the same `currentRendererTime` seemed to work but is gross?

gaearon:
This might have to do with the "terminal updates" work @acdlite is doing. Or maybe not. I don't know if this is still a problem but it came up in my time slicing demos.

PR #13855, opened 2018-10-15
Add support for React.pure in ReactDOMServer

alexmckenley:
Add support for REACT_PURE_TYPE in ReactDOMServer + tests

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRenderer.js
+++ b/packages/react-dom/src/server/ReactPartialRenderer.js
@@ -1004,18 +1004,12 @@ class ReactDOMServerRenderer {
           }
           case REACT_PURE_TYPE: {
             const element: ReactElement = ((nextChild: any): ReactElement);
-            let nextChildren;
-            if (
-              typeof elementType.render === 'object' && elementType.render !== null && elementType.render.$$typeof === REACT_FORWARD_REF_TYPE
-            ) {
-              nextChildren = toArray(
-                elementType.render.render(element.props, element.ref),
-              );
-            } else {
-              nextChildren = toArray(
-                elementType.render(element.props),
-              );
-            }
+            let nextChildren = [
+              React.createElement(
+                elementType.type,
+                Object.assign({ref: element.ref}, element.props),
+              ),
+            ];
             const frame: Frame = {
               type: null,
               domNamespace: parentNamespace,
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/__tests__/ReactDOMServerIntegrationPure-test.internal.js
```diff
@@ -0,0 +1,112 @@
+/**
```
gaearon: There's a lot of setup cost for initializing these test modules. Can we put this together with some other tests since there's only a few of test cases here?

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -945,7 +946,8 @@ class ReactDOMServerRenderer {
       }
       if (typeof elementType === 'object' && elementType !== null) {
         switch (elementType.$$typeof) {
-          case REACT_FORWARD_REF_TYPE: {
+          case REACT_FORWARD_REF_TYPE:
+          case REACT_PURE_TYPE: {
             const element: ReactElement = ((nextChild: any): ReactElement);
             const nextChildren = toArray(
               elementType.render(element.props, element.ref),
```
sebmarkbage: This is not correct. It is passing the `element.ref` second argument. It shouldn't get the second argument in the `pure` case.

gaearon: But https://github.com/facebook/react/pull/13822?

sebmarkbage: I didn't see that one. I'm not sure I approve. Let me think about it.

alexmckenley:
Moved the tests and branched the REACT_PURE_TYPE logic. I'm still passing the forward ref stuff for now to keep the behavior in sync with ReactDOM (#13855). Let me know if you want to change this behavior.

gaearon:
Yeah we're going to revert that and change so that:

* `pure` gets one argument
* It should be possible to render `pure(forwardRef())` (but not the other way around)

@acdlite has details

alexmckenley:
Updated my diff so `pure` works with both functions and forwardRef elements. The tests should pass once `pure` and react Fiber are updated to respect this new logic.

sebmarkbage:
I updated on top of #13903. This now uses a slow path to implement pure in a fully recursive form. This is unfortunate but that's the easiest given the structure of the partial renderer. It doesn't make it easy to make recursive custom types.

But we're rewriting this whole engine soon anyway.

PR #13857, opened 2018-10-16
[reconciler] ReactFiberNewContext import maxSigned31BitInt twice

yongningfu:
in the ReactFiberNewContext, import maxSigned31BitInt twice


yongningfu:
Thank you for your comment. MAX_SIGNED_31_BIT_INT does look better

PR #13860, opened 2018-10-16
Add warning about css shorthand properties collision(#6348)

### LINKED ISSUE #6348: Warn when using overlapping styles (e.g. border and borderBottom)
# This "good first bug" is taken by @YongPilMoon. Don't work on it unless that's you!

This has been discussed a few times before but I don’t think there was any conclusion, and PRs intending to solve the issue were not merged for various reasons.

I would like to close those PRs as outdated, and reboot the discussion about this.
Performance considered, the conclusion from @sebmarkbage and @syranide seems to be:

> We should consider not allowing conflicting style rules at all.

— https://github.com/facebook/react/pull/2013#issuecomment-57338177

> IMHO, all things considered it's better to just disallow overlapping and warn in dev.

— https://github.com/facebook/react/pull/4661#issuecomment-132996649

Radium by @ianobermiller came to the same conclusion in https://github.com/FormidableLabs/radium/issues/95 but there was some backlash afterwards. React Native seems to allow style expansions but only for a few attributes (e.g. `margin` and `padding`, but not `border`).

I’m closing old pull requests about this, and creating this issue to track implementation of the behavior we seem to agree upon: we should warn in `__DEV__` when `border` and `borderBottom` are used at the same time. We can discuss more specifics (should either be ignored? should we allow a few whitelisted properties but warn for others?) in this issue.

As a migration strategy, we can suggest people to use something like https://github.com/ActionIQ/style-builder if they really need those shortcuts. It’s also something we’ll need to decide upon when implementing integrated styling.

---

Related issues:
- #2231
- #2407
- #5030
- https://github.com/FormidableLabs/radium/issues/95

Related PRs:
- #2013
- #4661

YongPilMoon:
I only added background shorthand properties.
After code review, I would like to add rest of them.

Fixes #6348.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
@@ -2698,5 +2698,34 @@ describe('ReactDOMComponent', () => {
       ReactDOM.render(<Component />, container);
       expect(typeof container.onclick).not.toBe('function');
     });
+
+    it('should warn about css shorthand properties collision', () => {
+      const container = document.createElement('div');
+      const oldStyle = {
+        background: 'url(http://example.org/image/a.jpg) no-repeat center',
+        backgroundSize: '150px',
+        backgroundColor: 'red',
+      };
+
+      ReactDOM.render(<div style={oldStyle} />, container);
+      const stubStyle = container.firstChild.style;
+
+      expect(stubStyle.background).toEqual(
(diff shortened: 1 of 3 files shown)


sophiebits:
Thanks for sending this @YongPilMoon! I realized as I was reviewing that solving this correctly is a bit tricker than I had anticipated so I opted to finish it off myself in #14181. Feel free to comment on my PR if you notice any issues with it, and thanks for taking a first stab at this.

PR #13861, opened 2018-10-16
Remove unstable_read() in favor of direct dispatcher call

gaearon:
It's an advanced feature that shouldn't be used by application code directly. For example classes are supposed to use `contextType` (https://github.com/reactjs/reactjs.org/pull/1265) instead.

So we want to hide it a bit deeper. This makes it only available via direct dispatcher read. Also removes an extra runtime check.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberClassComponent.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.js
@@ -20,6 +20,7 @@ import {
 import ReactStrictModeWarnings from './ReactStrictModeWarnings';
 import {isMounted} from 'react-reconciler/reflection';
 import * as ReactInstanceMap from 'shared/ReactInstanceMap';
+import ReactSharedInternals from 'shared/ReactSharedInternals';
 import shallowEqual from 'shared/shallowEqual';
 import getComponentName from 'shared/getComponentName';
 import invariant from 'shared/invariant';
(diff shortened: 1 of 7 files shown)


TrySound:
This is sad. `.read()` api is quite useful for translations. Will you go with this at least in `react-cache` or we should wait for `adopt` syntax solution?

gaearon:
I mean, you can still use it in a library. Your translation library could literally do the same thing that `react-cache` would do.

```js
  function readContext(Context) {
    const dispatcher =
      React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.ReactCurrentOwner
        .currentDispatcher;
    return dispatcher.readContext(Context);
  }
```

TrySound:
Great! Thank you.

markerikson:
Uh... that seems really icky.  Why are you effectively _encouraging_ digging around in React's internals?  What's the reasoning for removing this?

markerikson:
So I'm starting to poke around at potential ways to re-revamp `connect` to use direct subscriptions again (per https://github.com/reduxjs/react-redux/issues/1177 ), and the very first question I have is, "how are we supposed to get the store reference out of context so we can subscribe in `cDM`?"

`Context.unstable_read()` still seems like it would have been the best answer to that question.

gaearon:
`contextType` in a class (16.6+), `useContext` in a function (coming). 

Don’t have a good answer for you that’s guaranteed to work earlier.

markerikson:
Yeah, I know.  That's why I'm diving off the deep end and experimenting with a hooks-based approach first.

`contextType` isn't exactly an option atm, because in v6 we allow the user to customize what context the component uses by passing it as a prop.

PR #13862, opened 2018-10-16
[WIP] Replay load/error events that happen earlier than commit

sebmarkbage:
If we create an img/image/link those might fire their load event before we commit. If that happens, we need to store the event so that we can replay it during the commit phase.

I had an idea to fix it but this is still WIP because I actually need to test if this approach works.

- [x] Write code
- [ ] Write unit tests
- [ ] Write fixtures

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMEarlyEvent-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMEarlyEvent-test.js
@@ -1,0 +1,79 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails react-core
+ */
+
+'use strict';
+
+let React;
+let ReactDOM;
+let ConcurrentMode;
+
+beforeEach(() => {
+  React = require('react');
+  ConcurrentMode = React.unstable_ConcurrentMode;
+  ReactDOM = require('react-dom');
+});
+
+describe('ReactDOMImg', () => {
+  let container;
+
+  beforeEach(() => {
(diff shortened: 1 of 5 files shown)


sophiebits:
Is it not possible to receive more than one event before mount?

sebmarkbage:
@sophiebits No, because you can't change the src. At least not without resuming of host nodes which we don't have yet. If there's only one resource, it either loads or errors.

sebmarkbage:
The timing of these events are a bit weird. I guess they will fire in the "layout" phase. So it'll fire before componentDidMount of the parent that rendered it. It's also before the ref on the image has fired so you don't even have a ref on it yet. This is consistent with onFocus/Blur events of auto-focused elements.

It's a bit weird that it is so late but not fully after initialization. However, it is good because it means you can setState at this point. setState before commit gets sketchy.

NE-SmallTown:
Does audio/video/script/source also need this?

packages/react-dom/src/__tests__/ReactDOMEarlyEvent-test.js
```diff
@@ -0,0 +1,79 @@
...
+
+    // TODO: Write test
+
+    ReactDOM.render(
+      <ConcurrentMode>
+        <img onLoad={onLoadSpy} />
+      </ConcurrentMode>,
+    );
+
+    // someHowGetAnImage.dispatchEvent(loadEvent);
```
sebmarkbage: @trueadm I think we’ll need to spy on document.createElement by wrapping it so we can detect which img instance React creates.

We can’t use a ref to simulate this because it doesn’t resolve in time.

gaearon:
This was fixed in https://github.com/facebook/react/pull/23316.

PR #13293, opened 2018-07-31
Context.set(newGlobalValue)

acdlite:
Adds a method to update a context's default (global) value. This schedules an update on all context consumers that are not nested inside a provider, across all roots and renderers. For most use cases, this replaces the need to inject a context provider at the top of each root.

(I've added an `unstable_` prefix until it's ready for public release.)

PR #13865, opened 2018-10-17
[react-cache] Remove `cache` as argument to `read`

acdlite:
Updated is API is `Resource.read(key)` instead of `Resource.read(cache, key)`.

The cache is read from context using `readContext`.

This also removes cache invalidation entirely (other than the default LRU mechanism), as well as the ability to have multiple caches. We'll add it back once `Context.write` (some version of #13293) lands and we can implement it the right way.

Since there's now only a single cache (the global one), we don't actually need to use context yet, but I've added a dummy context anyway so the user gets an error if they attempt to read outside the render phase.


packages/react-cache/src/__tests__/ReactCache-test.internal.js
```diff
@@ -0,0 +1,327 @@
...
+    // We've now rendered values 1, 2, 3, 4, 5, over our limit of 3. The least
+    // recently used values are 2 and 3. They will be evicted during the
+    // next sweep.
+    evictLRU();
+
+    root.update(
+      <Suspense>
+        <AsyncText ms={100} text={1} />
+        <AsyncText ms={100} text={2} />
+        <AsyncText ms={100} text={3} />
```
NE-SmallTown: Should we use or add a PureComponent of AsyncText case like :

```
<AsyncText ms={100} text={2} />
<AsyncText ms={100} text={3} />
<AsyncText ms={100} text={1} />
```
here? IMO, we need a case which just test the cache read.

acdlite: This test does read from the cache? And idk what you mean by "common reconciliation"

NE-SmallTown: Sorry. I mean, this is just test the function component(the `AsyncText`) but in real world, we have class component which can be `PureComponent` or normal `Component`, so in the common reconciliation, they will go to different switch case or codes, I think it maybe should be tested too.

packages/react-cache/src/ReactCache.js
```diff
@@ -10,394 +10,178 @@
...
+  if (entry === undefined) {
+    const thenable = fetch(input);
+    thenable.then(
       value => {
-        // Resource loaded successfully.
-        const resolvedRecord: ResolvedRecord<K, V> = (pendingRecord: any);
-        resolvedRecord.status = Resolved;
-        resolvedRecord.suspender = null;
-        resolvedRecord.value = value;
+        if (newResult.status === Pending) {
```
aweary: What is this `status` check is guarding against?

PR #13866, opened 2018-10-17
Add warning message when Suspense fallback is undefined

### LINKED ISSUE #13864: Suspense component should only capture if fallback prop is defined
*[Edit by @acdlite: Decided in comments below that we will change the semantics so that a missing `fallback` prop means the exception should propagate to the next parent (like a rethrow). That way a Suspense component can specify other props like `maxDuration` without needing to provide a fallback, too.]*

---

Bug
`<React.Suspense>` does not warn you if you omit a `fallback`). While redundant for TS/Flow usage,  I misspelled the `fallback` prop by accident in a playground and was tearing my hair out trying to figure out why things were not working as expected. 

Omit a `fallback`. 

React should warn during development if `fallback` is `undefined`.

Master

imbhargav5:
Aims to fix #13864 
- [x] Add warning message within reconciler
- [x] Add test case to show a warning when `fallback` prop is missing.
- [ ]  Fix the many test cases which do not have `fallback` as a prop in jest `development` environment. How should I go about this? 

Should I add a `fallback={null}` prop for existing test cases which don't have a fallback prop defined or should I add an `expect().toWarnDev` to all of them?

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -55,6 +55,7 @@ import getComponentName from 'shared/getComponentName';
 import ReactStrictModeWarnings from './ReactStrictModeWarnings';
 import warning from 'shared/warning';
 import warningWithoutStack from 'shared/warningWithoutStack';
+import lowPriorityWarning from 'shared/lowPriorityWarning';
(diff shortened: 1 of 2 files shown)


sebmarkbage:
I believe this was effectively fixed by #13879 since it should now error if there is no fallback specified anywhere above.

PR #13875, opened 2018-10-17
[RFC] FIX: Context pollutes mutable global state

leebyron:
Fixes #13874

This is just a straw-man approach for a fix for #13874 that avoids mutating global context by giving each context instance a unique key it can use to store its value in a context object that's scoped to the current render.


leebyron:
Closing in favor of #13877 which is a simpler and probably more efficient fix

PR #13877, opened 2018-10-17
FIX: Context pollutes mutable global state (with "thread local storage")

leebyron:
Fixes #13874

Alternative to #13875

This straw-man approach also avoids mutating global context by giving each ReactPartialRenderer its own mutable state and using context instances as keys into that mutable state.


leebyron:
Hey @sebmarkbage, how are you all thinking of the next version cut. Is there any chance of getting this into a patch release?

sebmarkbage:
Landed #14182 instead but would be good to perf test both to validate the theory.

PR #13879, opened 2018-10-18
Suspense component does not capture if `fallback` is not defined

### LINKED ISSUE #13864: Suspense component should only capture if fallback prop is defined

acdlite:
A missing fallback prop means the exception should propagate to the next
parent (like a rethrow). That way a Suspense component can specify other
props like maxDuration without needing to provide a fallback, too.

Closes #13864 

cc: @jaredpalmer

### DIFF
--- a/packages/react-reconciler/src/ReactFiberUnwindWork.js
+++ b/packages/react-reconciler/src/ReactFiberUnwindWork.js
@@ -205,9 +205,9 @@ function throwException(
     workInProgress = returnFiber;
     do {
       if (workInProgress.tag === SuspenseComponent) {
-        const state = workInProgress.memoizedState;
-        const didTimeout = state !== null && state.didTimeout;
-        if (!didTimeout) {
+        const fallback = workInProgress.memoizedProps.fallback;
+        const didTimeout = workInProgress.memoizedState;
+        if (fallback !== undefined && !didTimeout) {
           // Found the nearest boundary.
 
           // If the boundary is not in concurrent mode, we should not suspend, and
--- a/packages/react-reconciler/src/ReactFiberUnwindWork.js
+++ b/packages/react-reconciler/src/ReactFiberUnwindWork.js
@@ -207,7 +207,7 @@ function throwException(
       if (workInProgress.tag === SuspenseComponent) {
         const fallback = workInProgress.memoizedProps.fallback;
         const didTimeout = workInProgress.memoizedState;
-        if (fallback !== undefined && !didTimeout) {
+        if (!didTimeout && workInProgress.memoizedProps.fallback !== undefined) {
           // Found the nearest boundary.
 
           // If the boundary is not in concurrent mode, we should not suspend, and
--- a/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
@@ -106,7 +106,7 @@ describe('ReactSuspense', () => {
     function Foo() {
       ReactTestRenderer.unstable_yield('Foo');
       return (
-        <Suspense>
+        <Suspense fallback={<Text text="Loading..." />}>
           <Bar>
             <AsyncText text="A" ms={100} />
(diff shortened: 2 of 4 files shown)


packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -211,8 +211,9 @@ function throwException(
     workInProgress = returnFiber;
     do {
       if (workInProgress.tag === SuspenseComponent) {
+        const fallback = workInProgress.memoizedProps.fallback;
         const didTimeout = workInProgress.memoizedState;
-        if (!didTimeout) {
+        if (fallback !== undefined && !didTimeout) {
```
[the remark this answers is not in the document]
gaearon: Also what happens if you sometimes specify it and sometimes don't? Is that a legit or broken pattern? Does it need a warning, like controlled<->uncontrolled "flapping" input?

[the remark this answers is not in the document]
sophiebits: Prior art for our APIs (particularly defaultProps) is that undefined is treated as absent. Otherwise it is a pain to wrap Placeholder in your own component that passes a `fallback` through.

sebmarkbage:
```suggestion
        if (!didTimeout && workInProgress.memoizedProps.fallback !== undefined) {
```

packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
```diff
@@ -271,4 +272,49 @@ describe('ReactSuspense', () => {
...
+    jest.advanceTimersByTime(1000);
+    expect(ReactTestRenderer).toHaveYielded([]);
+    expect(root).toFlushAndYield([]);
+    expect(root).toMatchRenderedOutput('Loading...');
+
+    jest.advanceTimersByTime(5000);
+    expect(ReactTestRenderer).toHaveYielded(['Promise resolved [Hi]']);
+    expect(root).toFlushAndYield(['Hi']);
+    expect(root).toMatchRenderedOutput('Hi');
+  });
```
NE-SmallTown: This test case is weird, I don't think we need provide this usage of Suspense(at lease for this case behavior). In this case, we set the `maxDuration` option  but it is useless, seems we use the `maxDuration` of the outer Suspense(default `0`?), this is not expected.

So just from my personal perspective, we have 2 ways:
1. Warn about this usage if we allow to do this

or 

2. Use the `maxDuration`(if provided) of inner and `fallback` of outer

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -205,9 +205,9 @@ function throwException(
     workInProgress = returnFiber;
     do {
       if (workInProgress.tag === SuspenseComponent) {
-        const state = workInProgress.memoizedState;
-        const didTimeout = state !== null && state.didTimeout;
-        if (!didTimeout) {
+        const fallback = workInProgress.memoizedProps.fallback;
+        const didTimeout = workInProgress.memoizedState;
+        if (!didTimeout && workInProgress.memoizedProps.fallback !== undefined) {
```
bvaughn: Just noticed this. Looks like an oversight that you read from `workInProgress.memoizedProps.fallback` twice rather than use the `fallback` const.

gaearon: Seems like this code already changed.

PR #13885, opened 2018-10-19
Lazy components must use React.lazy

acdlite:
Removes support for using arbitrary promises as the type of a React element. Instead, promises must be wrapped in `React.lazy`. This gives us flexibility later if we need to change the protocol.

The reason is that promises do not provide a way to call their constructor multiple times. For example:

```js
const promiseForA = new Promise(resolve => {
  fetchA(a => resolve(a));
});
```

Given a reference to `promiseForA`, there's no way to call `fetchA` again. Calling `then` on the promise doesn't run the constructor again; it only attaches another listener.

In the future we will likely introduce an API like `React.eager` that is similar to `lazy` but eagerly calls the constructor. That gives us the ability to call the constructor multiple times. E.g. to increase the priority, or to retry if the first operation failed.


wheelo:
Awesome

PR #13886, opened 2018-10-19
React.lazy constructor must return result of a dynamic import

acdlite:
## Based on #13885

We may want to change the protocol later, so until then we'll be restrictive. Heuristic is to check for existence of `default`.

See #13885 for more details.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactServerRendering-test.js
+++ b/packages/react-dom/src/__tests__/ReactServerRendering-test.js
@@ -571,14 +571,19 @@ describe('ReactDOMServer', () => {
       ReactDOMServer.renderToString(<React.unstable_Suspense />);
     }).toThrow('ReactDOMServer does not yet support Suspense.');
 
+    async function fakeImport(result) {
+      return {default: result};
+    }
+
     expect(() => {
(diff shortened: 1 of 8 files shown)


NE-SmallTown:
If this is just for Component(from the codes and naming like `getResultFromResolvedLazyComponent`), should we add a check for the `default` field?

arianon:
Doesn't this make the case of importing a lazy component that isn't a module's default export a bit less obvious?

Before
```js
const FooComponent = React.lazy(() =>
  import("component-library").then(module => module.FooComponent)
);
```

After
```js
const FooComponent = React.lazy(() =>
  import("component-library").then(module => ({
    default: module.FooComponent
  }))
);
```

milesj:
Why not check for `__esModule` and then choose `default`? That would support Babel/TS compiled files out of the box.

```
// Babel exported default
if (typeof result === 'object' && result.__esModule && result.default) {
  return result.default;
}

// module.exports default
return result;
```

gaearon:
Please comment on the RFC if you have feedback. It contains the motivation and more details: https://github.com/reactjs/rfcs/pull/64

PR #13896, opened 2018-10-19
Bugfix for #13886

acdlite:
Fixes a bug where a lazy component does not cache the result of its constructor.


packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
```diff
@@ -40,7 +44,7 @@ describe('ReactLazy', () => {
     expect(root).toFlushAndYield(['Loading...']);
     expect(root).toMatchRenderedOutput(null);
 
-    await LazyText;
+    await {};
```
gaearon: Uh?

PR #13897, opened 2018-10-19
Allow forwardRef to be composed with pure

sebmarkbage:
This uses a different approach than the moddable approach because that approach also started getting quite complex and affects tags and symbols.

Instead, I just add a `compare` field to forwardRef's element type and use `undefined` to represent the lack of pure wrapper and null to represent the default.

This way we don't affect the perf of pure and when we drop forwardRef we can just drop this path.

This also handles the composition of multiple pure wrappers with custom comparisons.


sebmarkbage:
Superseded by #13903 which is better since it allows any type to be passed to pure.

PR #13899, opened 2018-10-20
Class component can suspend without losing state outside concurrent mode

acdlite:
Outside of concurrent mode, schedules a force update on a suspended class component to force it to prevent it from bailing out and reusing the current fiber, which we know to be inconsistent.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -401,32 +401,41 @@ function updateClassComponent(
   }
   prepareToReadContext(workInProgress, renderExpirationTime);
 
+  const instance = workInProgress.stateNode;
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -401,32 +401,41 @@ function updateClassComponent(
...
-        workInProgress,
-        Component,
-        nextProps,
-        renderExpirationTime,
-      );
+  if (instance === null) {
+    if (current !== null) {
+      // An class component without an instance only mounts if it suspended
+      // inside a non- concurrent tree, in an inconsistent state. We want to
+      // tree it like a new mount, even though an empty version of it already
```
bvaughn: s/tree it/treat it/ 😄

PR #13900, opened 2018-10-20
Centralize props memoization

sebmarkbage:
We compare pendingProps against memoizedProps in beginWork so we already have strong assumptions that can't vary by branch. So we can simplify this which lets us get rid of the special case in lazy.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -230,7 +230,6 @@ function updateForwardRef(
     nextChildren,
     renderExpirationTime,
   );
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -314,7 +311,6 @@ function updateMode(
     nextChildren,
     renderExpirationTime,
   );
-  memoizeProps(workInProgress, nextChildren);
```
sebmarkbage: This one is interesting because it doesn't memoize its props. Only its children. This is a bug since it won't bail out ever.

PR #13901, opened 2018-10-20
Always bail out timed out children even if they receive an update

acdlite:
I moved resetChildExpirationTime to ReactFiberCompleteWork so we can insert an additional reset only for SuspenseComponent.


sophiebits:
wow, whoever paired on this with you must be really smart

PR #13902, opened 2018-10-20
Unfork Lazy Component Branches

sebmarkbage:
By adding an extra field for reconciliation purposes we can avoid the Lazy special cases. Normally it wouldn't be worth it but we can start using this technique for other things like pure.


packages/react-reconciler/src/ReactFiber.js
```diff
@@ -498,14 +510,27 @@ export function createFiberFromElement(
...
 
+  return fiber;
+}
+
+export function createFiberFromElement(
+  element: ReactElement,
+  mode: TypeOfMode,
+  expirationTime: ExpirationTime,
+): Fiber {
+  const fiber = createFiberFromElementWithoutDebugInfo(
```
acdlite: Why the two functions? Doesn't look like this is called anywhere else

sebmarkbage: So that I can do early returns in the other one. Compiler can restructure and get rid of it in prod.

acdlite: I believe you

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -724,79 +719,94 @@ function mountIndeterminateComponent(
...
+  ));
+  startWorkTimer(workInProgress);
+  const resolvedProps = resolveDefaultProps(Component, props);
+  let child;
+  switch (resolvedTag) {
+    case FunctionComponent: {
+      child = updateFunctionComponent(
+        null,
+        workInProgress,
+        Component,
```
acdlite: Should we get rid of this argument now?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -724,79 +719,94 @@ function mountIndeterminateComponent(
...
+  const resolvedTag = (workInProgress.tag = resolveLazyComponentTag(
+    workInProgress,
+    Component,
+  ));
+  startWorkTimer(workInProgress);
+  const resolvedProps = resolveDefaultProps(Component, props);
+  let child;
+  switch (resolvedTag) {
+    case FunctionComponent: {
+      child = updateFunctionComponent(
```
acdlite: These can be early returns

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -694,8 +684,6 @@ function completeWork(
       break;
     }
     case ForwardRef:
-    case ForwardRefLazy:
-      break;
```
acdlite: Oops, shouldn't have deleted the break, because now it falls through to `SuspenseComponent`. I'll revert.

PR #13903, opened 2018-10-20
Allow arbitrary types to be wrapped in pure

sebmarkbage:
First I made the PureComponent a slow path. This creates an outer fiber that container the pure check and an inner fiber that represents which ever type of component.

Then I added an optimized fast path for simple pure function components. This is special cased when there are no defaultProps and it's a simple function component instead of class. This doesn't require an extra fiber.

We could make it so that this also works with custom comparer but that means we have to go through one extra indirection to get to it. Maybe it's worth it, donno.

### DIFF
--- a/packages/react/src/__tests__/forwardRef-test.js
+++ b/packages/react/src/__tests__/forwardRef-test.js
@@ -226,4 +226,131 @@ describe('forwardRef', () => {
         '    in Foo (at **)',
     );
   });
+
+  it('should not bailout if forwardRef is not wrapped in pure', () => {
+    const Component = props => <div {...props} />;
+
+    let renderCount = 0;
+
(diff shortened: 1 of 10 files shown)


packages/react/src/__tests__/forwardRef-test.js
```diff
@@ -226,4 +226,131 @@ describe('forwardRef', () => {
...
+
+    expect(ref.current).toBe(null);
+    expect(differentRef.current.type).toBe('div');
+
+    ReactNoop.render(<RefForwardingComponent ref={ref} optional="bar" />);
+    ReactNoop.flush();
+    expect(renderCount).toBe(3);
+  });
+
+  it('should custom pure comparisons to compose', () => {
```
acdlite: Missing a word?

PR #13904, opened 2018-10-20
 Make NoWork big instead of small

sophiebits:
Look at all the code we don't need!

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -243,11 +243,7 @@ function updatePureComponent(
 ) {
   const render = Component.render;
 
-  if (
-    current !== null &&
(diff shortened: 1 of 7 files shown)


sophiebits:
bigger?! I guess we inline 1073741823 a bunch of times?

sophiebits:
Probably could change Closure hereabouts if we wanted to *not* inline it every time. But maybe that's slow. https://github.com/google/closure-compiler/blob/85643a17293ac09253dcf8017404136d28942fc4/src/com/google/javascript/jscomp/InlineVariables.java#L725

sebmarkbage:
I think I would prefer the inverse and start the timer at the max int going backwards.

That speaks to the notion of higher “priority” being a higher number. We always end up having to say “higher, and by higher I mean lower number”.

That would leave NoWork being falsy which is a nice property. And it would solve the file size issue.

PR #13912, opened 2018-10-21
Flip expiration times

sophiebits:
Now:

0 - NoWork
1 - Never
...
big - low pri things
even bigger - hi pri things
max 31-bit int - Sync

Per @sebmarkbage in https://github.com/facebook/react/pull/13904.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -241,8 +241,8 @@ if (__DEV__) {
   };
 }
 
-// Used to ensure computeUniqueAsyncExpiration is monotonically increasing.
-let lastUniqueAsyncExpiration: number = 0;
+// Used to ensure computeUniqueAsyncExpiration is monotonically decreasing.
+let lastUniqueAsyncExpiration: number = Sync - 1;
 
 // Represents the expiration time that incoming updates should use. (If this
 // is NoWork, use the default strategy: async updates in async mode, sync
@@ -1550,11 +1550,11 @@ function computeThreadID(
 function computeUniqueAsyncExpiration(): ExpirationTime {
   const currentTime = requestCurrentTime();
   let result = computeAsyncExpiration(currentTime);
-  if (result <= lastUniqueAsyncExpiration) {
+  if (result >= lastUniqueAsyncExpiration) {
     // Since we assume the current time monotonically increases, we only hit
     // this branch when computeUniqueAsyncExpiration is fired multiple times
     // within a 200ms window (or whatever the async bucket size is).
-    result = lastUniqueAsyncExpiration + 1;
+    result = lastUniqueAsyncExpiration - 1;
   }
   lastUniqueAsyncExpiration = result;
   return lastUniqueAsyncExpiration;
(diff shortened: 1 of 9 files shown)


sophiebits:
Want me to make them all negative? 🙄

min 31-bit int - Sync
large negative - hi pri things
less-large negative - low pri things
...
–1 - Never
0 - NoWork

Then 0 is in the right place and it goes in the right direction…

sebmarkbage:
I think we can find a mental model where this makes sense. E.g. a count down until the app stops working (NoWork).

acdlite:
I’m cool with this if we rename the field, too. Like timeRemaining or something with “priority” in it.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -855,18 +840,10 @@ function resetChildExpirationTime(
...
         newChildExpirationTime = childUpdateExpirationTime;
       }
-      if (
-        newChildExpirationTime === NoWork ||
-        (childChildExpirationTime !== NoWork &&
-          childChildExpirationTime < newChildExpirationTime)
-      ) {
+      if (childChildExpirationTime > newChildExpirationTime) {
         newChildExpirationTime = childChildExpirationTime;
       }
```
sophiebits: At this point we could even go so far as to do

```
newChildExpirationTime = Math.max(
  newChildExpirationTime,
  child.expirationTime,
  child.childExpirationTime,
);
```

so simple… wow

doge meme

sebmarkbage: 😢

sophiebits:
I kinda like the negative clock (that becomes less negative over time) actually… what about you two?

sebmarkbage:
Nah. Negative is weird. That's not Rust/WASM/et al friendly since we'd want the unsigned form there.

acdlite:
If we didn’t already have something called “priority” I would say let’s rename it to that and merge your PR as is.

I don’t like the “expirationTime” naming regardless because it only acts like an expiration time when we schedule the callback with Scheduler. Everywhere else it acts like a relative priority.

Precdence? Urgency? Rank? (I’m just reading the thesaurus)

sophiebits:
"Threshold" maybe.

acdlite:
Importance?

sebmarkbage:
Note that we also plan to add another public priority API which is not related to time (the priority of fallbacks returning to their unsuspended state).

gaearon:
Urgency?

sophiebits:
We're hopeless.

gaearon:
Oh I missed @acdlite already suggested urgency.

acdlite:
@sebastian yeah and I also don’t think we understand the relationship between the two concepts yet, either.

sophiebits:
restlessness

NE-SmallTown:
> If we didn’t already have something called “priority” I would say let’s rename it to that and merge your PR as is.
> I don’t like the “expirationTime” naming regardless because it only acts like an expiration time when we schedule the callback with Scheduler. Everywhere else it acts like a relative priority.

Agree, we use priority naming before but then use expirationTime to abstract too many things for processing starvation and competition better. But just from a developer perspective, the expirationTime is confused and hard to understand(lower means higher, higher means lower), and when you compare it to some names like `xxxxWork`, it's weird(a time to compare a work, what's that, oh, they are just numbers...)

PR #13918, opened 2018-10-22
Updated comment for getEventModifierState

### LINKED ISSUE #13916: Update getModifierState browser support comments
We have a few stale comments about getEventModifierState:

https://github.com/facebook/react/blob/663835a43a5a92ff63c3e3b43e4c9ff40dde173b/packages/react-dom/src/events/getEventModifierState.js#L24-L26

IE8 isn't an issue anymore, [but Safari probably still is](https://caniuse.com/#feat=keyboardevent-getmodifierstate). It'd be nice to figure out specifically why we still need this module and update the comment with that info.

powens:
Updating the comments for the `getEventModifierState` function; Changing the legacy browser support to mention Safari instead of IE8, and removing the note about Chrome not supporting Lock-Keys.

Closes #13916

### DIFF
--- a/packages/react-dom/src/events/getEventModifierState.js
+++ b/packages/react-dom/src/events/getEventModifierState.js
@@ -21,9 +21,9 @@ const modifierKeyToProp = {
   Shift: 'shiftKey',
 };
 
-// IE8 does not implement getModifierState so we simply map it to the only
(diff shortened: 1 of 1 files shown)


nhunzaker:
Hey! Sorry for the delay. This is great, thank you!

PR #13921, opened 2018-10-23
Try rendering again if a timed out tree receives an update

acdlite:
Found a bug related to suspending inside an already mounted tree. While investigating this I noticed we really don't have much coverage of suspended updates — nearly all are testing the initial mount. I think this would greatly benefit from some fuzz testing; still haven't thought of a good test case, though.

### DIFF
--- a/packages/jest-react/src/JestReact.js
+++ b/packages/jest-react/src/JestReact.js
@@ -35,19 +35,17 @@ function assertYieldsWereCleared(root) {
 }
 
 export function toFlushAndYield(root, expectedYields) {
+  assertYieldsWereCleared(root);
+  const actualYields = root.unstable_flushAll();
   return captureAssertion(() => {
-    assertYieldsWereCleared(root);
-    const actualYields = root.unstable_flushAll();
     expect(actualYields).toEqual(expectedYields);
   });
 }
 
 export function toFlushAndYieldThrough(root, expectedYields) {
+  assertYieldsWereCleared(root);
+  const actualYields = root.unstable_flushNumberOfYields(expectedYields.length);
   return captureAssertion(() => {
-    assertYieldsWereCleared(root);
-    const actualYields = root.unstable_flushNumberOfYields(
-      expectedYields.length,
-    );
     expect(actualYields).toEqual(expectedYields);
   });
 }
@@ -76,8 +74,8 @@ export function toHaveYielded(ReactTestRenderer, expectedYields) {
 }
 
 export function toFlushAndThrow(root, ...rest) {
+  assertYieldsWereCleared(root);
   return captureAssertion(() => {
-    assertYieldsWereCleared(root);
     expect(() => {
       root.unstable_flushAll();
     }).toThrow(...rest);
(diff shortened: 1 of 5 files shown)


packages/jest-react/src/JestReact.js
```diff
@@ -35,19 +35,17 @@ function assertYieldsWereCleared(root) {
 }
 
 export function toFlushAndYield(root, expectedYields) {
+  assertYieldsWereCleared(root);
```
acdlite: I moved these outside because `captureAssertion` was swallowing the stack trace.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1127,8 +1124,8 @@ function updateSuspenseComponent(
         // its fragment. We're going to skip over these entirely.
         const nextFallbackChildren = nextProps.fallback;
         const primaryChildFragment = createWorkInProgress(
-          currentFallbackChildFragment,
-          currentFallbackChildFragment.pendingProps,
+          currentPrimaryChildFragment,
+          currentPrimaryChildFragment.pendingProps,
```
acdlite: This is a fix for an oopsies that could have been its own PR, but it was required to get the new test I wrote to pass. I can split it out if I need to.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1481,23 +1479,45 @@ function beginWork(
...
+            const primaryChildFragment: Fiber = (workInProgress.child: any);
+            const primaryChildExpirationTime =
+              primaryChildFragment.childExpirationTime;
+            if (
+              primaryChildExpirationTime !== NoWork &&
+              primaryChildExpirationTime <= renderExpirationTime
+            ) {
+              // The primary children have pending work. Use the normal path
+              // to attempt to render the primary children again.
+              return updateSuspenseComponent(
```
acdlite: This is the main fix here.

packages/react-reconciler/src/ReactFiberSuspenseComponent.js
```diff
@@ -23,3 +24,17 @@ export type SuspenseState = {|
   // `didTimeout` because it's not set unless the boundary actually commits.
   timedOutAt: ExpirationTime,
 |};
+
+export function shouldCaptureSuspense(
```
acdlite: I split this out because it was getting hard to follow all the nested branches in the `throwException` function

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -407,9 +406,10 @@ function unwindWork(
         // Captured a suspense effect. Set the boundary's `alreadyCaptured`
         // state to true so we know to render the fallback.
         const current = workInProgress.alternate;
-        const currentState = current !== null ? current.memoizedState : null;
-        let nextState = workInProgress.memoizedState;
-        if (currentState === null) {
+        const currentState: SuspenseState | null =
+          current !== null ? current.memoizedState : null;
```
acdlite: I believe this will fix the issue in www, haven't confirmed yet. I'll try on my sandbox shortly.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1437,6 +1434,7 @@ function beginWork(
 ): Fiber | null {
   const updateExpirationTime = workInProgress.expirationTime;
 
+  // console.log(workInProgress.tag);
```
gaearon: oops

acdlite:
@sebmarkbage Hmm I remember hitting this issue with context and shrugging it off, not state though. (Oh, and context should just work now that the timed out children remain in the tree. Should add a test for that as a follow up.)

PR #13922, opened 2018-10-23
Remove unstable_ prefix from Suspense

sebmarkbage:
We are using it with lazy and the combination Suspense + lazy seems pretty stable. maxDuration is not but that's only enabled when you're in ConcurrentMode which is still unstable.

Don't forget to fix www.

### DIFF
--- a/fixtures/dom/src/components/fixtures/suspense/index.js
+++ b/fixtures/dom/src/components/fixtures/suspense/index.js
@@ -5,7 +5,7 @@ import TestCase from '../../TestCase';
 const React = window.React;
 const ReactDOM = window.ReactDOM;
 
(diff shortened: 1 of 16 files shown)


sebmarkbage:
Although I guess https://github.com/facebook/react/pull/13921 is a counter argument. :)

gaearon:
Would be nice to follow up to understand how it landed while breaking that test.

sophiebits:
Should maxDuration be prefixed?

sebmarkbage:
I think it's fine to leave `maxDuration`. You have to type `unstable_` somewhere to use it. We're going to have to add types for this API and changing that after the fact, especially with the poor versioning story there, it'll be a pain.

PR #13923, opened 2018-10-23
Fix lint

sebmarkbage:
oops


gaearon:
You can also remove `-internal` now

PR #13925, opened 2018-10-23
[Synchronous Suspense] Unmounting a suspended class component

acdlite:
Previously we assumed any mounted class component would have an instance, but that's not the case for classes that suspend outside of Concurrent mode.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -496,6 +496,9 @@ function commitUnmount(current: Fiber): void {
       safelyDetachRef(current);
       const instance = current.stateNode;
       if (
+        // Typically, a component that mounted will have an instance. However,
+        // outside of concurrent mode, a suspended component may commit without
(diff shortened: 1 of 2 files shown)


acdlite:
@gaearon Probably not, can't be sure. This category of bug related to committing fibers that didn't actually complete — it's an intentional deviation from our model, so I wouldn't be shocked if we found more.

sebmarkbage:
This doesn't make sense to me. I thought they could all go to indeterminate mode. I don't want to go hunt down adding special cases in the code that we'll have to remove again.

acdlite:
@sebmarkbage Classes don't use the indeterminate tag, except module-style components. That's probably a better way to fix it, though — a special tag that indicates the fiber is incomplete.

Are you ok with landing this in the meantime to unblock the release?

sebmarkbage:
I'm worried about this not being the last one.

acdlite:
Ok I'll open a different PR that adds a new type of work, `IncompleteComponent`

PR #13926, opened 2018-10-23
[Synchronous Suspense] Suspending a class outside concurrent mode

acdlite:
When a class component suspends during mount outside concurrent mode, change the tag so it's not mistaken for a completed component. For example, we should not call componentWillUnmount if it is deleted.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -33,7 +33,7 @@ import {
   MemoComponent,
   SimpleMemoComponent,
   LazyComponent,
-  IncompleteComponent,
+  IncompleteClassComponent,
 } from 'shared/ReactWorkTags';
 import {
   NoEffect,
@@ -841,9 +841,11 @@ function mountLazyComponent(
   return child;
 }
 
-function mountIncompleteComponent(
+function mountIncompleteClassComponent(
   _current,
   workInProgress,
+  Component,
+  nextProps,
   renderExpirationTime,
 ) {
   if (_current !== null) {
@@ -859,12 +861,6 @@ function mountIncompleteComponent(
 
   // Promote the fiber to a class and try rendering again.
   workInProgress.tag = ClassComponent;
-  const Component = workInProgress.type;
-  const unresolvedProps = workInProgress.pendingProps;
-  const resolvedProps =
-    workInProgress.elementType === Component
-      ? unresolvedProps
-      : resolveDefaultProps(Component, unresolvedProps);
 
   // The rest of this function is a fork of `updateClassComponent`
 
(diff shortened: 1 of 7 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -840,6 +841,68 @@ function mountLazyComponent(
...
+    // Since this is conceptually a new fiber, schedule a Placement effect
+    workInProgress.effectTag |= Placement;
+  }
+
+  // Promote the fiber to a class and try rendering again.
+  workInProgress.tag = ClassComponent;
+  const Component = workInProgress.type;
+  const unresolvedProps = workInProgress.pendingProps;
+  const resolvedProps =
+    workInProgress.elementType === Component
```
sebmarkbage: Any reason this is here instead of beginWork like the others?

acdlite:
@sebmarkbage Re: naming, I thought maybe this might be needed for other types in the future? But I suppose we'll rename it then.

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: @sebmarkbage Nah, I'll change

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -495,7 +499,13 @@ function commitUnmount(current: Fiber): void {
     case ClassComponent: {
       safelyDetachRef(current);
       const instance = current.stateNode;
-      if (typeof instance.componentWillUnmount === 'function') {
+      if (
+        // Typically, a component that mounted will have an instance. However,
+        // outside of concurrent mode, a suspended component may commit without
+        // an instance, so we need to check whether it exists.
+        instance !== null &&
```
gaearon: So why do we still need this check? I thought the separate tag is supposed to address it.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -762,7 +763,7 @@ function mountLazyComponent(
   renderExpirationTime,
 ) {
   if (_current !== null) {
-    // An indeterminate component only mounts if it suspended inside a non-
+    // An lazy component only mounts if it suspended inside a non-
```
aweary: A* lazy

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -840,6 +841,64 @@ function mountLazyComponent(
...
+function mountIncompleteClassComponent(
+  _current,
+  workInProgress,
+  Component,
+  nextProps,
+  renderExpirationTime,
+) {
+  if (_current !== null) {
+    // An incomplete component only mounts if it suspended inside a non-
+    // concurrent tree, in an inconsistent state. We want to tree it like
```
aweary: > We want to tree it

treat it?

PR #13928, opened 2018-10-23
Enable stable concurrent APIs flag for 16.7 alpha

bvaughn:
These APIs will no longer be "unstable" in the upcoming 16.7 release. Rename them behind a feature flag though, in case we need to publish a 16.6 minor release in the meanwhile.


sebmarkbage:
Let's land this after 16.6 is out.

bvaughn:
Yes of course. 😄 Just getting it ready.

bvaughn:
Ready to go now I guess 😄

PR #13930, opened 2018-10-23
Removed react-cache from the bundle list for now

bvaughn:
We don't want to release this with 16.6


bvaughn:
Fresh `yarn build && yarn test-prod` works for me locally FWIW.

PR #13931, opened 2018-10-23
Revert accidentally committed existence check

acdlite:
I originally added this check in #13925, which I later closed in favor of #13926, but I neglected to remove it.


sebmarkbage:
Is this synced?

PR #13935, opened 2018-10-24
react-dom: Update README.md

marcelometal:
Replaced 'var' with 'const'


gaearon:
Thanks but people reading this might be using ES5 too. I don't really see the point of this, especially for a commonly used package like React itself.

PR #13959, opened 2018-10-24
[react-is] add back proper AsyncMode symbol, for back compat

### LINKED ISSUE #13958: react-is breaking change, re AsyncMode
It seems like react-is v16.6 no longer is compatible with React v16.3 - v16.5, in that React v16.3 - v16.5 will use `Symbol(react.async_mode)`, but `react-is`'s `AsyncMode` export will instead be `Symbol(react.concurrent_mode)`.

It seems like the async mode symbol needs to remain the same in react-is 16.6+, and the `isAsyncMode` function would need to check *both* the AsyncMode and ConcurrentMode symbols.

ljharb:
- Partial revert of #13732
 - Fixes #13958.

I didn't add tests for `isAsyncMode` because #13732 didn't preserve any - happy to add them if needed.

### DIFF
--- a/packages/react-is/src/ReactIs.js
+++ b/packages/react-is/src/ReactIs.js
@@ -10,6 +10,7 @@
 'use strict';
 
 import {
+  REACT_ASYNC_MODE_TYPE,
   REACT_CONCURRENT_MODE_TYPE,
   REACT_CONTEXT_TYPE,
   REACT_ELEMENT_TYPE,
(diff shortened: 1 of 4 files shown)


ljharb:
I have confirmed that these changes cause my enzyme tests in react 16.3 to pass.

gaearon:
Seems okay. We shouldn't have exposed this without the prefix but since we did...

Issue #13960, opened 2018-10-25
React.lazy: defaultProps not merged into props in commit phase lifecycles


variadicintegrity:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
In a component loaded using React.lazy, default props are not merged into the props object for these lifecycles:
- componentDidMount
- componentDidUpdate
- componentWillUnmount
- getSnapshotBeforeUpdate

In shouldComponentUpdate, default props are merged into the nextProps argument, but not this.props.

Steps to reproduce:
1. Create a component that has a default value for a prop using defaultProps.
2. Use that prop in one of the above lifecycles.
3. Load the component using React.lazy.
4. Observe that the prop is not defined in the above lifecycles.

[CodeSandbox Link](https://codesandbox.io/s/vym3zl8063)

**What is the expected behavior?**
Default props should be merged into props for all lifecycles when using React.lazy.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
react 16.6.0
react-dom 16.6.0

Windows 10 Firefox 63.0
Windows 10 Chrome 70.0.3538.77

React.lazy was not in previous versions of React

aweary:
Thanks for the detailed report, this definitely looks like a bug.

PR #13961, opened 2018-10-25
Resolve defaultProps for Lazy components

aweary:
A (still) failing test and naive "fix" for https://github.com/facebook/react/issues/13960. I want to dig into this more, but hopefully I can get some input on the right approach for this.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -782,7 +782,10 @@ function mountLazyComponent(
   workInProgress.type = Component;
   const resolvedTag = (workInProgress.tag = resolveLazyComponentTag(Component));
   startWorkTimer(workInProgress);
-  const resolvedProps = resolveDefaultProps(Component, props);
+  const resolvedProps = (workInProgress.pendingProps = resolveDefaultProps(
+    Component,
+    props,
+  ));
   let child;
   switch (resolvedTag) {
     case FunctionComponent: {
--- a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
@@ -379,7 +379,7 @@ describe('ReactLazy', () => {
       'shouldComponentUpdate (this.props): A',
       'shouldComponentUpdate (nextProps): A',
       'A2', // render
-      'getSnapshotBeforeUpdate (this.props): undefined',
+      'getSnapshotBeforeUpdate (this.props): A',
       'getSnapshotBeforeUpdate (prevProps): A',
       'componentDidUpdate: A',
     ]);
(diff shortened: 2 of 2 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -782,7 +782,10 @@ function mountLazyComponent(
   workInProgress.type = Component;
   const resolvedTag = (workInProgress.tag = resolveLazyComponentTag(Component));
   startWorkTimer(workInProgress);
-  const resolvedProps = resolveDefaultProps(Component, props);
+  const resolvedProps = (workInProgress.pendingProps = resolveDefaultProps(
+    Component,
+    props,
```
aweary: The type and tag are overwritten, so my first thought was that the same should be done for `pendingProps`

aweary:
This solves part of the issue, but the second issue is that we rely on the correct props being resolved in `React.createElement`. Seems like there will need to be a check for lazy elements so that when a Fiber is created it's created with the resolved default props as well.

gaearon:
>but the second issue is that we rely on the correct props being resolved in React.createElement

When you say "we rely" what do you mean exactly? Rely in what way?

gaearon:
I'll continue this in https://github.com/facebook/react/pull/14108

aweary:
> When you say "we rely" what do you mean exactly? Rely in what way?

Default props are resolved in `React.createElement`. It's assumed that the element will have the resolved props at that point.

> I don't understand how this could be possible. If the component is lazy, we don't have its default props at first, do we? We're waiting for it to load.

The initial load happens inside `mountLazyComponent`. At that point the default props aren't available, but it suspends and retries and eventually it calls `resolveDefaultProps`.  It passes those resolved props to the corresponding `update{Type}Component` which is why the props are correct in the constructor and other lifecycle methods called at that point.

https://github.com/facebook/react/blob/d5d10d140ed8739c83f3cee5d43f0bf60eea1002/packages/react-reconciler/src/ReactFiberBeginWork.js#L769-L774

It was still broken for the initial commit phase, as props get red from the workInProgress. Setting `pendingProps` to the resolved props fixes that part of the issue.

Even then, it's still broken for updates. In `React.createElement` the lazy component has been fully resolved, but the default props are not parsed because it checks `type.defaultProps` and `type` is not the resolved component.  When the workInProgress is created from the element it uses `element.props`.

My conclusion was that, at some point in the update path, we'd need to check the `type` and resolve the default props. I was trying to find the right place for this so that non-lazy components didn't incur the cost of calling `resolveDefaultProps` as well, but it wasn't clear where that would be for updates.

gaearon:
>but the default props are not parsed because it checks type.defaultProps and type is not the resolved component

Would `fiber.elementType` work then?

gaearon:
(Feel free to push to my branch in https://github.com/facebook/react/pull/14108)

aweary:
> Would `fiber.elementType` work then?

We could check `fiber.elementType` before calling `createWorkInProgress` and then call `resolveDefaultProps` with `pendingProps` if it's a lazy component, but I was concerned about adding a property check there since it's a hot path for all element types.

PR #13968, opened 2018-10-25
Implement the experimental Hooks proposal behind a feature flag

sebmarkbage:
*Hooks* let you use all React features (such as state and lifecycle) without writing a class.

* **[Learn more about Hooks](https://reactjs.org/docs/hooks-intro.html)**
* **[Provide feedback on the Hooks RFC](https://github.com/reactjs/rfcs/pull/68)**

We have presented the Hooks proposal at React Conf 2018. This pull request implements all proposed built-in Hooks behind a feature flag, both for the client and the server renderer.

This is a very early time for Hooks and this proposal is experimental. We're merging it into master now so we can iterate on it. It is **turned on only in the 16.7 alpha** release so that you can try it and provide your feedback.

We have been using this code for a month in production so while the APIs are not stable, we don't expect the implementation to contain major bugs.

**Please keep the feedback and discussion contained [in the Hooks RFC](https://github.com/reactjs/rfcs/pull/68) (and _not_ in this PR) so that everyone can discuss it in a single place without checking multiple GitHub threads.**

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRendererHooks.js
+++ b/packages/react-dom/src/server/ReactPartialRendererHooks.js
@@ -365,8 +365,8 @@ export const Dispatcher = {
   useState,
   useMutationEffect,
   useLayoutEffect,
-  // useAPI is not run in the server environment
-  useAPI: noop,
+  // useImperativeMethods is not run in the server environment
+  useImperativeMethods: noop,
   // Callbacks are not run in the server environment.
   useCallback: noop,
   // Effects are not run in the server environment.
--- a/packages/react-dom/src/server/ReactPartialRendererHooks.js
+++ b/packages/react-dom/src/server/ReactPartialRendererHooks.js
@@ -50,7 +50,7 @@ const RE_RENDER_LIMIT = 25;
 function resolveCurrentlyRenderingComponent(): Object {
   invariant(
     currentlyRenderingComponent !== null,
-    'Hooks can only be called inside the body of a functional component.',
+    'Hooks can only be called inside the body of a function component.',
   );
   return currentlyRenderingComponent;
 }
@@ -105,7 +105,7 @@ export function finishHooks(
   children: any,
   refOrContext: any,
 ): any {
-  // This must be called after every functional component to prevent hooks from
+  // This must be called after every function component to prevent hooks from
   // being used in classes.
 
   while (didScheduleRenderPhaseUpdate) {
@@ -321,7 +321,7 @@ function dispatchAction<S, A>(
       lastRenderPhaseUpdate.next = update;
     }
   } else {
-    // This means an update has happened after the functional component has
+    // This means an update has happened after the function component has
     // returned. On the server this is a no-op. In React Fiber, the update
     // would be scheduled for a future render.
   }
--- a/packages/react-dom/src/server/ReactPartialRendererHooks.js
+++ b/packages/react-dom/src/server/ReactPartialRendererHooks.js
@@ -11,7 +11,7 @@ import type {ReactContext} from 'shared/ReactTypes';
 import invariant from 'shared/invariant';
 import warning from 'shared/warning';
 
-type BasicStateAction<S> = S | (S => S);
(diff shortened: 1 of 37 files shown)


NE-SmallTown:
> committed on 6 Sep

Oh, the last time you do this is the *Suspense*. 😄 

Great work !

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
```diff
@@ -0,0 +1,1960 @@
...
+        'Layout',
+        'Layout effect 1',
+      ]);
+      expect(ReactNoop.getChildren()).toEqual([
+        span('Passive'),
+        span('Layout'),
+      ]);
+    });
+
+    it('flushes passive effects even if siblings schedule a new root', () => {
```
philipp-spiess: I wonder how you come up with all of those edge cases. 😉

packages/react/src/ReactHooks.js
```diff
@@ -0,0 +1,120 @@
...
+    if ((Context: any)._context !== undefined) {
+      const realContext = (Context: any)._context;
+      // Don't deduplicate because this legitimately causes bugs
+      // and nobody should be using this in existing code.
+      if (realContext.Consumer === Context) {
+        warning(
+          false,
+          'Calling useContext(Context.Consumer) is not supported, may cause bugs, and will be ' +
+            'removed in a future major release. Did you mean to call useContext(Context) instead?',
+        );
```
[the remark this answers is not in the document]
trueadm: There’s no reason to use that pattern anymore with hooks.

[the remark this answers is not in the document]
Andarist: @trueadm actually this could be useful for libraries wanting to hide their providers 🤔

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
gaearon: Ran into bugs in product code.

packages/react/src/ReactHooks.js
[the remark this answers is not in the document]
gaearon: It's plausible one of returned values would be a Hook itself in the future.

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -0,0 +1,374 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+import type {ReactContext} from 'shared/ReactTypes';
+
+import invariant from 'shared/invariant';
+import warning from 'shared/warning';
+
+type BasicStateAction<S> = S | (S => S);
```
TrySound: This is more sound I think. In current case flow will always refine any value and function params and return value won't be checked. I had the same problem in powerplug.
```
type BasicStateAction<S> = (S => S) | S;
```

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -0,0 +1,374 @@
...
+  resolveCurrentlyRenderingComponent();
+  return context._currentValue;
+}
+
+function basicStateReducer<S>(state: S, action: BasicStateAction<S>): S {
+  return typeof action === 'function' ? action(state) : action;
+}
+
+export function useState<S>(
+  initialState: S | (() => S),
```
TrySound: Here the same problem. `(() => S)` is unreachable.

packages/react-dom/src/server/ReactPartialRendererHooks.js
acdlite: Yeah we found the same thing when testing at FB. Forgot to update in the upstream brach. :)

acdlite:
I'm merging this PR now so that we can continue to iterate on changes to Hooks in separate PRs. The feature flag for Hooks will remain disabled (so they won't be accessible in releases) until we come to a decision on the RFC.

PR #13974, opened 2018-10-25
Fixed a random bug in react-dom development

jld3103:
I found a random bug in react-dom, but only in development:
```
ERROR in ./node_modules/react-dom/cjs/react-dom.development.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
SyntaxError: node_modules/react-dom/cjs/react-dom.development.js: Unexpected token, expected ";" (13463:228)
```

A string was not escaped correctly, but only in development it throws this error.


jld3103:
The checks failed, but it wasn't my changed code.
Look at the bottom of the output and then please rerun the checks.

aweary:
It's using double quotes, so the single quotes don't need to be escaped. The syntax is valid, so the error you're getting has to be a result of some other issue. It seems like it was thrown by `babel-loader` which suggests you have an issue with your build setup.

Check our community support resources for help on that:
https://facebook.github.io/react/community/support.html

PR #13977, opened 2018-10-26
Add space before parentheses for coding consistency

nadim1992:


nhunzaker:
Thank you for taking the time to send this in, however we try to avoid merging to stylistic changes that do not also involve bug fixes. It makes it easier for us to track the history of updates and figure out when code last changed.

PR #14003, opened 2018-10-27
chore(tests): don't rely on jest fake timers scheduling real timers

SimenB:
I'm currently working on replacing Jest's custom fake timers implementation with Lolex (https://github.com/facebook/jest/pull/5171). As part of that, I wanted to test that React's tests still work. They didn't...

A quirk of Jest's current implementation is that we schedule `process.nextTick` and `setImmediate` on the real timers, even though they're faked. Then we just make sure to do nothing if Jest invoked the function when advancing time, or vice versa if Node triggered it before us. That behavior is intentionally _not_ re-implemented in the new version.

This PR changes from using `setImmediate` to `Promise.resolve().then()` to "wait 1 tick" which behaves the same.

### DIFF
--- a/packages/react/src/__tests__/ReactProfilerDOM-test.internal.js
+++ b/packages/react/src/__tests__/ReactProfilerDOM-test.internal.js
@@ -18,15 +18,6 @@ let ReactCache;
 function initEnvForAsyncTesting() {
   // Boilerplate copied from ReactDOMRoot-test
   // TODO pull this into helper method, reduce repetition.
-  const originalDateNow = Date.now;
-  global.Date.now = function() {
-    return originalDateNow();
-  };
-  global.requestAnimationFrame = function(cb) {
-    return setTimeout(() => {
-      cb(Date.now());
-    });
-  };
   const originalAddEventListener = global.addEventListener;
   let postMessageCallback;
   global.addEventListener = function(eventName, callback, useCapture) {
@@ -140,7 +131,7 @@ describe('ProfilerDOM', () => {
 
               // Evaluate in an unwrapped callback,
               // Because trace/wrap won't decrement the count within the wrapped callback.
-              setImmediate(() => {
+              Promise.resolve().then(() => {
                 expect(onInteractionTraced).toHaveBeenCalledTimes(1);
                 expect(
                   onInteractionScheduledWorkCompleted,
--- a/packages/react/src/__tests__/ReactProfilerDOM-test.internal.js
+++ b/packages/react/src/__tests__/ReactProfilerDOM-test.internal.js
@@ -18,6 +18,12 @@ let ReactCache;
 function initEnvForAsyncTesting() {
   // Boilerplate copied from ReactDOMRoot-test
   // TODO pull this into helper method, reduce repetition.
+  // TODO remove `requestAnimationFrame` when upgrading to Jest 24 with Lolex
+  global.requestAnimationFrame = function(cb) {
+    return setTimeout(() => {
+      cb(Date.now());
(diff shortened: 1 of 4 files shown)


packages/react/src/__tests__/ReactProfilerDOM-test.internal.js
```diff
@@ -18,10 +18,7 @@ let ReactCache;
 function initEnvForAsyncTesting() {
   // Boilerplate copied from ReactDOMRoot-test
   // TODO pull this into helper method, reduce repetition.
-  const originalDateNow = Date.now;
-  global.Date.now = function() {
-    return originalDateNow();
-  };
+  // TODO remove `requestAnimationFrame` when upgrading to Jest 24 with Lolex
```
SimenB: Using Jest 24, the test passes without faking `global.requestAnimationFrame`. Not sure if it's due to Lolex or just a newer version of JSDOM

packages/react/src/__tests__/ReactProfilerDOM-test.internal.js
```diff
@@ -18,10 +18,7 @@ let ReactCache;
 function initEnvForAsyncTesting() {
   // Boilerplate copied from ReactDOMRoot-test
   // TODO pull this into helper method, reduce repetition.
-  const originalDateNow = Date.now;
-  global.Date.now = function() {
```
SimenB: Lolex will fake this.

The test also passed when removing it when using Jest 23, not sure what it was supposed to do?

nhunzaker: This is curious! I'm not sure what this was for. I'm guessing that this was just an artifact copy/pasting a snippet from ReactDOMRoot-test. It's quite a bit more sophisticated:

https://github.com/facebook/react/blob/c84b9bf828c38e0b24042ebb9cd859292146a6fb/packages/react-dom/src/__tests__/ReactDOMRoot-test.js#L23-L45

I think this is safe to remove.

packages/react/src/__tests__/ReactProfilerDOM-test.internal.js
```diff
@@ -140,7 +137,7 @@ describe('ProfilerDOM', () => {
 
               // Evaluate in an unwrapped callback,
               // Because trace/wrap won't decrement the count within the wrapped callback.
-              setImmediate(() => {
+              Promise.resolve().then(() => {
```
nhunzaker: I kind of wish this could be re-used, but it's unclear where something like that could go. .

SimenB:
Seeing as the tests still pass, there's no reason not to merge, imo :)
Timeline is fuzzy, personally I hope for e.g. December 1st

packages/react/src/__tests__/ReactProfilerDOM-test.internal.js
SimenB: Could do an `await`, then you're guaranteed a tick even if promises are faked

nhunzaker: Naw, I think this is fine.

nhunzaker:
Sure. Let's run with it. Thanks!

acdlite:
@nhunzaker In the future, please at least tag the person who wrote the code that's being changed. Brian or me, in this case. Unless it's to unbreak master. Sometimes even small changes like this could have unforeseen consequences.

nhunzaker:
Sorry 😥. Sure thing!

SimenB:
Did it break anything, or is that just a precaution?

PR #14023, opened 2018-10-29
Add command to rollup to sync React Native renderer in OSS

empyrical:
This pull request adds a new `--sync-opensource` command to the rollup build script.

This differs from `--sync-fbsource`, because it tries to sync to this path:

    {PROVIDED PATH}/xplat/js/react-native-github/Libraries/Renderer/

While `--sync-opensource` will sync to:

    {PROVIDED PATH}/Libraries/Renderer/

This will make it easier for OSS users to update the React Native renderer to an alpha version of React for testing/tinkering. (Not recommended to send the RN repo bug reports as a result of doing this, however!)

Usage is like this:

```bash
> yarn build --sync-opensource=/path/to/react-native --type=RN_OSS
```

### DIFF
--- a/scripts/rollup/build.js
+++ b/scripts/rollup/build.js
@@ -61,6 +61,7 @@ const requestedBundleNames = (argv._[0] || '')
   .map(type => type.toLowerCase());
 const forcePrettyOutput = argv.pretty;
 const syncFBSourcePath = argv['sync-fbsource'];
+const syncOpenSourcePath = argv['sync-opensource'];
 const syncWWWPath = argv['sync-www'];
 const shouldExtractErrors = argv['extract-errors'];
 const errorCodeOpts = {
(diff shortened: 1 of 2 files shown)


empyrical:
Thanks for pointing that out. Updated the example!

bvaughn:
Is this change really necessary? Seems like this should suffice?
```sh
$ yarn build -- --type=RN_OSS

# The open source RN build will now be available at "./build/react-native/"
# You can "cp -r ./build/react-native /whatever/path/you/want"
```

empyrical:
That's true; I guess that this is a bit redundant. Since copying manually from the `build/react-native` folder seems to work just as good I may close this

empyrical:
Yeah, I think that's the better way; closing this

bvaughn:
No worries! 😄 Thanks!

PR #14024, opened 2018-10-29
Warn about variable number of dependencies

acdlite:
We don't check this in prod, since best practice is to always pass
these inline. But we should still warn in dev.

I also moved all the existing Hook tests to a separate file called ReactHooksWithNoopRenderer. New tests should use test renderer.

Second commit is the relevant one.


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -760,8 +761,19 @@ function dispatchAction<S, A>(
 }
 
 function inputsAreEqual(arr1, arr2) {
-  // Don't bother comparing lengths because these arrays are always
+  // Don't bother comparing lengths in prod because these arrays should be
   // passed inline.
+  if (__DEV__) {
+    warning(
+      arr1.length === arr2.length,
+      'Detected a variable number of dependencies. The length of the ' +
```
bvaughn: Seems like this error message should mention "hooks". It's a little cryptic.

acdlite: Perhaps "variable number of dependencies to a hook"?

acdlite: Updated

bvaughn: Yeah, I like that change.

acdlite:
Going to merge without squashing to preserve the history

PR #14025, opened 2018-10-30
[scheduler] Deadline object -> shouldYield

acdlite:
Instead of using a requestIdleCallback-style deadline object, expose a method Scheduler.shouldYield that returns true if there's a higher priority event in the queue.

### DIFF
--- a/packages/react-noop-renderer/src/createReactNoop.js
+++ b/packages/react-noop-renderer/src/createReactNoop.js
@@ -271,7 +271,7 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
           scheduledCallbackTimeout === -1 ||
           scheduledCallbackTimeout > newTimeout
         ) {
-          scheduledCallbackTimeout = newTimeout;
+          scheduledCallbackTimeout = elapsedTimeInMs + newTimeout;
         }
       }
       return 0;
@@ -284,6 +284,8 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
       scheduledCallback = null;
       scheduledCallbackTimeout = -1;
     },
+
+    shouldYield,
 
     scheduleTimeout: setTimeout,
     cancelTimeout: clearTimeout,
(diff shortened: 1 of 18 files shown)


packages/react-noop-renderer/src/createReactNoop.js
```diff
@@ -407,35 +409,44 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
 
   let yieldedValues = null;
 
+  let didYield;
   let unitsRemaining;
 
+  function shouldYield() {
+    if (
+      scheduledCallbackTimeout === -1 ||
+      elapsedTimeInMs > scheduledCallbackTimeout
```
bvaughn: ~~It's early and I'm probably being stupid, but shouldn't this be `elapsedTimeInMs < scheduledCallbackTimeout` ?~~

**Edit** Never mind. This is just saying that we should never yield if work has already timed out.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -2077,38 +2062,55 @@ function findHighestPriorityRoot() {
...
+  if (shouldYield()) {
+    didYield = true;
+    return true;
+  }
+  return false;
+}
+
+function performAsyncWork() {
+  try {
+    if (!shouldYieldToRenderer()) {
```
bvaughn: It's not initially obvious to me that `dl.didTimeout` is semantically equivalent to `shouldYieldToRenderer()` but I trust you here.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -2123,12 +2125,12 @@ function performWork(minExpirationTime: ExpirationTime, dl: Deadline | null) {
...
       (minExpirationTime === NoWork ||
         minExpirationTime >= nextFlushedExpirationTime) &&
-      (!deadlineDidExpire || currentRendererTime >= nextFlushedExpirationTime)
+      (!didYield || currentRendererTime >= nextFlushedExpirationTime)
     ) {
       performWorkOnRoot(
         nextFlushedRoot,
         nextFlushedExpirationTime,
-        currentRendererTime >= nextFlushedExpirationTime,
+        !(currentRendererTime >= nextFlushedExpirationTime),
```
bvaughn: Uh, why `! >=` rather than `<`?

packages/react-test-renderer/src/ReactTestRendererScheduling.js
```diff
@@ -31,52 +32,47 @@ export function setNowImplementation(implementation: () => number): void {
   nowImplementation = implementation;
 }
 
+export function shouldYield() {
+  if (
+    expectedNumberOfYields !== -1 &&
+    yieldedValues.length >= expectedNumberOfYields
+  ) {
+    // We at least as many values as expected. Stop rendering.
```
bvaughn: "We yielded at least as many"?

packages/react-reconciler/src/ReactFiberScheduler.js
acdlite: Planned to change it once I proved it worked, then forgot :D

PR #14028, opened 2018-10-30
auto assignment with loop statements

sunnibfgi:
Inject all hooks into the React


nhunzaker:
Thank you for taking the time to send this in, however we try to avoid merging to stylistic changes that do not also involve bug fixes. Particularly for fast-moving features, it makes it easier for us to track the history of updates and figure out when code last changed. 

Sorry :(

PR #14033, opened 2018-10-30
Show deprecated context object warnings usage in ReactDOM server

trueadm:
This is a follow up to https://github.com/facebook/react/pull/13829.

This PR adds deprecated context object usage warnings that mirror the same warnings as in  https://github.com/facebook/react/pull/13829. Specifically:

## dev when using consumer: 

- Before: When using `Context.Consumer`, it's the same as `Context` so nothing special is handled.

- After: When `Context.Consumer` is used the at the beginning of reconciliation, the fiber is given the `_context` field from the `Consumer`, that references to `Context`.

## dev when using context:

- Before: When using `Context`, it's the same as the previous case (but the wrong behaviour).

- After: When `Context` we check if `_context` field exists on it (which we added to the new proxy object) and because it won't (only Context.Consumer does) we trigger a new warning.

## prod when using consumer:

- Before & After: When using `Context.Consumer`, it's the same as `Context` so nothing special is handled.

## prod when using context:

- Before & After: When using `Context`, it's the same as the previous case so nothing special is handled.

This behaviour is backwards compatible with previous versions of React.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationNewContext-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationNewContext-test.js
@@ -249,5 +249,104 @@ describe('ReactDOMServerIntegration', () => {
       expect(e.querySelector('#language2').textContent).toBe('sanskrit');
       expect(e.querySelector('#language3').textContent).toBe('french');
     });
+
+    itRenders(
+      'should warn with an error message when using Context as consumer in DEV',
+      async render => {
+        const Theme = React.createContext('dark');
+        const Language = React.createContext('french');
+
+        const App = () => (
+          <div>
+            <Theme.Provider value="light">
+              <Language.Provider value="english">
+                <Theme.Provider value="dark">
+                  <Theme>{theme => <div id="theme1">{theme}</div>}</Theme>
+                </Theme.Provider>
+              </Language.Provider>
+            </Theme.Provider>
+          </div>
+        );
+        // We expect 1 error.
+        await render(<App />, 1);
+      },
+    );
+
+    // False positive regression test.
+    itRenders(
+      'should not warn when using Consumer from React < 16.6 with newer renderer',
+      async render => {
+        const Theme = React.createContext('dark');
+        const Language = React.createContext('french');
+        // React 16.5 and earlier didn't have a separate object.
+        Theme.Consumer = Theme;
+
+        const App = () => (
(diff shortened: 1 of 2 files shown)


trueadm:
Not sure why Circle CI is failing, but it seems unrelated to this PR.

engprodigy:
I'm trying to follow the changes as a follow up to PR #13829

bvaughn:
Looks like you need to run `yarn prettier` to update `packages/react-dom/src/server/ReactPartialRenderer.js` btw

packages/react-dom/src/__tests__/ReactDOMServerIntegrationNewContext-test.js
```diff
@@ -249,5 +249,104 @@ describe('ReactDOMServerIntegration', () => {
...
+                  </Theme.Consumer.Provider>
+                </Language.Provider>
+              </Theme.Provider>
+            </div>
+          );
+        };
+        // We expect 1 error.
+        await render(<App />, 1);
+      },
+    );
```
bvaughn: These tests seem a little weak since they don't assert on _what the error message actually is_. I guess `ReactDOMServerIntegrationTestUtils` doesn't provide a mechanism to do this currently, but it seems like it would be an improvement if e.g. `expectErrors` supported either a number or an error of partial messages.

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -1054,9 +1054,35 @@ class ReactDOMServerRenderer {
             return '';
           }
           case REACT_CONTEXT_TYPE: {
-            const consumer: ReactConsumer<any> = (nextChild: any);
-            const nextProps: any = consumer.props;
-            const nextValue = consumer.type._currentValue;
+            let reactContext: ReactContext<any> = (nextChild: any).type;
```
bvaughn: Seems a little odd that we're any-casting something that explicitly _might not be_ `ReactContext` to that type for DEV mode.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationNewContext-test.js
trueadm: I agree, this is something I stumbled into. It can be something we do as a follow up as I know that we're planning on making some changes to the server renderer and this might flow nicely with that workload.

PR #14036, opened 2018-10-30
Rename inputsAreEqual to areHookInputsEqual & move it to shared

Andarist:
I know its bikeshedding but the name of this function triggered my OCD a little bit because it sounds (to me) like a name for a boolean variable. Most commonly other functions with similar purpose are named in the codebase with verb as first work, i.e. `isValidElementType`.

While renaming it I've also noticed that this code was duplicated and already out of sync (one having a dev warning, one not having it) so I thought it would be also a good thing to move it to shared directory to keep a single source of truth.

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRendererHooks.js
+++ b/packages/react-dom/src/server/ReactPartialRendererHooks.js
@@ -7,6 +7,7 @@
  * @flow
  */
 import type {ReactContext} from 'shared/ReactTypes';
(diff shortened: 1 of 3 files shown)


sophiebits:
Seems OK to me. Shared means that it will get copied into each renderer which I guess is fine since that's effectively what we were manually doing before.

Hooks aren't stable so we're allowed to have breaking changes (and we generally feel OK adding warnings in minor releases anyway since they don't affect behavior).

sophiebits:
Thanks!

PR #14040, opened 2018-10-30
Don't lint against Hooks after conditional throw

### LINKED ISSUE #14038: eslint-plugin-react-hooks
```js
function Editor({ data: { page } }: EditorProps) {
  if (page == null) throw Error('missing page data in editor')
  const [editorValue, setEditorValue] = useState(() => {
}
```

I believe this should be a valid case, because throw.

sophiebits:
Seems like this should be OK. Fixes #14038.

Now when tracking paths, we completely ignore segments that end in a throw.


sophiebits:
It's possible that `thrownSegments` can help me? I don't know.

PR #14046, opened 2018-10-31
Add test for useReducer with indirection

jfo84:
I was a little surprised by the difference between client and server rendering when passing the updater function of `useState`, so I added a test for the `useReducer` case to improve the documentation

Cheers!


gaearon:
>by the difference between client and server rendering when passing the updater function of useState

What difference?

jfo84:
@gaearon The first test, the one where I changed the markup, fails if you change it to use the `itRenders` pattern and pass the render function instead of `serverRender`. The count is 3 in the client renders.

I assume that means that there's a mismatch between the client and server render when passing the updater function outside of the component where `useState` is called.

PR #14053, opened 2018-10-31
Implement pauseExecution, continueExecution, dumpQueue for Scheduler

mrkev:
- `pauseExecution`: pauses the scheduler from making progress running subsequent callbacks in the queue.
- `continueExecution`: continues executing callbacks from the queue.
- `dumpQueue`: returns an array of all scheduled callbacks. 

Note: at the moment tests don't pass, because `packages/scheduler/src/__tests__/SchedulerUMDBundle-test.internal.js` ensures the APIs exported across different modules is consistent, but I only added the functions to `umd/scheduler.development`. Please advise!

cc @n8schloss


packages/scheduler/src/Scheduler.js
```diff
@@ -31,6 +31,9 @@ var IDLE_PRIORITY = maxSigned31BitInt;
 var firstCallbackNode = null;
 
 var currentDidTimeout = false;
+// Pausing the scheduler is useful for debugging.
+var isSchedulerPaused = false;
```
mrkev: I think I might want to expose this variable too, since for debugging it's necessary to know if the thing is paused at all in the first place.

fixtures/scheduler/index.html
```diff
@@ -496,6 +524,38 @@ <h2>Tests:</h2>
...
+      updateTestResult(8, 'Done!');
+      displayTestResult(8);
+      checkTestResult(8);
+    })
+
+    // new size should be 2 now
+    updateTestResult(8, `Queue size: ${dumpQueue().length}.`);
+    displayTestResult(8);
+  });
+}
```
bvaughn: Thanks for adding a new integration test!

packages/scheduler/npm/umd/scheduler.development.js
```diff
@@ -67,6 +67,27 @@
...
+      arguments
+    );
+  }
+
+  function unstable_continueExecution() {
+    return global.React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Scheduler.unstable_continueExecution.apply(
+      this,
+      arguments
+    );
+  }
```
bvaughn: I strongly dislike the fact that the newly proposed API methods only exist on the DEV build.

Do other packages do this? I know that sometimes methods are no-ops in production builds, but having them be missing entirely feels weird.

I also think the fact that these methods exist only for DEV reinforces the notion that you could just use existing browser tooling (like the browser's built-in pause/resume functionality).

scripts/rollup/results.json
```diff
@@ -4,29 +4,29 @@
       "filename": "react.development.js",
       "bundleType": "UMD_DEV",
       "packageName": "react",
-      "size": 95936,
-      "gzip": 25258
+      "size": 99473,
+      "gzip": 25976
```
bvaughn: Let's revert the changes to this file.

packages/scheduler/src/Scheduler.js
```diff
@@ -336,6 +347,31 @@ function unstable_scheduleCallback(callback, deprecated_options) {
   return newNode;
 }
 
+function unstable_pauseExecution() {
+  isSchedulerPaused = true;
```
bvaughn: Why isn't the browser's built-in pause/resume functionality sufficient for whatever you're using this for?

packages/scheduler/src/Scheduler.js
```diff
@@ -336,6 +347,31 @@ function unstable_scheduleCallback(callback, deprecated_options) {
...
+
+  if (node !== null) {
+    do {
+      callbacks.push(node);
+      node = node.next;
+    } while (node !== firstCallbackNode);
+  }
+
+  return callbacks;
+}
```
bvaughn: Instead of exposing play, pause, and dump-queue– what if we just exposed the queue itself (`firstCallbackNode`) somewhere globally in DEV mode only? Then we could just write our own tiny helper function that read from it (if that was actually needed).

(This also assumes that we just use the browser's built-in play/pause script execution functionality.

acdlite:
Like @bvaughn  I'm unconvinced that `pauseExecution` and `resumeExecution` are necessary. What do they solve that couldn't be solved by DevTools breakpoints or `debugger` statements?

`dumpQueue` seems useful but also doesn't need to live in the Scheduler package.

packages/scheduler/npm/umd/scheduler.development.js
mrkev: I don't have an opinion on this. no-oping in production seems a good solution in my book.

bvaughn:
> Inspecting the current queue is useful for debugging, but things could change too quickly if not paused. Breakpoints and debugger statements pause all JS execution so it's probably hard to build custom in-page tools around these. How do I update my UI?

Why can't your custom tooling just log to the console? (Why would it need an in-page UI?)

> I'm thinking a useful workflow to understand and debug the scheduler might be stepping through the callbacks one at a time and seeing how the page/queue change after each one. Doing this with breakpoints would be hard since people would need to find the break on the scheduler's code, not their own. This is not something implemented in this PR, but depends on being able to soft-stop the scheduler.

Why would the ability to programmatically pause/unpause the scheduler make this any easier than just setting breakpoints (conditional if needed) in the Source panel and stepping through from there? Aren't you stepping through from a fixed point in code either way?

mrkev:
> Why can't your custom tooling just log to the console?

It can. 

> Why would the ability to programmatically pause/unpause the scheduler make this any easier than just setting breakpoints (conditional if needed) in the Source panel and stepping through from there? Aren't you stepping through from a fixed point in code either way?

Setting breakpoints doesn't allow me to view the scheduler state after each callback execution, unless I breakpoint in the scheduler itself, which would be unnecessary complicated for most devs.

@bvaughn this is why I asked in my original comment, how does tooling for understanding/debugging scheduled code look for you guys? What's your vision here? I want to align goals here.

bvaughn:
> > Why can't your custom tooling just log to the console?
>
> It can. Do we not want an in-page UI?

This PR is the first I've heard one mentioned. But insomuch as we hope for this package to eventually be adopted by browser vendors natively, I think additions like pause/continue/dump might hurt our chances of adoption. (Presumably if these APIs did become native, native browser DevTools would provide this functionality. In the meanwhile, I think it would be nice if we could get by with existing native tooling– pause/resume script execution, breakpoints, console log, etc.)

> Setting breakpoints doesn't allow me to view the scheduler state after each callback execution, unless I breakpoint in the scheduler itself, which would be unnecessary complicated for most devs.

I'm not sure I understand the last bit. Most devs probably wouldn't be debugging the scheduler itself, would they? Seems like they would be setting breakpoints in their own (app or framework) code.

bvaughn:
> In terms of browser vendor adoption, we're not asking browser vendors to standardize our implementation of scheduling, and I don't think that we'll be getting rid of the JSScheduler all together once somethings out.

Please correct me if I'm wrong.

I was under the impression that we essentially _were_ asking them to standardize what's in the "scheduler" package. The `JSScheduler` is a Facebook wrapper around that package, so it's reasonable to add layers to it, but this PR is proposing changes to the "scheduler" package.

> In terms of pausing and allowing developers to step through one function at a time, it's essential that when introducing new technologies and concepts to Facebook developers that we allow debugging tooling in order for folks (and for us) to be convinced that things are working correctly. Without some kind of debugging UI I guarantee that scheduling will get blamed for bugs it has nothing to do with.

This sounds reasonable.

Have we considered creating a custom web extension for this purpose? (I'm not sure how this would work initially. Just thinking out loud.)

n8schloss:
> > In terms of browser vendor adoption, we're not asking browser vendors to standardize our implementation of scheduling, and I don't think that we'll be getting rid of the JSScheduler all together once somethings out.
> 
> Please correct me if I'm wrong.
> 
> I was under the impression that we essentially _were_ asking them to standardize what's in the "scheduler" package. The `JSScheduler` is a Facebook wrapper around that package, so it's reasonable to add layers to it, but this PR is proposing changes to the "scheduler" package.
> 

Yes, but my point is that I think there will always be a place for some type of scheduler package even inside of React and it's there that this type of tooling should live. I do think that browsers will want to think about debugging tooling when implementing a scheduler, but that's orthogonal to this.

> > In terms of pausing and allowing developers to step through one function at a time, it's essential that when introducing new technologies and concepts to Facebook developers that we allow debugging tooling in order for folks (and for us) to be convinced that things are working correctly. Without some kind of debugging UI I guarantee that scheduling will get blamed for bugs it has nothing to do with.
> 
> This sounds reasonable.
> 
> Have we considered creating a custom web extension for this purpose? (I'm not sure how this would work initially. Just thinking out loud.)

Cool :) Yeah, I think an open source extension could eventually be neat here too! The extension would still need hooks into the scheduler though, so it doesn't change much for this diff imo.

bvaughn:
> Yes, but my point is that I think there will always be a place for some type of scheduler package even inside of React and it's there that this type of tooling should live. I do think that browsers will want to think about debugging tooling when implementing a scheduler, but that's orthogonal to this.

Agreed. I think the concern Andrew and I had was that adding non-essential code to this package may hurt our chance for browser adoption.

> Cool :) Yeah, I think an open source extension could eventually be neat here too! The extension would still need hooks into the scheduler though, so it doesn't change much for this diff imo.

Well I was asking because if there was a global var (in DEV mode) then perhaps the extension could read from it and provide its own UI to inspect the scheduled state while the scheduler was paused (using the browser's built-in pause script execution functionality).

Not sure entirely how this would work, since pausing script execution would also pause posted messages between an extension and the front-end, but...maybe it would be possible somehow.

bvaughn:
> Exactly :) Even in that case we're still going to need to let the extension hook into and pause the scheduler.

Why is this _necessarily_ true? I was suggesting that we could play/pause the scheduler using the browser's built-in "pause script execution" functionality.

Let's talk about this during the sync tomorrow? This is getting to be too much back and forth. 😄

n8schloss:
> Why is this _necessarily_ true? I was suggesting that we could play/pause the scheduler using the browser's built-in "pause script execution" functionality.
> 
> Let's talk about this during the sync tomorrow? This is getting to be too much back and forth. 😄

I'm not sure if we can do that and still get the debugger to display what we need and beyond that, it won't let us build the scheduler stepping mechanism that we want to build. So hooks into the scheduler are very important. Anyway, happy to chat about it tomorrow.

mrkev:
@acdlite @bvaughn updated with changes discussed

acdlite:
Please wrap in a feature flag, then I think this is ok to land.

PR #14065, opened 2018-11-01
Invalid actualDuration+treeBaseDuration for hidden+suspended trees

bvaughn:
While testing some of our new APIs in DevTools, I noticed that the profiler (both the API and the DevTools visual profiler) reported invalid `actualDuration` and `treeBaseDuration` values when suspending.

The `treeBaseDuration` was incorrect for the fallback commit. It only included the fallback subtree duration, when it should also include the time spent rendering the hidden subtree.

The `actualDuration` value was incorrect for both the fallback commit _and_ the final, resolved commit. It didn't include time spent rendering the component that actually read from the cache (the one that threw).

- [x] Fixed `treeBaseDuration` by propagating its value from the suspended tree to the `Fragment` React temporarily wraps around it when showing the fallback UI.
- [x] Fixed `actualDuration` by recording elapsed profiler time in the event of an error.
- [x] Fixed `actualDuration` in concurrent mode by propagating the time spent rendering the suspending component to its parent.

Also updated `ReactSuspensePlaceholder-test.internal` to cover these new cases.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -66,7 +66,12 @@ import {
 } from './ReactChildFiber';
 import {processUpdateQueue} from './ReactUpdateQueue';
 import {NoWork, Never} from './ReactFiberExpirationTime';
-import {ConcurrentMode, StrictMode, NoContext} from './ReactTypeOfMode';
+import {
+  ConcurrentMode,
+  NoContext,
+  ProfileMode,
+  StrictMode,
+} from './ReactTypeOfMode';
 import {
   shouldSetTextContent,
   shouldDeprioritizeSubtree,
@@ -743,7 +748,7 @@ function mountLazyComponent(
 ) {
   if (_current !== null) {
     // An lazy component only mounts if it suspended inside a non-
-    // concurrent tree, in an inconsistent state. We want to tree it like
+    // concurrent tree, in an inconsistent state. We want to treat it like
     // a new mount, even though an empty version of it already committed.
     // Disconnect the alternate pointers.
     _current.alternate = null;
@@ -829,7 +834,7 @@ function mountIncompleteClassComponent(
 ) {
   if (_current !== null) {
     // An incomplete component only mounts if it suspended inside a non-
-    // concurrent tree, in an inconsistent state. We want to tree it like
+    // concurrent tree, in an inconsistent state. We want to treat it like
     // a new mount, even though an empty version of it already committed.
     // Disconnect the alternate pointers.
     _current.alternate = null;
@@ -886,7 +891,7 @@ function mountIndeterminateComponent(
 ) {
   if (_current !== null) {
     // An indeterminate component only mounts if it suspended inside a non-
-    // concurrent tree, in an inconsistent state. We want to tree it like
+    // concurrent tree, in an inconsistent state. We want to treat it like
     // a new mount, even though an empty version of it already committed.
     // Disconnect the alternate pointers.
     _current.alternate = null;
@@ -1239,6 +1244,15 @@ function updateSuspenseComponent(
           NoWork,
           null,
         );
+
+        if (enableProfilerTimer && workInProgress.mode & ProfileMode) {
+          // Fiber treeBaseDuration must be at least as large as the sum of children treeBaseDurations.
+          // Otherwise the profiler's onRender metrics will be off,
+          // and the DevTools Profiler flamegraph will visually break as well.
+          primaryChildFragment.treeBaseDuration =
+            currentPrimaryChild.treeBaseDuration;
+        }
+
         primaryChildFragment.effectTag |= Placement;
         primaryChildFragment.child = currentPrimaryChild;
         currentPrimaryChild.return = primaryChildFragment;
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -753,7 +753,7 @@ function mountLazyComponent(
 ) {
   if (_current !== null) {
     // An lazy component only mounts if it suspended inside a non-
-    // concurrent tree, in an inconsistent state. We want to tree it like
```
sophiebits: 🌳

bvaughn:
Looks like resolving a suspended tree doesn't properly account for `actualDuration` either with regard to the `unstable_Profiler` durations. Will add more tests shortly.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1234,6 +1234,15 @@ function updateSuspenseComponent(
           NoWork,
           null,
         );
+
+        if (enableProfilerTimer && workInProgress.mode & ProfileMode) {
+          // Fiber treeBaseDuration must be at least as large as the sum of children treeBaseDurations.
+          // Otherwise the profiler's onRender metrics will be off,
+          // and the DevTools Profiler flamegraph will visually break as well.
+          primaryChildFragment.treeBaseDuration =
+            currentPrimaryChild.treeBaseDuration;
```
acdlite: Need to transfer from the siblings, too.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1234,6 +1234,15 @@ function updateSuspenseComponent(
           NoWork,
           null,
         );
+
+        if (enableProfilerTimer && workInProgress.mode & ProfileMode) {
+          // Fiber treeBaseDuration must be at least as large as the sum of children treeBaseDurations.
+          // Otherwise the profiler's onRender metrics will be off,
+          // and the DevTools Profiler flamegraph will visually break as well.
```
acdlite: Could you rewrite this comment in terms of what it means semantically instead of what will break if it goes wrong?

"treeBaseDuration is the sum of all the children durations" would be sufficient

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1282,6 +1282,12 @@ function renderRoot(root: FiberRoot, isYieldy: boolean): void {
         didFatal = true;
         onUncaughtError(thrownValue);
       } else {
+        if (enableProfilerTimer && nextUnitOfWork.mode & ProfileMode) {
+          // Record the time spent rendering before an error was thrown.
+          // This avoids inaccurate Profiler durations in the case of a suspended render.
+          stopProfilerTimerIfRunningAndRecordDelta(nextUnitOfWork, true);
```
acdlite: This gets called again in `completeUnitOfWork`, with no work in between. Seems wrong?

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1286,6 +1286,16 @@ function renderRoot(root: FiberRoot, isYieldy: boolean): void {
           // Record the time spent rendering before an error was thrown.
           // This avoids inaccurate Profiler durations in the case of a suspended render.
           stopProfilerTimerIfRunningAndRecordDelta(nextUnitOfWork, true);
+
+          // HACK Also propagate actualDuration for the time spent in the fiber that errored.
+          // This avoids inaccurate Profiler durations in the case of a suspended render.
+          // This happens automatically for sync renders, because of resetChildExpirationTime.
```
acdlite: Suggested alternative: bubble this in `completeUnitOfWork`, where `resetChildExpirationTime` would normally go for a completed fiber.

packages/react-reconciler/src/ReactFiberBeginWork.js
bvaughn: Good catch. Fixed and updated tests to cover this via 85c5831

packages/react-reconciler/src/ReactFiberScheduler.js
bvaughn: We [already have a similar bubbling there](https://github.com/facebook/react/blob/8b87ebf5b0053838fb58e4ad05c24e134030d75a/packages/react-reconciler/src/ReactFiberScheduler.js#L1059-L1070) to ensure that `actualDuration` bubbles all the way up to the `Suspense` component (or error boundary).

I think it's tricky to propagate this specific value (from the fiber that threw) in `completeUnitOfWork` without counting it too many times (since we hit that code path for ancestors too).

bvaughn: Scratch that, I misunderstood your initial suggestion. You're right!

bvaughn:
Rebased to account for improvements made in e9a2ec9 😄

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1246,6 +1251,18 @@ function updateSuspenseComponent(
           NoWork,
           null,
         );
+
```
acdlite: Suggested comment:

> Because `primaryChildFragment` is a new fiber that we're inserting as the parent of a new tree, we need to set its `treeBaseDuration`.

PR #14069, opened 2018-11-02
Warn for bad useEffect return value

sophiebits:
Mostly to catch this:

```js
useEffect(async () => {
  // ...
  return cleanup;
});
```

Is this too restrictive? Not sure if you would want to do like

```js
useEffect(() => ref.current.style.color = 'red');
```

which would give a false positive here. We can always relax it to only warn on Promises if people complain.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -295,8 +295,26 @@ function commitHookEffectList(
       if ((effect.tag & mountTag) !== NoHookEffect) {
         // Mount
         const create = effect.create;
-        const destroy = create();
-        effect.destroy = typeof destroy === 'function' ? destroy : null;
+        let destroy = create();
+        if (typeof destroy !== 'function') {
+          if (__DEV__) {
+            if (destroy !== null && destroy !== undefined) {
+              warningWithoutStack(
+                false,
+                'useEffect function must return a cleanup function or ' +
+                  'nothing.%s%s',
+                typeof destroy.then === 'function'
+                  ? ' Promises and useEffect(async () => ...) are not ' +
+                    'supported, but you can call an async function inside an ' +
+                    'effect.'
+                  : '',
+                getStackByFiberInDevAndProd(finishedWork),
+              );
+            }
+          }
+          destroy = null;
+        }
+        effect.destroy = destroy;
       }
       effect = effect.next;
     } while (effect !== firstEffect);
--- a/packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
@@ -51,4 +51,32 @@ describe('ReactHooks', () => {
     ]);
     expect(ReactTestRenderer).toHaveYielded(['Did commit: A, B']);
   });
+
+  it('warns for bad useEffect return values', () => {
+    const {useLayoutEffect} = React;
+    function App(props) {
+      useLayoutEffect(() => {
+        return props.return;
+      });
+      return null;
+    }
+    let root;
+
+    expect(() => {
+      root = ReactTestRenderer.create(<App return={17} />);
+    }).toWarnDev([
+      'Warning: useEffect function must return a cleanup function or ' +
+        'nothing.\n' +
+        '    in App (at **)',
+    ]);
+
+    expect(() => {
+      root.update(<App return={Promise.resolve()} />);
+    }).toWarnDev([
+      'Warning: useEffect function must return a cleanup function or ' +
+        'nothing. Promises and useEffect(async () => ...) are not supported, ' +
+        'but you can call an async function inside an effect.\n' +
+        '    in App (at **)',
+    ]);
+  });
 });


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -295,7 +295,20 @@ function commitHookEffectList(
...
+            warningWithoutStack(
+              false,
+              'useEffect function must return a cleanup function or nothing.%s%s',
+              typeof destroy.then === 'function'
+                ? ' Promises and async functions are not supported.'
+                : '',
+              getStackByFiberInDevAndProd(finishedWork),
+            );
+          }
+          destroy = null;
```
alexeyraspopov: In production the branch `if (typeof destroy !== 'function') {` will still be present with `destroy = null`. Is it worth keeping it since the next line checks the type again?

sophiebits: Thanks for the catch, I intended to remove the conditional from the next line.

FredyC:
I would like to also know what's the recommended approach for running async code in the effect. I am so used to have `async componentDidMount` and similar that this really feels rather restrictive suddenly.

What about adding `useAsyncEffect` which would be expecting this and as the tradeoff it wouldn't support cleanup as in most cases the async code does not need it. Or some other way to opt-out of this warning?

gaearon:
The intended solution for most async stuff is Suspense. It's unfortunate we don't have both yet so at first stages people are going to be doing `useEffect` more than intended. But that's mostly a transitional period.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -295,8 +295,21 @@ function commitHookEffectList(
       if ((effect.tag & mountTag) !== NoHookEffect) {
         // Mount
         const create = effect.create;
-        const destroy = create();
-        effect.destroy = typeof destroy === 'function' ? destroy : null;
+        let destroy = create();
+        if (typeof destroy !== 'function') {
+          if (__DEV__ && destroy !== null && destroy !== undefined) {
```
gaearon: Nit: I kinda perfer `if (__DEV__)` to be a separate line for easier visual scannning

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -295,8 +295,21 @@ function commitHookEffectList(
...
+            warningWithoutStack(
+              false,
+              'useEffect function must return a cleanup function or nothing.%s%s',
+              typeof destroy.then === 'function'
+                ? ' Promises and async functions are not supported.'
```
gaearon: Maybe add "but you can call an async function without waiting for it" or something. Or I expect we'll keep getting issues asking for this.

FredyC:
@gaearon 
> The intended solution for most async stuff is Suspense. It's unfortunate we don't have both yet so at first stages people are going to be doing `useEffect` more than intended. But that's mostly a transitional period.

I am still not sure what should we do at this moment? Running async function without `await`ing does not make much sense to me. Can you please provide some example?

FredyC:
I should probably point out that my intention is not to have `useEffect` wait for Promise resolution and then magically rerender. My point is that I might simply want to call something async and then just set state upon finishing it. It's basically an alternative for a missing Suspense.

```js
function Component() {
  const [data, setData] = useState(null)
  useEffect(async () => {
    const data = await fetchData()
    setData(data)
    // in this contrieved example I could probably do it like this
    fetchData().then(setData)
    // but it's usually more complicated than that.
  })
}
```

I can imagine this will be much prettier with Suspense, but since we can wrap this into a custom hook then changing to suspensified solution should be fairly easy, isn't that so?

arianon:
@FredyC Maybe you could try this.

```js
function Component() {
  const [data, setData] = useState(null)

  useEffect(() => {
	async function fetchAndDoStuffWithData() {
	  const data = await fetchData()
	  // do stuff here...
	  setData(data);
	}

    fetchAndDoStuffWithData();
  })
}
```

andrei-markeev:
yep so instead 
```js
useEffect(async () => await log("something"))
```
people will now have to write 
```js
useEffect(() => { (async () => await log("something"))() })
```
or use `.then` (which is even worse imo).

Nice solution to the problem (or is it really?) 🤔

phanirithvij:
As @arianon suggested [here](https://github.com/facebook/react/pull/14069#issuecomment-435491833) we can use a function inside it or it can also be a self-invoking function. 
```javascript
const Component = () => {
  const [data, setData] = useState(null);

  useEffect(() => {
	(async () => {
	  const data = await fetchData();
	  // do stuff here...
	  setData(data);
	})();
  });
}
```

FredyC:
I've pretty much settled with a custom hook. It has an obvious downside that I cannot use cleanup function there, but so far I haven't seen a case for an async effect that would need a cleanup. It just reads much better like this imo.

```jsx
const Component = () => {
  const [data, setData] = useState(null)

  useAsyncEffect(async () => {
    const data = await fetchData();
    // do stuff here...
    setData(data);
  }, [])
}
```

krzkaczor:
I created a small reusable hook, ready to copy:

```typescript
import { useEffect } from "react";

/**
 * Like useEffect but works with async functions and makes sure that errors will be reported
 */
export function useAsyncEffect(effect: () => Promise<any>) {
  useEffect(() => {
    effect().catch(e => console.warn("useAsyncEffect error", e));
  });
}

```

bmmpt:
Something like this would be nicer:
```
export const useEffectAsync = (
    effect: () => Promise<any>,
    deps?: DependencyList
): void => {
    useEffect(() => {
        effect().catch(e =>
            // eslint-disable-next-line no-console
            console.warn('useAsyncEffect error', e)
        );
    }, deps);
};
```

But it throws warnings:
  Line 12:  React Hook useEffect was passed a dependency list that is not an array literal. This means we can't statically verify whether you've passed the correct dependencies        react-hooks/exhaustive-deps
  Line 12:  React Hook useEffect has a missing dependency: 'effect'. Either include it or remove the dependency array. If 'effect' changes too often, find the parent component that defines it and
 wrap that definition in useCallback  react-hooks/exhaustive-deps

bmmpt:
@FredyC What does your custom hook useAsyncEffect look like?

FredyC:
@bmmpt Do you realize those warnings are just from friendly neighbor ESLint and you can disable those per line? There is nothing else to be done here.

bmmpt:
It would be nice if react hooks ESLint plugin was smart enough to suppress them if they're false positives. I'm not a big fan of disabling ESLint per line, unless there really is no other option.

FredyC:
@bmmpt I've never written an eslint plugin, but I can imagine it's not so easy to write heuristics that are so smart to avoid any false positives. The `useEffectAsync` is a custom hook you will be reusing. It's not like you will have disable comments all over the code. We are living in the world of tradeoffs, nothing is perfect.

bmmpt:
I'm just using hooks for a couple days now. Is passing deps from useEffectAsync into useEffect strictly necessary, or does React Hooks do it's magic if we just pass deps to useEffectAsync?
```
export const useEffectAsync = (
    effect: () => Promise<any>,
    // @ts-ignore
    // eslint-disable-next-line no-unused-vars
    deps?: DependencyList
): void => {
    useEffect(() => {
        effect().catch(e =>
            // eslint-disable-next-line no-console
            console.warn('useAsyncEffect error', e)
        );
    });
};
```
(And not without disabled warnings or errors :))

Issue #14073, opened 2018-11-02
Suspense fallback remounts when each child resolves


adwilk:
**Do you want to request a *feature* or report a *bug*?**
Not sure if its a bug or a feature request.

**What is the current behavior?**
The component given as a fallback to Suspense is remounted each time a lazy child resolves.

https://codesandbox.io/s/z6v6x3n1np
This example shows how a fallback component, which counts up, is mounted more than once.

**What is the expected behavior?**
The fallback is mounted once and is only unmounted when all lazy children are resolved.
In the example the counter would run up to 9.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
react@16.6.0

arianon:
Looks like expected behavior to me, ~~when `LazyText1` resolves, the `Suspense` component tries to render its children once again, but then `LazyText2` throws another promise, so it renders the fallback again, and so on.~~ Nope, all three promises are thrown in parallel, but every time a promise resolves, `Suspense` attempts to re-render the children, but seeing that some children are unresolved, it renders the fallback.

However, [in Concurrent Mode the observed behaviour aligns with your expected behaviour][1], with one apparent bug: the children don't render when `LazyText3`'s delay is over 8000 milliseconds, on my machine at least. I'm not sure what's happening, but I'll try to look into this further.

[1]: https://codesandbox.io/s/4lvn7ror8x

adwilk:
Here is a slightly less contrived example than what I shared above
https://codesandbox.io/s/2v8l3vn67p

Here there is a CSS loading spinner, due to the remounting of the fallback component the animation of the spinner is not smooth.

gaearon:
Seems like a bug.

gaearon:
I added a failing test in https://github.com/facebook/react/pull/14078

adwilk:
Thanks 👍

OndeVai:
Has this been fixed? I'm seeing the same issue now with React 17.0.1

gaearon:
That one was fixed, so you're seeing something different. File a bug with a minimal repro case.

OndeVai:
Actually, in my case, I have a Suspense (call it SuspenseA) nested several levels below another Suspense (call it SuspenseB). For  SuspenseB, SuspenseA renders its fallback UI only if I do not provide a fallback for SuspenseB. In my case I did not want any fallback UI for B at all. So, I ended up providing a null fallback and everything worked as desired. 

[...]

PR #14078, opened 2018-11-02
Add a failing test for remounting fallback in sync mode

gaearon:
This is a failing test for https://github.com/facebook/react/issues/14073.
See last line for the failing case.

I'm not actually sure it's a bug. It makes sense that if we render `null` in sync mode, we end up remounting. But it seems like far from ideal and I wonder if we can fix it.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
@@ -631,5 +631,65 @@ describe('ReactSuspense', () => {
       expect(root).toFlushAndYield(['Step: 3']);
       expect(root).toMatchRenderedOutput('Step: 3');
     });
+
+    it('does not remount the fallback while suspended children resolve in sync mode', () => {
+      let mounts = 0;
+      class ShouldMountOnce extends React.Component {
+        componentDidMount() {
+          mounts++;
+        }
+        render() {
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
```diff
@@ -631,5 +631,65 @@ describe('ReactSuspense', () => {
...
+        'Suspend! [Child 3]',
+        'Loading...',
+      ]);
+      jest.advanceTimersByTime(1000);
+      expect(ReactTestRenderer).toHaveYielded([
+        'Promise resolved [Child 2]',
+        'Child 2',
+        'Suspend! [Child 3]',
+        'Loading...',
+        'Suspend! [Child 3]', // TODO: why does this suspend twice?
```
gaearon: This was also weird

acdlite: It's a known flaw with sync mode. Pings aren't batched, so we get a synchronous retry for every individual ping. Added a comment in the rebased version of your test:

https://github.com/facebook/react/blob/74322975b74170fa4eabda5e5c0496a6c2a9356d/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js#L739-L741

Issue #14013, opened 2018-10-28
Suspense doesn't "resolve" for subsequent times if was suspended again 


YevheniiKravchenko:
**Do you want to request a *feature* or report a *bug*?**
Possibly a bug

**What is the current behavior?**
Repro demo: GitHub user search. 
Type one character and wait after the list is rendered - everything works correctly. Subsequent fetches work this way.
Type two characters faster than results for the first character are rendered and fallback will never go although the second request is successful. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**
https://codesandbox.io/s/k5v6ojz107

**What is the expected behavior?**
Suspense "resolves" with last resolved fetch and renders the list of users

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

react: 16.7.0-alpha.0
react-dom: 16.7.0-alpha.0
react-cache: 2.0.0-alpha.0

vramana:
![image](https://user-images.githubusercontent.com/3112509/47636768-872cf480-db7f-11e8-9350-f5ac9947f50b.png)

I think that's intentional behaviour. Adding a key on the Suspense will fix it.

aweary:
I don't think thats intentional, it should handle cases like this without having to add a `key`.

YevheniiKravchenko:
At least it should emit some warning, not just quietly show fallback forever.

gaearon:
This is definitely a bug

PR #14079, opened 2018-11-02
Add failing test for stuck Suspense fallback

gaearon:
Another issue that was discovered in https://github.com/facebook/react/issues/14073#issuecomment-435373550 (although this one is specific to concurrent mode). This just looks like a bug.

Setting second child's `ms` lower does not reproduce it.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
@@ -631,5 +631,33 @@ describe('ReactSuspense', () => {
       expect(root).toFlushAndYield(['Step: 3']);
       expect(root).toMatchRenderedOutput('Step: 3');
     });
+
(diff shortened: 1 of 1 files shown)


gaearon:
Might be the same problem as https://github.com/facebook/react/issues/14013? Don't know.

sophiebits:
What is supposed to set .alreadyCaptured to false? It seems like once it gets set to true, the only thing that can set it back to false (and thus trigger a retry of the primary content) is the commit phase, which never runs for a suspended update. Not sure if I'm missing some part of the model in my head.

cc @acdlite

sophiebits:
> Might be the same problem as #14013? Don't know.

Looks like it. Seems like once something pings once in concurrent mode, further pings end up being no-ops.

acdlite:
Fixed by https://github.com/facebook/react/pull/14083

Issue #13999, opened 2018-10-27
Suspense doesn't stop rendering fallback after Lazy resolves


Velenir:
**Do you want to request a *feature* or report a *bug*?**
bug, I believe, unless I misunderstand how it works
**What is the current behavior?**
Given a `Suspense` inside `Suspense` case like
```
<Suspense fallback="Fallback 1">
  <LazyText1 text="First text" />
  <Suspense fallback="Fallback 2">
    <LazyText2 text="Second text!" />
  </Suspense>
</Suspense>
```

Promises in `LazyTexts` start resolving at the same time, but if `LazyText2` is resolved first it never renders, `Fallback 2` stays.

Result:
```
First text

Fallback 2
```

Please see https://codesandbox.io/s/234o4zry7n

**What is the expected behavior?**

Every `LazyText` displayed in the end
```
First text

Second text!
```

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
react, react-dom 16.6.0

gaearon:
Looks like a bug to me.

PR #14083, opened 2018-11-03
[suspense] Avoid double commit by re-rendering immediately and reusing primary children

acdlite:
To support Suspense outside of concurrent mode, any component that starts rendering must commit synchronously without being interrupted. This means the normal path, where we unwind the stack and try again from the nearest Suspense boundary, won't work.

We used to have a special case where we commit the suspended tree in an incomplete state. Then, in a subsequent commit, we re-render using the fallback.

The first part — committing an incomplete tree — hasn't changed with this PR. But I've changed the second part — now we render the fallback children immediately, within the same commit.

Fixes https://github.com/facebook/react/issues/13999
Fixes https://github.com/facebook/react/issues/14013
Fixes https://github.com/facebook/react/issues/14073
Fixes https://github.com/facebook/react/pull/14078
Fixes https://github.com/facebook/react/pull/14079

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
@@ -9,6 +9,7 @@
 
 'use strict';
 
+let ReactFeatureFlags;
 let React;
 let ReactDOM;
 let Suspense;
@@ -20,6 +21,8 @@ describe('ReactDOMSuspensePlaceholder', () => {
 
   beforeEach(() => {
     jest.resetModules();
+    ReactFeatureFlags = require('shared/ReactFeatureFlags');
+    ReactFeatureFlags.enableHooks = true;
     React = require('react');
     ReactDOM = require('react-dom');
     ReactCache = require('react-cache');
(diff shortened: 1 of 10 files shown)


acdlite:
@gaearon Yep it fixes it

acdlite:
> Why is this fixing the bugs? Seems like we should be able to fix the bugs without changing the semantics?

It fixes two bugs. The first one, where React fails to switch back to the fallback even after a promise resolves, was just a mistake. I could split that into a separate PR if you like. (I wasn't even intending to fix that bug in this PR — I just happened to fix it in the course of refactoring.)

The other bug is that the fallback is always deleted on every ping. That one can't be fixed without changing semantics.

> It’s always slightly broken. Why is this brokenness better?

> it does two passes which is not possible in other cases so we’d have to think what other consequences that causes.

Two passes is how it works in concurrent mode, too. It's also how error boundaries work, both in and out of concurrent mode.

The difference is that instead of reusing the current primary tree, we update them in an inconsistent state and hide them. I'd argue this is less of a deviation from the concurrent version; the only difference now is whether we throw out the suspended work-in-progress tree or if we throw it out and revert to the current one.

As supporting evidence for why the updated model is simpler, note that the changes to the SuspenseState type. We can now track fewer pieces of state because everything happens within a single render phase: https://github.com/facebook/react/pull/14083/files#diff-2c5a7d418c1fe3bfba84500fc7231482

acdlite:
Also consider if I were refactoring in the opposite direction. The previous semantics make less sense — for one, because it forces us to delete and remount the fallback on every ping that doesn't completely resolve the tree.

But also because it's unnecessary. We can do everything within a single render phase. The only reason we used to not be able to do that is because 1) we needed to commit the primary tree in order to fire the effects, for compatibility outside strict mode 2) we needed to delete the tree in a subsequent, sync commit to prevent the user from seeing an inconsistent state.

Since #13823, though, we don't do step 2) anymore. Instead of deleting the inconsistent tree, we hide it, so that we can preserve the state.

So if we were deciding on the least-bad non-concurrent mode semantics to use today, there's no way we'd choose the model that's currently in master.

sebmarkbage:
> I'd argue this is less of a deviation from the concurrent version

The goal here is to deviate from sync mode as little as possible. Not to bring it into concurrent mode. The goal is to make it (almost) completely safe to adopt this in sync mode. It's a non-goal to make the sync mode the best that it can be - if that is causing further risk of breakages during adoption of Suspense.

acdlite:
Chatted offline, here's the summary:

- The key change in this PR is to render both sets of children — the suspended children and the fallback children — within a single commit. This fixes the issue where the fallback is deleted and remounted on every synchronous ping.
- The PR also happens to fix another bug (#13999 and duplicates) but that was an accident. I should have made this clearer in the PR description.
- The conceptual model (in non-concurrent mode) is that we always render both sets of children. Then, in the commit phase, we hide either the primary set or the fallback set depending on whether something in the primary set suspended.
- In our actual implementation, we cheat a bit in the non-suspended case by rendering nothing in place of the fallback, since we know it won't be needed.

Follow-up items:

- Should toggle the visibility of the children in the mutation phase, not the layout phase. That way if any parents in between the Suspense boundary and the host children read the children in a lifecycle, their visibility has already been updated.

gaearon:
Thanks for posting the summary. This is really helpful for understanding.

Issue #13876, opened 2018-10-17
Uncontrolled input type="checkbox" reflects updating `defaultChecked` in Edge and Safari


berlysia:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**

Updating the value of `defaultChecked` causes a change DOM property's `checked` value.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

https://codesandbox.io/s/yjop5zwmr9

**What is the expected behavior?**

Updating the value of `defaultChecked` should not affect to DOM property's `checked` value.
(no log in codesandbox is expected)

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

reproduce in React v15.2.0 and v16.5.2, ReactDOM v15.2.0 and v16.5.2.
not reproduce in v15.1

browser: Safari 12 on macOS Sierra, Edge 42 on Windows 10
not reproduce in Chrome(70, beta), Chrome(72, canary), Firefox(62), IE11 on Win10
smartphone browsers are not checked.

gaearon:
@aweary @nhunzaker Do you remember if this was intentional?

nhunzaker:
I am not sure if it was intentional, but this started happening with this commit: https://github.com/facebook/react/commit/f98e6c195568641fbca099e3cd2632a0d37d364f

I think this happens because the checked property never gets assigned so it never detaches from the defaultChecked assignment. I think this happens here:

https://github.com/facebook/react/blob/595b4f945b703128af7e65750726c0cbf073ca58/packages/react-dom/src/client/ReactDOMInput.js#L228-L233

This is curious! The comment says that it should only change the defaultChecked when the type changes, but it doesn't do any comparison for that. Maybe this should be:

```javascript
// When syncing the checked attribute, it only changes when it needs
// to be removed, such as transitioning from a checkbox into a text input
if (props.checked == null && props.defaultChecked != null) {
  // New: only change when the node type doesn't match
  if (props.type !== node.type) {
    node.defaultChecked = !!props.defaultChecked;
  }
}
``` 

More curious! I wrote this comment 😢

nhunzaker:
[...]

PR #14084, opened 2018-11-03
Remove checked attribute when it becomes null

nhunzaker:
This commit fixes an issue where removing input checked properties when switching input types did not remove the checked attribute.

This is the behavior for the React Fire fork of this behavior, however this case wasn't handled on the standard ReactDOM code paths.

I noticed this as a part of https://github.com/facebook/react/issues/13876, but it does not fix the original poster's issue.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMInput-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMInput-test.js
@@ -2036,4 +2036,25 @@ describe('ReactDOMInput', () => {
       expect(node.hasAttribute('value')).toBe(false);
     });
   });
+
+  it('removes the checked attribute when switching types on a controlled input', function() {
+    ReactDOM.render(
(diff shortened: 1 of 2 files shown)


gaearon:
Can you clarify:

* what this is broken today
* when did this get broken
* did it ever work

nhunzaker:
It appears that this never worked. I can reproduce this all the way down to React 0.4.0 (maybe 0.3 too).

I think we're just now catching this because it's rare in the wild, or awkward to write anyway. You'd have to do something like this:

```javascript
class Fixture extends React.Component {
  state = {
    toggled: false
  }

  componentDidMount() {
    this.setState({toggled: true});
  }

  render() {
    if (this.state.toggled) {
      return <input type="text" defaultValue="" />
    } else {
      return <input type="checkbox" defaultChecked={true} />
    }
  }
})
```

Additionally, it doesn't look like there's ever been an issue filed for this either (at least as far as I can tell).

PR #14085, opened 2018-11-03
Add Debug Tools Package for Introspection of Hooks

sebmarkbage:
Adds introspection of primitive and custom hooks in a single function component. If you feed it a standalone component it just gives you the initial/default values as the current value of each primitive. If you pass it a Fiber, it gives you the current values by reading the state from the Fiber.

Ideally we want this in DevTools but it requires so much implementation details that I don't think we'll want to maintain this with multi-version support. We'll probably have to move to snapshot versions of this package in DevTools, or just snapshoting all of DevTools. It's probably useful debug info for others too. So I decided to just make this a new separate package.

The algorithm goes something like this:

1) If running in Fiber mode, set up contexts of all parent providers and set the current hook to the first one of the Fiber. If running in standalone mode, skip this step.
2) Compute the stack trace from the root of the inspection function.
3) Set up a custom Dispatcher.
4) Compute the stack traces to all primitive hooks, using the custom Dispatcher, when used outside a component.
5) Execute the render function of the inspected component. This will call all hooks in order since they have to be unconditional. It's safe because the render function as no side-effects on its own.
6) Inside each primitive hook in the custom Dispatcher, log each primitive, it's current stateful value, and the stack trace at the time it was invoked. T
his will give us a list of all primitive hooks and their stacks.
7) For each primitive hook in the log, process the stack trace by removing any shared ancestors with the root of the inspection function. This ensures that the bottom of the stack is the render function. Also, for that particular primitive, remove the top of the stack that is shared with the stack when that primitive is called stand alone from a function component. This removes the call to the primitive itself. If the top most function call is called "useState" or whatever the primitive is called, remove that too since it's the React package wrapper around the dispatcher. The result should be a stack of only the custom hooks.
8) Loop through the list of primitive hooks, use the stack of custom hooks to form a tree structure. Shared parent paths means that they go into the same subnode.
9) ...
10) Profit.

Currently custom hooks doesn't have a "value" associated with them so you can only introspect the current values of primitives. In a follow up, we can add a DEV only API to share more information from custom hooks with the devtools. E.g. `useInspect(value)`


packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.internal.js
```diff
@@ -0,0 +1,167 @@
...
+ * @jest-environment node
+ */
+
+'use strict';
+
+let React;
+let ReactTestRenderer;
+let ReactDebugTools;
+
+describe('ReactHooksInspectionIntergration', () => {
```
sophiebits: add a custom hook integration test?

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -0,0 +1,478 @@
...
+    rootIndex - primitiveIndex < 2
+  ) {
+    // Something went wrong. Give up.
+    return null;
+  }
+  return hookStack.slice(primitiveIndex, rootIndex - 1);
+}
+
+function parseCustomHookName(functionName: string): string {
+  let startIndex = functionName.indexOf('.');
```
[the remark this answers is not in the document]
sebmarkbage: Good catch. Don’t know if any conditions report multiple levels but there is at least one.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -0,0 +1,478 @@
...
+    if (rootIndex !== -1) {
+      mostLikelyAncestorIndex = i;
+      return rootIndex;
+    }
+  }
+  return -1;
+}
+
+function isReactWrapper(functionName, primitiveName) {
+  let expectedPrimitiveName = 'use' + primitiveName;
```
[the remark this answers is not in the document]
sebmarkbage: I figured it would be unlikely for a custom hook to be called useAbuseState or something like that.

Not too happy about this heuristic but not sure what’s better.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -0,0 +1,478 @@
...
+export function inspectHooksOfFiber(fiber: Fiber) {
+  if (fiber.tag !== FunctionComponent && fiber.tag !== SimpleMemoComponent) {
+    throw new Error(
+      'Unknown Fiber. Needs to be a function component to inspect hooks.',
+    );
+  }
+  // Warm up the cache so that it doesn't consume the currentHook.
+  getPrimitiveStackCache();
+  let type = fiber.type;
+  let props = fiber.memoizedProps;
```
gaearon: Does this have default props resolved?

bvaughn:
> Ideally we want this in DevTools but it requires so much implementation details that I don't think we'll want to maintain this with multi-version support. We'll probably have to move to snapshot versions of this package in DevTools, or just snapshoting all of DevTools.

Related issue: facebook/react-devtools/issues/1214

If we were to rearchitect this package in the way I propose on that issue, would there still be value in releasing `react-debug-tools` as its own package?

> For each primitive hook in the log, process the stack trace by removing any shared ancestors with the root of the inspection function. This ensures that the bottom of the stack is the render function. Also, for that particular primitive, remove the top of the stack that is shared with the stack when that primitive is called stand alone from a function component. This removes the call to the primitive itself.

Nice! 👏

scripts/rollup/results.json
```diff
@@ -4,29 +4,29 @@
       "filename": "react.development.js",
       "bundleType": "UMD_DEV",
       "packageName": "react",
-      "size": 95936,
-      "gzip": 25258
+      "size": 98406,
+      "gzip": 25783
```
bvaughn: Revert this file?

packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.internal.js
bvaughn: Would be nice to have some basic testing of the stack parsing+intersection logic too.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -0,0 +1,530 @@
...
+      Dispatcher.useContext(({_currentValue: null}: any));
+      Dispatcher.useState(null);
+      Dispatcher.useReducer((s, a) => s, null);
+      Dispatcher.useRef(null);
+      Dispatcher.useMutationEffect(() => {});
+      Dispatcher.useLayoutEffect(() => {});
+      Dispatcher.useEffect(() => {});
+      Dispatcher.useImperativeMethods(undefined, () => null);
+      Dispatcher.useCallback(() => {});
+      Dispatcher.useMemo(() => null);
```
bvaughn: Would be nice if we had an automated check to ensure that the set of fake hooks here stays in sync with the set of real ones.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -0,0 +1,530 @@
...
+    primitive: 'Context',
+    stackError: new Error(),
+    value: context._currentValue,
+  });
+  return context._currentValue;
+}
+
+function useState<S>(
+  initialState: (() => S) | S,
+): [S, Dispatch<BasicStateAction<S>>] {
```
bvaughn: Would be nice if we had an automated check to ensure these hook signatures stayed in-sync with the real ones.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -0,0 +1,530 @@
...
+      Dispatcher.useState(null);
+      Dispatcher.useReducer((s, a) => s, null);
+      Dispatcher.useRef(null);
+      Dispatcher.useMutationEffect(() => {});
+      Dispatcher.useLayoutEffect(() => {});
+      Dispatcher.useEffect(() => {});
+      Dispatcher.useImperativeMethods(undefined, () => null);
+      Dispatcher.useCallback(() => {});
+      Dispatcher.useMemo(() => null);
+    } finally {
```
bvaughn: Why do we swallow errors in this way? An error in an given hook would block parsing of all subsequent hooks.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -0,0 +1,530 @@
...
+  return primitiveStackCache;
+}
+
+let currentHook: null | Hook = null;
+function nextHook(): null | Hook {
+  let hook = currentHook;
+  if (hook !== null) {
+    currentHook = hook.next;
+  }
+  return hook;
```
bvaughn: Seems a little confusing that a function named `nextHook` returns the current hook 😄

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -0,0 +1,530 @@
...
+// E.g. we can't assume that the root call shares all frames with the stack
+// of a hook call. A simple way to demonstrate this is wrapping `new Error()`
+// in a wrapper constructor like a polyfill. That'll add an extra frame.
+// Similar things can happen with the call to the dispatcher. The top frame
+// may not be the primitive. Likewise the primitive can have fewer stack frames
+// such as when a call to useState got inlined to use dispatcher.useState.
+//
+// We also can't assume that the last frame of the root call is the same
+// frame as the last frame of the hook call because long stack traces can be
+// truncated to a stack trace limit.
```
bvaughn: This comment is great. ^

Would be nice to have some tests for these cases.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -0,0 +1,530 @@
...
+  let current = fiber;
+  while (current) {
+    if (current.tag === ContextProvider) {
+      const providerType: ReactProviderType<any> = current.type;
+      const context: ReactContext<any> = providerType._context;
+      if (!contextMap.has(context)) {
+        // Store the current value that we're going to restore later.
+        contextMap.set(context, context._currentValue);
+        // Set the inner most provider value on the context.
+        context._currentValue = current.memoizedProps.value;
```
bvaughn: Why not just read the value from `contextMap` directly in `readContext` and `useContext` (rather than temporarily overriding `context._currentValue`)?

scripts/rollup/results.json
gaearon: We should just delete it and stop generating it on local builds :P

packages/react-debug-tools/src/ReactDebugHooks.js
sebmarkbage: We probably need a Flow type for this. There's not only one real one. There are at least three (Fiber, old SSR, new SSR).

packages/react-debug-tools/src/ReactDebugHooks.js
sebmarkbage: The type should do that https://github.com/facebook/react/blob/master/packages/react/src/ReactCurrentOwner.js#L25

packages/react-debug-tools/src/ReactDebugHooks.js
sebmarkbage: We don't swallow errors. This is a finally call.

packages/react-debug-tools/src/ReactDebugHooks.js
sebmarkbage: The contextMap stores the shadowed value, not the current one.

It's nice to use the same mechanism as Fiber because it allows the inspectHook thing to be called in a context where it context has already been set up with some other mechanism.

packages/react-debug-tools/src/ReactDebugHooks.js
gaearon: And shallow one!

packages/react-debug-tools/src/ReactDebugHooks.js
sophiebits: @bvaughn We only set `primitiveStackCache` if it doesn't throw.

bvaughn: 🤪 Right.

a-x-:
What is `useInspect`? there is not any useInspect in react source, google also don't know anything about it

--
Why not to create simple babel-transformation, that will be grab assosiated with hooks const-names?

bvaughn:
@a-x- The hook Sebastian was referring to was later added as `useDebugValue` (#14559)

PR #14095, opened 2018-11-05
Add back early return in setOffsets which was removed in 16.5.0

einarq:
The scenario is that we have a TinyMCE editor (which uses an iframe, which the relevant code changes in #13650 made changes related to), and this editor is shown in every row in a list component.
The editor supports pasting to add more rows, and when pasting a lot of items (100+), we get "Object expected", due to getSelection not being present on the Window object. Seems strange obviously, but there could be some security issue with Edge or something? Not sure.

Not sure how to test this so didn't add any unit tests for it. The early return existed pre 16.5.0, this PR adds it back.


gaearon:
Looks good. Thanks.

PR #13088, opened 2018-06-21
Fix crash during server render in react 16.4.1.

dustinsoftware:
This fixes a crash that only occurs in react-dom 16.4.1.

setTimeout and clearTimeout may not be available in some server-render environments (such as ChakraCore in React.NET), and loading ReactScheduler.js will cause a crash unless the existence of the variables are checked via a typeof comparison.

https://github.com/reactjs/React.NET/issues/555

The crash did not occur in 16.4.0, and the change appears to have been introduced here: https://github.com/facebook/react/pull/12931/files#diff-bbebc3357e1fb99ab13ad796e04b69a6L47

I tested this by using yarn link and running it with a local copy of React.NET. I am unsure the best way to unit test this change, since assigning null to `setTimeout` causes an immediate crash within the Node REPL.

PR #14103, opened 2018-11-05
Fix crash during server render in React 16.6.0

tnunes:
This fixes a bug introduced in React 16.6.0, which was not present in 16.5.2.

Runtimes without `setTimeout` and `clearTimeout` (e.g. ClearScript for .NET) are now crashing just by importing `react-dom`.

This replicates the fix for `scheduler` introduced in #13088 for `react-dom`, by checking for existence of `setTimeout` and `clearTimeout` in the runtime before using them.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -284,8 +284,13 @@ export function createTextInstance(
 }
 
 export const isPrimaryRenderer = true;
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks

tnunes:
Thank you @gaearon. Any ETA on when 16.6.1 (and the next 16.7 alpha) will be released?

gaearon:
Soon-ish

Issue #13820, opened 2018-10-10
componentDidCatch doesn't catch some invariants


mlityn:
**Do you want to request a *feature* or report a *bug*?**

A bug.

**What is the current behavior?**

componentDidCatch isn't called when a child component renders a void element containing another element. The app just crashes.
If it contains just text, the same error ("...is a void element tag and must neither have children nor use dangerouslySetInnerHTML") appears, but this time with componentDidCatch properly called.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

https://codesandbox.io/s/0042z5wrlp
The behavior is different when you change Comp1 with Comp2.

**What is the expected behavior?**

componentDidCatch should be called in both cases.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

16.5.2. 
In 16.4.2 it worked correctly.

askeroff:
Might it be because the Error Boundaries was intended to be used as 

```javascript
<ErrorBoundary>
  <ChildrenComponent>
</ErrorBoundary>
```

The way try-catch block works, that's what I inferred from documentation. After changing your example to the way as I described above it works and componentDidCatch is being called.

```javascript
import React from "react";
import ReactDOM from "react-dom";

const Comp1 = () => <input>some text</input>;
const Comp2 = () => (
  <input>
    <div />
  </input>
);

class App extends React.Component {
  state = { error: false };
  componentDidCatch() {
    this.setState({ error: true });
  }
  render() {
    return this.state.error ? <p>ERROR</p> : this.props.children;
  }
}

const rootElement = document.getElementById("root");
ReactDOM.render(
  <App>
    <Comp1 />
  </App>,
  rootElement
);


```

Additional note: there is a error-overlay, that is only visible in development, if you close it you will see your custom error message.

romainwn:
Hi, I would like to give a try to this issue !
Unfortunately, I'm used to React but not to the project itself, any idea where I should start looking ?

krzysztofla:
I also would like to know where to take a look :)

romainwn:
Humm, that is weird. When I tried the codesandbox example on a Windows a few hours earlier, I was able to reproduce the issue.
[...]

PR #12201, opened 2018-02-09
Add stack unwinding phase for handling errors

acdlite:
A rewrite of error handling, with semantics that more closely match stack unwinding.

Errors that are thrown during the render phase unwind to the nearest error boundary, like before. But rather than synchronously unmount the children before retrying, we restart the failed subtree within the same render phase. The failed children are still unmounted (as if all their keys changed) but without an extra commit.

Commit phase errors are different. They work by scheduling an error on the update queue of the error boundary. When we enter the render phase, the error is popped off the queue. The rest of the algorithm is the same.

This approach is designed to work for throwing non-errors, too, though that feature is not implemented yet.

Known differences from existing error handling semantics in synchronous mode:

- When a component throws an error, we continue rendering its siblings rather than unwind immediately.
- Because of the previous point, a single error boundary may capture more than one error during a single render, if multiple child branches fail.
- We call `componentDidCatch` for each error that a boundary captures, instead of just the first one.
- Error boundaries only catch error objects. Non-errors will propagate to the root and cause the tree to unmount, as if it’s an uncaught error.

Unresolved items:

- [x] Add `getDerivedStateFromCatch` lifecycle behind a feature flag. This is fired during the render phase so that we can retry without committing, as opposed to `componentDidCatch`, which fires in the commit phase. We'll keep `componentDidCatch` around for logging purposes. If `getDerivedStateFromCatch` is not implemented but `componentDidCatch` is, the fallback behavior is to unmount all the boundary's children. (This part is already implemented.) We may warn about the missing `getDerivedStateFromCatch` in a future release.
- [x] Existing code that schedules recovery updates inside `componentDidCatch` will have mostly the same semantics, with one key exception. We no longer track whether an error boundary failed in a previous commit. So it's now possible to fall into a recursive error loop, in the same way as an unguarded `setState` inside `componentDidUpdate`.

PR #14104, opened 2018-11-05
Recover from errors with a boundary in completion phase

gaearon:
Fixes https://github.com/facebook/react/issues/13820.

Dunno if this is an "idiomatic" fix. The issue is that we null out `nextUnitOfWork` as we traverse up during completion, but we also don't attempt to find an error boundary if `nextUnitOfWork` is null:

https://github.com/facebook/react/blob/b305c4e034bbb3b13df2028c0503a84f91e57455/packages/react-reconciler/src/ReactFiberScheduler.js#L1280-L1284

So errors in completion phase don't look for boundaries at all right now.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -1299,9 +1299,9 @@ function renderRoot(root: FiberRoot, isYieldy: boolean): void {
           (resetCurrentlyProcessingQueue: any)();
         }
 
-        if (!wasCompleting) {
-          const failedUnitOfWork: Fiber = nextUnitOfWork;
-          if (__DEV__ && replayFailedUnitOfWorkWithInvokeGuardedCallback) {
+        if (__DEV__ && replayFailedUnitOfWorkWithInvokeGuardedCallback) {
+          if (!wasCompleting) {
+            const failedUnitOfWork: Fiber = nextUnitOfWork;
             replayUnitOfWork(failedUnitOfWork, thrownValue, isYieldy);
           }
         }
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -254,6 +254,7 @@ let isWorking: boolean = false;
 // The next work in progress fiber that we're currently working on.
 let nextUnitOfWork: Fiber | null = null;
 let nextRoot: FiberRoot | null = null;
+let nextCompletingUnitOfWork: Fiber | null = null;
 // The time at which we're currently rendering work.
 let nextRenderExpirationTime: ExpirationTime = NoWork;
 let nextLatestAbsoluteTimeoutMs: number = -1;
@@ -947,10 +948,8 @@ function completeUnitOfWork(workInProgress: Fiber): Fiber | null {
     const siblingFiber = workInProgress.sibling;
 
     if ((workInProgress.effectTag & Incomplete) === NoEffect) {
-      // Prepare this field so we can find an error boundary in case completing throws.
-      nextUnitOfWork = workInProgress;
-
       // This fiber completed.
+      nextCompletingUnitOfWork = workInProgress;
       if (enableProfilerTimer) {
         if (workInProgress.mode & ProfileMode) {
           startProfilerTimer(workInProgress);
@@ -973,6 +972,8 @@ function completeUnitOfWork(workInProgress: Fiber): Fiber | null {
           nextRenderExpirationTime,
         );
       }
+      nextCompletingUnitOfWork = null;
+
       stopWorkTimer(workInProgress);
       resetChildExpirationTime(workInProgress, nextRenderExpirationTime);
       if (__DEV__) {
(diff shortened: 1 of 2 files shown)


gaearon:
Hmm. I was too confident. This causes some infinite loops.

gaearon:
Pushed a different fix. This doesn't fail any tests but is awkward.

gaearon:
This is less awkward. Ready for review. Still don't know if it's right approach.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1288,9 +1299,11 @@ function renderRoot(root: FiberRoot, isYieldy: boolean): void {
...
         }
 
-        const failedUnitOfWork: Fiber = nextUnitOfWork;
-        if (__DEV__ && replayFailedUnitOfWorkWithInvokeGuardedCallback) {
-          replayUnitOfWork(failedUnitOfWork, thrownValue, isYieldy);
+        if (!wasCompleting) {
+          const failedUnitOfWork: Fiber = nextUnitOfWork;
+          if (__DEV__ && replayFailedUnitOfWorkWithInvokeGuardedCallback) {
+            replayUnitOfWork(failedUnitOfWork, thrownValue, isYieldy);
+          }
```
blixt: Does this mean that the bug only happens when in DEV?

Also, couldn't this code be combined into the following block to avoid implying that the conditions/variables have any effect in production?

```typescript
        if (__DEV__) {
          // Reset global debug state
          // We assume this is defined in DEV
          (resetCurrentlyProcessingQueue: any)();

          if (!wasCompleting && replayFailedUnitOfWorkWithInvokeGuardedCallback) {
            const failedUnitOfWork: Fiber = nextUnitOfWork;
            replayUnitOfWork(failedUnitOfWork, thrownValue, isYieldy);
          }
        }
```

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -953,22 +954,30 @@ function completeUnitOfWork(workInProgress: Fiber): Fiber | null {
...
+        isCompleting = false;
 
         if (workInProgress.mode & ProfileMode) {
           // Update render duration assuming we didn't error.
           stopProfilerTimerIfRunningAndRecordDelta(workInProgress, false);
         }
       } else {
+        // Remember we're completing this unit so we can find a boundary if it fails.
+        isCompleting = true;
+        nextUnitOfWork = workInProgress;
```
blixt: Ineffective assignment?

packages/react-reconciler/src/ReactFiberScheduler.js
gaearon: I pushed another commit that clarifies which parts of the fix are DEV-only.

acdlite:
Rephrased, here's the alternate fix I'm suggesting. Either:

- Inside `completeUnitOfWork`, replace all occurrences of `workInProgress` with `nextUnitOfWork`.
- Or, always set `nextUnitOfWork = workInProgress` before calling `completeWork`.

gaearon:
>Or, always set nextUnitOfWork = workInProgress before calling completeWork.

Isn’t that what I’m doing?

acdlite:
Oh right, I read the first few commits and then didn't realize the subsequent ones changed it. :D

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1277,6 +1287,11 @@ function renderRoot(root: FiberRoot, isYieldy: boolean): void {
       resetContextDependences();
       resetHooks();
 
+      // Reset in case completion throws.
+      // This is only used in DEV and when replaying is on.
+      const mayReplay = mayReplayFailedUnitOfWork;
+      mayReplayFailedUnitOfWork = true;
```
acdlite: I don't love this boolean, especially because it's not dev only. But I'm more concerned that this replay logic is getting really convoluted and it seems there must be a better way to structure it.

Maybe we can move where `replayUnitOfWork` is called instead. If `replayUnitOfWork` is only called after a failed begin phase, we could fork `beginWork` to `beginWorkInDEV` which has a local try/catch.

```js
function beginWorkInDEV(current, workInProgress, renderExpirationTime) {
  try {
    beginWork(current, workInProgress, renderExpirationTime);
  } catch (e) {
    resetContextDependences();
    resetHooks();

    // Either call replayUnitOfWork or inline the whole thing here

    // Then throw to outer catch block
    throw e;
  }
}
```

There might be other implications that I'm forgetting, but I think we should give this a shot before piling more complexity onto this code path.

gaearon: >especially because it's not dev only

It is though. I think it would get DCE’d because we never read it outside DEV. The only reason I write to it outside a flag is just because it seems awkward to do extra wrapping because of scoping.

acdlite: Just checked — this is what I originally did, but @sebmarkbage asked that I move the extra `try` out of the hot path: https://github.com/facebook/react/pull/12201#discussion_r170399852

I feel we should reconsider given the additional complexity of distinguishing between the begin and complete phases.

gaearon: Pushed another commit that makes it clearer that part is DEV-only too. I'll look at your suggestion tomorrow/

sebmarkbage:
> let's revisit soon to see if there's a better way factor it.

let's

PR #14106, opened 2018-11-05
Fix the error message to reflect the method name

gaearon:
The method suggested in this error message doesn't exist. It tripped me more than once.

Does it get renamed somewhere? We should either fix the error message or fix the name.


bvaughn:
It is `unstable_toHaveYielded` for open source users. See here:
https://github.com/facebook/react/blob/master/packages/jest-react/src/JestReact.js#L57

We removed the "unstable_" prefix for ourselves for convenience here:
https://github.com/facebook/react/blob/master/scripts/jest/setupTests.js#L54

gaearon:
Ah. Okay. I can remember this.

PR #14109, opened 2018-11-05
Rename .internal tests that aren't using internals

sophiebits:


sophiebits:
There's ReactDOMSuspensePlaceholder but I left it because #14083 will add a feature flag usage back.

PR #14108, opened 2018-11-05
[WIP] Failing tests for lazy() not passing defaultProps

gaearon:
Based on https://github.com/facebook/react/pull/13961 by @aweary. Will fix https://github.com/facebook/react/issues/139609 (this is just failing tests for now).

I removed the incomplete fix and started by adding more test coverage. I haven't looked at how to fix this yet.

PR #14112, opened 2018-11-05
Fix lazy() with defaultProps

gaearon:
Builds on failing tests from https://github.com/facebook/react/pull/14108.
Fixes https://github.com/facebook/react/issues/13960.

I might have missed something. Just pushing it before going home. I haven't thought about the perf impact of this. Maybe we need to change the strategy and store them?

### DIFF
--- a/packages/react-reconciler/src/ReactFiberLazyComponent.js
+++ b/packages/react-reconciler/src/ReactFiberLazyComponent.js
@@ -12,7 +12,7 @@ import type {LazyComponent, Thenable} from 'shared/ReactLazyComponent';
 import {Resolved, Rejected, Pending} from 'shared/ReactLazyComponent';
 import warning from 'shared/warning';
 
-export function resolveDefaultProps(Component, baseProps) {
+export function resolveDefaultProps(Component: any, baseProps: Object): Object {
   if (Component && Component.defaultProps) {
     // Resolve default props. Taken from ReactElement
     const props = Object.assign({}, baseProps);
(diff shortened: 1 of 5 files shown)


gaearon:
Btw this probably means DevTools would also be wrong since they read off `memoizedProps`. And Enzyme & co.

packages/react-reconciler/src/ReactFiberLazyComponent.js
```diff
@@ -12,6 +12,21 @@ import type {LazyComponent, Thenable} from 'shared/ReactLazyComponent';
 import {Resolved, Rejected, Pending} from 'shared/ReactLazyComponent';
 import warning from 'shared/warning';
 
+export function resolveDefaultProps(Component, baseProps) {
```
sebmarkbage: I think this is not just used for lazy.

sebmarkbage: Even if it was this would be the wrong place for it.

sebmarkbage: I don’t like shared. I prefer copying to putting things in shared. Makes it impossible to special case and encourages abstracting.

sebmarkbage:
It made sense to do this lazily when we only did this on render. But now we do the work multiple times. Not sure that makes sense.

packages/react-reconciler/src/ReactFiberLazyComponent.js
gaearon: So where do you want it? I guess I can create a file in the reconciler package but I don’t see why it deserves that when conceptually it’s related to our design for lazy.

gaearon: I kept it in `LazyComponent` for now because I don't want to duplicate it in reconciler bundle twice.

gaearon:
>This should not be called all the time. Make it conditional on type not equaling elementType.

Pushed.

PR #14115, opened 2018-11-05
Implement {,un}hideInstance on RN renderer

sophiebits:
This is required to use lazy.

Test Plan:
* Verified lazy works on a real world use case (shows spinner, shows real content).
* Verified that if I change the primary content's styles to have `display: 'none'` then it never appears (i.e., the code in `unhide` reads the styles successfully)

### DIFF
--- a/packages/react-native-renderer/src/ReactFabricHostConfig.js
+++ b/packages/react-native-renderer/src/ReactFabricHostConfig.js
@@ -381,7 +381,16 @@ export function cloneHiddenInstance(
   props: Props,
   internalInstanceHandle: Object,
 ): Instance {
-  throw new Error('Not yet implemented.');
+  const viewConfig = instance.canonical.viewConfig;
+  const node = instance.node;
+  const updatePayload = ReactNativeAttributePayload.create(
+    {style: {display: 'none'}},
+    viewConfig.validAttributes,
+  );
+  return {
+    node: cloneNodeWithNewProps(node, updatePayload),
+    canonical: instance.canonical,
+  };
 }
 
 export function cloneUnhiddenInstance(
@@ -390,7 +399,17 @@ export function cloneUnhiddenInstance(
   props: Props,
   internalInstanceHandle: Object,
 ): Instance {
-  throw new Error('Not yet implemented.');
+  const viewConfig = instance.canonical.viewConfig;
+  const node = instance.node;
+  const updatePayload = ReactNativeAttributePayload.diff(
+    {...props, style: [props.style, {display: 'none'}]},
+    props,
+    viewConfig.validAttributes,
+  );
+  return {
+    node: cloneNodeWithNewProps(node, updatePayload),
+    canonical: instance.canonical,
+  };
 }
 
 export function createHiddenTextInstance(
--- a/packages/react-native-renderer/src/ReactNativeHostConfig.js
+++ b/packages/react-native-renderer/src/ReactNativeHostConfig.js
@@ -454,7 +454,16 @@ export function resetTextContent(instance: Instance): void {
 }
 
 export function hideInstance(instance: Instance): void {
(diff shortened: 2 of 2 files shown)


sebmarkbage:
Hm. We shouldn't assume that the flattening happens the way it does today. There is a configuration layer that determines that. Perhaps we can call `{ style: {display: ....} }` to ReactNativeAttributePayload somehow?

I was hoping that we'd be able to batch these into the normal properties update call so there is only one updateView call. I suspect this can lead to subtle bugs depending on the order of these updateView calls not getting flipped and it'll cause two updates where one is expected.

However, eventually this will all go away with Fabric.

We should add a fix to Fabric too at the same time so they line up. That one might be more straightforward.

sophiebits:
You mean it's OK to hardcode {style: {display}} but not {display}? I can change that.

sebmarkbage:
Yes. I'm uneasy with double updateView calls. But I think I'm ok with it, considering it's legacy in a Fabric world. So the only action items are 1) avoid the hard coding 2) fix fabric.

packages/react-native-renderer/src/ReactNativeHostConfig.js
```diff
@@ -454,15 +454,29 @@ export function resetTextContent(instance: Instance): void {
...
+  const viewConfig = instance.viewConfig;
+  UIManager.updateView(
+    instance._nativeTag,
+    viewConfig.uiViewClassName,
+    {display: 'none'},
+  );
 }
 
 export function hideTextInstance(textInstance: TextInstance): void {
   throw new Error('Not yet implemented.');
```
sebmarkbage: Can this happen? E.g. returning a string from render?

sophiebits: if you do

```
<Text><Suspense>{string}</Suspense></Text>
```

I think so. it seems weird, I don't know if this is complicated, and we never do this right now so I opted to skip it

sophiebits:
This OK? I don't love the diff call in unhide but I can't think of a correct way to do it better.

packages/react-native-renderer/src/ReactNativeHostConfig.js
sebmarkbage: I believe these are just like any other shadow node inside text from the native's perspective so I think we could send the display: none property here too.

cc @shergin

packages/react-native-renderer/src/ReactFabricHostConfig.js
```diff
@@ -390,7 +398,16 @@ export function cloneUnhiddenInstance(
   props: Props,
   internalInstanceHandle: Object,
 ): Instance {
-  throw new Error('Not yet implemented.');
+  const node = instance.node;
+  const updatePayload = ReactNativeAttributePayload.diff(
+    {...props, style: [props.style, {display: 'none'}]},
```
sebmarkbage: What does spread compile to?

sophiebits:
```
Object.assign({}, props, { style: [props.style, { display: "none" }] }),
```

which is polyfilled https://github.com/facebook/react-native/blob/cc90c20bf60c72fc1b42f3306636847ce7f478e5/Libraries/polyfills/Object.es6.js#L15

packages/react-native-renderer/src/ReactNativeHostConfig.js
shergin: I believe currently the only most outer `<Text>` supports `display: none`, but I think that's incorrect behavior (and I will fix it).

sebmarkbage: I believe in the DOM version we actually just set the string to empty string `""` while hidden and then reset it to its current value.

Seems like we could do the same for RN without needing to fix the native side.

PR #5997, opened 2016-02-08
Fix #5929. Resolve to default props when config key is set to undefined in cloneElement

truongduy134:
Fix #5929. Resolve to default props when config key is set to undefined in cloneElement

```
var Foo = React.createClass({
   getDefaultProps: function() {
      return {foo: 'bar'};
   },
   render: function() {
      return React.createElement('span', null, this.props.prop);
   },
});

clonedElement = React.cloneElement(React.createElement(Foo, {foo: 'foo'}), {foo: undefined});
console.log(element.props.foo); // Should be 'bar' instead of undefined
```

PR #14117, opened 2018-11-06
Reuse logic between createElement and cloneElement

jdlehman:
I have been going through the codebase to better understand how React works under the hood, when I noticed that `createElement` and `cloneElement` appeared to handle setting props differently, which didn't make sense to me based on what I know about React. After looking closer, I realized that they both actually do the same thing, but the difference in logic threw me off a bit at first glance.

This PR adds a minor refactor to reuse the shared logic between `createElement` and `cloneElement` so that it is more explicit that the props are set in the same way.

### DIFF
--- a/packages/react/src/ReactElement.js
+++ b/packages/react/src/ReactElement.js
@@ -337,6 +337,11 @@ export function cloneElement(element, config, children) {
     for (let i = 0; i < childrenLength; i++) {
       childArray[i] = arguments[i + 2];
     }
+    if (__DEV__) {
+      if (Object.freeze) {
+        Object.freeze(childArray);
+      }
+    }
     props.children = childArray;
   }
 
--- a/packages/react/src/ReactElement.js
+++ b/packages/react/src/ReactElement.js
@@ -88,6 +88,25 @@ function defineRefPropWarningGetter(props, displayName) {
   });
 }
 
+function setProps(config, props) {
+  for (let propName in config) {
+    if (
+      hasOwnProperty.call(config, propName) &&
+      !RESERVED_PROPS.hasOwnProperty(propName)
+    ) {
+      props[propName] = config[propName];
+    }
+  }
+}
+
+function setDefaultProps(defaultProps, props) {
+  for (let propName in defaultProps) {
+    if (props[propName] === undefined) {
+      props[propName] = defaultProps[propName];
+    }
+  }
+}
+
 /**
  * Factory method to create a new React element. This no longer adheres to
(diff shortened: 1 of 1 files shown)


jdlehman:
I also wanted to make a similar refactor around the logic for copying `children` to props, but I was afraid of causing a deoptimization since that particular code relies on `arguments`. If anyone has suggestions on a better way to handle this logic without deoptimizing the performance, it could be a good candidate to come through with this change as well.

packages/react/src/ReactElement.js
```diff
@@ -321,23 +324,7 @@ export function cloneElement(element, config, children) {
     }
 
     // Remaining properties override existing props
-    let defaultProps;
-    if (element.type && element.type.defaultProps) {
-      defaultProps = element.type.defaultProps;
-    }
-    for (propName in config) {
```
jcs090218: I am a bit confuse and worry that these lines will not produce the same result as `setProps` function do?

jdlehman: This is where I initially thought they were doing something different as well. This code was actually added later to fix a bug with default props (undefined values were not falling back to default props) that only existed in `cloneElement` via https://github.com/facebook/react/pull/5997/files.

The real difference structurally between how it was and how it is in this PR is when default props are set (if needed). Currently they are set while setting props, but before setting children on props. The order ultimately doesn't matter because default props will only be set when a value is undefined and clone element. We can also see from the tests passing that this functionality still behaves as expected.

Though actually, from playing with this more I this difference is an inconsistency when `children` is explicitly set to `undefined` and the component has a default prop for `children`. See the issue I just opened: https://github.com/facebook/react/issues/14136.

I'll add a test case depending on the outcome of that issue.

packages/react/src/ReactElement.js
```diff
@@ -88,6 +88,25 @@ function defineRefPropWarningGetter(props, displayName) {
...
+      !RESERVED_PROPS.hasOwnProperty(propName)
+    ) {
+      props[propName] = config[propName];
+    }
+  }
+}
+
+function setDefaultProps(defaultProps, props) {
+  for (let propName in defaultProps) {
+    if (props[propName] === undefined) {
```
sdevalapurkar: question/nit: could this just be
```suggestion
    if (props[propName] == undefined) {
```

jdlehman: I think here we just care if the prop value is `undefined`, otherwise explicitly setting a prop to `null` would fallback to the default prop since `null == undefined` is `true`. (perhaps it might be more clear to do `typeof props[propName] === "undefined"`)

PR #14118, opened 2018-11-06
Improve hydration fixture, support older versions of React

nhunzaker:
**What**

This PR makes some improvements to the DOM test fixtures to make it easier to test older versions of React.

Here's a link:
http://react-hydration-fixture.surge.sh/hydration

**Why**

After trying to figure out the origin of a few DOM bugs recently, it gave me some motivation to figure out how we can make that faster. I wanted to quickly test when a bug started or stopped, but discovered my only option was a lot of npm install + build/test across commits.

I think we can do that better, and I think the hydration fixture provides a good foundation for that. If we can figure out a way to host commit-level builds, we could very quickly pass through a range of commits to triage an issue in the browser. 

**Updates**

- Updating the App component to use a class instead of a function lets us support all fixtures down to 0.11
- For really old versions, I generalized the React version picker and added it to the hydration fixture. It can now override the current version of React and supports down to React 0.4.0
- Create a wrapper around findDOMNode for older React
- A small UX fix for the error message on the Hydration Fixture
- Additional documentation in the fixtures as to when key API changes happened. 

I don't want to support all the way back to 0.4.0 for all test fixtures, but I think having some utilities to do it quickly would be useful.

---

image

### DIFF
--- a/fixtures/dom/public/renderer.js
+++ b/fixtures/dom/public/renderer.js
@@ -13,6 +13,30 @@
   var renders = 0;
   var failed = false;
 
+  var needsReactDOM = getBooleanQueryParam('needsReactDOM');
+  var needsCreateElement = getBooleanQueryParam('needsCreateElement');
+
+  function unmountComponent(node) {
+    // ReactDOM was moved into a separate package in 0.14
+    if (needsReactDOM) {
+      ReactDOM.unmountComponentAtNode(node);
+    } else if (React.unmountComponentAtNode) {
+      React.unmountComponentAtNode(node);
+    } else {
+      // Unmounting for React 0.4 and lower
+      React.unmountAndReleaseReactRootNode(node);
+    }
+  }
+
+  function createElement(value) {
+    // React.createElement replaced function invocation in 0.12
+    if (needsCreateElement) {
+      return React.createElement(value);
+    } else {
+      return value();
+    }
+  }
+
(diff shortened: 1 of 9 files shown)


sophiebits:
One idea I've thought about that you may be interested in investigating: we may be able to use [Metro](https://facebook.github.io/metro/), the packager used by RN, to serve bundles of React dynamically from our repo. A lot of its configuration overlaps with Jest's which is why I'm hopeful this could work.

Basically we could end up with a way to start a server that vends built React bundles that fixtures could then depend on – they would not be exactly identical to our Rollup process, but they would likely be much faster and could support incremental building. (Metro has very fast reload times – so when switching between commits only a few files would rebuild, much like `yarn test MyTestFile --watch`.)

if you're interested in investigating this here is one example of using Metro on the web by one of the members of the team here at FB: https://github.com/rafeca/metro-sample-app.

nhunzaker:
That's very interesting. I can definitely look into this!

nhunzaker:
@philipp-spiess, @jquense, @aweary: it's been a while, but I wanted to see if we could land this on the DOM fixtures. I still plan to see if we could use the metro bundler for quick commit-level builds, however I'd like to close this out if possible and I think these are useful improvements.

nhunzaker:
@philipp-spiess no worries. Thanks ❤️

PR #14119, opened 2018-11-06
Restrict effect return type to a function or nothing

acdlite:
We already warn in dev if the wrong type is returned. This updates the Flow type.

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -147,7 +147,7 @@ function useRef<T>(initialValue: T): {current: T} {
 }
 
 function useLayoutEffect(
-  create: () => mixed,
+  create: () => (() => void) | void,
   inputs: Array<mixed> | void | null,
 ): void {
   nextHook();
@@ -159,7 +159,7 @@ function useLayoutEffect(
 }
 
 function useEffect(
-  create: () => mixed,
+  create: () => (() => void) | void,
   inputs: Array<mixed> | void | null,
 ): void {
   nextHook();
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -320,42 +320,40 @@ function commitHookEffectList(
       if ((effect.tag & unmountTag) !== NoHookEffect) {
         // Unmount
         const destroy = effect.destroy;
-        effect.destroy = null;
-        if (destroy !== null) {
+        effect.destroy = undefined;
+        if (destroy !== undefined) {
           destroy();
         }
       }
       if ((effect.tag & mountTag) !== NoHookEffect) {
         // Mount
         const create = effect.create;
-        let destroy = create();
-        if (typeof destroy !== 'function') {
-          if (__DEV__) {
-            if (destroy !== null && destroy !== undefined) {
-              warningWithoutStack(
-                false,
-                'useEffect function must return a cleanup function or ' +
-                  'nothing.%s%s',
-                typeof destroy.then === 'function'
-                  ? '\n\nIt looks like you wrote useEffect(async () => ...) or returned a Promise. ' +
-                    'Instead, you may write an async function separately ' +
-                    'and then call it from inside the effect:\n\n' +
-                    'async function fetchComment(commentId) {\n' +
-                    '  // You can await here\n' +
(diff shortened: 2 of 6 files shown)


acdlite:
> Also the return value from the destroy function itself should always be void.

That's can be annoying if you use an arrow function. For example, sometimes "destroy" -type functions return a boolean:

```js
useEffect(() => {
  mySet.add(thing);
  return () => mySet.delete(thing);
});
```

sebmarkbage:
`null` is not allowed instead of `undefined` in default props, default args etc. These things have a tendency to spread once they’ve leaked so better to narrow it early.

The argument for arrows in destroy also applies to `useEffect(() => foo());`

Might be better to just enforce a style that makes it clear that these are side-effectful void functions?

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -142,7 +142,7 @@ function useRef<T>(initialValue: T): {current: T} {
 }
 
 function useLayoutEffect(
-  create: () => mixed,
+  create: () => (() => void) | void,
```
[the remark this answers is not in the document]
jamesisaac: @Jessidhia `() => void` enforces a void return in Flow.  The effect you're talking about would be written as `() => mixed` in Flow.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -320,42 +320,40 @@ function commitHookEffectList(
...
-                    'useEffect(() => {\n' +
-                    '  fetchComment(commentId);\n' +
-                    '}, [commentId]);\n\n' +
-                    'In the future, React will provide a more idiomatic solution for data fetching ' +
-                    "that doesn't involve writing effects manually."
-                  : '',
-                getStackByFiberInDevAndProd(finishedWork),
-              );
-            }
+        effect.destroy = create();
```
acdlite: @sebmarkbage Usually we coerce missing values to null before storing them in our internal data structures, but I think that's because we usually accept either, and null is preferred because it's less likely to be unintentional. But in this case, since we don't accept null, I can skip the type check in prod. Let me know if this doesn't make sense.

sebmarkbage: I believe there has been times where V8 has treated undefined as effectively a missing property in the hidden class rather than a reified value. So setting to undefined might mess with the hidden class. Not sure though.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -320,42 +320,40 @@ function commitHookEffectList(
...
+
+        if (__DEV__) {
+          const destroy = effect.destroy;
+          if (destroy !== undefined && typeof destroy !== 'function') {
+            warningWithoutStack(
+              false,
+              'useEffect function must return a cleanup function or ' +
+                'nothing (undefined).%s%s',
```
sebmarkbage: Are there other common examples we can include here? So it doesn't sounds like a jargon/technical error message.

packages/react-reconciler/src/ReactFiberCommitWork.js
acdlite: Ok I'll leave it like this until we learn more, I suppose

packages/react-reconciler/src/ReactFiberCommitWork.js
acdlite: I'll add a branch specifically for `null`

acdlite:
Ok the warning message now has three branches, based on the type of the return value.

### If you return null

> An Effect function must not return anything besides a function, which is used for clean-up. You returned null. If your effect does not require clean up, return undefined (or nothing).

### If you return a promise

> An Effect function must not return anything besides a function, which is used for clean-up.
>
> It looks like you wrote useEffect(async () => ...) or returned a Promise.
Instead, you may write an async function separately and then call it from inside
the effect:
>
> ```
> async function fetchComment(commentId) {
>   // You can await here
> }
>
> useEffect(() => {
>   fetchComment(commentId);
> }, [commentId]);
> ```
> 
> In the future, React will provide a more idiomatic solution for data fetching that doesn't involve writing effects manually.

### If you return anything else that's neither a function nor undefined

> An Effect function must not return anything besides a function, which is used for clean-up. You returned: [value coerced to string]

***

I'll merge and we can bikeshed the messages more if they still need work.

PR #14124, opened 2018-11-06
fix typo & DCE

NE-SmallTown:


packages/react-dom/src/server/escapeTextForBrowser.js
```diff
@@ -56,7 +56,7 @@ function escapeHtml(string) {
 
   let escape;
   let html = '';
-  let index = 0;
```
NE-SmallTown: Below we do index = match.index, so it doesn't need assign here

scripts/rollup/shims/react-native/NativeMethodsMixin.js
```diff
@@ -14,7 +14,7 @@ const {
   __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED,
 } = require('ReactNative');
 
-import type {NativeMethodsMixinType} from 'ReactNativeTypes';
```
gaearon: These (and below) are wrong because they're written to be copied into RN repo.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -881,7 +881,6 @@ function unmountHostComponents(current): void {
         }
         parent = parent.return;
       }
-      currentParentIsValid = true;
```
NE-SmallTown: Every switch case we do `break`, so this line is unreachable

scripts/rollup/shims/react-native/NativeMethodsMixin.js
NE-SmallTown: Oh, don't know that, could we add some comments about that, I think that would be better?

NE-SmallTown: Done. Remove this and below

packages/react-reconciler/src/ReactFiberCommitWork.js
sophiebits: This looks reachable to me. The `break` takes us to exactly this line.

PR #14128, opened 2018-11-06
Add DEV-only checks for assumption about instance properties

gaearon:
I started depending on this in https://github.com/facebook/react/pull/14112. Adding some DEV-only checks so that we notice if we break this.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -54,6 +54,7 @@ import {
 import getComponentName from 'shared/getComponentName';
 import invariant from 'shared/invariant';
 import warningWithoutStack from 'shared/warningWithoutStack';
+import warning from 'shared/warning';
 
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -352,6 +369,22 @@ function commitLifeCycles(
           // We could update instance props and state here,
           // but instead we rely on them being set during last render.
           // TODO: revisit this when we implement resuming.
+          if (__DEV__) {
+            if (finishedWork.type === finishedWork.elementType) {
+              warning(
+                instance.props === finishedWork.memoizedProps,
```
acdlite: I thought the whole point of your previous PR is that `memoizedProps` does not include the resolved default props, but `instance.props` does?

acdlite: Oh duh, didn't notice the outer check :D

PR #14130, opened 2018-11-06
[WIP] Refactor beginWork replay

gaearon:
**WIP, doesn’t work**

I want to follow up on https://github.com/facebook/react/pull/14104#discussion_r231288310 and try to simplify this logic.

So far, I've tried to separate `beginWorkInDEV` that would call `beingWork` and fall back to `replay` in a `catch` block. A dozen tests are currently failing with a context mismatch or something like this, but I’m not sure why. I also realized that there might be a reason replaying calls `workLoop` rather than `beginWork` directly (which seems like it would be simpler?) — for example, `beginWork` might assume the profiler timer has already started. So we need to make sure that any initialization that happens prior to `beginWork` also happens on the replay code path, or revert to calling `workLoop`.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -275,19 +275,17 @@ let interruptedBy: Fiber | null = null;
 
 let stashedWorkInProgressProperties;
 let replayUnitOfWork;
-let mayReplayFailedUnitOfWork;
 let isReplayingFailedUnitOfWork;
 let originalReplayError;
 let rethrowOriginalError;
 if (__DEV__ && replayFailedUnitOfWorkWithInvokeGuardedCallback) {
   stashedWorkInProgressProperties = null;
-  mayReplayFailedUnitOfWork = true;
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1193,13 +1191,40 @@ function workLoop(isYieldy) {
...
+  workInProgress: Fiber,
+  renderExpirationTime: ExpirationTime,
+  isYieldy: boolean,
+): Fiber | null {
+  try {
+    return beginWork(current, workInProgress, renderExpirationTime);
+  } catch (thrownValue) {
+    resetContextDependences();
+    resetHooks();
+    if (nextUnitOfWork === null) {
```
acdlite: I don't think there's any way to hit this condition. `nextUnitOfWork` should always be `workInProgress`.

gaearon:
@acdlite Do you still care that I follow up on this refactor?

acdlite:
I addressed this when I rewrote the work loop last year so we can close this

PR #14133, opened 2018-11-07
[Synchronous Suspense] Reuse deletions from primary tree

acdlite:
Fixes a bug where deletion effects in the primary tree were dropped before entering the second render pass.

Because we no longer reset the effect list after the first render pass, I've also moved the deletion of the fallback children to the complete phase, after the tree successfully renders without suspending.

Will need to revisit this heuristic when we implement resuming.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -82,6 +82,7 @@ import {
   popHydrationState,
 } from './ReactFiberHydrationContext';
 import {ConcurrentMode, NoContext} from './ReactTypeOfMode';
+import {reconcileChildFibers} from './ReactChildFiber';
 
 function markUpdate(workInProgress: Fiber) {
   // Tag the fiber with an update effect. This turns a Placement into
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -993,7 +993,6 @@ function completeUnitOfWork(workInProgress: Fiber): Fiber | null {
 
       if (nextUnitOfWork !== null) {
         // Completing this fiber spawned new work. Work on that next.
-        nextUnitOfWork.firstEffect = nextUnitOfWork.lastEffect = null;
```
gaearon: So this was the thing that caused us to "forget" about deletions?

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -700,12 +701,30 @@ function completeWork(
...
       }
 
       const nextDidTimeout = nextState !== null;
       const prevDidTimeout = current !== null && current.memoizedState !== null;
+
+      if (current !== null && !nextDidTimeout && prevDidTimeout) {
+        // We just switched from the fallback to the normal children. Delete
+        // the fallback.
+        // TODO: Would it be better to store the fallback fragment on
+        // the stateNode during the begin phase?
```
gaearon: Why would it be better?

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -700,12 +701,30 @@ function completeWork(
...
+        const currentFallbackChild: Fiber | null = (current.child: any).sibling;
+        if (currentFallbackChild !== null) {
+          reconcileChildFibers(
```
gaearon: Would it be simpler if instead of deleting the fallback we kept it hidden? Although that seems bad from memory perspective.

packages/react-reconciler/src/ReactFiberScheduler.js
acdlite: Yes

packages/react-reconciler/src/ReactFiberCompleteWork.js
acdlite: To avoid the confusing series of checks this comment is wrapped in. Could be replaced by checking if the stateNode is null.

I'm not sure that's actually better though. Kinda gross either way.

packages/react-reconciler/src/ReactFiberCompleteWork.js
acdlite: It would be simpler because we would always render two fragments, instead of switching between fragments and no fragments like we do now. That's why `updateSuspenseComponent` has so much code in it; we don't do this type of custom reconciliation anywhere else, yet. Combined with the other weird things we do with Suspense (committing incomplete trees in sync mode, hiding instead of deleting after a timeout) it's gotten pretty complex. Trade off still seems worth it, but unfortunately that means it might take longer to flush out all these edge cases.

(I keep trying to think of a good fuzz, but I can't. Haven't given up though.)

acdlite:
Going to merge so we can try this out on www. @sebmarkbage let me know if this fix doesn't match what we verbally agreed to yesterday.

sebmarkbage:
This makes sense to me but is also scary. :/

PR #14143, opened 2018-11-07
Don't run danger on bad build

sophiebits:
sizebot comments can be confusing when not based on reality.

If results.json doesn't exist, danger will fail. This is what we want.


sophiebits:
Good enough error?

screenshot of sizebot comment explaining results.json was missing

PR #14144, opened 2018-11-08
[Fizz] New Server Rendering Infra

sebmarkbage:
This adds a bunch of new stuff related to the build process for the new server renderer (Fizz).

It adds a new package `react-stream`. The name is temporary (but is nicely owned by @aickin).

This is equivalent to `react-reconciler` but for server rendering. I considered putting them in the same package but they really shouldn't share anything and splitting them up into folders just make both seem more complicated.

This is used to create a two new entry points.

`react-dom/unstable-fizz` which is a server renderer for DOM. Can run in Node for Node streams or browsers for WHATWG streams.

There is also a `react-noop-rendering/server` for tests.

The interesting new twist here is that there is not just a HostConfig (Node.js, Service Worker, other). There's also a FormatConfig (DOM, RN, possibly variations of DOM).

Currently I just copy pasted all the FiberHostConfig overrides to also account for FizzHostConfig and FizzFormatConfig.

The `inlinedHostConfigs.js` model starts breaking down though. Currently the `dom` configuration tests the DOM format config + Node host config. However, that means we wouldn't get coverage on DOM format config + Service Worker host config. Maybe that's fine as long as the Service Worker host config is covered by one of the other configurations?


packages/react-stream/src/ReactFizzRenderer.js
```diff
@@ -0,0 +1,81 @@
+/**
```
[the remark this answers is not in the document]
gaearon: Umm.. Writer? Streamer? You know like cool kids on YouTube.

packages/react-stream/index.js
```diff
@@ -0,0 +1,26 @@
...
+// `react-stream/inline.<renderer>` import (which is *not* Flow typed) to
+// `react-stream/inline-typed` (which *is*) for the current renderer.
+// On CI, we run Flow checks for each renderer separately.
+
+'use strict';
+
+const ReactFizzRenderer = require('./src/ReactFizzRenderer');
+
+// TODO: decide on the top-level export form.
+// This is hacky but makes it work with both Rollup and Jest.
```
acdlite: Are we still doing this? I think we switched to

```js
export * from './src/ReactFizzRenderer';
```

@gaearon

acdlite: Oh, the other entries do that. Maybe I'm confused.

sebmarkbage: https://github.com/facebook/react/blob/master/packages/react-dom/index.js#L15-L16
https://github.com/facebook/react/blob/master/packages/react-reconciler/index.js#L25-L26

gaearon: I think we do that for packages that are ready to be ESM. But here the actual output will be a function anyway so it doesn’t matter. I guess could work either way.

packages/react-dom/package.json
```diff
@@ -29,6 +29,7 @@
     "server.js",
     "server.browser.js",
     "server.node.js",
+    "fizz.js",
```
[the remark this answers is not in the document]
sebmarkbage: hm. Why doesn't this break `ReactDOMFizzServer-test.js` in build tests? I'll need to look into that.

packages/react-stream/index.js
acdlite: My understanding based on absolutely nothing was that old packages do this because to switch to ESM would be a breaking change. But new packages don't have that restriction.

packages/react-dom/fizz.js
```diff
@@ -0,0 +1,12 @@
+/**
```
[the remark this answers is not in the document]
sebmarkbage: To the file name or the method names?

packages/react-stream/index.js
sebmarkbage: I'd love to fix this though. This is probably really bad for inlining/DCE.

packages/react-dom/fizz.js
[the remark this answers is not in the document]
gaearon: I’d add to the file

packages/react-stream/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "react-stream",
```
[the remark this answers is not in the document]
sebmarkbage: You mean the release? We still want it in the sync so we can try it out at FB.

packages/react-dom/package.json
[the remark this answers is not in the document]
gaearon: Because npm entry point (see npm folder) directly points the bundle. The entry point above (which you redirect) only used by Jest. So technically this change isn’t even necessary right now. But would be less confusing.

packages/react-stream/src/ReactFizzHostConfigNode.js
```diff
@@ -0,0 +1,44 @@
+/**
```
sebmarkbage: @gaearon I put this in `react-stream/src/` which is a bit weird because it's not part of the core algorithm. It's part of a possible host config. I'm not sure this makes sense to put it here but it's not really `shared` and we really shouldn't make `shared` a dumping ground.

However, I guess in the future we could have `react-stream` export preconfigured environments and all you fill in is the format config. E.g. `react-stream/index.node.js`, `react-stream/index.browser.js`.

So maybe that makes sense.

packages/react-dom/package.json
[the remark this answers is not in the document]
gaearon: For consistency you can make `npm/fizz` also do a re export. Like `npm/server` does.

packages/react-stream/package.json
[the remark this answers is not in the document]
gaearon: Ah yes I mean the release

packages/react-stream/src/ReactFizzHostConfigNode.js
gaearon: Sure. We also never got these things quite right from first try so I wouldn’t expect to do it now.

sebmarkbage:
I also added a browser build. https://github.com/facebook/react/pull/14144/commits/e65b22f304e3a3d8e86650f195a0222e2034a1fb

Ephem:
I'm moderately familiar with the current server renderer and after having gone through this I just have to add that I really like this approach for a new renderer! Makes so much sense to me.

packages/react-stream/src/ReactFizzHostConfigNode.js
gaearon: Maybe you can add a comment about this? I was just puzzled over this because I forgot.

sebmarkbage:
Since we landed the Fire infra we can land this too.

packages/react-stream/src/ReactFizzStreamer.js
```diff
@@ -0,0 +1,85 @@
...
+  let chunks = request.completedChunks;
+  request.completedChunks = [];
+
+  beginWriting(destination);
+  try {
+    for (let i = 0; i < chunks.length; i++) {
+      let chunk = chunks[i];
+      writeChunk(destination, chunk);
+    }
+  } finally {
```
sompylasar: Why  doesn't it keep the chunks that failed to be written (due to an exception) in `completedChunks` to be retried later?

packages/react-stream/src/ReactFizzStreamer.js
```diff
@@ -0,0 +1,85 @@
...
+    completeWriting(destination);
+  }
+  close(destination);
```
sompylasar: Why is `close(destination)` not called in case of an exception?

PR #14147, opened 2018-11-08
Suspense fuzz tester

acdlite:
The fuzzer works by generating a random tree of React elements. The tree two types of custom components:

- A Text component suspends rendering on initial mount for a fuzzy duration of time. It may update a fuzzy number of times; each update supsends for a fuzzy duration of time.
- A Container component wraps some children. It may remount its children a fuzzy number of times, by updating its key.

The tree may also include nested Suspense components.

After this tree is generated, the tester sets a flag to temporarily disable Text components from suspending. The tree is rendered synchronously. The output of this render is the expected output.

Then the tester flips the flag back to enable suspending. It renders the tree again. This time the Text components will suspend for the amount of time configured by the props. The tester waits until everything has resolved. The resolved output is then compared to the expected output generated in the previous step.

Finally, we render once more, but this time in concurrent mode. Once again, the resolved output is generated to the expected output.

I tested by commenting out various parts of the Suspense implementation to see if broke in the expected way. I also confirmed that it would have caught #14133, a recent bug related to deletions.


packages/react-reconciler/src/__tests__/ReactSuspenseFuzz-test.internal.js
```diff
@@ -0,0 +1,311 @@
...
+
+// const prettyFormatPkg = require('pretty-format');
+// function prettyFormat(thing) {
+//   prettyFormatPkg(thing, {
+//     plugins: [
+//       prettyFormatPkg.plugins.ReactElement,
+//       prettyFormatPkg.plugins.ReactTestComponent,
+//     ],
+//   });
+// }
```
acdlite: Leaving this here because it's useful when debugging a failing test case

gaearon:
>Then the tester flips the flag back to enable suspending. It renders the tree again. This time the Text components will suspend for the amount of time configured by the props. The tester waits until everything has resolved. The resolved output is then compared to the expected output generated in the previous step.

Technically this doesn't test that updates are handled consistently, does it? Would it be more solid if we instead incremented every value in the cache before suspending, and the test verified that all rendered output is original output + 1 after it settles?

acdlite:
@gaearon Could you rephrase? Are you concerned that the cache isn't being cleared in between renders? Each one uses a separate cache.

acdlite:
Here's an example test case. In this test, the Text component suspends for 2 seconds on initial mount. 100 ms after that, it updates itself by incrementing a counter in local state. That update will suspend for 200 ms.

The Container component will also schedule an update remount once after 250 ms, which will trigger the Text mount and updates again (since it's a fresh component).

https://github.com/facebook/react/blob/01c1093a390d790fa145d83c9049009498808d01/packages/react-reconciler/src/__tests__/ReactSuspenseFuzz-test.internal.js#L346-L354

The runner will keep advancing time until everything settles, both promises and updates.

dubzzz:
I know that this PR is not recent at all and I felt on the test while browsing the source code of React.

**Out of curiosity:** Why don't you used property based testing libraries to help you to generate those random tests?
They can be really helpful for those kind of tests in general as they offer what they call shrinking capabilities - _in other words whenever a test fails they try to produce a failing test with inputs simpler to understand (smaller most of the time)_.

You may have a look to [fast-check](https://github.com/dubzzz/fast-check), [jsverify](https://github.com/jsverify/jsverify) or [testcheck-js](https://github.com/leebyron/testcheck-js) for more details.

gaearon:
Wanna send a PR converting it? :-)

For me personally learning the API has been a barrier. I couldn’t actually figure out how to generate recursive things with testcheck-js.

dubzzz:
Yes, I will give it a try but with [fast-check](https://github.com/dubzzz/fast-check) as it handles recursive types (the lib is used by jest and jasmine)

dubzzz:
@gaearon Here is a first draft of the test migrated to property based testing. If you believe it can be useful, I can rework and clean a bit the code before opening a PR.

See https://github.com/dubzzz/react/commit/0b1df02347acc667195d404e8d15d764236aa9a8

gaearon:
Looks interesting. Send a PR and let’s discuss?

dubzzz:
I just opened the PR to move fuzz tests of ReactSuspense to property based testing. I will also migrate `ReactNewContext-test.js` (another fuzz-based test)

PR #14148, opened 2018-11-08
Use Entry in `yarn build ...` Instead of Label

sebmarkbage:
`yarn build core,dom` -> `yarn build react/index react-dom`

or

`yarn build core,dom-client` -> `yarn build react/,dom/index`

I've been frustrated that I can never guess what arguments `yarn build` expects so always end up building everything.

First, I improved the argument parsing a bit so that you don't have to separate the bundle names by commas. But commas still work. This ensures that `yarn build core dom` includes DOM.

It also ensures that spaces like `yarn build core, dom` doesn't build EVERYTHING by becoming an empty entry which matches everything.

I also got rid of label in bundles config. These follow a fairly arbitrary naming convention so I can never guess.

Instead I just use the name from npm entry name using fuzzy search.

There are some cases where this is weird. E.g. `yarn build react` builds everything since everything is called something with `react` in it.

There is one special case. The entry gets appended with `/index.js` at the end.

That allows to build things like `yarn build react/` to only build isomorphic. Or `yarn build react-dom/index` to exclude the server renderers where as `yarn build react-dom` includes server renderers.

### DIFF
--- a/scripts/rollup/build.js
+++ b/scripts/rollup/build.js
@@ -410,6 +410,12 @@ function getPlugins(
   ].filter(Boolean);
 }
 
+const isIndexName = /\/index(.js)?$/;
+
+function endsWith(str, suffix) {
+  return str.substr(str.length - suffix.length) === suffix;
+}
+
 function shouldSkipBundle(bundle, bundleType) {
   const shouldSkipBundleType = bundle.bundleTypes.indexOf(bundleType) === -1;
   if (shouldSkipBundleType) {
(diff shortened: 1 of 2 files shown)


sebmarkbage:
it's useful to do partial matches to some degree still. E.g. `jest` instead of knowing if it is `jest-react` or `react-jest`. Or `react-test` instead of `react-test-renderer`. Maybe we can come up with something clever in a follow up.

PR #14149, opened 2018-11-08
Deleted empty App.css

JoshuaScript:
Came across this empty App.css file & deleted it. Looks like the last commit in that folder was deleting vestigial CRA files, so I'm thinking this was a missed file.


JoshuaScript:
@vikaskyadav Yes, a search reveals that it isn't being used anywhere: https://github.com/facebook/react/search?q=app.css&unscoped_q=app.css

nhunzaker:
Ya. That's definitely the case. Thanks!

PR #14153, opened 2018-11-08
Fix to the memory leak issue with useEffect

trueadm:
This is a possible solution to the memory leak issues @localvoid alerted us to in this repro https://codesandbox.io/s/lz61v39r7.

I'm not 100% sure if the changes in this PR break something else, so it would be good to get @acdlite eye's on this. Specifically this PR makes the following changes:

- When processing an effect's `create` function, we `null` out the `create` field after. If we don't do this, the context of the `create` function closure causes memory retainment.
- Furthermore, if we don't pass the `create` function from `useEffect` to the dependencies array. Doing so, brings the memory retainment issue in the above point. Instead, we tag the `create` function with a field `__key` and assign a unique `symbol` to the function. We then use this symbol as the key in the dependencies.

I can confirm that in the cases of the repro the memory leak issues no longer exist with these two changes. Unfortunately, I don't know of any good way to make tests that can validate memory leaks. If anyone has ideas on how we might test these changes, please let me know.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -316,7 +316,8 @@ function commitHookEffectList(
       }
       if ((effect.tag & mountTag) !== NoHookEffect) {
         // Mount
-        const create = effect.create;
+        const create = ((effect.create: any): () => mixed);
+        effect.create = null;
         let destroy = create();
         if (typeof destroy !== 'function') {
           if (__DEV__) {
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -317,6 +317,8 @@ function commitHookEffectList(
       if ((effect.tag & mountTag) !== NoHookEffect) {
         // Mount
         const create = ((effect.create: any): () => mixed);
+        // Null out the `create` field, avoiding possible memory leaks
+        // due to retaining the function's closure context.
         effect.create = null;
         let destroy = create();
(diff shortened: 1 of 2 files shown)


trueadm:
@sophiebits In order to remove the memory leak on unmount, you need to `null` out the `memoizedState` from the fiber and alternate fiber in `detachFiber`. That doesn't stop the memory leak properly though, so this approach aims at proactively tackling that. In regards to the `__key` change, I actually thought that the key change would improve performance compared to always rerunning. Furthermore, when I tried the approach of always rerunning some hook tests fail, specifically `fires all mutation effects before firing any layout effects` and `force flushes passive effects before firing new layout effects`.

sebmarkbage:
The big question is if we want to be optimizing for the common case:

```
useEffect(() => { });
```

In this case this will always be a new function so all attempts to avoid recalling it will just slow it down.

Since it will always reexecute we ideally don’t need to even create a hook object for it. But we don’t know before we call it if it will also have a destructor associated with it. Unless we add those to a different list somehow.

sebmarkbage:
So this doesn’t work when you have a destroy function right?

```
useEffect(() => () => ...);
```

Still would leak since we need to store the callback somewhere.

Similarly what happens with `useCallback(...)` etc? Do they still leak?

I’m not sure what the bar is for when we’re ok with leaking vs not and which scenarios.

trueadm:
@sebmarkbage I'm not super sure about all cases. The underlying issue is here is retention of closure references and their contexts.

trueadm:
@sebmarkbage I tried the alternative approach as I outlined above: https://github.com/facebook/react/pull/14218

PR #14155, opened 2018-11-08
Add a low priority to the scheduler that still resolves

n8schloss:
Normal priority is for things that we still want the user to see and idle pri is for things that are more speculative and may never resolve. We do not currently have a priority for things (like logging or sending data back to servers) where we want to ensure that it resolves, but where we want other things to be able to happen first. This adds such a priority.

I was debating a bit between calling it 'low' priority or 'logging' priority. I went with calling it 'low' because you can use this for way more than logging, but opening to changing it.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -12,7 +12,8 @@
 var ImmediatePriority = 1;
 var UserBlockingPriority = 2;
 var NormalPriority = 3;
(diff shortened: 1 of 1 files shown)


sebmarkbage:
Seems reasonable to me. Thoughts, @acdlite?

acdlite:
Oh duh, we already that one. I thought it was still "offscreen" or something.

Let's merge and figure out the name later.

PR #14157, opened 2018-11-08
[Synchronous Suspense] Don't delete children of suspended component

acdlite:
Vestigial behavior that should have been removed in #13823.

Found using the Suspense fuzz tester in #14147.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberUnwindWork.js
+++ b/packages/react-reconciler/src/ReactFiberUnwindWork.js
@@ -70,7 +70,6 @@ import {
   LOW_PRIORITY_EXPIRATION,
 } from './ReactFiberExpirationTime';
 import {findEarliestOutstandingPriorityLevel} from './ReactFiberPendingPriority';
-import {reconcileChildren} from './ReactFiberBeginWork';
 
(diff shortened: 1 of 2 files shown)


acdlite:
@sebmarkbage I blame it on sloppiness on my part when I was updating the tests in #13823. But yeah, it's getting pretty complex, too.

There are a few things in our Suspense implementation that are unusual:

- Custom reconciliation. We conditionally wrap the children in a fragment while preserving their state.
- Outside concurrent mode, we commit the suspended tree in an incomplete state. By itself, this is similar to our `hidden` trick.
- Semantic divergence between concurrent and synchronous mode. Usually the only difference between the two modes is scheduling. There are some second order things that diverge, but synchronous mode is essentially a subset of the semantics of concurrent mode. But with Suspense, the semantics are forked.
- Hiding a tree instead of deleting it. Conceptually similar to the `hidden` trick except we have to do the hiding ourselves, by toggling the nearest host children.

Any one of these quirks by themselves isn't so complex, but we have to consider all of them together.

I still think they're worth it, on the whole, but if we ever decide it's too much, we might need to drop one of these constraints.

PR #6846, opened 2016-05-23
Tutorial ported to ES6 classes or stateless functions

ankitml:
Issue reference : https://github.com/facebook/react/issues/3710

*Before* submitting a pull request, please make sure the following is done...
1. Fork the repo and create your branch from `master`.
2. If you've added code that should be tested, add tests!
3. If you've changed APIs, update the documentation.
4. Ensure the test suite passes (`grunt test`).
5. Make sure your code lints (`grunt lint`) - we've done our best to make sure these rules match our internal linting guidelines.
6. If you haven't already, complete the [CLA](https://code.facebook.com/cla).

PR #14161, opened 2018-11-09
Improved suspense support in ReactDOMServer

alexmckenley:
I've updated the support for suspense in the ReactPartialRenderer. Instead of always rendering the fallback, the server renderer will attempt to render the children, and only revert to the fallback if rendering is suspended with a thrown promise. 

Let me know what you think.


alexmckenley:
Updated based on recommendations from @sophiebits. 💯

alexmckenley:
@sophiebits Thanks for the review! I pushed up the changes, but I dont see a merge button, seems like I may not have permission.

Tor01100:
dom/src/server/ReactPartialRenderer.js
index 643d53cf0c3..c5f124e7c4a 100644


Can't be running in js try using HTML script for the patch for #6846

PR #14164, opened 2018-11-09
Remove errant return assignment

### LINKED ISSUE #14162: "Cannot set property 'return' of null" while using Suspense and lazy

bug

Given code like this:
```jsx
<Suspense fallback={...}>
  {this.state.render && <LazyComponent />}
</Suspense>
```
my `LazyComponent` should download and render when I run `this.setState({ render: true })`. Instead, I get an internal React error about setting property `null` of `return`.

https://jsfiddle.net/cq9hfwe8/1/. Open the console to see the error message.

The combination of `Suspense` and `lazy` defer downloading and rendering my component until I set a state variable to `true`.

Tried on Windows in Chrome 71, React 16.6.1.

acdlite:
Oopsie!

This could have been avoided if our types were modeled correctly with Flow (using a disjoint union).

Fuzz tester didn't catch it because it does not generate cases where a Suspense component mounts with no children. I'll update it.

Fixes #14162

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -1244,7 +1244,7 @@ function updateSuspenseComponent(
     } else {
       // The current tree has not already timed out. That means the primary
       // children are not wrapped in a fragment fiber.
-      const currentPrimaryChild: Fiber = (current.child: any);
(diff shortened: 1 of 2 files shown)


acdlite:
@sophiebits You do right before you traverse down into them, which is not true in this case. I think this is some copypasta from a different branch.

Issue #14125, opened 2018-11-06
Autofocus Text puts cursor at end instead of beginning


Matthew-Goldberg:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
In React 16, a text input with a value and autofocus will set the cursor at the _end_ of the input.
`<input autoFocus={true} type="text" defaultValue="4444" />`

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

See the React 16 repro here: https://codepen.io/matthewg0/pen/XymjYo

**What is the expected behavior?**
In React 15, and in plain HTML, the behavior is that the cursor is placed at the _start_ of the input.
React 15: https://codepen.io/matthewg0/pen/NEGRzv
Plain HTML: https://codepen.io/matthewg0/pen/wQKoap

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
Reproduces as of React 16.  Worked as expected in React 15.

gaearon:
Want to look into a fix?

Matthew-Goldberg:
Sure thing, thanks for confirming.  I'll look into it.

mmarkelov:
@Matthew-Goldberg I just tried to figure out what is wrong with it, but I don't have any ideas with it. So if you stack on it - you could try [this](https://codesandbox.io/s/rmjmlmxv64) - works in all modern browsers

heyimalex:
Just to narrow it down, v16.2.0 works and v16.3.0 doesn't, so it's somewhere in there.

chasestarr:
bisected to [this commit](https://github.com/facebook/react/commit/fd69c239a0bb8fde412aa7c1c0cea9c7fda287a9)

Matthew-Goldberg:
@mmarkelov I think this will be hard to write a test, it seems `jsdom` is showing that the `selectionStart` is `0` even though in the browser the same code shows `selectionStart` is the end/length of the text value.

Matthew-Goldberg:
OK the `jsdom` issue is now fixed: https://github.com/jsdom/jsdom/pull/2799
When that is released we can upgrade the `jsdom` version to pick up the fix, and re-implement the test which should now be useful.

[...]

PR #14171, opened 2018-11-09
Fix input autofocus position

mmarkelov:
https://github.com/facebook/react/issues/14125

I am not sure that it is right approach. So it will be fine, if someone could check it. Maybe I miss something.


packages/react-dom/src/__tests__/ReactDOM-test.js
```diff
@@ -308,14 +308,15 @@ describe('ReactDOM', () => {
...
-          <input autoFocus={true} />
+          <input defaultValue="foo" autoFocus={true} />
           <p>The above input should be focused after mount.</p>
         </div>,
         container,
       );
 
       expect(inputFocusedAfterMount).toBe(true);
       expect(focusedElement.tagName).toBe('INPUT');
+      expect(focusedElement.selectionStart).toBe(0);
```
Matthew-Goldberg: Are you sure this test is useful?  It seems this assertion passes even without a fix to the actual problem.

xavierrutayisire:
Any news about this PR?
It would be nice to have this fix.

xavierrutayisire:
Hey, do you know if we could have this fix in an upcoming release?
I'm working on a project where the wrong autofocus position is a blocking problem.

xavierrutayisire:
Hi @mmarkelov @vikaskyadav @Matthew-Goldberg,

Do you know if this pull request can still be done soon?

Matthew-Goldberg:
The challenge was figuring out how to write a test; it seems jsdom is showing that the selectionStart is 0 even though in the browser the same code shows selectionStart is the end/length of the text value.  Maybe an issue with jsdom.

Matthew-Goldberg:
I have opened an issue with `jsdom` to see if we can get that library updated so we could write a valid test.
https://github.com/jsdom/jsdom/issues/2787

Matthew-Goldberg:
OK the `jsdom` issue is now fixed: https://github.com/jsdom/jsdom/pull/2799
When this is released we can upgrade the `jsdom` version to pick up the fix, and re-implement the test which should now be useful.

Matthew-Goldberg:
`jsdom` fix is released https://github.com/jsdom/jsdom/releases/tag/16.1.0

gaearon:
We're going to remove attribute syncing in one of the next majors. Then this will be solved. I don't think we should be adding more complexity here because it is already a lot.

Matthew-Goldberg:
@gaearon it may still be useful to implement the test @mapkeji added in 9d78fbe54a1c465bb2a36951ead5d8d91a73d8e3
At the time, this test would pass either way so it was not useful.
With the `jsdom` upgrade it should be useful.
In any case, can't add the working test unless the actual implementation matches spec.
But just a note in case the expectation is to restore this behavior with the attribute syncing change.

PR #14178, opened 2018-11-09
SimpleMemoComponent should warn if a ref is given

### LINKED ISSUE #13964: React.memo – attempting to assign a ref to a component wrapped in memo does not produce a friendly error message
Bug

Attempting to assign a ref to a component wrapped in React.memo does not produce a friendly error message. Attempting the same thing with a functional component *not* wrapped in memo gives a descriptive error ("Warning: Function components cannot be given refs. Attempts to access this ref will fail.").

I trust this is not needed for this case

The same error message as with a non-memo functional component should be displayed.

16.6.0

sophiebits:
Fixes #13964.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -250,6 +250,9 @@ function updateMemoComponent(
       // to a SimpleMemoComponent to allow fast path updates.
       workInProgress.tag = SimpleMemoComponent;
       workInProgress.type = type;
+      if (__DEV__) {
+        validateFunctionComponentInDev(workInProgress, type);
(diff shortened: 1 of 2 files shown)


sophiebits:
@bvaughn I don't understand what you mean. The test does test SimpleMemoComponent.

bvaughn:
Sorry. I wrote the exact opposite of what I meant to write. ☹️ We're explicitly testing the simple memo case but not the lazy+memo case.

bvaughn:
Hmm, well ok then.

sophiebits:
Pardon?

bvaughn:
😊 uh... please ignore me. wrong response / wrong issue.

Issue #8689, opened 2017-01-04
Shorthand CSS properties can cause style inconsistencies


bvaughn:
I'm reporting a **bug**.

## What is the current behavior?

1. Render a host component with a `style` that contains both a [shorthand CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties) and an over-writing longer form of the property (eg `overflow` and a conflicting `overflowX` and/or `overflowY` property).
2. Update the host component (re-render) with the same shorthand property (eg `overflow`) but remove the longer form property (eg `overflowX`).
3. The host component style is now invalid.

### Examples

This bug can be reproduced here: https://plnkr.co/edit/7O1xZUqbD13ST2BAxcm9?p=preview

*Update*: I have since updated the Plnkr to use `border` instead of `overflow` since it makes the problem more immediately obvious to spot.

#### Example
1. Render `style={{ overflow: 'hidden', overflowY: 'auto' }}`
1. Render `style={{ overflow: 'hidden' }}`
1. Expected: `<div style="overflow: hidden;" />`
1. Actual: `<div style="overflow-x: hidden;" />`

#### Alternate example
1. Render `style={{ overflow: 'auto', overflowX: 'hidden', overflowY: 'hidden' }}`
1. Render `style={{ overflow: 'auto' }}`
1. Expected: `<div style="overflow: auto" />`
1. Actual: `<div style="" />`

You can also reproduce this bug with properties like `margin`, `padding`, `border`, etc.

#### Caveats

Note that if the shorthand value changes between renders things work as expected (because React explicitly updates the shorthand style).

I know this is very edge-case behavior and so may not be worth fixing. I originally noticed it by way of issue bvaughn/react-virtualized/issues/525.

## Which versions of React / browser / OS are affected

This bug reproduces in Chrome, Firefox, and Safari using React 15.4.1 as well as the unreleased `ReactDOMFiber` renderer.

bvaughn:
I think the underlying cause here is that `ReactDOM` only updates the `style` attributes that have changed (eg `overflowX`, `overflowY`). But removing one or both of these properties from a `CSSStyleDeclaration` impacts the `overflow` property as well, resulting in a `CSSStyleDeclaration` instance that doesn't _necessarily_ match the most recently specified `style` object..

I was not able to reproduce this issue in a unit test (using Jest) because Jsdom doesn't trickle down the effects like the major browser vendors seem to (eg updating `overflowX` doesn't impact `overflow`).

[...]

PR #14181, opened 2018-11-09
Warn about conflicting style values during updates

### LINKED ISSUE #6348: Warn when using overlapping styles (e.g. border and borderBottom)

sophiebits:
This is one of the most insidious quirks of React DOM that people run into. Now we warn when we think an update is dangerous.

We still allow rendering `{background, backgroundSize}` with unchanging values, for example. But once you remove either one or change `background` (without changing `backgroundSize` at the same time), that's bad news. So we warn.

Fixes #6348.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
@@ -146,6 +146,141 @@ describe('ReactDOMComponent', () => {
       }
     });
 
+    it('should warn for conflicting CSS shorthand updates', () => {
+      const container = document.createElement('div');
+      ReactDOM.render(
+        <div style={{font: 'foo', fontStyle: 'bar'}} />,
+        container,
+      );
+      expect(() =>
+        ReactDOM.render(<div style={{font: 'foo'}} />, container),
+      ).toWarnDev(
+        'Warning: Removing a style property during rerender (fontStyle) ' +
+          'when a conflicting property is set (font) can lead to styling ' +
+          "bugs. To avoid this, don't mix shorthand and non-shorthand " +
+          'properties for the same value; instead, replace the shorthand ' +
+          'with separate values.' +
+          '\n    in div (at **)',
+      );
+
+      // These updates are OK and don't warn:
+      ReactDOM.render(
+        <div style={{font: 'qux', fontStyle: 'bar'}} />,
+        container,
+      );
+      ReactDOM.render(
+        <div style={{font: 'foo', fontStyle: 'baz'}} />,
+        container,
+      );
+
+      expect(() =>
+        ReactDOM.render(
+          <div style={{font: 'qux', fontStyle: 'baz'}} />,
+          container,
+        ),
+      ).toWarnDev(
+        'Warning: Updating a style property during rerender (font) when ' +
+          'a conflicting property is set (fontStyle) can lead to styling ' +
+          "bugs. To avoid this, don't mix shorthand and non-shorthand " +
+          'properties for the same value; instead, replace the shorthand ' +
+          'with separate values.' +
+          '\n    in div (at **)',
+      );
+      expect(() =>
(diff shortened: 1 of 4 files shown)


sophiebits:
@bvaughn This is probably one of the only bugs that both you (#8689) and I (https://github.com/facebook/react/issues/6348#issuecomment-391562336) have both run into. 😂 Hope this warning helps.

packages/react-dom/src/shared/CSSPropertyOperations.js
```diff
@@ -78,3 +85,103 @@ export function setValueForStyles(node, styles) {
...
+ *   {font: 'foo', fontVariant: 'bar'} -> {font: 'baz', fontVariant: 'bar'}
+ *   becomes .style.font = 'baz'
+ * Removing a shorthand property (longhand gets lost too):
+ *   {font: 'foo', fontVariant: 'bar'} -> {fontVariant: 'bar'}
+ *   becomes .style.font = ''
+ * Removing a longhand property (should revert to shorthand; doesn't):
+ *   {font: 'foo', fontVariant: 'bar'} -> {font: 'foo'}
+ *   becomes .style.fontVariant = ''
+ */
+export function validateShorthandPropertyCollisionInDev(
```
bvaughn: Seems like this could get noisy if we don't do any kind of de-duping. I wonder if this is something you considered and ruled out?

Guess de-duping would have a small impact on tests too.

packages/react-dom/src/__tests__/ReactDOMComponent-test.js
```diff
@@ -146,6 +146,147 @@ describe('ReactDOMComponent', () => {
...
+        container,
+      );
+      ReactDOM.render(
+        <div style={{font: 'foo', fontStyle: 'baz'}} />,
+        container,
+      );
+
+      expect(() =>
+        ReactDOM.render(
+          <div style={{font: 'qux', fontStyle: 'baz'}} />,
```
bvaughn: Ugh, this is subtle. Had to read these tests a couple times. Nice to have a warning for it.

packages/react-dom/src/shared/CSSPropertyOperations.js
sophiebits: I thought about it but I don't think this will be very common, it is a bad bug, and it is pretty easy to fix – so I think it's OK not to.

gaearon:
This is out in 16.13.
https://reactjs.org/blog/2020/03/02/react-v16.13.0.html

PR #14182, opened 2018-11-09
Use unique thread ID for each partial render to access Context

sebmarkbage:
Fixes #13874

Alternative to #13877

This first adds an allocator that keeps track of a unique ThreadID index for each currently executing partial renderer. IDs are not just growing but are reused as streams are destroyed.

This ensures that IDs are kept nice and compact.

_One minor breakage is that it is no longer safe to just let streams be GC:ed for clean up. Typically, they you would never drop a stream on the floor. It's either exhausted or errors._

This lets us use an "array" for each Context object to store the current values. The look up for these are fast because they're just looking up an offset in a tightly packed "array".

I don't use an actual Array object to store the values. Instead, I rely on that VMs (notably V8) treat storage of numeric index property access as a separate "elements" allocation.

This lets us avoid an extra indirection.

However, we must ensure that these arrays are not holey to preserve this feature.

To do that I store the `_threadCount` on each context (effectively it takes the place of the `.length` property on an array, and also lets me use that pun). It's unclear whether a "real" `.length` would be faster if it could avoid the internal bounds check.

This lets us first validate that the context has enough slots before we access the slot. If not, we fill in the slots with the default value.

This *should* be a fast approach, in theory, but I haven't actually confirmed that the builds don't deopt somewhere yet.

cc @leebyron

### DIFF
--- a/packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js
+++ b/packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js
@@ -18,11 +18,15 @@ class ReactMarkupReadableStream extends Readable {
     this.partialRenderer = new ReactPartialRenderer(element, makeStaticMarkup);
   }
 
+  _destroy() {
+    this.partialRenderer.destroy();
+  }
+
   _read(size) {
     try {
       this.push(this.partialRenderer.read(size));
     } catch (err) {
-      this.emit('error', err);
+      this.destroy(err);
     }
   }
 }
--- a/packages/react-dom/src/server/ReactDOMStringRenderer.js
+++ b/packages/react-dom/src/server/ReactDOMStringRenderer.js
@@ -14,8 +14,12 @@ import ReactPartialRenderer from './ReactPartialRenderer';
  */
 export function renderToString(element) {
   const renderer = new ReactPartialRenderer(element, false);
-  const markup = renderer.read(Infinity);
-  return markup;
+  try {
+    const markup = renderer.read(Infinity);
+    return markup;
+  } finally {
+    renderer.destroy();
+  }
 }
 
 /**
@@ -25,6 +29,10 @@ export function renderToString(element) {
  */
 export function renderToStaticMarkup(element) {
   const renderer = new ReactPartialRenderer(element, true);
-  const markup = renderer.read(Infinity);
-  return markup;
+  try {
+    const markup = renderer.read(Infinity);
+    return markup;
+  } finally {
+    renderer.destroy();
+  }
 }
--- a/packages/react-dom/src/server/ReactPartialRendererContext.js
+++ b/packages/react-dom/src/server/ReactPartialRendererContext.js
@@ -1,0 +1,104 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import type {ThreadID} from './ReactThreadIDAllocator';
+import type {ReactContext} from 'shared/ReactTypes';
+
+import {REACT_CONTEXT_TYPE} from 'shared/ReactSymbols';
+import ReactSharedInternals from 'shared/ReactSharedInternals';
+import getComponentName from 'shared/getComponentName';
+import warningWithoutStack from 'shared/warningWithoutStack';
+import checkPropTypes from 'prop-types/checkPropTypes';
+
+let ReactDebugCurrentFrame;
+if (__DEV__) {
+  ReactDebugCurrentFrame = ReactSharedInternals.ReactDebugCurrentFrame;
+}
+
+const didWarnAboutInvalidateContextType = {};
+
+export const emptyObject = {};
+if (__DEV__) {
+  Object.freeze(emptyObject);
+}
+
+function maskContext(type, context) {
+  const contextTypes = type.contextTypes;
+  if (!contextTypes) {
+    return emptyObject;
+  }
+  const maskedContext = {};
+  for (const contextName in contextTypes) {
+    maskedContext[contextName] = context[contextName];
+  }
+  return maskedContext;
+}
+
+function checkContextTypes(typeSpecs, values, location: string) {
+  if (__DEV__) {
+    checkPropTypes(
+      typeSpecs,
+      values,
+      location,
+      'Component',
+      ReactDebugCurrentFrame.getCurrentStack,
+    );
+  }
+}
+
+export function validateContextBounds(
+  context: ReactContext<any>,
+  threadID: ThreadID,
+) {
+  // If we don't have enough slots in this context to store this threadID,
+  // fill it in without leaving any holes to ensure that the VM optimizes
+  // this as non-holey index properties.
+  for (let i = context._threadCount; i <= threadID; i++) {
+    // We assume that this is the same as the defaultValue which might not be
+    // true if we're rendering inside a secondary renderer but they are
+    // secondary because these use cases are very rare.
+    context[i] = context._currentValue2;
+    context._threadCount = i + 1;
+  }
+}
+
+export function processContext(
+  type: Function,
+  context: Object,
+  threadID: ThreadID,
+) {
+  const contextType = type.contextType;
+  if (typeof contextType === 'object' && contextType !== null) {
+    if (__DEV__) {
+      if (contextType.$$typeof !== REACT_CONTEXT_TYPE) {
+        let name = getComponentName(type) || 'Component';
+        if (!didWarnAboutInvalidateContextType[name]) {
+          didWarnAboutInvalidateContextType[name] = true;
+          warningWithoutStack(
+            false,
+            '%s defines an invalid contextType. ' +
+              'contextType should point to the Context object returned by React.createContext(). ' +
+              'Did you accidentally pass the Context.Provider instead?',
+            name,
+          );
+        }
+      }
+    }
+    validateContextBounds(contextType, threadID);
+    return contextType[threadID];
+  } else {
+    const maskedContext = maskContext(type, context);
+    if (__DEV__) {
+      if (type.contextTypes) {
+        checkContextTypes(type.contextTypes, maskedContext, 'context');
+      }
+    }
+    return maskedContext;
+  }
+}
--- a/packages/react-dom/src/server/ReactThreadIDAllocator.js
+++ b/packages/react-dom/src/server/ReactThreadIDAllocator.js
@@ -1,0 +1,56 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+// Allocates a new index for each request. Tries to stay as compact as possible so that these
+// indices can be used to reference a tightly packaged array. As opposed to being used in a Map.
+// The first allocated index is 1.
+
+export type ThreadID = number;
+
+let nextAvailableThreadIDs = new Uint16Array(16);
+for (let i = 0; i < 15; i++) {
+  nextAvailableThreadIDs[i] = i + 1;
+}
+nextAvailableThreadIDs[15] = 0;
+
+function growThreadCountAndReturnNextAvailable() {
+  let oldArray = nextAvailableThreadIDs;
+  let oldSize = oldArray.length;
+  let newSize = oldSize * 2;
+  if (newSize > 0x10000) {
+    throw new Error(
+      'Maximum number of concurrent React renderers exceeded. ' +
+        'This can happen if you are not properly destroying the Readable provided by React. ' +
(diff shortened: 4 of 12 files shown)


packages/react-dom/src/server/ReactPartialRendererContext.js
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export function validateContextBounds(
+  context: ReactContext<any>,
+  threadID: ThreadID,
+) {
+  // If we don't have enough slots in this context to store this threadID,
+  // fill it in without leaving any holes to ensure that the VM optimizes
+  // this as non-holey index properties.
+  for (let i = context._threadCount; i <= threadID; i++) {
```
sophiebits: this is easier to read for me:

```js
while (context._threadCount < threadID) {
  context[context._threadCount++] = context.currentValue2;
}
``

File: packages/react/src/ReactContext.js [hunk 2]
```diff
@@ -42,6 +42,9 @@ export function createContext<T>(
     // Secondary renderers store their context values on separate fields.
     _currentValue: defaultValue,
     _currentValue2: defaultValue,
```
sophiebits: why even keep these any more instead of using "threads" on the client too? only the lack of global coordination?

packages/react-dom/src/server/ReactThreadIDAllocator.js
```diff
@@ -0,0 +1,56 @@
...
+for (let i = 0; i < 15; i++) {
+  nextAvailableThreadIDs[i] = i + 1;
+}
+nextAvailableThreadIDs[15] = 0;
+
+function growThreadCountAndReturnNextAvailable() {
+  let oldArray = nextAvailableThreadIDs;
+  let oldSize = oldArray.length;
+  let newSize = oldSize * 2;
+  if (newSize > 0x10000) {
```
sophiebits: nit: (2 << 16) probably easier to read

packages/react-dom/src/server/ReactThreadIDAllocator.js
```diff
@@ -0,0 +1,56 @@
...
+  return oldSize;
+}
+
+export function allocThreadID(): ThreadID {
+  let nextID = nextAvailableThreadIDs[0];
+  if (nextID === 0) {
+    return growThreadCountAndReturnNextAvailable();
+  }
+  nextAvailableThreadIDs[0] = nextAvailableThreadIDs[nextID];
+  return nextID;
```
sophiebits: Could return nextID - 1 (then +1 in free) so they appear to start at 0.

packages/react-dom/src/server/ReactThreadIDAllocator.js
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+// Allocates a new index for each request. Tries to stay as compact as possible so that these
+// indices can be used to reference a tightly packaged array. As opposed to being used in a Map.
+// The first allocated index is 1.
```
sebmarkbage: I think what we're going to do is use this same strategy on the client but primary renders like React DOM and React Native will use index 0 hard coded so they can skip any allocation and resizing logic. Then secondary renderers can use index 1 and above.

packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js
```diff
@@ -18,11 +18,15 @@ class ReactMarkupReadableStream extends Readable {
...
 
+  _destroy() {
+    this.partialRenderer.destroy();
+  }
+
   _read(size) {
     try {
       this.push(this.partialRenderer.read(size));
     } catch (err) {
-      this.emit('error', err);
```
sophiebits: is this breaking?

packages/react-dom/src/server/ReactThreadIDAllocator.js
```diff
@@ -0,0 +1,56 @@
...
+function growThreadCountAndReturnNextAvailable() {
+  let oldArray = nextAvailableThreadIDs;
+  let oldSize = oldArray.length;
+  let newSize = oldSize * 2;
+  if (newSize > 0x10000) {
+    throw new Error(
+      'Maximum number of concurrent React renderers exceeded. ' +
+        'This can happen if you are not properly destroying the Readable provided by React. ' +
+        'Ensure that you call .destroy() on it if you no longer want to read from it.',
+    );
```
sophiebits: ┏┓  
┃┃╱╲ in this
┃╱╱╲╲  house
╱╱╭╮╲╲  we 
▔▏┗┛▕▔    use
╱▔▔▔▔▔▔▔▔▔▔╲  
              invariants
╱╱┏┳┓╭╮┏┳┓ ╲╲ 
▔▏┗┻┛┃┃┗┻┛▕▔

also can we tweak:
> Ensure that you call .destroy() on it when you are finished reading from it.
current wording sounds a bit like a special case

packages/react-dom/src/server/ReactThreadIDAllocator.js
sebmarkbage: I wonder if this will mess with SMI optimizations. SMI - 1 might no longer be a SMI maybe?

Also just doing a bit of extra math is annoying. I think I'll use the primary renderer optimization as an excuse for why we shouldn't. :P

sebmarkbage:
Maybe we'll find something else to store in there. :D 

I suspect that we'll want to use these IDs for more things in the future. E.g. we have similar concepts in the interaction tracking. If we start using actual threads, then all our globals needs to be thread local but adding many thread locals globally is bad news so maybe things like current owner and dispatcher will use this too.

packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js
sophiebits: what is it fixing?

sebmarkbage:
We don't really have good coverage enough internally to really know if this is fast/safe in production environments. The best we can do at this point is probably just to release it in a patch.

packages/react-dom/src/server/ReactThreadIDAllocator.js
```diff
@@ -0,0 +1,58 @@
...
+function growThreadCountAndReturnNextAvailable() {
+  let oldArray = nextAvailableThreadIDs;
+  let oldSize = oldArray.length;
+  let newSize = oldSize * 2;
+  invariant(
+    newSize <= 0x10000,
+    'Maximum number of concurrent React renderers exceeded. ' +
+      'This can happen if you are not properly destroying the Readable provided by React. ' +
+      'Ensure that you call .destroy() on it if you no longer want to read from it.' +
+      ', and did not read to the end. If you use .pipe() this should be automatic.',
```
sophiebits: .,

sophiebits:
Instead of denoting the terminal ID as `nextIDs[id] === 0`, you could note it as `nextIDs[id] === id` and then instead of using `nextIDs[0]` to track the next available, store a separate int pointing to the next index. Then your IDs would start at 0.

packages/react-dom/src/server/ReactDOMStringRenderer.js
```diff
@@ -14,8 +14,12 @@ import ReactPartialRenderer from './ReactPartialRenderer';
  */
 export function renderToString(element) {
   const renderer = new ReactPartialRenderer(element, false);
-  const markup = renderer.read(Infinity);
-  return markup;
+  try {
+    const markup = renderer.read(Infinity);
+    return markup;
```
leebyron: Any reason why these aren't just `return renderer.read(Infinity)`?

leebyron:
I agree with @sophiebits that it seems like a useful optimization to have thread IDs start at 0 and grow larger rather than start at the high ID and shrink towards 0 - that avoids having to fill a bunch of holes on context objects for the common case of a single thread

leebyron:
I'm also curious to see if there's a meaningful performance difference between this and using a Map per renderer (especially native Map) and instead swapping in an implementation of useContext while performing partial rendering to use the correct Map

sophiebits:
@leebyron The IDs are increasing. They just start at 1.

trueadm:
It seems that this PR breaks a use case for new context, one where you have no parent provider, you get the value `undefined` which is incorrect.

gaearon:
Why didn’t tests catch this?

trueadm:
@gaearon I don't know enough about this concurrent SSR implementation. @sebmarkbage should be able to explain why?

packages/react-dom/src/server/ReactPartialRendererContext.js
```diff
@@ -0,0 +1,104 @@
...
+    // We assume that this is the same as the defaultValue which might not be
+    // true if we're rendering inside a secondary renderer but they are
+    // secondary because these use cases are very rare.
+    context[i] = context._currentValue2;
```
sophiebits: @trueadm This is meant to read the default value.

packages/react/src/ReactContext.js
sophiebits: (originally set here)

packages/react-dom/src/server/ReactPartialRendererContext.js
sebmarkbage: I’m not sure how the proxing works with this. Is it DEV only or both?

gaearon: What proxying?

gaearon:
@trueadm 

>I expect to get 10, but I instead get undefined.

Could you please explain the testing methodology? This test prints `10` if I add it to `ReactServerRendering-test.js`:

```js
    fit('waat', () => {
const MyContext = React.createContext(10);

function Component() {
  const { Consumer, Provider } = MyContext;
  return (
    <React.Fragment>
      <Consumer>
          {(value: number) => <span>{value}</span>}
      </Consumer>
    </React.Fragment>
  )
}

  console.log(ReactDOMServer.renderToString(<Component />))
    });
```

screen shot 2018-11-19 at 8 29 55 pm


I also checked that 

```js
const React = require('react')
const ReactDOMServer = require('react-dom/server')

const MyContext = React.createContext(10);

function Component() {
  const {
    Consumer,
    Provider
  } = MyContext;
  return React.createElement(React.Fragment, null, React.createElement(Consumer$
}
```

prints `<span>10</span>` with `node index.js`.

trueadm:
@gaearon I'm doing this like in your second example, except I've copied over the production server bundle and pasted it into the `node_modules/react-dom` directory so it's using the master build.

gaearon:
Are you sure you don't have two React copies?

packages/react/src/ReactContext.js
```diff
@@ -42,6 +42,9 @@ export function createContext<T>(
...
+    // Used to track how many concurrent renderers this context currently
+    // supports within in a single renderer. Such as parallel server rendering.
+    _threadCount: 0,
```
sophiebits: @trueadm did you copy over the new `react` package? this line is important.

gaearon: Oh. Does this mean `react-dom/server` is not compatible with `react@16.0.0`? We've tried to relax this requirement during 16.x.x release line. If this doesn't work we need to start bumping peer deps again.

trueadm: This was it! I'm sorry, I feel stupid now :(

trueadm:
Ignore me. I was using the wrong `react` package. Sorry for the waste of time.

gaearon:
It still points out an issue though. We intended to support mismatching versions of `react` and `react-dom` within a major, and so far it worked in 16. So we should probably fix.

packages/react/src/ReactContext.js
sebmarkbage: @gaearon We accidentally bumped peer deps anyway.

nomcopter:
This broke me in a minor version update (16.6.1 -> 16.6.3). I was calling `ReactDOMServer.renderToStaticMarkup(<Child />)` in a `Parent`'s render function in order to `dangerouslySetInnerHTML` in a `contentEditable` element. In 16.6.1 `Child` had access to the context that `Parent` had access to. 
In 16.6.3 `Child` no longer has access to `Parent`'s context. 

This is probably desirable, as the previous behavior seems weird, but it is likely worth flagging as a potential breaking change in the release notes. Unfortunately, it is already out in a minor version, although this use case seems very rare.

sebmarkbage:
@nomcopter Thanks for flagging. I think we’re going to consider that a bug fix. It is not intended or desirable. What is worse is that it could accidentally leak a context unintentionally in similar ways as the big this PR was meant to fix.

packages/react/src/ReactContext.js
gaearon: The fix is simple enough I did it anyway. We should relax peer deps back imo.

gaearon:
Filed https://github.com/facebook/react/issues/14292 to track it.

packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js
voxpelli: I submitted a PR to fix this: https://github.com/facebook/react/pull/14314

PR #14183, opened 2018-11-10
Simplify CSS shorthand property warning

sophiebits:
I figured out a simpler way to do #14181. It does allocate some but I think that's OK. Time complexity might even be better since we avoid the nested loops the old one had.

### DIFF
--- a/packages/react-dom/src/shared/CSSPropertyOperations.js
+++ b/packages/react-dom/src/shared/CSSPropertyOperations.js
@@ -5,11 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import {
-  overlappingShorthandsInDev,
-  longhandToShorthandInDev,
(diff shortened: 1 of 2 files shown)


sebmarkbage:
Are we ok releasing the first one in a patch release?

I think we should probably add this warning behind a feature flag since it is likely to be very noisy.

sophiebits:
I don't think this will be very noisy. I think it basically always indicates a real bug.

sebmarkbage:
In that case we should first turn it on internally and verify that assumption.

sophiebits:
Happy to. If we want to cut a patch first we can disable it temporarily I guess.

gaearon:
@astr0sl0th This PR has been merged a year ago. Please don’t leave reviews on old PRs, you’re spamming everyone’s notifications.

Issue #14102, opened 2018-11-05
Synthetic KeyboardEvent should support KeyboardEvent.code


jjspace:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
The current synthetic keyboard event does not support the [`KeyboardEvent.code`](https://www.w3.org/TR/uievents/#dom-keyboardevent-code) property.

**What is the expected behavior?**
The synthetic keyboard event should pass along the [`KeyboardEvent.code`](https://www.w3.org/TR/uievents/#dom-keyboardevent-code) property. This is currently in the WD of DOM Events but is part of replacing `keyCode` and `charCode` and is much more consistent and easy to use. This is currently only supported by FF and Chrome ([CanIUse](https://caniuse.com/#feat=keyboardevent-code)) so it may be a bit premature to fully integrate. However `keyCode`, `charCode` and `which` are being deprecated so this will eventually need to be added.

Edit: I spoke too quickly, CanIUse shows that FF, Chrome, Safari and Opera support it. IE, Edge and most mobile browsers do not.

philipp-spiess:
Seems like Safari also supports this (looking at caniuse.com). Do you want to send a PR? 🙂

charliesmart:
It seems like the [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) should also probably added. I'd be happy to send a PR for both if you don't hear back from @jjspace

jjspace:
@philipp-spiess is right, `key` is already there and implemented correctly (this was what I adapted my code to use in lieu of `code`) I'm happy to take a look at adding `code` but I wont have much time until the weekend if @charliesmart or someone else wants to do it before then. It doesn't seem like a large change apart from browser support.

jjspace:
I've created the PR to pass the `code` property along to the synthetic event. I looked into solutions to provide a polyfill to browsers without support like Edge and IE but the property is actually very complicated and needs access to the actual hardware keycodes as seen [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Code_values) so it's not really possible. 

This should also have a change in the [docs](https://reactjs.org/docs/events.html#keyboard-events). How do I go about updating those? Also, when I do would it make sense to comment on the partial lack of browser support?

mydiemho:
is there any updates on this?

kgabryje:
Is anyone working on this? Or can I take it?

[...]

PR #14191, opened 2018-11-11
Add KeyboardEvent.code to synthetic event

jjspace:
#14102 
Passing along `KeyboardEvent.code` from the native Keyboard Event. This is a very complex calculated value is not effected by keyboard layout unlike other KeyboardEvent values so there is no way to create a polyfill for browsers without support.

### DIFF
--- a/packages/react-dom/src/events/SyntheticKeyboardEvent.js
+++ b/packages/react-dom/src/events/SyntheticKeyboardEvent.js
@@ -16,6 +16,7 @@ import getEventModifierState from './getEventModifierState';
  */
 const SyntheticKeyboardEvent = SyntheticUIEvent.extend({
   key: getEventKey,
+  code: null,
   location: null,
   ctrlKey: null,
   shiftKey: null,
--- a/packages/react-dom/src/events/SyntheticKeyboardEvent.js
+++ b/packages/react-dom/src/events/SyntheticKeyboardEvent.js
@@ -8,6 +8,7 @@
 import SyntheticUIEvent from './SyntheticUIEvent';
 import getEventCharCode from './getEventCharCode';
 import getEventKey from './getEventKey';
+import getEventCode from './getEventCode';
 import getEventModifierState from './getEventModifierState';
 
 /**
@@ -16,7 +17,7 @@ import getEventModifierState from './getEventModifierState';
  */
 const SyntheticKeyboardEvent = SyntheticUIEvent.extend({
   key: getEventKey,
-  code: null,
+  code: getEventCode,
   location: null,
   ctrlKey: null,
   shiftKey: null,
--- a/packages/react-dom/src/events/__tests__/SyntheticKeyboardEvent-test.js
+++ b/packages/react-dom/src/events/__tests__/SyntheticKeyboardEvent-test.js
@@ -451,6 +451,30 @@ describe('SyntheticKeyboardEvent', () => {
         });
       });
     });
+
+    describe('code', () => {
+      it('should throw (in dev) when KeyboardEvent.code is undefined', () => {
+        const node = ReactDOM.render(
+          <input
+            onKeyDown={e => {
+              return e.code;
+            }}
+          />,
+          container,
+        );
+        let evt = new KeyboardEvent('keydown', {
+          code: undefined,
+          bubbles: true,
+          cancelable: true,
+        });
+
+        if (__DEV__ && evt.code === undefined) {
+          expect(() => node.dispatchEvent(evt)).toThrowError();
+        } else {
+          node.dispatchEvent(evt);
+        }
+      });
+    });
   });
 
   describe('EventInterface', () => {
--- a/packages/react-dom/src/events/getEventCode.js
+++ b/packages/react-dom/src/events/getEventCode.js
@@ -1,0 +1,15 @@
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Code_values
+ * @param {object} nativeEvent
+ * @returns {string} Standardized keyboard code value if defined natively
+ */
+function getEventCode(nativeEvent: KeyboardEvent): string {
+  if (__DEV__ && nativeEvent.code === undefined) {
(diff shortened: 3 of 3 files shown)


packages/react-dom/src/events/SyntheticKeyboardEvent.js
```diff
@@ -16,6 +16,7 @@ import getEventModifierState from './getEventModifierState';
  */
 const SyntheticKeyboardEvent = SyntheticUIEvent.extend({
   key: getEventKey,
+  code: null,
```
Kovensky: In `__DEV__`, could this be made into a getter that would throw if `code` is missing from the event?

tagalong420: (1) Yeah it could if the code is missing or incorrect.  
(2) If it did proceed to load it would load wrong and kick the user out of the server and or domain. 
(3) Codes for all links and sites haft to be correctly placed in the script and diagram/program.

jjspace: I added a getter for this to throw an error if `code` is not defined on the `KeyboardEvent` object. I also tried to add tests for this but I couldn't find a way to force code to be undefined on the native event within node, any suggestions/help with that?

jjspace:
Can someone take another look at this to see if it's ready to merge? @Kovensky @tagalong420 ?

packages/react-dom/src/events/__tests__/SyntheticKeyboardEvent-test.js
```diff
@@ -451,6 +451,30 @@ describe('SyntheticKeyboardEvent', () => {
...
+          />,
+          container,
+        );
+        let evt = new KeyboardEvent('keydown', {
+          code: undefined,
+          bubbles: true,
+          cancelable: true,
+        });
+
+        if (__DEV__ && evt.code === undefined) {
```
Kovensky: Didn't know you could use this in tests too 🤔 

Do tests get run twice for this, or does it just always take one of the paths?

jjspace: It seems like it should work, I was going off what I saw in other test files like this [here](https://github.com/facebook/react/blob/master/packages/react-dom/src/__tests__/ReactComponent-test.js#L53) It seems like it just runs once and chooses the path dependent on the env variables. 

But I also was unable to find a way to force the `code` property of the `KeyboardEvent` to be undefined to actually test it as the constructor will set it to `''` automatically if set to undefined like in line 466.

philipp-spiess: @Jessidhia 

> Do tests get run twice for this, or does it just always take one of the paths?

[They're run twice 🙂](https://github.com/facebook/react/blob/679402a66b09e129d06f415cd976d18727cc8590/package.json#L105)

packages/react-dom/src/events/__tests__/SyntheticKeyboardEvent-test.js
```diff
@@ -451,6 +451,30 @@ describe('SyntheticKeyboardEvent', () => {
...
+          expect(() => node.dispatchEvent(evt)).toThrowError();
```
philipp-spiess: This should use the infrastructure we have in place for dev only warnings. Check out [this](https://github.com/facebook/react/blob/679402a66b09e129d06f415cd976d18727cc8590/packages/react/src/__tests__/ReactJSXElementValidator-test.js#L229-L235) and [this](https://github.com/facebook/react/blob/679402a66b09e129d06f415cd976d18727cc8590/packages/react/src/ReactElementValidator.js#L300-L308).

This will also make sure that the warning is _not_ emitted in production and that the message is indeed correct.

packages/react-dom/src/events/getEventCode.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Code_values
+ * @param {object} nativeEvent
+ * @returns {string} Standardized keyboard code value if defined natively
+ */
+function getEventCode(nativeEvent: KeyboardEvent): string {
+  if (__DEV__ && nativeEvent.code === undefined) {
```
philipp-spiess: I'm not sure if our builds will properly dead-code-eliminate this branch. I'd rather split it like this:

```js
if (__DEV__) {
  warning(nativeEvent.code !== undefined, "Warning here");
}
```

Although it should probably use `warningWithoutStack()` since it is triggered within an event handler which does not have a component stack _I think_.

https://github.com/facebook/react/blob/master/packages/shared/warningWithoutStack.js

packages/react-dom/src/events/getEventCode.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Code_values
+ * @param {object} nativeEvent
+ * @returns {string} Standardized keyboard code value if defined natively
```
philipp-spiess: I think you can remove the JSDocs type annotations since we use Flow anyway.

This file should also add the usual copyright header and a flow hint:

https://github.com/facebook/react/blob/679402a66b09e129d06f415cd976d18727cc8590/packages/create-subscription/src/createSubscription.js#L1-L8

packages/react-dom/src/events/getEventCode.js
```diff
@@ -0,0 +1,15 @@
...
+ */
+function getEventCode(nativeEvent: KeyboardEvent): string {
+  if (__DEV__ && nativeEvent.code === undefined) {
+    throw new ReferenceError(
+      'KeyboardEvent.code does not exist in this environment',
```
philipp-spiess: The message is not particularly clear. If we go for it, we need to make sure that it explains what's going on. What about:

> KeyboardEvent.code is not supported in the browser you’re currently using. You should either add a polyfill or use KeyboardEvent.key instead.

PR #14194, opened 2018-11-12
give  `canUseDOM` with a possibility to be a constant

gfx:
https://webpack.js.org/plugins/define-plugin/

Webpack's DefinePlugin has the ability to replace `typeof expr` to a constant in compile-time, which should lead to better dead-code-elimination.


gfx:
Thanks ;)

Issue #14188, opened 2018-11-11
Suspense


will-hart:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

I have a component which listens for resize events (via a BlueprintJS ResizeSensor). When loading a component dynamically with `lazy` / `Suspense`, an exception occurs as the resize sensor appears to be unmounted:

```
Uncaught Error: Unable to find node on an unmounted component.
    at invariant (29.chunk.js:86295)
    at findCurrentFiberUsingSlowPath (29.chunk.js:90628)
    at findCurrentHostFiber (29.chunk.js:90640)
    at findHostInstanceWithWarning (29.chunk.js:106349)
    at findDOMNode (29.chunk.js:106869)
    at ResizeSensor.componentDidUpdate (29.chunk.js:10535)
```

The resize sensor [should be removing listeners on unmount](https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/resize-sensor/resizeSensor.tsx#L73).

Demo: https://codesandbox.io/s/n4241q075l
Related: https://github.com/palantir/blueprint/issues/3141

**What is the expected behavior?**

No exception is thrown.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

Tested with:
- React 16.6.1, fails
- React 16.6.0, works

arianon:
Nope, it's broken on master, and after bisecting, I confirmed the regression was caused by #14083

gaearon:
>EDIT: Something else I found is that before the regression, ResizeSensor.componentDidUpdate doesn't execute, but now it does.

This sounds like a separate problem, can you provide a test case for it?

gaearon:
Here's a minimal repro with ReactDOM:

```js
const Lazy = React.lazy(() => new Promise(resolve => resolve({
  default() {
    return 'hello';
  }
})))

class Child extends React.Component {
  componentDidUpdate() {
    ReactDOM.findDOMNode(this);
  }

  render() {
    return null;
  }
}

class App extends React.Component {
  state = {
    suspend: false
  };
  handleClick = () => {
    this.setState({ suspend: true });
  };
  render() {
    return (
      <React.Suspense fallback="Loading">
        <Child />
        <button onClick={this.handleClick}>Suspend</button>
        <Child />
        {this.state.suspend && <Lazy />}
      </React.Suspense>
    );
  }
}

ReactDOM.render(
  <App />,
  document.getElementById('container')
);
```

[...]

PR #14198, opened 2018-11-12
Fix findDOMNode incorrectly failing inside Suspense

gaearon:
Builds on https://github.com/facebook/react/pull/14197. Second test ~~is failing~~ is now fixed.

Based on the test in https://github.com/facebook/react/issues/14188#issuecomment-437705392 found by @arianon. Unlike the issue from https://github.com/facebook/react/pull/14197, this one seems like it existed even before the Suspense refactorings. So maybe it was always there.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
@@ -262,8 +262,8 @@ describe('ReactDOMSuspensePlaceholder', () => {
     }
 
     ReactDOM.render(<App />, container);
-    // expect(log).toEqual([]);
+    expect(log).toEqual(['cDM']);
     ReactDOM.render(<App />, container);
-    // expect(log).toEqual(['cDM']);
+    expect(log).toEqual(['cDM', 'cDU']);
   });
 });
--- a/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
@@ -222,4 +222,48 @@ describe('ReactDOMSuspensePlaceholder', () => {
     await Lazy;
     expect(log).toEqual(['cDU first', 'cDU second']);
   });
+
+  // Regression test for https://github.com/facebook/react/issues/14188
+  it('can call findDOMNode() in a suspended component commit phase (#2)', () => {
+    let suspendOnce = Promise.resolve();
+    function Suspend() {
+      if (suspendOnce) {
+        let promise = suspendOnce;
+        suspendOnce = null;
+        throw promise;
+      }
+      return null;
+    }
+
+    const log = [];
+    class Child extends React.Component {
+      componentDidMount() {
+        log.push('cDM');
+        ReactDOM.findDOMNode(this);
+      }
+
+      componentDidUpdate() {
+        log.push('cDU');
+        ReactDOM.findDOMNode(this);
+      }
+
+      render() {
+        return null;
+      }
+    }
+
+    function App() {
+      return (
+        <Suspense fallback="Loading">
+          <Suspend />
+          <Child />
+        </Suspense>
+      );
+    }
+
+    ReactDOM.render(<App />, container);
+    // expect(log).toEqual([]);
+    ReactDOM.render(<App />, container);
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.internal.js
```diff
@@ -157,4 +162,108 @@ describe('ReactDOMSuspensePlaceholder', () => {
...
+      return (
+        <Suspense fallback="Loading">
+          <Suspend />
+          <Child />
+        </Suspense>
+      );
+    }
+
+    ReactDOM.render(<App />, container);
+    // expect(log).toEqual([]);
```
gaearon: I wasn't sure what's expected here. We'll need to fill those out before merging.

sophiebits:
I pushed a commit that I think fixes it. If findCurrentFiberUsingSlowPath is called on a fiber that has an alternate whose parent has no alternate, it previously incorrectly thought it had reached the root.

We hit this now and not before because of the Fragment we add under each Suspense child – it doesn't end up having an alternate.

packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -117,16 +117,17 @@ export function findCurrentFiberUsingSlowPath(fiber: Fiber): Fiber | null {
...
-    // If both copies of the parent fiber point to the same child, we can
-    // assume that the child is current. This happens when we bailout on low
-    // priority: the bailed out fiber's child reuses the current child.
-    if (parentA.child === parentB.child) {
+    // If both copies of the parent fiber point to the same child (or if there
+    // is only a single parent fiber), we can assume that the child is current.
+    // This happens when we bailout on low priority: the bailed out fiber's
+    // child reuses the current child.
+    let parentB = parentA.alternate;
+    if (!parentB || parentA.child === parentB.child) {
```
gaearon: `=== null` pls?

sophiebits: That's probably better. I had copied the truthy check from the old code.

gaearon: Maybe Seb *wanted* it to be slower. Bad bad findDOMNode.

gaearon:
Could we be making similar assumptions about presence of alternate elsewhere? Such as in event system traversal logic (eg traverseEnterLeave) or DOM Component Tree?

sebmarkbage:
I need to take 7 hours to think about this. Not so sure about this.

But also if you don’t want it to error you should test that the return value is what you expect it to be.

It’s better to error than return the wrong value.

packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -117,16 +117,17 @@ export function findCurrentFiberUsingSlowPath(fiber: Fiber): Fiber | null {
   let b = alternate;
   while (true) {
     let parentA = a.return;
-    let parentB = parentA ? parentA.alternate : null;
-    if (!parentA || !parentB) {
+    if (!parentA) {
```
sebmarkbage: I don’t think the logic is designed to make assumptions about which side is which. That’s why it’s a and b instead of say current and workInProgress.

This seems to think that A has some different property than B. Why is this?

sophiebits: I believe the logic should have really been `!parentA && !parentB`. Then we should enter the next `if` if either is null. But it is impossible for parentA to be null and parentB to be non-null. Hence the asymmetry. I could write it in a more symmetric way but it would have dead branches.

sophiebits:
OK you're right it returns the wrong value. Dang.

sebmarkbage:
What even is the right value? It doesn’t have well defined behavior since conceptually the fallback/main content could be first or last or not exist in the tree.

sebmarkbage:
I think that in this case you need to skip past the fragment and look at the two parents of the fragment to figure out if they converge.

sophiebits:
I think so too. But is it supposed to return the hidden node?

sophiebits:
ugh the logic to support suspense properly here looks so complicated

sebmarkbage:
Note that Suspense isn’t even done yet. The hydration of Suspense is looking to triple the complexity of Suspense and that’s not even taking findDOMNode into account.

gaearon:
This was fixed in https://github.com/facebook/react/pull/15312

Issue #14159, opened 2018-11-08
React.memo and React.lazy ignore propTypes


zpr-sioslife:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
```PropTypes.*.isRequired``` with ```React.memo``` doesn't get triggered in the console when the required prop is undefined.

```jsx
import React, { memo } from "react";
import ReactDOM from "react-dom";
import PropTypes from "prop-types";

// ********************************************
// It should appear two propTypes console errors
// ********************************************

// without memo(..): it will throw the propTypes error like it should
const MemoButton = memo(() => {
  return <button>Memo Button</button>;
});
MemoButton.propTypes = {
  memocolor: PropTypes.string.isRequired
};

// it will throw the error in the console
const Button = () => {
  return <button>Button</button>;
};
Button.propTypes = {
  color: PropTypes.string.isRequired
};

const rootElement = document.getElementById("root");
ReactDOM.render(
  <>
    <MemoButton />
    <Button />
  </>,
  rootElement
);
```
[![Edit 501jppv64n](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/501jppv64n?expanddevtools=1)

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
```js
// affected versions
"prop-types": "15.6.2",
"react": "16.6.1",
"react-dom": "16.6.1",
```

Kovensky:
It should work if the `propTypes` are attached to the component that you give as an argument to `React.memo`.

mjmena:
I came across this issue in the 16.7 alpha as well. I've updated the above codeSandbox to show that adding propTypes before or after memoizing the component has no effect. 
```
"prop-types": "^15.6.2",
"react": "^16.7.0-alpha.0",
"react-dom": "^16.7.0-alpha.0",
```
[![Edit zky2km9703](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/zky2km9703)

gaearon:
We need to decide on what we actually want to support. Arguably with `memo` it is valid to decorate either side (depending on how you define it). With `lazy` decorating outer side seems shady and we should probably warn if you attempt to do so.

ljharb:
You'd want adding or removing `memo` to be transparent - ie, not a breaking change - so I'd expect propTypes, defaultProps, etc to all work the same on it.

FredyC:
[...]

PR #14203, opened 2018-11-12
Use getComponentName() to reduce duplication

gaearon:
We couldn't do it before because it used to take fiber and not type as argument. But it's been taking type as an argument for a while now so we can use it.

I don't actually care about duplication per se but it's easier to reuse this instead of having to remember to add this in both places. Note actually reading `propTypes` still diverges (https://github.com/facebook/react/issues/14159) and we'd need to decide on semantics before fixing it.

### DIFF
--- a/packages/react/src/ReactElementValidator.js
+++ b/packages/react/src/ReactElementValidator.js
@@ -185,21 +185,16 @@ function validateChildKeys(node, parentType) {
  */
 function validatePropTypes(element) {
   const type = element.type;
-  let name, propTypes;
+  const name = getComponentName(type);
+  let propTypes;
   if (typeof type === 'function') {
     // Class or function component
-    name = type.displayName || type.name;
     propTypes = type.propTypes;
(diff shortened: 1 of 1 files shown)


sebmarkbage:
I was actually going to change it to take a Fiber instead so that it can correctly pick names when type !== elementType.

bvaughn:
> I was actually going to change it to take a Fiber instead so that it can correctly pick names when type !== elementType.

This sounds nice.

packages/react/src/ReactElementValidator.js
```diff
@@ -197,9 +197,6 @@ function validatePropTypes(element) {
   ) {
     // ForwardRef
     const functionName = type.render.displayName || type.render.name || '';
```
bvaughn: This var is no longer used

gaearon:
@sebmarkbage We'll still need something for the server renderer.

PR #14205, opened 2018-11-12
Save CI-built node_modules as build artifacts

bvaughn:
Let's save the `node_modules` created by the build script so that the script @acdlite is working on can publish them to NPM.

Example output [here](https://circleci.com/gh/facebook/react/12639#artifacts/containers/0):
screen shot 2018-11-12 at 3 38 16 pm

Which extracts to:
screen shot 2018-11-12 at 4 43 14 pm

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,4 +42,4 @@ jobs:
             - node_modules
 
       - store_artifacts:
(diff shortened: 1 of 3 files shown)


bvaughn:
Maybe we should zip this directory up and store it compressed, to make it easier to download?

PR #14218, opened 2018-11-13
Clear memoizedState on unmount of fiber to avoid memory leak

trueadm:
This is an alternative approach to https://github.com/facebook/react/pull/14153. This PR aims at clearing the memory leak when the associated fibers containing the references to the closures (stored in `memoizedState`) still exist after being unmounted. The fix is to simply `null` out the `memoizedState` field which seems to clear up the retained functions.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -728,9 +728,11 @@ function detachFiber(current: Fiber) {
   // itself will be GC:ed when the parent updates the next time.
   current.return = null;
   current.child = null;
+  current.memoizedState = null;
   if (current.alternate) {
     current.alternate.child = null;
     current.alternate.return = null;
+    current.alternate.memoizedState = null;
   }
 }
 
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -729,10 +729,18 @@ function detachFiber(current: Fiber) {
   current.return = null;
   current.child = null;
   current.memoizedState = null;
+  current.updateQueue = null;
+  current.firstEffect = null;
+  current.lastEffect = null;
+  current.nextEffect = null;
   if (current.alternate) {
-    current.alternate.child = null;
     current.alternate.return = null;
+    current.alternate.child = null;
     current.alternate.memoizedState = null;
+    current.alternate.updateQueue = null;
+    current.alternate.firstEffect = null;
+    current.alternate.lastEffect = null;
+    current.alternate.nextEffect = null;
   }
 }
 
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -726,13 +726,13 @@ function detachFiber(current: Fiber) {
   // get GC:ed but we don't know which for sure which parent is the current
   // one so we'll settle for GC:ing the subtree of this child. This child
   // itself will be GC:ed when the parent updates the next time.
+  // We do not null out the 'nextEffect' field as it causes tests to fail.
   current.return = null;
   current.child = null;
   current.memoizedState = null;
   current.updateQueue = null;
   current.firstEffect = null;
   current.lastEffect = null;
-  current.nextEffect = null;
   if (current.alternate) {
     current.alternate.return = null;
     current.alternate.child = null;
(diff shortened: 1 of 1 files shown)


acdlite:
Oh and `sibling`

sophiebits:
I still don't understand why this is necessary. Isn't the expectation that all the pointers to the unmounted Fiber are removed? In which case you don't need to null out all its fields.

acdlite:
@sophiebits The parent alternate could still have a reference via its child pointer

sophiebits:
Can't we null out .return.alternate.child? If I understand our pooling correctly, we would never end up reusing the parent alternate fiber without re-cloning anyway.

acdlite:
The code comment explains that we don't detach the alternate pointer because we're not sure if it's current but that should be fixable. Excepting resuming.

acdlite:
That's a riskier change, though. Return pointers are confusing. Could be worth it.

sophiebits:
I'm all about risky.

sophiebits:
I guess it's not safe. With this diff

```diff
diff --git a/packages/react-reconciler/src/ReactChildFiber.js b/packages/react-reconciler/src/ReactChildFiber.js
index bed780c2b..2fdc77ee0 100644
--- a/packages/react-reconciler/src/ReactChildFiber.js
+++ b/packages/react-reconciler/src/ReactChildFiber.js
@@ -256,6 +256,7 @@ function ChildReconciler(shouldTrackSideEffects) {
     }
     childToDelete.nextEffect = null;
     childToDelete.effectTag = Deletion;
+    childToDelete.return = returnFiber;
   }
 
   function deleteRemainingChildren(
diff --git a/packages/react-reconciler/src/ReactFiberCommitWork.js b/packages/react-reconciler/src/ReactFiberCommitWork.js
index 55a39ee64..eee0659dc 100644
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -726,6 +726,10 @@ function detachFiber(current: Fiber) {
   // get GC:ed but we don't know which for sure which parent is the current
   // one so we'll settle for GC:ing the subtree of this child. This child
   // itself will be GC:ed when the parent updates the next time.
+  const returnAlternate = current.return.alternate;
+  if (returnAlternate !== null) {
+    returnAlternate.child = null;
+  }
   current.return = null;
   current.child = null;
   if (current.alternate) {
diff --git a/packages/react-reconciler/src/ReactFiberCompleteWork.js b/packages/react-reconciler/src/ReactFiberCompleteWork.js
index fbd3b5b65..4da60e633 100644
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -725,6 +725,7 @@ function completeWork(
             currentFallbackChild.nextEffect = null;
           }
           currentFallbackChild.effectTag = Deletion;
+          currentFallbackChild.return = current.child;
         }
       }
 
```

all but one of our tests pass. But I realized that some of our traversals – like in findDOMNode – can run in between render and commit on the alternate tree and will mutate the .return pointers. So we can't rely on this to do the right thing.

If we could store the return pointer for a deletion somewhere else where it wouldn't be clobbered then I guess it would work. (Or if we were to findCurrentFiberUsingSlowPath during commit which I imagine is not desirable.)

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -726,11 +726,20 @@ function detachFiber(current: Fiber) {
   // get GC:ed but we don't know which for sure which parent is the current
   // one so we'll settle for GC:ing the subtree of this child. This child
   // itself will be GC:ed when the parent updates the next time.
+  // We do not null out the 'nextEffect' field as it causes tests to fail.
```
gaearon: Do we have a better understanding of this? "Causes tests to fail" doesn't say anything about the intent and whether those tests _should_ fail or if there's some other mistake.

This feels dangerous to me if we don't understand it. How do you know clearing `firstEffect` is safe if `nextEffect` is unsafe? Maybe we just accidentally hit some cases in our tests but not others.

gaearon: In general I'd prefer that we don't use vague wording like "causes tests to fail" in the comments. Either we know why it fails and explain the conceptual reason. Or we don't know, and should understand before merging.

trueadm: @gaearon This PR isn't in a state to be merged. I was more asking @acdlite for some guidance on this, thus my comments.

trueadm:
Follow up PR created https://github.com/facebook/react/pull/14276.

PR #14219, opened 2018-11-13
Validate propTypes for lazy and memo

gaearon:
Fixes https://github.com/facebook/react/issues/14159

### DIFF
--- a/packages/react/src/ReactElementValidator.js
+++ b/packages/react/src/ReactElementValidator.js
@@ -185,21 +185,16 @@ function validateChildKeys(node, parentType) {
  */
 function validatePropTypes(element) {
   const type = element.type;
-  let name, propTypes;
+  const name = getComponentName(type);
+  let propTypes;
   if (typeof type === 'function') {
     // Class or function component
-    name = type.displayName || type.name;
     propTypes = type.propTypes;
   } else if (
     typeof type === 'object' &&
     type !== null &&
     type.$$typeof === REACT_FORWARD_REF_TYPE
   ) {
-    // ForwardRef
-    const functionName = type.render.displayName || type.render.name || '';
-    name =
-      type.displayName ||
-      (functionName !== '' ? `ForwardRef(${functionName})` : 'ForwardRef');
     propTypes = type.propTypes;
   } else {
     return;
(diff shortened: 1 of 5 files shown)


packages/react/src/ReactElementValidator.js
```diff
@@ -177,6 +180,32 @@ function validateChildKeys(node, parentType) {
...
+      case REACT_MEMO_TYPE:
+        innerType = type.type;
+        break;
+      case REACT_LAZY_TYPE: {
+        innerType = refineResolvedLazyComponent(type);
+        break;
+      }
+    }
+    const innerPropTypes = getPropTypes(innerType);
+    return {...innerPropTypes, ...type.propTypes};
```
sophiebits: Shouldn’t this intersect if there are two checkers for the same key?

gaearon: Hmm. Right. I wasn't sure what's the correct mental model but intersecting makes sense.

gaearon: Meh. This seems too complicated. Either would have to always re-create them, or cache them somewhere. I think shadowing is reasonable, no?

sophiebits:
@sebmarkbage Can you explain your thinking here?

sebmarkbage:
You shouldn't be assigning type definitions on the consuming side. This is very different from `memo` which is used on the definition side.

You shouldn't be declaring the definitions of a type at the consuming side since it is not the owner of the interface. In fact, before you load it, you don't necessarily even know what it'll be if it's dynamic.

The argument about parity when you wrap memo or not on the definition side is strong since it's the owner of the code, but the same doesn't apply with `lazy` since `lazy` is on the consuming side. If you just remove lazy, you'd be mutating the definition side.

Further more, we might not even have an special React object here in the future so doesn't seem very future proof. It might e.g. be an AssetReference. At one point is was a Promise. Not sure what the type of `eager` will be once we add that so it'd need to have parity with that.

In fact, we should also warn about `defaultProps` being assigned to the lazy wrapper.

sophiebits:
Makes sense. Thanks.

packages/react/src/ReactElementValidator.js
sebmarkbage: If shadowing is reasonable, why not the whole set? It should probably be a warning that the type sets are ambiguous and only one should be provided. Seems like an error.

gaearon:
Superseded by https://github.com/facebook/react/pull/14298

PR #14226, opened 2018-11-14
[scheduler] Expose message event handler and key

acdlite:
Scheduler uses a message event to schedule a callback that fires after paint. This happens on every frame until the queue is empty. An unfortunate consequence is that every other message event handler also gets called on every frame; even if they exit immediately, this adds up to significant per- frame overhead. Until there's a better way to insert work after paint, a workaround is to monkey-patch `addEventListener` so there's only a single listener that wraps all the other ones. When a message
event is fired, we check if the event is Scheduler event. If it is, we call Scheduler, and skip the other event handlers entirely. (We could also use this same wrapper to defer the priority of the message event handlers, though that's a more drastic a change.)

This is a bit hacky, though, and probably doesn't belong in this package. It can be implemented in a wrapper, though, if we expose both the message key used by Scheduler message events and the Scheduler message listener.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -447,6 +447,26 @@ var requestHostCallback;
 var cancelHostCallback;
 var shouldYieldToHost;
 
+// Scheduler uses a message event to schedule a callback that fires after paint.
+// This happens on every frame until the queue is empty. An unfortunate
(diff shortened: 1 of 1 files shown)


acdlite:
@sebmarkbage thought of a better approach: use a MessageChannel. TIL!

PR #14233, opened 2018-11-14
Fix TODO task about Flow

ryota-murakami:
Original TODO commented Flow Error is not happen while current Flow version.
So I worked remove original type cast and improve `container._reactRootContainer` type.

As far as I was confirmed `container._reactRootContainer` will be `Root` or `null`.

- [assign Root](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOM.js#L531)
- [assign null](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOM.js#L692)

therefore I changed optional type to `Root | null`.

Feel free to close this PR if you don't think it's useful😀


ryota-murakami:
@vikaskyadav Thank you so much your review!

PR #14234, opened 2018-11-14
[scheduler] Post to MessageChannel instead of window

acdlite:
Scheduler needs to schedule a task that fires after paint. To do this, it currently posts a message event to `window`. This happens on every frame until the queue is empty. An unfortunate consequence is that every other message event handler also gets called on every frame; even if they exit immediately, this adds up to significant per-frame overhead.

Instead, we'll create a MessageChannel and post to that, with a fallback to the old behavior if MessageChannel does not exist.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -533,13 +533,14 @@ if (typeof window !== 'undefined' && window._schedMock) {
   };
 
   // We use the postMessage trick to defer idle work until after the repaint.
+  var port = null;
   var messageKey =
     '__reactIdleCallback$' +
     Math.random()
       .toString(36)
       .slice(2);
   var idleTick = function(event) {
-    if (event.source !== window || event.data !== messageKey) {
+    if (event.source !== port || event.data !== messageKey) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


packages/scheduler/src/Scheduler.js
```diff
@@ -533,13 +533,14 @@ if (typeof window !== 'undefined' && window._schedMock) {
   };
 
   // We use the postMessage trick to defer idle work until after the repaint.
+  var port = null;
   var messageKey =
```
sebmarkbage: The overhead of this is no longer necessary in this mechanism since it's a unique channel.

packages/scheduler/src/Scheduler.js
```diff
@@ -649,6 +647,19 @@ if (typeof window !== 'undefined' && window._schedMock) {
...
 
+  if (typeof MessageChannel === 'function') {
+    // Use a MessageChannel, if support exists
+    var channel = new MessageChannel();
+    channel.port1.onmessage = idleTick;
+    port = channel.port2;
+  } else {
+    // Otherwise post a message to the window. This isn't ideal because message
+    // handlers will fire on every frame until the queue is empty, including
+    // some browser extensions.
```
sebmarkbage: Do we need this fallback? IE9? Do we even support that anyway?

acdlite: Idk, for some reason I thought FB did but I guess we don't any longer

PR #14236, opened 2018-11-15
auto injection using loop statements

sunnibfgi:


aweary:
Thanks, but direct property assignment is easier to understand and likely faster.

Issue #14239, opened 2018-11-15
Regression: React 16 automatically marks first item of a select with size > 1 as checked


kamituel:
**Do you want to request a *feature* or report a *bug*?**

Bug - regression

**What is the current behavior?**

In react 16 when creating a `<select size="3">` (size > 1) the first child <option> is automatically getting marked as selected. In React 16 there does not seem to be a way to specify no <option> gets selected by default

https://codesandbox.io/s/m76ozkwm89

**What is the expected behavior?**

In React 15 unless you marked an option to be selected <option selected> no options were selected by default.

https://codesandbox.io/s/6894m3k9w

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

Reproducible in Chrome 72 and Firefox 59.

**Note:** this is almost a duplicate of #12200. In fact, I copied issue text and codesandbox examples, just changing `multiple` to `size="3"`.

highskillzz:
Can I take this issue?

nhunzaker:
Interesting!

Here's a manual DOM example of what we're doing:
https://codepen.io/nhunzaker/pen/gqEvjj

It also looks like this is not necessary if a select dropdown is already in "multiple" mode".

nhunzaker:
@highskillzz sorry to get back late to this. https://github.com/facebook/react/pull/14242 is close to a fix.

gaearon:
Fixed in 16.8.5.

PR #14242, opened 2018-11-15
Don't set the first option as selected in select tag with `size` attribute 

kulek1:
Fixes https://github.com/facebook/react/issues/14239

`size` attribute needs to be added to `select` before `option`'s are inserted.
Currently with `size` greater than 1 and without `multiple` attribute added, the first option is selected which is not expected result in that case.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMSelect-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMSelect-test.js
@@ -362,6 +362,32 @@ describe('ReactDOMSelect', () => {
     expect(node.options[2].selected).toBe(true); // gorilla
   });
 
+  it('does not select an item when size is initially set to greater than 1', () => {
+    let stub = (
+      <select size="2">
+        <option value="monkey">A monkey!</option>
+        <option value="giraffe">A giraffe!</option>
+        <option value="gorilla">A gorilla!</option>
+      </select>
+    );
+    const container = document.createElement('div');
+    const select = ReactDOM.render(stub, container);
+
+    expect(select.options[0].selected).toBe(false);
+    expect(select.options[1].selected).toBe(false);
+    expect(select.options[2].selected).toBe(false);
+
+    // Note: There is an inconsistency between JSDOM and Chrome where
+    // Chrome reports an empty string when no value is selected for a
+    // single-select with a size greater than 0. JSDOM reports the first
+    // value
+    //
+    // This assertion exists only for clarity of JSDOM behavior:
+    expect(select.value).toBe('monkey'); // "" in Chrome
+    // Despite this, the selection index is correct:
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -398,14 +398,21 @@ export function createElement(
...
       // See https://github.com/facebook/react/issues/13222
-      if (type === 'select' && props.multiple) {
+      // and https://github.com/facebook/react/issues/14239
+      if (type === 'select') {
         const node = ((domElement: any): HTMLSelectElement);
-        node.multiple = true;
+        if (props.multiple) {
+          node.multiple = true;
+        } else if (props.size) {
+          node.size = props.size;
```
nhunzaker: Small thing, but could you add a comment indicating that assigning `size` early is only necessary when `multiple` is false?

nhunzaker: It sounds like, this does essentially the same thing as setting `multiple=true` before options are appended because they visually become similar:

https://codepen.io/nhunzaker/pen/VgRQEy

kulek1:
@nhunzaker Sure, go ahead :)

nhunzaker:
I added a clarifying comment and some additional information to the DOM text fixture.

I also added a basic unit test, but uncovered a difference in behavior between JSDOM and Chrome. I left some information about that in the test for future work, like if we revisit the decision to assign attributes to selects after child options are attached.

nhunzaker:
Also pinged the DOM team for a review, so I'm not just approving my own changes :)

packages/react-dom/src/__tests__/ReactDOMSelect-test.js
```diff
@@ -362,6 +362,32 @@ describe('ReactDOMSelect', () => {
     expect(node.options[2].selected).toBe(true); // gorilla
   });
 
+  it('does not select an item when size is initially set to greater than 1', () => {
+    let stub = (
```
kulek1: You can use `const` here.

nhunzaker:
@philipp-spiess Ya. I did some browser testing, but I can't recall how comprehensive it was. I can take a sweep with the new fixture.

packages/react-dom/src/__tests__/ReactDOMSelect-test.js
nhunzaker: Ya. Thanks!

nhunzaker:
Okay! This checks out in:

- Desktop Safari 8, 12
- iOS Safari 9, 12
- Chrome 49, 72
- Firefox 52, 65
- IE 9, 10, 11
- Edge 18

I checked for:

- Initial selection is blank
- Selections are possible
- Only one selection is possible 

Thanks @kulek1 for sending in this PR!

PR #14245, opened 2018-11-15
Wrap shorthand CSS property collision warning in feature flag

acdlite:
Disables the recently introduced (#14181) warning for shorthand CSS property collisions by wrapping in a feature flag. Let's hold off shipping this until at least the next minor.


sophiebits:
:(

PR #14249, opened 2018-11-16
Reduce scheduler serialization overhead

developit:
In the process of switching to `MessageChannel`, it seems the postMessage call was modified to pass `"*"` (originally the target origin value from `window.postMessage`). This actually ends up triggering serialization, whereas passing `undefined` bypasses.

To save some investigation, passing a Number like `0` still incurs serialization overhead - `undefined` has special behavior.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -618,7 +618,7 @@ if (typeof window !== 'undefined' && window._schedMock) {
     frameDeadline = rafTime + activeFrameTime;
     if (!isMessageEventScheduled) {
       isMessageEventScheduled = true;
-      port.postMessage('*');
+      port.postMessage(undefined);
     }
   };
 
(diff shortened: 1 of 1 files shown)


sebmarkbage:
what about passing zero arguments?

bvaughn:
> what about passing zero arguments?

`postMessage` errors if you don't provide exactly 1 argument.

acdlite:
What about `null`? Would that work? We like to use `null` if it's intentional.

I'll go ahead and merge, though. Thanks Jason!

developit:
I admit I hadn't had time to check `null`. I need to verify with Mathias but I don't think it triggers the bypass. Source is here:
https://cs.chromium.org/chromium/src/third_party/blink/renderer/bindings/core/v8/script_value.h?type=cs&g=0&l=74

sophiebits:
I wouldn’t expect IsEmpty to return true for undefined; was this observably faster?

sebmarkbage:
@developit For the formality, mind asking to be added to our CLA?

developit:
Yup, working on it!

@sophiebits yes - consistently 5x faster. I'll set up a proper benchmark.

sophiebits:
:o

developit:
Update on the CLA - I was told to just sign as an individual, which I've just done.

PR #14267, opened 2018-11-18
Ensure React object shape is monomorphic

trueadm:
The main `React` object's property call-sites suffer from deoptimizing because we assign properties to the object after its creation, thus changing the object's shape (hidden class). This moves all properties back into the object literal.

### DIFF
--- a/packages/react/src/React.js
+++ b/packages/react/src/React.js
@@ -48,7 +48,9 @@ import {
 import ReactSharedInternals from './ReactSharedInternals';
 import {enableStableConcurrentModeAPIs} from 'shared/ReactFeatureFlags';
 
-const React = {
+// Please make sure that no properties are added to this object after its
+// creation. This ensures the object keeps the same shape for performance reasons.
+export default {
   Children: {
     map,
     forEach,
@@ -78,27 +80,27 @@ const React = {
   version: ReactVersion,
 
   __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: ReactSharedInternals,
+
+  ConcurrentMode: enableStableConcurrentModeAPIs
+    ? REACT_CONCURRENT_MODE_TYPE
+    : null,
+  Profiler: enableStableConcurrentModeAPIs ? REACT_PROFILER_TYPE : null,
+
+  unstable_ConcurrentMode: !enableStableConcurrentModeAPIs
+    ? REACT_CONCURRENT_MODE_TYPE
+    : null,
+  unstable_Profiler: !enableStableConcurrentModeAPIs
+    ? REACT_PROFILER_TYPE
+    : null,
+
+  useCallback: enableHooks ? useCallback : null,
+  useContext: enableHooks ? useContext : null,
+  useEffect: enableHooks ? useEffect : null,
+  useImperativeMethods: enableHooks ? useImperativeMethods : null,
+  useLayoutEffect: enableHooks ? useLayoutEffect : null,
+  useMemo: enableHooks ? useMemo : null,
+  useMutationEffect: enableHooks ? useMutationEffect : null,
+  useReducer: enableHooks ? useReducer : null,
+  useRef: enableHooks ? useRef : null,
+  useState: enableHooks ? useState : null,
 };
-
-if (enableStableConcurrentModeAPIs) {
-  React.ConcurrentMode = REACT_CONCURRENT_MODE_TYPE;
-  React.Profiler = REACT_PROFILER_TYPE;
-} else {
-  React.unstable_ConcurrentMode = REACT_CONCURRENT_MODE_TYPE;
-  React.unstable_Profiler = REACT_PROFILER_TYPE;
-}
-
-if (enableHooks) {
-  React.useCallback = useCallback;
-  React.useContext = useContext;
-  React.useEffect = useEffect;
-  React.useImperativeMethods = useImperativeMethods;
-  React.useLayoutEffect = useLayoutEffect;
-  React.useMemo = useMemo;
-  React.useMutationEffect = useMutationEffect;
-  React.useReducer = useReducer;
-  React.useRef = useRef;
-  React.useState = useState;
-}
-
-export default React;
--- a/packages/react/src/React.js
+++ b/packages/react/src/React.js
@@ -50,57 +50,107 @@ import {enableStableConcurrentModeAPIs} from 'shared/ReactFeatureFlags';
 
 // Please make sure that no properties are added to this object after its
 // creation. This ensures the object keeps the same shape for performance reasons.
-export default {
-  Children: {
-    map,
-    forEach,
-    count,
-    toArray,
-    only,
-  },
+let React;
+// We should remove the need for two objects if hooks are enabled by default.
+if (enableHooks) {
+  React = {
+    Children: {
+      map,
+      forEach,
+      count,
+      toArray,
+      only,
+    },
 
-  createRef,
-  Component,
-  PureComponent,
+    createRef,
+    Component,
+    PureComponent,
 
-  createContext,
-  forwardRef,
-  lazy,
-  memo,
+    createContext,
+    forwardRef,
+    lazy,
+    memo,
 
-  Fragment: REACT_FRAGMENT_TYPE,
-  StrictMode: REACT_STRICT_MODE_TYPE,
-  Suspense: REACT_SUSPENSE_TYPE,
+    Fragment: REACT_FRAGMENT_TYPE,
+    StrictMode: REACT_STRICT_MODE_TYPE,
+    Suspense: REACT_SUSPENSE_TYPE,
 
-  createElement: __DEV__ ? createElementWithValidation : createElement,
-  cloneElement: __DEV__ ? cloneElementWithValidation : cloneElement,
-  createFactory: __DEV__ ? createFactoryWithValidation : createFactory,
-  isValidElement: isValidElement,
+    createElement: __DEV__ ? createElementWithValidation : createElement,
+    cloneElement: __DEV__ ? cloneElementWithValidation : cloneElement,
+    createFactory: __DEV__ ? createFactoryWithValidation : createFactory,
+    isValidElement: isValidElement,
 
-  version: ReactVersion,
+    version: ReactVersion,
 
-  __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: ReactSharedInternals,
+    __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: ReactSharedInternals,
 
-  ConcurrentMode: enableStableConcurrentModeAPIs
-    ? REACT_CONCURRENT_MODE_TYPE
-    : null,
-  Profiler: enableStableConcurrentModeAPIs ? REACT_PROFILER_TYPE : null,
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiber.js
```diff
@@ -253,13 +253,13 @@ function FiberNode(
...
-  if (enableProfilerTimer) {
-    this.actualDuration = 0;
-    this.actualStartTime = -1;
-    this.selfBaseDuration = 0;
-    this.treeBaseDuration = 0;
-  }
+  this.actualDuration = 0;
+  this.actualStartTime = -1;
+  this.selfBaseDuration = 0;
+  this.treeBaseDuration = 0;
```
bvaughn: I thought this conditional was stripped from production and production+profiling builds (as part of uglification).

Removing it will cause non-profiling production builds to increase memory size, no?

bvaughn: Looking at the CJS bundles in NPM, it looks to me like the condition _is_ being removed from all but the dev builds.

bvaughn: FWIW, my above comment only applies to builds where this flag is static ^

trueadm: I was testing this on the FB bundle, sorry, I should have stated. We need to somehow make this static for FB too otherwise it seems most, if not all of our fibers on FB code end up getting deoptimized on V8 :(

bvaughn: It is static for production bundles– just not the dev bundle because we don't dead code eliminate it.

We choose which bundle to serve based on a gatekeeper. (We switch at the level of the react-dom bundle.)

trueadm: I remove the change, as it's not the fix. Something else must be touching those properties. Thanks for the help, I'll keep digging.

packages/react/src/React.js
```diff
@@ -48,37 +48,57 @@ import {
...
+  this.useImperativeMethods = null;
+  this.useLayoutEffect = null;
+  this.useMemo = null;
+  this.useMutationEffect = null;
+  this.useReducer = null;
+  this.useRef = null;
+  this.useState = null;
+}
+
+const React = new ReactObject();
```
Kovensky: The same could be accomplished, without requiring a constructor or making this more difficult to convert to ES modules, by just using conditionals for each assignment.

```js
React.ConcurrentMode = enableStableConcurrentModeAPIs ? REACT_CONCURRENT_MODE_TYPE : undefined;
// et al
```

trueadm: I tried that and the object still depots, likely because it has quite a few properties.

packages/react/src/React.js
```diff
@@ -48,37 +48,57 @@ import {
...
 
 if (enableStableConcurrentModeAPIs) {
   React.ConcurrentMode = REACT_CONCURRENT_MODE_TYPE;
```
bvaughn: Do these lines not affect anything? (Any reason we shouldn't move them and the hooks assignments into the function too?)

packages/react/src/React.js
```diff
@@ -48,57 +48,61 @@ import {
 import ReactSharedInternals from './ReactSharedInternals';
 import {enableStableConcurrentModeAPIs} from 'shared/ReactFeatureFlags';
 
-const React = {
-  Children: {
+// This uses a function constructor to enforce a monomorphic object. Using an object literal
+// along with the conditional branches causes this object to deoptimize frequently otherwise.
+function ReactObject() {
+  this.Children = {
```
bvaughn: I guess `React.Children.*` isn't used often enough to matter?

sophiebits:
cc @bmeurer

gaearon:
I don't fully understand if it's meant to be for internal or external usage.

Externally I think this wouldn't solve the problem because most people use React via webpack which already has its own object indirection in the middle. It's like `ReactImport.default.createElement`. So it probably deopts badly anyway, and this wouldn't fix it.

trueadm:
@bmeurer The property callsites on the React object show the V8 deopt message "wrong map". Moving them all into a constructor seems to stop the message from showing up.

sebmarkbage:
If we switch to ES modules exports in our syntax, then Rollup will convert this to an object for CJS anyway. Most likely packagers that consume ES module directly will convert it to an object when it is imported as `*` which happens often, and we're back to square one. ES module exports also can't be conditional so we'd have to figure out how to deal with that. It would be weird to have an export with the value `undefined` when you inspect this module in DEV tools.

sebmarkbage:
Can we switch to use ES exports in our code and fix this in Rollup?

Andarist:
@sebmarkbage What kind of change you'd like to see for this in rollup? Their output (module.exports with properties) seems to be sane for common use case

bmeurer:
I'm really having a hard time understanding this change and why it makes things better. On a high level this seems like we have a bug in V8 somewhere, since wrt. shape changes there shouldn't be any difference in the two approaches. I guess what I'm missing is the link to bundlers and what this ends up doing in the end.

@trueadm Would it be possible to provide a standalone .js file (ideally runnable in `d8` or at least `node`) that illustrates the problem? I.e. showing *good behavior* with constructor approach and *bad behavior* with object literal?

trueadm:
After speaking to @bmeurer about this, I think I've identified the deopt areas in the case of the React object. We're changing the shape of React object after creating it, both in the above (where we assign the hooks and the ConcurrentMode and profiler properties) and internally at FB where we change the shape yet again.

If I make the conditional properties part of the object literal and change the internal areas where we change the shape of the React object, I no longer see the same "wrong map` deopt message for the object literal approach. So I'll update the PR to reflect my findings and make a diff to change the internal shape change deopt. I'm a bit curious as to why I saw different deopt messages when using a function constructor, but that might have been something I was misreading when profiling.

packages/react/src/React.js
```diff
@@ -48,57 +48,109 @@ import {
...
+    useCallback,
+    useContext,
+    useEffect,
+    useImperativeMethods,
+    useLayoutEffect,
+    useMemo,
+    useMutationEffect,
+    useReducer,
+    useRef,
+    useState,
```
bvaughn: Curious why we forked for `enableHooks` rather than using ternaries like for `enableStableConcurrentModeAPIs` and `__DEV__`? e.g.
```js
React = {
  Children: {
    map,
    forEach,
    count,
    toArray,
    only,
  },

  createRef,
  Component,
  PureComponent,

  createContext,
  forwardRef,
  lazy,
  memo,

  Fragment: REACT_FRAGMENT_TYPE,
  StrictMode: REACT_STRICT_MODE_TYPE,
  Suspense: REACT_SUSPENSE_TYPE,

  createElement: __DEV__ ? createElementWithValidation : createElement,
  cloneElement: __DEV__ ? cloneElementWithValidation : cloneElement,
  createFactory: __DEV__ ? createFactoryWithValidation : createFactory,
  isValidElement: isValidElement,

  version: ReactVersion,

  __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: ReactSharedInternals,

  ConcurrentMode: enableStableConcurrentModeAPIs
    ? REACT_CONCURRENT_MODE_TYPE
    : null,
  Profiler: enableStableConcurrentModeAPIs ? REACT_PROFILER_TYPE : null,

  unstable_ConcurrentMode: !enableStableConcurrentModeAPIs
    ? REACT_CONCURRENT_MODE_TYPE
    : null,
  unstable_Profiler: !enableStableConcurrentModeAPIs
    ? REACT_PROFILER_TYPE
    : null,

  useCallback enableHooks ? useCallback : null,
  useContext enableHooks ? useContext : null,
  useEffect enableHooks ? useEffect : null,
  useImperativeMethods enableHooks ? useImperativeMethods : null,
  useLayoutEffect enableHooks ? useLayoutEffect : null,
  useMemo enableHooks ? useMemo : null,
  useMutationEffect enableHooks ? useMutationEffect : null,
  useReducer enableHooks ? useReducer : null,
  useRef enableHooks ? useRef : null,
  useState enableHooks ? useState : null,
};
```
No big deal I guess, just curious.

trueadm: I did that originally, but @gaearon suggested that it unnecessarily exposed the hooks APIs as properties.

trueadm: I had to draw the line somewhere and they seemed less bad to expose. Forking the object 4 times would have been pretty horrible. :D

bvaughn: Yeah I agree completely. I wasn't suggesting that 😅 I don't like even this fork but I think we've already talked about it more than it's worth so I'll 🤐

gaearon: Hmm to be clear I don't think it's great to expose `ConcurrentMode` = null either.

bvaughn: Would you feel any differently about `undefined`?

bvaughn:
I disagree that that proposal is preferable to having a few `undefined` properties on the exported object. Why should alpha builds be de-opted?

gaearon:
>Why should alpha builds be de-opted?

They've always been deopted so it's not a new issue. You shouldn't ideally be using them in production anyway. 

>I disagree that that proposal is preferable to having a few undefined properties on the exported object. 

It's not objective but if it appears on the React object, it will likely appear in autocomplete (at least in some editors, and in Chrome). It will appear in tools that display APIs based on `Object.keys()`. It will look missing and broken so we'll start getting bug reports and complaints about missing documentation. It will also look "confirmed" so at least for Hooks, we'll see people saying "why did it ship in stable". `unstable_` sets expectations accordingly. It also feels shady that `React.hasOwnProperty('useState')` or `React.hasOwnProperty('ConcurrentMode')` changes value in a minor release not containing it. Even though it's unlikely someone will use this exact feature check.

Kovensky:
These look like unreasonable expectations for an ES module, if you're ever
planning to migrate to that.

bvaughn:
> They've always been deopted so it's not a new issue. You shouldn't ideally be using them in production anyway.

Yeah I guess that's fair, but since people compare e.g. performance of hooks vs classes, it sucks if we're knowingly continuing to deopt.

bvaughn:
Not sure I agree with hooks "looking confirmed", or at least not that it's more important than deopting. The `hasOwnProperty` concern is a strong point though.

gaearon:
>These look like unreasonable expectations for an ES module, if you're ever
planning to migrate to that.

I don't understand what this comment refers to.

>Not sure I agree with hooks "looking confirmed", or at least not that it's more important than deopting. 

Deopting affects only alpha. Shipping an API name (even without implementation) affects a stable release.

bvaughn:
> Deopting affects only alpha. Shipping an API name (even without implementation) affects a stable release.

Hm, potentially. Maybe. 😄 Anyway, you feel more strongly about this than I do so~ 🙇

sophiebits:
I don't understand why adding expando properties would make the existing fast properties slower.

Andarist:
> > These look like unreasonable expectations for an ES module, if you're ever
planning to migrate to that.

> I don't understand what this comment refers to.

I think what @Kovensky has meant is that with ES modules you won't be able to export things conditionally because of static nature of ES modules.

gaearon:
https://github.com/facebook/react/pull/14309

gaearon:
I merged https://github.com/facebook/react/pull/14309 which does this.

PR #14268, opened 2018-11-18
expose htmlattribute -> htmlAttribute mapping in react-dom

probablyup:
There are a ton of libraries that reimplement this functionality, leading to unnecessary bundle weight. We can make things easier on the community by simply exporting what react-dom itself uses.

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -39,6 +39,7 @@ import {enableStableConcurrentModeAPIs} from 'shared/ReactFeatureFlags';
 import * as ReactDOMComponentTree from './ReactDOMComponentTree';
 import {restoreControlledState} from './ReactDOMComponent';
 import * as ReactDOMEventListener from '../events/ReactDOMEventListener';
+import possibleStandardNames from '../shared/possibleStandardNames';
 import {
   ELEMENT_NODE,
   COMMENT_NODE,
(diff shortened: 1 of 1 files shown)


gaearon:
>There are a ton of libraries that reimplement this functionality

Why? That seems like the problem.

For example, `styled-components` reliance on this is a conceptual issue with its *API*. Not with the implementation.

probablyup:
Because React warns if you pass through an attribute it doesn’t recognize.
Many libraries try to do filtering to prevent this since end users
frequently complain when they see console warnings. Making the list that
React uses for the warning importable and deduped by the various libraries
is a nice gesture, although just dropping the warning would be better IMO.

gaearon:
>Because React warns if you pass through an attribute it doesn’t recognize.

This hasn't been true since React 16. React lets you pass any attributes through.

It does warn for supported attributes. However, you should just let React do its thing, and not try to detect it at a library level.

The root of the issue, to my knowledge, is that some libraries attempt to *filter DOM attributes*. That's a problem by itself because it doesn't let people pass custom attributes to the DOM. If those libraries don't attempt to do that, then the whole thing is not an issue.

I'm saying that attempting to do filtering before React *is* a design flaw in these libraries. That they need to fix. We shouldn't be making it more convenient to create more libraries with this design flaw.

Am I missing something?

probablyup:
> This hasn't been true since React 16. React lets you pass any attributes through.

That's also not accurate. This warning is emitted for "linkStyle", for example:

```
Warning: React does not recognize the linkStyle prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase linkstyle instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

gaearon:
>This warning is emitted for "linkStyle", for example:

Right, but it tells you what to do. If you pass `linkstyle` it won’t warn. In either case it *will* add it to the DOM.

More importantly, this warning occurs for *any* camelCase name outside of the ones supported by React. I don’t see how exporting a map of *known* attribute would help you with it. 

Maybe let’s start from another end. If React *did* provide an export like this, what would you do with it? In development and in production.

You describe it as a common need but the only library I consistently see bit by this is styled components — because its API decisions early on relies on there existing a difference between “DOM props” and “other props” which is not the case anymore in React.

PR #14276, opened 2018-11-19
Clear fields on unmount of fiber to avoid memory leak

trueadm:
This PR is a follow up to https://github.com/facebook/react/pull/14218. In that PR, fields that we were unsure about were `null`ed out. This changes the `null`ing out of fields to the two that we know are safe to `null` out – specifically, `memoizedState` and `updateQueue`. This stops the memory leak we encounter when the associated fibers containing the references to the closures (stored in memoizedState) still exist after being unmounted

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -732,8 +732,8 @@ function detachFiber(current: Fiber) {
   current.updateQueue = null;
   const alternate = current.alternate;
   if (alternate !== null) {
-    alternate.child = null;
     alternate.return = null;
+    alternate.child = null;
     alternate.memoizedState = null;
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -728,9 +728,14 @@ function detachFiber(current: Fiber) {
...
   current.child = null;
-  if (current.alternate) {
-    current.alternate.child = null;
-    current.alternate.return = null;
+  current.memoizedState = null;
+  current.updateQueue = null;
+  const alternate = current.alternate;
+  if (alternate !== null) {
+    alternate.child = null;
+    alternate.return = null;
```
gaearon: Nit: reorder to match the order above? So it's easier to see they're the same.

Issue #14224, opened 2018-11-13
Warning: Expected Instance Props to Match Memoized Props 


BigPrimeNumbers:
When updating `react` and `react-dom` from 16.6.0 to 16.6.1 or 16.6.2, I get new warnings that did not occur prior:

```
Warning: Expected instance props to match memoized props before componentDidMount. This is likely due to a bug in React. Please file an issue.
Warning: Expected instance props to match memoized props before componentDidUpdate. This is likely due to a bug in React. Please file an issue.
Warning: Expected instance props to match memoized props before processing the update queue. This is likely due to a bug in React. Please file an issue.
```
all from `react-dom.development.js?ec6d:506`

I am unable to give you examples from the code as I am not sure where this is occurring and the stacktrace comes from deep in `react-dom` and is unfollowable from my code, though it happens on every page visited within the app. Reverting back to 16.6.0 versions removes the warning completely. 

There exists at least one other with the same problem: https://stackoverflow.com/questions/53272887/expected-instance-props-to-match-memoized-props-in-react-16

dxcqcv:
I have the same problem, and I can reproducing example using my [yoman generator](https://github.com/dxcqcv/generator-webpack-humble)

so the steps is

1. ` npm install -g yo` (if you do not install yoman)
2. `npm install -g generator-webpack-humble `
3. `yo webpack-humble`
4.  select `use React with react-css-modules and react-redux and redux-thunk`
5. run `npm run dev` after all dependents installed
6. then you'll see `Warning: Expected Instance Props to Match Memoized ` in browser console

And I already found the key of problem is `componentDidMount` hook, if I removed it the warning is disappeared

hardchopshoppirate:
Hi there,
For me, it's reproduced with `react: 16.6.*` and `rxjs: 6.3.3`. I found that without initializing of rxjs it works fine only on `react: 16.6.1`. Maybe it will be helpful for you to reproduce the issue.
Thanks!

edgeadjei:
I'm having the same issue after updating both react and react-dom to `^16.6.3`.  And with a bundle size of `3.6 mb` excluding node modules reproducing the issue will be rough.

adamthompson922:
I have been experiencing the same issue and have isolated it to using the latest `react-css-modules` with `create-react-class`. When migrating to use ES6 class components, the error disappears.

Here is a [minimal example](https://codesandbox.io/s/xppzppmpv4).

[...]

PR #14277, opened 2018-11-19
Warn about reassigning this.props

gaearon:
Related to https://github.com/facebook/react/issues/14224.

This adds a new warning when you reassign `this.props` during rendering. This has never been supported but apparently some libraries (like deprecated `react-css-modules`) do that.

Since this causes confusing warnings in the commit phase (where we aren't sure if it's our bug or not), I'm adding an earlier warning that fires in the render phase and has a component stack.

As a bonus, I reworded and added component name to the commit phase warning so that it's less difficult to figure out which component causes it.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -126,7 +126,7 @@ let didWarnAboutBadClass;
 let didWarnAboutContextTypeOnFunctionComponent;
 let didWarnAboutGetDerivedStateOnFunctionComponent;
 let didWarnAboutFunctionRefs;
-let didWarnAboutReassigningProps;
+export let didWarnAboutReassigningProps;
 
 if (__DEV__) {
   didWarnAboutBadClass = {};
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -126,12 +126,14 @@ let didWarnAboutBadClass;
 let didWarnAboutContextTypeOnFunctionComponent;
 let didWarnAboutGetDerivedStateOnFunctionComponent;
 let didWarnAboutFunctionRefs;
+let didWarnAboutReassigningProps;
 
 if (__DEV__) {
   didWarnAboutBadClass = {};
   didWarnAboutContextTypeOnFunctionComponent = {};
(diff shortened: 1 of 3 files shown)


gaearon:
(I also contacted @gajus who is the maintainer of `react-css-modules` and asked if it's possible to cut a patch that at least restores props back.)

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -491,14 +493,25 @@ function updateClassComponent(
...
     shouldUpdate,
     hasContext,
     renderExpirationTime,
   );
+  if (__DEV__) {
+    let inst = workInProgress.stateNode;
+    if (inst.props !== nextProps) {
+      warning(
+        didWarnAboutReassigningProps,
+        'Reassigning `this.props` is not supported and will lead to bugs.',
```
trueadm: I think we can word this message better. `lead to bugs` sounds a bit ambiguous – what bugs? Maybe it's best to link to a docs page or something instead.

trueadm:
On this note, should we also do the same for `this.state`?

packages/react-reconciler/src/ReactFiberBeginWork.js
gaearon: I don’t know what bugs. Bugs with incorrect props?

gaearon:
I tweaked the warning and also made it so that the subsequent warning about possible React bug isn't shown if we think the component is resposible.

gaearon:
>On this note, should we also do the same for this.state?

Maybe but I haven't seen this pattern in the wild (unlike `this.props`) so not sure it's worth the extra checks.

ppivatto:
Perfcet!

Roboonl:
> (I also contacted @gajus who is the maintainer of `react-css-modules` and asked if it's possible to cut a patch that at least restores props back.)

Have you heard anything from @gajus by any chance?

gajus:
https://github.com/facebook/react/issues/14224#issuecomment-440072024

Roboonl:
Thanks @gajus, the warning seems resolved with that change, @gaearon could you confirm?

gajus:
It would be nice to add a comment explaining what it is doing. I don't understand the purpose of the finally block. If first assignment does not work, whats the reason the second would?

gaearon:
I think you misunderstood my proposal: https://github.com/facebook/react/issues/14224#issuecomment-443031980

PR #14282, opened 2018-11-19
Remove `import * as` pattern from the codebase

sebmarkbage:
Whenever we do this, Rollup needs to materialize this as an object. This causes it to also add the Babel compatibility property which is unnecessary bloat. However, since when we use these, we leak the object on the `this` binding which could deoptimize inlining/DCE. Most of the time this just works because GCC is smart enough to see that it doesn't leak but not always. By avoiding this pattern it's more predictable.

If we really need an object we should export default an object.

Currently there is an exception for DOMTopLevelEventTypes since listing out the imports is a PITA and it doesn't escape so it should get properly inlined. We should probably move to a different pattern to avoid this for consistency though.

Do we need to add this as a lint warning?

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -68,7 +68,12 @@ import invariant from 'shared/invariant';
 import warningWithoutStack from 'shared/warningWithoutStack';
 
 import ReactFiberInstrumentation from './ReactFiberInstrumentation';
-import * as ReactCurrentFiber from './ReactCurrentFiber';
+import {
+  getStackByFiberInDevAndProd,
+  phase as ReactCurrentFiberPhase,
+  resetCurrentFiber,
+  setCurrentFiber,
+} from './ReactCurrentFiber';
 import {
   now,
   scheduleDeferredCallback,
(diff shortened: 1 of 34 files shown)


packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -68,7 +68,12 @@ import invariant from 'shared/invariant';
 import warningWithoutStack from 'shared/warningWithoutStack';
 
 import ReactFiberInstrumentation from './ReactFiberInstrumentation';
-import * as ReactCurrentFiber from './ReactCurrentFiber';
+import {
+  getStackByFiberInDevAndProd,
+  phase as ReactCurrentFiberPhase,
```
bvaughn: I don't see a pattern in what you use an uppercase first letter for (e.g. `ReactCurrentFiberPhase`) and what stays lower case (e.g. `getInstance`)

sebmarkbage: Either if a) there's more than one with the same name, b) I'm scared of overlap from a local variable or something.

But I think we should probably rename these on a case-by-case basis to something designed for being used with named imports.

PR #14291, opened 2018-11-20
Fix support for mixing react-dom/server@16.6 and react@<16.6

gaearon:
This prevents an obscure issue if you use `react-dom/server` from 16.6 but `react` from 16.5 or earlier. We decided to support this a while ago and it's easy to keep working.

I didn't bother writing a test because I'm lazy.


sebmarkbage:
The idea here is that it should just do a hidden class check, then put the current value in the register and jump over it with the same label as it would use for the loop. Thereby making this code inline-friendly (in the VM).

This check would add additional logic that at least deals with the *other case* and might need more registers.

gaearon:
Doh. I misread the DEV closing brace. 😳

PR #14297, opened 2018-11-21
Use |0 to coerce to number

gaearon:
Follow-up to https://github.com/facebook/react/pull/14291. Should be less work, right?

I don't think we can make this simpler without reverting the fix. The bug is very subtle and I think it would suck to keep shipping it.


gaearon:
I think it's fine.

PR #14298, opened 2018-11-21
Validate propTypes for lazy() and memo() and warn about invalid patterns

gaearon:
Fixes https://github.com/facebook/react/issues/14159.

Supersedes https://github.com/facebook/react/pull/14219.

### Lazy

* We now warn about trying to assign `defaultProps` and `propTypes` to outer wrapper (but allow it).
  * When `propTypes` is on the outer `lazy()` wrapper, we have undefined behavior. In some cases outer `propTypes` might validate, in other it might not. The warning will discourage you regardless, and ask you to define it inside instead.
* We now validate `propTypes` defined on the inner lazy component. Coverage added.
* Better error message if you nest `lazy()`. Coverage added.

### Memo

* We now validate `propTypes` both on the inner and the outer component. Coverage added.

A bunch of the complexity is because for lazy code paths, we have to validate in the reconciler (because we don't have `defaultProps` until lazy resolves). We want to validate initial props too, and not just updates. So we can't just descend down in `createElement` when possible — as this would "lose" the props from the first lazy mount.

Additionally, in `memo` code paths we take care to validate both outer and inner props. This is a bit messy because of the "upgrade" scenario (which might also be combined with lazy). I added a bunch of tests to cover different combinations.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
@@ -510,7 +510,7 @@ describe('ReactLazy', () => {
     expect(root).toMatchRenderedOutput('Friends Bye');
   });
 
-  it('warns about defining propTypes on the outer wrapper', async () => {
+  it('warns about defining propTypes on the outer wrapper', () => {
     const LazyText = lazy(() => fakeImport(Text));
     expect(() => {
       LazyText.propTypes = {hello: () => {}};
--- a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
@@ -510,6 +510,18 @@ describe('ReactLazy', () => {
     expect(root).toMatchRenderedOutput('Friends Bye');
   });
 
+  it('warns about defining propTypes on the outer wrapper', async () => {
+    const LazyText = lazy(() => fakeImport(Text));
+    expect(() => {
+      LazyText.propTypes = {hello: () => {}};
+    }).toWarnDev(
+      'React.lazy(...): It is not supported to assign `propTypes` to ' +
+        'a lazy component import. Either specify them where the component ' +
+        'is defined, or create a wrapping component around it.',
+      {withoutStack: true},
+    );
+  });
+
   it('includes lazy-loaded component in warning stack', async () => {
     const LazyFoo = lazy(() => {
       ReactTestRenderer.unstable_yield('Started loading');
--- a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
@@ -619,12 +619,12 @@ describe('ReactLazy', () => {
     expect(root).toMatchRenderedOutput('22');
 
     // Update
-    root.update(
-      <Suspense fallback={<Text text="Loading..." />}>
-        <LazyAdd inner={false} outer={false} />
-      </Suspense>,
-    );
     expect(() => {
+      root.update(
+        <Suspense fallback={<Text text="Loading..." />}>
+          <LazyAdd inner={false} outer={false} />
+        </Suspense>,
+      );
       root.unstable_flushAll();
     }).toWarnDev(
       'Invalid prop `inner` of type `boolean` supplied to `Add`, expected `number`.',
@@ -804,12 +804,12 @@ describe('ReactLazy', () => {
     expect(root).toMatchRenderedOutput('Inner default text');
 
     // Update
-    root.update(
-      <Suspense fallback={<Text text="Loading..." />}>
-        <LazyText text={null} />
-      </Suspense>,
-    );
     expect(() => {
+      root.update(
+        <Suspense fallback={<Text text="Loading..." />}>
+          <LazyText text={null} />
+        </Suspense>,
+      );
       root.unstable_flushAll();
     }).toWarnDev(
       'The prop `text` is marked as required in `T`, but its value is `null`',
--- a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
@@ -570,6 +570,42 @@ describe('ReactLazy', () => {
     expect(root).toMatchRenderedOutput('22');
   });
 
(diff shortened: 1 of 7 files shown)


packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
```diff
@@ -461,6 +461,67 @@ describe('ReactLazy', () => {
...
+    root.update(
+      <Suspense fallback={<Text text="Loading..." />}>
+        <LazyText inner="Friends" />
+      </Suspense>,
+    );
+    expect(root).toFlushAndYield(['Friends Bye']);
+    expect(root).toMatchRenderedOutput('Friends Bye');
+  });
+
+  it('warns about defining propTypes on the outer wrapper', async () => {
```
raunofreiberg: Redundant async?

gaearon:
@sebmarkbage Ready for review.

gaearon:
Strengthened the test coverage a little bit. Now if I change any of the added checks, some tests fail.

gaearon:
>Can we put the logic in mountLazyComponent after calling readLazyComponentType instead of in each update* function? Then I think you only need special logic to handle SimpleMemo w/ lazy.

That wouldn't handle updates, would it? They get re-tagged. Alternatively I can make `createElement` always try to resolve lazy props for validation but then this logic is duplicated far away.

sophiebits:
Hm good point. Maybe that is the simplest though? So that createElement always recurses as deep as it synchronously can – and then when we get new info allowing us to go deeper (i.e., in mountLazyComponent) we check the new depth. Not sure if that would result in duplicate warnings.

gaearon:
If I remember correctly that's what I started with, and then I bumped into issues with prop resolving (e.g. memo). I can try again for sure.

gaearon:
OK, one way to make it work is to simply call `createElement` from the `Begin` branch, and consolidate all logic in `createElement`. Will push something soon.

gaearon:
Should be ready for another review

packages/react/src/ReactLazy.js
```diff
@@ -8,13 +8,52 @@
...
+  let lazyType = {
     $$typeof: REACT_LAZY_TYPE,
     _ctor: ctor,
     // React uses these fields to store the result.
     _status: -1,
     _result: null,
   };
+
+  if (__DEV__) {
+    // In production, this would just set it on the object.
```
sophiebits: I guess it would be enumerable in prod but not in dev? prolly ok tho

packages/react/src/ReactElementValidator.js
```diff
@@ -177,57 +180,93 @@ function validateChildKeys(node, parentType) {
...
-    name =
-      type.displayName ||
-      (functionName !== '' ? `ForwardRef(${functionName})` : 'ForwardRef');
     propTypes = type.propTypes;
-  } else {
-    return;
+  } else if (typeof type === 'object') {
+    switch (type.$$typeof) {
+      case REACT_FORWARD_REF_TYPE: {
+        propTypes = type.propTypes;
```
sophiebits: do we warn against propTypes on the inner fn?

packages/react/src/ReactElementValidator.js
```diff
@@ -177,57 +180,93 @@ function validateChildKeys(node, parentType) {
...
-      'getDefaultProps is only used on classic React.createClass ' +
-        'definitions. Use a static property named `defaultProps` instead.',
-    );
+  // A type may have an inner type (e.g. React.memo or React.lazy).
+  // That's what the recursive case is for.
+  if (innerType) {
+    // Inner type may have its own defaultProps
+    let innerProps = props;
+    if (innerType && innerType.defaultProps) {
+      innerProps = {...props};
```
sophiebits: feels like a bit of a waste we need to do this and throw it away, I guess

sophiebits:
I wanna think on this a little more

packages/react/src/ReactLazy.js
gaearon: I figured it's bad regardless but can change

packages/react/src/ReactElementValidator.js
gaearon: Not the common case though? I'm assuming `defaultProps` is not super common by itself, and especially when wrapped.

packages/react/src/ReactElementValidator.js
gaearon: yes

packages/react/src/ReactLazy.js
gaearon: fixed

sebmarkbage:
Currently we do a lot of validation in the createElement callsite because it gives a nicer stack trace. However, we also have the compiler meta data that adds source info to JSX so we can recover that information anyway. We also have the ability to synthesize component stacks that give a decent error.

PropTypes is not a first-class concept. Static types has won. defaultProps is likely a dying concept given destructuring taking its place. Ideally key validation should be a static concept too like Hooks lints. Memo and lazy has forced us into moving the residuals of this into the reconciler anyway. This means that we've already made breaking runtime concessions. Reading `element.props` is different depending on if the component was lazy or wrapped in memo. So this whole early evaluation thing is kind of broken already. The inlining across modules that it was meant to support never really happened neither.

There are benefits to simplifying the semantics of JSX callsites like avoiding the runtime dependencies and just having this built-in in runtime helpers like babel-helpers. That'll make it a lot harder to make changes to. E.g. we couldn't have made any of the changes this PR does to ElementValidator. I think we should strive towards it being close to `type.bind(null, props)`.

So I think we should opt towards moving all this logic into the reconciler and not do any changes to the createElement validation.

gaearon:
So you’re for reverting https://github.com/facebook/react/commit/2e77ca47fe80ebe6595333542a8c5c138c68643f and merging?

I want to get this in ASAP because the longer we wait the more violations will uncover in next patch.

sebmarkbage:
@gaearon yes. I believe I reviewed the commit before that and it seemed to make sense overall.

Kovensky:
There is one notable place where defaultProps work differently from destructuring -- the React DevTools. If you do use defaultProps, the props with defaulted values show up in the devtools panel, but if you use destructuring omitted props don't show up at all.

This could probably be solved in another way through emitting additional compiler metadata in development builds, and if you have a false negative there it's still better than the current destructuring experience, but then the compiler has to know on which functions/classes it's ok to attach it to...

PR #14301, opened 2018-11-22
Enable hooks in fabric

sahrens:
because hooks are awesome.


brunolemos:
Were the [issues](https://github.com/facebook/react-native/issues/21967#issuecomment-438483674) fixed? E.g. the `useEffect` delay.

sebmarkbage:
I think you need to update this one too: https://github.com/facebook/react/blob/master/packages/shared/forks/ReactFeatureFlags.native-fabric-fb.js#L18

PR #14309, opened 2018-11-22
Prevent deopts from modifying exports object in stable builds

gaearon:
Spiritual successor to https://github.com/facebook/react/pull/14267.
Only fixes it for stable builds (which IMO is a fair tradeoff).

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -770,6 +770,7 @@ const ReactDOM: Object = {
 
   flushSync: flushSync,
 
+  unstable_createRoot: createRoot,
   unstable_flushControlled: flushControlled,
 
   __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: {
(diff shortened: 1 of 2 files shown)


Andarist:
Are those shapes defined at build time? Or a feature (like hooks) can be enabled at runtime? If it's static (build time) why not parse the entry point with babel and output a monomorphic shape for every type of the build?

gaearon:
Sounds like it would be complicated and not worth the effort? Especially given that it can introduce other bugs.

Andarist:
Sure - it would be slightly more convoluted, but looking at how many custom things are done here it wouldnt really increase the overall complexity of the build script that much - especially that such babel transform should be really easy and the code to write it would be minimal (I could help with it if you like this approach).

I totally understand if you don't want this - I only thought that this should be doable at (imho) low cost and that this option could be considered (& rejected if needed 😉).

gaearon:
If you send a PR I can take a look 🙂 maybe I’m overestimating complexity. I’m just worried it will make the boundary between what’s static and what dynamic more confusing, and potentially can break things.

And we’d need a different strategy for ESM anyway.

Andarist:
> If you send a PR I can take a look 🙂 maybe I’m overestimating complexity. I’m just worried it will make the boundary between what’s static and what dynamic more confusing, and potentially can break things.

Gonna certainly do it! We can also think how to test it - maybe require the created file after the build and compare exports shape snapshot?

> And we’d need a different strategy for ESM anyway.

Beauty of this solution is that it actually wouldnt require separate solution - creating exported shape at build time gives us freedom to export any shape we want. Removing object properties (with undefined/null values) isnt any different than removing whole export statements from the AST manipulation point of view :)

gaearon:
I'll just take it in — shouldn't harm and might help.

trueadm:
Checked this, deopt for "wrong map" doesn't appear :D nice work @gaearon

PR #14314, opened 2018-11-23
Fix regression: Errors not emitted in streams

voxpelli:
Regression introduced in #14182 resulted in errors no longer being emitted on streams, breaking consumers.

The cause is a faulty implementation of the `destroy()` method of `ReactMarkupReadableStream`, which resulted in no `error` event being thrown even though one expected it.

This patch fixes that + adds tests for the fixes.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactServerRendering-test.js
+++ b/packages/react-dom/src/__tests__/ReactServerRendering-test.js
@@ -556,6 +556,52 @@ describe('ReactDOMServer', () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


greaveselliott:
Hi, when will this change be released?

gaearon:
If you need it urgently you can build React from master and get the fix. I can't give you an exact days — we generally release things when we feel they're stable, and this means giving some time for other changes to shake out. Probably within a few weeks.

decompil3d:
Fix appears to be in 16.7.0

Issue #14326, opened 2018-11-26
Async useEffect is pretty much unreadable


linvain:
![heh](https://user-images.githubusercontent.com/42490556/49019303-ac247f80-f19e-11e8-8ed9-633b4e0628a7.png)

The only alternative is using `then()` chains which aren't very readable at their own.

gaearon:
I think you're making it more complicated than it needs to be.

```js
async function fetchMyAPI() {
  let url = 'http://something';
  let config = {};
  const response = await myFetch(url);
  console.log(response);
}

useEffect(() => {
  fetchMyAPI();
}, []);
```

Longer term we'll discourage this pattern because it encourages race conditions. Such as — anything could happen between your call starts and ends, and you could have gotten new props. Instead, we'll recommend Suspense for data fetching which will look more like

```js
const response = MyAPIResource.read();
```

and no effects. But in the meantime you can move the async stuff to a separate function and call it.

lxe:
What if I need to close over `setSomething(value returned from async)`?

```js 
async function fetchWidgets() {
  return await fetch('./api/widgets').then(r => r.json())
}

// ...
const [widgets, setWidgets] = useState([]);
// Need another async wrapper function to close over `setWidgets`?
useEffect(() => (async () => setWidgets(await fetchWidgets))());
```

RiJung:
@gaearon 
now I should use `useEffect`  instead of `componentDidMount` into a React Component with Hooks.

Scenario "initial unique call to a server":
To accomplished this, DependencyList ( second argument of useEffect) in useEffect should every time an empty array otherwise the application will send every state change a fetch call to the server.

it' means ,  this it the best practise to getData
```typescript
    useEffect(() => {
        console.log("useEffect, fetchData here");
    }, []);
````

Full Code here
<details>

```typescript 
import React, { useState, useEffect } from 'react';

export function CaseAssignDropDown() {
    const [count, setCount] = useState(0);
    useEffect(() => {
        console.log("useEffect, fetchData here");
    }, []);
    /*
    [] -> get data once
    [count] or removed [] -> get data each click
    */
    return (
        <div>
            <p>You clicked {count} times</p>
            <button onClick={() => setCount(count + 1)}>
                Click me
                </button>
        </div>
    );
}
``` 

</details>

gaearon:
Please ask on StackOverflow and feel free to link to your question from here. We don’t use issues for support because answers get lost on GitHub.

codeaid:
[...]

PR #14327, opened 2018-11-26
Make useEffect(async) warning more verbose

gaearon:
Adds a more detailed message. People still get tripped over this:

* https://github.com/facebook/react/issues/14326
* https://mobile.twitter.com/bardiarastin/status/1064144863388680192

New message:

screen shot 2018-11-26 at 3 53 12 pm

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -337,7 +337,7 @@ function commitHookEffectList(
                   'nothing.%s%s',
                 typeof destroy.then === 'function'
                   ? '\n\nIt looks like you wrote useEffect(async () => ...) or returned a Promise. ' +
-                    'Instead, you may write an async function separately, ' +
+                    'Instead, you may write an async function separately ' +
                     'and then call it from inside the effect:\n\n' +
                     'async function fetchComment(commentId) {\n' +
                     '  // You can await here\n' +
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -336,9 +336,17 @@ function commitHookEffectList(
                 'useEffect function must return a cleanup function or ' +
                   'nothing.%s%s',
                 typeof destroy.then === 'function'
-                  ? ' Promises and useEffect(async () => ...) are not ' +
-                    'supported, but you can call an async function inside an ' +
-                    'effect.'
+                  ? '\n\nIt looks like you wrote useEffect(async () => ...) or returned a Promise. ' +
+                    'Instead, you may write an async function separately, ' +
```
sophiebits: no comma after separately please

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -336,9 +336,17 @@ function commitHookEffectList(
...
+                    'and then call it from inside the effect:\n\n' +
+                    'async function fetchComment(commentId) {\n' +
+                    '  // You can await here\n' +
+                    '}\n\n' +
```
Kovensky: More of a question: would it be better to write this function inside the effect callback instead?

Supposedly, the engine would only have to evaluate the function declaration when the effect callback is called, but I'm not sure it'd help anyway as it'd still have to create the shared closure object.

At any rate, writing it outside the component entirely would definitely be better.

gaearon: I don't really have an opinion on this but it's nice to show it doesn't have to be inside.

alippai:
Is there a tracking bug for data fetching solution (possibly with error handling / timeout included)? Just being curious whether it will fit other use-cases too e.g Fetch API streaming body or Web Worker postMessage

PR #14329, opened 2018-11-26
[FIX] Allow shallow renderer to work with new API

mprobber:

Hi,
I really like the new `contextType` API in 16.6, but the shallow renderer doesn't seem to provide support for it.  This pull request allows you to pass in an object for `context`, and if `contextType` is defined, will shallow render the component with the provided context, as it did for the old context API.

### DIFF
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -144,11 +144,7 @@ class ReactShallowRenderer {
     this._rendering = true;
     this._element = element;
 
-    if (element.type.contextType) {
-      this._context = context;
-    } else {
-      this._context = getMaskedContext(element.type.contextTypes, context);
-    }
+    this._context = getContext(element, context);
 
     if (this._instance) {
       this._updateClassComponent(element, this._context);
@@ -377,6 +373,14 @@ function shouldConstruct(Component) {
   return !!(Component.prototype && Component.prototype.isReactComponent);
 }
 
+function getContextTypeContext(contextType, context) {
+  if (context !== undefined && context !== emptyObject) {
+    return context;
+  }
+
+  return contextType._currentValue;
+}
+
 function getMaskedContext(contextTypes, unmaskedContext) {
   if (!contextTypes) {
     return emptyObject;
@@ -386,6 +390,14 @@ function getMaskedContext(contextTypes, unmaskedContext) {
     context[key] = unmaskedContext[key];
   }
   return context;
+}
+
+function getContext(element, contextOption) {
+  if (element.type.contextType) {
+    return getContextTypeContext(element.type.contextType, contextOption);
+  }
+
+  return getMaskedContext(element.type.contextTypes, contextOption);
 }
 
 export default ReactShallowRenderer;
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -143,7 +143,12 @@ class ReactShallowRenderer {
 
     this._rendering = true;
     this._element = element;
-    this._context = getMaskedContext(element.type.contextTypes, context);
+
+    if (element.type.contextType) {
+      this._context = context;
+    } else {
+      this._context = getMaskedContext(element.type.contextTypes, context);
+    }
 
     if (this._instance) {
       this._updateClassComponent(element, this._context);
--- a/packages/react-test-renderer/src/__tests__/ReactShallowRenderer-test.js
+++ b/packages/react-test-renderer/src/__tests__/ReactShallowRenderer-test.js
@@ -594,6 +594,22 @@ describe('ReactShallowRenderer', () => {
     expect(result).toEqual(<div />);
   });
 
+  it('can shallowly render components with contextType and default context', () => {
+    const SimpleContext = React.createContext('hello world');
+
+    class SimpleComponent extends React.Component {
+      static contextType = SimpleContext;
+
+      render() {
+        return <div>{this.context}</div>;
+      }
+    }
+
+    const shallowRenderer = createRenderer();
+    const result = shallowRenderer.render(<SimpleComponent />);
+    expect(result).toEqual(<div>hello world</div>);
+  });
+
   it('can shallowly render components with contextType', () => {
     const SimpleContext = React.createContext('hello world');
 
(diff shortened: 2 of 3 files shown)


packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -143,7 +143,12 @@ class ReactShallowRenderer {
 
     this._rendering = true;
     this._element = element;
-    this._context = getMaskedContext(element.type.contextTypes, context);
+
+    if (element.type.contextType) {
```
gaearon: This should only be enabled for class components.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -143,7 +143,12 @@ class ReactShallowRenderer {
...
+      this._context = context;
```
gaearon: This is the legacy context object. It has no relation to the new context API you're implementing.

mprobber:
@gaearon

I agree with you that the shallow renderer should use the value of the default context if no context is provided, but I'm not really sure about rendering two levels deep and not being transparent about it in the shallow renderer for a select few components.

Looking at https://github.com/facebook/react/blob/c954efa70f44a44be9c33c60c57f87bea6f40a10/packages/react-reconciler/src/ReactFiberClassComponent.js#L362 there is no way for a component to support both the `contextType` and the legacy `contextTypes` APIs, so it doesn't seem like it would be problematic to use the same argument to set context on a component (just an object, used to set this.context).   I don't have too much context (*ba-dum-tss*) about implementation details though, so I'm curious if there's something I'm missing.  I'll update the PR so the default context renders if none is provided.

jfirebaugh:
> It doesn't seem like it would be problematic to use the same argument to set context on a component (just an object, used to set this.context).

Coming here from https://github.com/airbnb/enzyme/issues/1553, via https://stackoverflow.com/questions/53711166/unable-to-set-context-in-an-enzyme-test, and that's pretty much exactly how I expected it to work.

minznerjosh:
To @gaearon's point, using the second argument to support `contextType` would create some funkiness in enzyme.

It seems reasonable at first because, with the plain react shallow renderer, you're only ever rendering one component at a time, so you have a lot of control over what context you pass when you render a component; you _know_ not to pass legacy context if the component you're testing is using `contextType`.

But Enzyme has the concept of `dive()`ing, allowing you to render components further down in the tree. And, most importantly, when you `dive()` a component, the **legacy** context of any providers further up in the tree will be passed as context (via the second argument to `.render()`).

So, if this change were to be released, and no changes in Enzyme were made, if you were to `dive()` a component that uses `contextType` you would end up with a bunch of **legacy** context as `this.context` in your component.

Now, of course, Enzyme could be updated to inspect the component being `dive()`d to determine whether new or legacy context should be passed. But I still think it's worth calling out that this change is a little more "break-y" then it appears at first. And I think this potentially odd behavior is a decent smell for the fact that new context and legacy context are actually totally different APIs, and we maybe shouldn't overload the second arg to `.render()` to handle both.

mprobber:
@minznerjosh 

The only thing coming to mind to that would avoid overloading the second argument, and supporting `ContextType` would be we passing a Map of `React.Context`, value pairs, but that seems very, very clunky.

I'd think that a follow-up change should be made to enzyme regardless if this PR gets merged.  Currently, if you shallow render a component with `contextType`, `this.context` is the empty object, as opposed to the default value of the context.  So, while this would make `dive()` behave bizarrely for components that use `contextType`, they already behaves incorrectly under enzyme.  Regardless, I'd be happy to craft a pull request for enzyme when any changes are made.

Let me know if you come up with an API that doesn't overload the second argument to render that's not super clunky.  I'd be happy to implement it.

gaearon:
>But Enzyme has the concept of dive()ing

Oh my. This is the first time I'm hearing and reading about this API 😳This seems like a real can of wormsn.

ljharb:
`.dive()` is just sugar for a feature enzyme has always had since v1, the ability to find an arbitrary custom component and create a new shallow wrapper from that point.

gaearon:
The confusing thing to me is that the updates in the parent component wouldn't "reflect" in that "dived" version for `shallow`. But they would for `mount`. Unless I'm missing something.

minznerjosh:
@gaearon I'm pretty sure  you're correct about this. If you change state/props in the parent, you'd need to re-dive to see the changes reflected in the "dived" version. This, however, probably isn't _super_ surprising to users given the immutable nature of non-root enzyme wrappers.

However, it does present itself if you ever want to test component lifecycle methods in the dived child by manipulating the props/state of the parent.

ljharb:
That's also the case for both shallow and mount for any derived wrapper - ie, any `.find`-produced wrapper is not updated if the parent updates, you always have to re-find from the root.

gaearon:
Thanks for explaining!

DavyJohnes:
Guys, is there any workaround to pass context to component which uses new context API?

d4wei:
@mprobber @gaearon Any chance we might move forward with this, or decide on an alternative API in the near future? Having shallow rendering support the new `contextType` API would be really useful!

If we are thinking about alternatives, would passing a map of Provider/value pairs when creating an instance of `shallowRenderer`(instead of passing it when `render` is called) be a less clunky way to do this while avoiding the problems @minznerjosh described? I think in a certain sense, non-default values for contexts can also be interpreted as "configurations" for a `shallowRenderer` instance.

> Guys, is there any workaround to pass context to component which uses new context API?

@DaveVoyles If the context your component uses **happens to be an object**, you can use the legacy API to test it for now by setting an appropriate value for `contextTypes` on your component at test time. I've documented this workaround in the [Stack Overflow question](https://stackoverflow.com/a/56483320/11610682) that led me here.

comp615:
Wanted to throw my hat into the ring too, having a solution here would greatly help unblocking our upgrade path to remove old-style, legacy context. Right now we code-modded most of it away and the code is functioning, but tests are an absolute mess. Can explore the workaround detailed above, but that feels a bit worse.

mjancarik:
I created module for workaround https://www.npmjs.com/package/shallow-with-context. The module works well in our projects.

ljharb:
bump

gaearon:
The shallow renderer has been extracted so please port this change to https://github.com/NMinhNguyen/react-shallow-renderer if you'd like. Thank you!

PR #14330, opened 2018-11-26
Remove usage of `fbjs/lib/invariant` in ReactNativeViewConfigRegistry.

cpojer:
I'm trying to remove fbjs from react-native. This is the corresponding change necessary in the React repo.

I am unsure how to test this so I'm assuming that CI will tell me if something is wrong here. Also please tell me if the `invariant` entry in `package.json` shall go somewhere else. I'm assuming since this is synced to react-native for product uses that `invariant` as a dev dependency on the root is good enough for testing.

### DIFF
--- a/scripts/rollup/shims/react-native/ReactNativeViewConfigRegistry.js
+++ b/scripts/rollup/shims/react-native/ReactNativeViewConfigRegistry.js
@@ -15,7 +15,7 @@ import type {
   ViewConfigGetter,
 } from './ReactNativeTypes';
 
-const invariant = require('fbjs/lib/invariant');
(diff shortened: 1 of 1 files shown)


cpojer:
So you are saying just referring to invariant without adding it to the `react` repo is fine?

sophiebits:
I think so. https://github.com/facebook/react/blob/409066a0a14434df387d25971f09d33b9a5aa914/scripts/rollup/packaging.js#L87

cpojer:
Done! Although to be fair this makes my substantial PR look fairly unimpressive. I'm happy to add some more complex looking stuff somewhere.

PR #14336, opened 2018-11-27
Remove useMutationEffect

sophiebits:
useMutationEffect has problems (namely, refs aren't attached at the time that it runs) and we're not positive it's necessary. useLayoutEffect runs at the same time as componentDidMount/Update so it's sufficient for all existing use cases; it can be used in any case that useEffect happens too late. Until we figure out what we want to do, let's delete it.


gaearon:
sorry @kentcdodds

kentcdodds:
Haha, I'll update my blogpost :D

krave1986:
Maybe we can also update docs [here](https://reactjs.org/docs/hooks-faq.html#how-to-read-an-often-changing-value-from-usecallback). Code snippet explaining how to read an often-changing value from useCallback used this hook which is no longer available.

sophiebits:
That exact link has already been updated.

sanonz:
After removing useMutationEffect, what should we do about this situation?

https://stackblitz.com/edit/react-use-styles?file=useStyles.js

bvaughn:
> After removing useMutationEffect, what should we do about this situation?
> https://codesandbox.io/s/znx8kj1j0l

Someone asked the above question (but I guess it was deleted before I had a chance to respond?)

Anyway, in case someone else comes across this PR and wonders the same– it seemed worth re-emphasizing what Sophie said in the PR description:
> useLayoutEffect runs at the same time as componentDidMount/Update so it's sufficient for all existing use cases; it can be used in any case that useEffect happens too late.

Updating the above repro to use `useLayoutEffect` instead will avoid the temporary flash of unstyled content. Here's an updated repro:
https://codesandbox.io/s/v6joo6k9o5

PR #14339, opened 2018-11-27
Release script auto-determine the latest Canary build if none specified

bvaughn:
Seems like this would be useful if we want to integrate with the FB sync script. Also it's just convenient.

Now you can prepare a specified build:
```sh
./scripts/release/prepare-canary.js --build=12774
```

Or just prepare the most recent one from `master`:
```
./scripts/release/prepare-canary.js
```

### Demo
kapture 2018-11-27 at 15 07 40

### DIFF
--- a/scripts/release/prepare-canary-commands/get-latest-master-build-number.js
+++ b/scripts/release/prepare-canary-commands/get-latest-master-build-number.js
@@ -1,0 +1,22 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const http = require('request-promise-json');
+const {logPromise} = require('../utils');
+
+const run = async () => {
+  // https://circleci.com/docs/api/#recent-builds-for-a-project-branch
+  const metadataURL = `https://circleci.com/api/v1.1/project/github/facebook/react/tree/master`;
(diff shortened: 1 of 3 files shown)


scripts/release/prepare-canary-commands/get-latest-master-build-number.js
```diff
@@ -0,0 +1,22 @@
...
+'use strict';
+
+const http = require('request-promise-json');
+const {logPromise} = require('../utils');
+
+const run = async () => {
+  // https://circleci.com/docs/api/#recent-builds-for-a-project-branch
+  const metadataURL = `https://circleci.com/api/v1.1/project/github/facebook/react/tree/master`;
+  const metadata = await http.get(metadataURL, true);
+  const build = metadata.find(entry => entry.branch === 'master').build_num;
```
sophiebits: Do you want the latest completed+successful build instead of just the latest? Failed and incomplete both seem subpar.

bvaughn: 🤔 Good point.

bvaughn:
🕺 If anyone disagrees about this change I'll revert it.

PR #14359, opened 2018-11-30
[Fire] Add initial build infrastructure

gaearon:
This adds:

* `react-dom/unstable-fire` entry point
* `react-dom/src/fire` folder which is where we'll put the implementation files
* A new host config in `react-dom/src/fire/ReactFireHostConfig`. Currently it just re-exports the old one, but you can verify it works by uncommenting the `console.log` in there.
* A new entry in `bundles.js`, `forks.js`, and a new inlining alias `react-reconciler/inline.fire` for our host config. This also means there's `yarn flow fire` now.
* A new entry point in `react-dom/src/fire/ReactFire`. It is a copy-paste of `ReactDOM` entry point with fixed relative paths. Its main purpose it just to show that the setup works — we'll fully replace the actual content later. Importantly, it references `react-reconciler/inline.fire`.
* Running `yarn test-fire` and `yarn test-fire-prod` now stubs `react-dom` with `react-dom/unstable-fire`. You can verify this by commenting out `console.log`s in the Fire host config and in the Fire entry point.
* There's a new fixture now. To get it working you'd need to run `yarn build react/index,fire --type=UMD_DEV`.

After this PR is merged we can rebase @trueadm's initial work on top.

Note: I decided to go with `ReactFire` rather than `ReactDOMFire` because we already know it's DOM and Dominic feels it's already getting too verbose for his taste. :P

### DIFF
--- a/fixtures/unstable-fire/index.html
+++ b/fixtures/unstable-fire/index.html
@@ -1,0 +1,14 @@
+<html>
+  <body>
+    <script src="../../build/node_modules/react/umd/react.development.js"></script>
+    <script src="../../build/node_modules/react-dom/umd/react-dom-unstable-fire.development.js"></script>
+    <script src="https://unpkg.com/babel-standalone@6/babel.js"></script>
+    <div id="container"></div>
+    <script type="text/babel">
+      ReactFire.render(
+        <h1>Hello World!</h1>,
+        document.getElementById('container')
+      );
+    </script>
+  </body>
+</html>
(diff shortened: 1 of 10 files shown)


trueadm:
Looks good – we also need a ReactFire specific ReactFlag if possible, like I've done in https://github.com/trueadm/react/tree/react-dom-fire so we can disable some legacy tests around thing we no longer support.

gaearon:
>we also need a ReactFire specific ReactFlag if possible

Let's do together with main integration PR.

gaearon:
Let's call next a minor. https://github.com/reactjs/reactjs.org/pull/1447

gaearon:
(Which probably means you should land yours too?)

PR #14382, opened 2018-12-03
[WIP] Add base implementation of React Fire

trueadm:
This an initial work-in-progress PR for React Fire. Better description to come at some point along with a TODO checklist. For more context, see the tracking issue #13525.


FredyC:
Why not link https://github.com/facebook/react/issues/13525 in here to have at least some basic info :)

philipp-spiess:
I got to take a look at this over the weekend. Unfortunately I didn’t have the chance to play with this yet but I’d like to summarize the changes that I’ve noted so far (I only really checked the event side of things):

1. There’s still a synthetic event system in place. Instead of all the synthetic event objects we monkey patch the native event object before we emulate the capture and bubble phases like we used to before.
2. Events are still listened at the document right now - The plan is to work on those changes after the base implementation is merged.
3. Event shims now work differently and there’s only one for `onChange` and one for `onBeforeInput` from what I can see:
  https://github.com/facebook/react/blob/235b682331b1afc4abfe6d25fc5c73eebc586393/packages/react-dom/src/fire/polyfills/ReactFirePolyfilledEvents.js#L4-L9

Thank you @trueadm for answering my questions 🙂.

trueadm:
@philipp-spiess Sure, this PR is very early and things are highly likely to change.
@FredyC Good point.

packages/react-dom/src/events/__tests__/BeforeInputEventPlugin-test.js
```diff
@@ -810,9 +812,12 @@ describe('BeforeInputEventPlugin', () => {
     testContentEditableComponent(environments[1], scenarios);
   });
 
-  it('should extract onBeforeInput when simulating in env with no CompositionEvent on input[type=text]', () => {
-    testInputComponent(environments[2], scenarios);
-  });
+  // TODO investigate why this test fails for Fire
```
j-nolan: This test seems to pass. ChangeEventPlugin-test.internal.js is red though

trueadm: It's failing for me. Note, the way to run React Fire tests on this PR is: `yarn test-fire` or `yarn test-fire-prod`.

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -811,7 +810,8 @@ function createRoot(container: DOMContainer, options?: RootOptions): ReactRoot {
 
 if (enableStableConcurrentModeAPIs) {
   ReactDOM.createRoot = createRoot;
-  ReactDOM.unstable_createRoot = undefined;
+} else {
```
Andarist: wont this deopt ReactDOM object once again?

trueadm: Was mistakenly pulled from an old branch.

packages/react-dom/src/fire/ReactFireEventTraversal.js
```diff
@@ -0,0 +1,234 @@
...
+}
+
+/**
+ * Traverses the ID hierarchy and invokes the supplied `cb` on any IDs that
+ * should would receive a `mouseEnter` or `mouseLeave` event.
+ *
+ * Does not invoke the callback on the nearest common ancestor because nothing
+ * "entered" or "left" that element.
+ */
+export function traverseEnterLeave(
```
jquense: would this be a good moment/save implementation complexity to change up how this plugin is implemented, could avoid the special traversal

packages/react-dom/src/fire/ReactFireEventTraversal.js
```diff
@@ -0,0 +1,234 @@
...
+  );
+}
+
+function dispatchEventHandler(fiber, capturePhase, proxyContext) {
+  const domNode = getDOMNodeFromFiber(fiber);
+  // Html Nodes can be nested fe: span inside button in that scenario
+  // browser does not handle disabled attribute on parent,
+  // because the event listener is on document.body
+  // Don't process clicks on disabled elements
+  if (proxyContext.isClickEvent && domNode.disabled) {
```
jquense: isn't this true for any mouse/pointer event?

packages/react-dom/src/fire/ReactFireEvents.js
```diff
@@ -0,0 +1,322 @@
...
+  );
+
+  monkeyPatchNativeEvent(event, proxyContext);
+  batchedUpdates(dispatchEvent, proxyContext);
+}
+
+function stopPropagation() {
+  this.cancelBubble = true;
+  if (!this.immediatePropagationStopped) {
+    this.stopImmediatePropagation();
```
jquense: why does stopPropagation call stopImmediatePropagation (also IE support?)

jquense:
this is sweet !

packages/react-dom/src/fire/ReactFireEventTraversal.js
trueadm: This was taken from Inferno and I believe we have something similar in the current ReactDOM inside SimpleEventsPlugin.

packages/react-dom/src/fire/ReactFireEvents.js
trueadm: Was also taken from Inferno, need to check if we still need to do this, but last time I checked it was to get around some issues in IE.

packages/react-dom/src/fire/ReactFireEventTraversal.js
trueadm: Do you have any ideas how we can change this plugin? For now the idea was to leave the current logic mostly alone and try and refactor with the new codebase operational. It would be good if you have some ideas how to avoid doing all this extra work.

packages/react-dom/src/fire/ReactFireEvents.js
```diff
@@ -0,0 +1,351 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import {
+  addEventBubbleListener,
+  addEventCaptureListener,
+} from '../events/EventListener';
```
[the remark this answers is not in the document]
trueadm: This is true, but for phase 1 of React Fire, the idea is to keep attaching to the document so this will stay until we remove it in phase 2. :)

packages/react-dom/src/fire/ReactFireDevTools.js
```diff
@@ -0,0 +1,52 @@
...
+              (protocol === 'file:'
+                ? '\nYou might need to use a local HTTP server (instead of file://): ' +
+                  'https://fb.me/react-devtools-faq'
+                : ''),
+            'font-weight:bold',
+          );
+        }
+      }
+    }
+  }
```
Goodluckhf: I think, for better readability is better to reverse conditions. It'll lead to decrease level of nesting
something like this:
```
if (! __DEV__) return;
```

packages/react-dom/src/fire/ReactFireComponentProperties.js
```diff
@@ -0,0 +1,905 @@
...
+        // Unset styles on `lastProp` but not on `nextProp`.
+        for (styleName in lastProp) {
+          if (
+            lastProp.hasOwnProperty(styleName) &&
+            (!nextProp || !nextProp.hasOwnProperty(styleName))
+          ) {
+            if (!styleUpdates) {
+              styleUpdates = {};
+            }
+            styleUpdates[styleName] = '';
```
Goodluckhf: it looks like repeating code from line 306. Is it possible to simplify logic here, maybe to extract method?

packages/react-dom/src/fire/ReactFireComponentProperties.js
```diff
@@ -0,0 +1,905 @@
...
+      possibleStandardNames &&
+      !possibleStandardNames.hasOwnProperty(lowerCasedName)
+    ) {
+      return null;
+    }
+    return (
+      (possibleStandardNames && possibleStandardNames[lowerCasedName]) || null
+    );
+  }
+  return null;
```
Goodluckhf: IMHO here it's better to reverse condition too
```if (! __DEV__) return;```

File: packages/react-dom/src/fire/ReactFireComponentProperties.js [hunk 10]
```diff
@@ -0,0 +1,905 @@
...
+  node: Element,
+  name: string,
+  expected: mixed,
+  propertyInfo: PropertyInfo,
+): mixed {
+  if (__DEV__) {
+    if (propertyInfo.mustUseProperty) {
+      const {propertyName} = propertyInfo;
+      return (node: any)[propertyName];
+    } else {
```
Goodluckhf: `else` here is redundant. It's only add unnecessary nesting

packages/react-dom/src/fire/ReactFireDevTools.js
trueadm: It's nested for a reason – it's so the DEV code can get dead-code-eliminated by the React bundle build system for production environments.

trueadm: It’s not about the reverse condition. It’s about a different coding pattern - in this case an early return guard.

trueadm:
This was only an experiment, closing.

PR #14383, opened 2018-12-03
Prevent a v8 deopt when profiling

### LINKED ISSUE #14365: Possible v8 de-opt for profiling react-dom bundles
I've observed that when React's profiling mode is enabled, updating the profiling fields [here](https://github.com/facebook/react/blob/1d25aa5787d4e19704c049c3cfa985d3b5190e0d/packages/react-reconciler/src/ReactProfilerTimer.js#L46) and [here](https://github.com/facebook/react/blob/1d25aa5787d4e19704c049c3cfa985d3b5190e0d/packages/react-reconciler/src/ReactProfilerTimer.js#L69-L71) during the _render_ phase seems to cause a significant deopt in [`getHostSibling`](https://github.com/facebook/react/blob/1d25aa5787d4e19704c049c3cfa985d3b5190e0d/packages/react-reconciler/src/ReactFiberCommitWork.js#L858-L898) during the _commit_ phase for Chrome only. (Neither Safari nor Firefox seem to be affected by this.)

I have created a repro case with inline annotations here:
https://github.com/bvaughn/react-profiling-v8-deopt-repro#about-this-repro-case

(Note that the above repo is private. Please tag me here if you need access.)

bvaughn:
Resolves #14365

### DIFF
--- a/packages/react-reconciler/src/ReactFiber.js
+++ b/packages/react-reconciler/src/ReactFiber.js
@@ -254,6 +254,23 @@ function FiberNode(
   this.alternate = null;
 
   if (enableProfilerTimer) {
+    // Note: The following is done to avoid a v8 deopt.
+    //
+    // It is important to initialize the fields below with doubles.
+    // Otherwise Fibers will deopt and end up having separate shapes when
+    // doubles are later assigned to fields that initially contained smis.
+    // This is a bug in v8 having something to do with Object.preventExtension().
+    //
+    // Learn more about this deopt here:
+    // https://github.com/facebook/react/issues/14365
+    // https://bugs.chromium.org/p/v8/issues/detail?id=8538
+    this.actualDuration = Number.NaN;
+    this.actualStartTime = Number.NaN;
+    this.selfBaseDuration = Number.NaN;
+    this.treeBaseDuration = Number.NaN;
+
+    // It's okay to replace the initial doubles with smis after initialization.
+    // This simplifies other profiler code and doesn't trigger the deopt.
     this.actualDuration = 0;
     this.actualStartTime = -1;
     this.selfBaseDuration = 0;


trueadm:
Nice work finding that we can set them back to ints straight after. Seems like a simple fix to me. I was going to suggest just removing the preventExtensions logic, but this is better.

bvaughn:
Thanks Dom! (and thanks for the help digging into this on Friday)

gaearon:
🤣 

thanks for digging into this, you're both awesome

packages/react-reconciler/src/ReactFiber.js
```diff
@@ -254,6 +254,23 @@ function FiberNode(
   this.alternate = null;
 
   if (enableProfilerTimer) {
+    // Note: The following is done to avoid a v8 deopt.
```
mathiasbynens: Nit: `s/deopt/performance cliff/`

(It's not a deopt.)

bvaughn: Interesting. I was referring to the fact that the result is Fibers all having different shapes (which seemed like a deopt to me) but admittedly, I was just guessing at terminology 😄

mathiasbynens: To give some more background, a deopt happens when we're running optimized machine code generated by the optimizing compiler (TurboFan in V8's case), and one of the assumptions in the optimized code turns out to be invalid. In such cases we have to _deopt_ and go back to running bytecode in the interpreter (Ignition in V8's case). ([Our JSConf talk includes some more detail and visualizations that make this more clear, in case you're interested.](https://mathiasbynens.be/notes/shapes-ics))

This V8 bug however occurs before we even get to optimized code.

You can verify this yourself by running [our reduced test case](https://bugs.chromium.org/p/v8/issues/detail?id=8538) with `--no-opt`: it still reproduces.

PR #14389, opened 2018-12-04
Updated release script's suggested next version

bvaughn:
Based on team discussion during today's sync, the next version number will be:
* A minor bump if releasing from master
* A patch bump if releasing from a branch

The release script still enables user overrides for these values. This only affects the suggested next version.

### DIFF
--- a/scripts/release/prepare-canary-commands/download-build-artifacts.js
+++ b/scripts/release/prepare-canary-commands/download-build-artifacts.js
@@ -4,7 +4,7 @@
 
 const http = require('request-promise-json');
 const {exec} = require('child-process-promise');
-const {readdirSync} = require('fs');
+const {existsSync, readdirSync} = require('fs');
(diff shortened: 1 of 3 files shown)


scripts/release/prepare-canary-commands/download-build-artifacts.js
```diff
@@ -20,6 +20,10 @@ const run = async ({build, cwd}) => {
     entry => entry.path === 'home/circleci/project/node_modules.tgz'
   ).url;
 
+  if (!existsSync(join(cwd, 'build'))) {
+    await exec(`mkdir ./build`, {cwd});
+  }
```
bvaughn: This change (and the one in `check-out-packages`) was just fixing an edge case I noticed while testing these changes.

bvaughn:
This seems uncontroversial enough that I'm going to just land it.

PR #14392, opened 2018-12-04
TestRenderer toJSON should not expose the Suspense children Array wrapper

bvaughn:


packages/react-test-renderer/src/ReactTestRenderer.js
```diff
@@ -461,7 +461,15 @@ const ReactTestRendererFiber = {
         if (container.children.length === 1) {
           return toJSON(container.children[0]);
         }
-
+        if (
+          container.children.length === 2 &&
+          container.children[0].isHidden === true &&
```
sophiebits: Can we throw if we encounter .isHidden in any other context?

bvaughn: What do you mean by "in any other context"?

sophiebits: If I understand right, today we only use .isHidden for the two children of a Suspense component. If we ever change that we may want to update this logic in the test renderer code. So I thought it might be smart to assert that !isHidden except in this specific case. (For example, if arbitrary user code could make host components hidden today, then this logic would incorrectly catch those if there's ever two children with the first child hidden.)

bvaughn: > For example, if arbitrary user code could make host components hidden today

Ah. This `isHidden` attribute is set by the renderer's config though. It's not something users can directly set. So if we ever had an invalid `isHidden` setting, that would mean a bug in React (or at least the test renderer).

sophiebits: I'm trying to catch the case where we change React and forget about the test renderer.

bvaughn: I guess that makes sense...but it feels like that would be a kind of heavyweight way to guard against that.

I think we would need to update this function, as well as the `toJSON` recursive method it calls, to check every child for `isHidden === true` and if it ever finds them– verify that there is only one sibling, and that it's `isHidden === false`.

PR #14400, opened 2018-12-06
Dangerfile exits early if build failed

bvaughn:


sophiebits:
Does this mean if danger succeeds then you push an update and break the build, the old comment sticks around?

bvaughn:
🤔 I'm not sure. I can test it out on this PR by pushing a bad commit.

Maybe I need to explicitly print an error message using `danger.markdown` ..

bvaughn:
Given the current changes, if a build fails in the scenario you describe, @sophiebits, this comment will replace the previous one:
screen shot 2018-12-06 at 2 17 16 pm

If I just `process.exit(0)` without logging anything, I think the danger comment would just be removed (or not added to begin with). Which is preferable?

sophiebits:
I think showing the message is preferable. It's harder to get confused that way.

bvaughn:
Seems redundant, since there will also be an update from CI in that case:
screen shot 2018-12-06 at 2 22 12 pm

In either case, the stale/invalid results stats would be remove.

bvaughn:
I think it's better for the danger comment to be removed if the build fails. I think the red CI ❌ is enough feedback for that case.

Unless anyone objects, I think the behavior this PR currently exhibits is okay.

sophiebits:
ok, cool!

bvaughn:
No problem 😄

PR #14420, opened 2018-12-11
Add failing test of getDerivedStateFromError+retry

sophiebits:
Wanted to try to test out our error handling behavior. I don't fully understand the behavior here.


sophiebits:
Not urgent, assigning to Andrew for now to dig in later.

yongningfu:
@sophiebits although we use getDerivedStateFromError to fix the error, but we did not reset the global nextRenderDidError to false, so it will continue re-render with sync in the end of renderRoot

[when error, we set nextRenderDidError to true](https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberUnwindWork.js#L355)

[in the renderRoot end, we will retry](https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberScheduler.js#L1425)

PR #14421, opened 2018-12-11
Merged enableProfilerTimer & enableSchedulerTracing flags into enableProfiling

bvaughn:
In practice, we always turn these flags on and off together.

We can also use this new flag to determine when to track suspending components and durations (rather than adding a third flag and complication `ReactFiberRoot`.

Most of this change is pretty rote. The only file worth looking at more closely is `ReactProfiler-test.internal` where I simplified tests for a case that no longer exists (profiling enabled, tracing disabled). The relevant code related to this change can be seen in `ReactFiberCommitWork`.

[See this comment below for a better view of the actual changes to `ReactProfiler-test.internal`](https://github.com/facebook/react/pull/14421#issuecomment-446332734).


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -543,29 +539,18 @@ function commitLifeCycles(
...
-        } else {
-          onRender(
-            finishedWork.memoizedProps.id,
-            current === null ? 'mount' : 'update',
-            finishedWork.actualDuration,
-            finishedWork.treeBaseDuration,
-            finishedWork.actualStartTime,
-            getCommitTime(),
-          );
-        }
```
bvaughn: We never actually used this combination anywhere (in www, fbsource, or open source).

bvaughn:
Here's the whitespace-ignored diff for `packages/react/src/__tests__/ReactProfiler-test.internal.js`
```diff
diff --git a/packages/react/src/__tests__/ReactProfiler-test.internal.js b/packages/react/src/__tests__/ReactProfiler-test.internal.js
index a811cec41..b9e2709ee 100644
--- a/packages/react/src/__tests__/ReactProfiler-test.internal.js
+++ b/packages/react/src/__tests__/ReactProfiler-test.internal.js
@@ -25,8 +25,7 @@ let TextResource;
 let resourcePromise;
 
 function loadModules({
-  enableProfilerTimer = true,
-  enableSchedulerTracing = true,
+  enableProfiling = true,
   replayFailedUnitOfWorkWithInvokeGuardedCallback = false,
   useNoopRenderer = false,
 } = {}) {
@@ -39,8 +38,7 @@ function loadModules({
   ReactFeatureFlags = require('shared/ReactFeatureFlags');
   ReactFeatureFlags.debugRenderPhaseSideEffects = false;
   ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
-  ReactFeatureFlags.enableProfilerTimer = enableProfilerTimer;
-  ReactFeatureFlags.enableSchedulerTracing = enableSchedulerTracing;
+  ReactFeatureFlags.enableProfiling = enableProfiling;
   ReactFeatureFlags.replayFailedUnitOfWorkWithInvokeGuardedCallback = replayFailedUnitOfWorkWithInvokeGuardedCallback;
 
   React = require('react');
@@ -127,22 +125,19 @@ const mockDevToolsForTest = () => {
 
 describe('Profiler', () => {
   describe('works in profiling and non-profiling bundles', () => {
-    [true, false].forEach(enableSchedulerTracing => {
-      [true, false].forEach(enableProfilerTimer => {
-        describe(`enableSchedulerTracing:${
-          enableSchedulerTracing ? 'enabled' : 'disabled'
-        } enableProfilerTimer:${
-          enableProfilerTimer ? 'enabled' : 'disabled'
+    [true, false].forEach(enableProfiling => {
+      describe(`enableProfiling: ${
+        enableProfiling ? 'enabled' : 'disabled'
       }`, () => {
         beforeEach(() => {
           jest.resetModules();
 
-            loadModules({enableSchedulerTracing, enableProfilerTimer});
+          loadModules({enableProfiling});
         });
 
         // This will throw in production too,
         // But the test is only interested in verifying the DEV error message.
-          if (__DEV__ && enableProfilerTimer) {
+        if (__DEV__ && enableProfiling) {
           it('should warn if required params are missing', () => {
             expect(() => {
               expect(() => {
@@ -208,14 +203,12 @@ describe('Profiler', () => {
       });
     });
   });
-  });
 
-  [true, false].forEach(enableSchedulerTracing => {
   describe('onRender callback', () => {
     beforeEach(() => {
       jest.resetModules();
 
-        loadModules({enableSchedulerTracing});
+      loadModules({enableProfiling: true});
     });
 
     it('should handle errors thrown', () => {
@@ -310,14 +303,14 @@ describe('Profiler', () => {
 
       let [call] = callback.mock.calls;
 
-        expect(call).toHaveLength(enableSchedulerTracing ? 7 : 6);
+      expect(call).toHaveLength(7);
       expect(call[0]).toBe('test');
       expect(call[1]).toBe('mount');
       expect(call[2]).toBe(10); // actual time
       expect(call[3]).toBe(10); // base time
       expect(call[4]).toBe(5); // start time
       expect(call[5]).toBe(15); // commit time
-        expect(call[6]).toEqual(enableSchedulerTracing ? new Set() : undefined); // interaction events
+      expect(call[6]).toEqual(new Set()); // interaction events
 
       callback.mockReset();
 
@@ -333,14 +326,14 @@ describe('Profiler', () => {
 
       [call] = callback.mock.calls;
 
-        expect(call).toHaveLength(enableSchedulerTracing ? 7 : 6);
+      expect(call).toHaveLength(7);
       expect(call[0]).toBe('test');
       expect(call[1]).toBe('update');
       expect(call[2]).toBe(10); // actual time
       expect(call[3]).toBe(10); // base time
       expect(call[4]).toBe(35); // start time
       expect(call[5]).toBe(45); // commit time
-        expect(call[6]).toEqual(enableSchedulerTracing ? new Set() : undefined); // interaction events
+      expect(call[6]).toEqual(new Set()); // interaction events
 
       callback.mockReset();
 
@@ -356,14 +349,14 @@ describe('Profiler', () => {
 
       [call] = callback.mock.calls;
 
-        expect(call).toHaveLength(enableSchedulerTracing ? 7 : 6);
+      expect(call).toHaveLength(7);
       expect(call[0]).toBe('test');
       expect(call[1]).toBe('update');
       expect(call[2]).toBe(4); // actual time
       expect(call[3]).toBe(4); // base time
       expect(call[4]).toBe(65); // start time
       expect(call[5]).toBe(69); // commit time
-        expect(call[6]).toEqual(enableSchedulerTracing ? new Set() : undefined); // interaction events
+      expect(call[6]).toEqual(new Set()); // interaction events
     });
 
     it('includes render times of nested Profilers in their parent times', () => {
@@ -934,8 +927,7 @@ describe('Profiler', () => {
         expect(call[5]).toBe(380); // commit time
       });
 
-        [true, false].forEach(
-          replayFailedUnitOfWorkWithInvokeGuardedCallback => {
+      [true, false].forEach(replayFailedUnitOfWorkWithInvokeGuardedCallback => {
         describe(`replayFailedUnitOfWorkWithInvokeGuardedCallback ${
           replayFailedUnitOfWorkWithInvokeGuardedCallback
             ? 'enabled'
@@ -1124,8 +1116,7 @@ describe('Profiler', () => {
             ReactNoop.flush();
           });
         });
-          },
-        );
+      });
     });
 
     it('reflects the most recently rendered id value', () => {
@@ -1197,7 +1188,6 @@ describe('Profiler', () => {
       expect(callback).toHaveBeenCalledTimes(1);
     });
   });
-  });
 
   it('should handle interleaved async yields and batched commits', () => {
     jest.resetModules();
@@ -1278,7 +1268,7 @@ describe('Profiler', () => {
       jest.resetModules();
 
       loadModules({
-        enableSchedulerTracing: true,
+        enableProfiling: true,
         ...params,
       });
 
@@ -1555,7 +1545,7 @@ describe('Profiler', () => {
       let call = onRender.mock.calls[0];
       expect(call[0]).toEqual('test-profiler');
       expect(call[5]).toEqual(mockNow());
-      if (ReactFeatureFlags.enableSchedulerTracing) {
+      if (ReactFeatureFlags.enableProfiling) {
         expect(call[6]).toMatchInteractions([interactionCreation]);
       }
 
@@ -1627,7 +1617,7 @@ describe('Profiler', () => {
         call = onRender.mock.calls[0];
         expect(call[0]).toEqual('test-profiler');
         expect(call[5]).toEqual(mockNow());
-        if (ReactFeatureFlags.enableSchedulerTracing) {
+        if (ReactFeatureFlags.enableProfiling) {
           expect(call[6]).toMatchInteractions([interactionOne]);
         }
 
@@ -1660,7 +1650,7 @@ describe('Profiler', () => {
       call = onRender.mock.calls[0];
       expect(call[0]).toEqual('test-profiler');
       expect(call[5]).toEqual(mockNow());
-      if (ReactFeatureFlags.enableSchedulerTracing) {
+      if (ReactFeatureFlags.enableProfiling) {
         expect(call[6]).toMatchInteractions([]);
       }
 
@@ -1716,7 +1706,7 @@ describe('Profiler', () => {
       call = onRender.mock.calls[0];
       expect(call[0]).toEqual('test-profiler');
       expect(call[5]).toEqual(mockNow());
-      if (ReactFeatureFlags.enableSchedulerTracing) {
+      if (ReactFeatureFlags.enableProfiling) {
         expect(call[6]).toMatchInteractions([interactionTwo]);
       }
 
@@ -1850,7 +1840,7 @@ describe('Profiler', () => {
         expect(call[0]).toEqual('test');
         expect(call[5]).toEqual(mockNow());
         expect(call[6]).toMatchInteractions(
-          ReactFeatureFlags.enableSchedulerTracing
+          ReactFeatureFlags.enableProfiling
             ? [interactionLowPri, interactionHighPri]
             : [],
         );
@@ -1867,7 +1857,7 @@ describe('Profiler', () => {
         expect(call[0]).toEqual('test');
         expect(call[5]).toEqual(mockNow());
         expect(call[6]).toMatchInteractions(
-          ReactFeatureFlags.enableSchedulerTracing ? [interactionLowPri] : [],
+          ReactFeatureFlags.enableProfiling ? [interactionLowPri] : [],
         );
 
         expect(onInteractionTraced).toHaveBeenCalledTimes(2);
@@ -1977,13 +1967,13 @@ describe('Profiler', () => {
       expect(call[0]).toEqual('test');
       expect(call[5]).toEqual(firstCommitTime);
       expect(call[6]).toMatchInteractions(
-        ReactFeatureFlags.enableSchedulerTracing ? [interactionOne] : [],
+        ReactFeatureFlags.enableProfiling ? [interactionOne] : [],
       );
       call = onRender.mock.calls[1];
       expect(call[0]).toEqual('test');
       expect(call[5]).toEqual(mockNow());
       expect(call[6]).toMatchInteractions(
-        ReactFeatureFlags.enableSchedulerTracing ? [interactionOne] : [],
+        ReactFeatureFlags.enableProfiling ? [interactionOne] : [],
       );
 
       onRender.mockClear();
@@ -2040,13 +2030,13 @@ describe('Profiler', () => {
       expect(call[0]).toEqual('test');
       expect(call[5]).toEqual(firstCommitTime);
       expect(call[6]).toMatchInteractions(
-        ReactFeatureFlags.enableSchedulerTracing ? [interactionTwo] : [],
+        ReactFeatureFlags.enableProfiling ? [interactionTwo] : [],
       );
       call = onRender.mock.calls[1];
       expect(call[0]).toEqual('test');
       expect(call[5]).toEqual(mockNow());
       expect(call[6]).toMatchInteractions(
-        ReactFeatureFlags.enableSchedulerTracing ? [interactionTwo] : [],
+        ReactFeatureFlags.enableProfiling ? [interactionTwo] : [],
       );
 
       onRender.mockClear();
@@ -2105,13 +2095,13 @@ describe('Profiler', () => {
       expect(call[0]).toEqual('test');
       expect(call[5]).toEqual(firstCommitTime);
       expect(call[6]).toMatchInteractions(
-        ReactFeatureFlags.enableSchedulerTracing ? [interactionThree] : [],
+        ReactFeatureFlags.enableProfiling ? [interactionThree] : [],
       );
       call = onRender.mock.calls[1];
       expect(call[0]).toEqual('test');
       expect(call[5]).toEqual(mockNow());
       expect(call[6]).toMatchInteractions(
-        ReactFeatureFlags.enableSchedulerTracing ? [interactionThree] : [],
+        ReactFeatureFlags.enableProfiling ? [interactionThree] : [],
       );
     });
 
@@ -2172,7 +2162,7 @@ describe('Profiler', () => {
       let call = onRender.mock.calls[0];
       expect(call[0]).toEqual('test-profiler');
       expect(call[6]).toMatchInteractions(
-        ReactFeatureFlags.enableSchedulerTracing ? [interaction] : [],
+        ReactFeatureFlags.enableProfiling ? [interaction] : [],
       );
 
       expect(onInteractionTraced).toHaveBeenCalledTimes(1);
@@ -2260,7 +2250,7 @@ describe('Profiler', () => {
         let call = onRender.mock.calls[0];
         expect(call[0]).toEqual('test-profiler');
         expect(call[6]).toMatchInteractions(
-          ReactFeatureFlags.enableSchedulerTracing ? [interaction] : [],
+          ReactFeatureFlags.enableProfiling ? [interaction] : [],
         );
 
         expect(onInteractionTraced).toHaveBeenCalledTimes(1);
@@ -2283,7 +2273,7 @@ describe('Profiler', () => {
         call = onRender.mock.calls[2];
         expect(call[0]).toEqual('test-profiler');
         expect(call[6]).toMatchInteractions(
-          ReactFeatureFlags.enableSchedulerTracing ? [interaction] : [],
+          ReactFeatureFlags.enableProfiling ? [interaction] : [],
         );
 
         expect(onInteractionTraced).toHaveBeenCalledTimes(1);
```

sophiebits:
(I don't feel qualified to review this.)

bvaughn:
I don't think there's much context required. This PR just collapses two feature flags (that were always flipped on and off together) into a single one.

If there's any context I could provide by chatting that would give you more confidence in reviewing it, I'd be happy to for what it's worth. 😄

sophiebits:
Right, I don't feel qualified to decide if that is a good change.

bvaughn:
I see. I am confident that it is a good change, but I'll find someone else to review the PR.

bvaughn:
@sebmarkbage mentioned that it might be a good idea to remove the tracing flag in the future and just always enable that feature (even for prod). This change would still be okay, since we'd just be moving code outside of a feature flag (either tracing or the new profiling one).

sebmarkbage:
The issue is that it makes it a lot harder to figure out in the future which pieces should move out of the block.

One thing you could do is ensure that enableProfiling is never turned on without enableSchedulerTracing also being turned on. That way you don't have to handle the case where Profiling is on but Tracing isn't since that will never happen.

bvaughn:
> The issue is that it makes it a lot harder to figure out in the future which pieces should move out of the block.

I don't think that's true. We'll need to do a similar amount of work in either case if we choose to remove the `enableSchedulerTracing` feature flag. I think it's not a big amount of work though, because the flag is only used in a few files.

> One thing you could do is ensure that enableProfiling is never turned on without enableSchedulerTracing also being turned on. That way you don't have to handle the case where Profiling is on but Tracing isn't since that will never happen.

Aside from the fact that separate flags seemed to have outlasted their usefulness in this case– the original motivation for this merger was to avoid adding a _third_ flag for suspense metrics. Particularly because it would make places like [`ReactFiberRoot`](https://github.com/facebook/react/blob/2743fb7b23a5af42059a9b07c5f6048f941709d9/packages/react-reconciler/src/ReactFiberRoot.js#L112-L166) more complicated.

bvaughn:
Doesn't seem like anyone but me is in favor of this change 😄 so maybe we can consider #14444 as an alternative.

PR #14423, opened 2018-12-12
Add support for Suspense & lazy() to the react-is package

pleunv:
The "react-is" package appears to lack checks for `React.Suspense` & `React.lazy()`. Since libraries like enzyme depend on this to implement proper support for these element types I figured I'd take a stab at adding them.

I ran into some `yarn test` related issues on the way, as I seem to be unable to run the whole test suite (on Windows), only a `yarn test --watch ReactIs` does the trick. Should I file an issue for this?

### DIFF
--- a/packages/react-is/src/__tests__/ReactIs-test.js
+++ b/packages/react-is/src/__tests__/ReactIs-test.js
@@ -39,17 +39,17 @@ describe('ReactIs', () => {
     }
 
     const FunctionComponent = () => React.createElement('div');
-
     const ForwardRefComponent = React.forwardRef((props, ref) =>
       React.createElement(Component, {forwardedRef: ref, ...props}),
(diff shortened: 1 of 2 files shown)


packages/react-is/src/__tests__/ReactIs-test.js
```diff
@@ -152,6 +156,14 @@ describe('ReactIs', () => {
     expect(ReactIs.isMemo(Component)).toBe(false);
   });
 
+  it('should identify lazy', () => {
+    const Component = () => React.createElement('div');
+    const lazyComponent = React.lazy(() => Component);
```
bvaughn: nit: This should actually be `React.lazy(() => ({default: Component});` but that doesn't matter for the purposes of this test.

PR #14427, opened 2018-12-12
Inject overrideProps() fn to DevTools

bvaughn:
This function will enable editing props for function components, host nodes, and special types like `memo` and `forwardRef`.

Copying the update logic into React will also enable us to add more complex behavior in the future– such as the ability to override props in a way that won't get overridden unless the component is re-rendered with a _new_ props value.

## Demo
Here's a demo that shows me editing a `Context.Provider` value, `React.memo` and `React.forwardRef` props, host node props, and function component props.
kapture 2018-12-12 at 15 40 18

Unblocks facebook/react-devtools/pull/1249

### DIFF
--- a/packages/react-reconciler/src/ReactFiberReconciler.js
+++ b/packages/react-reconciler/src/ReactFiberReconciler.js
@@ -62,6 +62,7 @@ import {
   current as ReactCurrentFiberCurrent,
 } from './ReactCurrentFiber';
 import {StrictMode} from './ReactTypeOfMode';
+import {Sync} from './ReactFiberExpirationTime';
 
 type OpaqueRoot = FiberRoot;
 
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -339,10 +340,49 @@ export function findHostInstanceWithNoPortals(
...
+    return copyWithSetImpl(obj, path, 0, value);
+  };
+
+  // Support DevTools props for function components, forwardRef, memo, host components, etc.
+  overrideProps = (fiber: Fiber, path: Array<string | number>, value: any) => {
+    fiber.pendingProps = copyWithSet(fiber.memoizedProps, path, value);
+    if (fiber.alternate) {
+      fiber.alternate.pendingProps = fiber.pendingProps;
+    }
+    flushPassiveEffects();
```
sebmarkbage: Why is this after setting the pendingProps? What if these effects synchronously renderers an update to this tree?

bvaughn: Somewhat arbitrary (so I'll move it).

Seems a little weird either way, since such a synchronous re-renderer would mean we're reading from stale `memoizedProps` when we do our copuy.

PR #14429, opened 2018-12-13
Memoize promise listeners to prevent exponential growth

### LINKED ISSUE #14220: When using React.lazy will cause the GPU/CPU to run overloaded, and the page is very slow.
bug

page is very slow.

As we can see from this picture, React has been executing a work loop during the rendering suspend.

image

image

I expect the page not to be stuck

React16.6.3

acdlite:
Previously, React would attach a new listener every time a promise is thrown, regardless of whether the same listener was already attached during a previous render. Because React attempts to render every time a promise resolves, the number of listeners grows quickly.

This was especially bad in synchronous mode because the renders that happen when the promise pings are not batched together. So if a single promise has multiple listeners for the same root, there will be multiple renders, which in turn results in more listeners being added to the remaining unresolved promises. This results in exponential growth in the number of listeners with respect to the number of IO-bound components in a single render.

Fixes #14220

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -20,7 +20,9 @@ import type {ExpirationTime} from './ReactFiberExpirationTime';
 import type {CapturedValue, CapturedError} from './ReactCapturedValue';
 import type {SuspenseState} from './ReactFiberSuspenseComponent';
 import type {FunctionComponentUpdateQueue} from './ReactFiberHooks';
+import type {Thenable} from './ReactFiberScheduler';
 
+import {unstable_wrap as Schedule_tracing_wrap} from 'scheduler/tracing';
 import {
   enableHooks,
   enableSchedulerTracing,
@@ -88,6 +90,7 @@ import {
 import {
   captureCommitPhaseError,
   requestCurrentTime,
+  retryTimedOutBoundary,
 } from './ReactFiberScheduler';
 import {
   NoEffect as NoHookEffect,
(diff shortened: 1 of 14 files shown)


packages/react-reconciler/src/__tests__/ReactSuspensePlaceholder-test.internal.js
```diff
@@ -8,9 +8,9 @@
  * @jest-environment node
  */
 
-runPlaceholderTests('ReactSuspensePlaceholder (mutation)', () =>
-  require('react-noop-renderer'),
-);
+// runPlaceholderTests('ReactSuspensePlaceholder (mutation)', () =>
+//   require('react-noop-renderer'),
+// );
```
sebmarkbage: Why disable these?

acdlite: Commented it out while debugging, meant to add back. Although now I just noticed... I screwed up when I converted this to use the test renderer. It's not actually testing anything in concurrent mode anymore :facepalm:

acdlite: I'll fix that in a separate PR since it's unrelated

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1180,6 +1185,30 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
       if (primaryChildParent !== null) {
         hideOrUnhideAllChildren(primaryChildParent, newDidTimeout);
       }
+
+      // If this boundary just timed out, then it will have a set of thenables.
+      // For each thenable, attach a listener so that when it resolves, React
+      // attempts to re-render the boundary in the primary (pre-timeout) state.
```
sebmarkbage: I thought we wanted to do this work in the passive effect? Is there a reason not to?

sebmarkbage:
You still need to add the lint config to a few more places like RN.

packages/react-reconciler/src/ReactFiberCommitWork.js
acdlite: Only that it seemed like a riskier change. I'll add a follow-up.

Kovensky:
"Because React attempts to render every time a promise resolves" ah, so it's intentional. I thought it was a bug when I encountered this in test code (I thought React would wait for all thrown Promises to resolve before attempting to rerender a Suspense. Won't this just cause another suspend again?)

acdlite:
@Kovensky Every time a promise pings it means some part of the tree has been unblocked, so it’s worth rendering again, even before all the promises resolve, to fire the nested requests.

threepointone:
^ Just dropping a note here that the above issues were because of a bad polyfill, and have been 'resolved'. This PR is fine.

PR #14435, opened 2018-12-13
Enable hooks by default for FB React Native renderer

bvaughn:


packages/shared/forks/ReactFeatureFlags.native-fb.js
```diff
@@ -23,6 +22,7 @@ export const {
 } = require('ReactFeatureFlags');
```
sebmarkbage: The rest of these, are they still relevant to be dynamic or can we just turn them on or off?

yungsters: Only `debugRenderPhaseSideEffects` is currently dynamically configured internally.

We always set `debugRenderPhaseSideEffectsForStrictMode` to true.

bvaughn: Looks like `debugRenderPhaseSideEffects ` is a dynamic value, `debugRenderPhaseSideEffectsForStrictMode` is being set to true in `InitializeReactFeatureFlags`, and nothing is changing the default value of `warnAboutDeprecatedLifecycles`.

bvaughn:
I've pushed another commit to this branch that makes all but one of the flags (`debugRenderPhaseSideEffects`) static.

Here are the new flag values:

| flag | value | previous value? |
| :--- | :--- | :--- |
| `debugRenderPhaseSideEffects` | (dynamic) | (dynamic) |
| `debugRenderPhaseSideEffectsForStrictMode` | `true` |  `true` |
| `disableInputAttributeSyncing` | `false` | `false` |
| `enableHooks` | `true` | (dynamic) |
| `replayFailedUnitOfWorkWithInvokeGuardedCallback` | `__DEV__` | `false`  |
| `warnAboutDeprecatedLifecycles` | `true` | `true` |

PR #14444, opened 2018-12-14
Removed enableSchedulerTracing flag and turned tracing on for all builds

bvaughn:
Too soon? 😄 I didn't spent but a few minutes on this so feel free to close.

[Be sure to look with whitespace ignored](https://github.com/facebook/react/pull/14444/files?w=1)


packages/react-reconciler/src/ReactFiberRoot.js
```diff
@@ -115,65 +99,36 @@ export function createFiberRoot(
...
+    hydrate,
+    nextExpirationTimeToWorkOn: NoWork,
+    expirationTime: NoWork,
+    firstBatch: null,
+    nextScheduledRoot: null,
+
+    interactionThreadID: unstable_getThreadID(),
+    memoizedInteractions: new Set(),
+    pendingInteractionMap: new Map(),
+  }: FiberRoot);
```
bvaughn: This is a nice cleanup, although we'll probably end up restoring a similar fork when suspense metrics are added.

packages/scheduler/src/__tests__/Tracing-test.js
```diff
@@ -48,4 +59,287 @@ describe('Tracing', () => {
...
+    });
+
+    it('should expose a subscriber ref to be externally manipulated', () => {
+      SchedulerTracing.unstable_trace('outer event', currentTime, () => {
+        expect(SchedulerTracing.__subscriberRef).toEqual({
+          current: null,
+        });
+      });
+    });
+  });
```
bvaughn: All of these tests can now be run against the built bundle 👍

packages/scheduler/src/__tests__/TracingSubscriptions-test.js
```diff
@@ -612,10 +608,4 @@ describe('TracingSubscriptions', () => {
       expect(onInteractionTraced).not.toHaveBeenCalled();
     });
   });
-
-  describe('disabled', () => {
-    beforeEach(() => loadModules({enableSchedulerTracing: false}));
-
-    // TODO
-  });
 });
```
bvaughn: All of these tests can now be run against the built bundle 👍

bvaughn:
We'll need to do this eventually but we don't need to keep this PR open. It conflicts with everything at this point anyway.

PR #14448, opened 2018-12-17
Use ES6 rest parameters in function invariant.

yifei-fu:
Use ES6 rest parameters syntax in function invariant so that the error message supports more than 6 arguments.

### DIFF
--- a/packages/shared/invariant.js
+++ b/packages/shared/invariant.js
@@ -27,7 +27,7 @@ if (__DEV__) {
   };
 }
 
-export default function invariant(condition, format, a, b, c, d, e, f) {
+export default function invariant(condition, format, ...args) {
   validateFormat(format);
 
   if (!condition) {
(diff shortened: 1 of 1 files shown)


aweary:
Thanks, but this was done on purpose to avoid the overhead of the compiled output, which [iterates over `arguments`](https://babeljs.io/repl#?babili=false&browsers=chrome%20%3E%2060&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABDMA3AhgJxusUAUAtgM4DmANIgHQ1anECUiA3gFCKKYCmUImSdYqwC-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=false&presets=es2015%2Creact%2Cstage-2%2Cstage-3%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2)

Issue #10181, opened 2017-07-13
Firefox error: Permission denied to access property "target"


vineetgopal:
**Do you want to request a *feature* or report a *bug*?**
Bug.

**What is the current behavior?**
Certain events in Firefox 54.0 trigger the following error:

```
Error: Permission denied to access property "target"
  at target(~/react/lib/getEventTarget.js:23:0)
  at getEventTarget(~/react/lib/ReactEventListener.js:58:0)
  at method(~/react/lib/Transaction.js:138:0)
  at transaction(~/react/lib/ReactDefaultBatchingStrategy.js:63:0)
  at batchingStrategy(~/react/lib/ReactUpdates.js:98:0)
  at ReactUpdates(~/react/lib/ReactEventListener.js:150:0)
  at func(~/raven-js/src/raven.js:265:0)
```

Note that raven is just wrapping the call, and I don't think that should be relevant.

The specific line of code this happens on is:
```
function getEventTarget(nativeEvent) {
  var target = nativeEvent.target || nativeEvent.srcElement || window;
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Property_access_denied suggests it could be due to an iframe and cross-origin issues. We do have some iframes embedded, although I haven't been able to reproduce the error.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/84v837e9/).**
I was unfortunately not able to reproduce this, though it has happened for ~200 separate users since the Firefox 54.0 upgrade on June 13, 2017. If it helps, we reported a similar error a while ago to raven.js (https://github.com/getsentry/raven-js/issues/768) which was fixed using a try/catch.

**What is the expected behavior?**
No errors :)

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
React 15.3.2
Firefox 54.0, Windows and OSX

I think this is only an issue with the new Firefox, and not new behavior in React.

Manuelandro:
https://github.com/getsentry/raven-js/issues/768

aweary:
Without a reproduction, there's unfortunately not much we can do. Please feel free to ping me if you can provide a consistent repro with React. Thanks!

StephanieYHe:
We're experiencing this problem too, and it's affecting users on Firefox 54, 55 and 56. No repro as well but just wanted to note that this issue continues!

mqklin:
[...]

PR #14495, opened 2018-12-26
Catch case causing `Permission denied error` error to appear on Firefox

lamhieu-vk:
this issue is similar to the issue https://github.com/facebook/react/issues/10181

in Firefox, anchorNode and focusNode can be "anonymous divs", e.g. the up/down buttons on an <input type="number">. Anonymous divs do not seem to expose properties, triggering a "Permission denied error" if any of its properties are accessed. The only seemingly possible way to avoid erroring is to access a property that typically works for non-anonymous divs and catch any error that may otherwise arise. read more https://bugzilla.mozilla.org/show_bug.cgi?id=208427

please review and merge it, thank you!

_update: screenshot error record_ 

image

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponentTree.js
+++ b/packages/react-dom/src/client/ReactDOMComponentTree.js
@@ -23,18 +23,28 @@ export function precacheFiberNode(hostInst, node) {
  * ReactDOMTextComponent instance ancestor.
  */
 export function getClosestInstanceFromNode(node) {
-  if (node[internalInstanceKey]) {
-    return node[internalInstanceKey];
-  }
-
-  while (!node[internalInstanceKey]) {
-    if (node.parentNode) {
-      node = node.parentNode;
-    } else {
-      // Top of the tree. This node must not be part of a React tree (or is
-      // unmounted, potentially).
-      return null;
+  // In Firefox, anchorNode and focusNode can be "anonymous divs", e.g. the
+  // up/down buttons on an <input type="number">. Anonymous divs do not seem to
+  // expose properties, triggering a "Permission denied error" if any of its
+  // properties are accessed. The only seemingly possible way to avoid erroring
+  // is to access a property that typically works for non-anonymous divs and
+  // catch any error that may otherwise arise. See
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=208427
+  try {
+    if (node[internalInstanceKey]) {
+      return node[internalInstanceKey];
     }
+    while (!node[internalInstanceKey]) {
+      if (node.parentNode) {
+        node = node.parentNode;
+      } else {
+        // Top of the tree. This node must not be part of a React tree (or is
+        // unmounted, potentially).
+        return null;
+      }
+    }
+  } catch (e) {
+    return null;
   }
 
   let inst = node[internalInstanceKey];


lamhieu-vk:
@bvaughn @sophiebits I think it's great if you review or give me a comment for PR, many of my users are having problems with it! thank you very much!

bvaughn:
@lamhieu-vk Can you provide a repro case for Firefox that triggers this behavior?

packages/react-dom/src/client/ReactDOMComponentTree.js
```diff
@@ -23,18 +23,28 @@ export function precacheFiberNode(hostInst, node) {
...
-      // unmounted, potentially).
-      return null;
+  // In Firefox, anchorNode and focusNode can be "anonymous divs", e.g. the
+  // up/down buttons on an <input type="number">. Anonymous divs do not seem to
+  // expose properties, triggering a "Permission denied error" if any of its
+  // properties are accessed. The only seemingly possible way to avoid erroring
+  // is to access a property that typically works for non-anonymous divs and
+  // catch any error that may otherwise arise. See
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=208427
+  try {
```
gaearon: This is a hot path so I don't think we can afford to try-catch here.

gaearon:
From a cursory glance it looks to me like something Firefox should fix on their side. Is this a wrong conclusion? Is there a more up-to-date bug describing this?

lamhieu-vk:
@bvaughn It in my system, we have a logger in client side to log error throw in production!

lamhieu-vk:
hi @gaearon ,I don't think we should wait for Firefox to use it, but I have seen some parts in React code that work together.
https://github.com/facebook/react/blob/6c404d82aa082a008dba033eb3f0f9c6803b13c4/packages/react-dom/src/client/ReactDOMSelection.js#L33

packages/react-dom/src/client/ReactDOMComponentTree.js
lamhieu-vk: hi @gaearon , In the first, thank you for review this PR!
In our bug log, I realize this is where the error appears, it only appears in some versions of Firefox!

lamhieu-vk:
hi @gaearon , I tried to reproduce this error but I couldn't do it and it still appeared to my users, I noticed through the user error log! I noticed, it only appears on Firefox (Gecko) browsers and it doesn't work on other browsers!

ref for same prolem:
https://github.com/facebook/react/issues/10181
https://github.com/rollbar/rollbar.js/issues/391
https://github.com/angular/protractor/issues/4272
https://github.com/getsentry/sentry-javascript/issues/768

please give us your comment!

lamhieu-vk:
_updated: log error from client_

image

image

image

image

lamhieu-vk:
_updated: log error from client_

image

image

image

gaearon:
We need a reproducible demo.

lamhieu-vk:
I think this PR should close, many thanks for all to support me!

sangquangnguyen:
> I think this PR should close, many thanks for all to support me!

Hi @lh0x00 , could I know why this PR is close please? is the issue fixed yet?

lh0x00:
@sangquangnguyen nope, but I cannot create a small project to reproduce at that time. if u can reproduce it, can u send me for debug and re-open this PR (if the same issue).

PR #14503, opened 2018-12-29
Simplify wording of key warning

sophiebits:
I don't think "array or iterator" is adding anything, and it may well be confusing, especially since this is one of the first and most common warnings that devs see.


TryingToImprove:
I am not a fan of this change. Since there are no "lists" in javascript.

Searching for "JavaScript list" on Google will not provide the same results as "JavaScript iterator" or array.

gaearon:
If we change this here's our chance to also clarify "unique" only applies to siblings and not globally.

sophiebits:
Hmm I kinda liked the simplicity.

PR #14313, opened 2018-11-23
Adding isMemo check to react-is package

jintoppy:
This pull request is related to #14278 

What does it do?

Adds isMemo check to the react-is package.

Why?

Adding isMemo check will help to write a clean code for packages which are doing custom rendering, like Enzyme.

Please see the issue and PR raised on Enzyme

airbnb/enzyme#1875
airbnb/enzyme#1914

PR #12932, opened 2018-05-29
[react-is] Add elementType API

aweary:
Resolves https://github.com/facebook/react/issues/12882

Adds an API for checking what the element type would be if the value was passed to `React.createElement`. It's essentially `isValidElementType` except that it returns the element type.

Issue #12882, opened 2018-05-22
ReactIs.typeOf for non-elements


mridgway:
**What is the current behavior?**

`ReactIs.typeOf` currently only works for element types, it does not allow you to pass a raw `Component` or `ForwardRef` to know the type.

The use case for this is in `hoist-non-react-statics` I now need a [special cases for `ForwardRefs`](https://github.com/mridgway/hoist-non-react-statics/issues/48). To do this, I will need to know the type of the source and target components, but currently I would need to turn them into elements first.

All of the `ReactIs.is*` functions also have this issue since they use the `typeOf` function internally.

```js
const ForwardComponent = React.forwardRef(() => {});

ReactIs.typeOf(ForwardComponent); // undefined
ReactIs.typeOf(React.createElement(ForwardComponent)); // Symbol(react.forward_ref)
```

**What is the expected behavior?**

Ideally I could pass in just the Component and get the type of it: 
```js
ReactIs.typeOf(ForwardComponent) // Symbol(react.forward_ref)
```

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

ReactIs@16.3.2

aweary:
Hey @mridgway, thanks for bringing this up. As I understand it, the intention with `ReactIs.typeOf` is that it reads the types of different _elements_; it's meant to type check values that are returned from `React.createElement`. There's also `ReactIs.isValidElementType` which tells you whether a value can be _passed_ to `React.createElement`. So there's a certain symmetry there.

Changing `typeOf` to return the type for `ForwardRef` element type and `ForwardRef` elements could create problems if anyone relies on `typeOf` to only defined values for things that can be rendered (things returned from `React.createElement`). Maybe that's OK in this case, I'm not sure.

It sounds like what you want is something in between the two that tells you the type for element types.

```jsx
React.elementType('div')
// Symbol(react.element)
React.elementType(React.StrictMode)
// Symbol(react.strict_mode)
React.elementType(React.forwardRef(Component))
// Symbol(react.forward_ref)
```

Do you think an API like that would suffice for your use case(s)? @bvaughn what do you think?

bvaughn:
> As I understand it, the intention with `ReactIs.typeOf` is that it reads the types of different _elements_; it's meant to type check values that are returned from `React.createElement`.

That's correct.

[...]

PR #14546, opened 2019-01-08
Fix memo and lazy element types being considered elements

eps1lon:
`ReactIs.typeOf` is currently returning an element type for  `lazy` and `memo` even if the given value is not an element (i.e. return value of `React.createElement`) but and elementType. This is confusing since @sebmarkbage spoke against an `elementType` API in https://github.com/facebook/react/pull/12932#pullrequestreview-133808918. 

This means that it is currently possible to determine if a given elementType is a lazy or memo component. 

This PR changes the following behavior so that lazy and memo behave similar to e.g. forwardRef when passed to `is*` or `typeOf`.

```js
// current
const Forward = React.forwardRef(() => null);
const Memo = React.memo(() => null);

ReactIs.isValidElementType(Forward) === true;
ReactIs.isForward(Forward) === false;
ReactIs.isForward(<Forward />) === true;

ReactIs.isValidElementType(Memo) === true;
ReactIs.isMemo(Memo) === true;

// after
ReactIs.isMemo(Memo) === false;
ReactIs.isMemo(<Memo />) === true;
```

I just want to add that I would prefer if `ReactIs.typeOf(Memo)` *and* `ReactIs.typeOf(FowardRef)` would return the element type but as far as I understood @sebmarkbage both should return `undefined`.

Pinging @ljharb since it sounded like (https://github.com/facebook/react/pull/14313#pullrequestreview-178105229) that would be a breaking change for `enzyme`.

Related:
https://github.com/facebook/react/issues/12882#issuecomment-392267309
> `typeOf` checks the return value of `createElement`

or `ReactDOM.createPortal`.

### DIFF
--- a/packages/react-is/src/ReactIs.js
+++ b/packages/react-is/src/ReactIs.js
@@ -48,14 +48,14 @@ export function typeOf(object: any) {
             switch ($$typeofType) {
               case REACT_CONTEXT_TYPE:
               case REACT_FORWARD_REF_TYPE:
+              case REACT_LAZY_TYPE:
+              case REACT_MEMO_TYPE:
               case REACT_PROVIDER_TYPE:
                 return $$typeofType;
               default:
                 return $$typeof;
             }
         }
-      case REACT_LAZY_TYPE:
-      case REACT_MEMO_TYPE:
       case REACT_PORTAL_TYPE:
         return $$typeof;
     }
--- a/packages/react-is/src/__tests__/ReactIs-test.js
+++ b/packages/react-is/src/__tests__/ReactIs-test.js
@@ -150,17 +150,17 @@ describe('ReactIs', () => {
 
   it('should identify memo', () => {
     const Component = () => React.createElement('div');
-    const memoized = React.memo(Component);
-    expect(ReactIs.typeOf(memoized)).toBe(ReactIs.Memo);
-    expect(ReactIs.isMemo(memoized)).toBe(true);
+    const Memoized = React.memo(Component);
+    expect(ReactIs.typeOf(<Memoized />)).toBe(ReactIs.Memo);
+    expect(ReactIs.isMemo(<Memoized />)).toBe(true);
     expect(ReactIs.isMemo(Component)).toBe(false);
   });
 
   it('should identify lazy', () => {
     const Component = () => React.createElement('div');
-    const lazyComponent = React.lazy(() => Component);
-    expect(ReactIs.typeOf(lazyComponent)).toBe(ReactIs.Lazy);
-    expect(ReactIs.isLazy(lazyComponent)).toBe(true);
+    const LazyComponent = React.lazy(() => Component);
+    expect(ReactIs.typeOf(<LazyComponent />)).toBe(ReactIs.Lazy);
+    expect(ReactIs.isLazy(<LazyComponent />)).toBe(true);
     expect(ReactIs.isLazy(Component)).toBe(false);
   });
 


eps1lon:
So from my understanding there is is currently an asymmetry between e.g. `React.forwardRef` and `React.memo` with regards to `React.is*`. Before 16.7 we had the following signature: `React.isType(value: any): value is React.ReactElement` which basically meant that we would get something like `React.isForwardRef(React.createElement(Component)) === true`.

`ReactIs.isMemo` now accepts element types (or `React.ReactComponentType` if you're familiar with the typescript declarations for react) instead of elements which was rejected for `React.forwardRef` because "introspection is bad". So I'm curious why this introspection is ok for lazy and memo element types but bad for forwardRef types.

@ljharb react-devtools has no dependency on react-is. Looks like they're directly using the symbols: https://github.com/facebook/react-devtools/blob/4d30604d971b6969982c1346748c6721def0e618/backend/attachRendererFiber.js

> Neither memo nor lazy actually tries to be an invisible wrapper - own properties aren’t copied over, including defaultProps, and including the displayName.

Which is also true for `forwardRef` so I don't think that argument applies here.

ljharb:
I think the introspection should also be possible for forwardRef types.

ljharb:
Simply knowing that it's "lazy" - and not just a regular old Promise - would be sufficient introspection for enzyme (although providing any sort of info, like a name, would be super helpful too).

eps1lon:
I'm not really following that hole introspection is bad/ok argument. Can you link me some resources that explain that concept and why it might be bad for some compilers? 

I just can't see the pattern that disallows `ReactIs.isforwardRef(Component)` but allows `ReactIs.isMemo(Component)`.

ljharb:
That there's a current lack of consistency doesn't mean the right direction is removing functionality that enzyme depends on.

eps1lon:
> That there's a current lack of consistency doesn't mean the right direction is removing functionality that enzyme depends on.

I agree with that but I all did was follow the documented path that is currently ignored in the implementation. I'm all for a `is*Type` or `elementType` API similar to #12932.

machester4:
Hi, I'm having this problem, could it be related? [https://github.com/zeit/next.js/issues/6117](url)

bvaughn:
I did not see this discussion, but I have opened a PR to fix the inconsistent/mistaken behavior:
https://github.com/facebook/react/pull/17278

bvaughn:
Regarding @ljharb's concerns about "removing functionality that enzyme depends on", fixing this inconsistency would not prohibit libraries like Enzyme from using the following pattern:

```js
ReactIs.isForwardRef(Thing) ||
ReactIs.isValidElementType(Thing) && ReactIs.isForwardRef(<Thing />
```

bvaughn:
I've checked out Enzyme and installed a local build of `react-is` (with my change from https://github.com/facebook/react/pull/17278) and all of the tests pass without any code changes on the Enzyme side.

I think that's because Enzyme is _already_ [wrapping values in "fake elements"](https://github.com/airbnb/enzyme/blob/8e2d65cc4afdbca0df34308a75bb96fa64e46ce6/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js#L254-L256) before passing them to `react-is`.

bvaughn:
I've merged PR #17278 which has this fix as well (along with some other added tests). Sorry for missing this PR earlier. I wasn't tagged on it and I just didn't notice.

PR #14548, opened 2019-01-08
Separate current owner and dispatcher

bvaughn:
This is part tech-debt cleanup, since the dispatcher didn't really belong in the owner ref to begin with.

Also this will be used by the DevTools to support hooks. DevTools need a way to modify the current dispatcher, so renderers will need to inject this ref at runtime.

Note that external libraries depending on React internals may be impacted by this, but [there are migration strategies involving feature detection](https://github.com/facebook/react/pull/14548#issuecomment-452448465) and in general, we don't really support accessing the internals data structure.

### DIFF
--- a/packages/react-cache/src/ReactCache.js
+++ b/packages/react-cache/src/ReactCache.js
@@ -46,11 +46,12 @@ const Pending = 0;
 const Resolved = 1;
 const Rejected = 2;
 
-const currentOwner =
-  React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.ReactCurrentOwner;
+const ReactCurrentDispatcher =
+  React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
+    .ReactCurrentDispatcher;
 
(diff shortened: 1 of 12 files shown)


TrySound:
Could you mention this in changelog? Me and probably somebody else use currentOwner to read from context since v16.6

bvaughn:
I guess we can add an entry to the changelog about this, since it impacts e.g. Relay

They would need to change [this](https://github.com/facebook/relay/blob/d5170e9a08408e26bc41605900d86a32801a9543/packages/react-relay/modern/readContext.js) to something more like this:
```js
const React = require("React");

const {ReactCurrentDispatcher, ReactCurrentOwner}
  = React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;

function readContext<T>(Context: React.Context<T>): T {
  const dispatcher =
    ReactCurrentDispatcher != null
      ? ReactCurrentDispatcher.current
      : ReactCurrentOwner.currentDispatcher;
  return dispatcher.readContext(Context);
}

module.exports = readContext;
```

cc @kassens

Issue #14481, opened 2018-12-21
Please publish new alpha with v16.7 bugfix and Hooks


derek-duncan:
**Do you want to request a *feature* or report a *bug*?**
¯\_(ツ)_/¯

**What is the expected behavior?**
The blog post we pretty clear about Hooks not being available for a while, but for those of us who have already begun preparing for the hooks release, it would be great to have an updated alpha with the latest bugfixes included. Thank you!

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
v16.7.0-alpha.2 hooks release is not up to date with the latest version v16.7.

pelotom:
It's also misleading that `16.7.0` has been released which looks like a finalized version of `16.7.0-alpha.x`, but has features removed (i.e. hooks). I understand the reasoning behind why 16.7.0 doesn't have hooks, but it might mitigate some confusion to publish a new `16.8.0-alpha.0` which picks up where the previous alphas left off (and hopefully incorporates bugfixes from 16.7 of course).

aweary:
This seems reasonable, but keep in mind that we're heading into the holidays here so please be patient as it may take some time before anyone is available to address this. Thanks!

beheh:
@EvgenyOrekhov In that case you might want to specifically depend on `16.7.0-alpha.2` (or `=16.7.0-alpha.2`) instead of `^16.7.0-alpha.2` in your `package.json`.

cchamberlain:
Yep, can confirm that this breaks existing dependencies on hooks due to semver range taking the non alpha. Doing a `yarn add react@next` (without `-E`) adds `^16.7.0-alpha.2` to dependencies, which ends up resolving to `16.7.0` now. Simply publishing the same exact thing that's in `16.7.0-alpha.2` under a `16.8.0-alpha.0` version would save a lot of headaches. Could follow up with a patch for the other issue in a patch release to it.

dannyphillips:
Any updates on this? Is there a plan to release a new alpha with Hooks support?

gaearon:
`16.8.0-alpha.0` should be out. Don't forget to update *both* `react` and `react-dom`.

Kovensky:
Do react-is and react-test-tenderer also need matching versions?

gaearon:
Generally every update needs matching versions.

PR #14555, opened 2019-01-09
Prepare 16.8 alpha

gaearon:
To fix https://github.com/facebook/react/issues/14481.

This matches the feature set enabled in 16.7 alpha — even though we won't actually ship ConcurrentMode in the next minor. We'll need to have another discussion about what `next` tag means and whether it should be more like RCs (with `future` tag pointing to all-included builds). But I think this discussion can wait until we ship Hooks in a stable.

This release is mostly motivated by fixing https://github.com/facebook/react/issues/14481#issuecomment-451700486.


gaearon:
(The plan is not to merge this but to do a canary from this CI build. Then we promote that canary to `16.8.0-alpha.0`).

gaearon:
`16.8.0-alpha.0` is published.

PR #14556, opened 2019-01-09
react-debug-tools accepts currentDispatcher ref as param

bvaughn:
In support of https://github.com/facebook/react-devtools/pull/1272.

The `react-debug-tools` package can't import shared internals because it isn't bundled along with the application. Instead it needs to accept those internals from DevTools, which gets them from the renderer (#14550).

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -20,7 +20,7 @@ import {
   ForwardRef,
 } from 'shared/ReactWorkTags';
 
-const ReactCurrentDispatcher = ReactSharedInternals.ReactCurrentDispatcher;
+type CurrentDispatcherRef = typeof ReactSharedInternals.ReactCurrentDispatcher;
 
 // Used to track hooks called during a render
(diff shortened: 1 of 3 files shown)


bvaughn:
I think it would be nicer for `inspectHooksOfFiber` to accept a `props` as a parameter as well. (Currently it reads props from the  fiber, which means that DevTools needs logic to determine which is the _current_ fiber.)

I can hack around this in DevTools by comparing the generic `data.state` to the "opaque node"'s `memoizedState`, but that's kind of leaky and gross.

Thoughts?

sebmarkbage:
Can we make `currentDispatcher` an optional argument which defaults to looking up in `require('react')` but allows you to override it from DevTools? That way anyone using this tool bundled with their app such as if you install it from npm, it just works out of the box without having to drill into the secret object? But you still have the option to.

bvaughn:
Hm...I guess we could do that, although it would require DevTools to stub/mock the "react" module which seems awkward?

bvaughn:
Okay. Back to you @sebmarkbage

PR #14557, opened 2019-01-09
[WIP] Try to fix scheduler fixture

gaearon:
It's currently broken. Partially from #14025 but maybe there's bugs. I don't know yet.

I tried to fix it but 4 is still broken.

screen shot 2019-01-09 at 8 19 49 pm

I'm not sure if I need to change the expectations?

Also, 8 is broken but only if I go through every test before it first:

screen shot 2019-01-09 at 8 20 20 pm

It passes in isolation.


gaearon:
This might still be broken. In which case we need to revisit. Or maybe this is already fixed. I don't know.

PR #14559, opened 2019-01-10
Support configurable labels for custom hooks

bvaughn:
Builds on top of PR #14556 to add support for configurable labels for custom hooks (for DevTools) by way of a new, debug-only hook named `useDebugValue`.

Using the debounce hook from [usehooks.com](https://usehooks.com/) as an example:
```js
function useDebounce(value, delay) {
  const [debouncedValue, setDebouncedValue] = useState(value);

  // Show the debounced value inside of DevTools beside this hook:
  useDebugValue(debouncedValue);

  useEffect(
    () => {
      const handler = setTimeout(() => setDebouncedValue(value), delay);
      return () => clearTimeout(handler);
    },
    [value, delay]
  );

  return debouncedValue;
}
```

This would render like so:
screen shot 2019-01-09 at 9 51 25 pm


sebmarkbage:
When I first saw the name “label” I thought it would be a string to name the field itself - which I don’t think we should do. But perhaps others could be confused by the name as I was.

How about `useDebugValue`?

packages/react/src/ReactHooks.js
```diff
@@ -110,3 +110,8 @@ export function useImperativeMethods<T>(
   const dispatcher = resolveDispatcher();
   return dispatcher.useImperativeMethods(ref, create, inputs);
 }
+
+export function useDebugValueLabel(valueLabel: string) {
+  const dispatcher = resolveDispatcher();
+  return dispatcher.useDebugValueLabel(valueLabel);
```
sebmarkbage: Can we wrap this in `if (__DEV__) {`? That would make this function a noop which can be trivially eliminated by a bundler. At least with ES modules. So that would let us remove the callsite itself in prod bundles.

Kovensky:
The function it calls in the dispatcher itself is already a noop, unless
devtools are installed.

sebmarkbage:
Not all dispatchers are noops so there still needs to be a lookup of which dispatcher is active and that thing called. A call is not a noop.

I.e. a bundler doesn’t know that it will be a noop so it can’t delete it. A VM can’t neither because any dispatcher can be installed that isn’t a noop.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -588,6 +588,12 @@ export function useImperativeHandle<T>(
   }, nextInputs);
 }
 
+export function useDebugValue(valueLabel: string): void {
+  // This hook is normally a no-op.
+  // The react-debug-hooks package injects its own implementation
+  // so that e.g. DevTools can display customhook values.
```
gaearon: Typo: "custom hooks"

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -402,24 +412,53 @@ function buildTree(rootStack, readHookLog): HooksTree {
       subHooks: [],
     });
   }
+
+  // Associate custom hook values (useInpect() hook entries) with the correct hooks
```
gaearon: No longer called `useInspect`

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -402,24 +412,53 @@ function buildTree(rootStack, readHookLog): HooksTree {
...
+  rootChildren.forEach(hooksNode => rollupDebugValues(hooksNode));
+
   return rootChildren;
 }
 
+function rollupDebugValues(hooksNode: HooksNode): void {
```
gaearon: What does this function do exactly? I struggle to understand just from reading the code — maybe because of `filter` with a mutation and recursion inside. Not saying it needs to be changed but a brief comment about what it does to the tree would help.

packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.internal.js
```diff
@@ -212,6 +212,122 @@ describe('ReactHooksInspectionIntergration', () => {
     ]);
   });
 
+  describe('useDebugValue', () => {
```
gaearon: Should we also have a test for `useDebugValue` on top level of component? Outside a Hook.

gaearon: Also should we add tests for updates? To verify `inspectHooksOfFiber` can find latest version. (In case it matters — I don't understand this enough to tell yet.)

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -402,24 +412,53 @@ function buildTree(rootStack, readHookLog): HooksTree {
...
+  let useInpectHooksNodes: Array<HooksNode> = [];
+  hooksNode.subHooks = hooksNode.subHooks.filter(subHooksNode => {
+    if (subHooksNode.name === 'DebugValue') {
```
sebmarkbage: Maybe we should tag these nodes with something more unique to the built-in one so that it doesn't happen with custom hooks that happen to share the same name. E.g. You could also check that this is the terminal node. children.length === 0

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -402,24 +412,53 @@ function buildTree(rootStack, readHookLog): HooksTree {
...
+
+  // Associate custom hook values (useInpect() hook entries) with the correct hooks
+  rootChildren.forEach(hooksNode => rollupDebugValues(hooksNode));
+
   return rootChildren;
 }
 
+function rollupDebugValues(hooksNode: HooksNode): void {
+  let useInpectHooksNodes: Array<HooksNode> = [];
+  hooksNode.subHooks = hooksNode.subHooks.filter(subHooksNode => {
```
sebmarkbage: Maybe just make this into a for-loop. Like Dan, I find it confusing if something is both functional style and imperative. I assume this to be pure if it's using `filter`.

bvaughn:
> What happened to the transform function as second arg? Do we expect that we don't need that?

I forgot about it 😄 I'll add it.

packages/react-debug-tools/src/ReactDebugHooks.js
bvaughn: Good point. I meant to do this but forgot. 😁

bvaughn: This seems trickier than I thought because of how we're parsing the stack for function names and then comparing them to the primitive labels.

For example, I can name the local function something more unique (e.g.  `useDebugValue__REACT_INTERNAL`) but if it's assigned to the dispatcher as `useDebugValue`– then function name that's being parsed and passed to `isReactWrapper` is `useDebugValue`.

Maybe I'm overlooking something obvious. I'll come back to this.

bvaughn:
Ping! 😄

gaearon:
What about partial renderer (server), I don't see it in this PR? Should be a no-op too, otherwise it'll crash on use.

bvaughn:
Rats. It's easy to forget about that use case. Thanks for the catch! 😄

I'll pick this up tomorrow with another PR!

PR #14565, opened 2019-01-10
rename useImperativeMethods -> useImperativeHandle

threepointone:
just a project-wide search/replace. I'll send changes for docs in a bit too. 
I have done the builds, run the tests, linted and flow checked.


packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -174,7 +174,7 @@ function useImperativeMethods<T>(
     instance = ref.current;
   }
   hookLog.push({
-    primitive: 'ImperativeMethods',
+    primitive: 'ImperativeHandle',
```
gaearon: Heads up to @bvaughn in case DevTools reads this

bvaughn:
Thanks for the heads up Sunil. If only DevTools just "read" that 😅 

Right now, I've had to fork that file into DevTools entirely. So I'll need to copy over this change too.

Issue #14542, opened 2019-01-08
`useReducer` does not honor reducer function's default state value


sampathsris:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Please see the CodeSandBox here: https://codesandbox.io/s/v8y72ky0o5

Consider the following reducer:

```js
function reducer(state = 5, { type }) {
  switch (type) {
    case 'INCREMENT':
      return state + 1
    case 'DECREMENT':
      return state - 1
    default:
      return state
  }
}
```

Suppose we are trying to use the reducer in a component like so:

```jsx
function Counter() {
  const [count] = useReducer(reducer)
  return <span>{count}</span>
}
```

When rendered,  `count` is `undefined`.

**What is the expected behavior?**

One would expect `count` to be equal to `5`, because that is the default value for state in the reducer.

IMO, it would make sense if `initialState` argument for `useReducer` is undefined, to call `reducer(undefined, undefined)` to initialize the internal state.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

16.7.0-alpha.2

carlosrberto:
@sampathsris the default state value will be returned when an invalid action is dispatched.
The initial state for `useReducer` hook should be passed as a second argument to `useReducer`:
```js
useReducer(reducer, initialState);
```
I understand your point. When using Redux, in the moment when the store is created Redux dispatch an initialization action and all reducers return the default state.

Maybe something like this could be implemented in the `useReducer` hook.

ckknight:
`useReducer(reducer, undefined, { type: 'INIT' })` will work as you expect. It will execute the reducer initially, once, with a `{ type: 'INIT' }` action, thus falling through to the default case. Because `initialState` is provided as `undefined`, it will appropriately default to `5` as specified in the reducer.

carlosrberto:
@sampathsris I opened a [pull request](https://github.com/facebook/react/pull/14566) with a proposal to make this works.

gaearon:
I think `useReducer(reducer, undefined, { type: 'INIT' })` is what we suggest so that strong typing can work.

carlosrberto:
What do you think about #14566 @gaearon ?
Any chance of these changes be introduced ?

gaearon:
Did you see my comment above?

ckknight:
I don’t like the idea of an untyped action being dispatched implicitly without the ability to prevent that.

carlosrberto:
[...]

PR #14566, opened 2019-01-10
Use default reducer state value as initial state when using useReducer hook

carlosrberto:
## Use default reducer state value as initial state when using useReducer hook

Based on discussion from issue #14542.

If we provide the following reducer:

```js
function reducer(state = 5, { type }) {
  switch (type) {
    case 'INCREMENT':
      return state + 1;
    case 'DECREMENT':
      return state - 1;
    default:
      return state;
  }
}
```

Then in the component:

```js
function Counter() {
  const [count] = useReducer(reducer);
  return <span>{count}</span>;
}
```

We don't get count equal `5` in the initial render.

It works if we provide the `initialState` as second argument for `useReducer`:

```js
const [count] = useReducer(reducer, 5);
``` 

or like @ckknight mentioned if we call an initialization action:

```js
useReducer(reducer, undefined, { type: 'INIT' });
```

However I think it's a little redundant because we already set the default state value in the reducer function.

This pull request add some changes to use the default state value from the reducer function if no `initialState` was provided to `useReducer`.

I don't know if this behaviour was discussed during the implementation of `useReducer`, but if this is not the desired behaviour for `useReducer` feel free to close this pull request.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
@@ -209,6 +209,23 @@ describe('ReactDOMServerHooks', () => {
       expect(domNode.textContent).toEqual('0');
     });
 
+    itRenders('with default reducer state value', async render => {
+      function reducer(state = 0, action) {
+        return action === 'increment' ? state + 1 : state;
+      }
+      function Counter() {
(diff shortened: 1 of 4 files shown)


carlosrberto:
I'm closing this PR, since it's not the recommended way to get the initial state from reducer
https://reactjs.org/docs/hooks-reference.html#lazy-initialization

PR #14567, opened 2019-01-10
Add hooks support to ReactShallowRenderer

trueadm:
This PR adds React Hooks support to React's test Shallow Renderer. The implementation is mostly based on the ReactDOMServer hooks dispatcher implementation, with a few tweaks to account for the fact that shallow renderer instances do not share state. The ReactDOMServer implementation can be found here:

https://github.com/facebook/react/blob/master/packages/react-dom/src/server/ReactPartialRendererHooks.js#L340-L354

I've added a selection of tests too that confirm the expected behaviour works as intended.


packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -99,6 +118,259 @@ class ReactShallowRenderer {
     this._rendering = false;
     this._forcedUpdate = false;
     this._updater = new Updater(this);
+    if (enableHooks) {
+      this._dispatcher = ReactCurrentDispatcher.current = this._createDispatcher();
```
bvaughn: nit: Should we only set the `ReactCurrentDispatcher.current` to a non-null dispatcher when actually rendering? Not sure if this is strictly necessary, but it seems more semantically correct (and is more inline with other renderers).

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -99,6 +118,259 @@ class ReactShallowRenderer {
...
+        this._validateCurrentlyRenderingComponent();
+        return readContext(context);
+      },
+      useEffect: noOp,
+      useImperativeMethods: noOp,
+      useLayoutEffect: noOp,
+      useMemo,
+      useReducer,
+      useRef,
+      useState,
```
bvaughn: Needs `useDebugValue` (noop)

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -99,6 +118,259 @@ class ReactShallowRenderer {
...
+      } else {
+        return previousRef;
+      }
+    };
+
+    const readContext = <T>(
+      context: ReactContext<T>,
+      observedBits: void | number | boolean,
+    ): T => {
+      return context._currentValue;
```
bvaughn: The partial renderer impl varies from yours:
```js
function readContext<T>(
  context: ReactContext<T>,
): T {
  let threadID = currentThreadID;
  validateContextBounds(context, threadID);
  return context[threadID];
}
```

Not sure of this reason for this. Is it a mistake?

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -99,6 +118,259 @@ class ReactShallowRenderer {
...
+  }
+
+  _validateCurrentlyRenderingComponent() {
+    invariant(
+      this._currentlyRenderingComponent !== null,
+      'Hooks can only be called inside the body of a function component.',
+    );
+  }
+
+  _createDispatcher() {
```
bvaughn: We'll want to add a return type of `typeof Dispatcher` to this one #14599 is merged

packages/react-test-renderer/src/ReactShallowRenderer.js
trueadm: We use threads on the SSR version, but I believed we didn't for the DOM version. I'll take another look at this.

gaearon: I don't think this will be relevant to shallow.

bvaughn: Agreed. Just didn't have any context on the SSR case for this and wanted verification that it wouldn't cause any unforeseen trouble here.

gaearon:
Seems like this leaks state if you render different component types one after another.

```js
  it('should not leak state when component type changes', () => {
    function SomeComponent({defaultName}) {
      const [name] = React.useState(defaultName);

      return (
        <div>
          <p>
            Your name is: <span>{name}</span>
          </p>
        </div>
      );
    }

    function SomeOtherComponent({defaultName}) {
      const [name] = React.useState(defaultName);

      return (
        <div>
          <p>
            Your name is: <span>{name}</span>
          </p>
        </div>
      );
    }

    let result = shallowRenderer.render(
      <SomeComponent defaultName={'Dominic'} />,
    );
    expect(result).toEqual(
      <div>
        <p>
          Your name is: <span>Dominic</span>
        </p>
      </div>,
    );

    result = shallowRenderer.render(
      <SomeOtherComponent defaultName={'Dan'} />,
    );
    expect(result).toEqual(
      <div>
        <p>
          Your name is: <span>Dan</span>
        </p>
      </div>,
    );
  });
```

gaearon:
Another failing test. Seems like updating state on render only works once? Then it goes into an infinite loop. But I think it should have stabilized.

```js
  it('should repeat rendering until state stabilizes', () => {
    function SomeComponent({defaultName}) {
      const [name, updateName] = React.useState(defaultName);

      if (name !== 'Dan') {
        updateName('Dan');
      }

      if (name === 'Dan') {
        updateName('Andrew');
      }

      return (
        <div>
          <p>
            Your name is: <span>{name}</span>
          </p>
        </div>
      );
    }

    const result = shallowRenderer.render(
      <SomeComponent defaultName={'Dominic'} />,
    );

    expect(result).toEqual(
      <div>
        <p>
          Your name is: <span>Andrew</span>
        </p>
      </div>,
    );
  });
```

packages/react-test-renderer/src/ReactShallowRenderer.js
Kovensky: This sounds like a race condition to me 😓

trueadm:
I've made all the relevant changes, plus added Flow to the file. I wasn't sure about applying #14569 as per @gaearon's feedback above. The server side partial renderer doesn't have the same logic applied to it, so maybe I'll wait on feedback from @acdlite as to if we need it here too.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -99,6 +187,288 @@ class ReactShallowRenderer {
...
+
+    const useMemo = <T>(
+      nextCreate: () => T,
+      inputs: Array<mixed> | void | null,
+    ): T => {
+      this._validateCurrentlyRenderingComponent();
+      this._createWorkInProgressHook();
+
+      const nextInputs =
+        inputs !== undefined && inputs !== null ? inputs : [nextCreate];
```
gaearon: To match https://github.com/facebook/react/pull/14594, let's call them `deps` (not `inputs`) and make sure we use `null` rather than `[nextCreate]` in all cases. (That's what #14594 was about.)

trueadm: I can do that, but the reason I didn't was because #14594 didn't make that change to packages/react-dom/src/server/ReactPartialRendererHooks.js.

trueadm:
@gaearon I made that change but it broke things. I don't believe the same logic is meant to be here without changing the types for `areHookInputsEqual`. Given this function was already changed in #14599 to have these flow types, I believe this to be intentional. I've made the input -> deps rename though.

packages/react-test-renderer/src/ReactShallowRenderer.js
bvaughn: > This sounds like a race condition to me

What makes you say this?

Temporarily overriding the current dispatcher is a pattern we follow in several places. As long as   each place that overrides is careful to restore on completion, what's the race?

packages/react-test-renderer/src/ReactShallowRenderer.js
gaearon: I think that's just an omission.

gaearon:
I'm not sure I fully understand what you're saying.

What I'm saying is — since https://github.com/facebook/react/pull/14594, `useMemo` and `useCallback` doesn't compare the function itself. Previously they considered it as part of deps but not anymore. We need to ensure this behavior is replicated in all implementations. I haven't looked exactly at *how* to do it (maybe it's more than changing that one line) but it needs to be done for semantics to match.

Jessidhia:
FWIW, just as `useCallback` used to be just an expensive identity function when called without deps, `useMemo` without deps array is now just an expensive `.call()` (it was before too, but there was one corner case where it wouldn't be; that corner case is now gone).

I think the deps argument should be **required** for them.

I should make an issue about it when I'm not half-asleep... 😴

trueadm:
@gaearon I've made that change. I was more suggesting there was some confusion over the referenced PR – as it didn't apply the same changed to the partial renderer and I assumed that this was intentional.

gaearon:
lgtm

koba04:
@trueadm 
This Hooks support for ShallowRenderer isn't including `useEffect` because this seems to be based on ReactDOMServer. I think `useEffect` isn't required for ReactDOMServer, but it seems to make sense for a unit testing. Do you have a plan to implement `useEffect` on ShallowRenderer?

gaearon:
No — just like shallow renderer doesn’t implement componentDid lifecycles.

ljharb:
@gaearon does that also mean that server rendering won’t invoke useEffect hooks?

gaearon:
It doesn't.

kushalmahajan:
I came to this thread since hooks is not supported by enzyme. Now only shallow rendering is stuck in my project which uses hooks throughout for all things.

After reading I tried shallow rendering instead of Enzyme.

Here's the code that I tried with React Shallow Renderer but it doesn't works for me in Release 16.7.2-alpha for React and ReactDOM.

```
describe('component', () => {
    it('on Button click text changes', () => {
        const renderer = new ShallowRenderer();
        renderer.render(<CallsHistory header='jhad'> Sample text</CallsHistory>);
        const result = renderer.getRenderOutput();
        const button = result.find('.toggleBtn');
        expect(result.find('.buttonText').text()).toBe('No');
        button.simulate('click');
        expect(result.find('.buttonText').text()).toBe('Yes');
    });
})
```
Although, I'm not sure if I can use simulate here as in enzyme. So please let me know if there's a way to make this work as the error is same `Invariant Violation: Hooks can only be called inside the body of a function component.`

```
  button.simulate('click');
```

gaearon:
Merged, not released. ;-)

gaearon:
@ljharb I guess my comment was ambiguous. I meant that SSR does *not* invoke `useEffect`.

gaearon:
Lol. I hope within two weeks!

luskin:
@gaearon any update on this?

bvaughn:
We'll provide a big, noisy update once hooks are released. (Probably not before though.)

PR #14569, opened 2019-01-11
Allow useReducer to bail out of rendering by returning previous state

acdlite:
This is conceptually similar to `shouldComponentUpdate`, except because there could be multiple `useReducer` (or `useState`) Hooks in a single component, we can only bail out if none of the Hooks produce a new value. We also can't bail out if any the other types of inputs — state and context — have changed.

These optimizations rely on the constraint that components are pure functions of props, state, and context.

In some cases, we can bail out without entering the render phase by eagerly computing the next state and comparing it to the current one. This only works if we are absolutely certain that the queue is empty at the time of the update. In concurrent mode, this is difficult to determine, because there could be multiple copies of the queue and we don't know which one is current without doing lots of extra work, which would defeat the purpose of the optimization. However, in our implementation, there are at most only two copies of the queue, and if *both* are empty then we know that the current queue must be.

Summary of changes:
- We use the `PerformedWork` effect tag bit to track whether a component's props, state, or context has changed. This means we need to mark it dirty every time we detect that one of those inputs has a new value. Props and state are fairly straightforward, but context is tricky. I had to add a field to the context object that represents whether the context object has changed (`changedBits`). This results in additional runtime cost every time context is read.
  - Alternatively, we could add a field somewhere on the fiber or update queue that we mark while scanning from the provider. This trade would memory for runtime performance.
- If a component bails out, we can remove the update priority flag from the current fiber. Normally we only mutate the current tree in the commit phase, never in the render phase. In this case it's ok to do because the update priority field only affects whether React needs to visit that fiber. If we receive new props, state, or context, then the fiber will be visited again anyway.


packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
```diff
@@ -46,25 +46,25 @@ describe('ReactNewContext', () => {
...
+  //     },
+  // );
+  // sharedContextTests('useContext inside forwardRef component', Context =>
+  //   React.forwardRef(function Consumer(props, ref) {
+  //     const observedBits = props.unstable_observedBits;
+  //     const contextValue = useContext(Context, observedBits);
+  //     const render = props.children;
+  //     return render(contextValue);
+  //   }),
+  // );
```
sebmarkbage: Why are all these tests commented out?

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1757,6 +1757,7 @@ function scheduleWorkToRoot(fiber: Fiber, expirationTime): FiberRoot | null {
           interactions.forEach(interaction => {
             if (!pendingInteractions.has(interaction)) {
               // Update the pending async work count for previously unscheduled interaction.
+
```
sebmarkbage: unrelated. plz send a separate PR and tag the whole team.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1647,9 +1691,11 @@ function bailoutOnAlreadyFinishedWork(
 
   if (current !== null) {
     // Reuse previous context list
-    workInProgress.firstContextDependency = current.firstContextDependency;
+    workInProgress.contextDependencies = current.contextDependencies;
   }
 
+  workInProgress.effectTag &= ~PerformedWork;
```
sebmarkbage: Structuring it this way gets tricky and will make resuming harder. If any pass performed work, we need it to remain as marked as performed work. This risks clearing it if a second pass bails out.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -117,13 +126,16 @@ function resolveCurrentlyRenderingFiber(): Fiber {
   return currentlyRenderingFiber;
 }
 
-export function prepareToUseHooks(
+export function renderWithHooks(
```
sebmarkbage: Why this change of pattern?

acdlite: It made more sense in an earlier iteration of this PR. It's not strictly necessary but the re-render logic is less fragile this way. I can move this to https://github.com/facebook/react/pull/14591 if needed.

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: Why do you need a per-fiber bit? Can't you just use a global flag?

acdlite: Right, I suppose that’s true. I guess I figured since I had to reset `PerformedWork` in the bailout case anyway I might as well use it for both. But global flag is better.

gaearon:
We should also change the SSR, no?

For example

```js
function Foo(props) {
  const [name, setName] = useState('Dan');
  if (props.name === 'Andrew') {
    setName(props.name);
  }
  return name; 
}

renderToString(<Foo name="Andrew" />
```

would previously go into an infinite loop but I think should now stabilize.

acdlite:
@gaearon The eager bailout doesn't apply to re-renders, but that's an interesting idea.

acdlite:
I think we should only eagerly compute render phase updates and bail out if we're sure the bailout can be relied on completely and predictably. Not sure about that. Seems safer to always require the guard around the update.

sebmarkbage:
The guard provides a natural place to other code that doesn't need to be evaluated if nothing has changed so it seems good to encourage that.

sebmarkbage:
Although, counter-argument: This optimization isn’t really meant for useState. It’s mostly for useReducer where it is difficult to know if a reducer will yield the same value or not.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -506,18 +532,38 @@ function updateFunctionComponent(
...
+    bailoutHooks(current, workInProgress, renderExpirationTime);
+    return bailoutOnAlreadyFinishedWork(
+      current,
+      workInProgress,
+      renderExpirationTime,
+    );
   }
-  nextChildren = finishHooks(Component, nextProps, nextChildren, context);
 
-  // React DevTools reads this flag.
```
sebmarkbage: You can restore this comment.

thysultan:
Does this still allow for mutable data-strucutures where something akin to the following is used for performance reasons:

```js
const [state, setState] = useState(() => [1, 2, 3, 4, 5])

setState(state => {
    state[4] = 'swap'
    return state
})
```

brunolemos:
@thysultan maybe you should use useRef for that

thysultan:
@brunolemos Yes you could also return `NaN`, though that is assuming the equality function is not using the `Object.is` semantics, in which case it is would prove to be even harder. 

That aside using refs for what is intuitively state seems counter intuitive to the advise regarding using refs in effects.

You would also need a lot more coordination in that regard as opposed to passing a function callback to `setState`.

Mutable data-structures are still the best choice when dealing with high throughput, so it would go a long way to avoid a design that paves a path with these obstacles .

gaearon:
@thysultan Functional setState callback (first argument) never supported mutation. We intentionally invoke it twice in Strict Mode to make the issue more visible. React supports mutation as escape hatch, but if you mutate things, do it in event handlers and not in the functional updater.

PR #14571, opened 2019-01-11
throw an error on useContext(Consumer) / useContext(Provider)

threepointone:
this is an odd one, because the way we've implemented it. this PR introduces an invariant in dev mode (which won't throw in prod!) but we feel that's the best call given that this is already a hot path.

### DIFF
--- a/packages/react/src/ReactHooks.js
+++ b/packages/react/src/ReactHooks.js
@@ -22,7 +22,6 @@ function resolveDispatcher() {
   return dispatcher;
 }
 
-
 export function useContext<T>(
   Context: ReactContext<T>,
   observedBits: number | boolean | void,
@@ -49,7 +48,6 @@ export function useContext<T>(
   }
   return dispatcher.useContext(Context, observedBits);
 }
-
 
 export function useState<S>(initialState: (() => S) | S) {
   const dispatcher = resolveDispatcher();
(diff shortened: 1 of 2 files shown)


gaearon:
If we go this route we'd need to amend the tests to check for different things in DEV and PROD too. (You can run prod tests with `yarn test-prod <filename>`. Also lint needs to be fixed.

I'd personally be okay with making this a DEV-only invariant, similar to how mutating props only throws in DEV. The bugs resulting from violating this are very subtle.

packages/react/src/ReactHooks.js
```diff
@@ -33,19 +33,17 @@ export function useContext<T>(
       const realContext = (Context: any)._context;
       // Don't deduplicate because this legitimately causes bugs
```
gaearon: This comment is no longer relevant — deduplication refers to warnings. Instead there should be a justification for DEV-only invariant here.

threepointone:
I’ve been thinking about it and I don’t know if this is a good idea anymore. It’s to catch a mistaken pattern that causes undefined behaviour, and we warn about it anyway. Similar to misspelling componentWillReceiveProps. Maybe we don’t want this to throw an error. I assume flow will fail anyway?

threepointone:
taking a call to abandon this.

threepointone:
abandoning this, prefer the behaviour we already have

Issue #14039, opened 2018-10-30
Improve message "dispatcher.useState is not a function" when mixing react@next with react-dom@16.5.2


geoffdavis92:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
Getting `TypeError: dispatcher.useState is not a function` when trying to call `useState`; this is happening in a create-react-app, as well as in the CodeSandbox example below.

**Example:**
https://codesandbox.io/s/7kp55lwwpj

**What is the expected behavior?**
`useState` hook function should execute without error as defined in the [API documentation](https://reactjs.org/docs/hooks-reference.html#usestate)

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

**React**: v16.7.0-alpha.0
**React-DOM**: 16.5.2 (CSB), 16.7.0-alpha.0 (local)
**React-Scripts**: 2.0.3 (CSB), 2.1.0 (local)

sophiebits:
@geoffdavis92 You need to update react-dom to the same version.

We can likely have a better error for this though. Leaving this issue open to track that.

CarberryChai:
you can `yarn add react-dom@next` and then have a try……

geoffdavis92:
@sophiebits Thank you, that was the issue. For some reason I thought I tried both packages at `next` but I suppose I did not.

This is ready to be closed IMO.

threepointone:
Broadly, this is about warning when we have mismatched versions of react and react-dom installed, yes? This particular error happens right now because we have peerDependencies set with `"react": "^16.0.0"`. This should be flagged by npm when installing, imo we should update peerDependencies to `"react": "^16.8.0"` (or whatever version) when we release hooks. 

(we could verify versions in runtime too, but that seems like overkill)

I'll send out a PR closer to hooks launch to fix this.

threepointone:
discussed with dan a bit more, the peer dependency wouldn't fix it, since this actually lives on react which doesn't have a peer dep, sigh. I'll send a PR with a better error message.

threepointone:
the latest version (16.8.0-alpha.0) of react doesn't throw this error anymore (though it does throw an invariant `Hooks can only be called inside the body of a function component.` which isn't that much better).

PR #14572, opened 2019-01-11
warn when it looks like you haven't updated react-dom

threepointone:
sketching a possible "fix" for #14039 for discussion. 

- icky error message, I know. 
- can't think of a good test for this, open to suggestions. should I mock the dispatcher to trigger this specific error? sounds self serving. 
- should probably add a TODO to remove this in v17

### DIFF
--- a/packages/react/src/ReactHooks.js
+++ b/packages/react/src/ReactHooks.js
@@ -19,6 +19,13 @@ function resolveDispatcher() {
     dispatcher !== null,
     'Hooks can only be called inside the body of a function component.',
   );
+  if (__DEV__) {
+    warning(
+      typeof dispatcher.useState === 'function',
+      "It looks like you're using an older version of your renderer, " +
+        "make sure you've updated both react and react-<renderer> in your package.json",
+    );
+  }
   return dispatcher;
 }
 


threepointone:
I should add a flag that fires this warning only once.

gaearon:
How did you test this? It's likely specific message in https://github.com/facebook/react/issues/14039 is no longer what happens. We need to actually look at what happens when e.g. using `react@16.8.0-alpha.0` and `react-dom@16.7.0` / `react-dom@16.6.0` (it's even possible the messages are different).

threepointone:
I haven't tested this, setting it up right now. will report back.

gaearon:
We kind of wanted `react-dom` to keep working with `react` of the same major without breaking. I don't think a strict match is necessarily desired unless you're trying to use new features — in which case you should have *at least* that version of both.

aweary:
>  I don't think a strict match is necessarily desired unless you're trying to use new features

Why not? Using mismatched versions isn't something people do intentionally (afaik) and if a user is already updating one, the burden of updating the other should be low.

gaearon:
If it works fine, what's the point of being too sensitive about it? In general it doesn't cause issues anymore.

threepointone:
alright, I've verified, here'se how react/react-dom at various versions interact with a simple hook - 

react@16.7.0-alpha.0
* 0.0 - ?? fails before react even starts up,  
* 3.0, 4.0, invariant - `Hooks can only be called inside the body of a function component.`
* 5.0, 6.0 - `dispatcher.useState is not a function`
* 8.0 - ` Cannot set property 'current' of undefined`

react@16.8.0-alpha.0
* all older versions fail with “Hooks can only be called inside the body of a function component."
* 0.0 - 2.0 - throws
* 3.0 - 7.0-alpha - invariant 

honestly, this is kind of an edgecase squared, mismatched versions of react/react-dom *and* using hooks recently. maybe we shouldn't do anything.

aweary:
> If it works fine, what's the point of being too sensitive about it? In general it doesn't cause issues anymore.

It does cause issues though, and even if it causes _less_ issues now there is still a chance that it will break things in the future. I don’t think there’s any reason to think that this would be less likely in the future? We don’t have any fixtures looking for undefined behavior with mismatched versions before release AFAIK

A single warning doesn’t seem too sensitive either. There’s no case where we recommend using mismatched versions, so it would always signal a real issue to the user.

gaearon:
@threepointone It "throws" on usage only, right? If we don't use Hooks then it's ok?

gaearon:
>0.0 - 2.0 - throws

throws what?

>3.0 - 7.0-alpha - invariant

which invariant?

threepointone:
@gaearon - "Hooks can only be called inside the body of a function component.". they all throw. 0.0 - 2.0 throw it as a regular error (like, codesandbox reported it as a normal error), the rest as an invariant (with the 'invariant' warning).

gaearon:
OK. I think the most actionable thing here is to make a doc page (similar to "current owner" error) that explains possible causes (calling outside render, calling inside useMemo, duplicate React, wrong version of React), and add a link to the existing invariant. What do you think?

threepointone:
Yeah, I think that would gather these loose ends in one place. I'll do it.

threepointone:
closing

PR #14585, opened 2019-01-14
RFC: warn when returning different hooks on subsequent renders

threepointone:
^ like it says. adds a field to `Hook` to track hook 'type', and compares when cloning on further renders.

started this PR/mvp to discuss how you folks would like this implemented? 

I have run the tests (and added one!)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -47,7 +48,18 @@ type UpdateQueue<A> = {
...
 
+type HookType =
+  | 'state'
+  | 'reducer'
+  | 'context'
+  | 'ref'
+  | 'effect'
+  | 'callback'
+  | 'memo'
+  | 'imperative-handle';
```
acdlite: We typically use numbers of these types of tags. See ReactWorkTags for example.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -237,8 +249,9 @@ export function resetHooks(): void {
   numberOfReRenders = 0;
 }
 
-function createHook(): Hook {
+function createHook(hookType: HookType): Hook {
   return {
+    hookType,
```
acdlite: Let’s only include this field in DEV so it doesn’t cost memory in production.

threepointone:
I have an odd bug in my test, where even though I'm matching the warning message exactly, `toWarnDev` isn't matching it right. I'll try to figure it out at work tomorrow.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -47,7 +48,18 @@ type UpdateQueue<A> = {
...
+  | 'reducer'
+  | 'context'
+  | 'ref'
+  | 'effect'
+  | 'callback'
+  | 'memo'
+  | 'imperative-handle';
+
 export type Hook = {
+  hookType: HookType,
```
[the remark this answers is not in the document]
aweary: `ReactFiber` just [has a comment and prefixes the DEV only fields with `_`](https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiber.js#L202-L206). You could maybe call it `_debugType`?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -249,8 +262,13 @@ function createHook(): Hook {
   };
 }
 
-function cloneHook(hook: Hook): Hook {
+function cloneHook(hook: Hook, hookType: HookType): Hook {
```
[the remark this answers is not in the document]
aweary: What about a top-level DEV-only variable like `currentHookType`. Set it to the expected type before calling `createWorkInProgressHook` and then read that in `cloneHook` and `createHook`? Then it doesn't have to be passed around and it would be more easily DCE'd for the prod build.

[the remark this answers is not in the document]
threepointone: I decided to read off currentHook directly!

threepointone:
thanks for reviewing my PR on a sunday y'all :)

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
aweary: @threepointone with `currentHookType` you could also avoid passing `hookType` to `cloneHook` as well. I don't think GCC will [remove unused function arguments](https://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540compilation_level%2520SIMPLE_OPTIMIZATIONS%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250A%252F%252F%2520ADD%2520YOUR%2520CODE%2520HERE%250Afunction%2520foo(a%252C%2520unused)%2520%257B%250A%2520%2520alert('Hello%252C%2520'%2520%252B%2520name)%253B%250A%257D%250Ahello('New%2520user'%252C%2520'never%2520used')%253B%250A%250A), so the `HookType` arguments won't get removed from the prod build even though they're unused

. If you do:

```js
if (__DEV__) {
  currentHookType = RefHook
}
workInProgressHook = createWorkInProgressHook();
```

Then I believe it will get DCE'd in prod, adding no new bytes.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -247,10 +264,15 @@ function createHook(): Hook {
 
     next: null,
   };
+  if (__DEV__) {
+    invariant(currentHookType !== null, 'Forgot to set currentHookType');
```
threepointone: I might just remove this

threepointone:
a couple that are missing 
- useImperativeHandle
- useContext

packages/react-reconciler/src/ReactFiberHooks.js
aweary: I think its safe to remove, a comment is probably better? Writing to a top-level variable is a pretty common pattern in the codebase, and new hook types are rare enough that a comment is probably a sufficient safeguard

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -259,6 +281,14 @@ function cloneHook(hook: Hook): Hook {
 
     next: null,
   };
+
+  if (__DEV__) {
+    if (!currentHook || currentHook._debugType !== hook._debugType) {
+      warning(false, 'Bad hook order!');
+    }
+    nextHook._debugType = hook._debugType;
+  }
```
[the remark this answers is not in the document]
acdlite: In many cases, a Hook mismatch will cause a throw so it needs to happen in that path, too. Should add a test.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -259,6 +281,14 @@ function cloneHook(hook: Hook): Hook {
 
     next: null,
   };
+
+  if (__DEV__) {
+    if (!currentHook || currentHook._debugType !== hook._debugType) {
```
aweary: This should now read from `currentHookType` right?

aweary:
Looks like there's [some overlap](https://github.com/facebook/react/pull/14594/files#diff-1adf25deacc6168be0ac290f98561486R121) with https://github.com/facebook/react/pull/14594 for tracking the hook name/type in DEV

threepointone:
@aweary hrmm yeah you're right. I'll refactor. thanks for the catch!

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -259,6 +281,14 @@ function cloneHook(hook: Hook): Hook {
...
+      warning(false, 'Bad hook order!');
```
zhigang1992: Perhaps we can improve the error message here. Add a shortened link to [Rules of Hooks ](https://reactjs.org/docs/hooks-rules.html)?

threepointone:
I've updated the PR with changes regarding some of the comments, also adding support for useImperativeHandle and useContext. 

I had to cheat with useContext, and add a hook to the hooks-list so we could track it's type. acceptable approach?

(also, woah realizing you can read context from anywhere really, not just a hook :P this PR would break that, of course)

working on 2 bits now -
- better errors, at least reporting the expected:actual hook types. would be nice to do a diff with a previous list of hooks as suggested. 
- porting this logic to the partial renderer
- a few more tests, because this is a bit serious now ha.

threepointone:
oh huh, we don't clonehooks when server rendering. neato, not needed.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -306,19 +335,33 @@ export function resetHooks(): void {
 }
 
 function createHook(): Hook {
-  return {
-    memoizedState: null,
+  let hook: Hook = __DEV__
+    ? {
+        // $FlowFixMe
```
threepointone: I could probably do better than this

Jessidhia:
Another potential error prone case is render props. Whatever hooks are called in the render prop belong to the child, not to your component.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -327,6 +370,24 @@ function cloneHook(hook: Hook): Hook {
 
     next: null,
   };
+
+  if (__DEV__) {
+    if (currentHookType !== hook._debugType) {
+      warning(
+        false,
+        'React just detected that you called %s in a position previously called by %s. ' +
```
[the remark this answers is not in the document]
gaearon: For sure.

threepointone:
@Jessidhia I thought about it a bit, and in general that should be fine? any specific usecase that could cause a bug? I figured useContext *might*, but feels like it's expected behaviour

Jessidhia:
I tried looking out examples for components I thought would be at risk for causing the problem (e.g. `react-spring`'s `Transition`, their new hooks API notwithstanding) but they all seem to be written as class components. It'll at least error out nearly immediately there at runtime because it's a class component.

If you did have a function component that implemented a similar pattern, though (invoking `children` a variable number of times), the hooks invoked from the `children` callback would look like hooks belonging to the inner component. It's not a big deal if `children` is always called, but if it's conditionally or iteratively called...

threepointone:
Ah that's fair. This PR should detect those, I'll make sure.

threepointone:
@acdlite @bvaughn how do you feel about this - 
screenshot 2019-01-18 at 15 47 14

bvaughn:
Looks great! Although I think the component stack ("in App (...)") should be at the top. So like...

```
Warning: React has detected a change in the order of hooks called by component SomeComponent:
     in App (at **)

On the previous render, the following hooks were called:
  * useState
  * useReducer

On the most recent render, the following hooks were called:
  * useReducer
  * useState

This will cause an error if not fixed! For more information see: https://fb.me/rules-of-hooks
```

gaearon:
>Although I think the component stack ("in App (...)") should be at the top. So like...

It can be *really* long in some apps (like hundreds of lines). So far we've always put it at the bottom I think — any particular reason you feel differently about this one?

threepointone:
(also, I'm not sure how to move the stack to the middle of a warning?)

gaearon:
Also worth checking how it looks with a deep custom Hook tree. I'm a bit worried the low-level info without custom Hook names is not as useful.

threepointone:
I've updated the PR with a an attempt of the above. I've updated tests too, but seeing a warning get doubly logged, am working on it.

gaearon:
Maybe we can capture the stack (`new Error().stack`) on mismatch and print the stack trace of the first mismatching Hook? This can at least give you an idea of which Hooks to start looking at.

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
```diff
@@ -596,4 +596,95 @@ describe('ReactHooks', () => {
...
+      }
+      return null;
+      /* eslint-enable no-unused-vars */
+    }
+    let root = ReactTestRenderer.create(<App flip={false} />);
+    expect(() => {
+      expect(() => root.update(<App flip={true} />)).toThrow();
+    }).toWarnDev(
+      [
+        'Warning: React just detected that you changed the order of hooks called across separate renders.\n' +
```
threepointone: this should log just once

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
threepointone: check it out, I log the hook chains in currentHookMatches rn

threepointone:
(btw, there are some caveats with this approach, for eg, it won't be able to detect swamped hooks of the same type)

threepointone:
> Maybe we can capture the stack (`new Error().stack`) on mismatch and print the stack trace of the first mismatching Hook? This can at least give you an idea of which Hooks to start looking at.

ok lemme give that a shot

bvaughn:
> It can be really long in some apps (like hundreds of lines). So far we've always put it at the bottom I think — any particular reason you feel differently about this one?

Sorry I was in meetings.

Long stack is a fair consideration. Hm...

I didn't like the message shown above because the stack was just floating at the bottom, disconnected from the error itself.

bvaughn:
Maybe something more like?

```
Warning: React has detected a change in the order of hooks called by FooComponent.

On the previous render, the following hooks were called:
  * useState
  * useReducer

On the most recent render, the following hooks were called:
  * useReducer
  * useState

This will cause an error if not fixed! For more information see: https://fb.me/rules-of-hooks

This error occurred in:
     in BarComponent (at **)
     in App (at **)
```

gaearon:
Sure (I also think it's more obvious what's going on when stack is more than a single line — unlike on that screenshot :-)

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -170,6 +200,40 @@ function areHookInputsEqual(
   return true;
 }
 
+function verifyHookCallOrder() {
+  let mismatched = false;
```
gaearon: Want to wrap the whole body in DEV just to make the purpose clear?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -221,6 +285,8 @@ export function renderWithHooks(
           getComponentName(Component),
         );
       }
+      verifyHookCallOrder();
+      currentHookMatches = [];
```
gaearon: Both of these statements should be DEV-only?

gaearon: Seems like maybe we could avoid re-allocating `currentHookMatches` and just do `.length = 0` here.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -298,27 +364,44 @@ export function resetHooks(): void {
 
   if (__DEV__) {
     currentHookNameInDev = undefined;
+    verifyHookCallOrder();
   }
+  currentHookType = null;
+  currentHookMatches = [];
```
gaearon: Why not in DEV?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -298,27 +364,44 @@ export function resetHooks(): void {
...
   renderPhaseUpdates = null;
   numberOfReRenders = -1;
 }
 
 function createHook(): Hook {
-  return {
-    memoizedState: null,
+  let hook: Hook = __DEV__
+    ? {
+        // $FlowFixMe - we're certain we've set currentHookType by now
```
gaearon: We use FlowFixMe for bugs in Flow — I think this just needs to be conversion via any. `(foo: any): desiredtype`

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -298,27 +364,44 @@ export function resetHooks(): void {
...
   didScheduleRenderPhaseUpdate = false;
   renderPhaseUpdates = null;
   numberOfReRenders = -1;
 }
 
 function createHook(): Hook {
-  return {
-    memoizedState: null,
+  let hook: Hook = __DEV__
+    ? {
```
gaearon: For things like this it's nice to declare two types, `Hook` and `HookDev`. Then when you need a DEV-only field, do `((hook: any): HookDev).something = ...`

Example of this pattern: https://github.com/facebook/react/blob/8c1614a2fdbb841f65ebf403d7ece3a518c6a984/packages/react-dom/src/server/ReactPartialRenderer.js#L697-L699

threepointone:
I tweaked the error message, added an error stack and updated the PR, but need to grab a quick bite. will get a screenshot and stuff when I'm back.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -414,6 +415,16 @@ function cloneHook(hook: Hook): Hook {
   if (__DEV__) {
     // anything faster than this?
     currentHookMatches.push([currentHookType, hook._debugType]);
```
gaearon: You could use a single array with even/odd values.

threepointone:
I removed currentMatches to track hooks, and use the fiber's hooks lists for the warning now, so it's a bit cleaner. This should be good to unblock the sync, and I'll work on a possible diff tree/custom hooks in the next PR.

packages/react-reconciler/src/ReactFiberHooks.js
threepointone: I'll do this.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -389,7 +481,10 @@ export function useContext<T>(
   observedBits: void | number | boolean,
 ): T {
   if (__DEV__) {
-    currentHookNameInDev = 'useContext';
+    currentHookNameInDev = HookDevNames[ContextHook];
+    currentHookType = ContextHook;
+    createWorkInProgressHook();
+    currentHookType = null;
   }
```
threepointone: ^ calling out this dev vs prod difference. I think it should be ok?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -327,6 +404,21 @@ function cloneHook(hook: Hook): Hook {
...
+    if (currentHookType !== hook._debugType) {
+      currentHookMismatch =
+        currentHookMismatch ||
+        new Error('tracer').stack
+          // the first few lines are always 'react' code.
+          // todo - maybe show a nice tree of hooks?
+          .split('\n')
+          .slice(3, 10)
+          .join('\n');
+    }
```
threepointone: I'll add a screenshot of what this looks like.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -170,6 +200,39 @@ function areHookInputsEqual(
   return true;
 }
 
+function flushHookMismatchWarnings() {
```
threepointone: flushMismatchWarning?

threepointone:
I made a version where it would show a side by side diff of hook trees using debug-tools (incl. custom hooks), but it turned out to be too invasive and might be more trouble than it's worth (and you have to already have broken the linter by this point.) so pulled that out, rewrote this diff to show the side by side thing for low level hooks. I'll adjust a screenshot soonly.

threepointone:
How's this - 
screenshot 2019-01-21 at 18 32 48

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
```diff
@@ -651,11 +654,15 @@ describe('ReactHooks', () => {
     expect(() => {
       expect(() => root.update(<App flip={true} />)).toThrow('custom error');
     }).toWarnDev(
-    // TODO - this should log just once
-    // it logs it twice because of the replay logic in the scheduler I think?
-    // must fix in some way, but pretty edge casey so letting this for now  
+      // TODO - this should log just once
+      // it logs it twice because of the replay logic in the scheduler I think?
+      // must fix in some way, but pretty edge casey so letting this for now
```
gaearon: Why edge casey? Does this happen on every error?

gaearon: The easiest thing we could do is dedupe by message. Keep a DEV-only set of diffs where you emit the warning. Don't emit the same diff twice.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -299,34 +401,67 @@ export function resetHooks(): void {
...
+        memoizedState: hook.memoizedState,
+
+        baseState: hook.baseState,
+        queue: hook.queue,
+        baseUpdate: hook.baseUpdate,
+
+        next: null,
+      };
+
+  if (__DEV__ && !currentHookMismatch) {
```
gaearon: Nit: we try to leave `if (__DEV__)` as its own condition for it to jump out — and nest `if`s

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -170,6 +205,71 @@ function areHookInputsEqual(
...
+
+function flushHookMismatchWarnings() {
+  // we'll show the diff of the low level hooks,
+  // and a stack trace so the dev can locate where
+  // the first mismatch is coming from
+  if (__DEV__ && currentHookMismatch) {
+    const hookStackDiff = [];
+    let current = firstCurrentHook;
+    let previousOrder = [];
+    while (current) {
```
gaearon: Nit: for Fibers etc we use explicit `null` comparisons

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -170,6 +205,71 @@ function areHookInputsEqual(
...
+      previousOrder.push(HookDevNames[((current: any): HookDev)._debugType]);
+      current = current.next;
+    }
+    let workInProgress = firstWorkInProgressHook;
+    let nextOrder = [];
+    while (workInProgress) {
```
gaearon: Same nit

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -170,6 +205,71 @@ function areHookInputsEqual(
...
+      nextOrder.push(HookDevNames[((workInProgress: any): HookDev)._debugType]);
+      workInProgress = workInProgress.next;
+    }
+    // some bookkeeping for formatting the output table
+    const columnLength = Math.max(
+      '  Previous render'.length,
+      ...previousOrder.map(hook => hook.length),
```
gaearon: Nit: we try to avoid features that generated Babel output with helpers (even in DEV). Let's use `apply`?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -54,6 +54,37 @@ type UpdateQueue<S, A> = {
...
+const ContextHook = 2;
+const RefHook = 3;
+const EffectHook = 4;
+const LayoutEffectHook = 5;
+const CallbackHook = 6;
+const MemoHook = 7;
+const ImperativeHandleHook = 8;
+const DebugValueHook = 9;
+
+const HookDevNames = [
```
gaearon: I'd write these as switch so that the 1:1 correspondence is clear, and also so that we can (type) check if is exhaustive, `HookType -> string`

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -170,6 +205,71 @@ function areHookInputsEqual(
...
+    warning(
+      false,
+      'React has detected a change in the order of hooks called by %s.\n' +
+        '%s\n' +
+        '%s\n\n' +
+        'Error trace:\n' +
+        '%s\n\n' +
+        'This will lead to bugs and errors if not fixed. ' +
+        'For more information, read the rules of hooks: https://fb.me/rules-of-hooks\n',
+      getComponentName(((currentlyRenderingFiber: any): Fiber).type),
```
gaearon: Nit: this is a bit hidden away. Let's read it very early so it's clearly visible this function assumes the variable is set.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -170,6 +205,71 @@ function areHookInputsEqual(
...
+    const hookStackLength = Math.max(previousOrder.length, nextOrder.length);
+    for (let i = 0; i < hookStackLength; i++) {
+      hookStackDiff.push(
+        (previousOrder[i] !== nextOrder[i] ? '> ' : '  ') +
+          padEndSpaces(previousOrder[i], columnLength) +
+          ' ' +
+          nextOrder[i],
+      );
+    }
+    warning(
```
gaearon: Let's deduplicate the warning here.

gaearon:
Proposal for output format:

```js
function useCustom() {
  useEffect()
}

function App() {
  useState();
  useEffect();
  if (foo) {
    useCustom()
  }
  useReducer()
  useState()
}
```

```
Previous render      Next render
u

**Comment by gaearon** [comment]
Proposal for output format:

```js
function useCustom() {
  useState()
  useEffect()
}

function useCustom2() {
  useState()
}

function App() {
  useState();
  useEffect();
  if (foo) {
    useCustom1()
  }
  useCustom2()
  useReducer()
  useState()
}
```

```
Previous render      Next render
--------------------------------
1. useState             useState
2. useEffect            useEffect
3. useState             useState
4. useReducer           useEffect
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
The first Hook type mismatch occurred:
  at useEffect (Foo:xx)
  at useCustom2 (Foo:xx)
  at App (Foo:xx)

This error occurred in the following component:
  in App
```

threepointone:
done screenshot 2019-01-21 at 18 32 48

threepointone:
done 
screenshot 2019-01-21 at 20 52 22

packages/react-reconciler/src/ReactFiberHooks.js
threepointone: ugh I missed this, will get to in a bit

packages/react-reconciler/src/ReactFiberHooks.js
threepointone: missed this

packages/react-reconciler/src/ReactFiberHooks.js
threepointone: I thought about this, and maybe I don't think this is critical? we're not planning on changing this list anytime soon,  and it'll add complexity.

gaearon: Not saying this is critical but I'm also not sure how this will add complexity.

packages/react-reconciler/src/ReactFiberHooks.js
gaearon: What does this do and why?

threepointone: Right now, we don't add a ContextType Hook to the hooks list on a fiber, we just do a straight  context read. this means you can't trace it's previous/current positions when rendering. this above bit adds it the list (via `createWorkInProgressHook`) so it can track it when it gets called again.

packages/react-reconciler/src/ReactFiberHooks.js
threepointone: didn't want to add extra function calls and the comparisons when it was readily indexed and available here.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -54,6 +54,39 @@ type UpdateQueue<S, A> = {
...
+  'useReducer',
+  'useContext',
+  'useRef',
+  'useEffect',
+  'useLayoutEffect',
+  'useCallback',
+  'useMemo',
+  'useImperativeHandle',
+  'useDebugValue',
+];
```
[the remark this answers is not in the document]
threepointone: you ok with the string comparisons happening on every run?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -54,6 +54,39 @@ type UpdateQueue<S, A> = {
...
+
+let currentHookType: HookType | null = null;
+// the first instance of a hook mismatch in a component,
+// represented by a portion of it's stacktrace
+let currentHookMismatch = null;
+const PossiblyWeakSet = typeof WeakSet === 'function' ? WeakSet : Set;
```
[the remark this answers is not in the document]
threepointone: the set `hookMismatchFibers` is for deduping error messages on fibers, so we show a max of one warning per fiber, persisted across app lifetime. it never flushes itself out. I'll add a comment 

`currentHookMismatch` holds the last mismatch recorded, and gets flushed.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -170,6 +207,92 @@ function areHookInputsEqual(
   return true;
 }
 
+// till we have String::padEnd, a small function to
+// right-pad strings with spaces till a minimum length
+function padEndSpaces(string: string, length: number) {
```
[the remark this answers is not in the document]
threepointone: you mean the contents of the fn, right? `function padEndSpaces(){ if (__DEV__){...}}`

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
acdlite: Ah right. So usually the way we handle this is we de-dupe by component display name. That way if you have a list of components all with the same bug, you’ll only see a single warning. Refer to ReactFiberClassComponent.

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
acdlite: Yeah it’s dev and the strings will be interned anyway

threepointone:
alright, addressed all the feedback, and also managed to remove `currentHookType`, just using `currentHookNameInDev`. It does feel a bit nicer now.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -170,6 +196,95 @@ function areHookInputsEqual(
...
+    }
+  }
+}
+
+function flushHookMismatchWarnings() {
+  // we'll show the diff of the low level hooks,
+  // and a stack trace so the dev can locate where
+  // the first mismatch is coming from
+  if (__DEV__) {
+    if (currentHookMismatch) {
```
acdlite: Just noticed this: for future reference, we try to always explicitly compare to null or undefined. That way we distinguish between intentional and unintentional missing values, and comparisons are slightly faster.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -54,6 +54,28 @@ type UpdateQueue<S, A> = {
   eagerState: S | null,
 };
 
+opaque type HookType = string;
+
+const StateHook: HookType = 'useState';
```
acdlite: Don’t need these variables, just inline the strings

PR #14588, opened 2019-01-14
make a fork for ReactCurrentDispatcher

threepointone:
we need to 'fork' ReactCurrentDispatcher just like we did for ReactCurrentOwner. I'm testing this right now, putting the PR out there to make sure I copy-pasted right.


threepointone:
tests/types pass! woo.

Issue #14583, opened 2019-01-13
16.8.0-alpha.0 (and 16.7) IE11 Suspense doesn't stop rendering fallback after Lazy resolves


Kazimirkas:
**Do you want to request a *feature* or report a *bug*?**

report a bug

**What is the current behavior?**

When using Suspense, I see React.Suspense fallback layout instead of loaded component in IE 11 (IE network panel shows, that bundle was loaded)

I can not reproduce this bug in 16.7.0-alpha.2! there is everything ok.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**
```
const UserCabinet = React.lazy(() => import('PublicUser/components/UserCabinet/UserCabinet'));

function WaitingComponent(Component) {
  return props => (
    <React.Suspense fallback={<div>Loading...</div>}>
      <Component {...props} />
    </React.Suspense>
  );
}

class PublicRoutes extends React.Component<{}> {
  render() {
    return (
      <Switch>
        <PublicPage path="/login" component={AuthPage} />
        <PublicPage path="/signUp" component={SignUpPage} />
        <PrivatePage path="/cabinet" component={WaitingComponent(UserCabinet)} />
        <Redirect to="/login" />
      </Switch>
    );
  }
}
```
**What is the expected behavior?**
see the rendered component

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

Tested on IE11 on windows 7
It reproduces only in 16.8.0-alpha.0
I can not reproduce this bug in 16.7.0-alpha.2, there is everything ok.
In Chrome is everything ok in all versions.

Kazimirkas:
the problem was with requirements for additional polyfills in newer version

gaearon:
Hmm which ones? I wouldn't expect that.

gaearon:
Could be the same as https://github.com/facebook/react/issues/14570. A reproducing example would be very helpful; that's unexpected.

threepointone:
@Kazimirkas can you try running this codesandbox and see if it runs on IE11? https://codesandbox.io/s/oo3rn9z8wq if not on codesandbox, try to run it locally. thanks!

aweary:
[...]

PR #14592, opened 2019-01-15
Avoid new Set([iterable]) for thenables

aweary:
Fixes https://github.com/facebook/react/issues/14583

I wrote and [deployed](http://miniature-song.surge.sh/) [a small test app](https://gist.github.com/aweary/e5d0e9c354ab3ee34bc67440d899702f) and verified that this fixes the issue outlined in #14583 in IE11 on Windows 8.


gaearon:
Hmm. Now that I merged it, I realized ideally you wouldn't read field twice. Create a set, add the item, then set `workInProgress.updateQueue`. Mind tweaking?

PR #14594, opened 2019-01-15
Don't bother comparing constructor when deps are not provided

acdlite:
When no dependencies are passed to an effect hook, what we used to do is compare the effect constructor. If there was no change, then we would skip firing the effect. In practice, this is a useless optimization because the constructor will always be different when you pass an inline closure. And if you don't pass an inline closure, then you can't access any props or state.

There are some edge cases where an effect that doesn't close over props or state could be useful, like reference counting the number of mounted components. But those are rare and can be addressed by passing an empty array of dependencies.

By removing this "optimization," we can avoid retaining the constructor in the majority of cases where it's a closure that changes on every render.

I made corresponding changes to the other hooks that accept dependencies, too (useMemo, useCallback, and useImperativeHandle).


packages/shared/areHookInputsEqual.js
```diff
@@ -14,15 +14,17 @@ export default function areHookInputsEqual(arr1: any[], arr2: any[]) {
...
-    warning(
-      arr1.length === arr2.length,
-      'Detected a variable number of hook dependencies. The length of the ' +
-        'dependencies array should be constant between renders.\n\n' +
-        'Previous: %s\n' +
-        'Incoming: %s',
-      arr1.join(', '),
-      arr2.join(', '),
-    );
+    if (arr1 === null || arr2 === null || arr1.length !== arr2.length) {
```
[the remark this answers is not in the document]
acdlite: Yeah I figured that was fine? Idk

packages/shared/areHookInputsEqual.js
```diff
@@ -14,15 +14,17 @@ export default function areHookInputsEqual(arr1: any[], arr2: any[]) {
...
+        'Detected a variable number of hook dependencies. The length of the ' +
+          'dependencies array should be constant between renders.\n\n' +
+          'Previous: %s\n' +
+          'Incoming: %s',
+        arr1 === null ? '(empty)' : `[${arr1.join(', ')}]`,
+        arr2 === null ? '(empty)' : `[${arr2.join(', ')}]`,
+      );
+    }
   }
   for (let i = 0; i < arr1.length; i++) {
```
sebmarkbage: Somewhere around here I'm betting the VM will insert a type check anyway. I bet we can get away with just doing a proper null check on the one we're not sure about. That way you don't have to cheat the type system.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -599,45 +599,45 @@ export function useDebugValue(
...
-  const nextInputs =
-    inputs !== undefined && inputs !== null ? inputs : [nextCreate];
+  const nextDeps = deps === undefined ? null : deps;
 
   const prevState = workInProgressHook.memoizedState;
   if (prevState !== null) {
-    const prevInputs = prevState[1];
-    if (areHookInputsEqual(nextInputs, prevInputs)) {
+    // Assume these are defined. If they're not, areHookInputsEqual will warn.
+    const prevDeps = prevState[1];
```
sebmarkbage: This property access should move within the conditional so we only do it if we need it.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -516,34 +516,36 @@ export function useRef<T>(initialValue: T): {current: T} {
...
-  let nextInputs = inputs !== undefined && inputs !== null ? inputs : [create];
+  const nextDeps = deps === undefined ? null : deps;
   let destroy = null;
   if (currentHook !== null) {
     const prevEffect = currentHook.memoizedState;
     destroy = prevEffect.destroy;
-    if (areHookInputsEqual(nextInputs, prevEffect.inputs)) {
-      pushEffect(NoHookEffect, create, destroy, nextInputs);
+    // Assume these are defined. If they're not, areHookInputsEqual will warn.
+    const prevDeps: Array<mixed> = (prevEffect.deps: any);
```
sebmarkbage: Same thing here. This should be only if `nextDeps !== null`.

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -57,6 +60,48 @@ function resolveCurrentlyRenderingComponent(): Object {
   return currentlyRenderingComponent;
 }
 
+function areHookInputsEqual(
```
threepointone: should this have been `areHookDepsEqual`?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -125,6 +128,48 @@ function resolveCurrentlyRenderingFiber(): Fiber {
...
+          'order and size of this array must remain constant.\n\n' +
+          'Previous: %s\n' +
+          'Incoming: %s',
+        currentHookNameInDev,
+        `[${nextDeps.join(', ')}]`,
+        `[${prevDeps.join(', ')}]`,
+      );
+    }
+  }
+  for (let i = 0; i < prevDeps.length && i < nextDeps.length; i++) {
```
Andarist: according to the previous check - those lengths should be the same, so shouldnt this iteration just assume that fact?

PR #14596, opened 2019-01-15
suggest useEffect when useLayoutEffect triggers in server rendering

threepointone:
Working on a separate PR for reactjs.org that this error will point to, started this PR to get approval on the copy. Works for everyone?


packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -281,7 +281,9 @@ export function useLayoutEffect(
       "be encoded into the server renderer's output format. This will lead " +
       'to a mismatch between the initial, non-hydrated UI and the intended ' +
       'UI. To avoid this, useLayoutEffect should only be used in ' +
```
gaearon: I think grammar gets a bit clunky. It's almost a run-on sentence.

threepointone: damn, thanks for that. love it.

threepointone: done

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -344,9 +344,12 @@ export function useLayoutEffect(
...
       "be encoded into the server renderer's output format. This will lead " +
-      'to a mismatch between the initial, non-hydrated UI and the intended ' +
-      'UI. To avoid this, useLayoutEffect should only be used in ' +
-      'components that render exclusively on the client.',
+      'to a mismatch between the initial, non-hydrated UI, and the intended ' +
+      'UI. If possible, replace useLayoutEffect with useEffect which ' +
+      "doesn't block the first paint. However, if this effect must run " +
+      'before the user sees anything, you can change this component to only ' +
+      'render on the client. To learn more, see ' +
+      'https://fb.me/react-warning-server-uselayouteffect',
```
sophiebits: this reads as very jargony to me… my suggestion:

```
      'useLayoutEffect is not supported during server rendering. This call ' +
        'is a no-op, but it likely indicates a bug because the effect will ' +
        'not run until your component is hydrated on the client. If this ' +
        'effect can be safely run after the component appears, replace ' +
        'useLayoutEffect with useEffect. If not, you can change this ' +
        'component to only render on the client. See ' +
        'https://fb.me/react-warning-server-uselayouteffect',
```

dimensi:
Hi, I do not understand the situation with this hook a bit. I use this hook to perform the animation synchronously with the state update, if I use useEffect, then I have jumps in the animation, because the animation library does not have time to start. Also, the documentation states that useLayoutEffect runs on the same phase as componentDidMount (that is, on the client side), and here my server issues complaints to me about my code. Why is that?
@gaearon, @sophiebits

gaearon:
Please ask a question in an issue, not a PR. When PR gets merged or closed, the discussion is hard to find — even if an issue is still active.

gaearon:
(You'll also need to provide a small runnable CodeSandbox example in your issue to better illustrate your problem.)

dimensi:
@gaearon ok

threepointone:
revis

threepointone:
closing this

PR #14599, opened 2019-01-15
Added Flow type to keep hooks dispatchers in-sync

bvaughn:
Follow up to #14597

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -641,20 +641,6 @@ export function useMemo<T>(
   return nextValue;
 }
 
-// Single source of truth that keeps various hooks APIs in-sync.
-export type DispatcherHooks = {
-  useCallback: typeof useCallback,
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -641,6 +641,20 @@ export function useMemo<T>(
   return nextValue;
 }
 
+// Single source of truth that keeps various hooks APIs in-sync.
+export type DispatcherHooks = {
```
bvaughn: Wasn't sure what to call this thing, since it's just the hooks portion of the dispatcher. Any better suggestions?

bvaughn:
Hm...maybe, although the ref also has the `Dispatcher` type.

My intended purpose of this new type was to prevent us from _forgetting_ to add hooks to one of these (as I did with the `useDebugValue` hook). I think the current check covers that. I'm not sure how adding one to the internals file would be able to do the same?

bvaughn:
How's this now, Dan?

PR #14608, opened 2019-01-16
Throw an error when using hooks inside useMemo, or .memo's comparison function

threepointone:
This PR adds a flag to 'pause' hooks while they're running, so we can throw an error whenever a hook is called inside 2 spots - useMemo's nextCreate function, and React.memo's comparison function. 
It's a bit icky, but wanted to confirm the approach with you. 

- need to add a test for verifying you can't useContext in the same scenarios 
- Need to add the same logic to the PartialRenderer

I could also rename pauseHooks to something more explicit, like pauseHooksWhileComparingMemos


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -342,13 +347,16 @@ function updateMemoComponent(
     // Default to shallow comparison
     let compare = Component.compare;
     compare = compare !== null ? compare : shallowEqual;
+    pauseHooks(true);
```
gaearon: I think the test passes without these calls. Memo doesn't call "prepare to use hooks" so we're not in an area where there's a current fiber. You should be able to just remove them.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -626,8 +633,9 @@ export function useMemo<T>(
       return prevState[0];
     }
   }
-
+  pauseHooks(true)
```
gaearon: Since we don't need `pauseHooks` elsewhere this can become local `isInUseMemo = true`.

packages/react-reconciler/src/ReactFiberBeginWork.js
threepointone: oh interesting, you're right they pass. I recollect how I broke this - I had a bad test, and 'fixed' it after I had my code in place. bleurgh, thanks for the catch.

packages/react-reconciler/src/ReactFiberHooks.js
threepointone: agreed, that was my first cut as well.

gaearon:
What about `useCallback` that's called during render phase? Sounds like we'd want to disable it there too?

gaearon:
Instead of adding a second check, let's just null out `currentlyRenderingFiber` right before the call into user code and then restore it?

threepointone:
> What about `useCallback` that's called during render phase? Sounds like we'd want to disable it there too?

the function itself isn't called during render, no? so I assume it should throw, but I'll add a test to verify.

threepointone:
> Instead of adding a second check, let's just null out `currentlyRenderingFiber` right before the call into user code and then restore it?

I'm happy to do this. either approach is similarly annoying tbh ha.

gaearon:
>I'm happy to do this. either approach is similarly annoying tbh ha.

I don't think they're exactly equivalent. This is a very hot path. Having one variable and doing just one check per Hook seems better than doing two. With this solution, only `useEffect` and the like "pay" with extra assignments. But you're not penalized for having many `useState`s which would each has to check `isInUseMemo` "just in case".

The downside is that it's an extra local variable (to hold the current Fiber) and Sebastian says this can go beyond the number of registers and cause a perf cliff. I don't know about this. I'd probably go with one extra variable though for less common hooks than an extra check for every hook.

---

>the function itself isn't called during render, no? so I assume it should throw, but I'll add a test to verify.

Ahh I see. I meant this is a valid code:

```js
const renderRow = useCallback((item) => {
  return <Row item={item} theme={theme} />
}, [theme])

<Table renderRow={renderRow} />

// in Table
rows.map(props.renderRow)
```

So callback itself _can_ execute during render phase and it could be nice if we could disallow calling Hooks in it. (They would be attached to the Table which is super confusing.) Reading context there is probably also bad because it would read Table's context rather than Row's.

But you're right we're not actually proxying the function so we can't add extra logic to it. Theoretically we *could* in DEV though. But we can talk about this another time.

threepointone:
Curious, have you ever got to a situation where comparisons are that expensive?

gaearon:
In React everything that executes every render is expensive. It's the hottest path in the application.

gaearon:
(In isolation, no, but these things add up and at some point you end up with too many — not knowing which ones are safe to remove. So we try to not add them whenever possible.)

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -108,7 +108,6 @@ let renderPhaseUpdates: Map<UpdateQueue<any>, Update<any>> | null = null;
 // Counter to prevent infinite loops.
 let numberOfReRenders: number = 0;
 const RE_RENDER_LIMIT = 25;
-
```
threepointone: revert

threepointone:
I made changes. I avoided the 'local' var check with yet another module-local var. 

I'll add tests for the server renderer now.

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -229,11 +229,14 @@ export function useReducer<S, A>(
   }
 }
 
+// a variable to store current component while we compare memo values
+let cachedCurrentlyRenderingComponent = null;
```
gaearon: Why is this top level rather than inside `useMemo`?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -616,11 +608,14 @@ export function useCallback<T>(
   return callback;
 }
 
+// a variable to store current component while we compare memo values
+let cachedCurrentlyRenderingFiber = null;
```
gaearon: Same q.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -633,9 +628,14 @@ export function useMemo<T>(
       return prevState[0];
     }
   }
-  pauseHooks(true)
+  
+  // null the reference to the component
```
gaearon: This comment is unnecessary, the line below says the same thing.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -633,9 +628,14 @@ export function useMemo<T>(
...
+  currentlyRenderingFiber = null;
+  // now, it'll throw if you try to call a hook inside nextCreate
   const nextValue = nextCreate();
-  pauseHooks(false)
+  // restore the current fiber
```
gaearon: Same

packages/react-dom/src/server/ReactPartialRendererHooks.js
threepointone: it's one less var to be gced, reuses the location across runs. you mentioned the local var has a cost I assumed you'd want to avoid?

threepointone: maybe I should just trust javascript engines

gaearon: Declaring variables pointing to shared objects has no effect on GC — I assume you mean the stack? I said it's a hypothetical concern but I'm not actually sure it matters and I think local function variable is fine.

aweary: I wish we had some way to validate these JS engine assumptions, it’d be interesting to see how often they’re true 🤔

gaearon: Well, having two checks is likely going to be more expensive than having one check, no matter what you do.

threepointone:
I've updated the PR addressing most of the comments, including adding tests for the server side stuff. one last bit is to prevent `dispatcher.readContext` in these scenarios, and I can't figure a clean way of doing this just yet, so I'll send it in a follow up PR.

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -196,7 +196,12 @@ export function useReducer<S, A>(
           // priority because it will always be the same as the current
           // render's.
           const action = update.action;
+
+          currentlyRenderingComponent = null;
+          // now, it'll throw if you try to call a hook inside the reducer
           newState = reducer(newState, action);
+          // restore the current component
+          currentlyRenderingComponent = cachedCurrentlyRenderingComponent;
           update = update.next;
```
gaearon: Nit: I would replace a comment structure like

```js
    currentlyRenderingComponent = null;
    // now, it'll throw if you try to call a hook inside the reducer
    newState = reducer(newState, action);
    // restore the current component
    currentlyRenderingComponent = cachedCurrentlyRenderingComponent;
```

with something like

```js
    // Temporarily clear to forbid calling Hooks.
    currentlyRenderingComponent = null;
    newState = reducer(newState, action);
    currentlyRenderingComponent = cachedCurrentlyRenderingComponent;
```

The pattern of replacing something and then setting again is very common in the codebase so we don't need to over-explain it — just enough to hint at the purpose when we clear.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -608,14 +616,11 @@ export function useCallback<T>(
...
 
-// a variable to store current component while we compare memo values
-let cachedCurrentlyRenderingFiber = null;
-
 export function useMemo<T>(
   nextCreate: () => T,
   inputs: Array<mixed> | void | null,
 ): T {
-  cachedCurrentlyRenderingFiber = currentlyRenderingFiber = resolveCurrentlyRenderingFiber();
+  let cachedCurrentlyRenderingFiber = (currentlyRenderingFiber = resolveCurrentlyRenderingFiber());
```
gaearon: Nit: Maybe just "fiber" everywhere? It's local so doesn't need too much introducing.

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
```diff
@@ -17,6 +17,9 @@ let ReactFeatureFlags;
 let ReactTestRenderer;
 let ReactDOMServer;
 
+const OutsideHookScopeError =
+  'Hooks can only be called inside the body of a function component';
```
gaearon: Nit: just copy paste it

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
```diff
@@ -122,27 +125,43 @@ describe('ReactHooks', () => {
...
+
+    function Simple() {
+      useState(0);
+      return 123;
+    }
+    let root = ReactTestRenderer.create(null);
+    expect(() => root.update(<App />)).toThrow(OutsideHookScopeError);
+    // we want to assure that no hook machinery has broken,
+    // so we listen for all errors on the next line
+    expect(() => root.update(<Simple />)).not.toThrow();
```
gaearon: If you want to ensure everything still works, worth asserting the output is legit.

packages/react-reconciler/src/ReactFiberHooks.js
gaearon: Nit: we accidentally lost this line

packages/react-reconciler/src/ReactFiberHooks.js
threepointone: didn't want to encourage reading from it, but I guess that's actually fine

packages/react-dom/src/server/ReactPartialRendererHooks.js
gaearon: It also helps to avoid newlines in between so that the starting and ending line of replacement are as close to each other as possible.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -448,7 +454,7 @@ export function useReducer<S, A>(
     const dispatch: Dispatch<A> = (queue.dispatch: any);
     return [workInProgressHook.memoizedState, dispatch];
   }
-
```
threepointone: revert

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -249,8 +254,10 @@ function useMemo<T>(
       return prevState[0];
     }
   }
-
```
threepointone: revert

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -626,8 +633,10 @@ export function useMemo<T>(
       return prevState[0];
     }
   }
-
```
threepointone: revert

threepointone:
alright, I'm happy with this, will land after ci finishes. thanks everyone!

threepointone:
@acdlite do you want to have a onceover, or should I dismiss your review? plsthx

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -512,6 +515,7 @@ export function useReducer<S, A>(
             const action = update.action;
             newState = reducer(newState, action);
           }
+          currentlyRenderingFiber = fiber;
```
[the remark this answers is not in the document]
threepointone: 😔yeah. sorry about that.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -509,8 +512,10 @@ export function useReducer<S, A>(
             // current reducer, we can use the eagerly computed state.
             newState = ((update.eagerState: any): S);
           } else {
+            currentlyRenderingFiber = null;
```
gaearon: Let's move this a line down and give it the usual comment.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -540,6 +545,7 @@ export function useReducer<S, A>(
     return [workInProgressHook.memoizedState, dispatch];
   }
 
+  currentlyRenderingFiber = null;
```
gaearon: Let's give it a comment for consistency.

threepointone:
I'll send a separate PR for guarding eagerReducer, I want to add a fresh test for that.

threepointone:
#fileatask ;(

PR #14609, opened 2019-01-16
Add ESLint rule playground

gaearon:
I found it rather annoying to iterate on our lint rules. Www/Nuclide has been too slow, and tests, while nice, make it too cumbersome to experiment with one particular use case, or just explore how the rule "feels" in practice.

This adds a "playground" fixture that lets you edit a file in an editor like VSCode and have any changes to rule source be reflected immediately on recheck. I had to make a hacky "proxy" rule for that to work because Node would cache the actual rule normally. The resulting workflow is pretty nice and I want to check this in — it'll definitely be useful to me but maybe to someone else too.

Obligatory gif:

Demo


bvaughn:
I don't think this seems to be working  for me. Maybe I'm misunderstanding something?

I've ran `yarn && yarn start`:
screen shot 2019-01-16 at 3 39 42 pm

When I modify the `RulesOfHooks` file– this watch re-runs successfully.

But I've opened the `fixtures/eslint/index.js` file and tried editing it– and nothing happens. No warnings show.

I have VS Code with the ESLint extension installed:
screen shot 2019-01-16 at 3 40 42 pm

gaearon:
Did you open `react` folder in VSCode or `react/fixtures/eslint`? It seems like it doesn't work unless you open the deep folder as project root. I'll add to README.

bvaughn:
Ah, that would be it then. I assumed you could open the full react project (since you would also be editing the rule).

gaearon:
I don't like Code so I'll be editing it in Sublime personally. :-)

But I think you can React as a folder to workspace and then it works.

screen shot 2019-01-16 at 11 50 40 pm

bvaughn:
> I don't like Code so I'll be editing it in Sublime personally. :-)

I'm also a Sublime user. I only fired up and tested Code because of your readme:
> 2. "File > Add Folder to Workspace" this specific folder in VSCode with ESLint extension

gaearon:
I figured most our users are on Code so we should probably test in that in general. Autofix support is nice too.

PR #14613, opened 2019-01-17
Fix shallow renderer set instance state after gDSFP before calling sCU

### LINKED ISSUE #14607: Bug: in shallow renderer `this.state` in `shouldComponentUpdate` will be updated by `getDeriveStateFromProps`
Bug

In `react-test-renderer/shallow`, the shallow renderer will set instance state in `getDerviedStateFromProps` so the `this.state` and `nextState` will be the same in `shouldComponentUpdate`, which is different with the behavior of real rendering.

given below class component:

```js

class SimpleComponent extends React.Component {
      constructor(props) {
        super(props);
        this.state = {
          value: props.value,
        };
      }

      static getDerivedStateFromProps(nextProps, prevState) {
        if (nextProps.value === prevState.value) {
          return null;
        }
        return { value: nextProps.value };
      }

      shouldComponentUpdate(nextProps, nextState) {
        console.log('shouldUpdate:', nextState.value !== this.state.value)
        return nextState.value !== this.state.value;
      }

      render() {
        return <div className={this.state.value} />;
      }
    }

```

render with ReactDOM, the `shouldComponentUpdate` will return true since `this.state` will be the old value `initial`:

```js
const divRef = React.createRef()
const div = document.createElement('div')

// in real test code we have to add a `ref={divRef}` in the `div` of SimpleComponent

const initialResult = ReactDOM.render(<SimpleComponent value="initial" />, div);
expect(divRef.current.classList[0]).toEqual("initial");

// will output `shouldUpdate: true`
const updatedResult = ReactDOM.render(<SimpleComponent value="updated" />, div);
// so the class name changes
expect(divRef.current.classList[0]).toEqual("updated");
```

In shallow renderer, the behavior is different:

```js
const initialResult = shallowRenderer.render(<SimpleComponent value="initial" />);
expect(initialResult).toEqual(<div>value:initial</div>);
// will not update, since in `shouldComponentUpdate` the `this.state` has been updated
// after `getDeriveStateFromProps` and as same as passed `nextState`
const updatedResult = shallowRenderer.render(<SimpleComponent value="updated" />);
// the following assert will fail
expect(updatedResult).toEqual(<div>value:updated</div>);
```

ShallowRenderer should not set the instance state in getDeriveStateFromProps, or in the `shouldComponentUpdate` we have no way to compare it.

I have tested this in master branch and 16.3, both not work.

By the way this problem is related to airbnb/enzyme#1970 . If it's recoginzed as a bug I can provide a PR in next several days.

chenesan:
This PR fixes #14607 which is related to airbnb/enzyme#1970 . Before this the shallow renderer will update instance state before shouldComponentUpdate if getDeriveStateFromProps return new partial state, make the passed `nextState` and old `this.state` the same in shouldComponentUpdate, which is different behavior with real rendering with ReactDOM.render (see the tests).


gaearon:
I think `_updateStateFromStaticLifecycle` might have been creating more confusion than what it's worth due to timing issues of what writes when. So I unwound it. I think this makes sense.

PR #14615, opened 2019-01-17
Warn when mixing createRoot() and old APIs

gaearon:
It's confusing when you mix those — e.g. if you render with `createRoot().render()` but then call `unmountComponentAtNode()`. Sometimes it warns with misleading warnings, and sometimes (in particular, if there's existing content) it didn't warn at all. We don't have guarantees around how it works either (e.g. unmounting doesn't work).

I added an explicit DEV-only flag for whether something is a new-style root, and I warn whenever you try to mix the API one way or the other. The message suggests the right API to use. I didn't change the behavior.

I didn't clean up false positive further warnings because the code for them is all over the place and it's hard to untangle the assumptions. So warning at entry point seemed like the easiest thing to do. We want to keep these worlds separate anyway.

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -160,9 +160,11 @@ setRestoreImplementation(restoreControlledState);
 export type DOMContainer =
   | (Element & {
       _reactRootContainer: ?Root,
+      _reactIsNewStyleRootDEV: ?boolean,
     })
   | (Document & {
       _reactRootContainer: ?Root,
+      _reactIsNewStyleRootDEV: ?boolean,
     });
 
 type Batch = FiberRootBatch & {
@@ -362,7 +364,7 @@ ReactWork.prototype._onCommit = function(): void {
 };
 
 function ReactRoot(
-  container: Container,
+  container: DOMContainer,
   isConcurrent: boolean,
   hydrate: boolean,
 ) {
@@ -652,6 +654,16 @@ const ReactDOM: Object = {
   },
 
   hydrate(element: React$Node, container: DOMContainer, callback: ?Function) {
+    if (__DEV__) {
+      warningWithoutStack(
+        !container._reactIsNewStyleRootDEV,
+        'You are calling ReactDOM.hydrate() on a container that was previously ' +
+          'managed by ReactDOM.%s(). This is not supported. ' +
+          'Did you mean to call root.render(element, {hydrate: true})?',
+        enableStableConcurrentModeAPIs ? 'createRoot' : 'unstable_createRoot',
+      );
+    }
+
     // TODO: throw or warn if we couldn't hydrate?
     return legacyRenderSubtreeIntoContainer(
       null,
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/client/ReactDOM.js
```diff
@@ -652,6 +647,19 @@ const ReactDOM: Object = {
...
   hydrate(element: React$Node, container: DOMContainer, callback: ?Function) {
+    invariant(
+      isValidContainer(container),
+      'Target container is not a DOM element.',
+    );
+    if (__DEV__) {
+      warningWithoutStack(
+        !container._reactIsNewStyleRootDEV,
+        'You are calling ReactDOM.hydrate() on a container that was previously ' +
+          'managed by ReactDOM.%s(). This is not supported. ' +
```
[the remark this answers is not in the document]
gaearon: "created by createRoot" sounded a bit tautological but I don't know if manage is better.

Container also wasn't technically "created" by createRoot. createRoot created the _root_ but it's the _container DOM node_ that we've previously passed to the other API. Not the root.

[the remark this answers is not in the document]
gaearon: I went with "passed to"

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -160,9 +159,11 @@ setRestoreImplementation(restoreControlledState);
 export type DOMContainer =
   | (Element & {
       _reactRootContainer: ?Root,
+      _reactHasBeenPassedToCreateRootDEV: ?boolean,
```
acdlite: Shouldn't need this because `_reactRootContainer` only exists on legacy roots.

gaearon: My primary goal was to detect that a container previously passed to `createRoot` is *then* passed to `unmountComponentAtNode`. I don't think I can do this without some kind of a marker on the *new* container. I can't check its first child because `createRoot` doesn't clear children (and so could mount at arbitrary point). Scanning all children seems not great.

gaearon:
I assume https://github.com/facebook/react/pull/14615#discussion_r makes sense but if it doesn't I can follow up.

packages/react-dom/src/client/ReactDOM.js
acdlite:
```js
const reactHasBeenPassedToCreateRoot = domContainer._reactRootContainer === null;
```

gaearon: But isn't that used by `ReactDOM.render` too? Since it's expressed via `createRoot` internally.

acdlite: `ReactDOM.render` uses the `ReactRoot` type, but it doesn't call `createRoot` directly.

The only place `_reactRootContainer` is set is in `legacyRenderSubtreeIntoContainer`. New roots don't need the expando because the root is retained in user space.

https://github.com/facebook/react/blob/b66e6e41e606745b30238689fa1a98e874b29e6e/packages/react-dom/src/client/ReactDOM.js#L556

gaearon: I think we're talking about two different cases here.

1. Detecting DOM container that's previously been used by `ReactDOM.render`. This is easy. That's what `_reactRootContainer` gives us.

2. Detecting DOM container that's previously been used by `ReactDOM.createRoot`. I don't see a way to do this currently. As you said, "new roots don't need the expando". That's why I couldn't figure out how to detect them, and added a new DEV field.

I need both checks. Not just one way around.

gaearon: > I suppose you're trying to distinguish between an empty DOM container that was never rendered by anything, and a DOM container that was rendered by createRoot, so you can provide a more helpful message.

Yep 👍

PR #14625, opened 2019-01-18
Warn when second callback is passed to setState/dispatch in Hooks

gaearon:
It's supported in classes and not in Hooks. People try to pass it and wonder why it doesn't work. Now we warn for that. This also lets us add it (or something else) later as a second argument.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -782,6 +782,15 @@ function dispatchAction<S, A>(
       'an infinite loop.',
   );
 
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -782,6 +782,15 @@ function dispatchAction<S, A>(
       'an infinite loop.',
   );
 
+  if (__DEV__) {
+    warning(
+      arguments.length <= 3,
```
gaearon: Could've been === 3 but I figured I don't want to punish users for our bugs.

PR #14626, opened 2019-01-18
Support sync thenables for lazy()

gaearon:
Why don't we? Currently they fail with a confusing error because of a race condition (status gets set to resolved but then the result gets overwritten by the next line). This should fix it.

I figured this might be useful for testing. See https://github.com/airbnb/enzyme/issues/1917#issuecomment-454208642. It's awkward that people look for workarounds like `waitForLazyLoaded` in a synchronous environment. Supporting sync thenables could be a nice solution to that.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberLazyComponent.js
+++ b/packages/react-reconciler/src/ReactFiberLazyComponent.js
@@ -47,6 +47,7 @@ export function readLazyComponentType<T>(lazyComponent: LazyComponent<T>): T {
       lazyComponent._status = Pending;
       const ctor = lazyComponent._ctor;
       const thenable = ctor();
+      lazyComponent._result = thenable;
       thenable.then(
         moduleObject => {
           if (lazyComponent._status === Pending) {
@@ -73,9 +74,9 @@ export function readLazyComponentType<T>(lazyComponent: LazyComponent<T>): T {
           }
         },
       );
-      // Don't race with a sync thenable
-      if (lazyComponent._status === Pending) {
-        lazyComponent._result = thenable;
+      // Check if it resolved synchronously
+      if (lazyComponent._status === Resolved) {
+        return lazyComponent._result;
       }
       throw thenable;
     }
--- a/packages/react-reconciler/src/ReactFiberLazyComponent.js
+++ b/packages/react-reconciler/src/ReactFiberLazyComponent.js
@@ -73,7 +73,10 @@ export function readLazyComponentType<T>(lazyComponent: LazyComponent<T>): T {
           }
         },
       );
-      lazyComponent._result = thenable;
+      // Don't race with a sync thenable
+      if (lazyComponent._status === Pending) {
+        lazyComponent._result = thenable;
+      }
       throw thenable;
     }
   }
--- a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
@@ -74,7 +74,7 @@ describe('ReactLazy', () => {
       </Suspense>,
     );
 
-    expect(ReactTestRenderer).toHaveYielded(['Loading...', 'Hi']);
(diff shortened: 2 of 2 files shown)


packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
```diff
@@ -61,6 +61,23 @@ describe('ReactLazy', () => {
...
+      },
+    }));
+
+    const root = ReactTestRenderer.create(
+      <Suspense fallback={<Text text="Loading..." />}>
+        <LazyText text="Hi" />
+      </Suspense>,
+    );
+
+    expect(ReactTestRenderer).toHaveYielded(['Loading...', 'Hi']);
```
acdlite:
```suggestion
    expect(ReactTestRenderer).toHaveYielded(['Hi']);
```

packages/react-reconciler/src/ReactFiberLazyComponent.js
```diff
@@ -73,7 +73,10 @@ export function readLazyComponentType<T>(lazyComponent: LazyComponent<T>): T {
           }
         },
       );
-      lazyComponent._result = thenable;
+      // Don't race with a sync thenable
+      if (lazyComponent._status === Pending) {
+        lazyComponent._result = thenable;
```
acdlite: Alt suggestion:

```js
// Check if it resolved synchronously
if (lazyComponent._status === Resolved) {
  return lazyComponent._result;
}
```

Since if it's already resolved we don't need to throw the thenable

gaearon: Aaaah. That's why it had the loading state. Thanks.

packages/react-reconciler/src/ReactFiberLazyComponent.js
```diff
@@ -47,6 +47,7 @@ export function readLazyComponentType<T>(lazyComponent: LazyComponent<T>): T {
       lazyComponent._status = Pending;
       const ctor = lazyComponent._ctor;
       const thenable = ctor();
+      lazyComponent._result = thenable;
```
gaearon: This is still needed because we read thenable from `lazyComponent._result` while it's pending. If we keep the assignment, it will overwrite sync thenables. But if we remove it at all, we will throw `null` instead of thenable. This fixes both cases.

packages/react-reconciler/src/ReactFiberLazyComponent.js
```diff
@@ -73,7 +74,10 @@ export function readLazyComponentType<T>(lazyComponent: LazyComponent<T>): T {
           }
         },
       );
-      lazyComponent._result = thenable;
+      // Check if it resolved synchronously
+      if (lazyComponent._status === Resolved) {
+        return lazyComponent._result;
+      }
```
gaearon: We could also do the same for failure. But doesn't seem worth the extra code to me. Could be a recursive call but then there's more risk of messing it up and going into a stack overflow. Meh.

gaearon:
I'll merge since this follows your suggestion.

packages/react-reconciler/src/ReactFiberLazyComponent.js
acdlite: You still need the assignment, but I don't think you need to move it up here. If you keep it right after the early return branch you added on line 78-80, then there's always a single assignment.

```js
if (lazyComponent._status === Resolved) {
  return lazyComponent._result;
}
lazyComponent._result = thenable;
throw thenable;
```

ntucker:
Does this make lazy() work in SSR?

gaearon:
No, `<Suspense>` is still unsupported. You’d also get a hydration mismatch if your server renders everything but the client lazy loads.

suanmei:
When will it be supported for ssr?

gaearon:
https://reactjs.org/blog/2018/11/27/react-16-roadmap.html#suspense-for-server-rendering

PR #14628, opened 2019-01-18
Add directory details to the package.json of all packages

greysteil:
Specifying the directory as part of the `repository` field in a `package.json` allows third party tools to provide better support when working with monorepos. For example, it allows them to correctly construct a commit diff for a specific package.

This format was accepted by npm in https://github.com/npm/rfcs/pull/19.


gaearon:
Are we sure this won't break some other tools relying on it being a string? I'd like to see some smaller packages do that first as pilot.

gaearon:
Seems good. Thanks.

PR #14632, opened 2019-01-18
Move lazy._result assignment

gaearon:
Follow-up to https://github.com/facebook/react/pull/14626.


gaearon:
Hmm. Actually this is wrong. It will infinite-loop for sync rejection because we're overwriting the `result`.

PR #14634, opened 2019-01-18
Upgrade Jest to 24.0.0-alpha.12

rickhanlonii:
## Overview

Upgrades to the Jest 24 beta release

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
@@ -2427,7 +2427,7 @@ describe('ReactDOMComponent', () => {
       const container = document.createElement('div');
 
       ReactDOM.render(<img src={obj} />, container);
-      expect(container.firstChild.src).toBe('hello');
+      expect(container.firstChild.src).toBe('http://localhost/hello');
 
       ReactDOM.render(<svg arabicForm={obj} />, container);
       expect(container.firstChild.getAttribute('arabic-form')).toBe('hello');
@@ -2461,7 +2461,7 @@ describe('ReactDOMComponent', () => {
       const child = Object.create(parent);
       const el = ReactTestUtils.renderIntoDocument(<img src={child} />);
 
-      expect(el.src).toBe('hello.jpg');
+      expect(el.src).toBe('http://localhost/hello.jpg');
     });
 
     it('assigns ajaxify (an important internal FB attribute)', function() {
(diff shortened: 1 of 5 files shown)


packages/react-dom/src/__tests__/ReactDOMComponent-test.js
```diff
@@ -2427,7 +2427,7 @@ describe('ReactDOMComponent', () => {
       const container = document.createElement('div');
 
       ReactDOM.render(<img src={obj} />, container);
-      expect(container.firstChild.src).toBe('hello');
+      expect(container.firstChild.src).toBe('http://localhost/hello');
```
rickhanlonii: jsdom added localhost as a default hostname

packages/react-test-renderer/src/__tests__/ReactTestRendererAsync-test.js
```diff
@@ -158,7 +158,7 @@ describe('ReactTestRendererAsync', () => {
 
       expect(() =>
         expect(renderer).toFlushAndYieldThrough(['foo', 'baz']),
-      ).toThrow('Expected value to equal:');
+      ).toThrow('Difference:');
```
rickhanlonii: The error matcher string changed, we no longer have the "Expected value to equal" text at the top of error outputs (shoutout to @pedrottimark)

scripts/jest/config.base.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
     '.*': require.resolve('./preprocessor.js'),
   },
   setupFiles: [require.resolve('./setupEnvironment.js')],
-  setupTestFrameworkScriptFile: require.resolve('./setupTests.js'),
+  setupFilesAfterEnv: [require.resolve('./setupTests.js')],
```
rickhanlonii: We changed this setting to allow multiple files and name it something that makes more sense (shoutout to @ddruker)

acdlite:
Before I review this can you explain why we're upgrading to an alpha? I don't have context

milesj:
@acdlite The biggest change in Jest 24 is Babel 7 support, so it would be good to test that it works. Jest 24 official will land soon(tm).

acdlite:
We're not even on Babel 7 yet so I don't get why we're upgrading Jest either. I'm not necessarily opposed, I'm just curious what the motivation is.

gaearon:
I told @rickhanlonii I hit a bug related to Promise rejection that was fixed in master so that's probably why. But agreed we probably should hold off until we use Babel 7. Sorry I didn't realize those changes were related.

rickhanlonii:
@milesj Jest supports babel 7 in Jest 23, in 24 we just _use_ babel 7 internally. There are over [100 features and bug fixes](https://github.com/facebook/jest/blob/44054afb917dd2a02f01adfc292375d5752d77f2/CHANGELOG.md#features) in 24, but I think the biggest is probably handling typescript by default

The context was to get a bug fix to @gaearon, but I'm cool closing until we release Jest 24 (soon). We can use this as a reference for what it would take to upgrade 👍

SimenB:
#12548 must land first - jest 24 won't work with babel 6

(might be possible to hack it by installing `babel-jest@23`, but I doubt you want to go down that rabbit hole)

PR #14639, opened 2019-01-19
Double-render function components in DEV in StrictMode

gaearon:
@acdlite noticed we're not doing this, but we should. It's DEV-only. Function components already need to not have observable side effects — so I think this is perfectly reasonable to do in a minor.

The test changes are a bit wonky. Maybe we could abstract this away but I felt it's only a few files for now and not too bad.


bvaughn:
This change might be observable in a way that breaks people's existing tests (if they happen to e.g. use mock functions for components and then verify the number of renders)

gaearon:
We can do that but for now I'll go with https://github.com/facebook/react/pull/14643 instead

PR #14643, opened 2019-01-21
Double-render function components with Hooks in DEV in StrictMode

gaearon:
Alternative to https://github.com/facebook/react/pull/14639. (See last commit.)

Less risky. Would make DEV faster for the stateless case. It's mostly Hooks abuse that we're worried about anyway. Added a test for `useMemo(fn, [])` specifically.

### DIFF
--- a/packages/react-art/src/__tests__/ReactART-test.js
+++ b/packages/react-art/src/__tests__/ReactART-test.js
@@ -385,7 +385,7 @@ describe('ReactART', () => {
       </CurrentRendererContext.Provider>,
     );
 
-    ReactNoop.flushThrough(__DEV__ ? ['A', 'A'] : ['A']);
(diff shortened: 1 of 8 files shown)


gaearon:
I'll merge to unblock the sybnc.

bvaughn:
Nice compromise, Dan

gaearon:
I already reverted

gaearon:
(Resubmitted in https://github.com/facebook/react/pull/14654)

bvaughn:
Hm?

This was just a follow up comment from #14639.

I just meant the concept of doing the double-rendering only for function components that used hooks.  Seems like a nice compromise from double rendering everything.

gaearon:
Just wanted to clarify we haven't actually landed this

Issue #14629, opened 2019-01-18
useImperativeHandle should warn when second arg isn't a function


lorenzomigliorero:
I've noticed a strange bug with the react redux `forwardRef` opt-in.
If i use it with a connected class component, everything is ok:
```javascript
const MyComponent = class Test extends React.Component {
  foo = () => console.log("Print foo from Test component");
  render() {
    return null;
  }
};

const ConnectedComponent = connect(
  null,
  null,
  null,
  { forwardRef: true }
)(MyComponent);

const store = createStore(() => {});

function App() {
  return (
    <Provider store={store}>
      <ConnectedComponent
        ref={ref => {
          if (ref) ref.foo();
        }}
      />
    </Provider>
  );
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```
If i use it with a connected functional component that use `forwardRef` with `useImperativeHandle`, i obtain a strange error: `create is not a function` in `commitHookEffectList` `react-dom` method.

```javascript
const MyComponent = React.forwardRef((props, ref) => {
  useImperativeHandle(ref, {
    foo: () => console.log("Print foo from Test component")
  });

  return null;
});

const ConnectedComponent = connect(
  null,
  null,
  null,
  { forwardRef: true }
)(MyComponent);

const store = createStore(() => {});

function App() {
  return (
    <Provider store={store}>
      <ConnectedComponent
        ref={ref => {
          if (ref) ref.foo();
        }}
      />
    </Provider>
  );
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```

I create a codepen to reproduce the issue: https://codesandbox.io/s/r7rpml460o

PS: Sorry for the cors error, but i don't find the way to add `react@next`as cdn

hamlim:
I believe the issue here is that `useImperativeHandle` expects the second argument to be a function that returns the reference:

```js
// before (your sample code)
useImperativeHandle(ref, {
  foo: () => console.log("Print foo from Test component")
});
// after
useImperativeHandle(ref, () => ({
  foo: () => console.log("Print foo from Test component")
}));
```

See: https://reactjs.org/docs/hooks-reference.html#useimperativehandle for more information about the API of the hook.

Maybe the error thrown could be more clear?

lorenzomigliorero:
You're right, `useImperativeHandle` accept a function as first argument.
A warning on wrong type argument could be a good idea.

Aberman12:
[...]

PR #14647, opened 2019-01-21
Warn about incorrect use of useImperativeHandle()

gaearon:
Fixes https://github.com/facebook/react/issues/14629.
I've made these mistakes quite a few times and I expect more people to do it.

This protects against:

* `useImperativeHandle({ focus })`
* `useImperativeHandle(ref, { focus })`

(Correct one is `useImperativeHandle(ref, () => ({ focus }))`.)

While adding these warnings I also noticed we don't set current fiber during some effect traversals. This seems like a regression. It caused warnings from Hooks during commit phase to have no component stack. So I fixed that too.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -674,6 +674,12 @@ export function useImperativeHandle<T>(
 ): void {
   if (__DEV__) {
     currentHookNameInDev = 'useImperativeHandle';
+    warning(
+      typeof create === 'function',
+      'Expected useImperativeHandle() second argument to be a function ' +
+        'that creates a handle. Instead received: %s.',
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -690,6 +696,14 @@ export function useImperativeHandle<T>(
       return () => refCallback(null);
     } else if (ref !== null && ref !== undefined) {
       const refObject = ref;
+      if (__DEV__) {
+        warning(
+          refObject.hasOwnProperty('current'),
+          'Expected useImperativeHandle() first argument to either be a ' +
+            'ref callback or React.createRef() object. Instead received: %s.',
```
gaearon: (Note: callback is handled by previous if-clause so we don't need to check for that)

PR #14648, opened 2019-01-21
Disallow reading context during useMemo etc

gaearon:
This extends https://github.com/facebook/react/pull/14608 to direct `readContext` calls and adds tests. It turned out it's not super easy.

I started by directing Dispatcher's `readContext` to `FiberHooks` implementation so that we can check `currentlyRenderingFiber` (which https://github.com/facebook/react/pull/14608 temporarily resets before calling user code from Hooks). But then I realized `currentlyRenderingFiber` is also `null` for classes and context consumers from which it's legit to `readContext()`.

I struggled a bit but then noticed we already rely on `renderExpirationTime` (per comment) to tell whether we're in a Hook-capable component or not. So I used that too, and only throw if `renderExpirationTime !== NoWork` (we’re using Hooks) and `currentlyRenderingFiber === null` (we’ve stepped into the user code).

While I was there, I also removed unnecessary `FiberBeginWork -> CurrentDispatcher -> readContext` indirection in favor of a direct call, as  `FiberBeginWork` itself is our own code.

This PR is client-side only. I started looking into SSR but I think there might be more warts there and I want to do it separately.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -14,7 +14,7 @@ import type {HookEffectTag} from './ReactHookEffectTags';
 
 import {NoWork} from './ReactFiberExpirationTime';
 import {enableHooks} from 'shared/ReactFeatureFlags';
-import {readContext} from './ReactFiberNewContext';
+import {readContext as readContextWithoutCheck} from './ReactFiberNewContext';
 import {
   Update as UpdateEffect,
   Passive as PassiveEffect,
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -771,6 +771,29 @@ export function useMemo<T>(
...
+  observedBits: void | number | boolean,
+): T {
+  // Forbid reading context inside Hooks.
+  // The outer check tells us whether we're inside a Hook like useMemo().
+  // However, it would also be true if we're rendering a class.
+  if (currentlyRenderingFiber === null) {
+    // The inner check tells us we're currently in renderWithHooks() phase
+    // rather than, for example, in a class or a context consumer.
+    // Then we know it should be an error.
+    if (renderExpirationTime !== NoWork) {
```
threepointone: til, good check.

PR #14649, opened 2019-01-21
Add test coverage for readContext() on the server

gaearon:
Only affects tests. Renames some variables and adds test coverage for SSR `readContext`.


gaearon:
Shouldn't be controversial

PR #14650, opened 2019-01-21
Forbid calling readContext() inside useMemo() for SSR

gaearon:
# Based on https://github.com/facebook/react/pull/14672

Adds the same warning for SSR.


gaearon:

PR #14653, opened 2019-01-21
Disallow reading context during useMemo etc

gaearon:
Reverts facebook/react#14651

In case we want to get it in

### DIFF
--- a/packages/react-reconciler/src/ReactFiberClassComponent.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.js
@@ -20,7 +20,6 @@ import {
 import ReactStrictModeWarnings from './ReactStrictModeWarnings';
 import {isMounted} from 'react-reconciler/reflection';
 import {get as getInstance, set as setInstance} from 'shared/ReactInstanceMap';
-import ReactSharedInternals from 'shared/ReactSharedInternals';
 import shallowEqual from 'shared/shallowEqual';
 import getComponentName from 'shared/getComponentName';
 import invariant from 'shared/invariant';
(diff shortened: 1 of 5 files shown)


gaearon:
I pushed another commit that introduces stashing to NewContext so that we don't need to rely on `expirationTime`.

@acdlite Please let me know if this is what you meant — and whether the edge case test I added in https://github.com/facebook/react/pull/14653/commits/093c8a27e9a20d28d85dddc7b34254c29af67af6 and https://github.com/facebook/react/pull/14653/commits/7cb26fbd25f82ff5098ee662b431ebe21dfb70e9 is sufficient. (I couldn't get it to fail with the old approach so I probably still misunderstand the issue.)

acdlite:
This looks good to me, gimme a few moments to try to break it and then I'll stamp

acdlite:
Also that reminds me we need a follow-up at some point to add the same guard to class component render phase functions. (Not super urgent, IMO, since that's already in master.)

gaearon:
Right but when (during class render) would we want to disable it? It’s already disabled outside of class render.

gaearon:
In other words — I thought the whole point of readContext is it *should* work in classes.

acdlite:
Not in a setState reducer though

gaearon:
Aaah.

gaearon:
What makes `setState` reducer conceptually different from gDSFP? The message says we support in in render phase lifecycles including render and gDSFP.

acdlite:
Left comment on the PR: https://github.com/facebook/react/pull/14670#issuecomment-456858534

sebmarkbage:
I think even this is too much for prod.

Hooks are about twice as slow as they should be already and this is adding to a bunch of micro optimizing we need to do.

In a better typed form we can guarantee that this never happens. The best way we have to ensure that right now is warnings.

As long as we can warn in all cases someone does this, I think it’s fine to leave it in production code since at worst you’re getting what you expected anyway.

gaearon:
Should I... revert? :-)

sebmarkbage:
Well you could land this and follow up by making all the stash things set a flag that triggers a warning in dev. :)

sebmarkbage:
I’m most concerned about useMemo and the initial state/action calls in useState/useReducer since those happens just automatically during render. They’re not trigger by an update. So they scale up by the size of the render tree.

I haven’t vetted all the reducer callsites if they only happen if there is an actual update in the queue or if they always happen.

But for consistency we might as well make them all DEV only.

gaearon:
Making it all DEV-only in https://github.com/facebook/react/pull/14677

PR #14657, opened 2019-01-22
Bump GCC

gaearon:
Let's see if this works.


gaearon:
@orta Any idea why Danger didn't run?

gaearon:
Generally yes. (Although I'd also look at the built diff.)

I hoped to see a size comparison but for some reason there was none. Let me try locally.

gaearon:
Oh maybe it's because there was zero change. 😄

Issue #14404, opened 2018-12-08
eslint-plugin-react-hooks should report errors inside unnamed functions


paboulos:
I want to report a bug for the hooks plugin.

**What is the current behavior?**
There was no error report after running eslint, but the component failed when running in the browser.
From the chrome dev console it reported "Uncaught Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement." 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React.
Here is a link to the github repo:
https://github.com/paboulos/react-hooks-eslint

**What is the expected behavior?**
Followed The Hooks API guide which says React hooks provides a linter plugin to enforce these rules automatically.Therefore it should have reported a usage violation when the eslint hooks plugin is specified. 
**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
Using window 10 OS and Chrome browser.
1. First ran npx create-react-app Hooks
2. Installed react 16.7.0-alpha.2 and react-dom 16.7.0-alpha.2
3. Installed eslint dev dependencies:
 "babel-eslint": "9.0.0",
    "babel-loader": "8.0.4",
    "eslint": "5.9.0",
    "eslint-config-airbnb": "17.1.0",
    "eslint-loader": "2.1.1",
    "eslint-plugin-import": "2.14.0",
    "eslint-plugin-jsx-a11y": "6.1.2",
    "eslint-plugin-react": "7.11.1",
    "eslint-plugin-react-hooks": "0.0.0"
4. Created the .eslintrc.json following the instructions from the Hooks API Doc
Then ran package script lint as follows: "npm run lint"
no errors reported.
Then ran package script start as follows: "npm start"
The React component CountHooks calls useState incorrectly and reports error in the browser dev console.

threepointone:
I wasn't able to reproduce this in codesandbox (ie - it shows as a lint error correctly - https://codesandbox.io/s/n4pjo8xvjj).

It also appears you've set up eslint with this rule 
```
 "react-hooks/rules-of-hooks": ["error","never"],
```
I'm not super familiar with eslint config options, but doesn't "never" mean the rule is disabled?

paboulos:
See below, and changed it to "error" as per the npm package doc for hooks, still doesn't work on windows 10 OS.
Are you saying there is no support on windows?
[...]

PR #14661, opened 2019-01-23
Fix issue with multiple code branches in hooks linter

### LINKED ISSUE #14362: react-hooks linter fails with unusual for loop
bug
React hooks linter fails with this piece of code

```js
function useHook() {
  const ref = React.useRef([1, 2, 3]);
  const res = []
  // valid because there are not unexpected returns
  if (ref.current) {
    for (let i = 0; i !== ref.current.length && !ref.current[i]; ++i ) {
      res.push(ref.current[i]);
    }
  }
  React.useLayoutEffect(() => {});
}
```

however a bit more usual for loop is handled as expected
```js
function useHook() {
  const ref = React.useRef([1, 2, 3]);
  const res = []
  if (ref.current) {
    for (let i = 0; i !== ref.current.length; ++i ) {
      res.push(ref.current[i]);
    }
  }
  React.useLayoutEffect(() => {});
}
```

Skip for looks without returns

eslint-plugin-react-hooks 0.0.0

Yurickh:
I changed the algorithm to a simple DFS so we can find the correct number of possible paths from the hook to the start of the function.

I'm not sure I should change the other (similar) functions as I couldn't think of a breaking example for them.

Fixes #14362

### DIFF
--- a/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
+++ b/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
@@ -110,29 +110,54 @@ export default {
          * Segments 1 and 2 have one path to the beginning of `MyComponent` and
          * segment 3 has two paths to the beginning of `MyComponent` since we
          * could have either taken the path of segment 1 or segment 2.
+         *
+         * Populates `cyclic` with cyclic segments.
          */
 
-        function countPathsFromStart(segment, visited = new Set()) {
-          if (codePath.thrownSegments.includes(segment)) {
-            return 0;
-          }
+        function countPathsFromStart(segment) {
+          const {cache} = countPathsFromStart;
+          let paths = cache.get(segment.id);
 
-          // We reached the destination
-          if (segment.prevSegments.length === 0) {
-            return 1;
-          }
-
-          let paths = 0;
-          visited.add(segment.id);
-
-          for (const prevSegment of segment.prevSegments) {
-            // Check if we already visited the segment so we don't fall into a cycle.
-            if (!visited.has(prevSegment.id)) {
-              paths += countPathsFromStart(prevSegment, visited);
+          // If `paths` is null then we've found a cycle! Add it to `cyclic` and
+          // any other segments which are a part of this cycle.
+          if (paths === null) {
+            if (cyclic.has(segment.id)) {
+              return 0;
+            } else {
+              cyclic.add(segment.id);
+              for (const prevSegment of segment.prevSegments) {
+                countPathsFromStart(prevSegment);
+              }
+              return 0;
             }
           }
 
-          visited.delete(segment.id);
+          // We have a cached `paths`. Return it.
+          if (paths !== undefined) {
+            return paths;
+          }
+
+          // Compute `paths` and cache it. Guarding against cycles.
+          cache.set(segment.id, null);
+          if (codePath.thrownSegments.includes(segment)) {
+            paths = 0;
+          } else if (segment.prevSegments.length === 0) {
+            paths = 1;
+          } else {
+            paths = 0;
+            for (const prevSegment of segment.prevSegments) {
+              paths += countPathsFromStart(prevSegment);
+            }
+          }
+
+          // If our segment is reachable then there should be at least one path
+          // to it from the start of our code path.
+          if (segment.reachable && paths === 0) {
+            cache.delete(segment.id);
+          } else {
+            cache.set(segment.id, paths);
+          }
+
           return paths;
         }
 
@@ -253,6 +278,7 @@ export default {
           return length;
         }
 
+        countPathsFromStart.cache = new Map();
         countPathsToEnd.cache = new Map();
         shortestPathLengthToStart.cache = new Map();
 
--- a/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
+++ b/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
@@ -110,47 +110,29 @@ export default {
          * Segments 1 and 2 have one path to the beginning of `MyComponent` and
          * segment 3 has two paths to the beginning of `MyComponent` since we
          * could have either taken the path of segment 1 or segment 2.
-         *
-         * Populates `cyclic` with cyclic segments.
          */
 
-        function countPathsFromStart(segment) {
-          const {cache} = countPathsFromStart;
-          let paths = cache.get(segment.id);
-
-          // If `paths` is null then we've found a cycle! Add it to `cyclic` and
-          // any other segments which are a part of this cycle.
-          if (paths === null) {
-            if (cyclic.has(segment.id)) {
-              return 0;
-            } else {
-              cyclic.add(segment.id);
-              for (const prevSegment of segment.prevSegments) {
-                countPathsFromStart(prevSegment);
-              }
-              return 0;
-            }
-          }
-
-          // We have a cached `paths`. Return it.
-          if (paths !== undefined) {
-            return paths;
-          }
-
-          // Compute `paths` and cache it. Guarding against cycles.
-          cache.set(segment.id, null);
+        function countPathsFromStart(segment, visited = new Set()) {
           if (codePath.thrownSegments.includes(segment)) {
-            paths = 0;
-          } else if (segment.prevSegments.length === 0) {
-            paths = 1;
-          } else {
-            paths = 0;
-            for (const prevSegment of segment.prevSegments) {
-              paths += countPathsFromStart(prevSegment);
+            return 0;
+          }
+
+          // We reached the destination
+          if (segment.prevSegments.length === 0) {
+            return 1;
+          }
+
+          let paths = 0;
+          visited.add(segment.id);
+
+          for (const prevSegment of segment.prevSegments) {
+            // Check if we already visited the segment so we don't fall into a cycle.
+            if (!visited.has(prevSegment.id)) {
+              paths += countPathsFromStart(prevSegment, visited);
             }
           }
-          cache.set(segment.id, paths);
 
+          visited.delete(segment.id);
           return paths;
         }
 
@@ -271,7 +253,6 @@ export default {
           return length;
         }
 
-        countPathsFromStart.cache = new Map();
         countPathsToEnd.cache = new Map();
         shortestPathLengthToStart.cache = new Map();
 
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/__tests__/__snapshots__/ReactIncrementalPerf-test.internal.js.snap
```diff
@@ -396,30 +396,30 @@ exports[`ReactDebugFiberPerf supports Suspense and lazy 2`] = `
 "
 `;
 
-exports[`ReactDebugFiberPerf supports portals 1`] = `
+exports[`ReactDebugFiberPerf supports memo 1`] = `
```
Yurickh: I'm not really sure why this changed, as I just run the tests, and didn't ask for snapshot updating at any time.

packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
```diff
@@ -110,47 +110,29 @@ export default {
...
-              return 0;
-            } else {
-              cyclic.add(segment.id);
-              for (const prevSegment of segment.prevSegments) {
-                countPathsFromStart(prevSegment);
-              }
-              return 0;
-            }
+        function countPathsFromStart(segment, visited = new Set()) {
+          if (codePath.thrownSegments.includes(segment)) {
```
hg-pyun: Is it okay `includes method` without polyfill?
https://developer.mozilla.org/ko/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

Yurickh: I guess so, since the linter always runs in a node environmemt.

Yurickh: Also, IIRC, thrownSegments isn’t an array here, it’s just array-like.

Jessidhia: Assuming this is an eslint@5 plugin, then yes as it's supported since node 6.

Yurickh: Since it is a direct extract from the original code, this shouldn't really be an issue.

hg-pyun: Okay, It looks like good.

packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
```diff
@@ -110,47 +110,29 @@ export default {
          * Segments 1 and 2 have one path to the beginning of `MyComponent` and
          * segment 3 has two paths to the beginning of `MyComponent` since we
          * could have either taken the path of segment 1 or segment 2.
-         *
-         * Populates `cyclic` with cyclic segments.
          */
 
-        function countPathsFromStart(segment) {
-          const {cache} = countPathsFromStart;
```
calebmer: The reason I use a cache here is that in complex components, time complexity can really start to spike up. Consider:

```js
function MyComponent() {
  if (a) {} else {}
  if (b) {} else {}
  if (c) {} else {}
  useHook();
}
```

Here we have 2<sup>3</sup> = 8 paths from `useHook()` to the start of `MyComponent`. Representing the following combinations of `a`, `b`, and `c`.

|a|b|c|
|-|-|-|
|true|true|true|
|false|true|true|
|true|false|true|
|false|true|false|
|true|false|false|
|false|true|false|
|false|false|true|
|false|false|false|

So we have a 2<sup>n</sup> exponential relationship, fun. Now remember that every `&&` and `||` (in the future perhaps also [`??`](https://github.com/tc39/proposal-nullish-coalescing)) introduces a condition since `false && expensive()` will not execute `expensive()`.

Let’s say we have a complex component that has 5 conditions placed in the component before 6 hooks all in the same segment. Without a cache we have to call `countPathsFromStart()` on 2<sup>5</sup> paths 6 times. With a cache we only need to call `countPathsFromStart()` on 2 × 5 segments because we cache the value for every segment so we only need to visit each segment once.

In big-O notation where “n” is the number of conditions and “h” is the number of hooks we have O(2n) time complexity with a cache and O(2<sup>n</sup> × h) without a cache.

To see what I mean in practice try adding the following component to the test suite. On this branch I waited about 10s and the test suite hadn’t finished. When I switched back to master the test suite finished in about 3s.

```js
function MyComponent() {
  // 40 conditions
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}
  if (c) {} else {}

  // 10 hooks
  useHook();
  useHook();
  useHook();
  useHook();
  useHook();
  useHook();
  useHook();
  useHook();
  useHook();
  useHook();
}
```

Yurickh: Hi! Thanks for the chime in, explaining the original rationale behind this piece of code, that was really enlightening.

Unfortunately, I'm not sure there's a way to work around this performance regression. Using the cache as is proved itself faulty, as issue proved.
I could just remove the item from the cache once we finished visiting it, but the complexity of the overall algorithm would be the same.
I guess the complexity of the problem of finding the number of paths between two nodes on a graph can't really be reduced here.

I'll think about how we could improve it for the case of multiple hooks, so the complexity is not so high.

calebmer:
@Yurickh I found a fix for the problem that does not involve removing the cache. In `countPathsFromStart()` make the following change:

```diff
               paths += countPathsFromStart(prevSegment);
             }
           }
-          cache.set(segment.id, paths);
+          // If our segment is reachable then there should be at least one path
+          // to it from the start of our code path.
+          if (segment.reachable && paths === 0) {
+            cache.delete(segment.id);
+          } else {
+            cache.set(segment.id, paths);
+          }

           return paths;
         }
```

If you choose to use this fix, I encourage you to think about why this is the correct solution. In theory, `segment.reachable && paths === 0` is a correct invariant to have for `countPathsFromStart()` but not for `countPathsToEnd()`. So if we detect a violation to this invariant we “forget” the computation which allows it to be re-computed later when we’ll get the correct result. (This has to do with how cycles are treated.)

Perhaps there is a more principled fix in the `if (paths === null) {` code. Since all we’re doing here is saying “we don’t like this result you gave us so try again later”.

If you choose to use this fix I also recommend adding the test case I included above to prevent future performance regressions.

calebmer:
A more in-depth look at this problem. To debug I added to the top of the `countPathsFromStart()` function. It will log the segment’s id along with some indentation so we can see the nesting of our calls. (I got this idea from one of Sophie’s tweets, actually)

```js
console.log('  '.repeat(i) + segment.id);
```

That gives us:

```
s2_6
  s2_2
    s2_1
    s2_5
      s2_4
        s2_3
          s2_2
  s2_3
```

@LosYear’s diagram from #14362 is really helpful for understanding this output, so I’ll include it:



Adding some annotations to the log:

```
s2_6
  s2_2
    s2_1              <---- first segment
    s2_5
      s2_4
        s2_3          <---- first call to countPathsFromStart(s2_3)
          s2_2        <---- cycle broken
  s2_3                <---- second call to countPathsFromStart(s2_3)
```

Our first call to `countPathsFromStart(s2_3)` looks at the segment which comes before it, `s2_2`. However, we’ve already visited `s2_2` so we’ve found a cycle! We break the cycle by returning 0 which means `s2_3` gets a count of 0. However, 0 is not the correct value for `s2_2`! You’ll see that the first time we visit `s2_2` it has two previous segments: `s2_1`, our initial segment, and `s2_5`. `s2_1` is the initial segment so it gives us 1 path. `s2_5` cycles back to `s2_2` so it gives us 0 paths. Therefore the correct number of paths for `s2_2` is 1, not 0. But we got 0 in our first call to `countPathsFromStart(s2_3)` because we were breaking the cycle.

So that’s why we delete the result when `segment.reachable && paths === 0` is violated. We know that if we call `countPathsFromStart(s2_3)` again that this time `s2_2` will have the correct value, 1, instead of 0.

packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
calebmer: I added some components below with a possible solution and the rationale behind that solution. TL;DR when `countPathsFromStart()` breaks a cycle it gives a temporary result of 0 to avoid looping forever.

sompylasar:
_I apologize in advance for the off-topic irrelevance of my question to this PR; there is no true "code review" feature of arbitrary piece of code at arbitrary time in GitHub yet, only "diff review"._

I was curious and looked around the code of the eslint rule, and found this comment which raised a question in me:
```
        // Gets the function name for our code path. If the function name is
        // `undefined` then we know either that we have an anonymous function
        // expression or our code path is not in a function. In both cases we
        // will want to error since neither are React function components or
        // hook functions.
        const codePathFunctionName = getFunctionName(codePathNode);
```
and then:
```
            } else if (codePathFunctionName) {
              // Custom message if we found an invalid function name.
              const message =
                `React Hook "${context.getSource(hook)}" is called in ` +
                `function "${context.getSource(codePathFunctionName)}" ` +
                'which is neither a React function component or a custom ' +
                'React Hook function.';
              context.report({node: hook, message});
```

I reviewed the `function getFunctionName(node) {` and it looks like it's only concerned with use cases of finding a function name of a hook, not a function component.

In particular, I may write a function component this way, using an anonymous function:
```
const MyComponentWithStyles = withStyles((props) => {
  // some useAnyHook() code here
});
```

Would it become a problem that the anonymous function inside a call argument is meant to be a React function component, but it's not annotated to be that? Is this a rule now to not use anonymous functions to define React function components?

Yurickh:
Hi, @sompylasar. In fact, you'll find an issue about your discoveries on https://github.com/facebook/react/issues/14404#issuecomment-455414668

sompylasar:
@Yurickh Thanks! It's exactly that. I searched for eslint hooks issues but missed this one (oh I realized I might have had is:pr in the filters).

gaearon:
Thank you!

PR #14670, opened 2019-01-23
Forbid readContext() during render-phase class setState() reducer form

gaearon:
## Depends on https://github.com/facebook/react/pull/14672

Addresses https://github.com/facebook/react/pull/14653#issuecomment-456582655 (I think?)

Not sure we want to actually do it but figured I'd send a PR.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
@@ -1347,6 +1347,31 @@ describe('ReactNewContext', () => {
       expect(ReactNoop.flush()).toEqual(['App', 'App#renderConsumer']);
       expect(ReactNoop.getChildren()).toEqual([span('goodbye')]);
     });
+
+    it('warns when reading context inside render phase class setState updater', () => {
+      const ThemeContext = React.createContext('light');
+
+      class Cls extends React.Component {
+        state = {};
+        render() {
+          this.setState(() => {
+            readContext(ThemeContext);
+          });
+          return null;
+        }
+      }
+
+      ReactNoop.render(<Cls />);
+      expect(ReactNoop.flush).toWarnDev(
+        [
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
```diff
@@ -1347,6 +1347,29 @@ describe('ReactNewContext', () => {
...
+          return null;
+        }
+      }
+
+      ReactNoop.render(<Cls />);
+      expect(() =>
+        expect(ReactNoop.flush).toThrow(
+          'Context can only be read while React is rendering',
+        ),
+      ).toWarnDev('Cannot update during an existing state transition', {
```
gaearon: We warn anyway so maybe it isn't worth it?

acdlite:
The reason to do this is because the semantics are currently undefined — what happens when context changes? Re-rendering will have no effect because the update already happened — and we want to preserve option value to possibly support it properly in the future. Like for code splitting reducers. Same reason we warn about reading context inside Hook reducer.

`getDerivedStateFromProps` is fine because we it every time.

gaearon:
That's true but we already unconditionally warn for render phase `setState` anyway. You can see it in the test. Usually if there's a warning, historically we didn't mind undefined behavior. It's meant to not officially be supported.

sebmarkbage:
Added a comment that I don’t think we should do this in prod.

https://github.com/facebook/react/pull/14653#issuecomment-456859212

gaearon:
Do we need an extra warning for this specifically? This test already warns with "Cannot update during an existing state transition".

gaearon:
I made this depend on https://github.com/facebook/react/pull/14672 and added a similar DEV-only warning on top.

gaearon:

gaearon:
Superseded by https://github.com/facebook/react/pull/14677

PR #14672, opened 2019-01-23
Make readContext() in Hooks DEV-only warning

gaearon:
Address https://github.com/facebook/react/pull/14653#issuecomment-456862761.


gaearon:

PR #14676, opened 2019-01-23
expose flushEffects on TestUtils

threepointone:
exposes flushEffects on TestUtils, letting testing libs to manually flush effects when testing their components. added a test. I'll update the docs/readmes if you folks think this a good approach.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactTestUtils-test.js
+++ b/packages/react-dom/src/__tests__/ReactTestUtils-test.js
@@ -14,6 +14,7 @@ let React;
 let ReactDOM;
 let ReactDOMServer;
 let ReactTestUtils;
+let ReactFeatureFlags;
 
 function getTestDocument(markup) {
   const doc = document.implementation.createHTMLDocument('');
@@ -28,6 +29,8 @@ function getTestDocument(markup) {
 
 describe('ReactTestUtils', () => {
   beforeEach(() => {
+    ReactFeatureFlags = require('shared/ReactFeatureFlags');
+    ReactFeatureFlags.enableHooks = true;
     createRenderer = require('react-test-renderer/shallow').createRenderer;
     React = require('react');
     ReactDOM = require('react-dom');
@@ -514,5 +517,28 @@ describe('ReactTestUtils', () => {
 
     ReactTestUtils.renderIntoDocument(<Component />);
     expect(mockArgs.length).toEqual(0);
+  });
+
+  it('flushEffects should flush effects, duh', () => {
+    let effected = false;
+    let layoutEffected = false;
+    function App() {
+      React.useEffect(() => {
+        effected = true;
+      });
+      React.useLayoutEffect(() => {
+        layoutEffected = true;
+      });
+      return null;
+    }
+    ReactTestUtils.renderIntoDocument(<App />);
+    // effects haven't fired yet
+    expect(effected).toBe(false);
+    // layout effects have, however
+    expect(layoutEffected).toBe(true);
+
+    ReactTestUtils.flushEffects();
+    // effects have now fired
+    expect(effected).toBe(true);
   });
 });
--- a/packages/react-dom/src/__tests__/ReactTestUtils-test.js
+++ b/packages/react-dom/src/__tests__/ReactTestUtils-test.js
@@ -14,7 +14,6 @@ let React;
 let ReactDOM;
 let ReactDOMServer;
 let ReactTestUtils;
-let ReactFeatureFlags;
 
 function getTestDocument(markup) {
   const doc = document.implementation.createHTMLDocument('');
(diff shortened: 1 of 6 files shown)


packages/react-dom/src/__tests__/ReactTestUtils-test.js
```diff
@@ -515,4 +518,27 @@ describe('ReactTestUtils', () => {
     ReactTestUtils.renderIntoDocument(<Component />);
     expect(mockArgs.length).toEqual(0);
   });
+
+  it('flushEffects should flush effects, duh', () => {
```
threepointone: I'll change this to be more formal :P

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -810,6 +811,7 @@ const ReactDOM: Object = {
   __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: {
     // Keep in sync with ReactDOMUnstableNativeDependencies.js
     // and ReactTestUtils.js. This is an array for better minification.
+    flushPassiveEffects,
```
gaearon: This won't minify the name

threepointone: I'll move it into the Events array

gaearon:
Can you check that `react-testing-library`'s tests would pass if you replace its hack with this?

threepointone:
alright I'll try that rn

threepointone:
had trouble setting up react-testing-library, but also checked https://github.com/kentcdodds/react-testing-library/blob/master/src/__tests__/render.js#L94-L99 and there's no coverage for any actual effects.

kentcdodds:
> there's no coverage for any actual effects.

The reason those tests are that way is because I didn't want to add tests for an experimental feature 😅

I'll give it a look :)

threepointone:
was about to reach out to you @kentcdodds. you could try building this and trying? or the next canary that includes this.

kentcdodds:
Just looking at this I'm pretty confident that this is exactly what we need.

The current (experimental) implementation of `flushEffects` in react-testing-library just does this: `ReactDOM.render(null, document.createElement('div'))` and the reason it does that is because it was the best way (causes the fewest side-effects) I could find to run `flushPassiveEffects` from outside of react.

I'm pretty confident that this is going to work. I'll confirm it soon.

kentcdodds:
I think that I'm actually going to pluck it off of `ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Events` directly though because I'd rather not import all of the test utils for this one thing 😬

kentcdodds:
Yup, this works great 💯

threepointone:
> I think that I'm actually going to pluck it off of ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Events. 
This *will* break unexpectedly. Is this so much really? https://unpkg.com/react-test-utils@0.0.1/

aweary:
@threepointone I believe it's part of the `react-dom` package via [`react-dom/test-utils`](https://unpkg.com/react-dom@16.7.0/cjs/react-dom-test-utils.development.js), [that package](https://www.npmjs.com/package/react-test-utils) isn't owned by us--but the point still stands. Unexpected breakage is no fun for anyone.

It might be nice to start fresh with a modern DOM testing package that doesn't have all the baggage of `ReactTestUtils`.

threepointone:
lol that's right, but my point remains (or is strengthened?) - there's no dependency cost to using it, it'll already be there.

kentcdodds:
It's mostly that using `react-dom/test-utils` gave me grief when trying to create the UMD build when I was still using the Simulate module. Maybe that problem has magically gone away? Or maybe I can figure out what was up 🤔

aweary:
@kentcdodds if it's still a problem, I'm happy to help figure it out! The fewer popular libraries using `__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED` the less grief there is for the React team.

threepointone:
btw fyi we're still still discussing what the api for this thing should be, there's some thoughts to what we can do better. it'll probably still be a helper on TestUtils, but the signature might be different. I'll keep y'all in the loop once we decide something, cheers

kentcdodds:
Interesting. How can the signature be different? Will it accept something to signify a specific part of the react tree to flush effects or something?

threepointone:
Thinking more about sequencing of tests, and what patterns we could unlock. I'll summarize here once there's clarity, promise

threepointone:
I'm going to close this pr, I'm spending some time this week trying to figure out what we'd like and how concurrent mode will affect these tests. thanks for the help @kentcdodds!

kentcdodds:
For you? Any. Time.

PR #14677, opened 2019-01-23
Make all readContext() and Hook-in-a-Hook checks DEV-only

gaearon:
* Changes it to be DEV-only
* Implements it for SSR
* Adds a similar DEV-only check for render phase class updates
* Changes all Hooks-in-Hooks warnings to similarly be DEV-only, removing the flaky current variable reassignment that led to bugs (includes a regression test)

### DIFF
--- a/packages/react-reconciler/src/ReactFiberNewContext.js
+++ b/packages/react-reconciler/src/ReactFiberNewContext.js
@@ -52,10 +52,8 @@ let currentlyRenderingFiber: Fiber | null = null;
 let lastContextDependency: ContextDependency<mixed> | null = null;
 let lastContextWithAllBitsObserved: ReactContext<any> | null = null;
 
-// We stash the variables above before entering user code in Hooks.
-let stashedCurrentlyRenderingFiber: Fiber | null = null;
-let stashedLastContextDependency: ContextDependency<mixed> | null = null;
-let stashedLastContextWithAllBitsObserved: ReactContext<any> | null = null;
+// Used in DEV to track whether reading context should warn.
+let areContextDependenciesStashedInDEV: boolean = false;
 
 export function resetContextDependences(): void {
   // This is called right before React yields execution, to ensure `readContext`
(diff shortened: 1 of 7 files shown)


gaearon:
Ready for review.

packages/react-reconciler/src/ReactFiberNewContext.js
```diff
@@ -52,37 +52,30 @@ let currentlyRenderingFiber: Fiber | null = null;
...
 
-export function stashContextDependencies(): void {
-  stashedCurrentlyRenderingFiber = currentlyRenderingFiber;
-  stashedLastContextDependency = lastContextDependency;
-  stashedLastContextWithAllBitsObserved = lastContextWithAllBitsObserved;
-
-  currentlyRenderingFiber = null;
-  lastContextDependency = null;
-  lastContextWithAllBitsObserved = null;
+export function stashContextDependenciesInDEV(): void {
```
sebmarkbage: These names don’t really make sense anymore.

How about enter/exitDisallowedContextRead?

gaearon:
A separate dispatcher sounds like a cleaner way to do it. I initially thought of replacing the dispatcher *while* we're inside a Hook like `useMemo` but then decided against it. I didn't think about wrapping the whole thing in DEV. Do you want to explore it in this PR, or land it first and I can iterate after 16.8?

gaearon:
I don't understand what you mean by "other hooks in the same scope".

sebmarkbage:
I think the way we allow currentlyRenderingFiber to be null is also wrong. That should also be done with the dispatcher technique.

I have a hard time reviewing these because it’s so spread out. If we’re confident that doing this whole refactor after 16.8, it’s fine but I can’t tell if that’s a breaking change or not from the code.

sebmarkbage:
For the currentlyRenderingFiber thing. It should always be non-null when the dispatcher is set because we’re always rendering something. We should instead swap out the whole dispatcher when we’re inside user code.

That way we also remove the whole resolveRenderingThingy() call in each hook.

sebmarkbage:
By “Other hooks in the same scope” I mean, it’s not clear to me from the code which hooks are disallowed and in what user code. Is it that most user code is covered by the “currentlyRenderingFiber === null” check and this is adding something additional or are there scopes where either one is in effect but not both?

gaearon:
I think there's some confusion here.

Here's a brief history of this code:

* Initially, `currentlyRenderingFiber === null` only happened in classes (and in general, when we're not rendering Hooks). It was never `null` while rendering Hook-capable components.

* https://github.com/facebook/react/pull/14608 repurposed `currentlyRenderingFiber === null`  check to also mean "inside user code".

* This PR backs out of that change, reverting `currentlyRenderingFiber === null` to its original meaning ("we're not rendering a function component"). But it adds DEV warnings with a separate flag for user code.

gaearon:
>By “Other hooks in the same scope” I mean, it’s not clear to me from the code which hooks are disallowed and in what user code. Is it that most user code is covered by the “currentlyRenderingFiber === null” check and this is adding something additional or are there scopes where either one is in effect but not both?

### Before #14608

* `currentlyRenderingFiber === null` **only** if we're **not** currently rendering a function component.
* If you call a Hook inside a Hook (e.g. `useMemo` from `useReducer`), nothing tells you it was wrong.

 ### Between #14608 and This PR

* `currentlyRenderingFiber === null` if we're not currently rendering a function component **or** if we entered user code from a Hook (e.g. either `fn` in `useMemo(fn)`, `fn` in `useReducer(fn)`, `fn` in `useState(fn)`, calls a Hook)
* If you call a Hook inside a Hook (e.g. `useMemo` from `useReducer`), you get an invariant.

 ### This PR
* `currentlyRenderingFiber === null` **only** if we're not currently rendering a function component.
* If you call a Hook inside a Hook (e.g. `useMemo` from `useReducer`), you get a warning.

PR #14679, opened 2019-01-23
Enable hooks!

bvaughn:
Commits are atomic:
* 7c54bff: Turn hooks on everywhere
* 021844a: Remove test overrides (and promote internal tests that only override the hooks flag)
* d1ffc25: Remove hooks flag entirely

### DIFF
--- a/packages/react-debug-tools/src/__tests__/ReactHooksInspection-test.js
+++ b/packages/react-debug-tools/src/__tests__/ReactHooksInspection-test.js
@@ -16,9 +16,6 @@ let ReactDebugTools;
 describe('ReactHooksInspection', () => {
   beforeEach(() => {
     jest.resetModules();
-    let ReactFeatureFlags = require('shared/ReactFeatureFlags');
-    // TODO: Switch this test to non-internal once the flag is on by default.
-    ReactFeatureFlags.enableHooks = true;
     React = require('react');
     ReactDebugTools = require('react-debug-tools');
   });
--- a/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
+++ b/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
@@ -17,9 +17,6 @@ let ReactDebugTools;
 describe('ReactHooksInspectionIntergration', () => {
   beforeEach(() => {
     jest.resetModules();
-    let ReactFeatureFlags = require('shared/ReactFeatureFlags');
-    // TODO: Switch this test to non-internal once the flag is on by default.
-    ReactFeatureFlags.enableHooks = true;
     React = require('react');
     ReactTestRenderer = require('react-test-renderer');
     ReactDebugTools = require('react-debug-tools');
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
@@ -38,7 +38,6 @@ function initModules() {
 
   ReactFeatureFlags = require('shared/ReactFeatureFlags');
   ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
-  ReactFeatureFlags.enableHooks = true;
   React = require('react');
   ReactDOM = require('react-dom');
   ReactDOMServer = require('react-dom/server');
--- a/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.js
@@ -9,7 +9,6 @@
 
 'use strict';
 
-let ReactFeatureFlags;
 let React;
 let ReactDOM;
 let Suspense;
@@ -21,8 +20,6 @@ describe('ReactDOMSuspensePlaceholder', () => {
 
   beforeEach(() => {
     jest.resetModules();
-    ReactFeatureFlags = require('shared/ReactFeatureFlags');
-    ReactFeatureFlags.enableHooks = true;
     React = require('react');
     ReactDOM = require('react-dom');
     ReactCache = require('react-cache');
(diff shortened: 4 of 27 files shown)


pureinx:
💪💪💪

zanedb:
YESSSSS I’M SO EXCITED

Daniel15:
HOOK ALL THE THINGS

int64ago:
Wow! I'm ready for it!

audiolion:
What caused the bundle size to go up?

bvaughn:
> What caused the bundle size to go up?

Hooks is a new set of APIs that were not included in (stable) builds before.

zaguiini:
Nice! When it’s going to be released?

ngtan:
I can't wait to bring it into my next release officially.

Daniel15:
> When is it going to be released?

Looks like @gaearon dropped a hint on Twitter recently:
[](https://twitter.com/dan_abramov/status/1086658019323269121)

Jemair:
Is there anything different from the beta version?

nanyang24:
I'm so excited！

zaguiini:
Also, what about React Native? When will it be ready?

aaronschubert0:
@zaguiini React Native usually syncs React every two weeks (on master), so I'd say once that's done and then wait for the next monthly RC

PaquitoSoft:
Hi there,

Fantastic news!

Are hooks supported in server side rendering?

Thanks

KBPsystem777:
Lets Hook the Future!

chaiyilin:
wondering what is the rhythm to update the react version in CRA?

gaearon:
You can always update React version independent of CRA in your project.

sakalx:
What is better from performance side
**useMemo or useState**

in case when need to do something only when props change
for example:

```javaScript
function parent({isValid}) {
  useMemo(() => {
    // do something
  }, [isValid]);
}
```
or
```javaScript
function parent({isValid}) {
  const [currStat, setStat] = useState(null);

  if (currStat !== isValid) {
    setStat(isValid);
    // do something
  }
}
```

Daniel15:
@sakalx You're probably better off asking questions somewhere people are more likely to see it 😃 See https://reactjs.org/community/support.html for some suggestions on where to get support. 

Reactiflux is pretty good if you want somewhere to chat to other people that use React: https://www.reactiflux.com/

bvaughn:
> What is better from performance side `useMemo` or `useState`
> in case when need to do something only when props change for example:

These two hooks are pretty different. You _could_ use the state hook in the way you mention above, but it's not what it's meant for.

As for which would perform better– in the above example, your call to `setStat` would cause another render (with the updated state value) which `useMemo` would avoid.

bvaughn:
> Never put side effects into `useMemo`.

I didn't read the original message as a side effect, rather– an expensive computation required in order to render

coderitual:
I have a stupid question. Can a function passed to `useEffect` be an async (eg to handle fetch with await)?

Async function inplicitly returns a promise and `useMemo` expects returned value to be a cleanup function which may be confusing.

zaguiini:
@coderitual no, because the return statement (a function or nothing) in an effect actually behaves something like componentWillUnmount and it’s synchronous, so you have to put an IIFE to work with async await inside an effect.

coderitual:
Yhm, I see. Thanks @zaguiini

gaearon:
@coderitual 

If you try to do `useEffect(async () => {})` you'll see a warning message that tells you exactly how to solve your problem. In general, my recommendation would be to try things and see what happens. :-)

screen shot 2019-01-26 at 9 03 05 pm

That said async chains in effects is usually a bad idea because props might change in the middle. So you need to be careful about handling race condition at every entry point.

Indeed Suspense will be a different solution that would make most data fetching in effects unnecessary.

coderitual:
@gaearon 

Oh, that's cool message. I realised something more important.

> That said async chains in effects is usually a bad idea because props might change in the middle. So you need to be careful about handling race condition at every entry point.

Didn't think about this but it's actually a really valid point. It looks like using many useEffects hooks "connected" by the state/prop variables is a way to go.

Gracias :)

zaguiini:
Just complementing @gaearon answer about promises: you most will most likely need an [abort controller](https://medium.com/@bramus/cancel-a-javascript-promise-with-abortcontroller-3540cbbda0a9) to prevent messing with state or something. Let me explain that with code:

```js
function fetchUser() {
  return new Promise((resolve) => {
    setTimeout(() => {
      resolve({
        name: 'Luis Felipe Zaguini',
      })
    }, 3000)
  })
}

function GreetUser({ id }) {
  const [user, setUser] = React.useState(undefined)

  React.useEffect(() => {
    const controller = new window.AbortController()

    fetchUser(id)
      .then((user) => {
        if(!controller.signal.aborted) {
          setUser(user)
        }
      })
      .catch(console.error)
    
    return () => {
      controller.abort()
    }
  }, [id])

  return (
    <div>
      {user ? `Hello, ${user.name}!` : 'Loading...'}
    </div>
  )
}
```

I think it's pretty clear what I'm trying to show. When the promise resolves, it will try to set the state. The promise will resolve even if the component is unmounted or the `id` changed.

See? With that, in the clean up function, we can "abort" that promise, avoiding the unmounted component state update warning. I don't think that's possible using async/await, not in that way, at least...

coderitual:
@zaguiini Hm, I am not sure if this is a good place for such discussion. Your example though doesn't use AbortController correctly. You are not aborting promise but just checking if the signal was raised after your promise resolved. The same behaviour you can achieve with async/await. The way Abort Controller should be used:

```
const controller = new AbortController();
const signal = controller.signal;
fetch(url, {signal}).then...
```

But thanks for reminding :)

zaguiini:
> @zaguiini Hm, I am not sure if this is a good place for such discussion. Your example though doesn't use AbortController correctly. You are not aborting promise but just checking if the signal was raised after your promise resolved. The same behaviour you can achieve with async/await. The way Abort Controller should be used:
> 
> ```
> const controller = new AbortController();
> const signal = controller.signal;
> fetch(url, {signal}).then...
> ```
> But thanks for reminding :)

It's ok. That was just an example :)

reyx7:
Finally, hooks will come out. 

```javascript
const [countDays, set] = useState(0);

useEffect(() => {
    setInterval(()=> set(countDays + 1),3600 * 24);
});
```

Thanks

PR #14680, opened 2019-01-23
Warn when creating multiple React roots on the same container

philipp-spiess:
We already warn when mixing `ReactDOM.render()` with `ReactDOM.createRoot()`. This seems like a good next step.

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -392,6 +392,7 @@ ReactRoot.prototype.unmount = function(callback: ?() => mixed): Work {
   callback = callback === undefined ? null : callback;
   if (__DEV__) {
     warnOnInvalidCallback(callback, 'render');
+    root.containerInfo._reactHasBeenPassedToCreateRootDEV = null;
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/client/ReactDOM.js
```diff
@@ -392,6 +392,7 @@ ReactRoot.prototype.unmount = function(callback: ?() => mixed): Work {
   callback = callback === undefined ? null : callback;
   if (__DEV__) {
     warnOnInvalidCallback(callback, 'render');
+    root.containerInfo._reactHasBeenPassedToCreateRootDEV = null;
```
philipp-spiess: I realized that we need to unset the prop so we can reuse the container after we've unmounted. I'm not sure if this is a good place to do so though since there can still be work pending - Maybe we add it into the `.then` callback? But that would mean that we can't do the following in sync:

```js
const root = ReactDOM.unstable_createRoot(container);
root.render(<App />);
root.unmount();
const root2 = ReactDOM.unstable_createRoot(container);
```

Which also seems bad.

PR #14681, opened 2019-01-23
Detect React roots by checking the parent fiber

### LINKED ISSUE #14535: Concurrent mode iOS hover behaviour
I hope this is the place to submit issues regarding concurrent mode.
I don't know if this is a bug or intended behaviour, but I have a situation where the my react app behaves differently depending on the use of concurrent mode.

Here is a link to a codesandbox demo: https://codesandbox.io/s/5xl0vnx11p

The behaviour on mobile Safari on iOS (12.1.2) is different whether using ReactDOM.render vs ReactDOM.createRoot. When ReactDOM.render is used, a tap results in an alert and after the alert is dismissed, the hover styles are applied. When using ReactDOM.createRoot, a tap results in the hover styles being applied. A second tap is needed to invoke the click handler.

(This double tap behaviour from iOS is described [here](https://css-tricks.com/annoying-mobile-double-tap-link-issue/))

Shouldn't the two render methods behave the same in this regard? or is this really intended behaviour?

BTW: A setTimout arround the mouseEnter event handler body does make ReactDOM.createRoot behave the same way as ReactDOM.render

philipp-spiess:
Fixes #14535
Related to #13778
Alternative to #14682

As @gaearon already noted, we can not rely on a container node having a `_reactRootContainer` to detect a React Root since the `createRoot()` API will not set it.

Furthermore, the `createRoot()` API is currently only setting a property on the container in DEV mode.

We could:

 1. Set a property in prod as well. (#14682)
 2. Pass in more information into the `appendChildToContainer()` config.

This PR is an attempt to implement 2. to visualize the outcome. It feels bad to do this though since non of the other renderers need that property and I’m not sure how stable the reconciler API is (i.e if we can just add properties like this).

### DIFF
--- a/packages/react-art/src/ReactARTHostConfig.js
+++ b/packages/react-art/src/ReactARTHostConfig.js
@@ -356,12 +356,14 @@ export function appendChild(parentInstance, child) {
   child.inject(parentInstance);
 }
 
-export function appendChildToContainer(parentInstance, child, containerIsRoot) {
+export function appendChildToContainer(parentInstance, child) {
   if (child.parentNode === parentInstance) {
     child.eject();
   }
   child.inject(parentInstance);
 }
+
+export const appendChildToPortalContainer = appendChildToContainer;
 
 export function insertBefore(parentInstance, child, beforeChild) {
   invariant(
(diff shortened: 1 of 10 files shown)


aweary:
I think adding a property or some other approach would be better. Changing the reconciler API to account for a renderer-specific problem doesn't seem ideal.

philipp-spiess:
Yeah, the other approach is a lot less invasive: #14682. I'd say we continue there. Thanks for your input!

philipp-spiess:
Let's not rule this out completely just yet (see https://github.com/facebook/react/pull/14682#issuecomment-457073195).

PR #14682, opened 2019-01-24
Detect React roots created by createRoot

### LINKED ISSUE #14535: Concurrent mode iOS hover behaviour

philipp-spiess:
Fixes #14535
Related to #13778
Alternative to #14681

As @gaearon already noted, we can not rely on a container node having a `_reactRootContainer` to detect a React Root since the `createRoot()` API will not set it.

Furthermore, the `createRoot()` API is currently only setting a property on the container in DEV mode.

We could:

 1. Set a property in prod as well.
 2. Pass in more information into the `appendChildToContainer()` config. (#14681)

This PR is an attempt to implement 1. It feels better than [the other approach](https://github.com/facebook/react/pull/14681) since we don't need to change the reconciler API.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
@@ -2683,6 +2683,23 @@ describe('ReactDOMComponent', () => {
           portalContainer,
         );
       }
+      const root = ReactDOM.unstable_createRoot(container);
+      root.render(<Component />);
+      jest.runAllTimers();
+
+      expect(typeof portalContainer.onclick).toBe('function');
+    });
+
+    it('adds onclick handler to a portal root mounted via a legacy React root', () => {
+      const container = document.createElement('div');
+      const portalContainer = document.createElement('div');
+
+      function Component() {
+        return ReactDOM.createPortal(
+          <div onClick={() => {}} />,
+          portalContainer,
+        );
+      }
 
       ReactDOM.render(<Component />, container);
       expect(typeof portalContainer.onclick).toBe('function');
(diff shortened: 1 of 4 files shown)


acdlite:
I would prefer to avoid relying on an expando

acdlite:
> we don't need to change the reconciler API

This is not necessarily a feature. The reconciler API might be flawed. If so, we shouldn't hack around it. Maybe the solution is to have separate methods for portals and roots.

philipp-spiess:
@acdlite Mhhhm it feels bad to change the reconciler API for this weird Safari bug but OTOH we only support portals for DOM right now, right? So perhaps there are differences between appending to a root and a portal and we just haven't noticed it yet. I'll update the other PR as well so we can compare both ideas.

acdlite:
> Mhhhm it feels bad to change the reconciler API for this weird Safari bug but OTOH we only support portals for DOM right now, right?

Yeah I'm not saying we should definitely change the reconciler API, I only want to make sure we consider the possibility that this could come up again, and that it's generally useful for the renderer to distinguish between roots and portals.

Issue #14669, opened 2019-01-23
Write a test for discrete events in Concurrent Mode


gaearon:
See this review comment: 

>Need one more test for the "counter" example: each discrete event must commit in sequence. So if you click on an increment button five times, the final result should always be five.

https://github.com/facebook/react/pull/14415#pullrequestreview-183478210

PR #14683, opened 2019-01-24
Add test for discrete event sequence in ConcurrentMode

aweary:
Resolves https://github.com/facebook/react/issues/14669

Adds a test that dispatches a sequence of discrete events (clicks) without flushing any work, then verifies that the final value is matches the expected final value in the sequence.

@gaearon I'm not positive this is exactly what #14669 meant but I think it's at least close 🤞

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
@@ -677,5 +677,43 @@ describe('ReactDOMFiberAsync', () => {
       // Therefore the form should have been submitted.
       expect(formSubmitted).toBe(true);
     });
+
+    it('discrete events commit in sequence', () => {
+      const buttonRef = React.createRef();
+      class Counter extends React.Component {
+        state = {counter: 0};
+        render() {
+          const {counter} = this.state;
+          return (
+            <button
+              ref={buttonRef}
+              onClick={() => this.setState({counter: counter + 1})}>
+              {counter}
+            </button>
+          );
+        }
+      }
+      ReactDOM.render(
+        <ConcurrentMode>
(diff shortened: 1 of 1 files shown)


packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
```diff
@@ -677,5 +677,43 @@ describe('ReactDOMFiberAsync', () => {
...
+        </ConcurrentMode>,
+        container,
+      );
+      let button = buttonRef.current;
+      // Dispatch a click event on the counter button to increment
+      function dispatchButtonClick() {
+        let firstEvent = document.createEvent('Event');
+        firstEvent.initEvent('click', true, true);
+        button.dispatchEvent(firstEvent);
+      }
```
philipp-spiess: Nit: You can do [`button.click();`](https://github.com/jsdom/jsdom/blob/16d3913eea0360c7757e75f7e266c3873c85b7dd/lib/jsdom/living/nodes/HTMLElement-impl.js#L34-L63) as well 🙂

packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
```diff
@@ -677,5 +677,43 @@ describe('ReactDOMFiberAsync', () => {
...
+      // Dispatch a sequence of 5 click events before flushing any work
+      dispatchButtonClick();
+      dispatchButtonClick();
+      dispatchButtonClick();
+      dispatchButtonClick();
+      dispatchButtonClick();
+      jest.runAllTimers();
```
Jessidhia: Also: before `jest.runAllTimers()` it should be `'4'`.

PR #14698, opened 2019-01-24
State update bug in concurrent mode

bvaughn:
I think this PR captures a bug that was reported internally at Facebook. I believe this code should work for both sync and concurrent mode, but the update fails for concurrent mode.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -257,7 +257,7 @@ function updateForwardRef(
     ) {
       // Only double-render components with Hooks
       if (workInProgress.memoizedState !== null) {
-        renderWithHooks(
+        nextChildren = renderWithHooks(
           current,
           workInProgress,
           render,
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1252,7 +1252,7 @@ function mountIndeterminateComponent(
       ) {
         // Only double-render components with Hooks
         if (workInProgress.memoizedState !== null) {
-          renderWithHooks(
+          value = renderWithHooks(
```
acdlite: Do the same for the other occurrences of `renderWithHook`

acdlite:
Explanation of why this fix works, for others who are interested:

When you call `renderWithHooks` during the initial render, React creates state queues for each of the State/Reducer Hooks. Those queues are bound to the `dispatch` method that is returned by the Hook, which is then closed over by event handlers, which in turn are included as part of the final `children` result that is returned by the component. These queues are also inherently coupled to the internal list of Hook nodes that is stored on `fiber.memoizedState`.

The cause of bug was that, in dev, we were calling `renderWithHooks` a second time, but we weren't using the children that correspond to that call.

PR #14701, opened 2019-01-26
Phased dispatcher

acdlite:
Refactor Hooks module to use separate dispatcher implementations for mount versus update. Main benefit is to speed up initial render by removing update-only branches from the mount path.


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -554,212 +594,234 @@ function basicStateReducer<S>(state: S, action: BasicStateAction<S>): S {
...
   reducer: (S, A) => S,
-  initialState: S,
-  initialAction: A | void | null,
+  initialState: void | S,
+  initialAction: void | null | A,
 ): [S, Dispatch<A>] {
-  if (__DEV__) {
-    if (reducer !== basicStateReducer) {
-      currentHookNameInDev = 'useReducer';
+  if (reducer === basicStateReducer) {
```
sebmarkbage: This seems like it can just move into `mountState`.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -554,212 +594,234 @@ function basicStateReducer<S>(state: S, action: BasicStateAction<S>): S {
...
+  hook: Hook,
+  reducer: (S, A) => S,
+  initialState: void | S,
+  initialAction: void | null | A,
+): [S, Dispatch<A>] {
+  const queue = hook.queue;
+
+  if (queue === null) {
+    return mountReducerImpl(hook, reducer, initialState, initialAction);
+  }
```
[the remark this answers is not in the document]
sebmarkbage: We can also just throw for the empty queue case since for the case where we append hooks to the end, we'll probably just want to detect that we've hit the end and then switching the dispatcher.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -859,19 +916,108 @@ function useEffectImpl(fiberEffectTag, hookEffectTag, create, deps): void {
...
+
+function mountEffect(
+  create: () => mixed,
+  deps: Array<mixed> | void | null,
+): void {
   if (__DEV__) {
-    currentHookNameInDev = null;
+    currentHookNameInDev = 'useEffect';
+  }
+  const hook = mountWorkInProgressHook();
```
sebmarkbage: Let's move all of these `mountWorkInProgressHook()` calls into `mountEffectImpl()`. That way this is just one function call indirections instead of two.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -859,19 +916,108 @@ function useEffectImpl(fiberEffectTag, hookEffectTag, create, deps): void {
...
+}
+
+function updateEffect(
+  create: () => mixed,
+  deps: Array<mixed> | void | null,
+): void {
+  if (__DEV__) {
+    currentHookNameInDev = 'useEffect';
   }
+  const hook = updateWorkInProgressHook();
```
sebmarkbage: Same thing as in the mount case. Let's put it in `updateEffectImpl()` so it's just one function call in here. (Later we can play around with bound functions.)

packages/react-reconciler/src/ReactFiberHooks.js
sebmarkbage: Alternatively, inline mountEffectImpl since it's so small.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -554,212 +594,234 @@ function basicStateReducer<S>(state: S, action: BasicStateAction<S>): S {
...
-  return useReducer(
-    basicStateReducer,
-    // useReducer has a special case to support lazy useState initializers
-    (initialState: any),
-  );
+  return readContext(context, observedBits);
 }
 
-export function useReducer<S, A>(
+function mountReducerImpl<S, A>(
```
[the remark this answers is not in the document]
acdlite: The weird inner branch didn't concern me because I figured it would go away once we change the lazy init API to the new one, but you're right, this is small enough that we should just inline it.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -788,219 +823,236 @@ function pushEffect(tag, create, destroy, deps) {
...
   deps: Array<mixed> | void | null,
 ): T {
-  if (__DEV__) {
-    currentHookNameInDev = 'useMemo';
-  }
-  currentlyRenderingFiber = resolveCurrentlyRenderingFiber();
-  workInProgressHook = createWorkInProgressHook();
-
+  const hook = mountWorkInProgressHook();
   const nextDeps = deps === undefined ? null : deps;
```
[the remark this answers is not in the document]
gaearon: Theoretically — could later make a smart compiler insert them automatically.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1120,12 +1168,317 @@ function dispatchAction<S, A>(
...
+      }
+      return mountLayoutEffect(create, deps);
+    },
+    useMemo<T>(create: () => T, deps: Array<mixed> | void | null): T {
+      currentHookNameInDev = 'useMemo';
+      if (hooksAccessForbiddenInDEV) {
+        warnInvalidHookAccess();
+      }
+      const prevHooksAccessForbiddenInDEV = hooksAccessForbiddenInDEV;
+      hooksAccessForbiddenInDEV = true;
```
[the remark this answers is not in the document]
acdlite: > Why not switch the dispatcher back to the ContextOnlyDispatcher?

We would have to do the switch in prod, too, which means adding additional code before and after every instance of those Hooks for something that will never happen if you use the Hooks lint rule, and will likely lead to a crash regardless.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1250,7 +1234,8 @@ function dispatchAction<S, A>(
       const eagerReducer = queue.eagerReducer;
       if (eagerReducer !== null) {
         if (__DEV__) {
-          ReactCurrentDispatcher.current = InvalidHooksDispatcherOnUpdateInDEV;
+          hooksAccessForbiddenInDEV = true;
```
sebmarkbage: One of the motivations for this refactor in the first place was to get rid of additional flags like this that is difficult to keep track of whether they're all assigned together in the same place and how the relate to assignments of the dispatcher. By adding this in again, I've now lost confidence that I understand these pairs again.

Can we make it just part of the dispatcher again?

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
sebmarkbage: The top comment is not addressed. The array tuple should be in the outer most function.

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
sebmarkbage: Ugh. GitHub showed me an old commit. Looks fixed it.

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
sebmarkbage: Ugh. GitHub showed me an old commit. Looking again...

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: I went with that approach originally but it became a problem after I made the change to switch back to the mount dispatcher once we reach the end of the list

https://github.com/facebook/react/blob/5a739725cd126e9cb1353279c64d46617712b3b7/packages/react-reconciler/src/ReactFiberHooks.js#L567-L572

because now the `finally` block will reset the dispatcher back to the old one.

An alternative would be to wrap each call site with a separate `try/finally` block but that leads to lots of forking in the code because you have to store a reference to the previous dispatcher. And it still suffers from the same problem because the nested hook could mutate the dispatcher — which is perhaps fine because that's a buggy code path, anyway, but it's a deviation from production behavior.

`hooksAccessForbiddenInDEV` seemed like the best trade off, and I don't think it's as bad as the previous boolean because it's completely localized within the DEV block that defines the DEV dispatcher.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -554,212 +584,210 @@ function basicStateReducer<S>(state: S, action: BasicStateAction<S>): S {
...
+  const dispatch: Dispatch<A> = (queue.dispatch = (dispatchAction.bind(
+    null,
+    // Flow doesn't know this is non-null, but we do.
+    ((currentlyRenderingFiber: any): Fiber),
+    queue,
+  ): any));
+  return [hook.memoizedState, dispatch];
+}
+
+function updateReducer<S, A>(reducer: (S, A) => S): [S, Dispatch<A>] {
```
sebmarkbage: This function gets exposed to always be called with three args. I'm not sure how much it matters in practice but just to be safe, we should try to ensure that the arg count is consistent so let's add the unused args here and pass them in to ensure type consistency. Same thing with useState...

packages/react-reconciler/src/ReactFiberHooks.js
sebmarkbage: I don't like that we're adding the progressive enhancement constraint before we need it. Isn't the simple fix here to just disable that feature and not reset it?

It seems like a problematic pattern in general if we can't use try/finally in that way since we're now deviating from the "algebraic effects" approach.

acdlite: > It seems like a problematic pattern in general if we can't use try/finally in that way since we're now deviating from the "algebraic effects" approach.

I don't think is a problem in general, it's only a problem because of the unsupported nested mutation, which is always a bug and only exists because we don't want to add a redundant check to the prod version.

acdlite: The other thing I considered was putting `createWorkInProgress` on the Dispatcher.

sebmarkbage: Once we have released the first version, we have plenty of time to figure out how we could structure the progressive enhancement case in a clever optimal way.

sebmarkbage: For real progressive enhancement we probably need a way to validate that it is a legit usage of expanding properties, e.g. by inserting a special primitive that when called creates a new tree. But we might also need the nested case. So I'm not sure this solution works anyway.

matthargett:
is this meant to merge before 16.8 publish?

matthargett:
Is this meant to merge for 16.8?

gaearon:
@matthargett 

Yes, we plan to get this and a few other PRs in before a stable version. If possible, I would like to ask you to refrain from asking questions about release schedule on a PR itself. It creates noise in the code review discussion and isn’t directly relevant to the PR. Instead, an appropriate venue might be a new issue, or a Twitter thread. Thank you! (There is no need to reply — please let’s keep this and other PR discussions focused on code.)

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: Resolved via chat: We don't support progressive enhancement yet anyway so let's throw and follow up once we get to that feature.

NE-SmallTown: @acdlite Could I ask what is "progressive enhancement" in this context?

Issue #14705, opened 2019-01-26
Context leaks to later renders when render stream destroyed early


overlookmotel:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**What is the current behavior?**

If a stream created with `.renderToNodeStream()` is destroyed before the render completes, memory can be retained unnecessarily by Contexts which were active at the time.

It's not a memory leak per se as memory use is not unbounded. There's a limit on number of threads in use and context slots are reused, so memory use can't increase endlessly.

Would only make a noticeable impact on an application which:

* renders using `.renderToNodeStream()` or `.renderToStaticMarkup()`
* stores large objects in Context (e.g. large data sets from fetch requests)
* experiences a spike in traffic which then recedes
* destroys a lot of streams prematurely (or has errors thrown in rendering which causes streams to be destroyed)

In the above case, when the spike hits, many threads are created for concurrent renderers. Each thread writes data into a slot of the Contexts it uses. When the threads are destroyed before the renders complete, the Context slots are not cleared, and so left filled with data which is no longer useful, but cannot be freed by GC.

When the spike recedes, this memory continues to be retained.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

I'll submit a PR with failing test case and fix shorty.

**What is the expected behavior?**

When stream is destroyed, it should clean up its resources.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

Issue introduced in 961eb65b4ba5de6bbfb6b8a075a924c284541177 which is in 16.7.0.

overlookmotel:
PR to fix: https://github.com/facebook/react/pull/14706

overlookmotel:
Actually, I've noticed that this bug can also cause one render's context to leak into another later render. This is likely to be more common and more serious than the original effect outlined above.

Steps to reproduce:

* Render 1 uses `.renderToNodeStream()`, it gets `threadID` 1
* Render 1 renders `<MyContext.Provider value={'XXX'}>`
* Render 1 stream is destroyed before `</MyContext.Provider>` is reached in the render
* `MyContext` slot for `threadID` 1 still contains 'XXX'
* Render 2 later on also gets allocated `threadID` 1
[...]

PR #14706, opened 2019-01-26
Fix react-dom/server context leaks when render stream destroyed early

overlookmotel:
Fix and test case for #14705.

Note: This makes `.popProvider`'s `provider` argument optional. It's only used in dev mode anyway, but I'm not sure if this is the wrong thing to do. Please excuse me - I've never used Flow before.

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRenderer.js
+++ b/packages/react-dom/src/server/ReactPartialRenderer.js
@@ -715,6 +715,7 @@ class ReactDOMServerRenderer {
   destroy() {
     if (!this.exhausted) {
       this.exhausted = true;
+      this.clearProviders();
       freeThreadID(this.threadID);
     }
   }
@@ -748,7 +749,7 @@ class ReactDOMServerRenderer {
     context[threadID] = provider.props.value;
   }
 
-  popProvider<T>(provider: ReactProvider<T>): void {
+  popProvider<T>(provider?: ReactProvider<T>): void {
     const index = this.contextIndex;
     if (__DEV__) {
       warningWithoutStack(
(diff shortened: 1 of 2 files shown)


overlookmotel:
2nd tested added for https://github.com/facebook/react/issues/14705#issuecomment-457883547.

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -776,6 +777,16 @@ class ReactDOMServerRenderer {
     context[this.threadID] = previousValue;
   }
 
+  clearProviders(): void {
+    while (this.contextIndex > -1) {
+      if (__DEV__) {
```
gaearon: Can you write the loop inline with minimal work it needs to do? For example reading previous value seems unnecessary except for the last one. This would also avoid the overloading you had to add to `popProvider`.

overlookmotel: I've pushed another commit which does as you request.

However, the previous value does need to be restored for each Provider, as each Context's current value is stored separately on that Context object.

Well, if there were several nested Providers for the same Context, then yes you could only restore just the last previous value. But I don't think this would be more performative, as `clearProviders()` would need to loop over the array twice - once to identify any repeated contexts, and then a 2nd time to restore whatever was the oldest previous value for each Context.

I think it's probably cheaper to redundantly write the value multiple times in these cases.

But if you want further changes, please just say so and I'll implement.

overlookmotel: NB This leaves `contextStack` and `contextValueStack` unaltered, and so retaining memory. But I guess since the `PartialRenderer` instance has been destroyed, it will be dereferenced and garbage collected shortly anyway.

overlookmotel:
Thank you!

PR #14717, opened 2019-01-29
Partial Hydration

sebmarkbage:
This adds a mechanism for partially hydrating a server rendered result while other parts of the page are still loading the code or data. This means that you can start interacting with parts of the screen while others are still hydrating.

## Model

In this model you always have to hydrate the root content first because it is what provides props to the children, which can be of arbitrary complexity. The model assumes that the root of the app is designed to be relatively shallow and then each abstraction gets progressively more complex the deeper it gets. To become interactive faster, components in the tree can themselves use progressive enhancement to add more complexity after initial hydration.

## API

This mechanism works on top of the `<Suspense fallback={...}>` API. With this PR, when we try to hydrate a Suspense component, we will immediately bail out and skip past it. We'll leave the server rendered content in place. Once we've finished a hydrating of everything not in a Suspense boundary, we'll commit it. Then the next frame we'll continue to hydrate the next Suspense boundary and so on. This is a breadth first hydration. This happens even if nothing suspends. 

The reason for the breadth first hydration is because we need to fully complete the parent before we can make any updates to children. Once we have committed a parent, we can now prioritize any of the Suspense "holes" left dehydrated inside it - independently.

The reason this works with Suspense boundaries is that the parents already have to be resilient to the children not being fully resolved, but also because we can at any point trigger the fallback state to return to a consistent state.

If the Suspense component rerenders with new props or new context as input, before we've fully hydrated, we have two problems:

1) We can no longer safely hydrate the subtree. React has a strong requirement that the initial render behaves just as it did on the server. This is a tradeoff that lets us avoid a lot of metadata added to the HTML. We can't make changes to a subtree without first rendering the initial state. After that we can make updates. In theory, we could store a snapshot of all contexts and props to solve this problem.

2) However, semantically, changes to props or Context should change whatever was rendered. E.g. if that switch is from dark to light mode, then we can't just leave the existing content in place. Similar things happen with certain layouts or media queries.

Therefore, the semantics here is that if that happens, then we'll delete the existing content and rerender it from scratch. If that suspends, we'll show the fallback.

This is an undesirable experience but reasonable compromise. To avoid this the product code must:

- Ensure that unrelated updates bail out early. E.g. using `memo` or `shouldComponentUpdate`.
- Avoid updates to top level Contexts.
- If that can't be avoided, use a low priority update in Concurrent Mode, and long suspense durations. This will delay this scenario for as long as possible.

## Quirks

This solution is susceptible to tearing issues, common in Flux stores, just like Concurrent Mode in general. E.g. if the store is mutated before the next level is hydrated, then we'll try to hydrate it with the wrong initial state. Therefore, stores need to be able to save a snapshot of their initial state for the duration of the hydration.

In the current version of this PR, hydration of suspense boundaries always gets scheduled as concurrent. I'm not sure if a non-concurrent mode of this even makes sense.

Hoisting state up to the root can be problematic because as these update they will pass their value down and rerender components that may not have fully hydrated yet which will put them in their fallback state. The key is to make any such state have a long expiration time and long suspense time so that if it happens, we have time to hydrate it beforehand. High-pri state should be local to components and not rerender at the top. Updates to top level state placed in Context will force the fallback state of the whole tree since it can affect everything and needs to be managed carefully.

For these reasons, it is important to carefully design the shell of the app so that different parts can operate independently for at least some period of time.

## Progress

Left to do in this PR:

- [x] Wrap in feature flag
- [x] Delete the content when any parent Context updates.
- [x] Deal with Suspense boundaries without fallback defined.

Follow up 1:

- [ ] If props or context has changed, first try hydrating at higher priority in case that lets us hydrate.
- [ ] Mark fallback content in the server rendered content. Wait to try to hydrate it until it switches to real content.
- [ ] Gracefully handle sync mode (including strict mode) or disable it completely.
- [ ] Add warnings for failing to hydrate suspense boundaries.

Follow up 2:

- [ ] Add mechanism to defer and replay events that happen in dehydrated trees.
- [ ] Use deferred events as hints for which trees to prioritize.
- [ ] If a deferred event can't be replayed on a hydrated tree within an expiration time, delete the content and show fallback.
- [ ] Consider hydrating each level at a higher priority than "offscreen", or let hydration take priority over updates to the same trees.


Ikhbar-Kebaa:
Hye

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -197,6 +240,8 @@ function throwException(
             earliestTimeoutMs = timeoutPropMs;
           }
         }
+      } else if (workInProgress.tag === DehydratedSuspenseComponent) {
+        // TODO
```
sebmarkbage: @acdlite I'm not sure exactly what I need to do here if anything. The timeout doesn't really come into play since failure to hydrate a suspended boundary doesn't timeout until it gets upgraded.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -432,6 +436,49 @@ export function removeChildFromContainer(
...
+    if (nextNode && nextNode.nodeType === COMMENT_NODE) {
+      let data = ((nextNode: any).data: string);
+      if (data === SUSPENSE_END_DATA) {
+        if (depth === 0) {
+          parentInstance.removeChild(nextNode);
+          return;
+        } else {
+          depth--;
+        }
+      } else if (nextNode === SUSPENSE_START_DATA) {
```
acdlite: I think this should be

```suggestion
      } else if (data === SUSPENSE_START_DATA) {
```

Indicates a test for clearing nested Suspense boundaries is missing?

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -495,15 +542,27 @@ export function canHydrateTextInstance(
...
 }
 
+export function canHydrateSuspenseInstance(
+  instance: HydratableInstance,
+): null | SuspenseInstance {
+  if (instance.nodeType !== COMMENT_NODE) {
+    // Empty strings are not parsed by HTML so there won't be a correct match here.
+    return null;
+  }
+  // This has now been refined to a text node.
```
acdlite:
```suggestion
  // This has now been refined to a suspense node.
```

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1598,6 +1612,81 @@ function updateSuspenseComponent(
...
+    return null;
+  }
+  // We use childExpirationTime to indicate that a child might depend on context, so if
+  // any context has changed, we need to treat is as if the input might have changed.
+  const hasContextChanged = current.childExpirationTime >= renderExpirationTime;
+
+  const prevProps = current.memoizedProps;
+  const nextProps = workInProgress.pendingProps;
+
+  if (hasContextChanged || prevProps !== nextProps) {
```
acdlite: We already compare `prevProps` to `nextProps` at the beginning of `beginWork`. And if they are different we set the module-level boolean `didReceiveUpdate` to true. (This approach would also account for legacy context changing, even though that's deprecated in strict mode anyway and I'm guessing partial hydration will only work in strict mode?)

packages/react-reconciler/src/ReactFiberSuspenseComponent.js
```diff
@@ -24,3 +24,13 @@ export function shouldCaptureSuspense(workInProgress: Fiber): boolean {
...
 }
+
+export function shouldCaptureDehydratedSuspense(
+  workInProgress: Fiber,
+): boolean {
+  // In order to capture, the Suspense component must have a fallback prop.
+  if (workInProgress.memoizedProps.fallback === undefined) {
+    return false;
+  }
+  return true;
```
acdlite: Should be able to replace calls to this function with `true` per our conversation

packages/react-reconciler/src/ReactFiberUnwindWork.js
acdlite: I don't think you need to do anything in this branch, but maybe add a comment explaining why: a dehydrated boundary should never time out. The only way the boundary would time out is if it's upgraded to a normal Suspense boundary.

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -331,6 +353,41 @@ function throwException(
...
+        workInProgress.expirationTime = renderExpirationTime;
+        return;
+      } else if (
+        enableSuspenseServerRenderer &&
+        workInProgress.tag === DehydratedSuspenseComponent &&
+        shouldCaptureDehydratedSuspense(workInProgress)
+      ) {
+        attachPingListener(root, renderExpirationTime, thenable);
+
+        // Since we already have a current fiber, we can eagerly add a ping listener.
```
acdlite:
```suggestion
        // Since we already have a current fiber, we can eagerly add a retry listener.
```

The terminology for "ping" versus "retry" isn't great but that's what we're using everything else

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -331,6 +353,41 @@ function throwException(
...
         renderDidSuspend(root, absoluteTimeoutMs, renderExpirationTime);
 
+        workInProgress.effectTag |= ShouldCapture;
+        workInProgress.expirationTime = renderExpirationTime;
+        return;
+      } else if (
+        enableSuspenseServerRenderer &&
+        workInProgress.tag === DehydratedSuspenseComponent &&
+        shouldCaptureDehydratedSuspense(workInProgress)
+      ) {
```
acdlite: Since this branch doesn't suspend (it doesn't call `renderDidSuspend`) I would expect React to keep rendering the same level over and over until the promise resolves. Is that what's happening?

sebmarkbage: No. What happens is that only the first path schedules remaining work at "Never" expiration time. Then if it throws, it doesn't suspend but it also doesn't leave any work on it. Instead it commits. Then it waits for the retry. The retry gets scheduled at normal priority. If that update also throws a promise, then it commits in the dehydrated state again and waits for the retry.

acdlite: I see so when something suspends inside a dehydrated Suspense boundary it always bails out and clears the expiration time. The ping/retry adds the expiration time back. There’s no need to suspend the commit because it’s not blocking anything.

sebmarkbage:
The bad case in this solution happens when a props/context changes which then causes that update to suspend. This triggers the fallback state. In the fully client-side solution, we keep the state while we load the missing data. However, in this case we'll delete the server rendered nodes and lose their state.

With one of the followups this won't happen as long as we can hydrate before the timeout happens because we can hydrate right before committing the suspended state.

It might seems like we should be able to hide the server rendered content and then hydrate it and show it. That's slightly different because if we commit the fallback state, we have now lost whatever was "current" right before that. We've lost the props and the context of all parent components which we need to hydrate.

In theory we could do something advanced in this case like snapshotting the props and values of all contexts. However, we'd have to keep this indefinitely and rely on that these snapshots are the only sources of data and that they're fully immutable. This adds a new constraint, that we're able to render states older than current.

Even if we could, we don't want to replay clicks when a fallback was rendered between. Since it's not seamless anyway. The only thing we'd gain is the ability to preserve state in uncontrolled forms.

This doesn't seem worth it.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1882,6 +1974,15 @@ function beginWork(
           }
           break;
         }
+        case DehydratedSuspenseComponent: {
+          if (enableSuspenseServerRenderer) {
+            // We know that this component will suspend again because if it has
+            // been unsuspended it has committed as a regular Suspense component.
+            // If it needs to be retried, it should have work scheduled on it.
+            workInProgress.effectTag |= DidCapture;
+            break;
```
bvaughn: What is this break statement doing inside of the `if` block?

PR #14720, opened 2019-01-29
Move Hook mismatch warning to first mismatch site

acdlite:
Allows us to localize the warning logic in one place.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -98,10 +98,6 @@ type HookType =
   | 'useImperativeHandle'
   | 'useDebugValue';
 
-// the first instance of a hook mismatch in a component,
-// represented by a portion of its stacktrace
-let currentHookMismatchInDev = null;
-
 let didWarnAboutMismatchedHooksForComponent;
 if (__DEV__) {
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -541,6 +488,16 @@ function createWorkInProgressHook(): Hook {
       currentHook = currentHook !== null ? currentHook.next : null;
     }
   }
+
+  if (__DEV__) {
+    if (
+      currentHook !== null &&
```
acdlite: Won't need to check if `currentHook` is null once #14701 lands

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -98,10 +98,6 @@ type HookType =
   | 'useImperativeHandle'
   | 'useDebugValue';
 
-// the first instance of a hook mismatch in a component,
-// represented by a portion of its stacktrace
-let currentHookMismatchInDev = null;
```
threepointone: we were including the trace so peeps could tell which (possibly custom hook) was the offending one. fine to remove?

threepointone: right, of course the warning would have its own trace. thanks!

PR #14723, opened 2019-01-30
Improve Reducer Hook's lazy init API

acdlite:


gaearon:
For posterity what’s the actual Flow signature with overload that we intend?

acdlite:
```js
declare function useReducer<S, A>(
  reducer: (S, A) => S,
  initialState: S,
): [S, Dispatch<A>];

declare function useReducer<S, I, A>(
  reducer: (S, A) => S,
  initialArg: I,
  init: (I) => S,
): [S, Dispatch<A>];
```

acdlite:
@sebmarkbage Good call, I edited my comment accordingly

bvaughn:
```js
declare function useReducer<S, A>(
  reducer: (S, A) => S,
  initialState: S,
  init: void,
): [S, Dispatch<A>];
```

What use case is this variation intended for?

gaearon:
Probably:

>That way it allows undefined to be passed. That’s useful when you’re forwarding arguments in wrappers like we do.

Like if you wrap `useReducer` from higher level API that may or may not receive an `init` argument I guess?

bvaughn:
Yeah maybe that's it. I wouldn't initially have expected that to require its own explicit variation.

Jessidhia:
Effectively, this acts as if the third argument defaults to the identity function? 🤔 

```ts
declare function useReducer<S, A>(
  reducer: (state: S, action: A) => S,
  initialState: S,
  initializer?: undefined
): [S, Dispatcher<A>]
declare function useReducer<S, A, I>(
  reducer: (state: S, action: A) => S,
  initializerArg: I,
  initializer: (arg: I) => S
): [S, Dispatcher<A>]
```

PR #13387, opened 2018-08-14
cloneElement: Added checker for undefined props.

aseem191:
This PR is an addition/feature that responds to https://github.com/facebook/react/issues/13381. It causes cloneElement to automatically skip past props whose values are undefined.

PR #14724, opened 2019-01-30
cloneElement: Added checker for undefined props

buoyantair:
Extending PR #13387  with revised changes.


buoyantair:
@matthargett yes I need to write tests yet..

buoyantair:
@matthargett Hi, I added the test. Do review :)

buoyantair:
Bump!

gaearon:
Thanks for the PR. Sorry for radio silence. While this is unfortunate, fixing this would be a breaking change and I don't think it's worth the churn at this point.

Issue #14708, opened 2019-01-26
hooks: useContext with useState not updating


leonardodino:
**Do you want to request a *feature* or report a *bug*?**

seems it's a **bug**. 😕 

**What is the current behavior?**

Nested context provider and `useContext` hooks seems to be conflicting, updates get discarded.

**What is the expected behavior?**

When connecting to a context, it should update whenever it's `value` changes.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

- **react**: `18.8.0-alpha.1` (also reproduced on `16.7.0-alpha.0`)
- **browser**: `chrome 71`
- **os**: macOS Sierra

---

### more details

While working on a cleanup of a localStorage "connection",
I tried to mix 2 articles ([`[1]`](https://reactjs.org/docs/hooks-faq.html#how-to-avoid-passing-callbacks-down) & [`[2]`](https://reactjs.org/docs/context.html#updating-context-from-a-nested-component)) from the official react documentation, I've implemented it with hooks, but the value seems not to be passing through.

I've put up a streamlined demo on [codesandbox `[3]`](https://codesandbox.io/s/0yzjr8vnrv).

The actual implementation is only a couple of lines more (parsing it from and stringifying it to JSON).

Workarounds that I found:
- If I create a new function on each render around the `setValue` function, it actually works. 
  - but this goes against the advice on [`[1]`](https://reactjs.org/docs/hooks-faq.html#how-to-avoid-passing-callbacks-down) about avoiding creating  new values.
- Migrate it to a class and use `componentDidUpdate` instead of `useEffect`.
  - I'm actually using this right now, as it works. Including saving a reference to the function in the state.

---

Is there anything that shouldn't work on the code below? the effect gets triggered with the changes,
but the value doesn't get updated on the components that consume via hook. see repro code [`[3]`](https://codesandbox.io/s/0yzjr8vnrv)

```javascript
const createLocalStorage = key => {
    const initialValue = localStorage.getItem(key)
    const ValueContext = createContext(initialValue)
    const SetterContext = createContext(() => {})

    const useStorage = () => [ValueContext, SetterContext].map(useContext)

    const Provider = ({children}) => {
        const [value, setValue] = useState(initialValue)

        useEffect(
            () => {
                console.log('effect', value)
                localStorage.setItem(key, value)
            },
            [value],
        )

        return (
            <ValueContext.Provider value={value}>
                <SetterContext.Provider value={setValue}>
                    {children}
[...]

PR #14729, opened 2019-01-30
Warn when second argument is passed to useContext

gaearon:
This leaves us the ability to later change its meaning. It also warns early avoids confusing cases like `[].map(useContext)` which accidentally supplies `observedBits` (https://github.com/facebook/react/issues/14708#issuecomment-457951903).

Internal `readContext` on the dispatcher doesn't have the warning.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
@@ -613,8 +613,114 @@ describe('ReactDOMServerHooks', () => {
   });
 
   describe('useContext', () => {
+    itThrowsWhenRendering(
+      'if used inside a class component',
+      async render => {
+        const Context = React.createContext({}, () => {});
+        class Counter extends React.Component {
+          render() {
+            let [count] = useContext(Context);
+            return <Text text={count} />;
+          }
+        }
+
+        return render(<Counter />);
+      },
+      'Hooks can only be called inside the body of a function component.',
+    );
+  });
+
+  itRenders(
+    'can use the same context multiple times in the same function',
+    async render => {
+      const Context = React.createContext({foo: 0, bar: 0, baz: 0});
+
+      function Provider(props) {
+        return (
+          <Context.Provider
+            value={{foo: props.foo, bar: props.bar, baz: props.baz}}>
+            {props.children}
+          </Context.Provider>
+        );
+      }
+
+      function FooAndBar() {
+        const {foo} = useContext(Context);
+        const {bar} = useContext(Context);
+        return <Text text={`Foo: ${foo}, Bar: ${bar}`} />;
+      }
+
+      function Baz() {
+        const {baz} = useContext(Context);
+        return <Text text={'Baz: ' + baz} />;
+      }
+
+      class Indirection extends React.Component {
+        render() {
+          return this.props.children;
+        }
+      }
+
+      function App(props) {
+        return (
+          <div>
+            <Provider foo={props.foo} bar={props.bar} baz={props.baz}>
+              <Indirection>
+                <Indirection>
+                  <FooAndBar />
+                </Indirection>
+                <Indirection>
+                  <Baz />
+                </Indirection>
+              </Indirection>
+            </Provider>
+          </div>
+        );
+      }
+
+      const domNode = await render(<App foo={1} bar={3} baz={5} />);
+      expect(clearYields()).toEqual(['Foo: 1, Bar: 3', 'Baz: 5']);
+      expect(domNode.childNodes.length).toBe(2);
+      expect(domNode.firstChild.tagName).toEqual('SPAN');
+      expect(domNode.firstChild.textContent).toEqual('Foo: 1, Bar: 3');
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
```diff
@@ -613,8 +613,114 @@ describe('ReactDOMServerHooks', () => {
   });
 
   describe('useContext', () => {
+    itThrowsWhenRendering(
```
gaearon: The diff confused git. This is not a new test.

All I did is move the `useContext()` test with bits into `readContext()` section below.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
```diff
@@ -613,8 +613,114 @@ describe('ReactDOMServerHooks', () => {
...
+      expect(clearYields()).toEqual(['Foo: 1, Bar: 3', 'Baz: 5']);
+      expect(domNode.childNodes.length).toBe(2);
+      expect(domNode.firstChild.tagName).toEqual('SPAN');
+      expect(domNode.firstChild.textContent).toEqual('Foo: 1, Bar: 3');
+      expect(domNode.lastChild.tagName).toEqual('SPAN');
+      expect(domNode.lastChild.textContent).toEqual('Baz: 5');
+    },
+  );
+
+  itRenders('warns when bitmask is passed to useContext', async render => {
```
gaearon: This is a new test.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
```diff
@@ -613,8 +613,114 @@ describe('ReactDOMServerHooks', () => {
...
+    function readContext(Context, observedBits) {
+      const dispatcher =
+        React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
+          .ReactCurrentDispatcher.current;
+      return dispatcher.readContext(Context, observedBits);
+    }
+
     itRenders(
-      'can use the same context multiple times in the same function',
+      'can read the same context multiple times in the same function',
```
gaearon: This is existing test which I made use `readContext` instead.

packages/react-reconciler/src/__tests__/ReactNewContext-test.internal.js
```diff
@@ -1300,81 +1324,7 @@ describe('ReactNewContext', () => {
   });
 
   describe('readContext', () => {
-    // Context consumer bails out on propagating "deep" updates when `value` hasn't changed.
```
gaearon: This test just moved below in the diff.

packages/react/src/ReactHooks.js
```diff
@@ -24,10 +24,23 @@ function resolveDispatcher() {
 
 export function useContext<T>(
   Context: ReactContext<T>,
-  observedBits: number | boolean | void,
+  unstable_observedBits: number | boolean | void,
 ) {
   const dispatcher = resolveDispatcher();
   if (__DEV__) {
+    warning(
```
gaearon: This is the actual change.

Yurickh:
Did you meant `useContext` in the PR's title?

benwiley4000:
@gaearon Does this make the `unstable_observedBits` argument to `useContext` *more* unstable than the `unstable_observedBits` prop to `createContext(...).Consumer`? Is there a chance of either being removed in a React 16.x release?

gaearon:
Yes there’s always a chance of unstable APIs getting removed.

It doesn’t make it more unstable than the Consumer prop version. It’s just easier to pass by accident and get weird results. Which is why this one warns.

benwiley4000:
Got it, thanks for the clarification!

benwiley4000:
@gaearon So I have another question -- is this meant to stop anyone from using `unstable_observedBits` in a library? Often with unstable APIs it seems understood that library maintainers may use it, but with this warning anyone consuming my library and using a development build of React will get this warning over and over and they will probably freak out. I'm considering adding a hook export to a library I maintain, but I don't want to do that if I can't use `unstable_observedBits` The warnings in the console I'm seeing add way too much noise.

windmaomao:
The problem I see is that i couldn't get the second flag working with a `memo`.  

I can see this flag is being used correctly to bail out a case when the bits not matched `(dependency.observedBits & changedBits) !== 0`.  But in general `Context.Provider` renders every children. Therefore unless i put a `memo` or make a custom `memoChild` to wrap the first child under the provider. I couldn't actually get any consumer bailout.

```
import { memo } from 'react'

const equalWithoutChildren = (prev, next) => {
  for (let k in prev) {
    if (k === 'children') continue
    if (prev[k] !== next[k]) return false
  }
  return true
}

const memoChild = (Component) => {
  const Memo = memo(({ children, ...props }) => {
    return <Component children={children} {...props} />
  }, equalWithoutChildren)

  return Memo
}

export default memoChild
```

Please let me know if i'm doing anything wrong. Also is this `useContextSelector` about to address? to make the context a bit cheaper to use? Thank you teams.

PR #14744, opened 2019-02-01
expose `TestUtils.act()` for batching actions in tests

threepointone:
### `act()` for testing react components
---

```js
// React DOM
import { act } from 'react-dom/test-utils';

// React Native
import { act } from 'react-test-renderer';
```

React, like other libraries, doesn't guarantee synchronous ordering and execution of _it's_ own work. eg - calling `this.setState()` inside a class doesn't actually set the component state immediately. In fact, it might not even update the state of the component in the same 'tick'. This isn't a problem when it comes to 'users'; they interact with React surfaces asynchronously, giving react and components plenty of time to 'resolve' to particular states.

However, tests are unique in that people write code, usually sequential, to interact with React components, and some assumptions they make won't hold true. Consider - with React's new `useEffect` hook, a component can run a side effect as soon as it 'starts up'. As a contrived example -

```jsx
function App(props) {
  useEffect(() => {
    props.callback();
  });
  return null;
}
```

Let's say you write a test for it like so -

```jsx
let called = false;
render(
  <App
    callback={() => {
      called = true;
    }}
  />,
  document.body
);
expect(called).toBe(true); // this fails, it's false instead!
```

The test would fail, which seems counterintuitive at first. But the docs explain it - "The function passed to useEffect will run after the render is committed to the screen." So while the effect has been queued into it's scheduler, it's up to React to decide when to run it. React only guarantees that it'll be run before the browser has reflected changes made to the dom to the user (ie - before the browser has 'painted' the screen)

You may be tempted to refactor this like so -

```jsx
// don't do this!
function App(props) {
  useLayoutEffect(() => {
    props.callback();
  });
  return null;
}
```

This would "work" in that your test would pass, but that's because you've explicitly using a render blocking effect where it possibly wasn't required. This is bad for a number of reasons, but in this context, it's bad because we're changing product behavior just to fix a test.

What can we do better?

Well, React could expose a helper, let's call it `act`, that guarantees the sequential execution of it's update queue. Let's rewrite the test -

```jsx
let called = false;
act(() => {
  // this 'scope' is safe to interact with the React component,
  // rendering and clicking as you please
  render(
    <App
      callback={() => {
        called = true;
      }}
    />,
    document.body
  );
});
// at this point, we can guarantee that effects have been executed,
// so we can make assertions
expect(called).toBe(true); // this passes now!
act(() => {
  // further interactions, like clicking buttons, scrolling, etc
});
// more assertions
```

Note - React still doesn't synchronously execute the effect (so you still can't put your `expect` statements _inside_ `act`), but it does guarantee to execute all enqueued effects right after `act` is called.

This is a nice mental model for separation of concerns when testing components - "React, here's a batch of code I'd like you to run at one go", followed by more code to test what React has actually 'done'.


threepointone:
cc @kentcdodds ^

threepointone:
moving it into TestUtils

gaearon:
Possible follow up: warn if we detect setState on a Hook in jsdom environment?

sebmarkbage:
If we're going to warn for setState on Hook outside of this thing, we should do it before release.

threepointone:
Noted. I'll send a followup in a bit.

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -611,6 +612,13 @@ function createPortal(
   return createPortalImpl(children, container, null, key);
 }
 
+// a helper for tests to scope their actions on the dom
+// batches and flushes effects and updates
+function batchedInteraction(callback: () => void) {
+  batchedUpdates(callback);
+  flushPassiveEffects();
+}
```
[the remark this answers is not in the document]
threepointone: agreed, I felt icky doing this. will change,

threepointone:
> We should bikeshed the name a bit more. interact is the wrong word

strong agree. it felt wrong right there. some options - `run`, `execute`, `batch`, `batchAndRun`. I would have also liked `do`

gaearon:
>I'm not sure I understand this.

We want to add a warning when you `setState` on a Hook-using component in jsdom outside of this "interact" scope. This is because you're usually testing the wrong behavior that doesn't occur in practice due to batching.

In fact this is already a problem in classes. `setState()` from test won't work like a real `setState()` in a class click handler. But it's too late to fix in classes since everybody does that. With Hooks we have a chance to explain this is bad, and point to the recommended solution.

sebmarkbage:
We shouldn't warn in all jsdom environments. Only test ones, such as jest. Sometimes jsdom can be used in some other esoteric use cases - e.g. to server render webcomponents.

sebmarkbage:
@threepointone Instead of detecting if we're inside "interact()", we can detect if we're batching updates or in concurrent mode. I.e. if we're in sync mode. It's unfortunate because we won't warn if you use `batchedUpdates()` which won't properly flush the effects but at least we catch the common case. That way we don't need to expose and new APIs from the ReactDOM bundle.

kentcdodds:
This makes sense. So correct me if I'm wrong but this basically means that if you interact with a component in any way that calls a state updater, you'll get a warning if it wasn't done within an `interact` callback (or whatever that ends up being called). That sounds good to me :+1: Makes the rules easier to follow. I'd love to play with this when it's ready!

gaearon:
>We shouldn't warn in all jsdom environments. Only test ones, such as jest

This is tricky because we don't have a way to detect. We can detect Jest by `global.it` or `global.expect` (or both). But not Ava which uses non-global helpers.

We could check `NODE_ENV`. But not everybody sets it. Also, we can't easily do this from inside our bundles because our build step would replace it. So it would need to be threaded through somehow.

sebmarkbage:
I suppose any jsdom server renderer should be using batchedUpdates anyway.

packages/react-dom/src/test-utils/ReactTestUtils.js
```diff
@@ -380,6 +380,11 @@ const ReactTestUtils = {
 
   Simulate: null,
   SimulateNative: {},
+
+  interact(callback: () => void) {
```
sebmarkbage: Drop the `inter`. Just call it `act`. It's cleaner.

sebmarkbage:
There is precedence in the `prepare`, `act`, `assert` style testing to call the phase where you are actually performing things "act" which is what you're meant to be doing in this callback. Followed by `expect()` calls later.

It's also short so it's not too bothersome to write all the time in all your tests.

packages/react-dom/src/test-utils/ReactTestUtils.js
```diff
@@ -380,6 +380,11 @@ const ReactTestUtils = {
...
+    ReactDOM.unstable_batchedUpdates(callback);
+    ReactDOM.render(null, document.createElement('div'));
```
sebmarkbage: Can we reuse a single node? Maybe put an inline React element instead so we don't bail out.

sebmarkbage:
I believe that in the nested case, these will flush at the outer `act(...)` boundary which makes sense. That way I can put these on the top level always, but I can also put them inside test utilities like "dispatch" helpers.

gaearon:
Canonical jsdom defection https://github.com/jsdom/jsdom/issues/1537#issuecomment-229405327

threepointone:
Question - Assuming the warning exists, how would I get this test to pass? 
```jsx
it('lets a ticker update', () => {
  function App(){
    let [toggle, setToggle] = useState(0)
    useEffect(() => {
      let timeout = setTimeout(() => {
        setToggle(1)
      }, 200);
      return () => clearTimeout(timeout)
    })
    return toggle
  }
  const el = document.createElement('div')
  act(() => {
    ReactDOM.render(<App />, el);
  })
  jest.advanceTimersByTime(250); // this warns!!!        
  expect(el.innerHTML).toBe("1")
})
```
for context - I wrote `ensureBatchingAndScheduleWork` in `ReactFiberScheduler`, that `dispatchAction` in `ReactFiberHooks` would call. Running this across our tests, I found a number of failures (because we use the pattern we want to prevent - getting a pointer to a hooks `setState` and calling it). filtering on only jsdom occurrences, that number got smaller, but led me to writing the above test. current work here - https://github.com/threepointone/react/commit/ac7416da8f8d570cab329c5a5769d543eeabd98f

threepointone:
one workaround is to isolate the render call and the timer advance, so this passes the test - 
```jsx
    act(() => {
      ReactDOM.render(<App />, container);
    });
    act(() => {
      jest.advanceTimersByTime(250); // this warns!!!  
    })
    
    expect(container.innerHTML).toBe('1');
```

(putting them both in the same act call _doesn't_ work, since the effect wouldn't have fired yet)

kentcdodds:
Good point @threepointone. I can see how this complicates the testing story quite a bit. Even experienced React engineers will struggle with writing tests for effects like this 🤔

threepointone:
@kentcdodds it might not be particularly bad for testing libraries, you could wrap your `render`/`fireEvent` with `act`.

kentcdodds:
Hmm... I like that idea. I wonder if there's a situation where you'd want to fire an event outside of act 🤔

I suppose if you did then we could expose a way for you to do so.

kentcdodds:
Hey, so would this work?

```js
let called = false;
// this 'scope' is safe to interact with the React component,
// rendering and clicking as you please
render(
  <App
    callback={() => {
      called = true;
    }}
  />,
  document.body
);
act(() => {});
// at this point, we can guarantee that effects have been executed,
// so we can make assertions
expect(called).toBe(true); // this passes now!
// further interactions, like clicking buttons, scrolling, etc
act(() => {});
// more assertions
```

Just exploring possibilities.

packages/react-test-renderer/src/ReactTestRenderer.js
```diff
@@ -17,8 +17,8 @@ import {
   updateContainer,
   flushSync,
   injectIntoDevTools,
+  batchedUpdates,
```
threepointone: is there a reason a generic batchedUpdates was exposed from TestRenderer?

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -65,6 +65,7 @@ import {
 import getComponentName from 'shared/getComponentName';
 import invariant from 'shared/invariant';
 import warningWithoutStack from 'shared/warningWithoutStack';
+import {canUseDOM} from 'shared/ExecutionEnvironment';
```
gaearon: It feels wrong to have any logic concerning the DOM (even checks) in the reconciler. It’s also why the warning false positives for the test renderer.

We should somehow make it specific to ReactDOM I think.

packages/react-dom/src/test-utils/ReactTestUtils.js
```diff
@@ -380,6 +384,11 @@ const ReactTestUtils = {
 
   Simulate: null,
   SimulateNative: {},
+
+  act(callback: () => void) {
+    ReactDOM.unstable_batchedUpdates(callback);
+    ReactDOM.render(actElement, actContainerElement);
```
gaearon: I then we should make it return result of calling batched updates (which by itself passes the callback retvalue).

This lets you write 

```js
const inst = act(() => ReactDOM.render(...)))
```

without awkward separate declaration.

packages/react-reconciler/src/ReactFiberScheduler.js
threepointone: I’d originally put the dom check inside ReactFiberHooks, I’ll move it back there. I’m thinking I’ll test the fiber’s internal instance and trst if it’s a dom element.

gaearon: ReactFiberHooks isn’t any better. It’s still in the reconciler. 

I meant to add a host config method that renderers can implement. Look at how server hydration warnings are done.

gaearon:
@kentcdodds I don’t understand what you’re trying to do with empty `act(() => {})` calls.

kentcdodds:
I was mostly just wondering if it would "work." Don't plan to do it.

threepointone:
Updated to use the HostConfig instead (thanks dan! til), avoiding a bunch of expensive tests and complexity. I now need to figure out how to get flow to pass for all the renderers.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -647,3 +648,20 @@ export function didNotFindHydratableTextInstance(
...
+
+// in a dom-like test environment, we want to warn if dispatchAction()
+// is called outside of a batchedUpdates/TestUtils.act(...) call.
+let ensureBatchedDispatchedActions = false;
+
+if (__DEV__) {
+  if (
+    canUseDOM &&
+    (navigator.userAgent.indexOf('Node.js') >= 0 ||
+      navigator.userAgent.indexOf('jsdom') >= 0)
```
SimenB: This _might_ be set to something else (https://jestjs.io/docs/en/configuration#testenvironmentoptions-object).

Might be a good idea to also check for some sign it's Jest (e.g. Jest can start to do `global.thisIsCatuallyJest = Symobl.for('jest-yes-this-is-jest')` or something). I saw the earlier talk about checking for `global.it` - at some point Jest will want to provide a global-less API so it'd be great to not couple anything more to it (although not in the immediate future)

threepointone: Changed it so that it only detects jest now.

SimenB: that might also go away at some point, but not in the foreseeable future. Thanks!

packages/react-test-renderer/src/ReactTestRenderer.js
```diff
@@ -557,6 +557,10 @@ const ReactTestRendererFiber = {
   /* eslint-enable camelcase */
 
   unstable_setNowImplementation: setNowImplementation,
+
+  act<X>(callback: () => X): X {
+    return batchedUpdates(callback);
```
sebmarkbage: Needs some way to flush passive effects.

keba06:
I support this fb page completely, this page shares interesting and fun videos, good for both men and women ... https://www.facebook.com/Keba06

packages/react-dom/src/__tests__/ReactTestUtils-test.js
```diff
@@ -515,4 +515,58 @@ describe('ReactTestUtils', () => {
...
+      let called = false;
+      ReactTestUtils.act(() => {
+        ReactDOM.render(
+          <App
+            callback={() => {
+              called = true;
+            }}
+          />,
+          container,
+        );
```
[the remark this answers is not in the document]
keba06: https://www.facebook.com/Keba06

packages/react-test-renderer/src/ReactTestRenderer.js
Jessidhia: Whe I wrote my own hacks tests I cheated by just doing two interactiveUpdates batches :innocent:

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1790,6 +1790,20 @@ function scheduleWorkToRoot(fiber: Fiber, expirationTime): FiberRoot | null {
   return root;
 }
 
+export function warnIfNotCurrentlyBatchingInDev(): void {
+  if (__DEV__) {
+    if (isBatchingUpdates === false) {
+      warningWithoutStack(
+        false,
+        'It looks like you are in a test environment, trying to ' +
```
gaearon: We'll need to remember to update the wording, e.g. "UI" not "ui", and also it's not clear where to import "act" from. It's also not clear what "unexpected UI in testing" means.

packages/react-test-renderer/src/ReactTestRenderer.js
```diff
@@ -557,8 +563,60 @@ const ReactTestRendererFiber = {
...
+    }
+    flushPassiveEffects();
+    // we want the user to not expect a return,
+    // but we want to warn if they use it like they can await on it.
+    return {
+      then() {
+        if (__DEV__) {
+          warningWithoutStack(
+            false,
+            'Do not await an act(...) call, it is not a promise',
```
gaearon: Nit: Capital Promise. Also probably you mean "it does not return a Promise".

packages/react-test-renderer/src/ReactTestRenderer.js
```diff
@@ -557,8 +563,60 @@ const ReactTestRendererFiber = {
...
+  act(callback: () => void): Thenable {
+    let result = batchedUpdates(callback);
+    // warnings copied from ReactTestUtils
+    if (__DEV__) {
+      if (result !== undefined) {
+        let addendum;
+        if (typeof result.then === 'function') {
+          addendum =
+            '\n\nIt looks like you wrote act(async () => ...) or returned a Promise. ' +
+            'Do not write async logic inside act(...)\n';
```
gaearon: Missing period at the end of the sentence.

Also maybe "Putting asynchronous logic inside act() is not yet supported".

gaearon: I think it's the other way around. If we *don't* make it return a thenable now, people can write `await act()` without realizing they're awaiting undefined. *Then* making it actually return a Promise *would* be a breaking change.

By warning on `await` now we're effectively enforcing that you don't depend on what it returns. Therefore we can later make it actually return a Promise in some cases.

packages/react-test-renderer/src/ReactTestRenderer.js
```diff
@@ -557,8 +563,60 @@ const ReactTestRendererFiber = {
   /* eslint-enable camelcase */
 
   unstable_setNowImplementation: setNowImplementation,
+
+  act(callback: () => void): Thenable {
+    // note: keep these warning messages in sync with
```
Jessidhia: in sync with...?

packages/react-test-renderer/src/ReactTestRenderer.js
```diff
@@ -557,8 +563,60 @@ const ReactTestRendererFiber = {
...
+    createNodeMock: defaultTestOptions.createNodeMock,
+    tag: 'CONTAINER',
+  },
+  true,
+  false,
+);
+
+function flushPassiveEffects() {
+  // Trick to flush passive effects without exposing an internal API:
+  // Create a throwaway root and schedule a dummy update on it.
```
Jessidhia: 🤯

This will work for passive effects but we'll probably need something else for ConcurrentMode.

I guess that's where the potential Promise-based API change would come in.

packages/react-test-renderer/src/ReactTestRenderer.js
threepointone: 🙃good catch, thanks

kentcdodds:
👏👏👏 great job Sunil!

In the future, I'm going to subscribe to all your PRs. Watching the commit messages come in is great fun 😁

ferdaber:
I'm super excited for this. I can't count how many times I've had to explain to my teammates to await a set timeout when testing effects, this will make it so much less work!

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -647,15 +660,20 @@ describe('ReactHooksWithNoopRenderer', () => {
...
-        'Passive',
-        'Layout',
-        'Layout effect 0',
-        'Passive effect',
-        'Layout',
-        'Layout effect 1',
-      ]);
+
+      act(() => {
+        expect(ReactNoop.flush()).toEqual([
```
threepointone: This is a bit confusing, will fix

willdurand:
A side-effect of this patch is that it broke one of our test cases that has the `@jest-environment node` directive. We setup Enzyme globally (in a `setupFilesAfterEnv` file) because all but one file use a browser env.

Because the stub element is initialized globally using `document` in this patch, the whole test file fails:

```
 FAIL  tests/unit/core/server/test_sriDataPlugin.js
  ● Test suite failed to run

    ReferenceError: document is not defined

      at node_modules/react-dom/cjs/react-dom-test-utils.development.js:944:27
      at Object.<anonymous> (node_modules/react-dom/cjs/react-dom-test-utils.development.js:1283:5)
      at Object.<anonymous> (node_modules/react-dom/test-utils.js:6:20)
      at Object.<anonymous> (node_modules/enzyme-adapter-react-16/build/ReactSixteenAdapter.js:31:18)
```

sakshigupta:
> A side-effect of this patch is that it broke one of our test cases that has the `@jest-environment node` directive. We setup Enzyme globally (in a `setupFilesAfterEnv` file) because all but one file use a browser env.
> 
> Because the stub element is initialized globally using `document` in this patch, the whole test file fails:
> 
> ```
>  FAIL  tests/unit/core/server/test_sriDataPlugin.js
>   ● Test suite failed to run
> 
>     ReferenceError: document is not defined
> 
>       at node_modules/react-dom/cjs/react-dom-test-utils.development.js:944:27
>       at Object.<anonymous> (node_modules/react-dom/cjs/react-dom-test-utils.development.js:1283:5)
>       at Object.<anonymous> (node_modules/react-dom/test-utils.js:6:20)
>       at Object.<anonymous> (node_modules/enzyme-adapter-react-16/build/ReactSixteenAdapter.js:31:18)
> ```

I am also facing the same issue, any fix for this?

Jessidhia:
Very dirty hack around it to get the `require` to work: `if (typeof document === 'undefined') { global.document = { createElement() { return null } } }`

But this may possibly crash elsewhere.

bvaughn:
Thanks for pointing this out, @sakshigupta. Sunil is going to look into it shortly.

NE-SmallTown:
> React only guarantees that it'll be run before the browser has reflected changes made to the dom to the user (ie - before the browser has 'painted' the screen)

I'm confused about this, especially the word **'before'**, for example:

```
function App(props) {
  React.useEffect(() => {
    props.callback();
  });
  return 'hi';
}

let called = false;
ReactDOM.render(
  <App
    callback={() => {
    	console.log('callback be called');
      called = true;
    }}
  />,
);
console.log(document.body.innerHTML);
```

When execute `console.log(document.body.innerHTML)`, we will see that the body has text 'hi' in the output string, it means this content has been painted to the screen at this moment, but the 'callback be called' will be called **after** that console.log instead 'before'

Jessidhia:
I think that was kind of a mixup.

The current guarantees are, AFAIK:

- `useEffect` will never block React modifying the DOM ("making a commit").
- each given `useEffect` will only be invoked once for each commit.
- any queued `useEffect` will all be forced to execute before any code re-enters React (event handlers, triggering `setState`, etc).

`useEffect` right now runs in its own microtask tick that is scheduled using `requestAnimationFrame`, and can be delayed for up to 5s before it is called, but that's an implementation detail.

PR #14746, opened 2019-02-01
[scheduler] Add method for forcing a lower framerate

n8schloss:
On very low end devices we may want to force scheduler to run at a lower framerate than 30fps even if the browser/device supports something faster. Lower spec devices may not be able to run enough tasks within 33ms to make meaningful progress, so this adds a function that lets developers force something larger.


packages/scheduler/src/Scheduler.js
```diff
@@ -634,10 +647,10 @@ if (globalValue && globalValue._schedMock) {
       nextFrameTime < activeFrameTime &&
       previousFrameTime < activeFrameTime
     ) {
-      if (nextFrameTime < 8) {
+      if (nextFrameTime < minFrameTime) {
```
matthargett: should we warn when in dev mode about the value being less than 8?

matthargett:
We were just talking about needing something like this internally on a CPU/memory bandwidth constrained React Native platform. My only nit is that a dev mode warning/safeguard should be put in if refresh rates >120hz is set  (and fallback to default if it is) to match the comments.

packages/scheduler/src/Scheduler.js
```diff
@@ -562,11 +565,21 @@ if (globalValue && globalValue._schedMock) {
...
 
   shouldYieldToHost = function() {
     return frameDeadline <= getCurrentTime();
   };
 
+  forceFrameTimes = function(options) {
+    if (options.minFrameTime) {
+      minFrameTime = options.minFrameTime;
+    }
+    if (options.startingFrameTime) {
```
acdlite: Can you explain why this part is necessary? I don't think I get it

n8schloss:
Alright, I totally failed at rebasing this correctly. This is like my second time using github for PR, I'll just make a new PR with my changes.

PR #14747, opened 2019-02-01
Tweak invalid Hook warning and error

gaearon:
- Adds a link to https://reactjs.org/warnings/invalid-hook-call-warning.html for the invariant. (I guess it's not a warning.. but I also don't want to introduce a new URL naming scheme. We don't have separate pages for errors right now. We can move it later if it causes confusion but I'm not sure it would.)

- Rewords the warning when we know more exactly what happened.

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRendererHooks.js
+++ b/packages/react-dom/src/server/ReactPartialRendererHooks.js
@@ -57,13 +57,15 @@ let currentHookNameInDev: ?string;
 function resolveCurrentlyRenderingComponent(): Object {
   invariant(
     currentlyRenderingComponent !== null,
-    'Hooks can only be called inside the body of a function component.',
+    'Hooks can only be called inside the body of a function component. ' +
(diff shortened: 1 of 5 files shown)


packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -57,13 +57,15 @@ let currentHookNameInDev: ?string;
...
-    'Hooks can only be called inside the body of a function component.',
+    'Hooks can only be called inside the body of a function component. ' +
+      '(https://fb.me/react-invalid-hook-call)',
   );
   if (__DEV__) {
     warning(
       !isInHookUserCodeInDev,
-      'Hooks can only be called inside the body of a function component. ' +
-        'Do not call Hooks inside other Hooks. For more information, see ' +
+      'Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks. ' +
```
threepointone: nice nice nice

PR #14748, opened 2019-02-02
fix typo

ryota-murakami:
that's it.


bvaughn:
I believe the semicolon is actually correct the way it was before 😅 Thanks though!

PR #14752, opened 2019-02-04
Use SameValue instead of === to check for dispatchAction equivalence

Jessidhia:
This allows a Reducer/State that happens to hold a `-0` or `NaN` to not bail out of the bail outs / incorrectly bail out. It's the same algorithm that is used to check the `deps` array in Effect/Memo/etc.

I'm not sure if this is sufficient test coverage, but it does what I expected (and, more importantly, fails if I undo the changes).

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -612,7 +612,7 @@ function updateReducer<S, I, A>(
 
         // Mark that the fiber performed work, but only if the new state is
         // different from the current state.
-        if (newState !== hook.memoizedState) {
+        if (!is(newState, hook.memoizedState)) {
           markWorkInProgressReceivedUpdate();
         }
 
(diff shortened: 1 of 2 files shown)


acdlite:
This makes sense to me, since we also use `Object.is` semantics for PureComponent, memo, context providers, deps, and so on. Thanks for the PR!

thomschke:
Immutable data structure libraries have their own equality check (like [immutable-js](http://facebook.github.io/immutable-js/docs/#/is)).

see -> https://github.com/facebook/react/pull/14569#issuecomment-458864678

sudhirj:
This really needs a giant warning saying "DO NOT USE ARRAYS AS STATE!!!". This completely breaks them:
```
const [list, setList] = useState([])
setList((l) => {
  l.push(1)
  return l
}
``` 
is a complete fail, because `Object.is` sees the same array returned, absolutely nothing happens.

PR #14756, opened 2019-02-04
[Wait for Hooks before merging] Scheduler.unstable_next

acdlite:
When a user initiates an action, they expect immediate feedback. They don't necessarily expect the action to finish immediately, but they do expect some sort of response from the UI within a timeframe that feels immediate, so they know their request was received.

In React, the recommended way to achieve this is to split expensive updates into two separate tasks. The first task contains work that needs to finish immediately. We refer to this as a "user blocking" update. An example is, when switching to a new tab in a tabbed interface, the new tab should change color immediately after it's clicked.

Everything that doesn't need to happen immediately (i.e. doesn't block the user) should happen in a second task. This task typically takes longer to finish, either because it's CPU intensive or because it relies on IO-bound resources. In the tab switching example, this includes loading and rendering the content that corresponds to the newly selected tab. We refer to this as a "normal" update, because this category tends to encompass most of the work in a React application.

To implement this pattern in React, you need a separate `setState` call for each of the two tasks. The first `setState` should happen directly inside the event handler. React will automatically infer that this update is user-blocking and schedule it at the corresponding priority.

The second update — the more expensive one that does the bulk of the work — should be wrapped in a call to `Scheduler.next`:

```js
const onClick = () => {
  // Local update that renders immediately
  setHighlightedTab(tab);
  Scheduler.next(() => {
    // Navigation update that might take a bit longer to finish
    dispatch({type: 'switch_tab', tab});
  });
};
```

Updates initiated inside the function passed to `Scheduler.next` are scheduled with "normal" priority, instead of user-blocking priority. (The function itself is immediately invoked, a la `batchedUpdates` and `flushSync`.)

The name "next" is intentionally vague about its exact semantics. It's meant to imply that the task scheduled inside passed function happen after the task scheduled directly inside the event handler.


packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1579,52 +1586,58 @@ function computeUniqueAsyncExpiration(): ExpirationTime {
...
-    if (
-      lowestPriorityPendingInteractiveExpirationTime === NoWork ||
-      expirationTime < lowestPriorityPendingInteractiveExpirationTime
-    ) {
-      lowestPriorityPendingInteractiveExpirationTime = expirationTime;
+
+    // If we're in the middle of rendering a tree, do not update at the same
+    // expiration time that is already rendering.
+    if (nextRoot !== null && expirationTime === nextRenderExpirationTime) {
+      expirationTime -= 1;
```
Jessidhia: One thing that confused me when I was reading this: is this before or after the `nextRenderExpirationTime`?

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1579,52 +1586,58 @@ function computeUniqueAsyncExpiration(): ExpirationTime {
...
-    } else {
-      // Updates during the render phase should expire at the same time as
-      // the work that is being rendered.
-      expirationTime = nextRenderExpirationTime;
-    }
+  if ((fiber.mode & ConcurrentMode) === NoContext) {
+    // Outside of concurrent mode, updates are always synchronous.
+    expirationTime = Sync;
+  } else if (isWorking && !isCommitting) {
+    // During render phase, updates expire during as the current render.
```
Jessidhia: "during the current render"? "as the current render"? "with the current render"?

packages/scheduler/src/Scheduler.js
```diff
@@ -264,6 +264,37 @@ function unstable_runWithPriority(priorityLevel, eventHandler) {
   }
 }
 
+function unstable_next(eventHandler) {
```
Jessidhia: This is IIUC an inlined version of:

```ts
return runWithPriority(
  Math.min(NormalPriority, getCurrentPriorityLevel()),
  eventHandler
)
```

?

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -793,9 +798,11 @@ function commitRoot(root: FiberRoot, finishedWork: Fiber): void {
       // TODO: Avoid this extra callback by mutating the tracing ref directly,
       // like we do at the beginning of commitRoot. I've opted not to do that
       // here because that code is still in flux.
-      callback = Schedule_tracing_wrap(callback);
+      callback = Scheduler_tracing_wrap(callback);
     }
-    passiveEffectCallbackHandle = schedulePassiveEffects(callback);
+    passiveEffectCallbackHandle = runWithPriority(NormalPriority, () => {
+      return schedulePassiveEffects(callback);
+    });
```
Jessidhia: When I was reading this code last week I was wondering how you'd be able to mess with the `schedulePassiveEffects` priority. Doesn't seem normally possible... unless you cause a `Sync` update I guess...

PR #14757, opened 2019-02-05
Avoid importing Scheduler directly

acdlite:
The reconciler should not depend directly on Scheduler. This adds it to the host config for the renderer instead.

(Except for `scheduler/tracing` imports, which are used only by the profiling build. I've left those imports as-is, though I'm open to directing those through the host config, too.)


packages/react-test-renderer/src/ReactTestHostConfig.js
```diff
@@ -210,6 +210,8 @@ export const shouldYield = TestRendererSchedulingShouldYield;
 export const scheduleTimeout = setTimeout;
 export const cancelTimeout = clearTimeout;
 export const noTimeout = -1;
+export const schedulePassiveEffects = scheduleDeferredCallback;
+export const cancelPassiveEffects = cancelDeferredCallback;
```
bvaughn: Seems a little weird now that we're exporting these multiple times with the same name ¯\_(ツ)_/¯ Oh well though.

packages/react-noop-renderer/src/createReactNoop.js
```diff
@@ -854,6 +870,16 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
       return yieldedValues;
     },
 
+    flushPassiveEffects() {
+      // Trick to flush passive effects without exposing an internal API:
+      // Create a throwaway root and schedule a dummy update on it.
+      const rootID = 'bloop';
```
bvaughn: tiny nit: This is unlikely to conflict, but should we maybe pick something a little longer and more unique?

acdlite: I gotcha

packages/react-test-renderer/src/ReactTestHostConfig.js
acdlite: Right this change doesn't have much immediate impact, but gives us flexibility to customize the implementation in the future. For example, noop renderer already has a forked implementation.

PR #14758, opened 2019-02-05
[TestUtils.act] fix return result checking 

threepointone:
fixes a bug from #14744 where returning null from `.act` would throw when null. Sorry! Added a test to be sure.


threepointone:
doing a merge in with master, gimme a minute

PR #14768, opened 2019-02-06
[TestUtils.act] warn when using TestUtils.act in node

### LINKED ISSUE #14764: 16.8 regression: react-dom/test-utils no longer require()-able in pure node
Bug. Behavior regression.

Package throws when being required in Node.

```
avi@localhost  ~/projects/r168  node
> require('react-dom/test-utils')
ReferenceError: document is not defined
    at /home/avi/projects/r168/node_modules/react-dom/cjs/react-dom-test-utils.development.js:944:27
    at Object.<anonymous> (/home/avi/projects/r168/node_modules/react-dom/cjs/react-dom-test-utils.development.js:1283:5)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
```

`react-dom/test-utils` should not throw in "pure" node (without `jsdom` providing fake document on global).

happens on 16.8.0
previous versions work.

threepointone:
- lazily initializes the stub element we use when flushing effects 
- adds a warning for when you try to run in in node (rather, non-document envs)
- fixes #14764

### DIFF
--- a/packages/react-dom/src/test-utils/ReactTestUtils.js
+++ b/packages/react-dom/src/test-utils/ReactTestUtils.js
@@ -392,16 +392,17 @@ const ReactTestUtils = {
 
   act(callback: () => void): Thenable {
     if (actContainerElement === null) {
-      // first, test whether we can actually create the stub element
-      invariant(
-        typeof document !== 'undefined' &&
-          document !== null &&
-          typeof document.createElement === 'function',
-        'It looks like you called TestUtils.act() in a non-browser-like environment' +
-          "(like node.js). This is not supported. If you're using" +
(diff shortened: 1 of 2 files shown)


joepuzzo:
Out of curiosity what is the release cycle for patch releases?

Issue #14763, opened 2019-02-06
react-test-renderer/shallow render throws TypeError: Cannot read property 'current' of undefined


vjrasane:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

```
TypeError: Cannot read property 'current' of undefined

       6 |   it('renders correctly', () => {
       7 |     const renderer = new ShallowRenderer()
    >  8 |     const render = renderer.render(
         |                             ^
       9 |       <StatusChip
      10 |         status={'Acknowledged'}
      11 |         text={'IN PROGRESS'}

      at ReactShallowRenderer.render (node_modules/react-test-renderer/cjs/react-test-renderer-shallow.development.js:703:53)
      at Object.render (src/components/StatusChip/index.snapshot.test.jsx:8:29)
```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

Run ShallowRenderers render method

**What is the expected behavior?**

render should return shallow render of the given component

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

"react": "16.6.3",
"react-test-renderer": "16.8.0",
 "enzyme": "3.7.0",
 "enzyme-adapter-react-16": "1.7.0",
 "jest": "24.0.0"

Workaround is to add:

  "resolutions": {
    "react-test-renderer": "16.7.0"
  }

(for yarn)

alyaptsev:
The same problem
https://github.com/facebook/react/blob/master/packages/react-test-renderer/src/ReactShallowRenderer.js#L44 `ReactCurrentDispatcher` is `undefined` 😨

DonikaV:
for npm you can just install **react-test-renderer** in your package.json with specific version. We just fixed our problems with this solution.

vjrasane:
seems like it works with the latest react version 16.8.0. My previously used version 16.6.3 was failing, probably because the transitive react-test-renderer version was newer and it required something that was not present in the old react version.

bvaughn:
I think this is due to us splitting the current owner/dispatcher refs (#14548)

gaearon:
[...]

PR #14770, opened 2019-02-06
Backwards compat fix for ReactCurrentDispatcher on older react versions

bvaughn:
Potential fix for #14763.

Current owner and dispatcher used to share the same ref, but PR #14548 split them out to better support the `react-debug-tools` package. This had the unintended side effect of newer renderers incompatible with versions of the `react` package that predate this change.

This PR adds a fallback property.

### DIFF
--- a/packages/shared/ReactSharedInternals.js
+++ b/packages/shared/ReactSharedInternals.js
@@ -10,4 +10,19 @@ import React from 'react';
 const ReactSharedInternals =
   React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;
 
+// Add fallback for newer renderers running with older react package versions.
+// Current owner and dispatcher used to share the same ref,
+// but PR #14548 split them out to better support the react-debug-tools package.
+if (!ReactSharedInternals.hasOwnProperty('ReactCurrentDispatcher')) {
+  const {ReactCurrentOwner} = ReactSharedInternals;
+  ReactSharedInternals.ReactCurrentDispatcher = {
+    get current() {
+      return ReactCurrentOwner.currentDispatcher;
+    },
+    set current(value) {
+      ReactCurrentOwner.currentDispatcher = value;
+    },
+  };
+}
+
 export default ReactSharedInternals;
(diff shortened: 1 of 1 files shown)


packages/shared/ReactSharedInternals.js
```diff
@@ -10,4 +10,19 @@ import React from 'react';
 const ReactSharedInternals =
   React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;
 
+// Add fallback for newer renderers running with older react package versions.
+// Current owner and dispatcher used to share the same ref,
+// but PR #14548 split them out to better support the react-debug-tools package.
+if (!ReactSharedInternals.hasOwnProperty('ReactCurrentDispatcher')) {
+  const {ReactCurrentOwner} = ReactSharedInternals;
+  ReactSharedInternals.ReactCurrentDispatcher = {
```
sebmarkbage: Can't this just be `{current: null}`? Why does it need to forward to the owner object?

bvaughn: Because I wasn't convinced that no 16.8+ renderer would ever have a valid use case for accessing current owner?

bvaughn: e.g. react-cache uses it

bvaughn:
Okay, I've replaced it with `current: null`

bvaughn:
We'll publish a bugfix release with this fix shortly

Vlad-Shevliakov:
Goddamn!

dzienisz:
@bvaughn do you think that this fix wouldn't happen if we could just test software a little bit more?

gaearon:
I’m not sure if you’re genuinely asking or implying we didn’t test React. We spent weeks ironing out remaining issues in this release and at some point you just gotta publish. (You were welcome to try alphas which have been available for months, and the latest one had the same issue.)

We’ll keep this particular situation in mind for the future (mismatching versions of React and ReactDOM). But we’re also humans and can make mistakes. If you don’t want our mistakes to break your code then you should be using lockfiles in a production environment.

PR #14800, opened 2019-02-08
Split ReactHooks-test into public and internal

gaearon:
Most of these tests can run against public bundles.

(In fact all of them could but that would make assertions a bit annoying in DEV.)

I split out the tests that needed no changes. The total number is the same (32).


gaearon:
Hmm seems like bundle tests don't quite pass. I'll check next week.

sophiebits:
I think your test failures are because ReactFeatureFlags.test-renderer (unlike dom) doesn't enable debugRenderPhaseSideEffectsForStrictMode.

gaearon:
meh, this got outdated and I don't care strongly enough about this

Issue #14782, opened 2019-02-07
useImperativeHandle behaves as if inputs are [] by default


AhmedElywa:
useImperativeHandle have bug in 16.8.1 if you want to access any state will get the initial state not current state but in 16.8.0-alpha.1 it work good 

You can see here https://codesandbox.io/s/xjl8znwpz open console when you change input value will get the currently value in console but if you clicked button get value from ref will get initial value (e)

If you try same here https://codesandbox.io/s/qxkll6po0j with 16.8.0-alpha.1 when clicked button get currently value

FredyC:
That's a pretty common "bug" caused by Javascript closures. React apparently is not recreating handle on every render so it will always see only the value from the first render.

You can work around that by using ref: https://codesandbox.io/s/8nr4x36zy9

FredyC:
I think this kind of problems should get its own section in the docs. Many people will get burned by this (me included) because it's just not that apparent without knowing this advanced JS behavior.

AhmedElywa:
But why work in alpha version?

AhmedElywa:
@FredyC I’m not use it to transfer value this just example 
I use here https://github.com/oahtech/rnsc/blob/master/src/components/Accordion/index.js

To control the state of expanded or collapsed

FredyC:
I think this is a mindset issue. You should be controlling components by calling methods on them. Instead, you should pass props that will change that state. I don't have a capacity to update that example of yours right now. Just try to fiddle with it and I am sure you will figure a way without imperative access.

MicahChalmer:
There's a third parameter to `useImperativeHandle` called `inputs` which is intended to be used in the same way as similar parameters to `useEffect` and `useMemo`--if you pass an array to it, it will re-create the handle whenever any of the inputs have changed since the last call.  I find that if I pass in what the closure I'm passing depends on, it works as expected.

The bug here (and I do think it's a bug) is this: unlike the other handles that have an `inputs` parameter (and unlike the previous behavior from the alphas, which I still think is the correct behavior,) if you don't pass `inputs` it defaults to ignoring the new closure you passed in.  It defaults to behaving as if you had passed `[]` to inputs.  The others default to _always_ calling it, which I believe is the right thing to do, for precisely the reason that it avoids this gotcha around values saved from previous renders.  The current behavior is "memoized by default" and that's wrong for the same reasons explained in other contexts.

[...]

PR #14801, opened 2019-02-08
Fix useImperativeHandle to have no deps by default

gaearon:
Fixes https://github.com/facebook/react/issues/14782. If nothing is provided, there should be _no_ dependencies — not `[ref]` alone.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -905,7 +905,7 @@ function mountImperativeHandle<T>(
 
   // TODO: If deps are provided, should we skip comparing the ref itself?
   const effectDeps =
-    deps !== null && deps !== undefined ? deps.concat([ref]) : undefined;
+    deps !== null && deps !== undefined ? deps.concat([ref]) : deps;
 
   return mountEffectImpl(
     UpdateEffect,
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -905,7 +905,7 @@ function mountImperativeHandle<T>(
 
   // TODO: If deps are provided, should we skip comparing the ref itself?
   const effectDeps =
-    deps !== null && deps !== undefined ? deps.concat([ref]) : [ref];
+    deps !== null && deps !== undefined ? deps.concat([ref]) : deps;
```
acdlite:
```suggestion
    deps !== null && deps !== undefined ? deps.concat([ref]) : null;
```

Nit: Should only be null, not undefined. I wondered why this wasn't a type error but it's because `mountEffectImpl` and `updateEffectImpl` contain their own type check; we might inline those later.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -931,7 +931,7 @@ function updateImperativeHandle<T>(
...
```
acdlite: Same as above

gaearon:
>You checked the shallow renderer and server renderer too?

Just checked — both are noops.

PR #14802, opened 2019-02-08
[ShallowRenderer] Queue/rerender on dispatched action after render component with hooks

rodrigopr:
This should help make shallow render a viable option for testing components that use hooks.

### DIFF
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -414,8 +414,7 @@ class ReactShallowRenderer {
       }
     } else {
       // This means an update has happened after the function component has
-      // returned. On the server this is a no-op. In React Fiber, the update
-      // would be scheduled for a future render.
+      // returned from a different component.
     }
   }
 
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -384,7 +384,7 @@ class ReactShallowRenderer {
         'an infinite loop.',
     );
 
-    if (componentIdentity === this._currentlyRenderingComponent) {
+    if (componentIdentity === this._previousComponentIdentity) {
       // This is a render phase update. Stash it in a lazily-created map of
       // queue -> linked list of updates. After this render pass, we'll restart
       // and apply the stashed updates on top of the work-in-progress hook.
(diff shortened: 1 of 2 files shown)


rodrigopr:
@gaearon, @trueadm, @acdlite, could you review this? (sorry to ping like this)
Being able to trigger hooks state changes (useState, useReducer) after render will allow us to keep most of our shallow tests intact if we migrate components to use hooks.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -384,7 +384,7 @@ class ReactShallowRenderer {
         'an infinite loop.',
     );
 
-    if (componentIdentity === this._currentlyRenderingComponent) {
+    if (componentIdentity === this._previousComponentIdentity) {
```
gaearon: Why is this change correct? It disagrees with what the comment inside says ("This is a render phase update"). What was it supposed to mean? Why was it changed?

gaearon:
I reverted in https://github.com/facebook/react/pull/14839 — I appreciate the PR but I wanted to cut a patch soon and it's not obvious to me that this change is correct. (Although you're right the existing code is not either.)

This code was originally written with no outside state updates in mind. It was written as 1:1 to the server renderer. It might need a larger refactoring to work with updates at arbitrary points in time.

At the least, we should rename the variables and rewrite the comments that indicate these are only render-phase updates. With this change the logic doesn't match the comments and variables, and I'm worried this will make the code so confusing that nobody will be able to fix further problems in it.

gaearon:
(I'd be happy if you could take this work. Let's just ensure the code matches the intention, and that we examine it all instead of minimal tweaks to "make it work".)

gaearon:
I filed https://github.com/facebook/react/issues/14840 to track this.

packages/react-test-renderer/src/ReactShallowRenderer.js
rodrigopr: Yeah, I missed updating the comment =/
Before this change `_dispatchAction` ignored calls after the render returned, this checks if its the same component and queue the action.

rodrigopr:
No problem @gaearon, I'll move the discussion to the issue.

gaearon:
Thanks!

trueadm:
Thanks for digging into this deeper. I have a larger fix ready that should handle this case properly.

PR #14804, opened 2019-02-09
Only call destroy when destroy is a function

bennyhobart:
Currently experiencing an issue when returning null from useEffect, when a component unmounts react throws an error.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -321,7 +321,7 @@ function commitHookEffectList(
         // Unmount
         const destroy = effect.destroy;
         effect.destroy = undefined;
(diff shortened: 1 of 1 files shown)


gaearon:
Returning `null` from `useEffect` is not supported and should have been forbidden by Flow types. Let's follow up internally. Thanks for PR!

steveluscher:
Should we maybe make this `__DEV__` check apply to the unmount cycle too to give people who don't use Flow can clean up their `useEffect` hooks?

https://github.com/bennyhobart/react/blob/34cef5d345b163f9f264aa5fb0ae87485199f0f1/packages/react-reconciler/src/ReactFiberCommitWork.js#L333-L364

gaearon:
Meaning, adding an extra warning before the crash? That sounds like a good idea.

PR #14808, opened 2019-02-09
Fix hydration with createRoot warning

sebmarkbage:
It's suggesting an API that doesn't exist.

`{hydrate: true}` is passed to `createRoot`, not `render`.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMRoot-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMRoot-test.js
@@ -410,7 +410,7 @@ describe('ReactDOMRoot', () => {
         // We care about this warning:
         'You are calling ReactDOM.hydrate() on a container that was previously ' +
           'passed to ReactDOM.unstable_createRoot(). This is not supported. ' +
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/__tests__/ReactDOMRoot-test.js
```diff
@@ -410,7 +410,7 @@ describe('ReactDOMRoot', () => {
         // We care about this warning:
         'You are calling ReactDOM.hydrate() on a container that was previously ' +
           'passed to ReactDOM.unstable_createRoot(). This is not supported. ' +
-          'Did you mean to call root.render(element, {hydrate: true})?',
+          'Did you mean to call createRoot(container, {hydrate: true}).render(element)?',
```
threepointone:
```suggestion
          'Did you mean to call unstable_createRoot(container, {hydrate: true}).render(element)?',
```

threepointone: or even `ReactDOM.unstable_createRoot(`?

Issue #14285, opened 2018-11-20
action does not dispatched(useReducer) when It is dispatched in useEffect's clean up callback


merongmerongmerong:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

action dispatched in return callback of `useEffect` seem to not work

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

https://codesandbox.io/s/5yqmo128v4

only foo -> baz is logged

```javascript
import React, { useState, useEffect, useReducer } from "react";
import ReactDOM from "react-dom";

import "./styles.css";

function reducer(state, action) {
  console.log("bar", action); // not logged
  // debugger
  return state;
}

function Foo({ value }) {
  const [state, dispatch] = useReducer(reducer, {});

  useEffect(
    () => {
      return () => {
        console.log("foo");
        // debugger
        dispatch({ type: "foo" });
        // debugger
        console.log("baz");
      };
    },
    [state, value]
  );

  return <p>{value}</p>;
}

function App() {
  const [value, set] = useState(0);

  return (
    <div className="App">
      <button onClick={() => set(value + 1)}>INC</button>
      {value % 2 ? <Foo value={value} /> : null}
    </div>
  );
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```

**What is the expected behavior?**
bar is logged in console
(foo -> baz -> bar`action`)

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

react: "16.7.0-alpha.2",
react-dom: "16.7.0-alpha.2"

aweary:
@merongmerongmerong what exactly would you expect to happen here? In this example the `Foo` component being unmounted when the effect's cleanup funciton is called. You can't update state for an unmounted component, so this behavior seems expected.

If you have an effect that gets cleaned up *in a component that isn't being unmounted* it calls the reducer as expected: https://codesandbox.io/s/vqr9olv4o7

In this case the behavior seems correct, but maybe it would be good to warn here.

merongmerongmerong:
@aweary 
[...]

PR #14809, opened 2019-02-09
Add test for calling dispatch from useEffect cleanup

tijwelch:
Adds a test regarding #14285 . When dispatch is called from useEffect clean-up function, expect that the reducer is called. This was not the case in the past and was the source of #14285

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
@@ -669,6 +669,37 @@ describe('ReactHooks', () => {
     }).toThrow('is not a function');
   });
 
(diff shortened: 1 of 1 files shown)


acdlite:
This is only observable if your reducer contains side effects. And if your reducer has side effects, you're going to have all sorts of problems, especially in concurrent mode — we double invoke reducers in dev to flush out these types of bugs.

Your test happens to pass because we have an optimization that will eagerly call a reducer in certain, limited scenarios. But it's not even the ideal behavior here, because there's no point calling a reducer for a component that is being unmounted. So you definitely shouldn't rely on this behavior.

tijwelch:
Thanks for the clarification.

Issue #14807, opened 2019-02-09
Shallow renderer does not support React.memo


ljharb:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
In normal rendering, you can memoize a class-based component: https://jsfiddle.net/586ea3cx/

With the shallow renderer, it seems like you can't: https://jsfiddle.net/odj217Lv/1/

This might be blocking https://github.com/airbnb/enzyme/pull/1914, in which tests are failing with `Cannot call a class as a function`. (it's tough to repro stuff with the shallow renderer)

Yurickh:
I would like to point out that React.memo being used with classes is not documented behaviour and _might not be_ part of React's public facing API.

You can check that all references to React.memo in the docs say it is meant to be used with functional components [here](https://reactjs.org/blog/2018/10/23/react-v-16-6.html#reactmemo) and [here](https://reactjs.org/docs/react-api.html#reactmemo).

Yurickh:
I would like to point out that React.memo being used with classes is not documented behaviour and _might not be_ part of React's public facing API.

You can check that all references to React.memo in the docs say it is meant to be used with functional components [here](https://reactjs.org/blog/2018/10/23/react-v-16-6.html#reactmemo) and [here](https://reactjs.org/docs/react-api.html#reactmemo).

ljharb:
Indeed; but it doesn’t throw when you try, and it seems to work fine in actual React - I’d expect it to work, or not work, the same everywhere.

gaearon:
`memo()` can be used with any component type. Happy to take a PR that adds support for it!

aweary:
This doesn't seem to be isolated to class components, it fails with any component using `memo()`: https://jsfiddle.net/f0j1e5tw/6/

ljharb:
The relevant tests in the open enzyme PR only fail for class components, fwiw.

gaearon:
Fixed in 16.8.5.

PR #14816, opened 2019-02-11
Support React.memo in ReactShallowRenderer

aweary:
Fixes https://github.com/facebook/react/issues/14807

* Adds support for `React.memo` in `React.ShallowRenderer`. Instead of using `element.type` directly it now checks if the element is `REACT_MEMO_TYPE` and uses the nested type if it is.

* Validates the propTypes for the inner component

For testing I just coped `ReactShallowRenderer-test` and wrapped every component in `React.memo` to validate everything worked. I don't think we should take approach, but it made it easy to find issues 🤷‍♂️maybe we can have a single set of tests that run with and without memo?

### DIFF
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -8,7 +8,7 @@
  */
 
 import React from 'react';
-import {isForwardRef, isMemo} from 'react-is';
+import {isForwardRef, isMemo, ForwardRef} from 'react-is';
 import describeComponentFrame from 'shared/describeComponentFrame';
 import getComponentName from 'shared/getComponentName';
 import shallowEqual from 'shared/shallowEqual';
@@ -589,18 +589,21 @@ class ReactShallowRenderer {
           ReactCurrentDispatcher.current = this._dispatcher;
           this._prepareToUseHooks(elementType);
           try {
-            if (isForwardRef(element)) {
+            // elementType could still be a ForwardRef if it was
+            // nested inside Memo.
+            if (elementType.$$typeof === ForwardRef) {
+              invariant(
+                typeof elementType.render === 'function',
+                'forwardRef requires a render function but was given %s.',
+                typeof elementType.render,
+              );
               this._rendered = elementType.render.call(
                 undefined,
                 element.props,
-                element.ref
+                element.ref,
               );
             } else {
-              this._rendered = elementType.call(
-                undefined,
-                element.props,
-                this._context,
-              );
+              this._rendered = elementType(element.props, this._context);
             }
           } finally {
             ReactCurrentDispatcher.current = prevDispatcher;
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -516,8 +516,9 @@ class ReactShallowRenderer {
     }
 
     const elementType = isMemo(element.type) ? element.type.type : element.type;
-    this._rendering = true;
+    const previousElement = this._element;
 
+    this._rendering = true;
     this._element = element;
     this._context = getMaskedContext(elementType.contextTypes, context);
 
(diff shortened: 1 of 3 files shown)


gaearon:
If we support `memo()` and `forwardRef()`, their combination `memo(forwardRef())` should also work. Does it?

aweary:
@gaearon I've added support for `memo(forwardRef())` and also  made sure that `forwardRef(memo())` fails as expected. 

What do you think of the testing strategy? Should I remove `ReactShallowRendererMemo-test` and just rely on the `memo` tests added to `ReactShallowRenderer-test`? Do you think it's worth running every test with and without `memo`?

packages/react-test-renderer/src/__tests__/ReactShallowRenderer-test.js
```diff
@@ -1454,4 +1454,115 @@ describe('ReactShallowRenderer', () => {
...
+    const SomeComponent = React.forwardRef(SomeMemoComponent);
+    const shallowRenderer = createRenderer();
+    expect(() => {
+      expect(() =>
+        shallowRenderer.render(<SomeComponent ref={testRef} />),
+      ).toWarnDev(
+        'Warning: forwardRef requires a render function but received ' +
+          'a `memo` component. Instead of forwardRef(memo(...)), use ' +
+          'memo(forwardRef(...))',
+        {withoutStack: true},
```
aweary: This test is failing saying `Expected test not to call console.error()`, but I'm calling `toWarnDev` here so I don't think that's right?

packages/react-test-renderer/src/__tests__/ReactShallowRenderer-test.js
```diff
@@ -1454,4 +1454,115 @@ describe('ReactShallowRenderer', () => {
...
+
+  it('should not call the comparison function with React.memo on the initial render', () => {
+    const areEqual = jest.fn(() => false);
+    const SomeComponent = React.memo(({foo}) => {
+      return <div>{foo}</div>;
+    }, areEqual);
+    const shallowRenderer = createRenderer();
+    shallowRenderer.render(<SomeComponent foo={1} />);
+    expect(areEqual).not.toHaveBeenCalled();
+    expect(shallowRenderer.getRenderOutput()).toEqual(<div>1</div>);
```
aweary: This test is failing here because the rendered output doesn't match. I verified that `getRenderOutput` is returning `<div>1</div>`. I guess its failing because they're not the same React elements, but we're using this same pattern in this test suite so I don't know why its failing here.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -500,7 +500,8 @@ class ReactShallowRenderer {
       element.type,
     );
     invariant(
-      isForwardRef(element) || typeof element.type === 'function',
+      isForwardRef(element) ||
+        (typeof element.type === 'function' || isMemo(element.type)),
```
ljharb: any reason not to use isValidElementType from react-is here, and asserting the type isn’t string, rather than repeating some of the logic?

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -545,14 +565,13 @@ class ReactShallowRenderer {
           }
         }
 
-        if (element.type.hasOwnProperty('contextTypes')) {
+        if (elementType.hasOwnProperty('contextTypes')) {
```
ljharb: you may want to use Object.prototype.hasOwnProperty.call here, since it’d otherwise work fine to pass a function component with a null prototype.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -561,19 +580,34 @@ class ReactShallowRenderer {
...
+        }
+        if (shouldRender) {
+          const prevDispatcher = ReactCurrentDispatcher.current;
+          ReactCurrentDispatcher.current = this._dispatcher;
+          this._prepareToUseHooks(elementType);
+          try {
+            // elementType could still be a ForwardRef if it was
+            // nested inside Memo.
+            this._rendered =
+              elementType.$$typeof === ForwardRef
```
ljharb: should this use `isForwardRef(element)` (and cache the isForwardRef status in a var, since it seems to be checked multiple times)?

aweary: It's a little confusing, but that doesn't work because `isForwardRef` expects a React Element for the `forwardRef`, not the actual object returned from `forwardRef` itself.

```js
const forwardRef = React.forwardRef(...);
isForwardRef(forwardRef) // false
isForwardRef(React.createElement(forwardRef)) // true
```

packages/react-test-renderer/src/ReactShallowRenderer.js
aweary: Probably because `isValidElementType` was introduced after this version of the shallow renderer. That's a good point though, because looking at `isValidElementType` catches more uncommon cases that React technically supports and that fail with the shallow renderer.

For example, [you can do](https://codesandbox.io/s/vqz0ryn7xy):

```js
const MemoFragment = React.memo(React.Fragment)
```

But rendering that with the shallow render will trigger this invariant. The same goes for other elements that have their own type like `React.Suspense`, `React.ConcurrentMode`, `React.StrictMode`, etc.

After this is approved I'll submit a follow up that switches to `isValidElementType` and adds test for all those cases.

aweary: Though it seems unlikely that anyone would be shallow rendering `React.memo(React.Fragment)` 😅

aweary:
@gaearon @trueadm how should the shallow renderer handle rendering the element types exported by `React` like `React.StrictMode`, `React.ConcurrentMode`, etc.

```js
shallowRenderer.render(<React.StrictMode>hmm</React.StrictMode>)
```

Currently they fail with the error:

```log
Invariant Violation: ReactShallowRenderer render(): Shallow rendering works only
with custom components, but the provided element type was `symbol`.
```
I think it's correct that it fails (there's no reason to shallow render them directly) but the error about a `symbol` element type is confusing. Should we report the actual symbol name instead? Should we fail with the same error if you try to render `React.memo(React.StrictMode)`?

```
the provided element type was `React.StrictMode`.
```

It looks like we also need to update `jest-react` or whatever utility is responsible for comparing JSX with `expect` because it reports these types as `UNDEFINED`

```
 Expected value to equal:
      <UNDEFINED>lkdol</UNDEFINED>
    Received:
      <UNDEFINED>lol</UNDEFINED>
```

ljharb:
ping on this; I'm going to update enzyme's react 16 adapter so it can handle memo for SFCs, but without this fix it'll still throw for class-based components. It'd be good to close that gap sooner than later.

trueadm:
> I think it's correct that it fails (there's no reason to shallow render them directly) but the error about a `symbol` element type is confusing. Should we report the actual symbol name instead?

I think this is all we should do here. Provide a better error message so they understand that these symbols are not supported.

packages/react-test-renderer/src/ReactShallowRenderer.js
gaearon: This codepath is for classes so should be ok. In fact it should probably use dot access since that's what we do in the real one.

PR #14819, opened 2019-02-11
Proposal: use functional component as a first example in README

alexeyraspopov:
Given the fact that the hooks are released now and allow to do more using functional components, would it be a perfect moment to consider functional components as a default way to build react components? Starting with just README the very first example becomes clearer and no longer pulls classes' complexity.


gaearon:
We'd still need to show classes prominently on the homepage for quite a while. But I'm not opposed to doing this in README.

Issue #14790, opened 2019-02-07
Warning when a memoized component uses both hooks and suspense


trojanowski:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

There is a warning shown in the console for a component which has all the following conditions met:

* it is memoized using `React.memo`
* it uses hooks
* it uses suspense (there is a promise thrown).

Text of that warning: `Warning: ComponentName: Rendered more hooks than during the previous render. This is not currently supported and may lead to unexpected behavior.`

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

https://codesandbox.io/s/71lr16j491

**What is the expected behavior?**

No warning should be shown because the rules of hooks are met. Also, the behavior should be consistent no matter if `React.memo` is used or not.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

React 16.8.1

gaearon:
My first read is this is a bug.

muqg:
I noticed this very same issue in version `16.8.1`. However in my case there is no memoized component between the place where a `Promise` is thrown and the `Suspense` component. 

The component declares two hooks on top, before the `Promise` is thrown and the warning appears. I moved the call throwing the `Promise` on top, before the hooks and the same thing happens. Warning disappears if there is no `Promise` thrown within the component.

A mock of my component is something like:
```js
function Component(props) {
  const [state, setState] = useState()
  useEffect(...)

  // Promise is thrown somewhere after these two
  // hooks on top, but before the render method.
  return (
    ...
  )
}
```

ex3ndr:
Can confirm this.

gaearon:
Fixed in 16.8.2.
https://codesandbox.io/s/vmmkmvo6pl

PR #14821, opened 2019-02-11
Remove false positive warning and add TODOs about `current` being non-null

gaearon:
Failing test for https://github.com/facebook/react/issues/14790.

Seems like we expect that alternate is always empty for Hook "mounts" but that's not the cause if a memo'd component has previously suspended.

Not sure what idiomatic fix is.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -220,6 +220,10 @@ function updateForwardRef(
   nextProps: any,
   renderExpirationTime: ExpirationTime,
 ) {
+  // TODO: current can be non-null here even if the component
+  // hasn't yet mounted. This happens after the first render suspends.
+  // We'll need to figure out if this is fine or can cause issues.
+
   if (__DEV__) {
     if (workInProgress.type !== workInProgress.elementType) {
       // Lazy component props can't be validated in createElement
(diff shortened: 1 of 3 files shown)


sebmarkbage:
`.alternate` should always be null for true mounts. Something may get upgraded but in that case it should disconnect the `.alternate`. So I think something else is broken if that's not the case.

gaearon:
Plain function:

```
First attempt
- mount indeterminate
  - throws
  - (stays indeterminate)

Second attempt
- mount indeterminate
  - current != null so resets alternate // <-- this is why there's no false positive for memo()-less components
  - gets tagged as functional
```

Memo:

```
First attempt
- update memo
  - gets tagged as simple
  - updates simple
    - throws

Second attempt
- updates simple
  - false positive <--- alternate wasn't reset
```

gaearon:
It seems like we use something else as an indicator whether we were suspended and should disconnect the alternate. For classes, we check the instance. For functions, we know we suspended because we're still Indeterminate. But for memo, we don't seem to use any indicator.

gaearon:
Actually looks like classes have a special `IncompleteClassComponent` tag for this.

acdlite:
We should be able to remove this warning entirely, now that we have this invariant:

https://github.com/facebook/react/blob/f24a0da6e0f59484e5aafd0825bb1a6ed27d7182/packages/react-reconciler/src/ReactFiberHooks.js#L488-L493

I think I meant to do this in my PR but I must have forgotten.

PR #14826, opened 2019-02-11
[scheduler] Add method for forcing a lower framerate

n8schloss:
This is a follow up to https://github.com/facebook/react/pull/14746. On very low end devices we may want to force scheduler to run at a lower framerate than 30fps even if the browser/device supports something faster. Lower spec devices may not be able to run enough tasks within 33ms to make meaningful progress, so this adds a function that lets developers force a specific frame rate.

### DIFF
--- a/fixtures/scheduler/index.html
+++ b/fixtures/scheduler/index.html
@@ -105,6 +105,20 @@
         <div><b>Actual:</b></div>
         <div id="test-8"></div>
     </li>
+    <li>
+      <p>Can force a specific framerate</p>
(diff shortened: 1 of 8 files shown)


matthargett:
I'd really like to see this exposed in the regular React build that ends up in React Native as well. One of the teams I work with was asking about this exact functionality.

philipp-spiess:
This seems to be in master now (https://github.com/facebook/react/commit/43c4e5f348eb5704464986e2dc3221e347041b82) so I guess this PR can be closed 🙂

Issue #14793, opened 2019-02-08
contextType apparently not working in dev mode server side rendering


EthanRutherford:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Crash during server render due to `this.context` being undefined. Bug only happens in SSR, and only in dev mode react. Issue appears neither in browser renders, nor in production mode server renders.

note: Using https://reactjs.net/ for c# server-side rendering, with bundled-in version of react disabled (so using the same version of react as the client bundle)

(simplified) application structure
```
// root component
const {Provider, Consumer} = createContext({});
const Root = props => return <Provider value={props.notNull}>{props.children}</Provider>;

// consumer
class ConsumingComponent extends Component {
  static contextType = Consumer;

  render() {
    if (this.context.someValue) return null;
    return <div/>;
  }
}

// render tree
<Root>...<ConsumingComponent/>...</Root>

// output of server render
JsRuntimeException: TypeError: unable to get property 'someValue' of undefined or null reference.
```

**What is the expected behavior?**
`this.context` in the consuming component should be the value passed to the Provider, and not `undefined`.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

The bug, as best as I can pin down, appeared in 16.6.3. The code works in 16.6..1, and 16.6.2 is listed as a broken release in the changelog.

hamlim:
The issue _could_ be that you are providing the `Consumer` as the `static contextType` value. Instead `static contextType` should be assigned the entire context object:

```js
const context = createContext({});
const {Provider, Consumer} = context;
class ConsumingComponent extends Component {
  static contextType = context; // assign `context` here, not `Consumer`

  render() {
    if (this.context.someValue) return null;
    return <div/>;
  }
}
```

I think there should be a warning for this, however it may only be for the client side renderer (i.e. it might not warn within react-dom/server, if that is what the c# rendering implementation uses).

gaearon:
@hamlin Good catch.

EthanRutherford:
That's a bit unfortunate then, I didn't expect I'd have to expose the entire context object. In my setup, I expose the context consumer for other components to use, but `Root` in the example is the only thing which should ever use the provider. 

This still seems like a bug to me, I'd expect consumers of a context to use `context.Consumer`, even if they are using `contextType`.

gaearon:
This is intentionally unsupported but the warning seems necessary.

EthanRutherford:
[...]

PR #12501, opened 2018-03-31
Prefix _context property on returned ReactContext from createContext …

Andarist:
Context 🤣:
https://twitter.com/AndaristRake/status/979835281871441920

`Provider`/`Consumer` are circular (with a comment about it, so I guess it's by design) which makes this possible:
```js
const ctx = React.createContext()
ctx.Consumer.Consumer.Consumer.Consumer.Provider
```

which is funny, but I don't know if anything can be done about it.

cc @acdlite

PR #14831, opened 2019-02-12
Warn for Context.Consumer with contextType

aweary:
Fixes https://github.com/facebook/react/issues/14793

Extends the `contextType` warning to include `Context.Consumer`. For both the `ReactDOM` and `ReactDomServer`.

Since `$$typeof` alone can't be used to distinguish consumers from context objects, the heuristic I used to identify context consumers was to check for the `_context` property which doesn't exist on the context object itself.

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRendererContext.js
+++ b/packages/react-dom/src/server/ReactPartialRendererContext.js
@@ -77,7 +77,10 @@ export function processContext(
   const contextType = type.contextType;
   if (typeof contextType === 'object' && contextType !== null) {
     if (__DEV__) {
-      if (contextType.$$typeof !== REACT_CONTEXT_TYPE) {
+      const isContextConsumer =
+        contextType.$$typeof === REACT_CONTEXT_TYPE &&
+        contextType._context !== undefined;
+      if (contextType.$$typeof !== REACT_CONTEXT_TYPE || isContextConsumer) {
         let name = getComponentName(type) || 'Component';
         if (!didWarnAboutInvalidateContextType[name]) {
           didWarnAboutInvalidateContextType[name] = true;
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/server/ReactPartialRendererContext.js
```diff
@@ -77,16 +77,20 @@ export function processContext(
   const contextType = type.contextType;
   if (typeof contextType === 'object' && contextType !== null) {
     if (__DEV__) {
-      if (contextType.$$typeof !== REACT_CONTEXT_TYPE) {
+      const isContextConsumer =
+        contextType.$$typeof === REACT_CONTEXT_TYPE &&
+        contextType._context !== undefined;
```
gaearon: Was it also called `context` in some patch release?

aweary: Yeah, it was changed to `_context` in https://github.com/facebook/react/pull/12501. Is it worth checking? I think a lot more would be broken if you were trying to render a tree created with a version of React that still used `context`

packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -515,17 +515,21 @@ function constructClassInstance(
   const contextType = ctor.contextType;
   if (typeof contextType === 'object' && contextType !== null) {
     if (__DEV__) {
+      const isContextConsumer =
+        contextType.$$typeof === REACT_CONTEXT_TYPE &&
+        contextType._context !== undefined;
```
gaearon: What if context value actually *is* undefined?

aweary: Isn't the context value tracked as `_currentValue` on this `_context` element?

gaearon: Oh right.

Issue #14840, opened 2019-02-13
Shallow renderer doesn't update on setState() with Hooks


gaearon:
See the issue and the test case with original fix in https://github.com/facebook/react/pull/14802.

We'll need to fix it properly, see https://github.com/facebook/react/pull/14802#issuecomment-463277621.

rodrigopr:
I will try to take another look at this, starting with adjusting comments/code to align with new behavior.
You mentioned it might need a larger refactoring, do you have anything in mind? Should it be a part of this change?

PR #14841, opened 2019-02-13
Support nonRenderPhase updates in ShallowRenderer

trueadm:
Addresses the issue with https://github.com/facebook/react/issues/14840.

Following up from the fix in https://github.com/facebook/react/pull/14802 – which although somewhat correct, it didn't make sense when aligned with the comments. I've addressed the comments and approach so it should fix the original issue whilst keeping non-render phase updates and render-phase updates separate.

### DIFF
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -459,6 +459,8 @@ class ReactShallowRenderer {
         }
         lastUpdate.next = update;
       }
+      // Finally, invoke a new render
+      this._renderer.render(this._renderer._element, this._renderer._context);
     }
   }
 
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -195,6 +195,7 @@ class ReactShallowRenderer {
     this._currentlyRenderingComponent = null;
     this._numberOfReRenders = 0;
     this._previousComponentIdentity = null;
+    this._nonRenderPhaseUpdates = null;
   }
 
   _context: null | Object;
@@ -211,6 +212,7 @@ class ReactShallowRenderer {
   _currentlyRenderingComponent: null | Object;
   _previousComponentIdentity: null | Object;
   _renderPhaseUpdates: Map<UpdateQueue<any>, Update<any>> | null;
+  _nonRenderPhaseUpdates: Map<UpdateQueue<any>, Update<any>> | null;
   _isReRender: boolean;
   _didScheduleRenderPhaseUpdate: boolean;
   _numberOfReRenders: number;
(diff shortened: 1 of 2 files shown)


packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -237,25 +268,34 @@ class ReactShallowRenderer {
...
+            this._nonRenderPhaseUpdates,
+            queue,
+            dispatch,
+          );
+          if (possibleStateAndDispatcher !== null) {
+            this._nonRenderPhaseUpdates = null;
+            // A sanity check to ensure two maps do not get used together
+            invariant(
+              this._didScheduleRenderPhaseUpdate === false,
+              'non-render phase updates should never occur in combination with render phase updates',
```
gaearon: Why would they not?

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -383,35 +423,42 @@ class ReactShallowRenderer {
...
       // This is a render phase update. Stash it in a lazily-created map of
       // queue -> linked list of updates. After this render pass, we'll restart
       // and apply the stashed updates on top of the work-in-progress hook.
       this._didScheduleRenderPhaseUpdate = true;
+      if (this._renderPhaseUpdates === null) {
+        this._renderPhaseUpdates = new Map();
+      }
+      updatesMap = this._renderPhaseUpdates;
+    } else if (componentIdentity === this._previousComponentIdentity) {
+      // This means an update has happened after the function component has
```
rodrigopr: Should it also rerender in this case?

It would be inline with class component behavior, covering tests like
```js
rendered.props.onClick();
expect(shallowRenderer.getRenderOutput()).toEqual(... updated ...);
```
which is what I think a lib like enzyme would do internally

trueadm: I'm not sure how Enzyme works, but in this case, it's triggering an update to the update queue. There is no re-render, which seems more in line with how shallow renderer works (you explicitly call `render`).

rodrigopr: But don't shallow-render `Updater` do this already? (immediately change the state and rerender)
https://github.com/facebook/react/blob/master/packages/react-test-renderer/src/ReactShallowRenderer.js#L154-L159

gaearon: Yeah setState _should_ trigger a render.

trueadm: Ah, my mistake. I've changed the PR to do a re-render.

ericaprieto:
useState seems to be working with this, but useEffect don't, is this intended?

gaearon:
AFAIK even old React lifecycle methods like `componentDidMount` aren't triggered by shallow renderer. They rely on DOM too often.

In general I wouldn't recommend using shallow renderer for anything more involved than asserting on the result. If you want to test `useEffect`, please [follow the recommended testing approach](https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks).

ericaprieto:
That approach feels weird to me... I need to know the internals of every sub component to be able to make assertions, and that becomes really messy as you go up in the tree. If a sub component changes, than my unit test has to change too... I don't know, maybe I'm missing something, but I've never seen an example showing how this approach scales.

packages/react-test-renderer/src/ReactShallowRenderer.js
trueadm: I couldn't think of a case where that would happen, at least in the case they do this invariant would hit. If you can think of a test for it I can enable it and add the logic for that code path :)

gaearon:
This is a long discussion and I'm not sure I'm ready to do it in this PR. :-)

>need to know the internals of every sub component to be able to make assertions

You need to test *what matters to the user*. This approach is described here: https://testing-library.com/docs/react-testing-library/intro.

The user doesn’t care how you split your app into components. If you refactor a `Panel` into `PanelBorder` and `PanelContent` that shouldn’t break your tests because from user’s perspective, nothing changed. Shallow rendering encourages writing tests that are too fragile and that *don’t really test much* — in many cases they mostly test that React itself works. That’s not super useful.

Of course you’re right that you don’t usually want a single test to get too deep. That’s where mocking helps. In Jest, you can do

```js
jest.mock('../MyButton', () => 'button')
```

or

```js
jest.mock('../SomeContent', () => (props) => '(My content)')
```

and that would shim the rendered component with a fake version. That allows you to pick where your test “ends“.  Other test runners may or may not provide similar capabilities.

In either case, this is a long discussion that’s irrelevant to this PR — so let’s have it elsewhere as long as we have a clear question and goals in mind. I’d rather this not turn into a giant argument about testing because people tend to be very opinionated about that.

ericaprieto:
I get what your saying, I thought about that a lot and eventually got to this mocking approach, but I always feel I'm doing wrong, you know? I guess I'm too obsessed with unit testing, maybe I need to change how I think about tests.

Anyway, you are right, this is not the place for this and thank you very much for answering!

lmerotta:
Any news on when this will be merged ?

trueadm:
Fixed in https://github.com/facebook/react/pull/15120

PR #14845, opened 2019-02-14
children can be a element

xp44mm:
children can be a element, so uniform it to array that contains single element.


threepointone:
does this fix some previously broken behaviour? This is an explicit choice made by react to prevent unneeded allocations, and is handled uniformly across the codebase.

xp44mm:
maybe children is so flexable more than i can thought. so i cancel and ended this pull request.

Issue #14849, opened 2019-02-14
Calling setState during component rendering causes incorrect previous state value in next state update


hlousekv:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Counter value is 1.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

https://codesandbox.io/s/7jzwjm090j

**What is the expected behavior?**
Counter value should be 2.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

Firefox/Chrome
16.8.1

gaearon:
This is a bug.

aaronjensen:
I get why this can be surprising, though I was under the impression that one should not call `setState` in a render call (because it's a side effect). Am I missing something?

gaearon:
https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops

agDmitry:
I have exactly [this case](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#alternative-1-reset-uncontrolled-component-with-an-id-prop). Implemented everything as described [here](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops). Spent a couple of hours double-checking everything before stumbling upon this issue. Hope it will be fixed soon because this is a common case in our app.

gaearon:
We always want to fix bugs soon. :-) There’s no need to add extra pressure — we do feel guilty around bugs in the wild and try to fix them as soon as possible.

We expect to cut a fix some time next week.

agDmitry:
> There’s no need to add extra pressure

Sorry, i didn't wanted it to sound like that (i'm still sometimes struggle with speaking/writing english). React team done an amazing work on hooks, thanks all of you for that.

I think that [this](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#alternative-2-reset-uncontrolled-component-with-an-instance-method) would be preferebale solution for avoiding this bug until it will be fixed.

michalkvasnicak:
Is this the same problem as described in this issue? https://codesandbox.io/s/50k546knwk

[...]

PR #14852, opened 2019-02-14
Don't discard render phase state updates with the eager reducer optimization

gaearon:
Reproduces https://github.com/facebook/react/issues/14849 in 2 out of 3 cases (with hoisted reducer and useState but not with inline reducer). Points to a bug in eager update logic. Not sure where it is yet.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -607,7 +607,6 @@ function updateReducer<S, I, A>(
         }
 
         hook.memoizedState = newState;
-
         // Don't persist the state accumlated from the render phase updates to
         // the base state unless the queue is empty.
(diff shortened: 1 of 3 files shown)


gaearon:
Pushed a potential fix in 6f44db2d534406556bf2233308bb93f3323dbd69. I don't really know if it's correct. It would really help to have a written description of how the eager optimization works and what assumptions it makes.

acdlite:
I wrote a partial description of how the optimization is supposed to work here. Maybe we can add more to it:

https://github.com/facebook/react/blob/9b891871ce7e0cf8d25d35dc9dea25a226fc2cc8/packages/react-reconciler/src/ReactFiberHooks.js#L1094-L1117

Specifically, should add that `eagerReducer` and `eagerState` represent the values during the last render attempt. If there's no pending work, then we know those are equal to the current value. Maybe we can think of a better name for those, too.

PR #14855, opened 2019-02-14
include component stack in 'act(...)' warning

threepointone:
Folks aren't seeing stack traces for their  calls in their jest setups, so this PR includes a component stack trace to make it easier to locate and wrap with `act(...)`.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -1815,8 +1815,9 @@ export function warnIfNotCurrentlyBatchingInDev(fiber: Fiber): void {
           '});\n' +
           '/* assert on the output */\n\n' +
           "This ensures that you're testing the behavior the user would see in the browser." +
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1815,8 +1815,9 @@ export function warnIfNotCurrentlyBatchingInDev(fiber: Fiber): void {
           '});\n' +
           '/* assert on the output */\n\n' +
           "This ensures that you're testing the behavior the user would see in the browser." +
-          ' Learn more at https://fb.me/react-wrap-tests-with-act',
+          ' Learn more at https://fb.me/react-wrap-tests-with-act\n%s',
```
gaearon: Is this newline necessary? I thought stack starts with a newline.

threepointone: fair, removed.

PR #14875, opened 2019-02-17
[refactor][client/ReactDOM] convert force Boolean casting to Flow Ano…

ryota-murakami:
Change as title.

### why
I think this is safer than before.

For example, add deliberately value change as a invalid for Logical Expression.

```diff
-    node &&
+  NaN &&
    (node.nodeType === ELEMENT_NODE ||
      node.nodeType === DOCUMENT_NODE ||
```

After that run ` yarn flow dom` command.  
As a result Flow can find illegal values by static analysis before runtime.  
In addition if assigned illegal values into the expression,
result value is forcibly cast to boolean, making debugging difficult.

- console output
```diff
ryota.murakami@ryotas-MacBook-Pro ~/f/react> yarn flow dom
yarn run v1.12.3
$ node ./scripts/tasks/flow.js dom
Running Flow on the dom renderer...
Launching Flow server for /Users/ryota.murakami/fork/react/scripts/flow/dom
Spawned flow server (pid=2118)
Logs will go to /private/tmp/flow/zSUserszSryota.murakamizSforkzSreactzSscriptszSflowzSdom.log
Monitor logs will go to /private/tmp/flow/zSUserszSryota.murakamizSforkzSreactzSscriptszSflowzSdom.monitor_log
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ ../../../packages/react-dom/src/client/ReactDOM.js:457:5

+ Cannot return
+ NaN && ((node.nodeType === ELEMENT_NODE) || (node.nodeType === DOCUMENT_NODE) || (node.nodeType === DOCUMENT_FRAGMENT_NO
+ DE) || ((node.nodeType === COMMENT_NODE) && (node.nodeValue === ' react-mou...'))) because number [1] is incompatible
+ with boolean [2].

     ../../../packages/react-dom/src/client/ReactDOM.js
     454│  */
 [2] 455│ function isValidContainer(node): boolean {
     456│   return (
     457│     NaN &&
     458│     (node.nodeType === ELEMENT_NODE ||
     459│       node.nodeType === DOCUMENT_NODE ||
     460│       node.nodeType === DOCUMENT_FRAGMENT_NODE ||
     461│       (node.nodeType === COMMENT_NODE &&
     462│         node.nodeValue === ' react-mount-point-unstable '))
     463│   );
     464│ }
     465│

     /private/tmp/flow/flowlib_bb4a3db/core.js
 [1]  11│ declare var NaN: number;



Found 1 error
Flow failed for the dom renderer
```

Thank you so much😀

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -452,8 +452,8 @@ ReactRoot.prototype.createBatch = function(): Batch {
  * @return {boolean} True if the DOM is a valid DOM node.
  * @internal
  */
-function isValidContainer(node) {
-  return !!(
+function isValidContainer(node): boolean {
+  return (
     node &&
     (node.nodeType === ELEMENT_NODE ||
       node.nodeType === DOCUMENT_NODE ||


packages/react-dom/src/client/ReactDOM.js
```diff
@@ -452,8 +452,8 @@ ReactRoot.prototype.createBatch = function(): Batch {
  * @return {boolean} True if the DOM is a valid DOM node.
  * @internal
  */
-function isValidContainer(node) {
-  return !!(
+function isValidContainer(node): boolean {
+  return (
```
Ailrun: This is not type-safe since without `!!`, this function can return any falsy value of `node`, like `null`.

ryota-murakami: @Ailurn When exactly the case you said, Flow is so effective.

I create live example,
Could you please look following link🙇‍♂️
https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBTAHgBzgJwBcwBjOAOwGdiBRAGRoFkaA5AFQH0WB5AERrABeMAEYA3JlwFiZKsTY0AGpx78hYAMwTseIqQrUwAYW6Nm7LnwHCAHNql7Zh3tyMBVcyqvqAnPd0yBsQu7p4cAGIASgCCAOJhqtai4uioUACu5CSEAJYUYDmUAGoAhjA5ACZGFIQlOeQY+AAU5HAVGACUAFxgAEZwcDAYJeRgAN6oYGD4GITp+KNNk1Ngre1gAGQby1MtbRgAdGsYbACeOBhCgsL0TKxeagA+jzsrx0f7ZxdXwiEe95Yni8Vm99h92l9Ltdfq5-hYonEEt5nq9du9jpCfsZTEi1FtUaD2uCMKUYOkodcwAByabDbIAWgAtnBMoR6Xh6mzMtQSr0htSOh1lh0JABfdDAYBgRh1cioQqkyrVci1eqNFrpGAwIVAA

aweary:
This isn't equivalent. The `boolean` return type is only a type annotation, it doesn't affect runtime behavior. We still need the `!!` to do the actual type coercion at runtime.

packages/react-dom/src/client/ReactDOM.js
Ailrun: @ryota-murakami The problem is using `null` is **a valid usage** of this function. You cannot remove valid usage for a wrong correctness.

ryota-murakami:
@aweary I understand mechanism of flow.
These are removed and during a transpile by [@babel/plugin-transform-flow-strip-types](https://babeljs.io/docs/en/babel-plugin-transform-flow-strip-types)

PR #14893, opened 2019-02-19
Updating copyright headers, dropping the year

nhunzaker:
Resubmitting @guuibayer's license updates from a non-master branch. It's out of date and I didn't get a response, but I think the licences should be updated.


nhunzaker:
@threepointone Caught two more places in the source. I also noticed that builds are generating the old license, but I can't figure out where that's coming from.

nhunzaker:
Got it. This is coming through the prop-types package:
https://github.com/facebook/prop-types/blob/master/index.js#L2

Want me to update it there as well?

threepointone:
yeah we should update it there too, but there's no hurry

PR #14894, opened 2019-02-20
Add a warning to useRef/useState/useReducer invoked with a missing argument instead of undefined

Jessidhia:
As per the discussion on https://twitter.com/dan_abramov/status/1097844767956918274

### DIFF
--- a/packages/react/src/ReactHooks.js
+++ b/packages/react/src/ReactHooks.js
@@ -80,6 +80,7 @@ export function useReducer<S, I, A>(
 }
 
 export function useRef<T>(initialValue: T): {current: T} {
+  const dispatcher = resolveDispatcher();
   if (__DEV__) {
     // this is the only place where, if the user gives no arguments,
     // the argument will actually be missing instead of just being undefined.
@@ -89,7 +90,6 @@ export function useRef<T>(initialValue: T): {current: T} {
       arguments.length,
     );
   }
-  const dispatcher = resolveDispatcher();
   return dispatcher.useRef(initialValue);
 }
 
--- a/packages/react/src/ReactHooks.js
+++ b/packages/react/src/ReactHooks.js
@@ -67,6 +67,15 @@ export function useContext<T>(
 
 export function useState<S>(initialState: (() => S) | S) {
   const dispatcher = resolveDispatcher();
+  if (__DEV__) {
+    // this is the only place where, if the user gives no arguments,
+    // the argument will actually be missing instead of just being undefined.
+    warning(
+      arguments.length === 1,
+      'Expected exactly one argument to React.useState(initialState), instead received %s.',
+      arguments.length,
+    );
+  }
   return dispatcher.useState(initialState);
 }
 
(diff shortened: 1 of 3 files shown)


packages/react/src/ReactHooks.js
```diff
@@ -81,6 +90,15 @@ export function useReducer<S, I, A>(
...
+  if (__DEV__) {
+    // this is the only place where, if the user gives no arguments,
+    // the argument will actually be missing instead of just being undefined.
+    warning(
+      arguments.length === 1,
+      'Expected exactly one argument to React.useRef(initialValue), instead received %s.',
+      arguments.length,
+    );
+  }
   return dispatcher.useRef(initialValue);
```
Jessidhia: I thought of adding the warnings to `mountRef` / `mountState` instead but, because of how the dispatcher is invoked, by the time things reach them the `arguments.length` will already be `1`. `ReactHooks.js` is the only place that sees the `arguments` exactly as provided by the user.

sebmarkbage:
I don’t know about this one.

Even though we certainly encourage this in a stricter environment like Flow/TypeScript. I feel like it goes against the carefree dynamic nature that might cause someone to stick to plain JavaScript. I don’t know if we should explicitly forbid it like this is doing.

There are also concerns with relying on argument lengths in general. We explicitly always allow undefined to be passed where fewer arguments are allowed. This important when forwarding and wrapping.

E.g. it should always be fine to do `let wrappedUseState = initial => { somethingOther(); return useState(initial); }`

There might be similar practical concerns in the reverse case too.

Jessidhia:
Hmm. Conversely, then, it should be possible to allow you to pass 0 arguments to the typescript types. Through overloads, it is possible to force the result to be and/or include `| undefined`.

I wrote tests for how something like this would work:

```ts
// $ExpectType MutableRefObject<number>
React.useRef(0);

// these are not very useful (can't assign anything else to .current)
// but it's the only safe way to resolve them
// $ExpectType MutableRefObject<null>
React.useRef(null);
// $ExpectType MutableRefObject<undefined>
React.useRef(undefined);

// |null convenience overload
// it should _not_ be mutable if the generic argument doesn't include null
// $ExpectType RefObject<number>
React.useRef<number>(null);
// but it should be mutable if it does (i.e. is not the convenience overload)
// $ExpectType MutableRefObject<number | null>
React.useRef<number | null>(null);

// |undefined convenience overload
// with no contextual type or generic argument it should default to undefined only (not {} or unknown!)
// $ExpectType MutableRefObject<undefined>
React.useRef();
// $ExpectType MutableRefObject<number | undefined>
React.useRef<number>(Math.random() as number | undefined);
// $ExpectType MutableRefObject<number | undefined>
React.useRef<number>();
```

Jessidhia:
Do you have similar thoughts w.r.t. `createContext`? I remember this discussion also coming up with that, but probably because you create contexts a lot less often than you use hooks it's not as high a priority.

sebmarkbage:
The same thing applies to createContext. We don’t enforce arity. But technically you probably should always provide one arg.

It looks like we also made a mistake with createContext because the second argument is not actually meant to be fully supported yet. We might want to warn if it is provided.

Which is what we did for the second argument to useContext. In that case it is because we want to reserve the option to add semantics to that argument in the future. But we don’t technically forbid you to pass more arguments than needed in most other methods.

PR #14900, opened 2019-02-20
Use the canonical nativeTag for Fabric's setNativeProps

TheSavior:
I'm going to be working on migrating the API for setNativeProps and wanted to add a better assertion for the current setNativeProps behavior. I found that with the Fabric renderer the first argument to setNativeProps was `undefined`.

I'm pretty sure we will be changing this later as Fabric won't be calling `UIManager.updateView` at all but this can at least be changed for now.

### DIFF
--- a/packages/react-native-renderer/src/ReactNativeComponent.js
+++ b/packages/react-native-renderer/src/ReactNativeComponent.js
@@ -153,6 +153,7 @@ export default function(
         return;
       }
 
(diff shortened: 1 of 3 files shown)


sebmarkbage:
You need to run `yarn prettier` first but looks good.

TheSavior:
Beat you to it. 👍

PR #14901, opened 2019-02-20
Adding setNativeProps tests for NativeMethodsMixin

TheSavior:
We are going to add a warning for setNativeProps on NativeMethodsMixin so I'm adding a test for the current behavior to be able to check for the warning in a follow up.

I'm also adding a test for the Fabric renderer which required for checking `canonical`. This isn't super relevant because NativeMethodsMixin won't work for Fabric at all, but this will probably help during the migration.

### DIFF
--- a/packages/react-native-renderer/src/__tests__/ReactFabric-test.internal.js
+++ b/packages/react-native-renderer/src/__tests__/ReactFabric-test.internal.js
@@ -12,10 +12,12 @@
 
 let React;
 let ReactFabric;
+let createReactClass;
 let createReactNativeComponentClass;
 let UIManager;
 let FabricUIManager;
 let StrictMode;
+let NativeMethodsMixin;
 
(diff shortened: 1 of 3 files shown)


packages/react-native-renderer/src/__tests__/ReactFabric-test.internal.js
```diff
@@ -30,8 +32,16 @@ describe('ReactFabric', () => {
     ReactFabric = require('react-native-renderer/fabric');
     FabricUIManager = require('FabricUIManager');
     UIManager = require('UIManager');
+    createReactClass = require('create-react-class/factory')(
```
TheSavior: I copied this from https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/react/src/__tests__/createReactClassIntegration-test.internal.js#L24-L27

PR #14906, opened 2019-02-20
Support editable useState hooks in DevTools

bvaughn:
Expose a new `overrideHookState` to be injected into DevTools to enable state and reducer hooks to be editable in DevTools. Also update `ReactDebugHooks` package to support this functionality.

We should release the `react-debug-tools` package soon.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberReconciler.js
+++ b/packages/react-reconciler/src/ReactFiberReconciler.js
@@ -341,6 +341,7 @@ export function findHostInstanceWithNoPortals(
   return hostFiber.stateNode;
 }
 
+let overrideHook = null;
 let overrideProps = null;
 
 if (__DEV__) {
@@ -368,6 +369,24 @@ if (__DEV__) {
     return copyWithSetImpl(obj, path, 0, value);
   };
 
+  // Support DevTools editable hooks state.
+  overrideHook = (
+    fiber: Fiber,
+    nativeHookIndex: number,
+    path: Array<string | number>,
+    value: any,
+  ) => {
+    let currentHook = fiber.memoizedState;
+    while (currentHook !== null && nativeHookIndex > 0) {
+      currentHook = currentHook.next;
+      nativeHookIndex--;
+    }
+    if (currentHook !== null) {
+      let updatedState = copyWithSet(currentHook.memoizedState, path, value);
+      currentHook.queue.dispatch(updatedState);
+    }
+  };
+
   // Support DevTools props for function components, forwardRef, memo, host components, etc.
   overrideProps = (fiber: Fiber, path: Array<string | number>, value: any) => {
     flushPassiveEffects();
@@ -385,6 +404,7 @@ export function injectIntoDevTools(devToolsConfig: DevToolsConfig): boolean {
 
   return injectInternals({
     ...devToolsConfig,
+    overrideHook,
     overrideProps,
     currentDispatcherRef: ReactCurrentDispatcher,
     findHostInstanceByFiber(fiber: Fiber): Instance | TextInstance | null {
--- a/packages/react-reconciler/src/ReactFiberReconciler.js
+++ b/packages/react-reconciler/src/ReactFiberReconciler.js
@@ -372,14 +372,16 @@ if (__DEV__) {
   // Support DevTools editable values for useState and useReducer.
   overrideHookState = (
     fiber: Fiber,
-    index: number,
+    id: number,
     path: Array<string | number>,
     value: any,
   ) => {
+    // For now, the "id" of stateful hooks is just the stateful hook index.
+    // This may change in the future with e.g. nested hooks.
     let currentHook = fiber.memoizedState;
-    while (currentHook !== null && index > 0) {
+    while (currentHook !== null && id > 0) {
       currentHook = currentHook.next;
-      index--;
+      id--;
     }
     if (currentHook !== null) {
       flushPassiveEffects();
--- a/packages/react-reconciler/src/ReactFiberReconciler.js
+++ b/packages/react-reconciler/src/ReactFiberReconciler.js
@@ -372,14 +372,14 @@ if (__DEV__) {
   // Support DevTools editable hooks state.
   overrideHook = (
     fiber: Fiber,
-    nativeHookIndex: number,
+    index: number,
     path: Array<string | number>,
     value: any,
   ) => {
     let currentHook = fiber.memoizedState;
-    while (currentHook !== null && nativeHookIndex > 0) {
+    while (currentHook !== null && index > 0) {
       currentHook = currentHook.next;
-      nativeHookIndex--;
+      index--;
     }
     if (currentHook !== null) {
       let updatedState = copyWithSet(currentHook.memoizedState, path, value);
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -368,6 +369,24 @@ if (__DEV__) {
...
+    value: any,
+  ) => {
+    let currentHook = fiber.memoizedState;
+    while (currentHook !== null && nativeHookIndex > 0) {
+      currentHook = currentHook.next;
+      nativeHookIndex--;
+    }
+    if (currentHook !== null) {
+      let updatedState = copyWithSet(currentHook.memoizedState, path, value);
+      currentHook.queue.dispatch(updatedState);
```
acdlite: This will work for `useState` but not `useReducer`, which I assume is intentional, correct?

sebmarkbage: We really should support useReducer if we support useState since that’s the recommended one. Don’t want people choosing useState over useReducer based on DevTools.

bvaughn: Interesting. I thought a little (not much) about `useReducer` support but didn't think it was necessary.

bvaughn:
> Like we meant to do with props.

I don't think we _meant_ to do this with props. It's something we've discussed but never committed to (at least not that I recall, although I'm forgetful).

I think it could be nice in some cases, but I'm personally not convinced that it's useful often enough to justify the added complexity of building it to be honest. At least not at this stage of the rewrite.

sebmarkbage:
I’d be happy to add the infra myself to help out adding this to props. I’d rather have an approach that is conceptually consistent than relies on subtle details of rerendering that breaks during refactors. We’re dealing with the exact problem for setNativeProps on Fabric now and it’s a pain.

The difference with state is that conceptually you can update it to anything and it is persistent. So we don’t have to do it for state but I haven’t thought much about the tradeoffs yet.

bvaughn:
I remain unconvinced that the added complexity of masking is worth it as a feature. Based on my recent Twitter poll (which isn't necessarily conclusive– but it did have 1,500 votes so I think it's at least worth consideration) only 19% of people use the feature regularly to begin with, and most of them only use it for toggling booleans to e.g. test visual state.

Maybe you and I can talk about this today over a coffee or something and come to a consensus that unblocks me for now @sebmarkbage.

bvaughn:
Okay. I've updated the implementation so that both `useState` and `useReducer` values can be edited. There is an edge case here (similar to editable class props and state) where pending updates might cause odd behavior, but I don't think we need to handle that. I don't think enough people use this feature (or use it for anything more complicated than toggling basic values when iterating locally) that we need to address the more complex possible edge cases.

I also added an explicit `isEditable` param to the hook metadata (which isn't strictly necessary but felt more appropriate). I could combine the `index` and `isEditable` values, but I kind of like the more explicit approach.

bvaughn:
I've pushed a version of the DevTools built with this React change to https://react-devtools-experimental.now.sh/ if you'd like to test out the editable `useState` and `useReducer` values.

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -368,6 +369,27 @@ if (__DEV__) {
     return copyWithSetImpl(obj, path, 0, value);
   };
 
+  // Support DevTools editable values for useState and useReducer.
+  overrideHook = (
```
sebmarkbage: Let's call this something specific to state since there are many other hooks. E.g. might want to override changed bits etc. for debugging context not updating. For Focal we might want to flip the booleans which are not state.

`overrideHookState`?

sebmarkbage: I guess the `isEditable` just indicates that the "value" can be edited as is in whatever form it is. Maybe `overrideHookDebugValue` whatever the representation of debug tools' value is?

bvaughn: Are you suggesting I rename the injected `overrideHook` method to `overrideHookDebugValue`? I think that sounds confusingly close to `useDebugValue` even though they aren't actually related. Maybe I'm misunderstanding you.

`overrideHookState` seems okay though. Probably more clearly indicates its purpose.

bvaughn: I think I've complied with part of your request (renaming `overrideHook` to `overrideHookState`) but I'm still a little fuzzy on the other comment. Would you like me to also rename `isEditable` to something like `isStateEditable`?

bvaughn: Chatted offline and the consensus was this:
* Replace `index` and `isEditable` properties with a single property, e.g. `overrideState` that is either a function or `null`.
* DevTools will pass through the renderer-injected values to `react-debug-tools` when inspecting a fiber.
* For hooks that are editable, `react-debug-tools` will add an `overrideState` function that closes over the necessary info.

bvaughn: Actually implementing the changes above is proving to be kind of complicated, and I'm having second thoughts about it being a good idea.

bvaughn: The thing that makes this change tricky is that it requires maintaining and coordinating between several data structures, all of which are updated async.

### Inspecting hooks

React itself has the hooks list stored in `memoizedState`. These hooks are what we actually need to update.

`ReactDebugHooks` has its own inspected hooks object, consisting of a pointer to the "native" hook, the name (e.g. "State"), value, array of sub-hooks (if it's a custom hook), and an `overrideState` function that closes of things needed to e.g. modify a state or reducer hook.

The frontend needs its own representation, but since this is sent across the Bridge– we need to replace the `overrideState` function with an ID that lets us map back to the original closure. So we send a third structure– with an id, name, value, sub-hooks array, and an `isStateEditable` boolean indicating whether there's an `overrideState` function.

### Overriding state

The frontend starts by sending an "overrideHookState" message with the hook id, path, override value, and the renderer ID (so the backend can find the right injected internals).

The agent uses the renderer ID to pass this info along to the appropriate renderer interface, which needs to be able to find the original inspected hook (from `ReactDebugHooks`, using our ID) so it can call the `overrideState` method.

Then the `overrideState` method needs a way to find the current fiber, and to call the renderer's injected `overrideHookState` method with that fiber along with the "native" hook, path, and value.

bvaughn: @sebmarkbage Let's talk about this again. I don't think this change is a positive one after looking into how I would implement it. I think it adds a lot of complexity without adding enough value to offset it.

bvaughn:
Looks like the approach I was using of updating `baseState` and `memoizedState` and then scheduling sync work stopped working between 16.8 alpha 1 and 16.8.0. The work now doesn't flush until something else triggers an update. Need to do some digging to figure out what changed between the two releases.

bvaughn:
For now I'm just pushing a fix to the bailout issue mentioned in my previous comment and a rename of the inspected hook `isEditable` field to `isStateEditable`.

paramsinghvc:
It is made editable now, but when I change something in the State hook in devtools and press enter, it doesn't seem to reflect on the UI, and neither it's updated when I log using "Log this component data in console" button. Is this a bug?

PR #14907, opened 2019-02-20
Adding ReactNative.setNativeProps that takes a ref

TheSavior:
As part of https://github.com/react-native-community/discussions-and-proposals/issues/72 this adds a new top level API on ReactNative that we can use to setNativeProps.

This can only be called with a ref to a native component and will warn/no-op if called with an invalid component ref.

Other than jest tests, I also verified this worked by running 
```
yarn build native --type=RN_FB_DEV
```
and copied the output files into FB's codebase. I added a console.warn to ensure it was using the renderer I copied in, modified `react-native-implementation.js` to expose this and then verified `setNativeProps` worked correctly from the top level for the paper renderer via a playground app.


packages/react-native-renderer/src/ReactNativeSetNativeProps.js
```diff
@@ -0,0 +1,43 @@
...
+
+import {create} from './ReactNativeAttributePayload';
+import {warnForStyleProps} from './NativeMethodsMixinUtils';
+
+import warningWithoutStack from 'shared/warningWithoutStack';
+
+// Module provided by RN:
+import UIManager from 'UIManager';
+
+export function setNativeProps(handle: any, nativeProps: Object) {
```
TheSavior: Is there a better flow type I should be using for `handle`?

sebmarkbage: We used to pretend that native host instances were ReactNativeComponent. That’s not technically true because they’re not `instanceof`.

With forwardRef, we get the inner type. I would’ve expected that you have to supply the ref type to Flow. What does Flow think the type of a ref is?

Ideally it would be an opaque type which will help catch anyone trying to call something else on it.

sebmarkbage:
Prettier.

packages/react-native-renderer/src/ReactNativeSetNativeProps.js
```diff
@@ -0,0 +1,43 @@
+/**
```
sebmarkbage: Since you’re going to add more of these, can we name it something more general so we don’t need a file per method?

packages/react-native-renderer/src/ReactNativeSetNativeProps.js
TheSavior: Alright, seems like a problem for another PR at another time. :)

PR #14912, opened 2019-02-21
Deprecate ref.setNativeProps in favor of ReactNative.setNativeProps

TheSavior:
Related to: react-native-community/discussions-and-proposals#72

We want people to call the new top level API for setNativeProps that was added in https://github.com/facebook/react/pull/14907. This PR adds deprecation warnings for the current methods. 

I need to think about how I want to handle the migration of core and internal and could use some insight. I could:

- hold off on landing this and instead update core to the new API before any warnings land
- land this and add a yellow box ignore internally as we are converting core. If this change goes out with 0.60 then we have time. We might want to land this in 0.59 to have the warnings out as soon as possible although that would also mean we'd need to cherry pick the fixes for these warnings in core into 0.59

### DIFF
--- a/packages/shared/forks/ReactFeatureFlags.native-fb.js
+++ b/packages/shared/forks/ReactFeatureFlags.native-fb.js
@@ -13,9 +13,7 @@ import typeof * as FeatureFlagsType from 'shared/ReactFeatureFlags';
 import typeof * as FeatureFlagsShimType from './ReactFeatureFlags.native-fb';
 
 // Re-export dynamic flags from the fbsource version.
-export const {
-  debugRenderPhaseSideEffects,
-} = require('ReactFeatureFlags');
+export const {debugRenderPhaseSideEffects} = require('ReactFeatureFlags');
 
 // The rest of the flags are static for better dead code elimination.
 export const enableUserTimingAPI = __DEV__;
--- a/packages/shared/forks/ReactFeatureFlags.native-fb.js
+++ b/packages/shared/forks/ReactFeatureFlags.native-fb.js
@@ -13,7 +13,9 @@ import typeof * as FeatureFlagsType from 'shared/ReactFeatureFlags';
 import typeof * as FeatureFlagsShimType from './ReactFeatureFlags.native-fb';
 
 // Re-export dynamic flags from the fbsource version.
-export const {debugRenderPhaseSideEffects} = require('ReactFeatureFlags');
+export const {
+  debugRenderPhaseSideEffects,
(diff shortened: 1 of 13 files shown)


cpojer:
How long will it take to convert all of RN Core to use the new API? I feel like adding a warning for everyone in the community and silencing them internally isn't great :(

TheSavior:
Well, just loading the internal Playground view with this change popped up 115 warnings. Sounds like we should migrate the core before landing this.

TheSavior:
I'm thinking that I will probably want to land this and add the warning to the yellowbox ignores for core RN so that nobody sees the error but it will be easy for us to comment out the ignore line and fix them. Right now in order to see these warnings we have to do a sync first which is painful.

sebmarkbage:
We do have separate builds from FB and open source. If you want, you could add this behind a feature flag that we only enable for FB.

Add it here: https://github.com/facebook/react/blob/master/packages/shared/ReactFeatureFlags.js

Enable here: https://github.com/facebook/react/blob/master/scripts/rollup/shims/react-native-fb/ReactFeatureFlags.js

packages/shared/forks/ReactFeatureFlags.native-fb.js
```diff
@@ -27,6 +27,7 @@ export const debugRenderPhaseSideEffectsForStrictMode = true;
 export const disableInputAttributeSyncing = false;
 export const replayFailedUnitOfWorkWithInvokeGuardedCallback = __DEV__;
 export const warnAboutDeprecatedLifecycles = true;
+export const warnAboutDeprecatedSetNativeProps = false;
```
sebmarkbage: Hm. Is this the canonical one?

TheSavior: What do you mean by canonical? It seems like to get Flow to pass these consts have to be defined in every file

TheSavior: Ah, nvm. I understand. I need to re-export the one from FeatureFlags. I'm testing this change more closely to ensure I can actually flip this on in FBSource.

sebmarkbage: I think we should actually remove it from the dynamic one if possible. Whenever possible we should always use static flags and drive to remove the dynamic ones.

TheSavior:
Verified that these warnings are now on in FB:
simulator screen shot - iphone xs - 2019-02-25 at 12 32 36

Issue #14904, opened 2019-02-20
controlled input cursor jumps to end (again)


ljcrapo:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

when typing in a controlled input, the cursor always jumps to the end. This was an old issue that seems to have resurfaced.

[this code pen](https://codepen.io/gaearon/pen/VmmPgp?editors=0010) used in the docs [here](https://reactjs.org/docs/forms.html#controlled-components)  has the problem in all browsers as far as I have been able to test.

**What is the expected behavior?**

because we are using the state to update the component as soon as it's changed, the input element should be able to keep the cursor in the same place.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

I'm at latest (16.8.2) and I tested on Chrome, FireFox, and Edge on Windows

as far as I know, this was working at some point, though I don't know how long ago. possibly even before "Fiber"

gaearon:
This is quite weird. Same code works for me on CodeSandbox: https://codesandbox.io/s/vqn79m09r3

Not sure what's going on here.

gaearon:
Seems UMD-specific.

gaearon:
Thank you very much for reporting this. It's a serious regression and exposed some lacking coverage in our test suite regarding UMD bundles. I have a fix ready in https://github.com/facebook/react/pull/14914.

ljcrapo:
Great! thank you for the fix.

gaearon:
This is fixed in 16.8.3. Unfortunately unpkg doesn't seem to pick up the 16.8.3 version yet from its default redirect.

ljcrapo:
That's OK, I replaced the resources with the specific release artifacts to test and it works great now. 👍

ranneyd:
> This is quite weird. Same code works for me on CodeSandbox: https://codesandbox.io/s/vqn79m09r3
> 
> Not sure what's going on here.

@gaearon it flipped for me and I'm actually having the same problem OP reported in _your_ CodeSandbox, but _not_ in _his_ codepen. Should that be?

Chrome 74

Manubi:
I've updated from react 16.8.1 to 16.8.6. Still having the same problem...

CodeMedic42:
I am running into this same issue. I have created the following fiddle which is a boiled down version of what I have.

https://jsfiddle.net/CodeMedic42/87f9gnow/

[...]

PR #14914, opened 2019-02-21
Fix UMD builds by re-exporting the scheduler priorities

gaearon:
Closes https://github.com/facebook/react/issues/14904.

Fixes the regression to UMD builds since https://github.com/facebook/react/pull/14756 which caused all interactive updates in 16.8.2 to be scheduled with low priority.

bug

We had a test that checked for those but it used to ignore priorities because it wasn't important at the time. I'm making the test check those too.

Since the existing test *also* verifies the Scheduler UMD bundles, I added getters to those to make sure the API matches up. But that's not the important part of this fix. The important part is just adding those constants to "scheduler internals" for UMD.

fix

Note this means that even ReactDOM 16.8.3+ won't work correctly with older React UMDs. This might be surprising but it breaks often enough for UMDs that it's probably okay.

### DIFF
--- a/packages/react/src/ReactSharedInternals.js
+++ b/packages/react/src/ReactSharedInternals.js
@@ -18,6 +18,11 @@ import {
   unstable_continueExecution,
   unstable_wrapCallback,
   unstable_getCurrentPriorityLevel,
+  unstable_IdlePriority,
+  unstable_ImmediatePriority,
+  unstable_LowPriority,
+  unstable_NormalPriority,
+  unstable_UserBlockingPriority,
 } from 'scheduler';
 import {
   __interactionsRef,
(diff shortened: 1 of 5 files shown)


packages/react/src/ReactSharedInternals.js
```diff
@@ -60,6 +65,11 @@ if (__UMD__) {
       unstable_pauseExecution,
       unstable_continueExecution,
       unstable_getCurrentPriorityLevel,
+      unstable_IdlePriority,
+      unstable_ImmediatePriority,
+      unstable_LowPriority,
+      unstable_NormalPriority,
+      unstable_UserBlockingPriority,
```
gaearon: This is the important part of the fix. The rest isn't.

nhunzaker:
I built a version of the fixtures with this change:
https://react-dom-umd-fix.surge.sh/text-inputs

Looks good to me, :+1:

PR #14916, opened 2019-02-21
[ESLint] Warn against assignments from inside Hooks

gaearon:
These are usually bugs and work by accident. You're supposed to use refs instead.

I thought about pulling this into its own rule. But then it's niche enough that it's hard to come up with a name that makes sense. And when it fires, the exhaustive deps warning becomes confusing. So I decided to make this mistake "cancel out" the exhaustive deps warning until you clarify the intent.

screen shot 2019-02-21 at 5 16 52 pm

I plan to add a few more "niche" cases into this rule. Such as using `ref.current` in cleanup phase.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -1617,6 +1617,7 @@ const tests = {
           useEffect(() => {
             value = 42;
             value2 = 100;
+            value = 43;
             console.log(value2);
             console.log(value3);
             setTimeout(() => {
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -1608,6 +1607,124 @@ const tests = {
...
+          }, [value, value2, value3]);
+        }
+      `,
+      errors: [
+        // value
+        `Assigning a variable from inside a React useEffect Hook ` +
+          `will not persist between re-renders. ` +
+          `If it's only needed by this Hook, move the variable inside it. ` +
+          `Alternatively, declare a ref with the useRef Hook, ` +
+          `and keep the mutable value in its 'current' property.`,
```
gaearon: I could include the name but this point to a specific place so I figured it's not super important. Could add it tho.

Issue #14920, opened 2019-02-21
[ESLint] Feedback for 'exhaustive-deps' lint rule


gaearon:
## What is this

This is a new ESLint rule that verifies the list of dependencies for Hooks like `useEffect` and similar, protecting against the stale closure pitfalls. For most cases it has an autofix. We'll add more documentation over the next weeks.

![autofix demo](https://user-images.githubusercontent.com/810438/53197370-f7b9be80-3611-11e9-9d97-fc69285000d0.gif)

## Installation

```
yarn add eslint-plugin-react-hooks@next
# or
npm install eslint-plugin-react-hooks@next
```

ESLint config:

```js
{
  "plugins": ["react-hooks"],
  // ...
  "rules": {
    "react-hooks/rules-of-hooks": 'error',
    "react-hooks/exhaustive-deps": 'warn' // <--- THIS IS THE NEW RULE
  }
}
```

Simple test case to verify the rule works:

```js
function Foo(props) {
  useEffect(() => {
    console.log(props.name);
  }, []); // <-- should error and offer autofix to [props.name]
}
``` 
## The lint rule complains but my code is fine!

**If this new `react-hooks/exhaustive-deps` lint rule fires for you but you think your code is correct**, please post in this issue.

----

# BEFORE YOU POST A COMMENT

**Please** include these three things:

1. A CodeSandbox demonstrating a **minimal code example** that still expresses your intent (not "foo bar" but **actual UI pattern** you're implementing).
2. An explanation of the **steps** a user does and what you expect to see on the screen.
3. An explanation of the **intended API** of your Hook/component.

![please](https://media1.giphy.com/media/14cNSEhRyTX8Ag/giphy.gif?cid=3640f6095c700c9b774a644a51e2089e)

## But my case is simple, I don't want to include those things!

It might be simple to you — but it’s not at all simple to us. **If your comment doesn't include either of them (e.g. no CodeSandbox link), we will hide your comment** because it’s very hard to track the discussion otherwise. Thank you for respecting everyone’s time by including them.

The end goal of this thread is to find common scenarios and transform them into better docs and warnings. This can only happen when enough details are available. Drive-by comments with incomplete code snippets significantly drive down the quality of the discussion — to the point that it's not worth it.

billyjanitsch:
[CodeSandbox](https://codesandbox.io/s/nqy69ol00)

[...]

PR #14930, opened 2019-02-22
[ESLint] Add more cases to exhaustive-deps rule

gaearon:
* Adds a better message for when you try to pass literals (saw some confusion in https://github.com/facebook/react/issues/14920#issuecomment-466226093).
* Adds a new warning for accessing `ref.current` in effect cleanup (for example, https://github.com/facebook/react/issues/14920 suffers from this bug). The heuristic is that we only warn when you see you don't assign `ref.current = ` yourself. Then we assume it’s managed by React.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -446,6 +446,122 @@ const tests = {
         }
       `,
     },
+    {
+      // Valid because we assign ref.current
+      // ourselves. Therefore it's likely not
+      // a ref managed by React.
+      code: `
+        function MyComponent() {
+          const myRef = useRef();
+          useEffect(() => {
+            const handleMove = () => {};
+            myRef.current = 42;
+            return () => {
+              console.log(myRef.current.toString())
+            };
+          }, []);
+          return <div />;
+        }
+      `,
+    },
+    {
+      // Valid because we assign ref.current
+      // ourselves. Therefore it's likely not
+      // a ref managed by React.
+      code: `
+        function useMyThing(myRef) {
+          useEffect(() => {
+            const handleMove = () => {};
+            myRef.current = 42;
+            return () => {
+              console.log(myRef.current.toString())
+            };
+          }, [myRef]);
+        }
+      `,
+    },
+    {
+      // Valid because the ref is captured.
+      code: `
+        function MyComponent() {
+          const myRef = useRef();
+          useEffect(() => {
+            const handleMove = () => {};
+            const node = myRef.current;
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -446,6 +446,122 @@ const tests = {
...
+            return () => node.removeEventListener('mousemove', handleMove);
+          }, []);
+          return <div ref={myRef} />;
+        }
+      `,
+    },
+    {
+      // Valid because we assign ref.current
+      // ourselves. Therefore it's likely not
+      // a ref managed by React.
```
bvaughn: This comment and the code below seem out of sync. Maybe bad copy-pasta? Maybe should mirror the comment above?

> Valid because the ref is captured.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -197,6 +182,47 @@ export default {
...
+          }
+          // We only want to warn about React-managed refs.
+          if (foundCurrentAssignment) {
+            return;
+          }
+          context.report({
+            node: dependencyNode.parent.property,
+            message:
+              `Accessing '${dependency}.current' during ` +
+              `the effect cleanup is likely a mistake because by this time React ` +
```
eps1lon: A cultural suggestion: Explain what is happening and why rather than telling someone directly they made a mistake. I suspect people might get overly defensive if the computer tells them they made a mistake.
```suggestion
              `the effect cleanup will likely cause a TypeError because by this time React ` +
```
No opinion what should we call it e.g. "null pointer exception", "reading a property from `null`" etc.

gaearon: TypeError isn't the main problem (easy to work around). A bigger problem is that you'll always keep getting the *previous* ref value since it's being read too early. The challenge is to explain this without freaking people out about how "complex" it is.

gaearon: Oops, I misread. This is a different warning from the one I was thinking about. In this case the problem is that you're reading the *next* ref value when you meant to access the *previous* one.

eps1lon: Right, maybe 
```suggestion
              `the effect cleanup will likely use a different ref value because by this time React ` +
```

My point was more about not actually saying that this "is a mistake". That's already implied by reporting that a lint rule has been violated. As far as I know the other messages don't add this is likely a mistake either.

gaearon: Sounds reasonable.

PR #14931, opened 2019-02-22
[scheduler] Pass didTimeout argument to callbacks

acdlite:
As I prepare to refactor the Fiber scheduler, I've noticed some quirks in our implementation. This PR addresses one of them.

---

There's no reason for a timed out Scheduler callback to check `shouldYield`, because the value will always be false until the work has completed. The `didTimeout` argument provides this information to the callback so it can avoid the redundant checks.

React's existing check for whether a callback has timed out didn't make any sense, but happened to work anyway. I don't think the wrongness of the old implementation is observable via React APIs but it's incoherent regardless.


NE-SmallTown:
Any info about the refactor?

PR #14938, opened 2019-02-23
Fix debug tools crashing on inspecting a component using useContext

bdbch:
This pull request should fix a bug with the `useContext` function breaking the devtools because the useContext function is  not calling the `nextHook()` function.

Example Sandbox:
https://codesandbox.io/s/z2qok540m

Related issue:
https://github.com/facebookincubator/redux-react-hook/issues/34


bvaughn:
I caught the issue in a test and added a fix via #14940

PR #14940, opened 2019-02-24
Fixed incompatibility between react-debug-tools and useContext()

bvaughn:
An issue was reported in facebookincubator/redux-react-hook/issues/34 where DevTools caused a runtime error when inspecting a component that used a `useContext` hook. I determined the cause to be due to the fact that [the `useContext` hook inside of `ReactDebugHooks` doesn't call `nextHook()` to advance the list](https://github.com/facebook/react/blob/ba708fa79b3db6481b7886f9fdb6bb776d0c2fb9/packages/react-debug-tools/src/ReactDebugHooks.js#L92-L102), which causes subsequent hooks to be mismatched.

Initially, I thought the fix would be to simply add that call– _but_ React itself is inconsistent with how it treats `useContext` between development and production builds. This presents a problem for the `react-debug-tools` package– it either breaks in development or production mode.

I've addressed this by refactoring our hooks ordering checks to use a separate, DEV only list (stored on fibers as `_debugHookTypes`). This way we don't have to rely on a hook being added to the hooks list in order to be validated, and code like `ReactDebugHooks` does not have to worry about inconsistent behavior between DEV and PROD bundles, (and we avoid adding additional overhead to PROD bundles).

This changed caught a couple of additional warnings in existing tests. I've beefed out or test coverage in this area as well for going forward.

A few alternative fixes were considered:
1. Update the `useContext` implementation in React to use `mountContext`/`updateContext` (instead of calling `readContext` directly) so that it's _consistent_ between modes.
   * **Cons**: This adds some small additional overhead in production mode that is otherwise unnecessary (and may not be beneficial to the majority of users). This also does not fix DevTools for the existing 16.8 releases.
1. DevTools disables hooks inspection in production mode.
   * **Cons**: This makes DevTools slightly less useful, but that's already the case in production mode due to minification/mangling.
   * **Pros**: This is a backwards compatible fix (meaning `react-debug-tools` will work with earlier hook releases too).
1. DevTools passes the `buildType` flag to `inspectHooks` (telling it whether React is running in production mode). `ReactDebugHooks.useContext` then conditional calls `nextHook()` based on this flag.
   * **Cons**: This feels a little fragile and the API feels a little awkward (since there's already a third optional param).
   * **Pros**: This is a backwards compatible fix.

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -93,6 +93,10 @@ function useContext<T>(
   context: ReactContext<T>,
   observedBits: void | number | boolean,
 ): T {
+  if (__DEV__) {
+    // ReactFiberHooks only adds context to the hooks list in DEV.
+    nextHook();
+  }
   hookLog.push({
     primitive: 'Context',
     stackError: new Error(),
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -93,7 +93,6 @@ function useContext<T>(
   context: ReactContext<T>,
   observedBits: void | number | boolean,
 ): T {
-  nextHook();
   hookLog.push({
     primitive: 'Context',
     stackError: new Error(),
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -93,10 +93,7 @@ function useContext<T>(
   context: ReactContext<T>,
   observedBits: void | number | boolean,
 ): T {
-  if (__DEV__) {
-    // ReactFiberHooks only adds context to the hooks list in DEV.
-    nextHook();
-  }
+  nextHook();
   hookLog.push({
     primitive: 'Context',
     stackError: new Error(),
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -182,22 +182,25 @@ let currentHookNameInDev: ?HookType = null;
 // The list stores the order of hooks used during the initial render (mount).
 // Subsequent renders (updates) reference this list.
 let hookTypesDev: Array<HookType> | null = null;
+let hookTypesUpdateIndexDev: number = -1;
 
-// In DEV, this index is used to compare the order of hooks between update and mount.
-// The index is null during mount (to distinguish between mount and update).
-let hookTypesUpdateIndexDev: number | null = null;
-
-function checkHookTypesDev() {
+function mountHookTypesDev() {
   if (__DEV__) {
     const hookName = ((currentHookNameInDev: any): HookType);
 
-    if (hookTypesUpdateIndexDev === null) {
-      if (hookTypesDev === null) {
-        hookTypesDev = [hookName];
-      } else {
-        hookTypesDev.push(hookName);
-      }
-    } else if (hookTypesDev !== null) {
+    if (hookTypesDev === null) {
+      hookTypesDev = [hookName];
+    } else {
+      hookTypesDev.push(hookName);
+    }
+  }
+}
+
+function updateHookTypesDev() {
+  if (__DEV__) {
+    const hookName = ((currentHookNameInDev: any): HookType);
+
+    if (hookTypesDev !== null) {
       hookTypesUpdateIndexDev++;
       if (hookTypesDev[hookTypesUpdateIndexDev] !== hookName) {
         warnOnHookMismatchInDev(hookName);
@@ -323,7 +326,7 @@ export function renderWithHooks(
       current !== null
         ? ((current._debugHookTypes: any): Array<HookType>)
         : null;
-    hookTypesUpdateIndexDev = hookTypesDev === null ? null : -1;
+    hookTypesUpdateIndexDev = -1;
   }
 
   // The following should have already been reset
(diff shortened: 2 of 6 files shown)


bdbch:
Great! Thanks a lot!

smapress:
Well done! Thx!

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -93,6 +93,7 @@ function useContext<T>(
   context: ReactContext<T>,
   observedBits: void | number | boolean,
 ): T {
+  nextHook();
```
threepointone: not sure if relevant, but we add the workinprogresshook to the list only when __DEV__. so `hook.next` behaves differently in prod vs dev https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberHooks.js#L527-L529

bvaughn: Good catch. My test had a blindspot here since the null hook would just re-initialize the state.

bvaughn: Hmm... 🤔 I think this might be trickier than a simple `__DEV__` check can account for. It's not a matter of whether `react-debug-tools` is itself in production or development mode, but whether the renderer it's interacting with is.

threepointone:
Maybe we can change it to add the hook to the list whether dev or prod. I remember adding it only so we could do the warning for when the order doesn’t match across renders. It doesn’t serve any other purpose.

bvaughn:
Yeah, I'm looking at that now.

bvaughn:
Previously in production mode, `useContext` pointed directly at `readContext`. I've changed it to point to `mountContext`/`updateContext` instead (and I've changed those to both call `mountWorkInProgressHook`/`updateWorkInProgressHook` unconditionally). This adds additional overhead in production mode that would be nice to avoid, but without it– the `react-debug-tools` package can't properly inspect components in (react) production mode.

If this change is unacceptable, then we have a few alternatives:
1. Disable hooks inspection in production mode in DevTools. (This would be a bit unfortunate but not the end of the world.)
1. Pass a param to `react-debug-tools` telling it whether React is running in production mode (aka the renderer's `bundleType`) and then base the `useContext` conditional `nextHook()` call on that. (This would be kind of hacky but avoids adding overhead to React in production mode for a feature that might not be used by many people.)

I would be interested in hearing @sebmarkbage's thoughts on this.

gaearon:
I'd err on the side of keeping context calls very cheap and accepting some integration fragility (DevTools pass the DEV flag) as a tradeoff. With a test that shouldn't even be that fragile.

sebmarkbage:
My M.O. in any case like this is to error on the side of adding more implementation complexity to the DEV side and keep the prod side clean.

In this case, let’s take a step back and ask why DEV differs.

It’s easy to assume that context is the odd one out here and that all other hooks goes into the hooks list. That’s not true conceptually though.

Some of the hooks we have now don’t have to be stateful (eg useEffect without a dep list nor destructor). In the future, the hooks we want to add will also not be stateful.

So the list we have right now is not so much the canonical “hooks list”. It’s only the “stateful hooks list”. In fact, in the future some of them might have more than one entry.

I don’t know exactly but I’m guessing that the only reason @acdlite added context to this set in DEV is that it makes it easier to issue warnings that enforce call order and preserve our option to keep them. However, by reusing the production mechanism for this, I think we’ve also accidentally introduced a bug (prod/dev behavior) that is observable through the invariant calls that only happen in one path.

The lesson learned from “current owner” is that it’s a bad idea to rely on subtle production behavior for dev warnings because they’re not always going to overlap. That’s why we have “current debug frame” instead for the warning info.

I think it would be appropriate to do the same here if we want to preserve some of the dev warnings. If we want to conceptually model the warnings as the user facing api contract (every primitive hook call is one hook in that exact order). Then we should explicitly model that in dev. Which requires dev to have a separate hooks list that overlaps mostly with the stateful hooks list (but not exactly).

bvaughn:
> I think it would be appropriate to do the same here if we want to preserve some of the dev warnings. If we want to conceptually model the warnings as the user facing api contract (every primitive hook call is one hook in that exact order). Then we should explicitly model that in dev. Which requires dev to have a separate hooks list that overlaps mostly with the stateful hooks list (but not exactly).

Interesting. I hadn't considered this option, but I think I dig it. I'll look into it this morning.

bvaughn:
Looks like the DEV warning is already broken for any hook types other than `useState` and `useReducer`. Rather than our DEV warning, an invariant will be triggered:

> Invariant Violation: Should have a queue. This is likely a bug in React. Please file an issue.

I'm going to expand our hook ordering test to make sure it covers all of the hook types.

bvaughn:
Okay. I've refactored this to a DEV-only list of hook names to verify ordering. I've also updated our tests to cover more cases (including several that failed prior to this change).

There are a couple of potential follow up things we could do (which I left TODO comments for).

bvaughn:
Rebased and updated DEV mode to also check for mount vs update using the `nextCurrentHook`.

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
```diff
@@ -997,16 +995,23 @@ describe('ReactHooks', () => {
...
+      'Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks',
+      'Warning: React has detected a change in the order of Hooks called by App. ' +
+        'This will lead to bugs and errors if not fixed. For more information, ' +
+        'read the Rules of Hooks: https://fb.me/rules-of-hooks\n\n' +
+        '   Previous render            Next render\n' +
+        '   ------------------------------------------------------\n' +
+        '1. useReducer                 useReducer\n' +
+        '2. useState                   useRef\n' +
+        '   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n\n',
+    ]);
```
bvaughn: This change wasn't strictly necessary but it seems like this test is a bit more robust (and reads better) if it uses our `toWarnDev` check.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1216,20 +1248,120 @@ if (__DEV__) {
...
+  HooksDispatcherOnMountWithHookTypesInDEV = {
+    readContext<T>(
+      context: ReactContext<T>,
+      observedBits: void | number | boolean,
+    ): T {
+      return readContext(context, observedBits);
+    },
+
+    useCallback<T>(callback: T, deps: Array<mixed> | void | null): T {
+      currentHookNameInDev = 'useCallback';
```
acdlite: I think we can get rid of this global now, right? By passing the name directly to the function on the next line.

bvaughn: No, we still use it in the `areHookInputsEqual` function.

acdlite: Oh right, forgot it was used there, too. Never mind then!

bdbch:
👏

bvaughn:
The fix for this issue was released with v16.8.4

Raphael20735:
Admin

Em ter, 26 de fev de 2019 15:58, Brian Vaughn <notifications@github.com>
escreveu:

PR #14853, opened 2019-02-14
await act(async () => ...)

threepointone:
TODO - 
- [x] ReactDOM, TestRenderer, NoopRenderer
- [x] tests for other renderers, I missed them last time around 
- [x] the warnings are placeholders, still thinking about what would be better, open to suggestions
- [x] implications on concurrent mode

acting

(repasting from the group)

I hacked together an asynchronous version of `act(...)`, and it's kinda nice. 

You've seen the synchronous version -

```js
act(() => {
  // updates and stuff
})
// make assertions 
```

This still works, and gives the same warnings. But if you pass an `async` function - 
```js
await act(async() => {
  // updates and stuff
})
// expect commits and effects to be flushed 
// make assertions 
```
Neat! I set it up so if you *don't* `await` the result from `act`, a warning gets triggered (with `setImmediate`) to do so. That makes it a bit harder to have rogue async `act()` calls in the ether.

You can nest `act()` calls - 

```js
await act(async() => {
  // nest synchronous calls
  act(() => {
    // updates and such    
  })
  // as before, updates and effects are flushed 

  // make assertions 
  await sleep(500) // or for a promise to resolve, or whatever
  // more assertions maybe 

  // nest asynchronous calls too 
  await act(async () => {
    // mutations and such 
    // more awaits 
    // and maybe an event or two
  })
  // more assertions 
})
```

I implemented a cheap form of unrolling safety, so if a previous `act()` gets closed before any subsequent `act()` calls, a warning gets triggered. This should prevent most interleaving attempts, and maintain a tree-like shape of `act()` blocks. 

pros -
- works with async/await, solves oss problems cleanly 
- the sync api is preserved 
- the warning is preserved
- works with fake timers/fb 

cons - 
- can't guarantee batching after the first await in an act block (this will get better when concurrent?)
- less restrictive than the sync model, and starts to feel more opt-in than opt-out (eg- someone could just wrap their entire test with an `act()` call... which might be fine?)
- exposes a secret api on react dom to implement it, dunno how you feel about that

PR #14955, opened 2019-02-26
Warn if ReactNoop.act is not awaited

acdlite:
Based on #14952

Updates the noop renderer tests to conform to the changes coming in #14853. It overlaps slightly with that PR but it doesn't implement the logic to warn when you nest async `act` calls; it only adds a warning if you call `act` without awaiting it.

I'm not sure if we should merge this PR now or if we should wait for #14853 to land first. It may make sense to merge this one and then rebase Sunil's forthcoming changes on top of that.

In other words, @threepointone, feel free to ignore this PR if you like — it only exists to potentially make your life easier.

### DIFF
--- a/packages/react-noop-renderer/src/createReactNoop.js
+++ b/packages/react-noop-renderer/src/createReactNoop.js
@@ -538,6 +538,8 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
   let didStop: boolean = false;
   let expectedNumberOfYields: number = -1;
 
+  let actDepth = 0;
+
   function shouldYield() {
     if (
       expectedNumberOfYields !== -1 &&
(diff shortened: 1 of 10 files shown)


packages/react-noop-renderer/src/createReactNoop.js
```diff
@@ -832,38 +830,52 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
...
+          if (
+            result !== undefined &&
+            result !== null &&
+            typeof result.then === 'function'
+          ) {
+            result.then(() => {
+              actDepth--;
+              if (actDepth === 0) {
+                // TODO: Should keep flushing until queue is empty
+                ReactNoop.unstable_flushWithoutYielding();
```
threepointone: this would simplify tests too (which I assume was the reason you did this). til, thanks!

acdlite:
Closing in favor of #14853

PR #14964, opened 2019-02-27
Add new mock build of Scheduler with flush, yield API

acdlite:
Test environments need a way to take control of the Scheduler queue and incrementally flush work. Our current tests accomplish this either using dynamic injection, or by using Jest's fake timers feature. Both of these options are fragile and rely too much on implementation details.

In this new approach, we have a separate build of Scheduler that is specifically designed for test environments. We mock the default implementation like we would any other module; in our case, via Jest. This special build has methods like `flushAll` and `yieldValue` that control when work is flushed. These methods are based on equivalent methods we've been using to write incremental React tests. Eventually we may want to migrate the React tests to interact with the mock Scheduler directly, instead of going through the host config like we currently do.

For now, I'm using our custom static injection infrastructure to create the two builds of Scheduler — a default build for DOM (which falls back to a naive timer based implementation), and the new mock build. I did it this way because it allows me to share most of the implementation, which isn't specific to a host environment — e.g. everything related to the priority queue. It may be better to duplicate the shared code instead, especially considering that future environments (like React Native) may have entirely forked implementations. I'd prefer to wait until the implementation stabilizes before worrying about that, but I'm open to changing this now if we decide it's important enough.


scripts/jest/config.build.js
```diff
@@ -31,12 +31,20 @@ moduleNameMapper[
...
-  // Named entry points
-  moduleNameMapper[
-    `^${name}/(.*)$`
-  ] = `<rootDir>/build/node_modules/${name}/$1`;
+  if (name === 'scheduler') {
+    // Scheduler is a special case because we mock it by default, but unmock
+    // for specific modules.
+    moduleNameMapper['^scheduler$'] = `<rootDir>/build/node_modules/scheduler`;
+    moduleNameMapper['^scheduler/tracing$'] =
+      '<rootDir>/build/node_modules/scheduler/tracing';
```
sebmarkbage: How is this not just the same as below? Can you get rid of this special case?

acdlite: Yeah, got it working by changing the regex to only match the top level directory per package.

PR #14968, opened 2019-02-27
Avoid dynamic dispatch for scheduler calls

gaearon:
I noticed our prod bundles have code like:

screen shot 2019-02-27 at 4 00 48 pm

This is because `scheduler` is a CommonJS external module. Rollup calls named exports on these with dot notation. This seems very non-ideal in hot paths.

I'm not sure if there's an easy way to fix it on the build level. But scheduler is the only external module we use and it's easy to fix at callsites. So I just did that. Star imports for everything else stay an anti-pattern.

This also removes some re-export junk in the UMD bundle.

Fixed version:

screen shot 2019-02-27 at 4 02 16 pm

### DIFF
--- a/packages/react-art/src/ReactARTHostConfig.js
+++ b/packages/react-art/src/ReactARTHostConfig.js
@@ -5,21 +5,23 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import {
-  unstable_scheduleCallback as scheduleDeferredCallback,
-  unstable_cancelCallback as cancelDeferredCallback,
-} from 'scheduler';
-export {
-  unstable_now as now,
-  unstable_scheduleCallback as scheduleDeferredCallback,
-  unstable_shouldYield as shouldYield,
-  unstable_cancelCallback as cancelDeferredCallback,
-} from 'scheduler';
 import Transform from 'art/core/transform';
 import Mode from 'art/modes/current';
+import * as Scheduler from 'scheduler';
 import invariant from 'shared/invariant';
 
 import {TYPES, EVENT_TYPES, childrenAsString} from './ReactARTInternals';
+
+// Intentionally not named imports because Rollup would
+// use dynamic dispatch for CommonJS interop named imports.
+const {
+  unstable_now: now,
+  unstable_scheduleCallback: scheduleDeferredCallback,
+  unstable_shouldYield: shouldYield,
+  unstable_cancelCallback: cancelDeferredCallback,
+} = Scheduler;
+
+export {now, scheduleDeferredCallback, shouldYield, cancelDeferredCallback};
 
 const pooledTransform = new Transform();
 
--- a/packages/react-cache/src/LRU.js
+++ b/packages/react-cache/src/LRU.js
@@ -7,7 +7,11 @@
  * @flow
  */
 
-import {unstable_scheduleCallback as scheduleCallback} from 'scheduler';
+import * as Scheduler from 'scheduler';
+
+// Intentionally not named imports because Rollup would
+// use dynamic dispatch for CommonJS interop named imports.
+const {unstable_scheduleCallback: scheduleCallback} = Scheduler;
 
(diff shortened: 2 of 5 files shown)


acdlite:
Why did the size go up?

sebmarkbage:
We should track this for when we switch to ES modules though.

This is a semantic bug since it passes the `this` value. Maybe this fix can be done in Rollup if it hasn’t been already?

gaearon:
>Why did the size go up?

Looks like it only did for UMDs — because they *already* didn't suffer from this indirection. (They didn't do `require`.) 

So in UMD the change ended up adding a few lines of junk:

```diff
diff --git a/a.js b/a.js
index b714163..e4dda72 100644
--- a/a.js
+++ b/a.js
@@ -1009,7 +1009,7 @@
             c = c.focusOffset;
             try {
               b.nodeType, e.nodeType;
-            } catch (aj) {
+            } catch (lj) {
               b = null;
               break a;
             }
@@ -5891,7 +5891,19 @@
     SelectEventPlugin: Zi,
     BeforeInputEventPlugin: Ki
   });
-  var Xc = void 0,
+  var R = da.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Scheduler,
+    $i = R.unstable_cancelCallback,
+    aj = R.unstable_now,
+    bj = R.unstable_scheduleCallback,
+    cj = R.unstable_shouldYield,
+    dj = R.unstable_runWithPriority,
+    ej = R.unstable_getCurrentPriorityLevel,
+    fj = R.unstable_ImmediatePriority,
+    gj = R.unstable_UserBlockingPriority,
+    hj = R.unstable_NormalPriority,
+    ij = R.unstable_LowPriority,
+    jj = R.unstable_IdlePriority,
+    Xc = void 0,
     Nf = (function(a) {
       return "undefined" !== typeof MSApp && MSApp.execUnsafeLocalFunction
         ? function(b, c, d, e) {
@@ -5964,9 +5976,9 @@
       strokeOpacity: !0,
       strokeWidth: !0
     },
-    $i = ["Webkit", "ms", "Moz", "O"];
+    kj = ["Webkit", "ms", "Moz", "O"];
   Object.keys(yb).forEach(function(a) {
-    $i.forEach(function(b) {
+    kj.forEach(function(b) {
       b = b + a.charAt(0).toUpperCase() + a.substring(1);
       yb[b] = yb[a];
     });
@@ -5991,18 +6003,10 @@
         wbr: !0
       }
     ),
-    R = da.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Scheduler,
-    Ug = R.unstable_cancelCallback,
-    te = R.unstable_now,
-    Vg = R.unstable_scheduleCallback,
-    Oc = R.unstable_shouldYield,
-    Nc = R.unstable_runWithPriority,
-    yi = R.unstable_getCurrentPriorityLevel,
-    re = R.unstable_ImmediatePriority,
-    se = R.unstable_UserBlockingPriority,
-    Kg = R.unstable_NormalPriority,
-    zi = R.unstable_LowPriority,
-    Ai = R.unstable_IdlePriority,
+    te = aj,
+    Vg = bj,
+    Oc = cj,
+    Ug = $i,
     me = null,
     ne = null,
     Di = "function" === typeof setTimeout ? setTimeout : void 0,
@@ -6324,6 +6328,13 @@
   };
   var ji = "function" === typeof WeakSet ? WeakSet : Set,
     wi = "function" === typeof WeakMap ? WeakMap : Map,
+    yi = ej,
+    Nc = dj,
+    re = fj,
+    se = gj,
+    Kg = hj,
+    zi = ij,
+    Ai = jj,
     pe = Ma.ReactCurrentDispatcher,
     Jg = Ma.ReactCurrentOwner,
     ye = 1073741822,
```

The Node and WWW builds got smaller, however, as expected.

PR #14970, opened 2019-02-28
Replace test renderer's fake Scheduler implementation with mock build

acdlite:
The test renderer has its own mock implementation of the Scheduler interface, with the ability to partially render work in tests. Now that this functionality has been lifted into a proper mock Scheduler build, we can use that instead.

The changes to the test are mostly a straightforward find-and-replace. There were a few that relied on dynamic injection to override the `now` HostConfig method. I updated those tests to use Jest module mocking instead.


NE-SmallTown:
related: #14969

PR #14971, opened 2019-02-28
Swap expect(ReactNoop) for expect(Scheduler)

acdlite:
In the previous commits, I upgraded our custom Jest matchers for the noop and test renderers to use Scheduler under the hood.

Now that all these matchers are using Scheduler, we can drop support for passing ReactNoop and test roots and always pass Scheduler directly.


packages/react-reconciler/src/__tests__/ReactIncrementalSideEffects-test.internal.js
```diff
@@ -599,7 +601,7 @@ describe('ReactIncrementalSideEffects', () => {
     // render some higher priority work. The middle content will bailout so
     // it remains untouched which means that it should reuse it next time.
     ReactNoop.render(<Foo text="foo" step={1} />);
-    expect(ReactNoop).toFlushAndYield(['Foo', 'Content', 'Bar', 'Bar']);
+    expect(Scheduler).toFlushAndYield(['Foo', 'Content', 'Bar', 'Bar']);
```
NE-SmallTown: This seems very confused to me because `shouldComponentUpdate` returns `false` but the renderer still choose to render it(the `Content` component) which is very different to the DOM renderer, any details would you mind to share?

PR #14974, opened 2019-02-28
Fix act() tests by properly unmounting the container

philipp-spiess:
While playing around with the new `act()` API I discovered a test that
was unexpectedly failing. After some digging I noticed that the test is
passing when run in isolation and that the test suite can be fixe by
properly calling `ReactDOM.unmountComponentAtNode()` to clean up after
every test run.

You can reproduce the issue by simply commenting out the call to
`unmountComponentAtNode` in the `afterEach` callback. The newly added
test will fail but pass in isolation.

I thought it would be a good idea to fix this upstream so nobody runs
into this issue again.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
+++ b/packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
@@ -11,6 +11,7 @@ let React;
 let ReactDOM;
 let ReactTestUtils;
 let act;
+let container;
 
 jest.useRealTimers();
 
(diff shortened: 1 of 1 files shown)


packages/react-dom/src/__tests__/ReactTestUtils-test.js
```diff
@@ -678,6 +665,39 @@ describe('ReactTestUtils', () => {
     );
   });
 
+  it('flushes immediate re-renders with act but advances timers later', () => {
```
threepointone: act doesn't advance timers at all, so this title is a bit wrong.

philipp-spiess: Right, removed that part.

philipp-spiess:
@threepointone Rebased my PR but the issue I noticed earlier is no longer happening (the missing `ReactDOM.unmountComponentAtNode()` no longer causes differences in behavior - Which is great!). Let me know if you think the test case adds value and if we should get the changes in.

threepointone:
Thank you, it just slipped my mind! I do think your structure is better, I'll merge it in today.

threepointone:
lol @sizebot go home, you're drunk

gaearon:
@threepointone For the future — it's a good idea to always do cleanup in `afterEach` because you want to clean up even if a test failed and didn't reach the end.

PR #14982, opened 2019-02-28
[ESLint] Deduplicate suggested dependencies

gaearon:
The previous algorithm was pretty ad-hoc which resulted in edge cases like when it suggested `[props, props.onChange]` when `[props]` alone would suffice. This is a rewrite of it that instead maintains a data structure representing property chains. We mark the nodes that are *required* (i.e. used in the code), and then mark the nodes that are *satisfied* (i.e. specified in the dependency list).

That lets us precisely calculate the minimal necessary deps. (However, for effects we still allow extraneous deps when they’re already specified.) It should also open the door to making the rule smarter in the future (e.g. we should be able to extend this logic to suggest a parent object if we use >5 its properties or similar).

See inline comments and new tests.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -33,33 +33,33 @@ const tests = {
   valid: [
     {
       code: `
-      function MyComponent() {
-        const local = {};
-        useEffect(() => {
-          console.log(local);
-        });
-      }
-    `,
-    },
-    {
-      code: `
-      function MyComponent() {
-        useEffect(() => {
+        function MyComponent() {
           const local = {};
-          console.log(local);
-        }, []);
-      }
-    `,
+          useEffect(() => {
+            console.log(local);
+          });
+        }
+      `,
     },
     {
       code: `
-      function MyComponent() {
-        const local = {};
-        useEffect(() => {
-          console.log(local);
-        }, [local]);
-      }
-    `,
+        function MyComponent() {
+          useEffect(() => {
+            const local = {};
+            console.log(local);
+          }, []);
+        }
+      `,
+    },
+    {
+      code: `
+        function MyComponent() {
+          const local = {};
+          useEffect(() => {
+            console.log(local);
+          }, [local]);
+        }
+      `,
     },
     {
       // OK because `props` wasn't defined.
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -601,6 +599,24 @@ const tests = {
...
+          }, [props.foo]);
+          const fn3 = useCallback(() => {
+            console.log(props.foo.bar.baz);
+          }, [props.foo.bar]);
+          const fn4 = useCallback(() => {
+            console.log(props.foo.bar.baz);
+          }, [props.foo.bar.baz]);
+        }
+      `,
+    },
```
bvaughn: Nice test! 👍

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -1263,6 +1277,156 @@ const tests = {
...
+          const fn = useCallback(() => {
+            console.log(props.foo.bar.baz);
+          }, [props.foo.bar.baz, props.foo]);
+        }
+      `,
+      output: `
+        function MyComponent(props) {
+          const fn = useCallback(() => {
+            console.log(props.foo.bar.baz);
+          }, [props.foo]);
```
bvaughn: I guess this is the right approach, since we kind of have to assume immutability for much of this stuff to work. Seems like it's at least possible that `props.foo.bar.baz` was specified because `foo` gets mutated, but I guess we just don't support that case.

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -1910,20 +2074,18 @@ const tests = {
...
       `,
-      // TODO: [props.onChange] is superfluous. Fix to just [props.onChange].
       output: `
         function MyComponent(props) {
           useEffect(() => {
             if (props.onChange) {
               props.onChange();
             }
-          }, [props, props.onChange]);
+          }, [props]);
```
bvaughn: Nice 👍

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -306,77 +306,82 @@ export default {
...
+          // missingDependencies,
+        } = collectRecommendations({
+          dependencies,
+          declaredDependencies: [], // Pretend we don't know
+          optionalDependencies,
+          isEffect,
+        }));
+      }
+
+      // Alphabetize the suggestions, but only if deps were already alphabetized.
```
bvaughn: ❤️

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
gaearon: Yeah. (This was already the case before this PR btw.)

bvaughn: Yeah, this particular before and after example just made me think about it more closely :)

PR #14984, opened 2019-03-01
Import Scheduler directly, not via host config

acdlite:
We currently schedule asynchronous tasks via the host config. (The host config is a static/build-time dependency injection system that varies across different renderers — DOM, native, test, and so on.) Instead of calling platform APIs like `requestIdleCallback` directly, each renderer implements a method called `scheduleDeferredCallback`.

We've since discovered that when scheduling tasks, it's crucial that React work is placed in the same queue as other, non-React work on the main thread. Otherwise, you easily end up in a starvation scenario where rendering is constantly interrupted by less important tasks. You need a centralized coordinator that is used both by React and by other frameworks and application code. This coordinator must also have a consistent API across all the different host environments, for convention's sake and so product code is portable — e.g. so the same component can work in both React Native and React Native Web.

This turned into the Scheduler package. We will have different builds of Scheduler for each of our target platforms. With this approach, we treat Scheduler like a built-in platform primitive that exists wherever React is supported.

Now that we have this consistent interface, the indirection of the host config no longer makes sense for the purpose of scheduling tasks. In fact, we explicitly do not want renderers to schedule tasks via any system except the Scheduler package.

So, this PR removes `scheduleDeferredCallback` and its associated methods from the host config in favor of directly importing Scheduler.


sebmarkbage:
The key here is that we expect isomorphic React code to directly schedule work with the scheduler package and not go through the React package to do so.

Even though you technically could build a renderer with a different host config injection, it would no longer accept idiomatic React code since idiomatic React code would depend on “scheduler”.

PR #14991, opened 2019-03-01
Add Paris Climate Agreement badge for FB to README

tomj:
Facebook is a signatory to [_'We Are Still In'_](https://www.wearestillin.com/signatories) which pledges corporate support for the Paris Climate Agreement.  Would it be against the principles of the React project to acknowledge this on the `README` with a badge?

[Climate rating](https://www.wearestillin.com/signatories)

Climate change is an immense and time sensitive issue which affects us all.  Acknowledging Paris on the repo would provide reassurance to contributors and users that as the host/guardian of React, FB is committed to a positive future for the planet.  The badge links to [wearestillin.com](https://wearestillin.com/) - the declaration which Facebook signed acknowledging their commitment to the Paris Agreement.

I understand this isn't your typical PR and is my first contribution to React.  But every bit of awareness about climate change makes an impact and this is a pretty positive diff from just a few ASCII characters 🙌🌍


cpojer:
Thank you for this PR. I think it's really cool that such a badge exists, and I'm glad that Facebook is part of it but I don't think we need to add this to each of the projects on our GitHub org.

PR #14992, opened 2019-03-01
Revert #14756 changes to ReactFiberScheduler

acdlite:
#14756 PR introduced some bugs in concurrent mode during internal testing. I believe it's specifically related to the changes to ReactFiberScheduler. Until we figure out a proper solution, I'm going to try reverting it.

Not totally certain this is sufficient to unbreak the bugs we found, but I'm using this branch to determine that.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -12,6 +12,11 @@ import type {Batch, FiberRoot} from './ReactFiberRoot';
 import type {ExpirationTime} from './ReactFiberExpirationTime';
 import type {Interaction} from 'scheduler/src/Tracing';
 
+// Intentionally not named imports because Rollup would use dynamic dispatch for
+// CommonJS interop named imports.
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -172,19 +171,6 @@ export type Thenable = {
   then(resolve: () => mixed, reject?: () => mixed): mixed,
 };
 
-// Intentionally not named imports because Rollup would
```
gaearon: Mind keeping these?

acdlite: They're not being used in this module anymore. Do you mean comment these lines out?

acdlite:
Ok after testing in a sandbox, I believe this fixes/unbreaks at least one of the bugs we found internally.

packages/react-reconciler/src/ReactFiberScheduler.js
acdlite: I'll comment them out so we don't forget next time we import Scheduler in this module

PR #14993, opened 2019-03-01
[ESLint] Suggest to destructure props when they are only used as members

gaearon:
This addresses some confusion we've seen around `props.foo()` asking for `props` to be included. It's also best practice anyway. But only show the warning if `props` isn't used as an object. (And only if it's actually missing — so that it doesn't complain for valid code.)

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -2391,6 +2391,29 @@ const tests = {
     {
       code: `
         function MyComponent(props) {
+          useEffect(() => {
+            props.onChange();
+            externalCall(props);
+          }, []);
+        }
+      `,
+      output: `
+        function MyComponent(props) {
+          useEffect(() => {
+            props.onChange();
+            externalCall(props);
+          }, [props]);
+        }
+      `,
+      // Don't suggest to destructure props here since you can't.
+      errors: [
+        "React Hook useEffect has a missing dependency: 'props'. " +
+          'Either include it or remove the dependency array.',
+      ],
+    },
+    {
+      code: `
+        function MyComponent(props) {
           let value;
           let value2;
           let value3;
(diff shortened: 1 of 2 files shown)


gaearon:
I don't think it's worth warning if it's called something other than `props` (personally have never seen that).

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -2224,7 +2224,189 @@ const tests = {
         }
       `,
       errors: [
-        // TODO: make this message clearer since it's not obvious why.
+        "React Hook useEffect has a missing dependency: 'props'. " +
```
Andarist: This error message kinda surprises me when taking into account our recent discussion ( https://github.com/facebook/react/pull/14876 ) around props referential equality etc. 

If `props` have "unstable" reference then IMHO you shouldnt ever suggest using it as hook's dependency 🤔 The latter advice sounds way more appropriate (destructuring)

gaearon: `props` is OK as an escape hatch when you just want the rule to shut up. It's still stable if you only set state, for example. Better than nothing.

Andarist: Shouldnt the advice get reordered then? First propose destructuring and only latter propose using `props` directly.

You are calling it an escape hatch to shut up the rule - so it sounds to me that this is actually not something which you'd like to recommend.

gaearon: You're welcome to send a PR. :-) There are more worrying things I'm currently focused on than this particular message.

Andarist: Sure, might prepare one - was just validating my thinking so far :)

PR #14996, opened 2019-03-02
[ESLint] Treat functions that don't capture anything as static

gaearon:
The exact heuristics will take me some time to figure out. This is just the first minimal draft.

It lifts the restriction on specifying functions as deps where it's safe. In particular, a function like

```js
  function tick() {
    setCount(c => c + 1);
  }
```

would be considered safe because it doesn't capture any render-phase values except knowingly static ones (we only do this if `setCount` comes from `useState` Hook).

So we could omit it:

```js
  useEffect(() => {
    let id = setInterval(() => {
      tick();
    }, 1000);
    return () => clearInterval(id);
  }, []); // No warning
```

But if we change `tick` to be

```js
  function tick() {
    setCount(count + 1);
  }
```

it would warn, asking you to add `tick` to dependencies.

We only do this check one level deep. So it helps for helper functions that only set state or dispatch, but you might end up in a situation where you can't add a prop to one of them without triggering warnings for all effects using it. Ideally we might want to push you to `useReducer` a bit sooner but I haven't figured out what a good flow is.

**This PR makes the rule strictly more relaxed. It doesn't add any new behaviors, it just fires the warnings less often in cases where it's safe.**

In follow-ups, we'll need to tweak the behavior and warnings to be more useful for functions. Such as we might want to detect the above special case and *suggest the updater form* if it would fix the warning. Similarly, we could detect always-new deps (like `tick`) and in that case suggest wrapping in `useCallback`.

### DIFF
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -34,6 +34,22 @@ export default {
         : undefined;
     const options = {additionalHooks};
 
+    // Should be shared between visitors.
+    let staticKnownValueCache = new WeakMap();
+    let functionWithoutCapturedValueCache = new WeakMap();
+    function memoizeWithWeakMap(fn, map) {
+      return function(arg) {
+        if (map.has(arg)) {
+          // to verify cache hits:
+          // console.log(arg.name)
+          return map.get(arg);
+        }
+        const result = fn(arg);
+        map.set(arg, result);
+        return result;
+      };
+    }
+
     return {
       FunctionExpression: visitFunctionExpression,
       ArrowFunctionExpression: visitFunctionExpression,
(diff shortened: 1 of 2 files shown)


gaearon:
Perhaps, a better alternative could be to **make the autofix wrap safe functions into `useCallback` for you**. That teaches you about the pattern early and works transitively as you add more functions.

The upside of that would be that if you add a prop to some deep function in the call stack, it propagates to immediate callers. So you don't have an avalanche of effects to fix. But the downside is you don't see *which* effects depend on that prop transitively. However, that may be good. Because effects might get reported with coarser deps than necessary otherwise. Having `useCallback` at every level helps you confront which of those actually need those props.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -105,6 +121,153 @@ export default {
...
+
+      // Some are known to be static based on Hook calls.
+      // const [state, setState] = useState() / React.useState()
+      //               ^^^ true for this reference
+      // const [state, dispatch] = useReducer() / React.useReducer()
+      //               ^^^ true for this reference
+      // const ref = useRef()
+      //       ^^^ true for this reference
+      // False for everything else.
+      function isStaticKnownHookValue(resolved) {
```
gaearon: This function just moved into the closure. It's the same implementation with a slightly different argument.

PR #15003, opened 2019-03-03
refactor (es6): eslint-plugin-react-hooks

gkal19:
- Introduce `destructuring binding`
- `includes` for readability


gkal19:
```
$ ./scripts/circleci/track_stats.sh
facts-tracker: Branch is not master, exiting...
```

gaearon:
Thanks but I don't think this helps readability. We generally avoid purely stylistic PRs too.

Issue #14201, opened 2018-11-12
Automated incremental React releases


bvaughn:
This ticket covers the changes/additions we would need to make to our existing [release scripts](https://github.com/facebook/react/tree/master/scripts/release) to support automatically publishing canaries from master.

### Changes to CI scripts
- [x] #14244: Update CI scripts and `scripts/release/build.js` to share code for building and generating artifacts. (There's a lot of redundancy now.)
- [x] #14244: Auto-generate a checksum file and add it to each NPM package (as well as the `files` array in the `package.json`). This will enable us to later verify that that a specific package version corresponds with a specific Git revision to help ensure that we don't accidentally publish incompatible sibling packages.
- [x] #14244: Auto-generate a unique version number based on the Git revision that can be used for a canary ([like this](https://github.com/facebook/react/blob/1f6d3438034162a6ae3b4d96027aaeaa065d41eb/scripts/continuous-release/index.js#L486-L502)).
- [x] #14205: Circle CI already builds all of our packages _and_ runs tests again both source and the built modules. This is _most_ of what our [build script](https://github.com/facebook/react/blob/master/scripts/release/build.js) currently does. We should go ahead and save the resulting builds to Circle CI so that we can just download and publish them as-is.

### Changes to local scripts
- [x] #14225: Create a script to **prepare a canary release** from a CI artifact.
  - [x] Require inputs: Circle CI build identifier (e.g. [12639](https://circleci.com/gh/facebook/react/12639#artifacts/containers/0)).
  - [x] It should download and extract the artifact to `build/node_modules`
- [x] #14244: Create a script tp **create a canary release** from the current Git branch and revision.
  - [x] Resulting output should be the same as the canary release script.
- [x] #14255: Create a script tp **prepare a stable release** from a published canary.
  - [x] #14255: Require inputs: NPM version number of a published canary release.
  - [x] #14255: It should scan the local [`packages` directory](https://github.com/facebook/react/tree/master/packages) to determine the list of _non-private_ packages, and then check out the canary version of each package to `build/node_modules`.
  - [x] #14255: It should prompt for a version number for each package. (By default, it could fill in the latest major+minor by querying NPM.) This should be done up front because of inter-package dependencies (e.g. `react` depends on `scheduler`).
  - [x] #14255: It should replace the canary version number (in both `package.json` and in the source for renderers) with the specified version. It should also replace version numbers for `dependencies`/`peerDependencies`.
[...]

PR #15005, opened 2019-03-03
fix(auto-version-update): update root package version while publishing

### LINKED ISSUE #14985: package have wrong version
package.json have wrong version.
The current release is 16.8.3, but the package.json version is 16.6.1.

nutboltu:
Fixes #14985 

Root Cause: #14201 

In `scripts/release/publish-commands` scripts we only updated the packages versions and the version in `ReactVersion.js`. But not the root package version.

**Possible Solutions**

- [ ] Updating the root package version number as per react version.
- [x] Remove version from root package json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,5 @@
 {
   "private": true,
-  "version": "16.6.1",
   "workspaces": [
     "packages/*"
   ],
--- a/scripts/release/publish-commands/update-stable-version-numbers.js
+++ b/scripts/release/publish-commands/update-stable-version-numbers.js
@@ -36,14 +36,21 @@ const run = async ({cwd, packages, tags}) => {
 
   // Update the shared React version source file.
   const sourceReactVersionPath = join(cwd, 'packages/shared/ReactVersion.js');
-  const {version} = await readJson(
+  const {version: reactVersion} = await readJson(
     join(nodeModulesPath, 'react', 'package.json')
   );
   const sourceReactVersion = readFileSync(
     sourceReactVersionPath,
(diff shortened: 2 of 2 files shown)


bvaughn:
I think we should just remove the `version` field entirely, since it's only in-sync with a few of the packages to begin with. Want to update the PR to do that?

nutboltu:
@bvaughn I am going to remove the `version` from root package json

bvaughn:
Thanks~

NE-SmallTown:
I don't think we should remove this, because sometimes I want to check the react version(not on npm), so if I can't find the version field, it would be confusing

bvaughn:
> sometimes I want to check the react version on github(not on npm)

To be honest, that's not really a use case I think we are too worried about supporting, but you can check it here instead:
https://github.com/facebook/react/blob/master/packages/react/package.json#L7

gaearon:
You can find the version here: https://reactjs.org

Top right corner

nutboltu:
It's easy to find the version in [CHANGELOG.md](https://github.com/facebook/react/blob/master/CHANGELOG.md). Also you can see it in the README
react_version

NE-SmallTown:
Yea, I could find it at many places, changelog, release, package/react/package.json, reactjs.org, even in twitter. But from my perspective, it's a very common case that a open source js project describe it's version in the root package.json, right? I know there are some monorepos like react/babel remove the version, but add it is not a complicate cost, but will give someone a better experience.

But maybe I worry too much :)

PR #15013, opened 2019-03-05
Run Placeholder tests in persistent mode, too

acdlite:
I screwed up a few months ago when I converted the Suspense placeholder tests to use test renderer instead of noop renderer — I forgot that we were using noop renderer because it supports both mutation and persistent mode. This PR corrects that mistake by converting them back to the noop renderer. This exposed bugs in our persistent implementation. (We don't use Suspense in persistent mode anywhere yet, but it's what Fabric will use.)

This indicates we should stop using test renderer for our internal React tests, and that we should probably be running all our tests in both modes by default. (I also found a bug related to profiler in persistent mode.) But I'll leave that for a follow up.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -16,6 +16,7 @@ import type {
   Props,
   Container,
   ChildSet,
+  HostContext,
 } from './ReactFiberHostConfig';
 
 import {
@@ -50,6 +51,7 @@ import {
 import invariant from 'shared/invariant';
 
 import {
+  getRootHostContext,
   createInstance,
   createTextInstance,
   createHiddenTextInstance,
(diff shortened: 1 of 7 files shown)


packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -577,12 +630,19 @@ function completeWork(
...
+        workInProgress,
+        rootContainerInstance,
+        rootHostContext,
+      );
       break;
     }
     case HostComponent: {
-      popHostContext(workInProgress);
       const rootContainerInstance = getRootHostContainer();
+      const childHostContext = getHostContext();
```
sebmarkbage: This seems fragile. We typically don't have to assume that a context needs to be passed in. How should I know when I can use context and not?

What prevents you from popping it later?

acdlite: The test I wrote :D

sebmarkbage: I'm not interested in the bugs but why we can't restructure the code to also fix it.

acdlite: So the only alternative I could think of was to have expose a `getParentContext` function from the stack module. This seemed less bad? Idk

Issue #14983, opened 2019-02-28
[reconciler] re-instanciating object in commitUpdate


drcmda:
**Do you want to request a *feature* or report a *bug*?**

feature, this is for https://github.com/drcmda/react-three-fiber

**What is the current behavior?**

commitUpdate is supposed to simply update the instance, it seems that it cannot "exchange" the instance. 

**What is the expected behavior?**

I need to re-instanciate the instance, just writing props to it is not enough in my case. Basically i am trying this:

```js
  commitUpdate(instance, updatePayload, type, oldProps, newProps, fiber) {
    if (instance.isObject3D) {
      applyProps(instance, newProps, oldProps)
    } else {
      // This is a data object, let's extract critical information about it
      const parent = instance.__parent
      const { args: argsNew = [], ...restNew } = newProps
      const { args: argsOld = [], ...restOld } = oldProps
      // If it has new props or arguments, then it needs to be re-instanciated
      if (argsNew.some((value, index) => value !== argsOld[index])) {
        // First it gets removed from its parent
        removeChild(parent, instance)
        // Next we create a new instance and append it again
        const newInstance = createInstance(instance.type, newProps)
        appendChild(parent, newInstance)
        // Switch fiber node (???)
        fiber.stateNode = newInstance // <---- this doesn't work
```

This is a THREE data object which i am trying to map. Sometimes these objects cannot be mutated, THREE expects them to be re-created. One example of this is: https://threejs.org/docs/index.html#api/en/geometries/PlaneGeometry 

> Any modification after instantiation does not change the geometry.

I am creating the object in JSX like so:

```jsx
<planeGeometry name="geometry" args={[width, height]} />
```

When the args change the instance has to be recreated, no other way around unfortunately.

It would be amazing if commitUpdate would be allowed to return an instance. And if it does the reconciler writes it into the fiber node. I am trying fiber.stateNode here, but it doesn't help. Next time commitUpdate is called it passes on the old instance again, which doesn't have a parent any longer.

I don't really understand how these nodes work but the stacktrace says  

```jsx
commitWork(_current2, nextEffect)
```

where nextEffect ends up being "fiber" in the commitUpdate. "_current2" seems to be the one that still clings to the old stateNode.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

React 16.8.3

drcmda:
Oh, i just tried something that did actually work:

```jsx
        // Switch fiber node
[...]

PR #15021, opened 2019-03-05
Allow commitUpdate to re-instantiate a stateNode

drcmda:
This PR extends the reconcilers commitUpdate function with a function that can optionally signal that the host intents to re-instantiate the instance. See: https://github.com/facebook/react/issues/14983

This is how it could look from the end-users perspective, exchangeInstance is optional:

```jsx
commitUpdate(instance, updatePayload, type, oldProps, newProps, fiber, exchangeInstance) {
  if (!shallowEqual(newProps, oldProps)) {
    const newInstance = new Catalogue[type](newProps)
    exchangeInstance(newInstance)
  }
}
```

### The problem

Some targets do not allow mutation of instances after creating, or at least mutation doesn't have any effect. One such case is THREEJS where [some objects](https://threejs.org/docs/index.html#api/en/geometries/PlaneGeometry) have that restriction and they are expected to be re-created:

> Any modification after instantiation does not change the geometry.

This prevents my [reconciler](https://github.com/drcmda/react-three-fiber) from updating props.

Dan has suggested wrapping these objects, which would allow us to exchange instances. I did this, but now users get served different refs, which is confusing.

```jsx
<mesh ref={mesh}> // mesh.current === THREE.Mesh
  <geometry ref={geo} /> // geo.current.object === THEE.Geometry
```

Another suggestion was to export these objects as forwardref function, so users will have to do this:

```jsx
import { Geometry } from 'react-three-fiber'

<mesh ref={mesh}> // mesh.current === THREE.Mesh
  <Geometry ref={geo} /> // geo.current=== THEE.Geometry
```

But that's also confusing imo, maybe even more than wrapping instances. If i can i would like to abstract the problem away, it's three's problem, not the end-users. 

The third option would be to use Proxies, but we can't drop IE.

This is just meant to have a more technical discussion, seeing if it would maybe be possible for the reconciler to support such platforms. The PR works, but this is just a stab into the dark.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1174,7 +1174,7 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
             newInstance => {
               // Some hosts cannot mutate instances, in that case the reconciler can
               // call this method in order to signal that it has created a new instance.
-              [finishedWork, current$$1].forEach(fiber => {
+              [finishedWork, current].forEach(fiber => {
                 if (fiber !== null) {
                   fiber.stateNode = newInstance;
                   if (fiber.ref) {
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1171,6 +1171,22 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
             oldProps,
             newProps,
             finishedWork,
+            newInstance => {
+              // Some hosts cannot mutate instances, in that case the reconciler can
+              // call this method in order to signal that it has created a new instance.
+              [finishedWork, current$$1].forEach(fiber => {
+                if (fiber !== null) {
+                  fiber.stateNode = newInstance;
+                  if (fiber.ref) {
+                    if (typeof fiber.ref === 'function') {
+                      fiber.ref(newInstance);
+                    } else {
+                      fiber.ref.current = newInstance;
+                    }
+                  }
+                }
+              });
+            },
           );
         }
       }
--- a/scripts/rollup/results.json
+++ b/scripts/rollup/results.json
@@ -4,29 +4,29 @@
       "filename": "react.development.js",
       "bundleType": "UMD_DEV",
       "packageName": "react",
-      "size": 102032,
-      "gzip": 26457
+      "size": 102060,
+      "gzip": 26495
     },
     {
       "filename": "react.production.min.js",
       "bundleType": "UMD_PROD",
       "packageName": "react",
-      "size": 12564,
-      "gzip": 4826
+      "size": 12609,
+      "gzip": 4834
     },
     {
       "filename": "react.development.js",
       "bundleType": "NODE_DEV",
       "packageName": "react",
-      "size": 63704,
-      "gzip": 17181
+      "size": 63707,
+      "gzip": 17178
     },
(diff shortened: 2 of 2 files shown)


gaearon:
What happens to the parent/child relationships both ways?

drcmda:
these objects, at least in my case, do not have parents or any relations. a vector3 is just a class. i let react handle the relations through the fibernodes.

gaearon:
Any host nodes can have parents and children. If you want to implement it on a generic level then you need to have an idea of how that affects parent/child relationships in general case and doesn't leave them messed up.

gaearon:
In particular, what happens if you have one of these things in a tree, then it gets "replaced", then we add/move/remove its children or move/remove itself in the parent.

drcmda:
a vector doesnt have children. as far as i understand the fiber node does. replacing the statenode doesn't affect the children because at no point do actual children written into it. the fiber node stays put, even with an exchanged reference to the host instance it wraps around. here's an example:

```jsx
function appendChild(parentInstance, child) {
  if (child) {
    if (child.isObject3D) parentInstance.add(child)
  }
}

function removeChild(parentInstance, child) {
  if (child) {
    if (child.isObject3D) {
      parentInstance.remove(child)
      if (child.dispose) child.dispose()
    }
  }
}
```

If it's not a scene object it doesn't get added anywhere, i let react handle children through fiber.

gaearon:
>If these objects could have children then it would be up to commitupdate to take care of it, that also wouldn't be a problem, it would track relationships and call its own addChild/removeChild methods like so:

I don't have confidence that *React itself* won't get confused later.

Try using this method in ReactDOM, e.g., to remount divs when their classname changes. Then try to add and remove their children while also changing their classname. See if you can find edge cases.

drcmda:
> I don't have confidence that React itself won't get confused later.

Technically there's no reason for react to be confused. It doesn't know the host, that's why it passes on the responsibility to add/remove to the config holder and any contact that it has to make with it host goes through that layer of responsibility. Wouldn't be any different with the dom.

If we could establish that stateNode can safely be mutated as long as the reconciler guides it, that would allow us to steer platforms that are not entirely prop/update driven.

drcmda:
After wrabgling too much with side-effects of wrapping objects and the downsides of the other possible workarounds i have decided to go forward with hacking into react internals:

https://github.com/drcmda/react-three-fiber/blob/521ea40010ee5827e5f73ce859e2f4d39ea4a8d1/src/reconciler.js#L227-L238

I think that's the only solution that's clean on the outside and doesn't put a burden on the users shoulders.

It would be super appreciated if maybe @sebmarkbage could have a quick look. Is fiber and fiber.alternate a given, and what is "alternate" good for? And does react hold other references to the instance? 

Other than that, i think this PR can be closed?

michalochman:
@drcmda is there a reason you don't want to use the `key` prop for that? If you just want to create new instance and mount it instead of updating old instance then set the `key` to some kind of hash of your `newProps`. There are edge cases where it would not work, but most of the time it will.

drcmda:
@michalochman i am not sure, but can i change the key prop inside the reconciler before the commit phase? if you mean a userland key swap, i wouldn't like putting technical debt on the users shoulder. There's no reason the user should be concerned about what happens underneath, they assign new arguments and everything just works:

```jsx
<planeBufferGeometry args={[width, height]} />
```

```jsx
<planeBufferGeometry key={`${width}-${height}`} args={[width, height]} />
```

I guess the latter could be my escape hatch once react changes their internal api.

Issue #15012, opened 2019-03-04
`data-reactroot=""` lost in server render if element inside `<Context.Provider>`


overlookmotel:
**Do you want to request a *feature* or report a *bug*?**

Bug.

**What is the current behavior?**

Calling `.renderToString()` on `<React.Fragment><div>Hello!</div></React.Fragment>` produces `<div data-reactroot="">Hello!</div>`.

However, many other combinations of DOM root element within "invisible" React elements are rendered without `data-reactroot=""`.

e.g. all of the following produce `<div>Hello!</div>` (without `data-reactroot=""`):

```
<React.Fragment><React.Fragment><div>Hello!</div></React.Fragment></React.Fragment>

<Context.Provider><div>Hello!</div></Context.Provider>

<Context.Consumer>{() => <div>Hello!</div>}</Context.Consumer>

<React.StrictMode><div>Hello!</div></React.StrictMode>
```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

See above.

**What is the expected behavior?**

I assume that any of the above examples should produce markup including `data-reactroot=""` on the `div` element.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

Issue present in all versions >= 16.7.0, possibly earlier versions too.

This is a minor issue - I guess would only affect where the markup is rehydrated on client side using `.render()` rather than `.hydrate()` - usage which is discouraged. However, using `.render()` to rehydrate is still officially supported.

I can implement a fix and make a PR. But can anyone confirm that the expected behavior is what I think it is, and that this is indeed a bug?

gaearon:
This does sound like a bug. A fix would be welcome.

overlookmotel:
OK great. Thanks for swift response. I'll submit a PR soon as I can - should be in next couple of weeks.

overlookmotel:
I've submitted a PR #15023.

overlookmotel:
@gaearon Anyone had a chance to look at my fix for this in PR #15023?

ianks:
Now that react-apollo is using the Context API, our builds are broken because of this.

gaearon:
[...]

PR #15023, opened 2019-03-05
Fix SSR missing data-reactroot attribute inside invisible React elements

overlookmotel:
PR for #15012.

```js
const React = require('react'),
  {renderToString} = require('react-dom/server');

const Context = React.createContext();

const html = renderToString(
  <Context.Provider>
    <div>Hello!</div>
  </Context.Provider>
);
```

Before this PR, outputs:

```
'<div>Hello!</div>'
```

with this PR:

```
'<div data-reactroot="">Hello!</div>'
```

Implementation is through addition of a `.domRootFrame` property on `ReactDOMServerRenderer` instance.

* `domRootFrame` is `null` if no DOM element in stack above currently rendering element.
* When a entering a root DOM element, `domRootFrame` is set to the stack frame for that DOM element.
* When exiting root DOM element, `domRootFrame` is set back to `null` (the root element is identified by checking if the frame being popped from the stack is equal to `domRootFrame`)

I am not clear what desired behavior is when there are multiple DOM roots e.g.:

```
<React.Fragment>
  <div>One</div>
  <div>Two</div>
</React.Fragment>
```

I guess both `<div>`s should get `data-reactroot=""` markup, but not sure what the result on client side would be!

Any changes you want made, please shout.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationFragment-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationFragment-test.js
@@ -102,5 +102,76 @@ describe('ReactDOMServerIntegration', () => {
     itRenders('an empty fragment', async render => {
       expect(await render(<React.Fragment />)).toBe(null);
     });
+
+    // Regression tests for https://github.com/facebook/react/issues/15012
+    describe('creates markup with data-reactroot="" attribute', () => {
+      it('within fragment', () => {
+        const App = () => (
+          <React.Fragment>
+            <div>
+              <div>Hello</div>
+            </div>
+          </React.Fragment>
+        );
+
+        expect(ReactDOMServer.renderToString(<App />)).toBe(
+          '<div data-reactroot=""><div>Hello</div></div>',
+        );
+      });
+
+      it('within nested fragments', () => {
+        const App = () => (
+          <React.Fragment>
+            <React.Fragment>
+              <div>
+                <div>Hello</div>
+              </div>
+            </React.Fragment>
+          </React.Fragment>
+        );
+
+        expect(ReactDOMServer.renderToString(<App />)).toBe(
+          '<div data-reactroot=""><div>Hello</div></div>',
+        );
+      });
+
+      it('with multiple DOM roots', () => {
+        const App = () => (
+          <React.Fragment>
+            <div>
+              <div>One</div>
+            </div>
+            <div>
+              <div>Two</div>
+            </div>
+          </React.Fragment>
+        );
+
+        expect(ReactDOMServer.renderToString(<App />)).toBe(
+          '<div data-reactroot=""><div>One</div></div><div data-reactroot=""><div>Two</div></div>',
+        );
+      });
+
+      it('with multiple DOM roots in multiple fragments', () => {
+        const App = () => (
+          <React.Fragment>
+            <React.Fragment>
+              <div>
+                <div>One</div>
+              </div>
+            </React.Fragment>
+            <React.Fragment>
+              <div>
+                <div>Two</div>
+              </div>
+            </React.Fragment>
+          </React.Fragment>
+        );
+
+        expect(ReactDOMServer.renderToString(<App />)).toBe(
+          '<div data-reactroot=""><div>One</div></div><div data-reactroot=""><div>Two</div></div>',
+        );
+      });
+    });
   });
 });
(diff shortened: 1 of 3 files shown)


overlookmotel:
I altered the implementation to use an integer property `.domRootIndex` on `ReactDOMServerRenderer` instance.

Behavior remains the same.

If `domRootIndex === 0`, no DOM nodes exist above in the stack, so next DOM element encountered is the root and should have `data-reactroot` attribute. If `domRootIndex !== 0`, we are already inside a DOM node, so any other DOM elements are not the root.

* `domRootIndex` is `0` initially.
* When entering first DOM element, `domRootIndex` is set to length of stack after DOM element's frame is pushed onto stack.
* When exiting DOM root element, `domRootIndex` is set back to `0` (it's determined whether the frame being popped represents the DOM root element by testing if length of stack is equal to `domRootIndex`).

ianks:
I'm not familiar with the React codebase so I can't comment on implementation, but overall the specs cover the cases where we are experiencing problems. I would also say that solving this issue is hugely important because in many cases, a visible react root will not exist since HOC will open need to have hydration code before rendering a component. Would love to see this get merged :smile:

ianks:
Any progress on this?

overlookmotel:
Anyone got a moment to look at this? @gaearon confirmed this was a bug, and this PR fixes it, so would be great if it'd be possible to look at merging it.

Happy to make any changes you need made to get it fit for merging.

kiddkai:
Adding this attribute will make `render` function work exactly the same as `hydrate` function. 

By the implementation in the client side:

https://github.com/facebook/react/blob/1d48b4a68485ce870711e6baa98e5c9f5f213fdf/packages/react-dom/src/fire/ReactFire.js#L505

Even though `render` is currently not recommended, it removes everything from the container and render everything again on the client-side, which make sense for some use cases that SSR has some mismatch content with client render.

kiddkai:
Adding this attribute will make `render` function work exactly the same as `hydrate` function. 

By the implementation in the client side:

https://github.com/facebook/react/blob/1d48b4a68485ce870711e6baa98e5c9f5f213fdf/packages/react-dom/src/fire/ReactFire.js#L505

Even though `render` is currently not recommended, it removes everything from the container and render everything again on the client-side, which make sense for some use cases that SSR has some mismatch content with client render.

overlookmotel:
Hi @kiddkai. Just to clarify: are you saying this would be a bad thing?

kiddkai:
Hi @overlookmotel, IMO it is a bad thing.

For some huge applications that haven't migrate to hydrate compatible code, it could be devastating. For some framework like `next.js`, they manually put `data-reactroot` into the top level element. Which will force `render` turn into `hydrate` mode. If you want the hydrate behaviour. It's better either use `render` with manual `data-reactroot` or just `hydrate`.

overlookmotel:
@kiddkai If you don't want `data-reactroot` in your server-rendered HTML, can't you just use `.renderToStaticMarkup()` rather than `.renderToString()`?

My point with #15012 is that the current behavior is inconsistent.

The following produce `<div data-reactroot="">Hello!</div>`:

* `renderToString(<div>Hello!</div>)`
* `renderToString(<React.Fragment><div>Hello!</div></React.Fragment>)`

and the following produce just `<div>Hello!</div>`:

* `renderToString(<React.Fragment><React.Fragment><div>Hello!</div></React.Fragment></React.Fragment>)`
* `renderToString(<Context.Provider><div>Hello!</div></Context.Provider>)`

This does not look like an intentional decision to me, but a bug.

Actually, we should probably discuss this on #15012 rather than here, as that issue describes the bug (or what I say is a bug). This PR is just a fix for the issue.

overlookmotel:
Please don't close this, Mr Robot.

@gaearon did confirm this as a bug in #15012, so I'm hoping it may be merged at some point.

gaearon:
Thanks for PR.

>I am not clear what desired behavior is when there are multiple DOM roots 

I think that's the crux of this issue. Emitting two nodes tagged as roots isn't what we want, I think.

The reason this whole attribute even matters is for compatibility with hydrating via `ReactDOM.render` call. However we've already deprecated that behavior in favor of `ReactDOM.hydrate`. So I'm not sure this is worth fixing. What do you think?

overlookmotel:
Please keep this fresh, Mr Bot.

ianks:
whats the hold up on merging this?

overlookmotel:
Closing as @gaearon has indicated on #15012 that they do not want to fix.

gaearon:
Sorry for the churn. I appreciate you doing the work. We’ve gotten swamped with other stuff and now that I revisit it’s clear we want to go a different way anyway.

PR #15029, opened 2019-03-06
Run persistent mode tests in CI

acdlite:
## Based on #15013

Adds a command to run React Noop tests in persistent mode.

```
yarn test-persistent
```

This command will run in CI.

Includes a fix for a bug I found when running the tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "postinstall": "node node_modules/fbjs-scripts/node/check-dev-engines.js package.json && node ./scripts/flow/createFlowConfigs.js",
     "debug-test": "cross-env NODE_ENV=development node --inspect-brk node_modules/.bin/jest --config ./scripts/jest/config.source.js --runInBand",
     "test": "cross-env NODE_ENV=development jest --config ./scripts/jest/config.source.js",
+    "test-persistent": "cross-env NODE_ENV=development jest --config ./scripts/jest/config.source-persistent.js",
     "test-fire": "cross-env NODE_ENV=development jest --config ./scripts/jest/config.source-fire.js",
     "test-prod": "cross-env NODE_ENV=production jest --config ./scripts/jest/config.source.js",
     "test-fire-prod": "cross-env NODE_ENV=production jest --config ./scripts/jest/config.source-fire.js",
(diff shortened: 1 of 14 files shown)


scripts/jest/setupPersistent.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
```
sebmarkbage: Do we need both these identical files? setupPersistent and setupTests.persistent. What's the difference?

package.json
```diff
@@ -80,11 +81,11 @@
     "rollup-plugin-replace": "^2.0.0",
     "rollup-plugin-strip-banner": "^0.2.0",
     "semver": "^5.5.0",
+    "shelljs": "^0.8.3",
```
sebmarkbage: #carmackno We need fewer deps. This is unnecessarily complicated.

acdlite: Sure, I'll just delete that last commit then

scripts/jest/setupPersistent.js
acdlite: Idk I copy-pasted the other test entry points

acdlite: Oh no wait, that's a rename mistake. I'll delete whichever one is superfluous.

package.json
acdlite: For context I would have just used `exec` or whatever but something something Windows contributors whatever I don't care enough it's CI anyway

sebmarkbage: This was a bit of a last minute Friday drive-by review. The dependency is just one issue. The other issue is its reliance on file system information in general. We should leave that to jest since it's something that does a lot of complex work around invalidation, file watching, module resolution etc. It also fairly frequently changes.

So without this diff, it just runs more than it needs to on CI wasting cycles but should be equivalent otherwise?

acdlite:
I don't think this will affect CI times because we have four parallel instances and I put this task in the one that previously ran the shortest. I was more interested in the time it takes to run `test-persistent` locally.

package.json
acdlite: Yeah

acdlite:
Screen Shot 2019-03-11 at 10 54 59 AM

Persistent tests run in container 0, which is still not close to being the slowest one.

PR #12919, opened 2018-05-28
Always use event capturing. Remove media events.

nhunzaker:
**What**

I finally had a chance to circle back on what I learned investigating event listener attachment. 

In short, we distinguished between captured and bubbled events, however I believe this is not necessary as of IE9+. Captured do not appear to exhibit the same edge cases as bubbled events, allowing us to simplify event code.

**Why do we want to do this?**

I think we can probably drastically reduce the code for eagerly attaching listeners for events that do not bubble. So far, all events that do not bubble appear to capture on parents elements. 

Longer term, [I wonder if we can remove the `.bind()` when attaching listeners](https://github.com/facebook/react/blob/master/packages/react-dom/src/events/ReactDOMEventListener.js#L180), and maybe even remove top level types.

**More info**

This commit changes event subscription such that all events are attached with capturing. This prevents the need to eagerly attach media event listeners to video, audio, and source tags, and simplifies the event subscription process.

I think we can probably extend this to more event types, maybe [most/all eager event attachments](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMFiberComponent.js#L451-L516), but that will require substantial testing. Until that time, this commit lays the foundation for that work by demonstrating that capturing is applicable in cases where bubbling does not occur.

**Test plan**

1. Verify fixtures on https://react-capture-all-events.surge.sh/media-events
2. Verify fixtures on https://react-capture-all-events.surge.sh/input-change-events
3. Verify fixtures on https://react-capture-all-events.surge.sh/pointer-events

I believe these to be the key event types to test, however this change definitely affects all event listeners.

**I've tested so far in:**

- Internet Explorer: 9, 11
- Edge 16
- Firefox: 49, 60
- Chrome 47, 66
- Safari 9, 11.1

PR #15036, opened 2019-03-06
Add infrastructure for passive/non-passive event support for future API exploration

trueadm:
## Summary
This PR makes a bunch of changes to the underlying implementation for how ReactDOM listens to events on the DOM. Notably, it opens up the possibility of exploring the proper handling of passive/non-passive event listeners. As to not break existing event listeners, they now fall under "legacy", where they work 100% as they do now without any changes in control flow. Also some old code has been tidied up and Flow annotations have been added where they were previously missing.

## Why do we need this?
For future event API exploration we need to properly hook into controlling how event listeners are registered with respect to if they are passive or non-passive. The way ReactDOM deals with events now is to not specify if an event is passive or non-passive. By doing so, we let the browser determine the behaviour behind the scenes. Common cases like when `touchmove` is registered by React to the `document` mean browsers actually make the events `passive` behind the scenes. This is the ideal case, but in the future we also need ways of controlling this behaviour for future event API work.

## Validate this works
This PR is a part of a larger part of work which has implementations and tests that validate that the passive/non-passive logic works. In the interests of keeping this PR slim, I've intentionally omitted that code. In further follow up PRs, I'll add that logic to validate the behaviour of this PR. I know this isn't the most ideal case, but it was the best approach I could come up with.

## Polyfills
In the future, if we decide to ship an API that makes use of passive events, support for `addEventListener` and specifically the third argument on `addEventListener` where the argument is an object needs to supported. There is a [polyfill](https://github.com/WICG/EventListenerOptions/blob/gh-pages/EventListenerOptions.polyfill.js) we could advise people to use or we could try and check for this in DEV and warn. I'm not fully sure on the path we should take yet - maybe it's something we should consider in the future once we're ready to ship said features.


packages/react-dom/src/events/ReactBrowserEventEmitter.js
```diff
@@ -84,22 +85,39 @@ import isEventSupported from './isEventSupported';
...
+  Document | Element | Node,
+  ElementListeningObject,
+> = new WeakMap();
 
-/**
- * To ensure no conflicts with other potential React instances on the page
- */
-const topListenersIDKey = '_reactListenersID' + ('' + Math.random()).slice(2);
+// We store legacy events we listen where we don't use
+// passive/non-passive options on the event listener.
```
philipp-spiess: It seems like those legacy events are also stored in the `ElementListeningObject`. Did I miss something or is this comment outdated?

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -125,70 +136,66 @@ export function isEnabled() {
...
+    const listener = dispatch.bind(null, topLevelType, null);
+    // We don't listen for passive/non-passive
+    addEventBubbleListener(element, rawEventName, listener, null);
+  } else {
+    // Check if interactive and wrap in interactiveUpdates
+    const passiveListener = dispatch.bind(null, topLevelType, true);
+    const activeListener = dispatch.bind(null, topLevelType, false);
+    // We listen to the same event for both passive/non-passive
+    addEventBubbleListener(element, rawEventName, passiveListener, true);
+    addEventBubbleListener(element, rawEventName, activeListener, false);
```
philipp-spiess: If the browser doesn't support passive/non passive events, we would potentially adding two _different_ (since we used `.bind`) event listeners here so they might double fire. This is of course pure speculation but it might need a feature test here as well (and for the capture case).

packages/react-dom/src/events/EventListener.js
```diff
@@ -8,17 +8,33 @@
...
+  isPassive: boolean | null,
 ): void {
-  element.addEventListener(eventType, listener, false);
+  if (isPassive === null) {
+    element.addEventListener(eventType, listener, false);
+  } else {
+    element.addEventListener(eventType, listener, {
+      passive: isPassive,
+      capture: false,
+    });
```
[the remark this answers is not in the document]
trueadm: Good idea. Given the new API is still a while a way from being mainstream, it might be a hard limitation for this new API, as there appears to be no way to polyfill the behaviour.

packages/react-dom/src/events/ReactBrowserEventEmitter.js
```diff
@@ -84,22 +85,39 @@ import isEventSupported from './isEventSupported';
  *    React Core     .  General Purpose Event Plugin System
  */
 
-const alreadyListeningTo = {};
-let reactTopListenersCounter = 0;
+const elementListeningObjects: WeakMap<
+  Document | Element | Node,
+  ElementListeningObject,
+> = new WeakMap();
```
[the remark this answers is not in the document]
trueadm: I'll make that change. Nice one.

packages/react-dom/src/events/ReactBrowserEventEmitter.js
trueadm: We store legacy and nonLegacy in the same object, which goes into the Map.

packages/react-dom/src/events/ReactDOMEventListener.js
trueadm: This is intentional. We want to double listen so the new event APIs in the future can control heuristics given both. They'd likely guard one path given an API. For example `onTouchMovePassive` vs `onTouchMoveActive`.

packages/react-dom/src/events/ReactBrowserEventEmitter.js
philipp-spiess: Makes sense. I think I was confused by:

> We store [..] on the event listener.

packages/react-dom/src/events/ReactDOMEventListener.js
philipp-spiess: Yeah you're right the listeners will only call the respective callbacks so we need them both to fire. My bad. :-)

packages/react-dom/src/events/ReactBrowserEventEmitter.js
trueadm: I've updated the comment to be more clear.

packages/react-dom/src/events/EventListener.js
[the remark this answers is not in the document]
trueadm: I'm not sure how we can test this at runtime to know if the various options are supported without adding quite a few more bytes. Is it worth it? We could just recommend a polyfill like we do in other places for those who need backwards support. https://github.com/WICG/EventListenerOptions/blob/gh-pages/EventListenerOptions.polyfill.js

packages/react-dom/src/events/ReactBrowserEventEmitter.js
```diff
@@ -93,10 +93,15 @@ const elementListeningObjects:
...
-// them in nonLegacy so they do not conflict.
+// This object will contain both legacy and non legacy events.
+// In the case of legacy events, we register an event listener
+// without passing an object with { passive, capture } etc to
+// third argument of the addEventListener call (we use a boolean).
+// For non legacy events, we register an event listener with
+// an object (third argument) with the passive property. We also
+// double listen for the event on the element, as we need to check
+// both paths (where passive is both true and false) so we can
+// handle logic in either case.
```
philipp-spiess: Thanks ❤️  Super clear now.

packages/react-dom/src/events/EventListener.js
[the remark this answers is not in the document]
philipp-spiess: Looking at this polyfill I think it will be complicated to install it in apps that use React to render into a different JavaScript context (`<iframe>`, etc). I will have to do some testing to know for sure, though.

On MDN (https://developer.mozilla.org/de/docs/Web/API/EventTarget/addEventListener) it also seems that the `options` object is not implemented in Edge and IE11 which would affect quite a few users. I don't 100% trust this data though. If you want I can do some cross-browser tests to see which browsers support the options prop? 🙂 

---

> [..] know if the various options are supported 

I think that it is enough if we only test for passive event support. There are no other options that we would set. If passive events are not supported, we don't need the `options` prop.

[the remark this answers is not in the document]
trueadm: We could just fall back to not supplying options, which would allow `preventDefault` to work in all cases (not just nonPassive). What we could do, is set a flag to state this is the case and block `preventDefault()` on the SyntheticEvent object.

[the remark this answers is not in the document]
nhunzaker: Could we just capture all events?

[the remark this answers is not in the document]
nhunzaker: Here we go:
https://github.com/facebook/react/pull/12919

@philipp-spiess has a nice summary of the problems with my original attempt here:
https://github.com/facebook/react/pull/12919#issuecomment-395224674

packages/react-dom/src/events/ReactBrowserEventEmitter.js
```diff
@@ -84,22 +85,47 @@ import isEventSupported from './isEventSupported';
...
 
-const alreadyListeningTo = {};
-let reactTopListenersCounter = 0;
+const PossiblyWeakMap = typeof WeakMap === 'function' ? WeakMap : Map;
+const elementListeningObjects:
+  | WeakMap
+  | Map<
+      Document | Element | Node,
+      ElementListeningObject,
+    > = new PossiblyWeakMap();
```
nhunzaker: Ahh, :fire:. This is a great update!

nhunzaker: I've been thinking about something similar for wrapper state meta data that is attached to inputs, like here:

https://github.com/facebook/react/blob/d0289c7e3a2dfc349dcce7f9eb3dee22464e97bd/packages/react-dom/src/client/ReactDOMInput.js#L118-L126

packages/react-dom/src/events/ReactBrowserEventEmitter.js
```diff
@@ -125,62 +151,74 @@ function getListeningForDocument(mountAt: any) {
...
+        // We listen to them on the target DOM elements.
+        // Some of them bubble so we don't want them to fire twice.
+        break;
+      default:
+        // By default, listen on the top level to all non-media events.
+        // Media events don't bubble so adding the listener wouldn't do anything.
+        const isMediaEvent = mediaEventTypes.indexOf(dependency) !== -1;
+        if (!isMediaEvent) {
+          trapBubbledEvent(dependency, mountAt, isLegacy);
+        }
```
nhunzaker: Back to the capture thing. If we could ensure execution order is maintained, we could just capture all the media events and get rid of this block.

That's out of the context of this PR, but still I wonder.

packages/react-dom/src/events/EventListener.js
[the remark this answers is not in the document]
trueadm: Won't capturing all events lead to the same issues pointed out in those threads? I'm happy to change it, if it makes sense to do so without any side-effects.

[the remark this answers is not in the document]
nhunzaker: Yes, sorry; I didn't give a very helpful comment either. 

At the time, I stopped work on this because it wasn't clear if there was a way forward. However I was curious if you had any ideas.

It might just be a dead end.

trueadm:
As per the request of @gaearon, I've run a bunch of performance tests on this PR vs master. I slightly optimized the approach and now I see a slight performance gain in Chrome and Safari. No noticeable differences in Edge and Firefox.

I'd really appreciate it if we could get this PR landed soon.

nhunzaker:
Where did you land on feature checking event listener options objects? Adding a feature detection like the [MDN recommendation](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Safely_detecting_option_support) seems fine and with minimal impact.

Also if passive events won't be publicly exposed for a while maybe it could be kicked down the road a bit.

Does this need additional testing? I'd be happy to give it a run across browsers.

trueadm:
@nhunzaker I added that support just now. It would be awesome if you could do some manual tests for it too :)

packages/events/EventPluginHub.js
```diff
@@ -163,17 +163,27 @@ function extractEvents(
   targetInst: null | Fiber,
   nativeEvent: AnyNativeEvent,
   nativeEventTarget: EventTarget,
+  passive: null | boolean,
```
gaearon: Can we avoid null | boolean? I imagine this would cause deopts.
For example, it could be an enum (0, 1, 2).

packages/react-dom/src/events/BeforeInputEventPlugin.js
```diff
@@ -462,6 +462,8 @@ function extractBeforeInputEvent(
 const BeforeInputEventPlugin = {
   eventTypes: eventTypes,
 
+  isLegacy: true,
```
nhunzaker: I'm curious how this flag relates to interactive updates and if this will end up being a more general "noPassive" flag or something.

packages/events/EventPluginHub.js
```diff
@@ -163,17 +163,27 @@ function extractEvents(
...
 ): Array<ReactSyntheticEvent> | ReactSyntheticEvent | null {
   let events = null;
   for (let i = 0; i < plugins.length; i++) {
     // Not every plugin in the ordering may be loaded at runtime.
     const possiblePlugin: PluginModule<AnyNativeEvent> = plugins[i];
     if (possiblePlugin) {
+      // Check if the plugin supports legacy or non legacy events
+      // based on the passive flag being null or a boolean
+      if (
+        (possiblePlugin.isLegacy && passive !== null) ||
```
gaearon: To avoid checking every plugin every time, can we have two lists of plugins, and switch early depending on the event?

packages/events/ResponderEventPlugin.js
```diff
@@ -497,6 +497,8 @@ const ResponderEventPlugin = {
 
   eventTypes: eventTypes,
 
+  isLegacy: true,
```
gaearon: What does "legacy" really mean here? That it can't handle passive events? Some other constraints?

packages/react-dom/src/events/ReactBrowserEventEmitter.js
```diff
@@ -125,62 +150,74 @@ function getListeningForDocument(mountAt: any) {
  */
 export function listenTo(
   registrationName: string,
-  mountAt: Document | Element,
-) {
-  const isListening = getListeningForDocument(mountAt);
+  mountAt: Document | Element | Node,
+  isLegacy: boolean,
+): void {
```
gaearon: This changes behavior from `isLegacy` effectively being `true` (because it didn't exist) to `false` by default. Are we sure we have no internal callsites calling this? Are they unaffected?

packages/react-dom/src/events/ReactBrowserEventEmitter.js
```diff
@@ -84,22 +85,46 @@ import isEventSupported from './isEventSupported';
...
+  isLegacy: boolean,
+): Set<DOMTopLevelEventType> {
+  let listeningObject = elementListeningObjects.get(element);
+  if (listeningObject === undefined) {
+    listeningObject = createElementListeningObject();
+    elementListeningObjects.set(element, listeningObject);
   }
-  return alreadyListeningTo[mountAt[topListenersIDKey]];
+  const listeningKey = isLegacy ? 'legacy' : 'nonLegacy';
+  return listeningObject[listeningKey];
```
gaearon: Can you make this if/else with two concrete names instead of access through a string?

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -125,70 +137,108 @@ export function isEnabled() {
...
+    // We don't listen for passive/non-passive
+    eventListener(element, rawEventName, listener, null);
+  } else {
+    if (passiveBrowserEventsSupported) {
+      // Check if interactive and wrap in interactiveUpdates
+      const activeListener = bindDispatch(dispatch, topLevelType, false);
+      const passiveListener = bindDispatch(dispatch, topLevelType, true);
+      // We listen to the same event for both passive/non-passive
+      eventListener(element, rawEventName, passiveListener, true);
+      eventListener(element, rawEventName, activeListener, false);
```
gaearon: Can you explain more about why we listen for both? Sorry I'm not quite following.

packages/react-dom/src/events/SimpleEventPlugin.js
```diff
@@ -206,6 +206,8 @@ const SimpleEventPlugin: PluginModule<MouseEvent> & {
 } = {
   eventTypes: eventTypes,
 
+  isLegacy: true,
```
gaearon: So which ones are not legacy? Are any?

packages/events/ResponderEventPlugin.js
trueadm: Yes that’s what it means

packages/react-dom/src/events/ReactBrowserEventEmitter.js
trueadm: I’ve check all callsites and updated them

packages/react-dom/src/events/ReactDOMEventListener.js
trueadm: We listen to both so we can handle both cases in future plugins event APIs. This is because we listen on the root ahead of time so we need both to handle specific logic that is only possible in one or the other

packages/react-dom/src/events/SimpleEventPlugin.js
trueadm: None yet.

trueadm:
I've updated the PR. I've removed the confusing `null | boolean` logic for `passive` and instead went with passing through an enum, which shows intent better. I've also addressed the PR feedback. There are no actors using the new passive/non-passive event system yet, but that will come soon with the experimental event API proposal I have brewing (which also doesn't require the event plugin system too).

nhunzaker:
I dug into some browser testing and realized that I wouldn't actually be able to verify passive event handlers. Do I have that right?

Still, the event listener tracking updates and no-passive code paths execute correctly for Chrome (49, 72), Firefox (65, 52), IE (9, 10), and iOS Safari 9.

nhunzaker:
Here's the pen I used to verify the feature check:

https://codepen.io/nhunzaker/pen/QoaVWX

Debug view is helpful:
https://s.codepen.io/nhunzaker/debug/QoaVWX/wQMPoNBqndqk

I verified the false/positive case by version for Firefox, Safari, IE, Edge, and Chrome:
https://caniuse.com/#feat=passive-event-listener

trueadm:
@nhunzaker In order to test this PR, you'd need to manually change the PR somewhat so it executes for another plugin. I'm working on a subsequent PR that adds the basis for that new event API. It essentially is a new API that doesn't use the event plugin system at all and instead has a more explicit system for moving the handling of events to outside of ReactDOM. Once we get this PR landed, I can add another PR that adds that basic system in place. That will also make this whole thing testable.

Thanks for checking the codepens, that's awesome :)

nhunzaker:
> It essentially is a new API that doesn't use the event plugin system at all and instead has a more explicit system for moving the handling of events to outside of ReactDOM.

Very cool. Looking forward to seeing this :+1:

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -494,41 +497,41 @@ export function setInitialProperties(
...
     case 'details':
-      trapBubbledEvent(TOP_TOGGLE, domElement);
+      trapBubbledEvent(TOP_TOGGLE, domElement, true);
       props = rawProps;
       break;
     case 'input':
       ReactDOMInputInitWrapperState(domElement, rawProps);
       props = ReactDOMInputGetHostProps(domElement, rawProps);
-      trapBubbledEvent(TOP_INVALID, domElement);
+      trapBubbledEvent(TOP_INVALID, domElement, true);
```
gaearon: Per our discussion, it seems like these code paths will never "want" to get two listeners attached/invoke. So instead of branching deeply and passing an argument through, let's fork the innermost implementation and call the fork from the new code when we want to. That will also likely let us DCE more based on a feature flag.

trueadm:
@gaearon @sebmarkbage I've re-written the implementation based on feedback you both gave me. It now uses bitwise flags to express the intention of the event system. Furthermore, everything should almost now get DCE due to `enableEventAPI`. Thanks!

trueadm:
Nice, this actually makes ReactDOM 0.1% smaller :P

packages/react-dom/src/events/forks/EventListener-www.js
```diff
@@ -16,16 +16,18 @@ export function addEventBubbleListener(
   element: Element,
   eventType: string,
   listener: Function,
+  passive?: boolean,
 ): void {
-  EventListenerWWW.listen(element, eventType, listener);
+  EventListenerWWW.listen(element, eventType, listener, passive);
```
sebmarkbage: This doesn't seem right. The internal EventListener doesn't accept a 4th argument (and Event.listen accepts a forth argument which is a priority and won't work with this).

trueadm: I’ll update the www version

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -125,70 +155,89 @@ export function isEnabled() {
...
+      : dispatchEvent;
+    const rawEventName = getRawEventName(topLevelType);
+    let eventFlags = RESPONDER_EVENT_SYSTEM;
+
+    // If passive option is not supported, then the event will be
+    // active and not passive, but we flag it as using not being
+    // supported too. This way the responder event plugins know,
+    // and can provide polyfills if needed.
+    if (passive) {
+      if (passiveBrowserEventsSupported) {
```
sebmarkbage: This seems like the wrong check to me. Shouldn't this case be inverted? This will mark it as active and PASSIVE_NOT_SUPPORTED only if passive events *are* supported.

trueadm: Well spotted!

packages/events/EventSystemFlags.js
```diff
@@ -0,0 +1,16 @@
...
+ * @flow
+ */
+
+export type EventSystemFlags = number;
+
+export const PLUGIN_EVENT_SYSTEM = 1;
+export const RESPONDER_EVENT_SYSTEM = 1 << 1;
+export const IS_PASSIVE = 1 << 2;
+export const IS_ACTIVE = 1 << 3;
+export const PASSIVE_NOT_SUPPORTED = 1 << 4;
```
NE-SmallTown: why we don't use literal number here just like `maxSigned31BitInt`?

trueadm: These are used to do bitwise operations on, as there can be a combination of the above being used.

eps1lon: I think the question was why `1 << 4` is used over `16`. `1 << 4` makes the intent clearer. `maxSigned31BitInt` "solves" this by adding a comment. I don't think it matters what pattern is used.

NE-SmallTown: yea, just like @eps1lon said

trueadm: This pattern is clear, as we may add more in the future, which is simply `1 << 5`, `1 << 6` etc. We use this approach throughout the React codebase and we've verified that Google Closure Compiler simplifies the constant value.

PR #15055, opened 2019-03-07
[ESLint] Tweak setState updater message and add useEffect(async) warning

gaearon:
What it says.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -949,6 +949,29 @@ const tests = {
         }
       `,
     },
+    {
+      code: `
+        function App() {
+          const [query, setQuery] = useState('react');
+          const [state, setState] = useState(null);
+          useEffect(() => {
+            let ignore = false;
+            fetchSomething();
+            async function fetchSomething() {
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -2304,7 +2327,7 @@ const tests = {
       errors: [
         "React Hook useEffect has a missing dependency: 'state'. " +
           'Either include it or remove the dependency array. ' +
-          `You can also write 'setState(state => ...)' ` +
+          `You can also write 'setState(s => ...)' ` +
```
threepointone: why did you decide to do a single character here?

gaearon: Showed it to people and they were confused seeing the same variable name because it wasn't obvious it's being shadowed. This explanation clicked tho.

gaearon: I guess I'm open to other changes, just want to avoid the shadowing. If people think this is confusing we can brainstorm more options.

threepointone: No I trust your judgement, I was just curious. It makes sense. Thanks!
(ashamed to say I usually use `x =>...`, which people around me hate)

PR #15061, opened 2019-03-08
[Partial Hydration] Render client-only content at normal priority

sebmarkbage:
If the dehydrated suspense boundary's fallback content is terminal there is nothing to show. We need to get actual content on the screen soon.

If we deprioritize that work to offscreen, then the timeout heuristics will be wrong.

Therefore, if we have no current and we're already at terminal fallback state we'll immediately schedule a deletion and upgrade to real suspense.

Ideally, if we're in a pending state, we should also just register the listener instead of retrying at offscreen, but meh.

The first commit is from another branch I needed to factor out but might as well include it here.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
@@ -762,6 +762,67 @@ describe('ReactDOMServerPartialHydration', () => {
     expect(ref.current).toBe(span);
   });
 
+  it('replaces the fallback within the maxDuration if there is a nested suspense in a nested suspense', async () => {
+    let suspend = false;
+    let promise = new Promise(resolvePromise => {});
+    let ref = React.createRef();
+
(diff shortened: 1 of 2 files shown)


sebmarkbage:
I think this is still a problem if there is one more suspense boundary around this which has content. In that scenario that boundary will hydrate at offscreen and the same thing happens.

sebmarkbage:
There is a conflict between our normal heuristic that dehydrated boundaries should be hydrated at offscreen priority. We can do that because there is no reason to hydrate content faster than that if it’s already being displayed.

However if there is nested fallback content that can only be displayed by client rendering it which has higher pri. But we can’t get there unless we first render the parent which is lower pri.

sebmarkbage:
The new approach just schedules the client content at higher pri once it gets there. It means that it can be delayed compared to when it would normally be rendered but that seems fine. That only happens if you start interacting anyway.

This is kind of non-idiomatic anyway because you’re expected to server render everything in the normal case.

Issue #15057, opened 2019-03-07
setState in useEffect causing a "React state update on an unmounted component" warning


KidkArolis:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
The warning is triggered after an intricate sequence of events. I stumbled upon it by accident, assumed it was an error in my code, but then step by step removed everything in my application until I was left with a fairly small bit of code that doesn't seem to be doing anything illegal from the API point of view, yet is triggering a warning.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem:**

https://codesandbox.io/s/q87882qv64

The example is my real application code trimmed down as much as possible to demonstrate the warning. So some of the code might be a bit nonsensical/contrived, but that's because I removed lots of surrounding code leaving only the relevant bits for reproducing the issue.

In other words, there might weird looking uses of useEffect and weird sequencing of things, but that sort of falls out of how I've structured my routes, state, components in the full app, etc.

**What is the expected behavior?**

I would like to know if

a) is this a React bug that I stumbled upon that should be fixed?
b) is this something I'm doing "wrong" and how I could fix that in my application (i.e. is this a real memory leak being caused because of the way I structured the code)

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

16.8.4

KidkArolis:
Just to add a bit about why I think this is a bug.

The recommendation from the warning is to cleanup any subscriptions inside useEffect in the effect destroy function. But in this codesandbox, you should see that useEffect destroy is only called **after** the setState is called. There is no way to clean that up afaik since we're not yet aware that the component is being unmounted.

gaearon:
This code is definitely buggy btw. You’re passing [] as effect deps but close over both atom and selector. That’s not legit. I suggest you to adopt the `exhaustive-deps` rule and fix those bugs first.

Implementing subscriptions correctly is actually somewhat tricky. Check this for inspiration: https://gist.github.com/bvaughn/e25397f70e8c65b0ae0d7c90b731b189

Does that work better?

KidkArolis:
I'll continue working on this codesandbox to try and minimise the code and isolate the issue better.

I am aware of some of the subscription pitfalls, this code is not what I use in production version. This is just an attempt at finding a minimal set of code that triggers this warning without using the React APIs "incorrectly" (except for the thing you pointed about about effect's deps, but I doubt that's the issue, I'll check).

[...]

Issue #14369, opened 2018-12-01
setState hook inside useEffect can cause unavoidable warning  Can't perform a React state update


istarkov:
**BUG**

**What is the current behavior?**

Example: https://codesandbox.io/s/6y1x2zr21n clicking on OK button cause `Warning: Can't perform a React state update on an unmounted component.`

The problem that unsubscribe is called during B event `setVisible(v => false);` call, see logs:

```
SET VISIBLE BEFORE 
UNSUBSCRIBE 
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in Child (created by Holder)
SET VISIBLE AFTER 
```

In our case we have this even without RAF call, but on `transitionend` DOM event.
(It's occurred randomly and rare in our codebase as transitionend event should be called exactly at needed time, but example showed what happens)
Seems like it occurred only if you have a `setState` call during useEffect callback like `setRefresh(v => v + 1);` (_inside provided code_) (_after rewriting our codebase to avoid setState calls in useEffect the error has gone_)

Code
```javascript
import React from "react";
import ReactDOM from "react-dom";
import mitt from "mitt";

const emitter = mitt();

const Child = () => {
  const [visible, setVisible] = React.useReducer((s, a) => a, true);
  React.useEffect(() => {
    const handle = () => {
      console.log("SET VISIBLE BEFORE");
      setVisible(v => false); // <--- THIS CALL CAUSES UNSUBSCRIBE AND WARNING ABOUT STATE
      console.log("SET VISIBLE AFTER");
    };
    emitter.on("B", handle);
    return () => {
      console.log("UNSUBSCRIBE");
      emitter.off("B", handle);
    };
  }, []);

  return <div>{visible && <h1>CHILD TEXT</h1>}</div>;
};

const Holder = () => {
  const [refresh, setRefresh] = React.useState(0);
  const visible = React.useRef(true);
  React.useEffect(() => {
    if (refresh === 1) {
      visible.current = false;
      setRefresh(v => v + 1); // <--- This state change from effect caused problems
    }
    const handle = () => {
      setRefresh(v => v + 1);
    };
    emitter.on("A", handle);
    return () => {
      emitter.off("A", handle);
    };
  });

  return <div>{visible.current && <Child />}</div>;
};

function App() {
  return (
    <div>
      <Holder />
      <button
        onClick={() => {
[...]

PR #15067, opened 2019-03-08
Failing test for false positive "setState on unmounted component" warning

gaearon:
Failing test for https://github.com/facebook/react/issues/15057.
Possibly relates to some cases of https://github.com/facebook/react/issues/14369.

The crux of the issue is that **even if we use a guard variable, sometimes an effect can't `setState` without triggering a warning**. In particular, this happens if:

* We set state in a component (e.g. from an event emitter call)
* We flush previous effect before we the state gets set
* Flushing previous effect caused a state change in a parent which resulted in component unmounting
* The scheduler thinks we *called* `setState` on an unmounted component, and warns

I'm not sure how to fix it. I imagine I'd expect that if `flushPassiveEffects` [here](https://github.com/facebook/react/blob/d0289c7e3a2dfc349dcce7f9eb3dee22464e97bd/packages/react-reconciler/src/ReactFiberHooks.js#L1109) led to unmounting of that component, we ideally shouldn't warn [here](https://github.com/facebook/react/blob/d0289c7e3a2dfc349dcce7f9eb3dee22464e97bd/packages/react-reconciler/src/ReactFiberScheduler.js#L1860).

I'm not sure if there are any bigger issues in cases where it's *not* unmounted though. Is this just a stray warning or is there a larger potential concern?

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
@@ -1699,6 +1699,57 @@ describe('ReactHooks', () => {
     ).toThrow('Hello');
   });
 
+  it('does not fire a false positive warning when previous effect unmounts the component', () => {
+    let {useState, useEffect} = React;
+    let globalListener;
+
+    function A() {
+      const [show, setShow] = useState(true);
+      function hideMe() {
+        setShow(false);
+      }
+      return show ? <B hideMe={hideMe} /> : null;
+    }
+
+    function B(props) {
+      return <C {...props} />;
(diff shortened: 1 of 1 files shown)


acdlite:
We could just not warn if there are pending passive effects.

gaearon:
https://github.com/facebook/react/pull/15067

acdlite:
I think this is the same as the serial/discrete events problem. Added a comment in your PR: https://github.com/facebook/react/pull/15122#issuecomment-473591116

gaearon:
This will be fixed in 16.9

PR #15071, opened 2019-03-09
Compile invariant directly to throw expressions

acdlite:
*This is the first of two PRs to improve our error extraction process.*

Ok I went down a bit of a rabbit hole.

It started with this todo: 

https://github.com/facebook/react/blob/d0289c7e3a2dfc349dcce7f9eb3dee22464e97bd/packages/react-reconciler/src/ReactFiberUnwindWork.js#L390-L398

For some context, production builds of React do not include error messages. We strip them out at compile time and [replace them with error codes](https://reactjs.org/blog/2016/07/11/introducing-reacts-error-code-system.html) and a link to a URL that displays the full message. Our error messages can be as verbose as we want without increasing  code size. Here's an example: https://reactjs.org/docs/error-decoder.html?invariant=109&args%5B%5D=Foo.

The error code script works by finding calls to our `invariant` module, looking up the message in a map, and replacing it with the corresponding code. Because we use `invariant` throughout our codebase to assert expected behavior, including for user facing errors, this mostly works really well.

But error in the excerpted code above doesn't use `invariant`. It's part of React's error handling path; we can't use `invariant` there because invariant will immediately throw. Instead, we create an error object and pass it along to the next step of the error handling algorithm. That means this error is not being replaced with an error code.

While thinking about the best way to solve this, I collected some other flaws in the way we do error extraction:

- The `invariant` module is a Facebook-ism that's not commonly used by other developers. It's a marginal burden for first time contributors to learn how to use it. For example, `invariant` throws an error when its first argument evaluates to `false`, but most people are accustomed to checking for the positive case: `if (cond) throw Error(msg)` as opposed to `invariant(!cond, msg)`. It's also not obvious that you're supposed to use `invariant` for errors if you've never contributed to React before.
- `invariant` was originally meant to assert some property of the program that is always supposed to be true. These types of errors should only surface to the user if there's a bug in the program itself. Over time, both Facebook and React have started using it for user facing errors, too.
- There's minimal enforcement for using `invariant` in our codebase. We have our sizebot that reports increases in bundle size, but if a PR already increases bundle size for other reasons, a stray `Error` could increase it slightly further without being noticed by the reviewer.
- As described above, `invariant` can only be used to immediately throw an error. It's not suited for cases where you need to create an error and pass it along.

### Compile `invariant` directly to throw expressions.

This PR turns `invariant` into a compile-time only module. It turns code like this:

```js
invariant(condition, 'A %s message that contains %s', adj, noun);
```

into this:

```js
if (!condition) {
  if (__DEV__) {
    throw ReactError(`A ${adj} message that contains ${noun}`);
  } else {
    throw ReactErrorProd(ERR_CODE, [adj, noun]);
  }
}
```

The only thing `ReactError` does is return an error whose name is set to "Invariant Violation" to match the existing behavior.

`ReactProdError` is a special version used in production that throws a minified error code, with a link to see to expanded form. This replaces the `reactProdInvariant` module.

The runtime behavior is the same as it is today.

### Compile normal `Error` constructors to `ReactError` and `ReactErrorProd`

I'll open this part as a separate PR.

### Use a lint rule to enforce that error messages are written in the correct format

This will go in the next PR, too.


sebmarkbage:
AFAIK, we still need to compile to `invariant(...)` calls for the FB builds because we have a different transform internally that deals with those. Has that changed?

acdlite:
Not sure, I was hoping one of you had more context on that.

acdlite:
Looks like Logview will still get the messages but our `invariant` module does other stuff I'm less sure about so I'll compile `throw ReactProdError` to `invariant` in the FB builds.

sebmarkbage:
I'll yield to @gaearon (www) and @sahrens (fbsource) to review this.

packages/shared/ReactError.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ */
+
+// Do not require this module directly! Use a normal error constructor with
+// template literal strings. The messages will be converted to ReactError during
+// build, and in production they will be minified.
+
+function ReactError(message) {
+  const error = new Error(message);
+  error.name = 'Invariant Violation';
```
gaearon: While we're at it, can we just say "React Error"? Is that a breaking change? 😄

scripts/error-codes/__tests__/__snapshots__/minify-error-messages.js.snap
```diff
@@ -0,0 +1,96 @@
...
+"import _ReactErrorProd from 'shared/ReactErrorProd';
+import _ReactError from 'shared/ReactError';
+
+import invariant from 'shared/invariant';
+(function () {
+  if (!condition) {
+    if (__DEV__) {
+      throw _ReactError(\`Expected a component class, got \${Foo}.\${Bar}\`);
+    } else {
+      throw _ReactErrorProd('18', [Foo, Bar]);
```
gaearon: While we're here, can we make them actual numbers?

packages/shared/ReactError.js
acdlite: I'd be comfortable changing in a minor

acdlite:
I looked into it more, and in www, the only things the `invariant` module does is:

- Strip out the error message in production. We specifically don't want this behavior because it removes our error codes. @gaearon previously built a way to opt us out of this transform but compiling directly to `throw` statements sidesteps this problem.
- Remove the `invariant` call from the stack trace. Again, by compiling directly to `throw` statements instead of adding another indirection, we avoid this problem completely.

I'll look into fbsource, too.

acdlite:
AFAICT the version in fbsource does nothing special at all.

packages/shared/ReactError.js
sebmarkbage: I think it's a breaking change. It breaks continuity in logs.

scripts/error-codes/extract-errors.js
```diff
@@ -76,17 +76,20 @@ module.exports = function(opts) {
...
-
-            existingErrorMap[errorMsgLiteral] = '' + currentID++;
+            addtoErrorMap(errorMsgLiteral);
           }
         },
       },
     });
   }
 
+  function addtoErrorMap(errorMsgLiteral) {
```
gaearon: Nit: addToErrorMap

scripts/error-codes/minify-error-messages.js
```diff
@@ -0,0 +1,133 @@
...
+
+          // Import ReactErrorProd
+          const reactErrorProdIdentfier = file.addImport(
+            'shared/ReactErrorProd',
+            'default',
+            'ReactErrorProd'
+          );
+
+          // Outputs:
+          //   throw ReactErrorProd(ERR_CODE, [adj, noun]);
```
gaearon: Can we make it `throw ReactErrorProd(ERR_CODE, adj, noun);`? Then wouldn't need `[]` in minified output on every callsite.

acdlite: I wasn't sure about this part so I just asked Sebastian and his rationale was that using arguments will make ReactErrorProd slightly slower, but using an array will likely make all the functions that throw slightly slower to compile, so it's hard to say which way is better. But since ReactErrorProd is in an error path, and fewer bytes is generally better, no array is good.

PR #15072, opened 2019-03-09
[Outdated; needs revision] Add support for normal errors to error code infra

acdlite:
## Based on #15071

*Refer to the base PR for more context*

---

This compiles normal error constructors (`new Error(...)` and `Error(...)`) to `ReactError` in and `ReactErrorProd`, adding support to our error code replacement infra.

It also adds a lint rule to enforce that error messages are written in an extraction-friendly format.

Error messages must be a concatenation of strings and string literals. This is partly an artificial restriction, because I could generate a template string even from a dynamic expression, but requiring the use of a template literal for dynamic values helps prevent mistakes like `const msg = 'string'; throw new Error(msg)`.

You can add `extract-error/skip` to messages that don't need to be replaced with an error code.


acdlite:
Closing because the code is outdated but we should still do this

PR #15078, opened 2019-03-10
[eslint] Wording tweaks

sophiebits:
I think these are a little clearer. @gaearon asked me to take a look.

### DIFF
--- a/fixtures/eslint/.eslintrc.json
+++ b/fixtures/eslint/.eslintrc.json
@@ -1,7 +1,7 @@
 {
   "root": true,
   "parserOptions": {
-    "ecmaVersion": 6,
+    "ecmaVersion": 8,
     "sourceType": "module",
     "ecmaFeatures": {
       "jsx": true
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -100,7 +100,7 @@ export default {
             message:
               `React Hook ${reactiveHookName} does nothing when called with ` +
               `only one argument. Did you forget to pass an array of ` +
-              `dependencies?.`,
+              `dependencies?`,
           });
         }
         return;
@@ -455,7 +455,7 @@ export default {
             message:
               `The ref value '${dependency}.current' will likely have ` +
               `changed by the time this effect cleanup function runs. If ` +
-              `this ref points to a component rendered by React, copy ` +
+              `this ref points to a node rendered by React, copy ` +
               `'${dependency}.current' to a variable inside the effect, and ` +
               `use that variable in the cleanup function.`,
           });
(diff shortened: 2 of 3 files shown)


sophiebits:
I didn't update the test file yet.

fixtures/eslint/.eslintrc.json
```diff
@@ -1,7 +1,7 @@
 {
   "root": true,
   "parserOptions": {
-    "ecmaVersion": 6,
+    "ecmaVersion": 8,
```
sophiebits: Can't use async arrow functions without this.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -501,28 +502,14 @@ export default {
...
-                  context.report({
-                    node: declaredDependencyNode,
-                    message:
-                      `The '${
-                        declaredDependencyNode.raw
-                      }' literal is not a valid dependency ` +
-                      'because it never changes. You can safely remove it.',
-                  });
-                }
+                context.report({
```
sophiebits: I'm not 100% sure about removing the string special case. I understand people could be confused by the normal message? But I'm worried that it's confusing in the case that it doesn't correspond to an actual variable. (Maybe we could check that it looks like a valid identifier/member expression?)

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -453,11 +453,11 @@ export default {
...
             node: dependencyNode.parent.property,
             message:
-              `Accessing '${dependency}.current' during the effect cleanup ` +
-              `will likely read a different ref value because by this time React ` +
-              `has already updated the ref. If this ref is managed by React, store ` +
-              `'${dependency}.current' in a variable inside ` +
-              `the effect itself and refer to that variable from the cleanup function.`,
+              `The ref value '${dependency}.current' will likely have ` +
+              `changed by the time this effect cleanup function runs. If ` +
+              `this ref points to a component rendered by React, copy ` +
```
gaearon: "a component" — do you think people will assume divs aren't component?

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
gaearon: I think ideally we'd special case values that are in scope in this message.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -644,7 +630,10 @@ export default {
                 fn.name.name
               }' definition into its own useCallback() Hook.`;
           }
+          // TODO: What if the function needs to change on every render anyway?
+          // Should we suggest removing effect deps as an appropriate fix too?
           context.report({
+            // TODO: Why not report this at the dependency site?
```
gaearon: I thought maybe it's worth reporting at both. But I want to report it here because you usually should start fixing it here. (Such as by either cutting it to paste later — outside component or inside effect, or by wrapping in useCallback.) I'm worried that reporting at the dependency site will make people think they're supposed to remove it from the array and silence the rule.

gaearon:
Uhh I don't know why React shows this as "closed"

sophiebits:
Thanks for getting the tests for me.

PR #15087, opened 2019-03-12
Add ESLint warning for createRef deps

philipp-spiess:
Alternative to #15086

After pitching my idea on [Twitter][], @gaearon mentioned that there might be a valid use case for calling `createRef()` inside function components.

As an alternative idea we could add it to the ESLint rule so that it be silenced easily.

The current implementation is very rudimentary (I somehow have the feeling that I should not pass those dependencies down to the very bottom but report earlier). Any input is greatly appreciated.

I'm not sure about making it auto fixable since `useRef()` does have a different meaning than `createRef()` so we should have a human look over it I believe.

[Twitter]: https://twitter.com/dan_abramov/status/1105245485487611904

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -972,6 +972,16 @@ const tests = {
         }
       `,
     },
+    {
+      code: `
+        const ref = React.createRef();
+        function Thing() {
+          useEffect(() => {
+            console.log(ref.current);
+          }, []);
+        }
+      `,
+    },
   ],
   invalid: [
     {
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -4431,6 +4441,72 @@ const tests = {
...
+          useEffect(() => {
+            console.log(ref.current);
+          }, []);
+        }
+      `,
+      errors: [
+        `React Hook useEffect has a dependency that was created by ` +
+          `creatRef(). This will create a new ref object on every re-render ` +
+          `so that the hook will always need to re-run. Did you mean to use ` +
+          `useRef() instead?`,
```
philipp-spiess: I think in this case we need a different message as it _does not_ have a dependency and won't re-run on every render. I'd say we still encourage this pattern even if the function captures only the first ref value and not the ones on re-render.

PR #15094, opened 2019-03-12
Make setNativeProps a no-op with Fabric renderer

TheSavior:
We aren't sure if we want to implement setNativeProps for Fabric. We are making it a no-op with a warning so we can catch these use cases and migrate them to other approaches. Many of the calls are JS driven animations which we are converting to use the native driver.


aleclarson:
@TheSavior Is there a public discussion about the removal of `setNativeProps`?

PR #15101, opened 2019-03-13
update schedule/ to scheduler/ in error message

Jephuff:
The logging about scheduler/tracing were still referencing schedule/tracing. 
This adds the `r`

Test Plan:
----------
ran these commands
```
yarn test
yarn test-prod
yarn prettier
yarn lint
yarn flow test
```

Changelog:
----------
[General] [Changed] - Updated reference to schedule/tracing to scheduler/tracing in error messages


gaearon:
It's not correct to update this file manually. Instead you'll want to update the error message itself in the code. If it's fixed in the code that there's nothing to correct.

Jephuff:
Ahh yes, sorry about that!

PR #15108, opened 2019-03-14
Adds experimental event API scaffolding

trueadm:
## Summary

This PR adds some scaffolding to React needed for the event API experiment that we're testing out internally at Facebook. The API will be behind a `enableEventAPI` flag to ensure the logic is kept in isolation.

This scaffolding in this PR doesn't add any functionality. This PR inserts placeholders, symbols for the new fibers, functions and tags/types in preparation for follow up PRs that add the implementations. In the interests of making this experimental event API easier to review and consume, I've opted to create lots of smaller PRs.

This PR is only one part of the work to add the event API, so expect other PRs that expand on this and fill in the implementations. https://github.com/facebook/react/pull/15036 is the previous part of work.

I plan on writing up an RFC explaining this event API, along with how the internals work (such as the new symbols) once we've had a chance to iron out some details internally.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -43,6 +43,7 @@ import {
   IncompleteClassComponent,
   MemoComponent,
   SimpleMemoComponent,
+  EventTarget,
 } from 'shared/ReactWorkTags';
 import {
   invokeGuardedCallback,
@@ -90,6 +91,7 @@ import {
   hideTextInstance,
   unhideInstance,
   unhideTextInstance,
+  handleEventTarget,
 } from './ReactFiberHostConfig';
 import {
   captureCommitPhaseError,
@@ -299,6 +301,7 @@ function commitBeforeMutationLifeCycles(
     case HostText:
     case HostPortal:
     case IncompleteClassComponent:
+    case EventTarget:
       // Nothing to do for these component types
       return;
     default: {
@@ -584,8 +587,8 @@ function commitLifeCycles(
       return;
     }
     case SuspenseComponent:
-      break;
     case IncompleteClassComponent:
+    case EventTarget:
       break;
     default: {
       invariant(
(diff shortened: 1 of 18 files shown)


philipp-spiess:
Now I'm _really_ curious about what you've worked on internally at Facebook.

aweary:
@philipp-spiess https://www.facebook.com/careers/ 😉

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1213,6 +1216,12 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
     case IncompleteClassComponent: {
       return;
     }
+    case EventTarget: {
+      const newProps = finishedWork.memoizedProps;
+      const type = finishedWork.type.type;
+      handleEventTarget(type, newProps, finishedWork);
```
sebmarkbage: This block should be wrapped in the feature flag check so that it can remove everything including the property access etc. That'll also ensure DCE of the function itself once it has more content.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -762,6 +765,16 @@ function completeWork(
       }
       break;
     }
+    case EventComponent: {
+      const rootContainerInstance = getRootHostContainer();
+      const responder = workInProgress.type.responder;
+      handleEventComponent(responder, rootContainerInstance, workInProgress);
```
sebmarkbage: Same here. Wrap in feature flag check here.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -762,6 +765,16 @@ function completeWork(
...
+      break;
+    }
+    case EventTarget: {
+      markUpdate(workInProgress);
```
sebmarkbage: Here too.

packages/react-reconciler/src/ReactFiberCommitWork.js
trueadm: I was going to do this in a follow up PR anyway. I can put in a other PR to do this sooner. :)

PR #15111, opened 2019-03-14
Remove facts tracker

kaylangan:
It looks like Travis is no longer used in this project so I've deleted/updated the references to Travis.


gaearon:
I was thinking we should just remove this "facts tracker", we don't use it anyway.

kaylangan:
Sounds good. I removed facts tracker and updated the PR title.

gaearon:
Thanks. If we ever want this again, we have version control. :-)

PR #15112, opened 2019-03-14
Adds more scaffolding for experimental event API

trueadm:
This PR is a follow up to https://github.com/facebook/react/pull/15108. This adds the rest of the experimental event API scaffolding logic. Like the previous PR, this is only intended for internal FB testing right now. I also added flags around everything, so bundle size shouldn't regress other than a few bytes (if it does, I've done something wrong with DCE).

I've also added relevant fiber tests to ensure the warnings fire as expected.


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1923,6 +1926,70 @@ function updateContextConsumer(
...
+    workInProgress,
+    nextChildren,
+    renderExpirationTime,
+  );
+  if (__DEV__) {
+    // Check event node has any text nodes as direct nodes and warn
+    let child = workInProgress.child;
+    while (child !== null) {
+      if (child.tag === HostText) {
+        warningWithoutStack(
```
sebmarkbage: This won’t warn if the node is nested in a function component for example. We could use the host context for this instead. We don’t have to do that in this PR but can you add a test for the warning and maybe a commented out test for the nested case? That way it’s easy to do that change later.

sebmarkbage: Why exclude the stack here? Seems useful to have the stack in the warning.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1923,6 +1926,70 @@ function updateContextConsumer(
...
+    'Event target components must be direct child of event components',
+  );
+  // These invariants only occur in DEV to reduce overhead in production
+  if (__DEV__ && nextProps.type === 'touch-hit') {
+    let childrenCount = 0;
+    let child = workInProgress.child;
+    while (child !== null) {
+      if (child.tag === HostText) {
+        childrenCount++;
+        invariant(
```
sebmarkbage: This should be a warning if it’s behind a DEV flag. Let’s also add a test that covers this.

Not only does it let us test the warning but also make sure the rest of the basic stuff here works too.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -766,13 +769,17 @@ function completeWork(
...
+        const rootContainerInstance = getRootHostContainer();
+        const responder = workInProgress.type.responder;
+        handleEventComponent(responder, rootContainerInstance, workInProgress);
+      }
       break;
     }
     case EventTarget: {
-      markUpdate(workInProgress);
+      if (enableEventAPI) {
+        markUpdate(workInProgress);
```
sebmarkbage: This is fine as a start but longer term it would be good if we could avoid doing this in the commit phase for initial renders.

For host components we add events and stuff during render phase during initial mount (`current === null`). We only do stuff in the commit phase for updates.

trueadm: In this case, the event target needs to mutate the direct DOM elements of the event target. This is only known till commit phase as the children might not have been created/attached in complete phase. Well at least that's the issue I ran into when testing this, maybe I'm missing something though.

trueadm:
I've addressed the feedback. I'll follow up on any other concerns in subsequent PRs. :)

packages/react-dom/src/client/validateDOMNesting.js
```diff
@@ -159,10 +162,18 @@ if (__DEV__) {
     dlItemTagAutoclosing: null,
   };
 
-  updatedAncestorInfo = function(oldInfo, tag) {
+  updatedAncestorInfo = function(oldInfo, tag: string | Symbol) {
```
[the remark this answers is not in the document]
trueadm: I originally did that, but it meant updating the host context object to include another property, which felt a bit over-engineered when technically event components are related to the DOM ancestor tree too.

packages/react-dom/src/shared/DOMNamespaces.js
```diff
@@ -29,8 +33,11 @@ export function getIntrinsicNamespace(type: string): string {
 
 export function getChildNamespace(
   parentNamespace: string | null,
-  type: string,
+  type: string | Symbol,
 ): string {
+  if (enableEventAPI && type === REACT_EVENT_COMPONENT_TYPE) {
```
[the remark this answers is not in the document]
trueadm: My same comment applies. Splitting this function up might also make some parts more confusion. I'll give it another pass by though.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1923,6 +1926,63 @@ function updateContextConsumer(
...
+    nextChildren,
+    renderExpirationTime,
+  );
+  const parent = workInProgress.return;
+  invariant(
+    parent !== null && parent.tag === Event,
+    'Event target components must be direct children of event components',
+  );
+  // These warnings only occur in DEV to reduce overhead in production
+  if (__DEV__ && nextProps.type === 'touch-hit') {
```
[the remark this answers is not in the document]
trueadm: I can change it to a symbol. For now there is only one valid type and that is touch hit. I'll wrap the logic in a type.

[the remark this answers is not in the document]
gaearon: I guess I got confused because in another place, what used to always be a string became `string | symbol`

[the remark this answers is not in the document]
trueadm: Updated this and made it a symbol.

packages/react-dom/src/client/validateDOMNesting.js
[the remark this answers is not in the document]
trueadm: Updated and changed it to use its own logic without using `updatedAncestorInfo`.

packages/react-reconciler/src/ReactFiberHostContext.js
```diff
@@ -79,7 +82,12 @@ function pushHostContext(fiber: Fiber): void {
     rootInstanceStackCursor.current,
   );
   const context: HostContext = requiredContext(contextStackCursor.current);
-  const nextContext = getChildHostContext(context, fiber.type, rootInstance);
+  let type = fiber.type;
+  // For React event components, we get the typeof field (the symbol)
```
gaearon: Why?

trueadm: Because we then don't need to expose the fiber work tag logic in the renderer for the fiber.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -142,14 +148,17 @@ export function getRootHostContext(
...
     const ancestorInfo = updatedAncestorInfo(null, validatedTag);
+    if (enableEventAPI) {
+      return {namespace, ancestorInfo, isEventComponent: false};
+    }
     return {namespace, ancestorInfo};
   }
   return namespace;
 }
 
 export function getChildHostContext(
```
sebmarkbage: It would be better to have a `getChildHostContextForEvent` function in the HostConfig. That way there can be a default noop version that just returns the parent context that will be inlined into beginWork.

We also don't have to consider the complexities of events when dealing with just simple host components.

packages/react-reconciler/src/ReactFiberHostContext.js
```diff
@@ -79,7 +82,12 @@ function pushHostContext(fiber: Fiber): void {
...
+  if (enableEventAPI && fiber.tag === EventComponent) {
+    type = type.$$typeof;
+  }
+  const nextContext = getChildHostContext(context, type, rootInstance);
 
   // Don't push this Fiber's context unless it's unique.
   if (context === nextContext) {
```
sebmarkbage: If you create a separate pushHostContextForEvent function then for renderers where we call `getChildHostContextForEvent` and that's an identity function that just returns the same context, then this should be able to be DCE since it will know that the context is always the same.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1923,6 +1929,64 @@ function updateContextConsumer(
...
+    renderExpirationTime,
+  );
+  const parent = workInProgress.return;
+  invariant(
+    parent !== null && parent.tag === Event,
+    'Event target components must be direct children of event components',
+  );
+  const eventTargetType = nextProps.type;
+  // These warnings only occur in DEV to reduce overhead in production
+  if (__DEV__ && eventTargetType === REACT_EVENT_TARGET_TOUCH_HIT) {
```
sebmarkbage: This can't happen at all right now since nothing uses the Symbol. Can we get a test of how you expect this to be used even if it's a light version of some internal implementation details of the plugin?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1923,6 +1929,64 @@ function updateContextConsumer(
...
+    current,
+    workInProgress,
+    nextChildren,
+    renderExpirationTime,
+  );
+  const parent = workInProgress.return;
+  invariant(
+    parent !== null && parent.tag === Event,
+    'Event target components must be direct children of event components',
+  );
```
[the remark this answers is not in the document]
trueadm: Your example would work, the parent needs to be an EventComponent and in the above case, it is.

trueadm:
@sebmarkbage I've added tests and moved all the logic checking into host context. This simplifies a bunch of cases and moves the checking logic into the renderer, which makes far more sense.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -165,6 +180,27 @@ export function getChildHostContext(
...
+    if (event.$$typeof === REACT_EVENT_COMPONENT_TYPE) {
+      return {...parentHostContextDev, isEventComponent: true};
+    } else if (event.$$typeof === REACT_EVENT_TARGET_TYPE) {
+      warning(
+        typeof parentHostContext === 'function' &&
+          parentHostContext !== null &&
+          parentHostContext.$$typeof === REACT_EVENT_COMPONENT_TYPE,
+        'validateDOMNesting: React event targets must be direct children of event components.',
+      );
+      return {...parentHostContextDev, isEventTarget: true, hostNodeCount: 0};
```
sebmarkbage: Let's explicitly list out the props you want to copy here so it's clear what overrides are expected and what is expected to reset.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -165,6 +180,27 @@ export function getChildHostContext(
   return getChildNamespace(parentNamespace, type);
 }
 
+export function getChildHostContextForEvent(
+  parentHostContext: HostContext,
+  event: ReactEventComponent | ReactEventTarget,
+): HostContext {
+  if (__DEV__) {
+    const parentHostContextDev = ((parentHostContext: any): HostContextDev);
+    if (event.$$typeof === REACT_EVENT_COMPONENT_TYPE) {
```
sebmarkbage: It might be nicer to pass a flag or the symbol as an argument and whatever else you need as an argument. That way the various host configs aren't so tied to the implementation details of the VDOM and the structure of this object. E.g. for createInstance we pass `type` and `props` as explicit arguments instead of passing the whole React Element.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -193,6 +229,17 @@ export function createInstance(
   if (__DEV__) {
     // TODO: take namespace into account when validating.
     const hostContextDev = ((hostContext: any): HostContextDev);
+    if (__DEV__ && enableEventAPI) {
+      if (hostContextDev.isEventTarget) {
+        (hostContextDev: any).hostNodeCount++;
```
sebmarkbage: Hm. This won't work. Mutation in here is not safe because we could do multiple passes over the parent or render them out of order in a Suspense case.

It might be better to place this warning in the place where you actually attach the event listeners and the error happens. I'm not sure where you're planning to attach the listeners but if you do it in the commit phase and plan on traversing the Fibers to find the DOM node, then you can do the warning there too. Another possible place is in the complete phase for the EventTarget, because that's when all children are guaranteed to have been completed.

trueadm:
@sebmarkbage I had to make the test `internal` because it writes to the feature flags to enable the events API. Tests that do this must always been internal it seems (according to comments in the config file for the test runner).

packages/react-dom/src/shared/DOMNamespaces.js
```diff
@@ -31,6 +35,9 @@ export function getChildNamespace(
   parentNamespace: string | null,
   type: string,
 ): string {
+  if (enableEventAPI && type === REACT_EVENT_COMPONENT_TYPE) {
```
sebmarkbage: You can revert this whole block because it can't happen anymore. The `type` is a string and `REACT_EVENT_COMPONENT_TYPE` is a symbol.

packages/shared/getElementFromTouchHitTarget.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+import type {Fiber} from 'react-reconciler/src/ReactFiber';
+
+import {HostComponent} from 'shared/ReactWorkTags';
+import warning from 'shared/warning';
+
+export default function getElementFromTouchHitTarget(
+  targetFiber: Fiber,
+): null | Element {
```
sebmarkbage: The return type here should be `Instance` from HostConfig since it's whatever the host config defined it to be. Which happens to be Element for DOM but you're calling it from noop too.

packages/react-reconciler/src/__tests__/ReactFiberEvents-test-internal.js
```diff
@@ -0,0 +1,361 @@
...
+          <div>Hello world</div>
+        </EventComponent>
+      );
+
+      ReactNoop.render(<Test />);
+      expect(Scheduler).toFlushWithoutYielding();
+      expect(ReactNoop).toMatchRenderedOutput(<div>Hello world</div>);
+    });
+
+    it('should warn when an event component has a direct text child', () => {
```
NE-SmallTown: has -> does't have?

Issue #15113, opened 2019-03-14
eslint-plugin-react-hooks useCallback circular dependencies?


KhodorAmmar:
*bug/question*

**What is the current behavior?**

When there's a function that mentions itself wrapped with useCallback or useEffect, eslint-plugin-react-hooks will add the function to the list of dependencies causing no-use-before-define warnings

Example:

```js
const updateAndClean = useCallback(() => {
  console.log("update something")
  document.removeEventListener("event", updateAndClean);
});
```

The idea of the function above is to clean after itself when there's no need for the event anymore. I'm sure there are better ways to do that, but for the sake of the issue let's presume this is valid code.

eslint-plugin-react-hooks would propose adding "updateAndClean" to the list of dependencies:

```js
const updateAndClean = useCallback(() => {
  console.log("update something")
  document.removeEventListener("event", updateAndClean);
}, [updateAndClean]);
```
Causing no-use-before-define warning:
`Line 17:  'updateAndClean' was used before it was defined  no-use-before-define` 

Questions are:
- Is this ok to leave as is? would "using the function before defining it" break any useCallback optimisation? 
- Can eslint-plugin-react-hooks not automatically fix that since it causes other warnings?
- Knowing the function needs to run once, can I pass an empty array to useCallback?

gaearon:
Huh. This does seem silly / a bug.

gaearon:
Do you want to send a fix? I think we should ever consider a function to be its own dependency.

PR #15115, opened 2019-03-15
[ESLint] Don't suggest a function as its own dep

gaearon:
Fixes https://github.com/facebook/react/issues/15113.


gaearon:
I guess there's still a weird case with circular deps where I'm not sure what's the canonical solution is. But maybe let's wait till someone reports it so that we can see the real code people try to write.

PR #15119, opened 2019-03-15
Add feature flag to disable yielding

sebmarkbage:
Useful for debugging.

We might want to have this as an option to createRoot in the future.


gaearon:
(Fails because CircleCI is down and our master builds don't kick in: https://status.circleci.com/)

PR #15120, opened 2019-03-15
[Shallow] Implement setState for Hooks and remount on type change

### LINKED ISSUE #14840: Shallow renderer doesn't update on setState() with Hooks

gaearon:
Fixes #14840.

Alternative to https://github.com/facebook/react/pull/14841. While reviewing it, I realized some of the complexity in how we implemented it is relates to an existing bug in the shallow renderer. In particularly, rendering twice with different element type kept reusing the old class instance. I fixed this in 50f22627ef91772456b36f52786c7cf2f1774db5. (Now it behaves consistently on unmount/mount and on rendering a different element type.) While this has been broken for a long while, the old behavior of ignoring the type change plainly doesn't make sense — so if it breaks any tests, it means those tests were testing the wrong thing.

This allowed me to remove a bunch of fields and simplify the logic to match shallow renderer assumptions. In the last commit (4972adbc71bfc6c8bbe05c3b6da1a09fda4043b5), I'm adding support for state updates to shallow renderer. I think it's a bit simpler than what we tried in #14841.

I checked these changes don't break Enzyme test suite.

### DIFF
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -178,6 +178,10 @@ class ReactShallowRenderer {
   };
 
   constructor() {
+    this._reset();
+  }
+
+  _reset() {
     this._context = null;
     this._element = null;
     this._instance = null;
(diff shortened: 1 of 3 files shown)


pgangwani:
@gaearon  does this include all hooks compatibility for Shallow Render. Can we take latest version to retest failed testcases  in  airbnb/enzyme#2041

gaearon:
Feel free to test it. If you see an issue please file it.

PR #15121, opened 2019-03-15
Change passive checker to use defineProperty

trueadm:
This converts the `passive` check to use `defineProperty` so that this works in older browsers that do not support newer syntax.


trueadm:
Validated it still works as intended.

PR #15122, opened 2019-03-15
Suppress unmounted setState warning after flushing passive effects

gaearon:
Fixes https://github.com/facebook/react/issues/15057. (I verified sandbox doesn't warn anymore.)

If `setState(foo)` caused passive effects to flush, and those in turn caused the component to unmount, there is no way somebody could have foreseen that and guarded `setState` early. Therefore, flushing passive effects temporarily disables the warning until after the next time the work is scheduled.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -181,6 +181,7 @@ const {ReactCurrentDispatcher, ReactCurrentOwner} = ReactSharedInternals;
 
 let didWarnAboutStateTransition;
 let didWarnSetStateChildContext;
+let suppressUpdateOnUnmountedWarning;
 let warnAboutUpdateOnUnmounted;
 let warnAboutInvalidUpdates;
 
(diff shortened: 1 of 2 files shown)


acdlite:
I've been thinking about this fix and I'm worried it exposes a deeper flaw about when we flush passive effects. Currently, we check for pending passive effects inside the `setState` method before we add additional updates to the queue, in case those pending effects also add things to the queue.

However, the `setState` method is too late, because the event that caused the update might not have ever fired had the passive effects flushed before we got there.

This seems similar to the discrete/serial events problem. When a serial update comes in, and there's already a pending serial update, we have to do it *before* we call the user-provided event handlers. Because the event handlers themselves might change as a result of the pending update.

So, I wonder if a better fix would be to remove `flushPassiveEffects` from `setState` and instead flush them in the same place we flush serial events. This function:

https://github.com/facebook/react/blob/d926936f0b5ebe160819717924d7a124c2f65471/packages/react-reconciler/src/ReactFiberScheduler.js#L2568-L2577

 And then all other updates are assumed to be non-serial, so the order in the queue shouldn't matter.

sebmarkbage:
Yea silencing is not the fix here. It surfaces a more fundamental issue that can surface itself with other effects than just setState.

PR #15123, opened 2019-03-16
Add a failing test for potential useReducer bug

gaearon:
https://mobile.twitter.com/pomber/status/1106667842844401670

cc @pomber


gaearon:
The difference I'm seeing between 11->21 and 21->33 is that 11->21 didn't get `markWorkInProgressReceivedUpdate` during the `useReducer` call.

Screen Shot 2019-03-16 at 12 48 11 AM

PR #15124, opened 2019-03-16
Eager bailout optimization should always compare to latest reducer

acdlite:
Based on a bug report by @pomber

https://mobile.twitter.com/pomber/status/1106667842844401670

The issue was that the queue contains a reference to the last rendered reducer, but we weren't updating it in all cases, so sometimes it was stale, triggering the bailout incorrectly.

The original bug report (https://codesandbox.io/s/yj605043yv) had an additional bug in it that made it a bit more confusing, which is it did not specify an initial state. I did not include that part in my test case because it wasn't relevant to the actual underlying bug.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -627,6 +627,8 @@ function updateReducer<S, I, A>(
     'Should have a queue. This is likely a bug in React. Please file an issue.',
   );
 
+  queue.eagerReducer = reducer;
+
   if (numberOfReRenders > 0) {
     // This is a re-render. Apply the new render phase updates to the previous
     // work-in-progress hook.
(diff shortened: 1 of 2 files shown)


gaearon:
I tried fixing it like this, and it also passed tests: https://github.com/facebook/react/commit/6670512082e1c83de535f3b9ddb3425fa4fb8bc5. Is my fix wrong? If yes, can we add a new case that this naïve fix would fail?

acdlite:
@gaearon That's essentially the same fix except in yours doesn't account for the render phase updates path. So I think it would break if an earlier state in the same component gets a render phase update, and then a subsequent inline reducer closes over that. I'll see if I can write a test case where that would fail.

acdlite:
Added additional test that fails if reducer is not updated in the render phase updates branch.

tjallingt:
> The original bug report (https://codesandbox.io/s/yj605043yv) had an additional bug in it that made it a bit more confusing, which is it did not specify an initial state.

This seems like something worth logging a warning for?

PR #15126, opened 2019-03-16
ReactNative's ref.measureLayout now takes a ref

TheSavior:
The number is the reactTag which is found in product code with `findNodeHandle`. In order to get rid of `findNodeHandle` this needs to be able to be called with a ref. This PR now allows measureLayout to be passed a ref which must be a ref to a native node. 

In Paper we will get the nativeTag for this ref, in Fabric, in a follow up, this will pull the node off the ref.canonical and call FabricUIManger with it.

### DIFF
--- a/packages/react-native-renderer/src/__tests__/ReactFabric-test.internal.js
+++ b/packages/react-native-renderer/src/__tests__/ReactFabric-test.internal.js
@@ -316,6 +316,132 @@ describe('ReactFabric', () => {
     });
   });
 
+  it('should support ref in ref.measureLayout', () => {
+    const View = createReactNativeComponentClass('RCTView', () => ({
+      validAttributes: {foo: true},
+      uiViewClassName: 'RCTView',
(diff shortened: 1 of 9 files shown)


TheSavior:
Lets not change NativeMethodsMixin or ReactNative.NativeComponent. Let's not make those more powerful now.

packages/react-native-renderer/src/__tests__/ReactFabric-test.internal.js
```diff
@@ -316,6 +316,61 @@ describe('ReactFabric', () => {
     });
   });
 
+  it('should support ref in ref.measureLayout', () => {
+    const View = createReactNativeComponentClass('RCTView', () => ({
+      validAttributes: {foo: true},
+      uiViewClassName: 'RCTView',
+    }));
+
+    [View].forEach(Component => {
```
bvaughn: 😆

PR #15130, opened 2019-03-17
test(eslint-plugin-react-hooks): add coverage for unused custom hook

RedTn:
Adds coverage for https://coveralls.io/builds/22217376/source?filename=packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L1297


ahtee:
Can you add a quick comment explaining what this does in the code as well to match the rest of the checks

PR #15139, opened 2019-03-18
Add more info to invalid hook call error message

jaredpalmer:
This adds more information to the invalid hook call error message summarizing the 3 most common reasons. I am not sure 100% this change is necessary for the server renderer, but included it there as well for completeness.

> Related Twitter thread which prompted the change: https://twitter.com/ryanflorence/status/1106809967779012609?s=20

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -278,8 +278,12 @@ function warnOnHookMismatchInDev(currentHookName: HookType) {
 function throwInvalidHookError() {
   invariant(
     false,
-    'Hooks can only be called inside the body of a function component. ' +
-      '(https://fb.me/react-invalid-hook-call)',
+    'Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for' +
+      ' one of the following reasons:\n' +
+      '1. You might have mismatching versions of React and React DOM\n' +
+      '2. You might be breaking the Rules of Hooks\n' +
+      '3. You might have more than one copy of React in the same app\n' +
+      'See https://fb.me/react-invalid-hook-call for more information.',
   );
 }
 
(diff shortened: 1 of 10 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -278,8 +278,12 @@ function warnOnHookMismatchInDev(currentHookName: HookType) {
 function throwInvalidHookError() {
   invariant(
     false,
-    'Hooks can only be called inside the body of a function component. ' +
-      '(https://fb.me/react-invalid-hook-call)',
+    'Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for' +
+      ' one of the following reasons:\n' +
+      '1. You might have mismatching versions of React and React DOM\n' +
```
gaearon: Technically this can fire in RN too, so the message should probably say "of React and the renderer (such as React DOM)"

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -278,8 +278,12 @@ function warnOnHookMismatchInDev(currentHookName: HookType) {
...
+      '2. You might be breaking the Rules of Hooks\n' +
+      '3. You might have more than one copy of React in the same app\n' +
+      'See https://fb.me/react-invalid-hook-call for more information.',
```
gaearon: Maybe "for tips about how to debug and fix this problem". To call more attention to the link.

packages/react-reconciler/src/ReactFiberHooks.js
jaredpalmer: I copied from the React docs. Should we also maybe update those as well?

jaredpalmer: Will open a different PR there.

gaearon: In the React docs it's more acceptable because React docs in generally use DOM for examples. But seeing "React DOM" in your native console is weird.

gaearon:
Looks good, thanks.

PR #15141, opened 2019-03-18
Provide new jsx transform target for reactjs/rfcs#107

rickyvetter:
This exposes two new top-level functions on `React` - `jsx`, and `jsxs`. These are targets for JSX transformers which open up many simplifications to element creation described in https://github.com/reactjs/rfcs/pull/107.

The intent is to be backwards-compatible with the existing `createElement` call and to migrate transformers over so that most people are using this new transform. Once there are actual transforms written against this api, the more substantive changes can be tested.

### DIFF
--- a/packages/react/src/React.js
+++ b/packages/react/src/React.js
@@ -22,6 +22,7 @@ import {
   createFactory,
   cloneElement,
   isValidElement,
+  jsx,
 } from './ReactElement';
 import {createContext} from './ReactContext';
 import {lazy} from './ReactLazy';
@@ -43,10 +44,16 @@ import {
   createElementWithValidation,
   createFactoryWithValidation,
   cloneElementWithValidation,
+  jsxWithValidation,
+  jsxWithValidationStatic,
+  jsxWithValidationDynamic,
 } from './ReactElementValidator';
 import ReactSharedInternals from './ReactSharedInternals';
 import {error, warn} from './withComponentStack';
-import {enableStableConcurrentModeAPIs} from 'shared/ReactFeatureFlags';
+import {
+  enableStableConcurrentModeAPIs,
+  enableJSXTransformAPI,
+} from 'shared/ReactFeatureFlags';
 
 const React = {
   Children: {
@@ -105,6 +112,19 @@ const React = {
 if (enableStableConcurrentModeAPIs) {
   React.ConcurrentMode = REACT_CONCURRENT_MODE_TYPE;
   React.unstable_ConcurrentMode = undefined;
+}
+
+if (enableJSXTransformAPI) {
+  if (__DEV__) {
+    React.jsxDEV = jsxWithValidation;
+    React.jsx = jsxWithValidationDynamic;
+    React.jsxs = jsxWithValidationStatic;
+  } else {
+    React.jsx = jsx;
+    // we may want to special case jsxs internally to take advantage of static children.
+    // for now we can ship identical prod functions
+    React.jsxs = jsx;
+  }
 }
 
 export default React;
--- a/packages/react/src/ReactElement.js
+++ b/packages/react/src/ReactElement.js
@@ -95,8 +95,10 @@ function defineRefPropWarningGetter(props, displayName) {
  * if something is a React Element.
  *
  * @param {*} type
+ * @param {*} props
  * @param {*} key
  * @param {string|object} ref
+ * @param {*} owner
  * @param {*} self A *temporary* helper to detect places where `this` is
  * different from the `owner` when React.createElement is called, so that we
  * can warn. We want to get rid of owner and replace string `ref`s with arrow
@@ -104,11 +106,9 @@ function defineRefPropWarningGetter(props, displayName) {
  * change in behavior.
  * @param {*} source An annotation object (added by a transpiler or otherwise)
  * indicating filename, line number, and/or other information.
- * @param {*} owner
- * @param {*} props
  * @internal
  */
-const ReactElement = function(type, key, ref, self, source, owner, props) {
+const ReactElement = function(type, props, key, ref, owner, self, source) {
   const element = {
     // This tag allows us to uniquely identify this as a React Element
     $$typeof: REACT_ELEMENT_TYPE,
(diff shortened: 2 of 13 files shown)


packages/react/src/React.js
```diff
@@ -83,6 +85,10 @@ const React = {
   createElement: __DEV__ ? createElementWithValidation : createElement,
   cloneElement: __DEV__ ? cloneElementWithValidation : cloneElement,
   createFactory: __DEV__ ? createFactoryWithValidation : createFactory,
+  jsx: __DEV__ ? jsxWithValidation.bind(null, false) : jsx,
```
sebmarkbage: Can you add a feature flag in ReactFeatureFlags for this and then only conditionally add these if the flag is on? Like `enableStableConcurrentModeAPIs` is doing below.

It allows us to release from master once it lands.

You might have to find other ReactFeatureFlags forked files and it there too. E.g. you can turn it on in the ReactFeatureFlags.www.js

rickyvetter:
I suspect that tests are failing because CI tests against the bundles and uses `ReactFeatureFlags.readonly.js`. I don't know how to repro this locally and didn't see any special way other feature flags get around this. Recommendations would be appreciated!

gaearon:
You can add `.internal.js` suffix to test files that use feature flags. They won't run on bundles.

packages/react/src/ReactElement.js
```diff
@@ -164,6 +164,91 @@ const ReactElement = function(type, key, ref, self, source, owner, props) {
...
+        defineKeyPropWarningGetter(props, displayName);
+      }
+      if (ref) {
+        defineRefPropWarningGetter(props, displayName);
+      }
+    }
+
+    // these aren't explicitly defined because they are implementation details of dev time transform
+    // shouldn't really be exposed as part of the api. TODO: should they be?
+    if (arguments[3] !== undefined) {
```
sebmarkbage: I think we probably should have a separate DEV only method for these because we want to ensure that it doesn't even work in prod so that nobody accidentally enables this in a prod environment. E.g. jsxDEV(...)

Additionally, we probably want to reserve the last arguments for the "flags" proposal.

packages/react/src/ReactElement.js
```diff
@@ -164,6 +164,91 @@ const ReactElement = function(type, key, ref, self, source, owner, props) {
...
+        hasOwnProperty.call(config, propName) &&
+        !RESERVED_PROPS.hasOwnProperty(propName)
+      ) {
+        props[propName] = config[propName];
+      }
+    }
+  }
+
+  if (key === null && maybeKey !== undefined) {
+    key = maybeKey;
```
sebmarkbage: So there's an interesting case here. In the intermediate step we can't actually support both `<div {...props} key="key" />` and `<div key="key" {...props} />` at the same time here. In case the props has a `key`. They have different semantics but we don't have a way to distinguish them in this function. Therefore I think we should only allow one and have the other compile to createElement and warn in the compiler/lint (or fail to compile).

I believe it makes the most sense that the new function only supports the `<div {...props} key="key" />` form. That way the known static key always win in terms of the React key and it continues to work the same when we update to the new semantics that doesn't allow a key to be spread in.

That means that this block should always override, even if the `key` is non-null here.

packages/react/src/ReactElementValidator.js
```diff
@@ -259,6 +261,105 @@ function validateFragmentProps(fragment) {
...
+        validateChildKeys(children, type);
+      }
+    }
+  }
+
+  if (props.key !== undefined) {
+    warning(
+      false,
+      'React.jsx: passing key as part of props is a deprecated pattern. ' +
+        'Please pass using the explicit third argument to React.jsx instead.',
```
sebmarkbage: Most people will see this warning only in the context of JSX so the message should reflect that. E.g "Spreading a key to JSX is a deprecated pattern. Explicitly pass a key after spreading props in your JSX call. E.g. <ComponentName {...props} key={key} />"

packages/react/src/ReactElementValidator.js
```diff
@@ -259,6 +261,105 @@ function validateFragmentProps(fragment) {
...
+      false,
+      'React.jsx: type is invalid -- expected a string (for ' +
+        'built-in components) or a class/function (for composite ' +
+        'components) but got: %s.%s',
+      typeString,
+      info,
+    );
+  }
+
+  const element = jsx.apply(this, Array.prototype.slice.call(arguments, 1));
```
sebmarkbage: If we do the thing where jsxDEV is its own special thing, then we don't need this to be called with arity. Just call it jsx(type, props, key).

packages/react/src/ReactElement.js
```diff
@@ -164,6 +164,91 @@ const ReactElement = function(type, key, ref, self, source, owner, props) {
...
+
+  let key = null;
+  let ref = null;
+  let self = null;
+  let source = null;
+
+  // TODO: I don't allow null/undefined at dev time
+  // I think that decision makes sense because the goal is to pass props
+  // straight through. Is it React's responsibility to create an empty object
+  // if you pass null/undefined?
```
sebmarkbage: Yea, I think we can assume that the config is always defined since this should not be called directly by users. It might be nice if empty props shared the same empty object for easier bailout but that could be guaranteed by the compiler e.g. by importing a shared object. That way we don't have to branch for it here.

packages/react/src/ReactElement.js
rickyvetter: Great. Makes sense. Should we have the `'' + key` casting behavior too? I think since the key is often a variable the transform will have a hard time verifying that it's a string - so it's not as easy to manage as the non-nullable config/props. I'll add it as well.

packages/react/src/ReactElement.js
rickyvetter: By flags proposal do you mean the static/not static flag? If we do that do the two functions (`jsx`/`jsxs`) still make sense?

`jsxDEV` is a bit tricky to write. But I'll put up changes to show how I'm imagining it.

packages/react/src/ReactElementValidator.js
rickyvetter: I believe ElementValidator only runs in dev, so this would actually call jsxDEV with source and self. Check the newest code - not sure it's the best way, but I think it has the ideal api, where jsxDEV is opt in for source/self but jsx in dev will still give you all of the warnings.

packages/react/src/ReactElement.js
```diff
@@ -164,6 +164,132 @@ const ReactElement = function(type, key, ref, self, source, owner, props) {
...
+  if (type && type.defaultProps) {
+    const defaultProps = type.defaultProps;
+    for (propName in defaultProps) {
+      if (props[propName] === undefined) {
+        props[propName] = defaultProps[propName];
+      }
+    }
+  }
+
+  // TODO: is removing owner safe?
```
sebmarkbage: We need to continue supporting string refs and owners because people need to be able to upgrade the transform without any other code other than bumping a minor of React. E.g. bumping a Babel version should break your React app.

rickyvetter:
There were a couple new feature flag forks added for the new scheduler - I thought maybe missing those had caused this issue, but after adding them it seems that sizebot is still reporting an increase (although only on NODE_PROD - does the absence of a row here mean there's no increase?). Looking at `scripts/rollup/forks.js` it seems like the logic is correct to return false on this feature flag for all non-WWW builds. So yes, it seems like there is an issue with the DCE setup but I'm having trouble building locally (bad Java version - playing around with it) and can't inspect the bundles that CI produced so I'm not sure how to proceed.

sebmarkbage:
Hm. I can't repro locally neither.

packages/react/src/ReactElement.js
```diff
@@ -95,20 +95,20 @@ function defineRefPropWarningGetter(props, displayName) {
...
  * functions, and as long as `this` and owner are the same, there will be no
  * change in behavior.
  * @param {*} source An annotation object (added by a transpiler or otherwise)
  * indicating filename, line number, and/or other information.
- * @param {*} owner
- * @param {*} props
  * @internal
  */
-const ReactElement = function(type, key, ref, self, source, owner, props) {
+const ReactElement = function(type, props, key, ref, owner, self, source) {
```
sebmarkbage: Ok, so I downloaded the artifacts. Somehow, this is not getting inlined anymore. That's why the file increases.

I'm not sure if it's called in more places and maybe it stops trying to inline it before it's able DCE the other paths.

It could also be the argument order. You could try restoring the argument order. It might have an effect on the compiler since the property reads of mutable values are not pure operations.

rickyvetter: Hey looks like reordering fixes it! Wild. Now there is only a size regression on WWW prod which is to be expected since the feature is turned on there.

sebmarkbage:
Neat! Feel free to "Squash and merge".

Issue #13969, opened 2018-10-25
16.6 Context API not working in class component


dericgw:
**Do you want to request a *feature* or report a *bug*?**
Quite possibly a bug (or maybe confusion about the current API)

I am using the new Context API as well as the new `static contextType` in React 16.6. I am passing context down a couple components deep but when I attempt to access the context within the component, the object is empty (~only the default value passed into `createContext` is being displayed~). This is happening in a current feature I am working on at my job, so I cannot display that code, but I did create a Codesandbox with the gist of the problem.

Here is a demonstration of the behavior: https://codesandbox.io/s/r4myz959ro

I would expect to be able to access the current values of the context. This way, if those values change, I would always have the most recent values. Now, maybe this is expected behavior, however, it would be confusing if it is.

React 16.6
ReactDOM 16.6

javascrewpt:
I don't believe you're using the Context API correctly. Here's how I usually implement it:
https://codesandbox.io/s/v0x6kr0py7

dericgw:
@javascrewpt While I do believe that is one way to use the Context API, have you looked at this: https://reactjs.org/docs/context.html#classcontexttype ?

By the looks of that, I should be able to do what I am attempting to do.

javascrewpt:
Thanks, didn't know that. What I did figure was when you export context from index.js and import it in that component, it's an empty object. Then I tried to export a regular object with a few properties and import it the same and it was an empty object again. So I added an additional file and export just the context with the default values and now it seems to work. https://codesandbox.io/s/xjj2nq5r34

dericgw:
@javascrewpt That is a solution, but the approach seems to avoid the original issue. Also, the top-level component could be used to store state and also provide methods that manipulate that state. In your last example, you would need to jump through more hoops to account for that. From what I can tell, this is the whole purpose of the changes to the context API in 16.6:

> We’ve heard feedback that adopting the new render prop API can be difficult in class components. So we’ve add a convenience API to consume a context value from within a class component.

I appreciate your help. I would like to address the root of the issue or gain clarification of the intentions of the API. If the API is not supposed to work like I am attempting to use it, then I would propose (and even help) in clarifying the documentation so that others do not run into the same issue.

javascrewpt:
[...]

PR #15142, opened 2019-03-18
Improve warning for invalid class contextType

gaearon:
This adds extra warning for when `contextType` exists but is `undefined`. This can happen due to module cycles in legacy environments (see https://github.com/facebook/react/issues/13969) and can be very difficult to debug.

SSR and regular version are copy pasted.

I separated the DEV check outside the actual code because unlike in the prod code, we actually *do* want to warn for `undefined`-but-not-truthy code path.

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRendererContext.js
+++ b/packages/react-dom/src/server/ReactPartialRendererContext.js
@@ -10,18 +10,18 @@
 import type {ThreadID} from './ReactThreadIDAllocator';
 import type {ReactContext} from 'shared/ReactTypes';
 
-import {REACT_CONTEXT_TYPE} from 'shared/ReactSymbols';
+import {REACT_CONTEXT_TYPE, REACT_PROVIDER_TYPE} from 'shared/ReactSymbols';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import getComponentName from 'shared/getComponentName';
 import warningWithoutStack from 'shared/warningWithoutStack';
 import checkPropTypes from 'prop-types/checkPropTypes';
 
 let ReactDebugCurrentFrame;
+let didWarnAboutInvalidateContextType;
 if (__DEV__) {
   ReactDebugCurrentFrame = ReactSharedInternals.ReactDebugCurrentFrame;
+  didWarnAboutInvalidateContextType = new Set();
 }
-
-const didWarnAboutInvalidateContextType = {};
 
 export const emptyObject = {};
 if (__DEV__) {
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -513,27 +513,52 @@ function constructClassInstance(
...
+        if (contextType === null) {
+          addendum = ' However, it is set to null.';
+        } else if (contextType === undefined) {
+          addendum =
+            ' However, it is set to undefined. ' +
+            'This can be caused by a typo or by mixing up named and default imports. ' +
+            'This can also happen due to a circular dependency, so ' +
+            'try moving the createContext() call to a separate file.';
+        } else if (typeof contextType !== 'object') {
+          addendum = ' However, it is set to ' + typeof contextType + '.';
```
aweary: I think we can add "a" here to make it a little clearer

```js
addendum = ' However, it is set to a ' + typeof contextType + '.';
```
That way it reports "set to a string" instead of "set to string", which sounds better. Same goes for numbers, symbols, and functions.

packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -513,27 +513,52 @@ function constructClassInstance(
...
+        } else if (contextType.$$typeof === REACT_PROVIDER_TYPE) {
+          addendum = ' Did you accidentally pass the Context.Provider instead?';
+        } else if (contextType._context !== undefined) {
+          // <Context.Consumer>
+          addendum = ' Did you accidentally pass the Context.Consumer instead?';
+        } else {
```
aweary: Add another branch for arrays?

```js
else if (Array.isArray(contextType)) {
  addendum = 'However, it is set to an array.'
}
```

gaearon: Seems very unlikely to happen

aweary: 🤷‍♂️ seems like it could happen if you import the wrong value. I figured that if it happens it would be weird to see.

> However, it is set to an object with keys {0, 1, 2, 3, 4, 5, 6...100000}

packages/react-dom/src/server/ReactPartialRendererContext.js
```diff
@@ -75,26 +75,50 @@ export function processContext(
...
-            name,
-            isContextConsumer ? 'Consumer' : 'Provider',
-          );
+        contextType._context === undefined; // Not a <Context.Consumer>
+
+      if (!isValid && !didWarnAboutInvalidateContextType.has(type)) {
+        didWarnAboutInvalidateContextType.add(type);
+
+        let addendum = '';
+        if (contextType === null) {
```
sebmarkbage: Sometimes it is nice to be able to conditionally set it:

```js
class Foo {
  static contextType = isServerEnvironment ? null : Context;
}
```

This provides no way to do that while also ensuring you stay declarative.

Is `null` really a common mistake? I'm guessing not since it's the only one you don't have a descriptive message for.

gaearon: Nope. Agree let’s omit it.

packages/react-reconciler/src/ReactFiberClassComponent.js
gaearon: Arrays are pretty rare as export values so I don't think it's worth handling.

PR #15143, opened 2019-03-18
Test that we don't suspend when disabling yielding

sebmarkbage:
Just adding a test to confirm this works.


acdlite:
Thanks for adding this because I think this behavior is an implementation accident — it's the desired behavior, but I when I implemented this I didn't consider the case where we don't yield but the work hasn't expired.

PR #15145, opened 2019-03-19
Deprecate module pattern (factory) components

sebmarkbage:
This is part of https://github.com/reactjs/rfcs/pull/107 but it's not the only reason we want to deprecate it.

It requires a fairly complex code path in React that is forked from how class components are initialized. It makes function components slower, adds code bloat to React and is a maintenance burden.

It was really kind of supported from the beginning. It was originally added at a time where createClass and class-like abstractions were more popular. We also hit quirks with compile-to-JS languages when we changed how our class abstractions worked. E.g. we had to make accommodations to support ScalaJS. To avoid future problems, we kept this style of class working since it's really the simplest form of constructor in JavaScript.

However, since then when we ultimately supported both ES classes (must be called with `new`) and arrow function components (cannot be called with `new`), we had to start getting more selective about which components we call with `new` and which we don't.

In practice, I don't think this pattern has been widely used even by compile to JS languages or class-like abstractions.

In any official capacity, its support has only been briefly noted at the bottom of the [React v0.13.0 Beta 1](https://reactjs.org/blog/2015/01/27/react-v0.13.0-beta-1.html#other-languages) post.

It is currently used by at least [Relay Classic](https://github.com/facebook/relay/blob/master/packages/react-relay/buildReactRelayContainer.js#L52-L73) so some libraries will need to upgrade.

A neat trick is that you can almost always just add:

```js
Foo.prototype = React.Component.prototype;
```

This tricks React's class detection into thinking it's a class and we'll call it with `new`, which means that this doesn't work on arrow functions.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactComponentLifeCycle-test.js
+++ b/packages/react-dom/src/__tests__/ReactComponentLifeCycle-test.js
@@ -977,7 +977,16 @@ describe('ReactComponentLifeCycle', () => {
     };
 
     const div = document.createElement('div');
-    ReactDOM.render(<Parent ref={c => c && log.push('ref')} />, div);
+    expect(() =>
+      ReactDOM.render(<Parent ref={c => c && log.push('ref')} />, div),
+    ).toWarnDev(
+      'Warning: The <Parent /> component appears to be a function component that returns a class instance. ' +
+        'Change Parent to a class that extends React.Component instead. ' +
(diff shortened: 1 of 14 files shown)


edvinerikson:
The Relay Classic reference seems to be used in Relay Modern, but that's the only case in OSS that I've seen using this pattern.

sebmarkbage:
I have a diff open internally to fix the Relay caller.

PR #15146, opened 2019-03-19
feat(eslint-plugin-react-hooks): autofix inject deps on useMemo or useCallback

RedTn:
This feat injects dependencies onto useMemo/useCallback if the second argument is not provided and it has a dependency.
I was thinking not to break code that already uses useMemo without the second argument, that the fix shouldn't be applied unless the developer explicitly wants that change, but I'm open to better ideas.
Ideally there would also be a fix option where it just removes the useMemo/useCallback hook if there are no deps required

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -1087,7 +1087,7 @@ const tests = {
       ],
     },
     {
-      // Infer deps onto hooks that need them
+      // Does nothing to deps on useMemo/useCallback if no options specified
       code: `
         function MyComponent(props) {
           let foo = 42;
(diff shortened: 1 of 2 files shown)


gaearon:
Not super sure how I feel about this. Maybe file an issue for discussion first? Ideally we wouldn't want to have many config options with different behaviors.

*However*, a definitely useful fix would be to improve bare function `useCallback` autofix wrapping (which already exists) to infer deps. Can you start with that instead?

RedTn:
@gaearon Can file issue to discuss and work on the latter.

> However, a definitely useful fix would be to improve bare function useCallback autofix wrapping (which already exists) to infer deps. Can you start with that instead?

Sure, will remove the config option and update the PR to autofix always and only on **useCallback**.
Should I keep the existing functionality where if the hook doesn't have any deps it will inject an empty array?

If I misinterpreted feel free to let me know

gaearon:
>Should I keep the existing functionality where if the hook doesn't have any deps it will inject an empty array?

I'd prefer to move that to a separate PR, and focus this one on fixing the existing case for a start. Thanks!

PR #15150, opened 2019-03-19
Adds react-events package for internal testing

trueadm:
This PR adds a `react-events` package that will act as the home for the user-facing APIs for the experimental event API. This package is only intended for internal testing right now. This also adds a bundle to the build script so we can build the `react-events` package to consume internally.

This PR will be a dependency for https://github.com/facebook/react/pull/15112.

### DIFF
--- a/packages/react-events/index.js
+++ b/packages/react-events/index.js
@@ -1,0 +1,16 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+'use strict';
+
+const ReactEvents = require('./src/ReactEvents');
+
+// TODO: decide on the top-level export form.
+// This is hacky but makes it work with both Rollup and Jest.
+module.exports = ReactEvents.default || ReactEvents;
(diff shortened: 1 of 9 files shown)


packages/react-events/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "react-events",
+  "description": "React is a JavaScript library for building user interfaces.",
+  "keywords": [
+    "react"
+  ],
+  "version": "16.8.4",
```
gaearon: Do you want to start this at `0.1.0` instead?

packages/react-events/index.js
```diff
@@ -0,0 +1,16 @@
...
+ * @flow
+ */
+
+'use strict';
+
+const ReactEvents = require('./src/ReactEvents');
+
+// TODO: decide on the top-level export form.
+// This is hacky but makes it work with both Rollup and Jest.
+module.exports = ReactEvents.default || ReactEvents;
```
acdlite: I don't think we've been doing this on our new packages. See: https://github.com/facebook/react/blob/master/packages/react-cache/index.js

gaearon: Ah good catch

trueadm: Indeed! https://github.com/facebook/react/pull/15152

PR #15155, opened 2019-03-19
Removed unnecessary null check

JonasNiestroj:
In my opinion this null check is not necessary. Object.assign handles null and undefined values.

Its not a big change but i think it makes the code a little bit more readable.


aweary:
Removing this check would break the optimization where returning `null` from a state updater signals that no state update has occurred.

Issue #14387, opened 2018-12-04
useEffect(effect, [ref.current]) is prematurely re-running


1000hz:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
It seems `useEffect(effect, [ref.current])` is re-running under certain circumstances without the current ref identity changing.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React:**
https://codesandbox.io/s/qkzl9xjj44

**What is the expected behavior?**
In this example, the effect should only be run once, not re-run (and thus cleaned up) upon clicking the button for the first time. The `isActive` state should be returned to false on mouseup. `CLEAN UP!` should not be logged to the console since the ref hasn't been reassigned.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
16.7.0-alpha.2 (affects all environments)

brookslybrand:
I don't believe this is actually a bug. The problem is, ref is not actually defined when you pass it in, but since you're passing a reference that gets assigned, it actually gets assigned when the effect kicks off.

Where you have `[ref.current]` in your `useEffect` call (line 28), try replacing it with this and and you'll see what I mean:
```
[
  (() => {
    console.log("ref to compare with", ref.current);
    return ref.current;
  })()
]
```

The ref is undefined when you first pass it because the component hasn't actually mounted, and thus it has nothing to reference.

This simple solution is to this is to use [`useLayoutEffect`](https://reactjs.org/docs/hooks-reference.html#uselayouteffect)

> The signature is identical to useEffect, but it fires synchronously after all DOM mutations.

1000hz:
Ah, right. React checks the effect dependencies during render, but defers running the effect until after commit/paint. So it doesn't recheck the deps until the next render, which happens after `setIsActive(true)`.

So while I guess this isn't a bug per se, it still might be something that could be addressed. Maybe after commit, React could re-run the render function to update the dependencies (but not actually re-render) if it's not too expensive and doesn't cause any other unwanted side effects. I did notice `useLayoutEffect` happened to skirt around the issue in this particular example, but I'm more concerned with fixing this unintuitive behavior if possible than fixing the implementation of that hook.

1000hz:
[...]

PR #15157, opened 2019-03-19
Clean up unmounted fiber references for GC

paulshen:
This change cleans up Fiber pointers to help with garbage collection.

This clears Fiber's `stateNode` pointer when committing unmounts. This removes a potential pointer back to host nodes (e.g. DOM) and component instances.

This cleans `nextEffect` pointer when we're done processing the effect chain (either when flushing passive effects or after committing layout effects). Without clearing this, it's possible to retain effect pointers to unmounted nodes.

A continuation of #14218. Detaching `nextEffect` needs to happen after [committing host effects](https://github.com/facebook/react/blob/daeda44d8f5abdb7354742bb69a967302d34d7f9/packages/react-reconciler/src/ReactFiberScheduler.js#L729) because the effect chain is needed for lifecycles and ref callbacks.


bvaughn:
Purely from a DevTools perspective, this change looks okay (since we call `onCommitFiberUnmount` before this clean up happens). Haven't considered beyond that scope yet.

gaearon:
Can you provide an example demonstrating in which scenario this solves a problem? Thanks.

bvaughn:
> This helps remove memory leaks where effect chains point to old nodes

I'm not sure I agree with describing this as a "memory leak" since there are only ever two fibers (current and alternate). Nothing is "leaked", just "retained" for longer than necessary in some cases.

paulshen:
@gaearon I'm having trouble creating an isolated repro but I can share a real world repro.

## Repro steps
1. Log into Discord (https://discordapp.com/app) in your browser.
2. Click this [link](https://discordapp.com/__development/link?s=WNvogz5xJxIH39rY0SopUpAG4LnvzbeoR0Gu9hSVK7s%3D.eyJ0YXJnZXRCdWlsZE92ZXJyaWRlIjp7ImRpc2NvcmRfd2ViIjp7InR5cGUiOiJpZCIsImlkIjoiMTdkNDdhYSJ9fSwicmVsZWFzZUNoYW5uZWwiOiJzdGFibGUiLCJ2YWxpZEZvclVzZXJJZHMiOltdLCJleHBpcmVzQXQiOiJGcmksIDE5IEFwciAyMDE5IDExOjI2OjEzIEdNVCJ9). This puts you in a custom build with `react-dom` NPM 16.8.4.
3. You can confirm custom build by hovering over gear icon in lower right.
image
4. Go into a server ([example](https://www.reactiflux.com/)). Repeatedly switch between two text channels in a server while profiling memory usage. See memory usage go up a lot.

Screenshot 2019-03-19 17 54 53

## Repro with this PR

Repeat with this [custom build](https://discordapp.com/__development/link?s=HckwoKI55tmt%2BoB0yRoSc4FI%2FH2%2FZAzdlL0j8X8ZcGk%3D.eyJ0YXJnZXRCdWlsZE92ZXJyaWRlIjp7ImRpc2NvcmRfd2ViIjp7InR5cGUiOiJpZCIsImlkIjoiYTY2YjdjNCJ9fSwicmVsZWFzZUNoYW5uZWwiOiJzdGFibGUiLCJ2YWxpZEZvclVzZXJJZHMiOltdLCJleHBpcmVzQXQiOiJGcmksIDE5IEFwciAyMDE5IDExOjI2OjUwIEdNVCJ9) which uses `react-dom` from this PR.

See that memory is freed when you switch between channels.

Screenshot 2019-03-19 18 13 17

## Investigation

When you have a Fiber node reference, it links to all the other nodes in that tree as well as old host nodes via `stateNode`.

@bvaughn Regarding "memory leak", I've found instances where mounted nodes have `nextEffect` references to unmounted nodes. This seems possible if a subset of the React tree changes (e.g. one pane of many). Components outside the changed tree are not touched and can continue to hold `nextEffect` pointers to old nodes within the changed tree. In Discord, the sidebar does not update when you change URL routes from one channel to another within the same server. I confirmed this by observing `nextEffect`.

Does this sound plausible? I'd love to get an isolated repro and will keep trying. Thanks for looking!

sebmarkbage:
I think we should probably reset the nextEffect pointer to null during the last traversal over the effect list in the commit phase (most likely passive effects).

This gets more complex in concurrent mode renders can be interrupted tho.

paulshen:
@sebmarkbage The PR, as is, takes the conservative approach of removing only `Deletion` effects from the `nextEffect` chain. It's done in `commitAllLifeCycles`, which already has a side effect of tracking for existence of passive effects. I tweaked the code to get a reference to `firstPassiveEffect` in case we're starting with one or more `Deletion` effects.

paulshen:
@acdlite looks like there's an active `Scheduler` rewrite! It might make this PR moot but I think unwinding `nextEffect` pointers is still relevant in the new version.

matthargett:
@acdlite can you confirm that #15196 will ultimately resolve this issue? it's not obvious to me that it does in current draft form, but it's a draft after all :)

paulshen:
@sebmarkbage @gaearon I've updated this to clear all `nextEffect` pointers after they're no longer needed (either flushing passive effects or after committing layout effects). This seems simpler than removing only `Deletion` nodes.

I've also updated this to handle the new fiber scheduler. Let me know if there's anything I can do to move this forward. Thanks!

matthewwithanm:
@sebmarkbage pointed me to this thread so I thought I'd share this publicly FWIW:

I recently upgrade the WhatsApp web client from React 16.3.1 to 16.8.4 and we got a huge increase in memory consumption: every conversation component we showed was retained forever. We ended up tracing this back to our animation library holding on to a reference to two DOM nodes. These nodes referenced a linked list of Fibers (via `__reactInternalInstance$jf6dr7u9fbf`), each of which held on to a component instance.

While it doesn't exactly mirror what's going on in our app, I managed to create a minimal repro:

https://codepen.io/matthewwithanm/pen/LvmqJP

I get that this isn't technically a memory leak on React's part, but React made a tiny, very bounded leak (2 nodes) into an unbounded one that scaled linearly with the number of screens visited (!)

Also—and this is just for my own curiosity—is there a reason (perf?) the DOM -> React association couldn't be made with a WeakMap instead? I know that this PR would address our problem by keeping the retained fibers to one per node but it still feels a little icky that they're left with references to React at all.

andreieftimie:
We're seeing the same thing as @matthewwithanm points out above, applying this PR on a local fork solved the memory leak issue for us.

raRaRa:
Can you confirm if it helps with the leak here: https://github.com/facebook/react/issues/15583

AFAIK that particular leak is within Google Chrome itself, but it seems to be worse in React.

tariqporter:
This seems to resolve the issue we have in our application

https://github.com/facebook/react/issues/15637

but only with the generated development build

`build/node_modules/react/cjs/react.development.js`
`build/node_modules/react-dom/cjs/react-dom.development.js`

and not with the production build

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -756,6 +756,12 @@ function commitUnmount(current: Fiber): void {
       }
     }
   }
+
+  // Remove reference for GC
+  current.stateNode = null;
+  if (current.alternate != null) {
```
tariqporter: Should this be `if (current.alternate !== null) {`?

RoystonS:
We've just spent the last 6 days tracking down a severe memory leak, which has ended up blocking our product release. We eventually tracked it down to an issue with `componentDidMount` on class components in React.

After some effort, we've managed to boil the leak down to a trivial repro case consisting of a basic Gatsby starter and a single component that repeatedly re-renders either a class component with a `componentDidMount` function or a function component with an equivalent `useEffect`.  The function component works fine, but the class component leaks through a chain of `.nextEffect` links.

Here's the repro case:

https://github.com/jonnycornwell/potential_react_leak
(specific commit: https://github.com/jonnycornwell/potential_react_leak/commit/bb432464b57453acc8b332b138dffb62aaa0157e)

It's (production-)built and deployed at:
https://jonnycornwell.github.io/potential_react_leak/

To reproduce, simply click the 'Bump count' button for the 'leaking example'. It's merely re-rendering a child component (by changing the React `key`). That child component allocates a 10Mb ArrayBuffer and contains a `componentDidMount` lifecycle method. None of those buffers ever get GCed.  Click it lots of times, and you'll see lots of memory leaking, even after explicit GC.  Do the same for the 'non leaking example', and as you might expect, it doesn't leak.

The issue reproduces easily in both Chrome (v75) and Firefox (68.0b9).

@gaearon, in https://github.com/facebook/react/pull/15157#issuecomment-474633410 you asked for a case where this patch would make a difference?   We applied this patch (to the React 16.8.5 codebase (as that's what our product is using) and it fixed our leak.

We've also tried applying it to React 16.8.6, as used by the above Gatsby-based demo, and it fixes the leak there too.

Here's a Chrome memory snapshot comparison of the leak in action, focused on one of the leaked 10Mb ArrayBuffers after several re-renders. The 'Retained Size' column, in particular is interesting as it shows that each new render is holding on to the previous render data, growing by 10Mb each time:

ReactLeak

CharlieHess:
At Slack, we've also had a new release at risk due to ballooning memory concerns. We're on `16.8.3`, and seeing footprints like this:

image

With very repeatable steps. Retainers showed a lot of `FiberNode`s in a chain. After applying this patch, a GC reclaims everything reliably:

image

amadeus:
We (Discord) spent a couple weeks recently digging into this issue more. 

The TL;DR is that we still don't totally know the `cause` of the leak, but we do have a few hunches. We did figure out a way to negate it - but it took a lot more than this PR. You can see our forks changes here: https://github.com/discordapp/react/tree/16.8.6-the-big-memory.

One of our learnings was the changes in this PR did not go far enough. In fact, there are even a few inconsistencies/bugs in this PR because when clearing the `stateNode`s - we didn't account for parts of the switch statement returning early.

One major hunch we have - is that any DOM nodes that might be passed around in props or state, will end up triggering leaks:
https://codesandbox.io/s/compassionate-fermat-3cq5b

There's also a known bug in Chrome surrounding video tags:
https://bugs.chromium.org/p/chromium/issues/detail?id=969049
which only gets compounded by these fiber references
https://github.com/facebook/react/issues/15583

It's also quite common for our app to trigger major UI transitions (going from one server to another basically unmounts most of the app and then mounts a whole new version.  We have some pet theories that perhaps these reference chains are so long that the GC traversals just can't take enough time to fully collect, or just end up too deep for them ever to be resolved.

paulshen:
To follow up @amadeus, we spent a lot of time investigating this and were unable to find the root cause. I still don't know whether our memory leak was caused in userland or something else. 🤷‍♀ 

I opened a PR where we ultimately landed. The PR here was actually not enough for us; we weren't memory-leak free until we cleared the host node pointers as well (_hides_).

https://github.com/discordapp/react/pull/1

If you're desperate like we were, you can try using the build `"react-dom": "discordapp/react#2526c1bf020a3ecb51c62ffeadbf7a2ac86d1550"`. It's very hacky and I'm not proud of it but it did unblock us.

RoystonS:
Ok, so this is a fairly major issue, negatively affecting the releases of a number of companies (and presumably others who haven't done the memory testing), and we have a repro example and a partial fix for it.

What do we need to do to take this further? It would be great to get a real investigation of this by a core React member to see if this is the correct fix, but _a_ fix is urgently needed.  Do we need to raise a separate memory leak issue to highlight this? I'm just slightly concerned that all this is hiding down at the bottom of a PR and not getting attention, but I'm also not keen to create more admin work just for the sake of it.

sebmarkbage:
I recognize this as a problem but we can't play whack-a-mole with the fixes. Especially if there are different fixes that aren't sufficient for everyone. We need to understand the source of the problem.

@RoystonS Thanks for your detailed repro and the heap snapshot illustrating the root. This is very actionable. We next need to dig into why a Gatsby environment is causing this problem and the kind of pattern causing it.

For new people coming onto this thread, I'd like to see something like:

> We had code like this:

```
let node;
function Foo() {
  return <div ref={n => { if( n) node = n; } /></div>;
}
```

> We didn't expected it to be a big deal.

or

>  We used a <video /> tag and while looking into the heap snapshot, we found that it was still around

Ideally with a screenshot of the leak like @RoystonS's great example.

These could be possible explanations, but is this the common pattern? I'm not sure everyone coming onto this thread has the same issue.

I think the issue that @RoystonS is highlighting is interesting because it's showing that a static tree somehow holds onto multiple copies of a single instance. That shouldn't be possible because React 1) shouldn't be creating multiple instances of a single component in the first place 2) shouldn't even create multiple versions of the Fibers.

While this PR might paper over some of the memory leaks, the source issue here seems much worse and deeper and would still be around.

amadeus:
@sebmarkbage I want to make sure this particular link isn’t missed, it shows both a very common use case and a very easy way to repro (dom refs used as props):

https://codesandbox.io/s/compassionate-fermat-3cq5b

sebmarkbage:
@amadeus Yea. That's an interesting one because it ultimately is retained in React props as opposed to global state.

That pattern is one that we've been looking at actively discouraging though. Passing an imperative DOM element is likely to get your into trouble anyway. E.g. it's easy to read from it in the render pass. You also now require your parents to have gotten a ref materialized and cause a rerender which is really bad for perf.

I think the way this pattern tends to work better is by passing a `createRef()` or `useRef()` box down as the prop and to the `<div />`. That way the ref box does get cleaned up on unmount.

In general, patterns that hold onto DOM nodes, is tricky because to clean it up we essentially have to say that "React can't rely on GC for garbage collection." and we have to change everything to do manual deallocation. Effectively creating the worst of both worlds.

I believe it would be plausible to detach our back pointers from DOM nodes in a very specific case where we also clean up refs. That's not guaranteed to work in all cases you have a ref to a DOM node but probably most of them.

The stronger guarantee is probably just to pass the ref box as a prop instead.

MarshallOfSound:
@sebmarkbage Hey I'm also at Slack with @CharlieHess

I'm not sure we're going to be able to isolate our exact leak down too much especially not as minimal as the two existing repros in this thread.  But the facts as I understand them right now are:

* We have a combination of refs and props being leaked on as far as I can see every state transition.  I can share some screenshots of the heapdump if that would help but not the heapdump itself.
  * These refs and props were being retained by fibers not leaked into the global scope like in your first example
  * None of the leaked elements are Video elements, they're mostly divs and spans
* After we started shipping the patches from https://github.com/facebook/react/pull/15157#issuecomment-503413083 our memory usage plummetted by a noticeable amount (we have the metrics to prove it)

If there are specific questions you have I can try to get some answers, our internal repro is very simple but involves our entire app (which is not simple 😆)

matthewwithanm:
> In general, patterns that hold onto DOM nodes, is tricky because to clean it up we essentially have to say that "React can't rely on GC for garbage collection." and we have to change everything to do manual deallocation. Effectively creating the worst of both worlds.

I know this is much easier said than done and that there are tradeoffs, but should retaining memory characteristics of leaks be a goal of the design? User code aside, things like browser bugs and misbehaving extensions might make the environment too unreliable to rely on no nodes being retained. Something like leaking a single node is usually considered not worth worrying about, but this flips that on its head making every leak, no matter how small, a release blocker.

sebmarkbage:
@matthewwithanm Yea, this is similar to security features. They're a second level of defense. However, the second level defense doesn't have to be air tight since it will always be an incomplete solution.

The question to me is what scale of issues the most aggressive form of this actually would fix in practice. I think the actual cases where the full solution is needed is extremely rare.

For example, if there are other closures attached to the DOM nodes such as event listeners and it leaks anyway, it wouldn't fix anything. This is common when you're getting a ref to do manual manipulation on. At the same time, maybe the lesser fix of only detaching it at the same time ref cleanup happens during unmount is sufficient for most of the simple cases that also happen to be sufficient.

That's why it's so important to get a sense for the patterns where this actually happens.

Another dimension I'm thinking about is if maybe we can start getting rid of refs on DOM nodes and instead make this pattern work with nodes created by user space and inserted into React. The new event system also can attach listeners differently which will need to be considered.

We should also be up-front that this PR isn't the "all memory leaks" solution. We've unfortunately had this become the discussion thread for all leaks but there are several different ones here.

packages/react-reconciler/src/ReactFiberCommitWork.js
rapzo: Probably because of this:
```
> const a = {}
undefined
> a.x != null
false
> a.x == null
true
> a.x === null
false
```

It's a nice way to cover `null` and `undefined` scenarios.

RoystonS:
@sebmarkbage it's been 2 weeks since anybody contributed to this thread.  What could we do that would help progress this?

You mentioned that this PR thread has become a discussion thread for a variety of leaks (but it sounds like the proposed fix does have an impact on all or most of them), and so you - quite rightly - want to be able to distinguish them.

What would you like people to do next? Would you like people to post specific feedback here, or would you like people to raise separate issues?

sebmarkbage:
I spawned another issue and tried to gather my thoughts on that one. https://github.com/facebook/react/issues/16087

Regarding the issue demonstrated in https://github.com/jonnycornwell/potential_react_leak, the next steps would be to find out what in Gatsby is actually causing this to leak since just React doesn't seem to trigger it. It's important to understand exactly the code pattern causing this because we don't know if this PR is a complete solution or if it just happened to minimize the effect enough in the test apps to not be noticed.

I can imagine some patterns that are sufficiently covered by this PR though so probably worth getting it in. I wonder if we could make some kind of regression test. Even if it's a DOM fixture.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -756,6 +756,12 @@ function commitUnmount(current: Fiber): void {
       }
     }
   }
+
+  // Remove reference for GC
+  current.stateNode = null;
```
sebmarkbage: We won't want to do this indiscriminately for all types of Fibers. They might not all have this field. We're looking to use Flow disjoint unions to enforce that they're consistently used for each type. They don't all use it in the same way where this makes sense.

We should be more specific and only do this for the type where it's needed.

However, which type is this supposed to clean up? Is it just DOM nodes? Is it class instance references? What about the equivalent pattern for Hooks? It's hard to tell without knowing more about the particular pattern this fix is solving for.

I'm guessing that this is somehow meant to release DOM nodes that might be part of larger subtrees and therefore might have back pointers into Fibers. However, where is the root that hold onto the first Fiber that might point back into those anyway?

packages/react-reconciler/src/ReactFiberScheduler.new.js
```diff
@@ -1404,6 +1404,15 @@ function commitRootImpl(root, expirationTime) {
     rootWithPendingPassiveEffects = root;
     pendingPassiveEffectsExpirationTime = expirationTime;
   } else {
+    // We are done with the effect chain at this point so let's clear the
+    // nextEffect pointers to assist with GC. If we have passive effects, we'll
+    // clear this in flushPassiveEffects.
+    nextEffect = firstEffect;
+    while (nextEffect !== null) {
+      const nextNextEffect = nextEffect.nextEffect;
+      nextEffect.nextEffect = null;
```
sebmarkbage: I think this part of the fix makes sense because otherwise if this node (nextEffect) never gets any updates ever again, it'll now have a reference to a node that might later get deleted. Should be pretty easy to come up with a reduced test case for this.

paulshen: I was able to come up with an isolated repro. https://github.com/paulshen/react-next-effect-leak/blob/master/src/App.js Our belief was correct that `nextEffect` pointers can point to zombie nodes. Here is a running sandbox https://etc33.codesandbox.io/. Examine the `nextEffect` pointer of `<Sidebar>` after clicking "Remount InnerBody" several times. You can see that it remains pointing at the original (and unmounted) `<InnerBody>`.

I tried to create this repro in the past but was bamboozled by `cloneChildFibers` and https://github.com/facebook/react/blob/67e3f3fb6e342f95f00215c84d5d013d7b0e1b33/packages/react-reconciler/src/ReactFiber.js#L418

brad-decker:
Also confirming that this patch resolves the memory leak that we are experiencing in our application. @sebmarkbage the information here and posted in the umbrella issue has lead to us doing a lot to mitigate the impact of the leak, so thank you for the guidance on what can compound the issue. Still, we feel we are chipping away slowly at the problem, and it’s all worth while to do even in the absence of this issue, but with a looming launch date we are anxiously awaiting this fix.

sebmarkbage:
@brad-decker Mind outlining what application pattern actually caused the leak similar to how others have posted a screenshot from the Chrome heap snapshots? Maybe you're still leaking large DOM trees and maybe we can help you fix that too rather than just mitigate the leak.

packages/react-reconciler/src/ReactFiberCommitWork.js
sebmarkbage: This particular example is rooted in a `nextEffect`. With your other fix, that would be reset to null. So would that be sufficient to fix this particular leak?

Maybe we can split out this part to a separate PR and just land the nextEffect part like you suggested? That way we can see if that is sufficient or not.

packages/react-reconciler/src/ReactFiberScheduler.new.js
sebmarkbage: Maybe just add a comment next to it with a link to your repro. We don't need to add any infra just for this. We can figure that out later.

brad-decker:
@sebmarkbage absolutely thank you. 

We use Nextjs and discovered this issue because on mobile when a user flips back and forth between two of our pages, ones that are a real-world use case / hot path of user traffic, the page crashes. The number of dom nodes in memory is increasing over time with each page view.

image

I am very novice at using the heap snapshot to determine what are the culprits of memory retention, but so far we have discovered a lot of the issues due to realizing that we were using hooks inappropriately and closing over refs and state. We fixed a lot of this via the exhaustive deps rule, but i'm sure we are still doing some things in various files that are not helping the situation.

What i've done to diagnose the issue:
1. I chose two pages in our application and removed all of the layout that encapsulates them, to verify it wasn't anything higher than the page level holding onto memory. In Next land that means I removed my _app and _document so that each page is isolated.
2. I added links at the top of the page from one page to the other.
3. Flipping back and forth between pages, taking snapshots, hunting down retainers and figuring out why they are holding onto nodes. 
4. Removed whole sections of the page entirely to see if i could isolate components that were more problematic than others, etc.

I would love some guidance on how to best work through this. I see 'stateNode' and 'nextEffect' as common retainers of memory but i'm not familiar enough with the architecture to understand the significance of these items. 

image

image

image

image

I have a heapshapshot i can send if it would help more than screenshots. Again pretty novice here.

sebmarkbage:
@brad-decker This is actually super helpful because this is showing a new kind of leak that we haven't seen before in this thread. It looks like a closure it closing over an internal React Fiber data structure. That is interesting because you probably aren't getting a handle to it yourself. Most likely it's React closing over its own internals. It only does this one case: The dispatch function returned from useReducer which is also used to implement setState from useState.

My guess is that you're leaking a dispatch or setState function in an event listener that you're attaching to the window. This would be common in things listening to global key press events or click-outside solutions. You might be using a third party tool that does that.

An example that would cause this:

```
let [value, setValue] = useState(null);
useEffect(() => {
  window.addEventListener('keyup', () => {
    setValue(...);
  });
});
```

This is a pretty bad leak that you probably want to fix anyway since you don't want keypresses to potentially cause unexpected side-effects due to this leak.

However, it is super helpful to me because it's a pretty common bug. It's something we can try to add monitoring and logging for, and we can find a fix that is stronger than the one proposed in this PR.

brad-decker:
@sebmarkbage the only success I have had in totally resolving our memory issues is by removing the Link component from next js everywhere. I have had a few false positives in my journey so I'm really not sure if I'm on the right track.

image

https://github.com/zeit/next.js/blob/canary/packages/next/client/link.tsx

matthewwithanm:
> ```js
>   // useMemo to prevent creating this function more than once
>   useEffect(() => {
>     window.addEventListener('resize', handler);
>     return () => {
>       window.removeEventListener('resise', handler);
>     };
>   }, [handler]);
> ```

hey @brad-decker! it looks like "resize" is spelled wrong in the `removeEventListener()` call so it's never actually being removed! (one reason why the EventTarget API is less than 💯)

bvaughn:
> hey @brad-decker! it looks like "resize" is spelled wrong in the removeEventListener() call so it's never actually being removed! (one reason why the EventTarget API is less than 💯)

That would cause a leak to everything in that function's scope right, since `window` would have a reference.

brad-decker:
Well, dang. Thanks for pointing that out! I am kicking myself for that typo :P

brad-decker:
https://github.com/facebook/react/issues/14387 reading through this and trying to understand the implications of accessing a ref in an effect. The ref is a mutable pointer but its part of the component scope, yeah? Do we put refs in the dependency array of an effect? Using my previous example, the one with the typo, if the handler was created in the effect as per the recommendation in the hooks FAQ for dealing with function, wouldn't I need to also include ref as a dependency? Same question, i suppose, with useCallback?

bvaughn:
> Do we put refs in the dependency array of an effect?

No, it's not necessary to specify refs in the dependency array because refs are expected to be mutated by user code (outside of React, without triggering a re-render).

brad-decker:
@sebmarkbage @bvaughn @timneutkens @timer it turns out that we have a compounding problem with Next and React. The code snippet above with the unfortunate typo was a result of me trying to hack to find the memory leak, that code did not exist prior. While it did help me from continuing to go down a rabbit hole that was caused by my own typo in my search for the underlying leak, that was not the root cause.

on line: https://github.com/zeit/next.js/blob/canary/packages/next/client/link.tsx#L101 of next/link an element is set as the key of a map, this element comes from a ref passed to a child component by the Link component. Link basically forwards props to its child so we can add Next magic to 'a' tags or even 'div' tags. The problem seems to be that either next/link itself is closing over a fiber node by doing this, or something in our code that is adjacent to the Links in some way is holding onto the node. 

How I have confirmed this: 
1. I checked out our master branch to make sure I was in the original state of our memory leak.
2. in node_modules I found the link file and deleted the listeners.set call, and nothing else.
3. ran performance recording for 25 seconds, and during twenty of those seconds, I flipped back and forth between two pages.

image

### before

image (1)

### after
image (2)

we still have minor memory leaks in our application, as illustrated by the gradual upwards trend, but it's significantly less prominent now with very clear successful garbage collecting without the listener being set in the map and no code change on our end.

--- 

thoughts?

sebmarkbage:
This line in Link makes me skeptical: https://github.com/zeit/next.js/blob/canary/packages/next/client/link.tsx#L117

componentDidMount fires *after* the ref callbacks of children. The idea is that they need to be wired before they can be referenced in componentDidMount.

https://github.com/zeit/next.js/blob/canary/packages/next/client/link.tsx#L124

So that would imply that the clean up function gets overridden before it can be called.

Another issue is that makes this code a bit fragile is if handleRef gets called multiple times it doesn't clean up the old ones. E.g. if props switches or something. I don't see that happening now but seems like it would be easy to make such as mistake.

timneutkens:
Having @ijjk look into it!

brad-decker:
Update, tested the above application against changes made in https://github.com/zeit/next.js/pull/7943 by @ijjk and it seems to be resolved. We have also started fixing a myriad of "minor" leaks that have an additive effect. All of them are related to things already brought up by @sebmarkbage in this and the umbrella thread. Thanks for all of the assistance, @sebmarkbage @bvaughn @matthewwithanm @timneutkens and @ijjk

packages/react-reconciler/src/ReactFiberScheduler.new.js
paulshen: @sebmarkbage in case it was missed, I opened #16115

linusthe3rd:
Just to note for folks following this thread (as the information can be easy to miss), the author of this PR opened another PR (#16115) to address part of the leak. 

This was merged in last week and looks to help my application's memory usage a lot better.

eltonchan:
> Can you provide an example demonstrating in which scenario this solves a problem? Thanks.

My application memory leak is for this reason

Fabbok1x:
Any progress on this?

eltonchan:
> Can you provide an example demonstrating in which scenario this solves a problem? Thanks.

When will I plan to merge this?

abirmingham:
I am noticing a large disparity between the memory leak in my application when NODE_ENV=development versus NODE_ENV=production.

Here is the development leak:
image

Here is the production leak:
image

As you can see, the leak is ~3x worse in production (

zebralight:
I tried using this fork to resolve some memory leak issues I've been having but I actually ended up getting better results with 16.9 than with this.  For a bit of context, I have an app with react router where switching between two components will lead to an accumulation of stale fiber nodes.

PR #15159, opened 2019-03-20
test(accumulate): add test suite for accumulate function

jberry93:
Should increase [coverage](https://coveralls.io/builds/22279329/source?filename=packages/events/accumulate.js) a tad for the highlighted loc 👍


gaearon:
We try not to add tests like this that use private APIs anymore. Instead, we test the public API. In this case, this implementation is already covered by the actual events tests.

It doesn't hurt that we added this, but for the future, we made a concentrated effort to *get away* from these kinds of tests. Thanks for sending the PR though! I understand it's not very obvious because it doesn't match how many JS projects do testing.

PR #15168, opened 2019-03-20
Add internal logic for listening to event responders

trueadm:

This PR adds logic to listen to events defined in event responders, following on from previous PRs:

- https://github.com/facebook/react/pull/15036
- https://github.com/facebook/react/pull/15112
- https://github.com/facebook/react/pull/15150

This PR is quite difficult to test in isolation without creating unit tests that depend on internals (which I know we're moving away from as a team). In the next PR, I'll add further logic that connects the event firing to appropriate events that make for good public API tests. There isn't much to this PR to review other than a single function where most of the work happens, so hopefully it should be transparent as to what happens.l

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -14,6 +14,7 @@ import warning from 'shared/warning';
 import {canUseDOM} from 'shared/ExecutionEnvironment';
 import warningWithoutStack from 'shared/warningWithoutStack';
 import type {ReactEventResponder} from 'shared/ReactTypes';
+import type {DOMTopLevelEventType} from 'events/TopLevelEventTypes';
 
 import {
   getValueForAttribute,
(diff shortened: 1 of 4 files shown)


trueadm:
Looks like I’ll need to wrap the function in the event api flag to DCE properly. I’ll akso fix the Flow error too, when I get back later.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1267,3 +1276,65 @@ export function restoreControlledState(
...
   }
 }
+
+export function listenToEventResponderEvents(
+  eventResponder: ReactEventResponder,
+  element: Element | Document,
+): void {
+  if (enableEventAPI) {
+    const {targetEventTypes} = eventResponder;
+    // Get the listening set for this element. We use this to track
```
threepointone: just a nit, but I think we should standardise on using capital S Set, else it's easy to get wrong (especially for noobs like me)
```suggestion
    // Get the listening Set for this element. We use this to track
```

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1267,3 +1276,65 @@ export function restoreControlledState(
...
+    const listeningSet = getListeningSetForElement(element);
+
+    // Go through each target event type of the event responder
+    for (let i = 0, length = targetEventTypes.length; i < length; ++i) {
+      const targetEventType = targetEventTypes[i];
+      let topLevelType;
+      let capture = false;
+      let passive = true;
+
+      // By default, if no event config object is provided (only a string),
```
threepointone:
```suggestion
      // If no event config object is provided (i.e. - only a string),
```

PR #15169, opened 2019-03-20
Fork performWork instead of using boolean flag

sebmarkbage:
I inline it into performAsyncWork instead.

Code that was only relevant to the async callback had leaked into the
performWork call which is an indication that this was a bad abstraction
and therefore the wrong place to DRY.

By inlining I also discovered that minExpirationTime is actually irrelevant
in the yieldy case so we can clean that up.


acdlite:
Test?

sebmarkbage:
I don't have a reduced test case atm.

PR #15170, opened 2019-03-20
React.warn() and React.error()

bvaughn:
Library authors often [request](https://twitter.com/ryanflorence/status/1108022074730110976) a way to log warnings that include the React "component stack". This PR adds two new top-level APIs to for this purpose: `React.warn` and `React.error`

These methods avoid exposing the stack directly so that people aren't tempted to parse it. (They can obviously still do this by intercepting calls to `console.error` but at least the API doesn't encourage it.) Instead they wrap `console.error` and `console.warn` and append the component stack when available.

### Usage

Example usage ([borrowed from react-window](https://github.com/bvaughn/react-window/blob/c53622fd48f82eac12421b9f4013ef73f5b28b0d/src/createListComponent.js#L578-L599)):
```js
import { warn } from "react";

function List(props) {
  if (process.env.NODE_ENV !== "production") {
    if (props.innerTagName != null || props.outerTagName != null) {
      warn(
        "The innerTagName and outerTagName props have been deprecated. " +
          "Please use the innerElementType and outerElementType props instead."
      );
    }
  }

  // ...
}
```

Which might log something like:
```
The innerTagName and outerTagName props have been deprecated. Please use the innerElementType and outerElementType props instead. 
    in List (at App.js:14)
    in div (at App.js:13)
    in App (at src/index.js:7)
```

### Behavior

The behavior of these new methods is as follows:

Build | phase | calls `console` method? | appends component stack?
----- | ----- | ----- | -----
development | render, commit | ✓ | ✓
development | other | ✓ | ✗
production | any | ✗ | ✗

### DIFF
--- a/packages/react/src/withComponentStack.js
+++ b/packages/react/src/withComponentStack.js
@@ -17,11 +17,12 @@ if (__DEV__) {
   error = function() {
     const stack = ReactDebugCurrentFrame.getStackAddendum();
     if (stack !== '') {
-      const args = [];
-      for (let i = 0; i < arguments.length; i++) {
+      const length = arguments.length;
+      const args = new Array(length + 1);
+      for (let i = 0; i < length; i++) {
         args[i] = arguments[i];
       }
-      args.push(stack);
+      args[length] = stack;
       console.error.apply(console, args);
     } else {
       console.error.apply(console, arguments);
@@ -31,11 +32,12 @@ if (__DEV__) {
   warn = function() {
     const stack = ReactDebugCurrentFrame.getStackAddendum();
     if (stack !== '') {
-      const args = [];
-      for (let i = 0; i < arguments.length; i++) {
+      const length = arguments.length;
+      const args = new Array(length + 1);
+      for (let i = 0; i < length; i++) {
         args[i] = arguments[i];
       }
-      args.push(stack);
+      args[length] = stack;
       console.warn.apply(console, args);
     } else {
       console.warn.apply(console, arguments);
--- a/packages/react/src/withComponentStack.js
+++ b/packages/react/src/withComponentStack.js
@@ -7,10 +7,14 @@
 
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 
-let warnWithComponentStack = (...args) => console.error(...args);
+function noop() {}
+
+let error = noop;
+let warn = noop;
 if (__DEV__) {
-  warnWithComponentStack = (...args) => {
-    const ReactDebugCurrentFrame = ReactSharedInternals.ReactDebugCurrentFrame;
+  const ReactDebugCurrentFrame = ReactSharedInternals.ReactDebugCurrentFrame;
+
+  error = (...args) => {
     const stack = ReactDebugCurrentFrame.getStackAddendum();
     if (stack !== '') {
       console.error(...args, stack);
@@ -18,6 +22,15 @@ if (__DEV__) {
       console.error(...args);
     }
   };
+
+  warn = (...args) => {
+    const stack = ReactDebugCurrentFrame.getStackAddendum();
+    if (stack !== '') {
+      console.warn(...args, stack);
+    } else {
+      console.warn(...args);
+    }
+  };
 }
 
-export default warnWithComponentStack;
+export {error, warn};
--- a/packages/react/src/withComponentStack.js
+++ b/packages/react/src/withComponentStack.js
@@ -15,28 +15,30 @@ if (__DEV__) {
   const ReactDebugCurrentFrame = ReactSharedInternals.ReactDebugCurrentFrame;
 
   error = function() {
-    const args = [];
-    for (let i = 0; i < arguments.length; i++) {
-      args[i] = arguments[i];
-    }
     const stack = ReactDebugCurrentFrame.getStackAddendum();
     if (stack !== '') {
-      console.error.apply(console, args.concat(stack));
-    } else {
+      const args = [];
+      for (let i = 0; i < arguments.length; i++) {
+        args[i] = arguments[i];
+      }
+      args.push(stack);
       console.error.apply(console, args);
+    } else {
+      console.error.apply(console, arguments);
     }
   };
 
   warn = function() {
-    const args = [];
-    for (let i = 0; i < arguments.length; i++) {
-      args[i] = arguments[i];
-    }
     const stack = ReactDebugCurrentFrame.getStackAddendum();
     if (stack !== '') {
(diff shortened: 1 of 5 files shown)


bvaughn:
Follow up:
* `React.warn` and `React.error` (for `console.warn` and `console.error`)
* Pass-throughs in DEV outside of render/commit phases.
* No-ops in production.

bvaughn:
Okay, PR has been updated~

sophiebits:
Why make it a no-op in production, vs logging the message sans stack?

bvaughn:
> Why make it a no-op in production, vs logging the message sans stack?

To gently encourage people to use this API for DEV-only messaging.

eps1lon:
This makes library life so much easier :pray: 

Could we change the signature to be equal to existing APIs such as [`warning`](https://www.npmjs.com/package/warning) or [`invariant`](https://www.npmjs.com/package/invariant)? 

In it's current form `React.warn` requires two levels of nesting to not appear in production. Would be nice if we could just™
```diff
- warning(someInvariant, message)
+ React.warn(someInvariant, message)
```

gaearon:
Tbh our experience with `warning` / `invariant` APIs is people find them confusing (easy to mess up the condition, and you can't do args lazily without a build transform. We might be eventually migrating away from them in React as well.

bvaughn:
> Could we change the signature to be equal to existing APIs such as `warning` or `invariant`?

I've personally always found those signatures confusing. I think these methods being 1:1 with their `console` counterparts is going to be easier for open source users. More familiar.

bvaughn:
> In it's current form React.warn requires two levels of nesting to not appear in production. 

Also to be clear, the idea is for these to be DEV-only warnings. So you would want some form of `process.env.NODE_ENV` check so they are removed entirely from production builds.

eps1lon:
> I've personally always found those signatures confusing.

100% agreed here. I'm never comfortable once I have to negate boolean logic. To many moving parts. 

> More familiar.

I would disagree here looking at the download stats for `warning` and the number of dependents.

> Also to be clear, the idea is for these to be DEV-only warnings. So you would want some form of process.env.NODE_ENV check so they are removed entirely from production builds.

That's where `babel-plugin-transform-dev-warning` comes into play. This would already work by switching
```diff
- import warning from 'warning';
+ import { warn as warning } from 'react';
```
if the signatures were the same.

But I understand if you want to use the opportunity to create a better API. We can still wrap them with a custom function that has the same API as `warning`
```js
function warning(condition, ...reactWarnArgs) {
  if (!condition) {
    React.warn(...reactWarnArgs)
  }
}
```

gaearon:
>I would disagree here looking at the download stats for warning and the number of dependents.

A lot of this was just cargo cult from early on. I and other people saw React using `warning`, it seemed "smart" and we copied it, and others spread the pattern. Especially because at the time you would "deduplicate" the logic this way. (You can't anymore since flat bundles.) I don't think this pattern is actually that useful or anything.

bvaughn:
I think wrapping (if you want to mimic the `warning` API) is a good solution 👍 I think the built-in API mirroring e.g. `console.warn` is going to be easiest to understand for the most people.

packages/react/src/withComponentStack.js
```diff
@@ -0,0 +1,36 @@
...
+import ReactSharedInternals from 'shared/ReactSharedInternals';
+
+function noop() {}
+
+let error = noop;
+let warn = noop;
+if (__DEV__) {
+  const ReactDebugCurrentFrame = ReactSharedInternals.ReactDebugCurrentFrame;
+
+  error = (...args) => {
```
sebmarkbage: Let's write out the ES5 code. I don't trust the Babel plugin to do a better job than we do.

oliviertassinari:
> That's where babel-plugin-transform-dev-warning comes into play.

@eps1lon Maybe we should [deprecate the library]( https://github.com/oliviertassinari/babel-plugin-transform-dev-warning/issues/7)? If React is making the if branch explicit, adding an extra `process.env.NODE_ENV !== "production"` should be simple. The pull request description's issue could be written like this:

```jsx
import { warn } from "react";

function List(props) {
  if ((props.innerTagName != null || props.outerTagName != null) && process.env.NODE_ENV !== "production") {
    warn(
      "The innerTagName and outerTagName props have been deprecated. " +
        "Please use the innerElementType and outerElementType props instead."
    );
  }

  // ...
}
```

bvaughn:
> The pull request description's issue could be written like this:

Wouldn't the result end up being something more like this?

```js
import { warn } from "react";

function List(props) {
  // This check wouldn't be removed, so it would bloat the bundle.
  if (props.innerTagName != null || props.outerTagName != null) {
    if (process.env.NODE_ENV !== "production") {
      warn(
        "The innerTagName and outerTagName props have been deprecated. " +
          "Please use the innerElementType and outerElementType props instead."
      );
    }
  }

  // ...
}
```

Andarist:
Env check should always be the first thing in the if statement - because as @bvaughn has noticed other conditions might not get removed properly if you put the env check non-first. In this case minifier wont assume in most cases that getters, such as props.innerTagName here, are side-effect free.

diegohaz:
Will this print the error/warn multiple times if the component is re-rendered?

Andarist:
@diegohaz looking at the implementation - yes, it's just a wrapper around console.warn/error:
https://github.com/facebook/react/pull/15170/files#diff-5ddbfb03321128deb9f08b4daf40fe11R17

bvaughn:
> Will this print the error/warn multiple times if the component is re-rendered?

Yes. Up to you to do your own de-duping. I suggest [using a `WeapSet`](https://github.com/bvaughn/react-window/blob/89287838dc7960e4838081786e0a4423f6607029/src/createListComponent.js#L110-L119) or something similar for this.

mrmckeb:
Would this be slated for 16.9? Looks great!

gaearon:
Yeah planned for 16.9

sahrens:
Awesome! I'm very excited to finally have something here. Did you consider including the "welcome to debugging in react" throw/catch for `error`? What about exposing a utility for getting the stack addendum itself?

bvaughn:
> Did you consider including the "welcome to debugging in react" throw/catch for `error`?

No, I don't think was considered, since we're just calling `console.error` rather than actually throwing something.

> What about exposing a utility for getting the stack addendum itself?

This was considered briefly and TBH I don't remember the discussion super clearly. I think we wanted to avoid an API that might encourage people parsing the stack and building things on top of it. (You can obviously still intercept `console` method calls to get the stack _if you really want to_.)

PR #15172, opened 2019-03-20
Mark Profiler API as stable

bvaughn:
The `Profiler` RFC (https://github.com/reactjs/rfcs/pull/51) is 10 months old and API hasn't changed in months. It seems like that we'll want to add a new API for `Suspense` metrics, but as a new prop that should not impact the existing `onRender` prop.

Maybe it's time to mark this API as stable? Opening for discussion purposes.

(If this lands, I will update and close out the RFC.)


bvaughn:
:heart:

markerikson:
Per my comment on Twitter the other day, it would be nice to have metrics for commit times too (especially if sync `setState` updates are causing those to extend).

kevinSuttle:
Awesome! Did this land in 16.8.5?

bvaughn:
No, 16.9 😄

kevinSuttle:
Oh, nice. Good to know. Thanks!

PR #15173, opened 2019-03-20
Performance improvement: ReactDOMServer `resolve` class method

overlookmotel:
At present the server-side renderer has a hot path through a function called `resolve()` which contains a lengthy closure `processChild()`.

This PR converts `resolve()` and `processChild()` to prototype methods of `ReactDOMServerRenderer` class, to remove this closure.

Using a [fairly naive benchmark](https://github.com/overlookmotel/reactdom-server-bench), I am seeing the following performance improvements:

* Node 11: ~4%
* Node 10: ~4%
* Node 8: ~4%
* Node 6: ~15%

If there is an official benchmark I can use to test this on, please let me know.

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRenderer.js
+++ b/packages/react-dom/src/server/ReactPartialRenderer.js
@@ -397,278 +397,6 @@ function validateRenderResult(child, type) {
   }
 }
 
-function resolve(
-  child: mixed,
-  context: Object,
-  threadID: ThreadID,
-): {|
-  child: mixed,
-  context: Object,
-|} {
-  while (React.isValidElement(child)) {
-    // Safe because we just checked it's an element.
-    let element: ReactElement = (child: any);
-    let Component = element.type;
-    if (__DEV__) {
-      pushElementToDebugStack(element);
-    }
-    if (typeof Component !== 'function') {
-      break;
-    }
-    processChild(element, Component);
-  }
-
-  // Extra closure so queue and replace can be captured properly
-  function processChild(element, Component) {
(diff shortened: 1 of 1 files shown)


packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -1200,6 +928,286 @@ class ReactDOMServerRenderer {
...
+      if (__DEV__) {
+        pushElementToDebugStack(element);
+      }
+      if (typeof Component !== 'function') {
+        break;
+      }
+      ({child, context} = this.resolveNext(element, context, Component));
+    }
+
+    return {child, context};
```
gaearon: Can we just return the thing returned from `resolveNext` here? Since we already have it.

jacobp100:
Could this not be made simpler by just un-nesting `processChild` and passing in `threadID` as an argument? Then we could avoid the prototype lookup and the closure.

gaearon:
Yeah I'd like to avoid dynamic dispatch if possible.

overlookmotel:
Is there any official (or more official than mine) benchmark I can run to determine what's fastest?

overlookmotel:
Please do not close, Mr Robot.

gaearon:
>Is there any official (or more official than mine) benchmark I can run to determine what's fastest?

Not really. We have https://github.com/facebook/react/tree/master/scripts/bench but it's pretty outdated and doesn't server render.

gaearon:
Could you try this please:

>Could this be made simpler by just un-nesting processChild and passing in threadID as an argument? Then we could avoid the prototype lookup and the closure.

?

PR #15179, opened 2019-03-21
Add part of the event responder system for experimental event API

trueadm:

This PR adds logic to handle triggering event responders from events that are listened to at a top level. This follows on from previous PRs:

- https://github.com/facebook/react/pull/15036
- https://github.com/facebook/react/pull/15112
- https://github.com/facebook/react/pull/15150
- https://github.com/facebook/react/pull/15168

This PR also adds tests that validate the triggering of event responder `handleEvent` function.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -44,7 +44,6 @@ import {
 import dangerousStyleValue from '../shared/dangerousStyleValue';
 
 import type {DOMContainer} from './ReactDOM';
-import {currentEventComponentFibers} from '../events/DOMEventResponderSystem';
 import type {ReactEventResponder} from 'shared/ReactTypes';
 import {
   REACT_EVENT_COMPONENT_TYPE,
@@ -52,7 +51,6 @@ import {
   REACT_EVENT_TARGET_TOUCH_HIT,
 } from 'shared/ReactSymbols';
 import getElementFromTouchHitTarget from 'shared/getElementFromTouchHitTarget';
-import type {Fiber} from 'react-reconciler/src/ReactFiber';
 
 export type Type = string;
 export type Props = {
@@ -865,16 +863,6 @@ export function handleEventComponent(
   internalInstanceHandle: Object,
 ): void {
   if (enableEventAPI) {
-    // We keep the event responder hub up-to-do with the state of
-    // event component fiber tree (it uses this to know what events to fire).
-    // To do this, we update the current event component fiber and remove
-    // the alternate fiber if it exists.
-    const eventComponetFiber = ((internalInstanceHandle: any): Fiber);
-    currentEventComponentFibers.add(eventComponetFiber);
-    const alternateFiber = eventComponetFiber.alternate;
-    if (alternateFiber !== null) {
-      currentEventComponentFibers.delete(alternateFiber);
-    }
     const rootElement = rootContainerInstance.ownerDocument;
     listenToEventResponderEvents(eventResponder, rootElement);
   }
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -864,18 +864,20 @@ export function handleEventComponent(
   rootContainerInstance: Container,
   internalInstanceHandle: Object,
 ): void {
-  // We keep the event responder hub up-to-do with the state of
-  // event component fiber tree (it uses this to know what events to fire).
-  // To do this, we update the current event component fiber and remove
-  // the alternate fiber if it exists.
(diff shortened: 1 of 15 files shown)


threepointone:
^ tests breaking, known issue?

trueadm:
Yep. I’ll fix when I get home :)

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -862,6 +864,16 @@ export function handleEventComponent(
   rootContainerInstance: Container,
   internalInstanceHandle: Object,
 ): void {
+  // We keep the event responder hub up-to-do with the state of
+  // event component fiber tree (it uses this to know what events to fire).
+  // To do this, we update the current event component fiber and remove
+  // the alternate fiber if it exists.
+  const eventComponetFiber = ((internalInstanceHandle: any): Fiber);
```
gaearon: Wrap this in flag pls?

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -0,0 +1,114 @@
...
+  eventSystemFlags: EventSystemFlags,
+): void {
+  // TODO add proper event context
+  let context = ({}: any);
+  let node = targetFiber;
+  // Traverse up the fiber tree till we find event component fibers.
+  while (node !== null) {
+    if (node.tag === EventComponent) {
+      if (node.alternate !== null && !currentEventComponentFibers.has(node)) {
+        node = node.alternate;
```
gaearon: What's the logic here? How do you choose between current and alternate and why? Can you write the high level algorithm as a comment in more detail?

trueadm: I've updated the code and added a very detailed explanation.

trueadm:
I've added more test coverage and added comments around some of the logic involved in handling the active event component fibers.

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -0,0 +1,241 @@
...
+  this.eventType = topLevelType;
+  this.eventTarget = nativeEventTarget;
+  this._flags = eventSystemFlags;
+  this._fiber = null;
+  this._responder = null;
+  this._discreteEvents = [];
+  this._nonDiscreteEvents = [];
+}
+
+DOMEventResponderContext.prototype.isPassive = function(): boolean {
```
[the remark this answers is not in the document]
trueadm: `context` is passed to the event responder module so that the user can call `isPassive` etc without having to know about the React internals. I thought about exposing these functions via a public API on React, but that felt like over-kill, especially as the responder modules are external to React.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -862,29 +864,43 @@ export function handleEventComponent(
...
 ): void {
-  const rootElement = rootContainerInstance.ownerDocument;
-  listenToEventResponderEvents(eventResponder, rootElement);
+  if (enableEventAPI) {
+    // We keep the event responder hub up-to-do with the state of
+    // event component fiber tree (it uses this to know what events to fire).
+    // To do this, we update the current event component fiber and remove
+    // the alternate fiber if it exists.
+    const eventComponetFiber = ((internalInstanceHandle: any): Fiber);
+    currentEventComponentFibers.add(eventComponetFiber);
```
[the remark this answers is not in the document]
trueadm: Good point, I'll update this tomorrow.

packages/react-dom/src/events/DOMEventResponderSystem.js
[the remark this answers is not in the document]
sebmarkbage: Cool. I don't fully see how everything will fit together. We can come back to this later once more stuff is in place.

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -0,0 +1,197 @@
...
+  if (validEventTypesForResponder === undefined) {
+    validEventTypesForResponder = createValidEventTypeSet(
+      responder.targetEventTypes,
+    );
+    eventResponderValidEventTypes.set(responder, validEventTypesForResponder);
+  }
+  if (!validEventTypesForResponder.has(topLevelType)) {
+    return;
+  }
+  let state = stateMap.get(responder);
```
sebmarkbage: The responder can't change because the ReactEventComponent (`fiber.type` here) is effectively frozen as I understand it.

The stateMap lives on the Fiber.

So there is a 1:1 relationship between the map and the responder.

So there can only ever be one value in the stateMap.

Can we just make it a lazily initialized field instead of a Map?

trueadm:
I spoke to @sebmarkbage and he was happy with getting this PR in so I'm merging.

Issue #14635, opened 2019-01-19
UMD build doesn't work via import in the browser (TypeError: Cannot set property 'React' of undefined)


vp2177:
When `import`ing the UMD build (both production and development) in the browser – that is, from a `<script type="module">` – an error is thrown: `TypeError: Cannot set property 'React' of undefined`.

The relevant code:

```js
12: (function (global, factory) {
13:	typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
14:	typeof define === 'function' && define.amd ? define(factory) :
15:	(global.React = factory());
16: }(this, (function () { 'use strict';
```

Throws on line 15 as `global` ≣ `this` is `undefined` in a module context.  
Is this as simple to fix as passing `this || self` on line 16, or is there a catch?

**Do you want to request a *feature* or report a *bug*?** 
*bug*

**What is the current behavior?**
`TypeError: Cannot set property 'React' of undefined`

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**
<https://jsfiddle.net/dja3wfc7/> (**_Updated_**)

**What is the expected behavior?**
I would expect the `import` to work, as the same file works if included via `<script src=...>`.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
React 16.7.0 (latest stable) – I have found another UMD related bug, #3037, which pertains to React **0.12.2**. I have tried that version and indeed it works, it can be `import`ed.
Chrome 71, Safari 12 tested (probably all browsers)

gaearon:
It's not expected that UMD imports would work with `import` — you want ESM builds (which we don't offer yet).

vp2177:
@gaearon I have looked into why the UMD build doesn't work, as I thought it should. (Of course an ESM build would be better, but in the meantime...)

It is because rollup `^0.52.1` is used for the build. I haven't tracked down when this was fixed in rollup but the current version (`1.1.2`) produces a correct UMD, which works in a module context.

Is it feasible to upgrade the rollup dependency?

gaearon:
I don't remember a particular reason. If you send a PR upgrading Rollup we can look at it!

kunukn:
> I don't remember a particular reason. If you send a PR upgrading Rollup we can look at it!

https://github.com/facebook/react/pull/15037

Issue #2575, opened 2014-11-20
Issues with ES6 classes


ceane:
I'm having issues with ES6 classes in React. It seems that esprima-harmony differs from how 6to5 is implementing classes, as well as how classes are implemented in beta versions of browsers.

With esprima-harmony I am able to use classes (using `React.createClass(MyClass.prototype)`), however,

**If I use any of these methods:**

**with 6to5**

``` javascript
React.createClass(MyComponent.prototype);
```

**In Chrome 39 and IE11 Windows 10 Preview**

``` javascript
React.createClass(MyComponent);
```

I get this error: `Error: Invariant Violation: createClass(...): Class specification must implement a`render`method.`. And if I try using `React.createFactory` in 6to5 I get this error: `TypeError: undefined is not an object (evaluating 'nextElement.props')`.  

I can't figure out what would be so different between Object.defineProperty and directly defining prototypes, other than the fact that properties set on the prototype aren't automatically enumerable.

**ES6 code**

``` javascript

class MyComponent {
  render() {
    return null;
  }
}
```

**esprima-harmony compiles:**

``` javascript
function MyComponent() {}
MyComponent.prototype.render = function() {return null;}
```

**6to5 compiles:**

``` javascript
var _classProps = function(child, staticProps, instanceProps) {
  if (staticProps) Object.defineProperties(child, staticProps);
  if (instanceProps) Object.defineProperties(child.prototype, instanceProps);
};

var MyComponent = (function() {
  var MyComponent = 
     function MyComponent() {};

  _classProps(MyComponent, null, {
    render: {
      writable: true,

      value: function() { 
        return null;
      }
    }
  });

  return MyComponent;
})();
```

How should I resolve this issue?

mvpspl619:
Should this be re-opened because of the latest announcement given below ?

Unfortunately/fortunately, 6to5 already made necessary changes to make them not enumerable. 

React version being used: 0.12.2.

```
class _Header {

	renderItems () {
		return (
				<ul>
					<li><Link className="btn btn-primary" to="newidea">New</Link></li>
					<li><Link className="btn btn-primary" to="ideas">Ideas</Link></li>
				</ul>
		);
	}

	render () {
		return (
			<div>
				<h1 className="dashboard-header">Idea Store</h1>
				<br />
				{this.renderItems()}
			</div>
		)
	}
};

export default {Header : React.createClass(_Header.prototype)};
```

https://twitter.com/webreflection/status/560929583974088704

[...]

Issue #2293, opened 2014-10-06
React mutates event objects


cjohansen:
I have some code that accesses event.target asynchronously, and to my surprise React is mutating the event object, removing useful properties synchronously after emitting the event. This seems like broken behavior to me:

``` html
<!DOCTYPE html>
<html>
  <head>
    <title>React mutating event objects</title>
  </head>
  <body>
    <h1>React mutating event objects</h1>
    <div id="ui"></div>
    <script src="http://fb.me/react-0.11.2.js"></script>
    <script>
      var d = React.DOM;

      var Clicker = React.createClass({
          render: function () {
              return d.div({},
                           d.a({
                               href: "#",
                               onClick: function (e) {
                                   e.preventDefault();
                                   this.setProps({syncTarget: e.target});
                                   setTimeout(function () {
                                       this.setProps({asyncTarget: e.target});
                                   }.bind(this), 10);
                               }.bind(this)
                           }, "Click it"),
                           d.p({}, "Sync target: " + this.props.syncTarget),
                           d.p({}, "Async target: " + this.props.asyncTarget));
          }
      });

      React.renderComponent(Clicker(), document.getElementById("ui"));
    </script>
  </body>
</html>
```

I would expect the sync and async targets to be the same, but if you run the above, async target will be `null`.

PR #2141, opened 2014-09-04
[idea] Automatically detect property/attribute for HTMLDOMPropertyConfig

syranide:
Proper implementation of #1512

It does not get rid of `MUST_USE_PROPERTY` (see #2140, #2202) and `MUST_USE_ATTRIBUTE` because I'm not 100% sure if we might want to use it or not, but I assume there's no need for it.

I also removed a test that I'm not sure what to do about. It relied on really hacky behavior and I don't really see a simple way of keeping it alive (not really familiar with testing), but please shout if you think otherwise.

This is beneficial for a lot of reasons, one being that we ensure DOM attributes always reflect the intended value. `MUST_USE_PROPERTY` was previously used for some attributes not supported by all browsers/elements, which when updated would just add a custom property to the node object rather than the intended attribute.

Half-related: #1848 and #1657

PR #15181, opened 2019-03-21
Upgrade rollup from 0.5.2 to 1.7.0

bjacobel:
After coming across https://github.com/facebook/react/issues/14635, I thought I would follow @gaearon's suggestion there to submit a pull request to upgrade Rollup to v1.x. 

There are a few breaking changes in this upgrade path, as described in [Rollup's changelog.](https://github.com/rollup/rollup/blob/master/CHANGELOG.md). I've included them below, but I don't have a lot of context for what any of them mean.

<details>

- Revert class id preservation from #2025 (#2048)
- Refactor missing export plugin hook (#1987)
- Deprecate the legacy option and thus IE8 support (#2141)
optimizeChunks is renamed to experimentalOptimizeChunks to reflect this feature is not production-ready yet (#2575)
- Several (mostly deprecated) options have been removed or renamed (#2293, #2409):
    ```
    banner -> output.banner
    dest -> output.file
    entry -> input
    experimentalCodeSplitting -> now always active
    experimentalDynamicImport -> now always active
    experimentalPreserveModules -> preserveModules
    exports -> output.exports
    extend -> output.extend
    footer -> output.footer
    format -> output.format
    freeze -> output.freeze
    globals -> output.globals
    indent -> output.indent
    interop -> output.interop
    intro -> output.intro
    load -> use plugin API
    moduleName -> output.name
    name -> output.name
    noConflict -> output.noConflict
    output.moduleId -> output.amd.id
    outro -> output.outro
    paths -> output.paths
    preferConst -> output.preferConst
    pureExternalModules -> treeshake.pureExternalModules
    resolveExternal -> use plugin API
    resolveId -> use plugin API
    sourcemap -> output.sourcemap
    sourceMap -> output.sourcemap
    sourceMapFile -> output.sourcemapFile
    strict -> output.strict
    targets -> use output as an array
    transform -> use plugin API
    useStrict -> output.strict
    ```
- In general, output options can no longer be used as input options (#2409)
bundle.generate and bundle.write now return a new format (#2293)
Several plugin hooks have become deprecated and will display warnings when used (#2409):
    ```
    transformBundle
    transformChunk
    ongenerate
    onwrite
    ```
- Plugin transform dependencies are deprecated in favour of using the this.addWatchFile plugin context function (#2409)
- Accessing this.watcher in plugin hooks is deprecated in favour of the watchChange plugin hook and the this.addWatchFile plugin context function (#2409)
- Using dynamic import statements will by default create a new chunk unless inlineDynamicImports is used (#2293)
- Rollup now uses acorn@6 which means that acorn plugins must be compatible with this version; acorn is now external for non-browser builds to make plugins work (#2293)
</details>

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "random-seed": "^0.3.0",
     "react-lifecycles-compat": "^3.0.2",
     "rimraf": "^2.6.1",
-    "rollup": "^0.52.1",
+    "rollup": "^1.7.0",
     "rollup-plugin-babel": "^3.0.1",
     "rollup-plugin-commonjs": "^8.2.6",
     "rollup-plugin-node-resolve": "^2.1.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -101,6 +101,16 @@
   dependencies:
     "@types/whatwg-streams" "^0.0.6"
 
+"@types/estree@0.0.39":
+  version "0.0.39"
+  resolved "https://npm.hubteam.com/npm-nexus/repository/npm-all/@types/estree/-/estree-0.0.39.tgz#e177e699ee1b8c22d23174caaa7422644389509f"
+  integrity sha512-EYNwp3bU+98cpU4lAWYYL7Zz+2gryWH1qbdDTidVd6hkiR6weksdbMadyXKXNPEkQFhXM+hVO9ZygomHXp+AIw==
+
+"@types/node@^11.9.5":
+  version "11.11.4"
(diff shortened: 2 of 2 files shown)


kunukn:
There is also this PR for updating the Rollup version https://github.com/facebook/react/pull/15037

bjacobel:
@kunukn Thanks for linking that - it looks like that PR is more complete and farther along, so I will close this one and ask contributors/reviewers to focus their efforts on that one.

PR #15184, opened 2019-03-21
[ESLint] Warn about setState directly in dep-less useEffect

gaearon:
This warns about the common mistake of using setState directly in an effect without a dependency list — which too easily leads to loops. This is the only case where I propose that we'd warn for dep-less useEffect. It looks like this:

Screen Recording 2019-03-21 at 08 53 PM

It won't warn you about `setState` nested in another function (it'll assume you know what you're doing). Only top level ones right in the effect.

It doesn't check for early returns. I considered this but then thought the pattern is still too fragile. With autofix it's not too annoying anyway.

### DIFF
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -468,9 +468,101 @@ export default {
         },
       );
 
-      if (!declaredDependenciesNode) {
+      // Warn about assigning to variables in the outer scope.
+      // Those are usually bugs.
+      let staleAssignments = new Set();
+      function reportStaleAssignment(writeExpr, key) {
+        if (staleAssignments.has(key)) {
+          return;
+        }
+        staleAssignments.add(key);
+        context.report({
+          node: writeExpr,
+          message:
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -468,9 +468,101 @@ export default {
         },
       );
 
+      // Warn about assigning to variables in the outer scope.
+      // Those are usually bugs.
```
gaearon: This code just moved from below with no changes.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -468,9 +468,101 @@ export default {
...
+
+      if (staleAssignments.size > 0) {
+        // The intent isn't clear so we'll wait until you fix those first.
+        return;
+      }
+
       if (!declaredDependenciesNode) {
+        // Check if there are any top-level setState() calls.
+        // Those tend to lead to infinite loops.
+        let setStateInsideEffectWithoutDeps = null;
```
gaearon: This code is new.

PR #15186, opened 2019-03-21
Enabled warnAboutDeprecatedLifecycles flag by default

bvaughn:
Also updated tests to account for this. (Mostly this involved merging some "internal" tests that are no longer internal.)

Note that we still only warn about "unsafe" lifecycle methods in strict mode, but now we will always warn about the "legacy" lifecycles.


bvaughn:
Ping!

bvaughn:
rightbackatyabuddy

packages/react-dom/src/__tests__/ReactComponentLifeCycle-test.js
```diff
@@ -1324,4 +1387,87 @@ describe('ReactComponentLifeCycle', () => {
...
+      }
+    }
+
+    const container = document.createElement('div');
+    expect(() =>
+      ReactDOM.render(<MyComponent x={1} />, container),
+    ).toLowPriorityWarnDev(
+      [
+        'componentWillMount is deprecated and will be removed in the next major version. ' +
+          'Use componentDidMount instead. As a temporary workaround, ' +
```
gaearon: "As a temporary workaround" — people will 100% read it as "UNSAFE_ version will be removed in 17". We've already seen this confusion despite explicit assurances it won't be. We need to emphasize this in the message.

bvaughn: The workaround is still temporary, in that using the (renamed) method is still _unsafe_.

If you have proposals for an alternate error message, post a PR, but I don't think we should make it sound like ongoing use of the unsafe lifecycles is totally fine either.

sebmarkbage: I'm not sure it's better to error on the side of less scary.

It is really unsafe even with "batched" Suspense too, and just generally a risk factor in server rendering. Not just concurrent mode. So you'll like be left behind one way or another.

So even though we probably won't completely remove it, it's probably a good idea to treat it almost as bad as if we would.

Does it matter if you can upgrade to React 18 but not use any of the features in it?

PR #15192, opened 2019-03-22
React v16.8.5

gaearon:
This is based on 16.8.4 branch with the addition of:

* Improve async useEffect warning (#15104)
* Don't set the first option as selected in select tag with `size` attribute  (#14242)
* Fix shallow renderer not allowing hooks in forwardRef render functions (#15100)
* Use same example code for async effect warning (#15118)
* Support React.memo in ReactShallowRenderer (#14816)
* [Shallow] Implement setState for Hooks and remount on type change (#15120)
* Link to useLayoutEffect gist in a warning (#15158)
* Add more info to invalid hook call error message (#15139)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,23 @@
   </summary>
 </details>
 
+## 16.8.5 (March 22, 2019)
+
+### React DOM
+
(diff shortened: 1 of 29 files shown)


gaearon:
Released.

PR #15195, opened 2019-03-22
Fix a crash in Suspense with findDOMNode

gaearon:
Another attempt at https://github.com/facebook/react/pull/14198.
Fixes the internal crash we've recently observed.

I know this fix might have weird semantics (which DOM node should it return for a suspended tree?) but it still seems better to not crash in this case IMO.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberTreeReflection.js
+++ b/packages/react-reconciler/src/ReactFiberTreeReflection.js
@@ -21,6 +21,8 @@ import {
   HostRoot,
   HostPortal,
   HostText,
+  Fragment,
+  SuspenseComponent,
 } from 'shared/ReactWorkTags';
 import {NoEffect, Placement} from 'shared/ReactSideEffectTags';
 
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -119,8 +121,27 @@ export function findCurrentFiberUsingSlowPath(fiber: Fiber): Fiber | null {
...
+        if (
+          maybeSuspense &&
+          maybeSuspense.tag === SuspenseComponent &&
+          // If state isn't null, it timed out and we have two Fragment children.
+          maybeSuspense.memoizedState !== null
+        ) {
+          parentA = maybeSuspense;
+          parentB = maybeSuspense;
+          a = maybeSuspenseFragment;
+          b = maybeSuspenseFragment;
```
gaearon: I don't know if it's legit to have them pointing to the same one. I haven't really thought whether this makes sense.

packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -119,8 +121,27 @@ export function findCurrentFiberUsingSlowPath(fiber: Fiber): Fiber | null {
...
+      if (maybeSuspenseFragment && maybeSuspenseFragment.tag === Fragment) {
+        const maybeSuspense = maybeSuspenseFragment.return;
+        if (
+          maybeSuspense &&
+          maybeSuspense.tag === SuspenseComponent &&
+          // If state isn't null, it timed out and we have two Fragment children.
+          maybeSuspense.memoizedState !== null
+        ) {
+          parentA = maybeSuspense;
+          parentB = maybeSuspense;
```
sebmarkbage: I believe it’s possible for the suspense to have an alternate and we might be coming from the disconnected tree.

gaearon:
Fixed in https://github.com/facebook/react/pull/15312

PR #15196, opened 2019-03-22
Switch to new ReactFiberScheduler implementation

acdlite:
## Based on https://github.com/facebook/react/pull/15151

Sets the `enableNewScheduler` feature flag to true.

`yarn test` now runs against the new implementation, so use `yarn test-old-scheduler` to test the old implementation. Both commands will run in CI, in case we need to revert.


acdlite:
Never mind. I'm going to create two separate builds and test behind a GK instead.

PR #15197, opened 2019-03-22
Added invariant for null/undefined create in useEffect, useLayoutEffect

### LINKED ISSUE #15194: Better error message when useEffect is called with no parameters

Feature

Suppose you give an `undefined` object or forget to provide parameters to `useEffect`:

```js
const App = () => {
    React.useEffect();
    return "";
};

ReactDOM.render(<App />, document.getElementById('container'));
```

On JSFiddle (https://jsfiddle.net/g39javzq), this error stack is logged:

```
Uncaught TypeError: create is not a function
    at commitHookEffectList (react-dom.development.js:17372)
    at commitPassiveHookEffects (react-dom.development.js:17396)
    at HTMLUnknownElement.callCallback (react-dom.development.js:143)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:193)
    at invokeGuardedCallback (react-dom.development.js:250)
    at commitPassiveEffects (react-dom.development.js:18863)
    at wrapped (react.development.js:1353)
    at flushFirstCallback (react.development.js:587)
    at flushWork (react.development.js:699)
    at MessagePort.channel.port1.onmessage (react.development.js:1093)
```

...though on a local project in React 16.8.4, this stack is logged:

```
Uncaught TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
    at _iterableToArrayLimit (my-project.chunk.js:123)
    at _slicedToArray (my-project.chunk.js:456)
    at MyComponent (myFile.js:7)
```

Neither of these were useful during the 5 minutes it took to find the culprit `useEffect` that should have been a `useState`.

In development, we check for an `undefined` `deps` argument in `useEffect` and give a better warning?

React 16.8.4, Chrome on Mac.

JoshuaKGoldberg:
I considered putting this somewhere closer to where `create` is used, but also wanted the complaint to be specific to the function & parameter name...

Fixes #15194.


JoshuaKGoldberg:
okie dokie, following up now...

image

gaearon:
We try to avoid DEV-only invariants because they can lead to inconsistent behavior between dev and prod. Ideally we would throw right before it would throw anyway. In that case maybe it’s accepable.

JoshuaKGoldberg:
@gaearon I'm not super sure if there's a change you wanted, but I'm interpreting that to mean that this should remove the `if (__DEV__)`? Applied in the latest commit 😄

acdlite:
Could you make it a dev-only warning, instead of an invariant? The idea is to avoid an extra check in production, and to make sure the behavior is consistent across dev and prod.

JoshuaKGoldberg:
@acdlite sure - how would you like me to do that? I'm not familiar with the terminology here; does a dev-only warning mean using the `warning` from `shared/warming`?

packages/react/src/ReactHooks.js
```diff
@@ -93,6 +93,15 @@ export function useEffect(
   inputs: Array<mixed> | void | null,
 ) {
   const dispatcher = resolveDispatcher();
+
+  if (__DEV__) {
+    warning(
+      create != null,
+      'useEffect must be called with a create function, but null or undefined was passed.',
+    );
+    return;
```
gaearon: Shouldn't return early because that changes the behavior between DEV and PROD.

alexanderkjeldaas:
#15369 is somewhat related to more error checking, but for the return value.

JoshuaKGoldberg:
Oh, I forgot about this! I'll tackle it soon. 🚀

JoshuaKGoldberg:
Mmhh, no, I don't have time to push this through. Best of wishes!

Issue #15088, opened 2019-03-12
useReducer - eagerReducer optimization discussion/questions


Andarist:
I'd like to continue the discussion started by me under a recent blog post by Dan as encouraged by Dan 😉 https://github.com/gaearon/overreacted.io/commit/99bfdca459ff4094ee523c7419b58989d18bc594#r32694433

Just to summarize what I've stumbled upon when experimenting with useReducer after reading that hoisted & declared in render reducers are treated differently (I've wanted to explore how they are handled by React):
1. I have no idea how to reenter eagerReducer calculation after first scheduled work (& after render gets fully processed & committed). This might very well be just me not understanding how fibers work - but currently I'm confused by this. It doesn't enter this code branch because `fiber.expirationTime` is not 0 (NoWork) and the work gets scheduled right away. Any pointers regarding this? Is this valid? Might this be a bug somewhere?
2. The logic around reducer bailouts is somewhat iffy for me - maybe it's just a matter of mentioning those in the documentation:
- not every reducer update can bailout from rendering. If the action queue gets processed in the render phase then it's just not possible, we are already rendering after all. 
- action queue gets processed in the render phase for reducers declared inside render, this means that any new state computation might depend on the "fresh" props. This is not the case for the bailout mechanism though - it's only possible to perform a bailout when dispatching an action (so when we do not have access to the fresh props). Should this restriction be mentioned in the docs? IMHO this behaviour is inconsistent - for the greater good, so it's acceptable but I think it should be documented because it's slightly inconsistent & people might trip over this.

**Note** I'm happy to provide documentation changes if needed, I'd like to discuss those points first to get a better understanding of things.

gaearon:
>I have no idea how to reenter eagerReducer calculation after first scheduled work

I don't know what that means. What are you trying to do? Preferably describe it from user perspective than impl details.

>The logic around reducer bailouts is somewhat iffy for me 

Why does it matter to you? These heuristics can change over time. They shouldn't affect how your reducer works.

> If the action queue gets processed in the render phase then it's just not possible, we are already rendering after all.

We are rendering *that component* but the bailout still determines whether we'll recurse on children or not. That's the important part.

Andarist:
> We are rendering that component but the bailout still determines whether we'll recurse on children or not. That's the important part.

[...]

PR #15198, opened 2019-03-23
Do not keep bailouted actions which might lead to reducer computing wrong value on update later

Andarist:
This is a fix for what I think is a bug which I've reported in https://github.com/facebook/react/issues/15088

The issue might be observed here https://codesandbox.io/s/n97q1o6nq4 . Just increment the counter -> disable the whole thing -> increment few times (those are going to bail out) -> enable the counter -> observe a counter incrementing by the amount of actions which have bailouted previously.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -1123,21 +1123,6 @@ function dispatchAction<S, A>(
       next: null,
     };
 
-    // Append the update to the end of the list.
-    const last = queue.last;
-    if (last === null) {
-      // This is the first update. Create a circular list.
-      update.next = update;
-    } else {
-      const first = last.next;
-      if (first !== null) {
-        // Still circular.
-        update.next = first;
-      }
-      last.next = update;
-    }
-    queue.last = update;
-
     if (
       fiber.expirationTime === NoWork &&
       (alternate === null || alternate.expirationTime === NoWork)
@@ -1182,6 +1167,22 @@ function dispatchAction<S, A>(
         warnIfNotCurrentlyBatchingInDev(fiber);
       }
     }
+
+    // Append the update to the end of the list.
+    const last = queue.last;
+    if (last === null) {
+      // This is the first update. Create a circular list.
+      update.next = update;
+    } else {
+      const first = last.next;
+      if (first !== null) {
+        // Still circular.
+        update.next = first;
+      }
+      last.next = update;
+    }
+    queue.last = update;
+
     scheduleWork(fiber, expirationTime);
   }
 }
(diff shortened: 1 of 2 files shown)


acdlite:
This is an interesting bug, but I don't think it's the right fix. I think I have a better one but I'm on my phone right now so I'll follow up later, or on Monday.

Andarist:
Might not be the best, although the fix is pretty straightforward and doesn't seem like a hacky solution. If you have any thoughts on how this could be improved I'm all ears - I'd love to work on this if possible. Enjoy your weekend!

acdlite:
Test case for you to consider: A parent and a child are both updated in the same batch. The child bails out early, but the parent doesn’t. The parent re-renders and passes a new prop to the child, which closes over the new prop in its reducer. What do you think should happen?

Andarist:
Didn't consider that! I assume that React's assumption is that scheduled work in a batch should execute top-down, therefore eager bailout lower in the tree should get ignored (or rather - the action should get processed by the reducer in the render phase if a parent has provided new props to a child which enable the child's state update).

I've added a failing test case for this 👍 Do you have any possible solutions in mind? I could try to explore this fully on my own but for sure it would take me way longer as I'm not that familiar with the codebase here. I would appreciate any pointers so I could continue working on a fix.

Andarist:
@acdlite any tips on how I could implement proper fix for this?

Andarist:
@acdlite friendly 🏓 would love to explore fixing this but atm I'm not sure what the appropriate approach to tackle this would be

gaearon:
I think @acdlite would probably have already replied or fixed this if he had a strategy in mind. We appreciate the help but I think you might need to try something and look at in more depth to form a proposal if you want to move it forward.

sebmarkbage:
I think we should probably disable this optimization for useReducer and split out the implementation of useState to a separate implementation since that can't change the reducer it's  always safe to precompute the state. We should also forbid suspending in a setState reducer. This simplifies the implementation in terms of number of fields we need to keep around and should fix this issue by always rerendering reducers.

Andarist:
@sebmarkbage let me know if I could help with that - the codebase is complex though and I probably would not be able to make those changes on my own without some guidance from the React team.

Andarist:
The issue still persists: https://codesandbox.io/s/silly-paper-22org

gaearon:
@andarist do you want to take a shot at the above suggestion?

Andarist:
@gaearon yes! but frankly - I'm not sure what is being exactly suggested right now. Should I "simply" remove this optimization from useReducer and provide a new implementation for useState (not depending on useReducer)? 

Ono concern I have about removing this optimization is that we can't know if a reducer is static or dynamic, so if we remove this entirely we'd have to always schedule a full component render for each dispatched action. Is that correct?

gaearon:
>Should I "simply" remove this optimization from useReducer and provide a new implementation for useState (not depending on useReducer)?

Maybe let's start with this and then figure out if we need to do something else.

Andarist:
Closing this as it is supposed to be fixed with other reconciler stuff as per [this comment](https://github.com/facebook/react/pull/18241#issuecomment-607523516)

Issue #11877, opened 2017-12-18
Number input gets cleared when typing period as decimal mark


Hilzu:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
My OS and browser are configured to a locale that uses comma as the decimal mark (Finland for those interested). In the codepen below when I accidentally type a period after some number the whole input gets cleared. This is somehow related to the parent component's state being updated because if the `defaultValue` prop is removed from the `<Input>` component the input doesn't get cleared.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template for React 16: https://jsfiddle.net/Luktwrdm/, template for React 15: https://jsfiddle.net/hmbg7e9w/).**
https://codepen.io/anon/pen/aEOgNL?editors=0010

**What is the expected behavior?**
The input should retain its visible value even if it would be invalid (can't be converted to a number).

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
React 16.2.0 and Chrome 63 on macOS 10.12.6.

Tested in Safari 11.0.2 and typing a period kinda works. After typing the first number after the period it gets converted to a comma but the cursor jumps to the beginning.

Also tested in Firefox 57.0.1 but it seems to think that period is the correct decimal mark.

gaearon:
cc @nhunzaker

gaearon:
Can you check if older React versions also had this problem?

Hilzu:
Ported the example to the React 15 fiddle and the behaviour seems identical in Chrome, Safari and Firefox https://jsfiddle.net/hmbg7e9w/2/

nhunzaker:
Ah this is super fun...  What is the best way to compare the equality of a number when one is sourced as a string and the other a floating point number?

**For some background:**

Number inputs do not support the text selection API. We try to avoid touching the `value` property of number inputs as much as possible because setting a new value means the cursor jumps to the end while someone is typing. We can't restore user's text position after a value changes. 

Since number inputs are fussy (particularly in Chrome), we use loose equality checks, like `"1.2" == 1.2` for this check, because usually controlled number inputs are stored as numeric values in component state. 

The problem is that loose type coercion doesn't appear to locale sensitive. 

**For my own benefit, it would be great if you could confirm the following behavior:**

1. When the input is set to `"1,2"`, what do browsers report using `value.valueAsNumber`?
[...]

PR #15211, opened 2019-03-25
Ignore update when React and number input isNaN()

nhunzaker:
This commit prevents an issue where assigning NaN to an invalid input causes a text wipe that disrupts text entry:

nan-input

This happens if referencing `input.valueAsNumber` directly, or by calling `Number(input.value)` in the onChange callback of an input. This results in a comparison of `NaN` to `""`, which always returns `true`, assigning a blank string to the input and clearing the input's text.

It's also worth mentioning that assigning NaN to an input raises this warning:

image

However we still allow the value to fall through and get assigned to the input.

This code is pretty gnarly, and I don't love adding more knots, but I believe this addresses https://github.com/facebook/react/issues/11877 and avoids some bad user experience.

## Test Plan

Take a look at the number input fixtures and the new fixture at the bottom here:
https://react-dom-fixtures-nan-inputs.surge.sh/number-inputs

### DIFF
--- a/packages/react-dom/src/client/ReactDOMInput.js
+++ b/packages/react-dom/src/client/ReactDOMInput.js
@@ -177,13 +177,22 @@ export function updateWrapper(element: Element, props: Object) {
 
   if (value != null) {
     if (type === 'number') {
-      if (
-        (value === 0 && node.value === '') ||
-        // We explicitly want to coerce to number here if possible.
-        // eslint-disable-next-line
-        node.value != (value: any)
-      ) {
-        node.value = toString(value);
+      let nodeValue = node.value;
+
+      // If the React prop value is NaN and current input value is invalid,
+      // it means both React and the DOM have invalid states (reported as
+      // empty). In this case, do not assign the value property, which would
+      // clear user input. The resulting state remains the same while still
+      // maintaining user expectations.
+      if (!(isNaN(value) && nodeValue === '')) {
+        if (
+          (value === 0 && nodeValue === '') ||
+          // We explicitly want to coerce to number here if possible.
+          // eslint-disable-next-line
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/client/ReactDOMInput.js
```diff
@@ -177,13 +177,22 @@ export function updateWrapper(element: Element, props: Object) {
...
-      ) {
-        node.value = toString(value);
+      let nodeValue = node.value;
+
+      // If the React prop value is NaN and current input value is invalid,
+      // it means both React and the DOM have invalid states (reported as
+      // empty). In this case, do not assign the value property, which would
+      // clear user input. The resulting state remains the same while still
+      // maintaining user expectations.
+      if (!(isNaN(value) && nodeValue === '')) {
```
nhunzaker: This actually trips up on browser that don't support number inputs.

I don't think we should execute this code path if the browser doesn't support number inputs. We can handle that by just checking `input.type` in IE9, it reports "text" instead of "number".

aweary: Is IE9 the only browser we support that doesn't support number inputs? I feel like we're getting close to the point where we should consider dropping IE9 support.

nhunzaker: Ya. IE9 is the only one. I think we're getting close too - should we consider it?

philipp-spiess: Looking at https://caniuse.com/usage-table: How come IE9 is still at 0.2%? 🙈

fhrf: It's been more than a year, IE9 is still 0.22% 👀

Issue #15019, opened 2019-03-05
The browser crashes when use React.lazy return Promise.resolve(undefined)


angular-moon:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
The browser crashes 

```js
const Loading = () => <div>loading...</div>;
const Component = lazy(() => Promise.resolve(undefined));

function App() {
  return (
    <div className="App">
      <Suspense fallback={<Loading />}>
        <Component />
      </Suspense>
    </div>
  );
}
```
*I can't provide an example of CodeSandbox, which would cause the browser to crash*

**What is the expected behavior?**
throw Error

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
react@16.8.3
react-dom@16.8.3
chrome 72.0.3626.119

gaearon:
If the browser itself crashes then it's a browser bug. We can't fix it.

angular-moon:
emm..., I mean the browser isn't responding, memory is up, and it feels like the program might be in an infinite loop

gaearon:
Ok I see what you mean now. You’re saying React enters an infinite loop it lazy import resolves to undefined. That sounds like something we’d want to fix (even though it’s a bug in the application code).

daychongyang:
I have the same problem after using React.lazy for code splitting,  sometimes there is memory soaring, the page is unresponsive, and the console is madly reporting an error.  Has this problem been solved in the current release? @gaearon @raunofreiberg   (⊙ˍ⊙)

PR #15222, opened 2019-03-27
Fix React.lazy infinite loop on passing undefined

raunofreiberg:
Related to #15019 

The following code no longer produces a loop.

```
const Component = React.lazy(() => Promise.resolve(undefined));
      ReactDOM.render(
        <div>
          <h1>Hello World!</h1>
          <React.Suspense fallback="..">
            <Component />
          </React.Suspense>
        </div>,
      );
```

Instead, an error is shown: (tried on dev.html standalone fixture)

```
Uncaught Invariant Violation: Element type is invalid. Received a promise that resolves to: undefined. Lazy element type must resolve to a class or function.
```

### DIFF
--- a/packages/react-reconciler/src/ReactFiberLazyComponent.js
+++ b/packages/react-reconciler/src/ReactFiberLazyComponent.js
@@ -50,9 +50,9 @@ export function readLazyComponentType<T>(lazyComponent: LazyComponent<T>): T {
       thenable.then(
         moduleObject => {
           if (lazyComponent._status === Pending) {
-            const defaultExport = moduleObject.default;
+            const _result = moduleObject ? moduleObject.default : moduleObject;
             if (__DEV__) {
-              if (defaultExport === undefined) {
+              if (result === undefined) {
                 warning(
                   false,
                   'lazy: Expected the result of a dynamic import() call. ' +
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactFiberLazyComponent.js
```diff
@@ -50,9 +50,9 @@ export function readLazyComponentType<T>(lazyComponent: LazyComponent<T>): T {
       thenable.then(
         moduleObject => {
           if (lazyComponent._status === Pending) {
-            const defaultExport = moduleObject.default;
+            const _result = moduleObject ? moduleObject.default : moduleObject;
```
eps1lon:
```suggestion
            const _result = moduleObject != null ? moduleObject.default : moduleObject;
```
It's probably a bit exotic to check for strings or numbers but it's closer to the original behavior.

packages/react-reconciler/src/ReactFiberLazyComponent.js
```diff
@@ -50,9 +50,9 @@ export function readLazyComponentType<T>(lazyComponent: LazyComponent<T>): T {
...
             if (__DEV__) {
-              if (defaultExport === undefined) {
+              if (result === undefined) {
```
eps1lon:
```suggestion
              if (_result === undefined) {
```
`result` is from the upper scope.

daychongyang:
Didn't it progress?

raunofreiberg:
@gaearon Ping

cseas:
I'm experiencing the error in #18768 with react 17.0.1

PR #15228, opened 2019-03-27
Adds experimental event component responder surfaces

trueadm:

This PR adds Press and Hover responder surfaces, some additional event context APIs and fixes some bugs surfaced from the additional Press tests added.

#15036
#15112
#15150
#15168
#15179

This PR also adds tests that validate the triggering of event responder handleEvent function.


packages/react-events/src/Press.js
```diff
@@ -0,0 +1,332 @@
...
+  },
+  handleEvent(
+    context: EventResponderContext,
+    props: Object,
+    state: PressState,
+  ): void {
+    const {eventTarget, eventType, event} = context;
+
+    switch (eventType) {
+      case 'keydown': {
```
necolas: I think we should consider making keyboards trigger `pressin`, `pressout`, `press`, and even `longpress` events.

trueadm: Will do this at a later point, in a follow up.

trueadm:
I checked manually and this PR does not add any size compared to master. The results.json in master needs updating.

sebmarkbage:
We don't compare against the results.json file but a rebuilt version. I tested locally and the (small) build increase shown by sizebot above seems legit.

trueadm:
@sebmarkbage I saw it was approved earlier, we also spoke about it earlier too. None of these code paths should get used on FB, but I’ll be sure to sanity check this.

 The size difference doesn’t make sense. If I remove the changes in this PR and build, I still get 0.3% increase locally.

sebmarkbage:
I checked out this commit built and then checked out the previous commit and built that. Then compared the two bundles.

I never compare against the checked in JSON. That one never matches up.

PR #15232, opened 2019-03-28
Allow DevTools to toggle Suspense fallbacks

gaearon:
What it says. Not sure if this is sufficient yet. I'll build it out a bit more before we get this in.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -77,6 +77,7 @@ import {
   cloneChildFibers,
 } from './ReactChildFiber';
 import {processUpdateQueue} from './ReactUpdateQueue';
+import {shouldSuspend} from './ReactFiberDevToolsHook';
 import {
   NoWork,
   Never,
@@ -1391,6 +1392,12 @@ function updateSuspenseComponent(
 ) {
   const mode = workInProgress.mode;
   const nextProps = workInProgress.pendingProps;
+
+  if (__DEV__) {
+    if (shouldSuspend(workInProgress)) {
+      workInProgress.effectTag |= DidCapture;
+    }
+  }
 
   // We should attempt to render the primary children unless this boundary
   // already suspended during this render (`alreadyCaptured` is true).
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -77,7 +77,6 @@ import {
   cloneChildFibers,
 } from './ReactChildFiber';
 import {processUpdateQueue} from './ReactUpdateQueue';
-import {shouldSuspend} from './ReactFiberDevToolsHook';
 import {
   NoWork,
   Never,
@@ -97,6 +96,7 @@ import {
   registerSuspenseInstanceRetry,
 } from './ReactFiberHostConfig';
 import type {SuspenseInstance} from './ReactFiberHostConfig';
+import {shouldSuspend} from './ReactFiberReconciler';
 import {
   pushHostContext,
   pushHostContainer,
--- a/packages/react-reconciler/src/ReactFiberDevToolsHook.js
+++ b/packages/react-reconciler/src/ReactFiberDevToolsHook.js
@@ -16,9 +16,6 @@ declare var __REACT_DEVTOOLS_GLOBAL_HOOK__: Object | void;
 
 let onCommitFiberRoot = null;
 let onCommitFiberUnmount = null;
-let shouldSuspendFiber = function() {
-  return false;
-};
 let hasLoggedError = false;
 
 function catchErrors(fn) {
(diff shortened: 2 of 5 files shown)


packages/react-reconciler/src/ReactFiberDevToolsHook.js
```diff
@@ -71,6 +74,14 @@ export function injectInternals(internals: Object): boolean {
     onCommitFiberUnmount = catchErrors(fiber =>
       hook.onCommitFiberUnmount(rendererID, fiber),
     );
+    onCommitFiberRoot = catchErrors(root =>
+      hook.onCommitFiberRoot(rendererID, root),
+    );
```
sebmarkbage: Duplicate.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1389,6 +1390,10 @@ function updateSuspenseComponent(
   const mode = workInProgress.mode;
   const nextProps = workInProgress.pendingProps;
 
+  if (shouldSuspend(workInProgress)) {
```
sebmarkbage: Should be wrapped in `__DEV__`.

packages/react-reconciler/src/ReactFiberDevToolsHook.js
```diff
@@ -71,6 +74,14 @@ export function injectInternals(internals: Object): boolean {
...
+    if (hook.shouldSuspendFiber) {
+      shouldSuspendFiber = catchErrors(fiber =>
+        hook.shouldSuspendFiber(rendererID, fiber),
+      );
+    }
```
sebmarkbage: Should be wrapped in `__DEV__` too.

acdlite:
Make sure you test in both sync mode and concurrent mode

gaearon:
DevTools https://github.com/bvaughn/react-devtools-experimental/pull/36

bvaughn:
I don't think this is sufficient for DevTools to do feature detection.

gaearon:
>We should probably add a new prop to where we do the injection that signals whether the current version+build supports this override behavior. This way DevTools won't display a confusing button that doesn't work.

Yeah.

>Looking at the implementation, I'm curious: why not inject a method e.g. suspendFiber(true|false) that works like overrideProps in that it both triggers a re-render and could be used as our signal for whether the build of React supports this action?

That could work. However, I wanted to make it so that the condition can be adjusted dynamically, and doesn't necessarily have to be "here's a Set of Fibers". For example it would be nice if we could do a quick check based on some predicate. Leaving it as a method in DevTools hook makes that a bit more flexible.

sebmarkbage:
During an initial render you can have this pause boundaries that didn’t exist yet.

Even in a UI that has boundaries, switching boundary can have side-effects that deletes things in the tree you switched out of.

gaearon:
>During an initial render you can have this pause boundaries that didn’t exist yet.

You'd need the Hook to say "false" before the renderer is injected for that. Which is something we need for Profiler Reload+Start anyway I guess?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1389,6 +1390,12 @@ function updateSuspenseComponent(
   const mode = workInProgress.mode;
   const nextProps = workInProgress.pendingProps;
 
+  if (__DEV__) {
+    if (shouldSuspend(workInProgress)) {
+      workInProgress.effectTag |= DidCapture;
```
acdlite: Suggestion: replace this with

```js
throwException(
  root,
  workInProgress, // returnFiber
  null, // sourceFiber
  dummyPromiseThatNeverResolves,
  renderExpirationTime
);
```

This will require refactoring `throwException` a bit, or perhaps splitting it into multiple functions, since it currently assumes that you have a source fiber.

bvaughn:
I hadn't realized that you were thinking of supporting a feature like "reload and step through suspense boundaries". Given that, deferring  to DevTools does make more sense.

I'm not entirely comfortable with the current solution for injection in the reload-and-do-something case, but it's reasonable to assume this could piggy back off of whatever solution we end up with there.

gaearon:
I updated the PR to a different API. Now it exposes `overrideSuspense` instead of asking the Hook to implement `shouldSuspend`. Calling `overrideSuspense` is how you supply a custom implementation.

The upside of this is that it makes it easy to feature-test whether a particular React renderer supports this feature from DevTools.

Note that while this moves the override to render injection (as opposed to reading a field from the Global Hook), it doesn't actually make it any harder to "pause" the first render. If we wanted, we could call `overrideSuspense` at the Global Hook level during the synchronous injection.

I've also added a CM test.

gaearon:
>Still not super thrilled about the force re-render bit in the DevTools PR that accompanies this, but I can get over it 😄

After conversation with Sebastian, I plan to make a few more changes:

* Rename `overrideSuspense` to `setSuspenseHandler`
* Add `scheduleUpdate` that removes the need for `overrideProps` hack

In the future, we'll move away from the `overrideProps` mechanism to (out of scope for this PR) `setPropsHandler` which would let us delegate the overriding logic to DevTools itself. That would let us implement overrides as shadowing instead of the current fragile mechanism.

bvaughn:
Nice nice! I like both of those changes 👍

> In the future, we'll move away from the overrideProps mechanism to (out of scope for this PR) setPropsHandler which would let us delegate the overriding logic to DevTools itself. That would let us implement overrides as shadowing instead of the current fragile mechanism.

Interesting. I'd love to talk more about this. I'm a little wary of the shadowing idea. (We've talked about this a bit in the past, either with you or Sebastian I forget.)

PR #15242, opened 2019-03-28
Adds React event component and React event target support to SSR renderer

trueadm:

This PR adds support for server side rendering for the experimental event API along with tests. I also added some ReactDOM renderer tests to the same test file to be consistent and give more coverage.

#15036
#15112
#15150
#15168
#15179 
#15228

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRenderer.js
+++ b/packages/react-dom/src/server/ReactPartialRenderer.js
@@ -22,6 +22,7 @@ import ReactSharedInternals from 'shared/ReactSharedInternals';
 import {
   warnAboutDeprecatedLifecycles,
   enableSuspenseServerRenderer,
+  enableEventAPI,
 } from 'shared/ReactFeatureFlags';
 
 import {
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/__tests__/ReactFiberEvents-test-internal.js
```diff
@@ -5,7 +5,6 @@
  * LICENSE file in the root directory of this source tree.
  *
  * @emails react-core
- * @jest-environment node
```
trueadm: This existed from a copy and paste from a previous test file. We don't want it here as we want Jest to give the JSDOM for `document` and using this setting removes JSDOM support for this test.

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -1162,6 +1165,27 @@ class ReactDOMServerRenderer {
...
+                childIndex: 0,
+                context: context,
+                footer: '',
+              };
+              if (__DEV__) {
+                ((frame: any): FrameDev).debugElementStack = [];
+              }
+              this.stack.push(frame);
+            }
+            return '';
```
sebmarkbage: We can throw if enableEventAPI isn't enabled. File size isn't a big issue and better reject than silently allow it.

PR #15246, opened 2019-03-28
Fix longpress in experimental Press event module

necolas:
The `longpress` event is dispatched during a press interaction, rather than after it has ended.

The `longPressCancelsPress` prop can be used to prevent 'press' being dispatched if `longpress` has already been dispatched.

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -80,6 +80,15 @@ function dispatchPressInEvents(
           true,
         );
       }
+      if (props.onLongPress) {
+        const longPressEventListener = e => {
+          props.onLongPress(e);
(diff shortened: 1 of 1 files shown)


packages/react-events/src/Press.js
```diff
@@ -256,7 +250,7 @@ const PressResponder = {
             ) {
               return;
             }
-            // Ignore right-clicks
+            // Ignore middle- and right-clicks
```
trueadm: Typo with the additional `-`

necolas: Intentional as per https://english.stackexchange.com/questions/113422/how-to-use-hyphens-appropriately-when-listing-multiple-hyphenated-terms

PR #15248, opened 2019-03-29
Add additional event API responder surfaces

trueadm:

This PR adds more event responder surfaces, fills in the TODOs around the responder ownership model and also fixes some bugs found with internal user testing.

### DIFF
--- a/packages/react-events/src/Drag.js
+++ b/packages/react-events/src/Drag.js
@@ -1,0 +1,187 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import type {EventResponderContext} from 'events/EventTypes';
+import {REACT_EVENT_COMPONENT_TYPE} from 'shared/ReactSymbols';
+
+const targetEventTypes = ['pointerdown', 'pointercancel'];
+const rootEventTypes = ['pointerup', {name: 'pointermove', passive: false}];
+
+type DragState = {
+  dragTarget: null | EventTarget,
+  isPointerDown: boolean,
+  isDragging: boolean,
+  startX: number,
+  startY: number,
+  x: number,
+  y: number,
+};
+
+// In the case we don't have PointerEvents (Safari), we listen to touch events
+// too
+if (typeof window !== 'undefined' && window.PointerEvent === undefined) {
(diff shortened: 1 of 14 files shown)


philipp-spiess:
👀 I really like how this is shaping up.

packages/react-events/src/Drag.js
```diff
@@ -0,0 +1,187 @@
...
+          const x = (obj: any).screenX;
+          const y = (obj: any).screenY;
+          state.x = x;
+          state.y = y;
+          if (!state.isDragging && x !== state.startX && y !== state.startY) {
+            let shouldEnableDragging = true;
+
+            if (
+              props.onShouldClaimOwnership &&
+              props.onShouldClaimOwnership()
```
necolas: This prop name doesn't seem quite right. In the RN responder system there are `onStartShouldSetResponder` and `onMoveShouldSetResponder` props to determine whether the view wants to become the responder after a start or move event.

packages/react-events/src/Drag.js
```diff
@@ -0,0 +1,187 @@
...
+                false,
+                eventData,
+              );
+            }
+            (event: any).preventDefault();
+          }
+        }
+        break;
+      }
+      case 'pointercancel':
```
necolas: we might eventually want a separate event to be called if the drag is cancelled

packages/react-events/src/Focus.js
```diff
@@ -0,0 +1,101 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import type {EventResponderContext} from 'events/EventTypes';
+import {REACT_EVENT_COMPONENT_TYPE} from 'shared/ReactSymbols';
+
+const targetEventTypes = [
+  {name: 'focus', passive: true, capture: true},
```
necolas: curious why these need to be passive and capture

packages/react-events/src/Swipe.js
```diff
@@ -0,0 +1,217 @@
...
+          if (props.onShouldClaimOwnership) {
+            context.releaseOwnership(state.swipeTarget);
+          }
+          const direction = state.direction;
+          const lastDirection = state.lastDirection;
+          if (direction !== lastDirection) {
+            if (props.onSwipeLeft && direction === 3) {
+              dispatchSwipeEvent(
+                context,
+                'swipeleft',
```
necolas: I think it might be preferable to have a single event like `onSwipe` with event data that provides the direction, velocity, etc. That would be simpler to work with in userland than potentially listening to 4 events.

PR #15256, opened 2019-03-29
Fix on(Long)PressChange events in experimental press event API

necolas:
Make sure that `onPressChange` is only called if `longPressCancelsPress` is `false`.
And make sure that `onLongPressChange` is called when a long press ends.

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -80,7 +80,7 @@ function dispatchPressInEvents(
           true,
         );
       }
-      if (props.onLongPress) {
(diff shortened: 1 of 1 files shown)


packages/react-events/src/Press.js
```diff
@@ -110,7 +110,7 @@ function dispatchPressOutEvents(
       true,
     );
   }
-  if (props.onPressChange) {
+  if (props.onPressChange && !longPressCancelsPress) {
```
trueadm: Missing `props`

Issue #15255, opened 2019-03-29
Profiler marks


bvaughn:
Can we expose a lighter weight set of Performance "marks" for people consuming browser Performance tracing? e.g. when a particular `Profiler` commits. This should be a lot less heavyweight than the full mark-and-measure stuff.

Brooooooklyn:
Is this something like **Application level tracing**?

PR #15260, opened 2019-03-29
Log User Timing API entries for Profiler nodes in the production+prof…

bvaughn:
Resolves issue #15255 cc @bgirard 

React uses the [User Timing API](https://developer.mozilla.org/en-US/docs/Web/API/User_Timing_API) to mark and measure time spent working on various things for development builds. This can be a useful source of information when it comes to cross-referencing browser performance traces. We avoid doing this in production mode though, due to the performance overhead of using the API.

I believe that we could make an exception for our [profiling bundles](https://fb.me/react-profiling) and log measurements for `Profiler` nodes only. This should be a lot less heavyweight than the full development measurements, while still providing a potentially useful signal in the trace logs.

Here's an initial pass at this. This change has a minimal impact on the development bundle behavior- (including the `Profiler` id as part of the display name)- which seemed safe enough. I also verified that none of the timer related code leaks into the production bundle.

### Development bundle

Screen Shot 2019-03-29 at 1 25 51 PM

### Profiling bundle

Screen Shot 2019-03-29 at 1 30 42 PM

### DIFF
--- a/packages/react-reconciler/src/ReactDebugFiberPerf.js
+++ b/packages/react-reconciler/src/ReactDebugFiberPerf.js
@@ -9,7 +9,10 @@
 
 import type {Fiber} from './ReactFiber';
 
-import {enableUserTimingAPI} from 'shared/ReactFeatureFlags';
+import {
+  enableProfilerTimer,
+  enableUserTimingAPI,
+} from 'shared/ReactFeatureFlags';
 import getComponentName from 'shared/getComponentName';
 import {
   HostRoot,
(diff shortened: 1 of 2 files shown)


bgirard:
Looks great!

bvaughn:
> What's the contract of this API?

This is a good question. To be honest, I didn't intend to suggest a contract here really. It just seemed like this would offer some potentially useful additional data points for some of our real world users.

> If it heavily loses its usefulness when it's not reliable and we can't keep it reliable, then it might be more a source a noise than not.

Fair enough.

It seems like this is pain point / a gap for users currently though. If you'd like, I think I could put together an alternate proposal that's a bit more robust to the deep re-rendering concern.

bvaughn:
I guess this PR was contentious so it never landed. Closing for now.

Issue #15257, opened 2019-03-29
[Umbrella] React Flare


necolas:
This issue tracks tasks related to the React DOM implementation of the experimental React Events API (`react-events`) aka React Flare (our internal code-name for the project). The idea is to extend React's event system to include high-level events that allow for consistent cross-device and cross-platform behavior.

*Note: For now this is completely experimental and won't affect the open source builds of React.*

## Core

- [ ] Document specific UX patterns that this enables and fixes. (inc. from my workplace post on styles, Dan & Brian dev tools, apps, etc.)
- [x] Every user-facing event should have a `timeStamp`. Should we provide a standard way to derive data that might be common across events and require different logic for different positional information, e.g., point, delta, offset, etc?
- [ ] How to test discrete vs continuous events?
- [ ] Should we polyfill pointer events or not dispatch emulated touch/mouse events in the core, to avoid multiple event components needing to implement the same logic to ignore emulated events?
- [x] Investigate native pointer capture for retargeting events - https://www.w3.org/TR/pointerevents/#pointer-capture. This could be particularly useful for drag/swipe UX.
- [x] Limit all event components to have a single host node as child? If not enforced for all components, we might need this for Hover/Focus/Press at least.
- FYI https://github.com/facebook/react/pull/12919
- [x] Optimize bundle output.
- [x] Replace magic object export with `React.unstable_createEventComponent(type: Symbol | number, responder: ReactEventResponder, displayName: ?string)`
- [x] Document core API for building responders
- [x] Stop propagation by default but scope to event module types.
- [x] Investigate ways to listen to root events without first needing to wait for a target event to be dispatched to the event component.
- [x] Remove `listener` from event object passed to callbacks like `onPress`
- [x] Ensure discrete events do not unnecessarily force flush previous non-discrete events.
- [x] Ensure event responder get unmounted correctly and pending events (e.g. long press) properly get removed when doing so. Maybe an unmount lifecycle method needs adding to event responders?
- [x] Create a new synthetic event and dispatch mechanism that relies on the event being generated in the event responder module, rather than from React. This allows for the event to be better typed and have a much simpler implementation for how React fires the event when dispatched + it means we can avoid pulling in the current SyntheticEvent system and all its dependencies.

## Ownership

- [ ] Decide on prop name equivalents to RN `onStartShouldSetResponder` and `onMoveShouldSetResponder`.

## Focus module

- [ ] Determine the potential use cases for DOM `focusin` and `focusout` events.
[...]

PR #15261, opened 2019-03-29
Add event touch hit target hit slop logic to experimental event API

trueadm:

This PR adds the necessary logic to mount and update the hit slop for touch hit targets. It does this by creating an element, absolutely positioning it inside a child and then appending it the target DOM element. This required moving some parts around to add the necessary fields required to properly ensure that we can create the element in the right namespace. There are also some TODOs that will be tackled in the coming weeks.


packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1343,3 +1352,157 @@ export function listenToEventResponderEventTypes(
...
+          top,
+          bottom,
+          rootContainerInstance,
+          parentNamespace,
+        );
+        // We need to make the target relative so we can make the hit slop
+        // element inside it absolutely position around the target.
+        // TODO add a dev check for the computed style and warn if it isn't
+        // compatible.
+        element.style.position = 'relative';
```
trueadm: @necolas We need to come up with a good solution for this. We should mark this in the umbrella issue and track it as to how we can validate it and make sure stacking order etc is good.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1343,3 +1352,127 @@ export function listenToEventResponderEventTypes(
...
+        }
+        const top = nextProps.top;
+        if (lastProps.top !== top) {
+          if (top === undefined) {
+            hitSlopElementStyle.top = '';
+          } else {
+            hitSlopElementStyle.top = `-${top}px`;
+          }
+        }
+        const bottom = nextProps.left;
```
Janetx: should this be `const bottom = nextProps.bottom` ?

trueadm: Fixed :)

packages/react-art/src/ReactARTHostConfig.js
```diff
@@ -445,7 +445,9 @@ export function handleEventComponent(
 
 export function handleEventTarget(
   type: Symbol | number,
-  props: Props,
+  lastProps: Props,
```
necolas: I think we should keep the name as `props` if these are the currently-in-use props (rather than the previous props)

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1343,3 +1352,157 @@ export function listenToEventResponderEventTypes(
...
+  const hitSlopElement = createElement(
+    'div',
+    emptyObject,
+    rootContainerInstance,
+    parentNamespace,
+  );
+  const hitSlopElementStyle = ((hitSlopElement: any): HTMLElement).style;
+
+  hitSlopElementStyle.position = 'absolute';
+  hitSlopElementStyle.display = 'block';
```
necolas: Not necessary as an absolutely positioned element will default to `block`

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1343,3 +1352,157 @@ export function listenToEventResponderEventTypes(
...
+  if (top !== undefined) {
```
necolas: The value could also be `null`. When a `value == null` we should default to `0px`

packages/react-art/src/ReactARTHostConfig.js
trueadm: We use lastProps, nextProps naming through all the host configs.

trueadm:
@sebmarkbage In regards to SSR, there is a task on here https://github.com/facebook/react/issues/15257 to do it as a separate point.

In regards to `insertBefore`, why wouldn't that work?

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1343,3 +1352,151 @@ export function listenToEventResponderEventTypes(
...
+        );
+      } else {
+        parentNamespace = ((hostContext: any): HostContextProd);
+      }
+
+      const element = ((node: any): HTMLElement);
+      // We update the event target state node to be that of the element.
+      // We can then diff this entry to determine if we need to add the
+      // hit slop element, or change the dimensions of the hit slop.
+      const lastElement = internalInstanceHandle.stateNode;
```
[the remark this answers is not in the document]
trueadm: I wasn't aware of this constraint. I can re-work it so that `stateNode` is an object ref instead. Then the object ref is passed to this function in host config; would that make more sense?

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1343,3 +1352,151 @@ export function listenToEventResponderEventTypes(
...
+          top,
+          bottom,
+          rootContainerInstance,
+          parentNamespace,
+        );
+        // We need to make the target relative so we can make the hit slop
+        // element inside it absolutely position around the target.
+        // TODO add a dev check for the computed style and warn if it isn't
+        // compatible.
+        element.style.position = 'relative';
```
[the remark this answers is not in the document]
trueadm: Yes, this was a bit talking point with @necolas too. There's no cheap way to get around having to do a computed lookup to find what the position is. Do you have any ideas on how we might solve this problem?

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1343,3 +1352,151 @@ export function listenToEventResponderEventTypes(
...
+  top: number | void | null,
+  bottom: number | void | null,
+  rootContainerInstance: Element | Document,
+  parentNamespace: string,
+): Element {
+  const hitSlopElement = createElement(
+    'div',
+    emptyObject,
+    rootContainerInstance,
+    parentNamespace,
```
sebmarkbage: This won't work if the parentNamespace isn't HTML so you should throw if you don't intend to support other namespaces.

sebmarkbage:
If the EventTarget's element gets a child added at the end, then React will think it's the last child so we use parentNode.appendChild(newNode) to insert it. That will put it at the end of the list but I think you want the hit slop to always be the last child in the children list.

```
<EventTarget>
  <div>
    Hi
    {c ? <span>I will be after the hit slop</span> : null}
  </div>
</EventTarget>
```

Can you add this test?

trueadm:
@sebmarkbage As the hit slop is absolutely positioned, I'm not sure it matters. It might mess with CSS styles though. With your idea around `<EventTarget>` being a child of the `<div>` how would it work?

trueadm:
I'm closing this and will open another PR with a different implementation.

Issue #15267, opened 2019-03-30
Add support to React DevTools for react-events


bvaughn:
This could mean a lot of things, longer term, but for the short term I think we need a way to uniquely identify types of `EventComponent` (something akin to the `displayName` attribute that custom components support). The information currently available on the Fiber `type` doesn't provide that.

As for `EventTarget` types, I see we have a unique symbol for the touch hit type (the only current target type?)- do we plan to add more unique symbols for future target types? If so, this would be sufficient to identify these.

trueadm:
EventTarget can indeed have more types, but there's only one right now – plus the user can create their own. I have a PR up that adds these fields to both objects: https://github.com/facebook/react/issues/15267.

necolas:
Closing as we're no longer developing the Flare API. See https://github.com/facebook/react/issues/15257#issuecomment-565408356

PR #15268, opened 2019-03-30
Adds displayName to EventComponent and EventTarget

trueadm:
Follow up to https://github.com/facebook/react/issues/15267. Adds the `displayName` field to EventComponent and EventTarget objects to provide this information from components using the experimental event API. This also adds support for the event API components showing in `getComponentName`.


gayathrisharma95:
On 30-Mar-2019 22:43, "Sunil Pai" <notifications@github.com> wrote:

*@threepointone* approved this pull request.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/facebook/react/pull/15268#pullrequestreview-220829013>,
or mute the thread

trueadm:
@gaearon I could add support for this in `getComponentName` too, but the names aren't internal to React and are user-defined and given the modules live in user-space, I thought this would be the simplest approach.

bvaughn:
> Any reason we don’t do it the way we normally do? By adding to getComponentName. Tools like DevTools can vendor that code.

The problem with this, as described in #15267, is that there isn't sufficient information for `EventComponent` tags to identify them any more specifically. This would hamstring DevTools ability to later show which events a component is subscribed to.

Technically we don't need the `displayName` attribute (or any change) for the `EventTarget` though, since we can uniquely identify that by the `elementType.type`.

trueadm:
Yep, as Brian said, the issue here is third party usage. I've updated the `getComponentName` implementation to understand the built-in `REACT_EVENT_TARGET_TOUCH_HIT`, otherwise it falls back to using `displayName` (which is now optional).

packages/shared/getComponentName.js
```diff
@@ -87,7 +91,19 @@ function getComponentName(type: mixed): string | null {
         if (resolvedThenable) {
           return getComponentName(resolvedThenable);
         }
+        break;
       }
+      case REACT_EVENT_COMPONENT_TYPE:
+        return ((type: any): ReactEventComponent).displayName;
+      case REACT_EVENT_TARGET_TYPE:
+        const eventTarget = ((type: any): ReactEventTarget);
+        if (eventTarget.type === REACT_EVENT_TARGET_TOUCH_HIT) {
```
bvaughn: How will this work? Hm... since we have a constant for our own, built-in touch target type- what value will users put there?

trueadm: For user created `EventTarget`s, they'd use their own symbol. That obviously won't be picked up by `getComponentName`, so they'd also need to supply a `displayName` field too.

bvaughn: Right, right. My question was unclear. I guess I was really asking- why we need to special case `REACT_EVENT_TARGET_TOUCH_HIT` (to treat it differently than other `REACT_EVENT_TARGET_TYPE`s) given that it wouldn't be possible for us to do a similar thing for custom target-types. I haven't thought a ton about the event system, just seemed like a potential hole.

I agree though that for display name purposes, using an optional `displayName` attribute should be fine.

trueadm: `REACT_EVENT_TARGET_TOUCH_HIT` has special properties that mean it gets handled by the ReactDOM internals differently, specifically that in this other PR to do with hit slop https://github.com/facebook/react/pull/15261.

bvaughn: Yeah, I saw that. (I searched the source for it before commenting above.) The question I was trying to ask was if we were comfortable that it's the only target type that will require such special handling. Sounds like the idea is to build onto the API if we find other cases in the future?

trueadm: Sorry, I should have asked if you had search the source before hand. I imagine we'd add more internal types as we go and as we need them and thus add more cases in `getComponentName`. They might also need special code paths in ReactDOM too, some may not.

To be honest, the previous change to this PR wasn't necessary, we could just use `displayName` in all cases and not have a special switch in `getComponentName`. I'm not too bothered on that. What do you think we should do?

trueadm:
I can confirm after testing locally that we want to keep the `getComponentName` logic. It does indeed help debugging component stack traces, otherwise you get `unknown` appear for the event API components.

packages/shared/getComponentName.js
```diff
@@ -87,7 +91,19 @@ function getComponentName(type: mixed): string | null {
         if (resolvedThenable) {
           return getComponentName(resolvedThenable);
         }
+        break;
       }
+      case REACT_EVENT_COMPONENT_TYPE:
+        return ((type: any): ReactEventComponent).displayName;
```
bvaughn: This should probably check for `undefined` before returning, just like the target.  (We probably don't want to _force_ custom event components to have `displayNames`.)

PR #15271, opened 2019-03-30
Update ReactDomEventListener

Imadra:
targetInst can be undefined, so check not if it is nonnull, but if it is truthy


trueadm:
Can you add a test case for where `getClosestInstanceFromNode` should return `undefined`. I believe the intention of `getClosestInstanceFromNode` was to always return `null` or a node so maybe this is a bug – thus a test would be good.

cpojer:
@Imadra can you add the test case that @trueadm was asking for?

PR #15272, opened 2019-03-30
Remove maxDuration

sebmarkbage:
We're going to replace it with a heuristic and different mechanism instead.

This PR just hard codes 150ms in place and fixes the tests. A follow up will refactor a bit more.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberUnwindWork.js
+++ b/packages/react-reconciler/src/ReactFiberUnwindWork.js
@@ -229,16 +229,12 @@ function throwException(
             break;
           }
         }
-        let timeoutPropMs = workInProgress.pendingProps.maxDuration;
-        if (typeof timeoutPropMs === 'number') {
-          if (timeoutPropMs <= 0) {
-            earliestTimeoutMs = 0;
-          } else if (
-            earliestTimeoutMs === -1 ||
-            timeoutPropMs < earliestTimeoutMs
-          ) {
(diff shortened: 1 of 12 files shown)


ghengeveld:
Could you explain what the heuristic and different mechanism will entail? I'm looking to integrate [React Async](https://react-async.dev) with Suspense, but it seems to still be in flux. It's hard to work towards compatibility with a moving target.

gaearon:
Agree it’s hard — which is why using Suspense for data fetching in Concurrent Mode is not marked as stable. If you’re doing it you’re kind of on your own. We’re still figuring out some things in the design.

gaearon:
More on what’s stable and what isn’t:

https://reactjs.org/blog/2018/11/27/react-16-roadmap.html

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -229,16 +229,12 @@ function throwException(
...
-        if (typeof timeoutPropMs === 'number') {
-          if (timeoutPropMs <= 0) {
-            earliestTimeoutMs = 0;
-          } else if (
-            earliestTimeoutMs === -1 ||
-            timeoutPropMs < earliestTimeoutMs
-          ) {
-            earliestTimeoutMs = timeoutPropMs;
-          }
+        const defaultSuspenseTimeout = 150;
```
NE-SmallTown: If it's a global const(config), why not lift it to the top?

PR #15273, opened 2019-03-30
Delete Suspense Fixture

sebmarkbage:
As far as I can tell this build is broken. I get a build error claiming invalid syntax (for Flow features). Let's fix or delete. If I can't test, I can't patch it up when I break it.


shellscape:
@gaearon reminder to add that back in. the 16.x roadmap points to the now-removed directory.

PR #15274, opened 2019-03-30
Remove Flow types from Scheduler and Tracing

sebmarkbage:
The idea is to make the scheduler as unopinionated and uncontroversial as possible.

Currently it's a bit of a mix and match and also use our build tooling. This keeps spreading.

Let's make a decision here either way and stick to it.

### DIFF
--- a/scripts/flow/environment.js
+++ b/scripts/flow/environment.js
@@ -41,3 +41,78 @@ declare module 'EventListener' {
     capture: (target: Element, type: string, callback: Function) => mixed,
   };
 }
+
+// Libdefs for Scheduler
+
+declare module 'scheduler/tracing' {
+  declare type Interaction = {|
+    __count: number,
+    id: number,
+    name: string,
+    timestamp: number,
+  |};
+
+  declare type Subscriber = {
+    // A new interaction has been created via the trace() method.
+    onInteractionTraced: (interaction: Interaction) => void,
+
+    // All scheduled async work for an interaction has finished.
+    onInteractionScheduledWorkCompleted: (interaction: Interaction) => void,
+
+    // New async work has been scheduled for a set of interactions.
+    // When this work is later run, onWorkStarted/onWorkStopped will be called.
+    // A batch of async/yieldy work may be scheduled multiple times before completing.
+    // In that case, onWorkScheduled may be called more than once before onWorkStopped.
+    // Work is scheduled by a "thread" which is identified by a unique ID.
+    onWorkScheduled: (interactions: Set<Interaction>, threadID: number) => void,
+
+    // A batch of scheduled work has been canceled.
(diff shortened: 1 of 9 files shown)


scripts/flow/environment.js
```diff
@@ -41,3 +41,78 @@ declare module 'EventListener' {
     capture: (target: Element, type: string, callback: Function) => mixed,
   };
 }
+
+// Libdefs for Scheduler
+
+declare module 'scheduler/tracing' {
```
sebmarkbage: This can go into flowtyped.

bvaughn:
Mixed feels about this.

acdlite:
I thought we agreed we’d make this decision after the API stabilizes and if/when we’re sure we want to push other frameworks (outside React ecosystem) to adopt it. Both are still up in the air.

bvaughn:
Agreed. Feels like removing the types now will just make the next few iterations a little more painful.

Usamaliaquat123:
@gaearon  please tell me how can i start contributing in react libr any docs thanks man

scripts/flow/environment.js
```diff
@@ -41,3 +41,78 @@ declare module 'EventListener' {
...
+
+  declare function unstable_getCurrent(): Set<Interaction> | null;
+
+  declare function unstable_getThreadID(): number;
+
+  declare function unstable_trace(
+    name: string,
+    timestamp: number,
+    callback: Function,
+    threadID: number,
```
eps1lon: This argument is optional in the implementation  https://github.com/facebook/react/blob/63f9f60376816bde524dcd15f0094cb8eb49188a/packages/scheduler/src/Tracing.js#L64 and in `unstable_wrap`. Not sure which one is right.

scripts/flow/environment.js
```diff
@@ -41,3 +41,78 @@ declare module 'EventListener' {
...
+    // In that case, onWorkStarted may be called more than once before onWorkStopped.
+    // Work is done by a "thread" which is identified by a unique ID.
+    onWorkStarted: (interactions: Set<Interaction>, threadID: number) => void,
+
+    // A batch of work has completed for a set of interactions.
+    // Work is done by a "thread" which is identified by a unique ID.
+    onWorkStopped: (interactions: Set<Interaction>, threadID: number) => void,
+  };
+
+  declare type InteractionsRef = {
```
eps1lon: This is typed like `enableSchedulerTracing` is `true` but `  declare function unstable_getCurrent(): Set<Interaction> | null;` is typed as `enableSchedulerTracing` is either `true` or `false`.

Should these types cover all builds of `scheduler/tracing` or only `tracing.development`. If it should cover all builds this should be nullable.

gaearon:
Stale.

PR #15283, opened 2019-04-01
Rename hover props in experimental event API and write unit tests

behzad888:
Note: this is for an experimental event API and writing tests.

 - `onHoverIn` -> `onHoverStart`
 - `onHoverOut` -> `onHoverEnd`
 - Write unit tests

### DIFF
--- a/packages/react-events/src/Hover.js
+++ b/packages/react-events/src/Hover.js
@@ -29,19 +29,19 @@ if (typeof window !== 'undefined' && window.PointerEvent === undefined) {
   targetEventTypes.push('touchstart', 'mouseover', 'mouseout');
 }
 
-function dispatchHoverInEvents(
+function dispatchHoverStartEvents(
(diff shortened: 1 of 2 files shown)


behzad888:
> Thank you for this contribution – it's awesome! It all looks good to me. I'm also a bit surprised to be honest that people externally are playing with this API. I'd be interested to know how it works out for you too. :)

LGTM.
I'm trying to prepare a blog about react experimental event API. In that, I want to how to use and how it works out.

gaearon:
You'll want to be *very* clear it's completely experimental and not really usable outside of this repository yet. We might scrap the whole thing altogether.

trueadm:
Like Dan said, this isn’t to be used right now and is likely to change in API frequently too. Furthermore, we might not even use it. I’d recommend holding off for now. :)

behzad888:
Ok. Thank you for your response.

PR #15288, opened 2019-04-01
Add Press event prop types and fix a check in Safari

necolas:
* Adds Press prop types
* Creates constants for default delay ms
* Fixes a mouse pointer check for Safari (no pointer events)

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -10,6 +10,11 @@
 import type {EventResponderContext} from 'events/EventTypes';
 import {REACT_EVENT_COMPONENT_TYPE} from 'shared/ReactSymbols';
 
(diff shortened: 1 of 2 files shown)


packages/react-events/src/Press.js
```diff
@@ -26,6 +31,23 @@ if (typeof window !== 'undefined' && window.PointerEvent === undefined) {
   rootEventTypes.push({name: 'mouseup', passive: false});
 }
 
+type PressProps = {
+  disabled: boolean,
+  delayLongPress: number,
+  delayPress: number,
+  delayPressEnd: number,
+  delayPressStart: number,
+  hitSlop: Object,
```
necolas: Remove this prop for now https://github.com/facebook/react/pull/15261

Issue #14254, opened 2018-11-16
React.lazy does not allow retrying a rejected promise


tnunes:
**Do you want to request a *feature* or report a *bug*?**

It can be seen as a feature or a bug, depending on angle. Let's say it's an enhancement to how `lazy` works.

**What is the current behavior?**

When using `React.lazy`, if the given promise rejects while trying to asynchronously load a component, it's no longer possible to retry loading the component chunk because `lazy` internally caches the promise rejection.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

This does not seem to work great in CodeSandbox because it's using service workers, which get in the way when simulating offline mode, yet this small app illustrates the issue: https://codesandbox.io/s/v8921j642l

**What is the expected behavior?**

A promise rejection should not be cached by `lazy` and another attempt to render the component should call the function again, giving it the chance to return a new promise.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

AFAIK all version of React that include `lazy`.

threepointone:
`React.lazy` accepts any function that returns a module, so you could do something like this - 

```jsx 
const Something = React.lazy(async () => {
  let mod;
  while(!mod){ 
   mod = await import('./something')
  }
  // this does infinite retries, 
  // you could modify it to do it only thrice, 
  // or add backoff/network detection logic, etc 
  return mod 
})
```
(I haven't tested the above, but I think it should work?)

tnunes:
@threepointone that would immediately retry loading a failed module until if eventually fails and gives up, which could address part of the problem.

The part we can't currently address unless we change `lazy` implementation is the scenario where we want to retry loading a previously failed async module load after the initial attempt(s) gave up.

Imagine an app code-split by route:

1. The user loads the initial route, navigates around while loading new routes on demand and then eventually goes offline, or has connectivity issues.
2. Any new route requested during this period will fail to load (which is expected) and we want to be able to gracefully show an error letting the user know we couldn't load it right now.
3. The user is free to continue using the application and navigate back to the routes that have already been loaded.
4. At a later time, if the user goes back to the previously failed route, we want to retry loading the module from the network and render it if successful.

[...]

PR #15296, opened 2019-04-02
Retry Lazy promise when rejected

varoot:
This solves #14254 by resetting `_status` on the lazy component if the promise is rejected.
This will cause the promise function to be re-triggered when the lazy component is re-rendered.


varoot:
The tests are passing on my machine but it seems to be failing on CircleCI. Does anyone know why?

leeseean:
are you kiding us?

varoot:
I fixed it now

Xesenix:
And what if someone wants to display message like there is connection problem with button that when you click retries?

PR #15297, opened 2019-04-02
writing unit tests in experimental event Drag API

behzad888:
Note: this is for an experimental event Drag API and writing tests.

 - Add `onDragStart` event
 - write unit tests for `onDragStart`, `onDragEnd`, `onDragChange` and `onDragMove`


trueadm:
This is good, but I feel we should probably get in https://github.com/facebook/react/pull/15295 first and then rebase this PR on top along with revise event Flow typings, rather than the other way around.

Issue #15301, opened 2019-04-03
renderWithHooks may initialize workInProgressHook at the beginning


Jokcy:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
example here https://codesandbox.io/s/34mvmoln65
Based on issue #15219 I found the reason of this issue is that, when `ErrorThrower` throws `renderWithHooks` just break down and keep `workInProgressHook` stay `useMemo`. Because of the `componentDidCatch` in `ErrorHandler` we can go on the render phase. But when we reach the `StatefulComponent` we got `useMemo` as `workInProgressHook` and it make `StatefulComponent` with first hook `useMemo`.

I guess we can fix this just set `workInProgressHook` to null at the beginning of `renderWithHooks`. I post this issue because I see this comment in `renderWithHooks`, based on this example, all this variables may also not be reset when start `renderWithHooks`, so maybe it's better to put them back?

```js
// The following should have already been reset
  // currentHook = null;
  // workInProgressHook = null;

  // remainingExpirationTime = NoWork;
  // componentUpdateQueue = null;

  // didScheduleRenderPhaseUpdate = false;
  // renderPhaseUpdates = null;
  // numberOfReRenders = 0;
  // sideEffectTag = 0;
```

If no other concerns about this, I just want to put them back to fix this issue.

gaearon:
As I wrote in https://github.com/facebook/react/pull/15302#pullrequestreview-222261586, test cases first would be very helpful!

PR #15302, opened 2019-04-03
fix 15301 15219

### LINKED ISSUE #15219: Combination of componentDidCatch and hooks throws "Should have a queue. This is likely a bug in React."

Bug

If a component with hooks throws an error in its render function *after* the hooks have been defined, and that render error is caught via a `componentDidCatch` in a parent component, any subsequent components will have their hook order jumbled up on the next render.

This results in the app crashing with an **"Uncaught Invariant Violation: Should have a queue. This is likely a bug in React. Please file an issue."** error (or different messages depending on the specific hooks used)

https://codesandbox.io/s/34mvmoln65

(once loaded, open the dev tools console and click the 'Trigger re-render' button)

Relevant source:

```jsx
function App(props) {
  const [, setCounter] = React.useState(0);
  return (
    <div>
      <ErrorHandler>
        <ErrorThrower />
      </ErrorHandler>
      <StatefulComponent />
      <button onClick={() => setCounter(value => value + 1)}>
        Trigger re-render
      </button>
    </div>
  );
}

function ErrorThrower() {
  React.useMemo(() => undefined, []);
  if (true) {
    throw new Error("!!!");
  }
  return <p>[Error component]</p>;
}

function StatefulComponent() {
  React.useState(null);
  return <p>[Stateful component]</p>;
}

class ErrorHandler extends React.Component {
  ...
  componentDidCatch(error) {
    ...
  }
}
```

The app should not crash, seeing as the `componentDidCatch()` ought to catch the render error and allow the rest of the app to render as normal

This will presumably affect all versions of React that include the current Hooks implementation (v16.8 onwards)

Jokcy:
fix #15219 and more info in #15301


Jokcy:
@gaearon It seems I ran into some issue with react-test-renderer. code here: https://codesandbox.io/s/r58yl52o74

I recently trying to test this issue with this code:

```js
```

Jokcy:
I also test the react-dom case in local with workInProgressHook reset at beginning and it will pass. Since I can't change node_modules code in codesandbox I can't show the example here.

Jokcy:
HA... I found this is not an issue of `renderWithHooks`😂. The reason of `workInProgressHook` not being reset is that React trying to `replayUnitOfWork` in development. 

Normally without `replayUnitOfWork` we catch error and `resetHooks` at `renderRoot`. But when replay we run the `ErrorThrow` component with `invokeGuardedCallback` which will not `resetHooks`. So~~~

I guess I need to change the commit content now.

Jokcy:
Maybe we just need to add `resetHooks` after replay?

Jokcy:
@gaearon I push this to add `resetHooks` after `replay`. I'm waiting for some discuss about how to test this case and also if we should make `replay` acting same in test-renderer  to test this case, since invokeGuardedCallback depend on dom event but test-renderer did not

Jokcy:
I don't understand why I just comment then it will close the issue...

PR #15307, opened 2019-04-03
mark react-events as private so the publish script skips it (for now)

threepointone:
we don't have this package just yet, and aren't ready to publish anything to it, so marking it as private so it doesn't interfere with our publish process.


ihsanali01 (commented):
Thanks

PR #15308, opened 2019-04-03
Add full TouchHitTarget hit slop (experimental event API) to ReactDOM

trueadm:
This PR is a complete replacement for the work suggested and originally started in https://github.com/facebook/react/pull/15261.

This PR aims add doing the following:

- Support ReactDOM rendering `<TouchHitTarget>` elements. These elements can have optional numerical and nullable `top`, `left`, `right` and `bottom` props.
- Support server-side rendering of these the hit targets along with their hit slop properties.
- Support hydration from server-side rendered content for these hit targets.
- Host configs now get additional methods: `getEventTargetChildElement` and `commitEventTarget`.
- Tests mainly covering ReactDOM/ReactDOMServer support.

Furthermore, I found that we can't actually test the hit slop logic in JSDOM because it doesn't support a bunch of the DOM APIs that `<TouchHitTarget>` and respective event responders use. We'll need to come up with a different strategy for that, maybe @necolas has some ideas.

Ref #15257

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -1341,10 +1341,10 @@ export function listenToEventResponderEventTypes(
 }
 
 export function createTouchHitTargetElement(
-  bottom: number | void | null,
-  left: number | void | null,
-  right: number | void | null,
-  top: number | void | null,
+  bottom: number,
+  left: number,
+  right: number,
+  top: number,
   rootContainerInstance: Element | Document,
   parentNamespace: string,
 ): HTMLElement {
@@ -1357,20 +1357,48 @@ export function createTouchHitTargetElement(
   const touchHitTargetElementStyle = touchHitTargetElement.style;
 
   touchHitTargetElementStyle.position = 'absolute';
-  if (top != null) {
+  if (top !== 0) {
     touchHitTargetElementStyle.top = `-${top}px`;
   }
-  if (left != null) {
+  if (left !== 0) {
     touchHitTargetElementStyle.left = `-${left}px`;
   }
-  if (right != null) {
+  if (right !== 0) {
     touchHitTargetElementStyle.right = `-${right}px`;
   }
-  if (bottom != null) {
+  if (bottom !== 0) {
     touchHitTargetElementStyle.bottom = `-${bottom}px`;
   }
 
   return touchHitTargetElement;
+}
+
+export function updateTouchHitTargetElement(
+  oldBottom: number,
+  oldLeft: number,
+  oldRight: number,
+  oldTop: number,
+  newBottom: number,
+  newLeft: number,
+  newRight: number,
+  newTop: number,
+  touchHitTargetElement: Element,
+) {
+  const touchHitTargetElementStyle = ((touchHitTargetElement: any): HTMLElement)
+    .style;
+
+  if (oldTop !== newTop) {
+    touchHitTargetElementStyle.top = `-${newTop}px`;
+  }
+  if (oldLeft !== newLeft) {
+    touchHitTargetElementStyle.left = `-${newLeft}px`;
+  }
+  if (oldRight !== newRight) {
+    touchHitTargetElementStyle.right = `-${newRight}px`;
+  }
+  if (oldBottom !== newBottom) {
+    touchHitTargetElementStyle.bottom = `-${newBottom}px`;
+  }
 }
 
 // We can remove this once the event API is stable and out of a flag
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -98,6 +98,7 @@ const AUTOFOCUS = 'autoFocus';
 const CHILDREN = 'children';
 const STYLE = 'style';
 const HTML = '__html';
+const EMPTY_OBJECT = {};
 
 const {html: HTML_NAMESPACE} = Namespaces;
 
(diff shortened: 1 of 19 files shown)


trueadm:
The bundle size is interesting, although the non-min size goes up from 105.45 KB -> 105.52 KB, gzip actually goes down from 33.59 KB -> 33.56 KB. I'll try and remove the non-gzip size though.

panergongzi:
so good

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1981,9 +1984,14 @@ function updateEventComponent(current, workInProgress, renderExpirationTime) {
...
 function updateEventTarget(current, workInProgress, renderExpirationTime) {
+  const type = workInProgress.type.type;
   const nextProps = workInProgress.pendingProps;
   let nextChildren = nextProps.children;
 
+  if (type === REACT_EVENT_TARGET_TOUCH_HIT && current === null) {
+    tryToClaimNextHydratableInstance(workInProgress);
+  }
+
   reconcileChildren(
```
sebmarkbage: What does it mean to pass children to an EventTarget in this new model? I thought these would just be leafy?

trueadm: I believe hydration tests failed unless I had this – was I wrong in that thinking? EventTargets could have children, for now there is only a single EventTarget and that is a TouchHitTarget, there will be many more in the future; some that can have children, some that cannot.

trueadm: Although thinking about it. We could make them leafy, but we’d need to have special cases for void host components:

<input><FocusTarget /></input>

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1339,6 +1340,67 @@ export function listenToEventResponderEventTypes(
...
+  const touchHitTargetElement = ((createElement(
+    'div',
+    EMPTY_OBJECT,
+    rootContainerInstance,
+    parentNamespace,
+  ): any): HTMLElement);
+  const touchHitTargetElementStyle = touchHitTargetElement.style;
+
+  touchHitTargetElementStyle.position = 'absolute';
+  if (top !== 0) {
```
necolas: We will always need to set each CSS position including for `0`. Here's an [example](https://codepen.io/anon/pen/GLorbm) that will break if `top:0` is removed from the CSS.

I also realised that the effective hitslop is reduced if a border is applied to the container, as the `0` position of the absolutely-position div is determined by the `padding-box`.

packages/react-events/src/__tests__/TouchHitTarget-test.internal.js
```diff
@@ -318,5 +324,315 @@ describe('TouchHitTarget', () => {
...
+        <EventComponent>
+          <div>
+            <TouchHitTarget top={10} left={10} right={10} bottom={10} />
+          </div>
+        </EventComponent>
+      );
+
+      let output = ReactDOMServer.renderToString(<Test />);
+      expect(output).toBe(
+        '<div><div style="position:absolute;-10px;-10px;-10px;-10px;"></div></div>',
```
necolas: The ReactDOMServer expected strings are missing CSS property names

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -1168,6 +1169,26 @@ class ReactDOMServerRenderer {
...
+                const props = nextElement.props;
+                const bottom = props.bottom || 0;
+                const left = props.left || 0;
+                const right = props.right || 0;
+                const top = props.top || 0;
+
+                if (bottom === 0 && left === 0 && right === 0 && top === 0) {
+                  return '';
+                }
+                let topString = top ? `-${top}px;` : '';
```
necolas: Missing the CSS property name in these strings

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1339,6 +1340,67 @@ export function listenToEventResponderEventTypes(
...
+): HTMLElement {
+  const touchHitTargetElement = ((createElement(
+    'div',
+    EMPTY_OBJECT,
+    rootContainerInstance,
+    parentNamespace,
+  ): any): HTMLElement);
+  const touchHitTargetElementStyle = touchHitTargetElement.style;
+
+  touchHitTargetElementStyle.position = 'absolute';
```
necolas: The element should also have `zIndex = -1` to make sure it never lies over other content in the container. But like `position` this means that the container needs to have its own `zIndex` set to create a relative stacking context, otherwise this element could fall behind the container.

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -1168,6 +1169,26 @@ class ReactDOMServerRenderer {
...
+                let leftString = left ? `-${left}px;` : '';
+                let rightString = right ? `-${right}px;` : '';
+                let bottomString = bottom ? `-${bottom}px;` : '';
+
+                return `<div style="position:absolute;${topString}${leftString}${rightString}${bottomString}"></div>`;
```
necolas: It's worth noting that this could result in an undesirable experience for mouse-based interactions if the client-side JS is available.

trueadm: Not sure how we tackle that one. Maybe @sebmarkbage had some ideas. I think you can maybe get away using some clever `pointer` media queries in a stylesheet.

packages/react-dom/src/client/ReactDOMComponent.js
trueadm: This breaks the hit target for me (I can no longer press it). Let's link up on this one next week.

trueadm:
Build is failing because of some strange `danger` script error. Seems unrelated to this PR.

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -110,7 +113,31 @@ const eventResponderContext: ResponderContext = {
...
+    }
+    const target = doc.elementFromPoint(x, y);
+    if (target === null) {
+      return false;
+    }
+    const childFiber = getClosestInstanceFromNode(target);
+    if (childFiber === null) {
+      return false;
+    }
+    const parentFiber = childFiber.return;
```
sebmarkbage: Normally we avoid traversing Fibers outside the reconciler but we already cheat this rule in the existing event system - for the same reasons as here. So this is ok.

PR #15310, opened 2019-04-03
Store entire build directory as CI artifacts

acdlite:
Updates the Circle CI config to store Facebook bundles as build artifacts. We already do this for our npm packages.

**Update:** Might as well store entire `build/` directory


acdlite:
Good call

acdlite:
Ok it now stores the `build/` directory as a tarball. The `node_modules` tarball used by the release script is unchanged.
Screen Shot 2019-04-03 at 4 37 40 PM

PR #15312, opened 2019-04-03
Alternate fix for "Fix a crash in Suspense with findDOMNode"

acdlite:
This doesn't rely on checking the tag. When the alternate of a parent is missing, it assumes it's a fragment indirection and moves onto the next parent fiber.

I based my commit on top of @gaearon's PR #15195.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberTreeReflection.js
+++ b/packages/react-reconciler/src/ReactFiberTreeReflection.js
@@ -21,6 +21,8 @@ import {
   HostRoot,
   HostPortal,
   HostText,
+  Fragment,
+  SuspenseComponent,
 } from 'shared/ReactWorkTags';
 import {NoEffect, Placement} from 'shared/ReactSideEffectTags';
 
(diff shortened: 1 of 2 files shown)


bvaughn:
Note to self: Assuming we land this, make sure we sync it to the new DevTools package too https://github.com/bvaughn/react-devtools-experimental/blob/07bf8e53c1e6fbf6503bebc170622ab161cf3b8f/src/backend/renderer.js#L1105-L1106

acdlite:
Oops forgot to run Flow

packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -117,11 +117,19 @@ export function findCurrentFiberUsingSlowPath(fiber: Fiber): Fiber | null {
   let b = alternate;
   while (true) {
     let parentA = a.return;
```
sebmarkbage: What if parentA here is a Suspense boundary with no fragment and its alternate is a Suspense boundary with a fragment?

In that case the children won't be representing the same thing. However, they also will never be equal so maybe it doesn't matter?

acdlite: That was my thinking, at least

acdlite:
I'll squash my commits (but leave @gaearon's separate) and land via command line so @gaearon also gets credit

bvaughn:
Thanks folks!

Synced to DevTools via https://github.com/bvaughn/react-devtools-experimental/commit/7fbdfd9

xiaoshuangLi:
I upgrade to react@16.9.0 and react-dom@16.9.0, but still get this error.

image

PR #15313, opened 2019-04-03
Match react-devtools display name behavior for React.memo

eps1lon:
```diff
// anonymous render
React.memo(() => null)
- Component
+ Memo
// named render
React.memo(function Inner() { return null })
- Inner
+ Memo(Inner)

```

Showcase to verify matching behavior with react-devtools: https://codesandbox.io/s/lpx7k7wv4z

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
@@ -867,6 +867,7 @@ describe('ReactLazy', () => {
       return props.inner + props.outer;
     };
     Add = React.memo(Add);
+    Add.displayName = 'Add';
     Add.propTypes = {
       inner: PropTypes.number.isRequired,
       innerWithDefault: PropTypes.number.isRequired,
@@ -882,6 +883,7 @@ describe('ReactLazy', () => {
       return props.inner + props.outer;
     };
     Add = React.memo(Add);
+    Add.displayName = 'Add';
     Add.propTypes = {
       inner: PropTypes.number.isRequired,
     };
@@ -945,7 +947,7 @@ describe('ReactLazy', () => {
     expect(() => {
       expect(Scheduler).toFlushAndYield(['Inner default text']);
     }).toErrorDev(
-      'The prop `text` is marked as required in `T`, but its value is `undefined`',
+      'The prop `text` is marked as required in `Memo(T)`, but its value is `undefined`',
     );
     expect(root).toMatchRenderedOutput('Inner default text');
 
@@ -958,7 +960,7 @@ describe('ReactLazy', () => {
       );
       expect(Scheduler).toFlushAndYield([null]);
     }).toErrorDev(
-      'The prop `text` is marked as required in `T`, but its value is `null`',
+      'The prop `text` is marked as required in `Memo(T)`, but its value is `null`',
     );
     expect(root).toMatchRenderedOutput(null);
   });
@@ -1050,6 +1052,7 @@ describe('ReactLazy', () => {
     Add.defaultProps = {
       inner: 2,
     };
+    Add.displayName = 'Add';
     const LazyAdd = lazy(() => fakeImport(Add));
     const root = ReactTestRenderer.create(
       <Suspense fallback={<Text text="Loading..." />}>
--- a/packages/react-reconciler/src/__tests__/ReactMemo-test.js
+++ b/packages/react-reconciler/src/__tests__/ReactMemo-test.js
@@ -365,7 +365,7 @@ describe('memo', () => {
           expect(Scheduler).toFlushWithoutYielding();
         }).toErrorDev(
           // Outer props are checked in createElement
-          'Invalid prop `outer` of type `string` supplied to `FnInner`, expected `number`.',
+          'Invalid prop `outer` of type `string` supplied to `Memo(FnInner)`, expected `number`.',
         );
 
         // Update
@@ -374,7 +374,7 @@ describe('memo', () => {
           expect(Scheduler).toFlushWithoutYielding();
         }).toErrorDev(
           // Outer props are checked in createElement
-          'Invalid prop `outer` of type `boolean` supplied to `FnInner`, expected `number`.',
+          'Invalid prop `outer` of type `boolean` supplied to `Memo(FnInner)`, expected `number`.',
         );
       });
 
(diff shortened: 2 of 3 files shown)


acdlite:
Oh haha I thought you were @bvaughn. Sorry :D If you want to change the signature of `getComponentType` then that's fine, but what you have is fine, too.

packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
```diff
@@ -792,6 +792,7 @@ describe('ReactLazy', () => {
     Add.defaultProps = {
       innerWithDefault: 42,
     };
+    Add.displayName = 'Add';
```
acdlite: Let's leave these tests as-is and write a separate one that tests `displayName`.

eps1lon: This is only there because the test function matches the display name. Since not every component is a memo component the test would fail.

eps1lon: i.e. this is not there to test `displayName` but to prevent refactoring the test.

eps1lon:
I think the devtools get the name from the fiber as well. If we can use a fiber instead we can we get rid of the component name fork in react-devtools as well?

packages/shared/getComponentName.js
```diff
@@ -34,10 +34,10 @@ function getWrappedName(
   innerType: any,
   wrapperName: string,
 ): string {
-  const functionName = innerType.displayName || innerType.name || '';
+  const innerName = getComponentName(innerType) || '';
```
eps1lon: This may be a bit overkill for render functions of `forwardRef`. We could extract the logic of `innerName` to the call-site of `getComponentName`.

packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
eps1lon: The alternative would be to use a wildcard matcher in https://github.com/facebook/react/blob/4aa22411f49d833589f38271ec4a4d800903114d/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js#L681
instead of expecting an exact component name.

eps1lon: > Let's leave these tests as-is and write a separate one that tests displayName.

Added an explicit test for memo and displayName in 012af538b6b532a8e2403c373b178e360e0f6791.

I could not use displayName in this test but then I would have to change the test to match `*` as the displayName (some components have the memo wrapper some don't). I don't think that's a good idea.

Not sure what else I should do here.

bvaughn:
> If we can use a fiber instead we can we get rid of the component name fork in react-devtools as well?

Not any time soon, since DevTools source lives in a different repo and (more importantly) needs to support multiple versions of React- unlike this help function, which only needs to support the current version at any given time.

packages/react-reconciler/src/__tests__/ReactIncrementalPerf-test.internal.js
```diff
@@ -584,6 +584,7 @@ describe('ReactDebugFiberPerf', () => {
       const MemoFoo = React.memo(function Foo() {
         return <div />;
       });
+      MemoFoo.displayName = 'Foo';
```
eps1lon: [`beginFiberMark`](https://github.com/facebook/react/blob/f3a14951ab9bccfd59ca977493b72321b24e50a5/packages/react-reconciler/src/ReactDebugFiberPerf.js#L122) will be called with the initial Fiber. However, in [`updateMemoComponent`](https://github.com/facebook/react/blob/4c78ac0b9df88edec73492f92f09d5438dd74c4d/packages/react-reconciler/src/ReactFiberBeginWork.js#L352) we will upgrade the Fiber to a `SimpleMemoComponent` at which point the Fiber is indistinguishable for `getComponentName` from a function component because with mutated the type. [`endFiberMark`](https://github.com/facebook/react/blob/f3a14951ab9bccfd59ca977493b72321b24e50a5/packages/react-reconciler/src/ReactDebugFiberPerf.js#L154) will receive a `SimpleMemoComponent` and `getComponentName` assumes it's a plain function component.

Not sure if this is considered a defect (assuming that the fiber type in `beginFiberWork` is the same we get in `endFiberWork`). Should we we add something `_debugOriginalType` field to the Fiber or rather change `react-devtools` to display `Foo` rather than `Memo(Foo)`? Mutating the `displayName` of the type when upgrading the Fiber might also work.

Any thoughts @bvaughn ?

packages/react-reconciler/src/ReactDebugFiberPerf.js
```diff
@@ -119,11 +120,19 @@ const getFiberLabel = (
   }
 };
 
+const getFiberName = (fiber: Fiber): string | null => {
+  if (fiber.tag === SimpleMemoComponent) {
+    return getComponentName(fiber.elementType);
```
eps1lon: Tested

```diff
-getComponentName(someFiber.type)
+getComponentName(someFiber.elementType)
```
on all source files in `master` and no test broke. This may be the better option if other types of Fibers change their `type` during work i.e. `getComponentName` should only be called with element types not fiber types (enforceable with types?).

eps1lon: #15638 would remove the necessity for this branch.

Airkro:
`forwardRef` should do the same.

eps1lon:
I'll resolve merge conflicts.

alflennik:
I just ran into this issue and I would love to see this PR make it in! I see that it's been sitting here for some time now.

eps1lon:
I'll resolve the conflicts though I don't think this will be considered from my experience.

eps1lon:
@bvaughn Could you take a look at this?

packages/react-reconciler/src/__tests__/ReactMemo-test.js
```diff
@@ -532,7 +532,7 @@ describe('memo', () => {
         ReactNoop.render(<MemoComponent optional="foo" />),
       ).toErrorDev(
         'Warning: Failed prop type: The prop `required` is marked as required in ' +
-          '`Foo`, but its value is `undefined`.\n' +
+          '`Bar`, but its value is `undefined`.\n' +
```
eps1lon, line 535: This test was added in #18925 but I don't think this should've been the expected behavior @bvaughn.

For `React.memo` we ignore `displayName` on the wrapper if we have a `displayName` on the wrapped component.
For `React.forwardRef` we honor `displayName` on the wrapper even if we have `displayName` on the wrapped component.

I think it is confusing that `React.memo` and `React.forwardRef` behave differently with regard to `displayName`.

PR #15322, opened 2019-04-04
Experimental event API: refactor responder modules for lifecycle inclusion

trueadm:
This PR moves around some parts and refactors the event responder system to cater for event responders having new lifecycle events (to be added in a follow up PR). The big difference is that `handleEvent` -> `onEvent` and it now has another argument, `event`.

I had feedback that it was confusing why the `context` object had so many concerns, so that's been taken care of now – it only deals with the context of the responder dealing with events, the actual event properties are in the `event` object.

I've also cleaned up and removed all the `prototype` methods and instead opted for static methods and using closures.

Ref #15257


packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -373,7 +164,9 @@ export function runResponderEventsInBatch(
   eventSystemFlags: EventSystemFlags,
 ): void {
   if (enableEventAPI) {
-    const context = new DOMEventResponderContext(
+    const currentEventResponderContext = ((eventResponderContext: any): Object);
+    currentEventResponderContext._eventQueue = createEventQueue();
+    const responerEvent = new DOMEventResponderEvent(
```
necolas: typo: responer -> responder

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -20,45 +21,237 @@ import type {
...
+  dispatchEvent(
+    possibleEventObject: Object,
+    {
+      capture,
+      discrete,
+      stopPropagation,
+    }: {
+      capture?: boolean,
+      discrete?: boolean,
+      stopPropagation?: boolean,
```
necolas: We could extract this to a named type

PR #15323, opened 2019-04-04
[React Native] Add tests to paper renderer for measure, measureLayout

TheSavior:
We will be changing these functions for Fabric so starting by adding some tests to validate the current behavior.

### DIFF
--- a/packages/react-native-renderer/src/__tests__/ReactNativeMount-test.internal.js
+++ b/packages/react-native-renderer/src/__tests__/ReactNativeMount-test.internal.js
@@ -249,6 +249,106 @@ describe('ReactNative', () => {
     });
   });
 
+  it('should call UIManager.measure on ref.measure', () => {
+    const View = createReactNativeComponentClass('RCTView', () => ({
+      validAttributes: {foo: true},
+      uiViewClassName: 'RCTView',
+    }));
+
+    class Subclass extends ReactNative.NativeComponent {
+      render() {
+        return <View>{this.props.children}</View>;
+      }
+    }
+
+    const CreateClass = createReactClass({
+      mixins: [NativeMethodsMixin],
+      render() {
+        return <View>{this.props.children}</View>;
+      },
+    });
+
+    [View, Subclass, CreateClass].forEach(Component => {
+      UIManager.measure.mockClear();
+
+      let viewRef;
+      ReactNative.render(
+        <Component
+          ref={ref => {
+            viewRef = ref;
+          }}
+        />,
+        11,
+      );
+
+      expect(UIManager.measure).not.toBeCalled();
+
+      const successCallback = jest.fn();
+      viewRef.measure(successCallback);
+
+      expect(UIManager.measure).toHaveBeenCalledTimes(1);
+      expect(UIManager.measure).toHaveBeenCalledWith(
+        expect.any(Number),
+        expect.any(Function),
+      );
+
+      const args = UIManager.measure.mock.calls[0];
+      expect(successCallback).not.toBeCalled();
+      args[1]('success');
+      expect(successCallback).toBeCalledWith('success');
+    });
+  });
+
+  it('should call UIManager.measureInWindow on ref.measureInWindow', () => {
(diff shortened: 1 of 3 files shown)


packages/react-native-renderer/src/__tests__/ReactNativeMount-test.internal.js
```diff
@@ -249,6 +249,106 @@ describe('ReactNative', () => {
...
+      expect(UIManager.measure).toHaveBeenCalledTimes(1);
+      expect(UIManager.measure).toHaveBeenCalledWith(
+        expect.any(Number),
+        expect.any(Function),
+      );
+
+      const args = UIManager.measure.mock.calls[0];
+      expect(successCallback).not.toBeCalled();
+      args[1]('success');
+      expect(successCallback).toBeCalledWith('success');
```
bvaughn: This seems a little janky too.

packages/react-native-renderer/src/__tests__/ReactNativeMount-test.internal.js
```diff
@@ -249,6 +249,106 @@ describe('ReactNative', () => {
...
+      expect(UIManager.measure).not.toBeCalled();
+
+      const successCallback = jest.fn();
+      viewRef.measure(successCallback);
+
+      expect(UIManager.measure).toHaveBeenCalledTimes(1);
+      expect(UIManager.measure).toHaveBeenCalledWith(
+        expect.any(Number),
+        expect.any(Function),
+      );
```
bvaughn: This seems a little janky. Like it's kind of testing internal API details.

matthargett: testing the module’s interactions with the public API of its collaborators, rather than the aggregate behavior of said multiple modules, is how you decouple a test like this from implementation details.

from Xunit patterns book:

> A Mock Object provides the system under test with both indirect input and a way to verify indirect output. All the expectations are configured before the calling of the method under test.

Asserting that no outbound calls to UIManager.measure happen before the viewRef.measure call is a good pinning test and test as documentation IMO.

packages/react-native-renderer/src/__tests__/ReactNativeMount-test.internal.js
matthargett: Here, I agree with you. The coupling is tight as a side effect of an array being used for the arguments instead of a hash with a named key. Given the design of the code, I’m not sure what else to do except make a test helper function that wraps this detail and give it an intention-revealing name.

packages/react-native-renderer/src/__tests__/ReactNativeMount-test.internal.js
bvaughn: I think my comments were  probably swapped.  _This_ seems janky and the other one seems like it's monkeying with internals a bit too much for my comfort.

packages/react-native-renderer/src/__tests__/ReactNativeMount-test.internal.js
TheSavior: The specific reason I wanted to test this is because one of the UIManager measure calls takes the arguments `fail, success`, but the host component takes the arguments `success, fail` and flips the order when passed to the UIManager. I wanted to validate that this stays the correct order because it is trivially easy to break (or think that the existing code is wrong).

I can't assert that the callback *is* the callback that is passed in because it is wrapped, so calling it and validating the argument felt like the most appropriate thing.

packages/react-native-renderer/src/__tests__/ReactNativeMount-test.internal.js
TheSavior: I'm going to change this to instead have a mock implementation that will throw if passed anything other than a number in the first argument and to call the passed callback with arguments. Then my assertion will just validate that the success callback was called with the expected arguments.

TheSavior:
Alright, I updated the tests to match the new style we talked about.

PR #15324, opened 2019-04-04
[React Native] measure calls will now call FabricUIManager

TheSavior:
This PR is based on https://github.com/facebook/react/pull/15323, split for review.

The Fabric renderer was previously calling the paper UIManager's measure calls and passing the react tag. This PR changes the renderer to now call FabricUIManager passing the node instead.

In #15126 Seb and I were thinking that we shouldn't support components created with NativeMethodsMixin and ReactNative.NativeComponent to be able to take a ref in measureLayout.

While that is still ideal, we can't reasonably make that change because the signatures of FabricHostComponent, FiberHostComponent, NativeMethodsMixin, and ReactNativeNativeComponent all have identical type signatures as enforced in each file with this:

```
(ReactFabricHostComponent.prototype: NativeMethodsMixinType);
```

Ideally we'd type host components differently than the other methods so that we can use flow to enforce that an argument is a ref to a host component and not one of the others. Unfortunately, we are pretty far from being able to do that today because all the core components in React Native are typed like this:

```
(requireNativeComponent('AndroidSwitch'): Class<ReactNative.NativeComponent<Props>>)
```

We will need to fix this and update all the ref types throughout the codebase to be more strict in order to properly type the argument to measureLayout as a ref to a host component. This is a much bigger project and out of scope of this PR. We should make that change at some point.

In order to keep all the types consistent in this PR measureLayout is taking a reactTag or object and the checks are done at runtime.

---------------

The additions to the FabricUIManager types are consistent with D14732142.


TheSavior:
I need to track down these lint errors:

```
/home/circleci/project/build/react-native/oss/ReactFabric-dev.js
  21229:9  error  'FabricUIManager__default' is not defined  no-undef
```

I think I'm probably missing some define somewhere.

bvaughn:
I think this is failing because the `FabricUIManager` module does not have a default export?

packages/react-native-renderer/src/ReactNativeTypes.js
```diff
@@ -89,7 +89,7 @@ class ReactNativeComponent<Props> extends React.Component<Props> {
   measure(callback: MeasureOnSuccessCallback): void {}
   measureInWindow(callback: MeasureInWindowOnSuccessCallback): void {}
   measureLayout(
-    relativeToNativeNode: number,
+    relativeToNativeNode: number | Object,
```
sebmarkbage: Let's make this `number | ReactNativeComponent` and cast it on the inside. It's not correct but more specific.

PR #15330, opened 2019-04-04
Experimental event API: Swipe module with tests

behzad888:
Note: this is for an experimental event Swip API and writing tests.

  - Determine event object data 
```
    pointerType: 'mouse' | 'pointer' | 'touch'
    initial: { x: number, y: number },
    point: { x: number, y: number },
    delta:  { x: number, y: number }  
    direction: 'up' | 'left' | 'down' | 'right',
```
  - Combine onSwipe{Left,Right,Up,Down} into onSwipe
  - Write unit tests 
    - `onSwipeStart`
    - `onSwipeEnd`
    - `onSwipe`
    - `Directions`
    - `nested responders`
    - `PointerType`

And about `pointerType` I think we should set it by `context`.

Ref #15257

### DIFF
--- a/packages/react-events/src/Swipe.js
+++ b/packages/react-events/src/Swipe.js
@@ -39,7 +39,7 @@ type EventData = {
   initial: Point,
   delta: Point,
   point: Point,
-  direction: SwipeDirection,
+  direction: null | SwipeDirection,
 };
 
 type SwipeEvent = {|
@@ -93,8 +93,7 @@ function dispatchSwipStartEvent(
     },
     initial: {x: state.startX, y: state.startY},
     point,
-    //TODO: fill pointerType
-    // pointerType: ... ,
+    pointerType: state.swipeTarget,
     direction: state.direction,
   };
   dispatchSwipeEvent(
@@ -120,8 +119,7 @@ function dispatchSwipeMoveEvent(
     },
     initial: {x: state.startX, y: state.startY},
     point,
-    //TODO: fill pointerType
-    // pointerType: ... ,
+    pointerType: state.swipeTarget,
     direction: state.direction,
   };
   dispatchSwipeEvent(
--- a/packages/react-events/src/Swipe.js
+++ b/packages/react-events/src/Swipe.js
@@ -24,17 +24,29 @@ if (typeof window !== 'undefined' && window.PointerEvent === undefined) {
 }
 
 type EventData = {
-  diffX: number,
-  diffY: number,
+  pointerType: PointerType,
+  delta: {
+    x: number,
+    y: numer,
+  },
+  direction: SwipeDirection,
 };
-type SwipeEventType = 'swipeleft' | 'swiperight' | 'swipeend' | 'swipemove';
+type PointerType = 'mouse' | 'pointer' | 'touch';
+
+type SwipeEventType = 'swipestart' | 'swipeend' | 'swipemove';
+
+type SwipeDirection = 'up' | 'left' | 'down' | 'right';
+
+type Point = {|
+  x: number,
+  y: number,
+|};
 
 type SwipeEvent = {|
   listener: SwipeEvent => void,
   target: Element | Document,
   type: SwipeEventType,
-  diffX?: number,
-  diffY?: number,
+  delta: Point,
 |};
 
 function createSwipeEvent(
@@ -64,10 +76,64 @@ function dispatchSwipeEvent(
   context.dispatchEvent(syntheticEvent, {discrete});
 }
 
+function dispatchSwipStartEvent(
+  context: ResponderContext,
+  props: Object,
+  state: SwipeState,
+  point: Point,
+) {
+  const eventData = {
+    delta: {
+      x: point.x - state.startX,
+      diffY: point.y - state.startY,
+    },
+    initial: {x: state.startX, y: state.startY},
+    point,
+    //TODO: fill pointerType
+    // pointerType: ... ,
+    direction: state.direction,
+  };
+  dispatchSwipeEvent(
+    context,
+    'swipestart',
+    props.onSwipeStart,
+    state,
+    true,
+    eventData,
+  );
+}
+
+function dispatchSwipeMoveEvent(
+  context: ResponderContext,
+  props: Object,
+  state: SwipeState,
+  point: Point,
+) {
+  const eventData = {
+    delta: {
+      x: point.x - state.startX,
+      diffY: point.y - state.startY,
+    },
+    initial: {x: state.startX, y: state.startY},
+    point,
+    //TODO: fill pointerType
+    // pointerType: ... ,
+    direction: state.direction,
+  };
+  dispatchSwipeEvent(
+    context,
+    'swipemove',
+    props.onSwipe,
+    state,
+    false,
+    eventData,
+  );
+}
+
 type SwipeState = {
-  direction: number,
+  direction: null | SwipeDirection,
   isSwiping: boolean,
-  lastDirection: number,
+  lastDirection: null | SwipeDirection,
   startX: number,
   startY: number,
   touchId: null | number,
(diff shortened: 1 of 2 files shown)


trueadm:
@behzad888 How do you mean by "And about pointerType I think we should set it by context.". Surely you should just set it on the Swipe event object that is created in the swipe responder module?

behzad888:
@trueadm 
I meant we have to check all events to set `pointerType`. Although, we can get that from `ResponderEvent`.

behzad888:
What's wrong with danger task?

packages/react-events/src/Swipe.js
```diff
@@ -23,18 +23,34 @@ if (typeof window !== 'undefined' && window.PointerEvent === undefined) {
   });
 }
 
+type PointerType = 'mouse' | 'pointer' | 'touch';
```
necolas: There is no `pointer` type. Did you mean `pen`?

packages/react-events/src/Swipe.js
```diff
@@ -64,10 +80,63 @@ function dispatchSwipeEvent(
   context.dispatchEvent(syntheticEvent, {discrete});
 }
 
+function dispatchSwipStartEvent(
```
necolas: typo: Swip -> Swipe

packages/react-events/src/Swipe.js
```diff
@@ -124,6 +194,17 @@ const SwipeResponder = {
             state.x = x;
             state.y = y;
             state.swipeTarget = target;
+            //use if statement to prevent Fallthrough warning
+            if (type === 'touchstart') {
+              state.pointerType = 'touch';
+            } else if (type === 'mousedown') {
+              state.pointerType = 'mouse';
+            } else if (type === 'pointerdown') {
+              state.pointerType = 'pointer';
```
necolas: this should be set to the value of `nativeEvent.pointerEvent`

packages/react-events/src/Swipe.js
```diff
@@ -159,26 +240,21 @@ const SwipeResponder = {
           }
           const x = (obj: any).screenX;
           const y = (obj: any).screenY;
-          if (x < state.x && props.onSwipeLeft) {
-            state.direction = 3;
-          } else if (x > state.x && props.onSwipeRight) {
-            state.direction = 1;
+          if (x < state.x) {
```
necolas: The logic here isn't quite right, as any vertical movement will cause the horizontal `direction` to be overwritten, even if the swipe is mainly along the horizontal axis. If a direction is to be reported, we need to determine whether it is "more" horizontal than vertical too.

packages/react-events/src/__tests__/Swipe-test.internal.js
```diff
@@ -0,0 +1,326 @@
...
+      const element = (
+        <Swipe onSwipeStart={onSwipeStart}>
+          <div ref={ref} />
+        </Swipe>
+      );
+      ReactDOM.render(element, container);
+    });
+
+    it('is called after "pointerdown" event', () => {
+      ref.current.dispatchEvent(createMouseEvent('pointerdown'));
```
necolas: This is dispatching a `MouseEvent`

packages/react-events/src/__tests__/Swipe-test.internal.js
```diff
@@ -0,0 +1,326 @@
...
+      const element = (
+        <Swipe onSwipe={onSwipe}>
+          <div ref={ref} />
+        </Swipe>
+      );
+      ReactDOM.render(element, container);
+    });
+
+    it('should be right', () => {
+      ref.current.dispatchEvent(createMouseEvent('mousedown'));
```
necolas: By default we should test against pointer events

packages/react-events/src/__tests__/Swipe-test.internal.js
```diff
@@ -0,0 +1,326 @@
...
+      ref.current.dispatchEvent(createMouseEvent('mousemove', 50, 0));
+      ref.current.dispatchEvent(createMouseEvent('mousemove', 60, 0));
+      ref.current.dispatchEvent(createMouseEvent('mouseup'));
+      expect(directions).toEqual(['left', 'right', 'right']);
+    });
+    it('directions in events in the correct order horizontal', () => {
+      ref.current.dispatchEvent(createMouseEvent('mousedown'));
+      ref.current.dispatchEvent(createMouseEvent('mousemove', 0, -50));
+      ref.current.dispatchEvent(createMouseEvent('mousemove', 0, 50));
+      ref.current.dispatchEvent(createMouseEvent('mousemove', 0, 60));
```
necolas: There are no tests that assert for events that have both a horizontal and vertical direction

packages/react-events/src/__tests__/Swipe-test.internal.js
```diff
@@ -0,0 +1,326 @@
...
+      ref.current.dispatchEvent(createMouseEvent('mouseup'));
+      expect(directions).toEqual(['down', 'up', 'up']);
+    });
+  });
+
+  describe('nested responders', () => {
```
necolas: @trueadm Nested swipe responders probably be a thing. Maybe swipe responders should always claim ownership?

packages/react-events/src/__tests__/Swipe-test.internal.js
```diff
@@ -0,0 +1,326 @@
...
+        'outer: onSwipeStart',
+        'inner: onSwipe',
+        'outer: onSwipe',
+        'inner: onSwipeEnd',
+        'outer: onSwipeEnd',
+      ]);
+    });
+  });
+
+  describe('PointerType', () => {
```
necolas: This isn't really a top-level test, as it can be folded into event object assertions for each of the callbacks

packages/react-events/src/Swipe.js
behzad888: please check new implement, thank you.

packages/react-events/src/__tests__/Swipe-test.internal.js
behzad888: based on https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events pointer event types are intentionally similar to mouse event types, however, PointerEvent API is not defined and document.createEvent('Event') deprecated and does not have screenX/Y option too. so, until finding a way to use Pointer Event API, I'm using MouseEvent.

packages/react-events/src/Swipe.js
behzad888: When a movement across the touch surface or screen should be considered a swipe. There are two variables at the distance traveled by the user's finger on the x or y-axis, and, the time it took. Based on these two factors, we can decide whether that action qualifies as a swipe and in what direction.

IMO we can have `distanceThreshold` props to weed out swipes and to be considered a swipe. 
For now, I'm using a default threshold with a value of 10 for sure. If you agree with me, let me do it.
what do you think?

packages/react-events/src/__tests__/Swipe-test.internal.js
behzad888: Attention, after write multiple touch identifier test units this would be necessary

packages/react-events/src/Swipe.js
```diff
@@ -159,26 +236,20 @@ const SwipeResponder = {
...
-            state.direction = 3;
-          } else if (x > state.x && props.onSwipeRight) {
-            state.direction = 1;
+          const distX = x - state.startX;
+          const distY = y - state.startY;
+
+          if (Math.abs(distX) >= DEFAULT_TRESHOLD_SWIP) {
+            state.direction = distX < 0 ? 'left' : 'right';
+          } else if (Math.abs(distY) >= DEFAULT_TRESHOLD_SWIP) {
+            state.direction = distY < 0 ? 'down' : 'up';
```
hartzis: currently, this logic "prefers" left/right swipes over up/down swipes?
Could/Should we check which abs is greater to determine the direction?
```
const absX = Math.abs(distX);
const absY = Math.abs(distY);
if (absX > DEFAULT_TRESHOLD_SWIP || absY > DEFAULT_TRESHOLD_SWIP) {
  if (absX > absY) {
    state.direction = distX < 0 ? 'left' : 'right';
  }
  state.direction = distY < 0 ? 'down' : 'up';
}
```

behzad888:
This could be close for now

PR #15334, opened 2019-04-05
ADD: disablePictureInPicture attribute for HTML5 videos

### LINKED ISSUE #15333: Support HTML5 Video Attribute disablePictureInPicture
Bug Report: `disablePictureInPicture` for html5 video elements are not supported

Eg:

```
    <video
    src="https://www.w3schools.com/tags/movie.mp4"
    disablePictureInPicture
    loop
    controls
    controlsList="nodownload"  
    ></video>
```

Should remove the 3 dots (more button) on videos on Chrome:

image

Instead, the attribute is not taken into consideration so the more menu stays:

image

Example:

https://jsfiddle.net/r1w4x7m0/2/

All / Latest, Chrome Browsers, All OSes.

Pull request for this issue has also been submitted: #15334

eek:
Fixes #15333

### DIFF
--- a/fixtures/attribute-behavior/src/attributes.js
+++ b/fixtures/attribute-behavior/src/attributes.js
@@ -451,7 +451,7 @@ const attributes = [
   {
     name: 'disablePictureInPicture',
     tagName: 'video',
-    read: getProperty('disablepictureinpicture')
+    read: getProperty('disablepictureinpicture'),
   },
   {
     name: 'display',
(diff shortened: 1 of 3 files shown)


eek:
Added a few changes @ahtee, it works properly.

I haven't added anything to AttributeTableSnapshot because I got `undefined` on all values, didn't know how to actually fix it.

image


One last thing I wasn't sure about was if I needed to add the `read` property in `attributes.js`: `read: getProperty('disablepictureinpicture')`

`autoPlay` has it 
but
`muted` doesn't have it

PR #15340, opened 2019-04-05
React events: add delay props to Press module

necolas:
* Adds support for `delayPressStart` and `delayPressEnd`.
* Adds tests for basic and integrated behaviour between delays and events.
* Updates documentation to include expected behaviour of `onPressStart` and `onPress` when used with delays.

Ref #15257

### DIFF
--- a/packages/react-events/README.md
+++ b/packages/react-events/README.md
@@ -3,6 +3,10 @@
 *This package is experimental. It is intended for use with the experimental React
 events API that is not available in open source builds.*
 
+Event components do not render a host node. They listen to native browser events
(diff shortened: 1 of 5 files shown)


packages/react-events/README.md
```diff
@@ -3,14 +3,31 @@
 *This package is experimental. It is intended for use with the experimental React
 events API that is not available in open source builds.*
 
+Event components do not render a host node. They listen to native browser events
+dispatched  on the host node of their child and transform those events into
```
NE-SmallTown: redundant space 😃

PR #15341, opened 2019-04-05
Profiler uses experimental Performance instructionCount API

bvaughn:
This PR changes the React `Profiler` to use an experimental `performance.instructionCount` API (instead of `performance.now`) if/when available. This change is gated behind a new `enableProfilerExperimentalInstructionCount` feature flag which has been turned off for open source builds (and is enabled only for Facebook builds).

This change has implications for the values passed to `Profiler` `onRender` (since these would be milliseconds _or_ instructions) but I think we can mostly ignore these, since it's highly unlikely the instruction count API would be enabled by default in Chrome for security reasons. Given that, it's likely that this data would only be used by e.g. companies like Facebook who have infra built up around it _or_ perhaps by developers launching Chromium with an opt-in flag (TBD). In either case, it's understood that the values being reported are _intentionally_ not timestamps.

Sebastian and I briefly discussed the possibility of just overriding `performance.now` to return this value so that the `Profiler` didn't need any special awareness of the new API, but I have concerns that (1) this would cause Profiler to be unable to track wall time and (2) potentially cause downstream breakage for e.g. throttling, debounce, etc.


sophiebits:
:o is this a new standard or just for browserlab/etc?

bvaughn:
This is something we're talking with the Chrome team about maybe adding behind an opt-in feature flag. (We're using it inside of FB at the moment as a more reliable way of detecting perf regressions.)

It has pretty exciting implications for open source people though! Assuming we could assign some "weight" to an "average" instruction, then we could pick targets/maxes for different classes of hardware, and Profiler could actually tell you "this commit is bad perf for X type of computer" etc.

Would be more actionable than what we're currently doing 😄

sophiebits:
This is interesting to me because in my experience instruction counts are (at least) an order of magnitude more stable than wall time for benchmarking. Most of the times I want this are locally, but if there's a way this could be on by default then it could also be useful for A/B testing performance, etc.

bvaughn:
The stability is why they're so appealing to use for perf regression testing. (Benoit has been looking into this within FB.)

Their loose couple with the specific hardware they happen to be running on is what makes me excited about using them in the Profiler (UI).

bvaughn:
Rebased and moved feature behind a flag.

NE-SmallTown:
Would you mind share any doc about this? I can't find something about it

bvaughn:
Not sure what the status on the instruction count stuff is, but @bgirard can ping me when we're ready to pick this up and actually land it.

PR #15343, opened 2019-04-06
sync expiration with scheduler

NE-SmallTown:


NE-SmallTown:
Seems doesn't work like the comment said, should I change the test or something else?

PR #15346, opened 2019-04-07
Experimental event API: Focus module

behzad888:
Note: 
determine eventData and pointerType 
```
{
  pointerType: 'mouse' | 'pen' | 'touch' | 'keyboard'
}
```

Ref #15257

### DIFF
--- a/packages/react-events/src/Focus.js
+++ b/packages/react-events/src/Focus.js
@@ -23,10 +23,17 @@ type FocusState = {
 
 type FocusEventType = 'focus' | 'blur' | 'focuschange';
 
+type PointerType = 'mouse' | 'pen' | 'touch' | 'keyboard';
+
+type EventData = {
+  pointerType: PointerType,
+};
+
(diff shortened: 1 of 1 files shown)


matthargett:
What about ‘gaze’ for AR/VR?l, and directional pad for TV remote/game controller?

necolas:
We're still figuring things out. Not sure it's worth your time submitting PRs for experimental APIs that are in flux.

jhampton:
> What about ‘gaze’ for AR/VR?l, and directional pad for TV remote/game controller?

Agreed @matthargett .  At the risk of being a pedant, perhaps a `dwell` or `hover` could be an analog for the same intent?

necolas:
Thanks for taking interest in this work. But I'm closing this PR as master has diverged quite a lot and we're still figuring out event data internally. This approach doesn't work either as native `focus`/`blur` events do not contain `pointerType` information on the event object.

PR #15347, opened 2019-04-07
Experimental event API: Hover module eventdata

behzad888:
Note:
determine eventData and pointerType
```
{
  pointerType: 'mouse',
  point: { x: number, y: number }
}
```
Ref #15257


matthargett:
what about gaze and pen?

necolas:
Thanks for taking interest in this work. But I'm closing this PR as master has diverged quite a lot and we're still figuring out event data internally.

PR #15358, opened 2019-04-09
Track Event Time as the Start Time for Suspense

sebmarkbage:
The new suspense model is based on the last "event time" that influenced the suspended render.

The "event time" is the time of the "update" (root render, setState, dispatch) - which we infer from its expiration time.

If this is a "retry", then there is no "update" and the event time is the time we committed the fallback. This event time is only accounted for if we actually switch from fallback to content, i.e. make progress. If we retry and don't make further progress, then this event time isn't accounted for.

To make the math simple and fast, I model the committed time as an "normal pri expiration time" at the time it committed.

Since there can be multiple updates and retries in the same batch, we use the most recent time for our heuristic. That ensures that we at least never wait longer than we otherwise would've.

Since we can't tell the difference between committed and suspended updates, this PR will error on the side of including committed updates.

For now, I still use the hard coded 150ms suspense time starting from the event time. In a follow up, I'll add the Just Noticeable Difference heuristic instead.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1275,13 +1275,13 @@ function commitSuspenseComponent(finishedWork: Fiber) {
   } else {
     newDidTimeout = true;
     primaryChildParent = finishedWork.child;
-    if (newState.timedOutAt === NoWork) {
+    if (newState.fallbackExpirationTime === NoWork) {
       // If the children had not already timed out, record the time.
       // This is used to compute the elapsed time during subsequent
       // attempts to render the children.
       // We model this as a normal pri expiration time since that's
       // how we infer start time for updates.
-      newState.timedOutAt = computeAsyncExpirationNoBucket(
+      newState.fallbackExpirationTime = computeAsyncExpirationNoBucket(
         requestCurrentTime(),
       );
     }
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -63,7 +63,10 @@ import invariant from 'shared/invariant';
 import warningWithoutStack from 'shared/warningWithoutStack';
 import warning from 'shared/warning';
 
-import {NoWork} from './ReactFiberExpirationTime';
+import {
+  NoWork,
+  computeAsyncExpirationNoBucket,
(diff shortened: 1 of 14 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1273,11 +1276,15 @@ function commitSuspenseComponent(finishedWork: Fiber) {
...
     primaryChildParent = finishedWork.child;
-    if (newState.timedOutAt === NoWork) {
+    if (newState.fallbackUpdate === NoWork) {
       // If the children had not already timed out, record the time.
       // This is used to compute the elapsed time during subsequent
       // attempts to render the children.
-      newState.timedOutAt = requestCurrentTime();
+      // We model this as a normal pri expiration time since that's
+      // how we infer start time for updates.
+      newState.fallbackUpdate = computeAsyncExpirationNoBucket(
```
acdlite: Naming nit: `fallbackTime`?

packages/react-reconciler/src/ReactFiberScheduler.new.js
```diff
@@ -878,26 +882,30 @@ function renderRoot(
...
+        }
+        // If workInProgressRootMostRecentEventTime is Sync, that means we didn't
+        // track any event times. That can happen if we retried but nothing switched
+        // from fallback to content. There's no reason to delay doing no work.
+        if (workInProgressRootMostRecentEventTime !== Sync) {
+          const msUntilTimeout = computeMsUntilTimeout(
+            workInProgressRootMostRecentEventTime,
           );
-          return null;
+          if (msUntilTimeout > 0) {
```
acdlite: I think we can get rid of this check now that we have the outer check? `computeMsUntilTimeout` will always return a positive value. (It returns 0 sometimes now, but it won't once we account for JND.)

sebmarkbage: It still returns 0 for disableYielding atm but that should probably be hoisted out. Can move it in the JND follow up.

packages/react-reconciler/src/ReactFiberCommitWork.js
sebmarkbage: I went with `fallbackExpirationTime`. I can't think of anything good. It's as if I made an update to switch to the content state at this time.

PR #15363, opened 2019-04-09
Include full error messages in React Native build

acdlite:
The React Native build does not minify error messages in production, but it still needs to run the error messages transform to compile `invariant` calls to `ReactError`. To do this, I added a `noMinify` option to the Babel plugin. I also renamed it from `minify-error-messages` to the more generic `transform-error-messages`.

### DIFF
--- a/scripts/error-codes/__tests__/__snapshots__/transform-error-messages.js.snap
+++ b/scripts/error-codes/__tests__/__snapshots__/transform-error-messages.js.snap
@@ -94,3 +94,14 @@ import invariant from 'shared/invariant';
   }
 })();"
 `;
+
+exports[`error transform should support noMinify option 1`] = `
+"import _ReactError from 'shared/ReactError';
+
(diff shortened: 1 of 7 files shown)


scripts/error-codes/__tests__/__snapshots__/transform-error-messages.js.snap
```diff
@@ -94,3 +94,14 @@ import invariant from 'shared/invariant';
   }
 })();"
 `;
+
+exports[`error transform should support noMinify option 1`] = `
+"import _ReactError from 'shared/ReactError';
+
+import invariant from 'shared/invariant';
```
TheSavior: Is this supposed to remove the import or does some other dead code elimination step do that?

acdlite: Yeah that happens in a subsequent step of the Rollup script

PR #15364, opened 2019-04-10
Move EventTypes to ReactTypes

trueadm:
This PR moves in the Flow type annotations from `events/EventTypes` to `shared/ReactTypes` to enable the RN build to properly consume these types without dependency misalignment.


packages/shared/ReactTypes.js
```diff
@@ -128,3 +126,47 @@ export type ReactEventTarget = {|
   displayName?: string,
   type: Symbol | number,
 |};
+
+type AnyNativeEvent = Event | KeyboardEvent | MouseEvent | Touch;
+
+export type ReactResponderEvent = {
+  nativeEvent: AnyNativeEvent,
+  target: Element | Document,
```
TheSavior: Are these global flow types? Will flow pass with these in fbsource/React Native?

trueadm: These are globals for Flow. We can revise them when we come to the RN work to add support for these events, but this should pass CI now.

PR #15366, opened 2019-04-10
Include new scheduler in test renderer www build

acdlite:
Changes `enableNewScheduler` from a static flag to a dynamic one in www's build of test renderer, so we can run the tests in either mode.


acdlite:
Decided to land the whole thing instead

PR #15367, opened 2019-04-10
Apply the Just Noticeable Difference to suspense timeouts

sebmarkbage:
Increase the suspended time based on how long we've already waited.

The theory is that a person [can't tell the difference between small differences in time](https://en.wikipedia.org/wiki/Just-noticeable_difference). Therefore, if we wait a bit longer than necessary that won't translate to a noticeable negative difference in the experience. However, waiting for longer might mean that we can avoid showing an intermediate loading state which is a positive difference. The longer we have already waited, the harder it is to tell small differences in time. Therefore, the longer we've already waited, the longer we can wait additionally. At some point we have to give up though. We pick a train model where the next boundary commits at a consistent schedule.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.new.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.new.js
@@ -160,6 +160,8 @@ import {
 } from 'shared/ReactErrorUtils';
 import {onCommitRoot} from './ReactFiberDevToolsHook';
 
+const ceil = Math.ceil;
+
 const {
   ReactCurrentDispatcher,
   ReactCurrentOwner,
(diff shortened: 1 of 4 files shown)


sebmarkbage:
The problem with this PR is that since we currently estimate the event time of "hi pri" updates as 5 seconds back in time. We think that we've waited a long time so if the hi-pri update suspends we'll wait a long time. Ideally we'd estimate these times correctly, but another approach is that we cap the timeout at the expiration time like we used it.

packages/react-reconciler/src/ReactFiberScheduler.new.js
```diff
@@ -1815,7 +1819,35 @@ export function resolveRetryThenable(boundaryFiber: Fiber, thenable: Thenable) {
   retryTimedOutBoundary(boundaryFiber);
 }
 
-function computeMsUntilTimeout(mostRecentEventTime: ExpirationTime) {
+// Computes the next Just Noticeable Difference (JND) boundary.
+// The theory is that a person can't tell the difference between small differences in time.
+// Therefore, if we wait a bit longer than necessary that won't translate to a noticeable
+// difference in the experience. However, waiting for longer might mean that we can avoid
```
NE-SmallTown: can -> can't?

PR #15374, opened 2019-04-10
React events: add onPressMove and pressRetentionOffset to Press

necolas:
This implementation differs from equivalents in React Native in the following ways:

1. A move during a press will not cancel onLongPress.
2. A move to outside the retention target will cancel all pending events and not reactivate when moved back within the retention target.

We can change this in the future but I think we'll want to refactor the state machine first.

Ref #15257

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -21,8 +21,14 @@ type PressProps = {
   onPress: (e: PressEvent) => void,
   onPressChange: boolean => void,
   onPressEnd: (e: PressEvent) => void,
+  onPressMove: (e: PressEvent) => void,
   onPressStart: (e: PressEvent) => void,
-  pressRententionOffset: Object,
+  pressRetentionOffset: {
+    top: number,
+    right: number,
+    bottom: number,
+    left: number,
+  },
 };
 
 type PressState = {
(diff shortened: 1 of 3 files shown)


packages/react-events/src/Press.js
```diff
@@ -67,13 +87,18 @@ const targetEventTypes = [
...
 
 // If PointerEvents is not supported (e.g., Safari), also listen to touch and mouse events.
 if (typeof window !== 'undefined' && window.PointerEvent === undefined) {
-  targetEventTypes.push('touchstart', 'touchend', 'mousedown', 'touchcancel');
-  rootEventTypes.push({name: 'mouseup', passive: false});
+  targetEventTypes.push('touchstart', 'touchend', 'touchcancel', 'mousedown');
+  rootEventTypes.push(
+    {name: 'mouseup', passive: false},
+    'touchmove',
+    'mousemove',
```
trueadm: These should probably be passive events.

packages/react-events/src/Press.js
```diff
@@ -67,13 +87,18 @@ const targetEventTypes = [
 const rootEventTypes = [
   {name: 'keyup', passive: false},
   {name: 'pointerup', passive: false},
+  'pointermove',
```
trueadm: This should probably be passive too.

packages/react-events/src/Press.js
necolas: Oh it's not passive by default?

PR #15381, opened 2019-04-11
Experimental Event API: Remove TouchHitTarget SSR logic to prevent issues with mouse events

trueadm:
This PR removes affects the experimental event API and its behaviour when server side rendering `TouchHitTarget`s. Specifically, it changes the rendering so that hit slop elements are no longer serialized in SSR output. Instead we rely the hydration logic for the `TouchHitTarget` to add in the hit slop element to the client. This change comes after internal discussions with @sebmarkbage and @necolas, where we agreed this was the right change for now; we can revisit this again in the future if needed.

This change fixes mouse events incorrectly firing before hydration has occurred. For example, if you have a `<button>` that had a `TouchHitTarget` on it, it would render with an inner absolutely positioned `<div>` that "expanded" the hit boundaries of its parent `<button>`. We then handle mouse and touch events in the event responders, allowing touch events to use the expanded boundary as intended and filtering out mouse events (as they shouldn't use the expanded boundaries). Unfortunately, as this logic of filtering only gets initialized upon hydration of a component, the incorrect behaviour happens with mouse events on SSR content before this fix.

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRenderer.js
+++ b/packages/react-dom/src/server/ReactPartialRenderer.js
@@ -1173,24 +1173,15 @@ class ReactDOMServerRenderer {
                 elementType.$$typeof === REACT_EVENT_TARGET_TYPE &&
                 elementType.type === REACT_EVENT_TARGET_TOUCH_HIT
               ) {
-                const props = nextElement.props;
-                const bottom = props.bottom || 0;
-                const left = props.left || 0;
-                const right = props.right || 0;
-                const top = props.top || 0;
-
-                if (bottom === 0 && left === 0 && right === 0 && top === 0) {
-                  return '';
-                }
-                let topString = top ? `-${top}px` : '0px';
-                let leftString = left ? `-${left}px` : '0px';
-                let rightString = right ? `-${right}px` : '0x';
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -1173,24 +1173,15 @@ class ReactDOMServerRenderer {
...
-                );
+                // We do not render a hit slop element anymore. Instead we rely
+                // on hydration adding in the hit slop element. The previous
+                // logic had a bug where rendering a hit slop at SSR meant that
+                // mouse events incorrectly registered events on the hit slop
+                // even though it designed to be used for touch events only.
+                // The logic that filters out mouse events from the hit slop
+                // is handled in event responder modules, which only get
+                // initialized upon hydration.
+                return '';
```
sebmarkbage: Won't this cause a hydration warning now? Can you add back tests to ensure that SSR+hydration of a hit target actually still works?

This can also be problematic when this isn't the very last child of the parent. If there is a next sibling that's a div `<Parent><HitTarget /><div /></Parent>`, then we'll try to hydrate that sibling as the hit target instead of this one.

That's why I expected you to leave the div in place but without the top/left/bottom/right styles and then only add the styles in when we commit.

PR #15384, opened 2019-04-11
Experimental Event API: adds `stopPropagation` by default to Press

trueadm:
This PR adds the default behaviour of stopping propagation on Press events, with the ability to override that feature using the `stopPropagation` prop. I also had to refactor the event propagation system to account for async event contexts, which means aligning `context.setTimeout` with `context.clearTimeout` so we can share actual timeouts where they have matching delays within the same context.

### DIFF
--- a/packages/react-events/src/__tests__/Press-test.internal.js
+++ b/packages/react-events/src/__tests__/Press-test.internal.js
@@ -381,6 +381,38 @@ describe('Event responder: Press', () => {
       expect(onPressChange).toHaveBeenCalledTimes(2);
       expect(onPressChange).toHaveBeenCalledWith(false);
     });
+
+    it('is called but does not bubble', () => {
+      const element = (
+        <Press onPressChange={onPressChange}>
+          <Press onPressChange={onPressChange}>
+            <div ref={ref} />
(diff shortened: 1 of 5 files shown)


packages/react-events/src/__tests__/Press-test.internal.js
```diff
@@ -477,6 +539,38 @@ describe('Event responder: Press', () => {
       expect(onLongPress).not.toBeCalled();
     });
 
+    it('is called but does not bubble', () => {
```
necolas: Let's perform all the bubbling tests in the "nested responders" block

trueadm: I was going to remove the nested responders block, as I thought it would be best to be about type of event. :/ I don't mind though.

necolas: Sure. But it seems we could test the bubbling/not-bubbling for all events in a single test.

trueadm: I'll move in a follow up tomorrow, will need some re-organizing of the test file.

PR #15385, opened 2019-04-11
Ensure TouchHitTarget element is server side rendered with hit slop

trueadm:
This is a follow up to #15381, specifically @sebmarkbage's comment regarding the mismatch in the missing hit target element and the usage of `pointer-events: none`. This now re-applies the hit slop element and its hit slop properties, with the addition of making `pointer-events: none` so they get excluded in hit target calculations till after hydration has completed. This should hopefully reduce the amount of reflows to the DOM too.

To get hydration updates for `TouchHitTarget` a special `hydrateTouchHitTarget` prop gets passed around, otherwise ReactDOM doesn't correctly set `pointer-events` to `auto` after.


packages/react-events/src/__tests__/TouchHitTarget-test.internal.js
```diff
@@ -330,7 +330,7 @@ describe('TouchHitTarget', () => {
       ReactDOM.render(<Test />, container);
       expect(Scheduler).toFlushWithoutYielding();
       expect(container.innerHTML).toBe(
-        '<div><div style="position: absolute; z-index: -1; bottom: -10px; ' +
+        '<div><div style="position: absolute; z-index: -1; pointer-events: auto; bottom: -10px; ' +
```
necolas: Can we exclude this style altogether when hydrating?

sebmarkbage: The question is what is the least likely to cause us to invalidate the layout. This needs testing but I was hoping that the pointer events flip would be the least likely to cause massive invalidation when we hydrate it.

trueadm: I can take a look again but this was done for a specific reason at the time.

trueadm: Fixed. I realized we can set `pointer-events` to `null` and it will remove the `none` value from SSR. :)

trueadm:
@necolas @sebmarkbage Would be good to get another review on this PR when you both have time.

PR #15386, opened 2019-04-11
React events: Press event fixes

necolas:
1. Fix hiding context menu for longpress via touch.
2. Fix scrolling of viewport for longpress via spacebar key.
3. Add tests for anchor-related behaviour and preventDefault.
4. Add a deactivation delay for forced activation
5. Add `pointerType` to Press events.

NOTE: this currently extends `pointerType` with non-standard types. We could consider using a different name, like `inputMode` or something.

NOTE: React Native doesn't have a deactivation delay for forced activation, but this is possibly because of the async bridge meaning that the events aren't dispatched sync.

Ref #15257

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -354,16 +354,17 @@ function calculateResponderRegion(target, props) {
 }
 
 function getPointerType(nativeEvent: any) {
-  if (nativeEvent.pointerType != null) {
-    return nativeEvent.pointerType;
+  const {type, pointerType} = nativeEvent;
+  if (pointerType != null) {
+    return pointerType;
   }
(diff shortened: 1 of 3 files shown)


packages/react-events/src/Press.js
```diff
@@ -325,6 +340,22 @@ function calculateResponderRegion(target, props) {
   };
 }
 
+function getPointerType(nativeEvent: any) {
+  if (nativeEvent.pointerType != null) {
+    return nativeEvent.pointerType;
+  }
+  if (nativeEvent.type.indexOf('mouse') > -1) {
```
trueadm: Could `type` be pulled out to its own variable? Seems like a lot of property accesses each time otherwise and should reduce prod code size.

PR #15387, opened 2019-04-11
Replace old Fiber Scheduler with new one

acdlite:
The new Fiber Scheduler has been running in Facebook for several days without issues. Let's switch to it.

- Remove `enableNewScheduler` feature flag
- Rename `ReactFiberScheduler.new.js` to `ReactFiberScheduler.js`
- Remove `react-dom/unstable-new-scheduler` build (and corresponding www build)
- Delete dead code
  - `ReactFiberScheduler.old.js`
  - `ReactFiberPendingPriority.js`, which was only called by the old scheduler
  - Tests that checked `enableNewScheduler` flag


MuYunyun:
Is the new implementation of Fiber Scheduler can be used independently?  I'm looking forward for using it in some day!:smile:

PR #15388, opened 2019-04-11
React events: add onHoverMove support

necolas:
Adds `onHoverMove`, related test, and documentation.

Ref #15257

### DIFF
--- a/packages/react-events/src/Hover.js
+++ b/packages/react-events/src/Hover.js
@@ -19,6 +19,7 @@ type HoverProps = {
   delayHoverStart: number,
   onHoverChange: boolean => void,
   onHoverEnd: (e: HoverEvent) => void,
+  onHoverMove: (e: HoverEvent) => void,
   onHoverStart: (e: HoverEvent) => void,
 };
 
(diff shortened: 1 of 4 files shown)


packages/react-events/src/Hover.js
```diff
@@ -265,16 +277,26 @@ const HoverResponder = {
...
+              dispatchHoverEndEvents(event, context, props, state);
+              state.isInHitSlop = true;
+            } else {
+              if (props.onHoverMove) {
+                const syntheticEvent = createHoverEvent(
+                  'hovermove',
+                  event.target,
+                  props.onHoverMove,
+                );
+                context.dispatchEvent(syntheticEvent, {discrete: true});
```
trueadm: Hover move shouldn't be discrete. Come to think of it, is press move discrete? We should also ensure that is not either.

thers:
Is it really OK that it does not respect the shape of an element? Suppose we have an element with `border-radius: 50%`, `mouseover` event respect it's shape, as in does not fire when mouse is within bounding rectangle but not over content inside border. Current behavior will probably lead to confusion of developers that this particular event does not take into account shape of an element.

PR #15391, opened 2019-04-11
Experimental Event API: Remove "listener" from event objects

trueadm:
This PR removes the logic for attaching a `listener` field to event objects, which leaks to the user function. Instead `dispatchEvent` now takes a listener argument instead. I also tidied up the resetting of `current` context values to avoid memory leaks and added an `invariant` to deal with edge cases at runtime.


packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -589,3 +613,11 @@ export function unmountEventResponder(
     ownershipChangeListeners.delete(eventComponentInstance);
   }
 }
+
+function validateResponderContext(): void {
+  invariant(
+    currentEventQueue && currentInstance,
+    'An event responder context was used outside of an event cycle. ' +
+      'Use context.setTimeout() to use asynchronous responder context outside of event cycle .',
```
necolas: 👍

PR #15393, opened 2019-04-11
Fix CI

acdlite:
CI is currently failing. I think it's because I pushed a commit directly to master without first opening a PR. Generally we shouldn't do this, but sometimes it makes sense, like if you want to merge a branch without squashing the commits.

I'll use this branch to try out some strategies for fixing it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,4 +47,7 @@ jobs:
           path: ./build.tgz
 
       - store_artifacts:
+          path: ./scripts/rollup/results.json
+
+      - store_artifacts:
           path: ./scripts/error-codes/codes.json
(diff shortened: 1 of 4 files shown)


acdlite:
Ok this is what I would expect to see:

Screen Shot 2019-04-11 at 4 31 37 PM

Subsequent PRs will fetch the results from CircleCI.

acdlite:
[build_14405_step_106_container_0.txt](https://github.com/facebook/react/files/3071001/build_14405_step_106_container_0.txt)

It fails because the base commit is broken... I suppose I should make that a warn instead :D

acdlite:
Ok I think this will fix master. I'll merge and see.

PR #15395, opened 2019-04-11
Bump scheduler version to 0.14.0

acdlite:
Releasing this early so React Native can depend on it


packages/scheduler/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "scheduler",
-  "version": "0.13.6",
+  "version": "0.14.0",
```
serenissimus: build failed

PR #15399, opened 2019-04-12

threepointone:
via #15319 

So tl;dr  this PR - 

- uses a renderer specific object (specifically, the `flushPassiveEffects` function itself) as the value of `ReactShouldWarnActingUpdates.current`
- checks this value on 'initial' render 
- checks this value when setting a state hook's value
- adds a fixture folder for `act()`

- using the 'wrong' `act()` shouldn't silence the warning:
  We do this by using a sigil unique to the renderer (as an elegant hack, we use `flushPassiveEffects` as the sigil, removing the need to make a new object). `act()` sets this object before the callback, and unsets it when it's over. Then in addition to `warnIfNotCurrentlyActingUpdatesInDev()`, we also check `warnIfNotScopedWithMatchingAct()` to confirm that`ReactShouldWarnActingUpdates.current` matches the expected sigil.

- using the wrong `act()` logs a warning:
  Using the same above method, we can check whether you're using the right version of act() for your code.

Now, I first added this check only for state hook updates, but it didn't reliably catch the common failure case. Consider the following component -

```jsx
function App() {
  let [state, setState] = useState(0);
  async function ticker() {
    await null;
    setState(x => x + 1);
  }
  useEffect(() => {
    ticker();
  }, [Math.min(state, 4)]);
  return state;
}
```

Let's write a test for it using the shiny new async `act()`

```jsx
const el = document.createElement("div");
await ReactTestUtils.act(async () => {
  ReactDOM.render(React.createElement(App), el);
});
// all 5 ticks present and accounted for
console.log(el.innerHTML); // 5!
```

This is the golden path - use the correct act with the matching renderer, and you'll get expected behaviour. Here's a quick diagram of the timeline it goes through

image

Of note, because we can check every time after calling `flushPassiveEffects()`, we can guarantee that the `act()` 'scope' will stay open until the effects queue is drained. Good.

Now, let's use a mismatching `act()`-

```jsx
await ReactTestUtils.act(async () => {
  ReactTestRenderer.create(React.createElement(App));
});
```

Now, let's say we'd added our sigil check _only for updates_, you'd think it would still trigger the warning. However, the timing of things has changed. There are 2 scenarios of how these will be sequenced out. The first, which is the 'good' version -

image

Because we can't use `flushPassiveEffects()` as expected (ie - it's just a no op for other renderer instances), we have to rely on the browser/jest environment to flush to the 'screen', and then the effects/updates fire. In this 'good' case, at least one set state call happens inside the act scope, so we can do the sigil check and warn that they aren't using the right act version.

However, the bad news is that this happens super rarely (in my rough estimation, only 1 in 20 'successes'). In reality, it usually happens outside the scope of the `act()` scope.

image

Here, you'll see that the effects fire after the `act()` scope has closed, so when we do the sigil check, we can only warn that the dev hasn't wrapped their code with `act()` (which will confuse them, since they think they already have)

So how do we fix this? Well, I noticed that most of these failures happen at the very start, when they initialise (`TestRenderer.create()`, `ReactDOM.render()`, etc). I believe that if we also add just the sigil identity check in the reconciler's `.createContainer()`, we should be able to warn _for most cases_ asap that they're using the wrong version (and ofc, `createContainer` is synchronous and won't escape the `act()` scope).

A quirk with our codebase is we don't allow multiple renderers in the same test, so I made a fixture folder `fixtures/act` to test these scenarios (see index.test.js for details). I used react-dom / react-test-renderer for the tests but I should probably add react-art as well. 

Questions - 

- In a previous iteration of this PR, I made a separate new object as the sigil, but then replaced it with using `flushPassiveEffects` as the sigil itself. ok/not ok?


threepointone:
hmm @sizebot you have my attention

acdlite:
@threepointone sizebot compares to the nearest common ancestor of master and your branch (https://github.com/facebook/react/commit/e5c59359c4ccbae31756db4c6f45f2a56cbe2a47). Because you've merged from master several times since then, the size reduction includes stuff that has already landed. This is one of the reasons I prefer to rebase upstream changes into my branch instead of merging them.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -2077,11 +2078,38 @@ function warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber) {
...
+    ) {
+      // it looks like we're using the wrong matching act(), so log a warning
+      warningWithoutStack(
+        false,
+        "It looks like you're using the wrong act() around your interactions.\n" +
+          'Be sure to use the matching version of act() corresponding to your renderer. e.g. -\n' +
+          "for react-dom, import {act} from 'react-test-utils';\n" +
+          'for react-test-renderer, const {act} = TestRenderer.' +
+          '%s',
+        getStackByFiberInDevAndProd(fiber),
```
threepointone: this doesn't work as 'expected', should probably leave out the component stack here.

threepointone:
Open questions - 
[ ] this adds a new export onto react-dom's secret export. ok/not ok?
[ ] I need to add some tests for the art/dom combo, and I'd be curious to see how nested updates would be 'solved'. 
[ ] I dunno, better warnings maybe? I'll take another spin at it later today.

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
```diff
@@ -61,18 +62,18 @@ function flushEffectsAndMicroTasks(onDone: (err: ?Error) => void) {
...
+    ReactShouldWarnActingUpdates.current = ReactActingUpdatesSigil;
   }
 
   function onDone() {
     if (__DEV__) {
       actingUpdatesScopeDepth--;
-      if (actingUpdatesScopeDepth === 0) {
-        ReactShouldWarnActingUpdates.current = false;
-      }
+      ReactShouldWarnActingUpdates.current = previousActingUpdatesSigil;
```
gaearon: What happens if we throw in the middle? Does this get restored?

threepointone: yup, `onDone()` gets called for errors too, and passes the error upwards
https://github.com/facebook/react/blob/7caa1b2a0ef1d935971edeb202d776fbefd1ffa2/packages/react-dom/src/test-utils/ReactTestUtilsAct.js#L154-L155
https://github.com/facebook/react/blob/7caa1b2a0ef1d935971edeb202d776fbefd1ffa2/packages/react-dom/src/test-utils/ReactTestUtilsAct.js#L133-L134

threepointone:
What if I used flushPassiveEffects as the sigil value? 🤔 it would remove the need for the extra exports.

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
```diff
@@ -103,9 +103,11 @@ function act(callback: () => Thenable) {
             if (called === false) {
               warningWithoutStack(
                 null,
-                'You called act(async () => ...) without await. ' +
+                'You called act(async () => ...) without awaiting its result. ' +
                   'This could lead to unexpected testing behaviour, interleaving multiple act ' +
-                  'calls and mixing their scopes. You should - await act(async () => ...);',
+                  'calls and mixing their scopes. You should await asynchronous act() ' +
+                  'calls, like so -\n' +
```
gaearon: We don't really use hyphens at the end of a sentence like this. Let's replace ", like so -" with ":"?

packages/react/src/ReactShouldWarnActingUpdates.js
```diff
@@ -0,0 +1,16 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * Used by act() to track whether you're outside an act() scope.
+ * This is an object so we can track identity of the renderer.
+ */
+
+const ReactShouldWarnActingUpdates = {
+  current: (null: null | {}), // {} should probably be an opaque type
```
threepointone: hrmm this should change to the signature for flushPassiveEffects, but why didn't flow complain?

fixtures/act/package.json
```diff
@@ -0,0 +1,27 @@
+{
```
acdlite: Is there a reason this needs to be a separate fixture package, instead of adding it to the DOM fixtures one? It would be nice if we could start consolidating our fixtures into as few as possible.

packages/react-test-renderer/src/ReactTestRendererAct.js
```diff
@@ -141,7 +144,8 @@ function act(callback: () => Thenable) {
         if (__DEV__) {
           warningWithoutStack(
             false,
-            'Do not await the result of calling act(...) with sync logic, it is not a Promise.',
+            'Do not await the result of calling a synchronous act(...), it is not a Promise. \n' +
```
acdlite: Do we even need this warning anymore?

acdlite: Also I would have done this in a separate PR to avoid giving me the opportunity to nitpick about it in this one :D

packages/react-test-renderer/src/ReactTestRendererAct.js
```diff
@@ -42,18 +42,20 @@ function flushEffectsAndMicroTasks(onDone: (err: ?Error) => void) {
...
   let previousActingUpdatesScopeDepth;
+  let previousActingUpdatesSigil;
   if (__DEV__) {
     previousActingUpdatesScopeDepth = actingUpdatesScopeDepth;
+    previousActingUpdatesSigil = ReactShouldWarnActingUpdates.current;
     actingUpdatesScopeDepth++;
-    ReactShouldWarnActingUpdates.current = true;
+    // we use the function flushPassiveEffects directly as the sigil,
+    // since it's unique to a renderer
+    ReactShouldWarnActingUpdates.current = flushPassiveEffects;
```
acdlite: Let's use an empty object that's local to this module instead.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -2077,23 +2077,58 @@ function warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber) {
   }
 }
 
+export function warnIfNotScopedWithMatchingAct(fiber: Fiber): void {
```
acdlite: This warning should live next to the rest of the `act` warning logic.

packages/react-test-renderer/src/ReactTestRendererAct.js
eps1lon: [Using await for a non-promise should be fine](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#Description). If there is some other issue with this it should be included in the warning message. In its current form it is misleading.

packages/react/src/ReactShouldWarnActingUpdates.js
threepointone: Yeah I already trued that locally, but it didn't warn. I'll keep poking at it.

fixtures/act/package.json
threepointone: No real reason, I'll merge it with the dom fixture

packages/react-test-renderer/src/ReactTestRendererAct.js
threepointone: ReactFiberScheduler wouldn't be able to read it when it needs to compare though? in `warnIfNotCurrentlyActingUpdatesInDev` 
https://github.com/facebook/react/blob/7f3fd8f8667f75821e02037aa3c229fd942baf39/packages/react-reconciler/src/ReactFiberScheduler.js#L2086

packages/react-reconciler/src/ReactFiberScheduler.js
threepointone: This needs to be accessible from 2 modules - ReactFiberHooks (while dispatching a state update) and ReactFiberReconciler (when creating a new container).

packages/react-test-renderer/src/ReactTestRendererAct.js
threepointone: What would the behaviour be if `await`ed? Should it run flushPassiveEffectsAndMicrotasks? I... guess it should?

fixtures/act/package.json
threepointone: hrmm actually, I needed this specifically to set up a test environment differently from the dom fixture. the dom fixture uses built versions via public, while the act one needs it in node_modules. open to suggestions, but I just tried and it's pretty annoying. I expect this to also host some stuff I plan to do to test www style setups too. would like to keep it.

packages/react-test-renderer/src/ReactTestRendererAct.js
threepointone: In any case, this shouldn't block this PR, I'll do it in a followup PR if needed, and if critical for 16.9, fair?

threepointone: This is a little harder than I expected. sync act() cleans up on exiting, so if any updates are fired on awaiting a sync act, they'll trigger the missing act() warning. using the async () => version seems cleanest.

fixtures/act/package.json
acdlite: Mainly I would like it to be the same package with the same package.json and node_modules. Ideally it would be a single app and single command that runs both, but even if they are separate commands that's better than nothing. Historically what happens is people will add these new fixtures and then nobody else opens them until somebody happens to notice that they are broken, by which time the original person has either lost context on what the desired behavior is, or they left the team entirely. The easier it is to run these fixtures, the more likely people will, I think.

fixtures/act/src/index.test.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import TestUtils from 'react-dom/test-utils';
+import TestRenderer from 'react-test-renderer';
```
acdlite: Do these fixture tests run when you run `yarn test`? If not, why are they in here? Can we move them to `packages/`?

packages/react-reconciler/src/ReactFiberScheduler.js
acdlite: Ah I misunderstood what this warning was. I got confused and thought it was about warning when you interleave multiple acts.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -2077,23 +2077,58 @@ function warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber) {
...
+// so we have a a step counter for when we descend/ascend from
+// act() calls, and test on it for when to warn
+// It's a tuple with a single value. Look into ReactTestUtilsAct as an
+// example of how we change the value
+
 function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
   if (__DEV__) {
     if (
       workPhase === NotWorking &&
-      ReactShouldWarnActingUpdates.current === false
```
acdlite: Oops, confusing that it used to warn if `ReactShouldWarnActingUpdates` was `false` :D

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -2077,23 +2077,58 @@ function warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber) {
...
+      ReactShouldWarnActingUpdates.current !== flushPassiveEffects
```
acdlite: Can we change this name to something more appropriate for how it's now used? `ReactActingRendererSigil`, perhaps.

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -276,7 +277,16 @@ export function createContainer(
   isConcurrent: boolean,
   hydrate: boolean,
 ): OpaqueRoot {
-  return createFiberRoot(containerInfo, isConcurrent, hydrate);
+  const fiberRoot = createFiberRoot(containerInfo, isConcurrent, hydrate);
+  if (__DEV__) {
+    // jest isn't a 'global', it's just exposed to tests via a wrapped function
+    // further, this isn't a test file, so flow doesn't recognize the symbol. So...
+    // $FlowExpectedError - because requirements don't give a damn about your type sigs.
+    if ('undefined' !== typeof jest) {
```
acdlite: Off-topic: Is there a task somewhere to support other test frameworks besides Jest? What's the plan?

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -276,7 +277,16 @@ export function createContainer(
...
+      warnIfNotScopedWithMatchingAct(fiberRoot.current);
```
acdlite: Why does this warning fire but not the other one?

fixtures/act/src/index.test.js
gaearon: Our setup doesn't support multiple renderers in the same file without also resetting shared React module state (which would break the warning).

packages/react-reconciler/src/ReactFiberScheduler.js
threepointone: that's fair, will do

packages/react-reconciler/src/ReactFiberReconciler.js
threepointone: we only discussed it offhand. filed https://github.com/facebook/react/issues/15439

packages/react-reconciler/src/ReactFiberReconciler.js
threepointone: I don't understand the question? In case a create and update are both wrapped by the wrong act(), then this warning will fire twice. But as explained in the description, the second one is less likely to fire, especially when async. We fire warnings for every violation.

fixtures/act/src/index.test.js
threepointone: ^ hence this fixture. open to ideas, but this seemed cleanest (for something that's probably an edge case)

packages/react-test-renderer/src/ReactTestRendererAct.js
threepointone: I started work on this in a separate branch, but it's tricky so far. awaiting inserts one microtask and could miss an update. failing code here rn https://github.com/threepointone/react/tree/await-sync-act will keep poking at it separately.

fixtures/act/package.json
threepointone: alright I updated the PR to merge the act/dom folders. someone still has to run yarn test in the dom folder, but at least it's all in one place

packages/react-reconciler/src/ReactFiberScheduler.js
threepointone: done

fixtures/act/src/index.test.js
acdlite: Sounds like a problem with the jest config. At the very least can we run this in CI? I will never remember to run this command manually.

packages/react-reconciler/src/ReactFiberReconciler.js
acdlite: `createContainer` isn't the right place for it because it's not an update. It's often accompanied by an update  (`ReactDOM.render` and `TestRenderer.create` call both `createContainer` and `updateContainer` one after the other) but our newer APIs have shifted to a model where you create a root before scheduling updates on in, e.g. `ReactDOM.createRoot`, which creates a root but doesn't schedule any work.

But I think there's a higher level problem...

> But as explained in the description, the second one is less likely to fire, especially when async. We fire warnings for every violation.

I did read your description, and I think I understood most of it, but I don't know if I agree with your root cause analysis.

Is it correct to rephrase the problem like this? APIs that schedule an update but existed before hooks — `this.setState`, `ReactDOM.render`, `TestRenderer.create` — don't trigger a warning if you don't wrap them in `act`, because we didn't want to start firing warnings in old tests. However, these types of updates could still result in passive effects, and if you don't wrap them in `act`, the effects will silently remain unflushed without a warning.

So it seems to me that the solution is to warn if a passive effect is scheduled outside of `act`.

acdlite:
> Because we can't use flushPassiveEffects() as expected (ie - it's just a no op for other renderer instances), we have to rely on the browser/jest environment to flush to the 'screen', and then the effects/updates fire.

Something to keep in mind is that passive effects are scheduled with Scheduler, and our recommendation will be to use the mock Scheduler build in tests. Which means effects that aren't wrapped by `act` will never be flushed; you can't rely on the "browser/jest environment" to eventually flush them in a subsequent event.

acdlite:
^ I believe the reason you didn't encounter this problem in your tests is because the separate fixture test suite you created doesn't use the mock Scheduler build.

packages/react-reconciler/src/ReactFiberReconciler.js
sebmarkbage: I don't know about the suggested solution. That's a bit controversial.

However, this is clearly not the right place for this warning since creating a root doesn't mean I will render it yet. It just means I have a reference to it for when I will.

threepointone: - How about in updateContainer then?
- I could also add it explicitly for each renderer's 'render' - so ReactDOM.render(), root.render(), TestRenderer.create()/.update(), etc. This will also mean I'll have to export the warning on the reconciler.

acdlite: Let's continue discussion here: https://github.com/facebook/react/issues/15472

fixtures/act/src/index.test.js
gaearon: Not sure what you mean by a problem with jest config. It doesn’t support so intentionally because of how we mock the inlined host configs.

threepointone: I've updated the PR to run these tests on ci

packages/react-reconciler/src/ReactFiberReconciler.js
threepointone: I updated the PR to do the check inside updateContainer(), looks good

threepointone:
@acdlite how would you feel about landing this before the `flushAll()` work anyway? It's useful enough that I want to make further PRs on top of it I think. I could make another branch off this but ehhh.

fixtures/act/package.json
threepointone: this now runs in ci

packages/react-reconciler/src/ReactFiberReconciler.js
threepointone: also a PR for when effects get queued outside an act https://github.com/facebook/react/pull/15631

threepointone:
bump

acdlite:
I can't review this until you resolve the conflicts

threepointone:
abandoning this for #15756

PR #15401, opened 2019-04-12
Double compress using GCC

sebmarkbage:
Let's see what sizebot says about this.

### DIFF
--- a/scripts/rollup/plugins/closure-plugin.js
+++ b/scripts/rollup/plugins/closure-plugin.js
@@ -28,8 +28,13 @@ module.exports = function closure(flags = {}) {
       flags = Object.assign({}, flags, {js: tempPath});
       await writeFileAsync(tempPath, code, 'utf8');
       const compiledCode = await compile(flags);
+      // Write the result
+      await writeFileAsync(tempPath, compiledCode, 'utf8');
+      // Compile again a second pass to ensure we've done as much inlining
+      // as we can.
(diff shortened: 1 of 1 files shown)


sebmarkbage:
ffs

acdlite:
Well there goes my theory that CircleCI would be more reliable than Paul's server :D  https://14437-10270250-gh.circle-artifacts.com/3/home/circleci/project/scripts/rollup/results.json

sebmarkbage:
Locally it seems dom prod min CJS goes from 106606 -> 106547 (34083 -> 34006 gzip). Not much, but the question still remains if it's worth it due to any future feature flags we might add, and whether it impacts runtime perf.

trueadm:
@sebmarkbage There's also the case that it will probably break the readability of the FB prod bundles. What do you they look like after? Also what impact is there on the bundle build times? It's already pretty slow.

threepointone:
Closing since we're not doing this (Please reopen if I'm wrong)

gaearon:
I think we could? If it saves size. Just need to compare the output manually.

sebmarkbage:
We probably shouldn't do this for FB bundles since we expect a second pass by our bundler anyway and we already compromise if that one is not smart enough. Basically, we could solve it by compiling it again in FB infra. I guess the file size there is just noisy variables.

There's still an increase in the "react" package which is interesting to look at.

threepointone:
omg 🤪

PR #15403, opened 2019-04-12
Update README.md

blackwolf08:
ES6-ES7 should be emphasised for future proofing Docs


blackwolf08:
completed my CLA

threepointone:
I don't think we want to do this unless we do it for all files (and probably not even then). We use https://reactjs.org for all public facing documentation anyway. But thank you!

PR #15404, opened 2019-04-12
Stop tracking bundle sizes

acdlite:
Deletes `scripts/rollup/results.json` from the repo and adds the path `.gitignore`. After running the build script, CI will move the file to `build/` before storing as an artifact. This is an extra precaution to ensure that sizebot compares against the correct sizes.


acdlite:
Looks like it worked:

Screen Shot 2019-04-12 at 1 32 40 PM

PR #15413, opened 2019-04-14
setValueForStyles error handling

Imadra:
I believe that function setValueForStyles throws incomprehensible error on some browsers because of stylename being undefined on these browsers. I added an error message for this.


sophiebits:
I don't think this is right so I'm going to close, but maybe if you provide more details about what you ran into then someone can help you better.

PR #15423, opened 2019-04-15
Disable Just Noticeable Difference heuristic

acdlite:
c25c59c808d90bfa78c40d0a19527d384a490cc8 resulted in bugs during internal testing. I don't know why yet, but removing the Just Noticeable Difference heuristic was sufficient to fix/hide the bugs again. In the interest of fixing this downstream quickly, let's revert it and land it again once we've figured out the root cause.

Specially, what I've done is this PR is revert `computeMsUntilTimeout` to what is what in the last known non-buggy commit: https://github.com/facebook/react/blob/3e2e930d62ac7166695e0c5b1538ebb7a41cd3da/packages/react-reconciler/src/ReactFiberScheduler.new.js#L1818-L1833

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
@@ -1568,7 +1568,7 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     );
   });
 
-  it('suspends for longer if something took a long (CPU bound) time to render', async () => {
+  it.skip('suspends for longer if something took a long (CPU bound) time to render', async () => {
     function Foo() {
       Scheduler.yieldValue('Foo');
       return (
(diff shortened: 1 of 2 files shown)


acdlite:
Oh wait this isn't right. One sec.

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1724,11 +1724,3 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     expect(ReactNoop.getChildren()).toEqual([span('Loading...')]);
   });
 });
-
-// TODO:
```
acdlite: This is unrelated; deleting because I noticed it and they are outdated.

acdlite:
Closing because I found a proper fix. Will open a new PR.

acdlite:
Actual fix landed in: https://github.com/facebook/react/pull/15424

PR #15424, opened 2019-04-15
Cancel pending commit before starting on root

acdlite:
Moves the cancelTimeout call to right before creating a new work-in-progress root. Fixes a class of bugs where a pending commit is not cancelled, causing an incomplete tree to accidentally commit.

In the interest of fixing downstream bugs quickly, I'm landing this without a test case; I'll add one in a follow up.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
@@ -1724,11 +1724,3 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     expect(ReactNoop.getChildren()).toEqual([span('Loading...')]);
   });
 });
-
-// TODO:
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1724,11 +1724,3 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     expect(ReactNoop.getChildren()).toEqual([span('Loading...')]);
   });
 });
-
-// TODO:
```
acdlite: This is unrelated. I happened to notice them now. Deleting these todos because they are outdated.

acdlite:
Got a verbal approval from @bvaughn to land this as he was packing up to leave

blling:
What does ’“leave” mean ?

bvaughn:
Sometimes after working 8 or 9 hours, I go home to rest.

Issue #15418, opened 2019-04-15
Field type="email" with multiple attribute cursor jumps to start


jordyvg:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
While adding multiple comma separated email addresses the cursor will jump to te beginning the input field. First i thought it was a [Formik](https://github.com/jaredpalmer/formik/issues/1428) issue, but it seems to be a React one.

Example pure React: https://codesandbox.io/embed/0y06zo7l8p
Example React with Formik: https://codesandbox.io/embed/k57zw7wz5
MDN docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email#multiple

**What is the expected behavior?**
Cursor stays on current position.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
Chrome 73, Safari 12.1, Opera 60.0
React 16.7.0

Maybe #14551 related?

Nizarius:
May I take this bug for PR?

jordyvg:
@Nizarius for sure

justingish:
Any progress update on this? I see that a PR was opened months ago, but it's still open. I'm still seeing this behavior today.

karthikpriyadarshan-alation:
The PR is still open. Can I work on this?

Nizarius:
@karthikpriyadarshan-alation thank you for your answer! Do you want to work with my PR where it's fixed but needs a regression fixture to https://github.com/facebook/react/tree/master/fixtures/dom and a detailed test plan. I can give you rights to work with my repository.

gaearon:
See https://github.com/facebook/react/pull/18459. We should fix it but I think this needs to wait a bit more.

josepharhar:
I made a fix in chrome which should fix this problem which should be released in chrome 85, try it out on canary: https://www.google.com/chrome/canary/

gaearon:
@josepharhar Do you know if there are similar fixes for other browsers?

josepharhar:
> @josepharhar Do you know if there are similar fixes for other browsers?

Firefox was never affected by this issue and I am planning on merging the fix to WebKit soon.

josepharhar:
I *finally* landed the fix in webkit. I don't know anything about safari's release cycle or when it will get to stable.

PR #15425, opened 2019-04-16
Fix: fixed cursor jumping for multiple email input

Nizarius:
Fixed: #15418

I should mention that it is not the perfect solution because the bug itself is created by specific browsers (chrome, safari, opera):

If input's type is "email" and it contains comma and whitespace, the browsers give the wrong input's value: `a, b` =>  `a,b`, although the specification shows that multiple email input can contain whitespace between comma and the next email: https://html.spec.whatwg.org/multipage/input.html#the-multiple-attribute. And even if we try to set the right value programmatically  it won't.

So when React tries to set defaultValue on input while user's typing, it breaks browser's understanding of where caret should be. 

Considering React already contains the similar handling for number input, I suggest to make it for multiple email inputs too. Of course, on input blur input all spaces after commas will be removed, but at least it doesn't interrupt user's typing and provides much better UX experience. 

P.S. Is it such necessary to set defaultValue for input on blur? Maybe we can turn it off for this case until browsers don't fix this bug?

### DIFF
--- a/packages/react-dom/src/client/ReactDOMInput.js
+++ b/packages/react-dom/src/client/ReactDOMInput.js
@@ -412,9 +412,10 @@ export function setDefaultValue(
   type: ?string,
   value: *,
 ) {
+  const isMultipleEmail = type === 'email' && node.hasAttribute('multiple');
   if (
-    // Focused number inputs synchronize on blur. See ChangeEventPlugin.js
-    type !== 'number' ||
+    // Focused number and email inputs synchronize on blur. See ChangeEventPlugin.js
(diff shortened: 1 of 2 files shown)


ohjames:
This is a nice change, but I guess it needs a test?

Nizarius:
Hi! Are there any comments from maintainers?

gaearon:
Could you add a regression fixture to https://github.com/facebook/react/tree/master/fixtures/dom and a detailed test plan so that we know how to verify this change is correct? Thank you.

PR #15428, opened 2019-04-16
React events: fix nested Hover components error

necolas:
Introduces a test that reproduces the error and unexpected output. The error was being caused by the `hoverTarget` being set to `null` even if the module bailed out of dispatching the hover end events. 

How to achieving the desired results for nested Hover will require further consideration.

Ref #15257


necolas:
Updated the patch to remove the addition of support for stopPropagation (no functional difference to the `Hover` module at the moment). `Press` is listening to `pointermove` on the root, and at the moment if `Hover` prevents bubbling `Press` stops working correctly.

PR #15431, opened 2019-04-17
tweak unsafe lifecycle warnings

threepointone:
- explicitly recommend actions as a way to silence the warning
- be clear that the renaming only works for non strict modes 
- include codemod suggestion 

tbh it's still not great, but started a PR to bikeshed. lemme know how to make this better.

This is what it looks like - (updated 18/05)
image


bvaughn:
Possible alternate wording. Don't know if it's great either, just sharing for discussion.

> componentWillMount is deprecated and will be removed in the next major version.
> 
> We suggest doing one of the following to silence this warning:
> 
> 1. Our recommended upgrade path is to replace "componentWillMount" with "componentDidMount". Example code is available at: https://fb.me/react-async-component-lifecycle-hooks
> 
> 2. To silence the warning, rename "componentWillMount" to "UNSAFE_componentWillMount". Note that the warning will still be logged in strict mode. To automatically rename all unsafe lifecycle methods in your codebase, run `npx react-codemod rename-unsafe-lifecycles <path/to/code>` in your command line.
> 
> Please update the following components: SomeComponent, AnotherComponent
> 
> Learn more about this warning here:
> https://fb.me/react-async-component-lifecycle-hooks

threepointone:
Updated with suggestions, did the same for the server rendered warning which I'd missed, updated tests. I also updated the screenshot in the description. ^

threepointone:
a followup 'todo' - make a warnings page we can point folks to, that's basically the blogpost, but reordered to emphasize solutions/suggestions.

threepointone:
I forgot to add `Note that the warning will still be logged in strict mode.` which I actually like. Will update.

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -573,13 +573,17 @@ function resolve(
...
-                  'removed in the next major version. Read about the motivations ' +
-                  'behind this change: ' +
-                  'https://fb.me/react-async-component-lifecycle-hooks' +
-                  '\n\n' +
-                  'As a temporary workaround, you can rename to ' +
-                  'UNSAFE_componentWillMount instead.',
+                'componentWillMount has been renamed to UNSAFE_componentWillMount, ' +
+                  "and the old name won't work in the next major version of React.\n" +
+                  'We suggest doing one of the following:\n' +
+                  '- To rename all deprecated lifecycles to their new names, you can run ' +
```
bvaughn: I'm concerned about this being our first "suggestion" in the list since it's _not actually_ our primary recommendation.

bvaughn: I think people might read this list as if it's prioritized. Could we move the rename bit to the _end_?

threepointone: Can indeed. Will do.

threepointone:
updated PR and screenshot

packages/react-dom/src/server/ReactPartialRenderer.js
threepointone: done

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -141,12 +141,18 @@ if (__DEV__) {
...
       lowPriorityWarning(
         false,
-        'componentWillMount is deprecated and will be removed in the next major version. ' +
-          'Use componentDidMount instead. As a temporary workaround, ' +
-          'you can rename to UNSAFE_componentWillMount.' +
-          '\n\nPlease update the following components: %s' +
-          '\n\nLearn more about this warning here:' +
-          '\nhttps://fb.me/react-async-component-lifecycle-hooks',
+        'componentWillMount has been renamed to UNSAFE_componentWillMount, ' +
+          "and the old name won't work in the next major version of React.\n" +
```
bvaughn: nit: two "\n\n" (here and below)

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -141,12 +141,18 @@ if (__DEV__) {
...
+          'We suggest doing one of the following:\n' +
+          '- If you initialize state in componentWillMount, move this logic into the constructor.\n' +
+          '- If you fetch data or perform other side effects in componentWillMount, ' +
+          'move this logic into componentDidMount.\n' +
+          '- To rename all deprecated lifecycles to their new names, you can run ' +
```
bvaughn: Maybe we should give a hint as to why folks might not want to just rename and ignore? Maybe we could replace

"_To rename all deprecated lifecycles to their new names, ..._"

...with

"_To prevent this warning from being logged to the console, ..._"

...so it sounds a little more like you're just choosing to ignore the warning without fixing the problem?

(here and below)

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -141,12 +141,18 @@ if (__DEV__) {
...
+          '`npx react-codemod rename-unsafe-lifecycles <path/to/code>` in your project folder. ' +
+          '(Note that the warning will still be logged in strict mode.)\n' +
+          '\nPlease update the following components: %s\n' +
+          '\nLearn about this warning, with more examples and suggestions here:\n' +
```
bvaughn: nit: Maybe just "_Learn more about this warning here:_" ? (here and below)

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -164,11 +170,22 @@ if (__DEV__) {
...
-          '\nhttps://fb.me/react-async-component-lifecycle-hooks',
+        'componentWillReceiveProps has been renamed to UNSAFE_componentWillReceiveProps, ' +
+          "and the old name won't work in the next major version of React.\n" +
+          'We suggest doing one of the following:\n' +
+          "- If you're updating state whenever props change, " +
+          'move this logic into static getDerivedStateFromProps.\n' +
+          '- If you fetch data or perform other side effects in componentWillReceiveProps, ' +
+          'move this logic into componentDidUpdate.\n' +
+          '- Refactor your code to not use derived state at all, as described at ' +
+          // todo - this should be an fb.me link
```
bvaughn: Just do this? Only takes 10 seconds? 😄

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -164,11 +170,22 @@ if (__DEV__) {
...
-          '\n\nLearn more about this warning here:' +
-          '\nhttps://fb.me/react-async-component-lifecycle-hooks',
+        'componentWillReceiveProps has been renamed to UNSAFE_componentWillReceiveProps, ' +
+          "and the old name won't work in the next major version of React.\n" +
+          'We suggest doing one of the following:\n' +
+          "- If you're updating state whenever props change, " +
+          'move this logic into static getDerivedStateFromProps.\n' +
+          '- If you fetch data or perform other side effects in componentWillReceiveProps, ' +
+          'move this logic into componentDidUpdate.\n' +
+          '- Refactor your code to not use derived state at all, as described at ' +
```
bvaughn: Maybe something like?

> Refactor your code to use memoization techniques instead of derived state.
> Learn more at: https://fb.me/react-derived-state

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -186,12 +203,19 @@ if (__DEV__) {
...
-          'you can rename to UNSAFE_componentWillUpdate.' +
-          '\n\nPlease update the following components: %s' +
-          '\n\nLearn more about this warning here:' +
-          '\nhttps://fb.me/react-async-component-lifecycle-hooks',
+        'componentWillUpdate has been renamed to UNSAFE_componentWillUpdate, ' +
+          "and the old name won't work in the next major version of React.\n" +
+          'We suggest doing one of the following:\n' +
+          '- If you fetch data or perform other side effects in componentWillUpdate, ' +
+          'move this logic into componentDidUpdate.\n' +
+          "- If you're reading DOM properties before an update, " +
```
bvaughn: Legit use cases for `getSnapshotBeforeUpdate` seem so uncommon, I wonder if it's worth mentioning here at all? I don't have a strong preference one way or another, but a slight preference to maybe not mention it here for fear of encouraging bad patterns for those maybe lacking context of when the method should be used.

threepointone: I think I'll leave it here. We seem fairly comprehensive for the others, it would be frustrating if the dev had to go to a doc to discover a new api for just this one case.

bvaughn: I guess I'm concerned that if people just move stuff to that method, without the added context of _why_ it exists, they might be tempted to do mutations/side effects in it.

packages/react-reconciler/src/ReactStrictModeWarnings.js
threepointone: How about "To ignore this warning from being logged to the console..."

packages/react-reconciler/src/ReactStrictModeWarnings.js
threepointone: They would probably try componentDidUpdate before that though? I dunno tbh. I could remove this, but seems like we should use same reasoning to remove getDerivedStateFromProps if so.

packages/react-reconciler/src/ReactStrictModeWarnings.js
threepointone: rather, "to disable this warning..."

threepointone: It's so wordy :(

threepointone: I think that we're saying the warning will still log in strict mode, and that we've moved this to the bottom of the list, is sufficient. I'll fight you on this briannson.

threepointone:
updated PR and screenshot

sophiebits:
I would not try to explain when to use which lifecycle method in the warning text. I think it is just too small a space and we should encourage people to read the blog posts instead. My suggestion would be to write something along the following lines:

In non-strict mode:

    Use of componentWillMount is not recommended and may indicate bugs in your code.

    * Rename componentWillMount in your code to UNSAFE_componentWillMount to suppress this warning. In React 17.x, only the UNSAFE_ name will work.
    * To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles /path/to/code`.
    * Move component logic from componentWillMount into componentDidMount (preferred in most cases) or the constructor. See https://fb.me/react-async-component-lifecycle-hooks for details.
    
    Please update the following components: %s

In strict mode if componentWillMount is used:

    Use of componentWillMount and UNSAFE_componentWillMount is not recommended and may indicate bugs in your code.

    * Rename componentWillMount in your code to UNSAFE_componentWillMount. In React 17.x, only the UNSAFE_ name will work.
    * To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles /path/to/code`.
    * Move component logic from componentWillMount into componentDidMount (preferred in most cases) or the constructor. See https://fb.me/react-async-component-lifecycle-hooks for details.
    
    Please update the following components: %s

In strict mode if componentWillMount is not used:

    Use of UNSAFE_componentWillMount is not recommended and may indicate bugs in your code.

    Move component logic from UNSAFE_componentWillMount into componentDidMount (preferred in most cases) or the constructor. See https://fb.me/react-async-component-lifecycle-hooks for details.
    
    Please update the following components: %s

I think it's also important there is a place to learn more about the auto-rename script. People are gonna be a little hesitant unless they know what it'll do. (Which files will it run on? etc)

threepointone:
Sorry I got a while to get back to this. Some thoughts - 

> I would not try to explain when to use which lifecycle method in the warning text. 

Trying to reconcile this with the suggestion, it's only one sentence shorter than what the PR currently has. I think a quick reference of alternatives is indeed valuable, and don't want folks to have to read a long post for every warning (and the link is pretty obvious, so they *will* go there if they don't comprehend immediately what to do.

I agree we should have different warning with strict/non strict mode, hacking on it asap.

RyanLamansky:
This messaging doesn't harmonize well with the current state of the server renderer, which works with `componentWillMount` but _not_ `componentDidMount`.

threepointone:
abandoning this for https://github.com/facebook/react/pull/16103, which uses all the feedback from here, and then some. see y'all there!

PR #15449, opened 2019-04-19
React Events: extract shared functions

necolas:
Moving functions that are used across the event modules to a shared module. This is currently being inlined in each event module. We can determine which, if any, of these functions could be moved into the core and exposed via `context`. And we could avoid inlining any remaining shared code that doesn't belong in core.


NE-SmallTown:
I think rename the file name to a 'shared' directory maybe will be more common

PR #15450, opened 2019-04-19
React events: consolidate logic of Hover event component

necolas:
Stacks on #15449. Demo integrating this patch: https://fburl.com/nfye4390

Minor refactor of `Hover` and additional regression coverage.

Ref #15257


necolas:
It should start passing once I merge the commit it depends on

PR #15451, opened 2019-04-19
React Events: consolidate logic for Press event component

necolas:
Stacks on #15449. Demo integrating this patch: https://fburl.com/nfye4390

Refactor of `Press` and additional regression coverage.

1. The logic for "start", "move", "end", and "cancel" events is consolidated into a single block to reduce duplication and improve consistency of the UX across input-types. Also reduces code size.

2. The bailout logic for anchor tags is removed since we `preventDefault` for `click` by default. We can discuss scenarios where it makes sense to limit functionality around interactions on anchor tags.

3. The logic for ignoring emulated events is simplified and improved. Pointer events can produce emulated touch (immediately after pointer) and mouse events (delayed) which is now accounted for and tested.

Ref #15257

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -13,6 +13,11 @@ import type {
   ReactResponderDispatchEventOptions,
 } from 'shared/ReactTypes';
 import {REACT_EVENT_COMPONENT_TYPE} from 'shared/ReactSymbols';
+import {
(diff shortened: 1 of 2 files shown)


necolas:
Added a link to an interactive demo running this patch. Will start doing this for most PRs that can benefit from reviewers checking interactions for themselves.

PR #15462, opened 2019-04-20
Experimental event API: rework the propagation system for event components

trueadm:
After much discussion with @necolas and quite a bit of internal testing, we came up with an idea to improve the experimental event API and its propagation system. Rather than return a `boolean` from the event responder `onHandle` to indicate if an event should be propagated, all events are now propagated depending on the responder and the event component.

Responders now have the `stopLocalPropagation` option (required). This tells the event system what it should do we encountering this particular event responder. If this flag is set to `true`, then when we encounter the same event responder during propagation, we skip calling the event listeners on that responder for that event component.

Furthermore, there was some confusion over the phases on the event object, and so now `onHandle` has been split into three methods: `onEvent`, `onEventCapture` and `onRootEvent`.

Ref #15257

### DIFF
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -18,6 +18,7 @@ import {
   HostComponent,
 } from 'shared/ReactWorkTags';
 import type {
+  ReactEventResponder,
   ReactEventResponderEventType,
   ReactEventComponentInstance,
   ReactResponderContext,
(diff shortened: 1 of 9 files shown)


trueadm:
@necolas it would be good if you could validate this PR locally when you get a chance against your codesandbox. Thanks!

trueadm:
@necolas I've fixed the conflicts with master and applied the name change that you suggested. I'll let you test it first – if it's all good, feel free to merge. :)

PR #15464, opened 2019-04-21
React Native: Allow Views to be nested inside of Text

rigdern:
This feature is now supported on both iOS and Android. The Android feature was merged a couple of weeks ago: https://github.com/facebook/react-native/pull/23195.

Here's some sample React Native code that utilizes the inline view feature:

```js
export default class App extends Component<Props> {
  render() {
    return (
      <View style={styles.background}>
        <Text>
          This is an{' '}
          <View style={{backgroundColor: 'steelblue', width: 10, height: 10}} />
          {' '}inline view.
        </Text>
      </View>
    );
  }
}
```

### Output Before PR (Android)

image

### Output After PR (Android)

image

Adam Comella
Microsoft Corp.

### DIFF
--- a/packages/react-native-renderer/src/ReactFabricHostConfig.js
+++ b/packages/react-native-renderer/src/ReactFabricHostConfig.js
@@ -191,11 +191,6 @@ export function createInstance(
     }
   }
 
-  invariant(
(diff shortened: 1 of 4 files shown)


Dimpletest:
Test comment

PR #15466, opened 2019-04-21
Add function to find all host instances for fiber, not just the first

sompylasar:
Modified package: `react-reconciler`
- `ReactFiberReconciler`'s `injectIntoDevTools` now injects one more function: `findHostInstancesByFiber`
- `ReactFiberTreeReflection` now exports one more function `findAllCurrentHostFibers`

This is required to support highlighting all corresponding DOM elements of Fragments in `react-devtools-experimental`:
https://github.com/bvaughn/react-devtools-experimental/issues/131

CC @bvaughn @sophiebits

### DIFF
--- a/packages/react-debug-tools/src/__tests__/ReactDevToolsHooksIntegration-test.js
+++ b/packages/react-debug-tools/src/__tests__/ReactDevToolsHooksIntegration-test.js
@@ -19,6 +19,8 @@ describe('React hooks DevTools integration', () => {
   let overrideHookState;
   let scheduleUpdate;
   let setSuspenseHandler;
+  let findHostInstanceByFiber;
+  let findHostInstancesByFiber;
 
   beforeEach(() => {
     global.__REACT_DEVTOOLS_GLOBAL_HOOK__ = {
@@ -26,6 +28,8 @@ describe('React hooks DevTools integration', () => {
         overrideHookState = injected.overrideHookState;
         scheduleUpdate = injected.scheduleUpdate;
         setSuspenseHandler = injected.setSuspenseHandler;
+        findHostInstanceByFiber = injected.findHostInstanceByFiber;
+        findHostInstancesByFiber = injected.findHostInstancesByFiber;
       },
       supportsFiber: true,
       onCommitFiberRoot: () => {},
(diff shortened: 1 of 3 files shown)


sompylasar:
Makes sense, I'll move the new one to DevTools. The original will likely be hard to delete with respect to backwards compatibility with all the previous versions of DevTools. If that's okay I can do that, too.

gaearon:
Should we just vendor this code into DevTools? We already have getCurrentFiber copied there because otherwise we don’t know the current props for inspection. If we have it there anyway I don’t see what we’re gaining by passing this through. It’s not like we depend on internals any less.

sompylasar:
> Should we just vendor this code into DevTools? We already have getCurrentFiber copied there because otherwise we don’t know the current props for inspection. If we have it there anyway I don’t see what we’re gaining by passing this through. It’s not like we depend on internals any less.
> 
> If we vendor it then it’ll work in production too. Which I think is valuable for DevTools.

Yes, what @sebmarkbage said. My only concern is the older versions of DevTools. If this function (the singular) disappears from React, these old versions will break. If we're okay to assume that developers only use the latest DevTools, it's fine to remove the functions from React and add to DevTools (to the new "experimental" one, not sure if the old one will be supported independently).

bvaughn:
> Should we just vendor this code into DevTools? 

Yes.

> If this function (the singular) disappears from React, these old versions will break.

Older DevTools will break with  newer React releases regardless. React maintaining backwards compat with legacy DevTools isn't a use case we intend to support.

sophiebits:
We should make sure there is always a stable/released devtools that works with the latest released React, but other than that we don't guarantee that new React will be compatible with old devtools. It does mean that we either need to release the new devtools or update the current ones before we can delete findHostInstanceByFiber from React. (Which means not committing that change to master so as to not block any more releases from happening.)

bvaughn:
>  We should make sure there is always a stable/released devtools that works with the latest released React

Absolutely.

But we regularly make backwards  breaking  changes to DevTools with  minor React  releases- e.g.  any time we add a new component  type or change the  way suspense works, e tc.

> It does mean that we either need to release the new devtools or update the current ones before we can delete findHostInstanceByFiber from React.

This is one of the  reasons I want to move (this  version of)  DevTools  into the React repo eventually, so we'll have automated test  coverage to prevent  accidental breaks.

sophiebits:
I think we’re saying the same thing?

bvaughn:
Yes. I was adding context.

sompylasar:
OK so the plan is:
- [x] 1. Close this PR.
- [ ] 2. Move the new (plural) function to the new DevTools PR that needs it https://github.com/bvaughn/react-devtools-experimental/issues/131
- [ ] 3. Open a new PR to https://github.com/facebook/react-devtools that adds the old (singular) function there.
- [ ] 4. Open a new PR to React that removes the old (singular) function.

sompylasar:
### 2.

Moved `findHostInstancesByFiber` (plural) to `react-devtools-experimental`. https://github.com/bvaughn/react-devtools-experimental/issues/131

### 3.

Copied `findHostInstanceByFiber` (singular) to `react-devtools`. https://github.com/facebook/react-devtools/pull/1339

`react-devtools-experimental` already had a copy of `findCurrentFiberUsingSlowPath` that the `findAllCurrentHostFibers` is based on: https://github.com/bvaughn/react-devtools-experimental/blob/942f67c9369cc587708bd0cde1e61099ab9ed8f3/src/backend/renderer.js#L1298-L1479

So I copied it to `react-devtools` likewise.

### 4.

`findHostInstanceByFiber` is delegating to `findCurrentHostFiber` which is used in React not only for DevTools, but in many other places.

So we can remove `findHostInstanceByFiber` from `injectIntoDevTools` -> `injectInternals` but cannot remove `findCurrentHostFiber`.

Issue #13097, opened 2018-06-22
Hydrating a Portal errors and duplicates contents


marcusdarmstrong:
**Do you want to request a *feature* or report a *bug*?**

Probably bug, but arguably a feature request, I suppose.

**What is the current behavior?**

I've attempted my best effort at a fiddle that shows off the particular issue. Obviously server side rendering is impossible via JSFiddle, but the markup should be equivalent to having rendered `Test` into a div with id `test-1` during server side render.

https://jsfiddle.net/y8o5n2zg/

As seen in the fiddle, an attempt to ReactDOM.hydrate() a portal results in:
> `Warning: Expected server HTML to contain a matching text node for "Hello World" in <div>.`

Additionally, after failing to hydrate, React renders the component and appends it resulting in a duplicated section of DOM:
> `<div id="test-1">Hello WorldHello World</div>`

**What is the expected behavior?**

In an ideal world, calling hydrate on a component that has portals would allow those DOM containers to hydrate into the components they were rendered with.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

I've only tested this in 16.4.1, but I've confirmed the behavior in Chrome and Firefox. Given that I'm really looking at an edge case here I doubt it worked previously.

#### *Why* I'm doing this edge-case-y nonsense:

We're currently using multiple React roots on our pages (as some portions of the pages are not rendered by React yet), most of which are server-side rendered. We'd like to be able to hydrate them into a single React root on page, so that we can share contexts between them without difficulty and without repeating those context components in memory (in some cases we can have a good number of roots on the page—20-30, perhaps?).

In searching, I found a few potentially related bugs (#12615, #10713, #11169), but it seemed like these really didn't line up with my (hopefully valid?) use case.

Thanks!

MaxMotovilov:
Nearly identical use case here. For now, resolving by a kludge in the component rendered into a portal (render self with `display:none`, detect duplicate, delete, show self).

gaearon:
When you hydrate, the initial render should match your server render. But portals are currently not supported on the server. Therefore, hydrating a portal doesn't make sense with current limitations.

I think you want to do something like:

```js
state = {mounted: false};

componentDidMount() {
  this.setState({ mounted: true });
}

render() {
  return <div>{this.state.mounted && ReactDOM.createPortal(...)}</div>
}
```

Does that make sense? Same workaround as you need to use when your client render doesn't match.

marcusdarmstrong:
[...]

PR #15473, opened 2019-04-23
Issues/12615

VariableVasasMT:

For Issue: https://github.com/facebook/react/issues/12615

### DIFF
--- a/fixtures/ssr/src/components/App.js
+++ b/fixtures/ssr/src/components/App.js
@@ -1,14 +1,9 @@
-import React, {useContext, useState, Suspense} from 'react';
+import React, {useState} from 'react';
 
 import Chrome from './Chrome';
 import Page from './Page';
 import Page2 from './Page2';
-import Theme from './Theme';
-
-function LoadingIndicator() {
-  let theme = useContext(Theme);
-  return <div className={theme + '-loading'}>Loading...</div>;
-}
+import CreatePortal from './CreatePortal';
 
 export default function App({assets}) {
   let [CurrentPage, switchPage] = useState(() => Page);
@@ -23,9 +18,8 @@ export default function App({assets}) {
         <a className="link" onClick={() => switchPage(() => Page2)}>
           Page 2
         </a>
-        <Suspense fallback={<LoadingIndicator />}>
-          <CurrentPage />
-        </Suspense>
+        <CurrentPage />
+        <CreatePortal />
       </div>
     </Chrome>
   );
--- a/fixtures/ssr/src/components/CreatePortal.js
+++ b/fixtures/ssr/src/components/CreatePortal.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import ReactDom from 'react-dom';
+
+export default class CreatePortal extends React.Component {
+  render() {
+    if (typeof document === 'undefined') {
+      return null;
+    }
+    const root = document.body;
+    return ReactDom.createPortal(<div>Hello Portal</div>, root);
+  }
+}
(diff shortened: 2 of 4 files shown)


VariableVasasMT:
I havent added test as of now, because I dont know how to.
partial hydration throws direct error, its hydration through loadable component using ssr, causes this issue, I made changed to ssr fixtures to show that it works, but cant find a way to modify following to make test suite do the same thing

```
it('should be able to use lazy components after hydrating with portal', async () => {
    class Child extends React.Component {
      render() {
        if (!document.getElementById('root')) {
          return null;
        }
        const root = document.getElementById('root');
        return ReactDOM.createPortal('World', root);
      }
    }

    const Lazy = React.lazy(
      () =>
        new Promise(resolve => {
          setTimeout(
            () =>
              resolve({
                default: () => <Child />,
              }),
            1000,
          );
        }),
    );
    class HelloWorld extends React.Component {
      state = {isClient: false};
      componentDidMount() {
        this.setState({
          isClient: true,
        });
      }
      render() {
        return (
          <div id="root">
            Hello{' '}
            {this.state.isClient && (
              <React.Suspense fallback="loading">
                <Lazy />
              </React.Suspense>
            )}
          </div>
        );
      }
    }

    element.innerHTML = ReactDOMServer.renderToString(<HelloWorld />);
    expect(element.textContent).toBe('Hello ');

    ReactDOM.hydrate(<HelloWorld />, element);
    expect(element.textContent).toBe('Hello loading');

    jest.runAllTimers();
    await Promise.resolve();
    Scheduler.flushAll();
    console.log();
    expect(element.textContent).toBe('Hello world');
  });
```

VariableVasasMT:
more so my issue with creating test is that a document has been declared globaly

cpojer:
Does it make sense to throw when trying to render a portal on the server or should they just be ignored instead?

ljharb:
If they throw, then a component using portal can’t be server rendered - but if they’re ignored, then there’ll be a client-server dom mismatch on initial render. What would be ideal for me is if they’re ignored on the server for rendering, and rendered on the client but ignored for the purposes of the mismatch algorithm, if I’m understanding things correctly.

VariableVasasMT:
@cpojer @ljharb  on the fair note, this specific issue is not server related, the hydration occurs on client, I dont think if there would be dom mismatch because of it, or may be I dont understand enough.

ljharb:
My understanding is that hydration is the initial render on the client on top of the initial server render, and effectively that must not produce a dom diff (altho i think there’s some wiggle room in there to account for subtle html differences). Which means, that if something is rendered in hydration, that i suspect it needs to have also been rendered on the server - or, the diffing algorithm needs to special-case allowing it.

VariableVasasMT:
>diffing algorithm needs to special-case allowing it.
@ljharb I agree with the latter, but is it worth it considering https://github.com/facebook/react/issues/13097 is open?
As much I understood, HostPortal is not passed to fibre or render tree
```
case HostPortal:
          pushHostContainer(
            workInProgress,
            workInProgress.stateNode.containerInfo,
          );
          break;
```
after this we loose the context of Portal, I think but I am not sure,
it should be handled like 
```
case SuspenseComponent: 
```
which does this
```
updateDehydratedSuspenseComponent
```

I am not sure though.

ljharb:
I'm not really sure ¯\\\_(ツ)_/¯ but presumably the first one to land shouldn't presuppose the existence of the other :-)

VariableVasasMT:
closing this, as I think this pull request is not important

PR #15474, opened 2019-04-23
Use stricter equality check

gaearon:
Found a few leftovers. Seems like it's nice to fix if it's in the tight paths.

### DIFF
--- a/packages/react-reconciler/src/ReactChildFiber.js
+++ b/packages/react-reconciler/src/ReactChildFiber.js
@@ -831,7 +831,7 @@ function ChildReconciler(shouldTrackSideEffects) {
           newChildren[newIdx],
           expirationTime,
         );
-        if (!newFiber) {
+        if (newFiber === null) {
           continue;
         }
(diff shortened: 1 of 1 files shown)


pbondoer:
Just a curious passerby (so if you don't have time to reply I completely understand), but what's the benefit of using triple-equality here? Does the implicit boolean cast impact performance (though I would find that strange...)? Or does it not matter much / at all, and is simply to not rely on truthy / falsy and be declarative and explicit? Perhaps for code readability and/or consistency in the codebase?

gaearon:
There is very minor performance impact because double equals is like comparing to null, undefined, and document.all (yeah for realz). Maybe it doesn’t matter but this is also a very tight code path. The more explicit we are, the easier it is for VM to run code fast.

pbondoer:
@gaeron Double equality? It seems to me that you used triple equality here :o?

gaearon:
Sorry I mean the implicit boolean coercion in the condition. That thing will check for “falsy” which is too broad and includes weird cases.

PR #15479, opened 2019-04-23
Measure callback timeout relative to current time

acdlite:
Fixes a bug where the timeout passed to `scheduleCallback` represented an absolute timestamp, instead of the amount of time until that timestamp is reached. The solution is to subtract the current time from the expiration.

The bug wasn't caught by other tests because we use virtual times that default to 0, and most tests don't advance time.

I also moved the `initialTimeMs` offset to the `SchedulerWithReactIntegration` module so that we don't have to remember to subtract the offset every time. (We should consider upstreaming this to the Scheduler package.)

### DIFF
--- a/packages/react-reconciler/src/SchedulerWithReactIntegration.js
+++ b/packages/react-reconciler/src/SchedulerWithReactIntegration.js
@@ -65,7 +65,6 @@ export const IdlePriority: ReactPriorityLevel = 95;
 // NoPriority is the absence of priority. Also React-only.
 export const NoPriority: ReactPriorityLevel = 90;
 
-export const now = Scheduler_now;
 export const shouldYield = disableYielding
   ? () => false // Never yield when `disableYielding` is on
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/SchedulerWithReactIntegration.js
```diff
@@ -65,14 +65,24 @@ export const IdlePriority: ReactPriorityLevel = 95;
...
 
-export const now = Scheduler_now;
 export const shouldYield = disableYielding
   ? () => false // Never yield when `disableYielding` is on
   : Scheduler_shouldYield;
 
 let immediateQueue: Array<SchedulerCallback> | null = null;
 let immediateQueueCallbackNode: mixed | null = null;
 let isFlushingImmediate: boolean = false;
+let initialTimeMs: number = Scheduler_now();
```
NE-SmallTown: Should this be `const`?

PR #15487, opened 2019-04-24
Event API: Add `FocusScope` surface

trueadm:
This adds a new `FocusScope` event API surface/responder. For now, `FocusScope` is a very basic implementation that only supports `autoFocus`, `restoreFocus` and `trap`. When you tab in a trapped `FocusScope`, you're limited to the scope of the event component. Pressing tab on the last focusable element of a `FocusScope` that is trapped will result in the focus moving to the first focusable element of the `FocusScope`. Furthermore, elements that aren't focusable (i.e. with `tabIndex` set to `-1`) will be skipped in all cases and elements nested in portals will traverse as expected via the React fiber tree rather than the DOM tree.

Example of `FocusScope` with focus trapping and a few other props:

```jsx
<div className="Dialog">
  <div className="Dialog-inner">
    <FocusScope trap={true} autoFocus={true} restoreFocus={true}>
      <h2>Focus Trapped Modal</h2>
      <input placeholder="You can focus me" />
      <div className="FakeButton" tabIndex={0}>
        I am a div but can also be focused
      </div>
      <input placeholder="Focus is blocked" tabIndex={-1} />
      <Press onPress={close} tabIndex={0}>
        <div className="FakeButton">Close Modal</div>
      </Press>
    </FocusScope>
  </div>
</div>;

```


packages/react-events/src/FocusScope.js
```diff
@@ -0,0 +1,146 @@
...
+        if (context.isTargetWithinEventComponent(target)) {
+          state.currentFocusedNode = ((target: any): HTMLElement);
+        } else {
+          if (state.currentFocusedNode !== null) {
+            state.currentFocusedNode.focus();
+          } else {
+            focusFirstChildEventTarget(context, state);
+          }
+        }
+      } else if (type === 'keydown' && nativeEvent.key === 'Tab') {
```
necolas: might need to check that ctrl, alt, or meta aren't also being held when key is `Tab`

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -322,61 +324,65 @@ const eventResponderContext: ReactResponderContext = {
...
+  }
+  if (type === 'a' || type === 'area') {
+    return !!memoizedProps.href;
+  }
+  return (
+    type === 'button' ||
+    type === 'textarea' ||
+    type === 'input' ||
+    type === 'object' ||
+    type === 'select'
```
giuseppeg: The list is missing a few elements: iframe, embed and `contenteditable`. 
Should this component skip subtrees that have the `inert` attribute set on the root?

packages/react-events/src/FocusScope.js
```diff
@@ -0,0 +1,157 @@
...
+      }
+    }
+  },
+  onMount(
+    context: ReactResponderContext,
+    props: FocusScopeProps,
+    state: FocusScopeState,
+  ) {
+    if (props.restoreFocus) {
+      state.nodeToRestore = document.activeElement;
```
giuseppeg: For this component to work inside of iframes this should be the ownerDocument of the current node.

packages/react-dom/src/events/DOMEventResponderSystem.js
trueadm: We can deal with `inert` in a future PR. It's a bit more involved as what if a parent tree has `inert` and its sub-trees have the `FocusScope`?

packages/react-events/src/FocusScope.js
trueadm: Thanks for the feedback. I addressed these points in https://github.com/facebook/react/pull/15496 :)

packages/react-dom/src/events/DOMEventResponderSystem.js
giuseppeg: Agreed that is not straightforward, I guess that `FocusScope` should noop when an ancestor has the `inert` attribute i.e. (force) disable trap and restore.

I am afraid that because of Portals React has to polyfill `inert` and can't rely on the native implementation.

Andarist:
Work around the Events API seems to be even more exciting for me as a developer than Concurrent Mode etc, and I'm stoked about the latter 🔥 . 

Great work!

packages/react-dom/src/events/DOMEventResponderSystem.js
marcysutton: It would be pretty great if React polyfilled `inert` anyway, as it is only supported in Chrome behind a flag.

necolas: WICG have a polyfill https://github.com/WICG/inert

mattiamanzati:
@trueadm Will the focusscope remain only a react-dom thing or will it be expanded also over react native eventually? I would love that! Thanks for making this happening to everyone contributing!

trueadm:
@mattiamanzati it’s very early to say but that’s definitely our long term ambition with this new event API.

giuseppeg:
@trueadm FWIW back then I forgot to mention that that in order for elements to be focusable they should be visible. At work I do this:

```js
export function getFocusableElements(node: HTMLElement): Array<HTMLElement> {
  return Array.prototype.filter.call(
    node.querySelectorAll(FOCUSABLE_ELEMENTS_SELECTOR),
    child =>
      !(child.inert || child.hasAttribute('inert') || child.getAttribute('tabindex') === '-1') &&
      !!(child.offsetWidth || child.offsetHeight || child.getClientRects().length),
  )
}
```

trueadm:
We don’t want to use query selectors. They don’t work with suspense and portals correctly.

giuseppeg:
@trueadm oh yea definitely, that part is not relevant. The filter function filters out non visible elements this way `!!(child.offsetWidth || child.offsetHeight || child.getClientRects().length)`. Obviously you'd need the actual DOM node and this thing can trigger layout

ryankshaw:
sorry, noob question, I see this is merged to master, does that mean it is this something we can expect to be in react 17 but won't be in any 16.x version?

trueadm:
@ryankshaw No, this won't be in React 17. It's not even in React's master branch now and has since been removed as it was only a short experiment.

PR #15490, opened 2019-04-24
[React Native] Inline calls to FabricUIManager in shared code

TheSavior:
My previous commit (https://github.com/facebook/react/commit/1b2159acc34d9ca2c950e53bfc46db385a75dbad) caused crashes when running in Fabric when navigating from a paper screen to a fabric screen. 

The NativeMethodsMixin and ReactNativeComponent code is shared in both the Paper renderer and Fabric renderer. The previous commit caused the top of the paper renderer to include `require('FabricUIManager')`. 

`FabricUIManager.js` is essentially just:
```
module.exports = global.nativeFabricUIManager;
```

Because this code was getting called at the top of the paper renderer, `FabricUIManager` was exporting `undefined`. 

When transitioning to a Fabric screen, c++ installs that global, and then the FabricRenderer has this check:

```
if (FabricUIManager.registerEventHandler) {
  ...
}
```

This manifested as a crash `cannot access registerEventHandle of undefined`. This is because FabricUIManager was cached from being accessed too early. 

## The "right" fix
We will fix this correctly by having native always, on startup, define and set the object:
```
global.nativeFabricUIManager = {
  createNode,
  cloneNode,
  measure,
  measureInWindow,
  // ...
} 
```

When starting with the paper renderer, these functions will just throw.

Then, when Fabric is loaded, it will replace the implementation of those functions with the actual correct implementation.

This will ensure there is no timing issues with when JavaScript reads and caches the value of the global or any of the functions.

We will need to ensure we replace the implementations of the functions and not redefine them because we will need to support JS reading the functions early like this:

```
const {createNode, measure} = global.nativeFabricUIManager;
```
Reading them early, and then calling them after Fabric has loaded will need to work. 

## The "short term" fix

This PR includes the short term fix which is to inline these requires so that it isn't called early. Normally this would happen by default as we run everything inline requires however this file is blacklisted from inline requires in React Native as it is faster this way. 

## Test Plan:
Opened Marketplace (this worked before this change)
Navigating from Catalyst (paper) to Fabric RNTester (fabric). This crashed before with the error above and no longer crashes.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -125,6 +125,12 @@ module.exports = {
         // https://github.com/jest-community/eslint-plugin-jest
         'jest/no-focused-tests': ERROR,
       }
+    },
+    {
+      files: ['packages/react-native-renderer/**/*.js'],
+      globals: {
+        nativeFabricUIManager: true,
+      }
     }
   ],
 
--- a/packages/react-native-renderer/src/NativeMethodsMixin.js
+++ b/packages/react-native-renderer/src/NativeMethodsMixin.js
@@ -88,7 +88,7 @@ export default function(
         // We can't call FabricUIManager here because it won't be loaded in paper
         // at initialization time. See https://github.com/facebook/react/pull/15490
         // for more info.
-        global.nativeFabricUIManager.measure(
+        nativeFabricUIManager.measure(
           maybeInstance.node,
           mountSafeCallback_NOT_REALLY_SAFE(this, callback),
         );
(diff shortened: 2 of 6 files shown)


bvaughn:
We already have a few inline requires for our www/fbsource forks. I didn't think there was harm  adding another one  temporarily.

sebmarkbage:
> We already have a few inline requires for our www/fbsource forks. I didn't think there was harm adding another one temporarily.

Previous bad ideas are never good precedence for future bad ideas. That's how small bad ideas leak and become large bad ideas.

bvaughn:
Uh, ok. I was not aware that we had established this particular thing as a "bad idea" 😄

TheSavior:
> I'm surprised lint and Flow lets you do this.
> 
> Would be nice to declare it as a top level global property here:
> 
> https://github.com/facebook/react/blob/master/scripts/flow/react-native-host-hooks.js#L100

Apparently accessing something with `global.foo` doesn't complain, but just accessing `foo` does. 

Also, apparently `maybeInstance.node` as well as the `mountSafeCallback_NOT_REALLY_SAFE` return are both typed as any, so defining the flow types don't help catch anything. We should fix that at some point.

PR #15492, opened 2019-04-25
Warn when suspending at wrong priority

acdlite:
Adds a warning when a user-blocking update is suspended.

Ideally, all we would need to do is check the current priority level. But we currently have no rigorous way to distinguish work that was scheduled at user-blocking priority from work that expired a bit and was "upgraded" to a higher priority. That's because we don't schedule separate callbacks for every level, only the highest priority level per root. The priority of subsequent levels is inferred from the expiration time, but this is an imprecise heuristic.

However, we do store the last discrete pending update per root. So we can reliably compare to that one. (If we broaden this warning to include high pri updates that aren't discrete, then this won't be sufficient.)

My rationale is that it's better for this warning to have false negatives than false positives.

Potential follow-ups:
- Bikeshed more on the message. I don't like what I landed on that much but I think it's good enough to start.
- Include the names of the components that updated. (The ideal place to fire the warning is during the setState call but we don't know if something will suspend until the render phase. Maybe we could be clever and warn during a subsequent update to the same component?)

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -698,6 +698,7 @@ function prepareFreshStack(root, expirationTime) {
 
   if (__DEV__) {
     ReactStrictModeWarnings.discardPendingWarnings();
+    componentsWithSuspendedDiscreteUpdates = null;
   }
 }
 
@@ -817,6 +818,14 @@ function renderRoot(
         }
 
         const returnFiber = sourceFiber.return;
+
+        checkForWrongSuspensePriorityInDEV(
+          thrownValue,
+          root,
+          sourceFiber,
+          expirationTime,
+        );
+
         throwException(
           root,
           returnFiber,
@@ -1225,6 +1234,7 @@ function commitRoot(root, expirationTime) {
 function commitRootImpl(root, expirationTime) {
   flushPassiveEffects();
   flushRenderPhaseStrictModeWarningsInDEV();
+  flushSuspensePriorityWarningInDEV();
 
   invariant(
     workPhase !== RenderPhase && workPhase !== CommitPhase,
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -814,6 +815,14 @@ function renderRoot(
         }
 
         const returnFiber = sourceFiber.return;
+
+        checkForWrongSuspensePriorityInDEV(
```
sebmarkbage: Let's move this check into `throwException` instead of duplicating the brand check etc. I'm moving around things in throwException and I wouldn't have thought to look here for another "thennable" thing that might be related to my change.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -2111,6 +2121,82 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
...
+      // have no rigorous way to distinguish work that was scheduled at user-
+      // blocking priority from work that expired a bit and was "upgraded" to
+      // a higher priority. That's because we don't schedule separate callbacks
+      // for every level, only the highest priority level per root. The priority
+      // of subsequent levels is inferred from the expiration time, but this is
+      // an imprecise heuristic.
+      //
+      // However, we do store the last discrete pending update per root. So we
+      // can reliably compare to that one. (If we broaden this warning to include
+      // high pri updates that aren't discrete, then this won't be sufficient.)
```
sebmarkbage: We really should have this warning too since it's pretty important to fix the mousemove events to have this property and we really shouldn't suspend there.

sebmarkbage: We can get back to it later. I suspect I'll need to track some new field for suspendIfNeeded. Maybe we can use that.

PR #15496, opened 2019-04-25
Event API: follow up fixes for FocusScope + context changes

trueadm:
This PR contains some follow up fixes as per comments in https://github.com/facebook/react/pull/15487 from @giuseppeg. This also adds a new helper method for getting the current active document `context.getActiveDocument`, which removes a bunch of code in places. I also added `context.isTargetWithinEventResponderScope` and `context.requestResponderOwnership` to deal with ownership control of the same responder.

### DIFF
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -86,7 +86,7 @@ const eventListeners:
     > = new PossiblyWeakMap();
 
 let currentTimers = new Map();
-let currentLocalOwners = new Map();
+let currentResponderOwners = new Map();
 let currentOwner = null;
 let currentInstance: null | ReactEventComponentInstance = null;
 let currentEventQueue: null | EventQueue = null;
@@ -289,7 +289,7 @@ const eventResponderContext: ReactResponderContext = {
       .responder;
     return (
       currentOwner === currentInstance ||
-      currentLocalOwners.get(responder) === currentInstance
+      currentResponderOwners.get(responder) === currentInstance
     );
   },
   requestGlobalOwnership(): boolean {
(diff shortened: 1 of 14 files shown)


packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -71,6 +71,7 @@ const rootEventTypesToEventComponentInstances: Map<
   DOMTopLevelEventType | string,
   Set<ReactEventComponentInstance>,
 > = new Map();
+global.foo = rootEventTypesToEventComponentInstances;
```
necolas: heh

packages/react-events/src/FocusScope.js
```diff
@@ -37,95 +35,112 @@ function focusFirstChildEventTarget(
...
+    }
+  },
+  onMount(
+    context: ReactResponderContext,
+    props: FocusScopeProps,
+    state: FocusScopeState,
+  ): void {
     if (props.restoreFocus) {
-      state.nodeToRestore = document.activeElement;
+      state.nodeToRestore = context.getActiveDocument().activeElement;
```
necolas: I think we'll need to do something like check for `blur` events that have a `relatedTarget` within the event component subtree, that way we can know the last element focused (`blurEvent.target`) before focus moved into the FocusScope instance. This should give the expected result in cases where the FocusScope is mounted well before the focus moves into it.

packages/react-events/src/FocusScope.js
```diff
@@ -135,17 +150,23 @@ const FocusScopeResponder = {
     context: ReactResponderContext,
     props: FocusScopeProps,
     state: FocusScopeState,
-  ) {
-    if (props.restoreFocus && state.nodeToRestore !== null) {
+  ): void {
+    if (
+      props.restoreFocus &&
```
necolas: Perhaps we could both simplify the responsibilities of this component and make it more flexible for users if we instead had a callback API so we can leave it up to people to shift the focus themselves, in case they need custom behaviour.

```js
const onFocusIn = (event) => {
  // if there's a cancel button in the scope, move focus there instead of to the first focusable element
  if (cancelRef.current != null) {
    focusElement(cancelRef.current);
  }
};

const onFocusOut = (event) => {
  if (event.target != null) {
    // restore focus if menu open-button if it is still open
    focusElement(event.target);
  } else {
    // shift focus to menu open-button if menu auto-collapsed when modal opened
    focusElement(fallbackElement);
  }
};

<FocusScope onFocusIn={onFocusIn} onFocusOut={onFocusOut}>

packages/react-events/src/FocusScope.js
trueadm: Sounds good. We should do this later at some point.

packages/react-events/src/FocusScope.js
trueadm: Sounds like a feature I was going to add anyway. I just added this to add a simple autofocus for modal cases.

PR #15498, opened 2019-04-25
upgrade prettier to 1.17.0

threepointone:
prettier started failing with my sublime text plugin; noted a fix was available. updated prettier, and ran it on all files. also updates .prettierrc.js's to use 'babel' as a parser as per it's new config.

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -8,7 +8,7 @@ module.exports = {
   jsxBracketSameLine: true,
   trailingComma: 'es5',
   printWidth: 80,
-  parser: 'babylon',
+  parser: 'babel',
 
   overrides: [
     {
--- a/fixtures/attribute-behavior/src/App.js
+++ b/fixtures/attribute-behavior/src/App.js
@@ -866,14 +866,12 @@ class App extends React.Component {
     // Sort
     switch (sortOrder) {
       case ALPHABETICAL:
-        return filteredAttributes.sort(
-          (attr1, attr2) =>
-            attr1.name.toLowerCase() < attr2.name.toLowerCase() ? -1 : 1
+        return filteredAttributes.sort((attr1, attr2) =>
+          attr1.name.toLowerCase() < attr2.name.toLowerCase() ? -1 : 1
         );
       case REV_ALPHABETICAL:
-        return filteredAttributes.sort(
-          (attr1, attr2) =>
-            attr1.name.toLowerCase() < attr2.name.toLowerCase() ? 1 : -1
+        return filteredAttributes.sort((attr1, attr2) =>
+          attr1.name.toLowerCase() < attr2.name.toLowerCase() ? 1 : -1
         );
       case GROUPED_BY_ROW_PATTERN: {
         return filteredAttributes.sort((attr1, attr2) => {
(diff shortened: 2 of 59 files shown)


threepointone:
(I should just burn down my fork and start over. stupid merges.)

packages/events/EventPluginUtils.js
```diff
@@ -40,15 +40,15 @@ if (__DEV__) {
...
+      ? 1
+      : 0;
 
     const instancesIsArr = Array.isArray(dispatchInstances);
     const instancesLen = instancesIsArr
       ? dispatchInstances.length
       : dispatchInstances
-        ? 1
-        : 0;
+      ? 1
```
threepointone: dunno how I feel about this. I kinda like indents in nested ternaries.

threepointone:
I'll figure out the CI thing

.prettierrc.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   jsxBracketSameLine: true,
   trailingComma: 'es5',
   printWidth: 80,
-  parser: 'babylon',
+  parser: 'babel',
```
milesj: IMO, you should omit the parser and just let it figure it out automatically.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationElements-test.js
```diff
@@ -255,8 +255,7 @@ describe('ReactDOMServerIntegration', () => {
           <div>
             {[['a'], 'b']}
             <div>
```
ahtee: strange if prettier did this, since the above patterns have more of a `new line` pattern... like
```
<a href=''>
   span
</a>
```

.prettierrc.js
ahtee: How would it automatically figure out the parser without this? 

> Prettier automatically infers the parser from the input file path, so you shouldn't have to change this setting.

https://prettier.io/docs/en/options.html#parser

milesj: You don't need to define `parser`. Prettier will automatically change the `parser` based on the file type and extension.

fixtures/attribute-behavior/src/App.js
```diff
@@ -866,14 +866,12 @@ class App extends React.Component {
     // Sort
     switch (sortOrder) {
       case ALPHABETICAL:
```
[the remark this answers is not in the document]
threepointone: yup, I didn't make any changes myself.

.prettierrc.js
ahtee: I suppose it doesn't hurt too bad to have it explicitly stated here then.

milesj: Do note that it will break if Prettier is ever used on more than just `.js` files.

keyz: @threepointone use `babel-flow` here to let it handle callable generics properly: https://tinyurl.com/y2o3mpvu

package.json
```diff
@@ -67,7 +68,7 @@
     "mkdirp": "^0.5.1",
     "ncp": "^2.0.0",
     "object-assign": "^4.1.1",
-    "prettier": "1.13.7",
+    "prettier": "1.17.0",
```
SimenB: 1.18 is out 👀

threepointone:
Closing this

PR #15500, opened 2019-04-25
React events: fix press end event dispatching

necolas:
This patch fixes an issue related to determining whether the end event occurs
within the responder region. Previously we only checked if the event target was
within the responder region for moves, otherwise we checked if the target was
within the event component. Since the dimensions of the child element can
change after activation, we need to recalculate the responder region before
deactivation as well if the target is not within the event component.

Ref #15257

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -52,10 +52,16 @@ type PressState = {
   isPressWithinResponderRegion: boolean,
   longPressTimeout: null | Symbol,
   pointerType: PointerType,
-  pressTarget: null | Element | Document,
+  pressTarget: null | Element,
   pressEndTimeout: null | Symbol,
   pressStartTimeout: null | Symbol,
-  responderRegion: null | $ReadOnly<{|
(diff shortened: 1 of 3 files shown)


sebmarkbage:
Regression tests?

packages/react-events/src/Press.js
```diff
@@ -551,18 +577,34 @@ const PressResponder = {
...
       case 'touchend': {
         if (state.isPressed) {
-          // Ignore unrelated keyboard events
+          // Ignore unrelated keyboard events and verify press is within
+          // responder region for non-keyboard events
           if (pointerType === 'keyboard') {
             if (!isValidKeyPress(nativeEvent.key)) {
               return;
             }
+            // If the event isn't within the event target, check if we're still within the responder region.
```
trueadm: Shouldn't this be:

`If the event target isn't within the press target`?

necolas:
What additional test coverage do you want?

trueadm:
@necolas I see you added tests. I was probably looking at an older commit.

PR #15502, opened 2019-04-25
Add Batched Mode

acdlite:
React has an unfortunate quirk where updates are sometimes synchronous — where React starts rendering immediately within the call stack of `setState` — and sometimes batched, where updates are flushed at the end of the current event. Any update that originates within the call stack of the React event system is batched. This encompasses most updates, since most updates originate from an event handler like `onClick` or `onChange`. It also includes updates triggered by lifecycle methods or effects. But there are also updates that originate outside React's event system, like timer events, network events, and microtasks (promise resolution handlers). These are not batched, which results in both worse performance (multiple render passes instead of single one) and confusing semantics.

Ideally all updates would be batched by default. Unfortunately, it's easy for components to accidentally rely on this behavior, so changing it could break existing apps in subtle ways.

One way to move to a batched-by-default model is to opt into Concurrent Mode (still experimental). But Concurrent Mode introduces additional semantic changes that apps may not be ready to adopt.

This commit introduces an additional mode called Batched Mode. Batched Mode enables a batched-by-default model that defers all updates to the next React event. Once it begins rendering, React will not yield to the browser until the entire render is finished.

Batched Mode is superset of Strict Mode. It fires all the same warnings. It also drops the forked Suspense behavior used by Legacy Mode, in favor of the proper semantics used by Concurrent Mode.

I have not added any public APIs that expose the new mode yet. I'll do that in subsequent commits.

### DIFF
--- a/packages/react-noop-renderer/src/createReactNoop.js
+++ b/packages/react-noop-renderer/src/createReactNoop.js
@@ -18,6 +18,7 @@ import type {Thenable} from 'react-reconciler/src/ReactFiberScheduler';
 import type {Fiber} from 'react-reconciler/src/ReactFiber';
 import type {UpdateQueue} from 'react-reconciler/src/ReactUpdateQueue';
 import type {ReactNodeList} from 'shared/ReactTypes';
+import type {RootTag} from 'shared/ReactRootTags';
 
 import * as Scheduler from 'scheduler/unstable_mock';
 import {createPortal} from 'shared/ReactPortal';
@@ -32,6 +33,7 @@ import enqueueTask from 'shared/enqueueTask';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import warningWithoutStack from 'shared/warningWithoutStack';
 import {enableEventAPI} from 'shared/ReactFeatureFlags';
+import {ConcurrentRoot, BatchedRoot, LegacyRoot} from 'shared/ReactRootTags';
 
 type EventTargetChildElement = {
   type: string,
@@ -901,21 +903,12 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
       return getPendingChildren(container);
     },
 
-    getOrCreateRootContainer(
-      rootID: string = DEFAULT_ROOT_ID,
-      isBatched: boolean,
-      isConcurrent: boolean,
-    ) {
+    getOrCreateRootContainer(rootID: string = DEFAULT_ROOT_ID, tag: RootTag) {
       let root = roots.get(rootID);
       if (!root) {
         const container = {rootID: rootID, pendingChildren: [], children: []};
         rootContainers.set(rootID, container);
-        root = NoopRenderer.createContainer(
-          container,
-          isBatched,
-          isConcurrent,
-          false,
-        );
+        root = NoopRenderer.createContainer(container, tag, false);
         roots.set(rootID, root);
       }
       return root.current.stateNode.containerInfo;
@@ -923,8 +916,6 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
 
     // TODO: Replace ReactNoop.render with createRoot + root.render
     createRoot() {
-      const isBatched = true;
-      const isConcurrent = true;
       const container = {
         rootID: '' + idCounter++,
         pendingChildren: [],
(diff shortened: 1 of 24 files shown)


packages/react-noop-renderer/src/createReactNoop.js
```diff
@@ -832,77 +832,160 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
...
+          isBatched,
+          isConcurrent,
+          false,
+        );
         roots.set(rootID, root);
       }
       return root.current.stateNode.containerInfo;
     },
 
+    // TODO: Remplace ReactNoop.render with createRoot + root.render
```
acdlite: TODO: Replace "Remplace" with "Replace"

sophiebits:
Yasss. I found myself wishing for batching literally earlier today. Hope this gets released soon.

Why couple it to strict mode? It doesn't require the strict semantics, does it? It feels to me like strict and batched are orthogonal (but both required for concurrent compatibility).

* strict: (compatibility with) once a render is started, it might not be always finished
* batched: once setState is called, there might be time before the state update happens (already the case in an event handler)

I am wary of too many flags though… 😬

acdlite:
@sophiebits It does require Strict Mode because of Suspense. The quirky, forked semantics we use for Suspense in Legacy Mode today aren't compatible with the newer features we're planning to add, so we want to move everyone over to the "proper" semantics used by Concurrent Mode instead. We might even add

> I am wary of too many flags though… 😬

Yeah I share this concern.

sebmarkbage:
> Suspense in Legacy Mode today aren't compatible with the newer features

The more immediate problem is that it isn't compatible across mode but also the Legacy Mode experience is much worse semantics. E.g. it breaks a lot of subtle patterns like that you can't rely on layout being available in componentDidMount, all refs possibly being null for a bit. This makes it much harder than it needs to be to write other code in a Suspense tree, but also if you write code that is compatible with Batched or Concurrent Mode, it might not be compatible in Legacy Mode (as in it'll crash). So the downgrade scenario is really bad.

I like to think of Suspense not existing in Legacy Mode. It's just a Batched Mode feature.

acdlite:
I don't think the implementation would look that different. If/when we make it per root, I would replace `fiber.mode & BatchedMode` with something more like `root.mode & BatchedMode`.

acdlite:
To-dos after discussing with @sebmarkbage:

- We're planning to get rid of the `<ConcurrentMode />` component API in favor of opting in the entire root. That means we don't need to store the mode on the Fiber. I'm going to move that work into a separate PR, but in anticipation of that change, the `createContainer` API exposed by the reconciler should accept an enum of different root types instead of separate `isBatched` and `isConcurrent` booleans, which don't make sense because you can't actually mix them. I had previously considered that this should be a bitmask, but since there are only three distinct types, an enum is probably less confusing.
- Batched Mode should support deprioritizing `hidden` trees, though React still shouldn't yield when it comes back to work on the next level.
- Batched Mode should schedule work at the same priority level that would be schedule in Concurrent Mode. This means that Batched Mode is almost identical to `disableYielding` feature flag, except its per root and it also affects Suspense semantics.

sebmarkbage:
> Batched Mode should schedule work at the same priority level that would be schedule in Concurrent Mode.

Wait this isn't quite right. Batched Mode should only ever have three different expiration times. Sync, Batched or Offscreen. All of which are fixed constants and not.

Scheduler.next, high vs normal pri event etc. shouldn't matter. (Discrete matters for when we flush early though. Batched is discrete.)

The level scheduled with the scheduler is kind of arbitrary. Can be Normal for Batched. Offscreen can be lower I guess.

acdlite:
That's what I meant, I'm not getting rid of the `Batched` expiration time.

acdlite:
Changed my mind. I think Scheduler priority for Batched updates should be Immediate (except for hidden subtrees which will be downgraded to Idle). Mostly for implementation reasons: need to be able to infer the priority level given an expiration time (`inferPriorityFromExpirationTime`). I think there's an argument that it should be Normal, and another argument that Immediate actually does make sense, but the priority system is underdesigned right now anyway, so I'm going to go with the thing that is most straightforward to implement.

acdlite:
I'm going to deal with `hidden` to a separate PR because I want to remove <ConcurrentMode /> first.

acdlite:
So this is ready for another review.

NE-SmallTown:
> Ideally all updates would be batched by default. Unfortunately, it's easy for components to accidentally rely on this behavior, so changing it could break existing apps in subtle ways.

I'm a little confused here, how could you do to implement this? I.e. if something out of control of React(timer, network...), how to make it still be batched?(I mean, not more extra api like `flush()`)

NE-SmallTown:
What does 'React event' mean in 'to the next React event'?

PR #15504, opened 2019-04-25
Add ReactDOM.unstable_createSyncRoot

acdlite:
- `ReactDOM.unstable_createRoot` creates a Concurrent Mode root.
- `ReactDOM.unstable_createSyncRoot` creates a Batched Mode root. It
does not support `createBatch`.
- `ReactDOM.render` creates a Legacy Mode root. It will eventually be
deprecated and possibly moved to a separate entry point, like
`react-dom/legacy`.

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -53,7 +53,7 @@ import {
   accumulateTwoPhaseDispatches,
   accumulateDirectDispatches,
 } from 'events/EventPropagators';
-import {LegacyRoot, ConcurrentRoot} from 'shared/ReactRootTags';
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/client/ReactDOM.js
```diff
@@ -186,17 +186,15 @@ type Batch = FiberRootBatch & {
...
-    children: ReactNodeList,
-    callback: ?() => mixed,
-  ): Work,
-  createBatch(): Batch,
 
   _internalRoot: FiberRoot,
 };
 
-function ReactBatch(root: ReactRoot) {
+type RootWithBatchingAPI = Root & {
```
sebmarkbage: Can we call this "ReactRoot", and instead call the other one "ReactSyncRoot"? These are just internal but these names tend to leak into official types in Flow/TypeScript.

PR #15505, opened 2019-04-25
React Events: core API documentation

necolas:
Initial docs for the core API.

@trueadm Feel free to make edits. I wasn't entirely sure how to describe some of the new ownership APIs (global vs responder)

Ref #15257

### DIFF
--- a/packages/react-events/docs/Hover.md
+++ b/packages/react-events/docs/Hover.md
@@ -1,0 +1,74 @@
+## Hover
+
+The `Hover` module responds to hover events on the element it wraps. Hover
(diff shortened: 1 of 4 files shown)


packages/react-events/docs/Hover.md
```diff
@@ -0,0 +1,74 @@
+## Hover
+
+The `Hover` module responds to hover events on the element it wraps. Hover
+events are only dispatched for `mouse` pointer types. Hover begins when the
```
TrySound: mouse and pen like in specified type?

PR #15516, opened 2019-04-26
React Events: add onFocusVisibleChange to Focus

necolas:
Called when focus visibility changes. Focus is only considered visible if a
focus event occurs after keyboard navigation. This provides a way for people to
provide visual focus styles for keyboard accessible UIs without those styles
appearing if focus is triggered by mouse, touch, pen.

Still working on incorporating a couple more details around when focus is considered visible.

Ref #15257

### DIFF
--- a/packages/react-events/src/Focus.js
+++ b/packages/react-events/src/Focus.js
@@ -19,14 +19,16 @@ type FocusProps = {
   onBlur: (e: FocusEvent) => void,
   onFocus: (e: FocusEvent) => void,
   onFocusChange: boolean => void,
+  onFocusVisibleChange: boolean => void,
 };
 
 type FocusState = {
-  isFocused: boolean,
   focusTarget: null | Element | Document,
+  isFocused: boolean,
+  isLocalFocusVisible: boolean,
 };
 
-type FocusEventType = 'focus' | 'blur' | 'focuschange';
+type FocusEventType = 'focus' | 'blur' | 'focuschange' | 'focusvisiblechange';
 
 type FocusEvent = {|
   target: Element | Document,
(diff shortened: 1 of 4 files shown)


packages/react-events/src/Focus.js
```diff
@@ -145,6 +191,57 @@ const FocusResponder = {
...
+
+      case 'keydown':
+      case 'keypress':
+      case 'keyup': {
+        const nativeEvent = event.nativeEvent;
+        if (
+          nativeEvent.key === 'Tab' &&
+          !(nativeEvent.metaKey || nativeEvent.altKey || nativeEvent.ctrlKey)
+        ) {
+          isGlobalFocusVisible = true;
```
necolas: It's probably not ideal to have global focus state being checked / modified in every instance. Should this be handled by a module outside of Flare (using native APIs) that we query for global focus visibility info? Any other ideas?

trueadm: We could add a `context.isEventComponentWithinResponderScope` API that checks if a parent event component in the tree has the same responder and bail-out of this logic?

necolas:
I noticed a couple of odd things happening while working on–and unrelated to changes in–this branch. First, the `onRootEvent` handler for `Focus` seems to be called twice for each `keydown` and `keyup` event. Second, in Safari pressing any of the Shift, Ctrl, Opt, Cmd keys while focused is causing 'mousemove' events to be dispatched to Focus's `onRootEvent` handler.

https://fburl.com/anpad6mb

trueadm:
@necolas This wasn't merged before the other PR that should have addressed the double event issue (the PR for using proper keys for events). I wonder if that fixes this issue. The other bugs sounds like yet another Safari thing we need to handle as an edge case. :/

Issue #15512, opened 2019-04-26
Suspense commits pending trees in Concurrent Mode under heavy load


gaearon:
Seems similar to https://github.com/facebook/react/pull/15424.
Minimal repro case: https://codesandbox.io/s/jzq8roj51y.

To reproduce, open in new tab (https://jzq8roj51y.codesandbox.io/), let first item's details to load, and then hold-press "down". You'll likely (not always) see Details tab flicker and later crash/warn in console.

![Screen Recording 2019-04-26 at 04 29 PM](https://user-images.githubusercontent.com/810438/56821850-5a644c00-6847-11e9-9fb9-061677cdf864.gif)

MarioCadenas:
I've been trying to reproduce this, and the only way i can reproduce it, is by connecting to React devtools, but I can't still get the console error.

![demo](https://im.ezgif.com/tmp/ezgif-1-c674fd786edf.gif)

submetu:
I too had to connect to the React Dev Tools in order to produce this bug

gaearon:
I can't repro this anymore with either version.

PR #15517, opened 2019-04-26
Invariant that throws when committing wrong tree

acdlite:
If React finishes rendering a tree, delays committing it (e.g. Suspense), then subsequently starts over or renders a new tree, the pending tree is no longer valid. That's because rendering a new work-in progress mutates the old one in place.

The current structure of the work loop makes this hard to reason about because, although `renderRoot` and `commitRoot` are separate functions, they can't be interleaved. If they are interleaved by accident, it either results in inconsistent render output or invariant violations that are hard to debug.

This commit adds an invariant that throws if the new tree is the same as the old one. This won't prevent all bugs of this class, but it should catch the most common kind.

To implement the invariant, I store the finished tree on a field on the root. We already had a field for this, but it was only being used for the unstable `createBatch` feature.

A more rigorous way to address this type of problem could be to unify `renderRoot` and `commitRoot` into a single function, so that it's harder to accidentally interleave the two phases. I plan to do something like this in a follow-up.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberRoot.js
+++ b/packages/react-reconciler/src/ReactFiberRoot.js
@@ -42,7 +42,7 @@ type BaseFiberRootProperties = {|
     | Map<Thenable, Set<ExpirationTime>>
     | null,
 
-  pendingCommitExpirationTime: ExpirationTime,
+  finishedExpirationTime: ExpirationTime,
   // A finished work-in-progress HostRoot that's ready to be committed.
   finishedWork: Fiber | null,
   // Timeout handle returned by setTimeout. Used to cancel a pending timeout, if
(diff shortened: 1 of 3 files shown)


gaearon:
Was it expected to actually _fix_ https://github.com/facebook/react/issues/15512? I still see the number glitching but with this change, I can't get neither the invariant nor the warning to trigger anymore.

gaearon:
However, I am also *not* seeing `commitRoot` and `renderRoot` being interleaved on master when the error happens:

Screen Shot 2019-04-27 at 02 16 28
Screen Shot 2019-04-27 at 02 16 40

Screen Shot 2019-04-27 at 02 17 25

Does this mean the root cause is different, or is my method flawed?

acdlite:
It fixes it but I didn’t isolate it to a test case yet. I know what’s happening at a high level but I don’t know what specific sequence of events needs to happen to trigger it. I’ll look into it more on Monday.

PR #15519, opened 2019-04-27
Flush act() only on exiting outermost call

threepointone:
Just starting this PR up to discuss the changes. The actual code changes aren't major. The big (breaking?) change is that nested acts won't work anymore.


acdlite:
Will wait for https://github.com/facebook/react/pull/15591 to land before reviewing (or I can review now if you base this one on top of that one)

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
```diff
@@ -119,6 +118,13 @@ function act(callback: () => Thenable) {
         called = true;
         result.then(
           () => {
+            if (actingUpdatesScopeDepth !== 1) {
```
acdlite: Is there any scenario where `actingUpdatesScopeDepth` would decrement twice, and this would be less than 1? Maybe should change to `actingUpdatesScopeDepth < 1` just in case.

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -285,14 +285,13 @@ describe('ReactTestUtils.act()', () => {
         return ctr;
       }
       const el = document.createElement('div');
+      act(() => {
```
acdlite: Why did you move this to a separate `act`? I would expect moving the assertion outside `act` is sufficient.

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -371,18 +370,23 @@ describe('ReactTestUtils.act()', () => {
...
+      await act(async () => {}); // #just-react-things
 
       expect(div.innerHTML).toBe('1');
       expect(ctr).toBe(2);
+
+      // this may seem odd, but it matches browser behaviour -
+      // a flash of "0" followed by "1"
+
+      // now you probably wouldn't write a test like this for your app
+      // you'd await act(async () => { ReactDOM.render(<App/> })
```
acdlite: So then why did you write it this way?

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -371,18 +370,23 @@ describe('ReactTestUtils.act()', () => {
...
-        expect(ctr).toBe(1);
+      act(() => {
+        ReactDOM.render(<App callback={() => ctr++} />, div);
       });
-      // this may seem odd, but it matches user behaviour -
-      // a flash of "0" followed by "1"
+      expect(div.innerHTML).toBe('0');
+      expect(ctr).toBe(1);
+
+      await act(async () => {}); // #just-react-things
```
acdlite: Why did you remove the `await` from the earlier `act`?

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
threepointone: the effect wouldn't flush effects until the act exits, only after which the timeout in the component would start, failing the assertion. hence, 2 act()s.

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
threepointone: I wanted to specifically test the intermediate sync state. this test is named badly (and frankly on looking at it now, it doesn't seem super useful).

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
threepointone: like mentioned, just wanted to make sure the intermediate state was what I'd expected.

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
threepointone: rather, actingUpdatesScopeDepth > 1

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
acdlite: It's not clear to me at all what behavior you're actually trying to test. More comments could help.

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
acdlite: Why did you remove the `await` then?

threepointone:
so many merge conflicts, so I redid it here https://github.com/facebook/react/pull/15682

PR #15521, opened 2019-04-28
Update link badge

Hongarc:
This commit will auto-detect your GitHub license


threepointone:
We don’t expect to ever change the license, so this wouldn’t be very useful (and likely slower). Thanks tho!

PR #15529, opened 2019-04-29
remove redundant if statement in onmessage handler  & fix typo

NE-SmallTown:
fix something as the title said

### DIFF
--- a/packages/scheduler/src/forks/SchedulerHostConfig.default.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.default.js
@@ -152,7 +152,7 @@ if (
   // We use the postMessage trick to defer idle work until after the repaint.
   const channel = new MessageChannel();
   const port = channel.port2;
-  channel.port1.onmessage = function(event) {
+  channel.port1.onmessage = function() {
     isMessageEventScheduled = false;
 
     const prevScheduledCallback = scheduledHostCallback;
@@ -172,11 +172,6 @@ if (
         didTimeout = true;
       } else {
         // No timeout.
-        if (!isAnimationFrameScheduled) {
-          // Schedule another animation callback so we retry later.
-          isAnimationFrameScheduled = true;
-          requestAnimationFrameWithTimeout(animationTick);
-        }
         // Exit without invoking the callback.
         scheduledHostCallback = prevScheduledCallback;
         timeoutTime = prevTimeoutTime;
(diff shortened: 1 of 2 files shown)


packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -152,48 +152,6 @@ if (
...
-      if (prevTimeoutTime !== -1 && prevTimeoutTime <= currentTime) {
-        // Exceeded the timeout. Invoke the callback even though there's no
-        // time left.
-        didTimeout = true;
-      } else {
-        // No timeout.
-        if (!isAnimationFrameScheduled) {
-          // Schedule another animation callback so we retry later.
-          isAnimationFrameScheduled = true;
-          requestAnimationFrameWithTimeout(animationTick);
```
NE-SmallTown: there is no `animationTick` because we declare `animationTick` below(use `const`), the reason there is no throw is seems that the we will never enter the `if (!isAnimationFrameScheduled) {` statement(I don't know why), so just for now, I just move the whole function under the `animationTick  declaration`

acdlite:
Can you write a unit test please?

NE-SmallTown:
(maybe)I will do that when I wake up :)

packages/scheduler/src/forks/SchedulerHostConfig.default.js
acdlite: This is inside a clousre. `animationTick` is outside the closure. It will already be defined by the time the `onmessage` handler fires.

NE-SmallTown:
@acdlite comment below

```
if (frameDeadline - currentTime <= 0) {
  // There's no time left in this idle period. Check if the callback has
  // a timeout and whether it's been exceeded.
 
  if (prevTimeoutTime !== -1 && prevTimeoutTime <= currentTime) {
    // Exceeded the timeout. Invoke the callback even though there's no
    // time left.
    didTimeout = true;
  } else {
    // No timeout.     
    --------->>> any time we enter this block, 'isAnimationFrameScheduled' will be true 
    --------->>> because before this we call requestHostCallback which will do the same 
    --------->>> logic as below 5 lines, so I think we can just remove below 5 lines?
    if (!isAnimationFrameScheduled) {
        // Schedule another animation callback so we retry later.
        isAnimationFrameScheduled = true;
        requestAnimationFrameWithTimeout(animationTick);
     }

     // Exit without invoking the callback.
     scheduledHostCallback = prevScheduledCallback;
     timeoutTime = prevTimeoutTime;
     return;
   }
}
```

packages/scheduler/src/forks/SchedulerHostConfig.default.js
NE-SmallTown: Ah, yea, sorry, I didn't notice that

PR #15532, opened 2019-04-29
Remove <React.unstable_ConcurrentMode /> element type

acdlite:
We decided to remove this because we're not convinced it's practical to opt a particular subtree into Concurrent Mode without opting in the entire root. The recommendation is to use `createRoot` instead.


NE-SmallTown:
Could we preserve this(concurrent mode) and just allow it be used in root?

acdlite:
@NE-SmallTown It's not impossible, but it creates weird edge cases that we'd prefer not to support, given that we haven't found a compelling enough use case for it.

PR #15533, opened 2019-04-29
Support `hidden` deprioritization in Batched Mode

acdlite:
Hidden trees will tear and commit in a separate render, as in Concurrent Mode. Unlike Concurrent Mode, once the hidden tree begins, it will finish without yielding.


acdlite:
Will consider as part of larger Hidden API rethink

PR #15545, opened 2019-04-30
React Events: fix hover state of element with descendants

necolas:
Ensure the hover state is preserved while moving the pointer over descendants
of a Hover component.

Ref #15257


necolas:
See #15546

trueadm:
I covered this in https://github.com/facebook/react/pull/15546 too haha.

PR #15560, opened 2019-05-03
Event API: Support press reentry for pointer events

trueadm:
This PR adds support to the `Press` event responder, so when a user presses into a hit zone, moves their pointer out of the zone, then re-enters, that the zone re-activates again. I've added tests to demonstrate this too.


necolas:
This should happen for touch and pen too, it's part of the RN Pressable behavior

trueadm:
Okay I can update the PR :)

trueadm:
@necolas I've made that change.

PR #15565, opened 2019-05-04
add useStateObject

ofirzBM:
useStateObject is using useState for create single object and exposing it's properties as accessors.
It removes the use of [val, setVal] array and use property accessor for both get and set.

Usage:
`const state = getStateObject({ test: 1 });

const update = (val) => {
	state.test = val;
}
return <div>{state.test}</div>`

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
@@ -1814,4 +1814,23 @@ describe('ReactHooks', () => {
     Scheduler.flushAll();
     expect(root).toMatchRenderedOutput('hello');
   });
+          
+  it('useStateObject: builds a state object', () => {
+    const {useStateObject, useEffect} = React;
+
+    function Counter() {
+      const state = useStateObject({ count1: 0, count2: 0});
+      useEffect( () => {
+        state.count1 ++;
+        state.count2 += 2;
+      });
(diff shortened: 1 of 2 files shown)


ofirzBM:
> Is there a reason you believe this should be in core and not just a library?

Thank you for your feedback,
I don't think so, but I believe this approach is easier to use and read.

threepointone:
It’s unlikely we’re going to add this to our list of primitives. I’m going to close this PR, I would recommend you make a reusable library of this instead.

ofirzBM:
> It’s unlikely we’re going to add this to our list of primitives. I’m going to close this PR, I would recommend you make a reusable library of this instead.

Sad, but thanks again for your time.
Beside that, how do you find this approach? is it easy to understand? use?

threepointone:
It’s not bad! Don’t be discouraged that this didn’t land in react, it’s not a commentary on your code or abilities at all. React tries very hard to be a minimal api with only the basic blocks so that developers and users can make their own abstractions like yours. It would make sense to be directly in react only if it couldn’t be solved with an abstraction, or is so commonly used that it only makes sense to include it by default. 

So yes, your abstraction looks useful. As some feedback - while react devs also usually try to stay away from mutative apis where they manually mutate objects like above, but libraries like mobx show that there’s still appeal for the approach. There’s definitely an audience for your api, you just need to find it :)

Good luck!

PR #15567, opened 2019-05-04
Fix interaction tracing for batched update mounts

### LINKED ISSUE #15566: Batched updates break interaction tracing for mounts
This test works:
```js
let interaction;

const onRender = jest.fn();

SchedulerTracing.unstable_trace("initial_event", performance.now(), () => {
  const interactions = SchedulerTracing.unstable_getCurrent();
  expect(interactions.size).toBe(1);
  interaction = Array.from(interactions)[0];

  ReactDOM.render(
    <React.Profiler id="profiler" onRender={onRender}>
      <Text text="Hi" />
    </React.Profiler>,
    document.createElement("div")
  );
});

expect(interaction).toBeDefined();
expect(onRender).toHaveBeenCalledTimes(1);
expect(onRender.mock.calls[0][6]).toContain(interaction);
```

But if you wrap it in `act`, it fails:
```js
let interaction;

const onRender = jest.fn();

TestUtils.act(() => {
  SchedulerTracing.unstable_trace("initial_event", performance.now(), () => {
    const interactions = SchedulerTracing.unstable_getCurrent();
    expect(interactions.size).toBe(1);
    interaction = Array.from(interactions)[0];

    ReactDOM.render(
      <React.Profiler id="profiler" onRender={onRender}>
        <Text text="Hi" />
      </React.Profiler>,
      document.createElement("div")
    );
  });

  expect(interaction).toBeDefined();
  expect(onRender).toHaveBeenCalledTimes(1);
  expect(onRender.mock.calls[0][6]).toContain(interaction);
});
```

In the latter case, inside of `ReactFiberScheduler`, the interactions `Set` is empty.
https://github.com/facebook/react/blob/2e5d1a8b9e2c29418a27b24306e4c8d5f8681f4f/packages/react-reconciler/src/ReactFiberScheduler.js#L2119-L2126

Note that this does not seem to impact _updates_, only the initial _mount_. I suspect we just need to wrap a callback or something minor.

I came across this while trying to write profiling tests for the new DevTools. (https://github.com/bvaughn/react-devtools-experimental/pull/258)

bvaughn:
When a root is _mounting_ inside of a batched update (including the `act` test helper) the interactions `Set` in of `ReactFiberScheduler` is empty. This does not impact _updates_, only _mounts_.

The reason for this is that in that case, we never recorded the pending interactions on the root.

Technically the bug was related to batched updates, not `act`, but it seemed worth adding test coverage for both just in case there were changes in the future.

I came across this while trying to write profiling tests for the new DevTools (https://github.com/bvaughn/react-devtools-experimental/pull/258). I've tested this change in that PR and verified that it fixes the disconnected interactions.

Resolves #15566

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "@mattiasbuelens/web-streams-polyfill": "0.1.0"
   },
   "devEngines": {
-    "node": "8.x || 9.x || 10.x || 11.x"
+    "node": "8.x || 9.x || 10.x || 11.x || 12.x"
   },
   "jest": {
     "testRegex": "/scripts/jest/dont-run-jest-directly\\.js$"
--- a/packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
+++ b/packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
@@ -10,6 +10,7 @@
 let React;
 let ReactDOM;
 let ReactTestUtils;
+let SchedulerTracing;
 let act;
 let container;
 
@@ -29,6 +30,7 @@ describe('ReactTestUtils.act()', () => {
     React = require('react');
     ReactDOM = require('react-dom');
     ReactTestUtils = require('react-dom/test-utils');
+    SchedulerTracing = require('scheduler/tracing');
     act = ReactTestUtils.act;
     container = document.createElement('div');
     document.body.appendChild(container);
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -87,7 +87,7 @@
     "@mattiasbuelens/web-streams-polyfill": "0.1.0"
   },
   "devEngines": {
-    "node": "8.x || 9.x || 10.x || 11.x"
+    "node": "8.x || 9.x || 10.x || 11.x || 12.x"
```
bvaughn: This wasn't strictly necessary ^ but... 🤷‍♂

Issue #11805, opened 2017-12-08
Run benchmark trials in random order


gaearon:
Our current benchmark script runs all trials in a specific order.
You can try it by running:

```
cd scripts/bench
yarn
yarn start
```

(Note: it takes a looong time to finish.)

I think running all trials in a deterministic order like we do now may actually introduce issues. Such as if some other process starts grabbing CPU in the middle of the trials, a specific set of trials (e.g. "HN" benchmark on "base" repo)  will be affected, and the results skewed.

But if we ran all trials in a random order, I would expect any temporary changes to be evenly distributed between random runs, and thus not favor any particular version or benchmark. Wouldn't that be better?

Does this make sense? Tagging as a good issue to look into, but I don't promise this would be easy.

mannanali413:
@gaearon i can pick this up. :)

gaearon:
Sounds great!

mannanali413:
@gaearon please have a look at the PR and let me know if the randomisation introduced would meet the requirement.
Cheers :)

agasnikov:
I'm on it

PR #15571, opened 2019-05-05
Allow running benchmarks in interleaved (local/remote) or fully random order

agasnikov:
**This PR addresses issue #11805**

Major changes:
- The control flow is noticeably changed. Individual becnhmarks can't be prepared/run one by one, so the process now consists of the following steps:
   - build React bundles for local and remote repos;
   - define the list of benchmarks to be run (for local and remote);
   - build all benchmarks, copy them to the server root folder;
   - prepare the runner queue (interleaved or random);
   - start HTTP server and run individual benchmarks according to the queue;
   - gather stats, calculate averages;
   - print results;
- Both local and remote benchmarks now served with the same server instance simultaneously. It allows running benchmarks in any order;
- All benchmarks are served from the `/scripts/bench/public` folder, where they are copied on the preparation step;
- Support for mismatching list of becnhmarks on local and remote branch (i.e. a new benchmark is created locally and not exists on the remote master);
- Support for running benchmarks in the interleaved (default) or random (pass `--order=random`) order;
- **Fix:** the original script seems to be serving only from the local branch becnhmark folder, regardless whether it's local or remote becnhmark run;

### DIFF
--- a/scripts/bench/.gitignore
+++ b/scripts/bench/.gitignore
@@ -1,2 +1,4 @@
 react-dom.production.min.js
 react.production.min.js
+public/
+remote-repo/
--- a/scripts/bench/benchmark.js
+++ b/scripts/bench/benchmark.js
@@ -7,17 +7,13 @@ const stats = require('stats-analysis');
 const config = require('lighthouse/lighthouse-core/config/perf-config');
 const spawn = require('child_process').spawn;
 const os = require('os');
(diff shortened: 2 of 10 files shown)


agasnikov:
Could someone help me out with the failed CI checks? As far as I understand, it failed on the `node ./scripts/prettier/index` command, but `yarn run prettier-all` throws no errors for me.

PR #15576, opened 2019-05-06
Add React.unstable_createEventComponent

necolas:
Introduces a new unstable API for creating event components from event responders.

Ref #15257

### DIFF
--- a/packages/react/src/ReactCreateEventComponent.js
+++ b/packages/react/src/ReactCreateEventComponent.js
@@ -1,0 +1,36 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ * @flow
+ */
+
+import type {ReactEventComponent, ReactEventResponder} from 'shared/ReactTypes';
+import {enableEventAPI} from 'shared/ReactFeatureFlags';
+
+import {REACT_EVENT_COMPONENT_TYPE} from 'shared/ReactSymbols';
+
+type EventComponentConfig = {
+  displayName: string,
+  props: null | Object,
+  responder: ReactEventResponder,
+};
+
+export function createEventComponent({
+  displayName,
+  responder,
+}: EventComponentConfig): ?ReactEventComponent {
+  if (enableEventAPI) {
+    const eventComponent = {
+      $$typeof: REACT_EVENT_COMPONENT_TYPE,
+      displayName: displayName,
+      props: null,
+      responder: responder,
+    };
+    if (__DEV__) {
+      Object.freeze(eventComponent);
+    }
+    return eventComponent;
+  }
+}
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -899,9 +899,6 @@ export function mountEventComponent(
     const rootContainerInstance = ((eventComponentInstance.rootInstance: any): Container);
     const rootElement = rootContainerInstance.ownerDocument;
     const responder = eventComponentInstance.responder;
-    if (__DEV__) {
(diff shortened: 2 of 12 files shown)


necolas:
The bundle size of `react` and all the `react-event` modules has increased. Not sure what the right way to set this up is

packages/react/src/ReactCreateEventComponent.js
```diff
@@ -0,0 +1,34 @@
...
+  props: null | Object,
+  responder: ReactEventResponder,
+};
+
+export function createEventComponent({
+  displayName,
+  props,
+  responder,
+}: EventComponentConfig): ReactEventComponent {
+  const eventComponent = {
```
trueadm: Out of curiosity if you wrap this entire function body in the flag, does size go down?

packages/react/src/ReactCreateEventComponent.js
```diff
@@ -0,0 +1,34 @@
...
+    $$typeof: REACT_EVENT_COMPONENT_TYPE,
+    displayName: displayName,
+    props: props,
```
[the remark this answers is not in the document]
necolas: should/shouldn't?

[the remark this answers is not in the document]
trueadm: Shouldn’t sorry.

packages/react/src/ReactCreateEventComponent.js
```diff
@@ -0,0 +1,34 @@
...
+    responder: responder,
+  };
+  if (__DEV__) {
+    Object.freeze(eventComponent);
```
trueadm: Given we freeze here. We can remove the freeze operation in FiberComplete phase.

packages/shared/ReactTypes.js
```diff
@@ -136,7 +136,7 @@ export type ReactEventComponentInstance = {|
 
 export type ReactEventComponent = {|
   $$typeof: Symbol | number,
-  displayName?: string,
+  displayName: string,
```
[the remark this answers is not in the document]
trueadm: I.e.

const Press = createEventComponent({ ... });
Press.displayName = “Press”

Just an idea.

necolas:
@trueadm I don't understand why the bundle sizes for individual event components almost doubles with this change

trueadm:
@necolas it’s because it’s bundling all of React with each event module. If you go to bundles.js, you can add React as an external dependency to avoid this.

packages/react/src/ReactCreateEventComponent.js
```diff
@@ -0,0 +1,36 @@
...
+import {REACT_EVENT_COMPONENT_TYPE} from 'shared/ReactSymbols';
+
+type EventComponentConfig = {
+  displayName: string,
+  props: null | Object,
+  responder: ReactEventResponder,
+};
+
+export function createEventComponent({
+  displayName,
```
trueadm: I think it was maybe better how it was before, as we're creating less objects:

```
export function createEventComponent(responder: ReactEventResponder, displayName?: string)
```

We should also freeze the `responder` too:

```
if (__DEV__) {
  Object.freeze(responder);
}
```

necolas:
Looks like the sizebot comment got deleted. I'll open another PR to get latest report

packages/react/src/ReactCreateEventComponent.js
trueadm: Sorry, I forgot it got moved: https://github.com/facebook/react/blob/master/packages/react-dom/src/events/DOMEventResponderSystem.js#L466

trueadm: > I don't think using an object for a one-off function call is going to be a problem. We used an object for export previously. But happy to change it.

It's not a big deal, but just means less bytes and one less object created which seems better given the API for this function only has 2 arguments now

necolas:
https://github.com/facebook/react/pull/15580

PR #15578, opened 2019-05-07
Add Suspense Boundary Context (and unstable_avoidThisFallback)

sebmarkbage:
This adds a new context for Suspense boundaries. This will be used to implement longer suspense in cases it's undesirable to show a certain fallbacks state and SuspenseList.

The context is a bitmask split into two parts. A shallow part that only applies to the direct suspense boundaries below. A subtree part that applies to the entire subtree deeply. Currently there are only two flags but I expect we'll need more in the future. I already have a couple of ideas (like tagging dehydrated boundaries and optimizations for suspense list).

The interesting case that this PR models is the difference if suspense happens in a tree that is already showing its content on the screen or if the content of the current suspense boundary is invisible.

If the current suspense boundary is showing content, then switching to the fallback would hide that content. This is always undesirable. In a follow up, I will add an option to suspend the commit for longer in this case. This will kick in regardless if there is a `unstable_avoidThisFallback` boundary or not. Having one of these avoided boundaries is completely optional.

However, in the case where we cannot suspend the commit for longer (e.g. because there was no busy spinner or we're in a sync mode), the flag `unstable_avoidThisFallback` can be used as a back up boundary to avoid completely falling back to the outer most boundary. It's never a good idea to have one so I'm not sure we'll actually want to keep this API but this lets us play with it. I didn't bother adding too many tests for it.

During initial render (or if the tree was previously hidden) then a `unstable_avoidThisFallback` boundary behaves as if there was no boundary there at all. The outer one catches it.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -1425,7 +1425,7 @@ function updateSuspenseComponent(
     )
   ) {
     // This either already captured or is a new mount that was forced into its fallback
-    // state by a parernt.
+    // state by a parent.
     const attemptedState: SuspenseState | null = workInProgress.memoizedState;
     // Something in this boundary's subtree already suspended. Switch to
     // rendering the fallback children.
@@ -1438,7 +1438,7 @@ function updateSuspenseComponent(
     nextDidTimeout = true;
     workInProgress.effectTag &= ~DidCapture;
   } else {
-    // Attempting the main content the main content
+    // Attempting the main content
     if (current === null || current.memoizedState !== null) {
       // This is a new mount or this boundary is already showing a fallback state.
       // Mark this subtree context as having at least one invisible parent that could
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -12,6 +12,7 @@ import type {Fiber} from './ReactFiber';
 import type {FiberRoot} from './ReactFiberRoot';
 import type {ExpirationTime} from './ReactFiberExpirationTime';
 import type {SuspenseState} from './ReactFiberSuspenseComponent';
+import type {SuspenseContext} from './ReactFiberSuspenseContext';
 
 import checkPropTypes from 'prop-types/checkPropTypes';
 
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1394,32 +1405,62 @@ function updateSuspenseComponent(
...
       fallbackExpirationTime:
-        nextState !== null ? nextState.fallbackExpirationTime : NoWork,
+        attemptedState !== null
+          ? attemptedState.fallbackExpirationTime
+          : NoWork,
     };
     nextDidTimeout = true;
     workInProgress.effectTag &= ~DidCapture;
+  } else {
+    // Attempting the main content the main content
```
acdlite:
```suggestion
    // Attempting the main content
```

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1394,32 +1405,62 @@ function updateSuspenseComponent(
...
+
+  if (
+    (workInProgress.effectTag & DidCapture) !== NoEffect ||
+    hasSuspenseContext(
+      suspenseContext,
+      (ForceSuspenseFallback: SuspenseContext),
+    )
+  ) {
+    // This either already captured or is a new mount that was forced into its fallback
+    // state by a parernt.
```
acdlite:
```suggestion
    // state by a parent.
```

packages/react-reconciler/src/ReactFiberSuspenseContext.js
```diff
@@ -0,0 +1,83 @@
...
+
+// Subtree Flags:
+
+// InvisibleParentSuspenseContext indicates that one of our parent Suspense
+// boundaries is not currently showing visible main content.
+// Either because it is already showing a fallback or is not mounted at all.
+// We can use this to determine if it is desirable to trigger a fallback at
+// the parent. If not, then we might need to trigger undesirable boundaries
+// and/or suspend the commit to avoid hiding the parent content.
+export const InvisibleParentSuspenseContext: SubtreeSuspenseContext = 0b01;
```
acdlite: I find this name confusing but don't have a better suggestion

Issue #15275, opened 2019-03-31
Add option in shallow renderer to run effects/componentDidUpdate/componentDidMount


bdwain:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
the shallow renderer does not run componentDidUpdate, componentDidMount, or useEffect functions. (I'll call them effect functions for short)

**What is the expected behavior?**
See [this enzyme issue](https://github.com/airbnb/enzyme/issues/1938#issuecomment-476137018) for more details about where this request is coming from. But the general idea is that it is often nice to run effect functions even when shallow rendering, rather than having to use full rendering on those specific tests. 

If the shallow renderer provided an option to run the effect functions, it would allow people who test with shallow rendering to more easily test their components. Enzyme currently supports this in class components by calling componentDidUpdate/mount directly on the component instance, but this would be a much harder thing to do for hooks, since they are usually anonymous.

Enzyme used to not allow this at all, but then added an option to turn on this behavior in their shallow renderer, before finally turning it on by default and then adding an option to turn it off. It didn't seem to cause too many issues for them, so I think this approach could work well for the React shallow renderer as well. Obviously I'm only asking for an option to turn it on now, not to change the default or anything.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
All and no it was never supported AFAIK

bdwain:
Any more thoughts on this? I'd be happy to attempt a PR to add this option if there's is no opposition and it's just a matter of prioritizing it.

bdwain:
I'll start working on a PR

agrasley:
Any chance we could get some eyes from a maintainer on the pull request for this issue? It's been sitting there for a while and it would be nice to have this support for testing hooks in libraries like enzyme. Thanks!

hweeks:
This feature is vital for us to be able to move over to some of these newer paradigms in React. Is there anything we can do to help bump this ticket for visibility?

pmoeller91:
Likewise this is a big deal. Hooks support on shallow rendering is crucial.

yanlipnican:
I hope this get done soon. We would like to use hooks.

nagman:
This feature is crucial. As my whole project only uses hooks, I can test none of `useEffect` functions.

Could this issue be assigned to someone?

ohjames:
@bdwain how's your attempt at the PR going? I'm also looking into it at the moment.

ohjames:
[...]

Issue #1616, opened 2014-05-28
Add find/scry by ID to test utils


feifanzhou:
TestUtils currently has `scryRenderedDOMComponentsWithClass` and the `find` variant, but nothing like `findRenderedDOMComponentWithID`. This functionality would make testing a lot easier.

spicyj:
We're not planning to add more utils like this to TestUtils. You can always implement this locally in your project as a helper if you want to, but we're planning to move towards shallow testing (i.e., TestUtils. createRenderer) in the future.

PR #15589, opened 2019-05-08
Feedback needed: add option to enable effects and lifecycle methods in shallow renderer

bdwain:
This PR is not complete yet, and once I've confirmed I'm on the right track I can close it and reopen it when it's actually finished if that is preferred. But I wanted to make sure this general approach is going to work.

This addresses issue #15275 by adding an option in the shallow renderer to run componentDidUpdate, componentDidMount, useEffect, and useLayoutEffect. This will allow libraries like enzyme to support running lifecycle methods when shallow rendering in components that have hooks. They already support this in class components by running the lifecycle instance methods directly, but that won't work for useEffect, and it seems like the support for that should live in the shallow renderer.

### DIFF
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -173,11 +173,12 @@ function basicStateReducer<S>(state: S, action: BasicStateAction<S>): S {
 }
 
 class ReactShallowRenderer {
-  static createRenderer = function() {
-    return new ReactShallowRenderer();
+  static createRenderer = function(useLifecycleMethods) {
+    return new ReactShallowRenderer(useLifecycleMethods);
   };
 
-  constructor() {
+  constructor(useLifecycleMethods = false) {
+    this._useLifecycleMethods = useLifecycleMethods;
     this._reset();
   }
 
@@ -366,6 +367,13 @@ class ReactShallowRenderer {
       return fn;
     };
 
+    const useEffect = (fn: Function, deps: Array<mixed> | void | null): Function => {
+      this._validateCurrentlyRenderingComponent();
+      this._createWorkInProgressHook();
+
+      //fill in impl
+    };
+
     return {
       readContext,
       useCallback: (identity: any),
@@ -374,9 +382,9 @@ class ReactShallowRenderer {
         return readContext(context);
       },
       useDebugValue: noOp,
-      useEffect: noOp,
+      useEffect: this._useLifecycleMethods ? useEffect : noOp,
       useImperativeHandle: noOp,
-      useLayoutEffect: noOp,
+      useLayoutEffect: this._useLifecycleMethods ? useEffect : noOp,
       useMemo,
       useReducer,
       useRef,
@@ -674,8 +682,13 @@ class ReactShallowRenderer {
     }
 
     this._rendered = this._instance.render();
-    // Intentionally do not call componentDidMount()
+    // Force user to opt in to calling componentDidMount()
     // because DOM refs are not available.
+    if(this._useLifecycleMethods) {
+      if (typeof this._instance.componentDidMount === 'function') {
+        this._instance.componentDidMount();
+      }
+    }
   }
 
   _updateClassComponent(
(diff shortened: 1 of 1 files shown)


packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -366,6 +367,13 @@ class ReactShallowRenderer {
       return fn;
     };
 
+    const useEffect = (fn: Function, deps: Array<mixed> | void | null): Function => {
+      this._validateCurrentlyRenderingComponent();
+      this._createWorkInProgressHook();
+
+      //fill in impl
```
bdwain: This is the part I'm trying to still figure out how to do. I think I have a general idea of how things work, but before I went any further I wanted to confirm that this makes sense.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -173,11 +173,12 @@ function basicStateReducer<S>(state: S, action: BasicStateAction<S>): S {
...
 
 class ReactShallowRenderer {
-  static createRenderer = function() {
-    return new ReactShallowRenderer();
+  static createRenderer = function(useLifecycleMethods) {
+    return new ReactShallowRenderer(useLifecycleMethods);
   };
 
-  constructor() {
+  constructor(useLifecycleMethods = false) {
```
bdwain: The way this option is passed can be changed. I just did this for now because it was easy.

louisscruz:
Are there any eyes on this? The approach here seems reasonable, and Enzyme seems to be blocked on this (or something like it) getting in.

bdwain:
@gaearon we discussed this in [this enzyme issue](https://github.com/airbnb/enzyme/issues/1938#issuecomment-476137018). can you point me to the right person to give feedback on this please? I think this would allow a lot of people to start using hooks that are currently blocked from it because they use shallow rendering in tests.

audiolion:
Wouldn't this always just be a poor man's version of enzyme's `mount` that could potentially have bugs where behavior is different? I know other libraries that use this approach create a dummy component and insert the hook(s) into it and run the full lifecycle to test them.

bdwain:
@audiolion isn't that true for all shallow rendering? I'm just proposing calling the functions in useEffect instead of doing nothing. We already simulate the behavior of other hooks, so useEffect seems appropriate to do too. And I'm suggesting exposing the behavior behind an option in the shallow renderer to avoid a breaking change.

Also, just to be clear, this is not to enable testing of custom hooks (though I suppose it could help with that too). It's to enable testing of components that call useEffect (or some other hook that calls useEffect).

Joonpark13:
Just want to chime in and say it would be great to be able to test components that have `useEffect` with `shallow` in enzyme. Would love to get someone from the react team's eyes on this!

bdwain:
Is there a way to get this PR looked at? It's been open for a while now?

agrasley:
Would love to see a maintainer put some eyes on this. Would make the hooks testing story with libraries like enzyme much better.

hweeks:
Another bump on getting some maintainers to take a look at this change.

Vetrano89:
Hey has anyone looked at this yet? :godmode: 🙏 📿

Alphy11:
@acdlite Sorry to ping you on it, but it does not seem to have any sort of response, seems to be a relatively simple change, and is blocking pretty much anyone using shallow render from using hooks.

Joonpark13:
Hate to keep bumping this but I think it's preventing a lot of people from using `useEffect` and `shallow`. Would it be possible to at least get some kind of indication that this will be looked at in the future? Or if this is not something that the team is willing to consider, then a comment saying so, so we can find a workaround?

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -674,8 +682,13 @@ class ReactShallowRenderer {
     }
 
     this._rendered = this._instance.render();
-    // Intentionally do not call componentDidMount()
+    // Force user to opt in to calling componentDidMount()
     // because DOM refs are not available.
+    if(this._useLifecycleMethods) {
```
intergalactic-overlords: spacing

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -760,9 +773,21 @@ class ReactShallowRenderer {
 
     if (shouldUpdate) {
       this._rendered = this._instance.render();
+
+      // Force user to opt in to calling componentDidUpdate() and 
+      // getSnapshotBeforeUpdate because DOM refs are not available.
+      if(this._useLifecycleMethods) {
```
intergalactic-overlords: spacing

packages/react-test-renderer/src/ReactShallowRenderer.js
bdwain: I'm still trying to get feedback on the implementation plan. I'll clean up spacing and things like that at the end

ohjames:
I wrote a complete version of this work here: https://github.com/facebook/react/pull/1616

It might be a good idea to close this PR so we can concentrate on the complete fix and avoid spreading the gaze of the maintainers too thinly?

Joonpark13:
> I wrote a complete version of this work here: #1616

Thanks for this! You meant to say #16168 though, I think?

bdwain:
closing this in favor of #16188

PR #15591, opened 2019-05-08
`act()` - s / flushPassiveEffects / Scheduler.unstable_flushWithoutYielding

threepointone:
a first crack at flushing the scheduler manually from inside `act()`. 
- uses `unstable_flushWithoutYielding()` directly instead of `flushAll()` 
- for tests to pass, I mostly replaced `toFlushAndYield(...)` with `toHaveYielded()`. 
- for a few tests that tested the state of the tree before flushing effects (but still after updates), I replaced `act()` with `batchedUpdates()`.

Open questions - 
- uh, is this right? - edit: yes. 
- `unstable_flushWithoutYielding` is an odd name, especially when it's followed by `toHaveYielded`. Or am I doing something wrong here? Or should one of these named differently? edit: yup, the names are confusing. will change in a followup PR

todo
- [x] add a warning when scheduler isn't mocked (but fallback to a version that works in sync mode)
- [x] make a umd build of scheduler/unstable_mock 
- [x] more tests for concurrent mode
- [ ] make a docs page/link explaining more


packages/shared/forks/Scheduler.umd.js
```diff
@@ -28,6 +28,7 @@ const {
   unstable_LowPriority,
   unstable_IdlePriority,
   unstable_forceFrameRate,
+  unstable_flushWithoutYielding,
```
threepointone: this is silly, because we don't ship this on the actual scheduler, only the mock. (also, means our tests don't run in browsers anymore). thinking of workarounds; will probably have to make a umd build of the mock version

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -503,4 +503,23 @@ describe('ReactTestUtils.act()', () => {
       });
     }
   });
+  describe('concurrent mode', () => {
```
threepointone: I will of course flesh this out.

packages/react-test-renderer/src/ReactTestRendererAct.js
```diff
@@ -27,9 +25,9 @@ let actingUpdatesScopeDepth = 0;
 
 function flushEffectsAndMicroTasks(onDone: (err: ?Error) => void) {
   try {
-    flushPassiveEffects();
+    Scheduler.unstable_flushWithoutYielding();
     enqueueTask(() => {
-      if (flushPassiveEffects()) {
+      if (hasPendingEffects()) {
```
acdlite: Maybe this should check if any work is scheduled, not just effects? `Scheduler.hasPendingWork()`

acdlite:
> unstable_flushWithoutYielding is an odd name, especially when it's followed by toHaveYielded. Or am I doing something wrong here? Or should one of these named differently?

I agree the naming sucks. We'll rename them eventually; suggestions welcome.

Need to throw a helpful error when Scheduler is not the mock build.

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
```diff
@@ -40,16 +41,32 @@ const {ReactShouldWarnActingUpdates} = ReactSharedInternals;
...
+          'For more info, visit https://fb.me/react-mock-scheduler',
+      );
+      hasWarnedAboutMissingMockScheduler = true;
+    }
 
-function flushEffectsAndMicroTasks(onDone: (err: ?Error) => void) {
+    while (flushPassiveEffects()) {}
+  };
+
+function flushWorkAndMicroTasks(onDone: (err: ?Error) => void) {
```
bvaughn: How would this actually flush micro tasks (in an env like Jest)?

threepointone: that's what enqueueTask is for https://github.com/facebook/react/blob/master/packages/shared/enqueueTask.js, it waits for a so-called 'macro' task before continuing. Any specific concern?

threepointone:
@bvaughn my current thinking is to have a `advanceTimersByTime` helper, that internally would actually call smaller timeouts, flushing between every run. I'm not committing to any specifics on it just yet, but am spending today/tomorrow writing out some tests including that scenario.

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
bvaughn: My concern is that if nothing tells Jest to advance it's timers, that macro task will never fire, and the test will just time out.

bvaughn:
I guess I'll wait and see what you have in mind then. I assume `advanceTimersByTime` will only work with the mock scheduler's `now` function though, and not with e.g. `performance.now`/`Date.now`? I guess that's okay, but it will require user code to also avoid using those functions if tests need predictable timing.

SimenB:
>I assume `advanceTimersByTime` will only work with the mock scheduler's `now` function though, and not with e.g. `performance.now`/`Date.now`?

FWIW, I'm working on and off (mostly off...) on landing [Lolex](https://yarnpkg.com/en/package/lolex) in Jest, which will add make sure `Date` (`new Date()` & `Date.now()`) and `performance.now()` are also mocked and controlled by `advanceTimersByTime`: https://github.com/facebook/jest/pull/7776. I'm hopeful we'll land it for Jest 25.

Not that this affects you at all _now_, but possibly something to keep in mind. If you do want real timings while faking time you'd have to get a reference to the original `performance.now` before installing fake timers. Alternatively, it might make sense for Jest to expose some way of timing things that's unaffected by fake timers being installed? Not sure

threepointone:
I updated the PR to run the TestUtilsAct-test tests in both heritage/concurrent mode. I recommend turning off whitespace changes when reading.

threepointone:
opening this up for review, but still unsure how to get the scheduler to unmock. will keep poking at it.

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -31,476 +55,502 @@ describe('ReactTestUtils.act()', () => {
...
     });
+  });
 
-    it('should flush effects recursively', () => {
-      function App() {
-        let [ctr, setCtr] = React.useState(0);
-        React.useEffect(() => {
-          if (ctr < 5) {
-            setCtr(x => x + 1);
+  [false, true].forEach(concurrent => {
```
acdlite: See above comment

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -13,6 +13,30 @@ let ReactTestUtils;
 let SchedulerTracing;
 let act;
 let container;
+let isConcurrentMode;
+let concurrentRoot = null;
+let ReactFeatureFlags;
+
+function render(el, dom) {
+  if (isConcurrentMode) {
+    concurrentRoot = ReactDOM.unstable_createRoot(dom);
```
acdlite: I really dislike test abstractions like this. Especially when combined with mutable module state like `isConcurrentMode`. It makes writing tests easier, but it makes debugging failing tests harder.

If you do want to share code, I would pass `render` and `unmount` to a helper function that runs the tests cases, like we do in `ReactFiberNewContext-test`. Something like:

```js
runActTests(renderConcurrent, unmountConcurrent, 'concurrent');
runActTests(renderLegacySync, unmountLegacySync, 'legacy sync');
```

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -13,6 +13,30 @@ let ReactTestUtils;
...
+    concurrentRoot.render(el);
+    return;
+  } else {
+    return ReactDOM.render(el, dom);
+  }
+}
+
+function unmount(el) {
```
acdlite: Same here

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -31,476 +55,502 @@ describe('ReactTestUtils.act()', () => {
...
-            return () => clearTimeout(timeout);
-          }, []);
-          return toggle;
-        }
+          let calledLog = [];
+          act(() => {
+            render(
+              <App
+                callback={() => {
+                  calledLog.push(calledLog.length);
```
acdlite: Nit: Why this prop indirection? Why does it log the length? Might seem like a silly nit to pick but little stuff like this makes it hard for me to infer which parts of the test are important and which are incidental.

threepointone: now that we have a mock scheduler, I should just use .yieldValue and co

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -31,476 +55,502 @@ describe('ReactTestUtils.act()', () => {
...
-          />,
-          document.createElement('div'),
-        );
-      });
-
-      expect(calledLog).toEqual([0]);
+  describe('mocked scheduler', () => {
+    beforeEach(() => {
+      ReactFeatureFlags.warnAboutMissingMockScheduler = true;
+      // todo - this doesn't actually work
```
threepointone: @acdlite I'm kinda struggling here - nothing I'm trying is able to unmock the mocked scheduler. does anything here (in the commented code) look out of place to you? plsthx!

acdlite: After unmocking it, you need to re-require the modules. The variables still point to the old versions.

threepointone:
alright, I updated the PR 
- pass the mock scheduler warning test,
- rewrite some tests to use Scheduler.yieldValue
- structure concurrent/legacy suites neatly
- I also merged from master, and rewrote some failing tests. of note, I had to swap a couple of act calls out with batchedUpdates for a couple of tests that tested in between state.

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
```diff
@@ -347,14 +348,17 @@ describe('ReactHooks', () => {
       });
       return <Child text={text} />;
     }
+    let root;
+    act(() => {
+      root = ReactTestRenderer.create(null, {unstable_isConcurrent: true});
```
acdlite: Nit: If you move the `ReactTestRenderer.create` call outside `act`, does it warn? Maybe we can special case `null`? (I don't like how test renderer's `create` also updates at the same time, and I'm accustomed to passing null as a workaround.)

threepointone: It does not warn when you pass null, we only warn on update calls. It will warn if you try to render something with an effect ofc  (once https://github.com/facebook/react/pull/15631 lands) To clarify, this is the behaviour you want, yeah? I'll move it out in this test.

threepointone:
alright, I'm going to land this. there are 2 major todos -
- docs, which I'm working on 
- discussed with @gaearon, and it was apparent that the missing mock scheduler warning only makes sense in concurrent mode (and the messaging is a but off). maybe this is something we should sovle with the testing build. in any case, am landing the messaging as is, but with the feature flag turned off for oss. will reach out to discuss next steps separately.

packages/react-test-renderer/src/ReactTestRendererAct.js
```diff
@@ -14,23 +14,42 @@ import {
...
+    if (warnAboutMissingMockScheduler === true) {
+      if (hasWarnedAboutMissingMockScheduler === false) {
+        warningWithoutStack(
+          null,
+          'Starting from React v17, the "scheduler" module will need to be mocked ' +
+            'to guarantee consistent behaviour across tests and browsers. To fix this, add the following ' +
+            "to the top of your tests, or in your framework's global config file -\n\n" +
+            'As an example, for jest - \n' +
+            "jest.mock('scheduler', () => require.requireActual('scheduler/unstable_mock'));\n\n" +
+            'For more info, visit https://fb.me/react-mock-scheduler',
```
kentcdodds: FYI, this short URL was never created 😅 I get redirected to facebook.com

threepointone: Good catch, we should land https://github.com/reactjs/reactjs.org/pull/2171 (preview: https://deploy-preview-2171--reactjs.netlify.com/docs/mock-scheduler.html)

PR #15586, opened 2019-05-08
Experimental Event API: Press event properties

trueadm:
This PR adds some `Press` responder event properties that are missing in relation to:

- clientX
- clientY
- screenX
- screenY
- pageX
- pageY
- timeStamp

We attempt to get these from the `nativeEvent` object, except in the case of delayed events, where we calculate the values based on nativeEvent's `timeStamp` value.

This PR depends on https://github.com/facebook/react/pull/15587.


necolas:
Do you mind if we split this PR into separate PR's? The two changes aren't really related.

PR #15587, opened 2019-05-08
Ensure touch events are properly handled for pageX and pageY

trueadm:
Whilst working with an iOS device, I noticed that we were incorrectly trying to access the co-ords for `pageX` and `pageY` on the `nativeEvent` object. They don't exist in this place, rather this change makes it so we access them on the `changedTouches` array (if it exists). We only take them from the first event, we need to add support multiple touches in the future.

Ref #15257

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -406,12 +406,12 @@ function isTouchEvent(nativeEvent: Event): boolean {
 }
 
 function getTouchFromPressEvent(nativeEvent: TouchEvent): Touch {
-  const {changedTouches, touches} = event.nativeEvent;
+  const {changedTouches, touches} = nativeEvent;
   return changedTouches.length > 0
     ? changedTouches[0]
     : touches.length > 0
       ? touches[0]
-      : event.nativeEvent;
+      : (nativeEvent: any);
 }
 
 function getEventPageCoords(
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -190,14 +190,13 @@ function dispatchLongPressChangeEvent(
 
 function activate(event, context, props, state) {
   const nativeEvent: any = event.nativeEvent;
-  const pageX = nativeEvent.pageX;
-  const pageY = nativeEvent.pageY;
+  const {x, y} = getEventPageCoords(nativeEvent);
   const wasActivePressed = state.isActivePressed;
   state.isActivePressed = true;
-  if (pageX != null && pageY != null) {
+  if (x !== null && y !== null) {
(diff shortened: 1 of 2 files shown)


packages/react-events/src/Press.js
```diff
@@ -402,12 +401,36 @@ function calculateResponderRegion(target: Element, props: PressProps) {
...
 
+function isTouchEvent(nativeEvent: Event): boolean {
+  return Array.isArray((nativeEvent: any).changedTouches);
+}
+
+function getEventPageCoords(
+  nativeEvent: Event,
+): {x: null | number, y: null | number} {
+  let eventObject = (nativeEvent: any);
+  if (isTouchEvent(eventObject)) {
```
necolas: The equivalent in RN is:

```
const getTouchFromPressEvent = (event: PressEvent) => {	
  const {changedTouches, touches} = event.nativeEvent;	
  return changedTouches.length > 0	
    ? changedTouches[0]	
    : touches.length > 0	
    ? touches[0]	
    : event.nativeEvent;	
};
```

packages/react-events/src/Press.js
```diff
@@ -402,12 +401,36 @@ function calculateResponderRegion(target: Element, props: PressProps) {
...
+    const touchEvents = eventObject.changedTouches;
+    if (touchEvents.length === 0) {
+      throw new Error('TouchEvent does not have any changed touches');
+    }
+    // TODO: we need to support multi-touch in the future, this is a temp
```
necolas: How does multi-touch related to Press?

packages/react-events/src/__tests__/Press-test.internal.js
```diff
@@ -1309,30 +1291,404 @@ describe('Event responder: Press', () => {
...
+        'onPressStart',
+        'onPressChange',
+        'onPressEnd',
+        'onPressChange',
+        'onPress',
+      ]);
+    });
+  });
+
+  describe('press with movement (touch events fallback)', () => {
```
necolas: It's a little difficult to follow what happened in this part of the diff. Is this duplicating the entire pointer events test block? If so, I don't think we need to do that (and we don't for mouse events)

trueadm: It doesn't duplicate it all, a lot of it, but I changed a few tests too and remove a few there were mouse related.

packages/react-events/src/Press.js
trueadm: As we discussed in our meeting, we need to consider the user pressing multiple Press event responders at the same time. I'll address this TODO tomorrow in a follow up.

PR #15593, opened 2019-05-09
Add withSuspenseConfig API

sebmarkbage:
This adds `unstable_withSuspenseConfig(callback, config)` to the `react` package. This is a primitive that can be used to implement "busy spinners". __I recommend reviewing the code commit by commit.__

This sets the config as a global state available to all React renderers during the scope of the synchronously executed callback (first arg). If you don't pass a config, it disables suspense (for longer than the default JND).

If any updates are scheduled (render, setState, dispatch) during this scope, then they get this "suspense config" associated with the update.

If during a render pass, we process one of those updates, we mark this render pass as having a "suspense config".

If during a render pass, we are forced to turn already visible content back into a fallback, we mark this render as suspended with a delay if possible. (This doesn't not happen during initial render of new content.)

If we have both marked this render as having a suspense config and it was suspended with delay, then we extend the suspended time beyond JND all the way to the timeout defined in the suspense config.

The suspense config has these options:

```
{
  timeoutMs: number,
  loadingDelayMs?: number,
  minLoadingDurationMs?: number,
}
```

The last two are used when we have actually completed a render fairly quickly. If we have exceeded `loadingDelayMs` but have not yet exceeded `loadingDelayMs + minLoadingDurationMs` then we suspend the commit until that time even though we could commit early. This can be used to implement busy spinners that delay when they first show but want to stay visible for a minimum time if they do show, to avoid flickering.

Tasks:
- [x] Implementation
- [ ] Tests

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -220,6 +220,7 @@ ReactBatch.prototype.render = function(children: ReactNodeList) {
     internalRoot,
     null,
     expirationTime,
+    null,
     work._onCommit,
   );
   return work;
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -971,6 +971,27 @@ function renderRoot(
     }
     case RootCompleted: {
       // The work completed. Ready to commit.
+      if (
+        !isSync &&
+        workInProgressRootLatestProcessedExpirationTime !== Sync &&
+        workInProgressRootCanSuspendUsingConfig !== null
+      ) {
+        // If we have exceeded the minimum loading delay, which probably
+        // means we have shown a spinner already, we might have to suspend
+        // a bit longer to ensure that the spinner is shown for enough time.
+        const msUntilTimeout = computeMsUntilSuspenseLoadingDelay(
+          workInProgressRootLatestProcessedExpirationTime,
+          expirationTime,
+          workInProgressRootCanSuspendUsingConfig,
+        );
+        if (msUntilTimeout > 10) {
+          root.timeoutHandle = scheduleTimeout(
+            commitRoot.bind(null, root),
+            msUntilTimeout,
+          );
+          return null;
+        }
+      }
       return commitRoot.bind(null, root);
     }
     default: {
@@ -1952,6 +1973,39 @@ function jnd(timeElapsed: number) {
               : ceil(timeElapsed / 1960) * 1960;
 }
 
+function computeMsUntilSuspenseLoadingDelay(
+  mostRecentEventTime: ExpirationTime,
+  committedExpirationTime: ExpirationTime,
+  suspenseConfig: SuspenseConfig,
+) {
+  if (disableYielding) {
+    // Timeout immediately when yielding is disabled.
+    return 0;
+  }
+
+  const minLoadingDurationMs = (suspenseConfig.minLoadingDurationMs: any) | 0;
+  if (minLoadingDurationMs <= 0) {
+    return 0;
+  }
+  const loadingDelayMs = (suspenseConfig.loadingDelayMs: any) | 0;
+
+  // Compute the time until this render pass would expire.
+  const currentTimeMs: number = now();
+  const eventTimeMs: number = inferTimeFromExpirationTime(
+    mostRecentEventTime,
+    suspenseConfig,
+  );
+  const timeElapsed = currentTimeMs - eventTimeMs;
+  if (timeElapsed <= loadingDelayMs) {
+    // If we haven't yet waited longer than the initial delay, we don't
+    // have to wait any additional time.
+    return 0;
+  }
+  const msUntilTimeout = timeElapsed - loadingDelayMs - minLoadingDurationMs;
+  // This is the value that is passed to `setTimeout`.
+  return msUntilTimeout;
+}
+
 function computeMsUntilTimeout(
   mostRecentEventTime: ExpirationTime,
   committedExpirationTime: ExpirationTime,
@@ -1976,7 +2030,7 @@ function computeMsUntilTimeout(
     mostRecentEventTime,
     suspenseConfig,
   );
-  const timeElapsed = currentTimeMs - eventTimeMs;
+  let timeElapsed = currentTimeMs - eventTimeMs;
   if (timeElapsed < 0) {
     // We get this wrong some time since we estimate the time.
     timeElapsed = 0;
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -210,7 +210,8 @@ let workInProgressRootExitStatus: RootExitStatus = RootIncomplete;
 // This is conceptually a time stamp but expressed in terms of an ExpirationTime
 // because we deal mostly with expiration times in the hot path, so this avoids
 // the conversion happening in the hot path.
-let workInProgressRootMostRecentEventTime: ExpirationTime = Sync;
+let workInProgressRootLatestProcessedExpirationTime: ExpirationTime = Sync;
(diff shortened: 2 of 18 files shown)


gaearon:
Why does this example show a "user blocking update suspended" if I click more than once?
https://gist.github.com/gaearon/a9db55f4341ebe31c9f59c9bfab63333

sebmarkbage:
@gaearon You need scheduler.next too.

gaearon:
OK this does work better. https://gist.github.com/gaearon/15f4b0f8265a1988428f7f60f347d80d

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -220,6 +220,7 @@ ReactBatch.prototype.render = function(children: ReactNodeList) {
     internalRoot,
     null,
     expirationTime,
+    null,
```
sebmarkbage: This means that batches never suspend, but I'm not sure it makes sense for them to suspend anyway since they control the commit anyway. I'm not sure when the promise should resolve in this new multi-commit world.

packages/react/src/ReactBatchConfig.js
```diff
@@ -0,0 +1,29 @@
...
+const DefaultSuspenseConfig: SuspenseConfig = {
+  timeoutMs: 5000,
+  loadingDelayMs: 0,
+  minLoadingDurationMs: 0,
+};
+
+// Within the scope of the callback, mark all updates as being allowed to suspend.
+export function suspendIfNeeded(scope: () => void, config?: SuspenseConfig) {
+  const previousConfig = ReactCurrentBatchConfig.suspense;
+  ReactCurrentBatchConfig.suspense = config || DefaultSuspenseConfig;
```
acdlite: Nit: Explicit comparison?

packages/react/src/ReactBatchConfig.js
```diff
@@ -0,0 +1,29 @@
...
+  try {
+    scope();
```
acdlite: Should this return? For consistency with `flushSync`, `next`, `runWithPriority`, et al.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -946,9 +951,17 @@ function renderRoot(
   }
 }
 
-export function markRenderEventTime(expirationTime: ExpirationTime): void {
-  if (expirationTime < workInProgressRootMostRecentEventTime) {
-    workInProgressRootMostRecentEventTime = expirationTime;
+export function markRenderEventTimeAndConfig(
```
acdlite: Naming suggestion: markBatchConfig

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -988,11 +997,19 @@ export function renderDidError() {
...
+function inferTimeFromExpirationTime(
+  expirationTime: ExpirationTime,
+  suspenseConfig: null | SuspenseConfig,
+): number {
   // We don't know exactly when the update was scheduled, but we can infer an
   // approximate start time from the expiration time.
   const earliestExpirationTimeMs = expirationTimeToMs(expirationTime);
-  return earliestExpirationTimeMs - LOW_PRIORITY_EXPIRATION;
+  return (
+    earliestExpirationTimeMs -
```
acdlite: Let's store the start time on the batch config so we don't have to infer it? Can do in a follow up.

acdlite: However the start time computation is pretty broken right now and it might be nice to get it soon so we can rule it out as a source of possible wonkiness when we get bug reports.

packages/react/src/ReactBatchConfig.js
sebmarkbage: What do we do elsewhere? I don't want any bad types like undefined or false to leak into the implementation since it might not immediately throw only deopt, and I don't want to add a bunch of prod runtime code since this should be a fairly very light operation.

packages/react/src/ReactBatchConfig.js
sebmarkbage: I guess. What is that ever used for? I could imagine things like async functions but that seems like very confusing bad practice to use an async function here.

packages/react-reconciler/src/ReactFiberScheduler.js
sebmarkbage: I don't want to have to create a new object for the batch too.

acdlite: Can it not be the same object as `suspenseConfig`?

packages/react/src/ReactBatchConfig.js
acdlite: Elsewhere we check for null and undefined explicitly, like

```js
ReactCurrentBatchConfig.suspense = config !== undefined && config !== null ? config : DefaultSuspenseConfig;
```

acdlite: I guess I'm wondering why you've chosen a different pattern here, and whether we should go change the other sites. Like `root.render`: 

https://github.com/facebook/react/blob/3d8b836e22981063d0396e04eefcb10fdec600ee/packages/react-dom/src/client/ReactDOM.js#L372-L378

packages/react-reconciler/src/ReactFiberScheduler.js
sebmarkbage: `suspenseConfig` is a user space object so mutating it gets fragile. I also expect it to mostly be a hoisted shared object. E.g. the default one right now but also the defaults in hooks would just use a hoisted config.

We could have another field for that but that requires me to track another thing for the render path. It's probably the right strategy eventually but I think we need to think about it some more.

packages/react/src/ReactBatchConfig.js
sebmarkbage: Now I wonder if `null` should be resetting it to not suspend? I guess you can also pass in timeout: 0 but then that opts you out of JND.

acdlite:
Is this ready for review? Don't see any tests

gaearon:
How do you decide what to put on React (this) vs Scheduler (next)? I understand React gives a stronger sintleton guarantee but as a user the distinction seems blurred to me. Especially because a bunch of similar modifiers live on the renderer (flushSync).

sebmarkbage:
@gaearon So for React vs Renderer. Things that have the same type signature regardless of renderer is a good candidate. findDOMNode and createPortal are not.

There are things that you expect to call only in an environment interop situation that can remain on the environment. E.g. flushSync. So it doesn’t really need to be agnostic. It’ll only be called from code that is written specifically for an environment. Additionally it might even know which particular root might be affected.

There are things that have the same type signature and is expected to be called in the source of isomorphic Components a lot. That’s why this api goes on React. Another good candidate would’ve been batchedUpdates if we weren’t planning on moving away from it anyway.

Now as for what goes on scheduler vs React. Scheduler needs to be lean and not have more concepts than everyone can agree on. For example it doesn’t know anything about batching or suspense. Those are React concepts and we can’t just add them to scheduler. Scheduler is things that you’re expected to call even if you’re not even using React or if you have a library that isn’t coupled to React.

Eg Redux should call things in scheduler but not in React. Redux-React should call things in React or scheduler but not the renderer.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -952,6 +974,27 @@ function renderRoot(
...
+      ) {
+        // If we have exceeded the minimum loading delay, which probably
+        // means we have shown a spinner already, we might have to suspend
+        // a bit longer to ensure that the spinner is shown for enough time.
+        const msUntilTimeout = computeMsUntilSuspenseLoadingDelay(
+          workInProgressRootLatestProcessedExpirationTime,
+          expirationTime,
+          workInProgressRootCanSuspendUsingConfig,
+        );
+        if (msUntilTimeout > 10) {
```
NE-SmallTown: what does 10 means?

PR #15599, opened 2019-05-09
Event API: ensure we pop context for event system fibers

trueadm:
This PR fixes an issue I stumbled across when running this internally. I'm not sure how to repro this in a test, as I don't know what conditions are needed exactly with the test renderer to invoke the `unwindInterruptedWork` function. Will follow up with someone to figure one out.


trueadm:
I will add some tests in a follow up. Landing to unblock internal issues.

gaearon:
To find what breaks, you can comment out _other_ lines in this function, see which existing tests break, and write similar tests. Unwinding is used for Suspense and/or error handling afaik.

PR #15604, opened 2019-05-10
[react-native] Use path-based imports instead of Haste for the RN renderer

ide:
To move React Native to standard path-based imports instead of Haste, the RN renderer that is generated from the code in this repo needs to use path-based imports as well since the generated code is vendored by RN. This commit makes it so the interface between the generated renderers and RN does not rely on Haste and instead uses a private interface explicitly defined by RN. This inverts control of the abstraction so that RN decides the internals to export rather than React deciding what to import.

On RN's side, a new module named `react-native/Libraries/ReactPrivate/ReactNativePrivateInterface` explicitly exports the modules used by the renderers in this repo. (There is also a private module for InitializeCore so that we can import it just for the side effects.) On React's side, the various renderer modules access RN internals through the explicit private interface.

The Rollup configuration becomes slimmer since the only external package is now `react-native`, and the individual modules are instead listed out in `ReactNativePrivateInterface`.

Task description: https://github.com/facebook/react-native/issues/24770
Sister RN PR (needs to land before this one): https://github.com/facebook/react-native/pull/24782

Test Plan: Ran unit tests and Flow in this repo. Generated the renderers and manually copied them over to the RN repo. Ran the RN tests and launched the RNTester app.


ide:
@cpojer Since the RN side of this work in https://github.com/facebook/react-native/pull/24807 has landed (thank you!), the renderers generated by this commit should be safe to sync out to RN whenever the next sync happens. Because we want to migrate away from Haste in RN 0.61, this means this commit needs to land and a React renderer sync needs to happen sometime before the next release and ideally with enough time to land & smoke test the follow-up commits that update the Jest/Flow/etc configs.

ide:
Added a commit to replace all the FabricUIManager imports with global `nativeFabricUIManager` references.

sebmarkbage:
Who is doing the sync of these changes to RN?

cpojer:
If there is a way to sync only the changes from this PR then I’m happy to do it next week. Otherwise I was gonna wait until the next time we need to do a sync.

ide:
@cpojer I think it wouldn't be too hard to check out the copy of React that was used for the last sync, cherry-pick these changes, and then copy them over to the RN repo. Here is a PR: https://github.com/facebook/react-native/pull/25010.

PR #15611, opened 2019-05-10
Event API: add key modifiers to Press events

trueadm:
This PR adds keyboard modifier booleans (`altKey`, `altKey`, `metaKey` and `shiftKey`) to Press events, so user-land gets access to this information. Furthermore, I also added `x` and `y` to Press events too (alias for `clientX` and `clientY`).

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -91,6 +91,12 @@ type PressEvent = {|
   pageY: null | number,
   screenX: null | number,
   screenY: null | number,
+  x: null | number,
+  y: null | number,
+  altKey: boolean,
+  ctrlKey: boolean,
+  metaKey: boolean,
+  shiftKey: boolean,
 |};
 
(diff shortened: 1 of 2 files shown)


packages/react-events/src/Press.js
```diff
@@ -148,9 +154,14 @@ function createPressEvent(
...
   let screenX = null;
   let screenY = null;
+  let altKey = false;
+  let ctrlKey = false;
+  let metaKey = false;
+  let shiftKey = false;
 
   if (event) {
     const nativeEvent = (event.nativeEvent: any);
+    ({altKey, ctrlKey, metaKey, shiftKey} = nativeEvent);
```
threepointone: why the extra `()` here? prettier?

trueadm: You need it, otherwise it doesn't assign to those bindings. Prettier will add them in anyway.

PR #15619, opened 2019-05-11
[Fresh] Set up initial scaffolding

gaearon:
This sets up some scaffolding for Fresh.

In particular:

* A DevTools Hook handler that lets you schedule hot updates by providing a "type to family" map.
* A test suite that verifies we can make a hot update while preserving state.
* We only add overhead (and fork the behavior) if the "type to family" map was injected.

The latest version of a component is resolved through a pointer in the family map. More detailed design I'm aiming for: https://gist.github.com/gaearon/a4d9fb3e6ea487a9296a8d2d9a6e3bf2.

There's a bunch of other TODOs in the code. I've addressed some of them in the subsequent pushed commits. However even outstanding TODOs don't necessarily need to block landing the initial scaffold.

You might want to read individual commits.

### In this PR

- [x] Test fixture setup
- [x] Render latest resolved type
- [x] Consider family when reconciling
- [x] Support forwardRef and memo
- [x] Ignore bailouts for edited components
- [x] Add a way to force remount of arbitrary fiber
- [x] Skip bailouts for any parent fibers when forcing child remount
- [x] Reset effects, callbacks, and memo for edited components
- [x] Use a first-class mechanism for remounting a fiber
- [x] Remount currently failed boundaries during hot reload

### Follow-ups

- [ ] Deal with classes
- [ ] Deal with lazy components, Suspense, and context consumers (?)
- [ ] Handle edge case type switches (e.g. simple memo -> non-simple memo)
- [ ] Ensure DevTools inspection sees latest versions
- [ ] Report which host nodes have been affected for visual feedback
- [ ] Babel plugin to generate registration calls which are manual in tests

### DIFF
--- a/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
@@ -83,8 +83,9 @@ describe('ReactFresh', () => {
     // TODO: invalidation based on signatures.
   }
 
-  it('can preserve state', () => {
+  it('can preserve state for compatible types', () => {
     if (__DEV__) {
+      let HelloV1;
       render(() => {
         function Hello() {
           const [val, setVal] = React.useState(0);
@@ -94,6 +95,7 @@ describe('ReactFresh', () => {
             </p>
           );
         }
+        HelloV1 = Hello;
         __register__(Hello, 'Hello');
         return Hello;
       });
@@ -108,6 +110,7 @@ describe('ReactFresh', () => {
       expect(el.textContent).toBe('1');
 
       // Perform a hot update.
+      let HelloV2;
       patch(() => {
         function Hello() {
           const [val, setVal] = React.useState(0);
@@ -117,6 +120,7 @@ describe('ReactFresh', () => {
             </p>
           );
         }
+        HelloV2 = Hello;
         __register__(Hello, 'Hello');
         return Hello;
       });
@@ -134,7 +138,48 @@ describe('ReactFresh', () => {
       expect(el.textContent).toBe('2');
       expect(el.style.color).toBe('red');
 
-      // TODO: test reconciliation with old type.
+      // Perform top-down renders with both fresh and stale types.
+      // Neither should change the state or color.
+      // They should always resolve to the latest version.
+      render(() => {
+        return HelloV1;
+      });
+      render(() => {
+        return HelloV2;
+      });
+      render(() => {
+        return HelloV1;
+      });
+      expect(container.firstChild).toBe(el);
+      expect(el.textContent).toBe('2');
+      expect(el.style.color).toBe('red');
+
+      // Bump the state again.
+      act(() => {
+        el.dispatchEvent(new MouseEvent('click', {bubbles: true}));
+      });
+      expect(container.firstChild).toBe(el);
+      expect(el.textContent).toBe('3');
+      expect(el.style.color).toBe('red');
+
+      // Finally, a render with incompatible type should reset it.
+      render(() => {
+        function Hello() {
+          const [val, setVal] = React.useState(0);
+          return (
+            <p style={{color: 'blue'}} onClick={() => setVal(val + 1)}>
+              {val}
+            </p>
+          );
+        }
+        // No register call.
+        // This is considered a new type.
+        return Hello;
+      });
+      expect(container.firstChild).not.toBe(el);
+      const newEl = container.firstChild;
+      expect(newEl.textContent).toBe('0');
+      expect(newEl.style.color).toBe('blue');
     }
   });
 });
--- a/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
@@ -1351,7 +1351,7 @@ describe('ReactFresh', () => {
     if (__DEV__) {
       let useEffectWithEmptyArrayCalls = 0;
 
-      let HelloV1 = render(() => {
+      render(() => {
         function Hello() {
           const [val, setVal] = React.useState(0);
           const tranformed = React.useMemo(() => val * 2, [val]);
@@ -1383,7 +1383,7 @@ describe('ReactFresh', () => {
       expect(useEffectWithEmptyArrayCalls).toBe(1); // useEffect didn't re-run
 
       // Perform a hot update.
-      let HelloV2 = patch(() => {
+      patch(() => {
         function Hello() {
           const [val, setVal] = React.useState(0);
           const tranformed = React.useMemo(() => val * 10, [val]);
--- a/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
@@ -955,7 +955,7 @@ describe('ReactFresh', () => {
     }
   });
 
-  it('can force remount by changing signature', () => {
+  xit('can force remount by changing signature', () => {
     if (__DEV__) {
       let HelloV1 = render(() => {
         function Hello() {
@@ -1086,31 +1086,31 @@ describe('ReactFresh', () => {
     }
   });
 
-  it('can remount on signature change within a <root> wrapper', () => {
+  xit('can remount on signature change within a <root> wrapper', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => Hello);
     }
   });
 
-  it('can remount on signature change within a simple memo wrapper', () => {
+  xit('can remount on signature change within a simple memo wrapper', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => React.memo(Hello));
     }
   });
 
-  it('can remount on signature change within forwardRef', () => {
+  xit('can remount on signature change within forwardRef', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => React.forwardRef(Hello));
     }
   });
 
-  it('can remount on signature change within forwardRef render function', () => {
+  xit('can remount on signature change within forwardRef render function', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => React.forwardRef(() => <Hello />));
     }
   });
 
-  it('can remount on signature change within nested memo', () => {
+  xit('can remount on signature change within nested memo', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello =>
         React.memo(React.memo(React.memo(Hello))),
@@ -1118,13 +1118,13 @@ describe('ReactFresh', () => {
     }
   });
 
-  it('can remount on signature change within a memo wrapper and custom comparison', () => {
+  xit('can remount on signature change within a memo wrapper and custom comparison', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => React.memo(Hello, () => true));
     }
   });
 
-  it('can remount on signature change within a class', () => {
+  xit('can remount on signature change within a class', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => {
         const child = <Hello />;
@@ -1137,7 +1137,7 @@ describe('ReactFresh', () => {
     }
   });
 
-  it('can remount on signature change within a context provider', () => {
+  xit('can remount on signature change within a context provider', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => {
         const Context = React.createContext();
@@ -1153,7 +1153,7 @@ describe('ReactFresh', () => {
     }
   });
 
-  it('can remount on signature change within a context consumer', () => {
+  xit('can remount on signature change within a context consumer', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => {
         const Context = React.createContext();
@@ -1165,7 +1165,7 @@ describe('ReactFresh', () => {
     }
   });
 
-  it('can remount on signature change within a suspense node', () => {
+  xit('can remount on signature change within a suspense node', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => {
         // TODO: we'll probably want to test fallback trees too.
@@ -1181,7 +1181,7 @@ describe('ReactFresh', () => {
     }
   });
 
-  it('can remount on signature change within a mode node', () => {
+  xit('can remount on signature change within a mode node', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => {
         const child = (
@@ -1196,7 +1196,7 @@ describe('ReactFresh', () => {
     }
   });
 
-  it('can remount on signature change within a fragment node', () => {
+  xit('can remount on signature change within a fragment node', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => {
         const child = (
@@ -1211,7 +1211,7 @@ describe('ReactFresh', () => {
     }
   });
 
-  it('can remount on signature change within a profiler node', () => {
+  xit('can remount on signature change within a profiler node', () => {
     if (__DEV__) {
       testRemountingWithWrapper(Hello => {
         const child = <Hello />;
--- a/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
@@ -744,4 +744,63 @@ describe('ReactFresh', () => {
       expect(newEl.style.color).toBe('blue');
     }
   });
+
+  it('does not re-render ancestor components unnecessarily during a hot update', () => {
+    if (__DEV__) {
+      let appRenders = 0;
+
+      render(() => {
+        function Hello() {
+          const [val, setVal] = React.useState(0);
+          return (
+            <p style={{color: 'blue'}} onClick={() => setVal(val + 1)}>
+              {val}
+            </p>
+          );
+        }
+        __register__(Hello, 'Hello');
+        function App() {
+          appRenders++;
+          return <Hello />;
+        }
+        __register__(App, 'App');
+        return App;
+      });
+
+      expect(appRenders).toBe(1);
+
+      // Bump the state before patching.
+      const el = container.firstChild;
+      expect(el.textContent).toBe('0');
+      expect(el.style.color).toBe('blue');
+      act(() => {
+        el.dispatchEvent(new MouseEvent('click', {bubbles: true}));
+      });
+      expect(el.textContent).toBe('1');
+
+      // No re-renders from the top.
+      expect(appRenders).toBe(1);
+
+      // Perform a hot update for Hello only.
+      patch(() => {
+        function Hello() {
+          const [val, setVal] = React.useState(0);
+          return (
+            <p style={{color: 'red'}} onClick={() => setVal(val + 1)}>
+              {val}
+            </p>
+          );
+        }
+        __register__(Hello, 'Hello');
+      });
+
+      // Assert the state was preserved but color changed.
+      expect(container.firstChild).toBe(el);
+      expect(el.textContent).toBe('1');
+      expect(el.style.color).toBe('red');
+
+      // Still no re-renders from the top.
+      expect(appRenders).toBe(1);
+    }
+  });
 });
--- a/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
@@ -7,6 +7,8 @@
  * @emails react-core
  */
 
+/* eslint-disable no-for-of-loops/no-for-of-loops */
+
 'use strict';
 
 let React;
@@ -17,8 +19,10 @@ describe('ReactFresh', () => {
   let container;
   let familiesByID;
   let familiesByType;
+  let newFamilies;
   let updatedFamilies;
   let performHotReload;
+  let signaturesByType;
 
   beforeEach(() => {
     let enableHotReloading;
@@ -46,10 +50,11 @@ describe('ReactFresh', () => {
 
     if (__DEV__) {
       const {scheduleHotUpdate} = enableHotReloading(familiesByType);
-      performHotReload = function() {
+      performHotReload = function(staleFamilies) {
         scheduleHotUpdate({
           root: lastRoot,
           updatedFamilies,
+          staleFamilies,
         });
       };
     }
@@ -60,9 +65,20 @@ describe('ReactFresh', () => {
   });
 
   function render(version, props) {
+    newFamilies = new Set();
     updatedFamilies = new Set();
+    signaturesByType = new Map();
     const Component = version();
+
+    // Fill in the signatures.
+    for (let family of newFamilies) {
+      const latestSignature = signaturesByType.get(family.currentType) || null;
+      family.currentSignature = latestSignature;
+    }
+
+    newFamilies = null;
     updatedFamilies = null;
+    signaturesByType = null;
     act(() => {
       ReactDOM.render(<Component {...props} />, container);
     });
@@ -70,10 +86,32 @@ describe('ReactFresh', () => {
   }
 
   function patch(version) {
+    // Will be filled in by __register__ calls in user code.
+    newFamilies = new Set();
     updatedFamilies = new Set();
+    signaturesByType = new Map();
     const Component = version();
-    performHotReload();
+
+    // Fill in the signatures.
+    for (let family of newFamilies) {
+      const latestSignature = signaturesByType.get(family.currentType) || null;
+      family.currentSignature = latestSignature;
+    }
+    // Now that all registration and signatures are collected,
+    // find which registrations changed their signatures since last time.
+    const staleFamilies = new Set();
+    for (let family of updatedFamilies) {
+      const latestSignature = signaturesByType.get(family.currentType) || null;
+      if (family.currentSignature !== latestSignature) {
+        family.currentSignature = latestSignature;
+        staleFamilies.add(family);
+      }
+    }
+
+    performHotReload(staleFamilies);
+    newFamilies = null;
     updatedFamilies = null;
+    signaturesByType = null;
     return Component;
   }
 
@@ -85,16 +123,17 @@ describe('ReactFresh', () => {
     let isNew = false;
     if (family === undefined) {
       isNew = true;
-      family = {current: null};
+      family = {currentType: type, currentSignature: null};
       familiesByID.set(id, family);
     }
-    const prevType = family.current;
-    family.current = type;
+    const prevType = family.currentType;
     if (isNew) {
       // The first time a type is registered, we don't need
       // any special reconciliation logic. So we won't add it to the map.
       // Instead, this will happen the firt time it is edited.
+      newFamilies.add(family);
     } else {
+      family.currentType = type;
       // Point both previous and next types to this family.
       familiesByType.set(prevType, family);
       familiesByType.set(type, family);
@@ -112,6 +151,10 @@ describe('ReactFresh', () => {
           break;
       }
     }
+  }
+
+  function __signature__(type, signature) {
+    signaturesByType.set(type, signature);
   }
 
   it('can preserve state for compatible types', () => {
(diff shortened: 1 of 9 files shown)


packages/react-reconciler/src/ReactFiberHotReloading.js
```diff
@@ -0,0 +1,258 @@
...
+  scheduleHotUpdate: HotUpdate => void,
+|};
+
+// Avoid any overhead even in DEV when hot reloading is off:
+export let resolveTypeForHotReloading = (type: any): any => type;
+export let isCompatibleFamilyForHotReloading = (
+  fiber: Fiber,
+  element: ReactElement,
+) => false;
+export let isInvalidatedForHotReloading = (fiber: Fiber | null): boolean =>
```
acdlite: Closure is likely smart enough to inline this, recognize that it always returns false, and strip it out of the conditional check, but relying on seems a bit flaky.

Separately, `export let` could cause Rollup to wrap this module in a closure.

A more predictable pattern would be:

```js
export function isInvalidatedForHotReloading(type) {
  // supportHotReloading is injected by DevTools hook
  if (__DEV__ && supportsHotReloading) {
    // Implementation body
  }
}
```

This will compile to a no-op in prod, which is less likely to confuse Closure.

packages/react-dom/src/__tests__/ReactFresh-test.internal.js
```diff
@@ -0,0 +1,1421 @@
...
+  let newFamilies;
+  let updatedFamilies;
+  let performHotReload;
+  let signaturesByType;
+
+  beforeEach(() => {
+    let enableHotReloading;
+    let lastRoot;
+    global.__REACT_DEVTOOLS_GLOBAL_HOOK__ = {
+      supportsFiber: true,
```
acdlite: If this sets `supportsHotReloading` to `true`, then we can avoid the `enableHotReloading` stuff.

packages/react-dom/src/__tests__/ReactFresh-test.internal.js
```diff
@@ -0,0 +1,1421 @@
...
+    ReactDOM = require('react-dom');
+    act = require('react-dom/test-utils').act;
+    container = document.createElement('div');
+    document.body.appendChild(container);
+
+    familiesByID = new Map();
+    familiesByType = new WeakMap();
+
+    if (__DEV__) {
+      const {scheduleHotUpdate} = enableHotReloading(familiesByType);
```
acdlite: `scheduleHotUpdate` can be injected into DevTools the same as `overrideHookState` and `overrideProps`.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -292,6 +294,13 @@ function areHookInputsEqual(
   nextDeps: Array<mixed>,
   prevDeps: Array<mixed> | null,
 ) {
+  if (__DEV__) {
+    // Force to remount effects, callbacks, and memo for hot reloaded components.
+    if (isInvalidatedForHotReloading(currentlyRenderingFiber)) {
```
acdlite: Why is this in this function? I would expect it to be in `renderWithHooks`, since that is called once per component instead of once per hook with deps.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -338,6 +347,10 @@ export function renderWithHooks(
   refOrContext: any,
   nextRenderExpirationTime: ExpirationTime,
 ): any {
+  if (__DEV__) {
+    Component = resolveTypeForHotReloading(Component);
```
acdlite: This makes conceptual sense but it means we have to remember not to read `fiber.type` anywhere else. We faced a similar problem with `React.lazy`, which is one of the reasons we split the `type` field into `elementType` (the thing we compare during reconciliation) and `type` (the implementation of the render phase).

Let's move this to `cloneFiber` and `createFiber` and assign the resolved type to the `fiber.type` field.

gaearon:
>The reason I like this is that it lets you remove isCompatibleFamilyForHotReloading from the reconciliation hot path. It also matches what we do for React.lazy.

FWIW I checked this on several products, and it's not *that* hot of a path. The equality check is true ~1000x times more often than false, so usually we don't even go into there. That's not to say it's completely free, but just that it's even less hot than I anticipated.

acdlite:
@gaearon It's not the `isCompatibleFamilyForHotReloading` function itself that is reason for caution, since as you say it's rare that it's ever called. It's the extra check and function call that surrounds it.

packages/react-reconciler/src/ReactFiberHooks.js
gaearon: I want to reset every Hook with dependencies, but keep state of other Hooks. So `useEffect`, `useCallback`, `useMemo`, `useImperativeHandle` should behave as if their deps changed (even if they were `[]` in the first place). But `useRef`, `useState`, `useReducer` should retain their values.

I wasn't sure what's an easy way to write this in `renderWitHooks`. I'd have to traverse the list and tear them down manually, but they don't have "types" in the list so I'd need to add those?

It seemed easier to just override the comparison to force any dep comparison to fail.

packages/react-dom/src/__tests__/ReactFresh-test.internal.js
gaearon: I wanted three things: 

1. No overhead if hot reloading is supported
2. A way for the hot runtime to inject family map into React
3. A way for the hot runtime to plug into global Hook

Note hot runtime isn't part of React DevTools. It just reuses the same mechanism. So it'll have to play nicely with DevTools itself. How does that happen?

If we make this `supportsHotReloading: true`, I supposed it'll be `false` by default in DevTools, and then hot runtime would mutate it to true? Note it would still need to override `inject` in order to gain access to `scheduleHotUpdate`, as well as `onCommitFiberRoot` to track roots.

How does the family map get injected into React? A separate method like `injectHotReloadingMap`?

Maybe family map could be a part of first `scheduleHotReload` call. While the map would have to be the same every time, this would mean we may zero price until you actually use the feature the first time.

packages/react-reconciler/src/ReactFiberHooks.js
gaearon: How confident are we that those will always run before we do some work? Are there any cases at all where they wouldn't? Stale types make me nervous.

My justification for this was that I only ever want to support hot reloading for one thing — rendering function components. And rendering ultimately happens here.

For example I don't want to accidentally mess with types for classes. Or even for resolving default props. There's also a bunch of `fiber.type === fiber.elementType` checks that add extra work like resolving, and I hoped not to cause those codepaths to fork after a reload. Seems more confusing.

How confident are you that this would be better?

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: Addressed.

packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -233,6 +234,21 @@ function warnOnFunctionType() {
   );
 }
 
+function isCompatibleType(fiber: Fiber, element: ReactElement): boolean {
+  if (__DEV__) {
+    // Keep the first check inline as it's overwhelmingly true in practice.
```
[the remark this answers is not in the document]
gaearon: Addressed.

packages/react-dom/src/__tests__/ReactFresh-test.internal.js
gaearon: Addressed.

packages/react-dom/src/__tests__/ReactFresh-test.internal.js
gaearon: Addressed.

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: I was thinking you would set a module variable called something like `ignorePreviousDependencies = isInvalidatedForHotReloading(fiber)` inside  `renderWithHooks`, and then use that in `areHookInputsEqual`.

I suppose it's not a big deal since it's DEV but comparing deps is a fairly hot code path, and this adds an unconditional map lookup.

acdlite: I generally don't like adding new module level variables but since this one would only be set at the beginning of `renderWithHooks`, and remain the same throughout the entire phase, seems ok.

gaearon: Ah I see what you mean now. Thanks for explaining. This makes sense.

gaearon: Pushed.

packages/react-reconciler/src/ReactFiberHooks.js
gaearon: Fixed.

packages/react-reconciler/src/__tests__/ReactIncrementalErrorReplay-test.internal.js
```diff
@@ -10,21 +10,22 @@
 'use strict';
 
 describe('ReactIncrementalErrorReplay-test', () => {
+  const React = require('react');
+  const ReactTestRenderer = require('react-test-renderer');
```
gaearon: This test didn't load anymore without a real renderer because internal module transitive deps now reached the host config.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -357,18 +361,22 @@ function updateMemoComponent(
       // SimpleMemoComponent codepath doesn't resolve outer props either.
       Component.defaultProps === undefined
     ) {
+      let resolvedType = type;
+      if (__DEV__) {
+        resolvedType = resolveFunctionForHotReloading(type);
```
sebmarkbage: Does this works for updates against the SimpleMemoComponent? Does the thing in ReactFiber know to deal with unwrapping the "Memo" HoC and update the `.type`?

packages/react-reconciler/src/ReactFiberHotReloading.js
```diff
@@ -0,0 +1,312 @@
...
+              // delete them too until we find a non-memo parent.
+              // TODO: we might need something similar for Suspense
+              // or its special "fake fragment" node.
+              fiberToRemount = parent;
+            } else {
+              break;
+            }
+          }
+          // Schedule the parent.
+          const parent = fiberToRemount.return;
```
acdlite: Can you explain to me why the parents need to force re-render? Isn't it sufficient that the component whose code changed re-renders without bailing out?

(Sorry I didn't review this part last time; I must have skimmed over it for some reason.)

packages/react-reconciler/src/ReactFiberBeginWork.js
gaearon: My understanding is:

1. First time we "upgrade" to simple memo, type gets resolved on this line.
2. Next time we work on it, `createWorkInProgress` resolves it for simple memo.

The thing in ReactFiber doesn't deal with regular memo at all because we only need resolving for simple memo. Regular memo doesn't call into user code (except the comparison function, but I don't aim to reload it).

packages/react-reconciler/src/ReactFiberHotReloading.js
```diff
@@ -0,0 +1,312 @@
...
+        return true;
+      }
+    }
+    return false;
+  } else {
+    return false;
+  }
+}
+
+export function shouldForceRenderForHotReloading(fiber: Fiber | null): boolean {
```
sebmarkbage: Forcing a rerender is a very loosely defined concept. We need to define this more clearly and we shouldn't call it "force" because it doesn't mean anything. E.g. if the rerender returns the same thing, does that mean we can bail out? What if we can determine early that nothing has changed based on a compiler driven memoization in the component? When are we allowed to bail out?

I think this shouldn't actually be treated as a forced rerender. It should be modeled as changing the "type".

`current.type !== workInProgress.type`

We already do this for `current.ref !== workInProgress.ref` and `current.props !== workInProgress.props`. This is just the same thing for a third field. The rest should follow.

What breaks down if you model it that way?

packages/react-reconciler/src/ReactFiberHotReloading.js
acdlite: Your gist includes this line

> Currently there are some bailouts that are impossible to override (like on a root).

but that doesn't apply to deep updates. Can you clarify what you mean here?

gaearon: Note there are two different codepaths.

```js
if (staleFamilies.has(family)) {
  fiberToRemount.elementType = 'DELETED';
  // schedule parent
} else if (updatedFamilies.has(family)) {
  // schedule self
}
```

The first code path is about *forcing remount (despite equivalent type)*, whereas the second path is about *forcing re-render*.

Forcing remount is necessary in case you change order of Hooks. We could also later use that to force remount edited classes. To force remount, we mutate `elementType` to `'DELETED'` and *schedule the parent*.

For normal edits, we don't force remount. We just force re-render of same fiber.

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: The thing that I want to happen is for it to not skip `reconcileFibers`. Then it'll "discover" its child type is now `'DELETED'`, and force it to remount.

acdlite: Why does is triggering `reconcileFibers` in the parent a requirement? In case the tag needs to change?

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: Can you explain how you expect the data structures for memo to work? There are two. The render function and the object around it. createWorkInProgress will resolve the memo object around it but that's wrong because the type of the `.type` field has now changed and it needs the render function now.

packages/react-reconciler/src/ReactFiberHotReloading.js
acdlite: The approach I would suggest off the top of my head (haven't thought through this fully so sorry if there's something obviously wrong)

- Schedule on update the reloaded fiber. Don't do anything special to the parent.
- When you get to `createWorkInProgress`, instead of cloning from `current`, create a new work-in-progress fiber that's disconnected (no alternate).
- For each of the old fiber's children, schedule a deletion effect.

gaearon: Something's gotta tag that child as receiving a deletion? I thought that's what `reconcileChildFibers` does. I'm fuzzy on how child reconciliation works.

sebmarkbage: I need the ability to remount a child in production for dehydrated suspense boundaries. The upgrade logic has a bug otherwise. So we need that anyway but not implemented like this.

We should think of a different technique than rerendering the parent to do this.

gaearon: Also, I need to insert the new retried child at its place.

gaearon: lol github comments are so out of order

gaearon: OK, I can think more about this

gaearon: Btw I'll also need to remount all currently failed boundaries on hot update. Planned to keep them in a weak set. This is so that you can recover after simple errors without refresh.

packages/react-reconciler/src/ReactFiberBeginWork.js
gaearon: My understanding is that normal Memo never runs rendering function by itself. It always either delegates to child fiber, or gets upgraded:

* Either it's Memo-that-does-a-bailout + FunctionComponent (or whatever's in memo).
* Or it gets upgraded to SimpleMemo.

So in practice only FunctionComponent and SimpleMemo need to resolve the function. Conveniently, that's what their type is (a function).

Memo doesn't need to resolve a function because it doesn't run it.

(ForwardRef is an exception because it also runs user code *and* its type is a wrapper object. That's why ForwardRef has a special codepath where we recreate the wrapper object if we only have a new inner function alone.)

The wrapper object doesn't play a role at all, except at the point of `__register__` call (which will does another `__register__` for the inner function).

packages/react-reconciler/src/ReactFiberHotReloading.js
sebmarkbage: Can you explain the scenario when you need this? We’re trying to guess over here and not succeeding.

gaearon: The scenario where I need what exactly?

gaearon: I have two reasons to want a "force remount" functionality:

* Changing Hooks signature. If some function gets a different set of Hooks, I want to force it to remount without even trying a hot render. However, I don't have a reliable way to propagate new function identity to all callers. So I want to tell React "remount all things with this family during this hot update".

* Failed error boundaries. If my hot edit led to a runtime error (as it often would due to typos etc), closest error boundary will activate together with global runtime error overlay. This is good. But once I save again, I want to "retry". That component doesn't exist anymore, and there's no way to tell error boundary to "reset state" (reset to what?) But I can remount it. That's what I want to do for failed boundaries during a hot update.

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: >but that isn't relevant to deep updates (setState). Can you clarify what you mean here?

This was about forcing remount of a child. (Which needs a re-render of the parent.) Indeed if we follow a different approach for forcing remount, this comment is outdated.

packages/react-reconciler/src/ReactFiberHotReloading.js
acdlite: @gaearon It works by scheduling a `Deletion` effect on the fiber that's deleted and adding them to the parent's effect list. You can do that manually without reconciling the parent.

packages/react-reconciler/src/ReactFiber.js
```diff
@@ -431,6 +436,25 @@ export function createWorkInProgress(
     workInProgress.treeBaseDuration = current.treeBaseDuration;
   }
 
+  if (__DEV__) {
+    switch (workInProgress.tag) {
+      case IndeterminateComponent:
+      case FunctionComponent:
+      case SimpleMemoComponent:
+        workInProgress.type = resolveFunctionForHotReloading(
+          workInProgress.type,
```
sebmarkbage: I believe this should be `current.type`.

I think this answers my question around simple memo. You're not getting a new `type` from `element.type` but reading it from the fiber which is the underlying render function. I would've expected you to read the type again from element but I guess this is fine too. It just means that if we ever start getting updatable types during reconciliation this would break but we can deal with that then.

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: I’d also need to schedule an insertion for the new version, right?

sebmarkbage: An error boundary doesn’t always setState on itself. Sometimes it sets state on a parent. How do you plan to deal with that?

sebmarkbage: > Changing Hooks signature

I think there's some subtle nuances to think about here. I originally thought of this as the "type" changing the family but I think that it explicitly can't use that mechanism. It must preserve the same family because some parent component can still have a React element that references the old function. Rendering a new instance using that element should get the new implementation.

Therefore, what you're looking for here is explicitly to reset the state rendered by the component but the "self" identity from the perspective of the outside conceptually still the same. In fact, you could argue that it shouldn't even delete the old fiber, but remount it in place (although granted, deleting and remounting is might still be the easiest way to do that).

packages/react-reconciler/src/ReactFiberHotReloading.js
```diff
@@ -0,0 +1,312 @@
...
+  if (__DEV__) {
+    if (fiber === null) {
+      return false;
+    }
+    if (fibersWithForcedRender === null) {
+      // Not hot reloading now.
+      return false;
+    }
+    return fibersWithForcedRender.has(fiber);
```
sebmarkbage: I don't think we should model these as synchronous. This should work just as well as an async operation. This is trying to cheat the async system by being clever and now that is giving me a lot of homework to figure out if that is actually safe and when it might not be.

Comparing `current.type !== workInProgress.type` solves one of the cases and works in async. Let's see if we can also solve the other one.

(Maybe we still do them sync for purposes of avoiding yielding bugs but that could trigger warnings e.g. against suspending in sync mode.)

sebmarkbage: Actually, isn't assuming synchronous rendering already broken for deprioritized subtrees like hidden and dehydrated trees? flushSync doesn't force those into Sync priority.

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: >An error boundary doesn’t always setState on itself. Sometimes it sets state on a parent. How do you plan to deal with that?

That's a good point. But I don't have a clear sense of how common that is in practice — especially with the future move to `getDerivedStateFromError`. In that world it seems like it's still a good enough heuristic to reset just the boundary alone.

>It must preserve the same family because some parent component can still have a React element that references the old function. Rendering a new instance using that element should get the new implementation.

Yes, I came to the same realization. I think we don't disagree here? That's what I tried to model.

I agree it would be nice to "remount in place" but practically speaking I didn't want to get very "custom" in reconciliation unless we agreed that I'd need to.

gaearon:
So as I understand, the outstanding blockers are:

* Get rid of the synchronous Set mechanism. Add a test for offscreen trees to confirm.
* Get rid of the remounting-via-parent implementation. Do something custom. Once that's in place, we don't need to skip bailouts as much, and can probably stick with comparing types.

gaearon:
Verified the new offscreen test passes with commented out logic enabled on top of https://github.com/facebook/react/pull/15666.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2106,18 +2115,99 @@ function bailoutOnAlreadyFinishedWork(
...
 function beginWork(
   current: Fiber | null,
   workInProgress: Fiber,
   renderExpirationTime: ExpirationTime,
 ): Fiber | null {
   const updateExpirationTime = workInProgress.expirationTime;
 
+  if (__DEV__) {
+    if (workInProgress._debugNeedsRemount && current !== null) {
+      // This will restart the begin phase with a new fiber.
```
acdlite: Any reason this returns instead of re-assigning `workInProgress` variable? I can see potential problems with either approach but I wonder why you chose this one.

gaearon: It’s what Seb suggested in the chat? I thought reassigning is confusing since outer code doesn’t know about it. This seemed a bit more explicit.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2106,18 +2115,99 @@ function bailoutOnAlreadyFinishedWork(
...
+
   if (current !== null) {
     const oldProps = current.memoizedProps;
     const newProps = workInProgress.pendingProps;
 
-    if (oldProps !== newProps || hasLegacyContextChanged()) {
+    if (
+      oldProps !== newProps ||
+      hasLegacyContextChanged() ||
+      (__DEV__ ? workInProgress.type !== current.type : false)
```
acdlite: I find this comparison confusing. I assume it's to handle the case where you force a fiber to re-render without bailing out. Could you use an explicit flag instead? E.g. Rename `_debugNeedsRemount` and make it an enum type.

packages/react-reconciler/src/ReactFiberHotReloading.js
```diff
@@ -0,0 +1,258 @@
...
+        type !== null &&
+        type !== undefined &&
+        typeof type.render === 'function'
+      ) {
+        // ForwardRef is special because its resolved .type is an object,
+        // but it's possible that we only have its inner render function in the map.
+        // If that inner render function is different, we'll build a new forwardRef type.
+        const currentRender = resolveFunctionForHotReloading(type.render);
+        if (type.render !== currentRender) {
+          return {...type, render: currentRender};
```
acdlite: Nit: No spread. You can use the Flow type in `ReactTypes` to make sure it stays in sync.

packages/react-reconciler/src/ReactFiberHotReloading.js
```diff
@@ -0,0 +1,258 @@
...
+}
+
+export function scheduleHotUpdate(hotUpdate: HotUpdate): void {
+  if (__DEV__) {
+    // TODO: warn if its identity changes over time?
+    familiesByType = hotUpdate.familiesByType;
+
+    const {root, staleFamilies, updatedFamilies} = hotUpdate;
+    flushPassiveEffects();
+    flushSync(() => {
```
acdlite: Does this still need to be sync?

packages/react-reconciler/src/ReactFiberHotReloading.js
```diff
@@ -0,0 +1,258 @@
...
+      scheduleFibersWithFamiliesRecursively(
+        root.current,
+        updatedFamilies,
+        staleFamilies,
+      );
+    });
+    flushPassiveEffects();
```
acdlite: What's this for?

packages/react-reconciler/src/ReactFiberHotReloading.js
```diff
@@ -0,0 +1,258 @@
...
+    if (familiesByType === null) {
+      throw new Error('Expected familiesByType to be set during hot reload.');
+    }
+
+    if (candidateType !== null) {
+      const family = familiesByType.get(candidateType);
+      if (family !== undefined) {
+        if (staleFamilies.has(family)) {
+          fiber._debugNeedsRemount = true;
+          scheduleWork(fiber, Sync);
```
acdlite: Same here. Do the hot updates need to be sync?

packages/react-reconciler/src/ReactFiberBeginWork.js
gaearon: But Seb suggested to model it like this above? It used to be a separate thing (although a Set rather than a field so I guess field would still be better than a Set).

I can change but I'm also not sure it's productive to keep changing it back and forth depending on who's reviewing. :-)

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: It doesn't _have to_ be but for the feedback loop to feel good you don't really want to wait. This also makes it easier to do post-processing like highlight affected nodes visually.

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: I can remove. I figured I want to do as much as I can related to the update synchronously but I don't have any particular justification except that the developer is waiting for reload and its effects to finish. I guess it's fair we don't want to deviate from real behavior too much.

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: They don't have to be but the developer is waiting and there's nothing happening that could be more important at that moment. It also makes it simpler to highlight updated nodes visually after the fact. (Unless I get notified about completion somehow.)

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2106,18 +2115,99 @@ function bailoutOnAlreadyFinishedWork(
...
+    }
+    current.nextEffect = null;
+    current.effectTag = Deletion;
+
+    newWorkInProgress.effectTag |= Placement;
+
+    // Restart work from the new fiber.
+    return newWorkInProgress;
+  } else {
+    return null;
```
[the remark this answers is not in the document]
gaearon: I wrote it this way in case it accidentally gets into the bundle. But I guess we'd have to pay for that code in either case and throwing would help detect it sooner.

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: I didn't want to lose `displayName` that might be declared on outer type. But I guess I can copy it over if it exists.

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: Removed

packages/react-reconciler/src/ReactFiberHotReloading.js
gaearon: Removed spread

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: I think it makes conceptual sense but if we're going to model that way then I find it odd that it's only applied in the paths that are specific to hot reloading. For example, it's not obvious why you compare types in `updateSimpleMemoComponent` but not `updateMemoComponent`. I know the reason, because I have context on this PR, but it's not readily apparent.

If you add a comment above each of these comparison then I'm ok with it.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2106,18 +2115,102 @@ function bailoutOnAlreadyFinishedWork(
...
+        if (prevSibling === null) {
+          throw new Error('Expected to find the previous sibling.');
+        }
+      }
+      prevSibling.sibling = newWorkInProgress;
+    }
+
+    // Delete the old fiber and place the new one.
+    // Since the old fiber is disconnected, we have to schedule it manually.
+    const last = returnFiber.lastEffect;
```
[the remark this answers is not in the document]
gaearon: This is copy paste from dehydrated boundary code below btw. So if we change we probably should in both.

PR #15631, opened 2019-05-12
warn if passive effects get queued outside of an act() call.

threepointone:
This version uses a feature flag to disable the warning in tests. It isn't a viable options since we'd have to disable a whole lot of tests that run in yarn test-build See https://github.com/facebook/react/pull/15763 instead 

---
based on #15591. this might be a bit controversial, but I agree with the reasoning. of note, we don't modify our own tests to satisfy this warning, instead using a feature flag to disable the warning for some tests. this is because we're testing the actual sequencing of work in these tests, and don't want to flush everything with act().

you probably want to see just this commit - https://github.com/facebook/react/pull/15631/commits/9f7a346ee33f161e033484a3336ffdf966b82324

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -1604,7 +1604,7 @@ function commitLayoutEffects(
 
 export function flushPassiveEffects() {
   if (rootWithPendingPassiveEffects === null) {
-    return;
+    return false;
   }
   const root = rootWithPendingPassiveEffects;
   const expirationTime = pendingPassiveEffectsExpirationTime;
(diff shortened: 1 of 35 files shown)


packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -2088,7 +2089,32 @@ function warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber) {
...
 }
 
-function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
+export function warnIfNotCurrentlyActingEffectsInDEV(fiber: Fiber): void {
+  if (__DEV__) {
+    if (warnAboutUnactedEffectsinDEV) {
+      if (ReactShouldWarnActingUpdates.current === false) {
+        warningWithoutStack(
+          false,
+          'An effect from %s inside a test was not wrapped in act(...).\n\n' +
```
acdlite: Nit: This makes it sound like `useEffect` is supposed to be wrapped in `act`.

acdlite:
^ approved once base PR is merged

threepointone:
I'm rewriting this to (hopefully) not need the feature flag. it'll mean rewriting a lot  of the tests.

acdlite:
@threepointone Let's land the change before updating a bunch of tests? I would prefer not to couple those.

threepointone:
made an attempt at doing this without the feature flag here - https://github.com/facebook/react/pull/15763

threepointone:
abandoning this for https://github.com/facebook/react/pull/15763

PR #15633, opened 2019-05-13
Event API: ensure preventDefault works for nested targets

trueadm:
This PR fixes a case with the Press event responder, where `preventDefault` would not be triggered when an anchor element's child was the target (not the anchor element itself). This uses a `isTargetWithinHostComponent` under-the-hood to enable this functionality.


packages/react-events/src/__tests__/Press-test.internal.js
```diff
@@ -1965,6 +1965,25 @@ describe('Event responder: Press', () => {
       expect(preventDefault).toBeCalled();
     });
 
+    it.only('prevents native behaviour by default #2', () => {
```
threepointone: stray .only

PR #15634, opened 2019-05-13
Event API: Focus onFocusVisble

trueadm:
This PR adds `onFocusVisble` to the Focus module. We already have `onFocusVisbleChange`, so there wasn't much needed to add this. Added relevant tests to confirm it works as intended too. We may not need this – we could add `pointerType` to `focus` and `blur` event objects instead.

### DIFF
--- a/packages/react-events/src/Focus.js
+++ b/packages/react-events/src/Focus.js
@@ -29,7 +29,12 @@ type FocusState = {
   isLocalFocusVisible: boolean,
 };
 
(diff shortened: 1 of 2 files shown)


ahtee:
Not too sure I like this, as it can be confused with `onFocusVisibleChange` I suppose this addition could not have been apart of the existing property (or method :) )?

trueadm:
Yeah, I think a better approach is adding `pointerType` to the event object. I'll close this and put up another PR with that approach tomorrow.

PR #15636, opened 2019-05-13
Add test for nested avoided boundaries

sebmarkbage:
I thought that this might be broken but looks like it's not. I still suspect that we'd have to do the rethrowing (of multiple captured promises) in the complete phase. I can't think of a case where this breaks though. I think this is because the potential bug only happens if we're in an "invisible" tree but we only want that to count if the invisible tree is a preferred boundary. In the case an avoided boundary is the only thing that is invisible, it doesn't matter because we want to use the inner most boundary anyway in those cases.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
@@ -1572,6 +1572,37 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     expect(ReactNoop.getChildren()).toEqual([span('A'), span('C'), span('B')]);
   });
 
+  it('favors showing the inner fallback for nested top level avoided fallback', async () => {
+    function Foo({showB}) {
+      Scheduler.yieldValue('Foo');
+      return (
+        <Suspense
+          unstable_avoidThisFallback={true}
+          fallback={<Text text="Loading A..." />}>
+          <Text text="A" />
+          <Suspense
+            unstable_avoidThisFallback={true}
+            fallback={<Text text="Loading B..." />}>
+            <AsyncText text="B" ms={5000} />
+          </Suspense>
+        </Suspense>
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1842,6 +1842,80 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     expect(ReactNoop.getChildren()).toEqual([span('A'), span('C'), span('B')]);
   });
 
+  it('favors showing the inner boundary for nested top level avoided boundaries', async () => {
```
ahtee: So for this, the inner suspense block should show in the parent Suspense when nesting, or is there additional detail for this test...?

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1842,6 +1842,80 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
+      'Loading B...',
+    ]);
+    // Flush to skip suspended time.
+    Scheduler.advanceTime(600);
+    await advanceTimers(600);
+
+    expect(ReactNoop.getChildren()).toEqual([span('A'), span('Loading B...')]);
+  });
+
+  it('keeps showing an avoided parent boundary if it is already showing', async () => {
```
acdlite: I don't understand what this is meant to test. Maybe the term "avoided parent boundary" is throwing me off. Are you referring to the primary contents or the fallback? I assumed fallback when I read the test name, but the test itself doesn't ever show the parent boundary's fallback.

acdlite: Oh I suppose this is testing that we never switch to "Initial load." Makes sense.

Issue #14319, opened 2018-11-24
API for display name on forwardRef, memo and potential future exotic components


eps1lon:
TL;DR: Can you expose [shared/getComponentName](https://github.com/facebook/react/blob/master/packages/shared/getComponentName.js)?

**Do you want to request a *feature* or report a *bug*?**
Expose an API to get the display name of every component (in `__DEV__` only).

**What is the current behavior?**
Most of the ecosystem still uses `Component.displayName || Component.name || someFallbackName` 
 (with some branching depending on the type of `Component`) when setting the display name of an enhanced component i.e. `connect()(WrappedComponent)` will result in `"connect(WrappedComponent)"` as a `displayName`.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**
Since components created by `forwardRef` or `memo` are not actual functions these higher-order components are not able to determine a proper display name while `react-devtools` is able to:

https://codesandbox.io/s/zqj9v50243

- `react-redux` creates `"connect(Component)"`
- `react-router` creates `"withRouter(undefined)"`

**What is the expected behavior?**
The new "exotic-components" should work with the existing 3rd party libraries WRT to `displayName`.

Now there are a couple of solutions to this issue:
1. **Edit:** Expose [shared/getComponentName](https://github.com/facebook/react/blob/master/packages/shared/getComponentName.js)
2. This is the responsibility of the ecosystem. It should provide a solution and maintain it. Somewhat blocked by #12882, related: #12932
3. Grant access to the functionality used in `react-devtools` (or would this only work on the fibers?)
4. Set a `name` (or `displayName` no preference here) property on those "exotic-components" (don't know how to call them). Naive implementation e.g.: `name: 'ForwardRef(' + fn.name + ')'`.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
I guess this started with `forwardRef` in 16.3.

martinmckenna:
+1 here. My current solution to smoke testing React.memo components is to add a custom attribute such as

`qa-my-child-component`

and then with jest-enzyme

`shallow(<MyComponent />).find('[qa-my-child-component]').toHaveLength(1)`.

Ideally, I'd like to do

[...]

PR #13197, opened 2018-07-12
Remove ad-hoc forks of getComponentName() and fix it

gaearon:
1. Fixes `getComponentName()` for a few cases where it was incorrectly reading the type (https://github.com/facebook/react/commit/a482fb41c35448cc1f2cc2e1e7b655a11e8f8957). We didn't notice because we rarely call it for anything other than user-defined components.
2. Changes its signature to require just the `type` alone rather than the Fiber. This might seem like it makes the calling code longer. However we often *already* have the type destructured anyway when we call it. I also remember from writing some of that code that it was awkward to always *have to* pass a Fiber around when I only needed a type.
3. This lets us remove two ad-hoc forks of `getComponentName()` that don't handle all the cases it does.

PR #15638, opened 2019-05-13
refactor(getComponentName): Use element types instead of fiber types

eps1lon:
Since fiber types can change during work (see MemoComponent -> SimpleMemoComponent) it seems more appropriate to use the elementType instead to determine a component name.

Makes #15313 easier (see https://github.com/facebook/react/pull/15313#discussion_r275157235). It would also enable #14319 since we no longer rely on internal data structures (fiber types) but public element types.

### DIFF
--- a/packages/shared/getComponentName.js
+++ b/packages/shared/getComponentName.js
@@ -41,26 +41,27 @@ function getContextName(type: ReactContext<any>) {
   return type.displayName || 'Context';
 }
 
-function getComponentName(type: mixed): string | null {
-  if (type == null) {
+function getComponentName(elementType: mixed): string | null {
+  if (elementType == null) {
     // Host root, text node or just invalid type.
     return null;
   }
   if (__DEV__) {
-    if (typeof (type: any).tag === 'number') {
+    // TODO: protect against potential false positives from user-defined statics
+    if (typeof (elementType: any).tag === 'number') {
       console.error(
         'Received an unexpected object in getComponentName(). ' +
           'This is likely a bug in React. Please file an issue.',
       );
     }
   }
-  if (typeof type === 'function') {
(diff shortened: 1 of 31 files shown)


packages/shared/getComponentName.js
```diff
@@ -43,27 +43,28 @@ function getWrappedName(
...
-  if (type == null) {
+function getComponentName(elementType: mixed): string | null {
+  if (elementType == null) {
     // Host root, text node or just invalid type.
     return null;
   }
   if (__DEV__) {
-    if (typeof (type: any).tag === 'number') {
+    // TODO: protect against potential false positives from user-defined statics
+    if (typeof (elementType: any).tag === 'number') {
```
eps1lon: The check was introduced in #13197. Seems like this was meant as a helper for the transition period since it was used on the full fiber previously.

It's probably better to [remove this altogether](https://github.com/eps1lon/react/commit/2be613fc1cb9f286ec434e1692820ea325fba9c3). An unknown component name should appear in the tests leading to a fail instead of a different error message.

[Using `_debugID`](https://github.com/eps1lon/react/commit/4403e4bfb40efc7ca7446144c64bd3d7f5d02cfb) is likely best to reduce the number of false positives.

[Using `isValidElementType`](https://github.com/eps1lon/react/commit/a2737cdb2f59ca033b2aa37ee2f1e1bee608c40d) leads to breakage in the [internal React.lazy tests](https://github.com/facebook/react/blob/9055e31e5c82d03f0a365c459f7bc79e402dbef5/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js#L562).

eps1lon:
I'll resolve merge conflicts.

eps1lon:
Closing due to missing interest.

PR #15640, opened 2019-05-14
Separate Offscreen and Idle expiration times

acdlite:
Offscreen work has some special semantics regarding tearing and Suspense. We don't necessarily want these semantics when scheduling an update at Scheduler's "idle" priority level. This splits the `Never` expiration time into separate `Offscreen` and `Idle` expiration times.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScheduler.js
+++ b/packages/react-reconciler/src/ReactFiberScheduler.js
@@ -90,7 +90,7 @@ import {
 import {
   NoWork,
   Sync,
-  Never,
+  Idle,
   msToExpirationTime,
   expirationTimeToMs,
   computeInteractiveExpiration,
@@ -98,6 +98,7 @@ import {
   inferPriorityFromExpirationTime,
   LOW_PRIORITY_EXPIRATION,
   Batched,
+  Offscreen,
 } from './ReactFiberExpirationTime';
 import {beginWork as originalBeginWork} from './ReactFiberBeginWork';
 import {completeWork} from './ReactFiberCompleteWork';
@@ -293,7 +294,7 @@ export function computeExpirationForFiber(
       expirationTime = computeAsyncExpiration(currentTime);
       break;
     case IdlePriority:
-      expirationTime = Never;
+      expirationTime = Idle;
       break;
     default:
       invariant(false, 'Expected a valid priority level');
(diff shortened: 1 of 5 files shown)


NE-SmallTown:
but the semantic of word "Offscreen" seems not appropriate for "expiration time"

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -909,7 +914,7 @@ function renderRoot(
       return commitRoot.bind(null, root);
     }
     case RootSuspended: {
-      if (!isSync) {
+      if (!isSync && expirationTime !== Offscreen) {
```
sebmarkbage: This should condition should never happen, right? I'm not sure this is actually the correct way of modeling it because if we did batch other work in with this level for some other reason, then you would still want to suspend.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -1795,6 +1800,14 @@ export function pingSuspendedRoot(
     return;
   }
 
+  if (suspendedTime === Offscreen) {
```
sebmarkbage: Why special case this at all? Should this be an invariant?

sebmarkbage: This should restart the render if a ping is in the middle of an offscreen render.

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1616,4 +1616,66 @@ describe('ReactSuspenseWithNoopRenderer', () => {
 
     expect(ReactNoop.getChildren()).toEqual([span('Loading A...')]);
   });
+
```
sebmarkbage: Can we add a variant of this test that tests this with an update explicitly scheduled at Idle priority?

sebmarkbage:
@gaearon’s other fix reminded me that we can’t just rely on Offscreen having no updates in it for ensuring that it doesn’t get pinged. We should add some test where the update happens inside the hidden subtree and suspends. In that case there’s a processed update in the Offscreen render pass.

acdlite:
Outdated

PR #15644, opened 2019-05-14
flush pending passive effects early on an interactive update

threepointone:
related to #15057

ported over Dan's PR from https://github.com/facebook/react/pull/15067 to discuss this. 

- is our recommendation to folks to use `interactiveUpdates(() => {...})` for all 'outside' updates?
- alternately, what should the OP code look like? this situation of 'do something when a value becomes something' comes up often. `emitEffect`?


threepointone:
abandoning for #15650

PR #15645, opened 2019-05-14
Event API: adds pointerType to Focus events

trueadm:
This PR is a follow up from the ideas in https://github.com/facebook/react/pull/15634. Specifically, rather than create a new type of event, this PR adds a new `pointerType` property to all Focus responder module events. This field lets the user know what type of input device was used to trigger a focus – `touch`, `pen`, `mouse` and `keyboard`.

### DIFF
--- a/packages/react-events/src/Focus.js
+++ b/packages/react-events/src/Focus.js
@@ -26,13 +26,16 @@ type FocusState = {
   focusTarget: null | Element | Document,
   isFocused: boolean,
   isLocalFocusVisible: boolean,
+  pointerType: PointerType,
 };
 
+type PointerType = '' | 'mouse' | 'keyboard' | 'pen' | 'touch';
 type FocusEventType = 'focus' | 'blur' | 'focuschange' | 'focusvisiblechange';
 
 type FocusEvent = {|
   target: Element | Document,
   type: FocusEventType,
+  pointerType: PointerType,
   timeStamp: number,
 |};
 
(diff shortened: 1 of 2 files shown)


packages/react-events/src/Focus.js
```diff
@@ -26,13 +26,16 @@ type FocusState = {
   focusTarget: null | Element | Document,
   isFocused: boolean,
   isLocalFocusVisible: boolean,
+  pointerType: PointerType,
 };
 
+type PointerType = '' | 'mouse' | 'keyboard' | 'pen' | 'touch';
```
TrySound: Why empty string?

trueadm: https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType

> If the device type cannot be detected by the browser, the value can be an empty string ("")

packages/react-events/src/Focus.js
```diff
@@ -148,6 +184,32 @@ function unmountResponder(
   }
 }
 
+function handleRootPointerEvent(
+  type: string,
```
necolas: We can get type and target from event rather than passing them to this function

Andarist:
Is this a good name for this property? Keyboard is actually not considered to be a pointer device by various HTML etc APIs. Including this one https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType

packages/react-events/src/Focus.js
```diff
@@ -249,6 +305,7 @@ const FocusResponder = {
           nativeEvent.key === 'Tab' &&
           !(nativeEvent.metaKey || nativeEvent.altKey || nativeEvent.ctrlKey)
         ) {
```
Andarist: Im wondering if this is a correct condition. MacOS (and therefore Safari) wont focus i.e. radios and checkboxes on Tab alone. U need to press down Tab+Alt for that

PR #15646, opened 2019-05-14
Event API: Add responder allowMultipleHostChildren flag

trueadm:
This PR adds the `allowMultipleHostChildren` option to event responder modules. This allows for DEV time validation of event components and their respective host component children (if they have any). This is required for `Press`, `Hover` and other core event modules where there is a hard constraint on the amount of host components a responder supports.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -851,7 +851,7 @@ function completeWork(
             }
             warning(
               hostChildrenCount < 2,
-              'An event component "%s" was found to have multiple host children.',
+              'A "<%s>" event component cannot contain multiple host children.',
               getComponentName(workInProgress.type),
             );
           }
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -97,6 +97,7 @@ import {
 import {markRenderEventTime, renderDidSuspend} from './ReactFiberScheduler';
 import {getEventComponentHostChildrenCount} from './ReactFiberEvents';
 import getComponentName from 'shared/getComponentName';
+import warning from 'shared/warning';
 
 function markUpdate(workInProgress: Fiber) {
   // Tag the fiber with an update effect. This turns a Placement into
@@ -813,11 +814,11 @@ function completeWork(
 
         if (eventComponentInstance === null) {
           let responderState = null;
-          if (!responder.allowMultipleHostChildren) {
+          if (__DEV__ && !responder.allowMultipleHostChildren) {
             const hostChildrenCount = getEventComponentHostChildrenCount(
               workInProgress,
             );
-            invariant(
+            warning(
               hostChildrenCount < 2,
               'A "<%s>" event component cannot contain multiple host children.',
               getComponentName(workInProgress.type),
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -827,6 +827,12 @@ function completeWork(
                   return;
                 }
                 node = fallbackNode;
+              } else if (node.tag === SuspenseComponent) {
+                const fallbackNode = ((node.child: any): Fiber).child;
+                if (fallbackNode === null) {
+                  return;
+                }
+                node = fallbackNode;
               }
               if (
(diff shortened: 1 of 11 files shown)


packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -811,6 +813,46 @@ function completeWork(
...
+              if (
+                node.tag === HostComponent ||
+                node.tag === HostText ||
+                node.tag === HostPortal
+              ) {
+                hostChildrenCount++;
+              } else {
+                const child = node.child;
+                if (child !== null) {
+                  node = child;
```
gaearon: It is not generally safe to step into a child and then follow the return pointer back because it may point to a different alternate. When we do that, we usually mutate the `return` pointer so we can get back in the right tree. Same with `sibling`.

An alternative to mutating could be to use recursion (since it's dev-only).

trueadm: This is in the complete phase though, so it shouldn't be unsafe right? You're right about recursion though, I'll change to that as it's also much easier to read.

gaearon:
What is your intention for Suspense nodes? Suspense has both primary and fallback tree, but only one is visible at a time. If both contain a single div, I think this would give a false positive. Would you want to handle that?

trueadm:
@gaearon Can you explain more what you mean by the false positive? Maybe I'm wrong here, but given that only one tree is visible at a time, the child fibers of the suspense nodes would always only point to the visible fibers at the complete phase? The logic in the function of this PR should only check the visible tree.

gaearon:
@trueadm 

>Maybe I'm wrong here, but given that only one tree is visible at a time, the child fibers of the suspense nodes would always only point to the visible fibers at the complete phase?

No, this is not guaranteed. If we previously had a real tree but then we suspended, we keep _both_ of them as children. The main tree gets hidden via CSS but both exist in the DOM too.

What should the behavior be then?

trueadm:
@gaearon Let's bring this up in today's sync.

trueadm:
I've updated the PR and added the logic to handle that case. Thanks for the help :)

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -811,6 +817,44 @@ function completeWork(
...
+              if (sibling !== null) {
+                findAllHostChildren(sibling);
+              }
+            };
+            if (workInProgress.child !== null) {
+              findAllHostChildren(workInProgress.child);
+            }
+            warning(
+              hostChildrenCount < 2,
+              'An event component "%s" was found to have multiple host children.',
```
necolas: Little detail. I think the message implies that all event components must have a single child. Could be phrased something like…

```suggestion
              'A "<%s>" event component cannot contain multiple host children.',
```

I think the UX consequences could be severe enough to cause a SEV if ignored. Should we throw an error like `React.Children.only` does?

trueadm: An error means that we have to move this out of the DEV condition. I wanted to avoid this as this check is expensive.

necolas: It can't be a DEV-only error?

trueadm: This is too expensive to do in prod IMO.

trueadm: I guess I can move back to the older version that didn't use recursion and make it an invariant, it just feels like a relatively expensive check to do on every fiber complete phase – especially as the tree below is likely to be quite large before the first host component is known.

trueadm:
Sure. I can change it back. The current version definitely does update return punters however.

packages/react-reconciler/src/ReactFiberEvents.js
```diff
@@ -0,0 +1,84 @@
...
+
+export function getSuspenseFiberFromTimedOutChild(fiber: Fiber): Fiber {
+  return ((((fiber.return: any): Fiber).return: any): Fiber);
+}
+
+export function getEventComponentHostChildrenCount(
+  eventComponentFiber: Fiber,
+): number {
+  let hostChildrenCount = 0;
+
```
gaearon: All of these are DEV-only. Can we make it explicit by wrapping their content in `__DEV__`? And maybe `InDEV` suffixes. So that we don't accidentally try to use them elsewhere.

Issue #14750, opened 2019-02-03
Simultaneous key events in effect handled out of order


stuartkeith:
**Do you want to request a *feature* or report a *bug*?**

Report a bug.

**What is the current behavior?**

I have an app that's registering event listeners for `window`'s key events (via `useEffect`). Those event listeners are triggering state updates (via `useState`). I think I have found a bug where simultaneous key events occurring in the same frame (whether down or up) will be handled out of order, causing state to becoming out of sync.

Take the following simple app (https://codesandbox.io/s/1z3v9zrk4j). I've kept this as keyup only for simplicity.

```
function App() {
  const [keys, setKeys] = useState([]);

  console.log('App', keys);

  const onKeyUp = function (event) {
    console.log('onKeyUp', event.key, keys);

    setKeys([...keys, event.key]);
  };

  useEffect(function () {
    console.log('effect', keys);

    window.addEventListener('keyup', onKeyUp);

    return function () {
      console.log('removing event listener', keys);

      window.removeEventListener('keyup', onKeyUp);
    };
  });

  return <p>{keys.join(', ')}</p>;
}
```

If I press down any two keys, e.g. the "q" and "w" keys, and then release them at precisely the same time, the following happens:

- The `keyup` event listener for `w` is called, which in turn calls `setKeys` with `['w']`
- `App` is re-rendered with `keys === ['w']`
- The `keyup` event listener for `q` is called, which in turn calls `setKeys` with `['q']`
- The effect's cleanup function is called, removing the event listener with `keys === []`
- The effect is run again, the event listener being added with `keys === ['w']`
- `App` is re-rendered with `keys === ['q']`
- The effect's cleanup function is called, removing the event listener with `keys ===['w']`
- The effect is run again, the event listener being added with `keys === ['q']`

This results in `keys === ['q']`. The render with `w` has been lost.

With three keys, only two keys are reliably shown. Four keys - only two are reliably shown.

If I add another `useState` call, the first `useState` has no issues - all keys are reliably detected. See https://codesandbox.io/s/0yo51n5wv:

```
function App() {
  const [keys, setKeys] = useState([]); 
  const [dummy, setDummy] = useState('foo');

  console.log("rendering App", keys);

  const onKeyUp = function(event) {
    console.log("onKeyUp event received", event.key, keys);

    setKeys([...keys, event.key]);
    setDummy('foo');
  };

  useEffect(function() {
    console.log("adding event listener", keys);

    window.addEventListener("keyup", onKeyUp);

[...]

PR #15650, opened 2019-05-14
Fix serial passive effects

acdlite:
Flush passive effects before discrete events. Currently, we check for pending passive effects inside the `setState` method before we add additional updates to the queue, in case those pending effects also add things to the queue.

However, the `setState` method is too late, because the event that caused the update might not have ever fired had the passive effects flushed before we got there.

This is the same as the discrete/serial events problem. When a serial update comes in, and there's already a pending serial update, we have to do it before we call the user-provided event handlers. Because the event handlers themselves might change as a result of the pending update.

This commit moves the `flushPassiveEffects` call to before the discrete event handlers are called, and removes it from the `setState` method. Non-discrete events will not cause passive effects to flush, which is fine, since by definition they are not order dependent.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMHooks-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMHooks-test.js
@@ -72,42 +72,6 @@ describe('ReactDOMHooks', () => {
     expect(container3.textContent).toBe('6');
   });
 
-  it('can batch synchronous work inside effects with other work', () => {
-    let otherContainer = document.createElement('div');
-
-    let calledA = false;
-    function A() {
-      calledA = true;
-      return 'A';
-    }
-
-    let calledB = false;
-    function B() {
-      calledB = true;
-      return 'B';
-    }
-
-    let _set;
-    function Foo() {
-      _set = React.useState(0)[1];
-      React.useEffect(() => {
-        ReactDOM.render(<A />, otherContainer);
-      });
-      return null;
-    }
-
-    ReactDOM.render(<Foo />, container);
-    ReactDOM.unstable_batchedUpdates(() => {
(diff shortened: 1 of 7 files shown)


acdlite:
@threepointone This is a duplicate of https://github.com/facebook/react/pull/15644 that I started last week but never opened a PR for. I don't care which one we merge as long as you incorporate the semantic changes and tests in this one.

acdlite:
I think the only difference  between our PRs is that I removed the `flushPassiveEffects` call from `setState`, for the reason described above.

gaearon:
If you need to treat setState as serial (eg from keyboard window event) how do you opt in? Are making interactiveUpdates a public API? Do we rename it?

acdlite:
@gaearon In a real test I would simulate a click event. Since ReactNoop doesn't have an event system, I used `interactiveUpdates`. I don't consider `interactiveUpdates` a public API but eventually we will need some way to create a custom discrete event handler. I think it will be two functions instead of one: a function to wrap the discrete event handler, and a function to flush pending discrete updates.

gaearon:
I guess I’m asking what’s the upgrade path for people who wanted to rely on serial behavior and can’t anymore with this change. Shouldn’t that upgrade path be a part of the same release?

acdlite:
We already lack a first class way to create custom serial event handlers (e.g. key events in  https://github.com/facebook/react/issues/14750). You can get most of the way there with event delegation: attach a single native event listener and wrap all the handlers in `batchedUpdates` or `flushSync`.

You'd still need a way to force flush passive effects. We could expose `unstable_flushPassiveEffects` but I'm not confident enough to make that stable.

Regardless, that problem exists independently of the problem fixed by this PR (where `flushPassiveEffects` gets called after the event handler has already started).

PR #15651, opened 2019-05-14
Event API: ensure getFocusableElementsInScope handles suspended trees

trueadm:
This PR fixes a bug in `getFocusableElementsInScope` where suspense trees weren't correctly handled during fiber tree traversal. With these changes, `getFocusableElementsInScope` now properly collects focusable elements in regards to Suspense fallbacks.


necolas:
Nice!

PR #15658, opened 2019-05-15
Track which fibers scheduled the current render work

bvaughn:
This PR introduces the concept of tracking which Fiber(s) schedule work with React.

Tracked Fibers are called "updaters" and are exposed to DevTools via a `memoizedUpdaters` property on the `ReactFiberRoot`. The implementation of this feature follows a vaguely similar approach as interaction tracing, but does not require reference counting since there is no subscriptions API.

This change is in support of a new DevTools Profiler feature that shows which Fiber(s) scheduled the selected commit in the Profiler.

All changes have been gated behind a new feature flag, `enableUpdaterTracking`, which is enabled for Profiling builds by default. We also only track updaters when DevTools has been detected, to avoid doing unnecessary work.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -37,6 +37,7 @@ import {
   enableStrictEffects,
   deletedTreeCleanUpLevel,
   enableSuspenseLayoutEffectSemantics,
+  enableUpdaterTracking,
 } from 'shared/ReactFeatureFlags';
 import {
   FunctionComponent,
@@ -89,7 +90,7 @@ import {
   resetCurrentFiber as resetCurrentDebugFiberInDEV,
   setCurrentFiber as setCurrentDebugFiberInDEV,
 } from './ReactCurrentFiber';
-
+import {isDevToolsPresent} from './ReactFiberDevToolsHook.new';
 import {onCommitUnmount} from './ReactFiberDevToolsHook.new';
 import {resolveDefaultProps} from './ReactFiberLazyComponent.new';
 import {
@@ -137,6 +138,7 @@ import {
   resolveRetryWakeable,
   markCommitTimeOfFallback,
   enqueuePendingPassiveProfilerEffect,
+  restorePendingUpdaters,
 } from './ReactFiberWorkLoop.new';
 import {
   NoFlags as NoHookEffect,
@@ -161,6 +163,10 @@ let offscreenSubtreeWasHidden: boolean = false;
 const PossiblyWeakSet = typeof WeakSet === 'function' ? WeakSet : Set;
 
 let nextEffect: Fiber | null = null;
+
+// Used for Profiling builds to track updaters.
+let inProgressLanes: Lanes | null = null;
+let inProgressRoot: FiberRoot | null = null;
 
 const callComponentWillUnmountWithTimer = function(current, instance) {
   instance.props = current.memoizedProps;
@@ -2094,6 +2100,20 @@ function attachSuspenseRetryListeners(finishedWork: Fiber) {
           }
         }
         retryCache.add(wakeable);
+
+        if (enableUpdaterTracking) {
+          if (isDevToolsPresent) {
+            if (inProgressLanes !== null && inProgressRoot !== null) {
+              // If we have pending work still, associate the original updaters with it.
+              restorePendingUpdaters(inProgressRoot, inProgressLanes);
+            } else {
+              throw Error(
+                'Expected finished root and lanes to be set. This is a bug in React.',
+              );
+            }
+          }
+        }
+
         wakeable.then(retry, retry);
       }
     });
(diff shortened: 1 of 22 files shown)


bvaughn:
Hm. I think there's a gap here with flushing pending updates that can cause the updater to get lost. Haven't been able to catch it in a test yet.

bvaughn:
Okay. I think I captured a variant of the bug in a new test (failing before) and fixed.

bvaughn:
Rebased after #15664 landed.

acdlite:
I'm going to defer to @sebmarkbage to review this since I think you discussed the overall approach with him already

bvaughn:
Lucky @sebmarkbage 🎉

bvaughn:
Ping @sebmarkbage 😄

bvaughn:
Ping @sebmarkbage again 👋

bvaughn:
I think this PR is still worth at least considering. Leaving open for now until @sebmarkbage indicates level of interest. (If interested, I'll rebase.)

bvaughn:
This would have been a useful feature 😄  It's unfortunate that it never got reviewed.

gaearon:
I can review it! I don't think I've seen it before. I doubt Seb sees GH mentions. :)

bvaughn:
> I don't think I've seen it before.

Screen Shot 2021-03-29 at 10 24 26 PM

🤣  At least I'm not the only one who does this!

Thanks for the offer Dan! I'll clean it up sometime this week and ping you.

gaearon:
It’s been... a while.

bvaughn:
Well fortunately, only....everything...has changed since I wrote this. Should not be a problem.

bvaughn:
I'm returning this to draft status while I work on rebasing and updating the expiration times logic to lanes logic.

packages/react-reconciler/src/ReactFiberLane.new.js
```diff
@@ -742,6 +747,53 @@ export function getBumpedLaneForHydration(
...
+          });
+          updaters.clear();
+        }
+
+        lanes &= ~lane;
+      }
+    }
+  }
+}
+
```
bvaughn, line 796: It seemed best to put these functions in `ReactFiberLane` so they had access to `laneToIndex` to make lanes->indices mapping easier.

bvaughn:
Okay I think it's ready for a look 😄 

(I assume I'l have to add some gating to my test once CI shows me which builds fail but otherwise...)

sebmarkbage:
Didn't see pings on GH.

I think I must have forgotten a lot of the context here.

But as I read it, this is really mostly because we want to track it through retries of Suspense and Offscreen. Is that right?

Because otherwise we can infer this info in the devtools when we loop over the tree after a commit. If it's just for that commit.

This is similar to how the Cache has to attach itself to a Suspense boundary.

I wonder if this would be more scoped if we did it that way. We could store the parents that participated in an update on an internal context and then snapshot that on the Suspense/Offscreen/Hydrating boundary.

Then devtools would pick up this info in the traversal of the commit either from the updated hook or the suspense boundary that was retries.

There's a subtle difference also if you create this list when something is called with setState on it vs if it participated in the render. Because things that called setState but that thing was deleted or in a hidden subtree didn't contribute to the rendering (just the cloning) of that commit. Not sure which one is better.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -2061,7 +2094,11 @@ function commitMutationEffects_complete(root: FiberRoot) {
   }
 }
 
-function commitMutationEffectsOnFiber(finishedWork: Fiber, root: FiberRoot) {
+function commitMutationEffectsOnFiber(
+  finishedWork: Fiber,
+  root: FiberRoot,
+  committedLanes: Lanes,
```
sebmarkbage, line 2100: It would be better to store this on a module variable so we don't have to pay the cost of the extra writing to the stack in prod.

bvaughn:
> But as I read it, this is really mostly because we want to track it through retries of Suspense and Offscreen. Is that right?

Yes.

> I wonder if this would be more scoped if we did it that way. We could store the parents that participated in an update on an internal context and then snapshot that on the Suspense/Offscreen/Hydrating boundary.
>
> Then devtools would pick up this info in the traversal of the commit either from the updated hook or the suspense boundary that was retries.

Some initial concern at the idea of maintaining that over time, across React versions. Seems like the internal logic of Suspense (and Offscreen) has already changed a lot and will likely be changing more (e.g. with effects semantics) – something explicit feels more future proof.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
bvaughn: Sure.

Seems a little arbitrary, since we're already passing around at least the root to most of these functions, but I'll make the minimal change for now and use module scope for the root and lanes.

bvaughn: Done.

bvaughn:
CI failure is unrelated to code changes:
> Error response from daemon: Get https://registry-1.docker.io/v2/circleci/openjdk/manifests/8-jdk-node-browsers: Get https://auth.docker.io/token?account=circlecidockerdefaultreadonly&scope=repository%3Acircleci%2Fopenjdk%3Apull&service=registry.docker.io: EOF

bvaughn:
Can I get a review of this please? :)

sebmarkbage (approved):
We can discuss alternative approaches for a longer term solution but don't want to block the UI features.

bvaughn:
Much appreciated!

PR #15659, opened 2019-05-15
Feature flag to revert #15650

acdlite:
PR #15650 is a bugfix but it's technically a semantic change that could cause regressions. I don't think it will be an issue, since the previous behavior was both broken and incoherent, but out of an abundance of caution, let's wrap it in a flag so we can easily revert it if necessary.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberClassComponent.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.js
@@ -52,7 +52,9 @@ import {
   requestCurrentTime,
   computeExpirationForFiber,
   scheduleWork,
+  flushPassiveEffects,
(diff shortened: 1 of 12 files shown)


packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -193,6 +194,9 @@ const classComponentUpdater = {
       update.callback = callback;
     }
 
+    if (revertPassiveEffectsChange) {
```
threepointone: did you forget to import this flag in this module? failing tests for the same reason.

acdlite: Oops, yes

PR #15664, opened 2019-05-15
Inform DevTools of commit priority level

bvaughn:
React now informs DevTools of the commit priority level (so the Profiler can display and potentially filter on the info).

### Demo
priority-Kapture 2019-05-15 at 15 21 57

### DIFF
--- a/packages/react-reconciler/src/ReactFiberDevToolsHook.js
+++ b/packages/react-reconciler/src/ReactFiberDevToolsHook.js
@@ -7,8 +7,12 @@
  * @flow
  */
 
+import {requestCurrentTime} from './ReactFiberScheduler';
+import {inferPriorityFromExpirationTime} from './ReactFiberExpirationTime';
+
 import type {Fiber} from './ReactFiber';
 import type {FiberRoot} from './ReactFiberRoot';
+import type {ExpirationTime} from './ReactFiberExpirationTime';
 
 import warningWithoutStack from 'shared/warningWithoutStack';
 
@@ -17,23 +21,6 @@ declare var __REACT_DEVTOOLS_GLOBAL_HOOK__: Object | void;
 let onCommitFiberRoot = null;
 let onCommitFiberUnmount = null;
 let hasLoggedError = false;
-
-function catchErrors(fn) {
-  return function(arg) {
-    try {
-      return fn(arg);
-    } catch (err) {
-      if (__DEV__ && !hasLoggedError) {
-        hasLoggedError = true;
-        warningWithoutStack(
-          false,
-          'React DevTools encountered an error: %s',
-          err,
-        );
-      }
-    }
-  };
-}
 
 export const isDevToolsPresent =
   typeof __REACT_DEVTOOLS_GLOBAL_HOOK__ !== 'undefined';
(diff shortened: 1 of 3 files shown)


bvaughn:
🤔 Looks like importing `ReactFiberScheduler` from `ReactFiberDevToolsHook` causes `ReactIncrementalErrorReplay-test.internal` to fail with an invariant I don't understand yet:

> Invariant Violation: This module must be shimmed by a specific renderer.

packages/react-reconciler/src/ReactFiberDevToolsHook.js
```diff
@@ -65,12 +52,39 @@ export function injectInternals(internals: Object): boolean {
...
-      hook.onCommitFiberRoot(rendererID, root),
-    );
-    onCommitFiberUnmount = catchErrors(fiber =>
-      hook.onCommitFiberUnmount(rendererID, fiber),
-    );
+    onCommitFiberRoot = (root, expirationTime) => {
+      const currentTime = requestCurrentTime();
+      const priorityLevel = inferPriorityFromExpirationTime(
+        currentTime,
+        expirationTime,
```
gaearon: Should these be DEV/PROFILE only?

bvaughn: Interesting. Yeah probably wouldn't hurt!

packages/react-reconciler/src/ReactFiberDevToolsHook.js
```diff
@@ -65,12 +53,43 @@ export function injectInternals(internals: Object): boolean {
...
-      hook.onCommitFiberUnmount(rendererID, fiber),
-    );
+    onCommitFiberRoot = (root, expirationTime) => {
+      try {
+        if (enableProfilerTimer) {
+          const currentTime = requestCurrentTime();
+          const priorityLevel = inferPriorityFromExpirationTime(
+            currentTime,
+            expirationTime,
+          );
```
bvaughn: Andrew mentioned maybe we could use [`getCurrentPriorityLevel()`](https://github.com/facebook/react/blob/9c6de716d028f17736d0892d8a3d8f3ac2cb62bd/packages/react-reconciler/src/SchedulerWithReactIntegration.js#L87-L102) here rather than inferring the priority level. I don't _think_ the priority level a Fiber was scheduled with will necessarily match the current priority level though. (Maybe I misunderstood his suggestion.)

packages/react-reconciler/src/__tests__/ReactIncrementalErrorReplay-test.internal.js
```diff
@@ -11,6 +11,10 @@
 
 describe('ReactIncrementalErrorReplay-test', () => {
   it('copies all keys when stashing potentially failing work', () => {
+    // Include a renderer to ensure host config files are properly shimmed.
+    // Otherwise transient imports may cause an invariant.
+    require('react-dom');
```
gaearon: I landed another PR which needed this fix so you should be able to revert this file

bvaughn: Oh, cool thanks.

packages/react-reconciler/src/ReactFiberDevToolsHook.js
acdlite: If you call `getCurrentPriorityLevel` inside `commitRoot` (not `commitRootImpl`, since that's always run at `ImmediatePriority` — it has to be the wrapper) then it will match the priority of the commit, which is not necessarily the same thing as the priority at which something was scheduled because multiple priorities can get batched.

This approach seems fine, though.

It's possible we may eventually track the priority on the update object.

PR #15666, opened 2019-05-16
Bail out of updates in offscreen trees

gaearon:
Currently `setState` inside `flushSync` will flush within an offscreen tree. This is because our offscreen bailout is buried into `updateHostComponent` which doesn't run in the jump-over bailout codepath as we go towards the deep update. This duplicates the check there. The new test fails on master. (I'm not sure I understand the change needed in another test so this needs some caution.)

I've also flipped the condition in both places to check the mode first. Since this is likely a more common first check that will be false.

Note this adds some overhead in CM to the jump-over bailout codepath. Maybe this is the time to add `<Offscreen />` instead?

### DIFF
--- a/packages/react-dom/src/__tests__/ReactUpdates-test.js
+++ b/packages/react-dom/src/__tests__/ReactUpdates-test.js
@@ -1323,18 +1323,27 @@ describe('ReactUpdates', () => {
 
     const root = ReactDOM.unstable_createRoot(container);
     root.render(<Foo />);
-    expect(Scheduler).toFlushAndYieldThrough(
-      ['Foo', __DEV__ && 'Foo', 'Baz', 'Foo#effect'].filter(Boolean),
-    );
+    if (__DEV__) {
+      expect(Scheduler).toFlushAndYieldThrough([
+        'Foo',
+        'Foo',
+        'Baz',
+        'Foo#effect',
+      ]);
+    } else {
+      expect(Scheduler).toFlushAndYieldThrough(['Foo', 'Baz', 'Foo#effect']);
+    }
 
     const hiddenDiv = container.firstChild.firstChild;
     expect(hiddenDiv.hidden).toBe(true);
     expect(hiddenDiv.innerHTML).toBe('');
 
     // Run offscreen update
-    expect(Scheduler).toFlushAndYield(
-      ['Bar', __DEV__ && 'Bar'].filter(Boolean),
-    );
+    if (__DEV__) {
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/__tests__/ReactUpdates-test.js
```diff
@@ -1290,6 +1290,67 @@ describe('ReactUpdates', () => {
...
+          </div>
+          <Baz />
+        </div>
+      );
+    }
+
+    const root = ReactDOM.unstable_createRoot(container);
+    root.render(<Foo />);
+    expect(Scheduler).toFlushAndYieldThrough(
+      ['Foo', __DEV__ && 'Foo', 'Baz', 'Foo#effect'].filter(Boolean),
```
acdlite: 😐

OK

🙂

gaearon: Better than making the test internal only just to flip this flag.

gaearon: (We should figure out something even better though. A bunch of tests don't run on bundles jut because of this inconvenience.)

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2132,6 +2132,15 @@ function beginWork(
...
+          if (
+            workInProgress.mode & ConcurrentMode &&
+            renderExpirationTime !== Never &&
+            shouldDeprioritizeSubtree(workInProgress.type, newProps)
+          ) {
+            // Schedule this fiber to re-render at offscreen priority. Then bailout.
+            workInProgress.expirationTime = workInProgress.childExpirationTime = Never;
+            return null;
+          }
           pushHostContext(workInProgress);
```
acdlite: Oh wait, I think this context push needs to go above the return

packages/react-dom/src/__tests__/ReactUpdates-test.js
acdlite: I would have done

```js
if (__DEV__) {
  expect(Scheduler).toFlushAndYieldThrough(['Foo', 'Foo', 'Baz', 'Foo#effect']);
} else {
  expect(Scheduler).toFlushAndYieldThrough(['Foo', 'Baz', 'Foo#effect']);
}
```

packages/react-reconciler/src/ReactFiberBeginWork.js
gaearon: fixd

PR #15671, opened 2019-05-16
Event API: ensure calculateResponderRegion accounts for page offset

trueadm:
This PR fixes issues where `Press` was not getting fired correctly due to co-ords of the responder not correctly accounting for the fact `getBoundingClientRect`'s position is relative to the viewport, not the page. This adds the page offsets to the calculations to ensure consistency when scrolling etc.

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -428,6 +428,25 @@ function calculateDelayMS(delay: ?number, min = 0, fallback = 0) {
   return Math.max(min, maybeNumber != null ? maybeNumber : fallback);
 }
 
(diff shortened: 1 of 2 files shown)


packages/react-events/src/Press.js
```diff
@@ -441,11 +441,13 @@ function calculateResponderRegion(
   );
 
   const clientRect = target.getBoundingClientRect();
+  const viewportOffsetX = window.pageXOffset;
```
necolas: Is the result still correct if the element is position fixed at the top of the viewpoint?

trueadm: I've changed this since.

PR #15677, opened 2019-05-17
Event API: ensure simulated mouse events are not propagated

trueadm:
When dealing with touch events, browsers "simulate" mouse events from them. This can be problematic for the experimental event API, as it means that when we've dispatched a "Press" during the "pointerup" event, a later "mouseup" or "click" event can cause side-effects that would never have happened by using "click" (as it's the last event). I've allowed "click" for now, only so that this doesn't affect tracking tooling that relies upon "click" working for touch events too; so only "mousedown" and "mouseup" and blocked when using Press.

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -117,6 +117,7 @@ const targetEventTypes = [
   // We need to preventDefault on pointerdown for mouse/pen events
   // that are in hit target area but not the element area.
   {name: 'pointerdown', passive: false},
+  'mousedown',
 ];
 const rootEventTypes = [
   'keyup',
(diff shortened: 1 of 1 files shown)


packages/react-events/src/Press.js
```diff
@@ -125,17 +126,19 @@ const rootEventTypes = [
   'scroll',
   'pointercancel',
 ];
+const baseRootEventTypes = [
+  // Used as a 'cancel' signal for mouse interactions
```
necolas: Either this comment should exist above `dragstart` below, or `dragstart` should be removed if it's not used anymore.

packages/react-events/src/Press.js
```diff
@@ -117,6 +117,7 @@ const targetEventTypes = [
   // We need to preventDefault on pointerdown for mouse/pen events
   // that are in hit target area but not the element area.
   {name: 'pointerdown', passive: false},
+  'mousedown',
```
necolas: Let's add a comment explaining why there's a `mousedown` dependency here.

packages/react-events/src/Press.js
```diff
@@ -125,17 +126,19 @@ const rootEventTypes = [
   'scroll',
   'pointercancel',
 ];
+const baseRootEventTypes = [
+  // Used as a 'cancel' signal for mouse interactions
+  {name: 'mouseup', passive: false},
```
necolas: Let's add a comment explaining why there's a `mouseup` root dependency here.

trueadm:
I think I'll go about this another way. React Flare listens to the document at a capture layer, which means that we may be blocking other third-party event tooling.

PR #15679, opened 2019-05-19
Fix issue with Tab+alt not being considered as isGlobalFocusVisible c…

Andarist:
cc @trueadm @necolas

### DIFF
--- a/packages/react-events/src/Focus.js
+++ b/packages/react-events/src/Focus.js
@@ -39,6 +39,8 @@ type FocusEvent = {|
   timeStamp: number,
 |};
 
+const isMac = /^Mac/.test(navigator.platform);
+
 const targetEventTypes = [
   {name: 'focus', passive: true},
   {name: 'blur', passive: true},
@@ -304,7 +306,11 @@ const FocusResponder = {
         const nativeEvent = event.nativeEvent;
         if (
           nativeEvent.key === 'Tab' &&
-          !(nativeEvent.metaKey || nativeEvent.altKey || nativeEvent.ctrlKey)
+          !(
+            nativeEvent.metaKey ||
+            (!isMac && nativeEvent.altKey) ||
+            nativeEvent.ctrlKey
+          )
         ) {
           state.pointerType = 'keyboard';
           isGlobalFocusVisible = true;
--- a/packages/react-events/src/Focus.js
+++ b/packages/react-events/src/Focus.js
@@ -39,7 +39,10 @@ type FocusEvent = {|
   timeStamp: number,
 |};
 
-const isMac = /^Mac/.test(navigator.platform);
+const isMac =
+  typeof window !== 'undefined' && window.navigator != null
+    ? /^Mac/.test(window.navigator.platform)
+    : false;
 
 const targetEventTypes = [
   {name: 'focus', passive: true},
--- a/packages/react-events/src/__tests__/Focus-test.internal.js
+++ b/packages/react-events/src/__tests__/Focus-test.internal.js
@@ -39,16 +39,20 @@ const createPointerEvent = (type, data) => {
   return event;
 };
 
+const modulesInit = () => {
+  ReactFeatureFlags = require('shared/ReactFeatureFlags');
+  ReactFeatureFlags.enableEventAPI = true;
+  React = require('react');
+  ReactDOM = require('react-dom');
+  Focus = require('react-events/focus');
+};
+
 describe('Focus event responder', () => {
   let container;
 
   beforeEach(() => {
     jest.resetModules();
-    ReactFeatureFlags = require('shared/ReactFeatureFlags');
-    ReactFeatureFlags.enableEventAPI = true;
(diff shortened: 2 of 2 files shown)


necolas:
What's this for please?

Andarist:
It's for focus type tracking - to be precise to determine which pointerType has originated focus. Without distinguishing this check (altKey) per platform you incorrectly lose info about Tab+altKey triggering focus on Mac. 

I could understand something in the wrong way, but according to my quick tests it would be an issue.

trueadm:
Could you add some tests please? :)

necolas:
Please can you also provide adequate context for this patch. For example, explaining why this is applied only to Mac, i.e., what's the difference in expected behaviour on Windows and Linux.

Andarist:
> Could you add some tests please? :)

Sure thing, gonna add them in following days.

> Please can you also provide adequate context for this patch. For example, explaining why this is applied only to Mac, i.e., what's the difference in expected behaviour on Windows and Linux.

Sorry that I haven't explained it properly right away! 

I've encountered this problem last week while implementing custom radio/checkboxes, our QA has reported that focusing with Tab doesn't work properly on Safari for those (radio/checkbox). To my big surprise I have found out that Safari (follow Mac behaviour) doesn't handle this like other browsers.

When pressing Tab alone it just jumps to the next text-like input (and probably some lists too), not to next focusable element. To jump to the next focusable element you have to press Tab+alt. I've checked it on the system level and I've checked out all major browsers on Mac - they all handle Tab OK (except Safari) but they also allow for Tab+alt (like the host system - Mac).

Seems like this behaviour can be configured on Mac, but the default is what I have described. Take a look at the bottom of the keyboard's Shortcus section of the system preferences:
Screen Shot 2019-05-21 at 00 15 56

I've checked out at home how exactly it behaves on Windows on my wife's computer - and each additional key with Tab triggers different action, so the condition as of before this PR was OK for Windows - I yet have to check Linux, but I expect it to behave exactly the same as Windows in this regard.

Andarist:
I've added a test for this particular scenario. I wanted to add an "opposite" test - for non-Mac - that it doesn't set this property etc, but because tests are relying on the manual event dispatching this is tricky. 

I'm not particularly fond of how I have added this test (it results in quite a duplication of the setup/teardown code), but I've tried bunch of possibilities and in the end I've decided this is the most straightforward one. Can't add a conditional setup to a single test in the middle of the setup chain. 

If you want me to refactor this anyhow - let me know.

packages/react-events/src/__tests__/Focus-test.internal.js
```diff
@@ -327,3 +327,65 @@ describe('Focus event responder', () => {
     expect(Focus.displayName).toBe('Focus');
   });
 });
+
+describe('Focus event responder - Mac', () => {
```
necolas: You shouldn't need to create another top-level `describe`, setup the feature flags, etc. Put the test into the relevant `onFocus` test block that already exists above.

Andarist: Totally agree, I wasn't sure how to approach mocking navigator for this single test. The problem is that setup for each test is done in beforeEach and ideally I would like to mock this BEFORE common beforeEach, but this is not quite possible - or I don't know how to do it, but after searching through jest documentation I couldn't find a way to do this. Any ideas?

necolas: Oh because you need the `navigator` mocked before `Focus` is evaluated? Hmm, not sure. I'll look into it

Andarist: yeah, that's exactly the "problem" with putting this into existing describe block

necolas: Have a look at how it's being done here: https://github.com/facebook/react/pull/15761. Once you reuse that pattern for this PR we can merge it. Thanks!

packages/react-events/src/Focus.js
```diff
@@ -39,6 +39,8 @@ type FocusEvent = {|
   timeStamp: number,
 |};
 
+const isMac = /^Mac/.test(navigator.platform);
```
necolas: Needs to be safe to evaluate this on the server: https://github.com/facebook/react/blob/0f7cc2b/packages/react-events/src/Press.js#L104-L107

packages/react-events/src/__tests__/Focus-test.internal.js
necolas: See https://github.com/facebook/react/blob/0f7cc2ba845e34d7a312819e27bbfa88cdb7d49f/packages/react-events/src/__tests__/Press-test.internal.js#L39-L46

Andarist: Thanks for the tip! Done.

Andarist:
@necolas should be ready to merge now (if CI passes), thanks for the help :)

PR #15681, opened 2019-05-20
[Fresh] Support lazy() and add Suspense tests

gaearon:
This resolves a TODO about supporting `lazy`.

For now this only lets you edit a file *wrapped with* `lazy` — but not the file *using* `lazy`. That would still need some follow-up work.

Also adds test coverage for hot reloading Suspended trees.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
@@ -895,6 +895,597 @@ describe('ReactFresh', () => {
     }
   });
 
+  it('can preserve state for lazy after resolution', async () => {
+    if (__DEV__) {
+      let promise;
+      let OuterV1 = render(() => {
+        function Hello() {
+          const [val, setVal] = React.useState(0);
+          return (
+            <p style={{color: 'blue'}} onClick={() => setVal(val + 1)}>
+              {val}
+            </p>
+          );
+        }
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/__tests__/ReactFresh-test.internal.js
```diff
@@ -895,6 +895,597 @@ describe('ReactFresh', () => {
...
+            <React.Suspense fallback={<p>Loading</p>}>
+              <Outer />
+            </React.Suspense>
+          );
+        }
+
+        return App;
+      });
+
+      expect(container.textContent).toBe('Loading');
```
threepointone: why not use act() for this?
(and a couple more further down)

gaearon: Use how?

gaearon: I thought act doesn't flush Suspense. I need recipes :P

threepointone: replace 
```jsx
jest.runAllTimers();
await promise;
Scheduler.flushAll();
```

with 
```jsx
await act(async () => {
  jest.runAllTimers()
})
```

in general, we shouldn't have to call `Scheduler.flushAll()` by hand

PR #15682, opened 2019-05-20
flush only on exiting outermost act()

threepointone:
^ like it says. we flush only on exiting the outermost act().

my previous PR (https://github.com/facebook/react/pull/15519) ended up in merge conflict hell, so I just redid it.

### DIFF
--- a/packages/react-dom/src/test-utils/ReactTestUtilsAct.js
+++ b/packages/react-dom/src/test-utils/ReactTestUtilsAct.js
@@ -84,16 +84,15 @@ function flushWorkAndMicroTasks(onDone: (err: ?Error) => void) {
 let actingUpdatesScopeDepth = 0;
 
 function act(callback: () => Thenable) {
-  let previousActingUpdatesScopeDepth;
+  let previousActingUpdatesScopeDepth = actingUpdatesScopeDepth;
+  actingUpdatesScopeDepth++;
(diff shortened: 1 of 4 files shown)


acdlite:
> my previous PR (#15519) ended up in merge conflict hell, so I just redid it.

Thanks! That makes it much easier for me to review, too. (Personally I find rebasing on master makes it easier to maintain stacked PRs, compared to merging, but if you prefer merging keep doing that.)

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
```diff
@@ -171,7 +177,11 @@ function act(callback: () => Thenable) {
 
     // flush effects until none remain, and cleanup
     try {
-      flushWork();
+      if (actingUpdatesScopeDepth === 1) {
```
acdlite: Nit:
```suggestion
      if (actingUpdatesScopeDepth <= 1) {
```

threepointone: really? it shouldn't do that if it's 0 (and it ideally should never be 0 here). going to push back on this.

threepointone: tbh, it should probably throw an error if it ever gets there

PR #15687, opened 2019-05-20
[Flare] Update interactiveUpdates flushing heuristics

trueadm:
This PR changes the logic that occurs in `interactiveUpdates`, to account for a scenario that proved problematic in the testing of React Flare – the mixing of both event systems on the same app. Note: this change is behind the experimental event API. Also included in this PR are:

- `interactiveUpdates` -> `discreteUpdates`
- `flushInteractiveUpdates` -> `flushDiscreteUpdates`
- Adds `batchedEventUpdates` and a new `BatchedEventPhase` to scheduler

## Discrete events use timeStamp to manage flushing

Now that the Flare event system and the existing event system co-exists, we've noticed far more flushing is occurring; which is problematic for cases where this behaviour was non-existant with only the existing event system. To counter this a `timeStamp` heuristic has been added by reading the `timeStamp` from the `window` object. We will probably want to change this to be an argument (rather than reading from window) in the future, but I didn't want to introduce too many changes in this PR.

The `timeStamp` is an important field on events, as it allows us to tell what events were part of the same real-world interaction. For example: `mouseup`, `pointerup` and `click` all should have the same `timeStamp`. If we know that they're all the same `timeStamp`, we can do a single flush for them all, which improves compatibility with Flare and also should improve runtime performance as we're flushing sync less.

### DIFF
--- a/packages/events/ReactGenericBatching.js
+++ b/packages/events/ReactGenericBatching.js
@@ -9,7 +9,6 @@ import {
   needsStateRestore,
   restoreStateIfNeeded,
 } from './ReactControlledComponent';
-import {enableEventAPI} from 'shared/ReactFeatureFlags';
 
 // Used as a way to call batchedUpdates when we don't have a reference to
 // the renderer. Such as when we're dispatching events or if third party
@@ -25,6 +24,7 @@ let discreteUpdatesImpl = function(fn, a, b, c) {
   return fn(a, b, c);
 };
 let flushDiscreteUpdatesImpl = function() {};
+let batchedEventUpdatesImpl = batchedUpdatesImpl;
 
 let isBatching = false;
 export function batchedUpdates(fn, bookkeeping) {
@@ -53,25 +53,48 @@ export function batchedUpdates(fn, bookkeeping) {
   }
 }
 
+export function batchedEventUpdates(fn, bookkeeping) {
+  if (isBatching) {
+    // If we are currently inside another batch, we need to wait until it
+    // fully completes before restoring state.
+    return fn(bookkeeping);
+  }
+  isBatching = true;
+  try {
+    return batchedEventUpdatesImpl(fn, bookkeeping);
+  } finally {
+    // Here we wait until all updates have propagated, which is important
+    // when using controlled components within layers:
+    // https://github.com/facebook/react/issues/1698
+    // Then we restore state of any controlled component.
+    isBatching = false;
+    const controlledComponentsHavePendingUpdates = needsStateRestore();
+    if (controlledComponentsHavePendingUpdates) {
+      // If a controlled event was fired, we may need to restore the state of
+      // the DOM node back to the controlled value. This is necessary when React
+      // bails out of the update without touching the DOM.
+      flushDiscreteUpdatesImpl();
+      restoreStateIfNeeded();
+    }
+  }
+}
+
 export function discreteUpdates(fn, a, b, c) {
   return discreteUpdatesImpl(fn, a, b, c);
 }
 
-let lastInteractiveTimeStamp = 0;
-
-export function flushDiscreteUpdates(timeStamp?: number) {
-  if (!enableEventAPI || !timeStamp || lastInteractiveTimeStamp !== timeStamp) {
-    lastInteractiveTimeStamp = timeStamp;
-    return flushDiscreteUpdatesImpl();
-  }
+export function flushDiscreteUpdates() {
+  return flushDiscreteUpdatesImpl();
 }
 
 export function setBatchingImplementation(
   _batchedUpdatesImpl,
   _discreteUpdatesImpl,
   _flushDiscreteUpdatesImpl,
+  _batchedEventUpdatesImpl,
 ) {
   batchedUpdatesImpl = _batchedUpdatesImpl;
   discreteUpdatesImpl = _discreteUpdatesImpl;
   flushDiscreteUpdatesImpl = _flushDiscreteUpdatesImpl;
+  batchedEventUpdatesImpl = _batchedEventUpdatesImpl;
 }
--- a/packages/events/ReactGenericBatching.js
+++ b/packages/events/ReactGenericBatching.js
@@ -27,6 +27,23 @@ let flushDiscreteUpdatesImpl = function() {};
 let batchedEventUpdatesImpl = batchedUpdatesImpl;
 
 let isBatching = false;
+
+function batchedUpdatesFinally() {
+  // Here we wait until all updates have propagated, which is important
+  // when using controlled components within layers:
+  // https://github.com/facebook/react/issues/1698
+  // Then we restore state of any controlled component.
+  isBatching = false;
+  const controlledComponentsHavePendingUpdates = needsStateRestore();
+  if (controlledComponentsHavePendingUpdates) {
+    // If a controlled event was fired, we may need to restore the state of
+    // the DOM node back to the controlled value. This is necessary when React
+    // bails out of the update without touching the DOM.
+    flushDiscreteUpdatesImpl();
+    restoreStateIfNeeded();
+  }
+}
+
 export function batchedUpdates(fn, bookkeeping) {
   if (isBatching) {
     // If we are currently inside another batch, we need to wait until it
@@ -37,19 +54,7 @@ export function batchedUpdates(fn, bookkeeping) {
   try {
     return batchedUpdatesImpl(fn, bookkeeping);
   } finally {
-    // Here we wait until all updates have propagated, which is important
-    // when using controlled components within layers:
-    // https://github.com/facebook/react/issues/1698
-    // Then we restore state of any controlled component.
-    isBatching = false;
-    const controlledComponentsHavePendingUpdates = needsStateRestore();
-    if (controlledComponentsHavePendingUpdates) {
-      // If a controlled event was fired, we may need to restore the state of
-      // the DOM node back to the controlled value. This is necessary when React
-      // bails out of the update without touching the DOM.
-      flushDiscreteUpdatesImpl();
-      restoreStateIfNeeded();
-    }
+    batchedUpdatesFinally();
   }
 }
 
@@ -63,19 +68,7 @@ export function batchedEventUpdates(fn, bookkeeping) {
   try {
     return batchedEventUpdatesImpl(fn, bookkeeping);
   } finally {
-    // Here we wait until all updates have propagated, which is important
-    // when using controlled components within layers:
-    // https://github.com/facebook/react/issues/1698
-    // Then we restore state of any controlled component.
-    isBatching = false;
-    const controlledComponentsHavePendingUpdates = needsStateRestore();
-    if (controlledComponentsHavePendingUpdates) {
-      // If a controlled event was fired, we may need to restore the state of
-      // the DOM node back to the controlled value. This is necessary when React
-      // bails out of the update without touching the DOM.
-      flushDiscreteUpdatesImpl();
-      restoreStateIfNeeded();
-    }
+    batchedUpdatesFinally();
   }
 }
 
--- a/packages/events/ReactGenericBatching.js
+++ b/packages/events/ReactGenericBatching.js
@@ -9,6 +9,7 @@ import {
   needsStateRestore,
   restoreStateIfNeeded,
 } from './ReactControlledComponent';
+import {enableEventAPI} from 'shared/ReactFeatureFlags';
 
 // Used as a way to call batchedUpdates when we don't have a reference to
 // the renderer. Such as when we're dispatching events or if third party
@@ -17,13 +18,13 @@ import {
 // scheduled work and instead do synchronous work.
 
 // Defaults
-let _batchedUpdatesImpl = function(fn, bookkeeping) {
+let batchedUpdatesImpl = function(fn, bookkeeping) {
   return fn(bookkeeping);
 };
-let _interactiveUpdatesImpl = function(fn, a, b, c) {
+let discreteUpdatesImpl = function(fn, a, b, c) {
   return fn(a, b, c);
 };
-let _flushInteractiveUpdatesImpl = function() {};
+let flushDiscreteUpdatesImpl = function() {};
 
 let isBatching = false;
 export function batchedUpdates(fn, bookkeeping) {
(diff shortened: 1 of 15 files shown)


packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -604,20 +605,33 @@ export function deferredUpdates<A>(fn: () => A): A {
...
   fn: (A, B, C) => R,
   a: A,
   b: B,
   c: C,
 ): R {
-  if (workPhase === NotWorking) {
-    // TODO: Remove this call. Instead of doing this automatically, the caller
-    // should explicitly call flushInteractiveUpdates.
-    flushPendingDiscreteUpdates();
+  const currentEvent = typeof window !== 'undefined' && window.event;
```
acdlite: Let's lift this logic into React DOM/Fire. The renderer can export two methods: `interactiveUpdates` and `flushInteractiveUpdates`.

Also maybe we should go with "discrete"? We've been using both but I kinda like "discrete" more. Not required to decide on that in this PR, though.

packages/react-events/src/__tests__/Press-test.internal.js
```diff
@@ -2301,4 +2303,58 @@ describe('Event responder: Press', () => {
...
+    dispatchEvent('pointerdown');
+    dispatchEvent('pointerup');
+    dispatchEvent('click');
+
+    expect(renderCounts).toBe(2);
+    Scheduler.flushAll();
+    expect(renderCounts).toBe(4);
+
+    document.body.removeChild(newContainer);
+  });
```
acdlite: Can you add a test that includes a subsequent event sequence? To check that `interactiveEventTimeStamp` gets reset properly.

acdlite:
Also how confident are we that this works in non-Chrome browsers?

packages/react-events/src/__tests__/Press-test.internal.js
trueadm: Added some further tests to verify this works.

packages/react-reconciler/src/ReactFiberScheduler.js
trueadm: Done :)

trueadm:
@acdlite I checked it Safari, FF, Chrome and Edge. All looking good with this change. I need to confirm with IE, but this is behind a flag anyway, so there should be no immediate risk.

packages/react-events/src/__tests__/Press-test.internal.js
```diff
@@ -2347,13 +2345,32 @@ describe('Event responder: Press', () => {
...
-    expect(renderCounts).toBe(4);
+    if (__DEV__) {
+      expect(renderCounts).toBe(4);
+    } else {
+      expect(renderCounts).toBe(2);
+    }
+
+    dispatchEvent('pointerdown', 100);
+    dispatchEvent('pointerup', 100);
+    // Ensure the timeStamp logic works
```
necolas: Is the Press event component the best place to test this logic?

trueadm: It’s fine for now. All responders should have some variant of this too.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -566,8 +566,8 @@ export function flushRoot(root: FiberRoot, expirationTime: ExpirationTime) {
   flushSyncCallbackQueue();
 }
 
-export function flushInteractiveUpdates() {
-  if (workPhase === RenderPhase || workPhase === CommitPhase) {
+export function flushDiscreteUpdates() {
+  if (workPhase !== NotWorking) {
```
acdlite: Why did you change this? It should flush inside `unstable_batchedUpdates`, too.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -604,17 +604,12 @@ export function deferredUpdates<A>(fn: () => A): A {
...
   b: B,
   c: C,
 ): R {
-  if (workPhase === NotWorking) {
-    // TODO: Remove this call. Instead of doing this automatically, the caller
-    // should explicitly call flushInteractiveUpdates.
-    flushPendingDiscreteUpdates();
-  }
   if (!revertPassiveEffectsChange) {
     // TODO: Remove this call for the same reason as above.
```
acdlite: Let's remove this `flushPassiveEffects` call now.

packages/events/ReactGenericBatching.js
```diff
@@ -46,26 +47,31 @@ export function batchedUpdates(fn, bookkeeping) {
...
 
-export function interactiveUpdates(fn, a, b, c) {
-  return _interactiveUpdatesImpl(fn, a, b, c);
+export function discreteUpdates(fn, a, b, c) {
+  return discreteUpdatesImpl(fn, a, b, c);
 }
 
-export function flushInteractiveUpdates() {
-  return _flushInteractiveUpdatesImpl();
+let lastInteractiveTimeStamp = 0;
```
acdlite: Hmm is this the right place to put this? (cc: @sebmarkbage) Seems good enough to merge but it wouldn't have been my first guess.

packages/react-reconciler/src/ReactFiberScheduler.js
trueadm: In the previous `interactiveUpdates` function, we only called `flushInteractiveUpdates` when `workPhase` was not `NotWorking`. So this was to duplicate that behaviour or some `act` tests fails. Would you prefer we had another version of this function that is only called on the event path and thus revert `flushDiscreteUpdates` back to the old conditional?

packages/react-reconciler/src/ReactFiberScheduler.js
trueadm: Remove `flushPassiveEffects` altogether and the flag, as it won't be used anywhere?

packages/events/ReactGenericBatching.js
trueadm: Seemed like the most sane place to put it, but I'm not bothered, it just meant less indirection on the existing codebase.

packages/react-reconciler/src/ReactFiberScheduler.js
trueadm: Actually, I think I know what you meant now. Changes are reflected in latest commit.

acdlite: The purpose of the check is to handle nested event dispatches, like `element.click()` or `element.focus()`, that are invoked inside a React lifecycle. In that case `workPhase === CommitPhase`; we can't synchronously flush pending discrete updates in that case because we're already in the middle of another render. So we exit and do nothing. Arguably, we should defer the nested event dispatch to a microtask, but that would be a breaking change in legacy mode. That seems outside the scope of this PR, though.

However, there's no valid case for `workPhase` to be `RenderPhase`. Without the check, this will trigger an internal invariant, but we should have a user-facing error with a helpful message.

So this is what I would suggest:

```js
if (workPhase === CommitPhase) {
  // We're inside the commit phase, so we can't synchronously flush pending
  // work. This is probably a nested event dispatch triggered by a
  // lifecycle/effect, like `el.focus()`. Exit.
  return;
}

if (workPhase === RenderPhase) {
  invariant(
    false,
    'unstable_flushDiscreteUpdates: Cannot flush updates when React is '
    + 'already rendering.'
  );
}

// Proceed with flushing
```

packages/react-reconciler/src/ReactFiberScheduler.js
acdlite: Yeah

acdlite: It's not a breaking change because this is only observable via the event system or via `unstable_interactiveUpdates`. After this PR, the event system will call `flushDiscreteUpdates`, which itself calls `flushPassiveEffects` behind a flag. So the behavior will be the same.

packages/react-reconciler/src/ReactFiberScheduler.js
acdlite: Oops didn't see your comment while I was typing :D

acdlite:
Can you add a test for:

```js
ReactDOM.unstable_batchedUpdates(() => {
  // This should flush synchronously
  ReactDOM.unstable_flushDiscreteUpdates();
  setState();
});
```

Or replace `batchedUpdates` with a click handler:

```js
function onClick() {
  // This should flush synchronously
  ReactDOM.unstable_flushDiscreteUpdates();
  setState();
};
```

I believe your current implementation of `flushDiscreteUpdates` would fail that test. See https://github.com/facebook/react/pull/15687#discussion_r285792879 for suggested implementation.

acdlite:
Might also want to add a test case that includes an imperative `el.click()` inside an effect.

acdlite:
Per our chat thread, I'd also be OK with adding a TODO above the early return for `BatchPhase` to remind ourselves to come back to it.

packages/react-events/src/__tests__/Press-test.internal.js
necolas: It's that the changes are being made to the core but the tests are here. And those tests have `__DEV__`-specific results

packages/events/ReactGenericBatching.js
trueadm: Addressed this in latest commit.

trueadm:
I've added more tests, updated tests, added `BatchedEventPhase` and also did various other things to tidy up the code.

packages/react-reconciler/src/ReactFiberScheduler.js
```diff
@@ -566,14 +567,20 @@ export function flushRoot(root: FiberRoot, expirationTime: ExpirationTime) {
...
 }
 
-export function flushInteractiveUpdates() {
-  if (workPhase === RenderPhase || workPhase === CommitPhase) {
-    // Can't synchronously flush interactive updates if React is already
-    // working. This is currently a no-op.
-    // TODO: Should we fire a warning? This happens if you synchronously invoke
-    // an input event inside an effect, like with `element.click()`.
+export function flushDiscreteUpdates() {
+  if (workPhase === CommitPhase || workPhase === BatchedPhase) {
```
acdlite: Why is this a no-op in `batchedUpdates`? Is there a test that failed otherwise?

trueadm: Yep, a bunch of tests failed from memory.

tjallingt:
I've previously encountered some issues related to Event.timeStamp in Vue with forks of older Chrome versions (specifically QtWebEngine) where the timeStamp would be some very large **negative** number.

Might be interesting to take a look at these:
 
https://github.com/vuejs/vue/issues/9681
https://github.com/vuejs/vue/commit/7591b9dc6dde314f2d32dcd7a8355f696a330979
https://github.com/vuejs/vue/blob/dev/src/platforms/web/runtime/modules/events.js#L42-L90

trueadm:
I saw some prior issues in Firefox, where the time stamps where using different precision timers (this longer numbers for some events than others). I believe this got fixed in Firefox after 52; there was a bug report I doh f earlier that stated the issue was resolved.

trueadm:
I saw some prior issues in Firefox, where the time stamps where using different precision timers (this longer numbers for some events than others). I believe this got fixed in Firefox after 52; there was a bug report I doh f earlier that stated the issue was resolved.

packages/react-reconciler/src/ReactFiberScheduler.js
acdlite: Can you put a TODO there? I'll take a look later. Doesn't have to block merge.

packages/react-events/src/__tests__/Press-test.internal.js
trueadm: We can follow up after and make tests in core too. I don't want to block this any longer given I'm on PTO in a few days. Maybe we can add this to our TODO list otherwise.

packages/react-reconciler/src/ReactFiberScheduler.js
trueadm: Added a TODO. Let's resync with this once I'm back from PTO; this at least unblocks us internally.

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -990,3 +998,29 @@ export function generateListeningKey(
   const passiveKey = passive ? '_passive' : '_active';
   return `${topLevelType}${passiveKey}`;
 }
+
+let lastDiscreteEventTimeStamp = 0;
+let useTimeStampHeuristic = true;
+
+if (canUseDOM) {
+  const firefoxUA = navigator.userAgent.match(/Firefox\/(.*)$/);
```
acdlite: What will the behavior be in Firefox then? Will it be broken, or merely less optimal?

Also perhaps there's a way to detect this that doesn't rely on UA sniffing. Like by dispatching a click event on a dummy node and checking the timestamps of the events.

trueadm: Actually, I just managed to test on FF 51 again, we don't need this anymore. The timestamps might be different resolutions, as pointed out with Vue, but we don't care for the relevance in terms of difference (> or <), we only care for the equality (===).

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -1001,6 +1001,16 @@ export function generateListeningKey(
...
+  // event.timeStamp isn't overly reliable due to inconsistencies in
+  // how different browsers have historically provided the time stamp.
+  // Some browsers provide high-resolution time stamps for all events,
+  // some provide low-resoltion time stamps for all events. FF < 52
+  // even mixes both time stamps together. Some browsers even report
+  // negative time stamps or time stamps that are 0 (iOS9) in some cases.
+  // Given we are only comparing two time stamps with equality (!==),
+  // we are safe from the resolution differences. If the time stamp is 0
+  // we bail-out of preventing the flush, which shouldn't have any
+  // effect on the desired output, other than we over-flush.
```
acdlite: The last sentence implies that it's only a performance change, which isn't true. It can affect semantics as well, such as if an earlier flush removes or adds event listeners that are fired in the subsequent flush.

acdlite:
^ We need to think about this heuristic more before we ship this to open source but since this is gated behind the Flare feature flag, it's good to merge so we can test it internally.

PR #15688, opened 2019-05-20
[React Fresh] Return host nodes for visual feedback

gaearon:
This PR adds a return value to `scheduleHotUpdate`:

```js
{|
  hostNodesForVisualFeedback: Array<Instance>
|}
```

This is an array of host nodes which can be used by the hot reloading runtime to show visual feedback related to the edited UI. Here's a gif to give you an idea of a possible way to use this. Note how the updated components are briefly highlighted in the browser right after edit:

Screen Recording 2019-05-20 at 08 25 PM

(Don't pay too much attention to the concrete visual feedback here; this is something we'll play more with. If this contextual feedback ends up being too much, we can always remove it and use a simple indicator, but I want to experiment more with it.)

The implementation goes as follows:

* We mark "affected fibers" as we traverse the tree to schedule hot updates.
  * We only mark them one level deep. If `Parent` and `Child` were edited in the same file, we'll only mark the `Parent`. This is because in the visual indication we'll likely want to coalesce children with parents anyway.
* During the hot update, we keep track of unmounted fibers, and their return references.
  * This will let us walk the parent path for disconnected fibers if necessary (see below why).
* After the hot update batch is flushed, we determine the host nodes:
  * If a fiber has host nodes below it, we add them to the set.
  * If a fiber has no host nodes below it, we add its closest parent host node to the set.
  * If a fiber has been unmounted, we perform this procedure for its closest still mounted parent instead.
  * If a fiber's parent is root, we use the root host node.

### No Component-Specific Visual Feedback

Maybe this is overkill and we just want an indicator in the screen corner or something. Then we can delete this code. I'm not sure and I wanted to play more with this.

### Move It Into Consuming Runtime

We could just return `affectedFibers` and call it a day. Then the consuming code could find the host nodes by copy pasting a similar implementation. That's actually how I started this. However, it's annoying to keep in sync with work tags. And the implementation is a bit tricky (e.g. tracking deletions). So if this is the workflow we want, we'd probably need to centralize it. Why not put it with the rest of the logic we're already testing then? So I put it here.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactFresh-test.internal.js
@@ -46,6 +46,7 @@ describe('ReactFresh', () => {
     Scheduler = require('scheduler');
     act = require('react-dom/test-utils').act;
     container = document.createElement('div');
+    container.className = 'container';
     document.body.appendChild(container);
 
     familiesByID = new Map();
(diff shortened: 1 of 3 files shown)


acdlite:
Did you consider using an effect tag? Then you can mutate during the actual commit phase, and you don't need an extra traversal.

gaearon:
This approach is flawed because the visual feedback should happen _before_ the reload is finished. Then it makes sense.

PR #15698, opened 2019-05-21
[Fresh] Set up infra for runtime and Babel plugin

gaearon:
* Adds a new `react-fresh` top level package.
  - Name TBD, I just need somewhere to put my stuff in.
* Adds a new `react-fresh/babel` entry point.
  - This will be the Babel plugin. It's a noop now. I added an empty snapshot test.
* Adds a new `react-fresh/runtime` entry point.
  - This will be the hot reloading runtime that manages component registration and Families.
  - It will be shared between web and RN implementations.
  - The first version I'm checking in is extracted from what we used in the test suite.

I want to keep it all in this repo so that I can write integration tests which run against latest React, and verify all pieces work together. This also makes it easier to ensure we actually fix this if we break it.

The next follow-up will be to actually write the Babel plugin, and add some kind of integration testing.


packages/react-fresh/babel.js
```diff
@@ -0,0 +1,13 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
+
+const ReactFreshBabelPlugin = require('./src/ReactFreshBabelPlugin');
+
+// This is hacky but makes it work with both Rollup and Jest.
+module.exports = ReactFreshBabelPlugin.default || ReactFreshBabelPlugin;
```
acdlite: I thought new packages can use ES modules.

gaearon: I don't think we enabled ESM output for anything yet because of how DEV/PROD split is done.

PR #15700, opened 2019-05-21
Fix missing return pointer assignment

acdlite:
I found a bug using the fuzz tester that manifested as incorrect ordering of children in the host tree, but whose root cause was a missing `return` pointer assignment on a work-in-progress fiber.

Usually return pointers are set during reconciliation (`reconcileChildFibers`) but this particular assignment happens inside the custom reconciliation implementation used by Suspense boundaries.

I would not be surprised if there were similar bugs related to incorrect return pointers. You're supposed to update the return pointer whenever a work-in-progress fiber is created, but there's nothing in the contract of the `createFiber` or `createWorkInProgress` function that implies this. I propose that we update their signatures to accept the return fiber as an argument. I will do this in a follow-up.

In this commit, I rearranged `updateSuspenseComponent` slightly so that every call to `createWorkInProgress` or a `createFiber*` function is immediately followed by a return pointer assignment.

I hardcoded the fuzz test case that surfaced the bug.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseFuzz-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseFuzz-test.internal.js
@@ -5,7 +5,7 @@ let Scheduler;
 let ReactFeatureFlags;
 let Random;
 
-const SEED = 0;
+const SEED = process.env.TEST_SEED || 'default';
 
 const prettyFormatPkg = require('pretty-format');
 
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/__tests__/ReactSuspenseFuzz-test.internal.js
```diff
@@ -5,7 +5,7 @@ let Scheduler;
 let ReactFeatureFlags;
 let Random;
 
-const SEED = 0;
+const SEED = process.env.TEST_SEED || 'default';
```
acdlite: Oops, didn't mean to commit this part

PR #15701, opened 2019-05-21
Flare: ensure discrete event flush is disabled during initial DOM mount

trueadm:
When testing https://github.com/facebook/react/pull/15687 internally, we encountered some issues where the invariant in `flushDiscreteUpdates` fired during legit tests.

The reason for this was because ReactDOM sets properties and attributes on elements on initial mount, which can also fire events on the DOM. We don't want these events to flush discrete events, so we now set a flag for this that we enable before setting DOM properties and then disable after setting them.

Furthermore, we felt that the `invariant` was maybe too strong, so this has been downgraded to a warning in DEV.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMEventListener-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMEventListener-test.js
@@ -474,7 +474,7 @@ describe('ReactDOMEventListener', () => {
     }
   });
 
-  it.only('should not warn when setting media attributes that trigger discrete events', () => {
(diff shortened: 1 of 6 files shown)


acdlite:
Maybe we should just revert the warning entirely if it's causing problems? Until we can verify that this approach is correct, and write tests. If the point is to unblock the sync then we should change as little as possible about the current behavior.

trueadm:
@acdlite I've added a regression test that confirms the expected behaviour.

trueadm:
We will come back to this one in the future. Closing this in favour of https://github.com/facebook/react/pull/15702

PR #15702, opened 2019-05-21
Flare: change flushDiscreteUpdates invariant to warning

trueadm:
When testing #15687 internally, we encountered some issues where the invariant in flushDiscreteUpdates fired during legit tests. On further inspection, using `invariant` was maybe too strong, so this has been downgraded to a `warning` in DEV.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMFiber-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMFiber-test.js
@@ -1026,14 +1026,8 @@ describe('ReactDOMFiber', () => {
 
   it('should not update event handlers until commit', () => {
     let ops = [];
-    let eventErrors = [];
     const handlerA = () => ops.push('A');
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/__tests__/ReactDOMFiber-test.js
```diff
@@ -1052,7 +1046,11 @@ describe('ReactDOMFiber', () => {
     class Click extends React.Component {
       constructor() {
         super();
-        node.click();
+        expect(() => {
+          node.click();
+        }).toWarnDev(
+          'Warning: unstable_flushDiscreteUpdates: Cannot flush updates when React is already rendering.',
```
acdlite: Should bikeshed this before it hits open source

trueadm: I'll leave you the key to the bikeshed. Help yourself.

PR #15704, opened 2019-05-22
Parallelize CircleCI jobs using workflows

acdlite:
Updates the CircleCI config to use the workflows feature to run jobs in parallel, instead of the `parallelism` option. This change alone doesn't improve the overall build time much, since almost all of the total time is spent running the Rollup script, which runs entirely sequentially. But it does improve reporting, and should make it easier to add additional parallel jobs in the future.

(My original motivation for this PR was because I want to use a workflow trigger to run a chron job for the fuzz tester.)

PRs report more granular information about the CI status:
Screen Shot 2019-05-21 at 8 41 42 PM

The Workflows view includes a nice flowchart. Note how the Rollup script takes the bulk of the time, as you would expect. (Parallelizing that step would be an easy win.)

Screen Shot 2019-05-21 at 8 34 12 PM


sophiebits:
ooooo very nice

bvaughn:
Slick!

PR #15707, opened 2019-05-22
[Flare] Account for fixed elements in getAbsoluteBoundingClientRect

trueadm:
This alter's the Press event responder so it can properly account for fixed elements that have scrollable regions. I've added inline comments explaining the logic and also tested this internally and it fixes a bunch of bugs.


packages/react-events/src/Press.js
```diff
@@ -436,10 +436,32 @@ function getAbsoluteBoundingClientRect(
   let offsetY = 0;
 
   // Traverse through all offset nodes
-  while (node != null && node.nodeType !== Node.DOCUMENT_NODE) {
-    offsetX += (node: any).scrollLeft;
-    offsetY += (node: any).scrollTop;
-    node = node.parentNode;
+  while (node != null) {
+    const parent = node.parentNode;
+    const scrollTop = node.scrollTop;
```
necolas: Since these causes the browser to synchronously reflow it could be worth moving this inline in the condition below so it's avoided when possible

trueadm: It's potentially used twice though, so accessing it twice is worse than once?

trueadm: Actually, I wasn't re-using the stored variable, now I am. As it is used on either path.

PR #15708, opened 2019-05-22
[Flare] update getEventCurrentTarget to use fiber tree

trueadm:
This PR updates the implementation of `getEventCurrentTarget` to use fiber tree rather than the DOM tree. This should be faster and result in less edge-cases around portals and other React abstractions.


packages/react-events/src/Focus.js
```diff
@@ -202,7 +202,7 @@ function handleRootPointerEvent(
   // Focus should stop being visible if a pointer is used on the element
   // after it was focused using a keyboard.
   if (
-    state.focusTarget === context.getEventCurrentTarget(event) &&
+    state.focusTarget === context.getEventCurrentTarget(event.target) &&
```
necolas: I think we should continue passing the event rather than the target

PR #15711, opened 2019-05-22
[Fresh] Initial Babel plugin implementation

gaearon:
This adds a basic implementation of the Babel plugin. It's incomplete and will be expanded as we go.

Its takes code like

```js
function Foo() {
  // ...
}

let Bar = () => {
  // ...
}
```

and turns it into something like

```js
function Foo() {
  // ...
}
_c = Foo;

let Bar = _c2 = () => {
  // ...
}

var _c;
var _c2;

__register__(_c, 'Foo');
__register__(_c2, 'Bar');
```

We rely on var and function declaration hoisting here. The registration all happens at the bottom so that if module init throws, we throw away the broken components.

The transform handles some basic cases like top-level functions and arrows (normal or exported). In the future I'll add support for some HOC patterns like `memo`. The transform intentionally injects variable in the middle of assignment so that in theory we could even do things like `_c3 = forwardRef(_c2 = memo(_c1 = function() { }))` without changing code semantics.

In the follow-ups, I'll handle more cases and tie this with our integration tests so we can verify the whole thing works as expected.

### DIFF
--- a/packages/react-fresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-fresh/src/ReactFreshBabelPlugin.js
@@ -29,7 +29,7 @@ export default function(babel) {
   }
 
   const buildRegistrationCall = template(`
-    __register__(HANDLE, PERSISTENT_ID)
+    __register__(HANDLE, PERSISTENT_ID);
   `);
 
   function isComponentishName(name) {
@@ -39,7 +39,7 @@ export default function(babel) {
   function isComponentish(node) {
     switch (node.type) {
       case 'FunctionDeclaration':
-        return isComponentishName(node.id.name);
+        return node.id !== null && isComponentishName(node.id.name);
       case 'VariableDeclarator':
         return (
           isComponentishName(node.id.name) &&
@@ -78,7 +78,9 @@ export default function(babel) {
         const functionName = path.node.id.name;
         const handle = createRegistration(programPath, functionName);
         insertAfterPath.insertAfter(
-          t.assignmentExpression('=', handle, path.node.id),
+          t.expressionStatement(
+            t.assignmentExpression('=', handle, path.node.id),
+          ),
         );
       },
       VariableDeclaration(path) {
--- a/packages/react-fresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-fresh/src/ReactFreshBabelPlugin.js
@@ -94,10 +94,6 @@ export default function(babel) {
           default:
             return;
         }
-        const declarations = path.node.declarations;
-        if (declarations.length !== 1) {
-          return;
-        }
         const declPath = path.get('declarations');
         if (declPath.length !== 1) {
           return;
--- a/packages/react-fresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-fresh/src/ReactFreshBabelPlugin.js
@@ -65,17 +65,28 @@ export default function(babel) {
   return {
     visitor: {
       FunctionDeclaration(path) {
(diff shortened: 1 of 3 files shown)


gaearon:
I initially thought component IDs would include the file name (e.g. `__register__(Foo, 'Foo.js$Foo')`. However I don't think this is actually a good idea. There is no guarantee that Babel receives correct file names. File name could be missing in `babel.transform` call for your integration — or worse, it could be the same for multiple files (e.g. a fake stub).

Instead I think the filename should be handled at the module runtime level. Whether we're in Metro/Webpack/something else, we should be able to know the module ID of currently initializing module. This is guaranteed to be correct because it's what the bundler actually uses. We'll also need this module ID anyway for preserving context identity.

So it seems like something Babel plugin doesn't need to handle.

bvaughn:
> We rely on var and function declaration hoisting here. The registration all happens at the bottom so that if module init throws, we throw away the broken components.

Is it important that the `var`s themselves are also declared at the bottom and rely on hoisting? Seems like this should only be necessary fro the registration?

(Still reviewing... just an initial question.)

packages/react-fresh/src/__tests__/__snapshots__/ReactFreshBabelPlugin-test.js.snap
```diff
@@ -1,3 +1,184 @@
...
+export function sum() {}
+export const Bad = 42;
+
+var _c;
+
+var _c2;
+
+var _c3;
+
+__register__(_c, 'Hello');
```
bvaughn: Wonder why it sometimes uses single quotes and other times uses double quotes with escape slashes...

gaearon:
Not strictly necessary. But since at least some need to be at the bottom (the ones in expressions like `const Foo = _c = ...`) I figured let's put all of them here. In general it would be good if we can move as much as possible down so that this transform isn't too noisy. Otherwise it can be a pain to debug when you don't have sourcemaps.

packages/react-fresh/src/__tests__/ReactFreshBabelPlugin-test.js
```diff
@@ -12,12 +12,170 @@ let freshPlugin = require('react-fresh/babel');
 
 function transform(input, options = {}) {
   return babel.transform(input, {
-    plugins: [[freshPlugin]],
+    babelrc: false,
+    plugins: ['syntax-jsx', freshPlugin],
   }).code;
 }
 
 describe('ReactFreshBabelPlugin', () => {
```
bvaughn: Nice tests.

packages/react-fresh/src/ReactFreshBabelPlugin.js
```diff
@@ -7,9 +7,127 @@
...
+    });
+    return handle;
+  }
+
+  const buildRegistrationCall = template(`
+    __register__(HANDLE, PERSISTENT_ID)
+  `);
+
+  function isComponentishName(name) {
+    return typeof name === 'string' && name[0] >= 'A' && name[0] <= 'Z';
```
bvaughn: 😆

gaearon:
Pushed a few bugfixes. I coalesced all variable definitions into one so that it doesn't take as many lines. Also checked that our own test suite works with the plugin enabled.

bvaughn:
> Not strictly necessary. But since at least some need to be at the bottom...

Gotcha! That's what I assumed. I was just curious. 😄Thanks for confirming.

> I coalesced all variable definitions into one so that it doesn't take as many lines. 

Sounds like a nice tweak.

PR #15716, opened 2019-05-22
Parallelizes the build script across multiple processes

acdlite:
## Based on #15757  and #15758 

Parallelizes the build script across multiple processes.

Screen Shot 2019-05-28 at 6 45 26 PM

(Our CircleCI plan supports up to 300 parallel processes (!!!) per job, but there's no point making the build job faster than the test job, so I capped it at 20.)

To do this, I needed to change how we generate the bundle size JSON file. I changed the build script to generate a separate size file per build. A downstream CI job consolidates them into a single `bundle-sizes.json` that contains the combined size information for every build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,6 +118,7 @@ jobs:
   build:
     docker: *docker
     environment: *environment
+    parallelism: 20
     steps:
       - checkout
       - *restore_yarn_cache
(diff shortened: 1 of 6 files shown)


acdlite:
Error code extraction is another blocker. Don't feel like dealing with that now. I'll pick this up some other time, not super crucial right now.

PR #15724, opened 2019-05-23
[Fresh] Babel plugin now handles HOCs + add integration tests

gaearon:
This is a follow-up to https://github.com/facebook/react/pull/15711 that teaches the Babel plugin to handle more cases including common patterns for higher-order components like `memo()`, `styled`, or chains of them. Concretely, we:

* Consider `A` declaration a likely component if it was used in `<A />` JSX call or `createElement(A)`
* Recognize patterns like `hoc1(hoc2(hoc3(() => { })))` and register each of them, as long as the outer definition is PascalCase.

Keep in mind registering is a no-op if something never gets rendered so false positives are acceptable. We might want to reconsider that later when we use the information about registrations to decide whether hot reload should propagate or stop at a module boundary (because all exports were registered). In that case false positives are not ideal. We can add smarter heuristics later, such as bailing out for functions that obviously aren't components. (E.g. when `prototype` is accessed or there's unexpected arguments or return value.)

I'm also adding a first integration test file. This one is different because it tests _both_ React, the runtime, and the Babel plugin together. It's as close as we have to an end-to-end test now. It's a bit harder to debug when it breaks (more moving pieces) but it also gives us more confidence that complex patterns actually work, and a way to reliably track regressions.

### DIFF
--- a/packages/react-fresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-fresh/src/ReactFreshBabelPlugin.js
@@ -105,8 +105,35 @@ export default function(babel) {
         if (!isComponentishName(name)) {
           return false;
         }
+        if (init.type === 'Identifier' || init.type === 'MemberExpression') {
+          return false;
+        }
         const initPath = path.get('init');
-        return findInnerComponents(inferredName, initPath, callback);
+        const foundInside = findInnerComponents(
+          inferredName,
+          initPath,
+          callback,
+        );
+        if (foundInside) {
+          return true;
+        }
+        // See if this identifier is used in JSX. Then it's a component.
+        const binding = path.scope.getBinding(name);
+        if (binding === undefined) {
+          return;
+        }
+        const referencePaths = binding.referencePaths;
+        for (let i = 0; i < referencePaths.length; i++) {
+          const ref = referencePaths[i];
+          if (
+            ref.node.type === 'JSXIdentifier' &&
+            ref.parent.type === 'JSXOpeningElement'
+          ) {
+            // const X = ... + later <X />
+            callback(inferredName, init, initPath);
+            return true;
+          }
+        }
       }
     }
     return false;
--- a/packages/react-fresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-fresh/src/ReactFreshBabelPlugin.js
@@ -36,18 +36,10 @@ export default function(babel) {
     const node = path.node;
     switch (node.type) {
       case 'FunctionDeclaration': {
-        const id = node.id;
-        if (id === null) {
-          return false;
-        }
-        const name = id.name;
-        if (!isComponentishName(name)) {
-          return false;
-        }
         // function Foo() {}
         // export function Foo() {}
         // export default function Foo() {}
-        callback(name, id, null);
+        callback(inferredName, node.id, null);
         return true;
       }
       case 'ArrowFunctionExpression': {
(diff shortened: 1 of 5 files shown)


packages/react-fresh/src/ReactFreshBabelPlugin.js
```diff
@@ -32,52 +32,173 @@ export default function(babel) {
...
-        return (
-          isComponentishName(node.id.name) &&
-          node.init !== null &&
-          (node.init.type === 'FunctionExpression' ||
-            (node.init.type === 'ArrowFunctionExpression' &&
-              node.init.body.type !== 'ArrowFunctionExpression'))
+      case 'FunctionDeclaration': {
+        const id = node.id;
+        if (id === null) {
+          return false;
```
Jessidhia: This excludes default export functions that are declared like `export default function() {}`. Is this intended?

gaearon: Yeah. Without a name we don’t know if it’s likely a component or just some random utility.

Maybe we could support that. But then, you probably should name your components anyway? So they show up in stacks correctly. I know there are Babel plugins that do this by filename but honestly is it too much to ask that people give names to their functions? 🙂

packages/react-fresh/src/ReactFreshBabelPlugin.js
```diff
@@ -32,52 +32,173 @@ export default function(babel) {
...
+      case 'VariableDeclarator': {
+        const init = node.init;
+        if (init === null) {
+          return false;
+        }
+        const name = node.id.name;
+        if (!isComponentishName(name)) {
+          return false;
+        }
+        if (init.type === 'Identifier' || init.type === 'MemberExpression') {
```
bvaughn: Components are occasionally defined as member expressions (I've done it for list items a time or two).

packages/react-fresh/src/ReactFreshBabelPlugin.js
```diff
@@ -32,52 +32,173 @@ export default function(babel) {
...
+                break;
+              case 'MemberExpression':
+                fnName = callee.property.name;
+                break;
+            }
+            switch (fnName) {
+              case 'createElement':
+              case 'jsx':
+              case 'jsxDEV':
+              case 'jsxs':
```
bvaughn: Nice

packages/react-fresh/src/__tests__/ReactFreshBabelPlugin-test.js
```diff
@@ -186,4 +186,103 @@ describe('ReactFreshBabelPlugin', () => {
...
+  it('registers likely HOCs with inline functions', () => {
+    expect(
+      transform(`
+        const A = forwardRef(function() {
+          return <h1>Foo</h1>;
+        });
+        const B = memo(React.forwardRef(() => {
+          return <h1>Foo</h1>;
+        }));
+        export default React.memo(forwardRef((props, ref) => {
```
bvaughn: I was initially surprised that the anonymous inner functions in `A` and `B` get registered, but the default export's anonymous function doesn't. I guess this is because the first two cases have a name that we can track back to and the third one doesn't.

_But_ if we changed the third one to be a named function, it would still get skipped.

This would match:
```js
export default function Foo() {
  return null;
}
```
And this would match:
```js
const Foo = function Foo(props, ref) {
  return null;
}
export default forwardRef(Foo);
```
But this would not:
```js
export default forwardRef(function Foo(props, ref) {
  return null;
});
```

I think the latter is a pattern I've used from time to time! 😁

packages/react-fresh/src/ReactFreshBabelPlugin.js
bvaughn: I think it's more than just `export default function() {}`. It also excludes _named_ functions exported with HOCs around them.

For example, this would match:
```js
export default function Foo() {
  return null;
}
```
And this would match:
```js
const Foo = function Foo(props, ref) {
  return null;
}
export default forwardRef(Foo);
```
But this would not:
```js
export default forwardRef(function Foo(props, ref) {
  return null;
});
```

I think the latter is a pattern I've used from time to time! 😁

gaearon: I'll be working more on these heuristics. I agree they don't make complete sense now.

Maybe I'll lean towards over-registering and then use some other heuristic for whether to propagate an update.

gaearon: Pushed a fix

gaearon:
>I forget: what's the plan for class components?

The plan is for them to always reset state. The exact mechanic is TBD.

packages/react-fresh/src/__tests__/ReactFreshBabelPlugin-test.js
gaearon: Fixed

PR #15725, opened 2019-05-23
Rename ReactFiberScheduler -> ReactFiberWorkLoop and extract throwException from Unwind

sebmarkbage:
This has been bothering me for a while.


packages/react-reconciler/src/ReactFiberThrow.js
```diff
@@ -0,0 +1,383 @@
...
+        }
+        break;
+      default:
+        break;
+    }
+    workInProgress = workInProgress.return;
+  } while (workInProgress !== null);
+}
+
+export {throwException, createRootErrorUpdate, createClassErrorUpdate};
```
acdlite: Any reason for putting the exports at the bottom? We're really inconsistent about this. Maybe we should pick one and stick with it.

sebmarkbage: This is just copy pasta so at least it's consistent with what it was.

I think the primary reason exports at the bottom are left over from the CommonJS form.

ronald11126:
L
Sent from Yahoo Mail on Android 
 
  On Thu, May 23, 2019 at 5:37 PM, Sebastian Markbåge<notifications@github.com> wrote:   

@sebmarkbage commented on this pull request.

In packages/react-reconciler/src/ReactFiberThrow.js:
> +            workInProgress,
+            errorInfo,
+            renderExpirationTime,
+          );
+          enqueueCapturedUpdate(workInProgress, update);
+          return;
+        }
+        break;
+      default:
+        break;
+    }
+    workInProgress = workInProgress.return;
+  } while (workInProgress !== null);
+}
+
+export {throwException, createRootErrorUpdate, createClassErrorUpdate};

This is just copy pasta so at least it's consistent with what it was.

I think the primary reason exports at the bottom are left over from the CommonJS form.

PR #15733, opened 2019-05-24
[Fresh] Generate signatures for Hooks

gaearon:
This adds support for editing components with Hooks.

State preservation already works by default. The goal here is to _not_ preserve state in cases where it is known to lead to problems. Such as reordering or deleting Hooks. We want to detect when that happens and remount.

This is implemented by generating `__signature__` calls for functions that contains calls to Hooks:

```js
function ButtonV1() {
  const [pressed, setPressed] = useState(false)
}
__signature__(ButtonV1, 'useState{[pressed, setPressed]}')

// -----------
// After edit
// -----------

function ButtonV1() {
  const [hover, setHover] = useState(false)
  const [pressed, setPressed] = useState(false)
  // ...
}
__signature__(ButtonV1, 'useState{[hover, setHover]}\nuseState{[pressed, setPressed]}')
```

If a signature changes, the Fresh runtime will tell React to remount that component type.

The signature contains names of all Hooks. Hooks that have a left hand side like `const [foo, setFoo]` have its source embedded in the signature. This catches renames (which often indicate invasive edits like changing state variable type).

If you use custom Hooks, we add them as the last argument:

```js
import useMyStuff from './useMyStuff'

function Foo() {
  const stuff = useMyStuff()
}
__signature__(
  Foo,
  'useMyStuff{stuff}',
  () => [useMyStuff] // <--- references, not strings!
);
```

This lets us recurse into each custom Hook when comparing signatures, so that if `useMyStuff` itself gets edited, we can remount the components using it.

The last argument is lazily evaluated (a function) for two reasons:

1. The reference might not be valid at the definition time because the Hook itself could be declared later in the file.
2. In React Native, we'll want to avoid triggering inline requires at module init time to avoid differences with production behavior.

The PR is easier to [read without whitespace](https://github.com/facebook/react/pull/15733/files?w=1) because some logic moved. You might also want to check commits separately.

### DIFF
--- a/packages/react-fresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-fresh/src/ReactFreshBabelPlugin.js
@@ -8,7 +8,7 @@
 'use strict';
 
 export default function(babel) {
-  const {types: t, template} = babel;
+  const {types: t} = babel;
 
   const registrationsByProgramPath = new Map();
   function createRegistration(programPath, persistentID) {
@@ -23,10 +23,6 @@ export default function(babel) {
     });
     return handle;
   }
-
-  const buildRegistrationCall = template(`
-    __register__(HANDLE, PERSISTENT_ID);
-  `);
 
   function isComponentishName(name) {
     return typeof name === 'string' && name[0] >= 'A' && name[0] <= 'Z';
@@ -158,6 +154,33 @@ export default function(babel) {
     return false;
   }
 
+  let hookCalls = new WeakMap();
+  let signedExpressions = new WeakSet();
+
+  function recordHookCall(functionNode, hookCallPath) {
+    if (!hookCalls.has(functionNode)) {
+      hookCalls.set(functionNode, []);
+    }
+    let hookCallsForFn = hookCalls.get(functionNode);
+    let key = '';
+    if (hookCallPath.parent.type === 'VariableDeclarator') {
+      // TODO: if there is no LHS, consider some other heuristic.
+      key = hookCallPath.parentPath.get('id').getSource();
+    }
+    hookCallsForFn.push({
+      name: hookCallPath.node.callee.name,
+      key,
+    });
+  }
+
+  function getHookCallsSignature(functionNode) {
+    const fnHookCalls = hookCalls.get(functionNode);
+    if (fnHookCalls === undefined) {
+      return null;
+    }
+    return fnHookCalls.map(call => call.name + '{' + call.key + '}').join('\n');
+  }
+
   return {
     visitor: {
       ExportDefaultDeclaration(path) {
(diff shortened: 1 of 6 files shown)


packages/react-fresh/src/ReactFreshBabelPlugin.js
```diff
@@ -158,6 +163,74 @@ export default function(babel) {
...
+      case 'useEffect':
+      case 'React.useEffect':
+      case 'useLayoutEffect':
+      case 'React.useLayoutEffect':
+      case 'useMemo':
+      case 'React.useMemo':
+      case 'useCallback':
+      case 'React.useCallback':
+      case 'useImperativeMethods':
+      case 'React.useImperativeMethods':
```
bvaughn: Why the decision to omit `useRef`, `useContext`, and `useDebugValue`?

packages/react-fresh/src/ReactFreshBabelPlugin.js
```diff
@@ -252,29 +468,99 @@ export default function(babel) {
...
+            if (
+              (targetExpr.type === 'ArrowFunctionExpression' ||
+                targetExpr.type === 'FunctionExpression') &&
+              targetPath.parent.type === 'VariableDeclarator'
+            ) {
+              // Special case when a function would get an inferred name:
+              // let Foo = () => {}
+              // let Foo = function() {}
+              // We'll register it on next line so that
+              // we don't mess up the inferred 'Foo' function name.
```
bvaughn: Nice.

packages/react-fresh/src/__tests__/ReactFreshBabelPlugin-test.js
```diff
@@ -295,7 +297,89 @@ describe('ReactFreshBabelPlugin', () => {
...
+            return <h1 ref={ref}>{foo}</h1>;
+          };
+        }
+
+        export let C = hoc();
+    `),
+    ).toMatchSnapshot();
+  });
+
+  it('includes custom hooks into the signatures', () => {
```
bvaughn: Nice

packages/react-fresh/src/ReactFreshBabelPlugin.js
gaearon: I forgot about them lol

bvaughn: 😂

packages/react-fresh/src/__tests__/__snapshots__/ReactFreshBabelPlugin-test.js.snap
```diff
@@ -1,5 +1,61 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`ReactFreshBabelPlugin generates signatures for function declarations calling hooks 1`] = `
+"
```
SimenB: you can use https://www.npmjs.com/package/jest-snapshot-serializer-raw to avoid the escaped `"` and extra wrapping `"`. We've done the same in Jest: https://github.com/facebook/jest/pull/7651

Makes them a bit easier to read, IMO 🙂

Issue #15732, opened 2019-05-24
“Rendered fewer hooks than expected” fired in an error boundary component without hooks


calebmer:
I’ve seen the “rendered fewer hooks than expected” invariant a couple times now pointing at a component rendered as a part of an error boundary that just caught an error when that component does not use hooks at all.

I’ve been able to fix this issue by un-commenting:

https://github.com/facebook/react/blob/8af1f879299e88f9e370c1f01d90efaeefcefd27/packages/react-reconciler/src/ReactFiberHooks.js#L372-L382

I haven’t been able to produce a minimal test-case yet 😣. I’m reporting anyway with the temporary fix that worked for me in case you have any insight on what a minimal test-case could be. I’ll update this issue if I keep running into the problem and I can create a minimal test-case.

calebmer:
I ran into the issue again and was able to produce a small repro: https://codesandbox.io/s/intelligent-cannon-f6bqx

Pressing the “throw” button triggers the error.

```
import React, { useCallback, useState } from "react";
import ReactDOM from "react-dom";

import "./styles.css";

function App() {
  return (
    <ErrorBoundary>
      <Thrower />
    </ErrorBoundary>
  );
}

class ErrorBoundary extends React.Component {
  state = {
    hasError: false
  };

  static getDerivedStateFromError() {
    return {
      hasError: true
    };
  }

  render() {
    return (
      <Wrapper>{this.state.hasError ? null : this.props.children}</Wrapper>
    );
  }
}

function Thrower() {
  const [shouldThrow, setShouldThrow] = useState(false);

  if (shouldThrow) {
    throw new Error("Throw!");
  }

  return (
    <button
      onClick={useCallback(() => {
        setShouldThrow(true);
      }, [])}
    >
      Throw!
    </button>
  );
}

function Wrapper({ children }) {
  return <>{children}</>;
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```

Pressing the “throw” button results in the invariant:

```
Uncaught Invariant Violation: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
    at invariant (https://f6bqx.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:55:15)
    at renderWithHooks (https://f6bqx.codesandbox.io/node_modules/react-dom/cjs/react-dom.development.js:13010:28)
```

With a component stack of:

```
The above error occurred in the <Wrapper> component:
[...]

PR #15747, opened 2019-05-27
Adds test for #15732.

calebmer:
A regression test for #15732. It’s broken on the 16.8.6 branch but works on master :tada:


gaearon:
Thancc

PR #15748, opened 2019-05-28
Adding new a warning message to Context.Consumer

laurosilvacom:
Adding warnings message to be more suggestive to the correct implementation.


avin-kavish:
I like the current version as is. It's a lot more polite.

threepointone:
Thank you for suggestion, but we use and prefer the question form more widely.

PR #15750, opened 2019-05-28
[Flare] Deeply prevent default on anchor elements

trueadm:
This PR adds support for deeply checking elements from within and event responder module. This is needed for the case where an element occurs outside the immediate event responder scope:

```jsx
<a href="#">
  <Press onPress={onPress}>
    <button ref={buttonRef} />
  </Press>
</a>
```

In the above case, the anchor is outside of the `Press` responder scope, so `isTargetWithinHostComponent` would always return `false`. With a new `deep` argument that can be passed to `isTargetWithinHostComponent`, the logic can now check deeper, outside the responder scope

### DIFF
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -424,11 +424,12 @@ const eventResponderContext: ReactResponderContext = {
   isTargetWithinHostComponent(
     target: Element | Document,
     elementType: string,
+    deep: boolean,
   ): boolean {
     validateResponderContext();
     let fiber = getClosestInstanceFromNode(target);
     while (fiber !== null) {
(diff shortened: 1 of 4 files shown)


necolas:
The example markup looks pretty suspect. We should talk about whether patterns like that make sense to support

trueadm:
@necolas The most obvious case is where you have nested Press responders.

necolas:
That's not what the example code is showing. And this seems to be related to the hack that nests interactive DOM elements

trueadm:
@necolas I took the example from the test in the code, but essentially it's this:

```jsx
<Press>
  <a>
    <Press>
      <button>
    </Press>
  </a>
</Press>
```

Sorry for the lack of clarity in my comment.

necolas:
OK yeah that makes sense and is due to the dodgy FB5 hack for nesting interactive elements

PR #15756, opened 2019-05-28

threepointone:
via #15319

(redid #15399, removed the changes that didn't even belong in it.)

So tl;dr this PR -

- uses a renderer specific object as the value of ReactCurrentActingRendererSigil.current
- checks this value on 'updateContainer`
- checks this value when setting a state hook's value
- adds a fixture folder for act() (which is run on CI)

- using the wrong act() shouldn't silence the 'missing act' warning
- using the wrong act() logs a warning tha you're, er, using the wrong act()

(Please see  #15399 for the long spiel on the mechanics for this.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
     "test-prod-build": "yarn test-build-prod",
     "test-build": "cross-env NODE_ENV=development jest --config ./scripts/jest/config.build.js",
     "test-build-prod": "cross-env NODE_ENV=production jest --config ./scripts/jest/config.build.js",
+    "test-dom-fixture": "cd fixtures/dom && yarn && yarn prestart && yarn test",
     "flow": "node ./scripts/tasks/flow.js",
     "flow-ci": "node ./scripts/tasks/flow-ci.js",
     "prettier": "node ./scripts/prettier/index.js write-changed",
--- a/packages/react-dom/src/test-utils/ReactTestUtilsAct.js
+++ b/packages/react-dom/src/test-utils/ReactTestUtilsAct.js
@@ -85,16 +85,19 @@ let actingUpdatesScopeDepth = 0;
 
 function act(callback: () => Thenable) {
   let previousActingUpdatesScopeDepth = actingUpdatesScopeDepth;
-  let previousActingUpdatesSigil = ReactActingRendererSigil.current;
+  let previousActingUpdatesSigil;
   actingUpdatesScopeDepth++;
   // we use the function flushPassiveEffects directly as the sigil,
   // since it's unique to a renderer
-  ReactActingRendererSigil.current = flushPassiveEffects;
+  if (__DEV__) {
+    previousActingUpdatesSigil = ReactActingRendererSigil.current;
+    ReactActingRendererSigil.current = flushPassiveEffects;
+  }
 
   function onDone() {
     actingUpdatesScopeDepth--;
-    ReactActingRendererSigil.current = previousActingUpdatesSigil;
     if (__DEV__) {
+      ReactActingRendererSigil.current = previousActingUpdatesSigil;
       if (actingUpdatesScopeDepth > previousActingUpdatesScopeDepth) {
         // if it's _less than_ previousActingUpdatesScopeDepth, then we can assume the 'other' one has warned
         warningWithoutStack(
--- a/packages/react-dom/src/test-utils/ReactTestUtilsAct.js
+++ b/packages/react-dom/src/test-utils/ReactTestUtilsAct.js
@@ -37,7 +37,7 @@ const [
 
 const batchedUpdates = ReactDOM.unstable_batchedUpdates;
 
-const {ReactShouldWarnActingUpdates} = ReactSharedInternals;
+const {ReactActingRendererSigil} = ReactSharedInternals;
 
 // this implementation should be exactly the same in
 // ReactTestUtilsAct.js, ReactTestRendererAct.js, createReactNoop.js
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -107,6 +107,7 @@
     "test-prod-build": "yarn test-build-prod",
     "test-build": "cross-env NODE_ENV=development jest --config ./scripts/jest/config.build.js",
     "test-build-prod": "cross-env NODE_ENV=production jest --config ./scripts/jest/config.build.js",
+    "test-dom-fixture": "cd fixtures/dom && yarn && yarn prestart && yarn test",
```
acdlite: Create a CI job for this please. Most people will not remember to run it.

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
```diff
@@ -85,17 +85,19 @@ let actingUpdatesScopeDepth = 0;
...
   if (__DEV__) {
-    ReactShouldWarnActingUpdates.current = true;
+    previousActingUpdatesSigil = ReactActingRendererSigil.current;
+    ReactActingRendererSigil.current = flushPassiveEffects;
   }
 
   function onDone() {
     actingUpdatesScopeDepth--;
     if (__DEV__) {
-      if (actingUpdatesScopeDepth === 0) {
```
acdlite: I don't get how you were able to remove this depth check. Can you explain?

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
```diff
@@ -85,17 +85,19 @@ let actingUpdatesScopeDepth = 0;
...
 function act(callback: () => Thenable) {
   let previousActingUpdatesScopeDepth = actingUpdatesScopeDepth;
+  let previousActingUpdatesSigil;
   actingUpdatesScopeDepth++;
+  // we use the function flushPassiveEffects directly as the sigil,
+  // since it's unique to a renderer
   if (__DEV__) {
-    ReactShouldWarnActingUpdates.current = true;
+    previousActingUpdatesSigil = ReactActingRendererSigil.current;
+    ReactActingRendererSigil.current = flushPassiveEffects;
```
acdlite: Could you use a separate export instead? I find this unnecessarily confusing for no benefit, given that it's dev-only.

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
threepointone: ReactActingRendererSigil.current starts with a `.current` value of `null`. Inside `warnIfNotCurrentlyActingUpdatesInDev`, we check whether `.current === flushPassiveEffects)`
As the act() scope opens, it sets the value to flushPassiveEffects which isn't null, *until* the last scope, when previousActingUpdatesSigil gets set as null, triggering the warning.

We could have had the same pattern when the sigil was just a boolean, by saving `.current` as a `previousActingSigilBoolean` or something.

acdlite: Makes sense

package.json
threepointone: I did! It's in https://github.com/facebook/react/pull/15756/files#diff-1d37e48f9ceff6d8030570cd36286a61R167

acdlite: Oh you're right. Missed it!

threepointone:
updated the PR to use an explicit export (from `ReactFiberWorkLoop`) as the value of the sigil. I'll land this later tonight.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2276,11 +2276,46 @@ function warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber) {
...
+// as the value of ReactCurrentActingRendererSigil.current
+// This identity lets us identify (ha!) when the wrong renderer's act()
+// wraps anothers' updates/effects
+export const ReactActingRendererSigil = {};
+
+export function warnIfNotScopedWithMatchingAct(fiber: Fiber): void {
+  if (__DEV__) {
+    if (
+      ReactCurrentActingRendererSigil.current !== null &&
+      // use the function flushPassiveEffects directly as the sigil
```
gaearon: Outdated comment

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2276,11 +2276,46 @@ function warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber) {
...
+      // so this comparison is expected here
+      ReactCurrentActingRendererSigil.current !== ReactActingRendererSigil
+    ) {
+      // it looks like we're using the wrong matching act(), so log a warning
```
gaearon: This comment kinda repeats the warning?

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2276,11 +2276,46 @@ function warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber) {
...
+      warningWithoutStack(
+        false,
+        "It looks like you're using the wrong act() around your test interactions.\n" +
+          'Be sure to use the matching version of act() corresponding to your renderer:\n\n' +
+          '// for react-dom:\n' +
+          "import {act} from 'react-test-utils';\n" +
```
threepointone: there's no such package either 🤦‍♂️

PR #15758, opened 2019-05-29
Don't extract errors in CI

acdlite:
Removes `--extract-errors` argument from CI build script command. Instead, the author is expected to run `yarn extract-errors` locally or manually edit the error code map.

The lint rule should be sufficient to catch unminified errors, but as an extra precaution, I added a post build step that greps the production bundles. The post-build step works even if someone disables the lint rule for a specific line or file.


bvaughn:
> Instead, the author is expected to run yarn extract-errors locally or manually edit the error code map.

Seems easy to forget. I would expect an update to the release script's logged instructions to remind us to do this.

PR #15763, opened 2019-05-29
warn if passive effects get queued outside of an act() call.

threepointone:
This adds a warning when effects are queued outside of an act() scope. Please see this diff with whitespace off (add `?w=1` to your url)

While the code itself isn't much (it adds the warning to `mountEffect()` and `updateEffect()` in `ReactFiberHooks`), it does change a lot of our tests. We follow a bad-ish pattern here, which is doing asserts inside `act()` scopes, but it makes sense for *us* because we're testing intermediate states, and we're manually flush/yield what we need in these tests.

This commit has *one* last failing test. Working on it, might reach out to one of you for help.

Please *please* look closely at the tests I've changed; I went about it with a rough-ish hand, wrapping whatever started warning with `act()` calls. For bits that seemed a bit hard to wrap with `act()` without drastically changing the test, I explicitly caught the missing `act()` warnings with `expect(...).toWarnDev([...])` statements.

An important question - Is this worth it? Consider an app built with both classical, and functional components. Assume that the class ones are parents to functional children. As we add hooks and effects to children, tests that used to pass for class components will start failing, meaning people will have to update tests for older components they might not be interested in touching.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMHooks-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMHooks-test.js
@@ -55,7 +55,6 @@ describe('ReactDOMHooks', () => {
 
     // we explicitly catch the missing act() warnings
     // to simulate this tricky repro
-    // todo - is this ok?
     expect(() => {
       ReactDOM.render(<Example1 n={1} />, container);
       expect(container.textContent).toBe('1');
--- a/packages/react-dom/src/__tests__/ReactDOMHooks-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMHooks-test.js
@@ -11,6 +11,7 @@
 
 let React;
 let ReactDOM;
+let act;
 let Scheduler;
 
 describe('ReactDOMHooks', () => {
@@ -21,6 +22,7 @@ describe('ReactDOMHooks', () => {
 
     React = require('react');
     ReactDOM = require('react-dom');
+    act = require('react-dom/test-utils').act;
     Scheduler = require('scheduler');
 
     container = document.createElement('div');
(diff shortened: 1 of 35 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2276,6 +2276,29 @@ function warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber) {
   }
 }
 
+export function warnIfNotCurrentlyActingEffectsInDEV(fiber: Fiber): void {
+  if (__DEV__) {
+    if (ReactShouldWarnActingUpdates.current === false) {
+      warningWithoutStack(
+        false,
+        'Your test just caused an effect from %s, but was not wrapped in act(...).\n\n' +
```
threepointone: This should be a better message, punting on that till I get approval on the rest of this.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
```diff
@@ -546,10 +548,12 @@ describe('ReactDOMServerHooks', () => {
         });
         return <Text text={'Count: ' + props.count} />;
       }
-      const domNode = await render(<Counter count={0} />);
-      expect(clearYields()).toEqual(['Count: 0']);
-      expect(domNode.tagName).toEqual('SPAN');
-      expect(domNode.textContent).toEqual('Count: 0');
+      await act(async () => {
+        const domNode = await render(<Counter count={0} />);
```
threepointone: one variant of this test fails "renders should ignore effects on the server with client render on top of good server markup"

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -1966,28 +2043,33 @@ describe('ReactHooksWithNoopRenderer', () => {
...
-      expect(Scheduler).toFlushAndYieldThrough(['Sync effect']);
-      ReactNoop.flushPassiveEffects();
+      act(() => {
+        ReactNoop.render(<App showMore={false} />, () =>
+          Scheduler.yieldValue('Sync effect'),
+        );
+        expect(Scheduler).toFlushAndYieldThrough(['Sync effect']);
+      });
+
+      // ReactNoop.flushPassiveEffects();
```
threepointone: stray comment

packages/react-dom/src/__tests__/ReactUpdates-test.js
```diff
@@ -1618,8 +1621,11 @@ describe('ReactUpdates', () => {
       let stack = null;
       let originalConsoleError = console.error;
       console.error = (e, s) => {
-        error = e;
-        stack = s;
+        // skip the missing act() error
+        if (e.slice(0, 40) !== 'Warning: Your test just caused an effect') {
```
acdlite: Push to a log of warnings and assert on the log at the end

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
acdlite: Remove this plz

threepointone:
working on this, changes planned.

threepointone:
fixed the failing test, and amended the other one; see this commit for details https://github.com/facebook/react/pull/15763/commits/46c0741a1cc960a00acde51f8d974155fa9f1ec9 I'll work on a better warning message and land this soon.

packages/react-dom/src/__tests__/ReactDOMHooks-test.js
```diff
@@ -53,23 +53,33 @@ describe('ReactDOMHooks', () => {
...
-    expect(container.textContent).toBe('2');
-    expect(container2.textContent).toBe('2'); // Not flushed yet
-    expect(container3.textContent).toBe('3'); // Not flushed yet
-    Scheduler.flushAll();
-    expect(container.textContent).toBe('2');
-    expect(container2.textContent).toBe('4');
-    expect(container3.textContent).toBe('6');
+    // we explicitly catch the missing act() warnings
+    // to simulate this tricky repro
+    // todo - is this ok?
```
threepointone: this isn't a todo anymore. will remove.

PR #15766, opened 2019-05-29
[Flare] Support legacy browser Spacebar key value

necolas:
IE 9,10,11 and Firefox < 37 can report spacebar key as `Spacebar`.


necolas:
> I wonder if we also need to add the event.key polyfill

I don't think that's been needed for a while https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#Browser_compatibility

PR #15771, opened 2019-05-30
Fix sizebot

acdlite:
The previous naming scheme for the partial size info files used the filename of the bundle. However, there are cases where multiple bundles have the same filename. This meant whichever bundle finishes last overwrites the previous ones with the same name.

The updated naming scheme is `bundle-sizes-<CI_NODE_INDEX>.json`. Instead of generating a separate info file per bundle, it now creates one per process.


scripts/rollup/build.js
```diff
@@ -634,6 +634,8 @@ function handleRollupError(error) {
 }
 
 async function buildEverything() {
+  await asyncRimRaf('build');
```
acdlite: I accidentally removed this in the previous PR so I'm adding it back

PR #15773, opened 2019-05-30
Increase parallelism again

acdlite:
Increased to 50 because let's try it and see? Post-build jobs are still the last to finish so if this makes the build job faster it will decrease the total CI time, too.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,7 +118,7 @@ jobs:
   build:
     docker: *docker
     environment: *environment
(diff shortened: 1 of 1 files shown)


acdlite:
Meh doesn't help much.

The issue is that the prod renderer bundles take way longer than everything else, so if one container happens to get assigned many of those bundles then the whole job can only be as fast as that container.

Could address by distributing the prod bundles better. Closing this for now.

Issue #266, opened 2013-08-14
.bind(null, ...) should work on autobound methods for partial application


sophiebits:
Currently it complains if the context isn't equal to `this`. Seems like `null` should be okay too.

PR #15778, opened 2019-05-30
chore: upgrade to jest 24

SimenB:
This won't work on CI since `pretty-format` uses `react-is` internally, and since yarn finds `react-is` in this repo, it won't install it. I don't know how to tell yarn to install a version of `react-is` into `pretty-format`s `node_modules`? To test this locally, I just did `cp -R ../jest/node_modules/react-is node_modules/pretty-format/node_modules` which worked fine. Maybe @arcanis can help?

All tests pass locally 🙂 

---

~This contains #15779 with the upgrade to jest 24 applied on top of it~


package.json
```diff
@@ -4,12 +4,12 @@
     "packages/*"
   ],
   "devDependencies": {
+    "@mattiasbuelens/web-streams-polyfill": "0.1.0",
     "art": "^0.10.1",
     "babel-cli": "^6.6.5",
     "babel-code-frame": "^6.26.0",
     "babel-core": "^6.0.0",
     "babel-eslint": "^10.0.0",
-    "babel-jest": "^23.0.1",
```
SimenB: you don't use this, you have your own transformer

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -521,7 +521,7 @@ const tests = {
       // Valid because the ref is captured.
       code: `
         function useMyThing(myRef) {
-          const myRef = useRef();
+          myRef = useRef();
```
SimenB: syntax error raised by babel since `myRef` existed already

packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
```diff
@@ -95,7 +95,7 @@ eslintTester.run('react-hooks', ReactHooksESLintRule, {
       useHook2 = () => { useState(); };
       ({useHook: () => { useState(); }});
       ({useHook() { useState(); }});
-      const {useHook = () => { useState(); }} = {};
+      const {useHook3 = () => { useState(); }} = {};
```
SimenB: syntax error raised by babel since `useHook` existed already

packages/react-dom/src/__tests__/ReactDOMServerIntegrationElements-test.js
```diff
@@ -648,7 +648,7 @@ describe('ReactDOMServerIntegration', () => {
       });
     });
 
-    describe('component hierarchies', async function() {
+    describe('component hierarchies', function() {
```
SimenB: `describe` must return `void`

packages/react-test-renderer/src/__tests__/ReactTestRendererAsync-test.js
```diff
@@ -160,7 +160,7 @@ describe('ReactTestRendererAsync', () => {
 
       expect(() =>
         expect(Scheduler).toFlushAndYieldThrough(['foo', 'baz']),
-      ).toThrow('Expected value to equal:');
+      ).toThrow('// deep equality');
```
SimenB: we've changed the error messages

packages/react/src/__tests__/ReactCoffeeScriptClass-test.coffee
```diff
@@ -543,3 +543,4 @@ describe 'ReactCoffeeScriptClass', ->
     node = ReactDOM.findDOMNode(instance)
     expect(node).toBe container.firstChild
     undefined
+  undefined
```
SimenB: same as above, `describe` must return `undefined`

packages/react/src/__tests__/createReactClassIntegration-test.js
```diff
@@ -453,7 +453,7 @@ describe('create-react-class-integration', () => {
     expect(() =>
       ReactDOM.render(<Foo foo="foo" />, document.createElement('div')),
     ).toWarnDev(
-      'Component: getDerivedStateFromProps() is defined as an instance method ' +
+      'Foo: getDerivedStateFromProps() is defined as an instance method ' +
```
SimenB: probably due to us having source maps now

scripts/jest/config.base.js
```diff
@@ -12,12 +12,13 @@ module.exports = {
...
-  setupTestFrameworkScriptFile: require.resolve('./setupTests.js'),
+  setupFilesAfterEnv: [require.resolve('./setupTests.js')],
   // Only include files directly in __tests__, not in nested folders.
   testRegex: '/__tests__/[^/]*(\\.js|\\.coffee|[^d]\\.ts)$',
   moduleFileExtensions: ['js', 'json', 'node', 'coffee', 'ts'],
   rootDir: process.cwd(),
   roots: ['<rootDir>/packages', '<rootDir>/scripts'],
   collectCoverageFrom: ['packages/**/*.js'],
   timers: 'fake',
+  testURL: 'about:blank',
```
SimenB: the default changed in some Jest release from `about:blank` to `http://localhost` - that broke a test you had for `href`

scripts/jest/config.base.js
```diff
@@ -12,12 +12,13 @@ module.exports = {
     '.*': require.resolve('./preprocessor.js'),
   },
   setupFiles: [require.resolve('./setupEnvironment.js')],
-  setupTestFrameworkScriptFile: require.resolve('./setupTests.js'),
+  setupFilesAfterEnv: [require.resolve('./setupTests.js')],
```
SimenB: renamed option

scripts/jest/preprocessor.js
```diff
@@ -53,6 +53,9 @@ module.exports = {
     if (filePath.match(/\.ts$/) && !filePath.match(/\.d\.ts$/)) {
       return tsPreprocessor.compile(src, filePath);
     }
+    if (filePath.match(/\.json$/) ) {
```
SimenB: Jest passes json files through transformers now, and babel didn't like it 🙂

scripts/jest/preprocessor.js
```diff
@@ -70,7 +73,7 @@ module.exports = {
               }
             : {}
         )
-      ).code;
```
SimenB: this gives jest source maps from babel

arcanis:
> since yarn finds `react-is` in this repo, it won't install it. I don't know how to tell yarn to install a version of `react-is` into `pretty-format`s `node_modules`?

Not sure I understand - you mean that `/node_modules/react-is` exists, so Yarn doesn't create `/node_modules/pretty-format/node_modules/react-is`? Yarn should only do this if the version at the top-level is compatible with the version requested by `pretty-format`. Isn't it the case?

SimenB:
>Not sure I understand - you mean that `/node_modules/react-is` exists, so Yarn doesn't create `/node_modules/pretty-format/node_modules/react-is`?

Correct.

>Yarn should only do this if the version at the top-level is compatible with the version requested by `pretty-format`. Isn't it the case?

It is semver compatible, but the package has not been built/compiled to cjs so it doesn't work. Its main point to this file: https://github.com/facebook/react/blob/3b2302253f13c9cd049fac10784e10b6c582e3b6/packages/react-is/index.js

I'd like to tell yarn to ignore the `react-is` from the workspace and install it into `pretty-format` (at least, I think that's what I want 😀)

arcanis:
Oh I see, so you want to ignore the local workspace .. hm .. this should be doable with [resolutions](https://yarnpkg.com/en/docs/selective-version-resolutions). Something like this, maybe?

```json
{
  "resolutions": {
    "pretty-format/react-is": "npm:react-is@<insert the semver range>"
  }
}
```

SimenB:
That sorta works - it does indeed install from the npm registry, but it installs it to `node_modules/react-is` meaning the one in the workspace is not there. That might break other code importing it (via some haste resolution or some such)?

arcanis:
Damn, that's likely a bug in the `resolutions` implementation (fwiw we have tests on this in the v2 trunk, so I know it works properly there) 😢

I don't see an easy fix then ... maybe a `resolutions` entry on `pretty-format` instead, to point to a local copy that would list `react-is@npm:^<range>`? It starts to be a bit hacky 🤔

It's interesting we never got this problem before, though!

threepointone:
Now's a good time for us to land this. What do you folks think is the best option right to resolve this pretty-format/react-is issue?

SimenB:
I'm not sure :( I didn't try out @arcanis's last suggestion since he said it was hacky. Maybe give that a go? I'm on vacation for a few more days, so if you have time, that'd be great. Otherwise I'll give it a whack some time next week

(Landing #15779 in the meantime would be great, to reduce this diff)

threepointone:
Cool cool, enjoy your vacation! I’ll poke at this myself soon

packages/react/src/__tests__/createReactClassIntegration-test.js
SimenB: nope, just babel 7 being better 🙂

SimenB:
I've rebased this now that the other PR landed, but it still has the same `react-is` issue. Not sure how to deal with it, tbh

bvaughn:
Figuring out a way to land this might make #266 a little easier to land, but I'm going to try to forge forward and not block on this. 😄

SimenB:
I played around some more with this, and got nowhere 😅 One thing that _works_ is https://github.com/facebook/react/pull/15778#issuecomment-497321320, but as mentioned that means that any other module in this monorepo that does `require('react-is');` will get the one from npm instead of the local one. Grepping shows this only happens in one spot, but still: https://github.com/facebook/react/blob/6f86294e686e172fb8da32880f70fa038d644c74/packages/react-test-renderer/src/ReactShallowRenderer.js#L11.

I'll push a commit with it. Thoughts on `moduleNameMapper` in jest config to point `react-is` to a local one? Then at least for tests you'll get the local version

SimenB:
Bah, that broke the build... Tests passed, though! I don't know how to solve this beyond manually installing `react-is` from npm into `node_modules/pretty-format/node_modules` via `postinstall`. Or use yarn 2 if it's ready? It's in RC as of 3 hours ago, that means "stable" in JS land, right?

https://github.com/yarnpkg/berry/releases/tag/2019-08-16

arcanis:
Fwiw I don't think the React repo will be able to switch unless Flow adds for in-zip loading, or someone writes a linker plugin for Yarn 2 that basically unplugs everything (which is doable).

Feel free to ping me on the Yarnpkg discord channel if you want to discuss it further.

SimenB:
That last commit is about as far from yarn v2 compatible you could get I think :P CI passed, though! :tada:

I think I've spent as much time as I'm willing to on this now, unless someone comes up with any fresh new ideas

scripts/yarn/downloadReactIsForPrettyFormat.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const prettyFormatNodeModulesReactIsDir = path.join(
+  path.dirname(prettyFormatPkgPath),
+  'node_modules/react-is'
+);
+
+rimraf.sync(prettyFormatNodeModulesReactIsDir);
+
+pacote
```
SimenB: could possibly check in the tarball to avoid the network hit? Not sure if you want this at all, so I went with what's easiest/quickest/shortest

bvaughn:
> One thing that works is #15778 (comment), but as mentioned that means that any other module in this monorepo that does require('react-is'); will get the one from npm instead of the local one. Grepping shows this only happens in one spot

The new DevTools package (#16381) uses it too, unfortunately.

SimenB:
My current solution should work for it - I didn't go with the linked one 🙂

SimenB:
Or did you mean that devtools requires `react-is` normally, and fails somehow? If so the same approach as in this PR should work

SimenB:
@threepointone thoughts on this?

SimenB:
rebased

threepointone:
I’ll bring this up in our weekly sync this week, we’ll take a resolution on it one way or the other.

SimenB:
Sounds good! Looking forward to it (hopefully) landing 😀

arcanis:
Btw, wouldn't it be simpler to just write this one module in standard js? It sounds quite small.

Another thought - would it work to run Jest through babel-register, configuring it to only transpile this one folder?

SimenB:
Can't really comment on the first one. As for babel register though, I _think_ it wouldn't work when jest spins up workers to run tests.

---

I have no idea how I broke danger...

SimenB:
OK, I broke danger by giving it babel 7 instead of 6. One quickfix is

```diff
diff --git c/package.json w/package.json
index db625cd6e..90561d8d5 100644
--- c/package.json
+++ w/package.json
@@ -34,6 +34,7 @@
     "@babel/traverse": "^7.0.0",
     "@mattiasbuelens/web-streams-polyfill": "0.1.0",
     "art": "^0.10.1",
+    "babel-core": "^6.26.3",
     "babel-eslint": "^10.0.0",
     "babel-plugin-syntax-trailing-function-commas": "^6.5.0",
     "babylon": "6.18.0",
diff --git c/yarn.lock w/yarn.lock
index 86804bbbf..83f231c6b 100644
--- c/yarn.lock
+++ w/yarn.lock
@@ -2100,7 +2100,7 @@ babel-code-frame@^6.16.0, babel-code-frame@^6.26.0:
     esutils "^2.0.2"
     js-tokens "^3.0.2"
 
-babel-core@^6.26.0, babel-core@^6.7.2:
+babel-core@^6.26.0, babel-core@^6.26.3, babel-core@^6.7.2:
   version "6.26.3"
   resolved "https://registry.yarnpkg.com/babel-core/-/babel-core-6.26.3.tgz#b2e2f09e342d0f0c88e2f02e067794125e75c207"
   dependencies:
```

however, the real fix is probably to upgrade danger - v3.0.4 in this repo is 6 majors behind

SimenB:
🎉

SimenB:
Rebased and squashed the extra commits to migrate devtools over. So now it's one commit migrating to v24, and the one commit I'm rather hesitant to have my name next to as a separate one for the hacky `react-is` workaround. Should make it easier to review at least!

(I often use this repo to test out breaking changes in Jest itself, especially around fake timers, which is why I wanna see it upgraded. 🙂 Makes it easier)

packages/react-devtools-extensions/package.json
```diff
@@ -25,7 +25,6 @@
     "archiver": "^3.0.0",
     "babel-core": "^7.0.0-bridge",
     "babel-eslint": "^9.0.0",
-    "babel-jest": "^24.7.1",
```
SimenB: these are just remnants after migrating it into this monorepo, I guess? They're unused

packages/react-devtools-extensions/package.json
```diff
@@ -25,7 +25,6 @@
     "archiver": "^3.0.0",
     "babel-core": "^7.0.0-bridge",
```
SimenB: this is quite possible also redundant now? not sure... but if it was here for jest@23, then it can go 🙂

SimenB:
@threepointone any news from your meeting?

SimenB:
@threepointone ping 🙂

threepointone:
I've approved, but could you rebase one last time and make sure nothing's broken? I don't expect it to be, but want to be certain before I hit merge :)

SimenB:
Awesome! I'll give it a rebase tonight or tomorrow

SimenB:
rebased, and CI still seems happy 👍

threepointone:
bing bang bosh, it's in!

SimenB:
woo, thanks! 🎉

threepointone:
No, thank you! We can finally start using inline snapshots yay 😃

PR #15779, opened 2019-05-30
chore: make tests compatible with Jest 24

SimenB:
This is #15778 without the actual upgrade to Jest 24 (as that breaks CI). I'm hopeful this will pass so it can land, then the diff for Jest 24 is just config changes.

Note that this does upgrade jest to the latest 23 release, which is why the lockfile has so many changes

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -521,7 +521,7 @@ const tests = {
       // Valid because the ref is captured.
       code: `
         function useMyThing(myRef) {
-          const myRef = useRef();
+          myRef = useRef();
           useEffect(() => {
             const handleMove = () => {};
             const node = myRef.current;
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -521,7 +521,6 @@ const tests = {
       // Valid because the ref is captured.
       code: `
         function useMyThing(myRef) {
-          myRef = useRef();
           useEffect(() => {
             const handleMove = () => {};
             const node = myRef.current;
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationElements-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationElements-test.js
@@ -651,7 +651,7 @@ describe('ReactDOMServerIntegration', () => {
       });
     });
 
-    describe('component hierarchies', async function() {
+    describe('component hierarchies', function() {
       itRenders('single child hierarchies of components', async render => {
         const Component = props => <div>{props.children}</div>;
         let e = await render(
--- a/packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
+++ b/packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
@@ -37,13 +37,7 @@ describe('ChangeEventPlugin', () => {
     ReactFeatureFlags = require('shared/ReactFeatureFlags');
     // TODO pull this into helper method, reduce repetition.
     // mock the browser APIs which are used in schedule:
-    // - requestAnimationFrame should pass the DOMHighResTimeStamp argument
     // - calling 'window.postMessage' should actually fire postmessage handlers
-    global.requestAnimationFrame = function(cb) {
-      return setTimeout(() => {
-        cb(Date.now());
-      });
-    };
     const originalAddEventListener = global.addEventListener;
     let postMessageCallback;
     global.addEventListener = function(eventName, callback, useCapture) {
--- a/packages/react/src/__tests__/ReactCoffeeScriptClass-test.coffee
+++ b/packages/react/src/__tests__/ReactCoffeeScriptClass-test.coffee
@@ -543,3 +543,4 @@ describe 'ReactCoffeeScriptClass', ->
     node = ReactDOM.findDOMNode(instance)
     expect(node).toBe container.firstChild
     undefined
+  undefined
--- a/scripts/jest/config.base.js
+++ b/scripts/jest/config.base.js
@@ -21,4 +21,11 @@ module.exports = {
   collectCoverageFrom: ['packages/**/*.js'],
   timers: 'fake',
   snapshotSerializers: [require.resolve('jest-snapshot-serializer-raw')],
+  // Jest changed from `about:blank` to `http://localhost` default in 24.5 (https://github.com/facebook/jest/pull/6792)
+  // in order to address https://github.com/facebook/jest/issues/6766. If one uses `about:blank` in JSDOM@11.12 or
+  // newer, it fails with `SecurityError: localStorage is not available for opaque origins`. However, some of React's
+  // tests depend on `about:blank` being the domain (for e.g. `url` in `img` tags). So we set `about:blank` here to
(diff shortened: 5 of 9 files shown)


scripts/jest/setupEnvironment.js
```diff
@@ -9,10 +9,6 @@ global.__PROFILE__ = NODE_ENV === 'development';
 global.__UMD__ = false;
 
 if (typeof window !== 'undefined') {
-  global.requestAnimationFrame = function(callback) {
```
SimenB: jsdom comes with raf

packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
```diff
@@ -36,13 +36,7 @@ describe('ChangeEventPlugin', () => {
     ReactFeatureFlags = require('shared/ReactFeatureFlags');
     // TODO pull this into helper method, reduce repetition.
     // mock the browser APIs which are used in schedule:
-    // - requestAnimationFrame should pass the DOMHighResTimeStamp argument
     // - calling 'window.postMessage' should actually fire postmessage handlers
-    global.requestAnimationFrame = function(cb) {
-      return setTimeout(() => {
-        cb(Date.now());
```
SimenB: the raf from jsdom passes in the timestamp. is this safe to remove?

SimenB:
@threepointone I've rebased this PR. I'd love to see it land, then landing the upgrade to Jest 24 should be simpler as the diff will be cleaner

scripts/jest/config.base.js
```diff
@@ -21,4 +21,5 @@ module.exports = {
   collectCoverageFrom: ['packages/**/*.js'],
   timers: 'fake',
   snapshotSerializers: [require.resolve('jest-snapshot-serializer-raw')],
+  testURL: 'about:blank',
```
SimenB: Note that doing this breaks in JSDOM@>=11.5 with `SecurityError: localStorage is not available for opaque origins` which is why the default has changed in Jest. Can try to do something for this when Jest 25 is released, which will force a newer version of JSDOM.

For now, I've manually made sure the entry in the lockfile is ~11.4.0, which does not have this issue

threepointone: Could you add a short TODO here?

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -521,7 +521,7 @@ const tests = {
       // Valid because the ref is captured.
       code: `
         function useMyThing(myRef) {
-          const myRef = useRef();
+          myRef = useRef();
```
threepointone: could you explain this change?

threepointone: (as in, to me :) I 'm not sure I understand why this const was removed.)

scripts/jest/config.base.js
SimenB: Done!

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
SimenB: `myRef` comes in as a parameter to the function. In babel 7 (and at runtime) it's a syntax error to redeclare a variable as const.

`Uncaught SyntaxError: Identifier 'myRef' has already been declared`.

This surfaces now because I've bumped `@babel/parser` as part of this PR (and `babel-eslint`, which uses that parser under the hood, is used as the parser for these eslint tests)

SimenB: The _wonderful_ error Jest outputs without this change:

image

scripts/jest/config.base.js
SimenB: ergh, expect it wasn't particularly short 😅

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
threepointone: Haha I missed that it was already an arg. Thanks!

SimenB: so an alternative in this case is to remove the parameter, but I'm not sure what this test asserts TBH

SimenB: that's the reason for all the changes in the eslint tests, btw

package.json
```diff
@@ -10,7 +10,6 @@
     "babel-code-frame": "^6.26.0",
     "babel-core": "^6.0.0",
     "babel-eslint": "^10.0.0",
-    "babel-jest": "^23.0.1",
```
SimenB: I mentioned this in the Jest 24 PR, but for completeness: you don't need this dependency as you implement your own transformer which does not use `babel-jest`: https://github.com/facebook/react/blob/c4f0b93703be3d165de78db7a381d6cca2203b2f/scripts/jest/preprocessor.js

scripts/jest/preprocessor.js
```diff
@@ -70,7 +70,7 @@ module.exports = {
               }
             : {}
         )
-      ).code;
```
SimenB: removing this ensures you return source maps to Jest, for better errors and debugging. This is why we now have `Foo` in some of the stack trace tests instead of the generic `Component`

package.json
SimenB: removed in #16297

SimenB:
@threepointone congrats on the 16.9 release 😀 I've rebased this after #16297 was merged

packages/react-dom/src/__tests__/ReactDOMServerIntegrationElements-test.js
```diff
@@ -651,7 +651,7 @@ describe('ReactDOMServerIntegration', () => {
       });
     });
 
-    describe('component hierarchies', async function() {
```
SimenB: `describe` must be `void` (in reality it must be synchronous) in Jest. `async` here was always an error, but it throws with jest 24.

https://jestjs.io/docs/en/troubleshooting#defining-tests

packages/react/src/__tests__/ReactCoffeeScriptClass-test.coffee
```diff
@@ -543,3 +543,4 @@ describe 'ReactCoffeeScriptClass', ->
     node = ReactDOM.findDOMNode(instance)
     expect(node).toBe container.firstChild
     undefined
+  undefined
```
SimenB: same as the `async` describe - the return value must be `void`.

SimenB:
Since I was messing with the jest setup, I also sent #16332 which seems to have uncovered some erroneously passing tests

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
threepointone: I checked, and this line could be removed completely. 
```suggestion
```

SimenB: perfect. the other eslint rule changes look alright?

threepointone:
I merged the other one before this one, and this now has a conflict with yarn.lock. Sorry! 😅If you could rebase and make sure the lockfile is fine, I'll merge this in. Thank you!

SimenB:
Rebased! 🙂

SimenB:
🎉

PR #15780, opened 2019-05-30
[FIX] - ReactDOM Add back version property to client

ealush:
### Change details
Adding back `.version` property to ReactDOM. It [used to exist in React15](https://github.com/facebook/react/blob/15-stable/src/renderers/dom/ReactDOM.js#L32) and got deleted. It still [exists in ReactDOM/server](https://github.com/facebook/react/blob/3b2302253f13c9cd049fac10784e10b6c582e3b6/packages/react-dom/src/server/ReactDOMServerNode.js#L22), and it appears to be [typed in @typed/react-dom](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-dom/index.d.ts#L25) so I am assuming it got deleted by mistake.

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -159,6 +159,8 @@ const ReactDOM: Object = {
       IsThisRendererActing,
     ],
   },
+
+  version: ReactVersion,
 };
 
 if (exposeConcurrentModeAPIs) {


ealush:
Ping:
@threepointone, @trueadm 
Can you take a look at it? It can really help when debugging version mismatch.

michasherman:
Any updates on this?

omrilotan:
Is this going to be added?

milesj:
Why is this necessary? Just do `require('react-dom/package.json').version`.

ealush:
@milesj, First, this is the browser interface, in the browser I don't have direct access to package.json. You can expose it, of course, but when debugging production applications this is not something you want to do.

Second, as I noted in the description of this pr All other interfaces (react, react-dom/server) still have it, and this is the only one where it doesn't exist. @typed also assumes it is there.

All these lead me to believe that the version key got deleted by mistake when moving from react 15 to 16.

ealush:
TBH I had no idea you could detect ReactDOM's version from devtools.

When I created the PR we were debugging a production site with multiple versions of React present on the same page - some with legacy React15 code and some with React16. We experienced a React/ReactDOM version mismatch which resulted in hooks errors.

When trying to debug it on the server both react and react-dom exposed their version, but in the client only react did so we couldn't really tell which version of React our code was using. The errors were not present in development.

Working on this issue I assumed it got deleted by mistake since all other APIs still maintained the version property.

gaearon:
Out in 16.13

PR #15796, opened 2019-05-31
[Events] Nested discrete events across systems

acdlite:
## Based on #15794

If an event in the old system is dispatched synchronously within an event from the new system, or vice versa, and the inner event is a discrete update, React should not flush pending discrete updates before firing the inner event's handlers, even if the outer event is not discrete.

Another way of saying this is that nested events should never force React to flush discrete updates.

Arguably, if the outer event is not a discrete event, then the inner event _should_ flush the pending events. However, that would be a breaking change. I would argue this isn't so bad, however, given that nested events are pretty rare. They don't fit nicely into our event model regardless, since we don't support nested React renders. In the future we should consider warning when events are nested.

### DIFF
--- a/packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
+++ b/packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
@@ -9,7 +9,7 @@
 
 'use strict';
 
-const React = require('react');
+let React = require('react');
 let ReactDOM = require('react-dom');
 let ReactFeatureFlags;
 let Scheduler;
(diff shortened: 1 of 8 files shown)


packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
```diff
@@ -748,5 +752,40 @@ describe('ChangeEventPlugin', () => {
       expect(ops).toEqual(['render: ']);
       expect(input.value).toBe('');
     });
+
+    xit('mouse enter/leave should be user-blocking but not discrete', async () => {
```
trueadm: xit

acdlite:
@trueadm I went through the test plan you gave me via chat and didn't encounter any issues.

PR #15797, opened 2019-06-01
Fix error logging in getDerivedStateFromProps

rickhanlonii:
## Overview

This PR fixes an bug in `getDerivedStateFromError` where errors are not logged when an error occurs.

## Motivation

In React Native, when using `componentDidCatch`, any error that is caught will display a RedBox and on dismiss show the error boundary:

cDC

When using `getDerivedStateFromError`, no RedBox is shown, only the error boundary:

gDSFE

With this fix, `getDerivedStateFromError` performs the same as `componentDidCatch` with a RedBox that dismisses to reveal the error boundary:

gDSFE_fixed

## Testing

I want to confirm the testing strategy before adding all of the tests in these files, so I've added two example tests:

- **ReactLegacyErrorBoundaries** test will pass without this change
- **ReactErrorBoundaries** test will **fail** without this change

### DIFF
--- a/packages/react-dom/src/__tests__/ReactErrorBoundaries-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactErrorBoundaries-test.internal.js
@@ -37,9 +37,14 @@ describe('ReactErrorBoundaries', () => {
   let RetryErrorBoundary;
   let Normal;
 
+  afterEach(() => {
+    console.error.mockRestore();
+  });
+
   beforeEach(() => {
     jest.useFakeTimers();
     jest.resetModules();
+    jest.spyOn(console, 'error');
     PropTypes = require('prop-types');
     ReactFeatureFlags = require('shared/ReactFeatureFlags');
     ReactFeatureFlags.replayFailedUnitOfWorkWithInvokeGuardedCallback = false;
@@ -653,6 +658,13 @@ describe('ReactErrorBoundaries', () => {
       container,
     );
     expect(container.firstChild.textContent).toBe('Caught an error: Hello.');
+    expect(console.error).toHaveBeenCalledWith(
+      __DEV__
+        ? expect.stringMatching(
+            /^The above error occurred in the <BrokenRender> component/,
+          )
+        : new Error('Hello'),
+    );
     expect(log).toEqual([
       'ErrorBoundary constructor',
       'ErrorBoundary componentWillMount',
--- a/packages/react-dom/src/__tests__/ReactErrorBoundaries-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactErrorBoundaries-test.internal.js
@@ -37,14 +37,9 @@ describe('ReactErrorBoundaries', () => {
   let RetryErrorBoundary;
   let Normal;
 
-  afterEach(() => {
-    console.error.mockRestore();
-  });
-
   beforeEach(() => {
     jest.useFakeTimers();
     jest.resetModules();
-    jest.spyOn(console, 'error');
     PropTypes = require('prop-types');
     ReactFeatureFlags = require('shared/ReactFeatureFlags');
     ReactFeatureFlags.replayFailedUnitOfWorkWithInvokeGuardedCallback = false;
@@ -649,6 +644,39 @@ describe('ReactErrorBoundaries', () => {
     expect(container3.firstChild).toBe(null);
   });
 
+  it('logs a single error when using error boundary', () => {
+    const container = document.createElement('div');
+    expect(() =>
+      ReactDOM.render(
+        <ErrorBoundary>
+          <BrokenRender />
+        </ErrorBoundary>,
+        container,
+      ),
+    ).toWarnDev('The above error occurred in the <BrokenRender> component:', {
+      logAllErrors: true,
+    });
+
+    expect(container.firstChild.textContent).toBe('Caught an error: Hello.');
+    expect(log).toEqual([
+      'ErrorBoundary constructor',
+      'ErrorBoundary componentWillMount',
+      'ErrorBoundary render success',
+      'BrokenRender constructor',
+      'BrokenRender componentWillMount',
+      'BrokenRender render [!]',
+      // Catch and render an error message
+      'ErrorBoundary static getDerivedStateFromError',
+      'ErrorBoundary componentWillMount',
+      'ErrorBoundary render error',
+      'ErrorBoundary componentDidMount',
+    ]);
+
+    log.length = 0;
+    ReactDOM.unmountComponentAtNode(container);
+    expect(log).toEqual(['ErrorBoundary componentWillUnmount']);
+  });
+
   it('renders an error state if child throws in render', () => {
     const container = document.createElement('div');
     ReactDOM.render(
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/__tests__/ReactErrorBoundaries-test.internal.js
```diff
@@ -653,6 +657,13 @@ describe('ReactErrorBoundaries', () => {
       container,
     );
     expect(container.firstChild.textContent).toBe('Caught an error: Hello.');
+    expect(console.error).toHaveBeenCalledWith(
```
gaearon: Does our `toWarnDev()` matcher work for this case?

packages/react-reconciler/src/ReactFiberThrow.js
```diff
@@ -102,6 +102,7 @@ function createClassErrorUpdate(
   if (typeof getDerivedStateFromError === 'function') {
     const error = errorInfo.value;
     update.payload = () => {
+      logError(fiber, errorInfo);
```
gaearon: What if I have _both_ gDSFE and cDC? We'll want to log once.

packages/react-dom/src/__tests__/ReactErrorBoundaries-test.internal.js
```diff
@@ -653,6 +657,13 @@ describe('ReactErrorBoundaries', () => {
...
+      __DEV__
+        ? expect.stringMatching(
+            /^The above error occurred in the <BrokenRender> component/,
+          )
+        : new Error('Hello'),
+    );
```
rickhanlonii: @bvaughn looks like you added the tests here, what do you think about this strategy? I copied this from elsewhere, toWarnDev doesn't seem to work for this:

```js
expect(() => {
  ReactDOM.render(
    <ErrorBoundary>
      <BrokenRender />
    </ErrorBoundary>,
    container,
  );
}).toWarnDev('The above error occurred in the <BrokenRender> component');
```

Here's the error (with added logs of what was actually logged)

Screen Shot 2019-06-01 at 1 31 04 AM

acdlite:
The idea behind this design is that you should be able to catch an error and gracefully recover. Not all errors should result in a red box.

packages/react-dom/src/__tests__/ReactErrorBoundaries-test.internal.js
rickhanlonii: I don't think so, see https://github.com/facebook/react/pull/15797#discussion_r289583960

rickhanlonii:
Ah yes of course, that makes sense. cc @gaearon

acdlite:
Didn't mean to shut down the discussion! I can imagine an argument for why we'd want to provide a way to observe all errors, even if they are caught. But idk if `logError` is the best mechanism.

acdlite:
For example, the Chrome DevTools has both "Break on exceptions" and "Break on caught exceptions." Could RN have some way to configure a similar setting?

gaearon:
Let's leave this open until we figure out some solution. I don't think this is RN-specific either.

In general presence of error boundary shouldn't silence unintentional exceptions IMO. At least things like TypeError or ReferenceError etc.

packages/react-reconciler/src/ReactFiberThrow.js
bvaughn: Seems like it would be worth calling `logError` in this case, so long as we check that cDC doesn't also exist.

bvaughn:
> In general presence of error boundary shouldn't silence unintentional exceptions IMO. At least things like TypeError or ReferenceError etc.

I'm not sure that quite describes what's happening here. An error boundary won't _swallow_ an error. The difference is just the additional DEV-only console logging.

That being said, I think adding a call to `logError` in gDSFE (so long as the component doesn't also defined cDC) makes sense to me.

gaearon:
We’ve been very intentional in the past about not relying on user-written error boundaries to be good, and showing the environment-specific dialog (www, CRA, RN) anyway.

I’m not saying this can’t be revisited but I’m a little surprised we decided to make a new method the point where it behaves differently. Since that method has been the recommended migration instruction from the old one for the fallback case. But I don’t think we discussed that we’re relaxing our earlier intentional stance that we don’t fully “trust” user written boundaries in DEV. If we’re intentionally trusting them more let’s chat more about it.

bvaughn:
I'm confused by your response, Dan. I suggested we add the call to cover gDSFE as well...

gaearon:
I don’t think we disagree 🙂 My response was mostly to Andrew’s point — but also that I do consider this “swallowing” in the sense that we were very intentionally doing that DEV call in the past.

bvaughn:
Gotcha! I think that not calling the `logError` method was an oversight on our part. I just wanted to clarify that we aren't "swallowing" an error in that...the app will not silently fail. 😁

sebmarkbage:
@gaearon Want to make sure we reopen this or something like it?

Logging as a side effect in prod should only rely on cDU.

This, however, is for showing an informative error that something went wrong that you should probably fix.

We should make sure that this doesn’t trigger dialogs but instead just an expandable bar or something like that doesn’t cover the UI. A developer must be able to see what the user would see otherwise it can lead to bad bugs where the resulting UI is something that the user can interact with but the developer would never go/see past the error. We’d had such bugs in the past.

That way it is a bit more informative than fatal. That can be useful for recoverable true errors like I/O failing.

We’ve also talked about gDSFE being used for control flow. For example `throw Redirect();` deep in the tree when you realize that you can’t actually show something for a particular ID.

The problem with this technique is that it’s not actually useful today because any intermediate boundary would catch it. So we need some kind of “matching” semantics which allows these special throws to by-pass catch all intermediates.

If we’re codifying that anyway then we can disambiguate errors from “algebraic effects”.

I.e. this particular path would always be true errors and should call logError.

packages/react-dom/src/__tests__/ReactErrorBoundaries-test.internal.js
```diff
@@ -644,6 +644,39 @@ describe('ReactErrorBoundaries', () => {
...
+    const container = document.createElement('div');
+    expect(() =>
+      ReactDOM.render(
+        <ErrorBoundary>
+          <BrokenRender />
+        </ErrorBoundary>,
+        container,
+      ),
+    ).toWarnDev('The above error occurred in the <BrokenRender> component:', {
+      logAllErrors: true,
```
rickhanlonii: Added a new option for `toWarnDev` which will log all errors

Without this, `toWarnDev` skips "The above error" errors [here](https://github.com/facebook/react/blob/b2adcfba32d133aabdbc182bba1001e729060abe/scripts/jest/shouldIgnoreConsoleError.js#L11)

packages/react-dom/src/__tests__/ReactLegacyErrorBoundaries-test.internal.js
```diff
@@ -486,6 +487,57 @@ describe('ReactLegacyErrorBoundaries', () => {
       },
     };
 
+    BothErrorBoundaries = class extends React.Component {
```
rickhanlonii: One of the ErrorBoundaries-test files needed to have an error boundary that used both the old and new methods, I chose the Legacy test because if Legacy is removed the test can be deleted

packages/react-reconciler/src/ReactFiberThrow.js
```diff
@@ -119,10 +119,12 @@ function createClassErrorUpdate(
         // TODO: Warn in strict mode if getDerivedStateFromError is
         // not defined.
         markLegacyErrorBoundaryAsFailed(this);
+
+        // Only log here if componentDidCatch is the only error boundary method defined
+        logError(fiber, errorInfo);
```
rickhanlonii: Move this inside this block to prevent double logging 👍

gaearon:
Thanks.

gaearon:
Filed spinoffs: https://github.com/facebook/react/issues/15989 and https://github.com/facebook/react/issues/15988.

PR #15806, opened 2019-06-03
chore: use jest-serializer-raw for react-fresh snapshots

SimenB:
As mentioned in https://github.com/facebook/react/pull/15733/files#r289843934. I personally find the snapshots easier to read without the escaped junk

Zero hard feelings if you disagree, this took less than 5 minutes to put together 🙂

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "packages/*"
   ],
   "devDependencies": {
+    "@mattiasbuelens/web-streams-polyfill": "0.1.0",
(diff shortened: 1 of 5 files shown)


gaearon:
Why isn't this the default?

SimenB:
Not sure... There is https://github.com/facebook/jest/pull/2482 which implemented the current behavior and https://github.com/facebook/jest/pull/4183 which tried to fix it, although I don't remember why it stalled

PR #15811, opened 2019-06-04
[Events] Add support for events that are both user-blocking and continuous

acdlite:
Some events, like hover, are more important than the default, but they are not discrete (i.e. they don't need to be processed serially).

I've only implemented this in the legacy event system. I'll leave Flare for a follow-up.

Absent resuming, I'm also not sure if this will cause starvation problems, so I put it behind a feature flag.


trueadm:
Note: for the Flare `hover` events we dispatch them as discrete events here:

https://github.com/facebook/react/blob/master/packages/react-events/src/Hover.js#L89

PR #15816, opened 2019-06-04
allow nested `act()`s from different renderers

threepointone:
Sometimes apps can be tested with multiple renderers at once. Examples - 
- react-art being used inside react-dom. This was 'fixed' by disabling missing act warnings for embedded renders https://github.com/facebook/react/pull/15975
- ReactDOM.render being used inside another component tree. The parent component will be rendered using ReactTestRenderer.create for a snapshot test or something.
- a ReactDOM instance interacting with a ReactTestRenderer instance (like for the new devtools)

This PR is for the other 2 usecases. It replaces the sigil reference with a stack that holds all sigils as act scopes open/close, and compares against the stack for when to show the missing act() warning.
 
Open question - At what point should flushPassiveEffects be called? On exiting the last act of any type? Or for a given type?

### DIFF
--- a/fixtures/dom/src/index.test.js
+++ b/fixtures/dom/src/index.test.js
@@ -14,7 +14,6 @@ import ARTSVGMode from 'art/modes/svg';
 import ARTCurrentMode from 'art/modes/current';
 import TestUtils from 'react-dom/test-utils';
 import TestRenderer from 'react-test-renderer';
-
 ARTCurrentMode.setCurrent(ARTSVGMode);
 
 global.__DEV__ = process.env.NODE_ENV !== 'production';
@@ -156,5 +155,13 @@ it('does not warn when nesting react-act inside react-dom', () => {
 it('does not warn when nesting react-act inside react-test-renderer', () => {
   TestRenderer.act(() => {
     TestRenderer.create(<ARTTest />);
+  });
+});
+
+it("doesn't warn if you use nested acts from different renderers", () => {
+  TestRenderer.act(() => {
+    TestUtils.act(() => {
+      TestRenderer.create(<App />);
+    });
   });
 });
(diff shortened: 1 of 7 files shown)


threepointone:
^ need to fix CI since I'm using toWarnDev in the fixture, will probably just fork it into the fixtures folder

fixtures/dom/src/index.test.js
```diff
@@ -11,21 +11,12 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import TestUtils from 'react-dom/test-utils';
 import TestRenderer from 'react-test-renderer';
+global.__DEV__ = process.env.NODE_ENV !== 'production';
 
-let spy;
-beforeEach(() => {
-  spy = jest.spyOn(console, 'error').mockImplementation(() => {});
+expect.extend({
```
threepointone:
```suggestion
expect.extend(require('../../../scripts/jest/matchers/toWarnDev'))
```

packages/react/src/ReactCurrentActingRendererSigil.js
```diff
@@ -8,12 +8,12 @@
...
- * We use a renderer's flushPassiveEffects as the sigil value
- * so we can track identity of the renderer.
+ * We maintain a 'stack' of renderer specific sigils
+ * corresponding to act() calls, so we can track whenever an update
+ * happens inside/outside of one.
  */
 
 const ReactCurrentActingRendererSigil = {
-  current: (null: null | (() => boolean)),
+  current: ([]: Array<{}>),
```
acdlite: This is conceptually a Set, not an Array, so (assuming this is the approach we go with) let's make it an Array.

threepointone: I started this with a Set, until I realized it's *not* one. It's a stack. Consider when leaving an act() scope, how could I know if the sigil has to be removed from the set?

threepointone:
Open question - At what point should flushPassiveEffects be called? On exiting the last act of any type? Or for a given type?

threepointone:
abandoning this for https://github.com/facebook/react/pull/16039

PR #15819, opened 2019-06-04
[Flare] Add event position properties to Hover responder

trueadm:
This PR adds mouse event co-ords to Hover responder module events. Specifically, `clientX`, `clientY`, `pageX`, `pageY`, `x` and `y`. These are all taken from the native event object.

### DIFF
--- a/packages/react-events/src/Hover.js
+++ b/packages/react-events/src/Hover.js
@@ -43,6 +43,14 @@ type HoverEvent = {|
   target: Element | Document,
   type: HoverEventType,
   timeStamp: number,
+  clientX: null | number,
+  clientY: null | number,
+  pageX: null | number,
+  pageY: null | number,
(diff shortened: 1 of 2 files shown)


packages/react-events/src/Hover.js
```diff
@@ -43,6 +43,14 @@ type HoverEvent = {|
   target: Element | Document,
   type: HoverEventType,
   timeStamp: number,
+  clientX: null | number,
```
necolas: If we don't need to support existing use of `clientX` (for hover) at FB maybe we shouldn't include it on the event objects

trueadm: True, but I wasn't 100% sure if it wouldn't be used. I'll leave it in for now, we can take a stab at removing a bunch of these later and using higher level versions possibly.

PR #15823, opened 2019-06-04
[Flare] EventPriority enum

acdlite:
Same as #15811 but for Flare.

I don't know what the public API for setting the event priority should be. Right now it accepts a numeric enum, but is this what we want? Maybe it should be a string enum? I've punted on this until I receive feedback.


acdlite:
Another option is to keep the `isDiscrete` option to `dispatchEvent` as-is, and call `Scheduler.runWithPriority` directly.

necolas:
lgtm

PR #15834, opened 2019-06-06
[Flare] Fix ES6 issues with IE11

trueadm:
This PR fixes two issues encountered with Flare when used on IE11:

- There can be bad Map/Set polyfills used that try to add properties to the objects when used as keys. If the object that is used as a key is non-extensible (it was frozen via `Object.freeze`) then this will cause errors. We already checked this in other parts of the codebase, so we just need to apply the same check here before we make `responder` frozen.
-  `Symbol` might not be polyfilled, which was previously required by React Flare. We didn't actually need to use this though, as we can use incrementing integers instead (this is what native `setTimeout` etc do in browsers).


packages/react/src/ReactCreateEventComponent.js
```diff
@@ -11,12 +11,34 @@ import {enableEventAPI} from 'shared/ReactFeatureFlags';
...
+
+if (__DEV__) {
+  hasBadMapPolyfill = false;
+  try {
+    const frozenObject = Object.freeze({});
+    const testMap = new Map([[frozenObject, null]]);
+    const testSet = new Set([frozenObject]);
+    // This is necessary for Rollup to not consider these unused.
+    // https://github.com/rollup/rollup/issues/1771
+    // TODO: we can remove these if Rollup fixes the bug.
```
bvaughn: IIR the fix for this landed this week so we can remove it soon 😄

526117:
ในวันที่ ศ. 7 มิ.ย. 2019 00:45 Dominic Gannaway <notifications@github.com>
เขียนว่า:

PR #15838, opened 2019-06-07
Update noop to closure function to up speed

Himself65:
```js
function noop1() {

}

const noop2 = () => {
}

console.time('test1')
for (let i = 0; i < 500000; i++) {
  noop1()
}
console.timeEnd('test1')


console.time('test2')
for (let i = 0; i < 500000; i++) {
  noop2()
}
console.timeEnd('test2')
```

```
test1: 5.570068359375ms
test2: 2.18701171875ms
```

### DIFF
--- a/packages/react/src/withComponentStack.js
+++ b/packages/react/src/withComponentStack.js
@@ -7,7 +7,7 @@
 
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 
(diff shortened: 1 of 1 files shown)


trueadm:
This is a rather inefficient micro-benchmark. If you switch noop2 to run first, rather than noop1, you will see the opposite results.

PR #15840, opened 2019-06-07
[Events] Adds existing event system behind enablePluginEventSystem flag

trueadm:
_Note: this PR should make no changes to existing logic anywhere. Everything should continue to work as it does today. Also we don't expect to turn this flag of for the long foreseeable future._

This PR takes the existing event system call-sites and modules and wraps them in a `enablePluginEventSystem` flag. With this flag in place we can understand internally what parts of our application code use what events parts of the heritage event system. This flag also allows us to track metrics on code size of React Flare vs the heritage event system, as code size is important for us to manage.

This PR also renames the existing React Flare event flag to something more descriptive - `enableResponderEventSystem`.


trueadm:
I need to look into why the sizes have gone up. It's likely something with GCC and the fact there are now flags there that weren't there before. @sebmarkbage any ideas, I think you ran into something similar in the past where inlining changed because of flags?

necolas:
Closing as we're no longer developing this API. See https://github.com/facebook/react/issues/15257#issuecomment-565408356

PR #15843, opened 2019-06-08
Fix useCallback not work properly in shallow renderer

### LINKED ISSUE #15774: useCallback doesn't memoize callback in shallow renderer

bug

In shallow renderer, The returned callback from `useCallback` is not the same one between two rerendering even though the dependencies are the same. But `useMemo` will work as expected.

For example, the following test won't work:

```js
    function SomeComponent() {
      const noop = React.useCallback(() => {}, []);

      return (
        <div onClick={noop} />
      );
    }

    let firstResult = shallowRenderer.render(<SomeComponent />);
    let secondResult = shallowRenderer.render(<SomeComponent />);

    expect(firstResult).toEqual(secondResult);
```

If `useCallback` returned the same callback between two rendering (I think) the assertion should pass.

but the same (almost) logic will work with `useMemo`, in the shallow renderer test suite (See https://github.com/facebook/react/blob/master/packages/react-test-renderer/src/__tests__/ReactShallowRendererHooks-test.js#L273-L291).

I created a branch in https://github.com/chenesan/react/tree/usecallback-not-work-properly-in-shallow-renderer and you can run the failed shallow renderer test in https://github.com/chenesan/react/commit/b2dff28909247342cfcb9ced2fbe24a0f9ca08b7#diff-d9a78422c03941578ae9ba487e8132cb .

`useCallback` should return the memoized callback when the dependencies unchanged even in shallow renderer.

I tried to look into this and I found out that in shallow renderer `useCallback` just returns the original callback argument (See https://github.com/facebook/react/blob/master/packages/react-test-renderer/src/ReactShallowRenderer.js#L365-L371) but `useMemo` will compare the dependencies between rendering. I'm not sure if it's intended (So it's expected that `useCallback` will not memoize callback in shallow renderer). If it's triaged as a bug I'm glad to send a pr for this :)

react@16.8.6

chenesan:
Fix #15774 .
Just copy and paste the same logic as `useMemo` in shallow renderer; Not sure if it's better to extract the comparison logic into common function or just keep it repeat.

### DIFF
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -337,7 +337,7 @@ class ReactShallowRenderer {
 
     const useCallback = <T>(
       callback: T,
-      deps: Array<mixed> | void | null
+      deps: Array<mixed> | void | null,
     ): T => {
       this._validateCurrentlyRenderingComponent();
       this._createWorkInProgressHook();
@@ -361,7 +361,7 @@ class ReactShallowRenderer {
       (this._workInProgressHook: any).memoizedState = [nextValue, nextDeps];
       return nextValue;
     };
-        
+
     const useRef = <T>(initialValue: T): {current: T} => {
       this._validateCurrentlyRenderingComponent();
       this._createWorkInProgressHook();
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -335,6 +335,33 @@ class ReactShallowRenderer {
       return nextValue;
     };
 
+    const useCallback = <T>(
+      callback: T,
+      deps: Array<mixed> | void | null
+    ): T => {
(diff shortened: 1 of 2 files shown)


chenesan:
@threepointone @gaearon sorry for pinging. any news on this?

threepointone:
I don't think we're going to add this functionality to the shallow renderer for now, sorry. I truly appreciate the effort.

ljharb:
@threepointone why not?

threepointone:
Sorry, I think I read the conclusion on the linked task wrongly. I'll reopen this, but I'm still not certain what our plans are for adding this to the shallow renderer.

ljharb:
This isn't stale, it's just that collaborators haven't made a decision yet.

(Please don't use stalebots, they're incredibly user-hostile)

gaearon:
The shallow renderer has moved to https://github.com/NMinhNguyen/react-shallow-renderer. If this is still relevant, please feel free to file this issue there. Thanks!

PR #15848, opened 2019-06-09
[Flare] Add Focus within support

devongovett:
This adds a `within` prop to the `Focus` event component, which makes events fire when descendants are focused and blurred in addition to the immediate child. This allows additional flexibility when building more complex components. The `within` name aligns with the CSS [:focus-within](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within) pseudo class.

For example, in a Date Picker component I'm working on, the focus ring CSS class is added to an outer element when one of the descendent elements is selected. Doing it with one `Focus` component wrapper is easier than wrapping each individual focusable element, propagating the focus state (and focus visible state) upwards and maintaining state manually on the common parent.

Screen Shot 2019-06-09 at 3 42 42 PM

Screen Shot 2019-06-09 at 3 43 52 PM

As you can see in the DOM structure above, the div with `role="combobox"` gets the focus ring class applied to it, but the individual divs with `role="spinbutton"` are actually focusable. This PR allows me to wrap the entire combobox in a `Focus` component with the `within` prop set in order to apply the focus ring class rather than each individual spinbutton.


devongovett:
We'd also need an `onFocusVisibleWithinChange` then too, along with `onFocusWithin` and `onBlurWithin`?

devongovett:
My thought on that was you probably need either one or the other, not both. If your child is directly focusable, then you want focus events for that immediate child, otherwise you might want focus events for descendants. That's why I did it as a prop rather than a whole new set of events. But willing to change it if you think it's clearer.

necolas:
I think this feature needs to be able to provide an equivalent to `:focus-within` styling without altering the existing semantics around "focus".

Here's an example of how I think it should work: https://codesandbox.io/s/experimental-event-api-n32y8

devongovett:
What if you want to combine focus within and focus visible? i.e. in the date picker example I showed above, the focus ring gets bolder on focus visible. But the immediate child of the `<Focus>` is not focusable itself. It would be nice to combine the two states in this case such that a class could be applied when focus is visible within the component.

The current PR allows that by changing the behavior of the events to apply to descendants so that I can handle `onFocusChange` to add the initial focus ring, and `onFocusVisibleChange` to add the bolder focus ring. As far as I can tell this wouldn't be possible without adding an additional `onFocusVisibleWithinChange` event or the `within` option I added. The `onFocusWithinChange` and `onFocusVisibleChange` events cannot be combined, because `onFocusVisibleChange` would never fire on the parent since it is not directly focusable.

FWIW, I don't think this is directly possible in CSS either (combining :focus-visible and :focus-within).

necolas:
> What if you want to combine focus within and focus visible?

That's not what the patch is doing though; it's changing what the current API means entirely without providing `focusWithin` support. I think we should provide a means to add the `focusWithin` semantics (as I did for `focusVisible`), but `onFocusVisibleChange` should not be called when elements arbitrarily deep in the subtree are focused.

If you want to change the styles on your combobox based on whether one of its children is visibly focused you can track that state across multiple `Focus` components. If there are use cases where that becomes too complex, we could either add new callbacks for the different concepts or build a new event component for the purpose (e.g., `FocusScope` is distinct from `Focus`)

devongovett:
It's opt in, but sure, it does change the behavior when you opt in. What about passing both pieces of info to a single callback? i.e. `onFocusChange` could receive both whether or not the immediate child is focused, along with whether global focus visible state is set. Same thing for `onFocusWithinChange`.

devongovett:
Sorry, hit the close button by accident. `onFocusWithinChange` is fine with me, I just have no way of knowing whether the global focus visible state is set.

necolas:
At one point I kicked around the idea of making the "change" event handlers take a second argument for meta data. A bit like this:

```
const [ focused, updateFocus ] = useState(null)
const [ focusVisible, updateFocusVisible ] = useState(null)

<Focus
  onFocusChange={(bool, info) => {
    updateFocus(bool);
    updateFocusVisible(info.pointerType === 'keyboard');
  }}
>
  <button style={focusVisible ? styles.focusVisible : null} />
</Focus>
```

It was focus visibility that got me thinking about doing this. But the state of focus visibility can change independent of focus, i.e., mouse click after keyboard focus. That leaves the UI stuck in the focus visible state when it shouldn't be. (You can see this bug on mobile.twitter.com, e.g., the top nav links) There would be a similar limitation to an API like `const onFocusWithinChange = (bool, { focusVisible }) => {}`. That event doesn't fire every time focus moves within the subtree. So you would need `onFocusVisibleWithinChange` to correctly expose the state.

devongovett:
Yeah... maybe `onFocusChange` could pass an object argument with the state: `onFocusChange({ isFocused: bool, isFocusWithin: bool, isFocusVisible: bool })`. If any of those properties changed, the event would fire again.

necolas:
It would still need `isFocusVisibleWithin` too. What you're suggest is a variant on the API I described, but bundled into a single callback that is getting called a lot, rather than individual callbacks:

```
<Focus onFocusChange={(
  isFocused, isFocusVisible, isFocusWithin, isFocusVisibleWithin
) => {
  updateFocus(isFocused);
  updateFocusVisible(isFocusVisible);
  updateFocusWithin(isFocusWithin);
  updateFocusVisibleWithin(isFocusVisibleWithin)
}} />

// vs

<Focus
  onFocusChange={updateFocus}
  onFocusVisibleChange={updateFocusVisible}
  onFocusWithinChange={updateFocusWithin}
  onFocusVisibleWithinChange={updateFocusVisibleWithin}
/>
```

devongovett:
Focus visibility is global state, so I think `isFocusVisible` would be the same across all `Focus` components. So you could determine if focus was visible within with `isFocusWithin && isFocusVisible`, and if focus is visible on the direct child with `isFocused && isFocusVisible`.

The downside to this API as you said is that it might be called a lot more than needed for a particular usecase, e.g. if you don't use focus visibility or focus within state. You seem pretty convinced that separate callbacks is the way to go, so I'm happy to go that way as well. Just wanted to make sure the alternatives were explored.

necolas:
> Focus visibility is global state, so I think isFocusVisible would be the same across all Focus components. 

Yes focus visibility is global but we're interested in whether a particular component should visibly display focus state. And that's what `:focus-visible` is about.

> So you could determine if focus was visible within with isFocusWithin && isFocusVisible, and if focus is visible on the direct child with isFocused && isFocusVisible.

Not quite, because `isFocusWithin` is true when `isFocused` is true.

> You seem pretty convinced that separate callbacks is the way to go

I think I've pointed out the problems with the alternative suggestions. The combobox can even be implemented with the existing `<Focus>` API using multiple components, but I think the concept of `focus-within` is worth adding and we can do that without changing everything else.

devongovett:
Ok I've updated the PR to implement `onFocusWithinChange` and `onFocusVisibleWithinChange`.

packages/react-events/src/Focus.js
```diff
@@ -224,11 +291,13 @@ const FocusResponder = {
...
       isFocused: false,
       isLocalFocusVisible: false,
+      isFocusWithin: false,
+      isFocusWithinVisible: false,
       pointerType: '',
     };
   },
   allowMultipleHostChildren: false,
-  stopLocalPropagation: true,
+  stopLocalPropagation: false,
```
necolas: This shouldn't be necessary

packages/react-events/src/__tests__/Focus-test.internal.js
```diff
@@ -258,17 +261,63 @@ describe('Focus event responder', () => {
...
+    it('is called after "blur" and "focus" events immediate child', () => {
+      ref.current.dispatchEvent(createFocusEvent('focus'));
+      expect(onFocusWithinChange).toHaveBeenCalledTimes(1);
+      expect(onFocusWithinChange).toHaveBeenCalledWith(true);
+      ref.current.dispatchEvent(createFocusEvent('blur'));
+      expect(onFocusWithinChange).toHaveBeenCalledTimes(2);
+      expect(onFocusWithinChange).toHaveBeenCalledWith(false);
+    });
+
+    it('is called after "blur" and "focus" events on descendants', () => {
```
necolas: There should be test coverage to ensure the callback is not called every time focus moves within the subtree

packages/react-events/src/__tests__/Focus-test.internal.js
```diff
@@ -351,6 +471,46 @@ describe('Focus event responder', () => {
         'inner: onFocusChange',
       ]);
     });
+
+    it('propagates focus within events', () => {
```
necolas: events -> components

This isn't what we discussed so I'd like to understand why we should change event propagation for this component

packages/react-events/src/Focus.js
devongovett: The problem was if you had nested `<Focus>` components. Then focus within wouldn't work on the outer one since event propagation was stopped at the inner one.

packages/react-events/src/__tests__/Focus-test.internal.js
devongovett: See previous comment about propagation change

packages/react-events/src/Focus.js
trueadm: That's the desired behaviour here. We don't want <Focus> to propagate ideally.

devongovett: Normal focus events aren't triggered on parents because of the check that the event target equals the immediate child of the focus component. But with focus within, I presume it should work like the `:focus-within` CSS selector, which is true when any descendant is focused. If there are nested `<Focus>` components, that would break the way focus within works and prevent focus within from being triggered since events are stopped at the inner most <Focus>.

trueadm: That makes sense, but you're changing the rest of how `<Focus>` works by changing `stopLocalPropagation`. You could make `stopLocalPropagation` `false`, but then you'd need to add additional logic into the responder callbacks to properly ensure local propagation is stopped. `FocusScope` actually gives an example of doing this by taking over responder ownership via `requestResponderOwnership`, which would block other event components of the same responder type from being propagated to.

devongovett: Ok... I can look. Should there be tests for that behavior? I assumed that since no tests failed when changing that setting that it was still working correctly. Also assumed it was just `event.stopPropagation()` but I guess that's also not true.

trueadm: `event.stopPropagation` doesn't mean anything to React as we don't use DOM propagation. We should add more test coverage to `Focus` as it's a bit lacking.

necolas: Focus has logic that ignores events if they aren't dispatched on the currentTarget. So I don't think there's missing test coverage or that managing ownership will be necessary. Changing local propagation would only make a difference for focus-within, it's just about which option makes the most sense.

For example, Hover intentionally doesn't pierce event component boundaries even though you might think it should at first. I mentioned in the prior discussion that we could in principle support both behaviours for focus-within, without altering the behaviour of any other callbacks either way. I'm sympathetic to the behaviour you chose to implement here by default

I'm not feeling well today but will test this PR out internally soon. Thanks!

trueadm: @necolas The issue is that if you nest `<Focus>` event components with `stopLocalPropagation` set to `false:

```jsx
<Focus>
  <div>
    <Focus>
      <span>
```
Then trigger a focus on the `<span>`, then inner `<div>` will be the `currentTarget`, not the outer `<span>`.

necolas: The behaviour of everything we've currently implemented should remain unchanged.

necolas:
I think this implementation has a bug with `onFocusVisibleWithinChange`, as it is getting called twice (false and true) when focus moves within the subtree. It shouldn't be called again until focus moves out of the subtree altogether or a non-keyboard pointer type is used. There's no test coverage in the PR for this scenario so I suspect you'll find tests fail when you add them. Demo: https://codesandbox.io/s/experimental-event-api-03xu4

packages/react-events/src/Focus.js
devongovett: So do you recommend following @trueadm's advice above about using responder ownership similar to `FocusScope` to implement this? How do I verify that it's working?

necolas:
I don't think you need to track isFocusVisibleWithin separately on state either

packages/react-events/src/Focus.js
trueadm: I don't think we should change this. Let's leave this the same as it is for now. I want to think more on how nested <Focus> and other components should be handled, but I don't think that work should be coupled with this PR.

necolas:
@devongovett Are you planning to update this PR or should we pick it up from here?

devongovett:
Yeah sorry, just haven't had a chance yet. I'll try to get to it in the next couple days.

devongovett:
@necolas @trueadm Updated:

* Left behavior of nested Focus event responders as is. Focus within with nested Focus components is now unsupported.
* Fixed bug where `onFocusVisibleWithinChange` was called as focus moved within the subtree.
* Stopped tracking `isFocusVisibleWithin` separately in the state.

necolas:
Latest patch seems to have regressed how focus is calculated https://codesandbox.io/s/experimental-event-api-03xu4

packages/react-events/src/__tests__/Focus-test.internal.js
```diff
@@ -258,17 +261,76 @@ describe('Focus event responder', () => {
       expect(onFocusChange).toHaveBeenCalledTimes(2);
       expect(onFocusChange).toHaveBeenCalledWith(false);
     });
+
+    it('is not called after "blur" and "focus" events on nested descendants', () => {
```
necolas: This is incorrect behaviour

devongovett: Why? It is verifying that `onFocusChange` is not called when focusing a descendant rather than the immediate child of the `<Focus>` component?

necolas: Oh I missed that it wasn't starting with the root being focused. Given the UX is quite broken by the patch it's odd that no tests failed.

devongovett: Yeah, I'll look into it and add more tests.

526117:
ในวันที่ จ. 10 มิ.ย. 2019 05:54 น. Devon Govett <notifications@github.com>
เขียนว่า:

necolas:
Closing this as I've implemented focus-within as a separate responder in #16152.

PR #15849, opened 2019-06-10
[Flare] First pass at an implementation of experimental FocusManager API

devongovett:
This is a first pass at implementing the `FocusManager` API specified in https://github.com/reactjs/rfcs/pull/109. It goes along with the already implemented `FocusScope` in react-events.

I adjusted the API slightly, and will be updating the RFC to match. In particular, the `focusNext` and `focusNextTabStop` methods have been merged, along with the corresponding previous methods. There is now an options object that can be passed to specify whether the element must be tabbable or only focusable. In addition, there are options to enable wrapping behavior at the ends of a scope, along with starting the search from a particular element rather than only `document.activeElement`. The methods also return the element they focused if any, which allows the calling code to do something with that information.

The API is now:

```javascript
type FocusManagerOptions = {
  from?: HTMLElement,
  tabbable?: boolean,
  wrap?: boolean,
};

interface FocusManager {
  focusNext(opts: FocusManagerOptions): HTMLElement,
  focusPrevious(opts: FocusManagerOptions): HTMLElement
}
```

Perhaps we will also want `focusFirst` and `focusLast` methods at some point as well. Happy to hear whatever feedback you have on the API.

Most of the implementation is moved from the existing code for `FocusScope`. There were a few things I wasn't sure about though:

* How it should be exposed. For now, it is available as `ReactDOM.unstable_FocusManager` if the event API feature flag is enabled. Perhaps it should be exposed as part of `react-events` instead of `react-dom`, but I didn't see a good way of exposing the necessary internals.
* How to detect `FocusScope` event components in the tree. For now, I added an `isFocusScope` option to the event responder, but this is kind of a hack. I didn't want react-dom to have a dependency on react-events, so I wasn't sure how else to do it.
* What should happen if you call `focusNext({from: node})`, where node contains focusable children? Should it focus the first item within that node, or go to the next focusable element within a sibling?

Thanks in advance for your feedback! 🙏


devongovett:
I'm not exactly sure why the `test_build` check failed. The `test_source` check is passing... It's getting `Cannot assign to read only property 'enableEventAPI' of object '#<Object>'`.

tjallingt:
I think its running tests on the bundles and since the feature flags aren't bundled it is failing.
I believe that the fix is to rename `FocusManager-test.js` to `FocusManager-test.internal.js`

devongovett:
Ah ok, thanks @tjallingt!

devongovett:
That seems to have fixed everything except fire. Maybe there is another flag for that somewhere...

tjallingt:
React Fire is a reimplementation of react dom right? Since FocusManager is exported on react dom you probably also need to change the export of Fire to include unstable_FocusManager to make it available to the fire tests.

devongovett:
Awesome, thanks for the help @tjallingt! 🎉

trueadm:
This looks awesome and I'll take a proper deep dive tomorrow. One thing I was thinking though - why is `FocusManager` coupled with ReactDOM? I thought it might make sense to couple it with `FocusScope`. i.e.

```js
import { FocusScope, FocusManager } from 'react-events/focus-scope';
```

or 

```js
import FocusScope from 'react-events/focus-scope';
const FocusManager = FocusScope.Manager;
```

You wouldn't really ever use one without the other too, but also it addresses some of the code-smells you had in regards to ReactDOM having to know implementation details relating to specific event responder modules (thus thus `.isFocusScope` hack).

Lastly, this isn't only relevant for ReactDOM. We might want the same high-level APIs to be exposed to other surfaces, such as React Native and React VR. In which case, they'd both be consuming the same APIs - so it doesn't make as much sense to be importing the manager from `ReactDOM`.

devongovett:
@trueadm yeah I totally agree - it should ideally be in react-events. I just couldn't find a good way to expose the necessary internals from react-dom for react-events to access, e.g. `getClosestInstanceFromNode` etc. In the existing event components in react-events, there is a context which allows access to those things, but since FocusManager is a singleton it doesn't have access to that context. So we'd need another way of exposing those internals.

If you have any guidance around how to do that, I'd be happy to update it.

trueadm:
@devongovett One way would be to make it explicit:

```jsx
let focusManager = null;
<FocusScope getFocusManager={fm => focusManager = fm}>
   ...
</FocusScope>
```

devongovett:
Definitely a possibility - I thought of maybe using a ref for it (make a ref to the FocusScope). It makes the API a little harder to use though IMO. Currently, the scope is found automatically based on the node you pass in, or the active element if no option is given. This means you don't need to know what scope an element is in, just that you need to move the focus.

Also, sometimes a scope is defined further up the tree, but the events to move focus (e.g. arrow key handlers) are further down. Not the worst thing in the world, but it would require either manual propagation of those key events up to the parent with the FocusScope, or passing a FocusManager instance down the tree to the component that needs it.

I guess this is an API ergonomics question more than anything.

trueadm:
@devongovett I guess I was just hinting at a ref there. Automatic scope is nice, but really, if the manager is tied to the current FocusScope, that magic might actually be a problem and causes bugs further on down the line. If you want to do `focusFirst` or anything like that, it's definitely coupled to a specific `FocusScope`, which means in many ways, `FocusManager` isn't actually global, but more of a context-like object.

devongovett:
Just saw your updated post. I think both usecases would be valuable to support. One usecase is to move focus within a scope: for example a keyboard navigable list component with up and down arrow support. In many cases, you don't know how many items will appear in that list, and shouldn't need to manually pass the keys of each item to `focusNext` and `focusPrevious` just to move within the scope.

Another usecase is to move focus between focus scopes like in your example of panels. I think the idea of using keys for that seems good. In addition, focusing a particular element by key might also be useful in case you know exactly which element to focus and want to avoid refs.

Definitely open to expanding on this API, I was mostly focused on the first usecase for this initial implementation.

devongovett:
Perhaps a hook like API to get the parent focus manager?

```javascript
let focusManager = useFocusManager();

let onKeyDown = e => {
  if (e.key === 'ArrowRight') {
    focusManager.focusNext();
  }
};
```

tjallingt:
@trueadm at that point why not something like this, which basically brings us back to how imperatively focusing dom nodes. (serious question, im not sure of, and want to learn about the implications of this)

```js
import {FocusScope} from 'react-events/focus-scope'

function LeftPanel() {
  let leftRef = useRef(null);
  let centerRef = useRef(null);
  let rightRef = useRef(null);

  function focusLeft() {
    if (isRTL) {
      centerRef.current.focusScope();
    } else {
      leftRef.current.focusScope();
    }
  }

  function focusRight() {
    ...
  }

  return (
    <FocusScope ref={leftRef} onLeftArrow={focusLeft} onRightArrow={focusRight}>
      ....
    </FocusScope>
  )
}
```

trueadm:
@tjallingt That means you need to pass around refs through your app. The nice benefit of keys is that you don't have to do that. You could maybe make use of either protocols, I see a validity in maybe having both for the cases where you need them.

tjallingt:
Sorry i removed my post since the discussion already caught up to what i was asking about (thanks for answering explicitly though :)

trueadm:
> Perhaps a hook like API to get the parent focus manager?

@devongovett That's a neat way. We can totally do that using hooks :)

> Definitely open to expanding on this API, I was mostly focused on the first usecase for this initial implementation.

That's fine. I just feel we should stay away from using DOM references. It makes this API hard to bring over to RN and other surfaces as soon as we make it about the DOM (which is definitely something we want to look into doing next half).

devongovett:
Alright. I'll try experimenting with a hooks based API to get the focus manager. Would you recommend implementing that with context? Can event components expose a value to context somehow? My knowledge of react internals is very limited so far. 😉

trueadm:
@devongovett The idea was to always have a way of using event components via hooks. We never focused on that this half, but the `createEventComponent` API always allowed for `props` which would be the default props passed to the event component. i.e.:

```js
function useFocus() {
  const focusManager = useRef(null);
  const FocusScope = useEventComponent(FocusScope, { ref: focusManager });

  return [FocusScope, focusManager ];
}

const [FocusScope, focusMananger] = useFocus();

<FocusScope>
   ...
</FocusScope>
```

Does that make sense? I can add a `React.unstable_useReactEventComponent` this week. I already designed this ages ago, like I said, so this should make more sense here.

devongovett:
Oh, so rather than importing `FocusScope` and using it directly, you'd have to call some sort of factory function for it? Wouldn't that mean you still have to pass the focus manager it returns down the tree yourself? Maybe I'm misunderstanding.

I guess I was more thinking you'd use FocusScope the same way as is currently implemented, and import a `useFocusManager` hook to use in child components to access the focus manager for the current subtree (like `useContext`).

```javascript
import {FocusScope, useFocusManager} from 'react-events/focus-scope';

function Parent() {
  return (
    <FocusScope>
      <Child />
    </FocusScope>
  );
}

function Child() {
  let focusManager = useFocusManager();
  let onKeyDown = e => {
    if (e.key === 'ArrowRight') {
      focusManager.focusNext();
   }
  };

  return <div onKeyDown={onKeyDown}>Child</div>;
}

**Comment by trueadm** [comment]
@devongovett Yeah, that was what I meant at the end of my comment. It's definitely possible to do, just will require some thought on how to best do it. :)

Give me tonight/tomorrow to think about such an API.

**Comment by trueadm** [comment]
Thinking about this more, I feel that you'd probably use React context to capture the focus manager instead of introducing new primitive hooks into React. Introducing new primitive hooks that are so coupled to a specific event component implementation is risky and somewhat brittle to change – so I think think we should avoid that.

**Comment by devongovett** [comment]
Yeah I agree. What if react implemented it internally using context though:

```javascript
const FocusScopeInternal = React.unstable_createEventComponent(
  FocusScopeResponder,
  'FocusScope',
);

const FocusContext = React.createContext();

export function FocusScope(props) {
  let focusManagerContext = {/* ... */};

  return (
    <FocusContext.Provider value={focusManagerContext}>
      <FocusScopeInternal {...props} />
    </FocusContext.Provider>
}

export function useFocusManager() {
  return useContext(FocusContext);
}
```

I guess the only small downside is that you'd see two components in dev tools instead of just one, but maybe there is a way to hide that.

trueadm:
@devongovett I don't think we want to start embedding context into `FocusScope`. However, `FocusScope.js` could export a functional component that does make use of context. 

```jsx
export function FocusScopeWithManager(props) {
  return React.createElement(
    FocusContext.Provider, {
      value: focusManagerContext,
      children: React.createElement(FocusScope, props);
    }
  );
}
```

devongovett:
Aside from basic focus containment, I think usage with a focus manager will be the most common use for a FocusScope. So not sure why we'd want two different components. I think that might get confusing.

trueadm:
@devongovett They are different components though: one is a functional component and the other is an event component. Event components can be consumed in different ways, so they both need to be exported from the event module.

devongovett:
How can event components expose an interface to a ref? Do we need to add an API to do that? Can the `onMount` responder method somehow assign things? I tried to track how refs work in other places, but didn't come to a conclusion on how best to handle it in this case.

trueadm:
@devongovett How do you mean? Event responders have access to the `props` of the event component in `onMount`, they could pass a reference to the focus manager there.

devongovett:
```javascript
let ref = React.createRef();
<FocusScope ref={ref}>
  {/* ... */}
</FocusScope>
```

Can FocusScope expose an interface to the ref somehow?

trueadm:
@devongovett Not yet, but I'll add something in a follow up soon. :)

trueadm:
@devongovett I've closed that PR I just linked. Thinking about it more, adding power functions is probably not the right move just yet – if only FocusScope is going to use context then changing the core API probably isn't going to scale well and just make for cases where it causes more issues than it solves. So `FocusScope` should probably export:

`export default FocusScope` as you showed above using a functional component wrapped
`export function useFocusManager` as you showed above
`export function useFocusScope` which will use the upcoming `useEventComponent` core hook which I'll be adding

devongovett:
@trueadm I've updated the PR to implement this with ref and hooks based access to the focus manager as discussed. Feels much cleaner now I think.

The focus-scope module now exports the following:

* `FocusScopeEventComponent` - the original event component. Perhaps to be replaced by `useFocusScope` later?
* `FocusScope` - a function component that exposes a ref to a focus manager, and wraps the event component in a context provider.
* `useFocusManager` - a hook to retrieve the current focus manager from context.

I couldn't see a way to implement this without adding an additional prop to the event component as well to get the focus manager (as you suggested above). An intermediary focus manager is exposed to the ref and to context, and on mount we get the internal one from the event component. If you attempt to call a focus manager function before the component is mounted, you'll get an error.

Also, because of the way the modules are exposed as CommonJS, I had to adjust the exports to use named exports instead of default exports. Would have had to add properties to the default exported component or something. I think it's clearer when there are multiple exports to use named exports anyway, but I realize this is a breaking change from the current implementation, so we can change it if needed.

trueadm:
@devongovett can you give some examples of usage with some code? I also back off adding ref support to event components in my other PR. So I wanted to see how you were doing it here. I guess you’re using a prop like you said?

devongovett:
The tests have some good examples. But, there are basically two ways:

Using a ref to access the focus manager in the same component as it is rendered:

```jsx
let List = (props) => {
  let focusManagerRef = React.createRef();
  let onKeyDown = (e) => {
    switch (e.key) {
      case 'ArrowUp':
        e.preventDefault();
        focusManagerRef.current.focusPrevious();
        break;
      case 'ArrowDown':
        e.preventDefault();
        focusManagerRef.current.focusNext();
        break;
    }
  };

  return (
    <FocusScope ref={focusManagerRef}>
      <ul onKeyDown={onKeyDown}>
        <li tabIndex="0">Item 1</li>
        <li tabIndex="0">Item 2</li>
        <li tabIndex="0">Item 3</li>
      </ul>
    </FocusScope>
  );
};
```

Accessing the focus manager in a child component via hooks:

```jsx
let DateRangePicker = (props) => {
  return (
    <FocusScope>
      <div>
        <DateField value={props.startDate} />
        &emdash;
        <DateField value={props.endDate} />
      </div>
    </FocusScope>
  );
};

let DateField = (props) => {
  return (
    <Fragment>
      <DateSegment type="year" />
      <DateSegment type="month" />
      <DateSegment type="day" />
    </Fragment>
  );
};

let DateSegment = (props) => {
  let focusManager = useFocusManager();
  let onKeyDown = (e) => {
    switch (e.key) {
      case 'ArrowUp':
        e.preventDefault();
        focusManager.focusPrevious();
        break;
      case 'ArrowDown':
        e.preventDefault();
        focusManager.focusNext();
        break;
    }
  };

  return <div tabIndex="0" onKeyDown={onKeyDown}>{/* ... */}</div>
};
```

The implementation uses a prop to get the focus manager from the internal event component, yeah. https://github.com/facebook/react/pull/15849/files#diff-587e37bf8623dff5ba8e829346a9046dR360

trueadm:
@devongovett That looks good, except you won't be able to pass a ref to `FocusScope`.

```jsx
<FocusScope ref={focusManagerRef}>
```

Instead use a prop:

``jsx
<FocusScope managerRef={focusManagerRef}>
```

Then you can access this in the event responders `onMount` and `onUnmount` methods, where you can update the `object.value`. I backed out of supporting `ref` as it's something we don't want to encourage with event components. A ref would also not make sense here, as you'd expect the `ref` to give the underlying `FocusScope` object, not its focus manager object.

**Comment by devongovett** [comment]
Hmm... as discussed earlier it's not an actual event component though, it's a function component wrapping the event component, so a ref works fine. Uses `useImperativeHandle` internally to expose the focus manager to the ref.

**Comment by trueadm** [comment]
@devongovett Why add the overhead of `useImperativeHandle` when you can just use a simple prop? Event though behind the scenes it's a function component, you're giving the impression to the user that these event components can now use refs, which is really confusing IMO and definitely something we should avoid here. I feel that `focusManagerRef` or some other explicit name is the best path forward for what we're trying to off the user.

**Comment by devongovett** [comment]
Isn't the focus manager meant to be the imperative controller for the focus scope? And aren't refs basically meant to expose imperative controllers for components? For DOM components, that's the underlying DOM node sure, but you can also call imperative methods on it. `useImperativeHandle` seems like it was designed for exactly this purpose. I guess I'm not sure about the overhead you mentioned. We could go with another name, but this seems to me like exactly what refs were meant for. The other event components don't currently expose a controller object, but perhaps they might in the future? I feel like it's more confusing to have separate ref props for each one rather than a consistent interface for accessing imperative controllers for components in general.

**Comment by trueadm** [comment]
@devongovett Let me explain what I'm saying with some code:

```jsx
let ListContainer = (props) => {
  return (
    <FocusScope>
      <List {...props} />
    </FocusScope>
  );
};

let List = (props) => {
  const mananger = useFocusManager();

  let onKeyDown = (e) => {
    switch (e.key) {
      case 'ArrowUp':
        e.preventDefault();
        mananger.focusPrevious();
        break;
      case 'ArrowDown':
        e.preventDefault();
        fmananger.focusNext();
        break;
    }
  };

  return (
    <ul onKeyDown={onKeyDown}>
      <li tabIndex="0">Item 1</li>
      <li tabIndex="0">Item 2</li>
      <li tabIndex="0">Item 3</li>
    </ul>
  );
};
```

The implementation of these looks like this:

```jsx

const FocusScopeImpl = React.unstable_createEventComponent(FocusScopeResponder, 'FocusScope');
const FocusManager = React.createContext(null);

function createFocusManager() {
  return {
    // ...
  };
}

export function FocusScope({ autoFocus, contain, children }) {
  const focusManager = useRef(() => createFocusManager()));
  return (
    <FocusManager.Provider value={focusManager}>
      <FocusScopeImpl
        autoFocus={autoFocus}
        contain={contain}
        // let the FocusScope responder know about
        // its associated mananger
        focusManager={focusManager}
        children={children}
      />
    </FocusManager.Provider>
  )
}

export function useFocusManager() {
  return useContext(FocusManager);
}
```

devongovett:
Hmm ok... that's not what we discussed earlier in this thread. Do you have an example for why you think accessing via `ref` is bad? I didn't quite follow this last sentence:

> There is a very good reason for this, it discourages access focus managers in child sub-trees, which I feel will lead to more confusion and many instances of hard to find bugs.

Isn't the hook for accessing focus managers in child sub-trees?

trueadm:
@devongovett > ok... that's not what we discussed earlier in this thread

I think we were talking about different things. I was trying to come up with an approach that didn't use a `ref` on the `FocusScope` but did use a ref internally.

> Isn't the hook for accessing focus managers in child sub-trees? 

No, they access the immediate parent instance in the current branch of the tree, rather than the child of a tree. It's only context after-all.

> Do you have an example for why you think accessing via ref is bad?

They're simply not needed here. Having one way of accessing the focus manager, via the hook is much better than having many different ways (hooks + ref) IMO. Plus, I'd really like a future world where React did not have to use refs, and instead could use hooks, effects and declarative UI via props to describe everything. Given event components are new, I think we should try harder before just use the escape hatch approach of `ref`, which really doesn't give any signal to me. `useFocusManager` does exactly what it says on the tin!

devongovett:
I'm all for getting rid of refs, but it might be confusing for people that they need to wrap their component in another component in order to use a focus manager. Calling `useFocusManager` in the same component where they render the `<FocusScope>` would give them back the parent focus manager instead, which might not have been what they thought. 🤷‍♂ 

I'll go ahead and change it. Just thought I'd throw that out there.

devongovett:
@trueadm Ref support removed.

trueadm:
There are some merge conflicts.

devongovett:
@trueadm Fixed the conflicts.

devongovett:
... and now there are more. Let me know when you're about ready to merge it and I'll resolve again. 😉

packages/react-events/src/FocusScope.js
```diff
@@ -157,4 +329,50 @@ const FocusScopeResponder = {
...
-export default ReactDOM.unstable_createEvent(FocusScopeResponder, 'FocusScope');
+export const FocusScopeEventComponent = ReactDOM.unstable_createEvent(
+  FocusScopeResponder,
+  'FocusScope',
+);
+
+const FocusScopeContext: React.Context<FocusManager> = React.createContext();
+
+export function FocusScope(props: FocusScopeProps) {
+  let internalFocusManager: ?FocusManager;
```
aweary: Is it intentional that this gets reset every time `FocusScope` renders? Maybe this should be stored in a ref?

packages/react-events/src/FocusScope.js
```diff
@@ -157,4 +329,50 @@ const FocusScopeResponder = {
...
+      return internalFocusManager.focusNext(options);
+    },
+    focusPrevious(options: FocusManagerOptions = {}) {
+      invariant(
+        internalFocusManager != null,
+        'Attempt to use a focus manager method on an unmounted component.',
+      );
+      return internalFocusManager.focusPrevious(options);
+    },
+  };
```
aweary: `focusManager` should be memoized if it's being passed down through context.

packages/react-events/src/__tests__/FocusScope-test.internal.js
```diff
@@ -241,4 +245,466 @@ describe('FocusScope event responder', () => {
...
+            </div>
+          );
+        };
+
+        ReactDOM.render(<ScopeParent />, container);
+
+        focusManager.focusPrevious({from: divRef.current});
+        expect(document.activeElement).toBe(inputRef.current);
+      });
+    });
```
aweary: @devongovett can you add a test for nested `FocusScope`s?

devongovett:
@aweary well... I already suggested two possible APIs to do this in this PR which were both rejected (singleton, and refs), so I think a follow up would be needed. I'd love to hear other suggestions!

packages/react-events/src/FocusScope.js
devongovett: That's a good suggestion

devongovett:
Fixed the conflicts again and updated to incorporate @aweary's comments.

devongovett:
Closing since this won’t be merged. Will follow up with @trueadm separately on the current state of focus management in React.

PR #15871, opened 2019-06-12
[Flare] Clean up of Press responder + better touch support

trueadm:
This PR aims at fixed a bunch of bugs with iOS in `Press`. Notably: ensuring the responder region is correctly adhered to and also accounting for multiple presses for all browsers. There have also been numerous small performance optimizations and now the runtime of a press to the `onPress` firing is around 20% less.

Another area I've tackled is the behaviour of pointer events. Previously, we used to fully leverage pointer events, but this ended up not being what we really wanted. `pointercancel` fires too frequently in most cases and makes for a detached user-experience when "touching" the UI compared to native "touching". We now switch to using `touchstart`, `touchmove` and `touchend` again, with `pointerType` co-ordination so we don't lose the pointer types.

I've also added basic support for `trackpad` pointer events. More testing is needed for this on iOS 13 though.

Demo URL: https://bxlrc.codesandbox.io/

### DIFF
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -189,10 +189,9 @@ const eventResponderContext: ReactResponderContext = {
     eventListeners.set(eventObject, listener);
     eventQueue.events.push(eventObject);
   },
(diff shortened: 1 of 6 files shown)


trueadm:
Closing. Will pull parts of the PR out into other PRs to reduce risk of landing such a big PR.

PR #15872, opened 2019-06-12
Interaction tracing works across hidden and SSR hydration boundaries

bvaughn:
Prior to this PR, interaction tracing does not work correctly for:
* Hidden subtrees
* SSR hydration containing one or more suspense boundaries

In both cases, interactions were not (reliably<sup>1</sup>) carried across any yielded work, which means the `Profiler` did not (reliably) report them and the subscription API prematurely reported an interaction as "complete" (no more outstanding work). This PR adds several new tests and fixes this behavior.

The decision to trace interactions across hidden tree boundaries may be somewhat controversial. It could be argued that idle work should not be included in a traced interaction because it might result in an interaction intended to trace _high priority_ work being drawn out because it also includes _low priority_ work. However I think this is the right decision because you can always opt out of this behavior by scheduling the low priority (hidden) work within a `unstable_clear()` scope.

<sup>1</sup> - In the case of hidden subtrees, interactions would be traced across boundaries under certain conditions (if there was other outstanding work scheduled for a given interaction). This was more of an unintentional side effect though. After this PR the behavior is _predictable_.


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1699,13 +1694,40 @@ function commitRootImpl(root) {
       currentTime,
       remainingExpirationTime,
     );
+
+    let prevInteractions: Set<Interaction> | null = null;
+    if (enableSchedulerTracing) {
+      if (remainingExpirationTime === Never) {
+        // Temporarily restore interactions in the case of e.g. hidden subtrees and suspended hydration.
+        // Don't restore otherwise or it would blur interrupting high-pri udpates with low-pri work.
```
bvaughn: I think this is the right approach to trace hidden work without blending high/low pri updates.

bvaughn: Hmm... this might still cause undesirable blurring if the idle/never work was there _before_ a high pri interruption. That seems harder to prevent. Is it worth trying to?

bvaughn:
FWIW Wooseok verified this fixes the issue he originally reported against FB5+SSR.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1698,15 +1696,38 @@ function commitRootImpl(root) {
       currentTime,
       remainingExpirationTime,
     );
+
+    if (enableSchedulerTracing) {
+      // Temporarily restore interactions in the case of e.g. hidden subtrees and suspended hydration.
+      // Don't restore otherwise or it would blur interrupting high-pri udpates with low-pri work.
```
acdlite: This comment is outdated now that you're scheduling the interactions directly

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1698,15 +1696,38 @@ function commitRootImpl(root) {
...
+      // Wait to do this until after remaining work has been scheduled,
+      // so that we don't prematurely signal complete for interactions when there's e.g. hidden work.
+      finishPendingInteractions(root, expirationTime);
+    }
+  }
+
   onCommitRoot(finishedWork.stateNode, expirationTime);
 
+  if (enableSchedulerTracing) {
+    reschedulePendingInteractionsAtNeverPriority = false;
```
acdlite: Feels fishy that this is in a separate branch from the one above where you check if it's true

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2507,14 +2528,17 @@ function computeThreadID(root, expirationTime) {
   return expirationTime * 1000 + root.interactionThreadID;
 }
 
-function schedulePendingInteraction(root, expirationTime) {
-  // This is called when work is scheduled on a root. It sets up a pending
-  // interaction, which is completed once the work commits.
+export function markPendingInteractionsToBeRescheduledAtNeverPriority() {
```
acdlite: Naming suggestion: `markDidDeprioritizeIdleSubtree`

acdlite:
Oh and you need to add the new variable (`reschedulePendingInteractionsAtNeverPriority`) to `prepareFreshStack`, too, so it gets reset when there's an interruption.

packages/react-reconciler/src/ReactFiberWorkLoop.js
bvaughn: This is in a broader branch. Seemed best to always unset it, but maybe I should just do that in `prepareFreshStack`

bvaughn:
Right. I was just coming to that conclusion too.

PR #15875, opened 2019-06-12
[Flare] Adds event responder getTransformedProps

trueadm:
This adds an additional power feature to event responders in the experimental event system. Specifically:

- `getTransformedProps` allows event responders to transform incoming props and output a slightly different set. This can be useful for when event responders want to utilize other React primitives, like context. It can also be a way to compose event components with wrappers that are used for logging or tracking at DEV time.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -2067,15 +2067,24 @@ function updateContextConsumer(
 }
 
 function updateEventComponent(current, workInProgress, renderExpirationTime) {
-  const nextProps = workInProgress.pendingProps;
-  let nextChildren = nextProps.children;
+  const eventResponder = workInProgress.type.responder;
+  const transformProps = eventResponder.transformProps;
+  let nextPendingProps = workInProgress.pendingProps;
+  if (typeof transformProps === 'function') {
+    workInProgress.pendingProps = nextPendingProps = transformProps(
+      nextPendingProps,
+    );
+  }
+  if (nextPendingProps !== null) {
+    const nextChildren = nextPendingProps.children;
 
-  reconcileChildren(
-    current,
-    workInProgress,
-    nextChildren,
-    renderExpirationTime,
-  );
+    reconcileChildren(
+      current,
+      workInProgress,
+      nextChildren,
+      renderExpirationTime,
+    );
+  }
   pushHostContextForEventComponent(workInProgress);
   return workInProgress.child;
 }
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2067,15 +2067,24 @@ function updateContextConsumer(
 }
 
 function updateEventComponent(current, workInProgress, renderExpirationTime) {
-  const nextProps = workInProgress.pendingProps;
-  let nextChildren = nextProps.children;
+  const eventResponder = workInProgress.type.responder;
+  const getTransformedProps = eventResponder.getTransformedProps;
+  let nextPendingProps = workInProgress.pendingProps;
+  if (typeof getTransformedProps === 'function') {
+    workInProgress.pendingProps = nextPendingProps = getTransformedProps(
```
sebmarkbage: I'm not sure this will actually remain safe. I'm about to add some code that does some minor resuming which might pick up these transformed props in a second pass over the component as the input.

trueadm:
@sebmarkbage The use case for `getTransformedProps ` was for when the event responder needed to wrap children in additional logic, for example React Context. For FocusScope, it could do this to add a context that allows components down the tree find the parent FocusScope focus manager, which would be in context (that was the the use case for adding it).

In terms of the ref use-case. Can you explain how it doesn't handle it? We update the `ref` on the event component instance after each update in the complete phase. That is the `ref` that the event responder invokes, which should always be safe and up-to-date. We never use the one on the fiber directly.

packages/react-reconciler/src/ReactFiberBeginWork.js
trueadm: Okay, let me know what I need to change to make this work, assuming I can address your concerns with this API.

sebmarkbage:
There's nothing preventing a custom event to expose itself as a true component.

E.g.

```
FocusScopeImpl = createEventComponent(...);
FocusScope = props => {
  return <FocusScopeImpl ...>
    <FocusContext value={something}>
      {children}
    </FocusContext>
  </FocusScopeImpl>
};
```

Then it can do all kind of custom manipulation, add state, refs and other features.

Replicating everything you can do in user space, in internals gets complicated very quickly.

trueadm:
This was brought up in the other discussion. I agree, except you're now exporting a functional component rather than an event component. We'd also need a way of accessing the original event component to be able to compose it directly (with upcoming event component hooks etc).

trueadm:
I'm going to close this and go back to the original approach of just leaving it to the user without the props transform and see how far we can go without needing these.

PR #15877, opened 2019-06-13
Construct Error at invariant call site for clearer stack traces

motiz88:
# Summary

Currently, when a React invariant fails, it throws an `Error` object constructed inside the `ReactError` (or `ReactErrorProd`) function. Because the stack trace is captured at the point of Error's construction, it always begins with the same frame for ReactError, no matter specific invariant actually failed.

This PR moves the Error constructor call to the location of the invariant, removing the stack frame inside ReactError.

# Test plan

Added a test in `ReactError-test.internal.js`.

---

NOTE: The PR summary has been significantly edited a couple of times; see edit history for more context.

### DIFF
--- a/packages/shared/ReactError.js
+++ b/packages/shared/ReactError.js
@@ -6,12 +6,11 @@
  *
  */
 
-// Do not require this module directly! Use a normal error constructor with
+// Do not require this module directly! Use normal `invariant` calls with
 // template literal strings. The messages will be converted to ReactError during
 // build, and in production they will be minified.
 
-function ReactError(message) {
-  const error = new Error(message);
+function ReactError(error) {
   error.name = 'Invariant Violation';
   return error;
 }
--- a/packages/shared/ReactErrorProd.js
+++ b/packages/shared/ReactErrorProd.js
@@ -6,20 +6,21 @@
  *
  */
 
-// Do not require this module directly! Use a normal error constructor with
+// Do not require this module directly! Use normal `invariant` calls with
 // template literal strings. The messages will be converted to ReactError during
 // build, and in production they will be minified.
 
-function ReactErrorProd(code) {
+function ReactErrorProd(error) {
+  const code = error.message;
   let url = 'https://reactjs.org/docs/error-decoder.html?invariant=' + code;
   for (let i = 1; i < arguments.length; i++) {
     url += '&args[]=' + encodeURIComponent(arguments[i]);
   }
-  return new Error(
+  error.message =
     `Minified React error #${code}; visit ${url} for the full message or ` +
-      'use the non-minified dev environment for full errors and additional ' +
-      'helpful warnings. ',
-  );
+    'use the non-minified dev environment for full errors and additional ' +
+    'helpful warnings. ';
+  return error;
 }
 
 export default ReactErrorProd;
--- a/packages/shared/__tests__/ReactErrorProd-test.internal.js
+++ b/packages/shared/__tests__/ReactErrorProd-test.internal.js
@@ -36,7 +36,7 @@ describe('ReactErrorProd', () => {
 
   it('should throw with the correct number of `%s`s in the URL', () => {
     expect(function() {
-      throw ReactErrorProd(124, 'foo', 'bar');
+      throw ReactErrorProd(Error(124), 'foo', 'bar');
     }).toThrowError(
       'Minified React error #124; visit ' +
         'https://reactjs.org/docs/error-decoder.html?invariant=124&args[]=foo&args[]=bar' +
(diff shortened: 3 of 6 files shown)


acdlite:
`ReactError` and `ReactErrorProd` are different from `invariant` in that they are error constructors, not functions that throw.

```js
throw ReactError();
```

So there's no extra frame that needs popping.

acdlite:
Closing. If I'm mistaken in my reasoning, please let me know with additional details and I'll reopen.

motiz88:
@acdlite, I'm not at my laptop now but I'm quite certain that's incorrect. They are functions that create an Error instance (which is where the stack gets captured) and return it, so there is in fact an extra frame there.

I'm seeing this with the error reporting for FB's RN apps - all the renderer invariants currently get aggregated by that first frame inside ReactError.

acdlite:
> They are functions that create an Error instance

Ah right. Ugh... Maybe we should make them actual constructors.

Lemme think about this before merging `framesToPop` because this affects open source users, too. Would like a solution that works for everyone.

acdlite:
What if we changed the error transform so that the error is passed to `ReactError` as an argument?

```js
throw ReactError(new Error(msg));
```

motiz88:
I'd be happy with changing the transform - that sounds like the best way forward for now. A few thoughts:

> make them actual constructors

That gets tricky depending on the execution environment. In a native ES2015 environment like Chrome, extending Error works as you'd expect, but under Babel there are some additional frames on the stack. Currently dealing with this internally :)

> this affects open source users, too

Just for full context - the `framesToPop` fix would work for RN users in open source as well. I agree that it's an unnecessary change as far as non-RN users of React are concerned, but given the precedent of `invariant` I'd bet on it being at least harmless.

Overall, `framesToPop` isn't a general enough mechanism to standardise or advocate for, but I'm hoping to come up with something better in the next few months.

gaearon:
The bundle size increase is not great. :-(

Can we make this FB-only or something?

motiz88:
Maybe we can compensate for the increase by making `ReactError` actually throw. I'll try that first.

motiz88:
Hmm, that didn't work the way I wanted it to. I've also tried moving `Error()` to be its own argument instead of wrapping the message, hoping that would be more compressible, but the improvement wasn't significant.

Options I see:

1. Roll back to 1488297 (`framesToPop` approach), which doesn't change the `invariant` call sites and consequently has a minimal size footprint.
2. Only modify `ReactError` (the `__DEV__` / `noMinify: true` path), leaving `ReactErrorProd` alone. This would satisfy RN's needs since we don't minify errors when building for RN, but ideally I'd want this in the `FB_WWW_PROD` bundle too.
3. Pass down a new flag into the `invariant` transform, only set it for FB bundles, and create `ReactError.fb.js` and `ReactErrorProd.fb.js` files that have the new interface. This adds the most complexity but only affects FB.
4. 💡 Just fork `ReactError` and `ReactErrorProd` for FB and go the `framesToPop` route, since all of FB's use cases support this convention.

4 is my favourite option now. Any objections?

motiz88:
Open questions:

1. How to test this change? I can think of a few approaches but they're all kinda churny, so I'm going to wait for someone to voice an opinion before proceeding.
2. Should we enable this fork in RN OSS too? I'm leaning towards yes (since RN's error handler supports `framesToPop`) but I'd definitely want this to be covered by a good test in that case.

acdlite:
Re: the version where you construct the error at the callsite, I inspected the build output and it looks like it only goes up because `Error` does not get minified.

```diff
- throw z(85)
+ throw z(Error(85))
```

This doesn't seem like a huge deal to me. Bundle size as a metric is mostly a proxy for parse time, which I wouldn't expect this to contribute to significantly.

However, I wonder what the increase would look like if you alias `Error` to a variable name?

> Maybe we can compensate for the increase by making ReactError actually throw. I'll try that first.

It used to work like that, like `invariant`, but we changed it to the current version so we can eventually support using normal errors in the source (among other reasons). See https://github.com/facebook/react/pull/15072

acdlite:
Also I would still like to find a solution that works for people outside Facebook and React Native.

motiz88:
If we find the bundle size increase to be acceptable, we have a couple of general solutions ready to go right here in the PR.

Aliasing `Error` occurred to me, but I'm not sure what a good way to automate that would be. It's also, strictly speaking, a change of behaviour (imagine someone overwriting `global.Error` after loading React).

I'll implement any changes you folks think are necessary.

acdlite:
> imagine someone overwriting global.Error after loading React

Meh, don't care :D

> Aliasing Error occurred to me, but I'm not sure what a good way to automate that would be

Import a module that exports Error? :D 

I will stamp a version that passes the error in, preferably with Error minified but without it is also fine.

motiz88:
@acdlite: Rolled back to 1488297 and then rebased on master as 765e068. I have a local version that aliases `Error` to a variable (using the module approach & Rollup's neat hoisting) but it doesn't quite seem to pay off in gzip size. Let's go with this as-is?

PR #15878, opened 2019-06-13
[Flare] Refactor of Press to fix various issues

trueadm:
This PR is a follow up of https://github.com/facebook/react/pull/15871. I intended to break this all down into smaller PRs, but was unable to do it without breaking the tests. Everything is quite coupled. I was however, able to extract out the changes we wanted from #1571 without changing how everything fundamentally works. This PR does not drop using Pointer Events and switching to touch events like the other PR and works like everything does today – so this is far less risky.

Notably, the changes and fixes are as follows:

- We now check the `target` correctly from touch events using `document.elementFromPoint`, fixing several iOS issues.
- We now ensure both pointer and touch events match their ID. This allows for multiple touches to occur without causing them to conflict with one-another (this also applies to pointer events).
- There was a regression with `contextmenu` blocking long press on touch devices. This fixes that.
- We now expose `pointerId` and `pointerType` on the responder `event` object. This is far cleaner and faster than how we did this before, plus it's less code! For cases where we don't have a `pointerId`, the default is `null`.
- I cleaned up some parts of the code that duplicated conditions and tried to unify them into hoisted declarations with nice names.
- I pulled in the tests from #15871 (they have much better touch input mocking) and manually inlined `elementFromPoint` into the Jest tests because JSDOM doesn't have support for it.

Example URL: https://codesandbox.io/s/experimental-event-api-v2-2rqnr

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -789,10 +789,7 @@ const PressResponder = {
         if (state.pointerType !== pointerType) {
           return;
         }
-        if (
-          type === 'pointermove' &&
-          activePointerId !== pointerId
-        ) {
+        if (type === 'pointermove' && activePointerId !== pointerId) {
           return;
         } else if (type === 'touchmove') {
           touchEvent = getTouchById(nativeEvent, activePointerId);
@@ -854,10 +851,7 @@ const PressResponder = {
         if (isPressed) {
           let isKeyboardEvent = false;
           let touchEvent;
-          if (
-            type === 'pointerup' &&
-            activePointerId !== pointerId
-          ) {
+          if (type === 'pointerup' && activePointerId !== pointerId) {
             return;
           } else if (type === 'touchend') {
             touchEvent = getTouchById(nativeEvent, activePointerId);
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -8,6 +8,7 @@
  */
 
 import type {
+  PointerType,
   ReactResponderEvent,
   ReactResponderContext,
 } from 'shared/ReactTypes';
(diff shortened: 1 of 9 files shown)


packages/react-events/src/Press.js
```diff
@@ -571,11 +570,56 @@ function removeRootEventTypes(
...
+  }
+  return null;
+}
+
+function getTouchTarget(context: ReactResponderContext, touchEvent: Touch) {
+  const doc = context.getActiveDocument();
+  return doc.elementFromPoint(touchEvent.clientX, touchEvent.clientY);
+}
+
+function updatePressResponderRegion(
```
necolas: nit: should probably be called updateIsPressWithinResponderRegion and it looks like the `isPressWithinResponderRegion` function is only used within this new function now, so maybe that can be inlined here?

packages/react-events/src/Press.js
```diff
@@ -659,18 +704,23 @@ const PressResponder = {
           // data around for handling of the context menu
           state.pointerType = pointerType;
           state.pressTarget = context.getEventCurrentTarget(event);
+          if (isPointerEvent) {
+            state.activePointerId = pointerId;
+          } else if (isTouchEvent) {
+            const touchEvent = getTouchFromPressEvent(nativeEvent);
+            state.activePointerId = touchEvent.identifier;
```
necolas: FYI, might be worth looking into which versions of Safari are still reusing touch identifiers - https://github.com/necolas/react-native-web/blob/36dacb2052efdab2a28655773dc76934157d9134/packages/react-native-web/src/modules/normalizeNativeEvent/index.js#L15-L22

trueadm: I'm not sure I fully follow it, is this a well known bug? It seems huge if the identifiers can be the same for events that aren't the same?!

necolas: Safe it ignore. I can't find anything about it online or in the original commit (>3 years ago). In fact, it sounds like Safari always produces unique identifiers whereas Chrome will recycle them (but presumably in a safe way).

PR #15882, opened 2019-06-14
[Suspense] Use style.setProperty to set display

acdlite:
Follow up to #15861.

Turns out you can't set `!important` using a normal property assignment. You have to use `style.setProperty`.

Maybe Andrew *should* just learn CSS.

IE9 doesn't support `style.setProperty` so we'll fall back to setting `display: none` without `important`, like we did before #15861. Our advice for apps that need to support IE9 will be to avoid using `!important`. Which seems like good advice in general, but IANACSSE.

Tested on FB and using our Suspense DOM fixture.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.js
@@ -83,25 +83,25 @@ describe('ReactDOMSuspensePlaceholder', () => {
           <div ref={divs[1]}>
             <AsyncText ms={500} text="B" />
           </div>
-          <div style={{display: 'block'}} ref={divs[2]}>
+          <div style={{display: 'inline'}} ref={divs[2]}>
             <Text text="C" />
           </div>
         </Suspense>
       );
     }
     ReactDOM.render(<App />, container);
-    expect(divs[0].current.style.display).toEqual('none !important');
-    expect(divs[1].current.style.display).toEqual('none !important');
-    expect(divs[2].current.style.display).toEqual('none !important');
+    expect(window.getComputedStyle(divs[0].current).display).toEqual('none');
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/__tests__/ReactDOMSuspensePlaceholder-test.js
```diff
@@ -83,25 +83,25 @@ describe('ReactDOMSuspensePlaceholder', () => {
           <div ref={divs[1]}>
             <AsyncText ms={500} text="B" />
           </div>
-          <div style={{display: 'block'}} ref={divs[2]}>
+          <div style={{display: 'span'}} ref={divs[2]}>
```
sebmarkbage: ?

acdlite: I changed it to `span` since `block` is the default display for divs.

acdlite: Ahaha I meant `inline`.

acdlite: Thanks, brain!

acdlite: I was talking to my brain, not you.

ryota-murakami: But `!important` only accept from `style.setProperty` is really first time killer except holder of CSS degree😅
Thanks find out this @acdlite 😀

MuYunyun: Following up Andrew to learn CSS. :smile：

NE-SmallTown:
@threepointone approved these changes 6 days ago 👀

PR #13321, opened 2018-08-03
Provide esm entry point for react-is

TrySound:
Refs
  https://github.com/facebook/react/issues/13272#issuecomment-409939608
  https://github.com/reactjs/rfcs/pull/38

In this PR I modified build script to generate two more bundles
`react-is.development.mjs` and `react-is.production.min.mjs` and
entry point `index.mjs` which reexports all named values depending on
production/development environments like this

```
import * as dev from './esm/development.mjs'
import * as prod from './esm/production.min.mjs'

export var name =
  process.env.NODE_ENV !== 'production' ? dev.name : prod.name;
```

I replaced closure wrapper with vendor plugin which is able to process
es modules by replacing them before and returning them after closure
minification.

To prove treeshakability of dual entry point we can generate rollup
bundle with `import {} from 'package'`. If nothing will left then
user bundler is able to eliminate unused code from dual entry point.

PR #15885, opened 2019-06-14
Make React compatible with ES Modules and Pika

sdegutis:
See https://www.pika.dev/about/ for basic information about Pika.

When searching for React (https://www.pika.dev/search?q=react) in Pika's interface, it shows:

> EXACT MATCH!
> react      
> Package found! However, no ES "module" entrypoint was detected in its package.json manifest. If possible, explore the site for a more web-friendly alternative.

It looks like this commit in my PR is all that's needed, because React is fortunately already written using ES Module syntax.

### DIFF
--- a/packages/react-dom/package.json
+++ b/packages/react-dom/package.json
@@ -3,6 +3,7 @@
   "version": "16.8.6",
   "description": "React package for working with the DOM.",
   "main": "index.js",
+  "module": "src/client/ReactDOM.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/facebook/react.git",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -8,6 +8,7 @@
   "homepage": "https://reactjs.org/",
   "bugs": "https://github.com/facebook/react/issues",
   "license": "MIT",
+  "module": "src/React.js",
   "files": [
     "LICENSE",
     "README.md",


sdegutis:
Ok I just signed the CA.

sdegutis:
I added a second commit that makes the same exact change for ReactDOM. This PR should now be sufficient for most (all?) React users to now use Pika and ES Modules to load React.

TrySound:
This won't work. Sources are not ready for consuming.
This is the first step for this https://github.com/facebook/react/pull/15037
This is the second which is blocked til upgrade is merged.
https://github.com/facebook/react/pull/13321

sdegutis:
If it's true, that's a shame. Guess we'll have to wait to use React with Pika and use Preact@next in the meantime. Thanks @TrySound. Feel free to close this PR if it doesn't make sense in context of being merged after those two PRs are merged.

TrySound:
From what I see pika is able to provide own version of react for components.
Here react is imported from `https://cdn.pika.dev/_/react/v16`
https://storage.googleapis.com/pika-web-packages-02/rifm/0.9.1/dist-es2019/rifm.min.js

sdegutis:
Oh perfect, how did I miss that!

FredKSchott:
For anyone else who finds this issue, we've added a note for how to use React with @pika/web
https://github.com/pikapkg/web#a-note-on-react

You can also search "react esm" on pika.dev to find a few different options for React & ReactDOM, published as ESM:
https://www.pika.dev/search?q=react%20esm

jadjoubran:
Thanks @FredKSchott, though it seems that react-esm has been deleted

FredKSchott:
Can you share a link to the deleted thing? I see react-esm & @reactesm both available: 

https://www.npmjs.com/package/react-esm
https://www.npmjs.com/package/@reactesm/react

jadjoubran:
Sorry I think I drew the wrong conclusion, however it's **react-esm** is not working on PikaCDN, it returns a 404:
https://cdn.pika.dev/react-esm/v16
Screenshot 2019-10-07 at 12 23 50

FredKSchott:
Ah, that looks like a bug in the CDN resolver logic. Thanks for sharing, I'll get that fixed asap

jadjoubran:
awesome thanks a lot!

PR #15889, opened 2019-06-14
[Flare] remove stopLocalPropagation option + modify responder ownership

trueadm:
After lots of internal discussion and also quite a lot of discussion on GitHub in relation to Focus, it occurred to me that the model that React Flare tries to keep consistent can easily be broken by a few features that leave the architecture up to abuse. Notably, the notion of event components not leaking propagation to other event components of the same type. It has fixed many issues that the system it replaced so rather than make it `optional`, this PR makes it the always default. `stopLocalPropagation` flag has been removed and now it applies to all event responders.

That also means that we no longer have any need for responder ownership, because by having `stopLocalPropagation` be the default, we also get that for free – so lots of code to kill off!

In the case of `FocusScope` and some tests that need to bypass the local propagation in very defined cases (a FocusScope loses focus, and needs to focus to bubble to the next FocusScope), there is a new context method called `continueLocalPropagation`.

Overall, this PR simplifies many of the concepts of this event system and keeps things more rigid, leading people to question their ideas before they try and hack the system to work like the heritage event system.

### DIFF
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -107,11 +107,8 @@ const eventListeners:
       ($Shape<PartialEventObject>) => void,
     > = new PossiblyWeakMap();
 
-const responderOwners: Map<
-  ReactEventResponder,
-  ReactEventComponentInstance,
-> = new Map();
 let globalOwner = null;
+let continueLocalPropagation = false;
 
 let currentTimeStamp = 0;
 let currentTimers = new Map();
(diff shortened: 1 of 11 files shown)


packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -453,6 +434,10 @@ const eventResponderContext: ReactResponderContext = {
     }
     return false;
   },
+  cotinueLocalPropagation() {
```
necolas: cotinue -> continue

trueadm: LOL :D

PR #15890, opened 2019-06-14
[Flare] Move unstable_createEventComponent to ReactDOM

trueadm:
This PR does two things:

- Moves `React.unstable_createEventComponent` to `ReactDOM.unstable_createEventComponent`.
- Breaks up the new event system Flow types by core and ReactDOM, thus making it easier in the future to have different Flow types per renderer/platform.

### DIFF
--- a/packages/shared/ReactTypes.js
+++ b/packages/shared/ReactTypes.js
@@ -7,6 +7,8 @@
  * @flow
  */
 
+import type {EventResponder} from 'react-reconciler/src/ReactFiberHostConfig';
+
 export type ReactNode =
   | React$Element<any>
   | ReactPortal
@@ -84,7 +86,7 @@ export type RefObject = {|
 export type ReactEventComponentInstance = {|
   currentFiber: mixed,
   props: null | Object,
-  responder: mixed,
+  responder: EventResponder,
   rootEventTypes: null | Set<string>,
   rootInstance: mixed,
   state: null | Object,
@@ -94,7 +96,7 @@ export type ReactEventComponent = {|
   $$typeof: Symbol | number,
   displayName: string,
   props: null | Object,
-  responder: mixed,
+  responder: EventResponder,
 |};
 
 export type ReactEventTarget = {|
--- a/packages/shared/ReactTypes.js
+++ b/packages/shared/ReactTypes.js
@@ -81,54 +81,10 @@ export type RefObject = {|
   current: any,
 |};
 
-export type ReactEventResponderEventType =
-  | string
-  | {name: string, passive?: boolean};
-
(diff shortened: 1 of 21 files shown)


packages/shared/ReactTypes.js
```diff
@@ -81,54 +81,10 @@ export type RefObject = {|
...
-    props: null | Object,
-    state: null | Object,
-  ) => void,
-};
-
 export type ReactEventComponentInstance = {|
   currentFiber: mixed,
   props: null | Object,
-  responder: ReactEventResponder,
+  responder: mixed,
```
sebmarkbage: I believe you should be able to reference a type defined in a host config here. Similar to how we import Instance in the reconciler even though the actually type is renderer specific.

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberCommitWork.js#L11
https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMHostConfig.js#L85

So you could potentially reexport it through the host config to get to it from here.

That way `yarn flow dom` tests this more strictly. Other host configs would have to export something else here. Like any or mixed.

scripts/rollup/bundles.js
```diff
@@ -465,7 +465,7 @@ const bundles = [
     moduleType: NON_FIBER_RENDERER,
     entry: 'react-events/press',
     global: 'ReactEventsPress',
-    externals: ['react'],
+    externals: ['react-dom'],
```
sebmarkbage: I believe this means that we should add it as a dep in the package.json.

https://github.com/facebook/react/blob/master/packages/react-events/package.json#L40

I  know yarn pnp relied on this package information for module resolution.

That's unfortunate since it's not a dep for the whole package. Just when used in a DOM environment. Not sure how mixed RN / web code deals with this. cc @necolas

trueadm: This isn't a big deal for now. We aren't going to be publishing `react-events` anytime soon. This was one of the reasons I tried to avoid this approach to begin with, but we could get around this with some kind of build step, it just means that these can't really ever go in `react`, but must be their own separate package – otherwise `react` will pull in ever renderer out there.

packages/shared/ReactTypes.js
trueadm: Awesome, thank you! This helps a lot. I do wonder about the react-native sync though. I think they copy over the ReactTypes.js as is, which would now break. @TheSavior maybe we can change how this process is done slightly (you pointed this out to me last time), as now this file depends on `react` repo dependencies. Maybe when doing the RN sync, we can change these fields to any, as these are all build artifacts anyway?

PR #15892, opened 2019-06-14
Fix outdated test comments

acdlite:
Comments for this unit test are confusing because they refer to an earlier version of the same test.


gaearon:
Oof maybe we should stop writing comments

acdlite:
Let me first compose a tweetstorm detailing why you're wrong

acdlite:
My commit count for the month went up! #impact

PR #6462, opened 2016-04-08
Re-added support for attaching events to document fragments

Wildhoney:
I thought I'd jump straight in and make the change :+1:

In the [open issue](https://github.com/facebook/react/issues/6456#issuecomment-207527336) it was mentioned that @spicyj wrote this part of the code, so it would be awesome if he &mdash; or anybody else &mdash; could chime in about the optimal way of implementing this. Thanks!

PR #15894, opened 2019-06-15
Fix: React cannot render in ShadowRoot

Jack-Works:
Fixes https://github.com/facebook/react/issues/19558

This pr fix:

- <del>ReactDOM.findDOMNode not working well with a `ShadowRoot` node</del>
- ReactDOM ensureListening work in this senario

```html
<custom-element>
    #ShadowRoot: open
        <div>
            <!-- React mount on this node, not the ShadowRoot -->
        </div>
</custom-element>
```

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -271,10 +271,11 @@ export function ensureListeningTo(
   // want to register events to document fragments or documents
   // with the modern plugin event system.
   invariant(
-    (rootContainerElement != null &&
-      rootContainerElement.nodeType === ELEMENT_NODE) ||
-      (rootContainerElement.nodeType === DOCUMENT_FRAGMENT_NODE &&
-        (rootContainerElement: ShadowRoot).mode),
+    rootContainerElement != null &&
+      (rootContainerElement.nodeType === ELEMENT_NODE ||
+        (rootContainerElement.nodeType === DOCUMENT_FRAGMENT_NODE &&
+          // $FlowFixMe GH issue #8457
+          ((rootContainerElement: any): ShadowRoot).mode)),
     'ensureListeningTo(): received a container that was not an element node. ' +
       'This is likely a bug in React.',
   );
@@ -286,7 +287,7 @@ export function ensureListeningTo(
 }
 
 function getOwnerDocumentFromRootContainer(
-  rootContainerElement: Element | Document,
+  rootContainerElement: Element | Document | ShadowRoot,
 ): Document {
   return rootContainerElement.nodeType === DOCUMENT_NODE
     ? (rootContainerElement: any)
@@ -396,7 +397,7 @@ function updateDOMProperties(
 export function createElement(
   type: string,
   props: Object,
-  rootContainerElement: Element | Document,
+  rootContainerElement: Element | Document | ShadowRoot,
   parentNamespace: string,
 ): Element {
   let isCustomComponentTag;
@@ -502,7 +503,7 @@ export function createElement(
 
 export function createTextNode(
   text: string,
-  rootContainerElement: Element | Document,
+  rootContainerElement: Element | Document | ShadowRoot,
 ): Text {
   return getOwnerDocumentFromRootContainer(rootContainerElement).createTextNode(
     text,
@@ -1212,7 +1213,7 @@ export function warnForUnmatchedText(textNode: Text, text: string) {
 }
 
 export function warnForDeletedHydratableElement(
-  parentNode: Element | Document,
+  parentNode: Element | Document | ShadowRoot,
   child: Element,
 ) {
   if (__DEV__) {
@@ -1229,7 +1230,7 @@ export function warnForDeletedHydratableElement(
 }
 
 export function warnForDeletedHydratableText(
-  parentNode: Element | Document,
+  parentNode: Element | Document | ShadowRoot,
   child: Text,
 ) {
   if (__DEV__) {
@@ -1246,7 +1247,7 @@ export function warnForDeletedHydratableText(
 }
 
 export function warnForInsertedHydratedElement(
-  parentNode: Element | Document,
+  parentNode: Element | Document | ShadowRoot,
   tag: string,
   props: Object,
 ) {
@@ -1264,7 +1265,7 @@ export function warnForInsertedHydratedElement(
 }
 
 export function warnForInsertedHydratedText(
-  parentNode: Element | Document,
+  parentNode: Element | Document | ShadowRoot,
   text: string,
 ) {
   if (__DEV__) {
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -13,7 +13,6 @@ import {
 } from '../events/EventRegistry';
 
 import {canUseDOM} from 'shared/ExecutionEnvironment';
-import invariant from 'shared/invariant';
 
 import {
   getValueForAttribute,
@@ -257,7 +256,7 @@ if (__DEV__) {
 }
 
 export function ensureListeningTo(
-  rootContainerInstance: Element | Node | ShadowRoot,
+  rootContainerInstance: Element | Node,
   reactPropEvent: string,
   targetElement: Element | null,
 ): void {
@@ -267,18 +266,19 @@ export function ensureListeningTo(
     rootContainerInstance.nodeType === COMMENT_NODE
       ? rootContainerInstance.parentNode
       : rootContainerInstance;
-  // Containers should only ever be element nodes. We do not
-  // want to register events to document fragments or documents
-  // with the modern plugin event system.
-  invariant(
-    rootContainerElement != null &&
-      (rootContainerElement.nodeType === ELEMENT_NODE ||
-        (rootContainerElement.nodeType === DOCUMENT_FRAGMENT_NODE &&
-          // $FlowFixMe GH issue #8457
-          ((rootContainerElement: any): ShadowRoot).mode)),
-    'ensureListeningTo(): received a container that was not an element node. ' +
-      'This is likely a bug in React.',
-  );
+  if (__DEV__) {
+    if (
+      rootContainerElement == null ||
+      (rootContainerElement.nodeType !== ELEMENT_NODE &&
+        // This is to support rendering into a ShadowRoot:
+        rootContainerElement.nodeType !== DOCUMENT_FRAGMENT_NODE)
+    ) {
+      console.error(
+        'ensureListeningTo(): received a container that was not an element node. ' +
+          'This is likely a bug in React. Please file an issue.',
+      );
+    }
+  }
   listenToReactEvent(
     reactPropEvent,
     ((rootContainerElement: any): Element),
@@ -287,7 +287,7 @@ export function ensureListeningTo(
 }
 
 function getOwnerDocumentFromRootContainer(
-  rootContainerElement: Element | Document | ShadowRoot,
+  rootContainerElement: Element | Document,
 ): Document {
   return rootContainerElement.nodeType === DOCUMENT_NODE
     ? (rootContainerElement: any)
@@ -312,7 +312,7 @@ export function trapClickOnNonInteractiveElement(node: HTMLElement) {
 function setInitialDOMProperties(
   tag: string,
   domElement: Element,
-  rootContainerElement: Element | Document | ShadowRoot,
+  rootContainerElement: Element | Document,
   nextProps: Object,
   isCustomComponentTag: boolean,
 ): void {
@@ -397,7 +397,7 @@ function updateDOMProperties(
 export function createElement(
   type: string,
   props: Object,
-  rootContainerElement: Element | Document | ShadowRoot,
+  rootContainerElement: Element | Document,
   parentNamespace: string,
 ): Element {
   let isCustomComponentTag;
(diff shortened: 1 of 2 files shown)


sravannerella:
Is this going to fix the input event propagation of shadow DOM?

Jack-Works:
> Is this going to fix the input event propagation of shadow DOM?

Let me have a try. And I also need a rebase on the react codebase

Jack-Works:
> Is this going to fix the input event propagation of shadow DOM?

I don't get your meaning by "fix the input event propagation". My change fix the following example. Checkout if it is what you mean

```html
<html>

<body>
  <script src="../../../build/node_modules/react/umd/react.development.js"></script>
  <!-- The latest version on unpkg won't work on this example. -->
  <!-- <script src="https://unpkg.com/react-dom/umd/react-dom.development.js"></script> -->
  <script src="../../../build/node_modules/react-dom/umd/react-dom.development.js"></script>
  <script src="https://unpkg.com/babel-standalone@6/babel.js"></script>
  <div id="container"></div>
  <custom-element>
    #ShadowRoot: open
    <div>
      <!-- React mount on this node, not the ShadowRoot -->
    </div>
  </custom-element>
  <script>
    var s = document.querySelector('custom-element').attachShadow({ mode: 'open' })
    var d = document.createElement('div')
    s.appendChild(d)
  </script>
  <script type="text/babel">
    function Comp() {
      const [s, ss] = React.useState('input')
      return <span>{s}<input value={s} onChange={e => ss(e.target.value)} /></span>
    }
      ReactDOM.render(
        <Comp />,
        d
      );
    </script>
</body>

</html>
```

Jack-Works:
Don't close it.

bvaughn:
cc @trueadm on this one

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -264,14 +264,31 @@ if (__DEV__) {
   };
 }
 
+/**
+ * If the root container is inside of a ShadowRoot, events will lost.
```
trueadm: typo: "events will lost."

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -264,14 +264,31 @@ if (__DEV__) {
...
+ * Recursively access its' parentNode to check if it's inside a ShadowRoot.
```
trueadm: "its" rather than "its'"

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -264,14 +264,31 @@ if (__DEV__) {
...
+ * then attach it to a ShadowRoot, events will still lost.
+ */
+function isInDocumentFragment(element?: Element | Node | null) {
+  if (!element) {
+    return false;
+  }
+  if (element.nodeType === DOCUMENT_FRAGMENT_NODE) {
+    return true;
+  }
+  return isInDocumentFragment(element.parentNode);
```
trueadm: It's better to do this in a while loop rather than recursively.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -264,14 +264,31 @@ if (__DEV__) {
...
+}
+
 function ensureListeningTo(
   rootContainerElement: Element | Node,
   registrationName: string,
 ): void {
-  const isDocumentOrFragment =
+  const isDocumentOrInDocumentFragment =
```
trueadm: `isDocumentOrInFragment` sounds better.

Jack-Works:
It seems like jsdom doesn't support shadow dom ( https://github.com/jsdom/jsdom/issues/1030 ) so I can't write a test for that

packages/react-dom/src/client/ReactDOMComponent.js
Jack-Works: hmm, a dom has a limited set of parents so the recursive chain won't be too long. And this recursive function doesn't introduce any new variables in itself and have clear quit condition so the recursive is safe I think

trueadm: A DOM node can have a huge DOM tree, with thousands of parents. I still feel we should use a while loop here, as the extra bytes will be neglible.

trueadm:
Looking at this again, it clashes with some of the work we're looking into now in regards to moving event registration away from the document and to roots instead. I'd rather come back to this PR in the future (if needed).

Jack-Works:
> Looking at this again, it clashes with some of the work we're looking into now in regards to moving event registration away from the document and to roots instead. I'd rather come back to this PR in the future (if needed).

https://github.com/facebook/react/commit/fc01ed92aeb731aa9bf8b4343edd166405a09e3c if so, can you cherry pick this commit at least?

Jack-Works:
In https://github.com/facebook/react/commit/62861bbcc752c63728592fdc45078854f9aae161#diff-adeaf9850a891faf89a84c73308e0e0bR284 @trueadm mentioned: 
> We do not want to register events to document fragments or documents with the modern plugin event system. 

Why? Rejecting document fragments will disable the ability that React to render in a ShadowRoot.

Jack-Works:
@shivaylamba @trueadm rebase. 

Due to conflict with React new event system, the other 3 commits are removed. Only ce0f27213a301c3e584387040c93a50e77e3b76d (fix `findDOMNode` think DocumentFragment is a "ReactComponent") left.

Again, why throws on document fragments container? It's useful when try to wrap the React component into a web component.

Jack-Works:
bump

trueadm:
> Due to conflict with React new event system, the other 3 commits are removed. Only ce0f272 (fix findDOMNode think DocumentFragment is a "ReactComponent") left.

So are those changed not needed anymore with the new event system?

> Again, why throws on document fragments container? It's useful when trying to wrap the React component into a web component.

I'm not sure what you mean here. Document fragments are only ever temporary. As soon as you append them into the document they lose their children and becomes useless. You can still use fragments to insert children, you should just not be using them with React managed elements.

I noticed this PR only contains the changes for findDOMNode. Given that findDOMNode is deprecated, I'm not sure how important these changes are, as we ideally want folks to move off this API.

Jack-Works:
> So are those changed not needed anymore with the new event system?

I have no idea. I'll try it when I have time for it. (Is the new event system stable now?)

image


> I'm not sure what you mean here. Document fragments are only ever temporary. As soon as you append them into the document they lose their children and becomes useless. You can still use fragments to insert children, you should just not be using them with React managed elements.

As you can see, ShadowRoot is a kind of DocumentFragment. Therefore disabling DocumentFragment also removes the support for ShadowRoot.

> I noticed this PR only contains the changes for findDOMNode. 

Yes, the change in the old event system has removed.

> Given that findDOMNode is deprecated, I'm not sure how important these changes are, as we ideally want folks to move off this API.

I'm not using `findDOMNode` but the usage is in my dependencies so it's not easy to fix.

trueadm:
> I'm not using findDOMNode but the usage is in my dependencies so it's not easy to fix.

Let's remove the findDOMNode changes then.

> As you can see, ShadowRoot is a kind of DocumentFragment. Therefore disabling DocumentFragment also removes the support for ShadowRoot.

Gotcha. So in this case, we should add a check for cases were we want to disallow a normal document fragment, but allow shadow roots? I believe we can check for the existance of the `mode` property rather than using `instanceof`. `instanceof` won't work between iframes.

I'll wait for you to see if it works. It would be good if you could add a test around this, you could make a fragment and then add the correct `mode` and other shadow root properties, to ensure it passes and doesn't break.

Jack-Works:
> Let's remove the findDOMNode changes then.

Hmm I mean I can't know if all of my dependencies are using this function it's better to fix it in react.

> It would be good if you could add a test around this

Where does the test run in? IIRC JSDom didn't implement ShadowDOM yet.

trueadm:
> Hmm I mean I can't know if all of my dependencies are using this function it's better to fix it in react.

I mean, can we put that change in another PR as I feel the two changes are big enough to be tested separately. The change to findDOMNode is unrelated to the event system and given both these changes are probably breaking changes, it would be good to break them up if that's okay?

> Where does the test run in? IIRC JSDom didn't implement ShadowDOM yet.

Ah yes, JSDOM hasn't caught up. You can probably add a fixture test that makes use of the real DOM and demonstrates this use-case. We have a bunch of other fixture tests you can look at for pointers.

gaearon:
`findDOMNode` is effectively deprecated in modern code. There is no need to make changes to it.

Jack-Works:
@trueadm I made a fix, but I'm not sure if this fix covers all usage with ShadowRoot. I have tried the following mounting ways:

```html
<!-- Working! -->
#ShadowRoot: open (React Root)
    <div onClick={x => console.log(x)}>Content</div>

<!-- Working! -->
#ShadowRoot: open
    <div> (React Root)
        <div onClick={x => console.log(x)}>Content</div>
    </div>

<!-- Working! -->
#ShadowRoot: open (React Portal Root)
    <div onClick={x => console.log(x)}>Content</div>

<!-- Working! -->
#ShadowRoot: open
    <div> (React Portal Root)
        <div onClick={x => console.log(x)}>Content</div>
    </div>
```

But I found a strange problem:

```jsx
function App() {
  return <h1 onClick={console.log}>h1</h1>;
  // console.log is replaced by "disableLog" so I can't log any event.
  return <h1 onClick={x => console.log(x)}>h1</h1>;
  // have to do this
}
```

trueadm:
> console.log is replaced by "disableLog" so I can't log any event.

That's to be expected, React replaces `console.log` during render. Just keep with the callback approach, as in reality the render should be side-effect free and `console.log` is actually a form of side-effect (and one only used really for debugging).

Jack-Works:
The console replacement part is confusing to me. You can't cancel all side effects after all, this is making people(me) confused and doesn't really resolve the side effect problem.

I'll resolve the CI error and add tests later this week!

trueadm:
> The console replacement part is confusing to me. You can't cancel all side effects after all, this is making people(me) confused and doesn't really resolve the side effect problem.

This is something we do in DEV to avoid confusion around double renders (to catch side-effects that might cause problems for concurrent mode). Last time I checked, in regards to console.log, we over-ride it so you don't get duplicate console logs. Duplicate console logs might be even more confusing otherwise.

bvaughn:
@trueadm is correct. React only overrides the console methods for the second render, which it only does in DEV. This prevents duplicate logs from being in the console and confusing people. We understand that it might be confusing as well to see a `console.log` statement that doesn't print anything– but we think that's a less common case.

Jack-Works:
> React only overrides the console methods for the second render, which it only does in DEV. This prevents duplicate logs from being in the console and confusing people.

Actually, duplicate logs will warn the developer about they shouldn't do side effect things out of `useEffect`. It actually helps developers to figure out what is "double render" or things about the concurrent mode. I think React should at lease preserve the log and warn the developer instead of drop everything they logged, it's super confusing

bvaughn:
> Actually, duplicate logs will warn the developer about they shouldn't do side effect things

We test changes like this within Facebook before we deploy them to open source. From early testing experience, duplicate logs confuse developers into thinking that either React or their application code is broken and triggering multiple renderers. Silencing the log dramatically reduces this.

Jack-Works:
> Silencing the log dramatically reduces this.

It's not resolving the problem. It's hiding the problem. And people (me) start to thinking `return <h1 onClick={console.log}>h1</h1>;` why the event is not fired.

Another approach might be, if it is not in the rendering stage, let `disableLog` function behave like a normal log, this will reduce the harm.

trueadm:
I feel like this discussion had gone off track. If you have concerns about the behavior around the `console.log` heuristics, maybe it's best to put these concerns in another issue?

Jack-Works:
Discussion about this moved to #19537

eps1lon:
This PR fixes https://github.com/facebook/react/issues/19558

Jack-Works:
@trueadm Hi I have fixed the flow type problem, but I have no idea about how to add tests.

The basic idea is

```js
const x = document.querySelector('#root');
const shadow = x.attachShadow({mode: 'open'});

ReactDOM.unstable_createRoot(shadow).render(<App />);
function App() {
  return <h1 onClick={console.trace}>h1</h1>;
}
```

If the click invokes the function, the test passes.

gaearon:
Apparently we broke this before and it was last fixed in https://github.com/facebook/react/pull/6462.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -256,7 +257,7 @@ if (__DEV__) {
 }
 
 export function ensureListeningTo(
-  rootContainerInstance: Element | Node,
+  rootContainerInstance: Element | Node | ShadowRoot,
```
gaearon, line 260: This is unnecessary. `ShadowRoot` is already a `Node`. (So is element, fwiw)

gaearon:
I pushed a slightly different fix:

- The comment right above this code was stale and repeated the code, so I removed it
- There is no need for this to be a production error, so I changed it to a DEV warning
  - This also means we don't have to pay for this bugfix with production bytes
- Removed the type changes since this doesn't affect anything else and we'd like to keep types simpler

gaearon:
Btw sorry this took long to review; I misunderstood the report in https://github.com/facebook/react/issues/19558 and thought this was already broken in 16 because both links led to the same sandbox.

eps1lon:
> Btw sorry this took long to review; I misunderstood the report in #19558 and thought this was already broken in 16 because both links led to the same sandbox.

My bad, sorry. I have to be more careful when forking codesandbox for different React versions

gaearon:
Haha no problem, I do that all the time

PR #15899, opened 2019-06-16
Add additional check in warning for changed bits so users can use bigint in their code

zmitry:
I wanted to use big int in my context value but it seems that some minor checks will throw an error but it seems to be pretty minor check which will allow end user to try big int for observed bits


acdlite:
Thanks but we intentionally limited this to 31 bits, for both speed and memory reasons. It's also an unstable feature so you probably shouldn't rely on it unless you're prepared for us to remove it in the future.

PR #15902, opened 2019-06-17
Add SuspenseList Component

sebmarkbage:
This component lets you coordinate the loading sequence of nested Suspense boundaries. How this happens is defined by the `revealOrder` property. By default, it behaves like a noop so the nested boundaries just do what they normally would.

It operates on the nearest most Suspense boundary but you can actually have intermediate components and DOM nodes. If the nearest nested boundary is another SuspenseList, it operates on the nested Suspense boundary within that.

This first PR implements the `"together"` option which ensures that all boundaries resolve together. If any of them trigger a fallback, they all do.

Subsequent PRs will implement ordered rendering row-by-row. However, even in those options when there are insertions in the middle or existing rows triggering new fallbacks in the middle, there will be a "together" pass to resolve those first before the row-by-row kicks in. So the `"together"` mode is the primitive needed for those modes too.

The mechanism for this mode is that we do a full render pass first, and then scan the result to see if there are any fallbacks. If there are, we do a full second render pass (including re-reconciliation). There are many complex scenarios that can happen such as having existing trees staying in fallbacks, and new props coming in.

The second pass currently deletes all work that we did in the first pass so it's quite in efficient but this particular mode is expected to be rare and not have large subtrees. It's expected that there are not that many abstractions until we hit the first Suspense boundary. We're Suspended anyway.

Unfortunately this does a shallow scan even if we're just updating something nested inside the tree. I'd like to add an optimization if nothing has suspended during the first render pass and we know from previous results that nothing has suspended. Not quite sure exactly when that's safe yet.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -31,6 +31,7 @@ import {
   ContextConsumer,
   Profiler,
   SuspenseComponent,
+  SuspenseListComponent,
   DehydratedSuspenseComponent,
   MemoComponent,
   SimpleMemoComponent,
@@ -1918,6 +1919,22 @@ function updateDehydratedSuspenseComponent(
   }
 }
 
+function updateSuspenseListComponent(
+  current: Fiber | null,
+  workInProgress: Fiber,
+  renderExpirationTime: ExpirationTime,
+) {
+  // TODO
+  const nextChildren = workInProgress.pendingProps.children;
+  reconcileChildren(
+    current,
+    workInProgress,
+    nextChildren,
+    renderExpirationTime,
+  );
+  return workInProgress.child;
+}
+
 function updatePortalComponent(
   current: Fiber | null,
   workInProgress: Fiber,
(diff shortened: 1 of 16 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1918,6 +1936,147 @@ function updateDehydratedSuspenseComponent(
...
+      // If we previously forced a fallback, we need to schedule work
+      // on any nested boundaries to let them know to try to render
+      // again. This is the same as context updating.
+      propagateSuspenseContextChange(
+        workInProgress,
+        nextChildFibers,
+        renderExpirationTime,
+      );
+    }
+    workInProgress.memoizedState = false;
```
gaearon: Is this going to mess with hidden class?

sebmarkbage: Is this a new comment? What happened to the other one that I already answered?

sebmarkbage:
I realized that this probably doesn't work in legacy mode. I'll probably just disable it somehow.

packages/react-reconciler/src/ReactFiberBeginWork.js
gaearon: Lol. I thought I hadn't send it because I couldn't find it today either (and didn't see your reply to it).

sebmarkbage: I made it an object since I will do that in the next PR anyway.

PR #15904, opened 2019-06-17
[Fresh] Rename findHostNodesForHotUpdate > findHostInstancesForHotpdate

gaearon:
This doesn't really matter because it wasn't released yet and isn't used anywhere.
Renaming for consistency with other DevTools methods which call them "instances".


NE-SmallTown:
The issue title 'findHostInstancesForHotpdate' should be 'findHostInstancesForHotdate'

PR #15909, opened 2019-06-18
[Flare] add disableContextMenu to Press

necolas:
Provides a way to entirely disable the native context menu.


trueadm:
This would change the previous behavior of preventing the context menu from appearing on a long tap (touch). Is that intended?

necolas:
How does it change that behaviour?

trueadm:
Sorry, ignore me, misread it on my phone!

PR #15910, opened 2019-06-18
[Scheduler] Use continuation pattern for posting host callback

acdlite:
As I prepare to implement integrated timers, I noticed some peculiarities in the Scheduler implementation that could afford to be cleaned up.

This is a refactor and shouldn't affect any observable behavior (except maybe accidental bugs); mostly it removes some concepts that existed in earlier iterations of Scheduler and are no longer needed.

The main change is to how the DOM implementation schedules an additional callback before yielding to the main thread. It used to follow the same code path for scheduling task; now it has its own branch directly inside the message event handler. The special case for error handling — where we call `postMessage` immediately without waiting for rAF — has similarly been localized inside the catch block of the message event handler.

It probably helps to read the commits one at a time.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -11,7 +11,6 @@
 import {enableSchedulerDebugging} from './SchedulerFeatureFlags';
 import {
   requestHostCallback,
-  cancelHostCallback,
   shouldYieldToHost,
   getCurrentTime,
   forceFrameRate,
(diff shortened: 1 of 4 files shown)


packages/scheduler/src/Scheduler.js
```diff
@@ -92,57 +90,56 @@ function flushTask(task) {
...
     } else {
       var nextAfterContinuation = null;
-      var task = firstTask;
+      var t = firstTask;
       do {
-        if (task.expirationTime >= expirationTime) {
-          // This task expires at or after the continuation. We will insert the
-          // continuation *before* this task.
-          nextAfterContinuation = task;
+        if (absoluteTimeout <= t.startTime + t.timeout) {
```
sebmarkbage: This forces you to do a computation each time you want to check it instead of only once. Is there a reason you don't just do startTime and expirationTime? In case the task doesn't gets added to the queue at exactly the startTime? In that case it seems like it should be increasing in priority since it's running behind.

acdlite: Yeah that makes sense. I think originally I thought I would the timeout after creation.

PR #15911, opened 2019-06-18
[Scheduler] Integrated timers

acdlite:
## Based on #15910 

Adds a `delay` option to `scheduleCallback`. When specified, the task is not scheduled until after the delay has elapsed.

Delayed tasks are scheduled on a timer queue maintained by Scheduler, instead of directly with the browser. The main benefit is to reduce the number of native browser timers that Scheduler's `message` event handler has to contend with; so, after yielding to the browser at the end of the frame, Scheduler will more quickly regain control of the main thread. Because we're able to flush the timer queue without yielding to browser timer events, there's also less task switching overhead (though in the absence of `isInputPending`, this is mostly a theoretical win since we yield every frame regardless).

If the queue of non-delayed tasks is non-empty — that is, if there is pending CPU bound work — Scheduler is able to avoid a browser timer entirely by periodically checking its own timer queue while flushing tasks (inside the `message` event handler). Once the CPU-bound work is complete, if there are still pending delayed tasks, Scheduler will schedule a single browser timer that fires once the earliest delay has elapsed.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -11,6 +11,8 @@
 import {enableSchedulerDebugging} from './SchedulerFeatureFlags';
 import {
   requestHostCallback,
+  requestHostTimeout,
+  cancelHostTimeout,
   shouldYieldToHost,
   getCurrentTime,
   forceFrameRate,
@@ -39,6 +41,7 @@ var IDLE_PRIORITY = maxSigned31BitInt;
 
 // Tasks are stored as a circular, doubly linked list.
 var firstTask = null;
+var firstDelayedTask = null;
 
 // Pausing the scheduler is useful for debugging.
 var isSchedulerPaused = false;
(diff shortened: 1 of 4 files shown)


packages/scheduler/src/Scheduler.js
```diff
@@ -72,28 +72,28 @@ function scheduleHostCallbackIfNeeded() {
   }
 }
 
-function flushFirstTask() {
-  const currentlyFlushingTask = firstTask;
-
+function flushTask(task) {
```
acdlite: Realizing now that the commit message doesn't really make sense anymore. Originally I thought I would need to be able to flush a task at an arbitrary point in the list, but that turned out not to be the case. I'll keep the change for type safety reasons, though.

acdlite:
sizebot is a bit misleading: The `react` package is unaffected except for the UMD build which includes a copy of `scheduler`.

Scheduler did go up a bit, but that was expected since it includes a new feature.

PR #15916, opened 2019-06-18
[Flare] Press events include defaultPrevented

necolas:
Also changes the behaviour of 'disableContextMenu' so that 'onContextMenu' is still called.

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -19,7 +19,6 @@ import {DiscreteEvent, UserBlockingEvent} from 'shared/ReactTypes';
 
 type PressProps = {
   disabled: boolean,
-  disableContextMenu: boolean,
   delayLongPress: number,
   delayPressEnd: number,
   delayPressStart: number,
@@ -38,6 +37,7 @@ type PressProps = {
     bottom: number,
     left: number,
   },
+  preventContextMenu: boolean,
   preventDefault: boolean,
   stopPropagation: boolean,
 };
@@ -72,6 +72,7 @@ type PressState = {
   |}>,
   ignoreEmulatedMouseEvents: boolean,
   activePointerId: null | number,
+  shouldPreventClick: boolean,
 };
 
 type PressEventType =
@@ -85,6 +86,7 @@ type PressEventType =
   | 'contextmenu';
 
 type PressEvent = {|
+  defaultPrevented: boolean,
   target: Element | Document,
   type: PressEventType,
   pointerType: PointerType,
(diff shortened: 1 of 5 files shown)


packages/react-events/src/Press.js
```diff
@@ -863,6 +871,30 @@ const PressResponder = {
...
+            // Check "open in new window/tab" and "open context menu" key modifiers
+            const preventDefault = props.preventDefault;
+            if (
+              preventDefault !== false &&
+              !shiftKey &&
+              !metaKey &&
+              !ctrlKey &&
+              !altKey
+            ) {
+              state.shouldPreventDefault = true;
```
necolas: This state needs to be reset somewhere

trueadm: It should be reset in `pointerup` right?

necolas: This is within the `pointerup` event processing. It looks like this code is only for handling `click`, so I can make it explicitly for that and reset it in the `click` block.

trueadm: No I mean, we should reset `shouldPreventDefault` to `false` in `pointerup` if we don't set it to `true`. That way we don't wrongly prevent things we shouldn't.

packages/react-events/src/Press.js
```diff
@@ -906,27 +939,9 @@ const PressResponder = {
...
-            !shiftKey &&
-            !metaKey &&
-            !ctrlKey &&
-            !altKey
-          ) {
-            nativeEvent.preventDefault();
-          }
+        if (state.shouldPreventClick) {
+          nativeEvent.preventDefault();
+          state.shouldPreventClick = false;
```
trueadm: We don't need this anymore. :)

trueadm:
On a side note: `disableContextMenu` -> `preventContextMenu` seems somewhat more fitting now that we're not disabling the `contentmenu` event from firing, rather we're just preventing an the native occurrence of it.

necolas:
This doesn't change any of the existing logic around that. And the `Press` event component doesn't accept multiple children.

trueadm:
@necolas I had a typo in my above example, I updated it after you commented (you were too quick).

necolas:
I'll check for any unexpected changes internally before landing this

necolas:
Looked into some of the questions we had…

The emulated `click` event shouldn't be dispatched multiple times for multi-touch interactions ([source](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Using_Touch_Events)).

> Browsers typically dispatch emulated mouse and click events when there is only a single active touch point. Multi-touch interactions involving two or more active touch points will usually only generate touch events. 

And `click` won't be dispatched if the `*down` event didn't start in the elements subtree.

trueadm:
@necolas Thanks for sanity checking it! :)

PR #15917, opened 2019-06-18
[Flare] Relable Press preventDefault + stopPropagation props

trueadm:
This PR does the follow:

- Renames`disableContextMenu` to `disableNativeContextMenu` and no longer prevents the `onContextMenu` event from firing if the prop is `false`.
- Renames `preventDefault` to `disableNativePropagation`
- Renames `stopPropagation` to `disableNativePropagation`

The idea being, is that using a slightly different name from the native DOM methods makes it clear that these work in slightly different ways that fit this new event system's needs, rather than needs of the user writing imperative DOM event handling.

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -19,7 +19,9 @@ import {DiscreteEvent, UserBlockingEvent} from 'shared/ReactTypes';
 
 type PressProps = {
   disabled: boolean,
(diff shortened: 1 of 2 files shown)


trueadm:
Closing for now. Let's reduce churn on this, but keep a reference to this PR so we can come back to it if needed.

PR #15918, opened 2019-06-19
Add forwards and backwards options to SuspenseList

sebmarkbage:
Builds on top of #15902

I'm not super happy with this implementation yet but it seems to work. I also need more tests.

So putting it out there as an early review.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -2005,7 +2005,7 @@ function findLastContentRow(firstChild: null | Fiber): null | Fiber {
   return lastContentRow;
 }
 
-type SuspenseListRevealOrder = 'forwards' | 'together' | void;
+type SuspenseListRevealOrder = 'forwards' | 'backwards' | 'together' | void;
 
 function updateSuspenseListComponent(
   current: Fiber | null,
@@ -2055,6 +2055,7 @@ function updateSuspenseListComponent(
     );
     suspenseListState = {
       didSuspend: true,
+      isBackwards: false,
       rendering: null,
       last: null,
       tail: null,
@@ -2109,6 +2110,7 @@ function updateSuspenseListComponent(
         if (suspenseListState === null) {
           suspenseListState = {
             didSuspend: false,
+            isBackwards: false,
             rendering: null,
             last: lastContentRow,
             tail: tail,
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1918,6 +1940,259 @@ function updateDehydratedSuspenseComponent(
...
+      lastContentRow = row;
+    }
+    row = row.sibling;
+  }
+  return lastContentRow;
+}
+
+type SuspenseListRevealOrder = 'forwards' | 'backwards' | 'together' | void;
+
+function updateSuspenseListComponent(
```
acdlite: Can you put a comment up here with an overview of the different render passes so the next person has some idea what's going on? Plz

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1918,6 +1940,259 @@ function updateDehydratedSuspenseComponent(
...
+
+  let shouldForceFallback = hasSuspenseContext(
+    suspenseContext,
+    (ForceSuspenseFallback: SuspenseContext),
+  );
+
+  if ((workInProgress.effectTag & DidCapture) !== NoEffect) {
+    // This is the second pass. In this pass, we should force the
+    // fallbacks in place.
+    shouldForceFallback = true;
```
acdlite: Factoring nit: This function might be easier to follow if you got rid of this boolean and put the first pass in the `else` branch of this condition (or vice versa). Since nearly all the code below only applies to the second pass.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -834,6 +921,90 @@ function completeWork(
...
+            workInProgress.effectTag |= DidCapture;
+            // Reset the effect list before doing the second pass since that's now invalid.
+            workInProgress.firstEffect = workInProgress.lastEffect = null;
+            // Schedule work so we know not to bail out.
+            workInProgress.expirationTime = renderExpirationTime;
+            return workInProgress;
+          }
+        } else {
+          // Append the rendered row to the child list.
+          let rendered = suspenseListState.rendering;
```
acdlite: Nit: Hoist this above the first branch so we only read it once.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -834,6 +921,90 @@ function completeWork(
...
+          if (!suspenseListState.didSuspend) {
+            suspenseListState.didSuspend = isShowingAnyFallbacks(rendered);
```
acdlite: Same here I guess :D

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -834,6 +921,90 @@ function completeWork(
...
+            if (previousSibling !== null) {
+              previousSibling.sibling = rendered;
+            } else {
+              workInProgress.child = rendered;
+            }
+            suspenseListState.last = rendered;
+          }
+        }
+
+        if (suspenseListState !== null && suspenseListState.tail !== null) {
```
acdlite: and here #ireviewcodegood

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -121,13 +125,16 @@ import {
...
+  setShallowSuspenseContext,
 } from './ReactFiberSuspenseContext';
+import {isShowingAnyFallbacks} from './ReactFiberSuspenseComponent';
 import {
   pushProvider,
   propagateContextChange,
   readContext,
   prepareToReadContext,
   calculateChangedBits,
+  scheduleWorkOnParentPath,
```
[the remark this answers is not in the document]
sebmarkbage: The other method does more things and doesn't bailout until it hits the root since it needs to return the root and schedule work on the root and possibly with scheduler. This path doesn't need that. So I think I'll leave this in and hope that the VM can inline it instead.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1433,6 +1442,22 @@ function validateFunctionComponentInDev(workInProgress: Fiber, Component: any) {
 // TODO: This is now an empty object. Should we just make it a boolean?
 const SUSPENDED_MARKER: SuspenseState = ({}: any);
 
+function shouldRemainOnFallback(
```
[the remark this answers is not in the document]
sebmarkbage: I found it difficult to read, especially if it's unnamed. The logic is fairly isolated to the suspense list case where as the effect tag is about which pass we're in.

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: I'll leave all the factoring stuff for now. I need to add the other features and then I'll do another pass to do all the optimizations which will likely require much further refactoring.

PR #15927, opened 2019-06-19
[Flare] Add useEvent hook implementation

trueadm:
This PR adds a new primitive hook to React called `useEvent`. For now it's behind the React Flare flag and is labelled `unstabled_useEvent` on the React object. This implementation might have bugs in, as it's not been hooked up and tested with the various event responder modules, like Press, Hover etc. This isn't intended to be a bug-free PR, more it's an initial implementation that should aim to get the wiring right between fibers and the hook.

The principal way of using this, is to leverage existing event components within a component. For example:

```jsx
import React, {unstable_useEvent} from 'react';
import Press from 'react-events/press';

function MyComponent() {
  unstable_useEvent(Press, {onPress: handlePressFromHook});

  function handlePressFromHook() {
    console.log('hook');
  } 
  function handlePressFromEventComponent() {
    console.log('event component');
  } 
  
  return (
    <Press onPress={handlePressFromEventComponent}>
      <div>Press me</div>
    </Press>
  );
}
```

Any event components in the **same** branch of an event component (not above the event component), will fire for the same type of event component. If there are no event components used in the branch of the function component that has the hooks, then the hook props will never fire. So this won't work:

```jsx
function MyComponent() {
  unstable_useEvent(Press, {onPress: handlePressFromHook});
  
  // You need to use a <Press> somewhere in this branch
  return (
    <div>Press me</div>
  );
}
```

What is powerful about the hook form of event components, is that they can track events occurring in their tree without needing to worry about propagation rules. All events will propagate to the hook form of these events, which is different to the event component form, where events stop propagation to the same type of event components.

Event responders can opt out of support hooks, for example `FocusScope` will not support hook form right now, so the responder module has `allowEventHooks` set to `false`.

The hook forms thus allow for control of events on children, where their implementation is not fully known or controller by the parent component. The only exception to this, is if an event responder takes global ownership, then no other events components or hooks will fire – this is intentional.

Furthermore, `displayName` is now no longer on event components and has instead been moved to the responders. This is to allow for hooks to have display names, via the responder.

### DIFF
--- a/packages/react-reconciler/src/ReactFiber.js
+++ b/packages/react-reconciler/src/ReactFiber.js
@@ -21,7 +21,7 @@ import type {TypeOfMode} from './ReactTypeOfMode';
 import type {SideEffectTag} from 'shared/ReactSideEffectTags';
 import type {ExpirationTime} from './ReactFiberExpirationTime';
 import type {UpdateQueue} from './ReactUpdateQueue';
-import type {ContextDependencyList} from './ReactFiberNewContext';
+import type {ContextDependency} from './ReactFiberNewContext';
 import type {HookType} from './ReactFiberHooks';
 import type {ReactEventComponentInstance} from 'shared/ReactTypes';
 
@@ -103,6 +103,12 @@ if (__DEV__) {
   }
 }
 
+export type Dependencies = {
+  expirationTime: ExpirationTime,
+  firstContext: ContextDependency<mixed> | null,
+  events: Array<ReactEventComponentInstance> | null,
+};
+
 // A Fiber is work on a Component that needs to be done or was done. There can
 // be more than one per component.
 export type Fiber = {|
@@ -163,11 +169,8 @@ export type Fiber = {|
   // The state used to create the output
   memoizedState: any,
 
-  // A linked-list of contexts that this fiber depends on
-  contextDependencies: ContextDependencyList | null,
-
-  // An array of event component instances for this fiber
-  events: Array<ReactEventComponentInstance> | null,
+  // An object of dependencies for this fiber
+  dependencies: Dependencies,
 
   // Bitfield that describes properties about the fiber and its subtree. E.g.
   // the ConcurrentMode flag indicates whether the subtree should be async-by-
@@ -267,8 +270,11 @@ function FiberNode(
   this.memoizedProps = null;
   this.updateQueue = null;
   this.memoizedState = null;
-  this.contextDependencies = null;
-  this.events = null;
+  this.dependencies = {
+    expirationTime: 0,
+    firstContext: null,
+    events: null,
+  };
 
   this.mode = mode;
 
@@ -437,8 +443,7 @@ export function createWorkInProgress(
   workInProgress.memoizedProps = current.memoizedProps;
   workInProgress.memoizedState = current.memoizedState;
   workInProgress.updateQueue = current.updateQueue;
-  workInProgress.contextDependencies = current.contextDependencies;
-  workInProgress.events = current.events;
+  workInProgress.dependencies = current.dependencies;
 
   // These will be overridden during the parent's reconciliation
   workInProgress.sibling = current.sibling;
@@ -829,8 +834,7 @@ export function assignFiberPropertiesInDEV(
   target.memoizedProps = source.memoizedProps;
   target.updateQueue = source.updateQueue;
   target.memoizedState = source.memoizedState;
-  target.contextDependencies = source.contextDependencies;
-  target.events = source.events;
+  target.dependencies = source.dependencies;
   target.mode = source.mode;
   target.effectTag = source.effectTag;
   target.nextEffect = source.nextEffect;
--- a/packages/react-reconciler/src/ReactFiber.js
+++ b/packages/react-reconciler/src/ReactFiber.js
@@ -23,6 +23,7 @@ import type {ExpirationTime} from './ReactFiberExpirationTime';
 import type {UpdateQueue} from './ReactUpdateQueue';
 import type {ContextDependencyList} from './ReactFiberNewContext';
 import type {HookType} from './ReactFiberHooks';
+import type {ReactEventComponentInstance} from 'shared/ReactTypes';
 
 import invariant from 'shared/invariant';
 import warningWithoutStack from 'shared/warningWithoutStack';
@@ -163,6 +164,9 @@ export type Fiber = {|
   // A linked-list of contexts that this fiber depends on
   contextDependencies: ContextDependencyList | null,
 
+  // An array of event component instances for this fiber
+  events: Array<ReactEventComponentInstance> | null,
+
   // Bitfield that describes properties about the fiber and its subtree. E.g.
   // the ConcurrentMode flag indicates whether the subtree should be async-by-
   // default. When a fiber is created, it inherits the mode of its
@@ -262,6 +266,7 @@ function FiberNode(
   this.updateQueue = null;
   this.memoizedState = null;
   this.contextDependencies = null;
+  this.events = null;
 
   this.mode = mode;
 
@@ -431,6 +436,7 @@ export function createWorkInProgress(
   workInProgress.memoizedState = current.memoizedState;
   workInProgress.updateQueue = current.updateQueue;
   workInProgress.contextDependencies = current.contextDependencies;
+  workInProgress.events = current.events;
 
   // These will be overridden during the parent's reconciliation
   workInProgress.sibling = current.sibling;
@@ -796,6 +802,7 @@ export function assignFiberPropertiesInDEV(
   target.updateQueue = source.updateQueue;
   target.memoizedState = source.memoizedState;
   target.contextDependencies = source.contextDependencies;
+  target.events = source.events;
   target.mode = source.mode;
   target.effectTag = source.effectTag;
   target.nextEffect = source.nextEffect;
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -215,6 +215,10 @@ function useMemo<T>(
   return value;
 }
 
+function useEvent() {
+  throw new Error('TODO: not yet implemented');
+}
+
 const Dispatcher: DispatcherType = {
   readContext,
   useCallback,
@@ -227,6 +231,7 @@ const Dispatcher: DispatcherType = {
   useReducer,
   useRef,
   useState,
+  useEvent,
 };
 
 // Inspect
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -709,7 +709,7 @@ function getTargetEventResponderInstances(
   while (node !== null) {
     // Traverse up the fiber tree till we find event component fibers.
     const tag = node.tag;
-    const events = node.events;
+    const events = node.dependencies.events;
 
     if (tag === EventComponent) {
       const eventComponentInstance = node.stateNode;
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -16,6 +16,7 @@ import {
   EventComponent,
   EventTarget as EventTargetWorkTag,
   HostComponent,
+  FunctionComponent,
 } from 'shared/ReactWorkTags';
 import type {
   ReactEventComponentInstance,
@@ -117,6 +118,7 @@ let currentTimers = new Map();
 let currentInstance: null | ReactEventComponentInstance = null;
 let currentEventQueue: null | EventQueue = null;
 let currentTimerIDCounter = 0;
+let currentDocument: null | Document = null;
 
 const eventResponderContext: ReactDOMResponderContext = {
   dispatchEvent(
@@ -217,15 +219,41 @@ const eventResponderContext: ReactDOMResponderContext = {
     }
     return false;
   },
-  isTargetWithinEventComponent,
-  isTargetWithinEventResponderScope(target: Element | Document): boolean {
+  isTargetWithinEventComponent(target: Element | Document): boolean {
     validateResponderContext();
-    const responder = ((currentInstance: any): ReactEventComponentInstance)
-      .responder;
     if (target != null) {
       let fiber = getClosestInstanceFromNode(target);
+      const currentFiber = ((currentInstance: any): ReactEventComponentInstance)
+        .currentFiber;
+
       while (fiber !== null) {
-        if (fiber.stateNode === currentInstance) {
+        if (fiber.tag === EventComponent) {
+          // Switch to the current fiber tree
+          fiber = fiber.stateNode.currentFiber;
+        }
+        if (fiber === currentFiber || fiber.stateNode === currentInstance) {
+          return true;
+        }
+        fiber = fiber.return;
+      }
+    }
+    return false;
+  },
+  isTargetWithinEventResponderScope(target: Element | Document): boolean {
+    validateResponderContext();
+    const componentInstance = ((currentInstance: any): ReactEventComponentInstance);
+    const responder = componentInstance.responder;
+
+    if (target != null) {
+      let fiber = getClosestInstanceFromNode(target);
+      const currentFiber = ((currentInstance: any): ReactEventComponentInstance)
+        .currentFiber;
+      while (fiber !== null) {
+        if (fiber.tag === EventComponent) {
+          // Switch to the current fiber tree
+          fiber = fiber.stateNode.currentFiber;
+        }
+        if (fiber === currentFiber || fiber.stateNode === currentInstance) {
           return true;
         }
         if (
@@ -384,9 +412,15 @@ const eventResponderContext: ReactDOMResponderContext = {
     const target = event.target;
     let fiber = getClosestInstanceFromNode(target);
     let hostComponent = target;
+    const currentResponder = ((currentInstance: any): ReactEventComponentInstance)
+      .responder;
 
     while (fiber !== null) {
-      if (fiber.stateNode === currentInstance) {
+      const stateNode = fiber.stateNode;
+      if (
+        fiber.tag === EventComponent &&
+        (stateNode === null || stateNode.responder === currentResponder)
+      ) {
         break;
       }
       if (fiber.tag === HostComponent) {
@@ -407,8 +441,16 @@ const eventResponderContext: ReactDOMResponderContext = {
   ): boolean {
     validateResponderContext();
     let fiber = getClosestInstanceFromNode(target);
+    const currentResponder = ((currentInstance: any): ReactEventComponentInstance)
+      .responder;
+
     while (fiber !== null) {
-      if (!deep && fiber.stateNode === currentInstance) {
+      const stateNode = fiber.stateNode;
+      if (
+        !deep &&
+        (fiber.tag === EventComponent &&
+          (stateNode === null || stateNode.responder === currentResponder))
+      ) {
         return false;
       }
       if (fiber.tag === HostComponent && fiber.type === elementType) {
@@ -451,24 +493,8 @@ function collectFocusableElements(
   }
 }
 
-function isTargetWithinEventComponent(target: Element | Document): boolean {
-  validateResponderContext();
-  if (target != null) {
-    let fiber = getClosestInstanceFromNode(target);
-    while (fiber !== null) {
-      if (fiber.stateNode === currentInstance) {
-        return true;
-      }
-      fiber = fiber.return;
-    }
-  }
-  return false;
-}
-
 function getActiveDocument(): Document {
-  const eventComponentInstance = ((currentInstance: any): ReactEventComponentInstance);
-  const rootElement = ((eventComponentInstance.rootInstance: any): Element);
-  return rootElement.ownerDocument;
+  return ((currentDocument: any): Document);
 }
 
 function releaseOwnershipForEventComponentInstance(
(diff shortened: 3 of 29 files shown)


TrySound:
Could you provide some use case for hook form over component only form?

trueadm:
@TrySound How do you mean? You can use the hook form to get events deep in a tree. Event components don't propagate events up the tree to other event components of the same type. Using hooks you can get access to these events that would normally wouldn't propagate.

TrySound:
So will this allow me to close popup menu on every item press like this?
```js
const Component = ({ closeMenu, doAction1, doAction2 }) => {
  useEvent(Press, { onPress: closeMenu })
  return <>
    <Press onPress={doAction1}>
      <button>do action 1</button>
     </Press>
    <Press onPress={doAction2}>
      <button>do action 2</button>
    </Press>
  </>
}
```

trueadm:
@TrySound Yep, that's right :)

alazzawimahmoud:
This will decline the need for lots of custom hooks out there, very promising!

kimochg:
Would it support custom event? if so, it brings a total new control flow to react. For now, upper  component need pass down event handler through props to be notified the changes from descendant components, which leads to hoist some states to upper component. If I understand this pr correctly, would it eliminate the state hoisting staff, in this way, the states keep inside descendant component, while ancestor component can observe them?

devongovett:
I like the idea here, but the API feels a bit backwards from what we're used to with hooks, so I thought it might be worth pointing out.

All of the existing hooks in React get data from or affect parent components or the component they are called from: never their children. For example, `useContext` gets data from a parent, and `useEffect` affects the current component. These new event hooks work the other way: they get data (events) from children. To me, it feels a bit weird to use hooks as the API for event bubbling as opposed to components. It breaks the React model of unidirectional data flow.

It's also a bit confusing that there are two ways of using events: components and hooks.

* The difference between them and when to use one or the other is unclear at first glance. The only major difference is event propagation as far as I can tell.
* It's unclear (e.g. in the above example by @TrySound) what the order of events is since the hook is not part of the tree. Does `closeMenu` run before or after `doAction1` or `doAction2`?
* It's unclear if event propagation can be controlled in any way. Can `doAction1` prevent propagation to `closeMenu`? Is it possible for the hook to prevent further propagation to parent hooks? There are very legit use cases for this.

Event components with normal event propagation feel like the more natural way. In addition to feeling closer to the way the DOM tree works, the event components are more powerful and less confusing:

* Event components are a visible part of the rendered tree rather than being an invisible wrapper.
* You can wrap event components around only one subtree rendered in a component, whereas the hooks would affect the entire component tree.
* Since they are visibly part of the tree, event components might prevent bugs. For example, if you added a new wrapper to an existing component with a new subtree that shouldn't receive events, the event component would still only wrap the part that should, whereas the hook would now wrap the new part as well. With hooks, you'd need to factor this into two components.

Rewriting @TrySound's example, it could look like this:

```javascript
const Menu = ({ closeMenu, doAction1, doAction2 }) => {
  return <Press onPress={closeMenu}>
    <Press onPress={doAction1}>
      <button>do action 1</button>
     </Press>
    <Press onPress={doAction2}>
      <button>do action 2</button>
    </Press>
  </Press>
}
```

Assuming the events passed to those handlers had a normal `stopPropagation` method, they could easily have all of the power I described earlier without a lot of the confusion.

I'd be interested to hear more of the reasoning around this, and why simply adding event propagation support to event components was not the way forward. What does this enable that simply using components does not?

sebmarkbage:
There are a bunch of reasons why these primitives are worth while. Exploring.

While the implementation uses bubbling you can also implement this using context by wrapping a context at each level. Eg think passing a dispatcher down. Another way to think about states that this could implement is reading a path like if this component is on a path towards the hovered target node. That’s observing a state similarly to useContext.

What hooks doesn’t do is add/change behavior of the tree and this doesn’t neither. It just observes it. That’s also an important distinction from how normal bubbling works. It actually changes the behavior of touch events and such.

There are a number of reasons we want this for complex cases but to me, the most compelling in the simplest ones.

Doing a `:hover` or `:focus-children` pseudo-selector is really simple in CSS.

Doing the same thing with rich components is a pita.

This primitive let us implement simple effects with really small syntax.

```
function ButtonWithHoverState() {
let isHovered = useHover();
return <CustomButton>
  <Label />
  {isHovered ? <Overlay /> : null}
</CustomButton>;
}
```

This syntax is already a bit much. Any solution adding refs or explicit wiring or callbacks to this is way too much.

sebmarkbage:
Another thing that stopPropagation have to deal with is what to do for server rendering. Especially for partial hydration where parents might be hydrated but not all children. The programmatic api needs to become way more declarative anyway. So the existing DOM approach is likely not going to be sufficient regardless.

devongovett:
This doesn't seem too different to me, but it is much clearer about:

1. Which elements it applies to.
2. When it will cause a component to re-render (where the state lives).

```jsx
function ButtonWithHoverState() {
  let [isHovered, setHovered] = useState(false);
  return <Hover onHoverChange={setHovered}>
    <CustomButton>
      <Label />
      {isHovered ? <Overlay /> : null}
    </CustomButton>
  </Hover>;
}
```

Not sure what you meant about stopPropagation on the server. How would user events like press and hover fire during server rendering?

necolas:
These hooks are complementary to, and must be used with event components in the tree. Many common scenarios, like the ones we're supporting internally, do not need the hooks at all.

The issue with partial hydration of server rendered HTML is related to replaying events and the difficulties caused by imperative calls to preventDefault and stopPropagation in event handlers.

mindplay-dk:
Might be a silly question, but why is it an element? It doesn't emit an actual DOM element, it just attaches handlers to the child(ren), right?

trueadm:
@mindplay-dk 

> Might be a silly question, but why is it an element? It doesn't emit an actual DOM element, it just attaches handlers to the child(ren), right?

These are only React elements, they're not DOM nodes. In fact, that's what is great about this new event system – you can encapsulate some forms of events without needing to associate them to a container. This is because we don't use the DOM node tree to handle event propagation, but rather the internal React fiber tree. This also allows us to add custom events without needing to worry about conflicting with native events that exist already or might get added in the future.

@devongovett

I originally had the same reservations about this. After working with this for a while and integrating this event system into FB5, it occurred to me that it was used to how DOM events work today and this new event system is a very different system. You need to forget how events work today on the DOM and have an open mind with React Flare. Very much like when React came around, and people had reservations, this is the same sort of mind-set change.

Event hooks aren't really anything super special, they're actually very like events components. For example take this:

```jsx
function MyComponent() {
  useEvent(Press)
  return (
    <Press>
      <Press>
  )
}
```
When this component and it's sub-tree gets reconciled by React, it actually forms the following ordering for when a "press" event comes in:

<Press> -> <Press> -> Press Hook

The important difference is that event components like (<Press>), never propagate events to other event components of the same type. This is to ensure the encapsulation is kept localized within the event component. React Flare doesn't allow users to do `event.stopPropagation()`, it's all declaratively handled in the form of component composition. So what if someone wants to listen to a nested `<Press>`? There would be no way of doing it, unless we then invalidated our rule and made event components optionally propagate. You'll remember that this was once the case in development, but we ran into issues internally and soon realized it was a bad design choice.

We then took a step back and realized that we can have two forms of encapsulation, localized within event components, and localized within function component sub-trees. How do you apply encapsulation to a function component sub-tree? The obvious is to use hooks, as that would allow us to "hook" the fiber representing the function component to the event responder in React's event system. Now, you have a way of "listening" to events of a specific responder, constrained by a component sub-tree – something we found we needed in a bunch of places internally. It also allows people to create powerful abstractions in the form of custom hooks that not only leverage things like state, context and refs, but also events! Not only that, but this system should also work with React Native and server-side rendering hydration (event replaying before concurrent hydration has hydrated a sub-tree).

benjamingr:
Hey @trueadm nice reply. I am very worried moving further away from native events and how the DOM works because writing automation for React is already super hard as it is because of how different the event system is from native events.

The DOM already has two mechanisms to deal with the hover issue you mentioned that allow overriding the default (which is not stopping propagation):
 - Some events (like scroll) don't bubble. Some events contain bubbling and non-bubbling alternatives (like blur and focusout).
 - You can capture events (an onHoverCapture) and the component you wrap cannot catch them before you.

I like the idea and the exploration and I think the syntax is interesting but this looks like a shift further away from native events. This breaks two major things:

 - The Chrome devtools tooling doesn't work with React particularly well, for example the event listeners tab shows the event delegation and EventPlugin arch. Moreoever it shows it for the "wrong" events (like how mouseenter listens to mouseover and mouseout).
 - Accessibility tools for people who have a hard time using a computer the "traditional" way have to pick up on this. 

Not to mention the fact react renders asynchronously and may handle events asynchronously means that establishing any causality for why something happened for debugging is significantly harder. 

I am biased because I have more experience building platforms and tools than apps and I am just one person and my opinion isn't any more valid than others' - I am just concerned about these issues.

gaearon:
>Accessibility tools for people who have a hard time using a computer the "traditional" way have to pick up on this.

Note that the intent is quite the opposite — all of these APIs are designed with extra effort to make accessible experiences the default.

These are valid concerns. At this point it's an experimental, completely additive (and unstable) API and React isn't "moving away" from anything. We can have a more informed broader discussion with more context and explanations after we can experiment with this in real products. For now it's a bit difficult to give all context (this is just in a very raw experimental stage), and I'm worried that you might be implying the communication on this PR is all we'll do, or that things are somehow set in stone. That's not the case.

benjamingr:
@gaearon oh, I just expressed worry regarding this API and how it interacts with automation and accessibility tools.

I am not saying "don't do it". This is more of a "this is a cool idea and sounds like a fun experiment, here are my concerns".

I am sorry it came as if I am opposed to this, I will communicate more clearly in the future in these sort of issues when I engage. I would never intentionally oppose to a feature without even trying it first and playing with the code (that would be mean and unproductive of me IMO) .

Anyway as always thanks for clarifying things and taking the time to emphasise this is still not set in stone. I am all for experimentation and doing cool things. I hope we can discuss the concerns I've raised (here or elsewhere) in a constructive way that does not attack or harm the experimentation  work and effort.

trueadm:
@benjamingr You made some good points, I'll try and address them accordingly:

- It's true that debugging with Chrome inspector is harder, but to be honest it's a quite terrible. That's why we have React Dev Tools. There's a new Dev Tools coming in the future that has full support for events that is going to be really appealing. We also plan on surfacing some additional features that will make event debugging much easier than before.
- We're planning on building a bespoke testing library that works great with this system. Rather than writing tests where you imperatively call `dispatchEvent` over and over, you'll be able to declaratively define an "interaction". This will then apply the w3c spec for what events would apply (mouse, touch, pen etc built in), so you get a really good testing interface. We need this internally anyway, as the `dispatchEvent` approach with JSDOM just doesn't work right.
- The new event responders have full control on how events are dispatched in terms of batching and compatibility with concurrent rendering. They also have full control over browser passive events. This isn't available in current event system.
- Accessibility is baked into this project. Everything we're doing has a strong emphasis on ensuring screen readers and interaction tooling (Voice Over on iOS for example) is taken into consideration both now and later on in the project.

As @gaearon said, nothing is set in stone. We may change how these features work at anytime, based off internal findings. That doesn't mean we can't discuss pain points now though – it's great to have community feedback on early ideas!

benjamingr:
@trueadm thanks for the thoughtful reply, it is good to learn that you are taking automation and accessibility seriously and are testing them.

>  There's a new Dev Tools coming in the future that has full support for events that is going to be really appealing.

Would it be possible to integrate into them in production builds? As far as I know at the moment React and React DevTools find each-other by a global variable which might not exist and usually doesn't in production.

A lot of time issues reproduce only in certain environment or builds and people resort into using the native devtools for debugging. I agree that the React DevTools provide a lot of information missing from the Chrome DevTools but moving even further away from native events means that the Chrome tools (that are always available) become less useful. 

I think this can be resolved technologically by making React easier to find though and is not a strong argument against diverging further from native events.

> We're planning on building a bespoke testing library that works great with this system. 

Yes please! 🙇  I would prefer it if this library was as low-level as possible where you provide basic primitives (find this component from an html element, fire a react event on it). This would absolutely alleviate my concern if it worked and kept some sort of ABI between React versions.

A lot of us are doing grey and black box testing which would need access to this library. I think that if the person writing the test and the person writing the app are different people then a different event system makes things harder. Tools with money (like Testim - my employer) have no issue dealing with this but tools that are much smaller and not backed by VC funds have a hard time with these issues.

Just try writing a cross-browser black-box test for Draft.js :] 

> The new event responders have full control on how events are dispatched in terms of batching and compatibility with concurrent rendering. They also have full control over browser passive events. This isn't available in current event system.

That's a good point. I am excited by the premise of React being able to make event handlers passive automatically which could help with a bunch of common footguns.

> Accessibility is baked into this project

I will try checking this out and ask a few friends who use assistive technologies to use the web to give feedback and I'll report if they say anything interesting.

Thanks for caring about this. I would really appreciate a way (for accessibility tooling) to figure out what caused what change (like synchronous native code does with a MutationObserver) to establish causality across usage for example:

 - "Would you like to discard the current changes?"

vs. 
 - "Modal, 'Would you like to discard the current changes?', triggered by clicking on Button 'Reset'"

It is an example of something that is possible with regular web technologies but is harder with React. This is not something Flare needs to fix and I do not believe Flare makes this harder to do but it would be great if we had accessibility tooling baked into the new devtools you are working on.

(This is just my opinion which is biased having worked with people who use assistive technologies a lot, I am not an expert in accessibility and it is possible this sort of feature isn't needed by most of the user base and does not need to be in 'core' react).

gaearon:
>Would it be possible to integrate into them in production builds? As far as I know at the moment React and React DevTools find each-other by a global variable which might not exist and usually doesn't in production.

This doesn't sound accurate. React DevTools works in production, and ReactDOM connects to it regardless.

benjamingr:
> This doesn't sound accurate. React DevTools works in production, and ReactDOM connects to it regardless.

I thought that React devtools just detects  `__REACT_DEVTOOLS_GLOBAL_HOOK__` or `window.React` and polls for `__REACT_DEVTOOLS_GLOBAL_HOOK__` once per second. I edited my comment.

On closer inspection I see that the hook itself emits the events rather than just checking for its presence. It would be useful to be able to consume this functionality programatically as well in order to build tools on top of it (an event giving you a reference to event tooling) the way Chrome extensions can connect to the devtools protocol.

In any case I feel this is a tangent and it distracts from the discussion about flare so I will hide this comment shortly so discussion can focus on the new event system and I have striked-through my earlier comment.

trueadm:
@benjamingr 

> I think this can be resolved technologically by making React easier to find though and is not a strong argument against diverging further from native events.

This new event system is a big departure from thinking of things in terms of "native events". Instead of thinking about handling events as "let me known when someone does onMouseDown", it's about the interaction and intent.

Event components like `<Press>` listen to a dozen native events to provide a single `onPress` consistently, there's no way it would be possible to provide the power of React Flare if we just passed native events to users. We're mostly leveraging Pointer Events too, so we can handle multi-touch across multiple input devices. It's awesome being able to have native-like feedback and trusting `<Press>` to work the same regardless of platform or input device (touch, pen, trackpad, keyboard and mouse).

When dealing with focus and modals, you'd use `<FocusScope>` which can contain the focus to its children and handle controlling and managing how focus works. When dealing with forms and inputs, we'll probably introduce `<FormScope>` that can really impact how form handling works today. These are all made possible by moving more into higher-level events (or interactions) than moving into lower-level events.

Lastly, these event modules are in user-land, thus you only pay the cost for their implementation if you actually use them in your bundle. They can also be bundle split, to reduce impact on initial render. This would not have been possible had we moved more towards a low-level architecture – you'd end up repeating the same thing we're attempting to do in your components (as everyone seems to do) without catching all the edge-cases and caveats that we've spend months finding and fixing. The aim is to cover all your bases so you can build better UIs without React and the DOM getting in your way.

devongovett:
@trueadm FWIW, I am totally in favor of the new event system. I love the work you've done on the event components so far, and I think they are going to improve the overall UX of all react apps in a major way. ❤️ 

I am still slightly concerned about the hooks API specifically though, and also want to ensure that usecases previously covered by the ability to stop propagation are properly covered by this new event model. I think many of my points about the clarity of hooks as opposed to components in my original comment still stand, in terms of API ergonomics. Would love to hear your thoughts on them.

In terms of propagation, I think there are some use cases that might be challenging to implement without some kind of control over propagation. I am totally aware that the current imperative methods have their challenges, so by no means do I think that is the right API necessarily. I do think the power they allow is necessary to expose though.

Imagine a confirmation dialog with a button, that when clicked, performs some validation. If the validation is successful, then the dialog closes, otherwise it remains open and displays an error. Assume for the sake of argument that this validation is too expensive to perform while the user is interacting with the content of the dialog. Originally, this dialog was implemented without validation by having `<Press>` responders wrapping the cancel and confirm buttons, with an event hook to handle closing the dialog on both actions at the dialog component level.

When validation is added which could prevent the dialog from closing, we would need some way to prevent the event from propagating to the hook press handler. With no way to do that, we would now have to refactor the dialog to not use the hook form at all, and instead pass a prop down to each button, and manually chain the close handler with the individual cancel and confirmation handlers. If this dialog component were a generic component implemented by a third party for example, we would have no way to do this and would also need to reimplement the entire dialog component in order to make this happen.

Another issue would arise if you added a new interactive element to the dialog later on, which also happened to use a `<Press>` event responder internally. Suddenly, the dialog would start closing when you interacted with this element. You'd need to refactor it again in order to not use the hook form, and instead pass props down explicitly to the buttons.

Along these lines, I'm struggling to find an example where using the hook form would not cause problems further down the road, especially if you don't control all of the components in the tree. I think it might be safer to just manually pass props down to the event components further down the tree so that at some point in the middle you can "stop propagation" by not calling the parent's callback. This is obviously very tedious though, so I think it would be good to find an API to express this in a better way.

Finally, coupling the hooks to event components further down the tree could have extensibility problems. Imagine I'm using a third party component, and I want to make something happen when a user hovers over a particular child element. Since I cannot modify the code to insert a `<Hover>` event responder, this is impossible. With DOM event propagation, I could simply handle events on a parent and check the event's `target` property to determine what I need to do. With this new model, I'd need to contact the developers of whatever component I was trying to extend, or reimplement it myself.

trueadm:
@devongovett These are good points, but ultimately it comes down to use-cases. Using the hook event form isn't meant to replace event components, they're meant to complement one-another. In the use case you have above, you'd never use the `Press` hook to close the dialog box – in fact, I think people will rarely use the hooks in conditional cases like this (with `<Press>` anyway).

There are a few use-cases in general with events:

- This event needs to propagate for an action that doesn't disrupt my children. i.e. styling, logging, visual indications. This is a great use-case for hooks
- This event needs to propagate and has a disruptive action. i.e. close dialogs, route to a new page, cancel an action etc. This is a bad use-case for hooks, ideally these should be put into component context, state or props.

That's not to say there are cases in the future where we might expand on the second pointer. One area of research is by using another new component that React Flare introduces – Event Targets. These would allow you to accomplish the same things, sort of like a `useReducer` dispatch. Although in this example I've called them `TargetID` just as an example:

Note: this is hypothetical API, it doesn't exist and we may never implement it. Also `TargetID` components do not create any DOM nodes or anything, they're just indicators for event responders to feed back to the user (if they exist).

```jsx
import { TargetID } from 'react-events';

function MyDialog({ closeDialog }) {
  useEvent(Press, { onPress: handlePress });

  function handlePress(event, targetId) {
    if (targetId === 'close') {
      closeDialog();
    }
  }

  return (
    <DialogInner />
  );
}

function DialogInner() {
  return (
    <div>
      <Press onPress={() => ...}>
        <button>
          Save
          <TargetID value="open" />
        </button>
      </Press>
      <Press>
        <button>
          Close
          <TargetID value="close" />
        </button>
      </Press>
    </div>
  );
}
```

You could expand this to give you identifiers for forms etc, or to handle other complex state machines. As `EventID` is a shareable object, they don't need to be passed around as props. The contract is that they use IDs that are documented and don't change, but that might be too fragile too. It would be good to get some more ideas around this. @sebmarkbage and @necolas had some around this area.

trueadm:
I was speaking to @sebmarkbage about this again recently and I think I have an interesting extension to this PR, which kind of expand on my last post, but also addresses some of the issues with using the event hook in this PR. This concept relies on providing an `name` prop to event components. The `name` prop should be a string or number. This concept already exists today in the DOM, but is rarely used. Here's an example of how it might be used:

```jsx
function MyComponent({ cancel }) {
  useEvent(Press, { onPress: handlePress });
  function handlePress(event) {
    // event.path is an array
    const pathStr = event.path.join('-');

    if (pathStr === 'outer-close') {
      cancel();
    } else {
      confirm();
    }
  }

  return (
    <Press name="outer">
      <Press name="cancel">
        <button>Cancel</button>
      </Press>
      <Press name="confirm">
        <button>Confirm</button>
      </Press>
    </Press>
  )
}
```

Thoughts?

packages/react-reconciler/src/ReactFiber.js
```diff
@@ -163,6 +164,9 @@ export type Fiber = {|
   // A linked-list of contexts that this fiber depends on
   contextDependencies: ContextDependencyList | null,
 
+  // An array of event component instances for this fiber
+  events: Array<ReactEventComponentInstance> | null,
```
acdlite: Could you put this array on the `ContextDependencyList` object instead? That way we can avoid another `fiber` field.

And then we can rename `contextDependencies` to just `dependencies`, perhaps.

trueadm: Isn't `ContextDependencyList` a linked list though. I opted for an array here, which would be a type-mismatch. Also, what if a component uses both context and events?

acdlite: There's a wrapper list object where you could attach the array. So they wouldn't be part of the same list:

```js
type Dependencies = {
  firstContext: ContextDependency,
  events: Array<ReactEventComponentInstance>
}
```

packages/react-reconciler/src/ReactFiber.js
```diff
@@ -267,8 +270,11 @@ function FiberNode(
   this.memoizedProps = null;
   this.updateQueue = null;
   this.memoizedState = null;
-  this.contextDependencies = null;
-  this.events = null;
+  this.dependencies = {
```
acdlite: Let's create this object lazily? Most fibers will not need it.

acdlite:
I'll follow up on creating the dependency object lazily. Merging to unblock.

PR #15928, opened 2019-06-19
[Fresh] Track mounted roots via DevTools Hook

gaearon:
The previous Fresh runtime API forced you to pass a root handle. However, that presumes that the module runtime (such as Metro) somehow is aware of all roots on the page. This isn't the case.

We could have React keep track of mounted roots. However, that might not be necessary. Here's why.

We still need some way to let the module runtime schedule refreshes. Currently this is done via DevTools Global Hook injection. DevTools (or module runtime) can set up the Global Hook. The renderer injects `scheduleHotUpdate` in it. As a result, the module runtime can schedule refreshes.

**Since we already rely on the DevTools Hook for communication, we might as well use the fact that this Hook gives us information about all roots.** Indeed, that's how DevTools itself works.

## The Goal

This PR lets the module runtime do 

```js
ReactRefreshRuntime.injectIntoGlobalHook(window);
```

and then

```js
ReactRefreshRuntime.performReactRefresh(update);
```

without worrying about how to track mounted roots.

#### Adding `injectIntoGlobalHook` to Refresh Runtime

  - This either sets up a shimmed tiny version of the Global Hook (for web users without extension).
  - Or, if the Global Hook is already set up (for web users with extension, or for RN which always sets it up), monkeypatches it.

The goal in both cases is to grab the scheduling functionality and to track roots in the Runtime. This lets the module system (e.g. Metro) simply call `ReactRefreshRuntime.performReactRefresh()` and not worry about how this connects to React.

The only constraint is that `ReactRefreshRuntime.injectIntoGlobalHook()` needs to run before the renderer code. The module system can guarantee it because it owns the execution order.

I have added regression tests for tracking roots.

#### Tweaking the API exposed to DevTools

Previously, we had `scheduleHotUpdate(root, update)` which included the resolution function. However, I realized this doesn't really work if you don't have a `root`. For example, if you hot update a component, but no roots are mounted yet, you still want its *next* render to use the updated version. So  you _need_ to inject the resolution function anyway.

I changed the API to:

* `setRefreshHandler(handler)` — This sets up the family resolution function. It is called on first edit even if there are no roots.
* `scheduleRefresh(root, update)` — renamed from `scheduleHotReload`
* `findHostInstancesForRefresh` — renamed from `findHostInstancesForHotReload`

## Follow-ups

There's a few things I plan to change here later. One is to use host config for getting client rect. But this isn't critical to get in now, since I don't plan to implement this in the first Metro integration. The second one is to add a few convenience helpers to move more React-related logic out of the module runtime (e.g. Metro). I might push that a bit later but it doesn't block this PR.

### DIFF
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -137,119 +137,131 @@ function resolveFamily(type) {
 }
 
 export function performReactRefresh(): boolean {
-  if (pendingUpdates.length === 0) {
-    return false;
-  }
-
-  const staleFamilies = new Set();
-  const updatedFamilies = new Set();
-
-  const updates = pendingUpdates;
-  pendingUpdates = [];
-  updates.forEach(([family, nextType]) => {
-    // Now that we got a real edit, we can create associations
-    // that will be read by the React reconciler.
-    const prevType = family.current;
-    familiesByType.set(prevType, family);
-    familiesByType.set(nextType, family);
-    family.current = nextType;
-
-    // Determine whether this should be a re-render or a re-mount.
-    if (canPreserveStateBetween(prevType, nextType)) {
-      updatedFamilies.add(family);
-    } else {
-      staleFamilies.add(family);
+  if (__DEV__) {
+    if (pendingUpdates.length === 0) {
+      return false;
     }
-  });
 
-  const update: RefreshUpdate = {
-    updatedFamilies,
-    staleFamilies,
-  };
+    const staleFamilies = new Set();
+    const updatedFamilies = new Set();
 
-  if (typeof setRefreshHandler !== 'function') {
-    warningWithoutStack(
-      false,
-      'Could not find the setRefreshHandler() implementation. ' +
-        'This likely means that injectIntoGlobalHook() was either ' +
-        'called before the global DevTools hook was set up, or after the ' +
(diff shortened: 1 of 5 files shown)


gaearon:
Pushed a second commit. This just adds some helpers that are currently inlined in Metro in my draft. It doesn't make sense to inline them in Metro because they're more React-specific. And other integrations will need the same helpers. So I put them in the runtime too.

gaearon:
Verified this works in RN.

packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -202,23 +271,260 @@ export function setSignature(
...
+
+    // For React Web, the global hook will be set up by the extension.
+    // This will also run before us.
+    let hook = globalObject.__REACT_DEVTOOLS_GLOBAL_HOOK__;
+    if (hook === undefined) {
+      // However, if there is no DevTools extension, we'll need to set up the global hook ourselves.
+      // Note that in this case it's important that renderer code runs *after* this method call.
+      // Otherwise, the renderer will think that there is no global hook, and won't do the injection.
+      globalObject.__REACT_DEVTOOLS_GLOBAL_HOOK__ = hook = {
+        supportsFiber: true,
```
bvaughn: Why? Does refresh actually work with non-FIber versions?

bvaughn:
Changes seem good. Your recent prod-only errors are causing test-prod tests to fail though 😁

PR #15931, opened 2019-06-20
[Flare] Press: fix stale deactivation region state

necolas:
The responder region calculation logic wasn't updating the deactivation region during the lifetime of an event instance, causing incorrect behaviour when the current press ends outside the press target and if the press target has moved since the last time the first-and-only time the deactivation region was measured.

Note: we previously discussed adding a warning when the press target has a height or width of `0`, however, that's only a special case for a general "issue" (unrelated to the issue this patch address). If the children of a press target are positioned (absolutely or relatively) outside the press target, and the target (and its children) move during a press (e.g., a "squish" transform) then the "pressend" trigger could be occurring on an element outside the responder's subtree. One that happens, we fallback to checking if the x/y event coordinates overlap with the union of activation/deactivation press target regions. Those regions do not account for protruding children and the "press" event might be skipped even if it appears to fall within the press retention region.

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -420,6 +420,8 @@ function dispatchPressEndEvents(
       deactivate(event, context, props, state);
     }
   }
+
+  state.responderRegionOnDeactivation = null;
 }
 
 function dispatchCancel(
@@ -726,6 +728,7 @@ const PressResponder = {
             state.pressTarget,
             props,
           );
+          state.responderRegionOnDeactivation = null;
           state.isPressWithinResponderRegion = true;
           dispatchPressStartEvents(event, context, props, state);
           addRootEventTypes(context, state);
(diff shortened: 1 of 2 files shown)


necolas:
This area of the code is quite complicated, and it's because the end events dispatcher is checking `state.isPressWithinResponderRegion` that I moved the updating call to update the stored box coordinates. LMK if you've got any ideas about how to refactor this code to simplify updating and querying this state. Eventually ResizeObserver might be a better way to ensure we can keep the measurements up to date independent from the event phases.

packages/react-events/src/Press.js
```diff
@@ -391,6 +393,22 @@ function dispatchPressEndEvents(
     state.longPressTimeout = null;
   }
 
+  if (nativeEvent != null && target != null) {
+    if (event != null && event.nativeEvent.type !== 'keyup') {
```
trueadm: Can't we use event.type here rather than event.nativeEvent.type?

trueadm:
The logic for this is quite confusing at first but I think I have my head around it now. Please can you please add two regression tests (`touch` and `pointer`) that fail on current master put passes for this PR? I'm worried that a subtle change might break this.

Issue #15930, opened 2019-06-19
[Flare] onPressEnd not being called on keyup for Enter key


aweary:
@trueadm @necolas 

Using a recent build from `master` I noticed that `onPressEnd` on the `Press` component is not being called on `keyup` when pressing with `Enter`. Here's an example that reproduces the issue: http://react-press-event-keyup-bug.surge.sh/. Here's the source for this example:

```jsx
function Application() {
  const [isPressed, setPressed] = React.useState(false);
  const onPressStart = () => {
    setPressed(true);
  };
  const onPressEnd = () => {
    setPressed(false);
  };
  return (
    <Press onPressStart={onPressStart} onPressEnd={onPressEnd}>
      <button style={{ 
        margin: 48, fontSize: 28, color: isPressed ? "red" : "initial" }}>
        Press Me
      </button>
    </Press>
  );
}
```

**Steps to Reproduce**
1. Focus the button via Tab
2. Press the button by holding down Enter
3. Un-press the button by releasing Enter

`onPressEnd` is never called so the pressed state persists. If you follow the same steps using the Space key or the mouse you'll see it works as expected. The [unit test for this](https://github.com/facebook/react/blob/f4e1ac8cafb75dfab129b95f0eb39591579f3d03/packages/react-events/src/__tests__/Press-test.internal.js#L336) is passing for me with this build which is kinda weird.

aweary:
After some debugging I believe this is due to how `click` events are dispatched when using the Space versus Enter on button elements. With Space the `click` event doesn't get dispatched until `keyup`, while for Enter the `click` event is dispatched on `keydown`. 

This means that `removeRootEventTypes` is called immediately for Enter so the `keyup` event is never handled. This also ends up breaking future presses as the pressed state never gets reset.

aweary:
@trueadm looks like https://github.com/facebook/react/pull/15854 is the root cause, moving `removeRootEventTypes` to the `click` event doesn't work here since `click` comes before `keyup`

necolas:
Fixed by the above PR

PR #15938, opened 2019-06-20
[Flare] Fix keyboard keyup regression

trueadm:
This PR addresses a keyboard bug that was a recent regression found in https://github.com/facebook/react/issues/15930. The problem was that there was a false assumption made that `click` occurred after `keyup`, it's actually the other way around. I've also updated the unit test that tracks this logic and ensured it matched the browser behaviour correctly. @aweary This should work for you now, let me know if you have any other issues!

### DIFF
--- a/packages/react-events/src/Press.js
+++ b/packages/react-events/src/Press.js
@@ -871,6 +871,7 @@ const PressResponder = {
               return;
             }
             isKeyboardEvent = true;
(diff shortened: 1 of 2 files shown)


trueadm:
@aweary I'll try and think of a unit test for that in a follow up. Thanks!

PR #15939, opened 2019-06-20
[Fresh] Throw in prod and change annotation

gaearon:
1. Babel transform now refuses to run in any environment other than `development`. This is to detect misconfiguration early.
2. The runtime now throws at initialization if it's included in a production mode bundle. This should never happen and usually points at misconfiguration. I'll verify this doesn't happen in RN.
3. Changes `@hot reset` annotation to `@refresh reset` to match overall naming.

### DIFF
--- a/packages/react-refresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-refresh/src/ReactFreshBabelPlugin.js
@@ -8,6 +8,19 @@
 'use strict';
 
 export default function(babel) {
+  if (typeof babel.getEnv === 'function') {
+    // Only available in Babel 7.
+    const env = babel.getEnv();
+    if (env !== 'development') {
+      throw new Error(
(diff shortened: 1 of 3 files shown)


packages/react-refresh/src/ReactFreshBabelPlugin.js
```diff
@@ -8,6 +8,19 @@
...
 
 export default function(babel) {
+  if (typeof babel.getEnv === 'function') {
+    // Only available in Babel 7.
+    const env = babel.getEnv();
+    if (env !== 'development') {
+      throw new Error(
+        'React Refresh Babel transform should only be enabled in development environment. ' +
+          'Instead, the environment is: "' +
+          env +
```
trueadm: I guess Prettier did this, seems strange though. :/

gaearon: yea

PR #15940, opened 2019-06-20
[Fresh] Don't consider require-like calls to be likely HOCs

gaearon:
I tried running Fresh transform on FB5 to verify it doesn't introduce semantic differences. There is a case where that happens due to inline requires.

We currently consider calls like `const Foo = bar()` to be likely component factory calls when `Foo` is used in JSX. E.g. it could be `const Foo = createFluxContainer()`. So we register those with Fresh runtime.

However, in CommonJS those might be *imports*. `const Foo = require('Foo')`. In that case registering them is a bad idea. Both because it's redundant (their own module will register them first anyway), and because this triggers inline requires early. We don't want to trigger inline requires early because that can introduce module cycles in our environment.

As a solution, I'm making the heuristic stricter. The dynamic `import()` or function that start with `require` are no longer considered likely component factories, and their result won't be registered even if it's used in JSX. We'll also completely bail out for more complex expressions (e.g. ternaries). We will, however, explicitly whitelist template literals because they're used for things like Styled Components, e.g.

```js
const Foo = styled.div`...`
```

See snapshots.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "babel-jest": "^23.0.1",
     "babel-plugin-check-es2015-constants": "^6.5.0",
     "babel-plugin-external-helpers": "^6.22.0",
+    "babel-plugin-syntax-dynamic-import": "^6.18.0",
     "babel-plugin-syntax-trailing-function-commas": "^6.5.0",
     "babel-plugin-transform-async-to-generator": "^6.22.0",
     "babel-plugin-transform-class-properties": "^6.11.5",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
     "babel-jest": "^23.0.1",
     "babel-plugin-check-es2015-constants": "^6.5.0",
     "babel-plugin-external-helpers": "^6.22.0",
+    "babel-plugin-syntax-dynamic-import": "^6.18.0",
```
gaearon: Had to add this so I can use it in a test.

PR #15941, opened 2019-06-20
[Fresh] Make global names more obscure

gaearon:
Just a find-and-replace. The goal is to more surely avoid clashes.


gaearon:
This is just a rename of a mangled global so I think I'll just merge. :-)

PR #15943, opened 2019-06-21
[Flare] Move createEvent back to React object

trueadm:
Internally, we ran into issues with exporting `createEvent` on `ReactDOM`. Moving this back to `React` fixes the issue. This is because the `ReactDOM` module has side-effects that occur upon initialization of the module, which breaks in cases such as server-side rendering.

Lately, given the `useEvent` hook, it might make more sense for us to keep `createEvent` on the `React` object. In the RN integration PR, I found better ways to ensure Flow type safety whilst doing so, so it might be worth while looking at how we can incorporate a polymorphic `createEvent` on the `React` object, where the `responder` field might relate to different responders depending on the platform.

The event responders exported should always be the same shape between platforms anyway, the only thing that differs is the props they receive – so this should be safe to do.

@sebmarkbage I'd really like it if you could chime in on this one too :)

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -42,7 +42,6 @@ import {
 } from 'react-reconciler/inline.dom';
 import {createPortal as createPortalImpl} from 'shared/ReactPortal';
 import {canUseDOM} from 'shared/ExecutionEnvironment';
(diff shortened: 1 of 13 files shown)


necolas:
LGTM. Don't think the types need changing back from ReactDOM to React as the ReactDOM types are only used in ReactDOM files. And it sounds like it's still something to figure out.

PR #15944, opened 2019-06-21
Lazily initialize dependencies object

acdlite:
Most fibers do not have events or context, so we save memory lazily initializing this container node.

Follow-up from #15927

### DIFF
--- a/packages/react-reconciler/src/ReactFiber.js
+++ b/packages/react-reconciler/src/ReactFiber.js
@@ -169,8 +169,8 @@ export type Fiber = {|
   // The state used to create the output
   memoizedState: any,
 
-  // An object of dependencies for this fiber
-  dependencies: Dependencies,
+  // Dependencies (contexts, events) for this fiber, if it has any
+  dependencies: Dependencies | null,
 
   // Bitfield that describes properties about the fiber and its subtree. E.g.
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiber.js
```diff
@@ -443,7 +439,18 @@ export function createWorkInProgress(
...
+  // it cannot be shared with the current fiber.
+  const currentDependencies = current.dependencies;
+  workInProgress.dependencies =
+    currentDependencies === null
+      ? null
+      : {
+          expirationTime: currentDependencies.expirationTime,
+          firstContext: currentDependencies.firstContext,
+          events: currentDependencies.events,
+        };
```
acdlite: The reason this only became an issue now is because of this change: https://github.com/facebook/react/pull/15944/files#diff-ff9ee977e81fe580b4202130fd86caeaR322

It used to reset `workInProgress.contextDependencies`, which is fine because `workInProgress` is a clone. But after #15944 it resets `workInProgress.dependencies.firstContext`, which is a bug because `dependencies is shared with the current fiber.

I clone in the render phase instead, but since this is such a subtle thing to keep track of (I've hit a similar thing with the class update queue before, but even so it took me a while to figure this bug out), I opted to clone here instead. Something to keep in mind when we look at refactoring the Fiber data structure.

PR #15946, opened 2019-06-21
Expire rendering the tail of SuspenseList after a timeout

sebmarkbage:
This is the first Suspense feature that isn't actually dependent on IO.

The thinking here is that it's normal for a SuspenseList to show loading states, and it'll be designed to handle it one at a time.

However, sometimes there are lists with really big items that take a long time to CPU render. Since data can become available as we do that, it is likely that we have all the data and become CPU bound.

In that case, the list would naively just render until the end and then display all items at once. I think that's actually what you want for fast lists. However, for slow ones (like News Feed), you're better off showing a few rows at a time.

It's not necessarily one at a time because if you can do many in a short period of time and fit them all on the screen, then it's better to do them all at once than pop them in one at a time very quickly.

Therefore, I use a heuristic of trying to render as many rows as I can in 500ms before giving up. I originally thought of making it configurable or having a special revealOrder mode for this but I have sense that this is not something people should have to think about too much and there are other possible heuristics we can add here later (such as if it will be below the fold or not). It seems like this would mostly just carry over to most cases.

This timer starts before the first row of the tail and we only check it after. This ensures that we always make a little progress each attempt. An alternative approach could be to start the time before doing the head of the list but we don't want that being slow prevent us from making further progress.

Currently, I disable this optimization at Never priority because there's nothing intermediate that becomes visible anyway.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -113,7 +113,7 @@ import {
   enableEventAPI,
 } from 'shared/ReactFeatureFlags';
 import {
-  markDidDeprioritizeIdleSubtree,
+  markSpawnedWork,
   renderDidSuspend,
   renderDidSuspendDelayIfPossible,
 } from './ReactFiberWorkLoop';
@@ -911,7 +911,7 @@ function completeWork(
               'This is probably a bug in React.',
           );
           if (enableSchedulerTracing) {
-            markDidDeprioritizeIdleSubtree();
+            markSpawnedWork(Never);
           }
           skipPastDehydratedSuspenseInstance(workInProgress);
         } else if ((workInProgress.effectTag & DidCapture) === NoEffect) {
@@ -974,8 +974,12 @@ function completeWork(
               // them, then they really have the same priority as this render.
               // So we'll pick it back up the very next render pass once we've had
               // an opportunity to yield for paint.
-              workInProgress.expirationTime = workInProgress.childExpirationTime =
-                renderExpirationTime - 1;
+
+              const nextPriority = renderExpirationTime - 1;
+              workInProgress.expirationTime = workInProgress.childExpirationTime = nextPriority;
+              if (enableSchedulerTracing) {
+                markSpawnedWork(nextPriority);
+              }
             } else {
               suspenseListState.didSuspend = isShowingAnyFallbacks(rendered);
             }
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -24,6 +24,8 @@ import type {
 } from './ReactFiberSuspenseComponent';
 import type {SuspenseContext} from './ReactFiberSuspenseContext';
 
+import {now} from './SchedulerWithReactIntegration';
+
 import {
   IndeterminateComponent,
   FunctionComponent,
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -958,7 +961,24 @@ function completeWork(
...
+              // attempts to render the main content and only render fallbacks.
+              // The assumption is that this is usually faster.
+              suspenseListState.didSuspend = true;
+              // Since nothing actually suspended, there will nothing to ping this
+              // to get it started back up to attempt the next item. If we can show
+              // them, then they really have the same priority as this render.
+              // So we'll pick it back up the very next render pass once we've had
+              // an opportunity to yield for paint.
+              workInProgress.expirationTime = workInProgress.childExpirationTime =
+                renderExpirationTime - 1;
```
sebmarkbage: @bvaughn This leaves remaining work at the end similar to offscreen and deprioritization but it's not actually Never priority. Do I need to do something special to carry over interaction tracing in this case?

acdlite: Yeah see `markDidPrioritizeIdleTree`. Need to update that so it tracks the lowest remaining priority per render. And then call it here.

acdlite: Here’s the recent commit that implemented this: https://github.com/facebook/react/pull/15872/commits/603ef0e9cd2add2bd7ec7599edf9ace4045ed8ed

bvaughn: Thanks Andrew!

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
```diff
@@ -1027,4 +1027,66 @@ describe('ReactSuspenseList', () => {
...
+    // We've still not been able to show anything on the screen even though
+    // we have two items ready.
+    expect(ReactNoop).toMatchRenderedOutput(null);
+
+    // Time has now elapsed for so long that we're just going to give up
+    // rendering the rest of the content. So that we can at least show
+    // something.
+    expect(Scheduler).toFlushAndYieldThrough([
+      'Loading C',
+      'C', // I'll flush through into the next render so that the first commits.
```
bvaughn: This is a helpful inline comment. In general, I find over-flushing approaches like this confusing at first glance.

One thing that I think maybe reads a little clearer (though it's more verbose) is to yield in an effect:
https://github.com/facebook/react/blob/34ce57ae751e0952fd12ab532a3e5694445897ea/packages/react/src/__tests__/ReactDOMTracing-test.internal.js#L232-L236

Then flush through the effect, rather than the first part of the next render:
https://github.com/facebook/react/blob/34ce57ae751e0952fd12ab532a3e5694445897ea/packages/react/src/__tests__/ReactDOMTracing-test.internal.js#L265

Just sharing, obviously no big deal :D

packages/react-reconciler/src/ReactFiberCompleteWork.js
sebmarkbage: Yea I might need a hand with this. There is another variant of this that could happen with hydration where we want to do the inverse of this and schedule an update at a higher pri from a render and effectively restart this render.

bvaughn: Looks like you got it sorted out okay 👍 but let me know if you do need a hand with the follow up or something.

PR #15953, opened 2019-06-21
[Flare] Add Hooks to event modules

aweary:
Initial implementation for the various Hooks that will go along with the event components in `react-events`. I'll push another commit with some tests later today.

### DIFF
--- a/packages/react-events/drag.js
+++ b/packages/react-events/drag.js
@@ -9,6 +9,4 @@
 
 'use strict';
 
-const Drag = require('./src/dom/Drag');
-
-module.exports = Drag.default || Drag;
+module.exports = require('./src/dom/Drag');
--- a/packages/react-events/focus.js
+++ b/packages/react-events/focus.js
@@ -9,6 +9,4 @@
 
 'use strict';
 
-const Focus = require('./src/dom/Focus');
-
-module.exports = Focus.default || Focus;
+module.exports = require('./src/dom/Focus');
--- a/packages/react-events/hover.js
+++ b/packages/react-events/hover.js
@@ -9,6 +9,4 @@
 
 'use strict';
 
-const Hover = require('./src/dom/Hover');
-
-module.exports = Hover.default || Hover;
+module.exports = require('./src/dom/Hover');
--- a/packages/react-events/press.js
+++ b/packages/react-events/press.js
@@ -9,6 +9,4 @@
 
 'use strict';
 
-const Press = require('./src/dom/Press');
-
-module.exports = Press.default || Press;
+module.exports = require('./src/dom/Press');
--- a/packages/react-events/scroll.js
+++ b/packages/react-events/scroll.js
@@ -9,6 +9,4 @@
 
 'use strict';
 
-const Scroll = require('./src/dom/Scroll');
-
-module.exports = Scroll.default || Scroll;
+module.exports = require('./src/dom/Scroll');
--- a/packages/react-events/src/dom/Drag.js
+++ b/packages/react-events/src/dom/Drag.js
@@ -262,8 +262,8 @@ const DragResponder: ReactDOMEventResponder = {
   },
 };
 
-export function useDrag(props: Object): void {
-  React.unstable_useEvent(DragResponder, props);
-}
-
 export const Drag = React.unstable_createEvent(DragResponder);
+
+export function useDrag(props: Object): void {
+  React.unstable_useEvent(Drag, props);
+}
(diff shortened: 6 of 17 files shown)


trueadm:
This changes the modules from default exports, which will mean we need to change how the internal www files work to use these.

packages/react-events/src/Drag.js
```diff
@@ -262,4 +262,8 @@ const DragResponder: ReactDOMEventResponder = {
   },
 };
 
-export default React.unstable_createEvent(DragResponder);
+export function useDrag(props: Object) : void {
+  React.unstable_useEvent(props);
```
TrySound: DragResponder?

necolas:
Is it necessary to export these hooks from the responder modules?

TrySound:
I personally would prefer to use `useEvent(EventComponent)` directly without wrappers. This show explicit intention to get bubbled events.

aweary:
@necolas this was the approach @trueadm and I discussed, so you can do:
```js
import {useFocus} from 'react-events/focus'
```

@necolas do you have another approach in mind? Happy to take a different path.

> I personally would prefer to use useEvent(EventComponent) directly without wrappers. This show explicit intention to get bubbled events.

@TrySound you pass the responder to `useEvent`, not the event component. So it would be `useEvent(EventComponent.responder)` which is weird. `useEvent` will likely be used by responder authors to provide their own Hook API, rather than used directly by product code.

necolas:
If the `useEvent` hook doesn't take the result of `createEvent`, the API names probably aren't right.

trueadm:
@necolas We could change it to take event components instead. I left that an open ended question in the original PR and was waiting till we played more around with how we expect to consume these. Whilst we do that, we should expose a way of telling the event responder if it's being used with a hook or an event component - like I mentioned on our internal issue.

aweary:
If there's a use case for using `useEvent` with a responder that doesn't have an associated `EventComponent` then it makes sense for it to accept the responder directly, but I don't think there is? Reading `responder` off of the `EventComponent` makes the most sense to me.

trueadm:
Addressed these points in https://github.com/facebook/react/pull/15955.

packages/react-events/src/Press.js
```diff
@@ -982,4 +982,8 @@ const PressResponder: ReactDOMEventResponder = {
   },
 };
 
-export default React.unstable_createEvent(PressResponder);
+export const Press = React.unstable_createEvent(PressResponder);
+
+export function usePress(props: PressProps): void {
+  React.unstable_useEvent(Press, props);
```
trueadm: I think @sebmarkbage and @acdlite were thinking that if you didn't pass `props`, that the default would be:

```jsx
const [isPressed, updatePressed] = useState(false);
React.unstable_useEvent(Press, { onPressChange: updatePressed });
return isPressed;
```

Same thing for `focus` and `hover`.

packages/react-events/drag.js
```diff
@@ -9,6 +9,4 @@
 
 'use strict';
 
-const Drag = require('./src/Drag');
-
-module.exports = Drag.default || Drag;
+module.exports = require('./src/Drag');
```
TrySound: I would prefer to not use such bridges modules. I use rollup for building production app. Such modules require explicitly define `namedExports: { 'react-events/drag': ['Drag', 'useDrag'] }`.

packages/react-events/src/Press.js
necolas: Magic vs less magic. We'd have to have a return object like PressableState built into the responder module.

aweary: Having `usePress` and other event hooks be conditionally stateful seems confusing and counter to the general rule of avoiding conditional state in hooks. 

It also means that if I wanted both the `isPressed` state while also observing other behavior, then I'd have to redefine the `isPressed` myself. 

Maybe it should _always_ return the state, and users who don't care about it can just ignore the return value?

necolas: There's also more than one boolean state in most cases.

Press: pressed, longPressed
Focus: focused, focusVisible, focusWithin, focusVisibleWithin

packages/react-events/drag.js
necolas: We'll get to that if this ever becomes public, but this work is currently limited to experimentation at Facebook.

packages/react-events/src/Press.js
trueadm: Ah yeah, maybe it is `usePressed` then that they were speaking about.

PR #15959, opened 2019-06-22
[Scheduler] Yield less if there's no pending input

acdlite:
At the end of each frame, Scheduler yields control of the main thread so the browser can execute important tasks; most importantly, painting the screen and responding to user input. There's some overhead involved in regaining control of the main thread, so we'd like to yield as infrequently as possible to keep the UI responsive.

The reason we yield on every frame is because there's no way for us to know whether we're blocking user input.

[`isInputPending`](https://github.com/WICG/is-input-pending) is an experimental browser API that gives us this information. It tells us whether there's a pending user input, which also means it tells us if there's *not* a pending user input. We can use this signal to decide whether it's OK not to yield.

There's a max frame length after which we'll yield regardless, as a precaution against blocking non-input tasks that we don't know about.

### DIFF
--- a/packages/scheduler/src/forks/SchedulerHostConfig.default.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.default.js
@@ -160,8 +160,38 @@ if (
   let activeFrameTime = 33;
   let fpsLocked = false;
 
+  // TODO: Make this configurable
+  // TODO: Adjust this based on priority?
+  let maxFrameLength = 300;
+
+  const isInputPending =
+    navigator !== undefined &&
+    navigator.scheduling !== undefined &&
+    navigator.scheduling.isInputPending !== undefined
+      ? navigator.scheduling.isInputPending
+      : null;
+
   shouldYieldToHost = function() {
-    return frameDeadline <= getCurrentTime();
+    const currentTime = getCurrentTime();
+    if (currentTime < frameDeadline) {
+      // There's still time left in the frame.
+      return false;
+    } else {
+      // There's no time left in the frame. We may want to yield control of the
+      // main thread, so the browser can perform high priority tasks. The main
+      // ones are painting and user input. If we're certain there's no user
+      // input, then we can yield less often without making the app less
+      // responsive. We'll eventually yield regardless, since there could be
+      // other main thread tasks that we don't know about.
+      if (isInputPending !== null && !isInputPending()) {
+        // There's no pending input. Only yield if we've reached the max
+        // frame length.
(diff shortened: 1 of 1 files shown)


packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -160,8 +160,38 @@ if (
   let activeFrameTime = 33;
   let fpsLocked = false;
 
+  // TODO: Make this configurable
+  // TODO: Adjust this based on priority?
+  let maxFrameLength = 150;
```
sebmarkbage: Let’s start a bit higher to get good signal on this. 300? Enough time to get an I/O event in that restarts React.

sebmarkbage:
We might need to put this behind a flag. If we ship this in open source we have effectively sealed the fate for isInputPending characteristics and api. Might be good to actually have it shipping before we do that.

acdlite:
If I understand how the origin trial works, I don't think a flag is needed because the API is only available if you send down a special header, which @n8schloss put behind a GK. But I'll add one just in case since it's easy enough.

acdlite:
I don't think it matters at all in open source because you also have to be on the same origin.

acdlite:
Even still, the worst that would happen is the we hit the threshold and activate the kill switch:

> Usage of these experiments is constrained to remain below Chrome’s deprecation threshold (< 0.5% of all Chrome page loads) by a system which automatically disables the experiment on all origins if this threshold is exceeded.

https://github.com/GoogleChrome/OriginTrials/blob/gh-pages/explainer.md

acdlite:
I'll do the feature flag separately

n8schloss:
Btw, as implemented right now it'll check for *any* input event, but I'm not sure if that's what we want since it may check for events that don't warrant interrupting things. On Facebook when we ran the origin trial for this I used the list of events [here](https://github.com/facebook/react/blob/master/packages/react-dom/src/events/SimpleEventPlugin.js#L59-L136) with the addition of `mouseMove`.

acdlite:
@n8schloss Our theory is that it's fine yield to all input events since we're also waiting until the end of the frame, using the rAF loop.

Also, we recently landed a change where continuous input events (i.e. ones that aren't discrete) are batched but still scheduled at user-blocking priority. For example, rendering should yield to mouse enter/leave events so that hover states are responsive in FB5. Here's the list of the those events: https://github.com/facebook/react/blob/master/packages/react-dom/src/events/SimpleEventPlugin.js#L104-L119

In the worst case we'll never be yieldier than we are today.

PR #15960, opened 2019-06-22
[Scheduler] requestPaint

acdlite:
Signals to Scheduler that the browser needs to paint the screen. React will call it in the commit phase. Scheduler will yield at the end of the current frame, even if there is no pending input.

When `isInputPending` is not available, this has no effect, because we yield at the end of every frame regardless.

React will call `requestPaint` in the commit phase as long as there's at least one effect. We could choose not to call it if none of the effects are DOM mutations, but this is so rare that it doesn't seem worthwhile to bother checking.


packages/scheduler/src/forks/SchedulerHostConfig.mock.js
```diff
@@ -94,6 +98,31 @@ export function unstable_flushNumberOfYields(count: number): void {
   }
 }
 
+export function unstable_flushUntilNextPaint(): void {
+  if (isFlushing) {
+    throw new Error('Already flushing work.');
+  }
+  if (scheduledCallback !== null) {
+    const cb = scheduledCallback;
+    shouldYieldForPaint = true;
```
acdlite: I supposed a better solution is to reassign `shouldYieldToHost`

algebraic_effects.gif

packages/react-reconciler/src/SchedulerWithReactIntegration.js
```diff
@@ -63,6 +64,7 @@ export const IdlePriority: ReactPriorityLevel = 95;
 export const NoPriority: ReactPriorityLevel = 90;
 
 export const shouldYield = Scheduler_shouldYield;
+export const requestPaint = Scheduler_requestPaint;
```
sebmarkbage: We probably need to feature test this so it doesn’t break in RN or too terribly for mismatched versions.

PR #15966, opened 2019-06-23
[Fresh] Retry failed roots on refresh

gaearon:
This aligns the behavior for roots with error boundaries. When a root unmounts as a result of an error, if the Fresh runtime is running, it will record its last rendered element in a Map.

There is a new `scheduleRoot(root, element)` DevTools hook. Fresh runtime calls it on update for all failed roots with their last rendered elements. This offers us an opportunity to remount the tree.

We remove a root from the failed map if there is a further update on it that isn't caused by an error. In order to determine whether an update is caused by an error, we check `effectTag`. This is done before calling `onCommitFiberRoot` so that Fresh runtime itself doesn't depend on a particular tag constant.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberDevToolsHook.js
+++ b/packages/react-reconciler/src/ReactFiberDevToolsHook.js
@@ -15,6 +15,7 @@ import type {Fiber} from './ReactFiber';
 import type {FiberRoot} from './ReactFiberRoot';
 import type {ExpirationTime} from './ReactFiberExpirationTime';
 
+import {DidCapture} from 'shared/ReactSideEffectTags';
 import warningWithoutStack from 'shared/warningWithoutStack';
 
 declare var __REACT_DEVTOOLS_GLOBAL_HOOK__: Object | void;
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberDevToolsHook.js
```diff
@@ -55,15 +56,16 @@ export function injectInternals(internals: Object): boolean {
...
           const currentTime = requestCurrentTime();
           const priorityLevel = inferPriorityFromExpirationTime(
             currentTime,
             expirationTime,
           );
-          hook.onCommitFiberRoot(rendererID, root, priorityLevel);
+          hook.onCommitFiberRoot(rendererID, root, priorityLevel, didError);
         } else {
-          hook.onCommitFiberRoot(rendererID, root);
+          hook.onCommitFiberRoot(rendererID, root, undefined, didError);
```
bvaughn: Seems like we should pass `priorityLevel` and `didError` in both cases?

bvaughn: Ignore me.

gaearon: (Resolved in a chat)

PR #15967, opened 2019-06-23
feat(chore): Minor refactor

jamesgeorge007:
Replaced the `if-else` ladder with ternary operator thereby resulting in a more concise form.

### DIFF
--- a/packages/react/npm/index.js
+++ b/packages/react/npm/index.js
@@ -1,7 +1,4 @@
 'use strict';
 
-if (process.env.NODE_ENV === 'production') {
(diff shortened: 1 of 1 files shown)


jamesgeorge007:
cc @gaearon

gaearon:
Hi, we generally avoid stylistic changes like this. It also won't work because requires need to be static.

gaearon:
(Thanks for PR though!)

jamesgeorge007:
@gaearon It would be great if you could point me to a perfect one to go for.

PR #15974, opened 2019-06-24
Add explicit support for ESLint 6.0.0

### LINKED ISSUE #15971: [eslint-plugin-react-hooks] Compatibility with ESLint 6.0.0
feature

Using in my `package.json`:
```
    "eslint": "^6.0.0",
    "eslint-plugin-react-hooks": "^1.6.0",
```
Yarn emits:
```
warning " > eslint-plugin-react-hooks@1.6.0" has incorrect peer dependency "eslint@^3.0.0 || ^4.0.0 || ^5.0.0".
```

No warning.

AFAICT, there shouldn't be other change needed besides updating the peer dependency version in the `package.json`: https://eslint.org/docs/user-guide/migrating-to-6.0.0
so it looks like I could just ignore the warning, but I'd rather be sure the plugin is actually compatible with ESLint 6.0.0

n/a

tbroyer:
Preemptively update tests wrt 'parser' requiring an absolute path rather than a package name, even though the project is still using ESLint 4.

Tested locally by upgrading eslint (and all `eslint-*` dependencies) to their latest versions (using `yarn upgrade-interactive --latest`) and then running `yarn test`. Without the change to the tests, they expectedly failed due to the `parser` option not being an absolute path; with the change to the tests, they all pass.

Fixes #15971

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -12,7 +12,7 @@ const ReactHooksESLintPlugin = require('eslint-plugin-react-hooks');
 const ReactHooksESLintRule = ReactHooksESLintPlugin.rules['exhaustive-deps'];
 
 ESLintTester.setDefaultConfig({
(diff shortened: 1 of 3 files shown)


gaearon:
Thanks.

PR #15981, opened 2019-06-24
[Flare] Move all event responders to dom directory

trueadm:
All the current event responders are implementations for the DOM renderer. As we plan to have other implementations for other renderers (RN etc), they should be in their own directories.


josiahKayan:
Em seg, 24 de jun de 2019 18:43, sizebot <notifications@github.com>
escreveu:

PR #15984, opened 2019-06-25
feat(docs): Add Table of Contents

jamesgeorge007:
Included contents section for easy navigation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,15 @@ React is a JavaScript library for building user interfaces.
 
 [Learn how to use React in your own project](https://reactjs.org/docs/getting-started.html).
 
+## Contents
+
+- [Installation](#installation)
+- [Documentation](#documentation)
+- [Examples](#examples)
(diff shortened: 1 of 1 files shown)


jamesgeorge007:
cc @aweary

jamesgeorge007:
cc @aweary @gaearon

jamesgeorge007:
cc @threepointone @trueadm

threepointone:
Thanks for this, but the README file is already very short so it can be quickly glanced, with links to reactjs.org for more. I don't think we should add a TOC.

jamesgeorge007:
@threepointone Just take the case of the [[react-native](https://github.com/facebook/react-native/blob/master/README.md) repo. It still has a TOC offering better navigation experience and I guess the same is applicable here as well.
Thoughts :thinking:

threepointone:
Closing since I expressed how I think about it, but happy to revisit if this comes up again.

PR #15987, opened 2019-06-25
RFC: wrap react-test-renderer's .create and .update with act() by default

threepointone:
DO NOT MERGE: Doing a PR so I can test on fb first. 

You should read this whitespace changes off. 

This PR adds act() by default to react-test-renderer's `.create`, `.update`, and `.unmount` methods. 

We risk a major breaking change with having added warnings on unqueued effects outside of an act scope. Eg: A lot of people's snapshot tests will break; this would be good, if they didn't have to first wrap their renderer helper with `act()`, and then update snapshots.

Broadly, we could also make a case why this makes sense to any consumers; they'll be using it only in testing scenarios, which we'd want to wrap in `act()` anyway

For our internal tests, this had the effect that a larger portion of the test was just wrapped with an outer `act()`, letting us test incremental updates by advancing the `Scheduler` manually.

- In the second commit, I also fix act()'s flow signature
- in the third commit, I update our fixtures to account for the change

### DIFF
--- a/fixtures/dom/src/index.test.js
+++ b/fixtures/dom/src/index.test.js
@@ -76,25 +76,23 @@ it('warns when using the wrong act version - test + dom: updates', () => {
   ]);
 });
 
-it('warns when using the wrong act version - dom + test: .create()', () => {
+it('does *not* warn when using the wrong act version - dom + test: .create()', () => {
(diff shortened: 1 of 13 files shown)


threepointone:
wiping this out, sending a fresh one

PR #15990, opened 2019-06-25
Refactor 'ReactChildren.js'

olafur-palsson:
Mainly split up 'traverseAllChildrenImpl' which had already grown to 100 lines.

### DIFF
--- a/packages/react/src/ReactChildren.js
+++ b/packages/react/src/ReactChildren.js
@@ -43,7 +43,6 @@ function escape(key) {
  * pattern.
  */
 
-
 const userProvidedKeyEscapeRegex = /\/+/g;
 function escapeUserProvidedKey(text) {
   return ('' + text).replace(userProvidedKeyEscapeRegex, '$&/');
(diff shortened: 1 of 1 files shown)


aweary:
Thanks, but we don't enforce a maximum line count for functions. We also tend to avoid purely stylistic refactors to avoid the risk of introduction unexpected behavior.

olafur-palsson:
What about increasing readability? The code improves a lot slower if it's hard to read.

audiolion:
I would argue that the original was was easier to read. For instance, `childrenAreNull` replaced this section of code:

```js
const type = typeof children;
 
if (type === 'undefined' || type === 'boolean') {	
  // All of the above are perceived as null.	
  children = null;
}
```

Because we are dealing with React domain specific terminology (e.g. children), when I see

```js
if (childrenAreNull(type))
```

I am left wondering what that function does and how it works, so I need to jump to that function definition. Whereas the first way, it is very explicit about exactly what is being checked, and the comment informs us of the semantic meaning of that check. Note too, the comment says `perceived as null`, whereas your function name loses that meaning in translation with the definitive `childrenAreNull`

PR #15998, opened 2019-06-26
Slightly improve performance of hydration.

bmeurer:
Avoid loading nodeType and data couple times from the same node in a row,
but instead load them only once, which will help engines to run this code
faster, especially during startup of the application. The general approach
is still not ideal, since hydrating this way forces the browser engine
to materialize JavaScript wrapper objects for all DOM nodes, even if they
are not interesting to hydration itself.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -677,7 +677,7 @@ function isHydratableNode(node: Node) {
   }
   if (enableSuspenseServerRenderer) {
     if (nodeType === COMMENT_NODE) {
-      return false;
+      return true;
     }
     const nodeData = (node: any).data;
     return (
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -670,21 +670,31 @@ export function registerSuspenseInstanceRetry(
   instance._reactRetry = callback;
 }
 
+function isHydratableNode(node: Node) {
+  const nodeType = node.nodeType;
+  if (nodeType === ELEMENT_NODE || nodeType === TEXT_NODE) {
+    return true;
+  }
+  if (enableSuspenseServerRenderer) {
+    if (nodeType === COMMENT_NODE) {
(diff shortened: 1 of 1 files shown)


sebmarkbage:
In normal usage, it is not expected that there will be any non-hydratable nodes. So these would all be materialized anyway. This is here mostly as a precaution against browser extensions or proxies inserting unexpected content.

bmeurer:
Thanks. I'm wondering if it's possible to do something about this in general. While it's probably possible to optimize this a bit more with the current approach, having to materialize all these wrapper objects is always gonna remain a bottleneck.

cc @developit @ timneutkens

sebmarkbage:
One thing we could do is drop all validation but still preresolve references to the node. Is there a way to walk the DOM while not materializing JS wrappers or cheaper ones?

We used to have a lazy approach but that instead paid the cost when the user interacted with something and also added complexity and meta-data that made it overall not worth it.

We’re moving to a selective hydration approach driven by interaction which effectively makes the paths lazy but also able to warm up in preparation for an interaction. So the characteristics are about to change.

bmeurer:
Interesting. The lazy approach sounds like a viable path.

Not validating, i.e. not loading `nodeType` and doing these lower case comparisons can probably reduce worst case scenarios - where we right not have easily 600ms for hydration - significantly.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -670,44 +670,40 @@ export function registerSuspenseInstanceRetry(
...
-  ) {
-    node = node.nextSibling;
+  for (; node != null; node = node.nextSibling) {
+    const nodeType = node.nodeType;
+    if (nodeType === ELEMENT_NODE || nodeType === TEXT_NODE) {
+      break;
+    }
+    if (enableSuspenseServerRenderer) {
+      if (nodeType === COMMENT_NODE) {
+        break;
```
sebmarkbage: This should actually have been `continue`. Logic is hard. :)

acdlite: Also, test coverage is hard :D

PR #16000, opened 2019-06-26
Add option to includeNegativeTabIndex in getFocusableElementsInScope

voideanvalue:
The idea for this pull request came up when @trueadm and I were talking about being able to differentiate between focusable and tabbable elements (to implement something like roving tabindex for focus management as recommended within aria best practices guide: https://www.w3.org/TR/wai-aria-practices/#kbd_roving_tabindex).

The current `getFocusableElementsInScope` implementation treats elements with `tabIndex === -1` as not focusable. This introduces as boolean flag (`includeNegativeTabIndex`) that disables the `tabIndex < 0` filter. Also made the logic more generic to accommodate for `tabIndex` values other than `-1` and `0`. And I added some tests :)

### DIFF
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -525,12 +525,6 @@ function releaseOwnershipForEventComponentInstance(
   return false;
 }
 
-function isInteger(value: any): boolean {
-  return (
-    typeof value === 'number' && isFinite(value) && Math.floor(value) === value
-  );
-}
-
 function isFiberHostComponentFocusable(
   fiber: Fiber,
   includeNegativeTabIndex: boolean,
@@ -539,13 +533,13 @@ function isFiberHostComponentFocusable(
     return false;
   }
   const {type, memoizedProps} = fiber;
-  if (memoizedProps.disabled) {
+  if (
+    (!includeNegativeTabIndex && memoizedProps.tabIndex === -1) ||
+    memoizedProps.disabled
+  ) {
     return false;
   }
-  if (isInteger(memoizedProps.tabIndex)) {
-    return includeNegativeTabIndex ? true : memoizedProps.tabIndex >= 0;
-  }
-  if (memoizedProps.contentEditable === true) {
+  if (memoizedProps.tabIndex === 0 || memoizedProps.contentEditable === true) {
     return true;
   }
   if (type === 'a' || type === 'area') {
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -389,14 +389,20 @@ const eventResponderContext: ReactDOMResponderContext = {
       }
     }
   },
-  getFocusableElementsInScope(): Array<HTMLElement> {
+  getFocusableElementsInScope(
+    includeNegativeTabIndex: boolean,
+  ): Array<HTMLElement> {
     validateResponderContext();
     const focusableElements = [];
     const eventComponentInstance = ((currentInstance: any): ReactDOMEventComponentInstance);
     const child = ((eventComponentInstance.currentFiber: any): Fiber).child;
 
     if (child !== null) {
-      collectFocusableElements(child, focusableElements);
+      collectFocusableElements(
+        child,
+        focusableElements,
+        includeNegativeTabIndex,
+      );
     }
     return focusableElements;
   },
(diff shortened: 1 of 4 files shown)


necolas:
> Also made the logic more generic to accommodate for tabIndex values other than -1 and 0

I don't think we should look to support values other than `-1` or `0`, as positive values create accessibility problems.

We're also using this event component with UI components that set tabindex to `-1` by default, and have their tabindex indirectly managed via a `focusable` prop which will set the value to `0`. Are you using the `Focus` event component? Or are you building an event responder?

voideanvalue:
> I don't think we should look to support values other than -1 or 0, as positive values create accessibility problems.

That's a fair point. I'll undo that change and only accept `0` and `-1`.

> Are you using the `Focus` event component? Or are you building an event responder?

I'm playing with the `Focus` event component and also building an event responder (a variant of `FocusScope` to enable roving tabindex on lists). I can share a fully working prototype (once I get it to that stage) with you if you think that'd be useful to see!

necolas:
I'm not sure how this would be useful in a context where every View renders down to `<div tabindex="-1">`. The set of elements you'd get back would be every View's host node in the subtree. Would that be a problem? How are you hoping to use the result?

voideanvalue:
> where every View renders down to `<div tabindex="-1">`

If every `<View>` renders down to `<div tabindex="-1">`, I'd consider every `<View>` focusable (though not tabbable because of the negative `tabindex` value).

Basing this on reading https://html.spec.whatwg.org/multipage/interaction.html#attr-tabindex and https://html.spec.whatwg.org/multipage/interaction.html#focusable-area. Specifically:

> The tabindex content attribute allows authors to indicate that an element is supposed to be focusable, and whether it is supposed to be reachable using sequential focus navigation

> If the value is a negative integer
The user agent must set the element's tabindex focus flag, but should omit the element from the sequential focus navigation order.

> The following table describes what objects can be focusable areas ... Elements that have their tabindex focus flag set, that are not actually disabled

When I say a tabbable element (or a tab stop), I mean that the element is part of the "sequential focus navigation order" as defined by the spec.

Probably also worth noting that I feel that the name `FocusScope` doesn't accurately represent the its current implementation (it manages tabbability / the "sequential focus navigation order", not all kinds of focus).

___

> The set of elements you'd get back would be every View's host node in the subtree. Would that be a problem?

I'm not very familiar with the semantics of the `View` component or the recommended uses. If every `div` in a web app came from a `View` I'd consider it a problem, but more so because I don't expect every `div` to become a focusable element.

___

> How are you hoping to use the result?

My motivation for wanting to be able to distinguish between tabbable and focusable elements is rooted the use case of implementing roving tabindex (use tab key to move focus to across lists and arrow keys to move focus across list items within the current list).

___

An alternative to this pull request could be adding `getFocusableElementsInScope` and `getTabbableElementsInScope` to `ReactDOMResponderContext` (that'd be a breaking change).

trueadm:
You should check out the other PR that adds focus manager to FocusScope too. There might be ideas that can be expanded on if similar. https://github.com/facebook/react/pull/15849

necolas:
> When I say a tabbable element (or a tab stop), I mean that the element is part of the "sequential focus navigation order" as defined by the spec.

I understand the semantic differences, although "tabbable" doesn't communicate that the element can also receive focus from other user actions, like mouse or touch. That's one reason why we avoid using "tabbable" for multi-input APIs.

> If every div in a web app came from a View I'd consider it a problem, but more so because I don't expect every div to become a focusable element.

This is what I'm trying to understand; whether you might be making assumptions about what elements will have `tabindex="-1"` for this specific pattern. Flare is intended to provide building blocks for a cross-platform React API, and that would mean not writing web apps with `div`, etc., at all. We might want to make it so that host nodes rendered by `View` are programmatically focusable by default (via `tabindex=-1`), or use tabindex in other scenarios that don't overlap well with other assumptions.

What is it that you want to do with a list of "focusable" elements? And why is it about `tabindex=-1|0` that makes that the appropriate marker for building that list?

> An alternative to this pull request could be adding `getFocusableElementsInScope` and `getTabbableElementsInScope` to `ReactDOMResponderContext`

I considered this too. We're using `focusable` in React Native to indicate that a View should be user-focusable (via keyboard, touch, or mouse), with programmatic focus of being more widely available and straight-forward than it is when working directly with the DOM, where people are used to managing their own `tabindex` values to make elements like `div` focusable.

devongovett:
FWIW, I added a `tabbable` option already in my other PR for focus management: https://github.com/facebook/react/pull/15849/files#diff-587e37bf8623dff5ba8e829346a9046dR53

voideanvalue:
> "tabbable" doesn't communicate that the element can also receive focus from other user actions, like mouse or touch. That's one reason why we avoid using "tabbable" for multi-input APIs.

Fair point :). I hadn't considered that.

> whether you might be making assumptions about what elements will have `tabindex="-1"` for this specific pattern ... What is it that you want to do with a list of "focusable" elements? And what is it about `tabindex=-1|0` that makes that the appropriate marker for building that list?

I might be making assumptions here. I'm trying my best to conform to the HTML Spec and ARIA guidelines. Please let me know if I'm misunderstanding something here.

I'm trying to follow the ARIA guidelines for using roving tabindex for focus management (https://www.w3.org/TR/wai-aria-practices/#kbd_roving_tabindex). Here's a specific use case I want to enable:
```
<ul>
  <li><a href tabindex="-1">List 1: Link 1</a><li>
  <li><a href tabIndex="0">List 1: Link 2</a><li>
  <li><a href tabindex="-1">List 1: Link 3</a><li>
</ul>
<ul>
  <li><a href tabindex="0">List 2: Link 1</a><li>
  <li><a href tabIndex="-1">List 2: Link 2</a><li>
</ul>
```
When `List 1: Link 2` has focus:
- pressing tab would move focus to `List 2: Link 1` (which is the next tabstop)
- pressing down arrow (assuming vertical list orientation) would move focus to `List 1: Link 3` and swap tabindex values (`List 1: Link 3` gets tabindex 0 and `List 1: Link 2` gets tabindex -1)

I have not given sufficient thought to the desirability of proliferating `<div tabindex="-1" />` all over the app. It seems a bit odd at first glance.

___

I'll spend some time looking at #15849 and #16009.

PR #16002, opened 2019-06-27
[Fresh] Handle changes at module boundaries

gaearon:
Normally we use the registration mechanism to "contain" changes. We edit a component, there's a new version of it, we update the registration. Cool. We don't want to re-evaluate the world. Just that file alone, if we can.

What if nothing registered during that file's evaluation? If there's nothing to register, there's nothing to update.. Right? Well, not quite.

If we had an `App` class, initially we registered it at the module boundary with a fallback `"[App.js] exports.default"` ID. But if we rewrite it to a function (imagine I'm a product developer moving to Hooks), in the new version we'll register it twice — first via transform with `"[App.js] App"` and then with `"[App.js] exports.default"` ID. First registration wins. As a result, `"[App.js] exports.default"` ID _keeps pointing to the old `App` class_. That happens on RN now.

As a result, when you convert a class to a function, you keep seeing the class.

We could patch over it in different ways. But I think the core problem here is that _if the families changed at the module boundary, we can't safely stop at that module_. (Especially in an environment without inline requires.) We need to re-evaluate parent modules, and apply the same algorithm to them, so that they re-render and use the latest versions of exports.

So I'm adding a mechanism that the module system can use to do this. The new method is `getFamilyByType`. The module system can compare the previous and next exports. If some were added or removed, we need to propagate the update. (This also helps fix the case where a parent references a deleted export from a child — that should become an error by re-rendering the parent.) The new method also lets us propagate the update if *registered family* becomes different for that export. So that catches cases like class-to-function, function-to-class, wrap-in-HOC, export-a-different-function-component-from-the-same-file.

I added some integration tests to show the intent although I won't be fully modeling it here. That code will live in the module system.

### DIFF
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -37,19 +37,18 @@ if (!__DEV__) {
 
 // In old environments, we'll leak previous types after every edit.
 const PossiblyWeakMap = typeof WeakMap === 'function' ? WeakMap : Map;
-const PossiblyWeakSet = typeof WeakSet === 'function' ? WeakSet : Set;
 
 // We never remove these associations.
 // It's OK to reference families, but use WeakMap/Set for types.
 const allFamiliesByID: Map<string, Family> = new Map();
-// $FlowIssue
-const allTypes: WeakSet<any> | Set<any> = new PossiblyWeakSet();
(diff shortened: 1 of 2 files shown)


gaearon:
I've already synced this to RN anyway, and it's purely additive. So I'll get it in. Happy to follow up.

PR #16005, opened 2019-06-27
SuspenseList Optimizations

sebmarkbage:
This includes a bunch of refactoring and optimizations to SuspenseList. Individual commits help the review.

The biggest optimization focus on the common case of deep updates inside a fully completed tree.

In the case where nothing in the list was suspended and the children didn't change, we now avoid scanning through the list in the begin phase since we don't need to find a tail.

Additionally, if nothing new suspended in this whole render pass, we can also skip the scan in the complete phase.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -20,7 +20,7 @@ import type {
 import type {ReactEventComponentInstance} from 'shared/ReactTypes';
 import type {
   SuspenseState,
-  SuspenseListState,
+  SuspenseListRenderState,
 } from './ReactFiberSuspenseComponent';
 import type {SuspenseContext} from './ReactFiberSuspenseContext';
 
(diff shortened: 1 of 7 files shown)


packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -953,16 +959,20 @@ function completeWork(
...
-          const needsRerender = hasSuspendedChildrenAndNewContent(
-            workInProgress,
-            renderedChildren,
-          );
+
+          // We might be suspended if something in this render pass suspended, or
+          // something in the previous committed pass suspended. Otherwise,
+          // there's no chance so we can skip the expensive call to
+          // hasSuspendedChildrenAndNewContent.
+          let cannotBeSuspended =
```
acdlite: This seems like something we could put on the `effectTag` or a similar bitfield that gets bubbled up during complete/unwind.

sebmarkbage: Yea I wanted a tag on each child so that I can use it to efficiently determine the tail in beginWork.

PR #16007, opened 2019-06-27
Add tail="collapsed" option to SuspenseList

sebmarkbage:
Builds on top of #16005.

This add an option to SuspenseList which ensures that we avoid inserting any new rows at the tail of the list past some point.

The first option is "collapsed" which means that there is only one row visible at the end of the list. I plan on adding a "hidden" option which ensures that zero visible rows are at the end.

Note that the tail in terms of unfolding row-by-row is defined as any insertions at the bottom or any rows that updated to become suspended or were already inserted from previous commits. This option doesn't actually remove or hide those. However, this is a rare edge cases. Typically you're expected to clear the list for these cases.

The use case here is for streaming rendering of items. This lets you provide more rows to React than you have available yet. The main purpose of this is for server rendering where you can't update to add more. However, it also is useful so that React can prerender later rows while blocked on previous rows. This PR doesn't actually do that yet tho.

This option also allows to render CPU bound work, one item at a time without showing all the fallbacks. Thanks to the tail expiration time.

## Collapsing to the last row

The `tail="collapsed"` option uses the first new row for showing the fallback state. That's fairly efficient because we've rendered it when we tried the last row.

If you want to instead show the last row, you have to use a nested suspense list and the "hidden" option of the inner one:

```
<SuspenseList revealOrder="forward">
  <SuspenseList revealOrder="forward" tail="hidden">
    ...
  </SuspenseList>
  <Tail />
</SuspenseList>
```

Tail in this case can be a Suspense boundary that unsuspends when we're not expected to get more items added to the list.

```
function Blocker({suspendOn}) {
  if (!suspendOn.isResolved()) throw something();
  return null;
}
function Tail() {
  return <Suspense fallback={<Shimmer />}>
    <Blocker suspendOn={endOfTheList} />
  </Suspend>;
}
```

This technique ensures that if `endOfTheList` resolves on the server, the SSR streaming can hide the tail shimmer. No need for client rerenders. Placing it in an outer SuspenseList also ensures that once `endOfTheList` is resolved we don't hide the Shimmer before the rows in the list are fully loaded. Otherwise, the shimmer would hide first and then we'd wait for the inner rows but they're hidden.

This is a bit unfortunate since I think this is a pretty common use case for visualizing a tail load differently from suspended existing boundaries.

In theory we could add an option to collapse into the last row but it gets tricky. Because after the first commit, that row would now be mounted and now it's not a new insertion anymore. That complicates the semantics a bit. The collapsed tail would have to be something like everything after the last committed row that isn't the last one.

This approach doesn’t actually work:

```
<SuspenseList>
  {rows}
  <Tail />
</SuspenseList>
```

Because that’s actually just two rows since the fragment in that case is treated as a single row.

I feel like it's just clearer with two lists. That also makes it clearer what a "row" means. E.g. if your tail has multiple rows, it kind of just works.

I also suspect that at some point SuspenseList will accept a custom streaming data type and in that case it might be easier to think of the inner list as one stream and the outer as another.

Although a counter point is that virtualization might make this api difficult/impossible to use.

A possible option would be to add an explicit tail component:

```
<SuspenseList tailMarker={<Tail />}>
  ...
</SuspenseList>
```

However, that would require two fibers to be stored similar to how Suspense inserts extra fibers.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -14,6 +14,7 @@ import type {ExpirationTime} from './ReactFiberExpirationTime';
 import type {
   SuspenseState,
   SuspenseListRenderState,
+  SuspenseListTailMode,
 } from './ReactFiberSuspenseComponent';
 import type {SuspenseContext} from './ReactFiberSuspenseContext';
 
@@ -188,6 +189,7 @@ let didWarnAboutFunctionRefs;
 export let didWarnAboutReassigningProps;
 let didWarnAboutMaxDuration;
 let didWarnAboutRevealOrder;
+let didWarnAboutTailOptions;
 
 if (__DEV__) {
   didWarnAboutBadClass = {};
@@ -198,6 +200,7 @@ if (__DEV__) {
   didWarnAboutReassigningProps = false;
   didWarnAboutMaxDuration = false;
   didWarnAboutRevealOrder = {};
+  didWarnAboutTailOptions = {};
 }
 
 export function reconcileChildren(
@@ -2063,11 +2066,40 @@ function validateRevealOrder(revealOrder: SuspenseListRevealOrder) {
   }
 }
 
+function validateTailOptions(
+  tailMode: SuspenseListTailMode,
+  revealOrder: SuspenseListRevealOrder,
+) {
+  if (__DEV__) {
+    if (tailMode !== undefined && !didWarnAboutTailOptions[tailMode]) {
+      if (tailMode !== 'collapsed') {
+        didWarnAboutTailOptions[tailMode] = true;
+        warning(
+          false,
+          '"%s" is not a supported value for tail on <SuspenseList />. ' +
+            'Did you mean "collapsed"?',
+          tailMode,
+        );
+      } else if (revealOrder !== 'forwards' && revealOrder !== 'backwards') {
+        didWarnAboutTailOptions[tailMode] = true;
+        warning(
+          false,
+          '<SuspenseList tail="%s" /> is only valid if revealOrder is ' +
+            '"forwards" or "backwards". ' +
+            'Did you mean to specify revealOrder="forwards"?',
+          tailMode,
+        );
+      }
+    }
+  }
+}
+
 function initSuspenseListRenderState(
   workInProgress: Fiber,
   isBackwards: boolean,
   tail: null | Fiber,
   lastContentRow: null | Fiber,
+  tailMode: SuspenseListTailMode,
 ): void {
   let renderState: null | SuspenseListRenderState =
     workInProgress.memoizedState;
@@ -2078,6 +2110,7 @@ function initSuspenseListRenderState(
       last: lastContentRow,
       tail: tail,
       tailExpiration: 0,
+      tailMode: tailMode,
     };
   } else {
     // We can reuse the existing object from previous renders.
@@ -2086,6 +2119,7 @@ function initSuspenseListRenderState(
     renderState.last = lastContentRow;
     renderState.tail = tail;
     renderState.tailExpiration = 0;
+    renderState.tailMode = tailMode;
   }
 }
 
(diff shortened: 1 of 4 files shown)


bvaughn:
> A possible option would be to add an explicit tail component:
> 
> ```
><SuspenseList tailMarker={<Tail />}>
>  ...
> </SuspenseList>
> ```
>However, that would require two fibers to be stored similar to how Suspense inserts extra fibers.

This API was my initial intuition as well. I'm not sure I understand the statement about requiring two fibers. Wouldn't nested lists also require two fibers?

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
```diff
@@ -1089,4 +1088,503 @@ describe('ReactSuspenseList', () => {
...
+
+    await B.resolve();
+
+    expect(Scheduler).toFlushAndYield(['B', 'Suspend! [C]']);
+
+    // Incremental loading is suspended.
+    jest.advanceTimersByTime(500);
+
+    // Even though B is unsuspended, it's still in loading state because
+    // it is blocked by C.
```
bvaughn: Seems counter-intuitive to me that B is blocked by C in this case. Since we display both loading states, I would also have expected them to resolve in order.

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
```diff
@@ -1089,4 +1088,503 @@ describe('ReactSuspenseList', () => {
...
+    expect(Scheduler).toFlushAndYield(['C']);
+
+    expect(ReactNoop).toMatchRenderedOutput(
+      <Fragment>
+        <span>A</span>
+        <span>B</span>
+        <span>C</span>
+      </Fragment>,
+    );
+  });
```
bvaughn: Worth adding some tests that resolve things out of order?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2010,6 +2013,119 @@ function findLastContentRow(firstChild: null | Fiber): null | Fiber {
...
+    if (tailOptions !== undefined && !didWarnAboutTailOptions[tailOptions]) {
+      if (tailOptions !== 'collapsed') {
+        didWarnAboutTailOptions[tailOptions] = true;
+        warning(
+          false,
+          '"%s" is not a supported value for tail on <SuspenseList />. ' +
+            'Did you mean "collapsed"?',
+          tailOptions,
+        );
+      } else if (revealOrder !== 'forwards' && revealOrder !== 'backwards') {
```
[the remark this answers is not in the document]
sebmarkbage: So that you don't get both warnings at once. Seems more actionable to pick a legit value first. E.g. if you specified something like "none", it would be misleading to recommend a revealOrder.

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
sebmarkbage: This isn't new behavior in this PR and this is basically just replicating the same assertion as other tests. The rationale for this behavior is that in a variable size list you don't want to pop in things multiple time to shift the below content around multiple times. Therefore the "head" resolves in the "together" mode.

This is kind of a rare case though since insertions in the middle are rare. Let alone multiple of them. A more common variant is inserting at the top but in that case you don't want to start with the top most and then expand them one by one to shift things down. More likely it's better to just show them all as a unit.

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
sebmarkbage: Some of the other ones resolve out of order, but I don't really want to add every combination. That's more for a fuzz tester.

Implementation detail wise there's also not much to test because we don't even render the future items so we have not attached any listeners.

packages/react-reconciler/src/ReactFiberBeginWork.js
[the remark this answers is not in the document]
bvaughn: Ah. I guess that makes sense. I was worried it could feel a bit like warning-whack-a-mole if you fixed one just to see another pop up.

The `revealOrder` would still be invalid if you specified `tail="none"` though so wouldn't we still want to warn about it?

I dunno. I don't feel strongly here 😄 Was just asking.

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
bvaughn: Okay. Implementation details can always change though 😁

sebmarkbage:
Found and fixed one more bug. https://github.com/facebook/react/pull/16007/commits/ccdb8b00dac3d989cb1255e724966ff36962c4a7

PR #16008, opened 2019-06-27
Enable soucemaps while run 'yarn debug-test'

flyOstrich:
the question is :
when i ran 'yarn debug-test' and debug from chrome://inspect , i got minified codes like this:
`it('should return the only child', function () {
    var instance =
    React.createElement(WrapComponent, { __source: { fileName: _jsxFileName, lineNumber: 70 } },
      React.createElement('span', { __source: { fileName: _jsxFileName, lineNumber: 71 } }));
    expect(React.Children.only(instance.props.children)).toEqual(React.createElement('span', { __source: { fileName: _jsxFileName, lineNumber: 74 } }));
  });
`
actually,i want to get source code when i debug test , so i added  sourceMaps:"inline" to enable this feature

### DIFF
--- a/scripts/jest/preprocessor.js
+++ b/scripts/jest/preprocessor.js
@@ -43,6 +43,7 @@ const babelOptions = {
     require.resolve('../babel/transform-prevent-infinite-loops'),
   ],
   retainLines: true,
+  sourceMaps: 'inline',
 };
 
 module.exports = {


zzuu666:
This PR is useful. When I run `yarn debug-test` and wanna debug `ReactDOM.render(<div />, container)`.  but I got an error position in source code.
I added same code as this PR, my problem is solved.

Debug runtime env:
node: 10.15.1
yarn: 1.15.2
babel/*: 7.x

gaearon:
Can you add screenshots? I tried this and still got garbled code.

zzuu666:
This is my test debug code:
break-point

without `sourceMap: inline`, I don't get the correct position:
without

with `sourceMap: inline`, It's working:
with

PR #16011, opened 2019-06-27
[Flare] Remove event targets including TouchHitTarget

trueadm:
This PR removes event target components from React and all related tests. This should drop file by around 1kb min+gzip and also reduce the complexity and overhead in managing event targets.

Currently, we don't use event targets anywhere and we have no short-term plans on using them. So far, many of the ideas we've had in terms of using them has shown there to be better alternatives to be used instead. We planned to use touch hit slop internally but we found we never actually needed them to improve touch interactivity (mobile browsers add a native touch hit slop when used with touch events rather than emulated mouse events it seems).

In the future, if we need them again, we can revisit attaching them via another route that might be less specific from the use-case that these were originally intended (touch hit slop).


sebmarkbage:
If I do as I suggested of making all of the event components box-only (pointer-events: none children), then I suspect that we’ll want this back because it is a way to expand the hit target explicitly outside the box-only.

trueadm:
@sebmarkbage We tried doing something similar, having pointer-events: none caused issues on FB5 from what I recall when we tried it. Furthermore, I have a better approach that I want to run by you offline. :)

PR #16013, opened 2019-06-27
Wrap ReactTestRenderer .create and .update with act() by default

threepointone:
Read this PR with `?w=1`

(I had an older version of this PR, made a fresh one to get out of conflict hell)

This PR adds `act()` by default to react-test-renderer's `.create`, `.update`, and `.unmount` methods.

We risk a major breaking change with having added warnings on unqueued effects outside of an act scope. Eg: A lot of people's snapshot tests will break; this would be good, if they didn't have to first wrap their renderer helper with `act()`, and then update snapshots.

Broadly, we could also make a case why this makes sense to any consumers; they'll be using it only in testing scenarios, which we'd want to wrap in `act()` anyway

For our internal tests, this had the effect that a larger portion of the test was just wrapped with an outer `act()`, letting us test incremental updates by advancing the `Scheduler` manually.

### DIFF
--- a/fixtures/dom/src/index.test.js
+++ b/fixtures/dom/src/index.test.js
@@ -42,10 +42,10 @@ it('resets correctly across renderers', () => {
     React.useEffect(() => {}, []);
     return null;
   }
-  TestUtils.act(() => {
-    TestRenderer.act(() => {});
+  TestRenderer.act(() => {
+    TestUtils.act(() => {});
     expect(() => {
-      TestRenderer.create(<Effecty />);
+      ReactDOM.render(<Effecty />, document.createElement('div'));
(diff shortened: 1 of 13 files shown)


threepointone:
closing this. might revisit in v17, or not. who knows!

xugao:
@threepointone - what's the latest status on this change? it's not very clear to me why this change is postponed :)

threepointone:
It didn’t seem like a good idea then, can’t recollect why; probably because it would be unexpected behaviour compared to other renderers. I’m not a part of the team anymore, so I can’t comment on future plans.

xugao:
thanks for the fast response @threepointone !
@acdlite - is there any plan to re-activate this change?

Issue #15726, opened 2019-05-23
React.info() and React.log()


bvaughn:
Maybe we should add `.log` and `.info` wrappers (similar to our `.warn` and `.info`) that the DevTools could override/disable while re-rendering a function component to inspect its hooks.

The beta DevTools current behavior caused some confusion for one of the early adopters:
https://github.com/facebook/react-devtools/issues/1215#issuecomment-495382645

bvaughn:
I think my personal preference, after a little consideration, is the third option- `React.debug(callback)`

sompylasar:
I think the proposal tries to fight with the call side-effects (logging, hitting breakpoints) but does not address the original problem that DevTools is solving, which is "DevTools re-renders function components that use hooks in order to __build up metadata to display in the selected element panel.__" – why doesn't React expose getting this required metadata as an API, so it forces DevTools to use workarounds to gather hooks information (call the components and thus trigger their call side-effects)?

bvaughn:
> why doesn't React expose getting this required metadata as an API, so it forces DevTools to use workarounds to gather hooks information

There are two reasons for this design:
* DevTools needs a lot more information than React does. For example, it's useful for DevTools to be able to identify and display "custom hooks" (functions with names that start with "use" that call other hooks). Identifying custom hooks relies on the relatively "expensive" operation of parsing the call stack.
* Even for built in hooks, we would not want to add additional overhead to React for every component that uses hooks just to support the uncommon case of displaying hooks values for a component selected in DevTools at some point in time.

sebmarkbage:
@sompylasar Fixing that also doesn't fix the other issue. React recalls functions multiple times during DEV mode and in concurrent mode can rebase and call the function multiple times for a single render. The contract is that we can call the function as much as we want without side-effects.

bvaughn:
I think the original DevTools behavior that felt particularly broken was that we polled the component once a second. So it felt like there was an infinite loop. Now that we cache the results and only invalidate when the component is updated, it will "feel" more inline with React's normal rendering flow.

sompylasar:
> React recalls functions multiple times during DEV mode and in concurrent mode can rebase and call the function multiple times for a single render. The contract is that we can call the function as much as we want without side-effects.

This is somewhat expected because this is not a part of debugging, it's a part of the program being debugged. Ideally a debugger does not alter the behavior of the program being debugged, and that is what I am concerned about with renders being called more than usual when DevTools is present.

[...]

PR #16017, opened 2019-06-28
Replace React.error and React.warn with getComponentStack

bvaughn:
Proposed alternative to #15170

Several people have asked, "_Could you provide a method to access the component stack directly?_". I think our reasoning against this was that we didn't want people to _depend_ on that stack, or to try to parse it and build anything on top of it.

I don't think we considered the case of instrumenting error logging though. Facebook (and probably others) overrides `console.error` to capture the call stack and save it somewhere. A downside of `React.error` is that it would cause errors from different call sites to be grouped together (unless the instrumenting code was smart enough to skip over the React frame).

Overriding only these two console methods also feels like a bit of an incomplete solution (#15726) to the logging story.

This PR replaces `React.warn` and `React.error` with a method that returns the current component stack¹ which people can log to whichever console group they want.

¹ At the moment, this method always returns an empty string in prod.


packages/react/src/getComponentStack.js
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import ReactSharedInternals from 'shared/ReactSharedInternals';
+
+let getComponentStack = function() {
+  return '';
```
bvaughn: Should this not return anything? Seemed weird to return void/undefined in prod and an empty string when called outside of the render phase in DEV, so I opted to return empty string in both cases for now.

sebmarkbage:
It’s not just about the component stack but other things we can do with it like reorder and delay etc.

This doesn’t address the abuse problem. The reason this wasn’t earlier is because there were many really bad use cases people were intending to use this for.

bvaughn:
> It’s not just about the component stack but other things we can do with it like reorder and delay etc.

That's a topic for a broader logging strategy. I see it as separate from the more focused need of accessing a console stack. Do you disagree?

> This doesn’t address the abuse problem. The reason this wasn’t earlier is because there were many really bad use cases people were intending to use this for.

TBH bad actors would probably misuse e.g. `React.warn` as well by intercepting calls to `console.warn`. I think we can only advocate for **not**  doing that. On the other hand, the current approach (from #15170) doesn't account for existing error logging/aggregating use cases, which is the main reason I opened this PR in the first place...

bvaughn:
Sebastian and I had a productive conversation this afternoon regarding my concerns above.

The quick takeaway is this: I will be updating this PR to remove `React.warn` and `React.error` _without_ adding an additional `getComponentStack` method. We can revisit these methods later as part of a larger discussion around the console logging experience with React (see #15726) but we should not block 16.9 for this.

In the meanwhile, I will be adding functionality to React DevTools to intercept calls to `console.error`, `console.warn`, and `console.trace` and auto-append component stacks (in DEV mode). This seems like a nice 80/20 solution that doesn't limit our options for a future API.

Might also be worth looking into a log view within DevTools that groups and orders logs by commit, etc.

bvaughn:
Closing in favor of #16126

PR #16019, opened 2019-06-30
[Fresh] Clone a custom hook node before use

theKashey:
This is a small patch to react-refresh babel plugin to fix a problem we found in [react-hot-loader](https://github.com/gaearon/react-hot-loader/issues/1268).

In short - refresh babel plugin is using the "original" _babel node_ to put a custom hook into `signature`, and `common-js` babel plugin has a logical condition not to transform any node twice. As a result - custom hook reference would not be transformed. @dgreensp did an amazing job finding the root cause.

Solution - just `clone` node before the use. Look like there is no need for deep clone in this case.

Keep in mind - __no tests were added__. Unfortunately, this bug is not reproducible with babel 6, and only babel 7 is affected. To be more concrete - I've added tests, but as long as they are not testing anything - they are passing initially - I've removed them.

### DIFF
--- a/packages/react-refresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-refresh/src/ReactFreshBabelPlugin.js
@@ -245,7 +245,7 @@ export default function(babel) {
       key: fnHookCalls.map(call => call.name + '{' + call.key + '}').join('\n'),
       customHooks: fnHookCalls
         .filter(call => !isBuiltinHook(call.name))
-        .map(call => call.callee),
+        .map(call => t.cloneDeep(call.callee)),
     };
   }
 
--- a/packages/react-refresh/src/__tests__/ReactFreshBabelPlugin-test.js
+++ b/packages/react-refresh/src/__tests__/ReactFreshBabelPlugin-test.js
@@ -15,7 +15,12 @@ function transform(input, options = {}) {
   return wrap(
     babel.transform(input, {
       babelrc: false,
-      plugins: ['syntax-jsx', 'syntax-dynamic-import', freshPlugin],
+      plugins: [
+        'syntax-jsx',
+        'syntax-dynamic-import',
+        freshPlugin,
+        ...(options.plugins || []),
+      ],
     }).code,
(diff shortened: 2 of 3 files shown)


dgreensp:
I think deep clone is necessary for related cases.  I can repro today or tomorrow if you want.  I haven’t tried it yet, but I would try “import *” and importing a name foo and doing foo.useHook.

gaearon:
It would be nice to get a confirmed repro case that fails with Babel 7, even in a fork. Then we can keep the test, and it will become useful once we update to Babel 7. Which we should do anyway.

theKashey:
Ok - test would be green today, and would be green in the future, as long as the problem is "fixed". But, at least, it would prevent regression.

@dgreensp - It does not matter how you've imported "hook", only `node.callee` is use, and that's a flat node.

gaearon:
What if callee is something like `Foo.Bar.Baz.useSomething()`?

theKashey:
> What if callee is something like Foo.Bar.Baz.useSomething()?

Does not matter - only "top level" transformation matters here(and for this case),  le transformation from `Foo` to `_utils.Foo` is equal to `Foo.Bar.Baz` => `_utils.Foo.Bar.Baz`. 
Deep clone might be useful to prevent problem with yet unknown plugins, but they might not exists.

Anyway - if the goal is to provide more safe out of the box solution - let's use deep clone.

gaearon:
Yeah, feels a bit safer. Let's use the deep one.

dgreensp:
My intuition was that foo.bar.baz() is actually (((foo).bar).baz)(), so the callee is foo.bar.baz and the imported identifier is foo, a descendent node.  I tried to repro it real quick and couldn’t, but it sounds like you’re going to go with deep, so sounds good to me.

gaearon:
Thanks!

theKashey:
😅 someone got that _theoretical_ problem, which might require `cloneDeep` - https://github.com/gaearon/react-hot-loader/issues/1280

PR #16024, opened 2019-07-01
Add tail="hidden" option to SuspenseList

sebmarkbage:
Builds on top of #16007.

This is similar to `tail="collapsed"` except it hides all rows in the tail. The added complexity here is that sometimes we only know that we've hit the tail after we've already rendered this row. So we need undo the effects added by rendering the row.

Additionally, by not committing the boundary, there's nothing to retry. So we need to transfer the promise set to the list so that we get a retry if the only thing to ping was in the tail.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -92,7 +92,7 @@ import {
   ForceSuspenseFallback,
   setDefaultShallowSuspenseContext,
 } from './ReactFiberSuspenseContext';
-import {isShowingAnyFallbacks} from './ReactFiberSuspenseComponent';
+import {findFirstSuspended} from './ReactFiberSuspenseComponent';
 import {
   isContextProvider as isLegacyContextProvider,
   popContext as popLegacyContext,
@@ -1059,7 +1059,8 @@ function completeWork(
       } else {
         // Append the rendered row to the child list.
         if (!didSuspendAlready) {
-          if (isShowingAnyFallbacks(renderedTail)) {
+          let suspended = findFirstSuspended(renderedTail);
+          if (suspended !== null) {
             workInProgress.effectTag |= DidCapture;
             didSuspendAlready = true;
             cutOffTailIfNeeded(renderState, true);
@@ -1070,9 +1071,11 @@ function completeWork(
             ) {
               // We need to delete the row we just rendered.
               // Ensure we transfer the update queue to the parent.
-              // TODO: This just reuses the code from the other path but could
-              // be optimized better.
-              hasSuspendedChildrenAndNewContent(workInProgress, renderedTail);
+              let newThennables = suspended.updateQueue;
+              if (newThennables !== null) {
+                workInProgress.updateQueue = newThennables;
+                workInProgress.effectTag |= Update;
+              }
               // Reset the effect list to what it w as before we rendered this
               // child. The nested children have already appended themselves.
               let lastEffect = (workInProgress.lastEffect =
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -930,112 +1004,160 @@ function completeWork(
...
+            didSuspendAlready = true;
 
-              const nextPriority = renderExpirationTime - 1;
-              workInProgress.expirationTime = workInProgress.childExpirationTime = nextPriority;
-              if (enableSchedulerTracing) {
-                markSpawnedWork(nextPriority);
-              }
-            } else {
-              suspenseListState.didSuspend = isShowingAnyFallbacks(rendered);
+            cutOffTailIfNeeded(renderState);
```
sebmarkbage: This assumes that this didn't suspend but it might have and that would affect whether we keep the rendered row or not. We should probably move the isShowingAnyFallbacks branch up.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -930,112 +1004,160 @@ function completeWork(
...
+            // This might have been modified.
+            if (renderState.rendering === null) {
+              // Ensure we transfer the update queue to the parent.
+              // TODO: This just reuses the code from the other path but could
+              // be optimized better.
+              hasSuspendedChildrenAndNewContent(workInProgress, renderedTail);
+              // Reset the effect list to what it w as before we rendered this
+              // child. The nested children have already appended themselves.
+              workInProgress.lastEffect = renderState.lastEffect;
```
sebmarkbage: We probably need to set `workInProgress.lastEffect.nextEffect = null;`. Most likely this will get set anyway but just to be safe. E.g. the suspense list its gets added to the end of the list.

sebmarkbage:
SSR is pretty tricky with this approach, since we won't have DOM nodes for all the items in the list during reconciliation. If we assume that we have the same number of items in the list and that we suspend at at least the same places as the server, then at least the collapsed mode works out. Because we'll at most render the last tail item which will be in fallback state on the server and get replaced on the client. However, with "hidden" mode, we don't know that we've hit the end of the list yet. We need to render it first and when we try to do that, we get a hydration error.

sebmarkbage:
I think this is ready for review and landing now.

I ended up needing to do a lot of factoring to make it nicer and found a number of bugs along the way.

Particularly, to minimize file size I ended up removing one optimization which causes extra rerenders. https://github.com/facebook/react/pull/16024/commits/6ec55cb229d90853947028cdfdf66808fe3b1796 I'm not sure this is the right call but maybe with resuming this won't matter much.

I'll address the SSR case in a follow up.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -1070,9 +1071,11 @@ function completeWork(
             ) {
               // We need to delete the row we just rendered.
               // Ensure we transfer the update queue to the parent.
-              // TODO: This just reuses the code from the other path but could
-              // be optimized better.
-              hasSuspendedChildrenAndNewContent(workInProgress, renderedTail);
+              let newThennables = suspended.updateQueue;
+              if (newThennables !== null) {
+                workInProgress.updateQueue = newThennables;
```
acdlite: Is this approach something we'll want to rethink later? By always attaching thenables to the SuspenseList instead of the inner boundary, perhaps.

sebmarkbage: We could possibly do that in throwException, yea.

PR #16027, opened 2019-07-01
Flush sync bug

lunaruan:
This fixes bug where `flushSync` runs with `NormalPriority` instead of `ImmediatePriority`

### DIFF
--- a/packages/react-reconciler/src/SchedulerWithReactIntegration.js
+++ b/packages/react-reconciler/src/SchedulerWithReactIntegration.js
@@ -172,13 +172,16 @@ function flushSyncCallbackQueueImpl() {
     let i = 0;
     try {
       const isSync = true;
-      for (; i < syncQueue.length; i++) {
-        let callback = syncQueue[i];
-        do {
-          callback = callback(isSync);
-        } while (callback !== null);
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/SchedulerWithReactIntegration.js
```diff
@@ -172,13 +172,16 @@ function flushSyncCallbackQueueImpl() {
...
-      for (; i < syncQueue.length; i++) {
-        let callback = syncQueue[i];
-        do {
-          callback = callback(isSync);
-        } while (callback !== null);
-      }
-      syncQueue = null;
+      runWithPriority(ImmediatePriority, () => {
+        for (; i < syncQueue.length; i++) {
+          let callback = syncQueue[i];
```
acdlite: Flow doesn't know that `runWithPriority` runs calls the passed in function immediately, so it doesn't know `syncQueue` isn't null at the time you access it.

But I think if you create a constant binding to `syncQueue` right before, then it will work:

```js
const queue = syncQueue;
runWithPriority(ImmediatePriority, () => {
  // queue is guaranteed to be non-null
});
```

PR #16030, opened 2019-07-02
Update Suspense Priority Warning to Include Component that Triggered Update

lunaruan:
Improved warning whenever lower priority events (ex. data fetching, page load) happen during a high priority update (ex. hover/click events) to include:
1.) Name of component that triggered the high priority update or
2.) Information that the update was triggered on the root


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2509,30 +2510,78 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
...
+      while (WIPNode != null) {
+        // Add the component that triggered the suspense
+        const current = WIPNode.alternate;
+        if (current !== null) {
+          switch (WIPNode.tag) {
+            case HostRoot:
+            case ClassComponent:
+              // Loop through the component's update queue and see whether the component
+              // has triggered any high priority updates
+              let queueElem = (current.updateQueue || {}).firstUpdate;
```
acdlite: This should also be an explicit null check:

```js
const updateQueue = current.updateQueue;
if (updateQueue !== null) {
  let update = updateQueue.firstUpdate;
  while (update !== null) {
    // ...
  }
}
```

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2509,30 +2510,78 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
...
-      // negatives than false positives.
-      rootsWithPendingDiscreteUpdates !== null &&
-      workInProgressRoot !== null &&
-      renderExpirationTime ===
-        rootsWithPendingDiscreteUpdates.get(workInProgressRoot)
+      (currentPriorityLevel === UserBlockingPriority ||
+        currentPriorityLevel === ImmediatePriority)
     ) {
+      let WIPNode = sourceFiber;
+      while (WIPNode != null) {
```
acdlite: This should be an explicit null comparison: `!== null`

The same goes for all the other null comparisons in this function.

We use this convention in the React codebase because 1) explicit comparisons to null are faster than comparing to both `null` and `undefined` 2) since none of our types include `undefined` (with the exception of some dev-only fields), when `undefined` does show up somewhere, we can infer that it's probably a bug.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2546,20 +2595,35 @@ export function checkForWrongSuspensePriorityInDEV(sourceFiber: Fiber) {
...
+        .sort()
+        .join(', ');
+
+      const callerErroMessage = componentsThatCallSuspendedDiscreteUpdates.has(
+        HostRoot,
+      )
+        ? 'Component was suspended when root was mounted or updated'
+        : `The components that called suspense are: ${parentComponentNamesString}`;
+
+      componentsThatCallSuspendedDiscreteUpdates.clear();
```
acdlite: Set to `null` instead of clearing, same as `componentsWithSuspendedDiscreteUpdates` above.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2546,20 +2595,35 @@ export function checkForWrongSuspensePriorityInDEV(sourceFiber: Fiber) {
...
         componentNames.push(name);
       });
       componentsWithSuspendedDiscreteUpdates = null;
 
-      // TODO: A more helpful version of this message could include the names of
-      // the component that were updated, not the ones that suspended. To do
-      // that we'd need to track all the components that updated during this
-      // render, perhaps using the same mechanism as `markRenderEventTime`.
+      const parentComponentNamesString = Array.from(
+        componentsThatCallSuspendedDiscreteUpdates,
```
acdlite: Older browsers (and I think all versions of IE) don't support `Array.from`. That's why the code above uses a `forEach` loop instead.

In general, try to stay consistent with other parts of the codebase. But also don't be afraid to change existing code if you need to. For example, if we were to use `Array.from` here, then we would replace the `forEach` loop above, too.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2546,20 +2595,35 @@ export function checkForWrongSuspensePriorityInDEV(sourceFiber: Fiber) {
...
+      )
+        .sort()
+        .join(', ');
+
+      const callerErroMessage = componentsThatCallSuspendedDiscreteUpdates.has(
```
acdlite: Typo: Missing an "r" in "error"

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2509,30 +2510,78 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
 export const warnIfNotCurrentlyActingUpdatesInDev = warnIfNotCurrentlyActingUpdatesInDEV;
 
 let componentsWithSuspendedDiscreteUpdates = null;
+let componentsThatCallSuspendedDiscreteUpdates = new Set();
```
acdlite: This should be initialized to `null` so that it doesn't allocate memory on start up. Same as `componentsWithSuspendedDiscreteUpdates`.

Also let's rename these to remove the "discrete" since that no longer applies. Suggested names: `componentsThatTriggeredHighPriSuspend` and `componentsThatSuspendedAtHighPri`

packages/react-reconciler/src/ReactUpdateQueue.js
```diff
@@ -197,6 +202,7 @@ export function createUpdate(
   expirationTime: ExpirationTime,
   suspenseConfig: null | SuspenseConfig,
 ): Update<*> {
+  const priority = __DEV__ ? getCurrentPriorityLevel() : null;
```
acdlite: Same feedback I gave in ReactFiberUpdateQueue also applies here.

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -488,6 +488,7 @@ describe('ReactSuspenseWithNoopRenderer', () => {
   });
 
   it('tries rendering a lower priority pending update even if a higher priority one suspends', async () => {
+    spyOnDev(console, 'error');
```
acdlite: If you use `spyOnDev` then you should also assert the number and type of warnings at the end of the test. Otherwise this will suppress unexpected warnings.

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1616,6 +1620,81 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     expect(ReactNoop.getChildren()).toEqual([span('Loading...')]);
   });
 
+  it('warns when a low pri update suspends inside a high pri call for functional components', async () => {
```
acdlite: Terminology nit: Replace "call" with "update" since that's the term we usually use for this concept

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1616,6 +1620,81 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
+    }
+
+    let _setFoo;
+    function App() {
+      let [foo, setFoo] = React.useState(false);
+      _setFoo = setFoo;
+      return (
+        <Suspense fallback="Loading...">
+          {foo && <AsyncText text="A" />}
+          <B />
```
acdlite: Nit: I don't think B is an important part of this test, so you can remove it.

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1616,6 +1620,81 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
 
+  it('warns when a low pri update suspends inside a high pri call for functional components', async () => {
+    function B() {
+      return 'B';
+    }
+
+    let _setFoo;
+    function App() {
+      let [foo, setFoo] = React.useState(false);
+      _setFoo = setFoo;
```
acdlite: Nit: Call this "show" instead?

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1653,10 +1732,74 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
+    }
+
+    await ReactNoop.act(async () => {
+      ReactNoop.render(<App />);
+    });
+
+    // also make sure lowpriority is okay
+    await ReactNoop.act(async () => {
+      Scheduler.unstable_runWithPriority(
+        Scheduler.unstable_NormalPriority,
```
acdlite: Normal is the default priority (we're probably going to rename it to DefaultPriority, in fact), so you can remove this part of the test

packages/react-reconciler/src/ReactFiberWorkLoop.js
acdlite: This set also needs to be reset inside `prepareFreshStack`. Look for all the places that `componentsWithSuspendedDiscreteUpdates` is referenced and do the same thing.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2509,30 +2510,78 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
...
-      // My rationale is that it's better for this warning to have false
-      // negatives than false positives.
-      rootsWithPendingDiscreteUpdates !== null &&
-      workInProgressRoot !== null &&
-      renderExpirationTime ===
-        rootsWithPendingDiscreteUpdates.get(workInProgressRoot)
+      (currentPriorityLevel === UserBlockingPriority ||
+        currentPriorityLevel === ImmediatePriority)
     ) {
+      let WIPNode = sourceFiber;
```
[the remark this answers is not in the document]
lunaruan: workInProgress is already declared in the upper scope. Is renaming to `workInProgressNode` okay?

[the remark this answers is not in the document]
acdlite: Oh right. Yeah that's fine :D

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2520,58 +2522,138 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
...
+                  if (
+                    priorityLevel === UserBlockingPriority ||
+                    priorityLevel === ImmediatePriority
+                  ) {
+                    if (componentsThatTriggeredHighPriSuspend === null) {
+                      componentsThatTriggeredHighPriSuspend = new Set();
+                    }
+                    if (workInProgressNode.tag === HostRoot) {
+                      componentsThatTriggeredHighPriSuspend.add(
+                        workInProgressNode.tag,
```
acdlite: Ooh this is unfortunate. How about we add the whole fiber to the set? Then we can compute the names later during `flushSuspensePriorityWarningInDEV`.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2520,58 +2522,138 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
...
-      componentsWithSuspendedDiscreteUpdates = null;
+      componentsThatSuspendedAtHighPri = null;
+
+      const componentsThatTriggeredSuspendNames = [];
+      componentsThatTriggeredHighPriSuspend.forEach(name =>
+        componentsThatTriggeredSuspendNames.push(name),
+      );
+
+      const componentsThatTriggeredSuspendErrorMessage = componentsThatTriggeredHighPriSuspend.has(
+        HostRoot,
```
acdlite: How about we omit the root from this warning entirely? We can handle that separately inside `ReactDOM.render` and corresponding APIs (in a different PR), which will allow us to provide a more helpful warning for each renderer. This case is relatively rare, anyway.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2520,58 +2522,138 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
...
+      )
+        ? 'Component was suspended when root was mounted or updated'
+        : `The components that called suspense are: ${componentsThatTriggeredSuspendNames
```
acdlite: "The components that triggered the update"

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2520,58 +2522,128 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
...
+      (currentPriorityLevel === UserBlockingPriority ||
+        currentPriorityLevel === ImmediatePriority)
     ) {
+      let workInProgressNode = sourceFiber;
+      while (workInProgressNode !== null) {
+        // Add the component that triggered the suspense
+        const current = workInProgressNode.alternate;
+        if (current !== null) {
+          switch (workInProgressNode.tag) {
+            case HostRoot:
```
acdlite: Should remove this case entirely so you don't have to check it later

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2520,58 +2522,128 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
...
+                  if (
+                    priorityLevel === UserBlockingPriority ||
+                    priorityLevel === ImmediatePriority
+                  ) {
+                    if (componentsThatTriggeredHighPriSuspend === null) {
+                      componentsThatTriggeredHighPriSuspend = new Set();
+                    }
+                    // TODO: handle case where the component that triggers the high priority
+                    // suspend is in the HostRoot
+                    if (workInProgressNode.tag !== HostRoot) {
```
acdlite: Remove this check once you remove the case above

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2585,6 +2657,7 @@ function flushSuspensePriorityWarningInDEV() {
           'feedback, and another update to perform the actual change.',
         // TODO: Add link to React docs with more information, once it exists
         componentNames.sort().join(', '),
+        componentsThatTriggeredSuspendNames.sort().join(', '),
```
acdlite: This array could be empty. In that case we should skip the "The components that triggered the update: " part of the message.

packages/react-reconciler/src/ReactFiberWorkLoop.js
acdlite: Also let's add a TODO comment to remind us to warn about host roots separately (in `ReactDOM.render` et al)

PR #16035, opened 2019-07-02
flush suspense fallbacks at the end of act() calls.

threepointone:
- 'Saves' a reference to the handle+root in a Map whenever `scheduleTimeout` is called
- On exiting an `act()` scope, flushes them out, and clears the Map entry
- Added a test for the same, and updated another that expected differently. 

I wrote just the one test for this, but I'm keen to add more coverage. Any pointers to what other tests could be useful?

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -264,6 +264,8 @@ let spawnedWorkDuringRender: null | Array<ExpirationTime> = null;
 // receive the same expiration time. Otherwise we get tearing.
 let currentEventTime: ExpirationTime = NoWork;
 
+const SuspenseTimeouts: Map<TimeoutID, FiberRoot> = new Map();
+
 export function requestCurrentTime() {
   if ((executionContext & (RenderContext | CommitContext)) !== NoContext) {
     // We're inside React, so it's fine to read the actual time.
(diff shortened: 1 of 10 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -770,6 +772,8 @@ function prepareFreshStack(root, expirationTime) {
     root.timeoutHandle = noTimeout;
     // $FlowFixMe Complains noTimeout is not a TimeoutID, despite the check above
     cancelTimeout(timeoutHandle);
+    // $FlowFixMe Complains noTimeout is not a TimeoutID, despite the check above
+    SuspenseTimeouts.delete(timeoutHandle);
```
threepointone: It would've been nice to only have this for when in a jest environment. This is a place the testing specific build would have been great.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -264,6 +264,8 @@ let spawnedWorkDuringRender: null | Array<ExpirationTime> = null;
 // receive the same expiration time. Otherwise we get tearing.
 let currentEventTime: ExpirationTime = NoWork;
 
+const SuspenseTimeouts: Map<TimeoutID, FiberRoot> = new Map();
```
acdlite: This isn't gated behind a feature flag, or behind a DEV-check. Seems the only reason is for backwards compat with React DOM. But legacy code all runs in sync mode.

So how about we only enable this in the testing build? The normal build will warn if you're in concurrent mode and you're wrapped in `act` and you suspend.

threepointone: yeah I'm thinking that's a good idea too. this is ideal for being only in a testing build.

threepointone:
abandoning this to remove from active PRs, will revisit soon.

PR #16036, opened 2019-07-02
[Flare] Add more functionality to Scroll event responder

trueadm:
This PR adds more functionality to the `Scroll` event responder. Specifically:

- Adds the direction property on the event object
- Adds `onScrollDragStart` and `onScrollDragEnd`
- Removes `onMomentumScrollStart` and `onMomentumScrollEnd` Flow prop types.
- Fixed a bug when disabled never removed root events

I looked into how we might implement the momentum scroll events and ultimately decided it wasn't worth the logic to support. The issue is that trackpad events fire `wheel` and `scroll` for each bit of momentum and thus it's not possible to know if the user is doing this – or if it's a system setting (like on MacOS).

Furthermore, working out the difference in delta between scrolls and mousewheels is one possibility (as other third party libraries try do), but the amount of code it requires and the complexity in handling edge cases makes it non-trivial. Adding support for touch based momentum is simpler but then this feature feels half-baked. I'd much rather leave it as a TODO for later work.

### DIFF
--- a/packages/react-events/src/dom/Scroll.js
+++ b/packages/react-events/src/dom/Scroll.js
@@ -23,14 +23,16 @@ type ScrollProps = {
   onScroll: ScrollEvent => void,
   onScrollDragStart: ScrollEvent => void,
   onScrollDragEnd: ScrollEvent => void,
-  onScrollMomentumStart: ScrollEvent => void,
-  onScrollMomentumEnd: ScrollEvent => void,
 };
 
 type ScrollState = {
+  direction: ScrollDirection,
   pointerType: PointerType,
   scrollTarget: null | Element | Document,
-  isPointerDown: boolean,
+  isDragging: boolean,
+  isTouching: boolean,
+  scrollLeft: number,
+  scrollTop: number,
 };
 
 type ScrollEventType =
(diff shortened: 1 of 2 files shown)


packages/react-events/src/dom/Scroll.js
```diff
@@ -165,13 +230,19 @@ const ScrollResponder: ReactDOMEventResponder = {
         state.pointerType = pointerType;
         break;
       }
+      case 'wheel': {
+        state.pointerType = 'mouse';
```
necolas: Did you mention that `trackPad` fires both `scroll` and `wheel` events?

trueadm: I mentioned it in the summary of this PR. Did you mean for me to add a comment somewhere too?

necolas: Is the pointerType going to be correct here if trackPad is the source of the wheel event?

trueadm: There's no way to know if it's a trackpad or not. Mouse events can trigger wheel events too.

necolas: OK sounds like we'll just not have trackpad as a pointer type for scroll

PR #16038, opened 2019-07-02
[Fresh] Rename _getMountedRootCount to getMountedRootCount

gaearon:
I planned to initially only use it for testing. But I think it may also be useful to force a full reload if all roots are gone. Such as if the first mount failed and never committed. So I'm exposing it more officially.

### DIFF
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -474,8 +474,7 @@ export function injectIntoGlobalHook(globalObject: any): void {
   }
 }
 
-// Exposed for testing.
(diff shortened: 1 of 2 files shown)


sebmarkbage:
Seems like a vague reason for causing us future headaches. What’s this useful for?

gaearon:
I wanted to do something like 

```js
if (getRootCount() === 0) {
  forceFullReload()
}
```

which is useful for recovering from *initial render* errors in app with no error boundaries. (They never committed so there's nothing to retry.)

I really need just `hasMountedRoots` so I can do that?

gaearon:
Actually I guess just tracking commit count is sufficient for that.

PR #16039, opened 2019-07-02
allow nested `act()`s from different renderers

threepointone:
(an alternative to https://github.com/facebook/react/pull/15816/)

There are usecases where multiple renderers need to operate inside a single `act()` scope.
- `ReactDOM.render` being used inside another component tree. The parent component will be rendered using `ReactTestRenderer.create` for a snapshot test or something.
- a ReactDOM instance interacting with a ReactTestRenderer instance (like for the new devtools)

This PR changes the way the acting sigils operate to allow for this. It keeps 2 booleans, one attached to React, one attached to the renderer. `act()` changes these values, and the workloop reads them to decide which warning to trigger.

(I also renamed `shouldWarnUnactedUpdates` to `warnsIfNotActing`)


packages/react-dom/src/test-utils/ReactTestUtilsAct.js
```diff
@@ -86,17 +86,21 @@ let actingUpdatesScopeDepth = 0;
...
+  let previousReactIsActing;
+  let previousReactRendererIsActing;
   actingUpdatesScopeDepth++;
   if (__DEV__) {
-    previousActingUpdatesSigil = ReactCurrentActingRendererSigil.current;
-    ReactCurrentActingRendererSigil.current = ReactActingRendererSigil;
+    previousReactIsActing = ReactIsActing.current;
+    previousReactRendererIsActing = ReactIsActing.current;
+    ReactIsActing.current = true;
+    ReactRendererIsActing.current = true;
```
acdlite: Naming nit: `IsSomeRendererActing` and `IsThisRendererActing`

Issue #15924, opened 2019-06-19
[react-hooks/exhaustive-deps] False positive for nested dependencies in useEffect hook


johnloven:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
The following code

```javascript
const callback = useCallback(() => {
    console.log("CALLBACK");
}, []);

const wrapper = { callback };

useEffect(() => {
    wrapper.callback();
}, [wrapper.callback]);
```

gives the following `react-hooks/exhaustive-deps` warning

> React Hook useEffect has a missing dependency: 'wrapper'. Either include it or remove the dependency array

If the `useEffect` is replaced by the following

```javascript
useEffect(() => {
    const callbackReference = wrapper.callback;
    callbackReference();
}, [wrapper.callback]);
```

there are no warnings.

**What is the expected behavior?**
There should be no warnings in either case.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
react: 16.8.4
eslint-plugin-react-hooks: 1.6.0

bartlangelaan:
This issue is especially cumbersome for people that are trying out React Hooks for the first time. That's because a context provider that provides a setter would already create a rerender loop.

I've created a CodeSandbox to illustrate the use case here:
[![Edit react-issue-15924](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/react-issue-15924-yz8uq?fontsize=14)

I've tried to understand how the eslint rule works, but I'm not familiar with the eslint visitors so I've no clue where to look in the [long file](https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js).

miraage:
It is not a bug. It is the only thing that is expected. 
You are actually using "wrapper" variable, so it must be added to the deps list.

miraage:
> What is the expected behavior?
There should be no warnings in either case.

Should be warnings in both cases. You are actually using `wrapper` variable, so it must be added to deps list.

marceloadsj:
@miraage I'm not sure about it:

![image](https://user-images.githubusercontent.com/4874089/60454798-1c174e80-9c35-11e9-953a-fc230787f55f.png)

This happen with functions, but not with values.

miraage:
@marceloadsj eslint rule is seeking for variables used. But I totally agree with idea that tools should help people by covering common patterns, e.g. this case. 
[...]

Issue #15448, opened 2019-04-18
[react-hooks/exhaustive-deps] exhaustive-deps does not play well with object returning hooks


dantman:
**Do you want to request a *feature* or report a *bug*?** Bug

**What is the current behavior?**

exhaustive-deps does not play well with 3rd party hooks like useBoolean that return objects instead of arrays because they work better when not destructured.

For example in the following:
```js
import { useCallback } from 'react';
import { useBoolean, useInput } from 'react-hanger';
// ...
  const pending = useBoolean(false);
  const email = useInput();
  const submit = useCallback(() => {
    pending.setTrue();
    Promise.resolve()
      .then(() => console.log(email.value))
      .then(pending.setFalse);
  }, [email.value, pending.setTrue, pending.setFalse]);
```

exhaustive-deps complains with `React Hook useCallback has a missing dependency: 'pending'. Either include it or remove the dependency array`.

However this is a bad idea, since `pending` will change every time that `pending.value` changes resulting in the `submit` event handler changing each time. Despite the fact that it only uses `pending.setFalse` and `pending.setTrue` which are static and already included in the deps array. Destructuring `pending` would also 

**What is the expected behavior?**

exhaustive-deps should not complain because the exact deps are already present in the deps array.

universse:
Have you tried destructuring pending?
```
const { setTrue, setFalse } = useBoolean(false)

const submit = useCallback(() => {
    setTrue();
    Promise.resolve()
      .then(() => console.log(email.value))
      .then(setFalse);
  }, [email.value, setTrue, setFalse]);
```
From my experience it should work

dantman:
Oh, I forgot to finish my sentence.

Destructuring would defeat the simplicity of the hook. I have trimmed down the example, the real world usage this is from is a big-ish component. So you most definitely would not be using the `setTrue`/`setFalse` without any context of what they are actually for. The real destructuring would look more like.

```js
import { useCallback } from 'react';
import { useBoolean, useInput } from 'react-hanger';
// ...
  const {value: pending, setTrue: setPendingTrue, setFalse: setPendingFalse} = useBoolean(false);
  const email = useInput();
  const submit = useCallback(() => {
    setPendingTrue();
    Promise.resolve()
      .then(() => console.log(email.value))
      .then(setPendingFalse);
  }, [email.value, setPendingTrue, setPendingFalse]);

  // a pile of code
    {pending && <Loading />}
  // more code
```

universse:
[...]

PR #16040, opened 2019-07-02
[ESLint] Include functions instead of the objects they are part of when the result is unused

bartlangelaan:
This changes how `eslint-plugin-react-hooks/exhaustive-deps` works.

Take this component as an example:
```js
function App() {
  const myContext = useMyContext();
  useEffect(() => {
    myContext.setValue(e => e + 1);
  }, [myContext.setValue]);
  return null;
}
```

The exhaustive deps rule compains with the warning that `myContext` is missing from the deps array. However, because calling the function changes stuff in `myContext` adding it to the deps array would result in an endless re-render loop.

Please see #15924 for more information. For an interactive version, please see this CodeSandbox:
[Edit react-issue-15924](https://codesandbox.io/s/react-issue-15924-yz8uq?fontsize=14)

As long as we don't use the result of a function, it's fine to include the function itself to the deps array instead of including the object it comes from.

If anything should work differently as how I implemented it, please let me know and I'll try to adjust my code.

This solves issues #15924 and #15448.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -1024,6 +1024,18 @@ const tests = {
         }
       `,
     },
+    {
+      code: `
+        function MyComponent(props) {
+          const [skillsCount] = useState();
+          useEffect(() => {
+            if (skillsCount === 0 && !props.isEditMode) {
+              props.toggleEditMode();
+            }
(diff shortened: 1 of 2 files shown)


gaearon:
This is not correct. When you call `myContext.setValue`, you are technically passing `myContext` as `this` to `setValue`. So if it's not a dependency, you risk passing a stale value which might contain stale data. That leads to very confusing bugs.

If you don't intend to pass `this`, you can destructure it early:

```js
function App() {
  const {setValue} = useMyContext();
  useEffect(() => {
    setValue(e => e + 1);
  }, [setValue]);
  return null;
}
```

PR #16041, opened 2019-07-02
[fail] Only warn on unacted effects for strict / non sync modes

threepointone:
(basically, when `fiber.mode !== 0b0000`)

Warnings on unacted effects may be too noisy, especially for legacy apps. This PR fires the warning only when in a non sync mode (concurrent/batched), or when in strict mode. This should make gradually updating codebases easier too.

I also added batched mode tests to the act() suite.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
+++ b/packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
@@ -48,6 +48,20 @@ describe('ReactTestUtils.act()', () => {
     ReactDOM.unmountComponentAtNode(dom);
   }
   runActTests('legacy sync mode', renderSync, unmountSync);
+
+  // and then in batched mode
+  let batchedRoot;
+  function renderBatched(el, dom) {
+    batchedRoot = ReactDOM.unstable_createSyncRoot(dom);
+    batchedRoot.render(el);
+  }
+  function unmountBatched(dom) {
+    if (batchedRoot !== null) {
+      batchedRoot.unmount();
+      batchedRoot = null;
+    }
+  }
+  runActTests('batched mode', renderBatched, unmountBatched);
 });
 
 function runActTests(label, render, unmount) {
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2453,6 +2453,7 @@ export function warnIfNotCurrentlyActingEffectsInDEV(fiber: Fiber): void {
   if (__DEV__) {
     if (
       warnsIfNotActing === true &&
+      fiber.mode &&
```
acdlite: Do an explicit comparison for the modes we care about, please, so it doesn't accidentally break if we add another mode. Also makes it greppable.

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -68,6 +82,26 @@ function runActTests(label, render, unmount) {
...
+        function App() {
+          React.useEffect(() => {}, []);
+          return null;
+        }
+        expect(() => {
+          render(<App />, container);
+          // flush all queued work
+          Scheduler.unstable_flushAll();
+        }).toWarnDev(
+          label !== 'legacy sync mode'
```
acdlite: :headdesk: This is why I don't like abstractions in tests

How about we write a separate test for each mode? :D

packages/react-reconciler/src/ReactFiberWorkLoop.js
threepointone: the condition is for anything but sync mode (including future modes?). but yeah, maybe better to be obvious.

acdlite: Probably, but can't say for sure which modes we might add the future. Maybe the next one won't even be scheduling related.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2453,6 +2454,9 @@ export function warnIfNotCurrentlyActingEffectsInDEV(fiber: Fiber): void {
   if (__DEV__) {
     if (
       warnsIfNotActing === true &&
+      (fiber.mode & StrictMode ||
```
threepointone: should this fire for profiling mode as well? I... guess? will follow up as a separate PR if so.

threepointone: actually I'll get it in here right now, nvm

acdlite: It shouldn't fire if only ProfileMode is set, so this isn't right. It should only fire if either ConcurrentMode or BatchedMode or StrictMode are set. The convention we usually use for bitmask checks is:

```js
if ((mode & (ConcurrentMode | BatchedMode | StrictMode)) !== NoMode) {
  // ...
}

```

ConcurrentMode and BatchedMode are always accompanied by StrictMode is not. So technically it would be sufficient to check only for StrictMode.

```js
if ((mode & StrictMode) !== NoMode) {
  // ...
}
```

PR #16046, opened 2019-07-03
[Fire] Remove unused React fire fork

trueadm:
This PR removes the React Fire host config/reconciler fork of ReactDOM. The code isn't being used anywhere and has to be maintained every time we change the ReactDOM host config or object.


eps1lon:
New experimental react project called "Rain"? Since those usually start with "F" it seems like a new generation starting with "R". "R" as in "Rewrite in Rust"? :thinking:

PR #16048, opened 2019-07-03
Adds experimental fundamental interface

trueadm:
Note: this is a completely private API for internal experimentation use only.


trueadm:
The git history is messed up, closing and reopening another.

PR #16060, opened 2019-07-04
[Fresh] Prevent local variables from being used in the hook signature

theKashey:
The current implementation of the babel plugin would move any `use*` to the hook signature, except the "builtin" hooks.
That could lead to a false positive, when something, which _ducks_ like hook could not be used in the signature.

```js
const Component = ({useTheForce}) => {
  const useInnerHook = () => {useTheForce();};

  useInnerHook();
}
```

### Solution
Check that variable is reachable from the `signature` scope. That would not let track the "real" hook composition, but, at least, would not lead to runtime error.

### DIFF
--- a/packages/react-refresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-refresh/src/ReactFreshBabelPlugin.js
@@ -236,7 +236,7 @@ export default function(babel) {
     }
   }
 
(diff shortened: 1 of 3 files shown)


theKashey:
The required code was already implemented, just in a bit different form.

PR #16065, opened 2019-07-04
[Flare] Rename target/root events on event responders

trueadm:
With the very likely addition of first-class event targets, the existing naming of `targetEventTypes` and `rootEventTypes` becomes a bit confusing. This PR renames them to include `host`, which is exactly what they are – host events that we register on the specific platform. Event "types" was also a bit confusing too, as originally they were only strings, but then became objects where you could define if something was passive etc. This should hopefully make things easier to understand.


trueadm:
Actually I'm not sure about this fully fits with the event targets idea, as you'd actually have a different set of host events you listen to for each event target. Closing.

PR #16072, opened 2019-07-06
fix a puzzle variable name. rootSibling => rootChild.

gxvv:
The origin variable is temporary for storing `container.lastChild`. So **Child** is more accurate than **Sibling**.


Darshankatthige:
Guys someone hacked my friend Facebook id anyone here to resolve that issue

Please let me know I'll be waiting for the your reply please

PR #16090, opened 2019-07-09
Update comments

pixcai:
See files changed


threepointone:
I can see why this would confuse. The exisiting version is specifically done so fb tooling doesn't consider wrappers.js to be a generated file. Closing this.

PR #16094, opened 2019-07-10
Add warning when single item or nested arrays are used with SuspenseList

sebmarkbage:
SuspenseList is a new type of component because its children have "row" semantics. Each child component or fragment in the children array is its own row. Regardless if that component renders a fragment. Fragments or nested arrays are also considered a whole row for the whole list.

Another confusing thing is that a single slot in JSX passes that straight through but multiple slots are wrapped in an array, causing nested arrays.

Effectively this means that there's only really two legit ways to specify children to SuspenseList:

Static Children:

```js
<SuspenseList revealOrder="forwards">
  <A />
  <B />
  <C />
</SuspenseList>
```

Dynamic Children:

```js
<SuspenseList revealOrder="forwards">
  {items.map(item => <Item key={item.key} />)}
</SuspenseList>
```

A combination is not legit because it's confusing (this would be two rows, not N+1):

```js
<SuspenseList revealOrder="forwards">
  {items.map(item => <Item key={item.key} />)}
  <A />
</SuspenseList>
```

Therefore this adds a warning for that case.

This case is also confusing because we special case keyless single fragments to get unwrapped. So this actually will not be a single row but N rows.

```js
<SuspenseList revealOrder="forwards">
  <>{items.map(item => <Item key={item.key} />)}</>
</SuspenseList>
```

Passing a single static item also isn't really ever useful with the "forwards" option since there's nothing to coordinate. You can just use the "together" mode then.

I added a warning if we see any single item since that seems like a mistake.

Perhaps we should even fork child fiber to special case this type of reconciliation but that would cause more code so many a warning is sufficient for now.

Another quirky thing is that the "together" mode doesn't really have the same quirks. It would be reasonable even to render a single component that has many boundaries inside of it.

So arguably this is legit so I disable the warning for this mode:

```
<SuspenseList revealOrder="together">
  <Component />
</SuspenseList>
```

Where the Component contains something like:

```
<div>
  <header><Suspense>...</Suspense></header>
  <footer><Suspense>...</Suspense></footer>
</div>
```

This is unfortunate because now it gets trickier to enforce the types of children in a type system since it now depends on the revealOrder flag. That's what I get for putting the notion of a Suspense "group" and "list" in the same component. In the type system, we could say that even this case should include no nested arrays just for consistency. That's probably ok.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -67,7 +67,7 @@ import shallowEqual from 'shared/shallowEqual';
 import getComponentName from 'shared/getComponentName';
 import ReactStrictModeWarnings from './ReactStrictModeWarnings';
 import {refineResolvedLazyComponent} from 'shared/ReactLazyComponent';
-import {REACT_LAZY_TYPE} from 'shared/ReactSymbols';
+import {REACT_LAZY_TYPE, getIteratorFn} from 'shared/ReactSymbols';
 import warning from 'shared/warning';
 import warningWithoutStack from 'shared/warningWithoutStack';
 import {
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2094,6 +2094,69 @@ function validateTailOptions(
...
 }
 
+function validateSuspenseListNestedChild(childSlot: mixed) {
+  if (__DEV__) {
+    let isArray = Array.isArray(childSlot);
+    let isIterable = !isArray && typeof getIteratorFn(childSlot) === 'function';
+    if (isArray || isIterable) {
+      warning(
+        false,
+        'A nested %s was passed to <SuspenseList />. This whole array ' +
```
gaearon: “This whole %s”?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2094,6 +2094,69 @@ function validateTailOptions(
...
+          'will be treated as a single row in the list. If this is what you ' +
+          'want, wrap it in a fragment e.g. <>{array}</>. If you meant to ' +
+          'merge multiple arrays to a single list, try using React.Children.toArray(). ' +
```
gaearon: Maybe `toArray(array)`. It took me a bit to realize what you’re suggesting to do here.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2094,6 +2094,69 @@ function validateTailOptions(
...
+function validateSuspenseListNestedChild(childSlot: mixed) {
+  if (__DEV__) {
+    let isArray = Array.isArray(childSlot);
+    let isIterable = !isArray && typeof getIteratorFn(childSlot) === 'function';
+    if (isArray || isIterable) {
+      warning(
+        false,
+        'A nested %s was passed to <SuspenseList />. This whole array ' +
+          'will be treated as a single row in the list. If this is what you ' +
+          'want, wrap it in a fragment e.g. <>{array}</>. If you meant to ' +
```
gaearon: Comma before eg?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2094,6 +2094,69 @@ function validateTailOptions(
...
+  revealOrder: SuspenseListRevealOrder,
+) {
+  if (__DEV__) {
+    if (
+      (revealOrder === 'forwards' || revealOrder === 'backwards') &&
+      (children !== undefined && children !== null && children !== false)
+    ) {
+      if (Array.isArray(children)) {
+        for (let i = 0; i < children.length; i++) {
+          if (!validateSuspenseListNestedChild(children[i])) {
```
gaearon: Should we include index in the message? If you’re accidentally including an array (such as if some helper like `renderFoo` returns it under some conditions) you might find it difficult to figure out which item is causing it.

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: It would be more like `toArray([firstArray, secondArray])` but it's kind of hard to give a precise recommendation without knowing how this happened.

sebmarkbage:
Should we just drop the recommendation here and make it a link instead?

sebmarkbage:
I simplified the message by always suggesting the nested SuspenseList solution to simplify the recommendation. That lets you configure if you want it to be one by one or all together explicitly. It's also the recommendation for things like grids that are nested arrays. This neatly avoids the need for `toArray` too. The fragment wrapper is equivalent to nested SuspenseList with no revealOrder specified but it's kind of an unusual edge case.

PR #16095, opened 2019-07-10
support callback for batchedUpdates

yiminghe:
In some scenarios, there are nested batchedUpdates, and each update requires notification of render completion.

```js
batchedUpdates(()=>{

// .... long code

batchedUpdates(()=>{
  // some setState({});
});

// render is not completed because of batch

});
```

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -344,6 +344,7 @@ export function computeExpirationForFiber(
 }
 
 let lastUniqueAsyncExpiration = NoWork;
+
 export function computeUniqueAsyncExpiration(): ExpirationTime {
   const currentTime = requestCurrentTime();
   let result = computeAsyncExpiration(currentTime);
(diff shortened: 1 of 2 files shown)


threepointone:
Could you explain the need for this?

yiminghe:
First we create an custom render:

```js
const CustomReconciler = Reconciler(HostConfig);
var Render = {
  create(element) {
    const container = CustomReconciler.createContainer(++id, false, false);
    const entry = {
      update(newElement) {
        CustomReconciler.updateContainer(newElement, container, null, null);
      },
      unmount() {
        CustomReconciler.updateContainer(null, container, null);
      },
    };
    entry.update(element);
    return entry;
  },
  batchedUpdates(...args) {
    return CustomReconciler.batchedUpdates(...args);
  },
};
```

Then we can create an entry:

```js
const entry1 = Render.create(...);
function updateEntry(entry, newElement, fn) {
  Render.batchedUpdates(()=>{
    entry.update(newElement);
    fn();
  });
 // entry1 render is done, process logic
}

```

But when there are two entries, batchedUpdates may be nested


```
updateEntry(entry1, newElement1,()=>{
updateEntry(entry2, newElement2,()=>{
});
```

gaearon:
In the future batched updates will be default, so an explicit API to wait for their completion doesn't make sense to me. Thank you for the PR though!

PR #16100, opened 2019-07-10
[Flare] Press includes button type

necolas:
1. Allow auxillary button clicks (i.e., middle mouse button) to trigger 'onPressStart' and 'onPressEnd', but never 'onPress'.
2. Report the button type – 'primary' or 'auxillary' – on the press event.

### DIFF
--- a/packages/react-events/src/dom/Press.js
+++ b/packages/react-events/src/dom/Press.js
@@ -89,6 +89,7 @@ type PressEventType =
   | 'contextmenu';
 
 type PressEvent = {|
+  button: 'primary' | 'auxillary',
   defaultPrevented: boolean,
   target: Element | Document,
   type: PressEventType,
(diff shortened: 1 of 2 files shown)


packages/react-events/src/dom/Press.js
```diff
@@ -706,11 +712,11 @@ const PressResponder: ReactDOMEventResponder = {
             state.activePointerId = touchEvent.identifier;
           }
 
-          // Ignore any device buttons except left-mouse and touch/pen contact.
-          // Additionally we ignore left-mouse + ctrl-key with Macs as that
+          // Ignore any device buttons except primary/secondary and touch/pen contact.
```
necolas: s/secondary/auxillary

packages/react-events/src/dom/Press.js
```diff
@@ -962,7 +968,10 @@ const PressResponder: ReactDOMEventResponder = {
                 state,
               );
             }
-            if (state.isPressWithinResponderRegion) {
+            if (
+              state.isPressWithinResponderRegion &&
+              nativeEvent.button !== 1
```
trueadm: Can we hoist `button` out and re-use it for both conditions?

PR #16101, opened 2019-07-10
[Flare] Migrate to `useResponder` event system design 

trueadm:
This PR migrates away from the existing Flare event system design, to a slightly different design and API. Notably, there are now no Event Components, instead as the first of this new design there are only Event Responders. The only way to use an Event Responder is by using its hook form inside a functional component that you want to "tie" to becoming an event target for the responder in question.

Before you would do:

```jsx
  <Press onPress={onPress}>
    <Hover onHover={onHover}>
      <Focus onFocus={onFocus}>
        <div>
          {children}
        </div>
      </Focus>
    </Hover>
  </Press>
}
```

With this PR, you now do:

```jsx
  const targetRef= useRef(null);
  usePressResponder(targetRef, {onPress});
  useHoverResponder(targetRef, {onHover});
  useFocusResponder(targetRef, {onFocus});
  return <div ref={targetRef}>{children}</div>
}
```

## The reasons for this change:
- There is now an explicit contract between an event responder at the function component that owns it (using the hook design). This creates a boundary for any given responders inside the boundary, so they don't bubble up the fiber tree.
- There is also an explicit contract between a responder and a host target. Before, there was no clear way to provide a 1:1 contract.
- We create less fibers and objects with this model, whilst in the case before, each event component would have an associated fiber for each one in the tree.

## Issues with Event Components:

It's difficult to compose event components without falling into pitfalls. The majority if event components have a 1:1 mapping to a DOM node, meaning they need a static child that remains constant through the lifecycle of the event responder. This is easy to visualize in this case:

```jsx
<Focus>
  <button>Focus me!</button>
</Focus>
```

It's harder to track and warn about (without overhead) in cases where the node might resolve conditionally:

```jsx
<Focus>
  <SomeComponent />
</Focus>

function SomeComponent() {
  return condition ? <button>Focus me!</button> : <span>Focus me!</span>
}
```

It's even harder when the child might suspend or lazyily change. Not to mention, there's not a consistent story with event components. `<FocusScope>` might want to accept many children, yet is still an event component, whilst `<Focus>` clearly does not.

This updated design attempts to tackle some of those concerns.

## Note:

It's important to remember that this is still experimental and may change at any time. We expect `useResponder` to be used sparingly – almost as if it as private API – to be used in core primitive componets that are the building blocks for an app. In later PRs, we will expose `createListener` and `useListener` which will be used as more of a public facing API to be used in user-space components.


TrySound:
What about useEvent? How it looks now?

trueadm:
useEvent used event components - which this design does not. There are plans to have event listeners as part 2 of the design to come (if we go ahead with this new design) - which have both a hook and a component form. The main difference is they listen to Flare events (dispatched from event responders) rather than DOM events.

trueadm:
After some internal discussion, I've got come up with a slightly tweaked design change that warrants a new PR. I'll put this up tomorrow :)

PR #16104, opened 2019-07-10
Give unresolved lazy() a name in component stack

motiz88:
Currently, if a `React.lazy(...)` component fails to load, React reports the error with a component stack in which the lazy component is represented as `Unknown` ([example](https://codesandbox.io/s/unresolved-lazy-component-stack-qjsb3)).

With this PR, the name reported will instead be `Lazy` while the lazy component is unresolved. This PR also adds a test covering the (unchanged) case of a *resolved* lazy component that throws an error during rendering.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
@@ -1183,4 +1183,85 @@ describe('ReactLazy', () => {
       expect(Scheduler).toFlushAndYield([]);
     }).toErrorDev('Function components cannot be given refs');
   });
+
+  it('should error with a component stack naming the resolved component', async () => {
+    let componentStackMessage;
+
+    const LazyText = lazy(() =>
+      fakeImport(function ResolvedText() {
+        throw new Error('oh no');
+      }),
+    );
+
+    class ErrorBoundary extends React.Component {
+      state = {error: null};
+
+      componentDidCatch(error, errMessage) {
+        componentStackMessage = errMessage.componentStack;
+        this.setState({
+          error,
+        });
+      }
+
+      render() {
+        return this.state.error ? null : this.props.children;
+      }
+    }
+
+    ReactTestRenderer.create(
+      <ErrorBoundary>
+        <Suspense fallback={<Text text="Loading..." />}>
+          <LazyText text="Hi" />
+        </Suspense>
+      </ErrorBoundary>,
+      {unstable_isConcurrent: true},
+    );
+
+    expect(Scheduler).toFlushAndYield(['Loading...']);
+
+    try {
+      await Promise.resolve();
+    } catch (e) {}
+
+    expect(Scheduler).toFlushAndYield([]);
+
+    expect(componentStackMessage).toContain('in ResolvedText');
+  });
+
+  it('should error with a component stack containing Lazy if unresolved', () => {
+    let componentStackMessage;
+
+    const LazyText = lazy(() => ({
+      then(resolve, reject) {
+        reject(new Error('oh no'));
+      },
+    }));
+
+    class ErrorBoundary extends React.Component {
+      state = {error: null};
+
+      componentDidCatch(error, errMessage) {
+        componentStackMessage = errMessage.componentStack;
+        this.setState({
+          error,
+        });
+      }
+
+      render() {
+        return this.state.error ? null : this.props.children;
+      }
+    }
+
+    ReactTestRenderer.create(
+      <ErrorBoundary>
+        <Suspense fallback={<Text text="Loading..." />}>
+          <LazyText text="Hi" />
+        </Suspense>
+      </ErrorBoundary>,
+    );
+
+    expect(Scheduler).toHaveYielded([]);
+
+    expect(componentStackMessage).toContain('in Lazy');
+  });
 });
--- a/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
@@ -6,6 +6,15 @@ let ReactFeatureFlags;
 let Suspense;
 let lazy;
 
+function normalizeCodeLocInfo(str) {
+  return (
+    str &&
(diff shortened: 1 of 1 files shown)


motiz88:
cc @bvaughn who's been working on component stacks recently.

packages/react-reconciler/src/ReactCurrentFiber.js
```diff
@@ -37,7 +37,9 @@ function describeFiber(fiber: Fiber): string {
     default:
       const owner = fiber._debugOwner;
       const source = fiber._debugSource;
-      const name = getComponentName(fiber.type);
+      const name = fiber.type
+        ? getComponentName(fiber.type)
+        : getComponentName(fiber.elementType);
```
bvaughn: Hm... I think this change wouldn't do what your PR description says:

> The `displayName` support may be leveraged in a Babel plugin (see https://github.com/babel/babel/pull/10201) to infer the name from context and make this seamless for most users:

```diff
 const Foo = React.lazy(() => import('./FooImplementation'));
+Foo.displayName = 'Foo';
```

It *would* display "Lazy" rather than "Unknown", which would be an improvement, but to actually honor the `displayName` (if we wanted to) would also require a change to `getComponentName`)

packages/shared/getComponentName.js
```diff
@@ -90,7 +90,7 @@ function getComponentName(type: mixed): string | null {
         if (resolvedThenable) {
           return getComponentName(resolvedThenable);
         }
-        break;
+        return 'Lazy';
```
bvaughn: I think to fully support the `displayName` attribute you describe, we'd want to do this:
```js
case REACT_LAZY_TYPE: {
  const thenable: LazyComponent<mixed> = (type: any);
  const resolvedThenable = refineResolvedLazyComponent(thenable);
  if (resolvedThenable) {
    return getComponentName(resolvedThenable);
  } else {
    return thenable.displayName || 'Lazy';
  }
}
```

packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
```diff
@@ -1089,4 +1089,85 @@ describe('ReactLazy', () => {
...
+        <Suspense fallback={<Text text="Loading..." />}>
+          <LazyText text="Hi" />
+        </Suspense>
+      </ErrorBoundary>,
+    );
+
+    expect(Scheduler).toHaveYielded([]);
+
+    expect(componentStackMessage).toContain('in Lazy');
+  });
```
bvaughn: You'd probably want to add an additional test case for the named lazy:
```js
it('should error with a component stack containing Lazy with displayName if unresolved', () => {
  let componentStackMessage;

  const LazyText = lazy(() => ({
    then(resolve, reject) {
      reject(new Error('oh no'));
    },
  }));
  LazyText.displayName = 'LazyText';

  class ErrorBoundary extends React.Component {
    state = {error: null};

    componentDidCatch(error, errMessage) {
      componentStackMessage = errMessage.componentStack;
      this.setState({
        error,
      });
    }

    render() {
      return this.state.error ? null : this.props.children;
    }
  }

  ReactTestRenderer.create(
    <ErrorBoundary>
      <Suspense fallback={<Text text="Loading..." />}>
        <LazyText text="Hi" />
      </Suspense>
    </ErrorBoundary>,
  );

  expect(Scheduler).toHaveYielded([]);

  expect(componentStackMessage).toContain('in LazyText');
});
```

motiz88:
@bvaughn Oops, the description is stale. I did implement this with `displayName` support initially (45458ad) and then IIRC had an off-GitHub conversation with @gaearon where he fed back that we shouldn't add `displayName`.

bvaughn:
Oh I see. In that case then I think this looks reasonable.

gaearon:
Can we update the PR description to match what this currently does? I'm a bit confused because I don't remember it by now.

I think I was hesitant to support `displayName` for the same reason as it seems a shady pattern in general. Patching it on can mess with the hidden types. Someone will undoubtedly write a Babel plugin that adds it in production. Etc. I guess we already support that for `memo` and such so maybe this is not a strong argument.

motiz88:
Updated the description, sorry for the confusion.

For my part, I will happily bring back `displayName` support if you folks view it as acceptable. Personally my take on it is:
1. it solves a real (if niche) problem;
2. it doesn't meaningfully increase the size/complexity of React;
3. the cost to applications that don't use `displayName` is zero (or next to zero, see 2);
4. the size cost to applications that *do* use `displayName` is proportional to the number of lazy components they apply it to, which is the same deal as for `displayName` in general;
5. perf might not be a problem if `displayName` is applied uniformly (e.g. via a Babel plugin) - my understanding of hidden classes is that adding properties in a predictable order is a fast path;
6. until such a time as `displayName` is officially deprecated and something better is available to replace it, it makes sense to support it on all types of components.

The counterargument could be made that dropping `displayName` support on newer APIs _is_ a way of soft-deprecating it because it's too expensive to begin with. But a good alternative to it doesn't exist yet, so I personally think taking it away is premature.

gaearon:
Is this still a problem on master?

motiz88:
@gaearon Yes. I've just rebased and pushed an update.

packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
gaearon: @motiz88 Wanna do this?

motiz88: By "this" do you mean adding back `displayName` support? Yes, I'd be happy to.

gaearon: Hmm I misread the test. I thought @bvaughn meant adding a test that verifies a *resolved* `Lazy` shows the inner component name in the stack.

gaearon:
Let's add `displayName` support back. I was probably being unreasonable.

sebmarkbage:
I already fixed this for stacks in the other refactor so I took the liberty to rebase this.

I didn't do it for other uses of Component name since it's not clear what it should be. Especially if it's not that it's unresolved but that it errors when extracting the name. I'll land this new regression test though.

PR #16105, opened 2019-07-10
[Scheduler] Support inferring priority from stack

acdlite:
When executing a task, wraps the callback in an extra function whose name includes the current priority level. Profiling tools with access to function names can use this to determine the priority of the task.


.watchmanconfig
```diff
@@ -0,0 +1,2 @@
+{
+}
```
sebmarkbage: 👀

acdlite: Don't judge me

PR #16114, opened 2019-07-11
[Flare] Press: fix middle-click handling

necolas:
Make sure the root events are removed after middle-click completes.

(Wasn't able to create a failing unit test. Another reason to eventually move to browser functional tests.)


trueadm:
For a regression test you should be able to do pointerdown, pointerup, pointermove - all with button 1, clientX,clientY and mouse pointerType. At least that’s what the case was earlier, pressstart should incorrectly fail.

necolas:
Ah got it, needed to move out then back in

PR #16115, opened 2019-07-11
Clean up nextEffect pointer

paulshen:
This change clears the `nextEffect` reference when it is no longer needed. If there are no passive effects, we iterate through the effect chain after committing host/layout effects. Otherwise, we remove the pointer while flushing passive effects. Pulled from #15157

Without this change, it is possible for live Fiber nodes to have `nextEffect` pointers to unmounted Fiber nodes. 

## Repro

https://gist.github.com/paulshen/8a10c1ac7cd1b2983219e27a08d0f85b
Click "Increment" many times and notice that all unmounted `<InnerBody>` Fiber nodes are retained via a `nextEffect` chain from `<Sidebar>`'s Fiber node.

image

With this change, `<Sidebar>`'s Fiber node has an empty `nextEffect` reference.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -1712,6 +1712,16 @@ function commitRootImpl(root) {
     rootDoesHavePassiveEffects = false;
     rootWithPendingPassiveEffects = root;
     pendingPassiveEffectsExpirationTime = expirationTime;
+  } else {
+    // We are done with the effect chain at this point so let's clear the
+    // nextEffect pointers to assist with GC. If we have passive effects, we'll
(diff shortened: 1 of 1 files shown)


sebmarkbage:
Thanks @paulshen! Sorry about the long back and forth.

gaearon:
We've published `0.0.0-db3ae32b8` of all packages, which includes this fix. It is likely to be a release candidate for `16.9.0`. You can give it a try now and see whether it resolves the issue for you. Thanks.

jonnycornwell:
:+1: Thanks guys, finally got around to retesting and this definitely fixes the leak in the [gatsby starter](https://github.com/jonnycornwell/potential_react_leak) I made

gaearon:
16.9 is out too.

Issue #16009, opened 2019-06-27
[Flare] Rethinking Focus


trueadm:
I think we need to rethink how focus works in React. React Flare is the perfection opportunity to allow us to do this, so here are some of my thoughts. None of these ideas wouldn't be possible if it weren't for the great ideas from @sebmarkbage, @devongovett and @necolas have had. Furthermore, the discussions in https://github.com/facebook/react/pull/16000, https://github.com/facebook/react/pull/15848 and https://github.com/facebook/react/pull/15849 got me thinking on a better system.

## Focus is a mess on the DOM, so let's not use the DOM

Focusing on the DOM is a mess today. We couple ideas around ideas around things like `tabIndex` and whether a specific browser treats something as focusable. This is very much a hard-coded disaster where no one really agrees on a good formula for success. Not to mention, that this just doesn't translate well for a declarative UI. How does one tab to a specific node that isn't focusable? How does one use keyboard arrows to navigate a table using keyboard arrows?

Then there's implementation. Without relying on an attribute on an element or a `ref`, it's very hard to say: "Hey look, let's focus to this node, given this criteria". Not to mention the performance overhead of doing this: querying or "collecting" focusable elements is an expensive O(n) task, which doesn't scale for large applications well. I noticed that wrapping the an internal large app with `<FocusScope>` and then collecting all focusable nodes took over 850ms on Android using Chrome. Querying the DOM nodes took even longer.

Lastly, we can't use the DOM with React Native and the story for handling focus with React Flare is important. If we instead had a React system for handling focus, then both the web and RN would be consistent and performant.

## Accessible components

We already have the `<Focus>` and `<FocusScope>` event components. We could extend on React Flare and introduce a way of layering accessibility logic on to host components. In this I introduce a new API called `createAccessibleComponent`, but really it could be anything – ignore the naming! This is purely hypothetical discussion for now.

```jsx
// input is not focusable
<FocusScope>
  <input type="text" placeholder="Enter your username" /> 
</FocusScope>

const FocusableInput = ReactDOM.createAccessibleComponent((props, focusable) => {
  return <input tabIndex={focusable ? 0 : -1} {...props} />;
});

// now it's focusable
<FocusScope>
  <FocusableInput
    type="text"
    placeholder="Enter your username"
    focusable={true}
  />
</FocusScope>
```

If you don't use a `FocusScope`, then the normal DOM behaviour will continue to work as expected. `FocusScope` will only care about these new types of accessible component.

[...]

PR #16122, opened 2019-07-12
[Flare] Migrate from EventComponent to EventResponder

trueadm:
This is a follow up to https://github.com/facebook/react/pull/16101. In that PR, we introduced `useResponder`. After some careful consideration, this PR revises that design slightly so Event Responders are their own component that get used in the render tree rather than the `useResponder` hook.

This PR also removes `React.unstable_createEvent` and `React.unstable.useEvent`. We expect to add something similar to those APIs in subsequent follow up PRs.

Before you would do:

```jsx
  return (
    <Press onPress={onPress}>
      <Hover onHover={onHover}>
        <Focus onFocus={onFocus}>
          <div>
            {children}
          </div>
        </Focus>
      </Hover>
    </Press>
  )
}
```

With this PR and design, you would do this:

```jsx
  return (
    <div>
      <PressResponder onPress={onPress} />
      <HoverResponder onHover={onHover} />
      <FocusResponder onFocus={onFocus} />
      {children}
    </div>
  )
}
```

Press Responders are created in a similar way to how Event Components were created before:

```jsx
export PressResponder = React.unstable_createResponder(PressResponderImpl);
```

When it comes to the design, event responders "attach" themselves to their nearest host parent. So in the case of:

```jsx
<div>
  <React.Fragment>
    <FunctionWrapper>
      <EventResponder>
```
The `<EventResponder>` will attach itself to the `<div>`.

You can apply event responders to self-closing elements too (event responders do not created DOM nodes):

```jsx
<input><FocusResponder /></input>
```

## The reasons for this change:
- There is now an explicit contract between an event responder and a host component. This is the nearest host component parent to the event responder.
- We now resolve responders as a property of the thing we're interacting with, rather than wrapping around it. This provides an important constraint around how event responders are used, in that it makes extending an abstraction from the outside harder rather than easier.
- By making a responder a child of the target, we can handle conditional cases with ease (the event responder will unmount/mount).
- A target can only have a single type of event responder at a given time. If multiple event responders try to attach to the same target and they are the same event responder, only the first will register and the others will be no-ops (with dev warnings).
- Unlike the other design, using `ref`s is somewhat error prone. The same ref might be passed around and assign conditionally or multiple times, which would be difficult to detect and handle.


necolas:
How would you use this to mark specific parts of the host tree as the intended event target when a primitive is implemented via the "fundamental" API? With the existing model we can pass the responder components in as props even though that's not ideal.

These upward looking components could be used to create really confusing DX around events and styles, but we don't seem to have good alternative mechanisms in React today

trueadm:
For Fundamental, only the single instance gets associated to an actual fiber, so we could provide a context method to bind a responder inside the fundamental component implementation itself. You can also pass it through as children and if reconcile children is on, React will hook them up anyway.

By the existing API, you mean event components that we have today? Or did you mean the previous PR that had useResponder?

The inside-our design of these components is different for sure, but offer some interesting. Constraints because of their design that seems like one of the few really viable ways in React today. We can experiment with them internally for now, to see how they scale and get used - which is the only real practical indicator we have of their usefulness. I agree with your concerns though. @sebmarkbage might be able to explain how we might present them differently so they don’t look as awkward.

sebmarkbage:
I like this mechanism for some practical implementation reasons like that the way these connect are in a way that is essentially free since we deal with intermediates like this for DOM nodes anyway and works well with async creation. It also fits in neatly with React's reconciliation model in that if a host component unmounts, it'll by definition take its responder with it. So there's no confusion about what happens if a node changes in the middle of a sequence.

I also like it from some abstract composition notions. It's important for React to provide some form of encapsulation contract so that components can implement themselves how they want without something drilling through them. Once something is drilling through them, abstractions tend to become locked in place and implementation details are not safe to change. That tends to leak further out than the actual practical problems due to a fear that something *might* break. So you have to add things like defensive nodes. This is one of the problems with findDOMNode. EventComponent suffered from this as well because `<EventComponent><AbstractButton /></EventComponent>`. Now AbstractButton wasn't free to add multiple nodes or a fragment with a portal at the root. We used to think of this as a feature that you could extend components like this but there's nothing in here that says that it is the outer most node of the AbstractButton that is the best place to extend this component. This is similar to the issue with refs. We once considered having forwardRefs just pass  through to the inner most DOM node but we realized that many times the most suitable ref (for exposing things like focus) isn't always the outer most node. E.g. `let Input = () => <div><input /></div>`. We do have drill-through mechanisms like Context but those are opt-in from the implementation's perspective.

This is a good first step to build on since at least the mechanism satisfies some useful properties for building on.

That said, React isn't all about abstract composition rules and implementation details. I realize that this at first comes off as a confusing API and that's something we must address.

I think there are a few possible paths here:

- One possible path is just to keep the same mechanism but play with syntax sugar and different way of writing it. I don't think we should do this just yet until we've experimented with the mechanism or other approaches more. But even just making them not JSX could be one way to give them special status.

```js
<div>
  {PressResponder({onPress})}
  {HoverResponder({onHover})}
  {FocusResponder({onFocus})}
  {children}
</div>
```

- One is that we actually later on find more use cases like this to "augment" nodes/view in the tree and make that a first-class concept. Maybe like decorators for JSX. We did this Hooks. Hooks originally started out as just components in the reconciliation tree along with other things. After a while it was clear that this was a valuable enough concept to lift out into its own first-class concept and now they're "augmentations", a separate type of branch, in the fiber/reconciliation tree. We did that while preserving the composition and abstraction ability (custom hooks). We could potentially do something similar here.

- Another possible path, is that we don't change the API but we change the mental model. Instead of making the implementation details first-class, we can make the concept first-class. For example if more things (like Formation) buys into this notion of extension, then it becomes part of a larger conceptual story which is easier to explain in a larger concept. Particularly, I think part of what makes this confusing in the implementation details. If you think about it as something that renders into a DOM node, it's confusing how that DOM node would be able to capture events from its surrounding. So you have to make the leap that it somehow attaches itself to its parent. However, if you think of this from a visual design tool (like the classic MS Paint), then you can think of this as a Fill tool. You drop it in and it fills out the bounds of its available space with something "Pressable". Personally, I like that mental model better. It doesn't have to be how we explain it but I think there might be a way where presentation and the way you get introduced to this might be enough. We did fairly well with Hooks and that's what I hope will help some of our even more counter intuitive ideas like Concurrent/Suspense.

sebmarkbage:
@trueadm You don't have to fix this in this first PR but one thing to be aware of is that not all DOM components currently accept children other than something that can be toStringed.

https://github.com/facebook/react/blob/48f6594474c449b3ed62fadedda0ffad5e3a807a/packages/react-reconciler/src/ReactFiberBeginWork.js#L986-L998

https://github.com/facebook/react/blob/a457e02ae3a2d3903fcf8748380b1cc293a2445e/packages/react-dom/src/client/ReactDOMHostConfig.js#L309-L311

https://github.com/facebook/react/blob/a457e02ae3a2d3903fcf8748380b1cc293a2445e/packages/react-dom/src/client/ReactDOMTextarea.js#L59

I think the only case this matters for is `textarea`. We'd have to build a mechanism to allow responders to render into a textarea children. So you can attach things like key press and focus to `textarea`. Shouldn't be too hard.

trueadm:
@sebmarkbage I've added a test and a way to properly handle `children` with text content and textarea.

necolas:
> By the existing API, you mean event components that we have today? 

@trueadm I meant the non-fundamental API where primitives are just build with jsx divs from React DOM. Adding to the context/manager API sounds good.

> But even just making them not JSX could be one way to give them special status.

@sebmarkbage I like that idea and it gives us a bit of space to explore alternative syntax in the future

trueadm:
Another idea I had, which is another spin on the current idea process is to have a new node that makes these bindings explicit, rather than coupling them into children. `createBinding()` expects to arguments, a host node and an array of responders/styles to bind to the node:

```jsx
<View>
  {
    React.createBinding(
      <View>{children}</View>, [
      <PressResponder/>,
      <HoverResponder/>,
      <FocusResponder/>,
    ])
  }
</View>
```

If you have styling, like with formation:

```jsx
const ViewWithStyling = ({children, styling}) => (
  React.createBinding(
    <View>{children}</View>, [
    <PressResponder/>,
    <HoverResponder/>,
    <FocusResponder/>,
    <Border />,
    <BackgroundColor />,
    ...styling,
  ])
)

<View>
  <ViewWithStyling />
</View>
```

You can still compose styles and events like before, except they now have a clear connection. @sebmarkbage  made the point about being able to compose these to affect a parent, where `AppPressEffect` added styling and events and applied them based on some state changes:

```jsx
<div>
  {children}
  <AppPressEffect />
</div>
```

You could do this with some discrete functions instead, which is arguably more the way things look and feel today:

```jsx
<AppPressEffect>
  <div>{children}</div>
</AppPressEffect>

const AppPressEffect = ({children}) => {
  const [isPressed, updatePressed] = useState(false);

  return (
    React.createBinding(
      children, [
        <PressResponder onPressChange={updatePressed} />,
        isPressed && <PressedStyling />
       ]
  );
};
```

devongovett:
I don't think these responders should be components at all. As @sebmarkbage said, they are really more like decorators for host components. With both the original API (wrapping) and this one, it is somewhat unclear which node the events are bound to, and you lose explicit control. With the original, you don't have control over which element it applies to since it is always the nearest inner host component (similar problem as `forwardRef` solves), and with the new API, it is the nearest outer host component which could be problematic if someone passed in unknown children. Where children are forwarded and where events should be forwarded could be different places, for example children to an inner child and events to an outer one.

I liked the general idea from your original PR for this (#16101). Perhaps refs aren't the right API, but at least they are exactly explicit about where they go since you pass them as a prop to the host component directly. And you can already forward them to the correct host node on components with `forwardRef`.

It seems like the problem with refs (generally, not just for this use case) is that you can only have one per component. This makes it challenging to compose multiple unknown discrete pieces of functionality, i.e. a forwarded ref and an event responder. Perhaps host components could accept a new prop similar to refs but specific to event responders. This would allow refs to continue being passed for other purposes, while making event bindings explicit.

```jsx
  return (
    <div responders={[
      PressResponder({onPress}),
      HoverResponder({onHover}),
      FocusResponder({onFocus})
    ]}>
      {children}
    </div>
  );
}
```

This has several nice properties:

* It's explicit about which host component the responders apply to.
* It's separate from refs and children, which might need to be forwarded to different places than events.
* It's composable. You can spread additional responders passed in as props onto the right host nodes.
* It can be validated. You can verify that there are not multiple responders of the same type, though perhaps that could be allowed to allow easier composability of events rather than chaining functions?
* You can also verify that the same responder instance (returned from the functions) is not used on multiple host nodes if needed.

sebmarkbage:
The more I think about the more I think these *should* be components just like this PR.

That allows for

trueadm:
@devongovett Your last suggestion is very similar to my createBinding suggestion, in that they're explicitly defined. I opted not to use "responders" so so that you could bind other things other than events, such as styling (like Sebastian mentioned).

necolas:
I think this design still has some issues related to the implicit relationship to host nodes (which the current design also has, but in the opposite direction along the host node tree). I'll post something at work next week

trueadm:
I was thinking today: the real issue here is the implicit nature of the parent -> child. If we solve that problem, without compromising the ability to nicely compose things together, would that allow us to move forward?

One possible solution, is to explicitly define a host component's target for a given sub-tree. I was thinking a possible way to do this, was with a sort of render target props that is constrained to being only available for its children:

```jsx
React.createTarget(
  <div>
    {(target =>
      <EventResponder target={target} />
    )}
  </div>
);
```

However, it feels a bit cumbersome returning to a render prop pattern, plus it relies on having to be passed around a component tree via props. Another approach might be to use a more context-like API:

```jsx
const HeaderContainerTarget = React.createTarget('div');

function MyHeaderBar(props) {
  return (
    <HeaderContainerTarget {...props}>
      <div>
        <MyChild />
      </div>
    </HeaderContainerTarget>
  );
}

function MyChild() {
  return (
    <>
      <PressResponder target={HeaderContainerTarget} onPress={...} />
      <span>Hello world</span>
    </>
  );
}
```

A slight different alternative to the above approach based on hooks:

```jsx
const HeaderContainerTarget = React.createTarget('div');

function MyApp(props) {
  return (
    <HeaderContainerTarget {...props}>
      <div>
        <MyChild />
      </div>
    </HeaderContainerTarget>
  );
}

function MyChild() {
  usePressResponder(HeaderContainerTarget, pressProps);
  useHoverResponder(HeaderContainerTarget, hoverProps);
  return <span>Hello world</span>
}
```

devongovett:
Perhaps I can illustrate the concern in my previous comment with an example.

```jsx
function IconButton({children}) {
  return (
    <button>
      <Icon />
      <span class="button-label">{children}</span>
    </button>
  );
}
```

The problem here is that children are passed through as the button label into the `<span>`. If children contained any event responder components, then they would be attached to the `<span>` instead of the `<button>`. I think that's the issue with passing both through children. It's too hard to control where they are attached. If events were passed as some other prop, it would be easier.

devongovett:
Oops, just read your most recent comment more carefully @trueadm and I think you came to the same conclusion.

I think the `createTarget` idea you had might solve the issue so it's a little more explicit. But it still could be a bit confusing. You might still want to have children be forwarded to a different place than events. Or you might have multiple different targets inside your component where you want events to go. I still think separate props that you can pass through to whatever children you want would be the easiest and most explicit.

trueadm:
@devongovett My thinking was that you'd also pass the specific target through as props in those cases, so you're "opting" in for children to act on the target you're using, rather than children implciitly opting in parent nodes. You can use a target again in multiple places too, that's by design, except the target will be unavailable (like context is) outside its children sub-tree. I thought maybe making it use render-props might make this more clear, but I opted to avoid it here.

gut4:
Maybe use panResponder's design? https://facebook.github.io/react-native/docs/panresponder.html
```jsx
function PressableButton ({children, onPress, onHover, onFocus}) {
  const { bind } = useResponder({
    onPress: onPress,
    onHover: onHover,
    onFocus: onFocus
  })

  return (
    <div {...bind}>
      {children}
    </div>
  )
}
```

devongovett:
@gut4 the problem with that is when you want to have multiple responders on the same element and they both have e.g. `onClick` props. Then you have to merge the props somehow or the rightmost spread will win.

trueadm:
@gut4 Also React Flare doesn't use inline props on host components. React Flare has event responders which are competely separate entities from host components - that way we don't need to worry about conflicts with existing events (`onClick` etc). Plus as @devongovett explained, you might want many responders to listen to the same events, which is possible in Flare, but not with today's event system.

sebmarkbage:
Just a meta note. We need to be careful about creating too much ceremony. I feel like this happens every time we propose a new API. We tend to overthink the things we need to be defensive about. There's always an intuition about adding some more verbose API to make it more "explicit" when something is still unfamiliar but as soon as it becomes the standard way of doing things people immediately want to remove it. E.g. The Hooks RFC is full of suggestions about limiting functionality to be conservative and adding ceremony but the actual problems that people ended up being completely different ones. I think the goal here is making this API be powerful with few explicit steps without going overboard and causing subtle refactoring bugs as a result.

For example, we probably don't want to make it explicit on a per responder type level. Now all the components have to know exactly what extensions it allows. That's the opposite of composability. It should at least be a hierarchy of different types of responder. @trueadm's proposal is better because it's a single target per all responders.

There might be cases where refactoring problems needs to be considered such as adding a `<span>` to a `<button>` but maybe this can be addressed with something broader. We also have other things than runtime mechanisms that can help us. E.g. types (or lint). That can help keep the actual semantics simpler.

When you think about extensibility and composability, one way to think about it is what you can already do with more indirection and also what you would like to do with more abstractions.

The question of what is different between extending a parent vs child can be demonstrated by inserting an extra div.

```
<EventComponent>
  <Button>
    ...
```

vs.

```
<EventComponent>
  <div>
    <Button>
      ...
```

This means that you can always add extra event handlers for things like pointers by wrapping it.

However, this wrapping is not quite equivalent because the `<div>` can take up all the available space as its hit surface but the Button might render itself as smaller than the available space. The concept of the Button as a clickable thing might also just be a subset of what the outer bounds that the Button takes up. Therefore EventComponent is breaking the abstraction at a layout level.

Now if we look at child attaching to the parent:

```
<Button>
  <EventResponder />
  ...
```
vs.

```
<Button>
  <div>
    <EventResponder />
  ...
```

Just adding a div by itself won't actually allow this to work, however, any child gets to know the available space that the Button allows its children to take up. So you could make the div `position: absolute; left: 0; top: 0; bottom: 0; right: 0` or similar techniques. The DOM doesn't necessarily make this easy and predictable, but the mechanism is there.

All children have access to the available space in the parent. Therefore adding a pointer target to the parent doesn't add any capability that isn't already there through a nested child.

If you make custom hit testing (e.g. like in ART), this becomes very clear because a child component gets to reason about the available space in the parent and it can just test whether a point is within those bounds. Simple. `hitTest(point) { return isInBounds(point, this.availableSpace); }`

That's what leads me to the intuition that "available space" is a good "context type". It just so happens that this is more or less the thing that all layout aware nodes have in common. They just reason about that space. There's basically no parent that can accept children that wouldn't allow this. (SVG is a bit of different since they explicitly aren't layout aware but you can think of anything that accepts `fill` as providing layout bounds to children.)

All this is saying is that it's not drilling through some concept that doesn't already exist so it's not exposing a new power.

That doesn't mean that this always is the *correct* bounds - semantically. E.g. in the case of:

```
let Button = ({children}) => {
  return (
    <button>
      <Icon />
      <span class="button-label">{children}</span>
    </button>
  );
};
```

It's still possible to extend the span since it does have layout, and sometimes it's even what you want.

In this world, I'd probably design Button to be more explicit about this by providing an extension point where pointers extensions are supposed to be put:

```
let Button = ({children, hitTargetExtensions}) => {
  return (
    <button>
      <Icon />
      <span class="button-label">{children}</span>
      {hitTargetExtensions}
    </button>
  );
};
```

This also ensures that if the extension needs to render any additional DOM nodes, those are related to the hit target and not the span.

This becomes a pattern rather than adding to the feature set.

One concern is that might be hard to predict when designing the button. E.g. maybe you start out not having a span and then add one. However, that is already concern when designing reusable components. Often you end up wanting to add extension points elsewhere (like the icon) or wanting to split the children somehow. If you have to do this after the fact you can also use React.Children (or in the future useChildren or Call/Return to drill through abstractions and do a brand check).

The createTarget mechanism is a neat shortcut but is it sufficient? E.g. sometimes the pointer target isn't the same as the focusable target. Sometimes there's not even 1:1 pointer and focus. You can multiple focusable targets within one pressable component, or sometimes multiple different areas (such as oddly shaped targets) that are pressable but the item is conceptually one focusable thing. It also means that if my abstraction needs to add DOM nodes those will be in a different place than the place where my responders attach. I'm thinking overlays like sparkle effects or accessibility highlights.

```
let Button = ({children, hitTargetExtensions, focusExtensions}) => {
  return (
    <button>
      <Icon />
      <div>
        <span class="button-label">{children}</span>
        {hitTargetExtensions}
      </div>
      {focusExtensions}
    </button>
  );
};
```

Types can also work to help us here. Various responders can be typed differently in terms of children that are accepted. You can type a component to only accept DOM node children and not responders, or only specific types of responders.

E.g. if you start with generic children:

```
type Props = {|
  children: React.Node,
|};
let Button = ({children}: Props) => {
  ...
```

Then later realize that you need to split them out:

```
type Props = {|
  children: React.Nodes<React.DOMNode>,
  hitTargetExtensions: React.Nodes<React.DOMNode | React.PointerResponder>,
  focusExtensions: React.Nodes<React.TouchResponder>,
|};
let Button = ({children, hitTargetExtensions, focusExtensions}: Props) => {
  ...
```

You can now use the type system to find the places that pass responders as children.

Focus is a bit different than pointers here because it doesn't have to do with the "available space". It's not true that I can just add a `<div>` and that can intercept the parent's focus. However, I can always add a focusable DOM node as the first child and now I've effectively made that take focus in the same place as the parent. I can almost do the same thing just like I can almost do the same thing for click targets with `position: absolute`. We're just refining a capability that is already there. The thing that is a bit weird is if there already is a Focus responder on the parent but that's an issue to consider regardless.

That said, I'm not sure extending high level abstractions is necessarily a particularly common example here. I'd expect that you'd mostly build high level abstractions that have all these responders built in and it's the listeners on the outside that's the variable.

trueadm:
@sebmarkbage Nice write up. I do feel like we've gone in a circle though, as what you're proposing with specific props, i.e.:

```jsx
let Button = ({children, hitTargetExtensions, focusExtensions}) => {
  return (
    <button>
      <Icon />
      <div>
        <span class="button-label">{children}</span>
        {hitTargetExtensions}
      </div>
      {focusExtensions}
    </button>
  );
};
```

Feels like you're going back on the idea of composing directly via children somewhat. Especially if you take your earlier idea around blocking together styling and events:

```jsx
<div>
...
<TouchAndHoverEffectWithSparklingOverlay />
</div>
```

> Just a meta note. We need to be careful about creating too much ceremony. I feel like this happens every time we propose a new API. We tend to overthink the things we need to be defensive about. There's always an intuition about adding some more verbose API to make it more "explicit" when something is still unfamiliar but as soon as it becomes the standard way of doing things people immediately want to remove it. 

I think my `createTarget` proposal was sligthly misunderstood here too, I wasn't suggesting my proposal was just more explciit. It's actually more powerful and less verbose than previous ideas because it changes the mindset of how you go about writing a component. With your approach, just like I did, you'd be trying to tack it on to existing ideas/concepts/components and fail miserably when things don't work. With `createTarget`, you need to visualize your boundaries and setup targets to either be private or public, then expose them.

Furthermore, `createTarget` is much better for mutliple targets without worrying about assigning props for them, you can just use children. It tackles all the same problems you mentioned in your previous issue with the nice benefit in that also provides:

- Much better Flow/TS typing by default.
- Explicit targets by default
- Clearer seperation of boundaries
- Improved runtime performance
- Better composability via children

So taking your example in the reply above, we can re-write it with targets:

```jsx
export FocusTarget = React.createTarget('button');
export HitTarget = React.createTarget('div');

export const Button = ({children}) => {
  return (
    <FocusTarget>
      <Icon />
      <HitTarget>
        <span class="button-label">{children}</span>
      </HitTarget>
    </FocusTarget>
  );
};
```

Then if we wanted to assign focus or hit events responders, we can do that nicely:

```jsx
import {FocusTarget, HitTarget, Button} from 'button';

let ButtonExtentions = (
  {!isLoading && <Pressability target={HitTarget} />}
  <Focus target={FocusTarget} />
)

<Button>
  <ButtonExtension />
</Button>
```

IMO that reads and looks much more declarative too, as it enforces you think about what your targets should be when building a component. It also tackles a core problem here, which I think often goes missed. If you don't have discrete targets anad only use props, we have this problem that exists today (that we want to get rid of):

```jsx
let Component = props.Component;

<Component>
  <EventResponder />
</Component>
```

This is another issue I ran into when trying this code out in places. Conditionally switching host components is messy anyway, but it's a fairly frequent pattern. We would rightly unmount-mount in this case, but when it comes to some responders `focus`, `scroll` and `drag`, this can be a bit confusing. That's why I didn't opt for `createTarget` to define the host component in the render phase, and instead define it outside.

This discussion has proven to be very informative though! Let's keep progressing with thoughts and ideas!

sebmarkbage:
> What if the outer parent has a different hit rect than the inner child?

That's when you'd need to expose a different prop that can be delegated to the outer parent.

> we can provide control over what is private and public.

Part about my point about equivalency is that this the case regardless because once you accept a child, then that child can render something that takes up the equivalent space and does the same thing, or render something that accepts focus as a child. Like its own div. So you'd need to specify very defined constraints on the child that it only accepts a very narrow set of children that needs to be specifically defined what they're allowed to do. E.g. maybe they're not allowed to add children with any nested responders at all. That seems like a limitation that they type system needs to enforce though. From an API perspective you can instead by into it. Yea, you can extend it here, but maybe you shouldn't? In the common case, it leaves to a flexible API though.

> as it enforces you to think about what your targets should be when building a component

Is your proposal that this is required that you provide all targets or is there a default? E.g. that host components capture them otherwise. How do you provide the bounds for how deeply it leaks? It seems easy to forget one and now anything attaches to some parent high above the component.

> what if host components accept a function as a single child that returns its target for its sub-tree

Seems like the component exposing the render prop API would have to know ahead of time which split of targets it wants to expose to the children render function. Otherwise you end up with the same problem when you refactor and want to split them. At that point is seems equivalent to just adding an extra prop to me.

trueadm:
> So you'd need to specify very defined constraints on the child that it only accepts a very narrow set of children that needs to be specifically defined what they're allowed to do. E.g. maybe they're not allowed to add children with any nested responders at all. That seems like a limitation that they type system needs to enforce though. From an API perspective you can instead by into it. Yea, you can extend it here, but maybe you shouldn't? In the common case, it leaves to a flexible API though.

There is no way to enforce that properly today with Flow or TS. Especially as it becomes brittle to define such things when you wrap around layers and layers of composition. This has been a big issue in general. I think you'll just end up with everyone accepting this API is public and things in a sub-tree and manipulate the host node above it, and there's little that can be done to stop it.

> Is your proposal that this is required that you provide all targets or is there a default? E.g. that host components capture them otherwise. How do you provide the bounds for how deeply it leaks? It seems easy to forget one and now anything attaches to some parent high above the component.

I think you'd have a set of targets per component. Which is why the render prop API might make more sense here.

> Seems like the component exposing the render prop API would have to know ahead of time which split of targets it wants to expose to the children render function. Otherwise you end up with the same problem when you refactor and want to split them. At that point is seems equivalent to just adding an extra prop to me.

How do you mean know ahead of time? There's only one target from a host component. This target would only be accessible from with the sub-tree of the host component too, exposing it and using it outside will result in an invariant.

devongovett:
Nice writeup, @sebmarkbage. I think your argument relating layout to these event components helped. I think my main concern was around refactoring, but you're right that when you pass through children as in my button example, those children are implicitly dependent on their parent already to determine their layout. If you later refactor and add a new child, that could potentially affect the layout of the children being passed in.

I also agree that most of the time components will pre-bake functionality (e.g. button could still have an `onPress` prop using the press responder by default). I do really like @trueadm's `createTarget` proposal though in terms of allowing easy extensibility through children while ensuring responders are always attached to the the correct target. I think it might also fit in nicely with the `createAccessibleComponent` proposal from #16009. Perhaps one proposal could actually solve both issues?

trueadm:
Another PR is up that has a slightly different design that is different from that of this PR. #16163.

PR #16125, opened 2019-07-12
Prevent Closure from inlining flushTaskAtPriority

acdlite:


acdlite:
Outdated. Closure now supports `/** @noinline */`.

PR #16126, opened 2019-07-12
Remove React.error and React.warn

bvaughn:
Sebastian and I have just discussed [my concerns around `React.warn` and `React.error`](https://github.com/facebook/react/pull/16017). The quick takeaway is this: I am going to remove `React.warn` and `React.error` for now. We can revisit these methods later as part of a larger discussion around the console logging experience with React (see #15726) but we should not block 16.9 for this.

In the meanwhile, I will be adding functionality to React DevTools to intercept calls to `console.error`, `console.warn`, and `console.trace` and auto-append component stacks (in DEV mode). This seems like a nice 80/20 solution that doesn't limit our options for a future API.


eps1lon:
> In the meanwhile, I will be adding functionality to React DevTools to intercept calls to console.error, console.warn, and console.trace and auto-append component stacks (in DEV mode).

While this does sound like a good solution for experienced developers it still requires additional setup and navigation for novice developers. Debugging react shouldn't require a separate extension for the most basic things (like a "call"stack).

Is this really just removed because some rogue libraries might parse the component stack? Should we really optimize against possible hacks instead of for "normal" users?

bvaughn:
> While this does sound like a good solution for experienced developers it still requires additional setup and navigation for novice developers. Debugging react shouldn't require a separate extension for the most basic things (like a "call"stack).

React DevTools are not just for "experienced developers" but all levels 😄 The component stack, while _very helpful_, is not _required_ for debugging React apps. (It hasn't existed up until this point, after all.)

> Is this really just removed because some rogue libraries might parse the component stack? Should we really optimize against possible hacks instead of for "normal" users?

No. I linked to #16017 which specifies a few concerns. Probably my biggest at this moment is that overriding only these two console methods also feels like an _incomplete solution_ to the larger logging story (#15726).

eps1lon:
> React DevTools are not just for "experienced developers" but all levels 

I was concerned with the existing setup. If I start out with react it's highly unlikely I have the devtools set up.

> The component stack, while very helpful, is not required for debugging React apps. (It hasn't existed up until this point, after all.)

I'm not sure about the specific timeline but it existed for a long time in react itself (react specific warnings or prop-types). 

I fear this change is too tunnel visioned on react-devtools only. It doesn't ship in the browser by default.

bvaughn:
> If I start out with react it's highly unlikely I have the devtools set up.

In this scenario, the first time you ran React in development mode you would see this message in the console:

Screen Shot 2019-07-12 at 7 09 06 PM

> I'm not sure about the specific timeline but it existed for a long time in react itself (react specific warnings or prop-types).

The thing that has not existed (because we never actually released these APIs) is the ability to log _custom errors or warnings in user land_ with component stacks attached.

There's a balance to be struct between adding more things to React (which adds weight to the framework, even for the majority of people who use React apps and are _not developers_) and adding things to the surrounding tooling (which is opt-in and only impacts _developers_).

We don't make decisions like this lightly. There were several rounds of conversation between various parts of the team, and a couple of proposals (written by me). There is also still an outstanding issue (linked to in this PR) to come up with a more comprehensive logging plan.

This decision feels like the right set of trade offs at this moment in time though (in preparation for the 16.9 release).

babangsund:
@bvaughn Was this originally inspired by the warning and invariant? 😄

bvaughn:
@babangsund Not directly, not really. More so by `console.warn` and `console.error` TBH.

PR #16133, opened 2019-07-15
Inject getCurrentFiber() function to DevTools

bvaughn:
This returns the current value of ReactCurrentFiber and enables DevTools to append a custom (owner-only) component stack to warnings and errors in DEV mode.

This is an attempt to address concerns about component stacks being overly noisy in certain environments. It does feel like an improvement based on even my own test harness, e.g.
image

Supersedes #16127

### DIFF
--- a/packages/react-reconciler/src/ReactFiberReconciler.js
+++ b/packages/react-reconciler/src/ReactFiberReconciler.js
@@ -473,7 +473,7 @@ if (__DEV__) {
 
 export function injectIntoDevTools(devToolsConfig: DevToolsConfig): boolean {
   const {findFiberByHostInstance} = devToolsConfig;
-  const {ReactCurrentDispatcher, ReactDebugCurrentFrame} = ReactSharedInternals;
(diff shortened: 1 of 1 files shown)


sebmarkbage:
Current fiber should never be used because it's meant to support production runtime semantics (like string refs) which are not the same as the ones powering the debug frame. If you need something different than debug frame for some mechanism, then you should add something to debug frame separately from this field. That way current fiber's details can change and be removed from runtime separately from debug information. E.g. current fiber should really be null in class constructors but I think it's not because someone at some point wanted the debug info in the class constructor and subsequently negatively affected our string ref semantics.

That said, I don't get why debug frame isn't enough since the filtering of owners has to be done using the owner field on the fibers anyway.

bvaughn:
We chatted and I think you were confusing `ReactCurrentOwner` with `ReactCurrentFiber`. I think `ReactCurrentFiber` (the one we're using) is the _right_ one.

PR #16134, opened 2019-07-15
Add suspenseCallback feature for runtime tracing of loading states

bgirard:
This adds a 'SuspenseCallback' feature flag. When the property is set on
a suspense component it will be called during the commit phase with a
set of the immediate thenable for this component. This will allow user
code to build runtime tracing of the cause for a suspense boundary.

CC @sebmck

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -30,6 +30,7 @@ import {
   enableSuspenseServerRenderer,
   enableFlareAPI,
   enableFundamentalAPI,
+  enableSuspenseCallback,
 } from 'shared/ReactFeatureFlags';
 import {
   FunctionComponent,
@@ -1320,6 +1321,18 @@ function commitSuspenseComponent(finishedWork: Fiber) {
 
   if (supportsMutation && primaryChildParent !== null) {
     hideOrUnhideAllChildren(primaryChildParent, newDidTimeout);
+  }
+
+  if (enableSuspenseCallback && newState !== null) {
+    const suspenseCallback = finishedWork.memoizedProps.suspenseCallback;
+    if (typeof suspenseCallback === 'function') {
+      const thenables: Set<Thenable> | null = (finishedWork.updateQueue: any);
+      if (thenables !== null) {
+        suspenseCallback(new Set(thenables));
+      }
+    } else if (__DEV__) {
+      warning(false, 'Unexpected type for suspenseCallback.');
+    }
   }
 }
 
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -115,6 +115,7 @@ import {
 } from './ReactFiberHydrationContext';
 import {
   enableSchedulerTracing,
+  enableSuspenseCallback,
   enableSuspenseServerRenderer,
   enableFlareAPI,
   enableFundamentalAPI,
@@ -883,6 +884,14 @@ function completeWork(
           // is currently timed out, too.
           workInProgress.effectTag |= Update;
         }
+      }
+      if (
+        enableSuspenseCallback &&
+        workInProgress.updateQueue !== null &&
+        workInProgress.memoizedProps.suspenseCallback != null
+      ) {
+        // Always notify the callback
+        workInProgress.effectTag |= Update;
       }
       break;
     }
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseCallback-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseCallback-test.internal.js
@@ -1,0 +1,217 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails react-core
+ * @jest-environment node
+ */
+'use strict';
+
+let React;
+let ReactFeatureFlags;
+let ReactNoop;
+let Scheduler;
+
+describe('ReactSuspense', () => {
+  beforeEach(() => {
+    jest.resetModules();
+    ReactFeatureFlags = require('shared/ReactFeatureFlags');
+    ReactFeatureFlags.enableSuspenseCallback = true;
+
+    React = require('react');
+    ReactNoop = require('react-noop-renderer');
+    Scheduler = require('scheduler');
+  });
+
+  function text(t) {
+    return {text: t, hidden: false};
+  }
+
+  function createThenable() {
+    let completed = false;
+    const resolveRef = {current: null};
+    let promise = {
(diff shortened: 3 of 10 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1269,6 +1270,19 @@ function commitSuspenseComponent(finishedWork: Fiber) {
   if (supportsMutation && primaryChildParent !== null) {
     hideOrUnhideAllChildren(primaryChildParent, newDidTimeout);
   }
+
+  if (
+    enableSuspenseCallback &&
+    'suspenseCallback' in finishedWork.memoizedProps
```
sebmarkbage: You can skip this check since it will be implied by the get below.

sebmarkbage:
It's not obvious to me why this would break in the persistent renderer (used for RN Fabric). You may have stubbled upon an unrelated bug.

bgirard:
The failure is caused by this line:
https://github.com/facebook/react/blob/9f395904c6033598ba8bf47f5497fd6e5077c16d/packages/react-reconciler/src/ReactFiberCompleteWork.js#L837-L845

My vague understanding is that I'm relying on this line to sometimes cause an update incidentally. I need to decide when an update needs to actually happen and trigger it directly.

I can fix it by forcing `workInProgress.effectTag |= Update;` always. Are we expecting completeWork to only cause an update (commit?) only when needed? That's what I'm exploring right now. I think I want to compare the current/WIP fiber and cause an update if their updateQueue (thenablesSet) changed.

Does that sound right?

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -854,6 +855,10 @@ function completeWork(
           workInProgress.effectTag |= Update;
         }
       }
+      if (enableSuspenseCallback) {
+        // Always notify the callback
+        workInProgress.effectTag |= Update;
```
bvaughn: Might be nice to only schedule this effect when a callback has actually been provided.

packages/react-reconciler/src/__tests__/ReactSuspenseCallback-test.internal.js
```diff
@@ -0,0 +1,209 @@
...
+    ReactNoop.render(element);
+    expect(Scheduler).toFlushWithoutYielding();
+    expect(ReactNoop.getChildren()).toEqual([text('Waiting Tier 1')]);
+    expect(lastThenablesCallback1).toEqual(new Set([promise2]));
+
+    resolveRef2.current();
+    ReactNoop.render(element);
+    expect(Scheduler).toFlushWithoutYielding();
+    expect(ReactNoop.getChildren()).toEqual([text('Done'), text('Done')]);
+    expect(lastThenablesCallback1).toEqual(new Set([]));
```
sebmarkbage: It would be good to assert on the number of callbacks. One pattern that we use a lot is:
```
let ops = [];
const suspenseCallback1 = thenables => {
  ops.push(thenables);
};
...
expect(ops).toEqual([new Set([promise2])]);
ops = [];
...
expect(ops).toEqual([]);
```

That way you're asserting on the number of calls and get to see the wrong sequence in the logs.

I was originally thrown off by the error because I thought the persistent renderer was actually passing the wrong thenables here but now I realize that it just wasn't invoking it.

sebmarkbage:
Unrelated but I found a bug in the "train model". Because we aggressively mark this for updates, we commit too often which marks the last commit time and pushes out the train model's next commit further than it should.

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberCommitWork.js#L1266

packages/react-reconciler/src/__tests__/ReactSuspenseCallback-test.internal.js
bgirard: I was afraid of over-specifying my test. As-is I don't diff the thenables set before invoking the callback. So unrelated updates to the suspense component may cause more or less callback, which I think is related to the failure we got.

Should I ten diff the thenables? In the complete or commit phase?

sebmarkbage:
I think there's a disconnect somewhere.

Why do you need the callback to fire after everything has been shown? It doesn't have any promises to give you since it's not waiting on anything. Wouldn't it be fine to just ignore the callback in this case?

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1269,6 +1270,16 @@ function commitSuspenseComponent(finishedWork: Fiber) {
   if (supportsMutation && primaryChildParent !== null) {
     hideOrUnhideAllChildren(primaryChildParent, newDidTimeout);
   }
+
+  if (enableSuspenseCallback) {
+    const suspenseCallback = finishedWork.memoizedProps.suspenseCallback;
+    if (typeof suspenseCallback === 'function') {
+      const thenables: Set<Thenable> | null = (finishedWork.updateQueue: any);
+      suspenseCallback(new Set(thenables));
```
sebmarkbage: I would only invoke this callback if `newState` is non-null. Meaning only when it is in its "fallback state". I think that should be sufficient to reach parity between the persistent and mutable renderers.

bgirard: Yes, it would be fine not to fire the callback when not in the fallback state. The user code will probably just call `then` on the promise to track exactly when they complete and possibly watch the mount/unmount of the child/fallback.

Now do we want to specify, in the test, if the callback should fire if the thenableSet is identical? This means diffing the current/WIP thneableSet. If so should that would need to happen in Commit work before we call. If we don't specify it then when wont know for sure how many calls to expect in more tricky case. For the user code it's not a big deal because the client code will want to track promises individually anyways.

sebmarkbage: I don’t think we should necessarily dedupe because React shouldn’t be required to keep the cache of promises forever so we don’t want to keep that guarantee.

That said, the retryCache check is probably too late atm and we should move to the ReactFiberThrow as a separate optimization which has the side effect of this set not getting duplicates... mostly.

bgirard:
Applied the suggestions, thanks for the tip on fleshing out the callback behavior. I updated the user code to match with this and I think it looks decent. I'm getting great traces from this.

bgirard:
rebased, should be ready to go.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -884,6 +885,13 @@ function completeWork(
           workInProgress.effectTag |= Update;
         }
       }
+      if (
+        enableSuspenseCallback &&
+        workInProgress.memoizedProps.suspenseCallback != null
+      ) {
+        // Always notify the callback
+        workInProgress.effectTag |= Update;
```
sebmarkbage: Since this will only be called (for this purpose) if updateQueue is non-null, you should check `if (workInProgress.updateQueue !== null)` before adding this effect.

bgirard:
The build is stuck, but I rebased and retest against FB WWW.

PR #16141, opened 2019-07-16
Move ReactFiberErrorDialog RN fork into RN itself

motiz88:
Along with facebook/react-native#25671, moves the RN-specific implementation of `ReactFiberErrorDialog` from React to RN. This should only be merged (and subsequently synced to RN) _after_ facebook/react-native#25671 lands.

Also, with this, React doesn't use `ExceptionsManager` directly anymore, so we remove it from the RN private interface definition.


threepointone:
Sorry if I've misunderstood, but you did mention that this should only be merged only after https://github.com/facebook/react-native/pull/25671 lands, which it hasn't. Intentional? 

(I ask because I'm planning on cutting a new React alpha very soon)

motiz88:
@threepointone What I really meant is that this can't be synced to RN before the RN PR lands (side note: it's landing now). It's not actually a blocker to merging or to cutting a React release.

threepointone:
Good to know, thank you!

PR #16143, opened 2019-07-16
Support React.createContext in ReactShallowRenderer

### LINKED ISSUE #14442: ShallowRenderer does not work with Class.contextType

Bug

The shallow renderer from the 'react-test-renderer' package does not work with [Class.contextType](https://reactjs.org/docs/context.html#classcontexttype). The component always receives an empty object as the context.

The shallow renderer should forward the context provided to the render method to the rendered component's this.context.

16.6.3

**Steps to reproduce**

The test below can be used to reproduce the problem ([Open on codesandbox.io](https://codesandbox.io/s/oql0l6p95)). The first test reproduces the bug and the second test shows a workaround.

    import * as React from "react";
    import ShallowRenderer from 'react-test-renderer/shallow'
    import PropTypes from 'prop-types';
    
    const TestContext = React.createContext({message:"TEST1"});
    class TestShallow extends React.Component {
    
        static contextType = TestContext;
    
        componentDidMount() {
            console.log(this.context.message);
        }
    
        render() {
            return <h1>{this.context.message}</h1>;
        }
    }
    
    it('shallow render with contextType', () => {
    
        const result = new ShallowRenderer().render(<TestShallow />, {message:"TEST2"});
        expect(result).toEqual(<h1>TEST2</h1>); //FAILS
    
    });
    
    it('workaround shallow render with contextType', () => {
    
        TestShallow.contextTypes = {
            message: PropTypes.string
        };
    
        const result = new ShallowRenderer().render(<TestShallow />, {message:"TEST2"});
        expect(result).toEqual(<h1>TEST2</h1>); //WORKS
    
    });

The problem seems to be located [here](https://github.com/facebook/react/blob/2743fb7b23a5af42059a9b07c5f6048f941709d9/packages/react-test-renderer/src/ReactShallowRenderer.js#L146). The function getMaskedContext() checks the properties declared on type.contextTypes and filters out the properties from the context provided to the shallow render unless the types.contextTypes is declared.

hassaans:
Fixes #14442

- Adds support for `React.createContext` and `contextType` in React.ShallowRenderer.

- We're still supporting [contextTypes](https://reactjs.org/docs/context.html#legacy-api) in React 16.x releases, so the code for that was not removed.

### DIFF
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -536,7 +536,10 @@ class ReactShallowRenderer {
     this._element = element;
 
     if (elementType.contextType) {
-      this._context = context !== emptyObject ? context : elementType.contextType._currentValue;
+      this._context =
+        context !== emptyObject
+          ? context
+          : elementType.contextType._currentValue;
     } else {
       this._context = getMaskedContext(elementType.contextTypes, context);
     }
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -534,7 +534,12 @@ class ReactShallowRenderer {
 
     this._rendering = true;
     this._element = element;
-    this._context = getMaskedContext(elementType.contextTypes, context);
+
+    if (elementType.contextType) {
+      this._context = context !== emptyObject ? context : elementType.contextType._currentValue;
+    } else {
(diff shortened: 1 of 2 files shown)


ohjames:
Maybe merge these commits into one and suffix the message with the issue like `(#14442)`. Also recent commits to the shallow renderer start with `[Shallow]`.

garritfra:
Just scrolled through open issues and found #16143, which is blocked by this PR. Could someone take a look at this?

PR #16145, opened 2019-07-17
[Scheduler] Profiling features

acdlite:
Adds two new profiling features to Scheduler.

## Sample-based approach

The main feature is a way to read the current state of the Scheduler from a worker, using a shared array buffer. The shared array buffer contains (only in the profiling and development builds)

- the priority Scheduler at which is currently running (or 0 if it's not)
- the size of the queue
- the id of the currently running task (if applicable)
- a unique "run id" per block (if applicable)

Regarding the last point each synchronous block of Scheduler work is given a unique run ID. This is different than a task ID because a single task will have more than one run if it yields with a continuation.

The worker thread can use this information to construct a flamegraph. The advantage of the sample-based approach is that it scales in proportion to the available resources — if the CPU slows down, the sample rate slows down accordingly.

## Event-based approach

The disadvantage of the sample-based approach is that it doesn't during the very first part of page initialization, before the worker has initialized.

To address this problem, there's a separate feature to write to a log of events: `startLoggingProfilerEvents` and `stopLoggingProfilerEvents`.

`stopLoggingProfilerEvents` returns a typed, 32-bit integer array buffer (though not a shared array buffer). The log conforms to a custom event format. A profiler can then convert this log into a flamegraph.

An array buffer is used so that the log can be sent to a worker thread via `postMessage` and the structured clone algorithm.

The test suite works by printing a text representation of a Scheduler flamegraph. Example:

```
!!! Main thread              │          ██
Task 2 [User-blocking]       │        ░░░░██████
Task 1 [Normal]              │  ████████░░░░░░░░██████
```

## Both approaches combined

The suggested approach is to use the event log during page start up, then switch to the sample-based approach once the profiler worker is ready.


packages/scheduler/npm/umd/scheduler.development.js
```diff
@@ -144,5 +160,9 @@
       return global.React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
         .Scheduler.unstable_UserBlockingPriority;
     },
+    get unstable_sharedProfilingBuffer() {
+      return global.React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
+        .Scheduler.unstable_getFirstCallbackNode;
```
gaearon: Is this a copy paste mistake?

packages/scheduler/npm/umd/scheduler.production.min.js
```diff
@@ -138,5 +154,9 @@
       return global.React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
         .Scheduler.unstable_UserBlockingPriority;
     },
+    get unstable_sharedProfilingBuffer() {
+      return global.React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
+        .Scheduler.unstable_getFirstCallbackNode;
```
gaearon: Same question

packages/scheduler/npm/umd/scheduler.profiling.min.js
```diff
@@ -138,5 +154,9 @@
       return global.React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
         .Scheduler.unstable_UserBlockingPriority;
     },
+    get unstable_sharedProfilingBuffer() {
+      return global.React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED
+        .Scheduler.unstable_getFirstCallbackNode;
```
gaearon: Same question

packages/scheduler/npm/umd/scheduler.development.js
acdlite: Yes :)

packages/scheduler/src/SchedulerProfiling.js
```diff
@@ -0,0 +1,210 @@
...
+let runIdCounter: number = 0;
+let mainThreadIdCounter: number = 0;
+
+const profilingStateSize = 4;
+export const sharedProfilingBuffer =
+  // $FlowFixMe Flow doesn't know about SharedArrayBuffer
+  typeof SharedArrayBuffer === 'function'
+    ? new SharedArrayBuffer(profilingStateSize * Int32Array.BYTES_PER_ELEMENT)
+    : // $FlowFixMe Flow doesn't know about ArrayBuffer
+      new ArrayBuffer(profilingStateSize * Int32Array.BYTES_PER_ELEMENT);
```
gaearon: IE9 is still technically supported by us (with polyfills) but AFAIK doesn't have ArrayBuffer. Let's be careful adding new runtime dependencies in unconditional paths, especially for DEV features.

packages/scheduler/src/Scheduler.js
```diff
@@ -152,6 +190,8 @@ function flushWork(hasTimeRemaining, initialTime) {
     }
     // Return whether there's additional work
     if (currentTask !== null) {
+      markSchedulerSuspended(currentTime);
+      isHostCallbackScheduled = true;
```
gaearon: I've found this factoring very counterintuitive.

It's like `Scheduler` knows too much about the host config. It was surprising to see `isHostCallbackScheduled = true` here without any requests, and only later realize it "knows" host callback will be scheduled because it returned `true`, and it trusts host config to respect that.

There's similar implicitness in the error code path.

gaearon: The error code path seems inconsistent with what this code seems to imply.

The host config _does_ schedule a callback in the error case (if this throws).
But the `catch` below doesn't have the corresponding lines:

```js
            markSchedulerSuspended(currentTime);
            isHostCallbackScheduled = true;
```

Should it?

PR #16148, opened 2019-07-17
[Flare] Add Input event responder surface

trueadm:
This PR adds the `Input` event responder. This isn't an `<input />` DOM node, it's a responder surface to capture the following events:

- `onChange`
- `onValueChange`

I've taken almost all (minus a few that were not needed) tests from the existing ChangeEventPlugin from ReactDOM and they all pass correctly with this event responder, meaning you can do this:

```jsx
import {Input} from 'react-events/input';

function MyComponent() {
  const [value, setValue] = useState('');
  
  return (
    <Input onValueChange={setValue}>
      <input type="text" value={value} />
    </Input>
  );
}
```

You could abstract this into a hook and a text field component and it becomes really nice to compose:

```jsx
import {Input, useInput} from 'react-events/input';

function MyComponent() {
  const value = useInputValue();
  return <TextField value={value} />
}

function useInputValue() {
  const [value, onValueChange] = useState('');
  useInput({ onValueChange });
  return value;
}

function TextField({value}) {
  return (
    <Input>
      <input type="text" value={value} />
    </Input>
  );
}
```


packages/react-events/src/dom/Input.js
```diff
@@ -0,0 +1,323 @@
...
+  onKeyPress: (e: InputEvent) => void,
+  onBeforeChange: (e: InputEvent) => void,
+  onChange: (e: InputEvent) => void,
+  onValueChange: (value: string | boolean) => void,
+};
+
+type InputEvent = {|
+  data: string,
+  isComposing: boolean,
+  inputType: boolean,
```
TrySound: Why inputType is boolean? It should be a string, isn't it? Also looks like it's not supported in IE and Edge
https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/inputType

trueadm: That was a typo. We aren't looking to polyfill this though, if it exists we pass it on.

PR #16151, opened 2019-07-17
[Scheduler] Bugfix: Cancelling a continuation

acdlite:
Cancelling the original task should also cancel its continuation.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -144,14 +144,8 @@ function flushTask(task, currentTime) {
   // with the same priority and expiration as the just-finished callback.
   if (typeof continuationCallback === 'function') {
     var expirationTime = task.expirationTime;
-    var continuationTask = {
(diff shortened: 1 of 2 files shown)


packages/scheduler/src/Scheduler.js
```diff
@@ -144,14 +144,8 @@ function flushTask(task, currentTime) {
...
     var expirationTime = task.expirationTime;
-    var continuationTask = {
-      callback: continuationCallback,
-      priorityLevel: task.priorityLevel,
-      startTime: task.startTime,
-      expirationTime,
-      next: null,
-      previous: null,
-    };
+    var continuationTask = task;
```
Aprillion: is it desirable to mutate `task.callback`, `task.next` and `task.previous`? if not, the `task` object needs to be copied, not simply assigned.

acdlite: It's on purpose.

PR #16152, opened 2019-07-17
[Flare] Add FocusWithin responder

necolas:
FocusWithin is implemented as a separate responder, which keeps both focus
responders simple and allows for easier composition of behaviours.

Demo https://codesandbox.io/s/focuswithin-byl1q

Close #15848

### DIFF
--- a/packages/react-events/docs/FocusWithin.md
+++ b/packages/react-events/docs/FocusWithin.md
@@ -1,0 +1,57 @@
+# FocusWithin
+
+The `FocusWithin` module responds to focus and blur events on its child. Focus events
+are dispatched for all input types, with the exception of `onFocusVisibleChange`
+which is only dispatched when focusing with a keyboard.
+
+Focus events do not propagate between `FocusWithin` event responders.
+
+```js
+// Example
+const Button = (props) => {
+  const [ focusWithin, updateFocusWithin ] = useState(false);
+  const [ focusWithinVisible, updateFocusWithinVisible ] = useState(false);
+
+  return (
+    <FocusWithin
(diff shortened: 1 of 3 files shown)


trueadm:
Does it make sense to use focusIn/focusOut for the new responder? Shane about the code duplication too, the Focus module size is quite a bit bigger than before :/ we can come back to that though.

necolas:
> Does it make sense to use focusIn/focusOut for the new responder?

What do you mean?

> Shame about the code duplication too, the Focus module size is quite a bit bigger than before

It's only 0.2 KB larger gzipped after adding new functionality.

packages/react-events/docs/FocusWithin.md
```diff
@@ -0,0 +1,57 @@
...
+      onFocusWithinVisibleChange={updateFocusWithinVisible}
+    >
+      <button
+        children={props.children}
+        style={{
+          ...(focusWithin && focusWithinStyles),
+          ...(focusWithinVisible && focusWithinVisibleStyles)
+        }}
+      >
+    </Focus>
```
devongovett: Should be `</FocusWithin>`

sebmarkbage:
Hm. This is the exact use case that I was thinking about for the Focal hooks. `useFocus`, `useHover` etc. are all "within".

What's the use case for this that the useListener hooks wouldn't be sufficient for?

necolas:
In the current design? How are you expecting `useFocus` to work? If it's listening to focus/blur events coming from `<Focus>` in the subtree it's going to fire those pairs every time focus moves, which isn't what this does.

PR #16156, opened 2019-07-18
recommend react-codemod@next until we actually ship

threepointone:
This changest the UNSAFE_ lifecycle recommendation to `npx react-codemod@next ...`, in preparation for the release. The sequence of events is -
- publish the next version of react-codemod with the `next` tag
- publish the react alpha
- let it simmer for a week or so
- publish react-codemod stable
- another pr that removes `@next`
- release the kraken


threepointone:
This isn't needed anymore, since we published react-codemod

PR #5146, opened 2015-10-12
Don't execute <script> tags w/ createElement mode

spicyj:
Each script will execute at most once so we could also set `.textContent` to something like `;`, then add it to the document, then change the `.textContent`, but this seems like the simplest approach. See http://dev.w3.org/html5/spec-preview/the-script-element.html for details.

PR #16157, opened 2019-07-18
Add trusted types to react on client side

Siegrift:
## Trusted Types
[Trusted Types](https://github.com/WICG/trusted-types) ([spec](https://wicg.github.io/trusted-types/dist/spec/), [introductory article](https://developers.google.com/web/updates/2019/02/trusted-types)) is a new experimental DOM API implemented within the WICG , with a working [Chrome implementation](https://www.chromestatus.com/feature/5650088592408576).

The API creates a few new objects available on the global object in the browser, like most other web APIs ([impl in TS](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/trusted-types/index.d.ts) and in [Closure compiler](https://github.com/WICG/trusted-types/blob/master/externs/externs.js)).

Under certain conditions, controlled by a HTTP header (analogous to Content-Security-Policy behavior), the API can enable the enforcement - then it changes the signature of several DOM API functions and property setters, such that they accept specific object types, and reject strings. Colloquially, DOM API becomes strongly typed.

For example, with Trusted Types Element.innerHTML property setter accepts a TrustedHTML object.

Trusted Type objects stringify to their inner value. This API shape is a deliberate choice that enables existing web applications and libraries to gradually migrate from strings to Trusted Types without breaking functionality. In our example, it makes it possible to write the following:

```javascript
const policy = TrustedTypes.createPolicy('foo', { 
  createHTML: (s) => { /* some validation*/; return s} 
});

const trustedHTML = policy.createHTML('bar');
anElement.innerHTML = trustedHTML

anElement.innerHTML === 'bar'
```

The above code works regardless if the Trusted Types enforcement is enabled or not.

Reading from the DOM is unaffected, so Element.innerHTML getter returns a string. That's for practical reasons -- web applications read from DOM more often than they write to it, and only writing exposes the application to DOM XSS risks. Typing only the setters allows us to secure web applications with minimal code changes.

## Adding Trusted Types to React for client side
React applications rarely manipulate DOM directly as this is what React is for. That means that, if users want to opt in to Trusted Types, values assigned to execution sinks must be trusted. React on its own doesn't produce these execution sinks, except one [hack for IE](https://github.com/facebook/react/blob/b4178af81b02dfa0e898670ce564c4649fd8947f/packages/react-dom/src/client/setInnerHTML.js#L34). The only problem in React is the stringifying of values before assigning them to element properties or attributes. 

This PR enables applications to use Trusted Types on client side. The implementation removes stringifying of trusted values in reactDOM/client. It does so **only** when Trusted Types are available in global object. If they are available, there are special [is* functions](https://wicg.github.io/trusted-types/dist/spec/#dom-trustedtypepolicyfactory-ishtml) which are used to check if the value is trusted type or not. If the value is a trusted type, we return the value itself, otherwise we preserve the original logic and stringify the value.

## Problems
**We cannot use Trusted Types in Internet Explorer (IE) when using dangerouslySetInnerHTML on svg elements** - The cause is that in IE svg elements doesn't have innerHTML property, which means that to support innerHTML, react needs to use a [hack](https://github.com/facebook/react/blob/b4178af81b02dfa0e898670ce564c4649fd8947f/packages/react-dom/src/client/setInnerHTML.js#L34), which wraps the innerHTML of the svg and wraps them in an svg element and assigns this to innerHTML of some HTML element. This is a problem for Trusted Types because the original svg innerHTML value is stringified. There are a few workarounds for this:
a) Use default Trusted Type policy allowing creating svg nodes
b) Do not use svg with dangerouslySetInnerHTML but wrap the svg with div element and use dangerouslySetInnerHTML there 
     
## Reference
* [Draft: Add trusted types to react on server side](https://github.com/facebook/react/pull/16555)

## Update
* [TrustedURLs are deprecated](https://github.com/WICG/trusted-types/pull/204) and are going to be replaced with default policy call on javascript URLs ([which are deprecated in React and will not be supported in the future release](https://github.com/facebook/react/pull/15047/files)).
* [window.TrustedTypes is going to be renamed to window.trustedTypes](https://github.com/WICG/trusted-types/pull/205) (with lowercase **t**)


gaearon:
Thanks for the PR!

>Since Oct 12, 2015 React is making sure scripts are not run inside React components. It does so by creating a parser-inserted script via innerHTML, which never executes. This is a problem for Trusted Types, because innerHTML must use TrustedHTML value. This behavior in React is not documented and there are also no tests for it. We've decided to completely remove that part, but if it's a problem we can preserve the same behavior when Trusted Types are not available.

I think that behavior was always there. It's not new since Oct 12. This is because in the past, *we always used `innerHTML` for generating the markup*. The Oct 12 change was a precursor to #5146 which switched us to `document.createElement`. So it aligned the `createElement` behavior to the old `innerHTML` behavior. Regardless, enabling scripts to execute would be a major breaking change. I don't have a good sense of whether we want to do this in longer term, but for now we should find a way to avoid it if we can.

>Cannot use Trusted Types in Internet Explorer (IE) when using dangerouslySetInnerHTML on svg elements

I'm not following.. Does this mean IE has support for real Trusted Types? Is there any version of IE where this would matter?

packages/react-dom/src/shared/trustedTypesAwareToString.js
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @flow
+ */
+
+/**
+ * Returns true only if Trusted Types are available in global object and the value
+ * is a trusted type.
+ */
+function isTrustedTypesValue(value: any) {
+  if (typeof window.TrustedTypes === 'undefined') {
```
gaearon: Is there any reason for the `window.` qualifier here? Can we write plain `typeof TrustedTypes === 'undefined'`?

Generally in `react-dom` source we try to avoid assuming `window` exists because (depending on the code path) some `react-dom` logic may execute even on the server environment. Such as when some module does `require('react-dom')`. So we want to be careful to not introduce accidental reference errors there.

packages/react-dom/src/shared/trustedTypesAwareToString.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * We allow passing objects with toString method as element attributes or in dangerouslySetInnerHTML
+ * and we do validations that the value is safe. Once we do validation we want to use the validated
+ * value instead of the object (because object.toString may return something else on next call).
+ *
+ * If application uses Trusted Types we don't stringify trusted values, but preserve them as objects.
+ */
+function trustedTypesAwareToString(value: any) {
+  if (isTrustedTypesValue(value)) {
```
gaearon: This seems like it would go through a lot of checks even for the most common case: a string. Can we fast-path a string input first?

packages/react-dom/src/client/setInnerHTML.js
```diff
@@ -27,10 +28,20 @@ const setInnerHTML = createMicrosoftUnsafeLocalFunction(function(
...
       reusableSVGContainer || document.createElement('div');
+    if (__DEV__) {
+      warningWithoutStack(
+        typeof window.TrustedTypes === 'undefined',
+        "Using 'dangerouslySetInnerHTML' in an svg element with " +
+          'Trusted Types enabled in an Internet Explorer will cause ' +
+          'the trusted value to be converted to string. Assigning string ' +
+          "to 'innerHTML' will throw an error if Trusted Types are enforced. " +
+          "You can try to wrap your svg element inside a div and use 'dangerouslySetInnerHTML' " +
+          'on the enclosing div instead.',
```
gaearon: 1. In which situations can IE possibly have TT support? Is this about new Edge? Doesn't it use Blink, meaning it no longer has this issue?

2. Regardless, we usually don't want to warn about a pattern that's broken in one browser, just in that browser. Since developers would usually develop with their browser of choice and only later test it. Generally we want to prevent developers from writing potentially broken code early — instead of waiting for them to test in IE. If React API lets us do something, ideally it would work in all browsers. Let's figure out (1) first though.

packages/react-dom/src/shared/trustedTypesAwareToString.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/**
+ * We allow passing objects with toString method as element attributes or in dangerouslySetInnerHTML
+ * and we do validations that the value is safe. Once we do validation we want to use the validated
+ * value instead of the object (because object.toString may return something else on next call).
+ *
+ * If application uses Trusted Types we don't stringify trusted values, but preserve them as objects.
+ */
+function trustedTypesAwareToString(value: any) {
```
gaearon: Please annotate the return value type. I think currently it's typed as `any`, which means all types downwards are wrong. E.g. `toString()` from `ToStringValue` claims to return a string — but after this change, this is no longer guaranteed. We should encode this in types.

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -19,7 +20,7 @@ export opaque type ToStringValue =
 // around this limitation, we use an opaque type that can only be obtained by
 // passing the value through getToStringValue first.
 export function toString(value: ToStringValue): string {
```
gaearon: I'm wary of the layers here. Something calls `ToStringValue.toString` which itself calls `trustedTypesAwareToString`. It's hard to understand all the callsites and which are intentionally accepting trusted types, vs accidentally because they happen to use this module.

I propose to unify this module and `trustedTypesAwareToString.js` into one. And to be explicit about when we're guaranteed to get a string back vs when we may get a TT object.

packages/react-dom/src/client/setInnerHTML.js
Siegrift: > 1. In which situations can IE possibly have TT support? Is this about new Edge? Doesn't it use Blink, meaning it no longer has this issue?

IE applications can enable TT  if they load the [polyfill](https://github.com/WICG/trusted-types). 

> 2\. Regardless, we usually don't want to warn about a pattern that's broken in one browser, just in that browser. Since developers would usually develop with their browser of choice and only later test it.

Yeah, that is a good point. Easy way is to just log in every browser? The problem is that we couldn't figure out a solution without without introducing a policy for React that blesses hacky solution for IE.

packages/react-dom/src/shared/trustedTypesAwareToString.js
Siegrift: Yeah, my bad. 

Btw. are you ok with typing the return value as `string`? Be aware that TT are not equivalent to `string` and are missing methods like `.replace`

packages/react-dom/src/client/setInnerHTML.js
gaearon: >IE applications can enable TT if they load the polyfill.

Does that polyfill enable enforcement?

packages/react-dom/src/shared/trustedTypesAwareToString.js
gaearon: > Be aware that TT are not equivalent to string and are missing methods like .replace

Right — that's what I want to catch with types. So it can be some opaque type as long as we don't consider it a string.

Since Flow probably doesn't support TT at the moment, we can probably have a module with dangerous DOM sinks which are typed to accept our opaque type.

packages/react-dom/src/client/setInnerHTML.js
Siegrift: Whether Trusted Types are enforced or just in report only mode is up to the application. It can set the mode using the CSP header. The polyfill itself just provides Trusted Types API and patches the DOM sinks. _([reference](https://github.com/WICG/trusted-types#browsers))_

gaearon: Do I understand correctly that this issue would only affect websites enable TT polyfill in enforcement mode on IE? Since if it's not being enforced, `'<svg>' + html + '</svg>'` should just stringify?

Siegrift: Just to clarify -- The problem is not the stringification. Expression `'<svg>' + html + '</svg>'`is stringified independendly of whether Trusted Types are enforced or not. The problem is when you try to assign the **string** value to DOM sink _(e.g. innerHTML)_.

Otherwise your understanding is correct. It will break **iff**:
1) application **enforces** Trusted Types
2) uses `dangerouslySetInnerHTML` on svg nodes
3) runs on IE

Siegrift:
@gaearon I have completed the changes, can you have a look?

gaearon:
I will check on Monday, thank you!

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -415,11 +416,19 @@ export function createElement(
...
       const div = ownerDocument.createElement('div');
+      if (__DEV__) {
+        warningWithoutStack(
+          false,
+          'Encountered script tag while rendering React component. ' +
+            'Scripts inside React components are parser inserted into document ' +
+            'and they are never executed. Furthemore rendering script nodes ' +
+            'inside components breaks when using Trusted Types.',
+        );
+      }
```
gaearon: Is the “parser mode” protection relevant for scripts that have `src` or non standard value of `type` attribute? If not, maybe we can keep allowing those without warning (and not use innerHTML for them)?

koto: If you mean Trusted Types, we require typed value for the content regardless of the `type` attribute (or the presence of other attributes), as we'd be running into difficulties (e.g. what happens if the `src` gets cleared after the text has been modified). Discussion under https://github.com/WICG/trusted-types/issues/6.  

If you mean in HTML, it seems like the spec mandates setting the parser inserted flag unconditionally (https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-inhead). That said, it seems like Chrome doesn't comply: 

```
data:text/html,<script type=foo>alert(1)</script><script>s=document.scripts[0];document.head.removeChild(s);s.type='';document.head.appendChild(s);</script>
```

It is respected in FF though.

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -35,3 +35,48 @@ export function getToStringValue(value: mixed): ToStringValue {
...
+ * Trusted Types undergo a change where window.TrustedTypes was renamed to window.trustedTypes
+ * (https://github.com/WICG/trusted-types/pull/205).
+ */
+function isTrustedTypesValue(value: any): boolean {
+  // $FlowExpectedError - TrustedTypes are defined only in some browsers or with polyfill
+  if (typeof TrustedTypes !== 'undefined') {
+    return (
+      TrustedTypes.isHTML(value) ||
+      TrustedTypes.isScript(value) ||
+      TrustedTypes.isScriptURL(value)
```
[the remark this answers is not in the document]
Siegrift: This was "kinda" on purpose because trusted URLs will go away soon (https://github.com/WICG/trusted-types/pull/204). However I forgot to document this in the PR and also it's hard for other people to test this as chrome implementation and polyfill support trusted URLs. 

However it seems, that we might be able to ship this sooner than the change in Trusted Types, so I will add this behavior back conditionally. 

Thanks for pointing this out.

packages/react-dom/src/__tests__/ReactEmptyComponent-test.js
```diff
@@ -136,6 +136,8 @@ describe('ReactEmptyComponent', () => {
   });
 
   it('should distinguish between a script placeholder and an actual script tag', () => {
+    const consoleSpy = spyOnDev(console, 'error');
```
gaearon: For warnings, we use toWarnDev() matcher. Could you please use that?

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -35,3 +35,52 @@ export function getToStringValue(value: mixed): ToStringValue {
...
   }
 }
+
+/**
+ * Returns true only if Trusted Types are available in global object and the value is a trusted type.
+ *
+ * Trusted Types undergo a change where window.TrustedTypes was renamed to window.trustedTypes
+ * (https://github.com/WICG/trusted-types/pull/205).
+ */
+function isTrustedTypesValue(value: any): boolean {
```
gaearon: Can we make the function declaration itself conditional and check the globals at startup?

```
let isTTValue =
  typeof TT !== "undefined" ?
    impl1 :
    typeof tt !== "undefined" ?
      impl2 :
      impl3
```

Each implementation would assume a specific global exists, and the last one would be a shim that always returns false.

This is to reduce the runtime checks.

.eslintrc.js
```diff
@@ -149,5 +149,7 @@ module.exports = {
     spyOnProd: true,
     __PROFILE__: true,
     __UMD__: true,
+    TrustedTypes: true,
+    trustedTypes: true,
```
gaearon: Do we really need to support both?

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -35,3 +35,52 @@ export function getToStringValue(value: mixed): ToStringValue {
...
+export opaque type TrustedValue: {toString(): string} = {toString(): string};
+
+/**
+ * We allow passing objects with toString method as element attributes or in dangerouslySetInnerHTML
+ * and we do validations that the value is safe. Once we do validation we want to use the validated
+ * value instead of the object (because object.toString may return something else on next call).
+ *
+ * If application uses Trusted Types we don't stringify trusted values, but preserve them as objects.
+ */
+export function trustedTypesAwareToString(value: any): string | TrustedValue {
```
gaearon: Maybe `toStringOrTrustedType`?

.eslintrc.js
Siegrift: In .eslintrc yes. But I think you meant it in a more general way. Currently, the native chrome implementation uses TrustedTypes (with uppercase T) and that is the preferred way to test Trusted Types.

packages/react-dom/src/__tests__/ReactEmptyComponent-test.js
Siegrift: I have a problem with this as I already use `expect` to check that there is no error. I tried to google if I can chain expects in jest, but I didn't find how :/.

.eslintrc.js
gaearon: It's unfortunate for the bundle size to support both. Can we start with one version, and then switch over to another one when it reaches stable?

packages/react-dom/src/__tests__/ReactEmptyComponent-test.js
gaearon: >I already use expect to check that there is no error.

I'm not sure what this means, but if you mean `not.toThrow()`, you can just remove that completely. It doesn't mean anything. If it threw, the test would fail anyway.

.eslintrc.js
Siegrift: Yeah, we can do that. Users can set `window.trustedTypes = window.TrustedTypes` before bootstrapping React.

gaearon:
Should we put it behind a feature flag? See `ReactFeatureFlags`. The motivation is that we might want to tweak the semantics before getting it into a stable release. Putting the support behind a feature flag lets us continue to do release with it off, while testing with it on internally at FB. Then when we’re ready to commit to the semantics, we can flip the flag for everyone. What do you think?

koto:
How can clients use feature flags outside of FB? I assume a rebuild of React is needed to have one's own flag set (based on [this comment](https://github.com/facebook/react/pull/16254#issuecomment-517768661)?

We can put it behind a flag, but so far the changes look quite inobtrusive (it's only removing stringification for sinks, warning about SVG interpolation in IE and `<script>`s in JSX). Aflag might be an overkill for now. Your call though :)

gaearon:
>How can clients use feature flags outside of FB? I assume a rebuild of React is needed to have one's own flag set (based on this comment?

You'd need to do a build yourself, yes.

>A flag might be an overkill for now. Your call though :)

A flag is a safer option because if something is wrong, we can disable it instead of reverting the whole PR to unblock other work. We use flags for all non-trivial changes, and this one is a bit non-trivial. I don't mean a long-lived flag, but just for a few weeks so we can start testing this safely without worrying about it affecting release timing.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -417,11 +418,19 @@ export function createElement(
         type,
       );
     }
-
     if (type === 'script') {
       // Create the script via .innerHTML so its "parser-inserted" flag is
       // set to true and it does not execute
       const div = ownerDocument.createElement('div');
+      if (__DEV__) {
+        warningWithoutStack(
```
gaearon: Please deduplicate this warning with a boolean so it only ever fires once. (There's some examples in the codebase.)

This should also probably be `warning` rather than `warningWithoutStack`.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -417,11 +418,19 @@ export function createElement(
...
+          false,
+          'Encountered script tag while rendering React component. ' +
```
gaearon: Nit: Encountered _a_ script tag

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -417,11 +418,19 @@ export function createElement(
...
+            'Scripts inside React components are parser inserted into document ' +
+            'and they are never executed. Furthemore rendering script nodes ' +
+            'inside components breaks when using Trusted Types.',
+        );
+      }
```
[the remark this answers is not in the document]
Siegrift: Thanks for asking the community about this. 

Unfortunately, there is no workaround for Trusted Types (we want the content of the element to be TrustedScript unconditionally). Also in this particular case, the problem is the assignment to innerHTML of the div element.

But there is a workaround for not using the script tag. From the community responses I see, that the only legitimate use case is to have something in the DOM (which is not executed, but [some widgets may need the data](https://mobile.twitter.com/andytuba/status/1159984414681645057)). For this use case users could use [template element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) instead (it is supported everywhere apart from IE).

Warning only in cases when the script is should execute is not that easy. For example, there is also a `language` attribute which affects whether the script should be executed or not. You can read more about this in the [HTML scripting spec](https://html.spec.whatwg.org/multipage/scripting.html).

However, this warning is not only for Trusted Types and I think it might make sense to do it in a separate PR, too. WDYT?

The last thing I want to point out, that this is scripts don't execute only on client side. However on SSR they work (as they are just rendered to html string). I see some people use this ([ref1](https://mobile.twitter.com/jaquers/status/1160017562442305536) and [ref2](https://mobile.twitter.com/naholyr/status/1160101438468628480)). I see the use case for this, however this is fundamentally different from client side where the script is not executed. 

P.S.: I see that a lot of people were surprised by the fact that React doesn't execute scripts (me included) and this suggests that the use case is rare.

packages/react-dom/src/client/setInnerHTML.js
```diff
@@ -22,25 +24,42 @@ let reusableSVGContainer;
...
-    const svgNode = reusableSVGContainer.firstChild;
-    while (node.firstChild) {
-      node.removeChild(node.firstChild);
+  if (node.namespaceURI === Namespaces.svg) {
+    if (__DEV__) {
+      warningWithoutStack(
+        // Trusted Types undergo a change where window.TrustedTypes was renamed to window.trustedTypes
+        // (https://github.com/WICG/trusted-types/pull/205).
+        // $FlowExpectedError - TrustedTypes are defined only in some browsers or with polyfill
+        typeof TrustedTypes === 'undefined' &&
```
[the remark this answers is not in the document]
Siegrift: Yeah, sorry. I wasn't cautious enough.

packages/react-dom/src/client/ReactDOMComponent.js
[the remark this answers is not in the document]
gaearon: If we put the warning behind a feature flag (together with other code), that buys us time to consider the implications without making a final call now.

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -35,3 +37,42 @@ export function getToStringValue(value: mixed): ToStringValue {
       return '';
   }
 }
+
+/**
+ * Returns true only if Trusted Types are available in global object and the value is a trusted type.
+ */
+let isTrustedTypesValue: (value: any) => boolean;
+// $FlowExpectedError - TrustedTypes are defined only in some browsers or with polyfill
+if (typeof trustedTypes !== 'undefined') {
```
aweary: Lets add a check for `enableTrustedTypesIntegration` here as well, since we don't need this function if its disabled.

Going a bit further, it'd be ideal if `toStringOrTrustedType` would just export a function that only does string coercion when `enableTrustedTypesIntegration` is set to `false`, so we don't include unnecessary code in bundles built without the flag enabled.

packages/react-dom/src/client/ReactDOMComponent.js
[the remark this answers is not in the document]
Siegrift: I think it's ready :) Can you have a look if I didn't miss something?

packages/react-dom/src/client/ToStringValue.js
Siegrift: But this function is only used in `toStringOrTrustedType`. It shouldn't be included at all, because it will be never used and is also not exported.

gaearon: Someone can make a mistake later and bring it in. We tend to do feature flags at both call site and in the function.

The goal isn’t to ensure it’s “not included” but more to ensure that if someone calls it without a flag, it still has a reasonable behavior. Which might be throwing, returning some placeholder value, or whatever makes sense in a particular scenario.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -417,11 +422,20 @@ export function createElement(
         type,
       );
     }
-
     if (type === 'script') {
       // Create the script via .innerHTML so its "parser-inserted" flag is
       // set to true and it does not execute
       const div = ownerDocument.createElement('div');
+      if (__DEV__ && enableTrustedTypesIntegration && !didWarnScriptTags) {
```
gaearon: Nit: please keep DEV as its own block. And nest other conditions inside of it. This makes it easy to change logic inside of DEV block without accidentally moving some out when other conditions change.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -417,11 +422,20 @@ export function createElement(
...
+        warning(
+          false,
+          'Encountered a script tag while rendering React component. ' +
+            'Scripts inside React components are never executed when rendering' +
```
gaearon: Missing space at the end

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -417,11 +422,20 @@ export function createElement(
...
+            'on the client. Consider using tamplate tag instead ' +
```
gaearon: Type: should be “template”.

packages/react-dom/src/client/ToStringValue.js
gaearon: For example without the flag I’d just make it return false

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -35,3 +37,42 @@ export function getToStringValue(value: mixed): ToStringValue {
...
+ * We allow passing objects with toString method as element attributes or in dangerouslySetInnerHTML
+ * and we do validations that the value is safe. Once we do validation we want to use the validated
+ * value instead of the object (because object.toString may return something else on next call).
+ *
+ * If application uses Trusted Types we don't stringify trusted values, but preserve them as objects.
+ */
+export function toStringOrTrustedType(value: any): string | TrustedValue {
+  // fast-path string values as it's most frequent usage of the function
+  if (
+    enableTrustedTypesIntegration &&
```
gaearon: Nit: please keep feature flag in its own if condition. So that it’s easy to see static codepath forks. It’s fine to nest ifs when necessary.

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -35,3 +37,42 @@ export function getToStringValue(value: mixed): ToStringValue {
...
+ * Returns true only if Trusted Types are available in global object and the value is a trusted type.
+ */
+let isTrustedTypesValue: (value: any) => boolean;
+// $FlowExpectedError - TrustedTypes are defined only in some browsers or with polyfill
+if (typeof trustedTypes !== 'undefined') {
+  isTrustedTypesValue = (value: any) =>
+    trustedTypes.isHTML(value) ||
+    trustedTypes.isScript(value) ||
+    trustedTypes.isScriptURL(value) ||
+    // TrustedURLs are deprecated and will be removed soon: https://github.com/WICG/trusted-types/pull/204
```
gaearon: So we even want to support them then?

gaearon:
This is getting close. Exciting!

One more thing I’d like to ask you is to compare `yarn build --type=UMD_PROD --pretty react-dom` output in `build` folder before and after this PR. I’d expect the diff to be empty or minimal since the flag is off.

Also, do we have a server rendering story?

packages/react-dom/src/client/ToStringValue.js
Siegrift: Is it really necessary? I can see that you don't do this in the code base (for example here: https://github.com/facebook/react/blob/master/packages/react-dom/src/client/DOMPropertyOperations.js#L39). I think it will make the code too messy.

packages/react-dom/src/client/ToStringValue.js
Siegrift: I think at least for now we should. Both chrome and the polyfill currently support them and if we don't, trying out the integration will not work for the users (because the values will be stringified). They would need to create a default policy that would bless all URLs.

packages/react-dom/src/client/setInnerHTML.js
```diff
@@ -22,25 +25,38 @@ let reusableSVGContainer;
...
+          "You can try to wrap your svg element inside a div and use 'dangerouslySetInnerHTML' " +
+          'on the enclosing div instead.',
+      );
     }
-    while (svgNode.firstChild) {
-      node.appendChild(svgNode.firstChild);
+    if (!('innerHTML' in node)) {
+      reusableSVGContainer =
+        reusableSVGContainer || document.createElement('div');
+      reusableSVGContainer.innerHTML = '<svg>' + html.toString() + '</svg>';
```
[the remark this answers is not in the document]
Siegrift: Why should there be? If you have `'string' + variable`, js will call `toString` on that `variable` and not `valueOf`. Why do you think `valueOf` could be called?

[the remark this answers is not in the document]
koto: Actually Dan is right:
```
'' + {toString: () => '1', valueOf: () => '2'}
> '2'
```

`valueOf` is called implicitly earlier, in [ordinaryToPrimitive](http://www.ecma-international.org/ecma-262/#sec-ordinarytoprimitive). 

It's best to just check for this case. But I think `valueOf()` or `valueOf().toString()` would fix any inconsistencies.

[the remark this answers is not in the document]
Siegrift: Ahh, thanks @koto (sorry I musunderstood the spec).

However how do you mean `valueOf().toString()` would fix this? If you do `'' + ({toString: () => '1', valueOf: () => '2'}).valueOf().toString()` you get `"2"`.

I think its best to cast the `html` to `any` and (safely) preserve the original behavior.

packages/react-dom/src/client/DOMPropertyOperations.js
```diff
@@ -168,19 +170,21 @@ export function setValueForProperty(
     const {type} = propertyInfo;
     let attributeValue;
     if (type === BOOLEAN || (type === OVERLOADED_BOOLEAN && value === true)) {
+      // If attribute type is boolean, we know for sure it won't be an execution sink
+      // and we won't require Trusted Type here.
       attributeValue = '';
     } else {
       // `setAttribute` with objects becomes only `[object]` in IE8/9,
       // ('' + value) makes it output the correct toString()-value.
```
gaearon: Outdated comment

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -422,6 +427,18 @@ export function createElement(
       // Create the script via .innerHTML so its "parser-inserted" flag is
       // set to true and it does not execute
       const div = ownerDocument.createElement('div');
+      if (__DEV__) {
+        if (enableTrustedTypesIntegration && !didWarnScriptTags) {
```
gaearon: We’ve chatted more about this. What if when trusted types are on we actually execute the script if it’s a TT object? `<script>{ttObj}</script>`. As a new feature.

packages/react-dom/src/client/setInnerHTML.js
```diff
@@ -22,25 +25,39 @@ let reusableSVGContainer;
...
+          'Trusted Types enabled in an Internet Explorer will cause ' +
+          'the trusted value to be converted to string. Assigning string ' +
+          "to 'innerHTML' will throw an error if Trusted Types are enforced. " +
+          "You can try to wrap your svg element inside a div and use 'dangerouslySetInnerHTML' " +
+          'on the enclosing div instead.',
+      );
     }
-    while (svgNode.firstChild) {
-      node.appendChild(svgNode.firstChild);
+    if (!('innerHTML' in node)) {
```
gaearon: Does this need an else?

packages/react-dom/src/client/ReactDOMComponent.js
gaearon: Doesn’t need to be in this PR.

packages/react-dom/src/client/DOMPropertyOperations.js
Siegrift: Mhh, isn't it still valid? I mean that is the reason why we use `toStringOrTrustedType`.

packages/react-dom/src/client/setInnerHTML.js
Siegrift: Yeah :(

Sorry for keeping introducing a lot of these silly mistakes.

packages/react-dom/src/client/ReactDOMComponent.js
Siegrift: Yeah, however script would be executed only once so you would have to re-create it internally on every update.

packages/react-dom/src/client/DOMPropertyOperations.js
gaearon: Yeah but I mean since literally `('' + value)`  is not present, people might think this whole thing is irrelevant. But it's still relevant.

I think it's fine though.

packages/react-dom/src/client/setInnerHTML.js
```diff
@@ -22,25 +25,41 @@ let reusableSVGContainer;
...
-  if (node.namespaceURI === Namespaces.svg && !('innerHTML' in node)) {
-    reusableSVGContainer =
-      reusableSVGContainer || document.createElement('div');
-    reusableSVGContainer.innerHTML = '<svg>' + html + '</svg>';
-    const svgNode = reusableSVGContainer.firstChild;
-    while (node.firstChild) {
-      node.removeChild(node.firstChild);
+  if (node.namespaceURI === Namespaces.svg) {
+    if (enableTrustedTypesIntegration && __DEV__) {
+      warningWithoutStack(
```
gaearon: Let's use normal `warning` here? Stack would be nice.

packages/react-dom/src/client/__tests__/trustedTypes-test.internal.js
```diff
@@ -0,0 +1,105 @@
...
+      'Warning: Encountered a script tag while rendering React component. ' +
+        'Scripts inside React components are never executed when rendering ' +
+        'on the client. Consider using template tag instead ' +
+        '(https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template).\n' +
+        '    in script (at **)',
+    );
+
+    const spy = spyOnDev(console, 'error');
+    ReactDOM.render(<script>alert("I am not executed")</script>, container);
+    if (__DEV__) {
```
gaearon: Don't need the spy here. We already have a system that would fail the test if there's a warning without `toWarnDev` block around it.

Siegrift:
Resolved the minors :) looking forward to :ship:

gaearon:
Looks good, thanks.

gaearon:
@Siegrift @koto Can you please review this? https://github.com/facebook/react/pull/16795 I've made a few changes.

PR #16163, opened 2019-07-19
[Flare] Redesign core event system

trueadm:
This PR applies some core design changes as to how the React Flare event system works. This is a follow up to #16101 and #16122. This design tackles the problems already outlined in those PRs.

The design system now consists of two key parts:
- responders (created via `React.unstable_createResponder`)
- listeners (used via `React.unstable_useListener`)

Responders work pretty much like how they did before, except they're no longer consumed via Event Components, but are rather passed directly as an array to host components:

```jsx
import {PressResponder} from 'react-events/press';

<div responders={<PressResponder />} />
```

Host components can have multiple event responders, but any duplicate event responders are a warning and no-op. Event responders do not emit events, but they do accept optional props:

```jsx
import {PressResponder} from 'react-events/press';

<div responders={<PressResponder preventDefault={false} />} />
```

When you give a responder to a host component, there's a binding between that host component and the logic from within the relevant event responder implementation. The host component is also the "responder target" from there on too. Event responders act like a sort of event emitter, where they listen to native DOM events, but emit custom React Flare events.

Event listeners then pick up the React Flare events as the propagate through the React fiber tree. Event listeners are used exclusively in hook form:

```jsx
React.useListener(EventResponder, props)
```

Example:

```jsx
import {usePressListener, PressResponder} from 'react-events/press'l

function MyComponent() {
  usePressListener({ onPress: handlePress });

  function handlePress(e) {
    console.log('You pressed the div!');
  }

  return <div responders={<PressResonder />}>Press me!</div>
}
```

Events emitted from responders will terminate if they propagate the tree and hit another responder of the same type:

```jsx
function MyComponent() {
  // this won't get events from the inner div
  usePressListener({ onPress: handlePress });

  function handlePress(e) {
    console.log('You pressed the div!');
  }

  return (
    <div responders={<PressResonder />}>
        <div responders={<PressResonder />}>
          Inner div
        </div>
    </div>
  );
}
```

If the desired effect is to listen from events, an appropiate component boundary with a hook should be inserted, or another event responder should be used (i.e. `PressWithin`?).

If there are multiple responders in a sub-tree that can be intercepted by the same listener, like show:

```jsx
function MyComponent() {
  usePressListener({ onPress: handlePress });

  function handlePress(e) {
    console.log('You pressed the div!');
  }

  return (
    <React.Fragment>
      <div responders={<PressResonder />}>Div 1</div>
      <div responders={<PressResonder />}>Div 2</div>
    </React.Fragment>
  );
}
```

Then to only intercept a single event, the children should be broken into sub components that then listen to only the things it wants to be concerned about. It's also possible to determine the child being pressed by other mechanisms (`event.target` and maybe a future `event.responderProps`).

Compared to the previous event system design in Flare:

```jsx
  return (
    <Press onPress={onPress}>
      <Hover onHover={onHover}>
        <Focus onFocus={onFocus}>
          <div>
            {children}
          </div>
        </Focus>
      </Hover>
    </Press>
  )
}
```

With this PR and design, you would do this:

```jsx
  usePressResponder({onPress});
  useHoverResponder({onHover});
  useFocusResponder({onFocus});

  return (
    <div responders={[<PressResponder/ >, <FocusResponder />, <HoverResponder />]}>
      {children}
    </div>
  )
}
```

Note: there are no fibers created for this new event system design. Instead we attached objects to the existing fibers in the tree already, or use hooks.


TrySound:
Now it's less clear how to listen specific component for example in menu list and not really possible to "just wrap" any component. The component should support `responders` prop.

trueadm:
@TrySound How do you mean? You can wrap a component by making a new composite component that wraps whatever and uses hooks to listen. I showed an example above.

trueadm:
One question on the API for the `responders` prop. In this PR, you import the responder from the event module and then use it either as a function call or a reference to an object:

```jsx
// Reference to function object
<div responders={[PressResonder]} />

// Reference to object, generated from a function call
<div responders={[PressResonder({ ...someProps })]} />
```

This is made possible with the slightly confusing way I've made event modules export their responders:

```jsx
const PressResponderImpl = {
  ..
};

export function PressResponder(
  props: PressProps,
): {props: PressProps, responder: ReactDOMEventResponder} {
  return {
    props,
    responder: PressResponderImpl,
  };
}

PressResponder.props = {};
PressResponder.responder = PressResponderImpl;
```

I was thinking an alternative approach might be to use React Elements, where the `type` of the React Element is a direct reference to the `Responder` object itself (we'd need to add a `$$typeof` to the object, but this would be defined inline in the relevant module:


```jsx
<div responders={[<PressResponder />]} />
<div responders={[<PressResponder {...someProps} />]} />
```

We could even allow a single responder to be passed without the array syntax:

```jsx
<div responders={<PressResponder {...someProps} />} />
```

I'd like to hear thought from folks on what they prefer here. I'm leaning towards the React Element version, even if it comes with the cost of generating more objects at runtime.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -844,44 +815,38 @@ export function didNotFindHydratableSuspenseInstance(
...
-  // NO-OP, why might use this in the future
-}
-
-export function unmountEventComponent(
-  eventComponentInstance: ReactDOMEventComponentInstance,
+export function mountResponderInstance(
+  responder: ReactDOMEventResponder,
+  responderInstance: ReactDOMEventResponderInstance,
+  props: Object,
+  state: Object,
```
acdlite: Can we call this `stateNode` or something to distinguish it from React state? The key distinction to me is that it's mutable and lives outside React's update cycle

trueadm: `responderState`?

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -65,161 +66,94 @@ export function setListenToResponderEventTypes(
...
     eventPriority: EventPriority,
   ): void {
     validateResponderContext();
-    const {target, type, timeStamp} = possibleEventObject;
-
-    if (target == null || type == null || timeStamp == null) {
-      throw new Error(
-        'context.dispatchEvent: "target", "timeStamp", and "type" fields on event object are required.',
+    validateEventValue(eventValue);
+    // $FlowFixMe: Flow gets really confused with this line...
```
acdlite: It's because it's an opaque type. Let's make it not an opaque type, if we're going to compare them?

packages/react-dom/src/shared/ReactControlledValuePropTypes.js
```diff
@@ -36,7 +35,7 @@ if (__DEV__) {
         props.readOnly ||
         props.disabled ||
         props[propName] == null ||
-        enableFlareAPI
+        props.responders
```
acdlite: Can you explain why you removed the `enableFlareAPI` check? Wouldn't this affect someone who isn't using Flare but passes a `responders` prop to an input?

packages/shared/createEventResponder.js
```diff
@@ -0,0 +1,42 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ * @flow
+ */
+
+import type {ReactEventResponder} from 'shared/ReactTypes';
+import {REACT_RESPONDER_TYPE} from 'shared/ReactSymbols';
+
+export default function createEventResponder<E, C>(
+  displayName: string,
```
acdlite: Why is this a separate object instead of part of the config?

trueadm: So we ensure the hidden class is consistent, this improved perf in my microbenchmarks by around 15%.

packages/react-dom/src/shared/ReactControlledValuePropTypes.js
trueadm: Nice find, updated in latest commit.

packages/react-dom/src/events/DOMEventResponderSystem.js
trueadm: Done

packages/shared/createEventResponder.js
acdlite: Do we need the `displayName` in prod?

(Regardless, doesn't block this PR)

nathggns:
> Then to only intercept a single event, the children should be broken into sub components that then listen to only the things it wants to be concerned about

I appreciate that it probably simultaneously the case that it is too late and not the place to make bones about this detail, but I can see it leading to frustration/friction. React has been pretty good thus far about being unopinionated about the size of components. Can someone direct me to an explanation about why an API was chosen that requires you to break a component down to sub-components to attach two different event handlers of the same type to different host components? I know there’s some benefit or philosophy that I’m probably missing and I don’t want to unduly criticise when I don’t understand.

trueadm:
@nathggns It comes after research into how components are mostly used today with hooks. Given hooks are unconditiaonl, we never found a case internally where you wanted split the children. The size of components here isn't really the issue here, as you're not increasing size vs previous patterns (at least that I'm aware of).

PR #16164, opened 2019-07-19
Create ReactFabric.stopSurface and use that for bridgeless mode binding

ejanzer:
For Fabric we want to rename `unmountComponentAtNode` to `stopSurface` to be consistent with `startSurface`. Because `unmountComponentAtNode` is called from native code (see ReactInstanceManager), we can't just delete it yet. Once this change syncs, I'll clean up usage of `unmountComponentAtNode` in native code, then I'll come back and remove it completely.

For more context, see https://github.com/facebook/react-native/commit/305058178e144d4709606ea42d0edb9d864b33c5 (and additional discussion on D16273720 for fb employees). This PR also updates the binding I create in that diff for bridgeless mode to refer only to stopSurface.

### DIFF
--- a/scripts/rollup/shims/react-native/ReactFabric.js
+++ b/scripts/rollup/shims/react-native/ReactFabric.js
@@ -23,6 +23,10 @@ if (__DEV__) {
   ReactFabric = require('../implementations/ReactFabric-prod');
 }
 
-BatchedBridge.registerCallableModule('ReactFabric', ReactFabric);
+if (global.RN$Bridgeless) {
+  global.RN$stopSurface = ReactFabric.stopSurface;
+} else {
+  BatchedBridge.registerCallableModule('ReactFabric', ReactFabric);
+}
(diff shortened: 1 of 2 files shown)


scripts/rollup/shims/react-native/ReactFabric.js
```diff
@@ -23,6 +23,10 @@ if (__DEV__) {
   ReactFabric = require('../implementations/ReactFabric-prod');
 }
 
-BatchedBridge.registerCallableModule('ReactFabric', ReactFabric);
+if (global.RN$Bridgeless) {
```
ejanzer: I'm assuming that this file gets synced over, too?

bvaughn: Yes.

bvaughn:
Whoops sorry for forgetting to actually merge this.

PR #16166, opened 2019-07-20
[Scheduler] Use rIC to post first callback

acdlite:
Scheduler uses `requestAnimationFrame` to post tasks to the browser. If this happens at the beginning of a frame, the callback might not fire until the subsequent frame, even if the main thread is idle.

Our theory was that this wouldn't be an issue in practice, because once the first rAF fires, we schedule the next rAF as early as possible in that frame. Combined with our heuristic for computing frame deadlines, we shouldn't get any idle time in between frames — only before the *first* frame.

This reasoning holds true when you have a small number of large tasks, such as the ones posted by React. The idle time before the task starts is negligible relative to the lifetime of the entire task.

However, it does not hold if you have many small tasks posted over a span of time, perhaps spawned by a flurry of IO events. In this case, instead of single, contiguous rAF loop preceded by an idle frame, you get many rAF loops preceded by many idle frames. Our new theory is that this is contributing to poor main thread utilization during page loads.

To try to reclaim as much idle time as possible, this PR adds two experimental flags. The first one adds a `requestIdleCallback` call to start the rAF loop, which will fire before rAF if there's idle time left in the frame. (We still call rAF in case there isn't. We start working in whichever event fires first.)

The second flag tries a similar strategy using `setTimeout(fn, 0)`. If the timer fires before rAF, we'll assume that the main thread is idle.

If either `requestIdleCallback` or `setTimeout` fires before rAF, we'll immediately perform some work. Since we don't have a real frame time that we can use to compute the frame deadline, we might miss the next vsync, but this is fine because we can catch up during the next frame, by which point a real rAF will have fired and the loop can proceed the same way it does today.

Test plan: Try this on Facebook to see if it improves load times.

### DIFF
--- a/packages/scheduler/src/forks/SchedulerHostConfig.default.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.default.js
@@ -5,7 +5,11 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import {enableIsInputPending} from '../SchedulerFeatureFlags';
+import {
+  enableIsInputPending,
+  requestIdleCallbackBeforeFirstFrame as requestIdleCallbackBeforeFirstFrameFlag,
+  requestTimerEventBeforeFirstFrame,
+} from '../SchedulerFeatureFlags';
 
 // The DOM Scheduler implementation is similar to requestIdleCallback. It
 // works by scheduling a requestAnimationFrame, storing the time for the start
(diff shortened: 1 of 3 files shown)


packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -244,103 +207,157 @@ if (
...
+    prevRAFTime = rAFTime;
+    frameDeadline = rAFTime + frameLength;
+
+    port.postMessage(null);
   };
 
   requestHostCallback = function(callback) {
-    if (scheduledHostCallback === null) {
+    if (!isRAFLoopRunning) {
+      isRAFLoopRunning = false;
```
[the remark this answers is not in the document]
acdlite: Good catch, I fixed this locally but didn't commit it

packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -244,103 +207,157 @@ if (
...
+        if (!hasMoreWork) {
           scheduledHostCallback = null;
         }
       } catch (error) {
         // If a scheduler task throws, exit the current browser task so the
         // error can be observed, and post a new task as soon as possible
         // so we can continue where we left off.
-        isMessageEventScheduled = true;
-        port.postMessage(undefined);
+        port.postMessage(null);
```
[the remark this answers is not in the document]
acdlite: Not really

packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -244,103 +207,162 @@ if (
...
         throw error;
       }
-      // Yielding to the browser will give it a chance to paint, so we can
-      // reset this.
-      needsPaint = false;
     }
+    // Yielding to the browser will give it a chance to paint, so we can
+    // reset this.
+    needsPaint = false;
```
bvaughn: Is this still valid in the event that `performWorkUntilDeadline` gets called by timeout or rIC?

packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -244,103 +207,162 @@ if (
...
+            cancelIdleCallback(idleCallbackID);
+          }
+          frameDeadline = getCurrentTime() + frameLength;
+          performWorkUntilDeadline();
+        }, 0);
       }
     }
   };
 
   cancelHostCallback = function() {
```
bvaughn: Worth checking/clearing the timeout/rIC here?

packages/scheduler/src/forks/SchedulerHostConfig.default.js
acdlite: Not in that case but it is relevant in the case where `forceFrameRate` sets the frame length to an artificially large size. We should track the chosen frame length and the hardware frame length separately. Planning to do that in a follow up.

packages/scheduler/src/forks/SchedulerHostConfig.default.js
acdlite: `performWorkUntilDeadline` will exit immediately when `scheduledHostCallback` is null so it's fine. I wanted to keep the callback/timeout IDs local.

bvaughn: Yeah, I know it's a no-op in that case.

Seems reasonable.

Issue #16160, opened 2019-07-18
Bug: React DOM Server is mutating CSS variable names


rdhaliwal:
**Do you want to request a *feature* or report a *bug*?**

Bug

### Current Behaviour

 - `ReactDOMServer.renderToString` is mutating CSS Variable names that are inlined and in the `--camelCase` format into `--kebab-case`

### Expected Behaviour

 - `ReactDOMServer.renderToString` does not mutate the CSS Variable name, and keeps it in `--camelCase` format

### Demo

 - https://codesandbox.io/s/react-dom-server-debugging-28xsr

### Versions:

 - Issue appears in `react@16.8.6` &  `react-dom@16.8.6` 
 - Issue appears in `react@16.0.0` & `react-dom@16.0.0` 
 - (That's all I've tested so far)

### Other notes:

 - It's worth noting that `ReactDOM.render` is preserving the CSS Variable name and not mutating it. It only seems to only be an issue with `ReactDOMServer.renderToString`

bedakb:
I'd like to work on this one.

I think problem is the `createMarkupForStyles` function (`react-dom/server/ReactPartialRenderer.js`), because It runs `hyphenateStyleName` on custom css property so that's why CSS variable casing is mutated.

I assume solution should looks like this:

`serialized += delimiter + (isCustomProperty ? styleName : processStyleName(styleName)) + ':';`

PR #16167, opened 2019-07-20
Don't hyphenate custom CSS properties for ReactDOMServer

bedakb:
This PR fixes https://github.com/facebook/react/issues/16160

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRenderer.js
+++ b/packages/react-dom/src/server/ReactPartialRenderer.js
@@ -209,10 +209,11 @@ function validateDangerousTag(tag) {
 
 const styleNameCache = {};
 const processStyleName = function(styleName) {
+  const isCustomProperty = styleName.indexOf('--') === 0;
   if (styleNameCache.hasOwnProperty(styleName)) {
     return styleNameCache[styleName];
   }
-  const result = hyphenateStyleName(styleName);
+  const result = isCustomProperty ? styleName : hyphenateStyleName(styleName);
   styleNameCache[styleName] = result;
   return result;
 };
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -209,10 +209,11 @@ function validateDangerousTag(tag) {
 
 const styleNameCache = {};
 const processStyleName = function(styleName) {
+  const isCustomProperty = styleName.indexOf('--') === 0;
```
arianon: You likely meant to put this after the conditional return so it doesn't execute for every style name.

bedakb: thanks, resolved!

gaearon:
Please add a test to one of `ReactDOMServerIntegration` files. This will likely uncover that we still emit an incorrect warning. To fix the warning I think you'll need to make a similar fix [here](https://github.com/facebook/react/blob/2c4d61e1022ae383dd11fe237f6df8451e6f0310/packages/react-dom/src/shared/CSSPropertyOperations.js#L37).

bedakb:
@gaearon updated.
Hope It looks ok.

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -212,7 +212,8 @@ const processStyleName = function(styleName) {
   if (styleNameCache.hasOwnProperty(styleName)) {
     return styleNameCache[styleName];
   }
-  const result = hyphenateStyleName(styleName);
+  const isCustomProperty = styleName.indexOf('--') === 0;
```
gaearon: It's not ideal that we do this check again because `createMarkupForStyles` which calls us already does the same check anyway.

gaearon:
Looks good to me. Thanks.

PR #16168, opened 2019-07-21
[Shallow] Implement callEffects option to call effects from shallow renderer (#15275)

### LINKED ISSUE #15275: Add option in shallow renderer to run effects/componentDidUpdate/componentDidMount

ohjames:
Fixes the effects related part of #15275 by allowing the user to tell the shallow renderer to call effect hooks.

~It could do with a couple more tests but I wanted to get feedback on the approach and was also worried about putting any more time into it given that the maintainers have not responded to the corresponding issue in the months it's been open.~

Added a full set of tests.

### DIFF
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -349,7 +349,7 @@ class ReactShallowRenderer {
             cleanup: memoizedState.cleanup,
             run:
               inputs == null ||
-              shouldRunEffectsBasedOnInputs(memoizedState.inputs, inputs),
+              !areHookInputsEqual(inputs, memoizedState.inputs),
           };
         }
       }
@@ -688,7 +688,10 @@ class ReactShallowRenderer {
             ReactCurrentDispatcher.current = prevDispatcher;
           }
           this._finishHooks(element, context);
-          this._callEffectsIfDesired();
+          if (this._options.callEffects) {
+            this._callEffects(true);
+            this._callEffects(false);
+          }
         }
       }
     }
@@ -747,15 +750,6 @@ class ReactShallowRenderer {
     this._rendered = this._instance.render();
     // Intentionally do not call componentDidMount()
     // because DOM refs are not available.
-  }
-
-  _callEffectsIfDesired() {
-    if (!this._options.callEffects) {
-      return;
-    }
-
-    this._callEffects(true);
-    this._callEffects(false);
   }
 
   _callEffects(callLayoutEffects: boolean) {
@@ -920,17 +914,6 @@ function getMaskedContext(contextTypes, unmaskedContext) {
     context[key] = unmaskedContext[key];
   }
   return context;
-}
-
-function shouldRunEffectsBasedOnInputs(
-  before: Array<mixed> | void | null,
-  after: Array<mixed>,
-) {
-  if (before == null || before.length !== after.length) {
-    return true;
-  }
-
-  return before.some((value, i) => after[i] !== value);
 }
 
 export default ReactShallowRenderer;
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -175,12 +175,21 @@ function basicStateReducer<S>(state: S, action: BasicStateAction<S>): S {
   return typeof action === 'function' ? action(state) : action;
 }
 
+type ShallowRenderOptions = {
+  callEffects: boolean,
+};
+
+const isEffectHook = Symbol();
+
 class ReactShallowRenderer {
-  static createRenderer = function() {
-    return new ReactShallowRenderer();
+  static createRenderer = function(
+    options: ShallowRenderOptions = {callEffects: false},
+  ) {
+    return new ReactShallowRenderer(options);
   };
 
-  constructor() {
+  constructor(options: ShallowRenderOptions) {
+    this._options = options;
     this._reset();
   }
 
@@ -202,6 +211,7 @@ class ReactShallowRenderer {
     this._numberOfReRenders = 0;
   }
 
+  _options: ShallowRenderOptions;
   _context: null | Object;
   _newState: null | Object;
   _instance: any;
@@ -311,6 +321,54 @@ class ReactShallowRenderer {
       );
     };
 
+    const useGenericEffect = (
+      create: () => (() => void) | void,
+      inputs: Array<mixed> | void | null,
+      isLayoutEffect: boolean,
+    ) => {
+      this._validateCurrentlyRenderingComponent();
+      this._createWorkInProgressHook();
+
+      if (this._workInProgressHook !== null) {
+        if (this._workInProgressHook.memoizedState == null) {
+          this._workInProgressHook.memoizedState = {
+            isEffectHook,
+            isLayoutEffect,
+            create,
+            inputs,
+            cleanup: null,
+            run: true,
+          };
+        } else {
+          const {memoizedState} = this._workInProgressHook;
+          this._workInProgressHook.memoizedState = {
+            isEffectHook,
+            isLayoutEffect,
+            create,
+            inputs,
+            cleanup: memoizedState.cleanup,
+            run:
+              inputs == null ||
+              shouldRunEffectsBasedOnInputs(memoizedState.inputs, inputs),
+          };
+        }
+      }
+    };
+
+    const useEffect = (
+      create: () => (() => void) | void,
+      inputs: Array<mixed> | void | null,
+    ) => {
+      useGenericEffect(create, inputs, false);
+    };
+
+    const useLayoutEffect = (
+      create: () => (() => void) | void,
+      inputs: Array<mixed> | void | null,
+    ) => {
+      useGenericEffect(create, inputs, true);
+    };
+
     const useMemo = <T>(
       nextCreate: () => T,
       deps: Array<mixed> | void | null,
@@ -385,9 +443,9 @@ class ReactShallowRenderer {
         return readContext(context);
       },
       useDebugValue: noOp,
-      useEffect: noOp,
+      useEffect: this._options.callEffects ? useEffect : noOp,
       useImperativeHandle: noOp,
-      useLayoutEffect: noOp,
+      useLayoutEffect: this._options.callEffects ? useLayoutEffect : noOp,
       useMemo,
       useReducer,
       useRef,
@@ -630,6 +688,7 @@ class ReactShallowRenderer {
             ReactCurrentDispatcher.current = prevDispatcher;
           }
           this._finishHooks(element, context);
+          this._callEffectsIfDesired();
         }
       }
     }
@@ -688,6 +747,36 @@ class ReactShallowRenderer {
     this._rendered = this._instance.render();
     // Intentionally do not call componentDidMount()
     // because DOM refs are not available.
+  }
+
+  _callEffectsIfDesired() {
+    if (!this._options.callEffects) {
+      return;
+    }
+
+    this._callEffects(true);
+    this._callEffects(false);
+  }
+
+  _callEffects(callLayoutEffects: boolean) {
+    for (
+      let hook = this._firstWorkInProgressHook;
+      hook !== null;
+      hook = hook.next
+    ) {
+      const {memoizedState} = hook;
+      if (
+        memoizedState != null &&
+        memoizedState.isEffectHook === isEffectHook &&
+        memoizedState.isLayoutEffect === callLayoutEffects &&
+        memoizedState.run
+      ) {
+        if (memoizedState.cleanup) {
+          memoizedState.cleanup();
+        }
+        memoizedState.cleanup = memoizedState.create();
+      }
+    }
   }
 
   _updateClassComponent(
@@ -831,6 +920,17 @@ function getMaskedContext(contextTypes, unmaskedContext) {
     context[key] = unmaskedContext[key];
   }
   return context;
+}
+
+function shouldRunEffectsBasedOnInputs(
+  before: Array<mixed> | void | null,
+  after: Array<mixed>,
+) {
+  if (before == null || before.length !== after.length) {
+    return true;
+  }
+
+  return before.some((value, i) => after[i] !== value);
 }
 
 export default ReactShallowRenderer;
--- a/packages/react-test-renderer/src/ReactShallowRenderer.js
+++ b/packages/react-test-renderer/src/ReactShallowRenderer.js
@@ -336,7 +336,7 @@ class ReactShallowRenderer {
             isLayoutEffect,
             create,
             inputs,
-            cleanup: null,
+            destroy: null,
             run: true,
           };
         } else {
@@ -346,7 +346,7 @@ class ReactShallowRenderer {
             isLayoutEffect,
             create,
             inputs,
-            cleanup: memoizedState.cleanup,
+            destroy: memoizedState.destroy,
             run:
               inputs == null ||
               !areHookInputsEqual(inputs, memoizedState.inputs),
@@ -765,10 +765,10 @@ class ReactShallowRenderer {
         memoizedState.isLayoutEffect === callLayoutEffects &&
         memoizedState.run
       ) {
-        if (memoizedState.cleanup) {
-          memoizedState.cleanup();
+        if (memoizedState.destroy) {
+          memoizedState.destroy();
         }
-        memoizedState.cleanup = memoizedState.create();
+        memoizedState.destroy = memoizedState.create();
       }
     }
   }
--- a/packages/react-test-renderer/src/__tests__/ReactShallowRendererHooks-test.js
+++ b/packages/react-test-renderer/src/__tests__/ReactShallowRendererHooks-test.js
@@ -280,6 +280,81 @@ describe('ReactShallowRenderer with hooks', () => {
         'call effect',
       ]);
     });
+
+    it('should trigger effects and cleanup depending on inputs', () => {
+      let _setFriend;
+      const happenings = [];
+
+      function SomeComponent() {
+        const [friend, setFriend] = React.useState('Bons');
+        const [cat] = React.useState('Muskus');
+        _setFriend = setFriend;
+
+        React.useEffect(
+          () => {
+            happenings.push('call friend effect');
+            return () => {
+              happenings.push('cleanup friend effect');
+            };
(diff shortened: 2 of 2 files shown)


packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -91,6 +91,17 @@ function areHookInputsEqual(
...
 
+function shouldRunEffectsBasedOnInputs(
+  before: Array<mixed>,
+  after: Array<mixed> | void | null,
+) {
+  if (after == null || before.length !== after.length) {
+    return true;
+  }
+
+  return before.some((value, i) => (after: any)[i] !== value);
```
ohjames: flow makes me `: any` here as it thinks `after` can be `null` (despite the fact the function would have returned true already if that were the case.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -172,12 +183,21 @@ function basicStateReducer<S>(state: S, action: BasicStateAction<S>): S {
   return typeof action === 'function' ? action(state) : action;
 }
 
+type ShallowRenderOptions = {
+  callEffects: boolean,
+};
+
+const isEffectHook = Symbol();
```
ohjames: Ultra paranoia to prevent any other non effect hook from creating a memoized state that could look like an effect hook's memoized state.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -308,6 +329,54 @@ class ReactShallowRenderer {
...
+      inputs: Array<mixed> | void | null,
+      isLayoutEffect: boolean,
+    ) => {
+      this._validateCurrentlyRenderingComponent();
+      this._createWorkInProgressHook();
+
+      if (this._workInProgressHook !== null) {
+        if (this._workInProgressHook.memoizedState == null) {
+          this._workInProgressHook.memoizedState = {
+            isEffectHook,
```
ohjames: And here we see the ultra paranoid mode in effect.

packages/react-test-renderer/src/__tests__/ReactShallowRendererHooks-test.js
```diff
@@ -252,6 +252,36 @@ describe('ReactShallowRenderer with hooks', () => {
     expect(effectsCalled).toEqual([]);
   });
 
+  describe('when callEffects option is used', () => {
+    it('should trigger effects after render', () => {
+      let happenings = [];
```
ohjames: I called it `happenings` rather than `events` to make it obvious these are not DOM events or anything like that.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -308,6 +329,54 @@ class ReactShallowRenderer {
...
+              inputs == null ||
+              shouldRunEffectsBasedOnInputs(inputs, memoizedState.inputs),
+          };
+        }
+      }
+    };
+
+    const useEffect = (
+      create: () => (() => void) | void,
+      inputs: Array<mixed> | void | null,
```
ohjames: I believe `? Array<mixed>` is an equivalent type, but everywhere else I see the `useEffect` being specified, the type is written like this, so I followed suit.

bdwain:
thanks for figuring this out! If you want to copy the logic for componentDidMount and componentDidUpdate, you can see it at #15589

ohjames:
Completed the tests now.

ohjames:
@bdwain I thought it would be good to handle the life-cycle events in a different PR. Since all my code is using hooks now, I'm not so interested in them. Plus it's easy to call them from outside if you need them, I see that enzyme already does that.

euroclydon37:
Lol. I need this in my testing life.

ohjames:
Closed by accident, sorry.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -308,6 +318,54 @@ class ReactShallowRenderer {
...
+          this._workInProgressHook.memoizedState = {
+            isEffectHook,
+            isLayoutEffect,
+            create,
+            inputs,
+            destroy: null,
+            run: true,
+          };
+        } else {
+          const {memoizedState} = this._workInProgressHook;
```
ssunday: IMO, this is style, but I prefer the spaces around the brackets, ex: `const { memoizedState }`

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -619,6 +677,10 @@ class ReactShallowRenderer {
             ReactCurrentDispatcher.current = prevDispatcher;
           }
           this._finishHooks(element, context);
+          if (this._options.callEffects) {
+            this._callEffects(true);
```
ssunday: This basically just toggles it on and off—is there anyway to condense it into one func?

packages/react-test-renderer/src/ReactShallowRenderer.js
bdwain: react uses prettier right? it should make these decisions for us

ssunday: Yep, I see it now. It's set this way. Sorry!

packages/react-test-renderer/src/ReactShallowRenderer.js
ohjames: @ssunday that's not what it does, first it runs the layout effects, then the effects.

ssunday: Sorry, I misread some of the logic. Trying add comments to something very good. So eager to get this going!

ssunday:
@ohjames Thank you so much for doing this! I'm not as familiar with the react code base (clearly from my comments) but if there's anything you need help on to get this going let me know 😬So eager to have this functionality in place.

packages/react-test-renderer/src/ReactShallowRenderer.js
bdwain: Same!

ohjames:
@ssunday Nothing more we can really do for this until the maintainers have a look. They haven't commented on any of the related issues and they've been open for months. It's possible to get this working in your own extension of this library by hooking into private methods, that's what I'm doing for now. Enzyme could possibly do the same.

ssunday:
@ohjames Yeah, it'll take a while. With a draftjs change I was following, took a decent amount of time. But it happened!

For the hooking into private methods, could you link a gist or some examples of how to do that you could add to your PR comment? I think that helped with this [draftjs](https://github.com/facebook/draft-js/pull/2035) PR— provide a way for people to rest in the wild and confirm? Or post to the Enzyme issue related to this. If it's not too difficult?

nommuna:
This addition would help immensely when testing hooks with enzyme. Hopefully this gets added soon. 

@ohjames what do you mean by “hooking into private methods”?  Right now I am trying to find a good way to test private methods under a functional component and I’m not having any luck with it.

ljharb:
Functional components can't have "methods"; that's a term that applies only to classes. If you mean, functions created inside the function, or only accessible via closure, you can't and shouldn't directly test those.

bdwain:
yea i think @ohjames was referring to private methods on the shallow renderer itself

ohjames:
@ljharb I mean you can make the react test renderer support effects if you override one of it's "private" (underscore prefixed) methods and then flush the hooks after each call to render.

ljharb:
ah, sure, if hooking into fully public methods can be a reliable mechanism for enzyme, that would suffice pending the proper PRs being merged and released :-)

ohjames:
@ljharb I wouldn't call an underscore prefixed method "fully" public because the intention is private. But... well they didn't use a symbol or # to make it completely private ;)

ljharb:
@ohjames a symbol is also fully public, because it's also accessible from the outside :-) only closed-over variables and private fields are actually "private" in javascript.

ohjames:
@ljharb It can be done by overriding `_createDispatcher` so if that method would be acceptable to enzyme I can hook it in. Is there someone I should be @-ing from the react team to get some 👀 on this? Seems to be quite a lot of interest, but not a single issue on this topic has received a comment from a maintainer in 5 months. Maybe they would have dropped in to tell us it won't be accepted and not to waste any effort on it?

Joonpark13:
> Is there someone I should be @-ing from the react team to get some 👀 on this? Seems to be quite a lot of interest, but not a single issue on this topic has received a comment from a maintainer in 5 months. Maybe they would have dropped in to tell us it won't be accepted and not to waste any effort on it?

Just wanted to chime in and say I strongly feel the same way - it would be at least good to get a "hey we won't be considering this".

gaearon:
Sorry I haven't replied earlier!

Generally we've been using shallow renderer less and less at Facebook because it encourages brittle tests that verify implementation detail. So components that relied on it a lot introduced a lot of refactoring friction, and didn't catch valuable regressions anyway.

The main value proposition of shallow rendering is that you mock out the inner components. We see the value in this, but in our experience doing this at the module system level is more flexible and gives you more leeway when refactoring. For example, you can always `jest.mock('Button', () => 'button')` to replace a child `Button` component with a mock. We've mostly been happy with this approach.

Since we don't use the shallow renderer much anymore, we don't feel like we would be good stewards of its API, or to even recommend it. Additionally, Enzyme has been adding different behavior to it (such as calling class commit phase lifecycles) that we didn't originally plan. At least some of it was due to a misunderstanding (for example, React intends to guarantee refs are resolved in the commit phase, but running code with shallow renderer forces you to handle null gracefully).

At this point I think it would be better if we could move out the shallow renderer out of this repository. It's stagnating here, getting fixes very late, and causes frustration like in this PR. By design, it doesn't use almost any of the React code, so it should be easy to move.

If @davidmarkclements were kind to offer the `react-shallow-renderer` package name, we could externalize it there, and maybe it could be maintained in the Enzyme repository or elsewhere instead.

Would anyone like to propose a plan? Thanks.

ljharb:
That sounds like a great plan, as long as the react team can provide some kind of commitment to expose API as needed, and ideally include tests as part of core - so that enzyme (and the shallow renderer) doesn’t unintentionally break with every release.

gaearon:
What API do you expect to need? Shallow renderer doesn't depend on any other code so I don't expect it to need private APIs. The only one I can think of is the way Hooks Dispatcher is exposed from the `react` package — but that isn't Enzyme-specific.

gaearon:
We'd be happy to run some tests against a separate package fwiw. Similar to how we have integration tests with `create-react-class`.

ohjames:
The shallow renderer uses the following modules from "shared" that I don't think are externally reachable:

```js
import describeComponentFrame from 'shared/describeComponentFrame';
import getComponentName from 'shared/getComponentName';
import shallowEqual from 'shared/shallowEqual';
import invariant from 'shared/invariant';
import ReactSharedInternals from 'shared/ReactSharedInternals';
import is from 'shared/objectIs';
import type { ReactContext, ReactEventResponderListener } from 'shared/ReactTypes';
import type {ReactElement} from 'shared/ReactElementType';
```

ljharb:
Note that I've previously manually extracted `shallowEqual` as https://www.npmjs.com/package/enzyme-shallow-equal, and `getComponentName` as https://unpkg.com/browse/airbnb-prop-types@2.14.0/build/helpers/getComponentName.js - but it'd be *much* better if I could deprecate both of those to use official React packages, or alternatively, if React could use those instead of its current implementation.

Also, `is` exists as https://npmjs.com/object-is (same thing tho, it'd be ideal if React just used that package).

LBroy:
> Sorry I haven't replied earlier!
> 
> Generally we've been using shallow renderer less and less at Facebook because it encourages brittle tests that verify implementation detail. So components that relied on it a lot introduced a lot of refactoring friction, and didn't catch valuable regressions anyway for us. (Your experience might differ.)
> 
> The main value proposition of shallow rendering is that you mock out the inner components. We see the value in this, but in our experience doing this at the module system level gives you more leeway when refactoring. For example, you can always `jest.mock('Button', () => 'button')` to replace a child `Button` component with a mock. We've mostly been happy with this approach and found it more flexible.
> 
> Since we don't use the shallow renderer much anymore, we don't feel like we would be good stewards of its API, or even can recommend it. Additionally, Enzyme has been adding different behavior to it (such as calling class commit phase lifecycles) that we didn't originally plan. At least some of it was due to a misunderstanding (for example, React intends to guarantee refs are resolved in the commit phase, but running code with shallow renderer forces you to handle null gracefully). But there is also some mismatch between how Enzyme uses it internally, and how we initially thought it would be used.
> 
> **At this point I think it would be better if we could move out the shallow renderer out of this repository.** It's stagnating here, getting fixes very late, and causes frustration like in this PR. By design, it doesn't use almost any of the React code, so it should be easy to move. On the other hand, Enzyme is in active development and might as well want to take ownership of this piece.
> 
> If @davidmarkclements were kind to offer the `react-shallow-renderer` package name, we could externalize it there, and maybe it could be maintained in the Enzyme repository or elsewhere instead.
> 
> Would anyone like to propose a plan? Thanks.

What do you recommend using instead?

ljharb:
I believe part of the mismatch here is that Facebook largely/historically doesn't do serverside rendering, which is where enzyme (and the shallow renderer) is critical. I'm not aware of any scalable alternatives.

ohjames:
@ljharb @gaearon How about just renaming `_createDispatcher` to `createDispatcher`. By making it part of the public interface of the class, we can hook into it and implement whatever effects we want from outside?

In the long term, to extract the shallow renderer, couldn't the `shared` package just be published?

But please consider implementing my first proposal. In the react-native world the shallow renderer is very popular, not being able to use (test) hooks makes us very sad.

davidmfoley:
This would be quite useful to me and to other experienced TDD practitioners who work with react.

In addition to being on the order of 100x faster to run as compared to tests that require a stubbed DOM (i.e. JSDOM), shallow rendering allows for test-driving the designs of components much more efficiently/directly than the equivalent virtual DOM tests. Certain classes of components (for example: Providers that encapsulate state machines and provide to their children via context) are much, much easier to test-drive with shallow approaches. 

Please consider merging this or otherwise exposing hooks to easily integrate effects into shallow test tools like enzyme in a way that is reliable and relatively future-proof.

kumar303:
> Generally we've been using shallow renderer less and less at Facebook because it encourages brittle tests 

@gaearon just curious, is that because those specific components don't have static typing (like Flow) to enforce that nested components get instantiated correctly? Also, it's sometimes tricky in Flow to truly enforce component signature checks, e.g. when using HOCs.

It's true that shallow rendering won't expose any problems like passing invalid props but static typing *can* solve this.

> The main value proposition of shallow rendering is that you mock out the inner components. We see the value in this, but in our experience doing this at the module system level gives you more leeway when refactoring. For example, you can always `jest.mock('Button', () => 'button')` to replace a child `Button` component with a mock. We've mostly been happy with this approach and found it more flexible.

Huh. To me, it seems way more flexible to rely on shallow rendering for this because the test doesn't have to *know* which dependencies to mock. This is important for refactoring; a test should not break when you add/remove nested components since they are not relevant to the main component under test -- nested components already have their own responsibilities.

I also like how shallow rendering encourages the test author to *only* consider functionality that the component under test is responsible for.


> Since we don't use the shallow renderer much anymore, we don't feel like we would be good stewards of its API, or even can recommend it. 

That makes sense, I was just curious to understand why, in case I'm missing some flaws in shallow rendering.

thedanchez:
Any updates on this idea of creating a `react-shallow-renderer` package that can be maintained in isolation or within Enzyme?

davidmarkclements:
@gaearon btw I'm happy to donate the package name when you guys are ready.
I'd also like to talk to someone about contributing https://github.com/esxjs/esx to the project in some way - any pointers?

rmlevangelio:
This will make unit tests with React.16 easier

thedanchez:
Following up on this on a weekly basis to make sure things don't go stale. Any updates?

pyyding:
I'm late to the party but would like to say testing is important hence this pull request is important. Please merge. Thx 😊

ohjames:
I believe that Godspeed You! Black Emperor is far more important than this PR and you should go listen to them and stop relying on those soulless motherfuckers at Facebook to give a damn about you. Yours Sincerely, James Pike. https://m.youtube.com/watch?v=2c9TyeKnGGk

Fieel:
@ohjames, to be honest, this response is borderline infuriating and I don't even know what to reply. Wow.

gaearon:
@ohjames I recognize you're frustrated with the lack of movement on this but doing personal attacks on people working on this repository is not acceptable.

As I said in my earlier comment:

>At this point I think it would be better if we could move out the shallow renderer out of this repository.

There is literally nothing that would prevent you from doing so and collaborating with e.g. Enzyme on it. The shallow renderer can be copy pasted, published in its own package, and be developed there.

I'm sorry that you were waiting for us to merge this PR, but I thought we set the expectation clear right above by saying we don't intend to invest more into shallow renderer and expanding its API surface.

gaearon:
>I believe part of the mismatch here is that Facebook largely/historically doesn't do serverside rendering, which is where enzyme (and the shallow renderer) is critical. 

This is not correct btw; we do rely heavily on server rendering for the new Facebook.

gaearon:
I'm going to lock this PR because the discussion has turned in an unconstructive direction. Additionally, @ljharb said he can't comment on the PR for some reason (maybe a GitHub bug).

I opened a new issue to discuss this in https://github.com/facebook/react/issues/17321. I welcome everyone from this thread (but please let's keep this one civil).

PR #16169, opened 2019-07-21
Do not warn when DOM attribute values are undefined

### LINKED ISSUE #13381: React.cloneElement cannot remove existing props

A feature.

When an element is cloned with [React.cloneElement](https://reactjs.org/docs/react-api.html#cloneelement), it's possible to add new props or modify existing ones, but not to remove existing props. 

[Relevant code](https://github.com/facebook/react/blob/v16.8.0-alpha.1/packages/react/src/ReactElement.js#L333).

Example of how it works right now:

```
const element = React.createElement("a", {href: "http://github.com"});
const newElement = React.cloneElement(element, {href: undefined});
console.log(newElement.props); // {href: undefined}
```

**What is the desired behavior?**

It would be great to add some way to remove props (passing `undefined` as value?):

```
const element = React.createElement("a", {href: "http://github.com"});
const newElement = React.cloneElement(element, {href: undefined});
console.log(newElement.props); // {}
```

I guess I could use directly `React.createElement` but, AFAIK, I'll have also to worry about special attributes like `key` and `ref`. I'd rather not mess with internals.

I think it has worked this way in all React versions.

**What's your use case?**

I am applying a map transformation of elements recursively and I need to remove some virtual props before passing the real elements for React to render.  Console shows `React does not recognize the [unknownProp] prop on a DOM element` for those props, I'd want to avoid that.

stephan-noel:
Fixes #13381

Currently (React 16.8.6), passing `null` or `undefined` on a ReactDOMElement will result in that attribute being absent from the corresponding DOM element, however, the following warning still appears:

```
class App extends Component {
  render() {
    return <div helloWorld={undefined}> Hello World</div>;
  }
}
```
image

As explained in the referenced issue, this leads to warnings in dev environments when trying to remove props by using `cloneElement` , despite the prop not appearing in the actual DOM.

This pull request takes the approach @sophiebits recommended in that issue and will not give the error, only when the value of the prop is `undefined` (ie, when the prop is `null` or some other value, the previous behavior will be maintained).

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMAttribute-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMAttribute-test.js
@@ -126,5 +126,14 @@ describe('ReactDOM unknown attribute', () => {
 
       expect(el.firstChild.getAttribute('helloworld')).toBe('something');
     });
+
+    it('does not warn on camelCase unknown attributes with value of undefined', () => {
+      const el = document.createElement('div');
+
+      ReactDOM.render(<div helloWorld={undefined} />, el);
+
+      expect(el.firstChild.getAttribute('helloworld')).toBeNull();
+      expect(el.firstChild.getAttribute('helloWorld')).toBeNull();
+    });
   });
 });
--- a/packages/react-dom/src/__tests__/ReactDOMAttribute-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMAttribute-test.js
@@ -133,7 +133,6 @@ describe('ReactDOM unknown attribute', () => {
       ReactDOM.render(<div helloWorld={undefined} />, el);
 
       expect(el.firstChild.getAttribute('helloworld')).toBeNull();
-      expect(el.firstChild.getAttribute('helloWorld')).toBeNull();
     });
   });
 });
--- a/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
@@ -2375,8 +2375,6 @@ describe('ReactDOMComponent', () => {
 
     it('allows cased custom attributes and does not warn when value is undefined', function() {
       const el = ReactTestUtils.renderIntoDocument(<div fooBar={undefined} />);
-
-      expect(el.getAttribute('fooBar')).toBeNull();
       expect(el.getAttribute('foobar')).toBeNull();
     });
 
--- a/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
@@ -2373,6 +2373,13 @@ describe('ReactDOMComponent', () => {
       expect(el.getAttribute('foobar')).toBe('true');
     });
 
+    it('allows cased custom attributes and does not warn when value is undefined', function() {
+      const el = ReactTestUtils.renderIntoDocument(<div fooBar={undefined} />);
+
+      expect(el.getAttribute('fooBar')).toBeNull();
+      expect(el.getAttribute('foobar')).toBeNull();
+    });
+
     it('warns on NaN attributes', function() {
       let el;
       expect(() => {
--- a/packages/react-dom/src/shared/ReactDOMUnknownPropertyHook.js
+++ b/packages/react-dom/src/shared/ReactDOMUnknownPropertyHook.js
@@ -151,7 +151,7 @@ if (__DEV__) {
         warnedProperties[name] = true;
         return true;
       }
-    } else if (!isReserved && name !== lowerCasedName) {
+    } else if (value !== undefined && !isReserved && name !== lowerCasedName) {
       // Unknown attributes should have lowercase casing since that's how they
       // will be cased anyway with server rendering.
       console.error(


trueadm:
I think this would likely be a breaking change for React 16.x. This might be something we can do, but put behind a flag, to be released in a later version of React. The typical cases where this accidently happens is when people spread props, i.e. `<div {...someProps }>`. In those cases, they may have no realized that they were passing through and invalid prop and regardless of the value being `undefined` or not, they'd expect the existing behaviour of React 16.x to warn.

sophiebits:
We generally wouldn’t consider removing a warning to be a breaking change, as it’s very unlikely to break someone’s production apps.

trueadm:
@sophiebits Unfortuantely, that would be a breaking change internally for how we make with SSR warnings errors in DEV.

stephan-noel:
Thanks for the quick reply!  In that case, I will wait for the results of the sync before putting it behind a flag.

packages/react-dom/src/__tests__/ReactDOMAttribute-test.js
```diff
@@ -126,5 +126,14 @@ describe('ReactDOM unknown attribute', () => {
...
       expect(el.firstChild.getAttribute('helloworld')).toBe('something');
     });
+
+    it('does not warn on camelCase unknown attributes with value of undefined', () => {
+      const el = document.createElement('div');
+
+      ReactDOM.render(<div helloWorld={undefined} />, el);
+
+      expect(el.firstChild.getAttribute('helloworld')).toBeNull();
+      expect(el.firstChild.getAttribute('helloWorld')).toBeNull();
```
rodoabad: Quick question for context, why are you asserting it twice?

stephan-noel: Sorry about the late response @robarnold .  I'm actually not asserting the same thing twice, either here or in the other test.  Here for example, the "W" is capitalized in one but not the other.  

The reason for this is that if the value of the attribute is not undefined or null, the attribute will appear in the DOM, but lower-cased, not camel-cased as it is in the JSX.

image


image

I didn't want the test to depend on this detail however.

I suppose my test doesn't need to bother with this, and could avoid this detail by just using an attribute name that consists of a single, lower-cased word. 

What do you think?

yuanoook:
Why should we skip warning for undefined and null here?
helloWorld is not a recognized prop on an element, I think there should be a warning for this case, no matter what the value is.

stephan-noel:
@yuanoook , please see [sophibits's comment here](https://github.com/facebook/react/issues/13381#issuecomment-458335157) in the referenced issue.  I want to warn if `null `is passed as a value (as other tests verify) but I do not want to warn if `undefined `is passed as a value.  

In the tests, I am asserting `toBeNull()` because for some reason, even if you pass in `undefined`, `getAttribute` will return `null`.

I think `not.toWarnDev()` is clearer, but that approach of prefixing with `not `was not working and I don't think that custom matcher was meant to be used like that.

packages/react-dom/src/__tests__/ReactDOMAttribute-test.js
stephan-noel: @rodoabad ,  I made the changes I mentioned in my response so that  it won't be as confusing and consistent with the other tests.  I will resolve the conversation, please feel free to make more comments.  Thanks for the input.

stephan-noel:
@trueadm , any update on this issue, or anything further I can do to make this PR ready?

stephan-noel:
@eps1lon , I suppose it's an easy check to make, but [here](https://github.com/facebook/react/issues/13381#issuecomment-458335157) @sophiebits states 

> We try to treat undefined the same as absent basically everywhere in React

I don't think the same is true for `null` and so I think we should get some feedback from the core maintainers  about whether doing the same for `null` is a good idea.

gaearon:
A few counter-arguments that leave me hesitant about this:

- Now that most people fixed these warnings, muting it for a subset of cases feels like a step back. Now you're not gonna notice if you pass something unnecessary down the line while it's undefined. But it can unexpectedly show up when it's not undefined! Maybe in prod only (e.g. based on prod data) in an surprising way.
- The `<Operator>` use case in https://github.com/facebook/react/issues/13381#issuecomment-515672242 is the kind of API we generally discourage. It's confusing you can't insert a div or some layout component in the middle without breaking the logic. So maybe it's good that the warning adds some friction to this pattern.
- That said, I don't think even for that use case that the warning is inevitable. You can write `let {propForOperation, ...rest} = props` inside `Button` and then render `<button {...props} />`. No need for `cloneElement` or cloning/deleting props.
- In long term, we want to eventually make props opaque on elements. So `cloneElement` justification is dubious for that too. The original state use case of "I am applying a map transformation of elements recursively" is also not something we'd like to encourage in long term because it prevents optimizations.

So all in all, I'm struggling to find a compelling use case for this, but I see a few downsides due to less warning coverage. What are your thoughts?

stephan-noel:
> Now that most people fixed these warnings, muting it for a subset of cases feels like a step back. Now you're not gonna notice if you pass something unnecessary down the line while it's undefined. But it can unexpectedly show up when it's not undefined! Maybe in prod only (e.g. based on prod data) in an surprising way.

That's a good point.

> The <Operator> use case in #13381 (comment) is the kind of API we generally discourage. It's confusing you can't insert a div or some layout component in the middle without breaking the logic. So maybe it's good that the warning adds some friction to this pattern.

I tried to make the example more generic, but actually the API is more like:

```
<Sorter srortType={'asc'}>
      <CheckBoxFilter sortValue={'size'} />
      <div sortValue={'price'} > {price} </div>
      <ButtonFilter sortValue={'color'} />
      <Button sortValue={'brand'} />
</Sorter>
```
The Sorter component just changes their order.  I want to be able to sort the filter orders by simply passing a different value to sortType and maybe even allow sortType to be a function that will sort based on the sortValue.  The logic is maintained here if you insert any component into the middle I think, so it wouldn't be discouraged in this case, right?  Or maybe I am missing something?

> That said, I don't think even for that use case that the warning is inevitable. You can write let {propForOperation, ...rest} = props inside Button and then render <button {...props} />. No need for cloneElement or cloning/deleting props.

Yes, you are right, I would have to wrap the <div/> in a custom component however and if Button is from a 3rd party library, I would also have to wrap it as well even if that component will do nothing else but to prevent sortValue from being spread further down.  It's a bit inconvenient and not too elegant but it works.  So I guess this would be the only downside I see.


> In long term, we want to eventually make props opaque on elements. So cloneElement justification is dubious for that too. The original state use case of "I am applying a map transformation of elements recursively" is also not something we'd like to encourage in long term because it prevents optimizations.

Just to clarify, by "opaque" do you mean that when accessing props.children, you will not be able to access its props?  For example:

```
const MyComponent = (props) => {

  const firstChild = React.Children.toArray(props.children)[0];

  // Error cannot access props.foo, it is opaque
  const fooPropOfFirstChild = firstChild.props.foo;

}
```

If so, I'm curious as to the reasons.  Can you point me to more information on this and the reasoning behind that goal because I see this when looking at code in various well-known projects?


Overall, the potential danger of the downsides (especially the first point) seems to outweigh the upsides of less wrapping.

gaearon:
I think we're not going to do this. If it's spread down, that's a problem by itself. Then the fact that it's sometimes `undefined` just masks the problem.

PR #16172, opened 2019-07-22
[Flare] Ensure mouse events can use target to validate press

trueadm:
This PR fixes an interesting bug that we encountered internally using the `Press` responder. When using mouse events (via pointer events or mouse events), we have a guarantee that the event target will be that of the one we're moving/releasing from (unlike touch events, which can have the same target as when the press was initiated). This mechanic for mouse events lets us bypass the `getBoundingClientRect` logic, which is:

- expensive as it requires a layout reflow
- can be incorrect in cases where children of the `responderTarget` are absolutely positioned
- can sometimes report `width` and/or `height` as `0` when some CSS styles are applied

This behaviour already existed in the current Press implementation. But there were two bugs:

- `isTargetWithinNode` wasn't taking into account the `alternate` fiber
- we weren't setting `isPressWithinResponderRegion` to `true` if we knew the mouse event target was in the responder region (by using the targets).

This fixes both cases, which also fixes the internal bug we were seeing.

### DIFF
--- a/packages/react-events/src/dom/Press.js
+++ b/packages/react-events/src/dom/Press.js
@@ -837,20 +837,22 @@ const PressResponder: ReactDOMEventResponder = {
         }
         const pressTarget = state.pressTarget;
 
-        if (
-          pressTarget !== null &&
-          !targetIsDocument(pressTarget) &&
-          (pointerType !== 'mouse' ||
-            !context.isTargetWithinNode(target, pressTarget))
-        ) {
-          // Calculate the responder region we use for deactivation, as the
-          // element dimensions may have changed since activation.
-          updateIsPressWithinResponderRegion(
-            touchEvent || nativeEvent,
-            context,
-            props,
-            state,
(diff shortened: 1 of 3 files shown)


packages/react-events/src/dom/Press.js
```diff
@@ -837,20 +837,22 @@ const PressResponder: ReactDOMEventResponder = {
...
-          // element dimensions may have changed since activation.
-          updateIsPressWithinResponderRegion(
-            touchEvent || nativeEvent,
-            context,
-            props,
-            state,
-          );
+        if (pressTarget !== null && !targetIsDocument(pressTarget)) {
+          if (
+            pointerType === 'mouse' &&
```
necolas: Why is this checking for only mouse pointers?

trueadm: That's what it was doing before. Touch event pointers have the same target until released.

PR #16179, opened 2019-07-23
[Flare] Press: fix keyboard interactions

necolas:
Prevents Spacebar from scrolling the window.
Prevents Enter from triggering a navigation if `preventDefault` is true.
Fixes the emulated events unit test.

Demo: https://codesandbox.io/s/admiring-ride-rz96j30rp


packages/react-events/src/dom/Press.js
```diff
@@ -680,6 +680,8 @@ const PressResponder: ReactDOMEventResponder = {
       case 'mousedown':
       case 'touchstart': {
         if (!isPressed) {
+          state.shouldPreventClick = false;
```
trueadm: We want to move this down, otherwise we'll set this during simulated events too (which we didn't do before).

PR #16181, opened 2019-07-23
[act] with nested act()s from different renderers, flush work on exiting outermost act()

threepointone:
Given this snippet:
```jsx
TestRenderer.act(() => {
  TestUtils.act(() => {
    TestRenderer.create(<Effecty />);
  });
});
```
We want to make sure that all work is only flushed on exiting the outermost act().

Now, naively doing this based on actingScopeDepth would work with a mocked scheduler, where flushAll() would flush all work across renderers.

This doesn't work without mocking the scheduler though; and where flushing work only works per renderer. So we disable this behaviour for a non-mocked scenario. This seems like an ok tradeoff.

### DIFF
--- a/packages/react-noop-renderer/src/createReactNoop.js
+++ b/packages/react-noop-renderer/src/createReactNoop.js
@@ -600,6 +600,8 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
   // ReactTestUtilsAct.js, ReactTestRendererAct.js, createReactNoop.js
 
   let hasWarnedAboutMissingMockScheduler = false;
+  const isSchedulerMocked =
+    typeof Scheduler.unstable_flushAllWithoutAsserting === 'function';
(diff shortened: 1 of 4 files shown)


packages/react-noop-renderer/src/createReactNoop.js
```diff
@@ -770,7 +775,12 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
 
       // flush effects until none remain, and cleanup
       try {
-        if (actingUpdatesScopeDepth === 1) {
+        if (
+          actingUpdatesScopeDepth === 1 &&
+          (isSchedulerMocked === false ||
+            (isSchedulerMocked === true &&
```
acdlite: `isSchedulerMocked === true` always evaluates to true here because `isSchedulerMocked === false` already evaluated to false.

PR #16186, opened 2019-07-23
Edit Suspense Priority Warning Message

lunaruan:
move 'component that triggered the update' in suspense priority warning message to the beginning of the message instead of the middle for better readability

Message after update:
```
Warning: The components that triggered the update: COMPONENT_NAME

The following components suspended during a user-blocking update: COMPONENT_NAME

Updates triggered by user interactions (e.g. click events) are considered user-blocking by default. 
They should not suspend. Updates that can afford to take a bit longer should be wrapped with 
`Scheduler.next` (or an equivalent abstraction). This typically includes any update that shows new 
content, like a navigation.
````

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -2639,30 +2639,39 @@ function flushSuspensePriorityWarningInDEV() {
 
       componentsThatTriggeredHighPriSuspend = null;
 
-      const componentThatTriggeredSuspenseError =
-        componentsThatTriggeredSuspendNames.length > 0
-          ? 'The components that triggered the update: ' +
-            componentsThatTriggeredSuspendNames.sort().join(', ') +
-            '\n\n'
-          : '';
+      const componentNamesString = componentNames.sort().join(', ');
+      let componentThatTriggeredSuspenseError = '';
+      if (componentsThatTriggeredSuspendNames.length > 0) {
+        componentThatTriggeredSuspenseError =
+          'The following components triggered a user-blocking update:' +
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by acdlite** [comment]
Here's my suggested text:

```
The following components triggered a user-blocking update:

  ParentComponentThatUpdated

that was then suspended by:

  ComponentThatDependsOnNetworkData, AnotherOneOfThose, AndAnother

The fix is to split the update into multiple parts: a user-blocking update to
provide immediate feedback, and another update that triggers the bulk of
the changes.

Refer to the documentation for useSuspenseTransition to learn how to implement
this pattern.
```

**Comment by acdlite** [comment]
If it was triggered by the root, then the message should be:

```
A user-blocking update was suspended by:

  ComponentThatDependsOnNetworkData, AnotherOneOfThose, AndAnother

The fix is to split the update into multiple parts: a user-blocking update to
provide immediate feedback, and another update that triggers the bulk of
the changes.

Refer to the documentation for useSuspenseTransition to learn how to implement
this pattern.
```

and we'll add a separate warning inside `ReactDOM.render` et al, as discussed previously.

nuragic:
@acdlite @lunaruan 👋 👀  I'm getting this message but I can't find any reference to `useSuspenseTransition`... any tips? Thanks!

gaearon:
Are you using Concurrent Mode as an unstable API? It’s not officially supported.

nuragic:
@gaearon I know it's unstable yet, but we're using it already and it's good enough as far as we can see... we have some use case in which it improved performance a lot. Just wondering if I can find any reference about this warning, orherwise it's fine... I'll wait :)

PR #16194, opened 2019-07-24
Fix suspenseCallback type warning, add a test

bgirard:
Sorry for the churn. Added a test.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1330,7 +1330,9 @@ function commitSuspenseComponent(finishedWork: Fiber) {
         suspenseCallback(new Set(thenables));
       }
     } else if (__DEV__) {
-      warning(false, 'Unexpected type for suspenseCallback.');
+      if (suspenseCallback !== undefined) {
+        warning(false, 'Unexpected type for suspenseCallback.');
+      }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1329,7 +1329,7 @@ function commitSuspenseComponent(finishedWork: Fiber) {
       if (thenables !== null) {
         suspenseCallback(new Set(thenables));
       }
-    } else if (__DEV__) {
+    } else if (__DEV__ && suspenseCallback != null) {
```
gaearon: We try to use explicit comparisons instead of `!=`. Actually do we even want to support `null`? Maybe let's check for just `undefined`.

We also try to separate `__DEV__` checks from others. Maybe:

```js
else if (__DEV__) {
  if (suspenseCallback !== undefined) {
    warning(...)
  }
}
```

sebmarkbage: We usually check undefined and null separately (partly to avoid document.all being considered).

trueadm: Is the __DEV__ on a separate line still something we want? I was pulled up on it a few months ago and asked to move it inline (forgot where and who), but GCC appropriately handles it.

sebmarkbage: This was more a legacy pattern when we didn't trust the compiler. We trust GCC now though so we don't have to put it on a separate line anymore.

However, I often still separate it out for clarity. I found it hard to see what the scope of deleted things are when they're on the same line. Especially if it's not clear like `else if (__DEV__ && something || bar)`.

I've also seen cases where the order of execution is wrong like:

```
if (foo() && __DEV__) { // wrong order
}
```

So separating it out helps understanding whether the `foo()` call will remain or not.

PR #16196, opened 2019-07-24
update legacy context warning message

threepointone:
The link in the legacy context message doesn't point to anything related to context. This changes the link to point to https://fb.me/react-legacy-context, which points to https://reactjs.org/docs/context.html#legacy-api. Also adds a line that it'll probably be gone later.


sebmarkbage:
Maybe we should add a one liner to the old page too since many people might see the old one?

ahtee:
Is there a reason why it is a `fb.me` link instead of just the `reactjs.org` link?

PR #16198, opened 2019-07-25
[Scheduler] Test browser implementation details

acdlite:
The Scheduler implementation uses browser APIs like `MessageChannel`, `requestAnimationFrame`, and `setTimeout` to schedule work on the main thread. Most of our tests treat these as implementation details; however, the sequence and timing of these APIs are not precisely specified, and can vary wildly across browsers.

To prevent regressions, we need the ability to simulate specific edge cases that we may encounter in various browsers.

This adds a new test suite that mocks all browser methods used in our implementation. It assumes as little as possible about the order and timing of events. The only thing it assumes is that requestAnimationFrame is passed a frame time that is equal to or less than the time returned by performance.now. Everything else can be controlled at will.

It also includes Scheduler-specific invariants, e.g. only one rAF callback can be scheduled at a time.

It overlaps slightly with the existing SchedulerDOM-test suite, which also mocks the browser APIs, but exposes a higher-level set of testing primitives. I will consolidate the two suites in a follow-up.

### DIFF
--- a/packages/scheduler/src/__tests__/SchedulerBrowser-test.js
+++ b/packages/scheduler/src/__tests__/SchedulerBrowser-test.js
@@ -1,0 +1,350 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails react-core
+ * @jest-environment node
+ */
+
+/* eslint-disable no-for-of-loops/no-for-of-loops */
+
+'use strict';
+
+let Scheduler;
+let runtime;
+let performance;
+let scheduleCallback;
+let NormalPriority;
+
+// The Scheduler implementation uses browser APIs like `MessageChannel`,
+// `requestAnimationFrame`, and `setTimeout` to schedule work on the main
+// thread. Most of our tests treat these as implementation details; however, the
+// sequence and timing of these APIs are not precisely specified, and can vary
+// wildly across browsers.
+//
+// To prevent regressions, we need the ability to simulate specific edge cases
+// that we may encounter in various browsers.
+//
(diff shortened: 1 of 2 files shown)


packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -271,7 +261,12 @@ if (
     };
     rAFTimeoutID = setTimeout(onTimeout, frameLength * 3);
 
-    if (prevRAFTime !== -1) {
+    if (
+      prevRAFTime !== -1 &&
+      // Make sure this rAF time is different from the previous one. This check
+      // could fail if two rAFs fire in the same frame.
+      rAFTime - prevRAFTime > 0.1
```
acdlite: This particular change fixes a bug in Safari where we post a rAF from inside another rAF, and that rAF fires within the same frame. One of the tests I added covers this scenario.

packages/scheduler/src/__tests__/SchedulerBrowser-test.js
```diff
@@ -0,0 +1,350 @@
...
+    function advanceTimeToNextFrame() {
+      const targetTime =
+        Math.ceil(currentTime / VSYNC_INTERVAL) * VSYNC_INTERVAL;
+      if (targetTime === currentTime) {
+        currentTime += VSYNC_INTERVAL;
+      } else {
+        currentTime = targetTime;
+      }
+    }
+    function getMostRecentFrameNumber() {
```
bvaughn: trivial naming nit: `getCurrentFrameNumber`? "Most recent" feels like an odd name to use when `currentTime` is exactly at a frame boundary.

**Edit** I see how you're using this method in the tests below, and `getMostRecentFrameNumber` makes more sense as a name. I think the usage in `fireAnimationFrame` feels more like a "current" value but I don't feel strongly about this. :smile:

PR #16204, opened 2019-07-25
[Flare] Adds Keyboard event responder

trueadm:
This adds the Keyboard event responder. It's mostly just a pass-through of `keyDown` and `keyUp`, with polyfills and normalization for browsers that don't fully support the right `event.key` value.

### DIFF
--- a/packages/react-events/src/dom/Keyboard.js
+++ b/packages/react-events/src/dom/Keyboard.js
@@ -1,0 +1,206 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import type {
+  ReactDOMResponderEvent,
+  ReactDOMResponderContext,
+} from 'shared/ReactDOMTypes';
+
+import React from 'react';
+import {DiscreteEvent} from 'shared/ReactTypes';
+
+type KeyboardEventType = 'keydown' | 'keyup';
+
+type KeyboardListenerProps = {|
+  onKeyDown: (e: KeyboardEvent) => void,
+  onKeyUp: (e: KeyboardEvent) => void,
+|};
+
+type KeyboardResponderProps = {
+  disabled: boolean,
+};
+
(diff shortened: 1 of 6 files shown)


packages/react-events/src/dom/Keyboard.js
```diff
@@ -0,0 +1,206 @@
...
+  timeStamp: number,
+|};
+
+const targetEventTypes = ['keydown', 'keyup'];
+
+/**
+ * Normalization of deprecated HTML5 `key` values
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#Key_names
+ */
+const normalizeKey = {
```
necolas: Which browsers still expose the deprecated names?

packages/react-events/src/dom/Keyboard.js
```diff
@@ -0,0 +1,206 @@
...
+  Del: 'Delete',
+  Win: 'OS',
+  Menu: 'ContextMenu',
+  Apps: 'ContextMenu',
+  Scroll: 'ScrollLock',
+  MozPrintableKey: 'Unidentified',
+};
+
+/**
+ * Translation from legacy `keyCode` to HTML5 `key`
```
necolas: Which browsers is this targeting?

packages/react-events/src/dom/Keyboard.js
trueadm: I took this code from:

https://github.com/facebook/react/blob/master/packages/react-dom/src/events/getEventKey.js

I remember when working on React Fire that not supporting this broke many internal web driver tests, which use an older version of FF.

packages/react-events/src/dom/Keyboard.js
trueadm: We can probably remove this bit now actually, but keep the other one.

trueadm: Removed this in latest commit.

packages/react-events/src/dom/Keyboard.js
necolas: Ok we should probably revisit this because to cut down on legacy code

PR #16205, opened 2019-07-25
[Flare] Adds useListener implementation to ReactDebugHooks

trueadm:
This adds the implementation of `useListener` to `ReactDebugHooks`.

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -226,8 +226,9 @@ function useListener(
   hookProps: ?Object,
 ): void {
   const listenerProps = hookProps || emptyObject;
+  // Don't put the actual event responder object in, just its displayName
   const value = {
-    responder,
+    responder: responder.displayName,
     props: listenerProps,
   };
   hookLog.push({primitive: 'Listener', stackError: new Error(), value});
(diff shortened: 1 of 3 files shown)


packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -215,8 +221,16 @@ function useMemo<T>(
   return value;
 }
 
-function useListener() {
-  throw new Error('TODO: not yet implemented');
+function useListener(
+  responder: ReactEventResponder<any, any>,
+  hookProps: ?Object,
+): void {
+  const listenerProps = hookProps || emptyObject;
```
bvaughn: Any reason we couldn't just use `props: null` in the `hookLog` ?

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -215,8 +221,17 @@ function useMemo<T>(
...
+  // Don't put the actual event responder object in, just its displayName
+  const value = {
+    responder: responder.displayName,
```
bvaughn: Should we set a default/fallback name (we do elsewhere for event components)

trueadm: Event responders should always have a mandatory part of the `unstable_createResponder` API. I can add that though!

bvaughn: Ah! Okay well if we're _guaranteed_ to have a name then no need. I just wasn't sure, based on the precedent in the renderer interface. 😄

trueadm: I added it anyway, to be safe!

PR #16207, opened 2019-07-25
Warn when rendering tests in concurrent/sync mode without a mocked scheduler

threepointone:
Concurrent/Sync mode tests should always be run with a mocked scheduler (v17 or not). This PR adds a warning for the same. I'll put up a separate PR to the docs with a page detailing how to mock the scheduler.


acdlite:
I don't think we're going to call it "batched mode" publicly. We use that term now to distinguish it from  the legacy synchronous mode.

But in the future, only "batched" and concurrent modes will exist. In that world, "batched" doesn't make sense because concurrent mode is also batched. So our current plan for the naming of the three modes is:

- **Legacy mode** – `ReactDOM.render`
- **Sync mode** - `ReactDOM.createSyncRoot`
- **Concurrent mode** - `ReactDOM.createRoot`

I sometimes refer to the two synchronous modes as "legacy sync" and "batched sync" if there's ambiguity.

threepointone:
Noted, changed "Batched" for "Sync" in the messages.

threepointone:
I moved the one in the act() definition into this new one, much prefer it this way.

packages/react-dom/src/__tests__/ReactTestUtilsAct-test.internal.js
```diff
@@ -7,22 +7,93 @@
  * @emails react-core
  */
 
+let React;
```
threepointone: This test suite doesn’t have anything to do with act anymore, will rename

threepointone:
renamed some stuff for consistency 
warnAboutMissingMockScheduler -> warnAboutUnmockedScheduler
warnIfNotMockedScheduler -> warnIfUnmockedScheduler
didWarnAboutNotMockedScheduler -> didWarnAboutUnmockedScheduler

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2522,6 +2526,37 @@ function warnIfNotCurrentlyActingUpdatesInDEV(fiber: Fiber): void {
 
 export const warnIfNotCurrentlyActingUpdatesInDev = warnIfNotCurrentlyActingUpdatesInDEV;
 
+// in tests, we want to enforce a mocked scheduler.
+let didWarnAboutUnmockedScheduler = false;
+
+export function warnIfUnmockedScheduler(fiber: Fiber) {
+  if (__DEV__) {
+    if (didWarnAboutUnmockedScheduler === false) {
+      if (fiber.mode & BatchedMode || fiber.mode & ConcurrentMode) {
```
threepointone: should we trigger the warning for legacy sync mode + strict mode too?

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -47,6 +48,9 @@ import {
   scheduleSyncCallback,
 } from './SchedulerWithReactIntegration';
 
+// The scheduler is imported here *only* to detect whether it's been mocked
+import * as Scheduler from 'scheduler';
```
threepointone: I could have also exported `isSchedulerMocked` from `SchedulerWithReactIntegration.js` if we want to prevent this import here.

threepointone:
ping!

acdlite:
I find it weird that most developers will never interact with Scheduler, but we're asking them to mock it.

If we provide them with a testing build of React DOM (and the other renderers), would that replace the need to mock the Scheduler package? I think it would in the normal case. The exception is if your framework interacts with Scheduler directly (instead of via built-in APIs like `useSuspenseTransition`).

Btw, this doesn't affect 16.9, right? Since both `createRoot` and `createSyncRoot` are unstable.

threepointone:
So you’re right about your points in the short term. As you said, just the testing build will not be sufficient *in the future*, because a consumer may well use the scheduler directly via scheduler.next, etc. So in the future, we’ll end up asking them to mock the scheduler anyway. 

The reason I raised this PR, was to help early adopters with their tests. Assuming we won’t do another release for a few months, this could at least guide the shape of their tests and make it easier when we *do* make changes and demand a testing build etc. So let’s say, 6 months out, when we do ship, at least their tests wouldn’t break horribly. If we had testing builds right now, we could’ve suggested that, but this seemed like low effort-moderate win to be had now. Most, like 99%, of folks won’t see this warning, but the ones that do, tests will still work as expected for them.

threepointone:
(This isn’t really blocking 16.9, happy to punt, but figured it would be a nice-to-have)

threepointone:
btw, the linked doc (https://github.com/reactjs/reactjs.org/pull/2171, preview at https://deploy-preview-2171--reactjs.netlify.com/docs/mock-scheduler.html) tries to make the role of the scheduler clearer

packages/react-reconciler/src/ReactFiberWorkLoop.js
threepointone: I think it's a good idea to do so, since folks will use strict mode as a stepping stone to concurrent mode. changing.

threepointone: maybe not. and we can't tell when rendering if a child tree will be in strict mode.

threepointone:
pending item: land the doc PR, and set up the fb.me link. Will do.

PR #16210, opened 2019-07-25
Warn when Using DefaultProps on Function Components

lunaruan:
As part of the process to deprecate defaultProps on function components (as per a larger proposal outlined [here](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md)), add a warning whenever someone does this.


packages/react-dom/src/__tests__/ReactFunctionComponent-test.js
```diff
@@ -365,7 +386,11 @@ describe('ReactFunctionComponent', () => {
     );
   });
 
+  // TODO: deprecate default props support after we remove defaultProps
+  // from function components
```
gaearon: Didn't we plan to keep them for classes forever?

lunaruan: Oh sorry, poor phrasing. I just meant that I need to change this test at some point. Will rephrase!

packages/react-dom/src/__tests__/ReactFunctionComponent-test.js
```diff
@@ -25,6 +26,8 @@ describe('ReactFunctionComponent', () => {
     React = require('react');
     ReactDOM = require('react-dom');
     ReactTestUtils = require('react-dom/test-utils');
+    ReactFeatureFlags = require('shared/ReactFeatureFlags');
+    ReactFeatureFlags.warnAbouDefaultPropsOnFunctionComponents = true;
```
threepointone: s/`warnAbouDefaultPropsOnFunctionComponents`/`warnAboutDefaultPropsOnFunctionComponents` (note the extra t at the end of `About`)

packages/react-dom/src/__tests__/ReactFunctionComponent-test.js
```diff
@@ -25,6 +26,8 @@ describe('ReactFunctionComponent', () => {
...
   });
```
threepointone:
```suggestion

  afterEach(() => {
    ReactFeatureFlags.warnAboutDefaultPropsOnFunctionComponents = false;
  })
```

packages/react-dom/src/__tests__/ReactFunctionComponent-test.js
```diff
@@ -206,6 +209,24 @@ describe('ReactFunctionComponent', () => {
...
+    }
+
+    FunctionalComponent.defaultProps = {
+      testProp: true,
+    };
+
+    expect(() =>
+      ReactTestUtils.renderIntoDocument(<FunctionalComponent />),
+    ).toWarnDev(
+      'Warning: FunctionalComponent: Function components do not support defaultProps. ' +
```
ljharb: *do not*, or *will not* in a future version?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1424,6 +1427,23 @@ function validateFunctionComponentInDev(workInProgress: Fiber, Component: any) {
...
+    warnAbouDefaultPropsOnFunctionComponents &&
+    Component.defaultProps !== undefined
+  ) {
+    const componentName = getComponentName(Component) || 'Unknown';
+
+    if (!didWarnAboutDefaultPropsOnFunctionComponent[componentName]) {
+      warningWithoutStack(
+        false,
+        '%s: Function components do not support defaultProps. ' +
+          'Use Javascript default arguments instead.',
```
ljharb:
```suggestion
          'Use JavaScript default arguments instead.',
```

packages/react-dom/src/__tests__/ReactFunctionComponent-test.js
```diff
@@ -206,6 +209,24 @@ describe('ReactFunctionComponent', () => {
...
+        'Use Javascript default arguments instead.',
```
ljharb:
```suggestion
        'Use JavaScript default arguments instead.',
```

acdlite:
Tests suites that modify feature flags need an `internal` suffix.

The reason is that we have a CI job that runs our unit tests against the actual npm packages. `shared/ReactFeatureFlags` doesn't exist in those packages, because it gets inlined by our build pipeline.

acdlite:
Need to update all the feature flag files to include the new one you added. E.g. `ReactFeatureFlags.native-oss.js`. That's why the Flow and build CI jobs are failing.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1424,6 +1427,23 @@ function validateFunctionComponentInDev(workInProgress: Fiber, Component: any) {
...
+    warnAboutDefaultPropsOnFunctionComponents &&
+    Component.defaultProps !== undefined
+  ) {
+    const componentName = getComponentName(Component) || 'Unknown';
+
+    if (!didWarnAboutDefaultPropsOnFunctionComponent[componentName]) {
+      warningWithoutStack(
+        false,
+        '%s: Function components will not support defaultProps. ' +
+          'Use JavaScript default arguments instead.',
```
FredyC: Might be worth mentioning it's ES2015 feature ... or not. 🤷‍♂

packages/react-dom/src/__tests__/ReactDeprecationWarnings-test.internal.js
```diff
@@ -0,0 +1,46 @@
...
+
+    FunctionalComponent.defaultProps = {
+      testProp: true,
+    };
+
+    expect(() =>
+      ReactTestUtils.renderIntoDocument(<FunctionalComponent />),
+    ).toWarnDev(
+      'Warning: FunctionalComponent: Function components will not support defaultProps. ' +
+        'Use JavaScript default arguments instead.',
```
acdlite: Alt suggestion for first sentence: "Support for `defaultProps` will be removed from function components in a future release."

ljharb: ideally, a future *major* release

acdlite: Right, I assume that was implied but should probably be specific

Sawtaytoes:
I don't know why we're deprecating `defaultProps` on function components. I mean, I get the concept, but please look at this:

image

The second and third examples are functionally the same, but the first is different. If React renders `<MyComponent />` again, it will create a new default object for `value`.

Am I understanding this correctly?

What happens if you have `useEffect` calls like this:
```js
const [
  localValues,
  setLocalValues,
] = useState(values)

useEffect(
  () => {
    setLocalValues(
      values
    )
  },
  [values],
)

return (
  <div>
    {values.text}
  </div>
)
```

The first example would call the `useEffect` more times than you expect whereas the latter two only call it the one time (at mount).  While you may think "a few extra Virtual DOM diffs is no big deal", it will lead to many unexpected bugs in hooks because you're writing with the intention that props, which were never passed, are never changing.

I myself have run into this issue as well and kept wondering why my values kept getting overridden when `children` updated even though I wasn't passing any props.

FredyC:
@Sawtaytoes If you are into endless debate on this topic, be sure to read through https://github.com/reactjs/rfcs/pull/107. I am sure you will find your answers there.

gaearon:
The PR is for the implementation so any discussion here will get lost.

Sawtaytoes:
Where should I move the discussion? The RFC place?

gaearon:
Sure.

uloco:
When this gets depreacted, is there still a way to set the defaultProps globally for example the react-native <Text> element? Overriding default props of existing components from react / react-native, which I don't have control of is where this is really good to have it available as an option..

uloco:
It's exactly what I am talking about. I want the default `<Text />` element for android to always have the property `android_hyphenationFrequency` to be set to  `full`. Because I want all text elements throughout the app to have this behaviour. Should I really create a custom component for this?

Until now it works (in react 17) that you override it globally with `Text.defaultprops.android_hypenationFrequency = "full"`. This is also a common pattern for font scaling, because the defaults are just not that great here.

For me it feels like a weird pattern to have a different custom component for a behaviour that is so basic as this and needed in every single place always.

You could warn the user to not use this in components they create themselves, but being able to customize someone elses defaults would be nice imho...

ljharb:
Yes, you really should. Don’t mutate objects you don’t own.

PR #16212, opened 2019-07-25
Optimize objectIs

ku8ar:
Native Object.is implementation is much faster than polyfill.

~~https://jsperf.com/object-is-vs-polyfill/1~~

https://jsperf.com/object-is-vs-polyfill-better/1

Zrzut ekranu 2019-07-25 o 22 43 15

### DIFF
--- a/packages/shared/objectIs.js
+++ b/packages/shared/objectIs.js
@@ -17,4 +17,4 @@ function is(x: any, y: any) {
   );
 }
 
-export default is;
+export default typeof Object.is === 'function' ? Object.is :  is;
--- a/packages/shared/objectIs.js
+++ b/packages/shared/objectIs.js
@@ -17,6 +17,4 @@ function is(x: any, y: any) {
   );
 }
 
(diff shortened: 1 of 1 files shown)


intellild:
图片
Looks like safari's native Object.is is a disaster

ku8ar:
@probablyup @intellil OK, I know why so weak. I wrote the wrong test. Safari is very slow when reading the value of the `Object` object.

I have improved test and it is much better:
https://jsperf.com/object-is-vs-polyfill-better/1

Zrzut ekranu 2019-08-9 o 19 17 07

No difference on Safari and Firefox, but a great improvement on Chrome.
Probably the good Chrome result is due to some micro-optimization, during which Blink remembers the Object.is result when it gets the same values. But for React this means an advantage, because in most cases rerender is with same properties.

ku8ar:
@acdlite  hi! What do you think?

ku8ar:
Is there anybody out there?
@gaearon @trueadm

PR #16214, opened 2019-07-25
[scheduler] Yield many times per frame, no rAF

acdlite:
Adds experimental flag to yield many times per frame using a message event loop, instead of the current approach of guessing the next vsync and yielding at the end of the frame.

This new approach forgoes a `requestAnimationFrame` entirely. It posts a message event and performs a small amount of work (5ms) before yielding to the browser, regardless of where it might be in the vsync cycle. At the end of the event, if there's work left over, it posts another message event.

This should keep the main thread responsive even for really high frame rates. It also shouldn't matter if the hardware frame rate changes after page load (our current heuristic only detects if the frame rate increases, not decreases).

The main risk is that yielding more often will exacerbate main thread contention with other browser tasks.

I'm also not sure to what extent message events are throttled when the tab is backgrounded, relative to `requestAnimationFrame` or `setTimeout`. I'm starting with the assumption that `message` events fire with at least the same priority as timers, but I'll need to confirm.

Let's try it and see.


132yse:
Why not use setTimeOut 0 directly? What the difference between setTimeOut and Message Channel?

fxxjdedd:
> Why not use setTimeOut 0 directly? What the difference between setTimeOut and Message Channel?
Maybe this is the answer:
https://stackoverflow.com/questions/18826570/settimeout0-vs-window-postmessage-vs-messageport-postmessage

yisar:
@fxxjdedd Thanks for your link. but in this link, `window.postMessage` is better than `Message channel`
This is interesting. I still don't have a good answer.
https://github.com/facebook/react/pull/14234
see this PR.

li-kai:
@yisar I found a [stackoverflow post](https://stackoverflow.com/a/18832977) and a [mapbox PR](https://github.com/mapbox/mapbox-gl-js/pull/8677) that may explain why this approach was taken.

SyMind:
Why not use Promise, Promise polyfill was originally used in React? Now queueMicroTask is living standard, why not use it?

jddxf:
@SyMind Microtasks are continuously executed until the microtask queue on the event loop is empty which means a higher-pri task such as one for user interaction would be blocked by React.

franckchen:
> Why not use setTimeOut 0 directly? What the difference between setTimeOut and Message Channel?

We prefer MessageChannel because of the 4ms setTimeout clamping

beMySun:
> > Why not use setTimeOut 0 directly? What the difference between setTimeOut and Message Channel?
> 
> We prefer MessageChannel because of the 4ms setTimeout clamping

https://github.com/facebook/react/releases/tag/v16.10.0

lizuncong:
> > > Why not use setTimeOut 0 directly? What the difference between setTimeOut and Message Channel?
> > 
> > 
> > We prefer MessageChannel because of the 4ms setTimeout clamping
> 
> https://github.com/facebook/react/releases/tag/v16.10.0

I have discussed which APIs are suitable for task scheduling：https://github.com/lizuncong/mini-react/issues/22

beMySun:
Hello!  Your E-mail has been recieved and I will respond you as soon as possible.Best wishes!

PR #16215, opened 2019-07-26
[Bugfix] Check tag before calling hook effects

acdlite:
Adds a missing tag check before calling passive effects.

The reason we didn't catch this earlier is because it happens to accidentally work even without the check in most cases.

I added a regression test but it's very specific and weird and you should feel free to delete it later if the implementation details change and it causes you problems. I split this into the first commit to show that the test fails.

The second commit adds the tag check, to show that it fixes the test.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -55,10 +55,12 @@ import {
   clearCaughtError,
 } from 'shared/ReactErrorUtils';
 import {
+  NoEffect,
   ContentReset,
   Placement,
   Snapshot,
   Update,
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1953,7 +1953,15 @@ export function flushPassiveEffects() {
       resetCurrentDebugFiberInDEV();
     } else {
       try {
-        commitPassiveHookEffects(effect);
+        switch (effect.tag) {
```
gaearon: Not so easy. There's also DEV above behind `invokeGuardedCallback`.

PR #16217, opened 2019-07-26
Warn when Using String Refs

lunaruan:
As part of the process to deprecate string refs (as per a larger proposal outlined [here](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md)), expand string ref warning to fire for everyone who uses string refs (not just people on strict mode) to notify them of planned deprecation changes.


packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -114,21 +115,36 @@ function coerceRef(
...
   ) {
     if (__DEV__) {
-      if (returnFiber.mode & StrictMode) {
+      // TODO: Clean this up once we turn on the string ref warning for
+      // everyone, because the strict mode case will no longer be relevant
+      if (returnFiber.mode & StrictMode || warnAboutStringRefs) {
         const componentName = getComponentName(returnFiber.type) || 'Component';
-        if (!didWarnAboutStringRefInStrictMode[componentName]) {
+        if (!didWarnAboutStringRefs[componentName]) {
+          let stringRefWarning;
```
threepointone: can I recommend against constructing the message here like this? This part of the message, which is arguably more critical than the part after, won't be extracted into WARNINGS, and isn't as readable as it could be. Instead, I recommend having 2 calls to warningWithoutStack. Even though the text is duplicated, since they're right next to each other, we can be pretty sure we'll edit both if needed. As an example from another warning PR I'm working on https://github.com/facebook/react/pull/16207/files#diff-63c455475b8566c10993b2daa2c3211bR2535-R2555

packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -114,21 +115,38 @@ function coerceRef(
...
-            getStackByFiberInDevAndProd(returnFiber),
-          );
-          didWarnAboutStringRefInStrictMode[componentName] = true;
+        if (!didWarnAboutStringRefs[componentName]) {
+          if (warnAboutStringRefs) {
+            warningWithoutStack(
+              false,
+              'Component "%s" contains the string ref "%s". Support for string refs ' +
+                'will be removed in a future major release. We recommend using ' +
+                'createRef() instead.' +
```
acdlite: Maybe mention `useRef` instead? Or at least also.

gaearon: Yeah it’s a common mistake to put createRef into a function component

Aquinas74:
class Welcome extends React.Component {
  render() {
    return <h1>Hello, {this.props.name}</h1>;
  }
}

PR #16232, opened 2019-07-28
Remove outdated comment about gcc

epicfaace:
```
          // Works because `google-closure-compiler-js` is forked in Yarn lockfile.	
          // We can remove this if GCC merges my PR:	
          // https://github.com/google/closure-compiler/pull/2707	
          // and then the compiled version is released via `google-closure-compiler-js`.
```

It looks like the PR mentioned in that comment was merged quite some time ago, and React is no longer using a forked version of google-closure-compiler-js.


gaearon:
Thanks

PR #16240, opened 2019-07-29
[act] flush suspense fallbacks in tests

threepointone:
Opening this to share the approach andrew et al, and answer some pending questions.

In this PR, for tests (specifically, code inside an `act()` scope), we immediately trigger work that would have otherwise required a timeout. This makes it simpler to tests loading/spinner states, and makes tests resilient to changes in React.

For some of our tests(specifically, ReactSuspenseWithNoopRenderer-test.internal), we _don't_ want fallbacks to immediately trigger, because we're testing intermediate states and such. Added a feature flag `flushSuspenseFallbacksInTests` to disable this behaviour on a per case basis.

Open questions - 
- should react warn you if you try to suspend outside an act scope?

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -26,6 +26,7 @@ import {
   enableSchedulerTracing,
   revertPassiveEffectsChange,
   warnAboutUnmockedScheduler,
+  flushSuspenseFallbacksInTests,
 } from 'shared/ReactFeatureFlags';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import invariant from 'shared/invariant';
@@ -993,7 +994,7 @@ function renderRoot(
     case RootIncomplete: {
       invariant(false, 'Should have a work-in-progress.');
     }
-    // Flow knows about invariant, so it compains if I add a break statement,
+    // Flow knows about invariant, so it complains if I add a break statement,
     // but eslint doesn't know about invariant, so it complains if I do.
     // eslint-disable-next-line no-fallthrough
     case RootErrored: {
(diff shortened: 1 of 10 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -800,6 +801,21 @@ function prepareFreshStack(root, expirationTime) {
   }
 }
 
+// a wrapper around scheduleTimeout, that immediately
+// runs the callback when inside an act() scope
+function scheduleWorkTimeout(callback, period) {
+  if (flushSuspenseFallbacksInTests === true) {
```
acdlite: Can we go back to the previous version where you wrap each case of the switch statement with this check? Semantically it's *almost* the same, but not exactly. There are lots of subtle edge cases that happen in between. E.g. in the `RootSuspended` case, we sometimes restart instead of Suspended:

https://github.com/facebook/react/blob/55bc393f726ddf3def83ca5a43cfc38a45d2e421/packages/react-reconciler/src/ReactFiberWorkLoop.js#L1037-L1048

It's also less of a refactor hazard the next time someone edits this code.

threepointone: agreed, will do

threepointone:
I've updated the PR, but I do want to add at least a few tests (even if just a few) that just verify that this works in product-like code. will do before eod.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1023,7 +1024,14 @@ function renderRoot(
       // possible.
       const hasNotProcessedNewUpdates =
         workInProgressRootLatestProcessedExpirationTime === Sync;
-      if (hasNotProcessedNewUpdates && !isSync) {
+      if (
+        hasNotProcessedNewUpdates &&
+        !(
```
acdlite: Nit: Simplify these nested boolean expressions plz

threepointone: better?

PR #16241, opened 2019-07-29
[Fresh] Transfer refs when remounting

gaearon:
Previously, editing a class component would cause refs to it to get nulled after an edit.

This fixes the ref to be transferred on remount so it doesn't get lost. Added a (previously failing) regression test.

Thanks @TheSavior for the bug report.


bvaughn:
Nice!

PR #16245, opened 2019-07-29
[Scheduler] Store Tasks on a Min Binary Heap

acdlite:
Switches Scheduler's priority queue implementation (for both tasks and timers) to an array-based min binary heap.

This replaces the naive linked-list implementation that was left over from the queue we once used to schedule React roots. A list was arguably fine when it was only used for roots, since the total number of roots is usually small, and is only 1 in the common case of a single-page app.

Since Scheduler is now used for many types of JavaScript tasks (e.g. including timers), the total number of tasks can be much larger.

Heaps are the standard way to implement priority queues. Insertion is O(1) in the average case (append to the end) and O(log n) in the worst. Deletion is O(log n). Peek is O(1).

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -18,6 +18,7 @@ import {
   forceFrameRate,
   requestPaint,
 } from './SchedulerHostConfig';
+import {push, pop, peek} from './SchedulerMinHeap';
 
 // TODO: Use symbols?
 var ImmediatePriority = 1;
@@ -40,9 +41,12 @@ var LOW_PRIORITY_TIMEOUT = 10000;
 // Never times out
 var IDLE_PRIORITY = maxSigned31BitInt;
 
-// Tasks are stored as a circular, doubly linked list.
-var firstTask = null;
-var firstDelayedTask = null;
+// Tasks are stored on a min heap
+var taskQueue = [];
+var timerQueue = [];
+
+// Incrementing id counter. Used to maintain insertion order.
+var taskIdCounter = 0;
 
 // Pausing the scheduler is useful for debugging.
 var isSchedulerPaused = false;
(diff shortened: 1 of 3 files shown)


packages/scheduler/src/SchedulerMinHeap.js
```diff
@@ -0,0 +1,92 @@
+/**
```
sebmarkbage: Does this mean we've given up on having a single easy to reason about file?

acdlite: Personally I don’t think it’s worth it but I can move this into the same file if you want me to

To maximize adoption, I think polyfilling the native API is the better approach.

packages/scheduler/src/Scheduler.js
```diff
@@ -133,76 +125,34 @@ function flushTask(task, currentTime) {
...
   } finally {
     currentPriorityLevel = previousPriorityLevel;
     currentTask = previousTask;
   }
 
-  // A callback may return a continuation. The continuation should be scheduled
-  // with the same priority and expiration as the just-finished callback.
+  // A callback may return a continuation.
   if (typeof continuationCallback === 'function') {
-    var expirationTime = task.expirationTime;
```
acdlite: This whole branch is left over from before the "return a continuation" API existed. Instead of popping the task from the queue, checking if there's a continuation, and then rescheduling it, we can peek at the task and only remove it once it completes. If there's a continuation, then the task remains in its current position.

packages/scheduler/src/SchedulerMinHeap.js
```diff
@@ -0,0 +1,91 @@
...
+    } else {
+      // The parent is smaller. Exit.
+      return;
+    }
+  }
+}
+
+function siftDown(heap, node, index) {
+  const length = heap.length;
+  while (index < length) {
```
sophiebits: This is never false, right? Should it just be `while (true)`?

packages/scheduler/src/Scheduler.js
```diff
@@ -504,34 +403,14 @@ function unstable_continueExecution() {
...
-    }
-    var previous = task.previous;
-    previous.next = next;
-    next.previous = previous;
-  }
-
-  task.next = task.previous = null;
+  // Null out the callback to indicate the task has been canceled. (Can't remove
+  // from the queue because you can't remove arbitrary nodes from an array based
+  // heap, only the first one.)
```
sophiebits: Can't you sift up/down as appropriate?

acdlite: I don’t think so because I don’t know the index of the task.

packages/scheduler/src/SchedulerMinHeap.js
```diff
@@ -0,0 +1,91 @@
...
+export function peek(heap: Heap): Node | null {
+  const first = heap[0];
+  return first === undefined ? null : first;
+}
+
+export function pop(heap: Heap): Node | null {
+  const first = heap[0];
+  if (first !== undefined) {
+    const last = heap.pop();
+    if (last !== undefined && last !== first) {
```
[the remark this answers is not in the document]
acdlite: Yeah I think maybe this was to satisfy Flow? Will check

acdlite:
Will wait to merge this until after 16.9 is released

packages/scheduler/src/SchedulerMinHeap.js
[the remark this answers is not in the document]
acdlite: Nevermind, forgot Flow array access is intentionally unsafe for exactly this reason :D

PR #16253, opened 2019-07-30
A failing test for the revertPassiveEffectsChange, er,  change

threepointone:
Disabling the feature flag changed some behaviour that broke some fb tests. This is a reproduction of the issue as I understand it: A setState inside an effect, will cause an 'early' flush of all effects on second render.

Toggling the `revertPassiveEffectsChange` feature flag passes/fails the test.

Q: What is the expected behaviour in this case?


threepointone:
ok this looks like expected behaviour (tho it seems weird) closing this, will writeup thoughts later.

PR #16256, opened 2019-07-30
BugFix: Suspense priority warning firing when not supposed to

lunaruan:
Previously, the suspense priority warning was fired even if the Root wasn't suspended. Changed the warning to fire only when the root is suspended.

Also refactored the suspense priority warning so it's easier to read.


packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1738,6 +1710,38 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     );
   });
 
+  it('', async () => {
```
gaearon: Missing test name

packages/react/src/__tests__/ReactProfiler-test.internal.js
```diff
@@ -2733,10 +2733,7 @@ describe('Profiler', () => {
         if (__DEV__) {
           expect(console.error).toHaveBeenCalledTimes(1);
           expect(console.error.calls.argsFor(0)[0]).toContain(
-            'Warning: %s\n\nThe fix is to split the update',
-          );
-          expect(console.error.calls.argsFor(0)[1]).toContain(
-            'A user-blocking update was suspended by:',
+            'Warning: A user-blocking update was suspended by:',
           );
           expect(console.error.calls.argsFor(0)[1]).toContain('AsyncText');
```
gaearon: I know this code already existed but curious — any reason we’re not using `toWarnDev()` for this warning?

acdlite:
@sebmarkbage Luna and I discussed this before she submitted the PR. The issue is that if we wait until the complete phase, we've lost track of the components that suspended in that subtree. So we'll need to track them via some mechanism, like on the stack.

Or we could omit the names of the components that suspended from the message. I think the names of the components that triggered the update are the more relevant piece of information. Though arguably, the names of the components that suspended could also be helpful for debugging in the `useSuspenseDeferredValue` case.

acdlite:
I was kind of hoping we could omit the names of the components that suspended, given that the implementation is annoying, and rely React DevTools to surface this information instead, eventually.

sebmarkbage:
I think what you really want is *what* suspended. As in what is the Promise associated with. That's why we're now attaching debug info to the promises. Not the nearest component that threw one. With call through and other optimizations we might not know the exact component. We should only need try/catch at boundary levels, not nearest component. This is unfortunate leak.

So I'd be in favor removing that from the message.

acdlite:
Ok @lunaruan in that case you can leave the `checkForWrongSuspensePriorityInDEV` call it is, and only flush the warning in the `RootSuspended` or `RootSuspendedWithDelay` cases, as you've already done. Then remove the names of the components that threw from the message.

Should also add a test where the component that triggers the update is in between the Suspense boundary and the component that throws.

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1730,56 +1674,80 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
     );
   });
 
-  it('warns when suspending inside discrete update', async () => {
-    function A() {
-      Scheduler.unstable_yieldValue('A');
-      TextResource.read(['A', 1000]);
-      return 'A';
-    }
+  it('Does not warn when a component updates after the fallback is shown', async () => {
```
acdlite:
```suggestion
  it('does not warn about wrong Suspense priority if no new fallbacks are shown', async () => {
```

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2661,70 +2661,36 @@ export function checkForWrongSuspensePriorityInDEV(sourceFiber: Fiber) {
...
-          '\n\n' +
-          '  ' +
-          componentNamesString;
-      } else {
-        componentThatTriggeredSuspenseError =
-          'A user-blocking update was suspended by:' +
-          '\n\n' +
-          '  ' +
-          componentNamesString;
+            '\n\n' +
```
acdlite: Let's make this "...triggered a user-blocking update that suspended"

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1730,56 +1674,80 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
+    });
 
-    ReactNoop.discreteUpdates(() => ReactNoop.render(<App />));
-    expect(Scheduler).toFlushAndYieldThrough(['A']);
+    expect(Scheduler).toHaveYielded(['Suspend! [A]', 'Suspend! [B]']);
+  });
 
-    // Warning is not flushed until the commit phase
+  it(
+    'Warns when component that triggered update is between Suspense boundary ' +
```
acdlite:
```suggestion
    'Warns when component that triggered user-blocking update is between Suspense boundary ' +
```

packages/react/src/__tests__/ReactProfiler-test.internal.js
lunaruan: I think it was easier at the time to use console.error, because there was a lot of them. Do you prefer `toWarnDev()`?

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -1730,56 +1674,80 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
-        <Suspense fallback="Loading...">
-          <A />
-          <B />
-          <C />
-          <C />
-          <C />
-          <C />
-        </Suspense>
+    expect(Scheduler).toHaveYielded(['Suspend! [A]']);
+
```
acdlite: I would put another assertion here:

```js
expect(ReactNoop).toMatchRenderedOutput(<span prop="Loading..." />);
```

PR #16257, opened 2019-07-30
Bugfix:  Effects should never have higher than normal priority

acdlite:
The priority of passive effects is supposed to be the same as the priority of the render. (That was the original intended behavior, at least.) The first commit fixes a bug where the priority is sometimes wrong if the effects are flushed early.

However, the priority should really never be higher than Normal Priority.

The second commit changes the behavior so that effects always have normal priority, or lower if the render priority is lower (e.g. offscreen prerendering).

The implementation is a bit awkward because of the way `renderRoot`, `commitRoot`, and `flushPassiveEffects` are split. This is a known factoring problem that I'm planning to address once 16.9 is released.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSchedulerIntegration-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSchedulerIntegration-test.internal.js
@@ -170,6 +170,34 @@ describe('ReactSchedulerIntegration', () => {
     });
   });
 
+  it('passive effects have render priority even if they are flushed early', async () => {
+    const {useEffect} = React;
+    function ReadPriority({step}) {
+      Scheduler.unstable_yieldValue(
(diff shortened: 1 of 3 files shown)


acdlite:
There's a test failing; will look when I get back from coffee

packages/react-reconciler/src/__tests__/ReactSchedulerIntegration-test.internal.js
```diff
@@ -213,7 +256,4 @@ describe('ReactSchedulerIntegration', () => {
     Scheduler.unstable_flushUntilNextPaint();
     expect(Scheduler).toHaveYielded(['A', 'B', 'C']);
   });
-
-  // TODO
-  it.skip('passive effects have render priority even if they are flushed early', () => {});
```
acdlite: Should probably conduct regular audits of our TODOs :D

PR #16259, opened 2019-07-31
[act] Wrap IsThisRendererActing in DEV check

acdlite:
So that it doesn't leak into the production bundle. Follow-up to #16240.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
+++ b/packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
@@ -722,68 +722,74 @@ function runActTests(label, render, unmount, rerender) {
     });
 
     describe('suspense', () => {
-      it('triggers fallbacks if available', async () => {
-        let resolved = false;
-        let resolve;
-        const promise = new Promise(_resolve => {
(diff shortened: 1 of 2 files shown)


acdlite:
@threepointone Just to confirm, this isn't supposed to affect prod, right? There are some tests that are failing but I assume that was just an oversight in the last PR.

acdlite:
I'm going to assume that this doesn't affect prod and we can follow up later if that's not the case. All the `act` stuff should be dev-only, or alternatively, in a separate testing build.

threepointone:
We use this sigil not just for warnings now, but also for deciding when to flush, so it does actually affect prod?

acdlite:
Is this the only case? We really should not be leaking `act` behavior into the main production bundles.

acdlite:
What is the use case for running `act` tests in production?

threepointone:
I think the discussion we had was people might be running tests with prod builds? iirc. Happy to not support it if so, or move to making testing builds.

acdlite:
Since this only affects Concurrent and Sync mode I'll merge this and we can pick up discussion elsewhere.

PR #16261, opened 2019-07-31
Add missing check to unmocked Scheduler warning

acdlite:
The unmocked Scheduler warning doesn't actually check if Scheduler is mocked.

Only affects www because the `warnAboutUnmockedScheduler` feature flag is off in the open source build.


acdlite:
Merging to unblock sync

threepointone:
Sorry! 😣 Thanks for the fix!

PR #16264, opened 2019-07-31
[Flare] Tweaks to Flare system design and API

trueadm:
This PR makes some slight adjustments to the exist Flare event system design. After internal testing and from feedback, it was found that passing responders as a React Element to host components, then having detached listeners led to unnecessary boilerplate and some issues around composability. To address these concerns the follow changes have been made:

## `useListener` -> `useResponder`

There was confusion as to why `useListener` was actually using an Event Responder, but was named "useListener". Now the hook is `useResponder`, which takes an Event Responder and returns and Event Listener. 

## `responders` prop -> `listeners` prop

Now that the `useResponder` hook returns a discrete event listener object, this now gets passed to host components rather than supplying a responder as a React Element on the host component.

To demonstrate these changes, here's a before and after:

```jsx
// Before
function PressableButton({onPress}) {
  useListener(PressResponder, {onPress});

  return (
    <button responders={<PressResponder />}>Press me</button>
  );
}

// After
function PressableButton({onPress}) {
  const pressListener = useResponder(PressResponder, {onPress});

  return (
    <button listeners={pressListener}>Press me</button>
  );
}
```

This design change allows us to explore composing the sequence of listeners, allowing for more complex interactions. It also makes the bindings between listeners and their responders more explicit and less magical.


necolas:
Thanks this looks good! The implementation appears simpler too. I do have a few questions / comments.

> context.dispatchEvent

The implementation is still dispatching named events. Why do we need to do this? What is the payload `type` needed for now? And if it's necessary to treat the gesture lifecycle as a series of events internally, is that something we can better hide from the responder callbacks (i.e., not having the event `type` in the payload they receive)? In Flutter this object is called "details", in SwiftUI this is object is called "GestureState". That's how we're using it too – an object the reflects the current gesture state.

> Responder propagation control

We've discussed several ownership models but this wasn't one of them. I'm not sure refs are a practical way to negotiate ownership as the same ref would have to be passed around a lot. I don't think the term "propagation" is relevant in the context of ownership over a pointer / gesture. We should chat about this idea at work.

> Remove FocusScope

👍 But let's do this as a separate PR.

trueadm:
@necolas 

> The implementation is still dispatching named events

I didn't want to change the event object shapes as part of this PR.

> We've discussed several ownership models but this wasn't one of them. 

This isn't to do with the ownership model. It's just a way of allowing responders of the same type to not propagate. Without this our internal Pressability design will no longer work.

> But let's do this as a separate PR.

I wanted to avoid converting the FocusScope module and tests over.

necolas:
> I didn't want to change the event object shapes as part of this PR.

That makes perfect sense. But beyond this PR that's something I'd like to better understand. Something we can chat about in person.

> This isn't to do with the ownership model

I think this is still part of the system that determines which responders should be active, like the `behavior` concept. Should we keep the existing mechanism for now and revisit it in another PR after this gets merged?

> I wanted to avoid converting the FocusScope module and tests over.

No need to do that. We can merge a PR that removes FocusScope before this one gets applied to master.

trueadm:
> That makes perfect sense. But beyond this PR that's something I'd like to better understand. Something we can chat about in person.

Agreed. I thought this might be an area that you'd like to spend some time working on actually, but you'd want the design changes in before you did that.

> I think this is still part of the system that determines which responders should be active, like the behavior concept. Should we keep the existing mechanism for now and revisit it in another PR after this gets merged?

Agreed that it doesn't make much sense now. I'll remove that logic and the code for it. As long as we don't sync, this should be fine anyway.

trueadm:
I've restore the old logic for how responder propagation occurs.

Issue #15691, opened 2019-05-21
TypeError: Cannot read property 'body' of null


aralroca:
This method:

https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/react-dom/src/client/getActiveElement.js

Is throwing this error in jest:

![image](https://user-images.githubusercontent.com/13313058/58075486-20693a00-7ba8-11e9-91c4-73caaf9f9c68.png)

I realize that this method is checking only when the document is `undefined`. However, using jest looks that is setting sometimes the document to `null`, causing this error.

To reproduce it, I'm testing a callback that is using an async task:

```jsx
 const asyncTask = useCallback(async (data) => {
    dispatch({ type: 'FETCH' })

    try {
      const res = await MyService.asyncMethod(data)
      dispatch({ type: 'SUCCESS', data: res })
    } catch (e) {
      dispatch({ type: 'ERROR' }) // This dispatch throw this error in jest
    }
  }, [])
```

About the test, I'm just mocking the service and testing the different scenarios.

I imagine that is related that the `dispatch` from the `useReducer`, that it's used inside the catch, is fired when the component maybe is unmount? @gaearon In this case how can I do a safer dispatch inside a callback method?

aralroca:
I tried to do this:

```jsx
 const dispatchSafe = useSafeDispatch(dispatch)

 const asyncTask = useCallback(async (data) => {
    dispatchSafe({ type: 'FETCH' })

    try {
      const res = await MyService.asyncMethod(data)
      dispatchSafe({ type: 'SUCCESS', data: res })
    } catch (e) {
      dispatchSafe({ type: 'ERROR' }) // This dispatchSafe throw the error in jest
    }
  }, [])
```

When **useSafeDispatch** is:

```jsx
export default function useSafeDispatch(dispatch) {
  const dispatchSafe = useRef(dispatch)

  useEffect(() => () => {
    dispatchSafe.current = () => null
  }, [])

  return dispatchSafe.current
}
```

However, this doesn't solve the problem. Maybe I'm doing something wrong?

threepointone:
could you make a repro with codesandbox, or a git repo? can't say without more info.

cristiam86:
I'm also having this issue. In my case it only happens when there is a console.log inside the catch.

No error:
```js
async componentDidMount() {
  try {
    const helpMenu = await MenuService.getMenuAyuda();
    this.setState({ helpMenuArray: helpMenu });
[...]

PR #16266, opened 2019-07-31
Fix TypeError in getActiveElement

ishristov:
During tests (with react-testing-library) a random test randomly fails
with "TypeError: Cannot read property 'body' of null" in the getActiveElement.
It appears that sometimes the `doc` is null but the check in the file only
checks for undefined. This proposed change checks for both `null` or `undefined`
which should prevent the failures.

https://github.com/facebook/react/issues/15691

### DIFF
--- a/packages/react-dom/src/client/getActiveElement.js
+++ b/packages/react-dom/src/client/getActiveElement.js
@@ -9,7 +9,7 @@
 
 export default function getActiveElement(doc: ?Document): ?Element {
   doc = doc || (typeof document !== 'undefined' ? document : undefined);
-  if (doc == null) {
+  if (doc === null || doc === undefined) {
     return null;
   }
   try {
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Hey! Do you have a way for us to reproduce the issue?

packages/react-dom/src/client/getActiveElement.js
```diff
@@ -9,7 +9,7 @@
 
 export default function getActiveElement(doc: ?Document): ?Element {
   doc = doc || (typeof document !== 'undefined' ? document : undefined);
-  if (typeof doc === 'undefined') {
+  if (doc == null) {
```
philipp-spiess: We try to be as explicit as possible so if we want to check for `null` and `undefined` something like this would be better:

```suggestion
  if (doc === null || doc === undefined) {
```

ishristov: Done.

ishristov:
> Hey! Do you have a way for us to reproduce the issue?

I don't think I'll be able to do that easily or at all...

threepointone:
I think you’ll see these errors if you have async code that runs after the test exits. The fix should be in your test, not in react.

threepointone:
Closing since this should be fixed at the test, not in React.

PR #16268, opened 2019-07-31
use helper function to create react symbol

Yelmor:
This commit makes no functional change.
It brings:
1. cleaner code;
2. slightly reduce dist bundle size (~0.1kb).


gaearon:
Thanks but it doesn't appear to reduce bundle size in practice (see the bot message above).

PR #16269, opened 2019-07-31
Add a feature flag to disable legacy context

gaearon:
Adds a feature flag that disables the legacy context API. The feature flag is turned off in open source builds. Strict Mode is already a mechanism that warns about it, but this feature flag actually makes `this.context` always an empty object, and `context` argument to functions always `undefined`. So it's a way to enforce it. If we later offer "modern" builds, we can turn it on there by default. For now, I want to use this internally for newly written code.

No whitespace: https://github.com/facebook/react/compare/master...gaearon:ctx?expand=1&w=1

### DIFF
--- a/packages/react-reconciler/src/ReactFiberContext.js
+++ b/packages/react-reconciler/src/ReactFiberContext.js
@@ -134,13 +134,11 @@ function hasContextChanged(): boolean {
 }
 
 function isContextProvider(type: Function): boolean {
-  const childContextTypes = type.childContextTypes;
-  const hasChildContextTypes =
-    childContextTypes !== null && childContextTypes !== undefined;
   if (disableLegacyContext) {
     return false;
   } else {
-    return hasChildContextTypes;
+    const childContextTypes = type.childContextTypes;
+    return childContextTypes !== null && childContextTypes !== undefined;
   }
 }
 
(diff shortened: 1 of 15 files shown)


packages/react-reconciler/src/ReactFiberContext.js
```diff
@@ -46,259 +47,333 @@ function getUnmaskedContext(
...
 function cacheContext(
   workInProgress: Fiber,
   unmaskedContext: Object,
   maskedContext: Object,
 ): void {
-  const instance = workInProgress.stateNode;
-  instance.__reactInternalMemoizedUnmaskedChildContext = unmaskedContext;
-  instance.__reactInternalMemoizedMaskedChildContext = maskedContext;
+  if (disableLegacyContext) {
+    invariant(
```
acdlite: Can we make these dev-only warnings instead? Once everyone removes legacy context from their apps, we can remove the warning in a minor, whereas removing an invariant is an according-to-semver-pedants breaking change.

gaearon:
Previously, classes that weren't context consumers would get `{}`. You're saying it should be `undefined` too. That's a bigger breaking change but maybe it makes sense because it lets you find code with bad assumptions more easily.

acdlite:
> That's a bigger breaking change but maybe it makes sense because it lets you find code with bad assumptions more easily.

Right, and then we don't have to do a second breaking change later

gaearon:
We could still remove the feature but keep it always `{}`. Since it doesn't cost us anything to keep setting it once.

acdlite:
@gaearon It's the second argument to functions that I care more about, personally. We might want to use that for something else.

Don't really care about classes. What's a class? 🤷‍♀️

PR #16271, opened 2019-08-01
[Scheduler] Delete old rAF implementation

acdlite:
Our performance tests indicate that the message loop implementation has better CPU utilization than the current rAF one. This commit removes the feature flag and deletes the rAF implementation.

I removed the old SchedulerDOM tests because they assumed a rAF implementation. I ported the relevant tests to the SchedulerBrowser test suite.

Opening the PR now, but I won't merge this until after 16.9.


packages/scheduler/src/__tests__/SchedulerBrowser-test.js
```diff
@@ -0,0 +1,377 @@
...
+    });
+    scheduleCallback(NormalPriority, () => {
+      runtime.log('B');
+    });
+    runtime.assertLog(['Post Message']);
+    runtime.fireMessageEvent();
+    runtime.assertLog(['Message Event', 'A', 'B']);
+  });
+
+  it('multiple tasks with a yield in between', () => {
```
gaearon: This test breaks after the min heap refactor.

gaearon: Ok that's because it needs 4999 instead of 5000 now. Must be some comparison condition flip. Probably doesn't matter.

gaearon:
Since we want to keep both for now, I pulled all test except `delayed task` (which would need setTimeout setup fixes that break rAF tests) into #16271.

acdlite:
Will open new PR

PR #16282, opened 2019-08-02
Warn if you try to use `act()` in prod

threepointone:
We have behaviour divergence for act() between prod and dev (specifically, act() + concurrent mode does not flush fallbacks in prod. This doesn't affect anyone in OSS yet.)

We also don't have a good story for writing tests in prod (and what from what I gather, nobody really writes tests in prod mode).

We could have wiped out act() in prod builds, except that _we_ ourselves use act() for our tests when we run them in prod mode.

This PR is a compromise to all of this. We will log a warning if you try to use act() in prod mode, and we silence it in our test suites. (Any discrepancies in behaviour will still be caught by assertions). 

We will revisit this in a future version.

### DIFF
--- a/scripts/jest/shouldIgnoreConsoleError.js
+++ b/scripts/jest/shouldIgnoreConsoleError.js
@@ -25,6 +25,16 @@ module.exports = function shouldIgnoreConsoleError(format, args) {
       // They are noisy too so we'll try to ignore them.
       return true;
     }
+    if (
+      format.indexOf(
+        'act(...) is not supported in production builds of React'
+      ) === 0
+    ) {
+      // We don't yet support act() for prod builds, and warn for it.
+      // But we'd like to use act() ourselves for prod builds.
+      // Let's ignore the warning and #yolo.
+      return true;
+    }
   }
   // Looks legit
   return false;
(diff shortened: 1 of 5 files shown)


scripts/jest/shouldIgnoreConsoleError.js
```diff
@@ -25,6 +25,14 @@ module.exports = function shouldIgnoreConsoleError(format, args) {
       // They are noisy too so we'll try to ignore them.
       return true;
     }
+    if (
+      format.indexOf('act(...) is not supported in production builds of React.')
```
threepointone: huh this doesn’t actually match the error, odd that nothing failed

threepointone: silly me, fixed.

acdlite:
Tests?

threepointone:
doh! 😅 fixing asap

threepointone:
Added a test.

acdlite:
Merging so I can sync it to www

henryqdineen:
> from what I gather, nobody really writes tests in prod mode

We currently run our tests in prod mode in the CI and I noticed this warning while upgrading to 16.9.0. Will this be removed in a future version without deprecations? Any suggestions other than running tests against the DEV build? It would be a major infra change for us at this point. Thanks!

PR #16285, opened 2019-08-03
correct ReactContext.js if block writting at line21

fantasticsoul:
correct ReactContext.js if block writting at line21

### DIFF
--- a/packages/react/src/ReactContext.js
+++ b/packages/react/src/ReactContext.js
@@ -20,16 +20,15 @@ export function createContext<T>(
 ): ReactContext<T> {
   if (calculateChangedBits === undefined) {
     calculateChangedBits = null;
-  } else {
-    if (__DEV__) {
-      warningWithoutStack(
-        calculateChangedBits === null ||
-          typeof calculateChangedBits === 'function',
-        'createContext: Expected the optional second argument to be a ' +
-          'function. Instead received: %s',
-        calculateChangedBits,
-      );
-    }
+  }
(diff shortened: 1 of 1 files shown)


packages/react/src/ReactContext.js
```diff
@@ -20,16 +20,15 @@ export function createContext<T>(
...
-          typeof calculateChangedBits === 'function',
-        'createContext: Expected the optional second argument to be a ' +
-          'function. Instead received: %s',
-        calculateChangedBits,
-      );
-    }
+  }
+  if (__DEV__) {
+    warningWithoutStack(
+      calculateChangedBits === null ||
```
audiolion: this change could cause this warning to be thrown now where it wouldn't before, `calculateChangedBits` is `undefined`, then gets set to `null`, then you are in `__DEV__` and `calculateChangedBits === null` now passes.

acdlite: `warningWithoutStack` fires when its first argument is falsey. When `calculateChangedBits` is `null`, it does not warn.

So in either version, if `calculateChangedBits` is undefined, there's no warning. So the behavior is the same.

acdlite:
Thanks for the PR, but the behavior is the same either way. Explanation: https://github.com/facebook/react/pull/16285#discussion_r311183195

packages/react/src/ReactContext.js
audiolion: ah I assumed it called when `true`, my bad.

PR #16296, opened 2019-08-05
[react-events] Add ContextMenu responder

necolas:
A module for responding to contextmenu events. This functionality will be
removed from the Press responder in the future.


necolas:
Note: this responder uses a different approach to defining the native events it listens to, and surfaces an issue our existing responders.

The `ContextMenu` responder listens to *either* pointer events *or* fallback events. Doing this makes it easier to keep the PointerEvents and fallback implementations separate in more complicated responders, as the fallback events can be completely ignored. And it also makes it clear that the test suites needs to be run twice, with the correct environment simulated. 

In contrast, the existing responders listen to pointer events by default, and if those aren't supported they *add* fallback events. In the unit test environment, pointer events aren't supported so the responder always adds the fallback events! However, in the unit tests we still dispatch native pointer events in almost all the scenarios. This creates an unrealistic environment, where a browser is dispatching pointer events but React is also listening to the fallback events! As such, the existing unit tests are both imposing greater complexity on the PointerEvents implementation that would occur in reality (i.e., listening to fallback events and having to deal with them) and potentially covering up bugs if the PointerEvents path accidentally ends up depending on state changes caused by the fallback events (because those events won't be listened to outside of the unit test env).

engprodigy:
Hi @necolas is there any way I can test this experimental module in a react app? Thanks.

PR #16297, opened 2019-08-05
Babel 7

lunaruan:
For #12548 

Upgraded from Babel 6 to Babel 7.

The only significant change seems to be the way `@babel/plugin-transform-classes` handles classes differently from `babel-plugin-transform-es2015-classes`. In regular mode, the former injects a `_createClass` function that increases the bundle size, and in the latter it removes the safeguard checks. However, this is okay because we don't all classes in new features, and we want to deprecate class usage in the future in the react repo.

Differences between classes in Babel 6 (left) and Babel 7 (right)
image
image

Example diffs (6 on the left and 7 on the right):
[ReactDOM-prod.js (facebook-www)](https://www.diffchecker.com/l6yOxMlv)
[ReactDOM-dev.js (facebook-www)](https://www.diffchecker.com/qX91sRbT)
[react-dom.development.js (UMD)](https://www.diffchecker.com/nNATN138)
[react-dom.production.js (UMD)](https://www.diffchecker.com/qenU4uy8)
[react-dom.development.js (CommonJS)](https://www.diffchecker.com/KyACMHRa)
[react-dom.production.js (CommonJS)](https://www.diffchecker.com/7PjUMRjo)
[react-dom-server.node.production.min.js (CommonJS)](https://www.diffchecker.com/YYEcoNzs)
[react-dom-server.browser.production.min.js (UMD)](https://www.diffchecker.com/hkhunOI0)


package.json
```diff
@@ -4,38 +4,38 @@
...
     "art": "^0.10.1",
-    "babel-cli": "^6.6.5",
-    "babel-code-frame": "^6.26.0",
-    "babel-core": "^6.0.0",
     "babel-eslint": "^10.0.0",
-    "babel-jest": "^23.0.1",
-    "babel-plugin-check-es2015-constants": "^6.5.0",
-    "babel-plugin-external-helpers": "^6.22.0",
-    "babel-plugin-syntax-dynamic-import": "^6.18.0",
+    "babel-jest": "^24.8.0",
```
SimenB: you can remove `babel-jest` entirely, it's not used

babel.config.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+module.exports = {
+  presets: ['@babel/preset-react', '@babel/preset-flow'],
```
acdlite: Can we remove these presets and use the plugins directly? That way expected stuff can't sneak in.

packages/react-refresh/src/__tests__/ReactFreshBabelPlugin-test.js
```diff
@@ -7,25 +7,54 @@
...
         freshPlugin,
         ...(options.plugins || []),
       ],
     }).code,
   );
 }
 
+const oldEnv = babel.getEnv();
 describe('ReactFreshBabelPlugin', () => {
+  beforeEach(() => {
```
acdlite: The reason Fresh throws an error if you're not in development environment is because the production build of React does not support hot reloading. The production build of React uses a `process.env.NODE_ENV` check (that's what `__DEV__` compiles to), not `BABEL_ENV`. Technically, `BABEL_ENV` and `NODE_ENV` could be different, even though they're usually the same.

We should use `__DEV__`/`NODE_ENV` since that's what determines the version of React DOM.

```js
if (__DEV__) {
  // The tests
} else {
  it('throw error if environment is not development', ...);
}
```

babel.config.js
```diff
@@ -1,8 +1,11 @@
 'use strict';
 
 module.exports = {
-  presets: ['@babel/preset-react', '@babel/preset-flow'],
   plugins: [
+    '@babel/plugin-syntax-jsx',
+    '@babel/plugin-transform-react-jsx',
+    '@babel/plugin-transform-react-display-name',
```
acdlite: Let's remove this one. Don't need it.

package.json
lunaruan: Will do. Thanks!

bvaughn:
Nice work @lunaruan !

scripts/jest/config.source-persistent.js
```diff
@@ -9,6 +9,12 @@ module.exports = Object.assign({}, baseConfig, {
     'ReactIncrementalTriangle',
     'ReactIncrementalReflection',
     'forwardRef',
+    // ReactFreshBabelPlugin is only available for dev.
+    // We need two tests here because otherwise, ReactFreshBabelPlugin-test will
+    // fail due to obsolete snapshots
+    process.env.NODE_ENV === 'development'
+      ? '<rootDir>/packages/react-refresh/src/__tests__/ReactFreshBabelPluginProd-test.js'
+      : '<rootDir>/packages/react-refresh/src/__tests__/ReactFreshBabelPlugin-test.js',
```
gaearon: This is kinda hacky. Can we figure out another way without making exceptions at this level?

scripts/rollup/bundles.js
```diff
@@ -134,12 +134,26 @@ const bundles = [
     entry: 'react-dom/server.browser',
     global: 'ReactDOMServer',
     externals: ['react'],
+    babel: opts =>
+      Object.assign({}, opts, {
+        // Include JSX
```
gaearon: The commend doesn't match the intention.

TrySound:
I would merge this first https://github.com/facebook/react/pull/15037

PR #16302, opened 2019-08-06
[Fresh] Support multiple renderers at the same time

gaearon:
Fresh Runtime used to assume there's only one renderer injected at a time. That assumption is inaccurate. For example, on DOM+ART or Paper+Fabric surfaces it resulted in the "wrong" roots being passed to a renderer.

The fix is to store the renderer helpers per renderer ID, and associate mounted roots with them. Then we know which one to use.

I verified this fixes an issue we found in the Fabric experiment.

### DIFF
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -20,13 +20,19 @@ import type {
 import type {ReactNodeList} from 'shared/ReactTypes';
 
 import {REACT_MEMO_TYPE, REACT_FORWARD_REF_TYPE} from 'shared/ReactSymbols';
-import warningWithoutStack from 'shared/warningWithoutStack';
 
 type Signature = {|
   ownKey: string,
   forceReset: boolean,
   fullKey: string | null, // Contains keys of nested Hooks. Computed lazily.
   getCustomHooks: () => Array<Function>,
+|};
+
+type RendererHelpers = {|
+  findHostInstancesForRefresh: FindHostInstancesForRefresh,
+  scheduleRefresh: ScheduleRefresh,
+  scheduleRoot: ScheduleRoot,
+  setRefreshHandler: SetRefreshHandler,
 |};
 
 if (!__DEV__) {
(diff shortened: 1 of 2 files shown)


packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -397,11 +380,14 @@ export function injectIntoGlobalHook(globalObject: any): void {
       // However, if there is no DevTools extension, we'll need to set up the global hook ourselves.
       // Note that in this case it's important that renderer code runs *after* this method call.
       // Otherwise, the renderer will think that there is no global hook, and won't do the injection.
+      let nextID = 0;
       globalObject.__REACT_DEVTOOLS_GLOBAL_HOOK__ = hook = {
         supportsFiber: true,
-        inject() {},
+        inject(injected) {
+          return nextID++;
```
gaearon: This is not exactly what DevTools does but close enough.

packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -413,23 +399,31 @@ export function injectIntoGlobalHook(globalObject: any): void {
...
     hook.onCommitFiberRoot = function(
-      id: mixed,
+      id: number,
       root: FiberRoot,
       maybePriorityLevel: mixed,
       didError: boolean,
     ) {
+      const helpers = helpersByRendererID.get(id);
+      if (helpers === undefined) {
+        return;
```
gaearon: There can still be commits to renderers that don't support Fast Refresh. We ignore those.

gaearon:
Oops, forgot the test at first. Pushed it up.

packages/react-refresh/src/__tests__/ReactFreshMultipleRenderer-test.internal.js
```diff
@@ -0,0 +1,96 @@
...
+'use strict';
+
+jest.resetModules();
+let React = require('react');
+let ReactFreshRuntime = require('react-refresh/runtime');
+ReactFreshRuntime.injectIntoGlobalHook(global);
+let ReactDOM = require('react-dom');
+
+jest.resetModules();
+let ReactART = require('react-art');
```
gaearon: These are top level because I couldn't get ART to work otherwise. Dunno why.

packages/react-refresh/src/ReactFreshRuntime.js
bvaughn: This might be nice to leave as a comment.

packages/react-refresh/src/__tests__/ReactFreshMultipleRenderer-test.internal.js
bvaughn: Ah, that's unfortunate since it means you can't reset modules between tests.

gaearon: Yeah. I spent 20 minutes on this and then flipped the table

PR #16306, opened 2019-08-07
Reset hydration state after reentering

sebmarkbage:
We might reenter a hydration state, when attempting to hydrate a boundary. We need to ensure that we reset it to not hydrating once we exit it. Otherwise the next sibling will still be in hydration mode.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberUnwindWork.js
+++ b/packages/react-reconciler/src/ReactFiberUnwindWork.js
@@ -25,6 +25,7 @@ import {enableSuspenseServerRenderer} from 'shared/ReactFeatureFlags';
 
 import {popHostContainer, popHostContext} from './ReactFiberHostContext';
 import {popSuspenseContext} from './ReactFiberSuspenseContext';
+import {resetHydrationState} from './ReactFiberHydrationContext';
 import {
   isContextProvider as isLegacyContextProvider,
   popContext as popLegacyContext,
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -134,7 +139,6 @@ function unwindInterruptedWork(interruptedWork: Fiber) {
       break;
     case DehydratedSuspenseComponent:
       if (enableSuspenseServerRenderer) {
-        // TODO: popHydrationState
```
sebmarkbage: This one is not needed since restarting the root covers it. I think anything covered by resetFreshStack also doesn't need to be here.

packages/react-reconciler/src/ReactFiberUnwindWork.js
```diff
@@ -80,8 +81,12 @@ function unwindWork(
     }
     case DehydratedSuspenseComponent: {
       if (enableSuspenseServerRenderer) {
-        // TODO: popHydrationState
         popSuspenseContext(workInProgress);
+        if (workInProgress.alternate === null) {
+          // TODO: popHydrationState
```
sebmarkbage: I can't actually do this yet because this because popping causes hydration warnings. This is part of a larger bug that suspending in a dehydrated tree can leave the hydration context in an inconsistent state when we move to work on the sibling of the thing that suspended. That's a bigger issue to fix.

PR #16312, opened 2019-08-07
[act] remove obsolete container and warning

threepointone:
In a previous version of `act()`, we used a dummy dom element to flush effects. This doesn't need to exist anymore, and this PR removes it. The warning doesn't need to be there either (React will fire a wrong renderer act warning if needed).

### DIFF
--- a/packages/react-dom/src/test-utils/ReactTestUtils.js
+++ b/packages/react-dom/src/test-utils/ReactTestUtils.js
@@ -4,7 +4,6 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import type {Thenable} from 'react-reconciler/src/ReactFiberWorkLoop';
 
 import React from 'react';
 import ReactDOM from 'react-dom';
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/test-utils/ReactTestUtils.js
```diff
@@ -396,21 +393,17 @@ const ReactTestUtils = {
...
-              "If you're using TestRenderer for your tests, you should call " +
-              'ReactTestRenderer.act(...) instead of ReactTestUtils.act(...).',
-          );
-        }
+    if (__DEV__) {
+      // warn if we're trying to use this in something like node (without jsdom)
+      if (didWarnAboutActInNodejs === false) {
+        didWarnAboutActInNodejs = true;
+        warningWithoutStack(
+          typeof document !== 'undefined' && document !== null,
```
gaearon: I know it already exists but structure is confusing because the variable is called `didWarnAboutActInNodejs` but it would be true if we did _not_ warn too.

I'd do this instead:

```js
if (typeof document === 'undefined' || document === null) {
  warningWithoutStack(didWarnAboutActInNodejs, '...');
  didWarnAboutActInNodejs = true;
}
```

Or even two levels of nesting and just `warning(false, ...)`

threepointone: yeah I'll clean this up

threepointone: Fixed by nuking the whole thing ha. updated PR description.

Issue #12209, opened 2018-02-11
Warn if stringifying an attribute takes too long


gaearon:
With React 16 we don't have an attribute whitelist so both `src={obj}` and `myattribute={obj}` would be valid. The objects get stringified and added as attributes for smooth migration path because a lot of the existing code already depends on this behavior.

There is, however, one pitfall here. Sometimes you made do `<div {...rest}>` and not realize that `rest` includes an object whose stringifying is unusually expensive. For example a deeply nested Immutable Map. Now, this wouldn't produce an error, but it would slow down rendering for no good reason.

We could protect against this by putting `performance.now()` counters around the places where we stringify attributes. If stringification takes more than, say, 2 milliseconds, then something bad is going on, and we should probably warn.

Swieckowski:
@gaearon I'd like to claim this.

gaearon:
Sounds good!

Swieckowski:
Could you point in the right direction for where the migration path starts for attributes? I've been looking at the PRs mentioned in https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html to try figure out where to start looking, but it's been a little overwhelming.

aweary:
@Swieckowski I think you'll want to look at `setValueForProperty` in [`DOMPropertyOperations`](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/DOMPropertyOperations.js)

https://github.com/facebook/react/blob/fb85cf2e9c52a4b2999144a89a6ee8256aec55c7/packages/react-dom/src/client/DOMPropertyOperations.js#L164-L166

https://github.com/facebook/react/blob/fb85cf2e9c52a4b2999144a89a6ee8256aec55c7/packages/react-dom/src/client/DOMPropertyOperations.js#L136

Those two lines are where known attribute and unknown attribute values get stringified, respectively. I'm not sure if @gaearon intends to only warn for attributes that we don't recognize, but this should be a good place to start.

Keep in mind that this warning and timing logic should only happen in the DEV bundle. This is handled by wrapping code in `__DEV__` checks, which you can see examples of in the file linked above.

Hope that helps!

Swieckowski:
That's actually the file I was just looking at, but was a bit confused by the way it was being stringified and the (value: any) which I've come to the conclusion is flow.

Thanks a lot, it's good to know I was at least looking in the right place. Is there a reason that's the way the attribute value is being stringified this way?

Swieckowski:
[...]

PR #16315, opened 2019-08-08
fixing some Unperformant Stringify Warning

kambleaa007:

https://github.com/facebook/react/issues/12209

### DIFF
--- a/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
+++ b/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
@@ -157,6 +157,45 @@ describe('DOMPropertyOperations', () => {
     });
   });
 
+  it('should warn if custom attributes take too long to stringify', () => {
+    const container = document.createElement('div');
+    const attributeValue = {foo: 'bar'};
+    attributeValue.toString = function() {
+      // finds 2000th prime to waste time
+      nthPrime(1225);
+      let originalToString = Object.prototype.toString;
+      return originalToString.apply(this);
+    };
+
+     const nthPrime = n => {
+      let sieve = [2];
+      let current = 3;
+      let prime;
+
+       while (sieve.length < n) {
+        current += 2;
+        if (current % 2 === 0) {
+          continue;
+        }
+        prime = true;
+        for (let j in sieve) {
+          if (current % sieve[j] === 0) {
+            prime = false;
+            break;
+          }
+        }
+        if (prime) {
+          sieve.push(current);
+        }
+      }
+      return current;
+    };
+    expect(() =>
+      ReactDOM.render(<div data-foo={attributeValue} />, container),
+    ).toWarnDev('Stringifying your attribute is causing perfomance issues');
+  });
+
+
   describe('deleteValueForProperty', () => {
     it('should remove attributes for normal properties', () => {
       const container = document.createElement('div');
--- a/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
+++ b/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
@@ -162,7 +162,12 @@ describe('DOMPropertyOperations', () => {
     const attributeValue = {foo: 'bar'};
     attributeValue.toString = function() {
       // for loop to waste time
-      for (let i = 0; i < 54321; i++) {}
+      let n=0;
+      for (let i = 0; i < 1e6; i++) {
+        n = (n * i + 1) % 1e6
+      }
+      invariant(n === 577020, "…");
+      
       let originalToString = Object.prototype.toString;
       return originalToString.apply(this);
     };
--- a/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
+++ b/packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
@@ -167,12 +167,12 @@ describe('DOMPropertyOperations', () => {
       return originalToString.apply(this);
     };
 
-     const nthPrime = n => {
+    const nthPrime = n => {
       let sieve = [2];
       let current = 3;
       let prime;
 
-       while (sieve.length < n) {
+      while (sieve.length < n) {
         current += 2;
         if (current % 2 === 0) {
           continue;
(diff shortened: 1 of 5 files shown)


kambleaa007:
How to resolve
Error: Don't import fbjs/lib/warning (found in C:\Users\ashish.kamble\GIT\react\mainReact\react\packages\react-dom\src\client\DOMPropertyOperations.js). Use the utilities in packages/shared/ instead.
    at resolveId (C:\Users\ashish.kamble\GIT\react\mainReact\react\scripts\rollup\build.js:289:15)
    at promise.then.result (C:\Users\ashish.kamble\GIT\react\mainReact\react\node_modules\rollup\dist\rollup.js:1735:32)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

threepointone:
Can you explain when such a situation would arise? It seems fairly rare that a custom object like this is passed as a Dom attribute

packages/react-dom/src/client/DOMPropertyOperations.js
```diff
@@ -18,20 +18,40 @@ import {
 import sanitizeURL from '../shared/sanitizeURL';
 import {disableJavaScriptURLs} from 'shared/ReactFeatureFlags';
 
+import warning from '../shared/warnValidStyle';
```
threepointone:
```suggestion
import warning from 'shared/warning';
```
could you try this instead?

kambleaa007: Thanks, I am trying

packages/react-dom/src/client/DOMPropertyOperations.js
```diff
@@ -18,20 +18,40 @@ import {
...
+let stringifyWithPerformanceWarning;
+
 export function getValueForProperty(
   node: Element,
   name: string,
   expected: mixed,
   propertyInfo: PropertyInfo,
 ): mixed {
   if (__DEV__) {
+    stringifyWithPerformanceWarning = function(value) {
```
threepointone: Could you hoist this to a top level function? In prod, it won't have any references and will get removed, so it should be fine there.

packages/react-dom/src/client/DOMPropertyOperations.js
```diff
@@ -141,8 +161,13 @@ export function setValueForProperty(
       const attributeName = name;
       if (value === null) {
         node.removeAttribute(attributeName);
+      } else if (__DEV__) {
+        node.setAttribute(
+          attributeName,
+          stringifyWithPerformanceWarning(value),
```
threepointone: we could skip doing this with the warning if value.toString belongs to one of the primitives (Number, Boolean, String etc)

packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
```diff
@@ -157,6 +157,44 @@ describe('DOMPropertyOperations', () => {
     });
   });
 
+  it('should warn if custom attributes take too long to stringify', () => {
+    const container = document.createElement('div');
+    const attributeValue = {foo: 'bar'};
+    attributeValue.toString = function() {
+      // finds 2000th prime to waste time
```
threepointone: could you make this something simpler? Like a for loop that just runs many times. We shouldn't have to maintain a prime number generator in our tests :)

packages/react-dom/src/client/DOMPropertyOperations.js
kambleaa007: ok

packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
kambleaa007: added for loop

packages/react-dom/src/client/DOMPropertyOperations.js
kambleaa007: do you mean, directly doing `warning()` rather that `stringifyWithPerformanceWarning()`

packages/react-dom/src/client/DOMPropertyOperations.js
kambleaa007: @threepointone Do you know why my test cases failling, i just clones and `yarn test` its failing

packages/react-dom/src/__tests__/DOMPropertyOperations-test.js
```diff
@@ -157,6 +157,21 @@ describe('DOMPropertyOperations', () => {
     });
   });
 
+  it('should warn if attributes take too long to stringify', () => {
+    const container = document.createElement('div');
+    const attributeValue = {foo: 'bar'};
+    attributeValue.toString = function() {
+      // for loop to waste time
+      for (let i = 0; i < 54321; i++) {}
```
ooflorent: Actually this does not waste time because the loop does nothing. Most JavaScript engines detect this loop as a no-op and remove it from the codegen. Another way to waste time would be:

```js
let n = 0;
for (let i = 0; i < 1e6; i++) {
  n = (n * i + 1) % 1e6
}
invariant(n === 577020, "…")
```

What's important is to test `n` after the loop so we are sure the value has be computed.

kambleaa007: ok thanks

packages/react-dom/src/client/DOMPropertyOperations.js
haydaralrikabi:
```suggestion
import warning from '../shared/warnValidStyle';
import warning from 'shared/warning';
```
Try this

haydaralrikabi:
```suggestion
import warning from 'shared/warning';
```

packages/react-dom/src/client/DOMPropertyOperations.js
```diff
@@ -18,13 +18,32 @@ import {
 import sanitizeURL from '../shared/sanitizeURL';
 import {disableJavaScriptURLs} from 'shared/ReactFeatureFlags';
```
haydaralrikabi:
```suggestion
import warning from 'shared/warning';
```

PR #16320, opened 2019-08-08
[react-events] Refactor unit tests for Hover

necolas:
First pass at refactoring the tests and responders. This only touches `Hover`.

Demo: https://codesandbox.io/s/refactor-hover-responder-f9k2q

**Problem**

The existing responders listen to pointer events by default and add fallback events if PointerEvent is not supported. However, this complicates the responders and makes it easy to create a problematic unit test environment. jsdom doesn't support PointerEvent, which means that the responders end up listening to pointer events *and* fallback events in unit tests. This isn't a direct problem in production environments, because no browser will fire pointer events if they aren't supported. But in the unit test environment, we often dispatch event sequences taken from browsers that support pointer events. This means that what we're often testing is actually a (complex) scenario that cannot even occur in production: a responder that is listening to and receives both pointer events and fallback events. Not only does this risk making responders more complicated to implement but it could also hide bugs in implementations.

**Response**

Implement the responders so that they're only listening to *either* pointer events *or* fallback events, never both. This should make the default pointer events implementations significantly simpler and easier to test, as well as free to rely on the complete PointerEvents API. In the future it should also make DCE easier for target environments that are known to support PointerEvents, as we can use build tools with an equivalent of the runtime check. The fallback events (touch and mouse) need to coexist and be resilient to browser emulated events. Our unit tests should express a suite of high-level interactions that can be run in environments with and without PointerEvents support.

### DIFF
--- a/packages/react-events/src/dom/Hover.js
+++ b/packages/react-events/src/dom/Hover.js
@@ -10,6 +10,7 @@
 import type {
   ReactDOMResponderEvent,
   ReactDOMResponderContext,
+  PointerType,
 } from 'shared/ReactDOMTypes';
 import type {ReactEventResponderListener} from 'shared/ReactTypes';
 
@@ -38,6 +39,7 @@ type HoverState = {
 type HoverEventType = 'hoverstart' | 'hoverend' | 'hoverchange' | 'hovermove';
 
 type HoverEvent = {|
+  pointerType: PointerType,
   target: Element | Document,
   type: HoverEventType,
   timeStamp: number,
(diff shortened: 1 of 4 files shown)


packages/react-events/src/dom/Hover.js
```diff
@@ -237,95 +227,122 @@ const hoverResponderImpl = {
...
-              state.hoverTarget,
-            );
-            context.dispatchEvent(
-              syntheticEvent,
-              onHoverMove,
-              UserBlockingEvent,
-            );
+      case 'pointermove': {
+        if (state.isHovered) {
+          if (state.hoverTarget !== null) {
```
trueadm: can this line be merged with the line above?

PR #16325, opened 2019-08-08
Bump deps in packages/**/package.json

acdlite:


packages/jest-mock-scheduler/package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://reactjs.org/",
   "peerDependencies": {
     "jest": "^23.0.1",
-    "scheduler": "^0.11.0"
```
bvaughn: Should we really bump this one?

bvaughn: Was it intentional that we didn't raise it past 0.11 previously? Or was that a mistake?

acdlite: The API changed in 0.14 so yeah. It's private anyway so doesn't really matter.

acdlite:
I know sizebot I'm fixing it!

PR #16328, opened 2019-08-09
Upgrade ESLint so we can use JSX Fragment syntax

acdlite:
Now that we're using Babel 7, this is the last blocker.


.eslintrc.js
```diff
@@ -65,6 +60,9 @@ module.exports = {
     'no-var': ERROR,
     strict: ERROR,
 
+    // Enforced by Prettier
+    'max-len': OFF,
```
acdlite: I was getting some weird syntax error in a transitive module. Google didn't turn up anything so I decided to turn it off since the length is enforced by Prettier, anyway.

acdlite:
^ Oops didn't mean to delete that. Sizebot says the base commit is broken because the coverage went down slightly after the Babel 7 PR.

.eslintrc.js
threepointone: Prettier won’t warn for long strings tho, and won’t rewrite them either. Which... may be fine?

SimenB: fwiw, this also yelled at me for long line comments, which prettier doesn't touch

acdlite: I'll add a comment so the next person who looks at it knows why I turned it off.

PR #16334, opened 2019-08-09
[Flare] Rework the responder dispatching/batching mechanism

trueadm:
We want to change the event system so it is possible to handle and manage ownership/propagation of event responders as we process the fiber tree. In order to do this, we need to change how the current event responder system works so that events that are queued to dispatch happen after the processing of each individual responder, rather than at the end.

Previously, we built up an array of all user events that needed to be dispatched, along with their priority and dispatched them all together. This made it impossible for user events to have effects that make it possible to control propagation and ownership though, as by this stage, we've already processed the responders. To illustrate this:

```
Responder A
  ->
    Responder B
      ->
        Responder C
```

*Old Model:*

If a native event happens and Responder C gets activated, we then bubble to Responder B and then to Responder A. Along the way, each responder might `dispatchEvent`, so we'd store the user events in an array. Once all Responders A, B, C have been handled we then flush all the events we've collected.

*New Model:*

If a native event happens and Responder C gets activated, we then bubble to Responder B and then to Responder A. Along the way, each responder might `dispatchEvent`, if so, we now dispatch the user events at that point, rather than having to queue them up to flush at the end. We are able to do this by wrapping the entire event process in `batchedEventUpdates` and having a dedicated event that sets a flag for when we're in the process of executing a user event. The priority logic from before remains unchanged because of these tweaks.

I also removed the global ownership model, as this new mechanism allows for a much more refined approach that simply globally saying a responder is the only owner. It also conflicts with how this new mechanism works.


trueadm:
@necolas @acdlite @sebmarkbage Would be good to get a review on this PR when you folks have a chance.

PR #16332, opened 2019-08-09
chore: add eslint-plugin-jest's valid-expect rule

SimenB:
~There are some assertions here which just do `expect(thing)` without a matcher, meaning it's not actually asserting anything (it does not work like `assert`). I added `.toBe(true)` to them, and 2 of the tests are failing, meaning they are either wrong, or false negatives on master.~

The above is dated now that #16336 was merged, but it _did_ uncover them 🙂

PR #16336, opened 2019-08-09
fix some missing assertions

threepointone:
These were discovered by @SimenB in https://github.com/facebook/react/pull/16332. We weren't making actual assertions on some values. This PR makes the assertions, and fixes the tests.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
@@ -1034,13 +1034,13 @@ describe('ReactDOMComponent', () => {
       return (str + '').replace(/([.?*+\^$\[\]\\(){}|-])/g, '\\$1');
     }
 
-    function toHaveAttribute(actual, expected) {
+    function expectToHaveAttribute(actual, expected) {
       const [attr, value] = expected;
       let re = '(?:^|\\s)' + attr + '=[\\\'"]';
       if (typeof value !== 'undefined') {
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/__tests__/ReactDOMComponent-test.js
```diff
@@ -1034,33 +1034,31 @@ describe('ReactDOMComponent', () => {
...
 
-    function toHaveAttribute(actual, expected) {
+    function expectToHaveAttribute(actual, expected) {
       const [attr, value] = expected;
       let re = '(?:^|\\s)' + attr + '=[\\\'"]';
       if (typeof value !== 'undefined') {
         re += quoteRegexp(value) + '[\\\'"]';
       }
-      return new RegExp(re).test(actual);
+      expect(new RegExp(re).test(actual)).toBe(true);
```
SimenB: this _could_ be `expect(actual).toMatch(new RegExp(re))` fwiw. If it fails, it'll print `actual` instead of `expected false to be true`

threepointone: Could you include this in https://github.com/facebook/react/pull/16332?

SimenB: Certainly! Will update tomorrow

SimenB: Done

PR #16338, opened 2019-08-09
Release script supports publishing a subset of packages

bvaughn:
Sometimes it's useful to skip publishing one or more packages. In the past, we've done this by temporarily editing the local `package.json` to mark it as a "private" package, but this is kind of janky. The release script should support this since it can add some additional automated checks to make sure we don't publish something invalid.

### Step 1: Prepare canary
In this case, we still prepare all of the packages. (This greatly simplifies the following script steps.) For the packages we plan to "skip" as part of publishing, we just confirm the already published version numbers.
step-1-updatedKapture 2019-08-09 at 11 13 26

### Step 2: Publish filtered set of packages
When it is time to publish, we can specify one or more packages as "skipped". The updated script will confirm that we really want to skip them, and ensure that doing so wouldn't leave us in an invalid state (depending on something that hasn't already been released).

step-2-updated-Kapture 2019-08-09 at 11 39 38

Note that this is still an advanced technique and something that should be done carefully and with consideration.

### Guard against typos
invalid-package-Kapture 2019-08-09 at 9 21 03

### Guard against publishing a dependency on an unpublished version
invalid-skipped-version-Kapture 2019-08-09 at 9 22 35

### DIFF
--- a/scripts/release/create-canary-commands/build-artifacts.js
+++ b/scripts/release/create-canary-commands/build-artifacts.js
@@ -21,5 +21,5 @@ const run = async ({cwd, dry, tempDirectory}) => {
 };
 
 module.exports = async params => {
-  return logPromise(run(params), 'Building artifacts', 420000);
+  return logPromise(run(params), 'Building artifacts', 600000);
 };
--- a/scripts/release/publish-commands/print-follow-up-instructions.js
+++ b/scripts/release/publish-commands/print-follow-up-instructions.js
@@ -9,12 +9,10 @@ const {join} = require('path');
 const theme = require('../theme');
 const {execRead} = require('../utils');
 
-const run = async ({cwd, packages, skipPackages, tags}) => {
+const run = async ({cwd, packages, tags}) => {
   // All packages are built from a single source revision,
   // so it is safe to read build info from any one of them.
-  const arbitraryPackageName = packages.find(
-    name => !skipPackages.includes(name)
-  );
+  const arbitraryPackageName = packages[0];
   const {commit, environment, reactVersion} = readJsonSync(
     join(cwd, 'build', 'node_modules', arbitraryPackageName, 'build-info.json')
   );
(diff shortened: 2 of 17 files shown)


scripts/release/create-canary-commands/build-artifacts.js
```diff
@@ -21,5 +21,5 @@ const run = async ({cwd, dry, tempDirectory}) => {
 };
 
 module.exports = async params => {
-  return logPromise(run(params), 'Building artifacts', 420000);
+  return logPromise(run(params), 'Building artifacts', 600000);
```
bvaughn: Technically this change was unrelated but I noticed the previous default value was not very accurate, given the current average build time. (We've added more packages since the script was written.)

scripts/release/publish-commands/print-follow-up-instructions.js
```diff
@@ -50,7 +46,13 @@ const run = async ({cwd, packages, tags}) => {
         const packageName = packages[i];
         console.log(theme.path`• packages/%s/package.json`, packageName);
       }
-      console.log(theme.path`• packages/shared/ReactVersion.js`);
+      const status = await execRead(
+        'git diff packages/shared/ReactVersion.js',
+        {cwd}
+      );
+      if (status) {
```
bvaughn: It's possible this version wasn't actually updated (depending on what we skipped) so I added a check to avoid printing a confusing message.

bvaughn:
For example, this is the sequence of commands I plan to use to publish `use-subscription` from the latest canary.

#### Promote canary `use-subscription`
```sh
scripts/release/prepare-stable.js \
  --skipTests \
  --skipPackages \
      react \
      react-refresh \
      react-is \
      react-dom \
      react-test-renderer \
      create-subscription \
      scheduler \
      eslint-plugin-react-hooks \
      jest-react \
      react-art \
      react-reconciler
```

#### Publish `use-subscription`
```sh
scripts/release/publish.js \
  --tags next latest \
  --skipPackages \
      react \
      react-refresh \
      react-is \
      react-dom \
      react-test-renderer \
      create-subscription \
      scheduler \
      eslint-plugin-react-hooks \
      jest-react \
      react-art \
      react-reconciler
```

scripts/release/publish-commands/validate-skip-packages.js
```diff
@@ -0,0 +1,59 @@
...
+    return;
+  }
+
+  const validateDependencies = async (name, dependencies) => {
+    if (!dependencies) {
+      return;
+    }
+
+    for (let dependency in dependencies) {
+      // Do we depend on a package thas has been skipped?
```
acdlite: Yay thank you!

PR #16340, opened 2019-08-09
Cleanup Babel PR (ReactFreshPlugin)

lunaruan:
This PR addresses the comments in #16297 

In particular, for `ReactFreshBabelPlugin`, @gaearon mentioned that we could modify the `ReactFreshBabelPlugin` to not throw an error if it is run in a test (if `expect` is a function it is most likely in a test environment) instead of setting a precedent for modifying the jest config.


packages/react-refresh/src/ReactFreshBabelPlugin.js
```diff
@@ -11,7 +11,7 @@ export default function(babel) {
   if (typeof babel.getEnv === 'function') {
     // Only available in Babel 7.
     const env = babel.getEnv();
-    if (env !== 'development') {
+    if (env !== 'development' && typeof expect !== 'function') {
```
acdlite: @gaearon My understanding was that the purpose of this error is to prevent people from enabling the Fresh plugin in their unit tests. Doesn't this change mean that consumers of this plugin who use Jest won't get the error below?

gaearon: We actually do run it in tests in fbsource. So I don't know that it's necessarily a bad idea.

The primary purpose is to prevent people from including it in prod setups.

PR #16343, opened 2019-08-09
[react-events] Focus/FocusWithin responders with fallbacks

necolas:
Separate the PointerEvent and fallback implementations.
Fix the unit tests to cover both PointerEvent and non-PointerEvent environments.
Fix the focus-visible related callbacks to get called when keys other than "Tab" are used.

### DIFF
--- a/packages/react-events/src/dom/Focus.js
+++ b/packages/react-events/src/dom/Focus.js
@@ -33,6 +33,7 @@ type FocusState = {
   isFocused: boolean,
   isFocusVisible: boolean,
   pointerType: PointerType,
+  isEmulatingMouseEvents: boolean,
 };
 
 type FocusProps = {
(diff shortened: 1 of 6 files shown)


packages/react-events/src/dom/Focus.js
```diff
@@ -66,25 +67,12 @@ const isMac =
...
-    'touchstart',
-    'touchend',
-  );
-}
+const hasPointerEvents =
+  typeof window !== 'undefined' && window.PointerEvent != null;
+
+const rootEventTypes = hasPointerEvents
+  ? ['keydown', 'keyup', 'pointermove', 'pointerdown', 'pointerup']
+  : ['keydown', 'keyup', 'mousedown', 'touchmove', 'touchstart', 'touchend'];
```
TrySound: mouseup is missing. Is it intentional?

PR #16346, opened 2019-08-10
Refactor Partial Hydration

sebmarkbage:
This is a pretty invasive refactor of the SuspenseComponent and DehydratedSuspenseComponent.

The primary purpose of this refactor is to avoid the hacky "upgrade" and "downgrade" by mutating the tag.

In the new model, a boundary is always represented by the same SuspenseComponent fiber. Inside it I store a "DehydratedFragment". The inner fiber represents the dehydrated nodes in the tree. This can be used to delete the whole thing or insert before it.

I also switched legacy mode to always client-render the boundary content since we can't hydrate partially in legacy mode. It also warns.

Otherwise, this (hopefully) shouldn't have an semantic differences.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -36,7 +36,6 @@ import {
   Profiler,
   SuspenseComponent,
   SuspenseListComponent,
-  DehydratedSuspenseComponent,
   MemoComponent,
   SimpleMemoComponent,
   LazyComponent,
@@ -1526,7 +1525,10 @@ function updateSuspenseComponent(
     workInProgress.effectTag &= ~DidCapture;
   } else {
     // Attempting the main content
-    if (current === null || current.memoizedState !== null) {
+    if (
+      current === null ||
+      (current.memoizedState: null | SuspenseState) !== null
+    ) {
       // This is a new mount or this boundary is already showing a fallback state.
       // Mark this subtree context as having at least one invisible parent that could
       // handle the fallback state.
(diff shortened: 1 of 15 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1646,26 +1643,121 @@ function updateSuspenseComponent(
...
     // ensure the state of the primary children is preserved.
-    const prevState = current.memoizedState;
-    const prevDidTimeout = prevState !== null;
-    if (prevDidTimeout) {
+    const prevState: null | SuspenseState = current.memoizedState;
+    if (prevState !== null) {
+      if (enableSuspenseServerRenderer) {
+        const dehydrated = prevState.dehydrated;
+        if (dehydrated !== null) {
+          if (!didSuspend) {
```
acdlite: More comments

scripts/error-codes/codes.json
```diff
@@ -338,5 +338,7 @@
   "337": "An invalid event responder was provided to host component",
   "338": "ReactDOMServer does not yet support the fundamental API.",
   "339": "An invalid value was used as an event responder. Expect one or many event responders created via React.unstable_createResponer().",
-  "340": "An invalid value was used as an event listener. Expect one or many event listeners created via React.unstable_useResponer()."
+  "340": "An invalid value was used as an event listener. Expect one or many event listeners created via React.unstable_useResponer().",
```
acdlite: @trueadm Noticed a typo here. Should be `unstable_useResponder()`

trueadm: I’m on PTO. cc @necolas

acdlite: Oh yeah, sorry. I mean I can do it too :D

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1511,21 +1511,23 @@ function updateSuspenseComponent(
...
     // rendering the fallback children.
-    nextState = SUSPENDED_MARKER;
     nextDidTimeout = true;
     workInProgress.effectTag &= ~DidCapture;
   } else {
     // Attempting the main content
-    if (current === null || current.memoizedState !== null) {
+    if (
+      current === null ||
+      (current.memoizedState: null | SuspenseState) !== null
```
bvaughn: Why is this necessary? ^

scripts/error-codes/codes.json
acdlite: https://github.com/facebook/react/pull/16373

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: It's so that it becomes greppable.

Once we add types for fields like memoizedState it will also be used to ensure that those types are correct.

sebmarkbage: The type SuspenseState should probably be nullable by itself since it has meaning and would be a variant if we had those.

PR #16352, opened 2019-08-10
[Partial Hydration] Attempt hydration at a higher pri first if props/context changes

sebmarkbage:
__Builds on top of #16346__

I attempted this in an early PR but the approach of using mutation to change fiber tags was broken and lead to bugs. Now that's fixed in #16346 and I can finally rebase this fix.

If we haven't yet hydrated a boundary, we may still get props or context flowing into it. Currently, the delete the content and replace it with a content render in that case. Dropping events and state as a result.

When we're already rendering the future state, it is too late to hydrate since we can't continue rendering down with the new props/context as it wouldn't yield the same result as what was rendered on the server. However, assuming people use immutable data for hydration as they should, then we can solve this problem. We solve it by "going back in time" to an earlier point on the timeline and hydrate using those props/context. After that is done, we then render the new props/context on top of it.

In practice, this just means that we increase the priority of the hydration to slightly higher than the render that otherwise would have to delete the content.

We also mark this as a "bad" loading state which means that we'll also take advantage of our maximum allowed suspense time to wait for any remaining data/code that we'd need to complete the hydration. Meaning that we can do updates to parents before we have the code to the children and it still looks seamless to the user!

### DIFF
--- a/fixtures/ssr/src/components/Chrome.js
+++ b/fixtures/ssr/src/components/Chrome.js
@@ -26,7 +26,16 @@ export default class Chrome extends Component {
           <Theme.Provider value={this.state.theme}>
             {this.props.children}
             <div>
-              <ThemeToggleButton onChange={theme => this.setState({theme})} />
+              <ThemeToggleButton
+                onChange={theme => {
+                  React.unstable_withSuspenseConfig(
+                    () => {
+                      this.setState({theme});
+                    },
+                    {timeoutMs: 6000}
+                  );
+                }}
+              />
             </div>
           </Theme.Provider>
           <script
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -171,7 +171,7 @@ import {
 import {
   markSpawnedWork,
   requestCurrentTime,
-  retryTimedOutBoundary,
+  retryDehydratedSuspenseBoundary,
   scheduleWork,
 } from './ReactFiberWorkLoop';
 
@@ -2081,7 +2081,7 @@ function updateDehydratedSuspenseComponent(
     // Register a callback to retry this boundary once the server has sent the result.
     registerSuspenseInstanceRetry(
       suspenseInstance,
-      retryTimedOutBoundary.bind(null, current),
+      retryDehydratedSuspenseBoundary.bind(null, current),
     );
     return null;
   } else {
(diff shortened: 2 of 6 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1946,11 +2038,31 @@ function updateDehydratedSuspenseComponent(
...
+    // hydrate it. We might still be able to hydrate it using an earlier expiration time, if
+    // we are rendering at lower expiration than sync.
+    if (renderExpirationTime < Sync) {
+      if (suspenseState.retryTime <= renderExpirationTime) {
+        // This render is even higher pri than we've seen before, let's try again
+        // at even higher pri.
+        let attemptHydrationAtExpirationTime = renderExpirationTime + 1;
+        suspenseState.retryTime = attemptHydrationAtExpirationTime;
+        scheduleWork(current, attemptHydrationAtExpirationTime);
+        // TODO: Early abort this render.
```
sebmarkbage: @acdlite We don't currently have a way to abort a render while inside the work loop. I attempted something like https://github.com/sebmarkbage/react/commit/c9c6c0a4c57e9fa04adb9890ac9c3ca6ecebfcdc but I don't think that's the right approach.

This works fine as long as we yield every 5ms anyway but it's not ideal when there are longer work loop times e.g. due to isInputPending or similar.

fixtures/ssr/src/components/Chrome.js
```diff
@@ -26,7 +26,16 @@ export default class Chrome extends Component {
...
+              <ThemeToggleButton
+                onChange={theme => {
+                  React.unstable_withSuspenseConfig(
+                    () => {
+                      this.setState({theme});
+                    },
+                    {timeoutMs: 6000}
+                  );
+                }}
+              />
```
sebmarkbage: To run this fixture you need to turn on enableSuspenseServerRenderer in ReactFeatureFlags.

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: You could throw something and then handle the interruption in `ReactFiberThrow`

acdlite: Do you need to call `markSpawnedWork`?

sebmarkbage: I was thinking about that and I don't think I should call `markSpawnedWork` here because the hydration isn't actually part of this work that caused this update. The hydration is the work that is scheduled on Idle which isn't really part of this update.

Issue #16342, opened 2019-08-09
textarea does not show warning when switching from uncontrolled to controlled like inputs do


bradwestfall:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

While things like `<input>` correctly get a warning when switching from uncrontrolled to controlled, I'm noticing `<textarea>` does not

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

Here's a codesandbox. Type in the input field, we see error (correct), change to textarea and start over, type in field and we don't see the error (incorrect I think) https://codesandbox.io/s/recursing-dawn-jls8i

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

16.8

kdreagle:
Taking a crack at this as my first PR. Currently trying to get all the tests to pass.

karthikpriyadarshan-alation:
The PR is still open. Wondering if I can work on this since it's been open for a while?

karthikkp:
The PR is still open. Wondering if I can work on this since it's been open for a while?

bradwestfall:
@karthikkp I'd say if react team doesn't reply soon, just go ahead and submit a PR

wapgear:
I have a focused fix ready for the uncontrolled-to-controlled textarea warning and will submit a PR shortly.

PR #16354, opened 2019-08-11
Warn when textarea switches between controlled and uncontrolled

kdreagle:
Fixes issue #16342 

I think the select element may need this warning as well. If so I will probably create a PR for that shortly.


gaearon:
Sorry, this slipped through the cracks. Looks like https://github.com/facebook/react/pull/18681 picked it up with an updated message.

gaearon:
Thanks for the PR!

Issue #16003, opened 2019-06-27
[eslint-plugin-react-hooks] Crash when referencing "arguments"


Jessidhia:
**Do you want to request a *feature* or report a *bug*?**

Report a bug

**What is the current behavior?**

Referencing `arguments` from inside an arrow function (i.e. the `arguments` from the nearest non-arrow function) causes a crash in the eslint plugin.

```
TypeError: Cannot read property 'type' of undefined
Occurred while linting /.../src/react.tsx:92
    at gatherDependenciesRecursively (.../node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js:1032:21)
    at visitFunctionExpression (.../node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js:985:7)
    at .../node_modules/eslint/lib/util/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (.../node_modules/eslint/lib/util/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (.../node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (.../node_modules/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (.../node_modules/eslint/lib/util/node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (.../node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:632:23)
    at .../node_modules/eslint/lib/linter.js:752:32
```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

This doesn't reproduce in codesandbox as CRA hasn't been updated yet.

This snippet is enough to cause the crash:

```js
function useMyHook(/*...*/) {
  useEffect(() => {
    arguments // crash because reference.resolved.defs is empty
  }, [])
}
```

This, however, does not crash:

```js
function useMyHook(/*...*/) {
  useEffect(function() {
    arguments // ok
    return () => arguments // also ok
  }, [])
}
```

[...]

PR #16356, opened 2019-08-11
[eslint-plugin-react-hooks] Fixed crash when referencing arguments in arrow functions.

hristo-kanchev:
**Description**:
Referencing arguments from inside an arrow function caused a crash.

Fixes issue: #16003

### DIFF
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -399,7 +399,6 @@ export default {
 
           const def = reference.resolved.defs[0];
 
-          // Ignore if def is undefined - e.g. `arguments` binding.
           if (def == null) {
             continue;
           }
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -397,8 +397,14 @@ export default {
             });
           }
 
-          // Ignore references to the function itself as it's not defined yet.
           const def = reference.resolved.defs[0];
+
+          // Ignore if def is undefined - e.g. `arguments` binding.
+          if (def === undefined) {
```
gaearon: Let's make this sloppy `== null` and then remove `!= null` from next comparison.

hristo-kanchev: Good shout! I've updated the PR.

gaearon:
Thanks

PR #16359, opened 2019-08-11
Reset didReceiveUpdate in beginWork

sebmarkbage:
This is a bad bug. It means that we sometimes inherit
didReceiveUpdate from a previous component's begin.

Effectively this only means that we're overrendering in some cases.

We should refactor to get rid of this as a global flag.

I don’t have a test case because it only showed up as an intermediate state in another test. Maybe we can come up with something simplified.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -2846,6 +2846,8 @@ function beginWork(
         workInProgress,
         renderExpirationTime,
       );
+    } else {
+      didReceiveUpdate = false;
     }
   } else {
     didReceiveUpdate = false;


acdlite:
Chatted offline. This is a bug but it only affects the hooks bailout optimization when hydrating, and only if all of the following are true:

1) No new props
2) There's a pending update on the fiber
3) The update bails out during the render phase, which means it can't have bailed out due to the _eager_ bailout optimization (the one that happens inside `setState`)
4) The previous fiber that was processed also received an update.

So it's pretty rare. I'll wrote a test case. I'll submit separately.

`didReceiveUpdate` is also used by Suspense when hydrating, but I don't think it can affect that because of 3.

sebmarkbage:
A test case that hits this for Suspense is added in https://github.com/facebook/react/pull/16369

PR #16363, opened 2019-08-12
Add missing valid unitless CSS Property numbers

brandonkal:
I have added missing CSS properties that produce valid CSS without a unit.


trueadm:
I'm not sure we want to add more of these, as they add to the code size of ReactDOM for no real gain (you can easily add a `px` suffix in your app code). In the future, it might be that we remove all unitless CSS property numbers, so I'm going to close this for now. We can always revisit in the future.

Issue #16366, opened 2019-08-12
`act` should yield the value returned from the async function.


pfgray:
**Do you want to request a *feature* or report a *bug*?**
Feature

A couple of things have surprised me about the new asynchronous `act` API.

#### 1. Act doesn't return a promise.

Currently the `act` function returns an object with a `then` method that doesn't return a promise.

Recently, I tried to write something like:

```js
act(() => {
  return doSomethingAsynchronous()
})
.then(() => {
  // ...
})
.then(() => {
  // ...
})
```

The second `.then` call throws an error because the value that is returned from `act` isn't a full-fledged `Promise`, but rather just an object that contains a `.then` method that returns `undefined`. This is fine for simple `async`/`await` cases, but definitely caught me by surprise.

#### 2. Act discards the returned value.

Act also throws away the value returned from the async function callback, which makes it harder to use the values generated there. For example, if I wanted to use the value yielded by `doSomethingAsynchronous`, I would need to introduce a mutable variable:

```js
let returnedValue;
await act(async () => {
  returnedValue = await doSomethingAsynchronous()
})

// use returnedValue here
```

If `act` simply propogates the returned value from the asynnchronous callback, this will solve the confusing bits above, allowing for things like:

```
const returnedValue = await act(() => doSomethingAsynchronous())
```

threepointone:
I thought about this, and I empathise that it's not as clean as it could be, but the role of act() is not to represent a value/future, and it doesn't compose like it either. (For example, you can't have 2 async act scopes in parallel) It's to establish a 'scope' within which things happen, so they can be batched and executed as one, and guarantee a side effect (flushing effects and updates) after it exits. In a language like javascript that doesn't have support for "zones", this api is the best we can do. I would not want to make this function return a value just to save some characters. 


The workaround though, is pretty simple - make a helper that does what you want -
```jsx
import TestUtils from 'react-dom/test-utils'

async function act(callback){
  let ret;
  await TestUtils.act(async () => {
    ret = await callback();
  });
  return ret;
}

// ...

const returnedValue = await act(() => doSomethingAsynchronous())
 ```

pfgray:
Thanks for the workaround! I'm already employing something very simliar in my project.

[...]

PR #16367, opened 2019-08-12
Changing async act() to propagate returned value from async callback

pfgray:
Addresses #16366


threepointone:
Closing this since we closed the original issue too. But I promise it's on my mind.

PR #16379, opened 2019-08-13
Remove "Waiting for async callback" User Timing measurement

gaearon:
This measurement is already broken (mark/measure calls for it mismatch in Concurrent Mode). This makes debugging very annoying because "break on caught exceptions" stops there all the time.

It's difficult to keep it up to date because it models a pause rather than an actual work slice. This is why we keep breaking it. I propose that we just remove this timing, and fix it forward with Root Events.


gaearon:
The test warning I added uncovered a problem in _production_ mode which is fixed by the second commit. Note that it doesn't affect open source, but it affects Ads which uses User Timing for a subset of users in prod environment.

gaearon:
@acdlite said he has a WIP replacement for this on Scheduler level anyway

PR #16381, opened 2019-08-13
Import React DevTools v4 [DO NOT MERGE]

bvaughn:
The goal of this import is to preserve history from the DevTools v4 fork:
https://github.com/bvaughn/react-devtools-experimental

# Checklist
- [x] Import (with history preserved)
- [ ] Merge dependencies into root `package.json`/`yarn.lock` somehow
- [ ] Test all builds and artifacts
- [ ] Make sure CI doesn't break

# Preparation
Here's what I did...

## Prepare fork for merge
```sh
mkdir ~/Desktop/devtools-merge
cd ~/Desktop/devtools-merge

git init .
git commit -m "Initializing empty merge repo" --allow-empty  

git remote add -f fork git@github.com:bvaughn/react-devtools-experimental.git
git merge fork/master --allow-unrelated-histories

# Moved everything except "packages" into a new "extension" subdirectory
# to avoid merge conflicts with React repo folder structure
```

## Import into React
```sh
cd ~/Documents/git/react

git remote add devtools-merge ~/Desktop/devtools-merge
git fetch devtools-merge

git checkout -b devtools-v4-merge
git merge --allow-unrelated-histories devtools-merge/master

git remote rm devtools-merge
```

## Test history
It should now be possible to follow the history of even files that have been moved, e.g.
```sh
git log --follow extension/package.json 
```


SimenB:
You might need to tweak https://github.com/facebook/react/blob/868d02d6c65eb72da9013ab6c00d04a988d3a776/scripts/jest/preprocessor.js#L59-L72 to support your extra babel config file

bvaughn:
@SimenB Thanks for the pointer.

My original goal was to use a totally separate Jest entry point (with its own config) rather than trying to overload the already complicated root Jest configuration.

But that was breaking for reasons I couldn't understand, and assumed was related to the fact that the package wasn't being seen as "module" (because of its folder structure). After seeing your comment on the other pR, I think it might have actually been caused by `react-is` (since that's the package that was erroring).

I think I'll just try to add another fork in the root though because I think I exhausted the other idea.

bvaughn:
Just synced in the post 4.0.0 launch changes (https://github.com/bvaughn/react-devtools-experimental/compare/6eb04b2..2843a15)

bvaughn:
This PR is getting close to being landable, I think. Some of the new tests are failing though because they depend on both react DOM (to render the test app/code) and test renderer (to render DevTools). In particular, usage of hooks inside of test renderer causes the tests to fail.

MichaelDeBoey:
🎉🎉🎉

PR #16383, opened 2019-08-13
Partial support for React.lazy() in server renderer.

leebyron:
Provides partial support for React.lazy() components from the existing PartialRenderer server-side renderer.

Lazy components which are already resolved (or rejected), perhaps with something like `react-ssr-prepass`, can be continued into synchronously. If they have not yet been initialized, they'll be initialized before checking, opening the possibility to exploit this capability with a babel transform. If they're pending (which will typically be the case for a just initialized async ctor) then the existing invariant continues to be thrown.

### DIFF
--- a/packages/shared/ReactLazyComponent.js
+++ b/packages/shared/ReactLazyComponent.js
@@ -7,6 +7,8 @@
  * @flow
  */
 
+import warning from 'shared/warning';
+
 export type Thenable<T, R> = {
   then(resolve: (T) => mixed, reject: (mixed) => mixed): R,
 };
@@ -25,6 +27,7 @@ type ResolvedLazyComponent<T> = {
   _result: any,
 };
 
+export const Uninitialized = -1;
 export const Pending = 0;
 export const Resolved = 1;
 export const Rejected = 2;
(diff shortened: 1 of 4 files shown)


sebmarkbage:
I noticed you removed the `enableSuspenseServerRenderer` branch. Any particular reason for that? It seemed ok.

packages/shared/ReactLazyComponent.js
```diff
@@ -34,3 +37,40 @@ export function refineResolvedLazyComponent<T>(
 ): ResolvedLazyComponent<T> | null {
   return lazyComponent._status === Resolved ? lazyComponent._result : null;
 }
+
+export function initializeLazyComponentType(
+  lazyComponent: LazyComponent<any>,
+): void {
+  if (lazyComponent._status === Uninitialized) {
```
sebmarkbage: This could be a breaking change if this was using a non-matching version of React. That was the case in React 16.5.0 where this field was called _reactStatus. In this case I that release had lazy behind a feature flag. So iit's probably fine.

We typically are more resilient and check truthy or non-undefined or something like that and then add an expando in that case so that it at least works.

leebyron:
For the enableSuspenseServerRenderer branch, I wasn’t sure of the right behavior or test plan and figured it would be better to keep this one smaller and follow up with that change in another PR after this one

PR #16384, opened 2019-08-13
Make component stack last argument for deprecation warnings

gaearon:
Component stack is the last argument when we use `warning()`.

There are a few cases where we append stack manually. (Maybe some of them don't need to.) Those were originally written to be more readable. However, increasingly we're using tooling that treats last argument specially if it looks like a component stack. Both in RN and in FB5. We'll likely do the same in CRA/Next/etc.

So let's always put component stack last consistently. This makes it easier to build richer UIs and looks decent in console anyway.

Screen Shot 2019-08-13 at 10 57 24 PM


gaearon:
We should probably just migrate the rest to `warning()`.

PR #16388, opened 2019-08-14
Rename legacy "events" package to "legacy-events"

bvaughn:
This package conflicts with the built-in `events` Node package. We already had to add special casing for it, and I was having to add more in #16381 (since DevTools depends on `events`) so let's just rename it?


ashtonsix:
hey @bvaughn, does "legacy" in the package name indicate React will change how synthetic events  work soon?

PR #16391, opened 2019-08-14
Work around a Scheduler bug

gaearon:
- Comments out the broken cancellation logic until we figure out why.
- Fixes copy paste mistakes.

Follow-up to #16145 which introduced them.

### DIFF
--- a/packages/scheduler/src/forks/SchedulerHostConfig.default.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.default.js
@@ -406,7 +406,8 @@ if (
   };
 
   cancelHostCallback = function() {
-    scheduledHostCallback = null;
+    // TODO: Uncomment when we know why this breaks www.
+    // scheduledHostCallback = null;
   };
 
   requestHostTimeout = function(callback, ms) {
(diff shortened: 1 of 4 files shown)


packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -406,7 +406,8 @@ if (
   };
 
   cancelHostCallback = function() {
-    scheduledHostCallback = null;
+    // TODO: Uncomment when we know why this breaks www.
+    // scheduledHostCallback = null;
```
sebmarkbage: I'm not comfortable with this. This line wasn't added in the PR and there are other semantic changes in that PR but it was supposed to be an observational only change. The change also doesn't seem to be fully encapsulated behind the feature flag.

E.g. I don't understand this behavior change: https://github.com/facebook/react/commit/a34ca7bce69c2f321daef0b8650ad6e7cfce366a#diff-603a307ec39e05daabd1c651dc2ffb15R407

I should've reviewed more carefully.

gaearon:
This actually looks insufficient. The first problem went away but I found another one (a freeze). Let's revert.

PR #16392, opened 2019-08-14
Revert "[Scheduler] Profiling features"

gaearon:
Reverts facebook/react#16145

There are a few bugs we need to solve first.


sebmarkbage:
whoa. I didn't see that this affected `react.production.min.js` so much. That seems like another bug.

PR #16397, opened 2019-08-15
[react-events] Press: improve test coverage

necolas:
1. Run the tests in environments with and without PointerEvent support.
2. Improve test coverage to include both mouse and touch pointers.
3. Change `Press` responder so that it only listens to either pointer events or fallbacks events.

### DIFF
--- a/packages/react-events/src/dom/__tests__/Press-test.internal.js
+++ b/packages/react-events/src/dom/__tests__/Press-test.internal.js
@@ -9,6 +9,21 @@
 
 'use strict';
 
+import {
+  click,
+  dispatchPointerCancel,
+  dispatchPointerDown,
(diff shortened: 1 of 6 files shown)


packages/react-events/src/dom/__tests__/Press-test.internal.js
```diff
@@ -438,22 +310,34 @@ describe('Event responder: Press', () => {
         return <div ref={ref} listeners={listener} />;
       };
       ReactDOM.render(<Component />, container);
+      document.elementFromPoint = () => ref.current;
+    });
+
+    it('is called after pointer down and up: mouse', () => {
```
necolas: Collapse this test and the next by using `it.each(pointerTypesTable)`

Issue #3473, opened 2015-03-20
How Much XSS Vulnerability Protection is React Responsible For?


sebmarkbage:
There was a security hack that mentions React here: http://danlec.com/blog/xss-via-a-spoofed-react-element

Ultimately this is a server-side bug and NOT a bug in React itself. This issue is about figuring out if there is something we can do to mitigate issues when you have a JSON parsing bug or some server-side issue.

## isValidElement

React is designed to work with plain objects as input, and in fact, we're even getting rid of the `_isReactElement` as a way to verify this. We'll allow any JSON object. IMO, there is no problem with the verification here.

All string values are sanitized before inserted into the DOM (except for CSS styles which is a known wontfix issue for IE8).

In earlier versions we used `instanceof` checks but that didn't work well with multiple Reacts, it makes it difficult to optimize (inline objects are much faster) and couples JSX permanently to React, which we would like to avoid.

## dangerouslySetInnerHTML

One possible solution is to disable this feature and require it to be used imperatively (`React.findDOMNode(ref).innerHTML = ''`) which makes for worse performance at insertion time.

However, I don't believe this is the only bad thing once you can insert arbitrary HTML tags. It is certainly the easiest way to gain access to XSS though. You can also insert arbitrary Web Components which could expose data. You can render form elements that can potentially pass data.

## What else can we do?

Ultimately this is an issue where `<div>{userData}</div>` seems like a valid use case, but if your userData is compromised, it becomes dangerous.

Should React be responsible for protecting itself against arbitrary JSON as children?

yukinying:
Hi, Albert from Yahoo Security team here. I was working with @mridgway on studying this issue. Just wanna share more thoughts. 

Re: disabling dangeriouslySetInnerHTML, it is a good move, but it would not be sufficient. React allow inject "script" and "style" tag, or "style" attributes. All of those would allow script execution. Also, in general, we worried about UI redressing attack that attackers can create arbitrary overlay on a page that could steal all the input submitted.

sebmarkbage:
I don't think "script" tags will execute because we use innerHTML to create them. Do the "style" tag and "style" attributes allow you to execute (in context) scripts in modern browsers? I think it is only older versions of IE.

Regardless, the UI redressing attack is still an equally valid concern.

yukinying:
[...]

PR #16398, opened 2019-08-15
Add Experimental Flight Infrastructure

sebmarkbage:
This adds some plumbing for the Flight experiment.

We'll likely need this to (optionally) support a streaming protocol. This is the same as Fizz (streaming server rendering). There is quite a bit of wiring to make that work. This infra creates a build for each environment with zero overhead cost.

We have one build for Node which is the default. We currently have a separate build for a Browser environment (with browser stream support). This is useful for Service Workers or just debugging since the browser debugger is better than Node. We'll likely also have a FB specific one. 

ReactServerHostConfig configures the runtime environment and has various overrides.

Additionally, we likely want to special case the encoding format for some parts of the protocol depending on what you're rendering into. E.g. a host component tree can be flattened into raw HTML. Therefore, we want one entry point for each renderer type. Not a single one for the whole project. This entry point is `react-dom/unstable-flight`.

ReactServerFormatConfig configures the output format configuration.

Both of these are shared between Fizz and Flight. So they each get the host config and the format config compiled in.

The second two commits show the actually relevant files.

cc @timneutkens


devknoll:
How does colocation work with this? Isn't the mentioned example of `Foo.model.js` _by definition_ not colocated? Or are you imagining something akin to:

```
// Post.js
function Post({model}) {
  return (
    <div>
      {model.user}
      {model.text}
    </div>
  )
}
Post.getModelFromProps = async (props) => {
  const post = fetch(`/posts/${props.id}`)
  return {post, user: <User id={post.author_id}} />}
}

// User.js
function User({model}) {
  return <span>{model.name}</span>
}
User.getModelFromProps = async (props) => {
  const user = fetch(`/users/${props.id}`)
  return user;
}
```

(where e.g. `getModelFromProps` could be automatically _extracted_ into a `Post.model.js`, etc)

sebmarkbage:
Yea, collocated in the sense of two files next to each other. An entangled pair. Not in the same file. The rationale is that currently it's really hard to tell how different parts of a file might be entangled.  E.g. which imports are client-only or server-only? is this value visible to the client? Collocating in completely different parts of a file isn't much easier to keep track of. It's not as if it's in the same expression. A big part of this design is just the explainability. It's easy to tell what code is server and which is client this way.

gaearon:
Can you explain the API a bit? I’ve read the tests and still a bit confused how pieces are supposed to fit together and why host nodes are model output.

sebmarkbage:
@gaearon No. This is the infra PR, not a project review. :)

packages/react-server/src/__tests__/ReactFlight-test.js
```diff
@@ -0,0 +1,38 @@
...
+
+describe('ReactFlight', () => {
+  beforeEach(() => {
+    jest.resetModules();
+
+    React = require('react');
+    ReactNoopFlight = require('react-noop-renderer/flight');
+  });
+
+  it('can call resolve a model', () => {
```
josephsavona: nit: 'can resolve a model'

packages/react-server/src/ReactFlightStreamer.js
```diff
@@ -0,0 +1,144 @@
+/**
```
josephsavona: just curious, why put this in react-server instead of react-flight?

fixtures/flight-browser/index.html
```diff
@@ -0,0 +1,84 @@
...
+            <Text>hello</Text>
+            <Text>world</Text>
+          </div>
+        );
+      }
+
+      let model = {
+        title: 'Title',
+        content: {
+          __html: <HTML />,
```
josephsavona: Just checking that I'm following what's happening here: "model" values can render components, which are (at least for now) effectively rendered on the server and sent down as an HTML string.

packages/react-server/src/ReactFlightStreamer.js
sebmarkbage: Mostly because both Fizz and Flight depend on ReactServerHostConfig/ReactServerFormatConfig so it lets me avoid hoisting those concerns to a third package. It's quite likely that the two servers would be running in the same machine and interleaved. So you likely want to version them together too.

It gets awkward now that I'm adding a client API too which needs its own package or join `react-reconciler`.

It's not that important though since these are all not the encouraged APIs other than for custom builds. The public API is the `react-dom` entry points. We don't even own the `react-server` npm name so will likely change.

fixtures/flight-browser/index.html
sebmarkbage: Yea.  The HTML component is actually just a Model function. Then that model happens to contains "host" components (leaves) which are flattened into HTML. This will likely need to be much more clever to protect against XSS attacks like #3473 and to allow targeting nested nodes with client components etc.

sebmarkbage: The nice thing about this pattern is that if you have a tree of stateless functional components, the same code can be one shot rendered like this, or client rendered.

devknoll: > since it's just a Model function and can return arbitrary data structures

Can you clarify in your current mental model: Is a Model function actually a component (i.e. can I use hooks in it), or are you using `<HTML />` here for sugar?

sebmarkbage: A Model function can not use hooks. It's not a client-side component so it cannot have effects or state. (We could potentially let it have access to useContext though.)

devknoll: Can it suspend? If not, how can it fetch any async data?

I know this is just a plumbing PR, but it would be super helpful if you could include a small end-to-end example 😄

sebmarkbage: That’s one possible solution but that’s really not decided. It could be async functions, it could be something else. :)

fixtures/flight-browser/index.html
```diff
@@ -0,0 +1,84 @@
...
+        let charsReceived = 0;
+        let decoder = new TextDecoder();
+
+        let json = '';
+        reader.read().then(function processChunk({ done, value }) {
+          if (done) {
+            renderResult(json);
+            return;
+          }
+          json += decoder.decode(value);
```
sophiebits: I think this needs a {stream: true} in order to properly handle the case that a chunk splits a codepoint, and in the `if` above, add

```js
json += decoder.decode();
```

to finish. https://encoding.spec.whatwg.org/#example-end-of-stream

PR #16408, opened 2019-08-15
Enable MessageLoop implementation by default

gaearon:
This disables the RAF scheduler impl in favor of the MessageLoop one, which we've found better utilizes CPU. Unlike https://github.com/facebook/react/pull/16271, this keeps the RAF implementation for now just in case, but it's only in one specially marked test.

This matches what we already rolled out in www, except the rAF-specific flags (they're true on www but they're dead code in the rAF path anyway — and I didn't want to adjust the test for them).

My previous PR includes most of the extra test coverage for ML scheduler which I picked up from Andrew's earlier PR.

### DIFF
--- a/packages/scheduler/src/SchedulerFeatureFlags.js
+++ b/packages/scheduler/src/SchedulerFeatureFlags.js
@@ -10,5 +10,5 @@ export const enableSchedulerDebugging = false;
 export const enableIsInputPending = false;
 export const requestIdleCallbackBeforeFirstFrame = false;
 export const requestTimerEventBeforeFirstFrame = false;
-export const enableMessageLoopImplementation = false;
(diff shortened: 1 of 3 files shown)


sebmarkbage:
Maybe we need to make sure this works on RN now?

acdlite:
Let's go ahead and merge this?

gaearon:
Lemme check RN. What are we looking for? Just that effects fire?

gaearon:
I tested on RN a bit. useEffect works.
The fallback implementation is the same with or without the flag, so that should be fine.

brunolemos:
> that should be fine

famous last words 😄

phaistonian:
Is this change published on the latest or any of the pre-releases?

Issue #12051, opened 2018-01-19
Event Handlers on Fragments


sebmarkbage:
The use case is when an abstraction wants to listen to some event bubbling up and intercept the event somehow.

```js
function Abstraction(props) {
  return <div onClick={() => props.onAction()}>{props.children}</div>;
}
```

Perhaps there is some state and context involved too.

This works fine today as long as you can wrap it in a `<div />`. However, you can't necessarily always do that. I hear that some of this will be relaxed but conceptually I think the constraints will remain in many environment that you don't want a wrapper element around these.

```js
<ul>
  <Abstraction onAction={...}>
    <li>A</li>
    <li>B</li>
  </Abstraction>
</ul>
```

```js
<table>
  <tbody>
    <Abstraction onAction={...}>
      <tr><td>A</td></tr>
      <tr><td>B</td></tr>
    </Abstraction>
    <Abstraction onAction={...}>
      <tr><td>A</td></tr>
      <tr><td>B</td></tr>
    </Abstraction>
  </tbody>
</table>
```

It would be nice to be able to use fragments for this.

```js
function Abstraction(props) {
  return <Fragment onClick={() => props.onAction()}>{props.children}</Fragment>;
}
```

It is easy to implement with the synthetic event system but I suspect it is doable with other approaches too (including just inserting an element called `<fragment />` in the DOM).

To implement this we'd need somewhere to store the "current" props. We can use the `stateNode` on fragment fibers to hold that. We also need to schedule commit phase effects whenever the set of event handlers on a fragment changes. In the commit phase we update the stateNode on both Fibers to hold the "current props".

That way when we bubble up the fiber return pointers, we know that stateNode holds the current set of props and those will be the ones we extract event handlers from.

sophiebits:
We'd need something like `display: contents` in order to actually insert a DOM node. It's not clear that this is really feasible without the synthetic event system AFAICT.

sebmarkbage:
I believe there are plans to allow for more of these forms in browsers, and even if the normal DOM node solution doesn't work, a plausible future addition would be to have first class fragment nodes. @trueadm has been really passionate about this. So I think there's at least a plausible future where this capability just works natively.

[...]

PR #16425, opened 2019-08-16
fragment events

kyeotic:
This is a minimal working set for capturing/bubbling events on Fragements, as discussed in #12051. @bsidelinger912 helped me with this (thank you Ben!)

I'm opening this PR now to start a discussion, and to see if I am on the right track. There are still failing tests as a result of this change but I want to make sure this is something the team is still interested in before I invest more time to fix them.

I have proved this out with the included unit test, as well as with manual confirmation with a built version in the test fixture (which I didn't commit).

### DIFF
--- a/packages/react-reconciler/src/ReactChildFiber.js
+++ b/packages/react-reconciler/src/ReactChildFiber.js
@@ -517,7 +517,6 @@ function ChildReconciler(shouldTrackSideEffects) {
           return created;
         }
         case REACT_FRAGMENT_TYPE: {
-          console.log('fargment child creating');
           const created = createFiberFromFragment(
             newChild,
             returnFiber.mode,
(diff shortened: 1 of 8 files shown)


packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -516,6 +516,17 @@ function ChildReconciler(shouldTrackSideEffects) {
           created.return = returnFiber;
           return created;
         }
+        case REACT_FRAGMENT_TYPE: {
+          console.log('fargment child creating');
```
jacobp100:
```suggestion
```

kyeotic: removed

trueadm:
This isn't something we're looking to support with the existing event system, so I'm going to close this PR for now.

kyeotic:
@trueadm That's disappointing. Are there plans to support it in a future event system, or should I read that as React never intends to support this kind of abstraction?

trueadm:
@kyeotic It's something I'm actively investigating as part of our experimention with the new event system, although there's nothing planned in the short-term.

PR #16432, opened 2019-08-16
Babel Transform JSX to React.jsx/React.jsxDEV Plugin

lunaruan:
As per the [RFC to deprecate `createElement`](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md) , we want to remove `React.createElement(type, props, children)` and replace this with `React.jsx(type, props, key)`. This modifies the existing [`@babel/plugin-transform-react-jsx`](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-react-jsx/src/index.js) (and [helper](https://github.com/babel/babel/blob/master/packages/babel-helper-builder-react-jsx/src/index.js)) babel plugin to support `React.jsx` and `React.jsxDEV`. 

The main differences between `React.jsx`/`React.jsxDEV` and `React.createElement` are:
  1.) `key` is now passed as an explicit argument rather than through props
  3.) `children` are now passed through props rather than as an explicit argument
  4.) `props` must always be an object
  5.) `__source` and and `__self` are now passed as separate arguments into `React.jsxDEV` rather than through props

Part of the rationale for this change is that we to deprecate key spread through props because this is an expensive dynamic comparison operation. We want users instead always explicitly pass key as a prop. However, in the interim, we need a way to distinguish between `<div {...props} key={foo} />` and `<div key={foo} {...props} />`. Therefore, until we completely deprecate key spreading, we will use `React.createElement` to transform `<div {...props} key="Hi" />` and `React.jsx` to transform everything else.


packages/react-jsx-babel-plugin/__tests__/transform-jsx-to-react-jsx-test.js
```diff
@@ -0,0 +1,508 @@
...
+        ? [
+            '@babel/plugin-transform-react-jsx-source',
+            '@babel/plugin-transform-react-jsx-self',
+          ]
+        : []),
+      [
+        './packages/react-jsx-babel-plugin/transform-jsx-to-react-jsx',
+        {
+          useBuiltIns: true,
+          useCreateElement: false,
```
rickyvetter: Something to consider that we might also need a config for auto-imports.

packages/react/src/ReactElement.js
```diff
@@ -12,6 +12,7 @@ import {REACT_ELEMENT_TYPE} from 'shared/ReactSymbols';
 import ReactCurrentOwner from './ReactCurrentOwner';
 
 const hasOwnProperty = Object.prototype.hasOwnProperty;
+const freeze = Object.freeze;
```
rickyvetter: This is no longer wrapped in __DEV__ - does it still get compiled away in prod? Probably, but idk. Regardless I'd make sure it's a separate commit - doesn't fit with the rest of the changes.

packages/react/src/ReactElement.js
```diff
@@ -179,6 +180,16 @@ export function jsx(type, config, maybeKey) {
   let key = null;
   let ref = null;
 
+  // Currently, key can be spread in as a prop. This causes a potential
+  // issue if key is also explicitly declared (ie. <div {...props} key="Hi" />
+  // or <div key="Hi" {...props} /> ). We want to deprecate key spread,
+  // but as an intermediary step, we will use jsxDEV for everything except
+  // <div {...props} key="Hi" />, because we aren't currently able to tell if
+  // key is explicitly declared to be undefined or not.
+  if (maybeKey !== undefined) {
```
rickyvetter: I like lifting this, but I think it should be right next to the `hasValidKey` check on 197. Especially since this comment applies to both checks. Might be better to lift it as well to be above the `ref` assignment.

packages/react/src/ReactElement.js
```diff
@@ -239,6 +244,16 @@ export function jsxDEV(type, config, maybeKey, source, self) {
...
```
rickyvetter: Ditto on the lift.

packages/react/src/ReactElementValidator.js
```diff
@@ -338,16 +340,20 @@ export function jsxWithValidation(
...
+          validateChildKeys(children, type);
         }
       } else {
         validateChildKeys(children, type);
       }
     }
   }
 
-  if (props.key !== undefined) {
+  if (hasOwnProperty.call(props, 'key')) {
```
rickyvetter: 🔥

packages/react/src/ReactElementValidator.js
```diff
@@ -338,16 +340,20 @@ export function jsxWithValidation(
...
     if (children !== undefined) {
       if (isStaticChildren) {
-        for (let i = 0; i < children.length; i++) {
-          validateChildKeys(children[i], type);
+        if (Array.isArray(children)) {
+          for (let i = 0; i < children.length; i++) {
+            validateChildKeys(children[i], type);
+          }
+        } else {
+          validateChildKeys(children, type);
```
rickyvetter: Yeah good catch here.

packages/react-jsx-babel-plugin/__tests__/__snapshots__/transform-jsx-to-react-jsx-test.js.snap
```diff
@@ -0,0 +1,424 @@
...
+
+exports[`transform react to jsx display name variable declaration 1`] = `
+"var Component = React.createClass({
+  render: function render() {
+    return null;
+  }
+});"
+`;
+
+exports[`transform react to jsx fragment with no children 1`] = `"var x = React.jsx(React.Fragment, {});"`;
```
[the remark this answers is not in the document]
sebmarkbage: It wouldn't make sense if expected that to increase in scope. Fragment syntax is probably the only one that has special syntax so that makes sense.

`<></>` should auto-import.

It should not extend to things with identifiers though.

However, for example `<Fragment />` and `<Profiler />` shouldn't auto-import.

packages/react-jsx-babel-plugin/__tests__/__snapshots__/transform-jsx-to-react-jsx-test.js.snap
```diff
@@ -0,0 +1,424 @@
...
+  children: React.jsx(\\"span\\", {
+    // a double-slash comment
+    attr2: \\"bar\\"
+  })
+});"
+`;
+
+exports[`transform react to jsx should properly handle potentially null variables 1`] = `
+"var foo = null;
+var x = React.jsx(\\"div\\", Object.assign({}, foo));"
```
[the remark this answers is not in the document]
sebmarkbage: Looks like the upstream Babel 7 version switches been the built-in or Object.assign.

https://github.com/babel/babel/blob/master/packages/babel-helper-builder-react-jsx/src/index.js#L204-L206

So we should keep that aligned.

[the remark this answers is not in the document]
lunaruan: Yes, there is a `useBuiltIns` flag that can be set to use `Object.assign` or `extends` (Babel's version of object assignment). I'll add another test to make sure that the flag works though.

packages/react-jsx-babel-plugin/__tests__/__snapshots__/transform-jsx-to-react-jsx-test.js.snap
```diff
@@ -0,0 +1,424 @@
...
+
+exports[`transform react to jsx concatenates adjacent string literals 1`] = `
+"var x = React.jsx(\\"div\\", {
+  children: [\\"foo\\", \\"bar\\", \\"baz\\", React.jsx(\\"div\\", {
+    children: \\"buz bang\\"
+  }), \\"qux\\", null, \\"quack\\"]
+});"
+`;
+
+exports[`transform react to jsx display name assignment expression 1`] = `
```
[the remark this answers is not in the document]
lunaruan: Good catch! I copied a lot of the tests from [Babel](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx/test/fixtures/react). Previously, for `React.createClass`, we needed to add a `displayName` for debugging to distinguish between different components without a name, but now we don't need this anymore. Will remove these tests.

packages/react-jsx-babel-plugin/transform-jsx-to-react-jsx.js
```diff
@@ -0,0 +1,564 @@
...
+    // for React.jsx, key, __source (dev), and __self (dev) is passed in as
+    //a separate argument rather than in the args object. We go through the
+    // props and filter out these three keywords so we can pass them in
+    // as separate arguments later
+    for (let i = 0, len = openingPath.node.attributes.length; i < len; i++) {
+      const attr = openingPath.node.attributes[i];
+      if (t.isJSXAttribute(attr) && t.isJSXIdentifier(attr.name)) {
+        if (attr.name.name === 'key') {
+          key = convertAttribute(attr).value;
+        } else if (attr.name.name === '__source') {
```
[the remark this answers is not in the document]
lunaruan: Good question! There's a separate transform that adds [`__source`](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx-source) and [`__self`](https://babeljs.io/docs/en/babel-plugin-transform-react-jsx-self) as props. Usually, these plugins are used automatically in [`@babel/preset-react`](https://github.com/babel/babel/blob/11ed2e2bf572c5748b1bef47d748da97f7e0bedb/packages/babel-preset-react/src/index.js#L31)

packages/react-jsx-babel-plugin/__tests__/__snapshots__/transform-jsx-to-react-jsx-test.js.snap
```diff
@@ -0,0 +1,424 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`transform react to jsx React.fragment to set keys and source 1`] = `
+"var _jsxFileName = \\"\\";
+var x = React.jsxDEV(React.Fragment, {}, \\"foo\\", true, {
```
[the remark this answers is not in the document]
lunaruan: Yes, planning on adding that soon! I wanted to incrementally land these changes first though. LMK what you think.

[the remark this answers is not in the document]
rickyvetter: I'm fine with incrementally adding these features for sure. If that's the plan then I think this PR looks good as it stands.

packages/react-jsx-babel-plugin/__tests__/__snapshots__/transform-jsx-to-react-jsx-test.js.snap
```diff
@@ -0,0 +1,374 @@
...
+  lineNumber: 2
+}, this);"
+`;
+
+exports[`transform react to jsx properly handles keys 1`] = `
+"var x = React.jsx(\\"div\\", {
+  children: [React.jsx(\\"div\\", {}, \\"1\\"), React.jsx(\\"div\\", {
+    meow: \\"wolf\\"
+  }, \\"2\\"), React.jsx(\\"div\\", {}, \\"3\\")]
+});"
```
rickyvetter: Sorry I missed this - static callers should actually use [`jsxs`, not `jsx`](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#always-pass-children-as-props).

```js
var x = React.jsxs("div", {
  children: [React.jsx("div", {}, "1"), React.jsx("div", {
    meow: "wolf"
  }, "2"), React.jsx("div", {}, "3")]
});
```

packages/react/src/ReactElementValidator.js
rickyvetter: On second read-through this `jsxs` is why it doesn't check here - `jsxs` should only be used if children is statically known to be an array. It doesn't make sense otherwise. I think we should throw some other warning when this happens (stop using jsxs in userland unless you know what you're doing!)

lunaruan: I think the issue with that is that in the case of something like:
```
function foo() {
   const x = [<span />, <span />];
   return (
      <div>{x}</div>
   );
}
```
Theoretically, when you call this function, `div`'s children are statically known to be an array. However, babel doesn't know this, because Babel doesn't know what `x` is. Thus, we have to declare this as static. 

However, in this case:
```
function foo() {
   const x = <span />;
   return (
      <div>{x}</div>
   );
}
```
If `x` is actually a singleton, it would not be an array. Thus the extra check.

sebmarkbage: This doesn't seem right. isStaticChildren should only be true when the transform itself creates the array, i.e. not the above example. So it would always be an array.

packages/react-jsx-babel-plugin/transform-jsx-to-react-jsx.js
```diff
@@ -0,0 +1,564 @@
...
+  visitor.Program = {
+    enter(path, state) {
+      state.set(
+        'oldJSXIdentifier',
+        createIdentifierParser('React.createElement'),
+      );
+      state.set(
+        'jsxIdentifier',
+        createIdentifierParser(
+          state.opts.development ? 'React.jsxDEV' : 'React.jsx',
```
sebmarkbage: The transform currently never transforms to `React.jsxs`. `jsxs` doesn't have any runtime semantic differences in prod so it shouldn't matter. However, we should still do it in the prod transform because not everyone has separate development and production builds/transforms. So that lets us still keep the development warnings in such environment.

We should switch between `React.jsx` and `React.jsxs` based on if the children are static. I.e. same as `isStaticChildren`.

packages/react/src/ReactElementValidator.js
lunaruan: Ah yes, I got confused about what `isStaticChildren` referred to. Thanks for the clarification! Updated it and also added a warning as per @rickyvetter's suggestion

packages/react-jsx-babel-plugin/__tests__/transform-jsx-to-react-createElement-test.js
```diff
@@ -0,0 +1,355 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+/* eslint-disable quotes */
+'use strict';
+
+const babel = require('@babel/core');
+const codeFrame = require('@babel/code-frame');
+
+function transform(input, options) {
+  return babel.transform(input, {
```
gaearon: Check out how `FreshBabelPlugin` test does `let {wrap} = require('jest-snapshot-serializer-raw');`. This will make snapshots a lot more readable.

packages/react-jsx-babel-plugin/__tests__/transform-jsx-to-react-createElement-test.js
```diff
@@ -0,0 +1,355 @@
...
+          '\n' +
+          codeFrame.codeFrameColumns(
+            code,
+            {start: {line: 1, column: 2}},
+            {highlightCode: true}
+          )
+      )
+    );
+  });
+  it('should transform known hyphenated tags', () => {
```
gaearon: Nit: would you mind leaving newlines between tests

packages/react-jsx-babel-plugin/__tests__/transform-jsx-to-react-jsx-test.js
```diff
@@ -0,0 +1,436 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+/* eslint-disable quotes */
+'use strict';
+
+const babel = require('@babel/core');
+const codeFrame = require('@babel/code-frame');
+
+function transform(input, options) {
+  return babel.transform(input, {
```
gaearon: Same re: wrap(). Might make them easier to review.

packages/react-jsx-babel-plugin/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "react-jsx-babel-plugin",
+  "version": "1.0.0",
```
gaearon: Let's start with 0.1.0 to signal this isn't a thing people should use yet? Also we need to decide whether to mark it as `private` until we want others to try it.

`babel-plugin` shouldn't be the suffix. It should be a prefix: https://babeljs.io/docs/en/plugins/#plugin-shorthand. But we'll probably want to do it in upstream Babel plugin anyway — this is only for testing, right? In that case it should be `private`.

packages/react/src/ReactElementValidator.js
```diff
@@ -334,20 +336,30 @@ export function jsxWithValidation(
...
+        if (Array.isArray(children)) {
+          for (let i = 0; i < children.length; i++) {
+            validateChildKeys(children[i], type);
+          }
+        } else {
+          warning(
+            false,
+            'React.jsx: Static children should always be an array. ' +
+              'You are likely explicitly calling React.jsxs or React.jsxDEV. ' +
+              'Use the babel transform instead.',
```
gaearon: Nit: uppercase Babel

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
+'use strict';
+
+const esutils = require('esutils');
+
+function helper(babel, opts) {
+  const {types: t} = babel;
+
+  const visitor = {};
+
+  visitor.JSXNamespacedName = function(path, state) {
```
Jessidhia: Why write it this way, I wonder? 🤔

```jsx
const visitor = {
  JSXNamespacedName(path, state) {
    // ...
  },
  JSXSpreadChild(path) {
    // ...
  }
  // ...
}
```

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+  visitor.JSXSpreadChild = function(path) {
+    throw path.buildCodeFrameError(
+      'Spread children are not supported in React.',
+    );
+  };
+
+  visitor.JSXElement = {
+    exit(path, file) {
+      let callExpr;
+      if (file.opts.useCreateElement || useCreateElement(path)) {
```
Jessidhia: Starting a non-hook function name with `use` sounds like bad karma.

`shouldUseCreateElement`?

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+
+module.exports = function(babel) {
+  const {types: t} = babel;
+
+  const createIdentifierParser = id => () => {
+    return id
+      .split('.')
+      .map(name => t.identifier(name))
+      .reduce((object, property) => t.memberExpression(object, property));
+  };
```
Jessidhia: This function name is confusing. Looking at the name, I thought it did the same as https://github.com/babel/babel/blob/master/packages/babel-types/src/validators/buildMatchMemberExpression.js.

Jessidhia:
```ts
const { template } = babel

template.ast(id).expression
```

This does parse arbitrary code, though, but you only call it with predefined inputs.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+        createIdentifierParser(
+          state.opts.development ? 'React.jsxDEV' : 'React.jsxs',
+        ),
+      );
+      state.set('jsxFragIdentifier', createIdentifierParser('React.Fragment'));
+    },
+  };
+
+  visitor.JSXAttribute = function(path) {
+    if (t.isJSXElement(path.node.value)) {
```
Jessidhia: Ah, the weirdest part of the JSX spec nobody noticed nobody implemented.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+      attribs = objs[0];
+    } else {
+      // looks like we have multiple objects
+      if (!t.isObjectExpression(objs[0])) {
+        objs.unshift(t.objectExpression([]));
+      }
+
+      const expressionHelper = useBuiltIns
+        ? t.memberExpression(t.identifier('Object'), t.identifier('assign'))
+        : file.addHelper('extends');
```
Jessidhia: Possibly unpopular opinion: just write object spreads. Tell users to add `@babel/transform-object-rest-spread` if they need it.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
Jessidhia: Ah, I see, this is a fork of https://github.com/babel/babel/blob/master/packages/babel-helper-builder-react-jsx/src/index.js

packages/react/src/ReactElementValidator.js
Jessidhia: ~~No love for iterables?~~

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+    } else {
+      node.name = t.stringLiteral(node.name.name);
+    }
+
+    return t.inherits(t.objectProperty(node.name, value), node);
+  }
+
+  // We want to use React.createElement, even in the case of
+  // jsx, for <div {...props} key={key} /> to distinguish it
+  // from <div key={key} {...props} />. This is an intermediary
```
Jessidhia: I'd expect `<div key={key} {...props} />` to be the problematic one, as the `...props` spread could potentially override the value of `key`.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+
+  // Builds JSX into:
+  // Production: React.jsx(type, arguments, key)
+  // Development: React.jsxDEV(type, arguments, key, isStaticChildren, source, self)
+  function buildJSXElementCall(path, file) {
+    if (opts.filter && !opts.filter(path.node, file)) {
+      return;
+    }
+
+    const openingPath = path.get('openingElement');
```
jridgewell: Ditto things from `buildCreateElementCall`

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+    const useBuiltIns = file.opts.useBuiltIns || false;
+    if (typeof useBuiltIns !== 'boolean') {
+      throw new Error(
+        'transform-react-jsx currently only accepts a boolean option for ' +
+          'useBuiltIns (defaults to false)',
+      );
+    }
+
+    while (attribs.length) {
+      const prop = attribs.shift();
```
jridgewell: This is an n<sup>2</sup> algorithm. Why not a normal for loop?

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+    let attribs = [];
+    let key;
+    let source;
+    let self;
+
+    // for React.jsx, key, __source (dev), and __self (dev) is passed in as
+    // a separate argument rather than in the args object. We go through the
+    // props and filter out these three keywords so we can pass them in
+    // as separate arguments later
+    for (let i = 0, len = openingPath.node.attributes.length; i < len; i++) {
```
jridgewell: Nit: store `openingPath.node.attributes` in a variable. Also, `length` check.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+      const attr = openingPath.node.attributes[i];
+      if (t.isJSXAttribute(attr) && t.isJSXIdentifier(attr.name)) {
+        if (attr.name.name === 'key') {
+          key = convertAttribute(attr).value;
```
jridgewell: Is `<div key={foo} />` supported? How about `<div key />`?

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+      t.callExpression(
+        path.node.children.length > 1 ? state.staticCallee : state.callee,
+        args,
+      )
+    );
+  }
+
+  // Builds props for React.jsx. This function adds children into the props
+  // and ensures that props is always an object
+  function buildJSXOpeningElementAttributes(attribs, file, children) {
```
jridgewell: Ditto things from `buildCreateElementOpeningElementAttributes`.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+        );
+      }
+    }
+
+    pushProps(_props, objs);
+
+    if (objs.length === 1) {
+      // only one object
+      if (!t.isObjectExpression(objs[0])) {
+        // this could be null, and jsx expects props to be non-null
```
jridgewell: I think the comment is incorrect. It's either an `ObjectExpression` or the argument of a `JSXSpreadAttribute`.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+      } else {
+        attribs = objs[0];
+      }
+    } else {
+      // looks like we have multiple objects
+      if (!t.isObjectExpression(objs[0])) {
+        objs.unshift(t.objectExpression([]));
+      }
+
+      const expressionHelper = useBuiltIns
```
jridgewell: This could be hoisted out of both branches.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+  // Builds JSX into:
+  // Production: React.createElement(type, arguments, children)
+  // Development: React.createElement(type, arguments, children, source, self)
+  function buildCreateElementCall(path, file) {
+    if (opts.filter && !opts.filter(path.node, file)) {
+      return;
+    }
+
+    const openingPath = path.get('openingElement');
+    openingPath.parent.children = t.react.buildChildren(openingPath.parent);
```
jridgewell: It seems like we could just store this as a variable, instead of mutating the parent node.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+      const args = state.args;
+      if (t.react.isCompatTag(tagName)) {
+        args.push(t.stringLiteral(tagName));
+      } else {
+        args.push(state.tagExpr);
+      }
+    },
+
+    post(state, pass) {
+      state.callee = pass.get('jsxIdentifier')();
```
jridgewell: I don't understand the deferred pattern here, it's not expensive to create the member expressions.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+        : file.addHelper('extends');
+
+      // spread it
+      attribs = t.callExpression(expressionHelper, objs);
+    }
+
+    return attribs;
+  }
+
+  function buildCreateElementFragmentCall(path, file) {
```
jridgewell: Ditto things from `buildCreateElementCall`

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -0,0 +1,586 @@
...
+      value.value = value.value.replace(/\n\s+/g, ' ');
+
+      // "raw" JSXText should not be used from a StringLiteral because it needs to be escaped.
+      if (value.extra && value.extra.raw) {
+        delete value.extra.raw;
+      }
+    }
+
+    if (t.isJSXNamespacedName(node.name)) {
+      node.name = t.stringLiteral(
```
jridgewell: Maybe needs to inherit from `node.name`?

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
lunaruan: Thanks for the review! We want to make sure there are no behavior changes for users even though we are changing the transform. In the `createElement` case,`...props` does override the key, so this is okay. In the `jsx` case, `<div {...props} key={key} />` would transform into `jsx(div, props, undefined)`. However, we would not be certain whether key is `undefined` because it was not passed down as a prop or because it was explicitly called as `undefined`, so we would not be able to replicate the behavior of the potential override. Thus, `<div {...props} key={key} />` would be the problematic one.

lunaruan:
@jridgewell Thanks for the code review! Most of this code is forked from [@babel/plugin-transform-react-jsx](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-react-jsx/src/index.js) and associated [helper](https://github.com/babel/babel/blob/master/packages/babel-helper-builder-react-jsx/src/index.js).  This code is for experimentation purposes only. We plan to upstream these changes to Babel at some point in the future once we've proven out the concept, and we'll make changes to the original transform if need be.

packages/react-jsx-babel-plugin/package.json
lunaruan: Yes, this is only for testing purposes. When we publish the changes it will be to the actual Babel plugin. I'll mark this test version as private.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
lunaruan: A lot of this code was forked from the existing react jsx babel transform. If you'd like, you can put up a PR to modify that transform with your proposed changes.

packages/react-jsx-babel-plugin/src/TransformJSXToReactBabelPlugin.js
lunaruan: In the case of:
```
const props = null;
<div {...props} />
```
this would translate into `jsx(div, props, ...)`, which would equal `jsx(div, null, ...)`. Therefore, we have to transform the above to `jsx(div, Object.assign({}, props),...)` to ensure that the second argument is never null. You make a good point though that this comment is slightly confusing.

jridgewell:
👍 . If the intention is to upstream these changes, we can wait to do a full review until then.

PR #16433, opened 2019-08-16
[react-events] DOM event testing library

necolas:
This patch formalizes the mock native events and event sequences used in unit tests.

The `createEventTarget` function returns an object that can be used to dispatch native event sequences on the target without having to manually do so across all the scenarios we need to account for. Unit tests can be written as if we were only working with PointerEvent, but they will dispatch realistic native event sequences based on the execution environment (e.g., is PointerEvent supported?) and pointer type.

```
describe.each(environments)('Suite', (hasPointerEvents) => {
  beforeEach(() => {
    // setup
  });

  test.each(pointerTypes)('Test', (pointerType) => {
    const target = createEventTarget(node);
    target.pointerdown({pointerType});
    expect(callback).toBeCalled();
  });
});
```

Every native event that is dispatched now includes a complete object by default. The properties of the events can be customized. Properties that shouldn't be relied on in responder implementations are excluded from the mock native events to ensure tests will fail. Equivalent properties are normalized across different event types, e.g., 'pointerId' is converted to 'identifier' before a TouchEvent is dispatched.


packages/react-events/src/dom/__tests__/Press-test.internal.js
```diff
@@ -140,51 +127,54 @@ describe.each(environmentTable)('Press responder', hasPointerEvents => {
       );
     });
 
-    it('is not called after "pointermove" following auxillary-button press', () => {
-      const target = ref.current;
-      target.getBoundingClientRect = () => ({
+    it('is not called after pointer move following auxillary-button press', () => {
+      const node = ref.current;
+      const target = createEventTarget(node);
+      node.getBoundingClientRect = () => ({
```
necolas: Use `setBoundingClientRect`

packages/react-events/src/dom/__tests__/Press-test.internal.js
```diff
@@ -475,20 +463,21 @@ describe.each(environmentTable)('Press responder', hasPointerEvents => {
     it.each(pointerTypesTable)(
       'is called after pointer move: %s',
       pointerType => {
-        const target = ref.current;
-        target.getBoundingClientRect = () => ({
+        const node = ref.current;
+        node.getBoundingClientRect = () => ({
```
necolas: use target helper

packages/react-events/src/dom/testing-library/domEventSequences.js
```diff
@@ -0,0 +1,201 @@
...
+    if (hasPointerEvent()) {
+      dispatch(domEvents.pointerup(payload));
+      dispatch(domEvents.lostpointercapture(payload));
+      dispatch(domEvents.pointerout(payload));
+      dispatch(domEvents.pointerleave(payload));
+    }
+    dispatch(
+      domEvents.touchend(
+        payload,
+        payload != null ? payload.pointerId : undefined,
```
necolas: Remove

packages/react-events/src/dom/testing-library/domEventSequences.js
```diff
@@ -0,0 +1,201 @@
...
+  } else {
+    if (hasPointerEvent()) {
+      dispatch(domEvents.pointerover(payload));
+      dispatch(domEvents.pointerenter(payload));
+      dispatch(domEvents.pointerdown(payload));
+    }
+    dispatch(
+      domEvents.touchstart(
+        payload,
+        payload != null ? payload.pointerId : undefined,
```
necolas: Remove

packages/react-events/src/dom/testing-library/domEventSequences.js
```diff
@@ -0,0 +1,201 @@
...
+  if (hasPointerEvent()) {
+    dispatch(domEvents.pointermove(payload));
+  }
+  if (pointerType === 'mouse') {
+    dispatch(domEvents.mousemove(payload));
+  } else {
+    dispatch(
+      domEvents.touchmove(
+        payload,
+        payload != null ? payload.pointerId : undefined,
```
necolas: Remove

packages/react-events/src/dom/testing-library/index.js
```diff
@@ -0,0 +1,109 @@
...
+        left: x,
+        right: x + width,
+        top: y,
+        bottom: y + height,
+      };
+    };
+  },
+});
+
+module.exports = {
```
trueadm: Can't we just use ES module exports here? It's only going to be used in Jest, but it makes things more consistent with the other files in `src`.

necolas: Sure. Spent too much time in www :P

PR #16434, opened 2019-08-17
[Partial Hydration] Add "hydrationOptions" behind the enableSuspenseCallback flag

sebmarkbage:
This gets invoked when a boundary is either hydrated or if it is deleted because its props/context updated before getting hydrated or a parent got deleted before it mounted.

If there are nested boundaries in the tree that we haven't gotten to yet and the parent gets deleted, then no callback is fired. Only the top one.

If we have a hydration error that causes us to delete the parent node, then there also isn't a callback. That could however be a different callback we could add.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -607,7 +607,12 @@ function commitLifeCycles(
       }
       return;
     }
-    case SuspenseComponent:
+    case SuspenseComponent: {
+      if (enableSuspenseCallback) {
+        commitSuspenseHydrationCallbacks(finishedRoot, finishedWork);
+      }
+      return;
+    }
     case SuspenseListComponent:
     case IncompleteClassComponent:
     case FundamentalComponent:
@@ -644,7 +649,8 @@ function hideOrUnhideAllChildren(finishedWork, isHidden) {
         }
       } else if (
         node.tag === SuspenseComponent &&
-        node.memoizedState !== null
+        node.memoizedState !== null &&
+        node.memoizedState.dehydrated === null
       ) {
         // Found a nested Suspense component that timed out. Skip over the
         // primary child fragment, which should remain hidden.
@@ -719,6 +725,7 @@ function commitDetachRef(current: Fiber) {
 // deletion, so don't let them throw. Host-originating errors should
 // interrupt deletion, so it's okay
 function commitUnmount(
+  finishedRoot: FiberRoot,
   current: Fiber,
   renderPriorityLevel: ReactPriorityLevel,
 ): void {
@@ -801,7 +808,7 @@ function commitUnmount(
       // We are also not using this parent because
       // the portal will get pushed immediately.
       if (supportsMutation) {
-        unmountHostComponents(current, renderPriorityLevel);
+        unmountHostComponents(finishedRoot, current, renderPriorityLevel);
       } else if (supportsPersistence) {
         emptyPortalContainer(current);
       }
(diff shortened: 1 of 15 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -644,7 +649,8 @@ function hideOrUnhideAllChildren(finishedWork, isHidden) {
         }
       } else if (
         node.tag === SuspenseComponent &&
-        node.memoizedState !== null
+        node.memoizedState !== null &&
+        node.memoizedState.dehydrated == null
```
gaearon: Intentional sloppy?

sebmarkbage: No, good catch.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -644,7 +649,8 @@ function hideOrUnhideAllChildren(finishedWork, isHidden) {
         }
       } else if (
         node.tag === SuspenseComponent &&
-        node.memoizedState !== null
+        node.memoizedState !== null &&
+        node.memoizedState.dehydrated === null
```
gaearon: What is this about? I don't see you setting it in this PR.

Is this a behavior change? Wasn't the feature logging-only? Or was this always `null` anyway?

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -815,11 +822,24 @@ function commitUnmount(
...
         }
       }
+      return;
+    }
+    case DehydratedFragment: {
+      if (enableSuspenseCallback) {
+        const hydrationCallbacks = finishedRoot.hydrationCallbacks;
+        if (hydrationCallbacks !== null) {
+          const onDeleted = hydrationCallbacks.onDeleted;
+          if (onDeleted) {
```
gaearon: `!== null`?

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -859,6 +859,10 @@ function completeWork(
             if ((workInProgress.effectTag & DidCapture) === NoEffect) {
               // This boundary did not suspend so it's now hydrated and unsuspended.
               workInProgress.memoizedState = null;
+              if (enableSuspenseCallback) {
+                // Notify the callback.
+                workInProgress.effectTag |= Update;
```
gaearon: Seems a bit worrying to have a generic effect tag like this behind a flag since this may significantly fork execution path wherever we check for `Update`.

packages/react-reconciler/src/ReactFiberRoot.js
```diff
@@ -91,6 +100,7 @@ type ProfilingOnlyFiberRootProperties = {|
 export type FiberRoot = {
```
gaearon: While we're here, should this have been exact?

packages/react-reconciler/src/ReactFiberCommitWork.js
sebmarkbage: This probably should've been in #16346 but it was not observable because we weren't scheduling an Update effect in this scenario. Now that we do, we need to differentiate what purpose the Update effect has.

packages/react-reconciler/src/ReactFiberCommitWork.js
sebmarkbage: This is an optional property. In some other cases we check `typeof ... === 'function'` when it's a prop, but in some cases we check truthiness. Not sure which one to be consistent with so I went with the one that takes fewer bytes.

packages/react-reconciler/src/ReactFiberCompleteWork.js
sebmarkbage: Yea, that's a fair concern. I'm not very confident that this will be a lasting API form though so wouldn't want to do a big refactor yet nor should be pay the cost if it's not needed.

The event replaying might need this flag too and that might move out to a shared path sooner.

packages/react-reconciler/src/ReactFiberRoot.js
sebmarkbage: I don't think Flow supported the spreading as exact but arguably we shouldn't do the spreading neither. We don't for Fiber.

PR #16442, opened 2019-08-18
Remove note about future react scripts

amiralies:
react-scripts now includes this plugin


trueadm:
@gaearon does look right to you?

PR #16453, opened 2019-08-19
Don't trigger fallbacks in act()

threepointone:
We have some code that triggers suspense fallbacks when in an act() scope. This has caused a few annoying problems.
- We don't want this codepath in prod because it's testing specific.
- It's also on a 'hot' path, so we want to avoid code that we know will never be triggered
- At this point, we have divergent behaviour between dev and prod, which sucks. it also probably means act() can't be used for prod tests (and it's not intuitive why)

This PR simply removes that feature. The longer I stare at it, I'm sure it should either not be a feature of act(), or we should bite the bullet and let it exist on the prod path. Fallbacks can be triggered in tests with act(jest.runAllTimers) if needed (and I assume someone will make a helper)


threepointone:
Investigating other solutions, closing this PR

PR #16455, opened 2019-08-19
[ESLint] Forbid top-level use*() calls

gaearon:
In the initial release of `eslint-plugin-react-hooks/rules-of-hooks`, we forbid this:

```js
function foo() {
  useState(); // BAD: calling a Hook from a non-Hook
}
```

But we didn't forbid this:

```js
// top-level code

useState(); // ALSO BAD: calling a Hook from top level
```

This PR forbids it in the lint rule.

### Why did it work before?

We didn't ban it at the time for three reasons:

1. We didn't know how Hooks would be received by the community. Taking the whole `use*` prefix seemed dicey and we didn't want to push the convention harder than absolutely necessary.
2. It is a runtime crash anyway, so you find out about this immediately.
3. There's a `history` library whose 2.x versions had an API like this:

```js
const {createHistory, useBasename} = require('history-2.1.2');
const browserHistory = useBasename(createHistory)({
  basename: '/',
});
```

So banning this pattern would create a false positive for it.

### Why ban it now?

If your environment uses "inline requires" (opt-in on React Native), you might not get a crash at all. Instead, **the top level initialization of a module may run as a result of some component's render**, and so Hooks would accidentally "belong" to the parent. That's confusing. While "inline requires" aren't very common on the web, they're a powerful optimization, so it's plausible they will get used more often with time as build systems add support for them.

Even regardless of that risk, we can confidently say Hooks have been a successful rollout now. So whether the bugs themselves are a problem or not, **the `use` convention effectively already strongly implies it's a Hook**. From that perspective, `use*()` at top level is just confusing to look at.

If you run into a false positive, you might want to change that to something like this anyway:

```js
const {createHistory, useBasename: withBasename} = require('history-2.1.2');
const browserHistory = withBasename(createHistory)({
  basename: '/',
});
```
so that your coworkers don't think it's a Hook.

That particular API also doesn't exist in modern versions of `history` anyway. And I haven't found similar APIs in the wild.

### What about `MyLibrary.useFoo()`?

We warn about invalid use of `use*()` and `React.use*()`, but not `MyLibrary.use*()`. This is because initially we saw too many false positives from modules like `MyStore.useNewAnalytics()`.

The new top-level error respects the same rule. For example, it would warn for `useBasename()`, but not for `History.useBasename()`. So that's another escape hatch if you want to suppress it.

In the future, we might consider making it stricter, and banning invalid calls to `MyLibrary.use*()` too. But it's less common in general because the convention is to import Hooks directly without a namespace. It's also much more commonly encountered in open source, like `app.use()` from Express, or `jest.useFakeTimers()`. So maybe it's not worth it.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
@@ -232,21 +232,20 @@ const tests = {
       (class {i() { useState(); }});
     `,
     `
-      // Currently valid although we *could* consider these invalid.
-      // It doesn't make a lot of difference because it would crash early.
+      // Valid because they're not matching use[A-Z].
+      fooState();
       use();
       _use();
-      useState();
       _useState();
-      use42();
-      useHook();
       use_hook();
-      React.useState();
     `,
     `
-      // Regression test for the popular "history" library
-      const {createHistory, useBasename} = require('history-2.1.2');
-      const browserHistory = useBasename(createHistory)({
+      // This is grey area.
+      // Currently it's valid (although React.useCallback would fail here).
+      // We could also get stricter and disallow it, just like we did
+      // with non-namespace use*() top-level calls.
+      const History = require('history-2.1.2');
+      const browserHistory = History.useBasename(History.createHistory)({
         basename: '/',
       });
     `,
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
```diff
@@ -708,3 +771,42 @@ function genericError(hook) {
...
+      return only.indexOf(t) !== -1;
+    }
+    if (skipped.length > 0) {
+      return skipped.indexOf(t) === -1;
+    }
+    return true;
+  };
+  tests.valid = tests.valid.filter(predicate);
+  tests.invalid = tests.invalid.filter(predicate);
+}
```
bvaughn: Nice

gaearon:
Note this is a breaking change for the plugin since it would become more aggressive.

PR #16458, opened 2019-08-19
chore(ReactDomComponent): refactor code

jamesgeorge007:
Refactor code to a more concise version.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -264,6 +264,14 @@ if (__DEV__) {
   };
 }
 
+function enableFlareAPIIsSet(propKey) {
+  return (
+    (enableFlareAPI && propKey === LISTENERS) ||
+    propKey === SUPPRESS_CONTENT_EDITABLE_WARNING ||
+    propKey === SUPPRESS_HYDRATION_WARNING
+  );
+}
+
 function ensureListeningTo(
   rootContainerElement: Element | Node,
   registrationName: string,
@@ -340,11 +348,7 @@ function setInitialDOMProperties(
       } else if (typeof nextProp === 'number') {
         setTextContent(domElement, '' + nextProp);
       }
-    } else if (
-      (enableFlareAPI && propKey === LISTENERS) ||
-      propKey === SUPPRESS_CONTENT_EDITABLE_WARNING ||
-      propKey === SUPPRESS_HYDRATION_WARNING
-    ) {
+    } else if (enableFlareAPIIsSet(propKey)) {
       // Noop
     } else if (propKey === AUTOFOCUS) {
       // We polyfill it separately on the client during commit.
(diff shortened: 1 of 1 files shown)


packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -340,11 +346,7 @@ function setInitialDOMProperties(
       } else if (typeof nextProp === 'number') {
         setTextContent(domElement, '' + nextProp);
       }
-    } else if (
-      (enableFlareAPI && propKey === LISTENERS) ||
-      propKey === SUPPRESS_CONTENT_EDITABLE_WARNING ||
-      propKey === SUPPRESS_HYDRATION_WARNING
-    ) {
+    } else if (enableFlareAPIIsSet(propKey)) {
```
trueadm: These aren't all Flare APIs, so the naming is off.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1045,14 +1039,10 @@ export function diffHydratedProperties(
...
-        (enableFlareAPI && propKey === LISTENERS) ||
-        propKey === SUPPRESS_CONTENT_EDITABLE_WARNING ||
-        propKey === SUPPRESS_HYDRATION_WARNING ||
+        enableFlareAPIIsSet(propKey) ||
         // Controlled attributes are not validated
         // TODO: Only ignore them on controlled tags.
-        propKey === 'value' ||
-        propKey === 'checked' ||
-        propKey === 'selected'
+        ['value', 'checked', 'selected'].includes(propKey)
```
trueadm: This is far less efficient than using simple or conditions.

jamesgeorge007: should I revert?

packages/react-dom/src/client/ReactDOMComponent.js
jamesgeorge007: would you be kind to suggest a name for the respective variable?

jamesgeorge007: @trueadm ?

jamesgeorge007:
@trueadm could you give me a better picture regarding the CI failure :thinking:

trueadm:
Sure thing. I’m just home now so I’ll have a look and add more feedback tomorrow when I’m in the office :)

packages/react-dom/src/client/ReactDOMComponent.js
trueadm: Yep

packages/react-dom/src/client/ReactDOMComponent.js
trueadm: `isInternalPropDefinition`?

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -697,11 +701,7 @@ export function diffProperties(
       }
     } else if (propKey === DANGEROUSLY_SET_INNER_HTML || propKey === CHILDREN) {
       // Noop. This is handled by the clear text mechanism.
-    } else if (
-      (enableFlareAPI && propKey === LISTENERS) ||
-      propKey === SUPPRESS_CONTENT_EDITABLE_WARNING ||
-      propKey === SUPPRESS_HYDRATION_WARNING
-    ) {
+    } else if (enableFlareAPI(propKey)) {
```
trueadm: Typo here, you meant to use the function rather than `enableFlareAPI`

gaearon:
Thank you for the effort, but let's not do this.

We generally avoid purely stylistic refactors. They often introduce subtle bugs when some branch was a bit different but no one noticed.

The repetition here makes the code flexible. It's easy to move things around, add or remove or negate some conditions. As soon as they get abstracted into helpers, the code becomes harder to change. You may start having some conditions that never fire, but you might not realize that because they're behind an indirection.

This also starts to mess with dead code elimination. It's harder to predict which code gets included and which doesn't.

jamesgeorge007:
@trueadm @gaearon could you point me an appropriate one to work upon?

PR #16459, opened 2019-08-19
chore(ReactDomComponent): refactor code to avoid ambiguity

jamesgeorge007:
Minor refactor to avoid ambiguities.

### DIFF
--- a/packages/react-dom/src/shared/ReactControlledValuePropTypes.js
+++ b/packages/react-dom/src/shared/ReactControlledValuePropTypes.js
@@ -28,16 +28,20 @@ if (__DEV__) {
     submit: true,
   };
 
+  function propsAreConfigured(props, propName) {
+    return (
+      hasReadOnlyValue[props.type] ||
+      props.onChange ||
+      props.readOnly ||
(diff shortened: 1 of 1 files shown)


packages/react-dom/src/shared/ReactControlledValuePropTypes.js
```diff
@@ -15,6 +15,13 @@ const ReactControlledValuePropTypes = {
   checkPropTypes: null,
 };
 
+const propsAreConfigured = hasReadOnlyValue[props.type] ||
```
trueadm: This should be in the __DEV__ block, otherwise it won't be dead code removed for prod.

gaearon:
Thanks, but we'd rather not do this. Same reasons as https://github.com/facebook/react/pull/16458#issuecomment-522718684.

In fact this already introduces a bug because those snippets aren't identical. Luckily we had test coverage for those, but I hope this shows how easy it is to make a mistake in a "refactor" like this.

There was no ambiguity in the original code. There was some duplication, but we don't mind it. Especially in older code that has worked fine for years.

PR #16460, opened 2019-08-19
Throw on unhandled SSR suspending

gaearon:
This adds a throw for suspending outside of `Suspense`, equivalent to what we do on the client side.

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRenderer.js
+++ b/packages/react-dom/src/server/ReactPartialRenderer.js
@@ -883,7 +883,8 @@ class ReactDOMServerRenderer {
               const fallbackFrame = frame.fallbackFrame;
               invariant(
                 fallbackFrame,
-                'suspense fallback not found, something is broken',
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -910,6 +910,14 @@ class ReactDOMServerRenderer {
...
         } catch (err) {
           if (enableSuspenseServerRenderer && typeof err.then === 'function') {
+            invariant(
+              this.suspenseDepth > 0,
+              // TODO: include component name. This is a bit tricky with current factoring.
+              'A React component suspended while rendering, but no fallback UI was specified.\n' +
+                '\n' +
+                'Add a <Suspense fallback=...> component higher in the tree to ' +
+                'provide a loading indicator or placeholder to display.',
+            );
```
sebmarkbage: Shouldn't we also throw if we don't have `enableSuspenseServerRenderer` on?

PR #16479, opened 2019-08-19
[react-events] Rely on 'buttons' rather than 'button'

necolas:
The semantics of 'button' on events differs between PointerEvent and
MouseEvent, whereas they are the same for 'buttons'. Furthermore, 'buttons'
allows developers to determine when multiple buttons are pressed as the same
time.

https://w3c.github.io/pointerevents/#the-button-property

### DIFF
--- a/packages/react-events/src/dom/ContextMenu.js
+++ b/packages/react-events/src/dom/ContextMenu.js
@@ -28,20 +28,19 @@ type ContextMenuState = {
 };
 
 type ContextMenuEvent = {|
-  target: Element | Document,
-  type: 'contextmenu',
-  pointerType: PointerType,
-  timeStamp: number,
-  clientX: null | number,
-  clientY: null | number,
-  pageX: null | number,
-  pageY: null | number,
-  x: null | number,
-  y: null | number,
   altKey: boolean,
+  buttons: 0 | 1 | 2,
   ctrlKey: boolean,
   metaKey: boolean,
+  pageX: null | number,
+  pageY: null | number,
+  pointerType: PointerType,
   shiftKey: boolean,
+  target: Element | Document,
+  timeStamp: number,
+  type: 'contextmenu',
+  x: null | number,
+  y: null | number,
 |};
 
 const hasPointerEvents =
@@ -60,13 +59,13 @@ function dispatchContextMenuEvent(
 
   const gestureState = {
     altKey: nativeEvent.altKey,
-    button: nativeEvent.button === 0 ? 'primary' : 'auxillary',
-    ctrlKey: nativeEvent.altKey,
-    metaKey: nativeEvent.altKey,
-    pageX: nativeEvent.altKey,
-    pageY: nativeEvent.altKey,
+    buttons: nativeEvent.buttons != null ? nativeEvent.buttons : 0,
+    ctrlKey: nativeEvent.ctrlKey,
+    metaKey: nativeEvent.metaKey,
+    pageX: nativeEvent.pageX,
+    pageY: nativeEvent.pageY,
     pointerType,
-    shiftKey: nativeEvent.altKey,
+    shiftKey: nativeEvent.shiftKey,
     target,
     timeStamp,
     type: 'contextmenu',
--- a/packages/react-events/src/dom/Press.js
+++ b/packages/react-events/src/dom/Press.js
@@ -75,24 +75,24 @@ type PressEventType =
   | 'presschange';
 
 type PressEvent = {|
-  button: 'primary' | 'auxillary',
-  defaultPrevented: boolean,
-  target: Element | Document,
-  type: PressEventType,
-  pointerType: PointerType,
-  timeStamp: number,
+  altKey: boolean,
+  buttons: 0 | 1 | 4,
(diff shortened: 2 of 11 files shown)


packages/react-events/src/dom/Press.js
```diff
@@ -581,11 +586,12 @@ const pressResponderImpl = {
             state.activePointerId = touchEvent.identifier;
           }
 
-          // Ignore any device buttons except primary/auxillary and touch/pen contact.
+          // Ignore any device buttons except primary/secondary and touch/pen contact.
           // Additionally we ignore primary-button + ctrl-key with Macs as that
           // acts like right-click and opens the contextmenu.
           if (
-            nativeEvent.button > 1 ||
+            nativeEvent.buttons === 2 ||
```
necolas: @trueadm When this condition included `nativeEvent.buttons != null` it caused this `Input` test to fail: https://github.com/facebook/react/blob/4279455/packages/react-events/src/dom/__tests__/Input-test.internal.js#L1038. We probably shouldn't be using `Press` in the `Input` tests or testing core logic in the responder tests.

trueadm: I agree that maybe we shouldn't be including other responder tests in other test, but we should definitely have overlapping responder tests – meaning the original test is still very much valuable.

packages/react-events/src/dom/Press.js
```diff
@@ -75,24 +75,24 @@ type PressEventType =
...
 
 type PressEvent = {|
-  button: 'primary' | 'auxillary',
-  defaultPrevented: boolean,
-  target: Element | Document,
-  type: PressEventType,
-  pointerType: PointerType,
-  timeStamp: number,
+  altKey: boolean,
+  buttons: 1 | 4,
```
trueadm: Shouldn't this be `1 | 2 | 4` if its bitwise?

necolas: We don't use right-click anymore in this responder

necolas: Also looks like `pointerup` always have `buttons` as `0`, so will need that too.

packages/react-events/src/dom/ContextMenu.js
```diff
@@ -28,20 +28,19 @@ type ContextMenuState = {
...
-  pointerType: PointerType,
-  timeStamp: number,
-  clientX: null | number,
-  clientY: null | number,
-  pageX: null | number,
-  pageY: null | number,
-  x: null | number,
-  y: null | number,
   altKey: boolean,
+  buttons: null | 1 | 2,
```
necolas: can also be `0` (touch events)

packages/react-events/src/dom/ContextMenu.js
```diff
@@ -60,13 +59,13 @@ function dispatchContextMenuEvent(
 
   const gestureState = {
     altKey: nativeEvent.altKey,
-    button: nativeEvent.button === 0 ? 'primary' : 'auxillary',
-    ctrlKey: nativeEvent.altKey,
-    metaKey: nativeEvent.altKey,
-    pageX: nativeEvent.altKey,
-    pageY: nativeEvent.altKey,
+    buttons: nativeEvent.buttons != null ? nativeEvent.buttons : 1,
```
necolas: Default to `0` probably

packages/react-events/src/dom/testing-library/domEventSequences.js
```diff
@@ -147,7 +150,7 @@ export function pointermove(target, payload) {
 export function pointerup(target, defaultPayload) {
   const dispatch = arg => target.dispatchEvent(arg);
   const pointerType = getPointerType(defaultPayload);
-  const payload = {button: 0, buttons: 2, ...defaultPayload};
+  const payload = {buttons: buttonsType.primary, ...defaultPayload};
```
necolas: This should be `buttonsType.none`

packages/react-events/src/dom/testing-library/domEvents.js
```diff
@@ -314,15 +328,21 @@ export function pointerover(payload) {
 }
 
 export function pointerup(payload) {
-  return createPointerEvent('pointerup', {button: 0, buttons: 2, ...payload});
+  return createPointerEvent('pointerup', {
+    buttons: buttonsType.primary,
```
necolas: Use `none`

packages/react-events/src/dom/testing-library/domEvents.js
```diff
@@ -346,7 +366,10 @@ export function mouseover(payload) {
 }
 
 export function mouseup(payload) {
-  return createMouseEvent('mouseup', {button: 0, buttons: 2, ...payload});
+  return createMouseEvent('mouseup', {
+    buttons: buttonsType.primary,
```
necolas: Check this again, the browser probably makes this `0` (`none`) too

trueadm:
Seems like there's maybe a case we haven't handled here. Middle mouse buttons are reported with `event.button` of `1` but the `event.buttons` value is `0`.

necolas:
Reported as `0` where?

PR #16516, opened 2019-08-20
Use renderToStaticMarkup for tests

devknoll:
Make sure the tests under `renderToStaticMarkup` actually use `renderToStaticMarkup`, rather than `renderToString`.


trueadm:
This change is not necessary as it doesn't affect the output of the test case here. Thanks for looking anyway.

trueadm:
Update: looking through the PR, it seems completely reasonable – sorry for the confusion.

PR #16517, opened 2019-08-20
 [react-events] Press responder simplification 

necolas:
NOTE: this stacks on top of #16479 which I'll merge once we've internally synced what is currently in master. Preview-review should only look at the second commit in this PR "Press responder simplification".

* Remove 'pressRetentionOffset' so we don't have to perform layout measurements during moves.
* Maintain the pressed state while a pointer is down and moving, even if it moves outside the target.
* Determine whether to call 'onPress' based on whether the pointer was released over the target.

This aligns more closely with how the web natively deals with pointers and simplifies the responder.
We only need to call getBoundingClientRect for touch pointers. Further revisions could be made to
avoid target capturing for touch pointers with PointerEvents, which would leave the gBCR call only
for the fallback TouchEvents.

### DIFF
--- a/packages/react-events/src/dom/Press.js
+++ b/packages/react-events/src/dom/Press.js
@@ -22,12 +22,6 @@ import {DiscreteEvent, UserBlockingEvent} from 'shared/ReactTypes';
 
 type PressProps = {|
   disabled: boolean,
-  pressRetentionOffset: {
-    top: number,
-    right: number,
-    bottom: number,
-    left: number,
-  },
   preventDefault: boolean,
   stopPropagation: boolean,
   onPress: (e: PressEvent) => void,
@@ -46,21 +40,8 @@ type PressState = {
   isActivePressed: boolean,
   isActivePressStart: boolean,
   isPressed: boolean,
-  isPressWithinResponderRegion: boolean,
   pointerType: PointerType,
   pressTarget: null | Element | Document,
-  responderRegionOnActivation: null | $ReadOnly<{|
-    bottom: number,
-    left: number,
-    right: number,
-    top: number,
-  |}>,
-  responderRegionOnDeactivation: null | $ReadOnly<{|
-    bottom: number,
-    left: number,
-    right: number,
-    top: number,
-  |}>,
   ignoreEmulatedMouseEvents: boolean,
   activePointerId: null | number,
   shouldPreventClick: boolean,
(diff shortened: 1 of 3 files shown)


packages/react-events/src/dom/Press.js
```diff
@@ -660,45 +536,22 @@ const pressResponderImpl = {
...
-          } else {
-            // Calculate the responder region we use for deactivation, as the
-            // element dimensions may have changed since activation.
-            updateIsPressWithinResponderRegion(
-              touchEvent || nativeEvent,
+        if (isPressed) {
+          const onPressMove = props.onPressMove;
+
+          if (isFunction(onPressMove)) {
+            dispatchEvent(
```
trueadm: Won't this mean that we fire pressmove even if our pointer might be outside the target?

packages/react-events/src/dom/__tests__/Press-test.internal.js
```diff
@@ -698,10 +612,8 @@ describe.each(environmentTable)('Press responder', hasPointerEvents => {
         'onPressStart',
         'onPressChange',
         'onPressMove',
-        'onPressEnd',
-        'onPressChange',
-        'onPressStart',
-        'onPressChange',
+        'onPressMove',
```
trueadm: This relates to my comment above, surely if you move out, then end should fire?

trueadm:
> Maintain the pressed state while a pointer is down and moving, even if it moves outside the target.

This is a big change internally right? Are the plans to move the pointer moving behaviour to a `Tap` responder – if so, ignore my comments.

packages/react-events/src/dom/Press.js
necolas: Yep

packages/react-events/src/dom/__tests__/Press-test.internal.js
necolas: We can do that for events that aren't capturing the target. For now this is fine and consistent across pointers

necolas:
This is just how web buttons work anyway. I don't think anyone is depending on how press move works. And there's no other option for touch events without measuring the hit rect every move which is not good for perf

necolas:
I'm going to abandon this and create a new responder – `Tap` – so we can start fresh and be more intentional about swapping existing product code from `Press` to `Tap` when we're happy with the implementation.

PR #16532, opened 2019-08-21
[Partial Hydration] Dispatching events should not work until hydration commits

sebmarkbage:
This includes a bunch of clean up to set up for selective hydration.

Primarily the bug it fixes is that we attach event listeners while hydrating but these need to be noops (and later replayable) until we actually commit. Since the parent may not be mounted any setStates that we let through might become invalid.

There's no way around that we need some commit-phase effect to indicate when it's done. We could do these on all event listeners but it's sufficient to just do it at the root of the things we're hydrating.

In fact, we did that with HostRoot already by attaching a fake Placement effect tag on it.

This PR formalizes that by making Hydrating a new effect tag. It means that this is the beginning of a in-progress hydrating subtree. The tag gets removed in the commit phase to indicate that the tree is mounted. Same as Placement.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
@@ -25,6 +25,7 @@ describe('ReactDOMServerPartialHydration', () => {
     ReactFeatureFlags = require('shared/ReactFeatureFlags');
     ReactFeatureFlags.enableSuspenseServerRenderer = true;
     ReactFeatureFlags.enableSuspenseCallback = true;
+    ReactFeatureFlags.enableFlareAPI = true;
 
     React = require('react');
     ReactDOM = require('react-dom');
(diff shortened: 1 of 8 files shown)


packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
```diff
@@ -1729,4 +1729,91 @@ describe('ReactDOMServerPartialHydration', () => {
...
+    let root = ReactDOM.unstable_createRoot(container, {hydrate: true});
+    root.render(<App />);
+    Scheduler.unstable_flushAll();
+    jest.runAllTimers();
+
+    expect(container.textContent).toBe('Click meHello');
+
+    // We're now partially hydrated.
+    a.click();
+    expect(clicks).toBe(0);
```
sebmarkbage: This is the thing that this PR fixes. It used to let these through.

trueadm:
Please can you add a test for the new event system? You could add a test in `DOMEventResponderSystem-test.internal`. Notably, event responders mount during complete phase, so this will likely have to move to the commit phase now:

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberCompleteWork.js#L693-L774

sebmarkbage:
@trueadm That seems like a bigger issue we need to deal with. Particularly, this is not concurrent mode safe: https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberCompleteWork.js#L1346-L1347

trueadm:
That code was designed to work like the existing model for assigning props against the DOM instance, which is also done in the complete phase. I’m happy to change it, but it’s based on previous feedback that I guess no longer holds with concurrent mode?

sebmarkbage:
@trueadm It assigned against the DOM in the complete phase for initial mount (when there is no current). That normally work since you can't fire an event on something that doesn't exist in the DOM yet (except for hydration). That's the same thing that this PR is dealing with. That part is legit, so I'll add a test for the same thing for Flare.

However, for updates to an already mounted DOM component we do it in the commit phase since the old listeners remain valid until we're ready to commit. https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMHostConfig.js#L371

trueadm:
Seems like those two lines you highlighted need to be moved to the commit phase then and then it should work as expected. I can do that tomorrow :)

PR #16540, opened 2019-08-22
[react-events] Ensure updateEventListeners updates in commit phase

trueadm:
This is a follow up to the comment @sebmarkbage pointed out in https://github.com/facebook/react/pull/16532#issuecomment-523673963. Notably, we should only be updating EventResponder instances in the commit phase. EventResponder instances should continue to mount in the complete phase.

I also managed to resolve the issue that prevented us from moving the event fiber logic into the `ReactFiberEvents` file, as that logic is now shared between the complete and commit phase files.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -824,10 +824,9 @@ export function mountResponderInstance(
   responderProps: Object,
   responderState: Object,
   instance: Instance,
-  rootContainerInstance: Container,
 ): ReactDOMEventResponderInstance {
   // Listen to events
-  const doc = rootContainerInstance.ownerDocument;
+  const doc = instance.ownerDocument;
   const documentBody = doc.body || doc;
   const {
     rootEventTypes,
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -118,6 +118,7 @@ import {
 } from './ReactHookEffectTags';
 import {didWarnAboutReassigningProps} from './ReactFiberBeginWork';
 import {runWithPriority, NormalPriority} from './SchedulerWithReactIntegration';
+import {updateEventListeners} from './ReactFiberEvents';
 
 let didWarnAboutUndefinedSnapshotBeforeUpdate: Set<mixed> | null = null;
 if (__DEV__) {
@@ -1330,6 +1331,13 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
             newProps,
             finishedWork,
           );
+        }
+        if (enableFlareAPI) {
+          const prevListeners = oldProps.listeners;
+          const nextListeners = newProps.listeners;
+          if (prevListeners !== nextListeners) {
+            updateEventListeners(nextListeners, instance, finishedWork);
+          }
         }
       }
       return;
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -9,12 +9,7 @@
 
 import type {Fiber} from './ReactFiber';
 import type {ExpirationTime} from './ReactFiberExpirationTime';
-import type {
-  ReactEventResponder,
-  ReactEventResponderInstance,
-  ReactFundamentalComponentInstance,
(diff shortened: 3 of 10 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1331,6 +1332,13 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
             finishedWork,
           );
         }
+        if (enableFlareAPI) {
+          const prevListeners = oldProps.listeners;
+          const nextListeners = newProps.listeners;
+          if (prevListeners !== nextListeners) {
+            updateEventListeners(nextListeners, instance, finishedWork);
```
trueadm: This is the main change of this PR.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -693,14 +680,8 @@ function completeWork(
         if (enableFlareAPI) {
           const prevListeners = current.memoizedProps.listeners;
           const nextListeners = newProps.listeners;
-          const instance = workInProgress.stateNode;
           if (prevListeners !== nextListeners) {
-            updateEventListeners(
```
trueadm: This is where we used to update Event Responders

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -742,12 +723,7 @@ function completeWork(
             const instance = workInProgress.stateNode;
             const listeners = newProps.listeners;
             if (listeners != null) {
-              updateEventListeners(
-                listeners,
-                instance,
-                rootContainerInstance,
-                workInProgress,
-              );
+              updateEventListeners(listeners, instance, workInProgress);
```
trueadm: This is where we mount Event Responders

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -764,12 +740,7 @@ function completeWork(
           if (enableFlareAPI) {
             const listeners = newProps.listeners;
             if (listeners != null) {
-              updateEventListeners(
-                listeners,
-                instance,
-                rootContainerInstance,
-                workInProgress,
-              );
+              updateEventListeners(listeners, instance, workInProgress);
```
trueadm: This is where we mount Event Responders

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -1258,156 +1229,4 @@ function completeWork(
   return null;
 }
 
-function mountEventResponder(
```
trueadm: Moved this code block into ReactFiberEvents

packages/react-reconciler/src/ReactFiberEvents.js
```diff
@@ -1,66 +1,24 @@
-/**
- * Copyright (c) Facebook, Inc. and its affiliates.
```
necolas: Copyright notice should probably be preserved

sebmarkbage:
Looks like when you added back the `@flow` annotation, the coverage was enabled again and surfaced a Flow error.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -824,10 +824,9 @@ export function mountResponderInstance(
   responderProps: Object,
   responderState: Object,
   instance: Instance,
-  rootContainerInstance: Container,
 ): ReactDOMEventResponderInstance {
   // Listen to events
-  const doc = rootContainerInstance.ownerDocument;
```
sebmarkbage: I believe that we originally built the old event system around the  `rootContainerInstance` so that we had the option to move to a model where we attach the listeners on the `rootContainerInstance` instead of the `ownerDocument`. The idea was that if you attach it on the root, then it doesn't interfere with siblings trees and stopping propagation can stop it to non-React parents. That's why `rootContainerInstance` is passed around everywhere even though it doesn't really matter much which one is used (other than for portals into iframes).

trueadm: I know, but we never actually use it and it saves having to store the `rootContainerInstance` somewhere to be be passed in during the commit phase (which is why I removed it, as I couldn't see an easy way to get it in the commit phase again).

PR #16542, opened 2019-08-22
[Scheduler] Profiler Features (second try)

gaearon:
This is a revert of a revert of #16145 + my attempts to fix its issues.
You need to read individual commits.

The main fix has already landed independently in https://github.com/facebook/react/pull/16407.

However, there were a few more issues in #16145 that I'm fixing in this PR:

* Bad copy paste for re-exports (fixed in https://github.com/facebook/react/commit/935526e121add025ff74df8f7dde06de27bf7c2e)
* Init time dependency on ArrayBuffer which would break IE9 (fixed in ee548c3cc17dc58c000cf6b13278c4b08642050f)
* Potential deopt for tasks due to adding a field later (fixed in 9c30b0b695d81e9c43b296ab93d895e4416ef713)

There is a bigger conceptual issue though that I'm struggling with.

(Feel free to skip these details)

<details>

#16145 added a bunch of `pop()` calls — for example, when a task errors or get cancelled. My understanding is that this was done as an optimization. I.e. we don't need to schedule another loop if there's no actual tasks in the loop left.

That optimization doesn't seem like it would cover all cases. For example, if we have tasks [A, B, C], and then cancel C, B, and A in sequence, only A would be popped (because it's the first item). So this optimization seems "optional" and not something we can rely on.

However, the tests *do* rely on this optimization being present.
Without it, the "main thread" bar never stops:

```
 Scheduler › marks when a task is canceled

   !!! Main thread              │      ██
    Task 1 [Normal]              │██████░░🡐 canceled
    "
    Received:
      "
    !!! Main thread              │      ██████████████████████
    Task 1 [Normal]              │██████░░🡐 canceled


Scheduler › marks when a task errors

    Expected value to equal:
      "
    !!! Main thread              │
    Task 1 [Normal]              │██████🡐 errored
    "
    Received:
      "
    !!! Main thread              │        ████████████████████
    Task 1 [Normal]              │██████🡐 errored
    "

Scheduler › marks when multiple tasks are canceled

    Expected value to equal:
      "
    !!! Main thread              │      ██
    Task 1 [Normal]              │██████░░🡐 canceled
    Task 2 [Normal]              │░░░░░░░░🡐 canceled
    "
    Received:
      "
    !!! Main thread              │      ██████████████████████
    Task 1 [Normal]              │██████░░🡐 canceled
    Task 2 [Normal]              │░░░░░░░░🡐 canceled
    "
```

I don't know how to interpret this.

Is this optimization essential for profiler's _correctness_? In that case it seems dodgy that depending on the order in which you cancel callbacks, it would either work or not.

Or is this optimization purely an optimization? Are the test asserting on implementation details too aggressive? Why does the "Main Thread" bar never stop? Or are we missing a "suspended" event that should have happened in that case?

Finally, the logic to determine whether we "suspended" or "resumed" seems fragile to me. I don't know for sure what those terms are supposed to mean. For example, if a callback errors, did we "suspend"? It seems like no from the code, but I don't know why. Conceptually, I'd think of an "error" as "suspending" the scheduler loop. Unless "suspend" means something else here.

My last commit just removes the "optimization" (9a6440440ef9b0100579d70abfb1c51592922bfb). **It breaks some tests but I can't move further without understanding their intent.** Is it the code, the assertions, or the test setup that needs to change?

Was this more than an optimization? I'll leave it at that so someone else can pick it up. :-)

</details>

I chatted to Boone, and arrived at a solution that satisfies both of us. Basically, I wasn't sure what Suspended and Resumed events mean. Boone told me it's okay to treat them as "exit the loop" and "enter the loop".

So in the next commits I did the following:

* Removed the `pop()` optimization that affected non-profiling paths to reduce the risk. (9a6440440ef9b0100579d70abfb1c51592922bfb)
  - This broke tests because they seemed to assert on implementation details
* Strengthened tests to explicitly treat Suspend and Resumed as "exit" and "enter", and added assertions for that (8482f5de03f1621d6025bc0cf87da3d565c2f979)
  - This further broke tests because we didn't always emit these events in pairs
- Finally, made Suspended and Resumed match 1:1 to "exiting" and "entering" the flush (1f5f7df068a52aeb180797aa21dea1aef0329781)
  - I had to amend tests to match this new behavior. These changes make sense to me. We now flag "main thread" time as anything not spent in the Scheduler.

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -8,10 +8,14 @@
 
 /* eslint-disable no-var */
 
-import {enableSchedulerDebugging} from './SchedulerFeatureFlags';
 import {
-  requestHostCallback,
+  enableSchedulerDebugging,
+  enableProfiling,
+} from './SchedulerFeatureFlags';
+import {
+  requestHostCallback as requestHostCallbackWithoutProfiling,
   requestHostTimeout,
+  cancelHostCallback,
   cancelHostTimeout,
   shouldYieldToHost,
   getCurrentTime,
@@ -21,11 +25,26 @@ import {
 import {push, pop, peek} from './SchedulerMinHeap';
 
 // TODO: Use symbols?
-var ImmediatePriority = 1;
-var UserBlockingPriority = 2;
-var NormalPriority = 3;
-var LowPriority = 4;
-var IdlePriority = 5;
+import {
+  ImmediatePriority,
+  UserBlockingPriority,
+  NormalPriority,
+  LowPriority,
+  IdlePriority,
+} from './SchedulerPriorities';
+import {
+  sharedProfilingBuffer,
+  markTaskRun,
+  markTaskYield,
+  markTaskCompleted,
+  markTaskCanceled,
+  markTaskErrored,
+  markSchedulerSuspended,
+  markSchedulerUnsuspended,
+  markTaskStart,
+  stopLoggingProfilingEvents,
+  startLoggingProfilingEvents,
+} from './SchedulerProfiling';
 
 // Max 31 bit integer. The max integer size in V8 for 32-bit systems.
 // Math.pow(2, 30) - 1
@@ -60,14 +79,16 @@ var isPerformingWork = false;
 var isHostCallbackScheduled = false;
 var isHostTimeoutScheduled = false;
 
-function flushTask(task, callback, currentTime) {
-  currentPriorityLevel = task.priorityLevel;
-  var didUserCallbackTimeout = task.expirationTime <= currentTime;
-  var continuationCallback = callback(didUserCallbackTimeout);
-  return typeof continuationCallback === 'function'
-    ? continuationCallback
-    : null;
+function requestHostCallbackWithProfiling(cb, time) {
+  if (enableProfiling) {
+    markSchedulerSuspended(time);
+    requestHostCallbackWithoutProfiling(cb);
+  }
 }
+
+const requestHostCallback = enableProfiling
+  ? requestHostCallbackWithProfiling
+  : requestHostCallbackWithoutProfiling;
 
 function advanceTimers(currentTime) {
   // Check for tasks that are no longer delayed and add them to the queue.
@@ -81,6 +102,10 @@ function advanceTimers(currentTime) {
       pop(timerQueue);
       timer.sortIndex = timer.expirationTime;
       push(taskQueue, timer);
+      if (enableProfiling) {
+        markTaskStart(timer);
+        timer.isQueued = true;
+      }
     } else {
       // Remaining timers are pending.
       return;
@@ -96,7 +121,7 @@ function handleTimeout(currentTime) {
   if (!isHostCallbackScheduled) {
     if (peek(taskQueue) !== null) {
       isHostCallbackScheduled = true;
-      requestHostCallback(flushWork);
+      requestHostCallback(flushWork, currentTime);
     } else {
       const firstTimer = peek(timerQueue);
       if (firstTimer !== null) {
@@ -107,6 +132,10 @@ function handleTimeout(currentTime) {
 }
 
 function flushWork(hasTimeRemaining, initialTime) {
+  if (isHostCallbackScheduled) {
+    markSchedulerUnsuspended(initialTime);
+  }
+
   // We'll need a host callback the next time work is scheduled.
   isHostCallbackScheduled = false;
   if (isHostTimeoutScheduled) {
@@ -135,15 +164,24 @@ function flushWork(hasTimeRemaining, initialTime) {
       const callback = currentTask.callback;
       if (callback !== null) {
         currentTask.callback = null;
-        const continuation = flushTask(currentTask, callback, currentTime);
-        if (continuation !== null) {
-          currentTask.callback = continuation;
+        currentPriorityLevel = currentTask.priorityLevel;
+        const didUserCallbackTimeout =
+          currentTask.expirationTime <= currentTime;
+        markTaskRun(currentTask, currentTime);
+        const continuationCallback = callback(didUserCallbackTimeout);
+        currentTime = getCurrentTime();
+        if (typeof continuationCallback === 'function') {
+          currentTask.callback = continuationCallback;
+          markTaskYield(currentTask, currentTime);
         } else {
+          if (enableProfiling) {
+            markTaskCompleted(currentTask, currentTime);
+            currentTask.isQueued = false;
+          }
           if (currentTask === peek(taskQueue)) {
             pop(taskQueue);
           }
         }
-        currentTime = getCurrentTime();
         advanceTimers(currentTime);
       } else {
         pop(taskQueue);
@@ -152,6 +190,8 @@ function flushWork(hasTimeRemaining, initialTime) {
     }
     // Return whether there's additional work
     if (currentTask !== null) {
+      markSchedulerSuspended(currentTime);
+      isHostCallbackScheduled = true;
       return true;
     } else {
       let firstTimer = peek(timerQueue);
@@ -160,6 +200,18 @@ function flushWork(hasTimeRemaining, initialTime) {
       }
       return false;
     }
+  } catch (error) {
+    if (currentTask !== null) {
+      if (enableProfiling) {
+        const currentTime = getCurrentTime();
+        markTaskErrored(currentTask, currentTime);
+        currentTask.isQueued = false;
+      }
+      if (currentTask === peek(taskQueue)) {
+        pop(taskQueue);
+      }
+    }
+    throw error;
   } finally {
     currentTask = null;
     currentPriorityLevel = previousPriorityLevel;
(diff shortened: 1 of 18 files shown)


gaearon:
Regarding "suspended" event. I'm not sure if it's supposed to mean any time the loop stops. Or if it's only supposed to mean that work loop stopped but there's more work to do later.

gaearon:
This is ready for review. Production bundle diff: https://gist.github.com/gaearon/b3ee8dadf9d4fac5f407dfa9e3c247fc/revisions

packages/scheduler/src/Scheduler.js
```diff
@@ -135,15 +152,24 @@ function flushWork(hasTimeRemaining, initialTime) {
...
         } else {
+          if (enableProfiling) {
+            markTaskCompleted(currentTask, currentTime);
+            currentTask.isQueued = false;
+          }
           if (currentTask === peek(taskQueue)) {
             pop(taskQueue);
           }
         }
-        currentTime = getCurrentTime();
```
gaearon: I'm not sure why this removed. Bring it back doesn't fail any tests. :-/ I can revert this one if @sebmarkbage doesn't know either.

packages/scheduler/src/Scheduler.js
```diff
@@ -160,10 +186,23 @@ function flushWork(hasTimeRemaining, initialTime) {
...
     }
+  } catch (error) {
+    if (enableProfiling) {
+      if (currentTask !== null) {
+        const currentTime = getCurrentTime();
+        markTaskErrored(currentTask, currentTime);
+        currentTask.isQueued = false;
+      }
+    }
+    throw error;
```
gaearon: It's kind of non-ideal that in prod path we now just have `catch (err) { throw error; }`. I could restructure to make it profile-only but didn't bother.

packages/scheduler/src/Scheduler.js
gaearon: Actually I see now. It's because after inlining `flushTask` we already recalc time just above. So this became unnecessary.

packages/scheduler/src/SchedulerProfiling.js
```diff
@@ -0,0 +1,214 @@
...
+    : // $FlowFixMe Flow doesn't know about ArrayBuffer
+      typeof ArrayBuffer === 'function'
+      ? new ArrayBuffer(profilingStateSize * Int32Array.BYTES_PER_ELEMENT)
+      : null // Don't crash the init path on IE9
+  : null;
+
+const profilingState =
+  enableProfiling && sharedProfilingBuffer !== null
+    ? new Int32Array(sharedProfilingBuffer)
+    : null;
```
aweary: Why not just fallback to a regular array?

gaearon: shrug.png

packages/scheduler/src/Scheduler.js
sebmarkbage: Yea this is not great. Does this cause any deopts that we know of? Would be an extremely bad place to deopt.

Let's just hoist this body out into a separate function that is either wrapped in catch or not. It should get inlined by closure, however, maybe we don't actually want it to inline since that limits the deopt to the outer function in those older VMs.

packages/scheduler/src/Scheduler.js
```diff
@@ -261,21 +302,45 @@ function unstable_scheduleCallback(priorityLevel, callback, options) {
       typeof options.timeout === 'number'
         ? options.timeout
         : timeoutForPriorityLevel(priorityLevel);
+    // if (enableProfiling) {
+    //   var _label = options.label;
+    //   if (typeof _label === 'string') {
+    //     label = _label;
+    //   }
+    // }
```
sebmarkbage: ?

gaearon: That's from original PR. I can remove. I think it's a TODO.

gaearon: TODO itself:

https://github.com/facebook/react/blob/1f5f7df068a52aeb180797aa21dea1aef0329781/packages/scheduler/src/Scheduler.js#L292-L293

packages/scheduler/src/Scheduler.js
```diff
@@ -370,3 +447,16 @@ export {
   getCurrentTime as unstable_now,
   forceFrameRate as unstable_forceFrameRate,
 };
+
+export const unstable_startLoggingProfilingEvents = enableProfiling
```
sebmarkbage: This is unfortunate that this adds so many bytes. Don't we have a way to conditionally export?

gaearon: No but we could pull them all in their own namespace.

gaearon: I think we could pull all "advanced features" in a namespace for now.

packages/scheduler/src/Scheduler.js
```diff
@@ -261,21 +302,45 @@ function unstable_scheduleCallback(priorityLevel, callback, options) {
...
-    priorityLevel,
-    startTime,
-    expirationTime,
-    sortIndex: -1,
-  };
+  var newTask;
+  if (enableProfiling) {
+    newTask = {
+      // __PROFILER__ only
+      isQueued: false,
```
sebmarkbage: Just create all the normal fields above and then append this by mutation. Closure will know to inline it into the object literal.

packages/scheduler/src/Scheduler.js
```diff
@@ -261,21 +302,45 @@ function unstable_scheduleCallback(priorityLevel, callback, options) {
...
+      callback,
+      priorityLevel,
+      startTime,
+      expirationTime,
+      sortIndex: -1,
+    };
+  } else {
+    newTask = {
+      // !!! Keep these in sync with above branch !!!
+      id: ++taskIdCounter,
```
sebmarkbage: Why is this `++taskIdCounter` now instead of `taskIdCounter++`?

gaearon: I think this is unobservable. Maybe Andrew wanted to align it with the internal counter that's currently off by one? I can revert this bit.

packages/scheduler/src/Scheduler.js
gaearon: Wow I had no idea. I'll remove my change then.

packages/scheduler/src/Scheduler.js
gaearon: Oh this is because of

```js
profilingState[CURRENT_TASK_ID] = 0;
```

I bet. I could just change the first ID to `1` though.

packages/scheduler/src/SchedulerProfiling.js
sebmarkbage: That's an interesting approach because it would allow us to remove the null checks below and it kind of keeps working, except it wouldn't be able to be introspected since there's no buffer exposed. I like it just for byte savings though.

packages/scheduler/src/Scheduler.js
sebmarkbage: I think it's safe to remove this TODO because I don't think we should put this feature in the profiler itself but instead expose the ID so that someone outside can associate the ID with a label if they want.

packages/scheduler/src/Scheduler.js
gaearon: Resolved

packages/scheduler/src/SchedulerProfiling.js
gaearon: Is there any harm to always doing `new Int32Array(sharedProfilingBuffer)`? Even when it's off? We'd just not write to it when profiling is off. And when buffer is not supported, we'd pass `null` which would give us an empty array.

I guess it doesn't work in IE9 though so probably should be guarded.

gaearon:
New prod diff: https://gist.github.com/gaearon/145750f6358cfb00b3e0a9e60826566a/revisions

packages/scheduler/src/SchedulerProfiling.js
```diff
@@ -0,0 +1,203 @@
...
+    typeof SharedArrayBuffer === 'function'
+    ? new SharedArrayBuffer(profilingStateSize * Int32Array.BYTES_PER_ELEMENT)
+    : // $FlowFixMe Flow doesn't know about ArrayBuffer
+      typeof ArrayBuffer === 'function'
+      ? new ArrayBuffer(profilingStateSize * Int32Array.BYTES_PER_ELEMENT)
+      : null // Don't crash the init path on IE9
+  : null;
+
+const profilingState =
+  enableProfiling && sharedProfilingBuffer !== null
```
gaearon: Keeping this `sharedProfilingBuffer !== null` check because while `new Int32Array(null)` would work, IE10 which doesn't have `ArrayBuffer` *also* doesn't have `Int32Array`.

packages/scheduler/src/Scheduler.js
```diff
@@ -118,52 +135,82 @@ function flushWork(hasTimeRemaining, initialTime) {
...
-      if (callback !== null) {
-        currentTask.callback = null;
-        const continuation = flushTask(currentTask, callback, currentTime);
-        if (continuation !== null) {
-          currentTask.callback = continuation;
-        } else {
-          if (currentTask === peek(taskQueue)) {
-            pop(taskQueue);
-          }
+    if (enableProfiling) {
```
sebmarkbage: I figured you'd fork the finally too but I guess this works too.

Issue #16526, opened 2019-08-21
Custom hook and strange warnings 


brunoscopelliti:
I've a question about a behaviour that appears strange to me.

I am writing a custom hook.

```js
const useToggle =
  (initialState) => {
    const [isTrue, setState] = useState(initialState || false);

    const False =
      () => {
        setState(false);
      };

    const True =
      () => {
        setState(true);
      };

    return [isTrue, True, False];
  };
```

It works fine.

But if I try to rewrite `False`, and `True` using `setState.bind` I get the following warning in console.

```js
const False = setState.bind(null, false);

const True = setState.bind(null, true);
```

The warning:

```
Warning: State updates from the useState() and useReducer() Hooks don't support the second callback argument.
To execute a side effect after rendering, declare it in the component body with useEffect().
```

Is this the expected behaviour?

Why is that? The two snippet appear equivalent to me.

gaearon:
I'm guessing you're doing something like `onClick={True}` which results in `onClick={e => True(e)}` or `onClick={r => setState(true, e)}`. React thinks you're trying to pass a second argument to `setState` and warns it's not supported. Maybe we could scope down the warning for just the case when you pass a function.

brunoscopelliti:
Oh, I see.
So there's nothing particularly wrong about my custom hook.
@gaearon  Thank you for the super fast answer.

> Maybe we could scope down the warning for just the case when you pass a function.

It makes sense to me. If you think is a good idea, I can help with a PR in the coming days. Just let me know.

gaearon:
Sure, a PR sounds fine.

threepointone:
Closing since the fix landed. Thanks @brunoscopelliti!

PR #16543, opened 2019-08-22
Only warn in case the fourth argument is a function

brunoscopelliti:
Ref. #16526.

As per @gaearon 's suggestion, this PR scope down the warning to just the case when second argument is a function.


gaearon:
Thanks.

PR #16555, opened 2019-08-23
Add trusted types to react on server side

Siegrift:
## Trusted Types
[Trusted Types](https://github.com/WICG/trusted-types) ([spec](https://wicg.github.io/trusted-types/dist/spec/), [introductory article](https://developers.google.com/web/updates/2019/02/trusted-types)) is a new experimental DOM API implemented within the WICG , with a working [Chrome implementation](https://www.chromestatus.com/feature/5650088592408576).

The API creates a few new objects available on the global object in the browser, like most other web APIs ([impl in TS](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/trusted-types/index.d.ts) and in [Closure compiler](https://github.com/WICG/trusted-types/blob/master/externs/externs.js)).

Under certain conditions, controlled by a HTTP header (analogous to Content-Security-Policy behavior), the API can enable the enforcement - then it changes the signature of several DOM API functions and property setters, such that they accept specific object types, and reject strings. Colloquially, DOM API becomes strongly typed.

For example, with Trusted Types Element.innerHTML property setter accepts a TrustedHTML object.

Trusted Type objects stringify to their inner value. This API shape is a deliberate choice that enables existing web applications and libraries to gradually migrate from strings to Trusted Types without breaking functionality. In our example, it makes it possible to write the following:

```javascript
const policy = TrustedTypes.createPolicy('foo', { 
  createHTML: (s) => { /* some validation*/; return s} 
});

const trustedHTML = policy.createHTML('bar');
anElement.innerHTML = trustedHTML

anElement.innerHTML === 'bar'
```

The above code works regardless if the Trusted Types enforcement is enabled or not.

Reading from the DOM is unaffected, so Element.innerHTML getter returns a string. That's for practical reasons -- web applications read from DOM more often than they write to it, and only writing exposes the application to DOM XSS risks. Typing only the setters allows us to secure web applications with minimal code changes.

## Adding Trusted Types to React for server side
Unfortunately, there are no Trusted Types (TT) on server side. However, it’s really easy to introduce a reflected xss attack via server side rendering. Markup rendered on server side is rendered to string (without any DOM emulation) and returned as a response from the server and there is no way TT can prevent this attack. This creates inconsistency when rendering on client and server side (client side would fail with TT violation), which shouldn’t happen. Also, if application uses hot reloading, you will get a TT error after each reload (because React templates are re-rendered to DOM, this time on client).

This PR enables applications to use Trusted Types on server side. Functions inside [ReactDOMNodeStreamRenderer.js](https://github.com/facebook/react/blob/fc807720780962729df2ad1b0f53258603c020c7/packages/react-dom/src/server/ReactDOMStringRenderer.js) and [ReactDOMStringRenderer.js](https://github.com/facebook/react/blob/fc807720780962729df2ad1b0f53258603c020c7/packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js) in react-dom/server package now accept optional third parameter with [trusted types polyfill implementation](https://github.com/WICG/trusted-types). If TT are provided, they are enforced, **otherwise no behavioral change is made**. If TT are enforced, we check whether the values are trusted before creating the markup from them and throw an error otherwise.   

## Reference
* [Add trusted types to react on client side](https://github.com/facebook/react/pull/16157)


gaearon:
Apologies for closing. We've accidentally pushed a master branch to the repo (due to an old fork), then deleted it, and this lead to a bunch of PRs getting auto-closed. GitHub seems confused and unfortunately does not offer a way to reopen this against `main`. I can't promise when we'll get to reviewing, but if there is still interest in this PR, we'd appreciate a resubmit.

PR #16561, opened 2019-08-23
chore(ReactFiberWorkLoop): rephrase comment

jamesgeorge007:
Minor tweak.


trueadm:
This change isn't really correct either, as no error is going to be "received" as such. Thanks for taking the time anyway :)

PR #16575, opened 2019-08-26
Change Babel spread transform from loose -> strict

bvaughn:
The spread syntax should work for typed arrays, e.g.
```js

```

While importing DevTools, I noticed that some of the hydration tests were failing because the spread operator was returning unexpected values.
```js
[...new Uint8Array([1,2,3])]

// rather than this [1, 2, 3]
// it was returning this [[1, 2, 3]]
```

This seems to be because of the way we've configured the `@babel/plugin-transform-spread` "loose" mode. Docs say:
> In loose mode, all iterables are assumed to be arrays.

In the case of DevTools though, we _know_ we are dealing with iterables that aren't necessarily arrays.

Hopefully we can change this transform without negatively impacting the build? Let's compare the output from CI and see if this does anything unexpected.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -19,7 +19,7 @@ module.exports = {
     '@babel/plugin-transform-shorthand-properties',
     '@babel/plugin-transform-computed-properties',
     '@babel/plugin-transform-for-of',
-    ['@babel/plugin-transform-spread', {loose: true, useBuiltIns: true}],
+    ['@babel/plugin-transform-spread', {loose: false, useBuiltIns: true}],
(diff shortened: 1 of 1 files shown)


gaearon:
If I recall correctly, the non-loose version is the one that embeds an iterator-to-array helper at every spread callsite. It’s somewhat bloated so we should probably seek another solution anyway rather than enable it for React builds. If this was a hard requirement, it would be better to forbid the syntax completely for React bundles than have a chance of accidentally embedding a bloated helper.

PR #16578, opened 2019-08-27
prevent firefox marking required textareas invalid

### LINKED ISSUE #16402: textarea with `required` attribute renders in invalid state in FF
*Bug*

When a `textarea` has the `required` attribute it renders in the invalid state using FireFox. 

This behavior is visible here: https://codesandbox.io/s/rough-frog-f00ow. Remember to view it in FireFox.  

I expect the required field to validate on form submit and not before. 

I have tested this with FireFox 68 and 69 and React 16.8.6

This appears appears to be a re-occurrence of: https://github.com/facebook/react/issues/8395

halvves:
Bug was caused by an IE10/IE11 bugfix dealing with the placeholder attribute and textContent. Solved by avoiding the IE bugfix when textContent was empty.

Closes #16402 

## Test Plan

Open up DOM test fixtures and compare (in Firefox)...

local: http://screeching-degree.surge.sh/textareas
to latest: http://screeching-degree.surge.sh/textareas?version=16.9.0

...and verify that textareas are no longer marked invalid on load when required.

Also confirm that the original IE bugfix is still functioning as intended.

### DIFF
--- a/fixtures/dom/src/components/fixtures/textareas/index.js
+++ b/fixtures/dom/src/components/fixtures/textareas/index.js
@@ -54,7 +54,8 @@ export default class TextAreaFixtures extends React.Component {
             <b>
               <i>not</i>
             </b>{' '}
-            see a red aura, indicating the textarea is invalid.
+            see a red aura on initial page load, indicating the textarea is
+            invalid.
             <br />
             This aura looks roughly like:
             <textarea style={{boxShadow: '0 0 1px 1px red', marginLeft: 8}} />
--- a/fixtures/dom/src/components/fixtures/textareas/index.js
+++ b/fixtures/dom/src/components/fixtures/textareas/index.js
@@ -1,3 +1,4 @@
+import Fixture from '../../Fixture';
 import FixtureSet from '../../FixtureSet';
 import TestCase from '../../TestCase';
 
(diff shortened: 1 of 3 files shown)


fixtures/dom/src/components/fixtures/textareas/index.js
```diff
@@ -39,6 +40,43 @@ export default class TextAreaFixtures extends React.Component {
...
+          <TestCase.Steps>
+            <li>View this test in Firefox</li>
+          </TestCase.Steps>
+
+          <TestCase.ExpectedResult>
+            You should{' '}
+            <b>
+              <i>not</i>
+            </b>{' '}
+            see a red aura, indicating the textarea is invalid.
```
philipp-spiess: Can you explain this a bit better? I _do_ see a red outline specifically when I focus the "Empty value prop string" textarea, hit any character on the keyboard, and then move the focus to one of the other textareas. Is this expected?

packages/react-dom/src/client/ReactDOMTextarea.js
```diff
@@ -157,7 +157,9 @@ export function postMountWrapper(element: Element, props: Object) {
   // will populate textContent as well.
   // https://developer.microsoft.com/microsoft-edge/platform/issues/101525/
   if (textContent === node._wrapperState.initialValue) {
-    node.value = textContent;
+    if (textContent) {
```
philipp-spiess: (As a note: It seems like `Node.textContent` has to be either a string or `null`: https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#Return_value so the loose check seems OK, maybe we still want to have an explicit `textContent !== '' && textContent !== null` though)

fixtures/dom/src/components/fixtures/textareas/index.js
philipp-spiess: Ah that seems reasonable. Maybe update the instructions then to be clearer about this case. 

> I think these are setup this way to avoid needing to scaffold out the controlled input code. Would it be less confusing if they were set up this way?

Not quite sure what you mean by that but I think what we have now is already better than what we had 🙂

packages/react-dom/src/client/ReactDOMTextarea.js
halvves: updated to use more explicit check

fixtures/dom/src/components/fixtures/textareas/index.js
halvves: updated expected result to read "...see a red aura **_on initial page load_**, indicating the textarea is invalid..."

do you think this is sufficient clarification?

nhunzaker:
@halvves Thanks for being patient with us. This is good by me. Let's ship it.

PR #16587, opened 2019-08-27
[react-core] Add experimental React Scope component API

trueadm:
This PR adds an experimental `React.unstable_createScope` API for internal testing and usage only. This is to be used for testing with projects around handling focus management within React. In order to keep things small and sane for reivew this PR only has ReactDOM (client) support for now (other renderer support to follow).


trueadm:
@sebmarkbage The usage of the symbol in switch statements.

PR #16591, opened 2019-08-28
[Partial Hydration] Don't invoke listeners on parent of dehydrated event target

sebmarkbage:
If an event gets invoked on a child of a suspense boundary we haven't hydrated yet, that's an opportunity where we might want to consider replaying it instead. In the existing semantics there exist no replaying, instead things are ignored (see #16532). So this PR makes sure we don't dispatch these events to React's event system. They may still have been invoked on non-React nodes.

However the tricky part is that we don't readily know if a node is a non-React DOM node or if it's just a node we haven't gotten to yet. We can't just mark the node as dehydrated since the server streaming can update the content of dehydrated boundaries as they go and that would lose the markers.

However, typically there will at least be some React DOM node that is a parent of the Suspense boundary so that will normally become the target today. We also need to deal with the same case when there is a Suspense boundary at the root and while the root most level is concurrently hydrating.

In the case where there is a parent React DOM node, we don't know if the target node was a non-React DOM node that someone manually inserted or if it is a child of a dehydrated boundary.

The common case is that if it's a DOM node that someone messes with, it won't have any children so we can use that as a quick bailout to assume it's not a React node.

If it does have children, I backtrack on the previous siblings to see if we're nested inside a Suspense boundary (i.e. if we're inside two comment nodes). This could potentially be expensive if there are many previous siblings but most of the time there's only one direct DOM node inside a Suspense boundary, and it's unusual that it wouldn't be a Suspense boundary in this case. The worst case is that this is happening on a non-React node and that the React parent happens to have a child that renders null or something in it, and also that this has many children in it.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponentTree.js
+++ b/packages/react-dom/src/client/ReactDOMComponentTree.js
@@ -15,9 +15,14 @@ const randomKey = Math.random()
   .slice(2);
 const internalInstanceKey = '__reactInternalInstance$' + randomKey;
 const internalEventHandlersKey = '__reactEventHandlers$' + randomKey;
+const internalContainerInstanceKey = '__reactContainere$' + randomKey;
 
 export function precacheFiberNode(hostInst, node) {
   node[internalInstanceKey] = hostInst;
+}
+
+export function markContainerAsRoot(hostRoot, node) {
+  node[internalContainerInstanceKey] = hostRoot;
 }
 
 /**
@@ -70,6 +75,13 @@ export function getClosestInstanceFromNode(targetNode) {
           // check.
         }
       }
+      return targetInst;
+    }
+    // Next we'll check if this is a container root that could include
+    // React nodes in the future.
+    targetInst = parentNode[internalContainerInstanceKey];
+    if (targetInst) {
+      // If so, we return the HostRoot Fiber.
       return targetInst;
     }
     targetNode = parentNode;
--- a/packages/react-dom/src/client/ReactDOMComponentTree.js
+++ b/packages/react-dom/src/client/ReactDOMComponentTree.js
@@ -8,6 +8,8 @@
 import {HostComponent, HostText} from 'shared/ReactWorkTags';
 import invariant from 'shared/invariant';
 
+import {getParentSuspenseInstance} from './ReactDOMHostConfig';
+
 const randomKey = Math.random()
   .toString(36)
   .slice(2);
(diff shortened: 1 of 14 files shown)


trueadm:
If I'm understanding this right, this will re-target an incoming event to be that of a hydrated/active fiber and its nearest DOM node. If this is the case, is this really the behaviour we want? Shouldn't we proceed with building in event capturing and replaying?

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -141,8 +149,15 @@ function handleTopLevel(bookKeeping: BookKeepingInstance) {
...
-    bookKeeping.ancestors.push(ancestor);
-    ancestor = getClosestInstanceFromNode(root);
+    if (ancestor.tag === HostComponent || ancestor.tag === HostText) {
+      bookKeeping.ancestors.push(ancestor);
+    }
+    let parent = root.parentNode;
+    if (parent) {
+      ancestor = getClosestInstanceFromNode(parent);
+    } else {
+      ancestor = null;
```
trueadm: couldn't this just be `break`?

sebmarkbage:
I updated the root approach to track a separate expando on the container. That way we can differentiate between a DOM node in its role as a Container and its role as a leaf Instance.

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -141,7 +149,9 @@ function handleTopLevel(bookKeeping: BookKeepingInstance) {
     if (!root) {
       break;
     }
-    bookKeeping.ancestors.push(ancestor);
+    if (ancestor.tag === HostComponent || ancestor.tag === HostText) {
```
trueadm: Maybe create a binding for tag here rather than doing the property lookup twice?

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -311,16 +321,33 @@ export function dispatchEvent(
...
-    // event on a non-React tree). We might also consider queueing events and
-    // dispatching them after the mount.
-    targetInst = null;
+  if (targetInst !== null) {
+    let nearestMounted = getNearestMountedFiber(targetInst);
+    if (nearestMounted === null) {
+      // This tree has been unmounted already.
+      targetInst = null;
+    } else {
+      if (nearestMounted.tag === SuspenseComponent) {
```
trueadm: You could also cache the tag as a binding here too.

packages/react-dom/src/client/ReactDOMComponentTree.js
```diff
@@ -8,43 +8,84 @@
...
 const internalInstanceKey = '__reactInternalInstance$' + randomKey;
 const internalEventHandlersKey = '__reactEventHandlers$' + randomKey;
+const internalContainerInstanceKey = '__reactContainere$' + randomKey;
 
 export function precacheFiberNode(hostInst, node) {
   node[internalInstanceKey] = hostInst;
 }
 
+export function markContainerAsRoot(hostRoot, node) {
+  node[internalContainerInstanceKey] = hostRoot;
```
trueadm: I do wonder if its still the best strategy to keep adding additional properties to DOM nodes vs a WeakMap with an object that contains `{ instance, container, props }`. We could probably unify this with the already existing WeakMap we have on DOM nodes for checking for registered events: 

https://github.com/facebook/react/blob/master/packages/react-dom/src/events/ReactBrowserEventEmitter.js#L89

This doesn't need to be done in this PR, just an observation.

gaearon:
If I recall correctly we need this expando to fix iOS clicks in portals anyway.

packages/react-dom/src/client/ReactDOMComponentTree.js
sebmarkbage: The one for props is an unfortunate one. That was never intentional. I just realized too late I didn't have a convenient place to find the current Fiber. I don't remember exactly why I couldn't just update the fiber pointer to the correct alternate every time events change and then read from memoizedProps. There's probably something we can do to get rid of that one.

Regarding the new container one. It's very rare that something will both be a container and an instance so there will almost never be more than one expando. So for that case it's probably not worth adding an indirection for the very common case just to avoid two slots in the rare case.

sebmarkbage:
@gaearon Note that this doesn't actually add one for portals. Only roots. Partly because Portals don't support hydration anyway.

sebmarkbage:
I added a note that calling `getClosestInstanceFromNode` on a Container DOM node (or Suspense comment) doesn't actually return the HostRoot (nor SuspenseComponent). That's because conceptually the Fiber doesn't actually represent the Container DOM node. It represents a virtual thing that exists as the children of the Container DOM node. Similar to a Fragment.

Therefore to get to it, you have to pass a child of the DOM Container node.

This isn't really all that important nuance but it does allow me to take some short cuts (e.g. for the common case where the target is an instance) and gives me a convenient hint that a node is indeed dehydrated.

sebmarkbage:
I'll also note for future us, that this has a known problem when the Container is a Comment node. However, that is an unstable feature and currently doesn't work with hydration anyway. So that needs some further consideration (if we ever want to support that use case).

PR #16602, opened 2019-08-28
chore: upgrade to danger 9

SimenB:
#15778 is broken since Danger@3 loads babel 7 and only works with babel 6. Upgrading  danger fixes it, so here's a separate PR with it. 🤞 it passes

PR #16610, opened 2019-08-29
chore: Update nvmrc with latest lts

eps1lon:
On `master`:
```bash
$ nvm use
Now using node v8.4.0 (npm v5.3.0)
$ yarn
error gitlab@10.2.1: The engine "node" is incompatible with this module. Expected version ">=10.0.0". Got "8.4.0"
error Found incompatible module.
```

Updated with latest node lts release.


threepointone:
This seems reasonable, but could you help me understand what the gitlab connection is? Is it some deeper dependency we have? I didn't find it in `yarn.lock`, so I'm confused.

eps1lon:
TL;DR: It's a dependency of `danger` which was only recently bumped to 9. It probably got introduced during that PR

```bash
$ yarn why gitlab
yarn why v1.17.3
[1/4] Why do we have the module "gitlab"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "gitlab@10.2.1"
info Reasons this module exists
   - "_project_#danger" depends on it
   - Hoisted from "_project_#danger#gitlab"
info Disk size without dependencies: "812KB"
info Disk size with unique dependencies: "1.15MB"
info Disk size with transitive dependencies: "2.69MB"
info Number of shared dependencies: 18
Done in 1.73s.
```

eps1lon:
Yeah 2hr old merge: https://github.com/facebook/react/pull/16602/files#diff-8ee2343978836a779dc9f8d6b794c3b2R5834

bvaughn:
I don't have any objections to updating the default Node. I usually use v11 locally.

If our main concern is the warning being printed to the console, we could also use the `--ignore-engines` flag to suppress it. But we might as well just update from my POV.

threepointone:
Alright, let's get this in. Thanks!

(We also agree to blame @bvaughn if it breaks stuff. Don't tell him.)

eps1lon:
Just for posterity: node 8 reaches end of life in december, 10 is active LTS and 11 is no longer maintained (odd version numbers have a short lifespan, even numbers are LTS candidates). [Full node release cycle](https://github.com/nodejs/Release)

Issue #16527, opened 2019-08-21
DevTools: Profiler: Save profile does nothing on Firefox


deadmau6:
Hi, long time user, first time issuer.

I think I found a bug with the Profiler component. When I click on the `Save profile...` button at the top nothing happens and there appears to be no new files in my Downloads folder. I think it might be silently failing and that's why I am not getting any response.  Has anyone else ran into this issue?

Another thing I would like to mention is that I haven't looked through all of the documentation material about the new Profiler. I only read [this introductory blog post](https://reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html), so if this issue is addressed elsewhere I apologize and also request the source to that solution.

Thank you for your time!

**Versions:**
 * `React DevTools 4.0.5`
 * `Firefox: 68.0.2(64-bit)`
 * `Ubuntu: 18.04`

hristo-kanchev:
It seems that this is happening because of the HTMLElement generation in the `downloadFile` util function.

I'll investigate this further.

hristo-kanchev:
I fixed the issue. I'll open a PR later today :)

PR #16612, opened 2019-08-29
[DevTools] [Profiler]: Save profile now working in Firefox

hristo-kanchev:
Fixes: #16527 

**Description**:
Basically the reason for FF to block it with the previous implementation was that we didn't have a physical DOMNode for the download anchor tag.

With this fix we now pass a `HTMLAnchorElement` to the `downloadFile` function in order to set the correct attributes.

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/ProfilingImportExportButtons.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/ProfilingImportExportButtons.js
@@ -29,6 +29,7 @@ export default function ProfilingImportExportButtons() {
   const {profilerStore} = store;
 
   const inputRef = useRef<HTMLInputElement | null>(null);
+  const downloadRef = useRef<HTMLAnchorElement | null>(null);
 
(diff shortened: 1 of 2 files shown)


packages/react-devtools-shared/src/devtools/views/Profiler/ProfilingImportExportButtons.js
```diff
@@ -114,6 +116,7 @@ export default function ProfilingImportExportButtons() {
         onChange={handleFiles}
         tabIndex={-1}
       />
+      <a ref={downloadRef} className={styles.Input} />
```
bvaughn: nit: At this point we might consider renaming the `Input` class to e.g. `Hidden` or something but this is minor, given the small scope.

bvaughn:
Thank you for the fix! Just verified.

packages/react-devtools-shared/src/devtools/views/Profiler/ProfilingImportExportButtons.js
hristo-kanchev: Good shout. We can fix this in another PR. I'll reference this comment.

PR #16614, opened 2019-08-30
[Hydration Bugfix] Updates to dehydrated content when `disableSchedulerTimeoutBasedOnReactExpirationTime` is enabled

acdlite:
## The Bug

When server rendered content that hasn't finished hydrating yet ("dehydrated" content) receives an update (via props or context), React has a mechanism to force the content to hydrate before applying the update. It does this by increasing the priority of the hydration task from Idle to a level slightly higher than the current render. React will abort the current render, perform the hydration, then try the update again on top of the now-fully-hydrated content.

There are unit tests that cover this case. The bug starts happening when `disableSchedulerTimeoutBasedOnReactExpirationTime` is enabled. It turns out that the mechanism to interrupt the current rendering task depends on the hydration task having a slightly earlier timeout, because Scheduler tasks are sorted by their timeouts. When the hydration task is higher priority, it causes `shouldYield` to flip to true, forcing the render to yield execution and allowing the hydration task to start. (This is similar to how input events can interrupt normal priority renders.)

`disableSchedulerTimeoutBasedOnReactExpirationTime` breaks this mechanism, because when it is enabled, the timeout given to Scheduler is no longer based on React's internal expiration times. Effectively, all rendering tasks within the same priority category are first-in-first-out. So, the hydration task comes *after* the original task in the Scheduler queue, and therefore `shouldYield` will keep returning `false`, and the original task will run to completion. (See https://github.com/facebook/react/pull/16284 for more information on `disableSchedulerTimeoutBasedOnReactExpirationTime`.)

The first commit in this PR adds a regression test for this case.

## The Fix

There are several potential fixes. The one I've chosen is not ideal in the long term, but it's lower risk compared to the complete solution, which will likely require some refactoring of how rendering tasks are scheduled.

The work loop already has some logic to cancel a rendering task in favor of a higher priority one, using `Scheduler.cancelCallback`. It does this by comparing the React expiration times of each task, so it doesn't depend on the ordering of tasks in Scheduler. This works when the high priority task is received during an input event.

However, `Scheduler.cancelCallback` is currently a no-op when given an already-running task. It does not cause the task to stop execution, and if the task does yield with a continuation, then the continuation will run. Which means it won't work if React is already inside the render phase. (Note the distinction between "inside the render phase" versus "in an event that fires in between two chunks of a time sliced task.")

The fix in the second commit addresses both parts: canceling the current task causes `shouldYield` to return `true`, and if the canceled task returns a continuation, the continuation is ignored.

This is sufficient to fix the regression.

## Alternative Fixes

A proper fix would be to model interruptions of in-progress renders in such a way that it does not depend on Scheduler's semantics for canceling and yielding. However, because of the inherent risk involved in changing how rendering tasks are scheduled, I would prefer to land this smaller fix first before attempting a refactor.

(There's already a planned mini-refactor of the work loop, e.g. to optimize how pings and restarts are modeled. We can fold this into that larger change.)

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
@@ -497,6 +497,103 @@ describe('ReactDOMServerPartialHydration', () => {
     expect(span.className).toBe('hi');
   });
 
+  it('regression test: an update to dehydrated content forces a restart to hydrate it first', async () => {
+    // Replicates a bug where React would fail to restart and hydrate when
+    // `disableSchedulerTimeoutBasedOnReactExpirationTime` is true.
+    jest.resetModuleRegistry();
+
+    ReactFeatureFlags = require('shared/ReactFeatureFlags');
+    ReactFeatureFlags.enableSuspenseServerRenderer = true;
+    ReactFeatureFlags.enableSuspenseCallback = true;
+    ReactFeatureFlags.enableFlareAPI = true;
+    ReactFeatureFlags.debugRenderPhaseSideEffects = false;
+    ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
+    // NOTE: Once this feature flag is removed, the preceding test is likely
+    // sufficient to cover this regression case; however, it might be worth
+    // keeping this test around anyway as an extra precaution.
+    ReactFeatureFlags.disableSchedulerTimeoutBasedOnReactExpirationTime = true;
+
+    React = require('react');
+    ReactDOM = require('react-dom');
+    act = require('react-dom/test-utils').act;
+    ReactDOMServer = require('react-dom/server');
+    Scheduler = require('scheduler');
+    Suspense = React.Suspense;
+    SuspenseList = React.unstable_SuspenseList;
+
+    function Text({text}) {
+      Scheduler.unstable_yieldValue(
+        text + ' [' + Scheduler.unstable_getCurrentPriorityLevel() + ']',
+      );
+      return text;
+    }
+
+    function Parent({text, className}) {
+      Scheduler.unstable_yieldValue(
+        'Parent [' + Scheduler.unstable_getCurrentPriorityLevel() + ']',
+      );
+      return (
+        <div>
+          <Suspense fallback="Loading...">
+            <Text text="Sibling" />
+            <span className={className}>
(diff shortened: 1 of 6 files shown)


packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
```diff
@@ -497,6 +497,103 @@ describe('ReactDOMServerPartialHydration', () => {
...
 
+  it('regression test: an update to dehydrated content forces a restart to hydrate it first', async () => {
+    // Replicates a bug where React would fail to restart and hydrate when
+    // `disableSchedulerTimeoutBasedOnReactExpirationTime` is true.
+    jest.resetModuleRegistry();
+
+    ReactFeatureFlags = require('shared/ReactFeatureFlags');
+    ReactFeatureFlags.enableSuspenseServerRenderer = true;
+    ReactFeatureFlags.enableSuspenseCallback = true;
+    ReactFeatureFlags.enableFlareAPI = true;
```
trueadm: Was this intentional?

packages/scheduler/src/__tests__/Scheduler-test.js
```diff
@@ -288,6 +288,26 @@ describe('Scheduler', () => {
     expect(Scheduler).toFlushWithoutYielding();
   });
 
+  it('cancelling the currently running task', () => {
+    const task = scheduleCallback(NormalPriority, () => {
```
trueadm: Are there any other priorities we need to test cancelling on other than normal? Should there be a test to confirm that discrete priority events do not get cancelled?

acdlite: The priority shouldn't be relevant, but I'll add the same test at user-blocking priority as an extra precaution.

packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
acdlite: Probably not necessary, I just copy pasted from the top of the file for consistency: https://github.com/facebook/react/blob/83102b6e51af196dff7bd2c16e9530eca58f4bb0/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js#L23-L36

acdlite:
Superseded by https://github.com/facebook/react/pull/16771

PR #16621, opened 2019-08-30
[react-core] Add more support for experimental React Scope API

trueadm:
The React Scope API is an experimental internal API only. This PR enables the flag for internal www usage and also adds support for ReactDOMServer and ReactTestRenderer.


trueadm:
Note: the size increases are not to OSS ReactDOM, the sizebot's reporting summary is a bit skewed here. There is no size difference on the OSS ReactDOM.

PR #16623, opened 2019-08-30
use-subscription tearing fix

### LINKED ISSUE #16396: use-subscription causes UI tearing in some random cases

Bug

Referring to [this twitter conversation](https://twitter.com/milankinen/status/1161718618633248768?s=20), it seems that `use-subscription` can cause "UI tearing" in some random cases due to a (possible) race condition w.r.t. the combination of `subscribe` and `getCurrentValue` in internal usage.

Here is a minimalistic application demonstrating the behaviour: https://github.com/milankinen/use-subscription-tearing-demo

I'd expect counters in the example application to *always* render same number.

Tested React version: 16.9.0 (haven't tested with older ones)
OS: OSX 10.14.6
Hardware: MacBook Pro (Retina, 15-inch, Mid 2015) 2.8 GHz Intel Core i7 & 16GB RAM
Browser: Chrome Version 76.0.3809.100 (Official Build) (64-bit)

@bvaughn

bvaughn:
Adds a test (previously failing) and a fix for a potential tearing case reported in #16396.

Resolves #16396

### DIFF
--- a/packages/use-subscription/src/__tests__/useSubscription-test.internal.js
+++ b/packages/use-subscription/src/__tests__/useSubscription-test.internal.js
@@ -602,28 +602,22 @@ describe('useSubscription', () => {
       // This update will be eagerly evaluated,
       // so the tearing case this test is guarding against would not happen.
       mutate('B');
-      expect(Scheduler).toFlushAndYieldThrough(['render:first:B']);
-      mutate('C');
-      expect(Scheduler).toFlushAndYield([
-        'render:second:B',
-        'render:first:C',
-        'render:second:C',
-      ]);
+      expect(Scheduler).toFlushAndYield(['render:first:B', 'render:second:B']);
 
       // No more pending updates
       jest.runAllTimers();
 
-      // Partial update "C" -> "D"
-      // Interrupt with a second mutation "D" -> "E".
+      // Partial update "B" -> "C"
+      // Interrupt with a second mutation "C" -> "D".
       // This update will not be eagerly evaluated,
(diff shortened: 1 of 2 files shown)


packages/use-subscription/src/__tests__/useSubscription-test.internal.js
```diff
@@ -560,4 +563,68 @@ describe('useSubscription', () => {
...
+        'render:second:B',
+        'render:first:C',
+        'render:second:C',
+      ]);
+
+      // No more pending updates
+      jest.runAllTimers();
+
+      // Partial update "C" -> "D"
+      // Interrupt with a second mutation "D" -> "E" but this should not cause tearing.
```
bvaughn: This second interrupting mutation is required to trigger the previous failure. ([Described in more detail in this issue comment.](https://github.com/facebook/react/issues/16396#issuecomment-526684284)) I don't yet understand why this is but will be digging in more.

bvaughn: It looks like React eagerly computes the state initially, for the first update. It doesn't do this or subsequent updates though because `expirationTime !== NoWork`.

I'm not sure if this is expected, given that my test flushes all pending work.

bvaughn: @acdlite and I chatted about this offline and I have a better understanding of it now. I've updated the inline test comments. Going to hand this PR to him since he has full context now too.

bvaughn:
Ping @acdlite :)

PR #13080, opened 2018-06-20
Convert react to named exports

TrySound:
In this diff I'm making a step forward to esm support.

Here I export scoped variables instead of object. Conditions are
resolved with with ternary operator.

To solve the problem of using both `import * as React from 'react'`
and `import React from 'react'` I removed `__esModule` field with simple
regexp. In the future rollup release this will be achived with
[new rollup option](https://github.com/rollup/rollup/pull/2287).

New bundles are tested with webpack, rollup and babel-jest. All ways
produce same results.

There was a problem with reexports in https://github.com/facebook/react/issues/11526
and as you can see in examples below it's solved so we can migrate
react-dom to named exports aswell in the next PR.

before

```js
var React = {
  Children: {
    map: mapChildren,
    forEach: forEachChildren,
    count: countChildren,
    toArray: toArray,
    only: onlyChild
  },
  createRef: createRef,
  Component: Component,
  PureComponent: PureComponent,
  createContext: createContext,
  forwardRef: forwardRef,
  Fragment: REACT_FRAGMENT_TYPE,
  StrictMode: REACT_STRICT_MODE_TYPE,
  unstable_AsyncMode: REACT_ASYNC_MODE_TYPE,
  unstable_Profiler: REACT_PROFILER_TYPE,
  createElement: createElementWithValidation,
  cloneElement: cloneElementWithValidation,
  createFactory: createFactoryWithValidation,
  isValidElement: isValidElement,
  version: ReactVersion,
  __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: {
    ReactCurrentOwner: ReactCurrentOwner,
    // Used by renderers to avoid bundling object-assign twice in UMD bundles:
    assign: _assign
  }
};
if (enableSuspense) {
  React.Timeout = REACT_TIMEOUT_TYPE;
}
{
  _assign(React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED, {
    // These should not be included in production.
    ReactDebugCurrentFrame: ReactDebugCurrentFrame,
    // Shim for React DOM 16.0.0 which still destructured (but not used) this.
    // TODO: remove in React 17.0.
    ReactComponentTreeHook: {}
  });
}
var React$2 = Object.freeze({
	default: React
});
var React$3 = ( React$2 && React ) || React$2;
// TODO: decide on the top-level export form.
// This is hacky but makes it work with both Rollup and Jest.
[...]

PR #16626, opened 2019-08-30
Add Auto Import to Babel Plugin 

lunaruan:
This babel transform is a fork of the @babel/plugin-transform-react-jsx transform and is for experimentation purposes only. We don't plan to own this code in the future, and we will upstream this to Babel at some point once we've proven out the concept.

As per the [RFC](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md) to simplify element creation, we want to add the ability to auto import "react' directly from the babel plugin. This commit updates the babel plugin with **two options**:

1.) `importSource`: The React module to import from. Defaults to `react`.
2.) `shouldCacheImportFns`: If this is set to `true`, React functions that are imported will be cached at the top of the file (ex. `var _jsx = _react.jsx`) instead of calling the member expression (ex. `_react.jsx`) directly in the code. Defaults to `false`.
3.) `autoImport`: The type of import. Defaults to `none`.
      - `none`: Does not import React. JSX compiles to `React.jsx` etc.
      - `namespace`: `import * as _react from "react";`. JSX compiles to `_react.jsx` etc.
      - `default`: `import _default from "react";` JSX compiles to `_default.jsx` etc.
      - `namedExports`: `import {jsx as _jsx} from "react";` JSX compiles to `_jsx` etc.
      - `require`: `var _react = _interopRequireWildcard(require("react"));`. jSX compiles to `_react.jsx` etc.

`namespace`, `default`, and `namedExports` can only be used when `sourceType: module` and `require ` can only be used when `sourceType: script`.

We plan to test #16432 before landing this diff.


packages/babel-plugin-react-jsx/__tests__/TransformJSXToReactJSX-test.js
```diff
@@ -29,15 +30,307 @@ function transform(input, options) {
...
+    let _error;
+    const code = `var x = <div><span /></div>`;
+    try {
+      transform(code, {
+        autoImport: 'require',
+      });
+    } catch (error) {
+      _error = error;
+    }
+    expect(_error).toEqual(
```
threepointone: Did you test on the error here in this manner because expect(...).toThrow(...) was not sufficient to test the shape of this error?

packages/babel-plugin-react-jsx/__tests__/TransformJSXToReactJSX-test.js
```diff
@@ -29,15 +30,307 @@ function transform(input, options) {
...
+      new SyntaxError(
+        'undefined: Babel `sourceType` must be set to `script` for autoImport ' +
```
threepointone: Why does the error start with 'undefined'?

packages/babel-plugin-react-jsx/__tests__/TransformJSXToReactJSX-test.js
```diff
@@ -29,15 +30,307 @@ function transform(input, options) {
...
+    );
+  });
+
+  it('auto import can specify source', () => {
+    expect(
+      transform(`var x = <div><span /></div>`, {
+        autoImport: 'namespace',
+        importSource: 'foobar',
+      })
+    ).toMatchSnapshot();
```
threepointone: I wish we could use jest 24 so this could be `toMatchInlineSnapshot()` :( I'll try to fix it this week.

packages/babel-plugin-react-jsx/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -24,6 +24,42 @@
 'use strict';
 
 const esutils = require('esutils');
+const helperModuleImports = require('@babel/helper-module-imports');
+const isModule = helperModuleImports.isModule;
+const addNamespace = helperModuleImports.addNamespace;
+const addNamed = helperModuleImports.addNamed;
+const addDefault = helperModuleImports.addDefault;
```
threepointone:
```suggestion
const {isModule, addNamespace, addNamed, addDefault} = require('@babel/helper-module-imports');
```
I love destructuring imports :)

packages/babel-plugin-react-jsx/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -24,6 +24,42 @@
...
+  default: 'default', // import _default from "react";
+  namedExports: 'namedExports', // import { jsx } from "react";
+};
+
+// We want to use React.createElement, even in the case of
+// jsx, for <div {...props} key={key} /> to distinguish it
+// from <div key={key} {...props} />. This is an intermediary
+// step while we deprecate key spread from props. Afterwards,
+// we will remove createElement entirely
+function shouldUseCreateElement(path, t) {
```
threepointone:
```suggestion
function shouldUseCreateElement(path, types) {
```
I wasn't sure what `t` meant here without looking at it's callsite.

packages/babel-plugin-react-jsx/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -576,25 +614,108 @@ module.exports = function(babel) {
...
+
+    if (IMPORT_TYPES[state.autoImport] === undefined) {
+      throw path.buildCodeFrameError(
+        'autoImport must be one of the following: ' +
+          Object.keys(IMPORT_TYPES).join(', '),
+      );
+    }
+    if (state.useCreateElement) {
+      throw path.buildCodeFrameError(
+        'auto importing cannot be used with createElement. Consider setting ' +
```
threepointone:
```suggestion
        'autoImport cannot be used with `createElement()`. Consider setting ' +
```
Alternately, "Automatically importing..."?

packages/babel-plugin-react-jsx/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -576,25 +614,108 @@ module.exports = function(babel) {
...
+          '`useCreateElement` to false to use the new jsx function instead',
```
threepointone:
```suggestion
          '`useCreateElement` to `false` to use the `jsx()` instead.',
```

packages/babel-plugin-react-jsx/src/TransformJSXToReactBabelPlugin.js
lunaruan: Oops I meant to change this but I forgot. Good catch!

packages/babel-plugin-react-jsx/__tests__/TransformJSXToReactJSX-test.js
lunaruan: it looks like it's the filename if there is a `filename` associated with the code. `path.buildCodeFrameError` automatically adds this. Will remove from the test though

sebmarkbage:
Why does it prepend with `_interopRequireWildcard`? We don't want that since it's unnecessary extra code, and we probably don't support it in all our own environments and we know for sure what our own module exports so we don't need to do that at runtime for every module.

Jessidhia:
You'll either get an `_interopRequireWildcard` or an `_interopRequireDefault`, depending on whether you want to pretend `'react'` is an actual Module Namespace Object or whether it is a default-exported object.

According to the current semantics of commonjs interop, both with real ES modules and with babelmodules, `'react'` is itself commonjs-adapted-to-default-exported-object as it doesn't have a truthy `__esModule` key. babelmodules will additionally copy the keys from the default export to the simulated module namespace object if you import it as a namespace.

You'll probably have to deal with https://github.com/facebook/react/pull/13080 first if you want to reach any kind of consistency here; otherwise I recommend treating React as a default import instead of a namespace import.

sebmarkbage:
@Jessidhia Why can’t we just emit a require call without the interop since we’re a CommonJS module in this mode?

Jessidhia:
It's possible by giving options to the `helper-module-imports` calls.

https://github.com/babel/babel/blob/v7.5.5/packages/babel-helper-module-imports/src/import-injector.js#L21-L62

I'm not 100% sure this is public API as it's only documented here, but it could as well just be an omission from the docs since the jsdoc is pretty comprehensive.

packages/babel-plugin-react-jsx/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -24,6 +24,47 @@
...
+  addDefault,
+} = require('@babel/helper-module-imports');
+
+// These are all the valid auto import types (under the config autoImport)
+// that a user can specific
+const IMPORT_TYPES = {
+  none: 'none', // default option. Will not import anything
+  require: 'require', // var _react = require("react");
+  namespace: 'namespace', // import * as _react from "react";
+  default: 'default', // import _default from "react";
```
sebmarkbage: I think we should probably rename this to `defaultExport` to avoid confusion since this isn't the "default" (as in "unless specified") mode.

packages/babel-plugin-react-jsx/__tests__/__snapshots__/TransformJSXToReactJSX-test.js.snap
```diff
@@ -38,6 +38,167 @@ var div = React.jsx(Component, Object.assign({}, props, {
...
+      key: "4"
+    }))]
+  })
+});
+`;
+
+exports[`transform react to jsx auto import namespace 1`] = `
+import * as _react from "react";
+
+var x = _react.jsx(_react.Fragment, {
```
sebmarkbage: We could add an additional option to emit a variable to cache these at the top of the file.

```
import * as _react from "react";
const _fragment = _react.Fragment;
```
...
```
var x = _jsx(_Fragment, {
```

For a native ES module system this wouldn't matter but the way most module transpilers are implemented this becomes an object. Which may need an indirection to load the object and then the current function in the object for each call. Instead of just loading the current function.

Note that this also applies to `namedExports`. It should be something like:

```
import {jsx as __jsx, Fragment as __Fragment} from "react";
const _jsx = __jsx;
const _fragment = __Fragment;
```

Modern browser VMs may optimize for the inline cache containing this information perhaps. So I don't think this matters but it might for Hermes.

Additionally this approach would slow down initialization of the module scope and use more memory.

So I doubt this approach will actually be better overall but it would be nice to have an option so we can experiment with it and test which one is faster in practice.

rickyvetter: Ditto for shadowing question and tests on these variables if we add this!

packages/babel-plugin-react-jsx/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -576,25 +593,305 @@ module.exports = function(babel) {
...
+      if (state.shouldCacheImportFns) {
+        Object.keys(importMap).forEach(importName => {
+          const importPath = getNamedExportImportPath(
+            path,
+            importMap[importName],
+            importName,
+            state.source,
+          );
+
+          const newName = path.scope.generateUidIdentifier(importName);
```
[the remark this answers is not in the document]
lunaruan: `generateUidIdentifier` would generate an identifier that doesn't collide with any locally defined variables. So in that case it generates: 

```
import * as _react3 from "react";
var _jsx = _react3.jsx;

const Component = function ({
  thing,
  ..._react
}) {
  if (!thing) {
    var _react2 = "something useless";
    return _jsx("div", {});
  }

  ;
  return _jsx("span", {});
};
```

lunaruan:
This PR has not been abandoned. Still working on adding it to WWW.

packages/babel-plugin-react-jsx/src/TransformJSXToReactBabelPlugin.js
```diff
@@ -623,25 +643,342 @@ module.exports = function(babel) {
...
+
   visitor.Program = {
     enter(path, state) {
+      let autoImport = state.opts.autoImport || 'none';
+      let source = state.opts.importSource || 'react';
+      const CACHE_IMPORT_FNS = state.opts.shouldCacheImportFns || false;
+      const {file} = state;
+
+      if (file.ast.comments) {
+        for (const comment of (file.ast.comments: Array<Object>)) {
```
sebmarkbage: This is not a Flow file. Remove the annotation.
Also, let's just use `for (let i = 0; ...` loop to avoid the lint problem.

packages/babel-plugin-react-jsx/__tests__/__snapshots__/TransformJSXToReactJSX-test.js.snap
sophiebits: I believe I tested this once with React.createElement in V8 and found it didn't make a significant difference (IIRC slightly worse; definitely wasn't better or I would've gone through with it).

sebmarkbage: Yea I tested it as well and came to the same conclusion. So we’re not doing this.

The key is that as long as the VM optimizes for storing methods on the object hidden class it’s guaranteed to be worse once you use more than one method.

What I’m wondering next is if it would be worth while having React be a global property.

Issue #16501, opened 2019-08-20
Getting maximum call stack exceeded on backend.js when rendering many elements.


ryan-rushton:
## Details
Using version 4.0.2 (8/15/2019) I am getting a maximum call stack size exceeded when my app starts up. While my app using this is a little different I have replicated the issue using a fresh create-react-app. I will include the App.js code below.

## Steps to reproduce
1. Create a new app using version 3.0.1 of create-react-app
2. Replace the code in App.js with the code listed below
3. Inspect the console, there should be a stack trace from a maximum call stack exceeded error

## App.js
```
import React from 'react';
import logo from './logo.svg';
import './App.css';

const createDivs = num => {
    const rtn = [];
    for(let i = 0; i < num; i++) {
        const key = `div-${i}`;
        rtn.push(<div className={key} key={key} />)
    }
    return rtn;
}

function App() {
  return (
    <div className="App">
      <header className="App-header">
        <img src={logo} className="App-logo" alt="logo" />
        <p>
          Edit <code>src/App.js</code> and save to reload.
        </p>
        <a
          className="App-link"
          href="https://reactjs.org"
          target="_blank"
          rel="noopener noreferrer"
        >
          Learn React
        </a>
      </header>
      <div>
        {createDivs(15000)}
      </div>
    </div>
  );
}

export default App;
```

## Call Stack
<details>

<summary>backend.js:formatted:2097 Uncaught RangeError: Maximum call stack size exceeded
</summary>

```
    at ge (backend.js:formatted:2097)
    at Ge (backend.js:formatted:2435)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
    at Ge (backend.js:formatted:2478)
[...]

Issue #16491, opened 2019-08-19
DevTools: Too deep trees can cause a stack overflow


bvaughn:
With 10,000 item nesting in `DeeplyNestedComponents` case:

<img width="432" alt="Screen Shot 2019-04-06 at 15 16 02" src="https://user-images.githubusercontent.com/810438/55670638-f3bec480-587e-11e9-850d-8a247aa42f7c.png">

Not super realistic but we could fix this by avoiding recursion in the traversal.

---
Originally reported by @gaearon via https://github.com/bvaughn/react-devtools-experimental/issues/76

bvaughn:
Closing as duplicate of #16501 since that has more repro info.

PR #16627, opened 2019-08-30
Refactored backend renderer to remove most of the recursion

bvaughn:
DevTools doesn't currently handle extremely deep trees (#16491) or extremely wide trees (#16501) very well, due to recursion in the backend interface. This PR removes _most_ of that.

As a sanity test, I confirmed that after this refactor, DevTools was able to handle the following:
```js
const Child = ({ children = null }) => children;

const Deep = () => {
  let children = null;
  for (let i = 0; i < 15000; i++) {
    children = <Child>{children}</Child>;
  }
  return children;
};

const Wide = () => {
  let children = [];
  for (let i = 0; i < 15000; i++) {
    children.push(<Child key={i} />);
  }
  return children;
};
```

Migrated from https://github.com/bvaughn/react-devtools-experimental/pull/385

### DIFF
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -1481,9 +1481,7 @@ export function attach(
 
   // Returns whether closest unfiltered fiber parent needs to reset its child list.
   //
-  // TRICKY
-  // Although this method is recursive, refactoring it to be iterative would add a lot of complexity.
-  // Since it is not expected to operate on huge parts of the tree at once, it's probably okay?
+  // TODO Refactor this method to be iterative as well
   function updateFiberRecursively(
     nextFiber: Fiber,
     prevFiber: Fiber,
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -1227,6 +1227,9 @@ export function attach(
     }
   }
 
+  // TRICKY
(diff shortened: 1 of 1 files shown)


packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -1413,22 +1453,37 @@ export function attach(
...
+        }
       }
     }
   }
 
   // Returns whether closest unfiltered fiber parent needs to reset its child list.
+  //
+  // TRICKY
+  // Although this method is recursive, refactoring it to be iterative would add a lot of complexity.
+  // Since it is not expected to operate on huge parts of the tree at once, it's probably okay?
```
bvaughn: This method needs to be migrated too, but it's not going to be as straight forward as the other two since it depends on its own recursive return value.

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -1413,22 +1453,35 @@ export function attach(
...
+          child = child.sibling;
+        }
       }
     }
   }
 
   // Returns whether closest unfiltered fiber parent needs to reset its child list.
+  //
+  // TODO Refactor this method to be iterative as well
   function updateFiberRecursively(
```
bvaughn: This method needs to be migrated too, but it's not going to be as straight forward as the other two since it depends on its own recursive return value.

bvaughn:
We may want to do this eventually but I'm going to close for now.

PR #16628, opened 2019-08-31
[react-events] Tap responder

necolas:
This is a partial replacement for the 'Press' responder:

1. `useTap` is scoped to pointers (no keyboard support). Our current thinking is
that "responders" should be limited to working with pointers, and that they can
be combined with 'useKeyboard' in user-space. For example, we might create a
'usePress' hook in user-space that combines 'useTap' with 'useKeyboard' to react
to both pointers and keyboard interactions.

2. `useTap` cancels the gesture once the pointer moves over an element that is
not within the responder target's subtree. This differs from `usePress` (and
React Native), where the gesture remains active after the pointer exits the
target's subtree and is restarted once the pointer reenters. One of the
drawbacks with the `usePress` behavior is that it requires repeatedly measuring
DOM elements (which can cause jank) to perform hit region tests. `useTap` avoids
doing this and relies on `document.elementFromPoint` only to support the
TouchEvent fallbacks.

3. `useTap` calls `onTapUpdate` when the active gesture's state changes,
`onTapEnd` when the gesture successfully completes. and `onTapCancel` when it
fails. There is no `onTap` callback. `usePress` did not explicitly report back
when the gesture failed, and product developers were confused about the
difference between `onPress` and `onPressEnd`.

4. `useTap` explicitly separates the PointerEvent implementation from the
MouseEvent/TouchEvent fallback.

5. `useTap` has better unit test coverage . All pointer types and the fallback
environment are tested. The shape of the gesture state object is also defined
and tested.

[Demo](https://codesandbox.io/s/latest-responder-build-rz96j30rp)


packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -158,16 +158,21 @@ const eventResponderContext: ReactDOMResponderContext = {
     validateResponderContext();
     const childFiber = getClosestInstanceFromNode(childTarget);
     const parentFiber = getClosestInstanceFromNode(parentTarget);
-    const parentAlternateFiber = parentFiber.alternate;
 
-    let node = childFiber;
-    while (node !== null) {
-      if (node === parentFiber || node === parentAlternateFiber) {
-        return true;
+    if (childFiber != null && parentFiber != null) {
```
necolas: If this function received a `parentTarget` that was an `Element`, but not an element in React's fiber tree, it would throw an error trying to access `parentFiber.alternate`...

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -158,16 +158,21 @@ const eventResponderContext: ReactDOMResponderContext = {
...
+      while (node !== null) {
+        if (node === parentFiber || node === parentAlternateFiber) {
+          return true;
+        }
+        node = node.return;
       }
-      node = node.return;
+      return false;
+    } else {
+      // Fallback to DOM APIs
```
necolas: ...and you might end up with an unmanaged DOM node here is when the target is not managed by React, e.g., it's the document or any other DOM context that React is rendering into. At the moment some of the responders cancel the gesture if a root level scroll event came from an ancestral element.

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -436,6 +441,7 @@ function processTimers(
 
 function createDOMResponderEvent(
   topLevelType: string,
+  responderTarget: Element,
```
necolas: Any event going to a responder should always have a `responderTarget` set, and it does, but this Flow type means all the responders need to account for a `null` case.

packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -534,9 +533,15 @@ function traverseAndHandleEventResponderInstances(
             const onEvent = responder.onEvent;
             if (onEvent !== null) {
               currentInstance = responderInstance;
-              responderEvent.responderTarget = ((target: any):
```
necolas: The event was being mutated each time. If we want to reuse/freeze a single event object across all the responders perhaps we could move `responderTarget` to a context method like `getResponderTarget`.

packages/react-events/src/dom/shared/index.js
```diff
@@ -0,0 +1,43 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails react-core
+ */
+
+'use strict';
+
+import {DiscreteEvent, UserBlockingEvent} from 'shared/ReactTypes';
+
+export const hasPointerEvents =
```
necolas: At some point we'll need to avoid inlining common internal dependencies in the builds for each responder.

packages/react-events/src/dom/testing-library/domEnvironment.js
```diff
@@ -20,10 +20,18 @@ export function hasPointerEvent() {
 }
 
 export function setPointerEvent(bool) {
-  const mock = bool ? emptyFunction : undefined;
-  global.PointerEvent = mock;
-  global.HTMLElement.prototype.setPointerCapture = mock;
-  global.HTMLElement.prototype.releasePointerCapture = mock;
+  const pointerCaptureFn = name => id => {
+    if (typeof id !== 'number') {
+      console.error(`A pointerId must be passed to "${name}"`);
```
necolas: This makes sure the param type is correct because other the unit tests *not* would fail. And Flow also has a bug where the param type is incorrectly defined as `string`

packages/react-events/src/dom/testing-library/domEventSequences.js
```diff
@@ -153,10 +153,11 @@ export function pointermove(target, payload) {
   }
 }
 
-export function pointerup(target, defaultPayload) {
+export function pointerup(target, defaultPayload = {}) {
   const dispatch = arg => target.dispatchEvent(arg);
   const pointerType = getPointerType(defaultPayload);
-  const payload = {buttons: buttonsType.none, ...defaultPayload};
+  // eslint-disable-next-line no-unused-vars
+  const {buttons, ...payload} = defaultPayload;
```
necolas: This fixes a unit-test-breaking bug in the `pointerup` sequence that produced an incorrect `buttons` value for `mousedown` and produced a false-positive emulated mouse event filtering effect in the `Tap` responder.

packages/shared/ReactDOMTypes.js
```diff
@@ -30,7 +30,7 @@ export type ReactDOMResponderEvent = {
   passiveSupported: boolean,
   pointerId: null | number,
   pointerType: PointerType,
-  responderTarget: null | Element | Document,
+  responderTarget: Element,
```
necolas: All `responderTarget` should be `Element` type, and they are, but the previous Flow type meant all the responders needed to account for the other types.

TrySound:
Using useKeyboard in user space may lead to inconsistent UX. I thought the point of builtin usePress was to avoid opinions here. There was a case with scrolling down on pressing space. Is it considered in new implementation?

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,753 @@
...
+
+function isModifiedTap(event) {
+  const nativeEvent: any = event.nativeEvent;
+  const {altKey, ctrlKey, metaKey, shiftKey} = nativeEvent;
+  return (
+    altKey === true || ctrlKey === true || metaKey === true || shiftKey === true
+  );
+}
+
+function createInitialState() {
```
trueadm: Should have a Flow return type to make sure we don't accidentally have property mismatches.

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,753 @@
...
+  context: ReactDOMResponderContext,
+  state: TapState,
+): void {
+  if (state.rootEvents != null) {
+    context.removeRootEventTypes(state.rootEvents);
+    state.rootEvents = null;
+  }
+}
+
+function shouldActivate(event) {
```
trueadm: Can we add a Flow type for `event` here?

trueadm:
@TrySound We'll still provide `usePress` hook that uses a combination of these responders under-the-hood and we'll recommend people continue to use that hook that the case where it's the desired behavior.

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,615 @@
...
+}
+
+/**
+ * Managing pointers
+ */
+
+function isActivePointer(event, state) {
+  return (
+    state.pointerType === event.pointerType &&
+    state.pointerId === event.pointerId
```
necolas: This check is broken because `event.pointerId` is only populated for PointerEvent

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,615 @@
...
+      case 'mousedown':
+      case 'touchstart': {
+        if (!hasPointerEvents) {
+          if (eventType === 'mousedown' && state.ignoreEmulatedEvents) {
+            return;
+          }
+        }
+
+        if (!state.isActive && shouldActivate(event)) {
+          const eventPointerId = event.pointerId;
```
necolas: This check is broken because `event.pointerId` is only populated for PointerEvent

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,615 @@
...
+          // Make mouse and touch pointers consistent.
+          if (eventPointerId != null) {
```
necolas: Instead check for hasPointerEvents

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,615 @@
...
+            // Flow bug: https://github.com/facebook/flow/issues/8055
+            // $FlowExpectedError
+            eventTarget.releasePointerCapture(eventPointerId);
+          }
+
+          state.isActive = true;
+          state.buttons = nativeEvent.buttons;
+          state.pointerId = eventPointerId;
```
necolas: Probably make a utility to populate this, as each of the 3 event types is different w.r.t. pointer ids

packages/react-events/src/dom/Press.js
```diff
@@ -672,7 +672,10 @@ const pressResponderImpl = {
         if (previousPointerType !== pointerType) {
           return;
         }
-        if (type === 'pointermove' && activePointerId !== pointerId) {
+        if (
+          type === 'pointermove' &&
+          activePointerId !== nativeEvent.pointerId
```
trueadm: Why not extract pointerId above?

packages/react-events/src/dom/testing-library/index.js
```diff
@@ -75,7 +75,7 @@ const createEventTarget = node => ({
   pointerenter(payload) {
     domEventSequences.pointerenter(node, payload);
   },
-  // node dispatches exit & out events
+  // node dispatches exit & keave events
```
trueadm: keave?

scripts/rollup/bundles.js
```diff
@@ -629,6 +629,21 @@ const bundles = [
     global: 'ReactEventsSwipe',
     externals: ['react'],
   },
+
+  {
```
trueadm: make sure you add a `react-events/tap.js` and a `react-events/npm/tap.js`. You will also need to add a line in `react-events/package.json` to include the new file.

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,664 @@
...
+      metaKey: false,
+      pageX: 0,
+      pageY: 0,
+      pointerType: '',
+      pressure: 0,
+      screenX: 0,
+      screenY: 0,
+      shiftKey: false,
+      tangentialPressure: 0,
+      target: document.createElement('div'),
```
trueadm: Why are we creating an element here? Also the `document` might not be the right one, as React works across multiple documents.

necolas: This can just be `null` now that I've given up on not having all these ghost `null` types floating around

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,664 @@
...
+): null | Element | Document {
+  if (hasPointerEvents) {
+    return event.target;
+  } else {
+    if (event.pointerType === 'touch') {
+      const doc = context.getActiveDocument();
+      const nativeEvent: any = event.nativeEvent;
+      const touch = getTouchById(nativeEvent, state.activePointerId);
+      if (touch != null) {
+        return doc.elementFromPoint(touch.clientX, touch.clientY);
```
trueadm: Nice

packages/react-events/src/dom/Press.js
necolas: The value is only used (and not `null`) within the the PointerEvent branches.

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,664 @@
...
+      return (
+        state.pointerType === event.pointerType &&
+        activePointerId === eventPointerId
+      );
+    } else {
+      return true;
+    }
+  } else {
+    if (event.pointerType === 'touch') {
+      const activePointerId = state.activePointerId;
```
trueadm: This is duplicated above, we can probably store it as a `const`.

packages/react-events/src/dom/Tap.js
```diff
@@ -0,0 +1,664 @@
...
+      case 'pointerdown':
+      case 'mousedown':
+      case 'touchstart': {
+        if (hasPointerEvents) {
+          const pointerId = nativeEvent.pointerId;
+          state.activePointerId = pointerId;
+          // Make mouse and touch pointers consistent.
+          // Flow bug: https://github.com/facebook/flow/issues/8055
+          // $FlowExpectedError
+          eventTarget.releasePointerCapture(pointerId);
```
trueadm: Won't calling this throw on Safari/iOS as it's not on the prototype?

necolas: It's wrapped in `if (hasPointerEvents)`

trueadm: We might still need to guard against lack of support, i.e. Firefox Mobile:

https://developer.mozilla.org/en-US/docs/Web/API/Element/releasePointerCapture

necolas: Looks like Firefox for Android doesn't have PointerEvent enabled by default.

PR #16631, opened 2019-09-01
Complete one expiration test case

NE-SmallTown:
Just make it more complete though there is no bug right now

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactExpiration-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactExpiration-test.internal.js
@@ -244,6 +244,16 @@ describe('ReactExpiration', () => {
       '1 [C] [render]',
       '1 [D] [render]',
     ]);
+
+    expect(Scheduler).toFlushAndYieldThrough([
+      '1 [A] [commit]',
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/__tests__/ReactExpiration-test.internal.js
```diff
@@ -244,6 +244,16 @@ describe('ReactExpiration', () => {
...
+
+    expect(Scheduler).toFlushAndYieldThrough([
+      '1 [A] [commit]',
+      '1 [B] [commit]',
+      '1 [C] [commit]',
+      '1 [D] [commit]',
+    ]);
+
+    // begin the update after the first one finished
+    expect(Scheduler).toFlushAndYieldThrough(['2 [A] [render]']);
```
threepointone: wouldn't this continue rendering through B, C, D, etc? or did you just want to add confirmations of the previous commits, and this render starting?

NE-SmallTown: Yea, it will render B, C, D but we don't need to write the redundant through because we know that will happen

PR #16642, opened 2019-09-03
[react-events] Adds preventKeys support to Keyboard responder

trueadm:
This PR adds the functionality to natively `preventDefault` on the Keyboard responder. This is done via a `preventKeys` prop, which takes an array of key strings that should be preventing during the `keydown` event; along with a `defaultPrevented` flag on the callback's event object.

### DIFF
--- a/packages/react-events/src/dom/Keyboard.js
+++ b/packages/react-events/src/dom/Keyboard.js
@@ -42,6 +42,7 @@ type KeyboardEvent = {|
 
 const isArray = Array.isArray;
 const targetEventTypes = ['keydown_active', 'keyup'];
+const modifiers = ['altKey', 'ctrlKey', 'metaKey', 'shiftKey'];
 
 /**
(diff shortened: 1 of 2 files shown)


packages/react-events/src/dom/Keyboard.js
```diff
@@ -173,12 +185,25 @@ const keyboardResponderImpl = {
     context: ReactDOMResponderContext,
     props: KeyboardProps,
   ): void {
-    const {responderTarget, type} = event;
+    const {nativeEvent, responderTarget, type} = event;
 
     if (props.disabled) {
       return;
     }
+    let defaultPrevented = (nativeEvent: any).defaultPrevented === true;
```
necolas: We could define `const nativeEvent: any = event.nativeEvent` at the top of the function body so we don't have to do it in 3 different places in the code below.

trueadm: Addressed

PR #16644, opened 2019-09-03
Add Warning on React.createElement for Key Spread/React.jsx

lunaruan:
As part of the process of simplifying `React.createElement` as per [this RFC](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md) and to make sure the warning against spreading key (which is currently only in `React.jsx`) is received, we want to start issuing a warning, once the Babel plugin is updated, to ask users to:
   1.) stop spreading key (the only time that they would be redirected to `React.createElement` in the new Babel plugin)
   2.) upgrade babel (if they've only upgraded React and not Babel)
   3.) stop calling `React.createElement` directly and calling `React.jsx` instead.


acdlite:
Let's put the warning behind a feature flag?

And also maybe update the description/title so people reading through the issues don't get the wrong idea. Could also link to the RFC.

packages/react/src/ReactElement.js
```diff
@@ -341,6 +342,29 @@ export function createElement(type, config, children) {
...
+          : type;
+      warningWithoutStack(
+        false,
+        '%s: You are most likely getting this warning because ' +
+          'you are spreading an object on an element while passing it a explicit ' +
+          'key in the format `<div {...props} key={keyVal} />`. To fix this, ' +
+          'switch the key and the props spread to `<div key={keyVal} {...props} />`.' +
+          '\n\n' +
+          'If you are explicitly calling React.createElement, please switch to ' +
+          'React.jsx instead.' +
```
sebmarkbage: This is not correct. We don't want people to call `React.jsx` directly. That's an implementation detail for the JSX plugin. It's too easy to for example do stuff like `React.jsxs('div', { children: data.map(i => <div>Look ma, no keys</div>) })`.

If you're manually calling createElement, then that's fine. However, that probably means that we can't do the warning for the transform this way.

packages/react/src/__tests__/ReactElement-test.internal.js
```diff
@@ -0,0 +1,43 @@
...
+    React = require('react');
+    ReactDOM = require('react-dom');
+  });
+
+  it('should warn when createElement is accessed ', () => {
+    const container = document.createElement('div');
+    expect(() =>
+      ReactDOM.render(React.createElement('div'), container),
+    ).toWarnDev(
+      'Warning: div: You are most likely getting this warning because ' +
```
acdlite: Suggestion: Add an extra sentence to the beginning like

> A `key` attribute on a JSX element should come before an object spread

The rest of the message is good.

PR #16654, opened 2019-09-04
[react-events] Various core tweaks for event responder system

trueadm:
This PR introduces a few fixes and core changes that are required for having event listeners/responders on scope components:

- Removes `responderTarget` from the responder event object. Instead `context.getCurrentTarget()` has been introduced that achieves the same purpose, without needing to mutate the event object. Furthermore, this works with scopes, which do not have a direct DOM node (instead the nearest current DOM from the target is picked up).

- Fixes a bug @necolas found and fixed in another PR, which also came up in this with the responder target changes above, so they were needed here too.

- Scope components via `React.unstable_createScope` now accept event responders via the `listeners` prop, like host components can. This makes it easier to compose scopes that need to tight control over handling things like keyboard/focus interactions.

### DIFF
--- a/packages/react-dom/src/events/DOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DOMEventResponderSystem.js
@@ -12,7 +12,7 @@ import {
   PASSIVE_NOT_SUPPORTED,
 } from 'legacy-events/EventSystemFlags';
 import type {AnyNativeEvent} from 'legacy-events/PluginModuleType';
-import {HostComponent} from 'shared/ReactWorkTags';
+import {HostComponent, ScopeComponent} from 'shared/ReactWorkTags';
 import type {EventPriority} from 'shared/ReactTypes';
 import type {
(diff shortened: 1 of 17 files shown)


packages/react-dom/src/events/DOMEventResponderSystem.js
```diff
@@ -158,16 +160,20 @@ const eventResponderContext: ReactDOMResponderContext = {
     validateResponderContext();
     const childFiber = getClosestInstanceFromNode(childTarget);
     const parentFiber = getClosestInstanceFromNode(parentTarget);
-    const parentAlternateFiber = parentFiber.alternate;
 
-    let node = childFiber;
-    while (node !== null) {
-      if (node === parentFiber || node === parentAlternateFiber) {
-        return true;
+    if (childFiber != null && parentFiber != null) {
```
trueadm: This was the fix taken from the other PR.

PR #16656, opened 2019-09-04
Abstract away code duplicates

ch3rn0v:
Apparently there are a couple of mostly similar functions that could be refactored to remove duplicate pieces of code. The commit does that.


aweary:
We typically avoid introducing more abstractions just for the sake of reducing duplication. As you can see from @sizebot's comment, this doesn't have a noticeable affect on code size either.

Thanks though!

PR #16657, opened 2019-09-04
[react-events] Keyboard responder propagation handling

trueadm:
This adds support for conditional Keyboard responder propagation, for where there are nested keyboard responders. To allow a `onKeyDown` or `onKeyUp` event to propagate the callback is required to return `true` to propagate a single responder level. This is done by providing a new method on context called `continuePropagation` – which tells the responder propagation system to propagate a single level after processing the current responder (of the same responder type).

I also had to remove the usage of `invokeGuardedCallbackAndCatchFirstError` when calling the user-land event callback. This was because the existing `invokeGuardedCallbackAndCatchFirstError` doesn't support returning a value from the callback – which is what we need here. Attempting to change the mechanics of the function seemed far too invasive, so we can test this approach out internally to see if we run into any issues (we shouldn't though).

### DIFF
--- a/packages/legacy-events/ReactGenericBatching.js
+++ b/packages/legacy-events/ReactGenericBatching.js
@@ -11,8 +11,6 @@ import {
 } from './ReactControlledComponent';
 import {enableFlareAPI} from 'shared/ReactFeatureFlags';
 
-import {invokeGuardedCallbackAndCatchFirstError} from 'shared/ReactErrorUtils';
-
 // Used as a way to call batchedUpdates when we don't have a reference to
 // the renderer. Such as when we're dispatching events or if third party
 // libraries need to call batchedUpdates. Eventually, this API will go away when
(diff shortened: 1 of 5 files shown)


packages/legacy-events/ReactGenericBatching.js
```diff
@@ -77,12 +75,12 @@ export function batchedEventUpdates(fn, a, b) {
...
 }
 
-export function executeUserEventHandler(fn: any => void, value: any) {
+// This is for the React Flare event system
+export function executeUserEventHandler(fn: any => void, value: any): any {
   const previouslyInEventHandler = isInsideEventHandler;
   try {
     isInsideEventHandler = true;
-    const type = typeof value === 'object' && value !== null ? value.type : '';
-    invokeGuardedCallbackAndCatchFirstError(type, fn, undefined, value);
```
gaearon: Is there a follow-up to fix this? It seems important that event handler doesn't crash the rest of the code?

trueadm: Yeah, I plan on addressing it this week.

necolas: This was already fixed in a follow-up right? This guard is still there on master

trueadm: LOL. I did, I’m clearly losing the plot.

PR #16659, opened 2019-09-04
[Scheduler] Add "combined" www build

acdlite:
The Scheduler module in www is loaded before most of our other infra: too early to conditionally load a separate build in production versus development. Which means we're currently sending both a prod and dev bundle to every single user.

We should fix this on the Facebook infra side, but in the meantime, this updates the build script to generate a new type of bundle, `FB_WWW_COMBINED`, which preserves __DEV__ conditions.

### DIFF
--- a/scripts/rollup/forks.js
+++ b/scripts/rollup/forks.js
@@ -10,6 +10,7 @@ const UMD_PROFILING = bundleTypes.UMD_PROFILING;
 const FB_WWW_DEV = bundleTypes.FB_WWW_DEV;
 const FB_WWW_PROD = bundleTypes.FB_WWW_PROD;
 const FB_WWW_PROFILING = bundleTypes.FB_WWW_PROFILING;
+const FB_WWW_COMBINED = bundleTypes.FB_WWW_COMBINED;
 const RN_OSS_DEV = bundleTypes.RN_OSS_DEV;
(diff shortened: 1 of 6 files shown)


acdlite:
I should add more context — we don't actually use DEV checks in the Scheduler package currently. We could load the dev bundle in both dev and prod. However, that might not always be the case, so I've opened this PR as a suggestion for how to handle this if it comes up in the future.

bvaughn:
Maybe we should just add a lint check that fails if we ever add a `DEV` check to the `scheduler` package.

scripts/rollup/forks.js
```diff
@@ -158,11 +159,7 @@ const forks = Object.freeze({
   },
 
   'scheduler/src/SchedulerFeatureFlags': (bundleType, entry, dependencies) => {
-    if (
-      bundleType === FB_WWW_DEV ||
-      bundleType === FB_WWW_PROD ||
-      bundleType === FB_WWW_PROFILING
-    ) {
+    if (bundleType === FB_WWW_COMBINED) {
```
trueadm: Why did you remove the other two bundleTypes here?

acdlite: Because the purpose of the combined bundle is to replace the dev and prod bundles with a single bundle that includes both code paths

acdlite:
Outdated. Worked around this in www.

PR #16662, opened 2019-09-04
[react-events] Refine executeUserEventHandler

trueadm:
This is a follow up to #16657 where I wasn't quite happy about losing the usage of `invokeGuardedCallbackAndCatchFirstError`. This reinstates `invokeGuardedCallbackAndCatchFirstError` and works around the problem-space by using an inline function to work out the user callback return value.


packages/react-native-renderer/src/ReactFabricEventResponderSystem.js
```diff
@@ -106,7 +106,6 @@ const eventResponderContext: ReactNativeResponderContext = {
       }
       case ContinuousEvent: {
         executeUserEventHandler(eventListener, eventValue);
-        break;
```
necolas: We should probably leave the `break` in here and in the DOM version in case any new cases get added later

PR #16663, opened 2019-09-05
Track "pending" and "suspended" ranges

acdlite:
A FiberRoot can have pending work at many distinct priorities. (Note: we refer to these levels as "expiration times" to distinguish the concept from Scheduler's notion of priority levels, which represent broad categories of work. React expiration times are more granular. They're more like a concurrent thread ID, which also happens to correspond to a moment on a timeline. It's an overloaded concept and I'm handwaving over some of the details.)

Given a root, there's no convenient way to read all the pending levels in the entire tree, i.e. there's no single queue-like structure that tracks all the levels, because that granularity of information is not needed by our algorithms. Instead we track the subset of information that we actually need — most importantly, the highest priority level that exists in the entire tree.

Aside from that, the other information we track includes the range of pending levels that are known to be suspended, and therefore should not be worked on.

This is a refactor of how that information is tracked, and what each field represents:

- A *pending* level is work that is unfinished, or not yet committed. This includes work that is suspended from committing. `firstPendingTime` and `lastPendingTime` represent the range of pending work. (Previously, "pending" was the same as "not suspended.")
- A *suspended* level is work that did not complete because data was missing. `firstSuspendedTime` and `lastSuspendedTime` represent the range of suspended work. It is a subset of the pending range. (These fields are new to this commit.)
- `nextAfterSuspendedTime` represents the next known level that comes after the suspended range.

This commit doesn't change much in terms of observable behavior. The one change is that, when a level is suspended, React will continue working on the next known level instead of jumping straight to the last pending level. Subsequent commits will use this new structure for a more substantial refactor for how tasks are scheduled per root.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberRoot.js
+++ b/packages/react-reconciler/src/ReactFiberRoot.js
@@ -73,8 +73,15 @@ type BaseFiberRootProperties = {|
   firstPendingTime: ExpirationTime,
   // The latest pending expiration time that exists in the tree
   lastPendingTime: ExpirationTime,
-  // The time at which a suspended component pinged the root to render again
-  pingTime: ExpirationTime,
+  // The earliest suspended expiration time that exists in the tree
+  firstSuspendedTime: ExpirationTime,
+  // The latest suspended expiration time that exists in the tree
+  lastSuspendedTime: ExpirationTime,
+  // The next known expiration time after the suspended range
+  nextAfterSuspendedTime: ExpirationTime,
+  // The latest time at which a suspended component pinged the root to
+  // render again
+  lastPingedTime: ExpirationTime,
 |};
 
 // The following attributes are only used by interaction tracing builds.
@@ -120,7 +127,10 @@ function FiberRootNode(containerInfo, tag, hydrate) {
   this.callbackExpirationTime = NoWork;
   this.firstPendingTime = NoWork;
   this.lastPendingTime = NoWork;
-  this.pingTime = NoWork;
+  this.firstSuspendedTime = NoWork;
+  this.lastSuspendedTime = NoWork;
+  this.nextAfterSuspendedTime = NoWork;
+  this.lastPingedTime = NoWork;
 
   if (enableSchedulerTracing) {
     this.interactionThreadID = unstable_getThreadID();
@@ -150,4 +160,54 @@ export function createFiberRoot(
   uninitializedFiber.stateNode = root;
 
   return root;
+}
+
+export function isRootSuspendedAtTime(
+  root: FiberRoot,
+  expirationTime: ExpirationTime,
+): boolean {
+  const firstSuspendedTime = root.firstSuspendedTime;
+  const lastSuspendedTime = root.lastSuspendedTime;
+  return (
+    firstSuspendedTime !== NoWork &&
+    (firstSuspendedTime >= expirationTime &&
+      lastSuspendedTime <= expirationTime)
+  );
+}
+
+export function markRootSuspendedAtTime(
+  root: FiberRoot,
+  expirationTime: ExpirationTime,
+): void {
+  const firstSuspendedTime = root.firstSuspendedTime;
+  const lastSuspendedTime = root.lastSuspendedTime;
+  if (firstSuspendedTime < expirationTime) {
+    root.firstSuspendedTime = expirationTime;
+  }
+  if (lastSuspendedTime > expirationTime || firstSuspendedTime === NoWork) {
+    root.lastSuspendedTime = expirationTime;
+  }
+
+  if (expirationTime <= root.lastPingedTime) {
+    root.lastPingedTime = NoWork;
+  }
+}
+
+export function markRootUnsuspendedAtTime(
+  root: FiberRoot,
+  expirationTime: ExpirationTime,
+): void {
+  if (expirationTime <= root.lastSuspendedTime) {
+    // The entire suspended range is now unsuspended.
+    root.firstSuspendedTime = root.lastSuspendedTime = root.nextAfterSuspendedTime = NoWork;
+  } else if (expirationTime <= root.firstSuspendedTime) {
+    // Part of the suspended range is now unsuspended. Narrow the range to
+    // include everything between the unsuspended time (non-inclusive) and the
+    // last suspended time.
+    root.firstSuspendedTime = expirationTime - 1;
+  }
+
+  if (expirationTime <= root.lastPingedTime) {
+    root.lastPingedTime = NoWork;
+  }
 }
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -67,6 +67,11 @@ import {
 
 import {createWorkInProgress, assignFiberPropertiesInDEV} from './ReactFiber';
 import {
+  isRootSuspendedAtTime,
+  markRootSuspendedAtTime,
+  markRootUnsuspendedAtTime,
+} from './ReactFiberRoot';
+import {
   NoMode,
   StrictMode,
   ProfileMode,
@@ -377,8 +382,6 @@ export function scheduleUpdateOnFiber(
     return;
   }
 
-  root.pingTime = NoWork;
-
   checkForInterruption(fiber, expirationTime);
   recordScheduleUpdate();
 
@@ -492,6 +495,9 @@ function markUpdateTimeFromFiberToRoot(fiber, expirationTime) {
     if (lastPendingTime === NoWork || expirationTime < lastPendingTime) {
       root.lastPendingTime = expirationTime;
     }
+
+    // Mark that the root is no longer suspended at this time.
+    markRootUnsuspendedAtTime(root, expirationTime);
   }
 
   return root;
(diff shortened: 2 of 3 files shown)


packages/react-reconciler/src/ReactFiberRoot.js
```diff
@@ -151,3 +161,51 @@ export function createFiberRoot(
...
+  );
+}
+
+export function markRootSuspendedAtTime(
+  root: FiberRoot,
+  expirationTime: ExpirationTime,
+): void {
+  const firstSuspendedTime = root.firstSuspendedTime;
+  const lastSuspendedTime = root.lastSuspendedTime;
+  if (firstSuspendedTime < expirationTime) {
```
Jessidhia: These branches would be easier to read with the comparison flipped (compare `expirationTime` against the current values)

Jessidhia: Oh, right, ExpirationTime is a decreasing clock

packages/react-reconciler/src/ReactFiberRoot.js
```diff
@@ -151,3 +161,51 @@ export function createFiberRoot(
...
+export function markRootUnsuspendedAtTime(
+  root: FiberRoot,
+  expirationTime: ExpirationTime,
+): void {
+  if (expirationTime <= root.lastSuspendedTime) {
+    // The entire suspended range is now unsuspended.
+    root.firstSuspendedTime = root.lastSuspendedTime = root.nextAfterSuspendedTime = NoWork;
+  } else if (expirationTime <= root.firstSuspendedTime) {
+    // Part of the suspended range has been unsuspended.
+    root.firstSuspendedTime += 1;
```
Jessidhia: I don't understand why move the `firstSuspendedTime` "one earlier" 🤔

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1038,11 +1047,20 @@ function renderRoot(
             prepareFreshStack(root, expirationTime);
             return renderRoot.bind(null, root, expirationTime);
           }
-          const lastPendingTime = root.lastPendingTime;
-          if (lastPendingTime < expirationTime) {
+
+          const nextAfterSuspendedTime = root.nextAfterSuspendedTime;
```
Jessidhia:
```suggestion
          const { nextAfterSuspendedTime } = root
```

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1058,6 +1076,11 @@ function renderRoot(
       return commitRoot.bind(null, root);
     }
     case RootSuspendedWithDelay: {
+      markRootSuspendedAtTime(root, expirationTime);
+      const lastSuspendedTime = root.lastSuspendedTime;
```
Jessidhia:
```suggestion
      const { lastSuspendedTime } = root
```

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1077,11 +1100,20 @@ function renderRoot(
           prepareFreshStack(root, expirationTime);
           return renderRoot.bind(null, root, expirationTime);
         }
-        const lastPendingTime = root.lastPendingTime;
-        if (lastPendingTime < expirationTime) {
+
+        const nextAfterSuspendedTime = root.nextAfterSuspendedTime;
```
Jessidhia:
```suggestion
        const { nextAfterSuspendedTime } = root
```

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1425,6 +1457,14 @@ function completeUnitOfWork(unitOfWork: Fiber): Fiber | null {
   return null;
 }
 
+function getRemainingExpirationTime(fiber: Fiber) {
+  const updateExpirationTime = fiber.expirationTime;
+  const childExpirationTime = fiber.childExpirationTime;
+  return updateExpirationTime > childExpirationTime
+    ? updateExpirationTime
+    : childExpirationTime;
```
Jessidhia: `Math.max`?

sebmarkbage: We avoid using methods when there is syntax because it requires additional look ups where as a comparison can trivially be understood as a single op.

packages/react-reconciler/src/ReactFiberWorkLoop.js
sebmarkbage: We also tend to avoid modern syntax as we don’t trust the transforms to produce good results.

packages/react-reconciler/src/ReactFiberRoot.js
```diff
@@ -151,3 +161,51 @@ export function createFiberRoot(
...
+export function isRootSuspendedAtTime(
+  root: FiberRoot,
+  expirationTime: ExpirationTime,
+): boolean {
+  const firstSuspendedTime = root.firstSuspendedTime;
+  const lastSuspendedTime = root.lastSuspendedTime;
+  return (
+    firstSuspendedTime !== NoWork &&
+    (firstSuspendedTime >= expirationTime &&
+      lastSuspendedTime <= expirationTime)
```
NE-SmallTown: FYI, totally I think it's time to rename expirationTime because it's been talked a few months, but for now, since it's a 'time' could we compare them through the 'time order'? I.e. from small to large (`lastSuspendedTime <= expirationTime` && expirationTime <= firstSuspendedTime)

packages/react-reconciler/src/ReactFiberRoot.js
acdlite: Good catch, fixed it

packages/react-reconciler/src/ReactFiberRoot.js
```diff
@@ -151,3 +161,51 @@ export function createFiberRoot(
 
   return root;
 }
+
+export function isRootSuspendedAtTime(
+  root: FiberRoot,
+  expirationTime: ExpirationTime,
+): boolean {
+  const firstSuspendedTime = root.firstSuspendedTime;
+  const lastSuspendedTime = root.lastSuspendedTime;
```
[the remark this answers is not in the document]
acdlite: https://github.com/facebook/react/pull/16663#discussion_r321060138

packages/react-reconciler/src/ReactFiberRoot.js
```diff
@@ -151,3 +161,51 @@ export function createFiberRoot(
...
+      lastSuspendedTime <= expirationTime)
+  );
+}
+
+export function markRootSuspendedAtTime(
+  root: FiberRoot,
+  expirationTime: ExpirationTime,
+): void {
+  const firstSuspendedTime = root.firstSuspendedTime;
+  const lastSuspendedTime = root.lastSuspendedTime;
```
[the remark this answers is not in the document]
acdlite: https://github.com/facebook/react/pull/16663#discussion_r321060138

packages/react-reconciler/src/ReactFiberRoot.js
acdlite: ExpirationTime using to be increasing but we flipped it because it tends to result in fewer comparisons to `NoWork`: https://github.com/facebook/react/pull/13912

packages/react-reconciler/src/ReactFiberWorkLoop.js
acdlite: https://github.com/facebook/react/pull/16663#discussion_r321060138

packages/react-reconciler/src/ReactFiberWorkLoop.js
acdlite: https://github.com/facebook/react/pull/16663#discussion_r321060138

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -831,21 +830,25 @@ function renderRoot(
...
     // for that case, we'll wait until we complete.
     if (workInProgressRootHasPendingPing) {
       // We have a ping at this expiration. Let's restart to see if we get unblocked.
       prepareFreshStack(root, expirationTime);
-    } else {
-      const lastPendingTime = root.lastPendingTime;
-      if (lastPendingTime < expirationTime) {
-        // There's lower priority work. It might be unsuspended. Try rendering
-        // at that level immediately, while preserving the position in the queue.
-        return renderRoot.bind(null, root, lastPendingTime);
```
sebmarkbage: We're discussing this.  Seems like this refactoring changed things.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -831,21 +830,25 @@ function renderRoot(
...
+    // just going to try to recover at the pending time anyway so we might as
+    // well start doing that eagerly.
+    //
     // Ideally we should be able to do this even for retries but we don't yet
     // know if we're going to process an update which wants to commit earlier,
     // and this path happens very early so it would happen too often. Instead,
     // for that case, we'll wait until we complete.
     if (workInProgressRootHasPendingPing) {
       // We have a ping at this expiration. Let's restart to see if we get unblocked.
       prepareFreshStack(root, expirationTime);
```
sebmarkbage: This can be removed (maybe) because there exists one in the ping function already.

packages/react-reconciler/src/ReactFiberWorkLoop.js
sebmarkbage: This can be replaced by a prepareFreshStack in an update that happens during a yield.

If the current tree has already determined that it has remaining work at lower pri, we can use that info to eagerly throw to terminate the render and try at lower pri.

acdlite:

Issue #16629, opened 2019-08-31
Chrome Extension Background Page and useEffect Not Triggered


mikecann:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

When using `useEffect` from a background page in a chrome extension the body of the effect isnt triggered.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

See the demo project here: https://github.com/mikecann/bgpage-hooks-issue

**What is the expected behavior?**

`useEffect` should be triggered

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

It stopped working somewhere between react 16.8 and 16.9 , you can change the react and react_dom version in the demo project and see that it works as expected in 16.8

**Workaround**

For now the issue can be worked around as demonstrated: https://github.com/mikecann/bgpage-hooks-issue/blob/master/bg.tsx#L5

So im not really sure whats going on but perhaps because background pages arent actually visually rendered to the screen then React has optimized things so that effects arent triggered?

mikecann:
Perhaps related to: https://github.com/facebook/react/issues/16606 ??

Jack-Works:
> Perhaps related to: #16606 ??

That should be not related. #16606 only happened in Firefox content script, but this issue happened in Chome background page (and why run React in the background page?)

mikecann:
> > Perhaps related to: #16606 ??
> 
> That should be not related. #16606 only happened in Firefox content script, but this issue happened in Chome background page (and why run React in the background page?)

Well im wondering if its something to do with requestAnimationFrame? Are react useEffect's now triggered by requestAnimationFrame? Perhaps requestAnimationFrame doesnt work in background pages?

> and why run React in the background page?

Because I find the useState and state update cycle super useful for creating reactive logic in the background. I can create custom hooks that can be used in both the background and other extension pages.

Jack-Works:
Yes, if user cannot see the page, rAF will not be called

Jack-Works:
You can replace rAF with setTimeout 0 to make it work in background page.

Jack-Works:
And I have made a hooks without react demo few days ago

```ts
interface Hooks {
    useState<T>(initialValue: T): [T, (newVal: T) => void]
    useEffect(f: () => () => void): void
}
function useHooks(fn: (hooks: Hooks) => void) {
[...]

PR #16672, opened 2019-09-05
Remove experimental scheduler flags

gaearon:
Removes `requestIdleCallbackBeforeFirstFrame` and `requestTimerEventBeforeFirstFrame`.

Both were rAF-specific experiments but we went with MessageLoop instead. We left rAF temporarily in the code but since we never actually used these code paths, I think we should delete them to make the code easier to understand.

### DIFF
--- a/packages/scheduler/src/forks/SchedulerHostConfig.default.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.default.js
@@ -7,8 +7,6 @@
 
 import {
   enableIsInputPending,
-  requestIdleCallbackBeforeFirstFrame as requestIdleCallbackBeforeFirstFrameFlag,
-  requestTimerEventBeforeFirstFrame,
   enableMessageLoopImplementation,
 } from '../SchedulerFeatureFlags';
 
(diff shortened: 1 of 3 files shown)


packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -359,50 +351,8 @@ if (
...
-          );
-        }
-        // Alternate strategy to address the same problem. Scheduler a timer
-        // with no delay. If this fires before the rAF, that likely indicates
-        // that there's idle time before the next vsync. This isn't always the
-        // case, but we'll be aggressive and assume it is, as a trade off to
-        // prevent idle periods.
-        let idleTimeoutID;
-        if (requestTimerEventBeforeFirstFrame) {
-          idleTimeoutID = setTimeout(function onTimerEventBeforeFirstFrame() {
```
acdlite: Since you've removed this timeout, there's now no timeout fallback for the initial `rAF` call, like there is for subsequent rAFs: https://github.com/facebook/react/blob/ff006451ad792ed26793691402bf7b5f813205f7/packages/scheduler/src/forks/SchedulerHostConfig.default.js#L306-L311

But it's whatever since we're going to disable the rAF stuff anyway.

acdlite: That timeout should have already been in the `else` branch of this feature, anyway.

gaearon: >there's now no timeout fallback for the initial rAF call, like there is for subsequent rAFs

I guess that explains https://github.com/facebook/react/issues/16629?

PR #16678, opened 2019-09-06
Get next expiration time from FiberRoot

acdlite:
## Based on #16663

Given a FiberRoot, we should be able to determine the next expiration time that needs to be worked on, taking into account the levels that are pending, suspended, pinged, and so on.

This removes the `expirationTime` argument from `scheduleCallbackForRoot`, and renames it to `ensureRootIsScheduled` to reflect the new signature. The expiration time is instead read from the root using a new function, `getNextExpirationTimeToWorkOn`.

The next step will be to remove the `expirationTime` argument from `renderRoot`, too.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -67,6 +67,11 @@ import {
 
 import {createWorkInProgress, assignFiberPropertiesInDEV} from './ReactFiber';
 import {
+  isRootSuspendedAtTime,
+  markRootSuspendedAtTime,
+  markRootUnsuspendedAtTime,
+} from './ReactFiberRoot';
+import {
   NoMode,
   StrictMode,
   ProfileMode,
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -827,30 +871,6 @@ function renderRoot(
...
-      // We have a ping at this expiration. Let's restart to see if we get unblocked.
-      prepareFreshStack(root, expirationTime);
-    } else if (!isSync) {
-      // Check if there's work that isn't in the suspended range
-      const firstPendingTime = root.firstPendingTime;
-      if (!isRootSuspendedAtTime(root, firstPendingTime)) {
-        // There's a pending update that falls outside the range of
-        // suspended work.
-        if (firstPendingTime > expirationTime) {
-          return renderRoot.bind(null, root, firstPendingTime);
```
sebmarkbage: As mentioned in #16663. This needs a replacement (and a test).

acdlite:

PR #16689, opened 2019-09-06
[react-events] Tap: add maximumDistance prop

necolas:
A prop for configuring the maximum distance that the active pointer can move before the tap is cancelled.

### DIFF
--- a/packages/react-events/src/dom/Tap.js
+++ b/packages/react-events/src/dom/Tap.js
@@ -21,29 +21,33 @@ import {
   isMac,
   dispatchDiscreteEvent,
   dispatchUserBlockingEvent,
+  getTouchById,
+  hasModifierKey,
 } from './shared';
 
-type TapProps = {|
-  disabled: boolean,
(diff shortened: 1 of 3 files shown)


aweary:
@necolas I'm not sure its related to these changes, but long press and press via the Enter key don't seem to work for me in that demo (Chrome 76 on macOS)

necolas:
LongPress will be a separate responder. And see https://github.com/facebook/react/commit/9ce8711d5a7c84e71587c75d2935966023ee6dbb for related context

necolas:
> What happens if the maximumDistance is set to 0? I assume that wouldn't work here?

Why is that?

trueadm:
@necolas Unless I'm mistaken, `moveDistance > maxDistance` means that if you didn't move and max distance is `0`, then it would fail to trigger?

necolas:
Yeah I think if you moved back over the initial point, in that case it would not fire `onTapUpdate`

Issue #16484, opened 2019-08-19
DevTools: Should be able to add a prop that is not yet set


bvaughn:
Right now I can inspect a component and edit its props but I could not find a way to add a _new_ prop that wasn't already set.

---
Originally reported by @steveluscher via https://github.com/bvaughn/react-devtools-experimental/issues/300

hristo-kanchev:
I'm currently playing around with this and I would like to get some feedback regarding the visual representation of it.

I presume that we should be able to support more complex data structures, right?

![add-new-prop](https://user-images.githubusercontent.com/23095052/63599546-2107d700-c5c2-11e9-91ad-a057d57bc41c.gif)

bvaughn:
I like the "+" button.

I think we should probably use `JSON.parse` to accept new input types, so that would enable strings/ints/booleans/objects/arrays.

hristo-kanchev:
Awesome! Thanks @bvaughn. I'm on it 🚀

hristo-kanchev:
@bvaughn What do you think of this style of editing props?

It's aligned with the style editor. 😄 

If you are cool with this I can start preparing the branch. 🎉 

👇 

**Number -> String -> null -> Number**

![props-1](https://user-images.githubusercontent.com/23095052/64120611-f32e4980-cd9c-11e9-9fab-fd6c6970aea5.gif)

---

👇 
**String -> Object**

![props-2](https://user-images.githubusercontent.com/23095052/64120612-f32e4980-cd9c-11e9-8f07-0df02f4f2b3d.gif)

bvaughn:
That looks really nice.

hristo-kanchev:
> That looks really nice. Let's do it!

Cool! I'll try and get it finished by the end of the week. I'll ping you when it's done. 
Thanks for the feedback Brian!

hristo-kanchev:
**Update**: Almost done! Just finishing up some styling tweaks. PR should be open tomorrow :)

Issue #16476, opened 2019-08-19
DevTools: Improve props/state value editing interface


bvaughn:
Follow up to a discussion thread on #294 and a Messenger chat with @gaearon.

The current UI for editing props/state/hooks values has a couple of shortcomings:
* strings/numeric values are sometimes of an unclear type. (DevTools uses the correct `<input type>` but this is subtle. Showing e.g. quotation marks around strings could be more helpful.
* `null` initial values can only become strings at the moment. (DevTools does not attempt to parse numeric or boolean values and change the input type. If it did, we would probably also need to enable a way for you to opt back out of that input type somehow in case it was incorrect.)
* fields that maybe support multiple types (e.g. string | number) are locked into a single type (whichever type they happen to be initially).

Additional quirks that might be worth ironing out:
* Non-editable strings are currently wrapped in quotation marks but editable ones aren't.

---
Originally reported via https://github.com/bvaughn/react-devtools-experimental/issues/321

a-x-:
Also:
1. nested objects shown as 'Object' strings (also, when copy props button pressed)
    <img width="442" alt="Screenshot 2019-08-29 at 20 28 30" src="https://user-images.githubusercontent.com/6201068/63962732-15258480-ca9c-11e9-89b1-1de28151208c.png">
    <img width="496" alt="Screenshot 2019-08-29 at 20 28 52" src="https://user-images.githubusercontent.com/6201068/63962738-16ef4800-ca9c-11e9-8605-b479b57a9c0a.png">
    ```
    {...
        "accounts": ["Object", "Object", "Object"],
        "resetForms": "fn()"
    ...}
    ```

2. context menu on fn() missed. in previous devtools I was able to jump to func-prop definition source

bvaughn:
@a-x- The above comment doesn't really seem related to this issue. Please file a new issue?

> nested objects shown as 'Object' strings (also, when copy props button pressed)

Please also include:
* Which version of DevTools you have installed
* Which browser you are using
* An example project (Code Sandbox would be nice) that shows this bug. (I'm not seeing it in DevTools when I look now.)

If you look at [the demo app](https://react-devtools-experimental.now.sh/) and search for "InspectableElements", you'll see that nested arrays and objects are serialized correctly:

![Screen Shot 2019-08-29 at 11 08 41 AM](https://user-images.githubusercontent.com/29597/63965139-6195a380-ca4d-11e9-8999-69483f6549a1.png)

[...]

PR #16700, opened 2019-09-07
[DevTools] Support for adding props | Improved state/props value editing 

hristo-kanchev:
**Solves**:
- #16484 
- #16476 

As both issues are highly overlapping, @bvaughn and I decided to submit the fix for both in one PR.

**Changes**:
- Extracted `sanitizeForParse` to the `utils` file as it's used in multiple places
- Added new component - `EditableName`
- Added support for adding additional entries (currently only enabled for `prop`)
- Fixed the issue with locking values by `type`.
- String values are now surrounded by quotes.

**Screenshots**:

**Adding new props**
props-1
props-2
props-3

---

**Overwriting an existing prop value**
props-4

---

**Changing data type** - **null** -> **array**
props-5


packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
```diff
@@ -27,20 +28,38 @@ export default function EditableValue({
...
   const [hasPendingChanges, setHasPendingChanges] = useState(false);
-  const [editableValue, setEditableValue] = useState(value);
+  const [stringifiedValue, setStringifiedValue] = useState(
+    JSON.stringify(value),
+  );
+  const [editableValue, setEditableValue] = useState(stringifiedValue);
   const inputRef = useRef<HTMLInputElement | null>(null);
 
-  if (hasPendingChanges && editableValue === value) {
+  useEffect(
```
hristo-kanchev: If the value changes we need to update the stringified version of it.

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
```diff
@@ -67,71 +87,59 @@ export default function EditableValue({
...
     },
-    [editableValue, dataType, overrideValueFn, path, value],
+    [editableValue, isValid, dataType, overrideValueFn, path, value],
   );
 
-  // Render different input types based on the dataType
-  let type = 'text';
-  if (dataType === 'boolean') {
-    type = 'checkbox';
-  } else if (dataType === 'number') {
```
hristo-kanchev: We don't need this anymore. Now we have the ability to change the value from a number to an array for example.

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
```diff
@@ -67,71 +87,59 @@ export default function EditableValue({
...
   let placeholder = '';
-  if (value === null) {
-    placeholder = '(null)';
-  } else if (value === undefined) {
+  if (value === undefined) {
     placeholder = '(undefined)';
-  } else if (dataType === 'string') {
-    placeholder = '(string)';
+  } else {
+    placeholder = 'Enter valid JSON';
```
hristo-kanchev: I think it would be good to have this placeholder for non `undefined` values. What do you think @bvaughn ?

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
```diff
@@ -67,71 +87,59 @@ export default function EditableValue({
...
   }
 
   return (
     <Fragment>
       {dataType === 'boolean' && (
         <label className={styles.CheckboxLabel}>
           <input
-            checked={inputValue}
+            checked={inputValue === 'true'}
```
hristo-kanchev: inputValue is now the `stringifiedVersion` of `value`. That's why we need to make this check.

davidfurlong:
Hey there @hristo-kanchev,
I'd like to weigh in here as well as I was looking at this part of the code and submitted a PR a long time ago that attempted to fix these bits. I'm happy to submit a PR to fix the below off of this one or for you to look at these issues

**in the KeyValue component**
- New line characters (\n) are hidden from string renders (perhaps hidden characters should be shown but in a different color?)
- Infinity is rendered as nothing
- -Infinity is rendered as nothing
- NaN is rendered as nothing
- '4' and 4 render as the same (it appears your string change fixes this)
- empty string renders as (string) when "" is probably better
- If strings are now in quotation marks perhaps it makes sense for these changes: (null) -> null and (undefined) -> undefined

hristo-kanchev:
> * New line characters (\n) are hidden from string renders (perhaps hidden characters should be shown but in a different color?)
From what I can see `\n` characters are rendered. Or am I missing something?
Screenshot 2019-09-08 at 15 15 37


> * Infinity is rendered as nothing
> * -Infinity is rendered as nothing
> * NaN is rendered as nothing
With this new addition they will be rendered as `null` do we want to fix this in this PR, Brian? 
This is due to `JSON.stringify`.

> * empty string renders as (string) when "" is probably better
With this PR they will be rendered as ""

Screenshot 2019-09-08 at 15 19 19

> * If strings are now in quotation marks perhaps it makes sense for these changes: (null) -> null and (undefined) -> undefined
With this PR `null` will be rendered as `null` but undefined is indeed `(undefined)`
Screenshot 2019-09-08 at 15 20 51

I think it's up to Brian to decide. 
I don't mind adding the support for this in this PR, but my own personal opinion is that the above unsolved issues should be in different issue.

What do you think @bvaughn ?

packages/react-devtools-shared/src/devtools/views/Components/EditableKey.js
```diff
@@ -0,0 +1,78 @@
...
+
+type OverrideKeyFn = (path: Array<string | number>, value: any) => void;
+
+type EditableKeyProps = {|
+  key?: string,
+  overrideKeyFn: OverrideKeyFn,
+|};
+
+export default function EditableKey({
+  key = '',
```
bvaughn: `key` and `ref` can't be passed in as props though. (They are extracted and treated specially.) So this will _never_ be anything but undefined (or `""`).

Can we pick a different name for whatever this is? (One that isn't overloaded in React terminology.) I would just suggest naming the state `value` (and `setValue`).

I know this is copying the precedent set by` EditableValue`- but it's borderline mixing controlled and uncontrolled behavior in a way that seems similar similar to [this](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-erasing-state-when-props-change). Maybe we should name the incoming thing e.g. `initialValue` or something to be more clear?

packages/react-devtools-shared/src/devtools/views/Components/EditableKey.js
```diff
@@ -0,0 +1,78 @@
...
+  const [isValid, setIsValid] = useState(false);
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  useEffect(
+    () => {
+      if (inputRef.current !== null) {
+        inputRef.current.focus();
+      }
+    },
+    [inputRef],
```
bvaughn: You shouldn't need to specify a ref as a dependency.

packages/react-devtools-shared/src/devtools/views/Components/EditableKey.js
bvaughn: Hm. Did you mean to use this component for existing keys too? Looks like you're only using it for new keys. Just curious!

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
bvaughn: Is it expected that once a prop becomes a boolean, you can't change it back?
aaaaaKapture 2019-09-09 at 16 42 55

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
bvaughn: Looks like this?

https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-unconditionally-copying-props-to-state

packages/react-devtools-shared/src/devtools/views/Components/InspectedElementTree.js
```diff
@@ -73,6 +122,17 @@ export default function InspectedElementTree({
               value={value}
             />
           ))}
+        {entryToAdd && (
+          <div className={styles.AddEntry}>
+            <EditableKey overrideKeyFn={handleEntryAddName} />:
```
bvaughn: Is there a way to edit an attribute after it's been added? Looks like it's only editable when it's "new". I guess it's weird to rename a prop being passed in, because a new render would re-add the previous one.

Seems like it might be nice for us to provide a way to edit (or even delete) overridden props though.

packages/react-devtools-shared/src/devtools/views/Components/EditableKey.js
Jessidhia: How about calling it `name`, going by the `Object.getOwnPropertyNames` nomenclature?

packages/react-devtools-shared/src/devtools/views/Components/InspectedElementTree.js
Jessidhia: A "reset overrides" button and/or toggle would be a nice debugging feature.

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
Jessidhia: Rather, that looks like it should be a `useMemo`, or even not a hook at all. You call `JSON.stringify` for the (unused) initializer value on every render pass anyway.

packages/react-devtools-shared/src/devtools/views/Components/EditableKey.js
hristo-kanchev: @bvaughn I initially had this idea and played around with it in the `KeyValue` component, but I saw that the PR is getting chunky and decided to not implement it, as it will add more complexity. 
Maybe we can create a new issue for this one and after this get's merged we can start enhancing this feature? What do you think?

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
hristo-kanchev: I thought it would be cool to keep this as it is. 
Users can still override it via Add new prop -> type `name` and set the value to `"hello world"` for example.
Do you think it would be better if we just displayed `true` / `false` instead of the checkbox? 
That way we can change it more easily.

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
hristo-kanchev: I completely removed this bit. Thanks for the input!

packages/react-devtools-shared/src/devtools/views/Components/InspectedElementTree.js
hristo-kanchev: I really like the idea from @Jessidhia ! 🚀 

I've added support for `canResetEntries`

Demo:

reset-props

hristo-kanchev: @bvaughn Regarding the ability to edit a name of an already set prop, as I mentioned in the other comment, I think we should add this support in another PR.

hristo-kanchev:
Hey @bvaughn ! 
I addressed all of your comments. I'll be going on holidays from tomorrow, so I won't be able to add the additional feature requests that you and @Jessidhia request. 
Feel free to hijack this PR and add them if it's urgent to get them in. If not, I can add them myself next week. 

Thanks for the feedback!

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
bvaughn: > Users can still override it via Add new prop -> type name and set the value to "hello world" for example.

That seems pretty non-obvious though. (I don't think it would occur to people to do this).

> Do you think it would be better if we just displayed true / false instead of the checkbox?

Yeah 😄 I think we should do this.

bvaughn:
> > * Infinity is rendered as nothing
> > * -Infinity is rendered as nothing
> > * NaN is rendered as nothing
> 
> With this new addition they will be rendered as null do we want to fix this in this PR, Brian?
This is due to JSON.stringify.

Whoops, sorry. I missed this. Could we use a placeholder for these values?

bvaughn:
Thanks for the quick turnaround, @hristo-kanchev! I'll review this again this morning and we'll try to at least get an intermediate version landed for now since it's an improvement overall. We can always iterate on it later.

Enjoy your vacation~

bvaughn:
I think I've got the stringify/parse conditions handling at least the more common values:
Screen Shot 2019-09-10 at 11 10 56 AM

Not sure if we care to support the less common ones (e.g. -Infinity) but we can iterate later.

packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
bvaughn: I'll just go ahead and make this change.

packages/react-devtools-shared/src/devtools/views/hooks.js
```diff
@@ -8,11 +8,75 @@
...
+  editableValue: any,
+  hasPendingChanges: boolean,
+  isValid: boolean,
+  parsedValue: any,
+  reset: () => void,
+  update: (newValue: any) => void,
+|};
+
+// Convenience hook for working with an editable value that is validated via JSON.parse.
+export function useEditableValue(
```
bvaughn: We could potentially refactor the style editor to use this hook as well, if we replaced the inline calls to `smartStringify` and `smartParse` with injected callbacks.

bvaughn:
I think this change caused a regression (#16859). The `<EditableValue>` `value` prop was renamed to `initialValue` but the `HooksTree` usage wasn't updated. Flow should have caught this though, and it didn't. That's concerning.

PR #16709, opened 2019-09-09
Bind rAF and cAF to window (#16606)

### LINKED ISSUE #16606: TypeError: "'requestAnimationFrame' called on an object that does not implement interface Window."
**Do you want to request a *feature* or report a *bug*?** **Bug**

TypeError: "'requestAnimationFrame' called on an object that does not implement interface Window."

https://github.com/facebook/react/blob/0f6e3cd61cf4a5a1491bb3c92780936aebc2a146/packages/scheduler/src/forks/SchedulerHostConfig.default.js#L88

In the WebExtension content script, globalThis !== window.
So when `const requestAnimationFrame = window.requestAnimationFrame`, the `this` binding of `rAF` is lost.

In the after calling, `rAF` rebinds to the globalThis, but globalThis doesn't implements Window interface.

PoC: Install this extension in the Firefox and you will see the error reported.
```index.js
try {
    const raf = window.requestAnimationFrame
    raf(() => console.log('Okay.'))
} catch (e) {
    console.warn(e)
}
```

```manifest.json
{
    "name": "rAF in content script test",
    "version": "0.0.1",
    "manifest_version": 2,
    "content_scripts": [{ "js": ["/index.js"], "matches": ["<all_urls>"] }]
}
```

Firefox, in WebExtension. 

Related to https://github.com/facebook/react/issues/16605

liady:
When capturing local references of requestAnimationFrame and cancelAnimationFrame - bind them to the window object.
Fixes #16606

### DIFF
--- a/packages/scheduler/src/forks/SchedulerHostConfig.default.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.default.js
@@ -83,8 +83,10 @@ if (
   const Date = window.Date;
   const setTimeout = window.setTimeout;
   const clearTimeout = window.clearTimeout;
-  const requestAnimationFrame = window.requestAnimationFrame && window.requestAnimationFrame.bind(window);
-  const cancelAnimationFrame = window.cancelAnimationFrame && window.cancelAnimationFrame.bind(window);
+  const requestAnimationFrame =
+    window.requestAnimationFrame && window.requestAnimationFrame.bind(window);
+  const cancelAnimationFrame =
+    window.cancelAnimationFrame && window.cancelAnimationFrame.bind(window);
 
   if (typeof console !== 'undefined') {
     // TODO: Remove fb.me link
--- a/packages/scheduler/src/forks/SchedulerHostConfig.default.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.default.js
@@ -83,8 +83,8 @@ if (
   const Date = window.Date;
   const setTimeout = window.setTimeout;
   const clearTimeout = window.clearTimeout;
-  const requestAnimationFrame = window.requestAnimationFrame;
-  const cancelAnimationFrame = window.cancelAnimationFrame;
+  const requestAnimationFrame = window.requestAnimationFrame && window.requestAnimationFrame.bind(window);
+  const cancelAnimationFrame = window.cancelAnimationFrame && window.cancelAnimationFrame.bind(window);
 
   if (typeof console !== 'undefined') {
     // TODO: Remove fb.me link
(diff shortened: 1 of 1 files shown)


gaearon:
1. Please add a comment explaining why this was necessary. 
2. Has someone verified this indeed fixes the reported issue?

liady:
This issue occured when running in environments where window exists but the global `this` does not equal `window`.
In those environments - dereferencing the functions (`requestAnimationFrame = window.requestAnimationFrame`) loses the binding to `window` and since the global `this` is different it throws an error.
By explicitly binding the derefernced function to window - this error is avoided.

gaearon:
I understand the reasoning. I’d just like someone to manually verify that this fixes the observable issue by doing a build and checking that. Could you please do it?

gaearon:
Sorry I didn’t clarify. By “add a comment” I meant please add a comment in the source code. Just a couple of lines.

packages/scheduler/src/forks/SchedulerHostConfig.default.js
```diff
@@ -83,8 +83,12 @@ if (
   const Date = window.Date;
   const setTimeout = window.setTimeout;
   const clearTimeout = window.clearTimeout;
-  const requestAnimationFrame = window.requestAnimationFrame;
-  const cancelAnimationFrame = window.cancelAnimationFrame;
+  // Explicitly bind local references to `window`, in order to support
+  // environments where the global `this` is not the window (e.g extensions)
+  const requestAnimationFrame =
+    window.requestAnimationFrame && window.requestAnimationFrame.bind(window);
```
gaearon: Do you mind making these more explicit, i.e.

```
const requestAnimationFrame =
    typeof window.requestAnimationFrame === 'function' ? window.requestAnimationFrame.bind(window) : undefined;
```

I know it's super unlikely but would be nice to recover gracefully on startup if globals are clobbered with invalid values.

liady:
@gaearon I checked it.
With a very minimal web-extension content script (repo [here](https://github.com/liady/react-webextension-poc)):
```javascript
container.id = "raf-test";
document.body.appendChild(container);

ReactDOM.render(React.createElement('div', null, `Hello rAF`), container);
```

I Managed to reproduce it without the fix:
Notok

And with the fix it works:
ok2

packages/scheduler/src/forks/SchedulerHostConfig.default.js
liady: Good point, done.

liady:
@acdlite do you think this change should be reverted? Or should it remain there to at least fix this specific issue (when trying to run React inside a web extension)?

phaistonian:
@liady Did you manage to _fix_ this issue in your app/extension? Without using this PR, that is.

liady:
@phaistonian no, still waiting for the final verdict on this PR...

Issue #14088, opened 2018-11-04
eslint-plugin-react-hooks does not work with wrapped components


AlexanderShushunov:
``` js
export const Counter = () => {
    let count, setCount;
    if (0 === 1) {
        [count, setCount] = useState(0);
    }
    return count;
};
const wrapper = Component => Component;
export const WrappedCounter = wrapper(() => {
    let count, setCount;
    if (0 === 1) {
        [count, setCount] = useState(0);
    }
    return count;
});
```
Linter shows an error in the first component, but ignore the second one.

gaearon:
We should probably handle this. Although it’s not ideal that you pass an anonymous function since that doesn’t give you a chance to name it.

AlexanderShushunov:
I think, it is a very common case. We have an anonymous function which is wrapped by several HOCs. An application uses only the wrapped variant. So, we do not need to think of a name of the inner function.

sandiiarov:
Why you don't want to use named function?
1. It helps to debug a code in react dev tools.
2. It will solve the issue with eslint.

```js
export const WrappedCounter = wrapper(Counter);
```

AlexanderShushunov:
@sandiiarov Because 'There are only two hard things in Computer Science...'
The example above is super simple. In real life finding out an appropriate name is a hard thing.

gaearon:
@AlexanderShushunov Would you like to send a PR to the plugin to support this?

AlexanderShushunov:
@gaearon I am not very experienced in eslint rules writing, but I always want to write one )

lukyth:
@AlexanderShushunov Are you still working on this?

gaearon:
I guess not, @lukyth feel free to look into it

fortinmike:
Another interesting case that triggers this is using the built-in `React.memo` function:

```javascript
const MyComponent = React.memo(props => {
  const count = 0;
  if (count === 1) useMyHook(); // NOT detected
  return <div>Hello</div>;
});
```

```javascript
const MyComponent = props => {
  const count = 0;
  if (count === 1) useMyHook(); // Detected!
  return <div>Hello</div>;
};
```

AlexanderShushunov:
@lukyth I tried to do it. But it was a bit complicated than I thought )

FredyC:
It's interesting that this is not getting more attention considering it does not work with React built-ins alone. And no, I am not able to fix it, AST is still too magical land for me :)

gaearon:
[...]

PR #16712, opened 2019-09-09
Fix ESLint hook rules for wrapped components

Shadowlauch:
This PR fixes the issue that rules like "no hooks in conditional statements" do not work with wrapped functions like you would have with forwardRef. (#14088)

I added new tests for these cases and ensured that the existing tests still pass. Making PR to huge projects is quite new to me, so I hope I did not miss anything.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
@@ -351,6 +351,30 @@ const tests = {
   invalid: [
     {
       code: `
+      // Invalid because it's dangerous and might not warn otherwise.
+      // This *must* be invalid.
+      const Component = wrapperFunction(() => {
+        if (cond) {
+          useConditionalHook();
+        }
+      });
+      `,
+      errors: [conditionalError('useConditionalHook')],
+    },
+    {
+      code: `
+      // Invalid because it's dangerous and might not warn otherwise.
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
```diff
@@ -349,6 +349,30 @@ const tests = {
...
   ],
   invalid: [
+    {
+      code: `
+      // Invalid because it's dangerous and might not warn otherwise.
+      // This *must* be invalid.
+      const Component = wrapperFunction(() => {
+        if (cond) {
+          useConditionalHook();
+        }
```
Jessidhia: I know the current heuristic doesn't check for `return` to tell whether something is a component or not, but the absence of a `return` here makes it look suspicious to me as the one non-symbol value components are not allowed to return is `undefined`.

Shadowlauch: Should I just return null at the end of the component? To be honest I did not include a return because none of the other tests did.

Shadowlauch:
Is there anything I can do to get this merged?

gaearon:
We don't want to support this convention for anything other than built-in things like memo. Write your components like regular functions and wrap them. This ensures they have good names.

```js
function Foo() {
}

Foo = wrap(Foo)
```

PR #16714, opened 2019-09-09
Let's schedule the passive effects even earlier

sebmarkbage:
It turns out I needed to schedule mine in the mutation phase and there are also clean up life-cycles there.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -1819,7 +1819,8 @@ function commitRootImpl(root, renderPriorityLevel) {
 
 function commitBeforeMutationEffects() {
   while (nextEffect !== null) {
-    if ((nextEffect.effectTag & Snapshot) !== NoEffect) {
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1819,14 +1819,25 @@ function commitRootImpl(root, renderPriorityLevel) {
...
+    const effectTag = nextEffect.effectTag;
+    if ((effectTag & Snapshot) !== NoEffect) {
       setCurrentDebugFiberInDEV(nextEffect);
       recordEffect();
 
       const current = nextEffect.alternate;
       commitBeforeMutationEffectOnFiber(current, nextEffect);
 
       resetCurrentDebugFiberInDEV();
+    } else if ((effectTag & Passive) !== NoEffect) {
```
acdlite:
```suggestion
    }
    if ((effectTag & Passive) !== NoEffect) {
```

sebmarkbage: Nice catch.

PR #16715, opened 2019-09-09
Don't bind expiration time to render callback

acdlite:
## Based on #16663 and #16678 

This is a fragile pattern because there's only meant to be a single task per root, running at a single expiration time. Instead of binding the expiration time to the render task, or closing over it, we should determine the correct expiration time to work on using fields we store on the root object itself.

This removes the "return a continuation" pattern from the `renderRoot` function. Continuation handling is now handled by the wrapper function, which I've renamed from `runRootCallback` to `performWorkOnRoot`. That function is merely an entry point to `renderRoot`, so I've also removed the callback argument.

So to sum up, at at the beginning of each task, `performWorkOnRoot` determines which expiration time to work on, then calls `renderRoot`. And before exiting, it checks if it needs to schedule another task.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -226,6 +226,10 @@ let workInProgressRootExitStatus: RootExitStatus = RootIncomplete;
 let workInProgressRootLatestProcessedExpirationTime: ExpirationTime = Sync;
 let workInProgressRootLatestSuspenseTimeout: ExpirationTime = Sync;
 let workInProgressRootCanSuspendUsingConfig: null | SuspenseConfig = null;
+// The work left over by components that were visited during this render. Only
+// includes unprocessed updates, not work in bailed out children.
+let workInProgressRootNextUnprocessedUpdateTime: ExpirationTime = NoWork;
+
 // If we're pinged while rendering we don't always restart immediately.
 // This flag determines if it might be worthwhile to restart if an opportunity
 // happens latere.
@@ -484,21 +488,27 @@ function markUpdateTimeFromFiberToRoot(fiber, expirationTime) {
   }
 
   if (root !== null) {
-    if (workInProgressRootExitStatus === RootSuspendedWithDelay) {
-      // The root already suspended with a delay, which means this render
-      // definitely won't finish. Since we have a new update, let's mark it as
-      // suspended now, right before marking the incoming update. This has the
-      // effect of interrupting the current render and switching to the update.
-      // TODO: This happens to work when receiving an update during the render
-      // phase, because of the trick inside computeExpirationForFiber to
-      // subtract 1 from `renderExpirationTime` to move it into a
-      // separate bucket. But we should probably model it with an exception,
-      // using the same mechanism we use to force hydration of a subtree.
-      // TODO: This does not account for low pri updates that were already
-      // scheduled before the root started rendering. Need to track the next
-      // pending expiration time (perhaps by backtracking the return path) and
-      // then trigger a restart in the `renderDidSuspendDelayIfPossible` path.
-      markRootSuspendedAtTime(root, renderExpirationTime);
+    if (workInProgressRoot === root) {
+      // Received an update to a tree that's in the middle of rendering. Mark
+      // that's unprocessed work on this root.
+      markUnprocessedUpdateTime(expirationTime);
+
+      if (workInProgressRootExitStatus === RootSuspendedWithDelay) {
+        // The root already suspended with a delay, which means this render
+        // definitely won't finish. Since we have a new update, let's mark it as
+        // suspended now, right before marking the incoming update. This has the
+        // effect of interrupting the current render and switching to the update.
+        // TODO: This happens to work when receiving an update during the render
+        // phase, because of the trick inside computeExpirationForFiber to
+        // subtract 1 from `renderExpirationTime` to move it into a
+        // separate bucket. But we should probably model it with an exception,
+        // using the same mechanism we use to force hydration of a subtree.
+        // TODO: This does not account for low pri updates that were already
+        // scheduled before the root started rendering. Need to track the next
+        // pending expiration time (perhaps by backtracking the return path) and
+        // then trigger a restart in the `renderDidSuspendDelayIfPossible` path.
+        markRootSuspendedAtTime(root, renderExpirationTime);
+      }
     }
     // Mark that the root has a pending update.
     markRootUpdatedAtTime(root, expirationTime);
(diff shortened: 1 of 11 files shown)


packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -494,7 +494,7 @@ describe('ReactSuspenseWithNoopRenderer', () => {
         return <Text text="(empty)" />;
       }
       return (
-        <Suspense>
+        <Suspense fallback="Loading...">
```
acdlite: This test was testing the wrong thing, because a Suspense without a `fallback` will result in an error. So what this was actually testing was the "on error, try again at the lowest pending level."

packages/react-reconciler/src/__tests__/ReactIncrementalErrorHandling-test.internal.js
```diff
@@ -256,56 +256,6 @@ describe('ReactIncrementalErrorHandling', () => {
     expect(ReactNoop.getChildren()).toEqual([span('Everything is fine.')]);
   });
 
-  it('on error, retries at a lower priority using the expiration of higher priority', () => {
```
acdlite: Intentionally removed this behavior; I'll add another test to describe what the new semantics are.

packages/react-reconciler/src/__tests__/__snapshots__/ReactIncrementalPerf-test.internal.js.snap
```diff
@@ -422,7 +422,10 @@ exports[`ReactDebugFiberPerf warns if an in-progress update is interrupted 1`] =
 `;
 
 exports[`ReactDebugFiberPerf warns if async work expires (starvation) 1`] = `
-"⚛ (Committing Changes)
+"⚛ (React Tree Reconciliation: Completed Root)
+  ⚛ Foo [mount]
```
acdlite: This was actually wrong before, because it improperly dropped the measurements from the render phase.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -602,27 +585,43 @@ function ensureRootIsScheduled(root: FiberRoot) {
...
+// This is the entry point for every concurrent task.
+function performWorkOnRoot(root, isSync) {
+  // Determine the next expiration time to work on, using the fields stored
+  // on the root.
+  let expirationTime = getNextRootExpirationTimeToWorkOn(root);
+  if (expirationTime !== NoWork) {
+    if (expirationTime !== Sync) {
+      // Since we know we're in a React event, we can clear the current
+      // event time. The next update will compute a new event time.
+      currentEventTime = NoWork;
```
Jessidhia: Is this a React event?

(is any queued async update considered an "event"?)

sebmarkbage: The "event" here refers to a scheduled callback with the scheduler. I.e. clean stack.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1063,38 +1035,44 @@ function renderRoot(
...
-            // restarting now instead.
-            prepareFreshStack(root, expirationTime);
-            return renderRoot.bind(null, root, expirationTime);
+            const lastPingedTime = root.lastPingedTime;
+            if (lastPingedTime === NoWork || lastPingedTime >= expirationTime) {
+              // This render was pinged but we didn't get to restart earlier so
+              // try restarting now instead.
+              root.lastPingedTime = expirationTime;
+              prepareFreshStack(root, expirationTime);
+              return;
```
sebmarkbage: If `lastPingedTime === expirationTime` then I wouldn't expect there to be a callback already scheduled, right? So how does this pick up work back up?

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1063,38 +1035,44 @@ function renderRoot(
...
 
-          const nextKnownPendingLevel = root.nextKnownPendingLevel;
-          if (nextKnownPendingLevel !== NoWork) {
-            // There's lower priority work. It might be unsuspended. Try rendering
-            // at that level.
-            return renderRoot.bind(null, root, nextKnownPendingLevel);
+          const nextTime = getNextRootExpirationTimeToWorkOn(root);
+          if (nextTime !== NoWork && nextTime !== expirationTime) {
+            // There's additional work on this root.
+            return;
```
sebmarkbage: It doesn't really hurt to schedule a callback that is just going to get canceled, right? Maybe it's ok to let it schedule a callback just in case we don't get called for a while to other work, we might as well commit?

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1063,38 +1035,44 @@ function renderRoot(
...
             lastSuspendedTime !== NoWork &&
             lastSuspendedTime !== expirationTime
           ) {
-            // We should prefer to render the fallback of at the last
-            // suspended level.
-            return renderRoot.bind(null, root, lastSuspendedTime);
+            // We should prefer to render the fallback of at the last suspended
+            // level. Ping the last suspended level to try rendering it again.
+            root.lastPingedTime = lastSuspendedTime;
+            return;
```
sebmarkbage: I'm paranoid about these not having a callback scheduled and us stalling.

acdlite: I put `ensureRootIsScheduled` in the finally block of `performWorkOnRoot` to try to account for this.

I think it could theoretically happen for some of the early flush paths, which bypass `performWorkOnRoot`. But since they bypass `performWorkOnRoot`, they also won’t cause an existing callback to be canceled.

packages/react-reconciler/src/ReactFiberWorkLoop.js
acdlite: https://github.com/facebook/react/pull/16715/files/6ace553825626774d9d9325be84389fe7fe5e33a..fd253c9b0662c48f9b780c6d48bc884c99cace7f#r322551325

packages/react-reconciler/src/ReactFiberWorkLoop.js
acdlite: https://github.com/facebook/react/pull/16715/files/6ace553825626774d9d9325be84389fe7fe5e33a..fd253c9b0662c48f9b780c6d48bc884c99cace7f#r322551325

acdlite:
Regarding your paranoia that we'll neglect to schedule a callback, I opened a follow-up PR that tries to prevent a refactor hazard: https://github.com/facebook/react/pull/16743

acdlite:

PR #16717, opened 2019-09-09
[react-reconciler] Flatten responder listeners array

necolas:
This supports nested arrays of listeners, allowing people to use responder
hooks that combine several other listeners. For example:

```
function usePress(config) {
  const tap = useTap({ ... });
  const keyboard = useKeyboard({ ... });
  // ...
  return [ tap, keyboard ];
}

function MyComponent(props) {
  const press = usePress({ ... });
  return <div listeners={[press]} />
}
```

### DIFF
--- a/packages/react-reconciler/src/ReactFiberEvents.js
+++ b/packages/react-reconciler/src/ReactFiberEvents.js
@@ -92,6 +92,18 @@ function mountEventResponder(
   respondersMap.set(responder, responderInstance);
 }
 
+function flatten(array: Array<any>, result: Array<any>): Array<any> {
+  for (let i = 0; i < array.length; i++) {
+    const value = array[i];
+    if (Array.isArray(value)) {
+      flatten(value, result);
+    } else {
+      result.push(value);
+    }
+  }
+  return result;
+}
+
 function updateEventListener(
   listener: ReactEventResponderListener<any, any>,
   fiber: Fiber,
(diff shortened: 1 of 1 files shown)


sebmarkbage:
I've always regretted adding these implicit flattening ones. I feel like the listeners API isn't a settled API by itself so making it more convenient to use isn't a high priority. In fact, it might be better to add more explicit combinatorial methods so that it's easier to upgrade once we settle on something.

packages/react-reconciler/src/ReactFiberEvents.js
```diff
@@ -165,9 +177,11 @@ export function updateEventListeners(
     if (respondersMap === null) {
       respondersMap = new Map();
     }
+
     if (isArray(listeners)) {
-      for (let i = 0, length = listeners.length; i < length; i++) {
-        const listener = listeners[i];
+      const flatListeners = flatten(listeners, []);
```
trueadm: We're doing this regardless of the listeners not needing to be flattened? This is not ideal.

necolas:
OK, I can move this into `View` for now given the objections.

> it's expected that it's only hidden in implementation details of higher level components for now

We want to start moving away from this approach to letting people use the gesture handlers directly.

> it might be better to add more explicit combinatorial methods 

That's what I'm trying to avoid doing for now, but that would be the longer-term idea.

> This is expensive and adds unnecessary overhead

This analysis seems unlikely to me in this case. The arrays appear on a limited number of elements in the document, and they are both relatively short and flat. In contrast, in React Native for Web every `style` is flattened every render, and `style` is usually defined on the majority of elements with values that are both longer and far more deeply nested that here...yet it is not a perf constraint.

trueadm:
> The arrays appear on a limited number of elements in the document, and they are both relatively short and flat. In contrast, in React Native for Web every style is flattened every render, and style is usually defined on the majority of elements with values that are both longer and far more deeply nested that here...yet it is not a perf constraint.

That sounds terrible tbh :(

Issue #16462, opened 2019-08-19
DevTools: Fully disable 0.14 support


bvaughn:
It’s confusing that 0.14 or earlier is in half-working state where it displays a tree (but incorrectly). We should detect it and fully disable if it doesn’t work. Or fix it.

If we go the route of disabling support:
1. DevTools should show a warning message that clearly indicates the version of React isn't supported. (This is probably a good idea for v13 and older anyway.)
2. DevTools should not throw any errors.

---
Originally reported by @gaearon via https://github.com/bvaughn/react-devtools-experimental/issues/384

danielhusar:
I can take a look if nobody started on this one yet :)

bvaughn:
It's all yours @danielhusar. Please let me know if you decide not to work on it for any reason, and I'll remove the "taken" label 😄

danielhusar:
Is there an easy way to get React version? Or do some feature detection that is only present from `0.15` up?

I'm looking into `window.__REACT_DEVTOOLS_GLOBAL_HOOK__`, but I don't see anything useful yet.

danielhusar:
@bvaughn Hey! Is there an easy way to get react version?
I'm looking into `window.__REACT_DEVTOOLS_GLOBAL_HOOK__.renderers` but looks like version has been only added in recent react versions. Or perhaps perform some feature detection?

It's also possible I'm looking into a totally wrong place :)

bvaughn:
I think you would have to feature detect it. Here's how the legacy extension did it:
https://github.com/facebook/react-devtools/blob/d839081f79f681617df7584a8ffd2f4163fa40b9/backend/attachRenderer.js#L24-L76

Based on this, I think checking for `renderer.ComponentTree` should be a sufficient way of differentiating between v15 and older. (But you'd want to double check me by testing it.)

danielhusar:
@bvaughn I have very early PR https://github.com/facebook/react/pull/16720 , would you be able to check if I'm remotely going in right direction? 

It's my first time in codebase and I'm bit struggling. I *think* I have the check for the extension there, but I have hardcoded values for standalone | frontend, as I haven't tracked down where to put those checks yet.

I was looking into adding some logic also inside `onBridgeOperations` in the store, but operations argument is just array of numbers, which I'm not sure have all the info for this.

bvaughn:
> I was looking into adding some logic also inside `onBridgeOperations` in the store, but `operations` argument is just array of numbers, which I'm not sure have all the info for this.

[...]

PR #16720, opened 2019-09-10
WIP: Show unsupported screen in dev tools for React version < 15.0.0

danielhusar:
refs https://github.com/facebook/react/issues/16462

Screen Shot 2019-09-09 at 20 51 00

### DIFF
--- a/packages/react-devtools-core/src/standalone.js
+++ b/packages/react-devtools-core/src/standalone.js
@@ -208,7 +208,10 @@ function initialize(socket: WebSocket) {
     socket.close();
   });
 
-  store = new Store(bridge, {supportsNativeInspection: false});
+  store = new Store(bridge, {
+    supportsNativeInspection: false,
+    supportsReact: true,
+  });
 
   log('Connected');
   reload();
--- a/packages/react-devtools-extensions/src/main.js
+++ b/packages/react-devtools-extensions/src/main.js
@@ -55,8 +55,9 @@ function createPanelIfReactLoaded() {
   }
 
   chrome.devtools.inspectedWindow.eval(
-    'window.__REACT_DEVTOOLS_GLOBAL_HOOK__ && window.__REACT_DEVTOOLS_GLOBAL_HOOK__.renderers.size > 0',
-    function(pageHasReact, error) {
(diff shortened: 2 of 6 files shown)


packages/react-devtools-core/src/standalone.js
```diff
@@ -208,7 +208,10 @@ function initialize(socket: WebSocket) {
     socket.close();
   });
 
-  store = new Store(bridge, {supportsNativeInspection: false});
+  store = new Store(bridge, {
+    supportsNativeInspection: false,
+    supportsReact: true,
```
bvaughn: I think this param name is a bit confusing. React DevTools always "support React" :smile: We just want to detect an _unsupported version_.

I think this probably doesn't belong as a store capability.

packages/react-devtools-extensions/src/main.js
```diff
@@ -130,6 +131,7 @@ function createPanelIfReactLoaded() {
           isProfiling,
           supportsReloadAndProfile: isChrome,
           supportsProfiling,
+          supportsReact: renderers.get(1) && !!renderers.get(1).ComponentTree,
```
bvaughn: Similar to above comment. I don't think this belongs as a store capability, so this can be reverted 😄

packages/react-devtools-shared/src/devtools/views/Components/Components.js
```diff
@@ -17,29 +17,36 @@ import portaledContent from '../portaledContent';
...
+                </NativeStyleContextController>
+              </div>
+              <ModalDialog />
+              <SettingsModal />
             </div>
-            <ModalDialog />
-            <SettingsModal />
-          </div>
+          ) : (
+            <UnsupportedReactVersion />
```
bvaughn: I think we should use the `ModalDialogContext` to show a message if we detect an unsupported React version. We do a similar thing for React Native in [`WarnIfLegacyBackendDetected`](https://github.com/facebook/react/blob/master/packages/react-devtools-shared/src/devtools/views/WarnIfLegacyBackendDetected.js) actually. Conceptually the two are kind of the same.

danielhusar:
Thanks!
I'll take a look at `attachRenderer`.
Would you still want to mount the dev tools and show the error message in there, or perhaps just simple console.log ?

bvaughn:
Closing in favor of #16897

PR #16730, opened 2019-09-10
[react-events] Remove stopPropagation (Press) + use document for delegation

trueadm:
This PR does two things:

- Removes the `stopPropagation` prop from the Press responder module. After internal inspection, this prop seems to have been misused internally because a misunderstanding on what it does. It thus makes sense to remove this prop, because it was always intended to be a temporary hack.
- Ensure that we always listen to captures on the `document`. Given we no longer can use `stopPropagation` we no longer have the issue of blocking other event systems already on the `document`.

### DIFF
--- a/packages/react-events/src/dom/Press.js
+++ b/packages/react-events/src/dom/Press.js
@@ -113,9 +113,7 @@ const DEFAULT_PRESS_RETENTION_OFFSET = {
 const targetEventTypes = hasPointerEvents
   ? [
       'keydown_active',
-      // We need to preventDefault on pointerdown for mouse/pen events
-      // that are in hit target area but not the element area.
-      'pointerdown_active',
+      'pointerdown',
       'click_active',
     ]
   : ['keydown_active', 'touchstart', 'mousedown', 'click_active'];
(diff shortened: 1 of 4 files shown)


necolas:
> Given we no longer can use stopPropagation we no longer have the issue of blocking other event systems already on the document.

Any "responder" can still do `nativeEvent.stopPropagation()`

trueadm:
@necolas True, but this was an explicit prop, so maybe my wording was a bit off. People can always do anything in the responders themselves.

necolas:
The point is that moving listeners back to the document means that the system can collide with external listeners also attached to the document. That's aside from remove the `stopPropagation` prop from Press, which we should definitely do.

trueadm:
@necolas I found a bunch of subtle bugs today when moving all to `body`. Previously, we accidentally assigned most events to `body` but with some on `document`. This meant that events like `scroll` worked on the document but don't work on body.

packages/react-events/src/dom/Press.js
```diff
@@ -132,8 +131,6 @@ const rootEventTypes = hasPointerEvents
       'touchcancel',
       // Used as a 'cancel' signal for mouse interactions
       'dragstart',
-      // We listen to this here so stopPropagation can
-      // block other mouseup events used internally
       'mouseup_active',
```
necolas: This doesn't need to be active anymore. Same for L118

PR #16732, opened 2019-09-10
[react-interactions] Add TabFocusContainer and TabbableScope UI components

trueadm:
This PR adds two experimental components to be used for internal focus management handling. Furthermore, this PR fixes a bug with scope component refs correctly detaching.

`TabFocusContainer` contains tabbable keyboard focus to within its children, where tabbing gets wrapped to the beginning.

`TabbableScope` collects host components that are keyboard tabbable so that `TabFocusContainer` can handle the tab navigation through said components.

### DIFF
--- a/packages/react-dom/src/client/focus/__tests__/TabFocusContainer-test.internal.js
+++ b/packages/react-dom/src/client/focus/__tests__/TabFocusContainer-test.internal.js
@@ -7,28 +7,11 @@
  * @flow
  */
 
+import {createEventTarget} from 'react-events/src/dom/testing-library';
+
 let React;
 let ReactFeatureFlags;
 let TabFocusContainer;
-
-const createTabForward = type => {
-  const event = new KeyboardEvent('keydown', {
-    key: 'Tab',
-    bubbles: true,
-    cancelable: true,
-  });
-  return event;
-};
-
-const createTabBackward = type => {
-  const event = new KeyboardEvent('keydown', {
-    key: 'Tab',
-    shiftKey: true,
-    bubbles: true,
-    cancelable: true,
-  });
-  return event;
-};
 
 describe('TabFocusContainer', () => {
   beforeEach(() => {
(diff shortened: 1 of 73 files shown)


necolas:
FWIW I still don't like the "tab" nomenclature given that these nodes can be reached by means other than pressing the Tab key on a keyboard.

trueadm:
@necolas Maybe it should be better named as `TabContain` or something then, as this is exclusively for that operation only.

necolas:
> this is exclusively for that operation only

so this component wouldn't work with e.g. a directional pad on a remote control?

packages/react-dom/src/client/focus/__tests__/TabFocusContainer-test.internal.js
```diff
@@ -0,0 +1,242 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+let React;
+let ReactFeatureFlags;
+let TabFocusContainer;
+
+const createTabForward = type => {
```
necolas: You could use the testing-library helpers for this

packages/react-dom/src/client/focus/__tests__/TabFocusContainer-test.internal.js
```diff
@@ -0,0 +1,242 @@
...
+          <button ref={buttonRef} />
+          <div ref={divRef} tabIndex={0} />
+          <input ref={input2Ref} tabIndex={-1} />
+          <button ref={butto2nRef} />
+        </TabFocusContainer>
+      );
+
+      ReactDOM.render(<Test />, container);
+      inputRef.current.focus();
+      document.activeElement.dispatchEvent(createTabForward());
```
necolas: You could use the testing-library helpers for this

trueadm: Sure. This was mostly a copy and paste from the old focus scope tests

PR #16738, opened 2019-09-10
[Fresh] Hash signatures

gaearon:
Fresh Babel plugin emits "signatures" that are strings containing names of Hooks and call metadata. For example:

```js
export default function App() {
  _s();

  const [foo, setFoo] = useState(0);
  React.useEffect(() => {});
  return <h1>{foo}</h1>;
}

_s(App, "useState{[foo, setFoo](0)}\\nuseEffect{}"); // <-- signature
```

Note the signature contains `[foo, setFoo](0)`. This ensures that if we change the `foo` name, or if we change the initial state (`0`), we force a remount. Including the initial argument source code is a special case for `useState` and `useReducer` since that usually signals an intentional attempt to remount.

Unfortunately, it's not always safe to emit an arbitrary string in some environments. For example, in www we have a transform that replaces `cx("foo")` with some string hash *regardless of whether it's inside a string or not*. It's just a regex. So if an expression like this becomes a part of a signature, www transforms can produce a syntax error.

To fix this, we can default to always emitting hashes for signatures. They're guaranteed to be short, and can't cause these kinds of issues. I left a fallback for ASTExplorer and for our own snapshot tests. (However, our integration tests verify the other branch still works.)

### DIFF
--- a/packages/react-refresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-refresh/src/ReactFreshBabelPlugin.js
@@ -7,7 +7,7 @@
 
 'use strict';
 
-export default function(babel, opts) {
+export default function(babel, opts = {}) {
   if (typeof babel.getEnv === 'function') {
     // Only available in Babel 7.
     const env = babel.getEnv();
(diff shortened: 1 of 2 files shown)


packages/react-refresh/src/ReactFreshBabelPlugin.js
```diff
@@ -300,7 +300,20 @@ export default function(babel, opts) {
       }
     });
 
-    const args = [node, t.stringLiteral(key)];
+    let finalKey = key;
+    if (typeof require === 'function' && !opts.emitFullSignatures) {
```
threepointone: (Only because I've been looking for examples for the elvis operator, feel free to ignore)
```suggestion
    if (typeof require === 'function' && !opts?.emitFullSignatures) {
```
Then you wouldn't need the default arguments for `opts`

gaearon: Wanna make sure I won't mess it up in other places where it's ready

PR #16751, opened 2019-09-11
[react-interactions] More Tab Focus control handling

trueadm:
This is a follow up PR to the previous initial work to handle basic tab focus control with `TabFocusController.js`s.

This PR adds a few more features to the experimental scope API and, most importantly, provides an internal imperative mechanism for controlling tab focus via an API. I've also made containment an optional prop on `TabFocusController`s and provided a way of exposing an imperative ref to scopes so these can easily be handed back via scope traversal.

In my next PR I plan on demonstrating taking this future and building components that allow for a powerful Grid based keyboard focusable system (using arrows + tab + modifiers).


packages/react-dom/src/client/focus/TabFocusController.js
```diff
@@ -0,0 +1,180 @@
...
+  getNextController: () => null | ControllerHandle,
+  getPreviousController: () => null | ControllerHandle,
+|};
+
+const {useImperativeHandle, useRef} = React;
+
+export const TabFocusController = React.forwardRef(
+  ({children, contain}: TabFocusControllerProps, ref): React.Node => {
+    const scopeRef = useRef(null);
+    const keyboard = useKeyboard({onKeyDown, preventKeys: ['tab']});
```
necolas: tab -> Tab

packages/react-dom/src/client/focus/TabFocusController.js
```diff
@@ -0,0 +1,180 @@
...
+      focusFirst,
+      focusNext,
+      focusPrevious,
+      getNextController,
+      getPreviousController,
+    };
+
+    useImperativeHandle(ref, () => controllerHandle);
+
+    function onKeyDown(event: KeyboardEvent): boolean {
```
necolas: Let's move this into the useKeyboard call

packages/react-dom/src/client/focus/TabFocusController.js
```diff
@@ -0,0 +1,180 @@
...
+
+    const controllerHandle: ControllerHandle = {
+      focusFirst,
+      focusNext,
+      focusPrevious,
+      getNextController,
+      getPreviousController,
+    };
+
+    useImperativeHandle(ref, () => controllerHandle);
```
necolas: Note this hook is not good for perf according to my tests

trueadm: No other way to do what we want here, other than using a class. :/

packages/react-dom/src/client/focus/TabFocusController.js
```diff
@@ -0,0 +1,181 @@
...
+type KeyboardEventType = 'keydown' | 'keyup';
+
+type KeyboardEvent = {|
+  altKey: boolean,
+  ctrlKey: boolean,
+  isComposing: boolean,
+  key: string,
+  location: number,
+  metaKey: boolean,
+  repeat: boolean,
```
necolas: AFAIK, `repeat` is unreliable. We could ditch `location`, `repeat`, and maybe even `isComposing` from these events

packages/react-dom/src/client/focus/TabFocusController.js
```diff
@@ -0,0 +1,181 @@
...
+        if (event.key !== 'Tab') {
+          return true;
+        }
+        if (event.shiftKey) {
+          return focusPrevious();
+        } else {
+          return focusNext();
+        }
+      },
+      preventKeys: ['Tab'],
```
necolas: I think you probably also need `[ 'Tab', { shiftKey: true } ]`

packages/react-dom/src/client/focus/TabFocusController.js
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    function getPreviousController(): null | ControllerHandle {
+      const tabbableScope = scopeRef.current;
+      const allScopes = tabbableScope.getChildrenFromRoot();
+      if (allScopes === null) {
+        return null;
+      }
+      const currentScopeIndex = allScopes.indexOf(tabbableScope);
+      if (currentScopeIndex === -1 || currentScopeIndex === 0) {
```
necolas: `<= 0`?

packages/react-dom/src/client/focus/__tests__/TabFocusController-test.internal.js
```diff
@@ -221,5 +215,44 @@ describe('TabFocusContainer', () => {
...
+          </TabFocusController>
+          <TabFocusController ref={secondFocusControllerRef}>
+            <input tabIndex={-1} />
+            <div ref={divRef} tabIndex={0} />
+          </TabFocusController>
+        </div>
+      );
+
+      ReactDOM.render(<Test />, container);
+      firstFocusControllerRef.current.focusFirst();
```
necolas: The tests are a bit easier to read if we avoid writing out `ref.current` each time:

```
const button = buttonRef.current;
const firstFocusController = firstFocusControllerRef.current;

firstFocusController.focusFirst();
expect(document.activeElement).toBe(button);
firstFocusController.focusNext();
expect(document.activeElement).toBe(button2);
```

PR #16752, opened 2019-09-11
Moved backend injection to the content script

onionymous:
Moved the backend injection logic to the content script. Previously this was done by using `chrome.devtools.inspectedWindow.eval()`:

```
const source = `
  // the prototype stuff is in case document.createElement has been modified
  (function () {
    var script = document.constructor.prototype.createElement.call(document, 'script');
    script.src = "${scriptName}";
    script.charset = "utf-8";
    document.documentElement.appendChild(script);
    script.parentNode.removeChild(script);
  })()
  `;

  chrome.devtools.inspectedWindow.eval(source, function(response, error)
...
```

Since this code runs in the context of the inspected page, it is subject to the page's CSP restrictions and will cause a [Trusted Types](https://github.com/WICG/trusted-types) violation. Content scripts, however, are not subject to the page's restrictions and can manipulate the DOM. This change will prevent a TrustedScriptURL violation on clients that have Trusted Types support.

### DIFF
--- a/packages/react-devtools-extensions/src/main.js
+++ b/packages/react-devtools-extensions/src/main.js
@@ -138,14 +138,10 @@ function createPanelIfReactLoaded() {
         chrome.devtools.inspectedWindow.eval(
           `window.postMessage({ source: 'react-devtools-inject-backend' });`, 
           function(response, error) {
-          if (error) {
-            console.log(error);
+            if (error) {
+              console.log(error);
+            }
           }
-
-          if (typeof done === 'function') {
-            done();
-          }
-        }
         );
 
         const viewElementSourceFunction = createViewElementSource(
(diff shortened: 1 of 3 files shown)


packages/react-devtools-extensions/src/main.js
```diff
@@ -4,7 +4,6 @@ import {createElement} from 'react';
 import {unstable_createRoot as createRoot, flushSync} from 'react-dom';
 import Bridge from 'react-devtools-shared/src/bridge';
 import Store from 'react-devtools-shared/src/devtools/store';
-import inject from './inject';
```
bvaughn: This was the only use of the `inject.js` module, so we can delete that now.

bvaughn:
Merged via 8f03109

PR #16755, opened 2019-09-11
[react-events] Tap support for virtual click

necolas:
A couple of issues:

1. This implementation can't be used to produce visual feedback as the entire sequence is flushed for the same event. We could wrap the "end" phase in a setTimeout to support this.

2. ~Changing the `targetEventTypes` to use `click` instead of `click_active` results in no `click` event being dispatched to `onRootEvent`, even though `rootEventTypes` contains `click_active`.~

### DIFF
--- a/packages/react-events/src/dom/Tap.js
+++ b/packages/react-events/src/dom/Tap.js
@@ -23,6 +23,7 @@ import {
   dispatchUserBlockingEvent,
   getTouchById,
   hasModifierKey,
+  isVirtualClick,
 } from './shared';
 
 type TapProps = $ReadOnly<{|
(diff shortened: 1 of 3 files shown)


trueadm:
> Changing the targetEventTypes to use click instead of click_active results in no click event being dispatched to onRootEvent, even though rootEventTypes contains click_active.

I can address this bug tomorrow. :)

necolas:
I tweaked the patch so that it doesn't need root click events.

I also realised that `useKeyboard` needs to call `preventDefault()` on `click` events that keyboard interactions might produce (e.g., on links), and that knowing when a "click" occurred might also be useful. Related, Microsoft added an `onClick` prop to RN's View specifically for keyboards. So I wonder if `useKeyboard` could add an `onClick` prop which could also be used to communicate back about these programmatic / screen-reader clicks instead of lumping that into `Tap`

necolas:
Closed in favor of #16780

Issue #16742, opened 2019-09-10
DevTools: Excessively long component names cause component tree to flatten


coreyleelarson:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
Our application uses an older version of Mobx, and its `inject` HOC creates excessively long component names (inject-Component-with-userStore-appStore-columnStore-...) when multiple stores are used. While navigating and scrolling down the component tree, when a components scrolls into frame whose name cannot fit horizontally, the entire tree flattens. This makes it difficult to discern between parent, sibling, and child components.

![devtools](https://user-images.githubusercontent.com/1900645/64652970-2606bc00-d3f3-11e9-8bcd-45ad2c43dd19.PNG)

**What is the expected behavior?**
I would expect the component tree to either scroll horizontally or truncate component names that are excessively long, while still including indentation for child components.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
React v16.3.0
Chrome v77
DevTools v4.0.6

9oelM:
Seems like a problem with `--indentation-size` variable used for css in DevTools. 
Will need a bit of research, but the problem should be somewhere around this.

## When you have `--indentation-size` set as 0px in parent html node
![image](https://user-images.githubusercontent.com/22465806/64674305-76d9eb80-d4ab-11e9-8e44-7dd0885e202f.png)

## When you drop `--indentation-size`
![image](https://user-images.githubusercontent.com/22465806/64674289-69246600-d4ab-11e9-86d1-ad7d9d7d0ab7.png)

You can see that removing that buggy `--indentation-size` css variable would make it look as expected.

9oelM:
Related:
https://github.com/facebook/react/blob/b438699d3620bff236282b049204e1221b3689e9/packages/react-devtools-shared/src/devtools/views/Components/Tree.js#L406

coreyleelarson:
@9oelM - I see. This appears to be where that variable is used:

https://github.com/facebook/react/blob/b438699d3620bff236282b049204e1221b3689e9/packages/react-devtools-shared/src/devtools/views/Components/Element.js#L120

What if we ensure `--indentation-size` is always greater than or equal to something like 5px?

bvaughn:
I don't know that supporting names like the ones shown in the attached screenshot (hundreds of characters long) is really a goal for DevTools.

[...]

PR #16756, opened 2019-09-11
Adds a minimum indentation size for components with extremely long names.

coreyleelarson:
Fixes https://github.com/facebook/react/issues/16742

In extreme cases, components with names that exceed the DevTools window width cause the dynamic indentation calculation to equal 0, flattening the entire component hierarchy. When this happens, it is difficult to discern between parent, sibling, and child components. This PR adds a minimum indentation size to ensure the tree is still visually usable in extreme cases, while preserving the regular dynamic indentation functionality under normal circumstances.

**Before:**

before

**After:**

after

**Happy Path:**

normal

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/Components.css
+++ b/packages/react-devtools-shared/src/devtools/views/Components/Components.css
@@ -11,7 +11,6 @@
 
 .TreeWrapper {
   flex: 0 0 65%;
-  overflow: auto;
 }
 
 .SelectedElementWrapper {
--- a/packages/react-devtools-shared/src/devtools/views/Components/Element.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/Element.js
@@ -117,7 +117,7 @@ export default function ElementView({data, index, style}: Props) {
         style={{
           // Left offset presents the appearance of a nested tree structure.
           // We must use padding rather than margin/left because of the selected background color.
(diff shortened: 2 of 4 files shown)


bvaughn:
This doesn't look sufficient. Deep trees would expand horizontally past the viewport and leave unscrollable white space.

coreyleelarson:
@bvaughn - You're absolutely right, I had not considered that scenario.

Removing the `.List { overflow-x: hidden !important }` seems to be the likely first step, however because `react-window` renders the container/elements with fixed widths and absolute positioning, the Active/Inactive background highlighting does not behave as desired.

 I'll have to think on this a bit when I've got more time.

Thanks for the feedback!

bvaughn:
Another thing to keep in mind is that re-enabling horizontal scroll means that we would need to restore measuring on-mount and remembering the widest measured item so that horizontal scroll didn't grow-and-shrink-and-grow as you scrolled vertically. (That is a bad UX.)

Here's how we used to do this:
https://github.com/bvaughn/react-devtools-experimental/commit/49f6a22b5b8a1b7d5e2a774262357e2992ad0ca0#diff-483c2df20fea61bc9557723ea45b633b

coreyleelarson:
Closing this pull request until I can rethink the implementation based on @bvaughn comments above. Wouldn't want this to be accidentally merged.

PR #16776, opened 2019-09-13
Support disabling interaction tracing for suspense promises

bvaughn:
If a thrown Promise has the `__reactDoNotTraceInteractions` attribute, React will not wrapped its callbacks to continue tracing any current interaction(s).

Example usage:
```js
function SomeComponent() {
  try {
    const value = SomeCache.read(/*...*/);
  } catch (maybePromise) {
    if (typeof value === "object" && typeof value.then === "function") {
      maybePromise.__reactDoNotTraceInteractions = true;
    }
    throw maybePromise;
  }
}
```

### DIFF
--- a/packages/react-reconciler/src/ReactFiberThrow.js
+++ b/packages/react-reconciler/src/ReactFiberThrow.js
@@ -174,7 +174,7 @@ function attachPingListener(
       renderExpirationTime,
     );
     if (enableSchedulerTracing) {
-      if ((thenable: any).__reactDoNotTraceInteractions !== true) {
+      if (thenable.__reactDoNotTraceInteractions !== true) {
         ping = Schedule_tracing_wrap(ping);
       }
     }
--- a/packages/react-reconciler/src/ReactFiberThrow.js
+++ b/packages/react-reconciler/src/ReactFiberThrow.js
@@ -174,7 +174,9 @@ function attachPingListener(
       renderExpirationTime,
     );
     if (enableSchedulerTracing) {
-      ping = Schedule_tracing_wrap(ping);
+      if ((thenable: any).__reactDoNotTraceInteractions !== true) {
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberThrow.js
```diff
@@ -174,7 +174,9 @@ function attachPingListener(
       renderExpirationTime,
     );
     if (enableSchedulerTracing) {
-      ping = Schedule_tracing_wrap(ping);
+      if (thenable.__reactDoNotTraceInteractions !== true) {
+        ping = Schedule_tracing_wrap(ping);
```
sebmarkbage: The ping part is a bit weird because it means that means that if we’re only suspended on this promise for the first commit I believe it will decrement and complete the interaction *before* we commit.

I think this only happens if this also happened to cause a “Delayed” suspended state (bad loading) since otherwise we don’t suspend the commit and it’ll continue until it commits.

However fixing this is tricky since we’d have to keep track of this promise and call it once we commit to detach.

bvaughn: > The ping part is a bit weird because it means that means that if we’re only suspended on this promise for the first commit I believe it will decrement and complete the interaction before we commit.

Hm... I don't think this is the case, but maybe you can point out what I'm overlooking? We only decrement in `finishPendingInteractions`, which runs after commit.

sebmarkbage: Another way to put it is, if this is ok, then it should be ok to remove this wrapper in all cases.

Because this ping is only relevant when there is an in progress render. If the in progress render keeps its own interactions alive then we never need this.

bvaughn: Hm...yeah, maybe wrapping the ping listener is never necessary.

PR #16779, opened 2019-09-13
[react-events] Keyboard calls preventDefault on 'click' events

necolas:
Make sure to call preventDefault for any 'click' events that follow a 'keydown'
event that matches 'preventKeys'.

### DIFF
--- a/packages/react-events/src/dom/Keyboard.js
+++ b/packages/react-events/src/dom/Keyboard.js
@@ -16,26 +16,32 @@ import React from 'react';
 import {DiscreteEvent} from 'shared/ReactTypes';
 import type {ReactEventResponderListener} from 'shared/ReactTypes';
 
-type KeyboardEventType = 'keydown' | 'keyup';
+type KeyboardEventType = 'keyboard:keydown' | 'keyboard:keyup';
 
-type KeyboardProps = {
+type KeyboardProps = {|
   disabled?: boolean,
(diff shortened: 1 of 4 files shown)


packages/react-events/src/dom/Keyboard.js
```diff
@@ -150,6 +156,7 @@ function createKeyboardEvent(
     isComposing,
     key: getEventKey(nativeEvent),
     metaKey,
+    pointerType: 'keyboard',
```
trueadm: Do we really need a pointerType for keyboard responder?

packages/react-events/src/dom/Keyboard.js
```diff
@@ -16,26 +16,32 @@ import React from 'react';
 import {DiscreteEvent} from 'shared/ReactTypes';
 import type {ReactEventResponderListener} from 'shared/ReactTypes';
 
-type KeyboardEventType = 'keydown' | 'keyup';
+type KeyboardEventType = 'keyboard:keydown' | 'keyboard:keyup';
```
trueadm: I guess the keyboard prefix is because we're also adding the keyboard pointer type. What do you expect this to expand to in the future?

necolas: I think we should scope event names to their responder, especially given that `keydown` and `keyup` are also the names of native events

PR #16780, opened 2019-09-13
[react-events] Keyboard support for virtual clicks

necolas:
This accounts for all clicks that are natively dispatched following relevant
keyboard interactions (e.g., key is "Enter"), as well as programmatic clicks,
and screen-reader virtual clicks.

(Alternative to https://github.com/facebook/react/pull/16755)

### DIFF
--- a/packages/react-events/src/dom/shared/index.js
+++ b/packages/react-events/src/dom/shared/index.js
@@ -70,3 +70,17 @@ export function hasModifierKey(event: ReactDOMResponderEvent): boolean {
     altKey === true || ctrlKey === true || metaKey === true || shiftKey === true
   );
 }
+
+// Keyboards, Assitive Technologies, and element.click() all produce "virtual"
+// clicks that do not include coordinates and "detail" is always 0 (where
+// pointer clicks are > 0).
+export function isVirtualClick(event: ReactDOMResponderEvent): boolean {
+  const nativeEvent: any = event.nativeEvent;
+  return (
+    nativeEvent.detail === 0 &&
+    nativeEvent.screenX === 0 &&
+    nativeEvent.screenY === 0 &&
+    nativeEvent.clientX === 0 &&
+    nativeEvent.clientY === 0
+  );
+}
--- a/packages/react-events/src/dom/Keyboard.js
+++ b/packages/react-events/src/dom/Keyboard.js
@@ -11,15 +11,20 @@ import type {
   ReactDOMResponderEvent,
   ReactDOMResponderContext,
 } from 'shared/ReactDOMTypes';
+import type {ReactEventResponderListener} from 'shared/ReactTypes';
 
 import React from 'react';
 import {DiscreteEvent} from 'shared/ReactTypes';
-import type {ReactEventResponderListener} from 'shared/ReactTypes';
+import {isVirtualClick} from './shared';
 
-type KeyboardEventType = 'keyboard:keydown' | 'keyboard:keyup';
+type KeyboardEventType =
+  | 'keyboard:click'
+  | 'keyboard:keydown'
+  | 'keyboard:keyup';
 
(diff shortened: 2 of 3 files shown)


necolas:
I think this is probably the better place for supporting these kinds of click events, as keyboard interactions also produce "virtual" clicks and you could argue the screen-reader created clicks are also "keyboard" triggered. Since most of the data in the click event is empty (just `0`), it fits better with the keyboard event type than the pointer event type and doesn't need to produce a complete "tap" sequence to preserve API expectations of `Tap`.

packages/react-events/src/dom/shared/index.js
```diff
@@ -70,3 +70,17 @@ export function hasModifierKey(event: ReactDOMResponderEvent): boolean {
...
+export function isVirtualClick(event: ReactDOMResponderEvent): boolean {
+  const nativeEvent: any = event.nativeEvent;
+  return (
+    nativeEvent.detail === 0 &&
+    nativeEvent.screenX === 0 &&
+    nativeEvent.screenY === 0 &&
+    nativeEvent.clientX === 0 &&
+    nativeEvent.clientY === 0
+  );
+}
```
craigkovatch: (Not sure what the etiquette on a merged PR is -- please let me know if these comments would be more appropriate in e.g. a new Issue.)

I don't believe this works cross-browser.

1) in IE(11), detail is always `0`, even for mouse-initiated click events. Can check for `!nativeEvent.pointerType` to work around this, since IE11 dispatches PointerEvents for these.

2) in Safari, the screen and client coordinates are present even for a keyboard-initiated click. The client coordinates seem to default to the "center" of the element.

What I've used internally at Tableau for this that works cross-browser is `return (nativeEvent.detail === 0 && !nativeEvent.pointerType);`

trueadm: This is really awesome feedback. Thank you cc @necolas seems like a valid thing we should be doing too.

necolas: How does `(nativeEvent.detail === 0 && !nativeEvent.pointerType)` work if that evaluates as `true` for mouse events as well as virtual clicks?

craigkovatch: My investigation showed that `nativeEvent.detail === 0` is sufficient for all evergreen browsers; adding `!nativeEvent.pointerEvent` covers the IE11 case. Is that more clear?

necolas: I still don't follow. The `click` event is of type `MouseEvent` which has no `pointerType` property. Are you saying IE11's `click` event is of type `PointerEvent`?

craigkovatch: > Are you saying IE11's click event is of type PointerEvent?
Yes, exactly

necolas: Thanks for confirming; we didn't know about that quirk. We'll be sure to incorporate your suggested fix. I like what Safari does here. I'm also curious to hear what you use this check for at Tableau

necolas: https://github.com/facebook/react/pull/16915

craigkovatch: > I'm also curious to hear what you use this check for at Tableau

It's a long, annoying story -- the short version is there an event normalizer (parallel to React) sitting on top of touch events that interferes with native keyboard click events on button elements. I needed this to differentiate the event sources so it could ignore those from the keyboard.

craigkovatch: @necolas little bit of unfortunate news -- it appears this will incorrectly eval to `true` for click events which are dispatched to an `<input>` when clicked on a wrapping `<label>`. Not sure if this is a problem for your purposes, but wanted to give you a heads-up.

PR #16781, opened 2019-09-13
[Scheduler] Prevent event log from growing unbounded

acdlite:
If a Scheduler profile runs without stopping, the event log will grow unbounded. Eventually it will run out of memory and the VM will throw an error.

To prevent this from happening, let's automatically stop the profiler once the log exceeds a certain limit. We'll also print a warning with advice to call `stopLoggingProfilingEvents` explicitly.

### DIFF
--- a/packages/scheduler/src/SchedulerProfiling.js
+++ b/packages/scheduler/src/SchedulerProfiling.js
@@ -44,7 +44,9 @@ if (enableProfiling) {
   profilingState[CURRENT_TASK_ID] = 0;
 }
 
-const INITIAL_EVENT_LOG_SIZE = 1000;
+// Bytes per element is 4
+const INITIAL_EVENT_LOG_SIZE = 131072;
+const MAX_EVENT_LOG_SIZE = 524288; // Equivalent to 2 megabytes
 
 let eventLogSize = 0;
 let eventLogBuffer = null;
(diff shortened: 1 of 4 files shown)


packages/scheduler/src/SchedulerProfiling.js
```diff
@@ -65,10 +67,16 @@ function logEvent(entries) {
     const offset = eventLogIndex;
     eventLogIndex += entries.length;
     if (eventLogIndex + 1 > eventLogSize) {
-      eventLogSize = eventLogIndex + 1;
-      const newEventLog = new Int32Array(
-        eventLogSize * Int32Array.BYTES_PER_ELEMENT,
-      );
+      eventLogSize *= 2;
+      if (eventLogSize > MAX_EVENT_LOG_SIZE) {
+        console.error(
```
acdlite: This message will increase the size of the profiling build. Seems fine? Idk.

scripts/babel/transform-prevent-infinite-loops.js
```diff
@@ -36,10 +36,18 @@ module.exports = ({types: t, template}) => {
     visitor: {
       'WhileStatement|ForStatement|DoWhileStatement': (path, file) => {
         const filename = file.file.opts.filename;
-        const MAX_ITERATIONS =
-          filename.indexOf('__tests__') === -1
-            ? MAX_SOURCE_ITERATIONS
-            : MAX_TEST_ITERATIONS;
+        const maxIterations = t.logicalExpression(
+          '||',
+          t.memberExpression(
```
acdlite: @gaearon Might want to review the changes I made in here. I wanted to increase the limit for a specific test so I could trigger the memory error.

acdlite:
Ok I'm going to land as-is though because I don't have any better ideas for how to test it right now, and I'd rather have a slow test than no test

PR #16784, opened 2019-09-14
[eslint] strip tailing property in assignments

### LINKED ISSUE #15510: [ESLint] Assignment like foo.bar.baz = X should warn about foo.bar instead
```js
  let foo = {}
  useEffect(() => {
    foo.bar.baz = 43;
  }, []);
```

This asks you to include `foo.bar.baz` into deps. But this doesn't make sense, as you write to it. Instead it should ask to include `foo.bar` into array.

If `foo` is a newly created object in render scope then we might additionally nudge you to `useRef` for it instead. Because otherwise it won't be shared between renders.

Zzzen:
closes #15510

>``` 
>let foo = {}
>useEffect(() => {
>   foo.bar.baz = 43;
>}, []);
>```
>This asks you to include foo.bar.baz into deps. But this doesn't make sense, as you write to it. Instead it should ask to include foo.bar into array.

### DIFF
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -1437,20 +1437,7 @@ function getDependency(node) {
     )
   ) {
     return getDependency(node.parent);
-  } else {
-    return stripTailingPropInAssignment(node);
-  }
-}
-
-/**
- * Assuming () means the passed/returned node:
- * (props) => (props)
- * (props.foo) => (props.foo)
- * (props.foo.bar) = X => (props.foo).bar = X
- * (props.foo.bar.baz) = X => (props.foo.bar).baz
- */
-function stripTailingPropInAssignment(node) {
-  if (
+  } else if (
     node.type === 'MemberExpression' &&
     node.parent &&
     node.parent.type === 'AssignmentExpression'
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -1438,6 +1438,25 @@ function getDependency(node) {
   ) {
     return getDependency(node.parent);
   } else {
+    return stripTailingPropInAssignment(node);
+  }
+}
+
+/**
+ * Assuming () means the passed/returned node:
+ * (props) => (props)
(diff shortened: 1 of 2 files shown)


gaearon:
Can you please rebase? We moved to the suggestions API so the test needs tweaks.

Zzzen:
👌 I'll try this weekend.

Zzzen:
Moved to the suggestions API. PTAL

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -1437,6 +1437,25 @@ function getDependency(node) {
     )
   ) {
     return getDependency(node.parent);
+  } else {
+    return stripTailingPropInAssignment(node);
```
gaearon, line 1441: What makes you think it's an assignment in this branch? Can we make it more specific and throw on something unexpected?

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -1437,6 +1437,25 @@ function getDependency(node) {
...
+}
+
+/**
+ * Assuming () means the passed/returned node:
+ * (props) => (props)
+ * (props.foo) => (props.foo)
+ * (props.foo.bar) = X => (props.foo).bar = X
+ * (props.foo.bar.baz) = X => (props.foo.bar).baz
+ */
+function stripTailingPropInAssignment(node) {
```
gaearon, line 1452: Just inline in this function into the parent one.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
Zzzen: 😅  `tryStrippingTrailingPropsInAssigment` could have been a better choice. I'm trying to enumerate all cases and confused by a test:

```js
function MyComponent(props) {
  useEffect(() => {
    if (props.foo.onChange) {
      props.foo.onChange();
    }
  }, []);
}
```

Why is `props.foo` a proper dependency? Why not `props.foo.onChange`?

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
Zzzen: 👌

gaearon:
Thanks.

PR #16794, opened 2019-09-16
[react-ui] Move experimental event+a11y work to react-ui package

trueadm:
This PR renamed the `react-events` package to `react-ui` and adds a top-level directory for each category. The `react-events` work goes into a top-level `events` directory, whilst all the accessibility component experiments go into `accessibility`. This also allows us to move relevant other parts of our experimental UI work into React Interactions, and other projects, into other top-level directories of `react-ui` without having to make them fit into existing directory structures (which is getting awkward).


trueadm:
@bvaughn We might decide to move this to the React package in the future, but for now this is all more of a package of experimental work we're playing around with internally. I do quite like `react-ui` though :P

packages/react-ui/accessibility/src/FocusGrid.js
```diff
@@ -7,11 +7,11 @@
  * @flow
  */
 
-import type {KeyboardEvent} from 'react-events/src/dom/Keyboard';
+import type {KeyboardEvent} from '../../events/src/dom/Keyboard';
```
necolas: Maybe these paths should also be `react-ui/events...` as that makes it easier to avoid inlining dependencies in the bundles

packages/react-ui/accessibility/src/FocusGrid.js
```diff
@@ -7,11 +7,11 @@
  * @flow
  */
 
-import type {KeyboardEvent} from 'react-events/src/dom/Keyboard';
+import type {KeyboardEvent} from 'react-ui/events/src/dom/Keyboard';
 
 import React from 'react';
 import {tabFocusableImpl} from './TabbableScope';
-import {useKeyboard} from 'react-events/keyboard';
+import {useKeyboard} from '../../events/keyboard';
```
necolas: Oops this is the non-type import that could get inlined

packages/react-ui/accessibility/src/ReactTabFocus.js
```diff
@@ -8,11 +8,11 @@
  */
 
 import type {ReactScopeMethods} from 'shared/ReactTypes';
-import type {KeyboardEvent} from 'react-events/src/dom/Keyboard';
+import type {KeyboardEvent} from 'react-ui/events/src/dom/Keyboard';
 
 import React from 'react';
 import {TabbableScope} from './TabbableScope';
-import {useKeyboard} from 'react-events/keyboard';
+import {useKeyboard} from '../../events/keyboard';
```
necolas: Here too

trueadm: This is okay. I plan on doing a follow up where I add the a11y components to build and will revise it there.

packages/react-ui/accessibility/src/FocusGrid.js
trueadm: This is okay. I plan on doing a follow up where I add the a11y components to build and will revise it there.

bvaughn:
I think this PR may have broken DevTools tests. Can you run ` yarn test-build-devtools`?

trueadm:
I had a similar issue when deleting the "react-events" directory locally. Git doesn't register directory removals as a change, so maybe we should update the script to check if the directory is empty?

PR #16795, opened 2019-09-16
Follow-up to initial Trusted Types support

gaearon:
This addresses a few things I've noticed since merging #16157.

* At FB we might want the flag to be dynamic while we're experimenting (e.g. depending on the user). We'd like to avoid overhead on every call for people who aren't opted in. So I want to check for TT flag early and branch based on that instead of doing it in `toStringOrTrustedType`.
  - My change sets `toStringOrTrustedType` to `toString` directly, and then adds the TT checks on top when the flag is on. I've inlined `isTrustedTypesValue` into that branch since it wasn't useful by itself.
* I've added a global to Flow to remove the suppressions.
* I'm destructuring `isHTML` and friends early to avoid property access hit in the hot path. I'm assuming `trustedTypes` methods don't care about `this` values. This might make mocking a bit more difficult but IMO this makes sense because we decide early on whether to use them or not. As opposed to TT support potentially "turning on" later in the app. This does mean, however, that TT needs to be initialized before ReactDOM to be acknowledged.
* I've revamped the tests so that they actually verify that we pass the TT objects to the DOM sinks. I've checked that replacing every `toStringOrTrustedType` call with `'' + value` now fails a particular test so we're covering them all.
* I've restructured `setInnerHTML` a bit to remove the duplicate `innerHTML` assignment which is easy to accidentally mess up. The special path now has an early return.

### DIFF
--- a/packages/react-dom/src/client/ToStringValue.js
+++ b/packages/react-dom/src/client/ToStringValue.js
@@ -38,22 +38,6 @@ export function getToStringValue(value: mixed): ToStringValue {
   }
 }
 
-/**
- * Returns true only if Trusted Types are available in global object and the value is a trusted type.
- */
-let isTrustedTypesValue: (value: any) => boolean;
-// $FlowExpectedError - TrustedTypes are defined only in some browsers or with polyfill
-if (enableTrustedTypesIntegration && typeof trustedTypes !== 'undefined') {
-  isTrustedTypesValue = (value: any) =>
-    trustedTypes.isHTML(value) ||
-    trustedTypes.isScript(value) ||
-    trustedTypes.isScriptURL(value) ||
-    // TrustedURLs are deprecated and will be removed soon: https://github.com/WICG/trusted-types/pull/204
-    (trustedTypes.isURL && trustedTypes.isURL(value));
-} else {
-  isTrustedTypesValue = () => false;
-}
-
 /** Trusted value is a wrapper for "safe" values which can be assigned to DOM execution sinks. */
 export opaque type TrustedValue: {toString(): string, valueOf(): string} = {
   toString(): string,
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/client/ToStringValue.js
```diff
@@ -67,15 +51,27 @@ export opaque type TrustedValue: {toString(): string, valueOf(): string} = {
...
-  }
+export let toStringOrTrustedType: any => string | TrustedValue = toString;
+if (enableTrustedTypesIntegration && typeof trustedTypes !== 'undefined') {
+  const isHTML = trustedTypes.isHTML;
+  const isScript = trustedTypes.isScript;
+  const isScriptURL = trustedTypes.isScriptURL;
+  // TrustedURLs are deprecated and will be removed soon: https://github.com/WICG/trusted-types/pull/204
+  const isURL = trustedTypes.isURL ? trustedTypes.isURL : value => false;
+  toStringOrTrustedType = value => {
+    if (typeof value === 'string') {
```
gaearon: I guess numbers should also go here.

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -67,15 +51,27 @@ export opaque type TrustedValue: {toString(): string, valueOf(): string} = {
...
-    // fast-path string values as it's most frequent usage of the function
-    typeof value !== 'string' &&
-    isTrustedTypesValue(value)
-  ) {
-    return value;
-  } else {
-    return '' + value;
-  }
+export let toStringOrTrustedType: any => string | TrustedValue = toString;
+if (enableTrustedTypesIntegration && typeof trustedTypes !== 'undefined') {
```
sebmarkbage: This seems like it would make this function not inlinable if the flag is on or dynamic which isn’t great because it used to be a simple `’’ +` that gets inlined.

packages/react-dom/src/client/__tests__/trustedTypes-test.internal.js
```diff
@@ -1,42 +1,219 @@
...
 
-  it('should not stringify trusted values', () => {
-    const trustedObject = {toString: () => 'I look like a trusted object'};
-    class Component extends React.Component {
-      state = {inner: undefined};
-      render() {
-        return <div dangerouslySetInnerHTML={{__html: this.state.inner}} />;
-      }
+  it('should not stringify trusted values for dangerouslySetInnerHTML', () => {
+    const ttObject1 = {
```
Siegrift: I would move this to beforeEach as it's being used in a few places.

packages/react-dom/src/client/ToStringValue.js
gaearon: That’s actually got me looking at it in the first place — it wasn’t getting inlined with a dynamic flag in FB bundle. Lemme think about this more.

gaearon: If the flag is _on_, it's already not being inlined, even before this PR. Probably because the body is bigger. Since we're doing a function call either way, I think a *smaller* function (as in this PR) would be better than a larger one. It's also a single call (in this PR) rather than two calls (as in before).

gaearon: I'm going to get this in because it's not being inlined either way with the flag on. If we want to fix this, we'll have to inline it manually at concrete callsites.

sebmarkbage: Why do we toString in user space instead of letting the browser do it again?

Seems like we only need it in some IE legacy thing and/or for the `javascript:` URL check so we should only ever do this at all in those cases.

gaearon: It's for IE9 I think? We could feature-test it I guess.

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -67,15 +51,21 @@ export opaque type TrustedValue: {toString(): string, valueOf(): string} = {
...
+  const isHTML = trustedTypes.isHTML;
+  const isScript = trustedTypes.isScript;
+  const isScriptURL = trustedTypes.isScriptURL;
+  // TrustedURLs are deprecated and will be removed soon: https://github.com/WICG/trusted-types/pull/204
+  const isURL = trustedTypes.isURL ? trustedTypes.isURL : value => false;
+  toStringOrTrustedType = value => {
+    if (
+      typeof value === 'object' &&
+      (isHTML(value) || isScript(value) || isScriptURL(value) || isURL(value))
```
koto: This fails with Illegal invocation error (is* functions need to be bound to the factory i.e. window.[tT]rustedTypes.

Issue #16685, opened 2019-09-06
Doesn't support profiler even with latest React version


iamgbayer:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
The profiler tab only shows this message even though I have a version greater than 16.5.

![Screen Shot 2019-09-06 at 13 26 47](https://user-images.githubusercontent.com/5483459/64444238-0d706c00-d0aa-11e9-8d32-810465a13433.png)

Something strange on the console is this message and just appears when the extension is active.

![Screen Shot 2019-09-06 at 13 25 17](https://user-images.githubusercontent.com/5483459/64444170-e74acc00-d0a9-11e9-8b7c-acaa31e50cea.png)

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
React **16.9.0**
React DevTools **4.0.6-a39d9c3**

bvaughn:
> The profiler tab only shows this message even though I have a version greater than 16.5.

Profiling requires either a DEV build or [the production+profiling build](http://fb.me/react-profiling). Are you running one of these?

iamgbayer:
I tested with two applications one using CRA and another using a custom Webpack config, both running on DEV build

Both return the same error.

threepointone:
Could you make a git repo that demonstrates this problem? It sounds like a configuration problem, but it's hard to comment without seeing the actual project.

bvaughn:
Yeah. We've tested a lot on this end, with several common setups (including CRA), and it works - so without having more info from your end on how to repro this, it's hard to help.

Just tested again to double check:
```js
create-react-app test-profiler
cd test-profiler
yarn start
```

Profiler works fine. The issue template asks:
> Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?

Could you clarify which OS, browser, and browser versions you're using please?

iamgbayer:
I tested with this repo below using CRA, but I think that be something on the application because of testing only with `create-react-app test-profiler` and starting the application, there was no problem.

This repo has the issue @threepointone, https://github.com/rentinga/client

I'm using an OSX High Sierra with Chrome 76.

LetItRock:
Hi! @bvaughn @iamgbayer 

I've found what was causing an issue here. In the sample application there is a code which does this: 
[...]

PR #16798, opened 2019-09-16
[react-devtools-shared] Added string type check for object name prop in getDisplayName function

LetItRock:
Fixes issue #16685

I've added a small fix in ```getDisplayName``` function from ```utils.js``` file for the case when ```type``` argument doesn't have a ```displayName``` property, but has ```name``` property with incorrect value ex. object. That case means that developer who is using react has in the code something like this: 
```
const Icon = () => {...}
Icon.name = { foo: 'bar' };
```
DevTools relay on displayName/name properties when displaying component names in the tree. 
```getDisplayName``` function requires displayName/name properties to be a string values, otherwise it will use fallback name - 'Anonymous'. 

Tests are included.
You can find longer description of what was wrong in the issue.

### DIFF
--- a/packages/react-devtools-shared/src/utils.js
+++ b/packages/react-devtools-shared/src/utils.js
@@ -46,17 +46,15 @@ export function getDisplayName(
     return nameFromCache;
   }
 
-  let displayName;
+  let displayName = fallbackName;
 
   // The displayName property is not guaranteed to be a string.
   // It's only safe to use for our purposes if it's a string.
   // github.com/facebook/react-devtools/issues/803
   if (typeof type.displayName === 'string') {
     displayName = type.displayName;
-  }
-
-  if (!displayName) {
-    displayName = type.name || fallbackName;
+  } else if (typeof type.name === 'string') {
+    displayName = type.name;
   }
 
   cachedDisplayNames.set(type, displayName);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rs-pavlo** [comment]
I have completed the CLA.

**Comment by LetItRock** [comment]
I've completed the CLA.

File: packages/react-devtools-shared/src/utils.js [hunk 1]
```diff
@@ -46,17 +46,15 @@ export function getDisplayName(
...
   // The displayName property is not guaranteed to be a string.
   // It's only safe to use for our purposes if it's a string.
   // github.com/facebook/react-devtools/issues/803
   if (typeof type.displayName === 'string') {
     displayName = type.displayName;
-  }
-
-  if (!displayName) {
-    displayName = type.name || fallbackName;
+  } else if (typeof type.name === 'string') {
```
bvaughn: The previous truthy check also prevented returning empty strings (for anonymous functions). This test would fail after this change:
```js
expect(getDisplayName(() => {})).toEqual('Anonymous');
```
I think the check should be:
```
typeof type.name === 'string' && type.name !== ''
```

rs-pavlo: yes, you right :( It would be enough to add one more test...

LetItRock: yes, you right :( It would be enough to add one more test...

bvaughn: No worries, I added it~

Issue #16753, opened 2019-09-11
RFC: Warning System Revamp


gaearon:
This is a proposal to change how the internal warning system works.

**Note: @walaura is already working on this, please don't send PRs.**

## Current System

React has a concept of "warnings". Conceptually, most of them should be treated as errors. They indicate bugs. For example, not fixing a "key" warning can result in very bad issues in production. For these "warnings", the only difference from a real error is that they don't throw and the checks are removed in production. Because they're expensive to do.

React warnings ultimately become `console.error` calls. In the source, they are expressed as `warning(cond, message, ...args)`. If the `cond` is false, the warning gets printed.

By default, calling `warning` will print a `console.error` with the current component stack appended at the end. Sometimes, we may not want the component stack. Maybe the warning is aggregated from many components (e.g. a StrictMode violation), and the stack is not relevant. In that case we have `warningWithoutStack`. It has the same API as `warning`, but doesn't append the stack. In fact, `warning(format, ...args)` is internally implemented as `warningWithoutStack(format + '%s', ...args, stack)`.

We also have a lesser-known `lowPriorityWarning` module. Unlike `warning` which uses `console.error`, `lowPriorityWarning` uses `console.warn`. It is "lower severity" (appears yellow in console), and we currently use it only for deprecation messages.

## Problems

* The naming is confusing. `warning`s are conceptually errors, and use `console.error`, but their name doesn't reflect that.
* The `warning(cond, ...)` API is confusing. It is easy to forget whether `cond` is supposed to be `true` or `false` for the warning to fire. (Answer: it fires on `false`.) Due to this confusion, a bunch of callsites just do `if (!cond) warning(false, ...)` to avoid thinking about this.
* `warning` and `lowPriorityWarning` have different default behavior. `warning` appends component stack by default, but `lowPriorityWarning` doesn't. This makes it difficult to "downgrade" a `warning` to `lowPriorityWarning` because we'd lose the stack and have to manually append it.
* Some `warning`s don't actually represent "errors" in practice. They have too high severity. It's not a huge deal in the console. But if you start hooking up the console to richer mechanisms (e.g. an error dialog), the difference becomes more annoying. You want to clearly separate *what's broken today* from *what may break tomorrow*.

## Ideal End State

* All React warnings are audited and split into two severities: `error` (stuff that is likely broken today) and `warn` (stuff that may break in the future).
  - This means we'll likely "downgrade" some mostly advisory warnings.
[...]

PR #16799, opened 2019-09-16
Warning system refactoring (part 1)

Jessidhia:
Step 1 of the refactoring proposed in https://github.com/facebook/react/issues/16753

It's pretty much the minimum for now, to make it easier to merge. The `toLowPriorityWarnDev` matcher usage checks still mention the `toWarnDev` name, though. It can be fixed by making the matcher name also an argument to the factory, or the matcher could be forked instead of using a factory.

### DIFF
--- a/packages/shared/lowPriorityWarning.js
+++ b/packages/shared/lowPriorityWarning.js
@@ -1,0 +1,32 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import lowPriorityWarningWithoutStack from 'shared/lowPriorityWarningWithoutStack';
+import ReactSharedInternals from 'shared/ReactSharedInternals';
+
+/**
+ * Similar to invariant but only logs a warning if the condition is not met.
+ * This can be used to log issues in development environments in critical
+ * paths. Removing the logging code for production environments will keep the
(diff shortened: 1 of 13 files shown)


scripts/print-warnings/print-warnings.js
```diff
@@ -53,7 +53,7 @@ function transform(file, enc, cb) {
           if (
             callee.isIdentifier({name: 'warning'}) ||
             callee.isIdentifier({name: 'warningWithoutStack'}) ||
-            callee.isIdentifier({name: 'lowPriorityWarning'})
+            callee.isIdentifier({name: 'lowPriorityWarningWithoutStack'})
```
gaearon: We should catch both variations?

packages/shared/lowPriorityWarning.js
```diff
@@ -5,47 +5,27 @@
...
-    }
-    if (!condition) {
-      printWarning(format, ...args);
+    if (condition) {
+      return;
     }
+    const ReactDebugCurrentFrame = ReactSharedInternals.ReactDebugCurrentFrame;
+    const stack = ReactDebugCurrentFrame.getStackAddendum();
+    // eslint-disable-next-line react-internal/warning-and-invariant-args
+    lowPriorityWarningWithoutStack(false, format + '%s', ...args, stack);
```
gaearon: This fails the warning extraction script that fails here trying to strintify it. How did we solve that in the warning module?

gaearon:
Lgtm, thanks!

PR #16804, opened 2019-09-17
[react-a11y] Add react-ui/accessibility to bundle build

trueadm:
In order to test the experimental accessibility components internally at FB, we need to be able to build the latest components so they get included in our syncs. This includes an accessibility bundle so we can test this out internally. Also, I made sure that our mappings for FB externals gets re-written appropriately, as our internal system won't recognise a `react-ui` package.


necolas:
Why are the accessibility-related components being bundled together rather than separately?

trueadm:
@necolas I'll split them up. Actually, if you could review my other PR then I can land that first, as that renames the files, it makes more sense.

PR #16806, opened 2019-09-17
[react-ui] FocusGrid -> ReactFocusTable + tweaks and fixes

trueadm:
This PR is the continuation of work for the experimental React Accessibility components. Notably, this PR changes:

- Rename `FocusGrid` -> `ReactFocusTable` to better reflect that it handles keyboard interactions (arrow keys) with layout matching that of a static table.
- Adds support for getting the props from a scope.
- Fixes a bunch of Flow issues
- Adds a `onNavigateOut` prop that can be used on FocusTables to control the behavior when a user keyboard interactions out of a table's bounds. The callback provides the `direction` and a helper method to select another table of the same type.


necolas:
I think "navigate" implies a change of screen and url, rather than focus exiting the component

trueadm:
@necolas Agreed, I'll change the summary of the PR to reflect.

Issue #16087, opened 2019-07-09
[Umbrella] Memory Leaks


sebmarkbage:
This issue is a summary of issues mentioned in https://github.com/facebook/react/pull/15157.

There are many different ways to create memory leaks with React since we give you access to the imperative power to do so. Most should be dealt with by clean up / unmount functions.

Some could be pure React bugs. Some could be related to the lack of clean up of render phase effects. Others could be related to leaks that exists but the way React works makes them larger than they otherwise would've.

# Actionable

I think there are at least two actionable patterns to address from #15157:

- [ ] If a handle on a DOM node is leaked, it takes the React tree with it. This is a fairly easy mistake to make and the effect is pretty high. What we would do here is special case DOM nodes with refs on them, and always detach their back pointer to the React Fiber, if it was ever fully mounted. We currently traverse these trees anyway when they get deleted. We want to stop doing this for most things but for nodes with a ref it seems minor to special case since they typically need to be invoked with null anyway.
- [ ] Investigate the source of the leak in https://github.com/jonnycornwell/potential_react_leak and fix the source.

# Unresolved

- [ ] Closing over setState/dispatch or class component instances to global state can leak too. Does this pattern warrant special casing too? Under what conditions?
- [ ] What patterns are actually covered by https://github.com/facebook/react/pull/15157? It can cut down on a potentially larger leak but how big is it really?
- [ ] What other issues remain after solving the actionable above? Let's make another pass investigating if people's original issues remain.

# Won't Fix

- Side-effects in class constructor, componentWillMount, componentWillReceiveProps, componentWillUpdate, getDerivedStateFrom... and render that store a reference to anything stateful outside React won't be able to clean up. This is [documented](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) in the 16.3 release and is a major design decision as part of the concurrent/suspense/error handling strategy.
- Effects/state retained temporarily in alternate fiber likely won't be fixed short term. This is due to how Fiber uses two trees and swaps between them. This can lead to additional values being retained until that tree gets some work on it to swap again. This was observed in the Hooks release and there are some confusing cases where a destroy function can hold onto more memory than expected in the closure. Typically this is solved by using a custom hook since that gets its own scope/closure.
- Props/child trees retained by alternate children. Similarly, children that was just removed can sometimes between retained by the alternate copy of that. That is until that node gets another update on it which clears out the old children. These cases are fairly unusual and fix themselves eventually as the app lives on.

[...]

PR #16807, opened 2019-09-17
[react-core] Clear more properties in detachFiber

trueadm:
This PR aims at tackling some potentially retained memory after a fiber gets detached. To do this, we clear down more fields on the fiber to be `null` in `detachFiber`.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -893,6 +893,7 @@ function commitNestedUnmounts(
 }
 
 function detachFiber(current: Fiber) {
+  const alternate = current.alternate;
   // Cut off the return pointers to disconnect it from the tree. Ideally, we
   // should clear the child pointer of the parent alternate to let this
   // get GC:ed but we don't know which for sure which parent is the current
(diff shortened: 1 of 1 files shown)


gaearon:
I think @sebmarkbage's point in https://github.com/facebook/react/pull/16115 and later https://github.com/facebook/react/issues/16087 is that we shouldn't play whack-a-mole with trying to clear as much as we can since it has a cost, and often masks the actual underlying issues. For each field clearing which "helps" we need to find a minimal reproducing case that demonstrates a leak.

sebmarkbage:
In particular, the question would be, this fiber itself has been deleted. So if it’s still retain, what is holding onto to it? However in this case we actually have the very answer to that question in the comment above in the code. It’s an architectural quirks that leaves this node alive for a bit longer so we clear it.

I believe we actually do have a way to clear the right parent now so we might be able to do that instead.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -903,13 +904,14 @@ function detachFiber(current: Fiber) {
   current.memoizedState = null;
   current.updateQueue = null;
   current.dependencies = null;
-  const alternate = current.alternate;
+  current.sibling = null;
+  current.alternate = null;
```
sebmarkbage: I believe this shouldn’t matter to set.

trueadm:
@gaearon I'd also not play whack-a-mole, but as per Sebastian's above comment, the reason we have to do this to begin with is because of the architectural quirks. If we have a better solution, then I'm all ears. :)

packages/react-reconciler/src/ReactFiberCommitWork.js
trueadm: It's not needed but it reduces the amount of code quite a bit by re-using the function (which should be hot). I can go back to the previous way if you feel it's better.

sebmarkbage:
When we add deleted fibers to the effect list. We know their parent’s work in progress. If we set the `.return` at that point then we’ll know which parent was the work in progress. Then we could do `current.return.child = null` here.

trueadm:
@sebmarkbage Makes sense. Shall we close this PR in favor of that approach? We can roll with this change until we land that change as a follow up.

sebmarkbage:
Yea let’s land this first in case that other approach doesn’t work.

PR #16809, opened 2019-09-17
[scheduler][profiler] Start time of delayed tasks

acdlite:
Fixes a bug in the Scheduler profiler where the start time of a delayed tasks is always 0.


packages/scheduler/src/Scheduler.js
```diff
@@ -91,7 +91,7 @@ function advanceTimers(currentTime) {
       timer.sortIndex = timer.expirationTime;
       push(taskQueue, timer);
       if (enableProfiling) {
-        markTaskStart(timer);
+        markTaskStart(timer, currentTime);
```
acdlite: Maybe it's time to add Flow back to this file :D

PR #16812, opened 2019-09-17
expose isHydrating on host config/ReactDOM

threepointone:
As part of an experiment, this PR moves `isHydrating` from ReactFiberHydrationContext to a ref on the host configs that support hydration (ie: react-dom and react-noop). It also exposes the same ref on react-dom's "internal" export. We will remove this export in the near future.


gaearon:
Why not expose a function on the reconciler instead? Like findHostInstance.

threepointone:
I could do that, will change.

threepointone:
closing this, @lunaruan is having a go at it.

PR #16821, opened 2019-09-18
[React Native] Fix for view config registrations

rickhanlonii:
## Overview
This change fixes a behavior in the React Native view config registry which would hide errors when registering components.

## Details
Before this change, even if we failed to register the view config for any reason (e.g. the events are invalid), we would still clear the callback for the view config.

This means, the next time that a user renders the component and we check for the view config, we would show this error:

> View config not found for name Slider

This error is wrong. The real error is masked (and may never be shown, if the registration happened before error handling is set up), and may be something like:

> Event cannot be both direct and bubbling: topChange

This PR fixes this behavior by clearing the view config callback only after the view config is successfully registered.

## Screen
Screen Shot 2019-09-18 at 2 28 04 PM

## Test

- Added test that fails without this change

### DIFF
--- a/packages/react-native-renderer/src/__tests__/ReactNativeEvents-test.internal.js
+++ b/packages/react-native-renderer/src/__tests__/ReactNativeEvents-test.internal.js
@@ -75,6 +75,49 @@ beforeEach(() => {
     .ReactNativeViewConfigRegistry.register;
 });
 
+it('fails to register the same event name with different types', () => {
+  const InvalidEvents = createReactNativeComponentClass('InvalidEvents', () => {
+    if (!__DEV__) {
+      // Simulate a registration error in prod.
+      throw new Error('Event cannot be both direct and bubbling: topChange');
+    }
+
+    // This view config has the same bubbling and direct event name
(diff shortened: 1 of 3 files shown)


packages/react-native-renderer/src/__tests__/ReactNativeEvents-test.internal.js
```diff
@@ -75,6 +75,49 @@ beforeEach(() => {
...
+  // we attempt to register the view config and fail.
+  expect(() => ReactNative.render(<InvalidEvents />, 1)).toThrow(
+    'Event cannot be both direct and bubbling: topChange',
+  );
+
+  // Continue to re-register the config and
+  // fail so that we don't mask the above failure.
+  expect(() => ReactNative.render(<InvalidEvents />, 1)).toThrow(
+    'Event cannot be both direct and bubbling: topChange',
+  );
```
rickhanlonii: Without the change in this PR, this test will fail because the view config is not found

Screen Shot 2019-09-18 at 3 30 12 PM

PR #16825, opened 2019-09-18
ESLint: exhaustive-deps rule doesn't catch useCallback issue

kentcdodds:
This is a bug report in the form of a PR. I could add the fix, but I just want to make sure that I'm not misunderstanding something. This test fails because the code doesn't give any warnings, but it should.

Specifically this should give a warning:

```javascript
const debouncedCb = useCallback(debounce(cb, {wait: 300}), []);
```

More context: https://twitter.com/kentcdodds/status/1174382572052180992

Let me know if you'd like me to continue with fixing this or if I'm missing something.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -4707,6 +4707,26 @@ const tests = {
           'Either include it or remove the dependency array.',
       ],
     },
+    {
+      code: `
+        import debounce from './debounce'
+
+        function Example({ cb }) {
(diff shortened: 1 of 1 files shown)


kentcdodds:
Honestly, this one's kinda tricky and I wouldn't be surprised if there were a case where we get false positives after "fixing" this one.

gaearon:
Should it just force you to always use the arrow form?

kentcdodds:
That would solve potential problems, though I have a feeling people would find that pretty annoying. I'm torn. I'd love to hear what others have to say about this.

gaearon:
For now I made it error and ask it to write it inline.

gaearon:
(Alternatively, you can declare it outside the call.)

PR #16833, opened 2019-09-19
[react-ui] Add preventDefault+ to Keyboard + update a11y components

trueadm:
This PR is a sort of spiritual follow up to https://github.com/facebook/react/pull/16822. Here are the changes outlined in this PR:

- Keyboard responder now provides `preventDefault` and `continuePropagation` to the user event. Keyboard still does not propagate to other keyboard responders by default, but provides a method to allow it to continue and extra level. I had to update tests respectfully, as we previous used the return value from the event (which was a bit confusing).
- Keyboard responder now does not `preventDefault` any events by default. Instead, the user can optionally control this by calling `preventDefault` like already done on the web. The `Press` responder has been updated to ensure that the same standard behavior exists as before.
- The accessibility components have been updated to properly make use of this new functionality.

### DIFF
--- a/packages/react-ui/accessibility/src/FocusTable.js
+++ b/packages/react-ui/accessibility/src/FocusTable.js
@@ -146,7 +146,7 @@ export function createFocusTable(): Array<React.Component> {
   function Cell({children}): FocusCellProps {
     const scopeRef = useRef(null);
     const keyboard = useKeyboard({
-      onKeyDown(event: KeyboardEvent): boolean {
+      onKeyDown(event: KeyboardEvent): void {
         const currentCell = scopeRef.current;
         switch (event.key) {
           case 'UpArrow': {
@@ -162,7 +162,7 @@ export function createFocusTable(): Array<React.Component> {
                 }
               }
             }
-            return false;
+            return;
           }
           case 'DownArrow': {
             const [cells, rowIndex] = getRowCells(currentCell);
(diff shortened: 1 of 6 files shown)


necolas:
Not really convinced about `event.continuePropagation()` rather than `event.stopPropagation()`. Why not have keyboard events propagate unless otherwise prevented from doing so by the callback handling a key?

trueadm:
@necolas Simply because it means reworking the core event system to have a way of stopping and starting propagation, which then confuses the action with the web equivalent. We won't actually be stopping native propagation here, but overwhelmingly that's what people thought it was doing when we had the declarative prop internally.

TBH I'd rather both `preventDefault` and `stopPropagation` were named differenty, so people knew it was different from how they already understand these methods working on React and the web today.

necolas:
> it means reworking the core event system to have a way of stopping and starting propagation

Before we had `return false` mean "stop propagation". I'd rather we keep that than invert everything and have to call `continuePropagation`

> We won't actually be stopping native propagation here, but overwhelmingly that's what people thought it was doing when we had the declarative prop internally.

Bike-shedding of names aside, people shouldn't expect that calling `stopPropagation` on an object (or returning false) is going to operate on a native DOM event when the callback isn't directly related to a DOM event in the first place.

trueadm:
Let's go with `stopPropagation` then. I think this would be a good one for you to involved in (I'll reach out directly).

Issue #16475, opened 2019-08-19
React DevTools: Welcome screen keeps showing up


proteriax:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
“Welcome to the new React DevTools!” message blocks the devtool panel every time the it is opened.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**
1. Open a website with React with DevTools installed.
2. Open the Component tab.
3. Dismiss the welcome screen.
4. Close the devtools and open it again.

**What is the expected behavior?**
Dismissing the “Welcome to the new React DevTools!” message should be permanent.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
DevTools: 4.0.5
Chrome: 77.0.3865.35

bvaughn:
Do you, by chance, have web storage disabled?

proteria:
I do block third party cookie access. Are DevTools preferences not stored at the extension storage itself?

bvaughn:
DevTools _does_ store preferences in the extension itself, but the `localStorage` API is disabled (for the extension) when web storage is disabled. This was a surprise to me, and something I reported as Chromium bug.

Unfortunately the lack of a synchronous API for preferences- combined with some timing concerns in order to support features like reload-and-profile- makes the `localStorage` API pretty appealing for lightweight preferences (including "has seen" state like this).

Pretty unfortunate that it causes the Welcome overlay to always reappear though :( That's no good.

proteria:
That’s surprising. `localStorage` works for my extensions even with web storage disabled.

bvaughn:
PRs are welcome if you'd like to dig in and lend a hand. :)

proteria:
Where’s the source code? It says it’s migrated too this repo but I don’t see it anywhere.

proteria:
Update: I fixed it by explicitly allowing `chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/` to store data under `chrome://settings/content/cookies`

bvaughn:
> Where’s the source code? It says it’s migrated to this repo but I don’t see it anywhere.

It's still a work in progress, PR #16381

Glad you found a workaround for now!

[...]

PR #16834, opened 2019-09-19
[DevTools] Remove Welcome dialog

gaearon:
"Fixes" https://github.com/facebook/react/issues/16475 by removing the welcome screen, as it's been enough time for people to auto-upgrade. It still sucks that there are cases where prefs aren't persisted, but let's at least remove the dialog.

Test plan:

Screen Shot 2019-09-19 at 2 31 26 PM


bvaughn:
I had imagined we'd keep this around for other potential major feature additions (just disabled via props) but we can always re-add it.

PR #16842, opened 2019-09-20
Rename experimental react-ui => react-interactions

gaearon:
Neither is intended as an actual name. It's a placeholder.

But ReactUI is something we might actually wanna use in the future. Using it as a codename now can make it more difficult to use later "for real". As we know, codenames leak and get a life of their own.

Is `react-interactions` the perfect name? No. But that's the point. If we started using it as a codename, let's keep it that way. And I guess it's OK to put windowing stuff there too until we figure out another place for it.


sophiebits:
Thanks for saving the reactui name!

Issue #16832, opened 2019-09-19
eslint-plugin-react-hooks: 'Hook is being called conditionally' error outside condition


cbdeveloper:
**Do you want to request a *feature* or report a *bug*?**

BUG (possibly)

**What is the current behavior?**

The plugin is showing this error:

>React Hook "useState" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return? (react-hooks/rules-of-hooks)eslint

But I don't think I'm calling any hooks conditionally.

The code:

https://codesandbox.io/s/exciting-bhabha-mqj7q

```
function App(props) {
  const someObject = { propA: true, propB: false };

  for (const propName in someObject) {
    if (propName === true) {
      console.log("something");
    } else {
      console.log("whatever");
    }
  }

  // THE PLUGIN ERROR MSG ON THIS useState
  const [myState, setMyState] = useState(null);

  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      <h2>Start editing to see some magic happen!</h2>
    </div>
  );
}
```

![image](https://user-images.githubusercontent.com/43407798/65239956-10bb1d00-dad8-11e9-88b9-ec517cd17645.png)

**What is the expected behavior?**

The plugin wouldn't show the error in this situation.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

![image](https://user-images.githubusercontent.com/43407798/65240081-5a0b6c80-dad8-11e9-8692-a9974342130e.png)

dankreiger:
I'm not deep enough into the source code yet to understand why, but I noticed that for this particular example:

`pathsFromStartToEnd` is `3`
`allPathsFromStartToEnd` is `2`

`possiblyHasEarlyReturn` is also `true`

This particular error is thrown when these values are unequal: see [source](https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L404)

M-Izadmehr:
@gaearon, I created an MR addressing and fixing this issue (#16853).

The main reason this happens is that the value of `allPathsFromStartToEnd` is calculated wrong.
In this example, the number of paths from start to end should be 3, but the function returns 2.

The reason this happens is because of flawed caching of cyclic paths. So in normal graph traversing, we need the path history as well, but it only takes a list of cyclic elements, and the elements leading to this path are not known. 
[...]

PR #16853, opened 2019-09-22
[eslint-plugin-react-hooks] Fix cyclic caching for loops containing a…

M-Izadmehr:
This MR Fixes a bug (#16832) in the `eslint-plugin-react-hooks`, where caching of cycles is performed wrong in some cases.

**Example:** https://codesandbox.io/s/exciting-bhabha-mqj7q

**Description**
If the user uses an `if` statement inside a `for ... of obj` statement, and uses a correct hook afterwards, the eslint shows an *error* that:
`React Hook "${hook}" is called conditionally. React Hooks must be called in the exact same order in every component render.`

**Reports**
I guess that it is already known issue because, inside the [ESLintRulesOfHooks.js - L562](https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js#L562), it is already known that in some cases caching of cyclic paths in code path, might result in the wrong errors.

Also (#16832) reports this case too.

**Reason**
Eslint uses `CodePathSegment` which is similar to doubly linked list. Based on the `CodePathSegment` AST different types of cyclic segments might happen. In the special case of [ForInStatement](https://eslint.org/docs/developer-guide/code-path-analysis#forinstatement) there are two paths from obj identifier to the next segment. In such a graph, in order to traverse all the paths from start to end, we need to not only remember the current segment, but also the whole path. 

The current code uses a simple caching just by name of segments (and not the list of segment history) as a result, whenever it faces a cycle, it breaks the cycle by returning 0. As a result, there might be several conditions, where the segment might have cyclic and not cyclic paths to end. In the current code, if the first path is cyclic all the non-cyclic ones are thrown away. And depending on the starting point and endpoint of AST, we might have different results.

For example for the demo code, the result of  [RulesOfHooks.js - L404](https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L291), the value of all `allPathsFromStartToEnd` is calculated as `2` instead of `3`

**Solution**
We need caching of responses and also saving cyclic paths for later logic, In the new code, I added a different caching mechanism, which also remembers the path history too. The effect of the performance is still negligible because even for [the most difficult test](https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js#L293) including `1099511627776` paths from start to end of the code, it only increases the time about 0.3s on MacBookPro.

### DIFF
--- a/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
+++ b/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
@@ -114,31 +114,33 @@ export default {
          * Populates `cyclic` with cyclic segments.
          */
 
-        function countPathsFromStart(segment) {
+        function countPathsFromStart(segment, pathHistory) {
           const {cache} = countPathsFromStart;
           let paths = cache.get(segment.id);
+          let pathList = new Set(pathHistory);
 
-          // If `paths` is null then we've found a cycle! Add it to `cyclic` and
-          // any other segments which are a part of this cycle.
-          if (paths === null) {
-            if (cyclic.has(segment.id)) {
-              return 0;
-            } else {
-              cyclic.add(segment.id);
-              for (const prevSegment of segment.prevSegments) {
-                countPathsFromStart(prevSegment);
-              }
-              return 0;
+          // If `pathList` includes the current segment then we've found a cycle!
+          // We need to fill `cyclic` with all segments inside cycle
+          if (pathList.has(segment.id)) {
+            const pathArray = [...pathList];
+            const cyclicSegments = pathArray.slice(
+              pathArray.indexOf(segment.id) + 1,
+            );
+            for (const cyclicSegment of cyclicSegments) {
+              cyclic.add(cyclicSegment)
             }
+
+            return 0;
           }
+
+          // add the current segment to pathList
+          pathList.add(segment.id);
 
           // We have a cached `paths`. Return it.
           if (paths !== undefined) {
             return paths;
           }
 
-          // Compute `paths` and cache it. Guarding against cycles.
-          cache.set(segment.id, null);
           if (codePath.thrownSegments.includes(segment)) {
             paths = 0;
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
```diff
@@ -114,39 +114,41 @@ export default {
...
-          if (paths === null) {
-            if (cyclic.has(segment.id)) {
-              return 0;
-            } else {
-              cyclic.add(segment.id);
-              for (const prevSegment of segment.prevSegments) {
-                countPathsFromStart(prevSegment);
-              }
-              return 0;
+          let pathList = new Set(pathHistory);
```
kidkkr: pathList doesn't seem to be reallocated.

```suggestion
          let pathList = new Set(pathHistory);
```

packages/eslint-plugin-react-hooks/src/RulesOfHooks.js
```diff
@@ -183,43 +185,45 @@ export default {
...
           } else {
             paths = 0;
             for (const nextSegment of segment.nextSegments) {
-              paths += countPathsToEnd(nextSegment);
+              paths += countPathsToEnd(nextSegment, pathList);
             }
           }
-          cache.set(segment.id, paths);
 
+          cache.set(segment.id, paths);
```
kidkkr: I think unnecessary diff in here.

M-Izadmehr: I am not sure why, but I am not sure why it considers this line as changed:)

TrySound:
@gaearon This problem is still relevant.

gaearon:
Thanks!

M-Izadmehr:
I almost forgot about this :D

PR #16863, opened 2019-09-23
Expose shared/getComponentName as React.getComponentName

koba04:
`getComponentName` is a handy helper function to get a component name for 3rd party libraries and custom renderers.

To get a proper component name, we have to deal with React internals like `shared/getComponentNam` does.
So I think it's beneficial for 3rd party developers to expose the function.
This PR exposes the API as `React.getComponentName` but it's ok to expose the API as an npm package like `react-component-name` instead of adding a function into `React`.

### DIFF
--- a/packages/react/src/React.js
+++ b/packages/react/src/React.js
@@ -13,6 +13,7 @@ import {
   REACT_SUSPENSE_TYPE,
   REACT_SUSPENSE_LIST_TYPE,
 } from 'shared/ReactSymbols';
+import getComponentName from 'shared/getComponentName';
 
 import {Component, PureComponent} from './ReactBaseClasses';
 import {createRef} from './ReactCreateRef';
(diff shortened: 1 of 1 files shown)


acdlite:
Hey thanks for the PR!

We don't want to expose any reflection APIs because it locks us into internal implementation details. Could you explain your use case?

koba04:
I'm working on creating a custom renderer using `react-reconciler` and want to display a component name for debugging.
But in my case, It was enough to display component names for host components.

Use-cases in 3rd party libraries are something like the following.
- https://github.com/styled-components/styled-components/blob/c3eedbeb79621acd668e15ed2d63107528ed81c3/packages/styled-components/src/utils/getComponentName.js
- https://github.com/styled-components/styled-components/blob/c3eedbeb79621acd668e15ed2d63107528ed81c3/packages/styled-components/src/utils/generateDisplayName.js

But it's not the purpose of this PR so I'll close this.
Thank you!

PR #16865, opened 2019-09-23
[work loop] Prevent work loop from being inlined

acdlite:
Uses Closure Compiler's `@noinline` directive.

See https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#noinline


gaearon:
Check that this actually worked? Rollup might mess with function comments as it moves them around.

acdlite:
Seems to work
Screen Shot 2019-09-23 at 11 09 40 AM

PR #16869, opened 2019-09-23
DevTools FB feature flag system

bvaughn:
Adds a basic feature flag system to DevTools extension builds, supporting `stable` and `prerelease` builds. This will enable early testing of prerelease features within the smaller pool of Facebook engineers using DevTools.


packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -907,11 +908,17 @@ export function attach(
     return null;
   }
 
-  function didHooksChange(prev: any, next: any): boolean {
+  function getChangedHooks(
+    prev: any,
+    next: any,
+  ): boolean | null | Array<number> {
```
LetItRock: I think that it would be better to split this function into two separate, like one saying if hooks even changed and another returning indicies... or another idea - it could always return array with indicies and if the flag is true then return that array - otherwise just just boolean checking length of that array... 
if we want to return faster from while loop than option one is better... wdyt?

bvaughn:
Seems potentially useful but not pressing so I'm going to close.

PR #16878, opened 2019-09-24
Update useEditableValue hook to sync external value changes

### LINKED ISSUE #16843: DevTools: showing wrong state

bvaughn:
This PR changes the `useEditableValue` to more intelligently reset the local/editable state value when external value changes.

Previously, the hook initialized local state (in `useState`) to mirror the prop/state value. Updates to the value were ignored though. (Once the state was initialized, it was never updated.) The new hook updates the local/editable state to mirror the external value *unless* there are already pending, local edits being made.

Note that this particular change very closely resembles [one of the derived state anti-patterns](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-erasing-state-when-props-change) outlined in a previous blog post. I think it might still be an acceptable approach for this specific scenario though.

### Here is a demo of the (broken) behavior before:
useEditableValue broken before behavior

### And here is the behavior after this change:
useEditableValue fixed after behavior

### Edge case
Because of the way the new hook works, there is a potentially odd edge case, but I think it's acceptable given the way the "reset" button works:
useEditableValue remaining weird edge case

Posting this PR for discussion purposes. Let's talk!

Resolves #16843

cc @hristo-kanchev

---

This was also broken for other use cases before. The example above was just one of the most awkward cases, but here is another:

### Before
useEditableValue before bug

### After
useEditableValue after

### DIFF
--- a/packages/react-devtools-shared/src/__tests__/useEditableValue-test.js
+++ b/packages/react-devtools-shared/src/__tests__/useEditableValue-test.js
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+describe('useEditableValue', () => {
+  let act;
+  let React;
+  let ReactDOM;
+  let useEditableValue;
+
+  beforeEach(() => {
+    const utils = require('./utils');
+    act = utils.act;
+
+    React = require('react');
+    ReactDOM = require('react-dom');
+
+    useEditableValue = require('../devtools/views/hooks').useEditableValue;
+  });
+
+  it('should override editable state when external props are updated', () => {
+    let state;
+
+    function Example({value}) {
+      const tuple = useEditableValue(value);
+      state = tuple[0];
+      return null;
+    }
+
+    const container = document.createElement('div');
+    ReactDOM.render(<Example value="foo" />, container);
+    expect(state.editableValue).toEqual('"foo"');
+    expect(state.externalValue).toEqual('foo');
+    expect(state.hasPendingChanges).toBe(false);
+
+    // If there are NO pending changes,
+    // an update to the external prop value should override the local/pending value.
+    ReactDOM.render(<Example value="bar" />, container);
+    expect(state.editableValue).toEqual('"bar"');
+    expect(state.externalValue).toEqual('bar');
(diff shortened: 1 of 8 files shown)


bvaughn:
(This was also broken for other use cases before. The example I showed in the PR video was just one of the most awkward cases.)

packages/react-devtools-shared/src/__tests__/useEditableValue-test.js
```diff
@@ -0,0 +1,84 @@
...
+    expect(state.hasPendingChanges).toBe(true);
+
+    // If there ARE pending changes,
+    // an update to the external prop value should NOT override the local/pending value.
+    ReactDOM.render(<Example value="bar" />, container);
+    expect(state.editableValue).toEqual('not-foo');
+    expect(state.externalValue).toEqual('bar');
+    expect(state.hasPendingChanges).toBe(true);
+  });
+});
```
gaearon: Wanna add a case for when serialization fails?

bvaughn: Sure, good call.

packages/react-devtools-shared/src/devtools/views/hooks.js
```diff
@@ -8,68 +8,102 @@
...
 
-  const reset = useCallback(
-    () => {
-      setEditableValue(smartStringify(initialValue));
-      setParsedValue(initialValue);
-      setIsValid(initialIsValid);
-    },
-    [initialValue, initialIsValid],
-  );
+  if (state.externalValue !== externalValue) {
```
bvaughn: psst, `NaN` will never equal `NaN` idiot

PR #16879, opened 2019-09-24
[React Native] Improve errors for invalid ViewConfig getter functions

JoshuaGross:
# Overview

Improve error messages around invalid ViewConfig getter functions.

# Details

As a followup to #16821, it is unlikely but possible for ViewConfig registry entries to be corrupted. I would like to surface this as early as possible, and include a little more information in exceptions that could be thrown later on.

### DIFF
--- a/packages/react-native-renderer/src/__tests__/ReactNativeError-test.internal.js
+++ b/packages/react-native-renderer/src/__tests__/ReactNativeError-test.internal.js
@@ -32,6 +32,17 @@ describe('ReactNativeError', () => {
         .computeComponentStackForErrorReporting;
   });
 
+  it('should throw error if null component registration getter is used', () => {
+    let error;
+    try {
+      createReactNativeComponentClass('View', null);
+    } catch (e) {
+      error = e;
+    }
+
+    expect(error.toString()).toBe('Invariant Violation: View config getter callback must be a function: View (received null)');
+  });
+
   it('should be able to extract a component stack from a native view', () => {
     const View = createReactNativeComponentClass('View', () => ({
       validAttributes: {foo: true},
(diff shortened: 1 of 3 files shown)


packages/react-native-renderer/src/__tests__/ReactNativeError-test.internal.js
```diff
@@ -32,6 +32,16 @@ describe('ReactNativeError', () => {
         .computeComponentStackForErrorReporting;
   });
 
+  it('should throw error if null component registration getter is used', () => {
+    expect(() => {
+      try {
+        createReactNativeComponentClass('View', null)
+      } catch (e) {
+        throw new Error(e.toString());
```
TheSavior: 🤔

packages/react-native-renderer/src/__tests__/ReactNativeError-test.internal.js
```diff
@@ -32,6 +32,16 @@ describe('ReactNativeError', () => {
...
+      }
+    }).toThrow('Invariant Violation: View config getter callback must be a function: View (received null)');
```
TheSavior: This doesn't read super great to me.

What about:

```
Invariant Violation: Cannot create component View, view config getter must be a function (received null)
```

JoshuaGross: Hmm, that's reasonable but it's the registry that's throwing this error, not the `createReactNativeComponentClass`

TheSavior: What about 
```
'Invariant Violation: View config getter callback for component `View` must be a function (received null)'
```

PR #16883, opened 2019-09-25
These flags are hard coded in our internal config

sebmarkbage:
By hard coding them here they can get properly DCE and don't have to be sent to users. The internal config should only have dynamic flags.

### DIFF
--- a/packages/shared/forks/ReactFeatureFlags.www.js
+++ b/packages/shared/forks/ReactFeatureFlags.www.js
@@ -14,17 +14,8 @@ import typeof * as FeatureFlagsShimType from './ReactFeatureFlags.www';
 export const {
   debugRenderPhaseSideEffects,
   debugRenderPhaseSideEffectsForStrictMode,
-  replayFailedUnitOfWorkWithInvokeGuardedCallback,
-  warnAboutDeprecatedLifecycles,
   disableInputAttributeSyncing,
-  warnAboutShorthandPropertyCollision,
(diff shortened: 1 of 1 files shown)


packages/shared/forks/ReactFeatureFlags.www.js
```diff
@@ -39,6 +30,16 @@ export const enableProfilerTimer = __PROFILE__;
 export const enableSchedulerTracing = __PROFILE__;
 export const enableSchedulerDebugging = true;
 
+export const replayFailedUnitOfWorkWithInvokeGuardedCallback = false;
+export const warnAboutDeprecatedLifecycles = true;
+export const warnAboutShorthandPropertyCollision = false;
+export const warnAboutDeprecatedSetNativeProps = false;
+export const disableLegacyContext = false;
```
gaearon: We should make a GK for this one. I can do in a follow-up.

packages/shared/forks/ReactFeatureFlags.www.js
```diff
@@ -39,6 +30,16 @@ export const enableProfilerTimer = __PROFILE__;
 export const enableSchedulerTracing = __PROFILE__;
 export const enableSchedulerDebugging = true;
 
+export const replayFailedUnitOfWorkWithInvokeGuardedCallback = false;
+export const warnAboutDeprecatedLifecycles = true;
+export const warnAboutShorthandPropertyCollision = false;
```
gaearon: This one too.

PR #16889, opened 2019-09-25
[react-interactions] Add Portal propagation configuration

trueadm:
This adds Portal propagation configuration support to the new event system. By default responder targets will not propagate through React Portals, unless a `targetPortalPropagation` flag is set to `true` on the responder object.


necolas:
Oh I'd assigned this task to myself

trueadm:
@necolas Ah my bad, sorry about that! Thanks for the review :)

sebmarkbage:
:( I really don't like this approach since it's now no longer safe to convert to and from a portal.

trueadm:
@sebmarkbage What approach do you think is better? Happy to revert and take an alternative approach?

necolas:
There isn't a problem when not using a portal. I'd rather make it so portals can be configured to prevent event bubbling but you opposed that too in the existing thread about the issues people had when migrating to the stable createPortal API

PR #16894, opened 2019-09-25
[react-interactions] Add more documentation for a11y components

trueadm:
This adds more internal documentation for the experimental accessibility components that we're using internally.

### DIFF
--- a/packages/react-interactions/accessibility/docs/FocusControl.md
+++ b/packages/react-interactions/accessibility/docs/FocusControl.md
@@ -1,0 +1,60 @@
+# FocusControl
+
+`FocusControl` is a module that exports a selection of helpful utility functions to be used
+in conjunction with the `ref` from a React Scope, such as `TabbableScope`.
+A ref from `FocusManager` can also be used instead.
+
+## Example
+
+```jsx
+const {
+  focusFirst,
+  focusNext,
+  focusPrevious,
+  getNextScope,
+  getPreviousScope,
+} = FocusControl;
+
+function KeyboarFocusMover(props) {
+  const scopeRef = useRef(null);
+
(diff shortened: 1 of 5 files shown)


packages/react-interactions/accessibility/docs/FocusControl.md
```diff
@@ -0,0 +1,60 @@
...
+```jsx
+const {
+  focusFirst,
+  focusNext,
+  focusPrevious,
+  getNextScope,
+  getPreviousScope,
+} = FocusControl;
+
+function KeyboarFocusMover(props) {
```
necolas: typo: Keyboard

packages/react-interactions/accessibility/docs/FocusManager.md
```diff
@@ -0,0 +1,39 @@
...
+
+`FocusManager` is a component that is designed to provide basic focus management
+control. These are the various props that `FocusManager` accepts:
+
+## Usage
+
+```jsx
+function MyDialog(props) {
+  return (
+    <TabbableScope containFocus={true} autoFocus={true}>
```
necolas: This example says `TabbableScope`

packages/react-interactions/accessibility/docs/TabbableScope.md
```diff
@@ -0,0 +1,58 @@
...
+}
+```
+
+## Implementation
+
+`TabbableScope` uses the experimental `React.unstable_createScope` API. The query
+function used for the scope is designed to collect DOM nodes that are tab focusable
+to the browser:
+
+```js
```
necolas: Could link to the implementation so we don't have to update this README if it changes

PR #16897, opened 2019-09-25
DevTools shows unsupported renderer version dialog

### LINKED ISSUE #16462: DevTools: Fully disable 0.14 support

bvaughn:
Rather than partially supporting v14, DevTools now shows an unsupported dialog with a link to instructions on how to downgrade DevTools to v3:

Screen Shot 2019-09-25 at 1 27 17 PM

I decided on this approach to detecting versions by looking at all v14 and v15 minor releases and by referencing [the old DevTools logic](https://github.com/facebook/react-devtools/blob/d839081f79f681617df7584a8ffd2f4163fa40b9/backend/attachRenderer.js#L24-L76).

I tested all minor release of v15 and v16 to ensure that the dialog didn't get shown. I tested the latest release of v14 and verified that it did.

Resolves #16462

### DIFF
--- a/packages/react-devtools-shared/src/devtools/store.js
+++ b/packages/react-devtools-shared/src/devtools/store.js
@@ -73,6 +73,7 @@ export default class Store extends EventEmitter<{|
   supportsNativeStyleEditor: [],
   supportsProfiling: [],
   supportsReloadAndProfile: [],
+  unsupportedRendererVersionDetected: [],
 |}> {
   _bridge: FrontendBridge;
 
@@ -124,6 +125,8 @@ export default class Store extends EventEmitter<{|
   _supportsNativeInspection: boolean = true;
   _supportsProfiling: boolean = false;
   _supportsReloadAndProfile: boolean = false;
+
+  _unsupportedRendererVersionDetected: boolean = false;
 
   // Total number of visible elements (within all roots).
   // Used for windowing purposes.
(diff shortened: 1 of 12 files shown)


packages/react-devtools-shared/src/devtools/store.js
```diff
@@ -337,6 +344,10 @@ export default class Store extends EventEmitter<{|
     return this._supportsReloadAndProfile && this._isBackendStorageAPISupported;
   }
 
+  get unsupportedRendererVersionDetected(): boolean {
```
trueadm: Is it normal that we use getters vs generic functions?

bvaughn: In the DevTools `Store` I guess it is. Any reason to prefer one over the other?

bvaughn: In DevTools `Store` I guess it is. Any reason to prefer one over the other?

trueadm: I guess it's fine as this won't be used on unsupported browsers anyway.

bvaughn: Nah, DevTools has the benefit of only being used in Chrome & Firefox

PR #16898, opened 2019-09-25
Add HostComponent type to ReactNative

TheSavior:
**Note: This change was already landed to React Native in https://github.com/facebook/react-native/commit/69c38e5a639f34620038ae5724426c92c355e509 so this commit just updates the types in React core to make the next sync not have conflicts.**

---------------

In React Native there are three types of "Native" components.

```
createReactClass with NativeMethodsMixin
```
```
class MyComponent extends ReactNative.NativeComponent
```
```
requireNativeComponent('RCTView')
```

The implementation for how to handle all three of these exists in the React Native Renderer. Refs attached to components created via these methods provide a set of functions such as
```
.measure
.measureInWindow
.measureLayout
.setNativeProps
```

These methods have been used for our core components in the repo to provide a consistent API. Many of the APIs in React Native require a `reactTag` to a host component. This is acquired by calling `findNodeHandle` with any component. `findNodeHandle` works with the first two approaches.

For a lot of our new Fabric APIs, we will require passing a ref to a HostComponent directly instead of relying on `findNodeHandle` to tunnel through the component tree as that behavior isn't safe with React concurrent mode.

The goal of this change is to enable us to differentiate between components created with `requireNativeComponent` and the other types. This will be needed to be able to safely type the new APIs.

For existing components that should support being a host component but need to use some JS behavior in a wrapper, they should use `forwardRef`. The majority of React Native's core components were migrated to use `forwardRef` last year. Components that can't use forwardRef will need to have a method like `getNativeRef()` to get access to the underlying host component ref.

Note, we will need follow up changes as well as changes to the React Renderer in the React repo to fully utilize this new type.

### DIFF
--- a/packages/react-native-renderer/src/NativeMethodsMixin.js
+++ b/packages/react-native-renderer/src/NativeMethodsMixin.js
@@ -161,7 +161,7 @@ export default function(
     measureLayout: function(
       relativeToNativeNode: number | Object,
       onSuccess: MeasureLayoutOnSuccessCallback,
-      onFail: () => void /* currently unused */,
+      onFail?: () => void /* currently unused */,
     ) {
       let maybeInstance;
 
--- a/packages/react-native-renderer/src/ReactFabricHostConfig.js
+++ b/packages/react-native-renderer/src/ReactFabricHostConfig.js
@@ -11,7 +11,7 @@ import type {
   MeasureInWindowOnSuccessCallback,
   MeasureLayoutOnSuccessCallback,
   MeasureOnSuccessCallback,
-  NativeMethodsMixinType,
+  NativeMethods,
   ReactNativeBaseComponentViewConfig,
   ReactNativeResponderEvent,
   ReactNativeResponderContext,
@@ -151,9 +151,9 @@ class ReactFabricHostComponent {
   }
 
   measureLayout(
-    relativeToNativeNode: number | Object,
+    relativeToNativeNode: number | ReactFabricHostComponent,
     onSuccess: MeasureLayoutOnSuccessCallback,
-    onFail: () => void /* currently unused */,
+    onFail?: () => void /* currently unused */,
   ) {
     if (
       typeof relativeToNativeNode === 'number' ||
@@ -186,7 +186,7 @@ class ReactFabricHostComponent {
 }
 
 // eslint-disable-next-line no-unused-expressions
-(ReactFabricHostComponent.prototype: NativeMethodsMixinType);
+(ReactFabricHostComponent.prototype: NativeMethods);
 
 export * from 'shared/HostConfigWithNoMutation';
 export * from 'shared/HostConfigWithNoHydration';
--- a/packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
+++ b/packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
@@ -11,7 +11,7 @@ import type {
   MeasureInWindowOnSuccessCallback,
   MeasureLayoutOnSuccessCallback,
   MeasureOnSuccessCallback,
-  NativeMethodsMixinType,
+  NativeMethods,
   ReactNativeBaseComponentViewConfig,
 } from './ReactNativeTypes';
 import type {Instance} from './ReactNativeHostConfig';
@@ -72,11 +72,11 @@ class ReactNativeFiberHostComponent {
   }
 
   measureLayout(
-    relativeToNativeNode: number | Object,
+    relativeToNativeNode: number | ReactNativeFiberHostComponent,
     onSuccess: MeasureLayoutOnSuccessCallback,
-    onFail: () => void /* currently unused */,
+    onFail?: () => void /* currently unused */,
   ) {
-    let relativeNode;
+    let relativeNode: ?number;
 
     if (typeof relativeToNativeNode === 'number') {
       // Already a node handle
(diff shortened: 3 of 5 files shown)


packages/react-native-renderer/src/NativeMethodsMixin.js
```diff
@@ -161,7 +161,7 @@ export default function(
     measureLayout: function(
       relativeToNativeNode: number | Object,
       onSuccess: MeasureLayoutOnSuccessCallback,
-      onFail: () => void /* currently unused */,
+      onFail?: () => void /* currently unused */,
```
TheSavior: This was always optional in the types used in React Native, but because the two types are duplicated in `ReactNativeTypes`, they must have fallen out of sync.

packages/react-native-renderer/src/ReactFabricHostConfig.js
```diff
@@ -151,9 +151,9 @@ class ReactFabricHostComponent {
   }
 
   measureLayout(
-    relativeToNativeNode: number | Object,
```
TheSavior: Yay!

packages/react-native-renderer/src/ReactNativeTypes.js
```diff
@@ -8,7 +8,7 @@
  * @flow
  */
 
-import React from 'react';
+import * as React from 'react';
```
TheSavior: This is needed because we are now importing types from React. `ElementRef`, and `AbstractComponent`.

packages/react-native-renderer/src/ReactNativeTypes.js
```diff
@@ -96,23 +96,34 @@ class ReactNativeComponent<Props> extends React.Component<Props> {
...
+    relativeToNativeNode: number | React.ElementRef<HostComponent<mixed>>,
     onSuccess: MeasureLayoutOnSuccessCallback,
-    onFail: () => void,
+    onFail?: () => void,
   ): void,
   setNativeProps(nativeProps: Object): void,
 };
 
+export type NativeMethodsMixinType = NativeMethods;
+export type HostComponent<T> = React.AbstractComponent<
```
TheSavior: So simple. So clean.

packages/react-native-renderer/src/ReactFabricHostConfig.js
bvaughn: I realize it's not part of this change, but should the `warningWithoutStack` inside of this function (and the one in `packages/react-native-renderer/src/ReactNativeFiberHostComponent.js`) be `__DEV__` gated?

packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
```diff
@@ -72,21 +72,25 @@ class ReactNativeFiberHostComponent {
...
     } else if (relativeToNativeNode._nativeTag) {
       relativeNode = relativeToNativeNode._nativeTag;
     } else if (
+      /* $FlowFixMe canonical doesn't exist on the node.
+       I think this branch is dead and will remove it in a followup */
       relativeToNativeNode.canonical &&
       relativeToNativeNode.canonical._nativeTag
     ) {
+      /* $FlowFixMe canonical doesn't exist on the node.
+       I think this branch is dead and will remove it in a followup */
```
bvaughn: I was wondering if the change from `Object` -> `ReactNativeFiberHostComponent ` made sense given `relativeToNativeNode.canonical` but I assume it's okay?

packages/react-native-renderer/src/ReactNativeTypes.js
```diff
@@ -8,7 +8,7 @@
  * @flow
  */
 
-import React from 'react';
+import React, {type ElementRef, type AbstractComponent} from 'react';
```
bvaughn: I didn't know you could mix ES imports and type imports this way 😮

packages/react-native-renderer/src/ReactNativeTypes.js
```diff
@@ -89,30 +89,41 @@ class ReactNativeComponent<Props> extends React.Component<Props> {
...
-    onFail: () => void,
+    onFail?: () => void,
   ): void,
   setNativeProps(nativeProps: Object): void,
 };
 
+export type NativeMethodsMixinType = NativeMethods;
+export type HostComponent<T> = AbstractComponent<
+  T,
+  $ReadOnly<$Exact<NativeMethods>>,
```
bvaughn: I don't think I understand the purpose of `$ReadOnly` or `$Exact` here.

packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
TheSavior: Yeah.

When I added this option I wrote tests that called this with createClass, ReactNativeNativeComponent and a host component. Nothing triggers this branch. That's why I think this is dead.

Worst case, nothing except some of our Fabric screens are using this API yet and we'll get quick signal if I'm wrong once we delete it. But it's unrelated to this type change.

packages/react-native-renderer/src/ReactNativeTypes.js
TheSavior: From @jbrown215:

> Definitely consider making the instance type of HostComponent read-only-- all of those function names seem like things you don't want people to overwrite. If the object really has no other properties, you can also make it exact. If it might, don't, otherwise spreading that object would be unsound (not that I can see a reason why someone would want to spread it?)

packages/react-native-renderer/src/ReactFabricHostConfig.js
TheSavior: warningWithoutStack internally has a dev check and is a no-op in dev.

PR #16903, opened 2019-09-26
[react-interactions] Add wrapping support to FocusList/FocusTable

trueadm:
This adds support to `FocusList` and `FocusTable` so that keyboard arrow focus moving can be wrapped back to the other side as requested internally. This also adds an optional `onKeyDown` prop for `TableCell` and `ListItem` so keyboard events can be intercepted.

### DIFF
--- a/packages/react-interactions/accessibility/src/FocusTable.js
+++ b/packages/react-interactions/accessibility/src/FocusTable.js
@@ -15,6 +15,7 @@ import {useKeyboard} from 'react-interactions/events/keyboard';
 
 type FocusCellProps = {
   children?: React.Node,
+  onKeyDown?: KeyboardEvent => void,
 };
 
 type FocusRowProps = {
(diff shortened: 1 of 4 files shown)


packages/react-interactions/accessibility/src/FocusTable.js
```diff
@@ -162,10 +186,15 @@ export function createFocusTable(scope: ReactScope): Array<React.Component> {
...
                 if (rowIndex > 0) {
                   const row = rows[rowIndex - 1];
-                  focusCellByIndex(row, cellIndex);
-                  event.preventDefault();
+                  focusCellByIndex(row, cellIndex, event);
                 } else if (rowIndex === 0) {
-                  triggerNavigateOut(currentCell, 'up', event);
+                  const wrap = getTableWrapProp(currentCell);
+                  if (wrap) {
+                  } else {
```
gaearon: This empty if looks a bit confusing? Is there a mistake?

trueadm: Fixed :)

PR #16909, opened 2019-09-26
expose isHydrating

lunaruan:
Added a function on `ReactFiberHydrationContext` to expose `isHydrating` in `ReactDOMFB`


lunaruan:
This change is most likely just going to be temporary and only used in one place, so it seemed easier to put it on the SECRET_INTERNALS object for now so it can also be easily removed.

PR #16912, opened 2019-09-26
Document `additionalHooks` option

airjp73:
Hi!

I noticed that there doesn't seem to be any documentation about this option for `exhaustive-deps`. Was this a deliberate decision or an oversight? Obviously, feel free to decline this if you don't want to do this or this is the wrong place.

I recently used this for a custom hook at work and it's pretty useful but I'm a little uneasy about using undocumented features.

### DIFF
--- a/packages/eslint-plugin-react-hooks/README.md
+++ b/packages/eslint-plugin-react-hooks/README.md
@@ -34,6 +34,20 @@ Then add it to your ESLint configuration:
 }
 ```
 
+## Configuration
+
+`exhaustive-deps` can be configured to validate custom hooks via the `additionalHooks` option.
+This option accepts a regex to match the names of custom hooks.
+
+```js
+{
+  "rules": {
+    // ...
+    "react-hooks/exhaustive-deps": ["warn", { "additionalHooks": "(useMyCustomHook|useMyOtherCustomHook)" }]
+  }
+}
+```
+
 ## Valid and Invalid Examples
 
 Please refer to the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) documentation and the [Hooks FAQ](https://reactjs.org/docs/hooks-faq.html#what-exactly-do-the-lint-rules-enforce) to learn more about this rule.


vkrol:
Related: https://github.com/reactjs/reactjs.org/issues/2350

wcandillon:
@airjp73 Thks for contributing this change. I hope this gets merged.

wcandillon:
This small PR would really be useful. Thank you for this @airjp73.

gaearon:
Most people really shouldn't use it. It's more of an escape hatch for tightly controlled cases. I think I'd prefer this stays undocumented. People who really need it will find it, but generally this isn't something you want to enable.

dclowd9901:
> Most people really shouldn't use it. It's more of an escape hatch for tightly controlled cases. I think I'd prefer this stays undocumented. People who really need it will find it, but generally this isn't something you want to enable.

With all due respect, this is silly reasoning. This is hardly a dangerous footgun, and is incredibly useful to devs who write hooks (basically any React developer up to this point). If an org finds this isn't an effective way to set standards internally, they can (and will) easily change the way they do things, but to make that decision for them is addressing a problem that doesn't exist.

mbest:
I'm trying to figure out how to use this option. Here is what I've added to package.json:

```json
  "eslintConfig": {
    "extends": "react-app",
    "rules": {
      "react-hooks/exhaustive-deps": [
        "warn",
        {
          "additionalHooks": "^(useDeepCompareLayoutEffect)$"
        }
      ]
    }
  },
```

But I don't get any warning. To test I replaced `useDeepCompareLayoutEffect` with `useLayoutEffect` in the code and do get a warning when running `yarn start`:

```
Compiled with warnings.

./src/core/providers/DataProvider.tsx
  Line 54:12:  React Hook useLayoutEffect has missing dependencies: 'otherProps' and 'setValue'.
  Either include them or remove the dependency array  react-hooks/exhaustive-deps

Search for the keywords to learn more about each warning.
To ignore, add // eslint-disable-next-line to the line before.
```

Any ideas?

gaearon:
FWIW I'd take a PR that enables the check for any Hook ending with Effect. That seems legit to me.

airjp73:
That sounds like a reasonable solution to me. I'd love to take a crack at that some time this week / weekend.

dclowd9901:
@mbest have you tried not matching start and end of line? I don't know if it matters, but I do not have that distinction on mine, and it works fine.

mbest:
Since we're using Create React App, I had to set the `EXTEND_ESLINT` environment variable to [enable customization of the eslint rules](https://create-react-app.dev/docs/setting-up-your-editor/#experimental-extending-the-eslint-config).

gaearon:
We've tried the heuristic, but I had to revert it because too many false positives are being reported (https://github.com/facebook/react/pull/19004). So I'm happy to take this PR as an alternative, if you'd like to resubmit it.

Issue #16848, opened 2019-09-21
React DOM SSR crashes since React-Dom 16.6.2


ernestognw:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
I've setup an SSR configuration using ReactDOM.renderToString() method with react-apollo, and it throws an Invariant violation as the following, with no more information about:

<img width="297" alt="Captura de pantalla 2019-09-21 a la(s) 6 29 37" src="https://user-images.githubusercontent.com/33379285/65372722-36136c80-dc39-11e9-8342-90369c2055e5.png">

Everything returns to normal when downgrading to react-dom@16.6.1

*My SSR App component*

<img width="600" alt="Captura de pantalla 2019-09-21 a la(s) 6 33 24" src="https://user-images.githubusercontent.com/33379285/65372762-b9cd5900-dc39-11e9-9045-143146e0baf9.png">

*Things I've tried*

- I have changed the react-dom package to serve the development version (./cjs/react-dom-server.node.development.js), just for debugging, and when importing that version, everything works fine, but when I return to the production version (./cjs/react-dom-server.node.production.min.js) the same error appears.

- I removed one by one, every component wrapping my `<App/>`, and the one that seems to be causing the error is the `<AuthProvider>`, which is a context implementation of mine. By removing some things on that component, i figured out that is a problem within a HOC provided by react-apollo, but I don't know exactly what it is because it doesn't show me any information.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

I actually do not have any way to reproduce it quickly.

**What is the expected behavior?**
Render correctly the website.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

React - 16.9.0
React-DOM - 16.9.0
Browsers - Safari and Chrome (tested)

Working if downgrading to React-DOM 16.6.1

soldovskij:
I think renderToString do not await callback from Context, have some issue

gaearon:
Closing since there's no repro

PR #16913, opened 2019-09-26
Critical bug at [server side render], failure when use [Context.Consumer] in [production] mode only.

soldovskij:
Hello everyone!
That fix for server side rendering when you use Context.Consumer.

Issues for example in react:
https://github.com/facebook/react/issues/16848

Issues for example in react-router:
https://github.com/ReactTraining/react-router/issues/6789
https://github.com/ReactTraining/react-router/issues/6704

Issues for example in redux-connect:
https://github.com/makeomatic/redux-connect/issues/137

Issues for example in next.js:
https://github.com/zeit/next.js/issues/7167

Bug was in `ReactDOMServerRenderer.js` since version `16.6.2` and affects.

- `renderToString`
- `renderToStaticMarkup` 
- `renderToNodeStream`
- `renderToStaticNodeStream`

To reproduce, only in `production` mode and total bundle, try

```
const ThemeContext = React.createContext('light');

const Button = () => {
  return (
    <ThemeContext.Provider value="dark">
      <ThemeContext.Consumer>
        {theme => <span>{theme}</span>}
      </ThemeContext.Consumer>
   </ThemeContext.Provider>
  )
}

const appHTML = ReactDomServer.renderToString(<Button />)

console.log(appHTML)
```

And you will see just `<span></span>` instead `<span>dark</span>`

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRenderer.js
+++ b/packages/react-dom/src/server/ReactPartialRenderer.js
@@ -1186,10 +1186,13 @@ class ReactDOMServerRenderer {
                     );
                   }
                 }
-              } else {
-                reactContext = (reactContext: any)._context;
               }
             }
+
+            if ((reactContext: any)._context !== undefined) {
+              reactContext = (reactContext: any)._context;
+            }
+
             const nextProps: any = (nextChild: any).props;
             const threadID = this.threadID;
             validateContextBounds(reactContext, threadID);


edvinerikson:
Would be nice with a test case as well 🙂

soldovskij:
> Would be nice with a test case as well 🙂

As far as I understand, the test is difficult to implement, since the error occurs after the separation files on `development` and `production`

soldovskij:
Don't know how, but test will be passed, but after build bug has effect.

soldovskij:
@sebmarkbage @trueadm that code between yours, can you chek it?

edvinerikson:
Probably need a test that runs on the production bundle. However I'm not very familiar how that works. Given that it's only the production bundle that has the bug, it might be a minifier bug causing this.

soldovskij:
Maybe, but this PR fix this behavior

soldovskij:
In deep, problem in object structure of react element. 
In dev it have {_context[thread_id], ...}, in prod {[thread_id], ...}. 
Code above my commit has comment about it.

I think problem arose by merge conflict. Condition for check this was shifted to dev block. 
But in real this condition need on both: dev and prod modes.

soldovskij:
@trueadm 
Can u try this?
1. git pull https://github.com/soldovskij/all-the-hooks/tree/srr-context 
2. yarn && yarn && start
3. open localhost:8080 and see `<span></span>` instead `<span>dark</span>`

as I understand it production mode

soldovskij:
Yes, you right. It is was collision in my bundle.
Good experience =)

trueadm:
Awesome. Glad I could help :)

milos2211:
I have same problem and i tried to create simple component with:
```const ThemeContext = React.createContext("light")
const Span = () => {
  return (
    <ThemeContext.Provider value="dark">
      <ThemeContext.Consumer>
        {theme => <span>{theme}</span>}
      </ThemeContext.Consumer>
    </ThemeContext.Provider>
  )
}
```
And this is working, but if i change value to object, like this: 
```const Span = () => {
  return (
    <ThemeContext.Provider value={{theme:"dark"}}>
      <ThemeContext.Consumer>
        {({theme}) => (<span>{theme}</span>)}
      </ThemeContext.Consumer>
    </ThemeContext.Provider>
  )
}
```
I get
> Cannot read property 'theme' of undefined

Both codes are working normally in DEV mode.

Any idea what could be wrong? 

Thanks.

milos2211:
I realize that i had multiple instances of Context. So when i fixed that, problem went away.

PR #16915, opened 2019-09-27
[react-interactions] Fix virtual click heuristic

necolas:


necolas:
Weird that `yarn flow dom` failed locally with errors like "Cannot resolve module react-interactions/events/keyboard." All good here though

PR #16918, opened 2019-09-27
Rename newState to oldState when use reducer

NE-SmallTown:
From https://redux.js.org/basics/reducers#handling-actions, this is how `reducer` means :)


NE-SmallTown:
sry, don't see the loop

PR #16935, opened 2019-09-27
[Selective Hydration] Increase priority for non-synchronous discrete events and retries

sebmarkbage:
If an increased priority dehydrated boundary commits as dehydrated because it is suspended, it no longer has any `expirationTime` scheduled. So when the thenable resolves, it gets the default priority. This increases the `retryTime`. This will almost immediately expire and effectively become sync. We might want to make this slightly higher than Normal but longer expiration time.

The upside of eagerly computing this (as opposed to when resolved) is that we can guarantee a predictable order of priorities. (First event in, highest pri.)

This PR also ensures that we increase the priority of subsequent events after the first can't be done synchronously. There's no reason to keep them as high as sync since at that point it is already too late since replaying has kicked in. So we schedule with user blocking pri for this case.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerSelectiveHydration-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerSelectiveHydration-test.internal.js
@@ -166,7 +166,7 @@ describe('ReactDOMServerSelectiveHydration', () => {
 
     container.innerHTML = finalHTML;
 
-    let span = container.getElementsByTagName('span')[3];
+    let spanD = container.getElementsByTagName('span')[3];
 
     suspend = true;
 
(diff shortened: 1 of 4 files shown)


trueadm:
I wonder, would this have future affects on Flare? It doesn't matter right now, but if you think it does a test that is skipped for Flare would be useful.

sebmarkbage:
The way I'm thinking about it is that we're testing the interaction between event system <-> replaying system in PartialHydration file which has tests for both. In this file we're testing the interaction between event replaying <-> reconciler priorities. We're actually testing the reconciler here, more than the events themselves. So we don't need to duplicate every combination of these 
tests.

PR #16936, opened 2019-09-28
[react-interactions] Tap cancels on second pointerdown

necolas:
This patch causes onTapCancel to be called whenever a second pointer interacts with the responder target.

### DIFF
--- a/packages/react-interactions/events/src/dom/__tests__/Tap-test.internal.js
+++ b/packages/react-interactions/events/src/dom/__tests__/Tap-test.internal.js
@@ -212,7 +212,7 @@ describeWithPointerEvent('Tap responder', hasPointerEvents => {
       const buttons = buttonsType.auxiliary;
       const target = createEventTarget(ref.current);
       target.pointerdown({buttons, pointerType});
-      target.pointerup({pointerType});
+      target.pointerup({buttons, pointerType});
       expect(onAuxiliaryTap).toHaveBeenCalledTimes(1);
     });
 
@@ -221,7 +221,7 @@ describeWithPointerEvent('Tap responder', hasPointerEvents => {
       const buttons = buttonsType.primary;
       const target = createEventTarget(ref.current);
       target.pointerdown({buttons, pointerType});
-      target.pointerup({metaKey: true, pointerType});
+      target.pointerup({buttons, metaKey: true, pointerType});
       expect(onAuxiliaryTap).toHaveBeenCalledTimes(1);
     });
   });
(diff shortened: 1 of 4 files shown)


packages/react-interactions/events/src/dom/testing-library/domEventSequences.js
```diff
@@ -18,6 +18,12 @@ function getPointerType(payload) {
   let pointerType = 'mouse';
   if (payload != null && payload.pointerType != null) {
     pointerType = payload.pointerType;
+  } else if (
+    Array.isArray(payload) &&
+    payload[0] != null &&
+    payload[0].pointerType != null
+  ) {
+    pointerType = payload[0].pointerType;
```
necolas: If it's an array it must be a TouchEvent so we don't need to pluck it from the payload (or include a `pointerType` from the unit tests call sites)

packages/react-interactions/events/src/dom/__tests__/Tap-test.internal.js
```diff
@@ -652,6 +652,23 @@ describeWithPointerEvent('Tap responder', hasPointerEvents => {
...
+      const pointerType = 'touch';
+      const target = createEventTarget(ref.current);
+      const buttons = buttonsType.primary;
+      target.pointerdown({buttons, pointerId: 1, pointerType});
+      if (hasPointerEvents) {
+        target.pointerdown({buttons, pointerId: 2, pointerType});
+      } else {
+        // TouchEvents
+        target.pointerdown([
+          {pointerId: 1, pointerType},
```
necolas: No point having pointerType here because it can only be a `touchstart` event

packages/react-interactions/events/src/dom/testing-library/domEventSequences.js
```diff
@@ -156,8 +167,12 @@ export function pointermove(target, payload) {
 export function pointerup(target, defaultPayload = {}) {
   const dispatch = arg => target.dispatchEvent(arg);
   const pointerType = getPointerType(defaultPayload);
-  // eslint-disable-next-line no-unused-vars
-  const {buttons, ...payload} = defaultPayload;
+
+  // Payload could be an array for TouchEvents
+  const payload = defaultPayload;
+  if (defaultPayload.buttons) {
+    delete payload.buttons;
```
trueadm: Delete is bit of an anti-pattern. It's always generally faster to create a mew object wijhout.

necolas: That's not important in these test utilities, but this code is mutating the original object which it shouldn't

PR #16937, opened 2019-09-28
[Selective Hydration] Prioritize the last continuous target

sebmarkbage:
This ensures that the current focus target is always hydrated first.

Slightly higher than the usual Never expiration time used for hydration. The priority increases with each new queued item so that the last always wins.

This is mostly to ensure that tooltips show up at a reasonable time.

If there's client rendered content that currently comes first though.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerSelectiveHydration-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerSelectiveHydration-test.internal.js
@@ -16,6 +16,57 @@ let Scheduler;
 let ReactFeatureFlags;
 let Suspense;
 
+function dispatchMouseHoverEvent(to, from) {
+  if (!to) {
+    to = null;
+  }
+  if (!from) {
+    from = null;
+  }
+  if (from) {
+    const mouseOutEvent = document.createEvent('MouseEvents');
+    mouseOutEvent.initMouseEvent(
+      'mouseout',
+      true,
+      true,
+      window,
+      0,
+      50,
+      50,
+      50,
+      50,
+      false,
+      false,
+      false,
+      false,
+      0,
+      to,
+    );
+    from.dispatchEvent(mouseOutEvent);
+  }
+  if (to) {
+    const mouseOverEvent = document.createEvent('MouseEvents');
+    mouseOverEvent.initMouseEvent(
+      'mouseover',
+      true,
+      true,
+      window,
+      0,
+      50,
+      50,
+      50,
+      50,
+      false,
+      false,
+      false,
+      false,
+      0,
+      from,
+    );
+    to.dispatchEvent(mouseOverEvent);
+  }
+}
+
 function dispatchClickEvent(target) {
   const mouseOutEvent = document.createEvent('MouseEvents');
   mouseOutEvent.initMouseEvent(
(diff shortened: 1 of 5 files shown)


packages/react-dom/src/__tests__/ReactDOMServerSelectiveHydration-test.internal.js
```diff
@@ -290,4 +341,103 @@ describe('ReactDOMServerSelectiveHydration', () => {
...
+      'D',
+      'Clicked D',
+      'C',
+      'Hover C',
+      'B',
+      'A',
+    ]);
+
+    document.body.removeChild(container);
+  });
```
bvaughn: Nice test 👍

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -421,6 +425,19 @@ export function attemptUserBlockingHydration(fiber: Fiber): void {
   markRetryTimeIfNotHydrated(fiber, expTime);
 }
 
+export function attemptContinuousHydration(fiber: Fiber): void {
+  if (fiber.tag !== SuspenseComponent) {
+    // We ignore HostRoots here because we can't increase
```
bvaughn: ~~This comment seems to have been copy-pasted from above- but I'm not sure I understand why we specifically mention the `HostRoot` type here, since we are ignoring everything other than `SuspenseComponent`.~~ Ah, I was missing the context of how `attemptToDispatchEvent()` worked.

packages/react-dom/src/events/ReactDOMEventReplaying.js
```diff
@@ -43,6 +43,12 @@ export function setAttemptUserBlockingHydration(fn: (fiber: Object) => void) {
   attemptUserBlockingHydration = fn;
 }
 
+let attemptContinuousHydration: (fiber: Object) => void;
+
+export function setAttemptContinuousHydration(fn: (fiber: Object) => void) {
```
bvaughn: It is not obvious to me why we're doing this indirect injection from `ReactDOM` -> `ReactDOMEventReplaying` but it looks like this is copying pre-existing behavior for `setAttemptSynchronousHydration` so I assume it's okay.

packages/react-reconciler/src/ReactFiberExpirationTime.js
```diff
@@ -32,6 +32,10 @@ export const Never = 1;
 // Idle is slightly higher priority than Never. It must completely finish in
 // order to be consistent.
 export const Idle = 2;
+// Continuous Hydration is a moving priority. It is slightly higher than Idle
+// and is used to increase priority of hover targets. It is increasing with
+// each usage so that last always wins.
+export let ContinuousHydration = 3;
```
bvaughn: Exporting this value seems a little weird, since we might schedule work at the value of `ContinuousHydration` at one point in time but then if we were to later compare the scheduled `expirationTime` to `ContinuousHydration` they wouldn't match.

packages/react-dom/src/events/ReactDOMEventReplaying.js
sebmarkbage: There's a cycle introduced by event replaying from the reconciler -> host config -> event system -> reconciler. It leads to a mess. I have some ideas for how we can avoid these dynamic injections but it involves getting rid of some other dynamic cases like these need to move into in the functions instead of overriding the functions values during module initialization: https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberCompleteWork.js#L148

packages/react-reconciler/src/ReactFiberExpirationTime.js
sebmarkbage: Makes sense. I'll avoid exporting it.

Issue #16938, opened 2019-09-28
React 16.10 broke Next.js/SSR applications


Timer:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**

React 16.10.0 has broken all Next.js applications (and potentially other SSR solutions).

It appears you cannot `hydrate` in conjunction with a client-side `<Suspense>` component.

> Expected to have a hydrated suspense instance. This error is likely caused by a bug in React. Please file an issue.

![image](https://user-images.githubusercontent.com/616428/65810696-7f087b00-e17b-11e9-9957-3481fb5853e3.png)

**CodeSandbox**: https://codesandbox.io/s/i66g1

**What is the expected behavior?**

Not entirely sure -- I'm opening this issue to discuss. The provided example worked in React 16.9.0 (and prior releases containing Suspense).

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

`react@16.10.0`/`react-dom@16.10.0` is broken.
`react@16.9.0`/`react-dom@16.9.0` works.

acdlite:
Thanks for reporting! We'll have a fix out soon; in the meantime, please downgrade back to 16.9.0.

acdlite:
I believe technically we don't support conditionally rendering a Suspense boundary locally. It just happened to (accidentally) work before.

We'll either fix so it matches the old behavior or provide a more graceful error or fallback behavior.

sebmarkbage:
The issue is that the error only occurred on the server. So if you conditionally rendered it, you’d miss (intentionally?) the error we logged to let you know not to use Suspense with SSR. The issue is that the accidental semantics this lead to is not actually the semantics we want SSR Suspense to have.

We’ll at least need this fallback behavior to behave reasonably when the server renders the boundary. https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberBeginWork.js#L1970

Now the question is if we can preserve the old accidental behavior if the server was conditionally excluding the boundary on the server. And if that’s even a good idea.

iksent:
I've just found that error `Minified React error #317` today only in my production build of Nextjs. There was no such error yesterday.
But I am having React `^16.8.6` and didn't update any packages.
Is it possible?

threepointone:
@iksent please file a separate issue. I’d recommend including a reproducing example on codesandbox or as a git repo.

phamducquanptit:
[...]

PR #16945, opened 2019-09-28
Client render Suspense content if there's no boundary match

sebmarkbage:
Without the `enableSuspenseServerRenderer` flag there will never be a boundary match. Also when it is enabled, there might not be a boundary match if something was conditionally rendered by mistake.

With this PR it will now client render the content of a Suspense boundary in that case and issue a DEV only hydration warning. This is the only sound semantics for this case.

Unfortunately, landing this will once again break #16938. It will be less bad though because at least it'll just work by client rendering the content instead of hydrating and issue a DEV only warning.

However, we must land this before enabling the `enableSuspenseServerRenderer` flag since it does this anyway.

I did notice that we special case `fallback={undefined}` due to our unfortunate semantics for that. So technically a workaround that works is actually setting the fallback to undefined on the server and during hydration. Then flip it on only after hydration. That could be a workaround if you want to be able to have a Suspense boundary work only after hydration for some reason.

It's kind of unfortunate but at least those semantics are internally consistent. So I added a test for that.

cc @Timer @timneutkens

### DIFF
--- a/packages/react-dom/src/__tests__/ReactServerRenderingHydration-test.js
+++ b/packages/react-dom/src/__tests__/ReactServerRenderingHydration-test.js
@@ -660,14 +660,45 @@ describe('ReactDOMServerHydration', () => {
     document.body.removeChild(parentContainer);
   });
 
-  it('regression test: Suspense + hydration in legacy mode ', () => {
+  it('Suspense + hydration in legacy mode', () => {
     const element = document.createElement('div');
     element.innerHTML = '<div>Hello World</div>';
+    let div = element.firstChild;
+    let ref = React.createRef();
+    expect(() =>
+      ReactDOM.hydrate(
+        <React.Suspense fallback={null}>
+          <div ref={ref}>Hello World</div>
+        </React.Suspense>,
+        element,
+      ),
+    ).toWarnDev(
+      'Warning: Did not expect server HTML to contain a <div> in <div>.',
+      {withoutStack: true},
+    );
+
+    // The content should've been client rendered and replaced the
+    // existing div.
+    expect(ref.current).not.toBe(div);
+    // The HTML should be the same though.
+    expect(element.innerHTML).toBe('<div>Hello World</div>');
+  });
+
+  it('Suspense + hydration in legacy mode with no fallback', () => {
+    const element = document.createElement('div');
+    element.innerHTML = '<div>Hello World</div>';
+    let div = element.firstChild;
+    let ref = React.createRef();
     ReactDOM.hydrate(
       <React.Suspense>
-        <div>Hello World</div>
+        <div ref={ref}>Hello World</div>
       </React.Suspense>,
       element,
     );
+
+    // Because this didn't have a fallback, it was hydrated as if it's
+    // not a Suspense boundary.
+    expect(ref.current).toBe(div);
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/__tests__/ReactServerRenderingHydration-test.js
```diff
@@ -660,14 +660,45 @@ describe('ReactDOMServerHydration', () => {
...
+    let ref = React.createRef();
+    expect(() =>
+      ReactDOM.hydrate(
+        <React.Suspense fallback={null}>
+          <div ref={ref}>Hello World</div>
+        </React.Suspense>,
+        element,
+      ),
+    ).toWarnDev(
+      'Warning: Did not expect server HTML to contain a <div> in <div>.',
```
sebmarkbage: This warns because of the superfluous div but before that it should warn about missing a Suspense boundary but it's currently missing that warning:

https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMHostConfig.js#L843

Timer:
FYI, Next.js' specific `<Suspense>` component did have a `fallback` defined. The reproduction I opened omitted it, sorry about that!

The behavior Next.js relied on, represented as a test:
```js
  it('Suspense + hydration in legacy mode with fallback', () => {
    element.innerHTML = '<div>Hello World</div>';
    let div = element.firstChild;
    let ref = React.createRef();

    ReactDOM.hydrate(
      <React.Suspense fallback={<div>Loading...</div>}>
        <div ref={ref}>Hello World</div>
      </React.Suspense>,
      element,
    );

    // Suspense component did not suspend during hydration, so
    // it was hydrated as if it was absent.
    expect(ref.current).toBe(div);
    expect(element.innerHTML).toBe('<div>Hello World</div>');
  });
```

☝️ this worked because Next.js guaranteed nothing in the tree would suspend during hydration / initial render.

Timer:
We released `next@9.0.7` in response to `react@16.10.0`, removing our top level Suspense wrapper: https://github.com/zeit/next.js/pull/8887.

Meaning, `next@9.0.7` is compatible with `react@16.10.0` and does not require `react@16.10.1`.

We don't necessarily need this old behavior to be supported -- the Suspense boundary was not used for any stable Next.js feature (it was flagged behind an experimental feature).

However, this has been released in Next.js stables for ~6 months (April 2nd).
If necessary, we could cut a patch for our `8.1.x` release line to remove the Suspense boundary.

Time (and cross-team coordination) could probably heal this, the `16.10.0` release was just painful because there was no `9.0.x` version compatible (and still no `8.1.x` version compatible).

acdlite:
@Timer 

> Time (and cross-team coordination) could probably heal this

Yeah we realized when discussing this incident that we really should be running smoke tests on the major React frameworks as part of our release plan.

> However, this has been released in Next.js for ~6 months (April 2nd).
If necessary, we could cut a patch for our `8.1.x` release line to remove the Suspense boundary.

That might be wise. We'll wait to land this change until after we figure out what to do in Next and the other projects that might be relying on this accidental behavior.

In the meantime, I marked 16.10.0 as deprecated so we should be fine for now.

timneutkens:
I'm not opposed to releasing a `8.1.1` that removes the suspense boundary too, even though Next.js 9 doesn't really break in most cases.

PR #16949, opened 2019-09-28
updated comment message with better wording

KirankumarAmbati:
The existing message wordings are little bit unclear. Made change to replace them with better wording.


KirankumarAmbati:
**CC:** @bvaughn @acdlite

Issue #16952, opened 2019-09-29
Devtools extension does not show elements from local files


nickmccurdy:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**
When a local file is loaded, the devtools extension displays a panel in devtools, but all elements are missing. There are no console warnings or errors. All elements display properly when using a server, but the documentation says a server is not necessary in this case.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

1. Put this in `index.html`
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8" />
    <title>Hello World</title>
    <script src="https://unpkg.com/react@16.10.1/umd/react.development.js"></script>
    <script src="https://unpkg.com/react-dom@16.10.1/umd/react-dom.development.js"></script>
  </head>
  <body>
    <div id="root"></div>
    <script type="text/babel">
      function HelloWorld() {
        return React.createElement("h1", null, "Hello, world!");
      }
      ReactDOM.render(
        React.createElement(HelloWorld, null),
        document.getElementById("root")
      );
    </script>
  </body>
</html>
```
2. Install React devtools extension and allow access to local file URLS
3. Open `index.html`
4. Open devtools
5. Open components panel

**What is the expected behavior?**
Hello element is displayed with default settings, and both Hello and h1 are displayed when the host type is not hidden in settings.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
- React/ReactDOM 16.10.1 UMD from Unpkg CDN
- React Developer Tools (Chrome extension) 4.1.2
- Chrome 77.0.3865.90
- macOS 10.14.6

bvaughn:
`chrome.devtools.inspectedWindow.eval` is throwing:
> DOMException: Failed to execute 'postMessage' on 'Window': Invalid target origin 'null' in a call to 'postMessage'.

Looks like a regression caused by #16900. Reverting that commit fixes DevTools.

cc @linshunghuang

linshunghuang:
Yes definitely caused by that change. My guess is we'll need to change the postMessage target from window.origin to '*'

bvaughn:
[...]

PR #16953, opened 2019-09-29
[DevTools] targetOrigin of postMessage needs to be '*' for local files

linshunghuang:
When we (extension) use `postMessage` to send a message to the content script (on the same window) to inject the backend script, the `targetOrigin` was set to `window.origin` here
https://github.com/facebook/react/blob/master/packages/react-devtools-extensions/src/main.js#L138

However, this won't work for local files (`window.origin` is `null`) so I'm changing the `targetOrigin` from `window.origin` to `'*'`. Other existing `postMessage` calls in contentScript.js does so as well.

https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#Using_window.postMessage_in_extensions:
> posting a message to a page at a file: URL currently requires that the targetOrigin argument be "*"

I tested with the local index.html file provided in https://github.com/facebook/react/issues/16952 (but removed `type="text/babel"`) and confirmed that the components tab was empty before this change, and working after this change.

### DIFF
--- a/packages/react-devtools-extensions/src/main.js
+++ b/packages/react-devtools-extensions/src/main.js
@@ -135,7 +135,7 @@ function createPanelIfReactLoaded() {
         // Initialize the backend only once the Store has been initialized.
         // Otherwise the Store may miss important initial tree op codes.
         chrome.devtools.inspectedWindow.eval(
-          `window.postMessage({ source: 'react-devtools-inject-backend' }, window.origin);`,
+          `window.postMessage({ source: 'react-devtools-inject-backend' }, '*');`,
           function(response, evalError) {
             if (evalError) {
               console.error(evalError);


packages/react-devtools-extensions/src/main.js
```diff
@@ -135,7 +135,7 @@ function createPanelIfReactLoaded() {
         // Initialize the backend only once the Store has been initialized.
         // Otherwise the Store may miss important initial tree op codes.
         chrome.devtools.inspectedWindow.eval(
-          `window.postMessage({ source: 'react-devtools-inject-backend' }, window.origin);`,
+          `window.postMessage({ source: 'react-devtools-inject-backend' }, '*');`,
```
bvaughn: Should we use `window.origin || "*"` or will `"*"` be sufficient for our Trusted Types case too?

linshunghuang: That looks good to me too but "*" will also be sufficient

bvaughn: Ok! Just wanted to verify. If "*" will cover both use cases, let's stick with that.

PR #16958, opened 2019-09-30
[react-interactions] Add tab handling to FocusList

trueadm:
This is a follow up to https://github.com/facebook/react/pull/16922, where we add the same tab keyboard control behavior to `FocusList`

### DIFF
--- a/packages/react-interactions/events/src/dom/testing-library/index.js
+++ b/packages/react-interactions/events/src/dom/testing-library/index.js
@@ -158,6 +158,33 @@ function testWithPointerType(message, testFn) {
   });
 }
 
+function emulateBrowserTab(backwards) {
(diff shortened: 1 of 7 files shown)


packages/react-interactions/events/src/dom/testing-library/index.js
```diff
@@ -158,6 +158,33 @@ function testWithPointerType(message, testFn) {
   });
 }
 
+function emulateBrowserTab(backwards) {
```
necolas: Please can you move this into the test file you're using it in, for now

trueadm: Sure. Will do in a follow up! :)

PR #16971, opened 2019-10-01
[react-interactions] Add allowModifiers flag to FocusList + FocusTable

trueadm:
This PR adds the `allowModifiers` flag to the `FocusList` and `FocusTable` components. This PR also changes the default behaviour for using arrow keys with modifier keys (they're now no-ops). I also moved some of the test code into a different module relating to `emulateBrowserTab` as per a follow up to @necolas's feedback.


packages/react-interactions/accessibility/src/FocusList.js
```diff
@@ -117,25 +127,34 @@ export function createFocusList(scope: ReactScope): Array<React.Component> {
...
-                  for (let i = 0; i < nodes.length; i++) {
-                    const node = nodes[i];
-                    if (node !== activeNode) {
-                      setElementCanTab(node, false);
-                    } else {
-                      setElementCanTab(node, true);
-                    }
+            const key = event.key;
+
+            if (key === 'Tab') {
```
bvaughn: Given that "shift" is a modifier, and it impacts "Tab" (reverses the direction) it seems odd that we check for `key === 'Tab'` _before_ bailing out if modifiers aren't allowed. Is this a bug?

trueadm: The modifiers logic is only for the keyboard arrow keys, not for Tab.

bvaughn: I see. That wasn't super clear to me just looking at the code and going off of the name of the prop. Maybe some inline comment clarifying the intent could be helpful?

packages/react-interactions/accessibility/src/FocusList.js
```diff
@@ -82,6 +83,13 @@ function getListProps(currentCell: ReactScopeMethods): Object {
   return {};
 }
 
+function hasModifierKey(event: KeyboardEvent): boolean {
```
necolas: This function is already in the 'shared' module

PR #16973, opened 2019-10-01
Fixing grammatical errors in error message

RaneWallin:
Fixing a small grammatical error in the error message, neither/nor clause and that/which confusion.


pconley:
you probably already figured this out, but you also need to change the test at 
https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js#L761

PR #16974, opened 2019-10-01
Add more descriptive error when a function call is passed to use memo

mattfwood:

image

Currently if you accidentally pass a function call to useMemo instead of a function reference, you get an error like the one above, which doesn't really point a dev in the right direction.

```javascript
// right:
useMemo(() => calculate(numbers), [numbers])

// wrong
useMemo(calculate(1,2), [numbers]) // => throws 'nextCreate is not a function'
```

I've added an invariant warning if anything other than a function is passed to `useMemo`


mattfwood:
Wanted to check on this, I recently had a handful of developers make this mistake while learning hooks and I think it's a fairly common place to make a mistake. I'm definitely open to re-wording or changing the warning or open to suggestions of how else it could be checked.

gaearon:
We can add a dev-only warning, but we shouldn't add more invariants to this hot path.

mattfwood:
@gaearon Got it, that makes sense. Would I be able to have this re-opened so I can look into making that change?

PR #16983, opened 2019-10-02
remove unused code in commitBeforeMutationLifeCycles

NE-SmallTown:
Since `Snapshot` is **only for class component**, and now we only call `commitBeforeMutationLifeCycles` when `(nextEffect.effectTag & Snapshot) !== NoEffect`, that means  we only call `commitBeforeMutationLifeCycles` **for class component**.

So, I think we can remove other fiber tag logic in `commitBeforeMutationLifeCycles`


NE-SmallTown:
close for wrong git commit order

PR #16989, opened 2019-10-02
Added trace updates feature (DOM only)

### LINKED ISSUE #16437: Devtools V4: Where is Highlight Updates?
If I understood correctly, this is the correct repository for devtools v4, right?

I just noticed that react devtool were updated. I'm missing the "Highlight Updates" function.
How can I activate it?

image

image

Version: 4.0.2 (8/15/2019)

bvaughn:
Resolves #16437

## Caveats
* It is only enabled in the browser extension (and `react-devtools-inline` NPM package) for now, so it only supports React DOM.
* It is not implemented for legacy renderers (v15) although it could be added by someone if they wanted to submit a follow up PR

## Demo
Trace updates demo

### DIFF
--- a/packages/react-devtools-shared/src/backend/agent.js
+++ b/packages/react-devtools-shared/src/backend/agent.js
@@ -204,22 +204,6 @@ export default class Agent extends EventEmitter<{|
     return null;
   }
 
-  getNodesForID(id: number): Array<Object> | null {
-    for (let rendererID in this._rendererInterfaces) {
-      const renderer = ((this._rendererInterfaces[
-        (rendererID: any)
-      ]: any): RendererInterface);
-
-      try {
-        const nodes = renderer.findNativeNodesForFiberID(id);
-        if (nodes != null) {
-          return nodes;
-        }
-      } catch (error) {}
-    }
-    return null;
-  }
-
   getProfilingData = ({rendererID}: {|rendererID: RendererID|}) => {
     const renderer = this._rendererInterfaces[rendererID];
     if (renderer == null) {
--- a/packages/react-devtools-shared/src/backend/agent.js
+++ b/packages/react-devtools-shared/src/backend/agent.js
@@ -21,10 +21,15 @@ import {
   sessionStorageSetItem,
 } from 'react-devtools-shared/src/storage';
 import setupHighlighter from './views/Highlighter';
+import {
+  initialize as setupTraceUpdates,
+  toggleEnabled as toggleTraceUpdatesEnabled,
+} from './views/TraceUpdates';
 import {patch as patchConsole, unpatch as unpatchConsole} from './console';
 
 import type {BackendBridge} from 'react-devtools-shared/src/bridge';
 import type {
+  FindNativeNodesForFiberID,
   InstanceAndStyle,
   NativeType,
   OwnersList,
@@ -87,6 +92,7 @@ export default class Agent extends EventEmitter<{|
   hideNativeHighlight: [],
   showNativeHighlight: [NativeType],
   shutdown: [],
+  traceUpdates: [Map<number, FindNativeNodesForFiberID>],
 |}> {
   _bridge: BackendBridge;
   _isProfiling: boolean = false;
@@ -143,6 +149,7 @@ export default class Agent extends EventEmitter<{|
       this.updateAppendComponentStack,
     );
     bridge.addListener('updateComponentFilters', this.updateComponentFilters);
+    bridge.addListener('updateTraceUpdates', this.updateTraceUpdates);
     bridge.addListener('viewElementSource', this.viewElementSource);
 
     if (this._isProfiling) {
@@ -159,6 +166,7 @@ export default class Agent extends EventEmitter<{|
     bridge.send('isBackendStorageAPISupported', isBackendStorageAPISupported);
 
     setupHighlighter(bridge, this);
+    setupTraceUpdates(this);
   }
 
   get rendererInterfaces(): {[key: RendererID]: RendererInterface} {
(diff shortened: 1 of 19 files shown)


packages/react-devtools-shared/src/backend/agent.js
```diff
@@ -196,6 +204,22 @@ export default class Agent extends EventEmitter<{|
...
+      try {
+        const nodes = renderer.findNativeNodesForFiberID(id);
+        if (nodes != null) {
+          return nodes;
+        }
+      } catch (error) {}
+    }
+    return null;
+  }
+
```
bvaughn: Whoops, I meant to remove this function. It isn't actually being used.

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -1430,11 +1463,35 @@ export function attach(
...
+        // If an ancestor updated, we should mark the nearest host nodes for highlighting.
+        if (elementType === ElementTypeHostComponent) {
+          traceUpdatesForNodes.add(nextFiber.stateNode);
+          traceNearestHostComponentUpdate = false;
+        }
+      } else {
+        if (
+          elementType === ElementTypeFunction ||
+          elementType === ElementTypeClass ||
+          elementType === ElementTypeContext
```
sebmarkbage: What if `traceNearestHostComponentUpdate` is false because the parent didn't rerender, but the props to the host component has changed. Seems like you'd want to include `ElementTypeHostComponent` here too, right?

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -1244,6 +1250,26 @@ export function attach(
...
+        ) {
+          // Otherwise if this is a traced ancestor, flag for the nearest host descendant(s).
+          traceNearestHostComponentUpdate = true;
```
sebmarkbage: This is a bit odd. If there are two nested HostComponents then only the outer one is marked but as soon as there's  an intermediate function it is not?

What's the expected effect for a mounted subtree? Should only the inserted Host node be highlighted or is it every node in the subtree?

packages/react-devtools-shared/src/backend/renderer.js
bvaughn: How would the props to a host component change without a parent component re-rendering?

packages/react-devtools-shared/src/backend/renderer.js
bvaughn: I think this is the desired behavior- that we only highlight the _nearest_ host components.

This is how the old feature worked. If a composite component or context type updated, we highlighted on the nearest node:
https://github.com/facebook/react-devtools/blob/d839081f79f681617df7584a8ffd2f4163fa40b9/plugins/TraceUpdates/TraceUpdatesBackendManager.js#L59-L66

packages/react-devtools-shared/src/backend/renderer.js
sebmarkbage: Good question. I guess it's special cases like: `<div hidden="foo"><div prop={...} /></div>`  The prop doesn't change in the same commit as the parent rerendering.

Also for nested host  components: `Foo = ({change}) => <div><div prop={change} /></div>`

The outer parent rerenders but the "traceNearestHostComponentUpdate" stops at the outer host component but you want to mark the inner one as having changed.

packages/react-devtools-shared/src/backend/renderer.js
sebmarkbage: That's for updates but this is the mount path.

bvaughn: My mistake. I misunderstood what you were saying. I agree that I should not flip this to true.

packages/react-devtools-shared/src/backend/renderer.js
bvaughn: WRT the nested example above, I think it still makes sense to highlight the outermost node, because that's the one associated with the `Foo` component (which is the thing that's actually re-rendering).

I think this makes sense because we highlight the outer `<div>` in the case where `Foo` re-renders but doesn't cause any host effects at all.

bvaughn: Chatted out of band and agreed that the current behavior is okay.

In the first case, it's likely that a nested component component would also re-render and so we would highlight its nodes.

In the second case, highlighting the outer `<div>` only is probably sufficient because this is just meant to be a coarse signal that a composite component re-rendered- not a fine grained tool like the Profiler.

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -1244,6 +1250,20 @@ export function attach(
       recordMount(fiber, parentFiber);
     }
 
+    if (traceUpdatesEnabled) {
+      const elementType = getElementTypeForFiber(fiber);
```
sebmarkbage: You can move this inside the `if (traceNearestHostComponentUpdate)` condition now that you only use it in one of  the branches.

slashwhatever:
So anyone using React 15 loses this feature?

bvaughn:
As the description says:
> It is not implemented for legacy renderers (v15) although it could be added by someone if they wanted to submit a follow up PR

Other than this, you could always install v3 of the DevTools extension for React v15.

cmacdonnacha:
Are there any docs to explain what the different colours represent when this setting is enabled?

cmacdonnacha:
> @cmacdonnacha The colours are synchronised with what you see in the Profiler and follow the same convention, moving from a cool blue for the least problematic (fewest renders) through green to orange for the most problematic (most renders).
> 
> You can see the full set of colours here: https://github.com/facebook/react/blob/main/packages/react-devtools-shared/src/backend/views/TraceUpdates/canvas.js

Thanks, but are they in the docs somewhere?

PR #16998, opened 2019-10-03
correct effect tag in LayoutEffect

NE-SmallTown:
Layout effect should use layout effect tag both for create and destroy rather than Mutation tag

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -34,6 +34,7 @@ import {
   MountLayout,
   UnmountPassive,
   MountPassive,
+  UnmountLayout,
 } from './ReactHookEffectTags';
 import {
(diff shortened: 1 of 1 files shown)


acdlite:
Looks like we're missing a test for the current behavior, though. The intent is that _all_ layout clean-up functions should fire before _any_ of the layout create effects _in the whole tree_.

NE-SmallTown:
Yea, but I think use `UnmountLayout` also could do that because we always call destroy func **first** and then call create func in `commitHooksEffectList`

NE-SmallTown:
Thanks. Could we call **all** unmount/destroy func first when we traverse the effect list in `commitHooksEffectList` and then call all mount/create func?

The main concern is that the tag "mismatch" is a little confusing. Anyway, if finally we don't want any change about this, I could add a test later

PR #17000, opened 2019-10-03
[react-interactions] Remove context.setTimeout & context.clearTimeout

trueadm:
This PR aims at reducing some of the size of the new responder event system. We have the concept of `setTimeout` and `clearTimeout` on responder context to handle async timers. This was used a while ago, but we removed any usage it from the event responder and thus have no real need to keep the code around anymore.


necolas:
It would be needed by LongPress

trueadm:
We can handle timers in the component rather than in the responder, which allows us to possibly use the scheduler cc @acdlite.

necolas:
In what component? LongPress is inherently related to timers

PR #17005, opened 2019-10-03
[React-DevTools] Support Shift+Enter to go to previous search result

Bo-Duke:
I added support for Shift + Enter while the search bar is in focus in the DevTools, it's a widely used shortcut to go to the previous search results instead of the next one.

As it wasn't a big change, I didn't create an issue, hope that's ok.


bvaughn:
Released today in v4.4.0

PR #17012, opened 2019-10-04
chore: upgrade jsdom

SimenB:
Following up on #15778, this is needed before being able to upgrade to Jest 25, which will have some significant performance improvements. Some tests for `style` properties are failing though, and I have no idea if it's due to jsdom or the tests themselves. I have some hope you'll know what's up yourselves 🤞  

I tried upgrading to the Jest 25 alpha btw, and it fails in the same way, so if it's an issue with jsdom itself, it has not been fixed for jsdom@15 and it'd be great to report it to them.

/cc @threepointone

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMComponent-test.js
@@ -2425,7 +2425,7 @@ describe('ReactDOMComponent', () => {
       const container = document.createElement('div');
 
       ReactDOM.render(<img src={obj} />, container);
-      expect(container.firstChild.src).toBe('hello');
+      expect(container.firstChild.src).toBe('http://localhost/hello');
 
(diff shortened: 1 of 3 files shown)


threepointone:
Bumping to extend life, this seems like one we’ll need

SimenB:
Jest 25 is released, but you cannot upgrade due to this error

eps1lon:
> Jest 25 is released, but you cannot upgrade due to this error

```diff
- element.style.foo
+ element.style.getPropertyValue('--foo')
```

which [works in a browser](https://codesandbox.io/s/css-custom-properties-access-423ni). The previous test wouldn't even pass in a browser (or at least not in chrome 79). 

I quickly tried it on this branch but had no success. Do you have a branch with jest 25 (and jsdom 15) that I can check-out?

SimenB:
I don't, but I can open up a separate PR for it, 2 sec

SimenB:
#17896

PR #17019, opened 2019-10-04
[react-interactions] Add FocusTable colSpan support

trueadm:
This PR adds support for `colSpan` on `FocusTable` cells. It works very much like how `<td colSpan={2}>` might work with the DOM. `colSpan` makes cells span a numerical number of columns, allowing for rows with a dynamic number of cells to continue to work.

This PR also removes the unnecessary `focusTableByID` and `id` logic, in favor of declarative nested tables (which compose better). Lastly, the `wrap` prop has been split into `wrapX` and `wrapY`, as per internal feedback.

### DIFF
--- a/packages/react-interactions/accessibility/src/FocusTable.js
+++ b/packages/react-interactions/accessibility/src/FocusTable.js
@@ -17,6 +17,7 @@ import setElementCanTab from './shared/setElementCanTab';
 type FocusCellProps = {
   children?: React.Node,
   onKeyDown?: KeyboardEvent => void,
+  colSpan?: number,
 };
 
 type FocusRowProps = {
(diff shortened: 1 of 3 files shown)


packages/react-interactions/accessibility/src/FocusTable.js
```diff
@@ -69,30 +70,59 @@ function focusScope(cell: ReactScopeMethods, event?: KeyboardEvent): void {
...
   if (row !== null && row.getProps().type === 'row') {
     const cells = row.getChildren();
     if (cells !== null) {
-      const rowIndex = cells.indexOf(currentCell);
-      return [cells, rowIndex];
+      const [rowIndex, rowIndexWithColSpan] = getCellIndexes(
+        cells,
+        currentCell,
+      );
+      return [cells, rowIndex, rowIndexWithColSpan];
```
elboman: Do we even need to return the `rowIndex` without the colSpan?

trueadm: Yeah, we need it for when we move left and right still.

trueadm:
Merging this as I plan to move these components to our internal codebase, away from Github.

PR #17023, opened 2019-10-05
[DevTools] Display RegExp values in props/state

Zirak:
Previously, when props/state contained a regexp, it was shown as an
empty object. This commit adds regexps as values in need of special
rehydration (like Symbols or TypedArrays), and display them as a user
might expect.

These values may be editable, but as regexps are not valid JSON it might
depart from the usual editing functionality, which may not be worth the
hassle just to support regexps.

Before:
image

After:
image

### DIFF
--- a/packages/react-devtools-shared/src/__tests__/__snapshots__/inspectedElementContext-test.js.snap
+++ b/packages/react-devtools-shared/src/__tests__/__snapshots__/inspectedElementContext-test.js.snap
@@ -503,6 +503,7 @@ exports[`InspectedElementContext should support complex data types: 1: Inspected
       "1": {}
     },
     "react_element": {},
+    "regexp": {},
     "set": {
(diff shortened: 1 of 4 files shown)


bvaughn:
Hey sorry for not seeing this sooner. I like this feature, but the underlying code has changed a lot since your PR. I'm going to try to resolve it locally. Stay tuned.

bvaughn:
I'm going to incorporate your branch, fix the conflicts, and add some more tests. Closing this PR in favor of #17690. (Your commit will stay in the history though!)

Zirak:
Awesome, thanks!

bvaughn:
FYI this change just went out as part of v4.4!

Zirak:
woo, thanks for the follow-up, especially on a small PR like this!

PR #17026, opened 2019-10-06
When copying tree to clipboard ensure the full tree is copied

### LINKED ISSUE #16924: react-devtools: tiny feature request (copy to clipboard related)
Hey guys, great job with the new devtools 💯 perhaps this request can be put in the backlog for a future release as I think it could be quite useful and could save a few steps for developers. 

- Copying data to clipboard stringifies all key-value pairs of an object
- When the keys' values happen to be objects or arrays, the values are given in constructor form rather than seeing the contents of that object/array

This is what was copied:
Screen Shot 2019-09-27 at 12 24 19

This is a pasted version in VSCode:
Screen Shot 2019-09-27 at 12 24 48

**What is the desired behavior?**
- When I click on "Copy to clipboard" in react-devtools, my copied object's data _keys_ and _values_ are in an unstringified format, or at least, I am presented the option to have this copied in an unstringified format
- The _values_ of each key is readable, e.g: if the value is an object, I can see the expanded object and all its key-value pairs clearly as shown here (this is logged into Chrome console from react-devtools)
Screen Shot 2019-09-27 at 12 29 11

ChrisDobby:
Added a new event - getDataForCopy - that accepts an element id and property as a parameter, gets that property of the element, formats it for copying and returns it in the dataForCopy.

Fixes #16924

### DIFF
--- a/packages/react-devtools-shared/src/backend/agent.js
+++ b/packages/react-devtools-shared/src/backend/agent.js
@@ -50,6 +50,9 @@ const debug = (methodName, ...args) => {
   }
 };
 
+const invalidRendererWarning = (rendererID: number, id: number) =>
+  `Invalid renderer id "${rendererID}" for element "${id}"`;
+
 type ElementAndRendererID = {|
   id: number,
   rendererID: number,
@@ -227,7 +230,7 @@ export default class Agent extends EventEmitter<{|
   getOwnersList = ({id, rendererID}: ElementAndRendererID) => {
     const renderer = this._rendererInterfaces[rendererID];
     if (renderer == null) {
-      console.warn(`Invalid renderer id "${rendererID}" for element "${id}"`);
+      console.warn(invalidRendererWarning(rendererID, id));
     } else {
       const owners = renderer.getOwnersList(id);
       this._bridge.send('ownersList', ({id, owners}: OwnersList));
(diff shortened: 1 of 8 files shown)


packages/react-devtools-shared/src/backend/agent.js
```diff
@@ -254,6 +261,18 @@ export default class Agent extends EventEmitter<{|
     }
   };
 
+  getDataForCopy = ({id, property, rendererID}: GetDataForCopyParams) => {
+    const renderer = this._rendererInterfaces[rendererID];
+    if (renderer == null) {
+      console.warn(`Invalid renderer id "${rendererID}" for element "${id}"`);
```
GavBaros: Tiny change request : line 267, `Invalid renderer id "${rendererID}" for element "${id}"` is repeated about 10 times in this file, is if worth it if we extract this to a meaningful const at the top of the file? 🙃  

This approach will eventually bring in mind issues (in a separate PR / NFR in the backlog ) such as:
- If we are repeating texts across a file, are those texts being used across other files?
- If so, how about centralising texts?
- Is centralising the texts worth it (i18n etc)? Will those texts eventually need translation as react grows to an international audience (eg: China)?

ChrisDobby: Sounds sensible to me.  Have updated the PR

PR #17035, opened 2019-10-07
Remove dormant createBatch experiment

acdlite:
In a hybrid React app with multiple roots, `createBatch` is used to coordinate an update to a root with its imperative container.

We've pivoted away from multi-root, hybrid React apps for now to focus on single root apps.

This PR removes the API from the codebase. It's possible we'll add back some version of this feature in the future.


sebmarkbage:
Are there any complexities in the workloop that can be simplified by removing this feature?

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -418,64 +224,24 @@ function ReactRoot(container: DOMContainer, options: void | RootOptions) {
 ReactRoot.prototype.render = ReactSyncRoot.prototype.render = function(
   children: ReactNodeList,
   callback: ?() => mixed,
-): Work {
```
acdlite: @sebmarkbage I also changed this to return nothing instead of a thenable. It still accepts a callback. I actually forgot that we supported both forms... I don't think this was intentional? Let's choose one, or neither. I don't have a preference; I went with the callback for now because it's the simpler implementation.

sebmarkbage:
I believe some of the exports from ReactFiberReconciler aren’t used by any renderer so we can delete those too. Although hopefully they are dce in the build already.

acdlite:
@sebmarkbage Any reason `attemptSynchronousHydration` is in the reconciler? If I move that to the work loop, I can get rid of `flushRoot`.

sebmarkbage:
It’s called by a renderer so needs to be exported by the reconciler since that’s the public api. But you can in-line it one way or another. It’s in the reconciler because it also does changes to the retry time and seemed like we didn’t want to bloat the work loop with too many concerns.

sebmarkbage:
You can maybe simplify flushRoot though to something simpler.

acdlite:
Ok I'll rename it and remove the invariant

PR #17042, opened 2019-10-08
[react-interactions] Adds more experimental Scope API methods

trueadm:
This PR adds two more internal API methods to React Scopes. Note: these are experimental and may change at any time. Notably, the changes are:

- Rename `getScopedNodes` -> `getAllNodes`. Rename to make way for new method below.
- Add `getFirstNode`. Gets only the 1st matching host component.
- Add `containsNode`. Validates that any given HTML node is within the sub-tree of a scope.

These APIs mainly aim at adding missing functionality that would result in computationally expensive work-arounds in user-land.

### DIFF
--- a/packages/react-interactions/accessibility/src/FocusGroup.js
+++ b/packages/react-interactions/accessibility/src/FocusGroup.js
@@ -30,7 +30,7 @@ type FocusGroupProps = {|
 const {useRef} = React;
 
 function focusGroupItem(cell: ReactScopeMethods, event: KeyboardEvent): void {
-  const firstScopedNode = cell.getFirstScopedNode();
+  const firstScopedNode = cell.getFirstNode();
   if (firstScopedNode !== null) {
     firstScopedNode.focus();
     event.preventDefault();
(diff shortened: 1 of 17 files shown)


packages/react-interactions/accessibility/src/FocusGroup.js
```diff
@@ -29,11 +29,13 @@ type FocusGroupProps = {|
 
 const {useRef} = React;
 
-function focusGroupItem(cell: ReactScopeMethods, event: KeyboardEvent): void {
-  const tabbableNodes = cell.getScopedNodes();
-  if (tabbableNodes !== null && tabbableNodes.length > 0) {
-    tabbableNodes[0].focus();
-    event.preventDefault();
+function focusGroupItem(cell: ReactScopeMethods, event?: KeyboardEvent): void {
```
germtb: I guess `event` is sometimes nullable but it was missed before?

trueadm: Well spotted!

Issue #9189, opened 2017-03-15
Expose DEV-mode warnings in devtools UI


bvaughn:
React has made recent developer experience improvements to lifecycle error handling and logging. Errors thrown during render can be recovered from using ~~`unstable_handleError`~~ `componentDidCatch`. Errors thrown during lifecycle methods are also automatically logged to the console with the component stack location to make them easier to identify (see #8785).

The team has discussed similar improvements for warnings- such as using a "yellow box" approach to make warnings stand out more in developer mode (see PRs #7360 and #8861). There has been some pushback though to the idea of React modifying the DOM for warning purposes (see issue #8784). I'm not sure how that will play out yet- but in the meanwhile, could we improve at least some of these use cases by making better use of the devtools?

For example, what if React exposed a new API that allowed associating a warning with one or more components in the devtools panel? (eg a method that- when called- recorded the current component stack and notified devtools if present) React could use this for things like missing or non-unique keys. 3rd party libraries may also benefit from this (eg react-virtualized could use this to warn about missing or incomplete positioning styles for cells).

Here's a rough outline of some of the features:
* Components with warnings could be highlighted in some emphasized way (eg yellow background) in devtools to make them easier to spot.
* A new toolbar option could be added to "Show All Warnings" (eg filter the tree view to show only components with warnings).
* Warning message could be shown inline in the settings panel for the selected component.

Here's a rough mockup: 
![screen shot 2017-03-15 at 3 47 31 pm](https://cloud.githubusercontent.com/assets/29597/23974108/bc3dabd6-0996-11e7-8d2a-e058c4cafe41.png)

Limitations:
* Not everyone uses devtools so this would not help a lot of people.
* This would not address issues like using the development mode of React for production sites.

Thoughts? Ideas? Suggestions?

gaearon:
Beautiful 😍

LXSMNSYC:
Any status on this? This seems to be a very helpful feature

bvaughn:
@LXSMNSYC No updates. This was just an idea that hasn't been high priority enough for me to experiment with it. The DevTools backend has all of the info needed to associate errors/warnings with specific components in the tree– but we'd need to:
* Pass the information from the backend to the frontend with a new type of message.
* Design and implement new UI in the Components tree and the inspected element to surface the messages.
* Figure out what to do if the component is filtered out of the tree. (Do we not show the warning? Do we show it on a parent or a descendent?)
[...]

PR #17053, opened 2019-10-09
[DO NOT MERGE] Prototype showing component errors/warnings within the Components tree

bvaughn:
Relates to #9189

Last night I played around a bit with the idea of surfacing dev warnings/errors in the Components tree UI:
Screen Shot 2019-10-08 at 9 03 32 PM

I'm not sure this is useful or compelling enough to warrant completing, and there are a few significant "TODOs", but I'm pushing in case myself or someone else decides to play with it some more in the future. (I'm not going to follow up with it for the time being.)

### DIFF
--- a/packages/react-devtools-shared/src/backend/agent.js
+++ b/packages/react-devtools-shared/src/backend/agent.js
@@ -37,7 +37,7 @@ import type {
   RendererID,
   RendererInterface,
 } from './types';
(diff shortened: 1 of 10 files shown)


mmontag:
This could be nice for troubleshooting those "unique key" errors.

bvaughn:
Interesting idea but not something I'm actively working on.

PR #17058, opened 2019-10-10
Migrate useDeferredValue and useTransition

lunaruan:
Migrated `useDeferredValue` and `useTransition` from Facebook's www repo into `ReactFiberHooks`.


sebmarkbage:
As part of this migration I intended to drop the “Suspense” part. Let’s just call them useDeferredValue and useTransition.

The Suspense part was to bring them into a brand but going forward, it’s just going to be annoying to have the longer name.

It’s also a misnomer since given this set up, this is the primary API to call into the Concurrent Mode APIs even if you don’t use Suspense at all.

sebmarkbage:
You can probably update this warning now. https://github.com/facebook/react/blob/5a71cbe7a9e192c23b847be6e0c575b6705f6939/packages/react-reconciler/src/ReactFiberWorkLoop.js#L2931-L2933

bvaughn:
We'll also need an accompanying PR to update the hooks API docs:
https://reactjs.org/docs/hooks-faq.html

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -25,21 +28,27 @@ let useCallback;
...
 
 describe('ReactHooksWithNoopRenderer', () => {
   beforeEach(() => {
     jest.resetModules();
+    jest.useFakeTimers();
 
     ReactFeatureFlags = require('shared/ReactFeatureFlags');
     ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
     ReactFeatureFlags.enableSchedulerTracing = true;
+    ReactFeatureFlags.flushSuspenseFallbacksInTests = false;
```
acdlite: Why is this necessary? Can we rewrite the test somehow? I can pair with you on this if needed

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
```diff
@@ -27,6 +27,7 @@ describe('ReactHooks', () => {
 
     ReactFeatureFlags = require('shared/ReactFeatureFlags');
     ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
+    ReactFeatureFlags.enableStableConcurrentModeAPIs = true;
```
acdlite: If you rebase on master, you can wrap the concurrent tests with an `__EXPERIMENTAL__` check instead.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1117,6 +1128,96 @@ function updateMemo<T>(
...
+  const [isPending, setPending] = mountState(false);
+  const startTransition = mountCallback(
+    callback => {
+      setPending(true);
+      Scheduler.unstable_next(() => {
+        const previousConfig = ReactCurrentBatchConfig.suspense;
+        ReactCurrentBatchConfig.suspense = config === undefined ? null : config;
+        try {
+          setPending(false);
+          callback();
```
Jessidhia: Does the callback run as a potentially separate commit, or does being inside `Scheduler.next` prevent non-explicit "unbatching"?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1117,6 +1128,96 @@ function updateMemo<T>(
...
+        } finally {
+          ReactCurrentBatchConfig.suspense = previousConfig;
+        }
+      });
+    },
+    [config, isPending],
+  );
+  return [startTransition, isPending];
```
Jessidhia: This is the only built-in hook producing a pair which returns a function as the first, instead of the second element. This will take some education.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1117,6 +1128,96 @@ function updateMemo<T>(
...
+    () => {
+      Scheduler.unstable_next(() => {
+        const previousConfig = ReactCurrentBatchConfig.suspense;
+        ReactCurrentBatchConfig.suspense = config === undefined ? null : config;
+        try {
+          setValue(value);
+        } finally {
+          ReactCurrentBatchConfig.suspense = previousConfig;
+        }
+      });
```
bvaughn: Why do we not pass the `TimeoutConfig` to `Scheduler.unstable_next` in the mount case, when we do in the update case?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1117,6 +1128,96 @@ function updateMemo<T>(
...
+    () => {
+      Scheduler.unstable_next(() => {
+        const previousConfig = ReactCurrentBatchConfig.suspense;
+        ReactCurrentBatchConfig.suspense = config === undefined ? null : config;
+        try {
+          setValue(value);
+        } finally {
+          ReactCurrentBatchConfig.suspense = previousConfig;
+        }
+      }, config);
```
sebmarkbage: This config doesn't seem right. Probably a copy paste error.

packages/react-reconciler/src/ReactFiberHooks.js
sebmarkbage: Nice catch. `unstable_next` doesn't accept a second argument. It must have been a copypasta from withSuspenseConfig.

lunaruan: Yeah, it was accidental copy/paste. Deleted it! :)

packages/react-reconciler/src/ReactFiberHooks.js
markerikson: That does seem odd.  Should it be flipped?

sebmarkbage: The first argument is useful without the second but not the inverse.

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -1922,6 +1975,212 @@ describe('ReactHooksWithNoopRenderer', () => {
...
+
+      act(() => {
+        _setText('B');
+      });
+      expect(Scheduler).toHaveYielded([
+        'B',
+        'A',
+        'B',
+        'Suspend! [B]',
+        'Loading',
```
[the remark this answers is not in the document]
sebmarkbage: The yields are not important. The output is important. If it was using Suspense, the output would be `[span('A'), span('Loading')]` which isn't desirable here.

sebmarkbage:
I wonder if we should remove `withSuspenseConfig`. It has the benefit that it can operate standalone from a specific component which might be useful in edge cases but we don't plan on documenting it early on.

If we do, then the question becomes whether we should move ReactCurrentBatchConfig into the renderer since it can now be local to a component which simplifies things a bit. Especially versioning but limits our ability to move it to become cross-renderer in the future.

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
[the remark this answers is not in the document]
Jessidhia: I'd expect `Suspense` to hold off on committing the subtree at all (so it'd stay showing `span('A')`) until the expirationTime is reached.

...though I guess `act()` forces an expirationTime flush.

ConAntonakos:
Are these being planned as hook versions for components like `Suspense` and `lazy` or am I way off?

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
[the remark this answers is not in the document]
sebmarkbage: In that case we couldn't commit the outer A so it would be `[span('A'), span('A')]`. We have another way to do that which is what `useTransition` is for.

We never commit anything in an inconsistent state by default. So either we commit A in both slots or B in both slots. Early explorations included ways of leaving some trees inconsistent but the downstream effects on that in a reactive system was too difficult to reason about.

`useDeferredValue` is a limited form of explicitly opting in to inconsistency for a particular value.

packages/react-reconciler/src/ReactFiberHooks.js
gaearon: The intention here is to:

1. Set pending false at high priority
2. Batch setting pending to false and _doing the actual state change_ at lower priority

As a result, you'll see pending immediately turn true, but then it turn false together with the actual state change you wanted to do.

packages/react-reconciler/src/__tests__/ReactHooks-test.internal.js
gaearon: This can be removed now

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -377,6 +377,28 @@ class ReactShallowRenderer {
...
+      return [startTransition, false];
+    };
+
+    // TODO: implement if we decide to keep the shallow renderer
+    const useDeferredValue = <T>(value: T, config): T => {
+      this._validateCurrentlyRenderingComponent();
+      warning(
+        false,
+        'useDeferredValue is not implemented for shallow renderers',
+      );
```
gaearon: I don't think we need these warnings. We should be able to just remove them and keep the rest of this code as is.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1117,6 +1128,96 @@ function updateMemo<T>(
...
+        ReactCurrentBatchConfig.suspense = config === undefined ? null : config;
+        try {
+          setPending(false);
+          callback();
+        } finally {
+          ReactCurrentBatchConfig.suspense = previousConfig;
+        }
+      });
+    },
+    [config, isPending],
```
bvaughn: Why do we pass `isPending` as a dependency here?

PR #17062, opened 2019-10-10
apply changes on editablevalue on blur feature implemented

muratcatal:
Added applying changes on EditableValue component on blur event.


packages/react-devtools-shared/src/devtools/views/Components/EditableValue.js
```diff
@@ -63,6 +61,17 @@ export default function EditableValue({
     }
   };
 
+  const handleBlur = event => {
+    event.stopPropagation();
```
bvaughn: Why do we stop prop?

bvaughn:

Thank you!

PR #17063, opened 2019-10-10
Pass prod error messages directly to constructor

acdlite:
Updates the `invariant` transform to pass an error message string directly to the Error constructor, instead of mutating the message property.

Turns this code:

```js
invariant(condition, 'A %s message that contains %s', adj, noun);
```

into this:

```js
if (!condition) {
  throw Error(
    __DEV__
      ? `A ${adj} message that contains ${noun}`
      : formatProdErrorMessage(ERR_CODE, adj, noun)
  );
}
```


packages/shared/formatProdErrorMessage.js
```diff
@@ -7,20 +7,19 @@
...
+function formatProdErrorMessage(code) {
   let url = 'https://reactjs.org/docs/error-decoder.html?invariant=' + code;
   for (let i = 1; i < arguments.length; i++) {
     url += '&args[]=' + encodeURIComponent(arguments[i]);
   }
-  error.message =
+  return (
     `Minified React error #${code}; visit ${url} for the full message or ` +
     'use the non-minified dev environment for full errors and additional ' +
-    'helpful warnings. ';
```
bvaughn: Wonder why we had a trailing space at the end before and if anything relied on it when appending to the error message. Doesn't look like it was intentional so I assume it's ok :)

acdlite: Yeah I think it was just a mistake :D

PR #17068, opened 2019-10-11
More accurate 'why did this render' wording

pelotom:
After using the new devtools profiler (which is awesome btw!) I discovered that one of my components was rerendering because "the parent component rendered", despite being `React.memo`ed and taking no props, which seemed like an impossibility (see [twitter thread](https://twitter.com/pelotom/status/1182482183769882626?s=20) documenting my confusion). I eventually discovered that it was in fact updating due to context changes, and @bvaughn [helpfully pointed out](https://twitter.com/brian_d_vaughn/status/1182654239627075584?s=20) that the message was a fallback used when the component rerendered due to reasons other than state and props changing.

This PR attempts to make the wording a little more accurate: in addition to adding the "or context changed" caveat, it seems like "Internal state changed" is a more accurate description than "Hooks changed", because updates due to the `useContext` hook don't fall in that category. Ideally in the special case where the component is memoized the message could definitively say the update was due to a context change, but I'm not sure how to detect that situation 🙂

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
@@ -167,7 +167,7 @@ function WhatChanged({
   if (changeDescription.didHooksChange) {
     changes.push(
       <div key="hooks" className={styles.WhatChangedItem}>
-        • Hooks changed
+        • Internal state changed
       </div>,
     );
   }
@@ -207,7 +207,7 @@ function WhatChanged({
   if (changes.length === 0) {
     changes.push(
       <div key="nothing" className={styles.WhatChangedItem}>
-        The parent component rendered.
+        The parent component rendered or context changed
       </div>,
     );
   }


packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
```diff
@@ -167,7 +167,7 @@ function WhatChanged({
   if (changeDescription.didHooksChange) {
     changes.push(
       <div key="hooks" className={styles.WhatChangedItem}>
-        • Hooks changed
+        • Internal state changed
```
bvaughn: Slight preference for "hooks" since people know what that means at least. "Internal state" sounds like it might refer to React internals or something else.

packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
```diff
@@ -207,7 +207,7 @@ function WhatChanged({
   if (changes.length === 0) {
     changes.push(
       <div key="nothing" className={styles.WhatChangedItem}>
-        The parent component rendered.
+        The parent component rendered or context changed
```
bvaughn: FWIW, we _can_ detect a context change for class components, so we probably only want to show this additional warning if we're inspecting a function component.

packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
pelotom: What about just "State changed"? What I found confusing was that changes due to `useContext` won't show up as "Hooks changed".

packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
pelotom: Sure, how do I detect that situation?

BTW, can't we also detect a context change for function components, simply by the process of elimination? If it's not due to state changes and it's memoized, then it must be due to context changes, right?

packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
bvaughn: Yeah, that's understandable. Most hooks aren't stateful in a way that triggers re-renders.

I'm not sure if "state changed" is better than "hooks changed" to be honest.

packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
bvaughn: > Sure, how do I detect that situation?

You can use the `type` attribute to differentiate between a class component (`ElementTypeClass`) and a function component (`ElementTypeFunction`, `ElementTypeMemo`, or `ElementTypeForwardRef`).

> BTW, can't we also detect a context change for function components, simply by the process of elimination? If it's not due to state or prop changes and it's memoized, then it must be due to context changes, right?

I think that may be a shaky assumption.

Currently the only two hooks that _directly_ trigger a re-render are `useReducer` and `useState` (which is built on top of `useReducer`). The rest of them are either _reactive_ or at least _indirect_ (like `useContext`). I'm not sure if this will always be the case though.

So I guess we _could_ show a more meaningful message if memoized props and state were shallowly equal _and_ element `type` was `ElementTypeMemo` but this feels fragile and wouldn't cover some more ambiguous cases. If we show a more specific message in some cases, but not in others- that alone could cause confusion. ("It's not a context change, because if it would DevTools would say so like it did before.")

PR #17070, opened 2019-10-11
Make uncontrolled -> controlled warning clearer

### LINKED ISSUE #17069: The warning for uncontrolled -> controlled inputs is confusing

Bug? Mild DX improvement.

Creating an input with an undefined or null value, then later passing a string value, triggers a warning about controlled inputs.

> A component is changing an uncontrolled input of type undefined to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa)

It's my experience in Reactiflux that this warning is unclear to developers with less familiarity with React, or at least with the "controlled" and "uncontrolled" terms.

vcarl:
Fixes #17069. Glad to iterate further on the specific language, but I've found that explaining in this way usually clears it up for people.


markerikson:
I would suggest adding a sentence like "This is likely caused by `value={something}` changing from `undefined` to a defined value".

sophiebits:
* The %s placeholder here is not the value type but rather the input type (checkbox, radio, etc). However I don’t think we need this info any more since we have robust component stacks in warnings. So let’s just remove that substitution and say “undefined value”.
* You’ll need to update any tests that are verifying this warning message.
* Other than that, this is a great improvement! Any chance you can make the above improvements? Then this can be merged.

vcarl:
Looks like this was upgraded to an error since I opened the PR. Fixing conflicts.

vcarl:
No rush on this, but anything else needed for it to land? cc @sophiebits @gaearon

gaearon:
@vcarl Sorry, there's a huge backlog and things slip through. I wanted to merge this but now the ref is gone. :-(

vcarl:
I restored the branch, but am unsubscribing from the PR. If it lands, cool! But I'm moving on.

gaearon:
Thanks!

PR #17071, opened 2019-10-11
Set up experimental builds

acdlite:
The experimental builds are packaged exactly like builds in the stable release channel: same file structure, entry points, and npm package names. The goal is to match what will eventually be released in stable as closely as possible, but with additional features turned on.

Versioning and Releasing
------------------------

The experimental builds will be published to the same registry and package names as the stable ones. However, they will be versioned using a separate scheme. Instead of semver versions, experimental releases will receive arbitrary version strings based on their content hashes. The motivation is to thwart attempts to use a version range to match against future experimental releases. The only way to install or depend on an experimental release is to refer to the specific version number.

Building
--------

I did not use the existing feature flag infra to configure the experimental builds. The reason is because feature flags are designed to configure a single package. They're not designed to generate multiple forks of the same package; for each set of feature flags, you must create a separate package configuration.

Instead, I've added a new build dimension called the **release channel**. By default, builds use the **stable** channel. There's also an **experimental** release channel. We have the option to add more in the future.

There are now two dimensions per artifact: build type (production, development, or profiling), and release channel (stable or experimental). These are separate dimensions because they are combinatorial: there are stable and experimental production builds, stable and experimental development builds, and so on.

You can add something to an experimental build by gating on `__EXPERIMENTAL__`, similar to how we use `__DEV__`. Anything inside these branches will be excluded from the stable builds. This gives us a low effort way to add experimental behavior in any package without setting up feature flags or configuring a new package.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,32 @@ aliases:
   - &attach_workspace
     at: build
 
+  - &process_artifacts
+    docker: *docker
+    environment: *environment
+    steps:
+      - checkout
+      - attach_workspace: *attach_workspace
+      - *restore_yarn_cache
+      - *run_yarn
+      - run: node ./scripts/rollup/consolidateBundleSizes.js
+      - run: ./scripts/circleci/upload_build.sh
+      - run: ./scripts/circleci/pack_and_store_artifact.sh
+      - store_artifacts:
+          path: ./node_modules.tgz
+      - store_artifacts:
+          path: ./build.tgz
+      - store_artifacts:
+          path: ./build/bundle-sizes.json
+      - store_artifacts:
+          # TODO: Update release script to use local file instead of pulling
+          # from artifacts.
+          path: ./scripts/error-codes/codes.json
+      - persist_to_workspace:
+          root: build
+          paths:
+            - bundle-sizes.json
+
 jobs:
   setup:
     docker: *docker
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -74,6 +100,18 @@ jobs:
       - *run_yarn
       - run: yarn test --maxWorkers=2
 
+  test_source_experimental:
```
bvaughn: I assume that since we're using parallel containers, this won't actually increase the duration of every CI job? That would not be great.

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -682,9 +671,8 @@ function warnIfReactDOMContainerInDEV(container) {
   if (__DEV__) {
     warningWithoutStack(
       !container._reactRootContainer,
-      'You are calling ReactDOM.%s() on a container that was previously ' +
+      'You are calling ReactDOM.createRoot() on a container that was previously ' +
```
bvaughn: I would not expect all fo these conditional names to change, given that we run this test against not-experimental builds as well. Why did they?

acdlite: See the first commit.

> The method names don't get stripped out of the production bundles
because they are passed as arguments to the error decoder.

Slightly unrelated but I did it this way instead of checking `__EXPERIMENTAL__` in the tests. We're probably going to remove the prefixed versions from the stable releases, anyway.

.circleci/config.yml
acdlite: The workflows run in parallel, yeah

acdlite:
I'll do release script changes in a follow-up. The existing `publish` script should actually work if you pass the build number in manually. The main thing I want to add is a check to the `prepare-stable` script that prevents you from publishing an experimental build as stable.

Issue #16968, opened 2019-10-01
React DevTools force re-render button


kentcdodds:
I'd like to have a button that forces a re-render of the selected component when clicked. For example, adding a button like this on the right of the component controls:

![image](https://user-images.githubusercontent.com/1500684/65933573-35b06980-e3cf-11e9-8245-e5bedac5c552.png)

This can be accomplished today by adding a new prop and changing the prop, but it's more work than I want to do.

This would be useful for use while profiling how a component performs with unnecessary re-renders. Right now I just have a button that's tied to a "forceRender" function:

```jsx
function useForceRerender() {
  const [, set] = React.useState()
  return React.useCallback(() => set({}), [])
}

function FilterComponent() {
  const forceRerender = useForceRerender()
 
  return (
    <>
      <button onClick={forceRerender}>force rerender</button>
      {/* more JSX */}
    </>
  )
}
```

Would be cool to have this built-in :)

hristo-kanchev:
Something like this maybe?

![forceupdate](https://user-images.githubusercontent.com/23095052/65963573-5aeab980-e45b-11e9-9a05-282bf6d6b832.gif)

Do you think the UI is getting cluttered over there, @bvaughn ?

If you are cool with this I can submit the PR.

kentcdodds:
That would be perfect for me 👍

bvaughn:
> Do you think the UI is getting cluttered over there, @bvaughn ?

Yup, that was my first thought when I saw the buttons take up slightly more space than the component name in your screenshot. Not sure what can be done about it that wouldn't add more clicks though (or make the hit area too small to be user friendly).

Let's not worry about the button space for now. I can revisit that UI.

> If you are cool with this I can submit the PR.

Sure, let's see a PR.

I wonder if "_Re-render component_" might be more intuitive to new users than "_Force update_"?

hristo-kanchev:
Yeah, "Re-render component" is better. I'll change the tooltip text 👍

hristo-kanchev:
Hey @bvaughn! Hope you are doing well.

I'm struggling to find a good way to re-render the component that will capture all use cases. 

I used `instance.forceUpdate` but this doesn't capture all use cases - for example functional components. 

I can push my code if you want to check it out and advise if possible?

bvaughn:
You should push it if you'd like someone to look it over. I don't know if I'll have any time in the near future though. Going to be pretty busy preparing for React conf.

bvaughn:
[...]

PR #17072, opened 2019-10-12
[DevTools] Implemented re-render mechanism for React 16.9+

hristo-kanchev:
Solves: #16968 

**Description**:
Added force re-render mechanism for components.

**Support**:
As @bvaughn suggested, we will **only support React 16.9+**
Versions that aren't supported won't show the `re-render` button in the UI.

**Tested on**:
- Versions that are < 16.9 do not show the `re-render` button in the UI.
- Versions that are >= 16.9 show the `re-render` button in the UI and successfully dispatch an update to the supported components types.

re-render


packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -2517,6 +2526,16 @@ export function attach(
     }
   }
 
+  function forceRerender(id: number) {
+    const fiber = findCurrentFiberUsingSlowPathById(id);
+
+    if (fiber !== null) {
+      if (typeof scheduleUpdate === 'function') {
```
trueadm: Can't this line be merged with the line above?

hristo-kanchev: Yup! I've pushed the changes.

packages/react-devtools-shared/src/backend/agent.js
```diff
@@ -125,6 +125,7 @@ export default class Agent extends EventEmitter<{|
     bridge.addListener('getOwnersList', this.getOwnersList);
     bridge.addListener('inspectElement', this.inspectElement);
     bridge.addListener('logElementToConsole', this.logElementToConsole);
+    bridge.addListener('forceRerender', this.forceRerender);
```
bvaughn: super nit: Maybe just "forceRender" ? The "re" seems unnecessary.

bvaughn:
I think we could improve the UX a bit though. When you click the button, there's no confirmation or signal that it's actually been clicked. Nothing happens.

Maybe we need to make it stay highlighted for a second :think: or something. Thoughts?

bvaughn:
I guess I still don't quite understand the use case of this feature, so I'd like to hold off merging it for the time being until I understand it better, and we work out the UX issue.

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -2247,6 +2253,9 @@ export function attach(
       // Can view component source location.
       canViewSource,
 
+      // Can force the component to re-render.
+      canForceRerender,
```
bvaughn: I think this should be conditional based on the element type. Right now it shows up for all types, including e.g. context providers and consumers, etc- re-rendering these doesn't make sense. Seems like this should be for class, function, forward ref, and _maybe_ memo components types only?

hristo-kanchev:
Heya @bvaughn ! I'll address your comments and after that we can park it for the time being, untill we get more feature requests like this one? What do you think?

bvaughn:
Sounds good, and thanks for being so understanding.

The code approach here seems fine. I'm just not sure the feature itself makes total sense yet, until I better understand the use case :)

bvaughn:
We chatted on #16968 and agreed to close this for now.

Thanks a bunch for the PR though!

PR #17081, opened 2019-10-14
refactoring: warning revamp part 2

Jessidhia:
Part 2 of https://github.com/facebook/react/issues/16753

Updating the babel plugin to handle the new syntax was a bit tricky.

`invariant` was also originally not part of the scope, but because the plugin that validated arguments to `warning` also expected `invariant` to have the same interface, I had to choose between forking the plugin or also transforming `invariant`.

Some of the expressions may look funny; they were generated by this codemod:

<details>

```js
'use strict';

module.exports = function transformer(file, api) {
  const j = api.jscodeshift;

  return j(file.source)
    .find(j.CallExpression)
    .filter(path => {
      const callee = path.value.callee;
      if (callee.type !== 'Identifier') {
        return false;
      }
      const name = callee.name;
      const scope = path.scope.lookup(name);

      return (
        // "isGlobal" just means file scope, even in Modules
        (scope === null || scope.isGlobal) &&
        (name === 'warning' ||
          name === 'warningWithoutStack' ||
          name === 'lowPriorityWarning' ||
          name === 'lowPriorityWarningWithoutStack' ||
          name === 'invariant')
      );
    })
    .forEach(path => {
      const firstArgument = path.value.arguments[0];
      if (firstArgument.type === 'StringLiteral') {
        // already transformed?
        return;
      }
      if (firstArgument.type === 'Literal') {
        if (firstArgument.value) {
          if (path.parentPath.value.type === 'ExpressionStatement') {
            path.prune();
          } else {
            path.replace(j.identifier('undefined'));
          }
          return;
        }
        path.value.arguments.shift();
        return;
      }

      path.value.arguments.shift();
      if (path.parentPath.value.type === 'ExpressionStatement') {
        return path.parentPath.replace(
          j.ifStatement(
            j.unaryExpression('!', firstArgument),
            j.blockStatement([path.parentPath.node])
          )
        );
      } else {
        return path.replace(
          j.sequenceExpression([
            j.logicalExpression('||', firstArgument, path.node),
            j.identifier('undefined'),
          ])
        );
      }
    })
    .toSource();
};
```
</details>

### DIFF
--- a/packages/create-subscription/src/createSubscription.js
+++ b/packages/create-subscription/src/createSubscription.js
@@ -36,14 +36,13 @@ export function createSubscription<Property, Value>(
 }> {
   const {getCurrentValue, subscribe} = config;
 
-  warningWithoutStack(
-    typeof getCurrentValue === 'function',
-    'Subscription must specify a getCurrentValue function',
-  );
-  warningWithoutStack(
-    typeof subscribe === 'function',
-    'Subscription must specify a subscribe function',
-  );
+  if (!(typeof getCurrentValue === 'function')) {
+    warningWithoutStack('Subscription must specify a getCurrentValue function');
+  }
+
+  if (!(typeof subscribe === 'function')) {
+    warningWithoutStack('Subscription must specify a subscribe function');
+  }
 
   type Props = {
     children: (value: Value) => React$Element<any>,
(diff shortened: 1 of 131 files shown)


gaearon:
Seems like something's up in prod environment — prod tests are failing?

acdlite:
> invariant was also originally not part of the scope, but because the plugin that validated arguments to warning also expected invariant to have the same interface, I had to choose between forking the plugin or also transforming invariant.

Can you please fork `invariant` instead? We have separate plans to replace those with `throw Error` expressions in the source and I'd rather not update those callsites twice.

acdlite:
I personally hate the `wrap-warning-with-env-check` plugin. Can we make an eslint rule instead of relying on a babel plugin to do it for us? The end game is to get to writing `console.error` directly.

gaearon:
Thanks for starting this! @walaura picked up your script and we got to completion in https://github.com/facebook/react/pull/17568. ❤️

PR #17082, opened 2019-10-14
[SuspenseList] Fix bugs with dropped Promises

sebmarkbage:
This includes two similar looking bug fixes to SuspenseList.

We must transfer any pending promises from inner boundary to list. For non-hidden modes, this boundary should commit so this shouldn't be needed but the nested boundary can make a second pass which forces these to be recreated without resuspending. In this case, the outer list assumes
that it can collect the inner promises to still rerender if needed.

Otherwise, the Promise never retries anything.

We must also propagate suspense "context" change to nested SuspenseLists. This bug looks similar to the previous one but is not based on the lack of retry but that the retry only happens on the outer boundary but the inner doesn't get a retry ping since it didn't know about its own promise after the second pass. So it bails out.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -2121,6 +2121,20 @@ function updateDehydratedSuspenseComponent(
   }
 }
 
+function scheduleWorkOnFiber(
+  fiber: Fiber,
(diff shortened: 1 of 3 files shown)


sebmarkbage:
It’s tricky to do that in throwException because you have to know the context at every level and you don’t know the context of the shadowed levels

PR #17083, opened 2019-10-14
Fixed a bug with illegal invocation for Trusted Types

koto:
`trustedTypes.is*` functions need to be bound to the `trustedTypes` object when called. This PR fixes the bug.


gaearon:
Hmm something is failing the tests.

I wonder if there’s more background on this spec decision I could learn from? Just curious.

koto:
There's no reason, I filed https://bugs.chromium.org/p/chromium/issues/detail?id=1014408 to track this in Chromium, but any fix we provide will take quite some time to reach Chrome stable :/ 

I fixed the test.

gaearon:
Cool, thanks

PR #17088, opened 2019-10-14
Remove concurrent apis from stable

acdlite:
Those who want to try concurrent mode should use the experimental builds instead.

I also turned on the Suspense server renderer and selective hydration in the experimental build.


packages/react-dom/src/client/ReactDOM.js
```diff
@@ -643,7 +624,7 @@ type RootOptions = {
   },
 };
 
-function createRoot(
+export function createRoot(
```
sebmarkbage: We shouldn't do exports here, right? Why would you?

acdlite: It's so I can add a prefixed version to the Facebook build:

https://github.com/facebook/react/blob/09d07e2bc5271b70e6e48cccbc56ca0c5b93ce8b/packages/react-dom/src/client/ReactDOMFB.js#L36-L37

I forgot this module is the public interface... I think it happens to work because only the `.default` export is reexported in the Common JS bundle, but maybe a better approach is to move the `createRoot` and other implementation stuff into a separate module.

acdlite: Ok after talking offline, I've removed this. www will use the experimental builds.

PR #17094, opened 2019-10-15
[babel-plugin-react-jsx] Avoid duplicate "children" key in props object

trueadm:
This PR fixes a duplicate "children" key bug in the `TransformJSXToReactJSX` Babel transform. Specifically, when we have JSX that looks like this:

```jsx
<Component children={1}>2</Component>
```

Before this PR, the output would have been:

```jsx
React.jsx(Component, {\n  children: 1,\n  children: "2"\n})
```

In the above output, the `children` property is duplicated in the object literal expression. In JS strict mode, this can cause errors in certain browsers (notably IE) where an object cannot contain duplicate property keys.

This PR changed the logic, so when this pattern is encountered (explicitly with `children` only) we create a new object and put the `children` property in it and then use `Object.assign` (or whatever the native helper is) to ensure the correct object is formed as an end-result. So the above would output:

```jsx
React.jsx(Component, Object.assign({
  children: 1
}, {
  children: "2"
}));
```

### DIFF
--- a/packages/babel-plugin-react-jsx/__tests__/TransformJSXToReactJSX-test.js
+++ b/packages/babel-plugin-react-jsx/__tests__/TransformJSXToReactJSX-test.js
@@ -480,4 +480,10 @@ describe('transform react to jsx', () => {
       transform(`<Component y={2} {...x} />`, {useBuiltIns: false})
     ).toMatchSnapshot();
   });
+
+  it('should not contain duplicate children key in props object', () => {
+    expect(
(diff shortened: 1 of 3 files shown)


trueadm:
@sebmarkbage The current JSX transform does not fix this case either, the fix actually comes from the ES2015 Babel preset that has a plugin that handles object literals with duplicate keys.

gaearon:
Yeah I initially thought that old transform handled this but we later realized it didn’t either so it seemed ok

PR #17097, opened 2019-10-15
DevTools does not eagerly patch console for RN

bvaughn:
Previously, the DevTools backend eagerly patched the console (to append ["component stacks"](https://github.com/facebook/react/blob/master/packages/react-devtools/CHANGELOG.md#component-stacks) to warnings and errors) even if no DevTools frontend was running. I did this because I thought the components stacks would be very helpful, and a lot of RN developers don't even run the DevTools UI to otherwise get them. (Also there would be no other way to get them for errors logged during mount.)

Unfortunately this had the observable effect of changing the source code line shown for the errors/warnings in debugger tooling (where the actual call to the native `console.error` is made) from `YellowBox.js` (the RN override) to the DevTools `backend.js` script. Some people complained, so this PR removes that default behavior<sup>1</sup>.

This should not affect the browser/DOM extension. I have tested to confirm this in the browser. I tested the change in RN by building the `react-native-core` backend and patching my local version of React Native.

Resolves https://github.com/facebook/react-native/issues/26788

<sup>1</sup> I'm not sure I understand why the previous behavior (of showing the `YellowBox.js` override location) was preferable to showing the DevTools backend. Neither is the location of the original caller. I've asked for clarification on the RN issue. If we decide to remove the DevTools override though, this is the way to do it.

### DIFF
--- a/packages/react-devtools-shared/src/hook.js
+++ b/packages/react-devtools-shared/src/hook.js
@@ -7,11 +7,6 @@
  * @flow
  */
 
-import {
-  patch as patchConsole,
-  registerRenderer as registerRendererWithConsole,
-} from './backend/console';
-
 import type {DevToolsHook} from 'react-devtools-shared/src/backend/types';
 
 declare var window: any;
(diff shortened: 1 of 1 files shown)


bvaughn:
I'm not sure I understand why the previous behavior (of showing the `YellowBox.js` override location) was preferable to the Dev have asked for clarification on the issue. If we decide to remove the DevTools override though, this is the way to do it.

gaearon:
>I'm not sure I understand why the previous behavior (of showing the YellowBox.js override location) was preferable to showing the DevTools backend

Huh, that's not at all what I thought the behavior was! What do you mean by this?

I was looking at this screenshot:



It has real filenames for all logs except the first one, not the YellowBox override location. Maybe you were looking at different lines?

gaearon:
To clarify, the relevant part is the red rectangle on the blurred screenshot.

bvaughn:
Ahh, I see. I misunderstood this to be about the component stack logs, but it seems this is actually about the fact that DevTools patches `console.log` too so it can disable logging when re-rendering to inspect hooks.

Hmm...maybe we should separate those.

gaearon:
Ohh I didn’t realize that’s what it is. Yeah. I think it’s OK to keep hijacking warn/error but probably not log.

gaearon:
Silly question. Can’t we only override console log *when* we re evaluate Hooks? And restore it right after.

bvaughn:
Yeah, maybe that override could be deferred until the first time a component was actually inspected (which would never happen unless someone was actually using the DevTools frontend).

So then the open question would be if we wanted to still keep eagerly overriding `console.warn`, `console.error` for component stacks (given that `YellowBox` already overrides those methods)

bvaughn:
Duh. I'm having trouble remembering my own code. We already _do_ delay patching `console.log` until we're re-rendering:
https://github.com/facebook/react/blob/4cb399a433771c84e861d5ca3d38a24733d23ad8/packages/react-devtools-shared/src/backend/renderer.js#L2291-L2313

gaearon:
https://github.com/facebook/react-native/issues/26788#issuecomment-542398749

PR #17099, opened 2019-10-15
Warn for update on different component in render

acdlite:
This warning already exists for class components, but not for functions.

It does not apply to render phase updates to the same component, which have special semantics that we do support.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -380,7 +380,7 @@ export function scheduleUpdateOnFiber(
   expirationTime: ExpirationTime,
 ) {
   checkForNestedUpdates();
-  warnAboutInvalidUpdatesOnClassComponentsInDEV(fiber);
+  warnAboutRenderPhaseUpdatesInDEV(fiber);
 
   const root = markUpdateTimeFromFiberToRoot(fiber, expirationTime);
   if (root === null) {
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2663,32 +2663,45 @@ if (__DEV__ && replayFailedUnitOfWorkWithInvokeGuardedCallback) {
...
+          case 'render':
+            if (didWarnAboutUpdateInRender) {
+              return;
+            }
+            warningWithoutStack(
+              false,
+              'Cannot update during an existing state transition (such as ' +
+                'within `render`). Render methods should be a pure function of ' +
+                'props and state.',
+            );
```
Jessidhia: Why use `warningWithoutStack` instead of `warning` here?

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -491,6 +491,50 @@ describe('ReactHooksWithNoopRenderer', () => {
       ]);
       expect(ReactNoop.getChildren()).toEqual([span(22)]);
     });
+
+    it('warns about render phase update on a different component', async () => {
```
Jessidhia: I presume there's already a test for it _not_ warning if it's the same component?

acdlite:
@sebmarkbage @gaearon Do we want to land this warning? If so I'll update it

gaearon:
@ermanzohre Can I ask you to please not “approve” old pull requests? Your actions are creating notification noise for everyone watching the repo. Thanks.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2781,30 +2781,44 @@ if (__DEV__ && replayFailedUnitOfWorkWithInvokeGuardedCallback) {
...
-      switch (ReactCurrentDebugFiberPhaseInDEV) {
-        case 'getChildContext':
-          if (didWarnAboutUpdateInGetChildContext) {
-            return;
-          }
+    if ((executionContext & RenderContext) !== NoContext) {
+      switch (fiber.tag) {
+        case FunctionComponent:
+        case ForwardRef:
+        case SimpleMemoComponent: {
```
sebmarkbage: I guess this should include Block too?

PR #17100, opened 2019-10-15
Separate sizebot for experimental builds

acdlite:
Configures the sizebot to leave a second comment that tracks the experimental build artifacts.


sebmarkbage:
Alternatively do a fan-in work flow thingy.

PR #17113, opened 2019-10-16
[react-interactions] Allow event.preventDefault on LegacyPress responder

trueadm:
This PR brings back `event.preventDefault` behavior that and removes the previous warning for its usage. Whilst this might not be the right long term strategy, this change unblocks internal cases where the native events need to be prevented to handle edge-cases that are exposed because of the mixing of event systems and quirks with the systems in place. Furthermore, this is intentionally only for the LegacyPress responder.

### DIFF
--- a/packages/react-interactions/events/src/dom/PressLegacy.js
+++ b/packages/react-interactions/events/src/dom/PressLegacy.js
@@ -157,9 +157,10 @@ function createPressEvent(
   let ctrlKey = false;
   let metaKey = false;
   let shiftKey = false;
+  let nativeEvent;
 
   if (event) {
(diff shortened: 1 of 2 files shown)


necolas:
> this change unblocks internal cases where the native events need to be prevented to handle edge-cases

What are those? And why did previous edge-cases not require this?

trueadm:
@necolas See internal task 55775588. I've also had another team come to me and want similar, to call `preventDefault` to stop text selection from changing when clicking an element that gains focus on `click` (like `button` elements).

necolas:
LegacyPress is likely too buggy for continued internal use and we should move them over to the Press module. So we might need to add this to other responders in that case

PR #17128, opened 2019-10-17
Add a failing test for SuspenseList bug

gaearon:
I don't know if it's a bug but it seems like it is?

If we render `AB` on mount and `CD` on update, I'd expect to see `CD`. But I see `ABCD`.


gaearon:
It seems like things being deleted in the first pass aren't still deleted in the second pass because we don't discover their deletion. Due to clearing the list's effect pointers.

PR #17144, opened 2019-10-18
Build script should default to experimental

acdlite:
`yarn build` defaults to building in experimental mode. To opt-out, set the `RELEASE_CHANNEL` environment variable to `stable`. This is the same as what we do when running tests via `yarn test`, to make local development easier.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,9 +149,13 @@ jobs:
       - checkout
       - *restore_yarn_cache
       - *run_yarn
-      - run: ./scripts/circleci/add_build_info_json.sh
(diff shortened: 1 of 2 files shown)


sebmarkbage:
Does this affect the release script in anyway? E.g. does it make us release experimental builds by default?

acdlite:
No, the release script pulls from CI artifacts. It doesn’t build locally.

Additionally, stable releases must be promoted from a canary, and the `prepare-stable` script will exit if you try to promote an experimental one.

So you would have to intentionally circumvent the normal release process in order to publish mess it up.

PR #17158, opened 2019-10-21
Use a microtask to detect end of event

acdlite:
Updates of like priority that occur within the same browser event are guaranteed to be batched. This is an important guarantee for event emitter or subscription-like patterns where a single browser event can spawn updates on many different components, all of which need to commit consistently.

To do this, we have a heuristic of marking the first "event time" of an update inside an event. Subsequent updates in the same event will reuse the same event time.

At the end of the event, the event time should be cleared. Failing to clear the event time has the effect of "freezing" the timeline, causing weird expiration quirks.

The event time used to be cleared right before entering the render phase. This mostly works, since every update will eventually be followed by a render task. However, it has some flaws: if there's lots of other non-React main thread work in the mean time, subsequent events could get overbatched. The heuristic is also a bit fragile; it happens to be the case that every call to `requestCurrentTime` is accompanied by a render task, but that may not necessarily always be the case, creating a subtle refactor hazard.

Instead of relying on the next render task to reset the current event time, this change schedules a microtask.

I didn't use an actual microtask because React does not currently require a Promise polyfill. Instead, I'm using a Scheduler task at immediate priority, which has almost the same semantics. (There's even an open question of whether we should get rid of Scheduler's "immediate" priority in favor of microtasks, but since it currently exists in the browser proposal, too, we can figure that out later.)

See https://github.com/facebook/react/pull/17159 for context on what prompted this PR

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -294,12 +294,14 @@ export function requestCurrentTime() {
     return msToExpirationTime(now());
   }
   // We're not inside React, so we may be in the middle of a browser event.
-  if (currentEventTime !== NoWork) {
+  if (currentEventTime === NoWork) {
     // Use the same start time for all updates until we enter React again.
-    return currentEventTime;
+    currentEventTime = msToExpirationTime(now());
+    scheduleCallback(ImmediatePriority, () => {
+      currentEventTime = NoWork;
+      return null;
+    });
   }
-  // This is the first update since React yielded. Compute a new start time.
-  currentEventTime = msToExpirationTime(now());
   return currentEventTime;
 }
 
--- a/packages/react-reconciler/src/__tests__/ReactExpiration-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactExpiration-test.internal.js
@@ -72,7 +72,7 @@ describe('ReactExpiration', () => {
     // Schedule an update.
     ReactNoop.render(<Text text="A" />);
     // Advance the timer.
-    Scheduler.unstable_advanceTime(2000);
+    Scheduler.unstable_advanceCPUBoundTime(2000);
     // Partially flush the the first update, then interrupt it.
     expect(Scheduler).toFlushAndYieldThrough(['A [render]']);
     interrupt();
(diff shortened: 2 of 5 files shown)


sebmarkbage:
From what I gathered from @gaearon in his repo it was indefinitely stalled until returning back to the tab and clicking something. Yet nothing was running on the thread.

In your description you mention that everything is accompanied by render tasks. But if that’s the case, how do we explain his bug? I believe Brian had a similar one. Not sure if we fixed that one.

Seems like we need to truly understand that bug until we have a fix.

Not sure about this micro task thing. We need to cut down on overhead as part of doing something as simple as setting state and we end up with a lot of scheduling now and a lot or computing of expirations.

Is there a simpler model here?

acdlite:
@sebmarkbage 

> Is there a simpler model here?

Believe it or not, I asked myself that question before I opened the PR!

With that out of the way...

> But if that’s the case, how do we explain his bug? I believe Brian had a similar one. Not sure if we fixed that one.

It's because the DevTools hook calls `requestCurrentTime()` inside its `onCommit` hook, which `requestCurrentTime` doesn't recognize as coming from inside a React event because it happens after the `CommitContext` flag gets reset.

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberDevToolsHook.js#L61

So SiMplEr fixes would include:

- fixing just this one bad caller in the DevTools hook. 
- not resetting the execution context until after the DevTools hook is called.
- resetting `currentEventTime` at both the beginning and end of `commitRoot`, and at the end of every other React entry point.

I'm not satisfied with those solutions because getting this wrong is subtle and the consequences when it goes wrong are bad.

> We need to cut down on overhead as part of doing something as simple as setting state and we end up with a lot of scheduling now and a lot or computing of expirations.

I don't think it adds that much relative overhead because it's only for the first update per task. So in the Flux case, it's constant time overhead. I think the worst case is if you have many separate Scheduler or browser tasks that all schedule individual updates really quickly. I could lower the priority from Immediate down to Normal/Default.

acdlite:
I think we should also have separate callers for reading the current time without freezing it, which is only needed when computing an update, versus reading the current time for any other reason.

Perhaps:

- `requestCurrentTimeForUpdate`
- `getCurrentTime`

acdlite:
Bug that motivated this change was fixed in https://github.com/facebook/react/pull/17158. I'm still concerned about the refactor hazard but I'll close this for now.

gaearon:
You mean it was fixed in https://github.com/facebook/react/pull/17160

PR #17159, opened 2019-10-21
Failing test: DevTools hook freezes timeline

acdlite:
The DevTools hook calls `requestCurrentTime` after the commit phase has ended, which has the accidental consequence of freezing the start time for subsequent updates. If enough time goes by, the next update will instantly expire.

I'll open potential fixes in separate PRs:

1. Use microtask to reset current event time: #17158
2. Don't call `requestCurrentTime` after commit phase has ended: #17160


acdlite:
Merged as part of #17160

PR #17160, opened 2019-10-21
Read current time without marking event start time

acdlite:
`requestCurrentTime` is only meant to be used for updates, because subsequent calls within the same event will receive the same time. Messing this up has bad consequences.

I renamed it to `requestCurrentTimeForUpdate` and created a new function that returns the current time without the batching heuristic, called `getCurrentTime`.

Swapping `requestCurrentTime` for `getCurrentTime` in the DevTools hook fixes the regression test added in #17159.


gaearon:
Wow. Would never think it's DevTools causing this 😄

PR #17164, opened 2019-10-21
Add guard to handle modified React elements with non-string keys

### LINKED ISSUE #17134: React Devtools should produce a better error message when integers are present as keys on react elements
Current behavior: React Devtools throws "RangError: Invalid Array Length" when integers are used as keys on react elements. 

Example: 

https://codesandbox.io/s/interesting-violet-v5c5j

https://v5c5j.csb.app/

Using anything but strings as keys is as far as I understand not even correct usage, but it would be great if react devtools checked a little bit earlier and had a nicer error than "RangeError: Invalid Array Length." It takes a long time to figure out from this message that one somehow managed to use integers as keys and needs to correct it.

I've only tested with Chrome and the latest version of react devtools as a chrome extension.

bvaughn:
Resolves #17134

I'm not convinced we should land this PR. Seems like maybe [this should be an unsupported use case](https://github.com/facebook/react/issues/17134#issuecomment-544734845). Interested in hearing what others thing though!

### DIFF
--- a/packages/react-devtools-shared/src/__tests__/__snapshots__/store-test.js.snap
+++ b/packages/react-devtools-shared/src/__tests__/__snapshots__/store-test.js.snap
@@ -607,3 +607,8 @@ exports[`Store should properly handle a root with no visible nodes: 1: mount 1`]
 `;
 
 exports[`Store should properly handle a root with no visible nodes: 2: add host nodes 1`] = `[root]`;
+
+exports[`Store should properly serialize non-string key values: 1: mount 1`] = `
+[root]
+    <Child key="123">
+`;
--- a/packages/react-devtools-shared/src/__tests__/store-test.js
+++ b/packages/react-devtools-shared/src/__tests__/store-test.js
@@ -837,4 +837,15 @@ describe('Store', () => {
     act(() => ReactDOM.unmountComponentAtNode(containerB));
     expect(store.supportsProfiling).toBe(false);
   });
+
+  it('should properly serialize non-string key values', () => {
+    const Child = () => null;
+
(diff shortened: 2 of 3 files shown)


alexkolson:
Wow! Thanks for the quick turnaround time on this. You rock! Again, I'm just a react user and not a contributor but I personally don't see much harm in adjusting devtools to help developers debug potential issues better. I'm not quite sure if I understand why this is an unsupported use case for devtools either. For React itself? Probably! But my hope would be that devtools gets better at telling me I'm doing something unsupported rather than _**devtools itself crashing**_ and being unusable due my unsupported or weird usage. I get that this could become a dangerous road and that you have to draw the line somewhere. Somehow, some way, one will always be able to purposely make devtools crash or throw an error. However, this particular case seems like something devtools can catch farther up relatively easily.

The thing is, devtools was the only thing crashing with integer keys. Our actual react application worked just fine. Or at least, appeared to be working from a react developer's perspective, perhaps there was something internally awry.

Even if the only change was that devtools still crashed with an error but instead of an obscure RangeError one got something like "Devtools encountered an error: unexpected integer key on element BlahBlahBlah. Keys must be strings." or something. 

Then again, I don't have nearly the experience or context that you do and everything I just said could be total craziness! I just really like devtools and find it super useful during react development, and I think better errors / error handling would be awesome!

threepointone:
Stamping to unblock, but happy to discuss more. 

I believe that devtools shouldn't throw (only) internal errors if the React app is misbehaving. So either we land this fix, OR we detect that it's been modified in this manner and log a warning/error before failing like it currently does (and suggest an alternative, like using React.cloneElement?)

PR #17165, opened 2019-10-21
Rename createSyncRoot to createBlockingRoot

gaearon:
"Sync" is easy to confuse with the Legacy mode. It also might sound too positive.

While I'm renaming `createSyncRoot` to `createBlockingRoot`, I also did some find-and-replace in comments and internal naming. Let's standardize on:

* Legacy Mode: The old one.
* Blocking Mode: createBlockingRoot
* Concurrent Mode: createRoot

### DIFF
--- a/packages/react-debug-tools/src/__tests__/ReactDevToolsHooksIntegration-test.js
+++ b/packages/react-debug-tools/src/__tests__/ReactDevToolsHooksIntegration-test.js
@@ -180,7 +180,7 @@ describe('React hooks DevTools integration', () => {
     }
   });
 
-  it('should support overriding suspense in sync mode', () => {
+  it('should support overriding suspense in legacy mode', () => {
     if (__DEV__) {
       // Lock the first render
       setSuspenseHandler(() => true);
--- a/packages/react-devtools-shared/src/__tests__/__snapshots__/storeStressSync-test.js.snap
+++ b/packages/react-devtools-shared/src/__tests__/__snapshots__/storeStressSync-test.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`StoreStress (Sync Mode) should handle a stress test for Suspense (Sync Mode) 1`] = `
+exports[`StoreStress (Legacy Mode) should handle a stress test for Suspense (Legacy Mode) 1`] = `
 [root]
   ▾ <Root>
       <X>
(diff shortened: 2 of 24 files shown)


acdlite:
Can we consider passing an option to `createRoot` instead, like `{disableConcurrentRendering: true}`? I think the original reason we went with a separate API was because we didn't want to expose `createBatch` to the non-concurrent version. But `createBatch` no longer exists.

acdlite:
I'm just not convinced "blocking" is a thing people will intuitively understand and I'm worried the additional terminology is more confusing than it's worth.

gaearon:
Concurrent Mode with "disabled concurrent rendering" isn't *that* much easier to explain IMO though — and we need to refer to this concept somehow when discussing adoption strategy.

acdlite:
But like what is the difference between "blocking" and "concurrent"? Are they mutually exclusive? Can you have one without the other? Technically legacy is also blocking. Why not  `createNonConcurrentRoot`? I don't get what "blocking" buys us as an additional terminology.

acdlite:
To clarify, I have two separate objections:

- Do we need to separate methods, instead of using an option? Given that the end game is to get rid of "blocking" and move everyone to concurrent.
- Is it worth adding a new term "blocking" when most people won't know what that means without an extra explanation, at which point we might as well keep using "concurrent."

acdlite:
OTOH I don't think it's any worse than what we already have, so I don't mind if we merge this for now. I would prefer to bikeshed it more before the actual release, though.

sebmarkbage:
There are a number of considerations for an options object:

- Future additions to the api (such as adding createBatch back) that are mode specific.
- Tree shaking, or clearly defined separate modules (like `/legacy`), makes it easier to create separate builds for it in the future.
- It encourages dynamic usage in wrappers. In fact that’s exactly what we do with hydration at FB.

That said, these arguments mostly apply to hydration too and we already said we didn’t want a combinatorial explosion.

sebmarkbage:
I mostly suggested it because Sync looked bad in the column. I’m all for bikeshedding this further before release.

However we should decide the name of the column now “Legacy Mode”, “??? Mode”, “Concurrent Mode”.

acdlite:
What I like about "legacy" is that describes the target use case as opposed to the technical characteristics. Maybe we could do the same here. "Compatibility Mode"?

sebmarkbage:
Tests failing.

gaearon:
Oops. Pushed

PR #17177, opened 2019-10-24
Update .mailmap

terabaud:
Update authors due to name change.


threepointone:
Sorry if I don’t understand, but what does this change? It seems to be an addition, hence I’m curious. 

Additionally, do you think we should verify this list across the repo’s history, with a script or something

terabaud:
I am mtf trans and there is one problem: before my transition, I contributed to React, so that my deadname appears in the AUTHORS file of this repository (L___ Rosema). 

The AUTHORS file is auto generated from a script using the git history:
https://github.com/facebook/react/blob/master/scripts/authors

See also: [git shortlog](https://git-scm.com/docs/git-shortlog)

For my past (and future) contributions to React, I'd like my deadname from the AUTHROS file to be removed and replaced by my proper name and email adress.  

The .mailmap file is used for overrides.

aweary:
Hey @terabaud, we don't really run the script to update AUTHORS very frequently. Please feel free to submit a PR that manually removes your deadname from the file.

PR #17178, opened 2019-10-24
Add CSS shorthands for horizontal & vertical variants for padding and margin

zhirzh:
Add CSS shorthands for padding and margin that are used in _react-native_:
* `verticalMargin` and `horizontalMargin`
* `verticalPadding` and `horizontalPadding`


abrahamcuenca:
Thanks for working on this @zhirzh this would be useful.

gaearon:
This can't be right. This code doesn't exist in RN.

PR #17183, opened 2019-10-25
fix: do not invoke function-like values in useDeferredValue

Jessidhia:
When reading the code I noticed that, because the `State` hook is used internally in the implementation, trying to hold a function or class value as a `DeferredValue` would cause it to be invoked as a lazy initializer, or updater function.

This fixes it by using a closure. I considered using a `typeof` check, but I'm not sure if adding the extra branches is useful. The branch could potentially avoid creating a closure on the common case, though. Not sure which is the best trade-off.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -1132,14 +1132,14 @@ function mountDeferredValue<T>(
   value: T,
   config: TimeoutConfig | void | null,
 ): T {
-  const [prevValue, setValue] = mountState(value);
+  const [prevValue, setValue] = mountState(() => value);
   mountEffect(
     () => {
       Scheduler.unstable_next(() => {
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1132,14 +1132,14 @@ function mountDeferredValue<T>(
   value: T,
   config: TimeoutConfig | void | null,
 ): T {
-  const [prevValue, setValue] = mountState(value);
+  const [prevValue, setValue] = mountState(() => value);
```
acdlite: If you use `mountReducer` and `updateReducer` instead, then you can write your own version of `basicStateReducer` and avoid these extra closures

acdlite: (where by "your own version of `basicStaticReducer`" I mean "a hoisted identity function" :D)

PR #17197, opened 2019-10-26
Test with AppVeyor with Node 11 and 12

Saibamen:
Test with Node 11 and 12

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,8 @@ environment:
   JAVA_HOME: C:\Program Files\Java\jdk1.8.0
   matrix:
     - nodejs_version: 10
+    - nodejs_version: 11
+    - nodejs_version: 12
 
 # Finish on first failed build
 matrix:


hang-yoko-tran:
s

Saibamen:
> 
> 
> s

s?

threepointone:
Hi! What is the purpose and motivation for this change? What does it do, or why is it useful?

Saibamen:
We need to test it with recent NodeJS LTS version: https://nodejs.org/en/download/

Saibamen:
> It usually works

Usually ;)
So we should test with latest LTS version of NodeJS in CI

acdlite:
Can we replace AppVeyor with a CircleCI workflow instead? My understanding is that we only use AppVeyor because CircleCI used to not support Windows. But now it does: https://circleci.com/blog/windows-general-availability-announcement/

Saibamen:
ping

PR #17206, opened 2019-10-29
[SuspenseList] Don't treat the last row in hidden as deleted if already mounted

sebmarkbage:
Already mounted rows that resuspend may be considered as part of a tail if they're at the end. However, for purposes of the tail="..." option they don't get deleted. We deal with that in cutOffTailIfNeeded.

However, if they're also the first to suspend in the "hidden" case, we have a special case that deletes the actual rendered row. This needs to consider if that row was already mounted or things go wrong.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
@@ -2023,7 +2023,7 @@ describe('ReactSuspenseList', () => {
     );
   });
 
-  it('eventually resolves two nested forwards suspense list with a hidden tail', async () => {
+  it('eventually resolves two nested forwards suspense lists with a hidden tail', async () => {
     let B = createAsyncText('B');
 
     function Foo({showB}) {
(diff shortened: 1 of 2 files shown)


sebmarkbage:
For the purposes of using the "tail" option you probably don't really ever care about already mounted components being part of the tail and we could just consider all of them as part of the head.

However, when you don't use it an update causes a resuspend it might be a bit confusing that they're now in "together" mode (since the head is always in "together" mode).

It would simplify things a bit if we just dropped this for updates though.

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
```diff
@@ -2135,4 +2135,159 @@ describe('ReactSuspenseList', () => {
       </div>,
     );
   });
+
+  it('eventually resolves two nested forwards suspense list with a hidden tail', async () => {
```
bvaughn: nit: "forwards suspense list**s**"

sebmarkbage: oops. this is a copy paste of the test above.

PR #17211, opened 2019-10-29
Dispatch commands to both UIManagers from both renderers

TheSavior:
ReactNative's internal module for [`codegenNativeCommands`](https://github.com/facebook/react-native/blob/master/Libraries/Utilities/codegenNativeCommands.js#L13) needs to dispatch to Fabric or paper depending on the node it was passed.

This makes the internal Renderer's exported `dispatchCommand` method both support components created with either renderer. 

There was a test file `ReactFabricAndNative-test.internal.js` that tests components created with Fabric passed to the Paper renderer.


TheSavior:
Since it is accessing the fabricUIManager directly from the global we shouldn't have any of those problems. When I forgot to do that and instead defined it at the top of the file, the tests failed. So I think this should be safe.

PR #17214, opened 2019-10-29
[react-interactions] Disable event listener guarding for Flare

trueadm:
I stumbled across a big internal issue relating to handling accessibility related focus management controls.

Specifically, we want to be able to track events that fire `focus` or `blur` events from mutations that ReactDOM makes to the DOM. So for example, if an `<input>` is focused and React unmounts this component (thus detatching it from the DOM tree and triggering respective `focus` and `blur` events) we want to be properly notified of such events (the `body` should gain `focus` in the above case). The reason why React today does not notify us on these events, is that we set an `enabled` flag during the commit phase of mutating the DOM. [The code for this can be found here](https://github.com/facebook/react/blob/master/packages/react-dom/src/events/ReactDOMEventListener.js#L331-L333).

Without such functionality, it's not possible to properly use React to track such changes appropriately – resulting into having manual DOM event listeners bypass React.

### DIFF
--- a/packages/react-dom/src/events/ReactDOMEventListener.js
+++ b/packages/react-dom/src/events/ReactDOMEventListener.js
@@ -328,7 +328,10 @@ export function dispatchEvent(
   eventSystemFlags: EventSystemFlags,
   nativeEvent: AnyNativeEvent,
 ): void {
-  if (!_enabled) {
+  // For Flare, need "blur" and "focus" events to fire in React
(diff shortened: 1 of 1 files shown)


trueadm:
> I think we'll need to go back to the drawing board here. One possible approach could be to simulate a blur if the previous focus is no longer in the document when we restore the selection/focus.

Thanks for this investigation. That does seem like the best solution, given that only Chrome is doing this for now.

PR #17216, opened 2019-10-29
Support package.json generation at build time

acdlite:
This commit adds the ability to limit the inclusion of certain files to the experimental releases, such as `react-dom/unstable-flight-server.js`. More generally, it supports generating different package.jsons depending on the build environment (release channel).

The package.json files that live at `packages/<package_name>/package.json` serve multiple purposes. The release scripts use them for guessing the next stable version numbers; they serve as templates for the actual package.jsons that get published to npm (the release script rewrites the version numbers first); during local development, the package source is symlinked into node_modules, which means the package.json is used for module resolution by Jest, Flow, Rollup, node scripts, and so on.

So I would rather not fork the packages entirely like we do with other files (host config, feature flags, www-specific modules), because it could get confusing trying to keep the copies in sync.

There are a few different approaches we could use. One option is to generate the package.jsons in a post-install step, like we do with the Flow configs.

But since the immediate requirement is only to conditionally generate the package.jsons that are published to npm, another option is to generate those files during the build step. Everything else can use the existing package.json files that live in the source.

So it works like this:

- If a package includes a `package.js`, the build script will use that instead of `package.json`.
- `package.js` is a module that returns a JavaScript object of keys and values. The build script will serialize this to JSON and write it to the artifacts directory.
- By convention, `package.js` files will import the `package.json` file from source and modify it before re-exporting.
- If `package.js` does not exist for a module, the build script will use `package.json` directly.

### DIFF
--- a/packages/react-dom/package.js
+++ b/packages/react-dom/package.js
@@ -1,0 +1,21 @@
+'use strict';
+
+const packageInfo = require('./package.json');
+
+if (__EXPERIMENTAL__) {
+  // Add experimental files
+  packageInfo.files.push(
+    'unstable-fizz.js',
+    'unstable-fizz.browser.js',
+    'unstable-fizz.node.js',
+    'unstable-flight-server.js',
+    'unstable-flight-server.browser.js',
+    'unstable-flight-server.node.js',
+  );
+
+  packageInfo.browser['./unstable-fizz.js'] = './unstable-fizz.browser.js';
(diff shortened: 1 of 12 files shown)


acdlite:
Hrrm I guess I need Jest to use the `package.js` files, too, so that module resolution when testing source matches module resolution when testing the builds.

acdlite:
Closing because it's not a priority right now. Would be nice to revisit in the future, though.

PR #17225, opened 2019-10-30
[react-interactions] Add handleSimulateChildBlur upon DOM node removal

trueadm:
Follow up to https://github.com/facebook/react/pull/17214.

This PR adds functionality to React Flare, in that a simulated `blur` event is fired when a DOM node being removed when it was actively focused (via the internal event system for Flare). The reason for doing it this way, opposed to within `restoreSelection` where it's far too late to trigger a `blur` event (as the DOM node has already been detached from the tree, so it's impossible to propagate such events).

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -55,6 +55,7 @@ import {
   addRootEventTypesForResponderInstance,
   mountEventResponder,
   unmountEventResponder,
+  dispatchEventForResponderEventSystem,
 } from '../events/DOMEventResponderSystem';
 import {retryIfBlockedOn} from '../events/ReactDOMEventReplaying';
 
@@ -108,6 +109,10 @@ import {
   enableFlareAPI,
   enableFundamentalAPI,
 } from 'shared/ReactFeatureFlags';
+import {
+  RESPONDER_EVENT_SYSTEM,
+  IS_PASSIVE,
+} from 'legacy-events/EventSystemFlags';
 
 let SUPPRESS_HYDRATION_WARNING;
 if (__DEV__) {
@@ -455,12 +460,20 @@ function handleSimulateChildBlur(
     selectionInformation &&
     child === selectionInformation.focusedElem
   ) {
-    const simulatedEvent: any = document.createEvent('CustomEvent');
-    simulatedEvent.initCustomEvent('blur', false, true);
-    simulatedEvent.relatedTarget = null;
-    ReactBrowserEventEmitterSetEnabled(true);
-    child.dispatchEvent(simulatedEvent);
-    ReactBrowserEventEmitterSetEnabled(false);
+    const targetFiber = getClosestInstanceFromNode(child);
+    // Simlulate a blur event to the React Flare responder system.
+    dispatchEventForResponderEventSystem(
+      'blur',
+      targetFiber,
+      ({
+        relatedTarget: null,
+        target: child,
+        timeStamp: Date.now(),
+        type: 'blur',
+      }: any),
+      ((child: any): Document | Element),
+      RESPONDER_EVENT_SYSTEM | IS_PASSIVE,
+    );
   }
 }
 
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -447,10 +447,28 @@ export function insertInContainerBefore(
   }
 }
 
+function handleSimulateChildBlur(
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -447,10 +447,28 @@ export function insertInContainerBefore(
...
+  if (
+    enableFlareAPI &&
+    selectionInformation &&
+    child === selectionInformation.focusedElem
+  ) {
+    const simulatedEvent: any = document.createEvent('CustomEvent');
+    simulatedEvent.initCustomEvent('blur', false, true);
+    simulatedEvent.relatedTarget = null;
+    ReactBrowserEventEmitterSetEnabled(true);
+    child.dispatchEvent(simulatedEvent);
```
sebmarkbage: I'm not sure we should actually dispatch on the real DOM. The issue with that is that now other non-React event listeners pick these up. In Firefox/Safari they'd unexpectedly see events that weren't there before. In Chrome they'd start seeing *double*  events. That's the reason I didn't use native dispatching in the replaying system to keep it isolated to React.

I'm also not sure it's the best to dispatch the events in the middle of the mutation sequence since the tree is inconsistent at this point. I'm not 100% sure the event system is resilient to working correctly at this point since we haven't updated to the new tree yet.

The other downside of doing this in here, as opposed to in the restore phase, is that now we add some overhead to all removeChild calls. This used to be inlined but will not need to be outlined and run extra checks for every removed node always. If we do it in the restore phase, it's once per commit instead of once per removed node.

I get that you want this to pick up the path before the deletion happens. However, by definition this node is a React managed node, so you can get the Fiber from it even after it's deleted from the DOM and that gives us the dispatch 
path.

trueadm: So I originally did exactly what you described above: instead of dispatching a native event, I trigger the Flare event responder system with `blur`.

The issue is, by that point the target is no longer in the tree. We need the target to be in the tree to ensure we can properly propagate this event correctly where multiple responders can react to such event – unfortunately by the restore phase fiber tree of the detached element is already deatched too (null return field). This ultimately left me with moving back to this shitty approach.

trueadm: The alternative is that we make this another event other than `blur`, like `reactblur` or something to avoid breaking existing logic.

trueadm: I have another approach...

trueadm: I've addressed in a follow up commit.

packages/react-interactions/events/src/dom/__tests__/FocusWithin-test.internal.js
```diff
@@ -140,28 +141,38 @@ describe.each(table)('FocusWithin responder', hasPointerEvents => {
       expect(onFocusWithinChange).toHaveBeenCalledTimes(2);
       expect(onFocusWithinChange).toHaveBeenCalledWith(false);
     });
+
+    it('is called after a focused element is unmounted', () => {
+      innerRef.current.focus();
```
necolas:
```suggestion
      const target = createEventTarget(innerRef.current);
      target.focus();
```

packages/react-interactions/events/src/dom/__tests__/FocusWithin-test.internal.js
```diff
@@ -258,6 +269,18 @@ describe.each(table)('FocusWithin responder', hasPointerEvents => {
       expect(onFocusWithinVisibleChange).toHaveBeenCalledTimes(2);
       expect(onFocusWithinVisibleChange).toHaveBeenCalledWith(false);
     });
+
+    it('is called after a focused element is unmounted', () => {
+      const inner = innerRef.current;
+      const target = createEventTarget(inner);
+      target.keydown({key: 'Tab'});
+      inner.focus();
```
necolas:
```suggestion
      const target = createEventTarget(innerRef.current);
      target.keydown({key: 'Tab'});
      target.focus();
```

trueadm: The changes to use `target.focus()` fail the tests, any idea why?

necolas: Maybe because it is just dispatching an event and not updating the value of `document.activeElement`? Perhaps jsdom does that behind the scenes when you use programmatic `.focus()`

trueadm:
@sebmarkbage Please can you take another look at this PR when you get a chance? Thanks :)

PR #17228, opened 2019-10-30
Rename findHostInstance_deprecated to findHostInstance_DEPRECATED

TheSavior:
@yungsters wants this to be `_DEPRECATED` to be more obnoxious than the lower case version. It's also more consistent with our internal files that we deprecate. React also has methods prefixed with `UNSAFE_` so the all caps seem reasonable.


yungsters:
Thanks. 👍

PR #17233, opened 2019-10-31
fix: make serializable data  in react-devtools

### LINKED ISSUE #17207: Bug: react-devtools TypeError: Do not know how to serialize a BigInt

TypeError: Do not know how to serialize a BigInt which makes it so the react dev tools cannot inspect the component props.

Steps to reproduce

1. Set a component prop to some value of type `BigInt`.
2. Open Chrome DevTools, then React Components view, try inspecting the component
3. There will be an exception [here](https://github.com/facebook/react/blob/b438699d3620bff236282b049204e1221b3689e9/packages/react-devtools-extensions/src/contentScript.js#L28-L38) (with the message above)

You can repro this with the following https://codesandbox.io/s/mystifying-cache-jshv3

Note that you need to repo via Chrome DevTools so that the bridge is active, if you use the codesandbox built-in DevTools the behavior is different.

That bigints are handled in a similar way to how symbols are handled. So that they don't crash when they cross a boundary, i.e. postMessage.

Latest version, hasn't worked before. That I know. Tested in Chrome and Firefox, although the behavior isn't exactly the same in Firefox as in Chrome, it doesn't appear to work in Firefox either.

I would like to propose a fix for this but I cannot find where in the source this would go. I don't care for editing capabilities and such, I just don't want the dev tools to give up on my just because I have BigInts in my code.

nutboltu:
Fixes #17207 

It seems `chrome.runtime` [`postMessage`](https://developer.chrome.com/extensions/messaging#connect) is trying to serialize the object but failed due to BigInt data type.( See the [Use within JSON](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) section)

**[UPDATE]**
Thanks @leidegre  for pointing out that we already hydrated our data type. I added data type `bigint` there.

I tested this with all the possible data type as below

- string
- number
- boolean
- object
- array
- undefined
- null
- Symbol
- BigInt


leidegre:
This is already taking place here. We just need to extend this with support for the BigInt type. It actually has its own `typeof` name `typeof 0n === "bigint"`.

https://github.com/facebook/react/blob/master/packages/react-devtools-shared/src/hydration.js#L93-L142

It should work probably close to the way symbols are currently being handled. They have the exact same problem unless treated the way they are.

nutboltu:
Thanks @leidegre , I completely missed that part. I added `bigint` data type in both hydration and view utils. Please have a look.

leidegre:
LGTM

leidegre:
@bvaughn Any chance we could get this fixed? We have a very clear problem statement with a purposed fix and this is really hurting me as I use BigInts everywhere in my React app... It's crashing the dev tools all the time...

nutboltu:
@bvaughn added `BigInt` in those tests and also updated the snapshots.

PR #17234, opened 2019-10-31
[Flight] Add Client Infrastructure

sebmarkbage:
This adds some infra to the client build so that we can parse a stream into an object that can suspend.

See the fixture for example.


packages/react-dom/src/client/flight/ReactFlightDOMClient.js
```diff
@@ -7,28 +7,80 @@
...
+    let chunk = request.responseText;
+    processStringChunk(response, chunk, processedLength);
+    processedLength = chunk.length;
+  }
+  function load(e: ProgressEvent): void {
+    progress(e);
+    complete(response);
+  }
+  function error(e: ProgressEvent): void {
+    reportGlobalError(response, new TypeError('Network error'));
```
gaearon: Why TypeError?

gaearon: Should there be any protection around this being called more than once? E.g. if some failure events fire in a succession.

sebmarkbage: This is supposed to mirror the type of `fetch` rejecting. Not sure what type that is.

There will be protection against multiple in the internals once we actually have an implementation.

PR #17245, opened 2019-10-31
Add preload() to React.lazy

mjackson:
This allows lazy components to be initialized ahead of render time. My use case is for preloading lazy components in React Router ahead of when they are needed. We could do this in a number of scenarios including when we first determine a component is going to be part of the render tree or when anticipating it will be.

Just submitting this draft to get feedback from the core team. If it's something you'd like, I can go ahead and write a test.

### DIFF
--- a/packages/react/src/ReactLazy.js
+++ b/packages/react/src/ReactLazy.js
@@ -7,6 +7,7 @@
 
 import type {LazyComponent, Thenable} from 'shared/ReactLazyComponent';
 
+import {initializeLazyComponentType} from 'shared/ReactLazyComponent';
 import {REACT_LAZY_TYPE} from 'shared/ReactSymbols';
 import warning from 'shared/warning';
 
@@ -17,6 +18,10 @@ export function lazy<T, R>(ctor: () => Thenable<T, R>): LazyComponent<T> {
     // React uses these fields to store the result.
     _status: -1,
     _result: null,
+    // Public API for initializing a lazy component ahead of render time.
+    preload() {
+      initializeLazyComponentType(lazyType);
+    },
   };
 
   if (__DEV__) {


raibima:
can't it be done already in userland?
```tsx
const preloadPage = () => import('./Page');
const Page = React.lazy(preloadPage);

// on link hover
preloadPage();
```

mjackson:
> can't it be done already in userland?

Yes, it can. But it's a lot less ergonomic. The API we are going for is this:

```js
let UserProfile = React.lazy(() => import('./UserProfile.js'));

let route = {
  path: '/user/:id',
  element: <UserProfile />
};
```

What I'd like to do is be able to preload a `route.element` w/out requiring the user to give me the preload function.

mjackson:
Also @raibima, the way the `React.lazy` code is currently written you will still suspend on the initial render even if you call the preload function yourself ahead of time. This is because the lazy component's `_status` will still be `Unresolved`, so it won't actually resolve until the next render.

Calling `initializeLazyComponentType` is the only way to avoid suspense entirely.

raibima:
ah, that makes sense. Thanks for clarifying @mjackson 👍

selfcontained:
This would also be sooooper helpful for promoting practices like preloading a lazy component in response to a mouse-over of a button that you know will render the component if clicked.  Totally solvable in user-land, but feels so useful that it'd be really nice to have built-in.  Currently we rely on our own wrapper around React.lazy to accomplish this, something like:

```js
export const LazyComponentWithPreload = (importStatement) => {
  const LazyComponent = React.lazy(importStatement);
  LazyComponent.preload = importStatement;
  return LazyComponent;
};
```

Then in practice it looks something like:

```jsx
const MyComponent = ReactLazyPreload(() =>
  import( /* webpackChunkName: "name-of-webpack-chunk" */ 'file-to-import');
);

<button onMouseOver={MyComponent.preload} />
```

mjackson:
Hiya @sebmarkbage, we talked briefly about this. Just wondering if you had any feedback on it. I'm planning on doing something like this in router v6, but we'll probably label it experimental unless there's support for it in core.

PR #17251, opened 2019-11-01
[Flight] Move around the Server side a bit

sebmarkbage:
__Builds on top of #17234.__

Just some clean up before I add more stuff.


packages/react-dom/src/__tests__/ReactFlightDOM-test.js
```diff
@@ -0,0 +1,92 @@
...
+      return model;
+    }
+
+    let {writable, readable} = getTestStream();
+    ReactFlightDOMServer.pipeToNodeWritable(<App />, writable);
+    let result = ReactFlightDOMClient.readFromReadableStream(readable);
+    await waitForSuspense(() => {
+      expect(result.model).toEqual({
+        html: '<div><span>hello</span><span>world</span></div>',
+      });
```
gaearon: Mind asserting we actually eventually got here?

packages/react-dom/src/__tests__/ReactFlightDOMBrowser-test.js
```diff
@@ -48,14 +56,19 @@ describe('ReactFlightDOM', () => {
...
+      };
+      return model;
+    }
+
+    let stream = ReactFlightDOMServer.renderToReadableStream(<App />);
+    let result = ReactFlightDOMClient.readFromReadableStream(stream);
+    await waitForSuspense(() => {
+      expect(result.model).toEqual({
+        html: '<div><span>hello</span><span>world</span></div>',
+      });
```
gaearon: Same

packages/react-dom/src/__tests__/ReactFlightDOM-test.js
sebmarkbage: There’s going to be a hundred of these. I’m not going to assert in each one that the testing environment works.

Jest fails it the promise to the test never resolves.

gaearon: I mean the promise could resolve without the inner expect actually having run successfully. If there’s a bug in waitForSuspense.

PR #17252, opened 2019-11-01
[Scheduler] Delete old rAF implementation

acdlite:
We've been using the message loop version for a while. Time to delete.


adasq:
What is the reason of removing rAF?

PR #17253, opened 2019-11-01
Don't show empty (no work) commits in Profiler

### LINKED ISSUE #16980: React DevTools recording commit without any component re-render
I'm struggling to make an isolated example of this, but the app where I found this is pretty simple so hopefully it's not too challenging to track down.

So I was profiling https://the-react-bookshelf.netlify.com (locally) and got this when I clicked on the "login" button:

image

image

The fact that there was no profile data for a commit is interesting. Each commit should be associated to a state update somewhere in the tree, and wherever that happened should trigger at least one component to re-render, but that didn't appear to happen here.

I also verified that I don't have any components filtered out:

image

And I didn't filter any commits either:

image

Here's the exported profile data:

https://gist.github.com/kentcdodds/dbff66043653333cd22cb9261a08550b

And here's the repo where you can pull it down and reproduce yourself: https://github.com/kentcdodds/bookshelf. The component we're looking at is here: https://github.com/kentcdodds/bookshelf/blob/master/src/unauthenticated-app.js

Sorry I can't give a more direct reproduction.

bvaughn:
Improves DevTools experience for #16980

This PR works around broken behavior (arguably) with the synthetic event system. It updates the DevTools backend to ignore commits for which React bailed out at the root without doing any work.

My logic for detecting this bailout case is as follows:
```js
const didBailoutAtRoot =
  root.current === root.current.alternate ||
  (alternate !== null &&
    alternate.expirationTime === 0 &&
    alternate.childExpirationTime === 0);
```

I initially tried checking for bailout in the same way as `bailoutOnAlreadyFinishedWork`:
https://github.com/facebook/react/blob/a2e05b6c148b25590884e8911d4d4acfcb76a487/packages/react-reconciler/src/ReactFiberBeginWork.js#L2735-L2741

But this had false positives (e.g. mounting, certain types of root updates). Based on my own testing, the approach above seems like a more reliable check. (It doesn't break any of the automated tests, and it "fixes" both of the failing cases reported in #16980.)

We should probably follow up to this issue by fixing the event system to not schedule these unnecessary updates, and/or updating React to not commit empty changes, but at least this change improves the current confusing profiling experience.

### DIFF
--- a/packages/react-devtools-shared/src/__tests__/profilerStore-test.js
+++ b/packages/react-devtools-shared/src/__tests__/profilerStore-test.js
@@ -79,4 +79,44 @@ describe('ProfilerStore', () => {
     store.profilerStore.profilingData = fauxProfilingData;
     expect(store.profilerStore.profilingData).toBe(fauxProfilingData);
   });
+
+  // This test covers current broken behavior (arguably) with the synthetic event system.
+  it('should filter empty commits', () => {
+    const inputRef = React.createRef();
+    const ControlledInput = () => {
+      const [name, setName] = React.useState('foo');
+      const handleChange = event => setName(event.target.value);
+      return <input ref={inputRef} value={name} onChange={handleChange} />;
+    };
+
+    const container = document.createElement('div');
+
+    // This element has to be in the <body> for the event system to work.
+    document.body.appendChild(container);
+
+    // It's important that this test uses legacy sync mode.
+    // The root API does not trigger this particular failing case.
+    ReactDOM.render(<ControlledInput />, container);
+
+    utils.act(() => store.profilerStore.startProfiling());
+
+    // Sets a value in a way that React doesn't see,
+    // so that a subsequent "change" event will trigger the event handler.
+    const setUntrackedValue = Object.getOwnPropertyDescriptor(
+      HTMLInputElement.prototype,
+      'value',
+    ).set;
+
+    const target = inputRef.current;
+    setUntrackedValue.call(target, 'bar');
+    target.dispatchEvent(new Event('input', {bubbles: true, cancelable: true}));
+    expect(target.value).toBe('bar');
+
+    utils.act(() => store.profilerStore.stopProfiling());
+
+    // Only one commit should have been recorded (in response to the "change" event).
+    const root = store.roots[0];
+    const data = store.profilerStore.getDataForRoot(root);
+    expect(data.commitData).toHaveLength(1);
+  });
 });
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -1643,6 +1643,7 @@ export function attach(
         }
       }
     }
+
     if (shouldIncludeInTree) {
       const isProfilingSupported = nextFiber.hasOwnProperty('treeBaseDuration');
       if (isProfilingSupported) {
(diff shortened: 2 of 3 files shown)


acdlite:
I'm going to assume you've chosen to put this check into the DevTools backend so that it works with existing releases of React, and users don't need to upgrade React in order to get the nicer experience. Which makes sense.

Regarding your heuristic. Mostly works, but you don't want to only compare to `NoWork`. You also want to check if the pending work is lower priority.

Also `root.current === root.current.alternate` can be removed because it is always false. A fiber's alternate will never point to itself. It's either `null` or a different fiber (its alternate).

So the check should look like:

```js
// DevTools backend
// In the commit phase, the alternate is the "previous current" fiber. So we can
// check its pending work before we started rendering.
const alternate = finishedWork.alternate;
const commitIsEmpty =
  alternate !== null &&
  alternate.expirationTime < finishedExpirationTime &&
  alternate.childExpirationTime < finishedExpirationTime;
```

I think this will cover almost all cases, but not 100%. Maybe like 99%. As a heuristic it's good enough, though.

The cases it won't catch is when there's an update with greater or equal priority between when we started rendering (i.e. when the HostRoot fiber was cloned) and when we committed it, because when scheduling the update, the `expirationTime` fields will have been dirtied again.

But the reason I don't think this is an issue in practice is that, in concurrent mode, a higher priority update would have also interrupted the render. So we would never get to the commit phase.

I also think it's not a problem in legacy and blocking modes, because even though those modes don't have interruptions, they also block events, so it's not possible to get a `setState`, anyway (unless you call `setState` during the render phase, which we forbid with a warning).

---

Anyway, this is a long-winded way of saying: this is a good enough check by itself, and it's nice that DevTools users won't have to upgrade React to get the fix.

But we should still fix it in the work loop by preventing React from commiting empty renders in the first place.

My suggestion for that is, right before calling `prepareFreshStack`, you can check if the HostRoot work-in-progress has pending update or child work. If it doesn't, exit without entering the work loop at all.

```js
// Inside `performConcurrentWorkOnRoot`/`performSyncWorkOnRoot`
if (root !== workInProgressRoot || expirationTime !== renderExpirationTime) {
  // Before starting a fresh render, check if we can bailout on the entire tree.
  // For non-root fibers, this bailout usually happens in the begin phase
  // of the parent component, but roots are special because they don't
  // have parents.
  const remainingExpirationTimeOnRoot =
    root.current.expirationTime > root.current.childExpirationTime
      ? root.current.expirationTime
      : root.current.childExpirationTime;
  if (remainingExpirationTime < renderExpirationTime) {
    // We can bailout without entering the work loop.
    markRootFinishedAtTime(root, expirationTime, remainingExpirationTimeOnRoot);
    return null;
  } else {
    // Create a fresh stack and continue to the work loop like normal.
    prepareFreshStack();
  }
}
```

acdlite:
Also we should figure out what's causing the empty renders in the first place

bvaughn:
> I'm going to assume you've chosen to put this check into the DevTools backend so that it works with existing releases of React, and users don't need to upgrade React in order to get the nicer experience. Which makes sense.

Yes. Thanks for the added context/explanation.

It would be ideal to avoid having to check the finished expiration time because DevTools doesn't have that info (so it wouldn't help older versions). I'll remove the current === alternate check since that was a misunderstanding.

> Also we should figure out what's causing the empty renders in the first place

Agreed. I will open a follow up issue after this one has been landed to track down the event system problem (and include the two repro cases).

bvaughn:
> Also we should figure out what's causing the empty renders in the first place

Adding the early bailout check described above breaks lots of tests (34 test suites, 154 test cases). Looks like many of them using the noop renderer for testing incremental/interruptions.

acdlite:
Could you open a draft PR so I can take a look?

bvaughn:
Yeah, sorry. That's my intention. I'm on call this week so I'm doing the syncs first to get them out of the way.

bvaughn:
Yeah, sorry. That was my intention but I'm on call this week and got side tracked by a sync.

Pushed it here:
https://github.com/facebook/react/pull/17267

Will dig in as soon as the sync is unblocked.

bvaughn:
Ah I think this was my mistake. Maybe I should have been using the _next_ `expirationTime` rather than the (previously set) `renderExpirationTime`)

PR #17255, opened 2019-11-02
[eslint] Check forwardRef callbacks

### LINKED ISSUE #17220: [eslint-plugin-react-hooks] Apply the rules of hooks to a forwardRef-wrapped component
Feature (to catch potential bugs)

The react-hooks/rules-of-hooks ESLint rule catches uses of hooks in conditionals in components, but does not recognise an anonymous function wrapped in forwardRef as a component. 

The following example breaks the rules of hooks, but isn't caught by the rule:
```jsx
// This should fail
const FancyButton = React.forwardRef((props, ref) => {
  if (props.fancy) {
    useCustomHook();
  }
  return <button ref={ref}>{props.children}</button>;
});
```

The above example should be caught by react-hooks/rules-of-hooks, and raise the "React Hook "useCustomHook" is called conditionally" error.

dprgarner:
Closes #17220 

This checks that the rules of hooks apply to callback arguments of `forwardRef`. 

I'm unsure whether this counts as a backwards-incompatible change or not. It's making the linter stricter, but I'd only expect new errors to be reported in code that's already breaking the rules of hooks. There is a potential edge-case of code with a forwardRef-wrapped component using a non-hook function with a name starting with "use"; this change would now return a false positive.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
@@ -142,21 +142,27 @@ const tests = {
       }
     `,
     `
-      // Valid because hooks can be used in forwardRef.
-      const FancyButton = React.forwardRef(() => {
-        return useHook();
+      // Valid because hooks can be used in anonymous arrow-function arguments
+      // to forwardRef.
+      const FancyButton = React.forwardRef((props, ref) => {
+        useHook();
+        return <button {...props} ref={ref} />
       });
     `,
     `
-      // Valid because hooks can be used in forwardRef.
-      const FancyButton = React.forwardRef(function () {
-        return useHook();
+      // Valid because hooks can be used in anonymous function arguments to
+      // forwardRef.
+      const FancyButton = React.forwardRef(function (props, ref) {
+        useHook();
+        return <button {...props} ref={ref} />
       });
     `,
     `
-      // Valid because hooks can be used in forwardRef.
-      const FancyButton = forwardRef(function () {
-        return useHook();
+      // Valid because hooks can be used in anonymous function arguments to
+      // forwardRef.
+      const FancyButton = forwardRef(function (props, ref) {
+        useHook();
+        return <button {...props} ref={ref} />
       });
     `,
     `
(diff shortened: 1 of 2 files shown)


dprgarner:
Thanks for the feedback. I'll add explicit props and ref arguments to the test cases, and generalise this to check `React.memo` too.

The [react/display-name][react-display-name] rule checks that components have a display name. In the codebase where I spotted this issue, we set `displayName` as a property on the component returned by `forwardRef`, which then causes this rule to pass. I don't know if it makes sense for this linting rule to be changed to require the `forwardRef` argument to be a named CapitalCase function, as setting the `displayName` property directly fixes the issue that the linting rule is checking for.

Should there be a linting rule enforcing CapitalCase function names in the arguments of forwardRef, though? Is the argument actually a component? The function signature isn't the same as for a functional component, and the [forwardRef docs][forward-ref-display-name] refer to the argument as a "render function" and don't use CapitalCase when naming it. Does this mean that the function argument to `forwardRef` shouldn't be using hooks at all?

[react-display-name]: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md
[forward-ref-display-name]: https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools

packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
```diff
@@ -695,6 +761,45 @@ const tests = {
...
+        // This *must* be invalid.
+        const MemoizedButton = memo(function(props) {
+          if (props.fancy) {
+            useCustomHook();
+          }
+          return <button>{props.children}</button>;
+        });
+      `,
+      errors: [conditionalError('useCustomHook')],
+    },
```
gaearon: Can we also add tests verifying how we handle `React.whatever(props => ...)` or `whatever(props => ...)`?

dprgarner: I've added some more tests. The existing behaviour is to skip checking unnamed callbacks  - there's a [regression test](https://github.com/facebook/react/pull/17255/files#diff-a822ea22538a2256c588fc355abbc79fR293) in place for this.

PR #17265, opened 2019-11-04
Types: findHostInstance_DEPRECATED returns React.ElementRef<HostComponent<mixed>>

TheSavior:
This was returning the wrong type and couldn't be used correctly. @lunaleaps found this issue when trying to use it in TouchableNativeFeedback for the first time.

### DIFF
--- a/packages/react-native-renderer/src/ReactFabric.js
+++ b/packages/react-native-renderer/src/ReactFabric.js
@@ -45,7 +45,7 @@ const ReactCurrentOwner = ReactSharedInternals.ReactCurrentOwner;
 
 function findHostInstance_DEPRECATED(
   componentOrHandle: any,
-): ?HostComponent<mixed> {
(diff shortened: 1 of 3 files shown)


packages/react-native-renderer/src/ReactFabric.js
```diff
@@ -45,7 +45,7 @@ const ReactCurrentOwner = ReactSharedInternals.ReactCurrentOwner;
 
 function findHostInstance_DEPRECATED(
   componentOrHandle: any,
-): ?HostComponent<mixed> {
+): ?React$ElementRef<HostComponent<mixed>> {
```
TheSavior: I used dollar signs here because these ones are inside the renderer and it doesn't seem to make sense to `import 'react'` at the top of this file. This seems to be consistent with the rest of the files in the React repo.

PR #17266, opened 2019-11-04
[react-interactions] Prevent duplicate onPress firing for keyboard Enter

trueadm:
Internally, we encountered an issue where the `LegacyPress` responder would double fire `onPress` when the `Enter` keyboard key was pressed. Looking into this deeply, the reason for this was because the browser triggers a `click` event which passes the logic in `isScreenReaderVirtualClick`.

I tried to find different discrepancies between screen reader clicks and the click that is fired from the browser when you press `Enter` but I couldn't see anything that stood out – so I took another approach and realized, by accident, the reason this doesn't happen when you press `Space` or ` ` key – we trigger `nativeEvent.preventDefault`. This means that the corresponding native `click` never gets triggered by the browser, but the responder still correctly fires the `onPress` callback – this time only once though!

I also noticed that is behavior didn't occur in the newer Press responder. I added support for the `Spacebar` key type there, but I also noticed a difference in `input` and `textarea` logic between the two responders that might be worth while tracking in a separate task – but maybe this is something @necolas has already considered.

### DIFF
--- a/packages/react-interactions/events/src/dom/Press.js
+++ b/packages/react-interactions/events/src/dom/Press.js
@@ -97,7 +97,7 @@ function isValidKey(e): boolean {
   const {key, target} = e;
   const {tagName, isContentEditable} = (target: any);
   return (
-    (key === 'Enter' || key === ' ') &&
+    (key === 'Enter' || key === ' ' || key === 'Spacebar') &&
     (tagName !== 'INPUT' &&
       tagName !== 'TEXTAREA' &&
       isContentEditable !== true)
--- a/packages/react-interactions/events/src/dom/Press.js
+++ b/packages/react-interactions/events/src/dom/Press.js
@@ -106,7 +106,10 @@ function isValidKey(e): boolean {
 
 function handlePreventDefault(preventDefault: ?boolean, e: any): void {
   const key = e.key;
-  if (preventDefault !== false && (key === ' ' || key === 'Enter')) {
+  if (
+    preventDefault !== false &&
+    (key === ' ' || key === 'Enter' || key === 'Spacebar')
+  ) {
     e.preventDefault();
   }
 }
--- a/packages/react-interactions/events/src/dom/PressLegacy.js
+++ b/packages/react-interactions/events/src/dom/PressLegacy.js
@@ -570,15 +570,14 @@ const pressResponderImpl = {
                 metaKey,
                 shiftKey,
               } = (nativeEvent: MouseEvent);
-              if (isValidKeyboardEvent(nativeEvent)) {
-                nativeEvent.preventDefault();
-              } else if (
+              if (
                 props.preventDefault !== false &&
                 !shiftKey &&
                 !metaKey &&
                 !ctrlKey &&
                 !altKey
               ) {
+                nativeEvent.preventDefault();
                 state.shouldPreventClick = true;
               }
             } else {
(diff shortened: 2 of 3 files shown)


packages/react-interactions/events/src/dom/PressLegacy.js
```diff
@@ -570,7 +570,7 @@ const pressResponderImpl = {
                 metaKey,
                 shiftKey,
               } = (nativeEvent: MouseEvent);
-              if (nativeEvent.key === ' ') {
+              if (isValidKeyboardEvent(nativeEvent)) {
```
necolas: This condition is already met on L566. This changes means the means that the `else if` clause on L575 will never be reached

packages/react-interactions/events/src/dom/__tests__/PressLegacy-test.internal.js
```diff
@@ -914,8 +914,7 @@ describe.each(environmentTable)('Press responder', hasPointerEvents => {
       ReactDOM.render(<Component />, container);
 
       const target = createEventTarget(ref.current);
-      target.keydown({key: 'Enter'});
-      target.click({preventDefault});
```
necolas: Why was the `click` event removed from this test? When Enter is pressed is certain host elements it will cause a `click` to be dispatched

packages/react-interactions/events/src/dom/Press.js
```diff
@@ -106,7 +106,10 @@ function isValidKey(e): boolean {
 
 function handlePreventDefault(preventDefault: ?boolean, e: any): void {
   const key = e.key;
-  if (preventDefault !== false && (key === ' ' || key === 'Enter')) {
+  if (
+    preventDefault !== false &&
+    (key === ' ' || key === 'Enter' || key === 'Spacebar')
```
necolas: Why is `Spacebar` added here? For some legacy browser? It would need to be added to L100 too.

packages/react-interactions/events/src/dom/__tests__/PressLegacy-test.internal.js
trueadm: Not if you `preventDefault`, meaning `click` will never fire.

packages/react-interactions/events/src/dom/PressLegacy.js
trueadm: Yeah, let me revise this somewhat.

packages/react-interactions/events/src/dom/Press.js
trueadm: For legacy browser support. I've added to the other code-site as per your suggestion.

trueadm:
@necolas You seem to know quite a bit about this logic. If you are happy and have time, please could you do some digging into it to help me? I'd really appreciate it :)

packages/react-interactions/events/src/dom/__tests__/PressLegacy-test.internal.js
necolas: The reason we preventDefault on Spacebar it because otherwise it scrolls the window. Should we preventDefault on keydown for Enter though?

trueadm: I think so, unless you know of any reason why we should not?

PR #17270, opened 2019-11-04
Remove `debugRenderPhaseSideEffects` flag

acdlite:
There are two similar flags, `debugRenderPhaseSideEffects` and `debugRenderPhaseSideEffectsForStrictMode`. The strict mode one is the only one that is actually used. I think originally the theory is that we would one day turn it on for all components, even outside strict mode. But what we'll do instead is migrate everyone to strict mode.

The only place `debugRenderPhaseSideEffects` was being used was in an internal test file. I rewrote those tests to use public APIs.


packages/react-reconciler/src/ReactFiberSuspenseConfig.js
```diff
@@ -11,11 +11,11 @@ import ReactSharedInternals from 'shared/ReactSharedInternals';
 
 const {ReactCurrentBatchConfig} = ReactSharedInternals;
 
-export type SuspenseConfig = {|
+export type SuspenseConfig = {
   timeoutMs: number,
   busyDelayMs?: number,
   busyMinDurationMs?: number,
-|};
+};
```
bvaughn: Seems unrelated but I assume this change was intentional?

acdlite: Ah nope didn't mean to stage that

TheSavior:
Should this callsite have been update too? Or is it different? https://github.com/facebook/react/blob/master/scripts/rollup/shims/react-native/ReactFeatureFlags.js

It looks like there are still three callsites of `debugRenderPhaseSideEffects` in the react repo

PR #17274, opened 2019-11-05
Fix devtools displaying Anonymous for memo of ref-forwarding components

### LINKED ISSUE #16722: memo name is Anonymous when passed the result of forwardRef

Bug

Consider the following code:

```js
const forwardRefComponentLike = function forwardRefComponentLike(_props, _ref) {
  return <span>FowardRef</span>;
};

const FowardRefComponent = React.forwardRef(forwardRefComponentLike, {});

const MemoOfFowardRefComponent = React.memo(FowardRefComponent);
```

The `MemoOfFowardRefComponent` component does not pick up the name of the ForwardRefComponent and is instead rendered in devtools as:

```
 Anonymous [Memo]
    forwardRefComponentLike [ForwardRef]
```

Screen Shot 2019-09-09 at 10 45 02 PM

See this codesandbox for a demo and a comparison of how React.memo handles regular functional components: https://codesandbox.io/s/react-forwardref-and-memo-combo-33vx7

I would expect this to pick up the name of the thing passed into React.Memo, as happens for regular components

```
forwardRefComponentLike [Memo]
```

Tested with React 16.9.0.

wsmd:
Resolves #16722 

This PR addresses the following issues:
- `react-devtools-shared` not displaying the proper displayName for `memo(forwardRef(C))`
- ~`react-is`'s `typeOf` not accounting for `forwardRef` components/functions. It's currently accounting for `forwardRef` elements only.~ (Fixed in #17278 🎉)

#### Problem: devtools not displaying the proper displayName for `memo(forwardRef(C))`

I encountered an issue with React dev-tools where it no longer infers the proper display name of a component wrapped with both `memo` and `forwardRef`. The display name in this case is shown as "Anonymous".

[I'm not the only one experiencing this!](#16722)

Consider the following example:

```jsx
const HelloWorld = () => <div>Hello, World!</div>;

const RefForwarding = React.forwardRef(HelloWorld);
const MemoizedRefForwarding = React.memo(RefForwarding);
```

When I render `<MemoizedRefForwarding />`, the React devtools shows the following:

Screen Shot 2019-11-04 at 7 47 01 PM

Notice how the `Memo` component is displayed as "Anonymous"

What I would expect is the following (taken after using the attempted fix):

Screen Shot 2019-11-04 at 7 46 22 PM

The `Memo` component is now properly displayed as "MemoizedRefForwarding"

#### Solution

I traced down the problem and found that `react-devtools-shared` has a [shared helper function `getDisplayName`](https://github.com/facebook/react/blob/0f3838a01b0fda0ac5fd054c6be13166697a113c/packages/react-devtools-shared/src/utils.js#L40-L62) that expects a function. This `getDisplayName` function is being [used by `getDisplayNameForFiber`](https://github.com/facebook/react/blob/0f3838a01b0fda0ac5fd054c6be13166697a113c/packages/react-devtools-shared/src/backend/renderer.js#L330) in `backend/renderer.js`.

While this works for most fibers with their `type` as the component function itself, `forwardRef` and `memo` have a slightly different `type`.

In the case of `memo(forwardRef(C))`, the `type` of the memo element is the `forwardRef` object, not the component function itself (which `getDisplayName` expects), so we need to extract that type off of `forwardRef`.

I solved this problem by adding a `resolveFiberType` that will attempt to "resolve" the type value, if the type of the fiber node happens to be an object, before passing it to `getDisplayName`. Since this is doing recursion it will cover `memo(C)`, `forwardRef(C)` , and `memo(forwardRef(C))`.

I'm not quite sure if that's the best possible way to approach this problem since it's my first time contributing to the React codebase 😬 so I'm not super familiar with it yet. I'd love some direction if there's a better way of handling this.

#### ~Problem (2) - `react-is`'s `typeOf` not accounting for `forwardRef` HOCs~

<details>
<summary>Outdated Comment (Issue was resolved in in #17278)</summary>
<br />
To fix the devtools issue mentioned above, I wanted to use `react-is`'s helpers. Specifically, `isMemo`, and `isForwardRef`. 

Looks like there's an outstanding issue with both of `ReactIs.isForwardRef` and `ReactIs.typeOf` where they don't respect a `forwardRef` HOC. It currently only works with elements created from a `fowardRef` component. 

```js
const RefForwardingComponent = React.forwardRef((props, ref) => null);

ReactIs.typeOf(RefForwardingComponent) // undefined
ReactIs.isForwardRef(RefForwardingComponent) // false

ReactIs.typeOf(<RefForwardingComponent />) // ReactIs.ForwardRef
ReactIs.isForwardRef(<RefForwardingComponent />) // true
```

I'm not sure if this is by design or intentional, but **it does seem inconsistent with how `isMemo` works**.

I made a 1-line change to `react-is` in this PR so that it respects `forwardRef` HOCs (the actual components/functions) not just the elements. Are there any pitfalls to know?

```js
const RefForwardingComponent = React.forwardRef((props, ref) => null);
 
// notice how I'm not passing an actual element
ReactIs.typeOf(RefForwardingComponent) // ReactIs.ForwardRef
ReactIs.isForwardRef(RefForwardingComponent) // true
```
</details>


packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -348,10 +348,11 @@ export function getInternalReactConstants(
       case FunctionComponent:
       case IndeterminateComponent:
         return getDisplayName(resolvedType);
+      case MemoComponent:
+      case SimpleMemoComponent:
       case ForwardRef:
         return (
-          resolvedType.displayName ||
-          getDisplayName(resolvedType.render, 'Anonymous')
+          resolvedType.displayName || getDisplayName(resolvedType, 'Anonymous')
```
wsmd: In cases of `memo` and `forwardRef`, we're delegating extracting the function/component type from those HOCs to the `getDisplayName` itself. I couldn't apply recursion here because `getDisplayNameForFiber` accepts a fiber node.

packages/react-is/src/__tests__/ReactIs-test.js
```diff
@@ -131,9 +133,10 @@ describe('ReactIs', () => {
 
   it('should identify memo', () => {
     const Component = () => React.createElement('div');
-    const memoized = React.memo(Component);
-    expect(ReactIs.typeOf(memoized)).toBe(ReactIs.Memo);
-    expect(ReactIs.isMemo(memoized)).toBe(true);
+    const Memoized = React.memo(Component);
+    expect(ReactIs.typeOf(Memoized)).toBe(ReactIs.Memo);
+    expect(ReactIs.typeOf(<Memoized />)).toBe(ReactIs.Element);
```
wsmd: > I'm not sure if this is by design or intentional, but it does seem inconsistent with how isMemo works.

As I mentioned above,

```js
// why is this ReactIs.ForwardRef
expect(ReactIs.typeOf(<RefForwardingComponent />)).toBe(ReactIs.ForwardRef);

// but this is ReactIs.Element 😕 
expect(ReactIs.typeOf(<Memoized />)).toBe(ReactIs.Element);
```

wsmd: Resolved in #17278

See https://github.com/facebook/react/pull/17278#discussion_r342690034

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -366,7 +385,7 @@ export function getInternalReactConstants(
         if (elementType.displayName) {
           return elementType.displayName;
         } else {
-          return getDisplayName(type, 'Anonymous');
+          return getDisplayName(resolvedType, 'Anonymous');
```
bvaughn: I had been thinking of suggesting a different approach that just kind of localized the check:
```diff
diff --git a/packages/react-devtools-shared/src/backend/renderer.js b/packages/react-devtools-shared/src/backend/renderer.js
index 761c78935..284183f9f 100644
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -365,6 +365,9 @@ export function getInternalReactConstants(
       case SimpleMemoComponent:
         if (elementType.displayName) {
           return elementType.displayName;
+        } else if (type.type != null && type.type.render != null) {
+          // Special case of a forwardRef() nested in a memo()
+          return getDisplayName(type.type.render, 'Anonymous');
         } else {
           return getDisplayName(type, 'Anonymous');
         }
```
I don't know that I really have a preference between the two approaches I guess. 😄 Yours is fine~

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -324,19 +324,39 @@ export function getInternalReactConstants(
...
+    MEMO_NUMBER,
+    MEMO_SYMBOL_STRING,
   } = ReactSymbols;
 
+  function resolveFiberType(type: any) {
+    // This is to support lazy components with a Promise as the type.
+    // see https://github.com/facebook/react/pull/13397
+    if (typeof type.then === 'function') {
+      return type._reactResult;
+    }
```
wsmd: I couldn't find any reference to `_reactResult` in the codebase. Is still relevant? Is it safe to remove?

bvaughn: I think it has sense been renamed to `_result` now that you mention it.

Someone can follow up with that in a separate PR.

cc @acdlite

packages/react-devtools-shared/src/backend/renderer.js
bvaughn: Tests added #17283

packages/react-devtools-shared/src/backend/renderer.js
bvaughn: I don't think this codepath is even used by DevTools anymore. By the time React DevTools gets called for the lazy component, it just gets the e.g. function.

PR #17278, opened 2019-11-05
Fix react-is memo and lazy type checks

bvaughn:
I think the `react-is` package is currently a little broken. This packages `typeOf` methods are meant to look at React _element_ types (e.g. `<MyComponent />` or `React.createElement(MyComponent)` not _component_ types (e.g. `MyComponent`), but it's inconsistent when it comes to lazy and memo types.

This inconsistency can be seen in shallow renderer as well with invariants like this:
```js
isForwardRef(element) || (typeof element.type === 'function' || isMemo(element.type)),
```
Rather than this:
```js
isForwardRef(element) || (typeof element.type === 'function' || isMemo(element)),
```

Is this change a bug fix or a breaking change? (Or both?) Let's discuss.

### DIFF
--- a/packages/react-is/src/__tests__/ReactIs-test.js
+++ b/packages/react-is/src/__tests__/ReactIs-test.js
@@ -71,6 +71,7 @@ describe('ReactIs', () => {
 
   it('should identify context consumers', () => {
     const Context = React.createContext(false);
+    expect(ReactIs.isValidElementType(Context.Consumer)).toBe(true);
     expect(ReactIs.typeOf(<Context.Consumer />)).toBe(ReactIs.ContextConsumer);
     expect(ReactIs.isContextConsumer(<Context.Consumer />)).toBe(true);
     expect(ReactIs.isContextConsumer(<Context.Provider />)).toBe(false);
@@ -79,6 +80,7 @@ describe('ReactIs', () => {
 
   it('should identify context providers', () => {
     const Context = React.createContext(false);
+    expect(ReactIs.isValidElementType(Context.Provider)).toBe(true);
     expect(ReactIs.typeOf(<Context.Provider />)).toBe(ReactIs.ContextProvider);
     expect(ReactIs.isContextProvider(<Context.Provider />)).toBe(true);
     expect(ReactIs.isContextProvider(<Context.Consumer />)).toBe(false);
@@ -106,6 +108,7 @@ describe('ReactIs', () => {
 
   it('should identify ref forwarding component', () => {
     const RefForwardingComponent = React.forwardRef((props, ref) => null);
+    expect(ReactIs.isValidElementType(RefForwardingComponent)).toBe(true);
     expect(ReactIs.typeOf(<RefForwardingComponent />)).toBe(ReactIs.ForwardRef);
     expect(ReactIs.isForwardRef(<RefForwardingComponent />)).toBe(true);
     expect(ReactIs.isForwardRef({type: ReactIs.StrictMode})).toBe(false);
@@ -113,6 +116,7 @@ describe('ReactIs', () => {
   });
 
   it('should identify fragments', () => {
+    expect(ReactIs.isValidElementType(React.Fragment)).toBe(true);
     expect(ReactIs.typeOf(<React.Fragment />)).toBe(ReactIs.Fragment);
     expect(ReactIs.isFragment(<React.Fragment />)).toBe(true);
     expect(ReactIs.isFragment({type: ReactIs.Fragment})).toBe(false);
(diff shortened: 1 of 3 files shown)


billyjanitsch:
@bvaughn have you seen the discussion in https://github.com/facebook/react/pull/14546?

Folks from the ecosystem were unsure whether the current behavior was intentional. But it's been around for long enough that several packages now depend on it. So it might be worth calling this a breaking change regardless of the original design intention.

bvaughn:
Thanks for the pointer, @billyjanitsch. I had not seen that discussion. I noticed this inconsistency while looking at a DevTools display complaint.

packages/react-is/src/__tests__/ReactIs-test.js
```diff
@@ -106,13 +108,15 @@ describe('ReactIs', () => {
 
   it('should identify ref forwarding component', () => {
     const RefForwardingComponent = React.forwardRef((props, ref) => null);
+    expect(ReactIs.isValidElementType(RefForwardingComponent)).toBe(true);
     expect(ReactIs.typeOf(<RefForwardingComponent />)).toBe(ReactIs.ForwardRef);
```
wsmd: I noticed an inconsistency here with how typeOf works when we pass a momozied element vs when we pass a ref-forwarding element. More info here https://github.com/facebook/react/pull/17274#discussion_r342368003

I think this PR addresses this inconsistency between the two types, but I'm still curious, what's the intended behavior for `typeOf(RefForwardingComponent)` and `typeOf(<RefForwardingComponent />)` – should we expect to get back `ReactIs.ForwardRef` in both cases? Same question goes for memoized elements/components.

For example:

```js
expect(ReactIs.typeOf(<Memoized />)).toBe(ReactIs.Memo);
expect(ReactIs.typeOf(Memoized)).toBe(ReactIs.Memo);

expect(ReactIs.typeOf(<RefForwardingComponent />)).toBe(ReactIs.ForwardRef);
expect(ReactIs.typeOf(RefForwardingComponent)).toBe(ReactIs.ForwardRef);
```

The proposed bug fix in #17274 relies on those assertions above being valid.

bvaughn: > I think this PR addresses this inconsistency between the two types, but I'm still curious, what's the intended behavior for typeOf(RefForwardingComponent) and typeOf(<RefForwardingComponent />) – should we expect to get back ReactIs.ForwardRef in both cases?

No. That's the primary purpose of this PR, to _not_ support that use case.

I've seen #17274. I have some draft comments on it but got distracted by this PR first.

wsmd: Thanks for the clarification @bvaughn! Exactly the answer I was looking for!

ljharb:
Heads up that v16.12 seems to have broken enzyme, which makes it arguably a breaking change. See https://github.com/airbnb/enzyme/pull/2292 for an example (`isMemo`, specifically).

bvaughn:
@ljharb FWIW I tagged you about this upcoming change on https://github.com/facebook/react/pull/14546#issuecomment-549913143

I also tested this change against Enzyme before merging:
> I've checked out Enzyme and installed a local build of `react-is` (with my change from https://github.com/facebook/react/pull/17278) and all of the tests pass without any code changes on the Enzyme side.

> I think that's because Enzyme is _already_ [wrapping values in "fake elements"](https://github.com/airbnb/enzyme/blob/8e2d65cc4afdbca0df34308a75bb96fa64e46ce6/packages/enzyme-adapter-react-16.3/src/ReactSixteenThreeAdapter.js#L254-L256) before passing them to `react-is`. So while that PR could be argued a breaking change, it doesn't seem to break Enzyme.

ljharb:
hm, i do recall that and i appreciate the care :-)

the isMemo checks are what’s breaking now; are you saying that the “fake elements” part is what needs to change?

ljharb:
It’s also possible that some other part of 16.12 is what’s breaking the memo tests, but that seems less likely.

bvaughn:
> the isMemo checks are what’s breaking now; are you saying that the “fake elements” part is what needs to change?

Hm, no I don't think so. It looks like it would be enough to fool the react-is check.

ljharb:
Pinning react-is to 16.11 in https://github.com/airbnb/enzyme/pull/2292 did fix it (the test failure is a linting thing), so it does seem to be related.

bvaughn:
> the test failure is a linting thing

bvaughn:
I'm not really sure what to make of the Enzyme CI failure, but it looks like there are 23 failing tests - none of which seeming to have anything to do with react-is on the surface. A lot of them seem to mention hooks:
Screen Shot 2019-11-26 at 9 38 00 AM

Screen Shot 2019-11-26 at 9 37 39 AM

ljharb:
The [previous run](https://travis-ci.org/airbnb/enzyme/jobs/616937071) had a number of memoization failures, also some Suspense ones. There may be some hook failures in other jobs, but i didn't get that far.

bvaughn:
Can you reduce the failure case to something smaller?

ljharb:
@bvaughn thanks for the offer! in the meantime a user has come up with a [working fix](https://github.com/airbnb/enzyme/pull/2300) so i'll move forward with that.

nighca:
So now.. which is the recommended way to check if a component (instead of an element) is memo component?

PR #17282, opened 2019-11-05
Update deepDiffer usage in React Native renderer

motiz88:
React Native exposes a function, `deepDiffer`, which `react-native-renderer` uses to compute prop updates. This comparison function has surprising behaviour when it comes to function values: it will treat any two functions as equal, entirely ignoring function identity. 

https://github.com/facebook/react-native/commit/99d229e186e1e5d08cb1fc417b4b911ade7aef33 added an explicit flag to switch this surprising behaviour on or off - the default behaviour is unchanged at the moment, but I plan to change it eventually once I've accounted for all existing call sites.

This PR opts `react-native-renderer` _into_ the surprising behaviour (it's the one place it actually makes sense, and probably where it originated - see e.g. [this comment](https://github.com/facebook/react/blob/053cf0fedc91a1507080afe43d3be354ec346e9e/packages/react-native-renderer/src/__tests__/ReactNativeAttributePayload-test.js#L209-L210)). I've also updated the vendored `deepDiffer` mock to match the version currently in React Native master, and added a test explicitly covering this behaviour.

### DIFF
--- a/packages/react-native-renderer/src/ReactNativeAttributePayload.js
+++ b/packages/react-native-renderer/src/ReactNativeAttributePayload.js
@@ -38,7 +38,7 @@ function defaultDiffer(prevProp: mixed, nextProp: mixed): boolean {
     return true;
   } else {
     // For objects and arrays, the default diffing algorithm is a deep compare
-    return deepDiffer(prevProp, nextProp);
+    return deepDiffer(prevProp, nextProp, {unsafelyIgnoreFunctions: true});
   }
 }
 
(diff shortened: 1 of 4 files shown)


packages/react-native-renderer/src/ReactNativeAttributePayload.js
```diff
@@ -38,7 +38,7 @@ function defaultDiffer(prevProp: mixed, nextProp: mixed): boolean {
     return true;
   } else {
     // For objects and arrays, the default diffing algorithm is a deep compare
-    return deepDiffer(prevProp, nextProp);
+    return deepDiffer(prevProp, nextProp, {unsafelyIgnoreFunctions: true});
```
gaearon: This is creating a new object every time. Given that it's static, can we please hoist it out?

PR #17283, opened 2019-11-05
Added a DevTools store test for component names

bvaughn:
Follow up to #17274

1. Add a Store test for display name handling for `memo`, `forwardRef`, and `lazy` components.
2. Remove outdated dead code for `React.lazy`.
3. Update DT tests to include HOC badge names in the serialized store

### DIFF
--- a/packages/react-devtools-shared/src/devtools/utils.js
+++ b/packages/react-devtools-shared/src/devtools/utils.js
@@ -7,6 +7,11 @@
  * @flow
  */
 
+import {
+  ElementTypeForwardRef,
+  ElementTypeMemo,
+} from 'react-devtools-shared/src/types';
+
 import type {Element} from './views/Components/types';
 import type Store from './store';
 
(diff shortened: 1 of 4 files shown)


packages/react-devtools-shared/src/devtools/utils.js
```diff
@@ -21,10 +26,25 @@ export function printElement(element: Element, includeWeight: boolean = false) {
...
+    }
+  } else if (element.type === ElementTypeForwardRef) {
+    if (hocDisplayNames === null) {
+      hocDisplayNames = ['ForwardRef'];
+    } else {
+      hocDisplayNames.push('ForwardRef');
+    }
+  }
+
+  let hocs = hocDisplayNames === null ? '' : ` [${hocDisplayNames.join('][')}]`;
```
bvaughn: Serialize the store in a format that includes the HOC badges, more like the `Badge` component does in the real DevTools UI. This will make our tests more meaningful.

threepointone: Yeah this is pretty. Nice.

ibarrae:
One question folks, has this been released already? I couldn't find any entry on the tags @bvaughn @threepointone

gaearon:
Why does it matter? This is an update to an internal test.

ibarrae:
> Why does it matter? This is an update to an internal test.

You are right, but I will assume that if this PR is not related, then #17274 has not been released either since they are both related.

gaearon:
Pretty sure https://github.com/facebook/react/pull/17274 is released.

ibarrae:
Awesome @gaearon thanks for the quick response :tada:

bvaughn:
> I couldn't find any entry on the tags

I'm not sure what you mean by this but #17274 is listed [in the changelog](https://github.com/facebook/react/blob/master/packages/react-devtools/CHANGELOG.md#bug-fixes-1) as being released back in November.

PR #17284, opened 2019-11-06
[Failing tests] startTransition invalidates downstream memoizations

acdlite:
Like `dispatch` or `setState`, we should return the same `startTransition` function on every render. This prevents it from invalidating downstream memoizations, like `useCallback`.

We do use `useCallback` internally so that a new `startTransition` is only created when the config object changes. However, if you pass an object, this does nothing because the object different every time.

Potential fixes:

- Instead of detecting changes in the config object, compare the fields (`timeoutMs` et al) directly. This fixes the worst case (inline config object) but still means we'll occasionally re-create `startTransition` functions.
- Track the current config object on an internal mutable field using an effect and read from that instead of binding a new method each time. This requires extra work in the commit phase, but combined with the previous option (comparing the fields instead of the entire config object) that should be rare.
- Instead of passing the config object to the `useTransition` hook, pass it to the `startTransition` function. This avoids needing to bind the config object to the method. Slightly less ergonomic, however, we're planning to change the default config object so that it's not needed in most cases. So it's only in special cases that you'd actually use one. **This is the most likely option**.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
@@ -2111,6 +2111,183 @@ describe('ReactHooksWithNoopRenderer', () => {
         ]);
       },
     );
+
+    it.experimental('always returns the same startTransition', async () => {
+      let transition;
+      function App() {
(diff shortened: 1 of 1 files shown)


acdlite:
Closed by #19719

PR #17285, opened 2019-11-06
[Flight] Basic Streaming Suspense Support

sebmarkbage:
This adds basic Suspense support to Flight by streaming the response.

First it sends one line with the root JSON model. Within the model holes are replaced with "$" followed by a hex ID. E.g. "$12AB3".

Subsequent lines begin with "J" followed by a hex ID.

So an example response may look something like:

```
{"some":{"json":"$1"},data:["$1","$2"]}
J1:{"more":"data"}
J2:{"even":"more"}
```

The client suspends if you try to access an unresolved value. Once a new line comes in it unsuspends that part of the model.

Currently this suspending is implemented as a getter but it could also be a Proxy. However, more likely is that we'll hide this behind an explicit API that unwraps both components and its data in the future.

## Errors

If the server errors while rendering one of the models that error is encoded with an "E" prefix:

```
{"bad":"$1","good":"$2"}
E1:{"message":"An error happened","stack": "\n..."}
J2:{"this":"worked"}
```

When this happens, the previously suspended part of the model now resolves and starts throwing an error on the client. This will then be caught by an error boundary on the client. Unrelated parts of the response can still continue rendering though.

If the connection dies, then all remaining pending promises resolve and start throwing an error which are caught by error boundaries. Previously resolved parts don't error though.

There's a prioritization scheme so errors are emitted at lower priority in the stream than models.

## Future Additions

I added a comment describing the protocol. The protocol also includes space for a few other types of data.

- _URL_: This can be used to emit URL rows at higher priority than models and can be used to preload other resources without waiting for or even parsing the JSON models yet.
- _Blob_: Traditionally sending binary data is a PITA in existing protocol. We should just have this built-in. It's especially efficient with modern stream APIs. This is important in data heavy use cases like data viz.
- _HTML_: Instead of encoding HTML inside JSON we can just send it as raw HTML for faster parsing and more efficient size. We can also encode that we generated this HTML (using React SSR with proper data structures) in the encoding. So we can by-pass dangerouslySetInnerHTML and just provide a React element that can be rendered directly.

### DIFF
--- a/packages/react-flight/src/ReactFlightClient.js
+++ b/packages/react-flight/src/ReactFlightClient.js
@@ -20,43 +20,113 @@ export type ReactModelRoot<T> = {|
   model: T,
 |};
 
-type OpaqueResponse = {
+type JSONValue = number | null | boolean | string | {[key: string]: JSONValue};
+
+const PENDING = 0;
+const RESOLVED = 1;
+const ERRORED = 2;
+
+type PendingChunk = {|
+  status: 0,
+  value: Promise<void>,
+  resolve: () => void,
+|};
+type ResolvedChunk = {|
+  status: 1,
+  value: mixed,
+  resolve: null,
+|};
+type ErroredChunk = {|
+  status: 2,
+  value: Error,
+  resolve: null,
+|};
+type Chunk = PendingChunk | ResolvedChunk | ErroredChunk;
+
+type OpaqueResponseWithoutDecoder = {
   source: Source,
-  modelRoot: ReactModelRoot<any>,
   partialRow: string,
+  modelRoot: ReactModelRoot<any>,
+  chunks: Map<number, Chunk>,
+  fromJSON: (key: string, value: JSONValue) => any,
+};
+
+type OpaqueResponse = OpaqueResponseWithoutDecoder & {
   stringDecoder: StringDecoder,
-  rootPing: () => void,
 };
 
 export function createResponse(source: Source): OpaqueResponse {
-  let modelRoot = {};
-  Object.defineProperty(
-    modelRoot,
-    'model',
-    ({
-      configurable: true,
-      enumerable: true,
-      get() {
-        throw rootPromise;
-      },
-    }: any),
-  );
+  let modelRoot: ReactModelRoot<any> = ({}: any);
+  let rootChunk: Chunk = createPendingChunk();
+  definePendingProperty(modelRoot, 'model', rootChunk);
+  let chunks: Map<number, Chunk> = new Map();
+  chunks.set(0, rootChunk);
 
-  let rootPing;
-  let rootPromise = new Promise(resolve => {
-    rootPing = resolve;
-  });
-
-  let response: OpaqueResponse = ({
+  let response: OpaqueResponse = (({
     source,
-    modelRoot,
     partialRow: '',
-    rootPing,
-  }: any);
+    modelRoot,
+    chunks: chunks,
+    fromJSON: function(key, value) {
+      return parseFromJSON(response, this, key, value);
+    },
+  }: OpaqueResponseWithoutDecoder): any);
   if (supportsBinaryStreams) {
     response.stringDecoder = createStringDecoder();
   }
   return response;
+}
+
+function createPendingChunk(): PendingChunk {
+  let resolve: () => void = (null: any);
+  let promise = new Promise(r => (resolve = r));
+  return {
+    status: PENDING,
+    value: promise,
+    resolve: resolve,
+  };
+}
+
+function createErrorChunk(error: Error): ErroredChunk {
+  return {
+    status: ERRORED,
+    value: error,
+    resolve: null,
+  };
+}
+
+function triggerErrorOnChunk(chunk: Chunk, error: Error): void {
+  if (chunk.status !== PENDING) {
+    // We already resolved. We didn't expect to see this.
+    return;
+  }
+  let resolve = chunk.resolve;
+  let erroredChunk: ErroredChunk = (chunk: any);
+  erroredChunk.status = ERRORED;
+  erroredChunk.value = error;
+  erroredChunk.resolve = null;
+  resolve();
+}
+
+function createResolvedChunk(value: mixed): ResolvedChunk {
+  return {
+    status: RESOLVED,
+    value: value,
+    resolve: null,
+  };
+}
+
+function resolveChunk(chunk: Chunk, value: mixed): void {
+  if (chunk.status !== PENDING) {
+    // We already resolved. We didn't expect to see this.
+    return;
+  }
+  let resolve = chunk.resolve;
+  let resolvedChunk: ResolvedChunk = (chunk: any);
+  resolvedChunk.status = RESOLVED;
+  resolvedChunk.value = value;
+  resolvedChunk.resolve = null;
+  resolve();
 }
 
 // Report that any missing chunks in the model is now going to throw this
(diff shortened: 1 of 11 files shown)


packages/react-server/src/ReactFlightServer.js
```diff
@@ -42,77 +44,241 @@ type ReactModelObject = {
...
+    // This is a host element. E.g. HTML.
+    return renderHostChildrenToString(element);
+  } else {
+    throw new Error('Unsupported type.');
+  }
+}
+
+function pingSegment(request: OpaqueRequest, segment: Segment): void {
+  let pingedSegments = request.pingedSegments;
+  pingedSegments.push(segment);
```
gaearon: Could this segment already be in the array? What would happen in that case? Is it okay to have it twice?

packages/react-server/src/ReactFlightServer.js
```diff
@@ -42,77 +44,241 @@ type ReactModelObject = {
...
+function retrySegment(request: OpaqueRequest, segment: Segment): void {
+  let value = segment.model;
+  try {
+    while (
+      typeof value === 'object' &&
+      value !== null &&
+      value.$$typeof === REACT_ELEMENT_TYPE
+    ) {
+      // If this is a nested model, there's no need to create another chunk,
+      // we can reuse the existing one and try again.
```
gaearon: Do I understand correctly this is just an optimization to avoid useless chunks like `J2:"$5"`? It has no impact on correctness?

packages/react-server/src/ReactFlightServer.js
```diff
@@ -42,77 +44,241 @@ type ReactModelObject = {
...
+    let json = stringify(value, request.toJSON);
+    let row;
+    let id = segment.id;
+    if (id === 0) {
+      row = json + '\n';
+    } else {
+      row = serializeRowHeader('J', id) + json + '\n';
+    }
+    request.completedJSONChunks.push(convertStringToBuffer(row));
+  } catch (x) {
```
gaearon: Nit: any downside to leaving try-catch around user code more closely? It's weird that making a typo here (e.g. `request.competedJSONChunks`) would lead to streaming an error chunk rather than a crash in the server. Although I guess we have similar undefined behavior in the reconciler code too.

packages/react-server/src/ReactFlightServer.js
```diff
@@ -42,77 +44,241 @@ type ReactModelObject = {
...
+    if (typeof x === 'object' && x !== null && typeof x.then === 'function') {
+      // Something suspended again, let's pick it back up later.
+      let ping = segment.ping;
+      x.then(ping, ping);
+      return;
+    } else {
+      // This errored, we need to serialize this error to the
```
gaearon: to the...?

packages/react-server/src/ReactFlightServer.js
```diff
@@ -42,77 +44,241 @@ type ReactModelObject = {
...
 function performWork(request: OpaqueRequest): void {
-  let rootModel = request.model;
-  request.model = null;
-  let json = JSON.stringify(rootModel, resolveModelToJSON);
-  request.completedChunks.push(convertStringToBuffer(json));
+  let pingedSegments = request.pingedSegments;
+  request.pingedSegments = [];
+  for (let i = 0; i < pingedSegments.length; i++) {
+    let segment = pingedSegments[i];
+    retrySegment(request, segment);
```
gaearon: Is this guaranteed to not be reentrant? What if we're dealing with a synchronous thenable? Do we assume pings always happen on the next tick?

packages/react-server/src/ReactFlightServer.js
```diff
@@ -42,77 +44,241 @@ type ReactModelObject = {
...
-    for (let i = 0; i < chunks.length; i++) {
-      let chunk = chunks[i];
-      writeChunk(destination, chunk);
+    let jsonChunks = request.completedJSONChunks;
+    let i = 0;
+    for (; i < jsonChunks.length; i++) {
+      request.pendingChunks--;
+      let chunk = jsonChunks[i];
+      if (!writeChunk(destination, chunk)) {
+        request.flowing = false;
```
gaearon: I don't understand what this means and what the `flowing` property is doing. Can you explain a bit?

packages/react-server/src/ReactFlightServer.js
```diff
@@ -235,7 +235,12 @@ function performWork(request: OpaqueRequest): void {
   }
 }
 
+let reentrant = false;
 function flushCompletedChunks(request: OpaqueRequest): void {
+  if (reentrant) {
+    return;
+  }
+  reentrant = true;
```
gaearon: I don't understand what this is for.

gaearon:
If connection dies, how do I implement a Retry button?

packages/react-flight/src/ReactFlightClient.js
```diff
@@ -20,100 +20,269 @@ export type ReactModelRoot<T> = {|
...
-  let model = JSON.parse(response.partialRow);
-  Object.defineProperty(modelRoot, 'model', {
-    value: model,
-  });
-  response.rootPing();
+  // In case there are any remaining unresolved chunks, they won't
+  // be resolved now. So we need to issue an error to those.
+  // Ideally we should be able to early bail out if we kept a
+  // ref count of pending chunks.
+  reportGlobalError(response, new Error('Connection closed.'));
```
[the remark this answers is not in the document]
sebmarkbage: We should probably implement the optimization where if all rows are satisfied we don’t even iterate over the set. In that case it’s not as confusing.

packages/react-server/src/ReactFlightServer.js
sebmarkbage: I had to hack it because I didn’t realize whatwg streams are reentrant. Will need a better refactor to support both types of a streams.

packages/react-server/src/ReactFlightServer.js
sebmarkbage: It’s ok to have it twice but it shouldn’t happen since a component only throws one promise at a time.

packages/react-server/src/ReactFlightServer.js
sebmarkbage: Yea barring any bugs.

packages/react-server/src/ReactFlightServer.js
sebmarkbage: toJSON can be invoked on objects here which is user code. It’s primarily meant to catch such things. Not implementation bugs.

packages/react-server/src/ReactFlightServer.js
sebmarkbage: thing

packages/react-server/src/ReactFlightServer.js
sebmarkbage: This whole function is invoked by the “scheduler” (setImmediate) so it’s never reentrant.

sebmarkbage: The primary purpose of the scheduler is to let multiple things get pinged and flushing to get in before we continue doing work in a batch.

packages/react-server/src/ReactFlightServer.js
sebmarkbage: Flowing means that we can keep writing more data to the buffer once we have it.

We don’t write eagerly to the output stream if it is full. That lets us emit higher pri things if they become available before the stream buffer becomes not full.

sebmarkbage:
Error boundary and do the request from the root again.

gaearon:
Let me rephrase the tests question. Do you want _me_ to write tests? 😄

gaearon:
I can also just make an app..

sebmarkbage:
If you want to write tests that would be great. Like the suspending and error cases.

gaearon:
Cool, I'll do that.

PR #17289, opened 2019-11-06
[Flight] Better compat with http.createServer

gaearon:
Don't know if we care in longer term but I use `http.createServer` for testing and that has older Node streams. They don't have `cork` and `uncork`. Also the `flush` on that object was misleading (because it was only used for flushing headers) so it fires a deprecation warning.

That's a few workarounds to make it work with `http.createServer`. Maybe we'll remove it later but that's the most convincing way to "really" test it that I found so far.


gaearon:
I guess we'll see what objects returned by Express etc are like. Maybe they shield from this and we won't need it.

PR #17291, opened 2019-11-06
[react-interactions] Change unmount blur logic to a dedicated event

trueadm:
This PR clears up some confusion internally and aligns the behavior of an unmounted node with focus to a specific unique event. This means we can also dispatch a new specific event for this action from the Focus Responder – removing the issues that might arise from mixing it with the existing `blur` and `focus` logic. I've opted to call the event `detachedvisiblenode` or `onDetachedVisibleNode` – but better names are surely out there!

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -452,7 +452,11 @@ export function insertInContainerBefore(
   }
 }
 
-function handleSimulateChildBlur(
+// This is a specific event for the React Flare
+// event system, so event responders can act
+// accordingly to a DOM node being unmounted that
+// previously had active document focus.
+function dispatchDetachedVisibleNodeEvent(
   child: Instance | TextInstance | SuspenseInstance,
 ): void {
   if (
@@ -463,13 +467,11 @@ function handleSimulateChildBlur(
     const targetFiber = getClosestInstanceFromNode(child);
     // Simlulate a blur event to the React Flare responder system.
     dispatchEventForResponderEventSystem(
-      'blur',
+      'detachedvisiblenode',
       targetFiber,
       ({
-        relatedTarget: null,
         target: child,
         timeStamp: Date.now(),
-        type: 'blur',
       }: any),
       ((child: any): Document | Element),
       RESPONDER_EVENT_SYSTEM | IS_PASSIVE,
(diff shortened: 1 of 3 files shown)


necolas:
Help me understand why we need this and what the internal confusion is?

trueadm:
@necolas It's the same reason we needed it before – we want to know when a node (that has active document focus) gets unmounted from the document – so we can restore focus to another nearby node. The issue with overloading it with `blur` and `focus`, is that we have no way of knowing from the other events if the heuristic was because it was actually from this codepath. That way we can internally GK this specific logic in case we run into issues.

necolas:
Why is it called `detachvisiblenode` rather than say `detachActiveElement`? Or why not add data to the `blur` event to indicate if it was caused by the node being removed from the DOM? The latter would avoid us making up some fake native event

trueadm:
@necolas `detachActiveElement` would be a better name – I can follow up with that name change next week. We don't want to use `blur` here because other even responders then need to know about the data being in existence to know there's something special about this event. All of this doesn't need to be perfect for now, as this is only a temporary solution to unblock internal solutions – until we come up with something that can happen outside of the mutation phase (like Sebastian mentioned in the previous diff that I mention in the summary).

I think a better approach would likely to be an entirely different mechanic that we can get around to thinking when we look at the event system as a whole again in the future. I've got a Quip with all these strange odd cases we've run into in regards to events this year – so it's all collected knowledge for requirements we need to handle at a later time.

necolas:
I don't follow why the new event is a better interim solution, or why a new callback is needed in the responder, when this seems to be a special type of blur event.

trueadm:
@necolas This is meant for a very bespoke usage right now for accessibility. I might prefix it all with DO_NOT_USE. 

With a new event system, none of this would be exposed inside a focus-like interface, instead the accessibility component would be able to ask ReactDOM to let it know directly – in a future world.

necolas:
Ok. Plenty of flare stuff has leaked into production dependencies already, including across platforms, so I'm not convinced what got merged is going to be restricted any better than alternatives that don't expand the API surface.

Do you want us to stamp flare patches with the expectation that we'll remove the entire system in the future?

trueadm:
I'm not disagreeing with your comments, I'm just saying that these can all be addressed in follow ups. What do you think the better strategy here is for us to apply in a follow up?

PR #17303, opened 2019-11-07
[RN] Hoist static deepDiffer options object

motiz88:
Hoists the options object literal from `defaultDiffer` to the module scope so it's not recreated on every prop update. Addresses https://github.com/facebook/react/pull/17282#discussion_r343677204.


gaearon:
Thanks!

PR #17307, opened 2019-11-07
[Flight] Basic Integration Test

gaearon:
This adds a basic integration test where we actually feed the model to ReactDOM components and use Suspense. The test verifies that we can stream data and reveal more content as it happens. We also verify errors work.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
+++ b/packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
@@ -53,7 +53,7 @@ describe('ReactFlightIntegration', () => {
     };
   }
 
-  it('should resolve the root', async () => {
+  it.experimental('should resolve the root', async () => {
     let {Suspense} = React;
 
     // Model
@@ -103,7 +103,7 @@ describe('ReactFlightIntegration', () => {
     );
   });
 
-  it('should not get confused by $', async () => {
+  it.experimental('should not get confused by $', async () => {
     let {Suspense} = React;
 
     // Model
@@ -138,7 +138,7 @@ describe('ReactFlightIntegration', () => {
     expect(container.innerHTML).toBe('<p>$1</p>');
   });
 
-  it('should progressively reveal chunks', async () => {
+  it.experimental('should progressively reveal chunks', async () => {
     let {Suspense} = React;
 
     class ErrorBoundary extends React.Component {
--- a/packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
+++ b/packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
@@ -13,6 +13,7 @@
 global.ReadableStream = require('@mattiasbuelens/web-streams-polyfill/ponyfill/es6').ReadableStream;
 global.TextDecoder = require('util').TextDecoder;
 
+let act;
 let Scheduler;
 let Stream;
 let React;
@@ -23,6 +24,7 @@ let ReactFlightDOMClient;
 describe('ReactFlightIntegration', () => {
   beforeEach(() => {
     jest.resetModules();
+    act = require('react-dom/test-utils').act;
     Scheduler = require('scheduler');
     Stream = require('stream');
     React = require('react');
@@ -47,18 +49,6 @@ describe('ReactFlightIntegration', () => {
       writable,
       readable,
     };
-  }
-
-  async function waitForDOMChange(container) {
-    let initialHTML = container.innerHTML;
-    for (let i = 0; i < 1000; i++) {
-      jest.runAllTimers();
-      await Promise.resolve();
-      Scheduler.unstable_flushAll();
-      if (container.innerHTML !== initialHTML) {
-        break;
-      }
-    }
   }
 
   it('should resolve the root', async () => {
@@ -102,7 +92,9 @@ describe('ReactFlightIntegration', () => {
     ReactDOM.render(<App result={result} />, container);
     expect(container.innerHTML).toBe('<h1>Loading...</h1>');
 
-    await waitForDOMChange(container);
+    await act(async () => {
+      jest.runAllTimers();
+    });
     expect(container.innerHTML).toBe(
       '<p><div><span>hello</span><span>world</span></div></p>',
     );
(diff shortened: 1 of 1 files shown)


packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
```diff
@@ -0,0 +1,319 @@
...
+        });
+      },
+    });
+    return {
+      writable,
+      readable,
+    };
+  }
+
+  async function waitForDOMChange(container) {
```
gaearon: This is kinda wonky but I wanted to check for next mutation specifically rather than manually wait for models. Since React itself would wait for models and I want to test it waits correctly.

sebmarkbage: Isn’t this what async act is for? Why isn’t it sufficient? It’s supposed to force flush even if it suspends and wait for nested promises. 

What’s missing and can we fix that?

gaearon: Lol I forgot about it. Let me try it

gaearon:
OK this should work. I need two of them to flush Promise ticks for both client and server.

gaearon:
ready for another

packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
```diff
@@ -0,0 +1,333 @@
...
+      '<div>:name::avatar:</div>' +
+        '<p style="">(loading sidebar)</p>' +
+        '<p style="">(loading posts)</p>' +
+        '<p>Game over</p>', // TODO: should not have message in prod.
+    );
+
+    // We can now show the sidebar.
+    resolvePhotosModel();
+    await act(async () => {
+      jest.runAllTimers();
```
acdlite: There are no timers in any of your test helpers so you don't need these calls.

I would put `resolvePhotosModel` inside the `act` call, even though it happens to work without it:

```js
await act(async () => {
  resolvePhotosModel();
});
```

gaearon: The Node server uses setImmediate for scheduling. So I need at least one per server work request I think.

sebmarkbage: We probably need a FlgihtServer.act(...) for that.

packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
```diff
@@ -0,0 +1,333 @@
...
+    });
+    expect(container.innerHTML).toBe(
+      '<div>:name::avatar:</div>' +
+        '<p style="">(loading sidebar)</p>' +
+        '<p style="">(loading posts)</p>' +
+        '<p>Game over</p>', // TODO: should not have message in prod.
+    );
+
+    // We can now show the sidebar.
+    resolvePhotosModel();
```
sebmarkbage: This should be inside the `act(...)` call. That's the side-effect that's acting.

packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
```diff
@@ -0,0 +1,333 @@
...
+        </Suspense>
+      );
+    }
+
+    let {writable, readable} = getTestStream();
+    ReactFlightDOMServer.pipeToNodeWritable(<RootModel />, writable);
+    let result = ReactFlightDOMClient.readFromReadableStream(readable);
+
+    let container = document.createElement('div');
+    ReactDOM.render(<App result={result} />, container);
```
sebmarkbage: We should only use concurrent mode in these tests. There are some features that might actually need it or might need a breaking change. E.g. thinking about repurposing the second argument to function components.

packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
```diff
@@ -0,0 +1,333 @@
...
+        '<p>Game over</p>', // TODO: should not have message in prod.
+    );
+
+    // We can now show the sidebar.
+    resolvePhotosModel();
+    await act(async () => {
+      jest.runAllTimers();
+    });
+    await act(async () => {
+      jest.runAllTimers();
```
sebmarkbage: Why the second one though?

packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
```diff
@@ -0,0 +1,333 @@
...
+        '<p>(loading)</p>',
+    );
+
+    // We can now show the details. Sidebar and posts are still loading.
+    resolveNameModel();
+    await act(async () => {
+      jest.runAllTimers();
+    });
+    await act(async () => {
+      jest.runAllTimers();
```
sebmarkbage: This is again two of them. This shouldn't be needed since act should be doing it recursively.

gaearon:
I pushed an update that shims `setImmediate` which lets us only have the necessary `act`s. Otherwise we'd still need two for each — because there's timers between Promises, and act can't help with that. Alternatively we could have [something like this](https://github.com/facebook/react/blob/38dd17ab98ce288fd0d0b68682a6df0f0a49e158/packages/react-devtools-shared/src/__tests__/utils.js#L51-L60) but I'd rather not.

gaearon:
With act, prod experimental build fails (?!)

packages/react-dom/src/__tests__/ReactFlightIntegration-test.js
```diff
@@ -0,0 +1,317 @@
...
+    let result = ReactFlightDOMClient.readFromReadableStream(readable);
+
+    let container = document.createElement('div');
+    let root = ReactDOM.createRoot(container);
+    act(() => {
+      root.render(<App result={result} />);
+    });
+    expect(container.innerHTML).toBe('<h1>Loading...</h1>');
+
+    await act(async () => {});
```
sebmarkbage: How does not doing anything help resolve it? Is it just because of a promise tick?

Does making the first one an async act avoid this?

gaearon: Yeah. If we make the first one async, we lost the opportunity to observe the initial loading state.

I guess I could resolve it manually to delay it.

gaearon: We have another test that verifies gradual loading states so I can make this one not observe it.

sebmarkbage: There might not be an intermediate state *without* act because the scheduler would wait until after the tick and we end up delaying a bit (JND). So not sure it's even correct to test this. Certainly not an important detail.

PR #17309, opened 2019-11-07
Pass suspense config to `startTransition` instead of `useTransition`

acdlite:
This avoids the problem of having to bind the config object to `startTransition`, invalidating downstream memoizations.

See #17284 for more details.

I'll need to codemod the existing callers in www, but that's not a blocker because www uses a userspace fork.


gaearon:
Wanna update the docs? 😄

gaearon:
j/k. I can do it.

acdlite:
@gaearon Not sure if we've reached consensus on the API change yet, just preparing this PR for when we do. But yeah I was planning to update the docs once it lands :)

acdlite:
Haven't abandoned this. Still need to decide on final API.

acdlite:
Closed by https://github.com/facebook/react/pull/19703

PR #17315, opened 2019-11-08
Fix missing file check

gaearon:
This fixes the case where you `yarn test-build` and have leftover empty dirs without `package.json`. I think the original check was meant to catch that case but it wasn't right.


gaearon:
I forgot that I fixed this and later fixed it again.

PR #17316, opened 2019-11-08
Add yarn build --unsafe-partial

gaearon:
We normally delete the `build` folder when rebuilding for consistency. This can be annoying if I make a tiny scoped edit to a particular package. This is an escape hatch so I can iterate faster when I know what I'm doing.

### DIFF
--- a/scripts/rollup/build.js
+++ b/scripts/rollup/build.js
@@ -646,7 +646,9 @@ function handleRollupError(error) {
 }
 
 async function buildEverything() {
-  await asyncRimRaf('build');
+  if (!argv.partial) {
+    await asyncRimRaf('build');
+  }
 
   // Run them serially for better console output
   // and to avoid any potential race conditions.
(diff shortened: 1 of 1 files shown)


gaearon:
You think so? I actually like the entry point granularity.

bvaughn:
The case I'd like to avoid is one of us testing an outdated/unrealistic local `node_module` (with stuff left over in the package from a previous build).

So e.g.
```
yarn build -- react,react-dom --type=NODE      
```

I think it would be safer to start with
```
 rm -rf build/node_modules/react
 rm -rf build/node_modules/react-dom
```

gaearon:
Maybe I'll rename to `partial-dangerous`? I really don't intend this to be used widely. Only for cases where you're debugging why some prod test is failing or similar.

gaearon:
Changed to `unsafe-partial`.

PR #17319, opened 2019-11-08
[Flight] End-to-End Fixture

gaearon:
Adds a fixture setup for Flight that's more realistic.

- Client: CRA + readFromFetch.
- Server: Express + pipeToNodeWritable.

This probably doesn't actually _stream_ now.
It's just barebones setup to render something.

I will be adding a more realistic endpoint, routing, and streaming in a follow-up.


sebmarkbage:
This fixture doesn't actually work for me. Complains about Flow syntax. Some Babel thing is not working.

Issue #15435, opened 2019-04-17
Add useReactPath hook for reflection and stable IDs


mike-marcacci:
**Do you want to request a *feature* or report a *bug*?**
Feature: this is a proposal that would solve #1137 and other similar cases by providing a reflection hook that retrieves the path in the react tree to a component.

**What is the current behavior?**
There is currently no mechanism for finding a component's path in the react component tree without using unstable, internal APIs.

**What is the expected behavior?**
There are a wide range of use-cases that require stable, unique namespaces that are stable across app instances.

My particular use case is rendering SVGs that include `<defs>` blocks. The challenge is that these definitions are all globally scoped. It's fairly easy to get around this by prefixing these definitions with an id that is stable and unique to the component instance:

```jsx
import { useMemo, ReactElement } from "react";

function gen(): string {
  return Math.random()
    .toString(36)
    .substr(2, 5);
}

function useUniqueId(): string {
  return useMemo(gen, []);
}

function SomeComponent(): ReactElement {
  const id = useUniqueId();
  return (
    <svg>
      <defs>
        <circle id={`${id}-myCircle`} cx="0" cy="0" r="5" />
      </defs>

      <use x="5" y="5" xlink:href={`#${id}-myCircle`} fill="red" />
    </svg>
  )
}

```

While this works well, the generated ID is different each time a component is instantiated. This means that it's not possible to safely rehydrate server-rendered SVGs, as the HTML will have the wrong prefix.

Instead, I would like a hook that provides the react path to the current component. For example:

```jsx
function SomeComponent(): ReactElement {
  const path = useReactPath();
  return <pre>{JSON.stringify(path, null, 2)}</pre>
}

// somewhere else:

<div>
  <div>Foo</div>
  <div>{[
    <div key="bar">Bar</div>,
    <div key="baz">
      Baz
      <SomeComponent />
    </div>
  ]}</div>
</div>
```

We would expect the path to be:
```js
[
  1,     // after the <div>Foo</div> block
  0,     // the interpolation block
  "baz", // the array specifies keys
  1      // after the Baz text block
]
```

This array could then be used much like an XPath for testing automation purposes, or hashed into a unique ID or prefix, for purposes like mine.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
This has never been possible.

nmain:
The specific use case you've outlined here could be solved with a different `useUniqueId` implementation:

```
let currentId = 0;

export function useUniqueId() {
[...]

PR #17322, opened 2019-11-08
Add useOpaqueIdentifier Hook

lunaruan:
We currently use unique IDs in a lot of places. Examples are:
	* `<label for="ID">`
	* `aria-labelledby`

This can cause some issues:
	1. If we server side render and then hydrate, this could cause an hydration ID mismatch
	2. If we server side render one part of the page and client side render another part of the page, the ID for one part could be different than the ID for another part even though they are supposed to be the same
	3. If we conditionally render something with an ID ,  this might also cause an ID mismatch because the ID will be different on other parts of the page

This PR creates a new hook `useUniqueId` that generates a different unique ID based on whether the hook was called on the server or client. If the hook is called during hydration, it generates an opaque object that will rerender the hook so that the IDs match.

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -27,7 +27,7 @@ import {NoMode} from 'react-reconciler/src/ReactTypeOfMode';
 
 import ErrorStackParser from 'error-stack-parser';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
-import {REACT_OPAQUE_OBJECT_TYPE} from 'shared/ReactSymbols';
+import {REACT_OPAQUE_ID_TYPE} from 'shared/ReactSymbols';
 import {
   FunctionComponent,
   SimpleMemoComponent,
@@ -331,7 +331,7 @@ function useOpaqueIdentifier(): OpaqueIDType | void {
     nextHook(); // Effect
   }
   let value = hook === null ? undefined : hook.memoizedState;
-  if (value && value.$$typeof === REACT_OPAQUE_OBJECT_TYPE) {
+  if (value && value.$$typeof === REACT_OPAQUE_ID_TYPE) {
     value = undefined;
   }
   hookLog.push({
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -18,12 +18,16 @@ import type {
   ReactScopeMethods,
 } from 'shared/ReactTypes';
 import type {Fiber} from 'react-reconciler/src/ReactFiber';
+import type {OpaqueIDType} from 'react-reconciler/src/ReactFiberHostConfig';
+
 import type {Hook, TimeoutConfig} from 'react-reconciler/src/ReactFiberHooks';
 import type {Dispatcher as DispatcherType} from 'react-reconciler/src/ReactFiberHooks';
 import type {SuspenseConfig} from 'react-reconciler/src/ReactFiberSuspenseConfig';
+import {NoMode} from 'react-reconciler/src/ReactTypeOfMode';
 
 import ErrorStackParser from 'error-stack-parser';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
+import {REACT_OPAQUE_OBJECT_TYPE} from 'shared/ReactSymbols';
 import {
   FunctionComponent,
   SimpleMemoComponent,
@@ -60,6 +64,8 @@ type BasicStateAction<S> = (S => S) | S;
 type Dispatch<A> = A => void;
 
 let primitiveStackCache: null | Map<string, Array<any>> = null;
+
+let currentFiber: Fiber | null = null;
 
 function getPrimitiveStackCache(): Map<string, Array<any>> {
   // This initializes a cache of all primitive hooks so that the top
@@ -319,6 +325,23 @@ function useDeferredValue<T>(value: T, config: TimeoutConfig | null | void): T {
   return value;
 }
 
+function useOpaqueIdentifier(): OpaqueIDType | void {
+  const hook = nextHook(); // State
+  if (currentFiber && currentFiber.mode === NoMode) {
+    nextHook(); // Effect
+  }
+  let value = hook === null ? undefined : hook.memoizedState;
+  if (value && value.$$typeof === REACT_OPAQUE_OBJECT_TYPE) {
+    value = undefined;
+  }
+  hookLog.push({
+    primitive: 'OpaqueIdentifier',
+    stackError: new Error(),
+    value,
+  });
+  return value;
+}
+
 const Dispatcher: DispatcherType = {
   readContext,
   useCallback,
@@ -336,6 +359,7 @@ const Dispatcher: DispatcherType = {
   useMutableSource,
   useDeferredValue,
   useEvent,
+  useOpaqueIdentifier,
 };
 
 // Inspect
@@ -683,6 +707,8 @@ export function inspectHooksOfFiber(
   if (currentDispatcher == null) {
     currentDispatcher = ReactSharedInternals.ReactCurrentDispatcher;
   }
+
+  currentFiber = fiber;
 
   if (
     fiber.tag !== FunctionComponent &&
--- a/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
+++ b/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
@@ -425,7 +425,7 @@ describe('ReactHooksInspectionIntegration', () => {
 
     it('should support composite useOpaqueIdentifier hook', () => {
       function Foo(props) {
-        const id = React.useOpaqueIdentifier();
+        const id = React.unstable_useOpaqueIdentifier();
         const [state] = React.useState(() => 'hello', []);
         return <div id={id}>{state}</div>;
       }
@@ -452,7 +452,7 @@ describe('ReactHooksInspectionIntegration', () => {
 
     it('should support composite useOpaqueIdentifier hook in concurrent mode', () => {
       function Foo(props) {
-        const id = React.useOpaqueIdentifier();
+        const id = React.unstable_useOpaqueIdentifier();
         const [state] = React.useState(() => 'hello', []);
         return <div id={id}>{state}</div>;
       }
--- a/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
+++ b/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
@@ -422,6 +422,66 @@ describe('ReactHooksInspectionIntegration', () => {
         },
       ]);
     });
+
+    it('should support composite useOpaqueIdentifier hook', () => {
+      function Foo(props) {
+        const id = React.useOpaqueIdentifier();
+        const [state] = React.useState(() => 'hello', []);
+        return <div id={id}>{state}</div>;
+      }
+
+      const renderer = ReactTestRenderer.create(<Foo />);
+      const childFiber = renderer.root.findByType(Foo)._currentFiber();
+      const tree = ReactDebugTools.inspectHooksOfFiber(childFiber);
+
+      expect(tree.length).toEqual(2);
+
+      expect(tree[0].id).toEqual(0);
+      expect(tree[0].isStateEditable).toEqual(false);
+      expect(tree[0].name).toEqual('OpaqueIdentifier');
+      expect(typeof tree[0].value).toBe('string');
+      expect(tree[0].value.startsWith('c_')).toBe(true);
+
+      expect(tree[1]).toEqual({
+        id: 1,
+        isStateEditable: true,
+        name: 'State',
+        value: 'hello',
+        subHooks: [],
+      });
+    });
+
+    it('should support composite useOpaqueIdentifier hook in concurrent mode', () => {
+      function Foo(props) {
+        const id = React.useOpaqueIdentifier();
+        const [state] = React.useState(() => 'hello', []);
+        return <div id={id}>{state}</div>;
+      }
+
+      const renderer = ReactTestRenderer.create(<Foo />, {
+        unstable_isConcurrent: true,
+      });
+      expect(Scheduler).toFlushWithoutYielding();
+
+      const childFiber = renderer.root.findByType(Foo)._currentFiber();
+      const tree = ReactDebugTools.inspectHooksOfFiber(childFiber);
+
+      expect(tree.length).toEqual(2);
+
+      expect(tree[0].id).toEqual(0);
+      expect(tree[0].isStateEditable).toEqual(false);
+      expect(tree[0].name).toEqual('OpaqueIdentifier');
+      expect(typeof tree[0].value).toBe('string');
+      expect(tree[0].value.startsWith('c_')).toBe(true);
+
+      expect(tree[1]).toEqual({
+        id: 1,
+        isStateEditable: true,
+        name: 'State',
+        value: 'hello',
+        subHooks: [],
+      });
+    });
   }
 
   describe('useDebugValue', () => {
--- a/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
+++ b/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
@@ -439,8 +439,7 @@ describe('ReactHooksInspectionIntegration', () => {
       expect(tree[0].id).toEqual(0);
       expect(tree[0].isStateEditable).toEqual(false);
       expect(tree[0].name).toEqual('OpaqueIdentifier');
-      expect(typeof tree[0].value).toBe('string');
-      expect(tree[0].value.startsWith('c_')).toBe(true);
+      expect((tree[0].value + '').startsWith('c_')).toBe(true);
 
       expect(tree[1]).toEqual({
         id: 1,
@@ -471,8 +470,7 @@ describe('ReactHooksInspectionIntegration', () => {
       expect(tree[0].id).toEqual(0);
       expect(tree[0].isStateEditable).toEqual(false);
       expect(tree[0].name).toEqual('OpaqueIdentifier');
-      expect(typeof tree[0].value).toBe('string');
-      expect(tree[0].value.startsWith('c_')).toBe(true);
+      expect((tree[0].value + '').startsWith('c_')).toBe(true);
 
       expect(tree[1]).toEqual({
         id: 1,
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
@@ -1643,8 +1643,6 @@ describe('ReactDOMServerHooks', () => {
         ).toErrorDev([
           'The object passed back from useOpaqueIdentifier is meant to be passed through to attributes only. ' +
             'Do not read the value directly.',
-          'The object passed back from useOpaqueIdentifier is meant to be passed through to attributes only. ' +
-            'Do not read the value directly.',
         ]);
       });
     });
@@ -1677,8 +1675,6 @@ describe('ReactDOMServerHooks', () => {
             'Do not read the value directly.',
         ),
       ).toErrorDev([
-        'The object passed back from useOpaqueIdentifier is meant to be passed through to attributes only. ' +
-          'Do not read the value directly.',
         'The object passed back from useOpaqueIdentifier is meant to be passed through to attributes only. ' +
           'Do not read the value directly.',
       ]);
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
@@ -18,6 +18,7 @@ let ReactFeatureFlags;
 let ReactDOM;
 let ReactDOMServer;
 let ReactTestUtils;
+let Scheduler;
 let useState;
 let useReducer;
 let useEffect;
@@ -28,6 +29,7 @@ let useRef;
 let useImperativeHandle;
 let useLayoutEffect;
 let useDebugValue;
+let useOpaqueIdentifier;
 let forwardRef;
 let yieldedValues;
 let yieldValue;
@@ -39,10 +41,12 @@ function initModules() {
 
   ReactFeatureFlags = require('shared/ReactFeatureFlags');
   ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
+  ReactFeatureFlags.flushSuspenseFallbacksInTests = false;
   React = require('react');
   ReactDOM = require('react-dom');
   ReactDOMServer = require('react-dom/server');
   ReactTestUtils = require('react-dom/test-utils');
+  Scheduler = require('scheduler');
   useState = React.useState;
   useReducer = React.useReducer;
   useEffect = React.useEffect;
@@ -53,6 +57,7 @@ function initModules() {
   useDebugValue = React.useDebugValue;
   useImperativeHandle = React.useImperativeHandle;
   useLayoutEffect = React.useLayoutEffect;
+  useOpaqueIdentifier = React.useOpaqueIdentifier;
   forwardRef = React.forwardRef;
 
   yieldedValues = [];
@@ -78,6 +83,9 @@ const {
   itRenders,
   itThrowsWhenRendering,
   serverRender,
+  streamRender,
+  clientCleanRender,
+  clientRenderOnServerString,
 } = ReactDOMServerIntegrationUtils(initModules);
 
 describe('ReactDOMServerHooks', () => {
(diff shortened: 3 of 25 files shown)


packages/react-dom/src/client/DOMPropertyOperations.js
```diff
@@ -110,7 +110,7 @@ export function getValueForAttribute(
       return expected === undefined ? undefined : null;
     }
     const value = node.getAttribute(name);
-    if (value === '' + (expected: any)) {
+    if (value === toString(expected)) {
```
acdlite: Why this change?

packages/react-dom/src/client/setAttribute.js
```diff
@@ -18,7 +19,11 @@ export function setAttribute(
   attributeName: string,
   attributeValue: string | TrustedValue,
 ) {
-  node.setAttribute(attributeName, (attributeValue: any));
+  if (attributeValue.$$typeof === REACT_OPAQUE_OBJECT_TYPE) {
```
acdlite: This works for updates, but need to handle similar case in `setInitialDOMProperties`, which is called during initial render.

packages/react-dom/src/client/DOMPropertyOperations.js
lunaruan: if `expected` is the opaque object, this would throw the string error. `toString` doesn't convert the opaque object, which makes this just return `value`, which is what it should be anyway.

packages/react-dom/src/client/setAttribute.js
lunaruan: I think `node.setAttribute` is called during `setInitialDOMProperties` as well to set attributes:

`setInitialDOMProperties` -> `setValueForProperty` -> `setAttribute`/`setAttributeNs`

packages/react-dom/src/__tests__/ReactDOMHooks-test.js
```diff
@@ -105,6 +109,145 @@ describe('ReactDOMHooks', () => {
...
+            <div id={id}>
+              <Child text="Child Two" />
+            </div>
+          </React.Suspense>
+        </div>
+      );
+    }
+
+    container.innerHTML =
+      '<div data-reactroot=""><div id="s_0">Child One</div><!--$!--><div>Fallback</div><!--/$--></div>';
```
acdlite: You should also assert that React didn't replace the original DOM node. I would grab a reference to the divs here, then after you hydrate, make sure they match the hydrated ones (you can use a ref for that part, or grab them out of the container again). Then do the same after the update.

packages/react-dom/src/__tests__/ReactDOMHooks-test.js
```diff
@@ -105,6 +109,145 @@ describe('ReactDOMHooks', () => {
...
+    );
+
+    suspend = false;
+    resolve();
+    await promise;
+
+    Scheduler.unstable_flushAll();
+    jest.runAllTimers();
+
+    expect(container.innerHTML).toMatchInlineSnapshot(
```
acdlite: The output is correct, but let's also test that there was only a single commit. You can do that with `useEffect`:

```js
const id = useUniqueID();
useEffect(() => {
  Scheduler.unstable_yieldValue('Did commit');
});
```

I expect this test will fail; let's write the test first and then I can work with you on how to fix it.

diegohaz:
I’d expect that calling it multiple times in the same component would produce the same ID so you would be able to use it in different custom hooks and they would point to the same ID. But it seems that calling `useUniqueID` multiple times in the same component will generate multiple IDs. What’s special about this that can’t be done outside React?

packages/react-dom/src/__tests__/ReactDOMHooks-test.js
```diff
@@ -105,6 +109,172 @@ describe('ReactDOMHooks', () => {
...
+        </div>
+      );
+    }
+
+    const root = ReactDOM.createRoot(container);
+    await act(async () => {
+      root.render(<App />);
+    });
+
+    expect(container.innerHTML).toMatchInlineSnapshot(
```
SimenB: if you do `expect(container).toMatchInlineSnapshot` `pretty-format` will format the inner html as html (newlines and less escaping, closer to how you'd write html manually).

If you explicitly want `innerHTML`, you can do `expect(wrap(container.innerHTML)).toMatchInlineSnapshot` to at least avoid the inline escaping and wrapping quotes. `wrap` comes from `jest-snapshot-serializer-raw` and is already used in `react-refresh`'s tests

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
```diff
@@ -843,4 +853,393 @@ describe('ReactDOMServerHooks', () => {
...
+            aria-labelledby="c_1"
+          />
+          <span
+            id="c_1"
+          />
+        </div>
+      `);
+    });
+
+    it('generates unique ids for client render on good server markup', async () => {
```
necolas: The snapshot below contains the server ids. Is this testing for preservation of ids from the server?

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
```diff
@@ -843,4 +853,393 @@ describe('ReactDOMServerHooks', () => {
...
+        const root = ReactDOM.createRoot(container, {hydrate: true});
+        await ReactTestUtils.act(async () => {
+          root.render(<App />);
+        });
+
+        expect(Scheduler).toHaveYielded(['Child did commit', 'Did commit']);
+        expect(Scheduler).toFlushAndYield([]);
+        jest.runAllTimers();
+
+        expect(divNode).toEqual(ref.current);
```
necolas: Do we still need this assertion is we have the inline snapshot below?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1218,6 +1227,52 @@ function updateTransition(
...
+        toString() {
+          throw Error(
+            'The object passed back from useUniqueID is meant to be passed through to ' +
+              'attributes only. Do not directly modify the output.',
+          );
+        },
+      };
+    } else {
+      return 'c_' + (clientId++).toString();
+    }
```
necolas: The body of this function looks identical to the mount function. Should we try to share it across mount and update?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1218,6 +1227,52 @@ function updateTransition(
...
 
+let clientId: number = 0;
+
+function mountUniqueId(): string | IdObject {
+  const [id, setId] = mountState(() => {
+    if (getIsHydrating()) {
+      return {
+        $$typeof: REACT_OPAQUE_OBJECT_TYPE,
+        _setId: () => {
+          setId(() => 'c_' + (clientId++).toString());
```
Jessidhia: This has the same problem as most of the "userspace" solutions: this ID is unstable and depends on commit order, especially in combination with Suspense and partial hydration.

One thing library authors often asked for is a way of getting the "tree position" of the element, which is better and relatively stable, but still not perfect; it'll probably work better for hydration and will be globally unique, but might get reused if a component is swapped by another for example.

eps1lon: I think with this solution it doesn't matter whether the counter changes between server and client since react is able to patch it up properly during hydration which wasn't possible before.

This hook is exclusively for IDREF attributes. It's not a generic, unique ID of your component instance.

Jessidhia: Ah, so the point is it not being seen as a hydration error but just as something to be patched 🤔

eps1lon: At least this is how I understood it. There's an argument to be made about hydration warnings for attributes in the first place (e.g. `<button disabled={!process.browser} />` but IDREFs seem like a safe place to start.

sebmarkbage: It’s not as simple as just patching. Partially because we don’t actually compare any props in hydration. But mostly because if you do it the naive way you may patch up one side but not the other side when partial hydration is being used.

This enforces that we force hydration of the things that needs patching on both sides.

We should have a test for that.

sebmarkbage:
How about `useOpaqueReference` or even `useOpaqueRef`?

It’s not necessarily an ID in its implementation. It’s just some way of referencing one node in two places.

The connection to `useRef` is a bit confusing but technically correct. It’s a reference to a mutable imperative box.

I’ve thought about having this in React Native as a way to pass a native box around. So that I can pass a reference to a native View instance to native APIs.

Unlike useRef, this kind of ref‘a current value would only be accessible from native code.

That’s very similar to how you can think about this use case too.

It’s a bit strange since you might expect to pass the target side like `<div ref={opaqueRef} />` instead of `<div id={opaqueRef} />`. Maybe we should even make that work. Requires a bit more code and special logic to do that though. It also means that the value can’t literally be a string since it would conflict with string refs.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1218,6 +1227,52 @@ function updateTransition(
...
+        },
+        toString() {
+          throw Error(
+            'The object passed back from useUniqueID is meant to be passed through to ' +
+              'attributes only. Do not directly modify the output.',
```
sebmarkbage: What does “don’t modify the output” mean? I’m just reading the value. Not modifying it.

packages/react-dom/src/client/setAttribute.js
```diff
@@ -16,9 +18,17 @@ import type {TrustedValue} from './ToStringValue';
 export function setAttribute(
   node: Element,
   attributeName: string,
-  attributeValue: string | TrustedValue,
+  attributeValue: string | TrustedValue | IdObject,
 ) {
-  node.setAttribute(attributeName, (attributeValue: any));
+  if (
+    typeof attributeValue === 'object' &&
```
sebmarkbage: This is probably React’s hottest path that we’re making slower (and not inlinable). Is this the best place we can do this? Where does TrustedValue do its check?

packages/react-reconciler/src/ReactFiberHooks.js
sebmarkbage: The update shouldn’t need to create the initializer (normally). Can we short cut this and avoid creating the unnecessary closure and array?

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: This should be an `invariant` since our error minification pipeline doesn't support error constructors, yet.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
sebmarkbage: Yea. I agree. Snapshot tests aren't a good way of testing these things. Especially when there are so many of them.

We don't provide any guarantees that this is actually the format we will use. In fact, these IDs are way too likely to conflict with a manual user provided ID. So we probably should change the format and we might have to change it  again.

A change in the format will yield a lot of manual work to verify if it's correct or not.

The tests also don't actually test the important properties that are useful to preserve regardless of which format we use. I.e. that a client generated ID can't overlap with a server generated ID. Does any of your tests fail if you just change it to be `react_` prefixed on both server and client? Or what if you accidentally make remove the auto-incrementing, does it fail any tests?

Instead of asserting against a snapshot, you could assert on the DOM that the IDs match. E.g. find the divs and spans and assert that `node.getAttribute('aria-labelledby')` matches `node.id` and that they don't match the other's IDs. That's all that really matter.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -260,6 +264,17 @@ function useDeferredValue<T>(value: T, config: TimeoutConfig | null | void): T {
   return value;
 }
 
+function useOpaqueReference(): string | IdObject {
+  const hook = nextHook();
+  const value = hook === null ? 'FAKE_ID' : hook.memoizedState;
```
sebmarkbage: The name  "FAKE_ID" doesn't say much and is kind of confusing if you do end up seeing it. What's Fake about it? Can I search for it? It also doesn't apply as a string when in React Native for example. Let's just leave it as `undefined`  if it's not there.

I believe this can also cause an error if you inspect a hydrated component since `toString` and `valueOf` will throw. We should probably detect if it's a `REACT_OPAQUE_OBJECT_TYPE` and do something special. Perhaps just leaving it as `undefined` as well.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
```diff
@@ -843,4 +853,384 @@ describe('ReactDOMServerHooks', () => {
...
+                </div>
+              </React.Suspense>
+            </div>
+          );
+        }
+
+        const container = document.createElement('div');
+
+        container.innerHTML =
+          '<div data-reactroot=""><div id="s_0">Child One</div><!--$!--><div>Fallback</div><!--/$--></div>';
```
sebmarkbage: I'd use `ReactDOMServer.renderToString(...)` to render this string so that you don't have to hard code the format.

Example: https://github.com/facebook/react/blob/9fe1031244903e442de179821f1d383a9f2a59f2/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js#L1049

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -60,7 +86,8 @@ if (enableTrustedTypesIntegration && typeof trustedTypes !== 'undefined') {
         trustedTypes.isScript(value) ||
         trustedTypes.isScriptURL(value) ||
         /* TrustedURLs are deprecated and will be removed soon: https://github.com/WICG/trusted-types/pull/204 */
-        (trustedTypes.isURL && trustedTypes.isURL(value)))
+        (trustedTypes.isURL && trustedTypes.isURL(value)) ||
+        value.$$typeof === REACT_OPAQUE_OBJECT_TYPE)
```
sebmarkbage: This doesn't seem right. This will be toString:ed by the DOM and throw an error anyway. No need to add a special case, is there? Just let it fail below.

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -44,14 +45,39 @@ export opaque type TrustedValue: {toString(): string, valueOf(): string} = {
...
 };
 
+export function evaluateToStringOrTrustedType(
+  value: mixed,
+): string | TrustedValue {
+  if (
+    value !== null &&
+    typeof value === 'object' &&
+    value.$$typeof === REACT_OPAQUE_OBJECT_TYPE &&
+    typeof value._setId === 'function'
```
sebmarkbage: You shouldn't need this extra check. You've asserted the type. Just cast it through `let obj: IdObject = (value: any)`.

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -44,14 +45,39 @@ export opaque type TrustedValue: {toString(): string, valueOf(): string} = {
...
+  ) {
+    value._setId();
+    return '';
+  } else {
+    return toStringOrTrustedType(value);
```
sebmarkbage: The point to collocate these was so that you could use the same type check to quickly determine if this is *not* an object which is the common case. However, by calling this second function, you have to check the type again in the other function. Leading to two checks instead of one, for the common case.

You should inline this here so that you can avoid extra type checks. However, the bigger question is why you need two functions at all.

packages/react-dom/src/client/ToStringValue.js
```diff
@@ -44,14 +45,39 @@ export opaque type TrustedValue: {toString(): string, valueOf(): string} = {
   valueOf(): string,
 };
 
+export function evaluateToStringOrTrustedType(
```
sebmarkbage: Does this need to be its own function? Can't it just be the same as `toStringOrTrustedType`?

`toStringOrTrustedType` is called in two other places.

This one seems like you might have missed: https://github.com/facebook/react/pull/17322/files#diff-214d5116dd8b5a7d184cdf5b2160f94cR183

Seems like it should also trigger this path.

The other one is here but I think that's actually already the wrong place to do this and it shouldn't toString this at all since the initial rendering doesn't: https://github.com/facebook/react/blob/9fe1031244903e442de179821f1d383a9f2a59f2/packages/react-dom/src/client/ReactDOMComponent.js#L792
But it doesn't hurt that we check for `REACT_OPAQUE_OBJECT_TYPE`. Even though the existing call is a bug in the first place.

That way we need less code overall.

packages/react-dom/src/client/DOMPropertyOperations.js
```diff
@@ -110,7 +114,7 @@ export function getValueForAttribute(
       return expected === undefined ? undefined : null;
     }
     const value = node.getAttribute(name);
-    if (value === '' + (expected: any)) {
+    if (value === toStringOrTrustedType(getToStringValue(expected))) {
```
sebmarkbage: This call site seems like it breaks with trusted types already.

But this fix doesn't seem right. Why call `getToStringValue`? Why would you need to ignore `function` values or `symbol`? `getAttribute` should never return those types.

`toStringOrTrustedType` also doesn't seem like the right fix neither.

For trusted types, the server generated string won't match the object on the client, so this wouldn't end up matching. Causing a hydration error. This will need to dig into the trusted type and inspect the value to see if it's equivalent.

For `REACT_OPAQUE_OBJECT_TYPE` the returned object will never match. So that will also be a hydration error.

It seems to me that this needs to just perform some more custom matching logic instead of trying to use the `toStringOrTrustedType` helper.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -197,6 +198,9 @@ if (__DEV__) {
     if (didWarnInvalidHydration) {
       return;
     }
+    if (clientValue && clientValue.$$typeof === REACT_OPAQUE_OBJECT_TYPE) {
```
sebmarkbage: This seems like the wrong place to do this check. If we get here it's because we already thought that there was an error. If you do the check in `getValueForAttribute` instead, then we don't have to treat it as an error in the first place.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1221,6 +1230,42 @@ function updateTransition(
   return [start, isPending];
 }
 
+let clientId: number = 0;
+
+function mountOpaqueReference(): string | IdObject {
+  const [id, setId] = mountState(() => {
+    if (getIsHydrating()) {
+      return {
+        $$typeof: REACT_OPAQUE_OBJECT_TYPE,
```
sebmarkbage: Other than its creation, this object is only ever referenced in `ReactDOM`. That's a good thing because on other platforms like the React Native use case, this object will look very different and have a very different signature.

To do that, though, we need to make the creation of this object something platform specific. Let's make a HostConfig method to create this object and only implement it in ReactDOM.

For React Native, we should make that method throw since this feature is not yet supported there. Later we could add some specific implementation.

You probably need to pass `setId` in as an argument to the `HostConfig` method.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1221,6 +1230,42 @@ function updateTransition(
...
+        valueOf() {
+          invariant(
+            false,
+            'The object passed back from useOpaqueReference is meant to be passed through to ' +
+              'attributes only. Do not read the value directly.',
+          );
+        },
+      };
+    } else {
+      return 'c_' + (clientId++).toString();
```
sebmarkbage: This should also move to `ReactDOM` since it's specific to how an "Opaque Reference" is constructed in the DOM.

Also, don't call `toString()`. Rely on `+ ''` casting instead like we do elsewhere. `.toString()` becomes a dynamic look up which can be monkey patched by `Number.prototype`. It's more bytes and less efficient.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js
```diff
@@ -843,4 +853,384 @@ describe('ReactDOMServerHooks', () => {
...
+            </div>
+            <div
+              id="c_0"
+            >
+              Child Two
+            </div>
+          </div>
+        </div>
+      `);
+    });
```
sebmarkbage: We should add a similar test where Child One is in a Suspense boundary that is still suspended. I.e. if we're only partially hydrated. Then we do an update that shows the second Child. This should suspend and wait for the suspense boundary to resolve before letting the update through.

It should not leave them inconsistent.

That test would ensure that we don't forget about this case and try to implement the "patching strategy approach" in the future.

https://github.com/facebook/react/pull/17322/files#r354445736

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -473,6 +474,12 @@ function useTransition(
   return [startTransition, false];
 }
 
+let serverId: number = 0;
+
+function useOpaqueReference(): string | IdObject {
+  return 's_' + (serverId++).toString();
```
sebmarkbage: Let's use `toString(36)` here. It'll save us a couple of bytes when we have many IDs.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1221,6 +1230,42 @@ function updateTransition(
...
 }
 
+let clientId: number = 0;
+
+function mountOpaqueReference(): string | IdObject {
+  const [id, setId] = mountState(() => {
+    if (getIsHydrating()) {
+      return {
+        $$typeof: REACT_OPAQUE_OBJECT_TYPE,
+        _setId: () => {
```
sebmarkbage: It is possible for this function to be called more than once if the object is referenced multiple times. This would eventually get rendered but unnecessarily queue up many updates and waste IDs that won't get used. We might want to dedupe so that if it's called more than once, we don't generate a new id each time.

sebmarkbage:
If we land https://github.com/facebook/react/pull/17774 then this will need to be rebased on it.

It removes any brand checks in React. This makes it a bit awkward to add just for this case.

An alternative approach could be that toString/valueOf calls setId when accessed and returning empty string. That way it happens automatically by the DOM calling these methods. Then we can add a DEV error if these are called in a different context than React calling them. E.g. we can detect this object in DEV only and do the special case but then that way there's no special brand checking in prod.

mike-marcacci:
I was just made aware of this PR which addresses the same problem I identified in #15435. However, my approach is a bit different in that I propose that an ID should be based on the component's path in react tree. My reasoning is that this is the only thing that is stable across the server and client and not dependent on rendering order/etc.

I would be interested to hear any thoughts on why the approach in this PR might be superior to my proposed strategy.

sebmarkbage:
I originally suggested that this be called `useOpaqueReference` because it's effectively a "native ref". In React Native, this could be implemented as a native box just like refs to contain the currently referenced native view or value. It's more like a ref than an ID since an ID is just an inefficient way to implement that. The longer name "Reference" was meant to differentiate it a bit from "Ref". It was also to discourage trying to use it as a string somewhere.

However, I think that was a little too clever. The flaw in naming is that it would seem like you could pass it into `ref={...}` instead of `id={...}`. That was actually my thinking that you'd pass it to `ref={...}` on a native view in RN. There are some practical reasons that maybe we shouldn't do that initially though.

So calling it "useOpaqueIdentifier" might be better for now.

I'm a bit concerned that it's too easy to accidentally use this for the wrong purpose (e.g. generate session IDs or seeds) if you only use this API client-side.

I wonder if we should let DEV mode sometimes (randomly?) return a wrapper object with a toString that warns unless it's React that uses it. For client rendering. So that you can't accidentally rely on it being a string when developing.

gaearon:
Let's not forget to add a DevTools regression test like in https://github.com/facebook/react/pull/18130.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -850,6 +852,12 @@ export function diffProperties(
         // to update this element.
         updatePayload = [];
       }
+    } else if (
+      typeof nextProp === 'object' &&
+      nextProp !== null &&
+      nextProp.$$typeof === REACT_OPAQUE_OBJECT_TYPE
+    ) {
+      throw nextProp.toString();
```
bvaughn: I don't think I understand why diff props throws here.

I guess we only hit this case if the values differed? Which is never expected?

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1233,6 +1241,12 @@ export function warnForUnmatchedText(textNode: Text, text: string) {
   }
 }
 
+export function getIsRenderingInDEV() {
```
bvaughn: What value does this function add? Looks like it's only called from 4 places, that themselves are already inside of a `__DEV__` conditional. Seems like we could just remove this indirection? (Maybe it doesn't matter since it would get inlined anyway but..)

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1524,6 +1535,46 @@ function rerenderTransition(
...
+  let hasSetId = false;
+
+  const [id, setId] = mountState(() => {
+    if (getIsHydrating()) {
+      return makeOpaqueHydratingObject(() => {
+        if (!hasSetId) {
+          setId(() => makeClientId());
+          hasSetId = true;
+        }
+      }, getComponentName(currentlyRenderingFiber.type));
```
bvaughn: Looks like we're *always* stringifying the component name, even though we *only* use the name for DEV warnings. Seems like we should only call `getComponnetName` in DEV mode, and probably only if we actually need to show a warning?

(So maybe we'd just pass the Fiber or the type?)

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1524,6 +1535,46 @@ function rerenderTransition(
...
+    } else {
+      return makeClientId();
+    }
+  });
+
+  if (currentlyRenderingFiber.mode === NoMode) {
```
bvaughn: I don't understand this conditional.

bvaughn:
😄

packages/react-dom/src/client/ReactDOMComponent.js
gaearon: Btw even if we intend to throw, we must never throw strings. They won't get proper stack traces in reporting.

packages/react-reconciler/src/ReactFiberHooks.js
lunaruan: AFAIK NoMode refers to Legacy mode. So if it's legacy mode, we don't want the opaque object to throw, so we use an effect to set the ID instead.

packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
```diff
@@ -422,6 +422,90 @@ describe('ReactHooksInspectionIntegration', () => {
...
+      expect(Scheduler).toFlushWithoutYielding();
+
+      let childFiber = renderer.root.findByType(Foo)._currentFiber();
+      let tree = ReactDebugTools.inspectHooksOfFiber(childFiber);
+      expect(tree).toEqual([
+        {
+          id: 0,
+          isStateEditable: false,
+          name: 'OpaqueIdentifier',
+          value: 'c_1',
```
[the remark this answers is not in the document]
lunaruan: We actually shouldn't care what this value is. I'll rewrite this test to reflect that as long as this value is a string it should be okay.

packages/react-dom/src/client/ReactDOMComponent.js
lunaruan: This throws here so that in concurrent mode we can rerender. We don't need to throw this, however, because nextProp.toString() itself should throw, so I'll remove the `throw`

packages/react-reconciler/src/ReactFiberHooks.js
gaearon: This assumes we won’t add other modes that are additive to Legacy. In fact I think we already have one — StrictMode. That’s not NoMode but it *is* legacy when on its own. 

I think what you want to check is `(mode & BlockingMode) == NoMode`. That says “doesn’t have BlockingMode bit”. That bit is present in all non-legacy modes so it’s absence means it’s legacy.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -849,6 +850,12 @@ export function diffProperties(
         // to update this element.
         updatePayload = [];
       }
+    } else if (
+      typeof nextProp === 'object' &&
+      nextProp !== null &&
+      nextProp.$$typeof === REACT_OPAQUE_OBJECT_TYPE
+    ) {
+      nextProp.toString();
```
gaearon: What is this call for? Trigger a warning? Or I thought that warning only fires during render?

lunaruan: No, the object only shows up during hydration. this function will set state and then throw an error, which causes React to synchronously retry the render, updating the object from a hydration object to something that resembles a string.

bvaughn: May benefit frmo an inline comment explaining the significance.

packages/shared/ReactSymbols.js
```diff
@@ -62,6 +62,12 @@ export const REACT_RESPONDER_TYPE = hasSymbol
   ? Symbol.for('react.responder')
   : 0xead6;
 export const REACT_SCOPE_TYPE = hasSymbol ? Symbol.for('react.scope') : 0xead7;
+export const REACT_OPAQUE_OBJECT_TYPE = hasSymbol
+  ? Symbol.for('react.opaqueObject')
```
sebmarkbage: Let's rename these to `'react.opaque.identifier'`.

packages/shared/ReactSymbols.js
```diff
@@ -62,6 +62,12 @@ export const REACT_RESPONDER_TYPE = hasSymbol
   ? Symbol.for('react.responder')
   : 0xead6;
 export const REACT_SCOPE_TYPE = hasSymbol ? Symbol.for('react.scope') : 0xead7;
+export const REACT_OPAQUE_OBJECT_TYPE = hasSymbol
```
sebmarkbage: And this to `REACT_OPAQUE_ID_TYPE`

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1143,6 +1161,108 @@ export function getInstanceFromNode(node: HTMLElement): null | Object {
   return getClosestInstanceFromNode(node) || null;
 }
 
+let clientId: number = 0;
+export function makeClientId(): OpaqueIDType {
+  const id = 'r:' + (clientId++).toString(36);
```
sebmarkbage: Maybe we should just return this string directly in PROD for perf.

packages/react-test-renderer/src/ReactTestHostConfig.js
```diff
@@ -380,6 +387,67 @@ export function beforeRemoveInstance(instance: any) {
...
+}
+
+export function getIsUpdatingOpaqueValueInRenderPhase() {
+  if (__DEV__) {
+    return isUpdatingOpaqueValueInRenderPhase;
+  }
+}
+
+function setIsUpdatingOpaqueValueInRenderPhase(isUpdating) {
+  isUpdatingOpaqueValueInRenderPhase = isUpdating;
```
sebmarkbage: This should be wrapped in `__DEV__`, right?

sebmarkbage:
One use case that we considered for this Hook is the notion of "shared element transitions" on React Native, and now in [Framer Motion](https://twitter.com/mattgperry/status/1243540350427332609).

I think in general, this use case isn't always very ergonomic because when you do shared element transitions it's often because the abstractions are fairly decoupled. It's a pain to thread some generated ID from a parent. The global string approach is quite convenient.

However, a global string suffers from the problem that you might render multiple instances of the same component on the same page. This is more likely on larger screens (or VR) where you fit more content.

In these cases the use cases are probably more scoped to a smaller part of the page and so funneling an auto-generated ID through wouldn't be so bad.

Of course, the approach used in for example Framer Motion doesn't work with Partial Hydration because the "from" might not be hydrated yet so it wouldn't know what to animate from. There's also nothing to force hydration of that path.

It seems like `useOpaqueIdentifier` might actually be useful for that use case too. However, it's a little different. For one, there's no ID in the actual HTML. The only purpose it would fill for that use case is to force hydration of any path that might read the ID.

The performance tradeoffs are also a bit different. We expect this to be rare for the aria/label case because both sides will exist from the beginning in almost all cases. For shared element transitions it's expected that only one side exists from the beginning and that eventually rendering the other side will be a new mount. So this will kick in the extra renders more frequently.

cc @InventingWithMonster

InventingWithMonster:
For context, as the API stands we do match components using a global string passed to the `magicId` prop.

> I think in general, this use case isn't always very ergonomic because when you do shared element transitions it's often because the abstractions are fairly decoupled. It's a pain to thread some generated ID from a parent. The global string approach is quite convenient.
>
> However, a global string suffers from the problem that you might render multiple instances of the same component on the same page. This is more likely on larger screens (or VR) where you fit more content.

We actually do render multiple instances of the same component on the same page. For example [clicking the red box in this demo](https://codesandbox.io/s/framer-motion-magic-motion-gallery-demo-pknk7) - the previous component is left in its original place to maintain layout. Likewise there's a crossfade mode where both elements are used in the shared element transition.

Because of this, I don't think I'd make shared element implementation that relied on `id` itself as I think it's conceptually different than a `visualId` (or `magicId` in our case.) I can also imagine use-cases where we go one -> many or many -> one, although I haven't explicitly supported this yet. Even so, this hook would be beneficial in that use-case, and threading its output through a tree is probably a separate concern that I assume can be solved via a Context-driven API.

packages/react/index.js
```diff
@@ -72,6 +72,7 @@ export {
   SuspenseList,
   unstable_withSuspenseConfig,
   block,
+  useOpaqueIdentifier,
```
acdlite, line 75: Remove this until it's stable

sebmarkbage:
> We actually do render multiple instances of the same component on the same page. For example clicking the red box in this demo - the previous component is left in its original place to maintain layout.

@InventingWithMonster The use case I had in mind was more like if you had two of these entire grids on the same page. https://codesandbox.io/s/framer-motion-magic-motion-gallery-demo-55ssm

This doesn't seem to work.

It's not the common case but it happens in very large apps where it's hard to see these things coming.

I guess this work as long as you have a separate SharedMagicMotion for each of them. I could see this being problematic if most of the time you have a local SharedMagicMotion but then once in a while you want for example a very top level thing to share the animation with a nested one. Like a profile picture in a navigation bar to animate in and out of the page. Then you have to hoist the scope of the SharedMagicMotion which then breaks other encapsulation.

packages/react/index.js
acdlite: Never mind, this file is only used for tests. `index.stable.js` is the stable interface.

theKashey:
According to the current implementation - everything is based on a single counter, even if different for client and server.
While server is running it will endlessly increment corresponding id, actually making ids unique every render; while client would always start from the scratch. So I would not worry much for SSR+hydrate, which was a primary goal for this PR. Look like it's safe.

But there are situation in a hybrid app, for example written using different framework/language, with some injections of React code(widgets or partial migration) – it could be a situation when two different  instances of React is used on a single page, resulting the same `ids`, with base 0, to be used more than once.
The same problem could occur with micro-services approach, when server response could be composed from pieces rendered by different services  independency - their results could again contain clashing ids.

From my understanding the problem above could be solved by prefixing `ids`, and that could not be done in a user space, as long as `id` is expected to be passed down to the attribute without any modification.

What about adding a `<Randomize>` component, which will generate a random UID, or support used-defined prefix, which would be applied to any `id` used below it. Then one could use it to prefix `ids` by `c:` or `r:`, not hardcode it as currently, as well as make ids really unique for multiple-Reacts case?

gaearon:
I also wonder if for a long-running service process we might ever run out of IDs.

theKashey:
Well, all this time (without React "support") every SSRed page with "unique ids"  had to be "synchronized" with the client. In other words - `counter` always start from zero. It's also "snapshots friendly".
If there any limitations to read and store this counter on the context level as a "mutable" variable?
I personally, with the current rise of a11n, would expect the future overuse of this function, so it has to be as performant as possible... but endlessly incrementing a single counter, which is actually a double, so it will break after a `MAX_SAFE_INTEGER` interactions.
However, I just calculated,- in my particular case(1000 Ids per page  x 300 RPS) I would have to keep the server running for roughly 900 years to breach that limit.

eps1lon:
> so it will break after a MAX_SAFE_INTEGER interactions.

Technically it's MAX_SAFE_INTEGER mounts. A re-render wouldn't increment the counter: https://8s2ms.csb.app/

On a naive level we could % MAX_SAFE_INTEGER the incremented value. That would at least solve the issue if don't have extremely long living instances and short lived instances in the same tree. I think the main concern here are server renderers, right? I can't imagine a client-side application that runs for years constantly with 60 frames per second (where each frame discards the old UI) with thousands of ids in the tree.

gaearon:
Do you mind filing an issue about possible clashes between multiple Reacts?

theKashey:
@gaearon - filled 3 cases out of my head. Probably there might be more.

eps1lon:
Are there plans for supporting multiple opaque identifiers in HTML attributes i.e. support IDREFS? Since aria was mentioned in the original description it seems like there should be support for `<button aria-labelledby={[id1, id2]} />` or a similar API. This would be used to implement a [collapsible listbox](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html):

```jsx
export default function App() {
  const labelId = React.unstable_useOpaqueIdentifier();
  const buttonId = React.unstable_useOpaqueIdentifier();
  const listboxId = React.unstable_useOpaqueIdentifier();
  return (
    <React.Fragment>
      <span id={labelId}>Choose an element:</span>
      <button
        aria-haspopup="listbox"
        aria-labelledby={[labelId, buttonId]}
        id={buttonId}
      >
        Neptunium
      </button>
      <ul aria-labelledby={labelId} id={listboxId} role="listbox" />
    </React.Fragment>
  );
}
````
-- https://codesandbox.io/s/useopaqueidentifier-for-idrefs-ocnm4

This currently almost works but it concatenates the ids with `","` (default toString of arrays) instead of `" "`.

gaearon:
@eps1lon Mind filing a new issue to discuss?

Arunkumar-R-R:
This is great

PR #17323, opened 2019-11-09
[Native] Add FeatureFlag to dispatch events with instance targets

TheSavior:
In React Native we want to try making the event target and currentTarget equal to a React Instance instead of a reactTag. This is needed to clean up some of the logic in ScrollResponder, TextInput, and TextInputState. This also will let these core components become compatible with Fabric.

There is precedent here as target and currentTarget on ReactDOM are HTMLElements that have .focus and .blur methods for example.

This API is behind a FeatureFlag that we can use to run an experiment internally.

It is worth noting that this will be a breaking change to ReactNative when rolled out. When searching the Facebook codebase I surprisingly found essentially *no* callsites referencing target/currentTarget except for our core components. That gives me some confidence that this breaking change is acceptable and doesn't require a different approach like putting the instance in a different key on the event.

The purpose of this experiment is to validate that assumption.

**Note: This PR doesn't have any logic to change `currentTarget`. I haven't investigated that one yet to know how to change that. Is that reasonable to do in a separate PR or do you think it should be part of the same one?**

## Test Plan
I did a local sync and ran these changes in the Facebook repo and tested this in both Fabric and Paper with the flag turned on and off. ScrollResponder and TextInput logic didn't throw any exceptions, was correctly handling target being both a number and instance.


TheSavior:
This PR depends on https://github.com/facebook/react/pull/17344 to change the flow types to allow null as the target.

TheSavior:
`currentTarget` will be changed in this stacked PR that I will be able to send once this is merged: https://github.com/TheSavior/react/pull/1/files

PR #17327, opened 2019-11-09
Fix '__DEV__ is not defined' error in react-art fixture and change message to use 'yarn' over 'npm'

ScottMorse:
I am using yarn 1.19.1, node v10.16.3, and npm 6.13.0.

* After running `yarn install` and `yarn build` and opening index.html, I received the error `__DEV__ is not defined` from the produced bundle.js file.  My change in the webpack config simply sets `__DEV__` to `true` to get rid of this error, allowing the successful rendering of the React logo animation after running the build.

*  I changed the message in index.html to suggest running `yarn install` and `yarn build` over `npm install` and `npm run build`, for two reasons:
    * The lockfile in the fixture is `yarn.lock`
    * The `"link:"` protocol in package.json is supported by yarn, not npm, so `npm install` fails with `EUNSUPPORTEDPROTOCOL`.

### DIFF
--- a/fixtures/art/index.html
+++ b/fixtures/art/index.html
@@ -9,8 +9,8 @@
     <p>If you're seeing this message, it means you haven't generated the bundle file for this example. Try running:</p>
 
     <pre><code>
-    npm install
-    npm run build
+    yarn install
+    yarn build
     </code></pre>
 
     <p>then reload the page.</p>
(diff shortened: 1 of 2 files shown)


gaearon:
What was actually using `__DEV__`?

gaearon:
My guess is you were trying to run this fixture with source, but it only works if you `yarn build` in React parent folder first.

PR #17328, opened 2019-11-09
Warn when createRoot() is called on existing root

gaearon:
The behavior is pretty confusing: https://codesandbox.io/s/frosty-satoshi-z87yi

So this adds a warning. Will be useful to people who replace `.render()` with `.createRoot().render()` without considering that they may be running it multiple times.

I don't warn if you `.unmount()` first. That seems okay-ish. Maybe you really don't have access to the root instance but at least you can free the previous one.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMRoot-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMRoot-test.js
@@ -212,4 +212,22 @@ describe('ReactDOMRoot', () => {
       {withoutStack: true},
     );
   });
+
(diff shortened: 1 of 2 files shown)


sebmarkbage:
We attach another container flag for purposes for the event system to know if we’re in a react root.

Can we reuse that?

gaearon:
Yeah I can unify them. That would also solve the iOS issue.

gaearon:
https://github.com/facebook/react/pull/17329

PR #17329, opened 2019-11-10
Warn when doing createRoot twice on the same node (another approach)

gaearon:
Alternative to #17328. Uses an existing flag. First commit is a refactoring, second adds a new warning.


gaearon:
Huh. The failure is funny. I forgot about the case of two renderers.

gaearon:
Re: async, I could follow this pattern:

https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOM.js#L545

PR #17330, opened 2019-11-10
Align root container onclick behavior with legacy mode

gaearon:
So... this isn't really important but we're inconsistent with how we set `onclick` on container nodes.

In Legacy Mode, we only set them on Portal Containers (but not Root Containers).
In modern Modes, we set them both on Portal Containers and Root Containers.

There's no particular good reason to set them on Root Containers (Portal Containers need it for `onclick` event to bubble on Safari). So I'm changing them to behave the same way as in Legacy Mode.

I guess it would also be ok to always set it. Another way to think of it is this is removing some usage of `_reactRootContainer` where the code implies "root" because it's really only "legacy root" and this isn't obvious. For example someone may change this logic later and assume all roots have `_reactRootContainer`.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMRoot-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMRoot-test.js
@@ -231,4 +231,39 @@ describe('ReactDOMRoot', () => {
     Scheduler.unstable_flushAll();
     ReactDOM.createRoot(container); // No warning
   });
+
+  it('sets noop onclick for Mobile Safari on portal containers', () => {
+    // We set noop onclick for portals (so clicks don't get lost in iOS Safari)
+    // but there's no reason to do this for top-level containers.
+
+    // Legacy Root
+    const portalContainer1 = document.createElement('div');
+    const rootContainer1 = document.createElement('div');
+    ReactDOM.render(
+      <div>{ReactDOM.createPortal(<div />, portalContainer1)}</div>,
+      rootContainer1,
+    );
+    expect(rootContainer1.onclick).toBe(null);
+    expect(portalContainer1.onclick).not.toBe(null);
+
+    // Blocking Root
+    const portalContainer2 = document.createElement('div');
+    const rootContainer2 = document.createElement('div');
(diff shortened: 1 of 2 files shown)


sebmarkbage:
Can you explain why we need it for only Portals?

gaearon:
If something has a React click listener we put onclick on that node directly. That’s sufficient to catch events from it. Or originating from its children.

But portals break out of DOM hierarchy. So if you click on a modal’s div, three might not be anything in the DOM tree above with an onclick handler. So safari will ignore it. However, the logical React hierarchy might have a click handler above the portal. We don’t want to miss it.

We don’t need these on roots because if there was no click handler by the time we reached the root, we know there’s nothing that needs to handle it in the React tree.

But we need them in portal containers because there might still be React listeners in the logical tree even if the physical tree has no onclick above.

sebmarkbage:
So in concurrent mode there's two more places we rely on bubbling for delegation:

1) Selective hydration. Emitting `onclick=""` attributes in the SSR should cover this use case sufficiently.
2) To force flush discrete work when the event hasn't committed yet. I believe this extra listener does help in that case.

Imagine that there isn't an onclick listener and then one is added in another discrete state update: `c ? <div /> : <div onClick={...} />`

If this change hasn't flushed yet, by the time the second click comes in, then Safari doesn't know that something might need the click and doesn't dispatch any event to us. If we always attach a click then Safari does dispatch the event and we can flush the previous work which attaches the event listeners in time to be invoked.

andi1984:
> But we need them in portal containers because there might still be React listeners in the logical tree even if the physical tree has no onclick above.

I'm not knowing the internals of React but for me, this seems to be the main situation where React maintains something which differs completely with the actual DOM. From a component wrapping level, it makes total sense to me within the React ecosystem, but are you sure that you want to maintain all those edge cases the users of React might run into because of that? Or is it actually no big deal?

I would personally have no problem if it would be clear that every DOM-based interaction between <Portal> (and its children) and outer components is not supported, but I agree it's nice to have this kind of interaction.

gaearon:
meh.

Issue #15080, opened 2019-03-10
Consider exporting `batchedUpdates` from React


markerikson:
ReactDOM and React Native both currently export an `unstable_batchedUpdates` API.  Because batching is a renderer/reconciliation-level concern, this API is exported by the renderer packages, not the core `react` package.

The React team has recently encouraged the Redux team to make use of `unstable_batchedUpdates` in React-Redux.  However, this becomes complicated due to how that API is being exported.

It's possible to deal with this at the bundler level.  Some experimentation shows that creating an alternate file with a `.native.js` extension will cause that to be picked up by the RN bundler, as in this example:

```js
// batch.js
import {unstable_batchedUpdates} from "./react-dom";

// ./react-dom.js
export {unstable_batchedUpdates} from "react-dom"

// ./react-dom.native.js
export {unstable_batchedUpdates} from "react-native"
```

However, this does not handle the case where an alternative React renderer is being used.  The [list of other React renderers is continuing to grow](https://github.com/chentsulin/awesome-react-renderer), which means that a React library that needs batching would have to deal with that situation in some way.  This becomes extremely complicated when you start considering variations on bundlers, module formats, and build environments.

It would be extremely beneficial if the React core itself exported a `batchedUpdates` API.  That could default to being a noop wrapper like `(callback) => callback()` if no suitable implementation was available.

I know that `unstable_batchedUpdates()` is, uh... "unstable".  However, the React team has stated that "it's the most stable of the unstable APIs", and "half of Facebook depends on this".

I think it would really help the ecosystem if some form of this API was solidified and exported from the core React package itself.

markerikson:
I suppose an alternative solution might be for userland packages to provide an alternate entry point for use with non-mainline renderers that would pull in a fallback noop, like:

```js
import {connect, Provider, batch} from "react-redux/alternate-renderers";
```

FredyC:
This is something we have been dealing with `mobx-react` for quite some time. The current solution resides in bundling for `react-dom` and then making a copy of that bundle and do hard replace of `react-dom` to `react-native`.

It's a viable solution, but far from pretty as it depends heavily on the building process and cannot be easily shared. Among the recent issues, this is problematic with Jest which is unable to recognize the RN bundle by default and fails. People have to manually configure Jest to map to RN build.

The ideal would be to have a single bundle which imports `unstable_batchedUpdates` from a single module and be done with it. Just as you are suggesting to have it in the `react` package.

[...]

Issue #14536, opened 2019-01-05
React.Suspense provide a lifecycle so components can handle the `display:none` removal


oliviertassinari:
**Do you want to request a *feature* or report a *bug*?**

It's a feature. 

**What is the current behavior?**

React.Suspense mounts its children with a `display: none` style if a promise is thrown. Once the thrown promise is resolved, React removes the `display: none` style.

**What is the expected behavior?**

The children components have no easy way to know when the `display: none` style is removed by React. This is problematic when one child component needs to read from the DOM layout to correctly display its elements. Most people wait for the `componentDidMount` callback to trigger, but because the element is `display: none`, it can't read any value from the DOM layout.

The issue was discovered in https://github.com/mui-org/material-ui/issues/14077. I believe that React should provide a lifecycle so the children components know when they are visible, that it's safe to do layout computations. 

The best workaround I'm aware of it to use the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) but it requires a polyfill on IE 11 and Safari.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

Version: 16.7.0-alpha.2

ryancogswell:
A new lifecycle doesn't feel like the right solution to me. In hooks implementations, the logic typically involved would be in `useLayoutEffect` and it seems that it would get convoluted to need to interact with a separate lifecycle for this `Suspense` scenario (brings back the kind of bugs that people tend to introduce with the separation of `componentDidMount/Update/Unmount`).

It seems that this really needs to be treated more like state of the `Suspense` element -- especially since it can toggle any number of times if anything causes a descendant to re-render in a way that now loads a lazy-loaded component (and of course many more scenarios once `Suspense` is supported for fetching, etc.). Then the appropriate way for descendants to use this information would be for it to be provided via some sort of `Context`. Perhaps there could be something like a `useSuspenseContext` hook that would return whether or not the element is currently contained within a `Suspense` that is in its fallback state. This boolean could then be leveraged in the `useLayoutEffect` dependency array. Alternatively, something like a `useSuspenseAwareLayoutEffect` could automatically execute whenever this boolean changes and/or provide the boolean as an argument to the effect function.

ryancogswell:
Perhaps someone on the React team has already put a lot of thought into this and knows what they plan to do and I'm just wasting brain cycles on it, but I want to follow-up on my earlier thoughts after some more consideration.

[...]

PR #17331, opened 2019-11-10
Split ReactDOM entry point

gaearon:
This is kind of subjective. I find it hard to navigate `ReactDOM.js`. It's long, mixes public API with implementation details, and mixes some injection with old APIs with new ones. It's hard to tell what's being used only by the old code, and what's only needed for legacy (or missing in modern code).

This is a refactor that does no functional changes. It splits `ReactDOM.js` into:

* `ReactDOM.js`: Entry point is lean and only does some injection and importing.
* `ReactDOMRoot.js`: Exports `createRoot`, `createBlockingRoot`, and `createLegacyRoot` (internal). This is the where the shared root implementation lives.
* `ReactDOMLegacy.js`: All legacy APIs like `findDOMNode`, `render`, `hydrate`, `unmountComponentAtNode`, `unstable_renderSubtreeIntoContainer`, live here. This includes some batching and legacy hyration heuristics code that isn't used by the modern roots.

Eventually we can move `ReactDOMLegacy.js` to a separate package entry point.

I've also renamed `BatchedRoot` to `BlockingRoot` (missed that during last refactoring).

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -7,30 +7,28 @@
  * @flow
  */
 
+import type {RootType} from './ReactDOMRoot';
 import type {ReactNodeList} from 'shared/ReactTypes';
-import type {RootTag} from 'shared/ReactRootTags';
-// TODO: This type is shared between the reconciler and ReactDOM, but will
-// eventually be lifted out to the renderer.
-import type {FiberRoot} from 'react-reconciler/src/ReactFiberRoot';
 
 import '../shared/checkReact';
 import './ReactDOMClientInjection';
+import {
+  findDOMNode,
+  render,
+  hydrate,
+  unstable_renderSubtreeIntoContainer,
+  unmountComponentAtNode,
+} from './ReactDOMLegacy';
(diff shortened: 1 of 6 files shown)


trueadm:
Rather than ReactDOMLegacy, maybe another name might be more fitting? Heritage? Legacy makes the APIs feel less relevant and important.

gaearon:
>Rather than ReactDOMLegacy, maybe another name might be more fitting? Heritage? Legacy makes the APIs feel less relevant and important.

This file is internal naming and is consistent with other places we call it "legacy mode" in the codebase. So I wouldn't worry too much about that here.

Generally this might be a discussion worth having (e.g. the docs also call it a "legacy mode" [here](https://reactjs.org/docs/concurrent-mode-adoption.html#migration-step-blocking-mode)). This discussion can probably be decoupled from this PR though. 

Early on I had the same concern. But given that we have a "half measure" Blocking Mode, it might make sense for the ecosystem to migrate to that when possible. In that sense, Legacy is truly "legacy" — it has semantics causing issues today (https://github.com/facebook/react/issues/14536, https://github.com/facebook/react/issues/15080), and it might be good to encourage moving away from it. We'll adjust the messaging as we get more info from people who try it.

Issue #17066, opened 2019-10-11
useEffect callback never called


pschiffmann:
I want to report a bug. My problem is that the callback function I pass to `useEffect()` is never called in my app. I uploaded a minimal app [here](https://github.com/pschiffmann/react-useeffect-issue) that reproduces the issue.

As you can see, I render a couple of nested components. After the initial render, I update the component state inside a `useEffect()` callback. However, this callback is only executed for the first two components, not for the third level component. The result is that the third and subsequent levels are not rendered at all.

I suspect that this is a bug in React and not a misuse on my side because doing any of the following changes will let the component tree render correctly in all levels:
- Don't use multiple React roots. If I remove the last (yellow) `ReactDOM.render()` call, then the second (red) component tree will render correctly.
- Don't conditionally render child components. Removing the `message !== DEFAULT_MESSAGE` check (`main.tsx`, line 20) causes the component trees to render correctly.
- Use `useLayoutEffect()` instead of `useEffect()`.

If you need additional information to reproduce the issue or have any questions, let me know. I'd like provide any help I can!

pschiffmann:
@kunukn Sorry, I forgot to add an effect condition. I changed the `useEffect()` call (see [diff](https://github.com/pschiffmann/react-useeffect-issue/commit/df2423229847afc99482fe0d6f484bf6ede0ca58)), but the result stays the same: Components level 3 and beyond don't render.

vkurchatkin:
Here is somewhat minimal reproduction:

```js
function useTest() {
  const [effect, setEffect] = useState(false);
  useEffect(() => {
    setEffect(true);
  }, []);

  return effect;
}

function A() {
  const effect = useTest();

  return (
    <div >
      {"" + effect}
      {effect && <B/>}
    </div>
  );
}

function B() {
  const effect = useTest();

  return (
    <div >
      {"" + effect}
      {effect && <C/>}
    </div>
  );
}


function C() {
  const effect = useTest();

  return (
    <div >
      {"" + effect}
    </div>
  );
}

function Other() {
  useEffect(() => {}, []);
  return null;
}


ReactDOM.render(
  <A/>,
  document.querySelector("#root1")
);

ReactDOM.render(
  <Other/>,
  document.querySelector("#root2")
);
```

WxSwen:
i met this issue too.
if i use useLayoutEffect or fallback to class by use componentdidmount will be done.
if you solve this problem, tell me;

pschiffmann:
[...]

PR #17335, opened 2019-11-10
test: fix for #17066

aleclarson:
This test is adapted from code in [this comment](https://github.com/facebook/react/issues/17066#issuecomment-541229219).

It passes with the current expectations, so I don't think I've successfully reproduced the bug yet. But this is a good starting point for anyone who wants to help.


acdlite:
This was fixed in https://github.com/facebook/react/issues/17066. Thanks for investigating!

Issue #17314, opened 2019-11-08
Concurrent Mode and UseSubscription with RxJS "lose" updates 


samcooke98:
**Do you want to request a *feature* or report a *bug*?**
**Bug** - I think? 

**What is the current behavior?**

In Concurrent Mode, it appears that if a render is interrupted, if a component is using `useSubscription` the interrupted update is lost, which leads to "tearing" 

The following codesandbox uses `useSubscription` with a RxJS `BehaviorSubject`, mimicking the example from here: https://www.npmjs.com/package/use-subscription#subscribing-to-observables 

In the sandbox, clicking on the "Increment Remote Count" button triggers the RxJs `BehaviorSubject` to increment. This is done outside of the React event handler (ie: via `window.addEventLIstener` and so the updates are not batched together. The update to render the numbers is artificially slowed down. 

If you click the "Increment Remote Count" button multiple times, the update works as expected. 

If you interrupt the update, via clicking the "increment local count", only the last number will update. 

So the Steps to reproduce look like: 
1. Click the "Increment Remote Count" button once
2. Before the update is committed to the DOM, click the "Increment Local Count" update. 
3. The first update is "lost" ie; the output looks like: 

![image](https://user-images.githubusercontent.com/787007/68468141-27255f80-0263-11ea-9afb-07964e21d9d9.png)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

https://jwenc.csb.app/ \ https://codesandbox.io/s/usesubscriptionconcurrentlosingupdates-jwenc 

**What is the expected behavior?**

I'd expect there to be a commit as the above screenshot, but I'd then expect there to be a follow-up commit that restores the consistency. In other words, I'd expect in the above picture for everything to be `1` 

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

`0.0.0-experimental-f6b8d31a7`

I'd be willing to try to take a stab at writing a React test for this, if needed?

dai-shi:
https://github.com/facebook/react/issues/16396#issuecomment-547193356
Here's codesandbox I created.

samcooke98:
Started looking into why this happens - https://github.com/facebook/react/pull/17336

[...]

PR #17213, opened 2019-10-29
DevTools standalone shell changes:

bvaughn:
Cosmetic changes to the RN shell:
1. Update React Native wording to mention recent changes to how DevTools and React Native are meant to connect.
2. Tweak styles slightly.

### Before
Screen Shot 2019-10-29 at 1 36 30 PM
### After
Screen Shot 2019-10-29 at 1 36 51 PM

### Before
Screen Shot 2019-10-29 at 1 42 00 PM

### After
Screen Shot 2019-10-29 at 1 57 25 PM

PR #17336, opened 2019-11-10
Add failing test for #17213

samcooke98:
I've tried to add a failing test for the issue described here. 
https://github.com/facebook/react/issues/17314

If you run it locally, and change `isMemo` to be false, the test will pass. The test will also pass if you change the line to 
`    const MemodSubbed = isMemo ? React.memo(Subbed, () => true) : Subbed;`

----

Below are my half-hearted attempts to look into this bug. 

It looks like when the `compare` function is defined it makes React call `createFiberFromTypeAndProps`. However, if the `compare` function is not defined, it means React goes down the codepath below; 

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberBeginWork.js#L380-L403

In other words, the bug is likely caused somewhere in `updateSimpleMemoComponent`

### DIFF
--- a/packages/use-subscription/src/__tests__/useSubscription-test.internal.js
+++ b/packages/use-subscription/src/__tests__/useSubscription-test.internal.js
@@ -52,6 +52,97 @@ describe('useSubscription', () => {
     return replaySubject;
   }
 
+  it('handles memo components', () => {
+    const observable = createBehaviorSubject('start');
+
+    function Child({value = 'default', id }) {
+      React.useEffect(
+        () => {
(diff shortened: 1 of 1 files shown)


gaearon:
Thank you for the test! Took us a while to get to this. Sorry this got autoclosed.
We think https://github.com/facebook/react/pull/18091 should fix it.

samcooke98:
Thanks @gaearon ! Andrew told me as well: https://github.com/facebook/react/issues/17314#issuecomment-589446273 :D

Issue #16550, opened 2019-08-23
Devtools scrollbar hides flamegraph


Curzy:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
scrollbar is covering flamegraph

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**
current
![current](https://user-images.githubusercontent.com/3931792/63563638-b09a8f00-c59c-11e9-95fc-627dcea023b6.png)

without scrollbar, there is more graphs hidden by scrollbar
![without scrollbar](https://user-images.githubusercontent.com/3931792/63563645-b8f2ca00-c59c-11e9-917e-8c63be90cb2d.png)

**What is the expected behavior?**
scrollbar does not cover flame graph or if flamegraph's height is long enough to use scrollbar, flamegraph container's width should shrink about scrollbars width 

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
16.9 / chrome 78.0.3887.7 / macOS 10.14.6 /

gejimayu:
i am working on this

gejimayu:
my merge request has been rejected :(
based on @bvaughn , this is not an issue
i think it's better to close this issue

Curzy:
scrollbar never automatically hides they always covering flame graphs 

in


Chrome 80.0.3968.0  / darkmode

bvaughn:
The issue shown in the screenshot looks like an OS bug of some sort. The scrollbar that's covering content is not the normal, partially transparent OS X scrollbar that only appears when you're actively scrolling.

I'm not sure this is something we _can_ fix, but always hiding scrollbar is not the solution to fixing this.

bvaughn:
Closing this as stale.

PR #17339, opened 2019-11-11
Devtools: Hide scrollbar on flame graph React Profiler

gejimayu:
Hide scrollbar to appear on flame graph React Profiler.

The reason is that the scrollbar seems to block some of the flame graph and does not provide any purpose.

This is the issue https://github.com/facebook/react/issues/16550

This is the end result

Chrome
chrome

Firefox
firefox


gejimayu:
@bvaughn may i ask for review? thanks in advance

gejimayu:
Hi thanks for the response.
I see, understood. What if i set the container width to make room scrollbar?

gejimayu:
Done.
Chrome
chrome
Firefox
firefox

Issue #17237, opened 2019-10-31
react-refresh: add options to override $RefreshReg$ and $RefreshSig$ for better System.js integration


vasiliicuhar:
**Do you want to request a *feature* or report a *bug*?**
Feature

Right now babel plugin emits globals: https://github.com/facebook/react/issues/16604

```js
window.$RefreshReg$ = () => {};
window.$RefreshSig$ = () => type => type;
```

It would be nice to have them configurable. That would allow to use `import.meta` in environments like SystemJS and have simpler implementation:

```js
import runtime from 'react-refresh/runtime'
runtime.injectIntoGlobalHook(window)

System.constructor.prototype.createContext = function (url) {
  return {
    url,
    $RefreshSig$: runtime.createSignatureFunctionForTransform,
    $RefreshReg$: (type, id) => {
      id = url + ' ' + id
      runtime.register(type, id)
    }
  };
};
```

If you don't mind I could create PR with changes to react-refresh/babel next week.

environment:
```js
{
    "systemjs": "^6.1.4",
    "react": "^16.11.0",
    "react-dom": "^16.11.0",
    "react-refresh": "^0.6.0"
 }
```

gaearon:
Happy to take a PR.

PR #17340, opened 2019-11-11
[Fresh] Add options to configure RefreshReg and RefreshSig identifiers

vasiliicuhar:
Allow to override `$RefreshReg$` and `$RefreshSig$` identifiers.
Main reason is to use `import.meta` in async environments (i.e. rollup + SystemJS) builds instead of relying on globals.

Usage: 
```js
{
  "plugins": [
    ["react-refresh/babel", {
      "refreshReg": "import.meta.refreshReg",
      "refreshSig": "import.meta.refreshSig"
    }]
  ]
}
```

Closes: #17237


vasiliicuhar:
I will fix typo in commit message alongside with subsequent changes. It spells configure wrong

PR #17341, opened 2019-11-11
[react-interactions] Expose host instance to Scope Query function

trueadm:
This PR is a follow up from #17293. Specifically, we added a 2nd argument to the `queryAllNodes ` Scope API. After some internal profiling, I noticed that this pushed engineers towards doing extra work after the query scope, which was a waste of CPU cycles.

To avoid doing extra work, this PR changes the API so that we now pass through the host instance to the query function. That means not only does the query function get the previously existing `type` and `props`, but it also gets a third argument `instance`. In the case of ReactDOM, this will be the DOM element instance.

This then allows for more precise querying of the scope without having to do wasted work after – it can be done as part of the query.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScope.js
+++ b/packages/react-reconciler/src/ReactFiberScope.js
@@ -35,17 +35,16 @@ const emptyObject = {};
 
 function collectScopedNodes(
   node: Fiber,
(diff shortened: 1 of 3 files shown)


trueadm:
@tatermelon CC for review.

PR #17347, opened 2019-11-12
[Bugfix] Passive effects triggered by synchronous renders in a multi-root app

acdlite:
The bug
-------

In a multi-root app, certain passive effects (`useEffect`) are never fired. See #17066.

The underlying problem
----------------------

The implicit contract of `flushPassiveEffects` is that, right after calling it, there should be no pending passive effects. In the normal case, in concurrent mode, this is true. But the current implementation fails to account for the case where a passive effect schedules synchronous work, which in turn schedules additional passive effects.

This led to `rootWithPendingPassiveEffects` being overwritten in the commit phase, because an assignment that assumed it was replacing null was actually replacing a reference to another root, which has the consequence of dropping passive effects on that root.

The fix
-------

The fix I've chosen here is, at the beginning of the commit phase, keep flushing passive effects in a loop until there are no more.

This doesn't not change the "public" implementation of `flushPassiveEffects`, though it arguably should work this way, too. I say "public" because it's only used by implementation layers on top of React which we control: mainly, the legacy version of `act` that does not use the mock Scheduler build. So there's probably still a bug in that `act` implementation.

I will address `act` in a follow-up. The ideal solution is to replace the legacy `act` with one implemented directly in the renderer, using a special testing-only build of React DOM. Since that requires a breaking change, we'll need an interim solution. We could make the "public" `act`
recursively flush effects in a loop, as I've done for the commit phase. However, I think a better solution is to stop automatically flushing the synchronous update queue at the end of `flushPassiveEffects`, and instead require the caller to explicitly call `flushSyncUpdateQueue` (or the equivalent) if needed. This follows the same pattern we use internally in the work loop, which is designed to avoid factoring hazards like the one that resulted in this bug.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
@@ -9,7 +9,7 @@
 
 'use strict';
 
-const React = require('react');
+let React;
 let ReactFeatureFlags = require('shared/ReactFeatureFlags');
 
 let ReactDOM;
@@ -26,6 +26,7 @@ describe('ReactDOMFiberAsync', () => {
   beforeEach(() => {
     jest.resetModules();
     container = document.createElement('div');
+    React = require('react');
     ReactDOM = require('react-dom');
     Scheduler = require('scheduler');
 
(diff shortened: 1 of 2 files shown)


gaearon:
Do we know when this regressed?

acdlite:
Not sure, I'll have to check. I'm going to cut a minor regardless, so I'll do that first. If enough people complain we can consider cutting some patches, too.

gaearon:
Can we close #17066 now? Can you verify its repro works now?

acdlite:
Verified with https://codesandbox.io/s/react-typescript-2thgm

PR #17350, opened 2019-11-12
feat: Add useHydrateableEffect

### LINKED ISSUE #14927: useLayoutEffect in ssr
Hi, I do not understand the situation with this hook a bit. I use this hook to perform the animation synchronously with the state update, if I use useEffect, then I have jumps in the animation, because the animation library does not have time to start. Also, the documentation states that useLayoutEffect runs on the same phase as componentDidMount (that is, on the client side), and here my server issues complaints to me about my code. Why is that?

https://codesandbox.io/s/oo47nj9mk9

_Originally posted by @dimensi in https://github.com/facebook/react/pull/14596#issuecomment-466023638_

eps1lon:
This is an attempt to resolve #14927

## Rationale
If the component requires a layout effect no current workaround solves the underlying issue. Explicit fallback views are most likely impossible (that's why we need layout in the first place) or the component itself is a sufficient fallback (e.g. the overall markup is renderable on the server but we need to adjust some minor visuals).

Since components using `useLayoutEffect` will always have an inconsistent UI before hydration we can potentially downgrade these effects to passive effects i.e. it's already inconsistent so we might as well make it inconsistent a little bit longer to make room for higher priority work.

## Behavior

- like useEffect when ssr
- like useEffect when hydrating
- like useLayoutEffect when rendering on the client
- like useLayoutEffect on updates


packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -1337,6 +1337,7 @@ function getReactiveHookCallbackIndex(calleeNode, options) {
   switch (node.name) {
     case 'useEffect':
     case 'useLayoutEffect':
+    case 'useHydrateableEffect':
```
eps1lon: Not so sure about the naming. We don't really hydrate an effect since it those are never executed on the server anyway. I only used hydration in the name to communicate that the behavior of the hook is tied to hydration.

eps1lon:
Closing since the original issue was resolved in https://github.com/facebook/react/issues/14927#issuecomment-572717051

PR #17354, opened 2019-11-13
[react-interactions] Refine custom active element blur logic

trueadm:
This is a follow up to https://github.com/facebook/react/pull/17291. Specifically, these PR tackles three things:

- Has a much better event name for what the actions occurring. Specifically: `onDetachedVisibleNode` -> `onBeforeFocusedElementDetached `
- Provides another callback that lets the user know the active blur has occurred on the removed node, this event is `onFocusedElementDetached `.
- Fixes a bug where where a focused active element is removed as part of a larger sub-tree, where the active element isn't removed directly, but an ancestor above it is. We now use the `beforeRemoveInstance` to correctly unify this logic (removing it from the `removeChild` and `removeChildFromContainer` methods)

## Why do we need two events? What is the for?

These callbacks are to help with our experimental internal accessibility work at Facebook. There are various ways to tackle this problem. I've been working closely with @tatermelon (Tatiana from the accessibility team) to guide this implementation internally and apply it.

The problem these tackle are when active elements (elements that are focused on the document) are unmounted. This is a big problem with accessibility and it's been a big problem that folks have tried to tackle for a while with hacks. The realism is that there's no real good way to do this in user-land today, which is evident by how many websites simply don't handle this use-case.

This problem is made more evident with the introduction of Suspense fallbacks, where content if focused for a spinner, but is then removed to show the actual real content after X time. We need some way to know that this has occurred, so we can re-align with the document and imperatively move focus to the right node.

Two events are essential here:
- `onBeforeFocusedElementDetached` fires before the active focused element is removed from the DOM. This is the perfect place to work out the "before" state, find what nodes are in view and do some calculations needed for the actual active blur event.
- `onFocusedElementDetached ` fires after all the changes have been committed and it's time to use the logic we stored in the before event, to work out what node we should imperatively focus next.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -465,16 +465,17 @@ export function insertInContainerBefore(
   }
 }
 
-function dispatchCustomFlareEvent(
-  type: string,
+function dispatchFlareDetachedBlurEvent(
+  elementDetached: boolean,
   targetInstance: null | Object,
   target: Element | Document,
 ): void {
   // Simlulate the custom event to the React Flare responder system.
   dispatchEventForResponderEventSystem(
-    type,
+    'blur',
     targetInstance,
     ({
+      elementDetached,
       target,
       timeStamp: Date.now(),
     }: any),
@@ -486,17 +487,13 @@ function dispatchCustomFlareEvent(
 function dispatchBeforeActiveElementBlur(element: HTMLElement): void {
   const targtInstance = getClosestInstanceFromNode(element);
   ((selectionInformation: any): SelectionInformation).blurredActiveElement = element;
-  dispatchCustomFlareEvent('beforeactiveelementblur', targtInstance, element);
+  dispatchFlareDetachedBlurEvent(false, targtInstance, element);
 }
 
 function dispatchActiveElementBlur(
   node: Instance | TextInstance | SuspenseInstance,
 ): void {
-  dispatchCustomFlareEvent(
-    'activeelementblur',
-    null,
-    ((node: any): HTMLElement),
-  );
+  dispatchFlareDetachedBlurEvent(true, null, ((node: any): HTMLElement));
 }
 
 // This is a specific event for the React Flare
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -104,6 +104,12 @@ export type ChildSet = void; // Unused
 export type TimeoutHandle = TimeoutID;
 export type NoTimeout = -1;
 
+type SelectionInformation = {|
+  blurredActiveElement: null | HTMLElement,
+  focusedElem: null | HTMLElement,
+  selectionRange: mixed,
+|};
+
 import {
   enableSuspenseServerRenderer,
   enableFlareAPI,
@@ -127,7 +133,7 @@ const SUSPENSE_FALLBACK_START_DATA = '$!';
 const STYLE = 'style';
 
 let eventsEnabled: ?boolean = null;
-let selectionInformation: ?mixed = null;
+let selectionInformation: null | SelectionInformation = null;
 
 function shouldAutoFocusHostComponent(type: string, props: Props): boolean {
   switch (type) {
@@ -205,6 +211,13 @@ export function prepareForCommit(containerInfo: Container): void {
 
 export function resetAfterCommit(containerInfo: Container): void {
   restoreSelection(selectionInformation);
+  if (enableFlareAPI) {
+    const blurredActiveElement = (selectionInformation: any)
+      .blurredActiveElement;
+    if (blurredActiveElement !== null) {
+      dispatchActiveElementBlur(blurredActiveElement);
+    }
+  }
   selectionInformation = null;
   ReactBrowserEventEmitterSetEnabled(eventsEnabled);
   eventsEnabled = null;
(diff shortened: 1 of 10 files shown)


packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -452,38 +465,60 @@ export function insertInContainerBefore(
   }
 }
 
+function dispatchCustomFlareEvent(
+  type: string,
+  targtInstance: null | Object,
```
tatermelon: nit: targetInstance

packages/react-interactions/events/src/dom/Focus.js
```diff
@@ -538,6 +556,23 @@ const focusWithinResponderImpl = {
     props: FocusWithinProps,
     state: FocusState,
   ): void {
+    if (event.type === 'activeelementblur') {
+      const onActiveElementBlur = (props.onActiveElementBlur: any);
+      if (isFunction(onActiveElementBlur)) {
+        const syntheticEvent = createFocusEvent(
+          context,
+          'activeelementblur',
```
necolas: Why do users of this responder need to know about this custom event rather than receiving the `blur` event that isn't reliably dispatched by browsers when the element is removed?

trueadm:
@necolas Thanks. I've addressed all your feedback :)

packages/react-interactions/events/src/dom/Focus.js
```diff
@@ -50,23 +50,26 @@ type FocusEventType =
   | 'blur'
   | 'focuschange'
   | 'focusvisiblechange'
-  | 'detachedvisiblenode';
+  | 'beforeactiveelementblur'
```
necolas: we should update these names too

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -104,6 +104,12 @@ export type ChildSet = void; // Unused
 export type TimeoutHandle = TimeoutID;
 export type NoTimeout = -1;
 
+type SelectionInformation = {|
+  blurredActiveElement: null | HTMLElement,
```
necolas:
```suggestion
  detachedActiveElement: null | HTMLElement,
````

File: packages/react-interactions/events/src/dom/Focus.js [hunk 5]
```diff
@@ -45,28 +45,25 @@ type FocusProps = {
...
+type FocusEventType = 'focus' | 'blur' | 'focuschange' | 'focusvisiblechange';
 
 type FocusWithinProps = {
   disabled?: boolean,
   onFocusWithin?: (e: FocusEvent) => void,
   onBlurWithin?: (e: FocusEvent) => void,
   onFocusWithinChange?: boolean => void,
   onFocusWithinVisibleChange?: boolean => void,
-  onDetachedVisibleNode?: (e: FocusEvent) => void,
+  onBeforeFocusedElementDetached?: (e: FocusEvent) => void,
```
necolas: `onBeforeBlurWithin`, only called when the element is going to be removed from the document

packages/react-interactions/events/src/dom/Focus.js
```diff
@@ -45,28 +45,25 @@ type FocusProps = {
...
+  onFocusedElementDetached?: (e: FocusEvent) => void,
```
necolas: This wouldn't be needed, `onBlurWithin` payload can include the `elementAttached` data

packages/react-interactions/events/src/dom/Focus.js
```diff
@@ -45,28 +45,25 @@ type FocusProps = {
...
 };
 
 type FocusWithinEventType =
   | 'focuswithinvisiblechange'
   | 'focuswithinchange'
   | 'blurwithin'
   | 'focuswithin'
-  | 'detachedvisiblenode';
+  | 'focusedelementdetached'
+  | 'beforefocusedelementdetached';
```
necolas: Replaced with `beforeblurwithin`

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -452,38 +465,57 @@ export function insertInContainerBefore(
...
+    target,
+    RESPONDER_EVENT_SYSTEM | IS_PASSIVE,
+  );
+}
+
+function dispatchBeforeActiveElementBlur(element: HTMLElement): void {
+  const targtInstance = getClosestInstanceFromNode(element);
+  ((selectionInformation: any): SelectionInformation).blurredActiveElement = element;
+  dispatchFlareDetachedBlurEvent(false, targtInstance, element);
+}
```
necolas:
```suggestion
function dispatchBeforeDetachedBlur(target: HTMLElement): void {
  const targtInstance = getClosestInstanceFromNode(element);
  selectionInformation.detachedActiveElement = target;
  
  dispatchEventForResponderEventSystem(
    'beforeblur',
    targetInstance,
    ({
      target,
      timeStamp: Date.now(),
    }: any),
    target,
    RESPONDER_EVENT_SYSTEM | IS_PASSIVE,
  );
```

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -205,6 +211,13 @@ export function prepareForCommit(containerInfo: Container): void {
 
 export function resetAfterCommit(containerInfo: Container): void {
   restoreSelection(selectionInformation);
+  if (enableFlareAPI) {
+    const blurredActiveElement = (selectionInformation: any)
+      .blurredActiveElement;
+    if (blurredActiveElement !== null) {
+      dispatchActiveElementBlur(blurredActiveElement);
+    }
```
necolas:
```suggestion
    const activeElementDetached = (selectionInformation: any).activeElementDetached;
    if (activeElementDetached !== null) {
      dispatchDetachedBlur(activeElementDetached);
    }
```

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -452,38 +465,57 @@ export function insertInContainerBefore(
...
+  const targtInstance = getClosestInstanceFromNode(element);
+  ((selectionInformation: any): SelectionInformation).blurredActiveElement = element;
+  dispatchFlareDetachedBlurEvent(false, targtInstance, element);
+}
+
+function dispatchActiveElementBlur(
+  node: Instance | TextInstance | SuspenseInstance,
+): void {
+  dispatchFlareDetachedBlurEvent(true, null, ((node: any): HTMLElement));
+}
```
necolas:
```suggestion
function dispatchBlur(target) {
  const targtInstance = getClosestInstanceFromNode(target);
  dispatchEventForResponderEventSystem(
    'blur',
    targetInstance,
    ({
      elementAttached: false,
      target,
      timeStamp: Date.now(),
    }: any),
    target,
    RESPONDER_EVENT_SYSTEM | IS_PASSIVE,
  );
```

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -452,38 +465,57 @@ export function insertInContainerBefore(
...
-      'detachedvisiblenode',
-      targetFiber,
-      ({
-        target: child,
-        timeStamp: Date.now(),
-      }: any),
-      ((child: any): Document | Element),
-      RESPONDER_EVENT_SYSTEM | IS_PASSIVE,
-    );
+    dispatchBeforeActiveElementBlur(((instance: any): HTMLElement));
```
necolas:
```suggestion
    dispatchBeforeDetachedBlur(((instance: any): HTMLElement));
```

packages/react-dom/src/client/ReactInputSelection.js
```diff
@@ -100,6 +100,8 @@ export function hasSelectionCapabilities(elem) {
 export function getSelectionInformation() {
   const focusedElem = getActiveElementDeep();
   return {
+    // Used by Flare
+    blurredActiveElement: null,
```
necolas:
```suggestion
    activeElementDetached: null,
```

trueadm:
@necolas Mind adding your changes as a follow up PR if you have the time?

PR #17362, opened 2019-11-14
Update Prettier parser name

lewislbr:
According to [Prettier's parser options](https://prettier.io/docs/en/options.html#parser), "babylon" was changed to "babel":

> "babel" (via @babel/parser) Named "babylon" until v1.16.0

So I just updated the name.


lewislbr:
Closing it at it was already addressed in #17811

PR #17369, opened 2019-11-14
Added unstable_do_not_use_getDOMNodes to Suspense boundaries

lunaruan:
Added ability to get child DOM nodes from suspense boundaries after the children of a suspense boundary unsuspends or during mount if children do not suspend


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1101,6 +1101,15 @@ function commitPlacement(finishedWork: Fiber): void {
...
     } else if (node.child !== null) {
+      if (
+        node.tag === SuspenseComponent &&
+        node.memoizedState === null &&
+        typeof node.memoizedProps.unstable_do_not_use_getDOMNodes === 'function'
+      ) {
+        node.memoizedProps.unstable_do_not_use_getDOMNodes(
+          getChildrenOfSuspenseBoundary(finishedWork),
+        );
+      }
```
acdlite: Need to pick off `unstable_do_not_use_getDOMNodes` before calling it, otherwise its `this` context will be the memoized props:

```js
const getDOMNodes = node.memoizedProps.unstable_do_not_use_getDOMNodes;
if (typeof getDOMNodes === 'function') {
  getDOMNodes(...)
}
```

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1477,6 +1492,49 @@ function commitSuspenseComponent(finishedWork: Fiber) {
   }
 }
 
+function getChildrenOfSuspenseBoundary(suspenseBoundary) {
```
acdlite: For updates, both this function and `hideOrUnhideAllChildren` will be called. Can you refactor so that `hideOrUnhideAllChildren` does both during updates? So we don't have to traverse twice.

packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
```diff
@@ -1322,4 +1322,161 @@ describe('ReactSuspense', () => {
...
+    expect(Scheduler).toHaveYielded([
+      'Suspend! [A]',
+      'Suspend! [B]',
+      'Suspend! [C]',
+      'Promise resolved [C]',
+      'Promise resolved [A]',
+    ]);
+    expect(Scheduler).toFlushAndYield(['A', 'C']);
+    expect(root).toMatchRenderedOutput('AHiLoadingC');
+    expect(children).toEqual([
```
acdlite: Need some tests that render host nodes, not just text. Also you can use `toMatchRenderedOutput`. That will make it easier to convert these to a different renderer, like noop, which is the plan for all the tests in this file. In fact... do you mind moving this to `ReactSuspenseWithNoopRenderer-test.internal.js` :D

packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js
```diff
@@ -1322,4 +1322,161 @@ describe('ReactSuspense', () => {
...
+    expect(Scheduler).toFlushAndYield(['B']);
+    expect(counter).toEqual(2);
+    expect(root).toMatchRenderedOutput('AB');
+    expect(children).toEqual([
+      {isHidden: false, tag: 'TEXT', text: 'A'},
+      {isHidden: false, tag: 'TEXT', text: 'B'},
+    ]);
+  });
+
+  it('does not unstable_do_not_use_getDOMNodes during updates if component does not suspend', () => {
```
necolas: "does not call..."

(A+ quality review)

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -2630,4 +2631,213 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
+    let _setShowText;
+    function App() {
+      const [showText, setShowText] = React.useState(false);
+      _setShowText = setShowText;
+      return (
+        <Suspense
+          fallback={'Loading'}
+          unstable_do_not_use_getHostNodes={suspenseChildren => {
+            children = suspenseChildren;
+            counter++;
```
acdlite: Push to a log instead of using a counter, please

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1090,6 +1158,19 @@ function commitPlacement(finishedWork: Fiber): void {
...
       // the portal directly.
     } else if (node.child !== null) {
+      if (
+        enableDoNotUseGetHostNodes &&
+        node.tag === SuspenseComponent &&
+        node.memoizedProps !== null &&
+        node.memoizedState === null &&
+        typeof node.memoizedProps.unstable_do_not_use_getHostNodes ===
+          'function'
+      ) {
```
acdlite: I don't get why any of this new code is necessary.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1447,8 +1528,25 @@ function commitSuspenseComponent(finishedWork: Fiber) {
...
+  if (primaryChildParent !== null) {
+    console.log('hi');
+    const getHostNodes =
+      finishedWork.memoizedProps.unstable_do_not_use_getHostNodes;
+    if (
+      enableDoNotUseGetHostNodes &&
+      !newDidTimeout &&
+      typeof getHostNodes === 'function'
+    ) {
+      console.log('nooooo');
```
acdlite: Oops

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1447,8 +1528,25 @@ function commitSuspenseComponent(finishedWork: Fiber) {
...
+      const children = getAndToggleHideForChildren(
+        primaryChildParent,
+        newDidTimeout,
+      );
+      getHostNodes(children);
+    } else {
+      console.log('wolf', newDidTimeout);
```
acdlite: Oops

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -21,6 +21,7 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
     ReactFeatureFlags.replayFailedUnitOfWorkWithInvokeGuardedCallback = false;
     ReactFeatureFlags.flushSuspenseFallbacksInTests = false;
+    ReactFeatureFlags.enableDoNotUseGetHostNodes = true;
```
acdlite: No new feature flags it tests. Use `it.experimental` instead.

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -2630,4 +2631,200 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
+    const log = [];
+    function Child() {
+      return 'Hi';
+    }
+    function App() {
+      return (
+        <Suspense
+          fallback={'Loading'}
+          unstable_do_not_use_getHostNodes={suspenseChildren => {
+            log.push(suspenseChildren);
```
acdlite: Use `Scheduler.unstable_yieldValue` instead

packages/shared/ReactFeatureFlags.js
```diff
@@ -92,3 +92,5 @@ export const enableTrustedTypesIntegration = false;
 
 // Flag to turn event.target and event.currentTarget in ReactNative from a reactTag to a component instance
 export const enableNativeTargetAsInstance = false;
+
+export const enableDoNotUseGetHostNodes = false;
```
acdlite:
```suggestion
export const enableDoNotUseGetHostNodes = __EXPERIMENTAL__;
```

acdlite: Do the same in the other files

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -686,6 +687,73 @@ function hideOrUnhideAllChildren(finishedWork, isHidden) {
   }
 }
 
+function getAndToggleHideForChildren(finishedWork, isHidden) {
```
acdlite: Put a comment here that explains this is a fork of `hideAndUnhideAllChildren`

packages/react-reconciler/src/ReactFiberCommitWork.js
acdlite: Oh this is to handle the initial render, but not in the place I expected. Let's handle this in `appendAllChildren` in the complete phase instead, since we already do a traversal there.

lunaruan:
Closing this for now because Sebastian came up with an alternative idea for TTI tracking that doesn't rely on React.

PR #17372, opened 2019-11-15
[Flight] Move Flight DOM to a Webpack Specific Package

sebmarkbage:
We'll have Webpack specific coupling so we need to ensure that it can be versioned separately from various Webpack versions. We'll also have builds for other bundlers in the future.

There's a matrix of configurations that include the server environment (node vs browser streams), client environment (browser vs RN), format (DOM vs RN), bundler (Webpack, Parcel, Rollup...) and possibly others. Each yielding a possible new build.

In practice, we only use a few of those combinations to create optimal builds.

Technically, the server doesn't have to be Webpack specific if the Babel transform is, but it might be good to have the transform be agnostic. E.g. it allows for the Node env and Webpack compile in parallel. So therefore I think it's better that the transform is agnostic and then we replace it with bundle meta data at runtime using a Webpack specific server.


packages/react-flight-dom-webpack/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "node": ">=0.10.0"
+  },
+  "peerDependencies": {
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0"
+  },
+  "dependencies": {
+    "loose-envify": "^1.1.0",
+    "object-assign": "^4.1.1",
+    "webpack": "^4.41.2"
```
gaearon: Should be a peer?

PR #17382, opened 2019-11-15
[Bugfix] Pending state is always user-blocking

### LINKED ISSUE #17272: When calling a useTransition startTransition callback outside of event handlers, isPending is never set to true

Bug

`isPending` is never set to true when calling `startTransition` within `useEffect`, but it *does* work properly when within a `useLayoutEffect`.

https://codesandbox.io/s/usetransition-useeffect-issues-p1j9s

Here's the correct behavior (accomplished via `useLayoutEffect`):

good

Here's the incorrect behavior (via `useEffect`):

bad

Note the difference is that the opacity never changes to 0.4 (which is determined based on the `isPending` state).

I expect them to both behave the same (at least as far as the user can observe).

- react@0.0.0-experimental-f6b8d31a7
- react-dom@0.0.0-experimental-f6b8d31a7

acdlite:
Fixes a bug where `isPending` is only set to `true` if `startTransition` is called from inside an input event. That's usually the case, but not always.

Now it works regardless of where you call it.

Fixes #17272

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -19,7 +19,6 @@ import type {HookEffectTag} from './ReactHookEffectTags';
 import type {SuspenseConfig} from './ReactFiberSuspenseConfig';
 import type {ReactPriorityLevel} from './SchedulerWithReactIntegration';
 
-import * as Scheduler from 'scheduler';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 
 import {NoWork} from './ReactFiberExpirationTime';
@@ -53,7 +52,12 @@ import getComponentName from 'shared/getComponentName';
 import is from 'shared/objectIs';
 import {markWorkInProgressReceivedUpdate} from './ReactFiberBeginWork';
 import {requestCurrentSuspenseConfig} from './ReactFiberSuspenseConfig';
-import {getCurrentPriorityLevel} from './SchedulerWithReactIntegration';
+import {
+  UserBlockingPriority,
+  NormalPriority,
+  runWithPriority,
+  getCurrentPriorityLevel,
+} from './SchedulerWithReactIntegration';
 
 const {ReactCurrentDispatcher, ReactCurrentBatchConfig} = ReactSharedInternals;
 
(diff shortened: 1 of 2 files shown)


gaearon:
Btw we track this in https://github.com/facebook/react/issues/17272

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1157,65 +1159,62 @@ function updateDeferredValue<T>(
...
+      }
     },
     [value, config],
   );
   return prevValue;
 }
 
+function startTransition(setPending, config, callback) {
+  const priorityLevel = getCurrentPriorityLevel();
+  runWithPriority(
```
sebmarkbage: I feel like we do this in a couple of internal places which seems like a very heavy weight way to do this. Going through the scheduler to do this. This adds to my suspicion that reading priorities from scheduler is the wrong model. It's not that scheduler dictates the priorities but it sometimes can be used as a default.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1157,65 +1159,62 @@ function updateDeferredValue<T>(
...
+    priorityLevel < UserBlockingPriority ? UserBlockingPriority : priorityLevel,
+    () => {
+      setPending(true);
+    },
+  );
+  runWithPriority(
+    priorityLevel > NormalPriority ? NormalPriority : priorityLevel,
```
sebmarkbage: Similarly here, it's not obvious to me that calling startTransition should change the running "current" queue of the scheduler. It should only change the priority inside React. They're not necessarily 1:1.

packages/react-reconciler/src/__tests__/ReactTransition-test.internal.js
```diff
@@ -0,0 +1,102 @@
...
+let ReactFeatureFlags;
+let React;
+let ReactNoop;
+let Scheduler;
+let Suspense;
+let useState;
+let useTransition;
+let act;
+
+describe('ReactHooksWithNoopRenderer', () => {
```
gaearon: Copy paste name

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: I agree, it's on my list of things to consider during expiration times refactor. Didn't want to couple it to this bugfix.

Issue #15520, opened 2019-04-27
Cannot suppress error boundary output using react-test-renderer


mpeyper:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

When following @gaearon's example on [how to suppress the error boundary logging in tests expecting errors](https://github.com/facebook/react/issues/11098#issuecomment-412682721), it appears to not be working in 16.8.6 (I'm not sure which versions after 16.4.3 is _has_ worked for).

[Here is a sandbox.](https://codesandbox.io/s/l5mrl747wm?expanddevtools=1&fontsize=14&module=%2Fsrc%2Fexample.test.js&previewwindow=tests)

I had to make a change to the `renderError` as it looks like the way codesandbox executes the script cause the `event.error` to be hidden from the code (it just had an object with `isTrusted: true` on it - google led me [here](https://stackoverflow.com/a/44862693/6902543) so I changed it to just count the number of errors).

I also made a change to be able to use `react-test-renderer` as well as `react-dom` as I'm actually after a way to suppress the error using the test renderer for an [issue raised in `react-hooks-testing-library`](https://github.com/mpeyper/react-hooks-testing-library/issues/50).

In my sandbox, you will see the errors

> The above error occurred in the <Darth> component:
> in Darth
> in TestBoundary
>
> React will try to recreate this component tree from scratch using the error boundary you provided, TestBoundary.

and

> The above error occurred in the <ObiWan> component:
> in ObiWan
> in TestBoundary
> 
> React will try to recreate this component tree from scratch using the error boundary you provided, TestBoundary.

get printed to the console, despite the `event.preventDefault()` call.

The other main thing to note is that the `react-test-renderer` test does not pass, which implies the error listener is not firing for it, which sort of makes sense to me as it's not using the DOM, so it's not erroring on the window, but my understanding of these things is somewhat limited.

**What is the expected behavior?**

Assuming the `expectRenderError`, `TestBoundary` and `handleTopLevelError` code is sufficient to suppress the output as indicated in #11098, I would expect the `react-dom` test to produce no output.

Assuming the same setup should also work for `react-test-renderer`, I would also expect that test to have no output.

If either of these assumptions are wrong, how should one go about suppressing the output in these tests?  

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

`react`: `16.8.6`
`react-dom`: `16.8.6`
`react-test-renderer`: `16.8.6`

[...]

PR #17383, opened 2019-11-16
[react-test-renderer] Silence error boundary console logs

milesj:
to: @trueadm @acdlite @bvaughn 
cc: @mpeyper

Solves https://github.com/facebook/react/issues/15520

Basically, when an error is thrown and an error boundary exists, `react-test-renderer` will *always* log the "The above error occurred in the component" message with `console.error`. There's currently no way to disable or hide it.

This PR makes a change (using forks) that entirely silences that log if an error boundary catches it. I'm unable to think of any situation where this log is desirable.

Would love to avoid this current hack: https://github.com/milesj/rut/blob/master/packages/rut/src/internals/patch.ts

### DIFF
--- a/packages/react-reconciler/src/forks/ReactFiberErrorDialog.test.js
+++ b/packages/react-reconciler/src/forks/ReactFiberErrorDialog.test.js
@@ -1,0 +1,30 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import type {CapturedError} from '../ReactCapturedValue';
+
+export type Features = {
+  errorBoundary: boolean;
+}
+
+const silenced: $Exact<Features> = {
+  errorBoundary: true,
+};
+
+export function silenceFeatures(features: Features) {
+  Object.assign(silenced, features);
+}
+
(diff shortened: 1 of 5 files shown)


milesj:
Weekend bump. @trueadm @acdlite @bvaughn

milesj:
Would love to get this in if possible.

milesj:
What's the best way to get someone to look at this?

mpeyper:
If it's any consolation, I'm watching this with much anticipation as well.

milesj:
@gaearon Ping.

mpeyper:
I've got a [new issue in `@testing-library/react-hooks`](https://github.com/testing-library/react-hooks-testing-library/issues/305) that looks like it needs me to reintroduce an error boundary to resolve it, but that's going to potentially cause a significant amount of output in my user's tests because of this issue and I'd really like to avoid it (as the issues are inevitably going to come to me about it).

At risk of sounding whingy, can someone please take a look at this?

mpeyper:
Bump, although I think the chances of anyone actually looking at this are quite slim now.

milesj:
Tried rebasing and it seems like a ton of internals have changed. This doesn't work anymore.

mpeyper:
That's frustrating. I'm sorry you've (and I've) had this experience and that it's gone unnoticed for so long to the point where we have to start from scratch again now.

PR #17385, opened 2019-11-16
Update eslint rule exhaustive deps to use new suggestions feature

### LINKED ISSUE #16313: [ESLint]react-hooks/exhaustive-deps rule autofix modifies code function, violating eslint best practices
Bug

Consider:
```jsx
function Example({ fetchData, someArg }) {
  let data = 'defaultValue';
  useEffect(() => {
    data = fetchData(someArg);
 }, [someArg]);

  return JSON.stringify(data);
}
```
auto fix will automatically change this to:
```jsx
function Example({ fetchData, someArg }) {
  let data = 'defaultValue';
  useEffect(() => {
    data = fetchData(someArg);
 }, [someArg, fetchData]);

  return JSON.stringify(data);
}
```

Here we have a simple data fetch component that should refetch data when parameters change, but not when the fetch function changes.  Whether this is right or not, it is legitimate code, that the lint fix will cause serious problems in the code if it is used like this:
```jsx
function ExmapleUsage({ fetchData }) {
  return <Example fetchData={( arg ) => fetchData('Hello World', arg)} someArg="Goodbye" /> 
}
```

[Eslint best practices](https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes) say that fix rules should not change functionality of code, so that you can safely run fix and expect no functional changes to be made.  This rule directly breaks that.  I as a repo maintainer see the auto fix as a greater risk than the problems the lint rule prevents.  If autofix was turned off, the rule would be entirely a positive.  
Unfortunately, [ESLint also has rejected the idea of disabling autofix for certain rules](https://github.com/eslint/eslint/issues/7549).  So not following best practices is not ideal.

Not a React Bug but a lint rule.

wdoug:
This pull request closes #16313

Assuming that https://github.com/eslint/eslint/pull/12384 moves forward and adds the suggestions feature to `eslint`, this pull request updates the `react-hooks/exhaustive-deps` `eslint` rule to use the suggestions feature. This will result in the rule no-longer modifying code functionality when running `eslint`'s autofixer while providing a new method to explicitly accept code suggestions for the rule.

For more details see: #16313

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -19,6 +19,16 @@ ESLintTester.setDefaultConfig({
   },
 });
 
+/**
+ * A string template tag that removes padding from the left side of multi-line strings
+ * @param {Array} strings array of code strings (only one expected)
+ */
+function normalizeIndent(strings) {
+  const codeLines = strings[0].split('\n');
+  const leftPadding = codeLines[1].match(/\s+/)[0];
+  return codeLines.map(line => line.substr(leftPadding.length)).join('\n');
+}
+
 // ***************************************************
 // For easier local testing, you can add to any case:
 // {
@@ -32,7 +42,7 @@ ESLintTester.setDefaultConfig({
 const tests = {
   valid: [
     {
-      code: `
+      code: normalizeIndent`
         function MyComponent() {
           const local = {};
           useEffect(() => {
@@ -42,7 +52,7 @@ const tests = {
       `,
     },
     {
-      code: `
+      code: normalizeIndent`
         function MyComponent() {
           useEffect(() => {
             const local = {};
@@ -52,7 +62,7 @@ const tests = {
       `,
     },
     {
-      code: `
+      code: normalizeIndent`
         function MyComponent() {
           const local = {};
           useEffect(() => {
@@ -67,7 +77,7 @@ const tests = {
       // to be an import that hasn't been added yet, or
       // a component-level variable. Ignore it until it
       //  gets defined (a different rule would flag it anyway).
-      code: `
+      code: normalizeIndent`
         function MyComponent() {
           useEffect(() => {
             console.log(props.foo);
@@ -76,7 +86,7 @@ const tests = {
       `,
     },
     {
-      code: `
+      code: normalizeIndent`
         function MyComponent() {
           const local1 = {};
           {
@@ -90,7 +100,7 @@ const tests = {
       `,
     },
     {
-      code: `
+      code: normalizeIndent`
         function MyComponent() {
           const local1 = {};
           {
@@ -104,7 +114,7 @@ const tests = {
       `,
     },
     {
-      code: `
+      code: normalizeIndent`
         function MyComponent() {
           const local1 = {};
           function MyNestedComponent() {
@@ -118,7 +128,7 @@ const tests = {
       `,
     },
     {
-      code: `
+      code: normalizeIndent`
         function MyComponent() {
           const local = {};
           useEffect(() => {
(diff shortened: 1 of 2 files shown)


gaearon:
Nice! Please ping this PR when it lands.

wdoug:
Suggestions have been merged. Waiting on a release now.

wdoug:
Suggestions released in [6.7.0](https://github.com/eslint/eslint/releases/tag/v6.7.0)

gaearon:
What happens if you have older ESLint? Do we just lose the suggestions?

Should we only fix this forward?

wdoug:
I was assuming that we would bump the major version of this package and possibly set the peer dependency of eslint to `>=6.7.0`. Probably adding some notes in the Readme would also be valuable. 

I don't think we'd want to try to support both suggestions and the autofixer at the same time based on something like the eslint version because then the functionality would unexpectedly change when someone updated their version of eslint. Also, I don't think we want to support the autofix option going forward as it is ultimately unsafe.

I personally think it would be better to have no fixes for folks that happen to have an older version of eslint. For example, at my company, I enabled the `exhaustive-deps` rule and then immediately disabled it when I realized that most everyone at my company has their editors set to automatically run eslint's autofixer and we have a ton of misuses of hooks dependencies that, without manual work to address them, would cause user-facing bugs when this rule's autofix would run. Because of our current setup, it is likely that many of these bugs would end up getting out to production and affecting actual users. In our case, it would be much better to have the rule warning with no assistance fixing the issues than have the rule attempt to fix them for us without us noticing potential breaking changes.

gaearon:
Just tried this with VS Code and couldn't get suggestions to show up. Does it mean extensions need to be updated?

wdoug:
Yes. I have opened [this issue](https://github.com/microsoft/vscode-eslint/issues/806#issuecomment-558078967) for the vscode extension

gaearon:
We probably shouldn’t merge until mainstream extensions catch up.

wdoug:
Is there any chance we could release a beta prerelease with the suggestions feature while we wait for extensions to catch up?

wdoug:
For reference, I'm hoping that [this `vscode-eslint` pull request](https://github.com/microsoft/vscode-eslint/pull/814) will get merged and released soon.

gaearon:
I’d feel comfortable releasing after VSCode plugin update is live. Can you check if there’s anything holding it back on that end?

gaearon:
Can you test that the rule works as expected with this release? https://github.com/microsoft/vscode-eslint/pull/814#issuecomment-587020489

gaearon:
This PR is published as `eslint-plugin-react-hooks@2.4.0-alpha.0`. Please confirm that it works.

gaearon:
Works in my testing.

gaearon:
`eslint-plugin-react-hooks@2.4.0` is out with this fix.

gaearon:
This is great work. Thank you for doing it.

wdoug:
My pleasure. It looks good on my end as well. Glad it finally got out.

nstepien:
I prefered the rule with autofix working. Would it be possible to add an option to make it autofixable again?
The VSCode extension update with suggestions support isn't out yet, so it's making things a little bit awkward at the moment.

LeopoldLerch:
> I prefered the rule with autofix working. Would it be possible to add an option to make it autofixable again?
> The VSCode extension update with suggestions support isn't out yet, so it's making things a little bit awkward at the moment.

Thing is, that is a common rule of eslint, that it does not change the code in regards of how it works. Which it in fact did with the autofix of the dependencies, expecially with useeffect. Altough a opt-in option for readding autofix could work, there is still the question if it SHOULD be there at all. 

The update of the extension is already on the way, you can install it using @next . Regular version should be updated on the weekend as there corresponding thread of the vscode extension suggests.
In the meanwhile you can still use the prior-version of the eslint-ruleset.

gaearon:
@nstepien You don't have to upgrade. :-) Feel free to revert to 2.3.0. We got this out early because https://github.com/facebook/react/issues/16313 has been causing pain for a while, and it can't wait more.

PR #17394, opened 2019-11-18
[react-interactions] Rename Flare listeners prop to DEPRECATED_flareListeners

trueadm:
This renames the `listeners` prop to `DEPRECATED_flareListeners`. This is for two reasons:

- to discourage internal usage of this experimental API
- to enable us to experiment with another event API that might make use of the `listeners` prop in the future


TrySound:
Do you completely reject this api? It looked very good from type system point of view. All low level event listeners could be eliminated in favour of single listeners prop.

trueadm:
@TrySound We don't necessarily reject the `listeners` prop and we will likely use something like it again, but with another experimental event system that is different from that of that we have now for React Flare.

PR #17398, opened 2019-11-18
[Flight] Chunks API

sebmarkbage:
__This is a very early experiment. We just need something to play with.__

This is a new core API but it's not really meant to be used without Flight. Instead Flight will produce these "chunks" from its payload. The primary purpose of this in core is to ensure the reconciler can update the same component instance with new data.

It can be used without Flight by running it in the same environment and it works fine. That is a nice property because a) it means you can get started running this code in a client-only environment and debug in browser. b) it describes the semantics in a standalone way which is how the type system thinks about it, so it's nice to be able to think about it this way.

This implementation doesn't support Context. Exposing React's context wouldn't be the same semantics as with Flight so it's not super easy and requires more code to actually support Context in this mode. That's unfortunate.

### DIFF
--- a/packages/react-reconciler/src/ReactChildFiber.js
+++ b/packages/react-reconciler/src/ReactChildFiber.js
@@ -19,6 +19,7 @@ import {
   REACT_ELEMENT_TYPE,
   REACT_FRAGMENT_TYPE,
   REACT_PORTAL_TYPE,
+  REACT_CHUNK_TYPE,
 } from 'shared/ReactSymbols';
 import {
   FunctionComponent,
@@ -26,9 +27,10 @@ import {
   HostText,
   HostPortal,
   Fragment,
+  Chunk,
 } from 'shared/ReactWorkTags';
 import invariant from 'shared/invariant';
-import {warnAboutStringRefs} from 'shared/ReactFeatureFlags';
+import {warnAboutStringRefs, enableChunksAPI} from 'shared/ReactFeatureFlags';
 
 import {
   createWorkInProgress,
@@ -392,32 +394,48 @@ function ChildReconciler(shouldTrackSideEffects) {
     element: ReactElement,
     expirationTime: ExpirationTime,
   ): Fiber {
-    if (
-      current !== null &&
-      (current.elementType === element.type ||
+    if (current !== null) {
+      if (
+        current.elementType === element.type ||
         // Keep this check inline so it only runs on the false path:
-        (__DEV__ ? isCompatibleFamilyForHotReloading(current, element) : false))
-    ) {
-      // Move based on index
-      const existing = useFiber(current, element.props, expirationTime);
-      existing.ref = coerceRef(returnFiber, current, element);
-      existing.return = returnFiber;
-      if (__DEV__) {
-        existing._debugSource = element._source;
-        existing._debugOwner = element._owner;
+        (__DEV__ ? isCompatibleFamilyForHotReloading(current, element) : false)
+      ) {
+        // Move based on index
+        const existing = useFiber(current, element.props, expirationTime);
+        existing.ref = coerceRef(returnFiber, current, element);
+        existing.return = returnFiber;
+        if (__DEV__) {
+          existing._debugSource = element._source;
+          existing._debugOwner = element._owner;
+        }
+        return existing;
+      } else if (
+        enableChunksAPI &&
+        current.tag === Chunk &&
+        element.type.$$typeof === REACT_CHUNK_TYPE &&
+        element.type.render === current.type.render
+      ) {
+        // Same as above but also update the .type field.
+        const existing = useFiber(current, element.props, expirationTime);
+        existing.ref = coerceRef(returnFiber, current, element);
+        existing.return = returnFiber;
+        existing.type = element.type;
+        if (__DEV__) {
+          existing._debugSource = element._source;
+          existing._debugOwner = element._owner;
+        }
+        return existing;
       }
-      return existing;
-    } else {
-      // Insert
-      const created = createFiberFromElement(
-        element,
-        returnFiber.mode,
-        expirationTime,
-      );
-      created.ref = coerceRef(returnFiber, current, element);
-      created.return = returnFiber;
-      return created;
     }
+    // Insert
+    const created = createFiberFromElement(
+      element,
+      returnFiber.mode,
+      expirationTime,
+    );
+    created.ref = coerceRef(returnFiber, current, element);
+    created.return = returnFiber;
+    return created;
   }
 
   function updatePortal(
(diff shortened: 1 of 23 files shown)


packages/react/src/chunk.js
```diff
@@ -0,0 +1,70 @@
...
+      );
+    }
+  }
+  return function(): Chunk<Props> {
+    let args = arguments;
+    return {
+      $$typeof: REACT_CHUNK_TYPE,
+      query: function() {
+        return query.apply(null, args);
+      },
```
sebmarkbage: In Flight, this function will be replaced by one that suspends on the data, as well as as a render that suspends on the code.

packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -396,32 +398,47 @@ function ChildReconciler(shouldTrackSideEffects) {
...
+        if (__DEV__) {
+          existing._debugSource = element._source;
+          existing._debugOwner = element._owner;
+        }
+        return existing;
+      } else if (
+        enableChunksAPI &&
+        current.tag === Chunk &&
+        element.type.$$typeof === REACT_CHUNK_TYPE &&
+        element.type.render === current.type.render
```
sebmarkbage: @gaearon There's a couple of places in this PR where I should probably do something for fast refresh but not sure what so I left that out.

gaearon: You can copy what we do for forwardRef. I can add it.

packages/react/src/chunk.js
```diff
@@ -0,0 +1,68 @@
...
+): (...args: Args) => Chunk<Props> {
+  if (__DEV__) {
+    if (typeof query !== 'function') {
+      warningWithoutStack(
+        false,
+        'Chunks require a query function but was given %s.',
+        query === null ? 'null' : typeof query,
+      );
+    }
+    if (render != null && render.$$typeof === REACT_MEMO_TYPE) {
```
gaearon: fowardRef too?

packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -1141,34 +1158,63 @@ function ChildReconciler(shouldTrackSideEffects) {
       // TODO: If key === null and child.key === null, then this only applies to
       // the first item in the list.
       if (child.key === key) {
-        if (
-          child.tag === Fragment
-            ? element.type === REACT_FRAGMENT_TYPE
```
gaearon: I don't think removing this check is correct.

As a result, `Fragment -> SomethingElse` change won't reset the child state.

Here's a small case to show it:

```js
  fit('should reset state when reconciling from fragment to another type', async () => {
    function Foo({children}) {
      return children;
    }

    function App({step}) {
      let Wrapper = step === 0 ? React.Fragment : Foo;
      return (
        <>
          <Wrapper>
            <Child />
          </Wrapper>
        </>
      )
    }

    let _setChildState;
    function Child() {
      const [x, setX] = React.useState('hello')
      _setChildState = setX;
      return x;
    }

    await ReactNoop.act(async () => {
      ReactNoop.render(<App step={0} />);
    });
    expect(ReactNoop.getChildren()[0].text).toBe('hello');

    // Change the child state
    await ReactNoop.act(async () => {
      _setChildState('goodbye');
    });
    expect(ReactNoop.getChildren()[0].text).toBe('goodbye');
    // Ensure it stays
    await ReactNoop.act(async () => {
      _setChildState('goodbye');
    });
    expect(ReactNoop.getChildren()[0].text).toBe('goodbye');

    // This should reset the child state.
    await ReactNoop.act(async () => {
      ReactNoop.render(<App step={1} />);
    });
    expect(ReactNoop.getChildren()[0].text).toBe('hello');
  });
```

gaearon: If you add a test for this, probably worth making it switch back to Frag too just for completeness.

gaearon: As a fix I think you want to wrap the Fragment case in `element.type === REACT_FRAGMENT_TYPE`. (And add a `break` to it?)

gaearon: Actually I think a break would be insufficient? Since you want to go to the fallthrough case. But you don't want to fall through to the Chunk (or any other case we add in the middle). So I think this needs more restructuring.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -687,6 +689,82 @@ function updateFunctionComponent(
...
+      }
+    }
+    setCurrentPhase(null);
+  } else {
+    nextChildren = renderWithHooks(
+      current,
+      workInProgress,
+      render,
+      nextProps,
+      data,
```
gaearon: This arg is currently called `refOrContext` which is a bit confusing. Maybe rename to something more generic?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1134,6 +1212,21 @@ function mountLazyComponent(
       );
       break;
     }
+    case Chunk: {
+      if (enableChunksAPI) {
```
gaearon: Test for lazy pls? This codepath is easy to break.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -3207,6 +3300,19 @@ function beginWork(
...
+        const props = workInProgress.pendingProps;
+        return updateChunk(
+          current,
+          workInProgress,
+          chunk,
+          props,
+          renderExpirationTime,
+        );
+      }
+    }
```
gaearon: Kind of inconsistent we don't throw for scopes/fundamental when their flags are off. Maybe add a `break` here to be consistent? Or fix the other ones to throw by removing their breaks.

packages/react-reconciler/src/ReactFiberBeginWork.js
gaearon: Alternatively maybe throw early to indicate we don't aim to support this yet.

packages/react-reconciler/src/ReactChildFiber.js
sebmarkbage: Yea this is just an oversight in the copy paste. I was wondering if we can just get rid of it though and have the elementType comparison.

acdlite: Might be something to consider before landing the JSX changes @lunaruan is working on.

devknoll:
QQ: From what I can see, this looks like it would work in legacy mode too, right? Is that intentional?

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: We do throw. If you look, this has the same semantics because the default case isn't a `default` but outside the block. `break` means default.

This should allow us to dead code eliminate the case statements since they all fall into the default case.

However, we're kind of inconsistent with where we do this. Pattern.

PR #17411, opened 2019-11-20
useTransition improvements

acdlite:
Stack of improvements to `useTransition`. Some of these will require a refactor of our "expiration time" model, and will therefore have implications beyond the scope of `useTransition`.

I'm planning to land this in stages. The first few commits are useful even before I start changing the expiration time model. They also block a concurrent mode release. Whereas the expiration times refactor is riskier, and it's possible they won't block the initial release.

On the other hand, breaking these changes into too many parts may be *more* risky, because instead of one batched risky change we'd have many individual risky changes.

For now I'll push everything to this branch. Once I'm further along I'll decide how to split it up.

## Summary

Not a comprehensive summary. I'll keep updating this section as I work.

### Changes that do not depend on expiration time refactor

- [x] When multiple transitions update the same queue, only the most recent
one should be considered pending. Example: If I switch tabs multiple times, only the last tab I click should display a pending state (e.g. an inline spinner).
  - Only affects `useTransition`. Could land this on its own, though it's most useful when combined with the next item.
- [x] When multiple transitions update the same queue, only the most recent one should be allowed to finish. Do not display intermediate states. Example: if you click on multiple tabs in quick succession, we should not switch to any tab that isn't the last one you clicked.

### Changes that do depend on expiration time refactor

TODO

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -125,6 +125,7 @@ import {
   throwException,
   createRootErrorUpdate,
   createClassErrorUpdate,
+  SuspendOnTask,
 } from './ReactFiberThrow';
 import {
   commitBeforeMutationLifeCycles as commitBeforeMutationEffectOnFiber,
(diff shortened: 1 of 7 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1299,6 +1309,16 @@ function handleError(root, thrownValue) {
         stopProfilerTimerIfRunningAndRecordDelta(workInProgress, true);
       }
 
+      // TODO: I think it's fine to use instanceof here because this exception
+      // is always thrown by the renderer? Or use a brand check.
+      if (thrownValue instanceof SuspendOnTask) {
```
acdlite: @sebmarkbage I assume you won't like this but I legit don't know what the best approach is. A brand check seems unnecessary since this is always thrown by the renderer and not across realms or packages.

acdlite:
For some reason the CodeSandbox CI job didn't run? Perhaps because I opened it as a draft? I'm going to close and re-open to see if that works.

PR #17414, opened 2019-11-21
[Fresh] Fix an infinite loop in an edge case

gaearon:
A good old mutation-while-iterating kind of bug. These maps and sets can be mutated during a commit, so we can't just `forEach` here. Otherwise we risk running into an infinite loop where processing each next item causes more items to be added to the list.

Added a regression test. Verified this fixes a problem on WWW.

### DIFF
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -195,8 +195,17 @@ export function performReactRefresh(): RefreshUpdate | null {
 
     let didError = false;
     let firstError = null;
-    failedRoots.forEach((element, root) => {
-      const helpers = helpersByRoot.get(root);
+
+    // We snapshot maps and sets that are mutated during commits.
+    // If we don't do this, there is a risk they will be mutated while
+    // we iterate over them. For example, trying to recover a failed root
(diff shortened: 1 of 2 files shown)


packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -195,8 +195,17 @@ export function performReactRefresh(): RefreshUpdate | null {
...
     let didError = false;
     let firstError = null;
-    failedRoots.forEach((element, root) => {
-      const helpers = helpersByRoot.get(root);
+
+    // We snapshot maps and sets that are mutated during commits.
+    // If we don't do this, there is a risk they will be mutated while
+    // we iterate over them. For example, trying to recover a failed root
+    // may cause another root to be added to the failed list -- an infinite loop.
+    let failedRootsSnapshot = new Map(failedRoots);
```
nihgwu: is it supposed that `new Map([iterable])` is polyfilled in `react-refresh`

gaearon: Would IE11 be the only one that can’t do it? I guess we could use something longer but tbh I don’t know if we care about supporting IE11 for a developer productivity feature like this.

threepointone:
Ouch, a classic bug.

PR #17418, opened 2019-11-21
useTransition improvements

acdlite:
- [x] When multiple transitions update the same queue, only the most recent
one should be considered pending. Example: If I switch tabs multiple times, only the last tab I click should display a pending state (e.g. an inline spinner).
- [x] When multiple transitions update the same queue, only the most recent one should be allowed to finish. Do not display intermediate states. Example: if you click on multiple tabs in quick succession, we should not switch to any tab that isn't the last one you clicked.

Here's the CodeSandbox I'm using to test the changes: https://codesandbox.io/s/elastic-hawking-69381. It's a tab switcher. Switch tabs and pay attention to the pending spinners.


acdlite:
@CompuIves Any idea why CodeSandbox CI isn't running on this PR?

acdlite:
Closed in favor of https://github.com/facebook/react/pull/17908

PR #17421, opened 2019-11-21
[react-interactions] Fix memory leak in event responder system

trueadm:
This should fix some issues with leaking fibers from root events in the event responder system. Specifically:

- Unsure we unmount listeners on Scope Nodes
- Unsure we memoize the responders map back on fibers


gaearon:
Can you make a small fixture to reproduce the leak?

trueadm:
@gaearon I'll add a regression test for the scopes as that's the main cause of this leak.

PR #17422, opened 2019-11-21
[react-interactions] Refine virtual click detection for FF+JAWS

trueadm:
After lots of internal testing, we found that JAWS with FF results in a case we didn't previously capture as a virtual click. Specifically, the `detail` property is `1` using JAWS but `mozInputSource` is `0`. We actually have this logic internally at FB, so it's reusing that (see https://fburl.com/diffusion/i2aajhpr).


tatermelon:
Thank you!! :)

PR #17427, opened 2019-11-21
fix(ReactElement): display warning for reserved default props

### LINKED ISSUE #17393: Bug: React doesn't warn about key|ref not a prop using defaultProps
bug

React warns about key|ref not being a prop:
```
Warning: Child: `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://fb.me/react-special-props)
```
However, when using `defaultProps` and _not_ providing a key|ref, the key|ref is used as a prop. Test:
```javascript
class Child extends React.Component {
  render() {
    return <div> {this.props.key} </div>;
  }
}
Child.defaultProps = {
  key: "example"
};
class Parent extends React.Component {
  render() {
    return (
      <div>
        <Child /> // <<<<< doesn't warn
        <Child key="1" /> // <<<<< warns
      </div>
    );
  }
}
```
This happens because: 
https://github.com/facebook/react/blob/3ad076472ce9108b9b8a6a6fe039244b74a34392/packages/react/src/ReactElement.js#L362-L384
assigning `defaultProps` doesn't take into account `RESERVED_PROPS`, so this condition `if (key || ref) {...` can be `false`

If `key|ref` is not a prop, it shouldn't map `defaultProps` reserved names to `props`

16.8.6

bpas247:
This displays a warning whenever an end user specifies `key` or `ref` in their default props definition.

Fixes #17393

EDIT: Changed purpose of pull request to not introduce any breaking changes.

### DIFF
--- a/packages/react/src/ReactElement.js
+++ b/packages/react/src/ReactElement.js
@@ -403,13 +403,17 @@ export function createElement(type, config, children) {
   if (type && type.defaultProps) {
     const defaultProps = type.defaultProps;
     for (propName in defaultProps) {
-      if (
-        props[propName] === undefined &&
-        propName !== 'ref' &&
-        propName !== 'key'
-      ) {
+      if (props[propName] === undefined) {
         props[propName] = defaultProps[propName];
       }
+      warningWithoutStack(
+        propName !== 'ref' && propName !== 'key',
+        '%s: %s is not a valid default prop to set. ' +
+          'Please remove this from your default props definition. ' +
+          'Note: the value is still being set, but this behavior will most likely be deprecated in future releases.',
+        type.displayName || type.name || 'Unkown',
+        propName,
+      );
     }
   }
   if (__DEV__) {
--- a/packages/react/src/ReactElement.js
+++ b/packages/react/src/ReactElement.js
@@ -406,14 +406,17 @@ export function createElement(type, config, children) {
       if (props[propName] === undefined) {
         props[propName] = defaultProps[propName];
       }
-      warningWithoutStack(
-        propName !== 'ref' && propName !== 'key',
-        '%s: %s is not a valid default prop to set. ' +
-          'Please remove this from your default props definition. ' +
-          'Note: the value is still being set, but this behavior will most likely be deprecated in future releases.',
-        type.displayName || type.name || 'Unkown',
-        propName,
-      );
+      if (__DEV__) {
+        if (propName === 'ref' || propName === 'key') {
+          console.error(
+            '%s: %s is not a valid default prop to set. ' +
+              'Please remove this from your default props definition. ' +
+              'Note: the value is still being set, but this behavior will most likely be deprecated in future releases.',
+            type.displayName || type.name || 'Unkown',
+            propName,
+          );
+        }
+      }
     }
   }
   if (__DEV__) {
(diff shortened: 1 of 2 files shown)


bpas247:
Hmm, starting to think about it and it's probably valuable to add some tests for this specific use case, even though the diff is small.

packages/react/src/ReactElement.js
```diff
@@ -363,7 +363,11 @@ export function createElement(type, config, children) {
   if (type && type.defaultProps) {
     const defaultProps = type.defaultProps;
     for (propName in defaultProps) {
-      if (props[propName] === undefined) {
+      if (
+        props[propName] === undefined &&
+        propName !== 'ref' &&
```
bpas247: Is it actually an issue for users to define a default prop for refs? I understand why this is a bad practice for keys, since they should be unique. but would this be a bad practice for refs too?

bpas247:
> This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contribution.

Speaking of which, I’m still waiting on code review for this PR 👀

gaearon:
Changing the behavior would be breaking. Can we start by making it a warning if you specify those?

packages/react/src/ReactElement.js
```diff
@@ -366,6 +366,14 @@ export function createElement(type, config, children) {
       if (props[propName] === undefined) {
         props[propName] = defaultProps[propName];
       }
+      warningWithoutStack(
```
gaearon: This helper doesn't exist anymore. Please rebase and use `console.error`.

packages/react/src/ReactElement.js
```diff
@@ -406,6 +406,17 @@ export function createElement(type, config, children) {
       if (props[propName] === undefined) {
         props[propName] = defaultProps[propName];
       }
+      if (__DEV__) {
+        if (propName === 'ref' || propName === 'key') {
+          console.error(
+            '%s: %s is not a valid default prop to set. ' +
```
gaearon, line 412: "is not a default prop to set" sounds confusing. Can you come up with a better wording?

packages/react/src/ReactElement.js
```diff
@@ -406,6 +406,17 @@ export function createElement(type, config, children) {
...
+              'Please remove this from your default props definition. ' +
+              'Note: the value is still being set, but this behavior will most likely be deprecated in future releases.',
```
gaearon, line 414: This is unnecessary. Just asking people to remove it is enough.

packages/react/src/__tests__/ReactElement-test.js
```diff
@@ -358,6 +358,28 @@ describe('ReactElement', () => {
     expect(instance.props.fruit).toBe('persimmon');
   });
 
+  // NOTE: We're explicitly not using JSX here. This is intended to test
+  // classic JS without JSX.
+  it('should display a warning when mapping key to a default prop', () => {
+    class Component extends React.Component {
+      render() {
+        return React.createElement('div', null, this.props.key);
```
gaearon, line 366: Since you render it, you probably want to assert it's in the DOM in your test as well?

bpas247: > Since you render it, you probably want to assert it's in the DOM in your test as well?

Hmm, I don't think rendering the key here adds anything to the test, so I could just not render it.

bpas247:
bump

53jk1 (approved):

ctititic:
i think react18 still have this problem, people can change key or ref by defaultProps with any warning

PR #17430, opened 2019-11-21
[Bugfix] Updates "un-committed" when rebasing

acdlite:
To enforce that updates that are committed can never be un-committed, even during a rebase, we need to track which updates are part of the rebase. The current implementation doesn't do this properly. It has a hidden assumption that, when rebasing, the next `renderExpirationTime` will represent a later expiration time that the original one. That's usually true, but it's not *always* true: there could be another update at even higher priority.

This requires two extra fields on the update queue. I really don't like that the update queue logic has gotten even more complicated. It's tempting to say that we should remove rebasing entirely, and that update queues must always be updated at the same priority. But I'm hesitant to jump to that conclusion — rebasing can be confusing in the edge cases, but it's also convenient. Enforcing single-priority queues would really hurt the ergonomics of useReducer, for example, where multiple values are centralized in a single queue. It especially hurts the ergonomics of classes, where there's only a single queue per class.

So it's something to think about, but I don't think "no more rebasing" is an acceptable answer on its own.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -20,6 +20,7 @@ import type {SuspenseConfig} from './ReactFiberSuspenseConfig';
 import type {ReactPriorityLevel} from './SchedulerWithReactIntegration';
 
 import ReactSharedInternals from 'shared/ReactSharedInternals';
+import {rebasedUpdatesNeverUncommit} from 'shared/ReactFeatureFlags';
 
 import {NoWork} from './ReactFiberExpirationTime';
 import {readContext} from './ReactFiberNewContext';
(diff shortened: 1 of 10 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -755,19 +766,36 @@ function updateReducer<S, I, A>(
...
+      ) {
         // Priority is insufficient. Skip this update. If this is the first
         // skipped update, the previous update/state is the new base
         // update/state.
-        if (!didSkip) {
-          didSkip = true;
+        if (newRebaseTime === NoWork) {
           newBaseUpdate = prevUpdate;
           newBaseState = newState;
+          newRebaseTime = renderExpirationTime;
```
acdlite: Ooh I think this should be the lower of `renderExpirationTime` and the old `rebaseTime`. In case there are two rebases in a row. I'll add a new test.

sebmarkbage: It would also have to be the further at of the RebaseEnd then, no?

acdlite: Actually I don't think this will work. There can be arbitrary numbers of rebases, each with their own ranges and expiration times.

Will need to evaluate.

acdlite:
The approach here won't work because of https://github.com/facebook/react/pull/17430#discussion_r350987813

Closing in favor of https://github.com/facebook/react/pull/17483 stack

PR #17438, opened 2019-11-23
[Fresh] Make all errors recoverable

gaearon:
The logic for recovering from errors was a bit shaky because we tried to guess what was last scheduled from the commit event. That was insufficient because:

1. Failure before the initial mount doesn't make commits at all
2. We can't easily distinguish a user-initiated unmount from a commit that happened for some other reason

The first is a current limitation of Fast Refresh which is annoying. The second leads to some edge cases where it can't reliably recover even on updates.

I'm changing the strategy to introduce a first-class DevTools hook that fires when a root gets scheduled an element. We track these in Fresh runtime in a WeakMap (if not available, retrying is off completely). This lets us recover from both initial and update errors, and also tell reliably what was the last scheduled element.

Because this DevTools hook is new, I had to add some checks. I also cleaned up an old instrumentation file that existed for a similar purpose but is unused. Also added a re-entrancy check to the Fresh runtime. Should be unnecessary but I'm starting to rely on it not being re-entrant.


packages/react-reconciler/src/ReactFiberDevToolsHook.js
```diff
@@ -54,6 +56,25 @@ export function injectInternals(internals: Object): boolean {
...
+      if (typeof hook.onScheduleFiberRoot === 'function') {
+        onScheduleFiberRoot = (root, children) => {
+          try {
+            hook.onScheduleFiberRoot(rendererID, root, children);
+          } catch (err) {
+            if (__DEV__ && !hasLoggedError) {
+              hasLoggedError = true;
+              warningWithoutStack(
+                false,
+                'React DevTools encountered an error: %s',
```
bvaughn: This warning is a bit misleading.

gaearon: It’s just copy paste from other “DevTools hooks”. We can change the message in all of them? I don’t think this one is particularly special other than the fact that it happens to be used by Fresh today. 

Technically there’s nothing preventing DevTools from using this one too — in fact I think it might help us remove some fragile logic in the renderer. (For new versions at least.) In that case we’d need to remove the DEV-only flag check though. 

In either case the issue would be filed in the same repo.

gaearon: In other words, we could apply the same argument to the “commit root” hook which has the same message but is used by both DevTools and Fresh.

bvaughn: I know it was copy pasted from the previous one, but in this particular case it's *always* incorrect.

bvaughn: I don't understand so much pushback against a wording suggestion :smile:

gaearon: I don't think it matters but I'm happy to stamp a PR that changes it :-) I don't mean to "push back", I was only explaining my reasoning for *why* I left it as is. I'm sorry if that explanation isn't useful.

bvaughn: It's cool. Thanks for the follow up PR.

PR #17439, opened 2019-11-23
react-test-renderer: improve findByType() error message

henryqdineen:
I noticed that The `findByType()` method throws unhelpful error messages when using a string as the `type` argument. For example:

```
testInstance.findByType('span');
// Expected 1 but found 2 instances with node type: "undefined"
```
I am also running into this when searching for components using `React.memo`. My solution was to use `getComponentName()`. Thanks!

### DIFF
--- a/packages/react-test-renderer/src/ReactTestRenderer.js
+++ b/packages/react-test-renderer/src/ReactTestRenderer.js
@@ -42,6 +42,7 @@ import {
   ScopeComponent,
 } from 'shared/ReactWorkTags';
 import invariant from 'shared/invariant';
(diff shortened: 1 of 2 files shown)


henryqdineen:
I still think this is a valid fix. Let me know if there is anything else needed from me.

threepointone:
Could you rebase this and add a test for the `Unknown` branch? thanks!

henryqdineen:
Thanks @threepointone. I just rebased and added the suggested test.

threepointone:
Looks good to me, thanks!

jen801 (commented):
>

PR #17455, opened 2019-11-26
Adjust SuspenseList CPU bound heuristic

sebmarkbage:
In SuspenseList we switch to rendering fallbacks (or stop rendering further rows in the case of tail="collapsed/hidden") if it takes more than 500ms to render the list. The limit of 500ms is similar to the train model and designed to be short enough to be in the not noticeable range.

This works well if each row is small because we time the 500ms range well. However, if we have a few large rows then we're likely to exceed the limit by a lot. E.g. two 480ms rows hits almost a second instead of 500ms.

This PR adjusts the heuristic to instead compute whether something has expired based on the render time of the last row. I.e. if we think rendering one more row would exceed the timeout, then we don't attempt.

This still works well for small rows and bails earlier for large rows.

The expiration is still based on the start of the list rather than the start of the render. It should probably be based on the start of the render but that's a bigger change and needs some thought.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -1114,6 +1114,8 @@ function completeWork(
               return null;
             }
           } else if (
+            // The time it took to render last row is greater than time until
+            // the expiration.
             now() * 2 - renderState.renderingStartTime >
               renderState.tailExpiration &&
             renderExpirationTime > Never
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -1114,7 +1114,8 @@ function completeWork(
               return null;
             }
           } else if (
-            now() > renderState.tailExpiration &&
+            now() * 2 - renderState.renderingStartTime >
```
acdlite: Inline comment that translates this into English plz

"The time it took to render last row is greater than time until the expiration"

PR #17456, opened 2019-11-26
Don't group Idle/Offscreen work with other work

sebmarkbage:
When we suspend we always try a lower level but we shouldn't try offscreen.

If nothing pings during this time then we’ll never return back to rendering the smaller set. This leads us to overrender a lot and assume offscreen heuristics for visible content.

Offscreen really shouldn’t ever render unless we’ve committed everything else. There’s no reason to since hidden and hydrating work never unblocks anything.

This change makes it so we leave cpu cycles on the table. Ideally we could warm up hidden work and then return to the previously prepared tree but since we can’t do that yet it’s better to not throw away the tree.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -25,6 +25,7 @@ import {
   warnAboutUnmockedScheduler,
   flushSuspenseFallbacksInTests,
   disableSchedulerTimeoutBasedOnReactExpirationTime,
+  enableTrainModelFix,
 } from 'shared/ReactFeatureFlags';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import invariant from 'shared/invariant';
@@ -539,9 +540,19 @@ function getNextRootExpirationTimeToWorkOn(root: FiberRoot): ExpirationTime {
   // on whichever is higher priority.
   const lastPingedTime = root.lastPingedTime;
   const nextKnownPendingLevel = root.nextKnownPendingLevel;
-  return lastPingedTime > nextKnownPendingLevel
-    ? lastPingedTime
-    : nextKnownPendingLevel;
+  const nextLevel =
+    lastPingedTime > nextKnownPendingLevel
+      ? lastPingedTime
+      : nextKnownPendingLevel;
+  if (
+    enableTrainModelFix &&
+    nextLevel <= Idle &&
+    firstPendingTime !== nextLevel
+  ) {
+    // Don't work on Idle/Never priority unless everything else is committed.
+    return NoWork;
+  }
+  return nextLevel;
 }
 
 // Use this function to schedule a task for a root. There's only one task per
@@ -2362,7 +2373,7 @@ export function pingSuspendedRoot(
   // Mark the time at which this ping was scheduled.
   root.lastPingedTime = suspendedTime;
 
-  if (root.finishedExpirationTime === suspendedTime) {
+  if (!enableTrainModelFix && root.finishedExpirationTime === suspendedTime) {
     // If there's a pending fallback waiting to commit, throw it away.
     root.finishedExpirationTime = NoWork;
     root.finishedWork = null;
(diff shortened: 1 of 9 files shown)


sebmarkbage:
I think we’ll also have to do something about `pingSuspendedRoot` because otherwise a ping on Offscreen could start rendering at Offscreen priority while the other ones are suspended.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -866,8 +873,14 @@ function finishConcurrentRender(
         }
 
         const nextTime = getNextRootExpirationTimeToWorkOn(root);
-        if (nextTime !== NoWork && nextTime !== expirationTime) {
-          // There's additional work on this root.
+        if (
+          nextTime !== NoWork &&
+          nextTime !== expirationTime &&
+          (!enableTrainModelFix || nextTime > Idle)
```
acdlite: Could put this check into `getNextRootExpirationTimeToWorkOn` instead? Then it will work with `pingSuspendedRoot`, too.

acdlite: More specifically, `getNextRootExpirationTimeToWorkOn` should never return `Never` if the root is suspended, which we can check by looking at `root.firstSuspendedTime` and `root.lastSuspendedTime`.

sebmarkbage: Are firstSuspendedTime and lastSuspendedTime reliable for this type of check? I think they just say that at some point this time was suspended, but it doesn't say that there is still work left on that level (e.g. if the updated component was deleted).

Also `pingSuspendedRoot` doesn't call `getNextRootExpirationTimeToWorkOn` so it wouldn't fix the issue that the finished work gets blown away.

acdlite: We clear `firstSuspendedTime` and `lastSuspendedTime` whenever we commit, so I think it is reliable for this check. Those fields are also how `pingSuspendedRoot` bails out of a ping early, if the pinged time has already finished.

> Also `pingSuspendedRoot` doesn't call `getNextRootExpirationTimeToWorkOn`

It calls `ensureRootIsScheduled` which in turn calls `getNextRootExpirationTimeToWorkOn`. All render passes begin by calling `getNextRootExpirationTimeToWorkOn` so  if you fix it there then you fix it for all cases.

acdlite: To be clear I think we'd still need to remove those lines in `pingSuspendedRoot`. Might be able to delete them? I think they are outdated.

sebmarkbage: I don't think it's that simple. Currently, `Never` can suspend. We also set `lastPingedTime` to `Never` to unblock such work. When we do that, we've lost whether that root was also pinged at a higher priority than that. So if `lastPingedTime` is `Never` then we can't safely return `NoWork` because it could've pinged at higher priority.

sebmarkbage: Actually I can’t seem to write a failing test for this.

If we start another render at higher pri then that unsuspends the Never pri so when the ping comes in it’s no longer suspended and so lastPingedTime doesn’t update.

Maybe this isn’t an issue after all.

That said, Never probably should never suspend the commit.

sebmarkbage: Actually I can’t seem to write a failing test for this.

If we start another render at higher pri then that unsuspends the Never pri so when the ping comes in it’s no longer suspended and so lastPingedTime doesn’t update.

Maybe this isn’t an issue after all.

That said, Never probably should never suspend the commit.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -539,9 +540,19 @@ function getNextRootExpirationTimeToWorkOn(root: FiberRoot): ExpirationTime {
...
+  const nextLevel =
+    lastPingedTime > nextKnownPendingLevel
+      ? lastPingedTime
+      : nextKnownPendingLevel;
+  if (
+    enableTrainModelFix &&
+    nextLevel <= Idle &&
+    firstPendingTime !== nextLevel
+  ) {
+    // Don't work on Idle/Never priority unless everything else is committed.
```
sebmarkbage: I'm a bit uneasy about this since if something goes wrong, it's not possible to unblock by pinging these levels. However, I don't *think* it ever happens now because these levels wouldn't ever be suspended if something else rendered at higher pri because that makes them unsuspended. I couldn't make a test that fails.

PR #17469, opened 2019-11-27
Remove workInProgressRootLatestSuspenseTimeout

lunaruan:
Remove `workInProgressRootLatestSuspenseTimeout` field from the work loop in preparation for splitting timeout from expiration time.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -232,7 +232,6 @@ let workInProgressRootFatalError: mixed = null;
 // because we deal mostly with expiration times in the hot path, so this avoids
 // the conversion happening in the hot path.
 let workInProgressRootLatestProcessedExpirationTime: ExpirationTime = Sync;
-let workInProgressRootLatestSuspenseTimeout: ExpirationTime = Sync;
 let workInProgressRootCanSuspendUsingConfig: null | SuspenseConfig = null;
 // The work left over by components that were visited during this render. Only
 // includes unprocessed updates, not work in bailed out children.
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -882,11 +881,21 @@ function finishConcurrentRender(
...
+        if (workInProgressRootCanSuspendUsingConfig !== null) {
           // We have processed a suspense config whose expiration time we
           // can use as the timeout.
-          msUntilTimeout =
-            expirationTimeToMs(workInProgressRootLatestSuspenseTimeout) - now();
+          if (
+            typeof workInProgressRootCanSuspendUsingConfig.busyDelayMs ===
+              'number' &&
+            typeof workInProgressRootCanSuspendUsingConfig.busyMinDurationMs ===
+              'number'
```
sebmarkbage: These numbers don’t have anything to do with the timeout.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1364,14 +1372,8 @@ export function markRenderEventTimeAndConfig(
     workInProgressRootLatestProcessedExpirationTime = expirationTime;
   }
   if (suspenseConfig !== null) {
-    if (
-      expirationTime < workInProgressRootLatestSuspenseTimeout &&
-      expirationTime > Idle
```
sebmarkbage: This condition seems dropped on the floor, why?

acdlite: Yeah I would guess this branch needs to stay as it was except replace `workInProgressRootSuspenseLatestTimeout` with `workInProgressRootCanSuspendUsingConfig.timeoutMs`.

Issue #17084, opened 2019-10-14
Devtools scrollbars should be dark in dark mode


Daniel15:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When Chrome Devtools are in "dark mode", all the standard panels have a dark scrollbar:
![](https://d.sb/2019/10/chrome_14-11.53.42.png)

However, the React Devtools still show a light scrollbar:
![](https://d.sb/2019/10/chrome_14-11.55.16.png)

**What is the expected behavior?**
Scrollbars should be styled similarly to the standard panels

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
Any React version
Chrome 77.0
Windows 10, version 1903.

Note: How to enable dark mode:
![](https://d.sb/2019/10/chrome_14-11.58.25.png)
![](https://d.sb/2019/10/chrome_14-11.59.29.png)

eoandersson:
I'd love to give this one a look. That is, if no one else is already working on it.

Daniel15:
@eoandersson - Sure! I don't think anyone is working on this at the moment. The readme has some instructions on how to build the devtools: https://github.com/facebook/react/tree/master/packages/react-devtools-extensions. Let us know if you get stuck 😄

eoandersson:
@Daniel15 - I've got a solution working using the `::-webkit-scrollbar` pseudo element. However, I have a few quick questions. 😄

1. I'm a little uncertain as to where I should place my new `::-webkit-scrollbar` CSS. I've  currently opted to place my new lines in `react-devtools-shared/src/devtools/views/DevTools.css`, though I've also considered the possibility of other locations, such as `react-devtools-shared/src/devtools/views/root.css`. What would be the ideal solution here?

2. In terms of styling, are there any guidelines to follow? For reference as to how the solution looks now, see the image below.

![Scrollbar-Themes](https://user-images.githubusercontent.com/24208371/66872404-34be3100-efa6-11e9-8d39-4e9b2138dfb4.png)

Daniel15:
> I'm a little uncertain as to where I should place my new ::-webkit-scrollbar CSS.

I'm not familiar with the devtools code so I'm not sure which file is best. @bvaughn might have a suggestion here. Otherwise, just place it wherever you feel is a decent place, and we can comment on it during code review 😃 

> In terms of styling, are there any guidelines to follow?

[...]

PR #17471, opened 2019-11-27
Fix appearance of devtool scrollbar in darkmode

eoandersson:
Fixed an issue where the devtool scrollbar didn't change its appearance based 
on light/dark mode.

[Issue #17084 ]

It should be noted that for some users the scrollbar is already switching between 
dark and light mode. However, this appears to be caused by either a Chrome or 
OS version, and as such the scrollbar differs from user to user. This change would 
ensure that all users have the same scrollbar and that it properly switches color 
based on the user's selected theme.

I've tried to maintain a similar shape and size to the original Chrome devtools 
scrollbar while using the already defined scrollbar color variables in React 
(Already used for Firefox).

---

_Chrome Devtools scrollbar_
ChromeScrollbar

_New React dark mode scrollbar_
ReactScrollbarDark

_New React light mode scrollbar_
ReactScrollbarLight

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/DevTools.css
+++ b/packages/react-devtools-shared/src/devtools/views/DevTools.css
@@ -52,3 +52,17 @@
     display: none;
   }
 }
+
+::-webkit-scrollbar-thumb  {
+  background-color: var(--color-scroll-thumb);
+  border-radius: 2px;
+}
+
+::-webkit-scrollbar-track {
+  background-color: var(--color-scroll-track);
+  border-left: 1px solid var(--color-border);
+}
+
+::-webkit-scrollbar {
+  width: 14px;
+}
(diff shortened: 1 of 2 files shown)


bvaughn:
Thank you for putting together this PR!

Unfortunately I don't think we should move forward with it, because it has a pretty significant (negative) impact on the default experience for Mac users.

For example, without this change- this is the UI I see in Chrome:
Screen Shot 2019-12-10 at 11 04 25 AM

With this change applied I see this in Chrome:
Screen Shot 2019-12-10 at 11 04 13 AM

Basically I've lost a combined total of 50px horizontal space to scroll bars that don't need to be there.

Especially given the fact that the bug reported in #17084 is minor/cosmetic and doesn't impact everyone, I don't think the fix justifies the downsides of lost screen space.

PR #17480, opened 2019-11-28
Use fewer global variables in Hooks

sebmarkbage:
Module level variables is an exception that we should use sparingly. It will make using multiple threads in a threaded implementation infeasible, harder to port to stricter languages, it's harder to reason about the performance and it is kind of hard to reason about whether these are allowed to diverge from the main state.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -187,8 +187,6 @@ let firstWorkInProgressHook: Hook | null = null;
 let workInProgressHook: Hook | null = null;
 let nextWorkInProgressHook: Hook | null = null;
 
-let remainingExpirationTime: ExpirationTime = NoWork;
-
 // Updates scheduled during render will trigger an immediate re-render at the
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -433,16 +420,25 @@ export function renderWithHooks(
...
   let children = Component(props, refOrContext);
 
   if (didScheduleRenderPhaseUpdate) {
+    // Counter to prevent infinite loops.
+    let numberOfReRenders: number = 0;
     do {
       didScheduleRenderPhaseUpdate = false;
+
+      invariant(
+        numberOfReRenders < RE_RENDER_LIMIT,
```
acdlite: The stack on this error is less useful now

sebmarkbage: In the commit I mentioned that we could add a DEV warning to help this. However, I think I could do a follow up of this that uses the dispatcher to error instead in this: https://github.com/facebook/react/pull/17484

PR #17483, opened 2019-11-28
Hooks Refactor Update Queue by Cloning Rebased Updates

sebmarkbage:
Builds on top of #17480.

Alternative to #17430.

This model splits the update queue into two (cyclic) linked lists. There is one each for current/workInProgress pair for rebase purposes. There is also one shared one.

When we process a queue we merge the shared onto the "current" one, and clone it onto the work in progress.

The cloned one doesn't include the beginning of the queue that is fully processed.

The cloned nodes that were processed get Sync priority so that they don't get dropped during future renders.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -20,7 +20,7 @@ import type {ReactPriorityLevel} from './SchedulerWithReactIntegration';
 
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 
-import {NoWork} from './ReactFiberExpirationTime';
+import {NoWork, Sync} from './ReactFiberExpirationTime';
 import {readContext} from './ReactFiberNewContext';
 import {createResponderListener} from './ReactFiberEvents';
 import {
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -689,94 +676,117 @@ function updateReducer<S, I, A>(
...
+      let baseFirst = baseQueue.next;
+      let pendingFirst = pendingQueue.next;
+      baseQueue.next = pendingFirst;
+      pendingQueue.next = baseFirst;
     }
-    first = baseUpdate.next;
-  } else {
-    first = last !== null ? last.next : null;
+    current.baseQueue = baseQueue = pendingQueue;
+    queue.pending = null;
```
acdlite: I love this trick

sebmarkbage:
We might want to add an optimization to avoid cloning the queue when something else renders at higher pri. In that case it’s unnecessary to clone it.

packages/react-reconciler/src/ReactFiberHooks.js
sebmarkbage: I believe we could use this same trick for expirationTimes. By storing it temporarily on a shared stateful object and then transferring it to “current”. That way we don’t need the alternate model when we backtrack and can just backtrack along on a shared object path.

acdlite: Yeah I used a similar trick for the eager bailout optimization. Where I reset current.expirationTime to NoWork after a bailout. Also in my useTransition PR to clear a pending transition. Such a nice strategy.

acdlite:
I squashed this with the class refactor and landed as https://github.com/facebook/react/commit/b617db3d966f678eb0b4aac6d96f7967b37a9e91 so it can be atomically reverted.

PR #17484, opened 2019-11-28
Remove renderPhaseUpdates Map

sebmarkbage:
Builds on top of #17483

Now that we have a separate queue for recently added updates, we can use this queue to pick up the render phase updates. As long as we clear them before committing.

TODO: I believe this might have a bug if you suspend during the second update but before we get to process this particular Hook. Since the update would remain in the queue in this case.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -504,14 +504,31 @@ export function bailoutHooks(
   }
 }
 
-export function resetHooks(): void {
+export function resetHooksAfterThrow(): void {
   // We can assume the previous dispatcher is always this one, since we set it
   // at the beginning of the render phase and there's no re-entrancy.
   ReactCurrentDispatcher.current = ContextOnlyDispatcher;
 
-  // This is used to reset the state of this module when a component throws.
-  // It's also called inside mountIndeterminateComponent if we determine the
-  // component is a module-style component.
+  if (didScheduleRenderPhaseUpdate) {
+    const current = (currentlyRenderingFiber: any).alternate;
+    if (current !== null) {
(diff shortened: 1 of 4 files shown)


sebmarkbage:
I should also make the infinite loop error use a dispatcher: https://github.com/facebook/react/pull/17480#discussion_r353366838

acdlite:
> TODO: I believe this might have a bug if you suspend during the second update but before we get to process this particular Hook. Since the update would remain in the queue in this case.

Inside `resetHooks`, which is called when something throws, scan through the partial list of work-in-progress hooks and clear the pending updates (`queue.pending = null`). We only need to do this if a re-render was scheduled.

acdlite:
Fixed the todo and merged after discussing out of band.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -508,14 +504,31 @@ export function bailoutHooks(
...
+export function resetHooksAfterThrow(): void {
   // We can assume the previous dispatcher is always this one, since we set it
   // at the beginning of the render phase and there's no re-entrancy.
   ReactCurrentDispatcher.current = ContextOnlyDispatcher;
 
-  // This is used to reset the state of this module when a component throws.
-  // It's also called inside mountIndeterminateComponent if we determine the
-  // component is a module-style component.
+  if (didScheduleRenderPhaseUpdate) {
+    const current = (currentlyRenderingFiber: any).alternate;
```
sebmarkbage: @acdlite What if this is render phase updates during initial mount?

acdlite: It doesn't matter then because either the whole fiber will get thrown out. Except in legacy mode goddamn it. I'll add a test in the other PR.

Issue #17476, opened 2019-11-27
Rendering a `String` class instance in an expression renders each letter as a text node


subyraman:
**Do you want to request a *feature* or report a *bug*?**

Report a bug

**What is the current behavior?**

If you attempt to render a `String`  in an expression, it will render each letter as a child node.

From an accessibility standpoint, this does matter. A screen reader like Apple's VoiceOver utility will attempt to read every text node, meaning that it will attempt to read each letter. This would mean an impacted experience for blind users. 

From a display perspective, this doesn't matter, because browsers will group the nodes next to each other, inline. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

https://codesandbox.io/s/string-constructor-issue-f5t5x

- Open the sandbox in a new window. 
- Examine the DOM. See that text nodes are created for each letter from the rendered `String` instance. 
- Open Apple's VoiceOver and read the page. Hear that all letters in the rendered `String` instance are spelled out, instead of the words being spoken. 

**What is the expected behavior?**

I would expect a `String` instance to be rendered the same as a `string` literal type. 

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

Happening on React DOM 16.8.6, Chrome 80.0, Mac OSX

kunukn:
The issue of rendering this.
```jsx
<p>{new String("Text inside of a string constructor")}</p>
```

Can be fixed using toString
```jsx
<p>{new String("Text inside of a string constructor").toString()}</p>
```

With given example, the Codesandbox shows a linter warning.
`Do not use String as a constructor.`

It is considered bad practice to use String as a constructor.

> There's almost never any reason to create the object version of any of them explicitly, and doing so can indeed lead to confusion

https://stackoverflow.com/a/41458207/815507

Is there a good reason to use string constructor for text rendering?
If there is, maybe we could use the toString method to solve this issue?

subyraman:
The application I work on uses extensions of the `String` type in order to allow for better auditing and control flow of our application code in TypeScript. `toString()` is not an option because type information would be erased. 

The linter opinion aside, the behavior of React in this case is unexpected and surprising. We had no reason to expect for it to render out child nodes for each letter.

[...]

PR #17493, opened 2019-11-29
Fix string object rendering

kunukn:
This PR fixes: #17476.

This is an example that fixes the issue. 

By design, it seem as rendering Objects are not allowed.
Typical render warning for objects. `Objects are not valid as a React child`

But because a string object is iterable, it get rendered with `reconcileChildrenIterator`

This has been used for inspiration of how to check for string
https://stackoverflow.com/q/4059147/815507
I could have used `newChild instanceof String` but thought that might be too expensive to use.

This PR works because `toString()` is applied by the existing code snipped.
`'' + newChild,`

### DIFF
--- a/packages/react-reconciler/src/ReactChildFiber.js
+++ b/packages/react-reconciler/src/ReactChildFiber.js
@@ -482,7 +482,11 @@ function ChildReconciler(shouldTrackSideEffects) {
     newChild: any,
     expirationTime: ExpirationTime,
   ): Fiber | null {
-    if (typeof newChild === 'string' || typeof newChild === 'number' || newChild instanceof String) {
+    if (
+      typeof newChild === 'string' ||
+      typeof newChild === 'number' ||
+      newChild instanceof String
+    ) {
       // Text nodes don't have keys. If the previous node is implicitly keyed
       // we can continue to replace it without aborting even if it is not a text
       // node.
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -482,7 +482,11 @@ function ChildReconciler(shouldTrackSideEffects) {
     newChild: any,
     expirationTime: ExpirationTime,
   ): Fiber | null {
-    if (typeof newChild === 'string' || typeof newChild === 'number') {
+    if (
+      typeof newChild === 'string' ||
+      typeof newChild === 'number' ||
+      (newChild && typeof newChild.valueOf() === 'string')
```
aweary: It seems unlikely that we'd want to add the overhead of calling `valueOf` in this code path. Especially because `valueOf` can end up executing arbitrary user code.

threepointone: Yeah I'm certain we wouldn't accept this PR for this edge case. Would recommend instead to call toString() on the object in app code, or use regular strings. The issues with typescript, should probably be fixed at the types level, not in react.

threepointone:
Going to close this because I don't think this is the right 'fix' to include.

packages/react-reconciler/src/ReactChildFiber.js
kunukn: Sounds logical to me.

For reference, here is a much better string checker.
https://github.com/lodash/lodash/blob/master/isString.js

>Is there a way to just special case String so it doesn't hit reconcileChildrenIterator?

Probably possible, could be added in getIteratorFn?

```jsx
export function getIteratorFn(maybeIterable: ?any): ?() => ?Iterator<*> {
  if (maybeIterable === null || typeof maybeIterable !== 'object') {
    return null;
  }
  if (maybeIterator instanceof String) { // <--- ?
    return null;
  }
...
```

PR #17496, opened 2019-12-01
add 'rollup-plugin-json-parse'

tjenkinson:
This adds [rollup-plugin-json-parse](https://github.com/tjenkinson/rollup-plugin-json-parse), which is configured to convert large objects to strings that are `JSON.parse`'d. This [should actually increase performance](https://github.com/tjenkinson/rollup-plugin-json-parse#why)

[Click here to see the diff of the build when the plugin is added](https://github.com/facebook/react/commit/ce13149c788706ed5a57213c071e8ebe7c8555f6)

Interesting to see that the real size goes up slightly, but the gzipped size is actually smaller.

Would be interesting if to see if it actually makes a speed difference here. Do you have any benchmarks that could be run?

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "rollup": "^0.52.1",
     "rollup-plugin-babel": "^4.0.1",
     "rollup-plugin-commonjs": "^8.2.6",
+    "rollup-plugin-json-parse": "^1.1.4",
     "rollup-plugin-node-resolve": "^2.1.1",
     "rollup-plugin-prettier": "^0.3.0",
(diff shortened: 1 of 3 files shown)


threepointone:
I think we actually tried this internally, and didn't see any wins. It's unlikely we'll add this dependency unless there's a clear performance difference. 

It's still an option for people to add this to their own application builds ofc, and include react in the configuration to run this transform, so we'd probably recommend people do it so instead of doing it as part of the react build. 

Closing this for now, but happy to reopen/discuss if you can show a performance win on a benchmark.

PR #17510, opened 2019-12-03
Refactor legacy update queue

acdlite:
## Based on #17483

Refactors legacy update queue to incorporate rebasing fix. Uses nearly the same approach as the hook update queue but has to handle a few other cases.

Managed to remove some older, poorly factored code related to error handling.

## To-do

- [x] One remaining unit test to fix
- [x] Fix Flow

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -20,7 +20,7 @@ import type {ReactPriorityLevel} from './SchedulerWithReactIntegration';
 
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 
-import {NoWork} from './ReactFiberExpirationTime';
+import {NoWork, Sync} from './ReactFiberExpirationTime';
 import {readContext} from './ReactFiberNewContext';
 import {createResponderListener} from './ReactFiberEvents';
 import {
@@ -108,13 +108,13 @@ type Update<S, A> = {
   action: A,
   eagerReducer: ((S, A) => S) | null,
   eagerState: S | null,
-  next: Update<S, A> | null,
+  next: Update<S, A>,
 
   priority?: ReactPriorityLevel,
 };
 
 type UpdateQueue<S, A> = {
-  last: Update<S, A> | null,
+  pending: Update<S, A> | null,
   dispatch: (A => mixed) | null,
   lastRenderedReducer: ((S, A) => S) | null,
   lastRenderedState: S | null,
@@ -144,7 +144,7 @@ export type Hook = {
   memoizedState: any,
 
   baseState: any,
-  baseUpdate: Update<any, any> | null,
+  baseQueue: Update<any, any> | null,
   queue: UpdateQueue<any, any> | null,
 
   next: Hook | null,
@@ -544,8 +544,8 @@ function mountWorkInProgressHook(): Hook {
     memoizedState: null,
 
     baseState: null,
+    baseQueue: null,
     queue: null,
-    baseUpdate: null,
 
     next: null,
   };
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1299,11 +1299,7 @@ function dispatchAction<S, A>(
       // This is the first update. Create a circular list.
       update.next = update;
     } else {
-      const first = pending.next;
-      if (first !== null) {
-        // Still circular.
-        update.next = first;
```
sebmarkbage: What's this about, we need this for it to stay circular, no?

acdlite: All I did was remove the `first !== null` check. Which is unnecessary because `update.next` is never null.

acdlite: I think the diff makes it look like I deleted the assignment but really all I deleted was the type check that surrounds it.

acdlite: Actually, to rephrase, it's not really a type check that I deleted. In the old implementation, the queue was only sometimes circular. It would eventually "unravel" to linear so that the old nodes could be garbage collected. But in the new implementation, the pending queue is *always* circular. So you don't need to check for if it's circular or not.

packages/create-subscription/src/__tests__/createSubscription-test.internal.js
```diff
@@ -160,7 +160,7 @@ describe('createSubscription', () => {
 
     it('should still work if unsubscription is managed incorrectly', async () => {
       const Subscription = createSubscription({
-        getCurrentValue: source => undefined,
+        getCurrentValue: source => source.value,
```
sebmarkbage: Is this a breaking change?

acdlite: I looked into this for a while and I think the old version was just wrong and it accidentally worked with the old implementation. I'm going to try to write a failing test for what's in master.

packages/react-reconciler/src/ReactUpdateQueue.js
```diff
@@ -117,27 +117,21 @@ export type Update<State> = {
...
-  lastCapturedUpdate: Update<State> | null,
-
-  firstEffect: Update<State> | null,
-  lastEffect: Update<State> | null,
-
-  firstCapturedEffect: Update<State> | null,
-  lastCapturedEffect: Update<State> | null,
+  baseQueue: Update<State> | null,
+  shared: SharedQueue<State>,
+  effects: Array<Update<State>> | null,
```
sebmarkbage: Why is this an array as opposed to a linked list like the hooks form? Harder to see how they correlate.

I was considering a refactor to arrays but with a different take (inline fields instead of objects) . However, I wanted to do that as a separate refactor.

acdlite: Idk I just chose one :D

This felt like a different trade off from the hooks one because these effects are relatively rare. Only when you pass a callback to `setState`. It's more like the `emitEffect` API.

acdlite: So it didn't seem worth putting a `nextEffect` field on the update type

packages/create-subscription/src/__tests__/createSubscription-test.internal.js
acdlite: Never mind, I found the bug. Will add a test.

acdlite:
I squashed this with the hooks refactor and landed as b617db3 so it can be atomically reverted.

PR #17513, opened 2019-12-03
[react-interactions] Upgrade passive event listeners to active listeners

trueadm:
One of the long standing things we've wanted to do when working on React Flare, was to be able to upgrade passive event listeners to active event listeners where possible. This removes the need to have two separate event listeners attached to the document at the same time.

If we first attach an active listener of a given type, and then later try to add a passive listener of the same type, we don't both – instead we handle the logic for both active and passive events inside the same active event listener. Furthermore, if we first attach a passive event for a given type, and then later try to add an active listener of the same type, we first remove the passive listener from the DOM and then we add the active listener to the DOM.

The logic in the responder system might seem a little convoluted because of how responders target/root events are designed. Thankfully, this won't be the case with the Listener API and we use a much simpler control flow when dealing with events with that API.

I also remove the responder logic for `passiveSupported` as it no longer felt relevant given this change, plus it was never used in any event responders. To enable the changes, we change the Listener Set to a Listener Map, so we can store the event listener callback we used when setting up the DOM event listener (this enables us to remove it again later).

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -14,7 +14,6 @@ import warning from 'shared/warning';
 import {canUseDOM} from 'shared/ExecutionEnvironment';
 import warningWithoutStack from 'shared/warningWithoutStack';
 import endsWith from 'shared/endsWith';
-import type {DOMTopLevelEventType} from 'legacy-events/TopLevelEventTypes';
 import {setListenToResponderEventTypes} from '../events/DOMEventResponderSystem';
 
 import {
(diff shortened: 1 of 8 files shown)


packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1307,12 +1309,12 @@ export function restoreControlledState(
 
 export function listenToEventResponderEventTypes(
   eventTypes: Array<string>,
-  element: Element | Document,
+  document: Document,
 ): void {
   if (enableFlareAPI) {
     // Get the listening Set for this element. We use this to track
```
necolas: It's not a `Set` anymore

Issue #16589, opened 2019-08-27
Confusing error when passing non-function to useMemo


aweary:
[Example](https://codesandbox.io/s/boring-albattani-grhhp).

When passing something like an object to `useMemo` the error isn't very helpful.

```
TypeError: nextCreate is not a function
    at mountMemo (https://grhhp.csb.app/node_modules/react-dom/cjs/react-dom.development.js:13459:19)
    ...
```

Most developers wont know what `nextCreate` is, and passing an object to `useMemo` doesn't seem _obviously_ wrong at first. We could add a `typeof` check that throws a better error, but maybe that's too much overhead? At the very least we could add a DEV warning with more helpful guidance.

threepointone:
Huh. This is usually caught by flow/typescript, but I can see why it might be confusing without a type system. Do you believe this is a common problem? If so, maybe a DEV warning would be good.

Tevinthuku:
@threepointone , If you resolve to have a dev warning for this, I could work on that

dhrubesh:
@threepointone Mind if I take this up?

duanebester:
@aweary I have an incoming PR for this

caspercasanova:
Why is this still open?

jeanpierrecarvalho:
> Why is this still open?

@duanebester

duanebester:
> Add DEV warning for when a non-function is passed to useMemo #17529

No one ever approved my PR, so I closed it

PR #17529, opened 2019-12-04
Add DEV warning for when a non-function is passed to useMemo

duanebester:
Before, the following would throw a TypeError for createNext() not being a function:

```
function App() {
  const memoized = React.useMemo({ foo: "bar" }, []);
  return <div>{memoized.foo}</div>;
}
```

Now we have a DEV warning that's more helpful:

```
Warning: useMemo received an argument that is not a function (instead, received `object`). When specified, this argument must be a function.
    in App (at src/index.js:7)
```

Addresses #16589

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -257,6 +257,23 @@ function checkDepsAreArrayDev(deps: mixed) {
   }
 }
 
+function checkArgumentIsFunctionDev(arg: mixed) {
+  if (__DEV__) {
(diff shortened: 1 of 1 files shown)


duanebester:
Should be a quick one @acdlite @aweary @philipp-spiess

bl00mber:
Hey @duanebester can you please tell how you have requested a review not being a member at the same time? :thinking:

duanebester:
@bl00mber one day github just gave me the option to request

bl00mber:
@duanebester that's interesting, thanks for answer!

PR #17536, opened 2019-12-05
Replace `wrap-warning-with-env-check` with an eslint plugin

walaura:
Part 1.5 of #16753

This PR removes the `__DEV__` wrapping part of `wrap-warning-with-env-check` and replaces it with a functionally identical eslint plugin (`no-production-logging`).The babel rule gets renamed to `lift-warning-conditional-argument` because that's what it does! In its current stage this will fail builds since we still got all the unwrapped calls in the actual codebase.

The plugin supports `&&`statements like `if( __DEV__ && React.suspenseIsEnabled)` but fails any top level `||`s. Have a look at the tests and code because I might have some unchecked assumptions about static code analysis here (and about how we strip out the `__DEV__` checks at compile time lol) but i think this covers the cases we wanna cover.

## Caveats

The generated development builds will include an extra block around the checks, from this original code
```jsx
if(suspenseEnabled){
  warning(2===2, 'example warning');
  ...
}
```

eslint (in this branch) will make the editor rewrite it to
```jsx
if(suspenseEnabled){
  if(__DEV__){
    warning(2===2, 'example warning');
  }
  ...
}
```

Which babel will transform (using ternaries) to
```jsx
if(suspenseEnabled){
  if(__DEV__){
    2===2 ? warning(false, 'example warning') : void 0;
  }
  ...
}
```

And then rollup will rewrite `__DEV__` to true (for a dev bundle)
```jsx
if(suspenseEnabled){
  if(true){
    2===2 ? warning(false, 'example warning') : void 0;
  }
  ...
}
```

And finally, commonjs remove the useless if (this seems like a happy side effect rather than something we are doing on purpose for the dev builds - as for prod we run the closure compiler). However it will keep the block scope:

```jsx
if(suspenseEnabled){
  {
    2===2 ? warning(false, 'example warning') : void 0;
  }
  ...
}
```

This makes sense to me, _this_ block scope is pretty absurd but we can't remove blocks automatically (at least not naively) because in some cases it _can_ affect the scope of variables.

We didn't have this before because babel would transform everything using ternaries, which don't start a block scope. Since we are moving the __DEV__ check to the userland now I don't think we want people writing this manually:

```jsx
if(suspenseEnabled){
    __DEV__ ? 2===2 ? warning(false, 'example warning') : void 0 : void 0;
  ...
}
```

## Next steps

How do we merge this? The way I see it we can either 

- Set the rule to warn, progressively fix the warnings. Don't like this because we don't have any rules set to warn? I might be wrong here? help??
- Prepare an eslint fix on this PR and apply it before merging. The builds will become green.


scripts/eslint-rules/no-production-logging.js
```diff
@@ -0,0 +1,75 @@
...
+      ) {
+        return true;
+      }
+    }
+  }
+
+  function report(node) {
+    context.report({
+      node: node,
+      message: 'Wrap {{identifier}} in a `if (__DEV__)` check',
```
philipp-spiess: Maybe add why so when new contributors see them they learn something? I.e "to ensure that warnings are omitted in production builds".

scripts/babel/wrap-warning-with-env-check.js
```diff
@@ -1,67 +0,0 @@
...
-            path.get('callee').isIdentifier({name: 'warningWithoutStack'})
-          ) {
-            // Turns this code:
-            //
-            // warning(condition, argument, argument);
-            //
-            // into this:
-            //
-            // if (__DEV__) {
-            //   if (!condition) {
```
philipp-spiess: This part of the babel plugin is no longer in the ESLint rule causing the dev packages to have the following diff:

```diff
- !(typeof trustedTypes === 'undefined') ? warning$1(false, "Using 'dangerouslySetInnerHTML' in an svg element with " + 'Trusted Types enabled in an Internet Explorer will cause ' + 'the trusted value to be converted to string. Assigning string ' + "to 'innerHTML' will throw an error if Trusted Types are enforced. " + "You can try to wrap your svg element inside a div and use 'dangerouslySetInnerHTML' " + 'on the enclosing div instead.') : void 0;
+ warning$1(typeof trustedTypes === 'undefined', "Using 'dangerouslySetInnerHTML' in an svg element with " + 'Trusted Types enabled in an Internet Explorer will cause ' + 'the trusted value to be converted to string. Assigning string ' + "to 'innerHTML' will throw an error if Trusted Types are enforced. " + "You can try to wrap your svg element inside a div and use 'dangerouslySetInnerHTML' " + 'on the enclosing div instead.');
```

walaura:
> I guess someone from the Core team also wants to take a look but other than that don't forget to run `eslint . --fix` before merging.

Will do!

as an addendum to the extra blocks - the automated tool is a bit naive and adds an if that in many cases can be redundant (for example, in the PR examples, a human could rewrite that as `if(suspenseEnabled) && __DEV__)`. This means that with time some of these extra blocks will go away!

packages/create-subscription/src/createSubscription.js
```diff
@@ -36,14 +36,18 @@ export function createSubscription<Property, Value>(
...
-  warningWithoutStack(
-    typeof subscribe === 'function',
-    'Subscription must specify a subscribe function',
-  );
+  if (__DEV__) {
+    warningWithoutStack(
+      typeof getCurrentValue === 'function',
+      'Subscription must specify a getCurrentValue function',
+    );
+  }
```
gaearon: Mind putting these into the same DEV block?

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -144,13 +144,15 @@ const ReactDOM: Object = {
   unstable_createPortal(...args) {
     if (!didWarnAboutUnstableCreatePortal) {
```
gaearon: Mind putting code like this inside the DEV block? Since it's only used by it.

packages/react-dom/src/client/ReactDOMSelect.js
```diff
@@ -49,21 +49,25 @@ function checkSelectPropTypes(props) {
...
     const isArray = Array.isArray(props[propName]);
     if (props.multiple && !isArray) {
-      warning(
-        false,
-        'The `%s` prop supplied to <select> must be an array if ' +
-          '`multiple` is true.%s',
-        propName,
-        getDeclarationErrorAddendum(),
-      );
+      if (__DEV__) {
```
gaearon: Mind putting DEV block around the body of this whole function?

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -393,15 +393,17 @@ export function useLayoutEffect(
...
-  warning(
-    false,
-    'useLayoutEffect does nothing on the server, because its effect cannot ' +
-      "be encoded into the server renderer's output format. This will lead " +
-      'to a mismatch between the initial, non-hydrated UI and the intended ' +
-      'UI. To avoid this, useLayoutEffect should only be used in ' +
-      'components that render exclusively on the client. ' +
-      'See https://fb.me/react-uselayouteffect-ssr for common fixes.',
-  );
+  if (__DEV__) {
```
gaearon: Mind combining it with previous DEV block?

packages/react-dom/src/shared/CSSPropertyOperations.js
```diff
@@ -148,17 +148,19 @@ export function validateShorthandPropertyCollisionInDev(
...
-        '%s a style property during rerender (%s) when a ' +
-          'conflicting property is set (%s) can lead to styling bugs. To ' +
-          "avoid this, don't mix shorthand and non-shorthand properties " +
-          'for the same value; instead, replace the shorthand with ' +
-          'separate values.',
-        isValueEmpty(styleUpdates[originalKey]) ? 'Removing' : 'Updating',
-        originalKey,
-        correctOriginalKey,
-      );
+      if (__DEV__) {
```
gaearon: Let's put DEV block around the whole function body

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -61,29 +61,33 @@ function areHookInputsEqual(
...
-      false,
-      'The final argument passed to %s changed size between renders. The ' +
-        'order and size of this array must remain constant.\n\n' +
-        'Previous: %s\n' +
-        'Incoming: %s',
-      currentHookNameInDev,
-      `[${nextDeps.join(', ')}]`,
-      `[${prevDeps.join(', ')}]`,
-    );
+    if (__DEV__) {
```
gaearon: Let's put this around the outer condition.

walaura:
Superceeded by #17540

PR #17538, opened 2019-12-06
HostText needs to copy over from current if it is unchanged

sebmarkbage:
stateNode is not one of the fields that gets copied over by createWorkInProgress.

This is bad for persistent mode since it's not the same instance.

In HostComponent we happened to always end up transferring the current stateNode when we bail.
However, we didn't do this for HostText.

Found by @JoshuaGross in Fabric.

TODO: Unit tests.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -536,6 +536,8 @@ if (supportsMutation) {
       // We'll have to mark it as having an effect, even though we won't use the effect for anything.
       // This lets the parents know that at least one of their children has changed.
       markUpdate(workInProgress);
+    } else {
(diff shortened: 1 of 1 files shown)


JoshuaGross:
Thanks for fixing this so quickly!

JoshuaGross:
@sebmarkbage Is there a plan to add tests to this and land / is anyone driving it? This will become/is a blocker for Fabric TextInput.

gaearon:
@JoshuaGross I was planning to do a sync after the holidays. This is currently blocked by lack of tests but we can land by then if @sebmarkbage doesn't write any.

JoshuaGross:
@gaearon Looks like we should just merge this in, since there haven't been any updates / tests

gaearon:
I got it in. I will run a sync once I’m back from PTO — on Monday.

PR #17540, opened 2019-12-06
Replace `wrap-warning-with-env-check` with an eslint plugin

walaura:
Part 1.5 of #16753

This PR removes the `__DEV__` wrapping part of `wrap-warning-with-env-check` and replaces it with a functionally identical eslint plugin (`no-production-logging`).The babel rule gets renamed to `lift-warning-conditional-argument` because that's what it does! In its current stage this will fail builds since we still got all the unwrapped calls in the actual codebase.

The plugin supports `&&`statements like `if( __DEV__ && React.suspenseIsEnabled)` but fails any top level `||`s. Have a look at the tests and code because I might have some unchecked assumptions about static code analysis here (and about how we strip out the `__DEV__` checks at compile time lol) but i think this covers the cases we wanna cover.

## Caveats

The generated development builds will include an extra block around the checks, from this original code
```jsx
if(suspenseEnabled){
  warning(2===2, 'example warning');
  ...
}
```

eslint (in this branch) will make the editor rewrite it to
```jsx
if(suspenseEnabled){
  if(__DEV__){
    warning(2===2, 'example warning');
  }
  ...
}
```

Which babel will transform (using ternaries) to
```jsx
if(suspenseEnabled){
  if(__DEV__){
    2===2 ? warning(false, 'example warning') : void 0;
  }
  ...
}
```

And then rollup will rewrite `__DEV__` to true (for a dev bundle)
```jsx
if(suspenseEnabled){
  if(true){
    2===2 ? warning(false, 'example warning') : void 0;
  }
  ...
}
```

And finally, commonjs remove the useless if (this seems like a happy side effect rather than something we are doing on purpose for the dev builds - as for prod we run the closure compiler). However it will keep the block scope:

```jsx
if(suspenseEnabled){
  {
    2===2 ? warning(false, 'example warning') : void 0;
  }
  ...
}
```

This makes sense to me, _this_ block scope is pretty absurd but we can't remove blocks automatically (at least not naively) because in some cases it _can_ affect the scope of variables.

We didn't have this before because babel would transform everything using ternaries, which don't start a block scope. Since we are moving the __DEV__ check to the userland now I don't think we want people writing this manually:

```jsx
if(suspenseEnabled){
    __DEV__ ? 2===2 ? warning(false, 'example warning') : void 0 : void 0;
  ...
}
```

## Next steps

How do we merge this? The way I see it we can either 

- Set the rule to warn, progressively fix the warnings. Don't like this because we don't have any rules set to warn? I might be wrong here? help??
- Prepare an eslint fix on this PR and apply it before merging. The builds will become green.

### DIFF
--- a/scripts/eslint-rules/__tests__/no-production-logging-test.internal.js
+++ b/scripts/eslint-rules/__tests__/no-production-logging-test.internal.js
@@ -1,0 +1,75 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails react-core
+ */
+
+'use strict';
+
+const rule = require('../no-production-logging');
+const RuleTester = require('eslint').RuleTester;
+const ruleTester = new RuleTester();
+
+ruleTester.run('no-production-logging', rule, {
+  valid: [
+    /* wrapped calls */
+    {
+      code: 'if (__DEV__) { warningWithoutStack(test) }',
+    },
+    {
+      code: 'if (__DEV__) { warning(test) }',
+    },
+    /* calls wrapped on an outer bound */
+    {
+      code: 'if (__DEV__) { if (potato) { while (true) { warning(test) } } }',
+    },
+    /* calls wrapped on an if with && */
+    {
+      code: 'if (banana && __DEV__ && potato && kitten) { warning(test) }',
+    },
+    /* don't do anything to arbitrary fn's or invariants */
+    {
+      code: 'normalFunctionCall(test)',
+    },
+    {
+      code: 'invariant(test)',
+    },
+  ],
+  invalid: [
+    {
+      code: 'warningWithoutStack(test)',
+      errors: [
+        {
+          message: `We don't emit warnings in production builds. Wrap warningWithoutStack() in an "if (__DEV__) {}" check`,
+        },
+      ],
+    },
+    {
+      code: 'if (potato) { warningWithoutStack(test) }',
+      errors: [
+        {
+          message: `We don't emit warnings in production builds. Wrap warningWithoutStack() in an "if (__DEV__) {}" check`,
+        },
+      ],
+    },
+    {
+      code: 'warning(test)',
+      errors: [
+        {
+          message: `We don't emit warnings in production builds. Wrap warning() in an "if (__DEV__) {}" check`,
+        },
+      ],
+    },
+    {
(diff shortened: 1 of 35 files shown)


scripts/eslint-rules/__tests__/no-production-logging-test.internal.js
```diff
@@ -0,0 +1,75 @@
...
+    {
+      code: 'if (__DEV__) { warning(test) }',
+    },
+    /* calls wrapped on an outer bound */
+    {
+      code: 'if (__DEV__) { if (potato) { while (true) { warning(test) } } }',
+    },
+    /* calls wrapped on an if with && */
+    {
+      code: 'if (banana && __DEV__ && potato && kitten) { warning(test) }',
```
gaearon: This feels too fragile and easy to get wrong, let's just disallow this.

scripts/eslint-rules/__tests__/no-production-logging-test.internal.js
```diff
@@ -0,0 +1,75 @@
...
+ * @emails react-core
+ */
+
+'use strict';
+
+const rule = require('../no-production-logging');
+const RuleTester = require('eslint').RuleTester;
+const ruleTester = new RuleTester();
+
+ruleTester.run('no-production-logging', rule, {
```
gaearon: Let's also check lowPriorityWarning (and its WithoutStack variant) in tests

scripts/eslint-rules/__tests__/no-production-logging-test.internal.js
```diff
@@ -0,0 +1,75 @@
...
+    /* wrapped calls */
+    {
+      code: 'if (__DEV__) { warningWithoutStack(test) }',
+    },
+    {
+      code: 'if (__DEV__) { warning(test) }',
+    },
+    /* calls wrapped on an outer bound */
+    {
+      code: 'if (__DEV__) { if (potato) { while (true) { warning(test) } } }',
```
gaearon: Nit: let's make this multiline so it's easier to read tests

scripts/eslint-rules/no-production-logging.js
```diff
@@ -0,0 +1,82 @@
...
+        if (node.operator === '&&') {
+          return [...traverseIf(node.left), ...traverseIf(node.right)];
+        }
+        return [];
+      default:
+        return [];
+    }
+  }
+
+  function hasIfInParents(node) {
```
gaearon: Nit: `hasIfDEVInParents`

PR #17541, opened 2019-12-06
Distinguish between an empty slot and a mismatch

jddxf:
Currently an empty slot is regarded as a mismatch between the slots. It triggers the slow path when we run into empty slots. Since empty slots are rather common, I believe it's worthwhile to distinguish between the two.

### DIFF
--- a/packages/react-reconciler/src/ReactChildFiber.js
+++ b/packages/react-reconciler/src/ReactChildFiber.js
@@ -546,8 +546,8 @@ function ChildReconciler(shouldTrackSideEffects) {
     oldFiber: Fiber | null,
     newChild: any,
     expirationTime: ExpirationTime,
-  ): Fiber | null | false {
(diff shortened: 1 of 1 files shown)


jddxf:
@gaearon Never knew that before. I'm considering to use another thing to indicate a mismatch, e.g. `undefined` or a unique empty object. Or I could add another function which would tell us whether the slots match. And we call `updateSlot` only when this function returns `true`. Which do you think fits better?

jddxf:
Finally I used `undefined` to indicate a mismatch. Not sure whether it hurts somehow since `undefined` is rarely used in the codebase.

gaearon:
I think we'd need some extensive benchmarking to confirm this is, in fact, better in practice. Can you come up with a plan? We have some old bench tooling in https://github.com/facebook/react/tree/master/scripts/bench, maybe that could be brought up-to-date and built upon.

PR #17542, opened 2019-12-06
Move scheduleHydration to root prototype

sebmarkbage:
We don't technically need the root since we'll find it through the DOM but we do need the root to have been created before we call this.

This API makes that a bit clearer. This was even asked about before so seems reasonable.

The only unfortunate thing is that it's not possible to do a generic scroll or event driven hydration without having access to a root.


gaearon:
So?

PR #17550, opened 2019-12-09
Fix comments typos

nicksp:
While investigating React Reconciler (for personal education) and its main phases: `ReactFiberBeginWork`, `ReactFiberCommitWork` and `ReactFiberCompleteWork` I spotted some typos in dev comments and thought I would commit back some of the sensible changes for folks who read comments 😛


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -728,7 +728,7 @@ function updateClassComponent(
   if (instance === null) {
     if (current !== null) {
       // An class component without an instance only mounts if it suspended
-      // inside a non- concurrent tree, in an inconsistent state. We want to
+      // inside a non-concurrent tree, in an inconsistent state. We want to
       // tree it like a new mount, even though an empty version of it already
```
gc: Should this be `treat it`?

nicksp: Good catch. Just fixed that.

PR #17554, opened 2019-12-09
[react-interactions] fix Press/Tap behavior for virtual middle clicks

necolas:
Tools like BetterTouchTool for macOS trigger middle-clicks with a 'buttons'
value that doesn't correspond to the middle-mouse button. To account for this
we also inspect the value of 'button'.

Close #17367

### DIFF
--- a/packages/react-interactions/events/src/dom/testing-library/domEnvironment.js
+++ b/packages/react-interactions/events/src/dom/testing-library/domEnvironment.js
@@ -65,6 +65,34 @@ export const platform = {
 };
 
 /**
+ * Button property
+ * This property only guarantees to indicate which buttons are pressed during events caused by pressing or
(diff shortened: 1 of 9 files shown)


packages/react-interactions/events/src/dom/testing-library/domEnvironment.js
```diff
@@ -64,6 +64,34 @@ export const platform = {
   },
 };
 
+/**
+ * Button property
+ * This property only guarantees to indicate which buttons are pressed during events caused by pressing or
+ * releasing one or multiple buttons. As such, it is not reliable for events such as 'mouseenter', 'mouseleave',
+ * 'mouseover', 'mouseout' or 'mousemove'. Furthermore, the semantics different for PointerEvent, where the value
```
trueadm: typo: the semantics different

PR #17556, opened 2019-12-09
React.unstable_avoidThisRender()

threepointone:
Introduces React.unstable_avoidThisRender(). 

This is one step towards implementing first class suspending behaviour for static resources like images/stylesheets. 

One of the problems with doing this naively in "userland", is because React is so aggressive on rerendering when the thenables resolve, we end up doing a bunch of unneeded rerenders. This sequence of PRs is to optimize the paths React executes in such cases.

How it's implemented
---

(There might be errors/a lack of understanding in this description)

`unstable_avoidThisRender()` is called during the render phase, which is similar to the 'throw' phase (without the actual throwing). When this happens:
- we look for the nearest Suspense boundary (that satisfies hasInvisibleParentBoundary === true), 
- and mark it's `effectTag` as `SoftCapture` (which is a variant of `DidCapture` for avoided renders). 
- We also attach a ping listener for cleaning up if the thenable resolves before we commit that boundary.

(These steps are similar to what we'd call if you threw an actual promise; ie - `throwException()`)

Soon after this, `completeWork` is called for the marked Suspense boundary. We use this instant to mark the boundary as expired, and return itself to render itself again.

Soon after, beginWork is called for that boundary. `SoftCapture` tells us that boundary has suspended. We clear the effectTag, initialize suspenseState/memoizedState on that boundary, and set `expirationTime` and `memoizedState.didAvoidRenderTime` as `renderExpirationTime - 1`; ie - we mark that boundary as 'expiring very soon'. 

Soon `unstable_avoidThisRender()` is called again, setting the effectTag back to `SoftCapture`. (This time in completeWork, we see that the effectTag and memoizedState are set, so can bypass returning itself.) 

After committing, the content rerender happens. (???) 

This time, in `completeWork()`, we call `renderDidSuspendDelayIfPossible()` to set the exit status on the tree. Currently, the actual content gets committed.

That's ...it? In the next PR, we'll add a new exitStatus, and actually show the fallback, passing the commented out tests in `ReactSuspenseWithNoopRenderer-test.internal.js`

TODO: address this review comment https://github.com/facebook/react/pull/17556/files#r358883141

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -7,6 +7,8 @@
  * @flow
  */
 
+import type {Thenable} from 'react-reconciler/src/ReactFiberWorkLoop';
+
 import type {
   ReactContext,
   ReactProviderType,
@@ -276,6 +278,10 @@ const Dispatcher: DispatcherType = {
   useResponder,
   useTransition,
   useDeferredValue,
+
+  avoidThisRender(thenable: Thenable) {
+    // no-op
+  },
 };
 
 // Inspect
--- a/packages/react-dom/src/server/ReactPartialRendererHooks.js
+++ b/packages/react-dom/src/server/ReactPartialRendererHooks.js
@@ -11,6 +11,9 @@ import type {
   Dispatcher as DispatcherType,
   TimeoutConfig,
 } from 'react-reconciler/src/ReactFiberHooks';
+
+import type {Thenable} from 'react-reconciler/src/ReactFiberWorkLoop';
+
 import type {ThreadID} from './ReactThreadIDAllocator';
 import type {
   ReactContext,
@@ -499,4 +502,7 @@ export const Dispatcher: DispatcherType = {
   useResponder,
   useDeferredValue,
   useTransition,
+  avoidThisRender(thenable: Thenable) {
+    // todo: this should trigger a fallback on SSR
+  },
 };
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -54,6 +54,7 @@ import {
   Update,
   Ref,
   Deletion,
+  SoftCapture,
 } from 'shared/ReactSideEffectTags';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import {
@@ -1563,6 +1564,7 @@ function validateFunctionComponentInDev(workInProgress: Fiber, Component: any) {
 const SUSPENDED_MARKER: SuspenseState = {
   dehydrated: null,
   retryTime: NoWork,
+  didAvoidRenderTime: NoWork,
 };
 
 function shouldRemainOnFallback(
@@ -1599,7 +1601,8 @@ function updateSuspenseComponent(
   let suspenseContext: SuspenseContext = suspenseStackCursor.current;
 
   let nextDidTimeout = false;
-  const didSuspend = (workInProgress.effectTag & DidCapture) !== NoEffect;
+  const didSuspend =
+    (workInProgress.effectTag & (SoftCapture | DidCapture)) !== NoEffect;
 
   if (
     didSuspend ||
@@ -1609,6 +1612,9 @@ function updateSuspenseComponent(
     // rendering the fallback children.
     nextDidTimeout = true;
     workInProgress.effectTag &= ~DidCapture;
+    // if SoftCapture is set, we don't clear it here
+    // since we read it later down below
+    // TODO: maybe it could be another module level global?
   } else {
     // Attempting the main content
     if (
@@ -1731,7 +1737,21 @@ function updateSuspenseComponent(
       primaryChildFragment.sibling = fallbackChildFragment;
       // Skip the primary children, and continue working on the
       // fallback children.
-      workInProgress.memoizedState = SUSPENDED_MARKER;
+
+      if (!(workInProgress.effectTag & SoftCapture)) {
+        workInProgress.memoizedState = SUSPENDED_MARKER;
+      } else {
+        workInProgress.effectTag &= ~SoftCapture;
+        const avoidedRenderExpiration = renderExpirationTime - 1; // verify/rename
+        // this means there's work to be done on this fiber "in the near future"
+        workInProgress.memoizedState = {
+          dehydrated: null,
+          retryTime: NoWork,
+          didAvoidRenderTime: avoidedRenderExpiration,
+        };
+
+        workInProgress.expirationTime = avoidedRenderExpiration;
+      }
       workInProgress.child = primaryChildFragment;
       return fallbackChildFragment;
     } else {
@@ -2996,6 +3016,7 @@ function beginWork(
             // whether to retry the primary children, or to skip over it and
             // go straight to the fallback. Check the priority of the primary
             // child fragment.
+
             const primaryChildFragment: Fiber = (workInProgress.child: any);
             const primaryChildExpirationTime =
               primaryChildFragment.childExpirationTime;
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1461,6 +1461,7 @@ function commitSuspenseComponent(finishedWork: Fiber) {
   if (enableSuspenseCallback && newState !== null) {
     const suspenseCallback = finishedWork.memoizedProps.suspenseCallback;
     if (typeof suspenseCallback === 'function') {
+      // TODO - what should suspenseCallback do if it's only images/avoidedThenables?
       const thenables: Set<Thenable> | null = (finishedWork.updateQueue: any);
       if (thenables !== null) {
         suspenseCallback(new Set(thenables));
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.js
@@ -60,6 +60,7 @@ import {
   NoEffect,
   DidCapture,
   Deletion,
+  SoftCapture,
 } from 'shared/ReactSideEffectTags';
 import invariant from 'shared/invariant';
 
@@ -870,6 +871,35 @@ function completeWork(
         workInProgress.expirationTime = renderExpirationTime;
         // Do not reset the effect list.
         return workInProgress;
+      } else if ((workInProgress.effectTag & SoftCapture) !== NoEffect) {
+        // something called avoidThisRender()
+        if (
+          current === null ||
+          (current.memoizedState &&
+            renderExpirationTime > current.memoizedState.didAvoidRenderTime)
+        ) {
+          // Something soft suspended, and this is the first "trigger rerender" pass.
+          // Re-render with the fallback children.
+          if (!workInProgress.memoizedState) {
+            // fallback isn't showing yet, trigger fallback pass
+            workInProgress.expirationTime = renderExpirationTime;
+            return workInProgress;
+          } else {
+            // TODO: we're never entering this branch because
+            // effectTag: SoftCapture is cleared after the previous pass
+            // does it matter?
+
+            // fallback's showing and we're in the
+            // render pass following the trigger pass
+            return null;
+          }
+        } else {
+          // second pass
+          // TODO: this doesn't actually delay right now?
+          renderDidSuspendDelayIfPossible();
+          // TODO: This also means SoftCapture tag isn't cleared from the fiber.
+          // does that matter?
+        }
       }
 
       const nextDidTimeout = nextState !== null;
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -13,6 +13,7 @@ import type {
   ReactEventResponderListener,
 } from 'shared/ReactTypes';
 import type {Fiber} from './ReactFiber';
+import type {Thenable} from './ReactFiberWorkLoop';
 import type {ExpirationTime} from './ReactFiberExpirationTime';
 import type {HookEffectTag} from './ReactHookEffectTags';
 import type {SuspenseConfig} from './ReactFiberSuspenseConfig';
@@ -43,6 +44,7 @@ import {
   warnIfNotScopedWithMatchingAct,
   markRenderEventTimeAndConfig,
   markUnprocessedUpdateTime,
+  avoidThisRender,
 } from './ReactFiberWorkLoop';
 
 import invariant from 'shared/invariant';
@@ -99,6 +101,8 @@ export type Dispatcher = {|
   useTransition(
     config: SuspenseConfig | void | null,
   ): [(() => void) => void, boolean],
+
+  avoidThisRender: (thenable: Thenable) => void,
 |};
 
 type Update<S, A> = {|
@@ -306,6 +310,14 @@ function throwInvalidHookError() {
       '2. You might be breaking the Rules of Hooks\n' +
       '3. You might have more than one copy of React in the same app\n' +
       'See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.',
+  );
+}
+
+function throwInvalidAvoidThisRenderError() {
+  invariant(
+    false,
+    'unstable_avoidThisRender can only be called inside of the ' +
+      'body of a function component, or during render() of a class component.',
   );
 }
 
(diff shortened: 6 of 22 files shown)


packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -92,6 +93,132 @@ describe('ReactSuspenseWithNoopRenderer', () => {
...
+            <AvoidedRendering />
+          </React.Suspense>
+        );
+      }
+
+      const root = ReactNoop.createRoot();
+
+      await ReactNoop.act(async () => {
+        root.render(<App />);
+      });
```
threepointone: This goes into an infinite loop right now, investigating why.

acdlite:
Overall approach looks good! Let me know if you need help with the infinite loop you're debugging

packages/shared/ReactFeatureFlags.js
```diff
@@ -94,3 +94,5 @@ export const enableTrustedTypesIntegration = false;
 
 // Flag to turn event.target and event.currentTarget in ReactNative from a reactTag to a component instance
 export const enableNativeTargetAsInstance = false;
+
+export const enableAvoidThisRenderAPI = false;
```
acdlite:
```suggestion
export const enableAvoidThisRenderAPI = __EXPERIMENTAL__;
```

packages/react-reconciler/src/ReactFiberThrow.js
```diff
@@ -175,6 +175,35 @@ function attachPingListener(
...
+  ) {
+    throwException(
+      root,
+      returnFiber,
+      sourceFiber,
+      avoidedThenable,
+      renderExpirationTime,
+    );
+  } else {
+    throw value;
```
acdlite: I think this part is wrong.

If we're inside already visible content, then we should continue rendering without triggering the parent fallback *and* delay commit.

So this should call `renderDidSuspendDelayIfPossible()`, which will change the root exit status to RootDidSuspendWithDelay. And it should attach a ping listener, so we can unblock the commit.

In the next PR, when we implement the ref counting strategy, this will instead change the root exit status to a new value (i.e. RootDidSuspendWithDelayRefCounting, but we'll think of a better name) and attach a special kind of ping listener that decrements the ref count.

acdlite: Need to make sure you have a test that covers this branch, btw. The test should mount a Suspense boundary, then in a `useTransition` update, mount a new image. The image will delay the commit instead of hiding the already-visible content.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -274,6 +276,11 @@ const Dispatcher: DispatcherType = {
   useResponder,
   useTransition,
   useDeferredValue,
+
+  unstable_avoidThisRender(thenable: Thenable) {
+    // this is just to satisfy flow, it should never be called.
+    // maybe I should put an invariant here
```
acdlite: Why should it never be called? It should be a no-op.

This special dispatcher is used to construct the DevTools hooks panel. DevTools executes the component function, which calls some hooks, then uses the log of hooks that were called to construct a tree of hooks. This will include any calls to `unstable_avoidThisRender`. We don't want to include `avoidThisRender` in the hook tree, which makes this kind of like `readContext`. But it also doesn't need to return anything, because it's a void function. So making it a no-op is correct.

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -501,4 +504,7 @@ export const Dispatcher: DispatcherType = {
   useResponder,
   useDeferredValue,
   useTransition,
+  unstable_avoidThisRender(thenable: Thenable) {
+    // no op on the server, for now
```
acdlite: Put a todo that says it should trigger the parent fallback

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1521,6 +1531,16 @@ function attachSuspenseRetryListeners(finishedWork: Fiber) {
...
+  } else {
+    let retryTime;
+    if (suspenseState !== null) {
+      retryTime = suspenseState.retryTime;
+    } else {
+      const currentTime = requestCurrentTimeForUpdate();
+      retryTime = computeExpirationForFiber(currentTime, finishedWork, null);
+    }
+
+    markUpdateTimeFromFiberToRoot(finishedWork, retryTime);
```
acdlite: I was wrong about this part. Re-rendering doesn't work because when we re-render it will trigger the same path (`hasInvisibleParentBoundary` is true), causing the infinite loop.

Lemme think about it and I'll get back to you.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1289,11 +1292,11 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
       }
       case SuspenseComponent: {
         commitSuspenseComponent(finishedWork);
-        attachSuspenseRetryListeners(finishedWork);
+        attachSuspenseRetryListeners(finishedWork, finishedWork.stateNode);
```
sebmarkbage: This isn't the right type. I think you meant to pass `memoizedState` but I think the factoring isn't right here because this function is meant to work with `SuspenseList` too and if it's doesn't, then I'm curious what will break in the `SuspenseList` case.

I think more likely, you'll just want to put your new logic in a separate function that you call from here and not mix it with `attachSuspenseRetryListeners`.

Currently you use the existence of the thennable Set to detect something but I don't think that means what you think it means. It's not a reliable signal for whether something is suspended. I'm not sure what the exact answer is but a good start would be to just move it out here to make it clear that it doesn't have anything to do with whether we attach listeners.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1521,6 +1531,16 @@ function attachSuspenseRetryListeners(finishedWork: Fiber) {
         thenable.then(retry, retry);
       }
     });
+  } else {
```
sebmarkbage: As I mentioned above, let's move this out to a separate function that isn't shared with SuspenseList since this is unrelated to attaching listeners.

packages/react-reconciler/src/ReactFiberThrow.js
```diff
@@ -175,6 +177,150 @@ function attachPingListener(
...
+        // in commit, add retry listener
+      }
+
+      return;
+    }
+    // This boundary already captured during this render. Continue to the next
+    // boundary.
+    workInProgress = workInProgress.return;
+  } while (workInProgress !== null);
+  // No boundary was found. Fallthrough to error mode.
```
acdlite: This is an error case. Very rare, not worth optimizing. I would just throw here.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1355,6 +1358,8 @@ export const ContextOnlyDispatcher: Dispatcher = {
   useResponder: throwInvalidHookError,
   useDeferredValue: throwInvalidHookError,
   useTransition: throwInvalidHookError,
+
+  unstable_avoidThisRender,
```
acdlite: This is wrong. Should be a method that throws an error. ("unstable_avoidThisRender() should not be called outside render...")

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -2036,5 +2053,7 @@ if (__DEV__) {
       updateHookTypesDev();
       return updateTransition(config);
     },
+
+    unstable_avoidThisRender,
```
acdlite: Similar note here. We shouldn't allow `avoidThisRender` inside setState reducers. Should warn.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1275,6 +1279,23 @@ function prepareFreshStack(root, expirationTime) {
   }
 }
 
+export function unstable_avoidThisRender(value: Thenable) {
+  invariant(
+    workInProgressRoot !== null &&
+      workInProgress !== null &&
```
acdlite: Pass `workInProgress` as an argument from the dispatcher method, where it's guaranteed to be defined, since the method is only accessible in the render phase.

packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
```diff
@@ -21,6 +21,7 @@ describe('ReactSuspenseWithNoopRenderer', () => {
     ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
     ReactFeatureFlags.replayFailedUnitOfWorkWithInvokeGuardedCallback = false;
     ReactFeatureFlags.flushSuspenseFallbacksInTests = false;
+    ReactFeatureFlags.enableAvoidThisRenderAPI = true;
```
acdlite: No new feature flags in tests. Instead, mark the tests that use it as experimental with `it.experimental`.

threepointone:
Later today, I'll flesh out the description, address feedback that I haven't yet, and maybe add another test.

packages/react-reconciler/src/ReactFiberThrow.js
threepointone: Added a warning instead.

packages/react/src/ReactAvoidThisRender.js
```diff
@@ -0,0 +1,31 @@
+/**
```
sebmarkbage: Just put this in ReactHooks where all the other hooks are. Also reuse the resolveDispatcher. I plan on removing the nice error message in prod anyway and let it fail with a "null" exception because it adds an unfortunate overhead in the Hook call and keeps it from inlining.

There are some that are their own file but I plan on just inlining everything. Our isomorphic file is embarrassingly large.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -99,6 +100,8 @@ export type Dispatcher = {
   useTransition(
     config: SuspenseConfig | void | null,
   ): [(() => void) => void, boolean],
+
+  unstable_avoidThisRender: typeof unstable_avoidThisRender,
```
sebmarkbage: Let's write out the expected signature explicitly. We also don't need to make this name "unstable" prefixed since it's a private API. It just adds bytes.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1639,7 +1646,24 @@ function updateSuspenseComponent(
       primaryChildFragment.sibling = fallbackChildFragment;
       // Skip the primary children, and continue working on the
       // fallback children.
-      workInProgress.memoizedState = SUSPENDED_MARKER;
+
+      if (workInProgress.effectTag & SoftCapture) {
+        // don't clear the effectTag yet, it'll be read in completeWork
```
sebmarkbage: This comment is wrong: `don't clear the effectTag yet` and then you clear it?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1639,7 +1646,24 @@ function updateSuspenseComponent(
...
+        // this means there's work to be done on this fiber "in the future"
+        workInProgress.memoizedState = {
+          dehydrated: null,
+          retryTime: NoWork,
+          didAvoidRenderTime: avoidedRenderExpiration,
+        };
+
+        workInProgress.expirationTime = avoidedRenderExpiration;
+      } else {
+        // workInProgress.effectTag & DidCapture is truthy
```
sebmarkbage: This comment is a bit too specific. It's just describing the technical details (also, we shouldn't actually be comparing truthiness but compare against zero so that's an oversight). You might want to describe it more as a higher level concept.

However, I think this points out a bug though. What if both are set? In completeWork, DidCapture wins, but in beginWork SoftCapture wins. That seems like it could cause havoc.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1639,7 +1646,24 @@ function updateSuspenseComponent(
       primaryChildFragment.sibling = fallbackChildFragment;
       // Skip the primary children, and continue working on the
       // fallback children.
-      workInProgress.memoizedState = SUSPENDED_MARKER;
+
+      if (workInProgress.effectTag & SoftCapture) {
+        // don't clear the effectTag yet, it'll be read in completeWork
+        workInProgress.effectTag &= ~SoftCapture;
+        const avoidedRenderExpiration = renderExpirationTime - 1; // verify/rename
```
sebmarkbage: What if the renderExpirationTime is Never? Then this will yield NoWork. It overflows. What prevents this from happening? If something is preventing it, then we should probably assert that. E.g. a DEV error.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2902,6 +2926,11 @@ function beginWork(
             // whether to retry the primary children, or to skip over it and
             // go straight to the fallback. Check the priority of the primary
             // child fragment.
+
+            // TODO: Around here, we need to detect that it's "2nd phase" of an
+            // avoided render (maybe just effectTag & SoftCapture ?),
+            // and just skip work for children altogether.
```
sebmarkbage: I think this comment is referring to something like resuming? But I'm not sure we have a reasonable safe way to do that atm or if it's even desirable. Another approach is to commit the children early and in that case there is no work to perform so it's automatic. Maybe just drop the comment until we have a proper plan.

packages/react-reconciler/src/ReactFiberCompleteWork.js
```diff
@@ -865,8 +866,36 @@ function completeWork(
...
+          // Something soft suspended, and this is the first "trigger rerender" pass.
+          // Re-render with the fallback children.
+          if (!workInProgress.memoizedState) {
+            // fallback isn't showing yet, trigger fallback pass
+            workInProgress.expirationTime = renderExpirationTime;
+            return workInProgress;
+          } else {
+            // TODO: we're never entering this branch because
+            // effectTag: SoftCapture is cleared after the previous pass
+            // does it matter?
```
sebmarkbage: You could make it an invariant that workInProgress.memoizedState is null. If it's not, then there's a bug somewhere since any throwing/avoidThisRender in the fallback should bubble past this. That way you also don't have to add a fake `return null`.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1275,6 +1279,23 @@ function prepareFreshStack(root, expirationTime) {
   }
 }
 
+export function unstable_avoidThisRender(value: Thenable) {
```
sebmarkbage: Drop this indirection. Just call `unstable_avoidThisRenderImpl` directly from ReactFiberHooks.

packages/react-reconciler/src/ReactFiberThrow.js
```diff
@@ -175,6 +176,64 @@ function attachPingListener(
   }
 }
 
+// unstable_avoidThisRender is much like throwException(), with some exceptions (ha!)
+export function unstable_avoidThisRenderImpl(
```
sebmarkbage: No need for the `unstable_` prefix here since it's internals.

packages/react/src/React.js
```diff
@@ -145,4 +148,8 @@ if (enableJSXTransformAPI) {
   }
 }
 
+if (enableAvoidThisRenderAPI) {
+  React.unstable_avoidThisRender = unstable_avoidThisRender;
```
sebmarkbage: This `React.unstable_avoidThisRender` is the only place you need the `unstable_` prefix.

packages/react-reconciler/src/ReactFiberWorkLoop.js
threepointone: I placed it here to have access to `workInProgressRoot`, `renderExpirationTime`, and `workInProgress`. Only `workInProgress` is in scope inside `ReactFiberHooks` (as `currentlyRenderingFiber`) Is there a simple way to get a hold of the former two?

packages/react-reconciler/src/ReactFiberBeginWork.js
threepointone: rewrote for DidCapture to win first

threepointone: scratch that, it fails tests. reverting and investigating.

gaearon:
Closing as stale.

PR #17560, opened 2019-12-09
Initialize update queue object on mount

acdlite:
Follow-up to b617db3d966f678eb0b4aac6d96f7967b37a9e91

Instead of lazily initializing update queue objects on the first update, class and host root queues are created on mount. This simplifies the logic for appending new updates and matches what we do for hooks.

### DIFF
--- a/packages/react-noop-renderer/src/createReactNoop.js
+++ b/packages/react-noop-renderer/src/createReactNoop.js
@@ -1142,20 +1142,33 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
 
       function logUpdateQueue(updateQueue: UpdateQueue<mixed>, depth) {
         log('  '.repeat(depth + 1) + 'QUEUED UPDATES');
-        const firstUpdate = updateQueue.firstUpdate;
-        if (!firstUpdate) {
+        const last = updateQueue.baseQueue;
+        if (last === null) {
           return;
         }
+        const first = last.next;
+        let update = first;
+        if (update !== null) {
+          do {
+            log(
+              '  '.repeat(depth + 1) + '~',
+              '[' + update.expirationTime + ']',
+            );
+          } while (update !== null && update !== first);
+        }
 
-        log(
-          '  '.repeat(depth + 1) + '~',
-          '[' + firstUpdate.expirationTime + ']',
-        );
-        while (firstUpdate.next) {
-          log(
-            '  '.repeat(depth + 1) + '~',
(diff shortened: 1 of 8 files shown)


sebmarkbage:
I don't think this should be on `.updateQueue` but on `.memoizedState`. The `.updateQueue` is treated as an effect queue for everything else at this point.

acdlite:
That makes sense but I didn't want to deal with the ramifications of other stuff that reads from it. Like DevTools. At least right now.

acdlite:
Basically I'm trying to keep the scope of this "bugfix" somewhat narrow. Can do a fuller refactor later.

packages/react-noop-renderer/src/createReactNoop.js
```diff
@@ -1142,20 +1142,33 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
...
         }
+        const first = last.next;
+        let update = first;
+        if (update !== null) {
+          do {
+            log(
+              '  '.repeat(depth + 1) + '~',
+              '[' + update.expirationTime + ']',
+            );
+          } while (update !== null && update !== first);
```
nicksp: Can `update` be changed while inside the loop? Do we need to additionally check for not null being inside `if (update !== null)`?

packages/react-noop-renderer/src/createReactNoop.js
```diff
@@ -1142,20 +1142,33 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
...
+        if (lastPending !== null) {
+          const firstPending = lastPending.next;
+          let pendingUpdate = firstPending;
+          if (pendingUpdate !== null) {
+            do {
+              log(
+                '  '.repeat(depth + 1) + '~',
+                '[' + pendingUpdate.expirationTime + ']',
+              );
+            } while (pendingUpdate !== null && pendingUpdate !== firstPending);
```
nicksp: Same question whether pendingUpdate needs to be checked for `! == null` when already inside `if` check

packages/react-noop-renderer/src/createReactNoop.js
acdlite: Good catch. It's missing a `update = update.next` line at the end of the loop. Honestly I don't think anyone actually uses this method anymore, so I'll probably just remove it.

nicksp: Yeah, I believe removal would be safe here and in the block below.

PR #17567, opened 2019-12-10
Show component location for selected element in bottom/right of props panel

### LINKED ISSUE #16497: DevTools: Show component file path

Feature.

The new DevTools doesn't show the component file path as did the v3 version.

It would be great to show the component file path, it's a really important feature when working in a large codebase.

React DevTools v4 is affected, v3 had it.

bvaughn:
Resolves #16497

Screen Shot 2019-12-10 at 8 46 02 AM

Rational for placement and formatting here:
https://github.com/facebook/react/issues/16497#issuecomment-564122209

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
@@ -7,6 +7,7 @@
  * @flow
  */
 
+import {copy} from 'clipboard-js';
 import React, {useCallback, useContext} from 'react';
 import {TreeDispatcherContext, TreeStateContext} from './TreeContext';
 import {BridgeContext, StoreContext} from '../context';
@@ -272,6 +273,7 @@ function InspectedElementView({
     hooks,
     owners,
     props,
+    source,
     state,
   } = inspectedElement;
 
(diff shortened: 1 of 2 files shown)


packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
```diff
@@ -403,6 +405,51 @@ function InspectedElementView({
...
+
+      {source !== null && (
+        <Source fileName={source.fileName} lineNumber={source.lineNumber} />
+      )}
+    </div>
+  );
+}
+
+// The function below (formatSourceForDisplay) is based on packages/shared/describeComponentFrame.js
+const BEFORE_SLASH_RE = /^(.*)[\\\/]/;
```
threepointone: Could you move this inside formatSourceForDisplay? Regexes are mutable, and I'm always wary of sharing them across function runs.

bvaughn: Sure. This follows the precedent in other files (like `describeComponentFrame`) but I don't mind making the change here.

gaearon:
Yay!

jschuler:
Is this feature still supported? I don't see the Source on selected elements

PR #17568, opened 2019-12-10
Remove the condition argument from warning()

walaura:
Step 2 of #16753 

- I've deleted the babel plugin (Woo!) and forked `warning-and-invariant-args` into `warning-args`and `invariant-args`.
- `toWarnDev` and `toLowPriorityWarnDev` didn't actually need any changes as they track the underlying console calls
- Some of the conditionals look wonky, not sure how to improve them without humans
- I don't have java on this laptop yet so can't build but it's installing rn


packages/create-subscription/src/createSubscription.js
```diff
@@ -37,14 +37,14 @@ export function createSubscription<Property, Value>(
...
   if (__DEV__) {
-    warningWithoutStack(
-      typeof getCurrentValue === 'function',
-      'Subscription must specify a getCurrentValue function',
-    );
-    warningWithoutStack(
-      typeof subscribe === 'function',
-      'Subscription must specify a subscribe function',
-    );
+    if (!(typeof getCurrentValue === 'function')) {
```
gaearon: Let's make a special case and convert these to `!==` negation directly.

packages/legacy-events/ResponderTouchHistoryStore.js
```diff
@@ -95,13 +95,14 @@ function resetTouchRecord(touchRecord: TouchRecord, touch: Touch): void {
...
   invariant(identifier != null, 'Touch object is missing identifier.');
   if (__DEV__) {
-    warningWithoutStack(
-      identifier <= MAX_TOUCH_BANK,
-      'Touch identifier %s is greater than maximum supported %s which causes ' +
-        'performance issues backfilling array locations for all of the indices.',
-      identifier,
-      MAX_TOUCH_BANK,
-    );
+    if (!(identifier <= MAX_TOUCH_BANK)) {
```
gaearon: Same for these. `<=` needs to flip to `>`, `<` needs to flip to `>=`, `>` needs to be `<=`, and `>=` needs to become `<`.

packages/react-cache/src/ReactCache.js
```diff
@@ -64,18 +64,23 @@ function readContext(Context, observedBits) {
 
 function identityHashFn(input) {
   if (__DEV__) {
-    warningWithoutStack(
-      typeof input === 'string' ||
+    if (
+      !(
```
gaearon: For more complex conditions like this it's fine to leave them as is.

walaura:
> * Let's detect `if (!(a === b))` early and instead output `if (a !== b)` and possibly the same for `<` and such.
> * Need to fix up `warningWithoutStack.www.js`. It currently re-exports `warning` from WWW. But it has the old API. So `warningWithoutStack.www.js` should now export a function that calls the WWW `warning` with `false` as first argument.

this should be good now but lemme recheck tomorrow

packages/shared/forks/lowPriorityWarningWithoutStack.www.js
```diff
@@ -6,4 +6,9 @@
  */
 
 // This "lowPriorityWarning" is an external module
-export default require('lowPriorityWarning');
+const lowPriorityWarning = require('lowPriorityWarning');
+
+export default function(...args) {
+  // eslint-disable-next-line react-internal/warning-args, react-internal/no-production-logging
+  return lowPriorityWarning(false, args);
```
gaearon: This one is interesting. Did we change this API in this PR? If not, what's the www API? Has this been wrong the whole time?

gaearon:
I still see the old `!(a === b)`, did you re-run the codemod after those changes?

packages/shared/forks/lowPriorityWarningWithoutStack.www.js
gaearon: actually ignore me, we did change its API

MuYunyun:
👍, the previous usage of warning is `warning(false, 'xxx')`, it increases the difficult of understanding the code.

In my option, it's naturally more to warn in the condition of true.

```js
if (true) {
  warn('xxx')
}
```

and why not write `waring(true, `xxx`)`? :smile:

PR #17579, opened 2019-12-12
Improved inspected element props with inline previews

### LINKED ISSUE #17073: [DevTools] polish hooks: complex values preview
We create many `useDebugValue` helpers, loggers and so one now, so we can feel less pain.

What matter (**Updated**)
- previewing complex values brifely like chrome devtools does it for arrays, sets, maps, objects,,,

I [wrote related twit](https://twitter.com/mxtnr/status/1178271362890240001) recently with one of my DebugValue helpers. screenshot out here:
Screenshot 2019-09-29 at 14 01 53
_my named hooks currently require a `label` option unless react babel cannot transform name of variable automatically_

bvaughn:
This mimics the inline preview shown by the browser console and dramatically improves the UX when inspecting deep values. I also updated tests to add more coverage for this new functionality.

There's a bit of hackiness in here (e.g. detecting `DataView` and `ArrayBuffer` types ) but I don't know of a better solution. Open to feedback!

Resolves #17073

# Demo
## Before
Demo: Before preview value changes

## After
Demo: After preview value changes

### DIFF
--- a/packages/react-devtools-shared/src/utils.js
+++ b/packages/react-devtools-shared/src/utils.js
@@ -10,6 +10,22 @@
 import Symbol from 'es6-symbol';
 import LRU from 'lru-cache';
 import {
+  isElement,
+  typeOf,
+  AsyncMode,
+  ConcurrentMode,
+  ContextConsumer,
+  ContextProvider,
+  ForwardRef,
+  Fragment,
+  Lazy,
+  Memo,
+  Portal,
+  Profiler,
+  StrictMode,
+  Suspense,
+} from 'react-is';
+import {
   TREE_OPERATION_ADD,
   TREE_OPERATION_REMOVE,
   TREE_OPERATION_REORDER_CHILDREN,
@@ -28,6 +44,8 @@ import {
   ElementTypeMemo,
 } from 'react-devtools-shared/src/types';
 import {localStorageGetItem, localStorageSetItem} from './storage';
+import {alphaSortEntries} from './devtools/views/utils';
+import {meta} from './hydration';
 
 import type {ComponentFilter, ElementType} from './types';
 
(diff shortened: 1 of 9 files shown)


packages/react-devtools-shared/src/utils.js
```diff
@@ -305,3 +323,277 @@ export function setInObject(
...
+      return `ArrayBuffer(${data.byteLength})`;
+    case 'data_view':
+      return `DataView(${data.buffer.byteLength})`;
+    case 'array':
+      if (showFormattedValue) {
+        const formatted = [];
+        for (let i = 0; i < data.length; i++) {
+          formatted[i] = formatDataForPreview(data[i], false);
+        }
+        return `[${truncateForDisplay(formatted.join(', '))}]`;
```
gaearon: I’m curious about what happens with very large arrays. Is there any work we’re doing that would be in vain? Like thousands of items. Should we break earlier?

packages/react-devtools-shared/src/utils.js
```diff
@@ -305,3 +323,277 @@ export function setInObject(
...
+          ? data[meta.size]
+          : data.length;
+        return `Array(${length})`;
+      }
+    case 'typed_array':
+      const shortName = `${data.constructor.name}(${data.length})`;
+      if (showFormattedValue) {
+        return `${shortName} [${truncateForDisplay(
+          data
+            .toString()
```
gaearon: Same question. If the data is very large, are we doing unnecessary work by toString’ing it all, only to take a slice later? Should we first take a slice with a reasonable limit?

packages/react-devtools-shared/src/utils.js
```diff
@@ -305,3 +323,277 @@ export function setInObject(
...
+          // To mimic their behavior, detect if we've been given an entries tuple.
+          //   Map(2) {"abc" => 123, "def" => 123}
+          //   Set(2) {"abc", 123}
+          if (Array.isArray(entryOrEntries)) {
+            const key = formatDataForPreview(entryOrEntries[0], true);
+            const value = formatDataForPreview(entryOrEntries[1], false);
+            valueStrings.push(`${key} => ${value}`);
+          } else {
+            valueStrings.push(formatDataForPreview(entryOrEntries, false));
+          }
```
gaearon: Same question. Should we break down earlier if we know we’re gonna go over the truncation limit anyway, and work for next items is wasted?

packages/react-devtools-shared/src/utils.js
```diff
@@ -305,3 +323,277 @@ export function setInObject(
...
+      return data.toString();
+    case 'object':
+      if (showFormattedValue) {
+        const formatted = [];
+        Object.keys(data)
+          .sort(alphaSortEntries)
+          .forEach(key => {
+            formatted.push(`${key}: ${formatDataForPreview(data[key], false)}`);
+          });
+        return `{${truncateForDisplay(formatted.join(', '))}}`;
```
gaearon: Same

packages/react-devtools-shared/src/utils.js
bvaughn: Ah, this is good insight. Definitely can add a short circuit here.

hristo-kanchev:
Nice one, @bvaughn 🤘 !!

Issue #17580, opened 2019-12-12
lastNativeEvent retains unmounted elements & fibers


bgirard:
I'm seeing memory kept around for longer then it needs to by lastNativeEvent:
https://github.com/facebook/react/blob/26ba38ae4b65375b97b561ec2d6a58ef597647ed/packages/react-dom/src/events/EnterLeaveEventPlugin.js#L49

In my case it's holding on to detached DOM nodes/Fiber.

I think we should be able to use WeakSet since we're doing a `===` check. Otherwise we should consider timing it out in some way.

bgirard:
Related #17006

bgirard:
Fixed by #17585

PR #17582, opened 2019-12-12
Hold on to the lastNativeEvent weakly so as not to retain fibers belonging to long-ago unmounted DOM nodes

steveluscher:
The event that we store in `lastNativeEvent` has a reference to its related DOM nodes which has a reference to some `FiberNode` instances. When a component unmounts we don't want to retain any of that.

This PR stores the `lastNativeEvents` that we've seen in a `WeakSet` so as not to cause them to be retained. Where `WeakSet` is not supported, we use a shim that does what the old thing used to do (and leaks like it too).

Here's a snapshot of one of these leaks in the Chrome Memory profiler. `tk` is `FiberNode` and `tl` is `lastNativeEvent`.

Screen Shot 2019-12-11 at 5 54 58 PM

This aims to address #17580. Credit for the idea to use `WeakSet` goes to @bgirard!

### DIFF
--- a/packages/react-dom/src/events/EnterLeaveEventPlugin.js
+++ b/packages/react-dom/src/events/EnterLeaveEventPlugin.js
@@ -42,11 +42,11 @@ const eventTypes = {
   },
 };
 
-// We track the lastNativeEvent to ensure that when we encounter
-// cases where we process the same nativeEvent multiple times,
-// which can happen when have multiple ancestors, that we don't
-// duplicate enter
-let lastNativeEvent;
+// We track the native events we've seen to ensure that when we encounter cases
+// where we process the same nativeEvent multiple times, which can happen when
+// have multiple ancestors, that we don't duplicate enter.
+const seenNativeEvents =
+  typeof WeakSet === 'function' ? new WeakSet() : new Set();
 
 const EnterLeaveEventPlugin = {
   eventTypes: eventTypes,
@@ -169,11 +169,18 @@ const EnterLeaveEventPlugin = {
 
     accumulateEnterLeaveDispatches(leave, enter, from, to);
 
-    if (nativeEvent === lastNativeEvent) {
-      lastNativeEvent = null;
+    if (seenNativeEvents.has(nativeEvent)) {
+      if (seenNativeEvents instanceof Set) {
+        seenNativeEvents.clear();
+      } else {
+        seenNativeEvents.delete(nativeEvent);
+      }
       return [leave];
     }
-    lastNativeEvent = nativeEvent;
+    if (seenNativeEvents instanceof Set) {
+      seenNativeEvents.clear();
+    }
+    seenNativeEvents.add(nativeEvent);
 
     return [leave, enter];
   },


packages/react-dom/src/events/EnterLeaveEventPlugin.js
```diff
@@ -169,11 +169,14 @@ const EnterLeaveEventPlugin = {
...
 
-    if (nativeEvent === lastNativeEvent) {
-      lastNativeEvent = null;
+    if (seenNativeEvents.has(nativeEvent)) {
+      seenNativeEvents.remove(nativeEvent);
       return [leave];
     }
-    lastNativeEvent = nativeEvent;
+    if (seenNativeEvents instanceof Set) {
+      seenNativeEvents.clear();
```
steveluscher: This replicates the ‘overwrite’ behavior of `lastNativeEvent = nativeEvent` for clients that don't support `WeakMap`.

packages/react-dom/src/events/EnterLeaveEventPlugin.js
```diff
@@ -169,11 +169,14 @@ const EnterLeaveEventPlugin = {
...
+    }
```
bgirard: Maybe:
seenNativeEvents = typeof WeakSet === 'function' ? new WeakSet() : new Set();

steveluscher: When `WeakSet` is supported then is there any harm to letting the set ‘grow’ infinitely? Something's presence in the set will be contingent on it being held on to somewhere else in your app, right?

When `WeakSet` is _not_ supported, I wonder if it's more efficient to call `clear()` rather than to construct a new `Set`.

bgirard: There's a small one. If you're leaking the thing you're holding a weak reference to, or retaining them on purpose, then this set is growing. If we only care about the last thing it's best to only track the last thing. It's unfortunate you can't simply clear a `WeakSet`.

packages/react-dom/src/events/EnterLeaveEventPlugin.js
```diff
@@ -169,11 +169,18 @@ const EnterLeaveEventPlugin = {
 
     accumulateEnterLeaveDispatches(leave, enter, from, to);
 
-    if (nativeEvent === lastNativeEvent) {
-      lastNativeEvent = null;
+    if (seenNativeEvents.has(nativeEvent)) {
+      if (seenNativeEvents instanceof Set) {
```
necolas: Can we get away with not performing the `instanceof` check each time? We should know at the point `seenNativeEvents` is defined whether or not we're working with a `Set` or `WeakSet`

sebmarkbage:
In generally we use expandos instead of weaksets as they've had better perf characteristics and doesn't require fallbacks implementations. E.g. `event._reactId = i++;` `lastSeenID = event._reactId;`.

However, in this case, we really shouldn't need to mess with the native events. It's just poorly structured code.

One solution would be to just pass a boolean or eventSystemFlags here that signifies that this is the first processed ancestor tree.

https://github.com/facebook/react/blob/acfe4b21b2621da3618f9096c8be06b32d6247ec/packages/react-dom/src/events/ReactDOMEventListener.js#L179

trueadm:
@sebmarkbage That's exactly what I did in my follow up PR! I noticed you added the flags to the EventPlugins, which made this fix slightly more obvious.

steveluscher:
Yay! Thanks @trueadm!

PR #17586, opened 2019-12-12
Add component stacks to (almost) all warnings

gaearon:
As the title says. This adds component stacks to all warnings except the ones where a package doesn't have peer or dep on React (such as `react-is`). Concretely, we switch over from `warningWithoutStack` to `warning`, and from `lowPriorityWarningWithoutStack` to `lowPriorityWarning`, wherever possible. Tests are adjusted accordingly.

There are three reasons for this.

1. Stacks are generally useful. More and consistent is better.
1. We want to move towards just writing `console.error` and `console.warn` in the source code, and having a Babel plugin call the wrapper with extra logic. So it's inconvenient to have this distinction.
1. Even for warnings that don't *currently* have stacks, they're actually already being added by React DevTools anyway. So the distinction hasn't mattered for a while.

What if the stack is misleading or not concrete enough? For example, maybe we have a better stack we want to pass. We do the same thing DevTools does in that case. If the last argument *already* looks like a component stack, we don't append an extra one.

Notably, this adds component stacks to SSR hydration mismatch warnings. They can still sometimes be misleading but it's better than nothing.

Some tests still have `{withoutStack: true}`. This is not because we *can't* emit them, but because in those particular tests we're just not inside of a tree. So we're keeping the assertion mechanism for that to avoid regressions.

In a follow-up, I will codemod to `console.error` / `console.warn` calls and a Babel plugin for the wrappers. This will also allow us to choose the right version of the wrapper depending on whether a package depends on React or not.


gaearon:
There's a very minor bundle size for non-React-using entry points increase due to more code paths pulling in the glue code that's only used in DEV. I'll fix that in a follow up that does this at the build time level.

PR #17587, opened 2019-12-12
Use console.error/warn() directly in the source

gaearon:
Picking up where https://github.com/facebook/react/pull/17586 left.

This replaces four custom modules in the source with direct calls:

- `shared/warning` -> `console.error`
- `shared/warningWithoutStack` -> `console.error`
- `shared/lowPriorityWarning` -> `console.warn`
- `shared/lowPriorityWarningWithoutStack` -> `console.warn`

**To review, you want to look at the second commit because the first one is just a find-replace.**

Here's a few points about the new system.

## Not Removed in Prod

We're not doing anything special to remove `console.error` or `console.warn` calls in prod anymore. If you keep them, they'll be there. We *do* have a linter plugin that tells you to wrap it in DEV. I audited existing cases where it fails, and added suppressions where it was intentional.

## Stack Is Injected Automatically

You don't need to choose whether to inject the stack. If you use `console.error`, it will be injected automatically, as long as the package from which you're calling it has access to the React object. Alternatively, it will keep your `console.error` as is in the source (no stack). DevTools would likely pick it up though anyway.

Note: in the future we might completely remove the stack injection and always rely on DevTools and similar mechanisms (e.g. RN integration) to inject it.

## Escape Hatches

If you want to use the "raw" `console.error` or `console.warn` for whatever reason, you can do it like this:

```js
console['error']('hahaha');
```

This will use the native `console.error` *in DEV too*. Practically, this means we won't try to append the stack, prepend the `Warning:` marker (which is currently relied on by RN), or go through the WWW whitelist. There's only a few places that we do this.

## Follow-ups

Out of scope of this PR, but after landing this I'm planning to rename `toWarnDev` to `toErrorDev` (because it's for `console.error`), and `toLowPriorityWarnDev` to `toWarnDev` (because it's for `console.warn`). Then I plan to audit the warnings and split them into these two levels according to whether they represent real issues or not.


gaearon:
https://github.com/facebook/react/pull/17599

aweary:
Screen Shot 2019-12-13 at 12 02 36 PM

PR #17591, opened 2019-12-13
Remove unstable_withSuspenseConfig

acdlite:
We don't need this now that the higher level APIs, `useTransition` and `useDeferredValue`, are built in.


sebmarkbage:
So I’ve actually started thinking of more use cases for this. It seems like the concept of transitions is growing more into “a predictable batch unit”. Which is also useful for animations. The name useTransition even alludes to animations which actually is a good thing because it’s a higher level concept onto which we can tie both loading and animations.

When things are expected to suspend, it’s best practice to have the flag (or use the difference between value and deferred value) to indicate that something is loading since it can take a while.

However when you want to start a transition for the purpose of triggering an animation, not so much. And it’s can happen in many places where getting to a hook is unnecessary or annoying.

So I’m thinking React.startTransition(...) could be a good name for this. With its primary purpose not being for suspense per se but for animations. It should go into the same special unique batch priority level as other transitions. Although it probably should still allow for suspending and a timeout.

acdlite:
Ok but until that exists can we delete this thing? Because in the meantime it’s an annoying implementation detail and lots of the Suspense tests are written in terms of it.

acdlite:
Also can you explain why the hook form is annoying if you need a hook for `setState` regardless?

sebmarkbage:
I think we need it pretty soon because of enter/exit animations.

Regardless, I don't want us to lose the implementation details as part of the refactoring since it's easy to build the implementation to be dependent on the connection to a Fiber. It also flushes out questions such as, can we keep this working cross-renderer in the sense that other renderers/roots aren't guaranteed to flush at the same time as the pending flag getting reset, but are guaranteed to have a separate priority level.

> Also can you explain why the hook form is annoying if you need a hook for setState regardless?

The point of `useTransition` is that you don't have direct access to the `useState` that is being affected. If you did, you could just `useDeferredValue`. So there's some kind of indirection between this caller and the target `setState`. In this caller there doesn't need to be any other Hooks.

acdlite:
Can I land only the test changes then?

acdlite:
Basically my main motivation for opening this PR now was to get the tests into a form that is closer to "idiomatic" React code

fixtures/ssr/src/components/Chrome.js
```diff
@@ -1,51 +1,42 @@
...
+      <body className={theme}>
+        <noscript
+          dangerouslySetInnerHTML={{
+            __html: `<b>Enable JavaScript to run this app.</b>`,
+          }}
+        />
+        <Theme.Provider value={theme}>
+          {props.children}
+          <div>
+            <ThemeToggleButton onChange={t => setTheme(t)} />
```
Jessidhia: typo for `setNextTheme`, or should `setNextTheme` be renamed 🤔

acdlite:
Closing. We decided to keep the API for now, even though it's internal-only, because it's likely we'll make some version of it public in the future (e.g. `React.startTransition`).

PR #17599, opened 2019-12-13
Use console directly instead of warning() modules

gaearon:
Picking up where #17568 and https://github.com/facebook/react/pull/17586 left.
Related to https://github.com/facebook/react/issues/16753.

This replaces four custom modules in the source with direct calls:

- `shared/warning` -> `console.error`
- `shared/warningWithoutStack` -> `console.error`
- `shared/lowPriorityWarning` -> `console.warn`
- `shared/lowPriorityWarningWithoutStack` -> `console.warn`

**To review, you want to look at the [second commit](https://github.com/facebook/react/pull/17599/commits/06478a670008c003e853888b608aa58eb9ae0a09?w=1) because the first one is just a find-replace.**

Here's a few points about the new system.

## Not Removed in Prod

We're not doing anything special to remove `console.error` or `console.warn` calls in prod anymore. If you keep them, they'll be there. We *do* have a linter plugin that tells you to wrap it in DEV. I audited existing cases where it fails, and added suppressions where it was intentional.

## Stack Is Injected Automatically

You don't need to choose whether to inject the stack. If you use `console.error`, it will be injected automatically, as long as the package from which you're calling it has access to the React object. Alternatively, it will keep your `console.error` as is in the source (no stack). DevTools would likely pick it up though anyway.

Note: in the future we might completely remove the stack injection and always rely on DevTools and similar mechanisms (e.g. RN integration) to inject it.

## Escape Hatches

If you want to use the "raw" `console.error` or `console.warn` for whatever reason, you can do it like this:

```js
console['error']('hahaha');
```

This will use the native `console.error` *in DEV too*. Practically, this means we won't try to append the stack, prepend the `Warning:` marker (which is currently relied on by RN), or go through the WWW whitelist. There's only a few places that we do this.

## Gotchas

We have two modules that don't have access to the isomorphic React package: `scheduler` and `react-is`. So we can't append the stacks. **This means `console.error` calls inside of these two packages won't get transformed even in DEV for real bundles — <s>they'll stay as `console.error`.** But in our tests, they would be transformed.</s> (I changed this to fail the build.)

We barely have any warnings there, and both are meant to be lightweight. So I think it's okay. The fact that they don't go through the WWW list can be confusing. But in longer term we want to get rid of the special `warning` entry point altogether. In that world, we would have the list be a list of warnings to *mute*, and pass everything else through.

To get there, we'll need to enforce that the React warning list matches WWW list 1:1. We'd need to get buy-in from other teams for that to work out. In the meantime, I don't expect we'll add a lot of new warnings to either `scheduler` or `react-is` anyway.

If this gotcha ends up being a pain, we can always have a second "compile target" that exists solely to make these warnings go through the WWW list anyway.

## Risks

Let's sync WWW carefully after this. We need to verify nothing broke and warnings still fire. I can do this.

## Follow-ups

Out of scope of this PR, but after landing this I'm planning to rename `toWarnDev` to `toErrorDev` (because it's for `console.error`), and `toLowPriorityWarnDev` to `toWarnDev` (because it's for `console.warn`). Then I plan to audit the warnings and split them into these two levels according to whether they represent real issues or not.

Additionally, I'll remove the cases where we detect presence of custom stack first before appending it. Instead, we'll set the current fiber.


acdlite:
If one of our goals is to be more rigorous about adding warnings, I wonder if we could take the same approach we use for the error transform, where we maintain the list of error messages in the source, and only transform the ones that have a matching entry. We use a lint rule to enforce that each message has a matching entry.

This makes adding/editing warnings slightly more annoying but it comes with some benefits:

- The escape hatch to opt out of transforming a warning is to suppress the lint error that occurs if you don't
- New warnings are more easily caught during code review. Makes it harder to sneak one by accidentally.
- We always have a list of warnings that is exactly in sync with the ones in the source code.

The downside is that adding a warning means also updating the warning list. But adding some inertia there could be a feature, too. We've gotten feedback that we have too many warnings of varying severity and no way to opt out of them, so it could behoove us to start being more intentional about adding them. (And if we ever do provide a way to opt-out of specific warnings, we would want to track them.)

packages/react-reconciler/src/ReactFiberErrorLogger.js
```diff
@@ -81,11 +78,12 @@ export function logCapturedError(capturedError: CapturedError): void {
...
     // has already printed it. Even if the application swallows the error, it is still
     // displayed by the browser thanks to the DEV-only fake event trick in ReactErrorUtils.
-    rawConsoleError(combinedMessage);
+    console['error'](combinedMessage); // Don't transform to our wrapper
   } else {
     // In production, we print the error directly.
     // This will include the message, the JS stack, and anything the browser wants to show.
     // We pass the error object instead of custom message so that the browser displays the error natively.
-    rawConsoleError(error);
+    // eslint-disable-next-line react-internal/no-production-logging
```
bvaughn: Is this disable-lint necessary?

packages/react/src/__tests__/ReactClassEquivalence-test.js
```diff
@@ -28,7 +28,8 @@ describe('ReactClassEquivalence', () => {
 function runJest(testFile) {
   const cwd = process.cwd();
   const extension = process.platform === 'win32' ? '.cmd' : '';
-  const result = spawnSync('yarn' + extension, ['test', testFile], {
+  const command = __DEV__ ? 'test' : 'test-prod';
```
bvaughn: Nice catch, but should we split it out into a different commit?

Your call!

scripts/eslint-rules/__tests__/no-production-logging-test.internal.js
```diff
@@ -18,28 +18,35 @@ ruleTester.run('no-production-logging', rule, {
...
-          warningWithoutStack(test, 'Oh no');
+          console.error('Hello %s', foo)
         }
       `,
     },
     {
       code: `
         if (__DEV__) {
-          lowPriorityWarning(test, 'Oh no');
+          console.error('Hello %s %s', foo, bar)
```
bvaughn: This change doesn't look equivalent. Shouldn't low-pri have mapped to `console.warn`?

scripts/eslint-rules/warning-args.js
```diff
@@ -44,17 +44,18 @@ module.exports = function(context) {
       // This could be a little smarter by checking context.getScope() to see
       // how warning/invariant was defined.
       const isWarning =
-        node.callee.type === 'Identifier' &&
-        (node.callee.name === 'warning' ||
-          node.callee.name === 'warningWithoutStack' ||
-          node.callee.name === 'lowPriorityWarning' ||
-          node.callee.name === 'lowPriorityWarningWithoutStack');
+        node.callee.type === 'MemberExpression' &&
+        node.callee.object.type === 'Identifier' &&
```
bvaughn: Should probably also check:
```js
node.callee.object.name === "console" &&
```
To avoid potential future collision with e.g.
```js
React.warn(...)
```

.eslintrc.js
```diff
@@ -135,6 +136,18 @@ module.exports = {
         'jest/valid-expect-in-promise': ERROR,
       },
     },
+    {
+      files: [
+        '**/__tests__/**/*.js',
+        'scripts/**/*.js',
+        'packages/*/npm/**/*.js',
+        'packages/react-devtools*/**/*.js'
```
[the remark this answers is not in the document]
gaearon: >Maybe we should intentionally exclude those packages, since the component stack logic would never be relevant?

I believe that’s what these lines are doing!

scripts/eslint-rules/__tests__/no-production-logging-test.internal.js
gaearon: That’s just git diff being funky. I added a few cases here.

gaearon:
@acdlite I’m down for this! Follow-up or here?

.eslintrc.js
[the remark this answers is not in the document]
bvaughn: Ah, duh. Yeah that would make more sense, now that I look again. okay 👍

acdlite:
@gaearon Follow up is fine, good to merge as-is

gaearon:
I pushed another commit that makes the "gotchas" clear by failing the build when `console.error/warn()` can't access the stack. In that case you're forced to use the escape hatch. In a follow-up, we can change that to use the warning list approach instead for the same purpose.

bvaughn:
This PR broke DevTools tests.

After this commit was merged, running `yarn test-build-devtools` fails with several errors, including:
```
    Configuration error:
    
    Could not locate module shared/consoleWithStackDev mapped as:
    /Users/bvaughn/Documents/git/react/build/node_modules/shared/consoleWithStackDev.
    
    Please check your configuration for these entries:
    {
      "moduleNameMapper": {
        "/^shared\/([^\/]+)$/": "/Users/bvaughn/Documents/git/react/build/node_modules/shared/$1"
      },
      "resolver": null
    }
```

PR #17608, opened 2019-12-15
DevTools context menu

bvaughn:
Inspired by built-in browser console options, this PR adds the ability to copy specific props/state/context/hooks values to the clipboard and to store them in a global variables (`$reactTemp`).

## Checklist
- [x] Add context menu component and hook.
- [x] Add backend commands for copying variables, storing global values, and inspecting attributes (works similarly to pre-existing pattern for inspecting React components).
- [x] Wire up context menu and backend commands to enable interacting with selected element props/state/hooks/context.
- [x] Add unit tests to cover new behavior.
- [x] Add `DevTools` props for enabling new behavior per environment. (Disabled for React Native for now.)

## New context menu options
Copy and store-as-global demo video

## Works for nested values
Demo of inspecting arbitrary props

### DIFF
--- a/packages/react-devtools-shared/src/backend/legacy/renderer.js
+++ b/packages/react-devtools-shared/src/backend/legacy/renderer.js
@@ -836,6 +836,16 @@ export function attach(
     }
   }
 
+  function prepareViewAttributeSource(
+    id: number,
+    path: Array<string | number>,
+  ): void {
+    const inspectedElement = inspectElementRaw(id);
+    if (inspectedElement !== null) {
+      window.$attribute = getInObject(inspectedElement, path);
+    }
+  }
+
   function prepareViewElementSource(id: number): void {
     const internalInstance = idToInternalInstanceMap.get(id);
     if (internalInstance == null) {
@@ -968,6 +978,7 @@ export function attach(
     inspectElement,
     logElementToConsole,
     overrideSuspense,
+    prepareViewAttributeSource,
     prepareViewElementSource,
     renderer,
     setInContext,
--- a/packages/react-devtools-shared/src/backend/legacy/renderer.js
+++ b/packages/react-devtools-shared/src/backend/legacy/renderer.js
@@ -649,14 +649,19 @@ export function attach(
     }
   }
 
-  function storeAsGlobal(id: number, path: Array<string | number>): void {
+  function storeAsGlobal(
+    id: number,
+    path: Array<string | number>,
+    count: number,
+  ): void {
     const inspectedElement = inspectElementRaw(id);
     if (inspectedElement !== null) {
       const value = getInObject(inspectedElement, path);
+      const key = `$reactTemp${count}`;
 
-      window.$reactTemp = value;
+      window[key] = value;
 
-      console.log('$reactTemp');
+      console.log(key);
       console.log(value);
     }
   }
--- a/packages/react-devtools-shared/src/backend/legacy/renderer.js
+++ b/packages/react-devtools-shared/src/backend/legacy/renderer.js
@@ -7,7 +7,6 @@
  * @flow
  */
 
-import {copy} from 'clipboard-js';
 import {
   ElementTypeClass,
   ElementTypeFunction,
@@ -16,7 +15,7 @@ import {
   ElementTypeOtherOrUnknown,
 } from 'react-devtools-shared/src/types';
 import {getUID, utfEncodeString, printOperationsArray} from '../../utils';
-import {cleanForBridge, copyWithSet, safeSerialize} from '../utils';
+import {cleanForBridge, copyToClipboard, copyWithSet} from '../utils';
 import {getDisplayName, getInObject} from 'react-devtools-shared/src/utils';
 import {
   __DEBUG__,
@@ -665,9 +664,7 @@ export function attach(
   function copyElementPath(id: number, path: Array<string | number>): void {
     const inspectedElement = inspectElementRaw(id);
     if (inspectedElement !== null) {
-      const value = getInObject(inspectedElement, path);
-
-      copy(safeSerialize(value));
+      copyToClipboard(getInObject(inspectedElement, path));
     }
   }
 
--- a/packages/react-devtools-shared/src/backend/legacy/renderer.js
+++ b/packages/react-devtools-shared/src/backend/legacy/renderer.js
@@ -7,6 +7,7 @@
  * @flow
  */
 
+import {copy} from 'clipboard-js';
 import {
   ElementTypeClass,
   ElementTypeFunction,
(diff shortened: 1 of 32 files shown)


packages/react-devtools-shared/src/backend/legacy/renderer.js
```diff
@@ -649,6 +649,25 @@ export function attach(
     }
   }
 
+  function storeAsGlobal(id: number, path: Array<string | number>): void {
+    const inspectedElement = inspectElementRaw(id);
+    if (inspectedElement !== null) {
+      const value = getInObject(inspectedElement, path);
+
+      window.$reactTemp = value;
```
gaearon: Maybe autoincrement like the browser does? So it's easy to save multiple.

packages/react-devtools-shared/src/devtools/ContextMenu/useContextMenu.js
```diff
@@ -0,0 +1,32 @@
...
+    () => {
+      if (ref.current !== null) {
+        const handleContextMenu = event => {
+          event.preventDefault();
+          event.stopPropagation();
+
+          const pageX =
+            event.clientX || (event.touches && event.touches[0].pageX);
+          const pageY =
+            event.clientY || (event.touches && event.touches[0].pageY);
```
gaearon: This confused me a bit. I was surprised the property is called `pageX/Y` everywhere else (e.g. in `showMenu` API and things it calls) because intuitively I expected `clientX/Y` to be used.

But it seems like we *do* use `clientX/Y` first.

What is the case where it doesn't exist? And does `pageX/Y` give us the same thing in that fallback? What if we already scrolled a bit, aren't they different in their meaning?

packages/react-devtools-shared/src/backend/legacy/renderer.js
bvaughn: I considered that but then I wasn't sure there was a compelling use case for saving multiple, and it gets harder to coordinate with multiple roots/renderers. Seems easy enough to re-assign (in console) if you want to save a second and compare.

bvaughn: I guess we could pass the auto-incremented value from the frontend...

packages/react-devtools-shared/src/devtools/ContextMenu/useContextMenu.js
bvaughn: TBH this is based on my looking at other context menu implementations. 😄 It seems to work from my testing.

gaearon: I'm confused because `client*` and `page*` are two different coordinate systems. If a page is scrolled down, `page*` will include how much you scrolled, but `client*` wouldn't.

If we use this value to calculate where to position the menu, it doesn't make sense to me that `page*` would be accurate. What matters is how far it is in the viewport — not relative to the *page*. So I would expect to see a bug there when this branch gets hit. When you say it works in testing, do you mean you specifically tested when

- the view is scrolled down, and
- we're using touch events

?

Because that's the case where I'd expect a bug.

My suggested change would be to always use `client*`, e.g. `event.clientX || (event.touches && event.touches[0].clientX)`. If we don't care to test the "touch" code path, it seems to me that we should start with what makes most sense (`client*` represents client coordinates), and only use `page*` if we have a good reason for that (my understanding is we don't have one?).

In either case, if normally we use `client*` (because most people using DevTools don't use touch), it would make sense to me to *rename the properties and arguments* to be `client*` instead of `page*`. They don't have effect on semantics but they are confusing when they don't match what you really get. E.g. I was looking at the measurement implementation and I was confident there was a bug there due to `page*` usage, but only later I found we're actually passing `e.client*` there.

packages/react-devtools-shared/src/backend/legacy/renderer.js
bvaughn: I guess the memory concern isn't all that big, since this is a debugging use case. I've added the auto-incremented behavior.

packages/react-devtools-shared/src/devtools/ContextMenu/useContextMenu.js
bvaughn: > Googling these exact lines didn't give me results.

Presumably because Prettier formatted these particular lines in a different way.

After thinking and testing a little more, I believe:
1. It probably doesn't matter which one we use, since the main DevTools window never scrolls, _but_
2. If it did scroll, we'd want to use `pageY` in both places (not just in the touch event) because otherwise our context menu would potentially be positioned too above where the cursor was.

packages/react-devtools-shared/src/backend/legacy/renderer.js
bvaughn: After some thought, I think storing separate variable can be practically useful, since you could store a value at one point, then store again after an update, then compare the two values in the console to see what changed.

bvaughn:
Okay folks. This is ready for review

packages/react-devtools-shared/src/backend/utils.js
```diff
@@ -52,3 +58,16 @@ export function copyWithSet(
   updated[key] = copyWithSet(obj[key], path, value, index + 1);
   return updated;
 }
+
+export function safeSerialize(data: any): string {
+  const cache = new Set();
+  return JSON.stringify(data, (key, value) => {
```
trueadm: Can you please explain what makes this safe?

packages/react-devtools-shared/src/devtools/ContextMenu/useContextMenu.js
```diff
@@ -0,0 +1,32 @@
+import {useContext, useEffect} from 'react';
+import {RegistryContext} from './Contexts';
+
+export default function useContextMenu({data, id, ref}) {
+  const {showMenu} = useContext(RegistryContext);
+
+  useEffect(
+    () => {
+      if (ref.current !== null) {
```
trueadm: Can ref.current change nodes between renders?

packages/react-devtools-shared/src/backend/utils.js
bvaughn: Maybe "safe" is a poor choice of words. This function protects against circular references, which would otherwise cause `JSON.stringify` to throw.

packages/react-devtools-shared/src/devtools/ContextMenu/useContextMenu.js
bvaughn: In our limited use of this hook, it would not be expected to change.

bvaughn:
Thanks for the quick feedback! I'll rename the "safe" function as soon as I can think of a better name.

packages/react-devtools-shared/src/backend/utils.js
trueadm: Ah okay, I was wondering in cases of using Symbols or functions etc, would it be safe for those too? Naming things is hard :/

bvaughn: It would be "safe" in that case, it would just not copy anything meaningful to the clipboard. (If a value can't be serialized, `JSON.stringify` would just return `undefined`.)

I renamed "safeSerialize" to "serializeToString" though and added an inline comment about its purpose.

Fausto95: That’s a cool trick @bvaughn 👌🏽

Daniel15:
Nice work!! 😄

PR #17613, opened 2019-12-15
fix(dev-tools): fix show correct displayName with forwardRef in Dev Tools

zthxxx:
allow set `displayName` after `React.forwardRef()`,

makesure Dev Tools show displayName as same as `getWrappedName` in `shared/getComponentName.js`.

It seems like a regressive bug and fixed in legacy react-devtools(https://github.com/facebook/react-devtools/pull/1154/files), but back now.

refs: 
- https://github.com/facebook/react-devtools/pull/1154/files
- emotion-js/emotion#1692

## **Before in React DevTools:**

image

## **After fixed in React DevTools**

image

### DIFF
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -364,18 +364,9 @@ export function getInternalReactConstants(
       case IndeterminateComponent:
         return getDisplayName(resolvedType);
       case ForwardRef:
-        // maybe set `displayName` after `React.forwardRef()`
-        //
-        // ```
-        // const Component = React.forwardRef(WrapComponent);
-        // Component.displayName = 'OtherName'
-        // ```
-        //
-        // `resolvedType` is `WrapComponent`
-        // `type` is Component
+        // Mirror https://github.com/facebook/react/blob/7c21bf72ace77094fd1910cc350a548287ef8350/packages/shared/getComponentName.js#L27-L37
         return (
           (type && type.displayName) ||
-          resolvedType.displayName ||
           getDisplayName(resolvedType, 'Anonymous')
         );
       case HostRoot:
@@ -388,11 +379,7 @@ export function getInternalReactConstants(
         return null;
       case MemoComponent:
       case SimpleMemoComponent:
-        if (elementType.displayName) {
-          return elementType.displayName;
-        } else {
-          return getDisplayName(resolvedType, 'Anonymous');
-        }
+        return getDisplayName(resolvedType, 'Anonymous');
       case SuspenseComponent:
         return 'Suspense';
       case SuspenseListComponent:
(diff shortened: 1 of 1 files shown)


zthxxx:
CLA signed.

zthxxx:
cc @gaearon

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -364,8 +364,19 @@ export function getInternalReactConstants(
       case IndeterminateComponent:
         return getDisplayName(resolvedType);
       case ForwardRef:
+        // maybe set `displayName` after `React.forwardRef()`
+        //
+        // ```
+        // const Component = React.forwardRef(WrapComponent);
+        // Component.displayName = 'OtherName'
```
nihgwu: I don't know how this pattern came, I used this way to define the displayName for ref-forwarded component before too, and found it didn't work, and then I read the documentation carefully and can't found that kind of usage https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools, so I changed my way to assign the displayName in my components

nihgwu: Yes I agree we should align the logic with https://github.com/facebook/react/blob/master/packages/shared/getComponentName.js#L34, that will make life much easier, in the meantime do we need to update the doc for this case

zthxxx: I know this way, it's so ugly.

And the reason why this docs exists is due to this bug.

Thinking what scene you need to use `React.forwardRef`, 
HOC, right?

So how can you set displayName after HOC returned in your code or babel?

nihgwu: For HOC I updated as the documented way
image

zthxxx: Your code's result I have wrote [at the beginning of this PR](https://github.com/facebook/react/pull/17613#issue-353278549)

All of HOC returned components have the same name.

Example, some actual HOC code in [emotion](https://github.com/emotion-js/emotion/blob/master/packages/core/src/context.js#L20-L37)

and the used: 
- https://github.com/emotion-js/emotion/blob/master/packages/core/src/jsx.js#L98-L112
- https://github.com/emotion-js/emotion/blob/master/packages/styled-base/src/index.js#L81
- https://github.com/emotion-js/emotion/blob/master/packages/core/src/global.js#L24-L55

zthxxx:
Thanks for your works, I think this requires a new release for [React DevTools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en).

zthxxx:
> Thanks for your works, I think this requires a new release for [React DevTools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en).

@bvaughn 
image

bvaughn:
Screen Shot 2019-12-29 at 8 47 43 AM

No need to keep pinging me. I've submitted an updated version, but I can't speed up the Chrome review process. :smile:

PR #17622, opened 2019-12-16
Test if circleci has problems with branch name

necolas:
Just testing if my branch names are causing problems (as this was a cause once upon a time)


necolas:
Nope, still only running one check

PR #17625, opened 2019-12-16
Remove renderPhaseUpdates Map

acdlite:
Follow up to #17484, which was reverted due to a bug found in www. ~~(I haven't actually found this bug yet, still trying to repro.)~~ Found and fixed

Also addresses a separate bug where updates that are thought to be render phase updates are cleared, but they are actually normal updates: https://github.com/facebook/react/commit/8a347ed024159a5307172ea633a4561160f4a6b9#r36440099

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -177,15 +177,10 @@ let currentlyRenderingFiber: Fiber = (null: any);
 let currentHook: Hook | null = null;
 let workInProgressHook: Hook | null = null;
 
-// Updates scheduled during render will trigger an immediate re-render at the
-// end of the current pass. We can't store these updates on the normal queue,
-// because if the work is aborted, they should be discarded. Because this is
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -509,23 +505,21 @@ export function resetHooksAfterThrow(): void {
...
-        }
-        hook = hook.next;
+    // There were render phase updates. These are only valid for this render
+    // phase, which we are now aborting. Remove the updates from the queues so
+    // they do not persist to the next render. Do not remove updates from hooks
+    // that weren't processed.
+    //
+    // Only reset the updates from the queue if it has a clone. If it does
+    // not have a clone, that means it wasn't processed, and the updates were
+    // scheduled before we entered the render phase.
```
acdlite: @sebmarkbage I think if we change the list to an array, we would just need some other mechanism to determine if it was processed or not. I added tests so we don't forget.

acdlite:
Ok I think this is ready for review now

Issue #17630, opened 2019-12-16
Fix broken DevTools tests


bvaughn:
PR #17599 broke `yarn test-build-devtools`. (b6c423daadaa35da3f34048628df9635505eecb1 is not broken,  0cf22a56a18790ef34c71bef14f64695c0498619 is broken)
```
    Configuration error:
    
    Could not locate module shared/consoleWithStackDev mapped as:
    /Users/bvaughn/Documents/git/react/build/node_modules/shared/consoleWithStackDev.
    
    Please check your configuration for these entries:
    {
      "moduleNameMapper": {
        "/^shared\/([^\/]+)$/": "/Users/bvaughn/Documents/git/react/build/node_modules/shared/$1"
      },
      "resolver": null
    }
```

Issue #17629, opened 2019-12-16
Update DevTools build process to use artifacts from CI


bvaughn:
Building (or even testing) DevTools currently requires building lots of artifacts locally e.g.
```
RELEASE_CHANNEL=experimental \
  yarn build \
    -- react/index,react-dom,react-is,react-debug-tools,scheduler \
    --type=NODE
```

To run the DevTools unit tests, you would also need to add `react-test-renderer` to the above list of targets.

### Pros
* Enables using DevTools to test local changes to react/renderers.

### Cons
* Slow to build renderers.
* Requires installing Java (which causes Mozilla testers to complain).
* Error message when you forget to build (or have an incomplete build) is very confusing when trying to e.g. build or run tests.

Maybe we should change the default build process to pull the latest (or a specific) build artifact from CI instead?

bvaughn:
I guess we could leverage the pre-existing release script for this.

1. **Step 1**: Build DevTools bundle and note artifact ID
```
$ ./scripts/release/prepare-canary.js 
✓ Determining latest Circle CI for the master branch
✓ Downloading artifacts from Circle CI for build 68717 
```

2. **Step 2**: Include artifact ID in instructions to Mozilla
```
./scripts/release/prepare-canary.js --build=68717
```

bvaughn:
Note that we would also need to update the `prepare-canary` command (if we used to) to make sure to pull the _experimental_ build, since DevTools depends on the `createRoot` API.

PR #17631, opened 2019-12-16
Fix test_build_devtools CI job to run test-build-devtools

gaearon:
We're testing the wrong thing in this job. Probably bad copy paste.
As a result, regressions like https://github.com/facebook/react/pull/17599 creep in.

First commit is a small fix that fixes running it locally after a package got renamed. (The command crashed for me locally otherwise, which made verifying the fix annoying.)

The second commit makes the right command run on CI. It fails, as expected.

The third commit fixes https://github.com/facebook/react/issues/17630.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -277,7 +277,7 @@ jobs:
       - *run_yarn
       - run:
           environment:
-            RELEASE_CHANNEL: stable
+            RELEASE_CHANNEL: experimental
           command: yarn test-build-devtools --maxWorkers=2
 
   test_dom_fixtures:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -376,7 +376,7 @@ workflows:
             - build
       - test_build_devtools:
           requires:
-            - build
+            - build_experimental
       - test_dom_fixtures:
           requires:
             - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -278,7 +278,7 @@ jobs:
       - run:
           environment:
             RELEASE_CHANNEL: stable
-          command: yarn test-build --maxWorkers=2
+          command: yarn test-build-devtools --maxWorkers=2
 
   test_dom_fixtures:
     docker: *docker
--- a/packages/react-devtools-shared/src/__tests__/inspectedElementContext-test.js
+++ b/packages/react-devtools-shared/src/__tests__/inspectedElementContext-test.js
@@ -514,6 +514,7 @@ describe('InspectedElementContext', () => {
     const arrayOfArrays = [[['abc', 123, true], []]];
     const div = document.createElement('div');
     const exampleFunction = () => {};
+    const exampleDateISO = '2019-12-31T23:42:42.000Z';
     const setShallow = new Set(['abc', 123]);
     const mapShallow = new Map([['name', 'Brian'], ['food', 'sushi']]);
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -278,7 +278,7 @@ jobs:
       - run:
           environment:
             RELEASE_CHANNEL: stable
-          command: yarn test-build --maxWorkers=2
+          command: yarn test-build-devtools --maxWorkers=2
```
bvaughn: lol whoops

scripts/jest/preprocessor.js
```diff
@@ -68,11 +68,14 @@ module.exports = {
...
+      const isInDevToolsPackages = !!filePath.match(
+        /\/packages\/react-devtools.*\//
+      );
       const testOnlyPlugins = [pathToBabelPluginAsyncToGenerator];
-      const sourceOnlyPlugins =
-        process.env.NODE_ENV === 'development'
-          ? [pathToBabelPluginReplaceConsoleCalls]
-          : [];
+      const sourceOnlyPlugins = [];
+      if (process.env.NODE_ENV === 'development' && !isInDevToolsPackages) {
```
bvaughn: ❤️

packages/react-devtools-shared/src/__tests__/inspectedElementContext-test.js
```diff
@@ -542,7 +543,7 @@ describe('InspectedElementContext', () => {
           // eslint-disable-next-line no-undef
           big_int={BigInt(123)}
           data_view={dataView}
-          date={new Date(123)}
+          date={new Date(exampleDateISO)}
```
bvaughn: 🤦‍♂

gaearon:
OK this one should actually fix it.

bvaughn:
Hm...looks like this has been broken for a while (since the stable/experimental stuff was split out) and we just haven't noticed. I think to build a version of React that DevTools can consume now we would need to run:
```
RELEASE_CHANNEL=experimental yarn build -- react/index,react-dom,react-is,react-debug-tools,scheduler,react-test-renderer --type=NODE
```
(Or just download the latest experimental CI artifact, which might be better.)

gaearon:
Do you mean we need to change the CI command to do this too?

bvaughn:
Since DevTools and its tests depend on the `createRoot` API, which is now only exported for experimental builds, yeah. We'd need to make sure we were feeding it an experimental build.

I think to do this, we'd need to make `test_build_devtools` depend on the `build_experimental` target rather than the current `build` target.

gaearon:
On CI, we already have `test_build_devtools` specify `build` job as a dependency so I don't think there's further action necessary. For local testing it would be nice to do this — or at least to have something warning you when you forget to.

gaearon:
>Since DevTools and its tests depend on the createRoot API, which is now only exported for experimental builds, yeah. We'd need to make sure we were feeding it an experimental build.

This is what I did in https://github.com/facebook/react/pull/17631/commits/2049b9438e9298a1350cf6c62b38d286fce2588f.

gaearon:
Ohh I see what you're saying. That this wouldn't be sufficient.

bvaughn:
Yeah, I think to get tests passing we need to use the experimental build command. Then I'll I'll follow up on the rest of this (the actual build/release process) as part of #17629

bvaughn:
Thanks Dan! <3

gaearon:
My "pleasure"!

PR #17632, opened 2019-12-17
Additional test infra changes for toErrorDev rename

gaearon:
I missed these two.


gaearon:
I'll just get it in — seems straightforward and don't want to regress for pending PRs.

Issue #17626, opened 2019-12-16
react-refresh + ReactDOM: hot reloading only works when bundling React


mxmul:
**Do you want to request a *feature* or report a *bug*?**

bug?

**What is the current behavior?**

_note: I am encountering this issue when using https://github.com/pmmmwh/react-refresh-webpack-plugin, but I believe it's an issue with react-refresh itself._

The react-refresh runtime [overrides](https://github.com/facebook/react/blob/7c21bf72ace77094fd1910cc350a548287ef8350/packages/react-refresh/src/ReactFreshRuntime.js#L459) `__REACT_DEVTOOLS_GLOBAL_HOOK__.inject` to get a reference to the React renderer. In my app, the `inject` method is never called however, because I load react/react-dom from a third-party CDN before my application code. This means that changed components are never actually refreshed in the DOM.

I believe the issue is that scripts are loaded in this order:

1. react-devtools sets up the global hook
2. react/react-dom are loaded on the page, and `inject()` is called
3. user code (which is instrumented with the react-refresh babel plugin) is loaded on the page, and `inject()` is monkey-patched

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

Follow the usage steps in the https://github.com/pmmmwh/react-refresh-webpack-plugin repo, but also add `react` and `react-dom` as externals in your Webpack build:
```
module.exports = {
  //...
  externals: {
    react: 'React',
    'react-dom': 'ReactDOM'
  }
};
```
then load those scripts onto the page from a CDN (e.g. unpkg or cdnjs) before the Webpack bundle.

**What is the expected behavior?**

`react-refresh` works the same whether React/ReactDOM are bundled with application code, or loaded via an external script.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

16.8.6 / Chrome / osx

gaearon:
I think one possible fix is to add some logic here:

https://github.com/facebook/react/blob/36a6e29bb3eead85e3500ba7269cbcd55516a8fb/packages/react-refresh/src/ReactFreshRuntime.js#L459

[...]

Issue #17552, opened 2019-12-09
react-refresh load from CDN?


otakustay:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

When `react-dom` is loaded from CDN like `<script src="https://cdn.jsdelivr.net/npm/react-dom@16.12.0/umd/react-dom.development.js"></script>`, `react-refresh` failed to inject hook into devtools.

I've created a related issue here: https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/13

We should find a way to invoke `injectIntoGlobalHook` function from `react-refresh/runtime`, however this file is in cjs format so we cannot currently do this in a simple way.

**What is the expected behavior?**

I'd like `react-refresh` to publish runtime as a umd bundle so we can reference it from CDN and put it before `react-dom`'s `<script>` element, then invoke `injectIntoGlobalHook` in the right place.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

- `react-dom`: 16.12.0
- `react-refresh`: 0.7.0

otakustay:
Is it possible that react-dom handle injection lazy on the first render call? This way externals tools can hook injections more easily

gaearon:
Can you check if https://github.com/facebook/react/pull/17633 fixes it?

gaearon:
I verified #17633 fixes this.

gaearon:
Fixed in `react-refresh@0.7.1`.

thetrime:
I realise this issue is already over a year old, but I'm not sure if this is really fixed. If you load react-dom first, then injectInternals() will return immediately if __REACT_DEVTOOLS_GLOBAL_HOOK__ is not defined. If we then load react-refresh, it will check if __REACT_DEVTOOLS_GLOBAL_HOOK__ is defined, and if not, define it by itself, but there doesn't seem to be any attempt to _re_-inject the hooks. The react-dom code doesn't set injectedHook except in injectInternals(), and if injectedHook is null then onCommitRoot() will have no effect.

The upshot is that loading react-dom as an external from a webpack project means it'll always load before react-refresh if that's bundled in (and it seemingly has to be, since I don't see it on a CDN).

[...]

PR #17633, opened 2019-12-17
[Fast Refresh] Support injecting runtime after renderer executes

gaearon:
I think this fixes https://github.com/facebook/react/issues/17626 and https://github.com/facebook/react/issues/17552.

I was assuming that Fresh runtime always executes before the renderer. But on the web this won't be the case if you load React from CDN and *then* run your webpack bundle (https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/13). And separating `<script>` tags can be annoying and complicated.

Instead, we can handle the simple case — when ReactDOM executes before the webpack bundle — gracefully. The DevTools global hook already has a [renderers map](https://github.com/facebook/react/blob/36a6e29bb3eead85e3500ba7269cbcd55516a8fb/packages/react-devtools-shared/src/hook.js#L280) so all we need to connect to them is to read from it. In fact that's how DevTools backends themselves [connect to already injected renderers](https://github.com/facebook/react/blob/36a6e29bb3eead85e3500ba7269cbcd55516a8fb/packages/react-devtools-shared/src/backend/index.js#L85-L86).

This still doesn't handle the case where you mount some roots *before* injecting the runtime. In that case they won't be "visible" to Fresh. Arguably this is more of a corner case and I wouldn't bother solving this yet. The webpack plugin should be able to ensure that injection happens before the actual app code runs.

I added a regression test.

### DIFF
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -435,6 +435,7 @@ export function injectIntoGlobalHook(globalObject: any): void {
       // Otherwise, the renderer will think that there is no global hook, and won't do the injection.
       let nextID = 0;
       globalObject.__REACT_DEVTOOLS_GLOBAL_HOOK__ = hook = {
+        renderers: new Map(),
         supportsFiber: true,
         inject(injected) {
           return nextID++;
@@ -467,6 +468,19 @@ export function injectIntoGlobalHook(globalObject: any): void {
       }
       return id;
     };
+
+    // Do the same for any already injected roots.
+    // This is useful if ReactDOM has already been initialized.
+    // https://github.com/facebook/react/issues/17626
+    hook.renderers.forEach((injected, id) => {
+      if (
+        typeof injected.scheduleRefresh === 'function' &&
+        typeof injected.setRefreshHandler === 'function'
+      ) {
+        // This version supports React Refresh.
+        helpersByRendererID.set(id, ((injected: any): RendererHelpers));
+      }
+    });
 
     // We also want to track currently mounted roots.
     const oldOnCommitFiberRoot = hook.onCommitFiberRoot;
--- a/packages/react-devtools-shared/src/hook.js
+++ b/packages/react-devtools-shared/src/hook.js
@@ -277,6 +277,8 @@ export function installHook(target: any): DevToolsHook | null {
   const hook: DevToolsHook = {
     rendererInterfaces,
     listeners,
+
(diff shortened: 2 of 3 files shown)


packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -435,6 +435,7 @@ export function injectIntoGlobalHook(globalObject: any): void {
       // Otherwise, the renderer will think that there is no global hook, and won't do the injection.
       let nextID = 0;
       globalObject.__REACT_DEVTOOLS_GLOBAL_HOOK__ = hook = {
+        renderers: new Map(),
```
gaearon: This is just so the lines below don't crash when we take this code branch. This particular Map isn't actually used further.

packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -468,6 +469,19 @@ export function injectIntoGlobalHook(globalObject: any): void {
       return id;
     };
 
+    // Do the same for any already injected roots.
```
gaearon: Note this is almost copy paste from the code above (462–468). I could extract a function but meh.

packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -468,6 +469,19 @@ export function injectIntoGlobalHook(globalObject: any): void {
...
+    // This is useful if ReactDOM has already been initialized.
+    // https://github.com/facebook/react/issues/17626
+    hook.renderers.forEach((injected, id) => {
+      if (
+        typeof injected.scheduleRefresh === 'function' &&
+        typeof injected.setRefreshHandler === 'function'
+      ) {
```
mxmul: I've applied these changes locally in my application (I had a pretty identical local branch going 😅), and it doesn't seem to work. The renderer instance has none of these properties at this point (react @ 16.8.6):

Screen Shot 2019-12-16 at 5 36 32 PM

gaearon: Can you make a minimal reproducing project?

gaearon: @mxmul Oh wait, this is expected — these were only added in 16.10 or so. You need to try with the latest version.

leidegre:
I can't get it to work unless I enable React Development Tools, is this expected behavior? Not a big deal but I can't seem to get it to do it's magic unless React Development Tools is runnig.

gaearon:
Sorry, not enough details. Please file a new issue.

PR #17640, opened 2019-12-17
Move DevTools CI job to experimental workflow

gaearon:
Before https://github.com/facebook/react/pull/17631, each commit would run 22 jobs on master:

Screen Shot 2019-12-17 at 6 13 05 PM

After it, only 10 jobs run on master:

Screen Shot 2019-12-17 at 6 13 00 PM

I can't figure out why, but putting DevTools build in the "stable" workflow is the only suspicious thing I can see in that PR. So let's try to change that and see if it helps.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -374,9 +374,6 @@ workflows:
       - test_build_prod:
           requires:
             - build
-      - test_build_devtools:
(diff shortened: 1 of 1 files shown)


gaearon:
This looks like it fixes the issue.

Issue #17389, opened 2019-11-17
Devtools resize Component State/Props pane


codercodingthecode:
**Do you want to request a *feature* or report a *bug*?**
I would like to request a feature

**What is the current behavior?**
Resizing the state/props pane in Component Tab

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

**What is the expected behavior?**
Allow resizing pane

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
Devtools 4.2.0 and Chrome Version 78.0.3904.97 

![image](https://user-images.githubusercontent.com/20182536/69003149-de088780-08c2-11ea-9f32-391faee6a557.png)

codercodingthecode:
I'd like to open a pr for this

bvaughn:
I'd be happy to review a PR for this.

hristo-kanchev:
Hey @bvaughn !

Are we still looking at adding this feature?

Maybe we can have a checkbox in Settings -> General that will allows the user to dock the Component State/Props/.. to the bottom instead of supporting resizing for that pane?

What do you think?

bvaughn:
Resizability, yes. Would still be open to adding that behavior in.

A checkbox to force a bottom docking position- not really a fan of that.

hristo-kanchev:
Cool @bvaughn. I'll try and open a PR by the end of this week. I'll ping you again when it's done.

bvaughn:
This issue is all yours! 😄

I've added the **"good first issue (taken)"** label so that others will know not to start work on the issue. If you change your mind about the issue, no worries! Just let me know so that I can remove the label and free it up for someone else to claim.

Cheers!

hristo-kanchev:
****Progress** so far:**

![horizontal-resize](https://user-images.githubusercontent.com/23095052/74434192-2b64a880-4e62-11ea-826c-3eef2a4fc10c.gif)

I have a question though, @bvaughn.

Should we also support vertical resizing when we are in this mode?

<img width="500" alt="Screenshot 2020-02-13 at 13 09 27" src="https://user-images.githubusercontent.com/23095052/74434236-48997700-4e62-11ea-8616-93d80509a960.png">

Thoughts?

[...]

PR #17643, opened 2019-12-18
Add DevTool Extensions Inspected Panel x-direction scrollbar

moonrailgun:
For someone who keep chrome devtool in right, width is always not enough
fix no scrollbar for Inspected Panel so we can drag it to view more info

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.css
+++ b/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.css
@@ -72,11 +72,17 @@
 }
 
 .InspectedElement {
-  overflow-x: hidden;
+  flex: 1 1 auto;
+  overflow-x: auto;
   overflow-y: auto;
   font-family: var(--font-family-monospace);
   font-size: var(--font-size-monospace-normal);
   line-height: var(--line-height-data);
+}
+
+.InspectedElementContent {
+  width: 100%;
+  min-width: max-content;
 }
 
 .Owner {
--- a/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
@@ -353,62 +353,64 @@ function InspectedElementView({
 
   return (
     <div className={styles.InspectedElement}>
-      <HocBadges element={element} />
-      <InspectedElementTree
-        label="props"
-        data={props}
-        inspectPath={inspectPropsPath}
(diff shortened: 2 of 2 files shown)


bvaughn:
Thanks for the POC and the suggestion!

This isn't a change I want to make though. I think overall, it would be worse for UX to introduce horizontal overflow into this panel. I don't think it's very intuitive, and it also negative impacts things like the HOC badges (which used to wrap but would now scroll).

I understand that limited space makes dev tooling more difficult to work with. This is why the React DevTools relayout when smaller than 600 pixels wide so that the selected element is *beneath* the component tree rather than beside it. I would be open to discussing additional potential responsive solutions like this if you have ideas!

moonrailgun:
As a Tool. I think is necessary to show more information rather than UX.

First, scrollbar is optional if devtool have more space to show. 

Second, I cant even work if its a deep data. We have to resize all chrome devtool panel to trigger 600px width and make `Inspected` panel show more information. Based on intuition, user should make panel width wider rather than narrower (but its looks like more effective to make it under 600px). And in this way, page maybe trigger more unuse DOM update and rerender.

Finally, I know scrollbar is not a good idea, but its a effective, actually, and simple solution to fix its problem. and no side-effect to other scene.  
 I think the perfect solution of it is we can handle panel size manual, just like normal chrome dom panel and style panel.

All in all, thanks for your review.

bvaughn:
> Second, I cant even work if its a deep data. We have to resize all chrome devtool panel to trigger 600px width and make Inspected panel show more information. Based on intuition, user should make panel width wider rather than narrower

Sounds like you dock your devtools to the side of the browser, rather than the bottom. Have you considered docking to the bottom as a way of getting more width? (Side docking has always felt too cramped to be useful, to me.)

> I think the perfect solution of it is we can handle panel size manual, just like normal chrome dom panel and style panel.

That seems like a better solution to this :smile: #17389

moonrailgun:
> Sounds like you dock your devtools to the side of the browser, rather than the bottom. Have you considered docking to the bottom as a way of getting more width? (Side docking has always felt too cramped to be useful, to me.)

Yep, I like make my devtool to the side. its helpful to remind me to take care of responsive layout. This is personal preference, dont go into seriously. 

Maybe you should merge this PR to fit this scene, its no affect to other scene. At least better than cant see more information. We should implement this issue in next step(this feature have more change and i cant implement it immediately, i will read code and implement it later if you dont want take more attention in it). 

Thanks for your review, React is a very nice framework

PR #17644, opened 2019-12-18
[react-interactions] Re-label Flare as deprecated and remove from RN

trueadm:
This PR cleans up a bunch of stuff in our code base relating to React Flare. Specifically, it renames all the Flare call-sites to be "DEPRECATED" so it's clear that these APIs are going to go away at some point and shouldn't be used internally. It also removes React Flare from RN – where it was never used internally.

If we are to move forward with other event system experiments, this makes it clearer where the old system (Flare) lives and where other systems live; whereas they might have been confused previously.


NE-SmallTown:
Would you mind tell a bit about the relationship between this PR and https://github.com/facebook/react/pull/17508? IMO, just like #17508 said, it learned from react-flare, but seems the new API like `unstable_createRootListener` is also the API in react-flare? If so, when we deprecate that,  is that indicate the experimental Listener API in #17508 has some new thoughts/change? Plz tell me if I'm wrong

trueadm:
@NE-SmallTown React Flare and the Listener API (as per that other PR) are separate event systems that have entirely different designs. There's some overlap in how things work, but that's more a side-effect of those things being positives taken from React Flare that we want to keep in the future. We are still revising things, so the Listener API PR might also change too, thus why we have not merged it.

NE-SmallTown:
Gotcha, thanks~

PR #17651, opened 2019-12-18
[react-interactions] Add Listener API + useEvent hook

trueadm:
Note: This API is intentionally meant to be a low-level way of creating events and assigning listeners to them. It's meant to be verbose so larger building blocks can be created on top of them.

This PR is an alternative solution and system to that of my other PR: https://github.com/facebook/react/pull/17508. Specifically, based off feedback internally, I've tried to tackle some of the problems that were brought up with the `createListener` approach in https://github.com/facebook/react/pull/17508:

- `createListener` largely depended on events being registered in the commit phase, meaning that there would likely be issues around needing to flush more to ensure we register DOM events. The new approach enforces all events are registered unconditionally via hooks in the render phase, mitigating this issue.
- `createListener` allowed for listeners to update and change their properties between renders, which again is problematic for performance and would also require more flushes to ensure we have committed the latest version of each listener.
- `createListener` had a complex diffing process to ensure we stored the latest listeners, but this meant that the process had additional overhead and memory usage – which is no longer the case with this PR.
- `createListener` required listeners to be put on nodes via the `listeners` prop. Furthermore, it required using arrays to combine multiple listeners, which some felt was not idealistic and might be confusing during debugging as to which listeners occurred at which stages. Also, there was general dislike to introducing another internal prop – as it would mean we'd have to first forbid `listeners` and wait for React 17 to introduce these APIs, as they might be used in the wild for other reasons (custom elements).
- `createListener` didn't provide an idiomatic way to conditionally add/remove root events (they're called delegated events in this new PR). With the new approach, there's a streamlined approach to ensure this is easier to do, and by default, no root events can be added unconditionally, which is a code-smell and a good cause of memory leaks/performance issues.

Taking the above points into consideration, the design of this new event system aims at bringing the same capabilities as described in https://github.com/facebook/react/pull/17508 whilst also providing some other nice features, that should allow for bigger event sub-systems to be built on top.

## `ReactDOM.useEvent`

This hook allows for the registration to a native DOM event, similar to that of `addEventListener` on the web. `useEvent` takes a given event `type` and registers it to the DOM then returns an object unique to that event that allows listeners to be attached to their targets in an effect or another event handler. The object provides three different methods to setup and handle listeners:

- `setListener(target: window | element, listener: ?(Event => void))` set a listener to be active for a given DOM node. The node must be a DOM node managed by React or it can be the `window` node for delegation purposes. If the listener is `null` or `undefined` then we remove the given listener for that DOM node or `window` object.
- `clear()` remove all listeners

The hook takes three arguments:

- `type` the DOM event to listen to
- `options` an optional object allowing for additional properties to be defined on the event listener. The options are:
  - `passive` provide an optional flag that tells the listener to register a passive DOM event listener or an active DOM event listener
  - `capture` provide an optional flag that tells the listener callback to fire in the capture phase or the bubble phase
  - `priority` provide an optional Scheduler priority that allows React to correct schedule the listener callback to fire at the correct priority.

## Note

For propagation, the same rules of `stopPropagation` and `stopImmediatePropagation` apply to these event listeners. These methods are actually monkey-patched, as we use the actual native DOM events with this system and API, rather than Synthetic Events. `currentTarget` and `eventPhase` are also respectfully monkey-patched to co-ordinate and align with the propagation system involved internally within React.

Furthermore, all event listeners are passive by default. If is desired to called `event.preventDefault` on an event listener, then the event listener should be made active via the `passive` option.

## Examples

An example of a basic clickable button:

```jsx
import {useRef, useEffect} from 'react';
import {useEvent} from 'react-dom';

function Button({children, onClick}) {
  const buttonRef = useRef(null);
  const clickEvent = useEvent('click');

  useEffect(() => {
    clickEvent.setListener(buttonRef.current, onClick);
  });

  return <button ref={buttonRef}>{children}</button>;
}
```

If you want to listen to events that are delegated to the window, you can do that:

```jsx
import {useRef, useEffect} from 'react';
import {useEvent} from 'react-dom';

function Button({children, onClick}) {
  const clickEvent = useEvent('click');

  useEffect(() => {
    // Pass in `window`, which is supported with this API
    // Note: `window` is not supported, as React doesn't
    // listen to events that high up.
    clickEvent.setListener(window, onClick);
  });

  return <button>{children}</button>;
}
```

If you wanted to extract the verbosity out of this into a custom hook, then it's possible to do so:

```jsx
import {useRef, useEffect} from 'react';
import {useEvent} from 'react-dom';

function useClick(ref, onClick) {
  const clickEvent = useEvent('click');

  useEffect(() => {
    clickEvent.setListener(ref.current, onClick);
  });

  return buttonRef;
}

function Button({children}) {
  const buttonRef = useRef(null);
  useClick(buttonRef , () => {
    console.log('Hello world!')
  });
  return <button ref={buttonRef}>{children}</button>;
}
```

A more complex button that tracks when the button is being pressed with the mouse:

```jsx
import {useRef, useEffect} from 'react';
import {useEvent} from 'react-dom';

function Button({children, onClick}) {
  const buttonRef = useRef(null);
  const [pressed, setPressed] = useState(false);
  const click = useEvent('click');
  const mouseUp = useEvent('mouseup');
  const mouseDown = useEvent('mousedown');

  useEffect(() => {
    const button = buttonRef.current;

    const handleMouseUp = () => {
      setPressed(false);
    };
    const handleMouseDown = () => {
      setPressed(true); 
    };

    click.setListener(button, onClick);
    if (pressed) {
      mouseUp.setListener(button, handleMouseUp);
    } else {
      mouseDown.setListener(button, handleMouseDown);
    }

    return () => {
      click.setListener(button, null);
      mouseDown.setListener(button, null);
      mouseUp.setListener(button, null);
    };
  }, [pressed, onClick]);

  return (
    <button ref={buttonRef} className={pressed && 'pressed'}>
      {children}
    </button>
  );
}
```

## What about the DOM's `element.addEventListener`?

In many respects, this low-level API was intentionally designed to be a replacement for the DOM's `addEventListener`. Not only does this new Listener API provide many nice benefits, like auto-recycling listeners on unmount, but it should also help prevent bugs that will likely occur when `addEventListener` is used in conjunction with Concurrent Mode.

For a detailed list of differences, here are just some of the key benefits of the Listener API vs the DOM's `addEventListener`:

- The event listeners automatically get recycled upon unmount
- The event listeners correctly batch updates and flush previous ones correctly, as well as co-ordinating priority levels with the internal scheduler
- The event listeners allow for alignment with the propagation of other events in the React event system, something not possible with manual DOM event listeners
- The event listeners will correctly work with React Portals, Suspense and Concurrent Mode, native event listeners do not.
- By ensuring we use a hook, like useEvent, we can determine the event needed during server-side render and ensure those events are registered on initial hydration ahead of time. This makes it possible to "replay" those events before the components themselves have concurrently hydrated on the client.

### DIFF
--- a/packages/legacy-events/EventSystemFlags.js
+++ b/packages/legacy-events/EventSystemFlags.js
@@ -11,8 +11,9 @@ export type EventSystemFlags = number;
 
 export const PLUGIN_EVENT_SYSTEM = 1;
 export const RESPONDER_EVENT_SYSTEM = 1 << 1;
-export const IS_PASSIVE = 1 << 2;
-export const IS_ACTIVE = 1 << 3;
-export const PASSIVE_NOT_SUPPORTED = 1 << 4;
-export const IS_REPLAYED = 1 << 5;
-export const IS_FIRST_ANCESTOR = 1 << 6;
+export const LISTENER_EVENT_SYSTEM = 1 << 2;
+export const IS_PASSIVE = 1 << 3;
+export const IS_ACTIVE = 1 << 4;
+export const PASSIVE_NOT_SUPPORTED = 1 << 5;
+export const IS_REPLAYED = 1 << 6;
+export const IS_FIRST_ANCESTOR = 1 << 7;
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -65,6 +65,8 @@ import {
 import {
   addResponderEventSystemEvent,
   removeActiveResponderEventSystemEvent,
+  addListenerSystemEvent,
+  removeListenerSystemEvent,
 } from '../events/ReactDOMEventListener.js';
 import {mediaEventTypes} from '../events/DOMTopLevelEventTypes';
 import {
@@ -89,6 +91,7 @@ import {validateProperties as validateUnknownProperties} from '../shared/ReactDO
 import {
   enableDeprecatedFlareAPI,
   enableTrustedTypesIntegration,
+  enableListenerAPI,
 } from 'shared/ReactFeatureFlags';
 
 let didWarnInvalidHydration = false;
@@ -1337,6 +1340,42 @@ export function listenToEventResponderEventTypes(
         );
         listenerMap.set(eventKey, eventListener);
       }
+    }
+  }
+}
+
+export function listenToEventListener(
+  type: string,
+  passive: boolean,
+  document: Document,
+): void {
+  if (enableListenerAPI) {
+    // Get the listening Map for this element. We use this to track
+    // what events we're listening to.
+    const listenerMap = getListenerMapForElement(document);
+    const passiveKey = type + '_passive';
+    const activeKey = type + '_active';
+    const eventKey = passive ? passiveKey : activeKey;
+
+    if (!listenerMap.has(eventKey)) {
+      if (passive) {
+        if (listenerMap.has(activeKey)) {
+          // If we have an active event listener, do not register
+          // a passive event listener. We use the same active event
+          // listener.
+          return;
+        } else {
+          // If we have a passive event listener, remove the
+          // existing passive event listener before we add the
+          // active event listener.
+          const passiveListener = listenerMap.get(passiveKey);
+          if (passiveListener != null) {
+            removeListenerSystemEvent(document, type, passiveListener);
+          }
+        }
+      }
+      const eventListener = addListenerSystemEvent(document, type, passive);
+      listenerMap.set(eventKey, eventListener);
     }
   }
 }
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -1344,15 +1344,14 @@ export function listenToEventResponderEventTypes(
   }
 }
 
-export function listenToEventListener(
+export function listenToListenerSystemEvent(
   type: string,
   passive: boolean,
-  document: Document,
 ): void {
   if (enableListenerAPI) {
     // Get the listening Map for this element. We use this to track
     // what events we're listening to.
-    const listenerMap = getListenerMapForElement(document);
+    const listenerMap = getListenerMapForElement(window);
     const passiveKey = type + '_passive';
     const activeKey = type + '_active';
     const eventKey = passive ? passiveKey : activeKey;
@@ -1374,7 +1373,7 @@ export function listenToEventListener(
           }
         }
       }
-      const eventListener = addListenerSystemEvent(document, type, passive);
+      const eventListener = addListenerSystemEvent(window, type, passive);
       listenerMap.set(eventKey, eventListener);
     }
   }
--- a/packages/react-dom/src/events/DOMEventProperties.js
+++ b/packages/react-dom/src/events/DOMEventProperties.js
@@ -223,3 +223,17 @@ export function getEventPriorityForPluginSystem(
   // for the event.
   return priority === undefined ? ContinuousEvent : priority;
 }
+
+export function getEventPriorityForListenerSystem(type: string): EventPriority {
+  const priority = eventPriorities.get(((type: any): TopLevelType));
+  if (priority !== undefined) {
+    return priority;
+  }
+  if (__DEV__) {
+    console.warn(
+      'The event "type" provided to useEffect() does not have a known priority type.' +
+        ' It is recommended to provide a "priority" option to specify a priority.',
+    );
+  }
+  return ContinuousEvent;
+}
--- a/packages/react-dom/src/events/DOMEventProperties.js
+++ b/packages/react-dom/src/events/DOMEventProperties.js
@@ -231,7 +231,7 @@ export function getEventPriorityForListenerSystem(type: string): EventPriority {
   }
   if (__DEV__) {
     console.warn(
-      'The event "type" provided to useEffect() does not have a known priority type.' +
+      'The event "type" provided to useEvent() does not have a known priority type.' +
         ' It is recommended to provide a "priority" option to specify a priority.',
     );
   }
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -33,6 +33,7 @@ import {
   Passive as PassiveEffect,
 } from 'shared/ReactSideEffectTags';
 import {
+  NoEffect as NoHookEffect,
   HasEffect as HookHasEffect,
   Layout as HookLayout,
   Passive as HookPassive,
@@ -1600,6 +1601,7 @@ const HooksDispatcherOnRerender: Dispatcher = {
   useResponder: createDeprecatedResponderListener,
   useDeferredValue: rerenderDeferredValue,
   useTransition: rerenderTransition,
+  useEvent: updateEventListener,
 };
 
 let HooksDispatcherOnMountInDEV: Dispatcher | null = null;
@@ -2115,6 +2117,11 @@ if (__DEV__) {
       updateHookTypesDev();
       return rerenderTransition(config);
     },
+    useEvent(event: ReactListenerEvent): ReactListenerMap {
+      currentHookNameInDev = 'useEvent';
+      updateHookTypesDev();
+      return updateEventListener(event);
+    },
   };
 
   InvalidNestedHooksDispatcherOnMountInDEV = {
@@ -2519,6 +2526,12 @@ if (__DEV__) {
       warnInvalidHookAccess();
       updateHookTypesDev();
       return rerenderTransition(config);
+    },
+    useEvent(event: ReactListenerEvent): ReactListenerMap {
+      currentHookNameInDev = 'useEvent';
+      warnInvalidHookAccess();
+      updateHookTypesDev();
+      return updateEventListener(event);
     },
   };
 }
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -17,6 +17,12 @@ import type {ExpirationTime} from './ReactFiberExpirationTime';
 import type {HookEffectTag} from './ReactHookEffectTags';
 import type {SuspenseConfig} from './ReactFiberSuspenseConfig';
 import type {ReactPriorityLevel} from './SchedulerWithReactIntegration';
+import type {
+  ReactListenerEvent,
+  ReactListenerMap,
+  ReactListener,
+  Container,
+} from './ReactFiberHostConfig';
 
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 
@@ -54,6 +60,15 @@ import {
   runWithPriority,
   getCurrentPriorityLevel,
 } from './SchedulerWithReactIntegration';
+import {
+  registerListenerEvent,
+  attachListenerToInstance,
+  detachListenerFromInstance,
+  validateReactListenerMapListener,
+  validateReactListenerDeleteListener,
+} from './ReactFiberHostConfig';
+import {getRootHostContainer} from './ReactFiberHostContext';
+import {enableListenerAPI} from 'shared/ReactFeatureFlags';
 
 const {ReactCurrentDispatcher, ReactCurrentBatchConfig} = ReactSharedInternals;
 
@@ -97,6 +112,7 @@ export type Dispatcher = {|
   useTransition(
     config: SuspenseConfig | void | null,
   ): [(() => void) => void, boolean],
+  useEvent(event: ReactListenerEvent): ReactListenerMap,
 |};
 
 type Update<S, A> = {|
@@ -129,7 +145,8 @@ export type HookType =
   | 'useDebugValue'
   | 'useResponder'
   | 'useDeferredValue'
-  | 'useTransition';
+  | 'useTransition'
+  | 'useEvent';
 
 let didWarnAboutMismatchedHooksForComponent;
 if (__DEV__) {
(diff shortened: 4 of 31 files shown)


necolas:
> `null` or `undefined` can be used for the listener for cases when a listener is not desired.

Why?

> Using `stopImmediatePropgation` will not do anything special, and instead will behave like `stopPropgatation` with a warning, as we feel that such a behavior is generally undesirable.

Please can you elaborate on why you consider it undesirable.

> An additional method called `event.skipPropagationForKind` allows for event listeners of a given `kind` to be skipped.

I don't understand this feature. Please can you provide more details.

trueadm:
@necolas Hopefully these address your points:

> Why?

It's desirable that listeners can be optional, for cases where you might reference `props.onClick` where the prop might be `null` or `undefined`. Furthermore, it might also be desirable that you listent to an event but don't want to action upon it (for example, a mouse event given you know you can use a pointer event). This can fast-path the reconcilation process and skip events that have nullish listener callbacks.

> Please can you elaborate on why you consider it undesirable.

It's undesirable, although completely do-able because of how hooks work today. If you have many `useEvent` hooks on a function that listen to a given event, stopping propagation immediately will terminate them all, which means that many of the given effects in the same function cannot be fired. I'm on the fence with this once, as I'm happy to add it back but _do we really need it?_. I'm not sold on needing it, for the extra debugging complexity it adds when things go wrong (this includes the DOM today and the projects I've worked on in the past).

> I don't understand this feature. Please can you provide more details.

This is a power feature, but one I added because the user-land alternative would be 2-3x more code. The logical scenario you might imagine is local propgation we had with React Flare. You can assign listeners to the same `kind`, such as a given unique `Symbol`. All listeners of that given `Symbol` then obey `skipPropgationForKind` when passed that exact symbol in terms of propagation order (capture, then bubble, as two different phases). This would allow user-land implementations of inclusive propagation to occur without needing lots of wiring in React components. I'm not too bothered about reverting this particular feature if we feel it's too elusive to the ideal of this PR though.

packages/shared/ReactFeatureFlags.js
```diff
@@ -51,9 +51,12 @@ export const exposeConcurrentModeAPIs = __EXPERIMENTAL__;
 
 export const warnAboutShorthandPropertyCollision = false;
 
-// Experimental React Flare event system and event components support.
+// Experimental React Flare event system.
 export const enableDeprecatedFlareAPI = false;
 
+// Experimental Listener system.
+export const enableListenerAPI = true;
```
gaearon: This should be false? Or `__EXPERIMENTAL__`?

packages/shared/forks/ReactFeatureFlags.www.js
```diff
@@ -74,6 +74,8 @@ function updateFlagOutsideOfReactCallStack() {
 
 export const enableDeprecatedFlareAPI = true;
 
+export const enableListenerAPI = false;
```
gaearon: This should be true though?

necolas:
> it might also be desirable that you listen to an event but don't want to action upon it (for example, a mouse event given you know you can use a pointer event).

Why would you want to listen to an event but not do anything with it?

> This can fast-path the reconcilation process and skip events that have nullish listener callbacks.

Is calling functions that don't do anything a significant issue? Doing that seems to be inherent to hooks.

> I'm not sold on needing it, for the extra debugging complexity it adds when things go wrong (this includes the DOM today and the projects I've worked on in the past).

That's interesting because I think a lot of this is also true for `stopPropagation`. If you can stop an event propagating to other elements you might leave them in a broken state. For example, if I call `stopPropagation` on a `mouseup` but not a `mousedown`, an element higher up the tree might be stuck in a "press-down" state if the "end" event never reaches it.

> This is a power feature, but one I added because the user-land alternative would be 2-3x more code…I'm not too bothered about reverting this particular feature if we feel it's too elusive to the ideal of this PR though.

Yeah I think we could start with a small footprint and see if there's anything we need to add as we start using it. In particular, I think modifying the event from the native event sets a bad precedent. Maybe we'd even be better off doing things like `ReactDOM.someSpecialBehavior(event)`

trueadm:
I can revert the skip logic, that’s fine. Do you feel I should re-add stopImmediatePropagation too? In terms of the nullish listener fallback, it’s something React supports today, I.e you can provide null to a given listener - if you feel strongly I can make it an invariant to do this - in that you must always provide a valid function as a fallback.

necolas:
> Do you feel I should re-add stopImmediatePropagation too?

For me it's a question of "what do we want this API to be?" If it is to be a low-level API over the native events, then I start wondering:

* Why is the hook called `useEvent` and not `useListener`?
* Should we modify or remove fields from the native event?
* Why can listeners "bind" to the function component? If that's useful in practical terms (is it?), perhaps it's something people should have to explicitly choose to do. At the moment it is the default pattern.

trueadm:
@necolas I've made those changes to the PR:

- I've added `stopImmediatePropagation`
- I've removed `kind` and the skipping of propagation

To address your further points:

> Why is the hook called useEvent and not useListener?

I'm not sure it really matters, `useEvent` is kind of short for `useEventListener`, but then writing out `useDelegatedEventListener` is just terrible IMO. If we feel strongly that `useListener` is better than `useEvent`, then I'm down with that.

> Why should we modify or remove fields from the native event?

We only monkey patch the native events that relate to propagation now. We no longer add any additional fields other than the internal field we need for tracking.

> Why can listeners "bind" to the function component? If that's useful in practical terms (is it?), perhaps it's something people should have to explicitly choose to do. At the moment it is the default pattern.

I've gone back and forth on this one many times. I originally wanted to make `bind` mandatory, but then marking the function as a bind target is not as simple as it sounds – passing a reference to function is messy (what's stopping you from putting a reference to another closure?) as was marking it some kind of enum.

As to whether this is useful – this is maybe the topic you feel most controversial. Ultimately, I do feel that the default use case is actually more useful than you think after playing round with this API on real code, especially for things not needing targets (focus, keyboard) and for components that are small and self-contained, it makes for less code if you only want to add a few events. Given there's a clear way to bind to DOM elements, and this API is likely to be consumed via some higher-level hook (`usePress(targetRef)`) then I don't see it being problematic.

trueadm:
I've redesign and refactored the code in this PR after much delibration and feedback today. Notably: this design is far more low-level and more like how events are added with `addEventListener` today. This is intentional and with this design, it actually makes it possible to be approaches like the `createListener` propsoal and PR with this design – plus it reduces the complexity and overhead needed when thinking about some of the problems to do with adding/removing events.

Here's another example of this looks like now:

```jsx
function Component() {
  const divRef = useRef(null)
  // Creates a clickEvent that acts like a-kind of
  // Event Map. This should always happen in the
  // render phase, so we can properly setup the listeners
  // unconditionally. Furthermore, we can detect
  // this listeners during SSR, and eagerly setup
  // the right event listeners on init, to better
  // enhance event replaying (for custom events etc).
  const clickEvent = useEvent('click', options)

  useEffect(() => {
    const div = divRef.current;
    // Set a listener for a given DOM node
    clickEvent.listen(div, nativeEvent => {
      console.log(nativeEvent);
    });
    // You can listen to the `document` or
    // any other node managed by React. Nodes
    // not managed by React will result in
    // an error being fired.
    clickEvent.listen(document, nativeEvent => {
      console.log(nativeEvent);
    });
    // You can also remove a listener
    clickEvent.unlisten(div);
    
    // If you have many listeners, you can clear them
    // all without having to know of them
    clickEvent.clear();

    // Note: listeners will automatically be
    // destroyed if the component with the hook
    // is detatched, or the DOM node that the
    // listener is attached to is detatched.
  }, [clickEvent])

  return <div ref={divRef} />
}
```

necolas:
Nice one. This design makes a lot more sense to me and addresses my previous questions. Even the name `useEvent` now has appropriate meaning.

It's straightforward to see how existing user-space hooks could be converted to use an api like this. It's very similar to how you would natively attach event listens in react components with the benefit of upfront registration.

migueloller:
Is the name of the hook set in stone? `useEvent` is nice and terse but it's a tad confusing given that the naming of other hooks are "use [whatever is returned]" by the hook. Something like `useEventManager` is a bit more verbose but more in the direction of what I'm attempting to suggest. Maybe `useEventTarget` (although this might also be confusing since the returned value is not an instance of `EventTarget` and does not have `addEventListener`, et al.). The thing that's returned is basically managing `EventTargets` that are passed in via `setListener`. Thoughts?

eloytoro:
What are the benefits of using this over just attaching event listeners normally?

Also, I think it would be better to update the example to include listening to changes to onClick in the effect, right now it only registers the function passed when the component mounts but not when it rerenders

trueadm:
> Is the name of the hook set in stone?

The name isn't set in stone, although I feel it's still better than other suggestions at this point. The object it returns back is more of an `EventListenerMap` than a manager, as the event is still coupled to the type and options passed when defining the hook.

> What are the benefits of using this over just attaching event listeners normally?

There are many benefits. Namely:
- The event listeners automatically get recycled upon unmount
- The event listeners correctly batch updates and flush previous ones correctly, as well as co-ordinating priority levels with the internal scheduler
- The event listeners allow for alignment with the propagation of other events in the React event system, something not possible with manual DOM event listeners
- The event listeners will correctly work with React Portals, Suspense and Concurrent Mode, native event listeners do not.

> Also, I think it would be better to update the example to include listening to changes to onClick in the effect, right now it only registers the function passed when the component mounts but not when it rerenders

What example exactly? Was it the one with conditionally adding `mouseup` and `mousedown`? I'll update that now, my bad.

necolas:
It's not necessary to return refs either:

```js
function useClick(ref, callback) {
  const click = useEvent('click');
  useEffect(() => {
    click.setListener(ref.current, callback);
  }, [ref, callback]);
}
```

Most components would never need to use this low-level API as they could consume event- or gesture-specific hooks:

```js
import useClick from './useClick';

function Button({children, onClick}) {
  const ref = useRef(null);
  useClick(ref, onClick);
  return <button ref={ref}>{children}</button>;
}
```

This is in line with earlier approaches to high-level gesture recognizers that we prototyped:

```js
import useTap from './useTap';

function Tappable({children, onTap, onTapStart, onTapEnd }) {
  const [gestureState, listeners] = useTap({ onTapStart, onTap, onTapEnd });
  return <button listeners={listeners}>{children}</button>;
}
```

Which could instead take the following form:

```js
import useTap from './useTap';

function Tappable({children, onTap, onTapStart, onTapEnd }) {
  const ref = useRef(null);
  const gestureState = useTap(ref, { onTapStart, onTap, onTapEnd });
  return <button ref={ref}>{children}</button>;
}
```

migueloller:
@trueadm, great point on `useEffect`. It does not return an effect, but instead you're just "managing" an effect with the function passed in and the returned clean up function. So yeah, I can see `useEvent` growing on people just like `useEffect` did. Thanks for the prompt response!

migueloller:
@necolas, those are all great points and it also makes adding multiple events more ergonomic as one can use a single ref but pass it into multiple hooks!

bvaughn:
Drive by random comment from someone who hasn't had a chance to review the code as a whole: I assume you chose "set" and "delete" to be very visibly different from "add" and "remove" - but it seems like a weird pair. Seems like "add" (or maybe "record") and "delete", or "set" and "unset", go together more naturally.

MuYunyun:
> For propagation, the same rules of stopPropgation and stopImmediatePropgation apply to these event listeners.

stopPropgation => stopPropagation, stopImmediatePropgation => stopImmediatePropagation 
@trueadm :smile:

TheSavior:
Do you have any thoughts on how you expect this to be typed with a type system? I’d really like to be able to verify that, for example, people don’t pass in a callback that expects a ScrollEvent to a something that will actually pass it a ClickEvent

trueadm:
> Drive by random comment from someone who hasn't had a chance to review the code as a whole: I assume you chose "set" and "delete" to be very visibly different from "add" and "remove" - but it seems like a weird pair. Seems like "add" (or maybe "record") and "delete", or "set" and "unset", go together more naturally.

Well, I look at the return of `useEvent` as being a `EventListenersMap` of sorts, where if it were a normal `Map`, you'd `.set()` and `.delete()` properties, but in this case you're setting listeners to their respective DOM nodes. The main reason for not going with `add`, is because that would mean you can add many listeners to the same DOM node, which is not true here:

```jsx
const divRef= useRef(null);
const click = useClick('click');
const click2 = useClick('click');

useEffect(() => {
  const div = divRef.current;
  const A = () => {};
  const B = () => {};

  // Set the `click` ListenerMap for `div` to be `A`
  click.setListener(div, A);
  // Set the `click` ListenerMap for `div` to be `B`,
  // this will overwrite the previous `A` set to `div`
  click.setListener(div, B);
  // This is fine, this won't overwrite `B` as it's a
  // different ListenerMap, so now we are listening
  // to both `A` and `B` on `div`
  click2.setListener(div, A);
});

return <div ref={divRef} />
```

This is an intentional design too, as you'd almost never want to register multiple listeners to the same DOM node (for the same type of event).

In terms of naming, I'm happy changing the names. They used to be `listen()` and `unlisten()` but feedback was that didn't seem to Map like in term of what they were doing. `unset()` sounds like a good call too, I'd prefer that to `deleteListener()` so maybe `unsetListener()`?

> Do you have any thoughts on how you expect this to be typed with a type system? I’d really like to be able to verify that, for example, people don’t pass in a callback that expects a ScrollEvent to a something that will actually pass it a ClickEvent

It would have to be something we have to speak to the TypeScript and Flow teams about. It would have to be modelled off the `type` field you pass `useEvent` I guess. How is ther DOM's `addEventListener` typed today?

necolas:
>  I look at the return of useEvent as being a EventListenersMap of sorts, where if it were a normal Map, you'd .set() and .delete() properties

👍

> you'd almost never want to register multiple listeners to the same DOM node (for the same type of event)

I don't think this is quite true. You'd probably rarely want to do that from *within the same custom hook*. One of the drawbacks of working with event listeners in ReactDOM today is that it's difficult to compose listeners. However, with this design you can call `useEvent` in your own hook and it won't collide with one from another hook, so the composition works.

trueadm:
> I don't think this is quite true. You'd probably rarely want to do that from within the same custom hook. One of the drawbacks of working with event listeners in ReactDOM today is that it's difficult to compose listeners. However, with this design you can call useEvent in your own hook and it won't collide with one from another hook, so the composition works.

Agreed. I don't think I worded my originaly reply here well as what you wrote is essentially what I was trying to get at.

bvaughn:
> Well, I look at the return of `useEvent` as being a `EventListenersMap` of sorts, where if it were a normal `Map`, you'd `.set()` and `.delete()` properties

Ah, that's an interesting perspective. It makes sense!

TrySound:
Does setListener still accept null as listener? in this case we probably do not need unset or delete. it can be just setListener(node, null)

trueadm:
@TrySound It does not support `null` and will warn if you do this. That's a good point though, as it would save the amount of APIs needed and improve code size slightly. I wonder what others think about this too.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1221,6 +1238,143 @@ function updateTransition(
...
+        const listeners = Array.from(listenerMap.values());
+        for (let i = 0; i < listeners.length; i++) {
+          detachListenerFromInstance(listeners[i]);
+        }
+        listenerMap.clear();
+      }
+    };
+
+    const destroy = (instance: Container) => {
+      listenerMap.delete(instance);
```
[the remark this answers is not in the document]
trueadm: `beforeRemoveInstance` only gets called on DOM nodes React manages, so it won't ever get called on `document`. Thus we should never leak the `documentListenerSet` as when the function component that has the `useEvent` unmounts, we have an effect that will clean up all listeners, including document listeners. Given we only store the listeners on the DOM node, it should GC without leaking.

trueadm:
Thanks for the awersome feedback @bvaughn! Much appreciated. Let me think about how we can support `window` events too, as it's a bit more involved as we only delegate to the `document` in the `capture` phase currently, which might not work the same if we use `document`.

TrySound:
I also interested in how ResizeObserver and IntersectionObserver can be integrated here?

trueadm:
@TrySound They are definitely out of scope for the intents of this API.

TheSavior:
> It would have to be something we have to speak to the TypeScript and Flow teams about. It would have to be modelled off the type field you pass useEvent I guess.

I'm a bit nervous about the possibility of this current API being typed and would feel much better if those discussions with flow / typescript happened before this API becomes public and locked in. 

> How is ther DOM's addEventListener typed today?
My understanding is that `addEventListener` isn't typed well on DOM, because that's a limitation of that API, not because people don't want it to be. In React Native, all of our events are strongly typed and we wouldn't want to give that up. 

Maybe the answer is that `react-dom` (and `react-native`) should exports typed variants like `useClick()` that can be used instead of `useEvent('click')`, and `useEvent` can only be used for the events that don't have typed variants.

trueadm:
> Maybe the answer is that react-dom (and react-native) should exports typed variants like useClick() that can be used instead of useEvent('click'), and useEvent can only be used for the events that don't have typed variants.

That sounds like something that can easily be done in user-land too without us having to ship hundreds of these with ReactDOM etc, which will only increase bundle size too.

arianon:
the `useEvent` API could be typed with type declarations such as:

in TS:
```typescript
interface EventMap {
  "click": MouseEvent;
  "mouseup": MouseEvent;
  "focus": FocusEvent;
  "blur": FocusEvent;
  "keyup": KeyboardEvent;
  // ... and so on
}

function useEvent<K extends keyof EventMap>(type: K, options?: EventOptions): ReactDOMListenerMap<EventMap[K]>
```

and in Flow:
```
declare function useEvent(type: MouseEventTypes, options?: EventOptions): ReactDOMListenerMap<MouseEvent>;
declare function useEvent(type: KeyboardEventTypes, options?: EventOptions): ReactDOMListenerMap<KeyboardEvent>;
declare function useEvent(type: FocusEventTypes, options?: EventOptions): ReactDOMListenerMap<FocusEvent>;
```

bvaughn:
No prob Dominic!

> That sounds like something that can easily be done in user-land too without us having to ship hundreds of these with ReactDOM etc

Strong agree. Big fan of solving this via generics or union types.

packages/react-dom/src/events/DOMEventProperties.js
```diff
@@ -187,3 +187,17 @@ export function getEventPriorityForPluginSystem(
...
 }
+
+export function getEventPriorityForListenerSystem(type: string): EventPriority {
+  const priority = eventPriorities.get(((type: any): TopLevelType));
+  if (priority !== undefined) {
+    return priority;
+  }
+  if (__DEV__) {
+    console.warn(
+      'The event "type" provided to useEffect() does not have a known priority type.' +
```
necolas: s/useEffect/useEvent

trueadm:
I've also updated the PR to now accept either `window` or a managed DOM element when using `setListener()`. Will need to do some more testing, but it seems to be fine. :)

necolas:
I don't think this API is sufficient for creating a global gesture system like the responder system in React Native. In that case you might only want to set up one listener for each touch event type on the window to track touches across the entire surface, and to coordinate how the interaction lock is managed across multiple views (accounting for views not managed by react).

trueadm:
@necolas How would this API be more suffecient for that use-case? Would you mind explaining what is missing?

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
trueadm: I added a comment below as to why we can do this without leaking. We only call destroy on non document nodes, which means they should never have a retained reference to the `documentListenerSet`, so it hsould never leak.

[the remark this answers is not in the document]
bvaughn: Hm... doesn't that mean listeners might leak via the `window`?

So if I set a listener on `window` and then forgot to call ~~`deleteListener`~~ (I guess it's `setListener` -> `null` now?) or `clear` during my cleanup function, wouldn't that leak a lot of stuff?

Your example comment above seems at odds with this:
```js
    // Note: listeners will automatically be
    // destroyed if the component with the hook
    // is detatched, or the DOM node that the
    // listener is attached to is detatched.
```

Maybe I'm missing or misunderstanding something though?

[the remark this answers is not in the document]
trueadm: For window listeners, they get cleaned up when the function component unmounted, as the hook has an effect that cleans up all listeners on unmount.

necolas:
I don't think there's anything wrong with this hooks API or that it necessarily needs (or can be) adjusted to support "global" systems like [modality tracking](https://github.com/facebook/react/blob/master/packages/react-interactions/events/src/dom/Focus.js#L119-L134) and [interaction-locks](https://github.com/facebook/react/blob/ab1a4f249e61045d523ddbbfb840e868afbbf785/packages/legacy-events/ResponderEventPlugin.js). They rely on a single coordinator that listens only to "top" level events on the window/document.

packages/legacy-events/EventSystemFlags.js
```diff
@@ -11,8 +11,9 @@ export type EventSystemFlags = number;
...
 export const RESPONDER_EVENT_SYSTEM = 1 << 1;
-export const IS_PASSIVE = 1 << 2;
-export const IS_ACTIVE = 1 << 3;
-export const PASSIVE_NOT_SUPPORTED = 1 << 4;
-export const IS_REPLAYED = 1 << 5;
-export const IS_FIRST_ANCESTOR = 1 << 6;
+export const LISTENER_EVENT_SYSTEM = 1 << 2;
+export const IS_PASSIVE = 1 << 3;
+export const IS_ACTIVE = 1 << 4;
+export const PASSIVE_NOT_SUPPORTED = 1 << 5;
```
simbathesailor: @trueadm Just curious, why shift operator are used , instead of just number for constants.

trueadm: Because multiple flags can be combined together, which is only possible with bitwise operators (Aka flags).

devknoll:
Out of curiosity: is there a chance that setting listeners in `useEffect` (rather than `useLayoutEffect`) could cause some events to be missed (however rare), or is there something else done to avoid that?

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1345,6 +1348,41 @@ export function listenToEventResponderEventTypes(
...
+          // If we have an active event listener, do not register
+          // a passive event listener. We use the same active event
+          // listener.
+          return;
+        } else {
+          // If we have a passive event listener, remove the
+          // existing passive event listener before we add the
+          // active event listener.
+          const passiveListener = listenerMap.get(passiveKey);
+          if (passiveListener != null) {
```
konekoya: Shouldn't this be `passiveListener !== null`?

trueadm: Nope. If the key doesn’t exist then the passiveListener is always undefined. This could probably change from being a nullish check to being a void check.

trueadm:
@devknoll You're right, in some cases you'd want to use `useLayoutEffect`, which works more like how events happen today in React. This is important for `focus` and `load` etc, but really, the most common cases can move into `useEffect` which means we do less work in the commit phase! :)

PR #17653, opened 2019-12-18
DevTools build script enhancements

### LINKED ISSUE #17629: Update DevTools build process to use artifacts from CI

bvaughn:
Resolves #17629

* Adds a helper script to download the latest experimental build from CI.
* Updates DevTools build instructions to reference this new script as well as the dependency on the experimental release channel.

New command:
```sh
./scripts/release/download-experimental-build.js
```


bvaughn:
Sure

packages/react-devtools-extensions/README.md
```diff
@@ -7,15 +7,28 @@ The easiest way to install this extension is as a browser add-on:
...
+### Prerequisite steps
+DevTools depends on local versions of several NPM packages<sup>1</sup> also in this workspace. You'll need to either build or download those packages first.
+
+<sup>1</sup> Note that at this time, an _experimental_ build is required because DevTools depends on the `createRoot` API.
+
+#### Build from source
+To build dependencies from source, run the following command from the root of the repository:
 ```sh
-yarn build -- react,react-dom,react-is,scheduler --type=NODE
+RELEASE_CHANNEL=experimental \
```
threepointone: I keep making versions of this line for myself too, seems like a useful enough command to have its own alias in scripts? (Maybe not in this PR, and I’m happy to do so myself)

packages/react-devtools-extensions/README.md
```diff
@@ -7,15 +7,28 @@ The easiest way to install this extension is as a browser add-on:
...
+  yarn build \
+    -- react/index,react-dom,react-is,react-debug-tools,scheduler \
+    --type=NODE
 ```
-
-Once the above packages have been built, you can build the extension by running:
+#### Download from CI
+To use the latest build from CI, run the following command from the root of the repository:
```
threepointone: It’s not clear that this is an alternative to the previous step.

bvaughn: Why not? They're subheaders under the same header.

bvaughn: I could make the header size smaller maybe?

packages/react-devtools-extensions/README.md
bvaughn: Sure. I'll add a new alias to package for DevTools. Not sure that's exactly what you're suggesting, but it may be worth doing anyway.

packages/react-devtools-extensions/README.md
threepointone: Is there a problem with just having one of them? Like, downloading from CI could be the suggested option, and you have a postscript that mentions that’s you can also build from source if needed. 

I’m thinking from perspective of someone following instructions to build the extension, who doesn’t understand what these deps are, or what context they provide. If *I* had to quickly make a build and followed these instructions, I might have done both steps. Ofc this is just a nit, so feel free to ignore.

packages/react-devtools-extensions/README.md
threepointone: Lemme think about it some more? Just seems so common, but also tribal/hearsay knowledge.

packages/react-devtools-extensions/README.md
bvaughn: Yes. It's not sufficient.

*We* probably want to build from source, so we can test both DevTools *and* React changes.

External contributors (or Mozilla reviewers) probably want to download a CI build because it's faster and their purpose is not to test React.

packages/react-devtools-extensions/README.md
bvaughn: Okay. For now though I added a `yarn build-for-react` command.

PR #17614, opened 2019-12-15
[react-interactions] Event testing library improvements

necolas:
Introduces a state machine around pointer events to produce more accurate mock
touch events. This allows multi-touch unit tests to be written entirely in
terms of mock pointer interactions, while producing the expected
'changedTouches', 'targetTouches', and 'touches' fields for mock touch events.

PR #17660, opened 2019-12-19
Create packages/dom-event-testing-library

necolas:
This is a follow-up to #17614. It moves the unit testing library for events into the `packages` directory so it can more easily be used in tests for other react packages, and mirrored internally to help with testing of event hooks we prototype in www.


bvaughn:
This is nifty.

PR #17666, opened 2019-12-19
Null stateNode after unmount

bvaughn:
For discussion purposes.

This change avoids a certain case where we over-retain Fibers after unmount:

In `detachFiber` we null out `child` and `sibling` pointers in an effort to clear references to the stale subtree, but-
1. We don't clear the `stateNode` pointer for host element fibers.
1. The `stateNode` has references to its previous host element children.
1. Each host element (the children) have references to their corresponding Fibers via the `__reactInternalInstance`* prop.

So we end up retaining the whole Fiber subtree through the single `stateNode` reference.

Although the cost of retaining these references is limited (e.g. it will go away the next time the parent fiber renders) it can still obscure investigations into other more serious, user space leaks.

An example of the previous behavior can be seen here:
https://codesandbox.io/s/sweet-leakey-rj2ii

### DIFF
--- a/packages/legacy-events/ReactControlledComponent.js
+++ b/packages/legacy-events/ReactControlledComponent.js
@@ -33,8 +33,12 @@ function restoreStateOfTarget(target) {
     'setRestoreImplementation() needs to be called to handle a target for controlled ' +
       'events. This error is likely caused by a bug in React. Please file an issue.',
   );
-  const props = getFiberCurrentPropsFromNode(internalInstance.stateNode);
-  restoreImpl(internalInstance.stateNode, internalInstance.type, props);
+  const stateNode = internalInstance.stateNode;
+  // Guard against Fiber being unmounted.
+  if (stateNode) {
+    const props = getFiberCurrentPropsFromNode(stateNode);
+    restoreImpl(internalInstance.stateNode, internalInstance.type, props);
+  }
 }
 
 export function setRestoreImplementation(
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -905,6 +905,7 @@ function detachFiber(current: Fiber) {
   current.lastEffect = null;
   current.pendingProps = null;
   current.memoizedProps = null;
+  current.stateNode = null;
   if (alternate !== null) {
     detachFiber(alternate);
   }


bvaughn:
> Might be worth checking on www, as there might be another call-site that needs a guard. :)

I kicked off an early sync with this change to see if any FB tests failed, and they did not.

bvaughn:
More context: this behavior is specific to class components.

The root of this leak is actually the `_reactInternalFiber` attribute, which is part of the `ReactInstanceMap`. This reference is needed in order for class component instances to schedule updates, but maybe another part of this PR should be to also clear this instance?

Although `ReactInstanceMap` exports a `remove` function, we never actually use it anywhere.

gaearon:
>The root of this leak is actually the _reactInternalFiber attribute

I think this is a bit of a misunderstanding. In this example, `_reactInternalFiber` points to the `App` Fiber. We can't *remove* that reference because `App` itself is not unmounted. So would be too early to remove it.

We're seeing this problem because `_reactInternalFiber` points to the `App` *alternate*. And the alternate holds onto the deleted children.

What if `_reactInternalFiber` pointed to the *current* `App` (which has `child = null`)? Would that help? Alas, no. It wouldn't help because **then we’d still see that Fiber as `_reactInternalFiber.alternate` instead**.

So the root problem here is really the architectural choice of using alternates. The tradeoff this choice is making is that *when children are deleted, the parent’s alternate will keep holding onto their Fibers until the next update* (when it would get overwritten).

We've been gradually cutting down on the consequences of this choice by clearing more and more fields on those child Fibers. So that retaining them isn’t as big of a deal.

But the fundamental problem isn’t that we have some leak we could easily remove. It’s the architectural choice of using alternates. It can be mitigated by... not using them. But that’s a big refactor.

bvaughn:
> This might still be good to get in regardless. If we’re clearing so many fields already, might as well do this one. This will reduce the effect of the leaks

This is my feeling after look at, and thinking about, this a bit more. This would still be a nice change to get in because it would minimize the impact of the current architecture and improve the signal/noise ratio for other investigations.

bvaughn:
I'm going to move forward with this change because I think it would help when investigating a reported potential DevTools leak, and because it does not seem to negatively impact any tests in this repo or in Facebook. Let's chat more about any other follow up items or concerns after the holidays!

kelly-tock:
anyone seeing this in react 17 as well?

eltonchan:
I can still encounter this problem in React 17 version. This does not necessarily happen, it is very random. Is this problem only with class components? Is it normal if use function component? @bvaughn

shaikh-kamran:
I am still having this issue in React 17, how to avoid this?

joankrastanov:
I am encountering this issue in React 17, does anyone know if this is fixed in future versions?

knownasilya:
Upgrading from 17 to 18 resolves my detached nodes issue, when doing performance analysis nodes kept growing, even though I was doing RR navigations. Updating to 18 I saw a drop in nodes 

v17
Screenshot 2024-12-24 at 11 17 08 AM

v18
Screenshot 2024-12-24 at 11 13 33 AM

PR #17668, opened 2019-12-19
Disable unsupported context menu options for Firefox

bvaughn:
Unfortunately the "copy to clipboard" and "go to definition" context menu options do not work for the Firefox add-on. I'm not sure why "go to definition" doesn't work. (Seems like it should but it just fails silently.) The clipboard option seems more obviously broken though.

We currently use [`document.execCommand("copy")`](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) which works for Chrome but fails for Firefox with the following error:
image

Unfortunately, using the newer [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write) fails for both browsers:
image

So for now it seems the best we can do is stick with the old `execCommand` API and disable the menu option for Firefox.

### Chrome demo
Context menu in Chrome

### Firefox demo
Context menu in Firefox

### DIFF
--- a/packages/react-devtools-extensions/src/main.js
+++ b/packages/react-devtools-extensions/src/main.js
@@ -211,12 +211,15 @@ function createPanelIfReactLoaded() {
               browserTheme: getBrowserTheme(),
               componentsPortalContainer,
               enabledInspectedElementContextMenu: true,
+              enabledInspectedElementContextMenuCopy: isChrome,
               overrideTab,
               profilerPortalContainer,
               showTabBar: false,
               store,
               warnIfUnsupportedVersionDetected: true,
-              viewAttributeSourceFunction,
+              viewAttributeSourceFunction: isChrome
+                ? viewAttributeSourceFunction
+                : null,
               viewElementSourceFunction,
             }),
           );
--- a/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
@@ -303,6 +303,7 @@ function InspectedElementView({
 
   const {
     isEnabledForInspectedElement,
+    supportsCopyOperation,
     viewAttributeSourceFunction,
   } = useContext(ContextMenuContext);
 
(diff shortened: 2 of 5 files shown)


bvaughn:
Tagging a few random reviewers who (I think) aren't yet on PTO 😄

digitarald:
> (Seems like it should but it just fails silently.) The clipboard option seems more obviously broken though.

`inspect` should work (per @jasonLaster). Maybe we can have a follow up to investigate.

bvaughn:
@digitarald `inspect` works when run manually in the console, but it does not seem to work when run via an extension.

jasonLaster:
@bvaughn that's disappointing, but maybe it shouldn't be too surprising as I did not test it. My hope was that console commands, would just work in the extension.

CC @nchevobbe @rpl

bvaughn:
@jasonLaster No worries! Here's a pointer to the code:
https://github.com/facebook/react/blob/1b9328cd9b516fb3d1bd125a6cbd1e9e8333ac5c/packages/react-devtools-extensions/src/main.js#L154-L171

The way this works currently is the frontend sends a `postMessage` command saying to inspect an attribute at a specific path (e.g. some-fiber-id -> props -> foo -> bar). The "renderer" (which is injected to run in the page in the same context as React itself) stores the value at the specified path in a global, `window.$attribute`. Then the extension itself asks the browser to `inspect()` it.

The extension is involved because JavaScript running in the document itself can't call `inspect()`.
> ReferenceError: inspect is not defined

In Firefox though, it seems to just silently fail when the extension calls `inspect()`.

rpl:
In firefox the interaction with the clipboard from an extension requires additional permissions ("clipboardRead" and/or "clipboardWrite"), some details about that are available in the following MDN doc page:
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard

In particular this section may be helpful: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard#Browser-specific_considerations

(and in case it may be also helpful, here are the related test cases from mozilla-central: https://searchfox.org/mozilla-central/source/toolkit/components/extensions/test/mochitest/test_ext_clipboard.html)

I just gave a quick look and (at least if I'm reading it correctly) currently `document.execCommand("copy")` seems to be executed in the webpage, but to leverage the extension permissions it should be executed from the extension content script attached to that page. 

About the issue with `inspect`, the current behavior on Firefox is the following:
- If the parameter passed to `inspect` is a DOM node, the developer toolbox should be switched to the  "Inspector" panel and the given element should be selected
- If the parameter is a javascript object, the given object should be shown in the webconsole sidebar panel (which should also be opened automatically)

`devtools.inspectedWindow.eval("inspect(...)")` seems to also be used by the extension to select the DOM element related to the currently selected React component instance, and that seems to be working as expected (I just tried to use that by building the react devtools extension from this repo)

And so I'm wondering what is set on the `$attr` property? is it a plain js object? which is the behavior expected by the React DevTools extension?

@jasonLaster @bvaughn would you mind to give me some steps to be able to reproduce this issue locally?
I gave a quick look and I was immediately able to trigger the issue with the clipboard but I didn't found the context menu item to trigger the issue with `inspect` (and I'm wondering if it has been already disabled on Firefox or I was just not looking to the right context menu).

bvaughn:
@rpl I tried adding the additional write permission but it didn't make a difference. I mentioned this in the Firefox Slack but I guess I forgot to add it to the PR. My bad.

bvaughn:
I'm going to merge this for now so the feature doesn't block an extension release to the browser stores. I'll be happy to follow up and re-add support for Firefox in a future PR if we figure out a path forward that satisfies the above constraints.

jasonLaster:
I filed #17681 as a follow up.

Issue #17229, opened 2019-10-30
onMouseEnter is fired on disabled buttons


diegohaz:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

`onMouseEnter` is fired on `disabled` buttons.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

https://codesandbox.io/s/react-onmouseenter-on-disabled-buttons-fskwd

**What is the expected behavior?**

`onMouseEnter` shouldn't be triggered on `disabled` buttons.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

React 16.11.0 (not sure when it started to happen)
Chrome/MacOS

edwincoronado:
Taking a look at this to see if I can fix it.

abrahamwilliam:
@edwincoronado @diegohaz 

I have identified the fix to this issue/Bug, just testing it.I am proceeding to it, please let me know if any issues.

abrahamwilliam:
@edwincoronado @diegohaz 

I have identified the fix to this issue/Bug, just testing it.I am proceeding to it, please let me know if any issues.

AlfredoGJ:
I submitted a pull request that solves this problem. If someone wants to take a look. Feedback is welcome.

gaearon:
Should be fixed in 16.13.
https://reactjs.org/blog/2020/03/02/react-v16.13.0.html

Issue #4251, opened 2015-06-29
mouseenter fires on disabled inputs whereas mouseleave does not


jquense:
There is an asymmetry to EnterLeave event plugin. Since `mouseenter` is created from the relativeTarget of the `mouseout` event it fires even though the target is disabled. Since the `mouseleave` is the inverse, i.e requires that the disabled element fire a mouseout, it doesn't fire a `mouseleave` for the disabled element.

I am pretty sure the correct behavior here is that neither event should fire if its target is disabled, since this mirrors `mouseout`. No idea if none-chrome browsers have the same behavior for which mouse events fire on disabled elements.

Additional caveat I just realized, React is probably also not firing `mousenter` events in the case where the mouse leaves a disabled element into a non disabled element

jquense:
So here is an initial attempt at a fix but I can't figure out how to properly use `EventPropagators` here. Is there a way I am missing to listen for child events (i.e mouseout/over) but also return an event that doesn't bubble itself?

```js
var EventConstants = require("./EventConstants");
var EventPropagators = require("./EventPropagators");
var SyntheticMouseEvent = require("./SyntheticMouseEvent");
var containsNode = require("./containsNode");

var ReactMount = require("./ReactMount");
var keyOf = require("./keyOf");

var topLevelTypes = EventConstants.topLevelTypes;
var getFirstReactDOM = ReactMount.getFirstReactDOM;

var eventTypes = {
  mouseEnter: {

    registrationName: keyOf({ onMouseEnter: null }),
    dependencies: [topLevelTypes.topMouseOut, topLevelTypes.topMouseOver]
  },
  mouseLeave: {
    registrationName: keyOf({ onMouseLeave: null }),
    dependencies: [topLevelTypes.topMouseOut, topLevelTypes.topMouseOver]
  }
};

var extractedEvents = [null, null];

var EnterLeaveEventPlugin = {

  eventTypes: eventTypes,

  /**
   * For almost every interaction we care about, there will be both a top-level
   * `mouseover` and `mouseout` event that occurs. Only use `mouseout` so that
   * we do not extract duplicate events. However, moving the mouse into the
   * browser from outside will not fire a `mouseout` event. In this case, we use
   * the `mouseover` top-level event.
   *
   * @param {string} topLevelType Record from `EventConstants`.
   * @param {DOMEventTarget} topLevelTarget The listening component root node.
   * @param {string} topLevelTargetID ID of `topLevelTarget`.
[...]

PR #17675, opened 2019-12-20
Fix onMouseEnter is fired on disabled buttons

AlfredoGJ:
This pull request fixes issue #17229. I modified `shouldPreventMouseEvent` function to include `onMouseEnter` and added a new test to `ReactBrowserEventEmmiter-test`.

### DIFF
--- a/packages/legacy-events/EventPluginHub.js
+++ b/packages/legacy-events/EventPluginHub.js
@@ -45,6 +45,7 @@ function shouldPreventMouseEvent(name, type, props) {
     case 'onMouseMoveCapture':
     case 'onMouseUp':
     case 'onMouseUpCapture':
+    case 'onMouseEnter':
       return !!(props.disabled && isInteractive(type));
     default:
       return false;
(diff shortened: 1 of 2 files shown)


AlfredoGJ:
@benbot Unlike `onMouseEnter`, `onMouseExit` performs well and has not the same bug. `onMouseExit` event is not even generated for a disabled button. I'm thinking that maybe none of these event names should be in that switch statement and the bug is in the event registration phase.

 I 'll be doing deeper research these days. Meanwhile, this could a useful workaround for someone.

mohsinulhaq:
This fixes https://github.com/facebook/react/issues/4251 as well, right?

LordSeb:
What if you want to display tooltip / help text why user cannot perform action connected with disabled button when user hovers the button? (my real world case). The fix should be to make onMouseExit enabled and have both enabled and not disable the onMouseEnter. If you know nice way (not hack like fake-disable button) how to solve my example case without onMouseEnter onMouseEnter I am free to discuss. But otherwise it should be both enabled in my opinion.

gaearon:
Can you please file a new issue to discuss?

ulvido:
I was using onMouseEnter on a disabled button. If user enters add button and the input is empty, I change the input placeholder to "write something first" and it was fine. what should i do now?

PR #17678, opened 2019-12-20
[react-dom] Refactor event priority handling to its own module

trueadm:
The team have mentioned a bunch of times that it felt wrong that DOM Event Priority handling was all coupled with the SimpleEventPlugin, especially as might want to extend on the priorities with newer features and refactors in the future.

This PR breaks out the priority logic into a dedicated module and makes uses of Maps for faster lookups (I benchmarked before and after and Maps were consistently faster than object property lookups in Chrome and Safari, neutral in Firefox). We should also get a slight code size saving by some of the code golfing that I did whilst refactoring the code.

### DIFF
--- a/packages/react-dom/src/events/DOMEventProperties.js
+++ b/packages/react-dom/src/events/DOMEventProperties.js
@@ -1,0 +1,189 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import type {EventPriority} from 'shared/ReactTypes';
+import type {
+  TopLevelType,
+  DOMTopLevelEventType,
+} from 'legacy-events/TopLevelEventTypes';
+import type {DispatchConfig} from 'legacy-events/ReactSyntheticEventType';
+
+import * as DOMTopLevelEventTypes from './DOMTopLevelEventTypes';
+import {
+  DiscreteEvent,
+  UserBlockingEvent,
+  ContinuousEvent,
+} from 'shared/ReactTypes';
+
+// Needed for SimpleEventPlugin, rather than
+// do it in two places, which duplicates logic
+// and increases the bundle size, we do it all
+// here once. If we remove or refactor the
+// SimpleEventPlugin, we should also remove or
+// update the below line.
+export const simpleEventPluginEventTypes = {};
+
+export const topLevelEventsToDispatchConfig: Map<
+  TopLevelType,
+  DispatchConfig,
+> = new Map();
+
+const eventPriorities = new Map();
+
+// We store all the DOMTopLevelEventTypes and their React Types in pairs of two.
+// Furthermore, we ignore prettier so we can keep the formatting sane.
+
+// prettier-ignore
+const discreteEvents = [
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/events/DOMEventPriority.js
```diff
@@ -0,0 +1,117 @@
...
+ * @flow
+ */
+
+import type {EventPriority} from 'shared/ReactTypes';
+import type {TopLevelType} from 'legacy-events/TopLevelEventTypes';
+
+import * as DOMTopLevelEventTypes from './DOMTopLevelEventTypes';
+import {DiscreteEvent, UserBlockingEvent, ContinuousEvent} from 'shared/ReactTypes';
+
+// We use a Set here rather than an Array or Object map.
```
philipp-spiess: Out of curiosity: Have you also tried a switch statement?

trueadm: I did and it was even faster, but not by much. The issue with switches was that it was also quite a bit more code size comared to before.

trueadm:
I did some benchmarking on FF and Safari and it looks like using multiple Sets is not as optimal as Chrome's V8. I've changed my approach to use Maps and moved things around so we don't do multiple lookups – this is now consistently faster in Chrome and Safari, neutral in Firefox.

packages/react-dom/src/events/DOMEventProperties.js
```diff
@@ -0,0 +1,175 @@
...
+} from 'legacy-events/TopLevelEventTypes';
+import type {DispatchConfig} from 'legacy-events/ReactSyntheticEventType';
+
+import * as DOMTopLevelEventTypes from './DOMTopLevelEventTypes';
+import {
+  DiscreteEvent,
+  UserBlockingEvent,
+  ContinuousEvent,
+} from 'shared/ReactTypes';
+import {simpleEventPluiginEventTypes} from './SimpleEventPlugin';
```
necolas: Typo: simpleEvent**Pluigin**EventTypes

But this file now imports the `SimpleEventPlugin` and creates a cyclic dependency between that and `DOMEventProperties`

packages/react-dom/src/events/DOMEventProperties.js
```diff
@@ -0,0 +1,184 @@
...
+ *   [TOP_ABORT, { sameConfig }],
+ * ]);
+ */
+
+function processTopEventTypesByPriority(
+  eventTypes: Array<DOMTopLevelEventType | string>,
+  priority: EventPriority,
+): void {
+  // As the event types are in pairs of two, we need to iterate
+  // through in twos.
```
philipp-spiess: 😮 neat!

necolas: eek though

packages/react-dom/src/events/DOMEventProperties.js
```diff
@@ -0,0 +1,189 @@
...
+  ContinuousEvent,
+} from 'shared/ReactTypes';
+
+// Needed for SimpleEventPlugin, rather than
+// do it in two places, which duplicates logic
+// and increases the bundle size, we do it all
+// here once. If we remove or refactor the
+// SimpleEventPlugin, we should also remove or
+// update the below line.
+export const simpleEventPluiginEventTypes = {};
```
necolas: Typo here, should be "Plugin"

PR #17690, opened 2019-12-22
[DevTools] Display RegExp values in props/state

bvaughn:
Replaces PR #17023

# Before
Screenshot of props panel before RegExp change

# After
Screenshot of props panel after RegExp change


bvaughn:
This change is pretty rote so I'm going to merge.

PR #17700, opened 2019-12-24
[react-test-renderer] Fix `findByType` broken for SimpleMemoComponent

### LINKED ISSUE #17301: react-test-renderer: the findByType method doesn't work with memo components
Feature

Whenever I try something like:

```
ReactTestRenderer.create(<SomeComponent />).root.findByType(SomeMemoComponent);
```

I get the following error: `No instances found with node type: "undefined"`. The only way I found for this to work was to reference the `type` property of memo components like this:

```
ReactTestRenderer.create(<SomeComponent />).root.findByType(SomeMemoComponent.type);
```

I am fine with this solution but then flow complains that `type` doesn't exist so I find myself fixing this with `$FlowFixMe` all over the place.

I would expect that passing a memo component to `findByType` would work. Or that flow would recognize the `type` property of memo components. I think both should work, specially the first option.

No, it never worked as far as I know.

kevin940726:
Fix #17301 .

The fix is simply check if the `tag` of the fiber node is `SimpleMemoComponent`, and check the `elementType` instead of `type`.

I'm not fully understand the difference between `elementType` and `type` in test-renderer though, looks like if we changed to always get the `elementType` instead of `type` would also work? Maybe it has something to do with `Lazy` and `Suspense`, but AFAIK `react-test-renderer` currently doesn't support them.

### DIFF
--- a/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
+++ b/packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
@@ -321,8 +321,7 @@ describe('ReactHooksInspectionIntegration', () => {
     }
     let Foo = React.memo(InnerFoo);
     let renderer = ReactTestRenderer.create(<Foo />);
-    // TODO: Test renderer findByType is broken for memo. Have to search for the inner.
-    let childFiber = renderer.root.findByType(InnerFoo)._currentFiber();
+    let childFiber = renderer.root.findByType(Foo)._currentFiber();
     let tree = ReactDebugTools.inspectHooksOfFiber(childFiber);
     expect(tree).toEqual([
       {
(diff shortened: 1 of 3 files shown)


gaearon:
This seems like it would break code searching by the *inner* type, right? I think we need a more fleshed out proposal for how you imagine this working ideally. You probably don't want different behavior for `SimpleMemo` and `Memo`. Overall, I think issues like this show that this whole API is flawed and not something we'd like to keep supporting in longer term.

bensampaio:
@gaearon what would be a better solution then?

Would it make sense to change the Flow types of React so that you can write `findByType(MyComponent.type)` without getting type errors? Or would you prefer that `findByType` accepts `memo` components as well like proposed in this MR?

PR #17703, opened 2019-12-24
chore: Refactor DangerJS

gabrielferreiraa:
- Removing unnecessary `else`
- Create aux function to check experimental release channel
- Import space


gaearon:
Thanks for PR, but I don't see why this is an improvement. This seems very subjective.

PR #17746, opened 2019-12-30
Build and store DevTools artifacts in Circle CI

bvaughn:
Might as well leverage Circle CI to test our DevTools builds, and if we're doing that- we might as well store the artifacts as well. Maybe in the future, we should release DevTools directly from CI artifacts as well, like we do for NPM packages.

Let's see if these changes actually work... 😄

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -193,6 +193,22 @@ jobs:
             - dist
             - sizes/*.json
 
+  build_devtools_and_process_artifacts:
+    docker: *docker
+    environment: *environment
+    parallelism: 20
+    steps:
+      - checkout
+      - attach_workspace: *attach_workspace
+      - *restore_yarn_cache
+      - *run_yarn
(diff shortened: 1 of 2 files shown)


bvaughn:
I think Circle CI is confused. Looks like it re-ran part of the build commands without cleaning up, so now some things are failing b'c files are left over from previous runs. I used to have the option to clear cache and re-run but I don't see it anymore.

bvaughn:
Yay Circle CI seems to have finally passed after the last round of changes.

xusai2014:
> Might as well leverage Circle CI to test our DevTools builds, and if we're doing that- we might as well store the artifacts as well. Maybe in the future, we should release DevTools directly from CI artifacts as well, like we do for NPM packages.
> 
> Let's see if these changes actually work... 😄

Hi Guys

Your ideas appeal to me. Could you describe the detail of the ideas？

PR #17757, opened 2020-01-02
Added missing guard to DevTools for Objects with null proto

### LINKED ISSUE #17754: DevTools can't inspect object without prototype

Bug

1. Select a component
2. The right panel (property panel?) always shows `Loading...` 
3. The following error is printed to the console.

```
backend.js:1 Uncaught TypeError: Cannot read property 'name' of undefined
    at O (backend.js:1)
    at s (backend.js:1)
    at s (backend.js:1)
    at l (backend.js:1)
    at Object.inspectElement (backend.js:6)
    at t.<anonymous> (backend.js:6)
    at t.r.emit (backend.js:6)
    at backend.js:32
    at t (backend.js:8)
```

image

I don't know why you minify the output, but the call stack points to here:

https://github.com/facebook/react/blob/f887d1aa27336baa0bc292158793a5a244c712b6/packages/react-devtools-shared/src/utils.js#L391

https://codesandbox.io/s/falling-wave-1j0qe

React DevTools displays the props of the selected component.

React 16.12.0
Chrome 79.0.3945.88
IDK

bvaughn:
This prevents a runtime error from occurring when these objects are inspected.

Resolves #17754

### DIFF
--- a/packages/react-devtools-shared/src/utils.js
+++ b/packages/react-devtools-shared/src/utils.js
@@ -388,7 +388,7 @@ export function getDataType(data: Object): DataType {
         return data.constructor.hasOwnProperty('BYTES_PER_ELEMENT')
           ? 'typed_array'
           : 'data_view';
-      } else if (data.constructor.name === 'ArrayBuffer') {
+      } else if (data.constructor && data.constructor.name === 'ArrayBuffer') {
         // HACK This ArrayBuffer check is gross; is there a better way?
         // We could try to create a new DataView with the value.
         // If it doesn't error, we know it's an ArrayBuffer,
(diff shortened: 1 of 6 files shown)


packages/react-devtools-shared/src/utils.js
```diff
@@ -388,15 +388,15 @@ export function getDataType(data: Object): DataType {
         return data.constructor.hasOwnProperty('BYTES_PER_ELEMENT')
           ? 'typed_array'
           : 'data_view';
-      } else if (data.constructor.name === 'ArrayBuffer') {
+      } else if (data.constructor && data.constructor.name === 'ArrayBuffer') {
         // HACK This ArrayBuffer check is gross; is there a better way?
```
james4388, line 392: Sorry but can we use `data instanceof ArrayBuffer` here?

PR #17758, opened 2020-01-02
Fixed CircleCI DevTools build artifact tar internal structure

bvaughn:
Oops. Previously the `tar.gz` file had nested folders instead of it, which made it awkward to work with.


threepointone:
I'm just unsure how this command makes a different internal structure in the compressed file. Feel free to ignore if it's not important.

bvaughn:
Before, unzipping the .tar.gz file would unpack to `./build/devtools/<stuff>`. Now it will unpack to `./<stuff>`.

Issue #16446, opened 2019-08-19
Error: "Could not find node with id "266" in commit tree"


ldcc-juj:
PLEASE INCLUDE REPRO INSTRUCTIONS AND EXAMPLE CODE

I got this error when I click 'Ranked'.

---------------------------------------------
Please do not remove the text below this line
---------------------------------------------

DevTools version: 4.0.4-3c6a219

Call stack: at chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:11:11441
    at Map.forEach (<anonymous>)
    at commitIndex (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:11:11387)
    at e.getRankedChartData (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:11:11920)
    at _i (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:56:277123)
    at Ha (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:43:55890)
    at Xl (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:43:98280)
    at Hl (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:43:84255)
    at Fl (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:43:81285)
    at chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:43:25363

Component stack: in _i
    in div
    in div
    in div
    in Or
    in Unknown
    in n
    in Unknown
    in div
    in div
    in Ha
    in le
    in ve
    in ko
    in Ul

bvaughn:
More detailed repro steps would be appreciated. What did you do up util you clicked "Ranked"?

Apollinaire:
Same here. I got this problem today on this profiling: https://gist.github.com/Apollinaire/effeb23efae1d603e9e0b6a8e39132e5
First time i had the error it said id 273, then I refreshed the page, the profiling was still present so I downloaded it to see if just the profiling reproduces the error. It does, but now it says id 257. 
Reproduction: I just opened the profiler, recorded, fiddled with my app, stopped the record, then looked at the commits. Everything was fine, until I clicked on the Ranked tab. I instantly saw the error message that got me here.
Version: `React Developer Tools
4.2.1 (11/27/2019)
[...]

PR #17267, opened 2019-11-04
Bailout without entering work loop for roots without work WIP

bvaughn:
Wraps up #16980

This addresses some edge cases where React currently does a no-op render and an empty/unnecessary commit.

- [x] Bailout without entering work loop for roots without work (see #16980).
- [x] Fix failing refs-test by wrapping updates with `act()`.
- [x] See if any other tests trigger this early bailout behavior to better understand the causes.

The tests listed below trigger this new early bailout code. Everyone is in `performSyncWorkOnRoot` (none trigger the new code in `performConcurrentWorkOnRoot`). I spot checked a couple of them to see why the new code is being hit, to see if it looked problematic. Below is my findings:

* `ReactES6Class-test` "_renders only once when setting state in componentWillMount_"
This one calls `performSyncWorkOnRoot()` twice. The first time is the callback passed from `legacyRenderSubtreeIntoContainer` to `unbatchedUpdates`. The second one (the one that bails out) is when `unbatchedUpdates` calls `flushSyncCallbackQueue`. This flush can be bailed out on.

* `RaectDOMInput-test` "_should control a value in reentrant events_"
This one hits the new codepath when it dispatch a discrete "input" event. It looks like when our test calls `node.dispatchEvent()` for the "input" event, something is actually dispatching a series of events (input, input, blur, focus) which causes more updates to be scheduled with React than necessary. Now we bail out after the first.

* `ReactCompositeComponent` "_should warn about `setState` in render_"
This one calls `setState` in render. Without this call, the bailout codepath doesn't get hit. Looks like the `setState` call leaves two things in the queue, so when the subsequent call to `flushSyncCallbackQueue` flushes them both, the second one is a no-op. The first thing gets added to the queue when the `setState` call is made. The second one by `commitRootImpl()` when it calls `ensureRootIsScheduled()` because `getNextRootExpirationTimeToWorkOn` returns a value that indicates there's more work.

---

* `ReactTestUtils.act()` > legacy mode › sync › flushes effects on every call
* `ReactTestUtils.act()` > blocking mode › sync › flushes effects on every call
* `ReactDOMInput` > should control a value in reentrant events
* `ReactDOMInput` > should control values in reentrant events with different targets
* `ReactDOMInput` > switching text inputs between numeric and string numbers › changes the number 2 to "2.0" using a change handler
* `ReactDOMInput` > should control radio buttons if the tree updates during render
* `ReactDOMInput` > assigning the value attribute on controlled inputs › always sets the attribute when values change on text inputs
[...]

PR #17759, opened 2020-01-02
Partial fix DevTools Profiler "Could not find node…" error

bvaughn:
DevTools records the following information while profiling:
1. A snapshot of the React tree when profiling started.
1. The [operations array](https://github.com/facebook/react/blob/master/packages/react-devtools/OVERVIEW.md#serializing-the-tree) for each commit.
1. Profiling metadata (e.g. durations, which values changed, etc.) for each commit.

It uses this information (snapshot + operations diff) to reconstruct the state of the application for a given commit as it's viewed in the Profiler UI. Because of this, it's very important that the operations and metadata arrays align. If they don't align, the profiler will be unable to correctly reconstruct the tree for a given commit, and it will likely throw errors (like "Could not find node…")

#16446 tracks a long-standing bug where the profiler is unable to correctly regenerate the tree for a commit. I believe it is because the two arrays I mentioned above are misaligned. I am still not entirely sure what causes the original bug, but I made it worse with PR #17253 by introducing another potential way for it to happen. This PR addresses that regression at least (and adds test coverage for it).

I will follow up this afternoon on the original #16446 issue. I think I may have [a lead](https://github.com/facebook/react/issues/16446#issuecomment-570289883) on what's happening at least, if not exactly an idea of how to reproduce it. I believe this fix is worth landing on its own though, since it seems to have made the bug easier to trigger.

### DIFF
--- a/packages/react-devtools-shared/src/__tests__/profilerStore-test.js
+++ b/packages/react-devtools-shared/src/__tests__/profilerStore-test.js
@@ -118,5 +118,6 @@ describe('ProfilerStore', () => {
     const root = store.roots[0];
     const data = store.profilerStore.getDataForRoot(root);
     expect(data.commitData).toHaveLength(1);
+    expect(data.operations).toHaveLength(1);
   });
 });
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -1021,10 +1021,30 @@ export function attach(
       pendingSimulatedUnmountedIDs.length === 0 &&
       pendingUnmountedRootID === null
     ) {
-      // If we're currently profiling, send an "operations" method even if there are no mutations to the tree.
-      // The frontend needs this no-op info to know how to reconstruct the tree for each commit,
-      // even if a particular commit didn't change the shape of the tree.
+      // If we aren't profiling, we can just bail out here.
+      // No use sending an empty update over the bridge.
       if (!isProfiling) {
+        return;
+      }
+
+      const current = root.current;
+      const alternate = current.alternate;
+
+      // Certain types of updates bail out at the root without doing any actual render work.
+      // React should probably not call the DevTools commit hook in this case,
+      // but if it does- we can detect it and filter them out from the profiler.
+      // NOTE: Keep this logic in sync with the one in handleCommitFiberRoot()
+      const didBailoutAtRoot =
+        alternate !== null &&
+        alternate.expirationTime === 0 &&
+        alternate.childExpirationTime === 0;
+
+      // The Profiler stores metadata for each commit and reconstructs the app tree per commit using:
+      // (1) an initial tree snapshot and
+      // (2) the operations array for each commit
+      // Because of this, it's important that the operations and metadata arrays align,
+      // So the logic that skips metadata for bailout commits should also apply to filter operations.
+      if (didBailoutAtRoot) {
         return;
       }
     }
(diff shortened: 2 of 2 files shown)


bvaughn:
Tagged a few people as reviewers because (probably) no one in particular has a ton of context to review this.

packages/react-devtools-shared/src/__tests__/profilerStore-test.js
```diff
@@ -118,5 +118,6 @@ describe('ProfilerStore', () => {
     const root = store.roots[0];
     const data = store.profilerStore.getDataForRoot(root);
     expect(data.commitData).toHaveLength(1);
+    expect(data.operations).toHaveLength(1);
```
bvaughn: This check fails without the newly-added bailout fix.

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -1021,12 +1021,32 @@ export function attach(
...
+        alternate !== null &&
+        alternate.expirationTime === 0 &&
+        alternate.childExpirationTime === 0;
+
+      // The Profiler stores metadata for each commit and reconstructs the app tree per commit using:
+      // (1) an initial tree snapshot and
+      // (2) the operations array for each commit
+      // Because of this, it's important that the operations and metadata arrays align,
+      // So the logic that skips metadata for bailout commits should also apply to filter operations.
+      if (didBailoutAtRoot) {
```
bvaughn: The added complexity for this bailout sucks. Probably worth it to avoid showing the empty commits in the Profiler and confusing people (see #16980).

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -1021,12 +1021,32 @@ export function attach(
...
+      const alternate = current.alternate;
+
+      // Certain types of updates bail out at the root without doing any actual render work.
+      // React should probably not call the DevTools commit hook in this case,
+      // but if it does- we can detect it and filter them out from the profiler.
+      // NOTE: Keep this logic in sync with the one in handleCommitFiberRoot()
+      const didBailoutAtRoot =
+        alternate !== null &&
+        alternate.expirationTime === 0 &&
+        alternate.childExpirationTime === 0;
```
gaearon: You already know this but this is very fragile and adds a dependency on fields we didn’t use before. I believe these might even be the fields Andrew is currently refactoring.

Can we make a proper fix in React itself instead?

bvaughn:
Okay, thanks Sunil. I'm going to move forward with this change since it *improves* the previous regression, but I'm still looking into a better approach that fixes the pre-existing bug as well.

packages/react-devtools-shared/src/backend/renderer.js
gaearon: Actually never mind, I see we already depend on these fields below. 😳

bvaughn: Oops, I literally *just* merged before your comment came in.

The reason I added this check in PR #17253 was because it could *also* improve the Profiler problem for older versions of React. If we think it's becoming unstable due to @acdlite refactoring, we should remove it and fix it properly in React: see PR #17267

bvaughn:
> and instead somehow filter out empty commits at a later stage. Such as during UI aggregation or in the store.

I'm working toward doing *just* that in a local branch. It's failing on a case that I haven't figured out yet though. This PR is just a bandaid fix for the meanwhile.

PR #17771, opened 2020-01-03
DevTools Profiler: Improve how empty commits are filtered

### LINKED ISSUE #16446: Error: "Could not find node with id "266" in commit tree"

bvaughn:
DevTools Profiler: Improve how empty commits are filtered

Previously, DevTools filtered empty commits on the backend, while profiling, through the use of a bailout heuristic that React currently happens to use. This approach was brittle and may have exacerbated the long-standing Profiler bug #16446.

This PR removes that heuristic and adds as a post-processing filtering pass instead. This removes the coupling between DevTools and a React internal implementation detail that may change.

I believe DevTools has two choices of criteria for this filtering:
* Filter commits that have no actual duration metadata.
* Filter commits that have no recorded operations (no mutations to the tree, no changed tree base durations).

I chose the first option, filtering by commits that have no reported metadata. It will miss an edge case, `unmountComponentAtNode`, but we would have nothing meaningful to show in the Profiler for that case anyway. (This particular case is why one of the snapshots changed with this commit.)

The second option, filtering by recorded operations, could potentially miss a more important edge case: where a component *did* render, but its `treeBaseDuration` didn't change. (In that event, there would be no tree `operations` sent.)

I believe this PR might resolve the longstanding issue #16446 but since it's very hard to reproduce, I'm not sure.

### DIFF
--- a/packages/react-devtools-shared/src/__tests__/__snapshots__/profilingCache-test.js.snap
+++ b/packages/react-devtools-shared/src/__tests__/__snapshots__/profilingCache-test.js.snap
@@ -1500,17 +1500,7 @@ Object {
 
 exports[`ProfilingCache should collect data for each root (including ones added or mounted after profiling started): Data for root Parent 3`] = `
 Object {
-  "commitData": Array [
-    Object {
-      "changeDescriptions": Map {},
-      "duration": 0,
-      "fiberActualDurations": Map {},
-      "fiberSelfDurations": Map {},
-      "interactionIDs": Array [],
-      "priorityLevel": "Immediate",
-      "timestamp": 34,
-    },
-  ],
+  "commitData": Array [],
   "displayName": "Parent",
   "initialTreeBaseDurations": Map {
     6 => 11,
@@ -1520,19 +1510,7 @@ Object {
   },
   "interactionCommits": Map {},
   "interactions": Map {},
-  "operations": Array [
-    Array [
-      1,
-      6,
-      0,
-      2,
-      4,
-      9,
-      8,
-      7,
-      6,
-    ],
-  ],
+  "operations": Array [],
   "rootID": 6,
   "snapshots": Map {
     6 => Object {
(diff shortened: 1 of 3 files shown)


packages/react-devtools-shared/src/__tests__/__snapshots__/profilingCache-test.js.snap
```diff
@@ -1520,19 +1510,7 @@ Object {
...
-    Array [
-      1,
-      6,
-      0,
-      2,
-      4,
-      9,
-      8,
-      7,
-      6,
```
bvaughn: Just to better annotate this change:
```sh
  1,  # rendered ID
  6,  # root ID
  0,  # string table isze
  2,  # TREE_OPERATION_REMOVE operation
  4,  # number of Fibers removed
  9,  # ids...
  8,  
  7,  
  6,
```

As explained in the PR description, it was expected.

PR #17772, opened 2020-01-03
[DO NOT MERGE] Root event marks

bvaughn:
Creating this work in progress PR to simplify the process of syncing the build to FB for testing.


bvaughn:
This PR is for testing. Don't need to actually keep it open.

PR #17773, opened 2020-01-04
Remove toString of dangerouslySetInnerHTML

sebmarkbage:
As far as I can tell, the toString call was added here:

https://github.com/facebook/react/commit/caae627cd557812d28d11237b34bff6c661ea8bc#diff-5574f655d491348f422bca600ff6711dR887

However, we don't toString the HTML in the initial creation. Only updates. It was never really needed.

Subsequently when we added Trusted Types, this needed to be changed to a special call but we really should just always let it pass through.

_Interestingly we should probably always let values pass through except for
when we need to do something special with the string which is rare._

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -85,7 +85,6 @@ import possibleStandardNames from '../shared/possibleStandardNames';
 import {validateProperties as validateARIAProperties} from '../shared/ReactDOMInvalidARIAHook';
 import {validateProperties as validateInputProperties} from '../shared/ReactDOMNullInputValuePropHook';
 import {validateProperties as validateUnknownProperties} from '../shared/ReactDOMUnknownPropertyHook';
-import {toStringOrTrustedType} from './ToStringValue';
 
 import {
   enableDeprecatedFlareAPI,
@@ -787,10 +786,7 @@ export function diffProperties(
       const lastHtml = lastProp ? lastProp[HTML] : undefined;
       if (nextHtml != null) {
         if (lastHtml !== nextHtml) {
-          (updatePayload = updatePayload || []).push(
-            propKey,
-            toStringOrTrustedType(nextHtml),
-          );
+          (updatePayload = updatePayload || []).push(propKey, nextHtml);
         }
       } else {
         // TODO: It might be too late to clear this if we have children


koto:
I only noticed this PR today. Thanks Sebastian!

To my understanding, there's still explicit stringification left in [DOMPropertyOperations](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/DOMPropertyOperations.js), that only is removed when the TT feature flag is enabled. Is there a reason why React explicitly stringifies those values?  Is it only because of IE8/9 issue where DOM APIs wouldn't stringify on their own, or is there some other reason?

The ulterior motive for the questions is of course enabling TT support by default in React, but I want to understand the existing constraints. For React DOM stringification is pretty much the only blocker.

gaearon:
Yes, I think this was because of IE9 (which we have not yet officially dropped).
We plan to enable the TT flag in 18.

koto:
If this is only about the IE9 bug, perhaps we could feature test for this bug once, instead of stringifying always? That should work and would be quite simple. The only backwards compatibility issue would be for environments that monkey patch DOM APIs as React might start passing non-string values. It has bitten us in the past, but feels like a reasonable change for a major version release.

gaearon:
What is the motivation? Is it to enable TT support sooner? If we made this change alone but did not enable the flag, what else would be missing for a full integration?

Siegrift:
> What is the motivation? Is it to enable TT support sooner?

Yes.

> If we made this change alone but did not enable the flag, what else would be missing for a full integration?

For client side integration the only problem is stringification of values which are later assigned to DOM sinks. This is already handled with TT support behind the feature flag. The missing part would be to handle user warnings when [using script tags in components](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMComponent.js#L431) and [when using dangerouslySetInnerHTML in svg element](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/setInnerHTML.js#L30).

gaearon:
I think we're open to a PR doing feature testing if you can verify it actually works on IE9. (You'd also need a few polyfills: Map/Set/rAF.)

koto:
> > If we made this change alone but did not enable the flag, what else would be missing for a full integration?
> 
> For client side integration the only problem is stringification of values which are later assigned to DOM sinks. This is already handled with TT support behind the feature flag. 

To clarify, removing the stringification would make majority of the apps TT-ready (i.e. applications wishing to enforce TT would add an appropriate header and start producing TT values when interpolating into sensitive sinks). It's quite likely this single change could be added without backwards compatibility issues - that's what we're trying to find out.

> The missing part would be to handle user warnings when [using script tags in components](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMComponent.js#L431) and [when using dangerouslySetInnerHTML in svg element](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/setInnerHTML.js#L30).

Only a warning in `__DEV__` mode in gated on the flag there, the behavior stays the same. I don't think this ever needs to change. Ultimately there's several ways of how to approach these:

- supporting these patterns and not warning devs at all, 
- documenting as a known issue, and describing workarounds (in React, or elsewhere)
- warning gated on a flag (current behavior)
- removing the support for these patterns (with deprecation warnings?)

All are perfectly valid - and these issues are not blocking for Trusted Type adopters, whereas stringification *is* blocking (not terribly, but some  patterns require workarounds e.g. one can't interpolate into `<iframe srcdoc>`).

PR #17774, opened 2020-01-04
Remove special casing of toString values when enableTrustedTypesIntegration

sebmarkbage:
As far as I can tell, we only toString in user space because of IE8/9. I tested in IE10 and IE11 and it doesn't seem to be an issue there.

We don't really support IE8/9 anymore and by the time this flag is on, it should be fine to stop supporting IE8/9. Arguably it's already ok to break it since we've done so with other IE8/9 things.


sebmarkbage:
I also removed the fancy typed abstractions. They actually provide more harm than they help. It makes it seem like you should go through them to get to a string or string like object but often you should just special case it at every callsite. Like this one: https://github.com/facebook/react/pull/17322/files#r362947154

The abstractions makes it seem like it can be used in other places but it's really only useful in one place and so we should inline it in that one place.

gaearon:
I think we do support IE9 when Map/Set/rAF polyfills are on. I don't think we added anything that would have broken it yet. But breaking it in 17 makes sense to me.

Issue #16456, opened 2019-08-19
React DevTools: styling overrides global styles


CompuIves:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
This is a very specific case - there are some styles leaking from react devtools, if you use it in your own application directly. Specifically these lines: https://github.com/facebook/react/blob/devtools-v4-merge/packages/react-devtools-shared/src/devtools/views/root.css#L174-L178.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

It's funny, I can give a *real* CodeSandbox editor now! You can see the styling on the editor elements here: https://codesandbox.io/s/new.

**What is the expected behavior?**

No leaking styles. Ideally the `* {` would be prefixed by an id or classname specific for the devtools. I think this should be possible with the styling used by the devtools. I can open a PR to fix this if everyone agrees on this fix.

cc @bvaughn

bvaughn:
Ah, interesting. Sorry about that!

> No leaking styles. Ideally the * { would be prefixed by an id or classname specific for the devtools. I think this should be possible with the styling used by the devtools.

I think think this might be trickier than that. I don't think that scoping to a style would effectively override the box sizing of element types _nested within_ DevTools, which is pretty important for layout in some cases.

A PR would be welcome though! :smile:

GasimGasimzada:
I tried to create a simple sample to show if using class name and wildcard selector together works: https://codepen.io/gasim/pen/mdyqEdd. 

Because wildcard is the last item to be applied within a style, I think nesting the styles within `.DevTools + *` is going to work (I will test it out tomorrow by building dev tools from the source). The only issue is if an element above the above DevTools component also has a similar selector (btw this is how CodeSandbox overrides React DevTools wildcard selector styles) :

```jsx
<div className="some-container">
  <div className="DevTools">...</div>
</div>
```

```css
.DevTools {
  box-sizing: border-box;
}

.some-container * {
  box-sizing: content-box;
}
```

Order matters here. If some container's CSS is loaded before .DevTools CSS, everything will be fine. Otherwise, the layout is going to break. However, without scoping the wildcard with a class name will fail anyways, regardless of the load order of selectors.

[...]

PR #17775, opened 2020-01-04
Scope dev tools wildcard styles within DevTools CSS class

GasimGasimzada:
Previously, wildcard selector was used to apply global styles that are related to DevTools. However, due to the fact that wildcard selectors will be used throughout the website, `react-devtools-inline` package bleeds out its styles outside its scope. So, in this PR, I am scoping these styles within `DevTools` main container's class name.

In this PR, I have added selectors for DevTools class and all the items within DevTools class.

Applied styles were tested using React DevTools app (package: `react-devtools`) by inspecting them through Electron's chrome inspector.

Referencing issue: https://github.com/facebook/react/issues/16456

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/DevTools.css
+++ b/packages/react-devtools-shared/src/devtools/views/DevTools.css
@@ -53,3 +53,9 @@
     display: none;
   }
 }
+
+.DevTools, .DevTools * {
+  box-sizing: border-box;
+
+  -webkit-font-smoothing: var(--font-smoothing);
+}
--- a/packages/react-devtools-shared/src/devtools/views/root.css
+++ b/packages/react-devtools-shared/src/devtools/views/root.css
@@ -182,9 +182,3 @@
   --interaction-commit-size: 10px;
   --interaction-label-width: 200px;
 }
-
-* {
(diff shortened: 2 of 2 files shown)


bvaughn:
Small thing I noticed is that this slightly breaks the test shell:
bug repro demo

bvaughn:
This change seems okay inside the extensions themselves.

bvaughn:
Significant regression inside of the standalone Electron app 😆 
Screen Shot 2020-01-06 at 10 30 41 AM

GasimGasimzada:
@bvaughn sorry, I am still getting used to Github Notifications (I have missed all the changes here somehow :angry:) 

> Significant regression inside of the standalone Electron app 😆

Was this one also fixed? Sorry for asking but what was the issue here? Everything seems fine in the screenshot...

bvaughn:
No worries, @GasimGasimzada. The problem shown in the screenshot is the large white pixel padding inside of the electron window. DevTools should fill that window, edge-to-edge. I fixed it before merging though!

GasimGasimzada:
Oh makes sense! Thank you for the fix!

bvaughn:

Thank you!

PR #17797, opened 2020-01-07
Renamed "ReactDOM-fb" imports to "ReactDOM" in www shims

bvaughn:
`ReactDOM-fb` was recently removed in www, which would cause all of these shims to break on the next sync.


scripts/rollup/shims/facebook-www/ReactDOMComponentTree_DO_NOT_USE.js
```diff
@@ -9,7 +9,7 @@
 
 const {
   __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED,
-} = require('ReactDOM-fb');
+} = require('ReactDOM');
```
gaearon: This file got deleted. Kill it here too?

gaearon: Meaning this shim.

bvaughn: Good call. Done.

PR #17798, opened 2020-01-07
Rename internal fields

sebmarkbage:
One day we'll use private fields for this but until then I'm just going to rename these every version until people get the hint.

This is actually better than if we change fields while keeping the same name. Because at least this will fail early (and often bailout) where as once found, people tend to assume that the data structure is there and as such throws random  errors to users.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponentTree.js
+++ b/packages/react-dom/src/client/ReactDOMComponentTree.js
@@ -18,9 +18,9 @@ import {getParentSuspenseInstance} from './ReactDOMHostConfig';
 const randomKey = Math.random()
   .toString(36)
   .slice(2);
-const internalInstanceKey = '__reactInternalInstance$' + randomKey;
-const internalEventHandlersKey = '__reactEventHandlers$' + randomKey;
-const internalContainerInstanceKey = '__reactContainere$' + randomKey;
+const internalInstanceKey = '__reactFiber$' + randomKey;
+const internalEventHandlersKey = '__reactEvents$' + randomKey;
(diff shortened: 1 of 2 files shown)


gaearon:
>I'm just going to rename these every version

Like.. manually?..

packages/react-dom/src/client/ReactDOMComponentTree.js
```diff
@@ -18,9 +18,9 @@ import {getParentSuspenseInstance} from './ReactDOMHostConfig';
 const randomKey = Math.random()
   .toString(36)
   .slice(2);
-const internalInstanceKey = '__reactInternalInstance$' + randomKey;
```
gaearon: https://github.com/facebook/react/blob/0545f366d4d6b5959f4bb172e810c745f74b9513/packages/react-devtools-shared/src/backend/views/Highlighter/Overlay.js#L284

sebmarkbage: We should never do these hacks in DevTools, we should use the injected APIs to do it or add a new injected API if we don't have one.

bvaughn: I think this (the overlay plugin) was ported without a ton of refactoring from the v3 codebase.

The nice thing about DevTools is that it lives in the React repo so we can clean any hacks like this up as needed 😄

#17822

bvaughn: P.S. #17841

gaearon:
We did this.

PR #17799, opened 2020-01-07
Revert "Revert "Dispatch commands to both UIManagers from both renderers (#17211)" (#17232)"

TheSavior:
This reverts commit d0fc0ba0a688950b8ab24a89f14888a19efa2444.

This is a reland of #17211.

This is our second attempt at making calls to dispatchCommand in Fabric go through the FabricUIManager. This had to be reverted before because the native code didn't have these commands implemented yet for Fabric. They do now so this is safe to land.


JoshuaGross:
Looks good to me 👍

PR #17805, opened 2020-01-08
Update Flow to 0.84

gaearon:
Updating to v0.84 first is a first step that avoids breaking changes in subsequent versions of Flow. v0.84 includes the new inexact object syntax, which is necessary to align with fbsource and unblock RN sync fix. Afterwards we can continue updating Flow in increments.


necolas:
Why upgrade to version 0.84 specifically?

gaearon:
They said it should be easy compared to some big breaking change after. It also gives us new inexact object syntax, which is necessary to align with fbsource and unblock RN sync fix.

gaearon:
We should upgrade to a later version too for sure but I started with this as MVP.

yarn.lock
```diff
@@ -4864,6 +4864,7 @@ eslint-plugin-no-unsafe-innerhtml@1.0.16:
 
 "eslint-plugin-react-internal@link:./scripts/eslint-rules":
   version "0.0.0"
+  uid ""
```
TrySound: Is this an accident?

gaearon: I just ran Yarn. No idea.

PR #17816, opened 2020-01-09
Add better test coverage for hooks in DevTools Profiler's change-detection

bvaughn:


packages/react-devtools-shared/src/__tests__/profilingCache-test.js
```diff
@@ -308,6 +308,130 @@ describe('ProfilingCache', () => {
...
+    );
+
+    // Third render has a changed reducer hook
+    utils.act(() => dispatch({type: 'invert'}));
+
+    // Fourth render has a changed state hook
+    utils.act(() => setState('def'));
+
+    // Fifth render has a changed context value, but no changed hook.
+    // Technically, DevTools will miss this "context" change since it only tracks legacy context.
```
bvaughn: This sucks, but there's no way around it short of DevTools overriding the current dispatcher's `readContext` method and tracking context values per-component, and that feels like too much overhead to justify.

bvaughn:
> Assuming you read the snapshots

Yes 😆

PR #17817, opened 2020-01-10
Fix issue template for questions

necolas:
Fixes the syntax for adding multiple labels.

See formatting is now correct: https://github.com/necolas/react/blob/github-question-template/.github/ISSUE_TEMPLATE/question.md


.github/ISSUE_TEMPLATE/question.md
```diff
@@ -20,7 +20,7 @@ If you have a coding question related to React and React DOM, it might be better
 
 ### https://www.reactiflux.com/
```
gaearon: Not everyone might be comfortable asking on a chat like Discord. Can we point folks here instead and let them choose? https://reactjs.org/community/support.html#popular-discussion-forums

PR #17824, opened 2020-01-12
[Fresh] Fix edge case with early function call

gaearon:
When a component calls custom Hooks, we "collect" a signature in two steps.

https://github.com/facebook/react/blob/64aae7b06fa47e126b0ba9c0ba9896caa803528e/packages/react-refresh/src/ReactFreshRuntime.js#L582-L603

For example, given this component:

```js
function App() {
  const x = useFancyState('x')
  const y = useFancyState('x')
}

function useFancyState(value) {
  return React.useState(value)
}
```

we want to:

1. First, establish that `App` has two `useFancyState` calls.
   - Our transform does this with a call like `_s(App, 'useFancyState(x) useFancyState(y)', () => [useFancyState, useFancyState])`
2. Then, when we're inside `App` for the first time, trigger the function to save their identities
   - Our transform does this with a call like `_s()`

We intentionally trigger the function to save function identities lazily on first render. This is so that we don't change the module definition order when you have inline requires.

We reuse the same function to minimize the transform noise. So we count calls. First call saves `() => [useFancyState, useFancyState]`, second call actually triggers it to get references. Next calls don't do anything.

This usually works, but it breaks in the corner case like this:

```js
render(<App />);

function App() { ... }
function useFancyState() { ... }
```

This is because we end up *inside* `App` before the *first* call injected by our transform runs. So we get the "inside" call before we get the "outside" call.

This is edge casey but can happen in environments like CodeSandbox where you write all code in one file.

The fix is to not rely on the call index, and instead to look at what we're passing. With this fix, we ignore all empty `_s()` calls until after we actually get the first `_s(App, ...)` registration call. To make this clearer, I'm changing a number tracking the call number to a status enum.

A regression test is added. This should fix [this broken case](https://codesandbox.io/s/invalid-value-used-as-weak-map-key-broken-react-fresh-y2mx3) on CodeSandbox. (After CodeSandbox updates `react-refresh` to the version with this fix.)

### DIFF
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -601,9 +601,14 @@ export function _getMountedRootCount() {
 //   'useState{[foo, setFoo]}(0)',
 //   () => [useCustomHook], /* Lazy to avoid triggering inline requires */
 // );
+type SignatureStatus = 'needsSignature' | 'needsCustomHooks' | 'resolved';
 export function createSignatureFunctionForTransform() {
   if (__DEV__) {
-    let call = 0;
+    // We'll fill in the signature in two steps.
+    // First, we'll know the signature itself. This happens outside the component.
+    // Then, we'll know the references to custom Hooks. This happens inside the component.
+    // After that, the returned function will be a fast path no-op.
+    let status: SignatureStatus = 'needsSignature';
     let savedType;
     let hasCustomHooks;
     return function<T>(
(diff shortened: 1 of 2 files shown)


packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -601,9 +601,14 @@ export function _getMountedRootCount() {
 //   'useState{[foo, setFoo]}(0)',
 //   () => [useCustomHook], /* Lazy to avoid triggering inline requires */
 // );
+type SignatureStatus = 'needsSignature' | 'needsCustomHooks' | 'resolved';
```
bvaughn: Nice readability change 👍

thgh:
Hi, I'm having this issue in Expo which seems to be using version 0.4.2. I'm using patch-package for now to add this line:
```
diff --git a/node_modules/react-refresh/cjs/react-refresh-runtime.development.js b/node_modules/react-refresh/cjs/react-refresh-runtime.development.js
index 3e017a2..27b49af 100644
--- a/node_modules/react-refresh/cjs/react-refresh-runtime.development.js
+++ b/node_modules/react-refresh/cjs/react-refresh-runtime.development.js
@@ -279,6 +279,7 @@ function register(type, id) {
   }
 }
 function setSignature(type, key) {
+  if (!key) return
   var forceReset = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : false;
   var getCustomHooks = arguments.length > 3 ? arguments[3] : undefined;

```

Would you mind to release this fix as 0.4.3?

gaearon:
What is the actual code triggering this in your case?

gaearon:
I released this PR as 0.4.3.

thgh:
Awesome! It works 😃 actual code:

```tsx
const App = () => <Text>{useUser()}</Text>

// imported
const [useUser, setUser] = createGlobalHook<string>('')
export { useUser, setUser }
fetchUser().then(setUser)

// imported
export function createGlobalHook(state) {
  let effects: any[] = []
  return [useGlobal, setGlobal, get]

  function useGlobal() {
    const [, effect] = useState(state)
    useEffect(() => {
      effects = effects.concat(effect)
      return () => effects = effects.filter(eff => eff !== effect)
    }, [])
    return state
  }

  function setGlobal(newState) {
    if (typeof newState === 'function') {
      state = newState(state)
    } else {
      state = newState
    }
    effects.forEach(s => s(state))
    return state
  }

  function get() {
    return state
  }
}
```

PR #17834, opened 2020-01-14
[fail] move IsThisRendererActing to own namespace

### LINKED ISSUE #17794: flushPassiveEffects and IsThisRendererActing are on the wrong "secret" object
They're currently attached to the "Events" namespace but they don't have anything to do with events. It needs to move to its own names on the secret object.

Once we delete Events, after ReactTestUtils and RNW are fixed, then we won't be able to clean up the array because these two are on it.

Additionally the IsSomeRendererActing field is in the production bundles.

We need to make all of these DEV-only.

threepointone:
The `Events` namespace on the secret object might be disappearing, so `IsThisRendererActing` can't live on it anymore. This PR -
- moves `IsThisRendererActing` from `ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Events` to `ReactDOM....Tests`

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -169,8 +169,8 @@ const ReactDOM: Object = {
       dispatchEvent,
       runEventsInBatch,
       flushPassiveEffects,
-      IsThisRendererActing,
     ],
+    Tests: [IsThisRendererActing],
   },
 };
 
--- a/packages/react-dom/src/test-utils/ReactTestUtilsAct.js
+++ b/packages/react-dom/src/test-utils/ReactTestUtilsAct.js
@@ -31,8 +31,11 @@ const [
   runEventsInBatch,
   /* eslint-enable no-unused-vars */
   flushPassiveEffects,
-  IsThisRendererActing,
 ] = ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Events;
+
+const [
+  IsThisRendererActing,
+] = ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Tests;
 
 const batchedUpdates = ReactDOM.unstable_batchedUpdates;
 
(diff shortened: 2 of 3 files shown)


packages/react-dom/src/client/ReactDOM.js
```diff
@@ -188,6 +187,11 @@ if (exposeConcurrentModeAPIs) {
     }
   };
 }
+if (__DEV__) {
+  ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Tests = [
```
threepointone: I could just call this something like `__SECRET_TEST_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED` on `ReactDOM` directly.

threepointone:
Bah, some tests failing with prod builds. Investigating.

threepointone:
I removed the __DEV__ checks, it looks like *we* use act() heavily across prod tests. That bit will take another PR and some discussion.

packages/react-dom/src/test-utils/ReactTestUtilsAct.js
```diff
@@ -31,9 +31,12 @@ const [
   runEventsInBatch,
   /* eslint-enable no-unused-vars */
   flushPassiveEffects,
-  IsThisRendererActing,
 ] = ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Events;
 
+const [
+  IsThisRendererActing,
+] = ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.Tests;
```
sebmarkbage: How about we just don't destructure nor set `IsThisRendererActing` in ReactTestUtilsAct for PROD?

It's only ever read by React in DEV mode so why bother writing it in PROD? Even if we do use the rest of act in prod.

sebmarkbage:
The issue I filed wasn't very clear but it refers to both `IsSomeRendererActing` and `IsThisRendererActing`.

I believe `IsSomeRendererActing` can also be made DEV only since it's only ever read in a DEV branch.

https://github.com/facebook/react/blob/a457e02ae3a2d3903fcf8748380b1cc293a2445e/packages/react/src/ReactSharedInternals.js#L19

https://github.com/facebook/react/blob/e706721490e50d0bd6af2cd933dbf857fd8b61ed/packages/react-reconciler/src/ReactFiberWorkLoop.js#L2717

threepointone:
closing this, will revisit right before 17

PR #17835, opened 2020-01-14
[react-interactions] Add DO_NOT_USE to Scope methods

trueadm:
In attempt to make it clearer about existing experimental APIs and their internal usage. In the long term, the Scopes API is expected to change/go-away but for now, given it's internal usage we'll have to find incremental ways to safely migrate to better options. This PR is one small step of many in order to do that. This PR makes the following changes:

- Adds a `DO_NOT_USE` prefix to most Scope API methods that are potentially dangerous and will likely change or go away.
- Provides a new experiemental method call `getChildContextValues` that attempts to work with existing React Context. It find the nearest child context values for a given type and returns them as part of an array. The idea is to internally use this approach and gradually deprecate and remove the DO_NOT_USE APIs.

### DIFF
--- a/packages/react-interactions/accessibility/README.md
+++ b/packages/react-interactions/accessibility/README.md
@@ -14,73 +14,10 @@ can be found [here](./docs).
 
 Note: React Scopes require the internal React flag `enableScopeAPI`.
 
-When creating a scope, a query function is required. The query function is used
-when collecting host nodes that match the criteria of the query function.
-
-```jsx
-// This query function only matches host nodes that have the type of "div"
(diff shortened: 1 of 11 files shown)


packages/react-interactions/accessibility/README.md
```diff
@@ -14,74 +14,11 @@ can be found [here](./docs).
 
 Note: React Scopes require the internal React flag `enableScopeAPI`.
 
-When creating a scope, a query function is required. The query function is used
```
trueadm: I'm removing this document as we really don't want people to be actively using any of these APIs right now.

PR #17837, opened 2020-01-14
Add support for Node v13 for development

jasonwilliams:
## Summary

I can't build with Node v13 installed.
This should fix that


gaearon:
Thanks Jason!

PR #17840, opened 2020-01-14
[Fresh] Support DevTools v3

gaearon:
I haven't verified but I think this would fix the issue we've seen with the old version of the global hook.

### DIFF
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -472,7 +472,17 @@ export function injectIntoGlobalHook(globalObject: any): void {
     // Do the same for any already injected roots.
     // This is useful if ReactDOM has already been initialized.
     // https://github.com/facebook/react/issues/17626
-    const renderers = hook.renderers || hook._renderers; // DevTools v4 || v3
+    let renderers = hook.renderers;
+    if (renderers === undefined && hook._renderers != null) {
+      // DevTools v3
+      renderers = new Map();
+      for (let key in hook._renderers) {
+        if (hook._renderers.hasOwnProperty(key)) {
+          renderers.set(key, hook._renderers[key]);
+        }
+      }
+    }
+
     renderers.forEach((injected, id) => {
       if (
         typeof injected.scheduleRefresh === 'function' &&
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -472,7 +472,8 @@ export function injectIntoGlobalHook(globalObject: any): void {
     // Do the same for any already injected roots.
     // This is useful if ReactDOM has already been initialized.
     // https://github.com/facebook/react/issues/17626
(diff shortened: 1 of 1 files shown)


packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -472,7 +472,8 @@ export function injectIntoGlobalHook(globalObject: any): void {
     // Do the same for any already injected roots.
     // This is useful if ReactDOM has already been initialized.
     // https://github.com/facebook/react/issues/17626
-    hook.renderers.forEach((injected, id) => {
+    const renderers = hook.renderers || hook._renderers; // DevTools v4 || v3
+    renderers.forEach((injected, id) => {
```
bvaughn: In v4, [`renderers` is a `Map`](https://github.com/facebook/react/blob/1e1a98942225771f86d8a81b5ac561a9fbff9263/packages/react-devtools-shared/src/hook.js#L275) but in v3 [`_renderers` is an `Object`](https://github.com/facebook/react-devtools/blob/d839081f79f681617df7584a8ffd2f4163fa40b9/backend/installGlobalHook.js#L127). So `.forEach()` wouldn't work on v3. You'd need to do a different loop for v3.

gaearon: I admire your diligence.

bvaughn: image

Mudassar045:
@gaearon  can possible `hook._renderers === undefined` ?

gaearon:
Yes, `!= null` checks for that.

Mudassar045:
> Yes, `!= null` checks for that.

JavaScript is so fishy... thanks I learned that `undefined == null` returns `true` and `undefined === null` returns `false`

packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -472,7 +472,18 @@ export function injectIntoGlobalHook(globalObject: any): void {
     // Do the same for any already injected roots.
     // This is useful if ReactDOM has already been initialized.
     // https://github.com/facebook/react/issues/17626
-    hook.renderers.forEach((injected, id) => {
+    let renderers = hook.renderers;
```
jamesgeorge007:
```suggestion
    let { renderers}  = hook;
```

jamesgeorge007: I'm a big fan of destructuring assignment approach. The former code snippet appears kinda redundant atleast IMO.

jamesgeorge007:
```suggestion
    let { renderers } = hook;
```

jamesgeorge007: I'm a big fan of destructuring assignment approach. The former code snippet appears kinda redundant atleast IMO.

gaearon:
Flow want something. I'll get back to this maybe.

PR #17842, opened 2020-01-14
Update to flow-bin@0.89

necolas:
## Summary

Incremental Flow update.
Versions above 0.89 introduce new types of error.

## Test Plan

`yarn flow dom`


bvaughn:
The DevTools `alphaSortEntries` Flow error is correctly pointing out that I'm passing it `.keys()` rather than `.entries()`. It just happens to work because of string index hashing. I'll shoot you a PR in a moment that fixes this.

PR #17852, opened 2020-01-16
Fresh cleanup

mamal72:
## Summary

I was trying to read react-refresh but it wasn't easy to read for me as there were a lot of big nested if blocks there with a one-liner else block.

## Test Plan

Nothing comes to my mind about this part.

### DIFF
--- a/packages/react-refresh/src/ReactFreshRuntime.js
+++ b/packages/react-refresh/src/ReactFreshRuntime.js
@@ -669,17 +669,17 @@ export function isLikelyComponentType(type: any): boolean {
       return typeof name === 'string' && /^[A-Z]/.test(name);
     }
     case 'object': {
-      if (type != null) {
-        switch (type.$$typeof) {
-          case REACT_FORWARD_REF_TYPE:
-          case REACT_MEMO_TYPE:
-            // Definitely React components.
-            return true;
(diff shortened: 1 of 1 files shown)


packages/react-refresh/src/ReactFreshRuntime.js
```diff
@@ -294,48 +294,47 @@ export function performReactRefresh(): RefreshUpdate | null {
...
-        case REACT_FORWARD_REF_TYPE:
-          register(type.render, id + '$render');
-          break;
-        case REACT_MEMO_TYPE:
-          register(type.type, id + '$type');
-          break;
-      }
-    }
-  } else {
+  if (!__DEV__) {
```
dai-shi: Oops. `__DEV__` becomes falsy in production... We need `else`.

mamal72: Can you explain it a bit more? Are transpilers somehow remove the dead codes in if conditions? If it's the case and they don't support this behavior, I think this PR should get closed. 🤔

mamal72: Yeah. 🤔

Thanks for the review! I'll try to find something else to clean or contribute. 👍

PR #17859, opened 2020-01-17
[react-interactions] Remove deprecated Scope APIs + update Focus components

trueadm:
This PR removes the deprecated Scope APIs that handled traversal. These APIs are no longer used internally, and instead now use React Context and the `scope.getChildContextValues()` API exclusively.

Unfortunately, I've decided to also remove the `accessibility` directory from the React repository for now, which includes the accessibility components. Many of the tests/docs/components were out of date compared to their latest versions, which were migrated to our internal FB repository last year. Keeping the components in sync proved to be very time-consuming because of several differences. Notably, our internal Jest tests for these components made usage of our internal testing infrastructure which is not available in open-source right now, plus there were usages of other bits of Facebook IP that I'd have to remove before syncing back over.

A more effecient long-term strategy will be to work with the partner teams involved and come up with a plan that involves moving these components back into open-source, probably under a different repository maybe, or maybe a different project relating to Facebook's accessibility work.


JanJakes:
> Unfortunately, I've decided to also remove the accessibility directory from the React repository for now
About a year later, I'd like to ask if there's any plan or progress to bring these back.

PR #17864, opened 2020-01-18
Add different string ref warning when owner and self are different

lunaruan:
When owner and self are different for string refs, we can't easily convert them to callback refs. This PR adds a warning for string refs for everyone when owner and self are different to tell users to manually update these refs.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDeprecationWarnings-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDeprecationWarnings-test.internal.js
@@ -69,4 +69,44 @@ describe('ReactDeprecationWarnings', () => {
         '\n    in Component (at **)',
     );
   });
+
+  it('should not warn when owner and self are the same for string refs', () => {
+    ReactFeatureFlags.warnAboutStringRefs = false;
+
+    class RefComponent extends React.Component {
+      render() {
+        return null;
+      }
+    }
+    class Component extends React.Component {
+      render() {
+        return <RefComponent ref="refComponent" __self={this} />;
+      }
+    }
+    ReactNoop.renderLegacySyncRoot(<Component />);
+    expect(Scheduler).toFlushWithoutYielding();
+  });
+
+  it('should warn when owner and self are different for string refs', () => {
+    class RefComponent extends React.Component {
+      render() {
+        return null;
+      }
+    }
+    class Component extends React.Component {
+      render() {
+        return <RefComponent ref="refComponent" __self={{}} />;
+      }
+    }
+
+    ReactNoop.render(<Component />);
+    expect(() => expect(Scheduler).toFlushWithoutYielding()).toErrorDev([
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -120,16 +120,33 @@ function coerceRef(
...
-              componentName,
-              mixedRef,
-              getStackByFiberInDevAndProd(returnFiber),
-            );
+            if (
+              element._owner &&
+              element._self &&
+              element._owner.stateNode !== element._self
+            ) {
+              console.error(
```
sebmarkbage: The reason we put the other warning in this file is because it's the only way to know if you're in StrictMode or not.

However, it's a sub-optimal place to put it because the error stack isn't as good. It doesn't point out the exact callsite. Since we don't need to check for StrictMode and can just always warn, you can instead put this in React.jsx and React.createElement so that we get better error stacks.

packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -120,16 +120,33 @@ function coerceRef(
...
+                'Owner and self do not match for the string ref "%s". Support for ' +
```
sebmarkbage: This doesn't really tell the user anything. They don't know what owner and self not matching means. How about:

_Support for string refs will be removed in a future major release. This case cannot be automatically converted to an arrow function. We ask you to manually fixing this case by using useRef() or createRef() instead. Learn more about using refs safely here: https://fb.me/react-strict-mode-string-ref%s_

packages/react-dom/src/__tests__/ReactDeprecationWarnings-test.internal.js
```diff
@@ -69,4 +69,50 @@ describe('ReactDeprecationWarnings', () => {
         '\n    in Component (at **)',
     );
   });
+
+  it('should not warn when owner and self are the same for string refs', () => {
```
sebmarkbage: This says _not_ but the test says it warns. Probably because it's concurrent mode which implies strict.

Try using `ReactNoop.renderLegacySyncRoot` instead.

packages/react/src/ReactElement.js
```diff
@@ -89,6 +96,33 @@ function defineRefPropWarningGetter(props, displayName) {
...
+      ReactCurrentOwner.current.stateNode !== config.__self
+    ) {
+      const componentName = getComponentName(ReactCurrentOwner.current.type);
+
+      if (!didWarnAboutStringRefs[componentName]) {
+        console.error(
+          'Component "%s" contains the string ref "%s". ' +
+            'Support for string refs will be removed in a future major release. ' +
+            'This case cannot be automatically converted to an arrow function. ' +
+            'We as you to manually fix this case by using useRef() or createRef() instead. ' +
```
sebmarkbage: *We ask

PR #17873, opened 2020-01-20
Add feature flag around React.createFactory

trueadm:
This PR adds a `disableCreateFactory` feature flag that allows us to disable `React.createFactory` from being exported from the React package. I also had to update all tests that made use of this API and move them to `React.createElement`.

### DIFF
--- a/packages/react-is/src/__tests__/ReactIs-test.js
+++ b/packages/react-is/src/__tests__/ReactIs-test.js
@@ -13,6 +13,8 @@ let React;
 let ReactDOM;
 let ReactIs;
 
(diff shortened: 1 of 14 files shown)


sebmarkbage:
Before we actually remove this, we’ll need to run a warning for a while. Should we also have a flag to add the warning?

trueadm:
I thought we’d most likely just add the warning without a flag. I’ve removed all internal usages, so this would be OSS only.

sebmarkbage:
We should add the warning now then. I'm scared we'll forget it if we leave it.

trueadm:
Agreed @sebmarkbage. PR is up for review: https://github.com/facebook/react/pull/17878

PR #17875, opened 2020-01-20
Remove FB specific React isomorphic build

sebmarkbage:
I removed unstable_SuspenseList internally. We don't need the FB build. I plan on also removing the ReactDOM fb specific entry. We shouldn't add any more FB specific internals nor APIs. If they're experimental they should go onto the experimental builds to avoid too many permutations.


trueadm:
In regards to the FB specific build, it's a bit more complex (unless I misunderstand you). @gaearon can confirm, but we have a different set of GCC settings for the FB prod build that doesn't mangle property/variable names, to make internal debugging easier. We also apply specific code shims/transforms to certain modules for the FB builds.

PR #17880, opened 2020-01-21
Add unstable_renderSubtreeIntoContainer and unstable_createPortal feature flags

trueadm:
This PR adds two feature flags:

- `disableUnstableRenderSubtreeIntoContainer` for `ReactDOM.unstable_renderSubtreeIntoContainer`
- `disableUnstableCreatePortal` for `ReactDOM.unstable_createPortal`

These disable the relevant unstable/deprecated ReactDOM APIs. Additionally `ReactDOM.unstable_renderSubtreeIntoContainer` has a deprecation warning message as follows:

`The ReactDOM.unstable_renderSubtreeIntoContainer() is deprecated and will be removed in a future major release. Consider using React Portals instead.`

I also changed the flowconfig to have the follow config, otherwise CI fails on this PR:

`server.max_workers=4`

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -53,7 +53,11 @@ import {
 } from 'legacy-events/EventPropagators';
 import ReactVersion from 'shared/ReactVersion';
 import invariant from 'shared/invariant';
-import {exposeConcurrentModeAPIs} from 'shared/ReactFeatureFlags';
+import {
+  exposeConcurrentModeAPIs,
+  disableUnstableCreatePortal,
+  disableUnstableRenderSubtreeIntoContainer,
+} from 'shared/ReactFeatureFlags';
 
 import {
   getInstanceFromNode,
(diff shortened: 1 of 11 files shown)


packages/react-dom/src/client/ReactDOM.js
```diff
@@ -189,6 +176,41 @@ if (exposeConcurrentModeAPIs) {
   };
 }
 
+if (!disableUnstableRenderSubtreeIntoContainer) {
+  ReactDOM.unstable_renderSubtreeIntoContainer = function(...args) {
+    if (__DEV__) {
+      if (!didWarnAboutUnstableRenderSubtreeIntoContainer) {
+        didWarnAboutUnstableRenderSubtreeIntoContainer = true;
+        console.warn(
+          'The ReactDOM.unstable_renderSubtreeIntoContainer() is deprecated ' +
```
bvaughn: I think the word "The" should be removed.

mzedeler:
Just a question: does this mean that portals will be deprecated in the future?

threepointone:
No, in fact the opposite. It means the unstable_ prefixed version for createPortal is removed. Use just createPortal instead.

PR #17882, opened 2020-01-21
Update legacy context warning message

trueadm:
After speaking to @gaearon, we spoke about how we can improve the warning message when using StrictMode with legacy context. Previously, we would output all the offending components and the component stack of the StrictMode, which wasn't that useful in practice. With the changes in this PR, we now output the component stack of the component that most frequently had legacy context warnings in StrictMoode.

### DIFF
--- a/packages/react-reconciler/src/ReactStrictModeWarnings.js
+++ b/packages/react-reconciler/src/ReactStrictModeWarnings.js
@@ -303,15 +303,15 @@ if (__DEV__) {
     }
     const type = fiber.type;
 
-    // Update legacy context counts
-    const warningCount = legacyContextCounts.get(type) || 0;
-    // Increase warning count by 1
-    legacyContextCounts.set(type, warningCount + 1);
-
     // Dedup strategy: Warn once per component
     if (didWarnAboutLegacyContext.has(type)) {
       return;
     }
+
+    // Update legacy context counts
+    const warningCount = legacyContextCounts.get(type) || 0;
+    // Increase warning count by 1
+    legacyContextCounts.set(type, warningCount + 1);
 
     let warningsForRoot = pendingLegacyContextWarning.get(strictRoot);
 
--- a/packages/react-reconciler/src/ReactStrictModeWarnings.js
+++ b/packages/react-reconciler/src/ReactStrictModeWarnings.js
@@ -286,6 +286,7 @@ if (__DEV__) {
   let pendingLegacyContextWarning: FiberToFiberComponentsMap = new Map();
 
   // Tracks components we have already warned about.
+  const legacyContextCounts = new Map();
   const didWarnAboutLegacyContext = new Set();
 
   ReactStrictModeWarnings.recordLegacyContextWarning = (
@@ -300,17 +301,23 @@ if (__DEV__) {
       );
       return;
     }
+    const type = fiber.type;
 
-    // Dedup strategy: Warn once per component.
-    if (didWarnAboutLegacyContext.has(fiber.type)) {
+    // Update legacy context counts
+    const warningCount = legacyContextCounts.get(type) || 0;
+    // Increase warning count by 1
+    legacyContextCounts.set(type, warningCount + 1);
+
+    // Dedup strategy: Warn once per component
+    if (didWarnAboutLegacyContext.has(type)) {
       return;
     }
 
     let warningsForRoot = pendingLegacyContextWarning.get(strictRoot);
 
     if (
-      fiber.type.contextTypes != null ||
-      fiber.type.childContextTypes != null ||
+      type.contextTypes != null ||
+      type.childContextTypes != null ||
       (instance !== null && typeof instance.getChildContext === 'function')
     ) {
       if (warningsForRoot === undefined) {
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -324,26 +322,46 @@ if (__DEV__) {
...
+            mostFrequentStack = stack;
+          }
+        }
 
         console.error(
           'Legacy context API has been detected within a strict-mode tree.' +
             '\n\nThe old API will be supported in all 16.x releases, but applications ' +
             'using it should migrate to the new version.' +
-            '\n\nPlease update the following components: %s' +
+            '\n\nPlease update the following component: %s' +
```
gaearon: Soo.. I'm not sure about this. I think it's still valuable to see a list of all components so you have a sense of how many there are to go. But with the stack pointing to the most frequent one.

bvaughn: Agreed. Only showing one at a time could be pretty frustrating- like whack-a-mole, where you begin to wonder if the list will ever end. I think showing *all* component names (like before) is probably best.

Not sure if "the most frequent one" is necessarily valuable since you'll need to fix them all if you want to be CM compatible, but showing the location of [whichever] one would probably be a nice addition.

trueadm: I agree. I’ll add it back, with the most frequent component appearing first.

trueadm: This is resolved. I'd appreciate another look :)

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -324,27 +324,49 @@ if (__DEV__) {
...
+        // Sort the stacks by their counts
+        stacks.sort((a, b) => b.count - a.count);
+
+        const mostFrequentStack = stacks[0];
+
+        if (mostFrequentStack) {
+          // We map to a Set to remove duplicate component names
+          const componentNames = Array.from(
+            new Set(stacks.map(stack => stack.name)),
+          ).join(', ');
```
bvaughn: It's not a huge deal, since this is DEV-only code, but I wonder if there isn't a slightly more efficient way of doing this. Rather than creating all of the component stacks and then doing a full sort based on count, maybe we could just change `didWarnAboutLegacyContext` from a `Set` to a `Map` that counts the number of times each fiber would have warned, and then look for the fiber with the highest count and generate its stack only?

packages/react-reconciler/src/__tests__/ReactIncremental-test.internal.js
```diff
@@ -1916,8 +1916,7 @@ describe('ReactIncremental', () => {
       'Legacy context API has been detected within a strict-mode tree.\n\n' +
         'The old API will be supported in all 16.x releases, but applications ' +
         'using it should migrate to the new version.\n\n' +
-        'Please update the following components: Intl, ShowBoth, ShowLocale',
-      {withoutStack: true},
+        'Please update the following components: Intl, ShowLocale, ShowBoth',
```
bvaughn: Should any of these tests verify that the stack is something reasonable (that it belongs to one of those components)?

packages/react-reconciler/src/ReactStrictModeWarnings.js
trueadm: I mentioned this to @gaearon and he mentioned that this wouldn't be reliable, as we should be using the stack instead because different fibers might make the same stack. In terms of effeciency this seemed fine given it was DEV, it's not ideal, but given we don't want to use the Fiber as the key, it seems sensible.

packages/react-reconciler/src/__tests__/ReactIncremental-test.internal.js
trueadm: I don't follow? The shown stack (there's only one) should always be the first component in the comma separated list.

packages/react-reconciler/src/ReactStrictModeWarnings.js
bvaughn: > different fibers might make the same stack

Hmm...given that the stack is essentially just the name (and potentially source location) of each fiber from the current one through its parent path, the chance of two fibers generating the same stack seems unlikely? Am I misunderstanding something? Can you mention an example of where this might happen?

packages/react-reconciler/src/__tests__/ReactIncremental-test.internal.js
bvaughn: My point is that none of these tests actually verify that. They verify that there is *a stack* but not what's in it.

trueadm: There is a test that does, I belive it's in ReactStrictMode-test.js

bvaughn: Ah, ok. Fair enough.

packages/react-reconciler/src/ReactStrictModeWarnings.js
trueadm: I can't remember that exact details of what @gaearon said, he might be able to give a detailed explanation as it's late here and I'm a bit drained. However, if you had an alternate fiber, you'd run into this issue on updates (because we switch fiber) which would generate the same stack but would obviously be a referentially different fiber.

bvaughn: Our check could be for `fiber` or `fiber.alternate` to account for this. (We do similar things in DevTools to account for this.)

I guess I'm curious to see what Dan says about this then.

packages/react-reconciler/src/__tests__/ReactIncremental-test.internal.js
```diff
@@ -2053,7 +2048,7 @@ describe('ReactIncremental', () => {
           'using it should migrate to the new version.\n\n' +
           'Please update the following components: Recurse',
       ],
-      {withoutStack: 1},
+      {withoutStack: 0},
```
gaearon: This can be removed?

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -324,27 +324,49 @@ if (__DEV__) {
...
-        );
+        // Get the stacks from our componentStacks Map
+        const stacks = Array.from(componentStacks.values());
+
+        // Sort the stacks by their counts
+        stacks.sort((a, b) => b.count - a.count);
+
+        const mostFrequentStack = stacks[0];
+
+        if (mostFrequentStack) {
```
gaearon: Maybe let's remove this condition and instead do `mostFrequentStack ? mostFrequentStack.stack : ''` below when passing arguments? Seems a bit less fragile.

packages/react-reconciler/src/ReactStrictModeWarnings.js
gaearon: My thinking was that if we have 100 fibers with the same stack (e.g. list items), I'd like to their stacks to have 100 weight. Instead of each of them having 1 weight. Because fixing legacy context would have effect on 100 subtrees.

>maybe we could just change didWarnAboutLegacyContext from a Set to a Map that counts the number of times each fiber would have warned

Wouldn't each fiber only ever warn once? I'm not quite following this. Or do you mean a map by type?

I think a map from fiber type to count makes sense to me. "Fix the most impactful type first." You wouldn't see the most common *stack* for that type though, but maybe it's OK?

bvaughn: Fair enough, I guess e.g. list items would have the same stack if we didn't have file name and line number info.

> Wouldn't each fiber only ever warn once? I'm not quite following this. Or do you mean a map by type?

Sorry, yes. I meant mapping by type.

trueadm: Sounds good.

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -324,16 +331,30 @@ if (__DEV__) {
...
-        );
+        const fibers = fiberArray.length;
+        let componentNames = '';
+
+        fiberArray
+          .sort((a, b) => {
+            const aCount = legacyContextCounts.get(a.type) || 0;
+            const bCount = legacyContextCounts.get(b.type) || 0;
+            return bCount - aCount;
+          })
```
bvaughn: We don't actually need to sort the entire array by count, do we? We could just do a single pass through and find the highest count.

And we could build up the component names string and populate the `didWarnAboutLegacyContext` map in the same pass we compared count.

bvaughn: Oh, I see. We were sorting by name before, and now we're showing the component names sorted by their count. I guess that makes sense then. 👍

trueadm:
@gaearon I've made those changes, let me know if you're happy with them.

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -300,17 +301,23 @@ if (__DEV__) {
...
 
-    // Dedup strategy: Warn once per component.
-    if (didWarnAboutLegacyContext.has(fiber.type)) {
+    // Update legacy context counts
+    const warningCount = legacyContextCounts.get(type) || 0;
+    // Increase warning count by 1
+    legacyContextCounts.set(type, warningCount + 1);
+
+    // Dedup strategy: Warn once per component
+    if (didWarnAboutLegacyContext.has(type)) {
```
bvaughn: Seems like we should do this check before incrementing the count.

Else we might end up including this type in the warning later, if another type has a context warning too?

trueadm: Good point. Thanks :)

packages/react-reconciler/src/ReactStrictModeWarnings.js
```diff
@@ -324,16 +331,30 @@ if (__DEV__) {
...
+
+        fiberArray
+          .sort((a, b) => {
+            const aCount = legacyContextCounts.get(a.type) || 0;
+            const bCount = legacyContextCounts.get(b.type) || 0;
+            return bCount - aCount;
+          })
+          .forEach((fiber, i) => {
+            const type = fiber.type;
+            if (didWarnAboutLegacyContext.has(type)) {
```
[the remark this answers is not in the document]
trueadm: I feel like I don't have enough context on this and the surrounding issues that are involved. I believe you wrote the original code here, so it makes sense why you'd be able to understand the semantics involved with the roots here too.

Those points are valid, and I'm sorry that I don't properly have the context you have to have addressed this already. I know you wrote most, if not all, of the original code and I must be slowing you down. Furthermore, I feel this task is also becoming a pandora's box – this is the forth refactor and I feel like I'm even further away from tackling the original problem.

I don't have a huge amount of bandwidth right now and I feel like I'm slowing progress, so I'll close this PR and stop stepping on your shoes. Sorry.

gaearon:
https://github.com/facebook/react/pull/17922

PR #17887, opened 2020-01-21
Make disable context flag dynamic for WWW

gaearon:
I want to make a GK to turn it off. To prevent more usages from being added.


sebmarkbage:
Let's discuss this because I'm not sure of a safe way you can use this. I imagine you want to turn it on based on a particular site's flag so a particular tree can't have it, but I don't know if that actually does what you think it does.

gaearon:
I guess let's not for now

PR #17888, opened 2020-01-21
Remove dynamic GKs for selective/train

sebmarkbage:
There are shipped/shipping.


gaearon:
Do we know if anything got better as a result?

sebmarkbage:
@gaearon Marginally for SSR hydration but it turns out to not be the major bottleneck so hard to tell how much.

PR #17889, opened 2020-01-22
Remove call to finishHooks when doing a static render.

### LINKED ISSUE #16416: SSR: Cannot set property 'memoizedState' of null

A bug?

> Cannot set property 'memoizedState' of null

```js
const processLink = html => {
  return renderToStaticMarkup(<Link />)
};

const RichText = ({ html }) => {
  const htmlProcessed = useMemo(() => processLink(html), [html]);
}
```
See https://codesandbox.io/s/cannot-set-property-memoizedstate-of-null-mrxfr

16.8~16.9

ijxsid2:
## Summary
Fixes #16416 . If you had a `renderToStaticMarkup` or `renderToString` inside the hook, it would cause all the hooks to finish and set `workInProgressHook` to `null`. In this PR, I am trying to call the `finishHooks` only when we're not doing a static markup. 

## Test Plan
Its my first issue so I might have missed something. I ran `yarn test`, `yarn linc` and `yarn flow`, they are all passing. Also, I am not sure how to add a test for this.

### DIFF
--- a/packages/react-dom/src/server/ReactPartialRenderer.js
+++ b/packages/react-dom/src/server/ReactPartialRenderer.js
@@ -416,7 +416,7 @@ function resolve(
   child: mixed,
   context: Object,
   threadID: ThreadID,
-  makeStaticMarkup: boolean
+  makeStaticMarkup: boolean,
 ): {|
   child: mixed,
   context: Object,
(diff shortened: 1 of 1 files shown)


bl00mber:
Hey @ijxsid2 can you please tell how you have requested a review not being a member at the same time? :thinking:

PR #17896, opened 2020-01-23
chore: upgrade to jest 25

SimenB:
## Summary

The changes to the test code relate to changes in JSDOM that come with Jest 25:

* Several JSDOM workarounds are no longer needed.
* Several tests made assertions to match incorrect JSDOM behavior (e.g. `setAttribute` calls) that JSDOM has now patched to match browsers.
  * https://codesandbox.io/s/resets-value-of-datetime-input-to-fix-bugs-in-ios-safari-1ppwh
* JSDOM no longer triggers default actions when dispatching click events. 
  * https://codesandbox.io/s/beautiful-cdn-ugn8f
* JSDOM fixed (jsdom/jsdom#2700) a bug so that calling `focus()` on an already focused element does not dispatch a `FocusEvent`.
* JSDOM now supports passive events.
* JSDOM has improved support for custom CSS properties.
  * But requires jsdom/cssstyle#112 to land to support webkit prefixed properties.

Closes #17012

## Test Plan

`yarn test`

PR #17897, opened 2020-01-23
chore: update to latest release of babel

SimenB:
## Summary

Fixes a couple of the test failures from #17896 and reduces the diff.

Updated babel by deleting everything `@babel/*` in the lockfile and running `yarn`

(Ironically my PR to Babel which broke the test: https://github.com/babel/babel/pull/10361)

## Test Plan

Green CI


necolas:
Thanks. We should probably bump the base versions of the dependencies in the `package.json` too https://github.com/facebook/react/blob/master/package.json#L7

SimenB:
@necolas any particular reason to? They're all in `devDependencies` so not user facing at all, meaning the version in the lockfile should be all that matters.

trueadm:
@SimenB We should update the package.json `devDependencies` anyway, but also it makes sense for users who might not be using Yarn?

SimenB:
@necolas @trueadm updated

trueadm:
Looks good, thank you!

PR #17898, opened 2020-01-23
[Do not merge] Provide React Element introspection warning

trueadm:
Example of how we might warn on React Element introspection


trueadm:
I don't think it's worth doing this at this point. The performance hit on DEV is huge, the Jest tests take 20% longer to run and this is likely to translate to a significant perf issues on DEV environments outside of Jest too (getters aren't fast, and having to invoke the introspection function makes things even slower). Furthermore, the complications this adds to infra and the escape hatches involved in preventing the warning really go against the wins that we want to add with prop introspection.

PR #17901, opened 2020-01-23
Update various monorepo dependencies

necolas:
# Summary

Updates various monorepo dependencies. Will update the closure compiler dependency separately.

## Test Plan

`yarn test`


package.json
```diff
@@ -33,19 +33,19 @@
     "@babel/preset-flow": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
     "@babel/traverse": "^7.0.0",
-    "@mattiasbuelens/web-streams-polyfill": "0.1.0",
-    "art": "^0.10.1",
+    "@mattiasbuelens/web-streams-polyfill": "^0.3.2",
```
trueadm: We should be careful with polyfills, as they might cause breakages if we use them in our bundle builds.

necolas: Are you saying you want me to revert this change?

trueadm: Looks to be okay

PR #17902, opened 2020-01-23
Update google-closure-compiler

necolas:
## Summary

Updates the google-closure-compiler dependency

## Test Plan

CI


trueadm:
I'm a bit worried that upgrading GCC is now causing our production bundles to fail tests. Maybe we should find out what the cause might be and report the issue back to GCC.

necolas:
Passing now. The previous test failures were a subset of the same tests that failed during the jest upgrade.

gaearon:
Thanks!

PR #17903, opened 2020-01-23
Use CommonJS outside of src

gaearon:
This might help fix the issue @trueadm was seeing when upgrading Jest.

### DIFF
--- a/packages/react-cache/index.js
+++ b/packages/react-cache/index.js
@@ -9,4 +9,6 @@
 
 'use strict';
 
-export * from './src/ReactCache';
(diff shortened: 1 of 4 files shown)


gaearon:
I guess this breaks the build. :-(

@trueadm If you prepare a PR with Jest upgrade, I can look at it.

acdlite:
Hmm is this because we don't run Babel on those files? Curious why this changed between Jest versions

necolas:
https://github.com/facebook/react/pull/17896

trueadm:
This isn't needed anymore. Thanks for the help anyway!

PR #17904, opened 2020-01-24
[WIP] Upgrade to Jest 25

trueadm:
This PR upgrades us to the latest version of Jest. Unfortunately, we run into issues doing this due to the fact newer versions of JSDOM require `testURL` to be `http://localhost`. Historically, this has been `about:blank` for our tests, because we depend on behavior differences in JSDOM that ensure our existing tests work as expected and changing the `testURL` (as I've done in this PR) now causes many of our tests to fail. As per the comment in our `config.base.js` by @SimenB :

```
  // Jest changed from `about:blank` to `http://localhost` default in 24.5 (https://github.com/facebook/jest/pull/6792)
  // in order to address https://github.com/facebook/jest/issues/6766. If one uses `about:blank` in JSDOM@11.12 or
  // newer, it fails with `SecurityError: localStorage is not available for opaque origins`. However, some of React's
  // tests depend on `about:blank` being the domain (for e.g. `url` in `img` tags). So we set `about:blank` here to
  // keep the current behavior and make sure to keep the version of JSDOM to version lower than 11.12. This will have
  // to be addressed properly when Jest 25 is released, as it will come with a newer version of JSDOM.
```

I'm unsure as to how we can resolve this. Maybe @SimenB has some ideas?

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "google-closure-compiler": "20190301.0.0",
     "gzip-size": "^3.0.0",
     "jasmine-check": "^1.0.0-rc.0",
-    "jest": "^24.9.0",
+    "jest": "^25.1.0",
     "jest-diff": "^24.9.0",
     "jest-snapshot-serializer-raw": "^1.1.0",
     "minimatch": "^3.0.4",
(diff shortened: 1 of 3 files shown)


trueadm:
For some reason I completely missed https://github.com/facebook/react/pull/17896. I'll close this PR in favor of that PR.

PR #17906, opened 2020-01-24
[react-interactions] use dom-event-testing-library for Drag tests

trueadm:
@necolas wrote an awesome testing library for DOM events, we should probably use it for the drag tests too. The PR replaces the hacky MouseEvents logic that was there before and makes the tests easier to read and more consistent with other responder tests. Hopefully it also helps with https://github.com/facebook/react/pull/17896.


necolas:
I removed these modules in https://github.com/facebook/react/pull/17907

PR #17908, opened 2020-01-24
Rewrite useTransition to better handle overlapping transitions

acdlite:
When the user interacts with the UI more quickly than it can respond, you usually don't want to update the screen with anything that isn't the last thing the user requested.

For example, if you click on multiple tabs in, we should not switch to any tab that isn't the last one you clicked. The last tab navigation supersedes all the previous ones.

However, this only applies to transitions that update the same part of the UI. For example, if you click a new tab, and in the meantime you click a drop-down menu, it doesn't matter whether the tab or the menu appears first. They are independent, parallel transitions.

To implement this behavior, we will treat transitions as **overlapping** if they share at least one state queue between them. The result of the most recent transition in an overlapping sequence is considered the **terminal** state. The rest are considered **intermediate** states. We will avoid showing intermediate states by batching them together with the terminal one.

When there are overlapping transitions, the `isPending` states of the `useTransition` hooks that correspond to the intermediate states are turned off (unless they happen to be the same `useTransition` hook as the terminal one).

---

I iterated on the implementation several times. I've squashed most of the commits to reduce noise when reviewing. You can see some of the unsquashed steps in https://github.com/facebook/react/pull/17418. If this is too much to review in a single PR, I can split the commits back up again; however, I don't think it's worth landing in more than two steps because there are only two useful features here. Increasing the granularity would increase the risk of landing without much benefit.

## Example (demo)

The example demo is a tab switcher. Pay attention to when the pending spinner disappears, and when the tab content is allowed to switch.

Before this PR: https://codesandbox.io/s/elastic-hawking-69381
After this PR: https://codesandbox.io/s/beautiful-goldstine-u157u

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -114,9 +114,9 @@ import {
   expirationTimeToMs,
   computeInteractiveExpiration,
   computeAsyncExpiration,
-  computeSuspenseExpiration,
-  inferPriorityFromExpirationTime,
+  computeSuspenseTimeout,
   LOW_PRIORITY_EXPIRATION,
+  inferPriorityFromExpirationTime,
   Batched,
   Idle,
 } from './ReactFiberExpirationTime';
@@ -232,7 +232,7 @@ let workInProgressRootFatalError: mixed = null;
 // This is conceptually a time stamp but expressed in terms of an ExpirationTime
 // because we deal mostly with expiration times in the hot path, so this avoids
 // the conversion happening in the hot path.
-let workInProgressRootLatestProcessedExpirationTime: ExpirationTime = Sync;
+let workInProgressRootLatestProcessedEventTime: ExpirationTime = Sync;
 let workInProgressRootLatestSuspenseTimeout: ExpirationTime = Sync;
 let workInProgressRootCanSuspendUsingConfig: null | SuspenseConfig = null;
 // The work left over by components that were visited during this render. Only
@@ -331,11 +331,9 @@ export function computeExpirationForFiber(
 
   let expirationTime;
   if (suspenseConfig !== null) {
-    // Compute an expiration time based on the Suspense timeout.
-    expirationTime = computeSuspenseExpiration(
-      currentTime,
-      suspenseConfig.timeoutMs | 0 || LOW_PRIORITY_EXPIRATION,
-    );
+    // If there's a SuspenseConfig, treat this as a concurrent update regardless
+    // of the priority.
+    expirationTime = computeAsyncExpiration(currentTime);
   } else {
     // Compute an expiration time based on the Scheduler priority.
     switch (priorityLevel) {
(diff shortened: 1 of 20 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1356,23 +1341,30 @@ export function markCommitTimeOfFallback() {
...
+    // Track the largest/latest timeout deadline in this batch.
+    // TODO: If there are two transitions in the same batch, shouldn't we
+    // choose the smaller one? Maybe this is because when an intermediate
+    // transition is superseded, we should ignore its suspense config, but
+    // we don't currently.
+    if (suspenseConfig !== null) {
+      // If `timeoutMs` is not specified, we default to 5 seconds.
+      // TODO: Should this default to a JND instead?
+      const timeoutMs = suspenseConfig.timeoutMs | 0 || LOW_PRIORITY_EXPIRATION;
+      const timeoutTime = computeSuspenseTimeout(eventTime, timeoutMs);
```
acdlite: Move default timeout to when the update is scheduled

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -331,11 +331,9 @@ export function computeExpirationForFiber(
...
   let expirationTime;
   if (suspenseConfig !== null) {
-    // Compute an expiration time based on the Suspense timeout.
-    expirationTime = computeSuspenseExpiration(
-      currentTime,
-      suspenseConfig.timeoutMs | 0 || LOW_PRIORITY_EXPIRATION,
-    );
+    // If there's a SuspenseConfig, treat this as a concurrent update regardless
+    // of the priority.
+    expirationTime = computeAsyncExpiration(currentTime);
```
acdlite: Split into two "lanes" depending on how large the timeout is.

- Smallish numbers get currentTime + 10sec
- "Infinite" numbers get really big number

acdlite:
## To-do

- [ ] Go back to using `useState` hook for `isPending` queue
- [ ] Use map of `intermediateTransitionTime -> terminalTransitionTime` instead of ranges of times. If a non-transition happens to fall in between, oh well.
- [ ] Inline `resolveTransitionTime` into `getNextExpirationTimeToWorkOn`

packages/react-reconciler/src/ReactFiberWorkLoop.js
acdlite: Nevermind. The reason this is resolved here is because `timeoutMs` is stored on the Suspense config object, which is owned by userspace. In the future we can track this on the root as a map of transition -> timeout.

acdlite:
Discussed with @sebmarkbage offline. Here's the plan:

- The ideal implementation requires a refactor of expiration times. The refactor will be significant, so I originally wanted to land the semantic changes in this PR first before moving onto a refactor. But since the interim solution in this PR is vastly different from the implementation post-refactor, we're going to do the refactor first.
- The first commit (decoupling expiration times and timeouts) is still useful so we should land that now. I'll split this into a new PR so this one can stay as a reference.

PR #17910, opened 2020-01-25
Implemented Profiler onCommit() and onPostCommit() hooks

bvaughn:
This PR adds two new callbacks to the Profiler, `onCommit` and `onPostCommit`, as proposed in [RFC 139](https://github.com/bvaughn/rfcs/blob/profiler-commit-durations/text/0000-profiler-measure-commit-durations.md) (https://github.com/reactjs/rfcs/pull/139).

- [x] Implement RFC.
- [ ] Add DevTools hook for capturing passive effect durations.

**When reviewing, use the [`?w=1`](https://github.com/facebook/react/pull/17910/files?w=1) feature flag to ignore whitespace**

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -29,6 +29,7 @@ import {
   deferPassiveEffectCleanupDuringUnmount,
   enableSchedulerTracing,
   enableProfilerTimer,
+  enableProfilerCommitHooks,
   enableSuspenseServerRenderer,
   enableDeprecatedFlareAPI,
   enableFundamentalAPI,
@@ -183,7 +184,11 @@ const callComponentWillUnmountWithTimer = function(current, instance) {
   startPhaseTimer(current, 'componentWillUnmount');
   instance.props = current.memoizedProps;
   instance.state = current.memoizedState;
-  if (enableProfilerTimer && current.mode & ProfileMode) {
+  if (
+    enableProfilerTimer &&
+    enableProfilerCommitHooks &&
+    current.mode & ProfileMode
+  ) {
     try {
       startLayoutEffectTimer();
       instance.componentWillUnmount();
@@ -447,7 +452,11 @@ export function commitPassiveHookEffects(finishedWork: Fiber): void {
         // TODO (#17945) We should call all passive destroy functions (for all fibers)
         // before calling any create functions. The current approach only serializes
         // these for a single fiber.
-        if (enableProfilerTimer && finishedWork.mode & ProfileMode) {
+        if (
+          enableProfilerTimer &&
+          enableProfilerCommitHooks &&
+          finishedWork.mode & ProfileMode
+        ) {
           try {
             startPassiveEffectTimer();
             commitHookEffectListUnmount(
@@ -480,7 +489,7 @@ export function commitPassiveEffectDurations(
   finishedRoot: FiberRoot,
   finishedWork: Fiber,
 ): void {
-  if (enableProfilerTimer) {
+  if (enableProfilerTimer && enableProfilerCommitHooks) {
     // Only Profilers with work in their subtree will have an Update effect scheduled.
     if ((finishedWork.effectTag & Update) !== NoEffect) {
       switch (finishedWork.tag) {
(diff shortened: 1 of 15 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -613,9 +719,74 @@ function commitLifeCycles(
...
+
+        // In the common case, we don't call onPostCommit() here.
+        // Typically, passive effects haven't been measured yet,
+        // so any durations we have recorded will be for the previous commit's effects.
+        // However if we have recorded passive effect durations,
+        // it means cleanup functions were run as part of an unmount.
+        // In that event we should to call onPostCommit() and bubble durations.
+        //
+        // TODO (bvaughn, acdlite) We plan to refactor passive effect cleanup timing as part of the CM umbrella.
+        // Hopefully we can remove this codepath after that refactor?
```
bvaughn: Do we have a task that I can link this TODO to? I didn't see one in the CM umbrella doc.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -772,16 +943,50 @@ function commitUnmount(
...
+            runWithPriority(priorityLevel, () => {
+              let effect = firstEffect;
+              do {
+                const destroy = effect.destroy;
+                if (destroy !== undefined) {
+                  // In order to support the Profiler's onCommit and onPostCommit API,
+                  // we need to attribute time spent in cleanup functions to the right effects group.
+                  // Unfortunately for effects with no changed dependencies, this will be NoHookEffect.
+                  //
+                  // TODO (bvaughn, acdlite) Revisit this when we refactor passive effects and unmount.
```
bvaughn: This section is kind of gross. I welcome suggestions for improving it!

In `updateEffectImpl`, we tag hooks with equal dependencies as `NoWork`:
https://github.com/facebook/react/blob/cf0081263ced42ddebe0a0d868701890448bfd4b/packages/react-reconciler/src/ReactFiberHooks.js#L939-L942

I think this makes sense in the common case, but it complicates the `onCommit` / `onPostCommit` implementation because I can't know which group to attribute time to when those effects are later being unmounted:
https://github.com/facebook/react/blob/cf0081263ced42ddebe0a0d868701890448bfd4b/packages/react-reconciler/src/ReactFiberCommitWork.js#L780

The cost of not being able to differentiate between the two is just that we'll end up not attributing some of the time spent in cleanup functions. This is probably not a dealbreaker for the API though.

bvaughn:
Ping 😄

bvaughn:
I have rebased this on top of PR #17947 to pull in the various passive effect fixes.

bvaughn:
Yes. I've been working on this off and on all afternoon. The recent passive effects changes require changing some of the interaction tracing tests (which I'm working on). Should have it ready to review by tomorrow morning.

bvaughn:
@sebmarkbage rebase complete.

Changes to the profiler tests look larger than they are, because I replaced all of the `onRender` callbacks in the interaction-tracing tests with `onPostCommit` callbacks (to make sure we didn't have any pending interaction traces still waiting to be completed).

I also updated tests to run with the deferred passive effects flag on and off, so be sure to use the `?w=1` query param when looking at the diff.

packages/react-reconciler/src/ReactProfilerTimer.js
```diff
@@ -77,9 +80,78 @@ function stopProfilerTimerIfRunningAndRecordDelta(
...
+    const elapsedTime = now() - passiveEffectStartTime;
+
+    passiveEffectStartTime = -1;
+
+    // Store duration on the next nearest Profiler ancestor.
+    let parentFiber = fiber.return;
+    while (parentFiber !== null) {
+      if (parentFiber.tag === Profiler) {
+        const parentStateNode = parentFiber.stateNode;
+        if (parentStateNode !== null) {
```
bvaughn: This check is something that we wouldn't need to do if we eagerly flushed passive destroy functions during unmount.

packages/react/src/__tests__/ReactProfiler-test.internal.js
```diff
@@ -182,1084 +183,2282 @@ describe('Profiler', () => {
     });
   });
 
-  [true, false].forEach(enableSchedulerTracing => {
-    describe('onRender callback', () => {
-      beforeEach(() => {
-        jest.resetModules();
+  [true, false].forEach(deferPassiveEffectCleanupDuringUnmount => {
```
bvaughn: This method of testing feature flags is controversial. I think it's nice to have the extra coverage, and these tests are already very verbose so the idea of duplicating them is not very appealing- but I don't want to die on this hill. I'll remove this and just test one of the flag settings if that's the team preference.

bvaughn:
Note to self: Add an additional feature flag for the commit hooks so we can turn them off if it proves to have a negative performance impact.

bvaughn:
FYI @sebmarkbage I've added a new feature flag (off by default) called `enableProfilerCommitHooks`. The new Profiler APIs are behind this feature flag.

PR #17916, opened 2020-01-27
Removed Root API callback params and added warnings

bvaughn:
The goal is to prevent introducing new uses of the update queue callback behavior (so that we can deprecate it later, as part of cleaning up legacy class component usage), Fortunately it looks like we weren't using these params internally anywhere, so I've just added tests that verify the new warning behavior.

I based the wording of this warning on a similar warning for hooks:
https://github.com/facebook/react/blob/cf0081263ced42ddebe0a0d868701890448bfd4b/packages/react-reconciler/src/ReactFiberHooks.js#L1265-L1278


bvaughn:
Cool. Happy to tweak it later.

PR #17919, opened 2020-01-28
Add semver dependency explicitly to DevTools

gaearon:
I couldn't get DevTools running on master.

Turns out, there was a `semver` dependency [used](https://github.com/facebook/react/blob/b7f217d40c6c8c973c86fabceeda238b38662171/packages/react-devtools-shared/src/backend/renderer.js#L10) but not specified in `package.json`. It worked accidentally, but `semver` 7.x appears incompatible with webpack for whatever reason.

So I'm pinning it back to 6.x.

Test plan: I can compile DevTools now.


gaearon:
I'll just get it in.

PR #17925, opened 2020-01-29
Flush `useEffect` clean up functions in the passive effects phase

bvaughn:
This PR cleans up our hook effect tags and also fixes the case where we were too eagerly flushing passive effect destroy functions on unmount. (As a side effect, we are also now reliably able to distinguish between passive and layout effects during unmount, which will improve the new Profiler methods added in #17910.)

The two major changes in this PR are:
* 9a38b12: Defer passive destroy functions during an unmount until other passive effects are flushed.
* 1c6be34: Cleanup effect tags.

This is a change in behavior/timing for destroy functions so it may cause code to break unexpectedly. For this reason, it has been added behind a killswitch feature flag (`deferPassiveEffectCleanupDuringUnmount`).

I've also identified some follow up work for passive effects, but to keep this PR small and easy to review, I've just added a TODO for this and will be tracking it with a separate issue, #17945.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -325,21 +325,14 @@ function commitBeforeMutationLifeCycles(
   );
 }
 
-function commitHookEffectList(
-  unmountTag: number,
-  mountTag: number,
-  finishedWork: Fiber,
-) {
+function commitHookEffectListUnmount(tag: number, finishedWork: Fiber) {
   const updateQueue: FunctionComponentUpdateQueue | null = (finishedWork.updateQueue: any);
   let lastEffect = updateQueue !== null ? updateQueue.lastEffect : null;
   if (lastEffect !== null) {
     const firstEffect = lastEffect.next;
     let effect = firstEffect;
     do {
-      if (
-        (effect.tag & HookHasEffect) !== NoHookEffect &&
-        (effect.tag & unmountTag) !== NoHookEffect
-      ) {
+      if ((effect.tag & tag) === tag) {
         // Unmount
         const destroy = effect.destroy;
         effect.destroy = undefined;
@@ -347,10 +340,19 @@ function commitHookEffectList(
           destroy();
         }
       }
-      if (
-        (effect.tag & HookHasEffect) !== NoHookEffect &&
-        (effect.tag & mountTag) !== NoHookEffect
-      ) {
+      effect = effect.next;
+    } while (effect !== firstEffect);
+  }
+}
+
+function commitHookEffectListMount(tag: number, finishedWork: Fiber) {
+  const updateQueue: FunctionComponentUpdateQueue | null = (finishedWork.updateQueue: any);
+  let lastEffect = updateQueue !== null ? updateQueue.lastEffect : null;
+  if (lastEffect !== null) {
+    const firstEffect = lastEffect.next;
+    let effect = firstEffect;
+    do {
+      if ((effect.tag & tag) === tag) {
         // Mount
         const create = effect.create;
         effect.destroy = create();
@@ -404,8 +406,8 @@ export function commitPassiveHookEffects(finishedWork: Fiber): void {
         // TODO (#17945) We should call all passive destroy functions (for all fibers)
         // before calling any create functions. The current approach only serializes
         // these for a single fiber.
-        commitHookEffectList(HookPassive, NoHookEffect, finishedWork);
-        commitHookEffectList(NoHookEffect, HookPassive, finishedWork);
+        commitHookEffectListUnmount(HookPassive | HookHasEffect, finishedWork);
+        commitHookEffectListMount(HookPassive | HookHasEffect, finishedWork);
         break;
       }
       default:
@@ -429,7 +431,7 @@ function commitLifeCycles(
       // This is done to prevent sibling component effects from interfering with each other,
       // e.g. a destroy function in one component should never override a ref set
       // by a create function in another component during the same commit.
-      commitHookEffectList(NoHookEffect, HookLayout, finishedWork);
+      commitHookEffectListMount(HookLayout | HookHasEffect, finishedWork);
       return;
     }
     case ClassComponent: {
@@ -1315,7 +1317,7 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
         // This prevents sibling component effects from interfering with each other,
         // e.g. a destroy function in one component should never override a ref set
         // by a create function in another component during the same commit.
-        commitHookEffectList(HookLayout, NoHookEffect, finishedWork);
+        commitHookEffectListUnmount(HookLayout | HookHasEffect, finishedWork);
         return;
       }
       case Profiler: {
@@ -1358,7 +1360,7 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
       // This prevents sibling component effects from interfering with each other,
       // e.g. a destroy function in one component should never override a ref set
       // by a create function in another component during the same commit.
-      commitHookEffectList(HookLayout, NoHookEffect, finishedWork);
+      commitHookEffectListUnmount(HookLayout | HookHasEffect, finishedWork);
       return;
     }
     case ClassComponent: {
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -26,6 +26,7 @@ import type {ReactPriorityLevel} from './SchedulerWithReactIntegration';
 
 import {unstable_wrap as Schedule_tracing_wrap} from 'scheduler/tracing';
 import {
+  deferPassiveEffectCleanupDuringUnmount,
   enableSchedulerTracing,
   enableProfilerTimer,
   enableSuspenseServerRenderer,
@@ -109,9 +110,11 @@ import {
   captureCommitPhaseError,
   resolveRetryThenable,
   markCommitTimeOfFallback,
+  enqueuePendingPassiveEffectDestroyFn,
 } from './ReactFiberWorkLoop';
 import {
   NoEffect as NoHookEffect,
+  NoEffectPassiveUnmountFiber,
   UnmountSnapshot,
   UnmountMutation,
   MountMutation,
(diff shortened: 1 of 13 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -756,32 +759,54 @@ function commitUnmount(
...
+                    safelyCallDestroy(current, destroy);
+                  }
+                } else {
+                  safelyCallDestroy(current, destroy);
+                }
               }
               effect = effect.next;
             } while (effect !== firstEffect);
-          });
+          } else {
```
bvaughn: This is the old code path, in case we enable the kill switch.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1285,8 +1310,10 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
       case MemoComponent:
       case SimpleMemoComponent:
       case Chunk: {
-        // Note: We currently never use MountMutation, but useLayout uses
-        // UnmountMutation.
+        // Note: We currently never use MountMutation, but useLayoutEffect uses UnmountMutation.
+        // This is to ensure ALL destroy fns are run before create fns,
+        // without requiring us to traverse the effects list an extra time during commit.
+        // This sequence prevents sibling destroy and create fns from interfering with each other.
```
bvaughn: I added these comments for myself, since it wasn't clear to me *why* we were using these tags.

packages/react-reconciler/src/ReactHookEffectTags.js
```diff
@@ -9,11 +9,12 @@
...
-export const UnmountPassive = /*       */ 0b10000000;
+export const NoEffect = /*                   */ 0b000000000;
+export const UnmountSnapshot = /*            */ 0b000000010;
+export const UnmountMutation = /*            */ 0b000000100;
+export const MountMutation = /*              */ 0b000001000;
+export const UnmountLayout = /*              */ 0b000010000;
+export const MountLayout = /*                */ 0b000100000;
+export const MountPassive = /*               */ 0b001000000;
+export const UnmountPassive = /*             */ 0b010000000;
+export const NoEffectPassiveUnmountFiber = /**/ 0b100000000;
```
bvaughn: The name of the new effect tag I added, `NoEffectPassiveUnmountFiber`, is kind of stupid. I'd welcome a solution for a better name. 🤪

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -926,7 +927,13 @@ function mountEffectImpl(fiberEffectTag, hookEffectTag, create, deps): void {
   hook.memoizedState = pushEffect(hookEffectTag, create, undefined, nextDeps);
 }
 
-function updateEffectImpl(fiberEffectTag, hookEffectTag, create, deps): void {
+function updateEffectImpl(
+  fiberEffectTag,
+  hookEffectTag,
+  noWorkUnmountFiberHookEffectTag,
```
bvaughn: This name isn't the greatest either. I welcome better suggestions.

The purpose of the changes to this function is to retain the ability to distinguish between layout and passive effects even if a render has no dependency changes. Previously, we would not be able to distinguish between the two during unmount.

packages/react-reconciler/src/ReactHookEffectTags.js
acdlite: I think it should be possible to simplify these tags by removing the Mount/Unmount . I think that's already handled because of `commitHookEffectList` accepts two separate arguments for which tags to mount and which tags to unmount.

Would it work if the only tags we had were this?

```js
// Represents whether effect should fire.
export const HasEffect = /*  */ 0b0001;
// Represents the phase in which the effect (not the clean-up) fires.
export const Snapshot = /*   */ 0b0010;
export const Layout = /*     */ 0b0100;
export const Passive = /*    */ 0b1000;
```

So during an update you would fire the clean up of any hook that has (for passive effects) that has `Passive | HasEffect`. When the fiber is deleted, you would only check `Passive`.

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -972,6 +973,92 @@ describe('ReactHooksWithNoopRenderer', () => {
...
+        expect(Scheduler).toFlushAndYield([
+          'passive create (no dependencies)',
+          'passive create',
+        ]);
+      });
+
+      // HACK
+      // This update is kind of gross since it exists to test an internal implementation detail:
+      // Effects without updating dependencies lose their layout/passive tag during an update.
+      // A special type of no-update tag (NoEffectPassiveUnmountFiber) is used to track these for later.
```
acdlite: To make it less "hacky" you could include the count in the log message. So that the effect actually depends on it.

Then make the other effect close over a different prop. `count2` or whatever.

Then during the update, you change one of the props but not the other. Only that effect should fire; the other one will bail out.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2176,6 +2178,19 @@ export function flushPassiveEffects() {
   }
 }
 
+export function enqueuePendingPassiveEffectDestroyFn(
+  destroy: () => void,
+): void {
+  if (deferPassiveEffectCleanupDuringUnmount) {
+    pendingUnmountedPassiveEffectDestroyFunctions.push(destroy);
+    rootDoesHavePassiveEffects = true;
+    scheduleCallback(NormalPriority, () => {
```
acdlite: This will schedule a callback for every clean-up effect. Need to check if `rootDoesHavePassiveEffects` is already true.

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
bvaughn: Yeah, sure. That"hacky" part is that I'm only doing this intermediate update because I *know* that used to cause the specific implementation to blur the passive and layout effects together.

packages/react-reconciler/src/ReactFiberWorkLoop.js
bvaughn: Ah, good call!

packages/react-reconciler/src/ReactHookEffectTags.js
bvaughn: I think this cleanup suggestion makes sense, although while looking into it I realized we have a potential bug with the sequence/timing of passive effects. (We don't call all destroy functions before create functions.) I'll take a pass at the effect tag cleanup *and* separating the passive effect destroy/create phases too. If it starts looking like that might be a bigger change though, I may split it out into a follow up PR.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -250,7 +248,6 @@ function commitBeforeMutationLifeCycles(
     case ForwardRef:
     case SimpleMemoComponent:
     case Chunk: {
-      commitHookEffectList(UnmountSnapshot, NoHookEffect, finishedWork);
```
bvaughn: We don't have a snapshot phase hook (yet) so this call was walking the effect list for no reason. I've removed it.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2193,6 +2210,20 @@ function flushPassiveEffectsImpl() {
...
+  if (deferPassiveEffectCleanupDuringUnmount) {
+    // Flush any pending passive effect destroy functions that belong to
+    // components that were unmounted during the most recent commit.
+    for (
+      let i = 0;
+      i < pendingUnmountedPassiveEffectDestroyFunctions.length;
+      i++
+    ) {
+      const destroy = pendingUnmountedPassiveEffectDestroyFunctions[i];
+      invokeGuardedCallback(null, destroy, null);
```
bvaughn: This loop approach has an unintentional side effect of being "better" than our normal destroy approach, since an error thrown by one destroy function won't prevent other destroy functions *on the same Fiber* from being called.

acdlite: Did you confirm that with a test? I would expect the current behavior to also unmount all the destroy functions on a fiber, even if one throws, because if one of them throws, then the fiber gets deleted, which triggers `commitNestedUnmount`, which then calls the remaining clean-up functions and wraps each call in a try catch: https://github.com/facebook/react/blob/38cd75861f44a40e686a39403bff39bf16fd796d/packages/react-reconciler/src/ReactFiberCommitWork.js#L780

bvaughn: Ah, yes I think you are right. Disregard!

packages/shared/ReactFeatureFlags.js
```diff
@@ -91,6 +91,12 @@ export const enableTrustedTypesIntegration = false;
 // Flag to turn event.target and event.currentTarget in ReactNative from a reactTag to a component instance
 export const enableNativeTargetAsInstance = false;
 
+// Controls behavior of deferred effect destroy functions during unmount.
+// Previously these functions were run during commit (along with layout effects).
+// Ideally we should delay these until after commit for performance reasons.
+// This flag provides a killswitch if that proves to break existing code somehow.
+export const deferPassiveEffectCleanupDuringUnmount = true;
```
bvaughn: Open question: should I set this to `false` for now (for open source releases)?

acdlite: Yeah probably, just to be safe

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -339,15 +336,21 @@ function commitHookEffectList(
...
         const destroy = effect.destroy;
         effect.destroy = undefined;
         if (destroy !== undefined) {
           destroy();
         }
       }
-      if ((effect.tag & mountTag) !== NoHookEffect) {
+      if (
+        (effect.tag & HookHasEffect) !== NoHookEffect &&
+        (effect.tag & mountTag) !== NoHookEffect
```
acdlite: Nit: You can save a check here if you assume `mountTag` and `unmountTag` already include `HookHasEffect`:

```js
(effect.tag & mountTag) === mountTag
```

packages/shared/ReactFeatureFlags.js
bvaughn: K, will do then.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -398,8 +401,11 @@ export function commitPassiveHookEffects(finishedWork: Fiber): void {
       case ForwardRef:
       case SimpleMemoComponent:
       case Chunk: {
-        commitHookEffectList(UnmountPassive, NoHookEffect, finishedWork);
-        commitHookEffectList(NoHookEffect, MountPassive, finishedWork);
+        // TODO (#17945) We should call all passive destroy functions (for all fibers)
+        // before calling any create functions. The current approach only serializes
+        // these for a single fiber.
+        commitHookEffectList(HookPassive, NoHookEffect, finishedWork);
+        commitHookEffectList(NoHookEffect, HookPassive, finishedWork);
```
acdlite: Following from previous comment, here you would combine the tag with `HookHasEffect` before passing them in:

```js
commitHookEffectList(HookPassive | HookHasEffect, NoHookEffect, finishedWork);
commitHookEffectList(NoHookEffect, HookPassive | HookHasEffect, finishedWork);
```

(I assume Closure inlines those values, or you could also hoist them out to module scope)

packages/react-reconciler/src/ReactFiberCommitWork.js
bvaughn: Hmm! But they specifically don't.

We use the tag to specify the *type* of effect even if there's no work currently required by the hook. This lets us properly differentiate between passive and layout effects in the unmount case. (`NoHookEffect` is used to signal that there's *also* work for that specific hook during the current commit.)

bvaughn: Ah, I see your comment below and I understand what you meant now. :smile: Sure I can make that change!

bvaughn: I think pre-combing would cause false positives.

This would fail for a layout tag:
```js
let tag = Layout | HasEffect;

tag & HasEffect && tag & Passive // false
tag & HasEffect | Passive // true
```

bvaughn: Please excuse my previous comment. I gotcha.

PR #17933, opened 2020-01-30
Fix Event Replaying in Flare by Eagerly Adding Active Listeners 

sebmarkbage:
This effectively reverts part of #17513. I duplicated the event replaying tests to now include Flare too, just like the partial hydration tests already did. The existing tests only covered passive events.

The primary problem is that if there is no active listener attached, then the `IS_ACTIVE` flag isn't passed into `DEPRECATED_dispatchEventForResponderEventSystem`. So the replaying of the events, like Press, that normally listen to active events doesn't work.

Replaying doesn't technically need to be active for all events. It's not guaranteed to always work to preventDefault if the code hasn't loaded yet. We could make the listeners attached as passive in the DOM but still pass the `IS_ACTIVE` flag. That would still "work" but it wouldn't actually work to prevent default.

The problem is that we do use a best-effort attempt to allow `preventDefault` to work. Most of the time it just works because we are able to synchronously hydrate within the event. That's also how we preventDefault on links to prevent the browser from taking over.

So I think ideally, at least _some_ event should still be attached as active such as `submit`, `click` and the `key` events. We don't have to make them all active. Notably the ones that can cause problems for scroll optimizations and such are the pointer events.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerSelectiveHydration-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerSelectiveHydration-test.internal.js
@@ -9,11 +9,14 @@
 
 'use strict';
 
+import {createEventTarget} from 'dom-event-testing-library';
+
 let React;
 let ReactDOM;
 let ReactDOMServer;
 let Scheduler;
 let Suspense;
+let usePress;
 
 function dispatchMouseHoverEvent(to, from) {
   if (!to) {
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
```diff
@@ -1888,7 +1888,7 @@ describe('ReactDOMServerPartialHydration', () => {
     const TestResponder = React.DEPRECATED_createResponder(
       'TestEventResponder',
       {
-        targetEventTypes: ['click'],
```
sebmarkbage: This test didn't catch it because apparently this means that it's listening to *both* active and passive. With the fix, now it actually fires twice which I think is a bug.

trueadm:
This all makes sense. I removed the eager event listeners before when maybe that was not right. We probably need to listen to both for Flare, but once we switch to useEvent this issue should go away.

PR #17934, opened 2020-01-30
Add Component Highlighting to Profiler DevTools

### LINKED ISSUE #17928: Profiler should highlight host components (e.g. DOM elements) on mouseover
Feature request from a DevTools user at Faceook:
> If I mouse over a node in the flame graph if you can highlight it in the view like the inspector that would be really amazing!

M-Izadmehr:
## Summary: Fix #17928 

Inside the devTools (Components) we have component highlighting, and by hovering the mouse on components it will highlight them in DOM. However, this feature was missing from `Profiler` (both FlameGraph Chart and Ranked Chart). The purpose of this PR is to add this functionality to the profiler.

## Demo
devtools


packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -115,9 +118,36 @@ function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
...
+  );
+
+  // Highlight last hovered element.
+  const handleElementMouseEnter = useCallback(
+    id => {
+      highlightNativeElement(id);
+    },
+    [highlightNativeElement],
+  );
+
```
strdr4605: This section is also repeated in `packages/react-devtools-shared/src/devtools/views/Profiler/CommitRanked.js` and also 
[packages/react-devtools-shared/src/devtools/views/Components/Tree.js](https://github.com/facebook/react/blob/e706721490e50d0bd6af2cd933dbf857fd8b61ed/packages/react-devtools-shared/src/devtools/views/Components/Tree.js).

Maybe better to create a separate hook file?

M-Izadmehr: Thanks for the comment, Yeah I absolutely agree,  I will try to use a hook for the highlighting functionality:)

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -115,9 +118,36 @@ function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
...
+          rendererID,
+          scrollIntoView: false,
+        });
+      }
+    },
+    [store, bridge],
+  );
+
+  // Highlight last hovered element.
+  const handleElementMouseEnter = useCallback(
```
strdr4605: @M-Izadmehr, can you please explain why you wrap in another `useCallback`?

bvaughn:
I'm going to hold off doing a review for now since it sounds like there are planned changes. :smile:

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
M-Izadmehr: Done:)
I removed the useless `useCallback`.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
M-Izadmehr: Done, now I extracted the highlighting logic to the hooks file, and created a hook called `useNativeElementHighlighter`, this hook exposes two methods: 
1. `highlightNativeElement`: method used to highlight an element using its id
2. `clearNativeElementHighlight`: method used to clear highlighting of elements

M-Izadmehr:
@bvaughn As a response to the reviews I made the following changes:
1.  I extracted the highlighting logic to the hooks file, and created a hook called `useNativeElementHighlighter`, which exposes two methods (`highlightNativeElement` and `clearNativeElementHighlight`).
2. Added clear highlighting when you leave the element
3. Added a logic when leaving an element, if another element is already selected use highlighting on that (see the gif)

ezgif-updated

bvaughn:
Might be worth profiling the profiler (in the test shell) before and after this change to see if we added any extra renders or otherwise did anything that would cause a regression in the flame chart views.

I'll review this sometime in the next couple of days though~

bvaughn:
> Might be worth profiling the profiler (in the test shell) before and after this change to see if we added any extra renders or otherwise did anything that would cause a regression in the flame chart views.

Any chance you were able to do this? :grin:

bvaughn:
I did a quick test. I don't see any big perf regressions, but I do see a UX problem: Highlights always get stuck. Looks like anytime I'm not moused over the root flamegraph node, its DOM element is highlighted - and mousing out of the graph doesn't clear the highlight either, leaving it stuck on the page.

M-Izadmehr:
Thanks for the comment :)
Yeah, I think that was a bad decision on my side because I thought it might be a good idea if a fiber is already selected and the user is not hovering an element, highlight the clicked fiber. However, it results in some bad implications:
1. If the user redirects to a new page (where the fiber does not exist anymore) the highlight still remains on the page.
2. Also, in this way, we should keep track of valid fibers
As a result, I removed that unnecessary logic (ao that logic will be similar to components devtools).

packages/react-devtools-shared/src/devtools/views/hooks.js
```diff
@@ -300,3 +302,29 @@ export function useSubscription<Value>({
 
   return state.value;
 }
+
+// provides highlighting and clearing highlights of elements based on ID
+export function useNativeElementHighlighter() {
+  const store = useContext(StoreContext);
+  const bridge = useContext(BridgeContext);
+
+  const highlightNativeElement = (id: number) => {
```
bvaughn: The previous version of `highlightNativeElement` was memoized with `useCallback`. This one isn't, which would cause the effect it's passed to in the `Tree` component to be run every render.

strdr4605: In this case, it probably is good to wrap it in a `useCallback` as before. 

But I saw several places where `useCallback` is wrapping simple functions like 
```js
const handleBlur = useCallback(() => setTreeFocused(false), []);
``` 
From [When to useMemo and useCallback](https://kentcdodds.com/blog/usememo-and-usecallback) article I understand that in these cases it doesn't give better performance. 
I want to know your opinion on this topic, to understand if I should use `useCallback` in such cases.

M-Izadmehr: Interesting,
I was wondering if it is possible to use profiler devTools while debugging profiler devTools? :D

I did a small study:
* used both codes (with `useCallback` and without `useCallback`) on a real project
* Then tried to highlight different elements by moving the mouse randomly
* Put `console.time/console.timeEnd` before calling `highlightNativeElement` and after sending to bridge.
* Gathered 800 data points and averaged them

And both cases almost took a similar amount of time on my Mac:
1. **with `useCallback`:** 0.01529866435s** (Average of 800 highlights)
1. **without `useCallback`:** 0.01466493762s** (Average of 800 highlights)

The results basically even show that using `useCallback` has a minor negative (or even negligible to be honest) effect on performance. I guess the trade-off of using `useCallback` is almost equivalent to the calculation done in the function. 

Either way, there is a flaw in my case study, or as a usual practice we prefer to use `useCallback` on all/most methods in the project, please let me know, I will add it to the method:))

M-Izadmehr:
Interesting that similar PR happened in `react-flame-graph`, I think the adding tooltips with useful information can be a great idea.

bvaughn:
Agreed that tooltips might be a nice addition. Could perhaps also help with feature request like #18014 - a better way to see info at a glance without having to click and zoom in/out.

Would that be something that you'd be interested in taking a shot at, @M-Izadmehr ? :)

packages/react-devtools-shared/src/devtools/views/hooks.js
bvaughn: > I was wondering if it is possible to use profiler devTools while debugging profiler devTools? 

Yes. You can use the DevTools extension to debug/profile the test shell (which runs an inline version of the DevTools).

My concern was that it would break memoization and cause more/unnecessary rendering. I'm not sure I understand how the console timing you describe above measures this. I think you would need to use a profiler to measure it.

Let me take another look sometime later this morning.

M-Izadmehr:
@bvaughn, Great, that emoji suggestion is also a nice idea,
Of course, it is something that personally helps me while debugging, I will try to add the tooltip one to this PR:)

bvaughn:
It's okay. Lets' leave this PR as is (once I find time to review it) and do that as a separate one? :)

M-Izadmehr:
Sounds like a plan:) 👍

packages/react-devtools-shared/src/devtools/views/hooks.js
strdr4605: > My concern was that it would break memoization and cause more/unnecessary rendering.

I think that even if with `useCallback` it takes more time to evaluate between `console.time/console.timeEnd` the purpose is to no rerender. But still, this means that we should use `useCallback` with every function that we pass to children (which sounds strange). 
So probably the best way is to first pass function without `useCallback` and if there are performance issues then optimize. 

As I see in [Tree.js](https://github.com/facebook/react/blob/master/packages/react-devtools-shared/src/devtools/views/Components/Tree.js) every child prop function is wrapped with `useCallback`. 
So interesting question is: What is the performance without `useCallback`? Is it good enough to not `useCallback`?

bvaughn: There are some unnecessary `useCallback` wrappers in this repo. DevTools was my the first app I wrote that used hooks, so there are some mistakes :smile:

In this particular case, `highlightNativeElement` *used to be a stable function*. Because it gets passed through to `itemData`, it seemed preferable for it to remained stable-ish to avoid unnecessarily re-rendering a lot of chart nodes which are memoized: https://github.com/facebook/react/blob/90be006da8e231279151e7b1518bb64c62e26851/packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraphListItem.js#L118

Because `store` and `bridge` are stable, callbacks that only use those values should be stable too.

I'm not sure how important this turns out to be in practice. This is why I had suggested React profiling :smile:

M-Izadmehr:
Thanks for the comments, 
I think, as two pull requests will have minor conflicts with each other as both are working on hovering the fiber and adding some custom event listeners to mouse, I think it might be better to first merge one of them and afterward I will apply the reviews to this one (as some of the changes are already fixed in the other one).

packages/react-devtools-shared/src/devtools/views/hooks.js
M-Izadmehr: Interesting, that makes sense:)
About this case that we were discussing and also other use cases of using the `useCallback` and memoizations, I can try to go through the code, and do some additional checks to see how these changes affect the number of updates on the page and also probably the performance.  Would it be something that we are interested in?

bvaughn:
Sounds good @M-Izadmehr. The other PR has been merged!

bvaughn:
@M-Izadmehr Still planning to update this PR? Or would you like me to finish it up?

M-Izadmehr:
@bvaughn Sorry last week was super busy with some releases, I actually wanted to do it in the weekend, does it sound good?:)

bvaughn:
Sure sounds fine.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -115,9 +122,21 @@ function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
     return null;
   }, [chartData, selectedFiberID, selectedChartNodeIndex]);
 
+  // Highlight last hovered element.
+  const handleElementMouseEnter = id => {
```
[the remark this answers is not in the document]
M-Izadmehr: Done.
In order to preserve the consistency (for mouse handlers, we seem to always use the name of mouse in the method), I used the second suggestion (`const handleMouseLeave = clearNativeElementHighlight`)

packages/react-devtools-shared/src/devtools/views/Components/Tree.js
```diff
@@ -269,9 +256,9 @@ export default function Tree(props: Props) {
     setIsNavigatingWithKeyboard(false);
   }, []);
 
-  const handleMouseLeave = useCallback(() => {
-    bridge.send('clearNativeElementHighlight');
-  }, [bridge]);
+  const handleMouseLeave = () => {
```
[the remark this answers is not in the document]
M-Izadmehr: Done.
In order to preserve the consistency (for mouse handlers, we seem to always use the name of the mouse in the method), I used the second suggestion (`const handleMouseLeave = clearNativeElementHighlight`)

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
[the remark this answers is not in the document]
M-Izadmehr: After merging the other PR, this method now does two actions:
1. Clearing the highlighting,
2. Clearing the `fiberData` used for the tooltip,
So,  I guess we can keep the wrapper

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -115,9 +122,21 @@ function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
...
+    highlightNativeElement(id);
+  };
+
+  // remove highlighting of element on mouse leave
+  const handleElementMouseLeave = () => {
```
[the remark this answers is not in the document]
M-Izadmehr: After merging the other PR, this method now does two actions:

Clearing the highlighting,
Clearing the fiberData used for the tooltip,
So, I guess we can keep the wrapper

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRanked.js
```diff
@@ -91,15 +94,31 @@ type Props = {|
...
+    clearNativeElementHighlight,
+  } = useNativeElementHighlighter();
 
   const selectedFiberIndex = useMemo(
     () => getNodeIndex(chartData, selectedFiberID),
     [chartData, selectedFiberID],
   );
 
+  // Highlight last hovered element.
+  const handleElementMouseEnter = id => {
```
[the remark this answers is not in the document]
M-Izadmehr: Same as above comment:)

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRanked.js
```diff
@@ -91,15 +94,31 @@ type Props = {|
...
+    highlightNativeElement(id);
+  };
+
+  // remove highlighting of element on mouse leave
+  const handleElementMouseLeave = () => {
```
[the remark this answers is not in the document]
M-Izadmehr: Also the same, I guess we can pull these two lines out into a hook, but I also have the feeling that it will be overkill to add a custom hook just for two lines, any suggestion?

packages/react-devtools-shared/src/devtools/views/hooks.js
M-Izadmehr: About this one, I checked it out, there is no additional re-renders fortunately, as you mentioned store and bridge are stable. So I think it makes sense to keep highlighter without `useCallback`

M-Izadmehr:
Sorry for the late update, I merged the other stuff to this PR, now it seems stable, I don't see any additional re-renders:)

Here is a demo with both tooltip and highlighting:
ezgif-7-3bd435cc231a

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -96,9 +99,13 @@ type Props = {|
 |};
 
 function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
-  const [hoveredFiberData, hoverFiber] = useState<number | null>(null);
+  const [hoveredFiberData, setHoveredFiberData] = useState<number | null>(null);
```
bvaughn: This Flow type looks wrong. (Looks like it's been wrong for a while though FWIW- which is a bit concerning.)

We aren't storing a `number` in this state, but an object with `id` and `name` props (`TooltipFiberData`).

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -121,10 +128,22 @@ function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
...
+  const handleElementMouseLeave = () => {
+    clearNativeElementHighlight(); // clear highlighting of element on mouse leave
+    setHoveredFiberData(null); // clear hovered fiber data for tooltip
+  };
+
   const itemData = useMemo<ItemData>(
     () => ({
       chartData,
-      hoverFiber,
+      isHovered: hoveredFiberData && hoveredFiberData.id,
```
bvaughn: Why is this check not just `hoveredFiberData !== null`?

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -140,7 +159,7 @@ function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
     }),
     [
       chartData,
-      hoverFiber,
+      setHoveredFiberData,
```
bvaughn: This looks wrong. The dependencies array has `setHoveredFiberData ` but the memoized value references `hoveredFiberData`.

I wonder why our lint rule isn't complaining about this.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -121,10 +128,22 @@ function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
...
+    setHoveredFiberData(null); // clear hovered fiber data for tooltip
+  };
+
   const itemData = useMemo<ItemData>(
     () => ({
       chartData,
-      hoverFiber,
+      isHovered: hoveredFiberData && hoveredFiberData.id,
+      onElementMouseEnter: handleElementMouseEnter,
+      onElementMouseLeave: handleElementMouseLeave,
```
bvaughn: This looks wrong. Neither `handleElementMouseEnter` nor `handleElementMouseLeave` are mentioned in the dependencies array.

I wonder why our lint rule isn't complaining about this.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRanked.js
```diff
@@ -94,19 +97,35 @@ type Props = {|
 |};
 
 function CommitRanked({chartData, commitTree, height, width}: Props) {
-  const [hoveredFiberData, hoverFiber] = useState<number | null>(null);
+  const [hoveredFiberData, setHoveredFiberData] = useState<number | null>(null);
```
bvaughn: Same comment as above wrt hook state type.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRanked.js
```diff
@@ -94,19 +97,35 @@ type Props = {|
...
+    setHoveredFiberData(null); // clear hovered fiber data for tooltip
+  };
+
   const itemData = useMemo<ItemData>(
     () => ({
       chartData,
-      hoverFiber,
+      isHovered: hoveredFiberData && hoveredFiberData.id,
+      onElementMouseEnter: handleElementMouseEnter,
+      onElementMouseLeave: handleElementMouseLeave,
```
bvaughn: Same comment as above wrt missing dependencies.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRanked.js
```diff
@@ -94,19 +97,35 @@ type Props = {|
...
+  const handleElementMouseLeave = () => {
+    clearNativeElementHighlight(); // clear highlighting of element on mouse leave
+    setHoveredFiberData(null); // clear hovered fiber data for tooltip
+  };
+
   const itemData = useMemo<ItemData>(
     () => ({
       chartData,
-      hoverFiber,
+      isHovered: hoveredFiberData && hoveredFiberData.id,
```
bvaughn: Same comment as above wrt `hoveredFiberData !== null`

bvaughn:
Some follow up from my review yesterday.

Turns out our Flow coverage was really broken for DevTools. Oops! Fixed with #18204.

The "rules of hooks" lint rule isn't being run on this repo, but should be. I'll look to add it with a separate PR.

M-Izadmehr:
Oh I see, I can also try to check it out if we are using it on all the packages

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
M-Izadmehr: Good catch

M-Izadmehr:
@bvaughn Yeah, you are right, we can check if the element is imported, I think in the tooltip we already have information on that part (only if we are listening for changes for that), we can re-use the logic there for the highlighter too.
One simple way without performance reduction is to only enable highlighter, only when `why rendered` is true.
I will play around with it to see if there is any optimal solution for this:)

gaearon:
@M-Izadmehr Would be great if we could get this landed. :-)

M-Izadmehr:
@gaearon Yeah, please don't close this one, I was really busy in the past month and couldn't come back to this issue, I will try to pick it up in the weekend (thanks to quarantine, we have so much free time now:) )

bvaughn:
Hey @M-Izadmehr, PR #18479 is is going to cause a bit of a conflict for you b'c it also separates the highlight behavior into a hook. Looks like the hook is the same as yours though, so it should be easy for you to rebase and just use it instead. Just wanted to give you a heads up!

gaearon:
@M-Izadmehr Friendly ping :-) Do you think you might be able to get this over the finish line?

bl00mber:
Hey @M-Izadmehr can you please explain how you managed to request a review and not be a member at the same time? :thinking:

M-Izadmehr:
@bl00mber when you are finished with your stuff, you can ask @bvaughn to check it out (as it is related to DevTools, and he is the DevTools god:) )

bvaughn:
@bl00mber I've reviewed several of your PRs in the past month or so :smile: 

It's important to keep in mind that there are only a half dozen or so React core members trying to keep up with PRs from > 1,000 contributors (while at the same time creating our own as well).

bl00mber:
@bvaughn yeah thanks, I've actually do not wanted to request a review _now_ (may be only on stalled PRs), was just curious how the author of this PR is able to do it.

PR #17942, opened 2020-01-30
StrictMode should call sCU twice in DEV

bvaughn:
This was an oversight before. Even our docs suggested that sCU should be called twice:
https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects


bvaughn:
> What is the difference between assigning and requesting a review? How can you “assign” a PR to someone? Odd, github.

It's odd, huh? I guess the idea is that you can pick multiple reviewers, then one fo them could self-assign.

I'm trying to use both fields to see if it helps get PRs faster :)

PR #17949, opened 2020-01-31
Bugfix: Expired partial tree infinite loops

acdlite:
Adds regression tests that reproduce a scenario where a partially completed tree expired then fell into an infinite loop.

The code change that exposed this bug made the assumption that if you call Scheduler's `shouldYield` from inside an expired task, Scheduler will always return `false`. But in fact, Scheduler sometimes returns `true` in that scenario, which is a bug.

The reason it worked before is that once a task timed out, React would always switch to a synchronous work loop without checking `shouldYield`.

My rationale for relying on `shouldYield` was to unify the code paths between a partially concurrent render (i.e. expires midway through) and a fully concurrent render, as opposed to a render that was synchronous the whole time. However, this bug indicates that we need a stronger guarantee within React for when tasks expire, given that the failure case is so catastrophic. Instead of relying on the result of a dynamic method call, we should use control flow to guarantee that the work is synchronously executed.

(We should also fix the Scheduler bug so that `shouldYield` always returns false inside an expired task, but I'll address that separately.)

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -645,7 +645,6 @@ function performConcurrentWorkOnRoot(root, didTimeout) {
   currentEventTime = NoWork;
 
   // Check if the render expired.
-  const expirationTime = getNextRootExpirationTimeToWorkOn(root);
   if (didTimeout) {
     // The render task took too long to complete. Mark the current time as
     // expired to synchronously render all expired work in a single batch.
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -644,26 +644,28 @@ function performConcurrentWorkOnRoot(root, didTimeout) {
...
-    markRootExpiredAtTime(root, currentTime);
+  // Check if the render expired.
+  const expirationTime = getNextRootExpirationTimeToWorkOn(root);
+  if (didTimeout) {
+    if (workInProgressRoot === root && renderExpirationTime !== NoWork) {
+      // We're already in the middle of working on this tree. Expire the tree at
+      // the already-rendering time so we don't throw out the partial work. If
+      // there's another expired level after that, we'll hit the `else` branch
+      // of this condition, which will batch together the remaining levels.
+      markRootExpiredAtTime(root, renderExpirationTime);
```
sebmarkbage: It seems a bit like this make the expired value lie. If we added more places where we check the current time, we would update this to something further out. It's also not clear to me whether when we're actually going to set its to its real value and stop temporarily lying.

Instead of doing this here. Can we just let it set it to the lower expired value like we did before? Then instead check in performSyncWorkOnRoot if it has expired work but the expired work is lower priority then the currently rendering work, first do a pass finishing the currently rendering work and then do a second pass at the expired priority and then possibly a third pass if it errors in the second pass?

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberWorkLoop.js#L992

acdlite:
@sebmarkbage Based on your feedback I updated the PR in two commits. The first one moves the partial tree check to `performSyncWorkOnRoot`. That one is pretty simple by itself.

The second one removes `didError` by performing multiple render passes within the same call to `performSyncWorkOnRoot` or `performConcurrentWorkOnRoot`. To do this, I factored out the render phase. I think this refactor makes a lot of sense, though it's inherently riskier.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1493,6 +1440,53 @@ function workLoopSync() {
   }
 }
 
+function renderRootConcurrent(root, expirationTime) {
```
sebmarkbage: Just to be clear, this is only called once so it didn't technically need to be extracted out, right?

acdlite: Yeah I only did it for symmetry purposes

PR #17952, opened 2020-02-01
Simplify Continuous Hydration Targets

sebmarkbage:
Let’s use a constant priority for this. This helps us avoid restarting a render when switching targets and simplifies the model.

The downside is that now we're not down-prioritizing the previous hover target. However, we think that's ok because it'll only do one level too much and then stop.

I also add test meant to show why it's tricky to merge both hydration levels. Having both levels co-exist works. However, if we deprioritize hydration using a single level, we might deprioritize the wrong thing.

It also tests that we don't down-prioritize if we're changing the hover in the middle of doing continuous priority work. We don't do the -1 trick for these and so we don't have to consider that in the new model. Just that we shouldn't start doing that.


acdlite:
Thanks!

PR #17967, opened 2020-02-03
[Mock Scheduler] Mimic browser's advanceTime

acdlite:
The mock Scheduler that we use in our tests has its own fake timer implementation. The `unstable_advanceTime` method advances the timeline.

Currently, a call to `unstable_advanceTime` will also flush any pending expired work. But that's not how it works in the browser: when a timer fires, the corresponding task is added to the Scheduler queue. However, we will still wait until the next message event before flushing it.

This commit changes `unstable_advanceTime` to more closely resemble the browser behavior, by removing the automatic flushing of expired work.

```js
// Before this commit
Scheduler.unstable_advanceTime(ms);

// Equivalent behavior after this commit
Scheduler.unstable_advanceTime(ms);
Scheduler.unstable_flushExpired();
```

The general principle is to prefer separate APIs for scheduling tasks and flushing them.

This change does not affect any public APIs. `unstable_advanceTime` is only used by our own test suite. It is not used by `act`.

However, we may need to update tests in www, like Relay's.

### DIFF
--- a/packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
+++ b/packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
@@ -779,6 +779,7 @@ describe('ChangeEventPlugin', () => {
 
           // 3s should be enough to expire the updates
           Scheduler.unstable_advanceTime(3000);
+          expect(Scheduler).toFlushExpired([]);
           expect(container.textContent).toEqual('hovered');
         });
       },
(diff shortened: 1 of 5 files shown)


sebmarkbage:
I think I'd prefer a helper that plays time forward for a certain time with a reasonable amount of flushes in between. It shouldn't just jump to the end as if the CPU was stalled for the whole time. Seems like we'd test very unrealistic scenarios that way. The order might matter.

bvaughn:
I dig @sebmarkbage's suggestion of a method like `toAdvanceTimeByAndToFlushAndYield(number, Array<string>)` to avoid the case of forgetting to flush.

acdlite:
Yeah I think that API would make sense because it's a test matcher which discourages us from calling it inside another task.

PR #17968, opened 2020-02-04
[Scheduler] shouldYield should always return false from inside an expired task

acdlite:
## Based on #17967

The current behavior of shouldYield is unspecified except in cases that happen to be relied on by the React work loop's implementation. However, some of the unspecified cases are important and the consequences of assuming the wrong behavior are really bad.

For example, it's natural to assume that `shouldYield` returns false when called from within an expired task. But the current behavior usually does that, unless there's a higher priority task or a main thread task (i.e. at the end of the 5ms message loop interval).

This fixes `shouldYield` so that it always returns false inside an expired task, regardless of the other tasks in the queue.

### DIFF
--- a/packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
+++ b/packages/react-dom/src/events/__tests__/ChangeEventPlugin-test.internal.js
@@ -779,6 +779,7 @@ describe('ChangeEventPlugin', () => {
 
           // 3s should be enough to expire the updates
           Scheduler.unstable_advanceTime(3000);
+          expect(Scheduler).toFlushExpired([]);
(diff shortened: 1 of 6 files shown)


sebmarkbage:
It seems to me that shouldYield is already way above its budget for such a hot function. I'd expect a single atomic read basically.

It also seems to me that we're getting rid of the notion of "expiration time" in the scheduler since we won't have a timeoutMs option and the native one won't work this way anyway so this doesn't translate.

I think we should just simplify this check to only yield every 5ms regardless if higher pri tasks are scheduled or anything else.

gaearon:
Closing as stale.

PR #17975, opened 2020-02-04
fix sizebot - point correctly to circleci artifact

threepointone:
similar to #17972, this should fix sizebot not reporting stats.


threepointone:
Cool, it works

PR #17977, opened 2020-02-05
Portals trigger Suspense boundaries in SSR and only render on client

lunaruan:
Currently during SSR, rendering a Portal throws an error. However, in Concurrent Mode, if a Portal is wrapped in a Suspense boundary, instead of throwing an err, we want to trigger the closest Suspense fallback instead. Then during hydration, we can render the Portal. If a Portal isn't wrapped in a Suspense boundary, we will still throw an error to remain consistent with the previous implementation.


lunaruan:
Wrong branch

PR #17978, opened 2020-02-05
Portals trigger Suspense boundaries in SSR and only render on client

lunaruan:
Currently during SSR, rendering a Portal throws an error. However, in Concurrent Mode, if a Portal is wrapped in a Suspense boundary, instead of throwing an err, we want to trigger the closest Suspense fallback instead. Then during hydration, we can render the Portal. If a Portal isn't wrapped in a Suspense boundary, we will still throw an error to remain consistent with the previous implementation.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
@@ -78,6 +78,7 @@ describe('ReactDOMServerPartialHydration', () => {
     ReactFeatureFlags.enableSuspenseCallback = true;
     ReactFeatureFlags.enableDeprecatedFlareAPI = true;
     ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
+    ReactFeatureFlags.enableSuspenseServerRenderer = true;
 
     React = require('react');
     ReactDOM = require('react-dom');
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
```diff
@@ -2563,4 +2564,49 @@ describe('ReactDOMServerPartialHydration', () => {
...
+    }
+
+    const markup = ReactDOMServer.renderToString(<App />);
+    expect(markup).toContain('Loading...');
+    expect(portalContainer.textContent).toBe('');
+
+    const container = document.createElement('div');
+    container.innerHTML = markup;
+
+    // Hydrating this cames the fallback to be removed and the Portal to be added.
```
threepointone:
```suggestion
    // Hydrating this removes the fallback and adds the Portal.
```

packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
```diff
@@ -2563,4 +2564,49 @@ describe('ReactDOMServerPartialHydration', () => {
...
+    function App() {
+      return (
+        <div>
+          {ReactDOM.createPortal(<div>{'Portal'}</div>, portalContainer)}
+        </div>
+      );
+    }
+
+    expect(() => ReactDOMServer.renderToString(<App />)).toThrow(
+      'Portals must have a fallback UI when being rendered on the server.\n\n' +
```
threepointone:
```suggestion
      'Portals must have a fallback UI when rendered on the server.\n\n' +
```

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -900,16 +900,29 @@ class ReactDOMServerRenderer {
...
-                // TODO: include component name. This is a bit tricky with current factoring.
-                'A React component suspended while rendering, but no fallback UI was specified.\n' +
-                  '\n' +
-                  'Add a <Suspense fallback=...> component higher in the tree to ' +
-                  'provide a loading indicator or placeholder to display.',
-              );
+              if (err === REACT_PORTAL_TYPE) {
+                invariant(
+                  this.suspenseDepth > 0,
+                  'Portals must have a fallback UI when being rendered on the server.\n' +
```
threepointone:
```suggestion
                  'Portals must have a fallback UI when rendered on the server.\n' +
```

packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
```diff
@@ -2563,4 +2564,49 @@ describe('ReactDOMServerPartialHydration', () => {
     // Now we're hydrated.
     expect(ref.current).not.toBe(null);
   });
+
+  it('should return fallback during server rendering when Portal has a suspense boundary ', () => {
+    const portalContainer = document.createElement('div');
```
threepointone: In a typical SSR setup, `document` won't be available, so `document.createElement` will throw. 

So it begs the question: what does a component that has a `createPortal` call looks like in an SSR setup? Because it won't look like these tests.

lunaruan: Good point! Will have to rethink this. Thanks for the feedback.

threepointone: @sema can createPortal accept null if document.createElement doesn’t exist? The partial renderer won’t even need it. In the browser, we could throw if an element isn’t passed.

sebmarkbage: It could but we currently don’t encourage any uses of “is server” conditionals since it’s easy to make mistakes that cause hydration issues as a result.

I was under the impression that we did something that rendered portals internally for SSR. But I think what we actually do is throw before we even get to it.

I think that’s probably the strategy here. Have something that throws before we even get to the code that needs to create a DOM node.

We just don’t have an API to do that.

PR #17915, opened 2020-01-27
Testing builds for React/ReactDOM

threepointone:
This PR introduces adds `react/testing` and `react-dom/testing`.
- changes infra to generate these builds
- exports on `ReactDOM.act` in these testing builds
- uses the new test builds in `fixtures/dom`

In the next PR -

- I'll use the new builds for all our own tests
- I'll replace usages of `TestUtils.act` with `ReactDOM.act`.

PR #17985, opened 2020-02-06
Use testing builds for our own tests

threepointone:
Following up from #17915 where we generated testing builds for `react-dom`, this PR uses those builds for our own tests.

- fixes for fb builds: 
  - adds a ReactFeatureFlags fork where `isTestEnvironment:true`
  - adds a `.fb` fork for the `testing.js` entrypoint
- adds `isTestEnvironment: true` to `ReactFeatureFlags.readonly`
- with jest, mocks `react-dom` with the testing version, both for source and builds
- uses `ReactDOM.act` in one of these tests to verify it actually works

This also meant I had to add `testing.js` to `react-dom`'s `package.json` (so we can still run `yarn test-build`). If you feel strongly about not exposing this yet, maybe I could rename it to `testing_DO_NOT_USE.js`?

I'll write a post before I land this, but this setup has an annoyance. If you add a new feature flag, and run your tests, your test will probably fail, since it's likely you've not added it to the `.testing` forks. There's nothing to warn this except for flow. which might not be running. Open to ideas for syncing these forks.

### DIFF
--- a/packages/react-dom/package.json
+++ b/packages/react-dom/package.json
@@ -30,6 +30,7 @@
     "README.md",
     "build-info.json",
     "index.js",
+    "testing.js",
     "profiling.js",
     "server.js",
     "server.browser.js",
--- a/packages/shared/forks/ReactFeatureFlags.readonly.js
+++ b/packages/shared/forks/ReactFeatureFlags.readonly.js
@@ -9,4 +9,4 @@
 // It lets us determine whether we're running in Fire mode without making tests internal.
 const ReactFeatureFlags = require('../ReactFeatureFlags');
 // Forbid writes because this wouldn't work with bundle tests.
-module.exports = Object.freeze({...ReactFeatureFlags});
+module.exports = Object.freeze({...ReactFeatureFlags, isTestEnvironment: true});
--- a/fixtures/dom/src/__tests__/nested-act-test.js
+++ b/fixtures/dom/src/__tests__/nested-act-test.js
@@ -13,10 +13,7 @@ let TestRenderer;
 
 global.__DEV__ = process.env.NODE_ENV !== 'production';
 
-jest.mock('react-dom', () =>
-  require.requireActual('react-dom/cjs/react-dom-testing.development.js')
-);
-// we'll replace the above with react/testing and react-dom/testing right before the next minor
+jest.mock('react-dom', () => require.requireActual('react-dom/testing'));
 
 expect.extend(require('../toWarnDev'));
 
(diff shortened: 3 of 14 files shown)


threepointone:
I’m worried the testing feature flags could go out of sync with their base files. Open to ideas to mitigate that.

scripts/rollup/build.js
```diff
@@ -347,6 +347,7 @@ function getPlugins(
     bundleType === RN_FB_DEV ||
     bundleType === RN_FB_PROD ||
     bundleType === RN_FB_PROFILING;
+  const isTestBuild = entry === 'react-dom/testing';
```
acdlite: Shouldn't this also be true for React Noop and React Test Renderer?

threepointone: I thought we'd delay it until we actually used __TEST__ in any of its code paths. Happy to add it right now tho.

threepointone: This isn't necessary anymore, since the renderers read the feature flag now

threepointone:
Updated the PR and description. Copy pasting a specific concern - 

I'll write a post before I land this, but this setup has an annoyance. If you add a new feature flag, and run your tests, your test will probably fail, since it's likely you've not added it to the `.testing` forks. There's nothing to warn this except for flow. which might not be running. Open to ideas for syncing these forks.

acdlite:
You can reexport one from the other:

```js
// shared/forks/ReactFeatureFlags.www.testing.js

export {
  flags,
  that,
  are,
  the,
  same
} from './ReactFeatureFlags.www';

export const differentFlag = 'forked';
```

threepointone:
I tried this and it failed because it kept referring to itself and failing the build. Maybe I didn’t try it hard enough, I’ll try again.

packages/react-dom/package.json
```diff
@@ -30,6 +30,7 @@
     "README.md",
     "build-info.json",
     "index.js",
+    "testing.js",
```
acdlite: Putting this here means it will go into the next public release. If you can't get rid of it because the bundle tests rely on it, maybe prefix with `unstable`?

packages/shared/forks/ReactFeatureFlags.readonly.js
```diff
@@ -9,4 +9,4 @@
 // It lets us determine whether we're running in Fire mode without making tests internal.
 const ReactFeatureFlags = require('../ReactFeatureFlags');
 // Forbid writes because this wouldn't work with bundle tests.
-module.exports = Object.freeze({...ReactFeatureFlags});
+module.exports = Object.freeze({...ReactFeatureFlags, isTestEnvironment: true});
```
acdlite: Don't need this anymore, right?

acdlite:
> I tried this and it failed because it kept referring to itself and failing the build.

You can move the default feature flags to their own file, `shared/forks/ReactFeatureFlags.default`. Then `shared/ReactFeatureFlags` can re-export that. And so can all the forks.

threepointone:
abandoning this for https://github.com/facebook/react/pull/18196

PR #17990, opened 2020-02-06
Add feature flag for removing children Map support

trueadm:
This adds a flag to disables support for Map as children (resulting in an invariant instead) and updates the existing warning.


packages/react/src/ReactChildren.js
```diff
@@ -158,14 +159,22 @@ function traverseAllChildrenImpl(
   } else {
     const iteratorFn = getIteratorFn(children);
     if (typeof iteratorFn === 'function') {
-      if (__DEV__) {
+      if (iteratorFn === children.entries) {
```
gaearon: After this diff, this check runs for everyone even outside of DEV. Can we move it inside the condition?

PR #12163, opened 2018-02-06
Determine event target in Shadow DOM correctly

marionebl:
## Motivation

Fixes an issue with events not being delegated correctly when originating from an element inside a Shadow DOM root as described in #9242. 

## Included changes

* Add handling of `Events` that are `.composed` as per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/composed)
* Add basic test case for `getEventTarget`
* Add fixture in `fixtures/dom`, as `attachShadow` is not supported in `jsdom` yet

PR #17991, opened 2020-02-06
Event not working when the react app is rendered inside the web component issue fix.

LevinsDurai:
Hi Team,

Im levins and a big fan of react. I'm facing some css problem while using react in embeddable widget which can be embedded in websites and the same as chrome extension as well. We have often faced CSS overriding issue, hence I have opted to choose shadowDOM to sort this out. Since react events are binding globally, I have faced a problem in accessing react events inside shadowDOM. After digging react repo, I have figured out the getEventTarget() and added the supported changes.

This will be helpful for extending the support for ours and which uses shadowDOM or v8 engine based browsers. 
Supported Browsers:
Almost all the Browsers using V8 engine. Even Edge was moving their browsers to V8 engine ( https://www.polymer-project.org/blog/2020-01-15-edge-79-release) .
)
Thanks and regards,
Levins.

### DIFF
--- a/packages/react-dom/src/events/getEventTarget.js
+++ b/packages/react-dom/src/events/getEventTarget.js
@@ -21,7 +21,7 @@ function getEventTarget(nativeEvent) {
   // This is for solve the webcomponent event not working issue. And it will works on all the browsers which are all support the shadowDom.
   // https://github.com/facebook/react/issues/9242
   let targetPath = nativeEvent.path || (nativeEvent.composedPath && nativeEvent.composedPath());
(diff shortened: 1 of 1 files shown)


LevinsDurai:
Here is the workable sample => https://oxbj4.csb.app/

trueadm:
The CI is failing, so looks like you might need to run prettier. Also please could you add a test case or at least a Fiddle where this change works compared to before?

Thank you :)

gaearon:
I’m really not sure about this solution. Please see the existing discussion in https://github.com/facebook/react/pull/12163. How does it address those concerns?

PR #17992, opened 2020-02-06
Fix Flow type for React Native

gaearon:
I changed the external type in https://github.com/facebook/react/pull/17805 but my change was not backwards-compatible. This rolls my change back, and instead fixes the internal type to better agree with the external type.


gaearon:
On a meta note, I have no idea how the dark magic with two different classes being compatible works, but I'm not sure I want to know.

PR #17996, opened 2020-02-07
Refactor commitPlacement to recursively insert nodes

trueadm:
This PR refactors the logice in `commitPlacement` so that we instead recursively traverse the nodes as we insert them. I noticed that `commitPlacement` was a hot function during the commit phase, and this should improve the runtime performance slightly, as we no longer need to mutate the fields on the fibers to point to different return pointers. Interestingly, this also improves the Flow coverage, which was `any` before, this has been refined and fixed.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1099,44 +1099,45 @@ function commitPlacement(finishedWork: Fiber): void {
   const before = getHostSibling(finishedWork);
   // We only have the top Fiber that was inserted but we need to recurse down its
   // children to find all the terminal nodes.
-  let node: Fiber = finishedWork;
-  while (true) {
-    const isHost = node.tag === HostComponent || node.tag === HostText;
-    if (isHost || (enableFundamentalAPI && node.tag === FundamentalComponent)) {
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1099,44 +1099,45 @@ function commitPlacement(finishedWork: Fiber): void {
...
-          insertBefore(parent, stateNode, before);
-        }
+  insertOrAppendPlacementNode(finishedWork, before, parent, isContainer);
+}
+
+function insertOrAppendPlacementNode(
+  node: Fiber,
+  before: ?Instance,
+  parent: Instance,
+  isContainer: boolean,
```
acdlite: This boolean never changes, so you can fork this function: `insertOrAppendPlacementNode` and `insertOrAppendPlacementNodeIntoContainer`. Then the Flow types will be correct, too.

trueadm: I thought about doing that, but was unsure about the code size increase.

trueadm: I've made the change anyway as per your suggestion :)

PR #18000, opened 2020-02-07
useMutableSource hook

bvaughn:
`useMutableSource()` enables React components to **safely** and **efficiently** read from a mutable external source in Concurrent Mode. The API will detect mutations that occur during a render to avoid tearing and it will automatically schedule updates when the source is mutated.

**RFC**: https://github.com/reactjs/rfcs/pull/147


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -375,10 +380,7 @@ export function computeExpirationForFiber(
   return expirationTime;
 }
 
-export function scheduleUpdateOnFiber(
-  fiber: Fiber,
-  expirationTime: ExpirationTime,
-) {
+function scheduleUpdateOnFiber(fiber: Fiber, expirationTime: ExpirationTime) {
```
bvaughn: This export was never referenced (only the `scheduleWork` alias below) so I removed it.

bvaughn:
This PR is ready for review. I've updated it to reflect the latest thinking about the API.

bvaughn:
I do not understand the Circle CI `lint_build` error. I'm tempted to say it's a CI problem and not related to this PR though.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +860,181 @@ function rerenderReducer<S, I, A>(
...
+    const suspenseConfig = requestCurrentSuspenseConfig();
+    const expirationTime = computeExpirationForFiber(
+      currentTime,
+      currentlyRenderingFiber,
+      suspenseConfig,
+    );
+
+    isSafeToReadFromSource =
+      pendingExpirationTime === NoWork ||
+      pendingExpirationTime >= expirationTime;
```
bvaughn: **Note to self**: If this component is newly mounting, we need to always check the version number as well, to account for the case of tearing between newly mounted components that have not yet subscribed to the store (and so may miss a tear).

I'll add a test for this (and update the code) tomorrow.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +860,181 @@ function rerenderReducer<S, I, A>(
...
+    const root = currentlyRenderingRoot;
+
+    const create = () => {
+      const scheduleUpdate = () => {
+        const alreadyScheduledExpirationTime = root.mutableSourcePendingUpdateMap.get(
+          source,
+        );
+
+        // If an update is already scheduled for this source, re-use the same priority.
+        if (alreadyScheduledExpirationTime !== undefined) {
```
acdlite: This check means that if you update at one priority, then again at a different priority, the first priority always wins. We should support multiple concurrent priorities, which is solved by switching to a state queue. (The test case I described in my other comment would cover this.)

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1261,6 +1263,8 @@ function prepareFreshStack(root, expirationTime) {
   workInProgressRootNextUnprocessedUpdateTime = NoWork;
   workInProgressRootHasPendingPing = false;
 
+  resetMutableSourceWorkInProgressVersions(isPrimaryRenderer);
```
acdlite: Let's move this to `completeWork` and `unwindWork` of HostRoot so it gets cleared when exiting the render phase, instead of when starting the next render phase that happens to follow that. (This is akin to popping context providers off the stack.)

packages/shared/ReactMutableSource.js
```diff
@@ -0,0 +1,134 @@
...
+      if (pendingExpirationTime <= expirationTime) {
+        root.mutableSourcePendingUpdateMap.delete(mutableSource);
+      }
+    },
+  );
+}
+
+export function resetWorkInProgressVersions(isPrimaryRenderer: boolean): void {
+  if (isPrimaryRenderer) {
+    workInProgressPrimarySources.forEach(mutableSource => {
```
acdlite: Nit: This is an array, not a Map or a Set, so let's use a for loop like we do elsewhere.

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -459,6 +463,15 @@ function useResponder(responder, props): ReactEventResponderListener<any, any> {
   };
 }
 
+function useMutableSource<Source, Snapshot>(
+  source: MutableSource<Source>,
+  config: MutableSourceHookConfig<Source, Snapshot>,
+): Snapshot {
+  resolveCurrentlyRenderingComponent();
+  const getSnapshot = config.getSnapshot;
+  return getSnapshot(source._source);
```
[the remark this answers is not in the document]
acdlite: I can’t think of a server use case that isn’t actually Suspense/Flight. So never mind.

packages/react-reconciler/src/ReactFiberWorkLoop.js
bvaughn: Sure.

packages/react-reconciler/src/ReactFiberHooks.js
bvaughn: Yes, this sounds right. I could avoid checking the expiration time if I already have a WIP version. Good suggestion.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -168,6 +184,10 @@ let renderExpirationTime: ExpirationTime = NoWork;
 // the work-in-progress hook.
 let currentlyRenderingFiber: Fiber = (null: any);
 
+// The work-in-progress root fiber.
+// Used by mutable source hook to detect tears from pending updates.
+let currentlyRenderingRoot: FiberRoot = (null: any);
```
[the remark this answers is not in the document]
sebmarkbage: The other functions are able to be executed mostly standalone and operate on a single Fiber. Passing the argument alone ensures that we do that.

I've been thinking about using a few synchronous passes a couple of times that don't change the workInProgressFiber in the work loop but instead calls begin/complete in its own loop. E.g. for computing exit transition states with different contexts. It would be nice to preserve that property.

With resuming or even smaller similar optimizations (like reusing completed suspended trees) it'll be very important that we don't accidentally use global state since a render of a subtree from one pass needs to be able to be reused in another. I think for gesture calculation that will also be a problem.

This whole approach breaks that rule but that's inherent to this special problem space and not general. We think that we can find a suitable workaround for this one special case.

However, in general we should never operate on the root in begin/complete. That's why I don't think we should be passing the root down through begin since it'll encourage us to use it in more places but this is the one legit case and even this one is on the edge.

Passing lots of unused variables on the stack also has perf implications.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +860,181 @@ function rerenderReducer<S, I, A>(
...
+
+  // Is it safe to read from this source during the current render?
+  let isSafeToReadFromSource = false;
+
+  if (pendingExpirationTime !== null) {
+    // If the source has pending updates, we can use the current render's expiration
+    // time to determine if it's safe to read again from the source.
+    const currentTime = requestCurrentTimeForUpdate();
+    const suspenseConfig = requestCurrentSuspenseConfig();
+    const expirationTime = computeExpirationForFiber(
```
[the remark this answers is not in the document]
bvaughn: Oops, yes that's right.

packages/shared/ReactMutableSource.js
```diff
@@ -0,0 +1,134 @@
...
+  root.mutableSourcePendingUpdateMap.forEach(
+    (pendingExpirationTime, mutableSource) => {
+      if (pendingExpirationTime <= expirationTime) {
+        root.mutableSourcePendingUpdateMap.delete(mutableSource);
+      }
+    },
+  );
+}
+
+export function resetWorkInProgressVersions(isPrimaryRenderer: boolean): void {
```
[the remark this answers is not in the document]
bvaughn: Yup good call.

packages/react-dom/src/server/ReactPartialRendererHooks.js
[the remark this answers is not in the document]
sebmarkbage: Hm. I think this is legit even in the existing server renderer. You can use a shared in memory cache that changes. E.g. another request might be invalidating it at the same time you're rendering.

I'm just not sure how to solve it since it requires a second pass. I think the best we could do is trigger Suspense boundaries and fallback to client rendering. That might be the TODO.

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
bvaughn: I didn't like passing it in all of the places either, but didn't feel comfortable adding a getter for it. (That seems to encourage other places to use it more than explicitly passing.)

I *think* Sebastian is agreeing with Andrew's suggestion but I'm not entirely sure because of:
> However, in general we should never operate on the root in begin/complete.

Pending a decision on [this comment](https://github.com/facebook/react/pull/18000#discussion_r379257652) (which I'm still thinking about), I need to access the current root.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -356,7 +356,9 @@ function commitHookEffectListMount(tag: number, finishedWork: Fiber) {
       if ((effect.tag & tag) === tag) {
         // Mount
         const create = effect.create;
-        effect.destroy = create();
+        if (typeof create === 'function') {
```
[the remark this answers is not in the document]
sebmarkbage: I don't think we should make this a variable type unnecessarily. The VM should be able to check that the "effect" object has a consistent type. The appropriate thing here is probably a bigger refactor but it's not worth introducing this discrepancy for this hook (and potentially making all other effects slower).

I'd prefer that we set this to an empty function in this implementation if we need to rather.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +860,181 @@ function rerenderReducer<S, I, A>(
...
+  const memoizedState: MutableSourceState<Source, Snapshot> = {
+    config,
+    destroy,
+    snapshot,
+    source,
+  };
+
+  hook.memoizedState = memoizedState;
+
+  if (prevSource !== source) {
```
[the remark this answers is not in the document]
bvaughn: Interesting question. My thinking had been that a change in config would really just be a change in `getSnapshot` dependencies, and not a change in `subscribe`. Sebastian's suggestion on the RFC to get rid of the wrapper `config` object and pass both callbacks directly would make this more explicit, in which case I will add a check for that here as well.

packages/react-reconciler/src/ReactFiberCommitWork.js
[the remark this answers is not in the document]
bvaughn: > I'd prefer that we set this to an empty function in this implementation if we need to rather.

Interesting. I had considered doing that but this seemed more appropriate.

I don't feel strongly about this though. Happy to make that change.

packages/shared/ReactMutableSource.js
```diff
@@ -0,0 +1,134 @@
...
+  subscribe: (source: Source, callback: Function) => () => void,
+|};
+
+// Work in progress version numbers only apply to a single render,
+// and should be reset before starting a new render.
+// This tracks which mutable sources need to be reset after a render.
+let workInProgressPrimarySources: Array<MutableSource<any>> = [];
+let workInProgressSecondarySources: Array<MutableSource<any>> = [];
+
+export function createMutableSource<Source: $NonMaybeType<mixed>>(
```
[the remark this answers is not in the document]
sebmarkbage: The use of `shared` is a last resort and not best practice. In fact, half of the files in there shouldn't be there. Including FeatureFlags that are starting to cause problems.

Even the Symbols one is causing problems because the code size duplication is quite significant. We were talking about moving them to be exports from Isomorphic instead. So I think I might do a pass trying to get rid of as much as possible from `/shared/`.

This file certainly shouldn't be in shared since `createMutableSource` and the type signatures should just move there. In terms of many files, I think we should start moving everything in isomorphic to the same file to be more careful about growing that file. It's unnecessarily big for things that can't be made lazy.

packages/react-reconciler/src/ReactFiberCommitWork.js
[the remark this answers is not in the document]
bvaughn: Just in case it's unclear, I don't think this code path is actually executed. The change was only needed currently to satisfy Flow (although that could change in the future of course).

> I'd prefer that we set this to an empty function in this implementation if we need to rather.

I considered doing that initially... Happy to make that change!

packages/shared/ReactMutableSource.js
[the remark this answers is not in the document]
bvaughn: Yeah, already moved (just not pushed yet)

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
acdlite: > Pending a decision on this comment (which I'm still thinking about), I need to access the current root.

To clarify, you'd have to access the root in that proposal, too. It would be `root.lastPendingMutationTime`.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +860,181 @@ function rerenderReducer<S, I, A>(
...
+          // Make sure reads during future renders will know there's a pending update.
+          // This will prevent a higher priority update from reading a newer version of the source,
+          // and causing a tear between that render and previous renders.
+          root.mutableSourcePendingUpdateMap.set(source, expirationTime);
+        }
+      };
+
+      // Was the source mutated between when we rendered and when we're subscribing?
+      // If so, we also need to schedule an update.
+      const maybeNewSnapshot = getSnapshot(source._source);
```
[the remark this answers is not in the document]
bvaughn: The version is global, for the whole store. It doesn't indicate a change in the snapshot value.

That being said, I could maybe optimize by only generating a new snapshot if the version has also changed. :)

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +860,181 @@ function rerenderReducer<S, I, A>(
...
+  } else {
+    // If there are no pending updates, we should still check the work-in-progress version.
+    // It's possible this source (or the part we are reading from) has just not yet been subscribed to.
+    const lastReadVersion = getWorkInProgressVersion(source, isPrimaryRenderer);
+    if (lastReadVersion === null) {
+      // This is the only case where we need to actually update the version number.
+      // A changed version number for a new source will throw and restart the render,
+      // at which point the work-in-progress version Map will be reinitialized.
+      // Once a source has been subscribed to, we use expiration time to determine safety.
+      setWorkInProgressVersion(source, version, isPrimaryRenderer);
```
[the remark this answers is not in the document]
bvaughn: > we only need to populate the work-in-progress version if this is a new mount, because updates don't rely on mutable state: they read from a queue

I don't think this reflects the case of a new source being used for an already-mounted component.

acdlite:
The lint build error is legit. I looked at the ReactFabric-dev artifact and found this:

```js
// ReactFabric-dev.js:11656
useMutableSource: function(source, config) {
  currentHookNameInDev = "useMutableSource";
  mountHookTypesDev();
  return Snapshot > config;
}
```

Looks like some sort of build error related to Flow types?

Phew that we have the lint build task! Also an argument for why we should be running more of our tests against the bundles (and therefore moving away from `.internal` test files).

packages/react-reconciler/src/ReactFiberHooks.js
bvaughn: Yes, my thinking on this was that scheduling more than one update is only valid if we eagerly read snapshot values. I think the key question is whether or not that's a positive change. If it is, then going with an update queue makes sense.

bvaughn:
> The lint build error is legit.

That's interesting! :smile: I initially assumed it was a transient issue yesterday. Just hadn't had the chance to dig into it yet today.

packages/react-reconciler/src/ReactFiberRoot.js
```diff
@@ -74,6 +75,9 @@ type BaseFiberRootProperties = {|
   // render again
   lastPingedTime: ExpirationTime,
   lastExpiredTime: ExpirationTime,
+  // Used by useMutableSource hook to avoid tearing within this root
+  // when external, mutable sources are read from during render.
+  mutableSourcePendingUpdateMap: MutableSourcePendingUpdateMap,
```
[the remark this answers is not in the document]
bvaughn: I don't object to making this change if it's more consistent with precedent.

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
acdlite: Yeah sorry, by "mount" I mean either a new hook instance or whenever the source or config changes.

packages/react-dom/src/server/ReactPartialRendererHooks.js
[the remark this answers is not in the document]
bvaughn: I'll leave a TODO for now.

packages/react-reconciler/src/ReactFiberWorkLoop.js
bvaughn: I think this suggested change (adb5a2f) causes problems for high-pri interruptions, since moving this would mean that the WIP version doesn't get reset between a low priority render and a flush sync.

bvaughn:
I've updated this diff and squashed it into two commits:
* The first is my initial implementation, with a non-update-queue based approach.
* The second commit uses the update queue by composing `useEffect` and `useState` internally. This second approach ended up being a little bigger than I would have expected due to the need to create mount, update, **and rerender** implementations.

markerikson:
@bvaughn : one quick observation on your stated latest implementation:

For React-Redux specifically, we've found that we _have_ to subscribe in a `useLayoutEffect()`, otherwise it leads to bugs (see https://github.com/reduxjs/react-redux/pull/1444 and https://github.com/reduxjs/react-redux/issues/1313 ).  It's possible that `useMutableSource`'s semantics are different enough that that might not be an issue for us using it in the future, but FYI.

bvaughn:
@markerikson Probably easiest if we chat on the RFC about design related concerns 😄 Just to keep them all in one place. I'll quote you and comment over there.

bvaughn:
Note to self: I think I might revert this part of my most recent commit:
> 2. The subscribe callback function now expects to be passed the updated snapshot value.

Anytime we read from a source that we have not yet subscribed to (either because we have a new source or a new `subscribe` function) we have to fall back to checking the version and pending updates for the source:
https://github.com/facebook/react/blob/84637cc92002543798990d1a8f322bcf9e202f0e/packages/react-reconciler/src/ReactFiberHooks.js#L863-L915

This check has an ambiguous cases where a component's selector *might* be able to safely read from a source, but we can't know for sure- so we throw and de-opt the entire render (including all pending updates at all priorities for the current root).

Anything that sends us down this path more often increases the likelihood of this bigger de-opt.

bvaughn:
After some consideration, I reverted this part of my previous commit:
> 2. The subscribe callback function now expects to be passed the updated snapshot value.

Anytime we read from a source that we have not yet subscribed to (either because we have a new source or a new `subscribe` function) we have to fall back to checking the version and pending updates for the source:
https://github.com/facebook/react/blob/84637cc92002543798990d1a8f322bcf9e202f0e/packages/react-reconciler/src/ReactFiberHooks.js#L863-L915

This check has an ambiguous cases where a component's selector *might* be able to safely read from a source, but we can't know for sure- so we throw and de-opt the entire render (including all pending updates at all priorities for the current root).

Anything that sends us down this path more often increases the likelihood of this bigger de-opt.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +853,273 @@ function rerenderReducer<S, I, A>(
...
+          // If the value hasn't changed, no update is needed.
+          // Return state as-is so React can bail out and avoid an unnecessary render.
+          if (prevState.snapshot === newSnapshot) {
+            return prevState;
+          }
+
+          setState({
+            ...prevState,
+            snapshot: newSnapshot,
+          });
```
JeremyRH: Should this be returning new state rather than calling setState in a setState callback? [I encounter an error in this CodeSandbox when incrementing the count twice.](https://codesandbox.io/s/sparkling-firefly-0r1xb)

dai-shi: I encountered probably the same error: `TypeError: Cannot read property 'source' of undefined`

bvaughn: Ah, this looks like maybe a rebase mistake of some sort...not sure what's going on here. Yes, it should just be a return. Thanks for pointing it out :)

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +853,282 @@ function rerenderReducer<S, I, A>(
...
+      subscribe,
+    };
+
+    if (shouldResubscribe) {
+      const handleChange = () => {
+        // It's possible that this callback will be invoked even after being unsubscribed,
+        // if it's removed as a result of a subscription event/update.
+        // In this case, React might log a DEV warning about an update from an unmounted component.
+        // We can avoid triggering that warning with this check.
+        if (didUnsubscribe) {
```
acdlite: Can you elaborate more on why this warning is not legit? Is this the issue where the mutation event happens while there are still pending passive effects?

If it's only that, let's skip this check and solve holistically in a separate task/PR. Perhaps by wrapping the warning in a check for pending passive unmounts, as we've discussed in the past.

acdlite: Never mind, just saw your other PR! 😍

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +853,282 @@ function rerenderReducer<S, I, A>(
...
+}
+
+type MutableSourceRef<Source, Snapshot> = {|
+  getSnapshot: null | MutableSourceGetSnapshotFn<Source, Snapshot>,
+  source: null | MutableSource<Source>,
+  subscribe: null | MutableSourceSubscribeFn<Source, Snapshot>,
+|};
+
+function useMutableSourceImpl<Source, Snapshot>(
+  useEffectImpl: (
```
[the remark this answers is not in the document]
bvaughn: This seemed like a reasonable alternative, but I'd be happy to change it. Would be minimal code impact- would just remove the extra function call.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +853,282 @@ function rerenderReducer<S, I, A>(
...
+        }
+
+        const latestGetSnapshot = ((ref.current
+          .getSnapshot: any): MutableSourceGetSnapshotFn<Source, Snapshot>);
+        let newSnapshot = null;
+        let snapshotDidThrow = false;
+        try {
+          newSnapshot = latestGetSnapshot(source._source);
+        } catch (error) {
+          snapshotDidThrow = true;
```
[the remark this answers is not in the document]
bvaughn: What's the benefit of this change?

Happy to make the change. Just curious.

packages/react-reconciler/src/ReactFiberHooks.js
bvaughn: Yeah, I think I can remove this check once I rebase on master :)

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -2188,6 +2189,864 @@ function loadModules({
         });
       });
 
+      describe('useMutableSource', () => {
```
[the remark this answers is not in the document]
bvaughn: Uh, sure. The test suite for this hook isn't particular large compared to other hooks though, especially `useEffect` :)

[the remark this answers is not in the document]
acdlite: Yeah we should probably start splitting out those tests, too :D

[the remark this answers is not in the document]
bvaughn: Fair enough. I split the mutable source tests out with b028a0b.

packages/react-reconciler/src/ReactFiberHooks.js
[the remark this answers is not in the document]
acdlite: Main benefit is not passing a closure in the case where there's no error. If you needed to pass a closure anyway (like to check for stale sources, which I don't think you do for the reasons described in my earlier comment), then this would still have the benefit of saving one extra check from the common path.

bvaughn:
> Let's set aside `getSnapshot` for now. I think you're right that we don't have to resubscribe when that changes, but I wonder if this optimization is worth the added complexity. Not sure, will have to think about it.

I think it is, since the potential deopt in `readFromUnsubcribedMutableSource` is a big one. Let's chat more after you've had a chance to think about it though! :smile:

bvaughn:
Okay. I've refactored this hook to remove the `useRef` dependency (in favor of `memoizedState`) and to manually clear the `useState` update queue in the event of a new source/subscription. This latter change removes some unnecessary intermediate renders, which is nice.

Still want to write a few more tests (Monday morning).

packages/react-reconciler/src/ReactMutableSource.js
```diff
@@ -0,0 +1,80 @@
...
+
+export function getPendingExpirationTime(root: FiberRoot): ExpirationTime {
+  return root.mutableSourcePendingUpdateTime;
+}
+
+export function setPendingExpirationTime(
+  root: FiberRoot,
+  expirationTime: ExpirationTime,
+): void {
+  root.mutableSourcePendingUpdateTime = expirationTime;
```
bvaughn: I still feel a little unsure about the suggestion to store all pending updates with a single expiration time.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +854,224 @@ function rerenderReducer<S, I, A>(
...
+
+    // Normally a new getSnapshot function implies a new snapshot value.
+    // This function should be memoized so that it only changes when its dependencies change.
+    // Tf the returned snapshot value is the same-
+    // it likely means getSnapshot is not memoized correctly (e.g. inline function).
+    // In this case we can avoid the expensive deopt and just update local component state.
+    if (snapshot !== maybeNewSnapshot) {
+      // Even if the snapshot value did change, it might be okay to read the new value.
+      // If it's not we'll throw and restart the render (to avoid tearing).
+      snapshot = readFromUnsubcribedMutableSource(source, getSnapshot);
```
bvaughn: FYI: I will want to fork `readFromUnsubcribedMutableSource` rather than calling it again (just in case snapshot calculation is expensive; there's no need to re-run that function) but for now I'm going to hold off on this change until I get feedback on the overall approach.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -224,6 +237,22 @@ function useMemo<T>(
...
+  getSnapshot: MutableSourceGetSnapshotFn<Source, Snapshot>,
+  subscribe: MutableSourceSubscribeFn<Source, Snapshot>,
+): Snapshot {
+  // useMutableSource() composes multiple hooks internally.
+  // Advance the current hook index the same number of times
+  // so that subsequent hooks have the right memoized state.
+  nextHook(); // MutableSource
+  nextHook(); // State
+  nextHook(); // Effect
+  const value = getSnapshot(source._source);
```
bvaughn: Note to self: This isn't quite right, since it might return a copy of the data that's too new.

bvaughn:
I am still slightly concerned about the sync work loop issue. For now I'm just going to leave a disabled test with a TODO about this though.

bvaughn:
Rebased to account for partial renderer deletion.

bvaughn:
Added a follow-up task for tackling SSR+hydration #18183

packages/shared/ReactTypes.js
```diff
@@ -187,3 +187,40 @@ export type ReactScopeInstance = {|
   fiber: Object,
   methods: null | ReactScopeMethods,
 |};
+
```
vko-online:
```suggestion

```

packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
```diff
@@ -0,0 +1,1250 @@
...
+          'a:new:two',
+          'b:new:two',
+          'Sync effect',
+        ]);
+      });
+    });
+
+    // TODO (useMutableSource) Re-enable this test once workLoopSync can handle
+    // the thrown error without getting stuck in a cycle.
+    xit('should throw and restart render if source and snapshot are unavailable during a sync update', () => {
```
acdlite: This should work now if you rebase on master

packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
```diff
@@ -0,0 +1,1250 @@
...
+  ReactFeatureFlags.enableProfilerTimer = true;
+  React = require('react');
+  ReactNoop = require('react-noop-renderer');
+  Scheduler = require('scheduler');
+  useMutableSource = React.useMutableSource;
+  act = ReactNoop.act;
+}
+
+//[true, false].forEach(newPassiveEffectsTiming => {
+[false].forEach(newPassiveEffectsTiming => {
```
acdlite: Since https://github.com/facebook/react/pull/18234 landed, these flags are on in the `www-variant` test run. So you can remove this wrapper.

packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
bvaughn: Just rebased and still fails. :(

packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
bvaughn: This is nice.

packages/react-reconciler/src/ReactMutableSource.js
```diff
@@ -0,0 +1,118 @@
...
+  if (isPrimaryRenderer) {
+    mutableSource._workInProgressVersionPrimary = version;
+    workInProgressPrimarySources.push(mutableSource);
+  } else {
+    mutableSource._workInProgressVersionSecondary = version;
+    workInProgressSecondarySources.push(mutableSource);
+  }
+}
+
+export function warnAboutMultipleRenderersDEV(
```
bvaughn: I chose to put this into its own, DEV-only function rather than putting it in `getWorkInProgressVersion` because the latter isn't called in some cases. (It's only called when reading from an unsubscribed source.) The explicit check seemed better.

packages/react/index.stable.js
```diff
@@ -23,6 +23,8 @@ export {
   useDebugValue,
   useLayoutEffect,
   useMemo,
+  useMutableSource,
+  createMutableSource,
```
acdlite: Ah, almost missed this. Let's remove these from stable until we're ready to release them.

bvaughn: Ah, right.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +855,223 @@ function rerenderReducer<S, I, A>(
...
+  }
+}
+
+function useMutableSource<Source, Snapshot>(
+  hook: Hook,
+  source: MutableSource<Source>,
+  getSnapshot: MutableSourceGetSnapshotFn<Source, Snapshot>,
+  subscribe: MutableSourceSubscribeFn<Source, Snapshot>,
+): Snapshot {
+  const root = ((getWorkInProgressRoot(): any): FiberRoot);
```
acdlite: Don't forget to remove this

bvaughn: We need it, remember? (We track pending expiration time on per `root`)

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +855,223 @@ function rerenderReducer<S, I, A>(
...
+      if (typeof unsubscribe !== 'function') {
+        console.warn(
+          'Mutable source subscribe function must return an unsubscribe function.',
+        );
+      }
+    }
+
+    // Check for a possible change between when we last rendered and when we just subscribed.
+    const maybeNewVersion = getVersion(source._source);
+    if (!Object.is(version, maybeNewVersion)) {
```
acdlite: We don't assume `Object.is` support. We use our own implementation elsewhere; it's imported as `is` in this file.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +855,223 @@ function rerenderReducer<S, I, A>(
...
+  // Maybe the source changed in a way that the old subscription ignored but the new one depends on.
+  //
+  // If the getSnapshot function changed, we also shouldn't rely on the update queue.
+  // It's possible that the underlying source was mutated between the when the last "change" event fired,
+  // and when the current render (with the new getSnapshot function) is processed.
+  //
+  // In both cases, we need to throw away pending udpates (since they are no longer relevant)
+  // and treat reading from the source as we do in the mount case.
+  if (
+    !Object.is(prevSource, source) ||
```
acdlite: Same here

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +855,223 @@ function rerenderReducer<S, I, A>(
...
+
+  hook.memoizedState = ({
+    refs,
+    source,
+    subscribe,
+  }: MutableSourceMemoizedState<Source, Snapshot>);
+
+  // Sync the values needed by our subscribe function after each commit.
+  dispatcher.useEffect(() => {
+    refs.fiber = fiber;
```
acdlite: `computeExpirationForFiber` accepts either fiber alternate, so we can close over `workInProgress` instead of tracking this one with a ref.

bvaughn: Oh, cool.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +855,223 @@ function rerenderReducer<S, I, A>(
...
+      const latestFiber = refs.fiber;
+      try {
+        setSnapshot(latestGetSnapshot(source._source));
+
+        // Record a pending mutable source update with the same expiration time.
+        const currentTime = requestCurrentTimeForUpdate();
+        const suspenseConfig = requestCurrentSuspenseConfig();
+        const expirationTime = computeExpirationForFiber(
+          currentTime,
+          latestFiber,
```
acdlite: This can be `currentlyRenderingFiber`, no ref needed. (Gasp... we're breaking the hook rule! :D)

bvaughn: Yup.... that's what you said in the comment right above, no? :)

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: You need it in `readFromUnsubcribedMutableSource` but I  don't see a reference anywhere in the outer function, except the invariant check.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +855,223 @@ function rerenderReducer<S, I, A>(
...
+          suspenseConfig,
+        );
+
+        setPendingExpirationTime(root, expirationTime);
```
acdlite: Oh right here.

Hmm.

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: Oh https://github.com/facebook/react/pull/18000/files#r389111550

bvaughn: `readFromUnsubcribedMutableSource` uses it [here](https://github.com/facebook/react/blob/967d80f27e3fe8285298253d10221dcd05100d4f/packages/react-reconciler/src/ReactFiberHooks.js#L894-L895) to *get* the pending expiration time, and this function uses it [here](https://github.com/facebook/react/blob/967d80f27e3fe8285298253d10221dcd05100d4f/packages/react-reconciler/src/ReactFiberHooks.js#L992) to *update* the pending expiration time.

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: Meh we can figure this out later. I think I'll need to during my expiration times refactor, anyway.

bvaughn: 😁 Ok.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -891,6 +885,12 @@ function readFromUnsubcribedMutableSource<Source, Snapshot>(
   if (currentRenderVersion !== null) {
     isSafeToReadFromSource = currentRenderVersion === version;
   } else {
+    const root = ((getWorkInProgressRoot(): any): FiberRoot);
```
acdlite: If you call this in the outer function, you might as well pass it as an argument. Not a big deal either way though because the function be inlined. But you'd only need a single type check.

scripts/error-codes/codes.json
```diff
@@ -346,5 +346,8 @@
   "345": "Root did not complete. This is a bug in React.",
   "346": "An event responder context was used outside of an event cycle.",
   "347": "Maps are not valid as a React child (found: %s). Consider converting children to an array of keyed ReactElements instead.",
-  "348": "ensureListeningTo(): received a container that was not an element node. This is likely a bug in React."
+  "348": "ensureListeningTo(): received a container that was not an element node. This is likely a bug in React.",
+  "349": "Expected a work-in-progress root. This is a bug in React. Please file an issue.",
+  "350": "Cannot read from mutable source during the current render without tearing. This is a bug in React. Please file an issue.",
+  "351": "Mutable source subscribe function must return an unsubscribe function."
```
acdlite: Can remove this entry now that it's a warning

packages/react-reconciler/src/ReactFiberHooks.js
bvaughn: Gah, bad habit from working in DevTools code for so long.

bvaughn:
Code Sandbox CI failure is unrelated
```
/tmp/7f74b520/scripts/rollup/build.js:44
  throw err;
  ^

Error: Error: EEXIST: file already exists, mkdir '/tmp/7f74b520/build/react-native/shims'
    at error (/tmp/7f74b520/scripts/rollup/utils.js:17:20)
    at cb (/tmp/7f74b520/node_modules/ncp/lib/ncp.js:255:23)
    at onError (/tmp/7f74b520/node_modules/ncp/lib/ncp.js:247:12)
    at /tmp/7f74b520/node_modules/ncp/lib/ncp.js:159:16
    at FSReqWrap.args [as oncomplete] (fs.js:140:20)
```
I've seen this one on other PRs. Kind of annoying since we have no way to restart that particular job that I'm aware of.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +855,214 @@ function rerenderReducer<S, I, A>(
...
+        setSnapshot(() => {
+          throw error;
+        });
+      }
+    };
+
+    const unsubscribe = subscribe(source._source, handleChange);
+    if (__DEV__) {
+      if (typeof unsubscribe !== 'function') {
+        console.warn(
```
acdlite: I think this should be console.error instead

packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
```diff
@@ -0,0 +1,1317 @@
...
+                  subscribe={defaultSubscribe}
+                />
+              </>,
+            );
+            expect(Scheduler).toFlushAndYieldThrough(['a:one']);
+
+            const PrevScheduler = Scheduler;
+
+            // Get a new copy of ReactNoop.
+            loadModules();
```
acdlite: Could use another renderer like ART, instead of loading noop twice, so you don't have to mutate the flags

packages/react/index.classic.fb.js
```diff
@@ -26,6 +26,8 @@ export {
   useReducer,
   useRef,
   useState,
+  useMutableSource,
```
acdlite: When we land this in www, I suggest prefixing these APIs in the entry point file until we're ready for everyone to  use them. Then tell people to only use it via a module. That way we can change the API without having to run a codemod.

packages/react-reconciler/src/ReactFiberHooks.js
bvaughn: Oh ok, I thought you'd suggested warn. Will change.

packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
bvaughn: I don't think our tests allow multiple renderers to be used without resetting things anyway, because of how the host config stuff works. I think that's why we do the module resetting elsewhere. :)

acdlite: Ah I got mixed and thought you were changing the flags in between runs. Got it :D

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: Yeah I'm still stuck calling them warning even though we use `console.error`

I hate that name

bvaughn:
I thought of a failing case (test soon to be pushed) that this hook triggers. Andrew and I chatted out of band. It seems to be a pre-existing bug that he's going to submit a fix for. In the meanwhile I'm going to hold off on landing this PR until the new test passes. :smile:

bvaughn:
I rebased my PR on top of #18265 to see if it fixes the failing tests recently added (a6d4eed).

It fixes [this one](https://github.com/facebook/react/blob/a6d4eed534cbbc480577fd52be19d669af679af4/packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js#L812-L843), which is good 👍

It does not fix [this one](https://github.com/facebook/react/blob/a6d4eed534cbbc480577fd52be19d669af679af4/packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js#L1348-L1404), but I wouldn't expect it to fix that test alone. The test has another problem: When a change of version is detected, React throws and unwinds- and is supposed to reset all WIP version numbers. This is done by looping through the mutable sources that were *set* during the current render and resetting them. In this case though, we never *set* the torn source for the second renderer, so resetting doesn't actually do anything.

I guess we have a couple of options here:
1. Reset the WIP version (in the hook) *before* throwing for the source that's torn. I *think* this may be the best option for this edge case.
1. Track *reads* in addition to *writes* so we'll be sure to get the torn source too. This would add overhead to the hotter read path though so it's probably not a great idea.
1. Something else?

acdlite:
I think this is fine. We are intentionally not supporting anything more than a primary renderer and a secondary renderer, to keep the simpler implementation. And the reason to use a warning instead of an error is so the check is stripped out in prod.

bvaughn:
Right. That was my thinking too.

Okay then. Good to go?

Sawtaytoes:
Is there gonna be any documentation about how to use this hook on the React docs site?

bvaughn:
Once it's in a stable release, yes. This is still only available in the experimental React release channel.

PR #18001, opened 2020-02-07
Added $FlowFixMe to DevTools shell for module we Flow-ignore

bvaughn:
Fix Flow error introduced in #17999


bvaughn:
Seems expected since #17999 broke master. This should fix it.

PR #18002, opened 2020-02-08
Re-enable shorthand CSS property collision warning

### LINKED ISSUE #16393: Shorthand CSS property collision should trigger a warning
I faced a "bug" today that made me spent 1h figuring out what was going on:

I'm using an external component that accepts a `color` prop in order to set the `background-color` of the root element, but this same component also accepts a `background` prop which I wasn't passing, and by default, it was set to `''`.

The result: React didn't throw an error nor a warning, however, the resulting element in the DOM didn't contain either `background-color` or `background`, and since the element had a default `background-color` coming from a CSS class, it took me a while to figure out why the color that I was passing wasn't being applied, and instead it was using the one from the CSS class.

See: https://codesandbox.io/s/react-example-8rxc8

What I reported above was the `static1` case.

I added other cases as a bonus, as when I was playing with this they also seemed weird to me. On `static2` I define the same properties, but because I change the order, it works. On the toggleable ones, initially I can see the background, but after changing it never appears anymore.

I'm not sure if I created those extra "test cases" correctly. My main concern is really around static1 not outputting anything on the console as a warning.

Related issues:
#6348
#8689

I wonder if #14181 (@sophiebits) should have covered this?

sophiebits:
Originally added in https://github.com/facebook/react/pull/14181; disabled in https://github.com/facebook/react/pull/14245. Intention was to enable it in React 16.7 but we forgot.

Fixes #16393.


sophiebits:
Renamed the test and removed the feature flag from it.

craigkovatch:
Three cheers for @sophiebits !!!

PR #17825, opened 2020-01-12
Grammar fix

haseebdaone:
Just going through the source to get a better understanding of React and saw a grammar mistake.

replaced 'add' with 'subtract'

PR #18007, opened 2020-02-10
Tweak comment wording

gaearon:
Reverts facebook/react#17825

I don't think this is a "grammar fix". This is a semantic change to what the comment says. It's misleading because the `MAGIC_NUMBER_OFFSET` is still being *added*, not subtracted. `A - B` does not mean "subtract `A`".

In general, changing the meaning of comments should be done very carefully because misleading comments are especially dangerous. Ideally, it's best to double-check what the author meant.


gaearon:
"Subtract from the offset" should be clearer and reflects what the code is doing.

PR #18008, opened 2020-02-10
`disableLegacyContext` as a dynamic flag for www

threepointone:
This PR makes `disableLegacyContext` a dynamic feature flag for fb/www. For FB internal usage only.


gaearon:
We did this differently.

PR #18009, opened 2020-02-10
Add FB_WWW_MODERN bundle types

gaearon:
So that we can start testing slimmer builds internally.

Currently these are 100% matching the FB_WWW classic builds. I verified by building both modes and diffing them. In a follow-up PR, I will fork the feature flags and make some other changes for them.


gaearon:
That was exactly what I suggested to Seb, but he said he wants another feature flag file instead. Either I misunderstood what he was proposing, or we might want to chat about this all together.

gaearon:
That said, I find the idea of “experimental” meaning different things in fbsource and WWW scary. That’s why I scoped it to clearly be a WWW only thing.

acdlite:
Might have been a misunderstanding. @sebmarkbage and I chatted offline. Let’s discuss at the meeting!

gaearon:
Cool, no worries! Just want to make sure we agree 100% before I rewrite :-)

gaearon:
>The names of all the files are identical except for which subdirectory they live in.

We'll have to enable Haste packages for this to work, right? I guess you need to either way because of the "two Reacts" thing.

PR #18012, opened 2020-02-10
Move getIsHydrating from FB fork to Experimental flag

gaearon:
This has no actual impact except that getIsHydrating is on the secrets object for everyone, and not just FB. This would let us completely get rid of the forked FB-specific entry point in the modern version. Instead of the forked entry point, we will use feature flags to remove individual features.

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -37,6 +37,7 @@ import {
   attemptHydrationAtCurrentPriority,
   act,
 } from 'react-reconciler/inline.dom';
+import {getIsHydrating} from 'react-reconciler/src/ReactFiberHydrationContext';
 import {createPortal as createPortalImpl} from 'shared/ReactPortal';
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/client/ReactDOM.js
```diff
@@ -179,6 +180,8 @@ if (exposeConcurrentModeAPIs) {
       queueExplicitHydrationTarget(target);
     }
   };
+
+  ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.getIsHydrating = getIsHydrating;
```
sebmarkbage: It's very important that this isn't exposed in the CM release. It won't be needed by the future API neither.

Instead, put it behind a specific feature flag that we won't enable.

gaearon:
Ahh I thought this was for hero tracing or something like this. Okay I'll disable and just remove this.

gaearon:
Ain't needed.

PR #18013, opened 2020-02-10
Disable classic features in the modern WWW build

gaearon:
# Builds on https://github.com/facebook/react/pull/18009 + https://github.com/facebook/react/pull/18012

This drops a bunch of features from the WWW modern builds. I set it to pretty aggressive so we can try that and see what we have to revert.

- No more forked entry point. Same ReactDOM as the rest of the world.
- No user timings.
- No EventListener wrapping.
- No legacy context.
- No textarea children.
- No renderSubtree.
- No unstable_ prefix for createPortal.
- No maps as children.
- No findDOMNode, render, hydrate, unmountComponentAtNode.

See individual commits for closer details.


packages/shared/forks/ReactFeatureFlags.www-modern.js
```diff
@@ -0,0 +1,86 @@
...
+
+export const enableChunksAPI = __EXPERIMENTAL__;
+
+export const disableJavaScriptURLs = true;
+
+export function addUserTimingListener() {
+  throw new Error('Not implemented.');
+}
+
+export const enableDeprecatedFlareAPI = true;
```
trueadm: Can we add a comment to group the two sets up more? I’m worried people will add to the bottom of this list and it will get muddled up otherwise.

gaearon: This is 100% copy paste from the other file. I agree we should clean it up but let's do this separately from this PR.

trueadm: Sounds good.

gaearon:
We did this differently.

PR #18021, opened 2020-02-11
Revert "Remove legacy www config from Rollup build"

bvaughn:
Reverts #18016

This change caused a Makehaste failure for syncs:
```
Cannot transform 'shared/react/ReactEventsFocus-dev.js': TypeError: Unsupported function type: ObjectMethod
...
```
The specific syntax change that causes the failure is going from this:
```js
var Focus = (Object.freeze || Object)({
  passiveBrowserEventsSupported: passiveBrowserEventsSupported,
```
to this:
```js
var Focus = Object.freeze({
  get passiveBrowserEventsSupported() {
    return passiveBrowserEventsSupported;
  },
```

### DIFF
--- a/scripts/rollup/build.js
+++ b/scripts/rollup/build.js
@@ -541,6 +541,11 @@ async function createBundle(bundle, bundleType) {
       bundle.moduleType,
       pureExternalModules
     ),
+    // We can't use getters in www.
+    legacy:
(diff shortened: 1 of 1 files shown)


bvaughn:
Closing in favor of #18022

PR #18022, opened 2020-02-11
Do not export passiveBrowserEventsSupported from Focus module

trueadm:
We should not be exporting `passiveBrowserEventsSupported` from this module.


threepointone:
Why re-include the legacy builds if you’re removing the problematic export?

PR #18023, opened 2020-02-12
Put React.jsx and React.jsxDEV behind experimental build

lunaruan:
This PR puts the `React.jsx` and `React.jsxDEV` (`enableJSXTransformAPI` feature flag) in the experimental build so that we can use it to test React Native.

### DIFF
--- a/packages/react/index.experimental.js
+++ b/packages/react/index.experimental.js
@@ -43,4 +43,8 @@ export {
   unstable_withSuspenseConfig,
   // enableBlocksAPI
   block,
+  // enableJSXTransformAPI
+  jsx,
+  jsxs,
+  jsxDEV,
 } from './src/React';


packages/react/src/React.js
```diff
@@ -141,18 +140,15 @@ if (enableScopeAPI) {
...
-    React.jsxs = jsxWithValidationStatic;
-  } else {
-    React.jsx = jsx;
-    // we may want to special case jsxs internally to take advantage of static children.
-    // for now we can ship identical prod functions
-    React.jsxs = jsx;
-  }
+if (__DEV__) {
+  React.jsxDEV = jsxWithValidation;
+  React.jsx = jsxWithValidationDynamic;
```
threepointone: Would you consider moving `.jsx` and ` .jsxs` up to Line 105? 
```jsx
  jsx: __DEV__ ? jsxWithValidationDynamic : jsx,
  jsxs: __DEV__ ? jsxWithValidationStatic : jsx,

```
and leave `React.jsxDEV = jsxWithValidation;` as it is here in this `__DEV__` check

Just for consistency with createElement.

sebmarkbage:
This warning is enabled in jsx:

https://github.com/facebook/react/blob/f2fd484afdee1e4e25ee453bb7a544fa0558d172/packages/react/src/ReactElementValidator.js#L371-L373

This means that upgrading to future Babel plugins would enable this warning. Should we wait to enable this until later so it becomes part of a React upgrade instead of a Babel upgrade?

threepointone:
The 'Spreading a key to JSX' warning is now behind a feature flag (https://github.com/facebook/react/pull/18074). Could we rebase and land this?

PR #18024, opened 2020-02-12
Revert "Remove legacy www config from Rollup build"

gaearon:
Reverts facebook/react#18016

I think we still need to do this. https://github.com/facebook/react/pull/18022 just solved one specific case, and we happen to not have more cases at the moment. But if making a change to the code can cause this problem to reappear, the problem is not solved.


gaearon:
Let's just fix this: D19880784.

gaearon:
I solved this in WWW instead.

PR #18025, opened 2020-02-12
Refactor DOM plugin system to fewer modules

trueadm:
This PR refactors and cleans up the existing logic around React's DOM event plugin system. This is primiary to make the system somewhat easier to fork and manage as part of the future work and investigation into possibly using React Roots as the delegation nodes rather than using only the `document`. **Please note, there should be no logic/functionaly changes from this PR, and all existing code paths should still work.**

Notably, this PR does the following:

- We have a new module called `DOMEventPluginSystem`, this is where the bulk of the DOM plugin event system sits. This file should be forkable for future work with relative ease.
- `EventPluginHub` has been removed. Instead the core functions have now moved to their respective renderers instead. This should make it easier to make changes to the DOM event system in the future without impacting other renderers that keep the existing plugin system.
- `getListener` has been added. This function was previously in `EventPluginHub`, but now is a single module and exported function, as it is required during the event propagation phase.
- `ReactDOMEventListener` previously had a lot of logic relating to the DOM plugin system. This has now been moved into `DOMEventPluginSystem`.
- We had a test called `EventPluginHub-test` that was more a test for invalid listeners, so the test has changed name to reflect this better.
- We directly inject plugins into the `EventPluginRegistry` rather than the `EventPluginHub`.
- The DOM event plugin order has been inlined into `ReactDOMClientInjection`. In future React versions, we might want to remove this module and instead inline the event plugins directly into the event system instead to save bytes and improve runtime performance (RNW currently requires this as of React 16).
- We should get a small size perf win for ReactDOM because we've removed some of the module indirection from before.


scripts/error-codes/codes.json
```diff
@@ -98,8 +98,8 @@
   "96": "EventPluginRegistry: Cannot inject event plugins that do not exist in the plugin ordering, `%s`.",
   "97": "EventPluginRegistry: Event plugins must implement an `extractEvents` method, but `%s` does not.",
   "98": "EventPluginRegistry: Failed to publish event `%s` for plugin `%s`.",
-  "99": "EventPluginHub: More than one plugin attempted to publish the same event name, `%s`.",
-  "100": "EventPluginHub: More than one plugin attempted to publish the same registration name, `%s`.",
+  "99": "EventPluginRegistry: More than one plugin attempted to publish the same event name, `%s`.",
+  "100": "EventPluginRegistry: More than one plugin attempted to publish the same registration name, `%s`.",
```
trueadm: I wasn't sure if we could change existing error code messages or if we had to make new ones?

packages/legacy-events/getListener.js
```diff
@@ -0,0 +1,74 @@
...
+    // Work in progress (ex: onload events in incremental mode).
+    return null;
+  }
+  const props = getFiberCurrentPropsFromNode(stateNode);
+  if (!props) {
+    // Work in progress.
+    return null;
+  }
+  listener = props[registrationName];
+  if (shouldPreventMouseEvent(registrationName, inst.type, props)) {
```
trueadm: It's a shame that this runs for Fabric and RN too. This is definitely something we should remove once we get to clean up the existing DOM event system.

packages/legacy-events/ResponderEventPlugin.js
```diff
@@ -282,7 +282,7 @@ to return true:wantsResponderID|                            |
                                +                            + */
 
 /**
- * A note about event ordering in the `EventPluginHub`.
+ * A note about event ordering in the `EventPluginRegistry`.
```
trueadm: We inject into the `EventPluginRegistry` now, so the comment changes are to reflect that.

scripts/error-codes/codes.json
gaearon: Either should be fine I think since their meaning is the same.

PR #18027, opened 2020-02-12
remove "Unreleased" section from CHANGELOG

threepointone:
This section is empty, and imo isn't really helpful in React's changelog. I'm honestly not sure why this is even here? Figured I'd start a discussion with a PR, or we can remove it right now.


gaearon:
It's there because we wanted to add items to changelog together with PRs. However, we never actually started doing that, and I guess we won't. So fine.

PR #18028, opened 2020-02-12
Add Modern WWW build

gaearon:
Alternative to https://github.com/facebook/react/pull/18013.

**This changes the meaning of release channels for the WWW builds:**

- The stable channel (which was unused) is now the build we ship on WWW.
- The experimental channel is now used for the new modern build. We won't use it in production yet, but we'll start playing with it. It hard-breaks a bunch of legacy features.

**This is unsafe to land** until the script is updated to pull in both versions and put them in the right folders. We'll also need to solve the Haste conflicts.

Easier to review individual commits.

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -56,6 +56,7 @@ import ReactVersion from 'shared/ReactVersion';
 import invariant from 'shared/invariant';
 import {
   exposeConcurrentModeAPIs,
+  disableLegacyReactDOMAPIs,
   disableUnstableCreatePortal,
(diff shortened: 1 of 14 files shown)


gaearon:
I’ll prepare a WWW diff for the sync script before merging. Do you have a Haste strategy in mind? We could add package.json that enable separate Haste packages. If I remember how that works correctly.

acdlite:
Hmm I was thinking you'd use `requireCond` but I can see how that would be tedious to set up for every possible artifact.

gaearon:
I mean more that providesModules names (and filenames) need to be unique. Unless you set yo something special like we do for node_modules.

PR #18029, opened 2020-02-12
Re-throw errors thrown by the renderer at the root in the complete phase

acdlite:
React treats errors thrown at the root as a fatal because there's no parent component that can capture it. (This is distinct from an "uncaught error" that isn't wrapped in an error boundary, because in that case we can fall back to deleting the whole tree – not great, but at least the error is contained to a single root, and React is left in a consistent state.)

It turns out we didn't have a test case for this path. The only way it can happen is if the renderer's host config throws. We had similar test cases for host components, but none for the host root.

This adds a new test case and fixes a bug where React would keep retrying the root because the `workInProgress` pointer was not advanced to the next fiber. (Which in this case is null, since it's the root.)

We could consider in the future trying to gracefully exit from certain types of root errors without leaving React in an inconsistent state. For example, we should be able to gracefully exit from errors thrown in the begin phase. For now, I'm treating it like an internal invariant and immediately exiting.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -1285,6 +1285,12 @@ function handleError(root, thrownValue) {
         // boundary.
         workInProgressRootExitStatus = RootFatalErrored;
         workInProgressRootFatalError = thrownValue;
+        // Set `workInProgress` to null. This represents advancing to the next
+        // sibling, or the parent if there are no siblings. But since the root
+        // has no siblings nor a parent, we set it to null. Usually this is
+        // handled by `completeUnitOfWork` or `unwindWork`, but since we're
+        // interntionally not calling those, we need set it here.
+        // TODO: Consider calling `unwindWork` to pop the contexts.
         workInProgress = null;
         return null;
       }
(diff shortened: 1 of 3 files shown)


sebmarkbage:
Your description doesn’t describe what this fix does?

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1285,6 +1285,7 @@ function handleError(root, thrownValue) {
         // boundary.
         workInProgressRootExitStatus = RootFatalErrored;
         workInProgressRootFatalError = thrownValue;
+        workInProgress = null;
```
sebmarkbage: What is happening here? Why does this work?

acdlite: > This adds a new test case and fixes a bug where React would keep retrying the root because the workInProgress pointer was not advanced to the next fiber. (Which in this case is null, since it's the root.)

acdlite: In other words, it falls into an infinite loop because `workInProgress` is stuck on the the host root fiber forever. Usually `workInProgress` gets advanced during either `beingWork` or `completeUnitOfWork` but since it keeps erroring that never happens.

sebmarkbage: Ah. Subtle. A common would be nice but not required.

acdlite: Ok I'll add one! I think there's a potential follow up to also call `unwindWork` here to pop the contexts. I originally didn't bother because it's not impossible even that call will throw, but if layout errors are common enough it might be worth doing more to prevent corrupted internal state.

PR #18030, opened 2020-02-12
Split recent passive effects changes into 2 flags

bvaughn:
Separate flags can now be used to opt passive effects into:
1. `deferPassiveEffectCleanupDuringUnmount`: Defer passive effect destroy functions on unmount to subsequent passive effects flush. (#17925)
2. `runAllPassiveEffectDestroysBeforeCreates`: Running all passive effect destroy functions (for all fibers) before create functions. (#17947)

This allows us to test the less risky feature (#17947) separately from the more risky one.

Note that:
* `deferPassiveEffectCleanupDuringUnmount` is ignored unless `runAllPassiveEffectDestroysBeforeCreates` is also enabled.
* Both flags remain off for the open source release.

I've updated a couple of tests that were impacted by the change to explicitly test with both feature flag values so we don't have blind spots.

**Be sure to use the [`?w=1` param](https://github.com/facebook/react/pull/18030/files?w=1) to see the significant lines changed (`+393 −99` rather than `+5,471 −5,177`).**

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
@@ -34,2759 +34,2950 @@ let forwardRef;
 let memo;
 let act;
 
-describe('ReactHooksWithNoopRenderer', () => {
-  beforeEach(() => {
-    jest.resetModules();
-    jest.useFakeTimers();
+function loadModules({
+  deferPassiveEffectCleanupDuringUnmount,
+  runAllPassiveEffectDestroysBeforeCreates,
+}) {
+  ReactFeatureFlags = require('shared/ReactFeatureFlags');
+  ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
+  ReactFeatureFlags.enableSchedulerTracing = true;
+  ReactFeatureFlags.flushSuspenseFallbacksInTests = false;
+  ReactFeatureFlags.deferPassiveEffectCleanupDuringUnmount = deferPassiveEffectCleanupDuringUnmount;
+  ReactFeatureFlags.runAllPassiveEffectDestroysBeforeCreates = runAllPassiveEffectDestroysBeforeCreates;
+  React = require('react');
+  ReactNoop = require('react-noop-renderer');
+  Scheduler = require('scheduler');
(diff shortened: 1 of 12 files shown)


packages/shared/ReactFeatureFlags.js
```diff
@@ -91,10 +91,19 @@ export const enableTrustedTypesIntegration = false;
...
+// before any create functions are run, similar to how layout effects work.
+// This flag provides a killswitch if that proves to break existing code somehow.
+export const runAllPassiveEffectDestroysBeforeCreates = false;
+
 // Controls behavior of deferred effect destroy functions during unmount.
 // Previously these functions were run during commit (along with layout effects).
 // Ideally we should delay these until after commit for performance reasons.
 // This flag provides a killswitch if that proves to break existing code somehow.
+//
+// WARNING Only enable this flag in conjunction with runAllPassiveEffectDestroysBeforeCreates.
```
bvaughn: This constraint is a little gross ☹️ but it keeps the branching complexity low.

acdlite: Instead of two booleans you could use an enum, to prevent the invalid combination

bvaughn: True. I considered that but our precedent has always been booleans for flags.

I don't feel strongly about this though.

acdlite: Alternative, in the code, you could also check for both booleans before you enable the "defer unmounts" feature.

Just want to make it harder for someone to accidentally toggle the wrong combination of GKs.

bvaughn: Yeah, I hear you. That seems reasonable.

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -34,2759 +34,2953 @@ let forwardRef;
...
+  useImperativeHandle = React.useImperativeHandle;
+  forwardRef = React.forwardRef;
+  memo = React.memo;
+  useTransition = React.useTransition;
+  useDeferredValue = React.useDeferredValue;
+  Suspense = React.Suspense;
+  act = ReactNoop.act;
+}
+
+[true, false].forEach(deferPassiveEffectCleanupDuringUnmount => {
```
threepointone: wow function 
     much pure

PR #18031, opened 2020-02-13
Change string refs in function component message

sebmarkbage:
This should refer to string refs specifically. The forwardRef part doesn't make any sense in this case. I think this was just an oversight. This case is when a string ref is used _from_ a function component, not _on_ a function component.


devknoll:
Wait, don't you still need to wrap a function component with `forwardRef` on a function component in order to pass it a `ref`?

scripts/error-codes/codes.json
```diff
@@ -307,7 +307,7 @@
   "306": "Element type is invalid. Received a promise that resolves to: %s. Lazy element type must resolve to a class or function.%s",
   "307": "Hooks can only be called inside the body of a function component. (https://fb.me/react-invalid-hook-call)",
   "308": "Context can only be read while React is rendering. In classes, you can read it in the render method or getDerivedStateFromProps. In function components, you can read it directly in the function body, but not inside Hooks like useReducer() or useMemo().",
-  "309": "Function components cannot have refs. Did you mean to use React.forwardRef()?",
+  "309": "Function components cannot have string refs. We recommend using useRef() instead. Learn more about using refs safely here: https://fb.me/react-strict-mode-string-ref",
```
gaearon: Should we update this doc to mention useRef?

PR #18034, opened 2020-02-13
alias React.jsxDEV to React.jsx in prod builds (behind a feature flag)

threepointone:
For FB, we need to make it so React.jsxDEV calls 'work' even with prod builds. Setting this PR up as a possible option/discussion point, while I add some more tests.

For testing, I forked `ReactJSXElementValidator-test.js` to `ReactJSXElementValidator-test.internal.js`, turned on the feature flag, and aliased `React.jsx` with `React.jsxDEV`. 

These tests will run in test-source and test-source-prod; if we change the function signature in the future, then one of these should fail. 

My understanding is this might be a bit of a stop gap while we figure out a solution for FB, so this might not be necessary in the future.

### DIFF
--- a/packages/react/src/React.js
+++ b/packages/react/src/React.js
@@ -65,6 +65,7 @@ import {
   exposeConcurrentModeAPIs,
   enableChunksAPI,
   disableCreateFactory,
+  jsxDEVIncluded,
 } from 'shared/ReactFeatureFlags';
 const React = {
   Children: {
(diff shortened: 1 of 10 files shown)


packages/react/src/__tests__/ReactElementJSX-test.internal.js
```diff
@@ -396,4 +396,14 @@ describe('ReactElement.jsx', () => {
     // TODO: an explicit expect for no warning?
     ReactDOM.render(React.jsx(Parent, {}), container);
   });
+
+  describe('jsxDEV in prod', () => {
+    beforeEach(() => {
+      ReactFeatureFlags.jsxDEVIncluded = true;
+    });
+
+    it.only('should work', () => {
```
gaearon: .only

PR #18035, opened 2020-02-13
feat: honor displayName of context types

eps1lon:
## Summary

Any component name in errors/warnings that occur during server-side rendering ignore the `displayName`. However, react-devtools does use this property when computing the component name. The [docs mention this property](https://reactjs.org/docs/context.html#contextdisplayname) but are very specific about dev tools usage. Is this limitation intentional and if so: why?

## Test Plan

Tested by triggering a prop-types error.


eps1lon:
Changed the testing approach and rebase. @bvaughn could you take a look?

packages/react/src/__tests__/ReactContext-test.js
```diff
@@ -0,0 +1,45 @@
...
+        </Context.Provider>,
+      );
+    }).toErrorDev(
+      'Warning: Failed prop type: The prop `dontPassToSeeErrorStack` is marked as required in `Validator`, but its value is `undefined`.\n' +
+        '    in Validator (at **)\n' +
+        '    in MyContextType.Consumer (at **)\n' +
+        '    in MyContextType.Provider (at **)',
+    );
+  });
+});
```
bvaughn: nit: This test should probably just be in [`packages/react/src/__tests__/ReactContextValidator-test`](https://github.com/facebook/react/blob/master/packages/react/src/__tests__/ReactContextValidator-test.js) with the other validation tests.

eps1lon: Done.

trueadm:
I've had to revert this PR. Details are in https://github.com/facebook/react/pull/18223.

bvaughn:
Thanks, Dominic!

PR #18041, opened 2020-02-14
React developer tools extension for Microsoft Edge 

hmohamme:
## Summary

Motivation:
I switched to Microsoft Edge browser (chromium version) for my web development work and noticed react development tools extension was missing. It was a major blocker for me to keep using edge so, I ported it to work with edge.

Changes: 
Created new directory for edge extension.
Modified code to point to this new directory.
Added commands to build edge extension.
Added instructions to test the extension.
Added a package to open edge browser when `yarn test:edge` is run.

Chrome extensions work out of the box with Edge, but to pass the Microsoft certification process, all brand names were changed to Microsoft Edge.

### DIFF
--- a/packages/react-devtools-extensions/edge/test.js
+++ b/packages/react-devtools-extensions/edge/test.js
@@ -2,12 +2,21 @@
 
 'use strict';
 
-const chromeLaunch = require('chrome-launch');
+const edge = require('windows-edge');
+ 
 const {resolve} = require('path');
 
 const EXTENSION_PATH = resolve('./edge/build/unpacked');
 const START_URL = 'https://facebook.github.io/react/';
 
-chromeLaunch(START_URL, {
-  args: [`--load-extension=${EXTENSION_PATH}`],
-});
+edge({ uri: START_URL }, (err, ps) => {
+  if (err) throw err;
+  ps.on('error', console.error);
+  ps.on('exit', (code) => {
+    // Browser exited
+  });
+  setTimeout(() => {
+    ps.kill();
+  }, 2000);
+})
+
--- a/packages/react-devtools-extensions/edge/test.js
+++ b/packages/react-devtools-extensions/edge/test.js
@@ -1,0 +1,13 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const chromeLaunch = require('chrome-launch');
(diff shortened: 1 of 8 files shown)


packages/react-devtools-extensions/package.json
```diff
@@ -37,6 +41,7 @@
     "web-ext": "^3.0.0",
     "webpack": "^4.41.2",
     "webpack-cli": "^3.3.10",
-    "webpack-dev-server": "^3.3.1"
+    "webpack-dev-server": "^3.3.1",
+    "windows-edge": "^1.0.1"
```
bvaughn: New dependencies in `package.json` should be accompanied by an updated `yarn.lock` but I can always add this after the fact :)

packages/react-devtools-extensions/edge/test.js
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
+
+'use strict';
+
+const edge = require('windows-edge');
```
bvaughn: This NPM package only supports the Windows platform, which means that most of us (myself included) can't actually use it 😄 I wasn't able to find a Mac compatible version of this when I last looked. Wondering if you know of any?

packages/react-devtools-extensions/package.json
hmohamme: I have package-lock.json but there is no yarn.lock. I will see if I can generate yarn.lock and push with my next PR.

packages/react-devtools-extensions/edge/test.js
hmohamme: Sorry about that. 

I found [this package ](https://www.npmjs.com/package/open) which might work with some code changes. I will find a mac to test and submit a new PR once ready.

packages/react-devtools-extensions/package.json
bvaughn: No worries. I've already updated the Yarn lock in master.

To update it you'd need to use `yarn` to add new dependencies rather than NPM :)

packages/react-devtools-extensions/edge/test.js
bvaughn: Excellent! Thank you!

PR #18043, opened 2020-02-15
Add component stack parser

motiz88:
## Summary

Adds a parser for component stack strings, acting roughly as the inverse of [`getStackByFiberInDevAndProd`](https://github.com/facebook/react/blob/dd8205cef9a3176ef29dd5904f1352561f7ef43a/packages/react-reconciler/src/ReactCurrentFiber.js#L49-L57). It's exposed only via `react-debug-tools` and there is no plan to make it an officially supported public API.

This API is intended to allow privileged consumers (React Native etc) to process the information in component stacks without [rolling their own parsers](https://github.com/facebook/react-native/blob/f3d37325cbeb25db2aae2f5aa7b37de7b8a06f82/Libraries/LogBox/Data/parseLogBoxLog.js#L126-L146) (and without requiring direct access to React's internal data structures). Additionally, this will allow React to change the format of component stacks as needed without breaking these consumers.

This specific design is informed by a previous discussion between @sebmarkbage, @gaearon, @rickhanlonii and myself. The implementation is the best I could come up with naively but I'm sure there are wrinkles to be ironed out.

## Context for this change

We want to make it possible to decode component stacks using a source map, as a sustainable alternative to adding `displayName` properties to components in minified builds. Thanks to [function maps in Metro](https://github.com/facebook/metro/commit/64fc81e676ea781234dfe20c2bfaabdb1bef6b9e) and the Hermes [`getFunctionLocation`](https://github.com/facebook/hermes/commit/7364f96f19f87ea9317ca65541a8f154e2e839ff) API, this will first be possible for React Native apps using Hermes, but can translate to other environments given the required support from build tooling and VMs.

The plan is for React to feature-detect `getFunctionLocation`, call it on each component and encode the result in the component stack string, as well as expose it in the output of this parser.

## Test Plan

Tests, Flow

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugTools.js
+++ b/packages/react-debug-tools/src/ReactDebugTools.js
@@ -8,5 +8,6 @@
  */
 
 import {inspectHooks, inspectHooksOfFiber} from './ReactDebugHooks';
+import {parseErrorInfo} from './ReactErrorInfoParser';
 
-export {inspectHooks, inspectHooksOfFiber};
+export {inspectHooks, inspectHooksOfFiber, parseErrorInfo};
--- a/packages/react-debug-tools/src/ReactErrorInfoParser.js
+++ b/packages/react-debug-tools/src/ReactErrorInfoParser.js
@@ -1,0 +1,52 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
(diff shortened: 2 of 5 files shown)


gaearon:
We should put this into a separate package so it doesn’t increase the byte size. Maybe into react-debug-tools.

motiz88:
Moved `ReactErrorInfoParser` into `react-debug-tools`.

Issue #18014, opened 2020-02-11
Profiler flamegraph could show icon/emoji for the re-render reasons. Making it easier to scan.


bgirard:
Right now when working on a large project/component tree it can be difficult to understand why a large subtree is re-rendering. You click on each component and a reason: 1) Re-rendered because of the parent, 2) Re-rendered because a property changed, 3) Re-rendered because a hook changed. We don't get notified if a context value changed but ideally we would in the future.

If the flamegraph added icons we could perhaps spot patterns and more quickly identify what's going on when a large tree re-renders. For example state is changed (📚) at the top causing components to re-render via props (📋). Here's an example

```
📚Parent View
📋 TabStrip
📋 Page
📚Content
📋 Story
```

Here we can quickly spot that the state is changing in two component, and 3 components are changing because of prop. That might focus the investigation into why the state of Parent View / Content are changing which are likely causing the other props to change.

M-Izadmehr:
@bvaughn @bgirard,
I went on and added the tooltip summary of each fiber on hover, I feel it helps a lot for the overview, regardless, I still feel that emoji suggestion is also a good one, but the problem was that when I tried to implement it for the really small components in the chart it doesn't add any value. But we can still add it for the larger components in the chart.

bgirard:
I think the tooltip is a better idea overall. Your PR screenshot looks really nice!

> the problem was that when I tried to implement it for the really small components in the chart it doesn't add any value

Me and Brian discussed that. If the component is small you can zoom in. If it's still small then it's probably not one that we care about anyways. We'd still get the visualization for the large components that matter the most.

But with your context menus we probably don't need this at all!

bvaughn:
Now that we have tooltips, I don't think these icons are necessary anymore so I'm going to close this issue out. (I'm not a huge fan of the icon idea anyway, although the underlying need of getting info without zooming in makes sense!)

PR #18048, opened 2020-02-16
Add fiber summary tooltip to devtools profiling

M-Izadmehr:
## Summary
The purpose of this MR is to address #18014 (when working on a large project/component tree it can be difficult to understand why a large subtree is re-rendering, as a result adding a tooltip fo devtools profiler, which makes the process of debugging and inspecting elements easier.)

As a result, in really small components we don't need to click on them to see the details in sidebar, now it is possible to see the summary in the tooltip.

## Results

ezgif com-video-to-gif (1)
ezgif com-video-to-gif


bvaughn:
Screencast of this looks great!

I'll review it soon.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js
```diff
@@ -52,6 +77,8 @@ function CommitRankedListItem({data, index, style}: Props) {
       key={node.id}
       label={node.label}
       onClick={handleClick}
+      onMouseOver={event => handleMouseOver(event, node)}
+      onMouseOut={event => handleMouseOut(event)}
```
strdr4605: Maybe
```suggestion
      onMouseOut={handleMouseOut}
```

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js
```diff
@@ -52,6 +77,8 @@ function CommitRankedListItem({data, index, style}: Props) {
       key={node.id}
       label={node.label}
       onClick={handleClick}
+      onMouseOver={event => handleMouseOver(event, node)}
```
strdr4605: As `node` can be accessed from closure, maybe 
```suggestion
      onMouseOver={handleMouseOver}
```

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js
```diff
@@ -38,6 +46,23 @@ function CommitRankedListItem({data, index, style}: Props) {
...
   );
 
+  const handleMouseOver = useCallback(
+    (event: SyntheticMouseEvent<*>, nodeData: ChartNodeType) => {
+      const {id, name} = nodeData;
+      event.stopPropagation();
+      hoverFiber({id, name});
+    },
+    [hoverFiber],
+  );
```
strdr4605: As`nodeData` can be accessed from closure, maybe
```js
  const handleMouseOver = useCallback(
    (event: SyntheticMouseEvent<*>) => {
      const {id, name} = node;
      event.stopPropagation();
      hoverFiber({id, name});
    },
    [hoverFiber, node],
  );
```

packages/react-devtools-shared/src/devtools/views/Profiler/HoveredFiberInfo.js
```diff
@@ -0,0 +1,78 @@
...
+      );
+
+      listItems.push(
+        <div key={commitIndex} className={styles.CurrentCommit}>
+          {formatTime(timestamp)}s for {formatDuration(duration)}ms
+        </div>,
+      );
+      break;
+    }
+  }
```
strdr4605: Looks like this loop only finds one element, so why the `listItems` array is needed?
Maybe something like this?
```js
  const item = useMemo(() => {
    const commitIndices = profilingCache.getFiberCommits({
      fiberID: ((id: any): number),
      rootID: ((rootID: any): number),
    });
    const isSelectedCommitIndexInCommitIndices = commitIndices.some((commitIndex) => selectedCommitIndex === commitIndex);
    if (isSelectedCommitIndexInCommitIndices) {
      const {duration, timestamp} = profilerStore.getCommitData(
        ((rootID: any): number),
        selectedCommitIndex,
      );
      return (
        <div key={commitIndex} className={styles.CurrentCommit}>
          {formatTime(timestamp)}s for {formatDuration(duration)}ms
        </div>
      );
    } else {
      return null;
    }
  }, [selectedCommitIndex]);
```
Also `item` is not a good name so should think on renaming it.

bvaughn:
Need to run into a meeting now so a full PR is pending, but I noticed the tooltip position flickers when I did a quick test, e.g.:
flickerKapture 2020-02-18 at 10 25 27

packages/react-devtools-shared/src/devtools/views/Components/Tooltip.js
```diff
@@ -0,0 +1,51 @@
...
+import styles from './Tooltip.css';
+
+const initialTooltipState = {height: 0, mouseX: 0, mouseY: 0, width: 0};
+
+export default function Tooltip({children, label}: any) {
+  const containerRef = useRef(null);
+  const [tooltipState, setTooltipState] = useState(initialTooltipState);
+  const tooltipRef = useSmartTooltip(tooltipState);
+
+  const onMouseMove = useCallback(
```
bvaughn: `useCallback` doesn't do anything in this case, so we can remove it :)

packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.js
```diff
@@ -41,12 +45,13 @@ export default function ChartNode({
...
   return (
     <g className={styles.Group} transform={`translate(${x},${y})`}>
-      <title>{label}</title>
       <rect
         width={width}
         height={height}
         fill={color}
         onClick={onClick}
+        onMouseOver={onMouseOver}
+        onMouseOut={onMouseOut}
```
bvaughn: nit: Not sure it matters in this case (SVG rect) but in general, I'd prefer using mouse-enter and mouse-leave rather than mouse-over and mouse-out

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraphListItem.js
```diff
@@ -104,6 +124,8 @@ function CommitFlamegraphListItem({data, index, style}: Props) {
             key={id}
             label={label}
             onClick={event => handleClick(event, id, name)}
+            onMouseOver={event => handleMouseOver(event, chartNode)}
+            onMouseOut={event => handleMouseOut(event)}
```
bvaughn: Same nit as above (enter & leave)

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraphListItem.js
```diff
@@ -43,6 +46,23 @@ function CommitFlamegraphListItem({data, index, style}: Props) {
     [selectFiber],
   );
 
+  const handleMouseOver = useCallback(
```
bvaughn: These `useCallback` for mouse handlers can be removed :)

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js
```diff
@@ -38,6 +46,23 @@ function CommitRankedListItem({data, index, style}: Props) {
     [node, selectFiber],
   );
 
+  const handleMouseOver = useCallback(
```
bvaughn: `useCallback` can be removed here too.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js
```diff
@@ -38,6 +46,23 @@ function CommitRankedListItem({data, index, style}: Props) {
...
+    (event: SyntheticMouseEvent<*>, nodeData: ChartNodeType) => {
+      const {id, name} = nodeData;
+      event.stopPropagation();
+      hoverFiber({id, name});
+    },
+    [hoverFiber],
+  );
+
+  const handleMouseOut = useCallback(
```
bvaughn: And here too :)

packages/react-devtools-shared/src/devtools/views/Profiler/HoveredFiberInfo.js
```diff
@@ -0,0 +1,78 @@
...
+  let i = 0;
+  for (i = 0; i < commitIndices.length; i++) {
+    const commitIndex = commitIndices[i];
+    if (selectedCommitIndex === commitIndex) {
+      const {duration, timestamp} = profilerStore.getCommitData(
+        ((rootID: any): number),
+        commitIndex,
+      );
+
+      listItems.push(
```
bvaughn: This doesn't need to be an array since it will only ever find 0 or 1 matches.

packages/react-devtools-shared/src/devtools/views/Profiler/HoveredFiberInfo.js
```diff
@@ -0,0 +1,78 @@
...
+      <div className={styles.Content}>
+        <ProfilerWhatChanged fiberID={((id: any): number)} />
+        {listItems.length > 0 && (
+          <Fragment>
+            <label className={styles.Label}>Rendered at</label>: {listItems}
+          </Fragment>
+        )}
+        {listItems.length === 0 && (
+          <div>Did not render during this profiling session.</div>
+        )}
```
bvaughn: If you replaced the list above with a single item, this could be re-written more simply as e.g. `{matchingListItem || <div>Did not render during this profiling session.</div>}`

M-Izadmehr:
@bvaughn @strdr4605 Thanks for great comments, I will try to check them and also the flicker issue and get back to you:)

bvaughn:
Thank YOU for the great work so far :) Much appreciated

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js
M-Izadmehr: Makes sense, I removed it as param and used the value from closure with a simple callback

packages/react-devtools-shared/src/devtools/views/Profiler/HoveredFiberInfo.js
M-Izadmehr: Fixed

packages/react-devtools-shared/src/devtools/views/Components/Tooltip.js
M-Izadmehr: Done

packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.js
M-Izadmehr: Yeah, agree, I changed them to `onMouseLeave/Enter`, I think we also don't need to `stopPropagation` anymore, I removed them too:)

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraphListItem.js
M-Izadmehr: Done

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js
M-Izadmehr: Done.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js
M-Izadmehr: Done.

packages/react-devtools-shared/src/devtools/views/Profiler/HoveredFiberInfo.js
M-Izadmehr: I had doubts about whether to change it to filter, or keep it as for, as I checked I saw that we are banning `for...in` in eslints, also I never saw a case where we use `filter/map`, is there a particular reason for that? :)

M-Izadmehr: Done by the way!

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraphListItem.js
M-Izadmehr: Done!

M-Izadmehr:
Yeah, good catch, the flicker was due to my last commit, initially, I was doing `onMouseMove` on every movement of the mouse and it caused two updates (1. due to setting state of tooltip position, 2. sync useLayoutEffect), which would result in re-render of the whole graph on change. The cause of this issue was that in my last commit I added a condition to *"only listen for movements only if there is an element hovered"* (which was not the best decision:) )
Now, I removed the `useEffectLayout`, and also `tooltipState`, and as a result, the `children` (the whole graph) never re-renders because of mouse movement, and flicker is also fixed:)
ezgif com-video-to-gif (2)

bvaughn:
Ooh, thanks for the quick turnaround! Let me take another look!

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraphListItem.js
```diff
@@ -43,6 +46,15 @@ function CommitFlamegraphListItem({data, index, style}: Props) {
...
   );
 
+  const handleMouseEnter = (nodeData: ChartNodeType) => {
+    const {id, name} = nodeData;
+    hoverFiber({id, name});
+  };
+
+  const handleMouseLeave = () => {
+    hoverFiber(null);
+  };
```
strdr4605: Looks like this is same functions as in https://github.com/facebook/react/blob/0bd6e07a970fc711855b184ca222ee6d1897ade3/packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js#L48

So for consistency use variant in https://github.com/facebook/react/blob/0bd6e07a970fc711855b184ca222ee6d1897ade3/packages/react-devtools-shared/src/devtools/views/Profiler/CommitRankedListItem.js#L48

M-Izadmehr: Thanks, I removed this useCallback too:)

bvaughn:

Thank you!

PR #18050, opened 2020-02-17
referee to Edge in condition

uxitten:
Just improved comment


threepointone:
We actually do show the message for Edgium (ie - the new Microsoft Edge browser), whose useragent passes this check because it spells it as 'Edg' and not 'Edge'. Besides, we've also recently published the extension for Edgium (https://microsoftedge.microsoft.com/addons/detail/gpphkfbcpidddadnkolkpfckpihlkkil). In any case, the comment is still valid. Thanks for the effort!

PR #18059, opened 2020-02-17
Add component location to component stacks under Hermes

motiz88:
NOTE: This is based on https://github.com/facebook/react/pull/18043 ([compare](https://github.com/motiz88/react/compare/error-info-parser...motiz88:error-info-with-source-location)).

## Summary

Prints the source location of each component in component stack traces. This is the location where the component is _defined_ (as opposed to _used_). It's obtained from a Hermes-specific API since there is no standard JS API to retrieve the location of an arbitrary function. With this, apps targeting Hermes will log component stacks that can be decoded using a source map.

## ~TODO~ Discussion topics

- [ ] Agree on serialisation format; currently it's an optional JSON object embedded in each individual stack trace line.
- [ ] Normalise terminology and update parameter/property names etc: `source` (element source location) vs `location` (component source location) is confusing.
- [x] Update as needed to match changes in https://github.com/facebook/react/pull/18043 (e.g. import the parser from `react-debug-tools` rather than from `ReactSharedInternals`).
- [ ] Decide whether we need to implement special support for drilling through user-defined HOCs (_probably_ not).
- [x] Complete the implementation in DevTools console wrapper.

## Test Plan

Added tests

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugTools.js
+++ b/packages/react-debug-tools/src/ReactDebugTools.js
@@ -8,5 +8,6 @@
  */
 
 import {inspectHooks, inspectHooksOfFiber} from './ReactDebugHooks';
+import {parseErrorInfo} from './ReactErrorInfoParser';
 
-export {inspectHooks, inspectHooksOfFiber};
+export {inspectHooks, inspectHooksOfFiber, parseErrorInfo};
--- a/packages/react-debug-tools/src/ReactErrorInfoParser.js
+++ b/packages/react-debug-tools/src/ReactErrorInfoParser.js
@@ -7,12 +7,13 @@
  * @flow
  */
 
-const FRAME_RE = /\n {4}in (.+?)(?: \(created by (.+?)\)| \(at (.*)\:([0-9]+?)\))?$/gm;
+const FRAME_RE = /\n {4}in (.+?)( \{.*\})?(?: \(created by (.+?)\)| \(at (.*)\:([0-9]+?)\))?$/gm;
 
 type DescribedFiber = {
   +name: string,
   +source: ?{+fileName: string, +lineNumber: number, ...},
   +owner: ?{+name: string, ...},
+  +location: mixed,
   ...
 };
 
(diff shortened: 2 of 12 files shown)


gaearon:
Do we have a plan here? Anything you're blocked on?

motiz88:
I was hoping to start a discussion and get big-picture feedback on both this and the base PR #18043, then go implement any requested changes along with the known TODOs.

gaearon:
In https://github.com/facebook/react/pull/18043, we need to move it to a separate package and remove it from the React object. It shouldn't bloat the bundle size.

In this PR, let's make `describeComponentFrame` take the type instead of a separate module and separate call everywhere? In general we want to reduce the indirection and the amount of "helper" files.

Then I think we get CI passing and do another review.

motiz88:
I'm rebasing and completing this now.

One issue is that there doesn't seem to be a way to import code from `shared/...` in `react-devtools-shared/...`. I need this to amend DevTools's fork of [describeComponentFrame](https://github.com/facebook/react/blob/master/packages/react-devtools-shared/src/backend/describeComponentFrame.js) with a call to getComponentLocation which lives under `shared/`. Any thoughts on the proper way to do this? @bvaughn maybe?

bvaughn:
> One issue is that there doesn't seem to be a way to import code from `shared/...` in `react-devtools-shared/...`

Importing from non-DevTools directories works fine for the most part. I think it's just a matter of Jest configuration module mapping. This should unblock you:

```diff
diff --git a/scripts/jest/config.build-devtools.js b/scripts/jest/config.build-devtools.js
index 9c8501486..30c95c4b7 100644
--- a/scripts/jest/config.build-devtools.js
+++ b/scripts/jest/config.build-devtools.js
@@ -43,6 +43,9 @@ packages.forEach(name => {
   ] = `<rootDir>/build/node_modules/${name}/$1`;
 });
 
+// Shared isn't a built package; it's a special case.
+moduleNameMapper['^shared\/([^\/]+)$'] = '<rootDir>/packages/shared/$1';
+
 module.exports = Object.assign({}, baseConfig, {
   // Redirect imports to the compiled bundles
   moduleNameMapper,
```

PR #18064, opened 2020-02-18
[react-interactions] Ensure onBeforeBlur fires for hideInstance

trueadm:
As part of the React Flare system, we have the notion of a React specific event called `onBeforeBlur`. This is meant to triggered just before a managed node of React gets removed or hidden from the UI. We already support the case where this fires before we removed the node, but we didn't have a codepath or support for when this can occur during Suspense. During Suspense, we `display: none` the content when showing the fallback, which means that the React Flare focus systems fail to know of the fact that `blur` is about to be triggered, which is required to properly handle focus management logic. This PR adds that code path in, so `onBeforeBlur` correctly fires during Suspense too.

On a separate node, we should move this logic and other existing `onBeforeBlur` logic out of the Flare flag sometime and make it its own core feature at some point.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -598,7 +598,7 @@ function instanceContainsElem(instance: Instance, element: HTMLElement) {
 
 export function hideInstance(instance: Instance): void {
   // Ensure we trigger `onBeforeBlur` if the active focused elment
-  // is ether the instance of a child of the instance. We need
+  // is ether the instance of a child or the instance. We need
   // to traverse the Fiber tree here rather than use node.contains()
   // as the child node might be inside a Portal.
   if (enableDeprecatedFlareAPI && selectionInformation) {
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -58,6 +58,17 @@ import {
 } from '../events/DeprecatedDOMEventResponderSystem';
 import {retryIfBlockedOn} from '../events/ReactDOMEventReplaying';
 
+import {
+  enableSuspenseServerRenderer,
+  enableDeprecatedFlareAPI,
+  enableFundamentalAPI,
+} from 'shared/ReactFeatureFlags';
+import {HostComponent} from 'shared/ReactWorkTags';
+import {
+  RESPONDER_EVENT_SYSTEM,
+  IS_PASSIVE,
+} from 'legacy-events/EventSystemFlags';
+
 export type Type = string;
 export type Props = {
   autoFocus?: boolean,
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -584,7 +585,28 @@ export function clearSuspenseBoundaryFromContainer(
   retryIfBlockedOn(container);
 }
 
+function instanceContainsElem(instance: Instance, element: HTMLElement) {
+  let fiber = getClosestInstanceFromNode(element);
+  while (fiber !== null) {
+    if (fiber.tag === HostComponent && fiber.stateNode === element) {
```
trueadm: It might be a bit strange that we expose Fiber details in the host config, but we do this in other places.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -584,7 +585,28 @@ export function clearSuspenseBoundaryFromContainer(
...
+      return true;
+    }
+    fiber = fiber.return;
+  }
+  return false;
+}
+
 export function hideInstance(instance: Instance): void {
+  // Ensure we trigger `onBeforeBlur` if the active focused elment
+  // is ether the instance of a child of the instance. We need
```
necolas: s/of/or

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -584,7 +585,28 @@ export function clearSuspenseBoundaryFromContainer(
...
+  // to traverse the Fiber tree here rather than use node.contains()
+  // as the child node might be inside a Portal.
+  if (enableDeprecatedFlareAPI && selectionInformation) {
+    const focusedElem = selectionInformation.focusedElem;
+    if (focusedElem !== null && instanceContainsElem(instance, focusedElem)) {
+      dispatchBeforeDetachedBlur(((focusedElem: any): HTMLElement));
```
necolas: Do we want to dispatch this event for the new event system too?

packages/react-interactions/events/src/dom/__tests__/FocusWithin-test.internal.js
```diff
@@ -336,6 +338,68 @@ describe.each(table)('FocusWithin responder', hasPointerEvents => {
         expect.objectContaining({isTargetAttached: false}),
       );
     });
+
+    it.experimental(
```
necolas: These kinds of unit tests that involve the scheduler don't feel like they belong in the responders to me. Doesn't really matter right now since all this is going away eventually. But it makes me think about how we decide where the boundaries lie, as did your PR that rearranged the DOM event plugin system and its tests for the better

packages/react-dom/src/client/ReactDOMHostConfig.js
trueadm: We will eventually.

packages/react-interactions/events/src/dom/__tests__/FocusWithin-test.internal.js
trueadm: Yeah, I thought that. It's best if this code moved out of Flare soon.

PR #18070, opened 2020-02-18
fix(react-devtools-shared): useDebugValue with complex types

### LINKED ISSUE #17955: Bug(devtools): complex types ignored sometimes in useDebugValue
React version: 0.0.0-241c4467e (current `next`) and 16.12.0
Devtools: 4.4.0 (1/3/2020). [...] Created from revision f749045a5 on 1/3/2020.

## Steps To Reproduce

1. Visit https://codesandbox.io/s/usedebugvalue-complex-types-3x877
2. Inspect `App` Component
3. `NoHooksAtAll` displays no hooks at all
4. `IgnoredComplexValue` displayes the custom hook but without a value
5. `NoHooksAtAll` displays the custom hook with its value

Link to code example:

## The current behavior
Complex values (and custom hooks) are sometimes not displayed

## The expected behavior

Implementation of components and hooks should not affect if these are displayed.

eps1lon:
Closes #17955

Using `react-devtools-shell`:
`master`:
Screenshot from 2020-03-09 23-01-20

`pr`:
Screenshot from 2020-03-09 23-00-55


bvaughn:
CI is failing on lint btw. Run `yarn prettier-all`.

Not sure what the question is atm on this PR?

eps1lon:
> Is there a question (for me) on this PR at the moment or...? I'm not sure.

It looks like the test is already working as expected. I'm asking how I can test the actual react components that render this as it seems like the patch in the OP is the actual fix.

scripts/jest/config.build-devtools.js
```diff
@@ -43,6 +43,9 @@ packages.forEach(name => {
   ] = `<rootDir>/build/node_modules/${name}/$1`;
 });
 
+// https://jestjs.io/docs/en/webpack#mocking-css-modules
+moduleNameMapper['\\.(css|less)$'] = 'identity-obj-proxy';
```
eps1lon: There's currently exists a `cssMock.js` but it doesn't seem to be used. With `identity-obj-proxy` 
> Then all your className lookups on the styles object will be returned as-is (e.g., styles.foobar === 'foobar'). 

-- https://jestjs.io/docs/en/webpack#mocking-css-modules

packages/react-devtools-shared/src/__tests__/HooksTree-test.js
```diff
@@ -0,0 +1,92 @@
...
+          <InspectedElementContextController>
+            {children}
+          </InspectedElementContextController>
+        </TreeContextController>
+      </StoreContext.Provider>
+    </BridgeContext.Provider>
+  );
+
+  it('shows the complex value of custom hooks with sub-hooks', () => {
+    // props from `InspectedElementContext display complex values of useDebugValue: DisplayedComplexValue 1` snapshot
```
eps1lon: To make it more explicit I could merge the added test from `inspectedElementContext` and instead of serializing the tree directly pass `inspectedElement.hooks` to the tested `HooksTree`. A core member should decide if this coupling should be explicit or not.

bvaughn:
>> Is there a question (for me) on this PR at the moment or...? I'm not sure.
>
> It looks like the test is already working as expected. I'm asking how I can test the actual react components that render this as it seems like the patch in the OP is the actual fix.

You can use the shell? You can add any hooks you want into the example app code the shell uses to verify the behavior in a real browser.

scripts/jest/config.build-devtools.js
bvaughn: Was this causing a problem?

eps1lon: In the new test I'm using HooksTree which imports some css file. Without mocking it, jest throws.

eps1lon:
> You can use the shell? You can add any hooks you want into the example app code the shell uses to verify the behavior in a real browser.

I'll try that. Though the current test makes me pretty confident that this is fixed. I would test it with the shell (hopefully shareable) if this would help you understand the fix.

scripts/jest/config.build-devtools.js
bvaughn: Ah, I see. I don't think we should import a DevTools view in a test like that. I'll comment above.

packages/react-devtools-shared/src/__tests__/HooksTree-test.js
```diff
@@ -0,0 +1,92 @@
...
+
+    const container = document.createElement('div');
+    ReactDOM.render(
+      <Contexts>
+        <HooksTree
+          hooks={[
```
bvaughn: What is this test meant to be testing? I'm not sure I see the value it adds.

Hard-coding the hooks metadata doesn't seem like it's a very comprehensive test. I think the existing pattern of testing via the contexts is a better end-to-end coverage. Less brittle. :)

packages/react-devtools-shared/src/__tests__/HooksTree-test.js
```diff
@@ -0,0 +1,92 @@
...
+          ]}
+        />
+      </Contexts>,
+      container,
+    );
+
+    const hook = container.querySelector('.NameValueRow');
+    // it's actually DebuggableHook: {foo:2} but the first colon is added by css
+    // which isn't available in the test
+    expect(hook.textContent).toEqual('DebuggableHook{foo: 2}');
```
bvaughn: This test seems kind of brittle.

I think we can probably just skip `HooksTree-test.js` :smile:

packages/react-devtools-shared/src/__tests__/HooksTree-test.js
eps1lon: That was my suggestion about. Directly injecting the tree.hooks into the prop. I couldn't find an existing e2e test for the react-devtools view.

packages/react-devtools-shared/src/__tests__/HooksTree-test.js
eps1lon: That's the only test that is failing though Brian. HooksTree contains the bug. Why do you think the actual textual content is brittle? If you have an issue with the selector then I could add a `data-testid`. Since the relevant markup isn't semantic at all I don't see other opportunities

packages/react-devtools-shared/src/__tests__/HooksTree-test.js
bvaughn: I don't think there's a lot of value in testing the view components.  DevTools tests the underlying business logic on the backend (and frontend via the contexts).

Let's remove this test :)

packages/react-devtools-shared/src/__tests__/HooksTree-test.js
bvaughn: Hard-coding a JSON object in a test is brittle. That object should be created by the backend. e.g. What happens if the backend structure changes in a way that would break things, but your test is written with the hard-coded JSON object?

I don't have time to dig deep on this right now unfortunately, so my feedback so far has been pretty shallow. In a few days I might have more time, but I need to wrap up some RFCs first.

If you disagree with my PR feedback so far, then let's just set it down for a while until I have time to really dig in.

eps1lon:
- removed view test
- keep backend test (passing on `master` as well) as suggested in https://github.com/facebook/react/pull/18070#discussion_r382142439
- verified manually locally using `react-devtools-shell` (I had some trouble understanding that this isn't some sort of "shell" or "terminal" but an actual app or fixture that can be used for (manual) testing)

packages/react-devtools-shared/src/__tests__/HooksTree-test.js
eps1lon: Ok I think I got the hang of how I can verify my devtools changes. This was a bit frustrating when starting out so apologies if this sounded more brisk than intended.

bvaughn: No worries.

bvaughn:
I'm confused. Testing `useDebugValue` with "complex types" like objects already seems to work in master in both the shell:
Screen Shot 2020-03-12 at 9 22 20 AM

...and the actual extension:
Screen Shot 2020-03-12 at 9 23 18 AM

I can also revert the change to `HooksTree` and newly added test still passes.

eps1lon:
> I can also revert the change to HooksTree and newly added test still passes.

There are no tests for HooksTree. There never were. The newly added test passes on master as I said because the problem wasn't in the backend but the view. 

> ...and the actual extension:
> Screen Shot 2020-03-12 at 9 23 18 AM

That is what we not want. We want it to be displayed without having to expand.



> I'm confused. Testing `useDebugValue` with "complex types" like objects already seems to work in master in both the shell:
> Screen Shot 2020-03-12 at 9 22 20 AM

Not sure where this screenshot is taken from.

The original post has a screenshot of before and after. Maybe this explains it better what was changed?

bvaughn:
> There are no tests for HooksTree. There never were. The newly added test passes on master as I said because the problem wasn't in the backend but the view.

Right. Maybe I just don't understand the purpose of the newly added test.

> That is what we not want. We want it to be displayed without having to expand.

Ah, your most recent response clarifies something for me.

I think you want the stringified version of the object to display when the hook is collapsed. The current design was intentional, to allow you to expand a complex value. I understand why you'd prefer the collapsed view too. I think a compromise between them might be best.

Let me revisit your PR to think about that.

packages/react-devtools-shared/src/devtools/views/Components/HooksTree.js
```diff
@@ -202,17 +202,11 @@ function HookView({canEditHooks, hook, id, inspectPath, path}: HookViewProps) {
               className={name !== '' ? styles.Name : styles.NameAnonymous}>
               {name || 'Anonymous'}
             </span>
+            <span className={styles.Value} onClick={toggleIsOpen}>
```
bvaughn: This should only be shown if the hook is collapsed. Otherwise, the `<KeyValue>` component should be rendered as it enables the complex value to be expanded and inspected.

So something like this:
```js
<div className={styles.Hook}>
  <div ref={contextMenuTriggerRef} className={styles.NameValueRow}>
    <ExpandCollapseToggle isOpen={isOpen} setIsOpen={setIsOpen} />
    <span
      onClick={toggleIsOpen}
      className={name !== '' ? styles.Name : styles.NameAnonymous}>
      {name || 'Anonymous'}
    </span>
    <span className={styles.Value} onClick={toggleIsOpen}>
      {isOpen || getMetaValueLabel(value)}
    </span>
  </div>
  <div className={styles.Children} hidden={!isOpen}>
    <KeyValue 
      depth={1} 
      alphaSort={false} 
      inspectPath={inspectPath} 
      name="DebugValue" 
      path={path.concat(['value'])} 
      pathRoot="hooks"  
      value={value} 
    />
    {subHooksView}
  </div>
</div>
```

bvaughn: To be clear, the label should work like this when toggling:
Custom hook label

eps1lon: Yep that's what I wanted. I guess in this PR the debug value is still displayed on the same line as "CustomObject:" when expanded?

bvaughn: Yeah, the string preview just stayed visible but the ability to expand the complex value below was gone. I added it back in 32bac9e :smile:

PR #18073, opened 2020-02-19
Add modern event system flag + rename legacy plugin module

trueadm:
This PR does four things:

- Cleans up `ReactBrowserEventEmitter-test.internal.js` a bit
- Re-labels `listenToEvent` to `legacyListenToEvent`
- Renames the module `DOMEventPluginSystem` to `DOMLegacyEventPluginSystem`
- Adds the `enableModernEventSystem` flag

### DIFF
--- a/packages/react-dom/src/__tests__/ReactBrowserEventEmitter-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactBrowserEventEmitter-test.internal.js
@@ -14,7 +14,7 @@ let EventPluginRegistry;
 let React;
 let ReactDOM;
 let ReactDOMComponentTree;
-let DOMEventPluginSystem;
+let listenToEvent;
 let ReactDOMEventListener;
 let ReactTestUtils;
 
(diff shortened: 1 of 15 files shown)


packages/react-dom/src/__tests__/ReactBrowserEventEmitter-test.internal.js
```diff
@@ -66,7 +65,8 @@ describe('ReactBrowserEventEmitter', () => {
     React = require('react');
     ReactDOM = require('react-dom');
     ReactDOMComponentTree = require('../client/ReactDOMComponentTree');
-    DOMEventPluginSystem = require('../events/DOMEventPluginSystem');
+    listenToEvent = require('../events/DOMLegacyEventPluginSystem')
```
threepointone: Could you keep this named as legacyListenToEvent? 
```suggestion
    legacyListenToEvent = require('../events/DOMLegacyEventPluginSystem')
```

trueadm: @threepointone The `ReactBrowserEventEmitter-test.internal.js` change is because I fork that file in the follow up that uses the flag. Also, I'd rather add the flag in this PR as it feels more churny having another PR that adds it but doesn't use it (lots of file changes makes the diff harder to review).

PR #18074, opened 2020-02-19
Feature Flag for React.jsx' "spreading a key to jsx" warning

threepointone:
Adds a feature flag for when `React.jsx` warns you about spreading a key into jsx. It's false for all builds, except as a dynamic flag for fb/www.

I also included the component name in the warning.

### DIFF
--- a/packages/react/src/ReactElementValidator.js
+++ b/packages/react/src/ReactElementValidator.js
@@ -21,6 +21,7 @@ import {
   REACT_FRAGMENT_TYPE,
   REACT_ELEMENT_TYPE,
 } from 'shared/ReactSymbols';
+import {warnAboutSpreadingKeyToJSX} from 'shared/ReactFeatureFlags';
 import checkPropTypes from 'prop-types/checkPropTypes';
 
 import ReactCurrentOwner from './ReactCurrentOwner';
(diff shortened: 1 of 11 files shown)


packages/react/src/ReactElementValidator.js
```diff
@@ -365,13 +369,20 @@ export function jsxWithValidation(
     }
   }
 
-  if (hasOwnProperty.call(props, 'key')) {
```
threepointone: moved this inside the __DEV__ block. It wasn't being DCEed, so that gets us a few bytes back too.

packages/react/src/ReactElementValidator.js
```diff
@@ -365,13 +369,20 @@ export function jsxWithValidation(
...
-    if (__DEV__) {
-      console.error(
-        'React.jsx: Spreading a key to JSX is a deprecated pattern. ' +
-          'Explicitly pass a key after spreading props in your JSX call. ' +
-          'E.g. <ComponentName {...props} key={key} />',
-      );
+  if (__DEV__) {
+    if (warnAboutSpreadingKeyToJSX) {
+      if (
+        didWarnAboutSpreadingKey[name] !== true &&
```
threepointone: oh YIKES, lint or flow didn't catch this undefined var (because `window.name` is an actual thing?) 😱

packages/react/src/ReactElementValidator.js
```diff
@@ -365,13 +369,20 @@ export function jsxWithValidation(
...
+        if (didWarnAboutSpreadingKey[name] !== true) {
+          didWarnAboutSpreadingKey[name] = true;
+          console.error(
+            'React.jsx: Spreading a key to JSX is a deprecated pattern. ' +
+              'Explicitly pass a key after spreading props in your JSX call. ' +
+              'E.g. <%s {...props} key={key} />',
+            name,
+          );
+        }
+      }
```
threepointone: What I don't like about this implementation, is if a common component is used 'wrongly' in several places, it'll warn only for the 'first' one it encounters. A dev will fix it, run code again, and then see the next one. SO it'll become a game of whack-a-mole to fix all the usages. The cache should probably be based on component+location.

packages/react/src/ReactElementValidator.js
gaearon: Wait, wasn’t this whole file supposed to be dev only? Are you saying this function is used in prod?

threepointone: Oh hmm you're right. I only glanced at sizebot, probably not relevant.

packages/react/src/ReactElementValidator.js
threepointone: punting on the deduplication for now.

packages/react/src/ReactElementValidator.js
threepointone: followup - https://github.com/facebook/react/pull/18076

PR #18075, opened 2020-02-19
Add container to event listener signature

trueadm:
Note: This PR is a pre-requisite for the modern root event system.

For the new event system that uses root containers to register nodes. For this to work, we need a sane way of knowing what the container is, so we can properly find ancestors.

We could use `nativeEvent.currentTarget` but that seems fragile, as the `currentTarget` can change when the event bubbles, or might not be there if we bypass native events entirely (if we ever decided to add `onBeforeBlur` to the plugin event system).

Instead, we can bind the `container` to the dispatch function, as we do with other arguments. This then allows us to know the container correctly in all cases. Furthermore, we pass this to the event replaying logic so it can properly replay the event with the correct `container`.

### DIFF
--- a/packages/react-dom/src/events/ReactDOMEventListener.js
+++ b/packages/react-dom/src/events/ReactDOMEventListener.js
@@ -119,6 +119,7 @@ export function addResponderEventSystemEvent(
     null,
     ((topLevelType: any): DOMTopLevelEventType),
     eventFlags,
+    document,
   );
   if (passiveBrowserEventsSupported) {
     addEventCaptureListenerWithPassiveFlag(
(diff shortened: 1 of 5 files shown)


packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -289,6 +325,7 @@ export function attemptToDispatchEvent(
   topLevelType: DOMTopLevelEventType,
   eventSystemFlags: EventSystemFlags,
   nativeEvent: AnyNativeEvent,
+  container: Document | Element | Node,
```
gaearon: All of the other signatures have it before `nativeEvent` so this was a bit confusing.

gaearon: I guess the ones below don't. It would be nice to preserve a consistent order because it's easy to mix them up. Hopefully Flow would catch that.

trueadm: Flow can catch it, and it did, because I've hacked this file abount so much in the last few days and it's saved my ass a bunch :)

I can refine the order in a follow up though, you're right, consistency helps here.

PR #18076, opened 2020-02-19
add no-restricted-globals to eslint config

threepointone:
Our current lint config assumes a browser environment, which means it won't warn you if you use a variable like `name` without declaring it earlier. This PR imports the same one used by create-react-app, and enables it against our codebase.


gaearon:
Did you verify it works as expected?

threepointone:
Yup. Added console.log(name) to an arbit module, ran yarn linc - errored!

PR #18077, opened 2020-02-19
Remove Flare bundles from build

trueadm:
These Flare responders are not used internally anywhere, so we should remove them both from the build and internally.


gaearon:
Do we want to keep the code?

trueadm:
@gaearon For now we can keep the code as it's handy for catching any possible accidental regressions to Flare, whilst I change parts of the event system via tests.

PR #18080, opened 2020-02-19
Add guard around FocusWithin responder root events

trueadm:
We've run into an interesting problem internally that is likely due to a strange set of events from the recent `onBeforeBlur` change, causing `addRootEventTypes` to try and add the same event before removing it. This change should guard the change to prevent the invariant from occuring in these cases.

I'll try and make a repro test tomorrow and see if there's an alternative way to fix this.

### DIFF
--- a/packages/react-interactions/events/src/dom/Focus.js
+++ b/packages/react-interactions/events/src/dom/Focus.js
@@ -36,7 +36,7 @@ type FocusState = {
   isFocused: boolean,
   isFocusVisible: boolean,
   pointerType: PointerType,
-  ...
+  addedRootEvents?: boolean,
 };
 
 type FocusProps = {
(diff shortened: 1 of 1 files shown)


packages/react-interactions/events/src/dom/Focus.js
```diff
@@ -36,7 +36,7 @@ type FocusState = {
   isFocused: boolean,
   isFocusVisible: boolean,
   pointerType: PointerType,
-  ...
+  addedRootEvents?: boolean,
```
gaearon: Does it need to be optional?

gaearon: Let's make it just boolean.

trueadm: We don't use it in the Focus responder, only the FocusWithin responder. I didn't want to add more code to the state of that responder, so I opted to keep it optional for now. Breaking all the call-sites down by alternative state would make this a far more complex PR.

PR #18084, opened 2020-02-20
Add modern event plugin system fork that uses roots

trueadm:
This PR adds a fork of the existing event plugin system. The fork is behind the flag `enableModernEventSystem`. The modern event system differs from the legacy event system in that it attaches events to the root DOM node rather than the document DOM node.

To make this alteration simpler, there is a fork of the different event systems. The old legacy event system that attaches event listeners to the document is in its own module. The modern event system has its own module and the code paths for these should clearly be defined through the codebase. Switching the `enableModernEventSystem` to `true` should yield the same results in the tests (they should all pass as expected).

The key differences are the underlying implementation, which is reflected in some of the tests that have been forked/changed/skipped to respect the difference in event system implementation. Tests that previously passed becuase of the expectation that React listened to all events across the document no longer hold true, thus why they've been forked. I've also added tests that stress portals and embedded roots, to ensure the propagation rules remain consinstant compared to before.

We also no longer attach event listeners directly to DOM nodes, mainly for input/form/media elements. Instead we move this to the capture phase on the root, allowing us to still capture incoming events, but without having to attach as many DOM event listeners. This is clearly a breaking change and we'll need to see how this progresses during internal and external testing.

I'll be doing extensive testing with this build internally to find any issues that might arise. Going forward, I plan on making the flag GKed internally, so we can further find and squash issues as they arise.

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -56,6 +56,11 @@ import {
   TOP_PROGRESS,
   TOP_PLAYING,
 } from './DOMTopLevelEventTypes';
+import {getClosestInstanceFromNode} from '../client/ReactDOMComponentTree';
+import {DOCUMENT_NODE, COMMENT_NODE} from '../shared/HTMLNodeType';
+
+import {enableLegacyFBPrimerSupport} from 'shared/ReactFeatureFlags';
+import {HostRoot, HostPortal} from 'shared/ReactWorkTags';
 
 const capturePhaseEvents = new Set([
   TOP_FOCUS,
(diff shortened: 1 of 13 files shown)


packages/react-dom/src/__tests__/ReactDOMEventListener-test.js
```diff
@@ -189,9 +194,21 @@ describe('ReactDOMEventListener', () => {
...
-        expect(mock.mock.calls[1][0]).toBe('Child');
+        if (ReactFeatureFlags.enableModernEventSystem) {
+          // As we have two roots, it means we have two event listeners.
+          // This also means we enter the event batching phase twice,
+          // flushing the child to be 1.
+          // TODO: can we improve on this? microtasks seem to problematic
+          // (I tried that). We don't have any good way of knowing if
+          // another event will occur because another event handler
+          // might invoke stopPropagation() along the way. This means
+          // we over-flush.
```
philipp-spiess: I have played around with this before and also did not find a good way to batch events. Micro tasks are also problematic indeed because they are exhausted before the next event listener is fired (https://jsfiddle.net/gm7r5fkp/) and this behaviour is different when you call the event listeners programatically.

But batched mode (is this still a thing?) or concurrent mode should fix this, right?

trueadm: I spoke in depth to @sebmarkbage last nigtht about this and we settled on the fact we'll just have to put up with over-flushing till we can get around to tuning this in the future with some capability in the scheduler (which might require platform features being added).

Batched mode is still definitely a thing in concurrent mode though, and it's still something you'd ideally like to be automatic for the user when calling events.

trueadm:
https://github.com/facebook/react/pull/18179 is a pre-requistie for this PR, so please review that before this PR.

trueadm:
Closing this as most of the work has now been merged to master in smaller PRs.

PR #18086, opened 2020-02-20
Rename Chunks API to Blocks

sebmarkbage:
Sounds like this is the name we're going with. Blocks is a general abstract term used for many times of units similar to Components. Chunks is more often used for a very specific mechanism in CS. This also helps us distinguish it from other "chunking" implementation details.

My only hesitation is that Block sounds too big where they can actually be very small. Lego-blocks is a good analogy though since they're small. Components used to thought of as fairly large too but become smaller with more use.

Another possible name could be Part but this is often used to represents things lower in the hierarchy than Component in Bill of Materials etc. Where as Blocks are higher than Component. Part could be a name for "Host component".

(h/t @devknoll)


devknoll:
> Lego-blocks is a good analogy though since they're small

React Bricks? 😅

gaearon:
🧱

MatanBobi:
I maybe dropping in with no context but based on the explanation @sebmarkbage gave above, maybe Tiles would fit? As Tiles don't have any size reference (at least to me), the only problem with tiles is that I don't recall them being related to CS but bricks are also not related..

josephsavona:
Seems reasonable to me. Should we consider naming conflicts? Are there any other projects in this space that might be using the name? That said, "block" is pretty generic so it seems fine. 

Re the other suggestions - bricks or tiles - the naming doesn't work as well. "Blocks" makes me think of kids blocks which are all different shapes/sizes (like UI components) whereas bricks/tiles are usually all the same size or just a very limited set.

devknoll:
Bricks wasn't a (very) serious suggestion 😅In fact, that both @sebmarkbage and @josephsavona (and most other people, probably) called Lego's blocks instead of bricks (the name for a Lego piece) is probably a good hint that blocks might be more appropriate 😛 

My concern with those types of specific/skeuomorphic names like Brick or Tile is that they end up being imperfect analogies at best, and misleading at worst, where at least Block seems generic enough.

sophiebits:
Looking forward to buying the React-branded LEGO blocks.

also, chunk + block = chonk?

alexeyraspopov:
@sebmarkbage, reading through the tests, I'm trying to come up with a basic example of this API. Not that I have any suggestions or pushing on something, just wanted to ask if I understood the idea correctly:

```javascript
function UserQuery(userId) {
  // some sort of cache that can throw promise
  let user = userCache.get(userId);
  return { user };
}

function UserProfile(props, data) {
  return <JSX />;
}

let loadUserProfile = React.block(UserQuery, UserProfile);

function UserProfileView({ userId }) {
  let Profile = loadUserProfile(userId);
  return (
    <Suspense>
      <Profile ... />
    </Suspense>
  )
}
```

If this is approximately what the API is going to look like, I wonder if `useContext` can be used inside the query function or how limited the function is?

gaearon:
Yeah we'd like to support some form of context there. But not supported yet.

Issue #17318, opened 2019-11-08
useTransition - startTransition does not work on React.memo when is SimpleMemoComponent


salvoravida:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
useTransition - startTransition do not work on React.memo when is SimpleMemoComponent

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

**What is the expected behavior?**
startTransition work with SimpleMemoComponent

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

threepointone:
Please make a reproducing example on codesandbox, it’s not clear what you’re asking.

salvoravida:
Click on "With Transition Start"
OK -> https://codesandbox.io/s/test-redux-in-concurrent-mode-with-react-redux-salvoravidareact-redux-9lun7
KO -> https://codesandbox.io/s/test-redux-in-concurrent-mode-with-react-redux-salvoravidareact-redux-68pbd

only line 85 is changed

```jsx
const _MyComponent = () => {
  const status = useSelector(s => s.status);

  if (status === STATUS.PENDING) {
    throw promise;
  }
  return null;
};

//_MyComponent.defaultProps = {};
const MyComponent = React.memo(_MyComponent);
```

the problem maybe is here in react-dom 
```jsx
function updateMemoComponent(current$$1, workInProgress, Component, nextProps, updateExpirationTime, renderExpirationTime) {
  if (current$$1 === null) {
    var type = Component.type;

    if (isSimpleFunctionComponent(type) && Component.compare === null && // SimpleMemoComponent codepath doesn't resolve outer props either.
    Component.defaultProps === undefined) {
      var resolvedType = type;

      {
        resolvedType = resolveFunctionForHotReloading(type);
      } // If this is a plain function component without default props,
      // and with only the default shallow comparison, we upgrade it
      // to a SimpleMemoComponent to allow fast path updates.


      workInProgress.tag = SimpleMemoComponent;
      workInProgress.type = resolvedType;

      {
        validateFunctionComponentInDev(workInProgress, type);
      }

[...]

Issue #17028, opened 2019-10-07
[Concurrent mode] Scheduled updates are lost when parent state update happens


albertogasparin:
**Do you want to request a *feature* or report a *bug*?**
Probably a bug

**What is the current behavior?**
When in concurrent mode (render with `unstable_createRoot`), triggering a state change via an event (eg click) on the parent makes React loose the scheduled updates in the children. See:

https://codesandbox.io/s/react-concurrent-state-bug-sp6zq

Not an expert on concurrent mode here, was just playing around with it. Noticed that when some interactions happen the work (probably moved to the other tree) gets lost if the parent re-renders. You can see by clicking anywhere in the sandbox page vs clicking on the button that changes some local state. But that is just my educated guess on what might be happening.

**What is the expected behavior?**

State updates that are scheduled in children components should happen and the DOM value should reflect the state.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

react: 16.10.2
react-dom: 16.10.2

dai-shi:
👍 Linking my comment https://github.com/facebook/react/issues/16396#issuecomment-545003364. I thought it was an issue with use-subscription back then, but it turns out to be a broader issue.

dai-shi:
Hey, you might want to try with the latest experimental channel.
In [my tool](https://github.com/dai-shi/will-this-react-global-state-work-in-concurrent-mode), `0.0.0-experimental-f42431abe` fixes some cases. But not all.

acdlite:
Thanks for the bug report!

I believe this was fixed by https://github.com/facebook/react/pull/18091

You can confirm using `react@0.0.0-experimental-ea6ed3dbb` and `react-dom@react@0.0.0-experimental-ea6ed3dbb`.

I'm going to close this issue, but if the bug persists please comment and we'll reopen.

PR #18091, opened 2020-02-20
Bugfix: `memo` drops lower pri updates on bail out

acdlite:
Fixes a bug where lower priority updates on a components wrapped with `memo` are sometimes left dangling in the queue without ever being processed, if they are preceded by a higher priority bailout.

Cause
-----

The pending update priority field is cleared at the beginning of `beginWork`. If there is remaining work at a lower priority level, it's expected that it will be accumulated on the work-in-progress fiber during the begin phase.

There's an exception where this assumption doesn't hold: SimpleMemoComponent contains a bailout that occurs *before* the component is evaluated and the update queues are processed, which means we don't accumulate the next priority level. When we complete the fiber, the work loop is left to believe that there's no remaining work.

Mitigation
----------

Since this only happens in a single case, a late bailout in SimpleMemoComponent, I've mitigated the bug in that code path by restoring the original update priority from the current fiber.

This same case does not apply to MemoComponent, because MemoComponent fibers do not contain hooks or update queues; rather, they wrap around an inner fiber that may contain those. However, I've added a test case for MemoComponent to protect against a possible future regression.

Possible next steps
-------------------

We should consider moving the update priority assignment in `beginWork` out of the common path and into each branch, to avoid similar bugs in the future.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -523,6 +523,20 @@ function updateSimpleMemoComponent(
     ) {
       didReceiveUpdate = false;
       if (updateExpirationTime < renderExpirationTime) {
+        // The pending update priority was cleared at the beginning of
+        // beginWork. We're about to bail out, but there might be additional
+        // updates at a lower priority. Usually, the priority level of the
+        // remaining updates is accumlated during the evaluation of the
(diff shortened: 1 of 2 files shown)


dai-shi:
Hi, possibly related with #17318 (and #17314 and #17028)?

acdlite:
@dai-shi Yes probably. I'll confirm in a sec.

PR #18094, opened 2020-02-21
[WIP] Move Everything to ES Modules in Builds

sebmarkbage:
It turns out that our loose reliance on CommonJS isn't really letting Rollup do its best work. It's also covering up a bunch of quirks that cause other issues and this ends up leaking. We really only intended it to be used to suck in requires into UMD builds. Unfortunately it also means that we're actually inlining in a bunch of builds accidentally too.

The usage of CommonJS at the export level also covers up a bunch of issues by converting it first into an intermediate layer of CommonJS and then back again.

I want to get to a point where we can remove the CommonJS plugin from Rollup. I was hoping we could just fork the object-assign and prop-types ones because 1) both of these are kind of frozen and we should ideally get rid of the dep in the next version 2) umds are likely to get dropped soon too anyway. Unfortunately react-art also supports UMD, for some reason, and it's harder to fork all of art. I might end up keeping the CommonJS plugin but only for ART for this reason until we can kill those UMD bundles.

We should also start preparing for actually exporting ES modules. Right now we're not very rigid with how we export things. We should be always exporting a default for now since that gives us the best CommonJS output. We can't start exporting the esModule flag since that would start breaking things. 
(Although I settled on just doing that for react-interactions because it was easier and it's FB only.)

However, this means that importing named imports etc. don't work. Including in our own code. Like DevTools. So we'll need to fix that up too.

There's a bunch to do here and I'm really just getting started.

### DIFF
--- a/packages/react-art/index.js
+++ b/packages/react-art/index.js
@@ -9,6 +9,28 @@
 
 'use strict';
 
-const ReactART = require('./src/ReactART');
+import {
+  ClippingRectangle,
+  Group,
+  Shape,
+  Path,
+  LinearGradient,
+  Pattern,
+  RadialGradient,
+  Surface,
+  Text,
+  Transform,
+} from './src/ReactART';
 
-module.exports = ReactART;
+export default {
+  ClippingRectangle,
+  Group,
+  Shape,
+  Path,
+  LinearGradient,
+  Pattern,
+  RadialGradient,
+  Surface,
+  Text,
+  Transform,
+};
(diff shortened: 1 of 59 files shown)


packages/react-art/index.js
```diff
@@ -9,6 +9,28 @@
 
 'use strict';
```
TrySound: Maybe remove use strict everywhere?

sebmarkbage:
Done in other PRs.

PR #18096, opened 2020-02-21
Don't warn about unmounted updates if pending passive unmount

bvaughn:
I recently landed a change to the timing of passive effect cleanup functions during unmount (see #17925). This change defers flushing of passive effects for unmounted components until later (whenever we next flush pending passive effects).

Since this change increases the likelihood of a (not actionable) state update warning for unmounted components, I've suppressed that warning for Fibers that have scheduled passive effect unmounts pending.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -18,6 +18,7 @@ import type {Effect as HookEffect} from './ReactFiberHooks';
 
 import {
   warnAboutDeprecatedLifecycles,
+  deferPassiveEffectCleanupDuringUnmount,
   runAllPassiveEffectDestroysBeforeCreates,
   enableUserTimingAPI,
   enableSuspenseServerRenderer,
@@ -2687,6 +2688,18 @@ function warnAboutUpdateOnUnmountedFiberInDEV(fiber) {
       // Only warn for user-defined components, not internal ones like Suspense.
       return;
     }
+
+    if (
+      deferPassiveEffectCleanupDuringUnmount &&
+      runAllPassiveEffectDestroysBeforeCreates
+    ) {
+      // If there are pending passive effects unmounts for this Fiber,
+      // we can assume that they would have prevented this update.
+      if (pendingPassiveHookEffectsUnmount.indexOf(fiber) >= 0) {
+        return;
+      }
+    }
+
     // We show the whole stack but dedupe on the top component's name because
     // the problematic code almost always lies inside that component.
     const componentName = getComponentName(fiber.type) || 'ReactComponent';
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2687,6 +2688,18 @@ function warnAboutUpdateOnUnmountedFiberInDEV(fiber) {
...
+      deferPassiveEffectCleanupDuringUnmount &&
+      runAllPassiveEffectDestroysBeforeCreates
+    ) {
+      // If there are pending passive effects unmounts for this Fiber,
+      // we can assume that they would have prevented this update.
+      if (pendingPassiveHookEffectsUnmount.includes(fiber)) {
+        return;
+      }
+    }
+
```
threepointone: odd, syntax highlighting seems to be broken for this file 
image

bvaughn: Huh. Yeah for me me too.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2687,6 +2688,18 @@ function warnAboutUpdateOnUnmountedFiberInDEV(fiber) {
...
       return;
     }
+
+    if (
+      deferPassiveEffectCleanupDuringUnmount &&
+      runAllPassiveEffectDestroysBeforeCreates
+    ) {
+      // If there are pending passive effects unmounts for this Fiber,
+      // we can assume that they would have prevented this update.
+      if (pendingPassiveHookEffectsUnmount.includes(fiber)) {
```
acdlite: Nit: `includes` is a new-ish API, so use `indexof` or something instead so this doesn't break.

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -1144,6 +1144,90 @@ function loadModules({
...
+              // Simulate an XHR completing, which will cause a state update-
+              // but should not log a warning.
+              completePendingRequest();
+
+              ReactNoop.flushPassiveEffects();
+              expect(Scheduler).toHaveYielded(['passive destroy']);
+            });
+          });
+
+          it('still warns about state updates for unmounted components with no pending passive unmounts', () => {
```
acdlite: Can you add a test for when there's a pending unmount on a different component?

bvaughn: Sure!

acdlite:
I'm personally OK with merging this but I do think we should continue to think about the other types of side effects that might occur in an unmount function but aren't a state update

bvaughn:
This change is narrowly scoped to only suppress warnings for things that happen between a committed unmount and flushing of passive effects. It won't block any warnings for anything else- including a `setState` call made from within the passive effects destroy function.

PR #18102, opened 2020-02-22
Codemod to import * as React from "react";

sebmarkbage:
This is the correct way to import React from an ES module since the ES module will not have a default export. Only named exports.

I'm doing some changes to named exports and this will stop working. For open source ES modules we'll probably publish some default object that warns as an upgrade path but we can't use that in our own code.

### DIFF
--- a/packages/create-subscription/src/createSubscription.js
+++ b/packages/create-subscription/src/createSubscription.js
@@ -7,7 +7,7 @@
  * @flow
  */
 
-import React from 'react';
+import * as React from 'react';
 import invariant from 'shared/invariant';
 
 type Unsubscribe = () => void;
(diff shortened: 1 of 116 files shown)


sebmarkbage:
As far I can tell this didn't change any bytes in the build.

NE-SmallTown:
> since the ES module will not have a default export. Only named exports

I'm a little bit confused here, as far as I know, ES module **has** default export,  you can find it in [MDN](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) or [spec](https://tc39.es/ecma262/#sec-exports-static-semantics-exportentries), so maybe I miss some thing about this issue context, please correct me if I'm wrong, thanks!

TrySound:
"React es module will not have default export"

PR #18106, opened 2020-02-22
Export React as Named Exports instead of CommonJS

sebmarkbage:
The goal is to ensure more consistency in our tooling (e.g. we didn't even Flow check any imports of React in our code), better compiler output and unblock the ability to publish ES modules.

We can't conditionally export names, so we can't use feature flags to determine the exports anymore.

Instead, in my approach the default index.js exports all exports. This allows Flow to check and tests based on flags to get all features. Then I use a file extension scheme to fork the exports based on the four build types we need.

```
.modern.fb.js - experimental FB builds
.classic.fb.js - stable FB builds
.experimental.js - experimental npm builds
.stable.js - stable npm builds
```

I'd review the individual commit messages.

The output builds:

Stable - https://github.com/sebmarkbage/react-builds/commit/ed86c04190bf37da91bcf9505c8e56f572091efa
Experimental - https://github.com/sebmarkbage/react-builds/commit/5c0f29c14b7117f4362204f68fd6d7fdffc81486


sebmarkbage:
The increase in pre-gzip size for the react package is because this causes Rollup to encode this as:

```
exports.a = ...;
exports.b = ...;
exports.c = ...;
```

Instead of:

```
module.exports = {
  a: ...,
  b: ...,
  c: ...,
};
```

Which I guess is more correct in a sense but unclear what is actually better for perf.

This probably doesn't matter much since gzip takes care of it and the other packages won't matter as much since they have much fewer exports.

joeldenning:
Apologies for jumping in here without full context, but I'm a bit worried about switching to only named exports for react and react-dom, which appears to be the direction the ESM bundling is going. Isn't that a hugely breaking change? The following code is so very common:

```js
import ReactDOM from 'react-dom';
import React from 'react';

React.createElement('div')
ReactDOM.render(...)
```

For those wishing to use the ESM bundle once it's finished, will they have to migrate all code to use named exports? Is a codemod planned to help with migration? Or am I misunderstanding this entirely?

TrySound:
@joeldenning Don't worry. This is commonjs only change. ESM will have separate entry points with with default exports.

ericmorand:
@TrySound, do you know when the change is planned for? Specifically, when will react comes with an ESM variant alongside CJS and UMD ones?

TrySound:
No idea. Probably when conditional exports will be widely supported. Either in react 18 or 19.

ericmorand:
Thanks for your answer. I'm wondering if it would be possible to just provide the sources as-is in the published packages, alongside the "compiled" CJS and UMD variants. This way we could simple import from the sources and create our own application using ES syntax.

TrySound:
Sources will not help. There are a lot of configurations on top of it like flow syntax, aliases to shared package which is not published and many more.

ericmorand:
It's sad. Sounds like we are years away to get rid of bundling to build a react app. I am perfectly aware that bundling is here to stay because we have a long way to go before ES modules are widely adopted by all browsers, but I'm talking here about not bundling when developing - it's not needed and bundling impairs the performance a lot for developers.

Is it planned that React migrate to TypeScript at one point?

TrySound:
Snowpack allows to not use bundler for development. And their service has ready to use esm version of react bundle.
https://www.snowpack.dev/

Issue #17997, opened 2020-02-07
Bug: React Dev Tools Firefox extension fails to detect React


Kein:
React extension version: 4.4.0

## Steps To Reproduce

1. Install Firefox 72.0.2x64
2. Go to Discordapp.com

## The current behavior
Extension reports: _"This page does not appears to be using React"_

## The expected behavior
React is detected (just like it is currently in Chrome/Chromium)

hristo-kanchev:
It's happening because the page is explicitly blocking our script that gets injected in the page. 
We are using that to detect if React is there or not.

`Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).` in `injectGlobalHook.js`

Kein:
What are my options here?

ruslan-shuster:
Can I take this issue?

threepointone:
@ruslan-shuster sure! assigning to you.

ineggapps:
Chrome Too...

ruslan-shuster:
@ineggapps may I ask what is the url you were trying to access?

Kein:
For me it works just fine in Chromium[ungoogled] 79.0.3945.88

ruslan-shuster:
@Kein the problem which you described is now more or less understood
1) Your backend returns `content-security-policy` header which forbids injecting inline scripts to the page
2) `react-devtools` does inject some `<script>` tag to register window hooks, because Chrome content scripts cannot directly modify the `window` object, and `react-devtools` needs to install some hooks to the window.

So, what I'm going to do (if react core team agrees with this solution), is to implement some dedicated solution for `firefox`, which allows to install window hooks without injecting `<script>` tags. 

You can see more discussion of the problem & solutions [here](https://stackoverflow.com/questions/12395722/can-the-window-object-be-modified-from-a-chrome-extension).

ruslan-shuster:
@threepointone I had some time investigating this issue, and in order to solve it some decisions should be taken which I, as the first-time contributor, don't feel empowered to take. I'd like to hear your opinion, and also opinion of @bvaughn, if possible. So, here is what happens:

1) There is a `content-security-policy` `script-src` directive, which specifies from which sources the scripts for the web page may come. On the issue reported, this directive was used to allow only specified origins and scripts with particular `nonce`.

[...]

PR #18107, opened 2020-02-22
Indicate CSP violation on global hook injection

ruslan-shuster:
## Summary

Currently, __REACT_DEVTOOLS_GLOBAL_HOOK__ is being injected to the window through adding <script> tag to the web page, since it's the easiest cross-platform way to interact with web page environment. However, there are some users who restrict possible sources for scripts via Content Security Policy, but still want their web page to interact with `react-devtools`. The solution proposed here uses the possibility to whitelist webhook mounting script by its hash (see more [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src)).

## Test Plan

Here is [the web page](https://discordapp.com/) from the [reported issue](https://github.com/facebook/react/issues/17997). As you can see, `React` is not detected.

Screen Shot 2020-02-22 at 17 33 08

Here is a `script-src` directive which prevents __REACT_DEVTOOLS_GLOBAL_HOOK__ from mounting. 

Screen Shot 2020-02-22 at 17 34 25

Here is a web page with just the same CSP(though established through <meta> tag), but the hook mounting script is whitelisted in `script-src` directive. As you may see, `react-devtools` acknowledge `React`.

Screen Shot 2020-02-22 at 17 40 20

### DIFF
--- a/packages/react-devtools-extensions/src/injectGlobalHook.js
+++ b/packages/react-devtools-extensions/src/injectGlobalHook.js
@@ -86,6 +86,12 @@ if (sessionStorageGetItem(SESSION_STORAGE_RELOAD_AND_PROFILE_KEY) === 'true') {
   injectCode(rendererCode);
 }
 
+const cspViolationError = `react-devtools failed to install a global hook.
+It might be that it happened due to the Content Security Policy settings of a web page.
+If you are using content-security-policy http header or appropriate <meta> tag
+and you want your webpage to be detected by react-devtools extension consider adding
+'sha256-MJaZqAYHrwcn5ORXadICcIxGe8B8BLsjLE2OWa9jg78=' to script-src directive.`;
+
 // Inject a __REACT_DEVTOOLS_GLOBAL_HOOK__ global for React to interact with.
 // Only do this for HTML documents though, to avoid e.g. breaking syntax highlighting for XML docs.
 if ('text/html' === document.contentType) {
@@ -96,6 +102,12 @@ if ('text/html' === document.contentType) {
       saveNativeValues +
       detectReact,
   );
+
+  if (
+    typeof window.wrappedJSObject.__REACT_DEVTOOLS_GLOBAL_HOOK__ === 'undefined'
+  ) {
+    console.error(cspViolationError);
+  }
 }
 
 if (typeof exportFunction === 'function') {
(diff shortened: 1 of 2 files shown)


packages/react-devtools-extensions/src/injectGlobalHook.js
```diff
@@ -96,6 +102,12 @@ if ('text/html' === document.contentType) {
       saveNativeValues +
       detectReact,
   );
+
+  if (
+    typeof window.wrappedJSObject.__REACT_DEVTOOLS_GLOBAL_HOOK__ === 'undefined'
```
gaearon: Seems like this could also happen for other reasons? Such as being in an iframe. That's likely a lot more common. So I worry this would set people on a wrong path.

ruslan-shuster: @gaearon I am sure you know more scenarios that can happen here, and I will check what will happen when the website is open in iframe. This is just more-or-less the solution we agreed upon with @bvaughn in the [original issue](https://github.com/facebook/react/issues/17997). But, from your experience, what do you think would be better solution? Obviously, CSP gains more popularity, and I'm pretty sure there will be more and more websites stumbling upon it... Do you thing that comment in README.md is sufficient?

Thanks for your time and attention, I hope we'll find a way to make it better!

bvaughn: Sorry for the confusion.

My assumption about the warning was that we would only want to log it if we thought it was likely that the failure was due to CSP, but how to determine this is not something I know much about.

ruslan-shuster: @bvaughn no problem, thanks for your help:)

I tried to find any solution to detect CSP preventing inline scripts on the page, but I did not find anything that would work ONLY for CSP problem. There is always some kind of side effect or corner cases. 

What do you think about just adding a couple of lines to the README as I have [here](https://github.com/facebook/react/pull/18107/files#diff-7d04776b03aa659131b7d780d6f9d83fR110)? Is it sufficient to mitigate the problem people complain about?

bvaughn: Seems like a reasonable alternative.

PR #18125, opened 2020-02-25
[react-interactions] Ensure blur to window disengages press

trueadm:
This PR fixes an issue that was found internally. Specifically, if you engage a press and then a browser dialog opens, we should ensure that the press disengages. Otherwise, without this change the press with remain engaged. Also added a regression test to ensure we keep this behavior in the future.


necolas:
Shouldn't this also be added to Tap / non-legacy Press? Those are the ones we would port in the future, so this detail should be incorporated imo

trueadm:
@necolas Definitely. The should both be ported to those responders at some point too.

PR #18126, opened 2020-02-25
Add @nolint to FB bundle headers

trueadm:
Let's add the @nolint header to the FB/RN bundles so that we don't run eslint on the bundles.

### DIFF
--- a/scripts/rollup/wrappers.js
+++ b/scripts/rollup/wrappers.js
@@ -133,6 +133,7 @@ ${source}`;
 ${license}
  *
  * @noflow
+ * @nolint
  * @preventMunge
  * @preserve-invariant-messages
  */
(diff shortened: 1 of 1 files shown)


scripts/rollup/wrappers.js
```diff
@@ -133,6 +133,7 @@ ${source}`;
 ${license}
  *
  * @noflow
+ * @notlint
```
gaearon: nolint, not notlint?

sebmarkbage:
What is this for? Does it still run the lint on the bundles in the React repo or just not in FB?

We just caught a bug that would’ve gone to production by running lint on the bundles last week.

gaearon:
This is just for the FB linting. (The directive is not respected by eslint — only by arc lint.)

PR #18127, opened 2020-02-25
Remove runtime dependency on prop-types

gaearon:
This is to unblock the ESM work. There is no good story for using a CJS dependency so let's just inline this back. I lost the last argument because `console.error` now "sees" the stack again.

### DIFF
--- a/packages/react-art/package.json
+++ b/packages/react-art/package.json
@@ -26,7 +26,6 @@
     "create-react-class": "^15.6.2",
     "loose-envify": "^1.1.0",
     "object-assign": "^4.1.1",
-    "prop-types": "^15.6.2",
     "scheduler": "^0.19.0"
   },
   "peerDependencies": {
--- a/packages/react-dom/package.json
+++ b/packages/react-dom/package.json
@@ -19,7 +19,6 @@
   "dependencies": {
     "loose-envify": "^1.1.0",
     "object-assign": "^4.1.1",
-    "prop-types": "^15.6.2",
     "scheduler": "^0.19.0"
   },
   "peerDependencies": {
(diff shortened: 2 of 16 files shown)


sebmarkbage:
Looks like the tests doesn’t  “see” the stack.

gaearon:
Yeah, only for shallow renderer. I think it just needs to play along nicely by setting the current stack function.

gaearon:
k this should fix it

gaearon:
>It would be nice if react-shallow-renderer did this too.

Why?

ljharb:
@gaearon wouldn't it make more sense to add native ESM to prop-types? I'm happy to do that.

gaearon:
I think that would be good in general but the concern is that the ecosystem is not ready for transitive usage of ESM. We don’t want to make them autodiscoverable yet, but then it’s hard to try them without manually aliasing everything. So we were hoping to make this very scoped. Since we don’t *really* need either of those deps in a meaningful way. We also wanted to make React itself have zero dependencies anyway.

ljharb:
Node 13.3+ supports it and it’s planned to be backported into node 12 in April, I’m not sure why the ecosystem wouldn’t be ready.

“zero dependencies” isn’t honestly reached by inlining dependencies :-/ I’m confused why that’s a valuable.

sebmarkbage:
Zero dependencies is a valuable property for predictability, security, risk of downstream breakages and being able to reason eg about cost. Also makes things easier for our maintainability.

It comes with tradeoffs too but if this is the only one left, it doesn’t provide much value to leave it in. Especially since it’s DEV only it doesn’t get duplicated in prod so it doesn’t come with the normal downside of code duplication in the production bundles.

bvaughn:
FYI it's been reported that this change broke behavior of `PropTypes.resetWarningCache()`, which makes sense.

PR #18130, opened 2020-02-25
Update ReactDebugHooks to handle composite hooks

bvaughn:
The `useState` hook has always composed the `useReducer` hook. 1:1 composition like this is fine.

But some more recent hooks (e.g. `useTransition`, `useDeferredValue`) compose multiple hooks internally. This breaks `react-debug-tools` because it causes off-by-N errors when the debug tools re-renders the function.

For example, if a component were to use the `useTransition` and `useMemo` hooks, the normal hooks dispatcher would create a list of first state, then callback, then memo hooks, but the debug tools package would expect a list of transition then memo. This can break user code and cause runtime errors in both the `react-debug-tools` package and in product code.

This PR fixes the currently broken hooks by updating debug tools to be aware of the composite hooks (how many times it should call `nextHook` essentially) and adds tests to make sure they don't get out of sync again. We'll need to add similar tests for future composite hooks (like `useMutableSource` https://github.com/facebook/react/pull/18000).

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -243,7 +243,11 @@ function useResponder(
 function useTransition(
   config: SuspenseConfig | null | void,
 ): [(() => void) => void, boolean] {
-  nextHook();
+  // useTransition() composes multiple hooks internally.
+  // Advance the current hook index the same number of times
+  // so that subsequent hooks have the right memoized state.
+  nextHook(); // State
+  nextHook(); // Callback
   hookLog.push({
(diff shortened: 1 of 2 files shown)


bvaughn:
Code Sandbox build is trippin

bvaughn:
> Might want to drop a line on useOpaqueIdentifier.

Not sure I understand. That one hasn't landed yet?

gaearon:
I think Seb meant to drop a line on the PR so we don't forget to address before merging it.

bvaughn:
Ha! That makes more sense. :) Looks like you already added a comment. Thanks

PR #18133, opened 2020-02-26
Add a flag to disable module pattern components

gaearon:
This was already very niche and the warning doesn't fire on WWW. Let's disable these completely in the modern WWW build and stop doing these three checks for every element returned from a function?


gaearon:
We can probably even hard-break in both WWW builds.

PR #18136, opened 2020-02-26
Convert React Native builds to named exports

sebmarkbage:
These don't need any forks because we always export the same things atm.
They still build to CJS though atm.

### DIFF
--- a/packages/react-native-renderer/src/ReactNativeTypes.js
+++ b/packages/react-native-renderer/src/ReactNativeTypes.js
@@ -100,8 +100,6 @@ type SecretInternalsType = {
   ...
 };
 
-type SecretInternalsFabricType = {...};
-
 /**
  * Flat ReactNative renderer bundles are too big for Flow to parse efficiently.
  * Provide minimal Flow typing for the high-level RN API and call it a day.
@@ -137,7 +135,6 @@ export type ReactFabricType = {
     callback: ?Function,
   ): any,
   unmountComponentAtNode(containerTag: number): any,
-  __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: SecretInternalsFabricType,
   ...
 };
 
--- a/packages/react-native-renderer/fabric.js
+++ b/packages/react-native-renderer/fabric.js
@@ -7,10 +7,10 @@
  * @flow
  */
 
-'use strict';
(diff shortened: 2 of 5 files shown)


sebmarkbage:
Diff https://github.com/sebmarkbage/react-builds/commit/f4ed2aa148b3ae8ee5118c6795986a0705ec63fe

packages/react-native-renderer/src/ReactNativeTypes.js
```diff
@@ -137,7 +135,6 @@ export type ReactFabricType = {
     callback: ?Function,
   ): any,
   unmountComponentAtNode(containerTag: number): any,
-  __SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED: SecretInternalsFabricType,
```
TheSavior: I think we still need to export an empty object here so that it matches the paper renderer. Of code reads a key off and checks if the value is defined, that would now throw in fabric right because the object would be undefined?

sebmarkbage: The one current caller doesn't check if it's defined.

https://github.com/facebook/react-native/blob/master/Libraries/Utilities/JSDevSupportModule.js#L23-L24

And we're not going to be adding this back. No more internals.

sebmarkbage:
> Having a test plan that sanity checks that this build still works in the Facebook app would be great though!

That's what the sync is for. We'll find out. :)

TheSavior:
> That's what the sync is for. We'll find out. :)

That puts the burden on someone else, at some point in the future which would require them identifying this is the problem, and reverting it. However, reverting it probably wouldn't be trivial with additional changes stacked on top. Also, once this lands all partial syncs will have conflicts so we'd be totally blocked on moving forward until it can be resolved.

Doing a test sync of a change like this and verifying that the change works seems like something that should be done before merging.

sebmarkbage:
> That puts the burden on someone else, at some point in the future which would require them identifying this is the problem, and reverting it. However, reverting it probably wouldn't be trivial with additional changes stacked on top. Also, once this lands all partial syncs will have conflicts so we'd be totally blocked on moving forward until it can be resolved.

I mean this is literally the point of the sync. To batch that burden on someone else. Every other thing we land has that problem.

PR #18137, opened 2020-02-26
Remove /testing entry point from 'react' package

sebmarkbage:
We're not actually building this entry point so it doesn't point to anything. I can't think of a reason we'd need to fork the isomorphic one. We don't really fork it for anything since it's so generic to work with all renderers.

Since /profiling doesn't have this, it might confuse the story if we made people alias two things for testing but not profiling.


threepointone:
This will come back in my next PR, but merged this to make it into .13. Thank you!

PR #18144, opened 2020-02-27
Replace ReactShallowRenderer with a dependency

### LINKED ISSUE #17321: [Shallow Renderer] Plan forward
Let's discuss what to do with the shallow renderer here. As I mentioned in https://github.com/facebook/react/pull/16168#issuecomment-518344985, we aren't using it much and don't consider it a best practice. So we aren't going to be very good stewards of its API going forward.

My proposal was for Enzyme or folks interested in it to copy the code into another repo, and continue maintaining it there under a difference package name. It would make sense to Enzyme to start depending on that fork. We would then deprecate `react-test-renderer/shallow` in favor of the community-maintained fork.

If you'd like to volunteer to set up a repo, please let us know in this issue!

cc @davidmarkclements regarding the proposed `react-shallow-renderer` package name.

NMinhNguyen:
Closes #17321.

## Summary

I've extracted the shallow renderer out into its own package - http://npm.im/react-shallow-renderer https://github.com/NMinhNguyen/react-shallow-renderer

As discussed with @gaearon on Twitter, I didn't bother with DEV/PROD builds - the UMD build is essentially a [DEV build](https://unpkg.com/browse/react-shallow-renderer@16.12.0/umd/react-shallow-renderer.js), and [CJS](https://unpkg.com/browse/react-shallow-renderer@16.12.0/cjs/react-shallow-renderer.js) and [ES](https://unpkg.com/browse/react-shallow-renderer@16.12.0/esm/react-shallow-renderer.js) just have `process.env.NODE_ENV` checks inline.

## Test Plan

CI


packages/react-test-renderer/shallow.js
```diff
@@ -9,7 +9,7 @@
 
 'use strict';
 
-const ReactShallowRenderer = require('./src/ReactShallowRenderer');
+const ReactShallowRenderer = require('react-shallow-renderer');
```
NMinhNguyen: Not sure if I should add a deprecation warning here (so that it would log whenever this module is imported).

NMinhNguyen: Related to this - not sure if I should've kept `ReactShallowRenderer.js` file and re-exported from there (perhaps, some dependency on tooling).

NMinhNguyen:
When comparing the bundles, I noticed that `getComponentName` has changed since I first extracted the code about a month ago due to https://github.com/facebook/react/pull/18086. I'm wondering if there's a better way to keep the two in sync 😕

packages/react-test-renderer/shallow.js
sebmarkbage: This is fine but can you see if converting it to an ES module works: https://github.com/facebook/react/pull/18145/files#diff-d9693da2f86ad47019c4f8b94d7e5c3d

I believe this would cause the tests to fail so they need to be updated in the same way as I did in that PR.

gaearon:
To be fair, adding support for new component types (if even desired) will be a more likely blocker than non-ideal getComponentName names. I don't even think getComponentName is that useful when the stack is almost flat.

NMinhNguyen:
For now, I can sync my fork with the new Block type and cut another release. `16.12.1`? `16.13.0`?

packages/react-test-renderer/package.json
```diff
@@ -22,6 +22,7 @@
     "object-assign": "^4.1.1",
     "prop-types": "^15.6.2",
     "react-is": "^16.8.6",
+    "react-shallow-renderer": "^16.12.0",
```
gaearon: A bit unfortunate we didn't start with 1.0.0 or similar. So that you could break it more often if necessary. Although I guess nothing prevents you from jumping to 17 later without waiting for us.

gaearon:
To avoid the versioning awkwardness, we can make it `20.0.0` and pretend this never happened.

gaearon:
Blocks are an experiment and definitely don't make sense in the context of shallow rendering. So feel free to ignore them.

packages/react-dom/src/__tests__/ReactTestUtils-test.js
```diff
@@ -9,11 +9,11 @@
 
 'use strict';
 
-let createRenderer;
-let React;
-let ReactDOM;
-let ReactDOMServer;
-let ReactTestUtils;
+import ReactShallowRenderer from 'react-test-renderer/shallow';
```
gaearon: Rebase on master to get this change?

packages/react-test-renderer/shallow.js
NMinhNguyen: Seems like it worked and Rollup also did its thing 🙂

packages/react-test-renderer/shallow.js
```diff
@@ -9,8 +9,4 @@
 
 'use strict';
 
-const ReactShallowRenderer = require('./src/ReactShallowRenderer');
-
-// TODO: decide on the top-level export form.
-// This is hacky but makes it work with both Rollup and Jest.
-module.exports = ReactShallowRenderer.default || ReactShallowRenderer;
+export {default} from 'react-shallow-renderer';
```
gaearon: So what does this do? Is it using an actual ES build? Is this relying on the `module` field? @sebmarkbage Didn't you have concern about doing this?

packages/react-test-renderer/shallow.js
```diff
@@ -7,4 +7,4 @@
  * @flow
  */
 
-export {default} from './src/ReactShallowRenderer';
```
gaearon: Don't forget to delete the impl :-)

NMinhNguyen: Rebasing gone wrong :(

packages/react-test-renderer/shallow.js
NMinhNguyen: https://gist.github.com/NMinhNguyen/9d6863f70b44b953092c842465b3be90

NMinhNguyen: Actually, I added `console.log({ importee, importer })` in the Rollup forks plugin and seems like it picked ESM :)

```
{
  importee: '/react/packages/react-test-renderer/shallow.js',
  importer: undefined
}
{
  importee: '/react/packages/react-test-renderer/shallow.js',
  importer: undefined
}
{
  importee: '/react/packages/react-test-renderer/shallow.js',
  importer: undefined
}
{
  importee: 'react-shallow-renderer',
  importer: '/react/packages/react-test-renderer/shallow.js'
}
{
  importee: './react-shallow-renderer',
  importer: '/react/node_modules/react-shallow-renderer/esm/index.js'
}
{
  importee: 'object-assign',
  importer: '/react/node_modules/react-shallow-renderer/esm/react-shallow-renderer.js'
}
{
  importee: 'react-is',
  importer: '/react/node_modules/react-shallow-renderer/esm/react-shallow-renderer.js'
}
{
  importee: 'prop-types/checkPropTypes',
  importer: '/react/node_modules/react-shallow-renderer/esm/react-shallow-renderer.js'
}
{
  importee: './lib/ReactPropTypesSecret',
  importer: '/react/node_modules/prop-types/checkPropTypes.js'
}
{
  importee: '\u0000commonjs-proxy-./lib/ReactPropTypesSecret',
  importer: '/react/node_modules/prop-types/checkPropTypes.js'
}
{
  importee: './lib/ReactPropTypesSecret',
  importer: '/react/node_modules/prop-types/checkPropTypes.js'
}
{
  importee: './src/ReactIs',
  importer: '/react/packages/react-is/index.js'
}
{
  importee: '/react/node_modules/prop-types/lib/ReactPropTypesSecret.js',
  importer: '\u0000commonjs-proxy-/react/node_modules/prop-types/lib/ReactPropTypesSecret.js'
}
{
  importee: 'shared/ReactSymbols',
  importer: '/react/packages/react-is/src/ReactIs.js'
}
{
  importee: 'shared/isValidElementType',
  importer: '/react/packages/react-is/src/ReactIs.js'
}
{
  importee: 'shared/ReactSymbols',
  importer: '/react/packages/shared/isValidElementType.js'
```

gaearon:
I think this looks good. I’ll check tomorrow.

NMinhNguyen:
> I've tested this in a fresh CRA project and also at FB and couldn't find any issues.

Just wondering what kinds of tests you ran exactly, so I could try and run them myself for future releases of the package (maybe even have them run on CI).

gaearon:
For fresh CRA, I just copy pasted the docs snippet and verified it still passes tests — whether I write ESM+Babel or CJS+require.

NMinhNguyen:
Cool, thanks! But be careful with assumptions about ESM - as far as I know, Jest doesn't support ESM (https://github.com/facebook/jest/issues/4842), so what it ends up importing is the CJS version anyway. In other words, it doesn't matter what syntax you use at the call site (in your tests).

sebmarkbage:
@NMinhNguyen The syntax matters because using the `import` syntax adds a compatibility check at runtime that looks for `esModule`.

The problem is that we run Jest against both against source as it is. In this case Jest (Babel) is responsible for converting the `export ....` part to CommonJS. We also run Jest against an existing bundle and in that case Rollup built the bundle. We couldn't use just `require(...)` in our tests because that didn't add the interop layer that made it work with both.

However, now we won't run tests against the source any more. Just the bundle published to npm. So we don't need that compatibility anymore. So now we probably could revert the changes I made to the tests in https://github.com/facebook/react/pull/18145

NMinhNguyen:
@sebmarkbage sorry if I'm misunderstanding, but in CRA which points at the published `react-shallow-renderer` package, I think Jest would pick up the CJS bundle which has no `esModule` (https://unpkg.com/browse/react-shallow-renderer@16.12.0/cjs/react-shallow-renderer.js), so the `import` syntax in the test shouldn't matter, right?

PR #18145, opened 2020-02-27
Convert the rest of react-dom and react-test-renderer to Named Exports

sebmarkbage:
Nothing interesting here except that ReactShallowRenderer currently exports a class with a static method instead of an object.

I think the public API is probably just meant to be createRenderer but currently the whole class is exposed. So this means that we have to keep it as default export. We could potentially also expose a named export for createRenderer but that's going to cause compatibility issues.

So I'm just going to make that export default.

Unfortunately Rollup and Babel (which powers Jest) disagree on how to import this. So to make it work I had to move the jest tests to imports.

This doesn't work with module resetting. Some tests weren't doing that anyway and the rest is just testing ReactShallowRenderer so meh.

Also see https://github.com/facebook/react/pull/18144

### DIFF
--- a/packages/react-dom/src/__tests__/ReactTestUtils-test.js
+++ b/packages/react-dom/src/__tests__/ReactTestUtils-test.js
@@ -9,11 +9,11 @@
 
 'use strict';
 
-let createRenderer;
-let React;
-let ReactDOM;
-let ReactDOMServer;
-let ReactTestUtils;
+import ReactShallowRenderer from 'react-test-renderer/shallow';
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import * as ReactDOMServer from 'react-dom/server';
+import * as ReactTestUtils from 'react-dom/test-utils';
 
 function getTestDocument(markup) {
   const doc = document.implementation.createHTMLDocument('');
@@ -27,14 +27,6 @@ function getTestDocument(markup) {
 }
 
 describe('ReactTestUtils', () => {
-  beforeEach(() => {
-    createRenderer = require('react-test-renderer/shallow').createRenderer;
-    React = require('react');
-    ReactDOM = require('react-dom');
-    ReactDOMServer = require('react-dom/server');
-    ReactTestUtils = require('react-dom/test-utils');
-  });
-
   it('Simulate should have locally attached media events', () => {
     expect(Object.keys(ReactTestUtils.Simulate).sort()).toMatchSnapshot();
   });
@@ -403,7 +395,7 @@ describe('ReactTestUtils', () => {
       }
 
       const handler = jest.fn().mockName('spy');
-      const shallowRenderer = createRenderer();
+      const shallowRenderer = ReactShallowRenderer.createRenderer();
       const result = shallowRenderer.render(
         <SomeComponent handleClick={handler} />,
       );
--- a/packages/react-test-renderer/shallow.js
+++ b/packages/react-test-renderer/shallow.js
@@ -7,10 +7,4 @@
  * @flow
  */
 
-'use strict';
-
-const ReactShallowRenderer = require('./src/ReactShallowRenderer');
-
-// TODO: decide on the top-level export form.
-// This is hacky but makes it work with both Rollup and Jest.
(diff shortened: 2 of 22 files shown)


packages/react-dom/src/__tests__/ReactTestUtils-test.js
```diff
@@ -9,11 +9,11 @@
...
-let createRenderer;
-let React;
-let ReactDOM;
-let ReactDOMServer;
-let ReactTestUtils;
+import ReactTestRenderer from 'react-test-renderer/shallow';
+import * as React from 'react';
+import * as ReactDOM from 'react-dom';
+import * as ReactDOMServer from 'react-dom/server';
+import * as ReactTestUtils from 'react-dom/test-utils';
```
gaearon: Huh?

gaearon: What's special about this file? Why didn't you have to convert all other tests importing TestUtils or ReactDOMServer to use imports?

packages/react-dom/src/__tests__/ReactTestUtils-test.js
```diff
@@ -9,11 +9,11 @@
 
 'use strict';
 
-let createRenderer;
-let React;
-let ReactDOM;
-let ReactDOMServer;
-let ReactTestUtils;
+import ReactTestRenderer from 'react-test-renderer/shallow';
```
gaearon: Okay this is what confused me. It should say `ReactShallowRenderer`.

packages/react-test-renderer/src/ReactShallowRenderer.js
```diff
@@ -856,4 +856,8 @@ function getMaskedContext(contextTypes, unmaskedContext) {
   return context;
 }
 
+// This should probably be a default export and a named export.
```
gaearon: I don't understand what this comment is saying. Which one should it be?

packages/react-test-renderer/shallow.js
```diff
@@ -7,10 +7,4 @@
...
  */
 
-'use strict';
-
-const ReactShallowRenderer = require('./src/ReactShallowRenderer');
-
-// TODO: decide on the top-level export form.
-// This is hacky but makes it work with both Rollup and Jest.
-module.exports = ReactShallowRenderer.default || ReactShallowRenderer;
+export {default} from './src/ReactShallowRenderer';
```
gaearon: Isn't this gonna break `const ShallowRenderer = require('react-test-renderer/shallow')`? Is this not a breaking change?

packages/react-dom/src/__tests__/ReactTestUtils-test.js
gaearon: OK I see now, we only do this where we import `ReactShallowRenderer`.

packages/react-dom/src/__tests__/ReactTestUtils-test.js
sebmarkbage: potato, potato... fixed :)

packages/react-test-renderer/src/ReactShallowRenderer.js
sebmarkbage: Both

packages/react-test-renderer/shallow.js
sebmarkbage: No, Rollup makes this into `module.exports = `. (Babel doesn't which is what Jest is struggling with.)

gaearon: 🤯 I had no idea

packages/react-test-renderer/src/ReactShallowRenderer.js
sebmarkbage: When we expose a proper ES module we can do:

```
export default ReactShallowRenderer;
export createRenderer;
```

It's just that if I do that, then Rollup doesn't do the magic thing to only use `module.exports = `

sebmarkbage:
https://github.com/sebmarkbage/react-builds/commit/fb86179870227b6cc489fa325208aaba1acf5c02

gaearon:
beautiful

PR #18149, opened 2020-02-27
test(getComponentName): Increase test coverage

eps1lon:
## Summary

Increases coverage of `getComponentName` to increase likely-hood of #15638 being accepted.

The usages of `getComponentName` that are still uncovered are included in 2979a97af8b12d7842f3824c2ca27cfe0649482b.  They either affect react-native (don't know how this is tested), affect messages for possibly internal bugs ("file an issue") or aren't tested to begin with.

It seems like https://github.com/facebook/react/blob/abc8fa964631fa934aaafdaaaae065780d2b1ec1/packages/react-dom/src/server/ReactPartialRenderer.js#L1321 is actually not reachable. Looking at https://github.com/facebook/react/blob/abc8fa964631fa934aaafdaaaae065780d2b1ec1/packages/react-dom/src/server/ReactPartialRenderer.js#L124 it seems like the owner name is forcefully set to null in other cases and I can't find any references to owners in react-dom/server. Looks like owners aren't avaiable in the server renderer and therefore https://github.com/facebook/react/blob/abc8fa964631fa934aaafdaaaae065780d2b1ec1/packages/react-dom/src/server/ReactPartialRenderer.js#L1321 is dead code?

## Test Plan

1. Replace `getComponentName(**)` with `null`
2. If tests pass then mark as `fixme` and push to verify with full CI
2. If tests fail then this line is covered. Everything is fine

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerLifecycles-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerLifecycles-test.js
@@ -213,7 +213,7 @@ describe('ReactDOMServerLifecycles', () => {
   it('should invoke both deprecated and new lifecycles if both are present', () => {
     const log = [];
 
-    class Component extends React.Component {
+    class MyComponent extends React.Component {
       componentWillMount() {
         log.push('componentWillMount');
       }
@@ -225,8 +225,10 @@ describe('ReactDOMServerLifecycles', () => {
       }
     }
 
-    expect(() => ReactDOMServer.renderToString(<Component />)).toWarnDev(
-      'componentWillMount has been renamed',
+    expect(() => ReactDOMServer.renderToString(<MyComponent />)).toWarnDev(
+      'componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.\n\n' +
+        '* Move code from componentWillMount to componentDidMount (preferred in most cases) or the constructor.\n\n' +
+        'Please update the following components: MyComponent',
     );
     expect(log).toEqual(['componentWillMount', 'UNSAFE_componentWillMount']);
   });
--- a/packages/react-dom/src/__tests__/ReactDOMServerLifecycles-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerLifecycles-test.js
@@ -213,7 +213,7 @@ describe('ReactDOMServerLifecycles', () => {
   it('should invoke both deprecated and new lifecycles if both are present', () => {
     const log = [];
 
-    class MyComponent extends React.Component {
+    class Component extends React.Component {
       componentWillMount() {
         log.push('componentWillMount');
       }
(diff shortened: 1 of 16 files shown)


eps1lon:
Hey @bvaughn is this ok if I ping you to review this? Seems like this concerns more of your work (dev tooling).

packages/react-dom/src/__tests__/ReactDOMServerLifecycles-test.js
```diff
@@ -225,8 +225,10 @@ describe('ReactDOMServerLifecycles', () => {
       }
     }
 
-    expect(() => ReactDOMServer.renderToString(<Component />)).toWarnDev(
-      'componentWillMount has been renamed',
+    expect(() => ReactDOMServer.renderToString(<MyComponent />)).toWarnDev(
+      'componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.\n\n' +
+        '* Move code from componentWillMount to componentDidMount (preferred in most cases) or the constructor.\n\n' +
+        'Please update the following components: MyComponent',
```
bvaughn: I'm not sure I understand why we updated this test. This test isn't about verifying warnings. It's testing behavior, and the `toWarnDev()` assertions are just there to suppress the (expected) warnings.

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -339,6 +339,42 @@ describe('ReactHooksWithNoopRenderer', () => {
...
+      expect(Scheduler).toFlushWithoutYielding();
+      expect(() => act(() => _updateCountA(1))).toErrorDev(
+        "Warning: Can't perform a React state update on an unmounted " +
+          'component. This is a no-op, but it indicates a memory leak in your ' +
+          'application. To fix, cancel all subscriptions and asynchronous ' +
+          'tasks in a useEffect cleanup function.\n' +
+          '    in CounterA (at **)',
+      );
+      // already cached so this logs no error
+      act(() => _updateCountA(1));
```
bvaughn: This isn't updating anything, since the state is already 1.
```suggestion
      act(() => _updateCountA(2));
```

packages/react-dom/src/__tests__/ReactDOMServerLifecycles-test.js
eps1lon: Hm, let me check the git blame for when this warning was added. Might be I got confused or the warning message wasn't checked in the first place.

bvaughn: No worries. I think maybe just revert this file and I'll merge the rest?

eps1lon: Ok so the reason I chose this test was that it was the first one I found that hit https://github.com/facebook/react/blob/4b5cb6bddfc7b25ad41f4c5a555b40250479edb7/packages/react-dom/src/server/ReactPartialRenderer.js#L577. But I agree that this is the wrong place. There's already a better place. 0ed6d5d90962c983e7ce07ff8fb773110c19afb2 should resolve this issue.

bvaughn: Agreed, the new location seems better.

PR #18152, opened 2020-02-27
Run test-prod job for experimental builds

gaearon:
We forgot to run it, so it regressed. This is supposed to fail. Once it fails, I'll push a fix.


gaearon:
OK this didn't actually regress, it was just Jest cache for me locally. But we still should get this in.

PR #18153, opened 2020-02-27
Don't build shallow renderer for FB

gaearon:
We're now getting it from npm (`react-shallow-renderer`) so we don't need this bundle.
Related diff: D20142456.

### DIFF
--- a/scripts/rollup/bundles.js
+++ b/scripts/rollup/bundles.js
@@ -295,7 +295,7 @@ const bundles = [
       }),
   },
   {
-    bundleTypes: [FB_WWW_DEV, NODE_DEV, NODE_PROD, UMD_DEV, UMD_PROD],
+    bundleTypes: [NODE_DEV, NODE_PROD, UMD_DEV, UMD_PROD],
     moduleType: NON_FIBER_RENDERER,
     entry: 'react-test-renderer/shallow',
     global: 'ReactShallowRenderer',


scripts/rollup/bundles.js
```diff
@@ -295,7 +295,7 @@ const bundles = [
       }),
   },
   {
-    bundleTypes: [FB_WWW_DEV, NODE_DEV, NODE_PROD, UMD_DEV, UMD_PROD],
+    bundleTypes: [NODE_DEV, NODE_PROD, UMD_DEV, UMD_PROD],
```
sebmarkbage: We really don't need any of these right? This is just building a forwarding file. We can just hard code that in the `/npm/` folder instead. https://github.com/facebook/react/blob/master/packages/react-test-renderer/npm/shallow.js

Like we do for other forwarding modules that are just blank. https://github.com/facebook/react/blob/master/packages/react-dom/npm/server.js

NMinhNguyen: I might be missing something, but I think you'd still need UMD. But otherwise, for Node you should be able to hardcode in `.npm/`.

sebmarkbage: ugh. UMD. yea.

sebmarkbage: https://github.com/facebook/react/pull/18157

PR #18155, opened 2020-02-27
Refactor Portal Fiber parts for event system changes

trueadm:
This PR makes some changes to how we store the root container instance in context, and instead makes it so we store the root fiber instead. I've also added a Flow node for portal containers and also made it so that portal contains also have a `current` field, like root fiber containers do (this is needed for later changes that make the new event system possible). Lastly, we no longer pass around a `container` DOM node through the event dispatch system and instead expect the root/portal fiber instance passed to the host config.

The reason for making these changes, is that the previous assumption in passing around the "root container" element to the event system falls down in a bunch of places. Notably, where we might use the same container DOM element for multiple roots (`ReactDOM.createPortal(..., document.body)` comes to mind here). So instead of passing around the DOM node, we can instead pass around the boxed container for either the root or portal fiber – allowing the event system to traverse the correct tree and apply the correct semantics for when we encounter multiple roots during event propagation.

Lastly, the changes to the host context stack are so that we can get the root/portal fiber directly. This was far less code and probably more performant that maintaining another stack just for root/portal fibers. We can easily get the container element from the root/portal fibers after, so there should be no difference in output compared to master.

### DIFF
--- a/packages/react-reconciler/src/ReactFiber.js
+++ b/packages/react-reconciler/src/ReactFiber.js
@@ -26,6 +26,7 @@ import type {UpdateQueue} from './ReactUpdateQueue';
 import type {ContextDependency} from './ReactFiberNewContext';
 import type {HookType} from './ReactFiberHooks';
 import type {SuspenseInstance} from './ReactFiberHostConfig';
+import type {FiberPortal} from './ReactFiberRoot';
 
 import invariant from 'shared/invariant';
 import {
@@ -902,11 +903,13 @@ export function createFiberFromPortal(
   const pendingProps = portal.children !== null ? portal.children : [];
   const fiber = createFiber(HostPortal, pendingProps, portal.key, mode);
   fiber.expirationTime = expirationTime;
-  fiber.stateNode = {
+  const portalContainer: FiberPortal = {
     containerInfo: portal.containerInfo,
+    current: fiber,
     pendingChildren: null, // Used by persistent updates
     implementation: portal.implementation,
   };
+  fiber.stateNode = portalContainer;
   return fiber;
 }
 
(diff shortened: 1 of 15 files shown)


trueadm:
@bvaughn Any idea why the devtools build fails for CI on this PR? Have I done something wrong or is it CI acting up again?

bvaughn:
> filesystem layer verification failed for digest sha256:c8a34c72f5385eac42dc033f17c21af4a6cd34c842d7cd9b0f6d9dcdffd8c404

No clue what this means.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -986,13 +986,16 @@ function commitContainer(finishedWork: Fiber) {
...
       const portalOrRoot: {
         containerInfo: Container,
         pendingChildren: ChildSet,
         ...
-      } = finishedWork.stateNode;
+      } = portalContainer;
       const {containerInfo, pendingChildren} = portalOrRoot;
       replaceContainerChildren(containerInfo, pendingChildren);
+      // Update the fiber on the portal container
+      portalContainer.current = finishedWork;
```
trueadm: This seemed like the best place to update the `current` pointer to the most recent fiber on the portal container.

packages/react-reconciler/src/ReactFiber.js
```diff
@@ -902,11 +903,13 @@ export function createFiberFromPortal(
   const pendingProps = portal.children !== null ? portal.children : [];
   const fiber = createFiber(HostPortal, pendingProps, portal.key, mode);
   fiber.expirationTime = expirationTime;
-  fiber.stateNode = {
+  const portalContainer: FiberPortal = {
     containerInfo: portal.containerInfo,
+    current: fiber,
```
trueadm: We now track the `current` fiber of the portal container.

packages/react-reconciler/src/ReactFiberCommitWork.js
sebmarkbage: This is only for the "persistent" mode (i.e. Fabric). There's no commit phase for Portals.

trueadm:
Thanks for the feedback @sebmarkbage. I'll go in a different direction.

PR #18156, opened 2020-02-27
[Fabric] Fix targetAsInstance dispatchEvent "cannot read property of null"

JoshuaGross:
## Summary

In some cases in prod testing, the `targetFiber` here in `dispatchEvent` does not have a stateNode. If we try to get `canonical` directly from stateNode, it will crash. We already check that `targetFiber` is non-null before continuing, I think it makes sense to check stateNode as well.

It's unclear to me why this is happening in the first place, though (why we'd have a targetFiber but not a stateNode). We may want to investigate and fix that in addition to/instead of this.

## Test Plan

`yarn test`
`yarn test-prod`
`yarn flow fabric`

I have an FB-internal prod test as well.

### DIFF
--- a/packages/react-native-renderer/src/ReactFabricEventEmitter.js
+++ b/packages/react-native-renderer/src/ReactFabricEventEmitter.js
@@ -88,6 +88,7 @@ export function dispatchEvent(
   if (enableNativeTargetAsInstance) {
     if (targetFiber != null) {
       const stateNode = targetFiber.stateNode;
+      // Guard against Fiber being unmounted
       if (stateNode != null) {
         eventTarget = stateNode.canonical;
       }
--- a/packages/react-native-renderer/src/ReactFabricEventEmitter.js
+++ b/packages/react-native-renderer/src/ReactFabricEventEmitter.js
@@ -86,7 +86,7 @@ export function dispatchEvent(
 
   let eventTarget = null;
   if (enableNativeTargetAsInstance) {
-    if (targetFiber != null) {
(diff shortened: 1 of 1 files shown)


gaearon:
Do you need this in the next sync?

JoshuaGross:
Eli thinks it “shouldn’t be possible” for stateNode to be null so I need to do a bit more investigation. I would probably just land the eventual solution as a partial sync if it’s urgent.

gaearon:
fwiw we do null stateNode since recently on unmount.

JoshuaGross:
@gaearon Oh interesting - so the case where it's crashing, it would potentially mean an event has been dispatched to an unmounted component?

JoshuaGross:
@gaearon If it seems to you that this is a reasonable check to make, and that we aren't covering up some earlier data corruption, then I'm fine landing this.

gaearon:
>so the case where it's crashing, it would potentially mean an event has been dispatched to an unmounted component

Possibly.

>If it seems to you that this is a reasonable check to make, and that we aren't covering up some earlier data corruption, then I'm fine landing this.

I don't know the event system enough to tell if that's ever supposed to happen. @trueadm might know.

trueadm:
It's definitely possible with sync events. An event can cause a sync update that causes an update that unmounts a fiber, then the event can propagate to the unmounted fiber. It's rare, but can happen and should probably be guarded against.

JoshuaGross:
Cool. Thanks for the context (cc @TheSavior)

PR #18157, opened 2020-02-27
Remove shallow Node builds

sebmarkbage:
This is just a forwarding module. We can hardcode it.

### DIFF
--- a/packages/react-test-renderer/npm/shallow.js
+++ b/packages/react-test-renderer/npm/shallow.js
@@ -1,7 +1,3 @@
 'use strict';
 
-if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./cjs/react-test-renderer-shallow.production.min.js');
-} else {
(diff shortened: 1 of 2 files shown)


packages/react-test-renderer/npm/shallow.js
```diff
@@ -1,7 +1,3 @@
 'use strict';
 
-if (process.env.NODE_ENV === 'production') {
-  module.exports = require('./cjs/react-test-renderer-shallow.production.min.js');
-} else {
-  module.exports = require('./cjs/react-test-renderer-shallow.development.js');
-}
+module.exports = require('./react-shallow-renderer');
```
NMinhNguyen: Shouldn't this not have `./`? If you want to forward to the npm package?

NMinhNguyen:
I mentioned in my original PR that I didn’t bother with a UMD_PROD build in my fork, so I guess your UMD_PROD build would be a minified version of a UMD_DEV build with things like warnings and non-minified errors, but I guess that’s fine? 😅

PR #18165, opened 2020-02-27
Move remaining things to named exports

sebmarkbage:
The interesting case here is the noop renderers. The wrappers around the reconciler now changed to use a local export that gets mutated.

ReactNoop and ReactNoopPersistent now have to destructure the object to list out the names it's going to export. We should probably refactor ReactNoop away from createReactNoop. Especially since it's also not Flow typed.

### DIFF
--- a/packages/react-art/Rectangle.js
+++ b/packages/react-art/Rectangle.js
@@ -7,8 +7,4 @@
  * @flow
  */
 
-'use strict';
-
-const Rectangle = require('./npm/Rectangle');
-
-module.exports = Rectangle;
+export {default} from './npm/Rectangle';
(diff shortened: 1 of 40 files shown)


packages/react-art/Rectangle.js
```diff
@@ -7,8 +7,4 @@
  * @flow
  */
 
-'use strict';
-
-const Rectangle = require('./npm/Rectangle');
-
-module.exports = Rectangle;
+export {default} from './npm/Rectangle';
```
gaearon: How does this work? Aren’t they still CJS?

sebmarkbage: Yea, for now. But once ART switches to ES modules we can switch the inner implementation too.

PR #18179, opened 2020-02-28
More event system cleanup and scaffolding

trueadm:
This PR is another step towards adding a forked modern event system that uses roots. Before we get to that stage, we need to do some cleanup and add some parts of the scaffolding. 

Specifically, this PR does the following:
- fork `ensureListeningTo` to accept modern/legacy events
- wrap `legacyTrapBubbledEvent` calls in a flag, as we won't do these in the modern event system (we use the capture phase instead).
- add `dispatchEventForLegacyPluginEventSystem` calls at `DOMEventListener` behind the flag, as the entrance point for dispatching modern event system events.
- add `listenToTopLevelEvent` calls at `ReactDOMEventReplaying` behind the flag, as the entrance point for replaying modern event system events.
- add `DOMModernPluginEventSystem` module. This doesn't do much for now, the rest of the code and tests will come in a follow up PR. It does register events accordingly however.
- various small changes to Flow types and cleanup of DOMEventListener

Note: there should be no functional changes to the existing legacy event system.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -10,6 +10,7 @@
 import {registrationNameModules} from 'legacy-events/EventPluginRegistry';
 import {canUseDOM} from 'shared/ExecutionEnvironment';
 import endsWith from 'shared/endsWith';
+import invariant from 'shared/invariant';
 import {setListenToResponderEventTypes} from '../events/DeprecatedDOMEventResponderSystem';
 
 import {
@@ -59,7 +60,6 @@ import {getListenerMapForElement} from '../events/DOMEventListenerMap';
 import {
   addResponderEventSystemEvent,
   removeActiveResponderEventSystemEvent,
-  trapBubbledEvent,
 } from '../events/ReactDOMEventListener.js';
 import {mediaEventTypes} from '../events/DOMTopLevelEventTypes';
 import {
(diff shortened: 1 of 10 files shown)


packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -260,16 +270,35 @@ if (__DEV__) {
...
+    const rootContainerElement =
+      rootContainerInstance.nodeType === COMMENT_NODE
+        ? rootContainerInstance.parentNode
+        : rootContainerInstance;
+    // Containers can only ever be element nodes. We do not
+    // want to register events to document fragments or documents
+    // with the modern plugin event system.
+    invariant(
+      rootContainerElement && rootContainerElement.nodeType === ELEMENT_NODE,
+      'ensureListeningTo(): recieved a container that was not an element node. ' +
```
threepointone:
```suggestion
      'ensureListeningTo(): received a container that was not an element node. ' +
```

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -260,16 +270,35 @@ if (__DEV__) {
...
+    // which should be an element.
+    const rootContainerElement =
+      rootContainerInstance.nodeType === COMMENT_NODE
+        ? rootContainerInstance.parentNode
+        : rootContainerInstance;
+    // Containers can only ever be element nodes. We do not
+    // want to register events to document fragments or documents
+    // with the modern plugin event system.
+    invariant(
+      rootContainerElement && rootContainerElement.nodeType === ELEMENT_NODE,
```
threepointone:
```suggestion
      rootContainerElement != null && rootContainerElement.nodeType === ELEMENT_NODE,
```
is this 'faster'? or would that be an unnecessary optimisation?

scripts/error-codes/codes.json
```diff
@@ -345,5 +345,6 @@
   "344": "Expected prepareToHydrateHostSuspenseInstance() to never be called. This error is likely caused by a bug in React. Please file an issue.",
   "345": "Root did not complete. This is a bug in React.",
   "346": "An event responder context was used outside of an event cycle.",
-  "347": "Maps are not valid as a React child (found: %s). Consider converting children to an array of keyed ReactElements instead."
+  "347": "Maps are not valid as a React child (found: %s). Consider converting children to an array of keyed ReactElements instead.",
+  "348": "ensureListeningTo(): recieved a container that was not an element node. This is likely a bug in React."
```
threepointone:
```suggestion
  "348": "ensureListeningTo(): received a container that was not an element node. This is likely a bug in React."
```

PR #18180, opened 2020-02-28
Check in a forked version of object-assign only for UMD builds

sebmarkbage:
This one uses ES modules so that we can inline it into UMD builds.

We could wait for object-assign to make an ESM export but we're going to remove this dependency and assume global polyfills in the next version anyway. However, we'd have to figure out how to keep the copyright header and it'll get counted in terms of byte size (even if other tooling removes it).

A lot of headache when we have our own implementation anyway. So I'll just use that.

Ours is not resilient to checking certain browser bugs but those browsers are mostly unused anyway. (Even FB breaks on them presumably.)

We also don't need to be resilient to Symbols since the way React uses it we shouldn't need to copy symbols

### DIFF
--- a/packages/shared/forks/object-assign.inline-umd.js
+++ b/packages/shared/forks/object-assign.inline-umd.js
@@ -1,0 +1,31 @@
+/**
+ * Copyright 2004-present Facebook. All Rights Reserved.
+ */
+
+const hasOwnProperty = Object.prototype.hasOwnProperty;
+
+const _assign = function(to, from) {
+  for (let key in from) {
+    if (hasOwnProperty.call(from, key)) {
+      to[key] = from[key];
+    }
+  }
+};
+
+export default Object.assign ||
+  function(target, sources) {
+    if (target == null) {
+      throw new TypeError('Object.assign target cannot be null or undefined');
+    }
+
+    const to = Object(target);
+
+    for (let nextIndex = 1; nextIndex < arguments.length; nextIndex++) {
+      const nextSource = arguments[nextIndex];
+      if (nextSource != null) {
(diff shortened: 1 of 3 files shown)


sebmarkbage:
I'll just take one of our own implementations then. Probably easier than figuring out the banner.

packages/shared/forks/object-assign.inline-umd.js
```diff
@@ -0,0 +1,31 @@
...
+    if (hasOwnProperty.call(from, key)) {
+      to[key] = from[key];
+    }
+  }
+};
+
+export default Object.assign ||
+  function(target, sources) {
+    if (target == null) {
+      throw new TypeError('Object.assign target cannot be null or undefined');
```
gaearon: Do we actually care about this? Would we expect to ever happen with our usage?

sebmarkbage: Maybe if it's a bug. Meh. I mostly just took it as-is from FB.

scripts/babel/transform-object-assign-require.js
```diff
@@ -28,6 +28,10 @@ module.exports = function autoImporter(babel) {
 
     visitor: {
       CallExpression: function(path, file) {
+        if (file.filename.indexOf('object-assign') !== -1) {
+          // Don't replace Object.assign if we're transforming object-assign
+          return;
```
sebmarkbage: Since we're now transforming the file itself, I needed to ensure this bails out.

sebmarkbage:
Tested fixtured and ensured that it conditionally picks up the native version.

PR #18181, opened 2020-02-28
Change ReactVersion from CJS to ES module

sebmarkbage:
We need this to get rid of CJS in the build tool chain.


bvaughn:
`scripts/tasks/version-check.js` requires this file, yes.

sebmarkbage:
That script doesn't work on master because `/package.json` doesn't have a version. Fixed it.

PR #18184, opened 2020-02-28
Refine isFiberSuspenseAndTimedOut

trueadm:
This changes the `isFiberSuspenseAndTimedOut` implementation based on Sebastian's feedback below:

> I think you might need to check for fiber.memoizedState.dehydrated === null here. If dehydrated is non-null it's a not-yet hydrated boundary. I believe child will be null in that case.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberScope.js
+++ b/packages/react-reconciler/src/ReactFiberScope.js
@@ -31,7 +31,9 @@ function isFiberSuspenseAndTimedOut(fiber: Fiber): boolean {
 }
 
 function getSuspenseFallbackChild(fiber: Fiber): Fiber | null {
-  return ((((fiber.child: any): Fiber).sibling: any): Fiber).child;
+  const child = fiber.child;
+  const sibling = child ? child.sibling : null;
+  return sibling ? sibling.child : null;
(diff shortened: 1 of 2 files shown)


sebmarkbage:
This might not throw but this still seems like it'll yield the wrong results. The assumptions are still wrong about isFiberSuspenseAndTimedOut

trueadm:
@sebmarkbage What are the correct assumptions? I can adjust accordingly.

packages/react-reconciler/src/ReactFiberScope.js
```diff
@@ -31,7 +31,9 @@ function isFiberSuspenseAndTimedOut(fiber: Fiber): boolean {
 }
```
sebmarkbage: I think you might need to check for `fiber.memoizedState.dehydrated === null` here.

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberSuspenseComponent.js#L37

If dehydrated is non-null it's a not-yet hydrating boundary.

trueadm: Addressed your feedback. Thank you :)

PR #18185, opened 2020-02-28
Update react-shallow-renderer from 16.12.0 to 16.13.0

NMinhNguyen:
## Summary

Published a new version with @gaearon's PR https://github.com/NMinhNguyen/react-shallow-renderer/pull/15 as well as a few cosmetic changes (https://github.com/NMinhNguyen/react-shallow-renderer/pull/14 and https://github.com/NMinhNguyen/react-shallow-renderer/pull/21).

Bundle diff:

https://gist.github.com/NMinhNguyen/11fd4d00bb182e1cf8c8a007ef0df133

## Test Plan

CI

PR #18186, opened 2020-02-29
Only use Rollup's CommonJS plugin for "react-art"

sebmarkbage:
This upgrades the shallow renderer dep, and properly make react-debug-tools and jest-react depend on object-assign as an external dependency instead of compiling it in.

This means that "art" is our only CJS dependency that needs to be compiled with Rollup. So we can now disable the CJS plugin for everything else. That ensures we don't reintroduce new CJS dependencies.

I'm hoping to port "art" to ESM or we can just drop the UMD build. Then we can get rid of the plugin completely.

### DIFF
--- a/packages/jest-react/package.json
+++ b/packages/jest-react/package.json
@@ -23,6 +23,9 @@
     "react": "^16.0.0",
     "react-test-renderer": "^16.0.0"
   },
+  "dependencies": {
+    "object-assign": "^4.1.1"
+  },
   "files": [
     "LICENSE",
     "README.md",
--- a/packages/react-debug-tools/package.json
+++ b/packages/react-debug-tools/package.json
@@ -29,6 +29,7 @@
     "react": "^16.0.0"
   },
   "dependencies": {
-    "error-stack-parser": "^2.0.2"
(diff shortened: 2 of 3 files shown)


NMinhNguyen:
I'll close https://github.com/facebook/react/pull/18185 😅

sebmarkbage:
@NMinhNguyen I merged yours and rebased mine.

NMinhNguyen:
I actually realised something.. I remember seeing an issue in `hoist-non-react-statics` https://github.com/mridgway/hoist-non-react-statics/issues/52 where it used to be published with `module` field and it broke another dependency that was doing `require('hoist-non-react-statics')` (https://github.com/kentor/react-click-outside/blob/e3e1337/index.js#L3) because webpack automatically favours ESM. I wonder if I should update the forwarding module (`/npm/`) to have a `.default` check? Perhaps there's some CJS setup somewhere that uses webpack (e.g. to serve bundles in the browser via Karma), and it could break. Didn't want to create a separate issue for this (kinda need more time to repro as well), but thought I'd mention it somewhere.

PR #18187, opened 2020-02-29
Bump react-shallow-renderer to 16.13.1

NMinhNguyen:
See https://github.com/webpack/webpack/issues/7973

### Most recent edit

I've now removed the `module` field from `react-shallow-renderer` to avoid ESM/CJS interop issues and published v16.13.1 of the package (see related PR https://github.com/NMinhNguyen/react-shallow-renderer/pull/34).

### Edit

Just came across https://github.com/webpack/webpack/issues/6584 as well

## Summary

This change fixes a plausible CJS <-> ESM interop issue in webpack that I mentioned in https://github.com/facebook/react/pull/18186#issuecomment-592794066 (see https://github.com/webpack/webpack/issues/7973 for details).

I was actually able to reproduce this issue by tweaking `node_modules` (to use `react-shallow-renderer`) and running `yarn test:karma` in Material-UI's repo - https://github.com/mui-org/material-ui/blob/aa0963253bd682d0c08f21114037749e3e9f0817/package.json#L36

Although it does pick up the ESM version of `react-shallow-renderer`, Enzyme uses `_interopRequireDefault` (https://unpkg.com/browse/enzyme-adapter-react-16@1.15.2/build/ReactSixteenAdapter.js) when `require()`-ing it so things don't actually break. But imagine a proper CJS setup (with webpack) where users are doing `require('react-test-renderer/shallow')`, they could get this `.default` key.

I'm not so concerned about direct imports of `react-shallow-renderer` (as in, not transitively via `react-test-renderer/shallow`) because it's a new package, so I don't think dropping the ESM build is worth it.

Note: this issue doesn't affect Jest (and as a result, CRA).

## Test Plan

I modified `react-test-renderer/shallow` like so in Material-UI's `node_modules`:

```js
const ReactShallowRenderer = require('react-shallow-renderer')

console.log('Resolved path:', require.resolve('react-shallow-renderer'))
console.log({__esModule: ReactShallowRenderer.__esModule, ReactShallowRenderer})

module.exports = ReactShallowRenderer;
```

And got the following output:

> HeadlessChrome 80.0.3987 (Mac OS X 10.15.1) LOG: 'Resolved path: ./node_modules/react-shallow-renderer/esm/index.js'
> HeadlessChrome 80.0.3987 (Mac OS X 10.15.1) LOG: Object{__esModule: true, ReactShallowRenderer: Object{default: function ReactShallowRenderer() { ... }}}

### DIFF
--- a/scripts/rollup/forks.js
+++ b/scripts/rollup/forks.js
@@ -47,6 +47,11 @@ const forks = Object.freeze({
     return 'shared/forks/object-assign.umd.js';
   },
 
+  'react-shallow-renderer': () => {
+    // Use ESM build of `react-shallow-renderer`.
+    return 'react-shallow-renderer/esm/index.js';
+  },
+
   // Without this fork, importing `shared/ReactSharedInternals` inside
   // the `react` package itself would not work due to a cyclical dependency.
   'shared/ReactSharedInternals': (bundleType, entry, dependencies) => {
--- a/packages/react-test-renderer/package.json
+++ b/packages/react-test-renderer/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "object-assign": "^4.1.1",
     "react-is": "^16.8.6",
-    "react-shallow-renderer": "^16.13.0",
+    "react-shallow-renderer": "^16.13.1",
     "scheduler": "^0.19.0"
   },
   "peerDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -10856,10 +10856,10 @@ react-native-web@^0.11.5:
     prop-types "^15.6.0"
     react-timer-mixin "^0.13.4"
 
-react-shallow-renderer@^16.13.0:
-  version "16.13.0"
-  resolved "https://registry.yarnpkg.com/react-shallow-renderer/-/react-shallow-renderer-16.13.0.tgz#18defe59ac922a9623fcbb8bed8db0da7c728fd4"
-  integrity sha512-qdFKyjjHSp5t4+Fs4479JqYDQnBdHi/IZEnGEmZZKZqoiwxjG4DKJPYd7YTaTmalUAqZlZR+K9ayz1+8tEctSg==
+react-shallow-renderer@^16.13.1:
+  version "16.13.1"
+  resolved "https://registry.yarnpkg.com/react-shallow-renderer/-/react-shallow-renderer-16.13.1.tgz#4cfd6dc0f05a8d4d261ff7a80e9b88f15491a00a"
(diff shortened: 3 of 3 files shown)


packages/react-test-renderer/npm/shallow.js
```diff
@@ -1,3 +1,8 @@
 'use strict';
 
-module.exports = require('react-shallow-renderer');
+var ReactShallowRenderer = require('react-shallow-renderer');
+
+// This is hacky but makes it work with webpack preferring to load ESM
+// even if imported via `require()`, and `react-shallow-renderer` ships with ESM.
+// See https://github.com/webpack/webpack/issues/7973
+module.exports = ReactShallowRenderer['default'] || ReactShallowRenderer;
```
NMinhNguyen: I've escaped `.default` as `['default']` intentionally (for super old setups 😅), but happy to make it look more modern.

NMinhNguyen: Ha, ESLint actually wants me to write `.default` anyway..

NMinhNguyen:
Alternatively, I could introduce a breaking change to `react-shallow-renderer` (making it v17.0.0) to only used named exports, and then we tweak `react-test-renderer` to re-export it accordingly (and without interop workarounds).

gaearon:
Exposing the class was always weird (and probably an accident given createRenderer exists). Unfortunately that's what the docs already show so it's likely people use it widely.

NMinhNguyen:
> Exposing the class was always weird (and probably an accident given createRenderer exists). Unfortunately that's what the docs already show so it's likely people use it widely.
 
Yeah I don’t think I should remove the class export. However, given it’s a new package (and it’s unlikely that a lot of people have adopted it already), I’m open to cutting a new major release that uses a named export to export the class. When/if people migrate from `react-test-tenderer/shallow` they’d have to update their imports anyway, and in my opinion it’s not that much more work to grab (destructure in CJS) `ReactShallowRenderer`.

@gaearon has suggested via Twitter DM to remove `module` field which I’m kinda open to doing as well, but I think I slightly prefer switching to named exports. @sebmarkbage do you have any thoughts on this? Removing `module` might mean re-enabling `commonjs` plugin, but maybe that’s an acceptable trade-off?

NMinhNguyen:
One more solution: follow the approach taken in https://github.com/mridgway/hoist-non-react-statics/pull/86 to ensure the CJS version always exports a `.default` property. This would be a breaking change for CJS but ESM would be unaffected.

scripts/rollup/forks.js
```diff
@@ -47,6 +47,11 @@ const forks = Object.freeze({
     return 'shared/forks/object-assign.umd.js';
   },
 
+  'react-shallow-renderer': () => {
```
NMinhNguyen: I can also update this to use the CJS plugin instead.

PR #18194, opened 2020-03-02
Refine PressLegacy blur logic to account for IE11 issue.

trueadm:
This PR fixes an issues that was created in https://github.com/facebook/react/pull/18125. Notbaly, the issue was that IE11 fires `blur` events with a `null` `relatedTarget` when you click a focusable node.

Really, we never really needed the `relatedTarget`. Instead we can use the other part of the logic that checks if the actively pressed target is blurred instead. Given that all our Pressable nodes internally are always focusable, this should provide consistent characteristics across all browsers.

I've also ported the changes/fix to the `Tap` and `Press` event modules.

### DIFF
--- a/packages/react-interactions/events/src/dom/PressLegacy.js
+++ b/packages/react-interactions/events/src/dom/PressLegacy.js
@@ -886,16 +886,9 @@ const pressResponderImpl = {
         break;
       }
       case 'blur': {
-        // If we encounter a blur event that moves focus to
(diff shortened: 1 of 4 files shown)


trueadm:
@necolas I'm going to land this one now as it fixes a high-pri internal issue. If you have any feedback or concerns, I'd be happy to follow up ASAP.

walaura:
tyyyyy!!!

PR #18195, opened 2020-03-02
Modern Event System: add plugin handling and forked paths

trueadm:
This PR adds more of the modern event system logic, enough that based single root/non-portal logic now works and removing the flag should result in most tests passing. Note: this PR does not affect the existing legacy event system, and the only real changes are to the new modern event system (which is behind a flag anyway). This PR consists of:

- `DOMModernPluginEventSystem.js` adds event plugin handling, which is basically the exact same code from the legacy event system (except refactored to use basic for loops)
- `DOMModernPluginEventSystem-test.internal.js` that showes basic capture/bubble events working correctly with the new system (on a single root/non-portal example)
- `ReactBrowserEventEmitter-test.internal.js` uses the correct existing depending on the modern event system flag
- `ReactDOMEventListener-test.js ` disables or alters the expected outcome when using the new system
- `ReactTreeTraversal-test.js` disables or alters the expected outcome when using the new system
- `EnterLeaveEventPlugin.js` alters the logic to account for the new event system having roots rather than a single document listener, plus disables the first ancestor logic for the modern event system.

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -12,6 +12,7 @@ import type {DOMTopLevelEventType} from 'legacy-events/TopLevelEventTypes';
 import type {EventSystemFlags} from 'legacy-events/EventSystemFlags';
 import type {Fiber} from 'react-reconciler/src/ReactFiber';
 import type {PluginModule} from 'legacy-events/PluginModuleType';
+import type {ReactSyntheticEvent} from 'legacy-events/ReactSyntheticEventType';
 
 import {registrationNameDependencies} from 'legacy-events/EventPluginRegistry';
 import {batchedEventUpdates} from 'legacy-events/ReactGenericBatching';
@@ -102,11 +103,11 @@ function dispatchEventsForPlugins(
   rootContainer: Element | Document,
 ): void {
   const nativeEventTarget = getEventTarget(nativeEvent);
-  const syntheticEvents = [];
+  const syntheticEvents: Array<ReactSyntheticEvent> = [];
 
   for (let i = 0; i < plugins.length; i++) {
     const possiblePlugin: PluginModule<AnyNativeEvent> = plugins[i];
-    if (possiblePlugin) {
+    if (possiblePlugin !== undefined) {
       const extractedEvents = possiblePlugin.extractEvents(
         topLevelType,
         targetInst,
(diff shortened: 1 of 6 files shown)


packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -87,6 +92,48 @@ const capturePhaseEvents = new Set([
...
+  eventSystemFlags: EventSystemFlags,
+  nativeEvent: AnyNativeEvent,
+  targetInst: null | Fiber,
+  rootContainer: Element | Document,
+): void {
+  const nativeEventTarget = getEventTarget(nativeEvent);
+  const syntheticEvents = [];
+
+  for (let i = 0; i < plugins.length; i++) {
+    const possiblePlugin: PluginModule<AnyNativeEvent> = plugins[i];
```
threepointone: why isn't plugins typed as Array<PluginModule<AnyNativeEvent>>? could avoid the truthy check. or am I missing how it's used?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -87,6 +92,48 @@ const capturePhaseEvents = new Set([
...
+    if (possiblePlugin) {
```
threepointone: *I* prefer not coercing these values, only because I've been burnt by it in the past. 
```suggestion
    if (possiblePlugin !=  null) {
```

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -87,6 +92,48 @@ const capturePhaseEvents = new Set([
...
+        topLevelType,
+        targetInst,
+        nativeEvent,
+        nativeEventTarget,
+        eventSystemFlags,
+        rootContainer,
+      );
+      if (extractedEvents) {
+        if (isArray(extractedEvents)) {
+          syntheticEvents.push(...(extractedEvents: any));
```
threepointone: why the typecasting here? doesn't flow infer from isArray?

packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -0,0 +1,126 @@
...
+ * @emails react-core
+ */
+
+'use strict';
+
+let React;
+let ReactFeatureFlags;
+let ReactDOM;
+
+function dispatchEvent(element, type) {
```
threepointone: why not have only `dispatchClickEvent`? `dispatchEvent` isn't used elsewhere in this test.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: PluginModule is nullish, so we still have to check for that.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: We have to do because of a Flow problem I couldn't wrap my head around. I added a comment.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: It's because the array has an empty first plugin that isn't included in normal builds (responder plugin), so 0 is undefined.

Issue #15439, opened 2019-04-18
act() should warn in testing frameworks besides jest 


threepointone:
**Do you want to request a *feature* or report a *bug*?**

feature request

**What is the current behavior?**

The warnings for missing act() warnings around updates only happen in jest. 

**What is the expected behavior?**

We should support other test runners/frameworks as well (like jasmine, karma, etc)

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

16.8.0+

eps1lon:
Considering https://github.com/reactwg/react-18/discussions/102 landed in React 18 it seems like we can close this.

In React 18 any `act` usage will will warn if `global.IS_REACT_ACT_ENVIRONMENT` is not set (no reliance on Jest) and no updates outside of `act` and `global.IS_REACT_ACT_ENVIRONMENT` will be considered updates in an e2e test environment i.e. no missing act warning

PR #18196, opened 2020-03-02
WIP: Bring back OSS testing builds

threepointone:
Throwing this at CI to see what breaks.

We removed this in https://github.com/facebook/react/pull/18138 to unblock 16.13. This PR brings it back, with some changes -
- testing builds are prefixed with `unstable-`, i.e - `react-dom/unstable-testing`
- added back to `bundles.js`
- reintroduces the `isTestEnvironment`feature flag. Not used to generate the actual builds, but is used for test specific logic
- flushes suspense fallbacks in prod for testing builds
- changes some tests in TestUtilsAct to reflect this


packages/react-dom/src/__tests__/ReactTestUtilsAct-test.js
```diff
@@ -792,7 +790,7 @@ function runActTests(label, render, unmount, rerender) {
       }
     });
     describe('warn in prod mode', () => {
-      it('warns if you try to use act() in prod mode', () => {
+      xit('warns if you try to use act() in prod mode', () => {
```
threepointone: I could probably remove this test. Or rewrite it to run across the dev/prod+testing matrix.

threepointone:
The fb testing builds increased in size because we were previously using the wrong feature flags.

packages/react/src/ReactSharedInternals.js
```diff
@@ -31,4 +31,10 @@ if (__DEV__) {
   });
 }
 
+if (__DEV__ || isTestEnvironment) {
+  Object.assign(ReactSharedInternals, {
```
threepointone:
```suggestion
ReactSharedInternals.IsSomeRendererActing = IsSomeRendererActing
```

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -962,7 +962,7 @@ function mountEffect(
 ): void {
   if (__DEV__) {
     // $FlowExpectedError - jest isn't a global, and isn't recognized outside of tests
-    if ('undefined' !== typeof jest) {
+    if (__TESTING__ || 'undefined' !== typeof jest) {
```
threepointone: this will "fix" https://github.com/facebook/react/issues/15439

packages/react/unstable-testing.modern.fb.js
```diff
@@ -0,0 +1,10 @@
+/**
```
threepointone: This file may be unnecessary. 🤔

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -789,7 +789,7 @@ function finishConcurrentRender(
         hasNotProcessedNewUpdates &&
         // do not delay if we're inside an act() scope
         !(
-          __DEV__ &&
+          (__DEV__ || __TESTING__) &&
```
acdlite: We need a plan for all the other warnings. Do we need to update all the other warning guards, too?

threepointone:
I don’t understand, how would this setup disable warnings? You’d see CI to break badly if that was the case. 

The act one I modified makes it so the fallback behaviour is the same in prod (but only in testing builds).

packages/react-reconciler/src/ReactFiberWorkLoop.js
threepointone: I don’t think we have to change anything?

eps1lon:
Has this been abandoned? I was a bit confused why I didn't get missing act warnings until I found out that these are only enabled for `jest`. Naively I wanted to propose to check `NODE_ENV`: ```diff
-if ('undefined' !== typeof jest) {
+if (process.env.NODE_ENV === 'test') {
```
which is set by `jest` as well (by default) but this [is somewhat controversial](https://github.com/facebook/jest/issues/3370).

If you still intend to ship extra testing builds (which I prefer over mocking scheduler) I'd like to help. Seems like this PR "only" needs to be updated with `master`?

PR #18201, opened 2020-03-03
Modern Event System: Support nested portal/root boundaries

trueadm:
This PR adds support for multiple nested portal/root boundaries to the modern event system fork. Specifically, where you embed another React root or portal into an existing React sub-tree. Event propagation and the filtering and finding of ancestors is an important role here, and similar to the legacy event system, requires special logic to find the right ancestors to play through the plugin event system.

I've added a bunch of tests and extensive comments around the parts that get added in this PR. Note: this PR should not affect any of the existing legacy event system. I've tested this internally and all tests relating to this behavior seems to work as expected, but given it's all behind a flag that is disabled, the risk should be minimal. Also, if you turn on the flag, all our internal Jest tests pass with this code.

Left to do:

- Add support for legacy internal FB Primer (along with flag)

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -19,6 +19,8 @@ import {batchedEventUpdates} from 'legacy-events/ReactGenericBatching';
 import {executeDispatchesInOrder} from 'legacy-events/EventPluginUtils';
 import {plugins} from 'legacy-events/EventPluginRegistry';
 
+import {HostRoot, HostPortal} from 'shared/ReactWorkTags';
+
 import {trapEventForPluginEventSystem} from './ReactDOMEventListener';
 import getEventTarget from './getEventTarget';
 import {getListenerMapForElement} from './DOMEventListenerMap';
@@ -56,9 +58,10 @@ import {
   TOP_PROGRESS,
   TOP_PLAYING,
 } from './DOMTopLevelEventTypes';
-import {DOCUMENT_NODE} from '../shared/HTMLNodeType';
 
 import {enableLegacyFBPrimerSupport} from 'shared/ReactFeatureFlags';
+import {getClosestInstanceFromNode} from '../client/ReactDOMComponentTree';
+import {DOCUMENT_NODE, COMMENT_NODE} from '../shared/HTMLNodeType';
 
 const capturePhaseEvents = new Set([
   TOP_FOCUS,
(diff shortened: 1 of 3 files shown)


packages/shared/ReactTreeTraversal.js
```diff
@@ -5,7 +5,35 @@
...
+      const portalNode = node.stateNode.containerInfo;
+      while (grandNode !== null) {
+        // If we find a root that is actually a parent in the DOM tree
+        // then we don't continue with getting the parent, as that root
+        // will have its own event listener.
+        if (
+          grandNode.tag === HostRoot &&
+          grandNode.stateNode.containerInfo.contains(portalNode)
+        ) {
+          return null;
```
sebmarkbage: So there are three scenarios:

- The event target is outside the portal so it only affects the root. In this case there's no problem. In either case.
- The event target is in the portal, and the portal itself is outside the root in the DOM. In this case we want it to still bubble through the React tree here which is what traverseTwoPhase normally does.
- The event target is in the portal, and the portal is itself within the root in the DOM. In this case, unless we do something, we risk double firing the event. That's the problem to solve.

The double firing happens because the event can fire using the listeners attached on the root after the portal fires. It can also fire due to the traversal in traverseTwoPhase while it's firing on the portal.

There are two alternative ways to fix this.

- Prevent the traversal from going through to the root in this scenario. I.e. this is the fix you have in this PR.
- Prevent the root event listener from handling the event if the target is in its own portal.

What if instead we prevent the root's event listeners from firing?

Like what if the loop you have in dispatchEventForPluginEventSystem stopped after it found a portal and it wasn't equal to the rootContainer? Then the root's listeners would not be used when the event happens inside a portal. It's always the portal's listeners that handles all events inside a portal.

trueadm: Thank you for the very detailed feedback. I was able to revise my current implementation to address your concerns. Let me know if I've missed anything. :)

sebmarkbage: Posted a follow up PR to show what I mean: https://github.com/facebook/react/pull/18215/

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -224,6 +227,45 @@ export function dispatchEventForPluginEventSystem(
...
+    // sub-tree for that root and make that our ancestor instance.
+    while (true) {
+      if (node === null) {
+        return;
+      }
+      const tag = node.tag;
+      if (tag === HostRoot || tag === HostPortal) {
+        const container = node.stateNode.containerInfo;
+        if (container !== rootContainer && tag === HostPortal) {
+          return;
```
sebmarkbage: I originally tried this but had to add this mess: https://github.com/facebook/react/pull/18215/files#diff-6b784480f6f3297c42bc2ee46e6f943R212

Because I think that this is not sufficient if the Portal happened to be a Portal from a different root then it shouldn't bail out since the portal won't be bubbling into that root. We should add a test for that.

The longer term fix is probably to have a unique `internalInstanceKey` for each root. That way we don't have to bother with collisions. It'd just be like each root is whole other library.

packages/shared/ReactFeatureFlags.js
```diff
@@ -127,7 +127,7 @@ export const disableMapsAsChildren = false;
 export const warnUnstableRenderSubtreeIntoContainer = false;
 
 // Modern event system where events get registered at roots
-export const enableModernEventSystem = false;
+export const enableModernEventSystem = true;
```
gaearon: Careful. :-)

PR #18202, opened 2020-03-03
Disallow unnecessary dependencies in empty useEffect hook

soonoo:
I admit that there are some valid use cases for over-specifying dependencies in `useEffect`.  
Current eslint plugin for hooks however allows over-specifying dependencies even if nothing is done in `useEffect`.  
https://github.com/facebook/react/blob/8e6a08ea4f3beb2127ec3d1595cf5f83df5816b6/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js#L624-L632

I think over-specifying dependencies in `useEffect` should be considered as valid pattern only when at least something is done in that hook.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -622,11 +622,13 @@ const tests = {
       `,
     },
     {
-      // It is valid for effects to over-specify their deps.
+      // It is valid for effects to over-specify their deps to do some extra works.
       code: normalizeIndent`
         function MyComponent(props) {
           const local = props.local;
-          useEffect(() => {}, [local]);
+          useEffect(() => {
+            scrollToTop();
+          }, [local]);
         }
       `,
     },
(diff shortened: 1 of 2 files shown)


gaearon:
If the Hook is completely empty, surely there's no difference between under-specifying deps or over-specifying them? It's unnecessary in either case. I'm not sure why it would make sense to complain about this case. I'm also not sure why somebody would write an empty Hook.

soonoo:
@gaearon You are right. There is no need for empty hook. Below are my opinion about writing hooks.

1. Empty hook is needless but it can be written in that way anyway. Current test cases are also stating about that. 

2. `useMemo` disallows over-specifying dependeicies even when memoizing empty code block. I think it would be better if rules were applied consistently.
```javascript
const [val, setVal] = useState('')
const func = useMemo(() => {
}, [val])
   ^^^^^  not allowed 
```

gaearon:
>useMemo disallows over-specifying dependeicies even when memoizing empty code block. I think it would be better if rules were applied consistently.

This isn't about it being empty. `useMemo` and `useCallback` disallows over-specifying dependencies in general — which is intentional, because you shouldn't have any reason to do that. Those Hooks bodies are supposed to be pure. Effects are different, because you *might* want to run an effect more often. It's not pure. So we allow over-specifying deps for effects.

PR #18203, opened 2020-03-03
Use accumulateTwoPhaseDispatchesSingle directly

trueadm:
I noticed this when working on the modern event system, so thought I'd add the improvement separately.

We only accumulate two phase dispatches on a single event. So passing it to `accumulateTwoPhaseDispatches` is unecessary in a bunch of places and is just overhead and runtime indirection.

Furthermore, with the modern event system we may want to add an additonal argument to `accumulateTwoPhaseDispatchesSingle` during event plugin phase, so this makes that far simpler to do.

### DIFF
--- a/packages/legacy-events/EventPropagators.js
+++ b/packages/legacy-events/EventPropagators.js
@@ -66,7 +66,7 @@ function accumulateDirectionalDispatches(inst, phase, event) {
  * single traversal for the entire collection of events because each event may
  * have a different target.
  */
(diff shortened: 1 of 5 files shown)


gaearon:
>i'm unaware of the specific consequences

If you're curious — each `accumulate*` function has a `accumulate*Single` version. The difference as I understand is that `accumulate*Single` accepts an event, but `accumulate*` accepts *either* event or an array. Since we're calling it with event alone in those cases, there was no need to call the other version.

PR #18204, opened 2020-03-03
Fix flow types

bvaughn:
While reviewing #17934, I spotted some Flow problems that should have been caught. After investigating, this turned out to be caused by a missing `@flow` pragma in the main `React.js` file which was causing a lot of React APIs to be `any` types for DevTools and the interactions package.

### DIFF
--- a/packages/react-devtools-shared/src/__tests__/inspectedElementContext-test.js
+++ b/packages/react-devtools-shared/src/__tests__/inspectedElementContext-test.js
@@ -148,7 +148,7 @@ describe('InspectedElementContext', () => {
 
     // Modern Context API
     const BoolContext = React.createContext(contextData.bool);
-    (BoolContext: any).displayName = 'BoolContext';
+    BoolContext.displayName = 'BoolContext';
 
     class ModernContextType extends React.Component<any> {
       static contextType = BoolContext;
@@ -158,7 +158,7 @@ describe('InspectedElementContext', () => {
     }
 
     const ModernContext = React.createContext();
-    (ModernContext: any).displayName = 'ModernContext';
+    ModernContext.displayName = 'ModernContext';
 
     const container = document.createElement('div');
     await utils.actAsync(() =>
--- a/packages/react-devtools-shared/src/__tests__/inspectedElementContext-test.js
+++ b/packages/react-devtools-shared/src/__tests__/inspectedElementContext-test.js
@@ -148,7 +148,7 @@ describe('InspectedElementContext', () => {
 
     // Modern Context API
     const BoolContext = React.createContext(contextData.bool);
-    BoolContext.displayName = 'BoolContext';
+    (BoolContext: any).displayName = 'BoolContext';
 
     class ModernContextType extends React.Component<any> {
       static contextType = BoolContext;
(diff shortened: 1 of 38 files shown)


packages/react-devtools-shared/src/__tests__/inspectedElementContext-test.js
```diff
@@ -158,7 +158,7 @@ describe('InspectedElementContext', () => {
     }
 
     const ModernContext = React.createContext();
-    ModernContext.displayName = 'ModernContext';
+    (ModernContext: any).displayName = 'ModernContext';
```
bvaughn: I guess I should just add `displayName` to our internal context type. Will do momentarily.

packages/react-devtools-shared/src/devtools/ContextMenu/useContextMenu.js
```diff
@@ -30,11 +31,11 @@ export default function useContextMenu({
...
         const pageX =
-          event.pageX ||
-          (event.touches && ((event: any): TouchEvent).touches[0].pageX);
+          (event: any).pageX ||
+          (event.touches && (event: any).touches[0].pageX);
         const pageY =
-          event.pageY ||
-          (event.touches && ((event: any): TouchEvent).touches[0].pageY);
+          (event: any).pageY ||
+          (event.touches && (event: any).touches[0].pageY);
```
bvaughn: Flow v97 doesn't go very deep here I guess. 🤷‍♂️

packages/react/src/React.js
```diff
@@ -3,6 +3,8 @@
  *
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
```
bvaughn: This was the linchpin change that uncovered all of the masked `any` type errors.

packages/react/src/ReactHooks.js
```diff
@@ -37,7 +37,7 @@ function resolveDispatcher() {
 export function useContext<T>(
   Context: ReactContext<T>,
   unstable_observedBits: number | boolean | void,
-) {
+): T {
```
bvaughn: Our internal definition of `useContext` was broken.

packages/shared/ReactTypes.js
```diff
@@ -66,6 +66,9 @@ export type ReactContext<T> = {
   // DEV only
   _currentRenderer?: Object | null,
   _currentRenderer2?: Object | null,
+  // This value may be added by application code
+  // to improve DEV tooling display names
+  displayName?: string,
```
bvaughn: This helps me avoid a bunch of `: any` casts for DevTools code and matches the [Flow definition](https://github.com/facebook/flow/blob/416d5bc1d75cf70acd81b4cd29b713581883fb6a/lib/react.js#L211-L212).

eps1lon: And would help #18035 :wink:

bvaughn:
CI failure looks unrelated...
Screen Shot 2020-03-03 at 11 30 27 AM

threepointone:
Don’t be certain that it’s unrelated, I saw this recently as a a sign that another task failed. Afk or I’d look for proof.

bvaughn:
The failure is a 401 being returned from the electron repo in its post-install task :smile: Seems unrelated.

threepointone:
Oh no lol

PR #18210, opened 2020-03-04
Modern Event System: Add support for internal FB Primer

trueadm:
This PR exposes a new flag to be used with the modern event system: `enableLegacyFBPrimerSupport`. This flag should be enabled in non-experimental internal bundles only.

When this flag is enabled, we attempt to detect internal FB Primer event heuristics, so we know to defer the internal React dispatching till after Primer has run. We do this using as one-time document event listener.

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -56,6 +56,9 @@ import {
   TOP_PROGRESS,
   TOP_PLAYING,
 } from './DOMTopLevelEventTypes';
+import {DOCUMENT_NODE} from '../shared/HTMLNodeType';
+
+import {enableLegacyFBPrimerSupport} from 'shared/ReactFeatureFlags';
 
 const capturePhaseEvents = new Set([
   TOP_FOCUS,
@@ -165,6 +168,44 @@ export function listenToEvent(
   }
 }
 
+const validFBLegacyPrimerRels = new Set([
+  'dialog',
+  'dialog-post',
+  'async',
+  'async-post',
+  'theater',
+  'toggle',
+]);
+
+function willDeferLaterForFBLegacyPrimer(nativeEvent: any): boolean {
+  let node = nativeEvent.target;
+  const type = nativeEvent.type;
+  if (type !== 'click') {
+    return false;
+  }
+  while (node !== null) {
+    // Primer works by intercepting a click event on an <a> element
+    // that has a "rel" attribute that matches one of the valid ones
+    // in the Set above. If we intercept this before Primer does, we
+    // will need to defer the current event till later and discontinue
+    // execution of the current event. To do this we can add a document
+    // event listener and continue again later after propagation.
+    if (node.tagName === 'A' && validFBLegacyPrimerRels.has(node.rel)) {
+      const legacyFBSupport = true;
+      const isCapture = nativeEvent.eventPhase === 1;
+      trapEventForPluginEventSystem(
+        document,
+        ((type: any): DOMTopLevelEventType),
+        isCapture,
+        legacyFBSupport,
+      );
+      return true;
+    }
+    node = node.parentNode;
+  }
+  return false;
+}
+
 export function dispatchEventForPluginEventSystem(
   topLevelType: DOMTopLevelEventType,
   eventSystemFlags: EventSystemFlags,
@@ -173,6 +214,17 @@ export function dispatchEventForPluginEventSystem(
   rootContainer: Document | Element,
 ): void {
   let ancestorInst = targetInst;
+  if (rootContainer.nodeType !== DOCUMENT_NODE) {
+    // If we detect the FB legacy primer system, we
+    // defer the event to the "document" with a one
+    // time event listener so we can defer the event.
+    if (
+      enableLegacyFBPrimerSupport &&
+      willDeferLaterForFBLegacyPrimer(nativeEvent)
+    ) {
+      return;
+    }
+  }
 
   batchedEventUpdates(() =>
     dispatchEventsForPlugins(
--- a/packages/react-dom/src/events/ReactDOMEventListener.js
+++ b/packages/react-dom/src/events/ReactDOMEventListener.js
@@ -56,6 +56,7 @@ import {passiveBrowserEventsSupported} from './checkPassiveEvents';
 import {
   enableDeprecatedFlareAPI,
   enableModernEventSystem,
+  enableLegacyFBPrimerSupport,
 } from 'shared/ReactFeatureFlags';
 import {
   UserBlockingEvent,
@@ -143,6 +144,7 @@ export function trapEventForPluginEventSystem(
   container: Document | Element,
   topLevelType: DOMTopLevelEventType,
   capture: boolean,
+  legacyFBSupport?: boolean,
 ): void {
   let listener;
   let listenerWrapper;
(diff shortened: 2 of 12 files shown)


packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -119,4 +119,35 @@ describe('DOMModernPluginEventSystem', () => {
...
+    ReactDOM.render(<Test />, container);
+
+    // Fake primer
+    document.addEventListener('click', e => {
+      if (e.target.rel === 'dialog') {
+        log.push('primer');
+      }
+    });
+
+    ReactFeatureFlags.enableLegacyFBPrimerSupport = true;
```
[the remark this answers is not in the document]
trueadm: I didn't ned to set it to false, as they're reset before each test. I'll make that change.

packages/shared/forks/ReactFeatureFlags.www.js
```diff
@@ -97,6 +97,8 @@ export const warnUnstableRenderSubtreeIntoContainer = false;
 
 export const enableModernEventSystem = false;
 
+export const enableLegacyFBPrimerSupport = __EXPERIMENTAL__;
```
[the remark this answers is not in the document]
trueadm: It's meant to be not for experimental builds, so I reversed the flag.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -173,6 +210,17 @@ export function dispatchEventForPluginEventSystem(
   rootContainer: Document | Element,
 ): void {
   let ancestorInst = targetInst;
+  if (rootContainer.nodeType !== DOCUMENT_NODE) {
```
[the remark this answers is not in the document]
trueadm: Well, it should run in all tests when we switch to the modern event system in non experiemental mode. We also run tests with experimental mode, so this won't run there so you get both code-paths. It just depends on the modern event system being on by default.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -165,6 +168,40 @@ export function listenToEvent(
...
+  let node = nativeEvent.target;
+
+  while (node !== null) {
+    // Primer works by intercepting a click event on an <a> element
+    // that has a "rel" attribute that matches one of the valid ones
+    // in the Set above. If we intercept this before Primer does, we
+    // will need to defer the current event till later and discontinue
+    // execution of the current event. To do this we can add a document
+    // event listener and continue again later after propagation.
+    if (node.tagName === 'A' && validFBLegacyPrimerRels.has(node.rel)) {
```
threepointone: will this traversal be run on every event? if so, is it cheap to do?

trueadm: Only for click events.

packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -119,4 +119,33 @@ describe('DOMModernPluginEventSystem', () => {
...
+    // Fake primer
+    document.addEventListener('click', e => {
+      if (e.target.rel === 'dialog') {
+        log.push('primer');
+      }
+    });
+    let aElement = aRef.current;
+    dispatchClickEvent(aElement);
+    expect(onClick).toHaveBeenCalledTimes(1);
+    expect(log).toEqual(['primer']);
```
bvaughn: nit This test might be slightly more robust if it verified that another React-managed ancestor node didn't also get the click event. (As it stands, this test could pass if we just didn't stop propagating anything ever.)

We might also add a test for an anchor tag without one of our whitelisted rels.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -165,6 +168,43 @@ export function listenToEvent(
...
+      const legacyFBSupport = true;
+      trapEventForPluginEventSystem(
+        document,
+        ((type: any): DOMTopLevelEventType),
+        false,
+        legacyFBSupport,
```
bvaughn: What am I missing about the `legacyFBSupport ` param? `trapEventForPluginEventSystem` only accepts 3 params. Does FB override this somewhere I'm not seeing? Is this a mistake?

trueadm: This PR makes `trapEventForPluginEventSystem` take 4 arguments. If you look through the PR you should see it being added. Sometimes the github whitespace makes this harder to see, sorry :/

packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
trueadm: I've updated this test to what I think you meant. Is that okay now?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
bvaughn: Weird. I searched the entire files page and only found the import and declaration on this line. Huh!

bvaughn: Ooh that's because the name was truncated from the diff view because of multi-line param formatting. Oops!

trueadm: Even now? That's some serious caching issue on GitHub's part. I frequently have to do Cmd+Shift+R to make Github update PRs these days though, so it's unsurprising.

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -166,10 +168,40 @@ export function trapEventForPluginEventSystem(
...
+  // due to requirement to provide compatibility with
+  // internal FB www event tooling. This works by removing
+  // the event listener as soon as it is invoked. We could
+  // also attempt to use the {once: true} param on
+  // addEventListener, but that requires support and some
+  // browsers do not support this today, and given this is
+  // to support legacy code patterns, it's likely they'll
+  // need support for such browsers.
+  if (enableLegacyFBPrimerSupport && legacyFBSupport) {
+    const originalListener = listener;
```
bvaughn: What happens if Primer code ever calls `stopImmediatePropagation`? This handler just hangs out until the next click event? (Not that this seems likely to happen I guess)

packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -119,4 +119,54 @@ describe('DOMModernPluginEventSystem', () => {
...
+      return (
+        <div onClick={onDivClick}>
+          <a ref={aRef} href="#" onClick={onClick} rel="dialog-foo">
+            Click me
+          </a>
+        </div>
+      );
+    }
+    ReactDOM.render(<Test2 />, container);
+    dispatchClickEvent(aElement);
```
bvaughn: This is still pointing to the old `<a>` ref.

This test looks good otherwise though.

packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
bvaughn: Yeah. The extra React wrapper check looks good.

packages/react-dom/src/events/ReactDOMEventListener.js
trueadm: It doesn't do that, thankfully.

packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
trueadm: That's intentional. The click should propagate to the `div`.

bvaughn: How is that intentional?

This test is meant to be testing the anchor element with `rel="dialog-foo"` but `aElement` is still pointing to the previous anchor element with `rel="dialog"`. (Your test isn't testing what it's meant to.)

trueadm: Gotcha, I added this change to https://github.com/facebook/react/pull/18201. Thanks for that!

PR #18212, opened 2020-03-04
react-devtools: move electron to optionalDependencies

threepointone:
When we yarn/ci, we download electron only because it's listed in react-devtools as a dependency. We don't seem to use it for any tests or bundles though, so it's non-essential for the build. Further the electron download point is flaky, leading to ci failures like this https://circleci.com/gh/facebook/react/95743 This PR simply moves electron to optionalDependencies, so the build doesn't fail even if the download fails.

### DIFF
--- a/packages/react-devtools/package.json
+++ b/packages/react-devtools/package.json
@@ -23,11 +23,13 @@
     "start": "node bin.js"
   },
   "dependencies": {
-    "cross-spawn": "^5.0.1",
-    "electron": "^5.0.0",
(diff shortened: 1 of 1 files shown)


threepointone:
If the download fails for them, they’d be in the same situation as yarn failing. They’d see the same logs, and devtools wouldn’t start up (and you could detect a failed download in the bin and warn specifically for that)

I’m not super sure myself tbh.

threepointone:
This isn’t the right solution, so I’m closing it

PR #18213, opened 2020-03-04
Fix instanceContainsElem bug

trueadm:
This was an obvious bug on reflection, and most definitely a typo. The function is designed to find if an element is contained within another element in the fiber tree, except the original function mistakenly used `element` instead of `instance`, so we just early return everytime.


trueadm:
I'll add a follow up test tomorrow when I'm in the office.

PR #18215, opened 2020-03-05
Modern Event System: Support nested portal/root boundaries - Alternative approach

sebmarkbage:
This is an alternative to https://github.com/facebook/react/pull/18201. Instead of letting the root handle its events and the portal handle its events, this always lets the portal handle any events that targets the portal.

There's a subtle difference in semantics between these approaches. I added a test to show the capture/bubble phase order for React events.

The other things is whether non-React intermediate components between a Portal and a Root, are allowed to stopPropagation to the root.

An argument for why they shouldn't is that Portals don't behave as if they're rendered at that place in the DOM. For better or worse. They behave as if they're rendered in the place in the React tree. So there is nothing in between that calls stopPropagation in that world. It's counter intuitive but it's how the Portals are currently designed.

This approach sticks closer to the preexisting semantics. For better or worse.

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -204,6 +204,29 @@ export function dispatchEventForPluginEventSystem(
         ) {
           break;
         }
+        if (node.tag === HostPortal) {
+          // The target is a portal, but it's not the rootContainer we're looking for.
+          // Normally portals handle their own events all the way down to the root.
+          // So we should be able to stop now. However, we don't know if this portal
+          // was part of *our* root.
+          let grandNode = node.return;
+          while (grandNode !== null) {
+            if (grandNode.tag === HostRoot || grandNode.tag === HostPortal) {
+              const grandContainer = grandNode.stateNode.containerInfo;
+              if (
+                grandContainer === rootContainer ||
+                (grandContainer.nodeType === COMMENT_NODE &&
+                  grandContainer.parentNode === rootContainer)
+              ) {
+                // This is the rootContainer we're looking for and we found it as
+                // a parent of the Portal. That means we can ignore it because the
+                // Portal will bubble through to us.
+                return;
+              }
+            }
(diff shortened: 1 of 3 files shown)


packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -119,4 +617,375 @@ describe('DOMModernPluginEventSystem', () => {
...
+
+    function Root() {
+      let portalTargetRef = React.useRef(null);
+      React.useLayoutEffect(() => {
+        portalTargetRef.current.appendChild(portalContainer);
+      });
+      return (
+        <div onClick={onClickRoot} onClickCapture={onClickCaptureRoot}>
+          <div ref={portalTargetRef} />
+          {ReactDOM.createPortal(<Portal />, portalContainer)}
```
sebmarkbage: Portal are designed so that if I switch this from this to just `<Portal />` then the behavior is preserved. Then it's more obvious why stopPropagation on the sibling doesn't stop it.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -174,6 +178,66 @@ export function dispatchEventForPluginEventSystem(
...
+          // So we should be able to stop now. However, we don't know if this portal
+          // was part of *our* root.
+          let grandNode = node.return;
+          while (grandNode !== null) {
+            if (grandNode.tag === HostRoot || grandNode.tag === HostPortal) {
+              const grandContainer = grandNode.stateNode.containerInfo;
+              if (
+                grandContainer === rootContainer ||
+                (grandContainer.nodeType === COMMENT_NODE &&
+                  grandContainer.parentNode === rootContainer)
```
trueadm: This is a repeat of the same code about 17 lines up. It could be a separate function.

trueadm:
To simplify things, I ported the changes from this PR into https://github.com/facebook/react/pull/18201.

PR #18216, opened 2020-03-05
Remove event pooling in the modern system

gaearon:
This is a proposal.

Event pooling saves allocations during high firing events, but adds a bit overhead in "releasing", "destroying" and "reusing" instances. I'm not sure this is actually beneficial. Nobody else is doing it so it's not like this is a common practice. I think maybe we can experiment with dropping it altogether. We didn't do this before because it's technically a breaking change, but since we're changing a bunch of things about events anyway, maybe we can sneak it in? It's less of a chance than the move to roots, for example.

Pooling is confusing. For example, not being to access `e.target` in the `setState` updater has been a pain. Although maybe reading that early is best practice so you capture the right value etc. But the way it currently fails is very unintuitive. So maybe it's worth fixing even if we add a lint warning against that. I think the conceptual simplicity win is still valuable.

Effectively this moves all pooling related code behind a flag. I kept `getPooled` because it's scattered across the codebase, but in modern mode this just calls the constructor immediately. I also kept the public API, but `persist` is a no-op and `isPersistent` always returns true in modern mode.

### DIFF
--- a/packages/legacy-events/SyntheticEvent.js
+++ b/packages/legacy-events/SyntheticEvent.js
@@ -9,6 +9,8 @@
 
 import invariant from 'shared/invariant';
 
+import {enableModernEventSystem} from 'shared/ReactFeatureFlags';
+
 const EVENT_POOL_SIZE = 10;
 
 /**
@@ -152,7 +154,10 @@ Object.assign(SyntheticEvent.prototype, {
    * won't be added back into the pool.
    */
   persist: function() {
-    this.isPersistent = functionThatReturnsTrue;
+    // Modern event system doesn't use pooling.
+    if (!enableModernEventSystem) {
+      this.isPersistent = functionThatReturnsTrue;
+    }
   },
 
   /**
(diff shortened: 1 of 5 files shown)


packages/legacy-events/SyntheticEvent.js
```diff
@@ -152,71 +154,79 @@ Object.assign(SyntheticEvent.prototype, {
...
+    : functionThatReturnsFalse,
 
   /**
    * `PooledClass` looks for `destructor` on each instance it releases.
    */
   destructor: function() {
-    const Interface = this.constructor.Interface;
-    for (const propName in Interface) {
+    // Modern event system doesn't use pooling.
+    if (!enableModernEventSystem) {
```
gaearon: Changes below are just whitespace. 

https://github.com/facebook/react/pull/18216/files?w=1

sebmarkbage:
> but since we're changing a bunch of things about events anyway, maybe we can sneak it in?

This is the wrong intuition and I think we should probably wait. There are a ton of things that we should be doing that we're not doing and we really need to hold the line hard. If we were to introduce breaking changes there are a number of other that would be higher pri than this.

Any small breaking change adds to the risk that people can't upgrade to the next major. If they can't upgrade to the next major they can't isolate a small subsection of their site. If they can't do that, they won't be able to adopt the next major after that.

*All* the breaking changes in the next major should only be in service of unlocking the ability to pin a subtree to an old version which is the point of the "modern system". I'd even go as far as saying we shouldn't enable new warnings in 17.0. Only in a future minor.

sebmarkbage:
Behind a different flag that we can do it next one would be make sense though.

packages/legacy-events/SyntheticEvent.js
```diff
@@ -152,71 +154,79 @@ Object.assign(SyntheticEvent.prototype, {
    * won't be added back into the pool.
    */
   persist: function() {
-    this.isPersistent = functionThatReturnsTrue;
+    // Modern event system doesn't use pooling.
+    if (!enableModernEventSystem) {
```
trueadm: We'd probably want to add a warning on using persist(), as it'll just be dead code in apps otherwise.

philipp-spiess:
I wonder if we can go further and remove the heavy synthetic event system that requires all property names to be hardcoded so we can copy them to a small object that only sets the target and prototype links to the nativeEvent?

gaearon:
@philipp-spiess This part of Seb's comment probably applies here:

>There are a ton of things that we should be doing that we're not doing and we really need to hold the line hard. If we were to introduce breaking changes there are a number of other that would be higher pri than this.

trueadm:
I made a follow up to this PR in https://github.com/facebook/react/pull/18969.

PR #18217, opened 2020-03-05
Remove _ctor field from Lazy components

sebmarkbage:
This field is not needed because it's only used before we've initialized, and we don't have anything else to store before we've initialized.

This isn't an important perf optimization by itself but this refactor is motivated by lining it up with the same work on Blocks.

### DIFF
--- a/packages/shared/ReactLazyComponent.js
+++ b/packages/shared/ReactLazyComponent.js
@@ -7,44 +7,14 @@
  * @flow
  */
 
-export type Thenable<T, R> = {
-  then(resolve: (T) => mixed, reject: (mixed) => mixed): R,
-};
+import type {
+  PendingLazyComponent,
+  ResolvedLazyComponent,
+  RejectedLazyComponent,
+  LazyComponent,
+} from 'react/src/ReactLazy';
 
-type UninitializedLazyComponent<T> = {
-  $$typeof: Symbol | number,
-  _status: -1,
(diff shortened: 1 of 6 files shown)


packages/shared/ReactLazyComponent.js
```diff
@@ -9,44 +9,66 @@
...
 }
 
 export function initializeLazyComponentType(
   lazyComponent: LazyComponent<any>,
 ): void {
   if (lazyComponent._status === Uninitialized) {
-    lazyComponent._status = Pending;
-    const ctor = lazyComponent._ctor;
+    let ctor = lazyComponent._result;
+    if (!ctor) {
```
acdlite: Coercion is OK here?

sebmarkbage: When we detect existence of something unknown things we tend to do this type of check.

packages/shared/ReactLazyComponent.js
```diff
@@ -7,46 +7,35 @@
...
-  _ctor: () => Thenable<{default: T, ...}, mixed>,
-  _status: 1,
-  _result: any,
-  ...
-};
-
-export const Uninitialized = -1;
-export const Pending = 0;
-export const Resolved = 1;
-export const Rejected = 2;
```
sebmarkbage: I also upstreamed all of this to the isomorphic package since it owns this data structure. I'll think a bit how we can move the other two helpers out of `/shared`.

PR #18220, opened 2020-03-05
[Blocks] Make it possible to have lazy initialized and lazy loaded Blocks

sebmarkbage:
This builds on a similar design as Lazy in https://github.com/facebook/react/pull/18217

The goal here is to make evaluating the query lazy but still allow render functions from different sources to reconcile against each other - even if they're lazy. In my first approach I hid the lazy loading into the "render" function wrapper which is not going to work since it doesn't reconcile against the same thing anymore then.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -18,7 +18,6 @@ import type {
   SuspenseListTailMode,
 } from './ReactFiberSuspenseComponent';
 import type {SuspenseContext} from './ReactFiberSuspenseContext';
-import type {} from 'react/src/block';
 
 import checkPropTypes from 'shared/checkPropTypes';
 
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -17,6 +18,7 @@ import type {
   SuspenseListTailMode,
 } from './ReactFiberSuspenseComponent';
 import type {SuspenseContext} from './ReactFiberSuspenseContext';
+import type {} from 'react/src/block';
```
gaearon: ?

sebmarkbage:
The changing of `_fn` is same as Lazy. It changes. Even the functions have different call signatures so Flow should catch it. The thing that Flow doesn't have is "state classes" so there's no way to transition to a different type while staying in the type system so I have to briefly use "any" when changing the state. (It also has implications on threading.)

The initialization is unfortunate. There's a lot going on here. We can't throw in child fiber atm because that would bubble to the wrong parent potentially. We would also prevent siblings from warming up. I'm also not happy with the extra tuples created or the "if thennable" type check.

It took me like a month to get to this point and I'm still not happy with it. :)

PR #18222, opened 2020-03-05
Break up require/import statements in strings

cpojer:
## Summary

We have a number of dependency extractors in JavaScript tools, like Jest but also in internal FB tooling, that are using RegExp to extract static dependencies and create a graph. I am trying to reduce the amount of extractors we have at FB and noticed the Jest one has a few issues that are hard to fix with regex. I realize that RegExp dependency tracking is inherently imprecise but hope we can bear some minor inconvenience in just a few places.

In this PR I'm adjusting the React files we use to make sure require and import statements don't get extracted by Jest. Right now Jest extracts them, and when it looks up dependencies in reverse it just ignores the ones that don't exist. I'd like to make that fail hard in some tools and this change is a pre-requisite for that.

## Test Plan
When running this [dependencyExtractor](https://github.com/facebook/jest/blob/master/packages/jest-haste-map/src/lib/dependencyExtractor.ts) from Jest (via jest-haste-map) on react-native, the React files (ReactFabric-dev.fb.js etc.) will no longer have invalid dependencies associated with them.


cpojer:
FB: The stack of D20282825 explains the reasons behind this change for FB employees.

acdlite:
A comment definitely, would be extra nice to have a test or lint rule. Otherwise this absolutely will regress.

PR #18223, opened 2020-03-05
Revert "feat: honor displayName of context types"

trueadm:
Reverts facebook/react#18035. This breaks an internal FB test where we try and set the `displayName` on the context type. Otherwise we get this runtime exception:

`TypeError: Cannot set property displayName of #<Object> which has only a getter`

### DIFF
--- a/packages/react/src/ReactContext.js
+++ b/packages/react/src/ReactContext.js
@@ -120,11 +120,6 @@ export function createContext<T>(
           return context.Consumer;
         },
       },
-      displayName: {
-        get() {
-          return context.displayName;
-        },
-      },
     });
     // $FlowFixMe: Flow complains about missing properties because it doesn't understand defineProperty
     context.Consumer = Consumer;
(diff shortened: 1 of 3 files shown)


eps1lon:
@trueadm Could you share what you're doing internally so that we can get this in properly?

eps1lon:
@trueadm Could you share what you're doing internally so that we can get this in properly?

eps1lon:
You mean you're setting the `displayName` on the consumer not context type, right? Like `Context.Consumer.displayName = 'Unused'`? Is this even supported by react-devtools?

gaearon:
Even if it's unsupported, this is a breaking change if not fixed. :-)

eps1lon:
> Even if it's unsupported, this is a breaking change if not fixed. :-)

Sure I can go ahead and add a warning to the setter. I just think this stance locks any object shape once released because someone might set unsupported properties. 

I just need to know what the expected behavior should be.

trueadm:
Sorry, I meant the context consumer object.

PR #18224, opened 2020-03-05
feat: honor displayName of context types

eps1lon:
## Summary

New iteration of #18035 which had to be reverted (#18223) since setting `Context.Consumer.displayName` would throw. 

Setting `Context.Consumer.displayName` ignores the given value but emits warning. As far as I can tell this doesn't do anything on `master`: https://codesandbox.io/s/priceless-hooks-lm0ky

## Test Plan

@trueadm should sign this off since this affects internal tests. If a warning is still too loud we could put it behind a flag that we turn off for the fb builds.

### DIFF
--- a/packages/react/src/ReactContext.js
+++ b/packages/react/src/ReactContext.js
@@ -120,6 +120,11 @@ export function createContext<T>(
           return context.Consumer;
         },
       },
+      displayName: {
+        get() {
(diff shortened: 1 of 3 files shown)


bvaughn:
I ran a local sync on this version and verified that the internal test the previous PR broke passes now.

PR #18233, opened 2020-03-05
[React Native] Add getInspectorDataForViewAtPoint

rickhanlonii:
## Summary
This diff adds `getInspectorDataForViewAtPoint`, accessible through the DevTools hook, which provides an interface to return inspector data about the view touched at a given point. This allows the React Native inspector to ask React for information about a view at an x,y touch point without any knowledge of how each renderer gets the information.

## Implementation

**Paper**
For Paper we've just moved the logic from the [InspectorOverlay](https://github.com/facebook/react-native/blob/c3d072955024824d7ff6113fc9f642d25c462f16/Libraries/Inspector/InspectorOverlay.js#L40-L50) into React Core. This will allow us to delete `getInspectorDataForViewTag` called [here](https://github.com/facebook/react-native/blob/c3d072955024824d7ff6113fc9f642d25c462f16/Libraries/Inspector/Inspector.js#L61) which is necessary since Fabric does not support React tags.

**Fabric**
For Fabric we add a new strategy which queries native for a node using `findNodeAtPoint`, passes the shadow node to `measure`, and then asks React for the fiber information.

## Test Plan

- Manually tested in Fabric and Paper for React Native.

### DIFF
--- a/packages/react-native-renderer/src/ReactNativeFiberInspector.js
+++ b/packages/react-native-renderer/src/ReactNativeFiberInspector.js
@@ -188,8 +188,8 @@ if (__DEV__) {
         inspectedView._internalInstanceHandle.stateNode.node,
         locationX,
         locationY,
-        shadowNode => {
-          if (shadowNode == null) {
+        internalInstanceHandle => {
+          if (internalInstanceHandle == null) {
             callback({
               pointerY: locationY,
               frame: {left: 0, top: 0, width: 0, height: 0},
@@ -198,9 +198,9 @@ if (__DEV__) {
           }
 
           closestInstance =
-            shadowNode.stateNode.canonical._internalInstanceHandle;
+            internalInstanceHandle.stateNode.canonical._internalInstanceHandle;
           nativeFabricUIManager.measure(
-            shadowNode.stateNode.node,
+            internalInstanceHandle.stateNode.node,
             (x, y, width, height, pageX, pageY) => {
               callback({
                 pointerY: locationY,
--- a/packages/react-native-renderer/src/ReactNativeFiberInspector.js
+++ b/packages/react-native-renderer/src/ReactNativeFiberInspector.js
@@ -253,7 +253,7 @@ if (__DEV__) {
   ): void => {
     invariant(
       false,
-      'getInspectorDataForViewAtPoint() is not available in production',
+      'getInspectorDataForViewAtPoint() is not available in production.',
     );
   };
 }
--- a/packages/react-native-renderer/src/ReactNativeFiberInspector.js
+++ b/packages/react-native-renderer/src/ReactNativeFiberInspector.js
@@ -8,6 +8,7 @@
  */
 
 import type {Fiber} from 'react-reconciler/src/ReactFiber';
+import type {ReactFabricType, HostComponent as ReactNativeHostComponent} from './ReactNativeTypes';
 
 import {
   findCurrentHostFiber,
@@ -27,6 +28,7 @@ if (__DEV__) {
 }
 
 let getInspectorDataForViewTag;
+let getInspectorDataForViewAtPoint;
 
 if (__DEV__) {
   const traverseOwnerTreeUp = function(hierarchy, instance: any) {
@@ -89,6 +91,34 @@ if (__DEV__) {
     }));
   };
 
+  const getInspectorDataForInstance = function(closestInstance, frame): Object {
+    // Handle case where user clicks outside of ReactNative
+    if (!closestInstance) {
+      return {
+        hierarchy: [],
+        props: emptyObject,
+        selection: null,
+        source: null,
+      };
+    }
+
+    const fiber = findCurrentFiberUsingSlowPath(closestInstance);
+    const fiberHierarchy = getOwnerHierarchy(fiber);
+    const instance = lastNonHostInstance(fiberHierarchy);
+    const hierarchy = createHierarchy(fiberHierarchy);
+    const props = getHostProps(instance);
+    const source = instance._debugSource;
+    const selection = fiberHierarchy.indexOf(instance);
+
+    return {
+      hierarchy,
+      frame,
+      props,
+      selection,
+      source,
+    };
+  }
+
   getInspectorDataForViewTag = function(viewTag: number): Object {
     const closestInstance = getClosestInstanceFromNode(viewTag);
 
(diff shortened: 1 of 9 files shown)


packages/react-native-renderer/src/ReactNativeFiberInspector.js
```diff
@@ -117,13 +150,91 @@ if (__DEV__) {
...
+            left,
+            top,
+            width,
+            height,
+          };
+          const inspectorData = getInspectorDataForInstance(
+            getClosestInstanceFromNode(nativeViewTag),
+            frame,
+          );
+          callback({...inspectorData, nativeViewTag});
```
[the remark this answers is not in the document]
rickhanlonii: Nice catch, this is used I just used the wrong name. This is returned so that we can pass the view tag to the DevTools in paper so we don't regress (the DevTools will need updated for this feature to work in Fabric).

packages/react-native-renderer/src/ReactNativeFiberInspector.js
```diff
@@ -117,13 +150,91 @@ if (__DEV__) {
...
+        [x, y],
+        (nativeViewTag, left, top, width, height) => {
+          frame = {
+            left,
+            top,
+            width,
+            height,
+          };
+          const inspectorData = getInspectorDataForInstance(
+            getClosestInstanceFromNode(nativeViewTag),
```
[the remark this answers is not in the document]
rickhanlonii: This is the first line in `getInspectorDataForViewTag` (the rest of that function is the same as `getInspectorDataForInstance`), I do this here so that it's easier to delete the tag version later (after this change that function is only exported.

packages/react-native-renderer/src/ReactNativeFiberInspector.js
```diff
@@ -89,6 +91,34 @@ if (__DEV__) {
     }));
   };
 
+  const getInspectorDataForInstance = function(closestInstance, frame): Object {
```
TheSavior: Since this is returning Object, you won't get type safety that `getInspectorDataForViewAtPoint` is giving you `TouchedViewDataAtPoint`

packages/react-native-renderer/src/ReactNativeFiberInspector.js
```diff
@@ -117,13 +147,93 @@ if (__DEV__) {
...
+      height: 0,
+    };
+
+    if (inspectedView._internalInstanceHandle != null) {
+      // For Fabric we can look up the instance handle directly and measure it.
+      nativeFabricUIManager.findNodeAtPoint(
+        inspectedView._internalInstanceHandle.stateNode.node,
+        locationX,
+        locationY,
+        shadowNode => {
```
TheSavior: We should probably call this `internalInstanceHandle`

packages/react-native-renderer/src/ReactNativeFiberInspector.js
```diff
@@ -117,13 +147,93 @@ if (__DEV__) {
...
+
+  getInspectorDataForViewAtPoint = function(
+    findNodeHandle: (componentOrHandle: any) => ?number,
+    inspectedView: Object,
+    locationX: number,
+    locationY: number,
+    callback: (viewData: TouchedViewDataAtPoint) => mixed,
+  ): void {
+    let closestInstance = null;
+    let frame = {
```
TheSavior: Do you still need to define this here since you define it inline in the callback?

scripts/error-codes/codes.json
```diff
@@ -346,5 +346,6 @@
   "345": "Root did not complete. This is a bug in React.",
   "346": "An event responder context was used outside of an event cycle.",
   "347": "Maps are not valid as a React child (found: %s). Consider converting children to an array of keyed ReactElements instead.",
-  "348": "ensureListeningTo(): received a container that was not an element node. This is likely a bug in React."
+  "348": "ensureListeningTo(): received a container that was not an element node. This is likely a bug in React.",
+  "349": "getInspectorDataForViewAtPoint() is not available in production"
```
TheSavior: It looks like these often end with a period.

packages/react-native-renderer/src/ReactNativeFiberInspector.js
rickhanlonii: Yes because of the early return but I moved it inline for the early return.

packages/react-native-renderer/src/ReactNativeFiberInspector.js
rickhanlonii: Nice catch, fixed.

packages/react-native-renderer/src/ReactNativeTypes.js
```diff
@@ -100,22 +100,44 @@ type SecretInternalsType = {
...
+
+type InspectorDataGetter = (
+  (componentOrHandle: any) => ?number,
+) => $ReadOnly<{|
+  measure: Function,
+  props: InspectorDataProps,
+  source: InspectorDataSource,
+|}>;
+
+export type InspectorData = $ReadOnly<{
```
TheSavior: Should these be exact objects?

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -10,6 +10,7 @@
 import type {Fiber} from './ReactFiber';
 import type {FiberRoot} from './ReactFiberRoot';
 import type {RootTag} from 'shared/ReactRootTags';
+import type {TouchedViewDataAtPoint} from 'react-native-renderer/src/ReactNativeTypes';
```
bvaughn: This import is causing CI to fail.

packages/react-native-renderer/src/ReactNativeTypes.js
```diff
@@ -100,6 +100,46 @@ type SecretInternalsType = {
...
+    getInspectorData: InspectorDataGetter,
+  |}>,
+  selection: ?number,
+  props: InspectorDataProps,
+  source: ?InspectorDataSource,
+|}>;
+
+export type TouchedViewDataAtPoint = $ReadOnly<{|
+  pointerY: number,
+  touchedViewTag?: number,
```
sebmarkbage: What is this used for? I thought we're getting rid of all these tags so seems odd to add new users of it and expose more implementation details about to change to future work. Shouldn't there be another mechanism?

sebmarkbage:
I reverted this to unblock the sync but it's probably an easy fix. Feel free to reland after fixing the Flow errors etc.

packages/react-reconciler/src/ReactFiberReconciler.js
sebmarkbage: We can't have dependencies on RN from the reconciler bundle. The folders are checked independently.

Also from a layering perspective it's probably not right.

bvaughn: I'm a little confused that some of the commits seemed to pass CI

packages/react-reconciler/src/ReactFiberReconciler.js
```diff
@@ -109,6 +110,13 @@ type DevToolsConfig = {|
   // This API is unfortunately RN-specific.
   // TODO: Change it to accept Fiber instead and type it properly.
   getInspectorDataForViewTag?: (tag: number) => Object,
+  // Used by RN in-app inspector.
+  getInspectorDataForViewAtPoint?: (
```
sebmarkbage: I don't think either of these two should be part of this type because they're not actually used by the core DevTools protocol. It's kind of an adhoc RN specific addition.

Instead we should make this a renderer specific object.

```
rendererConfig: RendererInspectionConfig
```

Then you can define the type for RendererInspectionConfig through the HostConfig mechanism. So that each renderer gets their own type. That can then be safely referenced in the react-native folder.

So Fabric and RN host configs gets this type:

```
export type RendererInspectionConfig = {
  getInspectorDataForViewTag?: (tag: number) => Object,
  getInspectorDataForViewAtPoint?: (...) => ...;
};
```

You then import that from here using:

```
import type {RendererInspectionConfig} from './ReactFiberHostConfig';
```

bvaughn: I like this suggestion.

packages/react-native-renderer/src/ReactNativeTypes.js
rickhanlonii: This is to support the current devtools behavior in Paper where tapping in the app will open the element in the tree. I was thinking we would add this now to prevent the regression, while we figure out the correct solution for Fabric and Paper moving forward.

https://github.com/facebook/react-native/blob/84b7b8d07645b528c784e0bacecf1293797e0ec7/Libraries/Inspector/Inspector.js#L209-L212

What do you think?

PR #18234, opened 2020-03-06
Add test run that uses www feature flags

acdlite:
In CI, we run our test suite against multiple build configurations. For example, we run our tests in both dev and prod, and in both the experimental and stable release channels. This is to prevent accidental deviations in behavior between the different builds. If there's an intentional deviation in behavior, the test author must account for them.

However, we currently don't run tests against the www builds. That's a problem, because it's common for features to land in www before they land anywhere else, including the experimental release channel. Typically we do this so we can gradually roll out the feature behind a flag before deciding to enable it.

The way we test those features today is by mutating the `shared/ReactFeatureFlags` module. There are a few downsides to this approach, though. The flag is only overridden for the specific tests or test suites where you apply the override. But usually what you want is to run *all* tests with the flag enabled, to protect against unexpected regressions.

Also, mutating the feature flags module only works when running the tests against source, not against the final build artifacts, because the ReactFeatureFlags module is inlined by the build script.

Instead, we should run the test suite against the www configuration, just like we do for prod, experimental, and so on. I've added a new command, `yarn test-www`. It automatically runs in CI.

Some of the www feature flags are dynamic; that is, they depend on a runtime condition (i.e. a GK). These flags are imported from an external module that lives in www. Those flags will be enabled for some clients and disabled for others, so we should run the tests against *both* modes.

So I've added a new global `__VARIANT__`, and a new test command `yarn test-www-variant`. `__VARIANT__` is set to false by default; when running `test-www-variant`, it's set to true.

If we were going for *really* comprehensive coverage, we would run the tests against every possible configuration of feature flags: 2 ^ numberOfFlags total combinations. That's not practical, though, so instead we only run against two combinations: once with `__VARIANT__` set to `true`, and once with it set to `false`. We generally assume that flags can be toggled independently, so in practice this should be enough.

You can also refer to `__VARIANT__` in tests to detect which mode you're running in. Or, you can import `shared/ReactFeatureFlags` and read the specific flag you can about. However, we should stop mutating that module going forward. Treat it as read-only.

In this commit, I have only setup the www tests to run against source. I'll leave running against build for a follow up.

Many of our tests currently assume they run only in the default configuration, and break when certain flags are toggled. Rather than fix these all up front, I've hard-coded the relevant flags to the default values. We can incrementally migrate those tests later.

### DIFF
--- a/packages/shared/forks/ReactFeatureFlags.www.js
+++ b/packages/shared/forks/ReactFeatureFlags.www.js
@@ -9,8 +9,11 @@
 
 import typeof * as FeatureFlagsType from 'shared/ReactFeatureFlags';
 import typeof * as ExportsType from './ReactFeatureFlags.www';
+import typeof * as DynamicFeatureFlags from './ReactFeatureFlags.www-dynamic';
 
 // Re-export dynamic flags from the www version.
+const dynamicFeatureFlags: DynamicFeatureFlags = require('ReactFeatureFlags');
+
 export const {
   debugRenderPhaseSideEffectsForStrictMode,
   deferPassiveEffectCleanupDuringUnmount,
@@ -20,8 +23,9 @@ export const {
   warnAboutShorthandPropertyCollision,
   disableSchedulerTimeoutBasedOnReactExpirationTime,
   warnAboutSpreadingKeyToJSX,
+  replayFailedUnitOfWorkWithInvokeGuardedCallback,
   enableModernEventSystem,
-} = require('ReactFeatureFlags');
+} = dynamicFeatureFlags;
 
 // On WWW, __EXPERIMENTAL__ is used for a new modern build.
 // It's not used anywhere in production yet.
@@ -39,7 +43,6 @@ export const enableProfilerCommitHooks = false;
 export const enableSchedulerTracing = __PROFILE__;
 export const enableSchedulerDebugging = true;
 
-export const replayFailedUnitOfWorkWithInvokeGuardedCallback = false;
 export const warnAboutDeprecatedLifecycles = true;
 export const disableLegacyContext = __EXPERIMENTAL__;
 export const warnAboutStringRefs = false;
--- a/scripts/jest/setupTests.www.js
+++ b/scripts/jest/setupTests.www.js
@@ -1,0 +1,23 @@
+'use strict';
+
+jest.mock('shared/ReactFeatureFlags', () => {
+  jest.mock(
+    'ReactFeatureFlags',
+    () => jest.requireActual('shared/forks/ReactFeatureFlags.www-dynamic'),
+    {virtual: true}
+  );
+
+  const wwwFlags = jest.requireActual('shared/forks/ReactFeatureFlags.www');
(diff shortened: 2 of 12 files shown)


scripts/jest/setupTests.www.js
```diff
@@ -0,0 +1,23 @@
...
+  const defaultFlags = jest.requireActual('shared/ReactFeatureFlags');
+
+  // TODO: Many tests were written before we started running them against the
+  // www configuration. Update those tests so that they work against the www
+  // configuration, too. Then remove these overrides.
+  wwwFlags.disableLegacyContext = defaultFlags.disableLegacyContext;
+  wwwFlags.warnAboutUnmockedScheduler = defaultFlags.warnAboutUnmockedScheduler;
+  wwwFlags.enableUserTimingAPI = defaultFlags.enableUserTimingAPI;
+  wwwFlags.disableJavaScriptURLs = defaultFlags.disableJavaScriptURLs;
+  wwwFlags.enableDeprecatedFlareAPI = defaultFlags.enableDeprecatedFlareAPI;
```
acdlite: @trueadm When I ran the tests with Flare enabled, this flag started failing sporadically:

```js
Summary of all failing tests
 FAIL  packages/react-dom/src/__tests__/ReactDOMFiber-test.js
  ● ReactDOMFiber › should not update event handlers until commit

    Expected warning was not recorded:
      "Warning: unstable_flushDiscreteUpdates: Cannot flush updates when React is already rendering."

      1051 |         expect(() => {
      1052 |           node.click();
    > 1053 |         }).toErrorDev(
           |            ^
      1054 |           'Warning: unstable_flushDiscreteUpdates: Cannot flush updates when React is already rendering.',
      1055 |         );
      1056 |       }
```

Details: https://circleci.com/gh/facebook/react/97022?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Not sure if this is an issue or not, but letting you know just in case. Since Flare is enabled in the www build.

packages/shared/forks/ReactFeatureFlags.www.js
```diff
@@ -9,19 +9,22 @@
...
 export const {
-  debugRenderPhaseSideEffectsForStrictMode,
   deferPassiveEffectCleanupDuringUnmount,
   disableInputAttributeSyncing,
   enableTrustedTypesIntegration,
   runAllPassiveEffectDestroysBeforeCreates,
   warnAboutShorthandPropertyCollision,
   disableSchedulerTimeoutBasedOnReactExpirationTime,
   warnAboutSpreadingKeyToJSX,
+  replayFailedUnitOfWorkWithInvokeGuardedCallback,
```
acdlite: Noticed this was off in the www build. I think maybe we wanted to ramp this up in case it breaks something? Seems unlikely, but to decouple it from the rest of the changes I've made it a dynamic flag.

scripts/jest/setupTests.www.js
trueadm: This is known, and it seemed like maybe a fragile test from memory as it’s related to the discrete flushing logic based on the timeStamp logic.

PR #18240, opened 2020-03-06
Refactor Host Config Infra (getting rid of .inline*.js)

sebmarkbage:
I kept forgetting what the various .inline files are supposed to do, and it made it hard to add more combinations.

The first thing this does is getting rid of the various renderer depending on the indirection pseudo entry point (inline.dom.js etc). Instead we just deep require the reconciler. We do this for other files anyway and there's nothing really preventing this. Instead, we can just use the entry-point of the actual renderer as the jest hook to initialize the correct host config.

The actual react-reconciler/index.js entry point should only represent the public API (currently a function wrapper published to npm). Not used by us.

There are no longer any untyped files so we don't need the special inline-typed form. The untyped files caused some uncertainty since once an untyped file gets into the system it leaks the any all over the place. It also wasn't a sufficient approach (I had some hacks to hide DOM).

Instead, for Flow, in this approach we explicitly select which paths we want to check for each renderer. The inverse gets ignored. So every path is still covered by some renderer.

These ignores are not untyped so if a dependency accidentally tries to reach into something that isn't covered, then that will be an error. Not silently leak "any".


sebmarkbage:
Oh, this also delete react-reconciler/persistent since that's just a flag on host config now.

scripts/shared/inlinedHostConfigs.js
```diff
@@ -16,52 +16,77 @@ module.exports = [
       'react-flight-dom-webpack/server.node',
       'react-flight-dom-webpack',
     ],
+    paths: [
+      'react-dom',
+      'react-dom/unstable-fizz',
+      'react-dom/unstable-fizz.node',
+      'react-dom/src/server/ReactDOMFizzServerNode.js', // react-dom/unstable-fizz.node
```
sebmarkbage: These are explicitly called out so they get ignored in "dom-browser" and vice versa.

PR #18241, opened 2020-03-06
Eager bailouts only for use state

### LINKED ISSUE #15088: useReducer - eagerReducer optimization discussion/questions

Andarist:
## Summary

fixes #15088 closes #15198

I've stumbled upon an edge case where bailed out actions were kept in the update queue and could be applied using the newly rendered reducer during future rerender. This can be observed here - https://codesandbox.io/s/silly-paper-22org . Just click the disable button, click the increment button multiple times and click the disable button once again. The counter will jump to the number of times you have clicked on the incrementing button while it really should still show 0.

@sebmarkbage has suggested ditching the eager bailout optimization for the useReducer case altogether [here](https://github.com/facebook/react/pull/15198#issuecomment-561364762) and @gaearon has [recently encouraged me](https://github.com/facebook/react/pull/15198#issuecomment-593970984) to just follow this suggestion.

@sebmarkbage has also suggested that "We should also forbid suspending in a setState reducer.", but I'm not sure what that means exactly and how it should be approached so any tips regarding that would be highly appreciated.

As to the implementation - I've managed to remove the need to store "lastRenderedReducer" at all and also I've decided to only append updates to the queue if the rerender is scheduled. In the current implementation on the master branch, the queue was slightly prone to a minor memory leak because theoretically, it was possible to append bailed out updates until the next rerender and that, theoretically, could never happen. This is not a real-life scenario, but it still seems better to avoid appending to the queue unnecessarily to avoid going through those updates during the next render.

## Test Plan

I've run tests, lint & flow dom tasks before sending this as PR.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -872,8 +872,6 @@ function updateState<S, I, A>(
     'Should have a queue. This is likely a bug in React. Please file an issue.',
   );
 
-  queue.lastRenderedReducer = basicStateReducer;
-
   const current: Hook = (currentHook: any);
 
   // The last rebase update that is NOT part of the base state.
@@ -1000,8 +998,6 @@ function rerenderState<S, I, A>(
     queue !== null,
     'Should have a queue. This is likely a bug in React. Please file an issue.',
   );
-
-  queue.lastRenderedReducer = basicStateReducer;
 
   // This is a re-render. Apply the new render phase updates to the previous
   // work-in-progress hook.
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -3173,6 +3123,59 @@ function loadModules({
         expect(ReactNoop).toMatchRenderedOutput('5');
       });
 
+      it('should not store dispatched items in the queue (TODO: better description)', () => {
```
Andarist: I'm not sure how to best describe this test as it sort of tests the mechanism that is no longer here (eager bailouts for use reducer).

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1296,22 +1303,73 @@ function dispatchAction<S, A>(
...
+      if ('undefined' !== typeof jest) {
+        warnIfNotScopedWithMatchingAct(fiber);
+        warnIfNotCurrentlyActingUpdatesInDev(fiber);
+      }
+    }
+    scheduleWork(fiber, expirationTime);
+  }
+}
+
+function setState<S>(
```
Andarist: For now this is being split from the `dispatchAction` but they both share pretty much the same implementation - besides the fact that `dispatchAction` has no eager bailout implemented. 

Both could be collapsed into a single function and the eager bailout could only be attempted for `reducer === basicStateReducer` but to do that I would have to bring back keeping track of `lastRenderedReducer`. 

Let me know which approach do you prefer.

gaearon:
So we've been looking at this. It's probably the right thing to do. But it's a scary change. So we'd want it behind a flag. The current implementation in this PR would be hard to put behind a flag. It would also be hard to maintain two versions in one place.

We plan to fork the reconciler with a few other breaking and more ambitious changes. So we'll probably fix this as a part of that refactor. Thank you for starting this work though (and for flagging the issue in the first place). It's been really helpful.

PR #18242, opened 2020-03-07
Add Relay Flight Build

sebmarkbage:
This will be used to host a Flight Server and Flight Client inside Relay. Currently, there's only an FB www build because I'm not sure how we'll want to sync it to the OSS repo yet. Most likely it'll end up being an RN build instead since RN is Relay's source of truth.

We'll need another host config option for this configuration so I added `dom-relay` for that.

This is currently encoding to a JSON array of strings but I'll refactor to a different format/layering in a follow up.

Renamed some flight files while I was at it.


josephsavona:
Thanks!

PR #18248, opened 2020-03-08
[DevTools] Add shortcut keys for tab switching

### LINKED ISSUE #18226: DevTools: Add shortcut keys for standalone version
Suggested keys:
* ⌘1 - Switch to "Components" tab
* ⌘2 - Switch to "Profiler" tab

Note that these shortcut keys should **only** be implemented for the [standalone DevTools](https://www.npmjs.com/package/react-devtools) since (1) the browser already uses these shortcut keys and (2) the "Component" and "Profiler" tabs are only part of DevTools in the standalone version anyway.

kerolloz:
## Summary

Fixes #18226

## Test Plan

gif

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/DevTools.js
+++ b/packages/react-devtools-shared/src/devtools/views/DevTools.js
@@ -107,10 +107,14 @@ export default function DevTools({
   viewAttributeSourceFunction,
   viewElementSourceFunction,
 }: Props) {
-  const [tab, setTab] = useLocalStorage<TabID>(
+  let [tab, setTab] = useLocalStorage<TabID>(
     'React::DevTools::defaultTab',
-    overrideTab != null ? overrideTab : defaultTab,
+    defaultTab,
   );
+
+  if (overrideTab != null) {
+    tab = overrideTab;
+  }
 
   const viewElementSource = useMemo(
     () => ({
--- a/packages/react-devtools-shared/src/devtools/views/DevTools.js
+++ b/packages/react-devtools-shared/src/devtools/views/DevTools.js
@@ -13,7 +13,7 @@ import '@reach/menu-button/styles.css';
 import '@reach/tooltip/styles.css';
 
 import * as React from 'react';
-import {useEffect, useMemo, useState, useRef} from 'react';
+import {useEffect, useMemo, useRef} from 'react';
 import Store from '../store';
 import {BridgeContext, ContextMenuContext, StoreContext} from './context';
 import Components from './Components/Components';
@@ -27,6 +27,7 @@ import {ModalDialogContextController} from './ModalDialog';
 import ReactLogo from './ReactLogo';
 import UnsupportedVersionDialog from './UnsupportedVersionDialog';
 import WarnIfLegacyBackendDetected from './WarnIfLegacyBackendDetected';
+import {useLocalStorage} from './hooks';
 
 import styles from './DevTools.css';
 
@@ -106,10 +107,10 @@ export default function DevTools({
   viewAttributeSourceFunction,
   viewElementSourceFunction,
 }: Props) {
-  const [tab, setTab] = useState(defaultTab);
-  if (overrideTab != null && overrideTab !== tab) {
-    setTab(overrideTab);
-  }
+  const [tab, setTab] = useLocalStorage<TabID>(
+    'React::DevTools::defaultTab',
+    overrideTab != null ? overrideTab : defaultTab,
+  );
 
   const viewElementSource = useMemo(
     () => ({
@@ -130,32 +131,36 @@ export default function DevTools({
   const devToolsRef = useRef<HTMLElement | null>(null);
 
   useEffect(() => {
-    if (devToolsRef.current === null) return () => {};
-
-    if (showTabBar) {
-      const ownerWindow = devToolsRef.current.ownerDocument.defaultView;
-      const handleKeyDown = (event: KeyboardEvent) => {
-        if (event.ctrlKey || event.metaKey) {
-          switch (event.key) {
-            case '1':
-              setTab(tabs[0].id);
-              event.preventDefault();
-              event.stopPropagation();
-              break;
-            case '2':
-              setTab(tabs[1].id);
-              event.preventDefault();
-              event.stopPropagation();
-              break;
-          }
-        }
-      };
-
-      ownerWindow.addEventListener('keydown', handleKeyDown);
-      return () => {
-        ownerWindow.removeEventListener('keydown', handleKeyDown);
-      };
+    if (!showTabBar) {
+      return;
     }
+
+    const div = devToolsRef.current;
+    if (div === null) {
+      return;
+    }
+
+    const ownerWindow = div.ownerDocument.defaultView;
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (event.ctrlKey || event.metaKey) {
+        switch (event.key) {
+          case '1':
+            setTab(tabs[0].id);
+            event.preventDefault();
+            event.stopPropagation();
+            break;
+          case '2':
+            setTab(tabs[1].id);
+            event.preventDefault();
+            event.stopPropagation();
+            break;
+        }
+      }
+    };
+    ownerWindow.addEventListener('keydown', handleKeyDown);
+    return () => {
+      ownerWindow.removeEventListener('keydown', handleKeyDown);
+    };
   }, [showTabBar]);
 
   useEffect(() => {
--- a/packages/react-devtools-shared/src/devtools/views/DevTools.js
+++ b/packages/react-devtools-shared/src/devtools/views/DevTools.js
@@ -13,7 +13,7 @@ import '@reach/menu-button/styles.css';
 import '@reach/tooltip/styles.css';
 
 import * as React from 'react';
-import {useEffect, useMemo, useState} from 'react';
+import {useEffect, useMemo, useState, useRef} from 'react';
 import Store from '../store';
 import {BridgeContext, ContextMenuContext, StoreContext} from './context';
 import Components from './Components/Components';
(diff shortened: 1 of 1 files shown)


kerolloz:
@bvaughn 
what do you think , any thoughts?

kerolloz:
@strdr4605
No need to reinvent the wheel to implement a simple task. Using a well tested package can save you tons of bugs. It's my opinion, I respect yours, I see what you are saying.

packages/react-devtools-shared/src/devtools/views/DevTools.js
```diff
@@ -127,6 +128,15 @@ export default function DevTools({
     [enabledInspectedElementContextMenu, viewAttributeSourceFunction],
   );
 
+  const switchTabKeys = ['ctrl+1', 'ctrl+2', 'command+1', 'command+2'];
+  Mousetrap.bind(switchTabKeys, (e, keyCombination) => {
```
bvaughn: This is a side effect, so it would need to be inside of a `useEffect`:
```js
useEffect(() => {
  if (showTabBar) {
    Mousetrap.bind(...)
  }
}, [showTabBar]);

**Comment by bvaughn** [comment]
@ermanzohre Approving PRs without leaving feedback or comments is not really helpful.

**Comment by kerolloz** [comment]
@bvaughn 
I don't get how should I bind to `ownerDocument.defaultView`.
What about the following approach so far? except for `window.addEventListener`.
```javascript
useEffect(() => {
  if (showTabBar) {
    window.addEventListener('keydown', (e: KeyboardEvent) => {
      if ((e.ctrlKey || e.metaKey) && '12'.includes(e.key)) {
        const tabNumber = +e.key - 1;
        if (tabs[tabNumber].id !== tab) {
          console.log('switching');
          setTab(tabs[tabNumber].id);
        }
      }
      // e.preventDefault();
      // e.stopPropagation();
    });
  }
}, [showTabBar, tab]);
```
I know it has one problem, it listens to the event multiple times

bvaughn:
> I don't get how should I bind to `ownerDocument.defaultView`.

Did you check out the commit I linked to? It has an example of doing this.

`'12'.includes(e.key)` looks like an unusual way of checking for "1" or "2" :smile:

Maybe something like this? (Note: I did not test it.)
```js
useEffect(() => {
  // Fin ownerWindow from a ref link in commit be9046a

  if (showTabBar) {
    ownerWindow.addEventListener("keydown", (event: KeyboardEvent) => {
      if (event.ctrlKey || event.metaKey) {
        switch (event.key) {
          case "1":
            setTab(tabs[0].id);
            event.preventDefault();
            event.stopPropagation();
            break;
          case "2":
            setTab(tabs[1].id);
            event.preventDefault();
            event.stopPropagation();
            break;
        }
      }
    });
  }

  return () => {
    // showTabBar should never change, so this cleanup shouldn't be run
    // but it's still good practice to add it anyway! :)
    ownerWindow.removeEventListener("keydown");
  };
}, [showTabBar]);

```

kerolloz:
Yes I checked the commit, but still not sure how to do it. I tried doing

```js
const devtoolsRef = useRef<HTMLElement|null>(null);

useEffect(() => {
  const ownerWindow = devtoolsRef.current.ownerDocument.defaultView;
  const handleKeyDown = (event: KeyboardEvent) => {
    if (event.ctrlKey || event.metaKey) {
      // only switch if choosen tab is not the current one
      const shouldSwitchTabs = tabs[+event.key - 1].id !== tab;
      if (shouldSwitchTabs) {
        switch (event.key) {
          case '1':
            setTab(tabs[0].id);
            event.preventDefault();
            event.stopPropagation();
            break;
          case '2':
            setTab(tabs[1].id);
            event.preventDefault();
            event.stopPropagation();
            break;
        }
      }
    }
  };
  if (showTabBar) {
    ownerWindow.addEventListener('keydown', handleKeyDown);
  }

  return () => {
    // showTabBar should never change, so this cleanup shouldn't be run
    // but it's still good practice to add it anyway! :)
    ownerWindow.removeEventListener('keydown', handleKeyDown);
  };
}, [showTabBar, tab, tabs, devtoolsRef]);


...

return (
    <BridgeContext.Provider value={bridge} ref={devtoolsRef}>
```

I get `TypeError: devtoolsRef.current is null`.

strdr4605:
Because you assign the devtoolsRef first with null you should check it in the `useEffect`

maybe this will help

```js
const devtoolsRef = useRef<HTMLElement|null>(null);

useEffect(() => {
  if (!devtoolsRef.current) return;

  ...  

}, [showTabBar, tab, tabs, devtoolsRef]);


...

return (
    <BridgeContext.Provider value={bridge} ref={devtoolsRef}>
```

Also, maybe we can rewrite the switch to:

```js
if (shouldSwitchTabs) {
  const tabIndex = ["1", "2"].indexOf(event.key);
  if (tabIndex > -1) {
    setTab(tabs[tabIndex].id);
    event.preventDefault();
    event.stopPropagation();
  }
}
```

bvaughn:
Refs don't really belong in the dependencies array. They aren't React-managed-state/props. Also `tab` really doesn't need to be in there either, nor `setTab` because it's stable :smile:

I'm not really a fan of unnecessary succinctness in cases like this 😄 I think a good old fashioned switch/case is more readable! 😁

kerolloz:
I think `tab` needs to be in the dependencies array because its value changes whenever the tab switches. I tried removing it, and it didn't work, because the value of tab gets changed but it doesn't inside `useEffect`.

strdr4605:
> I'm not really a fan of unnecessary succinctness in cases like this 😄 I think a good old fashioned switch/case is more readable! 😁

Then it's better to extract 
```js
setTab(tabs[0].id);
event.preventDefault();
event.stopPropagation();
```
in another function as it is repetion of code.

----

But I write `switch` statements in 2 cases :sweat_smile::
1. No other way is possible (like `objectWithCases[caseToMatch]`)
2. Pattern matching in ReasonML

kerolloz:
Thank you @strdr4605 
but `if (!devtoolsRef.current) return;` will not help bind the event

bvaughn:
> I think `tab` needs to be in the dependencies array because its value changes whenever the tab switches.

The current tab does not need to be referenced at all in the effect. If you call `setTab` with the same as the current value, React will just ignore it. That's why the example code I wrote didn't use it.

bvaughn:
> Then it's better to extract
> ...
> in another function as it is repetion of code.

No. Repetition is fine in small amounts. The code is not large or complicated. We don't need to write it in as few characters as possible.

kerolloz:
Amazing react ❤️

So now the only problem I am facing is using `devtoolsRef`

packages/react-devtools-shared/src/devtools/views/DevTools.js
```diff
@@ -127,6 +127,35 @@ export default function DevTools({
...
+          case '2':
+            setTab(tabs[1].id);
+            event.preventDefault();
+            event.stopPropagation();
+            break;
+        }
+      }
+    };
+
+    if (showTabBar) ownerWindow.addEventListener('keydown', handleKeyDown);
```
bvaughn: nit: Let's move this check to the start of the effect? No point to define `handleKeyDown` (or return a cleanup function) if we're not going to use it. :smile:

packages/react-devtools-shared/src/devtools/views/DevTools.js
```diff
@@ -127,6 +127,35 @@ export default function DevTools({
     [enabledInspectedElementContextMenu, viewAttributeSourceFunction],
   );
 
+  const devToolsRef = useRef<HTMLElement | null>(null);
+
+  useEffect(() => {
+    const ownerWindow = devToolsRef.current.ownerDocument.defaultView;
```
bvaughn: Flow is failing on CI because `devToolsRef.current` could be null. It's looking for a check like:
```js
const devTools = devToolsRef.current;
if (devTools !== null) {
  // ...
}
```

kerolloz:
@bvaughn 
I don't get why some tests has failed

packages/react-devtools-shared/src/devtools/views/DevTools.js
```diff
@@ -127,6 +127,33 @@ export default function DevTools({
...
+              break;
+            case '2':
+              setTab(tabs[1].id);
+              event.preventDefault();
+              event.stopPropagation();
+              break;
+          }
+        }
+      });
+    }
```
bvaughn: Looks like you removed the cleanup function by accident (where you remove the event listener).

kerolloz: > nit: Let's move this check to the start of the effect? No point to define `handleKeyDown` (or return a cleanup function) if we're not going to use it. smile

Sorry I don't understand. I thought we don't need the return statement.
Should I return an empty function?
```js 
return ()=>{};
```

bvaughn: Before, you were always returning a "cleanup function" (removing an event listener) even if you weren't attaching it. I suggested you move the bailout check (if `!showTabBar`) earlier to avoid doing any of that work.

But in the case that you are adding an event listener, you should return a cleanup function that removes it.

kerolloz: Oh my bad. I hope now everything is fine.

packages/react-devtools-shared/src/devtools/views/DevTools.js
```diff
@@ -127,6 +127,36 @@ export default function DevTools({
...
+            case '2':
+              setTab(tabs[1].id);
+              event.preventDefault();
+              event.stopPropagation();
+              break;
+          }
+        }
+      });
+      return () => {
+        ownerWindow.removeEventListener('keydown');
```
bvaughn: This won't actually work. `removeEventListener` needs to be passed the type *and* the handler function. You'll want to declare the handler as a separate variable, then pass it to both `addEventListener` and `removeEventListener`.

kerolloz: what about?
image

kerolloz:
Oh finally! :tada: 
Thank you @bvaughn for your patience. Such a welcoming and helpful community. :sparkles:
I learned a lot!

PR #18255, opened 2020-03-09
feat(devtools): Allow editing context value in Provider

eps1lon:
## Summary

Allow editing of context values for modern context types.

## Test Plan

Verified locally using packages/react-devtools-shell.

master:
devools-context-master

pr:
devools-context-pr

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
@@ -33,6 +33,7 @@ import {
   ElementTypeFunction,
   ElementTypeMemo,
   ElementTypeSuspense,
+  ElementTypeContext,
 } from 'react-devtools-shared/src/types';
 
 import styles from './SelectedElement.css';
@@ -362,6 +363,13 @@ function InspectedElementView({
         });
       }
     };
+  } else if (type === ElementTypeContext) {
+    overridePropsFn = (path: Array<string | number>, value: any) => {
+      const rendererID = store.getRendererIDForElement(id);
+      if (rendererID !== null) {
+        bridge.send('overrideProps', {id, path, rendererID, value});
+      }
+    };
   }
 
   return (
@@ -375,7 +383,9 @@ function InspectedElementView({
           overrideValueFn={overridePropsFn}
           pathRoot="props"
           showWhenEmpty={true}
-          canAddEntries={typeof overridePropsFn === 'function'}
+          canAddEntries={
+            typeof overridePropsFn === 'function' && type !== ElementTypeContext
+          }
         />
         {type === ElementTypeSuspense ? (
           <InspectedElementTree
--- a/packages/react-devtools-shell/src/app/InspectableElements/Contexts.js
+++ b/packages/react-devtools-shell/src/app/InspectableElements/Contexts.js
@@ -104,7 +104,9 @@ export default function Contexts() {
         <LegacyContextConsumer />
       </LegacyContextProvider>
       <ModernContext.Provider value={contextData}>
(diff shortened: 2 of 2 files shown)


packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
```diff
@@ -362,6 +363,13 @@ function InspectedElementView({
         });
       }
     };
+  } else if (type === ElementTypeContext) {
```
eps1lon: It's not clear if this should be added to the previous branch that is entered for function/memo/forwardRef components or if this would be the wrong abstraction.

eps1lon: I think this was the right call: https://github.com/facebook/react/pull/18257/files#diff-fc6493b7eb53b9380e69716f370968a7R367

gaearon:
Can we add some tests for this?

packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
gaearon, line 366: Is this type just Provider or Consumer too?

gaearon: Not sure. That one is because it was a separate concept for legacy. But this one is literally editing props.

eps1lon:
> Can we add some tests for this?

I'd love to but it wasn't clear to me where and how. In previous efforts to work on devtools no tests for the frontend were added: https://github.com/facebook/react/pull/18070/files

packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
eps1lon: Pretty sure this was vor Provider and Consumer. It's working under the assumption the Context.Consumer won't have editable props to begin with. But even if it had any I don't see an issue making them editable?

eps1lon:
bump

bvaughn:
@rachelnabors brought this to my attention, and I will be happy to review it– after #19774 lands. (I can handle the conflicts.)

eps1lon:
> (I can handle the conflicts in this PR myself if necessary.)

Does it make sense to rebase this PR on your branch at this point? Otherwise I can take care of tending to this PR after #19774 lands if I notice merge conflicts.

bvaughn:
Up to you. I'll likely rebase my branch at least once more after I resolve the last outstanding issue or two. After that, it's up to you if you feel up for handling the merge conflicts or handing them off to me. :smile:

bvaughn:
#19774 was just merged. Would you like to fix the conflict in `SelectedElement` @eps1lon? Or would you like me to? (Happy either way.)

eps1lon:
I'll take a look.

eps1lon:
Looks like the code has completely changed. Would take me a bit longer (over the next week) so if you want to get this in earlier it's probably best if you rebase this PR.

bvaughn:
No worries. I can take a look this afternoon.

bvaughn:
Ah, so to clarify: Editing the context value *of a context Provider* works in master now already.

Editing the context value *of a Consumer* works in some cases, but not all:
* You **can** edit class consumers that use the legacy `contextTypes` API.
* You **can** edit class consumers that use the more recent `contextType` API.
* You **can not** edit modern `<Context.Consumer>` values. (This would require changes on the backend to support– we'd want to follow the pattern used for editing function component props and inject a new renderer method, and I don't think that's worth it).

eps1lon:
> Editing the context value of a modern context Provider works in master now already.

Awesome, thanks!

PR #18257, opened 2020-03-09
 feat(devtools): Allow editing of context value in Consumer

eps1lon:
Review on per-commit basis advised

## Summary

Allow editing of the context value from a consumer if the value comes from a Provider component.

## Test Plan

Verified locally by using react-devtools-shell:

`master`:
devools-context-consumer-master

`pr`:
devools-context-consumer-pr

### DIFF
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -2722,14 +2722,56 @@ export function attach(
 
     const fiber = findCurrentFiberUsingSlowPathById(id);
     if (fiber !== null) {
-      const instance = fiber.stateNode;
-      if (path.length === 0) {
-        // Simple context value
-        instance.context = value;
+      const typeSymbol = getTypeSymbol(fiber.type);
+      if (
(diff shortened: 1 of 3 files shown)


eps1lon:
> I think maybe it should offer a "link" to jump to the appropriate Provider.

Yep, I was thinking this as well. Happy to abandon this one in favor of a "jump to provider" feature combined with #18255

> Although at that point you might as well just edit the appropriate state?

You mean like

```jsx
function App() {
  const [state] = React.useState({ color: 2 }); // <--- edit this hook instead?
  return <Context.Provider value={value} />
}
```
It's not always obvious where the value of a provider comes from. Could be a memo or useState or useReducer hook. I think editing the value directly is more straight forward.

eps1lon:
But I do feel like that being able to edit context in the context of the component is still important (useContext) here. Otherwise it's not easy to jump between the component props and the context for that particular component.

eps1lon:
bump

eps1lon:
Abandoning. Context.Consumer isn't that important. `useContext` would be more helpful but is more complicate last I looked.

PR #18260, opened 2020-03-10
[Flight] Split Streaming from Relay Implemenation

sebmarkbage:
This splits the Flight implementation into a general JSON chunk part and a streaming part. The default implementation goes to a binary/string stream API which varies between Node and W3C streams. The Relay implementation by-passes the serialization and instead gets passed JSON-serializable object chunks.

The server configs are layered like this:

- ReactFlightServer
  - ReactFlightServerConfig
    - ReactFlightServerConfigStream
      - ReactServerStreamConfig
        - ReactServerStreamConfigNode
        - ReactServerStreamConfigBrowser
    - ReactFlightDOMRelayServerHostConfig

The client is a bit simpler. The implementation just imports ReactFlightClient (used by Relay) or ReactFlightClientStream (used by Streams).

## Flight Relay Server Bindings

The API for the server is: 

```
function render(model: ReactModel, destination: Destination): void;
```

The first argument is something that can be returned from a Block. I.e. the root model. Something JSONable. This will change to require a Block. The second argument is whatever Relay wants to the target to be.

This also depends on an external file that we'll have to add to www: `ReactFlightDOMRelayServerIntegration.js`. This will needs to implement the following API:

```
 opaque type Destination;
 function emitModel(
    destination: Destination,
    id: number,
    json: JSONValue,
  ): void;
 function emitError(
    destination: Destination,
    id: number,
    message: string,
    stack: string,
  ): void;
function close(destination: Destination): void;
```

This will be called asynchronously as ReactFlight completes part of the model. Once there's no more to render, it'll call `close()` on the destination. Relay can then serialize these rows as it wants.

## Flight Relay Client Bindings

The client API is:

```
function createResponse(): Response;
function getModelRoot<T>(response: Response): {model: T};
function resolveModel(
  response: Response,
  id: number,
  json: JSONValue,
): void;
function resolveError(
  response: Response,
  id: number,
  message: string,
  stack: string,
): void;
function close(response: Response): void;
```

To use this, you first have to create a "response". You can get a model out of the response which will just have placeholders in it. This particular API will change as I get rid of "proxies". It’ll return a Block instead. The principle is that it can return something even before any data has loaded so you immediately get something that can suspend.

The resolveModel and resolveError are called as new rows stream in. `close` should be called when there's no more rows or the connection dies (even if it dies early).

### DIFF
--- a/packages/react-flight-dom-relay/src/ReactFlightDOMRelayClient.js
+++ b/packages/react-flight-dom-relay/src/ReactFlightDOMRelayClient.js
@@ -7,9 +7,7 @@
  * @flow
  */
 
-import type {ReactModelRoot} from 'react-client/src/ReactFlightClient';
-
-import type {Chunk} from './ReactFlightDOMRelayClientHostConfig';
+import type {Response, JSONValue} from 'react-client/src/ReactFlightClient';
 
 import {
   createResponse,
@@ -19,8 +17,6 @@ import {
   resolveErrorChunk,
   close,
 } from 'react-client/src/ReactFlightClient';
-
-type EncodedData = Array<Chunk>;
 
 function parseModel(response, targetObj, key, value) {
   if (typeof value === 'object' && value !== null) {
@@ -42,27 +38,17 @@ function parseModel(response, targetObj, key, value) {
   return parseModelFromJSON(response, targetObj, key, value);
 }
 
-function read<T>(data: EncodedData): ReactModelRoot<T> {
-  let response = createResponse();
-  for (let i = 0; i < data.length; i++) {
-    let chunk = data[i];
-    if (chunk.type === 'json') {
-      resolveModelChunk(
-        response,
-        chunk.id,
-        parseModel(response, {}, '', chunk.json),
-      );
-    } else {
-      resolveErrorChunk(
-        response,
-        chunk.id,
-        chunk.json.message,
-        chunk.json.stack,
-      );
-    }
-  }
-  close(response);
-  return getModelRoot(response);
+export {createResponse, getModelRoot, close};
+
+export function resolveModel(response: Response, id: number, json: JSONValue) {
+  resolveModelChunk(response, id, parseModel(response, {}, '', json));
 }
 
-export {read};
+export function resolveError(
+  response: Response,
+  id: number,
+  message: string,
+  stack: string,
+) {
+  resolveErrorChunk(response, id, message, stack);
+}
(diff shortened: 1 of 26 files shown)


packages/react-flight-dom-relay/src/ReactFlightDOMRelayClient.js
```diff
@@ -7,24 +7,48 @@
...
+  close,
 } from 'react-client/src/ReactFlightClient';
 
-type EncodedData = Array<string>;
-
-function read<T>(data: EncodedData): ReactModelRoot<T> {
-  let response = createResponse(data);
-  for (let i = 0; i < data.length; i++) {
-    processStringChunk(response, data[i], 0);
+function parseModel(response, targetObj, key, value) {
```
josephsavona: `key` seems unused?

josephsavona:
> The API for the server is: 
> `function render(model: ReactModel, destination: Destination): void;`
> The first argument is something that can be returned from a Block. I.e. the root model

To clarify, this is the result of calling a Block's load function, right? So it's the curried Block Query function plus its arguments.

sebmarkbage:
> To clarify, this is the result of calling a Block's load function, right? So it's the curried Block Query function plus its arguments.

Yea, that's correct. We still need a name for that. 😂

josephsavona:
> We still need a name for that.

Thunk?

josephsavona:
Sorry for multiple rounds of questions, but who calls the client methods like `resolveModel`? Is that all orchestrated by Flight?

packages/react-flight-dom-relay/src/ReactFlightDOMRelayClient.js
```diff
@@ -7,24 +7,48 @@
...
+          value,
+          innerKey,
+          value[innerKey],
+        );
+      }
+    }
   }
-  complete(response);
-  return getModelRoot(response);
+  return parseModelFromJSON(response, targetObj, key, value);
```
sebmarkbage: "key" is used here.

josephsavona: 🤦‍♂

sebmarkbage:
@josephsavona No. Relay calls resolveModel and resolveError when it receives that "chunk" from the server. So you encode model and error however you want on the server and then parse it on the client and feed it back into Flight.

josephsavona:
Ah got it, so inverse of the server. Cool cool.

PR #18262, opened 2020-03-10
RFC Implementation: Speculative work

gnoff:
## Summary

This is a long-lived PR to discuss development of an implementation for https://github.com/reactjs/rfcs/pull/150

The current state is a Proof-of-Concept implementation of speculative work along with context-selectors.

There are currently four experiments embedded in this PR. some will be removed soon

## selectors for useContext

flag `enableContextSelectors`

this replaces `readContext` as the implementation for `useContext` and replaces with a stateful mount and update hook which can track previous values. It is not inherently 'faster' by itself since it does not affect context propagation and therefore the selector has no opportunity to bail out of work but when used with `reifyNextWork` or `speculativeWork` it can help reduce the amount of work done on a render for components that only read a small part of the context value

what's next:
I plan on reimplementing outside of hooks so any context reader can use a selector and take advantage of more aggressive bailouts

## Context propagation by Reader

flag `enableContextReaderPropagation`

this feature replaces the tree-walking context propagation for one that tracks direct readers. It is analagous to each reader being a `useReducer` hook and having the `dispatchAction` function held by the Provider. propagation is then mostly just dispatching a new state. It's not literally using this functionality and there are some tricks to make it faster on mount and only do the more complicated tracking in advanced cases when the set of contexts read from changes from render to render.

The key benefit here is the growth in work time for propagation is O(readers) instead of O(size of tree)

what's next:
I plan on moving where we track whether we need advanced mode into the prepare step to solve some uncommon memory leaks
I plan on experimenting with a super fast mode where we only activate reader tracking after a provider has a changed value at least once. this would enable passive contexts to add no mount or unmount costs

## Reify Next Work

flag `enableReifyNextWork`

This is the primary feature of Speculative Work. When this flag is enabled the bailoutOnAlreadyCompleted work will be modified so that before deciding whether to go deeper into the subtree the boundary of the next fibers that have work will be visited with a walking algorithm similar to context propagation. when a fiber with work is found and if that fiber type is supported (FunctionComponent for instance) a semi-eager bailout can run which will determine if that speculative update is now reified? (i.e. an update went from "this might need to an update" to "there is no update" or "there is definitely" an update)

If reification happens we've hit a work boundary and move on. once we exhaust the sub-tree up to the next work boudary and return back to our bailout we see if there is still any work to do deeper. the rest of the bailout logic remains unchanged

this is superior to speculative mode because the tree walking algo is lighter since we aren't doing the full 'work' logic, and because it is simpler in that we don't have to juggle "are we doing work on current or on work in progress"

what's next:
explore doing a full render call during reifyNextWork and stashing the result. if work is needed we can use the precomputed workInProgress instead of doing it a second time. this is actually 'safe' because we no none of the intermediate work could have affected the render
Also look into how we might support class components

## DEPRECATED Speculative Work Mode

flag `enableSpeculativeWork`

This was my initial implementation of speculative work and it was good but `reifyNextWork` is better. Essentially if we did a bailoutOnAlreadyCompleted work we would not create workInProgress fibers for children but would instead start work on the current tree. Then if we got somewhere where an update on curren required workInProgress we would reify the ancestor current tree back to the last WIP fiber before we started speculative mode and continue from there.

This worked and did offer some performance enhancements in some cases but there are a lot of extra edge cases tracking whether you are in speculative mode or not and how to handle switching between those modes

This feature is technically compatible with reifyNextWork but doesn't really offer any additional performance advantages and so it should probably be abandoned

What's next:
remove from codebase


packages/react-reconciler/src/ReactFiber.js
```diff
@@ -277,6 +280,8 @@ function FiberNode(
   this.sibling = null;
   this.index = 0;
 
+  this.residue = null;
```
gnoff: based on discussion with @sebmarkbage will remove residue and replace with a childFiber-like comparison algorithm that uses identical semantics as the reconciliation step but does not actually reconcile.

This is so we can determine if we are returning identical children from a previous render and can avoid reifying the speculative work tree

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2885,11 +3091,131 @@ function remountFiber(
   }
 }
 
+function reifyWorkInProgress(current: Fiber, workInProgress: Fiber | null) {
```
gnoff: This is the heavy lift, it is basically the child cloning algo but done in reverse walking back up the return list to get to a state we would have been in if speculative work was not a feature

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2885,11 +3091,131 @@ function remountFiber(
...
+      workInProgress.return = null;
+    }
+
+    // visit each child of workInProgress and create a newChild
+    let child = workInProgress.child;
+    let parent = workInProgress;
+    let newChild = voidFiber;
+    while (child !== null) {
+      if (child === reifiedCurrentChild) {
+        newChild = newChild.sibling = parent.child = reifiedNewChild;
```
gnoff: using a technique like voidFiber allows this heavily linearized assignment. it's probably too cute and I have no idea if it actually matters for perf. if we didn't use it we'd have to have an additional conditional to treat wip.child different than child.sibling

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -2885,11 +3091,131 @@ function remountFiber(
...
+        alreadyCompletedSpeculativeChildFibers = false;
+      } else {
+        newChild = newChild.sibling = parent.child = createWorkInProgress(
+          child,
+          child.pendingProps,
+          child.expirationTime,
+        );
+        newChild.return = null;
+        if (alreadyCompletedSpeculativeChildFibers) {
+          resetWorkInProgressFromCompletedSpeculativFiberExpirationTimes(
```
gnoff: with my most recent commit this is probably not required. fibers already completed had their childExpirationTime's reset anyway so this is not likely doing anything anymore

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -614,6 +656,111 @@ function updateWorkInProgressHook(): Hook {
...
+  } else {
+    // selector is actually observedBits
+    let observedBits = selector;
+    let contextValue = readContext(context, observedBits);
+    memoizedState.contextValue = contextValue;
+    return contextValue;
+  }
+}
+
+function bailoutContext(hook: Hook): boolean {
```
gnoff: these bailout functions will be run outside the context of the function component execution itself so they have to be given the hook in consideration

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -837,6 +985,44 @@ function rerenderReducer<S, I, A>(
   return [newState, dispatch];
 }
 
+// @TODO this is a completely broken bailout implementation. figure out how
+// dispatchAction and updateReducer really work and implement something proper
+function bailoutReducer(hook, renderExpirationTime): boolean {
```
gnoff: going to take a pass as actually supporting proper expiration time management next here and elsewhere. Would love some advice on writing tests in react to fully exercise various concurrent edge cases so I can verify correctness

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -875,6 +1062,9 @@ function rerenderState<S>(
 }
 
 function pushEffect(tag, create, destroy, deps) {
+  if (enableSpeculativeWork && tag & HookHasEffect) {
+    markWorkInProgressPushedEffect();
```
gnoff: this is used by a certain kind of bailout where even if we called the function component we can still bailout if no observable effects were pushed

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1529,7 +1554,9 @@ function completeUnitOfWork(unitOfWork: Fiber): Fiber | null {
       if (
         returnFiber !== null &&
         // Do not append effects to parents if a sibling failed to complete
-        (returnFiber.effectTag & Incomplete) === NoEffect
+        (returnFiber.effectTag & Incomplete) === NoEffect &&
+        // Do not append effects to parent if workInProgress was speculative
+        workWasSpeculative !== true
```
gnoff: need to add feature flag gate here

packages/react/src/ReactElement.js
```diff
@@ -158,6 +159,12 @@ const ReactElement = function(type, key, ref, self, source, owner, props) {
     _owner: owner,
   };
 
+  if (enableSpeculativeWork) {
+    // allows referential equality checking without having to hang onto old
+    // elements
+    element.residue = {};
```
gnoff: as mentioned elsewhere, will get rid of residue and replace with a childFiber like comparison implementation

yisar:
There is no doubt that it's interesting. It's a good way to provide reference value through selector and update accurately in a lazy way.

But one thing, I think maybe this is not only applicable to Context, maybe we can also make props cooperate with selector? Or with a similar idea, built-in React.memo, so that components can also render accurately?

packages/react-reconciler/src/ReactFiberHooks.js
gaearon: Just to set expectations, @acdlite is planning a pretty invasive set of changes to expiration times

gnoff: Thanks @gaearon, I talked with Sebastian on Monday and he let me know the same and that this PR won't get a lot of attention for the near future given your more immediate plans. That's all good. @acdlite if you are willing to catch me up in the future when you know what it might look like I'd love to hear about it

gnoff:
@yisar at the moment selectors for context only work in function components using the useContext hook. users would be free to incorporate props into the selector without limitation much like they can with reducers in useReducer

PR #18267, opened 2020-03-10
ReactDOM.useEvent add flag and entry point

trueadm:
This PR adds the `enableUseEventAPI` flag and the experimental `ReactDOM.useEvent` entry point (but is not exposed in any of our bundles). This PR is just very early scaffolding and doesn't do much more, follow up PRs will add more parts of the `useEvent` API.

Please note: This API is intended for internal FB testing only.


trueadm:
@threepointone There's nothing to really test right now. The code in this PR is not used anywhere just yet. In a couple of PRs there will be something that can be tested. :)

threepointone:
Yeah haha I figured. :shipit:

PR #18270, opened 2020-03-10
Flare: Fix listener upgrade bug

trueadm:
This fixes a bug that was found internally when attempting to upgrade to an active event listener. We weren't properly removing the passive event listener. Also wired up logic for fb event listeners so we correctly remove those too.

### DIFF
--- a/packages/react-dom/src/events/ReactDOMEventListener.js
+++ b/packages/react-dom/src/events/ReactDOMEventListener.js
@@ -105,6 +105,7 @@ export function addResponderEventSystemEvent(
   } else {
     eventFlags |= IS_ACTIVE;
   }
+  let fbListener;
   // Check if interactive and wrap in discreteUpdates
   const listener = dispatchEvent.bind(
     null,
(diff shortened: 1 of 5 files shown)


packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -113,30 +114,34 @@ export function addResponderEventSystemEvent(
...
       topLevelType,
       listener,
       passive,
     );
   } else {
-    addEventCaptureListener(document, topLevelType, listener);
+    fbListener = addEventCaptureListener(document, topLevelType, listener);
   }
-  return listener;
+  return fbListener || listener;
```
threepointone: I'll be honest, this code is a bit confusing. I can trace it out and it looked ok, but I defer to your judgement here.

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -203,6 +192,29 @@ export function trapEventForPluginEventSystem(
...
+  // EventListener-www module in FB builds.
+  return fbListener || listener;
+}
+
+export function removeTrappedPassiveEventListener(
+  document: Document,
+  topLevelType: string,
+  listener: any => void,
+) {
+  if (listener.remove != null) {
```
sebmarkbage: This is not great because we're shipping this code even if it's not an FB listener. Like if Function.prototype.remove is added on the page. It also confuses the compiler. We should be sure which path we're using and consistently call that - statically

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -203,6 +192,29 @@ export function trapEventForPluginEventSystem(
...
+    listener.remove();
+  } else {
+    if (passiveBrowserEventsSupported) {
+      document.removeEventListener(topLevelType, listener, {
+        capture: true,
+        passive: true,
+      });
+    } else {
+      document.removeEventListener(topLevelType, listener, true);
```
sebmarkbage: This isn't going through the forked indirection which isn't the right layering. This whole function should all be moved to EventListener.js. That way it can be properly forked and you can statically choose which function to call.

PR #18271, opened 2020-03-10
ReactDOM.useEvent: Add more scaffolding for useEvent hook

trueadm:
This PR is a follow up to https://github.com/facebook/react/pull/18267. Specifically, this adds a bit more of the `useEvent` hook implementation (but most just leaves most the code as TODOs). Furthermore, there's a very simple test that validates that the hook works in terms of returning the same event listener map.

### DIFF
--- a/packages/react-debug-tools/src/ReactDebugHooks.js
+++ b/packages/react-debug-tools/src/ReactDebugHooks.js
@@ -39,6 +39,11 @@ type HookLogEntry = {
   ...
 };
 
+type ReactDebugListenerMap = {|
+  clear: () => void,
+  setListener: (instance: EventTarget, callback: ?(Event) => void) => void,
+|};
+
 let hookLog: Array<HookLogEntry> = [];
 
 // Primitives
@@ -256,6 +261,16 @@ function useTransition(
   return [callback => {}, false];
 }
 
+const noOp = () => {};
+
+function useEvent(event: any): ReactDebugListenerMap {
+  hookLog.push({primitive: 'Event', stackError: new Error(), value: event});
+  return {
+    clear: noOp,
+    setListener: noOp,
+  };
+}
+
 function useDeferredValue<T>(value: T, config: TimeoutConfig | null | void): T {
   // useDeferredValue() composes multiple hooks internally.
   // Advance the current hook index the same number of times
@@ -285,6 +300,7 @@ const Dispatcher: DispatcherType = {
   useResponder,
   useTransition,
   useDeferredValue,
+  useEvent,
 };
 
 // Inspect
--- a/packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
+++ b/packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
@@ -1039,4 +1039,40 @@ describe('DOMModernPluginEventSystem', () => {
     expect(log).toEqual([]);
     expect(onDivClick).toHaveBeenCalledTimes(0);
   });
+
+  describe('ReactDOM.useEvent', () => {
+    beforeEach(() => {
+      jest.resetModules();
+      ReactFeatureFlags = require('shared/ReactFeatureFlags');
+      ReactFeatureFlags.enableModernEventSystem = true;
+      ReactFeatureFlags.enableUseEventAPI = true;
+
+      React = require('react');
+      ReactDOM = require('react-dom');
+      Scheduler = require('scheduler');
+      ReactDOMServer = require('react-dom/server');
+    });
+
+    if (!__EXPERIMENTAL__) {
+      it("empty test so Jest doesn't complain", () => {});
+      return;
+    }
+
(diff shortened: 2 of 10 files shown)


packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -1039,4 +1039,35 @@ describe('DOMModernPluginEventSystem', () => {
...
+
+        listenerMaps.push(listenerMap);
+
+        return <div />;
+      }
+
+      ReactDOM.render(<Test />, container);
+      ReactDOM.render(<Test />, container);
+      expect(listenerMaps.length).toEqual(2);
+      expect(listenerMaps[1]).toEqual(listenerMaps[1]);
```
bvaughn:
```suggestion
      expect(listenerMaps[0]).toEqual(listenerMaps[1]);
```

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -256,6 +256,17 @@ function useTransition(
...
 
+const noOp = () => {};
+
+function useEvent(event: any): any {
+  hookLog.push({primitive: 'Event', stackError: new Error(), value: event});
+  return {
+    clear: noOp,
+    listen: noOp,
+    unlisten: noOp,
+  };
```
bvaughn: `useEvent` in `ReactFiberHooks` and `ReactPartialRendererHooks` returns an object with `clear` and `setListener` attributes. This one has `clear`, `listen`, and `unlisten`. Why the difference?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1369,6 +1377,73 @@ function dispatchAction<S, A>(
...
+      }
+      if (memoizedEvent.passive !== event.passive) {
+        console.warn(
+          'The "passive" option passed to the useEvent() hook was different between renders.' +
+            ' The "passive" option is static and should never change between renders.',
+        );
+      }
+    }
+    // In order to clear up upon the hook unmounting,
+    /// we ensure we push an effect that handles the use-case.
```
bvaughn: > In order to clear up upon the hook unmounting,

What does this mean? Not sure I understand.

nit
```suggestion
    // we ensure we push an effect that handles the use-case.
```

packages/react-debug-tools/src/ReactDebugHooks.js
trueadm: That's a mistake, my bad!

packages/react-reconciler/src/ReactFiberHooks.js
trueadm: I updated the comment. Hopefully it makes more sense now!

packages/react-debug-tools/src/ReactDebugHooks.js
bvaughn: I would have expected Flow to catch this. It's supposed to ensure that all of our hooks implementations are consistent. Why is the return type of this hook `any`?

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -474,6 +474,13 @@ function useTransition(
   return [startTransition, false];
 }
 
+function useEvent(event: any): any {
```
bvaughn: Same question for this. Why is the return type `any` rather than `ReactListenerMap`?

ecreeth:
@trueadm you have added **react-dom-16.13.0.tgz**. it's a mistake?

trueadm:
@ecreeth Nicly spotted. Thanks :)

packages/react-native-renderer/src/ReactNativeHostConfig.js
```diff
@@ -26,6 +26,10 @@ import ReactNativeFiberHostComponent from './ReactNativeFiberHostComponent';
 
 const {get: getViewConfigForType} = ReactNativeViewConfigRegistry;
 
+export type ReactListenerEvent = Object;
+export type ReactListenerMap = Object;
+export type ReactListener = Object;
```
bvaughn: I wonder if you considered using a shared type between the reconciler and non-reconciler based renderers?

Do we expect that the type of `ReactListenerMap` and `ReactListener` will vary between renderers? I guess they would if `ReactListenerEvent` varied... maybe it would be hard to manage this with Flow generics.

I'm just a little concerned that Flow doesn't have our back here to keep all of the renderers in sync.

trueadm: I considered it, but I think that will largely become something to consider once we actually add these to those renderers. Given this is only exported (for now anyway) on `ReactDOM`, I don't see this as a blocker. In the long-term, we'd probably expose slightly different API. Like you mention, doing this with generics can become a bit of a pain, so I wanted to keep it as `Object` for now, just to make this work-load easier to get into the internal testing phase.

PR #18272, opened 2020-03-10
Don't warn about unmounted state updates from a passive destroy function

bvaughn:
Suppress new warning introduced by PRs #17925 and #17947.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -199,13 +199,14 @@ const {
 
 type ExecutionContext = number;
 
-const NoContext = /*                    */ 0b000000;
-const BatchedContext = /*               */ 0b000001;
-const EventContext = /*                 */ 0b000010;
-const DiscreteEventContext = /*         */ 0b000100;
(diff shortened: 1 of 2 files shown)


acdlite:
Why?

bvaughn:
To be clear, I don't think this is the right change to make- but I've been trying (without much success) to get feedback from others on the team about this for a couple of days. Just posting this PR in case we end up deciding that the warning should be reverted for now.

acdlite:
Yeah my instinct is that it's not right, either. I think this would most likely happen if you dispatch a Flux action or something inside the unmount function, which then accidentally fires a `setState` on the component that is unmounting. Which isn't a great pattern, but even still, the fix would be to unsubscribe from the Flux store before dispatching the action.

Curious to hear from others if there are arguments in favor of suppressing the warning.

gaearon:
Should we be warning on getState in cleanup in general? Seems a bit annoying to fix when you only "see" the unmount one but the fix involves fixing both anyway.

bvaughn:
Yeah, I mentioned this in my internal thread. I actually think we should, because I think there aren't really any good cases for updating state in a cleanup function.

bvaughn:
Closing in favor of #18307.

PR #18274, opened 2020-03-11
ModernEventSystem: refactor accumulateTwoPhaseListeners

trueadm:
This PR refactors, optimizes and simplifies the logic that we currently have to accumulate two phase event listeners. Previously, we have `accumulateTwoPhaseDispatchesSingle` and `accumulateTwoPhaseDispatches`. The implentation in this PR works by using the notoriously slow `unshift` array operation

I often found it difficult to follow how they worked and also notice how they could be further optimized. I've commented the implementation so it should be easier to follow. I also ensured the `SyntheticEvent` object has the fields initialized, to avoid hidden class de-opts, which also allowed me to properly fix the Flow definitions and improve Flow coverage. Once we remove the `Responder` plugin, we can delete a whole lot of legacy code going forward.

Furthermore, given the on-going work for the `ReactDOM.useEvent`, it occured to me that we could use a single function that accumulated listeners and later, extend this function to work with `useEvent` listeners too.

## Performance

Some might be wondering why I'm using `unshift` vs alternative approaches that avoid using it. I benchmarked several different versions and the realism is that we never really build up that many listeners in real world apps (not enough to really stress `unshift` as being the bottleneck). Here are some perf measurements:

- Previous Implementation (before this PR): **0.030**ms - **0.355**ms
- This PR: **0.009**ms - **0.116**ms
- Alternative approach (not using unshift): **0.021**ms - **0.133**ms

### DIFF
--- a/packages/react-dom/src/events/BeforeInputEventPlugin.js
+++ b/packages/react-dom/src/events/BeforeInputEventPlugin.js
@@ -7,7 +7,6 @@
 
 import type {TopLevelType} from 'legacy-events/TopLevelEventTypes';
 
-import {accumulateTwoPhaseDispatchesSingle} from 'legacy-events/EventPropagators';
 import {canUseDOM} from 'shared/ExecutionEnvironment';
 
 import {
@@ -29,6 +28,7 @@ import {
 } from './FallbackCompositionState';
 import SyntheticCompositionEvent from './SyntheticCompositionEvent';
 import SyntheticInputEvent from './SyntheticInputEvent';
+import {accumulateTwoPhaseListeners} from './DOMModernPluginEventSystem';
 
 const END_KEYCODES = [9, 13, 27, 32]; // Tab, Return, Esc, Space
 const START_KEYCODE = 229;
@@ -276,7 +276,7 @@ function extractCompositionEvent(
     }
   }
 
-  accumulateTwoPhaseDispatchesSingle(event);
+  accumulateTwoPhaseListeners(event);
   return event;
 }
 
@@ -437,7 +437,7 @@ function extractBeforeInputEvent(
   );
 
   event.data = chars;
-  accumulateTwoPhaseDispatchesSingle(event);
+  accumulateTwoPhaseListeners(event);
   return event;
 }
 
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -19,8 +19,9 @@ import {registrationNameDependencies} from 'legacy-events/EventPluginRegistry';
 import {batchedEventUpdates} from 'legacy-events/ReactGenericBatching';
 import {executeDispatchesInOrder} from 'legacy-events/EventPluginUtils';
 import {plugins} from 'legacy-events/EventPluginRegistry';
+import getListener from 'legacy-events/getListener';
 
-import {HostRoot, HostPortal} from 'shared/ReactWorkTags';
+import {HostRoot, HostPortal, HostComponent} from 'shared/ReactWorkTags';
 
 import {addTrappedEventListener} from './ReactDOMEventListener';
 import getEventTarget from './getEventTarget';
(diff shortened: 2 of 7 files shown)


packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -305,3 +306,39 @@ export function attachElementListener(listener: ReactDOMListener): void {
...
+    return;
+  }
+  const dispatchListeners = (event._dispatchListeners = []);
+  const dispatchInstances = (event._dispatchInstances = []);
+  const {bubbled, captured} = phasedRegistrationNames;
+  let node = event._targetInst;
+
+  // Accumulate all instances and listeners via the target -> root path.
+  // In order to reduce allocations, we store the listeners and instances
+  // in pairs. i.e. [func, <div>, func2, <span>, ...]
```
philipp-spiess: Is the comment outdated? Seems like you're using two arrays now instead of one with pairs.

trueadm: Yeah, I fixed :)

trueadm:
@philipp-spiess I really want to stop attaching the events, but that's a breaking change and is likely to affect RNW and the Responder plugin. We also need to stop exporting the event system accumulation events. Once we do that, we should save a bunch of bytes.

bvaughn:
Screen Shot 2020-03-11 at 10 18 21 AM🤔

:think

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -305,3 +306,46 @@ export function attachElementListener(listener: ReactDOMListener): void {
...
+        hasListeners = true;
+        // Capture listeners/instances should go at the start, so we
+        // unshift them to the start of the array.
+        dispatchListeners.unshift(captureListener);
+        dispatchInstances.unshift(node);
+      }
+      const bubbleListener = getListener(node, bubbled);
+      if (bubbleListener != null) {
+        hasListeners = true;
+        // Bubble listeners/instances should go at the start, so we
```
bvaughn:
```suggestion
        // Bubble listeners/instances should go at the end, so we
```

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -305,3 +306,46 @@ export function attachElementListener(listener: ReactDOMListener): void {
...
+        // push them to the end of the array.
+        dispatchListeners.push(bubbleListener);
+        dispatchInstances.push(node);
```
bvaughn: This looks odd. You're unshifting so that `node` comes before `listener` but you're pushing in the opposite order. Mistake?

trueadm: I’m not sure I get you. I’m unshifting to insert both the node and listener at the start during capture and pushing both during bubble.

trueadm:
> Screen Shot 2020-03-11 at 10 18 21 AM
> 
> 🤔

Fixed. I copied the bit in brackets from the last point and forgot to remove the copied text.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
bvaughn: I am not very familiar with the surrounding code, but I assume that the sequence of node -> listener -> node -> listener is significant.

But since you do this above:
```js
dispatchListeners.unshift(captureListener);
dispatchInstances.unshift(node);
```

...and this below:
```js
  dispatchListeners.push(bubbleListener);
  dispatchInstances.push(node);
```

...you will end up with a sequence of node -> listener -> ... -> listener -> node. I assume this would break code elsewhere that destructures these tuples?

trueadm: But the nodes and listeners are pushed or unshifted into separate arrays?

bvaughn: Oh...gosh. 😐Yup. I was definitely reading both as the same array, even after copy-pasting. The monospace font and my own blindness got me. Sorry for the confusion.

packages/react-dom/src/events/BeforeInputEventPlugin.js
```diff
@@ -29,6 +28,7 @@ import {
 } from './FallbackCompositionState';
 import SyntheticCompositionEvent from './SyntheticCompositionEvent';
 import SyntheticInputEvent from './SyntheticInputEvent';
+import {accumulateTwoPhaseListeners} from './DOMModernPluginEventSystem';
```
gaearon: Why is shared code unconditionally calling an implementation from the modern event system? If it is standalone, should it have been shared? This seems error-prone.

gaearon: Also can we delete `accumulateTwoPhaseDispatchesSingle` and `traverseTwoPhase` now?

I would find it clearer if the new implementation moved to where they lived before. Given that it’s used in legacy system too now.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -305,3 +306,46 @@ export function attachElementListener(listener: ReactDOMListener): void {
...
+      }
+    }
+    node = node.return;
+  }
+  // To prevent allocation to the event unless we actually
+  // have listeners we use the flag we would have set above.
+  if (hasListeners) {
```
gaearon: Wouldn’t checking the array length give us the same result?

packages/react-dom/src/events/BeforeInputEventPlugin.js
trueadm: We can’t delete them just yet. They’re exported from ReactDOM in the hidden object.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: It would. Either work here. I can change to do that in a follow up.

PR #18278, opened 2020-03-11
Don't use closures in DevTools injection

sebmarkbage:
Nested closures are tricky. They're not super efficient and when they share scope between multiple closures they're hard for a compiler to optimize. It's also unclear how many versions will be created.

By hoisting things out an just make it simple calls the compiler can do a much better job.

## Before 

```js
"function" === typeof ik && ik(c.stateNode, d);
```
```js
"function" === typeof Qi && Qi(b);
```
```js
var ik = null,
  Qi = null;
function nk(a) {
  if ("undefined" === typeof __REACT_DEVTOOLS_GLOBAL_HOOK__) return !1;
  var b = __REACT_DEVTOOLS_GLOBAL_HOOK__;
  if (b.isDisabled || !b.supportsFiber) return !0;
  try {
    var c = b.inject(a);
    ik = function(a) {
      try {
        b.onCommitFiberRoot(c, a, void 0, 64 === (a.current.effectTag & 64));
      } catch (e) {}
    };
    Qi = function(a) {
      try {
        b.onCommitFiberUnmount(c, a);
      } catch (e) {}
    };
  } catch (d) {}
  return !0;
}
(function(a) {
  var b = a.findFiberByHostInstance;
  return nk(
    n({}, a, {
      overrideHookState: null,
      overrideProps: null,
      setSuspenseHandler: null,
      scheduleUpdate: null,
      currentDispatcherRef: Wa.ReactCurrentDispatcher,
      findHostInstanceByFiber: function(a) {
        a = ic(a);
        return null === a ? null : a.stateNode;
      },
      findFiberByHostInstance: function(a) {
        return b ? b(a) : null;
      },
      findHostInstancesForRefresh: null,
      scheduleRefresh: null,
      scheduleRoot: null,
      setRefreshHandler: null,
      getCurrentFiber: null
    })
  );
})({
  findFiberByHostInstance: uc,
  bundleType: 0,
  version: "16.13.0",
  rendererPackageName: "react-dom"
});
```

## After

```js
  if ("function" === typeof ik)
    try {
      ik(Qi, c, void 0, 64 === (c.current.effectTag & 64));
    } catch (ya) {}
```
```js
  if ("function" === typeof Pi)
    try {
      Pi(Qi, b);
    } catch (e) {}
```
```js
var Dk = {
    findFiberByHostInstance: tc,
    bundleType: 0,
    version: "16.13.0",
    rendererPackageName: "react-dom"
  };
var Ek = {
  bundleType: Dk.bundleType,
  version: Dk.version,
  rendererPackageName: Dk.rendererPackageName,
  getInspectorDataForViewTag: Dk.getInspectorDataForViewTag,
  overrideHookState: null,
  overrideProps: null,
  setSuspenseHandler: null,
  scheduleUpdate: null,
  currentDispatcherRef: Ya.ReactCurrentDispatcher,
  findHostInstanceByFiber: function(a) {
    a = hc(a);
    return null === a ? null : a.stateNode;
  },
  findFiberByHostInstance: Dk.findFiberByHostInstance || tk,
  findHostInstancesForRefresh: null,
  scheduleRefresh: null,
  scheduleRoot: null,
  setRefreshHandler: null,
  getCurrentFiber: null
};
if ("undefined" !== typeof __REACT_DEVTOOLS_GLOBAL_HOOK__) {
  var Fk = __REACT_DEVTOOLS_GLOBAL_HOOK__;
  if (!Fk.isDisabled && Fk.supportsFiber)
    try {
      (Qi = Fk.inject(Ek)),
        (ik = Fk.onCommitFiberRoot),
        (Pi = Fk.onCommitFiberUnmount);
    } catch (a) {}
}
```

Note that it's still sub-optimal because the config object gets created even though it should be able to be inlined by Closure Compiler. We should really switch this per-renderer config object to be a HostConfig the whole thing instead to avoid that.

The injected object is also created outside the condition so it's always created even if no devtools is present.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberDevToolsHook.js
+++ b/packages/react-reconciler/src/ReactFiberDevToolsHook.js
@@ -21,6 +21,7 @@ import {DidCapture} from './ReactSideEffectTags';
 declare var __REACT_DEVTOOLS_GLOBAL_HOOK__: Object | void;
 
 let onScheduleFiberRoot = null;
+let rendererID = null;
 let onCommitFiberRoot = null;
 let onCommitFiberUnmount = null;
 let hasLoggedError = false;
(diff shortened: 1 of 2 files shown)


sebmarkbage:
Looks like fast-refresh is mutating the hook object. :(

packages/react-reconciler/src/ReactFiberDevToolsHook.js
```diff
@@ -123,19 +65,64 @@ export function injectInternals(internals: Object): boolean {
 }
 
 export function onScheduleRoot(root: FiberRoot, children: ReactNodeList) {
-  if (typeof onScheduleFiberRoot === 'function') {
-    onScheduleFiberRoot(root, children);
+  if (__DEV__) {
+    if (
+      injectedHook &&
+      typeof injectedHook.onScheduleFiberRoot === 'function'
+    ) {
```
bvaughn: Could sent pointers to e.g. `onScheduleFiberRoot` directly so we wouldn't have to do the property access

```js
if (typeof onScheduleFiberRoot === 'function') {
  onScheduleFiberRoot(rendererID, root, chilren);
}
```

gaearon: I think Seb is saying that property gets mutated by Fast Refresh. (I don't actually remember if it does but I guess?)

gaearon:
So, landing this?

pgn-vole:
Thanks for this change. FYI - It is also fixing a bug with React Fast refresh which happen when `hook.onScheduleFiberRoot` is defined **after** React code is excuted. 
In this setup, Fast Refresh will not work properly because it didn't manage to register a root because React didn't called its callback.

PR #18282, opened 2020-03-11
ReactDOM.useEvent: more scaffolding changes

trueadm:
This PR doesn't really introduce any real changes, other than adding additional scaffolding needed for the `useEvent` API. Notably, there have been quite a few variable renames and Flow type changes, not to mention the only real visible code change to ensure that we're dealing with actual DOM nodes when finding ancestors (more guarding against the case where the target container is not a DOM node, i.e. a future `window` node).

Lastly, this adds `validateEventListenerInstance`, which will have tests added in a follow up when the event system is correctly wired up to the hooks (will happen in the next PR). This function only applies warnings and validation against the wrong signature of `useEvent`'s map object being used (`setListener`).

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -1085,7 +1085,7 @@ export function mountEventListener(listener: ReactDOMListener): void {
   if (enableUseEventAPI) {
     const {target} = listener;
     if (target === window) {
-      // TODO
+      // TODO (useEvent)
     } else {
       attachElementListener(listener);
     }
@@ -1096,14 +1096,14 @@ export function unmountEventListener(listener: ReactDOMListener): void {
   if (enableUseEventAPI) {
     const {target} = listener;
     if (target === window) {
-      // TODO
+      // TODO (useEvent)
     } else {
       detachElementListener(listener);
     }
   }
 }
 
-export function validateEventListenerInstance(
+export function validateEventListenerTarget(
   target: EventTarget,
   listener: ?(Event) => void,
 ): boolean {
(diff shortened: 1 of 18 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1631,6 +1636,14 @@ function mountEventListener(event: ReactListenerEvent): ReactListenerMap {
   if (enableUseEventAPI) {
     const hook = mountWorkInProgressHook();
 
+    // These are TODOs
+    // eslint-disable-next-line no-unused-expressions
+    mountHostEventListener;
+    // eslint-disable-next-line no-unused-expressions
+    unmountHostEventListener;
+    // eslint-disable-next-line no-unused-expressions
+    validateEventListenerInstance;
```
bvaughn: This seems a bit unusual and unnecessary :smile: Slight preference to leave this change out.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1075,3 +1080,63 @@ export function unmountFundamentalComponent(
 export function getInstanceFromNode(node: HTMLElement): null | Object {
   return getClosestInstanceFromNode(node) || null;
 }
+
+export function mountEventListener(listener: ReactDOMListener): void {
+  if (enableUseEventAPI) {
+    const {target} = listener;
+    if (target === window) {
+      // TODO
```
bvaughn: nit IMO it's_ nice to tag TODOs with a little more info to make them project grepable, e.g. `TODO (useEvent)`

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1075,3 +1080,63 @@ export function unmountFundamentalComponent(
...
+    } else {
+      detachElementListener(listener);
+    }
+  }
+}
+
+export function validateEventListenerInstance(
```
bvaughn: Just curious why you've renamed all of the params to `target` but this method calls it "instance"

PR #18285, opened 2020-03-12
Set up infra for react-reconciler fork

acdlite:
We're planning to land some significant refactors of the reconciler. We want to be able to gradually roll out the new implementation side-by-side with the existing one. So we'll create a short-lived fork of the react-reconciler package. Once the new implementation has stabilized, we'll delete the old implementation and promote the new one.

This means, for as long as the fork exists, we'll need to maintain two separate implementations. This sounds painful, but since the forks will still be largely the same, most changes will not require two separate implementations. In practice, you'll implement the change in the old fork and then copy paste it to the new one.

This commit only sets up the build and testing infrastructure. It does not actually fork any modules. I'll do that in subsequent PRs.

The forked version of the reconciler will be used to build a special version of React DOM. I've called this build ReactDOMForked. It's only built for www; there's no open source version.

The new reconciler is disabled by default. It's enabled in the `yarn test-www-variant` command. The reconciler fork isn't really related to the "variant" feature of the www builds, but I'm piggy backing on that concept to avoid having to add yet another testing dimension.

---

First commit is a file rename. Second commit includes the changes to infra.


acdlite:
Github wat

acdlite:
Must a be a weird bug. PR definitely wasn't merged :D

PR #18286, opened 2020-03-12
Wrap Components and Profiler tabs with box-size style too

### LINKED ISSUE #18284: Bug: DevTools Profiler doesn't show selected commit
I don't see the selection at all. My version is 4.5.0. I think this regressed.

Screenshot 2020-03-11 at 23 40 37

bvaughn:
For the browser extension, these views get rendered into portals and so they don't inherit the box-sizing style from the `.DevTools` wrapper element. This causes views like the Profiler commit selector to subtly break.

This regressed with 9cc094a.

Resolves #18284

Screen Shot 2020-03-11 at 6 35 52 PM

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/Components.css
+++ b/packages/react-devtools-shared/src/devtools/views/Components/Components.css
@@ -9,6 +9,11 @@
   font-family: var(--font-family-sans);
 }
 
+.Components, .Components * {
+  box-sizing: border-box;
+  -webkit-font-smoothing: var(--font-smoothing);
+}
+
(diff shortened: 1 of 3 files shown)


packages/react-devtools-shared/src/devtools/views/Components/Components.css
```diff
@@ -9,6 +9,11 @@
   font-family: var(--font-family-sans);
 }
 
+.Components, .Components * {
```
bvaughn: Wildcard selectors aren't great but... it's fine.

PR #18287, opened 2020-03-12
Add KeyboardEvent.code to synthetic event

### LINKED ISSUE #14102: Synthetic KeyboardEvent should support KeyboardEvent.code

bl00mber:
resolve #14102

### DIFF
--- a/packages/react-dom/src/events/SyntheticKeyboardEvent.js
+++ b/packages/react-dom/src/events/SyntheticKeyboardEvent.js
@@ -16,6 +16,12 @@ import getEventModifierState from './getEventModifierState';
  */
 const SyntheticKeyboardEvent = SyntheticUIEvent.extend({
   key: getEventKey,
+  code: function(event) {
+    if (event.type === 'keydown' || event.type === 'keyup') {
+      return event.code;
+    }
+    return 0;
+  },
   location: null,
   ctrlKey: null,
   shiftKey: null,
--- a/packages/react-dom/src/events/SyntheticKeyboardEvent.js
+++ b/packages/react-dom/src/events/SyntheticKeyboardEvent.js
@@ -16,12 +16,7 @@ import getEventModifierState from './getEventModifierState';
  */
 const SyntheticKeyboardEvent = SyntheticUIEvent.extend({
   key: getEventKey,
-  code: function(event) {
-    if (event.type === 'keydown' || event.type === 'keyup') {
-      return event.code;
-    }
-    return 0;
-  },
+  code: null,
   location: null,
   ctrlKey: null,
   shiftKey: null,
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/events/SyntheticKeyboardEvent.js
```diff
@@ -16,6 +16,12 @@ import getEventModifierState from './getEventModifierState';
  */
 const SyntheticKeyboardEvent = SyntheticUIEvent.extend({
   key: getEventKey,
+  code: function(event) {
+    if (event.type === 'keydown' || event.type === 'keyup') {
```
gaearon, line 20: Can we just have `code: null`? Why does it need special handling?

bl00mber: I assumed this structure have something with react internals or `KeyboardEvent` specification because lower on this interface similar blocks are placed. When `keypress` is executed without react, `code` is equal to output of `keydown` and `keyup` events. However, if it executed with react, the output will be `null`.

packages/react-dom/src/events/__tests__/SyntheticKeyboardEvent-test.js
```diff
@@ -451,6 +451,52 @@ describe('SyntheticKeyboardEvent', () => {
...
+        node.dispatchEvent(
+          new KeyboardEvent('keypress', {
+            code: 'KeyQ',
+            bubbles: true,
+            cancelable: true,
+          }),
+        );
+        expect(codeDown).toBe('KeyQ');
+        expect(codeUp).toBe('KeyQ');
+        expect(codePress).toBe(null);
```
gaearon, line 497: What does the spec say it should be?

bl00mber: The spec says it should be a [valid code](https://w3c.github.io/uievents/#event-type-keypress). For some reason onKeyPress callback do not called when dispatched without charCode! I added charCode and codePress initializes now.

PR #18289, opened 2020-03-12
Address feedback for accumulateTwoPhaseListeners

trueadm:
This is a follow up to the feedback @gaearon left on https://github.com/facebook/react/pull/18274. Notably:

- moves the `accumulateTwoPhaseListeners` function into its own file.
- check length rather than using a boolean variable to determine if the event's `_dispatchListeners` and `_dispatchInstances` should be updated.
- Inline `EventPropagators.js` functions that are not used by ReactDOM into the `ResponderPlugin` module, TestUtils and the React Native Bridge module.
- Flow type `EventPropagators.js` and remove unused functions

### DIFF
--- a/packages/legacy-events/ReactSyntheticEventType.js
+++ b/packages/legacy-events/ReactSyntheticEventType.js
@@ -31,8 +31,8 @@ export type ReactSyntheticEvent = {|
     nativeEventTarget: EventTarget,
   ) => ReactSyntheticEvent,
   isPersistent: () => boolean,
-  _dispatchInstances: null | Array<Fiber>,
-  _dispatchListeners: null | Array<Function>,
-  _targetInst: null | Fiber,
+  _dispatchInstances: null | Array<Fiber> | Fiber,
+  _dispatchListeners: null | Array<Function> | Function,
+  _targetInst: Fiber,
   type: string,
 |};
--- a/packages/react-native-renderer/src/ReactNativeBridgeEventPlugin.js
+++ b/packages/react-native-renderer/src/ReactNativeBridgeEventPlugin.js
@@ -9,21 +9,59 @@
 
 import type {AnyNativeEvent} from 'legacy-events/PluginModuleType';
 import type {EventSystemFlags} from 'legacy-events/EventSystemFlags';
-import {
-  accumulateTwoPhaseDispatches,
-  accumulateDirectDispatches,
-} from 'legacy-events/EventPropagators';
+import {accumulateDirectDispatches} from 'legacy-events/EventPropagators';
 import type {TopLevelType} from 'legacy-events/TopLevelEventTypes';
 import SyntheticEvent from 'legacy-events/SyntheticEvent';
 import invariant from 'shared/invariant';
 
 // Module provided by RN:
 import {ReactNativeViewConfigRegistry} from 'react-native/Libraries/ReactPrivate/ReactNativePrivateInterface';
+import accumulateInto from 'legacy-events/accumulateInto';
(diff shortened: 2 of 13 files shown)


gaearon:
> Unfortunately, we can't remove the old propagation code, because it's exported on ReactDOM's internals:

If it’s exported for the sake of TestUtils, can you copy it directly into TestUtils?

Same with Responser plugin. 

Then we won’t have two versions of this code in the main bundle.

trueadm:
@gaearon Is that a breaking change?

gaearon:
Does anything use it outside of TestUtils? Regarding Responder plugin, it depends on how that’s wired up on RNW. Can you find that on GitHub?

trueadm:
@gaearon Doesn't look like it. @necolas would know more for sure though.

trueadm:
@gaeron Too reduce risk, I've opted to inline the TestUtils, RN bridge and Responder bits of code they used from `EventPropagators.js` into their respective modules instead. Whilst at it, I also added Flow typing to `EventPropagators.js` to add some peace of mind. :)

necolas:
I don't think these changes will impact RNW. Hopefully will be dropping this dependency soon anyway

packages/legacy-events/ReactSyntheticEventType.js
```diff
@@ -31,8 +31,8 @@ export type ReactSyntheticEvent = {|
     nativeEventTarget: EventTarget,
   ) => ReactSyntheticEvent,
   isPersistent: () => boolean,
-  _dispatchInstances: null | Array<Fiber>,
-  _dispatchListeners: null | Array<Function>,
-  _targetInst: null | Fiber,
+  _dispatchInstances: null | Array<Fiber> | Fiber,
+  _dispatchListeners: null | Array<Function> | Function,
```
bvaughn: This type looks sketchy/unfortunate.

trueadm: It was a genuine fix. `_dispatchInstances`/`_dispatchListeners` can be a single node rather than array. We don't excipit it with the recent changes we made, but legacy `accumulate*` try to be clever in not allocating arrays where possible so the Flow type accounts for that now.

bvaughn: Yeah I noticed that about the accumulate method. Just seems like a very overloaded type, easy to mess up 😅

trueadm: Yeah, that's why I'm making these refactors. The sooner we move away from that style, the better. :)

bvaughn: Fair enough!

packages/react-native-renderer/src/ReactNativeBridgeEventPlugin.js
```diff
@@ -9,22 +9,60 @@
...
+  if (event && event.dispatchConfig.phasedRegistrationNames) {
+    traverseTwoPhase(event._targetInst, accumulateDirectionalDispatches, event);
+  }
+}
+
+function accumulateTwoPhaseDispatches(events) {
+  forEachAccumulated(events, accumulateTwoPhaseDispatchesSingle);
+}
+// End of inline
+type PropagationPhases = 'bubbled' | 'captured';
```
gaearon: Could we also use the new implementation here for RN?

trueadm: It's best right now to keep it separate IMO. The other version is likely to become tangled with the `useEvent` logic for ReactDOM, and we don't want that in RN.

PR #18292, opened 2020-03-12
ReactDOM.useEvent: Add DOM host event listener logic

trueadm:
This PR adds the DOM event listener logic. As the useEvent hook has not yet been wired up to the host config, it's not yet possible to add tests – but those will come in the next PR.

I've detailed with comments how each part of the event listener logic works with `useEvent` "listeners", so hopefully it should flow and makes sense for those reviewing. In essense, we have a new internal listeners Set that we attach to DOM target instances. Target instances, are those will be attached via `useEvent`'s `setListener` API. In a follow up PR, we'll use this listeners Set to find any attached listener functions, to know to fire them.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -69,7 +69,6 @@ import {
   enableUseEventAPI,
 } from 'shared/ReactFeatureFlags';
 import {HostComponent} from 'shared/ReactWorkTags';
-import invariant from 'shared/invariant';
 import {
   RESPONDER_EVENT_SYSTEM,
   IS_PASSIVE,
@@ -1093,14 +1092,6 @@ export function mountEventListener(listener: ReactDOMListener): void {
       // TODO (useEvent)
     } else if (isDOMElement(target)) {
       attachElementListener(listener);
-    } else {
-      // The user should never get to here or we missed something in
-      // validateEventListenerTarget below.
-      invariant(
-        false,
-        'A useEvent listener target instance was not handled. This error is ' +
-          'likely caused by a bug in React. Please file an issue.',
-      );
     }
   }
 }
@@ -1108,20 +1099,12 @@ export function mountEventListener(listener: ReactDOMListener): void {
 export function unmountEventListener(listener: ReactDOMListener): void {
   if (enableUseEventAPI) {
     const {target} = listener;
-    if (isDOMElement(target)) {
+    if (target === window) {
       // TODO (useEvent)
     } else if (isDOMDocument(target)) {
       // TODO (useEvent)
     } else if (isDOMElement(target)) {
       detachElementListener(listener);
-    } else {
-      // The user should never get to here or we missed something in
-      // validateEventListenerTarget below.
-      invariant(
-        false,
-        'A useEvent listener target instance was not handled. This error is ' +
-          'likely caused by a bug in React. Please file an issue.',
-      );
     }
   }
 }
@@ -1150,9 +1133,11 @@ export function validateEventListenerTarget(
     }
     if (__DEV__) {
       console.warn(
-        'Event listener method setListener() from useEvent() hook requires the first argument to be either a valid ' +
-          'DOM node that was rendered and managed by React, or either the "window" or "document" objects.' +
-          "If you are setting setListener from ref, ensure the ref's current value has been set and is not null.",
+        'Event listener method setListener() from useEvent() hook requires the first argument to be either:' +
+          '\n\n' +
+          '1. A valid DOM node that was rendered and managed by React' +
+          '2. The "window" object' +
+          '3. The "document" object',
       );
     }
   }
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -69,6 +69,7 @@ import {
   enableUseEventAPI,
 } from 'shared/ReactFeatureFlags';
 import {HostComponent} from 'shared/ReactWorkTags';
+import invariant from 'shared/invariant';
 import {
   RESPONDER_EVENT_SYSTEM,
   IS_PASSIVE,
@@ -76,6 +77,8 @@ import {
 import {
   attachElementListener,
   detachElementListener,
+  isDOMDocument,
+  isDOMElement,
 } from '../events/DOMModernPluginEventSystem';
 
 export type ReactListenerEvent = ReactDOMListenerEvent;
@@ -1086,8 +1089,18 @@ export function mountEventListener(listener: ReactDOMListener): void {
     const {target} = listener;
     if (target === window) {
       // TODO (useEvent)
-    } else {
+    } else if (isDOMDocument(target)) {
+      // TODO (useEvent)
+    } else if (isDOMElement(target)) {
       attachElementListener(listener);
+    } else {
+      // The user should never get to here or we missed something in
+      // validateEventListenerTarget below.
+      invariant(
+        false,
+        'A useEvent listener target instance was not handled. This error is ' +
+          'likely caused by a bug in React. Please file an issue.',
+      );
     }
   }
 }
@@ -1095,10 +1108,20 @@ export function mountEventListener(listener: ReactDOMListener): void {
 export function unmountEventListener(listener: ReactDOMListener): void {
   if (enableUseEventAPI) {
     const {target} = listener;
-    if (target === window) {
+    if (isDOMElement(target)) {
       // TODO (useEvent)
-    } else {
+    } else if (isDOMDocument(target)) {
+      // TODO (useEvent)
+    } else if (isDOMElement(target)) {
       detachElementListener(listener);
+    } else {
+      // The user should never get to here or we missed something in
+      // validateEventListenerTarget below.
+      invariant(
+        false,
+        'A useEvent listener target instance was not handled. This error is ' +
+          'likely caused by a bug in React. Please file an issue.',
+      );
     }
   }
 }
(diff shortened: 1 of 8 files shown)


packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1086,19 +1089,39 @@ export function mountEventListener(listener: ReactDOMListener): void {
...
+      // TODO (useEvent)
+    } else if (isDOMElement(target)) {
       attachElementListener(listener);
+    } else {
+      // The user should never get to here or we missed something in
+      // validateEventListenerTarget below.
+      invariant(
+        false,
+        'A useEvent listener target instance was not handled. This error is ' +
+          'likely caused by a bug in React. Please file an issue.',
```
bvaughn: Hm... I wonder if this invariant is necessary?

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1086,19 +1089,39 @@ export function mountEventListener(listener: ReactDOMListener): void {
...
+      );
     }
   }
 }
 
 export function unmountEventListener(listener: ReactDOMListener): void {
   if (enableUseEventAPI) {
     const {target} = listener;
-    if (target === window) {
+    if (isDOMElement(target)) {
```
bvaughn: You have `if (isDOMElement(target)) {` and `else if (isDOMElement(target)) {`

I think you meant for this one to be `target === window`.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1123,19 +1149,11 @@ export function validateEventListenerTarget(
...
-        console.warn(
-          'Event listener method setListener() from useEvent() hook requires the first argument to be a valid' +
-            ' DOM node that was rendered and managed by React or a "window" object. If this is' +
-            ' from a ref, ensure the ref value has been set before attaching.',
-        );
-      }
+      console.warn(
+        'Event listener method setListener() from useEvent() hook requires the first argument to be either a valid ' +
+          'DOM node that was rendered and managed by React, or either the "window" or "document" objects.' +
+          "If you are setting setListener from ref, ensure the ref's current value has been set and is not null.",
```
bvaughn: > Event listener method setListener() from useEvent() hook requires the first argument to be either a valid DOM node that was rendered and managed by React, or either the "window" or "document" objects. If you are setting setListener from ref, ensure the ref's current value has been set and is not null.

Small suggestion for tweaking the wording:
> Event listener method setListener() from useEvent() hook requires the first argument to be either:
>1. A valid DOM node that was rendered and managed by React
>2. The "window" object
>3. The "document" object
>
> If you are passing a ref to setListener, ensure the ref's current value is not null.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -218,6 +233,16 @@ function isMatchingRootContainer(
   );
 }
 
+export function isDOMElement(target: EventTarget): boolean {
+  const nodeType = ((target: any): Node).nodeType;
+  return nodeType != null && nodeType === ELEMENT_NODE;
```
bvaughn: Isn't the `!= null` check here unnecessary?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -218,6 +233,16 @@ function isMatchingRootContainer(
...
+}
+
+export function isDOMDocument(target: EventTarget): boolean {
+  const nodeType = ((target: any): Node).nodeType;
+  return nodeType === DOCUMENT_NODE;
```
bvaughn: Isn't the `!= null` check here unnecessary?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -308,12 +326,62 @@ export function dispatchEventForPluginEventSystem(
...
+  while (node !== null) {
+    const tag = node.tag;
+    // Once we encounter a host container or root container
+    // we can return their DOM instance.
+    if (tag === HostRoot || tag === HostPortal) {
+      return node.stateNode.containerInfo;
+    }
+    node = node.return;
+  }
+  return instance;
```
bvaughn: Is this sound? If we don't find a host or portal root, can we just return the closest instance safely? (I don't have a good mental model of this. Just asking.)

Would we ever expect to encounter this other than the `window` or `document`?

**Edit** I guess, based on your comments below, this fallback is ok.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -308,12 +326,62 @@ export function dispatchEventForPluginEventSystem(
...
+    containerEventTarget,
+    listenerMap,
+    passive,
+  );
+  // Get the internal listeners Set from the target instance.
+  let listeners = getListenersFromTarget(target);
+  // If we don't have any listeners, then we need to init them.
+  if (listeners === null) {
+    listeners = new Set();
+    initListenersSet(target, listeners);
```
bvaughn: Curiosity question: Why does `getListenerMapForElement` lazily initialize the Map but `initListenersSet` doesn't?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -308,12 +326,62 @@ export function dispatchEventForPluginEventSystem(
...
 
 export function detachElementListener(listener: ReactDOMListener): void {
-  // TODO
+  const {target} = listener;
+  // Get the internal listeners Set from the target instance.
+  const listeners = getListenersFromTarget(target);
+  if (listeners !== null) {
+    // Remove out listener from the listeners Set.
+    listeners.delete(listener);
+  }
```
bvaughn: Does this method need to clean up anything in the listener map?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: What `!= null`?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: `getListenerMapForElement` is something that has existed for a while. I avoiding lazily initializing the listener Set because we'll be doing a lot of lookups on various instances where they have no `useEvent` listeners, and creating and setting up a Set on those is far too much overhead when they'll likely never have `useEvent` listener Sets.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: We'll never encounter this with `window` or `document`. I've revised the logic to help make that clear.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: Flow complains otherwise.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: In the follow up PR we will clear any leaking `listeners`. The listeners Set itself can be GC'd.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
bvaughn: This comment was misplaced.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
bvaughn: Ugh. Cast it.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
bvaughn: It seems wrong that attaching adds things to map and set, but detaching only removes from set.

trueadm: Oh sorry, I missed your original point. Yep, I'll add clean-up here too.

bvaughn: > The listeners Map isn't new too, it's been around for years – and we've never cleared it down.

Right. I guess I was assuming we'd need some check/logic to determine if this was the last thing listening to it, and we could then clear it out...but if we never did this, I guess there's no need to start doing it now. Just feels like a bug when I look at it.

> Lastly, it's not a bad case here – to over-listen.

I guess not, but at the same time having an *active* listener attached could have perf implications?

Ultimately I trust your judgement here b'c you're much more familiar with the surrounding code than I am. Just raising things that look potentially wrong as I see it :)

trueadm: It's valid to bring this up. I brought this up with @sebmarkbage many moons ago and he explained it to me and as to why we do it. Might be worth bringing it up again some time in a design meeting.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1123,19 +1132,13 @@ export function validateEventListenerTarget(
...
-        console.warn(
-          'Event listener method setListener() from useEvent() hook requires the first argument to be a valid' +
-            ' DOM node that was rendered and managed by React or a "window" object. If this is' +
-            ' from a ref, ensure the ref value has been set before attaching.',
-        );
-      }
+      console.warn(
+        'Event listener method setListener() from useEvent() hook requires the first argument to be either:' +
+          '\n\n' +
+          '1. A valid DOM node that was rendered and managed by React' +
```
gaearon: This is missing newlines. You need a newline after each of these.

PR #18293, opened 2020-03-12
Add <progress> to DOM fixtures

### LINKED ISSUE #13518: Add <progress> to DOM fixtures
Maybe I missed it but I don't see this case in DOM fixtures:

https://github.com/facebook/react/pull/7359#discussion_r77518170

Seems important because it's easy to accidentally gloss over it if we stop syncing attributes and have to edit a bunch of jsdom tests.

bl00mber:
resolve #13518

gif


fixtures/dom/src/components/fixtures/progress/index.js
```diff
@@ -0,0 +1,95 @@
...
+                  {' '}
+                  value:{' '}
+                  {JSON.stringify(
+                    Math.round((this.state.value + Number.EPSILON) * 100) / 100
+                  )}
+                </span>
+              </fieldset>
+
+              <fieldset>
+                <legend>Uncontrolled</legend>
```
gaearon, line 81: What is this example about? Progress doesn't have change events, so the notion of Uncontrolled doesn't make sense here, right?

bl00mber: I thought it may have sense as a demonstration of empty bar :persevere:

gaearon:
Thanks.

PR #18296, opened 2020-03-13
Failing test case for useMutableSource

acdlite:
Scenario: `getSnapshot` changes during render. The render finishes and commits. But before the passive effects fire, the source is mutated again. The subscription still thinks the old `getSnapshot` is current.

I think this means that whenever `getSnapshot` changes, we need to check if there was a mutation since render. Similar to what we do when resubscribing.

### DIFF
--- a/packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
@@ -43,8 +43,8 @@ describe('useMutableSource', () => {
     const callbacksA = [];
     const callbacksB = [];
     let revision = 0;
-    let valueA = 'a:one';
-    let valueB = 'b:one';
+    let valueA = initialValueA;
+    let valueB = initialValueB;
 
     const subscribeHelper = (callbacks, callback) => {
       if (callbacks.indexOf(callback) < 0) {
@@ -630,7 +630,7 @@ describe('useMutableSource', () => {
     });
 
     it('should only update components whose subscriptions fire', () => {
-      const source = createComplexSource('one', 'one');
+      const source = createComplexSource('a:one', 'b:one');
       const mutableSource = createMutableSource(source);
 
       // Subscribe to part of the store.
@@ -672,7 +672,7 @@ describe('useMutableSource', () => {
     });
 
     it('should detect tearing in part of the store not yet subscribed to', () => {
-      const source = createComplexSource('one', 'one');
+      const source = createComplexSource('a:one', 'b:one');
       const mutableSource = createMutableSource(source);
 
       // Subscribe to part of the store.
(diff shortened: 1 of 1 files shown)


bvaughn:
Nice! Thanks for the test case. I'll fix this in the morning.

bvaughn:
Thanks again!

#18297

acdlite:
@bvaughn I pushed another test case. Mind taking a look at that, too?

bvaughn:
On it.

acdlite:
Also I probably shouldn't have written the description as if this were specific to passive effects, since it applies to any mutation in the render phase.

acdlite:
I think what the second test case indicates is that we aren't resetting the queue properly here:

```js
stateHook.baseQueue = null;
snapshot = readFromUnsubcribedMutableSource(root, source, getSnapshot);
stateHook.memoizedState = stateHook.baseState = snapshot;
```

That clears the updates, but what we also need to do is create an entire new `queue` object and re-bind the `setState` method. So that even if there are interleaved updates, they get pushed to the older queue.

bvaughn:
Yeah, I was noticing that.

Thanks for the pointer :smile:

acdlite:
More concretely:

```js
const newQueue = {
  pending: null,
  dispatch: null,
  lastRenderedReducer: basicStateReducer,
  lastRenderedState: whatever,
};
newQueue.dispatch = dispatchAction.bind(currentlyRenderingFiber, newQueue);

// Now when this becomes current, the previous queue and dispatch method
// are complete discarded, including any interleaving updates that occur.
hook.queue = newQueue;
```

bvaughn:
😆 Yes, thanks...

bvaughn:
From my investigation, the effect that updates the `getSnapshot` ref *does* run and queue a state update for the new test case- but the one in `handleChange` is overriding with the old `getSnapshot` value. So...disconnecting the old queue/dispatch function seems reasonable.

It doesn't fix this case though, and it also breaks several others 😅so I need to dig some more.

bvaughn:
It's interesting that removing the first hook "fixes" the problem above for the second hook too.

bvaughn:
I think that difference is because the eager state computation is only happening for the first state hook on a given fiber that schedules an update. So if `second` is the only one, its state gets eagerly evaluated during the mutation as "b0" and then later updates from "b0" -> "a1" when the `getSnapshot` syncing effect runs.

If `first` is the *first* state hook to update, `second` state doesn't get eagerly snapshot and when the `getSnapshot` syncing effect runs and schedules an update to the newly mutated value, the old change event handler value is what actually ends up in state.

bvaughn:
I think I can write a similarly broken test case with just the state hook, but maybe I'm misunderstanding something.

```js
it("repro", async () => {
  let setStateA;
  let setStateB;

  function App({ count }) {
    const [a, setA] = React.useState("initial-a");
    const [b, setB] = React.useState("initial-b");
    React.useEffect(() => {
      if (count === 1) {
        setA("cascading-a");
        setB("cascading-b");
      }
    }, [count]);
    setStateA = setA;
    setStateB = setB;
    return `${a}:${b}`;
  }

  const root = ReactNoop.createRoot();
  await act(async () => {
    root.render(<App count={0} />);
  });

  await act(async () => {
    root.render(<App count={1} />);
    // Render and finish the tree, but yield before passive effects fire.
    expect(Scheduler).toFlushUntilNextPaint([]);
    // This is at high priority so that it doesn't get batched with default
    // priority updates that might fire during the passive effect
    ReactNoop.discreteUpdates(() => {
      setStateB("high-pri-b");
    });
    expect(Scheduler).toFlushUntilNextPaint([]);
    expect(root.getChildrenAsJSX()).toEqual("cascading-a:high-pri-b");
  });
});
```

Rather than a final state of `"cascading-a:high-pri-b"`, the above results in a final state of `"initial-a:high-pri-b"`.

acdlite:
The test you wrote is correct. I think the confusion is because there's another paint after that one, at normal priority, that does the cascading updates. Here's that same test with more assertions added:

```js
it('repro', async () => {
  let setStateA;
  let setStateB;

  function App({count}) {
    const [a, setA] = React.useState('initial-a');
    const [b, setB] = React.useState('initial-b');
    React.useEffect(() => {
      if (count === 1) {
        setA('cascading-a');
        setB('cascading-b');
      }
    }, [count]);
    setStateA = setA;
    setStateB = setB;
    return `${a}:${b}`;
  }

  const root = ReactNoop.createRoot();
  await act(async () => {
    root.render(<App count={0} />);
  });

  await act(async () => {
    root.render(<App count={1} />);
    // Render and finish the tree, but yield before passive effects fire.
    expect(Scheduler).toFlushUntilNextPaint([]);
    expect(root).toMatchRenderedOutput('initial-a:initial-b');
    // This is at high priority so that it doesn't get batched with default
    // priority updates that might fire during the passive effect
    ReactNoop.discreteUpdates(() => {
      setStateB('high-pri-b');
    });
    expect(Scheduler).toFlushUntilNextPaint([]);
    // High pri updates have finished, but not the cascading ones,
    // which have normal pri
    expect(root).toMatchRenderedOutput('initial-a:high-pri-b');
  });
  // Now the final state has rendered, including the cascading updates
  expect(root).toMatchRenderedOutput('cascading-a:cascading-b');
});
```

bvaughn:
I see. Thanks for clarifying.

I wonder why I'm seeing the update clobber in the way I'm seeing for uMS's internal state hook.

bvaughn:
Okay, looking at this again today.

Disconnecting dispatch from the update queue works to prevent the old B selector from overriding the new A selector value when A is mutated- but the mutation to A is also missed temporarily (since the subscribe function is still using the previous `getSnapshot` at this point).

The effect that updates `getSnapshot` sees the change, and uses the newly bound `dispatchAction` method to schedule the "missed" update, but when the component renders- the "missed" update is scheduled at an insufficient expiration time, so it's skipped for that update (and tearing happens, for A<sub>0</sub> and A<sub>1</sub>).

acdlite:
Gah I did it again!

acdlite:
@bvaughn 

> I think the getSnapshot effect needs to schedule its missed update at a higher priority, to make sure it's included in the next render- that or just replace the intermediate update entirely (but that seems shady).

I think that's the right idea, except it's not only limited to high priority updates. It could be a lower priority update, too. (I pushed another test case to cover that.)

What's tricky is there could be multiple pending mutations at different priorities. None of them should be allowed to commit without also committing the new update.

We've been using the word "entanglement" to describe this behavior. We don't have a mechanism to do this without de-opting, currently, but I'm planning to add one as part of the `useTransition` improvements. (This is the "terminal updates" problem.)

In the meantime, I think we should synchronously flush the pending mutation updates. Then we can do the entanglement thing once that exists.

I'll push a commit to show what I mean.

packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
```diff
@@ -1397,6 +1397,116 @@ describe('useMutableSource', () => {
...
+        // so it doesn't interrupt the in-progress render.
+        Scheduler.unstable_runWithPriority(
+          Scheduler.unstable_IdlePriority,
+          () => {
+            source.valueB = '3';
+          },
+        );
+      });
+      expect(Scheduler).toHaveYielded([
+        'TODO: This currently tears. Fill in with correct values once bug',
```
bvaughn: I think this was meant to be a comment? :smile:

bvaughn:
Thanks for the added context, Andrew.

Using `markRootExpiredAtTime` means that tests two and three will temporarily tear. (I'm actually not seeing any difference in behavior for the tests whether I use `markRootExpiredAtTime` or not.)

PR #18297, opened 2020-03-13
useMutableSource: bugfix for new getSnapshot with mutation

bvaughn:
Pulls in @acdlite's failing tests from #18296.

Fixes one, leaves TODO comment for addressing the other two once we have a mechanism to do so. (Currently some temporary tearing is possible between passive effect flushes.)

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -75,7 +75,8 @@ import {
   getCurrentPriorityLevel,
 } from './SchedulerWithReactIntegration';
 import {
-  getPendingExpirationTime,
+  getFirstPendingExpirationTime,
+  getLastPendingExpirationTime,
   getWorkInProgressVersion,
   markSourceAsDirty,
   setPendingExpirationTime,
@@ -916,16 +917,14 @@ function readFromUnsubcribedMutableSource<Source, Snapshot>(
     isSafeToReadFromSource = currentRenderVersion === version;
   } else {
     // If there's no version, then we should fallback to checking the update time.
-    const pendingExpirationTime = getPendingExpirationTime(root);
+    const pendingExpirationTime = getFirstPendingExpirationTime(root);
 
     if (pendingExpirationTime === NoWork) {
       isSafeToReadFromSource = true;
     } else {
       // If the source has pending updates, we can use the current render's expiration
       // time to determine if it's safe to read again from the source.
-      isSafeToReadFromSource =
-        pendingExpirationTime === NoWork ||
-        pendingExpirationTime >= renderExpirationTime;
+      isSafeToReadFromSource = pendingExpirationTime >= renderExpirationTime;
     }
 
     if (isSafeToReadFromSource) {
@@ -974,7 +973,8 @@ function useMutableSource<Source, Snapshot>(
 
   const dispatcher = ReactCurrentDispatcher.current;
 
-  const [currentSnapshot, setSnapshot] = dispatcher.useState(() =>
+  // eslint-disable-next-line prefer-const
+  let [currentSnapshot, setSnapshot] = dispatcher.useState(() =>
     readFromUnsubcribedMutableSource(root, source, getSnapshot),
   );
   let snapshot = currentSnapshot;
@@ -1032,7 +1032,7 @@ function useMutableSource<Source, Snapshot>(
           // There is no mechanism currently to associate these updates though,
           // so for now we fall back to synchronously flushing all pending updates.
           // TODO: Improve this later.
-          markRootExpiredAtTime(root, getPendingExpirationTime(root));
+          markRootExpiredAtTime(root, getLastPendingExpirationTime(root));
         }
       }
     }
@@ -1091,7 +1091,7 @@ function useMutableSource<Source, Snapshot>(
         // We missed a mutation before committing.
         // It's possible that other components using this source also have pending updates scheduled.
         // In that case, we should ensure they all commit together.
-        markRootExpiredAtTime(root, getPendingExpirationTime(root));
+        markRootExpiredAtTime(root, getLastPendingExpirationTime(root));
       }
     }
 
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -999,56 +999,44 @@ function useMutableSource<Source, Snapshot>(
     subscribe,
   }: MutableSourceMemoizedState<Source, Snapshot>);
 
-  // Sync the values needed by our subscribe function after each commit.
+  // Sync the values needed by our subscription handler after each commit.
   dispatcher.useEffect(() => {
-    const didGetSnapshotChange = !is(refs.getSnapshot, getSnapshot);
     refs.getSnapshot = getSnapshot;
 
     // Normally the dispatch function for a state hook never changes,
-    // but in the case of this hook, it will change if getSnapshot changes.
-    // In that case, the subscription below will have cloesd over the previous function,
-    // so we use a ref to ensure that handleChange() always has the latest version.
+    // but this hook recreates the queue in certain cases  to avoid updates from stale sources.
+    // handleChange() below needs to reference the dispatch function without re-subscribing,
+    // so we use a ref to ensure that it always has the latest version.
     refs.setSnapshot = setSnapshot;
 
-    // This effect runs on mount, even though getSnapshot hasn't changed.
-    // In that case we can avoid the additional checks for a changed snapshot,
-    // because the subscription effect below will cover this.
-    if (didGetSnapshotChange) {
-      // Because getSnapshot is shared with subscriptions via a ref,
-      // we don't resubscribe when getSnapshot changes.
-      // This means that we also don't check for any missed mutations
-      // between the render and the passive commit though.
-      // So we need to check here, just like when we newly subscribe.
-      const maybeNewVersion = getVersion(source._source);
-      if (!is(version, maybeNewVersion)) {
-        const maybeNewSnapshot = getSnapshot(source._source);
-        if (!is(snapshot, maybeNewSnapshot)) {
-          setSnapshot(maybeNewSnapshot);
+    // Check for a possible change between when we last rendered now.
+    const maybeNewVersion = getVersion(source._source);
+    if (!is(version, maybeNewVersion)) {
+      const maybeNewSnapshot = getSnapshot(source._source);
+      if (!is(snapshot, maybeNewSnapshot)) {
+        setSnapshot(maybeNewSnapshot);
 
-          const currentTime = requestCurrentTimeForUpdate();
-          const suspenseConfig = requestCurrentSuspenseConfig();
-          const expirationTime = computeExpirationForFiber(
-            currentTime,
-            fiber,
-            suspenseConfig,
-          );
-          setPendingExpirationTime(root, expirationTime);
+        const currentTime = requestCurrentTimeForUpdate();
+        const suspenseConfig = requestCurrentSuspenseConfig();
+        const expirationTime = computeExpirationForFiber(
+          currentTime,
+          fiber,
+          suspenseConfig,
+        );
+        setPendingExpirationTime(root, expirationTime);
 
-          // If the source mutated between render and now,
-          // there may be state updates already scheduled from the old getSnapshot.
-          // Those updates should not commit without this value.
-          // There is no mechanism currently to associate these updates though,
-          // so for now we fall back to synchronously flushing all pending updates.
-          // TODO: Improve this later.
-          markRootExpiredAtTime(root, getLastPendingExpirationTime(root));
-        }
+        // If the source mutated between render and now,
+        // there may be state updates already scheduled from the old getSnapshot.
+        // Those updates should not commit without this value.
+        // There is no mechanism currently to associate these updates though,
+        // so for now we fall back to synchronously flushing all pending updates.
+        // TODO: Improve this later.
+        markRootExpiredAtTime(root, getLastPendingExpirationTime(root));
       }
     }
-  }, [getSnapshot]);
+  }, [getSnapshot, source, subscribe]);
 
-  // If we got a new source or subscribe function,
-  // we'll need to subscribe in a passive effect,
-  // and also check for any changes that fire between render and subscribe.
+  // If we got a new source or subscribe function, re-subscribe in a passive effect.
   dispatcher.useEffect(() => {
     const handleChange = () => {
       const latestGetSnapshot = refs.getSnapshot;
@@ -1089,29 +1077,6 @@ function useMutableSource<Source, Snapshot>(
       }
     }
 
-    // Check for a possible change between when we last rendered and when we just subscribed.
-    const maybeNewVersion = getVersion(source._source);
-    if (!is(version, maybeNewVersion)) {
-      const maybeNewSnapshot = getSnapshot(source._source);
-      if (!is(snapshot, maybeNewSnapshot)) {
-        setSnapshot(maybeNewSnapshot);
-
-        const currentTime = requestCurrentTimeForUpdate();
-        const suspenseConfig = requestCurrentSuspenseConfig();
-        const expirationTime = computeExpirationForFiber(
-          currentTime,
-          fiber,
-          suspenseConfig,
-        );
-        setPendingExpirationTime(root, expirationTime);
-
-        // We missed a mutation before committing.
-        // It's possible that other components using this source also have pending updates scheduled.
-        // In that case, we should ensure they all commit together.
-        markRootExpiredAtTime(root, getLastPendingExpirationTime(root));
-      }
-    }
-
     return unsubscribe;
   }, [source, subscribe]);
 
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -998,7 +999,38 @@ function useMutableSource<Source, Snapshot>(
...
+      const maybeNewVersion = getVersion(source._source);
+      if (!is(version, maybeNewVersion)) {
+        const maybeNewSnapshot = getSnapshot(source._source);
+        if (!is(snapshot, maybeNewSnapshot)) {
+          // If the source mutated between render and now,
+          // we may have a state update already scheduled from the old getSnapshot.
+          // It's important to ensure that this update is included with that one.
+          // TODO This sucks; what's the right way of doing this?
+          runWithPriority(UserBlockingPriority, () => {
+            setSnapshot(maybeNewSnapshot);
```
bvaughn: As I mentioned [in this comment](https://github.com/facebook/react/pull/18296#issuecomment-600238790), I think what I need to do here is schedule a state update with the same expiration time as the one that might be pending due to a mutation. (Either that or manually replace the pending update with this one, but that seemed a little shady too.)

Added this "fix" in the meanwhile, for discussion purposes. :smile:

bvaughn:
I know you've been busy @acdlite, but friendly ping on this :smile:

Dave was talking to me this afternoon about a bug he's seeing with Recoil and I'd like to at least sync this so we can see if the fix helped or not.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1025,9 +1062,11 @@ function useMutableSource<Source, Snapshot>(
         // e.g. it might try to read from a part of the store that no longer exists.
         // In this case we should still schedule an update with React.
         // Worst case the selector will throw again and then an error boundary will handle it.
-        setSnapshot(() => {
-          throw error;
-        });
+        latestSetSnapshot(
```
acdlite: I think you need to do the same `markRootExpiredAtTime` thing during initial mount and when resubscribing, too, to avoid a momentary flicker.

bvaughn: Hm...do we?

A change firing between render and subscription means that we missed an update, but we have other checks in place to avoid actually tearing between other components that read from the source.

bvaughn: Put another way, the reason we added it where we did was because there's a potential tear- but in the subscribe-on-commit case, there's not a tear, just a potential missed update. Also at that point, we would have already shown the older value because we subscribe in a passive effect.

Unless I'm misunderstanding what you're suggesting?

bvaughn:
I think this test captures the mount/flicker case we talked about:
```js
it("blah", async () => {
  const source = createSource("one");
  const mutableSource = createMutableSource(source);

  let committedA = null;
  let committedB = null;

  const onRender = () => {
    if (committedB !== null) {
      expect(committedA).toBe(committedB);
    }
  };

  function ComponentA() {
    const snapshot = useMutableSource(
      mutableSource,
      defaultGetSnapshot,
      defaultSubscribe
    );
    Scheduler.unstable_yieldValue(`a:${snapshot}`);
    React.useEffect(() => {
      committedA = snapshot;
    }, [snapshot]);
    return <div>{`a:${snapshot}`}</div>;
  }
  function ComponentB() {
    const snapshot = useMutableSource(
      mutableSource,
      defaultGetSnapshot,
      defaultSubscribe
    );
    Scheduler.unstable_yieldValue(`b:${snapshot}`);
    React.useEffect(() => {
      committedB = snapshot;
    }, [snapshot]);
    return <div>{`b:${snapshot}`}</div>;
  }

  // Mount ComponentA with data version 1
  act(() => {
    ReactNoop.render(
      <React.Profiler id="root" onRender={onRender}>
        <ComponentA />
      </React.Profiler>,
      () => Scheduler.unstable_yieldValue("Sync effect")
    );
  });
  expect(Scheduler).toHaveYielded(["a:one", "Sync effect"]);
  expect(source.listenerCount).toBe(1);

  // Mount ComponentB with version 1 (but don't commit it)
  act(() => {
    ReactNoop.render(
      <React.Profiler id="root" onRender={onRender}>
        <ComponentA />
        <ComponentB />
      </React.Profiler>,
      () => Scheduler.unstable_yieldValue("Sync effect")
    );
    expect(Scheduler).toFlushAndYieldThrough(["a:one", "b:one", "Sync effect"]);
    expect(source.listenerCount).toBe(1);

    // Mutate -> schedule update for ComponentA
    Scheduler.unstable_runWithPriority(Scheduler.unstable_IdlePriority, () => {
      source.value = "two";
    });

    // Commit ComponentB -> notice the change and schedule an update for ComponentB
    expect(Scheduler).toFlushAndYield(["a:two", "b:two"]);
    expect(source.listenerCount).toBe(2);
  });
});

```

bvaughn:
I think the problem with what we talked about - only storing the lowest priority (highest expiration time) pending update - is that we are trying to use a single value for two things:
1. Flush all pending work for a root in certain cases (using `markRootExpiredAtTime`). The lowest priority works well for this.
2. Determine if it's safe to read from a source when we haven't subscribed yet or when `getSnapshot` changes (by comparing the pending time to `renderExpirationTime`). The lowest priority doesn't work for this because it can give us a false positive.

packages/react-reconciler/src/ReactFiberHooks.js
bvaughn: Follow up on the Dave /Recoil comment: Been stepping through that today, and it looks like the Recoil getter is also a setter, which is causing a loop. (In other words I think it's a Recoil problem, not a uMS problem.)

acdlite:
For the other types of pending work, we track the range:

https://github.com/facebook/react/blob/58afba06636300fd652924e3033916fc96142e13/packages/react-reconciler/src/ReactFiberRoot.js#L66-L73

We can do the same thing here

bvaughn:
> For the other types of pending work, we track the range

Yeah. That's what I ended up thinking too. I'll take a pass at it soon.

bvaughn:
> For the other types of pending work, we track the range

Yeah. That's what I ended up thinking too. I'll take a pass at it soon.

FWIW, I did try doing a naive version of this (just tracking lowest and highest priority) and it didn't seem sufficient, but I'll take another pass.

bvaughn:
Gah. I had a small, stupid mistake in my first try at the separate expiration times, and it just took me 2 hours of staring at the code to spot it.

bvaughn:
Ok back to you @acdlite

packages/react-reconciler/src/ReactMutableSource.js
```diff
@@ -30,20 +30,42 @@ export function clearPendingUpdates(
   root: FiberRoot,
   expirationTime: ExpirationTime,
 ): void {
-  if (root.mutableSourcePendingUpdateTime <= expirationTime) {
-    root.mutableSourcePendingUpdateTime = NoWork;
+  if (root.mutableSourceFirstPendingUpdateTime <= expirationTime) {
+    root.mutableSourceFirstPendingUpdateTime = NoWork;
```
acdlite, line 34: Needs to account for when the `expirationTime` falls in between the first and last part of the range. Here's the equivalent code for normal updates:

https://github.com/facebook/react/blob/5e464546af4a32c5ae5e704f68bcf4b98a4c440a/packages/react-reconciler/src/ReactFiberRoot.js#L253-L258

packages/react-reconciler/src/ReactMutableSource.js
```diff
@@ -30,20 +30,42 @@ export function clearPendingUpdates(
...
+  // It's inclusive of all other pending updates.
+  root.mutableSourceLastPendingUpdateTime = Math.max(
+    root.mutableSourceLastPendingUpdateTime,
+    expirationTime,
+  );
+
+  if (root.mutableSourceFirstPendingUpdateTime === NoWork) {
+    root.mutableSourceFirstPendingUpdateTime = expirationTime;
+  } else {
+    root.mutableSourceFirstPendingUpdateTime = Math.min(
```
acdlite, line 64: Nit: use comparison operators instead of Math.max like we do elsewhere. Easier to optimize. As an example, note how this does an assignment every time this function is called, even if the value isn't changed.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1048,6 +1087,11 @@ function useMutableSource<Source, Snapshot>(
       const maybeNewSnapshot = getSnapshot(source._source);
       if (!is(snapshot, maybeNewSnapshot)) {
         setSnapshot(maybeNewSnapshot);
+
+        // We missed a mutation before committing.
+        // It's possible that other components using this source also have pending updates scheduled.
+        // In that case, we should ensure they all commit together.
+        markRootExpiredAtTime(root, getLastPendingExpirationTime(root));
```
acdlite, line 1094: Need to mark this new `setSnapshot` call as a pending mutation, too, right before marking it as expired.

packages/react-reconciler/src/ReactMutableSource.js
bvaughn: This change makes sense after we talked about it, although I can't wait for your expiration times refactor now because they make less sense than I thought before 😅

packages/react-reconciler/src/ReactFiberHooks.js
bvaughn: This change makes sense conceptually, but it breaks the [most recent test case](https://github.com/facebook/react/pull/18297#issuecomment-606898157) I added.

In this new test I...
1. Mount and flush component A with data version 1.
2. Mount component B with data version 1 (but flush passive effects).
3. Mutate the data (which schedules an update for already mounted component A.
4. Flush passive effects (which also schedules an update for the newly mounted component B).
5. Verify there was no tearing.

**Before** this change, both components A and B rendered together after we flushed passive effects.

**After** this change, component B renders and commits, then component A (and we tear in between). This is because- with the change in place, we're marking the root as expired with a higher priority (e.g. 1073741296 instead of 2/idle).

I think this indicates that we should be marking root as expired with the *first* expiration time rather than the *last* in this case? At least, doing that fixes this test (and also the other two failing tests that were tearing before). This also means that we aren't *directly* using last expiration time for anything at this point (although indicate it determines when we reset first expiration time in `clearPendingUpdates`).

bvaughn: I'm going to push this change as a separate commit (rather than squashing) just to give you the opportunity to review it more closely: 7c9181f

Back to you @acdlite

bvaughn:
> This is flipped.

Ah. Assuming you just mean in this one place (and not everywhere) then I think tracking the "first" time may not be necessary. (Same thing as I said with my comment above, but with the labels swapped since I still can't think of the `expirationTime` with the lowest value as being the "last" time.

Going to push this change as a separate commit as well, to make it easier for you to review in isolation: db511f4

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1065,11 +1126,31 @@ function useMutableSource<Source, Snapshot>(
...
   // In both cases, we need to throw away pending udpates (since they are no longer relevant)
   // and treat reading from the source as we do in the mount case.
+  const didGetSnapshotChange = !is(prevGetSnapshot, getSnapshot);
   if (
     !is(prevSource, source) ||
     !is(prevSubscribe, subscribe) ||
-    !is(prevGetSnapshot, getSnapshot)
+    didGetSnapshotChange
   ) {
+    if (didGetSnapshotChange) {
```
acdlite, line 1135: Can remove this nested check. Need to drop updates from new sources, too.

bvaughn: Yeah, I wasn't sure about that... I'll remove it :smile:

Thanks for the review!

bvaughn: Hm, I think removing this check actually breaks some subtle assumptions about the effects and when we need to update the `getSnapshot`/`setSnapshot` refs. I think it would cause us to drop updates for mutations after a source changed... need to think about it a bit.

acdlite: I think it works because when `source` or `subscribe` changes, you're going to unsubscribe/resubscribe.

When `getSnapshot` changes, you don't resubscribe but you do track the latest value with a ref. Which I think of as an optimization for resubscribing; you're still "changing" the subscription in both cases. So they are the same thing except for what we do in the commit phase. (For example, I think all your tests would pass if you removed the effect and ref that tracks `getSnapshot` and instead added `getSnapshot` as a dependency to the  resubscribe hook.)

bvaughn: Ah, I see how to fix it. Will let me simplify things a bit too...

acdlite: > For example, I think all your tests would pass if you removed the effect and ref that tracks `getSnapshot` and instead added `getSnapshot` as a dependency to the  resubscribe hook.

Just checked this to confirm and they do all pass

bvaughn: Okay, here's what 86da5ce did:

Recreating the queue and dispatch if `source` or `subscribe` changed too (not just `getSnapshot`) meant that the we needed to update `refs.setSnapshot` in that case too. (Before that effect was only run when `getSnapshot` changed.)

So I added `source` or `subscribe` as dependencies to the first effect so it would update the `dispatch` ref. Since we're now *also* running this effect whenever we resubscribe, I was also able to consolidate the check for a changed snapshot between render and passive effects into just the first effect (rather than both).

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -999,56 +999,44 @@ function useMutableSource<Source, Snapshot>(
...
   dispatcher.useEffect(() => {
-    const didGetSnapshotChange = !is(refs.getSnapshot, getSnapshot);
     refs.getSnapshot = getSnapshot;
 
     // Normally the dispatch function for a state hook never changes,
-    // but in the case of this hook, it will change if getSnapshot changes.
-    // In that case, the subscription below will have cloesd over the previous function,
-    // so we use a ref to ensure that handleChange() always has the latest version.
+    // but this hook recreates the queue in certain cases  to avoid updates from stale sources.
+    // handleChange() below needs to reference the dispatch function without re-subscribing,
```
acdlite, line 1008: I think if you go with this approach then you don't have to track `getSnapshot` or `setSnapshot` at all. Can use the closed over values, since `getSnapshot` is now a dep, and `setSnapshot` only changes when one of the other deps does.

packages/react-reconciler/src/ReactFiberHooks.js
bvaughn: Weird. I didn't see any of your interleaved commits until just now, after I left mine. Heh. Didn't mean to seem like I was ignoring you.

> For example, I think all your tests would pass if you removed the effect and ref that tracks getSnapshot and instead added getSnapshot as a dependency to the resubscribe hook.

This requires us to over-subscribe though, (since we can't share that effect without running the previous cleanup function and *unsubscribing*). Once we replace `useEffect` with our own custom pushed effect, we can optimize this though.

bvaughn: To be clear, I'm trying to avoid re-subscribing unless *actually necessary* (not just convenient) because subscriptions can be expensive for some of `useMutableSource`'s users.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1066,10 +1092,26 @@ function useMutableSource<Source, Snapshot>(
   // In both cases, we need to throw away pending udpates (since they are no longer relevant)
   // and treat reading from the source as we do in the mount case.
   if (
+    !is(prevGetSnapshot, getSnapshot) ||
     !is(prevSource, source) ||
-    !is(prevSubscribe, subscribe) ||
-    !is(prevGetSnapshot, getSnapshot)
+    !is(prevSubscribe, subscribe)
```
acdlite, line 1097: This is neat that this is now the same condition that the effect hook uses when comparing deps. In the future we could  "cheat" and read the deps off the effect hook instead of using a ref.

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: Edit: Got confused, thought you had removed the `getSnapshot` optimization. Never mind.

bvaughn: No worries. This is something we can further optimize when we replace the two composed effects with our own pushed effect. Then we can just use one single effect and selectively unsubscribe/subscribe only when necessary.

Sounds like that will be a fun little cleanup refactor anyway. :smile:

packages/react-reconciler/src/ReactFiberHooks.js
acdlite: Right, I can see why you did it this way. I think we might want to revisit the trade off in the future, though, since changing `getSnapshot` is expensive regardless because it leads to deopts. So it needs to be relatively stable. (I am kinda worried about this, since it's easy for users to mess up.) And if we assume it's relatively stable, then resubscribing might not be so bad.

bvaughn: > since changing getSnapshot is expensive regardless because it leads to deopts. So it needs to be relatively stable.

Yeah, absolutely. Maybe there's something we could do to detect when people are passing inline functions, so we could warn.

PR #18299, opened 2020-03-13
add jsx-runtime and jsx-dev-runtime modules

lunaruan:
This PR adds the `jsx-runtime` and `jsx-dev-runtime` modules for the JSX Babel Plugin. WWW still relies on `jsx/jsxs/jsxDEV` from the `"react"` module, so once we refactor the code to point to the runtime modules we will remove `jsx/jsxs/jsxDEV` from the `"react"` module.

### DIFF
--- a/packages/react/index.experimental.js
+++ b/packages/react/index.experimental.js
@@ -45,8 +45,4 @@ export {
   unstable_withSuspenseConfig,
   // enableBlocksAPI
   block,
-  // enableJSXTransformAPI
-  jsx,
-  jsxs,
-  jsxDEV,
 } from './src/React';
--- a/packages/react/index.js
+++ b/packages/react/index.js
@@ -76,7 +76,4 @@ export {
   DEPRECATED_createResponder,
   unstable_createFundamental,
   unstable_createScope,
-  jsx,
-  jsxs,
-  jsxDEV,
 } from './src/React';
--- a/packages/react/npm/jsx-runtime.js
+++ b/packages/react/npm/jsx-runtime.js
@@ -1,0 +1,7 @@
+'use strict';
+
+if (process.env.NODE_ENV === 'production') {
+  module.exports = require('./cjs/react-jsx-runtime.production.min.js');
+} else {
+  module.exports = require('./cjs/react-jsx-runtime.development.js');
+}
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -14,7 +14,9 @@
     "build-info.json",
     "index.js",
     "cjs/",
-    "umd/"
+    "umd/",
+    "jsx-runtime.js",
+    "jsx-dev-runtime.js"
   ],
   "main": "index.js",
   "repository": {
--- a/packages/react/src/React.js
+++ b/packages/react/src/React.js
@@ -24,7 +24,6 @@ import {
   createFactory as createFactoryProd,
   cloneElement as cloneElementProd,
   isValidElement,
-  jsx as jsxProd,
 } from './ReactElement';
 import {createContext} from './ReactContext';
 import {lazy} from './ReactLazy';
@@ -52,9 +51,6 @@ import {
   createElementWithValidation,
   createFactoryWithValidation,
   cloneElementWithValidation,
-  jsxWithValidation,
-  jsxWithValidationStatic,
-  jsxWithValidationDynamic,
 } from './ReactElementValidator';
 import createMutableSource from './createMutableSource';
 import ReactSharedInternals from './ReactSharedInternals';
@@ -66,12 +62,6 @@ import createScope from 'shared/createScope';
 const createElement = __DEV__ ? createElementWithValidation : createElementProd;
 const cloneElement = __DEV__ ? cloneElementWithValidation : cloneElementProd;
 const createFactory = __DEV__ ? createFactoryWithValidation : createFactoryProd;
-
-const jsxDEV = __DEV__ ? jsxWithValidation : undefined;
-const jsx = __DEV__ ? jsxWithValidationDynamic : jsxProd;
-// we may want to special case jsxs internally to take advantage of static children.
-// for now we can ship identical prod functions
-const jsxs = __DEV__ ? jsxWithValidationStatic : jsxProd;
 
 const Children = {
   map,
@@ -127,9 +117,4 @@ export {
   createFundamental as unstable_createFundamental,
   // enableScopeAPI
   createScope as unstable_createScope,
-  // enableJSXTransformAPI
-  jsx,
-  jsxs,
-  // TODO: jsxDEV should not be exposed as a name. We might want to move it to a different entry point.
-  jsxDEV,
 };
--- a/packages/react/src/__tests__/ReactElementJSX-test.js
+++ b/packages/react/src/__tests__/ReactElementJSX-test.js
@@ -12,6 +12,8 @@
 let React;
 let ReactDOM;
 let ReactTestUtils;
+let JSXRuntime;
+let JSXDEVRuntime;
 
 // NOTE: We're explicitly not using JSX here. This is intended to test
 // a new React.jsx api which does not have a JSX transformer yet.
@@ -29,6 +31,8 @@ describe('ReactElement.jsx', () => {
     global.Symbol = undefined;
 
     React = require('react');
+    JSXRuntime = require('react/jsx-runtime');
+    JSXDEVRuntime = require('react/jsx-dev-runtime');
     ReactDOM = require('react-dom');
     ReactTestUtils = require('react-dom/test-utils');
   });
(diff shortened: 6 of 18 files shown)


packages/react/src/__tests__/ReactElementJSX-test.internal.js
```diff
@@ -34,6 +35,7 @@ describe('ReactElement.jsx', () => {
     ReactFeatureFlags.warnAboutSpreadingKeyToJSX = true;
```
gaearon: If this test wasn't marked as "internal", it would fail on bundles as it should — because we haven't exposed these entry points on npm. To do that, check out how files in `npm/` folders in different packages work. Also note `"files"` array in `package.json` of each package.

gaearon: https://github.com/facebook/react/pull/18301

gaearon:
You'll also need to add them to `bundles.js` to actually build the bundles. See similar entry points (e.g. `react-dom/server` and its config) for expiration. Hit me up for any questions!

gaearon:
Pls rebase :-) I think this should fail now.

scripts/rollup/bundles.js
```diff
@@ -76,6 +76,44 @@ const bundles = [
     externals: [],
   },
 
+  /******* React Runtime *******/
+  {
+    bundleTypes: [
+      UMD_DEV,
+      UMD_PROD,
+      UMD_PROFILING,
```
gaearon: Did we decide we want to have UMD builds for these? I don't remember.

scripts/rollup/bundles.js
```diff
@@ -76,6 +76,44 @@ const bundles = [
...
+    moduleType: ISOMORPHIC,
+    entry: 'react/jsx-runtime',
+    global: 'JSXRuntime',
+    externals: ['react'],
+  },
+
+  /******* React DEV Runtime *******/
+  {
+    bundleTypes: [
+      UMD_DEV,
```
gaearon: Same q.

scripts/rollup/bundles.js
```diff
@@ -76,6 +76,44 @@ const bundles = [
     externals: [],
   },
 
+  /******* React Runtime *******/
```
gaearon: You probably mean JSX here

scripts/rollup/forks.js
```diff
@@ -34,7 +34,11 @@ const forks = Object.freeze({
       // happens. Other bundles just require('object-assign') anyway.
       return null;
     }
-    if (entry === 'react') {
+    if (
+      entry === 'react' ||
```
gaearon: Let's make it `entry === 'react' || entry.startsWith('react/')` so that we don't forget in case of a rename.

scripts/rollup/forks.js
```diff
@@ -50,7 +54,11 @@ const forks = Object.freeze({
   // Without this fork, importing `shared/ReactSharedInternals` inside
   // the `react` package itself would not work due to a cyclical dependency.
   'shared/ReactSharedInternals': (bundleType, entry, dependencies) => {
-    if (entry === 'react') {
+    if (
+      entry === 'react' ||
```
gaearon: Same.

packages/react/npm/jsx-runtime.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+if (process.env.NODE_ENV === 'production') {
+  module.exports = require('./cjs/react/jsx-runtime.production.min.js');
```
gaearon: This needs to say `react-jsx-runtime`, not `react/jsx-runtime`. Same below.

scripts/rollup/bundles.js
lunaruan: Good point. Just talked to Seb and we should get rid of them because UMD doesn't work with auto import, so there's no point in keeping these

packages/react/npm/jsx-runtime.js
lunaruan: Oops, I missed this one. Good call! :D

packages/react/src/__tests__/ReactElementJSX-test.js
```diff
@@ -29,6 +30,7 @@ describe('ReactElement.jsx', () => {
     global.Symbol = undefined;
 
     React = require('react');
+    jsxRuntime = require('react/jsx-runtime');
```
gaearon: Super nit: JSXRuntime in tests for capitalization consistency? I don't care tho

packages/react/package.json
```diff
@@ -29,6 +31,9 @@
     "loose-envify": "^1.1.0",
     "object-assign": "^4.1.1"
   },
+  "peerDependencies": {
```
gaearon: A package can't be its own peer. This isn't needed. If our scripts complain it's our scripts that should be fixed.

packages/react/src/React.js
```diff
@@ -52,27 +51,19 @@ import {
...
-  jsxWithValidation,
-  jsxWithValidationStatic,
-  jsxWithValidationDynamic,
 } from './ReactElementValidator';
 import createMutableSource from './createMutableSource';
 import ReactSharedInternals from './ReactSharedInternals';
 import createFundamental from 'shared/createFundamentalComponent';
 import createResponder from 'shared/createEventResponder';
 import createScope from 'shared/createScope';
+import {jsx, jsxs, jsxDEV} from './jsx/ReactJSX';
```
gaearon: We’re still re-exporting from the React object. I don’t think we want that. Remove these. Then remove them from all top-level `index*.js` entry points. Only add that import to `index.fb.js` and re-export them there.

gaearon: Otherwise we’ll effectively ship `React.jsx` by accident.

packages/react/index.experimental.js
```diff
@@ -45,8 +45,5 @@ export {
   unstable_withSuspenseConfig,
   // enableBlocksAPI
   block,
-  // enableJSXTransformAPI
-  jsx,
-  jsxs,
-  jsxDEV,
 } from './src/React';
+export {jsx, jsxs, jsxDEV} from './src/jsx/ReactJSX';
```
gaearon: I don't think we want these on `experimental`? This is not a FB build.

packages/react/index.js
```diff
@@ -76,7 +76,5 @@ export {
   DEPRECATED_createResponder,
   unstable_createFundamental,
   unstable_createScope,
-  jsx,
-  jsxs,
-  jsxDEV,
 } from './src/React';
+export {jsx, jsxs, jsxDEV} from './src/jsx/ReactJSX';
```
gaearon: Same, this puts them on React object again for the npm build. I don't think we want them here?

packages/react/src/jsx/ReactJSXElement.js
```diff
@@ -0,0 +1,358 @@
...
+const hasOwnProperty = Object.prototype.hasOwnProperty;
+
+const RESERVED_PROPS = {
+  key: true,
+  ref: true,
+  __self: true,
+  __source: true,
+};
+
+let specialPropKeyWarningShown,
```
gaearon: Nit: we generally leave these as individual let statements

packages/react/src/jsx/ReactJSXElement.js
```diff
@@ -0,0 +1,358 @@
...
+        );
+        didWarnAboutStringRefs[componentName] = true;
+      }
+    }
+  }
+}
+
+function defineKeyPropWarningGetter(props, displayName) {
+  const warnAboutAccessingKey = function() {
+    if (__DEV__) {
```
gaearon: Wrap the whole outer function body in DEV?

packages/react/src/jsx/ReactJSXElement.js
```diff
@@ -0,0 +1,358 @@
...
+  warnAboutAccessingKey.isReactWarning = true;
+  Object.defineProperty(props, 'key', {
+    get: warnAboutAccessingKey,
+    configurable: true,
+  });
+}
+
+function defineRefPropWarningGetter(props, displayName) {
+  const warnAboutAccessingRef = function() {
+    if (__DEV__) {
```
gaearon: Same.

packages/react/src/jsx/ReactJSXElement.js
```diff
@@ -0,0 +1,358 @@
...
+}
+
+/**
+ * Verifies the object is a ReactElement.
+ * See https://reactjs.org/docs/react-api.html#isvalidelement
+ * @param {?object} object
+ * @return {boolean} True if `object` is a ReactElement.
+ * @final
+ */
+export function isValidElement(object) {
```
gaearon: I think this is unused and should be deleted?

packages/react/src/jsx/ReactJSXElement.js
```diff
@@ -0,0 +1,345 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
```
gaearon: Would be nice to have `@flow` on this. Not blocking though because we didn't have it before either.

lunaruan: I can do this in a separate PR after landing the Babel transform!

midudev:
Hi 👋 I thought this change would land with the next version but I see it's not yet available. Do you have any idea when this will be available? Thanks! 🙇

gaearon:
@midudev Can you clarify why you need it?

midudev:
Thanks for the reply. Yes, trying to get working latest babel-preset-react with `automatic` runtime, I'm getting this error:

`Module not found: Can't resolve 'react/jsx-runtime' in '/Users/midudev/Dev/frontend-fc--web-server/src'`

I thought it was related to this PR as installing latest `react` version there's no `jsx-runtime` available to import. Maybe am I doing something wrong?

gaearon:
Right, I just wanted to confirm why you're asking. This indeed hasn't been released in stable yet. You can install `react@experimental` if you want, that one has it.

jetro4u:
I am using the ```0.0.0-experimental-f42431abe``` version of React for testing. I converted a version 17 App and on all the pages I am getting the ```React' must be in scope when using JSX``` error. Meanwhile I already follow the [Introducing the New JSX Transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports) tutorial to implement it. Is it that the the New JSX Transform is not available in the experimental version? Or is there a way to implement it?

PR #18301, opened 2020-03-13
Run ReactElementJSX-test against bundles

gaearon:
This is an important test. We should run it on bundles, not just source. See https://github.com/facebook/react/pull/18299#discussion_r392007055 for the kind of issue it could catch.

I'm removing the internal suffix and moving the only test that actually depends on feature flags by now into its own file. That one stays internal.


packages/react/src/__tests__/ReactElementJSXNewWarnings-test.internal.js
```diff
@@ -0,0 +1,68 @@
...
+  beforeEach(() => {
+    jest.resetModules();
+
+    // Delete the native Symbol if we have one to ensure we test the
+    // unpolyfilled environment.
+    originalSymbol = global.Symbol;
+    global.Symbol = undefined;
+
+    ReactFeatureFlags = require('shared/ReactFeatureFlags');
+    ReactFeatureFlags.warnAboutSpreadingKeyToJSX = true;
```
acdlite: This is already true in the `test-www-variant` run, so instead of moving this to a new file, you can wrap the test in a condition:

```js
if (require('shared/ReactFeatureFlags').warnAboutSpreadingKeyToJSX) {
  // test goes here
}
```

acdlite: See here:
https://github.com/facebook/react/blob/885ed469096d4da7ca038564b323e9b706f5a6d0/packages/shared/forks/ReactFeatureFlags.www-dynamic.js#L10-L18

gaearon: ohh smart

gaearon:
It would be nice to have some way to ensure that a test isn’t completely ignored by all runs because we forgot about giving it a variant.

acdlite:
Yeah that would be nice. Though this is probably only an issue for existing tests, since it'd be weird to write a test that you never run yourself.

Also, if we remove all the hardcoded overrides here: https://github.com/facebook/react/blob/885ed469096d4da7ca038564b323e9b706f5a6d0/packages/shared/forks/ReactFeatureFlags.www-dynamic.js#L25-L32

...and here: https://github.com/facebook/react/blob/885ed469096d4da7ca038564b323e9b706f5a6d0/scripts/jest/setupTests.www.js#L13-L22

then if a flag is shipped to any of our open source or www channels, it will definitely be tested. So I'm not too worried about this.

acdlite:
Put another way, it's not that big a deal if we neglect to test a configuration that doesn't actually exist in one of our release channels. It's not ideal, because there could be hidden bugs that don't get discovered until we turn the flag on somewhere. But it's more important to test against every possible configuration that actually gets shipped.

PR #18302, opened 2020-03-13
Revert "Move MS Windows build to CircleCI"

acdlite:
I'm reverting this because it slowed our CI times from ~6 minutes to ~23 minutes. That's because it runs the build command without concurrency.

I don't remember why the Windows jobs exist. I can see why testing on a Windows machine is useful, but I don't think we should have to build the artifacts on Windows. We could re-use the artifacts from the normal build job.

Let's figure it out then re-land.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,4 @@
-version: 2.1
-
-orbs:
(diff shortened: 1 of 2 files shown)


acdlite:
cc: @wittgenst

PR #18306, opened 2020-03-13
Warn if stringifying an attribute takes too long

### LINKED ISSUE #12209: Warn if stringifying an attribute takes too long

bl00mber:
close #12209


bl00mber:
up

PR #18307, opened 2020-03-13
Warn against state updates from useEffect destroy functions

bvaughn:
Add clearer warning text for the specific case of updating state from within a passive effect's destroy function.

Also fixed a broken conditional that was preventing some of these tests from running after the recent test/variant refactor.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -2818,7 +2818,7 @@ function warnAboutUpdateOnUnmountedFiberInDEV(fiber) {
     // If we are currently flushing passive effects, change the warning text.
     if ((executionContext & PassiveEffectContext) !== NoContext) {
       console.error(
-        "Can't perform a React state update from within a passive effect destroy callback. " +
+        "Can't perform a React state update from within a useEffect cleanup function. " +
           'To fix, move state updates to the useEffect() body in %s.%s',
         tag === ClassComponent
           ? 'the componentWillUnmount method'
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -199,13 +199,14 @@ const {
 
 type ExecutionContext = number;
 
-const NoContext = /*                    */ 0b000000;
-const BatchedContext = /*               */ 0b000001;
-const EventContext = /*                 */ 0b000010;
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2812,15 +2814,28 @@ function warnAboutUpdateOnUnmountedFiberInDEV(fiber) {
...
-        : 'a useEffect cleanup function',
-      getStackByFiberInDevAndProd(fiber),
-    );
+
+    // If we are currently flushing passive effects, change the warning text.
+    if ((executionContext & PassiveEffectContext) !== NoContext) {
+      console.error(
+        "Can't perform a React state update from within a passive effect destroy callback. " +
+          'To fix, move state updates to the useEffect() body in %s.%s',
+        tag === ClassComponent
```
gaearon: Is this even possible?

bvaughn: Yes.

gaearon: I meant is it possible to get this warning with a class component? Referring to this particular line. The message seems confusing in that case.

bvaughn: Oh I see. Nope. Only function components have passive effects.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -199,13 +199,14 @@ const {
...
-const RenderContext = /*                */ 0b010000;
-const CommitContext = /*                */ 0b100000;
+const NoContext = /*                    */ 0b0000000;
+const BatchedContext = /*               */ 0b0000001;
+const EventContext = /*                 */ 0b0000010;
+const DiscreteEventContext = /*         */ 0b0000100;
+const LegacyUnbatchedContext = /*       */ 0b0001000;
+const RenderContext = /*                */ 0b0010000;
+const CommitContext = /*                */ 0b0100000;
+const PassiveEffectContext = /*         */ 0b1000000;
```
acdlite: We shouldn’t be using this for a DEV only warning. We should use a different variable that is only in DEV to reduce the risk that it affects semantic behavior.

bvaughn: Fair enough.

bvaughn: #18309

PR #18309, opened 2020-03-14
Replace passive effect context with boolean

bvaughn:
Follow up to #18307

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -199,14 +199,13 @@ const {
 
 type ExecutionContext = number;
 
-const NoContext = /*                    */ 0b0000000;
-const BatchedContext = /*               */ 0b0000001;
-const EventContext = /*                 */ 0b0000010;
-const DiscreteEventContext = /*         */ 0b0000100;
-const LegacyUnbatchedContext = /*       */ 0b0001000;
-const RenderContext = /*                */ 0b0010000;
-const CommitContext = /*                */ 0b0100000;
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -299,6 +298,10 @@ let spawnedWorkDuringRender: null | Array<ExpirationTime> = null;
 // receive the same expiration time. Otherwise we get tearing.
 let currentEventTime: ExpirationTime = NoWork;
 
+// Dev only flag that tracks if passive effects are currently being flushed.
+// We warn about state updates for unmounted components differently in this case.
+let isFlushingPassiveEffects = false;
```
bvaughn: Preference on the name? (I have none.)

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2448,6 +2454,10 @@ function flushPassiveEffectsImpl() {
   nestedPassiveUpdateCount =
     rootWithPendingPassiveEffects === null ? 0 : nestedPassiveUpdateCount + 1;
 
+  if (__DEV__) {
+    isFlushingPassiveEffects = false;
```
gaearon: Do we have a guarantee code in between wouldn't throw even if user code throws (unless there's a bug in React)?

bvaughn: User code is try/catched.

Though I wonder if I should have moved this before `flushSyncCallbackQueue` since that's where we reset prev exceution context.

Issue #18101, opened 2020-02-21
Arbitrary log lines may appear in between an error and the subsequent "The above error…" message


sophiebits:
React version: 16.12

If one component throws, and one of its siblings logs something else to the console, that sibling's logs appear in between the throwing component and the extra React information about that throw.

This is hard to explain in words but easy to explain in code and pictures, so look:

Repro: https://codesandbox.io/s/boring-firefly-bu79e

![image](https://user-images.githubusercontent.com/6820/75078940-d822d380-54bb-11ea-99d2-c77fbc0c9a5b.png)

Instead, I would expect "The above error…" to be right below the uncaught window-level exception we raise. (Maybe this isn't possible without diverging dev vs. prod behavior? Not sure but thought I'd file…)

sophiebits:
Stillll relevant.

rachelnabors:
Can confirm it happens in Firefox as well

sophiebits:
Still happens on next (18.3.0-next-4fcc9184a-20230217):  https://codesandbox.io/s/pedantic-dubinsky-x6une4

buoyantair:
I'd like to give this one a try :D

sophiebits:
If you (or anyone) opens a PR for this feel free to tag me! It may be a tricky one though.

sophiebits:
It seems this no longer happens with latest error handling behavior.

PR #18315, opened 2020-03-15
Include component stack to invariants

### LINKED ISSUE #11619: Include component stack in invariants
IIRC we didn't include it because it was DEV-only. But it's not anymore.
Maybe let's start including it?

Errors are often more prominent than warnings, and it would be great to have this info in both.

bl00mber:
close #11619


sophiebits:
I was under the impression that we now tack on a component stack to all errors raised during rendering, in a second message that starts "The above error occurred in the <XYZ /> component:". Can you post before and after screenshots of the console?

bl00mber:
Yeah this is probably not relevant

Screenshot from 2020-03-20

sophiebits:
It is interesting that in your screenshot the "The above error" message appears _before_ the actual error. Perhaps related to #18101.

However, given that the stack trace does show up it probably does make sense to leave the behavior as-is. I'll close the original issue too. Sorry to make you spend time on something that didn't end up getting merged, but thank you for the contribution!

PR #18316, opened 2020-03-15
improve error message for cross-functional component updates

JoviDeCroock:
## Summary

This improves the error message used for "cross-functional component updates"


gaearon:
Can we use `workInProgress` which is already available locally rather than read current owner?

JoviDeCroock:
@gaearon Thanks for the advice, I had to guard it for Flow tho

PR #18329, opened 2020-03-17
Don't support older isomorphic React with newer renderers

gaearon:
I think we've already determined this is not gonna work as well as we hoped. Let's just stop supporting it. It's not a breaking change for newer versions of the renderers. We're just removing the guarantee implied by the peer version which we broke anyway several times before.

My actual goal is to remove the patching junk from the builds (see the end of the diff). Especially now that we're adding JSXRuntime and it's leaking there too.

### DIFF
--- a/packages/shared/ReactLazyComponent.js
+++ b/packages/shared/ReactLazyComponent.js
@@ -38,11 +38,7 @@ export function initializeLazyComponentType(
   lazyComponent: LazyComponent<any>,
 ): void {
   if (lazyComponent._status === Uninitialized) {
-    let ctor = lazyComponent._result;
-    if (!ctor) {
-      // TODO: Remove this later. THis only exists in case you use an older "react" package.
-      ctor = ((lazyComponent: any)._ctor: typeof ctor);
-    }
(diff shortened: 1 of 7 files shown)


gaearon:
>The case that is still tricky is that we support older "react" in React Native implicitly.

Huh?

packages/shared/ReactLazyComponent.js
```diff
@@ -38,11 +38,7 @@ export function initializeLazyComponentType(
   lazyComponent: LazyComponent<any>,
 ): void {
   if (lazyComponent._status === Uninitialized) {
-    let ctor = lazyComponent._result;
-    if (!ctor) {
-      // TODO: Remove this later. THis only exists in case you use an older "react" package.
-      ctor = ((lazyComponent: any)._ctor: typeof ctor);
-    }
+    const ctor = lazyComponent._result;
```
sebmarkbage: Keep in mind that until we release this in isomorphic, it will break RN OSS.

gaearon: Makes sense. I won't do another full sync until after we've cut a stable version.

Issue #18178, opened 2020-02-28
Bug: too hard to fix "Cannot update a component from inside the function body of a different component."


finom:
React version:

16.13.0

## Steps To Reproduce

1. Build a time machine.
2. Go to the year 2017.
3. Build a huge application of 10K lines of code.
4. Get 80 (!) dependencies at package.json file including ones that become no longer maintained.
5. Update React to the latest version at February 27, 2020.
6. Get tons of errors that you don't know how to fix.
7. Tell your client that fixes are going to take unknown time and it's going to cost $$$ + days or weeks of investigation or we're going to get stuck with the outdated version of React and related libraries forever which will cost more $$$ but later.

Being serious, the business I work for isn't interested on that at all. Obviously I'd never made it happen to get such warnings to appear if I'd get them earlier. Currently that's impossibly hard to make the errors to be fixed because I get them at many different cases and with a huge stack trace. I tried to fix at least one of the appearing errors and it already took a lot of time. I tried to debug some of used libraries but got no luck. 

Just one example:

![image](https://user-images.githubusercontent.com/1082083/75559100-8fcf5c80-5a4b-11ea-8173-4f0a62cc5de3.png)

There we can notice the use of an outdated react-router, an outdated redux-connect (which I had to put to the project source to fix errors of outdated `componentWillReceiveProps` method), some HOCs created by recompose etc. It isn't just a simple virtual DOM tree where I can walk thru components developed by me and search by `setState` string to fix the bug, that's way more complicated than that.

Please make an "UNSAFE" option to disable this error or provide a simpler way to find where the error is thrown 🙏

sebmarkbage:
I wish we had added this warning earlier. I'm sorry we didn't. It was an oversight with the introduction of Hooks. I believe this must be caused by newer code that uses Hooks, because there was already the same warning for classes much earlier so any earlier code would've already seen this warning.

Note that this will likely start hard failing in future versions of React. Intentional or otherwise (we've had a lot of bugs with this pattern). So regardless, you might end up getting stuck on an old version. If it's not possible to fix it, I'd recommend pinning to an older version of React.

However, that said, we want to help you and make it as easy as possible to fix these, including seeing if we can get help from library authors.

If you expand the little `>` arrow in the Chrome console you should also see an additional Stack trace (in addition to the Component stack in your screen shot). Can you post that too? That should show you the exact callsite that is causing a side-effect in render.

RodolfoSilva:
With me this appears when i use formik 

[...]

PR #18330, opened 2020-03-17
Don't fire the render phase update warning for class lifecycles

gaearon:
Fixes https://github.com/facebook/react/issues/18178#issuecomment-600132920.

In 16.13, we added a new warning: `Cannot update a component from inside a body of a function component`. But as we see from reports like https://github.com/facebook/react/issues/18178#issuecomment-600106450, it fired for updates **from inside of class lifecycles too**. While arguably these *are* real issues (they happen in render phase) it's pretty noisy for legacy code.

In this PR, I do the following:

- Change the wording (`from inside a body of a function component` -> `while rendering`). This more accurately describes when it would fire.
- Change the deduping originally added in https://github.com/facebook/react/pull/18316 to be more aggressive. In particular, I suggest to not fire more than one warning per callsite. This is because the actual target components tend to be disconnected and mostly unrelated due to abstractions like Redux. Let's focus on the part that *fires* the updates.
- Limit the warning to **function component bodies and class `render` lifecycles**. Everything else will get muted. This is consistent with how our existing class "setState from another component's render" warning works.
  - As a possible follow-up, we could make both of these fire for other render-phase lifecycles like `getDerivedStateFromProps`, `shouldComponentUpdate`, and `constructor`. Maybe in strict mode only. However, we should keep them muted for `componentWill*` lifecycles including render phase ones.

Here is a table of the changes:

Screenshot 2020-03-17 at 21 10 07

**Red** means new behavior, **blue** means revert to 16.12 behavior.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -1363,6 +1363,7 @@ function mountIndeterminateComponent(
       ReactStrictModeWarnings.recordLegacyContextWarning(workInProgress, null);
     }
 
+    setIsRendering(true);
     ReactCurrentOwner.current = workInProgress;
     value = renderWithHooks(
       null,
@@ -1372,6 +1373,7 @@ function mountIndeterminateComponent(
       context,
       renderExpirationTime,
     );
+    setIsRendering(false);
   } else {
     value = renderWithHooks(
       null,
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1372,6 +1373,7 @@ function mountIndeterminateComponent(
       context,
       renderExpirationTime,
     );
+    setIsRendering(false);
```
acdlite: If the component throws, `setIsRendering` won't be reset. This needs to be reset in the the work loop. Here: https://github.com/facebook/react/blob/8311cb5d24fd1648ddffb659bf9b3317f5f885d0/packages/react-reconciler/src/ReactFiberWorkLoop.js#L1226-L1229

acdlite: Never mind, gets reset by `resetCurrentDebugFiberInDEV

PR #18331, opened 2020-03-17
Fix UMD Builds (ReactSharedInternals)

lunaruan:
`ReactCurrentBatchConfig.suspense` does not exist in `ReactSharedInternals.umd`. This PR adds it.


gaearon:
>ReactCurrentBatchConfig.suspense does not exist in legacy renderer for fixturees

Why?

sebmarkbage:
My understanding is that ReactCurrentBatchConfig is exposed on shared internals in all flag configurations. So why would it be undefined?

gaearon:
(Note this PR doesn't "fix fixtures" anymore, it's more like it fixes UMD builds?)

lunaruan:
> Scary stuff. Let's come up with some plan to prevent this in the future? Like maybe a shared flowtype.

This makes sense. I'll add it to my to do list to add a `ReactSharedInternals` type :)

PR #18332, opened 2020-03-17
Refactor React.Children to reduce indirection

gaearon:
This has been bugging me for a while. I know we don't plan to change their behavior because it's a legacy pattern anyway (although we don't have an alternative either). But it's a bit awkward to see this indirection in the bundle. It's also very confusing to read.

The original code was written assuming this is perf critical and we must avoid closures. I don't think it is. This pattern isn't super common and when you have it, you probably don't want to iterate over thousands of these anyway. In either case, my last implementation has minimal use of closures too. I removed pooling too which is a bit over-engineered for this use case and might actually hurt perf. Unlike with events, it's not observable.

I re-implemented count, forEach, and toArray in terms of map. I know they have slightly different semantics so a naïve implementation wouldn't work but I think I got it right (and when I did it wrong, tests failed). The original rationale to implement them separately was to reuse a lower level primitive that doesn't require any extra allocations. But since both `forEach` and `count` have other quirks (like counting holes) I don't expect them to actually be used that much in critical paths anymore. So I think it's okay `count` allocates an array now and later throws it away.

Each commit is fairly mechanical. At the end, I have a single recursive function as a core of map, and other functions expressed through map. Instead of a tangled web of indirections. Read individual commits. Although I guess reading the whole thing also works because you can follow it now.

### DIFF
--- a/packages/react/src/ReactChildren.js
+++ b/packages/react/src/ReactChildren.js
@@ -51,13 +51,53 @@ function escapeUserProvidedKey(text) {
 }
 
 /**
- * @param {?*} children Children tree container.
- * @param {!string} nameSoFar Name of the key path so far.
- * @param {!function} callback Callback to invoke with each child found.
- * process.
- * @return {!number} The number of children in this subtree.
+ * Generate a key string that identifies a component within a set.
+ *
+ * @param {*} component A component that could contain a manual key.
+ * @param {number} index Index that is used if a manual key is not provided.
+ * @return {string}
  */
-function traverseAllChildren(children, nameSoFar, callback) {
+function getComponentKey(component, index) {
+  // Do some typechecking here since we call this blindly. We want to ensure
+  // that we don't block potential future ES APIs.
+  if (
+    typeof component === 'object' &&
+    component !== null &&
+    component.key != null
+  ) {
+    // Explicit key
+    return escape(component.key);
+  }
+  // Implicit key determined by the index in the set
+  return index.toString(36);
+}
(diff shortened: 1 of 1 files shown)


gaearon:
I'm pretty sure I'm right (c)

I can write a small one-off generative test thing to verify the results match between implementations. #covid #bored

gaearon:
I know it's subtle though. E.g. people likely depend even on the nested map unwrapping call order. I think I kept all of that but I agree it's worth testing more closely.

sophiebits:
wdyt of just dropping the pooling and closure avoidance but keeping traverseAllChildren with a different callback for map vs forEach vs count? I think that gets you 90% of the code simplification.

sophiebits:
and is nicer in some ways – eg: count doesn't need to make an array.

gaearon:
I guess that would be better than what we have today, but I'd still prioritize golfing code size over runtime perf of these helpers.

gaearon:
Worth noting traverse perf was super important when it was shared by the reconciler code. It's not anymore though.

gaearon:
Let's give it a try. The only discrepancy I found is that `React.Children.forEach(null)` no longer returns `null`. I didn't fix it because there's no logical reason why it should (this seems accidental), and relying on `forEach` return value is extremely unlikely (this is the only case where it's not `undefined`). I couldn't find any internal callsites doing it either.

gaearon:
https://github.com/facebook/react/pull/18380

PR #18333, opened 2020-03-18
Don't use EventListener Fork in Modern WWW Builds

sebmarkbage:
If [D20382065](https://our.internmc.facebook.com/intern/diff/D20382065/) successfully rolls out (has some issues still). Then we don't need this fork anymore.

We should start by at least removing it from Modern builds.

To do this I needed to fix the unsubscribing stuff mentioned in https://github.com/facebook/react/pull/18270#discussion_r394058260. Flare is now used by both the FB fork and without it so we should just compile to one or the other. I do that by always returning an unsubscribe listener and always passing that to a removeEventListener function. The indirections can then be compiled out.

### DIFF
--- a/packages/react-dom/src/events/ReactDOMEventListener.js
+++ b/packages/react-dom/src/events/ReactDOMEventListener.js
@@ -49,6 +49,7 @@ import {
   addEventCaptureListener,
   addEventCaptureListenerWithPassiveFlag,
   addEventBubbleListenerWithPassiveFlag,
+  removeEventListener,
 } from './EventListener';
 import getEventTarget from './getEventTarget';
 import {getClosestInstanceFromNode} from '../client/ReactDOMComponentTree';
@@ -108,7 +109,6 @@ export function addResponderEventSystemEvent(
   } else {
     eventFlags |= IS_ACTIVE;
   }
-  let fbListener;
   // Check if interactive and wrap in discreteUpdates
   const listener = dispatchEvent.bind(
     null,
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -259,20 +250,9 @@ export function removeTrappedEventListener(
...
   passive: void | boolean,
-) {
-  if (listener.remove != null) {
-    listener.remove();
-  } else {
-    const rawEventName = getRawEventName(topLevelType);
-    if (passiveBrowserEventsSupported) {
-      targetContainer.removeEventListener(rawEventName, listener, {
-        capture,
-        passive,
```
sebmarkbage: The passive flag [is not needed when removing](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#Matching_event_listeners_for_removal). Only the capture flag needs to be specified.

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -203,28 +199,26 @@ export function addTrappedEventListener(
       try {
         return originalListener.apply(this, p);
       } finally {
-        if (fbListener) {
-          fbListener.remove();
-        } else {
-          validTargetContainer.removeEventListener(
-            ((rawEventName: any): string),
-            (listener: any),
-          );
```
sebmarkbage: This wasn't specifying the capture flag but I fixed that.

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -259,20 +250,9 @@ export function removeTrappedEventListener(
   capture: boolean,
   listener: any => void,
   passive: void | boolean,
```
trueadm: This argument is not used anymore

trueadm:
I'm merging this as it feeds into https://github.com/facebook/react/pull/18336. :)

PR #18334, opened 2020-03-18
[Flight] Wire up bundler configs

sebmarkbage:
This allows different Flight server and clients to have different configs depending on bundler to serialize and resolve modules.

I'll use these in a follow up for Blocks.

### DIFF
--- a/packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
+++ b/packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+export type ModuleMetaData = {
+  id: string,
+  chunks: Array<string>,
+};
+
+type Thenable = {
+  then(resolve: () => mixed, reject: (mixed) => mixed): mixed,
+  ...
+};
+
+// The chunk cache contains all the chunks we've preloaded so far.
+// If they're still pending they're a thenable. This map also exists
+// in Webpack but unfortunately it's not exposed so we have to
+// replicate it in user space. null means that it has already loaded.
+const chunkCache: Map<string, null | Thenable> = new Map();
+
+// Returning null means that all dependencies are fulfilled and we
+// can synchronously require the module now. A thenable is returned
+// that when resolved, means we can try again.
+export function preloadModule(moduleData: ModuleMetaData): null | Thenable {
+  let moduleEntry = require.cache[moduleData.id];
+  if (moduleEntry) {
+    // Fast exit if this module has already been loaded.
+    return null;
+  }
+  let chunks = moduleData.chunks;
+  let anyRemainingThenable = null;
+  for (let i = 0; i < chunks.length; i++) {
+    let chunkId = chunks[i];
+    let entry = chunkCache.get(chunkId);
+    if (entry === undefined) {
+      anyRemainingThenable = __webpack_chunk_load__(chunkId);
+      chunkCache.set(chunkId, anyRemainingThenable);
+      anyRemainingThenable.then(chunkCache.set.bind(chunkCache, chunkId, null));
+    } else if (entry !== null) {
+      anyRemainingThenable = entry;
+    }
+  }
+  return anyRemainingThenable;
+}
+
+export function requireModule<T>(moduleData: ModuleMetaData): T {
+  return __webpack_require__(moduleData.id).default;
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -157,6 +157,13 @@ module.exports = {
         nativeFabricUIManager: true,
       },
     },
+    {
+      files: ['packages/react-flight-dom-webpack/**/*.js'],
+      globals: {
+        '__webpack_chunk_load__': true,
+        '__webpack_require__': true,
+      },
+    },
   ],
 
   globals: {
--- a/packages/react-client/src/ReactFlightClient.js
+++ b/packages/react-client/src/ReactFlightClient.js
@@ -9,6 +9,13 @@
 
 import {REACT_ELEMENT_TYPE} from 'shared/ReactSymbols';
 
+// import type {ModuleMetaData} from './ReactFlightClientHostConfig';
+
+// import {
+//   preloadModule,
+//   requireModule,
+// } from './ReactFlightClientHostConfig';
+
 export type ReactModelRoot<T> = {|
   model: T,
 |};
--- a/packages/react-client/src/ReactFlightClientHostConfigNoStream.js
+++ b/packages/react-client/src/ReactFlightClientHostConfigNoStream.js
@@ -1,0 +1,30 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+export type StringDecoder = void;
+
+export const supportsBinaryStreams = false;
+
+export function createStringDecoder(): void {
+  throw new Error('Should never be called');
+}
+
+export function readPartialStringChunk(
+  decoder: StringDecoder,
+  buffer: Uint8Array,
+): string {
+  throw new Error('Should never be called');
+}
+
+export function readFinalStringChunk(
+  decoder: StringDecoder,
+  buffer: Uint8Array,
+): string {
+  throw new Error('Should never be called');
+}
--- a/packages/react-client/src/forks/ReactFlightClientHostConfig.custom.js
+++ b/packages/react-client/src/forks/ReactFlightClientHostConfig.custom.js
@@ -24,6 +24,11 @@
 // really an argument to a top-level wrapping function.
 
 declare var $$$hostConfig: any;
+
+export opaque type ModuleMetaData = mixed; // eslint-disable-line no-undef
+export const preloadModule = $$$hostConfig.preloadModule;
+export const requireModule = $$$hostConfig.requireModule;
+
 export opaque type Source = mixed; // eslint-disable-line no-undef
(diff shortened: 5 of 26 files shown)


packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
```diff
@@ -0,0 +1,48 @@
...
+};
+
+export function requireModuleIfAvailable<T>(
+  moduleData: ModuleMetaData,
+): T | null {
+  let entry = require.cache[moduleData.id];
+  if (entry) {
+    return entry.exports.default;
+  }
+  // Ideally Webpack would let us inspect that all chunks have loaded and
```
gaearon: I chatted to @sokra about this and he says they can make it return the same Promise every time so we can put it in a WeakMap. Would this help us? My understanding is that this wouldn't be sufficient. If we just needed to track chunk completion status we could do it ourselves:

```js
__webpack_chunk_load__(chunk).then(() => { status[chunk] = true })
```

But that doesn't solve the problem of checking whether a particular has *already* loaded (either as the initial script on the page or as a part of some other request outside of Flight). I think that's what you want to check? In that case, I don't see a different way than exposing `installedChunks`. Maybe we can write a plugin that puts it on a global.

sebmarkbage: Avoiding every possible async point is not a goal here. Just to minimize them. As discussed, I think the key here is to avoid the waterfall of requests that happens when each React component asks for them one by one. However, in this new approach we have all the info about all the chunks we're at high likelihood going to need soon without executing the React components. That's kind of the point because we should eagerly prefetch them regardless.

So I'm going to refactor this to eagerly load all the chunks and then keep track on our side whether it has already loaded. That way even if one needs to suspend, in that gap, we'll be able to collect a lot of newly resolved chunks at once and then continue.

sebmarkbage: The key difference between this and the dynamic imports API is that loading the chunks doesn't immediately execute the module bodies. If we did, then executing these eagerly wouldn't be palatable.

So the deep Webpack integration is still useful.

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
```diff
@@ -0,0 +1,48 @@
...
+  // doesn't expose this information without creating a Promise first.
+  // This causes an unfortunate exponential many-pass render since each
+  // level in the tree will restart to initialize that module.
+  // This will lead to unacceptable performance in React and will need
+  // changes to Webpack.
+  return null;
+}
+
+export function waitForModule(moduleData: ModuleMetaData): Thenable {
+  let chunks = moduleData.chunks[0]; // First candidate
```
sebmarkbage: cc @gaearon

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
sokra: Note that chunk loading can be more than just loading javascript:
* adding stylesheets when using the mini-css-extract-webpack-plugin
* compiling wasm when using the sync wasm approach (deprecated in webpack 5)
* prefetching and preloading more chunks
* custom plugins would add additional logic here (e. g. I wrote a localization plugin which fetches localization data during chunk loading).

So direct `installedChunks` access would break when using more than JS.

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
```diff
@@ -0,0 +1,53 @@
...
+type Thenable = {
+  then(resolve: () => mixed, reject: (mixed) => mixed): mixed,
+  ...
+};
+
+// The chunk cache contains all the chunks we've preloaded so far.
+// If they're still pending they're a thenable. This map also exists
+// in Webpack but unfortunately it's not exposed so we have to
+// replicate it in user space. null means that it has already loaded.
+const chunkCache: Map<string, null | Thenable> = new Map();
```
sebmarkbage: @sokra In this approach, I'll always call `__webpack_chunk_load__` eagerly so there's always an opportunity for it to kick off async requests as a result. Always getting the same Promise back would be helpful since I wouldn't need to cache it. However, I also need to know if they've all been resolved so I know it's safe to call require.

This means I have to replicate this information in a user space map anyway.

It would be nice if there was a way to call `__webpack_chunk_load__`, so that there's something that spawns async work to kick off but synchronously know if it's already safe to call `__webpack_require__`. It's fine that we're forced to call `__webpack_chunk_load__` and even that it's always async, but if we have already done so in the past, it would be nice to query the webpack cache rather than our own.

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
```diff
@@ -0,0 +1,53 @@
...
+// that when resolved, means we can try again.
+export function preloadModule(moduleData: ModuleMetaData): null | Thenable {
+  let moduleEntry = require.cache[moduleData.id];
+  if (moduleEntry) {
+    // Fast exit if this module has already been loaded.
+    return null;
+  }
+  let chunks = moduleData.chunks;
+  let anyRemainingThenable = null;
+  for (let i = 0; i < chunks.length; i++) {
```
sokra: This code snippet doesn't work for more than 1 chunks, as `anyRemainingThenable` is overwritten in each iteration.

You probably want to use `Promise.all` here?

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
```diff
@@ -0,0 +1,53 @@
...
+// If they're still pending they're a thenable. This map also exists
+// in Webpack but unfortunately it's not exposed so we have to
+// replicate it in user space. null means that it has already loaded.
+const chunkCache: Map<string, null | Thenable> = new Map();
+
+// Returning null means that all dependencies are fulfilled and we
+// can synchronously require the module now. A thenable is returned
+// that when resolved, means we can try again.
+export function preloadModule(moduleData: ModuleMetaData): null | Thenable {
+  let moduleEntry = require.cache[moduleData.id];
```
sokra: Technically using `require.cache` and ESM is kind of illegal. It works in webpack in "automatic" JS mode (kind of mixed CJS and ESM), but `require.cache` is not available in strict ESM mode. At least it works for webpack <= 5.

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
```diff
@@ -0,0 +1,53 @@
...
+      anyRemainingThenable.then(chunkCache.set.bind(chunkCache, chunkId, null));
+    } else if (entry !== null) {
+      anyRemainingThenable = entry;
+    }
+  }
+  return anyRemainingThenable;
+}
+
+export function requireModule<T>(moduleData: ModuleMetaData): T {
+  return __webpack_require__(moduleData.id).default;
```
sokra: In production mode webpack may mangles exports names if certain conditions are met. This means `.default` might be `.n` or something else.

It depends on how the module is used and how you get the id. You could opt-out from these optimization, but the better idea is probably to add the mangled export name to the `ModuleMetaData`. I guess you extract this meta data with a webpack plugin. Here you can also access the final name.

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+export type ModuleMetaData = {
+  id: string,
+  chunks: Array<string>,
```
sokra:
```suggestion
  id: string | number,
  chunks: Array<string | number>,
```

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
sokra:
```suggestion
const chunkCache: Map<string | number, null | Thenable> = new Map();
```

I consider adding an extra argument to `__webpack_chunk_load__` which make it return `null` instead of a resolved Promise when the chunk is already loaded.

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
sebmarkbage: This will be recalled each time a new chunk resolves which will then get the next one and so on. However in practice it most likely won’t get as far as even retrying the second time until everything has been resolved.

So in practice we get one promise, then wait for it, schedule work to try again, in the meantime all of them resolve and then we try again just to confirm they all resolved. If not we’ll do the whole thing again.

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
sebmarkbage: Thanks! Yea that sounds like a good idea.

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
sebmarkbage: Have you considered exposing a Webpack specific helper to do something similar to this?

In this case I could just get rid of it since it’s just an optimization to avoid having to check many maps.

It might be nice to know for scheduling purposes though. E.g. we try to be very particular about when we call `__webpack_require__` since that can kick off a lot of synchronous work. So it’s nice to be able to know if the call will be free or hundreds of ms.

packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js
sebmarkbage: That would be great! Thank you.

PR #18336, opened 2020-03-18
Modern Event System: refactor legacy FB support logic

trueadm:
This is a pretty big refactor of the FB event support logic in the modern event system. Previously, the logic worked for only Primer, but on more internal cases occuring, it seems we should probably delegated all "click" events to after the internal tooling runs, which fixes all the known issues with the legacy FB event systems. To ensure this works better, I've also made all the modern event system tests run with and without the internal FB flag so we catch more cases with this change.

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -67,6 +67,7 @@ import {
   TOP_RATE_CHANGE,
   TOP_PROGRESS,
   TOP_PLAYING,
+  TOP_CLICK,
 } from './DOMTopLevelEventTypes';
 import {
   getClosestInstanceFromNode,
@@ -229,18 +230,18 @@ function willDeferLaterForLegacyFBSupport(
   topLevelType: DOMTopLevelEventType,
   targetContainer: EventTarget,
 ): boolean {
-  if ((topLevelType: any) !== 'click') {
+  if (topLevelType !== TOP_CLICK) {
     return false;
   }
   // We defer all click events with legacy FB support mode on.
   // This means we add a one time event listener to trigger
   // after the FB delegated listeners fire.
-  const legacyFBSupport = true;
+  const isDeferredListenerForLegacyFBSupport = true;
   addTrappedEventListener(
     targetContainer,
     topLevelType,
     false,
-    legacyFBSupport,
+    isDeferredListenerForLegacyFBSupport,
   );
   return true;
 }
@@ -290,7 +291,12 @@ export function dispatchEventForPluginEventSystem(
     // time event listener so we can defer the event.
     if (
       enableLegacyFBSupport &&
+      // We do not want to defer if the event system has already been
+      // set to LEGACY_FB_SUPPORT. LEGACY_FB_SUPPORT only gets set when
+      // we call willDeferLaterForLegacyFBSupport, thus not bailing out
+      // will result in endless cycles like an infinite loop.
       (eventSystemFlags & LEGACY_FB_SUPPORT) === 0 &&
+      // We also don't want to defer during event replaying.
       (eventSystemFlags & IS_REPLAYED) === 0 &&
       willDeferLaterForLegacyFBSupport(topLevelType, targetContainer)
     ) {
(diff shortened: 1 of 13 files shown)


packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -159,11 +160,15 @@ export function addTrappedEventListener(
   if (passive === true && !passiveBrowserEventsSupported) {
     passive = false;
   }
+  const eventSystemFlags =
+    enableLegacyFBSupport && legacyFBSupport
```
threepointone: this is a bit confusing because you'd imagine legacyFBSupport would always be true if enableLegacyFBSupport is true. I think it means this is a badly named argument. Something to consider for a future PR?

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -177,14 +182,17 @@ export function addTrappedEventListener(
     targetContainer = document;
   }
 
-  const validTargetContainer = ((targetContainer: any): EventTarget);
+  const targetContainerToUse =
```
threepointone: I don't think this rename is meaningfully different, but ok. Maybe just reassign on targetContainer?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -225,42 +225,24 @@ export function listenToEvent(
...
-
-function willDeferLaterForFBLegacyPrimer(nativeEvent: any): boolean {
-  let node = nativeEvent.target;
-  const type = nativeEvent.type;
-  if (type !== 'click') {
+function willDeferLaterForLegacyFBSupport(
+  topLevelType: DOMTopLevelEventType,
+  targetContainer: EventTarget,
+): boolean {
+  if ((topLevelType: any) !== 'click') {
```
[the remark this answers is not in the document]
trueadm: Yeah, I'll address this better.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -303,12 +285,14 @@ export function dispatchEventForPluginEventSystem(
...
     }
-    // If we detect the FB legacy primer system, we
+    // If we are using the legacy FB support flag, we
     // defer the event to the null with a one
     // time event listener so we can defer the event.
     if (
-      enableLegacyFBPrimerSupport &&
-      willDeferLaterForFBLegacyPrimer(nativeEvent)
+      enableLegacyFBSupport &&
+      (eventSystemFlags & LEGACY_FB_SUPPORT) === 0 &&
```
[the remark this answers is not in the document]
trueadm: We want the opposite here, we want to not invoke the logic if the incoming event has been set with , otherwise we'll be endlessly making event listeners and get stuck in a cycle.

packages/react-dom/src/events/ReactDOMEventListener.js
trueadm: I'll rename as part of this PR, good point though.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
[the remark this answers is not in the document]
threepointone: Oh ok I get it. I think the structure of the comment confused me - "if we are using legacy, then we defer event", and I didn't notice  willDeferLaterForLegacyFBSupport is a boolean returning function, I though it was inside the block ha.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -225,42 +226,24 @@ export function listenToEvent(
...
-      );
-      return true;
-    }
-    node = node.parentNode;
-  }
-  return false;
+  // We defer all click events with legacy FB support mode on.
+  // This means we add a one time event listener to trigger
+  // after the FB delegated listeners fire.
+  const legacyFBSupport = true;
```
threepointone:
```suggestion
  const isDeferredListenerForLegacyFBSupport = true;
```

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -225,42 +226,24 @@ export function listenToEvent(
...
+  addTrappedEventListener(
+    targetContainer,
+    topLevelType,
+    false,
+    legacyFBSupport,
```
threepointone:
```suggestion
    isDeferredListenerForLegacyFBSupport,
```

PR #18341, opened 2020-03-18
[eslint-plugin-react-hooks] Disallow hooks in class components

ianobermiller:
## Summary

Per discussion at Facebook, we think hooks have reached a tipping point where it is more valuable to lint against potential hooks in classes than to worry about false positives.

## Test Plan

```
# run from repo root
yarn test --watch RuleOfHooks
```

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
@@ -233,45 +233,6 @@ const tests = {
       }
     `,
     `
-      // Currently valid because we found this to be a common pattern
-      // for feature flag checks in existing components.
-      // We *could* make it invalid but that produces quite a few false positives.
-      // Why does it make sense to ignore it? Firstly, because using
-      // hooks in a class would cause a runtime error anyway.
-      // But why don't we care about the same kind of false positive in a functional
-      // component? Because even if it was a false positive, it would be confusing
-      // anyway. So it might make sense to rename a feature flag check in that case.
-      class ClassComponentWithFeatureFlag extends React.Component {
-        render() {
-          if (foo) {
-            useFeatureFlag();
-          }
-        }
-      }
-    `,
-    `
-      // Currently valid because we don't check for hooks in classes.
-      // See ClassComponentWithFeatureFlag for rationale.
-      // We *could* make it invalid if we don't regress that false positive.
-      class ClassComponentWithHook extends React.Component {
-        render() {
-          React.useState();
-        }
-      }
-    `,
-    `
-      // Currently valid.
-      // These are variations capturing the current heuristic--
-      // we only allow hooks in PascalCase, useFoo functions,
-      // or classes (due to common false positives and because they error anyway).
-      // We *could* make some of these invalid.
-      // They probably don't matter much.
-      (class {useHook = () => { useState(); }});
-      (class {useHook() { useState(); }});
-      (class {h = () => { useState(); }});
-      (class {i() { useState(); }});
-    `,
-    `
       // Valid because they're not matching use[A-Z].
       fooState();
       use();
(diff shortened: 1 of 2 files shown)


ianobermiller:
> I wonder if it's worth limiting it to React.(Pure?)Component subclasses and/or classes with a render method.

Running this on our internal codebase at Facebook, I caught a bug in some code that was using a hook in a non-React class. I think it may be safe to say if you are using this lint rule, you should stick to the convention that all `use*` functions are hooks.

gaearon:
ok let's do it

blixt:
We have multiple cases in our codebase which look something like

```typescript
class StatefulStuff {
    // ...
    // A bunch of stateful functionality that can change over time.
    // ...

    // A helper hook that can allow a functional component to re-render when state changes.
    useTheThing(key: string) {
        const [thing, setThing] = useState(this.map[key])
        useEffect(() => { /* Register an event handler, etc. */ })
        return thing
    }
}
```

What can be done to make this ESLint rule stop thinking that this 99% non-React code is a React component? I'm happy to help implement the detection provided some pointers, because unfortunately as it stands right now we're given 2 pretty bad choices for our code base:

1. Ignore the ESLint rule in every file defining classes like the one above (and miss legitimate problems in the use of hooks)
2. Extract the hooks outside the class, but pass the class instance as an argument to the hook (this is bad because it requires us to expose private fields that shouldn't be exposed so that the external hook can use them)

If there are other workarounds I'm all ears.

a-c-sreedhar-reddy:
Are hooks allowed in class based components?

juliusv:
I'm running into the same issue as @blixt above. I have a stateful non-React class for API accesses that I pass around the app, and I define a `useFetchAPI()` hook function inside of it that I only call from function components. It would be nice to not require adding explicit comments to disable ESLint for those lines.

kachkaev:
You can avoid linting errors the following way:

```ts

// A helper hook that can allow a functional component to re-render when state changes.
const useTheThing = (key: string) => {
  const [thing, setThing] = useState(this.map[key])
  useEffect(() => { /* Register an event handler, etc. */ })
  return thing
}

class StatefulStuff {
    // ...
    // A bunch of stateful functionality that can change over time.
    // ...

    useTheThing = useTheThing;
}
```

TBH, I cannot imagine a real use case that would need a hook function inside a non-react class. Implementing things this way may lead to confusion by new contributors to the codebase, needlessly making the learning curve steeper.

blixt:
@kachkaev Rebinding the function might be a working solution, though as your example uses an arrow function it cannot be rebound and will not work. I think it can be made to work with a regular `function` statement though.

> TBH, I cannot imagine a real use case that would need a hook function inside a non-react class. Implementing things this way may lead to confusion by new contributors to the codebase, needlessly making the learning curve steeper.

I think a regular class does very little to confuse people – it's traditional JavaScript syntax. If the user understands hooks already, I don't think `instance.useThing(…)` is a steeper learning curve than `useThing(…)`. I *do* think working around a linter by moving methods outside the class could lead to confusion, though, so of course it would be nicer for the linter to allow defining hook methods in non-React component classes. But I think we're mostly theorizing here, so I can only speak for the use case at my company where we use these extensively.

I can also tell you about a real use case that we're actively using across our entire React codebase, which led to my comment in the first place:

```javascript
// (Very simplified to save space.)
export class FeatureSet {
    #activeConfig
    #listeners = new Map()
    constructor(config) {
        this.#activeConfig = { ...config }
    }
    update(changes) {
        for (const name in changes) {
            const variant = changes[name]
            this.#activeConfig[name] = variant
            const set = this.#listeners.get(name)
            if (!set) continue
            set.forEach(listener => listener(variant))
        }
    }
    useCondition(name, condition) {
        // This is a pretty involved hook that is hydration aware among other things
        // and uses internal state of FeatureSet so it would not be nice to move it out.
    }
    // more methods for inspecting state, adding listeners, etc
}

// Example use

// In some shared module:
const experiments = new FeatureSet({ checkoutEncouragement: "off" })

// In some app logic somewhere:
experiments.update({ checkoutEncouragement: "on" })

// Component:
function BuyButton() {
    const encourageUser = experiments.useCondition("checkoutEncouragement", v => v === "on")
    return <button>{encourageUser ? "BUY NOW!" : "Continue"}</button>
}
```

This is pretty much how we use it, overly simplified of course. We have multiple types of feature sets (tracking platform, active features based on user profile, and experiments for testing). The class tracks several internals such as realtime updates to the current state, and also tracks build-time state for effective hydration of the application.

This could of course all be rewritten to use some singleton store that has functions accessing it, but it's not the ideal way to write it, and I see little reason to go to such efforts to appease a linter.

kachkaev:
The code above looks over-engineered to me. I’d recommend checking out this article: [Tidy TypeScript: Avoid traditional OOP patterns](https://fettblog.eu/tidy-typescript-avoid-traditional-oop/). You can also check out how [apollo-client](https://github.com/apollographql/apollo-client) or [react-query](https://github.com/tannerlinsley/react-query) are implemented. In short, they have class-based stateful cores with clean class-free react hooks around them.

Issue #18033, opened 2020-02-13
DevTools Profiler: "getCommitTree(): Unable to reconstruct tree for root "…" and commit …"


morfioce:
Describe what you were doing when the bug occurred:
1. 
2. 
3. 

---------------------------------------------
Please do not remove the text below this line
---------------------------------------------

DevTools version: 4.4.0-f749045a5

Call stack: at N (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:40:154255)
    at e.getCommitTree (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:40:157019)
    at yl (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:40:320890)
    at ei (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:58864)
    at Mi (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:67500)
    at yl (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:108161)
    at kc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:99474)
    at wc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:99399)
    at fc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:96173)
    at chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:45937

Component stack: in yl
    in div
    in div
    in div
    in vo
    in Unknown
    in n
    in Unknown
    in div
    in div
    in Qi
    in Ve
    in nn
    in Da
    in Yc

bvaughn:
It doesn't look like this bug report has enough info for one of us to reproduce it.

Please provide a CodeSandbox (https://codesandbox.io/s/new), a link to a repository on GitHub, or provide a minimal code example that reproduces the problem. Screenshots or videos can also be helpful if they help provide context on how to repro the bug.

Here are some tips for providing a minimal example: https://stackoverflow.com/help/mcve

esvyridov:
@bvaughn Looks like the issue is still reproducible after a PR which was done in https://github.com/facebook/react/issues/16412 . I can try to reproduce the steps from https://github.com/facebook/react/issues/16412 if it'll help somehow.

bvaughn:
[...]

PR #18345, opened 2020-03-19
fix(devtools): check if data for commit index exists

jyash97:
## Summary
After Reload and profiling this app: https://gsqzl.csb.app/, traverse through the commit. Devtools break and throw an error for accessing data from `undefined` or `null`.

Profiling Data: https://gist.github.com/jyash97/612bc8d38c2f7f9842c19ae5d270252c

For more ref: https://github.com/facebook/react/issues/18033#issuecomment-601004423

## Test Plan

Added the check if the commit data exists than only access the value.

Tested with firefox, chrome extension and with inline package.

Commands used

```
yarn build:chrome && yarn test:chrome
```

&

```
yarn start ( for inline & shell package )
```

Screenshot Recording of Chrome extension:  https://youtu.be/1UyEZcfbjqg

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/CommitTreeBuilder.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/CommitTreeBuilder.js
@@ -79,9 +79,12 @@ export function getCommitTree({
 
     // Construct the initial tree.
     recursivelyInitializeTree(rootID, 0, nodes, dataForRoot);
-
     // Mutate the tree
-    if (operations != null && commitIndex < operations.length) {
+    if (
+      operations != null &&
+      commitIndex < operations.length &&
+      operations[commitIndex]
+    ) {
       const commitTree = updateTree({nodes, rootID}, operations[commitIndex]);
 
       if (__DEBUG__) {
@@ -98,7 +101,11 @@ export function getCommitTree({
       rootID,
     });
 
-    if (operations != null && commitIndex < operations.length) {
+    if (
+      operations != null &&
+      commitIndex < operations.length &&
+      operations[commitIndex]
+    ) {
       const commitTree = updateTree(
         previousCommitTree,
         operations[commitIndex],


jyash97:
@bvaughn  Do check these when you get time and lmk if the PR makes sense.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitTreeBuilder.js
```diff
@@ -79,9 +79,12 @@ export function getCommitTree({
 
     // Construct the initial tree.
     recursivelyInitializeTree(rootID, 0, nodes, dataForRoot);
-
     // Mutate the tree
-    if (operations != null && commitIndex < operations.length) {
+    if (
+      operations != null &&
+      commitIndex < operations.length &&
+      operations[commitIndex]
```
bvaughn: I think we should find and fix the underlying cause here. Why is `operations[commitIndex]` `undefined`? That's unexpected.

jyash97:
Hey @bvaughn thanks for taking a look. 
Here id the Sandbox Source: https://codesandbox.io/s/getcommittree-o29bx

packages/react-devtools-shared/src/devtools/views/Profiler/CommitTreeBuilder.js
jyash97: On debugging the underlying issue, I encountered this issue so I thought it would be better to add the check and get the correct error.

bvaughn:
Hm. There is a duplicate key warning in this app. I don't understand why it manifests the way it does with the profiler, or why React adds extra nodes when it should just be swapping between two, but if you fix that warning- the behavior of your app is what I would expect (and the Profiler works fine).

bvaughn:
I've reduced the error case to poke around a little further though, to see if I can figure out why the Profiler is reacting the way it is. (See [here](https://codesandbox.io/s/getcommittree-mtgto).)
```js
const One = () => null;
const Two = () => null;

function App() {
  const [didMount, setDidMount] = useState(false);

  useEffect(() => {
    setDidMount(true);
  }, []);

  return (
    <>
      {didMount ? <Two /> : <One key={1} />}
      <One key={1} />
    </>
  );
}
```

bvaughn:
Ah, I see why the Profiler is erroring in this case. The invalid `key` situation causes a DevTools invariant:
https://github.com/facebook/react/blob/0140118e8e2766abaac2c833775434f83df81d56/packages/react-devtools-shared/src/devtools/store.js#L929-L933

This thrown error means that the Profiler code that normally processes the commit (after the Components store) doesn't run.

I may update this specifically, but in general- this case isn't really meant to be supported. I wish we "failed" in a more meaningful way though! Let me give it a little thought.

bvaughn:
I've opened an alternate "fix" for this: #18378

Want to check it out?

jyash97:
Hey @bvaughn thanks for the review and letting me know the underneath issue. I will surely take look at your PR 💯

PR #18346, opened 2020-03-19
ModernEventSystem: fix event replaying

trueadm:
If you run all the tests with modern event system flag:

```
export const enableModernEventSystem = true;
```

This PR, thanks to Seb now should fix those use-cases with the modern event system and the event replaying mechanics.

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -302,60 +302,62 @@ export function dispatchEventForPluginEventSystem(
     ) {
       return;
     }
-    // The below logic attempts to work out if we need to change
-    // the target fiber to a different ancestor. We had similar logic
-    // in the legacy event system, except the big difference between
-    // systems is that the modern event system now has an event listener
-    // attached to each React Root and React Portal Root. Together,
-    // the DOM nodes representing these roots are the "rootContainer".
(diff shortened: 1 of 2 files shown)


trueadm:
After speaking to @sebmarkbage it sounds like we need a larger fix.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -302,60 +302,62 @@ export function dispatchEventForPluginEventSystem(
...
-    // systems is that the modern event system now has an event listener
-    // attached to each React Root and React Portal Root. Together,
-    // the DOM nodes representing these roots are the "rootContainer".
-    // To figure out which ancestor instance we should use, we traverse
-    // up the fiber tree from the target instance and attempt to find
-    // root boundaries that match that of our current "rootContainer".
-    // If we find that "rootContainer", we find the parent fiber
-    // sub-tree for that root and make that our ancestor instance.
-    let node = targetInst;
+    if (targetInst !== null) {
```
sebmarkbage: Why was this needed?

trueadm: The event replaying system passed `null` for target fiber. If we don't guard against it we will just early return and never dispatch anything, which can cause issues if I'm not mistaken?

PR #18348, opened 2020-03-19
Revert accidental changes to package.json

acdlite:
The publish script was written before we switched to running patch releases out-of-band, so when updating the local package.json version numbers, it accidentally reverted other changes that have landed to master since 16.13 was released.

Comparison to before the automated changes. Only differences are to the changelog and the version numbers: https://github.com/facebook/react/compare/8311cb5d24fd1648ddffb659bf9b3317f5f885d0...c4eda2b

### DIFF
--- a/packages/jest-react/package.json
+++ b/packages/jest-react/package.json
@@ -23,6 +23,9 @@
     "react": "^16.0.0",
     "react-test-renderer": "^16.0.0"
   },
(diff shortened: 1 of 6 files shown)


acdlite:
CodeSandbox CI is failing with that transient error we already know about. Not important. Merging to fix master.

PR #18349, opened 2020-03-19
ModernEventSystem: refine Flow types

trueadm:
This PR cleans up some Flow typing any casts.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMUseEvent.js
+++ b/packages/react-dom/src/client/ReactDOMUseEvent.js
@@ -7,6 +7,7 @@
  * @flow
  */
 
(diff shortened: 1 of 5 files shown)


packages/react-dom/src/client/ReactDOMUseEvent.js
```diff
@@ -64,6 +65,9 @@ export function useEvent(
       priority = optionsPriority;
     }
   }
+  if (priority === undefined) {
+    priority = getEventPriorityForListenerSystem(type);
```
trueadm: Noticed this when refining the type here. We shouldn't do a lookup for priority if we supply one.

PR #18351, opened 2020-03-19
ReactDOM.useEvent: support custom types

trueadm:
This PR adds support for custom user events to the `useEvent` system. I've also added tests demonstrating them working as expected. I also added an unnecessary `phasedRegistrationNames == null` check in `accumulateTwoPhaseListeners` that never occurs at runtime. This essentially allows this:

```jsx
function Component() {
  const ref = useRef(null)
  const foo = ReactDOM.unstable_useEvent('foo')

  useEvent(() => {
    foo.setListener(ref.current, () => {
      console.log('Custom user events work!');
    });
  });

  return <span ref={ref}>;
}

const event = document.createEvent('Event');
event.initEvent('foo', true, true);
document. getElementsByName('span')[0].dispatchEvent(event);
```

### DIFF
--- a/packages/legacy-events/ReactSyntheticEventType.js
+++ b/packages/legacy-events/ReactSyntheticEventType.js
@@ -19,14 +19,21 @@ export type DispatchConfig = {|
     captured: null | string,
   |},
   registrationName?: string,
-  eventPriority?: EventPriority,
-  customEvent?: boolean,
+  eventPriority: EventPriority,
+|};
+
+export type CustomDispatchConfig = {|
+  phasedRegistrationNames: {|
+    bubbled: null,
+    captured: null,
+  |},
+  customEvent: true,
 |};
 
 export type ReactSyntheticEvent = {|
-  dispatchConfig: DispatchConfig,
+  dispatchConfig: DispatchConfig | CustomDispatchConfig,
   getPooled: (
-    dispatchConfig: DispatchConfig,
+    dispatchConfig: DispatchConfig | CustomDispatchConfig,
     targetInst: Fiber,
     nativeTarget: Event,
     nativeEventTarget: EventTarget,
--- a/packages/legacy-events/ReactSyntheticEventType.js
+++ b/packages/legacy-events/ReactSyntheticEventType.js
@@ -13,13 +13,14 @@ import type {EventPriority} from 'shared/ReactTypes';
 import type {TopLevelType} from './TopLevelEventTypes';
 
 export type DispatchConfig = {|
-  dependencies: Array<TopLevelType>,
-  phasedRegistrationNames?: {|
-    bubbled: string,
-    captured: string,
+  dependencies?: Array<TopLevelType>,
+  phasedRegistrationNames: {|
+    bubbled: null | string,
+    captured: null | string,
   |},
   registrationName?: string,
-  eventPriority: EventPriority,
+  eventPriority?: EventPriority,
+  customEvent?: boolean,
 |};
 
 export type ReactSyntheticEvent = {|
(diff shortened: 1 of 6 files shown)


packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -1782,6 +1786,74 @@ describe('DOMModernPluginEventSystem', () => {
...
+            const log = [];
+            const onCustomEvent = jest.fn(e =>
+              log.push(['bubble', e.currentTarget]),
+            );
+            const onCustomEventCapture = jest.fn(e =>
+              log.push(['capture', e.currentTarget]),
+            );
+
+            function Test() {
+              let custom;
```
threepointone:
```suggestion
              let customEventHandle;
```
maybe not 'Handle', but I do think this should be a descriptive name

packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -1782,6 +1786,74 @@ describe('DOMModernPluginEventSystem', () => {
...
+              }).toWarnDev(
+                'Warning: The event "type" provided to useEvent() does not have a known priority type. ' +
+                  'It is recommended to provide a "priority" option to specify a priority.',
+              );
+
+              custom = ReactDOM.unstable_useEvent('custom-event', {
+                priority: 0, // Discrete
+              });
+
+              const customCapture = ReactDOM.unstable_useEvent('custom-event', {
```
threepointone: same here

packages/react-dom/src/events/SimpleEventPlugin.js
```diff
@@ -172,7 +172,10 @@ const SimpleEventPlugin: PluginModule<MouseEvent> = {
         break;
       default:
         if (__DEV__) {
-          if (knownHTMLTopLevelTypes.indexOf(topLevelType) === -1) {
+          if (
+            knownHTMLTopLevelTypes.indexOf(topLevelType) === -1 &&
```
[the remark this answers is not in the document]
trueadm: It's dev only :)

packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -1782,6 +1786,74 @@ describe('DOMModernPluginEventSystem', () => {
...
+
+            function Test() {
+              let custom;
+
+              // Test that we get a warning when we don't provide an explicit priortiy
+              expect(() => {
+                custom = ReactDOM.unstable_useEvent('custom-event');
+              }).toWarnDev(
+                'Warning: The event "type" provided to useEvent() does not have a known priority type. ' +
+                  'It is recommended to provide a "priority" option to specify a priority.',
```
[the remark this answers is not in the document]
trueadm: The possible values are 0, 1, 2 – for Discrete, user blocking and continious. I think we'll need to expose them from an enum on React maybe, but right now I'm unsure how that would work.

packages/legacy-events/ReactSyntheticEventType.js
```diff
@@ -13,19 +13,27 @@ import type {EventPriority} from 'shared/ReactTypes';
...
   registrationName?: string,
   eventPriority: EventPriority,
 |};
 
+export type CustomDispatchConfig = {|
+  phasedRegistrationNames: {|
+    bubbled: null,
+    captured: null,
+  |},
+  customEvent?: boolean,
```
threepointone: won't this always be true?

PR #18355, opened 2020-03-20
ReactDOM.useEvent: add `EventTarget` support

trueadm:
This PR adds support to `useEvent` so that `setListener` can take objects that are valid `EventTarget`s. For example, you can use `window` or `document` with `useEvent` as of this PR:

```jsx
function Component() {
  const clickHandle = ReactDOM.unstable_useEvent('click')

  useEffect(() => {
    clickHandle.setListener(document, () => {
      console.log('it works!');
    });
  });

  return <div>Click anywhere on the document</div>
}
```

We check for a valid `EventTarget` by if it has an `addEventListener` property that is a method. Given that React DOM's managed elements will also have this, we always first check if the `EventTarget` is a React managed DOM node (by checking if it has a fiber) and if it is, we use the existing React fiber two phase dispatching to find `useEvent` and `on*` prop events.  

However, if the `EventTarget` is not managed by React, i.e. an object like `window`, `document`,  `XMLHttpRequest`, `AudioNode`, `AudioContext` etc then we instead add event listener directly to the `EventTarget` rather than the container root (a different listener for capture vs bubble phase too) and then dispatch events directly off the target rather than through the fiber tree. Given propagation should work as expected now that the modern event system uses React roots, events should occur in the correct order.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -1165,7 +1165,7 @@ export function validateEventListenerTarget(
     if (__DEV__) {
       console.warn(
         'Event listener method setListener() from useEvent() hook requires the first argument to be ' +
-          'a valid DOM EventTarget interface object. If using a ref, ensure the current value is not null.',
+          'a valid DOM EventTarget. If using a ref, ensure the current value is not null.',
       );
     }
   }
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -75,11 +75,13 @@ import {
   IS_PASSIVE,
 } from 'legacy-events/EventSystemFlags';
 import {
-  attachElementListener,
-  detachElementListener,
-  isDOMDocument,
-  isDOMElement,
+  isManagedDOMElement,
+  isValidEventTarget,
   listenToTopLevelEvent,
+  detachListenerFromManagedDOMElement,
+  attachListenerFromManagedDOMElement,
+  detachTargetEventListener,
+  attachTargetEventListener,
 } from '../events/DOMModernPluginEventSystem';
 import {getListenerMapForElement} from '../events/DOMEventListenerMap';
 
@@ -1122,12 +1124,10 @@ export function registerEvent(
 export function mountEventListener(listener: ReactDOMListener): void {
   if (enableUseEventAPI) {
     const {target} = listener;
-    if (target === window) {
-      // TODO (useEvent)
-    } else if (isDOMDocument(target)) {
-      // TODO (useEvent)
-    } else if (isDOMElement(target)) {
-      attachElementListener(listener);
+    if (isManagedDOMElement(target)) {
+      attachListenerFromManagedDOMElement(listener);
+    } else {
+      attachTargetEventListener(listener);
     }
   }
 }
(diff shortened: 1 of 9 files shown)


threepointone:
This seems to include changes from the last PR. Rebase please?

trueadm:
@threepointone Rebased, sorry about that!

TrySound:
useEvent? Maybe useEffect?
```
  useEvent(() => {
    clickHandle.setListener(document, () => {
      console.log('it works!');
    });
  });
```

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1169,11 +1164,8 @@ export function validateEventListenerTarget(
     }
     if (__DEV__) {
       console.warn(
-        'Event listener method setListener() from useEvent() hook requires the first argument to be either:' +
-          '\n\n' +
-          '1. A valid DOM node that was rendered and managed by React\n' +
-          '2. The "window" object\n' +
-          '3. The "document" object',
+        'Event listener method setListener() from useEvent() hook requires the first argument to be ' +
+          'a valid DOM EventTarget interface object. If using a ref, ensure the current value is not null.',
```
bvaughn: nit: "a valid DOM EventTarget." seems sufficient?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -132,6 +132,17 @@ const emptyDispatchConfigForCustomEvents: CustomDispatchConfig = {
 
 const isArray = Array.isArray;
 
+// $FlowFixMe: Flow struggles with this pattern
+const PossiblyWeakMap = typeof WeakMap === 'function' ? WeakMap : Map;
+// $FlowFixMe: Flow cannot handle polymorphic WeakMaps
```
bvaughn: Seems better to cast to `Map` (or `WeakMap`) so we have *some* type info rather than `$FlowFixMe`?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -198,11 +209,22 @@ export function listenToTopLevelEvent(
   listenerMap: ElementListenerMap,
   passive?: boolean,
   priority?: EventPriority,
+  capture?: boolean,
 ): void {
-  const listenerEntry = listenerMap.get(topLevelType);
+  // If we explicitly define capture, then these are for EventTarget objects,
+  // rather than React managed DOM elements. So we need to ensure we separate
+  // capture and non-capture events (for React managed DOM nodes we only use
+  // one or the other, never both).
```
bvaughn: > "*for React managed DOM nodes we only use one or the other, never both*"

I'm not sure I understand this comment. It sounds like it's saying that you can't attach both a capture *and* bubble event to a React managed DOM node, but looking at the implementation, that doesn't seem correct.

Maybe you could change the wording to something clearer?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: I get the same issue with Maps, I kind of gave up here.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: That is correct, we never attach a catpture and bubble event listener to React managed DOM nodes. We have a Set that determines which one we do use, based on the event name, and we use one one and we emulate the effect of bubble/capture when we dispatch a two phase event. I'll update the comment to include that info.

PR #18356, opened 2020-03-20
Fix to prevent from leaking error messages in prod environment (ReactFlightServer)

suthagar23:
## Summary

Fix to prevent from leaking error messages in prod environment (ReactFlightServer). Added changes based on the existing TODO, and removed it after the fix.

### DIFF
--- a/packages/react-server/src/ReactFlightServer.js
+++ b/packages/react-server/src/ReactFlightServer.js
@@ -176,7 +176,12 @@ export function resolveModelToJSON(
   return value;
 }
 
-function emitErrorChunk(request: Request, id: number, error: mixed, errorCode: string): void {
+function emitErrorChunk(
+  request: Request,
+  id: number,
+  error: mixed,
+  errorCode: string,
+): void {
   let message;
   let stack = '';
   if (__DEV__) {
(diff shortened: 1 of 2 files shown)


packages/react-server/src/ReactFlightServer.js
```diff
@@ -176,21 +176,29 @@ export function resolveModelToJSON(
...
   let message;
   let stack = '';
-  try {
-    if (error instanceof Error) {
-      message = '' + error.message;
-      stack = '' + error.stack;
-    } else {
-      message = 'Error: ' + (error: any);
+  if (__DEV__) {
+    console.log('111111111111111')
```
awibox: console.log should be deleted

packages/react-server/src/ReactFlightServer.js
```diff
@@ -176,21 +176,29 @@ export function resolveModelToJSON(
...
+        message = 'Error: ' + (error: any);
+      }
+    } catch (x) {
+      message = 'An error occurred but serializing the error message failed.';
     }
-  } catch (x) {
-    message = 'An error occurred but serializing the error message failed.';
+  } else {
+    message = 'An error occurred(' + (errorCode ? errorCode : 'UNKNOWN') + ')';
+    console.log('22222222222222', message)
```
awibox: console.log should be deleted

suthagar23:
Thanks @awibox for pointing it out. It was removed with the last commit.

PR #18360, opened 2020-03-21
Execute hooks in separate namespaces for concurrent server renderers

### LINKED ISSUE #16416: SSR: Cannot set property 'memoizedState' of null

bl00mber:
fix #16416


bl00mber:
@sebmarkbage 
> A possible fix could be to store currentlyRenderingComponent on the stack in the synchronous rendering entry points (renderToString) and then restore it to its previous state upon existing.

You mean only `currentlyRenderingComponent` or `currentlyRenderingComponent` and `workInProgressHook` (on which the error is currently thrown) or all variables associated with reassignment should be stored in the array of objects of variables within `ReactDOMStrongRenderer.js`, and pulled out for modification and evaluation by the `ReactPartialRendererHooks.js` methods using array index when needed? Or passed to the `new ReactPartialRenderer` constructor? Or should instead error be thrown on an attempt to use nested renderers?

bl00mber:
@sebmarkbage can you please specify what does it mean to "restore it to its previous state upon exiting"? I've tried to initialize additional hooks scope to the `ReactPartialRenderer.js`, but when it is passed to `ReactDOMServerRenderer` it needs to be used in Dispatcher which is a simple object, not a class which can be instantiated, so it shared the same scope and, respectively, the same reference to the hooksScope (if it will be assigned as a Disapatcher property)! Should Dispatcher be simply cloned?

sebmarkbage:
Something needs to be temporarily mutated and then restored upon exiting. I'm not sure exactly what. We use this pattern a lot:

```
let current;
function foo(otherThing) {
  let previous = current;
  current = otherThing;
  try {
    render();
  } finally {
    current = previous;
  }
}
```

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -64,6 +70,20 @@ let isInHookUserCodeInDev = false;
 // In DEV, this is the name of the currently executing primitive hook
 let currentHookNameInDev: ?string;
 
+export function getHookInternals() {
+  return {
+    currentlyRenderingComponent,
+    firstWorkInProgressHook,
+    workInProgressHook,
+  };
```
gaearon, line 78: What about `renderPhaseUpdates` and others? I think those would break too.

bl00mber: I added only variables which were different before modification in finishHooks. Should I add others?

gaearon: `renderPhaseUpdates` gets set to `null` there. So I think it might break if you don't.

One thing you can try is writing a test that triggers writes and reads to each of them. (E.g. by breaking them and finding existing tests that cover this code.) Then write a test that the same pattern works with nested render calls.

bl00mber: Seems like this pattern breaks when adding second response.

gaearon: Have you figured out why?

bl00mber: Test passes, but there is no isolation of variables. Because renderers are asynchronous, one of the renderers rewrites `workInProgressHook` to undefined in one of the `createWorkInProgressHook` calls. Perhaps its because during execution the wrong `workInProgressHook` used. When `hookInternals` are initialized for the second time, they are empty objects, despite comment in `prepareToUseHooks` which says they should be the `null` values (and `workInProgressHook` is of the Hook type in the second `prepareToUseHooks` call).

bl00mber:
Performance tests (1000000 static markups with useMemo hook)
before: **25782** ms
after: **26413** ms

PR #18362, opened 2020-03-21
Refactor Lazy Components to use teh Suspense (and wrap Blocks in Lazy)

sebmarkbage:
I wasn't happy with my approach in https://github.com/facebook/react/pull/18220

Exposing the status stuff leaks a lot of complexity downstream. It avoids taking on a virtual dispatch but also means that other code (like Flight and Blocks) can't take advantage of what Lazy already does.

Using the suspense protocol is actually really nice in terms of code size and complexity. The principle is simple:

- Expose a function to call to get the value. This lazily does any work it needs to do.
- If it can't, throw an error or promise.

The lazy component type can now be used to implement all kinds of lazy behavior. So I also use it to implement lazy Blocks.


packages/shared/getComponentName.js
```diff
@@ -90,12 +89,14 @@ function getComponentName(type: mixed): string | null {
...
       case REACT_LAZY_TYPE: {
-        const thenable: LazyComponent<mixed> = (type: any);
-        const resolvedThenable = refineResolvedLazyComponent(thenable);
-        if (resolvedThenable) {
-          return getComponentName(resolvedThenable);
+        const lazyComponent: LazyComponent<any, any> = (type: any);
+        let payload = lazyComponent._payload;
+        let init = lazyComponent._init;
+        try {
+          return getComponentName(init(payload));
```
sebmarkbage: Getting the component name can now have side-effects, causing data to load.

packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -1188,17 +1205,20 @@ function ChildReconciler(shouldTrackSideEffects) {
           }
           case Block:
             if (enableBlocksAPI) {
-              if (element.type.$$typeof === REACT_BLOCK_TYPE) {
+              let type = element.type;
+              if (type.$$typeof === REACT_LAZY_TYPE) {
+                type = resolveLazyType(type);
```
sebmarkbage: For updates, we attempt to initialize earlier to see if we can reconcile against the inner value. I only do this for Blocks right now to preserve backwards compatibility.

However, in non-legacy mode, we could reconcile against the resolved value of any type.

E.g. if I have two lazy components that resolve to the same thing, why shouldn't they reconcile against each other?

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -492,7 +487,14 @@ function updateSimpleMemoComponent(
...
         // so let's just skip over it to find memo() outer wrapper.
         // Inner props for memo are validated later.
-        outerMemoType = refineResolvedLazyComponent(outerMemoType);
+        const lazyComponent: LazyComponentType<any, any> = outerMemoType;
+        let payload = lazyComponent._payload;
+        let init = lazyComponent._init;
+        try {
+          outerMemoType = init(payload);
+        } catch (x) {
+          outerMemoType = null;
```
gaearon: Why not just let it throw? If we got here, doesn't mean it's resolved anyway?

packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
```diff
@@ -879,7 +879,13 @@ describe('ReactLazy', () => {
       },
     );
 
-    expect(Scheduler).toFlushAndYield(['Started loading', 'Loading...']);
+    if (__DEV__) {
+      // Getting the name for the warning cause the loading to start early.
```
gaearon: Don't know how I feel about this. Can you motivate?

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: Hm. I think you’re right. I think that we’d only get here if the init is not deterministic. I.e. if it resuspends. That’s a bug in the lazy code but could possibly happen as a bug in a “user provided lazy”.

That would be a prod changing behavior so probably best to keep this since it wouldn’t happen in prod (although the call wouldn’t happen neither).

packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js
sebmarkbage: I think we really should be refactoring the getComponentName stuff. One todo I have is to pass it a fiber because the type is not really enough. If we had the fiber we could read the .type off of that.

It’s called everywhere though so it’s some work to properly clean that up.

What annoys be a bit more is that we have that logic in prod. I suspect that’s what leads this PR to increase bundle size but just a little bit.

We probably should fork getComponentName and accompanying logic into specific DEV and prod versions for different use cases.

Regardless, I’m confident the fix is in the logging and not compromising the logic for lazy resolution.

packages/react-reconciler/src/ReactChildFiber.js
```diff
@@ -419,22 +436,22 @@ function ChildReconciler(shouldTrackSideEffects) {
...
-        current.tag === Block &&
-        element.type.$$typeof === REACT_BLOCK_TYPE
-      ) {
+      } else if (enableBlocksAPI && current.tag === Block) {
         // The new Block might not be initialized yet. We need to initialize
         // it in case initializing it turns out it would match.
-        initializeBlockComponentType(element.type);
+        let type = element.type;
+        if (type.$$typeof === REACT_LAZY_TYPE) {
+          type = resolveLazyType(type);
```
sebmarkbage, line 444: This got lost when we dropped Blocks. The benefit of Blocks was that we only needed this check if the tag was Block but now we need to do it for everything.

PR #18363, opened 2020-03-21
Move some files out of /shared and rename to upper case

sebmarkbage:
This moves some stuff out of shared and into their respective packages.

I changed the isomorphic "creator" types to use the same convention (upper case) with named exports. It was inconsistent before.

Things that are Fiber specific can live in the reconciler because any individual renderer is allowed to deep link into react-reconciler. At least ours are.


trueadm:
I believe the RN sync script pulls some bits out of shared, like ReactTypes.

sebmarkbage:
Yea it's only ReactTypes and ReactNativeTypes. However, I'll separately open a diff to stop syncing ReactTypes. We shouldn't be using that in RN because it's not consistent with the upstream Flow types. That causes issues and all current imports were incorrect and has been codemodded so they're now unused.

PR #18364, opened 2020-03-21
Rename CI jobs after the command line equivalent

sebmarkbage:
This makes it easier to know what to write on your command line to replicate a failure locally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
           paths:
             - ~/.cache/yarn
 
-  lint:
+  yarn_lint:
     docker: *docker
     environment: *environment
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -152,7 +152,7 @@ jobs:
             RELEASE_CHANNEL: stable
           command: yarn test-www --maxWorkers=2
 
-  test_source_www_variant_prod:
+  RELEASE_CHANNEL_stable_NODE_ENV_prod_yarn_test_www_variant:
```
gaearon: Is there a character limit? Might as well write out production.

dangerfile.js
```diff
@@ -38,7 +38,7 @@ const RELEASE_CHANNEL =
 const artifactsJobName =
   process.env.RELEASE_CHANNEL === 'experimental'
     ? 'process_artifacts_experimental'
-    : 'process_artifacts';
+    : 'process_artifacts_stable';
```
gaearon: Needs diffs to sync scripts or a note in the group for the next person.

sebmarkbage: Reverting this to avoid the thrash. This is not a command you run anyway.

PR #18366, opened 2020-03-21
Stop syncing ReactTypes to RN

sebmarkbage:
This is a really old one and all callers have since been codemodded away anyway because of problems.

This file is not really as rigorously maintained as the official Flow types but has a few more specifics. However, the inconsistency causes problems when you try to pass files typed using the built-in Flow typing for React and mix it with these.

We just happen to get away with it because we compile out the types. If we didn't we would hit those problems by even using these in our renderers. We should only use the official types externally.

Ideally we should tighten up the external types and then just use those internally too.

### DIFF
--- a/scripts/rollup/packaging.js
+++ b/scripts/rollup/packaging.js
@@ -87,11 +87,9 @@ async function copyWWWShims() {
 }
 
 async function copyRNShims() {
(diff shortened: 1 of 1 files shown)


sebmarkbage:
[Internal diff](https://our.internmc.facebook.com/intern/diff/D20583740/).

PR #18370, opened 2020-03-23
ReactDOM.useEvent: add support for beforeblur/afterblur

trueadm:
This ports the existing Flare logic for handling `beforeblur` and `blur` over to the ReactDOM.useEvent system. The only real differences are:

- We don't invoke events directly into the modern event system, like we do with the Flare events (we invoke the Flare event system directly). Instead, we trigger a native DOM custom event and let the DOM event listeners catch it. We have to do this because we now listen to both roots and also EventTarget nodes. If we only trigger them directly to the event system then propagation is out of sequence because we can have many roots/portals listening with their native event listeners. If we have multiple ReactDOM's of different versions, they should also be able to know of this event too – if we only invoke it internally, they'll never know of the `beforeblur` event.
- Given the last point, I opted to move away from `blur` and instead we now use `afterblur`. Technically, that is also the correct name as `blur` occurs after the DOM node is removed and we're firing this in the commit phase that is actually after all the `blur` events would have been anyway. We do this because we're now invoking native events and over-firing a common event such as `blur` triggers listeners that previously never fired and we definitely don't want that. We nominate the `relatedTarget` to tell us what the detached node was and we fire it only on the `document`.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -520,6 +520,9 @@ function dispatchBeforeDetachedBlur(target: HTMLElement): void {
       // to dispatch the "beforeblur" event.
       ReactBrowserEventEmitterSetEnabled(true);
       const event = createEvent(TOP_BEFORE_BLUR);
+      // Dispatch "beforeblur" directly on the target,
+      // so it gets picked up by the event system and
+      // can propagate through the React internal tree.
       target.dispatchEvent(event);
     } finally {
       ReactBrowserEventEmitterSetEnabled(false);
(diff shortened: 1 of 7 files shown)


packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -490,34 +492,76 @@ export function insertInContainerBefore(
...
+    // which would not normally be expecting that event. So we instead
+    // use a new type of event, like we did with "beforeblur". Also,
+    // technically this period is actually after "blur", as Chrome invokes
+    // "blur" natively just as the node gets removed, thus having a valid
+    // target. If we fire a custom blue at this point, it won't work as
+    // the node has been attached, thus the event system will not pick it
+    // up. We have to use the native system, rather than invoke internally,
+    // like we used to do with Flare because we listen on multiple roots
+    // and we don't know the ordering of propagation unless we invoke a
+    // native event.
```
necolas: We should describe this without reference to Flare since that's going to be removed

packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -2139,6 +2144,174 @@ describe('DOMModernPluginEventSystem', () => {
...
+            target.focus();
+            expect(onBeforeBlur).toHaveBeenCalledTimes(0);
+            expect(onAfterBlur).toHaveBeenCalledTimes(0);
+
+            ReactDOM.render(<Component show={false} />, container);
+            Scheduler.unstable_flushAll();
+
+            expect(onBeforeBlur).toHaveBeenCalledTimes(1);
+            expect(onAfterBlur).toHaveBeenCalledTimes(1);
+            expect(onAfterBlur).toHaveBeenCalledWith(
```
necolas: Might want to the test to cover that the order of the 3 blur events is as expected

trueadm: There's only two blur events to test though. We don't expect `blur` to fire (it does for Chrome, but it doesn't in other browsers).

packages/react-dom/src/client/ReactDOMHostConfig.js
trueadm: Removed this comment.

PR #18371, opened 2020-03-23
[Flight] Integrate Blocks into Flight

sebmarkbage:
The previous model was centered around JSX rendering random models. This moves to a Block centric approach. To do this, we need a way to pass modules to the client so this wires up the previous infra to resolve modules into meta data.

The server is expected to replace a Block with a compiler to instead return a "Server Block" which contains a ModuleReference and Query. We don't have the compiler parts yet so the tests mimics what the compiler would do for server code.

Now the expected Blocks-first programming model actually works.


packages/react-server/src/ReactFlightServer.js
```diff
@@ -81,26 +89,31 @@ export function createRequest(
...
   let type = element.type;
   let props = element.props;
   if (typeof type === 'function') {
-    // This is a nested view model.
+    // This is a server-side component.
     return type(props);
   } else if (typeof type === 'string') {
     // This is a host element. E.g. HTML.
     return [REACT_ELEMENT_TYPE, type, element.key, element.props];
+  } else if (type[0] === REACT_SERVER_BLOCK_TYPE) {
```
gaearon: What's this? `block` return value on server only? How will this be wired up? I see you copy pasted it into tests but it's still a bit confusing.

gaearon: Ohh I see you're saying the compiler would do that. Okay.

packages/react-client/src/__tests__/ReactFlight-test.js
```diff
@@ -10,7 +10,11 @@
 
 'use strict';
 
+const ReactFeatureFlags = require('shared/ReactFeatureFlags');
```
gaearon: @acdlite Did you say we need to put these inline now? I don't remember why.

packages/react-server/src/ReactFlightServer.js
sebmarkbage: Yea. I mean I could keep it as $$typeof style objects and then convert it to arrays as part of JSON stringify but I figured this could be a good place to start experimenting with first class tuple style objects instead.

packages/react-client/src/__tests__/ReactFlight-test.js
sebmarkbage: I copy-pasted this from elsewhere so if we do, we should codemod the rest of them.

packages/react-client/src/ReactFlightClient.js
```diff
@@ -138,10 +151,23 @@ export function reportGlobalError(response: Response, error: Error): void {
   });
 }
 
-function definePendingProperty(
+function readMaybeChunk<T>(maybeChunk: Chunk<T> | T): T {
+  if ((maybeChunk: any).$$typeof !== CHUNK_TYPE) {
```
gaearon: I've noticed `maybeChunk` is literally the `User` function in the Relay test. Is this intentional? We're missing some mocking that represents the module transport there?

packages/react-server/src/ReactFlightServer.js
sebmarkbage: I think the compiler output should probably depend on a helper that generates this.

I'm thinking `react/flight-runtime` with a block helper.

packages/react-client/src/ReactFlightClient.js
sebmarkbage: It wasn't intentional at first but I did notice it while writing the tests and decided to keep it. I think we should do some mocking here but we don't have it wired up yet so we're not sure what we're simulating yet.

Also, the tests are far from sufficient yet. Needs more async stuff. I need to write more but don't want to block on it.

packages/react-client/src/ReactFlightClient.js
```diff
@@ -197,6 +223,55 @@ function createElement(type, key, props): React$Element<any> {
...
+function initializeBlock<Props, Data>(
+  tuple: UninitializedBlockPayload<Data>,
+): BlockComponent<Props, Data> {
+  // Require module first and then data. The ordering matters.
+  let moduleMetaData: ModuleMetaData = readMaybeChunk(tuple[1]);
+  let moduleReference: ModuleReference<
+    BlockRenderFunction<Props, Data>,
+  > = resolveModuleReference(moduleMetaData);
+  // TODO: Do this earlier, as the chunk is resolved.
+  preloadModule(moduleReference);
```
gaearon: The comment before its implementation says

```
 Returning null means [...]
```

but retvalue is never used. Is it outdated?

Also, is it completely unobservable? I deleted this and nothing failed.

sebmarkbage: The comment is outdated.

It's not unobservable because of this: https://github.com/facebook/react/blob/master/packages/react-flight-dom-webpack/src/ReactFlightClientWebpackBundlerConfig.js#L60-L61

Lots more tests needs to be written at some point.

packages/react-client/src/ReactFlightClient.js
```diff
@@ -197,6 +223,55 @@ function createElement(type, key, props): React$Element<any> {
...
+];
+
+type Thenable<T> = {
+  then(resolve: (T) => mixed, reject?: (mixed) => mixed): Thenable<any>,
+};
+
+function initializeBlock<Props, Data>(
+  tuple: UninitializedBlockPayload<Data>,
+): BlockComponent<Props, Data> {
+  // Require module first and then data. The ordering matters.
```
gaearon: Can we test correct ordering?

sebmarkbage: Hm. Yea. We could test a side-effect in the require somehow. There's no module function that runs in the tests atm though so needs some infra or hook into the mocks somehow.

packages/react-client/src/__tests__/ReactFlight-test.js
```diff
@@ -19,24 +23,78 @@ describe('ReactFlight', () => {
...
+            {'A'}
+            {', '}
+            {'B'}
+          </div>
+        ),
+      },
+    });
   });
+
+  if (ReactFeatureFlags.enableBlocks) {
```
gaearon: None of this is running, you want `ReactFeatureFlags.enableBlocksAPI`

sebmarkbage: lol. This is going to be fun.

packages/react-client/src/ReactFlightClient.js
sebmarkbage: We also don't have any async tests yet.

packages/react-flight-dom-relay/src/__tests__/ReactFlightDOMRelay-test.internal.js
```diff
@@ -53,6 +41,81 @@ describe('ReactFlightDOMRelay', () => {
...
+    function User(props, data) {
+      return (
+        <span>
+          {props.greeting}, {data.name}
+        </span>
+      );
+    }
+    let loadUser = block(Query, User);
+    let model = {
+      User: loadUser('Seb', 'Smith'),
```
gaearon: lol

packages/react-flight-dom-relay/src/__tests__/ReactFlightDOMRelay-test.internal.js
```diff
@@ -53,6 +41,81 @@ describe('ReactFlightDOMRelay', () => {
...
+      let curriedQuery = () => {
+        return query(...args);
+      };
+      return [Symbol.for('react.server.block'), render, curriedQuery];
+    };
+  }
+
+  it('can render a server component', () => {
+    function Bar({text}) {
+      return text.toUpperCase();
```
gaearon: Is returning arrays from these supposed to work? I tried wrapping it in `[]` and got `Cannot assign to read only property 'children' of object '#<Object>'`

gaearon: Hmm actually this only happens when my `toEqual` fails. Maybe something's wrong with some test helper?

acdlite: This is a Jest bug introduced in 0.25.0 https://github.com/facebook/jest/issues/9531

acdlite: Fixed in master but not yet released https://github.com/facebook/jest/pull/9575

packages/react-flight-dom-webpack/src/__tests__/ReactFlightDOM-test.js
```diff
@@ -136,7 +163,11 @@ describe('ReactFlightDOM', () => {
     }
 
     let {writable, readable} = getTestStream();
-    ReactFlightDOMServer.pipeToNodeWritable(<RootModel />, writable);
+    ReactFlightDOMServer.pipeToNodeWritable(
+      <RootModel />,
```
gaearon: Should we delete these non-Block tests now? What is their use?

packages/react-flight-dom-relay/src/__tests__/ReactFlightDOMRelay-test.internal.js
sebmarkbage: There's more work to do on server components. If anything it's too permissive and allows anything to serialize here. It should take a separate path that only allows certain renderable things, same as the client.

Somewhat surprising that it doesn't allow an array though.

packages/react-flight-dom-webpack/src/__tests__/ReactFlightDOM-test.js
sebmarkbage: They're the only ones testing async right now but ideally they should be rewritten to a more idiomatic style.

packages/react-flight-dom-relay/src/__tests__/ReactFlightDOMRelay-test.internal.js
sebmarkbage: Actually I think that particular error actually just a problem with jest logging/matching. When things are not toEqual which includes elements, that are frozen, for some reason there's a write somewhere. So if it doesn't match you get a completely confusing error.

You probably just didn't update the toEqual case to match.

packages/react-server/src/ReactFlightServer.js
```diff
@@ -138,39 +151,95 @@ function escapeStringValue(value: string): string {
...
+    return escapeStringValue(value);
   }
 
+  // Resolve server components.
   while (
     typeof value === 'object' &&
     value !== null &&
     value.$$typeof === REACT_ELEMENT_TYPE
   ) {
+    // TODO: Concatenate keys of parents onto children.
```
gaearon: What case is this about?

packages/react-flight-dom-relay/src/__tests__/ReactFlightDOMRelay-test.internal.js
gaearon: Yeah you're right.

packages/react-server/src/ReactFlightServer.js
sebmarkbage: Server components render away but their keys should still count.

```
return c ? <Foo key="a" /> : <Foo key="b" />;
```

```
function Foo() {
  return <div key="meh" />;
}
```

This renders `['$', 'div', 'meh', null]` regardless. But if this is an update request, it should clear the state because the key has changed.

It's not going to be perfectly compatible semantics since the component type should be part of the equation too but that's a bit too much and not necessary since it doesn't have state.

I think a good compromise would be to concatenate the key onto the child. Possibly with some escaping rules. Similar to what React.Children does.

PR #18374, opened 2020-03-24
Flip the arguments of Blocks and make the query optional

sebmarkbage:
The Query is an optional overload so I'll put that at the end instead. It also lines up with where the "data" argument goes into the render.

Originally I had some thought around how it would be better for types to infer the result of the first argument flowing into the second. Maybe as you're typing it out. But the current type systems and IDE wouldn't take advantage of this anyway so doesn't matter.

cc @devknoll

### DIFF
--- a/packages/react-client/src/__tests__/ReactFlight-test.js
+++ b/packages/react-client/src/__tests__/ReactFlight-test.js
@@ -29,15 +29,15 @@ describe('ReactFlight', () => {
     act = ReactNoop.act;
   });
 
-  function block(render, query) {
+  function block(render, load) {
     return function(...args) {
-      if (query === undefined) {
+      if (load === undefined) {
         return [Symbol.for('react.server.block'), render];
       }
-      let curriedQuery = () => {
-        return query(...args);
+      let curriedLoad = () => {
+        return load(...args);
       };
-      return [Symbol.for('react.server.block'), render, curriedQuery];
+      return [Symbol.for('react.server.block'), render, curriedLoad];
     };
   }
 
@@ -73,7 +73,7 @@ describe('ReactFlight', () => {
   });
 
   if (ReactFeatureFlags.enableBlocksAPI) {
-    it('can transfer a Block to the client and render there, without a query', () => {
+    it('can transfer a Block to the client and render there, without data', () => {
       function User(props, data) {
         return (
           <span>
@@ -97,8 +97,8 @@ describe('ReactFlight', () => {
       expect(ReactNoop).toMatchRenderedOutput(<span>Hello undefined</span>);
     });
 
-    it('can transfer a Block to the client and render there, with a query', () => {
-      function query(firstName, lastName) {
+    it('can transfer a Block to the client and render there, with data', () => {
+      function load(firstName, lastName) {
         return {name: firstName + ' ' + lastName};
       }
       function User(props, data) {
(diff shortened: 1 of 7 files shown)


packages/react-client/src/__tests__/ReactFlight-test.js
```diff
@@ -29,8 +29,11 @@ describe('ReactFlight', () => {
...
   });
 
-  function block(query, render) {
+  function block(render, query) {
     return function(...args) {
+      if (query === undefined) {
+        return [Symbol.for('react.server.block'), render];
+      }
       let curriedQuery = () => {
         return query(...args);
```
sebmarkbage: Interestingly, if the query itself returns `undefined`, it will turn into `null` because of how JSON.stringify works on arrays. Fixing this would be kind of tricky but you're not really supposed to return `undefined` from a query anyway. We could warn for that same as we warn if you return undefined from render.

packages/react/src/ReactBlock.js
```diff
@@ -50,13 +50,13 @@ function lazyInitializer<Props, Args: Iterable<any>, Data>(
 }
 
 export function block<Args: Iterable<any>, Props, Data>(
-  query: BlockQueryFunction<Args, Data>,
   render: BlockRenderFunction<Props, Data>,
+  query?: BlockQueryFunction<Args, Data>,
```
sebmarkbage: When declared for external users, the overloaded type is:

```
declare export function block<Args: Iterable<any>, Props, Data>(
  render: BlockRenderFunction<Props, Data>,
  query: BlockQueryFunction<Args, Data>,
): (...args: Args) => Block<Props>;
declare export function block<Props>(
  render: BlockRenderFunction<Props, void>,
): () => Block<Props>;
```

sebmarkbage:
I also renamed `query` to `load`. That's really just a convention but that's the verb we use on the importing side by convention so I used that to mirror that part of the API.

packages/react-reconciler/src/__tests__/ReactBlocks-test.js
```diff
@@ -69,7 +91,7 @@ describe('ReactBlocks', () => {
       );
     }
 
-    let loadUser = block(Query, User);
+    let loadUser = block(User, load);
```
sebmarkbage: Unfortunate that the order gets reversed from the grammar. `User` + `load` = `loadUser`.

packages/react-client/src/__tests__/ReactFlight-test.js
```diff
@@ -70,8 +73,32 @@ describe('ReactFlight', () => {
...
+      function User(props, data) {
+        return (
+          <span>
+            {props.greeting} {typeof data}
+          </span>
+        );
+      }
+      let loadUser = block(User);
+      let model = {
+        User: loadUser('Seb', 'Smith'),
```
josephsavona: the type signature for loadUser here would disallow args, though, right?

packages/react-reconciler/src/__tests__/ReactBlocks-test.js
josephsavona: That's just a naming convention that we could change, though. It's probably helpful to use different names for the result of `block`, since calling `loadUser` here doesn't directly call `load`.

packages/react-client/src/__tests__/ReactFlight-test.js
josephsavona: > We could warn for that same as we warn if you return undefined from render.

yeah let's warn in that case

PR #18375, opened 2020-03-24
ReactDOM.useEvent: Add support for experimental scopes API

trueadm:
This is the last big piece of the `useEvent` API puzzle. It ports the existing behavior in React Flare, where you can attach listeners to the experimental Scopes API. This means that with `useEvent` you can attach events to a "scope" rather than a physical `EventTarget`, making way for us to be able to port the internal a11y components and their usages of this over to `useEvent`, allowing us to remove React Flare. This feature is a very important one, as it allows us to attach events to something that doesn't require a physical DOM element, as these can be problematic in places with `<table>` and `<select>`, plus adding in additional DOM wrapper elements can break the semantics of accessibility tools and CSS styling – which is why we offered this feature for Scopes with Flare.

### DIFF
--- a/packages/react-dom/src/events/accumulateTwoPhaseListeners.js
+++ b/packages/react-dom/src/events/accumulateTwoPhaseListeners.js
@@ -11,13 +11,19 @@ import type {DOMTopLevelEventType} from 'legacy-events/TopLevelEventTypes';
 import type {EventSystemFlags} from 'legacy-events/EventSystemFlags';
 import type {ReactSyntheticEvent} from 'legacy-events/ReactSyntheticEventType';
 
-import {HostComponent} from 'react-reconciler/src/ReactWorkTags';
-import {enableUseEventAPI} from 'shared/ReactFeatureFlags';
+import {
+  HostComponent,
+  ScopeComponent,
+} from 'react-reconciler/src/ReactWorkTags';
+import {enableUseEventAPI, enableScopeAPI} from 'shared/ReactFeatureFlags';
 
 import getListener from 'legacy-events/getListener';
 import {getListenersFromTarget} from '../client/ReactDOMComponentTree';
 import {IS_TARGET_EVENT_ONLY} from 'legacy-events/EventSystemFlags';
-import {eventTargetEventListenerStore} from './DOMModernPluginEventSystem';
+import {
+  eventTargetEventListenerStore,
+  reactScopeListenerStore,
+} from './DOMModernPluginEventSystem';
 
 export default function accumulateTwoPhaseListeners(
   event: ReactSyntheticEvent,
@@ -76,11 +82,13 @@ export default function accumulateTwoPhaseListeners(
     // usual two phase accumulation using the React fiber tree to pick up
     // all relevant useEvent and on* prop events.
     let node = event._targetInst;
+    let lastHostComponent;
 
     // Accumulate all instances and listeners via the target -> root path.
     while (node !== null) {
-      // We only care for listeners that are on HostComponents (i.e. <div>)
+      // Handle listeners that are on HostComponents (i.e. <div>)
       if (node.tag === HostComponent) {
+        lastHostComponent = node.stateNode;
         // For useEvent listenrs
         if (enableUseEventAPI && accumulateUseEventListeners) {
           // useEvent event listeners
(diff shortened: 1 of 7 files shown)


packages/react-dom/src/events/accumulateTwoPhaseListeners.js
```diff
@@ -127,6 +135,30 @@ export default function accumulateTwoPhaseListeners(
...
           }
         }
+      } else if (enableScopeAPI && node.tag === ScopeComponent) {
+        const reactScope = node.stateNode.methods;
+        const eventTypeMap = reactScopeListenerStore.get(reactScope);
+        if (eventTypeMap !== undefined) {
+          const type = ((event.type: any): DOMTopLevelEventType);
+          const listeners = eventTypeMap.get(type);
+          if (listeners !== undefined) {
+            const captureListeners = Array.from(listeners.captured);
```
threepointone: btw I noticed we use this in a few places right now, so we should probably add Array.from to this page https://reactjs.org/docs/javascript-environment-requirements.html

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
```diff
@@ -127,6 +135,30 @@ export default function accumulateTwoPhaseListeners(
             dispatchInstances.push(node);
           }
         }
+      } else if (enableScopeAPI && node.tag === ScopeComponent) {
```
threepointone: I'm assuming this logic is from flare's implementation?

trueadm: Yep

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
trueadm: Good point, although we don't need to add it till v17.

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
```diff
@@ -127,6 +135,30 @@ export default function accumulateTwoPhaseListeners(
...
+              const listener = captureListeners[i];
+              const {callback} = listener;
+              dispatchListeners.unshift(callback);
+              dispatchInstances.unshift(((lastHostComponent: any): Element));
+            }
+            for (let i = 0; i < bubbleListeners.length; i++) {
+              const listener = bubbleListeners[i];
+              const {callback} = listener;
+              dispatchListeners.push(callback);
+              dispatchInstances.push(((lastHostComponent: any): Element));
```
gaearon: As far as I can tell, this is what caused the production crashes with today’s sync. This is only set if we “met” a host component in the above traversal before we “met” a Scope. Does anything actually guarantee that? We should probably be more careful with `any` casts in the production codepaths.

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
```diff
@@ -76,11 +82,13 @@ export default function accumulateTwoPhaseListeners(
...
 
     // Accumulate all instances and listeners via the target -> root path.
     while (node !== null) {
-      // We only care for listeners that are on HostComponents (i.e. <div>)
+      // Handle listeners that are on HostComponents (i.e. <div>)
       if (node.tag === HostComponent) {
+        lastHostComponent = node.stateNode;
         // For useEvent listenrs
         if (enableUseEventAPI && accumulateUseEventListeners) {
           // useEvent event listeners
```
gaearon: Hmm actually the bug seems to be here. When the flag is on, we follow this code path, and call getListenersFromTarget on the state node. But for some reason it may be null at this point (unmounted?). I don’t understand why this is possible but it seems to happen from the trace.

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
gaearon: Never mind, see below. I got confused between two code paths. I’m still not sure this was safe though — could the innermost instance be a TextInstance for example, followed by a Scope? I think that could still crash.

trueadm: The target can only ever be a DOM element as you can't have listeners on text nodes.

gaearon: You can't have listeners, but could it be bubbling from a text node?

trueadm: Text nodes can't have children, so there should be no case where we'd ever propagate through into or from a text node.

gaearon: No, I mean something like this:

```
<button>
  <Scope>
    {'hello'}{'world'}
  </Scope>
</button>
```

If I dispatch an event on the text node, won't it trigger this case?

trueadm: How are you dispatching the event? Whilst you can manually create an bubbling event and dispatch it directly on a text node (using the Node.prototype.dispatchEvent), this won't ever occur from an actual real-world event that the user would do.

We only store the host component when it's an actual host component instance (by checking the tag). We don't store `lastHostComponent` for text host instances, which I feel is accurate here.

gaearon: >this won't ever occur from an actual real-world event that the user would do.

I wouldn't be so sure, I remember text nodes firing events in Firefox in some cases. Not saying it should be common but may happen in production.

trueadm: Added support for HostText + test in https://github.com/facebook/react/pull/18441.

PR #18376, opened 2020-03-24
Downgrade Jest to 24

acdlite:
There's a bug where `toEqual` throws when comparing readonly properties, which affects our JSX matchers.

https://github.com/facebook/jest/pull/9575

The bug is fixed in master, but not yet released.

Until then, downgrading to 24.


gaearon:
Not sure it's going to be that easy due to jsdom changes. The upgrade was rather involved: https://github.com/facebook/react/pull/17896.

acdlite:
Wonderful.

acdlite:
@SimenB Do you mind doing a patch release for https://github.com/facebook/jest/pull/9575?

acdlite:
Fortunately cf0081263ced42ddebe0a0d868701890448bfd4b reverts pretty cleanly. I thought it was a big diff but it's mostly only yarn.lock changes.

sebmarkbage:
Did you test if the bug actually fixes all the things? The fix doesn't seem fully correct to me. E.g. what happens if we implement a setter that throws instead of freezing. Which we do sometimes.

acdlite:
It fixes comparing JSX. I tested with `toMatchRenderedOutput`. What are the other things?

acdlite:
Oh you mean the fix in master, not the revert. I didn't test that, no.

sebmarkbage:
Comparing SyntheticEvent might fail. https://github.com/facebook/react/blob/30a998debf79f8c8efb3fa9b6ff5458fb319d7a1/packages/legacy-events/SyntheticEvent.js#L271

SimenB:
>@SimenB Do you mind doing a patch release for facebook/jest#9575?

I don't have publish access, only FB employees do. You can try to ping @scotthovestadt or @DiZy internally? Or @cpojer might be able to

gaearon:
Sorry for inconvenience @SimenB, I just followed up with our team and you should have access to all packages now.

SimenB:
Cool, thanks! I've published 25.2.0 now.

> Did you test if the bugfix actually fixes all the things? The fix doesn't seem fully correct to me. E.g. what happens if we implement a setter that throws instead of freezing. Which we do sometimes.

This probably isn't fixed though, maybe open up an issue with a test that fails which should pass?

gaearon:
Current blocker: https://github.com/facebook/jest/issues/9745

SimenB:
25.2.7 published with fix for ^. Thanks for the help landing it @gaearon!

PR #18378, opened 2020-03-24
Inlined DevTools event emitter impl

bvaughn:
DevTools previously used the NPM [`events` package](https://www.npmjs.com/package/events) for dispatching events. This package has an unfortunate flaw though- if a listener throws during event dispatch, no subsequent listeners are called. I've replaced that event dispatcher with my own implementation that ensures all listeners are called before it re-throws an error.

---

[Here](https://codesandbox.io/s/getcommittree-mtgto) is an example app that breaks the Profiler UI without this change:
```js
const One = () => null;
const Two = () => null;

function App() {
  const [didMount, setDidMount] = useState(false);

  useEffect(() => {
    setDidMount(true);
  }, []);

  return (
    <>
      {didMount ? <Two /> : <One key={1} />}
      <One key={1} />
    </>
  );
}
```

This breaks the Profiler because the invalid `key` causes a DevTools `Store` invariant:
https://github.com/facebook/react/blob/0140118e8e2766abaac2c833775434f83df81d56/packages/react-devtools-shared/src/devtools/store.js#L929-L933

The `Store` happens to be listening to "operations" events ahead of the `ProfilerStore`, so the latter never gets notified of the commit. This causes the commit data to become corrupt.

Screen Shot 2020-03-24 at 10 38 40 AM

With this change in place, both stores are notified before the error is rethrown:

Screen Shot 2020-03-24 at 10 38 09 AM

This might be the underlying cause of a few DevTools errors (e.g. #18033).

### DIFF
--- a/packages/react-devtools-shared/src/events.js
+++ b/packages/react-devtools-shared/src/events.js
@@ -8,7 +8,7 @@
  */
 
 export default class EventEmitter<Events: Object> {
-  listenersMap: Map<string, Set<Function>> = new Map();
+  listenersMap: Map<string, Array<Function>> = new Map();
 
   addListener<Event: $Keys<Events>>(
     event: Event,
@@ -16,12 +16,13 @@ export default class EventEmitter<Events: Object> {
   ): void {
     let listeners = this.listenersMap.get(event);
     if (listeners === undefined) {
-      listeners = new Set();
-
-      this.listenersMap.set(event, listeners);
+      this.listenersMap.set(event, [listener]);
+    } else {
+      const index = listeners.indexOf(listener);
+      if (index < 0) {
+        listeners.push(listener);
+      }
     }
-
-    listeners.add(listener);
   }
 
   emit<Event: $Keys<Events>>(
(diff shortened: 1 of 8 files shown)


packages/react-devtools-shared/src/events.js
```diff
@@ -0,0 +1,65 @@
...
+    listeners.add(listener);
+  }
+
+  emit<Event: $Keys<Events>>(
+    event: Event,
+    ...args: $ElementType<Events, Event>
+  ): void {
+    const listeners = this.listenersMap.get(event);
+    if (listeners !== undefined) {
+      let caughtError = null;
```
gaearon: Nit: this won't work with `throw null`. While unlikely, it is more solid to have a separate `didThrow` boolean, and use that for the rethrow condition.

bvaughn: yeah fair enough 😄

bvaughn:
> I think ideally I'd like to see us completely remove the dependency on this bizarre Node-ish API and instead use something more purpose-built and limited. I guess that's the intent here, to lock it down.

That's what I see this as :smile: I had already only been using a subset of the "events" package API (only the methods I added Flow types for) so I think this change is fairly safe. DevTools also has pretty good end-to-end test coverage which tests the agent+bridge mechanisms.

gaearon:
Do you have confidence there are no cases where we dispatch during a dispatch? Or subscribe during a dispatch? Those are thorny if you don't clone listeners before calling them.

bvaughn:
> Do you have confidence there are no cases where we dispatch during a dispatch? Or subscribe during a dispatch?

Yeah pretty much. DevTools listeners are mostly setup during init and removed during teardown. We don't really add and remove listeners on the fly, except for a couple of very specific one-off cases like [`UnsupportedVersionDialog`](https://github.com/facebook/react/blob/master/packages/react-devtools-shared/src/devtools/views/UnsupportedVersionDialog.js#L42).

> Those are thorny if you don't clone listeners before calling them.

My implementation now clones listeners before calling, FWIW. I added that after your previous comment because it seemed like a good precaution.

bvaughn:
Bottom line is: I'm comfortable with this change. It fixes a known serious problem. If it turns out to introduce another, we can always revert :smile:

gaearon:
Haha sure let's go. :-)

Issue #14551, opened 2019-01-08
Cursor jumps with controlled input with email type


cdbean:
**Do you want to request a *feature* or report a *bug*?**
a bug

**What is the current behavior?**

With an email input (`<input type="email" />`), the cursor jumps when there is space. Try the demo https://codesandbox.io/s/y7y93p2zx?moduleview=1. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

![jan-08-2019 15-25-44](https://user-images.githubusercontent.com/578035/50865307-0ac24900-135a-11e9-81d3-9eb05c937533.gif)

**What is the expected behavior?**
The cursor should remain in the right position as user types

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
React 16
Chrome and Safari, not firefox

cdbean:
Seems it is a known issue https://twitter.com/kentcdodds/status/1082705003955933184. I'll probably close this

gaearon:
@imdongchen For the future, please don't close issues because somebody reported them on Twitter. GitHub is where we track them, not Twitter. :-)

gaearon:
Tracking this in https://github.com/facebook/react/issues/15418.

PR #18379, opened 2020-03-24
Fix email cursor jump

### LINKED ISSUE #15418: Field type="email" with multiple attribute cursor jumps to start

bl00mber:
fix #15418, #14551

Version of React: 16.13.1
Browser: Chrome 80

**Previous behavior:**
After deleting of second email separated by space cursor jumps to the beginning.
When typing a space in an empty email input and then literal, the space is deleted and cursor jumps to the beginning.

email-jump

**Altered behavior:**
Cursor not moving.

email-fixed

### DIFF
--- a/fixtures/dom/src/components/Header.js
+++ b/fixtures/dom/src/components/Header.js
@@ -71,6 +71,7 @@ class Header extends React.Component {
                 <option value="/text-inputs">Text Inputs</option>
                 <option value="/number-inputs">Number Input</option>
                 <option value="/password-inputs">Password Input</option>
+                <option value="/email-inputs">Email Input</option>
                 <option value="/selects">Selects</option>
                 <option value="/textareas">Textareas</option>
                 <option value="/input-change-events">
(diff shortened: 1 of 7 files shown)


gaearon:
I've changed my mind. I think this is possibly more risky. See my comment in https://github.com/facebook/react/pull/18459. We should definitely fix it but as a part of a more coordinated effort when we can squash many bugs at once.

PR #18380, opened 2020-03-24
Fix a mistake in ReactChildren refactor

gaearon:
Of course I broke something in https://github.com/facebook/react/pull/18332. Luckily internal tests caught it.

>"there's probably at least not many bugs." — @sebmarkbage

I changed signature of one of the internal functions, but forgot to adjust the callsite. So there was a missing argument. **The fix for that is https://github.com/facebook/react/commit/20fc0bd3d8bac9c3869fba55ad3cf48996bdba8e.**

To prevent this from happening again and to turn my mistake into a learning experience, I added:

1. Regression test for `map` returning an array — we didn't have that. https://github.com/facebook/react/commit/344e56c89273c7b01da78467ddc88b63328cab8e
2. Flow types. https://github.com/facebook/react/commit/f66dc559599e3f4e7b786f4937a370296c55e844

### DIFF
--- a/packages/react/src/ReactChildren.js
+++ b/packages/react/src/ReactChildren.js
@@ -111,7 +111,7 @@ function mapIntoArray(children, array, escapedPrefix, nameSoFar, callback) {
       if (childKey != null) {
         escapedChildKey = escapeUserProvidedKey(childKey) + '/';
       }
-      mapIntoArray(mappedChild, array, escapedChildKey, c => c);
+      mapIntoArray(mappedChild, array, escapedChildKey, '', c => c);
     } else if (mappedChild != null) {
       if (isValidElement(mappedChild)) {
         mappedChild = cloneAndReplaceKey(
--- a/packages/react/src/ReactChildren.js
+++ b/packages/react/src/ReactChildren.js
@@ -3,7 +3,11 @@
  *
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
  */
+
+import type {ReactNodeList} from 'shared/ReactTypes';
 
 import invariant from 'shared/invariant';
 import {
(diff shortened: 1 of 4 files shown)


packages/react/src/ReactChildren.js
```diff
@@ -46,7 +50,7 @@ function escape(key) {
 let didWarnAboutMaps = false;
 
 const userProvidedKeyEscapeRegex = /\/+/g;
-function escapeUserProvidedKey(text) {
+function escapeUserProvidedKey(text: string): string {
   return ('' + text).replace(userProvidedKeyEscapeRegex, '$&/');
```
gaearon: I don't think this coercion is necessary because reading the key *from* an element always gives you a string (or null). But I decided not to touch this.

packages/react/src/ReactChildren.js
```diff
@@ -111,16 +121,18 @@ function mapIntoArray(children, array, escapedPrefix, nameSoFar, callback) {
       if (childKey != null) {
         escapedChildKey = escapeUserProvidedKey(childKey) + '/';
       }
-      mapIntoArray(mappedChild, array, escapedChildKey, c => c);
+      mapIntoArray(mappedChild, array, escapedChildKey, '', c => c);
```
gaearon: This is the fix. Oops.

packages/react/src/ReactChildren.js
```diff
@@ -25,7 +29,7 @@ const SUBSEPARATOR = ':';
  * @param {string} key to be escaped.
  * @return {string} the escaped key.
  */
-function escape(key) {
+function escape(key: any): string {
```
trueadm: Why is key any?

sebmarkbage: Yea, I'd type the other side, that is the broken one, as any. In this case `component: mixed` should maybe be `component: any` at the callsite.

sebmarkbage:
lol at that branch name.

Issue #18357, opened 2020-03-20
Bug: High-pri setState causes primary tree to get unhidden


gaearon:
Repro: https://codesandbox.io/s/fast-water-4i4jb

Notice the flash. I'd expect `suspended` to be `true` during the high pri render.

gaearon:
Repro with two sibling components: https://codesandbox.io/s/vigilant-nightingale-deb0u. During the high pri render, we don't go into ProfilePage at all.

PR #18384, opened 2020-03-25
[Bugfix] Dropped updates inside a suspended tree

### LINKED ISSUE #18353: Bug: Updates in the primary tree only unsuspend once
https://codesandbox.io/s/upbeat-frost-3tt25

1. Press Suspend, see "Loading..."
2. Wait five seconds, you'll see "Happy birthday" at some point 
3. You won't see it again no matter how long you wait

Expected: you'll see it every five seconds.

acdlite:
Fixes #18353.

When there are multiple updates at different priority levels inside a suspended subtree, all but the highest priority one is dropped after the highest one suspends.

We do have tests that cover this for updates that originate outside of the Suspense boundary, but not for updates that originate inside.

I'm surprised it's taken us this long to find this issue, but it makes sense in that transition updates usually originate outside the boundary or "seam" of the part of the UI that is transitioning.

---

~~This is related to https://github.com/facebook/react/issues/18353 but it only fixes part of it. There's another bug, which I believe is related to `setState` traversal skipping over the primary fragment fiber. Still working on that one.~~ Update: yep, that was it. Pushed another commit that and confirmed that it fixes the sandbox in #18353. CodeSandbox CI is currently down, but when it's back up I'll post an updated sandbox.

Also related to https://github.com/facebook/react/issues/18357, but likewise only fixes part of it.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -98,8 +98,6 @@ import {
 } from './ReactUpdateQueue';
 import {
   NoWork,
-  Idle,
-  ContinuousHydration,
   Never,
   Sync,
   computeAsyncExpiration,
@@ -181,6 +179,7 @@ import {
   scheduleUpdateOnFiber,
   renderDidSuspendDelayIfPossible,
   markUnprocessedUpdateTime,
+  getWorkInProgressRoot,
 } from './ReactFiberWorkLoop';
 
 const ReactCurrentOwner = ReactSharedInternals.ReactCurrentOwner;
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1590,6 +1592,42 @@ function shouldRemainOnFallback(
...
+  if ((workInProgress.mode & BlockingMode) !== NoMode) {
+    // The highest priority remaining work is part of this render. Since we only
+    // keep track of the highest level, we don't know if there's a lower
+    // priority level scheduled. As a compromise, we'll render at a really low
+    // priority that includes all the updates.
+    // TODO: If expirationTime were a bitmask where each bit represents a
+    // separate task thread, this would be: currentChildBits & ~renderBits
+    // TODO: We used to track the lowest pending level for the whole root, but
+    // we removed it to simplify the implementation. It might be worth adding
+    // it back for this use case, to avoid redundant passes.
```
acdlite: @sebmarkbage I think I'm going to do this

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -3006,6 +3064,42 @@ function beginWork(
...
+              // fiber is not part of the return path of the children, so when
+              // an update happens, the fragment doesn't get marked during
+              // setState. This is something we should consider addressing when
+              // we refactor the Fiber data structure. (There's a test with more
+              // details; to find it, comment out the following block and see
+              // which one fails.)
+              //
+              // As a workaround, we need to recompute the `childExpirationTime`
+              // by bubbling it up from the next level of children. This is
+              // based on similar logic in `resetChildExpirationTime`.
```
acdlite: An alternative fix would be to detect this during `setState`. I thought it might be better to localize the fix here since it's relatively rare but I'm having second thoughts.

gaearon:
Is it worth adding tests for similar scenarios but with updates originating inside the fallback just in case?

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: Instead of tracking it globally, why won't we just track it on this suspense boundary in the state?

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: I think this is the right call but I think we should just bite the bullet have have Suspense boundaries always have these fragments even when not suspended. It's too much code to manage (buggy and byte size). Especially with enter/exit animations it's going to get even trickier.

It adds a bit more memory but there's not that many suspense boundaries and the ones that do exist fairly often transition through this state anyway.

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: I'm going to track the "suspended time" locally but to track the "last pending time" locally I'd have to add stuff to the `setState` path. Doesn't seem worth it since the bitmask version will be much simpler.

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: That would definitely simplify things a lot

acdlite:
Added a fallback test case

PR #18386, opened 2020-03-25
feat(createContext): Include displayName in warning

eps1lon:
## Summary

[Seems like this is not only used internally at facebook](https://twitter.com/kentcdodds/status/1242678976645959681). I included the `displayName` that should be set when warning to make it more obvious where the warning is coming from.

This also explicitly documents the strategy by which this warning is deduped.

## Test Plan

- adjust test
- check usefulness in https://codesandbox.io/s/displayname-contextconsumer-warning-it2e8

```
react.development.js:319 Warning: Setting `displayName` on Context.Consumer has no effect. You should set it directly on the context with Context.displayName = 'Location.Consumer'.
printWarning @ react.development.js:319
react.development.js:319 Warning: Setting `displayName` on Context.Consumer has no effect. You should set it directly on the context with Context.displayName = 'Route.Consumer'.
```
Looks more useful to me.

### DIFF
--- a/packages/react/src/ReactContext.js
+++ b/packages/react/src/ReactContext.js
@@ -125,11 +125,12 @@ export function createContext<T>(
         get() {
           return context.displayName;
         },
-        set() {
+        set(displayName) {
           if (!hasWarnedAboutDisplayNameOnConsumer) {
             console.warn(
               'Setting `displayName` on Context.Consumer has no effect. ' +
-                "You should set it directly on the context with Context.displayName = 'NamedContext'.",
+                "You should set it directly on the context with Context.displayName = '%s'.",
+              displayName,
             );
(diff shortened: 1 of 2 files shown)


gaearon:
>Seems like this is not only used internally at facebook.

I don't know what you mean by this. Yes, it's a public warning. It concretely says that doing this is unsupported and you should assign to the context itself instead.

gaearon:
Makes sense.

eps1lon:
> It concretely says that doing this is unsupported and you should assign to the context itself instead.

It was just in reference to the original PR that had to be reverted because it broke tests at facebook internally (https://github.com/facebook/react/pull/18223#issue-384354627). I thought it wasn't used by others because it didn't have any effect.

PR #18388, opened 2020-03-25
Add getInspectorDataForViewAtPoint (take two)

rickhanlonii:
## Summary

**This PR relands https://github.com/facebook/react/pull/18233 with feedback applied.**

- Adds `rendererConfig` to the DevTool config for renderer-specific dev functions.
- Moves `getInspectorDataForViewTag` to `rendererConfig` for React Native only.
- Add `getInspectorDataForViewAtPoint` to the React Native host configs.

This new method provides an interface to return inspector data about the view touched at a given point. This allows the React Native inspector to ask React for information about a view at an x,y touch point without any knowledge of how each renderer gets the information.

## Implementation

**Paper**
For Paper we've just moved the logic from the [InspectorOverlay](https://github.com/facebook/react-native/blob/c3d072955024824d7ff6113fc9f642d25c462f16/Libraries/Inspector/InspectorOverlay.js#L40-L50) into React Core. This will allow us to delete `getInspectorDataForViewTag` called [here](https://github.com/facebook/react-native/blob/c3d072955024824d7ff6113fc9f642d25c462f16/Libraries/Inspector/Inspector.js#L61) which is necessary since Fabric does not support React tags.

**Fabric**
For Fabric we add a new strategy which queries native for a node using `findNodeAtPoint`, passes the shadow node to `measure`, and then asks React for the fiber information.

## Test Plan

- Manually tested in Fabric and Paper for React Native.

### DIFF
--- a/packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
+++ b/packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
@@ -33,19 +33,19 @@ import {
 class ReactNativeFiberHostComponent {
   _children: Array<Instance | number>;
   _nativeTag: number;
-  _internalFiberInstanceHandle: Object;
+  _internalFiberInstanceHandleDEV: Object;
   viewConfig: ReactNativeBaseComponentViewConfig<>;
 
   constructor(
     tag: number,
     viewConfig: ReactNativeBaseComponentViewConfig<>,
-    internalInstanceHandle: Object,
+    internalInstanceHandleDEV: Object,
   ) {
     this._nativeTag = tag;
     this._children = [];
     this.viewConfig = viewConfig;
     if (__DEV__) {
-      this._internalFiberInstanceHandle = internalInstanceHandle;
+      this._internalFiberInstanceHandleDEV = internalInstanceHandleDEV;
     }
   }
 
--- a/packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
+++ b/packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
@@ -44,7 +44,9 @@ class ReactNativeFiberHostComponent {
     this._nativeTag = tag;
     this._children = [];
     this.viewConfig = viewConfig;
-    this._internalFiberInstanceHandle = internalInstanceHandle;
+    if (__DEV__) {
+      this._internalFiberInstanceHandle = internalInstanceHandle;
+    }
   }
 
   blur() {
--- a/packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
+++ b/packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
@@ -33,12 +33,18 @@ import {
 class ReactNativeFiberHostComponent {
   _children: Array<Instance | number>;
   _nativeTag: number;
+  _internalFiberInstanceHandle: Object;
   viewConfig: ReactNativeBaseComponentViewConfig<>;
 
-  constructor(tag: number, viewConfig: ReactNativeBaseComponentViewConfig<>) {
+  constructor(
+    tag: number,
+    viewConfig: ReactNativeBaseComponentViewConfig<>,
+    internalInstanceHandle: Object,
+  ) {
     this._nativeTag = tag;
     this._children = [];
     this.viewConfig = viewConfig;
+    this._internalFiberInstanceHandle = internalInstanceHandle;
   }
 
   blur() {
--- a/packages/react-native-renderer/src/ReactNativeFiberInspector.js
+++ b/packages/react-native-renderer/src/ReactNativeFiberInspector.js
@@ -8,6 +8,7 @@
  */
 
 import type {Fiber} from 'react-reconciler/src/ReactFiber';
+import type {TouchedViewDataAtPoint, InspectorData} from './ReactNativeTypes';
 
 import {
   findCurrentHostFiber,
@@ -27,6 +28,7 @@ if (__DEV__) {
 }
 
 let getInspectorDataForViewTag;
+let getInspectorDataForViewAtPoint;
 
 if (__DEV__) {
   const traverseOwnerTreeUp = function(hierarchy, instance: any) {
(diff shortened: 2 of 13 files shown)


packages/react-native-renderer/src/ReactNativeFiberInspector.js
```diff
@@ -80,15 +82,68 @@ if (__DEV__) {
...
+          measure: callback => {
+            // If this is Fabric, we'll find a ShadowNode and use that to measure.
+            const hostFiber = findCurrentHostFiber(fiber);
+            const shadowNode =
+              hostFiber != null &&
+              hostFiber.stateNode !== null &&
+              hostFiber.stateNode.node;
+
+            if (shadowNode) {
+              nativeFabricUIManager.measure(shadowNode, function(
```
TheSavior: Could you pass the callback directly here?

packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
```diff
@@ -33,12 +33,18 @@ import {
...
-  constructor(tag: number, viewConfig: ReactNativeBaseComponentViewConfig<>) {
+  constructor(
+    tag: number,
+    viewConfig: ReactNativeBaseComponentViewConfig<>,
+    internalInstanceHandle: Object,
+  ) {
     this._nativeTag = tag;
     this._children = [];
     this.viewConfig = viewConfig;
+    this._internalFiberInstanceHandle = internalInstanceHandle;
```
gaearon: Is this only used in DEV? If yes, please call it `this._internalFiberInDEV` or something and make it DEV-only to avoid using an extra field.

TheSavior: We have the same field in prod (used in prod) on the fabric host component. Do you feel like making it dev is the right choice knowing it’s on the fabric host component too? It seems reasonable to make it dev only since it is only needed in dev afaik but just checking.

gaearon: If we don't use some field in PROD (whether or not it exists on other places), it's generally a good practice to make it DEV only to reduce the memory usage. If only due to the extra pointer.

packages/react-native-renderer/src/ReactNativeFiberInspector.js
```diff
@@ -117,13 +172,90 @@ if (__DEV__) {
...
+          );
+          callback({
+            ...inspectorData,
+            pointerY: locationY,
+            frame: {left, top, width, height},
+            touchedViewTag: nativeViewTag,
+          });
+        },
+      );
+    } else if (__DEV__) {
```
bvaughn: nit: This whole function is inside of an if-DEV check, so this is redundant.

packages/react-native-renderer/src/ReactNativeTypes.js
```diff
@@ -100,6 +100,46 @@ type SecretInternalsType = {
...
+  props: InspectorDataProps,
+  source: InspectorDataSource,
+|}>;
+
+export type InspectorData = $ReadOnly<{|
+  hierarchy: Array<{|
+    name: ?string,
+    getInspectorData: InspectorDataGetter,
+  |}>,
+  selection: ?number,
```
bvaughn: What does this value represent? Just curious.

rickhanlonii: It's `fiberHierarchy.indexOf(instance);`, I believe it's used to show which component is inspected in the hierarchy.

bvaughn: Yeah, I saw that's what it was being assigned to, I just ... didn't understand its purpose 😅and the name was kind of confusing.

rickhanlonii: Updated to `selectedIndex` 👍

packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
```diff
@@ -44,7 +44,9 @@ class ReactNativeFiberHostComponent {
     this._nativeTag = tag;
     this._children = [];
     this.viewConfig = viewConfig;
-    this._internalFiberInstanceHandle = internalInstanceHandle;
+    if (__DEV__) {
+      this._internalFiberInstanceHandle = internalInstanceHandle;
```
gaearon: Please include `DEV` in the name so no one is tempted to use this in prod.

rickhanlonii: Done thanks 👍

gaearon: I still see the old name?

packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
```diff
@@ -33,12 +33,20 @@ import {
...
+  constructor(
+    tag: number,
+    viewConfig: ReactNativeBaseComponentViewConfig<>,
+    internalInstanceHandle: Object,
+  ) {
     this._nativeTag = tag;
     this._children = [];
     this.viewConfig = viewConfig;
+    if (__DEV__) {
+      this._internalFiberInstanceHandle = internalInstanceHandle;
```
gaearon: Still needs rename here.

packages/react-reconciler/src/ReactFiberReconciler.old.js
```diff
@@ -107,10 +108,7 @@ type DevToolsConfig = {|
   // Note: this actually *does* depend on Fiber internal fields.
   // Used by "inspect clicked DOM element" in React DevTools.
   findFiberByHostInstance?: (instance: Instance | TextInstance) => Fiber | null,
-  // Used by RN in-app inspector.
-  // This API is unfortunately RN-specific.
-  // TODO: Change it to accept Fiber instead and type it properly.
-  getInspectorDataForViewTag?: (tag: number) => Object,
+  rendererConfig?: RendererInspectionConfig,
```
gaearon: You also need to change `ReactFiberReconciler.new.js`.

packages/react-native-renderer/src/ReactNativeFiberHostComponent.js
rickhanlonii: I didn't save the file, pushed

packages/react-reconciler/src/ReactFiberReconciler.old.js
rickhanlonii: Is that still in a branch? I dont see it on master?

gaearon: Ignore me. You're right.

PR #18391, opened 2020-03-25
Formalize the Wakeable and Thenable types

sebmarkbage:
We use two subsets of Promises throughout React APIs. This introduces the smallest subset - Wakeable. It's the thing that you can throw to suspend. It's something that can ping.

I also use a shared type for Thenable in the cases where we expect a value so we can be a bit more rigid with our us of them.

I also made Flight use fake Promises as the Wakeable that it throws.

### DIFF
--- a/packages/react-client/src/ReactFlightClient.js
+++ b/packages/react-client/src/ReactFlightClient.js
@@ -40,48 +40,62 @@ const PENDING = 0;
 const RESOLVED = 1;
 const ERRORED = 2;
 
-const CHUNK_TYPE = Symbol('flight.chunk');
+type PendingChunk = {
+  _status: 0,
+  _value: null | Array<() => mixed>,
+  then(resolve: () => mixed): void,
+};
+type ResolvedChunk<T> = {
+  _status: 1,
+  _value: T,
+  then(resolve: () => mixed): void,
+};
+type ErroredChunk = {
+  _status: 2,
+  _value: Error,
+  then(resolve: () => mixed): void,
+};
+type SomeChunk<T> = PendingChunk | ResolvedChunk<T> | ErroredChunk;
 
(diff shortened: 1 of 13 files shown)


packages/shared/ReactTypes.js
```diff
@@ -234,3 +234,21 @@ export type MutableSource<Source: $NonMaybeType<mixed>> = {|
   _currentPrimaryRenderer?: Object | null,
   _currentSecondaryRenderer?: Object | null,
 |};
+
+// The subset of a Thenable required by things thrown by Suspense.
+// This doesn't require a value to be passed to either handler.
+export interface Wakeable {
```
sebmarkbage: Thenable is called that way because you can call then on it.

Maybe this should be Waker or something.

sebmarkbage: I guess in Rust, a Waker is the thing you call to then (essentially). It's the thing to be woken up.

So I guess Wakeable is fine. It's the thing that will at some point trigger to wake up the Waker.

packages/react-client/src/ReactFlightClient.js
```diff
@@ -160,16 +165,19 @@ export function reportGlobalError<T>(
...
     return maybeChunk;
   }
-  let chunk: Chunk<T> = (maybeChunk: any);
-  if (chunk.status === RESOLVED) {
-    return chunk.value;
+  let chunk: SomeChunk<T> = (maybeChunk: any);
+  if (chunk._status === RESOLVED) {
+    return chunk._value;
+  } else if (chunk._status === PENDING) {
+    // eslint-disable-next-line no-throw-literal
```
sebmarkbage: Btw this seems like a bug in the no-throw-literal lint. It gets confused by the Flow annotation.

PR #18395, opened 2020-03-26
ReactDOM.useEvent: enable on internal www and add inspection test

trueadm:
This PR adds an inspection Jest test and also enables the `enableUseEventAPI` for internal FB www usage. This also fixes a small compatibility issue with `beforeblur` and React Flare, that meant we were previously incorrectly listening for a native `beforeblur` event to fire on the DOM. We shouldn't be listening for a native event as we emulate React Flare events.


packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -327,11 +327,11 @@ export function isManagedDOMElement(
 export function isValidEventTarget(
   target: EventTarget | ReactScopeMethods,
 ): boolean {
-  return typeof (target: any).addEventListener === 'function';
+  return typeof (target: Object).addEventListener === 'function';
 }
 
 export function isReactScope(target: EventTarget | ReactScopeMethods): boolean {
-  return typeof (target: any).getChildContextValues === 'function';
+  return typeof (target: Object).getChildContextValues === 'function';
```
bvaughn: I'm curious. What benefit does this change provide?

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1348,6 +1348,10 @@ export function listenToEventResponderEventTypes(
       const targetEventType = isPassive
         ? eventType
         : eventType.substring(0, eventType.length - 7);
+      // We don't listen to this as we actually emulate it in the host config
+      if (targetEventType === 'beforeblur') {
```
bvaughn: Seems kind of unfortunate to have a specific event type hard-coded in this loop like this. Do you expect we may be adding more checks like this one over time?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: Looks nicer on the eyes. I've also had a lot of lemsip today, so it may have something to do with it too.

packages/react-dom/src/client/ReactDOMComponent.js
trueadm: We will be killing Flare soon, so this should all go away. Mwhuhaha.

PR #18397, opened 2020-03-26
Remove "es6-symbol" package from DevTools

bvaughn:
This polyfill was copied over from DevTools v3 but given the [current browser compatibility for `Symbol`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#Browser_compatibility), I think it's safe to remove.

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/HooksTree.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/HooksTree.js
@@ -124,7 +124,7 @@ function HookView({canEditHooks, hook, id, inspectPath, path}: HookViewProps) {
         hook !== null &&
         typeof hook === 'object' &&
         hook.hasOwnProperty(meta.type)
-          ? hook[meta.type]
+          ? hook[(meta.type: any)]
           : typeof value,
     },
     id: 'SelectedElement',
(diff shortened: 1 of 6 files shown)


packages/react-devtools-shared/src/devtools/views/Components/HooksTree.js
```diff
@@ -124,7 +124,7 @@ function HookView({canEditHooks, hook, id, inspectPath, path}: HookViewProps) {
         hook !== null &&
         typeof hook === 'object' &&
         hook.hasOwnProperty(meta.type)
-          ? hook[meta.type]
+          ? hook[(meta.type: any)]
```
trueadm: Why was this change needed?

bvaughn: Flow complains otherwise:
> Cannot access computed property using Symbol

bvaughn:
> Thanks so much for cleaning this up! Do you mind deploying a new version of the devtools into xplat/js after merging this?

Landing things in xplat can be a hassle though, and this isn't super high priority for me at the moment. I'll take a shot at it though, assuming it doens't take *too* long.

PR #18401, opened 2020-03-27
Rules of hooks ignore list

OlafMerkert:
## Summary

For historic reasons, our project contains a couple of very widely used functions with names that the heuristics of the `rules-of-hooks` linter rule falsely consider to be hooks.

- Renaming these functions could be an alternative option, but would incur the cost of a major breaking change in our project. And we would have to find new names ...
- Adding eslint ignores would likewise be a significant effort, and we would have to add those ignores to new code as well.

The option added with this pull request would allow us to enable the checks by `rules-of-hooks`, configuring the known exceptions in a single place -- from where we could also easily remove them if we later decide to rename the offending functions.

## Test Plan

I added test cases to check that only function names on the ignore list are actually ignored.

If you feel more test cases are needed, I would be happy to add them if you give me a pointer on what still needs to be covered.

### DIFF
--- a/packages/eslint-plugin-react-hooks/README.md
+++ b/packages/eslint-plugin-react-hooks/README.md
@@ -51,6 +51,24 @@ Please refer to the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) 
 
 For feedback about the `exhaustive-deps` rule, please post in [this thread](https://github.com/facebook/react/issues/14920).
 
+## Options
+
+If your codebase contains functions with `use` prefix which are not React Hooks, and you are not able to rename them, you can configure `react-hooks/rules-of-hooks` to ignore them with the `ignoredNames` option:
+
+```js
+{
+  "plugins": [
+    // ...
+    "react-hooks"
+  ],
+  "rules": {
+    // ...
+    "react-hooks/rules-of-hooks": ["error", {ignoredNames: ["useThisIsNotAHook"]}],
+    "react-hooks/exhaustive-deps": "warn"
+  }
+}
+```
+
 ## License
 
 MIT
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
@@ -378,6 +378,13 @@ const tests = {
         const [myState, setMyState] = useState(null);
       }
     `,
+    {
+      code: `
+             // Valid because function name is on the whitelist
+             useNotAHook()
+      `,
+      options: [{ignoredNames: ['useNotAHook']}],
+    },
   ],
   invalid: [
     {
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
@@ -381,9 +381,23 @@ const tests = {
     {
       code: `
              // Valid because function name is on the whitelist
-             useNotAHook()
+             useNotAHook();
+             useNotAHookEither();
       `,
-      options: [{ignoredNames: ['useNotAHook']}],
+      options: [{ignoredNames: ['useNotAHook', 'useNotAHookEither']}],
+    },
+    {
+      code: `
+             // Valid because function name is on the whitelist
+             const buildSomething = () => {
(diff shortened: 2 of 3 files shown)


gaearon:
@OlafMerkert 

>Renaming these functions could be an alternative option, but would incur the cost of a major breaking change in our project. And we would have to find new names ...

Can you tell me more about why it would be a breaking change? Is this part of some public API?

OlafMerkert:
@gaearon Thank you for your fast response!

The "problematic" functions are not part of a public API in a strict sense, they are used and defined in just one git repository and even just one npm package.

They are part of the public API of our platform layer which is used throughout the code base. However, different parts of the code base are owned by different teams, so any changes to the platform API need to be communicated and coordinated.

It's less than 10 functions affected, and their total usages are less than 2000, so renaming would be an option just from a technical point of view. Unfortunately we have to support multiple release branches, and are a bit afraid of merge conflicts that could be caused by applying the renaming.

I hope this answers your question.

gaearon:
Here's my suggestion.

Instead of renaming the functions, you can introduce aliases. E.g. export both the old name like `useSomething` and a new name like `shouldUseSomething` (or whatever matches your use case). Then your client repos can upgrade at their own pace, adopt the lint rule, and once they have violations, instead of suppressing those they can switch to using the new export. With find-and-replace this shouldn't take any effort, and with gradual rollout you don't have to do this in all repos at once.

Eventually you won't have usages of the old methods and you can delete them. Or you can do like React does, and print a warning to the console for a while from them before deleting.

Now you might ask. Isn't this way more complicated than just adding an option? I'd like to not add this option because the whole value of this rule is in establishing a convention. A convention is only as strong as its following in practice. If people have an escape hatch from following the convention then it will gradually become more relaxed. There won't be enough pressure on existing code that doesn't adhere to it to get fixed. It would be too easy to add exceptions for new code clashing with the convention.

Yes, there is some resistance and effort in this. I think this effort is valuable because it makes the convention strong and ultimately allows you to say with confidence: "if this thing starts with `use`, I know it's a Hook". There is some cost for getting there but I think it's worth it. So my objection isn't technical. I just want to ensure that if you *do* use this lint rule, the convention doesn't get undermined.

As a last resort, you can always fork the lint rule.

Hope that makes some sense.

gaearon:
>Unfortunately we have to support multiple release branches, and are a bit afraid of merge conflicts that could be caused by applying the renaming.

Btw I empathize with that. But I think it's a more general problem you might need to find a solution for. E.g. React sometimes renames its exports and you have to run a codemod. So you need a strategy going forward regardless for doing sweeping atomic changes.

OlafMerkert:
Well, it was worth a try. 
Thank you for your extensive feedback, the idea with the alias is pretty nice, though it's probably overkill in our case. Anyway we should have plenty of options to choose from for our codebase.

It seems that my understanding of what the hook convention is was a bit off.
I understood the convention to just say that "every hook function should have the `use` prefix". But from your response, it seems that it also says "every function with the `use` prefix should be a hook". 

That second part feels a bit odd to me personally, because that means one library makes an exclusive and general claim on a function prefix.

gaearon:
>I understood the convention to just say that "every hook function should have the use prefix". But from your response, it seems that it also says "every function with the use prefix should be a hook".

Even if this wasn't a part of the convention, the first time you'd try to use a non-Hook `use` function in your render, it would become a problem. So you would have to resolve it somehow anyway. As you said, naming changes are costly. This is why we think it's better to enforce this right away and before it's a pattern that spreads around the codebase.

>That second part feels a bit odd to me personally, because that means one library makes an exclusive and general claim on a function prefix.

I hear you! I think conceptually in the React-land, you can think of `use` as almost a syntax. Maybe someday it will be. In that sense, it does make sense that in React apps it has a special meaning.

For better or worse, for now we have to stick with a convention. But once you have a convention, any deviation is confusing anyway. E.g. whenever you see a `use` call you now have to be paranoid if it's not in render scope. Is it a Hook or not? That's what the convention is for. It's designed to help you answer this question unambigously.

I wouldn't say React makes an "exclusive" claim here. You don't have to enable the lint rule, and it's not part of generic JS lint rule sets. Just a React specific one. For example, Vue might also choose to use the `use` convention, but without the lint rule. I agree it's confusing for mixed projects. But even there, you can turn lint rules on and off based on the folder or file extension. So this isn't a big problem either.

Hope that makes sense.

PR #18405, opened 2020-03-27
Refactor Enter/Leave listener accumulation

trueadm:
We already cleaned up and refactored two phase event accumulation in a previous PR, this PR cleans up and refactors the Enter/Leave logic so it's easier to follow and adapt for future work where we can refactor how events are accumulated and ties to synthetic events. This should also have a slight runtime performance win as we don't accumulate as many objects with this implementation.

I also inlined all the rest of the logic into RN and ResponderPlugin, as we no longer need a bunch of code for ReactDOM.


packages/react-dom/src/events/accumulateEnterLeaveListeners.js
```diff
@@ -0,0 +1,139 @@
...
+    return inst;
+  }
+  return null;
+}
+
+/**
+ * Return the lowest common ancestor of A and B, or null if they are in
+ * different trees.
+ */
+function getLowestCommonAncestor(instA: Fiber, instB: Fiber): Fiber | null {
```
necolas: Btw you inlined these functions in the responder event plugins too. They used to be in the tree traversal module

trueadm: I altered the PR to inline all this code into their relevant places so we don't need them for ReactDOM.

packages/legacy-events/ResponderEventPlugin.js
```diff
@@ -158,6 +151,98 @@ const eventTypes = {
...
+    instB = getParent(instB);
+  }
+  return false;
+}
+
+/**
+ * Return the parent instance of the passed-in instance.
+ */
+export function getParentInstance(inst) {
+  return getParent(inst);
```
gaearon: This function doesn’t do much

trueadm:
This also slightly reduces the prod size of ReactDOM by the looks of it (although not by much, 0.1%). @gaearon @necolas let me know if there's anything else that sticks out to you both please?

gaearon:
Let’s wait a bit before we figure out the bug in the recent event commit. Caught it after a sync.

packages/legacy-events/ResponderEventPlugin.js
```diff
@@ -158,6 +151,91 @@ const eventTypes = {
 // Start of inline: the below functions were inlined from
 // EventPropagator.js, as they deviated from ReactDOM's newer
 // implementations.
+
+function getParent(inst) {
```
threepointone, line 155: I get that this is 'legacy', but it's weird that fibers are called instances here (since we use that word for something else elsewhere. 

could you explain *why* this other traversal fns are inlined multiple times instead of in one module? Also why do some of them have types, and some don't?

trueadm: They're inlined because we don't want them in ReactDOM anymore and by pulling them out, it means we won't accidently break another module by changing something else. They're essentially legacy, and in a bunch of the cases, will likely be killed when their associated module drops them. Some have types because the module is Flow types, whilst before they came from a module that wasn't Flow typed.

PR #18407, opened 2020-03-27
Move accumulateEventTargetListeners to its own module/function

trueadm:
The logic for `accumulateEventTargetListeners` should really be in its own function/module and not coupled up in the `accumulateTwoPhaseListeners` logic – making it all much easier to follow.


threepointone:
boo codesandbox boo

PR #18409, opened 2020-03-27
Don't show destroy function state update warning when updating ancestors

bvaughn:
React can't directly detect a memory leak, but there are some clues that warn about one. One of these clues is when an unmounted React component tries to update its state. For example, if a component forgets to remove an event listener when unmounting, that listener may be called later and try to update state, at which point React would warn about the potential leak.

Warning signals like this are more useful if they're strong. For this reason, it's good to always avoid updating state from inside of an effect's cleanup function. Even when you know there is no potential leak, React has no way to know and so it will warn anyway.

In most cases we suggest moving state updates to the useEffect() body instead (to avoid triggering the warning). This works so long as the component is updating its own state (or the state of a descendant). However this will not work when a component updates its parent state in a cleanup function. If such a component is unmounted but its parent remains mounted, the state will be incorrect. For this reason, we now avoid showing the warning if a component is updating an ancestor.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -161,6 +161,7 @@ import {
 import getComponentName from 'shared/getComponentName';
 import ReactStrictModeWarnings from './ReactStrictModeWarnings';
 import {
+  current as currentDebugFiberInDEV,
   isRendering as ReactCurrentDebugFiberIsRenderingInDEV,
   resetCurrentFiber as resetCurrentDebugFiberInDEV,
   setCurrentFiber as setCurrentDebugFiberInDEV,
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2822,14 +2823,36 @@ function warnAboutUpdateOnUnmountedFiberInDEV(fiber) {
...
+      // In most cases we suggest moving state updates to the useEffect() body instead.
+      // This works so long as the component is updating its own state (or the state of a descendant).
+      //
+      // However this will not work when a component updates its parent state in a cleanup function.
+      // If such a component is unmounted but its parent remains mounted, the state will be out of sync.
+      // For this reason, we avoid showing the warning if a component is updating an ancestor.
+      let current = fiber;
+      while (current !== null) {
+        // currentDebugFiberInDEV owns the effect destroy function currently being invoked.
+        if (currentDebugFiberInDEV === current) {
```
gaearon: Do you need to compare to alternate too?

bvaughn: I'm not certain. Probably?

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2822,14 +2823,39 @@ function warnAboutUpdateOnUnmountedFiberInDEV(fiber) {
...
+      // Warning signals like this are more useful if they're strong.
+      // For this reason, it's good to always avoid updating state from inside of an effect's cleanup function.
+      // Even when you know there is no potential leak, React has no way to know and so it will warn anyway.
+      // In most cases we suggest moving state updates to the useEffect() body instead.
+      // This works so long as the component is updating its own state (or the state of a descendant).
+      //
+      // However this will not work when a component updates its parent state in a cleanup function.
+      // If such a component is unmounted but its parent remains mounted, the state will be out of sync.
+      // For this reason, we avoid showing the warning if a component is updating an ancestor.
+      let current = fiber;
```
gaearon: Nit: rename to node because “current” without any qualifiers has a specific meaning in other places in code (current tree), and it’s also a bit confusing when compared with “current debug fiber” (it’s hard to remember what both currents mean).

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2822,14 +2823,39 @@ function warnAboutUpdateOnUnmountedFiberInDEV(fiber) {
...
+        // currentDebugFiberInDEV owns the effect destroy function currently being invoked.
+        if (
+          currentDebugFiberInDEV === current ||
+          currentDebugFiberInDEV === current.alternate
+        ) {
+          console.error(
+            "Can't perform a React state update from within a useEffect cleanup function. " +
+              'To fix, move state updates to the useEffect() body in %s.%s',
+            tag === ClassComponent
+              ? 'the componentWillUnmount method'
```
gaearon: Is this branch possible? Why does it exist? The message will say “to the useEffect body in the componentWillUnmount method”. That doesn’t make sense.

bvaughn: Great question.

No, this branch isn't possible. I think the weird wording is an artifact of suggested wording changes on the previous PR. (I didn't actually change it any in this PR.)

I'll do a cleanup pass on it though before landing.

PR #18412, opened 2020-03-28
Reset lastEffect when resuming SuspenseList

sebmarkbage:
We store an effect pointer so we can backtrack in the effect list in some cases. This is a stateful variable. If we interrupt a render we need to reset it.

This field was added after the optimization was added and I didn't remember to reset it here.

Otherwise we end up not resetting the firstEffect so it points to a stale list. As a result children don't end up inserted like we think they were. Then we try to remove them it errors.

It would be nicer to just get rid of the effect list and use the tree for effects instead. Maybe we still need something for deletions tho.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
@@ -2300,4 +2300,189 @@ describe('ReactSuspenseList', () => {
     // remount.
     expect(previousInst).toBe(setAsyncB);
   });
+
+  it('is able to re-suspend the last rows during an update with hidden', async () => {
+    let AsyncB = createAsyncText('B');
+
+    let setAsyncB;
+
+    function B() {
+      let [shouldBeAsync, setAsync] = React.useState(false);
+      setAsyncB = setAsync;
+
+      return shouldBeAsync ? (
+        <Suspense fallback={<Text text="Loading B" />}>
+          <AsyncB />
+        </Suspense>
+      ) : (
+        <Text text="Sync B" />
+      );
+    }
+
+    function Foo({updateList}) {
+      return (
+        <SuspenseList revealOrder="forwards" tail="hidden">
+          <Suspense key="A" fallback={<Text text="Loading A" />}>
+            <Text text="A" />
+          </Suspense>
+          <B key="B" updateList={updateList} />
+        </SuspenseList>
+      );
+    }
+
+    ReactNoop.render(<Foo />);
+
+    expect(Scheduler).toFlushAndYield(['A', 'Sync B']);
+
+    expect(ReactNoop).toMatchRenderedOutput(
+      <>
+        <span>A</span>
+        <span>Sync B</span>
+      </>,
+    );
+
+    let previousInst = setAsyncB;
+
+    // During an update we suspend on B.
+    ReactNoop.act(() => setAsyncB(true));
+
+    expect(Scheduler).toHaveYielded([
+      'Suspend! [B]',
+      'Loading B',
+      // The second pass is the "force hide" pass
+      'Loading B',
+    ]);
+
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
```diff
@@ -2300,4 +2300,189 @@ describe('ReactSuspenseList', () => {
...
+      // Flush partially through.
+      expect(Scheduler).toFlushAndYieldThrough(['B', 'C']);
+
+      // Schedule another update at higher priority.
+      Scheduler.unstable_runWithPriority(
+        Scheduler.unstable_UserBlockingPriority,
+        () => updateHighPri(true),
+      );
+
+      // That will intercept the previous render.
```
gaearon: interrupt? :-)

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
```diff
@@ -2300,4 +2300,189 @@ describe('ReactSuspenseList', () => {
...
+
+    jest.runAllTimers();
+
+    expect(Scheduler).toHaveYielded(
+      __DEV__
+        ? [
+            // First attempt at high pri.
+            'Suspend! [A]',
+            'Loading A',
+            // Re-render at forced.
```
gaearon: What does "forced" mean here? Which branch does it correspond to? Explain like I'm five.

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
```diff
@@ -2300,4 +2300,189 @@ describe('ReactSuspenseList', () => {
...
+            'Suspend! [A]',
+            'Loading A',
+            // We auto-commit this on DEV.
```
gaearon: Why? "Flush fallbacks" flag?

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
```diff
@@ -2300,4 +2300,189 @@ describe('ReactSuspenseList', () => {
...
+            // Try again on low-pri.
+            'Suspend! [A]',
+            'Loading A',
+          ]
+        : [
```
gaearon: This test is verifying the same logs appear three or four times. But how do we know the comments are actually telling the truth? Can we include the boolean values in logs so that we know we're *actually* rendering at the priority suggested by comments?

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
sebmarkbage: SuspenseList sets the suspense context that during render “forces” the suspense boundary into fallback state.

In this case it’s only one boundary so it doesn’t need to but we don’t know. If there was another boundary in the row we would have to rerender that one so that’s the second pass.

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
sebmarkbage: Yea

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
sebmarkbage: I’d rather just delete this assertion. It’s unrelated. It’s just that Scheduler expects force me to assert. This is super suboptimal so it’ll change by heuristics a lot.

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.internal.js
gaearon: Ohh that makes sense! I didn’t realize it works this way.

PR #18430, opened 2020-03-29
feat(StrictMode): Double-invoke render for every component

### LINKED ISSUE #18426: Call all functions twice in StrictMode
React version: 16.3-16.13

## Steps To Reproduce

1. Render function component with side-effects and without hooks in StrictMode
2. Component only renders once

Link to code example: https://codesandbox.io/s/strictmode-w-and-wo-hooks-vgxvh

## The current behavior

StrictMode only renders function components with hooks twice following https://github.com/facebook/react/issues/15074#issuecomment-471197572

## The expected behavior

I think making [the distinction between components with and without hooks causes more confusion than it helps](https://github.com/mui-org/material-ui/issues/20313). Especially since the docs do not mention this. I amended [past efforts to improve the StrictMode docs](https://github.com/reactjs/reactjs.org/pull/2515) but since this has been ignored I'm not sure this helps. I would also work on making the behavior consistent across all function component types if this is possible.

eps1lon:
## Summary

Closes #18426.

Essentially #14639 rebased with `master` where concerns were voiced about existing tests.

## Test Plan

ReactHooks-internal is responsible for this test. This also affects any test counting renders.

### DIFF
--- a/packages/react-art/src/__tests__/ReactART-test.js
+++ b/packages/react-art/src/__tests__/ReactART-test.js
@@ -391,7 +391,7 @@ describe('ReactART', () => {
       </CurrentRendererContext.Provider>,
     );
 
-    expect(Scheduler).toFlushAndYieldThrough(['A']);
+    expect(Scheduler).toFlushAndYieldThrough(__DEV__ ? ['A', 'A'] : ['A']);
 
     ReactDOM.render(
       <Surface>
@@ -406,7 +406,9 @@ describe('ReactART', () => {
     expect(ops).toEqual([null, 'ART']);
 
     ops = [];
-    expect(Scheduler).toFlushAndYield(['B', 'C']);
+    expect(Scheduler).toFlushAndYield(
+      __DEV__ ? ['B', 'B', 'C', 'C'] : ['B', 'C'],
+    );
 
     expect(ops).toEqual(['Test']);
   });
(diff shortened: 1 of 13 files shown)


gaearon:
Seems like bundle tests fail.

eps1lon:
> Seems like bundle tests fail.

It seems like react-test-renderer is not using `beginWork` from the react-reconciler package in `test-build` (but is in `test`).

Apparently `test-build` rewrites some `react-reconciler/` imports but I'm a bit confused why `react-reconciler` sometimes imports via relative specifier (e.g. `./ReactFiberRoot`) and sometimes via deep specifier (e.g. `react-reconciler/src/ReactRootTags`). Maybe I'm looking in the wrong spot.

eps1lon:
Following https://github.com/facebook/react/pull/14639/files#diff-0b241de0dfec2d07bb2b2fef76b254d2R11-R24 though I don't think I understand the distinction between source and build tests.

gaearon:
>It seems like react-test-renderer is not using beginWork from the react-reconciler package in test-build (but is in test).

Not sure what you mean by this. In the bundle tests, `react-reconciler` is bundled *inside* of `react-test-renderer` bundle. Just like it's bundled inside of `react-dom`.

The difference is that you can't set feature flags for bundle tests. Because nothing will read them. They're already compiled away.

gaearon:
Looks good. Thanks.

Issue #18429, opened 2020-03-29
Bug: SuspenseList revealOrder="together" and error boundaries


eps1lon:
React version: 0.0.0-experimental-aae83a4b9

## Steps To Reproduce

1. Wrap and error boundary around a suspense boundary inside a `<SuspenseList revealOrder="together" />`
2. Crash with `Cannot read property 'shared' of null`

Link to code example:

https://codesandbox.io/s/quirky-cannon-e4t1c is forked from https://codesandbox.io/s/black-wind-byilt

## The current behavior

The following code causes the described crash:

```jsx
<SuspenseList revealOrder="together">
  <ErrorBoundary fallback={null}>
    <Suspense fallback={<h2>Loading posts...</h2>}>
      <ProfileTimeline resource={resource} />
    </Suspense>
  </ErrorBoundary>
  <Suspense fallback={<h2>Loading fun facts...</h2>}>
    <ProfileTrivia resource={resource} />
  </Suspense>
</SuspenseList>
```

`revealOrder="forwards"` does not crash as well as placing the error boundary inside the suspense boundary like so: 
```jsx
<SuspenseList revealOrder="together">
  <Suspense fallback={<h2>Loading posts...</h2>}>
    <ErrorBoundary fallback={null}>
      <ProfileTimeline resource={resource} />
    </ErrorBoundary>
  </Suspense>
  <Suspense fallback={<h2>Loading fun facts...</h2>}>
    <ProfileTrivia resource={resource} />
  </Suspense>
</SuspenseList>
```

## The expected behavior

The position of the ErrorBoundary is likely wrong in the first place (since I can move it around to not cause a crash while preserving the semantics I expect). Maybe this needs a descriptive warning/error why this is happening and how I should resolve it?

gaearon:
Sounds like a bug.

gaearon:
Would you mind sending this as a PR with a failing test? You can see existing SuspenseList tests as an inspiration.

eps1lon:
Added test in https://github.com/facebook/react/pull/18432

This is reproducible with a basic class component which means the assumption in 
https://github.com/facebook/react/blob/c5d2fc7127654e43de59fff865b74765a103c4a5/packages/react-reconciler/src/ReactUpdateQueue.js#L399-L401

does not hold. This guard was removed in https://github.com/facebook/react/commit/7bf40e1cfdb780788700a41bf30163fdb8d105a3#diff-b69c8f2165f95ef2355ac661c12fa407L867

[...]

PR #18432, opened 2020-03-29
fix: Prevent crash when resuming class component inside SuspenseList

eps1lon:
## Summary

Test for #18429 (includes a fix but it's unclear if the semantics are correct) 

## Test Plan

~Might be interesting add matchers for the case where one of the error boundaries is actually triggered.~ No longer concerned with error boundaries. They caused the crash originally but the bug could be reduced to basic class components.

### DIFF
--- a/packages/react-reconciler/src/ReactFiber.js
+++ b/packages/react-reconciler/src/ReactFiber.js
@@ -95,6 +95,8 @@ import {
   REACT_BLOCK_TYPE,
 } from 'shared/ReactSymbols';
 
+import {initializeUpdateQueue} from './ReactUpdateQueue';
+
 let hasBadMapPolyfill;
 
 if (__DEV__) {
@@ -546,6 +548,15 @@ export function resetWorkInProgress(
     workInProgress.child = null;
     workInProgress.memoizedProps = null;
     workInProgress.memoizedState = null;
+    // updateQueue must never be null on ClassComponent or HostRoot
+    if (
+      workInProgress.tag === ClassComponent ||
+      workInProgress.tag === HostRoot
+    ) {
+      initializeUpdateQueue(workInProgress);
+    } else {
+      workInProgress.updateQueue = null;
+    }
 
     workInProgress.dependencies = null;
 
--- a/packages/react-reconciler/src/ReactFiber.js
+++ b/packages/react-reconciler/src/ReactFiber.js
@@ -546,7 +546,6 @@ export function resetWorkInProgress(
     workInProgress.child = null;
     workInProgress.memoizedProps = null;
     workInProgress.memoizedState = null;
-    workInProgress.updateQueue = null;
 
     workInProgress.dependencies = null;
 
--- a/packages/react-reconciler/src/ReactFiberClassComponent.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.js
@@ -885,8 +885,13 @@ function resumeMountClassInstance(
     nextContext = getMaskedContext(workInProgress, nextLegacyUnmaskedContext);
   }
 
-  if (workInProgress.updateQueue === null) {
-    initializeUpdateQueue(workInProgress);
+  if (__DEV__) {
+    if (workInProgress.updateQueue === null) {
+      console.error(
+        'Class fiber is missing an updateQueue. ' +
+          'This error is likely caused by a bug in React. Please file an issue.',
+      );
+    }
   }
 
   const getDerivedStateFromProps = ctor.getDerivedStateFromProps;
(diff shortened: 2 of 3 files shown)


packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -885,6 +885,10 @@ function resumeMountClassInstance(
     nextContext = getMaskedContext(workInProgress, nextLegacyUnmaskedContext);
   }
 
+  if (workInProgress.updateQueue === null) {
+    initializeUpdateQueue(workInProgress);
```
eps1lon: This looks more like it's fixing symptoms.

The queue is emptied in `resetWorkInProgress` if `current === null`: 
https://github.com/facebook/react/blob/31a9e391f757d374ce0065da6735f08137151507/packages/react-reconciler/src/ReactFiber.js#L541-L549

In `updateClassComponent` we resume in `resumeMountClassInstance` if `current === null` which straight goes to `processUpdateQueue` assuming the `updateQueue` has been initialized (but it isn't anymore).

Note that not reseting the `updatedQueue` in `resetWorkInProgress` fixes this as well while not failing any other test.

acdlite: If the type were sound, the `updateQueue` on a class fiber should always be non-null. So my preference is to put it there instead.

But also it's classes so whatever :D

acdlite: If you do move the fix to `resetWorkInProgress`, I would add a dev-only check to `resumeMountClassInstance` to guard against a regression.

```js
if (__DEV__) {
  if (workInProgress.updateQueue === null) { 
    console.error('Internal error: Class fiber is missing an updateQueue');
  }
}
```

gaearon:
@eps1lon Wanna do this change? https://github.com/facebook/react/pull/18432#discussion_r400549699

eps1lon:
> @eps1lon Wanna do this change? [#18432 (comment)](https://github.com/facebook/react/pull/18432#discussion_r400549699)

1. `- workInProgress.updateQueue = null` 
2. 
  ```diff
  - workInProgress.updateQueue = null;
  + initializeUpdateQueue(workInProgress);
  ```

The first approach passed locally.

eps1lon:
I guess 2 it is (to have a sound type) if I follow 
> If the type were sound, the updateQueue on a class fiber should always be non-null.

gaearon:
Sorry I'm not sure what you're asking. Can you rephase?

eps1lon:
Only 1 passes so I guess this is answered for me.  I forgot a word there: "1.  remove `workInProgress.updateQueue = null`"

packages/react-reconciler/src/ReactFiberClassComponent.js
eps1lon: - don't clear updateQueue in resetWorkInProgress
- add warning if updateQueue is empty in resumeWorkInProgress:
  > Class fiber is missing an updateQueue. This error is likely caused by a bug in React. Please file an issue.

packages/react-reconciler/src/ReactFiber.js
```diff
@@ -546,7 +546,6 @@ export function resetWorkInProgress(
     workInProgress.child = null;
     workInProgress.memoizedProps = null;
     workInProgress.memoizedState = null;
-    workInProgress.updateQueue = null;
```
sebmarkbage: I think this is going to break other things that are likely not covered by tests. We’ve had many bugs related to not resetting enough.

eps1lon: I'd have to dive deeper to be able to fix this then. I tried `initializeUpdateQueue(workInProgress)` instead which caused a bunch of "Unexpected pop" or "Unexpected Fiber popped". Also tried 
```
if (workInProgress.updateQueue !== null) {
  initializeUpdateQueue(workInProgress);
}
``` 
which caused the same issue

eps1lon: Nevermind. VSCode has problems with flow and didn't realize `initializeUpdateQueue` was never defined because I imported from the same line as `import type`. The ReferenceError was just buried below 5 screens of console errors.

packages/react-reconciler/src/ReactFiber.js
```diff
@@ -546,7 +548,15 @@ export function resetWorkInProgress(
...
+    // updateQueue must never be null on ClassComponent or HostRoot
```
eps1lon: In line with https://github.com/facebook/react/blob/c5d2fc7127654e43de59fff865b74765a103c4a5/packages/react-reconciler/src/ReactUpdateQueue.js#L399

would it make more sense to move this logic into a `resetUpdateQueue` inside `ReactUpdatedQueue.js`?

sebmarkbage:
Thanks for the well written test! Opened another PR with a fix. https://github.com/facebook/react/pull/18448/files

The issue was that we were resuming in the first place. We shouldn't trigger the resume path here because it's not equipped for true resuming.

eps1lon:
Yeah my approach did not feel right. Thanks!

PR #18441, opened 2020-03-31
ReactDOM.useEvent: revert and add guard for null stateNode

trueadm:
This PR reverts the revert PR (https://github.com/facebook/react/pull/18438) and applies the appropiate fix. The previous problem was that we were trying to call `getListenersFromTarget` on the `instance`, which was taken from `node.stateNode`. I missed the case where the `stateNode` can be `null` (when the component gets unmounted), so we should make sure to check that the instance is not `null`. I also added some validation to ensure that `lastHostComponent` should never be `null` too, as per comments from @gaearon in https://github.com/facebook/react/pull/18375.


gaearon:
Could you go through all code behind useEffect flag to search for similar bugs?

trueadm:
@gaearon I deep a deep dive earlier today, I couldn't see anywhere else we access `stateNode` in relation to `useEvent`.

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
```diff
@@ -76,55 +87,99 @@ export default function accumulateTwoPhaseListeners(
...
+            const bubbleListener = getListener(node, bubbled);
+            if (bubbleListener != null) {
+              // Bubble listeners/instances should go at the end, so we
+              // push them to the end of the array.
+              dispatchListeners.push(bubbleListener);
+              dispatchInstances.push(node);
+            }
           }
+        } else if (tag === HostText) {
+          lastHostComponent = instance;
```
gaearon: I think in principle it's okay if we want to skip those (would the old event system catch them? I'm not sure). We should just make sure we don't crash.

trueadm: It won't break. Let me remove this block.

PR #18444, opened 2020-03-31
Enable new passive effect behavior for FB builds

bvaughn:
Previously this behavior was controlled by GKs. This PR updates the flags to be enabled statically. It also enables the flags in the test builds.


acdlite:
There's this file, too https://github.com/facebook/react/blob/76678a5fc6672fea9f0717efd6a008feb1d821f3/packages/shared/forks/ReactFeatureFlags.www-dynamic.js#L20-L23

bvaughn:
> There's this file, too

Gah, right. So many files. I think the right thing to do for that file is to just delete the flags from it?

trueadm:
Just FYI: I reverted this in https://github.com/facebook/react/pull/18467.

PR #18445, opened 2020-03-31
Remove disableMapsAsChildren flag

bvaughn:
Change warning to say the case is unsupported (not that it will be deprecated).

New warning:
> Using Maps as children is not supported. Use an array of keyed ReactElements instead.


gaearon:
There is another warning in `ChildFiber`. Let's make sure they say the same thing?

bvaughn:
I went ahead and deleted the error code as well, since the feature was never actually released.

gaearon:
In fact the one in `ChildFiber` is the important one. This one is only for Children utilities like map.

bvaughn:
Oh thanks Dan

PR #18446, opened 2020-03-31
Disable module components dynamically for WWW

gaearon:
So that we can turn it off with a GK and then make it static again.


packages/shared/forks/ReactFeatureFlags.www.js
```diff
@@ -16,6 +16,7 @@ const dynamicFeatureFlags: DynamicFeatureFlags = require('ReactFeatureFlags');
 
 export const {
   debugRenderPhaseSideEffectsForStrictMode,
+  disableModulePatternComponents,
```
bvaughn: Should also be added to `ReactFeatureFlags.www-dynamic.js`

acdlite:
Need to update ReactFeatureFlags.www-dynamic.js. I believe Flow will catch this.

acdlite:
Also need to delete the static export
https://github.com/facebook/react/blob/1cd921db7c8797159244773f43694e035f7b5aea/packages/shared/forks/ReactFeatureFlags.www.js#L69

packages/shared/forks/ReactFeatureFlags.www-dynamic.js
```diff
@@ -25,6 +25,7 @@ export const replayFailedUnitOfWorkWithInvokeGuardedCallback = __DEV__;
 // to __VARIANT__.
 export const enableTrustedTypesIntegration = false;
 export const warnAboutShorthandPropertyCollision = true;
+export const disableModulePatternComponents = false;
```
acdlite: Why is this hard coded?

acdlite: It's not a big deal in this case but generally we should be testing both versions of any dynamic flag using __VARIANT__

gaearon: I don't want to go around and wrap every existing test that uses them. Should I?

acdlite: I don't think it's that many, right? And we need to update the tests at some point anyway. Though I can see how it could be annoying to have to front load that work.

gaearon: I guess it's good to do this now because otherwise we won't know if we'll break this path for WWW only.

gaearon: This helped find a few loose ends (missing SSR check and missing warning). So thanks for the nudge.

PR #18448, opened 2020-03-31
Reset stateNode in resetWorkInProgress

sebmarkbage:
Alternative fix to #18432

`resetWorkInProgress` is supposed to put the Fiber into the same state as if it was just created by child fiber reconciliation. For newly created fibers, that means that stateNode is null.


gaearon:
Why does resuming code path still exist?

sebmarkbage:
Not sure. Maybe legacy mode suspense?

PR #18455, opened 2020-04-01
Fix ESLint rule crash

gaearon:
Reported internally. The issue is `useEffect(foo)` where `foo` is coming e.g. from props. That should be legit; we don't need to check it. However, `useCallback(foo)` or `useMemo(foo)` should warn because they're useless without deps.

### DIFF
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -93,6 +93,28 @@ export default {
       const reactiveHook = node.callee;
       const reactiveHookName = getNodeWithoutReactNamespace(reactiveHook).name;
       const declaredDependenciesNode = node.arguments[callbackIndex + 1];
+      const isEffect = /Effect($|[^a-z])/g.test(reactiveHookName);
+
+      // Check the declared dependencies for this reactive hook. If there is no
+      // second argument then the reactive callback will re-run on every render.
+      // So no need to check for dependency inclusion.
+      if (!declaredDependenciesNode && !isEffect) {
+        // These are only used for optimization.
+        if (
+          reactiveHookName === 'useMemo' ||
+          reactiveHookName === 'useCallback'
+        ) {
+          // TODO: Can this have a suggestion?
+          reportProblem({
+            node: reactiveHook,
+            message:
+              `React Hook ${reactiveHookName} does nothing when called with ` +
+              `only one argument. Did you forget to pass an array of ` +
+              `dependencies?`,
+          });
+        }
+        return;
+      }
 
       switch (callback.type) {
         case 'FunctionExpression':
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -93,6 +93,28 @@ export default {
       const reactiveHook = node.callee;
       const reactiveHookName = getNodeWithoutReactNamespace(reactiveHook).name;
       const declaredDependenciesNode = node.arguments[callbackIndex + 1];
+      const isEffect = /Effect($|[^a-z])/g.test(reactiveHookName);
```
sompylasar, line 96: Would "Effects" be a valid name for this? `useSomeCustomEffects(fn, deps)`

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -93,6 +93,28 @@ export default {
...
+        ) {
+          // TODO: Can this have a suggestion?
+          reportProblem({
+            node: reactiveHook,
+            message:
+              `React Hook ${reactiveHookName} does nothing when called with ` +
+              `only one argument. Did you forget to pass an array of ` +
+              `dependencies?`,
+          });
+        }
```
sompylasar, line 115: What about "else" here? Do we ignore such hooks?

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -101,13 +123,18 @@ export default {
             callback,
             declaredDependenciesNode,
```
sompylasar, line 124: It's confusing to run `visitFunctionWithDependencies(` without dependencies (`declaredDependenciesNode` can be `undefined` here). I checked, that function seems to correctly bail out if there are none, still confusing.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
gaearon: Not with the current check, but you can use `additionalHooks` for that.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
gaearon: Rest would be effects, right? For effects we do allow to not specify deps.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
gaearon: We need to run it because there are other checks before we can bail out. Those other checks require a function body but not a deps node.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
sompylasar: > Rest would be effects, right? For effects we do allow to not specify deps.

Seeing this code for the first time today, it was quite hard to read through and reason about its control flow (especially due to lots of early returns); maybe you're right and it's fine.

gaearon: It's a pretty complex rule. :-) I'd happily take refactorings provided they don't break tests.

PR #18457, opened 2020-04-01
Fix: Don't read primaryChild.childExpirationTime

acdlite:
This is a variant of the fix in 5a0f1d5014bbf7bba4d775eb4617a5a8d18ba31c. We can't rely on the primary fiber's `childExpirationTime` field to be correct.

In this case, we can read from the Suspense boundary fiber instead.  This will include updates that exist in the fallback fiber, but that's not a big deal; the important thing is that we don't drop updates.

Longer term, we'll likely remove fragment reparenting from the Suspense implementation. It's only there to save on memory but because it's the only place we do reparenting, it's been hard to maintain.

If we ever do add reparenting, what this bug demonstrates is that the state tree and the UI tree are not the same thing. It just happens to be that in current React (no reparenting), the "state ancestors" of a node are always subset of its "UI ancestors", which is why we can get away with only having a single return path.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -1646,16 +1646,9 @@ function shouldRemainOnFallback(
 function getRemainingWorkInPrimaryTree(
   current: Fiber,
   workInProgress: Fiber,
-  currentPrimaryChildFragment: Fiber | null,
   renderExpirationTime,
 ) {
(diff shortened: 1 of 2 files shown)


gaearon:
>We can't rely on the primary fiber's childExpirationTime field to be correct.

Is this the only place we rely on it?

acdlite:
> Is this the only place we rely on it?

I think so, yeah

Issue #17609, opened 2019-12-15
Controlled numeric input gets cleared when unfocused


kripod:
**Do you want to request a *feature* or report a *bug*?**
I would like to report a bug.

**What is the current behavior?**
A numeric input field gets cleared accidentally on several occasions.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

1. Open [this sandbox](https://codesandbox.io/embed/sad-rgb-mltuj)
2. Type “12.” into the field, with the trailing decimal separator
3. Unfocus the control
4. Append “.” to the field’s value to see “12..”
5. Unfocus the control once again and see that the number has completely disappeared

**What is the expected behavior?**

Similar to how uncontrolled inputs work (remove the `value` prop and then repeat the steps above), the input should not be cleared on blur.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

I’m using React 16.12.0 and experienced the same behavior with the latest version of Chrome and iOS Safari. Regarding this bug, I have no experience with previous versions of React.

AfterThreeYears:
i think this is HTML Input elememt features, because you input not number, so input well clean content.

caner-karaman:
This is not React behavior, Its come from browser

kripod:
This isn't a browser bug. I've just set up a sandbox with the exact same code, but using Preact instead of React: https://codesandbox.io/s/objective-antonelli-o0f2x

Input "12." -> Unfocus -> Append "." -> Unfocus works just fine with Preact, but React clears the input field unintentionally.

vcanales:
Some investigation led me to #29d9710892f773b4d5081f1b77820d40d79f69d3 which took me [here](https://github.com/facebook/react/issues/7253#issuecomment-236074326). This is definitely a bug following that logic, and it seems to be related to not checking for special cases of validity in the case of `input[type=number]` when `onBlur` happens.

That being said, this issue will go away when the changes made in #13526 are no longer under a feature flag (`disableInputAttributeSyncing`). I checked this by modifying your copy of `react-dom`, but you can probably build react with the flag set to `true`.

gaearon:
This is probably still a bug.

gaearon:
[...]

PR #18459, opened 2020-04-01
Revert "Fix email cursor jump (#18379)"

gaearon:
This reverts commit 9b88b78b3d61e8fe81968d59b74592d5a53540b7.

I've changed my mind. This is too scary to change because there are other unknowns around this whole area. Like we could be introducing more regressions like https://github.com/facebook/react/issues/17609. If we can't sit down and fix this holistically, the best we can do is to reduce changes to avoid churn in the workarounds.

Should we fix it eventually? Yes. With an approach that is well-thought out for all input types and doesn't cause other issues. Such as maybe disabling syncing would help here. We'll likely slate this for after 17.


gaearon:
I'm going to leave the fixture tests though. Those are good to have and will guide future work.

PR #18461, opened 2020-04-02
Run more flags in VARIANT tests

gaearon:
As it says.


packages/react-interactions/events/src/dom/__tests__/Input-test.internal.js
```diff
@@ -140,8 +140,15 @@ describe('Input event responder', () => {
       ref.current.dispatchEvent(
         new Event('input', {bubbles: true, cancelable: true}),
       );
-      expect(onChangeCalled).toBe(0);
-      expect(onValueChangeCalled).toBe(0);
+
+      if (ReactFeatureFlags.disableInputAttributeSyncing) {
+        // TODO: figure out why. This might be a bug.
```
gaearon, line 145: This isn't new. Copypaste from ChangeEventPlugin test from which this was copied.

gaearon:
Interesting. Modern event system has a few failures in the stable channel. Let me remove it for now.

PR #18464, opened 2020-04-02
ReactDOM.useEvent: fix scope propagation issue

trueadm:
This PR fixes an issue with `useEvent` and scopes (see attached test). Specifically, we need to pass through the instance of the scope, so we can properly track if we can stopPropagation. Previously we used the `currentTarget`, which is the last physical DOM element. In the cases of scopes, this can be problematic if we have:

```jsx
<Scope>
  <Scope>
    <div />
  </Scope>
</Scope>
```

To get around this we instead pass around an object, as `{ instance: Fiber | null, container: EventTarget }`. We use this object to maintain the instance and contain (currentTarget) separately. I also moved the logic for this into the modern event system module so we don't affect other event systems with this change.


packages/legacy-events/EventPluginUtils.js
```diff
@@ -80,20 +80,12 @@ export function executeDispatchesInOrder(event) {
     validateEventDispatches(event);
   }
   if (Array.isArray(dispatchListeners)) {
-    let previousInstance;
     for (let i = 0; i < dispatchListeners.length; i++) {
-      const instance = dispatchInstances[i];
```
gaearon, line 85: Can you explain what this code was doing? I'm not sure I understand why we would have the same instance twice, or how that relates to propagation. I see it was added very recently? Is this used by the legacy event system? Seems a bit scary to change it.

gaearon: I think this PR is essentially removing the modified code, and instead forking it for modern system only?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -165,6 +167,57 @@ export const reactScopeListenerStore: WeakMap<
...
+      // - as an object { instance: Fiber | null, container: EventTarget }
+      // The second form is used for useEvent types
+      if (dispatchInstance.container !== undefined) {
+        instance = ((dispatchInstance: any): DispatchInstance).instance;
+        currentTarget = ((dispatchInstance: any): DispatchInstance).container;
+      } else {
+        currentTarget = getNodeFromInstance(((dispatchInstance: any): Fiber));
+      }
+      // We check if the instance was the same as the last one,
+      // if it was, then we're still on the same instance thus
```
gaearon, line 205: What does it mean to be on the same instance? Several `useEvent` targeting the same node?

packages/legacy-events/EventPluginUtils.js
trueadm: This is reverting this file back to what it was before any of the changes from the modern event system were made to it. Like you say, we instead use a forked version instead.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -165,6 +167,57 @@ export const reactScopeListenerStore: WeakMap<
...
+  let previousInstance;
+
+  if (dispatchListeners !== null && dispatchInstances !== null) {
+    for (let i = 0; i < dispatchListeners.length; i++) {
+      const dispatchInstance = dispatchInstances[i];
+      const dispatchListener = dispatchListeners[i];
+      let instance = dispatchInstance;
+      let currentTarget;
+
+      // dispatchInstance can come in two forms:
```
gaearon, line 194: I find this confusing. I worry these special cases are making modern event system harder to follow than the legacy one. Imagine we add a `container` field to Fiber later and this will brek.

If we really need to keep more data here, can we always wrap them in objects?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: Exactly, it's for `useEvent` targeting the same node.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: Yeah, I see that. I'll revise.

gaearon: Or I guess a new field so that we don't put more things into `_dispatchInstances`. Like `_dispatchContainers`.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -165,6 +167,57 @@ export const reactScopeListenerStore: WeakMap<
...
+        currentTarget = ((dispatchInstance: any): DispatchInstance).container;
+      } else {
+        currentTarget = getNodeFromInstance(((dispatchInstance: any): Fiber));
+      }
+      // We check if the instance was the same as the last one,
+      // if it was, then we're still on the same instance thus
+      // propagation should not stop. If we add support for
+      // stopImmediatePropagation at some point, then we'll
+      // need to handle that case here differently.
+      if (instance !== previousInstance && event.isPropagationStopped()) {
```
gaearon, line 209: I'm still not sure I understand this piece of logic. We check for propagation for the first `useEvent` OK. Then we have another `useEvent` on the same node. Let's say this one calls `stopPropagation`. Would we ignore it? Why?

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
```diff
@@ -111,18 +111,19 @@ export default function accumulateTwoPhaseListeners(
...
           const captureListeners = Array.from(listeners.captured);
           const bubbleListeners = Array.from(listeners.bubbled);
+          const container = ((lastHostComponent: any): Element);
 
           for (let i = 0; i < captureListeners.length; i++) {
             const listener = captureListeners[i];
             const {callback} = listener;
             dispatchListeners.unshift(callback);
-            dispatchInstances.unshift(((lastHostComponent: any): Element));
+            dispatchInstances.unshift({instance: node, container});
```
gaearon, line 120: Can you explain in more detail what's different about Scopes that makes us need this? Why do we need to "remember" the closest "last host component", but we don't for regular events? What is special here? Why can't we figure out the DOM node later?

trueadm: Scopes don't have a physical DOM element representing them as a target, so in order to provide the user with one, we given them the nearest DOM node from the
the native event's original target – this is also how React Flare worked. We can't do this later, as we don't know which part of the tree we came in by – so we have to do this here.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: We'd ignore it as `stopPropagation` works like that on the DOM today. We'd need to add `stopImmediatePropagation` that works like how you probably want, where we can stop propagation even on the same instance.

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
gaearon: Is this about `target` or `currentTarget`?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: I was doing just that :)

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
trueadm: It's about the `currentTarget` on the event object for when events are on scopes. We want it to be the last "real" currentTarget that is in the DOM.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
gaearon: Hmm. I'm not sure I understand what you mean by DOM working like this.

Can you provide an example?

>We'd need to add stopImmediatePropagation that works like how you probably want

I am not referring to one `useEvent` preventing another `useEvent`.

I am talking about the second `useEvent` being able to prevent *further* propagation up the tree. Does that work?

trueadm: > I am talking about the second useEvent being able to prevent further propagation up the tree. Does that work?

Yes it does, by design.

> Hmm. I'm not sure I understand what you mean by DOM working like this.

I mean that it works in same way propagation works on the DOM. We move a pointer up and down a single-list (which is section of the node tree) and each level of the tree is a propagation layer. If something stops propagation, we don't exit processing of the list until we move the pointer to the next level.

gaearon: >Yes it does, by design.

Do you mind adding a test for this?

trueadm: That's what the test on this PR should be testing, unless we're talking about different things.

gaearon: OK I get it now. What you're saying is that by the time we reach the next node, we will run this check again, and so if it has been stopped by then, we won't dispatch for it.

packages/react-dom/src/events/accumulateEnterLeaveListeners.js
```diff
@@ -83,24 +83,26 @@ function accumulateEnterLeaveListenersForEvent(
...
     if (node === common) {
       break;
     }
-    const alternate = node.alternate;
+    const {alternate, stateNode, tag} = node;
     if (alternate !== null && alternate === common) {
       break;
     }
-    if (node.tag === HostComponent) {
+    if (stateNode !== null && tag === HostComponent) {
```
gaearon, line 97: Is this a behavior change?

Either it was never null. And in that case it's safe to use it.

Or it was sometimes null. And then excluding this component from traversal changes the behavior without any flags.

packages/legacy-events/SyntheticEvent.js
```diff
@@ -78,6 +78,7 @@ function SyntheticEvent(
   this.nativeEvent = nativeEvent;
   this._dispatchListeners = null;
   this._dispatchInstances = null;
+  this._dispatchContainers = null;
```
gaearon, line 81: This also needs to be cleared in `destructor`.  Otherwise we're risking a memory leak.

packages/react-dom/src/events/accumulateEnterLeaveListeners.js
trueadm: If we the stateNode is `null` it saves us processing the listener for the node as it's unmounted and we can avoid CPU cycles.

gaearon: What happens today if the node is unmounted in the middle of an event?

packages/react-dom/src/events/accumulateEventTargetListeners.js
```diff
@@ -32,7 +34,8 @@ export default function accumulateEventTargetListeners(
           const listener = captureListeners[i];
           const {callback} = listener;
           dispatchListeners.push(callback);
-          dispatchInstances.push(targetContainer);
+          dispatchInstances.push(null);
```
gaearon, line 37: Why are these null?

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
```diff
@@ -67,9 +68,11 @@ export default function accumulateTwoPhaseListeners(
               if (capture === true) {
                 dispatchListeners.unshift(callback);
                 dispatchInstances.unshift(node);
+                dispatchContainers.unshift(instance);
```
gaearon, line 71: The naming is really tripping me. We're pushing a "node" to "instances" but an "instance" to "containers". What is the mental model for what's actually in there? Can we rename these fields and variable to match, and clarify which are DOM nodes and which are Fibers and which are something else?

gaearon: I don't mean to be pedantic. I find it very difficult to wrap my mind around this, especially because we have several cases. Right now I don't even have the confidence any particular field always holds things of the same type.

packages/react-dom/src/events/accumulateEnterLeaveListeners.js
trueadm: If it were `null` today, we'd invoke `invokeGuardedCallbackAndCatchFirstError` with a `null` `currentTarget` on the synthetic event.

packages/react-dom/src/events/accumulateEventTargetListeners.js
trueadm: We don't have an instance for `EventTarget` listeners, because they don't need or might even have a backing fiber (i.e. `window` or `document`).

packages/react-dom/src/events/accumulateEnterLeaveListeners.js
gaearon: If this is purely optimization I'd prefer to leave it out. Seems like unnecessary risk if this code is shared by the legacy system unless we have a clear proof that this does not change observable behavior.

gaearon: >we'd invoke invokeGuardedCallbackAndCatchFirstError with a null currentTarget on the synthetic event.

That seems like observably different behavior? User code would get called.

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
trueadm: Valid point, I'll clear up the names.

gaearon: How about this:

```
_dispatchListeners
_dispatchFibers
_dispatchCurrentTargets
```

?

packages/react-dom/src/events/accumulateEnterLeaveListeners.js
trueadm: Actually scrap what I said above, we call `getListener` later and that again checks for `stateNode` and if it's `null` then we don't fire any user code. So this change should be fine and have no obserable behavior changes.

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
trueadm: I don't want to change from `dispatchInstances` as that's what the existing legacy system uses and changing that is far too risky. I can change the `currentContainers` to `_dispatchCurrentTargets` though.

packages/react-dom/src/events/accumulateTwoPhaseListeners.js
```diff
@@ -31,20 +31,22 @@ export default function accumulateTwoPhaseListeners(
...
 
   // Accumulate all instances and listeners via the target -> root path.
-  while (node !== null) {
-    const {stateNode: instance, tag} = node;
+  while (instance !== null) {
+    const {stateNode, tag} = instance;
     // Handle listeners that are on HostComponents (i.e. <div>)
-    if (instance !== null && tag === HostComponent) {
-      lastHostComponent = instance;
+    if (stateNode !== null && tag === HostComponent) {
```
gaearon, line 47: Let's make tag checks first here and elsewhere. We don't care about other state nodes anyway.

packages/legacy-events/ReactSyntheticEventType.js
```diff
@@ -40,8 +40,11 @@ export type ReactSyntheticEvent = {|
     nativeEventTarget: EventTarget,
   ) => ReactSyntheticEvent,
   isPersistent: () => boolean,
-  _dispatchInstances: null | Array<Fiber | EventTarget> | Fiber | EventTarget,
-  _dispatchListeners: null | Array<Function> | Function,
+  isPropagationStopped: () => boolean,
+  _dispatchInstances: null | Array<Fiber | null>,
```
gaearon, line 44: OK, we're back to these being strict. This is a lot clearer IMO.

Now can we revert the changes to `executeDispatch` in `EventPluginUtils` from https://github.com/facebook/react/commit/a600408b284e92ed3d801d7f4e37d10b5da06b10?

I suggest:

```diff
- event.currentTarget =
-    inst.tag !== undefined ? getNodeFromInstance(inst) : inst;
+ event.currentTarget = getNodeFromInstance(inst);
```

I guess this isn't even used by the modern code path anymore.

gaearon: When did it stop being possible to have a single `Fiber` in there? Is it still possible in legacy system?

trueadm: In the legacy system it is, but not in the modern system. Given we've forked it, the old code should continue to work. I've re-added the type anyway, even if it's going away very soon.

PR #18466, opened 2020-04-02
Modern Event System: refine flags and handling of enableLegacyFBSupport

trueadm:
This PR fixes some issues with `enableLegacyFBSupport` and the modern event system. I took the time to clean up how flags are passed around, allowing us to handle some specific cases better which were the cause of why some of our tests fail if run with specific www flags enabled. With these changes they should all now pass.

### DIFF
--- a/packages/react-dom/src/events/ReactDOMEventListener.js
+++ b/packages/react-dom/src/events/ReactDOMEventListener.js
@@ -37,12 +37,11 @@ import {
 import {HostRoot, SuspenseComponent} from 'react-reconciler/src/ReactWorkTags';
 import {
   type EventSystemFlags,
+  LEGACY_FB_SUPPORT,
   PLUGIN_EVENT_SYSTEM,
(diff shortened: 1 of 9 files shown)


packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -268,7 +261,14 @@ function dispatchDiscreteEvent(
   container,
   nativeEvent,
 ) {
-  flushDiscreteUpdatesIfNeeded(nativeEvent.timeStamp);
+  if (
+    !enableLegacyFBSupport ||
+    // If we have Legacy FB support, it means we've already
+    // flushed for this event and we don't need to do it again.
+    (enableLegacyFBSupport && (eventSystemFlags & LEGACY_FB_SUPPORT) === 0)
```
gaearon, line 268: `enableLegacyFBSupport &&` is superfluous here. The previous condition already excluded it.

PR #18476, opened 2020-04-03
[Flight] Lazily parse models and allow any value to suspend

sebmarkbage:
There's a bunch of refactoring here but there are two key features that this unlocks:

1) We now parse the model lazily instead of when it arrives. This improves initial perf since not all models will actually be used during initial render, and it allows this work to be properly scheduled along with other React work. This also allow us to suspend when parsing a model. We can use this to allow arbitrary values to suspend. I.e. any value can be replaced with a "reference" (e.g. "$123") and if it's not available, the whole model suspends. Instead of getters/proxies. Blocks are special in that they can suspend at the React component level and therefore don't suspend the whole model.
2) This in turn lets us emit incomplete React elements as future references. Therefore, server components can now suspend and then fill in the content later.

### DIFF
--- a/packages/react-client/src/ReactFlightClient.js
+++ b/packages/react-client/src/ReactFlightClient.js
@@ -87,7 +87,6 @@ Chunk.prototype.then = function<T>(resolve: () => mixed) {
 };
 
 export type ResponseBase = {
-  rootChunk: SomeChunk<any>,
   chunks: Map<number, SomeChunk<any>>,
   readRoot<T>(): T,
   ...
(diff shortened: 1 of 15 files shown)


packages/react-client/src/ReactFlightClient.js
```diff
@@ -69,49 +87,45 @@ Chunk.prototype.then = function<T>(resolve: () => mixed) {
     }
     chunk._value.push(resolve);
   } else {
+    // Uninitialized Chunks should never have been
```
devknoll, line 90: never been...? 😜

Issue #18472, opened 2020-04-02
DevTools: Hovering "Rendered by" list should highlight elements


gaearon:
This list is pretty awesome:

<img width="390" alt="Screenshot 2020-04-03 at 00 12 08" src="https://user-images.githubusercontent.com/810438/78308234-e0047780-753f-11ea-9d4f-1e2d31e5baa0.png">

But always struggle to guess which component in the owner list I need to jump to.

We should make hovering the owner list highlight components, just like the main tree view does.

@hristo-kanchev, interested?

hristo-kanchev:
Sure, @gaearon.

I'll try and have something in place by the end of the week.

I'll keep you updated :)

hristo-kanchev:
Something link this?

![OwnerButton-Hover](https://user-images.githubusercontent.com/23095052/78347574-fd1d6280-75a0-11ea-9d26-1331e8b35c77.gif)

hristo-kanchev:
I've updated the PR and added the highlight native element feature.

👇 

![HighlightRenderedBy](https://user-images.githubusercontent.com/23095052/78379498-c102f580-75d2-11ea-8b26-76fd4589fbb6.gif)

gaearon:
Amazing. This will help me a lot. Thank you so much.

PR #18479, opened 2020-04-03
[DevTools] - Highlight rendered by elements on hover.

hristo-kanchev:
##  Summary

Resolve: #18472 

What's been done:
- Extracted `highlightNativeElement` functionality in its own hook in order to reuse it in multiple components.
- Replaced old highlight code in `Tree` view with the `useHighlightNativeElement` hook.
- Added support for highlighting native elements in the `rendered by` section.
- Added hover state to `rendered by` elements for better visibility on what's being hovered - same as in `Tree` view. (@bvaughn I can revert this part, but I think it's a nice addition).

## Test Plan

Hovering over Rendered By elements successfully highlight the DOM elements.

HighlightRenderedBy

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.css
+++ b/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.css
@@ -114,9 +114,6 @@
 .OwnerContent {
   display: flex;
   align-items: center;
-  padding-top: 0;
-  padding-right: 0;
-  padding-bottom: 0;
   padding-left: 1rem;
   width: 100%;
   border-radius: 0.25rem;
--- a/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.css
+++ b/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.css
@@ -82,7 +82,6 @@
 .Owner {
   border-radius: 0.25rem;
   padding: 0.125rem 0.25rem;
-  cursor: pointer;
   background: none;
   border: none;
   display: block;
@@ -107,10 +106,24 @@
 }
 
 .OwnerButton {
+  cursor: pointer;
+  width: 100%;
+  padding: 0;
+}
+
+.OwnerContent {
   display: flex;
   align-items: center;
-  margin-left: 0.5rem;
-  padding: 0;
+  padding-top: 0;
+  padding-right: 0;
+  padding-bottom: 0;
+  padding-left: 1rem;
+  width: 100%;
+  border-radius: 0.25rem;
+}
+
+.OwnerContent:hover {
+  background-color: var(--color-background-hover);
 }
 
 .ContextMenuIcon {
--- a/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
@@ -26,6 +26,7 @@ import ViewElementSourceContext from './ViewElementSourceContext';
 import NativeStyleEditor from './NativeStyleEditor';
 import Toggle from '../Toggle';
 import Badge from './Badge';
+import {useHighlightNativeElement} from '../hooks';
 import {
   ComponentFilterElementType,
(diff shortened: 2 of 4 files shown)


gaearon:
Thanks for the PR!

I should've been clearer. This might be a good change on its own (I'll defer to @bvaughn on the visual design) but what I'm looking for is **for them to get highlighted in the DOM** on hover, just like highlighting them in the tree does. This way you can move your mouse down to see which grandparent you want to stop at (because it highlights the appropriate area).

Hope this makes sense!

hristo-kanchev:
> Thanks for the PR!
> 
> I should've been clearer. This might be a good change on its own (I'll defer to @bvaughn on the visual design) but what I'm looking for is **for them to get highlighted in the DOM** on hover, just like highlighting them in the tree does. This way you can move your mouse down to see which grandparent you want to stop at (because it highlights the appropriate area).
> 
> Hope this makes sense!

Oh!

I completely misunderstood you. 😆  

Okay, everything is clear now. 

Closing this PR for now.

hristo-kanchev:
Hey @gaearon @bvaughn 

Just reopened the PR again and updated the description as well.

I've added the highlight native element support to the `rendered by` elements and I also kept the hovering state for those elements (the accidental feature 😂).

I can remove the hover styling but I think it's a nice to have. Ping me if it needs reverting and I'll update the PR right away.

Cheers!

bvaughn:
> * Extracted `highlightNativeElement` functionality in its own hook in order to reuse it in multiple components.
> * Replaced old highlight code in Tree view with the `useHighlightNativeElement` hook.

This PR is going to clash with #17934 but that's ... probably fine, whichever one lands first can just rebase on top of the other since the new hooks are the same.

hristo-kanchev:
> > * Extracted `highlightNativeElement` functionality in its own hook in order to reuse it in multiple components.
> > * Replaced old highlight code in Tree view with the `useHighlightNativeElement` hook.
> 
> This PR is going to clash with #17934 but that's ... probably fine, whichever one lands first can just rebase on top of the other since the new hooks are the same.

I'll keep an eye on that one. Thanks for linking the PR!

packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.css
```diff
@@ -107,12 +106,26 @@
...
+}
+
+.OwnerContent {
   display: flex;
   align-items: center;
-  margin-left: 0.5rem;
-  padding: 0;
+  padding-top: 0;
+  padding-right: 0;
+  padding-bottom: 0;
```
bvaughn, line 119: nit: I don't think these three `padding` assignments are required, since `HTMLSpanElement`s don't have any default padding.

packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
```diff
@@ -532,18 +537,26 @@ function OwnerView({
     [dispatch, id],
   );
 
+  const onMouseEnter = () => highlightNativeElement(id);
+
+  const onMouseLeave = () => clearHighlightNativeElement();
```
bvaughn, line 542: nit: This wrapper function isn't necessary :smile:

```suggestion
  const onMouseLeave = clearHighlightNativeElement;
```

packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.css
hristo-kanchev: Woopsie. Those are some leftovers from when I initially started developing the feature. Thanks for making the changes.

packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
hristo-kanchev: We need a `facepalm` reaction on GitHub. 😄

bvaughn: I'd use that one all the time :)

PR #11534, opened 2017-11-12
Use defaultValue instead of setAttribute('value')

nhunzaker:
This commit replaces the method of synchronizing an input's value
attribute from using setAttribute to assigning defaultValue. This has
several benefits:

- Fixes issue where IE10+ and Edge password icon disappears (#7328)
- Fixes issue where toggling input types hides display value on dates
  in Safari (unreported). See [a gif](https://user-images.githubusercontent.com/590904/32701917-13b2911e-c7ac-11e7-8d82-9b777dc61a0f.gif).
- Removes mutationMethod behaviors from DOMPropertyOperations
- Removes the need to re-assign `min`, `max`, and other input props to ensure execution order. value and defaultValue always assign later.

This should replace https://github.com/facebook/react/pull/8266, which is out of date and the fork has been removed :(.

### Test Plan

Visit the fixtures here: http://react-value-attribute-change.surge.sh/

1. Test [text inputs](http://react-value-attribute-change.surge.sh/text-inputs)
2. Test [number inputs](http://react-value-attribute-change.surge.sh/number-inputs)
3. Test [password inputs](http://react-value-attribute-change.surge.sh/password-inputs) in IE10+ and Edge
4. Test [input change events](http://react-value-attribute-change.surge.sh/input-change-events)

I've tested these in the following browsers (oof):

- [x] Chrome 42
- [x] Chrome 61
- [x] IE 9
- [x] IE 10
- [x] IE 11
- [x] Edge 15
- [x] Edge 16
- [x] Firefox 56
- [x] Firefox ESR (47)
- [x] iOS Safari 6
- [x] iOS Safari 10
- [x] iOS Safari 11
- [x] Desktop Safari 11
- [x] Desktop Safari 10
- [x] Desktop Safari 7
- [x] Desktop Safari 6  
- [x] Chrome for Android (latest)

---

This should fix #7328 (https://github.com/facebook/react/pull/8266)

PR #7397, opened 2016-08-01
Fix <input> with type date/time/etc. issue on mobile browsers

keyanzhang:
Fixes #7233. 

In short, iOS Safari acts weird when dealing with HTML5 input types: assigning `node.defaultValue` doesn't let the browser update the UI, so we'll need to set `node.value` to a dummy value then reset it. 

Reviewers: @spicyj @jimfb 

Test Plan:
https://jsbin.com/sefihic/edit?js,output should work on iOS Safari and other mobile browsers (I tested with iOS Safari, iOS Chrome, and Android Browser)

PR #18480, opened 2020-04-03
Update to Jest 25

gaearon:
Revert of #18376. Re-applies #17896.

Bumped the version to include fixes for https://github.com/facebook/jest/issues/9531 and the setter issue @sebmarkbage pointed out (https://github.com/facebook/jest/issues/9745).

Verified our matchers work now.


sebmarkbage:
Can you explain the broken test fix? I don’t get it.

gaearon:
The test was originally written in https://github.com/facebook/react/pull/7397. For non-text input types, mobile browsers didn't honor `setAttribute('value', ...)` alone. So we need to assign `.value` first (and reset it) for later `setAttribute('value', ...)` to be respected. Note that the PR mentions `defaultValue` but that's a red herring. It was about `defaultValue` *React prop* but we were using `setAttribute` to implement it.

Then in https://github.com/facebook/react/pull/11534 we changed the implementation to actually use `.defaultValue =` instead of `setAttribute`. But we still needed this to happen after a `.value` assignment. Otherwise (presumably) it wouldn't be respected. I don't know if this was verified. This test (second one in https://github.com/facebook/react/commit/fd69c239a0bb8fde412aa7c1c0cea9c7fda287a9#diff-ec833522fe8ba426ff9bfe13e609a6e9) was missing a getter so it was causing some comparison to bail and reflected the actual order incorrectly. So it was fixed to listen to `value` assignments correctly and return it. Although it was still kind of broken because it didn't proxy to the real jsdom implementation. In either case, the purpose of this test was to verify we set `.value` first and we `setAttribute` later. (I don't actually know why `setAttribute` gets called there, but I presume it's because setting `.defaultValue` triggers it, and we didn't have a separate `.defaultValue` mock).

Then in https://github.com/facebook/react/pull/17896/files#diff-ec833522fe8ba426ff9bfe13e609a6e9, we upgraded jsdom. That got rid of the bogus `setAttribute` logs. I presume it's because `.defaultValue` is no longer implemented in terms of `setAttribute` in jsdom. However, the test has also lost its purpose because it no longer verifies `.value` is assigned before `.defaultValue`. Because it doesn't mock `.defaultValue` at all.

Then in https://github.com/facebook/react/pull/18376/files#diff-ec833522fe8ba426ff9bfe13e609a6e9 we temporarily added those lines back, but then we skipped this test. I presume Andrew wasn't sure what's up with it.

Now I'm reverting Andrew's skipping. I've added `.defaultValue` mock to validate what the test was originally written to validate. That is, that we set the `.value` first. Again, I don't actually know if the original problem affected `.defaultValue` at all, or if it was confined to `setAttribute`. So maybe this isn't necessary at all.

After I unskipped and fixed it, I've noticed it fails with the `disableAttributeSyncing` flag. This is because I've made that flag run on CI in https://github.com/facebook/react/pull/18461. It didn't run before, but now it runs in variant tests. It failed because it was still asserting `setAttribute` gets called. But jsdom doesn't call `setAttribute` for `.defaultValue` assignments anymore. So I fixed the assertion there, too.

By the way, this bug got fixed in both Chrome and Safari in 2018 so it's possible we can remove this. https://github.com/facebook/react/issues/7233#issuecomment-393165708

wi-ski:
Is this currently whats live on npm? Im chasing - what I believe was - this issue and am still experiencing it on the current npm package.

PR #18485, opened 2020-04-03
[Flight] Add webpack plugin build

sebmarkbage:
Eject the Flight fixture and add it there.

This doesn't actually do anything yet. Just some wiring.


scripts/rollup/bundles.js
```diff
@@ -246,6 +246,15 @@ const bundles = [
     externals: ['react'],
   },
 
+  /******* React DOM Flight Webpack Plugin *******/
+  {
+    bundleTypes: [NODE_DEV, NODE_PROD, UMD_DEV, UMD_PROD],
```
gaearon, line 251: UMD? :D

sebmarkbage: Copy pasta.

PR #18487, opened 2020-04-04
Refactor Error Dialog Logging

sebmarkbage:
This stops passing "componentName" and some unnecessary fields to the custom error dialogs. This information is arguably misleading.

This is the only use case we have for getComponentName in PROD other than getting stacks but I'm going to use a different technique for prod stacks which will let me get rid of it in prod if I just remove this.

You'd think that this wouldn't affect the prod bundle but the indirect way this is wired up means that we include the code that should only be used by FB forks in the npm bundles as unused code. So I cleaned that up as well.

### DIFF
--- a/packages/react-reconciler/src/forks/ReactFiberErrorDialog.native.js
+++ b/packages/react-reconciler/src/forks/ReactFiberErrorDialog.native.js
@@ -7,10 +7,14 @@
  * @flow
  */
 
-import type {CapturedError} from '../ReactCapturedValue';
+import type {Fiber} from '../ReactFiber';
+import type {CapturedValue} from '../ReactCapturedValue';
+
+import {ClassComponent} from '../ReactWorkTags';
 
 // Module provided by RN:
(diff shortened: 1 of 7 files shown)


packages/react-reconciler/src/forks/ReactFiberErrorDialog.native.js
```diff
@@ -7,17 +7,32 @@
...
 
-export function showErrorDialog(capturedError: CapturedError): boolean {
+export function showErrorDialog(
+  boundary: Fiber,
+  errorInfo: CapturedValue<mixed>,
+): boolean {
+  const capturedError = {
+    componentStack: errorInfo.stack !== null ? errorInfo.stack : '',
+    error: errorInfo.value,
+    errorBoundary:
```
gaearon, line 32: I think you can drop this field. https://github.com/facebook/react-native/blob/21733641a8f1da6d6721618a0180509ab2b080a1/Libraries/Core/ReactFiberErrorDialog.js#L31

And maybe remove things from the type on their end.

PR #18491, opened 2020-04-04
Fix bailout broken in lazy components due to default props resolving

### LINKED ISSUE #17151: Bailing out doesn't work properly in lazy components with default props
Bug

Bailing out doesn't work properly in lazy components with default props. It seems we're incorrectly [comparing unresolved props (oldProps) with resolved props (newProps)](https://github.com/facebook/react/blob/62b04cfa753076d5ffb1d74b855f8f8db36f5186/packages/react-reconciler/src/ReactFiberClassComponent.js#L1100).

https://codesandbox.io/s/stoic-curran-3otbb

In the example above, `componentDidUpdate` shouldn't have been called when the button is clicked.

jddxf:
Fix #17151 

We should never compare unresolved props with resolved props. Since comparing resolved props by reference doesn't make sense, we use unresolved props in that case. Otherwise, resolved props are used.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberClassComponent.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.js
@@ -997,11 +997,13 @@ function updateClassInstance(
 
   cloneUpdateQueue(current, workInProgress);
 
-  const oldProps = workInProgress.memoizedProps;
-  instance.props =
+  const unresolvedOldProps = workInProgress.memoizedProps;
+  const oldProps =
     workInProgress.type === workInProgress.elementType
-      ? oldProps
-      : resolveDefaultProps(workInProgress.type, oldProps);
+      ? unresolvedOldProps
+      : resolveDefaultProps(workInProgress.type, unresolvedOldProps);
+  instance.props = oldProps;
+  const unresolvedNewProps = workInProgress.pendingProps;
 
   const oldContext = instance.context;
   const contextType = ctor.contextType;
@@ -1029,7 +1031,10 @@ function updateClassInstance(
     (typeof instance.UNSAFE_componentWillReceiveProps === 'function' ||
       typeof instance.componentWillReceiveProps === 'function')
   ) {
-    if (oldProps !== newProps || oldContext !== nextContext) {
+    if (
+      unresolvedOldProps !== unresolvedNewProps ||
+      oldContext !== nextContext
+    ) {
       callComponentWillReceiveProps(
         workInProgress,
         instance,
@@ -1047,7 +1052,7 @@ function updateClassInstance(
   newState = workInProgress.memoizedState;
 
   if (
-    oldProps === newProps &&
+    unresolvedOldProps === unresolvedNewProps &&
     oldState === newState &&
     !hasContextChanged() &&
     !checkHasForceUpdateAfterProcessing()
@@ -1056,7 +1061,7 @@ function updateClassInstance(
     // effect even though we're bailing out, so that cWU/cDU are called.
     if (typeof instance.componentDidUpdate === 'function') {
       if (
-        oldProps !== current.memoizedProps ||
+        unresolvedOldProps !== current.memoizedProps ||
         oldState !== current.memoizedState
       ) {
         workInProgress.effectTag |= Update;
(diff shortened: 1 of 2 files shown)


bl00mber:
@jddxf you probably may want to change the way you pull updated changes form react original repo to your fork because when you pull them you leave a lot of references in all the other pull requests!!!

Screenshot from 2020-04-05 1

Screenshot from 2020-04-05 2

packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -1056,20 +1061,21 @@ function updateClassInstance(
...
     if (typeof instance.getSnapshotBeforeUpdate === 'function') {
       if (
-        oldProps !== current.memoizedProps ||
+        unresolvedOldProps !== current.memoizedProps ||
         oldState !== current.memoizedState
       ) {
         workInProgress.effectTag |= Snapshot;
       }
     }
+    instance.props = workInProgress.memoizedProps = newProps;
```
gaearon, line 1078: Why is this needed?

jddxf:
@bl00mber Sorry for that. I did force pushed the right commits but that didn't work. I had to delete the fork and it seems ok now.

packages/react-reconciler/src/ReactFiberClassComponent.js
jddxf: Otherwise, React would warn.
https://github.com/facebook/react/blob/1fd45437d781b1b905be2308447aa7793bbcb571/packages/react-reconciler/src/ReactFiberBeginWork.js#L859-L868

This also mirrors the logic at the end of the same function.
https://github.com/facebook/react/blob/1fd45437d781b1b905be2308447aa7793bbcb571/packages/react-reconciler/src/ReactFiberClassComponent.js#L1139-L1151

This was not needed before because it was already guaranteed by the condition in the branch.
https://github.com/facebook/react/blob/1fd45437d781b1b905be2308447aa7793bbcb571/packages/react-reconciler/src/ReactFiberClassComponent.js#L1050

packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -997,11 +997,13 @@ function updateClassInstance(
 
   cloneUpdateQueue(current, workInProgress);
 
-  const oldProps = workInProgress.memoizedProps;
-  instance.props =
+  const unresolvedOldProps = workInProgress.memoizedProps;
```
gaearon, line 1000: This says they're unresolved. But later we do this:

```js
instance.props = workInProgress.memoizedProps = newProps;
```

So next time we *would* have resolved props in `memoizedProps`? I don't quite follow the logic.

jddxf: Since we reset `workInProgress.memoizedProps` to `workInProgress.pendingProps` right after begin work, `workInProgress.memoizedProps` is always unresolved when we begin work.
https://github.com/facebook/react/blob/e69ca310ea56b6f2f510dd463902405285953432/packages/react-reconciler/src/ReactFiberWorkLoop.js#L1501

gaearon: Then can we assign `instance.props` only? Is there a reason for `workInPorgress.memoizedProps` assignment? Sorry I'm being dense, it's been a while since I looked a tthis.

jddxf: Never mind. But unfortunately we have to keep their values in sync. [Here](https://github.com/facebook/react/pull/18491#discussion_r403638636) is the reason.

gaearon: I still don't think it makes sense to assign either of them.

In the sCU branch, `instance.props` reassignment makes sense because the passed props actually *did* change. We want the most recent values inside of them to be accessible to the lifecycle methods, even if sCU "lies" about them not having changed.

In this branch, we've verified props are `===`. So the only reason they "changed" is because we resolved default props. That's not a compelling reason to give our components a different props object.

gaearon: If you change the warning condition to `if (shouldUpdate && inst.props !== nextProps) {` then the warning won't fire. It will also make sense because *the render wasn't called* — so the warning message would have been misleading anyway.

jddxf:
```
instance.props = workInProgress.memoizedProps = newProps;
```
This only gives our components a different props object when we've actually resolved default props. But in that case, components are already given a new props object.
https://github.com/facebook/react/blob/2dddd1e00cdadad1783aa55ac65a28634590e9c1/packages/react-reconciler/src/ReactFiberClassComponent.js#L1001-L1004
I think it's ok to remove this assignment if keeping `workInProgress.memoizedProps` and `instance.props` stay the same until beginWork finishes is not a **requirement**. And that seems true now. So let's change the warning condition instead.

jddxf:
I cannot update this PR now due to refork. Closed in favour of #18539.

Issue #7136, opened 2016-06-28
Refs - "object is not extensible"


biphobe:
I wanted to drop my string-based refs bindings and replace them with callback-based ones, however, when I tried to replace `ref="email"` with `ref={ (ref) => this.refs.email = ref }`, I received an error: `TypeError: Can't add property email, object is not extensible`.

I didn't find any informations about this in docs, but I found other [issue](https://github.com/facebook/react/issues/4934) in which @frederickfogerty did what I tried to do: `ref={ c => this.refs['wrapper'] = c }`

... and I assume that back in 22 Sep 2015 it worked, now it doesn't. Did something change and refs object became nonextensible? Is assigning refs manually through callbacks to `this.refs` considered a bad practice now?

syranide:
AFAIK this was never good practice or recommended. Just store it directly on the instance.

biphobe:
@syranide Can you elaborate? I don't really see how does storing references differs between the `this.refs.email`/`this.email`/`this.dom.email`/`...` variants, they are all stored within the instance.

satya164:
@biphobe `this.refs` is a frozen object. Assign your references to `this` directly.

frederickfogerty:
@biphobe scrolling down in that issue it says that you can't assign to refs, so it didn't work back then either. Sorry for the confusion, I will edit my comment on that issue.

biphobe:
@frederickfogerty You're right, I omitted the bottom half of the comments when scanning the issue, sorry.

Thanks for replies, @syranide, @satya164.

ipip2005:
@satya164 
Seems this error only shows in prod bundle, but omitted in debug bundle. In our code we check-in the change which touch the public refs, but with no errors. After it deployed (which uses React.min.js) it shows this error.
Any clues?

gaearon:
Please provide an example reproducing this.

wzup:
@gaearon 
> _Please provide an example reproducing this._

Here is example that throws very same error
Docs https://reactjs.org/docs/uncontrolled-components.html

```javascript

// I create ref here as in React docs
  constructor(props) {
    super(props);
    this.first_name = React.createRef(); // as in docs
  }

render() {
  return(
        <View style={{marginVertical: 10}}>
          <TextInput
            required={false}
            autoFocus={false}
            ref={ this.first_name } // Then I provide my ref here
            keyboardType={'default'}
            placeholder={'Your first name'}
            defaultValue={ personal.first_name }
          />
        </View>
[...]

PR #18493, opened 2020-04-04
Provide error handling for non-extensible ref assignment

### LINKED ISSUE #16365: ErrorBoundary rendering multiple copies of itself

bug

When a error occurs during the assignment of a `ref` (and maybe other conditions), a error boundary wrapping that error may get confused and it would renders itself multiple times inside the same parent. See https://codesandbox.io/s/stoic-fermi-6etqb which renders:

```html
<div id="root">
  <div class="boundary"><span>content</span></div>
  <div class="boundary"><span>error</span></div>
</div>
```

```html
<div id="root">
  <div class="boundary"><span>error</span></div>
</div>
```

bl00mber:
fix #16365, related #7136


bl00mber:
Because only assignment is wrapped, it should throw as usual. The ref itself is not wrapped to try...catch statement and not executed.

Screenshot from 2020-04-05

gaearon:
I mean a case like this:

```js
<div ref={() => {
  throw new Error('hi')
} />
```

Does it trigger the same problem with error boundaries?

If yes, we need to fix both.

If not, why?

bl00mber:
`throw...` throws in ref like in `componentDidMount`. The incorrect mechanics is probably related to duplicate of `The above error...` message. These messages are thrown from refs in both **commitAttachRef** and **safelyDetachRef** methods. One excessive message is thrown from **safelyDetachRef** which is called before **safelyCallComponentWillUnmount**. The problem may be in the fact that the ref is attached _despite_ throwing an error and then called again within **safelyDetachRef**. I've tried to fix it by removing incorrect ref from `finishedWork` so it not needs to be detached because nothing was attached.

For example, if ref throws is such a way
`(elem) => {this.inputField = elem; throw new Error('error')}`
it needs to detached through `ref(null)` anyway because it throws after ref attachment.

Does it seems correct?

## Before

1  Screenshot from 2020-04-05 errors

2  Screenshot from 2020-04-05 stacks

## After

3  Screenshot from 2020-04-06 CallbackThrowUpdated

bl00mber:
up

53jk1 (approved):

PR #18495, opened 2020-04-04
Refactor Component Stack Traces

sebmarkbage:
I'm going to play with a different way to extract prod stack traces from native stack traces as well as building Hermes specific ones. This doesn't do that yet but it's setting up some refactoring for it.

Notably this no longer uses the getComponentName helper. This means that getComponentName is now only DEV only and can be safely be DCE. Component stacks are the only way we expose names in PROD, outside of DEV tools.

This approach also relies on tag instead of a .type field existing. Some fibers don't need that type field for anything except this use case.

I also ensured to carefully wrap things in `__DEV__` that removes some of the paths related to "source info" that was also included in prod even though that's no longer possible in the jsxDEV transform.


packages/react-reconciler/src/ReactFiberComponentStack.js
```diff
@@ -0,0 +1,72 @@
...
+} from 'shared/ReactComponentStackFrame';
+
+function describeFiber(fiber: Fiber): string {
+  const owner: null | Function = __DEV__
+    ? fiber._debugOwner
+      ? fiber._debugOwner.type
+      : null
+    : null;
+  const source = __DEV__ ? fiber._debugSource : null;
+  switch (fiber.tag) {
```
sebmarkbage, line 38: The list of things that get included was kind of arbitrary and not fully consistent before because it was filtered in Fiber but not in SSR. This list seems like the most commonly useful.

packages/shared/ReactComponentStackFrame.js
```diff
@@ -0,0 +1,145 @@
...
+
+function describeComponentFrame(
+  name: null | string,
+  source: void | null | Source,
+  ownerName: null | string,
+) {
+  let sourceInfo = '';
+  if (source) {
+    const path = source.fileName;
+    let fileName = path.replace(BEFORE_SLASH_RE, '');
```
sebmarkbage, line 32: This should probably be wrapped in DEV.

sebmarkbage:
The "in StrictMode" approach doesn't really work because it doesn't apply to Concurrent and Batched mode environments which are also strict. Which we'll likely slowly adopt even in the legacy environments that are not considered modern. So I'm not sure what that is supposed to do exactly.

gaearon:
We could read the current fiber’s mode and pass it from the fork.

bvaughn:
Why should we expose any info about the mode though?

sebmarkbage:
I started down an implementation like this but I don't think mode is really a generally applicable thing here. It's really arbitrary what kind of filtering you want based on your adoption strategy. Seems like some kind of explicit Context is a more appropriate approach here.

sebmarkbage:
I guess you can just add a wrapper component of any unique name and use that name as the magic string.

gaearon:
D20870861

packages/react/src/__tests__/forwardRef-test.js
```diff
@@ -242,6 +242,33 @@ describe('forwardRef', () => {
     ).toErrorDev(
       'Warning: Failed prop type: The prop `required` is marked as required in ' +
         '`Foo`, but its value is `undefined`.\n' +
+        '    in Unknown (at **)',
+    );
+  });
+
+  it('should honor a displayName in stacks if set on the inner function', () => {
```
gaearon, line 249: This will make some people unhappy. Can we make it work with the outer one too in DEV? For all the wrappers: lazy, memo, forwardRef.

sebmarkbage: What is worse is that Firefox doesn't respect displayName in stacks at all so none of them will be respected in Firefox. However, that's also true for regular error stacks so they should just fix that there. The goal is to be no better or worse than native stacks - and ideally just like native stacks.

I don't really want to add code to parse and try to piece together a line by trying to inject a name in the middle of the browser native format somehow.

What we could do is transfer the displayName onto the inner function if one is not already defined. This would be invasive in the sense that it's mutating an object not owned by React.

sebmarkbage: I added a forwarding setter in DEV. https://github.com/facebook/react/pull/18495/commits/9054ad29cd844b0dba230dce0572a52dd75c8a53

Not for lazy though because you're suppose to use that on the consuming side, and that's not where you specify things like propTypes or displayName. Not sure they should really be used with forwardRef/memo neither. It should really be the inner function that you name.

sebmarkbage: Ofc we had a test that reused a component. This trick doesn't work in this case. I think it's unusual though. https://github.com/facebook/react/pull/18495/commits/544983849e32e4a2c58ac775880346378323fb18

packages/shared/ReactComponentStackFrame.js
```diff
@@ -0,0 +1,142 @@
...
+  }
+  return describeComponentFrame(name, source, ownerName);
+}
+
+export function describeClassComponentFrame(
+  ctor: Function,
+  source: void | null | Source,
+  ownerFn: void | null | Function,
+): string {
+  return describeFunctionComponentFrame(ctor, source, ownerFn);
```
bvaughn, line 69: This looks like a bug. `ownerFn` should not be passed directly. Why didn't Flow catch this?

bvaughn: Oh. Whoops. I misread this as calling `describeComponentFrame`. My bad.

PR #18496, opened 2020-04-05
[eslint] Consider `useRef() as ...` as constant

sophiebits:
Sometimes you need to use casts, eg: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/28884#issuecomment-471341041. This change ignores them and allows you to omit the ref object from the deps list.

Test Plan: unit tests

### DIFF
--- a/packages/eslint-plugin-react-hooks/package.json
+++ b/packages/eslint-plugin-react-hooks/package.json
@@ -25,10 +25,13 @@
     "url": "https://github.com/facebook/react/issues"
   },
   "engines": {
-    "node": ">=7"
+    "node": ">=10"
   },
   "homepage": "https://reactjs.org/",
   "peerDependencies": {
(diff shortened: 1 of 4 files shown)


mertxe78:
@codesandbox[bot]

packages/eslint-plugin-react-hooks/package.json
```diff
@@ -25,10 +25,13 @@
...
   },
   "engines": {
-    "node": ">=7"
+    "node": ">=10"
   },
   "homepage": "https://reactjs.org/",
   "peerDependencies": {
     "eslint": "^3.0.0 || ^4.0.0 || ^5.0.0 || ^6.0.0"
+  },
+  "dependencies": {
```
gaearon, line 34: devDeps?

sophiebits: fixed

PR #18498, opened 2020-04-05
Fix performance issue in react-devtools when highlight enabled

### LINKED ISSUE #17935: Bug: Excessive cpu usage of the page when react-devtools is active
When option "Highlight updates when components render" is activated the whole page repaints in rapid succession after the components state has been changed. It causes 100% CPU usage by the browser and unpleasant DX due low fps.

React version: 16.12.0
DevTools version 4.4.0-f749045a5

The sequance of actions is important:
1. Open react application
2. Open react-devtools
3. Check option "Highlight updates when components render" in react-devtools settings
4. Change the internal state of a component
5. In activity monitor there will be 100% cpu usage of the page, or check option "Paint flashing" in rendering pane of chrome devtools

image
image

The code example (to trigger the issue: 1) check the option "Highlight updates when components render" and 2) click on the button:

```
import React, {useState} from 'react';

function App() {
  const [count, setCount] = useState(0);

  return (
    <button onClick={() => setCount(count + 1)}>
      click #{count}
    </button>
  );
}

export default App;
```

## The current behavior
Excessive cpu usage of the page

## The expected behavior
Normal cpu usage of the page

Faelivrinx:
fix #17935 

## Problem
Problem with high CPU usage when option `Highlight updates when components render` by passing `Number.MAX_VALUE` to `setTimeout` argument. This causes an integer overflow (when using delays larger than 2,147,483,647ms), resulting in the timeout being executed immediately and continuous redrawing canvas with 1ms rate. 

## Solution
Condition prevent to invoke `setTimeout` when `expirationTimeout` (default `NEGATIVE_INFINITY`) is negative number.
```javascript
if (expirationTimeout > 0) {
    redrawTimeoutID = setTimeout(prepareToDraw, expirationTimeout - now);
  }
```
Using `Number.MAX_VALUE` as a default could be more confusing and required condition like`expirationTimeout !== Number.MAX_VALUE`. At first look it's hard to verify what does it mean.

In my opinion timeout should be set only if required expiration time has a value greater than zero and it's more intuitive.

### DIFF
--- a/packages/react-devtools-shared/src/backend/views/TraceUpdates/index.js
+++ b/packages/react-devtools-shared/src/backend/views/TraceUpdates/index.js
@@ -114,21 +114,21 @@ function prepareToDraw(): void {
   redrawTimeoutID = null;
 
   const now = getCurrentTime();
-  let expirationTimeout = Number.NEGATIVE_INFINITY;
+  let earliestExpiration = Number.MAX_VALUE;
 
   // Remove any items that have already expired.
   nodeToData.forEach((data, node) => {
     if (data.expirationTime < now) {
       nodeToData.delete(node);
     } else {
-      expirationTimeout = Math.max(expirationTimeout, data.expirationTime);
+      earliestExpiration = Math.min(earliestExpiration, data.expirationTime);
     }
   });
 
   draw(nodeToData);
 
-  if (expirationTimeout > 0) {
-    redrawTimeoutID = setTimeout(prepareToDraw, expirationTimeout - now);
+  if (earliestExpiration !== Number.MAX_VALUE) {
+    redrawTimeoutID = setTimeout(prepareToDraw, earliestExpiration - now);
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/react-devtools-shared/src/backend/views/TraceUpdates/index.js
```diff
@@ -114,20 +114,22 @@ function prepareToDraw(): void {
...
-  let earliestExpiration = Number.MAX_VALUE;
+  let expirationTimeout = Number.NEGATIVE_INFINITY;
 
   // Remove any items that have already expired.
   nodeToData.forEach((data, node) => {
     if (data.expirationTime < now) {
       nodeToData.delete(node);
     } else {
-      earliestExpiration = Math.min(earliestExpiration, data.expirationTime);
+      expirationTimeout = Math.max(expirationTimeout, data.expirationTime);
```
gaearon, line 124: This is wrong though because out of several nodes, it would pick the *later* expiration time. But we want to pick the *earliest* one.

gaearon:
Looks good. Did you verify this solves the bug?

Faelivrinx:
> Looks good. Did you verify this solves the bug?

Yes, CPU usage are definitely less  on my machine after change

PR #18503, opened 2020-04-06
React Event System: cleanup plugins + break out update batching logic

trueadm:
This PR cleans up a bunch of things with the existing plugin event system that should simplify things and also remove some of the indirection that exists when we try to share things between different renderers to keep things DRY but actually ship irrelevant parts of the code with that when we do.

- When the modern event system flag is enabled we now staticlly inject the event plugin modules – removing the need to recompute the list of plugins dynamically. In the future this will complement https://github.com/facebook/react/pull/18483.
- `ReactGenericBatching.js` has been forked for ReactDOM. That version is called `ReactDOMUpdateBatching.js` and contains all the ReactDOM parts, which have been removed from `ReactGenericBatching.js`.
- `ReactControlledComponent.js` has been moved into ReactDOM and has been renamed `ReactDOMControlledComponent.js` to reflect that no other renderer, but ReactDOM uses it.
- Updated `ReactTestUtils` with an inlined version of `runEventsInBatch` and removed the export from ReactDOM internals.

### DIFF
--- a/packages/legacy-events/EventPluginRegistry.js
+++ b/packages/legacy-events/EventPluginRegistry.js
@@ -241,3 +241,20 @@ export function injectEventPluginsByName(
     recomputePluginOrdering();
   }
 }
+
+export function injectEventPlugins(
+  eventPlugins: [PluginModule<AnyNativeEvent>],
+): void {
+  for (let i = 0; i < eventPlugins.length; i++) {
+    const pluginModule = eventPlugins[i];
+    plugins.push(pluginModule);
+    const publishedEvents = pluginModule.eventTypes;
+    for (const eventName in publishedEvents) {
+      publishEventForPlugin(
+        publishedEvents[eventName],
+        pluginModule,
+        eventName,
+      );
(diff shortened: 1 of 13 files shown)


gaearon:
>When the modern event system flag is enabled we now staticlly inject the event plugin modules – removing the need to recompute the list of plugins dynamically. We also no longer inject the ResponderPlugin as it was recently removed. This should remove a chunk of code when the flag is enabled by default.

Where does this happen?

trueadm:
@gaearon Ah, well spotted. `ReactDOMClientInjection.js` which I now included.

necolas:
> We also no longer inject the ResponderPlugin as it was recently removed.

What are you saying here? This plugin wasn't injected by react-dom before and hasn't been removed either.

trueadm:
> What are you saying here? This plugin wasn't injected by react-dom before and hasn't been removed either.

Ah, I thought https://github.com/facebook/react/pull/18483 was merged already.

PR #18509, opened 2020-04-06
Feature/#18504 create dangerous string for styles

### LINKED ISSUE #18504: Expose API like `createStyles` for converting style object to CSS string
## Feature Request

provide API on react-dom to convert style object to CSS string. the API could be used to build dynamic CSS easily.

### Approach 1

```js
import {createCSS} from 'react-dom'

const inlineStyleString = createCSS({
  overflow: 'hidden', 
  display: '-webkit-box',
  WebkitLineClamp: 2,
})

return <style>{`.clamp-text { ${inlineStyleString} }`}</style>
```

### Approach 2

Another approach is to only map the key and value from the original style object, and let user play with it.

```js
import {createStyles} from 'react-dom'

// return a object with key-value pairs of css rules
const cssStyleObject = createCSS({
  WebkitTransform: 'scale(2)',
})
// return { '-webkit-transform': 'scale(2)'  }

const inlineStyleString = Object.keys(cssStyleObject).reduce((serialized, key) => {
  serialized += `${key}: ${cssStyleObject[key]};`
  return serialized
}, '')

return <style>{`.clamp-text { ${inlineStyleString} }`}</style>
```

## Why

See other react styling library like **radium**, the way to build css is quite similar with react inline style if they need anything aligned with react like browser prefix such as `Webkit` or detect the unit less number for some special rule such as `line-height`, they have to re-implement the logic.

the style object in react is quite convenient, but it's can only used for inline style. hope react team could consider to expose it in the react-dom. might not the origin function name, but the same functionality.

SimonHarmonicMinor:
Resolves #18504
Defines the function that converts style object to style string. Can be useful to define custom css classes with `<style>` tag or passing styles directly to elements.

### DIFF
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -69,6 +69,7 @@ import {
   setAttemptHydrationAtCurrentPriority,
   queueExplicitHydrationTarget,
 } from '../events/ReactDOMEventReplaying';
+import createDangerousStringForStyles from './createDangerousStringForStyles';
 
 setAttemptSynchronousHydration(attemptSynchronousHydration);
 setAttemptUserBlockingHydration(attemptUserBlockingHydration);
@@ -215,6 +216,7 @@ export {
   unstable_createPortal,
   // enableUseEventAPI
   useEvent as unstable_useEvent,
+  createDangerousStringForStyles
 };
 
 const foundDevTools = injectIntoDevTools({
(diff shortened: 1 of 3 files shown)


gaearon:
Thanks for the PR, but generally saying you need to submit an [RFC](https://github.com/reactjs/rfcs) first since there hasn't been much discussion about this feature yet.

SimonHarmonicMinor:
> Thanks for the PR, but generally saying you need to submit an [RFC](https://github.com/reactjs/rfcs) first since there hasn't been much discussion about this feature yet.

Hello! I haven't known about this flow of discussing new features. Got it now.

PR #18510, opened 2020-04-06
Profiler tooltip shows self duration

bvaughn:
The tooltip duration was previously showing less important duration info (about the commit itself, not the thing being hovered). This PR fixes that! I also shifted things around a little to put the more important info (IMO) first.

### No advanced metrics
profiler-tooltip-before-Kapture 2020-04-06 at 9 13 45

### With advanced metrics
devtools-profiiler-before-plus-Kapture 2020-04-06 at 9 23 04

### No advanced metrics
devtools-profiler-after-Kapture 2020-04-06 at 9 21 58

### With advanced metrics
devtools-profiler-after-with-Kapture 2020-04-06 at 9 33 29

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/HoveredFiberInfo.css
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/HoveredFiberInfo.css
@@ -27,7 +27,7 @@
 }
 
 .CurrentCommit {
-  margin-top: 0.25rem;
+  margin: 0.25rem 0;
   display: block;
   width: 100%;
   text-align: left;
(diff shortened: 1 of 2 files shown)


gaearon:
Omg this is so much clearer. I could never understand what the old one was saying but I thought I'm the only one :D

gaearon:
Actually I'm still curious what `Committed at` means in the right pane. Is it the time it took to do a commit? The time of the commit relative to the time you started profiling? Something else?

bvaughn:
I didn't review the initial tooltip PR closely enough. I thought the duration we were already showing was this (for the fiber, rather than the commit). Oops.

> Actually I'm still curious what Committed at means in the right pane. Is it the time it took to do a commit? The time of the commit relative to the time you started profiling? Something else?

Time relative to when you clicked "record" in the profiler, yes.

PR #18513, opened 2020-04-06
Filter certain DOM attributes (e.g. src) if value is empty string

bvaughn:
### This is a breaking change!
So I have put it behind a new feature flag, `enableFilterEmptyStringAttributesDOM`.

## Scope

This flag affects the following attributes/elements:

| Attribute | Elements |
| --- | --- |
| `action` | `form` |
| `formAction` | `button`, `input` |
| `href` | `a`, `area`, `base`, `link` |
| `src` | `audio`, `embed`, `iframe`, `img`, `input`, `script`, `source,`, `track`, `video` |

### Behavior before
| React | HTML |
| --- | --- |
| `` | `` |

### Behavior after
| React | HTML |
| --- | --- |
| `` | `` |

## Implementation

I am not very familiar with the details of the DOM render, but I think it makes the most sense to add this to the `DOMProperty` file. It seems to fit with the other filtering logic we have there.

This approach has a possible downside, since the logic only applies to *attributes* and does not consider the element type. (It's possible we may want to make an exception for some of the impacted elements I listed above?)

Loosely related to PR #15047

### DIFF
--- a/packages/react-dom/src/shared/DOMProperty.js
+++ b/packages/react-dom/src/shared/DOMProperty.js
@@ -171,9 +171,10 @@ export function shouldRemoveAttribute(
       if (propertyInfo.removeEmptyString && value === '') {
         if (__DEV__) {
           console.error(
-            'Invalid value "" (empty string) for attribute `%s`. ' +
-              'This will be a no-op. Either do not render the element at all ' +
-              'or use `null` instead to indicate an empty value.',
+            'An empty string ("") was passed to the %s attribute. ' +
+              'To fix this, either do not render the element at all ' +
+              'or pass null to %s instead of an empty string.',
+            name,
             name,
           );
         }
--- a/packages/react-dom/src/shared/DOMProperty.js
+++ b/packages/react-dom/src/shared/DOMProperty.js
@@ -172,7 +172,8 @@ export function shouldRemoveAttribute(
         if (__DEV__) {
           console.error(
             'Invalid value "" (empty string) for attribute `%s`. ' +
-              'Use `null` instead to indicate an empty value.',
+              'This will be a no-op. Either do not render the element at all ' +
+              'or use `null` instead to indicate an empty value.',
             name,
           );
         }
--- a/packages/react-dom/src/shared/DOMProperty.js
+++ b/packages/react-dom/src/shared/DOMProperty.js
@@ -7,7 +7,10 @@
  * @flow
  */
 
-import {enableDeprecatedFlareAPI} from 'shared/ReactFeatureFlags';
+import {
+  enableDeprecatedFlareAPI,
+  enableFilterEmptyStringAttributesDOM,
+} from 'shared/ReactFeatureFlags';
 
 type PropertyType = 0 | 1 | 2 | 3 | 4 | 5 | 6;
 
@@ -52,6 +55,7 @@ export type PropertyInfo = {|
   +propertyName: string,
   +type: PropertyType,
   +sanitizeURL: boolean,
+  +removeEmptyString: boolean,
 |};
 
 /* eslint-disable max-len */
@@ -163,6 +167,19 @@ export function shouldRemoveAttribute(
     return false;
   }
   if (propertyInfo !== null) {
+    if (enableFilterEmptyStringAttributesDOM) {
+      if (propertyInfo.removeEmptyString && value === '') {
+        if (__DEV__) {
+          console.error(
+            'Invalid value "" (empty string) for attribute `%s`. ' +
+              'Use `null` instead to indicate an empty value.',
+            name,
+          );
+        }
+        return true;
+      }
+    }
+
     switch (propertyInfo.type) {
       case BOOLEAN:
         return !value;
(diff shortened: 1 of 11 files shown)


sebmarkbage:
Is `<base href="" />` useful?

sebmarkbage:
SSR should probably just work, in the sense that it just removes it from the HTML, since it calls this. https://github.com/facebook/react/blob/103378b1eada44561821b1c22ff54e0537cf9764/packages/react-dom/src/server/DOMMarkupOperations.js#L53

Note sure if that causes problems but I can't think of one.

packages/react-dom/src/shared/DOMProperty.js
```diff
@@ -146,6 +150,15 @@ export function shouldRemoveAttribute(
...
 ): boolean {
+  if (enableFilterEmptyStringAttributesDOM) {
+    if (
+      propertyInfo !== null &&
+      propertyInfo.removeEmptyString === true &&
+      value === ''
+    ) {
+      return true;
+    }
+  }
```
sebmarkbage, line 161: This should move below line 178 here into the `propertyInfo !== null` check so we only do that once.

However, we also probably should not filter this for isCustomComponentTag since custom elements could have all kinds of semantics where this matters. We only try to fix the built-ins.

sebmarkbage:
Should this be accompanied with a warning too since this is likely a bug and you should be setting the value to undefined or null if you want this behavior?

sebmarkbage:
Seems like a reasonable approach that's inline with sanitizeURL.

bvaughn:
> Is <base href="" /> useful?

I don't think so.

[Mozilla says](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base):
> If the document contains no `<base>` elements, `baseURI` defaults to `location.href`.

Based on my own testing (Chrome, Firefox, Edge) adding `<base href="">` does not affect the `document.baseURI` value (still the default `location.href`).

> SSR should probably just work, in the sense that it just removes it from the HTML, since it calls this.

Great. I was hoping that was the case, and tests seemed to comfirm.

packages/react-dom/src/shared/DOMProperty.js
bvaughn: Oh, yeah. Duh.

sebmarkbage:
> Based on my own testing (Chrome, Firefox, Edge) adding <base href=""> does not affect the document.baseURI value (still the default location.href).

What if there's another `<base href="somethingelse" />` before it?

```html
<base href="something/else/" />
<base href="" />
```
vs
```html
<base href="something/else/" />
<base />
```

bvaughn:
> What if there's another `<base href="somethingelse" />` before it?

Same behavior. The empty one does not modify the `document.baseURI` value.

bvaughn:
[The spec](https://www.w3.org/TR/html52/document-metadata.html#the-base-element) says:

> There must be no more than one `base` element per document.

JeromeDeLeon:
What about `class` attribute? Can it be added?

sebmarkbage:
The warning should probably recommend not rendering the element at all as the primary recommendation and only null if that doesn’t work (i.e. if someone is too lazy to refactor).

sebmarkbage:
I could see someone using `<a href=“”>` for a reload link or open this page in a new tab.

bvaughn:
> The warning should probably recommend not rendering the element at all as the primary recommendation and only null if that doesn’t work (i.e. if someone is too lazy to refactor).

Yeah, that sounds like kind of a long and confusing warning :wink: I can change the rec. to "don't render at all" though.

> I could see someone using `<a href=“”>` for a reload link or open this page in a new tab.

Yeah... that's the use case that seems most likely I guess. They could always use `<a href=“#”>` instead.

> What about `class` attribute? Can it be added?

@JeromeDeLeon No. This is something different.

bvaughn:
I dunno. Is this warning better?

> Invalid value "" (empty string) for attribute `%s`.  This will be a no-op. Either do not render the element at all or use `null` instead to indicate an empty value.

koenpunt:
I'm wondering why it would be necessary to filter these?

Zeko369:
@koenpunt to prevent unnecessary network requests, for example `` would make a GET request to the current url

gaearon:
@bvaughn We were recently going through React warnings with @rachelnabors and she mentioned “no-op” is unfamiliar jargon. We’ll need to take a pass over them. But for now, I suggest:

>An empty string ("") was passed to the `%s` attribute. This may cause the browser to download the whole page again over the network. To fix this, either do not render the element at all or pass null to `%s` instead of an empty string.

bvaughn:
Sounds good, Dan.

bvaughn:
I appreciate the re-worked error message, @gaearon! These things are difficult to write 😅 

That being said, I think the middle sentence of the one you mentioned is too `img`+`src` specific. Happy to use the first and third sentence though.

gaearon:
>I think the middle sentence of the one you mentioned is too img+src specific

Maybe, but can we leave this wording for `src` then? The reason I'm asking is because it looks like we're just being pedantic. It's not obvious from the warning just how bad it is.

bvaughn:
Sure.

PR #18515, opened 2020-04-07
Clear finished discrete updates during commit phase

### LINKED ISSUE #18020: Bug: event.preventDefault is wrecking havoc with startTransition
React version: 241c446

## Steps To Reproduce

https://codesandbox.io/s/romantic-gates-2hrjp

1. Click "Show A" to render the lazy component.  
2. Type into the input whose placeholder says it works. startTransition will call, inline loading will show for 2 seconds, then the Suspense boundary will trigger.
3. Click the "update" button - same as above - everything works
4. Now type into the input whose placeholder says it *doesn't* work. Note that the inline placeholder stays up for the entire time; the Suspense boundary never shows.
5. Now note that steps 2 or 3 will likely immediately show the suspense boundary. The bypassed rendering of the Suspense boundary from step 4 seems to be backed up, stuck somewhere, and is now unleashed.

Note that steps 2 and 3 are optional. No matter what, the preventDefault that's called in step 4 causes this incorrect behavior no matter whether steps 2 or 3 have been exercised. 

Link to code example:  see above

## The current behavior

startTransition should suspend after the timeout even if event.preventDefault is called.

## The expected behavior

it does not

jddxf:
Fix #18020 

If a root is finished at a priority lower than that of the latest pending discrete updates on it, these updates must have been finished so we can clear them now. Otherwise, a later call of `flushDiscreteUpdates` would start a new empty render pass which may cause a scheduled timeout to be cancelled.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -1215,6 +1215,9 @@ function handleError(root, thrownValue) {
       resetContextDependencies();
       resetHooksAfterThrow();
       resetCurrentDebugFiberInDEV();
+      // TODO: I found and added this missing line while investigating a
+      // separate issue. Write a regression test using string refs.
+      ReactCurrentOwner.current = null;
 
       if (workInProgress === null || workInProgress.return === null) {
         // Expected to be working on a non-root fiber. This is a fatal error
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -1782,6 +1782,19 @@ function commitRootImpl(root, renderPriorityLevel) {
     remainingExpirationTimeBeforeCommit,
   );
 
+  // Clear already finished discrete updates in case that a later call of
+  // `flushDiscreteUpdates` starts a useless render pass which may cancels
+  // a scheduled timeout.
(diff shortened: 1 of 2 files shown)


gaearon:
I'm not super familiar with this code but it seems a bit fragile. Are we treating the symptom or the cause?

You said:

>And no timeout would be scheduled again for some reason. Thus the placeholder is never showed.

Do we understand why?

acdlite:
Thanks for the unit test!

The underlying issue is that we're clearing the timeout without making sure the priority level that corresponds to that level is rescheduled.

It's usually fine to clear a pending timeout at the beginning of a new render (in `prepareFreshStack`) because when we receive an update at a certain priority level, we mark that level and anything lower as "unsuspended." Which means we'll attempt to render the pending level again after the high pri render has finished:

https://github.com/facebook/react/blob/8edcd03b6450ef1d6b78fdb14b8f75682ae0589e/packages/react-reconciler/src/ReactFiberRoot.js#L216-L226

However, `flushDiscreteUpdates` is a case where we start rendering without having received a new update or ping.

There are multiple ways we could fix this. My instinct is to fix it inside `prepareFreshStack`. We should check that the suspended priority level is marked in such a way that we will return to it later. Probably by marking it as pinged.

We should also your changes here, but let's fix the underlying issue first. I'll take a closer look at this.

acdlite:
Thanks again for the fix! 🎉

PR #18521, opened 2020-04-07
Experiment: investigate if clearing alternate improves GC profile

trueadm:
Here's maybe a controversial PR, but one I think we should use to assist our internal investigation. I've had a hunch for a long time that we sometimes leak fibers on the alternates. Although this change is not a fix for the underlying issue, if this proves to improve the memory profile, then it can point us in a direction and maybe provide temporary cover.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1117,6 +1117,8 @@ function detachFiber(current: Fiber) {
   current.stateNode = null;
   current.alternate = null;
   if (alternate !== null) {
+    // Stop the alternate from going back to the current fiber.
+    alternate.alternate = null;
     detachFiber(alternate);
   }
 }
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1115,7 +1115,6 @@ function detachFiber(current: Fiber) {
   current.pendingProps = null;
   current.memoizedProps = null;
   current.stateNode = null;
-  current.alternate = null;
   if (alternate !== null) {
     // Stop the alternate from going back to the current fiber.
     alternate.alternate = null;
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1115,6 +1115,7 @@ function detachFiber(current: Fiber) {
   current.pendingProps = null;
   current.memoizedProps = null;
   current.stateNode = null;
+  current.alternate = null;
   if (alternate !== null) {
     detachFiber(alternate);
   }


packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1115,6 +1115,7 @@ function detachFiber(current: Fiber) {
   current.pendingProps = null;
   current.memoizedProps = null;
   current.stateNode = null;
+  current.alternate = null;
   if (alternate !== null) {
     detachFiber(alternate);
```
bvaughn, line 1120: You'd need to move the null out after this conditional.

bvaughn: Heh. I just noticed that this will call current -> alternate -> current again.

Maybe we should set:
```js
alternate. alternate = null;
```
...before calling?

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1115,7 +1115,10 @@ function detachFiber(current: Fiber) {
   current.pendingProps = null;
   current.memoizedProps = null;
   current.stateNode = null;
+  current.alternate = null;
```
sebmarkbage, line 1118: This is already set earlier.

trueadm: LOL

trueadm: I removed that line. I'll keep the change Brian mentioned.

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1116,6 +1116,8 @@ function detachFiber(current: Fiber) {
   current.memoizedProps = null;
   current.stateNode = null;
   if (alternate !== null) {
+    // Stop the alternate from going back to the current fiber.
+    alternate.alternate = null;
```
gaearon, line 1120: This doesn't do anything either, right? We're passing it to `detachFiber` which does the same thing.

trueadm: It does the same thing, but it avoids an extra function call. So in the scheme of things, it's slightly better.

trueadm:
I'll close this for now.

packages/react-reconciler/src/ReactFiberCommitWork.js
bgirard: What about nextEffect? That one isn't cleared. Would it help to clear that one? The leak I'm observing seems that nextEffect is an important link in the (alternate->nextEffect) chain to get to Fiber that are much older than just the previous update.

Could the leak just be that we're missing that one?

trueadm: We can't clear that field here as it's used for traversal to the next effect. If we cleared it, it would break the reconciler work loop. I'm looking into attempting to try and clear it another way now.

packages/react-reconciler/src/ReactFiberCommitWork.js
bvaughn: We should still probably get this change in though :)
https://github.com/facebook/react/pull/18521#discussion_r404907057

bvaughn: Ah, it looks like you've included it on another PR.

PR #18522, opened 2020-04-07
DevTools Profiler: Fix "cannot read property 'memoizedState' of null"

### LINKED ISSUE #18514: Bug: Dev Tools: TypeError: Cannot read property 'memoizedState' of null
React version:

## Steps To Reproduce

Repro steps using confidential internal fb tool provided separately (https://fb.workplace.com/groups/2299331103613797/permalink/2605716262975278/)

Gist of the repro:
1) open app
2) open react dev tools to profiling tab
3) turn on "Record why each component rendered while profiling"
4) interact with app
5) following error occurs: "TypeError: Cannot read property 'memoizedState' of null"
6) inspecting the flamegraph now shows incomplete information

Based on the stack trace, it looks like the issue occurs specifically on this line:
https://github.com/facebook/react/blob/3e94bce765d355d74f6a60feb4addb6d196e3482/packages/react-devtools-shared/src/backend/console.js#L135

bvaughn:
Resolves #18514

cc @bradzacher

### DIFF
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -918,7 +918,7 @@ export function attach(
   }
 
   function didHooksChange(prev: any, next: any): boolean {
-    if (next == null) {
+    if (prev == null || next == null) {
       return false;
     }
 


gaearon:
Why does this fix it? :D

bvaughn:
The line that's throwing in the reported case is [this one](https://github.com/facebook/react/blob/3278d242184a13add3f25f683b77ef9a6a2305f3/packages/react-devtools-shared/src/backend/renderer.js#L933) (because `prev` is null). DevTools expects [this check](https://github.com/facebook/react/blob/3278d242184a13add3f25f683b77ef9a6a2305f3/packages/react-devtools-shared/src/backend/renderer.js#L803) to prevent that call, but for some reason we have a "mount" operation with a non-null `prevFiber` that has null `memoizedState`.

I'm not sure why this is TBH. :frown:

PR #18529, opened 2020-04-07
Investigation: clear all fibers with nextEffect

trueadm:
This PR aims are clearing the `nextEffect` field on fibers that have detatched. We can't clear the `nextEffect` field during the `detatchFiber` stage, because we need this field to traverse through the effect tree. However, we can store the fiber in an array and clear the fields later (if they have a `nextEffect`). We can use this internally to validate if this fixes the memory leak. It's a pretty heavy handed approach, but might help us until we refactor the `nextEffect` logic in the future.

I also made `detachFiber` a bit more optimized so it doesn't unnecessarily call itself again.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1097,7 +1097,7 @@ function commitNestedUnmounts(
   }
 }
 
-function detachFiber(current: Fiber) {
+function detachFiber(current: Fiber, detatchAlternate: boolean) {
   const alternate = current.alternate;
   // Cut off the return pointers to disconnect it from the tree. Ideally, we
   // should clear the child pointer of the parent alternate to let this
@@ -1115,8 +1115,8 @@ function detachFiber(current: Fiber) {
   current.pendingProps = null;
   current.memoizedProps = null;
   current.stateNode = null;
-  if (alternate !== null) {
-    detachFiber(alternate);
+  if (detatchAlternate && alternate !== null) {
+    detachFiber(alternate, false);
   }
 }
 
@@ -1513,7 +1513,7 @@ function commitDeletion(
     // Detach refs and call componentWillUnmount() on the whole subtree.
     commitNestedUnmounts(finishedRoot, current, renderPriorityLevel);
   }
-  detachFiber(current);
+  detachFiber(current, true);
 }
 
 function commitWork(current: Fiber | null, finishedWork: Fiber): void {
--- a/packages/react-reconciler/src/ReactFiberCommitWork.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.js
@@ -1097,8 +1097,7 @@ function commitNestedUnmounts(
   }
 }
 
-function detachFiber(current: Fiber, detatchAlternate: boolean) {
-  const alternate = current.alternate;
+function detachFiber(current: Fiber) {
   // Cut off the return pointers to disconnect it from the tree. Ideally, we
   // should clear the child pointer of the parent alternate to let this
   // get GC:ed but we don't know which for sure which parent is the current
@@ -1109,15 +1108,11 @@ function detachFiber(current: Fiber, detatchAlternate: boolean) {
   current.memoizedState = null;
   current.updateQueue = null;
   current.dependencies = null;
-  current.alternate = null;
   current.firstEffect = null;
   current.lastEffect = null;
   current.pendingProps = null;
   current.memoizedProps = null;
   current.stateNode = null;
-  if (detatchAlternate && alternate !== null) {
-    detachFiber(alternate, false);
-  }
 }
 
 function emptyPortalContainer(current: Fiber) {
@@ -1513,7 +1508,11 @@ function commitDeletion(
     // Detach refs and call componentWillUnmount() on the whole subtree.
     commitNestedUnmounts(finishedRoot, current, renderPriorityLevel);
   }
-  detachFiber(current, true);
+  detachFiber(current);
+  const alternate = current.alternate;
+  if (alternate !== null) {
+    detachFiber(alternate);
+  }
 }
 
 function commitWork(current: Fiber | null, finishedWork: Fiber): void {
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -1946,6 +1949,11 @@ function commitRootImpl(root, renderPriorityLevel) {
       nextEffect.nextEffect = null;
       nextEffect = nextNextEffect;
     }
+    // Clear the nextEffect pointers of an alternates
```
gaearon, line 1952: Do we need the loop above then? Can we unify them?

packages/react-reconciler/src/ReactFiberCommitWork.js
```diff
@@ -1115,8 +1115,8 @@ function detachFiber(current: Fiber) {
   current.pendingProps = null;
   current.memoizedProps = null;
   current.stateNode = null;
-  if (alternate !== null) {
-    detachFiber(alternate);
+  if (detatchAlternate && alternate !== null) {
+    detachFiber(alternate, false);
```
acdlite, line 1119: Instead of boolean, I would remove these lines and call `detachFiber` on each fiber separately.

```js
detachFiber(fiber);
if (fiber.alternate !== null) {
  detachFiber(fiber.alternate);
}
```

packages/react-reconciler/src/ReactFiberWorkLoop.js
gaearon, line 1952: The comment above says:

>If we have passive effects, we'll clear this in flushPassiveEffects.	

So I think we need to fix it there too. Otherwise it'll only work if you don't have passive effects. If you do, we'll keep adding things to the list and not clear them.

trueadm:
@sebmarkbage Unfortunately not, I originally tried that, however we still get fibers that have non-null `nextEffect` fields that have been unmounted but neither those code paths clear down.

packages/react-reconciler/src/ReactFiberWorkLoop.js
trueadm: I'm unsure of where you mean here. Adding the above logic into `flushPassiveEffects` causes issues. Maybe you know more here and can help me?

trueadm: Unfortunately, we still need to keep the loop above. They're looping through separate things. The loop I've added is essentially going through all unmounted fibers and clearing the `nextEffect` pointer. There might be some effect fibers that are not unmounted that still need their effects clearing, which is what the above loop does.

bgirard:
Liang says that this fixed the FiberNode leaks in our 4 navigation test.

sebmarkbage:
@trueadm Is that *because* we're clearing alternate during deletion so we end up losing the one to clear?

trueadm:
@sebmarkbage This is with me changing the deletion phase (`detatchFiber`) so it doesn't `null` out the `alternate`, otherwise it would do nothing.

sebmarkbage:
Could be move detachFiber to those two places? `if (effectTag & Deletion) detachFiber(...)`

That way we don't need the temporary array and all the clean up is consolidated to the last time the fiber is referenced.

packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -2111,6 +2115,14 @@ function commitMutationEffects(root: FiberRoot, renderPriorityLevel) {
       }
       case Deletion: {
         commitDeletion(root, nextEffect, renderPriorityLevel);
+        // If the effect has a nextEffect, add it to our list
+        if (nextEffect.nextEffect !== null) {
```
sebmarkbage, line 2119: Does this check do anything? Seems like it would leak if this was last in the list. The alternate could still have things to clear.

trueadm: That's a good point, let me try your other suggestion first.

trueadm:
@sebmarkbage I've updated the PR as per your above mention, but it seems like tests now fail. Have I done what you asked right?

sebmarkbage:
@trueadm Yea this looks good. It seems like this affects commit phase profiling somehow by clearing things earlier than expected but that seems like any of these fixes might risk. I'm not too familiar with that code.

@bvaughn Do you have anything insight here? Does the profiler need to read the deleted fibers after they've been deleted somehow?

trueadm:
@sebmarkbage I just tested this change internally and the effect list we loop through now to detatch fibers, misses fibers that would be detatched normally via the previous route I took. So it doesn't feel like a suffecient fix to me?

sebmarkbage:
I suspected maybe that was the case but I don't get why. How can that be since commitDeletion is using the same list? https://github.com/facebook/react/blob/ddc4b65cfe17b3f08ff9f18f8804ff5b663788c8/packages/react-reconciler/src/ReactFiberWorkLoop.js#L2142

trueadm:
@sebmarkbage There seems to be a fundamental bug here then, as that list most definitely does not match up with the `nextEffect` list. Maybe if it did, this bug wouldn't event exist?

sebmarkbage:
I think the test failures here might actually be fixes if you read the comments.

trueadm:
@sebmarkbage I've reverted my changes back to before your suggestion to apply `detatchNode` in those two places. This was based off internal testing from Liang that showed that this did not fix the memory leak.

sebmarkbage:
I think it might be related to the runAllPassiveEffectDestroysBeforeCreates flag.

This: https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberWorkLoop.js#L2398

Doesn't get called in this branch:
https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberWorkLoop.js#L2265-L2372

So we never clean it up.

bvaughn:
😦 Sorry

gaearon:
Ohhh so the whole issue might be because the leak fix wasn't ported to that flag?

trueadm:
Yep, I can confirm that after removing the flag, this is no longer issue internally.

bvaughn:
Well good find!

trueadm:
Closing this as it's no longer the right solution.

sebmarkbage:
Interestingly the accidental fix to the profiler timings mentioned above might be interesting to revisit.

PR #18531, opened 2020-04-07
Add experimental DebugTracing logger for internal use

bvaughn:
Ads debug logging capabilities (based on earlier [experimental profiler branch](https://github.com/bvaughn/react/tree/root-event-marks)) behind a new feature flag, `enableDebugTracing`.


packages/react-reconciler/src/ReactFiberThrow.js
```diff
@@ -194,6 +195,13 @@ function throwException(
     // This is a wakeable.
     const wakeable: Wakeable = (value: any);
 
+    if (__DEV__) {
+      if (enableDebugTracing) {
+        const name = getComponentName(sourceFiber.type) || 'Unknown';
+        logComponentSuspended(name);
```
gaearon, line 201: Can we include the thenable? For Relay, it has a name of the query.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1777,6 +1795,18 @@ function dispatchAction<S, A>(
       }
     }
     scheduleUpdateOnFiber(fiber, expirationTime);
+
+    if (__DEV__) {
```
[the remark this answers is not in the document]
bvaughn: Was thinking about potentially different messaging between the two.

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1777,6 +1795,18 @@ function dispatchAction<S, A>(
...
+      if (enableDebugTracing) {
+        const priorityLevel = inferPriorityFromExpirationTime(
+          currentTime,
+          expirationTime,
+        );
+        const label = priorityLevelToLabel(priorityLevel);
+        const name = getComponentName(fiber.type) || 'Unknown';
+        logStateUpdateScheduled(name, label);
```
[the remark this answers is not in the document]
bvaughn: I guess we could. Seems like it could be kind of noisy.

packages/react-reconciler/src/ReactFiberClassComponent.js
```diff
@@ -224,6 +242,18 @@ const classComponentUpdater = {
 
     enqueueUpdate(fiber, update);
     scheduleUpdateOnFiber(fiber, expirationTime);
+
+    if (__DEV__) {
+      if (enableDebugTracing) {
+        const priorityLevel = inferPriorityFromExpirationTime(
```
acdlite, line 248: Since this is DEV-only, you can read the `.priorityLevel` off the update object, without having to infer based on expiration time.

bvaughn: Ooh, good tip.

bvaughn:
Don't care about failing CI right now, just trying out an idea.

`enableDebugTracing` feature flag still turns this feature on and off, but now React won't auto-log anything (e.g. state updates, suspends) unless it's within a `<React.DebugTraceMode>` subtree.

Maybe this helps cut down on the (unrelated?) noise a bit from clients like Comet who have a lot of logs.

bvaughn:
Rebased and added to old/new variants.

packages/react-is/src/ReactIs.js
```diff
@@ -125,6 +127,9 @@ export function isFragment(object: any) {
 export function isLazy(object: any) {
   return typeOf(object) === REACT_LAZY_TYPE;
 }
+export function isDebugTraceMode(object: any) {
```
gaearon, line 130: Let's drop this. Don't want another `isAsyncMode`. :D

packages/react-reconciler/src/ReactFiber.new.js
```diff
@@ -488,6 +490,10 @@ export function createFiberFromTypeAndProps(
           expirationTime,
           key,
         );
+      case REACT_DEBUG_TRACE_MODE_TYPE:
```
gaearon, line 493: Any way to make this branch DEV-only, and fall through to unknown case otherwise?

packages/react-reconciler/src/ReactTypeOfMode.js
```diff
@@ -9,10 +9,11 @@
 
 export type TypeOfMode = number;
 
-export const NoMode = 0b0000;
-export const StrictMode = 0b0001;
+export const NoMode = 0b00000;
+export const StrictMode = 0b00001;
+export const DebugTraceMode = 0b10000;
```
gaearon, line 14: Let's sort them in an increasing order.

packages/react/src/React.js
```diff
@@ -96,6 +97,7 @@ export {
   REACT_FRAGMENT_TYPE as Fragment,
   REACT_PROFILER_TYPE as Profiler,
   REACT_STRICT_MODE_TYPE as StrictMode,
+  REACT_DEBUG_TRACE_MODE_TYPE as DebugTraceMode,
```
gaearon, line 100: Can this export be `undefined` outside of DEV and when the flag is off?

packages/react/index.js
```diff
@@ -59,6 +59,7 @@ export {
   createMutableSource,
   Fragment,
   Profiler,
+  DebugTraceMode,
```
gaearon, line 62: Wouldn't this add it to public API? Let's make sure it stays FB-only or at least fully behind a flag.

Also maybe add unstable_

packages/shared/ReactSymbols.js
```diff
@@ -48,6 +49,7 @@ if (typeof Symbol === 'function' && Symbol.for) {
   REACT_RESPONDER_TYPE = symbolFor('react.responder');
   REACT_SCOPE_TYPE = symbolFor('react.scope');
   REACT_OPAQUE_ID_TYPE = symbolFor('react.opaque.id');
+  REACT_DEBUG_TRACE_MODE_TYPE = symbolFor('react.log_mode');
```
gaearon, line 52: A bit inconsistent

gaearon:
Also need to fix SSR to pass it through.

gaearon:
This is also missing a crucial feature: tracking of Promises. I think it's an easy fix.

Here's what I hacked together:

```js
let wakeableIDs = new WeakMap();
let id = 0;
function getWakeableID(wakeable) {
  if (!wakeableIDs.has(wakeable)) {
    wakeableIDs.set(wakeable, id++);
  }
  return wakeableIDs.get(wakeable);
}

function logComponentSuspended(componentName, wakeable) {
  log(
    '[!] + ' + componentName + " suspended",
    getWakeableID(wakeable),
    wakeable.displayName,
  );
  wakeable.then(() => {
    log('[!] Promise fulfilled',
      getWakeableID(wakeable),
      wakeable.displayName,
    )
  }, () => {
    log('[!] Promise rejected',
      getWakeableID(wakeable),
      wakeable.displayName,
    )
  })
}
```

packages/react-is/src/ReactIs.js
bvaughn: Sure. Although it feels weird leaving `react-is` in an incomplete/broken state.

packages/react-reconciler/src/ReactFiber.new.js
bvaughn: Hm... well... we could fork the entire `switch` statement, but that doesn't seem good.

We could maybe...also pre-filter type, e.g.
```
let filteredType = type;
if (!__DEV__ && type === REACT_DEBUG_TRACE_MODE_TYPE) {
  filteredType = null;
}
```

But neither of those really feel like improvements.

packages/react/index.js
bvaughn: It's only exposed in the experimental (and FB) indexes. This is the main index, not the `stable` one.

packages/shared/ReactSymbols.js
bvaughn: Whoops. I forgot to rename from its initial :)

packages/react-is/src/ReactIs.js
bvaughn: This would also break DevTools display but maybe that's.. okay

packages/react/src/React.js
bvaughn: Yeah I can make it conditional if the feature flag is disabled.

bvaughn:
```js
function logComponentSuspended(componentName, wakeable) {
  log(
    '[!] + ' + componentName + " suspended",
    getWakeableID(wakeable),
    wakeable.displayName,
  );
  wakeable.then(() => {
    log('[!] Promise fulfilled',
      getWakeableID(wakeable),
      wakeable.displayName,
    )
  }, () => {
    log('[!] Promise rejected',
      getWakeableID(wakeable),
      wakeable.displayName,
    )
  })
}
```

Wakeables have `displayNames`? 😮 😕 

Sure I can add resolve/reject tracing

bvaughn:
Ok, back to you!

gaearon:
>Wakeables have displayNames? 😮 😕

Just the Relay ones. :D

gaearon:
What's up with tests?

bvaughn:
> What's up with tests?

I don't know. I still don't have a good understanding of how all of our various test combinations run and how to satisfy each of them I guess. I'll look into it...

packages/react-dom/src/__tests__/ReactDOMServerIntegrationModes-test.js
```diff
@@ -39,6 +39,53 @@ describe('ReactDOMServerIntegration', () => {
     resetModules();
   });
 
+  // Test pragmas don't support itRenders abstraction
+  if (require('shared/ReactFeatureFlags').enableDebugTracing) {
```
gaearon, line 43: Can we align feature flag closer to component name? One way or another. I keep mistyping either of them because "trace" vs "tracing". :D

bvaughn: Haha, yeah sure thing

bvaughn:
> What's up with tests?

I'm not sure *why* but the test failure seems to have been caused by the conditional I added to `ReactSymbols` --- presumably related to importing `ReactFeatureFlags` from that file? I'm not sure. For now, I just backed that change out.

bvaughn:
For fuck sake

packages/react-reconciler/src/ReactFiber.old.js
```diff
@@ -488,6 +490,10 @@ export function createFiberFromTypeAndProps(
           expirationTime,
           key,
         );
+      case REACT_DEBUG_TRACING_MODE_TYPE:
```
gaearon, line 493: Does it mean that if I somehow end up with undefined `type` in a stable release, after this change it would start matching `REACT_DEBUG_TRACING_MODE_TYPE` (which would also be `undefined` in stable), and thus I won't get to the catch-all default case that throws? This seems bad.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -378,6 +389,25 @@ export function computeExpirationForFiber(
   return expirationTime;
 }
 
+export function priorityLevelToLabel(
+  priorityLevel: ReactPriorityLevel,
+): string {
+  switch (priorityLevel) {
```
gaearon, line 395: Let's wrap in feature flag and DEV.

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
```diff
@@ -371,6 +382,25 @@ export function computeExpirationForFiber(
   return expirationTime;
 }
 
+export function priorityLevelToLabel(
+  priorityLevel: ReactPriorityLevel,
+): string {
+  switch (priorityLevel) {
```
gaearon, line 388: Let's wrap in feature flag and DEV.

packages/react/index.js
```diff
@@ -59,6 +59,7 @@ export {
   createMutableSource,
   Fragment,
   Profiler,
+  DebugTracingMode,
```
gaearon, line 62: Does it become enumerable (but undefined) export? If yes, it should be `unstable_`.

packages/react-reconciler/src/ReactFiber.old.js
bvaughn: `REACT_DEBUG_TRACING_MODE_TYPE` isn't undefined in stable, just `React.DebugTracingMode`.

bvaughn: (I backed out the conditional symbol export)

packages/shared/forks/ReactFeatureFlags.testing.www.js
```diff
@@ -11,6 +11,7 @@ import typeof * as FeatureFlagsType from 'shared/ReactFeatureFlags';
 import typeof * as ExportsType from './ReactFeatureFlags.testing.www';
 
 export const debugRenderPhaseSideEffectsForStrictMode = false;
+export const enableDebugTracing = __EXPERIMENTAL__;
```
gaearon, line 14: Let's just make it false.

bvaughn: Hopefully that won't make the test matrix gods angry.

packages/react/index.js
bvaughn: This will never be undefined. It's just not re-exported by the .stable index. (This index file itself is never used directly.)

Issue #18486, opened 2020-04-04
Bug: Dropped update when render phase update happens before suspending


gaearon:
Sandbox: https://codesandbox.io/s/cranky-kapitsa-op3jh

Repro: type "a" and then after a second "b"

Expected: Eventually you'll see "ab" in all fields. Pending "..." indicator goes away.

Actual: You'll keep seeing "a" in low pri field. Pending "..." indicator is stuck.

VakoNonikashvili:
Sandbox: https://codesandbox.io/s/nostalgic-gauss-xrn4o

Why didn't you update mirror text with useEffect?

gaearon:
Generally saying, render phase updates are better for derived state because you don’t need an extra rendering pass for the whole subtree.

In either case, it doesn’t matter how else I could write it. The problem is that this uncovers a bug in React, and bugs need to be fixed.

PR #18534, opened 2020-04-07
Add a failing test for #18486

gaearon:
Failing test for https://github.com/facebook/react/issues/18486.
I think this matches the sandbox issue.

The pending state gets stuck:

Screenshot 2020-04-08 at 00 46 10

If you comment out the render phase update and adjust the Scheduler yield checks, then it doesn't get stuck. So I think this demonstrates the issue but the actual yields would need to be adjusted according to the fix. I don't know where the bug is yet.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseWithNoopRenderer-test.internal.js
@@ -3645,4 +3645,137 @@ describe('ReactSuspenseWithNoopRenderer', () => {
       </>,
     );
   });
+
+  // Regression: https://github.com/facebook/react/issues/18486
+  it.experimental(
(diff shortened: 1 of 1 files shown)


gaearon:
Looks like after resolving `"a"` and rendering it, we're calling `markRootFinishedAtTime` with `remainingExpirationTimeBeforeCommit` being `0`. That sets `root.firstPendingTime` to `0`, so next time `getNextRootExpirationTimeToWorkOn` also [returns `0`](https://github.com/facebook/react/blob/948fad357b41002f7e9139211464470b3c0462e8/packages/react-reconciler/src/ReactFiberWorkLoop.js#L526), and we decide to [never work on `"b"`](https://github.com/facebook/react/blob/948fad357b41002f7e9139211464470b3c0462e8/packages/react-reconciler/src/ReactFiberWorkLoop.js#L565).

PR #18535, opened 2020-04-08
More robust fix for #18515

acdlite:
Adds a regression test for the same underlying bug as #18515 but using pings.

Test already passes, but I confirmed it fails if you revert the fix in #18515.

Also confirmed that this fix alone is sufficient to fix the bug. However, I've kept the previous fix as defense against a future regression, since we've had similar bugs around clearing Suspense timers before.

The fix is to always set `nextKnownPendingLevel` upon finishing a root, instead of only when it suspends.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.js
@@ -679,9 +679,10 @@ function performConcurrentWorkOnRoot(root, didTimeout) {
 
     // We now have a consistent tree. The next step is either to commit it,
     // or, if something suspended, wait to commit it after a timeout.
-    const finishedWork: Fiber = ((root.finishedWork =
-      root.current.alternate): any);
+    const finishedWork: Fiber = (root.current.alternate: any);
+    root.finishedWork = finishedWork;
     root.finishedExpirationTime = expirationTime;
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.js
```diff
@@ -979,9 +974,10 @@ function performSyncWorkOnRoot(root) {
 
   // We now have a consistent tree. Because this is a sync render, we
   // will commit it even if something suspended.
-  root.finishedWork = (root.current.alternate: any);
+  const finishedWork: Fiber = (root.current.alternate: any);
+  root.finishedWork = finishedWork;
   root.finishedExpirationTime = expirationTime;
-
+  root.nextKnownPendingLevel = getRemainingExpirationTime(finishedWork);
```
gaearon, line 980: Is it worth write a regression test that would fail if this *and* the defensive safeguard were removed? Looks like the new test covers the Concurrent render only. Not sure if this matters.

acdlite: The diff view is misleading because of collapsed sections. This line is the sync path. The concurrent path is here: https://github.com/facebook/react/pull/18535/files#diff-63c455475b8566c10993b2daa2c3211bR685

The test #18515 covers the sync path because the discrete update is expired but I can add another test since that one happens to be fixed by the other change.

PR #18537, opened 2020-04-08
Bugfix: Render phase update causes remaining updates in same component to be dropped

### LINKED ISSUE #18486: Bug: Dropped update when render phase update happens before suspending

acdlite:
Fixes #18534 
Fixes #18486

Render phase updates should not affect the `fiber.expirationTime` field. We don't have to set anything on the fiber because we're going to process the render phase update immediately.

We also shouldn't reset the `expirationTime` field in between render passes because it represents the remaining work left in the update queues. During the re-render, the updates that were skipped in the original pass are not processed again.

I think my original motivation for using this field for render phase updates was so I didn't have to add another module level variable.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.js
@@ -230,6 +230,11 @@ let workInProgressHook: Hook | null = null;
 // finished evaluating this component. This is an optimization so we know
 // whether we need to clear render phase updates after a throw.
 let didScheduleRenderPhaseUpdate: boolean = false;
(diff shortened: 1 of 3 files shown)


gaearon:
haha I wasn't too far

Screenshot 2020-04-08 at 03 04 00

Issue #18497, opened 2020-04-05
Bug: Concurrent mode processes state updates out of order


sophiebits:
React version: 0.0.0-experimental-[d7382b6c4](https://github.com/facebook/react/commits/d7382b6c43b63ce15ce091cf13db8cd1f3c4b7ae)

I have a sandbox that adds the next integer to an array every time you move your mouse: https://codesandbox.io/s/compassionate-currying-nccok.

Intermediate state updates get dropped, then come back later (eg: you would never expect it to render with [0, 2]):

![confused](https://user-images.githubusercontent.com/6820/78509992-f7fa1800-7746-11ea-80fb-0db0174ec0b0.png)

My expectation is that these are all the same priority level and thus should get processed in order. Am I doing something wrong?

cc @acdlite @gaearon

sophiebits:
Same issue with UserBlockingPriority: https://codesandbox.io/s/thirsty-neumann-x1wzx.

sophiebits:
Diagnosis in https://github.com/facebook/react/pull/18538#issuecomment-610786771.

gaearon:
Not sure if this is fixed for good but it doesn't repro on master.
https://codesandbox.io/s/youthful-gates-yq0jw

PR #18538, opened 2020-04-08
Add regression test for #18497

sophiebits:


packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js
```diff
@@ -3263,4 +3263,45 @@ describe('ReactHooksWithNoopRenderer', () => {
...
+      expect(Scheduler).toFlushAndYieldThrough(['Render []', 'Commit []']);
+      expect(ReactNoop).toMatchRenderedOutput('[]');
+    });
+
+    act(() => {
+      _dispatch(1);
+      expect(Scheduler).toFlushAndYieldThrough(['Render [1]']);
+
+      _dispatch(2);
+      expect(Scheduler).toFlushAndYieldThrough(['Commit [1]']);
```
sophiebits, line 3293: Not sure if this is expected either. Should the dispatch prevent this one from being committed? I can't remember how this is supposed to work.

sophiebits:
It looks like this logic is to blame:

https://github.com/facebook/react/blob/2dddd1e00cdadad1783aa55ac65a28634590e9c1/packages/react-reconciler/src/ReactFiberWorkLoop.js#L357-L365

It applies to the "2" update but not "3", causing the "2" update to have a lower priority than the "3" update, which is why we render with only [1,3].

gaearon:
Seems like you can work around it by making it discrete: https://codesandbox.io/s/agitated-ramanujan-hwxjg?file=/src/App.js

Seb says this particular case will likely go away after a planned big refactor Andrew is working on.

But also that we don't make very strong guarantees about which updates get skipped. We only make guarantees about terminal value (eventual consistency) and that discrete updates would flush in order (like inmy linked demo).

sophiebits:
My understanding of discrete is you effectively lose the concurrency support. The only reason I started using concurrent mode is that I wanted React to automatically batch the renders (more than one mouse move per render). Changing this to be discrete feels like a little bit of a waste.

So it isn't a guarantee that updates at the same priority are processed in order? That's completely contrary to all my past understanding of how concurrent mode works. I thought we subset only based on a priority threshold.

sophiebits:
(I thought discrete is only meant for if you need it to be flushed before the next event handler runs.)

gaearon:
That’s what I thought too but I’m hearing we are relaxing some of the previous constraints. There are cases where the current model has very unintuitive behavior, and Andrew is working on an alternative one that splits updates into “threads” that aren’t directly tied to priority levels. I haven’t looked in depth at these changes yet so it’s hard for me to summarise them or how exactly the thinking has evolved. Maybe we can ask Seb to do this.

Generally in the past few months we’ve been mostly focused on solving cases that we see come up in practical product usage. So I think that shifted the focus a bit from when we didn’t really know which constraints mattered and which create other issues. I’m curious to learn more about how you ran into this one and what the product case that exhibits the bug looked like.

sophiebits:
Drawing app. Add a new point every mousemove. This bug means that sometimes you are drawing:

image

And then it adds points in the past and the line jumps around:

image

sophiebits:
@gaearon Worth committing a regression test?

gaearon:
https://github.com/facebook/react/pull/19215

PR #18539, opened 2020-04-08
Fix bailout broken in lazy components due to default props resolving

### LINKED ISSUE #17151: Bailing out doesn't work properly in lazy components with default props

jddxf:
Fix #17151 

We should never compare unresolved props with resolved props. Since comparing resolved props by reference doesn't make sense, we use unresolved props in that case. Otherwise, resolved props are used.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -856,7 +856,7 @@ function updateClassComponent(
   );
   if (__DEV__) {
     const inst = workInProgress.stateNode;
-    if (inst.props !== nextProps) {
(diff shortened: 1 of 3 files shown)


gaearon:
Makes sense. Thanks!

PR #18545, opened 2020-04-08
useRef: Warn about reading or writing mutable values during render

bvaughn:
This warning is being tested within Facebook behind a feature flag to assess its impact, (e.g. how many warnings does it trigger, how many patterns (like `usePrevious`) does it flush out, do they all have safe alternate patterns, etc.)

If you have code that unconditionally mutates a ref during render, you might want to go ahead and change it. It can break things (like #19114).

If you have code that reads a ref during render (e.g. the `usePrevious` pattern described above) please don't feel a need to make any changes to that code for now. We'll provide guidance once we've done more research.

---

Reading or writing a ref value during render is only safe if you are implementing [the lazy initialization pattern](https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily).

Other types of reading are unsafe as the ref is a [mutable source](https://github.com/reactjs/rfcs/blob/master/text/0147-use-mutable-source.md).

Other types of writing are unsafe as they are effectively [side effects](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects).

This change also refactors `useTransition` to no longer use a ref hook, but instead manage its own (stable) hook state.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.new.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.new.js
@@ -27,6 +27,7 @@ import {
   enableNewReconciler,
   decoupleUpdatePriorityFromScheduler,
   enableDoubleInvokingEffects,
+  enableUseRefMutationWarning,
 } from 'shared/ReactFeatureFlags';
 
 import {
@@ -1197,14 +1198,92 @@ function pushEffect(tag, create, destroy, deps) {
   return effect;
 }
 
+let stackContainsErrorMessage: boolean | null = null;
+
+function getCallerStackFrame(): string {
+  const stackFrames = new Error('Error message').stack.split('\n');
+
+  // Some browsers (e.g. Chrome) include the error message in the stack
+  // but others (e.g. Firefox) do not.
+  if (stackContainsErrorMessage === null) {
+    stackContainsErrorMessage = stackFrames[0].includes('Error message');
+  }
+
+  return stackContainsErrorMessage
+    ? stackFrames.slice(3, 4).join('\n')
+    : stackFrames.slice(2, 3).join('\n');
+}
+
 function mountRef<T>(initialValue: T): {|current: T|} {
   const hook = mountWorkInProgressHook();
-  const ref = {current: initialValue};
-  if (__DEV__) {
-    Object.seal(ref);
+  if (enableUseRefMutationWarning) {
+    if (__DEV__) {
+      // Support lazy initialization pattern shown in docs.
+      // We need to store the caller stack frame so that we don't warn on subsequent renders.
+      let hasBeenInitialized = initialValue != null;
+      let lazyInitGetterStack = null;
+      let didCheckForLazyInit = false;
+
+      // Only warn once per component+hook.
+      let didWarnAboutRead = false;
+      let didWarnAboutWrite = false;
+
+      let current = initialValue;
+      const ref = {
+        get current() {
+          if (!hasBeenInitialized) {
+            didCheckForLazyInit = true;
+            lazyInitGetterStack = getCallerStackFrame();
+          } else if (currentlyRenderingFiber !== null && !didWarnAboutRead) {
+            if (
+              lazyInitGetterStack === null ||
+              lazyInitGetterStack !== getCallerStackFrame()
+            ) {
+              didWarnAboutRead = true;
+              console.warn(
+                '%s: Unsafe read of a mutable value during render.\n\n' +
+                  'Reading from a ref during render is only safe if:\n' +
+                  '1. The ref value has not been updated, or\n' +
+                  '2. The ref holds a lazily-initialized value that is only set once.\n',
+                getComponentName(currentlyRenderingFiber.type) || 'Unknown',
+              );
+            }
+          }
+          return current;
+        },
+        set current(value) {
+          if (currentlyRenderingFiber !== null && !didWarnAboutWrite) {
+            if (
+              hasBeenInitialized ||
+              (!hasBeenInitialized && !didCheckForLazyInit)
+            ) {
+              didWarnAboutWrite = true;
+              console.warn(
+                '%s: Unsafe write of a mutable value during render.\n\n' +
+                  'Writing to a ref during render is only safe if the ref holds ' +
+                  'a lazily-initialized value that is only set once.\n',
+                getComponentName(currentlyRenderingFiber.type) || 'Unknown',
+              );
+            }
+          }
+
+          hasBeenInitialized = true;
+          current = value;
+        },
+      };
+      Object.seal(ref);
+      hook.memoizedState = ref;
+      return ref;
+    } else {
+      const ref = {current: initialValue};
+      hook.memoizedState = ref;
+      return ref;
+    }
+  } else {
+    const ref = {current: initialValue};
+    hook.memoizedState = ref;
+    return ref;
   }
-  hook.memoizedState = ref;
-  return ref;
 }
 
 function updateRef<T>(initialValue: T): {|current: T|} {
@@ -1591,24 +1670,24 @@ function startTransition(setPending, callback) {
 
 function mountTransition(): [(() => void) => void, boolean] {
   const [isPending, setPending] = mountState(false);
-  // The `start` method can be stored on a ref, since `setPending`
-  // never changes.
+  // The `start` method never changes.
   const start = startTransition.bind(null, setPending);
-  mountRef(start);
+  const hook = mountWorkInProgressHook();
+  hook.memoizedState = start;
   return [start, isPending];
 }
 
 function updateTransition(): [(() => void) => void, boolean] {
   const [isPending] = updateState(false);
-  const startRef = updateRef();
-  const start: (() => void) => void = (startRef.current: any);
+  const hook = updateWorkInProgressHook();
+  const start = hook.memoizedState;
   return [start, isPending];
 }
 
 function rerenderTransition(): [(() => void) => void, boolean] {
   const [isPending] = rerenderState(false);
-  const startRef = updateRef();
-  const start: (() => void) => void = (startRef.current: any);
+  const hook = updateWorkInProgressHook();
+  const start = hook.memoizedState;
   return [start, isPending];
 }
 
--- a/packages/react-reconciler/src/ReactFiberHooks.new.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.new.js
@@ -27,7 +27,7 @@ import {
   enableNewReconciler,
   decoupleUpdatePriorityFromScheduler,
   enableDoubleInvokingEffects,
-  enableUseRefMutationWarning,
+  enableUseRefAccessWarning,
 } from 'shared/ReactFeatureFlags';
 
 import {
@@ -1216,7 +1216,7 @@ function getCallerStackFrame(): string {
 
 function mountRef<T>(initialValue: T): {|current: T|} {
   const hook = mountWorkInProgressHook();
-  if (enableUseRefMutationWarning) {
+  if (enableUseRefAccessWarning) {
     if (__DEV__) {
       // Support lazy initialization pattern shown in docs.
       // We need to store the caller stack frame so that we don't warn on subsequent renders.
--- a/packages/react-reconciler/src/ReactFiberHooks.old.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.old.js
@@ -26,7 +26,7 @@ import {
   enableSchedulingProfiler,
   enableNewReconciler,
   decoupleUpdatePriorityFromScheduler,
-  enableUseRefMutationWarning,
+  enableUseRefAccessWarning,
 } from 'shared/ReactFeatureFlags';
 
 import {NoMode, BlockingMode, DebugTracingMode} from './ReactTypeOfMode';
@@ -1194,7 +1194,7 @@ function getCallerStackFrame(): string {
 
 function mountRef<T>(initialValue: T): {|current: T|} {
   const hook = mountWorkInProgressHook();
-  if (enableUseRefMutationWarning) {
+  if (enableUseRefAccessWarning) {
     if (__DEV__) {
       // Support lazy initialization pattern shown in docs.
       // We need to store the caller stack frame so that we don't warn on subsequent renders.
--- a/packages/react-reconciler/src/ReactFiberHooks.old.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.old.js
@@ -26,6 +26,7 @@ import {
   enableSchedulingProfiler,
   enableNewReconciler,
   decoupleUpdatePriorityFromScheduler,
+  enableUseRefMutationWarning,
 } from 'shared/ReactFeatureFlags';
 
 import {NoMode, BlockingMode, DebugTracingMode} from './ReactTypeOfMode';
@@ -1175,14 +1176,92 @@ function pushEffect(tag, create, destroy, deps) {
   return effect;
 }
 
+let stackContainsErrorMessage: boolean | null = null;
+
+function getCallerStackFrame(): string {
+  const stackFrames = new Error('Error message').stack.split('\n');
+
+  // Some browsers (e.g. Chrome) include the error message in the stack
+  // but others (e.g. Firefox) do not.
+  if (stackContainsErrorMessage === null) {
+    stackContainsErrorMessage = stackFrames[0].includes('Error message');
+  }
+
+  return stackContainsErrorMessage
+    ? stackFrames.slice(3, 4).join('\n')
+    : stackFrames.slice(2, 3).join('\n');
+}
+
 function mountRef<T>(initialValue: T): {|current: T|} {
   const hook = mountWorkInProgressHook();
-  const ref = {current: initialValue};
-  if (__DEV__) {
-    Object.seal(ref);
+  if (enableUseRefMutationWarning) {
+    if (__DEV__) {
+      // Support lazy initialization pattern shown in docs.
+      // We need to store the caller stack frame so that we don't warn on subsequent renders.
+      let hasBeenInitialized = initialValue != null;
+      let lazyInitGetterStack = null;
+      let didCheckForLazyInit = false;
+
+      // Only warn once per component+hook.
+      let didWarnAboutRead = false;
+      let didWarnAboutWrite = false;
+
+      let current = initialValue;
+      const ref = {
+        get current() {
+          if (!hasBeenInitialized) {
+            didCheckForLazyInit = true;
+            lazyInitGetterStack = getCallerStackFrame();
+          } else if (currentlyRenderingFiber !== null && !didWarnAboutRead) {
+            if (
+              lazyInitGetterStack === null ||
+              lazyInitGetterStack !== getCallerStackFrame()
+            ) {
+              didWarnAboutRead = true;
+              console.warn(
+                '%s: Unsafe read of a mutable value during render.\n\n' +
+                  'Reading from a ref during render is only safe if:\n' +
+                  '1. The ref value has not been updated, or\n' +
+                  '2. The ref holds a lazily-initialized value that is only set once.\n',
+                getComponentName(currentlyRenderingFiber.type) || 'Unknown',
+              );
+            }
+          }
+          return current;
+        },
+        set current(value) {
+          if (currentlyRenderingFiber !== null && !didWarnAboutWrite) {
+            if (
+              hasBeenInitialized ||
+              (!hasBeenInitialized && !didCheckForLazyInit)
+            ) {
+              didWarnAboutWrite = true;
+              console.warn(
+                '%s: Unsafe write of a mutable value during render.\n\n' +
+                  'Writing to a ref during render is only safe if the ref holds ' +
+                  'a lazily-initialized value that is only set once.\n',
+                getComponentName(currentlyRenderingFiber.type) || 'Unknown',
+              );
+            }
+          }
+
+          hasBeenInitialized = true;
+          current = value;
+        },
+      };
+      Object.seal(ref);
+      hook.memoizedState = ref;
+      return ref;
+    } else {
+      const ref = {current: initialValue};
+      hook.memoizedState = ref;
+      return ref;
+    }
+  } else {
+    const ref = {current: initialValue};
+    hook.memoizedState = ref;
+    return ref;
   }
-  hook.memoizedState = ref;
-  return ref;
 }
 
 function updateRef<T>(initialValue: T): {|current: T|} {
@@ -1534,24 +1613,24 @@ function startTransition(setPending, callback) {
 
 function mountTransition(): [(() => void) => void, boolean] {
   const [isPending, setPending] = mountState(false);
-  // The `start` method can be stored on a ref, since `setPending`
-  // never changes.
+  // The `start` method never changes.
   const start = startTransition.bind(null, setPending);
-  mountRef(start);
+  const hook = mountWorkInProgressHook();
+  hook.memoizedState = start;
   return [start, isPending];
 }
 
 function updateTransition(): [(() => void) => void, boolean] {
   const [isPending] = updateState(false);
-  const startRef = updateRef();
-  const start: (() => void) => void = (startRef.current: any);
+  const hook = updateWorkInProgressHook();
+  const start = hook.memoizedState;
   return [start, isPending];
 }
 
 function rerenderTransition(): [(() => void) => void, boolean] {
   const [isPending] = rerenderState(false);
-  const startRef = updateRef();
-  const start: (() => void) => void = (startRef.current: any);
+  const hook = updateWorkInProgressHook();
+  const start = hook.memoizedState;
   return [start, isPending];
 }
 
--- a/packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.js
+++ b/packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.js
@@ -1536,7 +1536,7 @@ describe('ReactHooksWithNoopRenderer', () => {
     it('does not show a warning when a component updates a childs state from within passive unmount function', () => {
       function Parent() {
         Scheduler.unstable_yieldValue('Parent');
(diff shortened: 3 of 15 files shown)


packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1233,12 +1233,36 @@ function pushEffect(tag, create, destroy, deps) {
...
+    let shouldWarnAboutRead = false;
+    const ref = {
+      get current() {
+        if (shouldWarnAboutRead && currentlyRenderingFiber !== null) {
+          console.warn(
+            'useRef: Unsafe read of a mutable value during render.\n\n' +
+              'Reading from a ref during render is only safe if:\n' +
+              '1. The ref value has not been updated, or\n' +
+              '2. The ref holds a lazily-initialized value that is only set once.',
+          );
```
bvaughn, line 1247: I guess I should add component name and stack heh

markerikson:
Just to be clear, this only kicks if you both write _and_ read while rendering?  _Just_ reading is safe?

Does this warning also only kick in if they both happen within the same function component execution (ie, resets after rendering is done)?

packages/react-reconciler/src/ReactFiberHooks.js
```diff
@@ -1233,12 +1234,39 @@ function pushEffect(tag, create, destroy, deps) {
...
+              '1. The ref value has not been updated, or\n' +
+              '2. The ref holds a lazily-initialized value that is only set once.\n\n%s',
+            getComponentName(fiber.type) || 'Unknown',
+            getStackByFiberInDevAndProd(fiber),
+          );
+        }
+        return current;
+      },
+      set current(value) {
+        if (!shouldWarnAboutRead && current !== null) {
```
sompylasar, line 1256: Would this be valid if `current === undefined` (not `null`)? I've seen `useRef()` without args, even though `mountRef` requires an argument, it can be of void type.

Should it warn on writes within render but not within effect if `current !== initialValue` and `value !== initialValue`?

gaearon:
>Just to be clear, this only kicks if you both write and read while rendering? Just reading is safe?

No, it’s reading itself that is a problem. Since it’s effectively the same as reading from a random global variable. It is non-deterministic because whatever you’re going to get depends on when render was called. If React calls render at a slightly different time you can have a different result.

gaearon:
> Does this warning also only kick in if they both happen within the same function component execution (ie, resets after rendering is done)?

Can you explain this pattern? First, it’s hard to guarantee any resetting. You would have to try/finally your entire component. Second, if you only need a ref value temporarily during render, you could have used a regular variable.

packages/react-reconciler/src/__tests__/useRef-test.internal.js
```diff
@@ -0,0 +1,254 @@
...
+
+        return null;
+      }
+
+      act(() => {
+        ReactNoop.render(<Example />);
+      });
+    });
+
+    it('should not warn about reads witin effect', () => {
```
Andarist, line 219: typo
```suggestion
    it('should not warn about reads within effect', () => {
```

billyjanitsch:
Am I right to think that this warning would be triggered by the [recommended implementation of `usePrevious`](https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state)?

gaearon:
Looks like it.

Can you show some code snippets of how you use it?

packages/react-reconciler/src/ReactFiberHooks.js
bvaughn: > Should it warn on writes within render but not within effect if `current !== initialValue` and `value !== initialValue`?

I assume you mean reads, but...not. It *is* okay to read from the ref if it's changed from an initial `null` value (for the lazily initialization pattern) but *only if it's changed once*. If the initial value isn't `null` though, reading after any change could be problematic.

> Would this be valid if `current === undefined` (not `null`)? I've seen `useRef()` without args, even though `mountRef` requires an argument, it can be of void type.

Adding a case for `undefined` is probably reasonable.

sompylasar: No, I meant **writes**.

billyjanitsch:
@gaearon Sorry, just now finding time to write this up.

> Can you show some code snippets of how you use it?

Tbh, I most often see it used for derived state, which I know is better served by [this pattern](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops) (if at all) since it avoids a multi-pass render. However, the ability to set state during render is only ever mentioned once in a "Hooks FAQ" answer and nowhere in the API reference. `usePrevious` is documented and has accordingly found its way into various popular hook technical [writing](https://usehooks.com/usePrevious/) and [hook](https://github.com/streamich/react-use/blob/e30d9cfa493e910f59dfcc2da2bc7cfcf8a0075f/src/usePrevious.ts) [libraries](https://github.com/kitze/react-hanger/blob/dfc0c1865196ba04e2333227d06b3e20fb794750/src/usePrevious.ts), so I've found that, in practice, more developers are aware of it than of setting state during render and it often gets used for the same purpose. I know you'd want to steer people away from this but consider that the warning in this PR is not actionable for people using this for derived state who don't know that state can be set in render.

(Btw, I know that "React changes too often" chatter can be very frustrating given the team's focus on backwards compatibility, but I'm sure you can also understand that it could be discouraging to use a pattern sanctioned by the docs -- so much so that "it’s [possible](https://reactjs.org/docs/hooks-faq.html#how-to-get-the-previous-props-or-state) that in the future React will provide a usePrevious Hook out of the box since it’s a relatively common use case" -- only to get an error about it a few months later. I don't mean to be discouraging and I don't feel that way myself, but I hope this can give you a bit more empathy or patience towards folks who do.)

---

That said, I think I've seen a few legit uses of `usePrevious` in product code. Maybe I haven't sufficiently considered alternatives. The general case is wanting to perform some side effect in response to a changed value where the effect also depends on the previous value. 

#### Logging

Sometimes we want to log when a value has changed, due to a user action or external event (we don't care by which mechanism). Often we want to log the old value as well as the new one, because we're interested in the transition. We generally only care to log the change if it actually got painted to the screen. For instance, imagine a UI where the width of a panel can be resized by the user.

```js
const {width} = props
const prevWidth = usePrevious(width)

useEffect(() => {
  if (width !== prevWidth) {
    log('resize', {width, prevWidth})
  }
}, [width, prevWidth])
```

This could be rewritten to manually read/write to the ref in the effect, but how would you write something reusable that encapsulates what `usePrevious` is doing here?

#### Warning

A similar example is adding a dev-only warning when a prop that should be static (e.g. the initialization value of an uncontrolled component, like `defaultValue` for `input`) changes to indicate that the change will be ignored. The warning should include the old and new value.

https://github.com/facebook/react/pull/18547 is a nice change that lets us do it in render but only if we're using the native `console`.

#### Debugging

While debugging, sometimes we care about a render in which some value has changed from the previous commit. For example, we want to pause execution if a value has changed so we can poke around the scope, or we just want to be informed if a value changed, because we didn't expect that. Maybe we even want to know if a value changed twice across two commits. `usePrevious` serves all of these cases nicely without much code and I haven't been able to imagine a similarly simple alternative.

bvaughn:
Thanks for the detailed response, @billyjanitsch :bow:

As we developer concurrent mode APIs, and see them used more broadly at Facebook, we learn about patterns that can cause problems that we didn't know about initially. It's understandable that it can be frustrating when our guidance changes. For what it's worth, we do try to share our latest way of thinking about these APIs with the larger open source community as our understanding evolves. We'll also be working on a pretty substantial overhaul of the documentation over the next couple of months too, which will hopefully help.

At this point, I don't think we'll be able to move forward with this PR. There are probably too many pre-existing cases that would cause it to fire. We may end up investigating a lint rule instead, since that could be configured/disabled.

---

#### Logging

> This could be rewritten to manually read/write to the ref in the effect, but how would you write something reusable that encapsulates what usePrevious is doing here?

I think you could do something like that with a custom hook too, unless I'm misunderstanding.
```js
function useChangeCallback(value, onChange) {
  const prevValueRef = useRef(value);
  useEffect(() => {
    const prevValue = prevValueRef.current;
    if (prevValue !== value) {
      prevValueRef.current = value;
      onChange(value, prevValue);
    }
  });
}
```

#### Warning

> A similar example is adding a dev-only warning when a prop that should be static (e.g. the initialization value of an uncontrolled component, like defaultValue for input) changes to indicate that the change will be ignored.

That's possible to do even with this warning.
```js
function Example({ propThatShouldNotChange }) {
  const stableRef = useRef(propThatShouldNotChange);
  if (stableRef.current !== propThatShouldNotChange) {
    // Warn
  }
}
```
Because you never *write* to the ref (other than the initial value) reading is fine.

#### Debugging

I don't have a solution for this use case that wouldn't violate the warning this PR explores.

TrySound:
We actively use constant hook to get the first rendered value. It's also in docs
https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
```
export const useConstant = <T>(fn: () => T): T => {
  const ref = React.useRef<null | {| value: T |}>(null);
  if (ref.current == null) {
    // we instantiate { value } to not conflict with returned null
    ref.current = { value: fn() };
  }
  return ref.current.value;
};
```

bvaughn:
@TrySound This PR accounts for that pattern too. (The first time a value is set from `null` to something else doesn't count against the "unsafe to read later" rule.)

dai-shi:
> At this point, I don't think we'll be able to move forward with this PR. There are probably too many pre-existing cases that would cause it to fire. We may end up investigating a lint rule instead, since that could be configured/disabled.

It would be nice for us to intentionally test our apps and libs. What about opting in with StrictMode? Is it feasible? Could it be easier (and more correct) than investigating a new eslint rule?

```jsx
<StrictMode warnReadingMutableRefValue>
```

kof:
> There are probably too many pre-existing cases that would cause it to fire. We may end up investigating a lint rule instead, since that could be configured/disabled.

@bvaughn Sad we can't move forward with this because of that. Your comment made me think, why don't we add the ability to configure runtime warnings?

bvaughn:
A warning suppression system has been discussed in the past on at least one GH issue (though I don't remember the link). If I recall, there was no solid outcome from the discussion.

bvaughn:
I've updated this PR to warn about both reads and writes during render.

Reading or writing a ref value during render is only safe if you are implementing [the lazy initialization pattern](https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily).

Other types of reading are unsafe as the ref is a [mutable source](https://github.com/reactjs/rfcs/blob/master/text/0147-use-mutable-source.md).

Other types of writing are unsafe as they are effectively [side effects](https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects).

If this idea seems worth pursuing, I can wrap the warning in a feature flag to enable us to roll it out gradually.

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1200,26 +1200,39 @@ function pushEffect(tag, create, destroy, deps) {
 function mountRef<T>(initialValue: T): {|current: T|} {
   const hook = mountWorkInProgressHook();
   if (__DEV__) {
-    const fiber = currentlyRenderingFiber;
+    // Support lazy initialization pattern shown in docs.
+    let hasBeenInitialized = initialValue !== null;
```
gaearon, line 1204: What about `undefined`? Many people who don't use type systems just do `useRef()`.

bvaughn: Thought about that. Was concerned about this case though:
```js
const ref = useRef(someVarWithPossiblyUndefinedValueThatMightChange);
```
I think it would be okay if we treated `initialValue === null || arguments.length === 0` though.

gaearon: I guess I'm curious why `someVarWithPossiblyUndefinedValueThatMightChange` wouldn't be a problem for `null`.

bvaughn: It could be, but I couldn't think of a way to differentiate/detect the lazy init pattern that wouldn't false positive :)

billyjanitsch:
Noticed the activity here again. @bvaughn @gaearon is the plan to roll this out for the OSS build? I had noted that this would warn for the documented `usePrevious` implementation (https://github.com/facebook/react/pull/18545#issuecomment-611601144) but honestly I'd prefer this warning over that hook since it catches so many other problematic issues that are quite difficult to automate catching otherwise.

I'd just like to ask for clarity: should the `usePrevious` recommendation in the docs accordingly be considered "deprecated" / should I start nudging codebases away from it?

markerikson:
Similar "will this warn / is this bad?" question:

In React-Redux's `useSelector`, we specifically read from refs while rendering:

https://github.com/reduxjs/react-redux/blob/e9094e7800e7b464d1221d26714454291f8a9730/src/hooks/useSelector.js#L9-L46

Looking at that code, it looks like half of it is a variant of "lazy initialization" (actually seeing if we need to run the selector function right now, while rendering, if we don't have a cached result or if the selector function itself changed), as well as propagating selector errors by throwing them while rendering.

Based on Brian's description, I think the code we have there would cause this warning to trigger, but I think the logic there is also justifiable.

bvaughn:
@billyjanitsch We're still (or once again) discussing this warning as a team. I think it's a good idea, but it will likely cause a lot of churn. Not sure what the next steps will be yet. Our `usePrevious` example *would* trigger the read warning, although it would be a false positive (since that hook only updates the ref in an effect, which is safe). Maybe that indicates a need for me to further refine the warning heuristics here to ignore reads after a mutation inside of an effect callback. That's something for me to think about.

@markerikson It's the end of the day here so I don't feel up to digging through unfamiliar source and thinking about it deeply– but I was pondering earlier though if I had made the read warning overly strict. For instance, maybe it's okay to read from the ref after lazily initializing it *so long as the ref is only mutated once* (for the lazy initialization). I think that would be okay.

billyjanitsch:
>We're still (or once again) discussing this warning as a team.

@bvaughn Thanks for the transparency. That makes sense, and I don't mean to rush for an answer. I'd appreciate if you could post an update here whenever you do come to a conclusion. I solemnly swear not to publish any FUD-laden Medium posts. 🙂

bvaughn:
No worries. This PR will certainly be updated as our thinking evolves.

gaearon:
@bvaughn Curious if you compared this solution to what Seb proposed about the `.stack` trick. I haven't thought deeply about either of them but I think it would be interesting to compare what each can (or cannot) do and how they compare on false positives.

His suggestion was:

>Add a getter for `ref.current`. If it's written before read, warn, because that meant you unconditionally wrote to it. The first time it's read, don't warn but take a `new Error().stack` and save the first stack frame on the hook. If it's read in a subsequent render, ensure that it's the same stack frame, since otherwise it's not a simple lazy init pattern so warn. If it's ever written in a subsequent render, warn. Reset the bookkeeping to original state after "hide".

bvaughn:
@gaearon Where did Seb suggest that? Sounds like it might be a good idea, but I don't remember seeing that suggestion.

bvaughn:
Just pushed an update that adds some polish:
* Track the stack during lazy init to avoid warning on update.
* Only warn once per component+hook pair.

bvaughn:
Just pushed an update that adds some polish:
* Track the stack during lazy init to avoid warning on update.
* Only warn once per component+hook pair.
* More tests added.

gaearon:
>Where did Seb suggest that? 

Sorry, that was a chat thread between us. He recently referred to "using the stack" and I misunderstood your reply to it in the meeting as having already considered this, so I assumed you already saw the same idea. Oops :D

gaearon (commented):
Let's put it behind a flag?

bvaughn:
> that was a chat thread between us

My memory is like Swiss cheese sometimes 😅

bvaughn:
> Let's put it behind a flag?

Sure. I was being lazy and holding off until there was consensus around trying the idea. Just pushed a commit that adds a flag though.

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1197,14 +1198,83 @@ function pushEffect(tag, create, destroy, deps) {
...
+    .join('\n');
+}
+
 function mountRef<T>(initialValue: T): {|current: T|} {
   const hook = mountWorkInProgressHook();
-  const ref = {current: initialValue};
-  if (__DEV__) {
-    Object.seal(ref);
+  if (enableUseRefMutationWarning) {
+    if (__DEV__) {
```
bvaughn, line 1211: This nesting is to satisfy our no-logging-in-production lint rule.

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1197,14 +1198,83 @@ function pushEffect(tag, create, destroy, deps) {
   return effect;
 }
 
+function getCallerStackFrame(): string {
+  return new Error().stack
+    .split('\n')
+    .slice(3, 4)
+    .join('\n');
```
Aprillion, line 1205: different browsers have different stacks, e.g. Firefox and Safari do not include the error message in the stack, so the line with `Component@file.js:<line>:<column> is at index 2 in those browsers, not index 3 like in Chrome, Node and IE

=> maybe something like https://github.com/stacktracejs/error-stack-parser/blob/master/dist/error-stack-parser.js ?

bvaughn: Thanks for reminding me of that! 😅

kentcdodds:
Hey friends, would this warn for this type of situation? https://codesandbox.io/s/agitated-rubin-x2c1l?file=/src/index.js:380-595

If so, how would one create a DOM node for a portal? Would I have to manage this in a `useEffect` and not render until I get a chance to create the DOM node in the effect callback? Seems kinda weird.

markerikson:
@kentcdodds: based on the discussions here and on Twitter in the last couple days, that usage looks valid to me:

- It's a lazy init
- The value being assigned is GC-able by itself (ie, no `ref.current.goCleanYourselfUp()` methods that need to be called)

I haven't double-checked against the current logic in the PR, but that seems okay.

bvaughn:
> would this warn for this type of situation? https://codesandbox.io/s/agitated-rubin-x2c1l?file=/src/index.js:380-595

Yes, in its current form.

> If so, how would one create a DOM node for a portal?

You could use state, [like so](https://codesandbox.io/s/intelligent-dew-1l3p8?file=/src/index.js):
```js
const [container] = React.useState(() => document.createElement("div"));

React.useEffect(() => {
  modalRoot.append(container);
  return () => container.remove();
}, []);

```

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.js
```diff
@@ -3202,91 +3202,6 @@ describe('ReactHooksWithNoopRenderer', () => {
     });
   });
 
-  describe('useRef', () => {
```
acdlite, line 3205: Why were these deleted?

acdlite: Can you update them to not read instead? Or to assert on the warning?

bvaughn: All ref tests were moved into `useRef-test` since there are a lot of them. That big hooks test file kind of sucks.

packages/react-reconciler/src/ReactFiberHooks.old.js
```diff
@@ -1175,14 +1176,92 @@ function pushEffect(tag, create, destroy, deps) {
...
 function mountRef<T>(initialValue: T): {|current: T|} {
   const hook = mountWorkInProgressHook();
-  const ref = {current: initialValue};
-  if (__DEV__) {
-    Object.seal(ref);
+  if (enableUseRefMutationWarning) {
+    if (__DEV__) {
+      // Support lazy initialization pattern shown in docs.
+      // We need to store the caller stack frame so that we don't warn on subsequent renders.
+      let hasBeenInitialized = initialValue != null;
```
acdlite, line 1201: So this only supports lazy initializing object types? i.e. you couldn't do lazy init with `useRef(-1)`

packages/react-reconciler/src/ReactFiberHooks.old.js
```diff
@@ -1175,14 +1176,92 @@ function pushEffect(tag, create, destroy, deps) {
...
+      let didWarnAboutRead = false;
+      let didWarnAboutWrite = false;
+
+      let current = initialValue;
+      const ref = {
+        get current() {
+          if (!hasBeenInitialized) {
+            didCheckForLazyInit = true;
+            lazyInitGetterStack = getCallerStackFrame();
+          } else if (currentlyRenderingFiber !== null && !didWarnAboutRead) {
```
acdlite, line 1215: What if it's read inside a class component? Or written to. Should it warn?

packages/react-reconciler/src/__tests__/useRef-test.internal.js
```diff
@@ -0,0 +1,361 @@
...
+  let useState;
+
+  beforeEach(() => {
+    React = require('react');
+    ReactNoop = require('react-noop-renderer');
+    Scheduler = require('scheduler');
+
+    const ReactFeatureFlags = require('shared/ReactFeatureFlags');
+    ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
+    ReactFeatureFlags.enableUseRefMutationWarning = true;
```
acdlite, line 33: Why this instead of the gate pragma?

packages/shared/forks/ReactFeatureFlags.www-dynamic.js
```diff
@@ -48,3 +48,4 @@ export const enableTrustedTypesIntegration = false;
 export const disableSchedulerTimeoutBasedOnReactExpirationTime = false;
 
 export const enableDoubleInvokingEffects = false;
+export const enableUseRefMutationWarning = false;
```
acdlite, line 51: Maybe "access" warning since it covers both reads and writes

acdlite (approved):

packages/react-reconciler/src/ReactFiberHooks.old.js
bvaughn: Only supports `null` and `undefined`, yes.

packages/react-reconciler/src/ReactFiberHooks.old.js
bvaughn: Undefined I guess. I could expand the warning to include class components as well, if you think that's worth doing.

bvaughn:
Hey everyone 👋 I am going to merge this PR and roll it out behind a feature flag within Facebook to assess its impact, e.g. how many warnings does it trigger, how many patterns (like `usePrevious`) does it flush out, do they all have safe alternate patterns, etc.

If you have code that unconditionally mutates a ref during render, you might want to go ahead and change it. It can break things (like #19114).

If you have code that reads a ref during render (e.g. the `usePrevious` pattern described above) please don't feel a need to make any changes to that code for now. We'll provide guidance once we've done more research.

Andarist:
> > would this warn for this type of situation? https://codesandbox.io/s/agitated-rubin-x2c1l?file=/src/index.js:380-595

> Yes, in its current form.

Isn't this a classic lazy init pattern that is being tested [here](https://github.com/facebook/react/pull/18545/files#diff-2202ffbecc1ea5bb40922101ff453d687834222c0013aa9a71191857c77b6937R175)? 🤔 

> You could also use memo, like so:

Isn't this unsafe as `useMemo` doesn't have semantic guarantees and can be just freely reinitialized when React decides to do so (which includes recomputation during Fast Refresh)?

bvaughn:
> Isn't this a classic lazy init pattern that is being tested [here](https://github.com/facebook/react/pull/18545/files#diff-2202ffbecc1ea5bb40922101ff453d687834222c0013aa9a71191857c77b6937R175)? 🤔 

The reason that sandbox would warn with the semantics on this PR is not because of the lazy init pattern, but because of the read further down on line 60 (passing `ref.current` to `createPortal`).

> Isn't this unsafe for this use case as `useMemo` doesn't have semantic guarantees and can be just freely reinitialized when React decides to do so (which includes recomputation during Fast Refresh)? 

My code example above had a silly mistake in the deps array. Should have been this:
```js
const container = useMemo(() => document.createElement("div"), []);

useLayoutEffect(() => {
  modalRoot.append(container);
  return () => container.remove();
}, [container]);
```

If React *did* "forget" the memoized element between renders, the effect would clean up the old one and reparent the new one.

josepot:
> The reason that sandbox would warn with the semantics on this PR is not because of the lazy init pattern, but because of the read further down on line 60 (passing ref.current to createPortal).

But since that ref would always yield the same value, that would be a "false positive", correct? I mean, as long as the render function always behaves the same for the same props, state and context, then it's ok to read a value from a ref, right?

bvaughn:
Yes, it would be a false positive in this case.

The danger with refs is that they can be passed around and mutated outside of React's awareness (unlike state/reducer).

We could change the warning semantics to only warn on read if the ref was mutated *after* being initialized (not including lazy-init pattern). The danger there is that maybe there's a code path that triggers a second mutation that just doesn't run often. Maybe you'd rarely or never see it in dev. So you have a false sense of confidence that your component is safe, when really there's an unsafe read that might cause a bug in production if this infrequent code path gets triggered.

josepot:
> We could change the warning semantics to only warn on read if the ref was mutated after being initialized (not including lazy-init pattern). The danger there is that maybe there's a code path that triggers a second mutation that just doesn't run often. Maybe you'd rarely or never see it in dev. So you have a false sense of confidence that your component is safe, when really there's an unsafe read that might cause a bug in production if this infrequent code path gets triggered.

Sorry, my intention was not to criticize the work done on this PR. I agree that it's better to err on the side of caution.

I made that comment because I wasn't sure if you were proposing those alternatives because you deemed the original code to be unsafe or because you wanted to highlight alternative implementations that accomplished the same thing while avoiding warnings.

Thanks a lot for clarifying that.

PR #18547, opened 2020-04-08
Disable console.logs in the second render pass of DEV mode double render

sebmarkbage:
This disables console.log by temporarily patching the global console object during the second render pass when we double render in strict mode in DEV.

I could go either way on this. It is confusing that it double logs warnings, React, legacy react packages or third-party ones. However, if you do have a problem due to double rendering then you kind of want logs to debug it.

I also uses this override to detect this in our Scheduler.yieldValue helper so that it automatically doesn't double log which makes it easier to write tests that test render phases. This can be done without any internals being exposed so it works to test our build outputs.

The main motivation of this is actually a different PR where I want to double render to expose stack traces. In this case it's extra strange because it's extracting the error stack from the first one that causes new warnings that are not necessarily the same. So I think we need something there. We don't necessarily have to do the same for these ones.

This is ofc not exhaustive since there are so many other side-effects than just logs and you could have other APIs that show logs in UI etc.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
@@ -10,7 +10,6 @@
 'use strict';
 
 let React;
-let ReactFeatureFlags = require('shared/ReactFeatureFlags');
 
 let ReactDOM;
 let Scheduler;
@@ -149,8 +148,7 @@ describe('ReactDOMFiberAsync', () => {
   describe('concurrent mode', () => {
     beforeEach(() => {
       jest.resetModules();
-      ReactFeatureFlags = require('shared/ReactFeatureFlags');
-      ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
+
       ReactDOM = require('react-dom');
       Scheduler = require('scheduler');
     });
--- a/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
@@ -639,9 +639,7 @@ describe('ReactDOMFiberAsync', () => {
       expect(container.textContent).toEqual('');
 
       // Everything should render immediately in the next event
-      expect(Scheduler).toFlushExpired(
-        __DEV__ ? ['A', 'A', 'B', 'B', 'C', 'C'] : ['A', 'B', 'C'],
-      );
+      expect(Scheduler).toFlushExpired(['A', 'B', 'C']);
       expect(container.textContent).toEqual('ABC');
     });
   });
--- a/packages/react-reconciler/src/ReactFiberBeginWork.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.js
@@ -181,6 +181,8 @@ import {
   getWorkInProgressRoot,
 } from './ReactFiberWorkLoop';
 
+import {disableLogs, reenableLogs} from 'shared/ConsolePatchingDev';
+
 const ReactCurrentOwner = ReactSharedInternals.ReactCurrentOwner;
 
 let didReceiveUpdate: boolean = false;
@@ -320,14 +322,19 @@ function updateForwardRef(
       debugRenderPhaseSideEffectsForStrictMode &&
       workInProgress.mode & StrictMode
     ) {
-      nextChildren = renderWithHooks(
-        current,
-        workInProgress,
-        render,
-        nextProps,
-        ref,
-        renderExpirationTime,
-      );
+      disableLogs();
+      try {
+        nextChildren = renderWithHooks(
+          current,
+          workInProgress,
+          render,
+          nextProps,
+          ref,
+          renderExpirationTime,
+        );
+      } finally {
+        reenableLogs();
+      }
     }
     setIsRendering(false);
   } else {
@@ -658,14 +665,19 @@ function updateFunctionComponent(
       debugRenderPhaseSideEffectsForStrictMode &&
       workInProgress.mode & StrictMode
     ) {
-      nextChildren = renderWithHooks(
-        current,
-        workInProgress,
-        Component,
-        nextProps,
-        context,
-        renderExpirationTime,
-      );
+      disableLogs();
+      try {
+        nextChildren = renderWithHooks(
+          current,
+          workInProgress,
+          Component,
+          nextProps,
+          context,
+          renderExpirationTime,
+        );
+      } finally {
+        reenableLogs();
+      }
     }
     setIsRendering(false);
   } else {
@@ -731,14 +743,19 @@ function updateBlock<Props, Data>(
       debugRenderPhaseSideEffectsForStrictMode &&
       workInProgress.mode & StrictMode
     ) {
-      nextChildren = renderWithHooks(
-        current,
-        workInProgress,
-        render,
-        nextProps,
-        data,
-        renderExpirationTime,
-      );
+      disableLogs();
+      try {
+        nextChildren = renderWithHooks(
+          current,
+          workInProgress,
+          render,
+          nextProps,
+          data,
+          renderExpirationTime,
+        );
+      } finally {
+        reenableLogs();
+      }
     }
     setIsRendering(false);
   } else {
@@ -923,7 +940,12 @@ function finishClassComponent(
         debugRenderPhaseSideEffectsForStrictMode &&
         workInProgress.mode & StrictMode
       ) {
-        instance.render();
+        disableLogs();
+        try {
+          instance.render();
+        } finally {
+          reenableLogs();
+        }
       }
       setIsRendering(false);
     } else {
@@ -1485,14 +1507,19 @@ function mountIndeterminateComponent(
         debugRenderPhaseSideEffectsForStrictMode &&
         workInProgress.mode & StrictMode
       ) {
-        value = renderWithHooks(
-          null,
-          workInProgress,
-          Component,
-          props,
-          context,
-          renderExpirationTime,
-        );
+        disableLogs();
+        try {
+          value = renderWithHooks(
+            null,
+            workInProgress,
+            Component,
+            props,
+            context,
+            renderExpirationTime,
+          );
+        } finally {
+          reenableLogs();
+        }
       }
     }
     reconcileChildren(null, workInProgress, value, renderExpirationTime);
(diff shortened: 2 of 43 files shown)


gaearon:
I was just debugging some internals yesterday and wished we had this.

acdlite:
Could we do the same for `replayFailedUnitOfWorkWithInvokeGuardedCallback`?

packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.internal.js
```diff
@@ -149,8 +148,7 @@ describe('ReactDOMFiberAsync', () => {
   describe('concurrent mode', () => {
     beforeEach(() => {
       jest.resetModules();
-      ReactFeatureFlags = require('shared/ReactFeatureFlags');
-      ReactFeatureFlags.debugRenderPhaseSideEffectsForStrictMode = false;
```
acdlite, line 153: Can remove  `internal` suffix now

sebmarkbage: Yea I'll do a follow up to rename all these if I land this. I think the remaining ones are just using the `let ops = []` pattern and we can port those.

sebmarkbage:
> Could we do the same for replayFailedUnitOfWorkWithInvokeGuardedCallback?

I think we probably could but that one was kind of so that you could break on the exception. The problem then is your logs are disabled while on your break point. That seems highly confusing.

packages/react-reconciler/src/ReactFiberBeginWork.js
```diff
@@ -1485,14 +1507,19 @@ function mountIndeterminateComponent(
...
-          null,
-          workInProgress,
-          Component,
-          props,
-          context,
-          renderExpirationTime,
-        );
+        disableLogs();
+        try {
+          value = renderWithHooks(
```
acdlite, line 1512: I could foresee some confusion if the logs don't correspond to the value that is returned. It usually doesn't matter because it's supposed to be pure, but in the case where it's not, you might use `console` to debug it. So maybe they should be disabled for the first pass instead of the second.

acdlite:
> The problem then is your logs are disabled while on your break point

I was thinking you would disable the logs during the first pass and enable them for the second. In the case of an error, the second pass would use `invokeGuardedCallback`.

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: So I was actually surprised that we used the second value. I was thinking we should ignore the second value like we do for class constructors.

sebmarkbage: In my other PR's replaying I have to ignore it because I force an error inside the function. Similarly the react-debug-tools for Hooks debugging it's just replayed stand-alone.

Doesn't necessarily mean we have to do the same here. I guess that's how I look at these.

acdlite: It's intentional because the children are bound to the work-in-progress fiber via the `queue` object via the `dispatch` method, and `renderWithHooks` mutates the fiber.

It was a very confusing bug when we found it :D

sebmarkbage:
> I was thinking you would disable the logs during the first pass and enable them for the second. In the case of an error, the second pass would use invokeGuardedCallback.

That would mean that we would have to always replay the second pass. At least as long as there's any log. Since most of the time when nothing throws or if it's just throwing a Promise, there wouldn't be a second pass.

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: This one: https://github.com/facebook/react/pull/14698

acdlite:
> That would mean that we would have to always replay the second pass.

That seems fine? Isn't the only new case promises?

packages/react-reconciler/src/ReactFiberBeginWork.js
sebmarkbage: I guess that's why I had to update that test to get the right dispatcher.

sebmarkbage: I could go either way on if the second render should process the update queue and even leave any side-effects. :)

sebmarkbage: Maybe the second render should have a different dispatcher that warns if you call setState again since you shouldn't call setState again for the same input.

sebmarkbage:
> That seems fine? Isn't the only new case promises?

No if we disable the logs we'd have to detect that we had silenced a log and replay. So any little log would severely slow down debugging.

We can't replay them later because it would give them the wrong stacks frames.

acdlite:
I'm confused because we already replay the methods in dev. I don't get what's getting slower.

sebmarkbage:
Hm. I was under the impression that we replay the whole render but I guess it's just the begin phase. Is that even safe from a resuming perspective? Seems like we'd have a lot of latent bugs there. Probably SuspenseList.

sebmarkbage:
It seems like these two should just use the same mechanism. If the first pass errors, then we do the double render pass in invokeGuardedCallback

sebmarkbage:
We can do that in a follow up if we decide to land this. For now we need to decide on this to unblock the production stacks PR.

packages/react-reconciler/src/ReactFiberBeginWork.js
acdlite: > I could go either way on if the second render should process the update queue and even leave any side-effects  :)

It can work but it needs to include internal mutations/side-effects too. We'd have to make it so that `renderWithHooks` does not mutate any of the `fiber` fields. Or stash the result of the first pass (`memoizedState`, `expirationTime`, `effectTag`) before doing the second pass and then restore them.

acdlite: > Maybe the second render should have a different dispatcher that warns if you call setState again since you shouldn't call setState again for the same input.

Oh I didn't see that. Yeah I think that would work. Just call the function again with a special dispatcher that doesn't mutate anything.

sebmarkbage:
One fatal flaw of silencing the first pass is warning deduping. Almost none of our warnings would show up as it currently stands and also true if others use the same technique.

sebmarkbage:
This actually revealed that we're calling the fake setState updaters for classes first and silencing the warnings inside of them as a result.

packages/react-reconciler/src/ReactFiberBeginWork.js
bvaughn: > Maybe the second render should have a different dispatcher that warns if you call setState again since you shouldn't call setState again for the same input.

I might be misunderstanding this comment, but if the second call has state updates already applied, then we might miss out on triggering the observable side effects in the first place (e.g. like something that happens only when props/state meet a certain criteria).

orestis:
I know this is in the experimental branch and I'm not supposed to use it, but I found this *super* confusing when I was trying to figure out what was going on in some lifecycles :)

gaearon:
Yeah, I feel that. I could go either way on this.

If you don't know what's happening, this makes it doubly confusing.

But if you know what's happening, then *not* having this is annoying because you can't use console.log to debug things in a sequence. It's just too noisy.

orestis:
My underlying issue is that I had some other render side-effects that I thought were idempotent (useRef) but turns out not -- and the first instinct to pull out console.log made me super confused.

As always in software, there's the possibility of making this configurable, perhaps opt-out with a small start-up warning message letting people know it's a thing? I mean, the double-render behaviour is mentioned in the docs but I'm pretty sure many people miss it.

bvaughn:
> perhaps opt-out with a small start-up warning message letting people know it's a thing?

I think console messages like this are kind of traps. They seem really useful- and often are, initially, but then they can get really annoying because after you've seen them and know what's going on- the console message is still always there cluttering up your log.

Even if we could only show them exactly once (say using a mechanism like DevTools and `localStorage`) there's a chance you would *not* see them that one time, because people aren't always staring at their console- only when they're actively using it to debug.

gaearon:
The problem with console messages on startup is that people who need them the most won't read them, and they're noise for everybody else.

jamesknelson:
I feel like *some* kind of warning would be really nice here.

I've been pulling my hair out the last few days trying to figure out why something seemed to be running twice, while the console logs seemed to show that it was only running once. It got especially confusing as logs in async code triggered by the render were still appearing twice, making it seem like the issue was async related. I only managed to figure out what was going on by grepping for `console` in react-dom.development.js.

I'm not sure exactly what the best way to warn developers is, but I'd wager most developers wouldn't expect a global like the console to change without *some* kind of warning.

eps1lon:
> One suggestion would be for an export from react-dom that lets you enable/disable this behavior.

I think the plan is to add a devtools option for it. See https://twitter.com/dan_abramov/status/1292959451415302144.

Opening a new feature request for devtools would help with some context why the current behavior is problematic and why the previous behavior was problematic as well. Otherwise comments in an already merged PR might get lost.

bvaughn:
There's some prior art here, as DevTools already overrides the console to add component stacks (if the option is enabled).

Screen Shot 2020-08-20 at 9 05 15 AM

Might be worth putting together a detailed proposal of how this should work (as a new GitHub issue).

joepuzzo:
IMO this is kind of crazy. In order to teach others of this behavior I would previously create a simple example in code sandbox and log in a use state.

```js
useState(() => console.log('Init'))
```

Most devs ( myself included ) "learn by logging". Now they may be scratching their heads for hours until they stumble upon an article or post that describes their code is executing two times. This needs to be more transparent.

matthewma7:
Didn't aware the strict mode is double rendering, this change cost me last night chasing down what happened. 
Instead of muting the console.log, could we monkey patch the log and append something like "Strict Mode double render" to the beginning of the console.log?

infinnie:
> My underlying issue is that I had some other render side-effects that I thought were idempotent (useRef) but turns out not -- and the first instinct to pull out console.log made me super confused.
> 
> As always in software, there's the possibility of making this configurable, perhaps opt-out with a small start-up warning message letting people know it's a thing? I mean, the double-render behaviour is mentioned in the docs but I'm pretty sure many people miss it.

Yes the double render behavior is mentioned. But the console suppression isn’t.

gaearon:
To follow up on this, we've changed the behavior in 18 to be more intuitive.
https://github.com/facebook/react/issues/21783#issuecomment-1083412766

PR #18548, opened 2020-04-09
Copy React Reconciler modules to new fork

acdlite:
Builds on top of #18285. See that PR for context and motivation.

I suggest reviewing commit-by-commit.

-  The first commit renames the existing (`.old`) modules. I'll land this to master in its own commit to preserve the history/blame.
-  The second commit refactors some of the internal types so they can be shared across implementations.
-  The third commit adds the forked (`.new`) modules. Forked modules are given a suffix of `.old` or `new`. It does change any of the implementation. They are the same except for the imports.  The changes in this commit are autogenerated. (Refer to the commit message for exact steps.)
- The fourth commit points Rollup, Jest, and Flow at the new implementation's entry points.


acdlite:
I'm going to land this while everyone is sleeping 😆

Normally I wouldn't do this but I'm making an exception since this affects such a large surface area and landing this in the middle of the day would be pretty churny / annoying to rebase.

More rationalization: none of the implementation has changed, only infra stuff and shuffling of types.

PR #18555, opened 2020-04-09
Support `additionalProperties` in `react-hooks/exhaustive-deps`

osdnk:
## Summary

After adding `enableDangerousAutofixThisMayCauseInfiniteLoops` I wanted to use it and observed that `additionalProperties` in schema of `eslint` are not allowed.

## Test Plan
Add `react-hooks/exhaustive-deps: ["warn", { enableDangerousAutofixThisMayCauseInfiniteLoops : true }]`
in `eslint.rc`

### DIFF
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -23,7 +23,7 @@ export default {
     schema: [
       {
         type: 'object',
-        additionalProperties: false,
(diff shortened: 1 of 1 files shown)


gaearon:
This doesn't make sense to me. `enableDangerousAutofixThisMayCauseInfiniteLoops` *is* defined in properties, so why do we need to enable `additionalProperties`?

I think you're having troubles just because the version with `enableDangerousAutofixThisMayCauseInfiniteLoops` hasn't been released yet.

PR #18559, opened 2020-04-09
Exclude forwardRef and memo from stack frames

sebmarkbage:
We can't patch the row with the new trick. So they won't be called `Memo(Name)` in the stack anymore.

We could give these their own "built-in" stack frame since they're conceptually HoCs.  I.e.
```
  at Name
  at Memo
```

However, from a debugging perspective this is not very useful meta data and quite noisy. So I'm just going to exclude them.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactMemo-test.js
+++ b/packages/react-reconciler/src/__tests__/ReactMemo-test.js
@@ -392,12 +392,20 @@ describe('memo', () => {
         Outer.defaultProps = {outer: 0};
 
         // No warning expected because defaultProps satisfy both.
-        ReactNoop.render(<Outer />);
+        ReactNoop.render(
+          <div>
+            <Outer />
+          </div>,
+        );
         expect(Scheduler).toFlushWithoutYielding();
 
         // Mount
         expect(() => {
-          ReactNoop.render(<Outer inner="2" middle="3" outer="4" />);
+          ReactNoop.render(
+            <div>
+              <Outer inner="2" middle="3" outer="4" />
+            </div>,
+          );
           expect(Scheduler).toFlushWithoutYielding();
         }).toErrorDev([
           'Invalid prop `outer` of type `string` supplied to `Inner`, expected `number`.',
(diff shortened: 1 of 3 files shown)


gaearon:
You kept forwardRef?

packages/shared/ReactComponentStackFrame.js
```diff
@@ -121,7 +121,7 @@ export function describeUnknownElementTypeFrameInDEV(
       case REACT_FORWARD_REF_TYPE:
         return describeFunctionComponentFrame(type.render, source, ownerFn);
       case REACT_MEMO_TYPE:
-        return describeFunctionComponentFrame(type.type, source, ownerFn);
```
sebmarkbage, line 124: This was also a bug because the inner one might not be a function.

packages/react-reconciler/src/__tests__/ReactMemo-test.js
```diff
@@ -408,7 +416,9 @@ describe('memo', () => {
         // Update
         expect(() => {
           ReactNoop.render(
-            <Outer inner={false} middle={false} outer={false} />,
+            <div>
+              <Outer inner={false} middle={false} outer={false} />
+            </div>,
```
sebmarkbage, line 421: The warning no longer includes any stack because there's only a Memo. This triggers our internal warning about not expecting a stack on the warning. However, that's not realistic so let's try it with a stack.

sebmarkbage:
I kept SimpleMemoComponent and ForwardRef in the Fiber describing thing but they just describe their function name. It's only one Fiber for both the function and wrapper. I removed Memo because it's not the actual function, the function is represented by its own Fiber as a child so it's already there.

For elements I was already doing this by only naming the function.

packages/shared/ReactComponentStackFrame.js
```diff
@@ -121,7 +121,7 @@ export function describeUnknownElementTypeFrameInDEV(
...
+        return describeUnknownElementTypeFrameInDEV(type.type, source, ownerFn);
```
gaearon, line 124: Comment plz?

PR #18561, opened 2020-04-09
Build Component Stacks from Native Stack Frames

sebmarkbage:
Instead of using displayName/name and source location from jsxDEV, this uses a hack to extract the native stack frame of the function. That way it's zero-cost when an error is not hit and we can feel good about including it in production environments too.

It works by calling each function once with invalid input and intentionally causing it to throw and catching the error. This should be ok since they shouldn't have side-effects.

By doing so we get the stack frame of that function somewhere in the error's stack. We then compare it to a control stack to figure out which one is the next frame after that. This should be the frame of the component. We use that to build a stack.

This stack should be parsable by off-the-shelf cross-browser stack parsing libraries.

The idea is to throw as early as possible. Ideally it's the first line such as the built-in super call of a class or the destructuring of props in the argument position. It could however be slightly later which might be slightly confusing to see the line of the first hook for example. Source code aware tooling could adjust this number to the previous function definition before the offset. If it doesn't throw at all such as if no hooks and no props are used, then we can't figure out the line number.

This doesn't give us the offset of the JSX call so it's slightly worse than the source info transform in that regard. However, it has the benefit of yielding the client-side source position rather than file name source position. This allows linkifying the location on the client so you can jump to source. It does require source maps to do so but it can piggyback on standard error logging strategies. This PR make the flag take over both DEV and prod so the source location from the transform is not used for consistency.

If this ends up being considered worse than the JSX transform due to the lack of JSX callsites, I have another idea to use the second error recovery pass in DEV to collect exact JSX callsites using stack frames but that's way more complicated and code. We wouldn't ship that to prod anyway.

TODO:
- [x] Fix bug when a native class doesn't extend React.Component and can't be called
- [x] Add fixture for easy cross-browser testing
  - [ ] There's an issue with native classes in Safari because Safari adds the "construct" call to the stack.


packages/react/src/ReactDebugCurrentFrame.js
```diff
@@ -7,26 +7,20 @@
  * @flow
  */
```
bvaughn, line 8: The organization of this file is confusing to me.
1. We unconditionally export a setter which sets a variable that is not exported (and has no getter or is otherwise unused *except* for DEV mode which uses it in `getStackAddendum`).
2. We assign a duplicate of that same setter to `ReactDebugCurrentFrame` in DEV mode, but we don't assign a getter. We leave that up to the current renderer, which of course, doesn't even use the local value written by `ReactDebugCurrentFrame.setExtraStackFrame` because it has no way to access it. So instead it uses its own local current fiber value.

Looks like the only thing using the exported `setCurrentlyValidatingElement` function (now `setExtraStackFrame`) is `ReactElementValidator` - and **even it only uses it in DEV**. So why are we using this odd pattern?

I'm not super familiar with some of these files so maybe there's a subtle difference that I'm overlooking.

Suggestion: Update `ReactElementValidator` to use `ReactDebugCurrentFrame.*` too, then get rid of the ES export.

packages/react/src/ReactElementValidator.js
```diff
@@ -127,16 +144,16 @@ function validateExplicitKey(element, parentType) {
     )}.`;
   }
 
-  setCurrentlyValidatingElement(element);
   if (__DEV__) {
+    setCurrentlyValidatingElement(element);
```
bvaughn, line 148: Nice catch. I noticed this while writing my above comment.

packages/shared/ConsolePatchingDev.js
```diff
@@ -21,28 +22,34 @@ function disabledLog() {}
...
-    // $FlowFixMe Flow thinks console is immutable.
-    console.log = prevLog;
-    // $FlowFixMe Flow thinks console is immutable.
-    console.info = prevInfo;
-    // $FlowFixMe Flow thinks console is immutable.
-    console.warn = prevWarn;
-    // $FlowFixMe Flow thinks console is immutable.
-    console.error = prevError;
-    /* eslint-enable react-internal/no-production-logging */
+    disabledDepth--;
```
bvaughn, line 41: Should we throw or warn if we ever go < 0?

packages/shared/__tests__/describeComponentFrame-test.js
```diff
@@ -11,13 +11,19 @@
...
 let ReactDOM;
+const ReactFeatureFlags = require('shared/ReactFeatureFlags');
 
 describe('Component stack trace displaying', () => {
   beforeEach(() => {
     React = require('react');
     ReactDOM = require('react-dom');
   });
 
+  if (ReactFeatureFlags.enableComponentStackLocations) {
```
bvaughn, line 22: This doesn't look necessary (at least not yet?)

packages/shared/ReactComponentStackFrame.js
```diff
@@ -19,6 +21,144 @@ import {
...
+      }
+      for (; s >= 1 && c >= 0; s--, c--) {
+        // Next we find the first one that isn't the same which should be the
+        // frame that called our sample function.
+        if (sampleLines[s] !== controlLines[c]) {
+          // Return the line we found.
+          // V8 adds a "new" prefix for native classes. Let's remove it to make it prettier.
+          const frame = '\n' + sampleLines[s - 1].replace(' at new ', ' at ');
+          if (__DEV__) {
+            if (typeof fn === 'function') {
```
bvaughn, line 136: Why is this `typeof` check necessary?

packages/shared/__tests__/describeComponentFrame-test.js
sebmarkbage: I think I'll just remove this test later. It's not very useful for this model. This test fails with the stack on, partly because the warning gets deduped because it's always the same component and line being called.

packages/shared/ReactComponentStackFrame.js
sebmarkbage: Because it throws in WeakMap if the wrong type sneaks in. That kills the stack which is not good if we're warning because some wrong type. We have tests that cover that.

packages/react/src/ReactDebugCurrentFrame.js
```diff
@@ -7,26 +7,20 @@
...
-    currentlyValidatingElement = element;
+    currentExtraStackFrame = stack;
   }
 }
 
 if (__DEV__) {
-  ReactDebugCurrentFrame.setCurrentlyValidatingElement = function(
-    element: null | ReactElement,
-  ) {
+  ReactDebugCurrentFrame.setExtraStackFrame = function(stack: null | string) {
```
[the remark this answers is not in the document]
sebmarkbage: Since it's DEV it probably doesn't matter but in general that might not be great because one is meant to be called in this same package so it can be inlined. So that would avoid inlining it.

packages/react/src/ReactDebugCurrentFrame.js
sebmarkbage: It's complicated because this is used by `jsx-runtime` too which is a separate file that needs to import this from isomorphic. This needs some further refactoring in a follow up at some point.

For now I'm just removing a cycle that otherwise would break.

packages/shared/ReactComponentStackFrame.js
bvaughn: Gotcha. It looked odd to have `fn: Function` followed by a `typeof` check.

bvaughn:
This generates a funky stack for host components when I trigger an intentional error in the DevTools test shell:
```
Uncaught Error: Objects are not valid as a React child (found: object with keys {}). If you meant to render a collection of children, use an array instead.
    at :8080/label
    at :8080/li
    at ListItem (app.js:63661)
    at :8080/ul
    at List (app.js:63459)
```

The browser things ":8080/li" and ":8080/ul" should be clickable, and opens a new tab to e.g. http://localhost:8080/ul

sebmarkbage:
I'm confused by the edits. What's the problem. The last one seems fine?

bvaughn:
Sorry. The last one is fine. The top one is funky, but the top one is actually the browser's own doing. I was mistaken at first.

PR #18564, opened 2020-04-09
Refactor detachFiber to a later stage

trueadm:
This PR is kind of a follow up to these PRs:

https://github.com/facebook/react/pull/18529
https://github.com/facebook/react/pull/18556

Specifically, in https://github.com/facebook/react/pull/18556, which cleared the `sibling` Fiber field, we encountered a failing internal test. The test related to `onBeforeBlur` and using scopes. Specifically, the problem was that we were trying to traverse a fiber tree at a point where a bunch of the siblings were already "deatached" – as in we already cleared their fields. This wasn't ideal, as `onBeforeBlur` gets called during the same phase as where we detatch fibers so we want to be able to accurately traverse the fiber tree still.

This PR moves `detachFiber` to a later stage, as mentioned in the comment by @bvaughn.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -1110,25 +1110,6 @@ function commitNestedUnmounts(
   }
 }
 
-function detachFiber(fiber: Fiber) {
-  // Cut off the return pointers to disconnect it from the tree. Ideally, we
-  // should clear the child pointer of the parent alternate to let this
-  // get GC:ed but we don't know which for sure which parent is the current
-  // one so we'll settle for GC:ing the subtree of this child. This child
-  // itself will be GC:ed when the parent updates the next time.
-  fiber.return = null;
-  fiber.child = null;
-  fiber.memoizedState = null;
-  fiber.updateQueue = null;
-  fiber.dependencies = null;
-  fiber.alternate = null;
-  fiber.firstEffect = null;
-  fiber.lastEffect = null;
-  fiber.pendingProps = null;
-  fiber.memoizedProps = null;
-  fiber.stateNode = null;
-}
-
 function emptyPortalContainer(current: Fiber) {
   if (!supportsPersistence) {
     return;
@@ -1521,11 +1502,6 @@ function commitDeletion(
   } else {
     // Detach refs and call componentWillUnmount() on the whole subtree.
     commitNestedUnmounts(finishedRoot, current, renderPriorityLevel);
-  }
-  const alternate = current.alternate;
-  detachFiber(current);
-  if (alternate !== null) {
-    detachFiber(alternate);
   }
 }
 
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -1919,6 +1919,15 @@ function commitRootImpl(root, renderPriorityLevel) {
       }
     } while (nextEffect !== null);
 
+    nextEffect = firstEffect;
+    while (nextEffect !== null) {
+      const nextNextEffect = nextEffect.nextEffect;
+      if (nextEffect.effectTag & Deletion) {
+        detachFiberPair(nextEffect);
+      }
+      nextEffect = nextNextEffect;
+    }
+
     nextEffect = null;
 
     // Tell Scheduler to yield at the end of the frame, so the browser has an
@@ -3409,5 +3418,36 @@ export function act(callback: () => Thenable<mixed>): Thenable<void> {
         resolve();
       },
     };
+  }
+}
+
+function detachFiberPair(fiber: Fiber) {
+  const alternate = fiber.alternate;
+  detachFiber(fiber);
+  if (alternate !== null) {
+    detachFiber(alternate);
+  }
+}
+
+function detachFiber(fiber: Fiber) {
+  // Cut off the return pointers to disconnect it from the tree. Ideally, we
+  // should clear the child pointer of the parent alternate to let this
+  // get GC:ed but we don't know which for sure which parent is the current
+  // one so we'll settle for GC:ing the subtree of this child. This child
+  // itself will be GC:ed when the parent updates the next time.
+  fiber.alternate = null;
+  fiber.child = null;
+  fiber.dependencies = null;
+  fiber.firstEffect = null;
+  fiber.lastEffect = null;
+  fiber.memoizedProps = null;
+  fiber.memoizedState = null;
+  fiber.pendingProps = null;
+  fiber.return = null;
+  fiber.sibling = null;
+  fiber.stateNode = null;
+  fiber.updateQueue = null;
+  if (__DEV__) {
+    fiber._debugOwner = null;
   }
 }
--- a/packages/react-interactions/events/src/dom/__tests__/FocusWithin-test.internal.js
+++ b/packages/react-interactions/events/src/dom/__tests__/FocusWithin-test.internal.js
@@ -22,6 +22,7 @@ const initializeModules = hasPointerEvents => {
   setPointerEvent(hasPointerEvents);
   jest.resetModules();
   ReactFeatureFlags = require('shared/ReactFeatureFlags');
+  ReactFeatureFlags.enableScopeAPI = true;
   ReactFeatureFlags.enableDeprecatedFlareAPI = true;
   React = require('react');
   ReactDOM = require('react-dom');
@@ -362,6 +363,40 @@ describe.each(table)('FocusWithin responder', hasPointerEvents => {
       expect(onBlurWithin).toHaveBeenCalledWith(
         expect.objectContaining({isTargetAttached: false}),
       );
+    });
+
+    // @gate experimental
+    it('is called after a nested focused element is unmounted (with scope query)', () => {
+      const TestScope = React.unstable_createScope();
+      const testScopeQuery = (type, props) => true;
+      let targetNodes;
+      let targetNode;
+
+      const Component = ({show}) => {
+        const scopeRef = React.useRef(null);
+        const listener = useFocusWithin({
+          onBeforeBlurWithin(event) {
+            const scope = scopeRef.current;
+            targetNode = innerRef.current;
+            targetNodes = scope.DO_NOT_USE_queryAllNodes(testScopeQuery);
+          },
+        });
+
+        return (
+          <TestScope ref={scopeRef} DEPRECATED_flareListeners={[listener]}>
+            {show && <input ref={innerRef} />}
+          </TestScope>
+        );
+      };
+
+      ReactDOM.render(<Component show={true} />, container);
+
+      const inner = innerRef.current;
+      const target = createEventTarget(inner);
+      target.keydown({key: 'Tab'});
+      target.focus();
+      ReactDOM.render(<Component show={false} />, container);
+      expect(targetNodes).toEqual([targetNode]);
     });
 
     // @gate experimental
--- a/packages/react-reconciler/src/ReactFiberCommitWork.old.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.old.js
@@ -1110,25 +1110,6 @@ function commitNestedUnmounts(
   }
 }
 
-function detachFiber(fiber: Fiber) {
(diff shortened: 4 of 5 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1972,6 +1972,7 @@ function commitRootImpl(root, renderPriorityLevel) {
     while (nextEffect !== null) {
       const nextNextEffect = nextEffect.nextEffect;
       nextEffect.nextEffect = null;
```
gaearon, line 1974: Can this just be done inside `detachFiber` now?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2399,6 +2400,7 @@ function flushPassiveEffectsImpl() {
     const nextNextEffect = effect.nextEffect;
     // Remove nextEffect pointer to assist GC
     effect.nextEffect = null;
```
gaearon, line 2402: Same.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
trueadm: We do this for all effects, `detachFiber` only gets called on fibers that are flagged for deletion. So it's a different change.

trueadm:
It seems this causes a failing test on the www test for `ReactHooksWithNoopRenderer-test`. Any ideas anyone?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -3205,3 +3207,36 @@ function finishPendingInteractions(root, committedExpirationTime) {
     );
   }
 }
+
+function detachFiberPair(fiber: Fiber) {
+  if (fiber.effectTag & Deletion) {
```
sebmarkbage, line 3212: Does this function get inlined? If not, I wonder if we should put the conditional on the outside. My theory is that we save a bit because the inline cache type check is already done against the fiber at the parent callsite so it'll be quicker than to do another type check here. It doesn't matter if it's a Deletion but Deletions are only a fraction of all effects. So the common case is that this doesn't happen and in that case we can skip the function call and additional hidden class type check.

sebmarkbage:
Setting the `return` to `null` is a mechanism we use to figure out if something is unmounted during the setState traversal. We're probably doing that later now which covers up the warning. Only in www for some reason.

@bvaughn Would know more about the timing aspect of those.

trueadm:
@sebmarkbage It only happens with the `markUpdateTimeFromFiberToRoot` flag. Probably something related to `markUpdateTimeFromFiberToRoot` that has now changed.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -1522,10 +1503,10 @@ function commitDeletion(
     // Detach refs and call componentWillUnmount() on the whole subtree.
     commitNestedUnmounts(finishedRoot, current, renderPriorityLevel);
   }
+  current.return = null;
```
trueadm, line 1506: These are needed to make the failing test I mentioned in the comments of this PR pass, as per @sebmarkbage's response. If we could avoid doing this, it would be better.

bvaughn:
Sorry to chime in kind of late here, but I wouldn't expect the current change to work as is.

---

Before this change, the sequence of committing a root was:
1. before mutation stuff (e.g. `getSnapshotBeforeUpdate`)
1. mutation effects (e.g. placements, updates, deletes **and detach fibers**)
1. layout effects
1. cascade (if scheduled)

Then on a later tick
1. passive effects

---

After this change, the sequence of committing a root is:
1. before mutation stuff (e.g. `getSnapshotBeforeUpdate`)
1. mutation effects (e.g. placements, updates, deletes)
1. layout effects
1. cascade (if scheduled)

Then on a later tick
1. passive effects
1. **detach fibers**

---

I think this change is problematic for a few reasons:
* It means that any updates scheduled after deletion by a layout *or* passive effect would not get flagged as problematic using our current heuristic of a disconnected `return` pointer/path.

I don't think it would work to check the `effectTag` tag instead, because that wouldn't handle the case of a component being unmounted because one of its ancestors were deleted.

bvaughn:
Sorry to chime in kind of late here, but I wouldn't expect the current change to work as is.

---

Before this change, the sequence of committing a root was:
1. before mutation stuff (e.g. `getSnapshotBeforeUpdate`)
1. mutation effects (e.g. placements, updates, deletes **and detach fibers**)
1. layout effects
1. cascade (if scheduled)

Then on a later tick
1. passive effects

---

After this change, the sequence of committing a root is:
1. before mutation stuff (e.g. `getSnapshotBeforeUpdate`)
1. mutation effects (e.g. placements, updates, deletes)
1. layout effects
1. cascade (if scheduled)

Then on a later tick
1. passive effects
1. **detach fibers**

---

I think this change is problematic because it means that any updates scheduled after deletion by a layout *or* passive effect would not get flagged as problematic using our current heuristic of a disconnected `return` pointer/path.

I don't think it would work to just check the `effectTag` tag either, because that wouldn't handle the case of a component being unmounted because one of its ancestors were deleted. (In that case, the component itself wouldn't have a `Deletion` effect.) I think we would have to walk to the root and see if any ancestors had been scheduled for deletion, e.g.
```js
const root = markUpdateTimeFromFiberToRoot(fiber, expirationTime);
if (root === null) {
  warnAboutUpdateOnUnmountedFiberInDEV(fiber);
  return;
}

let current = fiber;
while (current !== null) {
  if (current.effectTag & Deletion) {
    warnAboutUpdateOnUnmountedFiberInDEV(fiber);
    return;
  }
  current = current.return;
}
```

But I don't think that would work either. I think it would fire prematurely for components in a tree scheduled for a deletion that has not yet been committed.

(It might also not work for unmounted roots?)

trueadm:
@bvaughn Thanks for the detailed responder. I guess there are two things we can do:

- When we use call `detachFiber`, we instead only `null` out the `return` pointer on the fiber and its alternate (I did that in one of the commits on this PR but then went back to what I had originally). That would alleviate one of the issues and allow us to detach the remaining properties later, which works for solutions that rely on recursive traversal of the tree from the top (like scopes), but for traversal that relies on going back up the tree, it wouldn't. So this wouldn't fix the changes I made to the ReactProfiler tests, which is unfortunate.
- We somehow signal that a fiber has been detached another way. Maybe we have another effect, or use another field? That way we can tell if something is unmounted, not by checking only if the `return` is `null` but by this other heuristic too – then this case would maybe work.

It's a bank holiday today and Monday, so I'll think about it over the weekend, but I'll likely not code anything till after the Easter break – and there's no rush, although it would be nice to understand the possible solutions and maybe someone else on the team has thought of something else too.

bvaughn:
> Specifically, the problem was that we were trying to traverse a fiber tree at a point where a bunch of the siblings were already "detached" – as in we already cleared their fields. This wasn't ideal, as **onBeforeBlur gets called during the same phase as where we detach fibers** so we want to be able to accurately traverse the fiber tree still.

If the problem is just one of timing- where you just needed all of the mutation effects to fire before the fibers were detached- then we could probably just move the detaching slightly later, rather than pushing it all the way out to where we flush passive effects.

For example, we could do it right after processing layout effects (although this would require walking the effects list another time so maybe it's not great).

```diff
diff --git a/packages/react-reconciler/src/ReactFiberCommitWork.old.js b/packages/react-reconciler/src/ReactFiberCommitWork.old.js
index 4482ee781..a2ce2bc24 100644
--- a/packages/react-reconciler/src/ReactFiberCommitWork.old.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.old.js
@@ -1110,25 +1110,6 @@ function commitNestedUnmounts(
   }
 }
 
-function detachFiber(fiber: Fiber) {
-  // Cut off the return pointers to disconnect it from the tree. Ideally, we
-  // should clear the child pointer of the parent alternate to let this
-  // get GC:ed but we don't know which for sure which parent is the current
-  // one so we'll settle for GC:ing the subtree of this child. This child
-  // itself will be GC:ed when the parent updates the next time.
-  fiber.return = null;
-  fiber.child = null;
-  fiber.memoizedState = null;
-  fiber.updateQueue = null;
-  fiber.dependencies = null;
-  fiber.alternate = null;
-  fiber.firstEffect = null;
-  fiber.lastEffect = null;
-  fiber.pendingProps = null;
-  fiber.memoizedProps = null;
-  fiber.stateNode = null;
-}
-
 function emptyPortalContainer(current: Fiber) {
   if (!supportsPersistence) {
     return;
@@ -1522,11 +1503,6 @@ function commitDeletion(
     // Detach refs and call componentWillUnmount() on the whole subtree.
     commitNestedUnmounts(finishedRoot, current, renderPriorityLevel);
   }
-  const alternate = current.alternate;
-  detachFiber(current);
-  if (alternate !== null) {
-    detachFiber(alternate);
-  }
 }
 
 function commitWork(current: Fiber | null, finishedWork: Fiber): void {
diff --git a/packages/react-reconciler/src/ReactFiberWorkLoop.old.js b/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
index f77dd308f..a05fe579e 100644
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
@@ -369,6 +369,37 @@ export function computeExpirationForFiber(
   return expirationTime;
 }
 
+function detachFiberPair(fiber: Fiber) {
+  const alternate = fiber.alternate;
+  detachFiber(fiber);
+  if (alternate !== null) {
+    detachFiber(alternate);
+  }
+}
+
+function detachFiber(fiber: Fiber) {
+  // Cut off the return pointers to disconnect it from the tree. Ideally, we
+  // should clear the child pointer of the parent alternate to let this
+  // get GC:ed but we don't know which for sure which parent is the current
+  // one so we'll settle for GC:ing the subtree of this child. This child
+  // itself will be GC:ed when the parent updates the next time.
+  fiber.alternate = null;
+  fiber.child = null;
+  fiber.dependencies = null;
+  fiber.firstEffect = null;
+  fiber.lastEffect = null;
+  fiber.memoizedProps = null;
+  fiber.memoizedState = null;
+  fiber.pendingProps = null;
+  fiber.return = null;
+  fiber.sibling = null;
+  fiber.stateNode = null;
+  fiber.updateQueue = null;
+  if (__DEV__) {
+    fiber._debugOwner = null;
+  }
+}
+
 export function scheduleUpdateOnFiber(
   fiber: Fiber,
   expirationTime: ExpirationTime,
@@ -1948,6 +1979,15 @@ function commitRootImpl(root, renderPriorityLevel) {
       }
     } while (nextEffect !== null);
 
+    nextEffect = firstEffect;
+    while (nextEffect !== null) {
+      const nextNextEffect = nextEffect.nextEffect;
+      if (nextEffect.effectTag & Deletion) {
+        detachFiberPair(nextEffect);
+      }
+      nextEffect = nextNextEffect;
+    }
+
     nextEffect = null;
 
     // Tell Scheduler to yield at the end of the frame, so the browser has an
```

trueadm:
@bvaughn Thanks for the feedback. I've applied the changes as you mentioned and things mostly work – I had to revert the changes I made to the profiler tests, as they no longer hold true now. Oddly, it seems the ReactART tests now fail with the recent changes, so I'll look into that and see why.

bvaughn:
No problem! Glad it was helpful :smile:

> I had to revert the changes I made to the profiler tests, as they no longer hold true now (as expected I guess).

Yeah, I'm not sure of a great way to "fix" those tests given the other things that depend on this sequencing.

sebmarkbage:
Yea I’m not sure this is the right fix given that it adds another traversal. With the removal of effect list traversal will be more expensive than it was.

I’m not sure the timing of beforeblur itself is correct given that it happens part way through a set of mutations. That makes that event much less versatile. Eg you can’t reason about the screen location you’re going from since the layout has already shifted.

sebmarkbage:
Possibly a quick fix might be to move beforeblur to the snapshot phase and somehow figure out if the active element is in a subtree of a Deletion effect.

trueadm:
I'll tackle this in another PR, as per Sebastian's suggestion.

PR #18566, opened 2020-04-09
Don't return from perform/completeUnitOfWork 

sebmarkbage:
Mutate workInProgress instead of returning. We started out returning it but we have since started mutating causing ambivalent behavior of what the methods were supposed to do since they're both mutating and returning. We now always mutate and never return anything from perform and complete.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
@@ -1208,7 +1208,7 @@ function prepareFreshStack(root, expirationTime) {
   }
 }
 
-function handleError(root, thrownValue) {
+function handleError(root, thrownValue): void {
   do {
     try {
       // Reset module-level state that was set during the render phase.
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.old.js
```diff
@@ -1517,45 +1517,47 @@ function performUnitOfWork(unitOfWork: Fiber): Fiber | null {
...
 
   ReactCurrentOwner.current = null;
-  return next;
 }
 
-function completeUnitOfWork(unitOfWork: Fiber): Fiber | null {
+function completeUnitOfWork(unitOfWork: Fiber): void {
   // Attempt to complete the current unit of work, then move to the next
   // sibling. If there are no more siblings, return to the parent fiber.
-  workInProgress = unitOfWork;
```
sebmarkbage, line 1530: This line was unnecessary since it's always the current workInProgress that got completed.

PR #18567, opened 2020-04-10
Bubble errors if processing the error itself errors

sebmarkbage:
If we've tried completing an incomplete boundary once and failed, we don't need to try again.

Avoids an infinite loop when something causes the error processing to fail.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -1202,6 +1202,7 @@ function prepareFreshStack(root, expirationTime) {
 
 function handleError(root, thrownValue): void {
   do {
+    let erroredWork = workInProgress;
     try {
       // Reset module-level state that was set during the render phase.
       resetContextDependencies();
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1228,24 +1229,34 @@ function handleError(root, thrownValue): void {
...
       );
-      completeUnitOfWork(workInProgress);
+      completeUnitOfWork(erroredWork);
     } catch (yetAnotherThrownValue) {
       // Something in the return path also threw.
       thrownValue = yetAnotherThrownValue;
+      erroredWork = workInProgress;
+      if (
+        erroredWork !== null &&
+        (erroredWork.effectTag & Incomplete) !== NoEffect
```
acdlite, line 1253: I don't think it matters, but I expected this to be `DidCapture`, since that's the one that says we attempted a second pass. Whereas `Incomplete` is set on all the fibers in between the source and the boundary.

Would it work to compare `workInProgress` to `erroredWork`?  And if they are the same, we know something in the error handling path went wrong?

acdlite: Sorry, I meant stash `workInProgress` before re-entering `completeUnitOfWork` and then compare to that.

sebmarkbage: It's not just boundaries that can cause this so DidCapture doesn't work but comparing workInProgress to erroredWork should work.

PR #18569, opened 2020-04-10
Overhauled release scripts and docs

bvaughn:
This PR updates our release script instructions to match our newer release process and CI task names.

It also renames several scripts to more clearly match their purpose:
* `create-next` ->`build-release-locally`
* `prepare-next` -> `prepare-release-from-ci`
* `promote-stable` -> `prepare-release-from-npm`

[**View formatted README**](https://github.com/bvaughn/react/blob/release-script-readme-update/scripts/release/README.md)

I re-ran all of the commands locally to make sure the renames/moves didn't break any imports.


scripts/release/README.md
```diff
@@ -26,13 +27,13 @@ If this is your first time running the release scripts, go to the `scripts/relea
 To prepare a build for a particular commit:
 1. Choose a commit from [the commit log](https://github.com/facebook/react/commits/master).
 2. Click the "“✓" icon and click the Circle CI "Details" link.
-3. Select the `build` job (**not** the `build_experimental` job; see the next section).
+3. Select the `process_artifacts ` job (**not** the `process_artifacts_experimental`job; see the next section)
```
acdlite, line 30: Really the only thing that matters is which workflow it's from because of this part:

https://github.com/facebook/react/blob/af1b039bdd5a8b5def5d51acad00b79e9b7b377c/scripts/release/utils.js#L44-L56

So the previous instructions happened to also work

scripts/release/README.md
```diff
@@ -65,10 +71,10 @@ If the OTP code expires while publishing, re-run this command and answer "y" to
 
 Stable releases should always be created from the "next" channel. This encourages better testing of the actual release artifacts and reduces the chance of unintended changes accidentally being included in a stable release.
 
-To prepare a stable release, choose a "next" version and run the [`prepare-stable`](#prepare-stable) script <sup>1</sup>:
+To prepare a stable release, choose a "next" version and run the [`promote-to-stable`](#promote-to-stable) script <sup>1</sup>:
 
 ```sh
-scripts/release/prepare-stable.js --version=0.0.0-241c4467e
+scripts/release/promote-to-stable.js --version=0.0.0-241c4467e
```
acdlite, line 77: Can we call this `download-stable` or something? Since it corresponds to `download-release`, only difference being that it downloads the tarballs from npm.

I was thinking at some point we should consolidate these scripts and use the `download-release` process for stable releases, too — i.e. always download from CI artifacts. The rationale for downloading from npm was to reduce the amount of processing in between the prerelease tarball and the final release tarball. But since they all come from the CI tarball, anyway, it actually seems safer to me to use the same script for both types of releases. Then the only additional point of failure is what versions are given to the version rewriting step.

scripts/release/README.md
```diff
@@ -107,13 +113,13 @@ Once you have cherry picked all of the commits you want to include in the releas
...
 
-Once CI is complete, follow the regular [**next**](#publishing-next) and [**promote to stable**](#publishing-a-stable-release) processes.
+Once CI is complete, follow the regular [**next**](#publishing-release) and [**promote to stable**](#publishing-a-stable-release) processes.
 
 <sup>1: The `build-info.json` artifact can also be used to identify the appropriate commit (e.g. [unpkg.com/react@16.8.3/build-info.json](https://unpkg.com/react@16.8.3/build-info.json) shows us that react version 16.8.3 was created from commit [`29b7b775f`](https://github.com/facebook/react/commit/29b7b775f)).</sup>
 
 # Scripts
 
-## `create-next`
+## `build-release`
```
acdlite, line 122: I forgot this script existed 😮Can we call it something less ambiguous, like `build-release-locally` or something? Just so it doesn't get confused with `prepare-release`.

I don't think I've ever used it because pushing to a PR to kick off a CI build tends to be faster even with all the extra checks because of the build parallelization.

scripts/release/README.md
bvaughn: Right. The reason I changed the target is because the new display strings made the old targets hard to find. (They got truncated in the popup.) :smile:

scripts/release/README.md
bvaughn: Happy to change the name but I don't like that one because the script doesn't download a *stable* release. It downloads a *next* release, and prepares it to be published as stable.

> I was thinking at some point we should consolidate these scripts and use the download-release process for stable releases, too — i.e. always download from CI artifacts. The rationale for downloading from npm was to reduce the amount of processing in between the prerelease tarball and the final release tarball. But since they all come from the CI tarball, anyway, it actually seems safer to me to use the same script for both types of releases. Then the only additional point of failure is what versions are given to the version rewriting step.

The one thing the current approach gives us that you the one you mention wouldn't is that it forces us to release a *next* release first (and presumably test it) before then promoting it to stable.

acdlite: > the script doesn't download a stable release. It downloads a next release, and prepares it to be published as stable.

That's true implementation-wise, but from the user's perspective I don't think it matters because the end result is the same: a stable release in the `build` directory ready to be published

> The one thing the current approach gives us that you the one you mention wouldn't is that it forces us to release a next release first (and presumably test it) before then promoting it to stable.

Right, though you could still enforce that by pinging npm to confirm that a prerelease was  published.

acdlite: Maybe `prepare-next` and `prepare-stable`? I don't really care what the verb is but I do think it would be nice to be symmetrical

bvaughn: > Maybe `prepare-`next and `prepare-stable`?

This is what we had before :wink: but it's not very descriptive/accurate, given that `prepare-next` is the script you used to both prepare a *next* release and an *experimental* release. I don't think we should have the NPM tag in the name unless the script is *only* used for that tag (like "stable")

bvaughn: Maybe `prepare-stable` is still okay since that's what it's doing.

acdlite: Oh right :D

How about `prepare-release` versus `prepare-prerelease`? A bit awkward but avoids the "stable" versus "experimental" confusion

acdlite: Or we could go the other way and add a separate command for experimental:

- `prepare-stable`
- `prepare-next`
- `prepare-experimental`

acdlite: Or (sorry just thinking out loud) a single `prepare-release` command with a `channel` argument.

bvaughn: Uh...sure, that sounds fine...I don't really want to re-read and update the README again though 😭

bvaughn: It's fine. I can update it one more time.

scripts/release/README.md
bvaughn: Sure. Renamed

scripts/release/README.md
bvaughn: Maybe?
* `build-release-locally`: builds local source
* `prepare-release-from-ci`: prepares CI artifacts to be published as whatever
* `prepare-release-from-npm`: prepares "next" to be promoted to "stable"

bvaughn: I don't know. In a way, it becomes harder to evaluate names the more possibilities we toss out. I think any fo the revisions I've pushed so far are an improvement over what was there before. I'd like to prefer making another major change (like combining the scripts) right now b'c it doesn't feel like the most valuable use of time.

bvaughn:
Thanks for the feedback! 🙇 I'm going to merge this change :smile: We can iterate more later!

Issue #17356, opened 2019-11-13
Memoized child of Suspense component doesn't update when Context updates.


jcarroll2007:
**Do you want to request a *feature* or report a *bug*?**
bug (I think?)

**What is the current behavior?**
```javascript
const [value, setValue] = useState("default");
return (
  <div className="App">
    <input value={value} onChange={e => setValue(e.target.value)} />
    <div>
      <Value.Provider value={value}>
        <Suspense fallback={<div>loading</div>}>
          <MemoizedChild />
         </Suspense>
       </Value.Provider>
     </div>
  </div>
)
```

When using a memoized functional component (`MemoizedChild` in above example) in conjunction with `Context` as a child of a `React.Suspense` component, there seems to be a bug in which `MemoizedChild` does not update when the context it uses changes. For the full example, see my codesandbox below.

In the codesandbox, if you change the value of the input, the new value is provided to the context which causes the hook used in `MemoizedChild` (`useValue`) to throw a promise. This flips `Suspense` to the fallback state and when the promise resolves `MemoizedChild`'s state is not updated with the proper context value because (I'm assuming) the memoized value of `MemoizedChild` is the one that contained the previous context value and technically no props have changed, so that makes sense why it wouldn't have updated. However, this seems like it would be unexpected behavior.

https://codesandbox.io/s/react-suspense-maybe-bug-sznbk

**What is the expected behavior?**
I would expect that `MemoizedChild` would be re-rendered with the new provided value.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
I'm assuming all of them that contain Suspense and memo. So, since 16.8?

aweary:
Thanks for the report @jcarroll2007, this does look like a bug. I [reproduced with the experimental build as well](https://codesandbox.io/s/react-suspense-maybe-bug-k2hdm) and also verified that this only affects legacy mode (`ReactDOM.render`), Concurrent and Blocking Mode both behave as expected.

aweary:
@gaearon can you take a look at this when you get a chance?

overlookmotel:
It appears to relate to the same component which calls `useContext()` throwing.

Here's the same example but with `MemoizedChild` split into two components - one with the `useContext()` hook, and a child *inside it* which throws.

https://codesandbox.io/s/react-suspense-maybe-bug-c62nh

This behaves as expected.

gaearon:
Would anyone like to turn this into a failing test? This will get us closer to finding a fix.

[...]

PR #18572, opened 2020-04-10
Failing test related to usage of Suspense and Context in memoized component

Tsdevendra1:
Failing test as requested related to #17356.

This is a failing test which shows that when useContext and throwing a promise are used in a memoized component, it fails to update after the promise has resolved.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberThrow.new.js
+++ b/packages/react-reconciler/src/ReactFiberThrow.new.js
@@ -21,6 +21,7 @@ import {
   HostRoot,
   SuspenseComponent,
   IncompleteClassComponent,
+  SimpleMemoComponent,
 } from './ReactWorkTags';
 import {
   DidCapture,
@@ -264,6 +265,10 @@ function throwException(
               update.tag = ForceUpdate;
               enqueueUpdate(sourceFiber, update);
             }
+          } else if (sourceFiber.tag === SimpleMemoComponent) {
+            const update = createUpdate(NoWork, Sync, null);
+            update.tag = ForceUpdate;
+            enqueueUpdate(sourceFiber, update);
           }
 
           // The source fiber did not complete. Mark it with Sync priority to
--- a/packages/react-reconciler/src/ReactFiberThrow.old.js
+++ b/packages/react-reconciler/src/ReactFiberThrow.old.js
@@ -21,6 +21,7 @@ import {
   HostRoot,
   SuspenseComponent,
   IncompleteClassComponent,
+  SimpleMemoComponent,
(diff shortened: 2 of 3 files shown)


acdlite:
Ah nice test case! Confirmed this is a bug in "legacy" mode Suspense. The behavior is correct in Concurrent Mode. (If you have a Concurrent Mode test case, please let me know.)

Note that technically, in Legacy Mode, we don't officially support suspending except for `React.lazy`. We'll fix this anyway though.

acdlite:
It's essentially the same bug that we fixed here for classes:

https://github.com/facebook/react/blob/98d410f5005988644d01c9ec79b7181c3dd6c847/packages/react-reconciler/src/ReactFiberThrow.new.js#L260-L265

Need to decide on least bad way to fix it. My initial instinct is a special effect tag that prevents a bail out (`ForceUpdate`). It's a hack but it works. That's how `this.forceUpdate` for classes used to work.

Then we would add that check to this  condition here, to prevent us from setting  `didReceiveUpdate` to false.

https://github.com/facebook/react/blob/98d410f5005988644d01c9ec79b7181c3dd6c847/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L517-L522

Tsdevendra1:
Hi @acdlite, I've added a test case for the concurrent mode which passes. I'll have a go at implementing the solution you outlined and getting back to you.

Tsdevendra1:
I think I'm a bit stuck on this or maybe I didn't understand what you suggested properly. When I try to enqueue a `ForceUpdate`, the `updateQueue` is null for the memo fiber (which according to this: **https://github.com/facebook/react/blob/32bb44c80ac64b3c96fa9e03b6ecb6dce6a0de6e/packages/react-reconciler/src/ReactUpdateQueue.new.js#L219** is because the fiber has been unmounted). But this means nothing gets added to the update queue and as a result I can't differentiate that fiber when `updateSimpleMemoComponent` is called after the promise is resolved. Am I going about this the wrong way?

On a separate note, there seem to be .old and .new versions of some files, should I be modifying both or just one? I asked because when I build my local react using `yarn build react/index,react-dom/index,scheduler --type=NODE` (suggested [here](https://reactjs.org/docs/how-to-contribute.html)) it seems to be using the .old version of the files, I'm not sure how to use the .new version of the files.

gaearon:
If you don't plan to keep working on this, would you mind keeping the failing test open as a reminder?

gaearon:
https://github.com/facebook/react/pull/19216

eps1lon:
Superseded by #19216.

PR #18574, opened 2020-04-10
Add test helpers for gating by feature

acdlite:
Adds a new test helper, `it.gate`. You pass it the name of a feature flag, and if the flag is enabled, it asserts that the test passes. If  the flag is off, it asserts that the test fails. Example:

```js
it.gate('enableNewReconciler', 'test that only passes in new reconciler', () => {
  // ... test goes here
});
```

(`test.gate` is equivalent and makes perhaps makes more grammatical sense, but I'm using `it.gate` in these examples since we use `it` everywhere in our tests.)

You can test multiple flags by passing an array:

```js
it.gate(
  ['enableNewReconciler', 'enableModernEventSystem'],
  'test that passes only in new reconciler with modern event system', () => {
    // ... test goes here
  }
);
```

If any of the flags are disabled, the test is expected to fail.

You can check if a flag is off by adding a `!` to the front:

```js
it.gate(
  '!enableNewReconciler',
  'test that passes only in the old reconciler', () => {
    // ... test goes here
  }
);
```

For anything more advanced, you can pass a function:

```js
it.gate(
  flags => flags.warnAboutSpreadingKeyToJSX || !__DEV__,
  'should warn when keys are passed as part of props', () => {
    // Test passes when the flag is enabled, or in prod since
    // there are no warnings in prod
  }
);
```

I also added aliases for `it.old` and `it.new`, for testing the different reconciler forks:

```js
it.old('test that passes only in the old reconciler', () => {
  // ... test goes here
});
it.new('test that passes only in the new reconciler', () => {
  // ... test goes here
});
```

##  Next steps

Longer term I think it would nice if these were implemented as pragmas instead:

```js
// @gate enableNewReconciler
it('test that passes only in the new reconciler', () => {
  // ... test goes here
});
```

### DIFF
--- a/packages/react/src/__tests__/ReactElementJSX-test.js
+++ b/packages/react/src/__tests__/ReactElementJSX-test.js
@@ -369,8 +369,10 @@ describe('ReactElement.jsx', () => {
     );
   });
 
-  if (require('shared/ReactFeatureFlags').warnAboutSpreadingKeyToJSX) {
-    it('should warn when keys are passed as part of props', () => {
+  it.gate(
+    flags => flags.warnAboutSpreadingKeyToJSX || !__DEV__,
+    'should warn when keys are passed as part of props',
+    () => {
       const container = document.createElement('div');
       class Child extends React.Component {
         render() {
@@ -391,8 +393,8 @@ describe('ReactElement.jsx', () => {
           'Explicitly pass a key after spreading props in your JSX call. ' +
           'E.g. <Child {...props} key={key} />',
       );
-    });
-  }
+    },
+  );
 
   it('should not warn when unkeyed children are passed to jsxs', () => {
     const container = document.createElement('div');
--- a/scripts/jest/installGatedTestHelpers.js
+++ b/scripts/jest/installGatedTestHelpers.js
@@ -1,0 +1,142 @@
+'use strict';
+
+/* eslint-disable no-for-of-loops/no-for-of-loops */
+
+const noop = () => {};
+
+async function expectTestToFail(callback, errorMsg) {
+  if (callback.length > 0) {
+    throw Error(
+      'Gated test helpers do not support the `done` callback. Return a ' +
+        'promise instead.'
+    );
+  }
+  try {
+    const maybePromise = callback();
+    if (
+      maybePromise !== undefined &&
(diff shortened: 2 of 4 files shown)


acdlite:
Longer term I think it would nice if these were implemented as pragmas instead:

```js
// @gate enableNewReconciler
it('test that passes only in the new reconciler', () => {
  // ... test goes here
});
```

packages/react/src/__tests__/ReactElementJSX-test.js
```diff
@@ -369,8 +369,10 @@ describe('ReactElement.jsx', () => {
     );
   });
 
-  if (require('shared/ReactFeatureFlags').warnAboutSpreadingKeyToJSX) {
-    it('should warn when keys are passed as part of props', () => {
+  it.gate(
+    'warnAboutSpreadingKeyToJSX',
+    'should warn when keys are passed as part of props',
```
acdlite, line 374: Hmm this one is interesting because it the gate doesn't actually fail in prod because there are no warnings in prod...

acdlite: I think I'll add support for checking  `__DEV__`, too, even though it's not a flag. That kind fits with how I was planning to eventually add support for gating on a channel.

scripts/jest/installGatedTestHelpers.js
```diff
@@ -0,0 +1,142 @@
...
+    );
+  } else {
+    testFn(testName, fn);
+  }
+}
+
+function installGatedTestHelpers(channel) {
+  const it = global.it;
+  const fit = global.fit;
+  const xit = global.xit;
```
sebmarkbage, line 87: I kind of was hoping to switch to the `test()` form. Does this patch that too? I guess it’s just an alias for `it`?

acdlite: Yeah `it` is an alias

acdlite:
I went ahead and opened an alternative PR for the pragma proposal: https://github.com/facebook/react/pull/18581

I prefer that one, but I'll keep this PR open for reference.

acdlite:
Closed by #18581

PR #18576, opened 2020-04-11
Store unique ID counter on SSR request object and Add Prefix for separate renders

lunaruan:
There is a worry that `useOpaqueIdentifier` might run out of unique IDs if running for long enough. This PR moves the unique ID counter so it's generated per server renderer object instead. For people who render different subtrees, this PR adds a `prefix` option to `renderToString`, `renderToStaticMarkup`, `renderToNodeStream`, and `renderToStaticNodeStream` so identifiers can be differentiated for each individual subtree.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -1078,11 +1078,6 @@ export function makeClientIdInDEV(warnOnAccessInDEV: () => void): OpaqueIDType {
   };
 }
 
-let serverId: number = 0;
-export function makeServerId(): OpaqueIDType {
-  return 'R:' + (serverId++).toString(36);
-}
-
 export function isOpaqueHydratingObject(value: mixed): boolean {
   return (
     value !== null &&
--- a/packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js
+++ b/packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js
@@ -4,6 +4,7 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
+import type {ServerOptions} from './ReactPartialRenderer';
 
 import {Readable} from 'stream';
 
@@ -11,11 +12,15 @@ import ReactPartialRenderer from './ReactPartialRenderer';
 
 // This is a Readable Node.js stream which wraps the ReactDOMPartialRenderer.
 class ReactMarkupReadableStream extends Readable {
-  constructor(element, makeStaticMarkup) {
+  constructor(element, makeStaticMarkup, options) {
     // Calls the stream.Readable(options) constructor. Consider exposing built-in
     // features like highWaterMark in the future.
     super({});
-    this.partialRenderer = new ReactPartialRenderer(element, makeStaticMarkup);
+    this.partialRenderer = new ReactPartialRenderer(
+      element,
+      makeStaticMarkup,
+      options,
+    );
   }
 
   _destroy(err, callback) {
@@ -36,8 +41,8 @@ class ReactMarkupReadableStream extends Readable {
  * server.
  * See https://reactjs.org/docs/react-dom-server.html#rendertonodestream
  */
-export function renderToNodeStream(element) {
-  return new ReactMarkupReadableStream(element, false);
+export function renderToNodeStream(element, options?: ServerOptions) {
+  return new ReactMarkupReadableStream(element, false, options);
 }
 
 /**
@@ -45,6 +50,6 @@ export function renderToNodeStream(element) {
  * such as data-react-id that React uses internally.
  * See https://reactjs.org/docs/react-dom-server.html#rendertostaticnodestream
  */
-export function renderToStaticNodeStream(element) {
-  return new ReactMarkupReadableStream(element, true);
+export function renderToStaticNodeStream(element, options?: ServerOptions) {
+  return new ReactMarkupReadableStream(element, true, options);
 }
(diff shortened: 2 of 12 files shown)


packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1174,9 +1174,8 @@ export function makeClientIdInDEV(warnOnAccessInDEV: () => void): OpaqueIDType {
   };
 }
 
-let serverId: number = 0;
-export function makeServerId(): OpaqueIDType {
-  return 'R:' + (serverId++).toString(36);
+export function makeServerId(prefix: ?string, serverId: number): OpaqueIDType {
```
gaearon, line 1177: Why is `prefix` optional? We might as well always pass a string from the outside.

packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js
```diff
@@ -36,15 +37,21 @@ class ReactMarkupReadableStream extends Readable {
  * server.
  * See https://reactjs.org/docs/react-dom-server.html#rendertonodestream
  */
-export function renderToNodeStream(element) {
-  return new ReactMarkupReadableStream(element, false);
+export function renderToNodeStream(
+  element,
+  options: PartialRendererOptions | void,
```
gaearon, line 42: Should it be a public API? I suppose people won't know to pass this. Can it be autogenerated or something?

packages/react-dom/src/client/ReactDOMHostConfig.js
lunaruan: We want to minimize the number of bytes we send, so we don't want to append a `prefix` unless necessary (ie for apps that render multiple subtrees on the server and then stitch them together).

packages/react-dom/src/server/ReactDOMNodeStreamRenderer.js
lunaruan: We won't know when to generate the `prefix` because we don't want to prepend one unless necessary on the server, so regardless we would have to add an option.

gaearon:
I'm not sure I care that much about the "stitching together" use case. Is it legit? I guess we use it.

I'm more concerned about clashes on the client. E.g. two React apps but only one of them uses server rendering. But their IDs clash. Do we have a plan for this?

packages/react-dom/src/client/ReactDOMHostConfig.js
bvaughn: Flow types suggest "prefix" is optional but "serverId" is required. Is this the intent?

Putting an optional param before a required param seems wrong.

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -838,6 +858,10 @@ class ReactDOMServerRenderer {
 
     const prevThreadID = currentThreadID;
     setCurrentThreadID(this.threadID);
+    const prevUniqueID = currentUniqueID;
+    setCurrentUniqueID(this.uniqueID);
```
sebmarkbage, line 862: You pass this value and increment it but you don't actually ever update `this.uniqueID`. For the synchronous renderToString this won't show up (except maybe for nested calls which we have a open issue about). However for two streaming renders that read multiple times, it'll reset to zero between each read.

You might want to write a unit test that catches this as since it could be pretty bad if we regress on it.

As a solution I'd just pass `this` instead. setCurrentRenderer(this) and read both the uniqueID, threadID and prefix from the renderer directly.

sebmarkbage:
@gaearon The plan for two React apps on the client is to use some kind of global on the client to coordinate this. Unlike the server, we actually have an opportunity to automatically detect that.

sebmarkbage:
> E.g. two React apps but only one of them uses server rendering. But their IDs clash. Do we have a plan for this?

To be clear server IDs use a different prefix (uppercase) than the client (lowercase) so that they never collide. The more likely clash is just two React versions on the same document. They'll immediately clash from the first ID.

sebmarkbage:
The client could also use a time or random based prefix. It's not as sensitive to long IDs as the server.

packages/react-dom/src/server/ReactPartialRenderer.js
```diff
@@ -79,6 +79,10 @@ import {validateProperties as validateARIAProperties} from '../shared/ReactDOMIn
 import {validateProperties as validateInputProperties} from '../shared/ReactDOMNullInputValuePropHook';
 import {validateProperties as validateUnknownProperties} from '../shared/ReactDOMUnknownPropertyHook';
 
+export type PartialRendererOptions = {
```
sebmarkbage, line 82: Let's give this type a name that you'd expect the external API to refer to because it's now part of the public API. Flow/TS will likely just copy this name upstream so lets name it what it should have.

`ServerOptions`?

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -495,15 +500,18 @@ function useTransition(
...
+  );
 }
 
 function noop(): void {}
 
-export let currentThreadID: ThreadID = 0;
-
-export function setCurrentThreadID(threadID: ThreadID) {
-  currentThreadID = threadID;
+export let currentPartialRenderer = {};
```
sebmarkbage, line 512: This type isn't valid. I think it just accidental works since it's an "open type".

Set this to null so it fails if something goes wrong. Let's fool Flow though so you don't need null checks.

```
export let currentPartialRenderer: PartialRenderer = (null: any);
```

PR #18580, opened 2020-04-11
Feature: Add support to `exhaustive-deps` rule for any hook ending with `Effect`

airjp73:
## Summary

Following up on @gaearon 's comment [here](https://github.com/facebook/react/pull/16912#issuecomment-611724673). My original motivation was to document the `additionalHooks` option in the `exhaustive-deps` lint rule. He suggested an alternate solution by linting any hook ending with `Effect`. This PR adds support for that.

## Test Plan

I find that eslint rule tests are pretty thorough so all my testing is just through those test cases. I copied all the tests I could find for the `additionalHooks` option and reworked them to use a `useXEffect` naming convention instead of the `additionalHooks` option. I also tweaked some of the `additionalHooks` tests that were testing for `useCustomEffect` and changed it to `useCustomHook` because `useCustomEffect` would get caught anyway after these changes.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -360,12 +360,41 @@ const tests = {
     {
       code: normalizeIndent`
         function MyComponent(props) {
+          useCustomHook(() => {
+            console.log(props.foo);
+          });
+        }
+      `,
+      options: [{additionalHooks: 'useCustomHook'}],
+    },
+    {
+      code: normalizeIndent`
+        function MyComponent(props) {
+          useCustomHook(() => {
+            console.log(props.foo);
+          }, [props.foo]);
+        }
+      `,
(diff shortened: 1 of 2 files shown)


karlvr:
Oh my code is so much nicer. Thank you thank you.

gaearon:
Thanks.

jquense:
This heuristic seems very prone to false positives, we have a bunch of custom "Effect" hooks that don't take dep arguments and now require pragma's to avoid warnings about potential infinite updates

Glinkis:
I am using [hooks.macro](https://github.com/yuchi/hooks.macro), which adds a few hook utilities that doesn't take a dependency array. This now incorrectly shows up as a warning.

gaearon:
One thing we could do is to only run checks if you have a function followed by an array, in addition to ending with Effect. But I'm leaning towards reverting this altogether.

gaearon:
I'm going to have to revert this because the heuristic has too many reasonable false positives.
https://github.com/facebook/react/pull/19004

gaearon:
This is removed in `eslint-plugin-react-hooks@4.0.3`. Thanks all for the input.

PR #18581, opened 2020-04-12
Add pragma for feature testing: @gate

acdlite:
## Alternative to #18574

The `@gate` pragma declares under which conditions a test is expected to pass.

If the gate condition passes, then the test runs normally (same as if there were no pragma).

If the conditional fails, then the test is **expected to fail**. This ensures that all tests run in all environments and reduces the likelihood of a test being completely skipped.

An alternative to `it.experimental` and similar proposals.

### Basic:

```js
// @gate enableBlocksAPI
test('passes only if Blocks API is available', () => {/*...*/})
```

### Negation:

```js
// @gate !disableLegacyContext
test('depends on a deprecated feature', () => {/*...*/})
```

### Multiple flags:

```js
// @gate enableNewReconciler
// @gate experimental
test('only passes in new fork and in experimental channel', () => {/*...*/})
```

### Logical operators (yes, I'm sorry):

```js
// @gate experimental && (enableNewReconciler || disableSchedulerTimeoutBasedOnReactExpirationTime)
test('experimental, doesn\'t work in old fork unless Scheduler timeout flag is disabled', () => {/*...*/})
```

### Strings, and comparison operators

No use case yet but I figure eventually we'd use this to gate on different release channels:

```js
// @gate channel ===  "experimental" || channel === "modern"
test('works in OSS experimental or www modern', () => {/*...*/})
```

## How does it work?

The output of the transform is pretty straightforward:

Input:
```js
// @gate a && (b || c)
test('some test', () => {/*...*/})
```

Output:

```js
_test_gate(ctx => ctx.a && (ctx.b || ctx.c), 'some test', () => {/*...*/});
```

It also works  with `it`, `it.only`, and `fit`. It leaves `it.skip` and  `xit` alone because those tests are disabled anyway.

`_test_gate` is a global method that I set up in our Jest config. It works about the same as the existing `it.experimental` helper.

The context (`ctx`) argument is whatever we want it to be. I set it up so that it throws if you try to access a flag that doesn't exist. I also added some shortcuts for common gating conditions, like `experimental` and `stable`; and `old` and `new`:

```js
// @gate experimental
test('experimental feature', () => {/*...*/})

// @gate stable
test('stable feature, deprecated in experimental', () => {/*...*/})

// @gate old
test('only passes in old reconciler', () => {/*...*/})

// @gate new
test('only passes in new reconciler', () => {/*...*/})
```

## Why implement this as a pragma instead of a runtime API?

- Doesn't require monkey patching built-in Jest methods. Instead it compiles to a runtime function that composes Jest's API.
- Will be easy to upgrade if Jest ever overhauls their API or we switch to a different testing framework (unlikely but who knows).
- It feels lightweight so hopefully people won't feel gross using it. For example, adding or removing a gate pragma will never affect the indentation of the test, unlike if you wrapped the test in a conditional block.

### DIFF
--- a/scripts/babel/__tests__/transform-test-gate-pragma-test.js
+++ b/scripts/babel/__tests__/transform-test-gate-pragma-test.js
@@ -166,6 +166,13 @@ describe('transform test-gate-pragma: actual runtime', () => {
   // These tests use the actual gating runtime used by the rest of our
   // test suite.
 
+  // @gate __DEV__
+  test('__DEV__', () => {
+    if (!__DEV__) {
+      throw Error("Doesn't work in production!");
+    }
+  });
+
   // Always should fail because of the unguarded console.error
   // @gate false
   test('works with console.error tracking', () => {
--- a/scripts/babel/__tests__/transform-test-gate-pragma-test.js
+++ b/scripts/babel/__tests__/transform-test-gate-pragma-test.js
@@ -1,0 +1,163 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+'use strict';
+
+describe('transform-test-gate-pragma', () => {
+  // Fake runtime
+  // eslint-disable-next-line no-unused-vars
+  const _test_gate = (gateFn, testName, cb) => {
+    test(testName, (...args) => {
+      shouldPass = gateFn(context);
+      return cb(...args);
+    });
+  };
+
+  // eslint-disable-next-line no-unused-vars
+  const _test_gate_focus = (gateFn, testName, cb) => {
+    // NOTE: Tests in this file are not actually focused because the calls to
+    // `test.only` and `fit` are compiled to `_test_gate_focus`. So if you want
+    // to focus something, swap the following `test` call for `test.only`.
+    test(testName, (...args) => {
+      shouldPass = gateFn(context);
+      isFocused = true;
+      return cb(...args);
+    });
+  };
+
+  // Feature flags, environment variables, etc. We can configure this in
+  // our test set up.
+  const context = {
+    flagThatIsOff: false,
+    flagThatIsOn: true,
+    environment: 'fake-environment',
+  };
+
+  let shouldPass;
+  let isFocused;
+  beforeEach(() => {
+    shouldPass = null;
+    isFocused = false;
+  });
+
+  test('no pragma', () => {
+    expect(shouldPass).toBe(null);
+  });
+
+  // unrelated comment
+  test('no pragma, unrelated comment', () => {
+    expect(shouldPass).toBe(null);
+  });
+
+  // @gate flagThatIsOn
+  test('basic positive test', () => {
+    expect(shouldPass).toBe(true);
+  });
+
+  // @gate flagThatIsOff
+  test('basic negative test', () => {
+    expect(shouldPass).toBe(false);
+  });
+
+  // @gate flagThatIsOn
+  it('it method', () => {
+    expect(shouldPass).toBe(true);
+  });
+
+  /* eslint-disable jest/no-focused-tests */
+
+  // @gate flagThatIsOn
+  test.only('test.only', () => {
+    expect(isFocused).toBe(true);
+    expect(shouldPass).toBe(true);
+  });
+
+  // @gate flagThatIsOff
+  it.only('it.only', () => {
+    expect(isFocused).toBe(true);
+    expect(shouldPass).toBe(false);
+  });
+
+  // @gate flagThatIsOn
+  fit('fit', () => {
+    expect(isFocused).toBe(true);
+    expect(shouldPass).toBe(true);
+  });
+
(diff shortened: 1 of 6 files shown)


acdlite:
I should maybe just swap out out the tiny parser I wrote for actual JavaScript. I started out with a more restrictive syntax with ANDs and ORs but then eventually said fuck it and copied JS.

scripts/babel/transform-test-gate-pragma.js
```diff
@@ -0,0 +1,328 @@
...
+          switch (callee.type) {
+            case 'Identifier': {
+              if (
+                callee.name === 'test' ||
+                callee.name === 'it' ||
+                callee.name === 'fit'
+              ) {
+                const comments = statement.leadingComments;
+                if (comments !== undefined) {
+                  const condition = blah(comments);
```
acdlite, line 277: Ahaha forgot to give this a name

Mathspy:
Sorry to barge in! I am curious if this was inspired by a very similar feature in another language. I know Sebastian mentioned taking inspiration from Rust in few occasions recently so I am wondering if you or the whole team are in the same boat
Thanks for this PR it inspired me to make something of the same vein ❤️ And thanks for continuous lovely work on React ❤️

acdlite:
@Mathspy I'm glad you find it interesting!

> I am curious if this was inspired by a very similar feature in another language

Hmm I'm sure I was inspired indirectly via exposure but I didn't have a specific feature in mind. It sounds like you're implying Rust has a similar feature. I am decently familiar with Rust but I actually haven't used it much. If you're aware of prior art, please let me know!

> Thanks for this PR it inspired me to make something of the same vein

Neat! I was thinking that if we use this for a while and like it, we could try to upstream this to Jest.

Mathspy:
Oh I see!
For prior art, in Rust there is a specific [attribute macro](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros) which work pretty much like this, in that it compile keeps the the module/function/statement that it is are placed over or removes it based on a predicate, here's an example very similar to this:

```rust
#[cfg(target_os = "macos")]
fn macos_only() {
 // ...
}
```
This marks the module with a [configuration predicate](https://doc.rust-lang.org/reference/conditional-compilation.html) telling Rust to compile it only for the target of MacOS, otherwise it's completely removed. Another example:
```rust
#[cfg(test)]
mod tests {
    // ...
}
```

This will be compiled only when running in test mode (similar `NODE_ENV=test` for JS/Node) this allows Rust's unit tests to live in the same file as the units they are testing without being forced to export (mark as public for Rust) some internals 

A place where I can see this really shining in JS is actually:
```rust
// marks module/function/statement to be conditionally compiled only in debug mode (AKA dev mode; not production)
#[cfg(debug_assertions)] 
```

I think that using a `// @dev` feature gate in JS would be much more elegant than:
```js
if (__DEV__) {
  // dev mode assertions
}
```

I am definitely going to try making `// @dev` and maybe even `// @test` sometime later this week!

acdlite:
Ah yes I was aware of that feature! I thought you meant more the "negative testing" aspect, where if the test doesn't pass the gating condition, you run it anyway and assert that it fails. That surely has prior art somewhere but I can't think of an example.

Mathspy:
Aha, I definitely share the feeling of "knowing" that specific aspect from somewhere too but just being unable to put my finger on it. If I happen to remember I will definitely leave a comment here!

scripts/jest/setupTests.js
```diff
@@ -231,15 +254,19 @@ if (process.env.REACT_CLASS_EQUIVALENCE_TEST) {
...
+      resetAllUnexpectedConsoleCalls();
       return;
     }
-    throw Error(
-      'Tests marked experimental are expected to fail, but this one passed.'
-    );
+    throw Error(errorMsg);
   };
 
+  // TODO: Deprecate these helpers in favor of @gate pragma
```
bvaughn, line 269: Why? What value does a custom pragma offer for iterating on a local focused test?

scripts/babel/__tests__/transform-test-gate-pragma-test.js
```diff
@@ -0,0 +1,187 @@
...
+    console.warn('Should cause test to fail');
+  });
+
+  // @gate false
+  test('works with console tracking if error is thrown before end of test', () => {
+    console.warn('Please stop that!');
+    console.error('Stop that!');
+    throw Error('I told you to stop!');
+  });
+});
```
bvaughn, line 187: Neat testing strategy ^

scripts/jest/setupTests.js
acdlite: Oh I meant it.experimental, not the focus helpers. Definitely not removing those!

bvaughn: Ooh, I misunderstood that comment!

That makes sense then.

acdlite:
I'm a little hesitant about the DSL aspect but I'm not too worried because I think in 99% percent of cases, you'll only use a a single flag or a combination of flags, no operators.

So my exit strategy is:

- Remove DSL and only support single flags. For more complex cases, add a new flag alias to the TestFlags file.
- Or, remove the pragma entirely by codemodding to if statements.

PR #18582, opened 2020-04-12
Codemod it.experimental to gate pragma

acdlite:
Codemods all uses of  `it.experimental` to the gate pragma added in #18581.

[**Quick review link**](https://github.com/facebook/react/compare/master...acdlite:remove-it-experimental?w=1)


acdlite:
It's annoying how many whitespace changes there are, because the line length went down. But this illustrates the advantage of the pragma form: adding or removing a gate doesn't affect the whitespace, because the pragma goes on a separate line.

PR #18585, opened 2020-04-13
Migrate conditional tests to gate pragma

acdlite:
## Based on #18582

I split this up from the previous PR because the changes in here weren't completely automated. I had to manually edit stuff.

I searched through the codebase for this pattern:

```js
describe('test suite', () => {
  if (!__EXPERIMENTAL__) { // or some other condition
    test("empty test so Jest doesn't complain", () => {});
    return;
  }

  // Unless we're in experimental mode, none of the tests in this block
  // will run.
})
```

and converted them to the `@gate` pragma instead.

The reason this pattern isn't preferred is because you end up disabling more tests than you need to.

[**Quick review link**](https://github.com/acdlite/react/compare/remove-it-experimental...migrate-conditional-tests?w=1)


packages/react/src/__tests__/ReactDOMTracing-test.internal.js
```diff
@@ -716,11 +717,10 @@ describe('ReactDOMTracing', () => {
         expect(
           onInteractionScheduledWorkCompleted,
         ).toHaveBeenLastNotifiedOfInteraction(interaction);
-
-        done();
       });
 
-      it('traces interaction across client-rendered hydration', async done => {
+      // @gate experimental
+      it('traces interaction across client-rendered hydration', async () => {
```
bvaughn, line 723: nit: I don't think this test needs to be async.

packages/react/src/__tests__/ReactDOMTracing-test.internal.js
```diff
@@ -599,7 +600,8 @@ describe('ReactDOMTracing', () => {
     });
 
     describe('hydration', () => {
-      it('traces interaction across hydration', async done => {
+      // @gate experimental
+      it('traces interaction across hydration', async () => {
```
bvaughn, line 604: nit: I don't think this test needs to be async.

PR #18586, opened 2020-04-13
DevTools: Switch between "Rendered At" renders using keyboard arrow keys

### LINKED ISSUE #16473: DevTools: Switch between "Rendered At" renders using keyboard arrow keys
My favorite devtools feature, "Why did this render?", has a remaining problem: it's too hard to flip through a long list of renders in the Rendered At list to find an interesting render. I know I can use the mouse (or Tab and then Space) to move to the next render, but that's painful if there are 50+ renders to work through.

Suggestion: when an item in the Rendered At list has focus, then up/down arrows should change the selected render (and scroll the pane up/down if the selected render isn't visible). 

This would be helpful for accessibility too.

I'll file a separate issue for keyboard support for the flamegraph view. Keyboard access to that view will be much harder to implement so it makes sense to do it separately.

---
Originally reported by @justingrant via https://github.com/bvaughn/react-devtools-experimental/issues/332

ChrisDobby:
## Summary

Fixes #16473 
Up and down arrow keys change the selected render in the sidebar

## Test Plan

keyboard-nav

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
@@ -8,7 +8,7 @@
  */
 
 import * as React from 'react';
-import {Fragment, useContext, useCallback, useEffect, useRef} from 'react';
+import {Fragment, useContext, useEffect, useRef} from 'react';
 import WhatChanged from './WhatChanged';
 import {ProfilerContext} from './ProfilerContext';
 import {formatDuration, formatTime} from './utils';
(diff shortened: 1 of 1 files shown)


packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
```diff
@@ -31,12 +31,68 @@ export default function SidebarSelectedFiberInfo(_: Props) {
...
+    event => {
+      switch (event.key) {
+        case 'ArrowUp':
+          if (selectedCommitIndex !== null) {
+            selectCommitIndex(
+              selectedCommitIndex > 0
+                ? selectedCommitIndex - 1
+                : commitIndices.length - 1,
+            );
+          }
```
bvaughn, line 52: Oooh, this isn't right. I didn't catch it during my first review, but the selected component might not be in every commit. We'll need to filter the list.

packages/react-devtools-shared/src/devtools/views/Profiler/SidebarSelectedFiberInfo.js
```diff
@@ -75,7 +112,7 @@ export default function SidebarSelectedFiberInfo(_: Props) {
           <ButtonIcon type="close" />
         </Button>
       </div>
-      <div className={styles.Content}>
+      <div className={styles.Content} onKeyDown={handleKeyDown}>
```
bvaughn, line 115: Looks like this event isn't firing for Firefox. I think that's because of a missing `tabIndex={0}`

justingrant:
This rocks. Thanks @ChrisDobby and @bvaughn !

PR #18596, opened 2020-04-14
Delete flushSuspenseFallbacksInTests flag

acdlite:
This was meant to be a temporary hack to unblock the `act` work, but it quickly spread throughout our tests.

What it's meant to do is force fallbacks to flush inside `act` even in Concurrent Mode. It does this by wrapping the `setTimeout` call in a check to see if it's in an `act` context. If so, it skips the delay and immediately commits the fallback.

Really this is only meant for our internal React tests that need to incrementally render. Nobody outside our team (and Relay) needs to do that, yet. Even if/when we do support that, it may or may not be with the same `flushAndYield` pattern we use internally.

However, even for our internal purposes, the behavior isn't right because a really common reason we flush work incrementally is to make assertions on the "suspended" state, before the fallback has committed. There's no way to do that from inside `act` with the behavior of this flag, because it causes the fallback to immediately commit. This has led us to *not* use `act` in a lot of our tests, or to write code that doesn't match what would actually happen in a real environment.

What we really want is for the fallbacks to be flushed at the *end* of the `act` scope. Not within it.

This only affects the noop and test renderer versions of `act`, which are implemented inside the reconciler. Whereas `ReactTestUtils.act` is implemented in "userspace" for backwards compatibility. This is fine because we didn't have any DOM Suspense tests that relied on this flag; they all use test renderer or noop.

In the future, we'll probably want to move always use the reconciler implementation of `act`. It will not affect the prod bundle, because we currently only plan to support `act` in dev. Though we still haven't completely figured that out. However, regardless of whether we support a production `act` for users, we'll still need to write internal React tests in production mode. For that use case, we'll likely add our own internal version of `act` that assumes a mock Scheduler and might rely on hacks that don't 100% align up with the public one.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -7,7 +7,7 @@
  * @flow
  */
 
-import type {Wakeable} from 'shared/ReactTypes';
+import type {Thenable, Wakeable} from 'shared/ReactTypes';
 import type {Fiber, FiberRoot} from './ReactInternalTypes';
 import type {ExpirationTime} from './ReactFiberExpirationTime.new';
 import type {ReactPriorityLevel} from './ReactInternalTypes';
@@ -179,6 +179,9 @@ import {
   clearCaughtError,
 } from 'shared/ReactErrorUtils';
 import {onCommitRoot} from './ReactFiberDevToolsHook.new';
+
+// Used by `act`
+import enqueueTask from 'shared/enqueueTask';
 
 const ceil = Math.ceil;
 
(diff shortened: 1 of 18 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -3197,3 +3187,207 @@ function finishPendingInteractions(root, committedExpirationTime) {
...
+      let didFlushWork = false;
+      while (flushPassiveEffects()) {
+        didFlushWork = true;
+      }
+      return didFlushWork;
+    } finally {
+      isFlushingAct = prevIsFlushing;
+    }
+  }
+}
```
acdlite, line 3228: This is the relevant change. I set `isFlushingAct` while the work is being flushed at the end. Then I check this where we used to check the "force fallback" feature flag.

acdlite:
@lunaruan One is built into the reconciler and relies on some internals. That's the implementation that's used by React Test Renderer and React Noop.

There's another one implemented inside ReactTestUtils. It has to account for backwards compatibility with older versions of React. So it has lots of fallback mechanisms to try to work as best as possible for existing users. It's really only used to test components that include `useEffect`, because those effects are "passive" — unlike everything else in React's legacy mode, they are scheduled in an async task. Tests need to flush those tasks before making assertions. So that's what `act` does: at the end of the `act` body, it flushes all the passive effects.

In Concurrent Mode, it goes further and flushes *all* pending React tasks, including rendering and suspended fallbacks. That can't be properly implemented without relying on internals. So they diverge slightly. Long term plan is to move everyone to the built-in implementation (the one I changed in this PR).

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -881,11 +872,7 @@ function finishConcurrentRender(
       // The work completed. Ready to commit.
       if (
         // do not delay if we're inside an act() scope
-        !(
-          __DEV__ &&
-          flushSuspenseFallbacksInTests &&
-          IsThisRendererActing.current
```
acdlite, line 884: Ah I got this slightly wrong. Fixing.

PR #18605, opened 2020-04-14
Adds clarity to test renderer expectOne error

iturgeon:
## Summary

For newcomers poking in the dark using ReactTestRenderer, the various find
functions are used without realizing they will throw an error when they
find anytying but a single match.

The previous error of `No instances found with node type: "xxx"`
is not as clear about expectations as when expectOne finds more than
one: `Expected 1 but found 2 instances with node type "xxx"`

This change simplifies the code by just using the second version and
adds clarity for people unfamiliar with RTR's find functions

The new version: `Expected 1 but found 0 instances with node type "xxx"`

This inherently tells the user how many are expected and how many are found.
At the same time it provides contextual clues that the previous version lacked.

### DIFF
--- a/packages/react-test-renderer/src/ReactTestRenderer.js
+++ b/packages/react-test-renderer/src/ReactTestRenderer.js
@@ -417,10 +417,7 @@ function expectOne(
     return all[0];
   }
 
-  const prefix =
-    all.length === 0
-      ? 'No instances found '
-      : `Expected 1 but found ${all.length} instances `;
+  const prefix = `Expected 1 but found ${all.length} instances `;
 
   throw new Error(prefix + message);
 }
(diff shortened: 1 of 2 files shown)


rickhanlonii:
Hey @iturgeon, thanks for working to improve the test assertions.

In this case I think that 

> No instances found

Is better than 

> Expected 1 but found 0 instances

This is similar to how of the Jest matchers work which change the language to try to be more clear in the case of `.not` conditions or counts like this. I'm going to close this but thanks for the feedback!

iturgeon:
@rickhanlonii I can see your point. I'll raise one last point before it dies, though. Perhaps implied in the pr, but not clearly articulated yet.

This came up after a coworker, who was moving back and forth between enzyme and RTR. They wrote a complete test with several asserts etc, ran it and got an error 'no instances found'. The error itself gave very little in the form of context clues to associate the with the test code they just wrote.

In contrast, the error with the count information also prints what instance type it was looking for. The instance name will often be easily visible in the test you just wrote as an argument to expectOne.

I took the simplicity route by combining the branches in output, but would still suggest the 'no instances found' branch would serve the test writer better if it left some breadcrumbs back to their code in the form of printing the instance name.

rickhanlonii:
Maybe I'm missing something, the existing error has the instance name:

> No instances found with node type: "Unknown"

PR #18607, opened 2020-04-14
Experimental test selector API

bvaughn:
Adds several new experimental APIs to aid with automated testing.

This is not an RFC, although the PR includes some basic documentation in a similar format.

## Types of selectors

Each of the methods below accepts an array of "selectors" that identifies a path (or paths) through a React tree. There are four basic selector types:
* **Component**: Matches `Fibers` with the specified React component type
* **Role**: Matches Host Instances matching the (explicit or implicit) accessibility role.
* **Test name**: Matches Host Instances with a `data-testname` attribute.
* **Text**: Matches Host Instances that **directly** contain the specified text.

There is also a special *lookahead* selector type that enables further matching within a path (without actually including the path in the result). This selector type was inspired by the [`:has()` CSS pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:has). It enables e.g. matching a `<section>` that contained a specific header text, then finding a like button within that `<section>`.

### `findAllNodes()`

Finds all Host Instances (e.g. `HTMLElement`) within a host subtree that match the specified selector criteria.

```js
function findAllNodes(
  hostRoot: Instance,
  selector: Array<Selector>
): Array<Instance>
```

#### How does it work?

* This method starts searching at the specified root, which must be either:
    * A host parent node above the React container (e.g. `document.body`) in which case, React will traverse the native tree until it finds the first React container, or...
    * A React container, or...
    * A React-rendered Host Instance with a `data-testname` attribute (e.g. a match from an earlier `findAllNodes` query).
* Starting from the root (or React container) React traverses the internal `Fiber` tree ¹ looking for all paths matching the specified selector.
    * While traversing the tree, each node will be compared to the current selector index.
        * If the criteria is satisfied, the selector index is advanced.
        * Otherwise the tree traversal continues.
    * Once all selector criteria have been satisfied, the remaining subtree is searched for the topmost Host Instances (might be several for Fragments).

¹ Traversing the `Fiber` tree (instead of the host tree) provides several benefits:
* Easier to implement and share cross-renderer logic.
* Portals (e.g. tooltips) can be matched even if they are in a different host tree.

#### What does it return?

* An array of the topmost Host Instances (e.g. `HTMLElement`s) that match the specified criteria. (This array will be empty if no matches are found.)

## `getFindAllNodesFailureDescription()`

Returns an error string describing the matched and unmatched portions of the selector query.

```js
function getFindAllNodesFailureDescription(
  hostRoot: Instance,
  selector: Array<Selector>
): string | null
```

#### How does it work?

* This methods processes selectors in the same way as findAllNodes.

#### What does it return?

* If the full selector query is able to be matched, this method returns null.
* Otherwise, it returns a string describing how far it was able to match. For example:
```
findAllNodes was able to match part of the selector:
  <App> > <Navigation>

No matching component was found for:
  <UserProfiler> > [data-testname="edit"]
```

### `findBoundingRects()`

For all React components within a host subtree that match the specified selector criteria, return a set of bounding boxes that covers the bounds of the nearest (shallowed) Host Instances within those trees.

```js
function findBoundingRects(
  hostRoot: Instance,
  selector: Array<Selector>
): Array<Rect>
```

#### How does it work?

* This methods processes selectors in the same way as `findAllNodes`.
* Within each matching path:
    * Find the first Host Instances in the matched component. (Might be several for Fragments.)
    * For each Host Instance, call a `getBoundingClientRect` Host Config method. Add the result to a list of matched boxes.
    * Optionally merge adjacent boxes and exclude boxes that are fully covered by other boxes. (This ensure less reliance on implementation details like how many nodes make up the tree.)
* Return the list of matches bounds.

#### What does it return?

* An array of bounding boxes relative to the viewport. E.g. `{ x: number, y: number, width: number, height: number }`
* (This array will be empty if no matches are found.) 

#### What can this API be used for?

* Clicking or hovering over the center of a Component.
* Take a snapshot of the Component for screen shot tests.
* Assert on the maximum size a component is expected to take up.
* Scrolling the Component into view.

### `observeVisibleRects()`

For all React components within a host subtree that match the specified selector criteria, observe if it’s bounding rect is visible in the viewport and is not occluded.

```js
function observeVisibleRects(
  hostRoot: Instance,
  selector: Array<Selector>,
  callback: (intersectingRects: Array<{ ratio: number, rect: Rect }>) => void,
  options: IntersectionObserverOptions,
): { disconnect(): void }
```

How does it work?

* Create a new [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/) or equivalent `HostConfig` specific API, passing a callback and options argument to it.
    *  Note: Host Config implementation needs to convert each entry into an object containing only `{ ratio, rect }`, otherwise this callback works the same as the DOM version.
* This methods processes selectors in the same way as `findAllNodes`.
* Within each matching path:
    * Find the first Host Instances in the matched component. (Might be several for Fragments.)
    * For each Host Instance, call the `observe(intersectionObserver, instance)` Host Config method to add it to the list to be observed.
* Attach an internal (for test builds only) callback to the React reconciler that is called every time React commits a new update.
    * When this callback is invoked, disconnect the `Observer` and start over from the top to attach a new one. This makes this observation live updating. This prevents accidentally relying on implementation details such as if a DOM node is reused or remounted.

#### What does it return?

* An object containing a `disconnect()` method. When this method is called, we disconnect the `IntersectionObserver` and disconnect the callback in the React reconciler.

#### What can this API be used for?

* Waiting for a Component to become visible before doing further operations.
* Asserting that a Component is already visible.

#### Why is this API necessary?

* Ideally `findBoundingRects` would be enough for this but these APIs don’t give access to where in the DOM this component is, nor really all the other things that might obscuring the bounding rect. This lets us solve this use case without exposing more implementation details and we can build in details such as if it’s obscured by something virtual (e.g. if an ART component is covered by another ART component within a canvas).

### focusWithin()

For all React components within a host subtree that match the specified selector criteria, set focus within the first focusable Host Instance (as if you started before this component in the tree and moved focus forwards one step).

```js
function focusWithin(
  hostRoot: Instance,
  selector: Array<Selector>
): boolean
```

The internal structure of a node is an implementation detail. However, you can start from the outside of a component and move forward (e.g. hit tab) to focus within a component. This has behavior that is defined, and so it can be thought of as part of the public API fo the component.

With the Focus Selectors API used for a11y we might be able to even expose more specific capabilities for the outside of a component move focus into a child.

#### How does it work?

* This methods processes selectors in the same way as `findAllNodes`.
* Within each matching path:
    * Find the first Host Instances in the matched component. (Might be several for Fragments.)
    * For each Host Instance:
        * Focus the Host Instance IIF it is considered Focusable by the Host Config.
        * Return true once we’ve found a focusable Host Instance.
* If we didn’t find a focusable Host Instance, return false.

#### What does it return?

* A boolean indicating if something focusable was found and focus was set on it.

#### What can this API be used for?

* Setting focus on the canonical part of a Component such as text input.

## Usage examples

Coming soon. (For now, refer to unit tests.)

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
@@ -14,7 +14,7 @@ describe('ReactDOMTestSelectors', () => {
 
   let act;
   let createComponentSelector;
-  //let createHasPsuedoClassSelector;
+  let createHasPsuedoClassSelector;
   let createRoleSelector;
   let createTextSelector;
   let createTestNameSelector;
@@ -35,7 +35,7 @@ describe('ReactDOMTestSelectors', () => {
     const ReactDOM = require('react-dom/testing');
     act = ReactDOM.act;
     createComponentSelector = ReactDOM.createComponentSelector;
-    //createHasPsuedoClassSelector = ReactDOM.createHasPsuedoClassSelector;
+    createHasPsuedoClassSelector = ReactDOM.createHasPsuedoClassSelector;
     createRoleSelector = ReactDOM.createRoleSelector;
     createTextSelector = ReactDOM.createTextSelector;
     createTestNameSelector = ReactDOM.createTestNameSelector;
@@ -400,6 +400,41 @@ describe('ReactDOMTestSelectors', () => {
       const matches = findAllNodes(document.body, [
         createComponentSelector(Example),
         createRoleSelector('checkbox'),
+      ]);
+      expect(matches).toHaveLength(1);
+      expect(matches[0].id).toBe('match');
+    });
+
+    it('should support searching ahead with the has() selector', () => {
+      function Example() {
+        return (
+          <div>
+            <article>
+              <h1>Should match</h1>
+              <p>
+                <button id="match">Like</button>
+              </p>
+            </article>
+            <article>
+              <h1>Should not match</h1>
+              <p>
+                <button>Like</button>
+              </p>
+            </article>
+          </div>
+        );
+      }
+
+      render(<Example />, container);
+
+      const matches = findAllNodes(document.body, [
+        createComponentSelector(Example),
+        createRoleSelector('article'),
+        createHasPsuedoClassSelector([
+          createRoleSelector('heading'),
+          createTextSelector('Should match'),
+        ]),
+        createRoleSelector('button'),
       ]);
       expect(matches).toHaveLength(1);
       expect(matches[0].id).toBe('match');
--- a/packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
@@ -937,9 +937,9 @@ No matching component was found for:
       expect(didFocus).toBe(false);
     });
 
-    it('should return false if the only focusable elements are visible', () => {
+    it('should return false if the only focusable elements are hidden', () => {
       function Example() {
-        return <button style={{display: 'none'}}>not clickable</button>;
+        return <button hidden={true}>not clickable</button>;
       }
 
       render(<Example />, container);
@@ -968,7 +968,7 @@ No matching component was found for:
       }
       function FirstChild() {
         return (
-          <button style={{display: 'none'}} onFocus={handleFirstFocus}>
+          <button hidden={true} onFocus={handleFirstFocus}>
             not clickable
           </button>
         );
--- a/packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
@@ -1,0 +1,1186 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @emails react-core
+ */
+
+'use strict';
+
+let React;
+let ReactDOM;
+
+describe('ReactDOMTestSelectors', () => {
+  let container;
+
+  beforeEach(() => {
+    jest.resetModules();
+
+    React = require('react');
+    ReactDOM = require('react-dom/testing');
+
+    container = document.createElement('div');
+    document.body.appendChild(container);
+  });
+
+  afterEach(() => {
+    document.body.removeChild(container);
+  });
+
+  describe('findAllNodes', () => {
+    it('should support searching from the document root', () => {
+      function Example() {
+        return (
+          <div>
+            <div data-testname="match" id="match" />
+          </div>
+        );
+      }
+
+      ReactDOM.render(<Example />, container);
+
+      const matches = ReactDOM.findAllNodes(document.body, [Example], 'match');
+      expect(matches).toHaveLength(1);
+      expect(matches[0].id).toBe('match');
+    });
+
+    it('should support searching from the container', () => {
+      function Example() {
+        return (
+          <div>
+            <div data-testname="match" id="match" />
+          </div>
+        );
+      }
+
+      ReactDOM.render(<Example />, container);
+
+      const matches = ReactDOM.findAllNodes(container, [Example], 'match');
+      expect(matches).toHaveLength(1);
+      expect(matches[0].id).toBe('match');
+    });
+
+    it('should support searching from a previous match', () => {
+      function Outer() {
+        return (
+          <div data-testname="outer" id="outer">
+            <Inner />
+          </div>
+        );
+      }
+
+      function Inner() {
+        return <div data-testname="inner" id="inner" />;
+      }
+
+      ReactDOM.render(<Outer />, container);
+
+      let matches = ReactDOM.findAllNodes(container, [Outer], 'outer');
+      expect(matches).toHaveLength(1);
+      expect(matches[0].id).toBe('outer');
+
+      matches = ReactDOM.findAllNodes(matches[0], [Inner], 'inner');
+      expect(matches).toHaveLength(1);
+      expect(matches[0].id).toBe('inner');
+    });
+
+    it('should support an multiple component types in the selector array', () => {
+      function Outer() {
+        return (
+          <>
+            <div data-testname="match" id="match1" />
+            <Middle />
+          </>
+        );
+      }
+      function Middle() {
+        return (
+          <>
+            <div data-testname="match" id="match2" />
+            <Inner />
+          </>
+        );
+      }
+      function Inner() {
+        return (
+          <>
+            <div data-testname="match" id="match3" />
+          </>
+        );
+      }
+
+      ReactDOM.render(<Outer />, container);
+
+      let matches = ReactDOM.findAllNodes(
+        document.body,
+        [Outer, Middle],
+        'match',
+      );
+      expect(matches).toHaveLength(2);
+      expect(matches.map(m => m.id).sort()).toEqual(['match2', 'match3']);
+
+      matches = ReactDOM.findAllNodes(
+        document.body,
+        [Outer, Middle, Inner],
+        'match',
+      );
+      expect(matches).toHaveLength(1);
+      expect(matches[0].id).toBe('match3');
+
+      matches = ReactDOM.findAllNodes(document.body, [Outer, Inner], 'match');
+      expect(matches).toHaveLength(1);
+      expect(matches[0].id).toBe('match3');
+    });
+
+    it('should support an empty component type selector array', () => {
+      function Example() {
+        return (
+          <div>
+            <div data-testname="match" id="match" />
+          </div>
+        );
+      }
+
+      ReactDOM.render(<Example />, container);
+
+      const matches = ReactDOM.findAllNodes(document.body, [], 'match');
+      expect(matches).toHaveLength(1);
+      expect(matches[0].id).toBe('match');
+    });
+
+    it('should find multiple matches', () => {
+      function Example1() {
+        return (
+          <div>
+            <div data-testname="match" id="match1" />
+          </div>
+        );
+      }
+
+      function Example2() {
+        return (
+          <div>
+            <div data-testname="match" id="match2" />
+            <div data-testname="match" id="match3" />
+          </div>
+        );
+      }
+
+      ReactDOM.render(
+        <>
+          <Example1 />
+          <Example2 />
+        </>,
+        container,
+      );
+
+      const matches = ReactDOM.findAllNodes(document.body, [], 'match');
+      expect(matches).toHaveLength(3);
+      expect(matches.map(m => m.id).sort()).toEqual([
+        'match1',
+        'match2',
+        'match3',
+      ]);
+    });
+
+    it('should find nested matches', () => {
+      function Example() {
+        return (
+          <div data-testname="match" id="match1">
+            <div data-testname="match" id="match2" />
+          </div>
+        );
+      }
+
+      ReactDOM.render(<Example />, container);
+
+      const matches = ReactDOM.findAllNodes(document.body, [Example], 'match');
+      expect(matches).toHaveLength(2);
+      expect(matches.map(m => m.id).sort()).toEqual(['match1', 'match2']);
+    });
+
+    it('should enforce the specific order of components types', () => {
+      function Outer() {
+        return (
+          <>
+            <div data-testname="match" id="match1" />
+            <Inner />
+          </>
+        );
+      }
+      function Inner() {
+        return <div data-testname="match" id="match1" />;
+      }
+
+      ReactDOM.render(<Outer />, container);
+
+      expect(
+        ReactDOM.findAllNodes(document.body, [Inner, Outer], 'match'),
+      ).toHaveLength(0);
+    });
+
+    it('should only search the subtree within the component types selector', () => {
+      function Outer() {
+        return (
+          <>
+            <div data-testname="match" id="match1" />
+            <Inner />
+          </>
+        );
+      }
+      function Inner() {
+        return <div />;
+      }
+
+      ReactDOM.render(<Outer />, container);
+
+      expect(
+        ReactDOM.findAllNodes(document.body, [Outer, Inner], 'match'),
+      ).toHaveLength(0);
+
+      expect(
+        ReactDOM.findAllNodes(document.body, [Inner], 'match'),
+      ).toHaveLength(0);
+    });
+
+    it('should not search within hidden subtrees', () => {
+      const ref1 = React.createRef(null);
+      const ref2 = React.createRef(null);
+
+      function Outer() {
+        return (
+          <>
+            <Inner />
+            <div hidden={true}>
+              <div ref={ref1} data-testname="match" />
+            </div>
+          </>
+        );
+      }
+      function Inner() {
+        return <div ref={ref2} data-testname="match" />;
+      }
+
+      ReactDOM.render(<Outer />, container);
+
+      const matches = ReactDOM.findAllNodes(document.body, [Outer], 'match');
+
+      expect(matches).toHaveLength(1);
+      expect(matches[0]).toBe(ref2.current);
+    });
+
+    it('should throw if no container can be found', () => {
+      expect(() => ReactDOM.findAllNodes(document.body, [], 'match')).toThrow(
+        'Could not find React container within specified host subtree.',
+      );
+    });
+
+    it('should throw if an invalid host root is specified', () => {
+      const ref = React.createRef();
+      function Example() {
+        return <div ref={ref} />;
+      }
+
+      ReactDOM.render(<Example />, container);
+
+      expect(() => ReactDOM.findAllNodes(ref.current, [], 'match')).toThrow(
+        'Invalid host root specified. Should be either a React container or a node previously returned by findAllNodes().',
+      );
+    });
+  });
+
+  describe('getFindAllNodesFailureDescription', () => {
+    it('should describe findAllNodes failures caused by the component type selector', () => {
+      function Outer() {
+        return <Middle />;
+      }
+      function Middle() {
+        return <div />;
+      }
+      function NotRendered() {
+        return <div data-testname="match" />;
+      }
+
+      ReactDOM.render(<Outer />, container);
+
+      const description = ReactDOM.getFindAllNodesFailureDescription(
+        document.body,
+        [Outer, Middle, NotRendered],
+        'match',
+      );
+
+      expect(description).toEqual(
+        `findAllNodes was able to match part of the selector:
+  Outer > Middle
+
+No matching component was found for:
+  NotRendered`,
+      );
+    });
+
+    it('should describe findAllNodes failures caused by a failed test-name match', () => {
+      function Example() {
+        return (
+          <>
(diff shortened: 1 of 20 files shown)


packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1271,3 +1280,183 @@ export function validateEventListenerTarget(
...
+
+export const supportsTestSelectors = true;
+
+export function findRootFiber(node: Instance): null | Fiber {
+  const stack = [node];
+  while (stack.length > 0) {
+    const current = stack.shift();
+    if (isContainerMarkedAsRoot(current)) {
+      const root = ((getInstanceFromNodeDOMTree(current): any): Fiber);
+      return root.stateNode.current;
```
acdlite, line 1292: We should try not to let the Fiber type leak into the host config. We already do in other places, unfortunately. Could this return the Container type (`root.stateNode`) instead?

bvaughn: Yeah still thinking through the right API for the host config methods (state node vs fiber). Easy to change :)

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1271,3 +1280,183 @@ export function validateEventListenerTarget(
   }
   return false;
 }
+
+export const supportsTestSelectors = true;
+
+export function findRootFiber(node: Instance): null | Fiber {
+  const stack = [node];
+  while (stack.length > 0) {
+    const current = stack.shift();
```
acdlite, line 1289: Nit: Increment an index (O(1)) instead of `shift`-ing (O(n) in the worst case)

bvaughn: Sure :smile: This PR isn't really ready for review btw.

acdlite: Haha missed that! Ok I'll stop :D

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -19,11 +25,14 @@ import type {
 } from '../shared/ReactDOMTypes';
 import type {ReactScopeMethods} from 'shared/ReactTypes';
 
+import {getRole} from 'dom-accessibility-api';
```
gaearon, line 28: I don't have high hopes that this would be guaranteed to be tree shaken for non-testing builds. This also has a bunch of compiled Babel junk like `sliced` helpers. The entry point also contains another import which isn't used here. I think ideally we would have an inlined implementation.

packages/react-dom/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "homepage": "https://reactjs.org/",
   "dependencies": {
+    "dom-accessibility-api": "^0.4.3",
```
gaearon, line 20: We've been trying to avoid adding more runtime deps. Since that makes the future move to ESM harder to coordinate. See also my other comment.

packages/react-dom/src/client/ReactDOMHostConfig.js
bvaughn: Yeah I understand. I thought about inlining it to begin with but decided I'd rather just make progress initially on the API and the tests, and then swap this out later on :smile:

eps1lon: Yeah this is transpiled pretty far down to support IE 11. The `getRole` helper was originally intended to be a temporary drop-in replacement for aria-query which had some bugs that are now resolved. Ideally aria-query should be favored in the long run since it is used throughout the React ecosystem. 

Having another fork for implicit roles should be avoided IMO. Inconsistencies in those can be pretty frustrating.

Regarding ESM: This should be usable as an ES module. I have extra tests that verify this for node 13.

bvaughn: I looked into using `aria-query` initially (since it seems like that's the main thing RTL is using) but the documentation was pretty non-existent, there weren't really any tests to refer to, and I wasn't sure how to properly use it.

bvaughn:
Okay, ready for a first pass of review and discussion I think.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1244,3 +1254,173 @@ export function validateEventListenerTarget(
...
+    //
+    // Despite this, it seems like some browsers (e.g. Chrome, Firefox) return -1 even for elements
+    // that don't accept focus, like HTMLImageElement or the outermost HTMLElement tag.
+    // I think this method should (at least for now) only concern itself with "sequentially focusable" elements.
+    // https://html.spec.whatwg.org/multipage/interaction.html#sequentially-focusable
+    return false;
+  }
+
+  if (element.offsetWidth === 0 || element.offsetHeight === 0) {
+    // Hidden items can't be focused.
```
eps1lon, line 1346: Do you mean they shouldn't be focuseable or they can't? Because they [get focused when calling focus() and are in tab order](https://codesandbox.io/s/tender-breeze-m9k1c?file=/src/App.js).

This would also mean that this wouldn't be usable in JSDOM in which [these values are always 0](https://github.com/jsdom/jsdom/issues/135) because it has no layout engine

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1244,3 +1254,173 @@ export function validateEventListenerTarget(
...
+  if (((node: any): HTMLInputElement).disabled === true) {
+    // Disabled inputs can't be focused.
+    return false;
+  }
+
+  const element = ((node: any): HTMLElement);
+
+  if (element.tabIndex === null || element.tabIndex < 0) {
+    // The HTML spec says that negative tab index values indicate an element should be,
+    // "click focusable but not sequentially focusable".
```
eps1lon, line 1331: Should the function name reflect that? Negative tab indices are also used to programmatically move focus.

packages/react-dom/src/client/ReactDOMHostConfig.js
bvaughn: Shouldn't be. This is simulating what a user would do when clicking on a page.

eps1lon: So it's only concerned with clicking? The description sounded like it cares about tab order and these elements are in tab-order.

bvaughn: Sorry, they significant part of that statement wasn't the "click" but rather- what a user would do to interact with the page. I was trying to differentiate between _what JavaScript can do_ and _what a user does_. (This was a lightweight attempt at identifying non-displayed stuff.)

I'm not sure about the Code Sandbox you linked. I guess maybe the condition could be changed from `||` to `&&`.

eps1lon: > I'm not sure about the Code Sandbox you linked.

You can tab through the page and the button gets focused even if it has `offsetHeight === 0 && offsetWidth === 0`. That's why I'm curious where this condition stems from.

bvaughn: I know you can tab through, I saw the sandbox :smile:

The previous version of the Sandbox you linked did not have an `offsetWidth` of 0 for me (which is why I mentioned maybe `&&` instead of `||`).

eps1lon: Yeah I added it after you mentioned the extra condition. Wasn't meant to trick you :smile:

bvaughn: Closing this loop: I just decided to remove my attempts at detecting if an element is focusable or not and instead just rely fully on the browser to decide what's focusable.

Can revisit this later if it ends up being a problem.

bvaughn:
Pinging potential reviewers :smile:

packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
```diff
@@ -0,0 +1,1427 @@
...
+          <Example2 />
+        </>,
+        container,
+      );
+
+      const matches = findAllNodes(document.body, [
+        createTestNameSelector('match'),
+      ]);
+      expect(matches).toHaveLength(3);
+      expect(matches.map(m => m.id).sort()).toEqual([
```
eps1lon, line 239: This is not necessarily a concern with the implementation but if this lands the documentation should warn that we shouldn't rely on the order since it's an implementation detail of findAllNodes. I guess technically this is the order in the react tree which isn't the same as in the DOM tree because e.g. portals?

packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
```diff
@@ -0,0 +1,1427 @@
...
+
+      render(<Outer />, container);
+
+      const matches = findAllNodes(document.body, [
+        createComponentSelector(Outer),
+        createTestNameSelector('match'),
+      ]);
+
+      expect(matches).toHaveLength(1);
+      expect(matches[0]).toBe(ref2.current);
```
eps1lon, line 315: Is there a specific reason why sometimes you compare against `id` attributes and sometimes against host instances?

packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
```diff
@@ -0,0 +1,1427 @@
...
+
+      const matches = findAllNodes(document.body, [
+        createComponentSelector(Example),
+        createTextSelector('bar'),
+      ]);
+      expect(matches).toHaveLength(1);
+      expect(matches[0].id).toBe('match');
+    });
+
+    it('should support filtering by explicit accessibiliy role', () => {
```
eps1lon, line 340: For `dom-accessibility-role` the first role in the `role` attribute was sufficient but in `@testing-library/dom` we had some requests supporting fallback roles e.g. `<div role="meter progressbar" />`. https://github.com/testing-library/dom-testing-library/issues/411 has some more context in case that is something you're interested in.

bvaughn:
Ping @sebmarkbage / @gaearon / @acdlite again.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1117,3 +1127,142 @@ export function preparePortalMount(portalInstance: Instance): void {
...
+  }
+
+  return false;
+}
+
+export function getTextContent(fiber: Fiber): string | null {
+  switch (fiber.tag) {
+    case HostComponent:
+      let textContent = '';
+      const childNodes = fiber.stateNode.childNodes;
```
trueadm, line 1168: Why can't we do `fiber.stateNode.textContent` here? What does it miss compared to this approach?

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1142,3 +1152,142 @@ export function getInstanceFromScope(
...
+  // but this would not handle the case where application code managed focus to automatically blur.
+  let didFocus = false;
+  const handleFocus = () => {
+    didFocus = true;
+  };
+
+  const element = ((node: any): HTMLElement);
+  try {
+    element.addEventListener('focus', handleFocus);
+    element.focus();
```
trueadm, line 1227: This won't cover the case where you want to focus an SVG element. For that you can do:

```
(element.focus || HTMLElement.prototype.focus).call(element);
```
Another issue is that internally, we don't normally use `node.focus`, instead opting to use our `FocusManager` module, which sets the `tabIndex` to `-1` before calling `.focus()`. We do this because the focus handling logic internally might use a different heuristic from what is normally "browser focusable".

I'm also a bit concerned that this will trigger unwanted side-effects, i.e. calling `focus` would trigger an event listener that updates state or logs something. Which would be possible, right?

packages/react-dom/src/client/ReactDOMHostConfig.js
bvaughn: Great question!

I think that would make the text selector type useless, since it would always match every root (since `textContent` includes all of the DOM subtree). The way this is implemented, we will only return the contents of the *immediate* `TEXT_NODE` children. (There's a test that verifies this behavior.) I think this makes the selector actually useful.

packages/react-dom/src/client/ReactDOMHostConfig.js
bvaughn: > Another issue is that internally, we don't normally use `node.focus`, instead opting to use our `FocusManager` module, which sets the `tabIndex` to `-1` before calling `.focus()`. We do this because the focus handling logic internally might use a different heuristic from what is normally "browser focusable".

We want "normal browser focusable" behavior in this case, because code is for e2e testing. We want it to behave as closely to what the user would get as possible.

> I'm also a bit concerned that this will trigger unwanted side-effects, i.e. calling `focus` would trigger an event listener that updates state or logs something. Which would be possible, right?

This is for e2e testing. If focusing something triggers side effects, we want that behavior. This method won't be calling focus on more than one focusable thing.

trueadm: > We want "normal browser focusable" behavior in this case, because code is for e2e testing. We want it to behave as closely to what the user would get as possible.

Gotcha.

> This is for e2e testing. If focusing something triggers side effects, we want that behavior. This method won't be calling focus on more than one focusable thing.

In which case, we should also maybe provide an argument to `focusWithin` that tells it to scroll or not. This is an important heuristic and one that native `HTMLElement.prototype.focus` takes too (although only in the latest browsers). Given this is likely to be only be used by Chrome, which supports this feature, it sounds like we should also provide it?

bvaughn: > > This is for e2e testing. If focusing something triggers side effects, we want that behavior. This method won't be calling focus on more than one focusable thing.
> 
> In which case, we should also maybe provide an argument to `focusWithin` that tells it to `preventScroll` or not. This is an important heuristic and one that native `HTMLElement.prototype.focus` takes too (although only in the latest browsers). Given this is likely to be only be used by Chrome, which supports this feature, it sounds like we should also provide it?

I'm not sure I understand how one follows from the other. Can you explain it to me?

This API is meant to let a test author *simulate* something the user would do with their mouse and/or keyboard. It's not meant as a replacement for programmatically calling `.focus()`.

trueadm: Now I'm confused. Why do you event want a `focusWithin` API at all then? I thought you wanted to manually invoke `focus` firing in e2e tests? Otherwise, if you're stating you only want to simulate user would do, then you never want to be doing this at all.

At a more fundamental level, there is focus with intent and focus without intent. Focusing with intent is when the user invokes it explicitly (clicking, tabbing, using arrows, VoiceOver etc) and without intent is when the focus is controlled by logic and usually imperatively handled (moving focus upon an animation completing, moving focus when something closes/opens etc).

If the API is meant to simulate what the user is doing, then you don't need this API. Simualting `click`, or `press tab` etc would achieve what you're trying to get here. In the caes you don't want to have focus with intent, then the UI logic should handle that.

bvaughn: > Now I'm confused. Why do you event want a `focusWithin` API at all then?

This API would enable you to write a test that e.g. focuses inside of sign in form and then enters test account authentication details. For example, with Puppeteer that might look like:

```js
focusWithin(document.body, [createComponentSelector(LoginForm)]);

await page.keyboard.type("test-user");
```

> If the API is meant to simulate what the user is doing, then you don't need this API. Simulating `click`, or press `tab` etc would achieve what you're trying to get here.

I think this comment misses the high-level purpose of the selector API: It's to help you write tests that aren't brittle and that don't require custom test-specific data attributes (which have proven to be brittle and difficult to maintain at Facebook scale).

trueadm: > This API would enable you to write a test that e.g. focuses inside of sign in form and then enters test account authentication details. For example, with Puppeteer that might look like:

What field on the form would it select? Wouldn't you instead get the Username input and "click" it?

> Accomplishing this- without using the new React selector path to scope down which part of the page/application you want to interact with- would be difficult 😄

That's how our e2e tests look like in most cases already internally. I'm not sure you need to have this additionaly functionality. Otherwise, you might make tests that are fragile and break easily with small changes to the UI.

bvaughn: > What field on the form would it select? Wouldn't you instead get the Username input and "click" it?

The first focusable one within the matched path, as described in the API docs in the PR description :smile: You probably wouldn't write the *actual exact syntax* I wrote. It was just an example.

> That's how our e2e tests look like in most cases already internally. I'm not sure you need to have this additionaly functionality. Otherwise, you might make tests that are fragile and break easily with small changes to the UI.

Check out the internal doc I sent you in work chat. These new APIs are meant to *fix* several of the problems with our internal tests.

packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
bvaughn: just how I wrote the tests.

packages/react-dom/src/client/ReactDOMHostConfig.js
trueadm: The internal doc makes more sense now. I see some of the things you want to tackle.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -1142,3 +1152,142 @@ export function getInstanceFromScope(
...
+  // Instead we'll just listen for a "focus" event to verify that focus was set.
+  //
+  // We could compare the node to document.activeElement after focus,
+  // but this would not handle the case where application code managed focus to automatically blur.
+  let didFocus = false;
+  const handleFocus = () => {
+    didFocus = true;
+  };
+
+  const element = ((node: any): HTMLElement);
```
trueadm, line 1224: You'll also need to check if the node has `tabIndex` of `-1` before doing this. If the node does, then we should skip it. You can still focus an element that has a `tabIndex` of `-1`.

bvaughn: I disagree. See previous conversation on this PR.

bvaughn: > You can still focus an element that has a tabIndex of -1.

The reason why the `tabIndex` check doesn't feel obviously correct is that it only applies to keyboard focus. Users can still click to focus to focus the thing, so maybe the test selector API *should* focus those too.

trueadm: That's true, although I can see a case where you might want one or the other. So maybe an argument or option can be supplied to `focusWithin` to enable that behavior? The same goes with `aria-hidden`, which is another common way to telling accessibility tooling "please do not focus within my sub-tree".

bvaughn:
Thanks a ton for the review and feedback, Dominic!

packages/react-dom/src/__tests__/ReactDOMTestSelectors-test.internal.js
bvaughn: Cool. Thanks for the pointer. Might as well handle that case too since it's easy enough!

PR #18608, opened 2020-04-14
Add regression tests where sync render causes later concurrent render to expire soon

jddxf:
Currently we only reset `currentEventTime` when we start concurrent render. So `currentEventTime` is retained after sync render. That may make later concurrent render expire soon if enough time has elapsed since the last assignment to `currentEventTime`.

I've thought about resetting `currentEventTime` in `ensureRootIsScheduled` when there's nothing to work on the root. But that breaks the test below https://github.com/facebook/react/blob/96203240d3de3c290f3d6013a8cbcc7b38dc1740/packages/react-reconciler/src/__tests__/ReactExpiration-test.js#L111-L113


gaearon:
This seems to pass now. I'll merge.

PR #18609, opened 2020-04-14
Move beforeblur phase to prepareForCommit

trueadm:
This is a follow up to https://github.com/facebook/react/pull/18564. Specifically, this moves the `beforeblur` logic in `ReactDOMHostConfig`, so that it occurs during `prepareForCommit` – avoiding the issues with conflicting with a partially mutated tree. To do this logic we pass through a helper function to `prepareForCommit` called `isInstanceInsideHiddenOrRemovedTree`. This helps us determine if the "active element" is within a sub-tree that is either hidden or removed.

Furthermore, the effeciency of this approach should be a nice improvement, as it means we have to do far less logic than before – along with a slight code size improvement.

I also applied the `detachFiber` changes from https://github.com/facebook/react/pull/18564 and included the regression test from that PR too.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -94,6 +94,10 @@ import {
 import {getListenerMapForElement} from '../events/DOMEventListenerMap';
 import {TOP_BEFORE_BLUR, TOP_AFTER_BLUR} from '../events/DOMTopLevelEventTypes';
 
+// TODO: This is an exposed internal, we should move this around
+// so this isn't the case.
+import {isFiberInsideHiddenOrRemovedTree} from 'react-reconciler/src/ReactFiberTreeReflection';
+
 export type ReactListenerEvent = ReactDOMListenerEvent;
 export type ReactListenerMap = ReactDOMListenerMap;
 export type ReactListener = ReactDOMListener;
@@ -246,17 +250,14 @@ export function getPublicInstance(instance: Instance): * {
   return instance;
 }
 
-export function prepareForCommit(
-  containerInfo: Container,
-  isInstanceInsideHiddenOrRemovedTree: (instance: Object) => boolean,
-): void {
+export function prepareForCommit(containerInfo: Container): void {
   eventsEnabled = ReactBrowserEventEmitterIsEnabled();
   selectionInformation = getSelectionInformation();
   if (enableDeprecatedFlareAPI || enableUseEventAPI) {
     const focusedElem = selectionInformation.focusedElem;
     if (focusedElem !== null) {
       const instance = getClosestInstanceFromNode(focusedElem);
-      if (instance !== null && isInstanceInsideHiddenOrRemovedTree(instance)) {
+      if (instance !== null && isFiberInsideHiddenOrRemovedTree(instance)) {
         dispatchBeforeDetachedBlur(focusedElem);
       }
     }
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -73,7 +73,6 @@ import {
   enableUseEventAPI,
   enableScopeAPI,
 } from 'shared/ReactFeatureFlags';
-import {HostComponent} from 'react-reconciler/src/ReactWorkTags';
 import {
   RESPONDER_EVENT_SYSTEM,
   IS_PASSIVE,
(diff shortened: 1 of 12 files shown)


packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -247,9 +246,21 @@ export function getPublicInstance(instance: Instance): * {
...
   selectionInformation = getSelectionInformation();
+  if (enableDeprecatedFlareAPI || enableUseEventAPI) {
+    const focusedElem = selectionInformation.focusedElem;
+    if (focusedElem !== null) {
+      const instance = getClosestInstanceFromNode(focusedElem);
+      if (instance !== null && isInstanceInsideHiddenOrRemovedTree(instance)) {
+        dispatchBeforeDetachedBlur(focusedElem);
+      }
+    }
+  }
```
bvaughn, line 263: Do we need to add a `try`/`finally` now (in case the code goes down the path of `DEPRECATED_dispatchEventForResponderEventSystem`)

trueadm: Thankfully we don't need it. User code is wrapped with a `try/catch/finally` block that we use from the legacy event system.

packages/react-reconciler/src/ReactFiberCommitWork.old.js
```diff
@@ -1116,17 +1116,21 @@ function detachFiber(fiber: Fiber) {
...
   fiber.dependencies = null;
-  fiber.alternate = null;
   fiber.firstEffect = null;
   fiber.lastEffect = null;
-  fiber.pendingProps = null;
   fiber.memoizedProps = null;
+  fiber.memoizedState = null;
+  fiber.pendingProps = null;
+  fiber.return = null;
+  fiber.sibling = null;
```
bvaughn, line 1128: This seems like a nice change 👍

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
```diff
@@ -1845,7 +1846,7 @@ function commitRootImpl(root, renderPriorityLevel) {
     // The first phase a "before mutation" phase. We use this phase to read the
     // state of the host tree right before we mutate it. This is where
     // getSnapshotBeforeUpdate is called.
-    prepareForCommit(root.containerInfo);
+    prepareForCommit(root.containerInfo, isFiberInsideHiddenOrRemovedTree);
```
sebmarkbage, line 1849: Does the function get inlined? Passing first class functions around sometimes deopts closure. Although mostly if you use a closure which this isn’t.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -571,20 +575,9 @@ function dispatchAfterDetachedBlur(target: HTMLElement): void {
...
   instance: Instance | TextInstance | SuspenseInstance,
 ): void {
-  if (
-    (enableDeprecatedFlareAPI || enableUseEventAPI) &&
-    selectionInformation &&
-    instance === selectionInformation.focusedElem
-  ) {
-    dispatchBeforeDetachedBlur(((instance: any): HTMLElement));
-  }
   if (enableUseEventAPI) {
```
gaearon, line 581: Is there any way we can get rid of the block below, and thus of this whole host method, now?

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
trueadm: How else can we pass it to the renderer without too much indirection? I intentionally wanted to keep it out so we didn't leak details.

packages/react-dom/src/client/ReactDOMHostConfig.js
trueadm: Unfortunately not, this is separate logic that can't move to the other block.

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
gaearon: I think you can just use `isFiberInsideHiddenOrRemovedTree` directly from the renderer? Kind of not ideal but ReactDOM already does this in a few places.

trueadm: I asked @sebmarkbage about this and he suggested not doing that. Maybe we should here though, just to make it more optimal. I’ll defer for now.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -95,6 +94,10 @@ import {
 import {getListenerMapForElement} from '../events/DOMEventListenerMap';
 import {TOP_BEFORE_BLUR, TOP_AFTER_BLUR} from '../events/DOMTopLevelEventTypes';
 
+// TODO: This is an exposed internal, we should move this around
+// so this isn't the case.
+import {isFiberInsideHiddenOrRemovedTree} from 'react-reconciler/src/ReactFiberTreeReflection';
```
sebmarkbage, line 99: cc @acdlite in case this affects your fork. I don't think it will.

packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -332,3 +332,23 @@ export function findCurrentHostFiberWithNoPortals(parent: Fiber): Fiber | null {
...
+export function isFiberSuspenseAndTimedOut(fiber: Fiber): boolean {
+  const memoizedState = fiber.memoizedState;
+  return (
+    fiber.tag === SuspenseComponent &&
+    memoizedState !== null &&
+    memoizedState.dehydrated === null
+  );
+}
+
+export function isFiberInsideHiddenOrRemovedTree(fiber: Fiber): boolean {
```
sebmarkbage, line 345: Maybe add a note that this is only safe to call in the commit phase when the return tree is consistent. Nowhere else.

PR #18612, opened 2020-04-14
Lanes

acdlite:
## Not for Merge

Stack of changes related to a refactor of React's internal concurrency and scheduling architecture.

I'm using this draft PR so I can push to something and kick off CI checks. It's not for review, but I'll accept feedback regardless.

This is very large set of changes, and many of the incremental steps aren't that impactful except to unblock later ones. I may shuffle and re-arrange things as new blockers emerge.

**When I'm ready for something to be reviewed for merge, I will open separate PRs.**


scripts/codemods/expiration-time-adjustments.js
```diff
@@ -0,0 +1,132 @@
+'use strict';
```
sebmarkbage, line 1: Do we really need to check these in? Isn't this one time?

packages/react-dom/src/events/ReactDOMEventReplaying.js
```diff
@@ -663,7 +663,7 @@ function scheduleCallbackIfUnblocked(
       // Schedule a callback to attempt replaying as many events as are
       // now unblocked. This first might not actually be unblocked yet.
       // We could check it early to avoid scheduling an unnecessary callback.
-      scheduleCallback(NormalPriority, replayUnblockedEvents);
+      scheduleCallback(UserBlockingPriority, replayUnblockedEvents);
```
sebmarkbage, line 666: What's this change? It doesn't seem right. The idea is to be able retry as much as possible and there can be other user blocking events that unblocks that. Retrying too much can be somewhat expensive.

scripts/codemods/expiration-time-adjustments.js
acdlite: I'm not going to merge them, I just put them here in the commit so when I rebase it's already there. Plan is to delete them in a later commit before squashing and merging.

acdlite: > I added the codemods I used to the `scripts` directory. I won't merge
these to master. I'll remove them in a subsequent commit. I'm only
committing them here so they show up in the PR for future reference.

packages/react-dom/src/events/ReactDOMEventReplaying.js
acdlite: I wrote this months ago so might be misremembering the details slightly, but it's so the events get put into the queue ahead of the rendering tasks. It's not related setting the priority context. I can take another look at it, I think when we chatted in person about this a while ago you suggested that it's fine that it doesn't interrupt the current render, similar to  upgrading during hydration. Or something like that.

scripts/codemods/expiration-time-adjustments.js
acdlite: I guess what I'll do delete them in the immediate next commit :)

packages/react-dom/src/events/ReactDOMEventReplaying.js
acdlite: Ended up reverting this. Wish I could remember the exact reason I thought it was necessary and why it's not anymore. Most likely I was just mistaken.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.js
```diff
@@ -1724,15 +1714,24 @@ describe('ReactDOMServerHooks', () => {
 
         ReactDOM.createRoot(container, {hydrate: true}).render(<App />);
 
-        expect(() =>
-          expect(() => Scheduler.unstable_flushAll()).toThrow(
+        if (gate(flags => flags.new)) {
```
fregante, line 1717: It

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.js
```diff
@@ -1724,15 +1714,24 @@ describe('ReactDOMServerHooks', () => {
...
+          expect(() => Scheduler.unstable_flushAll()).toErrorDev([
```
fregante, line 1718: works

gaearon:
This landed via other PRs.

joshribakoff-sm:
Thanks for cleaning up and clarifying Dan, that is very helpful for those of us who may be wanting to follow the history ❤️

PR #18615, opened 2020-04-15
Fork legacy and modern event plugins

trueadm:
This PR splits apart the event system plugins. In a previous PR, they were renamed with a `Legacy` prefix – this PR takes it further and creates a fork of each plugin for the modern event system, with the `Modern` prefix. In `ReactDOMClientInjection` we now use the right set of corresponding plugins depending if we are using the modern event system or the legacy event system.

This should help isolate changes and let us continue work on improving things in the modern event system without affecting how the plugins work with the legacy event system. Furthermore, the accumulation and propagation logic that was recently introduced has been isolated to only work with the modern event system. The previous logic that was removed a few weeks ago has been brought back, but put in the legacy plugin event system module. This means that the legacy event system should work exactly as it used to and the modern event system changes should now be completely isolated from the legacy code paths.

I also forked all the tests for the event plugins so they too have `Legacy` and `Modern` prefixes. The modern event systems enable the feature flag so we are able to possibly catch issues that relate to the event system being used for that test (we run the modern event system with the www flags, but that also includes other www flags that are enabled, which might hide issues).

There are some downsides to this approach:
- we'll need to remember to apply any relevant bugfixes to both sets of event plugins (modern and legacy)
- we'll introduce a few more KB of code size to FB whilst the flag is behind a GK

Both of these downsides are temporary, depending on how quickly we can roll out the modern event system.


trueadm:
This PR seems pretty huge, so I'll instead break up into multiple PRs.

PR #18617, opened 2020-04-15
Fix incorrect unmounted state update warning

bvaughn:
There was a recent report within Facebook of an invalid "_can't perform a React state update on an unmounted component_" warning that was eventually reduced to the fact that the Fiber being warned about was not tracked in the `pendingPassiveHookEffectsUnmount` array, but it's alternate **was**.

Unfortunately, because we detach fibers (which nulls the `.alternate` field) when we commit a deletion, any state updates scheduled between that point and the eventual passive effects flush won't have a way to check if there is a pending passive unmount effect scheduled for the alternate *unless we also explicitly track the alternates*.

~~This fix feels heavy handed and gross to me. Is there a better solution that I'm overlooking?~~ Rather than tracking both sets of Fibers, this PR uses a new DEV-only effect that indicates if the Fiber has a pending passive unmount effect scheduled.

Relates to PR #18096

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -116,7 +116,7 @@ import {
   Snapshot,
   Callback,
   Passive,
-  PassiveUnmountPending,
+  PassiveUnmountPendingDev,
   Incomplete,
   HostEffectMask,
   Hydrating,
@@ -2313,10 +2313,10 @@ export function enqueuePendingPassiveHookEffectUnmount(
     pendingPassiveHookEffectsUnmount.push(effect, fiber);
     if (__DEV__) {
       if (deferPassiveEffectCleanupDuringUnmount) {
-        fiber.effectTag |= PassiveUnmountPending;
+        fiber.effectTag |= PassiveUnmountPendingDev;
         const alternate = fiber.alternate;
         if (alternate !== null) {
-          alternate.effectTag |= PassiveUnmountPending;
+          alternate.effectTag |= PassiveUnmountPendingDev;
         }
       }
     }
@@ -2385,10 +2385,10 @@ function flushPassiveEffectsImpl() {
 
       if (__DEV__) {
         if (deferPassiveEffectCleanupDuringUnmount) {
-          fiber.effectTag &= ~PassiveUnmountPending;
+          fiber.effectTag &= ~PassiveUnmountPendingDev;
           const alternate = fiber.alternate;
           if (alternate !== null) {
-            alternate.effectTag &= ~PassiveUnmountPending;
+            alternate.effectTag &= ~PassiveUnmountPendingDev;
           }
         }
       }
@@ -2872,7 +2872,7 @@ function warnAboutUpdateOnUnmountedFiberInDEV(fiber) {
     ) {
       // If there are pending passive effects unmounts for this Fiber,
       // we can assume that they would have prevented this update.
-      if ((fiber.effectTag & PassiveUnmountPending) !== NoEffect) {
+      if ((fiber.effectTag & PassiveUnmountPendingDev) !== NoEffect) {
         return;
       }
     }
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -116,6 +116,7 @@ import {
   Snapshot,
   Callback,
   Passive,
+  PassiveUnmountPending,
   Incomplete,
   HostEffectMask,
   Hydrating,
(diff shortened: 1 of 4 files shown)


bvaughn:
Cool 👍  What is it?

acdlite:
Instead of pushing the fiber to `pendingPassiveHookEffectsUnmount` and then checking with `indexOf`, you can set a bit on its `effectTag`.

You can also set that bit on the `effectTag` of its current fiber. As a general rule you shouldn't mutate current fibers, but effect tags are fine because it already committed.

bvaughn:
That sounds like a nicer approach, @acdlite. Back to you.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2310,6 +2311,15 @@ export function enqueuePendingPassiveHookEffectUnmount(
 ): void {
   if (runAllPassiveEffectDestroysBeforeCreates) {
     pendingPassiveHookEffectsUnmount.push(effect, fiber);
+    if (__DEV__) {
+      if (deferPassiveEffectCleanupDuringUnmount) {
+        fiber.effectTag |= PassiveUnmountPending;
```
bvaughn, line 2316: Is it weird to only set this bit in DEV mode? It's only needed there.

gaearon: Add Dev suffic to the flag?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2310,6 +2311,15 @@ export function enqueuePendingPassiveHookEffectUnmount(
 ): void {
   if (runAllPassiveEffectDestroysBeforeCreates) {
     pendingPassiveHookEffectsUnmount.push(effect, fiber);
+    if (__DEV__) {
+      if (deferPassiveEffectCleanupDuringUnmount) {
+        fiber.effectTag |= PassiveUnmountPendingDev;
+        const alternate = fiber.alternate;
```
acdlite, line 2317: Nit: Can you pass `current` as an argument through `schedulePassiveEffects`? We try to limit the number of places we access the `.alternate` pointer, since one day we're hoping to remove i t.

bvaughn: Ah, sure, although this is going to require adding a second param to a few other functions (e.g. `commitUnmount`, `commitNestedUnmounts`) since those also call `enqueuePendingPassiveHookEffectUnmount`.

bvaughn: I could do that for `schedulePassiveEffects()` because `commitLifeCycles()` already has both fibers.

But we also call `enqueuePendingPassiveHookEffectUnmount()` from `commitUnmount()` which is called from `commitNestedUnmounts()` which traces all the through to `commitMutationEffects()`, so I'm still going to have to access `alternate` there (and add a bunch of additional params to the functions in between). 

Is this still the right change?

bvaughn: It looks like what we're calling "current" is inconsistent between `commitWork` and `commitDeletion` calls.

For e.g. `commitWork` we pass `nextEffect.alternate` as `current`:
https://github.com/facebook/react/blob/cfefc81ab2f5103bedc9e45701e8c00c0689f499/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L2192-L2193

But for `commitDeletion` we pass `nextEffect` as `current`:
https://github.com/facebook/react/blob/cfefc81ab2f5103bedc9e45701e8c00c0689f499/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L2214

bvaughn: We also call `commitNestedUnmounts` and `commitUnmount` from `unmountHostComponents` while traversing the tree, so we'd need to add a bunch of additional `alternate` access here to pass through that prop.

I think I'm going to *not* make this change for now. Seem reasonable?

acdlite: Right I guess “current” isn’t accurate, it’s just the “other fiber”

Yeah that makes sense. If we got rid of alternates, each fiber would have a shared instance object and we could put the flag on that.

sebmarkbage:
Turns out this is actually going to be what tips us over and would have to require us to add a whole other field since we'd run out of double effect tags. We probably need to revisit to see if we can do something less invasive.

bvaughn:
I'm not sure I fully understand the "whole other field" comment. I think I'm missing some context. Is there more written down somewhere or is it mostly just convos between you and Andrew?

Since this is DEV only, seems like we could use a less optimal, DEV-only strategy for this?

Issue #18618, opened 2020-04-15
Bug: disableRemotePlayback not recognized


tomprats:
Passing `disableRemotePlayback` as an attribute like so:

```
<video disableRemotePlayback={true}>
```

Produces the warning:

> Warning: React does not recognize the `disableRemotePlayback` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `disableremoteplayback` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

It seems like React should recognize it's similar to https://github.com/facebook/react/pull/15334.

React version: 16.13.1

## The current behavior

Warning and not rendering the attribute

## The expected behavior

Adding the attribute similar to how it'd render `disablePictureInPicture` with a final output of:

```
<video disableRemotePlayback ...>
```

When rendering it with:

```
<video disableRemotePlayback={true} ...>
```

tombrowndev:
Create a PR #18619

tombrowndev:
@tomprats This has been merged in. Feel free to close the issue,

Cheers

PR #18619, opened 2020-04-15
ADD: disableRemotePlayback attribute for HTML5 videos

tombrowndev:
This is a fix for issue #18618.

Given this has a wide browser support, I think this should be available.


tombrowndev:
I have now signed the CLA, do I need to remake this PR or will it recognise it?

PR #18624, opened 2020-04-16
Fix wrong command in config.yml

acdlite:
Oops


bvaughn:
Looks like fixing this broke tests 😆 Specifically- my new test!

bvaughn:
I think #18626 should fix it.

PR #18625, opened 2020-04-16
Pressable click fix

nmn:
### Motiation
The custom press tracking has a few bugs:
1. VoiceOver users (and possibly JAWS users) often use a custom keyboard shortcut to trigger clicks, those clicks are currently ignored.
2. On touchscreens, since the click is only ignored on the target itself, if you click on an overlay and it results in the overlay being dismissed, the click then occurs on the element behind the overlay causing strange ghost clicks.

### Fix
These issues are fixed by using the native click event and not using a custom implementation.

## Test Plan
Updated the unit tests, although some of them had to be deleted. Further, did manual testing to ensure that the bugs have been fixed and no new bugs exist. Specially when dealing with nested pressables.

### DIFF
--- a/packages/react-interactions/events/src/dom/PressLegacy.js
+++ b/packages/react-interactions/events/src/dom/PressLegacy.js
@@ -766,22 +766,6 @@ const pressResponderImpl = {
             removeRootEventTypes(context, state);
           }
 
-          // Determine whether to call preventDefault on subsequent native events.
-          if (
-            target !== null &&
-            context.isTargetWithinResponder(target) &&
-            context.isTargetWithinHostComponent(target, 'a')
-          ) {
-            const {
-              altKey,
-              ctrlKey,
-              metaKey,
-              shiftKey,
-            } = (nativeEvent: MouseEvent);
-            // Check "open in new window/tab" and "open context menu" key modifiers
-            const preventDefault = props.preventDefault;
-          }
-
           dispatchPressEndEvents(event, context, props, state);
           state.touchEvent = null;
         } else if (type === 'mouseup') {
--- a/packages/react-interactions/events/src/dom/PressLegacy.js
+++ b/packages/react-interactions/events/src/dom/PressLegacy.js
@@ -751,7 +751,6 @@ const pressResponderImpl = {
       case 'touchend': {
         if (isPressed) {
           const buttons = state.buttons;
-          let isKeyboardEvent = false;
           let touchEvent;
           if (
             type === 'pointerup' &&
@@ -770,7 +769,6 @@ const pressResponderImpl = {
             if (!isValidKeyboardEvent(nativeEvent)) {
               return;
             }
-            isKeyboardEvent = true;
             removeRootEventTypes(context, state);
           } else if (buttons === 4) {
             // Remove the root events here as no 'click' event is dispatched when this 'button' is pressed.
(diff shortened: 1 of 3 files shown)


packages/react-interactions/events/src/dom/PressLegacy.js
```diff
@@ -803,46 +801,7 @@ const pressResponderImpl = {
...
-                updateIsPressWithinResponderRegion(
-                  touchEvent || nativeEvent,
-                  context,
-                  props,
-                  state,
-                );
-              }
-            }
-
-            if (state.isPressWithinResponderRegion && buttons !== 4) {
```
necolas, line 835: You can also delete this state property

packages/react-interactions/events/src/dom/PressLegacy.js
```diff
@@ -803,46 +801,7 @@ const pressResponderImpl = {
...
-              if (
-                pointerType === 'mouse' &&
-                context.isTargetWithinNode(target, pressTarget)
-              ) {
-                state.isPressWithinResponderRegion = true;
-              } else {
-                // If the event target isn't within the press target, check if we're still
-                // within the responder region. The region may have changed if the
-                // element's layout was modified after activation.
-                updateIsPressWithinResponderRegion(
```
necolas, line 826: I think this function's definition can be deleted from the module

packages/react-interactions/events/src/dom/__tests__/PressLegacy-test.internal.js
```diff
@@ -497,21 +494,6 @@ describe.each(environmentTable)('Press responder', hasPointerEvents => {
       expect(onPress).not.toBeCalled();
     });
 
-    // @gate experimental
-    it('is called with modifier keys', () => {
```
necolas, line 501: why was this test deleted?

packages/react-interactions/events/src/dom/PressLegacy.js
nmn: state.isPressWithinResponderRegion is used for tracking isPressed and to fire pressStart, pressEnd and pressMove events, so I think it needs to stay for now.

necolas: Yeah it's doesn't need to be removed to get `onPress` backed by `click`. This is only used to reactivate buttons when a down-pointer moves in and out of the hit rect. We already stopped doing that in [Tap](https://github.com/facebook/react/blob/35867e8094367fdebc4711940eb709454047c2d4/packages/react-interactions/events/src/dom/Tap.js (the intended replacement for Press) because it's work that we can avoid doing. But no need to get involved in that, we can remove this separately.

trueadm:
@nmn I had to revert this PR due to it not properly calling preventDefault, resulting in full page reloads internally when testing it out on the latest www sync.

packages/react-interactions/events/src/dom/PressLegacy.js
```diff
@@ -770,79 +760,13 @@ const pressResponderImpl = {
...
-            const preventDefault = props.preventDefault;
-
-            if (
-              preventDefault !== false &&
-              !shiftKey &&
-              !metaKey &&
-              !ctrlKey &&
-              !altKey
-            ) {
-              state.shouldPreventClick = true;
```
trueadm, line 802: We shouldn't have removed this logic.

PR #18627, opened 2020-04-16
Remove redundant conditional statements

BetaSu:
## Summary

Remove redundant conditional statements

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -824,10 +824,7 @@ export function diffProperties(
         // inserted already.
       }
     } else if (propKey === CHILDREN) {
-      if (
-        lastProp !== nextProp &&
-        (typeof nextProp === 'string' || typeof nextProp === 'number')
-      ) {
+      if (typeof nextProp === 'string' || typeof nextProp === 'number') {
         (updatePayload = updatePayload || []).push(propKey, '' + nextProp);
       }
     } else if (


packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -824,10 +824,7 @@ export function diffProperties(
         // inserted already.
       }
     } else if (propKey === CHILDREN) {
-      if (
-        lastProp !== nextProp &&
```
bvaughn, line 828: This change seems nice, since we already check for this above

https://github.com/facebook/react/blob/c9e59f0a311c7aada51e4a5dd76dc01ea454259c/packages/react-dom/src/client/ReactDOMComponent.js#L767

packages/react-reconciler/src/ReactChildFiber.new.js
```diff
@@ -405,12 +405,11 @@ function ChildReconciler(shouldTrackSideEffects) {
       );
       created.return = returnFiber;
       return created;
-    } else {
```
bvaughn, line 408: This change is subjective and I think should be reverted.

packages/react-reconciler/src/ReactChildFiber.old.js
```diff
@@ -405,12 +405,11 @@ function ChildReconciler(shouldTrackSideEffects) {
       );
       created.return = returnFiber;
       return created;
-    } else {
```
bvaughn, line 408: Ditto.

PR #18630, opened 2020-04-16
refactor import just for more consistent

zhangenming:
.

## Summary
for more consistent


bvaughn:
Thank you for the suggestion! :bow: but I think this change is probably pretty subjective :smile: I think the way it is currently (in master) is okay!

Issue #18589, opened 2020-04-13
Bug: react test utils do not work together with esm module


just-boris:
React version: 16.13.1

## Steps To Reproduce

This code works fine when using plain Node.js, but fails: when [esm module](https://github.com/standard-things/esm) is enabled:

```js
const { act } = require('react-dom/test-utils');

act(async () => {
  console.log('something');
}).then(
  () => console.log('success'),
  () => console.log('error')
);
```

```
$ node test.js
something
success
```

```
$ node -r esm test.js
something
Warning: This browser does not have a MessageChannel implementation, so enqueuing tasks via await act(async () => ...) will fail. Please file an issue at https://github.com/facebook/react/issues if you encounter this warning.
error
```

The warning is caused by an exception happened in this `try/catch` block: https://github.com/facebook/react/blob/72d00ab623502983ebd7ac0756cf2787df109811/packages/shared/enqueueTask.js#L15-L23

This is happening because `esm` module brings its own implementation of the `module` object with slightly different behavior which requires `this` context to be properly preserved.

This one-line change should do the fix:

```diff
-enqueueTaskImpl = nodeRequire('timers').setImmediate;
+enqueueTaskImpl = nodeRequire.call(module, 'timers').setImmediate;
```

This issue has also been reported to React-testing-library, which has similar code: https://github.com/testing-library/react-testing-library/issues/614

**Why should this be a fix in React and not in ESM?**

React relied on undocumented behavior of Node modules loader. It was never guaranteed that it will work without proper `this` context.

In fact, it will not work as expected if you try it with any non built-in module

```js
const nodeRequire = module.require;

nodeRequire('timers'); // built-in module, works
require('react') // proper require, also works
nodeRequire('react'); // Error: Cannot find module 'react', even though the module is installed
```

The current code in React may break in future if `timers` module will be replaced with anything else, or if Node.js changes their internal implementation.

aweary:
Do you want to submit a PR @just-boris? It sounds like a reasonable thing for us to fix.

just-boris:
@aweary, I could, however I am not sure about testing part. What kind of tests would you expect for this?

PR #18632, opened 2020-04-16
Fix requiring timers module with non standard require

just-boris:
# Summary

A fix for #18589 

## Test Plan

Ran "steps to reproduce" from the report. Fails before and passes after the fix.


sebmarkbage:
@gaearon Let's rethink this whole thing. It has grown beyond its original state slowly by slowly until now being hacks on hacks and bloated.

PR #18634, opened 2020-04-16
feat(react-dom): Add enterKeyHint

eps1lon:
## Summary

```diff
-<input enterkeyhint="search" />
+<input enterKeyHint="search" />
```

- [WHATWG spec](https://html.spec.whatwg.org/multipage/interaction.html#attr-enterkeyhint)
- [Chrome Platform Status](https://www.chromestatus.com/feature/5654529808793600)
- supported in chrome 77, safari 13.1, [firefox intents to implement it](https://bugzilla.mozilla.org/show_bug.cgi?id=1490661)
## Test Plan

Ran /fixtures/attribute-behavior locally. Results available under https://drive.google.com/open?id=1j5PJgc-NcKXsGZh8GqBfHXHfdEwd9ii_.

Note: Should I update the fixture for `inputMode` as well? It is still reading from the attribute but the property is [supported as of chrome 66](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode). 

https://github.com/facebook/react/blob/8c75389319d479dc5b2f2cf4b2c7084549913084/fixtures/attribute-behavior/src/attributes.js#L955

Codesandbox using this PR: https://codesandbox.io/s/enterkeyhint-6l5t3

### DIFF
--- a/fixtures/attribute-behavior/AttributeTableSnapshot.md
+++ b/fixtures/attribute-behavior/AttributeTableSnapshot.md
@@ -3223,6 +3223,31 @@
 | `end=(null)`| (initial)| `<null>` |
 | `end=(undefined)`| (initial)| `<null>` |
 
+## `enterKeyHint` (on `<input>` inside `<div>`)
+| Test Case | Flags | Result |
+| --- | --- | --- |
+| `enterKeyHint=(string)`| (initial)| `<empty string>` |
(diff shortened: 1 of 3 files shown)


eps1lon:
deleted by accident

eps1lon:
bump

gaearon:
Looks like we forgot to merge. Thanks!

PR #18636, opened 2020-04-16
Unify Flare FocusWithin responder with useFocusWithin

trueadm:
This PR unifies the Flare `FocusWithin` responder with the recent `useFocusWithin` responder so the Flare version uses a fake `afterblur` rather the fake `blur`. This means we can easily A/B test the various implementations that rely on this internally. I've also added FocusRegion tests that confirm the functionality matches that as we expect internally too.


trueadm:
@necolas The tests were adapted and taken from internal code for `FocusRegion` that fully uses the functions as we do internally – helps cover regressions until we move this out of Github.

necolas:
We have a focus hook that is testing itself against a different user space abstraction that has to be built up in the test. The complexity suggests the divisions might not be right because you can't be sure of the behavior without doing this

trueadm:
I've removed the tests from this PR.

necolas:
I'm not saying whether or not the tests should be part of the PR :) Just that if these sorts of tests are needed to verify the behavior of this hook, then maybe we're missing something about where this functionality belongs

trueadm:
@necolas They're not needed to verify the behavior at all, I just brought them in to verify I didn't make any mistakes. I guess given that we plan on migrating these to FB soon, where they'll get the full test coverage of FocusRegion's tests, it makes sense to leave it till then and not bring them in right now.

PR #18637, opened 2020-04-16
Create copy of Legacy event plugins for Modern system

trueadm:
This PR is a cut out of https://github.com/facebook/react/pull/18615, so we break down the amount of changes in that PR, but have the same end goal.

This PR does four things:

- Creates a copy of the `Legacy` plugin modules with a `Modern` prefix (no content changes at all)
- Links `ReactDOMClientInjection.js` to the corresponding correct set of plugin modules
- Adds a `Legacy` prefix to the plugin module test files.
- Adds a copy of the plugin module test files, but with a `Modern` prefix and an `internal` extension (as a follow up PR will add feature flag mutations for the modern event system).


trueadm:
@bvaughn Yep, it's temporary for now but it gives us a bit more confidence too.

PR #18654, opened 2020-04-17
Fixed minor code issues

mikaelparsekyan:
Fixed inlined return statements warnings.


gaearon:
Thanks for the PR, but this is a matter of code style and in this case intentional.

PR #18655, opened 2020-04-17
Revert focus event PRs

trueadm:
This PR reverts two revent PRs relating to the Flare focus changes:

https://github.com/facebook/react/commit/e2ccbf0358fafc8eb4470a45f5bca1cb97a4bba2
https://github.com/facebook/react/commit/f24a9e70980c2b1d0eb8309eb751315ff39f96bd

This is because we have an e2e test failing on the www sync relating to `beforeblur` and `afterblur`. I'll see if the sync passes with these changes removed. It might also be related to:

https://github.com/facebook/react/commit/bf55ea743496f4fb71df982af0324e322f25046c

Once we have the sync passing, I'll try and dissect the exact problem.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -1355,6 +1355,10 @@ export function listenToEventResponderEventTypes(
       const targetEventType = isPassive
         ? eventType
         : eventType.substring(0, eventType.length - 7);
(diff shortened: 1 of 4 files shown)


trueadm:
I can confirm that this fixes the issue on the www sync.

trueadm:
Actually, I don't think this revert has fixed it. Looks like bf55ea7 is the candidate then :(

PR #18575, opened 2020-04-11
Fix Component Stacks for IE and Native Classes in Safari

sebmarkbage:
This is a follow up to https://github.com/facebook/react/pull/18561

It adds more test cases and includes a fix to Safari since Safari adds another stack frame for the construct call. Polyfills might do something similar so this tries to be resilient by skipping over those.

### Chrome:
```

    at Lazy
    at Component (http://localhost:8000/fixtures/stacks/Components.js:7:50)
    at div
    at Suspense
    at BabelClassWithFields (http://localhost:8000/fixtures/stacks/BabelClasses-compiled.js:54:31)
    at BabelClass (http://localhost:8000/fixtures/stacks/BabelClasses-compiled.js:36:29)
    at FrozenClass (http://localhost:8000/fixtures/stacks/Components.js:24:5)
    at NativeClass (http://localhost:8000/fixtures/stacks/Components.js:16:1)
    at SuspenseList
    at Custom Name (http://localhost:8000/fixtures/stacks/Components.js:11:23)
    at ErrorBoundary (http://localhost:8000/fixtures/stacks/Example.js:5:1)
    at Example (http://localhost:8000/fixtures/stacks/Example.js:33:21)
```
### Firefox:
```

Lazy
Component@http://localhost:8000/fixtures/stacks/Components.js:7:1
div
Suspense
BabelClassWithFields@http://localhost:8000/fixtures/stacks/BabelClasses-compiled.js:54:31
BabelClass@http://localhost:8000/fixtures/stacks/BabelClasses-compiled.js:36:29
FrozenClass@http://localhost:8000/fixtures/stacks/Components.js:24:5
NativeClass@http://localhost:8000/fixtures/stacks/Components.js:16:49
SuspenseList
DisplayName@http://localhost:8000/fixtures/stacks/Components.js:11:1
ErrorBoundary@http://localhost:8000/fixtures/stacks/Example.js:5:49
Example@http://localhost:8000/fixtures/stacks/Example.js:33:21
```
Note: Firefox doesn't respect displayName in stack traces.

### Safari:
```

Lazy
Component@http://localhost:8000/fixtures/stacks/Components.js:7:48
div
Suspense
BabelClassWithFields@http://localhost:8000/fixtures/stacks/BabelClasses-compiled.js:54:35
BabelClass@http://localhost:8000/fixtures/stacks/BabelClasses-compiled.js:36:34
FrozenClass@http://localhost:8000/fixtures/stacks/Components.js:24:10
NativeClass
[...]

PR #18656, opened 2020-04-17
DevTools: Improved "native" component stacks

bvaughn:
DevTools provides an option to automatically append "component stacks" to third party warnings (`console.warn`) and errors (`console.error`):

image

This works by...
1. DevTools intercepts calls to `console.warn` and `console.error`.
2. If React is currently rendering a component, then the error is related.
3. DevTools looks to see if the error (or warning) already contain a stack string
4. If it doesn't, it appends one.

This PR builds on top of some previous React PRs (#18561 and #18575) to:
* Properly detects the new, experimental component stack format for Chrome, Firefox, and Safari, and
* Use a fork of React's new component stack implementation so that the appended console stacks are properly formatted and clickable.

## Testing

I used Code Sandbox to create a [small test](https://codesandbox.io/s/falling-firefly-9ul27) that included a React warning (with component stack) and a user warning (without one).

#### React v16.9 (and newer)
Screen Shot 2020-04-20 at 3 06 13 PM

#### React v16.8 (and older)
Screen Shot 2020-04-20 at 3 06 53 PM

### Firefox
Screen Shot 2020-04-20 at 3 05 20 PM

### Safari <sup>2</sup>
Screen Shot 2020-04-20 at 3 03 04 PM

<sup>1</sup> Custom component stacks are now clickable in Chrome. They aren't clickable in Firefox but at least they look more native.

<sup>2</sup> Safari does not display newlines properly for any console params after the first. It just shows "⏎ ". Oh well. I guess this is okay since DevTools doesn't support Safari anyway. I used the local DevTools shell to test it all the same.


bvaughn:
Here's what it looks like if DevTools uses the same "native" stack approach as React:

### Chrome
Screen Shot 2020-04-17 at 11 14 55 AM

### Firefox
Screen Shot 2020-04-17 at 11 15 04 AM

### Safari
Screen Shot 2020-04-17 at 11 14 51 AM

packages/react-reconciler/src/ReactFiberReconciler.new.js
```diff
@@ -541,5 +541,6 @@ export function injectIntoDevTools(devToolsConfig: DevToolsConfig): boolean {
     setRefreshHandler: __DEV__ ? setRefreshHandler : null,
     // Enables DevTools to append owner stacks to error messages in DEV mode.
     getCurrentFiber: __DEV__ ? getCurrentFiberForDevTools : null,
+    getStackByFiberInDevAndProd: getStackByFiberInDevAndProd,
```
bvaughn, line 544: I guess I could only inject this is the feature flag is on... It's just so nice though 😅 I'd like to always inject it

bvaughn:
Why won't it let me add my own pending comment without another comment 🤔

packages/react-reconciler/src/ReactFiberReconciler.new.js
sebmarkbage: The current version copies the implementation in the devtools. Same thing with getComponentName. The debug tools for hooks is also injected.

That’s kind of a nice property. Sure it’s dependent on the data structures but so is everything else. I removed getComponentName in prod in this change so if we had injected it we would’ve had to keep it in the prod bundle just for devtools.

Can you import this from shared instead perhaps? Or we can reexport it from react-debug-tools. Is there anything stateful that would break that?

sebmarkbage: I think that it could also work on older react versions too and you don’t have to worry about the feature flag.

bvaughn: > Can you import this from shared instead perhaps?

Not unless I want to make React's feature flag system work in DevTools too (which I'd rather not).

> Or we can reexport it from react-debug-tools. Is there anything stateful that would break that?

This might work. I will look into it.

bvaughn: I think I can just make a DevTools feature flag fork and that might be enough to do this. Seems doable.

bvaughn:
This stack looks a little funky for duplicate key warning:
Screen Shot 2020-04-17 at 1 58 32 PM

bvaughn:
Okay. I have updated this PR to embed (rather than inject) `getStackByFiberInDevAndProd`. This is nice in that it will make this feature more consistent across older versions of React.

In order for DevTools to import from `shared` and `react-reconciler` packages, I had to stub out either:
* The `ReactFeatureFlags` file itself (a fork), or
* `__PROFILE__` and `__EXPERIMENTAL__`  globals

I chose the latter, although I committed them separately. Still, DevTools was already running against an experimental version of React so flipping the `__EXPERIMENTAL__` flag to always on should not really impact anything. (The only things DevTools imports from `shared` or `react-reconciler` anyway is this new component stack helper.)

packages/react-devtools-shared/src/backend/console.js
```diff
@@ -49,19 +52,16 @@ export function dangerous_setTargetConsoleForTesting(
...
   if (typeof findFiberByHostInstance !== 'function') {
     return;
   }
 
   if (typeof getCurrentFiber === 'function') {
-    const {getDisplayNameForFiber} = getInternalReactConstants(version);
-
     injectedRenderers.set(renderer, {
       getCurrentFiber,
-      getDisplayNameForFiber,
```
bvaughn, line 64: Hm... actually now that I think about it, won't this importing rather than injecting change cause problems, @sebmarkbage? b'c of multiple React versions.

sebmarkbage: Just forking seems like a better option then. It's not much code and that allows us to cleanly decouple it without compromising on PROD size.

sebmarkbage: Currently it's the same format for PROD and DEV but one idea would be to do a more advanced but different technique in DEV that rerenders the tree.

One thing we might want to consider is dropping the DEV stacks from the DEV build and only make it a separate package that you install that overrides console.error e.g. for jest environments like a light weight devtools. Then the DevTools injected form would be the only implementation of the DEV stacks which would be a different implementation than the prod one passed to error boundaries.

bvaughn: Forking (to support older versions too) will be more complicated but maybe it's the right thing to do. I'll take a pass at it today.

bvaughn:
PR updated to add a DevTools fork of the new methods. Back to you, @sebmarkbage.

sebmarkbage:
Looks like you have a test failure in case you want to parallelize fixing that while I look.

bvaughn:
Thanks!

packages/react-devtools-shared/src/backend/console.js
```diff
@@ -94,36 +108,31 @@ export function patch(): void {
         try {
           // If we are ever called with a string that already has a component stack, e.g. a React error/warning,
           // don't append a second stack.
+          const lastArg = args.length > 0 ? args[args.length - 1] : null;
           const alreadyHasComponentStack =
-            args.length > 0 && FRAME_REGEX.exec(args[args.length - 1]);
+            lastArg !== null &&
+            (PREFIX_REGEX.exec(lastArg) ||
+              ROW_COLUMN_NUMBER_REGEX.exec(lastArg));
```
sebmarkbage, line 115: I tend to prefer .test() since exec is stateful any returns a match object that you're not using here.

packages/react-devtools-shared/src/backend/console.js
```diff
@@ -7,21 +7,27 @@
...
 
 const APPEND_STACK_TO_METHODS = ['error', 'trace', 'warn'];
 
-const FRAME_REGEX = /\n {4}in /;
+// React's custom built component stack strings match "\s{4}in"
+// Chrome's prefix matches "\s{4}at"
+const PREFIX_REGEX = /\s{4}(in|at)\s{1}/;
+// Firefox and Safari have no prefix ("")
+// but we can fallback to looking for location info (e.g. "foo.js:12:345")
+const ROW_COLUMN_NUMBER_REGEX = /:\d+:\d+$/;
```
sebmarkbage, line 23: Hm. This will only match if the very last line has a line number but that's not always true since we sometimes bail out. It's also possible that the root is a built-in component. E.g. `<Suspense>`. Those won't have a line number. It's probably likely that at least one of them as one but not guaranteed.

packages/react-devtools-shared/src/backend/DevToolsComponentStackFrame.js
```diff
@@ -0,0 +1,298 @@
...
+let prefix;
+export function describeBuiltInComponentFrame(
+  name: string,
+  source: void | null | Source,
+  ownerFn: void | null | Function,
+): string {
+  if (prefix === undefined) {
+    // Extract the VM specific prefix used by each line.
+    try {
+      throw Error();
```
Andarist, line 47: nit: i don't think there is much need to throw this error, `.stack` is synchronously readable

packages/react-devtools-shared/src/backend/console.js
bvaughn: Right, this is not a great check but it's the fallback check. Do you have a better suggestion?

I can't just search for a word like "Suspense" obviously :wink:

packages/react-devtools-shared/src/backend/DevToolsComponentStackFrame.js
sebmarkbage: This is a left over from the upstream implementation where it is required because Internet Explorer requires that you throw the error to assign a stack to it.

In the fork it's not technically needed since we don't support Internet Explorer with DevTools but we could in theory have some user space hosted DevTools that could work in such environments.

bvaughn: FWIW I think the iframe version of DevTools does actually support IE (last I tried it)

packages/react-devtools-shared/src/backend/console.js
bvaughn: > It's also possible that the root is a built-in component. E.g. <Suspense>. Those won't have a line number. It's probably likely that at least one of the frames has one but not guaranteed.

Since the component stack is reverse, the "first" item would be a user component anyway- but this should match an entry on a line other than the first regardless.

I'm banking on there being at least one line with source info, or we won't have a way to differentiate between a stack and an arbitrary string.

sebmarkbage: `/:\d+:\d+(\n|$)/` would at least catch if there's any numeric patterns for any of the rows. That might be enough.

bvaughn: Ah right. I forgot the `m` flag to make `/:\d+:\d+$/` function that way.

Issue #18571, opened 2020-04-10
DevTools: Memo(ForwardRef()) and "Rendered By" List


gaearon:
If you have `memo(forwardRef(X))`, then the inner component won't have a "rendered by" list. This is because it technically doesn't have an owner. It is artificial.

I think we should ideally set up `_debugOwner` for these Fibers in DEV just so existing tooling can find them. Or special case them in DevTools.

gaearon:
It's especially bad because it breaks the "rendered by" chain for inner trees. They "stop" at the inner memo() content (unless it's SimpleMemo).

I think this might also be why RN inspector sometimes shows trees that stop abruptly without going all the way up to the root.

hanq08:
I think this not only happens in forwardRef, any non simple component wrapped in memo will have this issue. for example memo(class Component).

I'd like to work on this. I think the issue lies here potentially https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L421  Let me know if that's ok thanks.

bvaughn:
@hanq08 If you're interested in working on this, we'd welcome a PR.

eps1lon:
bump

bvaughn:
This issue has already been fixed in master. I confirmed it a few minutes ago by running a test on a local build.

PR #18659, opened 2020-04-17
Support inner component _debugOwner in memo

hanq08:
## Summary
Add _debugOwner to inner component of memo.
https://github.com/facebook/react/issues/18571

## Test Plan

one test added. Not familiar with the react test tool. let me know if it's good thanks.

### DIFF
--- a/packages/react-devtools-shared/src/__tests__/ownersListContext-test.js
+++ b/packages/react-devtools-shared/src/__tests__/ownersListContext-test.js
@@ -206,4 +206,42 @@ describe('OwnersListContext', () => {
 
     done();
   });
+
+  it('should include all owners for a component wrapped in react memo ', async done => {
+    const InnerComponent = React.forwardRef((props, ref) => <div ref={ref} />);
+    const Memo = React.memo(InnerComponent);
+    const Grandparent = () => {
+      const ref = React.createRef();
+      return <Memo ref={ref} />;
+    };
+
+    utils.act(() =>
+      ReactDOM.render(<Grandparent />, document.createElement('div')),
+    );
+
+    let didFinish = false;
+    function Suspender({owner}) {
+      const read = React.useContext(OwnersListContext);
+      const owners = read(owner.id);
+      didFinish = true;
+      expect(owners.length).toBe(3);
+      expect(owners).toMatchSnapshot(
+        `owners for "${(owner && owner.displayName) || ''}"`,
+      );
+      return null;
+    }
+
+    const wrapped = ((store.getElementAtIndex(2): any): Element);
+    await utils.actAsync(() =>
+      TestRenderer.create(
+        <Contexts defaultOwnerID={wrapped.id}>
+          <React.Suspense fallback={null}>
+            <Suspender owner={wrapped} />
+          </React.Suspense>
+        </Contexts>,
+      ),
+    );
+    expect(didFinish).toBe(true);
+    done();
+  });
 });
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -429,6 +429,9 @@ function updateMemoComponent(
     child.ref = workInProgress.ref;
     child.return = workInProgress;
     workInProgress.child = child;
+    if (__DEV__) {
+      child._debugOwner = workInProgress;
+    }
     return child;
   }
   if (__DEV__) {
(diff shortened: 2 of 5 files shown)


packages/react-reconciler/src/__tests__/ReactMemo-test.js
```diff
@@ -497,5 +497,21 @@ describe('memo', () => {
...
+    it('inner component of react memo should have _debugOwner set', async () => {
+      const InComponent = React.forwardRef((props, ref) => <div ref={ref} />);
+      const Outer = React.memo(InComponent);
+      const App = () => {
+        const ref = React.createRef();
+        return <Outer ref={ref}>Click me! </Outer>;
+      };
+      ReactNoop.render(<App />);
+      expect(Scheduler).toFlushWithoutYielding();
+      const innerFiber = ReactNoop.getRoot().current.child.child.child;
```
gaearon, line 510: We'd like to avoid tests like this. Is there any behavior that's observable without internal structures?

Maybe there is something testing "Rendered By" in DevTools folder itself?

hanq08: sure yeah it does feel wrong writing test like this. I'll have a look to see if there's somewhere else I can't check.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -429,6 +429,9 @@ function updateMemoComponent(
     child.ref = workInProgress.ref;
     child.return = workInProgress;
     workInProgress.child = child;
+    if (__DEV__) {
+      child._debugOwner = workInProgress;
+    }
```
aweary, line 434: You'll want to include this in `ReactFiberBeginWork.old.js` as well

packages/react-reconciler/src/__tests__/ReactMemo-test.js
aweary: @hanq08 check out the [ownerListContext-test.js](https://github.com/facebook/react/blob/master/packages/react-devtools-shared/src/__tests__/ownersListContext-test.js), I believe that's where you'll want to test this.

hanq08:
@aweary @gaearon could you have a look at this when available thanks

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -429,6 +429,9 @@ function updateMemoComponent(
     child.ref = workInProgress.ref;
     child.return = workInProgress;
     workInProgress.child = child;
+    if (__DEV__) {
+      child._debugOwner = workInProgress;
```
bvaughn, line 433: Can you tell me why this change was necessary? `_debugOwner` is set automatically for all element types.

The newly added test (from this PR) passes for me in master without this change.

packages/react-devtools-shared/src/__tests__/ownersListContext-test.js
```diff
@@ -206,4 +206,42 @@ describe('OwnersListContext', () => {
...
     done();
   });
+
+  it('should include all owners for a component wrapped in react memo ', async done => {
+    const InnerComponent = React.forwardRef((props, ref) => <div ref={ref} />);
+    const Memo = React.memo(InnerComponent);
+    const Grandparent = () => {
+      const ref = React.createRef();
+      return <Memo ref={ref} />;
+    };
```
bvaughn, line 216: nit Let's use a named function so we don't see "Anonymous" in the tests?
```suggestion
    const InnerComponent = (props, ref) => <div ref={ref} />;
    const ForwardRef = React.forwardRef(InnerComponent);
    const Memo = React.memo(ForwardRef);
    const Grandparent = () => {
      const ref = React.createRef();
      return <Memo ref={ref} />;
    };
```

packages/react-reconciler/src/ReactFiberBeginWork.old.js
```diff
@@ -429,6 +429,9 @@ function updateMemoComponent(
     child.ref = workInProgress.ref;
     child.return = workInProgress;
     workInProgress.child = child;
+    if (__DEV__) {
+      child._debugOwner = workInProgress;
+    }
```
bvaughn, line 434: Same question as above.

bvaughn:
Here's how `memo` and `forwardRef` look on a build of React created from master:
owner-list

The HOC badge names are a little messed up in the owners list (which I will look into) but the memo and forwardRef owners lists are in tact.

bvaughn:
I removed my previous comment because it seems to have been in error. Digging in more.

bvaughn:
Moving things over to #19556 since I can't push to the branch this is based off of. Will land there.

Issue #13825, opened 2018-10-11
 script executed when tag name contains capital character


zenany:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
The code is: 
```
class App extends React.Component {
  render() {
    return (
      <div>
        <sCript>alert('Hi');</sCript>
        <script>alert('script 2: this should not execute');</script>
      </div>
    );
  }
}
```
the script in ` <sCript>alert('Hi');</sCript>` executed.
![s_ 2018-10-11 15 56 00](https://user-images.githubusercontent.com/2378632/46789074-3a1ee500-cd6e-11e8-8b0f-1ab172857f7c.png)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

Demo in codepen: https://codepen.io/zenany/pen/rqmPYq

**What is the expected behavior?**

`<sCript>alert('Hi');</sCript>` should not processed as code, this may cause security risk.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

versions:
```
  "dependencies": {
    "react": "^16.5.2",
    "react-dom": "^16.5.2",
  },
```
browser: firefox, chrome

bvaughn:
> `<sCript>alert('Hi');</sCript>` should not processed as code, this may cause security risk.

Why do you say this?

```js
document.createElement('sCript') // <script></script>
```

Also this HTML (without React):
```html
<html>
<body>
<div>
<sCript>alert('foo');</sCript>
<script>alert('bar');</script>
</div>
</body>
</html>
```

Will show both "foo" and "bar" alerts.

HTML is mostly case insensitive, except for certain attributes.

bvaughn:
This issue looks like it's been filed as the result of a misunderstanding, so I'm closing it. Please correct me if I'm missing something though and we can re-open it! 😄

zenany:
> This issue looks like it's been filed as the result of a misunderstanding, so I'm closing it. Please correct me if I'm missing something though and we can re-open it! 😄

Hi, @bvaughn . Can you evaluate the case i described up is a bug or feature?

gaearon:
See https://github.com/facebook/react/pull/19018.

PR #18660, opened 2020-04-17
Fix mistyped script arbitrary code execution vulnerability

### LINKED ISSUE #15978: Using 'scRipt' in React.createElement's type parameter allows you to execute JavaScript code in the child parameter
**Do you want to request a feature or report a bug?**
Report a bug

It's possible to execute JavaScript code written in the child parameter of the `React.createElement` function, if the type parameter consists of a string with mixed casing, like so:

```javascript
React.createElement('scRipt', null, 'alert(1)');
```
https://codesandbox.io/s/sleepy-hodgkin-twxqe

It should not execute the code, but behave exactly like if you typed in 'script' in lowercase.

I would argue that this is unexpected behavior. Judging from the docs (["JSX Prevents Injection Attacks"](https://reactjs.org/docs/introducing-jsx.html#jsx-prevents-injection-attacks)) developers might get the impression, that it is generally safe to embed any user input using JSX. Additionally, if you try to execute JavaScript using script tags, it won't work. 

I _think_ developers might get the idea to give users the ability to write their own "safe" HTML code in comments and the like by parsing their input and converting it to React elements. JSX also allows you to use dynamic input as tag types.

Apart from the mentioned behavior, I'm not aware of any other `type` / `children` combination that would result in JavaScript code execution without a controllable props parameter. So a scenario where a user can control both of these parameters is most likely the only one where this leads to issues.

So I think the code below is likely the culprit (line 423). Please correct me if I'm wrong.

https://github.com/facebook/react/blob/fce15f14d38d680f3bdee3a124e9e94e91520e16/packages/react-dom/src/client/ReactDOMComponent.js#L423-L430

Imo it would make sense to convert the `type` variable to lowercase before comparing it to the 'script' string, so it works as anticipated, even if the casing of the input is user-controllable. Unfortunately I'm not familiar enough with the inner workings of React to submit a pull request of which I'm confident, that it doesn't break anything. This would be my naive solution for this issue:

```javascript
if (typeof type === 'string' && type.toLowerCase() === 'script') { /* ... */ }
```

I'd love to hear your thoughts on this. Is this behavior actually unexpected and if so, should it be fixed?

bl00mber:
fix #15978, #13825

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -430,13 +430,13 @@ export function createElement(
     namespaceURI = getIntrinsicNamespace(type);
   }
   if (namespaceURI === HTML_NAMESPACE) {
-    const loweCaseType = type.toLowerCase();
+    const lowerCaseType = type.toLowerCase();
     
     if (__DEV__) {
       isCustomComponentTag = isCustomComponent(type, props);
       // Should this check be gated by parent namespace? Not sure we want to
       // allow <SVG> or <mATH>.
-      if (!isCustomComponentTag && type !== loweCaseType) {
+      if (!isCustomComponentTag && type !== lowerCaseType) {
         console.error(
           '<%s /> is using incorrect casing. ' +
             'Use PascalCase for React components, ' +
@@ -446,7 +446,7 @@ export function createElement(
       }
     }
 
-    if (loweCaseType === 'script') {
+    if (lowerCaseType === 'script') {
       // Create the script via .innerHTML so its "parser-inserted" flag is
       // set to true and it does not execute
       const div = ownerDocument.createElement('div');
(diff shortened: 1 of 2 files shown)


sophiebits:
We don't support non-lowercase HTML tags (hence the warning), which is why this wasn't previously changed.

aweary:
True, but this case seems exceptional since it's about upholding the invariant that `script` tags are never executed.

bvaughn:
> We don't support non-lowercase HTML tags (hence the warning), which is why this wasn't previously changed.

I dismissed the initial report of this too but... I think it's probably worth fixing this specific case.

bvaughn:
Lesson to self: Don't use GitHub's edit-in-place feature. It seems convenient, but it is not.

gaearon:
I'm going to revert this. See https://github.com/facebook/react/pull/19018.

PR #18664, opened 2020-04-18
Make Flow work with your editor

acdlite:
We typecheck the reconciler against each one of our host configs. `yarn flow dom` checks it against the DOM renderer, `yarn flow native` checks it against the native renderer, and so on.

To do this, we generate separate flowconfig files.

Currently, there is no root-level host config, so running Flow directly via `flow` CLI doesn't work. You have to use the `yarn flow` command and pick a specific renderer.

A drawback of this design, though, is that our Flow setup doesn't work with other tooling. Namely, editor integrations.

I think the intent of this was maybe so you don't run Flow against a renderer than you intended, see it pass, and wrongly think you fixed all the errors. However, since they all run in CI, I don't think this is a big deal. In practice, I nearly always run Flow against the same renderer (DOM), and I'm guessing that's the most common workflow for others, too.

So what I've done in this commit is modify the `yarn flow` command to ~symlink~ [Edit: ended up just copying the file instead because Flow had trouble detecting changes to a symlinked config file] the generated `.flowconfig` file into the root directory. The editor integration will pick this up and show Flow information for whatever was the last renderer you checked.

Everything else about the setup is the same, and all the renderers will continue to be checked by CI.

## Test plan

- Confirmed Flow worked in my editor (vscode) after running `yarn flow dom` once
- Ran `yarn flow native` and confirmed it switched by adding a mistake to the native host config
- Ran `yarn flow-ci` to confirm it ran against each renderer, again by adding a mistake to see if it would pick up on it

### DIFF
--- a/scripts/flow/runFlow.js
+++ b/scripts/flow/runFlow.js
@@ -8,24 +8,65 @@
 'use strict';
 
 const chalk = require('chalk');
-const spawn = require('child_process').spawn;
+const {spawn} = require('child_process');
+const fs = require('fs');
 
+// TODO: This generates all the renderer configs at once. Originally this was
+// to allow the possibility of running multiple Flow processes in parallel, but
+// that never happened. If we did, we'd probably do this in CI, anyway, and run
+// on multiple machines. So instead we could remove this intermediate step and
+// generate only the config for the specified renderer.
 require('./createFlowConfigs');
 
 async function runFlow(renderer, args) {
   return new Promise(resolve => {
-    console.log(
-      'Running Flow on the ' + chalk.yellow(renderer) + ' renderer...',
-    );
     let cmd = __dirname + '/../../node_modules/.bin/flow';
     if (process.platform === 'win32') {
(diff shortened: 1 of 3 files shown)


scripts/flow/runFlow.js
```diff
@@ -21,11 +22,21 @@ async function runFlow(renderer, args) {
     if (process.platform === 'win32') {
       cmd = cmd.replace(/\//g, '\\') + '.cmd';
     }
+
+    // Symlink render flowconfig file to the root of the project so that it
+    // works with editor integrations. This means that the Flow config used by
+    // the editor will correspond to the last renderer you checked.
```
acdlite, line 28: Turns out Flow doesn't detect changes to the symlink so I'll need to be a bit more clever

acdlite: Ended up just copying the file instead

acdlite:
I made it log an error if the file was edited more recently than the template

PR #18665, opened 2020-04-18
Remove unused Flow suppression comment

acdlite:
^ dats what it do


acdlite:
Ah nvm it still fails in native. Whatever, it's fine as-is :D

PR #18668, opened 2020-04-18
Refine the heuristics around beforeblur/afterblur

trueadm:
This is a follow up to https://github.com/facebook/react/pull/18662. Notably, that PR was a temporary fix, which turned out to still have issues when testing locally on FB.

I listened to the advice from @sebmarkbage and we now handle the logic for determining if an active element fiber is in a hidden or deleted sub-tree in the "before mutation" phase. I also took the opportunity to break apart the logic into separate host config methods which match the intent nicely, without needing to leak fiber details to the host configs: so now we expost to new methods called `beforeActiveInstanceBlur` and `afterActiveInstanceBlur`.

We can always improve and optimize this logic once the refactor for creating fast-paths comes in. I also thought about using an array to contain all the fibers, but this turned out quite large in practice. Internally this array can be extremely larger in some cases, so we'd probably want to use a `Set`.

I went with an implementation that works: it seems to completely fix the issue internally and all tests relating to this now pass everytime (unlike before, where they'd sometimes fail).

I also added a bunch of regression tests to prevent this issue from occuring in the future.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -242,10 +242,10 @@ export function getPublicInstance(instance: Instance): * {
   return instance;
 }
 
-export function prepareForCommit(containerInfo: Container): Object | void {
+export function prepareForCommit(containerInfo: Container): Object | null {
   eventsEnabled = ReactBrowserEventEmitterIsEnabled();
   selectionInformation = getSelectionInformation();
-  let activeInstance;
+  let activeInstance = null;
   if (enableDeprecatedFlareAPI || enableUseEventAPI) {
     const focusedElem = selectionInformation.focusedElem;
     if (focusedElem !== null) {
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -91,10 +91,6 @@ import {
 import {getListenerMapForElement} from '../events/DOMEventListenerMap';
 import {TOP_BEFORE_BLUR, TOP_AFTER_BLUR} from '../events/DOMTopLevelEventTypes';
 
-// TODO: This is an exposed internal, we should move this around
-// so this isn't the case.
-import {isFiberInsideHiddenOrRemovedTree} from 'react-reconciler/src/ReactFiberTreeReflection';
-
 export type ReactListenerEvent = ReactDOMListenerEvent;
 export type ReactListenerMap = ReactDOMListenerMap;
 export type ReactListener = ReactDOMListener;
@@ -159,7 +155,6 @@ export opaque type OpaqueIDType =
     };
 
 type SelectionInformation = {|
-  activeElementDetached: null | HTMLElement,
   focusedElem: null | HTMLElement,
   selectionRange: mixed,
 |};
@@ -247,32 +242,40 @@ export function getPublicInstance(instance: Instance): * {
   return instance;
 }
 
-export function prepareForCommit(containerInfo: Container): void {
+export function prepareForCommit(containerInfo: Container): Object | void {
   eventsEnabled = ReactBrowserEventEmitterIsEnabled();
   selectionInformation = getSelectionInformation();
+  let activeInstance;
   if (enableDeprecatedFlareAPI || enableUseEventAPI) {
     const focusedElem = selectionInformation.focusedElem;
     if (focusedElem !== null) {
-      const instance = getClosestInstanceFromNode(focusedElem);
-      if (instance !== null && isFiberInsideHiddenOrRemovedTree(instance)) {
-        dispatchBeforeDetachedBlur(focusedElem);
-      }
+      activeInstance = getClosestInstanceFromNode(focusedElem);
     }
   }
   ReactBrowserEventEmitterSetEnabled(false);
+  return activeInstance;
+}
+
+export function beforeActiveInstanceBlur(): void {
+  if (enableDeprecatedFlareAPI || enableUseEventAPI) {
+    ReactBrowserEventEmitterSetEnabled(true);
+    dispatchBeforeDetachedBlur((selectionInformation: any).focusedElem);
+    ReactBrowserEventEmitterSetEnabled(false);
+  }
+}
+
+export function afterActiveInstanceBlur(): void {
+  if (enableDeprecatedFlareAPI || enableUseEventAPI) {
+    ReactBrowserEventEmitterSetEnabled(true);
+    dispatchAfterDetachedBlur((selectionInformation: any).focusedElem);
+    ReactBrowserEventEmitterSetEnabled(false);
+  }
 }
 
 export function resetAfterCommit(containerInfo: Container): void {
   restoreSelection(selectionInformation);
   ReactBrowserEventEmitterSetEnabled(eventsEnabled);
   eventsEnabled = null;
-  if (enableDeprecatedFlareAPI || enableUseEventAPI) {
-    const activeElementDetached = (selectionInformation: any)
-      .activeElementDetached;
-    if (activeElementDetached !== null) {
-      dispatchAfterDetachedBlur(activeElementDetached);
-    }
-  }
   selectionInformation = null;
 }
 
@@ -525,8 +528,6 @@ function createEvent(type: TopLevelType): Event {
 }
 
 function dispatchBeforeDetachedBlur(target: HTMLElement): void {
-  ((selectionInformation: any): SelectionInformation).activeElementDetached = target;
-
   if (enableDeprecatedFlareAPI || enableUseEventAPI) {
     const event = createEvent(TOP_BEFORE_BLUR);
     // Dispatch "beforeblur" directly on the target,
--- a/packages/react-interactions/events/src/dom/use-event/__tests__/useFocusWithin-test.internal.js
+++ b/packages/react-interactions/events/src/dom/use-event/__tests__/useFocusWithin-test.internal.js
@@ -42,17 +42,22 @@ const table = [[forcePointerEvents], [!forcePointerEvents]];
 
 describe.each(table)(`useFocus`, hasPointerEvents => {
   let container;
+  let container2;
 
   beforeEach(() => {
     initializeModules(hasPointerEvents);
     container = document.createElement('div');
     document.body.appendChild(container);
+    container2 = document.createElement('div');
+    document.body.appendChild(container2);
   });
 
   afterEach(() => {
     ReactDOM.render(null, container);
     document.body.removeChild(container);
+    document.body.removeChild(container2);
     container = null;
+    container2 = null;
   });
 
   describe('disabled', () => {
@@ -465,9 +470,6 @@ describe.each(table)(`useFocus`, hasPointerEvents => {
         );
       };
 
-      const container2 = document.createElement('div');
-      document.body.appendChild(container2);
-
       const root = ReactDOM.createRoot(container2);
 
       act(() => {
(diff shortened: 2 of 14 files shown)


packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -250,32 +245,94 @@ export function getPublicInstance(instance: Instance): * {
...
 
-export function prepareForCommit(containerInfo: Container): void {
+function createEvent(type: TopLevelType): Event {
+  const event = document.createEvent('Event');
+  event.initEvent(((type: any): string), false, false);
+  return event;
+}
+
+function dispatchBeforeDetachedBlur(target: HTMLElement): void {
+  const targetInstance = getClosestInstanceFromNode(target);
```
gaearon, line 255: If both flags are off, can we exit early so this function can be fully DCE?

packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -342,20 +342,52 @@ export function isFiberSuspenseAndTimedOut(fiber: Fiber): boolean {
...
+    if (node === parentFiber || node === parentFiberAlternate) {
       return true;
     }
-    lastChild = node;
     node = node.return;
   }
   return false;
 }
+
+function isFiberHiddenAndContains(
```
gaearon, line 357: It’s a bit unclear to me whether which fiber we’re testing for whether it’s hidden from the name.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1901,9 +1904,19 @@ function commitRootImpl(root, renderPriorityLevel) {
     // The first phase a "before mutation" phase. We use this phase to read the
     // state of the host tree right before we mutate it. This is where
     // getSnapshotBeforeUpdate is called.
-    prepareForCommit(root.containerInfo);
+    const activeInstanceFiber = prepareForCommit(root.containerInfo) || null;
```
gaearon, line 1907: Hmm. So far we’ve almost avoided having Fibers be part of host config signatures. Except for events. Is it unavoidable here?

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
```diff
@@ -1920,9 +1923,19 @@ function commitRootImpl(root, renderPriorityLevel) {
...
-    prepareForCommit(root.containerInfo);
+    const activeInstanceFiber = prepareForCommit(root.containerInfo) || null;
+    let shouldFireAfterActiveInstanceBlur = false;
+
     nextEffect = firstEffect;
     do {
+      if (
+        !shouldFireAfterActiveInstanceBlur &&
+        activeInstanceFiber !== null &&
+        isFiberHiddenOrDeletedAndContains(nextEffect, activeInstanceFiber)
```
gaearon, line 1934: This puts a traversal inside another traversal. Have we measured the impact of this?

packages/react-dom/src/client/ReactDOMHostConfig.js
trueadm: This function only gets called from where there are flags. It gets removed from the output without needing the flags here.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
trueadm: If we pass back the host element then we need to ask the host for its instance again, so it's pretty unavoidable here.

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
trueadm: We have traversal within traversal before when we used to do these checks in `hideInstance`, so the impact is roughtly equivalent. Like I mentioned, we could maybe avoid this by putting all the current effects in a `Set` or an array, but that seems more involved and error prone.

packages/react-reconciler/src/ReactFiberTreeReflection.js
trueadm: Updated functions to better reflect intent.

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
trueadm: As @sebmarkbage mentioned last night:

> We could check if an effect is a deletion in the snapshot phase, and if it is, backtrack the active fiber's return pointer to see if we find the same fiber on three. It's not great since it'll make many deletions slower when there is at least one focused element. But it's faster for commits with no deletions, the same for one deletion and linearly worse for many deletions.

gaearon: I think his comment implies you'd only do this for deletions. But the code not only backtracks for deletions, but for every fiber. Because it does two checks (the other one being "is in hidden tree"). That second check always traverses upwards.

gaearon: >We have traversal within traversal before when we used to do these checks in hideInstance, so the impact is roughtly equivalent. 

I don't understand this. `hideInstance` is only called for things we're actually hiding. It fires very rarely. Like once a second or something.

Whereas this traversal is embedded into the hottest path — effect traversal. There can be hundreds of effect for every commit, which itself may happen dozen of times per second.

trueadm: We no longer need to do this extra traversal as I moved the code that handles this into a different place `commitBeforeMutationEffects`.

packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -342,20 +342,45 @@ export function isFiberSuspenseAndTimedOut(fiber: Fiber): boolean {
...
+    if (node === parentFiber || node === parentFiberAlternate) {
       return true;
     }
-    lastChild = node;
     node = node.return;
   }
   return false;
 }
+
+function isFiberInHiddenTreeThatsContainsTargetFiber(
```
gaearon, line 357: Maybe `isFiberTimedOutSuspenseThatContainsTargetFiber`?

I think I just understood what this is checking. This isn't meant to find whether a Fiber is "somewhere in a timed out tree", this is meant to check whether the Fiber is literally the `<Suspense>` node itself that has timed out.

trueadm: Yes.

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
trueadm: > I don't understand this.

Do you understand this now, given your other comment? This logic was more complex before because of a subtle bug that I fixed – but I didn't remove this complexity from this location (it was the weekend and I got told off for working haha). I also rebased master and fixed the conflicts.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -247,32 +242,40 @@ export function getPublicInstance(instance: Instance): * {
...
   selectionInformation = getSelectionInformation();
+  let activeInstance;
   if (enableDeprecatedFlareAPI || enableUseEventAPI) {
     const focusedElem = selectionInformation.focusedElem;
     if (focusedElem !== null) {
-      const instance = getClosestInstanceFromNode(focusedElem);
-      if (instance !== null && isFiberInsideHiddenOrRemovedTree(instance)) {
-        dispatchBeforeDetachedBlur(focusedElem);
-      }
+      activeInstance = getClosestInstanceFromNode(focusedElem);
```
gaearon, line 252: This may give you null. So the return type seems a bit inconsistent (void or null or Fiber). Can we make it always `null | Fiber`?

trueadm: I originally did that, but then I had to change every renderer where we use this method to return null and it seemed a bit overdrawn. Happy to do that again though.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -247,32 +242,40 @@ export function getPublicInstance(instance: Instance): * {
   return instance;
 }
 
-export function prepareForCommit(containerInfo: Container): void {
+export function prepareForCommit(containerInfo: Container): Object | void {
   eventsEnabled = ReactBrowserEventEmitterIsEnabled();
   selectionInformation = getSelectionInformation();
+  let activeInstance;
```
gaearon, line 248: In this file, "instance" usually refers to the `Instance` type. So this is a bit confusing and easy to accidentally return an actual `Instance`, which type checker wouldn't catch.

Maybe `focusedInstanceHandle`, to match other `*InstanceHandle` in this file?

packages/react-interactions/events/src/dom/use-event/__tests__/useFocusWithin-test.internal.js
```diff
@@ -463,5 +497,78 @@ describe.each(table)(`useFocus`, hasPointerEvents => {
...
+      act(() => {
+        root.render(<Component />);
+      });
+      jest.runAllTimers();
+      expect(onBeforeBlurWithin).toHaveBeenCalledTimes(1);
+      expect(onAfterBlurWithin).toHaveBeenCalledTimes(1);
+
+      resolve();
+
+      document.body.removeChild(container2);
```
gaearon, line 571: This won't run if the test fails. Would that confuse the test? Generally, might be better to do in `afterEach`.

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
```diff
@@ -296,6 +299,9 @@ let currentEventTime: ExpirationTime = NoWork;
 // We warn about state updates for unmounted components differently in this case.
 let isFlushingPassiveEffects = false;
 
+let activeInstanceFiber = null;
+let shouldFireAfterActiveInstanceBlur = null;
```
gaearon, line 303: Should this start as `false`?

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
```diff
@@ -296,6 +299,9 @@ let currentEventTime: ExpirationTime = NoWork;
 // We warn about state updates for unmounted components differently in this case.
 let isFlushingPassiveEffects = false;
 
+let activeInstanceFiber = null;
```
gaearon, line 302: Can you add a type to this?

packages/react-dom/src/client/ReactDOMHostConfig.js
gaearon: Alternatively we could make it explicit in the type. Like `null | void | Object`. It's just easy to miss otherwise.

gaearon: We could also make it always return `undefined` instead of `null`.

trueadm: I've made it `null | Fiber`.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -301,8 +301,8 @@ let currentEventTime: ExpirationTime = NoWork;
 // We warn about state updates for unmounted components differently in this case.
 let isFlushingPassiveEffects = false;
 
-let activeInstanceFiber = null;
-let shouldFireAfterActiveInstanceBlur = null;
+let focusedInstanceHandle = null;
```
gaearon, line 304: Out of sync with the old file

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1924,6 +1932,9 @@ function commitRootImpl(root, renderPriorityLevel) {
       }
     } while (nextEffect !== null);
 
+    // We no longer need to track the active instance fiber
+    focusedInstanceHandle = null;
```
gaearon, line 1936: FWIW, in this file we do call them fibers. It's just the host config that's intended to mostly stay agnostic.

PR #18670, opened 2020-04-18
Repro bug for Google Closure Compiler

trueadm:
This strange set of changes causes our ReactDOM Prod bundle to have broken JS. 

Run these commands to see the error:

`yarn build --type=NODE_PROD`

`yarn test-prod-build ReactDOMServerIntegrationTextarea`

You'll see this error:

`TypeError: Cannot read property 'focusedElem' of null`

It seems to be related to `JSCompiler_inline_result`. Which incorrectly gets re-used from a `null` value.

Issue reported:  google/closure-compiler#3588

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "webpack": "^4.41.2"
   },
   "devEngines": {
-    "node": "8.x || 9.x || 10.x || 11.x || 12.x || 13.x"
+    "node": "*"
   },
   "jest": {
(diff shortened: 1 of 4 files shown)


gaearon:
Did you submit an issue report to GCC?

trueadm:
@gaearon Made an issue just now: https://github.com/google/closure-compiler/issues/3588

gaearon:
This is not a GCC but likely a Rollup issue.
https://github.com/rollup/rollup/issues/4455

Issue #18669, opened 2020-04-18
Add More Fuzz Tests


gaearon:
There’s some things we don’t have a sufficient coverage of. Currently we catch them from production product bugs but this is not sustainable. We’re hoping some refactors will drastically simplify the model — but nevertheless we should invest in better fuzz test coverage. That’s how we caught similar bugs before at an earlier stage.

One thing we’re lacking coverage for is what happens to Suspense boundaries as updates are dispatched at different priorities in different order, and what happens when we’ve had to yield. We need to verify that Suspense always “wakes up” when Promises are resolved and there’s nothing to be suspended on. We also need to test this in combination with render phase updates. 

Here’s examples of bugs that I want a fuzzer to catch: https://github.com/facebook/react/issues/18657 https://github.com/facebook/react/issues/18020 https://github.com/facebook/react/issues/18486 https://github.com/facebook/react/issues/18357 https://github.com/facebook/react/issues/18353 https://github.com/facebook/react/issues/18644 https://github.com/facebook/react/pull/18412. 

@dubzzz I believe you were interested in this? This would take some effort but would be a major contribution.

dubzzz:
Yes, I'll be really interested in helping on this subject.

For the moment, I already drafted some fuzz tests based on property based testing for `SuspenseList` ([here](https://github.com/dubzzz/react/blob/e2cb4776ea3634fab2021d83cb8259bde03c0f3a/packages/react-reconciler/src/__tests__/ReactSuspenseList-test.property.js)). They detected the issue #18661.

Waiting your feedback on the following scenarii and on #18673.

---

Here are the scenarii I am covering so far with those tests - _please let me know if my scenarii are based on wrong hypothesis_:

- should wait all components to be resolved before showing any in "together" mode

The test generates a random tree of components with only `SuspenseList` having `mode="together"` and `Suspense`.

If there is still one pending `Suspense`, it checks that the `SuspenseList` is still waiting and not rendering anything. Then it resolves out-of-order one of the pending `Suspense`.

As soon as everything gets resolved, the test checks that everything is displayed properly.

- should display components up-to the first unresolved one as resolved, next ones should be considered unresolved in "forward" mode

Close to the test described above but for `mode="forwards"`.

The check is obviously different: We expect all the `Suspense` up to the first unresolved one not included to be shown correctly. And starting from the first unresolved one we expect them to show the fallback.

[...]

Issue #18661, opened 2020-04-17
Bug: Inconsistencies between nested SuspenseList and single SuspenseList


dubzzz:
**React version:** 0.0.0-experimental-e5d06e34b

Hoping this can be helpful for you. Here is what looks to be a bug with concurrent mode and nested SuspenseList.

## Steps To Reproduce

_In concurrent mode_

0. Setup

Let's suppose we have three components `<A />` (**not** lazy loaded), `<B />` (**not** lazy loaded) and `<C />` (lazy loaded).

In other words:

```jsx
import A from './A';
import B from './B';
const C = React.lazy(() => import('./C'));
```

1. Render

```jsx
render(
  <SuspenseList key="1" revealOrder="forwards">
    <SuspenseList key="1.1" revealOrder="forwards">
      <Suspense key="1.1.a" fallback={<Text text="Loading A" />}>
        <A />
      </Suspense>
    </SuspenseList>
  </SuspenseList>
)
```

2. Update the component (component now shows A, B, and C)

```jsx
render(
  <SuspenseList key="1" revealOrder="forwards">
    <SuspenseList key="1.1" revealOrder="forwards">
      <Suspense key="1.1.a" fallback={<Text text="Loading A" />}>
        <A />
      </Suspense>
      <Suspense key="1.1.b" fallback={<Text text="Loading B" />}>
        <B />
      </Suspense>
      <Suspense key="1.1.c" fallback={<Text text="Loading C" />}>
        <C />
      </Suspense>
    </SuspenseList>
  </SuspenseList>
)
```

3. Output is: `A / Loading B / Loading C`. While `B` has already been loaded (not lazy loaded). If I understand well the behaviour of `forwards` I would have expect to have `A / B / Loading C` instead.

Please note that the behaviour is not the same if I do not use nested `<SuspenseList />`.

Live repro: https://dubzzz.github.io/react-suspenselist-bug/build/
Link to code example (started with CRA): https://github.com/dubzzz/react-suspenselist-bug

## The current behavior

Output is: `A / Loading B / Loading C` with nested `SuspenseList`
Output is: `A / B / Loading C` without nested `SuspenseList`

## The expected behavior

Output is: `A / B / Loading C` with both

**How did I found this bug?**

This _potential bug_ has been discovered while I was trying to run property based tests based on [fast-check](https://github.com/dubzzz/fast-check/) against React library.

[...]

PR #18673, opened 2020-04-18
Migrate ReactSuspense fuzz tests to Property Based Tests

dubzzz:
## Summary

As discussed with @gaearon in #14147 and in the context of #18669, here is a Pull Request to move some of the tests using home-made fuzzing to property based testing solutions.

The advantage of switching to a property based testing approach instead of manually generated random values is that property based testing frameworks come with shrinking capabilities by default. It means that anytime such framework detects an issue in a test (also referred as property) it will try to produce simpler values producing the error too.

As of today, there are mainly 3 leading libraries for property based testing: [fast-check](https://github.com/dubzzz/fast-check), [jsverify](https://github.com/jsverify/jsverify) or [testcheck-js](https://github.com/leebyron/testcheck-js). _fast-check_ is the choice added by this PR as it comes with features that would be pretty useful in the future: built-in helpers to detect race-conditions.

Depending on how much time it takes me to migrate `ReactNewContext-test.js` to _fast-check_, I might add this test to the patch too.

## Test Plan

Move to property based testing for ReactSuspense fuzz tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "eslint-plugin-no-for-of-loops": "^1.0.0",
     "eslint-plugin-react": "^6.7.1",
     "eslint-plugin-react-internal": "link:./scripts/eslint-rules",
+    "fast-check": "^1.24.1",
     "fbjs-scripts": "0.8.3",
     "filesize": "^6.0.1",
     "flow-bin": "0.97",
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseFuzz-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseFuzz-test.internal.js
@@ -3,9 +3,8 @@ let Suspense;
 let ReactNoop;
 let Scheduler;
 let ReactFeatureFlags;
-let Random;
 
-const SEED = process.env.FUZZ_TEST_SEED || 'default';
+const fc = require('fast-check');
 const prettyFormatPkg = require('pretty-format');
 
 function prettyFormat(thing) {
@@ -27,7 +26,6 @@ describe('ReactSuspenseFuzz', () => {
     Suspense = React.Suspense;
     ReactNoop = require('react-noop-renderer');
     Scheduler = require('scheduler');
-    Random = require('random-seed');
   });
 
   function createFuzzer() {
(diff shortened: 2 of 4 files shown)


acdlite:
I am very disinclined to add a library just for the shrinking capability. Is there some other advantage?

acdlite:
To elaborate, automatic shrinking solves a problem that I have not ever experienced. I'm skeptical of adding a layer of third-party abstractions around what is essentially a fancy way of generating random values.

I also don't see much point in converting an existing test file that works fine as-is. If you want to add a new fuzz tester that uses one of those libraries, I'll put up less of a fight. I'll still challenge you on the need to add a dependency. Maybe I'd merge the PR then rewrite it without the library :)

dubzzz:
I totally get your point.

I'd say the most important benefit you'll get from fast-check are not those migrated tests today. Or at least for the moment, as they are green, everything will look the same for you. Shrinking might be helpful and clearly you can leave without that. 

But I believe fast-check will be really helpful for tests concerning detecting race conditions in Suspense and SuspenseList. It already proved useful as it detected #18661 and I believe it can be a master piece to help you with issue #18669. Indeed, in addition to classical property based testing capabilities it comes with a built-in support to reschedule asynchrounous operations and user interactions. You may be interested to get a look to the tests I wrote for Suspense and SuspenseList on https://github.com/dubzzz/react/commit/e2cb4776ea3634fab2021d83cb8259bde03c0f3a (and described in the issue #18669 - and in this [code snippet](https://github.com/dubzzz/fast-check/blob/master/example/005-race/debounced-autocomplete/main.spec.tsx) showing how the scheduler feature could be used to test races on an autocomplete).

Something that might also be useful is that fast-check will by default start with "smaller" generated entries and move towards "larger" ones as the test goes (referred as [biased in the documentation](https://github.com/dubzzz/fast-check/blob/master/documentation/1-Guides/AdvancedArbitraries.md#biased-arbitraries)). While it may look useless, it is pretty helpful in general to detect bugs. **Some bugs will never be detected with full random**. For instance, if you want to check the stability of a sort you can say sortOnFirstName(sortOnLastName()) is sortOnFirstNameThenOnLastName(). If to test that with fuzz technics you only generate random arrays of objects with two strings it might takes very long to have a single case where first name is duplicated and even longer to find a bug on stability.

In case, it can help, fast-check is already used by Jest and Jasmine.

acdlite:
How are you distinguishing fuzz tests from property tests? What would you say is the key difference? In my view, our existing fuzz tests are property tests because they test properties (mostly consistency) that should always hold true. We call them “fuzz” tests because they use randomly generated values to do it.

It still sounds to me like you’re mostly extolling the benefits of property testing and not specifically these libraries. I don’t think I need to be sold on property tests, only on using a framework to do it.

PR #18674, opened 2020-04-18
Repro bug Nested SuspenseList may display fallbacks while the component is loaded on re-render

dubzzz:
## Summary

As asked by @gaearon on the issue #18661, here is a simple unit test to reproduce the issue: _Nested SuspenseList may display fallbacks while the component is loaded on re-render_.

As the bug is still there today, the test is red (as expected).

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactSuspenseList-test.js
+++ b/packages/react-reconciler/src/__tests__/ReactSuspenseList-test.js
@@ -847,8 +847,10 @@ describe('ReactSuspenseList', () => {
         <span>C</span>
       </>,
     );
-  }); // @gate experimental
-  it('displays each items in "forwards" order in nested SuspenseLists on re-render', async () => {
+  });
+
+  // @gate experimental
+  it('displays each items in "forwards" order in nested SuspenseLists after re-render', async () => {
     const A = createAsyncText('A');
     const B = createAsyncText('B');
     const C = createAsyncText('C');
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/__tests__/ReactSuspenseList-test.js
```diff
@@ -847,6 +847,83 @@ describe('ReactSuspenseList', () => {
...
+
+    ReactNoop.render(<Foo showMore={true} />);
+
+    expect(Scheduler).toFlushAndYield([
+      'Another child',
+      'A',
+      'B',
+      'Suspend! [C]',
+      'Loading C',
+    ]);
```
dubzzz, line 904: Not fully sure of this expectation but it seems to be correct as only C is supposed to suspend in that case. And all others should be called for a render.

Please note that the current value received at this point is:

```suggestion
    expect(Scheduler).toFlushAndYield([
      'Another child',
      'A',
      'B',
      'Suspend! [C]',
      'Loading C',
      'Another child',
      'A',
      'Loading B',
      'Loading C',
    ]);
```

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.js
```diff
@@ -847,6 +847,83 @@ describe('ReactSuspenseList', () => {
...
+
+    expect(ReactNoop).toMatchRenderedOutput(
+      <>
+        <span>Another child</span>
+        <span>A</span>
+        <span>B</span>
+        <span>Loading C</span>
+      </>,
+    );
```
dubzzz, line 913: At this point we currently receive:

```suggestion
    expect(ReactNoop).toMatchRenderedOutput(
      <>
        <span>Another child</span>
        <span>A</span>
        <span>Loading B</span>
        <span>Loading C</span>
      </>,
    );
```

Please also note that if you remove the first call to render, the expectations will all pass. The bug only happens on a second render.

eps1lon:
Bump

PR #18676, opened 2020-04-20
fix: skip dangerouslySetInnerHtml hydration warning if it's undefined

### LINKED ISSUE #17170: dangerouslySetInnerHTML, children, and a bogus hydration warning

Bug (initially reported at https://github.com/zeit/next.js/issues/9173).

If a component has both `dangerouslySetInnerHTML` and `children` props, and the component is used with just the `children` prop, on page load the client logs a warning beginning with ``Warning: Prop `dangerouslySetInnerHTML` did not match.``.

Interestingly:

- If just `dangerouslySetInnerHTML` is used there is no warning.
- Multiple instances with just a `children` prop results in only one warning, for the first occurance.

Demo: https://codesandbox.io/s/react-hydration-warning-demo-3je59

In a Next.js project, create a page with the following content:

```jsx
const TestComponent = ({ dangerouslySetInnerHTML, children }) => (
  <div dangerouslySetInnerHTML={dangerouslySetInnerHTML} children={children} />
)

export default () => <TestComponent>a</TestComponent>
```

Loading the page in a browser will result in this warning logging to the console:

```
Warning: Prop `dangerouslySetInnerHTML` did not match. Server: "a" Client: ""
```

Note that the hydration warning is bogus; using view source and the inspector you can see the SSR and client rendered HTML is identical and correct.

This works just the same as before but without a hydration warning:

```jsx
const TestComponent = ({ dangerouslySetInnerHTML, children }) => {
  const divProps = {}
  if (dangerouslySetInnerHTML)
    divProps.dangerouslySetInnerHTML = dangerouslySetInnerHTML
  if (children) divProps.children = children

  return <div {...divProps} />
}
```

There should be no hydration warning at first client render.

Not sure if the bug is present in old versions of React, but it can be seen with v16.11.0.

eps1lon:
## Summary

Closes #17170

## Test Plan

Added tests to `ReactServerRenderingHydration-test.js` that make sure that
1. no warning is issued if `dangerouslySetInnerHTML` is `undefined` and the markup matches
2. warnings are issued if `dangerouslySetInnerHTML` is `undefined` and the markup does not match.

Verify fix with codesandboxes in #17170:
- Packages built from codesandbox CI are not working with `react-dom/server` /cc @CompuIves 
- Without `react-dom/server`: https://codesandbox.io/s/dangerouslysetinnerhtml-undefined-e14qg?file=/src/index.js logs no more warnings as expected.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactServerRenderingHydration-test.js
+++ b/packages/react-dom/src/__tests__/ReactServerRenderingHydration-test.js
@@ -564,4 +564,25 @@ describe('ReactDOMServerHydration', () => {
 
     expect(domElement.innerHTML).toEqual(markup);
   });
+
+  it('should warns if innerHTML mismatches with dangerouslySetInnerHTML=undefined on the client', () => {
+    const domElement = document.createElement('div');
+    const markup = ReactDOMServer.renderToStaticMarkup(
+      <div dangerouslySetInnerHTML={{__html: '<p>server</p>'}} />,
+    );
+    domElement.innerHTML = markup;
+
+    expect(() => {
+      ReactDOM.hydrate(
+        <div dangerouslySetInnerHTML={undefined}>
+          <p>client</p>
+        </div>,
+        domElement,
+      );
+
+      expect(domElement.innerHTML).not.toEqual(markup);
+    }).toErrorDev(
+      'Warning: Text content did not match. Server: "server" Client: "client"',
+    );
+  });
 });
--- a/packages/react-dom/src/__tests__/ReactServerRenderingHydration-test.js
+++ b/packages/react-dom/src/__tests__/ReactServerRenderingHydration-test.js
@@ -565,7 +565,7 @@ describe('ReactDOMServerHydration', () => {
     expect(domElement.innerHTML).toEqual(markup);
   });
 
-  it('should warns if innerHTML mismatches with dangerouslySetInnerHTML=undefined on the client', () => {
+  it('should warn if innerHTML mismatches with dangerouslySetInnerHTML=undefined and children on the client', () => {
     const domElement = document.createElement('div');
     const markup = ReactDOMServer.renderToStaticMarkup(
       <div dangerouslySetInnerHTML={{__html: '<p>server</p>'}} />,
@@ -583,6 +583,22 @@ describe('ReactDOMServerHydration', () => {
       expect(domElement.innerHTML).not.toEqual(markup);
     }).toErrorDev(
       'Warning: Text content did not match. Server: "server" Client: "client"',
+    );
+  });
+
+  it('should warn if innerHTML mismatches with dangerouslySetInnerHTML=undefined on the client', () => {
+    const domElement = document.createElement('div');
+    const markup = ReactDOMServer.renderToStaticMarkup(
+      <div dangerouslySetInnerHTML={{__html: '<p>server</p>'}} />,
+    );
+    domElement.innerHTML = markup;
+
+    expect(() => {
+      ReactDOM.hydrate(<div dangerouslySetInnerHTML={undefined} />, domElement);
+
+      expect(domElement.innerHTML).not.toEqual(markup);
+    }).toErrorDev(
+      'Warning: Did not expect server HTML to contain a <p> in <div>',
     );
   });
 });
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/__tests__/ReactServerRenderingHydration-test.js
```diff
@@ -548,4 +548,41 @@ describe('ReactDOMServerHydration', () => {
...
+    );
+    const markup = ReactDOMServer.renderToStaticMarkup(reactElement);
+    domElement.innerHTML = markup;
+
+    ReactDOM.hydrate(reactElement, domElement);
+
+    expect(domElement.innerHTML).toEqual(markup);
+  });
+
+  it('should warns if innerHTML mismatches with dangerouslySetInnerHTML=undefined on the client', () => {
```
gaearon, line 568: should warn

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -1110,12 +1110,11 @@ export function diffHydratedProperties(
       } else if (propKey === DANGEROUSLY_SET_INNER_HTML) {
         const serverHTML = domElement.innerHTML;
         const nextHtml = nextProp ? nextProp[HTML] : undefined;
-        const expectedHTML = normalizeHTML(
-          domElement,
-          nextHtml != null ? nextHtml : '',
-        );
-        if (expectedHTML !== serverHTML) {
-          warnForPropDifference(propKey, serverHTML, expectedHTML);
+        if (nextHtml != null) {
```
gaearon, line 1113: So.. what if `nextHtml` is `null` or `undefined`, but it was not `null` on the server? Wouldn't this miss a warning?

eps1lon: I had the exact same question. This is pretty similar to the second test I added (https://github.com/facebook/react/pull/18676/files#diff-ab371863932cd2e8f0ba14ff2eaab380R568). I can change it not render anything on the client or add a new one that goes from `{ __html: "server" }` to `undefined`.

gaearon: Adding new one sounds good.

gaearon: I'm just asking how it works.

eps1lon: I didn't read much of the surrounding code and relied on the tests doing what I expected so I can't tell you how it works.

Maybe I should convert it to a table and `it.each` so that it's more visible  how the props change between client and server? The tests are mostly setup.

gaearon: We prefer explicitly written tests to `it.each`. Don't worry about duplication.

>I didn't read much of the surrounding code and relied on the tests doing what I expected so I can't tell you how it works.

I think we need to figure out how it works before we can be comfortable merging this. :-)

packages/react-dom/src/__tests__/ReactServerRenderingHydration-test.js
eps1lon: fixed

packages/react-dom/src/client/ReactDOMComponent.js
eps1lon: > I think we need to figure out how it works before we can be comfortable merging this. :-)

Sure thing. Let me trace this for each test case. Maybe this clears this up.

CompuIves:
> Packages built from codesandbox CI are not working with react-dom/server /cc @CompuIves

Could you shoot me a Twitter DM or email? I'm not sure what's broken, but want to look into it!

packages/react-dom/src/client/ReactDOMComponent.js
eps1lon: When tracing the code I wasn't sure if this might result in a loss of component stacks so I added these to the test to make sure.

It looks like the reconciler is going through all the instances that aren't hydrated and deletes them:
https://github.com/facebook/react/blob/dc630d3374e8b3d01cdf9e12a81eab9322ac6832/packages/react-reconciler/src/ReactFiberHydrationContext.old.js#L462

The host config is then responsible for warning about the deletion of hydrateable text/elements: https://github.com/facebook/react/blob/58c895e59cd2822a9a7a67606d5fe7da5e09ba2a/packages/react-dom/src/client/ReactDOMHostConfig.js#L957

So it's no longer warning about a mismatch in `dangerouslySetInnerHTML` but rather that the resulting html mismatches.

It seems like both approaches do not create a helpful error message. In the version on `master` it displays the wrong content of `dangerouslySetInnerHTML` thinking that it's `<p>server</p>` while this content came from the `children` prop. In the version on this PR we no longer see that `dangerouslySetInnerHTML` was part of the issue.

PR #18678, opened 2020-04-20
Add repro for select event inconsistency

philipp-spiess:
We have an issue in the modern select event plugin that is now no longer emitting `onSelect` when using CMD+a to select text on macOS.

The issue is that the `selectionchange` event is only emitted at the document and targets the document so without document level listeners, the modern event system doesn’t know it happens.

@trueadm asked me to make a repro 🙂


philipp-spiess:
Issue was resolved and the test were added in #18680.

PR #18680, opened 2020-04-20
Modern event system: fix selectionchange bug

trueadm:
This fixes the issue found in https://github.com/facebook/react/pull/18678.

Notably, `TOP_SELECTION_CHANGE` or the native event `selectionchange` should always be attached to the `document` when setting up the listener. We cannot attach to the React root – which is why we were seeing issues internally with the `SelectPlugin`.

### DIFF
--- a/packages/react-dom/src/events/plugins/ModernSelectEventPlugin.js
+++ b/packages/react-dom/src/events/plugins/ModernSelectEventPlugin.js
@@ -25,11 +25,22 @@ import getActiveElement from '../../client/getActiveElement';
 import {getNodeFromInstance} from '../../client/ReactDOMComponentTree';
 import {hasSelectionCapabilities} from '../../client/ReactInputSelection';
 import {DOCUMENT_NODE} from '../../shared/HTMLNodeType';
-import {isListeningToAllDependencies} from '../DOMEventListenerMap';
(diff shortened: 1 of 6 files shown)


packages/react-dom/src/events/plugins/ModernSelectEventPlugin.js
```diff
@@ -168,13 +169,19 @@ const SelectEventPlugin = {
...
+      // root.
+      !isListeningToEvent(TOP_SELECTION_CHANGE, doc) ||
+      // If we are handling TOP_SELECTION_CHANGE, then we don't need to
+      // check for the other dependencies, as TOP_SELECTION_CHANGE is only
+      // event attached from the onChange plugin and we don't expose an
+      // onSelectionChange event from React.
+      (topLevelType !== TOP_SELECTION_CHANGE &&
+        !isListeningToEvents(rootTargetDependencies, container))
     ) {
       return null;
```
philipp-spiess, line 186: It feels weird that we have to do this test here and not in other event plugins. 🤷‍♂️

trueadm: We don't do the `isListeningToEvents` logic check in other plugins.

PR #18681, opened 2020-04-20
Warn when textarea switches between controlled and uncontrolled

### LINKED ISSUE #16342: textarea does not show warning when switching from uncontrolled to controlled like inputs do

bl00mber:
resolve #16342

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMTextarea-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMTextarea-test.js
@@ -16,8 +16,8 @@ describe('ReactDOMTextarea', () => {
   let ReactDOM;
   let ReactDOMServer;
   let ReactTestUtils;
-
-  let renderTextarea;
+  let container;
+  let node;
 
   const ReactFeatureFlags = require('shared/ReactFeatureFlags');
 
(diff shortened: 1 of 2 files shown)


gaearon:
What is the concrete problem this is solving? I get that it's inconsistent but I'm trying to remember why we have a warning in the first place.

bl00mber:
To prevent more type errors on the user side? Because controlled can only use strings as value, while uncontrolled can use undefined.

Kirill486:
Arn't they ( ```<input>``` and ```<textarea>``` ) **inherit the same thing**?
Is there a chance that we can achieve the same by moving logic to the ancestor?

bl00mber:
@Kirill486 that's the good point. @gaearon what do you thing about that? Should warning logic be placed to the parent to also handle `<select />` case?

Kirill486:
@bl00mber , we may able manage to answer this question ourselves. The answer seem obvious for me)

bl00mber:
@Kirill486 it doesn't seem _obvious_ to me, because I do not know if this PR is needed in the first place. There are PRs which are not reviewed for years or not reviewed at all so I want to make sure that this PR may have some chance to be reviewed in the future and that's the main purpose of my question.

PR #18684, opened 2020-04-21
Upgrade fbjs-scripts

sebmarkbage:
This version takes into account the NODE_ENV as part of jest cache keys. This avoids flaky tests since we depend on different transforms in prod and dev.

### DIFF
--- a/packages/react-refresh/src/__tests__/ReactFreshBabelPlugin-test.js
+++ b/packages/react-refresh/src/__tests__/ReactFreshBabelPlugin-test.js
@@ -415,7 +415,7 @@ describe('ReactFreshBabelPlugin', () => {
         }
     `,
         {
-          plugins: ['transform-es2015-modules-commonjs'],
+          plugins: ['@babel/transform-modules-commonjs'],
         },
       ),
     ).toMatchSnapshot();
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -2430,104 +2618,6 @@ babel-eslint@^9.0.0:
     eslint-scope "3.7.1"
     eslint-visitor-keys "^1.0.0"
 
-babel-generator@^6.26.0:
```
sebmarkbage, line 2433: lol. Don't need these anymore.

packages/react-refresh/src/__tests__/ReactFreshBabelPlugin-test.js
```diff
@@ -415,7 +415,7 @@ describe('ReactFreshBabelPlugin', () => {
         }
     `,
         {
-          plugins: ['transform-es2015-modules-commonjs'],
```
sebmarkbage, line 418: @gaearon This is the Babel 6 version which was only accidentally available in our repo. This now uses the Babel 7 transform. Not sure what this is testing.

gaearon: I think this just makes sure the transform doesn't break it in some way. Makes sense to use the recent one.

PR #18685, opened 2020-04-21
Avoid passing custom stacks to console.error

sebmarkbage:
The detection if one of the args contains a stack was broken in the new stack format. Unfortunately the toWarnDev helper that was meant to detect that oversight used the same check.

This stops detecting stacks and instead always appends a stack.

Many of the custom stacks were actually unnecessary because they had the correct execution context already. For the ones that might be called outside the render phase, I set the debug fiber right before calling the warning to set the right context.

That way we never need to pass a custom stack as an argument.


packages/react-dom/src/__tests__/ReactComponent-test.js
```diff
@@ -470,20 +461,11 @@ describe('ReactComponent', () => {
...
-    } catch (e) {
-      ex = e;
-    }
-    expect(ex).toBeDefined();
-    expect(normalizeCodeLocInfo(ex.message)).toBe(
-      'Objects are not valid as a React child (found: object with keys {x, y, z}).' +
-        (__DEV__
-          ? ' If you meant to render a collection of children, use ' +
-            'an array instead.' +
-            '\n    in div (at **)'
```
gaearon, line 485: This one will likely regress full-screen errors in CRA and such, where the extra information ("the above error occurred") isn't shown. That's a problem by itself though.

sebmarkbage: How do those work today? Ours is using a fork so seems like some missing layering.

If implemented as error boundaries, the information is in the componentStack but I assume it needs a global hook?

gaearon: Yeah needs something global.

bvaughn:
This looks like a nice change!

PR #18689, opened 2020-04-21
Remove ReactDOM.useEvent and associated types+tests

trueadm:
This PR removes the `ReactDOM.unstable_useEvent` API from the codebase. We are close to finding a lower level primitive to creating events in React without needing to couple the API to hooks only and that exploration might bring back some of the logic we had here, but ideally we'd probably revisit things again without bringing in as much of the overhead we had here. We'll likely bring back the `useFocusWithin` and `useFocus` APIs once we have the alternative API available too.


packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -1083,12 +1083,11 @@ describe('DOMModernPluginEventSystem', () => {
           }
         });
 
-        describe('ReactDOM.useEvent', () => {
+        describe.skip('ReactDOM.useEvent', () => {
```
trueadm, line 1086: Intentionally skipping, as we'll bring these back with the new API and to avoid test churn on good tests I thought we'd leave them in for now.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -1020,9 +1019,7 @@ function commitUnmount(
       if (enableDeprecatedFlareAPI) {
         unmountDeprecatedResponderListeners(current);
       }
-      if (enableDeprecatedFlareAPI || enableUseEventAPI) {
-        beforeRemoveInstance(current.stateNode);
-      }
+      beforeRemoveInstance(current.stateNode);
```
gaearon, line 1022: Don't we need this for `enableDeprecatedFlareAPI` still?

packages/react-reconciler/src/ReactFiberCommitWork.old.js
```diff
@@ -1020,9 +1019,7 @@ function commitUnmount(
       if (enableDeprecatedFlareAPI) {
         unmountDeprecatedResponderListeners(current);
       }
-      if (enableDeprecatedFlareAPI || enableUseEventAPI) {
```
gaearon, line 1023: Same question, don't we still need `enableDeprecatedFlareAPI`?

packages/react-reconciler/src/ReactFiberCommitWork.new.js
trueadm: We don't need the flag anymore. The code that Flare used (beforeblur stuff) was moved out, but I forgot to remove this flag.

PR #18690, opened 2020-04-21
Clear fiber field from host instance on unmount

trueadm:
Given we're still seeing memory leaks internally and that we already traverse each host component when we detach its fiber – let's also clear the fiber field from the host instance. I did some quick checks and didn't notice any performance runtime implications, in fact I noticed that overall memory consumed was lower from this change (testing internally)!

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponentTree.js
+++ b/packages/react-dom/src/client/ReactDOMComponentTree.js
@@ -40,6 +40,12 @@ export function precacheFiberNode(
   (node: any)[internalInstanceKey] = hostInst;
 }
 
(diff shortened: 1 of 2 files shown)


gaearon:
How do you feel about only doing this for node with refs? Like Seb suggested in https://github.com/facebook/react/issues/16087:

>What we would do here is special case DOM nodes with refs on them, and always detach their back pointer to the React Fiber, if it was ever fully mounted. We currently traverse these trees anyway when they get deleted. We want to stop doing this for most things but for nodes with a ref it seems minor to special case since they typically need to be invoked with null anyway.

PR #18697, opened 2020-04-22
Temporarily Remove DebugTracing from the New Reconciler

sebmarkbage:
Sorry @bvaughn and @gaearon. 😢We can add it back.

Some of the concepts are changing. E.g. getting the current priority level isn't as obviously going to be in the same place. Priority field on update isn't useful anymore since the lanes convey more info. This code is also in all the places moving around. It'll be easier to just add this back than to try to keep them each time they move.

### DIFF
--- a/packages/react-reconciler/src/ReactUpdateQueue.new.js
+++ b/packages/react-reconciler/src/ReactUpdateQueue.new.js
@@ -121,9 +121,6 @@ export type Update<State> = {|
   callback: (() => mixed) | null,
 
   next: Update<State> | null,
-
-  // DEV only
-  priority?: ReactPriorityLevel,
 |};
 
 type SharedQueue<State> = {|
@@ -207,9 +204,6 @@ export function createUpdate(
 
     next: null,
   };
-  if (__DEV__) {
-    update.priority = getCurrentPriorityLevel();
-  }
   return update;
 }
 
--- a/packages/react-reconciler/src/ReactUpdateQueue.new.js
+++ b/packages/react-reconciler/src/ReactUpdateQueue.new.js
@@ -87,7 +87,6 @@
 import type {Fiber} from './ReactInternalTypes';
 import type {ExpirationTime} from './ReactFiberExpirationTime.new';
 import type {SuspenseConfig} from './ReactFiberSuspenseConfig';
-import type {ReactPriorityLevel} from './ReactInternalTypes';
 
 import {NoWork, Sync} from './ReactFiberExpirationTime.new';
 import {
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactUpdateQueue.new.js
```diff
@@ -207,9 +202,6 @@ export function createUpdate(
 
     next: null,
   };
-  if (__DEV__) {
-    update.priority = getCurrentPriorityLevel();
```
sebmarkbage, line 211: btw, are we actually using these in DevTools? I thought we added them for DevTools but I couldn't find any usage of them.

acdlite: It was added for the Suspense priority warning, which we later removed

acdlite: We don't need it in the new system because the priority can be determined from the lane without having to infer.

acdlite:
Assuming the tests pass. I assume they will because they are already gated on the flag.

packages/shared/forks/ReactFeatureFlags.www-dynamic.js
```diff
@@ -19,7 +19,7 @@ export const disableInputAttributeSyncing = __VARIANT__;
 export const enableFilterEmptyStringAttributesDOM = __VARIANT__;
 export const enableModernEventSystem = __VARIANT__;
 export const enableLegacyFBSupport = __VARIANT__;
-export const enableDebugTracing = __VARIANT__;
+export const enableDebugTracing = !__VARIANT__;
```
acdlite, line 22: What's the purpose of this?

acdlite: Oh nvm :)

acdlite: Funny that means the behavior in the old reconciler wasn't being tested :D

sebmarkbage: The new reconciler is enabled when the variant is true. This feature doesn't work so the tests don't pass. This ensures it's enabled in the other variant instead.

packages/react-reconciler/src/ReactUpdateQueue.new.js
bvaughn: I think the only place DevTools surfaces priority level info is in the Profiler, per commit:
https://github.com/facebook/react/blob/571f5ad2d695e6373ffbd8e86b27c11dad3d22cf/packages/react-reconciler/src/ReactFiberDevToolsHook.old.js#L85-L102

Maybe that's what you were thinking of?

PR #18708, opened 2020-04-23
Disable prepareStackTrace while we're generating stacks

sebmarkbage:
This could be used to do custom formatting of the stack trace in a way that isn't compatible with how we use it. So we disable it while we use it.

In theory we could call this ourselves with the result of our stack. It would be a lot of extra production code though. My personal opinion is that this should always be done server side instead of on the client. This is not infrastructure that is worth its code size.

We could expose a custom parser that converts it and passes it through prepareStackTrace as structured data. That way it's external and doesn't have to be built-in to React.

### DIFF
--- a/packages/shared/ReactComponentStackFrame.js
+++ b/packages/shared/ReactComponentStackFrame.js
@@ -80,6 +80,9 @@ export function describeNativeComponentFrame(
   let control;
 
   reentry = true;
+  const previousPrepareStackTrace = Error.prepareStackTrace;
+  // $FlowFixMe It does accept undefined.
+  Error.prepareStackTrace = undefined;
   let previousDispatcher;
   if (__DEV__) {
     previousDispatcher = ReactCurrentDispatcher.current;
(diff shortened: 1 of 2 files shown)


sebmarkbage:
@bvaughn Should we port this to DevTools too?

packages/shared/ReactComponentStackFrame.js
```diff
@@ -182,6 +185,7 @@ export function describeNativeComponentFrame(
     reentry = false;
     if (__DEV__) {
       ReactCurrentDispatcher.current = previousDispatcher;
+      Error.prepareStackTrace = previousPrepareStackTrace;
```
gaearon, line 188: Did you mean to put this outside of DEV?

bvaughn:
> @bvaughn Should we port this to DevTools too?

Yeah seems worth adding.

> We could expose a custom parser that converts it and passes it through prepareStackTrace as structured data. That way it's external and doesn't have to be built-in to React.

I'm not sure I understand this suggestion? We don't have enough info in the string (and so a custom parser wouldn't have enough info) to fully reconstruct a `CallSight` object.

PR #18710, opened 2020-04-23
Remove flushDiscreteUpdates, discreteUpdates and syncUpdates

sebmarkbage:
We're simplifying discrete by merging it with sync. As such `flushSync()` can be used to flush already pending discrete work. `flushSync(() => ...)` replaces discreteUpdates. syncUpdates could in theory be used to enter the discrete phase in a non-discrete event but might as well use `flushSync(() => ...)` then to simplify the API surface area.

Currently `flushSync` errors if nested. We might need to relax that and always flush on the inner one. Let's see if we hit it.

I'm going to remove `flushControlled` as well. I need to see if I can just use `flushSync` for that. In Draft's case I might be able to just let the new heuristics take over and make `flushControlled` a noop.


acdlite:
Is it intentional that you didn't actually remove the `flushDiscreteUpdates` implementation? Is the idea to remove the API only and then clean up the implementation later?

sebmarkbage:
The implementation is still used by the event system. I'll need to implement the new stuff and then remove it.

PR #18711, opened 2020-04-23
Always wrap Suspense children

acdlite:
This is stacked on top of the work I'm doing in my Lanes refactor branch. I would ideally like to land all of the previous steps, too; however, it's not a hard requirement. I can decouple them if needed. I'm open to discussion on which parts we should land now versus later.

But I would like to land cd7e394 in some form regardless, hence the separate PR.

Descriptions are in each commit message.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -1737,7 +1737,6 @@ function updateSuspenseComponent(
   workInProgress,
   renderExpirationTime,
 ) {
-  const mode = workInProgress.mode;
   const nextProps = workInProgress.pendingProps;
 
   // This is used by DevTools to force a boundary to suspend.
@@ -1749,7 +1748,7 @@ function updateSuspenseComponent(
 
   let suspenseContext: SuspenseContext = suspenseStackCursor.current;
 
-  let nextDidTimeout = false;
+  let showFallback = false;
   const didSuspend = (workInProgress.effectTag & DidCapture) !== NoEffect;
 
   if (
@@ -1763,7 +1762,7 @@ function updateSuspenseComponent(
   ) {
     // Something in this boundary's subtree already suspended. Switch to
     // rendering the fallback children.
-    nextDidTimeout = true;
+    showFallback = true;
     workInProgress.effectTag &= ~DidCapture;
   } else {
     // Attempting the main content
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/__tests__/ReactSuspensePlaceholder-test.internal.js
```diff
@@ -403,10 +403,21 @@ describe('ReactSuspensePlaceholder', () => {
...
         // The suspense update should only show the "Loading..." Fallback.
-        // Both durations should include 10ms spent rendering Fallback
-        // plus the 8ms rendering the (hidden) components.
-        expect(onRender.mock.calls[1][2]).toBe(18);
-        expect(onRender.mock.calls[1][3]).toBe(18);
+        // The actual duration should include 10ms spent rendering Fallback,
+        // plus the 8ms render all of the hidden, suspended subtree.
+        // Note from Andrew to Brian: I don't fully understand why this one
+        // diverges, but I checked and it matches the times we get when
+        // we run this same test in Concurrent Mode.
```
acdlite, line 410: cc: @bvaughn Does this look ok? I don't remember if the legacy behavior diverged intentionally or if it was an implementation compromise. Naively, it seems fine for it to match the numbers in Concurrent Mode but I trust your judgment more than mine.

bvaughn: Oh man. This reaches back a while...

I *think* the 18/15 numbers are more accurate, so if the *new* behavior also produces those for sync roots- that seems nice.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2157,44 +1955,276 @@ function updateSuspenseComponent(
...
+      // final amounts. This seems counterintuitive, since we're intentionally
+      // not measuring part of the render phase, but this makes it match what we
+      // do in Concurrent Mode.
+      primaryChildFragment.actualDuration = 0;
+      primaryChildFragment.actualStartTime = -1;
+      primaryChildFragment.selfBaseDuration = 0;
+      primaryChildFragment.treeBaseDuration = 0;
+    }
+
+    // Transfer the effects from the primary tree
```
sebmarkbage, line 2021: Why do we have to transfer the effects now but didn't before? I'd expect them to be transferred automatically after we complete, since the node is a child.

acdlite: The comment isn't quite right. It's to remove the current fallback fiber from the deletion list, which was added in the first pass. However, that's only relevant to updates, and this is a mount path, so you're right that I can remove this here. But it is necessary in the `updateSuspenseFallbackChildren` path.

What we used to do is only add the fallback fiber to the effect list once we get to the complete phase: https://github.com/facebook/react/pull/18711/files/d4ac75f601ce71d73893e895892fa6a3d1dbbaf4..cd7e3948d3a1159ec06ae1a326c0ea2547656c6c#diff-c29bccbec7fe889841f1b254f08d97c7L891-L910

I opted for doing it here instead since this means we don't have to do anything special in Concurrent Mode. It's only weird in legacy mode because we don't create a fresh work-in-progress fiber for the second pass, which would reset the effect list. Similarly, we don't reuse effects from the first pass.

Another approach which I just thought of would be to clear the Deletion bit from the tag but keep it in the list. Slightly less code but slightly more computation later.

I'll delete this from the mount path, leave the update path as-is, and fix the comment.

PR #18713, opened 2020-04-23
ReactDOMEventListener: clean up module

trueadm:
This PR tidies up `ReactDOMEventListener` and moves the different event listener approaches into their respective modules. This makes it easier to remove event systems in the future without breaking something else. Along the way, I noticed that we weren't applying the modern event system flag in the plugin tests properly, so this fixes that too.

### DIFF
--- a/packages/react-dom/src/events/plugins/__tests__/ModernChangeEventPlugin-test.internal.js
+++ b/packages/react-dom/src/events/plugins/__tests__/ModernChangeEventPlugin-test.internal.js
@@ -9,9 +9,9 @@
 
 'use strict';
 
-let React = require('react');
-let ReactDOM = require('react-dom');
-let TestUtils = require('react-dom/test-utils');
+let React;
+let ReactDOM;
+let TestUtils;
 let ReactFeatureFlags;
 let Scheduler;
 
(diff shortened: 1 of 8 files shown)


packages/react-dom/src/events/plugins/__tests__/ModernChangeEventPlugin-test.internal.js
```diff
@@ -53,7 +54,11 @@ describe('ChangeEventPlugin', () => {
         postMessageCallback(postMessageEvent);
       }
     };
-    jest.resetModules();
+    ReactFeatureFlags.enableModernEventSystem = true;
```
bvaughn, line 57: I think... you don't need this test to be `.internal` anymore since the `enableModernEventSystem` flag is marked as *variant* in `ReactFeatureFlags.www-dynamic`.

I think you should be able to remove the `.internal` suffix on this (and the below test) and just use the new pragma comment:
```js
// @gate enableModernEventSystem
```

trueadm: I tried this locally just now, for some reason some tests in files still break with this change. Namely `EnterLeave`. To make sure I'm sane, I break the old legacy event system whilst running the change, so I can see if we are using the modern system or the legacy – and it seems we are still using the legacy with `// @gate enableModernEventSystem`

I'd rather tackle why that might be the case in another PR. :)

PR #18722, opened 2020-04-23
[ESLint] Extend isHook to recognize those under PascalCase's namespace

cyan33:
## Summary

Context and discussion: https://fburl.com/ppb00y2s

TLDR: A few teams have the convention to put hooks under a namespace and the current implementation doesn't catch hooks that are written this way.

```
// is hook
useSomething();

// is not hook => doesn't recognize
Foo.useSomething();
```

This prevented us from catching the true positives and started to cause issues, like T65929958. We want to extend the definition of `isHook` so that we can safely catch those that are used:

- on top level / not in React component or other hooks
- called in class component
- being conditionally called

We intentionally don't include the namespace that are named in "camelCase" because there're a lot of false positive, eg:

```
jest.useFakeTimers
```

This change requires us to move a few test cases that are used to be regression tests and _were_ valid to invalid.

## Test Plan

```
# under react
> yarn test ESLintRulesOfHooks-test.js
```

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
@@ -380,6 +380,8 @@ const tests = {
     },
     {
       code: `
+        // This is a false positive (it's valid) that unfortunately 
+        // we cannot avoid. Prefer to rename it to not start with "use"
         class Foo extends Component {
           render() {
             if (cond) {
@@ -401,17 +403,6 @@ const tests = {
         }
       `,
       errors: [conditionalError('Namespace.useConditionalHook')],
-    },
-    {
-      code: `
-        // Extending the isHook check to [A-Z].*\.use
-        // to not let it be called on top level
-        const History = require('history-2.1.2');
-        const browserHistory = History.useBasename(History.createHistory)({
-          basename: '/',
-        });
-      `,
-      errors: [topLevelError('History.useBasename')],
     },
     {
       code: `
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
```diff
@@ -392,6 +351,68 @@ const tests = {
...
+        }
+      `,
+      errors: [classError('This.useHook'), classError('Super.useHook')],
+    },
+    {
+      code: `
+        class Foo extends Component {
+          render() {
+            if (cond) {
+              FooStore.useFeatureFlag();
```
gaearon, line 386: This is a confusing example because it was supposed to show _why_ we don't want to change this. I guess you could leave a comment saying this is an unfortunate false positive we can't avoid.

packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
```diff
@@ -392,6 +351,68 @@ const tests = {
...
+            Namespace.useConditionalHook();
+          }
+        }
+      `,
+      errors: [conditionalError('Namespace.useConditionalHook')],
+    },
+    {
+      code: `
+        // Extending the isHook check to [A-Z].*\.use
+        // to not let it be called on top level
```
gaearon, line 408: I don't understand this comment. I think you can just drop this example now because it is also a false positive. Or you could come up with another way to avoid flagging it (e.g. lowercase `history` or rename Hook-like function at destructuring) and move it back to `valid`.

PR #18723, opened 2020-04-23
Modern Event System: fix fiber ancestor issue with invalid containers

trueadm:
This PR addresses an issue that we found internally, where events may not fiber if we do some crazy ordering of containers, so that the parent contains the child container, that is a parent node but the opposite in terms of fiber order. I added a regression test to validate this and can confirm this fixes it internally.

### DIFF
--- a/packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
+++ b/packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
@@ -336,6 +336,68 @@ describe('DOMModernPluginEventSystem', () => {
           expect(log[5]).toEqual(['bubble', buttonElement]);
         });
 
+        it('handle propagation of click events between disjointed comment roots #3', () => {
+          const buttonRef = React.createRef();
+          const divRef = React.createRef();
+          const childRef = React.createRef();
+          const log = [];
+          const onClick = jest.fn(e => log.push(['bubble', e.currentTarget]));
+          const onClickCapture = jest.fn(e =>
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/events/__tests__/DOMModernPluginEventSystem-test.internal.js
```diff
@@ -336,6 +336,68 @@ describe('DOMModernPluginEventSystem', () => {
...
+          dispatchClickEvent(divElement);
+          expect(onClick).toHaveBeenCalledTimes(3);
+          expect(onClickCapture).toHaveBeenCalledTimes(3);
+          expect(log[2]).toEqual(['capture', buttonElement]);
+          expect(log[3]).toEqual(['bubble', buttonElement]);
+          expect(log[4]).toEqual(['capture', divElement]);
+          expect(log[5]).toEqual(['bubble', divElement]);
+
+          document.body.removeChild(containerParent);
+          document.body.appendChild(container);
```
sebmarkbage, line 398: I think you meant removeChild.

trueadm: Nope, I want to reparent the container, so it can be removed again later.

trueadm:
@sebmarkbage Actually, maybe this change is quite right. As theres a failing test in `ReactDOMServerPartialHydration` when the modern event flag is enabled with this change. Any ideas?

sebmarkbage:
It turned out to not be this. It's simply that they're moving React managed DOM node arounds manually which we don't support.

PR #18729, opened 2020-04-24
Move hide/unhide logic to new Offscreen fiber type

acdlite:
This adds a new internal Fiber type, Offscreen, and uses it in the Suspense implementation to hide/unhide children when switching between the primary and fallback trees. Most of the logic for hiding/unhiding already existed, so this PR mostly just moves it to the new type. Eventually, the idea is we'll expose an `<Offscreen />` component type (name TBD) that uses the same implementation. I have not added the public type yet.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -23,6 +23,7 @@ import type {UpdateQueue} from './ReactUpdateQueue.new';
 import type {FunctionComponentUpdateQueue} from './ReactFiberHooks.new';
 import type {Wakeable} from 'shared/ReactTypes';
 import type {ReactPriorityLevel} from './ReactInternalTypes';
+import type {OffscreenState} from './ReactFiberOffscreenComponent';
 
 import {unstable_wrap as Schedule_tracing_wrap} from 'scheduler/tracing';
 import {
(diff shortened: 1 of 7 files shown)


packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -1731,18 +1739,22 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
...
+    if (supportsMutation) {
+      // Hide the Offscreen component that contains the primary children. TODO:
+      // Ideally, this effect would have been scheduled on the Offscreen fiber
+      // itself. That's how unhiding works: the Offscreen component schedules an
+      // effect on itself. However, in this case, the component didn't complete,
+      // so the fiber was never added to the effect list in the normal path. We
+      // could have appended it to the effect list in the Suspense component's
+      // second pass, but doing it this way is less complicated. This would be
+      // simpler if we got rid of the effect list and traversed the tree, like
+      // we're planning to do.
```
sebmarkbage, line 1754: I think the traversal of the tree would bail out on effect tags so it wouldn't go down this node for the same reason. This simpler solution seems to apply to both cases though.

acdlite:
Yeah I think it's right. I didn't find any places where we checked the tag of the primary fragment.

acdlite:
Will merge manually so Luna gets credit for her commit

Issue #18183, opened 2020-02-28
useMutableSource and hydration


bvaughn:
Follow up to PR #18000 and RFC https://github.com/reactjs/rfcs/pull/147

The new `useMutableSource` hook will need at least one additional API to be able to support server rendering and hydration. This API would likely be on the React root (the object returned by `createRoot`) and would enable mutable sources to have their versions eagerly captured before hydration begins.

##### How is the version used?

Currently a work-in-progress version of each mutable source is stored on the source itself. This version enables us to [avoid tearing before a source has been subscribed to](https://github.com/bvaughn/rfcs/blob/useMutableSource/text/0000-use-mutable-source.md#reading-from-a-source-before-subscribing). This version number is lazily populated (the first time a source is read during a given render).

##### How will the version be used during hydration?

In the case of server rendering, this version will need to be eagerly populated for every source so that we can detect tearing between the version of the source used for the server respond, and the version we will eventually read from while hydrating on the client.

##### How will this work?

One way to do this would be to store an array on each React root of mutable source and version number pairs. Each time we start (or resume) hydration, we can iterate through this array and use it to initialize the work-in-progress version for each source.

Once all outstanding hydration work is finished, we can clear the array. (Although this will require an additional change as we do not currently have a way to detect that we've finished all pending hydration work.)

jeremy-deutsch:
How would the `registerMutableSourceForHydration` API in #18771 work with libraries that create mutable sources during render (ex. [this](https://codesandbox.io/s/react-redux-usemutablesource-eyxoe) or [this](https://codesandbox.io/s/blissful-babbage-24uu3))?

(posting here so as not to clutter the PR)

bvaughn:
This API wouldn't support a create-during-render pattern. You'd need to create the source at the module level. Otherwise a mutation might happen between the time when root hydration starts and when a particular component that reads from the source is hydrated.

I don't see an alternative to this though, without e.g. tightening constraints on "version" to be serializable.

PR #18730, opened 2020-04-24
Root API should clear non-empty roots before mounting

bvaughn:
Legacy render-into-subtree API removes children from a container before rendering into it:
https://github.com/facebook/react/blob/4b02b6611141255a8e8b6d8a997e87c865f06626/packages/react-dom/src/client/ReactDOMLegacy.js#L113-L138

The `createRoot` API **did not do this**. Instead it left previous children around in the document forever. This commit brings the `createRoot` API into alignment with `render` - with a small exception. Pre-existing container children don't get cleared **until the commit phase**.

Note that this PR introduces a new DOM host config method, `clearContainer`, that may be used by the `useMutableSource` hook as well in the case of hydration (see #18183)

### DIFF
--- a/packages/react-art/src/ReactARTHostConfig.js
+++ b/packages/react-art/src/ReactARTHostConfig.js
@@ -427,12 +427,8 @@ export function unhideTextInstance(textInstance, text): void {
   // Noop
 }
 
-export function clearContainer(container: Container): void {
-  // TODO This doesn't work for anything other than SVG.
-  // Is that okay?
-  while (container.lastChild != null) {
-    container.lastChild.eject();
-  }
+export function clearContainer(container) {
+  // TODO Implement this
 }
 
 export function DEPRECATED_mountResponderInstance(
--- a/packages/react-art/src/ReactARTHostConfig.js
+++ b/packages/react-art/src/ReactARTHostConfig.js
@@ -427,6 +427,14 @@ export function unhideTextInstance(textInstance, text): void {
   // Noop
 }
 
+export function clearContainer(container: Container): void {
+  // TODO This doesn't work for anything other than SVG.
+  // Is that okay?
+  while (container.lastChild != null) {
+    container.lastChild.eject();
+  }
+}
+
 export function DEPRECATED_mountResponderInstance(
   responder: ReactEventResponder<any, any>,
   responderInstance: ReactEventResponderInstance<any, any>,
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -634,6 +634,17 @@ export function unhideTextInstance(
   textInstance.nodeValue = text;
 }
 
+export function clearContainer(container: Container): void {
+  if (container.nodeType === ELEMENT_NODE) {
+    ((container: any): Element).innerHTML = '';
+  } else if (container.nodeType === DOCUMENT_NODE) {
+    const documentElement = ((container: any): Document).documentElement;
+    if (documentElement != null) {
+      documentElement.innerHTML = '';
+    }
+  }
+}
+
 // -------------------
 //     Hydration
 // -------------------
--- a/packages/react-dom/src/client/ReactDOMHostConfig.js
+++ b/packages/react-dom/src/client/ReactDOMHostConfig.js
@@ -636,11 +636,11 @@ export function unhideTextInstance(
 
 export function clearContainer(container: Container): void {
   if (container.nodeType === ELEMENT_NODE) {
-    ((container: any): Element).innerHTML = '';
+    ((container: any): Element).textContent = '';
   } else if (container.nodeType === DOCUMENT_NODE) {
-    const documentElement = ((container: any): Document).documentElement;
-    if (documentElement != null) {
-      documentElement.innerHTML = '';
+    const body = ((container: any): Document).body;
+    if (body != null) {
+      body.textContent = '';
     }
   }
 }
--- a/packages/react-noop-renderer/src/createReactNoop.js
+++ b/packages/react-noop-renderer/src/createReactNoop.js
@@ -155,6 +155,10 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
     insertInContainerOrInstanceBefore(parentInstance, child, beforeChild);
   }
 
+  function clearContainer(container: Container): void {
+    container.children.splice(0);
+  }
+
   function removeChildFromContainerOrInstance(
     parentInstance: Container | Instance,
     child: Instance | TextInstance,
@@ -502,6 +506,7 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
         insertInContainerBefore,
         removeChild,
         removeChildFromContainer,
+        clearContainer,
 
         hideInstance(instance: Instance): void {
           instance.hidden = true;
@@ -531,6 +536,7 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
         supportsPersistence: true,
 
         cloneInstance,
+        clearContainer,
 
         createContainerChildSet(
           container: Container,
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -70,6 +70,7 @@ import {
   Snapshot,
   Update,
   Passive,
+  ClearContainer,
 } from './ReactSideEffectTags';
 import getComponentName from 'shared/getComponentName';
 import invariant from 'shared/invariant';
@@ -113,6 +114,7 @@ import {
   commitHydratedContainer,
   commitHydratedSuspenseInstance,
   beforeRemoveInstance,
+  clearContainer,
 } from './ReactFiberHostConfig';
 import {
   captureCommitPhaseError,
@@ -295,7 +297,15 @@ function commitBeforeMutationLifeCycles(
       }
       return;
     }
-    case HostRoot:
+    case HostRoot: {
+      if (supportsMutation) {
+        if (finishedWork.effectTag & ClearContainer) {
+          const root = finishedWork.stateNode;
+          clearContainer(root.containerInfo);
+        }
+      }
+      return;
+    }
     case HostComponent:
     case HostText:
     case HostPortal:
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -70,7 +70,6 @@ import {
   Snapshot,
   Update,
   Passive,
-  ClearContainer,
 } from './ReactSideEffectTags';
 import getComponentName from 'shared/getComponentName';
 import invariant from 'shared/invariant';
@@ -299,7 +298,7 @@ function commitBeforeMutationLifeCycles(
     }
     case HostRoot: {
       if (supportsMutation) {
-        if (finishedWork.effectTag & ClearContainer) {
+        if (finishedWork.effectTag & Snapshot) {
           const root = finishedWork.stateNode;
           clearContainer(root.containerInfo);
         }
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
@@ -123,7 +123,6 @@ import {
   HostEffectMask,
   Hydrating,
   HydratingAndUpdate,
-  ClearContainer,
 } from './ReactSideEffectTags';
 import {
   NoWork,
@@ -2168,7 +2167,7 @@ function commitBeforeMutationEffects() {
       beforeActiveInstanceBlur();
     }
     const effectTag = nextEffect.effectTag;
-    if ((effectTag & (Snapshot | ClearContainer)) !== NoEffect) {
+    if ((effectTag & Snapshot) !== NoEffect) {
       setCurrentDebugFiberInDEV(nextEffect);
 
       const current = nextEffect.alternate;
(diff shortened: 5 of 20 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.old.js
```diff
@@ -1912,6 +1913,15 @@ function commitRootImpl(root, renderPriorityLevel) {
...
   }
 
+  if (root.clearContainerBeforeMount) {
+    // We are about to mount into a container that previous contained non-React elements.
+    // We should clear the previous contents before beginning.
+    // This mimics legacy render into subtree behavior in a way that is safe for concurrent mode.
+    // (It doesn't result in multiple obsevable mutations.)
+    root.clearContainerBeforeMount = false;
+    clearContainer(root.containerInfo);
+  }
```
bvaughn, line 1923: Feels a little goofy to check for this here. Any better suggestions?

sebmarkbage: I was going to say maybe we don't need to make it an option. Could just be the default like how ReactDOM.render works.

sebmarkbage: I was confused.

bvaughn:
Seb and I chatted about this out of band yesterday. He suggested trying a different approach- putting an effect on the `HostRoot` during the complete phase if `current.child` is null (meaning either a *mount* or an update after a previous render with no React children).

This *works* but is also a lot broader than my initial approach. Several tests failing. Going to dig in and see if the failures are *improvements* or road blocks for the approach.

bvaughn:
One test case this breaks is in `ReactServerRenderingHydration-test`:

```js
it('Suspense + hydration in legacy mode with no fallback', () => {
  element.innerHTML = '<div>Hello World</div>';
  const div = element.firstChild;
  const ref = React.createRef();
  ReactDOM.hydrate(
    <React.Suspense>
      <div ref={ref}>Hello World</div>
    </React.Suspense>,
    element,
  );

  // Because this didn't have a fallback, it was hydrated as if it's
  // not a Suspense boundary.
  expect(ref.current).toBe(div);
  expect(element.innerHTML).toBe('<div>Hello World</div>');
});
```

In this case, `popHydrationState()` does not think it's hydrating because `fiber` is the `HostRoot` but `hydrationParentFiber` points to the `Suspense` node. So as a result, we clear the container.

sebmarkbage:
Can you push the code for the new approach?

gaearon:
Last time we discussed this, I think the conclusion was that maybe it shouldn't clear *until commit*, at which point we should replace. So that you can leave an HTML shim there that would be replaced. But maybe this isn't that useful.

gaearon:
I guess that's what you're doing.

bvaughn:
> Last time we discussed this, I think the conclusion was that maybe it shouldn't clear until commit, at which point we should replace. So that you can leave an HTML shim there that would be replaced. But maybe this isn't that useful.

That's what this PR is implementing. (And what my local commit that I haven't pushed yet is implementing in a different way, via a new effect type).

bvaughn:
> Can you push the code for the new approach?

@sebmarkbage Sure. It breaks lots of persistence tests too which I haven't had the chance to dig into yet.

bvaughn:
Okay, I think every test is passing now except for the one I mentioned [above](https://github.com/facebook/react/pull/18730#issuecomment-619411233).

bvaughn:
> Can you reuse the Update effectTag?

`Update` effect would happen too late, right? The container would just end up clearing all of its newly created (by React) contents.

Unless you're suggesting I call `commitBeforeMutationLifeCycles` with `Update` effects too, and just handle the `HostRoot` there. That seems sketchy though because we *also* handle `Update` for `HostRoot` in `commitWork` --- so we would have to switch based on `root.hydrate` which feels pretty fragile.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -293,7 +295,15 @@ function commitBeforeMutationLifeCycles(
       }
       return;
     }
-    case HostRoot:
+    case HostRoot: {
+      if (supportsMutation) {
+        if (finishedWork.effectTag & ClearContainer) {
+          const root = finishedWork.stateNode;
+          clearContainer(root.containerInfo);
```
sebmarkbage, line 302: So I get that the problem is that if you do this in the mutation phase you’ll delete the ones we just added. However if we do it in the snapshot, then these snapshot life cycles and beforeblur observe an already mutated tree and can’t read from the true previous state and it will likely trigger an additional layout pass if they read the layout.

I wonder if we should instead just have the clearContainer traverse top down until it hits the first react managed node we know about.

Alternatively we could special case this and call it between the snapshot and mutation phases. However then that gets tricky if we also want to do this for portals.

sebmarkbage: I guess it would be after the snapshots since the root is last but it wouldn’t work for portals.

bvaughn: Yeah, special casing also felt kind of dirty.

As for handling it during snapshot, I mentioned that in my comment above. We *could* do that but it feels fragile.

sebmarkbage:
I think we can’t do this for Portals because people (FB) are using document.body as the container for many Portals. This is kind of unfortunate for a number of event related things and selective hydration but it’s pretty common.

So that settles that.

bvaughn:
It's unclear to me after your recent comment if you're still a fan of repurposing the `Update` tag to include this too, given the concern I raised. 😅

packages/react-reconciler/src/ReactFiberCommitWork.new.js
sebmarkbage: Can we at least just use the Snapshot effect tag for this case?

I want to use the tags as a hint which children to traverse into when we switch to the traversal for effects.

So Snapshot would be the one we look for to enter this path.

bvaughn: Ah. That sounds promising. Let me give it a shot.

bvaughn: Yes, that will work nicely and avoid the extra effect tag. Good suggestion.

bvaughn:
CI failure is expected still until I figure out what to do about that one failing test.

Probably going to step away for a bit and go running or something 😄

bvaughn:
Okay. Tests are passing.

Still an open question about React Native and React ART configs.

Think it's ready for review/feedback.

bvaughn:
DevTools CI failure is unrelated to this PR. Looks like a transient issue.

trueadm:
Was @sebmarkbage's comment above addressed?

> I think we can’t do this for Portals because people (FB) are using document.body as the container for many Portals.

bvaughn:
> Was @sebmarkbage's comment above addressed?
> 
> > I think we can’t do this for Portals because people (FB) are using document.body as the container for many Portals.

Seb's comment doesn't require any action, does it? This change only applies to `HostRoot`s. I don't do anything new for `HostPortal`s.

trueadm:
Thanks for clarifying!

Does this also cover the case where we mount to a comment node? I know we do it a lot internally, but I wasn't sure if we wanted to keep supporting it in the future.

bvaughn:
I only explicitly check for `ELEMENT_NODE` and `DOCUMENT_NODE`.

I don't think there would be anything to do here in the case of a comment node, since comment nodes can't contain children.

packages/react-art/src/ReactARTHostConfig.js
```diff
@@ -427,6 +427,14 @@ export function unhideTextInstance(textInstance, text): void {
   // Noop
 }
 
+export function clearContainer(container: Container): void {
+  // TODO This doesn't work for anything other than SVG.
+  // Is that okay?
```
bvaughn, line 432: Would be interested to hear @sebmarkbage's thoughts on the priority of this.

packages/react-test-renderer/src/ReactTestHostConfig.js
```diff
@@ -126,6 +126,10 @@ export function removeChild(
   parentInstance.children.splice(index, 1);
 }
 
+export function clearContainer(container: Container): void {
+  container.children.splice(0);
```
trueadm, line 130: I believe a faster way is to do:

`container.children.textContent = ''`

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -634,6 +634,17 @@ export function unhideTextInstance(
   textInstance.nodeValue = text;
 }
 
+export function clearContainer(container: Container): void {
+  if (container.nodeType === ELEMENT_NODE) {
+    ((container: any): Element).innerHTML = '';
```
trueadm, line 639: I believe a faster way is to do:

`container.textContent = ''`

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -634,6 +634,17 @@ export function unhideTextInstance(
...
+  } else if (container.nodeType === DOCUMENT_NODE) {
+    const documentElement = ((container: any): Document).documentElement;
+    if (documentElement != null) {
+      documentElement.innerHTML = '';
```
trueadm, line 643: I wonder if we should ever do this? Shouldn't we trigger an invariant if mounting into the `document` instead?

packages/react-noop-renderer/src/createReactNoop.js
```diff
@@ -155,6 +155,10 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
     insertInContainerOrInstanceBefore(parentInstance, child, beforeChild);
   }
 
+  function clearContainer(container: Container): void {
+    container.children.splice(0);
```
trueadm, line 159: Nice. :)

This is mucher faster than `children.length = 0`.

packages/react-art/src/ReactARTHostConfig.js
trueadm: Can we also try `container.textContent = ''` as per my other comment? I think it works for SVG too.

packages/react-dom/src/client/ReactDOMHostConfig.js
bvaughn: React DOM's `Container` type is `Element | Document`. I don't know the historical reasons for supporting the latter, but that change seems much larger scope than this PR.

bvaughn: That being said, maybe this change is too aggressive anyway for the Document container case, since it would also clear `<head>` tags. I wish I knew more about why we supported that in the first place.

trueadm: We don't clear the `document` in the legacy case, so we should probably not do it here either.

packages/react-dom/src/client/ReactDOMHostConfig.js
bvaughn: Huh. That's interesting to note.

```js
document.documentElement.textContent = '';
// results in:
// <html></html>

document.documentElement.innerHTML = '';
// results in:
// <html>
//    <head></head>
//    <body></body>
// </html>
```

packages/react-dom/src/client/ReactDOMHostConfig.js
bvaughn: Where do you see that?

https://github.com/facebook/react/blob/4b02b6611141255a8e8b6d8a997e87c865f06626/packages/react-dom/src/client/ReactDOMLegacy.js#L113-L138

Looks to me like it removes all element nodes in the container, regardless of the container type?

packages/react-art/src/ReactARTHostConfig.js
bvaughn: I'm going to just put a "TODO" here. I just noticed that my Flow types are meaningless anyway b'c the file is missing a @flow pragma and there is no `Container` type for ART (which supports SVG, Canvas, and WebGL).

packages/react-dom/src/client/ReactDOMHostConfig.js
trueadm: That isn't the same as what you're doing. You're clearing the HTML, that above code is clearing the body contents of the document. When you do `document.lastChild` it gets the node from the body.

trueadm: Ignore me, I noticed you're clearing `documentElement`. That should be fine here.

bvaughn: Your edited comment makes sense. The previous one
> We don't clear the document in the legacy case, so we should probably not do it here either.

Made me think that you were suggesting I should *nothing* in the `DOCUMENT_NODE` case.

Anyway you're commenting on an outdated commit. I had already changed it to clear only the body.

packages/react-dom/src/client/ReactDOMHostConfig.js
```diff
@@ -636,11 +636,11 @@ export function unhideTextInstance(
...
   if (container.nodeType === ELEMENT_NODE) {
-    ((container: any): Element).innerHTML = '';
+    ((container: any): Element).textContent = '';
   } else if (container.nodeType === DOCUMENT_NODE) {
-    const documentElement = ((container: any): Document).documentElement;
-    if (documentElement != null) {
-      documentElement.innerHTML = '';
+    const body = ((container: any): Document).body;
+    if (body != null) {
+      body.textContent = '';
```
trueadm, line 643: Ah, you've updated it. Sorry, I missed this change.

packages/react-dom/src/client/ReactDOMHostConfig.js
trueadm: Yeah I noticed. The mobile UI for Github sucks though. Sorry for the churn.

packages/react-art/src/ReactARTHostConfig.js
trueadm: Makes sense :)

packages/react-dom/src/client/ReactDOMHostConfig.js
bvaughn: It does!

No worries.

PR #18732, opened 2020-04-25
Make ExpirationTime an opaque type

acdlite:
This is a series of commits that turns ExpirationTime into an opaque type.

They don't really deliver much value on their own, but they unblock later steps in an ongoing refactor to our internal concurrency and scheduling architecture.


packages/react-reconciler/src/ReactFiberExpirationTime.new.js
```diff
@@ -18,129 +18,75 @@ import {
...
+) {
+  return a >= b;
+}
+
+export function isSameExpirationTime(
+  a: ExpirationTimeOpaque,
+  b: ExpirationTimeOpaque,
+) {
+  return a === b;
+}
```
NE-SmallTown, line 80: Could I ask that what's the purpose of adding these two helper? Seems not good for code readable, I think maybe we can keep using `>`, `<`, `===`?

gaearon: The purpose of this PR is to prepare for the PR where the data structure changes. (From numbest to a bit mask.)

NE-SmallTown: Yea, I think you mean the Lanes, but just like bit masks in `SideEffectTags` or `TypeOfMode`, we don't use any helper, right? Back to the ponit, here I just means for the simple compare like `isSameOrHigher` or `isSame`, we can just use `>=` or `===`, but for complex compare or other purpose like `bumpPriorityHigher `, we can use it as a helper

NE-SmallTown:
"Each lane will have a priority associated with it", if so, could we use the term "taskQueue" of HTML eventloop spec? Seems that's what the quote means.

packages/react-reconciler/src/ReactFiberExpirationTime.new.js
NE-SmallTown: Oh, I know, thansk to the Closure

PR #18737, opened 2020-04-25
Devtools: Show inspectedElement `key` in right pane

### LINKED ISSUE #18702: Improve UX of finding full `key` value
## The current behavior

The full value of the `key` is very difficult / impossible to find and use in the interface of the React Devtools.

Kapture 2020-04-22 at 11 47 47

Only managed to find it by accident :(

## The expected behavior

The `key` is visible in the props list to the right.

### Detailed Proposal

As mentioned below in https://github.com/facebook/react/issues/18702#issuecomment-617924196

Add a light divider and new section in the props panel to the right.

Potentially also add a question mark that shows an explanation about the fact that things in this section are not really props.

Ref (Original implementation): https://github.com/facebook/react-devtools/pull/328

karlhorky:
## Summary

Initial MVP for showing `key` in `InspectedElementView`

Wasn't able to get the tooling for Flow and the development server set up quickly, so I have skipped that for now.

Just want to check if this approach has any fundamental problems before investing further time.

Closes #18702

## Test Plan

TBD

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
@@ -243,7 +243,6 @@ function InspectedElementContextController({children}: Props) {
             hooks: hydrateHelper(hooks),
             props: hydrateHelper(props),
             state: hydrateHelper(state),
-            key: hydrateHelper(key),
           };
 
           element = store.getElementByID(id);
--- a/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
@@ -209,6 +209,7 @@ function InspectedElementContextController({children}: Props) {
             hooks,
             props,
             state,
+            key,
           } = ((data.value: any): InspectedElementBackend);
 
           const inspectedElement: InspectedElementFrontend = {
@@ -241,6 +242,7 @@ function InspectedElementContextController({children}: Props) {
             hooks: hydrateHelper(hooks),
             props: hydrateHelper(props),
             state: hydrateHelper(state),
+            key: hydrateHelper(key),
           };
 
           element = store.getElementByID(id);
--- a/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
@@ -219,6 +219,7 @@ function InspectedElementContextController({children}: Props) {
             canViewSource,
             hasLegacyContext,
             id,
+            key,
             source,
             type,
             owners:
--- a/packages/react-devtools-shared/src/backend/legacy/renderer.js
+++ b/packages/react-devtools-shared/src/backend/legacy/renderer.js
@@ -731,7 +731,7 @@ export function attach(
       return null;
     }
 
-    const {displayName} = getData(internalInstance);
+    const {displayName, key} = getData(internalInstance);
     const type = getElementType(internalInstance);
 
     let context = null;
@@ -788,6 +788,8 @@ export function attach(
       displayName: displayName,
 
       type: type,
+
+      key: key != null ? key : null,
 
       // Inspectable properties.
       context,
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -2143,6 +2143,7 @@ export function attach(
       _debugOwner,
       _debugSource,
       stateNode,
+      key,
       memoizedProps,
       memoizedState,
       tag,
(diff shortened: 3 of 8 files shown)


karlhorky:
@bvaughn what do you think of this (admittedly naïve) approach? Am I missing something? I didn't see where `inspectedElement` was getting set in the 30 minutes that I put into this, so I'm not sure if the additional `key` in the type is going to cause problems.

bvaughn:
Tried running it locally but I'm getting an error when I try to inspect anything:
Screen Shot 2020-04-25 at 11 27 00 AM

packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
```diff
@@ -241,6 +242,7 @@ function InspectedElementContextController({children}: Props) {
             hooks: hydrateHelper(hooks),
             props: hydrateHelper(props),
             state: hydrateHelper(state),
+            key: hydrateHelper(key),
```
bvaughn, line 245: We don't need to use this for key.

bvaughn:
This should get you started:
```diff
diff --git a/packages/react-devtools-shared/src/backend/legacy/renderer.js b/packages/react-devtools-shared/src/backend/legacy/renderer.js
index 093fa8312..8cd3d5d2c 100644
--- a/packages/react-devtools-shared/src/backend/legacy/renderer.js
+++ b/packages/react-devtools-shared/src/backend/legacy/renderer.js
@@ -731,7 +731,7 @@ export function attach(
       return null;
     }
 
-    const {displayName} = getData(internalInstance);
+    const {displayName, key} = getData(internalInstance);
     const type = getElementType(internalInstance);
 
     let context = null;
@@ -787,7 +787,9 @@ export function attach(
 
       displayName: displayName,
 
-      type: type,
+      type,
+
+      key: key != null ? key : null,
 
       // Inspectable properties.
       context,
diff --git a/packages/react-devtools-shared/src/backend/renderer.js b/packages/react-devtools-shared/src/backend/renderer.js
index ae25089d7..48487206f 100644
--- a/packages/react-devtools-shared/src/backend/renderer.js
+++ b/packages/react-devtools-shared/src/backend/renderer.js
@@ -2142,10 +2142,11 @@ export function attach(
     const {
       _debugOwner,
       _debugSource,
-      stateNode,
+      key,
       memoizedProps,
       memoizedState,
       tag,
+      stateNode,
       type,
     } = fiber;
 
@@ -2300,6 +2301,8 @@ export function attach(
       // Does the component have legacy context attached to it.
       hasLegacyContext,
 
+      key: key != null ? key : null,
+
       displayName: getDisplayNameForFiber(fiber),
       type: elementType,
 
diff --git a/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js b/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
index 86a9a06d3..c88cc3543 100644
--- a/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/InspectedElementContext.js
@@ -219,6 +219,7 @@ function InspectedElementContextController({children}: Props) {
             canViewSource,
             hasLegacyContext,
             id,
+            key,
             source,
             type,
             owners:
@@ -242,7 +243,6 @@ function InspectedElementContextController({children}: Props) {
             hooks: hydrateHelper(hooks),
             props: hydrateHelper(props),
             state: hydrateHelper(state),
-            key: hydrateHelper(key),
           };
 
           element = store.getElementByID(id);
```

However the layout looks kind of broken:
Screen Shot 2020-04-25 at 11 35 24 AM

Key should be *above* props (I think). Also I don't think it should be formatted like a prop. It's not a nested attribute of an object like props/state are. It's more of a top level thing.

karlhorky:
Oh, thanks for pointing me in the right direction!

Sorry, it was still incomplete yesterday - still didn't get Flow / the local Devtools build to work on my system :( Was just waiting for the CI Flow to point me in the right direction of where to add the property in the backend, but was pulled away with other things.

But I guess with these code changes, all problems should be addressed...? If so I will move the PR out of draft.

karlhorky:
@bvaughn does this look ok like this? Or should I iterate on anything?

bvaughn:
You're going to have to give me more time than a Sunday afternoon to review the PR :smile: (I try not to work on Sundays.)

karlhorky:
Good rule 💯  Sorry, that was impatient - didn't take timezone + weekend into account.

bvaughn:
What do you think about this treatment?

Screen Shot 2020-05-11 at 10 53 28 AM

Screen Shot 2020-05-11 at 10 50 09 AM

karlhorky:
Sure, sounds fine. Thanks for your effort here!

bvaughn:
No problem. Thanks for collaborating on this!

PR #18739, opened 2020-04-26
Removes duplicate if statement

JesuHrz:
## Summary
This PR refectories a block of code that is duplicated multiple times in this file, so that it can be used in some other part of the code without adding this block again

### DIFF
--- a/packages/react/src/ReactElementValidator.js
+++ b/packages/react/src/ReactElementValidator.js
@@ -278,6 +278,25 @@ function validateFragmentProps(fragment) {
   }
 }
 
+function getTypeString(type) {
+  let typeString;
+  let info = '';
+  if (type === null) {
+    typeString = 'null';
+  } else if (Array.isArray(type)) {
+    typeString = 'array';
+  } else if (type !== undefined && type.$$typeof === REACT_ELEMENT_TYPE) {
+    typeString = `<${getComponentName(type.type) || 'Unknown'} />`;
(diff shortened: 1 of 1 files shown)


packages/react/src/ReactElementValidator.js
```diff
@@ -278,6 +278,25 @@ function validateFragmentProps(fragment) {
   }
 }
 
+function getTypeString(type) {
+  let typeString;
+  let info = '';
```
aap88-afk, line 283: *(single=account)

packages/react/src/ReactElementValidator.js
```diff
@@ -310,18 +329,8 @@ export function jsxWithValidation(
       info += getDeclarationErrorAddendum();
     }
 
-    let typeString;
-    if (type === null) {
```
aap88-afk, line 314: Unknown

packages/react/src/ReactElementValidator.js
```diff
@@ -278,6 +278,25 @@ function validateFragmentProps(fragment) {
...
 }
 
+function getTypeString(type) {
+  let typeString;
+  let info = '';
+  if (type === null) {
+    typeString = 'null';
+  } else if (Array.isArray(type)) {
+    typeString = 'array';
+  } else if (type !== undefined && type.$$typeof === REACT_ELEMENT_TYPE) {
```
aap88-afk, line 288: }origin'=true

aap88-afk:
First

PR #18745, opened 2020-04-26
[DevTools] Add Component Highlighting to Profiler

### LINKED ISSUE #17928: Profiler should highlight host components (e.g. DOM elements) on mouseover

bl00mber:
close #17928
This is the update of #17934 if @M-Izadmehr don't mind!

- useNativeElementHighlighter → useHighlightNativeElement
  - highlightNativeElement → highlightNativeElement
  - clearNativeElementHighlight → clearHighlightNativeElement

- Add missing dependencies

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.js
@@ -15,7 +15,6 @@ type Props = {|
   color: string,
   height: number,
   isDimmed?: boolean,
-  isHovered?: boolean,
   label: string,
   onClick: (event: SyntheticMouseEvent<*>) => mixed,
   onDoubleClick?: (event: SyntheticMouseEvent<*>) => mixed,
@@ -34,7 +33,6 @@ export default function ChartNode({
   color,
   height,
   isDimmed = false,
-  isHovered = false,
   label,
   onClick,
   onMouseEnter,
@@ -57,7 +55,7 @@ export default function ChartNode({
         onDoubleClick={onDoubleClick}
         className={styles.Rect}
         style={{
-          opacity: isHovered ? 0.75 : isDimmed ? 0.5 : 1,
+          opacity: isDimmed ? 0.5 : 1,
         }}
       />
       {width >= minWidthToDisplay && (
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.js
@@ -15,6 +15,7 @@ type Props = {|
   color: string,
   height: number,
   isDimmed?: boolean,
+  isHovered?: boolean,
   label: string,
   onClick: (event: SyntheticMouseEvent<*>) => mixed,
   onDoubleClick?: (event: SyntheticMouseEvent<*>) => mixed,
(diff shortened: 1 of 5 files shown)


packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -123,10 +131,22 @@ function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
...
+    setHoveredFiberData(null); // clear hovered fiber data for tooltip
+  };
+
   const itemData = useMemo<ItemData>(
     () => ({
       chartData,
-      hoverFiber,
+      isHovered: hoveredFiberData !== null,
+      onElementMouseEnter: handleElementMouseEnter,
+      onElementMouseLeave: handleElementMouseLeave,
```
bvaughn, line 150: There's no point to `useMemo` if we're passing in functions that get recreated inline every time we render.

Since these functions are just attached as mouse event handlers, normally it would not be important to memoize them- but in this case, not memoizing means that we could be doing a lot of otherwise unnecessary re-renders.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitRanked.js
```diff
@@ -94,28 +97,54 @@ type Props = {|
...
+    setHoveredFiberData(null); // clear hovered fiber data for tooltip
+  };
+
   const itemData = useMemo<ItemData>(
     () => ({
       chartData,
-      hoverFiber,
+      isHovered: hoveredFiberData !== null,
+      onElementMouseEnter: handleElementMouseEnter,
+      onElementMouseLeave: handleElementMouseLeave,
```
bvaughn, line 132: Same comment as above wrt memoization.

packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.js
```diff
@@ -55,7 +57,7 @@ export default function ChartNode({
         onDoubleClick={onDoubleClick}
         className={styles.Rect}
         style={{
-          opacity: isDimmed ? 0.5 : 1,
+          opacity: isHovered ? 0.75 : isDimmed ? 0.5 : 1,
```
bvaughn: This doesn't work the way it's intended, I think. The way it's written, `isHovered` does not mean "this fiber is hovered" it just means "something is hovered" - so currently, everything dims at the same time, making the effect not useful.

packages/react-devtools-shared/src/devtools/views/Profiler/CommitFlamegraph.js
```diff
@@ -123,10 +131,22 @@ function CommitFlamegraph({chartData, commitTree, height, width}: Props) {
...
+  const handleElementMouseLeave = () => {
+    clearHighlightNativeElement(); // clear highlighting of element on mouse leave
+    setHoveredFiberData(null); // clear hovered fiber data for tooltip
+  };
+
   const itemData = useMemo<ItemData>(
     () => ({
       chartData,
-      hoverFiber,
+      isHovered: hoveredFiberData !== null,
```
bvaughn: This isn't that useful.

We could fix it by passing in the hovered fiber ID, but that would require us to invalidate our memoization every time the mouse moved over something new which...seems like it's not worth the added benefit. I'm going to just suggest removing this.

PR #18746, opened 2020-04-26
DevTools: Keyboard arrow-key navigation for the flamegraph view

### LINKED ISSUE #16472: DevTools: Keyboard arrow-key navigation for the flamegraph view
It would be good to have keyboard navigation on the flamegraph view to make it easier and faster to scoot around to different components in the same render. 

Given that flamegraph is implemented in SVG and given the complexity of navigating through an uneven grid, I suspect keyboard access is probably beyond the scope of near-future DevTools releases, but wanted to file this issue while it's top-of-mind. 

Expected behavior: 
* There should be a selected/focused UI of some sort around the active component. This could be a black or colored outline around the currently selected component, or perhaps a background pattern like diagonal hatching.  IMHO, this would be valuable even without keyboard support! 
* Up arrow should change selection to the parent component
* Left/Right arrows should move selection to the prev/next sibling components. Should wrap when at the beginning or end of the line.
* Tab should move focus to the right pane (this matches behavior of other DevTools tabs like Elements)
* Shift+Tab should move focus to the rightmost control in the toolbar (currently this is the "next commit" button)

This is a companion issue to #332 which asks for keyboard access for the "Rendered At" list. Compared to this suggestion, #332 should be much easier to implement so I split them.

---
Originally reported by @justingrant via https://github.com/bvaughn/react-devtools-experimental/issues/333

ChrisDobby:
## Summary

Fixes #16472

Selected node now displays with a diagonal hatch.

Up arrow moves to parent, left/right move to previous/next sibling and wrap at the end/beginning of the row.

## Test Plan

flamegraph-keyboard

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.css
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.css
@@ -29,3 +29,7 @@
   text-align: left;
   transition: all ease-in-out 250ms;
 }
+
+.SelectedPath {
+  stroke-width: 2.4;
+}
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/ChartNode.js
@@ -15,6 +15,7 @@ type Props = {|
   color: string,
   height: number,
   isDimmed?: boolean,
+  isSelected?: boolean,
   label: string,
   onClick: (event: SyntheticMouseEvent<*>) => mixed,
   onDoubleClick?: (event: SyntheticMouseEvent<*>) => mixed,
@@ -33,6 +34,7 @@ export default function ChartNode({
   color,
   height,
   isDimmed = false,
+  isSelected = false,
   label,
   onClick,
   onMouseEnter,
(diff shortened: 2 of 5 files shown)


ChrisDobby:
bump

bvaughn:
I rebased this and updated the PR to point to main.

I'm not sure the change feels right though. Left/right arrow navigation works okay. Up arrow navigation works to a point but then seems to deselect focus. Down arrow does nothing. (I think I would have expected it to select the first child, similar to how it works in the Components tab.)

bvaughn:
I greatly appreciate this contribution! 🙇🏼 

I think after testing the feature though, it doesn't feel like an intuitive way to navigate the profiling data, so I think I'm going to leave the UI as-is.

Thanks for understanding.

justingrant:
@bvaughn - If the two concerns you identified (1. don't remove focus when up-arrowing past the top; 2. down arrow should select the first child) are fixed, would you support this PR? Those seem like straightforward changes that could probably be made relatively easily. If you think the PR could be good with those changes made, would it be useful to leave the PR open and ask for changes instead of closing it, or best to open a new PR?

BTW, you can blame the lack of down-arrow support on me because I didn't mention down arrow at all in my original [suggestion](https://github.com/bvaughn/react-devtools-experimental/issues/333) that led to this PR. BTW, thanks so much @ChrisDobby for building this!

bvaughn:
@justingrant In general, I didn't find the feature very compelling when I played around with it. I'm not sure I see people wanting to navigate the flamegraph using arrow keys as a common use case.

That being said, if those issues were addressed...I'd at least reconsider.

ChrisDobby:
I'll take a look at those issues and see how it goes! I tried it again earlier and the first thing I did was use the down arrow expecting it to do something :-)

PR #18756, opened 2020-04-27
Add experimental ReactDOM.createEventHandle

trueadm:
This PR adds the experimental `ReactDOM.createEventHandle` API. This works similar to how the recently removed `ReactDOM.useEvent` works, and provides an event handle object that can be use to set and clear listeners.

```jsx
const clickHandle = ReactDOM.createEventHandle('click')

function Component() {
  const ref = useRef(null);

  useEffect(() => {
    clickHandle.setListener(ref.current, () => {
      console.log('click!');
    });
  });

  return <button ref={ref}>Click me</button>
}
```

The key difference is that we no longer need to do this as a hook. I've added an implementation of how you can do `useEvent` with this API, and that is actually used in the `useFocus` and `useFocusWithin` surfaces.

I know this is a large PR, but most of the code is being brought back, and is largely unchanged from how `useEvent` worked before (all the tests are basically the same, as is the implementation for handling events). Some codepaths have been unified, to avoid code duplication and improve the fast paths compared to before.

Why is this API better than the hook version? Well we found there to be lots of cases where you want to use `addEventListener` in an application, either in a React effect or in some other external system that connects to React. Given this API is meant to replace usages of `addEventListener`, as a form of "wrapper", we wanted an API that can be used as such. Having it limited to a hook means that it can only be used during the render phase of a component, which doesn't cover many cases we want to be able to provide cover for.

### DIFF
--- a/packages/react-dom/index.classic.fb.js
+++ b/packages/react-dom/index.classic.fb.js
@@ -36,4 +36,5 @@ export {
   unstable_scheduleHydration,
   unstable_renderSubtreeIntoContainer,
   unstable_createPortal,
+  unstable_createEventHandle,
 } from './src/client/ReactDOM';
--- a/packages/react-dom/index.js
+++ b/packages/react-dom/index.js
@@ -27,4 +27,5 @@ export {
   unstable_scheduleHydration,
   unstable_renderSubtreeIntoContainer,
   unstable_createPortal,
+  unstable_createEventHandle,
 } from './src/client/ReactDOM';
--- a/packages/react-dom/index.modern.fb.js
+++ b/packages/react-dom/index.modern.fb.js
@@ -19,4 +19,5 @@ export {
   createBlockingRoot as unstable_createBlockingRoot,
   unstable_flushControlled,
   unstable_scheduleHydration,
+  unstable_createEventHandle,
 } from './src/client/ReactDOM';
--- a/packages/react-dom/src/client/ReactDOM.js
+++ b/packages/react-dom/src/client/ReactDOM.js
@@ -20,6 +20,7 @@ import {
   unmountComponentAtNode,
 } from './ReactDOMLegacy';
 import {createRoot, createBlockingRoot, isValidContainer} from './ReactDOMRoot';
+import {createEventHandle} from './ReactDOMEventHandle';
 
 import {
   batchedEventUpdates,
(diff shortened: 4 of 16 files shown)


TrySound:
The downside of the new api is the necessity to add pure annotation before createEventHandle call to make expression treeshakable. Hook version didn't have this problem.

trueadm:
@TrySound Like I mentioned in the summary, only providing a hook version has a load of downsides. You can still roll your own `useEvent` and have it tree-shakable, see the example in this PR.

packages/react-dom/src/events/DOMEventProperties.js
```diff
@@ -236,7 +245,7 @@ export function getEventPriorityForListenerSystem(
   }
   if (__DEV__) {
     console.warn(
-      'The event "type" provided to useEvent() does not have a known priority type.' +
+      'The event "type" provided to createEventHandler() does not have a known priority type.' +
```
nihgwu, line 248: createEventHandle?

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -1029,7 +1029,9 @@ function commitUnmount(
       if (enableDeprecatedFlareAPI) {
         unmountDeprecatedResponderListeners(current);
       }
-      beforeRemoveInstance(current.stateNode);
+      if (current.ref !== null) {
```
gaearon, line 1032: Can we put this in the feature flag? The benefit is that custom renderer users won't have to implement a noop method.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -1072,7 +1074,9 @@ function commitUnmount(
           unmountDeprecatedResponderListeners(current);
         }
         const scopeInstance = current.stateNode;
-        prepareScopeUnmount(scopeInstance);
+        if (current.ref !== null) {
```
gaearon, line 1077: Same.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
trueadm: I can wrap it with a feature flag, but I still need to export it from each of the renderers or Flow throws an error.

Issue #14490, opened 2018-12-24
Lazy useRef instance variables


Diokuz:
I want to save a class instance, not just plain object, with `useRef` hook. And, I dont want to construct it each time hook function is invoked.

Is it possible to do that?

My thoughts were:

```js
const instanceRef = useRef(() => new Instance())
```
But `useRef` does not accept the _initial value_ as a function (`useState` and `useReducer` could do that).
Will that change in _alpha 3_?

nikilok:
@Diokuz useRef does take in an initial state.

https://reactjs.org/docs/hooks-reference.html#useref

In the above link, they just pass `null` as the initial value.

Diokuz:
@nikilok yes, but the question was about how to create lazy initial value (create it only once).

My current solution is:
```js
const [refInitialValue] = useState(() => new Instance)
const instanceRef = useRef(refInitialValue)
```

but that is a bit overhead.

TrySound:
@Diokuz Just check its existence
```js
const instance = React.useRef(null)
if (instance.current == null) {
  instance.current = {
    // whatever you need
  }
}
```

jamiewinder:
I created this custom hook to help with this limitation, someone might find it useful:
```typescript
import { useRef } from 'react';

const SENTINEL = {};

export function useRefFn<T>(init: () => T) {
    const ref = useRef<T | typeof SENTINEL>(SENTINEL);
    if (ref.current === SENTINEL) {
        ref.current = init();
    }
    return ref as React.MutableRefObject<T>;
}
```

Kovensky:
It happens to work in this case because your function is generic and so T is "like {}", but you need to use a Symbol for sentinels to create a `unique symbol` type. Using an empty object's type as an input makes it nearly the same as `any` but without `null | undefined`.

jamiewinder:
Sorry, I'm not following. It would work for null, undefined, or any value as far as I can see. 

I use an empty object as the sentinel as it makes it unique and distinct for the ===  check which sets the ref to the real initial value on the first invocation. It'd work it the sentinel were a symbol too, but I don't think it makes a difference one way or the other?

jamiewinder:
Late realisation: doesn't useMemo (with an empty array as the second parameter) do exactly this?

gaearon:
[...]

PR #18760, opened 2020-04-28
Lint rule: prefer-lazy-initialization

vikr01:
## Summary

Per discussion in https://github.com/facebook/react/issues/14490#issuecomment-454973512 and [the "How to create expensive objects lazily?" section in the Hooks FAQ](https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily), I've added a lint rule for prefer-lazy-initialization.

I've seen *tons* of cases of people using `useRef(new Map())`/`useState(new Set())`/etc.

This rule tries to behave a bit similar to the `exhaustive-deps` rule in that it allows `additionalHooks` as an option, and treats the first option as the initial value in those hooks.

This rule checks the direct parameters passed in (i.e. it wouldn't catch something like `useRef((new Foo).bar)`).

Not adding it to the recommended config or marking it as "recommended" to prevent breaking changes from those already linting their codebases with this plugin/recommended config. People will have to explicitly turn this rule on to use it.

Additional options include `classObjectInstantiation` and `functionCall` to explicitly turn off checking for creating a new class object or checking for a function call.

It will also ignore nested hooks, since those can't be lazily initialized.

Example cases:
```javascript
useRef(new Foo()); // fails
useState(new Foo()); // fails
useCustomHook(new Foo()); // bad, with options {additionalHooks: 'useCustomHook'}

useRef(new Foo()); // passes, with options {classObjectInstantiation: false}

useRef(useState(()=>new Foo()); // passes because callbacks/nested hooks are used
useRef(useState(new Foo())); // fails on `useState` but not `useRef`
```

## Test Plan

`yarn test ESLintRulePreferLazyInitialization`

Maybe this could be expanded in the future to auto-suggest turning the initial value in `useState` into a callback?


bgirard:
Here's a modified benchmark for the first render case:
https://jsperf.com/creating-a-function-vs-creating-a-map/4
compared to the original benchmark that covers re-renders:
https://jsperf.com/creating-a-function-vs-creating-a-map

This suggestion isn't clear cut. In particular it means slowing down first render by introducing an extra function, a tiny hit to JS size, and the devX impact of a lint. Generally first render performance is more important than the re-render.

Where the win is clear is when we're not avoiding just a 'new Map()' but a 'new VeryCostlyObject()' in which case this is a clear win.

Overall I don't have a strong opinion for or against.

Issue #18762, opened 2020-04-28
Bug: generated files in source tree for react-window node module


pravi:
react-devetools-shared module has a coy of react-window node module ->
https://github.com/facebook/react/tree/master/packages/react-devtools-shared/src/node_modules/react-window

Can this be added to devDependencies in package.json instead? Or if that is not possible, can it be replaced with its source from github repo and built as part of the build process?

React version:

16.13.1

## Steps To Reproduce

It an issue with vendored code in the source.

Link to code example:

## The current behavior

build dependency is embedded in the source tree

## The expected behavior

all build dependencies should be added to package.json and pulled via yarn install.

pravi:
Build is passing if dist directory (packages/react-devtools-shared/src/node_modules/react-window/dist) is dropped.

bvaughn:
It is an intentional fork with a couple of small concurrent mode specific tweaks added. Eventually I will port those changes to the main react-window (probably). For now the fork is intentional though.

bvaughn:
Now I'm second guessing myself if the forked changes were already upstreamed. I can't remember...

I'll reopen this for now, but I don't think this is a priority issue for me to look into 😅 so it's probably going to sit for a while.

PR #18763, opened 2020-04-28
Drop generated files (dist) from source tree (src is enough to build)

pravi:
## Summary

Fixes: #18762 

## Test Plan

Build was passing with dist removed.


pravi:
Close in favor of #18765

PR #18765, opened 2020-04-28
Add react-window as devDependency and drop embedded copy

### LINKED ISSUE #18762: Bug: generated files in source tree for react-window node module

pravi:
## Summary

Fixes #18762


yarn.lock
```diff
@@ -8726,6 +8726,11 @@ memfs-or-file-map-to-github-branch@^1.1.0:
   resolved "https://registry.yarnpkg.com/memfs-or-file-map-to-github-branch/-/memfs-or-file-map-to-github-branch-1.1.2.tgz#9d46c02481b7eca8e5ee8a94f170b7e0138cad67"
   integrity sha512-D2JKK2DTuVYQqquBWco3K6UfSVyVwmd58dgNqh+TgxHOZdTmR8I130gjMbVCkemDl/EzqDA62417cJxKL3/FFA==
 
+"memoize-one@>=3.1.1 <6":
+  version "5.1.1"
+  resolved "https://registry.yarnpkg.com/memoize-one/-/memoize-one-5.1.1.tgz#047b6e3199b508eaec03504de71229b8eb1d75c0"
+  integrity sha512-HKeeBpWvqiVJD57ZUAsJNm71eHTykffzcLZVYWiVfQeI1rJtuEaS7hQiEpWfVVk18donPwJEcFKIkCmPJNOhHA==
+
```
bl00mber, line 8733: This is the duplicate of a dep which is already presented below?

package.json
```diff
@@ -77,6 +77,7 @@
     "prop-types": "^15.6.2",
     "random-seed": "^0.3.0",
     "react-lifecycles-compat": "^3.0.4",
+    "react-window": "^1.8.5",
```
bl00mber, line 80: This dep needs to be placed in `react-devtools-shared`

bvaughn:
This fork is intentional as mentioned [here](https://github.com/facebook/react/issues/18762#issuecomment-620905600) :smile:

Thanks though!

PR #18766, opened 2020-04-28
Refactor ElementListenerMap to use expando properties

trueadm:
Note: Last year, I originally made the change that moved this code to use WeakMaps in https://github.com/facebook/react/pull/15036. 

I recently did some profiling and noticed that we are spending quite a bit of time accessing and setting Maps on `DOMEventListenerMap`'s WeakMap for each element when using the Modern Event System. The times weren't huge, they were just collectively many more `ms` than I would have liked. I profiled switching back to a `Map` and the times do go down, but then we end up likely leaking by making that change. Why didn't we notice this before? Well I believe it has to do with the fact that we almost always were doing lookups on the `document` as the WeakMap key, with very few other DOM nodes used as keys. Doing expando lookups on `document` is about equivalent in performance to doing lookups with a WeakMap (`document` has only a slow path?). Now that we've moved to roots, I'm now seeing a more noticeable perf difference.

With all that in mind, I've moved us back to using internal expando properties on DOM nodes instead, which fixes the performance issue I was seeing. I also don't see any regressions with the legacy event system doing this.


packages/react-dom/src/client/ReactDOMComponentTree.js
```diff
@@ -30,8 +31,19 @@ const randomKey = Math.random()
   .toString(36)
   .slice(2);
 const internalInstanceKey = '__reactFiber$' + randomKey;
-const internalEventHandlersKey = '__reactEvents$' + randomKey;
+const internalPropsKey = '__reactProps$' + randomKey;
```
bvaughn, line 34: The last time we renamed these fields, [we posted about it](https://fb.workplace.com/groups/react.core.team/permalink/3470160429699214/) to let our few semi-supported use cases know ahead of time that it was coming.

I don't *think* anything relied on `__reactEvents` (or previously `__reactEventHandlers`) so I think this change is probably fine, but cc @gaearon to confirm.

packages/react-dom/src/events/plugins/ModernSelectEventPlugin.js
```diff
@@ -144,6 +146,29 @@ function constructSelectEvent(nativeEvent, nativeEventTarget) {
...
+      return false;
+    }
+  }
+  return true;
+}
+
+function isListeningToEvent(
+  registrationName: string,
+  mountAt: Document | Element,
+  listenerMap?: ElementListenerMap,
```
bvaughn, line 166: This optional param seems awkward. I think it would be better just to inline the `listenerMap.has(registrationName)` check in `isListeningToEvents` and leave off this new param for the other callers.

packages/react-dom/src/client/ReactDOMComponentTree.js
trueadm: I looked internally and it looks good to go with this PR.

gaearon: Yeah this particular one is safe

gaearon: This wasn't a rename btw. The old name is still there below.

bvaughn: Right but it's been repurposed, so...kind of the same as a rename.

PR #18770, opened 2020-04-28
Refine ReactDOM "Container" type

trueadm:
The `Container` is never a `document` (anymore at least), so let's refine the type better.


bvaughn:
I don't know if this change is safe.

We have a lot of `DOCUMENT_NODE` and `DOCUMENT_FRAGMENT_NODE` checks in the DOM config files.

trueadm:
Yeah my bad, I forgot about the legacy paths.

PR #18771, opened 2020-04-28
useMutableSource hydration support

### LINKED ISSUE #18183: useMutableSource and hydration

bvaughn:
Add hydration support for `useMutableSource` (resolves #18183)

## Usage example
```js
// 1. Render app on the server...

// 2. Create root for hydration. and pre-register all mutable sources.
// This eagerly records source versions to guard against tearing during (async) hydration.
const root = ReactDOM.unstable_createRoot(container, {
  hydrate: true,
  mutableSources: [relayMutableSource],
});

// 3. Render and hydrate like normal.
root.render(<Application />);
```

## MVP checklist
- [x] Determine if the change in behavior to unsafe `useOpaqueIdentifier` reads inside of a `Suspense` boundary is valid (cc @lunaruan).
- [x] Get consensus on the new API.

## Follow up work
- [ ] #18835: Decide when to clear `mutableSourceEagerHydrationData`. We don't currently have a way to detect when all outstanding hydration work is finished. (It may never finish.) This means that in use cases where the version is the entire sore (e.g. an immutable store without a version "number") we would retain a reference to that store, preventing it from being GCed.


packages/react-dom/src/client/ReactDOMRoot.js
```diff
@@ -115,6 +116,13 @@ ReactDOMRoot.prototype.unmount = ReactDOMBlockingRoot.prototype.unmount = functi
   });
 };
 
+ReactDOMRoot.prototype.registerMutableSourceForHydration = ReactDOMBlockingRoot.prototype.registerMutableSourceForHydration = function(
+  mutableSource: MutableSource<any>,
```
acdlite, line 120: Have you considered making his an option to `createRoot`? Like `hydrate` is.

packages/react-reconciler/src/ReactFiberBeginWork.old.js
```diff
@@ -2239,6 +2254,14 @@ function updateDehydratedSuspenseComponent(
   // but after we've already committed once.
   warnIfHydrating();
 
+  if ((getExecutionContext() & RetryAfterError) !== NoContext) {
```
acdlite, line 2257: Would checking if `root.hydrate` is false work?

packages/react-dom/src/client/ReactDOMRoot.js
```diff
@@ -30,6 +29,8 @@ export type RootOptions = {
   ...
 };
 
+import {findHostInstanceWithNoPortals} from 'react-reconciler/src/ReactFiberReconciler';
+import {registerMutableSourceForHydration} from 'react-reconciler/src/ReactMutableSource';
```
acdlite, line 33: Could you export this via the reconciler module? Then you don't to configure an additional forked entry point

packages/react-dom/src/client/ReactDOMRoot.js
bvaughn: I hadn't considered it, no. Seems reasonable!

packages/react-reconciler/src/ReactFiberBeginWork.old.js
bvaughn: I'm...not sure. I think?

I actually haven't got a test for this code path yet. I need to add one. I'll revisit this change once I have one.

packages/react-dom/src/client/ReactDOMRoot.js
bvaughn: Sure.

packages/react-dom/src/client/ReactDOMRoot.js
acdlite: Though I wonder if it's ever the case that you need to create the root before you know what all the mutable sources are. Like if there's a source that's only used by a chunk of the app and that chunk doesn't get streamed in until later. Seems plausible because otherwise you have to know about all possible sources up front, which is bad for code splitting.

bvaughn: That's possible. Feels kind of odd to pass around the root on the other hand.

I don't have a strong preference between these two so I'll just leave it as is for the time being. Seb can share his thoughts.

sebmarkbage: The `root.render(...)` API can be used for updates. The root instance is generally an "alive" thing once created.

It wouldn't make sense to call `registerMutableSourceForHydration` after hydration has already started since bets are off then. It would have to start erroring based on temporal values which seems a bit like a foot gun when you refactor or change the timing of two async calls like bootloading the code that runs render and the core that registerMutableSourceForHydration in parallel causing a race condition.

For event replaying I was thinking that createRoot would be passed a list of events that can be replayed after they've been recorded before React loaded. These are "grandfathered" into the event replaying queue. Once createRoot is called, we've already attached the event listeners on the root so it's a good time to hand off.

Additionally I was thinking createRoot would be passed a "time" option to indicate what the time was at the server render to get consistent hydration of time based values. It also already takes callbacks for hydration purposes too.

I.e. there's already a grab bag of option that go into createRoot and that's really the point of handoff from prehydration work to hydration. So I think this should also be an option to createRoot.

packages/react-reconciler/src/ReactFiberBeginWork.old.js
sebmarkbage: No that wouldn't work because root.hydrate is no longer true after the first outer level is committed. At that point "dehydrated boundaries" are marked as still hydrating. `root.hydrate` is kind of like that the root is a dehydrated boundary.

bvaughn:
No objection. I had made and stashed that change earlier because I didn't feel strongly about it. Just pushed it.

packages/react-dom/src/client/ReactDOMRoot.js
bvaughn: > It wouldn't make sense to call `registerMutableSourceForHydration` after hydration has already started since bets are off then.

This is a strong reason. 👍 I'm convinced.

bvaughn:
(Back to you with the options change, @sebmarkbage )

packages/react-dom/src/client/ReactDOMRoot.js
```diff
@@ -27,6 +26,7 @@ export type RootOptions = {
     onDeleted?: (suspenseNode: Comment) => void,
     ...
```
sebmarkbage, line 27: I feel like the option should probably go into this object that has the hydration related stuff. It doesn't really do anything if you're not hydrating and you shouldn't feel obligated to put all your mutableSources here if they're not used for hydration.

packages/react-reconciler/src/ReactFiberRoot.new.js
```diff
@@ -46,12 +46,13 @@ function FiberRootNode(containerInfo, tag, hydrate) {
   this.pingedLanes = NoLanes;
   this.expiredLanes = NoLanes;
   this.mutableReadLanes = NoLanes;
-
   this.finishedLanes = NoLanes;
 
   this.entangledLanes = NoLanes;
   this.entanglements = createLaneMap(NoLanes);
 
+  this.mutableSourceEagerHydrationData = [];
```
sebmarkbage, line 54: This array is a bit unnecessary if we're not going to hydrate. In the case that you use many roots, it's probably because they're not hydrating. Maybe we can make this nullable?

Additionally, if this is a non-hydration compatible renderer, we probably shouldn't add this field at all.

Can you gate all of this code behind the `supportsHydration` HostConfig? That way we can produce an optional DOM build that strips out hydration easily.

packages/react-dom/src/client/ReactDOMRoot.js
bvaughn: Sure.

packages/react-reconciler/src/ReactFiberRoot.new.js
bvaughn: Sure.

bvaughn:
I think ec01142 is what you had in mind, but let me know if I misunderstood.

PR #18772, opened 2020-04-28
[DevTools] enable Electron interactions on Linux & auto copy script tags

bl00mber:
When `react-devtools` is runned on Linux, the user cannot interact with Electron window. Also it may be convenient to have button to copy script tag immediately instead of manually copying it.

Peek 2020-04-29 1

### DIFF
--- a/packages/react-devtools/app.html
+++ b/packages/react-devtools/app.html
@@ -30,6 +30,10 @@
         margin: 0;
       }
 
+      .input-container {
+        position: relative;
+      }
+
       .input {
         display: block;
         font-weight: 100;
@@ -37,6 +41,18 @@
         border: 1px solid #aaa;
         background-color: #fff;
         color: #666;
+        position: relative;
+        left: -8px;
+      }
+
+      .copy-btn {
+        position: absolute;
+        cursor: pointer;
+        top: 0px;
+        right: -10px;
+      }
+      .copy-btn:active {
+        opacity: 0.5;
       }
 
       .link {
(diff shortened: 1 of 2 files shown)


packages/react-devtools/app.js
```diff
@@ -25,7 +25,7 @@ app.on('ready', function() {
     width: 800,
     height: 600,
     icon: join(__dirname, 'icons/icon128.png'),
-    frame: false,
+    frame: process.platform === 'linux' ? true : false,
```
bvaughn, line 28: Can you explain why this change is necessary?

packages/react-devtools/app.html
```diff
@@ -116,8 +132,14 @@
           <div class="box-header">React DOM</div>
           <div class="box-content">
             Add one of the following:
```
bvaughn, line 134: Could we just change this to:

> Add one of the following (click to copy):

and then automatically copy the text for them? (without adding new buttons)

bl00mber:
> how would it be usable even once the DevTools loaded?

Seems to be related to `-webkit-app-region: drag;` although it was possible to fix by disabling frameless window. Reenabled frameless window and disabled drag on boxes instead.

packages/react-devtools/app.html
```diff
@@ -150,13 +162,13 @@
 
       const $localhost = $("#localhost");
       $localhost.innerText = `<script src="http://localhost:${port}"></` + 'script>';
-      $localhost.addEventListener('click', selectAll);
-      $localhost.addEventListener('focus', selectAll);
+      $localhost.addEventListener('click', selectAllAndCopy);
+      $localhost.addEventListener('focus', selectAllAndCopy);
```
bvaughn, line 189: Huh, I just noticed that this has the unexpected effect of re-focusing (and re-copying) every time you tab away from and back to the process/window. Probably fine though. Overall this seems like a nice improvement.

PR #18773, opened 2020-04-29
[Blocks] Add Client Fixture

gaearon:
This is not very interesting by itself.

Just adds a client-only fixture that we can play with to start extracting the pieces that need to be librarified, like `react/cache` or `react-data/fetch` (all names temporary).


fixtures/blocks/src/lib/cache.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {createContext} from 'react';
+
+// TODO: clean up and move to react/cache.
```
gaearon, line 10: Since these pieces don't really need internals and only exist for coordination, I figured it's less hassle to put them right into the fixture for now. And then when we have the API nailed down a bit more, upstream them in packages and maybe delete/replace the existing `react-cache` code with them.

sebmarkbage: It's fine but I've noticed that I tend to prefer to work the opposite way of this.

E.g. if you notice the Flight stuff I started by adding all the infra for custom builds and everything in its right place before I even wrote any code. That way you have all that out of the way upfront and then you can just get a good flow going adding feature after feature.

Otherwise the first time you want to write a unit test or what not, you won't be able to do it and then it becomes a blocker.

That's why I tend to want to front load all the easy stuff so you don't end up getting blocked and discouraged by that later when you're heads down.

gaearon: Is this "API" reasonable to put into entry points? Then I'll just do that.

PR #18774, opened 2020-04-29
[Blocks] Initial implementation of cache and data/fetch

gaearon:
This adds a few entry points (names might change):

- `react/unstable-cache` (isomorphic context for coordination)
- `react-data` (`createResource` will live here, not used yet)
- `react-data/fetch` (a rudimentary data provider for `fetch`)

The APIs will likely change a lot, this is just initial wiring.
I haven't added proper tests yet because for now I'm developing this primarily with the fixture (https://github.com/facebook/react/pull/18773).

I'm not touching `react-cache` for now except renaming the source code entry point to `ReactCacheOld.js` for disambiguation. When we reach parity I can delete it, but it's used by some tests now.

Test plan: https://github.com/facebook/react/pull/18773 works.


lautapercuspain:
Hi, my question is why React wants to implement data-fetch/cache functionality inside the library?
I know there're few other libraries that already handle this in an efficient way.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @flow
+ */
+
+import type {Wakeable} from 'shared/ReactTypes';
+
+import {readCache} from 'react/unstable-cache';
+
+const Pending = 0;
+const Resolved = 1;
```
chicoxyzzy, line 15: Probably `Fulfilled` would be a more correct word here

https://github.com/domenic/promises-unwrapping/blob/master/docs/states-and-fates.md#states
https://jakearchibald.com/2014/resolve-not-opposite-of-reject/

gaearon:
> Hi, my question is why React wants to implement data-fetch/cache functionality inside the library?

For now this is very rudimentary and just to get the example working. We’ll need to think about scope more. But we’ll probably want a few built-in simple bindings for the most common cases. At least as reference implementations.

packages/react-data/src/fetch/ReactDataFetch.js
chicoxyzzy: Also I found other usages of `Resolved` variable name in ReactCache, and ReactLazy. It seems that renaming won't be a breaking change. Let me know if I can help with this.

gaearon: Makes sense. We can rename this later, doesn't really matter how it's called internally.

sebmarkbage: The term Fulfilled makes sense in combination with the term Promise since you "fulfill a promise to someone".

But these are not Promises. So not sure we need to stick to the same terminology. Maybe we should go with something completely different to make that clear.

lautapercuspain: @seangransee So, fallowing the same reasoning it can be called,`Completed` or `Finished`.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -0,0 +1,108 @@
...
+    xhr.send();
+    throw wakeable;
+  }
+  const result: Result = entry;
+  switch (result.status) {
+    case Resolved:
+      return result.value;
+    case Pending:
+    case Rejected:
+      throw result.value;
```
sebmarkbage, line 103: You can just treat this case as default. Makes the code smaller and throws something which is better than returning an invalid type.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -0,0 +1,108 @@
...
+        responseJSON = JSON.parse(xhr.response);
+      } catch (err) {
+        const rejectedResult = ((result: any): RejectedResult);
+        rejectedResult.status = Rejected;
+        rejectedResult.value = err;
+        return;
+      }
+      const resolvedResult = ((result: any): ResolvedResult);
+      resolvedResult.status = Resolved;
+      resolvedResult.value = responseJSON;
```
sebmarkbage, line 80: The return value should a a Response object like the Fetch API.

sebmarkbage: The tricky part to polyfilling this with XHR is that we don't know if you're going to read it as binary data or string. I.e. if we should set it to `responseType = 'arraybuffer'`.

gaearon: So you check status and JSON.parse in userland? Kind of annoying for the common case. It's convenient when a library does this for you.

sebmarkbage:
We should have two builds for fetch. One for node and one for browser so that they can have separate builds with separate polyfills. Like we do with flight and dom server etc.

https://github.com/facebook/react/blob/master/packages/react-flight-dom-webpack/server.node.js

gaearon:
I'll get these in and then rebuild on top.

packages/react-data/src/fetch/ReactDataFetch.js
tigt: Or `Done`

PR #18778, opened 2020-04-29
Refine experimental Scopes API

trueadm:
This PR cleans up some of the types around the experimental Scopes API and simplies the data structure around how they're created. The main goal here is to cleanup the API for integration with `createEventHandle`.

With this PR, rather than create a `ScopeMethods` object, we create only a `ScopeInstance` object, which has the methods on – but attached from static functions rather than generated inline. In order for the methods to work with this change, it uses itself (the `ScopeInstance`) as a key to ask the renderer for it's associated fiber so it can carry out traversal.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponentTree.js
+++ b/packages/react-dom/src/client/ReactDOMComponentTree.js
@@ -8,6 +8,7 @@
  */
 
 import type {Fiber} from 'react-reconciler/src/ReactInternalTypes';
+import type {ReactScopeInstance} from 'shared/ReactTypes';
 import type {
   Container,
   TextInstance,
@@ -47,7 +48,7 @@ export type ElementListenerMapEntry = {
 
 export function precacheFiberNode(
   hostInst: Fiber,
-  node: Instance | TextInstance | SuspenseInstance,
+  node: Instance | TextInstance | SuspenseInstance | ReactScopeInstance,
 ): void {
   (node: any)[internalInstanceKey] = hostInst;
 }
@@ -204,4 +205,10 @@ export function getEventListenerMap(node: EventTarget): ElementListenerMap {
     elementListenerMap = (node: any)[internalEventHandlersKey] = new Map();
   }
   return elementListenerMap;
+}
+
+export function getFiberFromScopeInstance(
+  scope: ReactScopeInstance,
+): null | Fiber {
+  return (scope: any)[internalInstanceKey] || null;
 }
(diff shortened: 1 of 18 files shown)


packages/react-dom/src/client/ReactDOMComponentTree.js
```diff
@@ -205,3 +206,9 @@ export function getEventListenerMap(node: EventTarget): ElementListenerMap {
   }
   return elementListenerMap;
 }
+
+export function getFiberFromScopeInstance(
+  scope: ReactScopeInstance,
+): null | Fiber {
+  return (scope: any)[internalInstanceKey] || null;
```
bvaughn, line 213: Wrap with feature flag?

bvaughn:
What are the longer term plans for the Scope API and non-DOM renderers? Are there any?

trueadm:
> What are the longer term plans for the Scope API and non-DOM renderers? Are there any?

There are no plans to use the Scope API in anything other than the DOM right now (they also work with the test renderer). We want to refine the API once we remove Flare, so that they move to become `useScope`, so around there we'll likely expose a larger design document as to how it might help in a larger picture with other renderers (if they are interested in the features).

bvaughn:
> There are no plans to use the Scope API in anything other than the DOM right now

Interesting. I just saw all of the "noop" methods being added to host configs.

Wonder if we should create a `ReactFiberHostConfigWithNoScope` file that just exports shims for all of the Scopes APIs? That would be more inline with other things like hydration/mutation/persistence.

trueadm:
@bvaughn Ah that's a good point. I can totally do that tomorrow :)

packages/react-reconciler/src/ReactFiberCommitWork.old.js
```diff
@@ -1725,6 +1728,7 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
             updateDeprecatedEventListeners(nextListeners, finishedWork, null);
           }
         }
+        prepareScopeUpdate(scopeInstance, finishedWork);
```
bvaughn, line 1731: Why is this necessary to have here? Shouldn't the call in complete work be sufficient?

As an aside, I *think* this would be the only place where we have an indirect call to `precacheFiberNode` in the commit phase.

trueadm: This is necessary so we update the the fiber on scope. We do one in the complete phase for initial mount, and in the commit phase after each update (so we point to the current fiber tree). This is similar to the `updateFiberProps` logic we have in the other commit methods, except scopes don't have props.

PR #18781, opened 2020-04-29
Handle legacy root contains in clearContainer

trueadm:
When debugging an internal issue, I noticed that we have an issue where we clear down a root that has existing content. This particular fix checks in `clearContainer` if the container we're trying to clear has a legacy container and if it does, checks to see if it matches the root we expected.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.old.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.old.js
@@ -679,7 +679,7 @@ function completeWork(
           // If we hydrated, then we'll need to schedule an update for
           // the commit side-effects on the root.
           markUpdate(workInProgress);
-        } else if (!fiberRoot.hydrate) {
+        } else if (!fiberRoot.hydrate && workInProgress.child !== null) {
           // Schedule an effect to clear this container at the start of the next commit.
           // This handles the case of React rendering into a container with previous children.
           // It's also safe to do for updates too, because current.child would only be null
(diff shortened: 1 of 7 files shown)


packages/react-reconciler/src/ReactFiberCompleteWork.old.js
```diff
@@ -679,7 +679,7 @@ function completeWork(
           // If we hydrated, then we'll need to schedule an update for
           // the commit side-effects on the root.
           markUpdate(workInProgress);
-        } else if (!fiberRoot.hydrate) {
+        } else if (!fiberRoot.hydrate && workInProgress.child !== null) {
```
bvaughn, line 682: Hm...so the way this check is supposed to work is this:

If the current Fiber didn't render any children (`current.child === null`) then that means we are either mounting (in which case the container should be clear, or we should clear it) or updating (in which case, if `current.child` is null, our previous render didn't render any children- so clearing the container should be a noop anyway).

The work in progress fiber having a child just means "we are about to add children to this container".

gaearon:
Btw don't forget to update both files.

packages/react-reconciler/src/ReactFiberCompleteWork.old.js
trueadm: Shouldn't the work in progress have children? If it doesn't, aren't we rendering nothing over previously having nothing?

trueadm:
This was more for a reference to the debugging work with @bvaughn. I don't expect to merge it.

packages/react-reconciler/src/ReactFiberCompleteWork.old.js
bvaughn: Whether the work in progress has children is not relevant to the specific question of whether the container has non-React contents in it. That's what we're trying to clear here. If we know that React didn't add things to this container the last time we committed (or if we are mounting) then we clear the container to make sure there's nothing unexpected in it.

trueadm:
I've updated this PR with another possible fix.

bvaughn:
This PR fixes the regression we're seeing internally. I think there might be a better place to check for this though. Let me post a potential alternate PR here shortly...

PR #18782, opened 2020-04-29
Move legacy hidden API to new internal Fiber type

acdlite:
## Based on #18733 

[**Diff compared to previous PR in stack**](https://github.com/acdlite/react/compare/unhide-suspense-trees-without...legacy-hidden)

Adds a new LegacyHidden fiber type.

This only exists so we can clean up the internal implementation of `<div hidden={isHidden} />`, which is not a stable feature. The goal is to move everything to the new Offscreen type instead. However, Offscreen has different semantics, so before we can remove the legacy API, we have to migrate our internal usage at Facebook. So we'll need to maintain both temporarily.

If a host component receives a `hidden` prop, we wrap its children in an LegacyHidden fiber. This is similar to what we do for Suspense children.

The LegacyHidden type happens to share the same implementation as the new Offscreen type, for now, but using separate types allows us to fork the behavior later when we implement our planned changes to the Offscreen API.

There are two subtle semantic changes here. One is that the children of the host component wil have their visibility toggled using the same mechanism we use for Offscreen and Suspense: find the nearest host node children and give them a style of `display: none`. We didn't used to do this in the old API, because the `hidden` DOM attribute on the parent already hides them. So with this change, we're actually "overhiding" the children. I considered addressing this, but I figure I'll leave it as-is in case we want to expose the LegacyHidden component type temporarily to ease migration of Facebook's internal callers to the Offscreen type.

The other subtle semantic change is that, because of the extra fiber that wraps around the children, this pattern will cause the children to lose state:

```js
return isHidden ? <div hidden={true} /> : <div />;
```

The reason is that I didn't want to wrap every single host component in an extra fiber. So I only wrap them if a `hidden` prop exists. In the above example, that means the children are conditionally wrapped in an extra fiber, so they don't line up during reconciliation, so they get remounted every time `isHidden` changes.

The fix is to rewrite to:

```js
return <div hidden={isHidden} />;
```

I don't anticipate this will be a problem at Facebook, especially since we're only supposed to use `hidden` via a userspace wrapper component. (And since the bad pattern isn't very React-y, anyway.)

Again, the eventual goal is to delete this completely and replace it with Offscreen.

### DIFF
--- a/packages/react-reconciler/src/ReactFiber.new.js
+++ b/packages/react-reconciler/src/ReactFiber.new.js
@@ -56,6 +56,7 @@ import {
   ScopeComponent,
   Block,
   OffscreenComponent,
+  LegacyHiddenComponent,
 } from './ReactWorkTags';
 import getComponentName from 'shared/getComponentName';
 
@@ -90,6 +91,7 @@ import {
   REACT_SCOPE_TYPE,
   REACT_BLOCK_TYPE,
   REACT_OFFSCREEN_TYPE,
+  REACT_LEGACY_HIDDEN_TYPE,
 } from 'shared/ReactSymbols';
 
 export type {Fiber};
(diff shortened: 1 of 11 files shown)


acdlite:
Need to move the bailout here, too:

https://github.com/facebook/react/blob/b330a29a0378c5b7517ffd99341b745e61bbdc03/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L3263-L3280

acdlite:
@sebmarkbage I'm going to go ahead and merge this so Luna can build on top. It only affects the new reconciler. If you have feedback after the fact, I'll address in a follow up.

PR #18784, opened 2020-04-29
[Blocks] Use native fetch

gaearon:
Follow-up to https://github.com/facebook/react/pull/18774.
Rewrites from XHR to just using native fetch.

- Fetch now returns a Response object, a thin wrapper over native Response.
- New Response methods: `arrayBuffer()`, `blob()`, `text()`, `json()`.
  - I didn't add `formData()` because I'm not sure why it's useful for GETs.
  - <s>Note I currently forbid reading different types from the same URL. Could be fixable but let's wait until someone actually has a use case.</s> Delegating to the browser here so if it throws, it throws.
- New Response fields: `headers`, `ok`, `redirected`, `status`, `statusText`, `type`, `url`.
  - Maybe we could just expose the native object as a property.
- Fetch now passes the options through.
  - But it errors on a few that don't make sense:
    - `signal`: We're going to control this ourselves.
    - `body` and `method`: We haven't decided what to do with POST.

### DIFF
--- a/packages/react-data/src/fetch/ReactDataFetch.js
+++ b/packages/react-data/src/fetch/ReactDataFetch.js
@@ -94,34 +94,38 @@ function Response(nativeResponse) {
 Response.prototype = {
   constructor: Response,
   arrayBuffer() {
-    let entry = this._entries.get('arrayBuffer');
+    const entries = this._entries;
+    let entry = entries.get('arrayBuffer');
     if (!entry) {
       entry = createFromThenable(this._response.arrayBuffer());
-      this._entries.set('arrayBuffer', entry);
+      entries.set('arrayBuffer', entry);
     }
     return readResult(entry);
   },
   blob() {
-    let entry = this._entries.get('blob');
+    const entries = this._entries;
+    let entry = entries.get('blob');
     if (!entry) {
       entry = createFromThenable(this._response.blob());
-      this._entries.set('blob', entry);
+      entries.set('blob', entry);
     }
     return readResult(entry);
   },
   json() {
-    let entry = this._entries.get('json');
+    const entries = this._entries;
+    let entry = entries.get('json');
     if (!entry) {
       entry = createFromThenable(this._response.json());
-      this._entries.set('json', entry);
+      entries.set('json', entry);
     }
     return readResult(entry);
   },
   text() {
-    let entry = this._entries.get('text');
+    const entries = this._entries;
+    let entry = entries.get('text');
     if (!entry) {
       entry = createFromThenable(this._response.text());
-      this._entries.set('text', entry);
+      entries.set('text', entry);
     }
     return readResult(entry);
   },
--- a/packages/react-data/src/fetch/ReactDataFetch.js
+++ b/packages/react-data/src/fetch/ReactDataFetch.js
@@ -78,6 +78,14 @@ function readResult(result: Result) {
   }
 }
 
+function getJson(nativeResponse) {
+  return nativeResponse.json();
+}
+
+function getText(nativeResponse) {
+  return nativeResponse.json();
+}
+
 function Response(nativeResponse) {
   this.headers = nativeResponse.headers;
   this.ok = nativeResponse.ok;
@@ -87,29 +95,32 @@ function Response(nativeResponse) {
   this.type = nativeResponse.type;
   this.url = nativeResponse.url;
 
-  this.json = function() {
-    return read('json', () => nativeResponse.json());
-  };
+  this._entry = null;
+  this._consumed = null;
+  this._response = nativeResponse;
+}
 
-  this.text = function() {
-    return read('text', () => nativeResponse.text());
-  };
-
-  let entry: void | Result;
-  let consumedType: void | string;
-
-  function read(type, getThenable) {
-    if (consumedType != null && type !== consumedType) {
+Response.prototype = {
+  json() {
+    return this._read('json', getJson);
+  },
+  text() {
+    return this._read('text', getText);
+  },
+  _read(type: string, getThenable) {
+    const consumed = this._consumed;
+    if (consumed != null && type !== consumed) {
       throw new Error('Already read.');
     }
+    let entry = this._entry;
     if (entry == null) {
-      consumedType = type;
-      const thenable = getThenable();
-      entry = createFromThenable(thenable, r => r);
+      this._consumed = type;
+      const thenable = getThenable(this._response);
+      entry = this._entry = createFromThenable(thenable, r => r);
     }
     return readResult(entry);
-  }
-}
+  },
+};
 
 export function fetch(url: string, options: mixed): Object {
   const map = readResultMap();
--- a/packages/react-data/src/fetch/ReactDataFetch.js
+++ b/packages/react-data/src/fetch/ReactDataFetch.js
@@ -49,9 +49,9 @@ function readResultMap(): Map<string, Result> {
 function toResult(thenable): Result {
   const result: Result = {
     status: Pending,
-    value: (null: any),
+    value: thenable,
   };
-  result.value = thenable.then(
+  thenable.then(
     value => {
       if (result.status === Pending) {
         const resolvedResult = ((result: any): ResolvedResult);
--- a/packages/react-data/src/fetch/ReactDataFetch.js
+++ b/packages/react-data/src/fetch/ReactDataFetch.js
@@ -46,7 +46,7 @@ function readResultMap(): Map<string, Result> {
   return map;
 }
 
-function createFromThenable(thenable, wrapValue): Result {
+function createFromThenable(thenable): Result {
   const result: Result = {
     status: Pending,
     value: (null: any),
--- a/packages/react-data/src/fetch/ReactDataFetch.js
+++ b/packages/react-data/src/fetch/ReactDataFetch.js
@@ -32,6 +32,8 @@ type RejectedResult = {|
 
 type Result = PendingResult | ResolvedResult | RejectedResult;
 
+// TODO: this is a browser-only version. Add a separate Node entry point.
+const nativeFetch = window.fetch;
 const fetchKey = {};
 
 function readResultMap(): Map<string, Result> {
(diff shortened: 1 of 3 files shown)


packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,88 @@ function readResultMap(): Map<string, Result> {
...
+function Response(nativeResponse) {
+  this.headers = nativeResponse.headers;
+  this.ok = nativeResponse.ok;
+  this.redirected = nativeResponse.redirected;
+  this.status = nativeResponse.status;
+  this.statusText = nativeResponse.statusText;
+  this.type = nativeResponse.type;
+  this.url = nativeResponse.url;
+
+  this.json = function() {
```
sebmarkbage, line 90: Let's put these on the prototype like the original. (Perhaps even two chains.)

We typically don't but we're emulating someone else's API here and the more we add, the slower it'll get to not use a prototype.

sebmarkbage: It's probably fine to use classes that we transpile. We have a few other non-core modules that does that. Once we compile to ESM we can probably go native classes.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,99 @@ function readResultMap(): Map<string, Result> {
...
+  this.statusText = nativeResponse.statusText;
+  this.type = nativeResponse.type;
+  this.url = nativeResponse.url;
+
+  this._entry = null;
+  this._consumed = null;
+  this._response = nativeResponse;
+}
+
+Response.prototype = {
```
sebmarkbage, line 111: nit: `constructor: Response,`

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,99 @@ function readResultMap(): Map<string, Result> {
...
+  },
+  _read(type: string, getThenable) {
+    const consumed = this._consumed;
+    if (consumed != null && type !== consumed) {
+      throw new Error('Already read.');
+    }
+    let entry = this._entry;
+    if (entry == null) {
+      this._consumed = type;
+      const thenable = getThenable(this._response);
```
sebmarkbage, line 132: Passing first class functions break inlining and what not. Just make a switch in here.

I'd just duplicate the code instead though and then extract from the inside out. E.g. make make this a helper function and call that in both functions: 

```
    const consumed = this._consumed;
    if (consumed != null && type !== consumed) {
      throw new Error('Already read.');
    }
    this._consumed = type;
```

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,99 @@ function readResultMap(): Map<string, Result> {
...
+}
+
+Response.prototype = {
+  json() {
+    return this._read('json', getJson);
+  },
+  text() {
+    return this._read('text', getText);
+  },
+  _read(type: string, getThenable) {
```
sebmarkbage, line 124: I'd move this to closure scope so that nobody abuses it.

packages/react-data/src/fetch/ReactDataFetch.js
gaearon: I was code golfing here a bit because the byte size of this helper seems important. Maybe not so much.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,114 @@ function readResultMap(): Map<string, Result> {
...
+      if (options.method || options.body || options.signal) {
+        // TODO: wire up our own cancellation mechanism.
+        // TODO: figure out what to do with POST.
+        throw Error('Unsupported option');
+      }
+    }
+    const thenable = nativeFetch(url, options);
+    entry = createFromThenable(
+      thenable,
+      nativeResponse => new Response(nativeResponse),
```
sebmarkbage, line 154: The way I think of these async points is that it gets simpler if everything is just at the deepest point. Then you can just use Suspense on top of that.

E.g. this is just like a `let nativeResponse = read(url); return new Response(nativeResponse)` like any other wrapper object around this would do in a suspensey function.

The difference is that you want to also memoize that object. Memoizing in this world is just passing deps into a resource and getting value out.

```
let nativeResponse =  readFetch(url);
return readResponse(nativeResponse);
```

I.e. there's a cache from nativeResponse to Response.

You could implement that as an expando on the nativeResponse.

```
let nativeResponse = readResult(entry);
if (nativeResponse._reactResponse) {
  return nativeResponse._reactResponse;
} else {
  return nativeResponse._reactResponse = new Response(nativeResponse);
}
```

That way you avoid the closure and it's a pattern you can use for many cases even if you need the underlying value for many different wrappers.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,113 @@ function readResultMap(): Map<string, Result> {
...
+
+  this._entry = null;
+  this._format = null;
+  this._response = nativeResponse;
+}
+
+function getConsumed(response, format) {
+  const consumedFormat = response._format;
+  if (consumedFormat != null && format !== consumedFormat) {
+    throw new Error('Already read.');
```
sebmarkbage, line 98: Why do we have to implement this ourselves? Can't we just keep an entry per format and let the underlying error throw?

gaearon: Didn't want an extra object but yeah this would be cleaner

sebmarkbage: Also that way, when Safari supports clone, we could just call `this._response.clone().arrayBuffer()` to make it work.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,115 @@ function readResultMap(): Map<string, Result> {
...
-      status: Pending,
-      value: wakeable,
-    };
-    map.set(url, result);
-    const xhr = new XMLHttpRequest();
-    xhr.onload = function() {
-      // TODO: should we handle status codes?
-      if (result.status !== Pending) {
-        return;
+function createFromThenable(thenable, wrapValue): Result {
```
sebmarkbage, line 49: wrapValue is unused now.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,106 @@ function readResultMap(): Map<string, Result> {
...
+function Response(nativeResponse) {
+  this.headers = nativeResponse.headers;
+  this.ok = nativeResponse.ok;
+  this.redirected = nativeResponse.redirected;
+  this.status = nativeResponse.status;
+  this.statusText = nativeResponse.statusText;
+  this.type = nativeResponse.type;
+  this.url = nativeResponse.url;
+
+  this._entries = new Map();
```
sebmarkbage, line 90: This is a fixed set. It'd be more efficient to just have four fields for it.

gaearon: LOL this makes sese.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,98 @@ function readResultMap(): Map<string, Result> {
...
+    );
+  },
+  json() {
+    return readResult(
+      this._json || (this._json = toResult(this._response.json())),
+    );
+  },
+  text() {
+    return readResult(
+      this._text || (this._text = toResult(this._response.text())),
```
gaearon, line 117: I guess text could be `''` so that's not right.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,98 @@ function readResultMap(): Map<string, Result> {
...
+    );
+  },
+  blob() {
+    return readResult(
+      this._blob || (this._blob = toResult(this._response.blob())),
+    );
+  },
+  json() {
+    return readResult(
+      this._json || (this._json = toResult(this._response.json())),
```
gaearon, line 112: Can JSON response be null?

packages/react-data/src/fetch/ReactDataFetch.js
gaearon: Ignore me. These are entries. :D

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,98 @@ function readResultMap(): Map<string, Result> {
...
-    xhr.onload = function() {
-      // TODO: should we handle status codes?
-      if (result.status !== Pending) {
-        return;
+function toResult(thenable): Result {
+  const result: Result = {
+    status: Pending,
+    value: (null: any),
+  };
+  result.value = thenable.then(
```
sebmarkbage, line 54: I tend do use the original thenable as the value instead of the return value. The first one will be held onto anyway but in theory the new one created by this one can be released. I'm not sure what's better though because some polyfills special cases a single listener. Not sure what native does.

gaearon: I think I was worried that React would be able to run before our continuation. But that's not possible because they run in order anyway, right? And we schedule ours first.

gaearon: I'll change it because my original concern doesn't make sense, so it should be ok. If not we'll learn something.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -44,50 +46,98 @@ function readResultMap(): Map<string, Result> {
...
-    map.set(url, result);
-    const xhr = new XMLHttpRequest();
-    xhr.onload = function() {
-      // TODO: should we handle status codes?
-      if (result.status !== Pending) {
-        return;
+function toResult(thenable): Result {
+  const result: Result = {
+    status: Pending,
+    value: (null: any),
```
sebmarkbage: That way you can assign this to thenable without the any.

PR #18789, opened 2020-04-30
[Blocks] Cancellation

gaearon:
This is just the initial wiring. I don't know if it actually makes sense yet. I don't understand *where* and *when* we should cancel. My first naïve approach was to cancel in the reducer, but that doesn't work because I can't read from the same cache anymore. So I think I'm missing something in how this relates to Here/There. But posting anyway because it might be easier to discuss in code.

Maybe it would help to change my example app to a more traditional navigation model. E.g. master/detail with links.

### DIFF
--- a/fixtures/blocks/src/Post.js
+++ b/fixtures/blocks/src/Post.js
@@ -11,10 +11,7 @@ import {fetch, preload} from 'react-data/fetch';
 import loadComments from './Comments';
 
 function load(params) {
-  if (params.id < 3) {
-    const nextId = params.id + 1;
-    preload('http://localhost:3001/posts/' + nextId);
-  }
+  preload('http://localhost:3001/comments?postId=' + nextId);
   const postResponse = fetch('http://localhost:3001/posts/' + params.id);
   return {
     post: postResponse.json(),
--- a/packages/react-data/src/fetch/ReactDataFetch.js
+++ b/packages/react-data/src/fetch/ReactDataFetch.js
@@ -34,10 +34,15 @@ type Result = PendingResult | ResolvedResult | RejectedResult;
 
 // TODO: this is a browser-only version. Add a separate Node entry point.
 const nativeFetch = window.fetch;
+const hasAbortController = typeof window.AbortController === 'function';
 const fetchKey = {};
 
-function readResultMap(): Map<string, Result> {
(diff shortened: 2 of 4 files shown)


fixtures/blocks/src/Post.js
```diff
@@ -11,10 +11,7 @@ import {fetch, preload} from 'react-data/fetch';
 import loadComments from './Comments';
 
 function load(params) {
-  if (params.id < 3) {
-    const nextId = params.id + 1;
-    preload('http://localhost:3001/posts/' + nextId);
-  }
+  preload('http://localhost:3001/comments?postId=' + nextId);
```
gaearon, line 14: Changed this because it didn't make sense. Preload should be used for things in current request, not next pages. This wouldn't work with the server model.

packages/react-data/src/fetch/ReactDataFetch.js
```diff
@@ -120,16 +125,24 @@ Response.prototype = {
...
   if (!entry) {
     if (options) {
       if (options.method || options.body || options.signal) {
-        // TODO: wire up our own cancellation mechanism.
         // TODO: figure out what to do with POST.
         throw Error('Unsupported option');
       }
     }
+    if (hasAbortController) {
+      const controller = new AbortController();
```
gaearon, line 139: Maybe I could create one controller per Cache Map entry? Seems unnecessary to create many of them if they're all going down together.

packages/react/src/cache/ReactCache.js
```diff
@@ -13,15 +13,27 @@ import invariant from 'shared/invariant';
...
 const CacheContext: ReactContext<null | Cache> = createContext(null);
 
 function CacheImpl() {
   this.resources = new Map();
-  // TODO: cancellation token.
+  this.disposeCallbacks = [];
 }
+CacheImpl.prototype = {
+  constructor: CacheImpl,
+  dispose() {
```
sebmarkbage, line 29: If we do have a helper, we should make it a tree-shakable method because it's possible you get stuck with a cache impl as part of your dependencies even if you don't actually use it.  So It's nicer if this is just a protocol. It's likely that you only have one usage of the cache and in that case it might be better that it's inline.

We allow arbitrary access to the Map so why not this?.

PR #18792, opened 2020-04-30
Prevent stale legacy root from clearing a container (DRAFT)

bvaughn:
Fixes the legacy root bug we observed after the recent sync (caused by #18730).

Alternate fix to #18781.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMFiber-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMFiber-test.js
@@ -1280,4 +1280,36 @@ describe('ReactDOMFiber', () => {
     );
     expect(didCallOnChange).toBe(true);
   });
+
+  it('unmounted legacy roots should never clear newer root content from a container', () => {
+    const ref = React.createRef();
+
+    function OldApp() {
+      const hideOnFocus = () => {
+        // This app unmounts itself inside of a focus event.
+        ReactDOM.unmountComponentAtNode(container);
+      };
+
+      return (
+        <button onFocus={hideOnFocus} ref={ref}>
+          old
+        </button>
+      );
+    }
+
+    function NewApp() {
+      return <button ref={ref}>new</button>;
+    }
+
+    ReactDOM.render(<OldApp />, container);
+    ref.current.focus();
+
+    ReactDOM.render(<NewApp />, container);
+
+    // Calling focus again will flush previously scheduled discerete work for the old root-
+    // but this should not clear out the newly mounted app.
+    ref.current.focus();
+
+    expect(container.textContent).toBe('new');
+  });
 });
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.new.js
@@ -687,8 +687,21 @@ function completeWork(
           // It's also safe to do for updates too, because current.child would only be null
           // if the previous render was null (so the the container would already be empty).
           //
-          // The additional root.hydrate check is required for hydration in legacy mode with no fallback.
-          workInProgress.effectTag |= Snapshot;
+          // The additional root.hydrate check above is required for hydration in legacy mode with no fallback.
+          //
+          // The root container check below also avoids a potential legacy mode problem
+          // where unmounting from a container then rendering into it again
+          // can sometimes cause the container to be cleared after the new render.
+          const containerInfo = fiberRoot.containerInfo;
+          const legacyRootContainer =
+            containerInfo == null ? null : containerInfo._reactRootContainer;
+          if (
+            legacyRootContainer == null ||
+            legacyRootContainer._internalRoot == null ||
+            legacyRootContainer._internalRoot === fiberRoot
+          ) {
+            workInProgress.effectTag |= Snapshot;
+          }
         }
       }
       updateHostContainer(workInProgress);
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.old.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.old.js
@@ -685,8 +685,21 @@ function completeWork(
           // It's also safe to do for updates too, because current.child would only be null
           // if the previous render was null (so the the container would already be empty).
           //
-          // The additional root.hydrate check is required for hydration in legacy mode with no fallback.
-          workInProgress.effectTag |= Snapshot;
+          // The additional root.hydrate check above is required for hydration in legacy mode with no fallback.
+          //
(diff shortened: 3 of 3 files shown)


bvaughn:
The thing I'm not able to repro in a test yet (but that we're seeing within Facebook) is `completeWork()` running for the initial root *after* the new root has been committed. The sequence in both cases is:

```
unmountComponentAtNode() root 1
unmountComponentAtNode() -> unbatched update root 1
legacyRenderSubtreeIntoContainer() update root 1
render() null root
legacyRenderSubtreeIntoContainer() mount root 2
completeWork() Marking root to be cleared root 2
clearContainer() root 2
completeWork() root 1 // I can't reproduce this in a test though
```

trueadm:
@bvaughn Nice findings. It might have something to do with the fact that when we call `unmountComponentAtNode`, we schedule a discrete event on root 1. If you call `focus` later (after we mount root 2), it flushes the pending scheduled work in root 1, causing it to render again, thus causing it to hit the completion phase on root 1.

bvaughn:
> when we call unmountComponentAtNode, we schedule a discrete event on root 1

Why do we do this? Could you point me at where? I must be missing it.

> If you call `focus` later (after we mount root 2), it flushes the pending scheduled work in root 1, causing it to render again

I guess I'm missing the part where we schedule a discrete event on the first root, because triggering a focus on the 2nd root doesn't flush anything.

bvaughn:
I think I got it!

bvaughn:
Thanks @trueadm, that comment was very helpful in helping me realize the thing I was missing in my test.

trueadm:
This is awesome! I wonder if we use `createRoot` do we have a similar issue?

bvaughn:
> I wonder if we use createRoot do we have a similar issue?

No, I don't think so. This only applies to legacy roots.

Also if you tried to create a new root (`createRoot(container)`) before the old one finished flushing the unmount, we warn about it.

packages/react-reconciler/src/ReactFiberCompleteWork.new.js
```diff
@@ -687,8 +687,21 @@ function completeWork(
...
-          // The additional root.hydrate check is required for hydration in legacy mode with no fallback.
-          workInProgress.effectTag |= Snapshot;
+          // The additional root.hydrate check above is required for hydration in legacy mode with no fallback.
+          //
+          // The root container check below also avoids a potential legacy mode problem
+          // where unmounting from a container then rendering into it again
+          // can sometimes cause the container to be cleared after the new render.
+          const containerInfo = fiberRoot.containerInfo;
+          const legacyRootContainer =
+            containerInfo == null ? null : containerInfo._reactRootContainer;
```
trueadm, line 697: The only reason I opted to do this in the renderer in my PR was because this seemed host-specific. I see you reasoning for doing it earlier though, it avoids doing unecessary work before it's too late. Another approach could be to expose a host config function that does this check (as this is only related to the legacy roots, which doesn't necessarily affect other renderers) that can do this check for us.

bvaughn: Yeah, I guess it is DOM renderer specific (then again, so is `clearContainer` itself really). I just thought it would be better to do this check earlier rather than in the perf-sensitive commit phase. Will be great to remove this check entirely eventually.

bvaughn:
Thanks for the review, Dominic. I'm going to merge this to hopefully unblock you on the sync. (Let me know if you need someone to take over the sync btw.)

I can follow up later if there are concerns or suggested changes from others.

packages/react-dom/src/__tests__/ReactDOMFiber-test.js
```diff
@@ -1280,4 +1280,36 @@ describe('ReactDOMFiber', () => {
...
+    function NewApp() {
+      return <button ref={ref}>new</button>;
+    }
+
+    ReactDOM.render(<OldApp />, container);
+    ref.current.focus();
+
+    ReactDOM.render(<NewApp />, container);
+
+    // Calling focus again will flush previously scheduled discerete work for the old root-
```
sebmarkbage, line 1309: Something is off here because there's no concept of "discrete work" for legacy mode (and I'm removing it for CM too). At least there shouldn't be. Maybe for passive effects but that doesn't kick in here.

There's only Sync work in legacy mode.

What do you mean by discrete work here? It might that something is misnamed in a method or something.

packages/react-reconciler/src/ReactFiberCompleteWork.new.js
sebmarkbage, line 697: This field is DOM specific. Not sure what happens to this code in RN. I guess it would just schedule a Snapshot which is currently a noop?

bvaughn: Yes. RN doesn't have a way to clear its container anyway (since there's no native API).

packages/react-dom/src/__tests__/ReactDOMFiber-test.js
bvaughn: @sebmarkbage This comment was based on [Dominic's comment above](https://github.com/facebook/react/pull/18792#issuecomment-622052111).

trueadm: @sebmarkbage Upon calling `scheduleUpdateOnFiber`, you add an entry to `rootsWithPendingDiscreteUpdates` which is the problem in this issue. I'd assume your work of removing the concept of discrete work didn't include this particulat code-path?

packages/react-reconciler/src/ReactFiberCompleteWork.new.js
trueadm: I mentioned this too in my previous comment. It's something we can address in a follow up.

bvaughn: Even though it is kind of renderer-specific, I don't know if I agree that moving it into the host config (and commit phase) is a good follow up change. Happy to talk more about it though.

trueadm: Yeah, it sucks either way. I guess a complete phase legacy check is less bad as it has less overhead and we know we'll likely be removing it in the near future.

packages/react-dom/src/__tests__/ReactDOMFiber-test.js
sebmarkbage: I haven't landed that part yet. Still blocked on some stuff.

Concretely I'm worried that we've accidentally landed "discrete" semantics in legacy mode at some point. I.e. some subtle breaking change.

PR #18794, opened 2020-04-30
[react-refresh/babel] Always insert registrations after var

sophiebits:
Previously, we transformed

```
let Foo = styled.div``;
```

to

```
let Foo = _c1 = styled.div``;
```

and then babel-plugin-styled-components would infer `_c1` as the display name. Widen the existing case that applies to function expressions to apply to any type of variable declaration.

### DIFF
--- a/packages/react-refresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-refresh/src/ReactFreshBabelPlugin.js
@@ -689,16 +689,15 @@ export default function(babel, opts = {}) {
               return;
             }
             const handle = createRegistration(programPath, persistentID);
-            if (
-              (targetExpr.type === 'ArrowFunctionExpression' ||
-                targetExpr.type === 'FunctionExpression') &&
-              targetPath.parent.type === 'VariableDeclarator'
-            ) {
-              // Special case when a function would get an inferred name:
+            if (targetPath.parent.type === 'VariableDeclarator') {
+              // Special case when a variable would get an inferred name:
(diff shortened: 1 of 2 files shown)


packages/react-refresh/src/ReactFreshBabelPlugin.js
```diff
@@ -689,16 +689,15 @@ export default function(babel, opts = {}) {
               return;
             }
             const handle = createRegistration(programPath, persistentID);
-            if (
-              (targetExpr.type === 'ArrowFunctionExpression' ||
-                targetExpr.type === 'FunctionExpression') &&
-              targetPath.parent.type === 'VariableDeclarator'
-            ) {
-              // Special case when a function would get an inferred name:
+            if (targetPath.parent.type === 'VariableDeclarator') {
```
gaearon, line 692: Seems like comment on line 708 is now outdated? What does the else code path match now?

sophiebits: We add two references for a HOC, one inside and one outside. (findInnerComponents recurses with the same callback.) The else path still applies to the inside one. I'll update the comment.

sophiebits: i.e. you end up with

```
let Foo = hoc(_c = () => {})
_c2 = Foo;
```

We could probably do more to preserve the names of inner components too but this should be 95% of the common code patterns.

PR #18796, opened 2020-05-01
Initial Lanes implementation

acdlite:
*This is ready for review. I'm going to do another pass to add comments and look for anything unusual I may have missed, but overall I'm happy with its current state*.

All of the changes I've made in this PR are behind the `enableNewReconciler` flag. Merging this to master will not affect the open source builds or the build that we ship to Facebook.

The only build that is affected is the `ReactDOMForked` build, which is deployed to Facebook **behind an experimental flag (currently disabled for all users)**. We will use this flag to gradually roll out the new reconciler, and quickly roll it back if we find any problems.

Because we have those protections in place, what I'm aiming for with this initial PR is the **smallest possible atomic change that lands cleanly and doesn't rely on too many hacks**. The goal has not been to get every single test or feature passing, and it definitely is not to implement all the features that we intend to build on top of the new model. When possible, I have chosen to preserve existing semantics and defer changes to follow-up steps. (Listed in the section below.)

(I did not end up having to disable any tests, although if I had, that should not have necessarily been a merge blocker.)

For example, even though one of the primary goals of this project is to improve our model for parallel Suspense transitions, in this initial implementation, I have chosen to keep the same core heuristics for sequencing and flushing that existed in the ExpirationTimes model: low priority updates cannot finish without also finishing high priority ones.

Despite all these precautions, **because the scope of this refactor is inherently large, I do expect we will find regressions.** The flip side is that I also expect the new model to improve the stability of the codebase and make it easier to fix bugs when they arise.

## Technical description of Lanes

This is not a comprehensive description. That would take up many pages. What I've written here is a brief overview intended to help React team members translate from the old model (Expiration Times) to the new one (Lanes). I can write a longer technical document once more of follow-up steps done.

There are two primary advantages of the Lanes model over the Expiration Times model:

- Lanes decouple the concept of task prioritization ("Is task A higher priority than task B?") from task batching ("Is task A part of this group of tasks?").
- Lanes can express many distinct task threads with a single, 32-bit data type.

In the old model, to decide whether to include a given unit of work in the batch that's being worked on, we would compare their relative priorities:

```js
const isTaskIncludedInBatch = priorityOfTask >= priorityOfBatch;
```

This worked because of a constraint we imposed where lower priority tasks were not allowed to complete unless higher priority tasks are also included. Given priorities A > B > C, you couldn't work on B without also working on A; nor could you work on C without working on both B and A.

This constraint was designed before Suspense was a thing, and it made some sense in that world. When all your work is CPU bound, there's not much reason to work on tasks in any order other than by their priority. But when you introduce tasks that are IO-bound (i.e. Suspense), you can have a scenario where a higher priority IO-bound task blocks a lower-priority CPU-bound task from completing.

A similar flaw of Expiration Times is that it's limited in how we can express a group of multiple priority levels.

Using a Set object isn't practical, in terms of either memory or computation — the existence checks we're dealing with are extremely pervasive, so they need to be fast and use as little memory as possible.

As a compromise, often what we'd do instead is maintain a range of priority levels:

```js
const isTaskIncludedInBatch = taskPriority <= highestPriorityInRange && taskPriority >= lowestPriorityInRange;
```

Setting aside that this requires two separate fields, even this is quite limiting in its expressiveness. You can express a closed, continuous range of tasks. But you can't represent a finite set of distinct tasks. For example, given a range of tasks, how do you remove a task that lies in the middle of the range? Even in cases where we had designed a decent workaround, reasoning about groups of tasks this way became extremely confusing and prone to regressions.

At first by design, but then more by accident, the old model coupled the two concepts of 1) prioritization and 2) batching into a single data type. We were limited in our ability to express one except in terms that affected the other.

In the new model, we have decoupled those two concepts. Groups of tasks are instead expressed not as relative numbers, but as bitmasks:

```js
const isTaskIncludedInBatch = (task & batchOfTasks) !== 0;
```

The type of the bitmask that represents a task is called a `Lane`. The type of the bitmask that represents a batch is called `Lanes`. (Note: these names are not final. I know using a plural is a bit confusing, but since the name appears all over the place, I wanted something that was short. But I'm open to suggestions.)

In more concrete React terms, an update object scheduled by `setState` contains a `lane` field, a bitmask with a single bit enabled. This replaces the `update.expirationTime` field in the old model.

On the other hand, a fiber is not associated with only a single update, but potentially many. So it has a `lanes`  field, a bitmask with zero or more bits enabled (`fiber.expirationTime` in the old model); and a `childLanes` field (`fiber.childExpirationTime`).

Lanes is an opaque type. You can only perform direct bitmask manipulation inside the `ReactFiberLane` module. Elsewhere, you must import a helper function from that module. This is a trade off, but one that I think is ultimately worth it, since dealing with Lanes can be very subtle, and colocating all the logic will make it easier for us to tweak our heuristics without having to do a huge refactor (like this one) every time.

### Commonly seen Expiration Time fields, translated to Lanes

- `renderExpirationtime` -> `renderLanes`
- `update.expirationTime` -> `update.lane`
- `fiber.expirationTime` -> `fiber.lanes`
- `fiber.childExpirationTime` -> `fiber.childLanes`
- `root.firstPendingTime` and `root.lastPendingTime` -> `fiber.pendingLanes`

### Things that probably block us from rolling this out at Facebook
  - **Protection against starvation by high pri React updates.** If one update is interrupted by another, that update will never expire. We do get starvation protection from non-React work, via the Scheduler package. However, we shouldn't rely on that, either, since the native Scheduler may or may not have a timeout mechanism.

### Future improvements that will build on top

Not a comprehensive list

- **Entanglement**, a mechanism to enforce that a lane depends on another one finishing in the same batch. Improvements that will be unlocked by this feature include:
  - **Skipping ("canceling") intermediate `useTransition` states**. Only the most recent transition in a sequence of related transitions should be allowed to finish.
  - **De-opting `useMutableSource` less frequently**.
- **Working on lanes out of order**. This will require us to design a new model for retrying suspended lanes when data is received, since it's infeasible to try every possible combination. Improvements that will be unlocked by this feature include:
  - **Parallel UI transitions via `useTransition`**. Currently, a single transition can easily block other transitions from finishing, even if they are in totally unrelated parts of the tree.
  - **Arbitrarily long-lived transitions that run in the background, like for refreshing data**. Related to previous item.
- **Eagerly computing state updates scheduled via the `useState` hook**. We may do this via entanglement, or we may take advantage of the fact that there are a finite number of non-overlapping batches that an update might render into and precompute each one whenever an additional pending update is scheduled before the queue is flushed.


packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,677 @@
...
+    }
+    i++;
+    lane <<= 1;
+  }
+  return NoLane;
+}
+
+function getLowestPriorityLane(lanes: Lanes): Lane {
+  // This finds the most significant non-zero bit.
+  // TODO: Consider alternate implementations. Could use a map. Is Math.log2
```
jacobp100, line 524: I think `Math.clz32` could work!

acdlite: Oh I didn’t realize so many browsers supported this. We can use a fallback when it’s not there. Good idea, thanks!

packages/react-reconciler/src/ReactFiberReconciler.new.js
```diff
@@ -336,36 +337,33 @@ export function attemptSynchronousHydration(fiber: Fiber): void {
...
-      flushSync(() =>
-        scheduleUpdateOnFiber(fiber, (Sync: ExpirationTimeOpaque)),
-      );
+      flushSync(() => scheduleUpdateOnFiber(fiber, SyncLane));
       // If we're still blocked after this, we need to increase
       // the priority of any promises resolving within this
       // boundary so that they next attempt also has higher pri.
-      const retryExpTime = UserBlockingUpdateTime;
+      const retryExpTime = InputDiscreteHydrationLane;
       markRetryTimeIfNotHydrated(fiber, retryExpTime);
```
bvaughn, line 350:
```suggestion
      const retryLane = InputDiscreteHydrationLane;
      markRetryTimeIfNotHydrated(fiber, retryLane);
```

packages/react-reconciler/src/ReactFiberReconciler.new.js
```diff
@@ -249,7 +250,7 @@ export function updateContainer(
     }
   }
   const suspenseConfig = requestCurrentSuspenseConfig();
-  const expirationTime = requestUpdateExpirationTime(current, suspenseConfig);
+  const expirationTime = requestUpdateLane(current, suspenseConfig);
```
bvaughn, line 253:
```suggestion
  const lane = requestUpdateLane(current, suspenseConfig);
```

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -500,34 +468,33 @@ export function scheduleUpdateOnFiber(
     // This is the result of a discrete event. Track the lowest priority
     // discrete update per root so we can flush them early, if needed.
     if (rootsWithPendingDiscreteUpdates === null) {
```
bvaughn, line 470: This if/else needs to move up into the `else` above now (see PR #18797)

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1539,19 +1366,23 @@ export function renderHasNotSuspendedYet(): boolean {
...
+  if (workInProgressRoot === null) {
+    prepareFreshStack(root, lanes);
+    startWorkOnPendingInteractions(root, lanes);
+  } else if (workInProgressRoot !== root) {
+    workInProgressRoot.wipLanes = NoLanes;
+    prepareFreshStack(root, lanes);
+    startWorkOnPendingInteractions(root, lanes);
+  } else if (lanes !== root.wipLanes) {
+    startWorkOnPendingInteractions(root, lanes);
+    prepareFreshStack(root, lanes);
```
bvaughn, line 1385: Calling `prepareFreshStack` *after* doing something looks odd. Can you explain why we do it?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2769,9 +2532,14 @@ export function pingSuspendedRoot(
     pingCache.delete(wakeable);
   }
 
+  // TODO: Instead of passing the current time as an argument, we should read it
```
bvaughn, line 2535: nit "time"

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -3435,19 +3177,20 @@ function startWorkOnPendingInteractions(
...
-  const earliestRemainingTimeAfterCommit = root.firstPendingTime_opaque;
+  const remainingLanesAfterCommit = root.pendingLanes;
 
   let subscriber;
 
   try {
     subscriber = __subscriberRef.current;
     if (subscriber !== null && root.memoizedInteractions.size > 0) {
-      const threadID = computeThreadID(root, committedExpirationTime);
+      // FIXME: More than one lane can finish in a single commit.
```
bvaughn, line 3192: How often do you think this will break things?

bvaughn:
This change looks great, by the way. Really excited about the new direction.

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,677 @@
...
+}
+
+// This returns the highest priority pending lanes regardless of whether they
+// are suspended.
+export function getHighestPriorityPendingLanes(root: FiberRoot) {
+  return getHighestPriorityLanes(root.pendingLanes);
+}
+
+export function getLanesToRetrySynchronouslyOnError(root: FiberRoot): Lanes {
+  const everythingButOffscreen = root.pendingLanes & ~OffscreenLane;
```
lunaruan, line 378: Why are we retrying `IdleLanes` if we are not retrying `OffscreenLanes`?  I tried to grep to see where we were using `IdleLands` outside of `schedulerPriorityToLanePriority` and I couldn't find other use cases, so I might be missing something here.

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,677 @@
...
+// Includes all non-Idle updates
+const UpdateRangeEnd = 26;
+const NonIdleLanes = /*                                 */ 0b000011111111111111111111111111;
+
+export const IdleHydrationLane: Lane = /*               */ 0b000100000000000000000000000000;
+const IdleLanes: Lanes = /*                             */ 0b011100000000000000000000000000;
+const IdleUpdateRangeStart = 27;
+const IdleUpdateRangeEnd = 29;
+
+export const OffscreenLane: Lane = /*                   */ 0b100000000000000000000000000000;
```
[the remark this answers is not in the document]
bvaughn: "shelf"? 🤨 I'd be interested in hearing the reasoning behind this! 😆 Offscreen seemed pretty clear.

packages/react-reconciler/src/ReactFiberLane.js
acdlite: The idea is to include anything that might include a pending update that would fix the error that we’re trying to recover from. Offscreen never includes updates, only deferred subtrees.

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,677 @@
...
+export const NoLanes: Lanes = /*                        */ 0b000000000000000000000000000000;
+export const NoLane: Lane = /*                          */ 0b000000000000000000000000000000;
+
+export const SyncLane: Lane = /*                        */ 0b000000000000000000000000000001;
+const SyncUpdateRangeEnd = 1;
+export const SyncBatchedLane: Lane = /*                 */ 0b000000000000000000000000000010;
+const SyncBatchedUpdateRangeEnd = 2;
+
+export const InputDiscreteHydrationLane: Lane = /*      */ 0b000000000000000000000000000100;
+const InputDiscreteLanes: Lanes = /*                    */ 0b000000000000000000000000011100;
```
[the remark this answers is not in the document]
acdlite: One is for a trick we use during hydration. If we receive an update to a tree that hasn’t been hydrated yet, we restart at a slightly higher priority to hydrate it, then continue with the update on top. So every async priority has a hydration lane associated with it.

The other extra lane was reserved for a trick that SuspenseList does where it leaves behind work that is slightly lower priority, but that turned out to not be necessary, so I removed it: https://github.com/facebook/react/pull/18738

Regardless, this whole range is going to go away when we make discrete inputs synchronous, so we can redistribute them to other priorities.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
acdlite: I wasn’t sure about this. If it’s not already a problem then this doesn’t make it worse, those, since multiple expirations times could finish in the same batch, too.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -580,71 +570,57 @@ function updateOffscreenComponent(
...
       if (prevState !== null) {
-        const prevBaseTime = prevState.baseTime;
-        nextBaseTime = !isSameOrHigherPriority(
-          prevBaseTime,
-          renderExpirationTime,
-        )
-          ? prevBaseTime
-          : renderExpirationTime;
+        const prevBaseLanes = prevState.baseLanes;
+        nextBaseLanes = combineLanes(prevBaseLanes, renderLanes);
```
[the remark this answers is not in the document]
acdlite: It’s never too late

packages/react-reconciler/src/__tests__/ReactIncrementalUpdates-test.js
```diff
@@ -372,12 +392,26 @@ describe('ReactIncrementalUpdates', () => {
...
-        'render',
-      ]),
+      expect(Scheduler).toFlushAndYield(
+        gate(flags =>
+          flags.new
+            ? [
+                'setState updater',
+                // In the new reconciler, updates inside the render phase are
+                // treated as if they came from an event, so the update gets
+                // shifted to a subsequent render.
```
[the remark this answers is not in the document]
acdlite: useSubscription doesn’t call setState during the render phase right? I think the only place we do this intentionally is useOpaqueIdentifier. Everywhere else we warn.

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -730,22 +729,20 @@ function updateReducer<S, I, A>(
...
+        );
+        markSkippedUpdateLanes(updateLane);
       } else {
         // This update does have sufficient priority.
 
         if (newBaseQueueLast !== null) {
           const clone: Update<S, A> = {
             eventTime: updateEventTime,
-            expirationTime: Sync, // This update is going to be committed so we never want uncommit it.
+            lane: NoLane, // This update is going to be committed so we never want uncommit it.
```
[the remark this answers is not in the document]
acdlite: Behavior isn’t different, just had to change how I encoded it. It means that the update should never be skipped. I could make it a bitmask with all 1s but that’s technically not the right type.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -3451,19 +3237,15 @@ function beginWork(
   }
 
   // Before entering the begin phase, clear pending update priority.
-  // TODO: This assumes that we're about to evaluate the component and process
-  // the update queue. However, there's an exception: SimpleMemoComponent
-  // sometimes bails out later in the begin phase. This indicates that we should
-  // move this assignment out of the common path and into each branch.
-  workInProgress.expirationTime_opaque = NoWork;
+  workInProgress.lanes = NoLanes;
```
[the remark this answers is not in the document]
acdlite: Nah deleted accidentally, good catch

packages/react-reconciler/src/ReactFiberLane.js
acdlite: Another way to look at it is: we don't want to include Offscreen because that work is expensive; for example, it includes hydrating the remaining dehydrated boundaries. So since we know including Offscreen won't fix the error, anyway, might as well exclude it.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
acdlite: We call both of these function right before starting a new render at a new level. I don't think the order matters since `prepareFreshStack` doesn't touch `root.memoizedInteractions`, but I'll rearrange them for consistency with the other branches.

packages/react-reconciler/src/ReactFiberLane.js
lunaruan: That makes sense. What type of work would fall under `IdleLanes` then?

packages/react-reconciler/src/ReactFiberLane.js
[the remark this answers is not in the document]
lunaruan: Seb thinks that having an Offscreen component when something is not off screen is confusing. He is worried that people might misunderstand the name and assume that you only need to use `Offscreen` component if the elements are offscreen, which might result in people not wrapping something with Offscreen until the elements are hidden.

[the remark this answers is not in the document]
gaearon: `<KeepAlive>`, a la Vue?

packages/react-reconciler/src/__tests__/ReactIncrementalUpdates-test.js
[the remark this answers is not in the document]
bvaughn: The code I linked to above is `useSubscription` calling `setState` during the render phase, yes.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.js
```diff
@@ -1306,44 +1306,25 @@ describe('ReactDOMServerHooks', () => {
...
-                'Child One',
-                'Child Two',
-              ]);
-            }
-          });
+          expect(Scheduler).toFlushAndYieldThrough(
+            __DEV__
+              ? [
+                  'Child with ID',
+                  // Fallbacks are immdiately committed in TestUtils version
```
josecanhelp, line 1313: Spelling: immediately

packages/react-reconciler/src/ReactFiberLane.js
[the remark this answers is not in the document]
bvaughn: I think "shelf" might be a bit of a colloquialism that isn't very intuitive for a lot of people.

Isn't `Offscreen` always for things that might be **on or off** the screen? (e.g. list items that might be scrolled out of viewports, tab containers that might be hidden)

Is the concern that you'd always have an `Offscreen` wrapper around even the visible items in that case (since we don't support reparenting)?

Maybe something like `Conditional` would be more accurate/intuitive?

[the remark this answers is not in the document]
bvaughn: It does seem related to the concept of `keep-alive`...

packages/react-reconciler/src/ReactFiberLane.js
acdlite: Updates that don't have to finish with any particular urgency. Like maybe a background refresh of data. Honestly, we don't have a ton of use cases in practice yet. It might end up merging with the "long transition" lanes.

packages/react-reconciler/src/ReactFiberLane.js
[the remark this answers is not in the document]
acdlite: Agree that a public API called `Offscreen` is confusing. Similar to when we were trying to think of a name for the `Suspense` component.

The naming considerations are slightly different for this lane, though. The thing that makes `Offscreen` special, aside from it being really low priority, is that we're allowed to not finish it, because the work at that level is not visible to the user.

https://github.com/facebook/react/blob/17448e1e7f9131215dfddfff1aeddf3812073dd9/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L1654-L1659

A better solution, though, might be to move `resetChildLanes` (the function in that snippet) to the switch statement inside `completeWork`. Then the only thing remarkable about the Offscreen is that it has really low priority.

packages/react-reconciler/src/__tests__/ReactIncrementalUpdates-test.js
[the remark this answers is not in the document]
acdlite: Ah that's a hook update on itself. Those are fine. Should come up with some disambiguating term.

"Render phase" updates here refers to updates on another component, or a class component on itself. The ones that would trigger the `warnAboutRenderPhaseUpdatesInDEV` warning.

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,677 @@
...
+    // If we already suspended with a delay, then interrupting is fine. Don't
+    // bother waiting until the root is complete.
+    (wipLanes & suspendedLanes) === NoLanes
+  ) {
+    getHighestPriorityLanes(wipLanes);
+    const wipLanePriority = return_highestLanePriority;
+    if (nextLanePriority <= wipLanePriority) {
+      return wipLanes;
+    } else {
+      return_highestLanePriority = nextLanePriority;
```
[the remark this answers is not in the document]
acdlite: Yeah probably should for consistency

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,677 @@
...
+export function markRootMutableRead(root: FiberRoot, updateLane: Lane) {
+  root.mutableReadLanes |= updateLane & root.pendingLanes;
+}
+
+export function markRootFinished(root: FiberRoot, remainingLanes: Lanes) {
+  root.pendingLanes = remainingLanes;
+
+  // Let's try everything again
+  root.suspendedLanes = 0;
+  root.pingedLanes = 0;
```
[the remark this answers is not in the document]
acdlite: No reason, I'll change it

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,677 @@
...
+}
+
+export function higherPriorityLane(a: Lane, b: Lane) {
+  // This works because the bit ranges decrease in priority as you go left.
+  return a !== NoLane && a < b ? a : b;
+}
+
+// export function createLaneMap<T>(initial: T): LaneMap<T> {
+//   return new Array(TotalLanes).fill(initial);
+// }
```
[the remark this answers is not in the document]
acdlite: Yeah I'll delete for now. Originally was going to use this to store an expiration (like, the real kind) for each lane, but decided to defer to a follow up.

packages/react-reconciler/src/ReactFiberLane.js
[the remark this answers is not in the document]
NE-SmallTown: `<MaybeOrNotOffscreen />` 😆

packages/react-reconciler/src/__tests__/ReactIncrementalUpdates-test.js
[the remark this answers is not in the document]
bvaughn: Ah! Ok thanks for clarifying :smile:

packages/react-reconciler/src/ReactFiberLane.js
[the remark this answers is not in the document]
acdlite: I guess since it's only used by the inner function, it's fine to leave it.

Might get rid of the "register" thing. Too clever. If only we knew someone on TC39 who could advocate for stack allocated tuples!

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,682 @@
...
+  // it and we'll lose our progress. We should only do this if the new lanes are
+  // higher priority.
+  if (
+    wipLanes !== NoLanes &&
+    wipLanes !== nextLanes &&
+    // If we already suspended with a delay, then interrupting is fine. Don't
+    // bother waiting until the root is complete.
+    (wipLanes & suspendedLanes) === NoLanes
+  ) {
+    getHighestPriorityLanes(wipLanes);
```
NE-SmallTown, line 356: It's a 'get' function but we don't use the return value from it though that's intended, but is that will get confusing?

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,682 @@
...
+
+// Seems redundant, but it changes the type from a single lane (used for
+// updates) to a group of lanes (used for flushing work).
+export function laneToLanes(lane: Lane): Lanes {
+  return lane;
+}
+
+export function higherPriorityLane(a: Lane, b: Lane) {
+  // This works because the bit ranges decrease in priority as you go left.
+  return a !== NoLane && a < b ? a : b;
```
jddxf, line 555: Is it possible that `b` equals to `NoLane`?

acdlite: In practice no, it's always `SelectiveHydrationLane` but should probably rename/refactor this to something more specific so it's not used elsewhere.

acdlite: Good catch :)

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -7,16 +7,682 @@
...
+  return getHighestPriorityLanes(root.pendingLanes);
+}
+
+export function getLanesToRetrySynchronouslyOnError(root: FiberRoot): Lanes {
+  const everythingButOffscreen = root.pendingLanes & ~OffscreenLane;
+  if (everythingButOffscreen !== NoLanes) {
+    return everythingButOffscreen;
+  }
+  if (everythingButOffscreen & OffscreenLane) {
+    return OffscreenLane;
```
jddxf, line 380: Seems this branch is unreachable. Maybe `root.pendingLanes & OffscreenLane`?

Knove:
i need  a longer technical document :P~

awefeng:
Maybe I need an article about lanes to explain 5W1H，bigwig~~~~~

NotCoderJack:
is there an longer technical document about lanes ? Really need it! @acdlite

okmttdhr:
I wrote a short post; `What is "Lane" in React` https://dev.to/okmttdhr/what-is-lane-in-react-4np7
Please let me know if there are any corrections.

PR #18797, opened 2020-05-01
Don't "schedule" discrete work if we're scheduling sync work

sebmarkbage:
_This is an alternative fix to #18792. This is a convoluted bug. Settle in._

The problem is that we're adding discrete work during an unmount:

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberWorkLoop.old.js#L463-L480

Normally this is fine because once we complete the unmount we'll clear it here:

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberWorkLoop.old.js#L1879-L1887

That ensures that we never have any work scheduled on a root after it unmounts.

However in non-batched legacy mode the work that completes and clears it can be done synchronously here:

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberWorkLoop.old.js#L445

That's before we add it. So once it's done, we end up adding the discrete work after we've already unmounted. Leaving work remaining that later flushes and clears the container again.

So one bug is that we should be doing this before we run the sync work. Sync work or scheduling should always come at the end of a function to avoid bugs like this.

However, the other bug is that we're scheduling this for Sync work at all. We don't need to. So I moved it to only schedule discrete for non-Sync work.

That revealed another bug. We currently flush discrete work before controlled components which is before the batched updates wrapper in events. That also flushes sync work implicitly. However, it only does that if there's discrete work scheduled. We relied on this to actually make controlled components work. So when we no longer schedule discrete work, this no longer implicitly flushes.

So I make flushing discrete always flush sync too.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMFiberAsync-test.js
@@ -13,6 +13,7 @@ let React;
 
 let ReactDOM;
 let Scheduler;
+let act;
 
 const setUntrackedInputValue = Object.getOwnPropertyDescriptor(
   HTMLInputElement.prototype,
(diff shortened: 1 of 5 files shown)


trueadm:
I created a www sync with this PR: D21348348

PR #18799, opened 2020-05-01
Modern Event System: improve dispatching queue

trueadm:
One of the major pain points with how we deal with `SyntheticEvent` listener dispatching is how we allocate and store the three things we care about:

- the listener
- the instance
- the current target

We revised this a bit with the modern event system, but we still stuck each of these in an array on the event object (`_dispatchInstances`, `_dispatchListeners` and `_dispatchCurrentTargets`). Ideally, we shouldn't put any of these objects on the event object.

Another pain point is that plugin modules can either return `void | null | SyntheticEvent | Array<SyntheticEvent>`.

This PR changes all of that and simplifies only how the modern event system deals and handles this workflow. Instead of attaching what we care about to the event object (which means the awkward `unshift` logic for the capture phase) we pass around a `DispatchQueue` object instead. This queue contains an entry for each `SyntheticEvent` and its respective `capture` and `bubble` phases. We pass it to each plugin in order, and each plugin then populates the queue with the events, along with their details (listener, instance, current target). We then process through the queue once we've passed it to each plugin and dispatch all the events in correct order.

So we end up with clean types that look like this:

```js
type DispatchQueue = Array<DispatchQueueItem>;

type DispatchQueueItemPhaseEntry = {|
  instance: null | Fiber,
  listener: Function,
  currentTarget: EventTarget,
|};

type DispatchQueueItemPhase = Array<DispatchQueueItemPhaseEntry>;

type DispatchQueueItem = {|
  event: ReactSyntheticEvent,
  capture: DispatchQueueItemPhase,
  bubble: DispatchQueueItemPhase,
|};
```

We now have cleaner logic for how we extract events and their listeners for each of the event plugins, plus we also have cleaner logic for dealing with dispatching the events we've extracted, handling it all by phase. We also no longer need to leak the internals of how the dispatching or propagation works to the user via the `SyntheticEvent` object.

This will further brings wins for when we re-introduce the `createEventHandle` API, as it will vastly simplify all the bits around that.


packages/react-dom/src/events/DOMLegacyEventPluginSystem.js
```diff
@@ -191,9 +191,10 @@ function extractPluginEvents(
   eventSystemFlags: EventSystemFlags,
 ): Array<ReactSyntheticEvent> | ReactSyntheticEvent | null {
   let events = null;
-  for (let i = 0; i < plugins.length; i++) {
+  const legacyPlugins = ((plugins: any): Array<LegacyPluginModule<Event>>);
```
gaearon, line 194: Why the any?

trueadm: We're casting it here. It's not an `any`.

gaearon: Yeah but if you need the cast to go through `any`, it's not sound.

trueadm: Yeah, it's difficult to type here though as it depends on the feature flag depending on what the actual type will be. I don't know any other way of casting this to what we want, let me know if there is that you know of?

PR #18800, opened 2020-05-01
Update bundle configs to not build some packages

gaearon:
Removes old ReactCache from FB builds. (D21349384 checks in a frozen copy.)

Also removes all bundles that are currently unused on WWW (including new cache) so we stop getting automatic codemod diffs trying to remove them. Let's add them back when we need them.

### DIFF
--- a/scripts/rollup/bundles.js
+++ b/scripts/rollup/bundles.js
@@ -579,17 +579,12 @@ const bundles = [
 
   /******* React Cache (experimental, old) *******/
   {
-    bundleTypes: [
-      FB_WWW_DEV,
-      FB_WWW_PROD,
-      NODE_DEV,
-      NODE_PROD,
-      UMD_DEV,
-      UMD_PROD,
-    ],
+    // This is only used by our own tests.
+    // We can delete it later.
+    bundleTypes: [NODE_DEV, NODE_PROD],
     moduleType: ISOMORPHIC,
     entry: 'react-cache',
-    global: 'ReactCache',
+    global: 'ReactCacheOld',
     externals: ['react', 'scheduler'],
   },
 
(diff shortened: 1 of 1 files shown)


sebmarkbage:
It still build Flight classic right, which is unused?

gaearon:
>It still build Flight classic right, which is unused?

Can't parse your question :D I deleted the ones that our bot (D21346997) keeps trying to delete because they're unreachable.

scripts/rollup/bundles.js
```diff
@@ -318,7 +309,7 @@ const bundles = [
     ],
   },
   {
-    bundleTypes: [FB_WWW_DEV, FB_WWW_PROD],
+    bundleTypes: [], // TODO: FB_WWW_DEV, FB_WWW_PROD
```
sebmarkbage, line 312: Don't delete this please. `ReactFlightDOMRelayServerRuntime-dev.classic.js` is unused but `ReactFlightDOMRelayServerRuntime-dev.modern.js` is used.

gaearon: Ahhh. I get it now. Let me roll that back.

gaearon: fixed

scripts/rollup/bundles.js
```diff
@@ -197,7 +188,7 @@ const bundles = [
   /******* React DOM - www - Testing *******/
   {
     moduleType: RENDERER,
-    bundleTypes: [FB_WWW_DEV, FB_WWW_PROD, FB_WWW_PROFILING],
+    bundleTypes: [FB_WWW_DEV, FB_WWW_PROD],
```
sebmarkbage, line 191: This explains the missing CJS build which is good.

PR #18802, opened 2020-05-01
[DevTools] Improve HOC search UX

### LINKED ISSUE #16467: DevTools: Improve HOC search UX
Building on top of #360

We could probably improve the search UX in a couple of ways:

* Visually indicate when the "match" is because of a HOC name, e.g.
Screen Shot 2019-08-03 at 1 54 21 PM

* Match built-in HOC types (e.g. searching "me" should match `React.memo` HOCs, searching "f" should match `React.forwardRef`, etc.)

---
Originally reported via https://github.com/bvaughn/react-devtools-experimental/issues/365

bl00mber:
close #16467

Peek 2020-05-02


packages/react-devtools-shared/src/utils.js
```diff
@@ -260,9 +260,7 @@ export function separateDisplayNameAndHOCs(
 
   switch (type) {
     case ElementTypeClass:
-    case ElementTypeForwardRef:
```
bl00mber, line 263: This condition branch was invalid anyway (do not used in `<Badge />`, `printElement`), so it doesn't matter that match does not performed now in the cases of `ElementTypeForwardRef` and `ElementTypeMemo`.

packages/react-devtools-shared/src/devtools/utils.js
```diff
@@ -30,19 +25,6 @@ export function printElement(element: Element, includeWeight: boolean = false) {
...
-    } else {
-      hocDisplayNames.push('Memo');
-    }
-  } else if (element.type === ElementTypeForwardRef) {
-    if (hocDisplayNames === null) {
-      hocDisplayNames = ['ForwardRef'];
-    } else {
-      hocDisplayNames.push('ForwardRef');
-    }
-  }
```
bvaughn, line 45: Why this change?

bl00mber: This block is not needed anymore, because hisDisplayNames are already defined in the `packages/react-devtools-shared/src/utils.js` of this PR.

bvaughn: Sorry, my broader question encompassed the change in the utils file too. It looks like a change in behavior. Before, e.g. "Memo" would have been **just one of** the badges (`hocDisplayNames.push('Memo')`). With the change you made, it would be the **only** badge (`hocDisplayNames = ['Memo']`).

I'm asking why was that changed made. Seems like a regression?

bl00mber: I made this change because I thought it can cause something like `['Memo', 'Memo']` regarding second condition (push). I don't quite understand what this _particular_ condition branch means, I guess it tries to push to an empty array or create a new one, but when `hosDisplayNames` is initialized, it has only unconditional `null` value for these both types, so I do not see how (where) the code were previously able to reach the second condition (in which the name is pushed). And now, because `hosDisplayNames` is no longer a `null`, the first condition will also be (supposedly) always false, so it makes the first condition unreachable.

bvaughn: The logic before was this:
https://github.com/facebook/react/blob/62077431686a5c276a37f185fc5ebd1e128a7601/packages/react-devtools-shared/src/devtools/utils.js#L29-L45

First, it parses the `displayName` to see if there are any implied HOCs, then it checks the type. It's written as `if` + `if/else` intentionally (rather than `if/else if/else`) because it's possible that a `forwardRef` or `memo` component might also have a display name indicating additional HOC wrappers. It's not a *likely* case, but I thought it was worth handling.

bl00mber: HOC wrapper makes sense. I've updated the code to handle this case. If I place this logic to `react-devtools-shared/src/utils.js` only, does it seem correct, or it should be duplicated here? Maybe `unshift` fits better, not `push`? Because the actual name is the first array item, not the last one.

bvaughn: > If I place this logic to `react-devtools-shared/src/utils.js` only, does it seem correct, or it should be duplicated here?

It's fine if the logic is only in a single place. I'm more considered with not losing the behavior.

> Maybe `unshift` fits better, not `push`? Because the actual name is the first array item, not the last one.

Sure, I could see a case for using `unshift` instead.

bl00mber: > I'm more considered with not losing the behavior.

`hocDisplayNames` is not modified in any other parts of the package, so the behavior should be the same.

PR #18803, opened 2020-05-02
[Blocks] Add Shell to Fixture

gaearon:
This extends the fixture app a bit. Now it has two routes:

- Home (http://localhost:3000/)
- Profile (http://localhost:3000/profile)

On Home, we show Feed. A Feed consists of Posts with Comments.

On Profile, we show user name and Timeline. It also consists of Posts with Comments.

I started with this mostly to get a feel for working with Blocks and Server Components. And to have a playground for the Here and There stuff.

Note there is no actual server. I'm just trying to write it as if we had the transform.

I added a shell with a top-level tabbar that pushes to history and performs client-side navigation. Back button works.

Note that currently, there is only one cache. So this PR does not address the actual "Here and There" problem at all. That will be the next step.

Also note that **there is a bug**. I suspect it may be a bug in React. To repro, simply load `/` and click "Profile". It will start loading Profile but after it's ready, it will render Feed instead. Removing inner Suspense from Profile fixes it. So I think something is broken. I'm not sure if `yarn build` gives me old or new reconciler so not sure if that's related. But we should probably figure it out before merge.

<s>I could use some help with narrowing down the bug.</s> Extracted to https://github.com/facebook/react/pull/18809.


fixtures/blocks/delay.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+module.exports = (req, res, next) => {
+  if (req.query.delay) {
```
gaearon, line 9: This lets me add `?delay=XXXX` to simulate slow fetches for concrete API calls.

fixtures/blocks/src/server/ProfilePage.block.js
```diff
@@ -0,0 +1,37 @@
...
+function load(userId) {
+  return {
+    user: fetch(`/users/${userId}`).json(),
+    ProfileTimeline: loadProfileTimeline(userId),
+  };
+}
+
+// Client
+
+function ProfilePage(props, data) {
```
gaearon, line 26: I don’t think these should even be Blocks. Seems like Server Components would do?

sebmarkbage: This Block is loading its own data. Who knows if you're going to need to use client stuff later.

This part is a bit awkward because you could just return this from the load function. Making this render more or less an identity function. We probably should have the compiler identify that pattern and return the same function for all identity ones.

However, it's not necessarily better than client if this is a frequently repeated component.

I'm not sure how you pick between a Block and a server component since you kind of lock yourself in.

gaearon: >I'm not sure how you pick between a Block and a server component since you kind of lock yourself in.

Yeah. Another thing I just realized messes with my head is that Server Components are more like `load` functions than like `Render` functions. But their naming is like `Render` functions.

sebmarkbage: Load functions that return JSX directly, and is exported without a block wrapper, is almost like a Block except you can't pass the result to a JSX component.

fixtures/blocks/src/server/Post.block.js
```diff
@@ -0,0 +1,45 @@
...
+        borderRadius: 10,
+        marginBottom: 20,
+        padding: 20,
+        maxWidth: 500,
+      }}>
+      <h4 style={{marginTop: 0}}>{props.post.title}</h4>
+      <p>{props.post.body}</p>
+      <Suspense
+        fallback={<h5>Loading comments...</h5>}
+        unstable_avoidThisFallback={props.waitForComments}>
```
sebmarkbage, line 38: I was planning on turning this into its own component type. So this won't be possible. I also think it indicates an anti-pattern. Like you're trying to control the loading sequence imperatively.

gaearon: What is a good way to encode that I want them to appear with the post in Feed case (to avoid layout jumps), but I want them to load progressively for a Post Permalink case?

sebmarkbage: In that case, I guess you could conditionally render the two different component types and never switch between them.

Maybe add a wrapper to ensure that.

```js
function ResuspendBoundary({ children, fallback }) {
  return <Suspense fallback={fallback} unstable_avoidThisFallback={true}>{children}</Suspense>;
}
```

acutmore:
Tried to see what was causing the bug but didn't make much progress.
Could it have anything to do with `navigate` and `routeContext` never being stable, meaning the `<RouterProvider>` has a different context value on every render? Thinking that might be pushing a re-render of Feed onto the queue?

gaearon:
I think I have the bug here: https://github.com/facebook/react/pull/18809.

PR #18805, opened 2020-05-02
Scheduler does not wait for resolving dehydrated pending boundaries

ferlores:
## Summary

When hydrating if we [find](https://github.com/facebook/react/blob/914b57be27a8697d8ed83266466b30378af379f0/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L2482) a dehydrated node in [pending state](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMHostConfig.js#L695), we [register](https://github.com/facebook/react/blob/914b57be27a8697d8ed83266466b30378af379f0/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L2496) a retry [callback](https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOMHostConfig.js#L707) to resume hydration later. The scheduler trace does not wait for that callback to terminate the trace

## Test Plan

Run the test without wrapping `retryDehydratedSuspenseBoundary` to observe the problem

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -198,6 +198,7 @@ import {
   getWorkInProgressRoot,
   pushRenderLanes,
 } from './ReactFiberWorkLoop.new';
+import {unstable_wrap as Schedule_tracing_wrap} from 'scheduler/tracing';
 
 import {disableLogs, reenableLogs} from 'shared/ConsolePatchingDev';
 
@@ -2336,10 +2337,11 @@ function updateDehydratedSuspenseComponent(
     // Leave the child in place. I.e. the dehydrated fragment.
     workInProgress.child = current.child;
     // Register a callback to retry this boundary once the server has sent the result.
-    registerSuspenseInstanceRetry(
-      suspenseInstance,
-      retryDehydratedSuspenseBoundary.bind(null, current),
-    );
+    let retry = retryDehydratedSuspenseBoundary.bind(null, current);
+    if (enableSchedulerTracing) {
+      retry = Schedule_tracing_wrap(retry);
+    }
+    registerSuspenseInstanceRetry(suspenseInstance, retry);
     return null;
   } else {
     // This is the first attempt.
--- a/packages/react/src/__tests__/ReactDOMTracing-test.internal.js
+++ b/packages/react/src/__tests__/ReactDOMTracing-test.internal.js
@@ -720,6 +720,76 @@ describe('ReactDOMTracing', () => {
       });
 
       // @gate experimental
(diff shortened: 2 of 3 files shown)


packages/react/src/__tests__/ReactDOMTracing-test.internal.js
```diff
@@ -23,6 +23,11 @@ let onWorkScheduled;
 let onWorkStarted;
 let onWorkStopped;
 
+import {
+  SUSPENSE_START_DATA,
```
gaearon, line 27: Let's not do this in the test. Just inline them.

sebmarkbage:
I guess we currently traces interactions through the hydration even when deprioritizing. So this makes sense. https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberBeginWork.old.js#L2222

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2338,7 +2339,9 @@ function updateDehydratedSuspenseComponent(
     // Register a callback to retry this boundary once the server has sent the result.
     registerSuspenseInstanceRetry(
       suspenseInstance,
-      retryDehydratedSuspenseBoundary.bind(null, current),
+      Schedule_tracing_wrap(
```
sebmarkbage, line 2342: Actually, let's only do this if the `enableSchedulerTracing` is on. Like so:

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberCommitWork.old.js#L1822

We don't pull in a dependency on `scheduler/tracing` in the production build. Only in the profiling build.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2336,10 +2337,11 @@ function updateDehydratedSuspenseComponent(
...
     workInProgress.child = current.child;
     // Register a callback to retry this boundary once the server has sent the result.
-    registerSuspenseInstanceRetry(
-      suspenseInstance,
-      retryDehydratedSuspenseBoundary.bind(null, current),
-    );
+    let retry = retryDehydratedSuspenseBoundary.bind(null, current);
+    if (enableSchedulerTracing) {
+      retry = Schedule_tracing_wrap(retry);
+    }
```
ferlores, line 2343: I noticed that if I comment these lines the test still passes, but not when I comment the lines in  `ReactFiberBeginWork.old.js`. Is there a way to annotate the test so it also runs pulling this file?

sebmarkbage: CI runs .new.js

Issue #18798, opened 2020-05-01
Error: "Cannot read property 'duration' of undefined"


shiftyp:
Describe what you were doing when the bug occurred:
1. Add interaction tracing with unstable_trace
2. Record a profile, navigate to Profiler > Profiled Interactions
3. Error appears when scrolling view or immediately

![Kapture 2020-05-01 at 8 37 44](https://user-images.githubusercontent.com/131928/80805899-3688c400-8b87-11ea-88ca-3a7ae7de3589.gif)
---------------------------------------------
Please do not remove the text below this line
---------------------------------------------

DevTools version: 4.6.0-6cceaeb67

Call stack: at chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:40:344360
    at Array.map (<anonymous>)
    at chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:40:344166
    at ci (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:59620)
    at Ll (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:109960)
    at qc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:102381)
    at Hc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:102306)
    at Vc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:102171)
    at Tc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:98781)
    at chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:46014

Component stack: in Unknown
    in Unknown
    in div
    in div
    in n
    in div
    in bc
    in div
    in n
    in div
    in vc
    in div
    in div
    in div
    in So
    in Unknown
    in n
    in Unknown
    in div
    in div
    in rl
    in Ze
    in fn
    in Ga
    in _s

chiragswadia:
Looks like duplicate of https://github.com/facebook/react/issues/18786 🤔

shiftyp:
Here is a deployed repro case: https://react-example1.s3.amazonaws.com/index.html

Steps:

1. Load above URL
2. Open React Profiler
3. Start recording
1. Hit the "Fetch" button a few times
4. Stop Recording
5. navigate to the Profiled Interactions tab
[...]

Issue #18786, opened 2020-04-30
Error: "Cannot read property 'duration' of undefined"


sebmarkbage:
Describe what you were doing when the bug occurred:

Profiled the new FB. Scrolled down to tail loads.

---------------------------------------------
Please do not remove the text below this line
---------------------------------------------

DevTools version: 4.6.0-a2fb84beb

Call stack: at chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:40:345591
    at Array.map (<anonymous>)
    at chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:40:345397
    at Ai (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:32:62580)
    at zl (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:32:112694)
    at jc (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:32:104789)
    at Oc (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:32:104717)
    at Tc (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:32:104585)
    at gc (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:32:101042)
    at chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:32:47376

Component stack: at chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:40:344679
    at div
    at div
    at n (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:40:194307)
    at div
    at Cc (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:40:346311)
    at div
    at n (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:3:8163)
    at div
    at bc
    at div
    at div
    at div
    at Do (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:40:262081)
    at chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:40:364048
    at n (chrome-extension://dnjnjgbfilfphmojnmhliehogmojhclc/build/main.js:40:274563)
[...]

PR #18806, opened 2020-05-03
fix: Error: Cannot read property 'duration' of undefined

Amine-H:
## Summary

Fix for #18786 and #18798

Just a few things I should mention first, I am really new to the code base, ~not sure what was the intention to use the indexes from `commits` instead of `commitData`.
In other words if `commitData` has more commit than `commits` (interaction commits) then this PR would be showing more commits than whan It should.~

#### UPDATE:
I have updated my changes to "ignore" commits that have no corresponding `commitData`

## Test Plan

- Profile https://react-example1.s3.amazonaws.com/index.html
- Scroll down to the end of the list

Thanks @shiftyp for the clear reproduction steps

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/InteractionListItem.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/InteractionListItem.js
@@ -35,7 +35,7 @@ function InteractionListItem({data: itemData, index, style}: Props) {
     selectTab,
   } = itemData;
 
-  const {commitData } = dataForRoot;
+  const {commitData} = dataForRoot;
   const {interactions, lastInteractionTime, maxCommitDuration} = chartData;
 
   const interaction = interactions[index];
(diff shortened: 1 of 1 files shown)


Amine-H:
After further investigating the issue, I found out that the changes I pushed aren't a valid.

the issue is caused when you stop profiling while the page still doing some work. causing missing commit in `commitData` or `interactionCommits` as an unexisting `commit`

bvaughn:
Thanks for submitting this PR!

I'm a little concerned that while it fixes a symptom (out of range index) it doesn't fix the underlying cause (why is the data missing?)

Adding some logging to DevTools and trying the repro, I can see that e.g. the backend records 181 commits but then the later profiling data only contains 180 commits. So one of them is being filtered out for some reason or is not being sent across the bridge. Either way this is likely to cause bigger problems than just this specific fix, so I'm going to look into it further.

bvaughn:
Going to close this PR in favor of #18862!

Amine-H:
Alright ! that should do it! thanks :+1:

bvaughn:
No problem! Thank you!

PR #18808, opened 2020-05-03
Rename Flight to Transport

sebmarkbage:
Flight is still the codename for the implementation details (like Fiber).

However, now the public package is react-transport-... which is only intended to be used directly by integrators into meta-frameworks.


scripts/rollup/bundles.js
```diff
@@ -313,7 +313,7 @@ const bundles = [
   {
     bundleTypes: [FB_WWW_DEV, FB_WWW_PROD],
     moduleType: RENDERER,
-    entry: 'react-flight-dom-relay',
+    entry: 'react-transport-dom-relay',
     global: 'ReactFlightDOMRelayClient',
```
sebmarkbage, line 317: I kept this name so we don't break the sync but we should follow up to rename this internally.

PR #18809, opened 2020-05-03
[Blocks] Failing test for nested load

gaearon:
Extracted from https://github.com/facebook/react/pull/18803. I think this is a bug.

The symptom is that we keep seeing the old screen after it unsuspends.
Hosting `<Suspense>` from `Parent` into `App` "fixes" it.

### DIFF
--- a/packages/react-reconciler/src/__tests__/ReactBlocks-test.js
+++ b/packages/react-reconciler/src/__tests__/ReactBlocks-test.js
@@ -258,8 +258,9 @@ describe('ReactBlocks', () => {
     );
   });
 
+  // Regression test.
   // @gate experimental
-  it('can load a nested block after an update', async () => {
+  it('does not render stale data after ping', async () => {
     function Child() {
       return <span>Name: {readString('Sebastian')}</span>;
     }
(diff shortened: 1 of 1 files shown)


gaearon:
Hmm. Not sure yet. The reason results are different is because if we hoist `<Suspense>` above the block, it will trigger `reconcileChildFibers` on ping, which will trigger the `load` initializer with `'Sebastian'`. Otherwise we don't trigger `load` and get use old `load` result for rendering.

gaearon:
Child doesn't have to be a Block.

This still fails:

```js
    function Child() {
      return <span>Name: {readString("Sebastian")}</span>;
    }

    const loadParent = block(
      function Parent(props, data) {
        return (
          <Suspense fallback="Loading...">
            {data.name ? <Child /> : <span>Empty</span>}
          </Suspense>
        );
      },
      function load(name) {
        return { name };
      },
    );
```

I couldn't find a version that fails with completely removing Blocks though.

sebmarkbage:
That sounds like it might be a Blocks specific bug.

sebmarkbage:
I could see how this might happen in legacy mode if the lazy wrapper mounts as a "LazyComponent" but not in CM.

gaearon:
I've noticed `current.type !== workInProgress.type`, does that tell us anything?

gaearon:
If I add some more logs in beginWork for blocks:

```js
        console.log('render block')
        console.log('current', current.type._data)
        console.log('wip', workInProgress.type._data)
```

I see this:

```
    render block
    current { name: null }
    wip { name: 'Sebastian' }

    ---- ping ---

    render block
    current { name: 'Sebastian' }
    wip { name: null }
```

gaearon:
Huh. Notice the failure is DEV only.

gaearon:
If you remove this line, it passes:

https://github.com/facebook/react/blob/64d4b8420495ed6685b9a48ce0295fa63c592bb9/packages/react-reconciler/src/ReactFiberBeginWork.old.js#L3052

gaearon:
OK I think I understand what's going on.

`beginWork(Block)` assumes that `wip.type` is correct. Which is always correct if we reconcile its parent. And generally, if we actually got to *rendering* a Block and didn't jump over it, it *does* mean its parent was reconciled. A Block doesn't render on its own. (TODO: verify this is actually always the case.)

However, this condition, which was originally added for Fast Refresh, prevents bailout if `current.type !== workInProgress.type`. Which we happen to trigger because we updated the `type` field in the child reconciler last time. So now we actually "visit" our Block and try to render it using `workInProgress.type`. But that field has never been updated, because we rely on parent's reconciliation to do so. So we render it with a stale block `type`.

gaearon:
> A Block doesn't render on its own. (TODO: verify this is actually always the case.)

Hmm. This is actually not true. A Block *can* render on its own as a result of setState. So the problem is more general, and not tied to the Fast Refresh check.

It's just that we make an assumption that `wip.type` is up to date. But if we didn't reconcile the parent, it can be stale.

gaearon:
https://github.com/facebook/react/pull/18810

PR #18810, opened 2020-05-04
[Blocks] Fix stale data on updates

gaearon:
Includes tests from https://github.com/facebook/react/pull/18809.
The last commit is the fix.

Blocks store data on type, so we need to copy it over when cloning WIP. That's the fix.
Verified https://github.com/facebook/react/pull/18803 works with it too.

### DIFF
--- a/packages/react-reconciler/src/ReactFiber.old.js
+++ b/packages/react-reconciler/src/ReactFiber.old.js
@@ -277,6 +277,8 @@ export function createWorkInProgress(current: Fiber, pendingProps: any): Fiber {
     current.alternate = workInProgress;
   } else {
     workInProgress.pendingProps = pendingProps;
+    // Needed because Blocks store data on type.
+    workInProgress.type = current.type;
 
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiber.old.js
```diff
@@ -413,6 +415,8 @@ export function resetWorkInProgress(
     workInProgress.memoizedProps = current.memoizedProps;
     workInProgress.memoizedState = current.memoizedState;
     workInProgress.updateQueue = current.updateQueue;
+    // Needed because Blocks store data on type.
+    workInProgress.type = current.type;
```
gaearon, line 419: I don't have coverage for the SuspenseList codepath but I figured it would need to be updated too.

sebmarkbage:
In particular I think we still need elementType on the fiber for quick reconciliation. So maybe we should be reading the data off element type instead. Which could be a lazy component. That would require rereading it from the lazy.

PR #18814, opened 2020-05-04
[Memory Leaks] Detach Fibers from nodes with refs

gaearon:
A more restrained alternative to https://github.com/facebook/react/pull/18690.
Addresses @sebmarkbage's proposal in https://github.com/facebook/react/issues/16087:

> If a handle on a DOM node is leaked, it takes the React tree with it. This is a fairly easy mistake to make and the effect is pretty high. What we would do here is special case DOM nodes with refs on them, and always detach their back pointer to the React Fiber, if it was ever fully mounted. We currently traverse these trees anyway when they get deleted. We want to stop doing this for most things but for nodes with a ref it seems minor to special case since they typically need to be invoked with null anyway.

Note: this does not fix any memory leak **in React**; it just makes the impact of **userland** memory leaks less severe.

**[View without whitespace](https://github.com/facebook/react/compare/master...gaearon:detach-ref?w=1)**

I checked that this works manually:

Screenshot 2020-05-04 at 15 37 12

### DIFF
--- a/packages/react-native-renderer/src/ReactFabricHostConfig.js
+++ b/packages/react-native-renderer/src/ReactFabricHostConfig.js
@@ -342,6 +342,10 @@ export function shouldSetTextContent(type: string, props: Props): boolean {
   return false;
 }
 
+export function clearInstanceHandle(instance: Instance) {
+  instance.canonical._internalInstanceHandle = null;
+}
+
 // The Fabric renderer is secondary to the existing React Native renderer.
 export const isPrimaryRenderer = false;
 
--- a/packages/react-native-renderer/src/ReactNativeHostConfig.js
+++ b/packages/react-native-renderer/src/ReactNativeHostConfig.js
@@ -488,6 +488,10 @@ export function clearContainer(container: Container): void {
   // UIManager does not expose a "remove all" type method.
 }
 
+export function clearInstanceHandle(instance: Instance) {
+  uncacheFiberNode(instance);
+}
+
 export function unhideTextInstance(
   textInstance: TextInstance,
   text: string,
--- a/packages/react-art/src/ReactARTHostConfig.js
+++ b/packages/react-art/src/ReactARTHostConfig.js
@@ -322,6 +322,10 @@ export function resetTextContent(domElement) {
   // Noop
 }
 
+export function clearInstanceHandle(instance) {
+  // Noop
+}
+
 export function shouldDeprioritizeSubtree(type, props) {
   return false;
 }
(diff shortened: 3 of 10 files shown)


packages/react-native-renderer/src/ReactFabricHostConfig.js
```diff
@@ -342,6 +342,10 @@ export function shouldSetTextContent(type: string, props: Props): boolean {
   return false;
 }
 
+export function clearInstanceHandle(instance: Instance) {
+  instance.canonical._internalInstanceHandle = null;
```
gaearon, line 346: @sebmarkbage Is this okay?

packages/react-native-renderer/src/ReactNativeHostConfig.js
```diff
@@ -488,6 +488,10 @@ export function clearContainer(container: Container): void {
   // UIManager does not expose a "remove all" type method.
 }
 
+export function clearInstanceHandle(instance: Instance) {
+  uncacheFiberNode(instance);
```
gaearon, line 492: Curiously I've noticed Paper already does this for all deletions already. I can remove this?

trueadm: Where is that done?

trueadm, line 492: You'll need to import `uncacheFiberNode` though, right?

trueadm:
Thinking about it, we can also use this heurstic to clear the listeners of nodes that have listeners via `createEventHandle`. That's because we only attach those that have `ref`s, awesome! We can possibly do the same for scopes too.

packages/react-native-renderer/src/ReactNativeHostConfig.js
gaearon: https://github.com/facebook/react/blob/3e13d70984a0ff091e71c20e1b29209bd83342d4/packages/react-native-renderer/src/ReactNativeHostConfig.js#L77-L86

gaearon: No, it's already defined in this file. It turns out we already do this so it's unnecessary.

trueadm: Ah, makes sense to your other comment.

trueadm: Yeah, then we won't need the above line. We'll just need it for Fabric.

aweary:
> Let's hold on this for a bit until we figure out the hidden semantics

@sebmarkbage how are those hidden semantics coming along? 🙂

gaearon:
We are starting to experiment with what gets broken by double-invoking effects (https://github.com/facebook/react/pull/19523) which will inform the further work on Hidden. Early stage still.

aweary:
@gaearon any updates on the double-invoking effects experiment regarding this? we're still running on a fork that does some more aggressive `null`ing to avoid the hard-to-find memory leaks 👀

gaearon:
We're trying to roll it out atm.

PR #18815, opened 2020-05-04
chore: remove usage of deprecated Jest functions

SimenB:
## Summary

Jest 26 is coming today or tomorrow, and we've removed `require.requireActual`. I just ran the lint rule `jest/no-deprecated-functions` on the code base, which also fixes a couple of removals scheduled for Jest 27. Doesn't hurt to be prepared 🙂 

## Test Plan

Green CI


SimenB:
Hmm, weird failure... `jest.requireActual` has been around since Jest 17 or some such

SimenB:
@gaearon this is one of the blockers for Jest 26. Thoughts on how to move forward?

One solution is to not touch the parts that use an old version of Jest, which is probably the easiest

SimenB:
#19920

PR #18816, opened 2020-05-04
chore: upgrade to Jest 26

SimenB:
## Summary

As mentioned in #18815, it's incoming. I mostly wanna test the new (opt-in) fake timers, but I thought to open a PR with just the bump as well. Includes #18815, so you can choose to just merge this one if you want.

Draft for now as 26 is still alpha.

## Test Plan

Green CI


gaearon:
@SimenB We're running into a need for the latest jsdom, which would require us to get on Jest 26. Any chance you want to revive this PR? I'd be happy to assist with missing bits.

rickhanlonii:
Could we patch Jest 25 with a new jsdom patch version or does it include more changes then just this?

gaearon:
We're gonna go with a custom jsdom fork for the time being but this is a liability.

eps1lon:
Can you override the default environment `jest` is using in this repo and use [`jest-environment-jsdom-sixteen`](https://www.npmjs.com/package/jest-environment-jsdom-sixteen) instead? That way you can split up the work required for jest 26 into two parts: fixing broken test due to jsdom 16 and fixing broken tests due to jest 26.

SimenB:
`jest-environment-jsdom-sixteen` is for jest 25, so you can use that 👍

rickhanlonii:
PR to use jest-environment-jsdom-sixteen [here](https://github.com/facebook/react/pull/19288), we can land that after JSDOM publishes the fix.

SimenB:
#19920

Issue #18819, opened 2020-05-04
Bug: eslint-plugin-react-hooks cannot use optional chaining in deps


coolkev:
React version: 16.3.1
eslint-plugin-react-hooks version: 4.0.0

## Steps To Reproduce
Use ?. in dependency to any of the react hooks.
```
export default function App() {
    const x = Date.now() % 2 === 0 ? { test: true } : undefined;
  
    React.useEffect(() => {
      if (x?.test) {
        console.log('test');
      }
    }, [x?.test]);
  
    return (
      <div className="App">
        <h1>Hello CodeSandbox</h1>
      </div>
    );
  }
  
```

Link to code example:
I was unable to reproduce the behavior in codesandbox, it must be transforming the code or stripping out the ? before linting. It happens in VS Code.

## The current behavior
eslint errors occur:
React Hook React.useEffect has a missing dependency: 'x'. Either include it or remove the dependency array.eslintreact-hooks/exhaustive-deps
React Hook React.useEffect has a complex expression in the dependency array. Extract it to a separate variable so it can be statically checked.eslintreact-hooks/exhaustive-deps
![image](https://user-images.githubusercontent.com/319334/81008656-25021f00-8e08-11ea-8f91-5c533671d75a.png)

## The expected behavior
I would expect it to work just like it does currently for non-optional child member access.
![image](https://user-images.githubusercontent.com/319334/81008703-36e3c200-8e08-11ea-8aa9-56df8ccf9531.png)
When ? is removed, it no longer shows eslint errors, but obviously would fail at runtime if x is undefined.

blydiet:
Hey one, Idea is inside of x you can make it look more like this   
const x = Date.now() % 2 === 0 ? { test: true } : { test: undefined };
that should get the result you want I think.

Also,  if you're on  Firefox and you did not set privacy.reduceTimerPrecision  you will not get the right reduced time precision for the date.now() the numbers will be a little random 

check out the docs https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now

coolkev:
The example code was just to show a simple way to reproduce the problem of the hook reporting incorrectly. The data used in my dependency checks is actually from redux store and will contain undefined objects. I use the useEffect hook to handle fetching the data if it is not present, and it can happen at multiple levels.

For example: 
```
const OrderDetail = ({ orderId }) => {

    const order = useSelector(state => state.orders[orderId]);

[...]

PR #18820, opened 2020-05-04
eslint-plugin-react-hooks: allow OptionalMemberExpression in deps (#18819)

coolkev:
## Summary

Fixes issue #18819 

## Test Plan

Added test case, and verified the fix works in my own VS Code instance.


gaearon:
Thanks!

AviVahl:
@gaearon I believe this PR caused a regression.

consider:
```js
const someContext = {
    users: Math.random() > 0.5 ? [{name: 'test'}] : undefined,
};

const filteredUsers = useMemo(() => {
    return someContext.users?.filter((u) => u.name.startsWith('a'));
}, [someContext.users]);
```
New version of hook now tells us the dep should be `someContext.users?.filter`, which is wrong for many common use cases (e.g. different class instances sharing the same method from prototype).

PR #18825, opened 2020-05-05
Add unstable prefix to experimental APIs

acdlite:
We've been shipping unprefixed experimental APIs (like `createRoot` and `useTransition`) to the Experimental release channel, with the rationale that because these APIs do not appear in any stable release, we're free to change or remove them later without breaking any downstream projects.

What we didn't consider is that downstream projects might be tempted to use feature detection:

```js
const useTransition = React.useTransition || fallbackUseTransition;
```

This pattern assumes that the version of `useTransition` that exists in the Experimental channel today has the same API contract as the final `useTransition` API that we'll eventually ship to stable.

To discourage feature detection, I've added an `unstable_` prefix to all of our unstable APIs.

The Facebook builds still have the unprefixed APIs, though. We will continue to support those; if we make any breaking changes, we'll migrate the internal callers like we usually do. To make testing easier, I added the `unstable_`-prefixed APIs to the www builds, too. That way our tests can always use the prefixed ones without gating on the release channel.


leonardodino:
created a documentation update issue: 
https://github.com/reactjs/reactjs.org/issues/2960

Please close the issue if the team think it's best to avoid this change 😄

PR #18830, opened 2020-05-05
Modern Event System: fix EnterLeave plugin logic

trueadm:
Today we found an internal issue with the modern event system and the EnterLeave plugin. Specifically, we didn't track `mouseout` on the container, only in the portal. That's because the portal had a child contained `onMouseEnter`. This bug is actually very similar to that of the recent fix to https://github.com/facebook/react/pull/18720.

Making this change, so that we now register this plugin on each root container fixed that issue, but also brought about another issue. This was great though, as we had already solved this in the legacy EnterLeave plugin (I commented on it below). Taking that logic and bringing it to the modern event system fixed the failing issues and aligned the modern event system with the legacy one.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -9,6 +9,7 @@
 
 import {registrationNameModules} from 'legacy-events/EventPluginRegistry';
 import {canUseDOM} from 'shared/ExecutionEnvironment';
+import invariant from 'shared/invariant';
 import {
   setListenToResponderEventTypes,
   addResponderEventSystemEvent,
@@ -282,12 +283,13 @@ export function ensureListeningTo(
     // Containers should only ever be element nodes. We do not
     // want to register events to document fragments or documents
     // with the modern plugin event system.
-    if (
+    invariant(
       rootContainerElement != null &&
-      rootContainerElement.nodeType === ELEMENT_NODE
-    ) {
-      listenToEvent(registrationName, ((rootContainerElement: any): Element));
-    }
+        rootContainerElement.nodeType === ELEMENT_NODE,
+      'ensureListeningTo(): received a container that was not an element node. ' +
+        'This is likely a bug in React.',
+    );
+    listenToEvent(registrationName, ((rootContainerElement: any): Element));
   } else {
     // Legacy plugin event system path
     const isDocumentOrFragment =
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -9,7 +9,6 @@
 
 import {registrationNameModules} from 'legacy-events/EventPluginRegistry';
 import {canUseDOM} from 'shared/ExecutionEnvironment';
-import invariant from 'shared/invariant';
 import {
   setListenToResponderEventTypes,
   addResponderEventSystemEvent,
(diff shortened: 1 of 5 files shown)


packages/react-dom/src/client/ReactDOMRoot.js
```diff
@@ -129,6 +132,11 @@ function createRootImpl(
         ? container
         : container.ownerDocument;
     eagerlyTrapReplayableEvents(container, doc);
+  } else if (enableModernEventSystem) {
+    // We use ensureListeningTo because the container might
+    // be a COMMENT_NODE and this function handles events
+    // on comments for us.
+    ensureListeningTo(container, 'onMouseEnter');
```
trueadm, line 139: This is similar to what we did with the fix in #18720.

packages/react-dom/src/events/plugins/ModernEnterLeaveEventPlugin.js
```diff
@@ -165,6 +162,14 @@ const EnterLeaveEventPlugin = {
     enter.target = toNode;
     enter.relatedTarget = fromNode;
 
+    // If we are not processing the first ancestor, then we
+    // should not process the same nativeEvent again, as we
+    // will have already processed it in the first ancestor.
+    const nativeTargetInst = getClosestInstanceFromNode(nativeEventTarget);
```
trueadm, line 168: This is basically the same as what we do on:

https://github.com/facebook/react/blob/master/packages/react-dom/src/events/plugins/LegacyEnterLeaveEventPlugin.js#L168-L173

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -280,16 +279,15 @@ function ensureListeningTo(
...
+    if (
       rootContainerElement != null &&
-        rootContainerElement.nodeType === ELEMENT_NODE,
-      'ensureListeningTo(): received a container that was not an element node. ' +
-        'This is likely a bug in React.',
-    );
-    listenToEvent(registrationName, ((rootContainerElement: any): Element));
+      rootContainerElement.nodeType === ELEMENT_NODE
+    ) {
+      listenToEvent(registrationName, ((rootContainerElement: any): Element));
```
trueadm, line 289: We don't need an invariant here, we can use this more as a utility function. This let's us re-use this function and it's logic

packages/react-dom/src/client/ReactDOMRoot.js
```diff
@@ -129,6 +132,10 @@ function createRootImpl(
         ? container
         : container.ownerDocument;
     eagerlyTrapReplayableEvents(container, doc);
+  } else if (enableModernEventSystem) {
+    // We use ensureListeningTo because the container might
+    // be a COMMENT_NODE.
```
gaearon, line 137: What does this have to do with comment nodes? I'm not following. The test case doesn't use them so I'm not seeing the connection.

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -280,16 +279,15 @@ function ensureListeningTo(
       rootContainerInstance.nodeType === COMMENT_NODE
         ? rootContainerInstance.parentNode
         : rootContainerInstance;
-    // Containers can only ever be element nodes. We do not
+    // Containers should only ever be element nodes. We do not
     // want to register events to document fragments or documents
     // with the modern plugin event system.
-    invariant(
```
gaearon, line 286: Why relax the invariant?

trueadm: Yeah, I'll bring it back in.

packages/react-dom/src/client/ReactDOMRoot.js
trueadm: I've update the condition above. The comment is redundant now.

PR #18832, opened 2020-05-05
[Blocks Fixture] Add a tiny router

gaearon:
Just a small abstraction for route matching to get a feel for how it composes. The function returns parsed out "params", which include remaining `route`. The nested matches would use that instead of the root `route`.

### DIFF
--- a/fixtures/blocks/src/server/ProfilePage.block.js
+++ b/fixtures/blocks/src/server/ProfilePage.block.js
@@ -12,20 +12,16 @@ import {unstable_block as block, Suspense} from 'react';
 // Server
 
 import {fetch} from 'react-data/fetch';
+import {matchRoute} from './ServerRouter';
 import loadProfileBio from './ProfileBio.block';
 import loadProfileTimeline from './ProfileTimeline.block';
 
-function load(userId, tab) {
+function load({userId, route}) {
   const user = fetch(`/users/${userId}`).json();
-  let Tab;
-  switch (tab) {
-    case 'bio':
-      Tab = loadProfileBio(user);
-      break;
-    default:
-      Tab = loadProfileTimeline(userId);
-      break;
-  }
+  const Tab = matchRoute(route, [
+    ['/', loadProfileTimeline, userId],
+    ['/bio', loadProfileBio, user],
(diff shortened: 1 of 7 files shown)


fixtures/blocks/src/server/ProfilePage.block.js
```diff
@@ -12,20 +12,16 @@ import {unstable_block as block, Suspense} from 'react';
 // Server
 
 import {fetch} from 'react-data/fetch';
+import {matchRoute} from './ServerRouter';
 import loadProfileBio from './ProfileBio.block';
 import loadProfileTimeline from './ProfileTimeline.block';
 
-function load(userId, tab) {
+function load({userId, route}) {
```
gaearon, line 19: This `route` is the remainder.

fixtures/blocks/src/server/ProfilePage.block.js
```diff
@@ -12,20 +12,16 @@ import {unstable_block as block, Suspense} from 'react';
...
-  switch (tab) {
-    case 'bio':
-      Tab = loadProfileBio(user);
-      break;
-    default:
-      Tab = loadProfileTimeline(userId);
-      break;
-  }
+  const Tab = matchRoute(route, [
+    ['/', loadProfileTimeline, userId],
```
gaearon, line 22: Varargs after `load` will be passed after the first implicit `params` object.

fixtures/blocks/src/server/ServerRouter.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+/* eslint-disable import/first */
+
+function tryMatch(template, url) {
```
gaearon, line 9: Doesn't do anything fancy but extracts stuff like `:id` and lets me continue the match with `/profile/:id/*` where `*` matches anything and puts the remainder into params.

gaearon:
I just merged https://github.com/facebook/react/issues/18837 with both.

PR #18838, opened 2020-05-06
Warn if calling setState outside of render but before commit

sebmarkbage:
This indicates that there's a side-effect spawned during render, that then asynchronously tries to set state when the response comes back. E.g. kick off a request and then setState with the result. It's a common mistake when implementing Suspense.

This should be done in useEffect.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -3112,7 +3112,9 @@ function beginWork(
                 // been unsuspended it has committed as a resolved Suspense component.
                 // If it needs to be retried, it should have work scheduled on it.
                 workInProgress.effectTag |= DidCapture;
-                break;
+                // We should never render the children of a dehydrated boundary until we
+                // upgrade it. We return null instead of bailoutOnAlreadyFinishedWork.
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -3110,7 +3110,8 @@ function beginWork(
                 // been unsuspended it has committed as a resolved Suspense component.
                 // If it needs to be retried, it should have work scheduled on it.
                 workInProgress.effectTag |= DidCapture;
-                break;
+
+                return null;
```
sebmarkbage, line 3114: This revealed that if you do this, then it'll schedule work on the "children" of the Suspense boundary. Before it mounts. We then try to render into the children. The Suspense boundary hasn't changed, so we get here to bail out. The normal bail out tries to render children. The child is a DehydratedFragment that isn't a valid tag to render. (It has no beginWork). That causes an invariant.

We never expect to render these children so I just bail out by returning null instead.

bvaughn: Some form of that would have made a nice inline comment explaining the significance of the return 😅

PR #18845, opened 2020-05-06
[Release] Follow redirect when downloading tarball

acdlite:
Adds -L option to `curl` command.

See: https://curl.haxx.se/docs/manpage.html#-L

### DIFF
--- a/scripts/release/shared-commands/download-build-artifacts.js
+++ b/scripts/release/shared-commands/download-build-artifacts.js
@@ -30,7 +30,9 @@ const run = async ({build, cwd}) => {
 
   // Download and extract artifact
   await exec(`rm -rf ./build/node_modules*`, {cwd});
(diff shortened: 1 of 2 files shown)


scripts/release/shared-commands/download-build-artifacts.js
```diff
@@ -30,7 +30,9 @@ const run = async ({build, cwd}) => {
 
   // Download and extract artifact
   await exec(`rm -rf ./build/node_modules*`, {cwd});
-  await exec(`curl ${nodeModulesURL} --output ./build/node_modules.tgz`, {cwd});
+  await exec(`curl -L ${nodeModulesURL} --output ./build/node_modules.tgz`, {
```
bvaughn, line 33: Nice.

Should we also add here while we're at it?
https://github.com/facebook/react/blob/33c3af284cbbb228199f1e2bd369c6e8efd9763d/scripts/release/prepare-release-from-npm-commands/check-out-packages.js#L47-L48

PR #18849, opened 2020-05-06
Modern Event System: fix bug in EnterLeave

trueadm:
This fixes a bug I accidently introduced in https://github.com/facebook/react/pull/18830.

Notably, I hoisted out the `related` const, but only today noticed they were two different related constants. One was `nativeEvent.relatedTarget || nativeEvent.fromElement;` and the other was `nativeEvent.relatedTarget || nativeEvent.toElement;` and for some reason I read them as the same thing! :/

So I've reverted that change, fixing the issue we're seeing internally. I'll follow up with a regression test at some point.


gaearon:
Oops, I totally missed that as well. Good catch.

PR #18855, opened 2020-05-07
rough implementation of direct context updates

gnoff:
This is an early implementation of an alternative method of propagating context updates which avoids walking the Provider's sub-tree

Will update description with more detail soon but high level

## Implementation
- when creating a context dependency (a read) if no `reader` yet exists, attach one.
- readers provide Providers a list of fibers that have read from them
- on propagation, readers are evaluated for updates just like all fibers are today during the walk
- on render a fiber will copy readers into new context dependencies if the context item matches in the same "slot"
- If there is a discrepancy (rare) we stuff the old reader in a lazily created danglingReaders array and worry about cleaning up if this WIP fiber is committed and then renders again
- If during a render there are danglingReaders they are cleaned up in the prepareToReadContext phase
- on unmount all readers are cleaned up

## Pros:
* Providers hold reference to specific fibers that read from them, can jump straight to them rather than walk entire sub-tree. Perf scales at `O(# of readers)` instead of `O(size of tree)`
* Still maintains lightweight mount/unmount overhead

## Perf Impact
* holds reference to the first fiber that read from Provider so unknown whether that fiber or alternate are part of the current tree. need to examine both's dependency list which can in common cases lead to double evaluations (relatively minor given offsets in not walking tree) and in rare cases unnecessary renders (if context reads change from render to render then effectively the past two renders contexts can trigger an update)
* extra object allocation to set up the reader. paid once on mount typically
* small cleanup work on unmount. it shouldn't have major impacts on perf but it's commit phase so I know there is sensitivity here

## Notes
* a lot of the complexity comes from `readContext` since this is the only way you can read from a variable list of contexts on each render
* my new tests were mostly used to observe internal implementations working looking at now removed log statements. They probably shouldn't stick around since they don't test new functionality alone. not sure if it's worth writing tests to assert that for instance readers get cleaned up appropriately but if it is how do I don that and still just test the public api?
* the change to Depdendency_old type should be made explicit, i sorta shortcutted a bit but if this PR is interesting I can certainly refine it

### DIFF
--- a/packages/react-reconciler/src/ReactFiber.old.js
+++ b/packages/react-reconciler/src/ReactFiber.old.js
@@ -27,6 +27,7 @@ import {
   enableFundamentalAPI,
   enableScopeAPI,
   enableBlocksAPI,
+  enableDirectContextPropagation,
 } from 'shared/ReactFeatureFlags';
 import {NoEffect, Placement} from './ReactSideEffectTags';
 import {ConcurrentRoot, BlockingRoot} from './ReactRootTags';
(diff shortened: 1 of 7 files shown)


gnoff:
@sebmarkbage this was somewhat extracted from my PR on speculative work. It's not strictly necessary for any of that but seems perhaps a worthy goal to change the context prop perf scaling in a meaningful way. would love to know if I struck the right balance there while keeping mount, unmount, and "contexts that never update" performance penalties minimal

PR #18860, opened 2020-05-07
Disable Webpack setImmediate polyfill for DevTools

bvaughn:
This pollutes `window`.

I also went ahead and updated Webpack dependencies to the latest versions.


bvaughn:
Indeed.

sebmarkbage:
Should we use Rollup instead like we do for other things?

Webpack is fine for UI or whatever but Rollup's output is so much more predictable for things like the backend.

bvaughn:
I dunno, maybe. I hate mucking about with build config though.

I'll think about it.

PR #18861, opened 2020-05-07
Proposal: react-hooks/exhaustive-deps: a more customizable additionalHooks

dvoytenko:
## Summary

The current `additionalHooks` in the react-hooks/exhaustive-deps eslint plugin accepts a regular expression to match hook names. It always assumes the following format:

```
function useCustomHook(callback, deps) {}
```

However, this format is not always convenient when other arguments are available in the hook's signature, because the set of dependencies could potentially be large. In those cases, it'd be more convenient to place callback and deps as the last two arguments. For instance, the standard `useImperativeHandle` has an additional first argument:

```
function useImperativeHandle(ref, callback, deps) {}
```

To enable this, the proposal adds a new property: `additionalHooksMap` that takes a map keyed by a hook name and value specifying `callback` argument offset.

An example:
```
[{additionalHooksMap: {
    'useCustomEffect': 1,
    'useAnotherCustomEffect': 2,
    ...
}}]
```

## Test Plan

Added new tests for the new format to the eslint tests.


swissspidy:
There's a conflict now because of #18907

dvoytenko:
Thanks, @swissspidy. Rebased.

gaearon:
Mind updating? Let's get this in.

gaearon:
I think we might want to think more about the API here. What if you want to use a regex *and* put it in the map? The existing feature supports regexes. It is also a bit confusing to have two independent options. Why does one of them win? What if a regex matches in one but the key matches in the other one?

gaearon:
Could be something like

```
additionalHooks: [
  [{ test: /regex/, callbackIndex: 3 }],
  ...
]
```

dvoytenko:
> I think we might want to think more about the API here. What if you want to use a regex _and_ put it in the map? The existing feature supports regexes. It is also a bit confusing to have two independent options. Why does one of them win? What if a regex matches in one but the key matches in the other one?

Sure. That makes sense. I'll update.

dvoytenko:
> Could be something like
> 
> ```
> additionalHooks: [
>   [{ test: /regex/, callbackIndex: 3 }],
>   ...
> ]
> ```

@gaearon PTAL. I updated per your request. The linter schema is now either one of these:
- A string - the current mode.
- An array of objects in the shape `{test, callbackIndex}`

swissspidy:
@gaearon would this be good to go after a rebase?

swissspidy:
Opened #19644 to track this proposal in case it helps with visibility...

whatisaphone:
It might be simpler (for end users) to infer the callback index from callsite usage. All the hooks I've ever seen that take a callback have arguments of `(...optionalExtraArgs, callback, deps?)`. So the callback is always second-to-last, or if it's last, the deps array is undefined.

Are there any callback+deps hooks out in the wild that break this assumption?

dvoytenko:
@whatisaphone I think somewhere this was covered as well. Personally, I always order args to have callback and deps as the last 2 (or 1) args. However, the current behavior is exactly the opposite and so we decided to go with a configurable option for backward compatibility.

MatthijsBon:
I would like to add that `useDebounce` (https://github.com/streamich/react-use/blob/master/docs/useDebounce.md) has a different signature than what seems to be proposed here:
```
const [ ... ] = useDebounce(fn: Function, ms: number, deps: DependencyList = []);
```
I would say that specifying the index of the callback is a non-complete addition and would suggest to instead specify the index of the DependencyList?

georeith:
@MatthijsBon I think you can assume that dependency list is the last argument. I can see why callback may potentially be moved around but that feels like a pretty solid convention thats worth encouraging.

dvoytenko:
bump

andrenanninga:
Any chance of getting this or https://github.com/facebook/react/pull/21719 merged? Would love to see this feature.

swissspidy:
bump

vojtech-dobes:
@gaearon Is there a chance of this being merged and shipped? `exhaustive-deps` is now very strict and discourages designing a focused API for custom hook, because it's not possible to teach this rule on which position the callback is present.

kachkaev:
@dvoytenko would you be willing to re-create this PR against the `main` branch? Seems like this PR is closed because the `master` branch has been deleted.

swissspidy:
Quoting https://github.com/facebook/react/pull/21719#issuecomment-1286262543:

> That said, I don't think we will be adding support for this. In general, we've been leaning towards recommending that custom Hooks do not take dependency arrays at all, and that you rely on the built-in Hooks to do that instead.

PR #18863, opened 2020-05-07
[Blocks] Scaffolding react-fetch + first pass at node implementation

maisano:
## Summary

Categorically, there are two elements to this PR. The first is in moving the `react-data` package over to `react-fetch`, and removing the nested `react-data/fetch` entry point. I took some guesses at some of the more structural changes and names used here.

The second part is a low fidelity, first pass at building out a fetch-like for the server implementation of `react-fetch`. For starters, the Node-specific `react-fetch` is a duplicate of the browser version. The fetch implementation is by no means complete but provides at least a start to build from. For instance, only a tiny subset of the API has implemented here, and, practically speaking, much of it seems skippable for the purposes of this library. There are also some obvious gaps (notice the handful of TODOs), as well a Promise usage (instead of a thenable). I also didn't write any tests. The purpose of this was mostly to get some of the server implementation set up so it could built on top of.

Happy to change any of this, or throw any or all of it away.


gaearon:
Awesome, thank you so much!

I added a few tests and took the liberty of simplifying the implementation. In particular, it's better to support fewer features well. Since we don't really have coverage for anything, we need to start simple. Otherwise it will be hard to tell which pieces work and which don't. So I've removed the redirect support for now, as well as the ability to specify options. We can add them back on a case by case basis, but this time with tests.

PR #18864, opened 2020-05-07
Detect and prevent render starvation, per lane

acdlite:
If an update is CPU-bound for longer than expected according to its priority, we assume it's being starved by other work on the main thread.

To detect this, we keep track of the elapsed time using a fixed-size array where each slot corresponds to a lane. ~~What we actually store is the event time when the lane first became CPU-bound.~~ Changed it to store the expiration time, not the start time. See second commit for explanation.

Then, when receiving a new update or yielding to the main thread, we check how long each lane has been pending. If the time exceeds a threshold constant corresponding to its priority, we mark it as expired to force it to synchronously finish, instead of being interrupted again.

We don't want to mistake time elapsed while an update is IO-bound (waiting for data to resolve) for time when it is CPU-bound. So when a lane suspends, we clear its associated event time from the array. When it receives a signal to try again, either a ping or an update, we assign a new event time to restart the clock.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -2301,7 +2301,12 @@ function updateDehydratedSuspenseComponent(
         // is one of the very rare times where we mutate the current tree
         // during the render phase.
         suspenseState.retryLane = attemptHydrationAtLane;
-        scheduleUpdateOnFiber(current, attemptHydrationAtLane);
+        // TODO: The event time is used to prevent starvation if a lane takes
+        // too long to finish rendering. Ideally this would be the time at which
+        // we started rendering. For spawned work, it's not a big deal, though,
+        // because eventually the original update will expire.
+        const eventTime = -1;
+        scheduleUpdateOnFiber(current, attemptHydrationAtLane, eventTime);
       } else {
         // We have already tried to ping at a higher priority than we're rendering with
         // so if we got here, we must have failed to hydrate at those levels. We must
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -112,6 +112,7 @@ import {
   SyncLane,
   OffscreenLane,
   DefaultHydrationLane,
+  NoTimestamp,
   includesSomeLane,
   laneToLanes,
   removeLanes,
@@ -2301,11 +2302,8 @@ function updateDehydratedSuspenseComponent(
         // is one of the very rare times where we mutate the current tree
         // during the render phase.
         suspenseState.retryLane = attemptHydrationAtLane;
-        // TODO: The event time is used to prevent starvation if a lane takes
-        // too long to finish rendering. Ideally this would be the time at which
-        // we started rendering. For spawned work, it's not a big deal, though,
-        // because eventually the original update will expire.
-        const eventTime = -1;
+        // TODO: Ideally this would inherit the event time of the current render
+        const eventTime = NoTimestamp;
         scheduleUpdateOnFiber(current, attemptHydrationAtLane, eventTime);
       } else {
         // We have already tried to ping at a higher priority than we're rendering with
--- a/packages/react-reconciler/src/ReactFiberLane.js
+++ b/packages/react-reconciler/src/ReactFiberLane.js
@@ -29,6 +29,7 @@ export opaque type LanePriority =
   | 16;
 export opaque type Lanes = number;
 export opaque type Lane = number;
+export opaque type LaneMap<T> = Array<T>;
 
 import invariant from 'shared/invariant';
 
@@ -66,7 +67,7 @@ const IdleLanePriority: LanePriority = 2;
 
 const OffscreenLanePriority: LanePriority = 1;
 
-const NoLanePriority: LanePriority = 0;
+export const NoLanePriority: LanePriority = 0;
 
 const TotalLanes = 31;
 
@@ -116,6 +117,8 @@ const IdleUpdateRangeStart = 28;
 const IdleUpdateRangeEnd = 30;
 
 export const OffscreenLane: Lane = /*                   */ 0b1000000000000000000000000000000;
+
+export const NoTimestamp = -1;
 
 // "Registers" used to "return" multiple values
 // Used by getHighestPriorityLanes and getNextLanes:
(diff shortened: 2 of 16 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2301,7 +2301,12 @@ function updateDehydratedSuspenseComponent(
         // is one of the very rare times where we mutate the current tree
         // during the render phase.
         suspenseState.retryLane = attemptHydrationAtLane;
-        scheduleUpdateOnFiber(current, attemptHydrationAtLane);
+        // TODO: The event time is used to prevent starvation if a lane takes
+        // too long to finish rendering. Ideally this would be the time at which
+        // we started rendering. For spawned work, it's not a big deal, though,
```
bvaughn, line 2306: No big deal 😎

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -365,6 +366,46 @@ export function getNextLanes(root: FiberRoot, wipLanes: Lanes): Lanes {
...
+      const priority = return_highestLanePriority;
+      if (priority <= IdleHydrationLanePriority) {
+        // Anything idle priority or lower should never expire.
+      } else {
+        // User interactions should expire slightly more quickly.
+        const expirationHeuristic =
+          priority >= InputContinuousLanePriority ? 1000 : 5000;
+        if (eventTime + expirationHeuristic <= currentTime) {
+          // This lane expired
+          root.expiredLanes |= lane;
```
bvaughn, line 400: What are the implications for entangled updates if we only mark a subset of lanes as expired?

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -555,7 +596,15 @@ export function higherPriorityLane(a: Lane, b: Lane) {
   return a !== NoLane && a < b ? a : b;
 }
 
-export function markRootUpdated(root: FiberRoot, updateLane: Lane) {
+export function createLaneMap<T>(initial: T): LaneMap<T> {
```
bvaughn, line 599: Why `T` instead of a more concrete type? Do we expect to have other types of lane maps in the future?

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -570,17 +619,69 @@ export function markRootUpdated(root: FiberRoot, updateLane: Lane) {
...
   root.suspendedLanes &= higherPriorityLanes;
   root.pingedLanes &= higherPriorityLanes;
+
+  if (eventTime !== -1) {
+    const eventTimes = root.eventTimes;
+    let lanes = updateLane | unsuspendedLanes;
+    while (lanes > 0) {
+      const index = ctrz(lanes);
+
+      // Assumes event times are monotonically increasing. Last one wins.
```
bvaughn, line 636: Sorry, maybe this is me being dumb because it's early, but I'm not sure I understand this comment. Isn't the only assumption this code is making is that `eventTime` is positive? What does "last one wins" mean in this context?

Was this comment more applicable to old expiration times implementation?

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -570,17 +619,69 @@ export function markRootUpdated(root: FiberRoot, updateLane: Lane) {
...
   root.suspendedLanes |= suspendedLanes;
   root.pingedLanes &= ~suspendedLanes;
+
+  // The suspended lanes are no longer CPU-bound. Clear their event times.
+  const eventTimes = root.eventTimes;
+  let lanes = suspendedLanes;
+  while (lanes > 0) {
+    const index = ctrz(lanes);
+
+    eventTimes[index] = -1;
```
bvaughn, line 656: Curious why we use an explicit constant for e.g. `NoLane = 0` but we pass around a naked `-1` in so many places. Should we use a constant for this too (to make it easier to grep for later if nothing else)?

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -671,18 +785,25 @@ export function getBumpedLaneForHydration(
...
+function clz32Fallback(lanes) {
+  if (lanes === 0) {
     return 32;
   }
-  return (31 - ((log(asUint) / LN2) | 0)) | 0; // the "| 0" acts like math.floor
+  return (31 - ((log(lanes) / LN2) | 0)) | 0;
+}
+
+// Count trailing zeros. Only used on lanes, so assume input is an integer.
+function ctrz(lanes) {
```
bvaughn, line 801: nit: Why no Flow type?

packages/react-reconciler/src/ReactFiberLane.js
acdlite: Oops, I meant "first one wins"

The idea is that the times always increase so there's no need to check if an existing time is earlier than the new time. The existing time will always be earlier.

packages/react-reconciler/src/ReactFiberLane.js
acdlite: Yeah I can do that. It won't add any type safety, unfortunately, but it does make it more obvious.

packages/react-reconciler/src/ReactFiberLane.js
bvaughn: Haha, ok that makes a lot more sense. I even wondered if maybe that's what you meant but discarded it.

packages/react-reconciler/src/ReactFiberLane.js
acdlite: Typing with Lanes won't add type safety because inside this module, Lane is only an alias for number. If it were exported, however, then adding Lanes would prevent someone from passing a number that's not of type Lanes. I'll add it anyway, though, in case this ever gets exported (though I don't think we would).

packages/react-reconciler/src/ReactFiberLane.js
acdlite: Most of the ones I have in mind are numbers, but not necessarily. Like the interactions map, for example.

packages/react-reconciler/src/ReactFiberLane.js
acdlite: Entanglement would happen in `getNextLanes`. First we'd detect that a lane had expired, so we'd choose it to work on. Then we'd combine it with the entangled lanes before returning.

bvaughn: Makes sense!

PR #18865, opened 2020-05-08
Add Mounting/Unmounting Effects/Refs to Offscreen Component

lunaruan:
This PR expands on #18729 by introducing an `Offscreen` component. This component will eventually be used to hide or unhide children and mount/unmount all effects in the Offscreen subtree. This PR also changes the behavior of effects and refs in the Offscreen component when components are hidden. Previously, when a component is hidden, its effects and refs do not unmount. This PR changes this so that whenever a component is hidden, its effects and ref do unmount. Layout effects and refs will always be unmounted on hide. Passive effects currently do not unmount by default (though there is a mode in which they do unmount), but we are still deciding what the default behavior should be. 

Note: This PR does yet not modify behavior for the current Suspense implementation. This will be added in a subsequent PR.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -621,14 +621,79 @@ function updateOffscreenComponent(
     pushRenderLanes(workInProgress, subtreeRenderLanes);
   }
 
+  if (nextProps.mode === 'hidden') {
+    return null;
+  } else {
+    reconcileChildren(current, workInProgress, nextChildren, renderLanes);
+    return workInProgress.child;
+  }
+}
+
+function updateLegacyHiddenComponent(
+  current: Fiber | null,
+  workInProgress: Fiber,
+  renderLanes: Lanes,
+) {
+  const nextProps: OffscreenProps = workInProgress.pendingProps;
+  const nextChildren = nextProps.children;
+
+  const prevState: OffscreenState | null =
+    current !== null ? current.memoizedState : null;
+
+  if (nextProps.mode === 'hidden') {
+    if (!includesSomeLane(renderLanes, (OffscreenLane: Lane))) {
+      let nextBaseLanes;
+      if (prevState !== null) {
+        const prevBaseLanes = prevState.baseLanes;
+        nextBaseLanes = mergeLanes(prevBaseLanes, renderLanes);
+      } else {
+        nextBaseLanes = renderLanes;
+      }
+
+      // Schedule this fiber to re-render at offscreen priority. Then bailout.
+      if (enableSchedulerTracing) {
+        markSpawnedWork((OffscreenLane: Lane));
+      }
+      workInProgress.lanes = workInProgress.childLanes = laneToLanes(
+        OffscreenLane,
+      );
+      const nextState: OffscreenState = {
+        baseLanes: nextBaseLanes,
+      };
+      workInProgress.memoizedState = nextState;
+      // We're about to bail out, but we need to push this to the stack anyway
+      // to avoid a push/pop misalignment.
+      pushRenderLanes(workInProgress, nextBaseLanes);
+      return null;
+    } else {
+      // Rendering at offscreen, so we can clear the base lanes.
+      const nextState: OffscreenState = {
+        baseLanes: NoLanes,
+      };
+      workInProgress.memoizedState = nextState;
+      // Push the lanes that were skipped when we bailed out.
+      const subtreeRenderLanes =
+        prevState !== null ? prevState.baseLanes : renderLanes;
+      pushRenderLanes(workInProgress, subtreeRenderLanes);
+    }
+  } else {
+    let subtreeRenderLanes;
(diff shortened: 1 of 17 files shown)


packages/react-reconciler/src/ReactFiberOffscreenComponent.js
```diff
@@ -19,6 +19,7 @@ export type OffscreenProps = {|
   // called "Offscreen." Possible alt: <Visibility />?
   mode?: 'hidden' | 'visible' | null | void,
   children?: ReactNodeList,
+  cleanupPassiveEffects?: boolean,
```
lunaruan, line 22: Was thinking about incorporating this into `mode`, but I couldn't think of a good name. Open to suggestions!

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -550,7 +552,7 @@ function commitLifeCycles(
     case ClassComponent: {
       const instance = finishedWork.stateNode;
       if (finishedWork.effectTag & Update) {
-        if (current === null) {
+        if (current === null || finishedWork.effectTag & Placement) {
```
lunaruan, line 555: Not sure if `Placement` is the right tag to use here or if there's something better.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -621,15 +621,80 @@ function updateOffscreenComponent(
     pushRenderLanes(workInProgress, subtreeRenderLanes);
   }
 
+  if (nextProps.mode === 'hidden') {
+    return null;
```
acdlite, line 625: Unify this with the branch that checks for `hidden` above

packages/shared/ReactFeatureFlags.js
```diff
@@ -102,6 +102,8 @@ export const enableComponentStackLocations = __EXPERIMENTAL__;
 
 export const enableNewReconciler = false;
 
+export const enableHiddenAPI = false;
```
acdlite, line 105: I would call this something that isn't so easily confused with the old API. Even just `enableOffscreenAPI` is slightly better.

Also might as well expose it in the experimental builds. No harm in it.

packages/react/src/React.js
```diff
@@ -99,6 +100,8 @@ export {
   REACT_STRICT_MODE_TYPE as StrictMode,
   REACT_DEBUG_TRACING_MODE_TYPE as unstable_DebugTracingMode,
   REACT_SUSPENSE_TYPE as Suspense,
+  // enableHiddenAPI
+  REACT_OFFSCREEN_TYPE as unstable_Offscreen,
```
acdlite, line 104: Need to add this to `index.modern.fb.js` and `index.classic.fb.js`, too. The tests didn't fail because we don't current tests against run the compiled Facebook build artifacts, only against source.

packages/react-reconciler/src/ReactFiberOffscreenComponent.js
acdlite: Yeah let's make a mode. Name isn't that important for now, so I'd err on the side of the obnoxiously specific, like `hidden-with-aggressive-cleanup`.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -1742,8 +1924,19 @@ function commitWork(current: Fiber | null, finishedWork: Fiber): void {
     case OffscreenComponent:
     case LegacyHiddenComponent: {
       const newState: OffscreenState | null = finishedWork.memoizedState;
+      const newProps = finishedWork.memoizedProps;
```
acdlite, line 1927: These changes should only be made `OffscreenComponent`. Not `LegacyHiddenComponent`.

Also the whole thing can be wrapped in the feature flag, so it gets stripped out of the non-FB builds.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -830,6 +832,186 @@ function commitLifeCycles(
...
+            safelyCallComponentWillUnmount(fiber, instance);
+          }
+        } else {
+          if (fiber.ref !== null) {
+            fiber.effectTag |= Ref;
+          }
+
+          const instance = fiber.stateNode;
+          if (typeof instance.componentDidMount === 'function') {
+            fiber.effectTag |= Update | Placement;
```
acdlite, line 923: Hmm. This isn't going to work, because it assumes that the class fiber is already part of the effect list. That's not always the case. In the begin phase, we might have bailed out before reaching that node. Like if a parent were wrapped in `memo`.

Same goes for hooks effect below.

Need to think about the right solution. Ideally you would call the effect immediately within this function, but it's too early. We're in the mutation phase, not the layout phase.

We could push the effects into an array so we can fire them in the next phase. Like we do for passive effects.

gaearon:
Seems like this was landed via another set of PRs.

PR #18880, opened 2020-05-09
Fix Profiler root change error

### LINKED ISSUE #18010: DevTools Profiler: "Could not find commit data for root …"
The DevTools Profiler occasionally encounters the error:
> Could not find commit data for root "..." and commit ...

This issue is intermittent and we do not currently know hot to reproduce it. **If you can reproduce it** we would love to get any of the following information from you:
* Info about how you reproduce it. (Share your code or site with us?)
* An exported Profiler JSON that contains the bug. (This may be less useful, since the bug likely happens pre-export- but may still be helpful.)

bl00mber:
fix #18010 "Could not find commit data for root …"

How to reproduce:
1. Start profiling
1. Switch color (ReactNativeWeb)
1. Press button (Toggle)
1. Select lowest item on the Flamegraph
1. Change root to Toggle

Sidebar not updates fiberID and fiberName, then trying to get commit data which was on previously selected root, resulting in error.

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/ProfilerContext.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/ProfilerContext.js
@@ -22,7 +22,7 @@ import type {ProfilingDataFrontend} from './types';
 export type TabID = 'flame-chart' | 'ranked-chart' | 'interactions';
 
 export type Context = {|
-  // Which tab is selexted in the Profiler UI?
+  // Which tab is selected in the Profiler UI?
   selectedTabID: TabID,
   selectTab(id: TabID): void,
 
@@ -43,7 +43,7 @@ export type Context = {|
   // 1. The selected root in the Components tree (if it has any profiling data) or
   // 2. The first root in the list with profiling data.
   rootID: number | null,
-  setRootID: (id: number) => void,
+  setRootIDcleanFiber: (id: number) => void,
 
   // Controls whether commits are filtered by duration.
   // This value is controlled by a filter toggle UI in the Profiler toolbar.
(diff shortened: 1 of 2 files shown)


bl00mber:
Another way to fix this is always clean fiberID and fiberName to prevent further bugs related to this logic.

packages/react-devtools-shared/src/devtools/views/Profiler/ProfilerContext.js
```diff
@@ -129,6 +129,38 @@ function ProfilerContextController({children}: Props) {
...
+    (id: number | null, name: string | null) => {
+      selectFiberID(id);
+      selectFiberName(name);
+
+      // Sync selection to the Components tab for convenience.
+      if (id !== null) {
+        const element = store.getElementByID(id);
+
+        // Keep in mind that profiling data may be from a previous session.
+        // In that case, IDs may match up arbitrarily; to be safe, compare both ID and display name.
```
bvaughn, line 145: Ah, right. So this seems to come up in a few places. I know this check was copied from pre-existing code, but looking at it again- the heuristic is flawed. ID and name might still happen to match for imported data.

We should check something on the data itself to show that it was imported or not, and just skip looking in the store for such imported elements.

We could add a new field to the `ProfilingDataFrontend` type to track this. Interested in tackling this? (Totally fine if you aren't, I could do it.)

packages/react-devtools-shared/src/devtools/views/Profiler/ProfilerContext.js
```diff
@@ -43,7 +43,7 @@ export type Context = {|
   // 1. The selected root in the Components tree (if it has any profiling data) or
   // 2. The first root in the list with profiling data.
   rootID: number | null,
-  setRootID: (id: number) => void,
+  setRootIDcleanFiber: (id: number) => void,
```
bvaughn, line 46: I think `setRootIDcleanFiber` is kind of an awkward name. Prefer to just leave the name we pass down through the context as `setRootID` (and change the implementation to also clear the fiber ID like you did)

packages/react-devtools-shared/src/devtools/views/Profiler/ProfilerContext.js
```diff
@@ -237,7 +245,7 @@ function ProfilerContextController({children}: Props) {
       supportsProfiling,
 
       rootID,
-      setRootID,
+      setRootIDcleanFiber,
```
bvaughn, line 248: In other words, `setRootID: setRootIDAndClearFiber`

packages/react-devtools-shared/src/devtools/views/Profiler/ProfilerContext.js
bl00mber: Yeah, it could be interesting to do. Should this field be committed to this PR, the task is all in this comment?

bvaughn: Just to be clear, this doesn't *need* to be done- at least not as part of this PR.

bl00mber: I'll try to do it in the next few days, thanks!

bl00mber: https://github.com/facebook/react/pull/18880

PR #18886, opened 2020-05-11
Modern Event System: revise ancestor logic

trueadm:
This PR revises the logic for the modern event system and how it handles finding the target ancestor fiber. This is to tackle in internal issue that we found recently.

Specifically, this PR fixes an issue where the ancestor instance fiber was incorrectly let through when we encountered a root container that wasn't a `HostComponent` or `HostText`. In that case, we should keep traversing up the DOM tree of the contianer, checking each node if they do return a valid `HostComponent` or `HostText`. In the event that they don't (`getClosestInstanceFromNode` returns `null`) then we shouldn't process the event, as we're in the wrong container to begin with.

The underlying issue that this tackles is specifically where you move a ReactDOM root (reparent it) into an existing ReactDOM root. When you do this, React should align the events to the right roots, but it was letting some slip through because we weren't stopping execution upon not finding an ancestor that was a valid `HostComponent` or `HostText` – meaning the events were duplicated, causing this bug to surface internally. With this fix, the reparenting of the root now works as expected (we support the moving reparenting of roots or portals, just not React managed nodes that are in the tree).

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -417,13 +417,13 @@ export function dispatchEventForPluginEventSystem(
       // sub-tree for that root and make that our ancestor instance.
       let node = targetInst;
 
-      while (true) {
+      mainLoop: while (true) {
         if (node === null) {
           return;
         }
         const nodeTag = node.tag;
         if (nodeTag === HostRoot || nodeTag === HostPortal) {
-          const container = node.stateNode.containerInfo;
+          let container = node.stateNode.containerInfo;
           if (isMatchingRootContainer(container, targetContainerNode)) {
             break;
           }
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -449,18 +449,23 @@ export function dispatchEventForPluginEventSystem(
...
+            const parentNode = getClosestInstanceFromNode(container);
+            if (parentNode === null) {
+              return;
+            }
+            const parentTag = parentNode.tag;
+            if (parentTag === HostComponent || parentTag === HostText) {
+              node = ancestorInst = parentNode;
+              continue mainLoop;
+            }
+            container = container.parentNode;
```
sebmarkbage, line 467: Maybe add a comment about which nodes you expect to see here. One issue is that this might be a HostRoot, but this can also be a SuspenseComponent of a dehydrated boundary. Not sure if this is correct in that case or not.

trueadm: If we hit a `SuspenseComponent`, shouldn't we also continue traversing up to find other containers as they'll still be represented correctly in the DOM, right? I tried to "break" this with dehydrated boundaries, but couldn't, so assumed the heuristcs remained consistent for this case as is for HostComponent/HostText.

PR #18889, opened 2020-05-11
useMutableSource: "Entangle" instead of expiring

acdlite:
A lane is said to be entangled with another when it's not allowed to render in a batch that does not also include the other lane.

This commit implements entanglement for `useMutableSource`. If a source is mutated in between when it's read in the render phase, but before it's subscribed to in the commit phase, we must account for whether the same source has pending mutations elsewhere. The old subscriptions must not be allowed to re-render without also including the new subscription (and vice versa), to prevent tearing.

In the old reconciler, we did this by synchronously flushing all the pending subscription updates. This works, but isn't ideal. The new reconciler can entangle the updates without de-opting to sync.

In the future, we plan to use this same mechanism for other features, like skipping over intermediate `useTransition` states.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.new.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.new.js
@@ -32,7 +32,7 @@ import {
   isSubsetOfLanes,
   mergeLanes,
   removeLanes,
-  markRootExpired,
+  markRootEntangled,
   markRootMutableRead,
 } from './ReactFiberLane';
 import {readContext} from './ReactFiberNewContext.new';
@@ -999,12 +999,11 @@ function useMutableSource<Source, Snapshot>(
         markRootMutableRead(root, lane);
 
         // If the source mutated between render and now,
-        // there may be state updates already scheduled from the old getSnapshot.
-        // Those updates should not commit without this value.
-        // There is no mechanism currently to associate these updates though,
-        // so for now we fall back to synchronously flushing all pending updates.
-        // TODO: This should entangle the lanes instead of expiring everything.
-        markRootExpired(root, root.mutableReadLanes);
+        // there may be state updates already scheduled from the old source.
+        // Entangle the updates so that they render in the same batch.
+        // TODO: I think we need to entangle even if the snapshot matches,
+        // because there could have been an update to a different hook.
+        markRootEntangled(root, root.mutableReadLanes);
       }
     }
   }, [getSnapshot, source, subscribe]);
--- a/packages/react-reconciler/src/ReactFiberLane.js
+++ b/packages/react-reconciler/src/ReactFiberLane.js
@@ -365,6 +365,37 @@ export function getNextLanes(root: FiberRoot, wipLanes: Lanes): Lanes {
     }
   }
 
+  // Check for entangled lanes and add them to the batch.
+  //
+  // A lane is said to be entangled with another when it's not allowed to render
+  // in a batch that does not also include the other lane. Typically we do this
+  // when multiple updates have the same source, and we only want to respond to
+  // the most recent event from that source.
+  //
+  // Note that we apply entanglements *after* checking for partial work above.
+  // This means that if a lane is entangled during an interleaved event while
+  // it's already rendering, we won't interrupt it. This is intentional, since
+  // entanglement is usually "best effort": we'll try our best to render the
+  // lanes in the same batch, but it's not worth throwing out partially
+  // completed work in order to do it.
+  //
+  // For those exceptions where entanglement is semantically important, like
+  // useMutableSource, we should ensure that there is no partial work at the
+  // time we apply the entanglement.
+  const entangledLanes = root.entangledLanes;
+  if (entangledLanes !== NoLanes) {
+    const entanglements = root.entanglements;
+    let lanes = nextLanes & entangledLanes;
+    while (lanes > 0) {
+      const index = ctrz(lanes);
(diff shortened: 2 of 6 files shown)


packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1004,7 +1004,7 @@ function useMutableSource<Source, Snapshot>(
         // There is no mechanism currently to associate these updates though,
         // so for now we fall back to synchronously flushing all pending updates.
         // TODO: This should entangle the lanes instead of expiring everything.
```
bvaughn, line 1006: This TODO is now TODONE, right? 😉

Also the block comment above is kind of stale.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -673,6 +673,15 @@ function performConcurrentWorkOnRoot(root, didTimeout) {
...
   currentEventPendingLanes = NoLanes;
 
+  invariant(
+    (executionContext & (RenderContext | CommitContext)) === NoContext,
+    'Should not already be working.',
+  );
+
+  // Flush any pending passive effects before deciding which lanes to work on,
+  // in case they schedule additional work.
+  flushPassiveEffects();
```
bvaughn, line 683: 👍

packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
```diff
@@ -1481,21 +1481,39 @@ describe('useMutableSource', () => {
...
+      // Now there are two pending mutations at different priorities. But they
+      // both read the same verion of the mutable source, so we must render
+      // them simultaneously.
+      //
+      if (gate(flags => flags.new)) {
+        // In the new reconciler, we can do this with entanglement: when the
+        // high priority render starts, we'll also include the low pri work.
+        expect(Scheduler).toFlushAndYieldThrough([
+          'Parent: 3',
+          // Demonstrates that we can yield here
```
bvaughn, line 1500: 👍

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -365,6 +365,37 @@ export function getNextLanes(root: FiberRoot, wipLanes: Lanes): Lanes {
...
+  // in a batch that does not also include the other lane. Typically we do this
+  // when multiple updates have the same source, and we only want to respond to
+  // the most recent event from that source.
+  //
+  // Note that we apply entanglements *after* checking for partial work above.
+  // This means that if a lane is entangled during an interleaved event while
+  // it's already rendering, we won't interrupt it. This is intentional, since
+  // entanglement is usually "best effort": we'll try our best to render the
+  // lanes in the same batch, but it's not worth throwing out partially
+  // completed work in order to do it.
```
bvaughn, line 380: This comment seems at odds with how `useMutableSource` relies no entanglements to work.

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -705,6 +738,21 @@ export function markRootFinished(root: FiberRoot, remainingLanes: Lanes) {
...
+
+  const entanglements = root.entanglements;
+  let lanes = entangledLanes;
+  while (lanes > 0) {
+    const index = ctrz(lanes);
+    const lane = 1 << index;
+
+    entanglements[index] |= lanes;
+
+    lanes &= ~lane;
```
bvaughn, line 752: This mechanism of clearing each lane after we've processed it- so the next call to `ctrz ` advances the index- does it actually cause a bug in this case?

The result of this loop is that each processed lane isn't marked as being entangled with the preceding lanes, only the following ones.

I *think* that means we could miss updates in the check above?

acdlite: Yeah I just noticed this while working on a PR on top. Good catch!

packages/react-reconciler/src/ReactFiberLane.js
acdlite: For `useMutableSource` it's ok because the `markRootEntangled` call happens in the commit phase, where there's no partial work. I wrote this comment here as a reminder for the future.

The other case we have planned for entanglement is skipping intermediate `useTransition` updates, where you don't want to interrupt.

packages/react-reconciler/src/ReactFiberHooks.new.js
acdlite: I'll rewrite :)

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -999,12 +999,11 @@ function useMutableSource<Source, Snapshot>(
...
-        // there may be state updates already scheduled from the old getSnapshot.
-        // Those updates should not commit without this value.
-        // There is no mechanism currently to associate these updates though,
-        // so for now we fall back to synchronously flushing all pending updates.
-        // TODO: This should entangle the lanes instead of expiring everything.
-        markRootExpired(root, root.mutableReadLanes);
+        // there may be state updates already scheduled from the old source.
+        // Entangle the updates so that they render in the same batch.
+        // TODO: I think we need to entangle even if the snapshot matches,
+        // because there could have been an update to a different hook.
```
acdlite, line 1005: Noticed this while updating the comment. I'll open a separate PR with a test  and fix.

PR #18891, opened 2020-05-12
Expose LegacyHidden type and disable <div hidden /> API in new fork

acdlite:
I will use this internally at Facebook to migrate away from `<div hidden />`. The end goal is to migrate to the Offscreen type, but that has different semantics. This is an incremental step.

I disabled `<div hidden />` in the new fork and updated the tests to use an indirection that picks the correct API. I will remove this once the LegacyHidden (and/or Offscreen) type has landed in both implementations.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -75,6 +75,7 @@ import {
   warnAboutDefaultPropsOnFunctionComponents,
   enableScopeAPI,
   enableBlocksAPI,
+  warnAboutDOMHiddenAttribute,
 } from 'shared/ReactFeatureFlags';
 import invariant from 'shared/invariant';
 import shallowEqual from 'shared/shallowEqual';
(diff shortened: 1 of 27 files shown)


packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -1122,23 +1130,19 @@ function updateHostComponent(
...
-        children: nextChildren,
-        // Check the host config to see if the children are offscreen/hidden.
-        mode: shouldDeprioritizeSubtree(type, nextProps) ? 'hidden' : 'visible',
-      },
-      _owner: __DEV__ ? {} : null,
-    };
-    nextChildren = wrappedChildren;
+  if (__DEV__) {
+    if (
+      false &&
```
acdlite, line 1135: Oops, meant to move this into its own commit. One sec...

scripts/jest/TestFlags.js
```diff
@@ -73,6 +73,10 @@ function getTestFlags() {
       classic: releaseChannel === 'classic',
       www,
 
+      // Using this more specific flag so it's easier to clean up later
+      enableLegacyHiddenType:
+        featureFlags.enableNewReconciler === false || __EXPERIMENTAL__,
```
acdlite, line 78: Name is a bit misleading, but what this says is that the tests will pass if we're in the experimental branch of the new fork (because `unstable_LegacyHidden` is exposed) or if we're in the old fork (because the tests fall back to `<div hidden />` in the old fork).

PR #18897, opened 2020-05-12
deps: bump ESLint version to v7

koba04:
## Summary

This PR is to update ESLint version in `devDependencies` to v7.
The motivation is to verify whether `eslint-plugin-react-hooks` works fine with ESLint v7.
#18878 is a PR to bump the peerDependencies version, but it doesn't guarantee the behavior with ESLint v7.

I've also fixed some failed tests with ESLint v7.
The fails are caused by https://github.com/eslint/eslint/pull/12955, which makes `output` require when the rule has fixed the code.

## Test Plan

`yarn test` && `yarn lint`

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "cross-env": "^6.0.3",
     "danger": "^9.2.10",
     "error-stack-parser": "^2.0.6",
-    "eslint": "^6.8.0",
+    "eslint": "^7.0.0",
     "eslint-config-fbjs": "^1.1.1",
     "eslint-config-prettier": "^6.9.0",
     "eslint-plugin-babel": "^5.3.0",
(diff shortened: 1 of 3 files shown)


koba04:
All checks have passed🚀

gaearon:
thank you

trueadm:
This is now causing master to fail because during our validation script we get the error:

`No files matching the pattern "./build/node_modules/*/umd/*.js" were found.`

Oddly, there are definitely files in there, as it states this just above.

```
Checking if files exist in ./build/node_modules/*/umd/*.js...
Found 32 bundles.
```

gaearon:
I reverted in https://github.com/facebook/react/pull/18899 because this failed master builds. It seems like we only ran this task on master? @koba04 mind looking into this?

koba04:
@gaearon oh, I’ll work on it now!

koba04:
@gaearon @trueadm I've fixed it at #18900.

PR #18898, opened 2020-05-12
Modern Event System: Add scaffolding for createEventHandle

trueadm:
This PR extracts a bunch of the logic out from https://github.com/facebook/react/pull/18756 that implements that basic scaffolding for the `createEventHandle` API. Note this PR does not enable the flag, so it shouldn't affect anything internally. The other changes are just tweaks and nits that the other PR made, not core changes to any of the existing functionality or behaviors.

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -32,7 +32,6 @@ import {
   LEGACY_FB_SUPPORT,
   IS_REPLAYED,
   IS_TARGET_PHASE_ONLY,
-  USE_EVENT_SYSTEM,
 } from './EventSystemFlags';
 
 import {
@@ -78,6 +77,8 @@ import {
   TOP_PLAYING,
   TOP_CLICK,
   TOP_SELECTION_CHANGE,
+  TOP_BEFORE_BLUR,
+  TOP_AFTER_BLUR,
   getRawEventName,
 } from './DOMTopLevelEventTypes';
 import {
@@ -89,7 +90,10 @@ import {batchedEventUpdates} from './ReactDOMUpdateBatching';
 import getListener from './getListener';
 import {passiveBrowserEventsSupported} from './checkPassiveEvents';
 
-import {enableLegacyFBSupport} from 'shared/ReactFeatureFlags';
+import {
+  enableLegacyFBSupport,
+  enableCreateEventHandleAPI,
+} from 'shared/ReactFeatureFlags';
 import {
   invokeGuardedCallbackAndCatchFirstError,
   rethrowCaughtError,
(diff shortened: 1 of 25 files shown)


packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -707,3 +735,10 @@ export function accumulateEnterLeaveListeners(
     );
   }
 }
+
+export function getListenerMapKey(
+  topLevelType: DOMTopLevelEventType,
+  capture: boolean,
+): string {
+  return `${getRawEventName(topLevelType)}__${capture ? 'capture' : 'bubble'}`;
```
gaearon, line 743: This seems awkward. We already started from a name like this (`registrationName`). Then it gets turned into a `topLevelType`. And now we reconstruct it back. Could we pass it through instead?

gaearon: Alternatively, could two separate maps work?

gaearon: Can you briefly describe what the intention is with separating the keys?

trueadm: `getRawEventName` is just a pass through, for returning `topLevelType` as a `string`. It gets inlined and removed by GCC, so there's no cost to it. It's used to avoid having to cast with Flow each time.

We could use separate Maps, but that means allocating more objects and having more overhead, this way means we can just use a single Map. We'd also need three Maps, 1 for the type lookup, that returns an object that has two further Maps for capture and bubble.

gaearon: I mean that we started with an event name, like `onClickCapture`. Then that gets turned into `click` + a boolean. Then this gets turned into `click_capture`. I'm suggesting to pass the original thing through and use it as a key. Would that work?

trueadm: > Can you briefly describe what the intention is with separating the keys?

> I mean that we started with an event name, like onClickCapture. Then that gets turned into click + a boolean. Then this gets turned into click_capture. I'm suggesting to pass the original thing through and use it as a key. Would that work?

I guess these points are related and it requires some backstory.

Firstly, `topLevelType` isn't `onClickCapture`, it's literally `click`. Also `onClickCapture` doesn't really mean that we should use the actual DOM event capture phase, we only use the bubble phase and emulate the capture qualities. For this particular key logic, we want the "real" properties for a given event – as this is important for `createEventHandle` where we actually use the capture phase and bubble phase specified by the user, rather than our internal emulation system, when dealing with events attached to `EventTarget`s not managed by React.

This is also further complicated by the fact that we don't know the registration name by this point (`onClickCapture` for example) because there are two code paths that never pass this to us: event replaying and `createEventHandle`.

gaearon: > Firstly, topLevelType isn't onClickCapture, it's literally click. 

I wasn't referring to `topLevelType`, I was referring to `registrationName` in the external code which eventually gets turned into `topLevelType`.

>Also onClickCapture doesn't really mean that we should use the actual DOM event capture phase, we only use the bubble phase and emulate the capture qualities. For this particular key logic, we want the "real" properties for a given event – as this is important for createEventHandle where we actually use the capture phase and bubble phase specified by the user, rather than our internal emulation system, when dealing with events attached to EventTargets not managed by React.

OK, that makes sense.

trueadm: > I wasn't referring to topLevelType, I was referring to registrationName in the external code which eventually gets turned into topLevelType.

Sorry for that!

PR #18904, opened 2020-05-13
React Flare: fix PressLegacy preventDefault issue

trueadm:
We recently uncovered a bug in the PressLegacy responder in React Flare. @necolas found a bug that occurs because we call `preventDefault` for `Enter` key presses. We should have only been doing this for `Space` key presses, otherwise it means we never get a corresponding `click` fire, because calling `preventDefault` in the key press phase cancels out the click. This PR fixes that, and also fixes the test that had the wrong semantics.

### DIFF
--- a/packages/react-interactions/events/src/dom/PressLegacy.js
+++ b/packages/react-interactions/events/src/dom/PressLegacy.js
@@ -586,7 +586,11 @@ const pressResponderImpl = {
                 !ctrlKey &&
                 !altKey
               ) {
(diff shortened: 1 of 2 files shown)


engprodigy:
Hi and a good day @trueadm . Please, how can I run a sample of react interactions lets say context menu in my front end application with the required flag enabled in my code or in my react build? Or for now it is only available for internal testing? Thanks for your anticipated response.

PR #18909, opened 2020-05-13
Update README

kenshin1215:


bvaughn:
Hi. Not sure what this PR is for, but I don't think adding the extra "ㄋ" character (whatever it is) seems like a good change, so I'm going to close it. Sorry!

PR #18912, opened 2020-05-13
Fix useMutableSource tearing bug

acdlite:
If a source is mutated after initial read but before subscription is set up, it should still entangle all pending mutations even if snapshot of new subscription happens to match.

Test case illustrates how not doing this can lead to tearing.

The fix is to move the entanglement call outside of the block that checks if the snapshot has changed.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.old.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.old.js
@@ -986,15 +986,14 @@ function useMutableSource<Source, Snapshot>(
           suspenseConfig,
         );
         setPendingExpirationTime(root, expirationTime);
-
-        // If the source mutated between render and now,
-        // there may be state updates already scheduled from the old getSnapshot.
-        // Those updates should not commit without this value.
-        // There is no mechanism currently to associate these updates though,
-        // so for now we fall back to synchronously flushing all pending updates.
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberHooks.old.js
```diff
@@ -986,15 +986,14 @@ function useMutableSource<Source, Snapshot>(
...
-        // TODO: Improve this later.
-        markRootExpiredAtTime(root, getLastPendingExpirationTime(root));
       }
+      // If the source mutated between render and now,
+      // there may be state updates already scheduled from the old getSnapshot.
+      // Those updates should not commit without this value.
+      // There is no mechanism currently to associate these updates though,
+      // so for now we fall back to synchronously flushing all pending updates.
+      // TODO: Improve this later.
+      markRootExpiredAtTime(root, getLastPendingExpirationTime(root));
```
bvaughn, line 996: Ah, that's unfortunate. Even more deopts in old mode now.

acdlite:
> Correct me if I'm wrong though, but this tearing case would still be possible for any source where version is not an auto-incremented value (e.g. window.location, maybe even Redux stores).

Yeah sounds right to me. Something to think about.

PR #18913, opened 2020-05-13
Profiler: Skip reading element for imported data

bl00mber:
I've also replaced `export` to `import` in some places where the data is _imported_ actually (was the word `export` in those places used intentionally?)

### DIFF
--- a/packages/react-devtools-shared/src/__tests__/profilingUtils-test.js
+++ b/packages/react-devtools-shared/src/__tests__/profilingUtils-test.js
@@ -16,7 +16,7 @@ describe('profiling utils', () => {
 
   it('should throw if importing older/unsupported data', () => {
     expect(() =>
-      utils.prepareProfilingDataFrontendForImport(
+      utils.prepareProfilingDataFrontendFromExport(
         ({
           version: 0,
           dataForRoots: [],
--- a/packages/react-devtools-shared/src/__tests__/profilingUtils-test.js
+++ b/packages/react-devtools-shared/src/__tests__/profilingUtils-test.js
@@ -16,7 +16,7 @@ describe('profiling utils', () => {
 
   it('should throw if importing older/unsupported data', () => {
     expect(() =>
-      utils.prepareProfilingDataFrontendFromImport(
+      utils.prepareProfilingDataFrontendForImport(
         ({
           version: 0,
           dataForRoots: [],
--- a/packages/react-devtools-shared/src/__tests__/profilingUtils-test.js
+++ b/packages/react-devtools-shared/src/__tests__/profilingUtils-test.js
@@ -16,7 +16,7 @@ describe('profiling utils', () => {
 
   it('should throw if importing older/unsupported data', () => {
     expect(() =>
-      utils.prepareProfilingDataFrontendFromExport(
+      utils.prepareProfilingDataFrontendFromImport(
         ({
           version: 0,
           dataForRoots: [],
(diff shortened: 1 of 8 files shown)


packages/react-devtools-shared/src/__tests__/profilingUtils-test.js
```diff
@@ -16,7 +16,7 @@ describe('profiling utils', () => {
 
   it('should throw if importing older/unsupported data', () => {
     expect(() =>
-      utils.prepareProfilingDataFrontendFromExport(
+      utils.prepareProfilingDataFrontendFromImport(
```
bvaughn, line 19: This was more semantically correct before.

We're *importing* data, but we're reading *from an export*.

If you want to rename it because you find the name confusing, you could name it `prepareProfilingDataFrontendForImport` I guess.

packages/react-devtools-shared/src/devtools/store.js
```diff
@@ -416,6 +416,11 @@ export default class Store extends EventEmitter<{|
   }
 
   getElementByID(id: number): Element | null {
+    if (
+      this._profilerStore._dataFrontend &&
+      this._profilerStore._dataFrontend.imported === true
```
bvaughn, line 421: This is not right.

There are several places in the Components tab that call this method, not just the Profiler. This would break (disable) all of the uses in Components anytime you happened to have profiling data imported in  the Profiler tab.

packages/react-devtools-shared/src/devtools/views/Profiler/utils.js
```diff
@@ -99,21 +99,21 @@ export function prepareProfilingDataFrontendFromBackendAndStore(
...
 
-  return {dataForRoots};
+  return {dataForRoots, imported: false};
 }
 
 // Converts a Profiling data export into the format required by the Store.
-export function prepareProfilingDataFrontendFromExport(
-  profilingDataExport: ProfilingDataExport,
+export function prepareProfilingDataFrontendFromImport(
+  profilingDataImport: ProfilingDataExport,
```
bvaughn, line 107: This isn't really correct. That data you're receiving as a parameter is the previously *exported data*. You're *returning* "imported" data I guess, but...

bl00mber: But? When data is readed from previously exported data it becomes imported data. It may be subjective because the variable name can be explained as a logical entity and not a semantical so I can remove this if it is not incorrect.

packages/react-devtools-shared/src/devtools/store.js
bl00mber: I thought that was the main point, to disable selection on the Components tab?

bvaughn: No. The Components tab should still work for your current app. So things should still be able to ask for info about a particular component.

The Profiler tab just shouldn't expect to load a component from the `Store` for a previous session. So this check needs to be in the Profiler, where it calls `store.getElementByID`

packages/react-devtools-shared/src/devtools/views/Profiler/utils.js
bvaughn: That's *after* it's been processed though. The two formats are different because of serialization constraints. (That's why there are two separate flow types for them.)

packages/react-devtools-shared/src/__tests__/utils.js
```diff
@@ -194,18 +194,20 @@ export function exportImportHelper(bridge: FrontendBridge, store: Store): void {
   );
   const parsedProfilingDataExport = JSON.parse(serializedProfilingDataExport);
 
-  const profilingDataFrontend = prepareProfilingDataFrontendFromExport(
+  const profilingDataFrontendImport = prepareProfilingDataFrontendForImport(
```
bvaughn, line 197: This is still wrong, at least in my mental model. The data that's been prepared isn't an import anymore. It's just "profiling data", as can be seen below when it's assigned to
```js
profilerStore.profilingData = profilingDataFrontendImport;
```

I think these renames confuse me because they go against my mental model of this stuff.

PR #18919, opened 2020-05-14
Add LegacyHidden to server renderer

acdlite:
I think this makes sense, but not entirely sure.

When the tree is hidden, the server renderer renders nothing. The contents will be completely client rendered.

When the tree is visible it acts like a fragment.

The future streaming server renderer may want to pre-render these trees and send them down in chunks, as with Suspense boundaries.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
@@ -2809,4 +2809,62 @@ describe('ReactDOMServerPartialHydration', () => {
     // Now we're hydrated.
     expect(ref.current).not.toBe(null);
   });
+
+  // @gate experimental
+  // @gate new
+  it('renders a hidden LegacyHidden component', async () => {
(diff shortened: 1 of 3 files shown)


acdlite:
Hmm I think maybe this won't work if it hydrates at Offscreen priority, which would happen if it's nested inside a Suspense boundary.

sebmarkbage:
Note that I this is not how the legacy stuff works. So you’re changing behavior to support a feature that exists only so that we don’t change legacy behavior. However this probably has an existing bug in it when priority is not offscreen. Eg due to increased priority of hydration.

acdlite:
> Hydration is normally only suppose to happen in lanes between other priorities (for the scheduler at higher pri case) which needs dedicated lanes. So I think it would make sense for the default hydration to have its own.

Makes sense to me. Started working on this but it's a slightly bigger change since I need to audit our use of `OffscreenLane` as a heuristic, which is subtle and easy to mess up. So I'd prefer to do that inn a follow up and land this PR as-is to fix the hydration warnings and unblock the internal rollout.

Issue #18823, opened 2020-05-04
useMutableSource: allow getSnapshot to return a function


dai-shi:
Ref: https://github.com/reactjs/rfcs/pull/147#issuecomment-623740218

React version: 0.0.0-experimental-e5d06e34b

## Steps To Reproduce

1. createMutableSource that contains functions
2. getSnapshot returns one of the functions

Link to code example:

https://codesandbox.io/s/jovial-dew-1eg4t?file=/src/App.js

## The current behavior

If getSnapshot returns a function, it will run the function in render. This results in various errors.

My workaround: return an object wrapping a function, and destruct on caller.

> For now, you can work around this by returning a function that returns your function

I don't know how this works as workaround.

## The expected behavior

It would be nice if returning a function just works.

Note: this is not a rare use case in my library https://github.com/dai-shi/use-context-selector
because we often pass `[state, setState]` in a context value.

dai-shi:
Thanks. Understood. I'd try not-100%-safe workaround in my lib, or let the lib users do a workaround.

bvaughn:
Thanks for understanding!

dai-shi:
(For those who might come here later,) here's my workaround: https://github.com/dai-shi/use-context-selector/pull/12/commits/2fcf1a2bfb9552daedca11ecff030914e3089aa3#diff-f41e9d04a45c83f3b6f6e630f10117fe

PR #18921, opened 2020-05-14
useMutableSource: Allow returning function from getSnapshot

sophiebits:
basicStateReducer strikes again.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.new.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.new.js
@@ -1012,7 +1012,8 @@ function useMutableSource<Source, Snapshot>(
       const latestSetSnapshot = refs.setSnapshot;
 
       try {
-        latestSetSnapshot(latestGetSnapshot(source._source));
+        const value = latestGetSnapshot(source._source);
+        latestSetSnapshot(() => value);
 
         // Record a pending mutable source update with the same expiration time.
         const suspenseConfig = requestCurrentSuspenseConfig();
--- a/packages/react-reconciler/src/ReactFiberHooks.old.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.old.js
@@ -1004,7 +1004,8 @@ function useMutableSource<Source, Snapshot>(
       const latestSetSnapshot = refs.setSnapshot;
 
       try {
-        latestSetSnapshot(latestGetSnapshot(source._source));
+        const value = latestGetSnapshot(source._source);
+        latestSetSnapshot(() => value);
 
         // Record a pending mutable source update with the same expiration time.
         const currentTime = requestCurrentTimeForUpdate();
--- a/packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
+++ b/packages/react-reconciler/src/__tests__/useMutableSource-test.internal.js
@@ -1494,6 +1494,45 @@ describe('useMutableSource', () => {
   );
 
   // @gate experimental
+  it(
+    'can store functions in source',
+    async () => {
+      const source = createSource(() => 'a');
+      const mutableSource = createMutableSource(source);
+
+      const getSnapshot = () => source.value;
+
+      function Read() {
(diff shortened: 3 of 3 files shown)


dai-shi:
@sophiebits Does this seem to fix #18823?

sophiebits:
@dai-shi It should.

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1012,7 +1012,8 @@ function useMutableSource<Source, Snapshot>(
       const latestSetSnapshot = refs.setSnapshot;
 
       try {
-        latestSetSnapshot(latestGetSnapshot(source._source));
+        const value = latestGetSnapshot(source._source);
+        latestSetSnapshot(() => value);
 
         // Record a pending mutable source update with the same expiration time.
         const suspenseConfig = requestCurrentSuspenseConfig();
```
sophiebits, line 1019: (nb: we still need some wrapper for the exceptions a few lines below this – we could maybe make something different for that to avoid allocating the function in the success case, but… I feel like this is fine?)

acdlite:
Do we actually want to support this or is it more about defining a behavior that's currently poorly defined?

Using a function as your "snapshot" only works if it's pure/idempotent. It doesn't work if it reads from anything mutable (including the source itself).

I can't think of any legit use cases for this. Maybe to make an expensive computation lazy, but that's a subtle thing to get right because you have to pluck out the inputs before closing over them. I would consider that a very advanced use case, and there are other ways to get around it, like by wrapping the computation in an object: `{ readMemoizedComputation() }`

As an alternative, could we say we don't support functions as snapshots and warn in dev if we detect one?

sophiebits:
@dai-shi's use case of storing a constant function in your store like a state setter (https://github.com/dai-shi/use-context-selector/tree/v2) doesn't seem unreasonable to me, though it's not how I would write the code.

But more I noticed this while reading the code and I don't like things that are ostensibly agnostic to the value you put in them except fail on certain ones for no good reason – I assumed it was an oversight.

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1012,7 +1012,8 @@ function useMutableSource<Source, Snapshot>(
       const latestSetSnapshot = refs.setSnapshot;
 
       try {
-        latestSetSnapshot(latestGetSnapshot(source._source));
+        const value = latestGetSnapshot(source._source);
+        latestSetSnapshot(() => value);
```
bvaughn, line 1016: This is an issue that I chatted about with Daishi a few weeks back, which has this obvious fix but... I had hoped to avoid the [almost always] unnecessary function wrapper. Still don't know how I feel about it.

bvaughn: Andrew's comment just loaded in for me after writing mine.

I agree with the concern he mentions about opening up extra problem cases if the function itself closes over mutable values.

> As an alternative, could we say we don't support functions as snapshots and warn in dev if we detect one?

I tend to like this better.

acdlite:
Another option is to use `useReducer` instead. However, we have plans to fork the `useState` and `useReducer` implementations so we can optimize `useState` for simple cases that can quickly bail out. So that might not work longer term.

OTOH, we've talked about making it so `useState` won't support multiple priority levels; that is, if you update at multiple priorities, both of them will get batched together, because that allows us to eagerly compute the final value without worrying that the reducer might change (since it's always `basicStateReducer`). That trade off works best when the updates are local. Doesn't work as well when the updates are entangled with lots of other updates across the whole app, as would often be the case with Flux store-like sources that `useMutableSource` is designed for.

Regardless I still kinda prefer not supporting this and warning.

bvaughn:
I've posted a potential alternative for discussion #18933

bvaughn:
We decided not to support this, but to warn instead (#18933). Just forgot to close the issue.

PR #18928, opened 2020-05-15
Fix lazy not to throw undefined objects

### LINKED ISSUE #18768: Bug: `React.lazy` throws undefined instead of an `Error` object
React version: 16.13.1

## Steps To Reproduce

```js
import React, { Suspense, lazy } from "react";

const Async = lazy(() => {}); // <------

let App = () => {
	return (
		<Suspense fallback={"Loading"}>
			<Async />
		</Suspense>
	);
};

export default App;
```

Link to code example: https://codesandbox.io/s/elegant-fermi-bxwtb

Might be related to https://github.com/facebook/react/issues/15019, https://github.com/facebook/react/pull/15222

## Context

I know that this is incorrect code, but `lazy(() => {import(...)})` is a rather easy typo to make.
The main reason why I'm bringing this is up is that other tools (like JSDOM and Codesandbox with the link above) expect an `Error` object to be thrown. 

## The current behavior

**Input**: `const Async = lazy(async () => ({}));`
**Error**: (this is correct)
```
Warning: lazy: Expected the result of a dynamic import() call. Instead received: [object Object]

Your code should look like: 
  const MyComponent = lazy(() => import('./MyComponent'))
```

**Input**: `const Async = lazy(() => ({}));`
**Error**: (the exception is not an `Error` object)
```
Uncaught {}
```

**Input**: `const Async = lazy(() => {});`
**Error**:  (the exception is not an `Error` object)
```
Uncaught undefined
```

## The expected behavior

**Input**: `const Async = lazy(async () => ({}));`
**Error**:
```
Warning: lazy: Expected the result of a dynamic import() call. Instead received: [object Object]

Your code should look like: 
  const MyComponent = lazy(() => import('./MyComponent'))
```

**Input**: `const Async = lazy(() => ({}));`
**Error**:
```
Warning: lazy: Expected the result of a dynamic import() call. Instead received: undefined

Your code should look like: 
  const MyComponent = lazy(() => import('./MyComponent'))
```

**Input**: `const Async = lazy(() => {});`
**Error**:
```
Warning: lazy: Expected the result of a dynamic import() call. Instead received: [object Object]

Your code should look like: 
  const MyComponent = lazy(() => import('./MyComponent'))
```

b4h0-c4t:
## Summary
fixed #18768
This issue has two problems.

### First problem
`React.lazy` did not perform an undefined check for `moduleObject`.

input
```js
const Async = lazy(async () => {});
```

current
```
Uncaught undefined
```

fixed
```
Warning: lazy: Expected the result of a dynamic import() call. Instead received: undefined

Your code should look like: 
  const MyComponent = lazy(() => import('./MyComponent'))
```

### Second problem
`React.lazy` didn't check if the result of executing the argument was thenable.
input

```js
const Async = lazy(() => ({}));
```

current
```
Uncaught {}
```

fixed
```
lazy: Expects to receive an async function.
```

## Test Plan
`yarn test ReactLazy`


packages/react/src/ReactLazy.js
```diff
@@ -52,14 +53,18 @@ function lazyInitializer<T>(payload: Payload<T>): T {
...
     const thenable = ctor();
+    invariant(
+      thenable !== undefined && typeof thenable.then === 'function',
+      'lazy: Expects to receive an async function.',
+    );
     // Transition to the next state.
     const pending: PendingPayload = (payload: any);
     pending._status = Pending;
     pending._result = thenable;
     thenable.then(
```
sebmarkbage, line 64: What I don't get is why doesn't this throw a proper error instead of `undefined`?

packages/react/src/ReactLazy.js
```diff
@@ -52,14 +53,18 @@ function lazyInitializer<T>(payload: Payload<T>): T {
...
       moduleObject => {
         if (payload._status === Pending) {
-          const defaultExport = moduleObject.default;
+          const defaultExport = moduleObject && moduleObject.default;
```
sebmarkbage, line 67: Similarly, why isn't this property access throwing a proper error that then is passed to the error handler?

packages/react/src/ReactLazy.js
b4h0-c4t: I am assuming that 'invariant' is throwing an error, is it wrong?

packages/react/src/ReactLazy.js
b4h0-c4t: Should I use 'invariant' for the check?

packages/react/src/ReactLazy.js
b4h0-c4t: I think this problem has been resolved, as I believe the invariant is throwing an error properly.

packages/react/src/ReactLazy.js
b4h0-c4t: If throwing error (invariant or other) in thenable.then, it error to be UnhandledPromiseRejectionWarning.
I think that it is a better solution than before the fix.

b4h0-c4t:
@sebmarkbage 
About the second conversation, I got some solutions.
I would like you to tell me which one is best.

**1. Allowing the moduleObject to be undefined.**
It is the current code.
I have interpreted the existing behavior in an expanded way.

**2. Use invariant.**

```js
moduleObject => {
  if (payload._status === Pending) {
    invariant(
      moduleObject !== undefined,
      'lazy: lazy: Expected the result of a dynamic import() call. '
    );
    const defaultExport = moduleObject.default;
    // ... some pragram.
```

It can throw an error using invariant.
But, I think it is inappropriate because it causes UnhandledPromiseRejectionWarning.

**3. Use Rejected**

```js
moduleObject => {
  if (payload._status === Pending) {
    // Transition to the next state.
    if(moduleObject !== undefined){
      const defaultExport = moduleObject.default;
      const resolved: ResolvedPayload<T> = (payload: any);
      resolved._status = Resolved;
      resolved._result = defaultExport;
    } else {
      const rejected: RejectedPayload = (payload: any);
      rejected._status = Rejected;
      rejected._result = error;
    }
    // ... some pragram.
```

It takes the "rejected status" to onFulFill.
it looks natural, but as far as I can see, the error is being managed by invariant.

b4h0-c4t:
bump

VitorLuizC:
bump

zoeyzou:
What is the status of this PR? throwing undefined is still happening

b4h0-c4t:
I think my turn to be over, but what do the maintainers think?

scripts/error-codes/codes.json
```diff
@@ -356,7 +356,7 @@
   "365": "Invalid selector type %s specified.",
   "366": "ReactDOM.createEventHandle: setListener called on an target that did not have a corresponding root. This is likely a bug in React.",
   "367": "ReactDOM.createEventHandle: setListener called on an element target that is not managed by React. Ensure React rendered the DOM element.",
-  "368": "ReactDOM.createEventHandle: setListener called on an invalid target. Provide a valid EventTarget or an element managed by React.",
+  "368": "ReactDOM.createEventHandle: setListener called on an invalid target. Provide a vaid EventTarget or an element managed by React.",
```
bvaughn, line 359: Not reviewing this PR, just glancing over it and noticed this change that looks to have been an accident and should be reverted.

gaearon:
I think there is a misunderstanding here.

The original review asked: **why** is it throwing undefined instead of the original error propagating?

This question has not been answered as far as I can see. This is what you want to answer before making any code changes.

gaearon:
I sent https://github.com/facebook/react/pull/21639 with potential fix. Not sure if it works yet.

gaearon:
We ended up doing this differently (https://github.com/facebook/react/pull/21642), but thank you for nudging to get this done. I agree it was an annoying oversight.

PR #18931, opened 2020-05-15
Update inaccurate mapChildren comment

rickyvetter:
The function you provide will only be passed a child and an index. It will not be passed a key. This is confirmed in the source, the Flow types, and the jsdoc comments.

## Summary

I was actually looking for a child map api that would expose keys and noticed this comment. It gave me the wrong impression and took a few minutes to re-affirm my understanding of React.Children and how it works. Removing this inaccuracy to hopefully prevent a similar misunderstanding in the future.

## Test Plan

I ran absolutely no tests. This is a non-jsdoc comment-only change.

### DIFF
--- a/packages/react/src/ReactChildren.js
+++ b/packages/react/src/ReactChildren.js
@@ -212,7 +212,7 @@ type MapFunc = (child: ?React$Node) => ?ReactNodeList;
  *
  * See https://reactjs.org/docs/react-api.html#reactchildrenmap
  *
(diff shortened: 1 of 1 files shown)


gaearon:
Note it should care of merging with the old key automatically.

rickyvetter:
Thanks for the note. Reading through the source really changed how I view the "keyed-ness" of an element. I'm glad I did.

PR #18933, opened 2020-05-15
Warn if MutableSource snapshot is a function

### LINKED ISSUE #18823: useMutableSource: allow getSnapshot to return a function

bvaughn:
Resolves #18823; potential alternative of #18921.

`useMutableSource` does not properly support snapshots that are functions. In part this is because of how it is implemented internally (the function gets mistaken for a state updater function). To "fix" this we could just wrap another function around the returned snapshot, but this pattern seems problematic to begin with- because the function that gets returned might itself close over mutable values, which would defeat the purpose of using the hook in the first place.

This PR proposes adding a new DEV warning if the snapshot returned is a function. It does not change the behavior (meaning that a function could still "work" in some cases- but at least the current behavior prevents passing around a closure that may later become stale unless you're *really intentional about it* e.g. `() => () => {...}`).

Open to discussion on whether we should lock it down further and treat this as an invariant.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.new.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.new.js
@@ -917,7 +917,7 @@ function readFromUnsubcribedMutableSource<Source, Snapshot>(
     const snapshot = getSnapshot(source._source);
     if (__DEV__) {
       if (typeof snapshot === 'function') {
-        console.warn(
+        console.error(
           'Mutable source should not return a function as the snapshot value. ' +
             'Functions may close over mutable values and cause tearing.',
         );
@@ -1003,7 +1003,7 @@ function useMutableSource<Source, Snapshot>(
       const maybeNewSnapshot = getSnapshot(source._source);
       if (__DEV__) {
         if (typeof maybeNewSnapshot === 'function') {
-          console.warn(
+          console.error(
             'Mutable source should not return a function as the snapshot value. ' +
               'Functions may close over mutable values and cause tearing.',
           );
--- a/packages/react-reconciler/src/ReactFiberHooks.new.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.new.js
@@ -914,7 +914,16 @@ function readFromUnsubcribedMutableSource<Source, Snapshot>(
   }
 
   if (isSafeToReadFromSource) {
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -913,7 +913,16 @@ function readFromUnsubcribedMutableSource<Source, Snapshot>(
   }
 
   if (isSafeToReadFromSource) {
-    return getSnapshot(source._source);
+    const snapshot = getSnapshot(source._source);
+    if (__DEV__) {
+      if (typeof snapshot === 'function') {
+        console.warn(
+          'Mutable source should not return a functions as the snapshot value. ' +
+            'Functions may close over mutable values and cause tearing.',
```
bvaughn, line 921: This message probably isn't great. Happy to change it.

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -913,7 +913,16 @@ function readFromUnsubcribedMutableSource<Source, Snapshot>(
   }
 
   if (isSafeToReadFromSource) {
-    return getSnapshot(source._source);
+    const snapshot = getSnapshot(source._source);
+    if (__DEV__) {
+      if (typeof snapshot === 'function') {
+        console.warn(
```
sebmarkbage, line 919: Should it be console.error? Seems very error like if it might give wrong output and we might make it an invariant.

Issue #18822, opened 2020-05-04
Add devtools "Inspect React component" to right-click menu


sophiebits:
Noticed today that Vue's devtools have an awesome "Inspect Vue component" option when right-clicking on a webpage (added in https://github.com/vuejs/vue-devtools/pull/479).

The React devtools should have the same feature! It would be useful when debugging for sure.

ChrisDobby:
I'd like to have a go at this if its OK?

asheshvidyut7:
@sophiebits please review. https://github.com/facebook/react/pull/18939

locofocos:
I see the PR is almost done! Would love to see this feature. I came here searching to see if this feature existed. I appreciate the work everyone has put in on this so far.

PR #18939, opened 2020-05-17
Enable Context Menu on Chrome to inspect React Component

asheshvidyut7:
Towards https://github.com/facebook/react/issues/18822
 
This PR adds a context menu on chrome. Clicking on the menu, react component is selected on Components tab.  

It is my first PR, will need guidance. 

TrimmedVideo


packages/react-devtools-extensions/chrome/manifest.json
```diff
@@ -40,7 +40,7 @@
     "persistent": false
   },
 
-  "permissions": ["file:///*", "http://*/*", "https://*/*"],
+  "permissions": ["file:///*", "http://*/*", "https://*/*", "contextMenus"],
```
bvaughn, line 43: New permission added only to the Chrome manifest? How does this impact Firefox and Edge?

Can we add this new permission as an [`optional_permissions`](http://www.adambarth.com/experimental/crx/docs/permissions.html) instead?

Maybe we could add an item to the settings menu, e.g. "Enable context menu", that prompts the user to grant the permission. It's not as convenient or discoverable but I'm wary about adding new required permissions. Might cause problems or complaints for people updating.

asheshvidyut7: In this PR, there is no impact on Firefox and Edge. Target was only chrome. Should I add it for them too?
Let me update optional_permissions.

sktguha: Yes that's true. I think chrome will disable the extension, saying that the extension has been disabled as it requires additional permissions. Have to check if contextMenus can be requested as an optional permission. I think it should be

53jk1 (approved):

packages/react-devtools-extensions/chrome/manifest.json
rickhanlonii: @sktguha any update?

sktguha: Oh hi @rickhanlonii , actually I didn't work on this, I was just giving some PR comments.
I think @asheshvidyut7 is the author of the PR.

But I think contextMenu permission can be an optional permission as per the text below
https://developer.chrome.com/docs/extensions/reference/permissions/#step-2-declare-optional-permissions-in-the-manifest:~:text=Most%20Chrome%20extension%20permissions%20can%20be%20specified%20as%20optional%2C%20with%20the%20following%20exceptions.

nt294:
Is this still planned at all?

PR #18942, opened 2020-05-18
Proof-of-concept: Express render phase in terms of generator functions

acdlite:
Combines all the sub-phases of the render phase — begin, complete, and unwind — into a single generator function.

It works by pushing/popping generator objects onto a stack. This replaces the stack in ReactFiberStack.

The cursor objects used by ReactFiberStack are replaced by module level variables that are referenced via closure.

For our release builds, what we would do is compile the generator functions to normal, static functions that push/pop to the stack directly, skipping the overhead of the generator objects.

The first one I've ported is Context providers.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -7,7 +7,7 @@
  * @flow
  */
 
-import type {ReactProviderType, ReactContext} from 'shared/ReactTypes';
+import type {ReactContext} from 'shared/ReactTypes';
 import type {BlockComponent} from 'react/src/ReactBlock';
 import type {LazyComponent as LazyComponentType} from 'react/src/ReactLazy';
 import type {Fiber} from './ReactInternalTypes';
@@ -142,11 +142,8 @@ import {
 } from './ReactFiberSuspenseContext.new';
 import {findFirstSuspended} from './ReactFiberSuspenseComponent.new';
 import {
-  pushProvider,
-  propagateContextChange,
   readContext,
   prepareToReadContext,
-  calculateChangedBits,
   scheduleWorkOnParentPath,
 } from './ReactFiberNewContext.new';
 import {renderWithHooks, bailoutHooks} from './ReactFiberHooks.new';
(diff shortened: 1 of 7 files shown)


acdlite:
This is a rough idea of what I had in mind for the generator output. This is a first draft. I'll add more/better examples later.

Input:

```js
let stack = [];
let threadLocalVariable = 'initial value';

export function* renderSomeComponent(
  current: Fiber | null,
  workInProgress: Fiber,
  renderLanes: Lanes,
): RenderStateMachine {
  const prevValueOnStack = threadLocalVariable;
  threadLocalVariable = 'override subtree with new value';
  try {
    const newChildren = newProps.children;
    reconcileChildren(current, workInProgress, newChildren, renderLanes);
    yield workInProgress.child;
  } finally {
    threadLocalVariable = overrideWithSomeVariable;
  }
}
```

Output:
```js
let stack = [];
let threadLocalVariable = 'initial value';

function renderSomeComponent_0(current, workInProgress, renderLanes) {
  const prevValueOnStack = threadLocalVariable;
  threadLocalVariable = 'override subtree with new value';

  const newChildren = newProps.children;
  reconcileChildren(current, workInProgress, newChildren, renderLanes);
  
  const yieldedValue = workInProgress.child;

  // First slot is the continuation to use if the children render successfully
  // Second slot is if something throws
  stack.push(renderSomeComponent_1, renderSomeComponent_2);
  // Remaining slots are used for local variables
  stack.push(prevValueOnStack);

  return yieldedValue;
}

function renderSomeComponent_1(current, workInProgress, renderLanes) {
  // Load local variables
  const prevValueOnStack = stack[2];

  threadLocalVariable = prevValueOnStack;

  // Pop continuations
  array.pop();
  array.pop();

  // Pop local variables
  array.pop()
  return _context.finish(7);
}

function renderSomeComponent_2(current, workInProgress, renderLanes) {
  // Pop continuations
  array.pop();
  array.pop();

  // Pop local variables
  array.pop()
  return _context.stop();
}
```

chirgjn:
👋 @acdlite 
I'd like to pick this up.
What would be the best way to build upon this? Should I fork acdlite/react ?

acdlite:
@chirgjn Yeah you can fork acdlite:generator-component

awto:
This is very simple and fast to implement with [EffectfulJS](https://github.com/awto/effectfuljs), much simpler than rewriting Regenerator. Let me know if interested. But I also have a few comments if you choose anything else.

There is no way to handle variables captured from generators and finally continuations (`for-of` also needs finally continuation). If you don't use them - fine.

Using functions as state callbacks is much slower than `switch-case`. Effectful JS supports both, so you'll be able to play with, but I've checked a lot of implementations `switch-case` is always much faster. Esp if the function aren't in module's top level, so always recreated. 

Effectful JS can optionally move them to top level if needed, with  handling closure captured variables properly, but `switch-case` is anyway faster.

markerikson:
Out of curiosity, what's the intended goal / benefit of this approach?

tylerhou:
I wrote https://github.com/tylerhou/fiber (HN: https://news.ycombinator.com/item?id=29628772) which may be of interest.

Issue #16444, opened 2019-08-18
Refactor ProfilerContext to use reducer instead of multi-state


bvaughn:
The `ProfilerContext` is currently comprised of several pieces of related state, each managed with `useState`. This necessitates awkward checks like [this](https://github.com/bvaughn/react-devtools-experimental/blob/4697f5b37967b85b2c844044aeebb5b1a740875d/src/devtools/views/Profiler/ProfilerContext.js#L126-L131) or even worse like [this](https://github.com/bvaughn/react-devtools-experimental/blob/4697f5b37967b85b2c844044aeebb5b1a740875d/src/devtools/views/Profiler/SnapshotSelector.js#L62-L73) or [this](https://github.com/bvaughn/react-devtools-experimental/blob/source/src/devtools/views/Profiler/Profiler.js#L71-L83).

This context should be refactored to use a single reducer (`useReducer`) like `TreeContext`. This is a bit more involved at the moment because of suspense and the `ProfilerContext` being higher level than the suspense cache. Although maybe we could work around this by using some sort of [subscription](https://github.com/bvaughn/react-devtools-experimental/blob/4697f5b37967b85b2c844044aeebb5b1a740875d/src/devtools/views/Profiler/ProfilerContext.js#L118-L124)?

See related issues like #16441 and commit [4697f5b](https://github.com/bvaughn/react-devtools-experimental/commit/4697f5b37967b85b2c844044aeebb5b1a740875d).

PR #18948, opened 2020-05-18
Profiler: Fix error on profiling stop

bl00mber:
Fix browser console error which is thrown on profiling stop: `Warning: Cannot update a component ('ProfilerContextController') while rendering a different component ('SnapshotSelector')` by moving code of `SnapshotSelector` containing commit index update logic responsible for error to `ProfilerContext` (according to comment above that code)

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Profiler/ProfilerContext.js
+++ b/packages/react-devtools-shared/src/devtools/views/Profiler/ProfilerContext.js
@@ -8,7 +8,14 @@
  */
 
 import * as React from 'react';
-import {createContext, useCallback, useContext, useMemo, useState} from 'react';
+import {
+  createContext,
+  useCallback,
+  useContext,
+  useMemo,
+  useState,
+  useEffect,
+} from 'react';
 import {unstable_batchedUpdates as batchedUpdates} from 'react-dom';
 import {useLocalStorage, useSubscription} from '../hooks';
 import {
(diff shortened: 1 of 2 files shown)


packages/react-devtools-shared/src/devtools/views/Profiler/ProfilerContext.js
```diff
@@ -235,6 +247,29 @@ function ProfilerContextController({children}: Props) {
...
+        selectCommitIndex(null);
+      }
+    } else if (selectedFilteredCommitIndex >= numFilteredCommits) {
+      if (numFilteredCommits === 0) {
+        selectCommitIndex(null);
+      } else {
+        selectCommitIndex(numFilteredCommits - 1);
+      }
+    }
+  }, [selectedFilteredCommitIndex, numFilteredCommits]);
```
bl00mber, line 271: This is probably a wrong way to do it in ProfilerContext, how to implement it better? Maybe use EventEmitter in SnapshotSelector which will fire update events? Or create a reducer hook which will manage selectCommitIndex calls? Should be this reducer used instead of selectCommitIndex calls in all the places?

bl00mber: Just found related [issue](https://github.com/facebook/react/issues/16444), which state variables should manage that reducer?

PR #18954, opened 2020-05-19
DevTools editable props tweaks

bvaughn:
1. Made non-editable prop text higher contrast (easier to read)
2. Also makes it stand out as different from dimmer placeholder text for "new prop"

#### Light mode
Screen Shot 2020-05-19 at 9 11 05 AM

#### Dark mode
Screen Shot 2020-05-19 at 9 10 56 AM

#### Light mode
Screen Shot 2020-05-19 at 9 08 56 AM

#### Dark mode
Screen Shot 2020-05-19 at 9 10 20 AM

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/KeyValue.css
+++ b/packages/react-devtools-shared/src/devtools/views/Components/KeyValue.css
@@ -3,7 +3,7 @@
 }
 
 .Name {
-  color: var(--color-dim);
+  color: var(--color-attribute-name-not-editable);
   flex: 0 0 auto;
   cursor: default;
 }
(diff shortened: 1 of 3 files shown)


packages/react-devtools-shared/src/devtools/views/Components/KeyValue.css
```diff
@@ -3,7 +3,7 @@
 }
 
 .Name {
-  color: var(--color-dim);
+  color: var(--color-aattribute-name-not-editable);
```
gaearon, line 6: aattribute?

bvaughn: Yeah it's like "attribute" but better

Issue #18085, opened 2020-02-20
Bug: [eslint-plugin-react-hooks] In comment, the version notation is broken.


pvcresin:
eslint-plugin-react-hook@v2.4.0

At `cjs/eslint-plugin-react-hooks.development.js` and `cjs/eslint-plugin-react-hooks.production.min.js`, the version notation in the comment is broken.

v2.3.0 is ok.

Not fatal, but could be a bug.

## The current behavior

```js
/** @license React vundefined
```

## The expected behavior

```js
/** @license React v2.4.0
```

sophiebits:
Confirmed.

v2.3.0: https://unpkg.com/eslint-plugin-react-hooks@2.3.0/cjs/eslint-plugin-react-hooks.development.js

vundefined: https://unpkg.com/eslint-plugin-react-hooks@2.4.0/cjs/eslint-plugin-react-hooks.development.js

The root package.json doesn't have a `version` so I'm confused how this ever worked…

https://github.com/facebook/react/blob/78e816032c8af962343abbf384e06f3e9bae9269/scripts/rollup/wrappers.js#L67

threepointone:
There's a script that updates the versions ci-update-package-versions.js

sophiebits:
I see, it is written here:

https://github.com/facebook/react/blob/0e49074f7a5ba3981ffd28387c8b7f891c7bad24/scripts/release/utils.js#L221-L224

pvcresin:
It looks like that's been fixed in v3. 🎉

PR #18955, opened 2020-05-19
Add version to rollup build

bl00mber:
Replace vundefined in builds to actual version
Related #18085

How to reproduce the bug:
1. run `yarn build react/index --type=UMD`
1. open `/build/node_modules/react/umd/react.production.min.js`
1. run `yarn build react-debug-tools/index --type=NODE`
1. open `/build/node_modules/react-debug-tools/cjs/react-debug-tools.production.min.js`


aap88-afk:
Revert

bl00mber:
@gaearon

PR #18959, opened 2020-05-20
Transfer actualDuration only once for SuspenseList

sebmarkbage:
We used to transfer actualDuration from all children each time a fiber completes or unwinds.

SuspenseList renders one child at a time by adding one more to the list. Each time this happens, we accumulate the time of the previously added children *again*. This makes SuspenseList look exponentially (factorial? I don't math) more expensive than it is.

This fix moves the accumulation (and resetting child expiration time) to the end only when completeWork returns null. I.e. the last pass.

For certain multi-pass we want to keep the accumulated time between passes such as Suspense and error boundaries. In those cases, I explicitly transfer the actualDuration before the next pass starts.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
@@ -1643,13 +1643,14 @@ function completeUnitOfWork(unitOfWork: Fiber): void {
         stopProfilerTimerIfRunningAndRecordDelta(completedWork, false);
       }
       resetCurrentDebugFiberInDEV();
-      resetChildExpirationTime(completedWork);
 
       if (next !== null) {
         // Completing this fiber spawned new work. Work on that next.
         workInProgress = next;
         return;
       }
+
+      resetChildExpirationTime(completedWork);
 
       if (
         returnFiber !== null &&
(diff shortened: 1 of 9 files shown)


packages/react-reconciler/src/__tests__/ReactSuspenseList-test.js
```diff
@@ -2567,4 +2569,170 @@ describe('ReactSuspenseList', () => {
...
+
+    // actualDuration
+    expect(onRender.mock.calls[2][2]).toBe((1 + 4 + 5 + 3) * 2 + 3);
+    // treeBaseDuration
+    expect(onRender.mock.calls[2][3]).toBe(
+      1 +
+        4 +
+        3 +
+        3 +
+        /* Resuspending a boundary also includes the content in base duration but it shouldn't */ 5,
```
sebmarkbage, line 2717: @bvaughn Like I mentioned in chat, I think this might be a bug or at least inconsistent. Seems like base duration includes both for updates (resuspending) but not for initial renders that suspend.

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
```diff
@@ -1643,14 +1643,15 @@ function completeUnitOfWork(unitOfWork: Fiber): void {
...
       resetCurrentDebugFiberInDEV();
-      resetChildExpirationTime(completedWork);
 
       if (next !== null) {
         // Completing this fiber spawned new work. Work on that next.
         workInProgress = next;
         return;
       }
 
+      resetChildExpirationTime(completedWork);
```
sebmarkbage, line 1653: This is the main change and then the rest follows.

sophiebits:
can’t tell from your summary…

if each child has its own correct time but you’re counting the 1st child n times, 2nd child n – 1 times, etc, then it’s quadratic

conversely, if your time for child k includes the times for children 1 through k – 1, then it’s exponential

packages/react-reconciler/src/__tests__/ReactSuspenseList-test.js
bvaughn: Right, it seems not ideal that we count the hidden/suspended branch in the tree duration. It also makes sense *why* we do this though, because of how we accumulate child times.

I think to fix this we would need to make `resetChildExpirationTime` smart enough to not count both branches of `Suspense` component's children.

Want me to try to tackle this with a follow up?

PR #18963, opened 2020-05-20
DevTools: Add root and renderer version to inspected props panel

bvaughn:
Adds the root API and renderer version to the bottom of the "rendered by" list. (If this info is missing, we degrade gracefully.)

# Screenshots

Screen Shot 2020-05-20 at 10 04 22 AM

Screen Shot 2020-05-20 at 10 04 09 AM

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/Badge.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/Badge.js
@@ -26,11 +26,11 @@ export default function Badge({
   type,
   children,
 }: Props) {
-  let totalBadgeCount = 0;
-
-  if (hocDisplayNames !== null) {
-    totalBadgeCount += hocDisplayNames.length;
+  if (hocDisplayNames === null) {
+    return null;
   }
+
+  const totalBadgeCount = hocDisplayNames.length;
 
   return (
     <Fragment>
--- a/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
@@ -45,7 +45,7 @@ import type {
   InspectedElementContextType,
   StoreAsGlobal,
 } from './InspectedElementContext';
-import type {Element, InspectedElement} from './types';
+import type {Element, InspectedElement, Owner} from './types';
 import type {ElementType} from 'react-devtools-shared/src/types';
 
 export type Props = {||};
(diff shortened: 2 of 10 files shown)


packages/react-devtools-shared/src/devtools/views/Components/Badge.js
```diff
@@ -26,12 +26,12 @@ export default function Badge({
   type,
   children,
 }: Props) {
-  let totalBadgeCount = 0;
-
-  if (hocDisplayNames !== null) {
-    totalBadgeCount += hocDisplayNames.length;
+  if (hocDisplayNames === null) {
```
bvaughn, line 29: This fixes a regression from #18802 I think.

packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
```diff
@@ -374,6 +376,15 @@ function InspectedElementView({
     };
   }
 
+  const rendererLabel =
+    rendererPackageName !== null && rendererVersion !== null
+      ? `${rendererPackageName}@${rendererVersion}`
+      : null;
+  const showRenderedBy =
```
bvaughn, line 383: We'll always show this section now, except for v15.

packages/react-reconciler/src/ReactFiberRoot.new.js
```diff
@@ -60,6 +61,20 @@ function FiberRootNode(containerInfo, tag, hydrate) {
   if (enableSuspenseCallback) {
     this.hydrationCallbacks = null;
   }
+
+  if (__DEV__) {
+    switch (tag) {
+      case BlockingRoot:
+        this._debugRootType = 'createBlockingRoot()';
```
bvaughn, line 68: This minimizes the coupling between React and DevTools.

packages/react-devtools-shared/src/devtools/views/Components/SelectedElement.js
```diff
@@ -415,19 +426,27 @@ function InspectedElementView({
...
-                key={owner.id}
-                displayName={owner.displayName || 'Anonymous'}
-                hocDisplayNames={owner.hocDisplayNames}
-                id={owner.id}
-                isInStore={store.containsElement(owner.id)}
-                type={owner.type}
-              />
-            ))}
+            {owners !== null &&
+              owners.length > 0 &&
```
lunaruan, line 433: Are you intentionally checing this condition twice (also in `showRenderedBy`)

bvaughn: Yes for Flow. I guess I could technically get rid of the second `owners.length` check if I moved `owners !== null && owners.length > 0` above into a second variable but it seemed kind of not worth doing.

bvaughn: I'll just add the Flow casting to get rid of the second check though :smile: I don't mind.

PR #18966, opened 2020-05-20
Profiler: Don't count timed out (hidden) subtrees in base duration

bvaughn:
The Profiler previously over-reported the base duration value by including times pent in hidden subtrees.


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1725,6 +1725,18 @@ function resetChildLanes(completedWork: Fiber) {
       treeBaseDuration += child.treeBaseDuration;
       child = child.sibling;
     }
+
+    const isTimedOutSuspense =
```
bvaughn, line 1729: Implementation note: I chose to subtract the base duration value after the child loop, rather than doing the check inline, because a single check with a possible subtraction seemed better than a frequent check inside of the child loop. Happy to refactor if there's concern about the approach.

PR #18970, opened 2020-05-21
Clear fiber.sibling field when clearing nextEffect

trueadm:
In order to help an internal issue with leaking memory from fibers, we should clear the `sibling` field when we mark a fiber for deletion. We previously did this in #16820 but this caused internal errors because there are subtle use cases with the deprecated `findDOMNode` when we need the `sibling` for traversal.

Instead of clearing the `sibling` field at this point, we could also maybe clear it when we also clear the `nextEffect`. This should happen at a later time, after the `findDOMNode` logic has run, allowing the field to properly be cleared and the retained fiber to be GC'd.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -1994,7 +1994,7 @@ function commitRootImpl(root, renderPriorityLevel) {
       const nextNextEffect = nextEffect.nextEffect;
       nextEffect.nextEffect = null;
       if (nextEffect.effectTag & Deletion) {
-        nextEffect.sibling = null;
+        detachFiberAfterEffects(nextEffect);
       }
       nextEffect = nextNextEffect;
     }
@@ -2451,7 +2451,7 @@ function flushPassiveEffectsImpl() {
     // Remove nextEffect pointer to assist GC
     effect.nextEffect = null;
     if (effect.effectTag & Deletion) {
-      effect.sibling = null;
+      detachFiberAfterEffects(effect);
     }
     effect = nextNextEffect;
   }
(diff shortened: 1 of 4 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1993,6 +1993,9 @@ function commitRootImpl(root, renderPriorityLevel) {
     while (nextEffect !== null) {
       const nextNextEffect = nextEffect.nextEffect;
       nextEffect.nextEffect = null;
+      if (nextEffect.effectTag & Deletion) {
+        nextEffect.sibling = null;
```
bvaughn, line 1997: Seems a little unfortunate, since all other Fiber fields get cleared during mutation by `detachFiber`.

I wonder if we should just move the call to `detachFiber` here instead, and add `sibling = null` to it? That way we'd keep all of the nulling out of fields in the same place.

trueadm: I originally did that, but then we had a bunch of code that expects `return` to be `null` before now, so then I ended up having to add `fiber.return = null` in the commit phase and I just thought it seemed not that different (given we want to really detatch there, we're blocked because of `findDOMNode`).

bvaughn: D'oh

Maybe we could have two methods, e.g. `detachFiberMutation` and `detachFiberAfterEffects` (or something, I dunno, names are hard), so it's a little more explicit in the source than a buried `.sibling = null` ?

trueadm: I've made the renames to what you mentioned, seemed logical.

PR #18973, opened 2020-05-21
DevTools: Fix highlight updates Canvas side problem

### LINKED ISSUE #17432: Devtools : highlight box is shown too small.
A bug

When Chrome devtools is highlighting highlight box is shown too small.

image

Highlight box is shown actual template size.

Chrome 78.0.3904.97
React Developer Tools 4.2.0

bvaughn:
Previously we were setting the width/height of the canvas to the screen's available width and height, but the CSS size to the viewport size. This former is generally smaller than the latter, and on some websites (e.g. [everycode.store](https://everycode.store/)) this caused the drawn dimensions to be compressed and look bad.

The reason this behavior was particularly broken for the site above [is because of a CSS rule](https://twitter.com/marvinhagemeist/status/1263506164144316416):
image

So to be fair, an alternate fix would be to override this rule with our inline style.

Resolves #17432


bvaughn:
Thanks for digging in and helping uncover the underlying cause!

PR #18976, opened 2020-05-21
Separate _owner from _debugOwner

gaearon:
We rely on the element owner for two different behaviors:

1. Semantic meaning: If an element is created in a class render method, that instance becomes its String Ref Owner.
2. Debugging info: If an element gets created while rendering an arbitrary fiber, we want to "remember" that for DevTools and some warning messages.

These purposes are currently conflated because we only have `element._owner`. This field is set inconsistently, e.g. it's set in production only for classes, but in development it's set while rendering function components too. This makes it hard to remove in the future when we remove String Refs.

In this PR, we split these two purposes into `element._owner` (populated from `ReactCurrentOwner.current`) and `element._debugOwner` (populated from `ReactDebugCurrentFrame.debugOwner`).

Only String Refs now rely on `element._owner`, so in this PR we no longer set `ReactCurrentOwner.current` for other Fiber types. The DEV-only cases rely on `element._debugOwner`.

TODO:

- [ ] Possibly revert the MountInderterminateComponent case. We may need this if refs worked for Factory Components before.
- [ ] Change DevTools to rely on `_debugOwner` for "rendered by" breadcrumbs.
  - [ ] Perhaps, this means we need to flip it, add a PROD `_refOwner` field, and then keep `_owner` as the DEV-only one? Then we won't need to change DevTools.
- [ ] Maybe, put `isRendering` onto `ReactDebugCurrentFrame`.

### DIFF
--- a/packages/react-client/src/ReactFlightClient.js
+++ b/packages/react-client/src/ReactFlightClient.js
@@ -235,6 +235,12 @@ function createElement(type, key, props): React$Element<any> {
       writable: false,
       value: null,
     });
+    Object.defineProperty(element, '_debugOwner', {
+      configurable: false,
+      enumerable: false,
+      writable: false,
+      value: null,
+    });
   }
   return element;
 }
--- a/packages/react-dom/src/__tests__/ReactFunctionComponent-test.js
+++ b/packages/react-dom/src/__tests__/ReactFunctionComponent-test.js
@@ -158,16 +158,12 @@ describe('ReactFunctionComponent', () => {
     expect(function() {
       ReactTestUtils.renderIntoDocument(<Child test="test" />);
     }).toThrowError(
-      __DEV__
-        ? 'Function components cannot have string refs. We recommend using useRef() instead.'
-        : // It happens because we don't save _owner in production for
-          // function components.
-          'Element ref was specified as a string (me) but no owner was set. This could happen for one of' +
-            ' the following reasons:\n' +
-            '1. You may be adding a ref to a function component\n' +
-            "2. You may be adding a ref to a component that was not created inside a component's render method\n" +
-            '3. You have multiple copies of React loaded\n' +
-            'See https://fb.me/react-refs-must-have-owner for more information.',
+      'Element ref was specified as a string (me) but no owner was set. This could happen for one of' +
+        ' the following reasons:\n' +
+        '1. You may be adding a ref to a function component\n' +
+        "2. You may be adding a ref to a component that was not created inside a component's render method\n" +
+        '3. You have multiple copies of React loaded\n' +
+        'See https://fb.me/react-refs-must-have-owner for more information.',
     );
   });
 
--- a/packages/react-dom/src/client/ReactDOMLegacy.js
+++ b/packages/react-dom/src/client/ReactDOMLegacy.js
@@ -32,17 +32,19 @@ import {
   findHostInstance,
   findHostInstanceWithWarning,
 } from 'react-reconciler/src/ReactFiberReconciler';
+import {isRendering as ReactCurrentFiberIsRendering} from 'react-reconciler/src/ReactCurrentFiber';
 import getComponentName from 'shared/getComponentName';
 import invariant from 'shared/invariant';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import {has as hasInstance} from 'shared/ReactInstanceMap';
 
-const ReactCurrentOwner = ReactSharedInternals.ReactCurrentOwner;
+let ReactDebugCurrentFrame;
 
 let topLevelUpdateWarnings;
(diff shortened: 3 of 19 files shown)


packages/react-client/src/ReactFlightClient.js
```diff
@@ -235,6 +235,12 @@ function createElement(type, key, props): React$Element<any> {
       writable: false,
       value: null,
     });
+    Object.defineProperty(element, '_debugOwner', {
```
gaearon, line 238: Note: this is Flight. Ignore.

packages/react-dom/src/__tests__/ReactFunctionComponent-test.js
```diff
@@ -158,16 +158,12 @@ describe('ReactFunctionComponent', () => {
...
-        ? 'Function components cannot have string refs. We recommend using useRef() instead.'
-        : // It happens because we don't save _owner in production for
-          // function components.
-          'Element ref was specified as a string (me) but no owner was set. This could happen for one of' +
-            ' the following reasons:\n' +
-            '1. You may be adding a ref to a function component\n' +
-            "2. You may be adding a ref to a component that was not created inside a component's render method\n" +
-            '3. You have multiple copies of React loaded\n' +
-            'See https://fb.me/react-refs-must-have-owner for more information.',
+      'Element ref was specified as a string (me) but no owner was set. This could happen for one of' +
```
gaearon, line 161: This message was previously inconsistent between DEV and PROD. Now we show the same message.

packages/react-dom/src/client/ReactDOMLegacy.js
```diff
@@ -32,17 +32,19 @@ import {
   findHostInstance,
   findHostInstanceWithWarning,
 } from 'react-reconciler/src/ReactFiberReconciler';
+import {isRendering as ReactCurrentFiberIsRendering} from 'react-reconciler/src/ReactCurrentFiber';
```
gaearon, line 35: This is a bit awkward. One alternative is to put `isRendering` onto `ReactDebugCurrentFrame` itself.

packages/react-reconciler/src/ReactFiber.new.js
```diff
@@ -454,7 +454,7 @@ export function createFiberFromTypeAndProps(
   type: any, // React$ElementType
   key: null | string,
   pendingProps: any,
-  owner: null | Fiber,
+  debugOwner: null | Fiber,
```
gaearon, line 457: This was already only used for a better error message, and null in prod.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -1384,7 +1381,6 @@ function mountIndeterminateComponent(
     }
 
     setIsRendering(true);
-    ReactCurrentOwner.current = workInProgress;
```
gaearon, line 1387: This is probably unsafe and might break String Refs in Factory Components. Not sure if we support those.

packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -10,6 +10,7 @@
 import type {Fiber} from './ReactInternalTypes';
 import type {Container, SuspenseInstance} from './ReactFiberHostConfig';
 import type {SuspenseState} from './ReactFiberSuspenseComponent.old';
+import {isRendering as ReactCurrentFiberIsRendering} from './ReactCurrentFiber';
```
gaearon, line 13: This is actually a bit of a red flag because for third-party renderers, `/reflection` entry point won't "see" the value. So maybe this is a good motivation to move `isRendering` into `ReactDebugCurrentFrame`.

packages/shared/ReactElementType.js
```diff
@@ -22,6 +22,7 @@ export type ReactElement = {|
   _owner: any,
 
   // __DEV__
+  _debugOwner: any,
```
bvaughn, line 25: Could use an inline comment to differentiate `_owner` and `_debugOwner`.

bvaughn:
By the way this changes DevTools behavior (which is why CI tests are failing). We'll need to update DT as well.

gaearon:
> By the way this changes DevTools behavior (which is why CI tests are failing)

Yes, hence one of the TODOs. :-)

bvaughn:
Sorry, I didn't noticed the TODOs in the list.

packages/react-client/src/ReactFlightClient.js
aap88-afk: Just you

packages/react-dom/src/__tests__/ReactFunctionComponent-test.js
aap88-afk: Revert

PR #18992, opened 2020-05-24
Fix: bug (styles/lastProp).hasOwnProperty is not a function

eddyw:
## Summary

I've found a strange issue when using style objects with prototype `null` or that override `__proto__` which causes React to crash with:

```
[First props] Uncaught TypeError: styles.hasOwnProperty is not a function
[Diff props] Uncaught TypeError: lastProp.hasOwnProperty is not a function
```

## Test Plan

I added two tests that check this behavior. It basically happens in these cases:

1. Style object has prototype of `null`:
```js
const styles = Object.create(null)
return <span style={styles}>Hey</span>
```
This is because React does:
```js
styles.hasOwnProperty(styleName)
```
However, objects with prototype `null` don't have `hasOwnProperty` method.

2. Style object has `__proto__` that overrides `hasOwnProperty`
```js
const styles = { __proto__: { hasOwnProperty: '' } }
return <span style={styles}>Hey</span>
```
This is similar to the previous in that the prototype overrides `hasOwnProperty` method. In this case to something that is not a `function`.

These are probably edge cases but it doesn't seem safe to rely on `hasOwnProperty` method from the provided `style` prop. In my case, it was a port of a styled system that created objects with prototype `null`.

From what I understand, this doesn't or shouldn't affects `lastProps/nextProps` because create React elements method does already use `Object.prototype.hasOwnProperty` to create a shallow copy of props.

As a suggestion, maybe React could export `Object.prototype.hasOwnProperty` to a shared module as `has` or `hasOwn` method and use it everywhere in order to avoid this type of issues.

### DIFF
--- a/packages/react-dom/src/__tests__/CSSPropertyOperations-test.js
+++ b/packages/react-dom/src/__tests__/CSSPropertyOperations-test.js
@@ -271,7 +271,7 @@ describe('CSSPropertyOperations', () => {
     class Comp extends React.Component {
       render() {
         const style = {
-          __proto__: { hasOwnProperty: '' }
+          __proto__: {hasOwnProperty: ''},
         };
         return <div style={style} />;
       }
--- a/packages/react-dom/src/__tests__/CSSPropertyOperations-test.js
+++ b/packages/react-dom/src/__tests__/CSSPropertyOperations-test.js
@@ -254,4 +254,30 @@ describe('CSSPropertyOperations', () => {
 
     expect(root.children[0].style.getPropertyValue('--foo')).toEqual('5');
   });
+
+  it('should render when style object has `null` prototype', () => {
+    class Comp extends React.Component {
+      render() {
+        const style = Object.create(null);
+        return <div style={style} />;
+      }
+    }
+
+    const root = document.createElement('div');
+    ReactDOM.render(<Comp />, root);
+  });
+
+  it('should render when style object has `__proto__`', () => {
+    class Comp extends React.Component {
+      render() {
+        const style = {
+          __proto__: { hasOwnProperty: '' }
+        };
+        return <div style={style} />;
+      }
+    }
+
+    const root = document.createElement('div');
+    ReactDOM.render(<Comp />, root);
+  });
 });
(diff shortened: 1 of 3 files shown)


gaearon:
Thanks for the PR. Can you speak about the actual use case where you're running into this? In general proto-less objects aren't very common in JS and you can't get one accidentally, so it is not clear to me that there is value in going the extra mile to support them for this case. It's not that it hurts, but then we need to be very concrete about all the places where they're allowed and add coverage to avoid regressions. It's not obvious to me that this is worth the maintenance overhead.

eddyw:
@gaearon 👋 fair enough,

The style object it's generated by a styled system (custom but close API to this one: https://styled-system.com/). This styled system validates user generated styles and only allows certain CSS attributes & values. The objects are always created with `Object.create(null)` before assigning the properties. Its primary focus was/is to generate style sheets (CSS-in-JS). However, for certain libraries such as a third-party modal that generates it's own styles using the `style` object, I tried to plug our styled system into it, so it can still validate properties (if mixed with user generated), sort of like:
```js
const style = system({ ...fromLib, ...userProvided })
```
And that causes the issue.

It's in fact an edge case since I haven't actually found others that have this issue. 🤔 

If it's worth looking at, I also found this that is somewhat related:

```javascript
<div style={{ cssText: "color: hotpink;" }}>Something</div>
```
It'll cause mis-match in SSR because it'll be generated as `style="css-text:color: hotpink;"`. On client, this allows to set the style using strings and it'll bypass React checking the property values which shouldn't be that big of an issue in modern browsers .. I think (the value is escaped anyways).

To sum it up, 
- It doesn't feel safe to rely solely on `instance.hasOwnProperty` to check for a property value (Anywhere, not specifically `style`). Most as a sort of a best practice.
- React doesn't check which properties are set on `CSSStyleDeclaration` instance. So you could set all sort of illegal values `{ cssText: '', 0: 'color', length: 0 }` (SSR mismatch, client crashes)

All in all, after writing this, I don't know if maybe this should be user responsibility only or if maybe React could be improved a bit here 😅

gaearon:
I don't think there is a strong case for doing this for `style` values alone but then not doing it for e.g. `props` in general. Which is an improvement we could technically make, but it's also not clear that this is serving a broad enough need. I think it makes sense to do the opposite — assume that objects passed to React are plain objects with the default object prototype. That's the status quo and if you have a proposal for changing it across all cases, it might be worth writing an RFC for why it's useful.

PR #18994, opened 2020-05-25
[DevTools] Add DevTools forked Feature flags

### LINKED ISSUE #18843: "Pause on caught exceptions" enabled always breaks on this line
In Chrome, when I am using the current stable version of react devtools extenison (4.6.0 (3/26/2020)), if I have "Pause on caught exceptions" enabled, I always hit this breakpoint:

https://github.com/facebook/react/blob/c2d1561c60507189b43ae96c6b89580874367e36/packages/react-devtools-shared/src/hook.js#L185

`ReferenceError: i is not defined
    at Object.inject (<anonymous>:1:1116)
    at injectInternals (react-dom.development.js:9842)
    at injectIntoDevTools (react-dom.development.js:20715)
    at eval (react-dom.development.js:21244)`

It would be nice if the published extension had an options page so that I could disable this try/catch block. (for example enabling NODE_ENV=test or something). 

Or, perhaps that code block is not really relevant to the chrome extension case, and that code path could be ignored for the chrome extension.

bl00mber:
resolve #18843
another option is just use process.env.TARGET


bl00mber:
@bvaughn

packages/react-devtools-shared/config/resolveTarget.js
```diff
@@ -0,0 +1,64 @@
...
+    case 'backend':
+      flagsPath = 'TargetFlags.default';
+      break;
+    case 'extension':
+      flagsPath = 'TargetFlags.extension';
+      break;
+    default:
+      console.error('invalid TARGET:', target);
+      process.exit(1);
+  }
```
bvaughn, line 57:
```suggestion
  switch (target) {
    case 'local':
    case 'remote':
    case 'inline':
    case 'standalone':
    case 'backend':
      flagsPath = 'TargetFlags.default';
      break;
    case 'extension':
      flagsPath = 'TargetFlags.extension';
      break;
    default:
      console.error('invalid TARGET:', target);
      process.exit(1);
  }
```

packages/react-devtools-shared/config/resolveTarget.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+function validateWebpackTarget(target) {
```
bvaughn, line 9: This method seems unnecessary since `resolveTargetFlags` also errors and exits.

bvaughn (changes requested):
IMO this feels a little over-architected for what it is. If you renamed "target flags" to be "feature flags" (a parallel to `shared/ReactFeatureFlags`) then maybe it would be okay. I have had thoughts about adding a feature flag system for a while so that I could try new things out in the internal Facebook build without committing to releasing them to open source until it's stable.

packages/react-devtools-shared/src/hook.js
```diff
@@ -172,7 +173,7 @@ export function installHook(target: any): DevToolsHook | null {
     // In that case, we'll patch later (when the frontend attaches).
     //
     // Don't patch in test environments because we don't want to interfere with Jest's own console overrides.
-    if (process.env.NODE_ENV !== 'test') {
+    if (enableConsolePatching && process.env.NODE_ENV !== 'test') {
```
bvaughn, line 176: Doest his actually work for you? Or does the Webpack-generated import statement just throw an error too (for the same reason as is mentioned in the inline comment:
```
        // Normally we could check "typeof patchConsole === 'function'",
        // but Webpack wraps imports with an object (e.g. _backend_console__WEBPACK_IMPORTED_MODULE_0__)
        // and the object itself will be undefined as well for the reasons mentioned above,
        // so we use try/catch instead.
```

When I tested it I see an error being thrown (and swallowed):
> ReferenceError: react_devtools_target_flags__WEBPACK_IMPORTED_MODULE_1__ is not defined

bvaughn (commented):
This PR does not resolve #18843 for reasons I mentioned in my comment on `src/hooks`. (Now we just get an error from a different import.)

We *could* solve this problem by adding an inline `process.env.NODE_ENV` type check which is what I think I'll do.

I think there's still some value for continuing the momentum of this PR though and creating a DevTools feature flag system, so I'm going to take over this branch and finish out that part :smile: I think I'll make use of the flag for logging purposes anyway.

bvaughn:
Please hold off on additional commits.

bl00mber:
Thanks for the review! I've added a commit to rename TargetFlags to FeatureFlags, but I cannot yet confirm that this will work as intended - just waited for the confirmation to push forward before continuing working on the PR.

packages/react-devtools-extensions/webpack.config.js
```diff
@@ -49,9 +51,11 @@ module.exports = {
   },
   plugins: [
     new DefinePlugin({
-      __DEV__: false,
-      __PROFILE__: false,
+      __DEV__,
       __EXPERIMENTAL__: true,
+      __EXTENSION__: true,
```
bvaughn, line 56: This didn't need to be passed in as `process.NODE_ENV.TARGET`. It's only true for one of the Webpack config files (this one).

packages/react-devtools-shared/src/hook.js
```diff
@@ -172,7 +172,11 @@ export function installHook(target: any): DevToolsHook | null {
     // In that case, we'll patch later (when the frontend attaches).
     //
     // Don't patch in test environments because we don't want to interfere with Jest's own console overrides.
-    if (process.env.NODE_ENV !== 'test') {
+    //
+    // Note that because this function is inlined, this conditional check must only use static booleans.
+    // Otherwise the extension will throw with an undefined error.
+    // (See comments in the try/catch below for more context on inlining.)
+    if (!__EXTENSION__ && !__TEST__) {
```
bvaughn, line 179: This (boolean check) is what we needed to avoid runtime error.

bvaughn (approved):
Let's make sure CI is happy, then I'll merge.
Thanks for kicking this off!

Issue #19001, opened 2020-05-26
Bug: Cannot use **use...Effect** names for custom hooks definition


japutko:
I had following code:

```ts
export const useSummaryHeaderActionsEffect = (): void => {
    const dispatch = useDispatch();

    useEffect(() => {
        dispatch(fetchAction);
    }, [dispatch]);
};
```

which is called in component like:

```ts
export const SummaryHeader: FC = () => {
    useSummaryHeaderActionsEffect();
    ...
}
```

and I kept receiving this error on lint (some content anonymised):

```
TypeError: Cannot read property 'type' of undefined
Occurred while linting ...SummaryHeader.tsx:9
    at visitCallExpression (node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js:823:24)
    at node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (node_modules/eslint/lib/linter/node-event-generator.js:254:26)
    at NodeEventGenerator.applySelectors (node_modules/eslint/lib/linter/node-event-generator.js:283:22)
    at NodeEventGenerator.enterNode (node_modules/eslint/lib/linter/node-event-generator.js:297:14)
    at CodePathAnalyzer.enterNode (node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:634:23)
    at node_modules/eslint/lib/linter/linter.js:936:32
    at Array.forEach (<anonymous>)
    at runRules (node_modules/eslint/lib/linter/linter.js:931:15)
    at Linter._verifyWithoutProcessors (node_modules/eslint/lib/linter/linter.js:1157:31)
    at Linter._verifyWithConfigArray (node_modules/eslint/lib/linter/linter.js:1255:21)
    at Linter.verify (node_modules/eslint/lib/linter/linter.js:1210:25)
    at Linter.verifyAndFix (node_modules/eslint/lib/linter/linter.js:1400:29)
    at verifyText (node_modules/eslint/lib/cli-engine/cli-engine.js:230:48)
    at CLIEngine.executeOnFiles (node_modules/eslint/lib/cli-engine/cli-engine.js:798:28)
    at Object.execute (node_modules/eslint/lib/cli.js:212:111)
    at Object.<anonymous> (node_modules/eslint/bin/eslint.js:107:28)
    .
[...]

Issue #18888, opened 2020-05-11
[eslint-plugin-react-hooks] false positive with useFocusEffect


leethree:
Upgraded `eslint-plugin-react-hooks` to 4.0.0, it's showing errors on React-Navigation's `useFocusEffect` hook.

```js
import { useFocusEffect } from '@react-navigation/native';

  useFocusEffect(
    React.useCallback(() => {
      const unsubscribe = API.subscribe(userId, user => setUser(user));
      return () => unsubscribe();
    }, [userId])
  );
```
(example copied from [`React-Navigation` documentation](https://reactnavigation.org/docs/use-focus-effect/))

## The current behavior

Lint error:

```
React Hook useFocusEffect received a function whose dependencies are unknown. Pass an inline function instead.
```

## The expected behavior

The rules shouldn't check the function because it's not a React `useEffect` hook.

gaearon:
>The rules shouldn't check the function because it's not a React useEffect hook.

Sigh. People have been asking for the opposite, so we added this for every function ending with `Effect`.

vkurchatkin:
Admittedly,  `useFocusEffect` is weird. It could easily work exactly like `useEffect` instead of requiring `useCallback`. Also you could probably just use `useIsFocused` + `useEffect` to avoid this issue

leethree:
> Sigh. People have been asking for the opposite, so we added this for every function ending with `Effect`.

😢 I completely understand it. But it's difficult to ask all library authors to follow the same rules. Maybe adding an option would help.

leethree:
Sounds reasonable to me.

> we'd reserve Effect suffix for things that mirror the dependency API.

Would be great if this could be added to documentation. I would file an issue to `react-navigation` asking them to change the hook name 😄

surgeboris:
> Sigh. People have been asking for the opposite, so we added this for every function ending with Effect.

It is very nice indeed that you're implementing features people were asking for, but it's really worth to describe those features [in the changelog](https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/CHANGELOG.md) at least (because not everyone is aware what others were asking for).

gaearon:
@surgeboris When I added this changelog, I took twenty minutes to go through all the past PRs, remember the context behind them, and reconstruct it. I'm sorry that I missed one of them.

gaearon:
Fixed changelog: https://github.com/facebook/react/commit/14e554b3103cc009c86f48ed927f2552649600a6.

[...]

Issue #18858, opened 2020-05-07
[eslint-plugin-react-hooks] async functions should be allowed for custom effects


vkurchatkin:
Consider the following code:

```js
useAsyncEffect(async () => {
  await foo()
}, []);
```

Eslint give the following error:

```js
  11:18  error  Effect callbacks are synchronous to prevent race conditions. Put the async function inside:

useEffect(() => {
  async function fetchData() {
    // You can await here
    const response = await MyAPI.getData(someId);
    // ...
  }
  fetchData();
}, [someId]); // Or [] if effect doesn't need props or state
```

There are 2 problems with that:

1. The error is provided by `react-hooks/exhaustive-deps`, but this has nothing to do with deps or exhaustiveness. I can't disable this warning without disable actually checking for exhaustive deps.
2. In general the error is bogus. Statement "Effect callbacks are synchronous to prevent race conditions" is only true for built-in effects, and is definitely not true for `useAsyncEffect`, which built specifically to support asynchronous callbacks.

eslint-plugin-react-hooks version: 4.0.0

## The current behavior

`react-hooks/exhaustive-deps` gives an error for this code

## The expected behavior

`react-hooks/exhaustive-deps` should not give an error for this code

threepointone:
https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L104 is where the check happens. 

iiuc, this is to allow passing back a cleanup/cancel function, which is much harder with your abstraction. I hope you've considered those ramifications and have your own workarounds. 

I dunno whether the team will change the regex, would be interested to hear from them.

vkurchatkin:
Just wanted to offer my perspective. I consider it plain wrong to place restrictions like this on my hooks and if this is intended behaviour,  my only option is to drop `exhaustive-deps` completely

threepointone:
You could also rename your hook. An Effect has a specific shape and meaning in React, so maybe co-opting it might not be the right move.

vkurchatkin:
> An Effect has a specific shape and meaning in React, so maybe co-opting it might not be the right move.

Technically, I could, I've thought of that. But `useAsyncEffect` is the obvious name for what it does. I don't want to choose objectively worse name because of a linter rule.

gaearon:
Fixed in `eslint-plugin-react-hooks@4.0.3` by removing the flawed heuristic.

whatisaphone:
[...]

PR #19004, opened 2020-05-26
Revert "Add support to `exhaustive-deps` rule for any hook ending with `Effect`"

gaearon:
Reverts #18580.

I thought we might try this but there's both bugs in the implementation (https://github.com/facebook/react/issues/19001) and too many false positives (https://github.com/facebook/react/pull/18580#issuecomment-633513512, https://github.com/facebook/react/pull/18580#issuecomment-632269045, https://github.com/facebook/react/issues/18858, https://github.com/facebook/react/issues/18888). I think it's clear this heuristic hasn't worked out.


swissspidy:
I'd like to propose #18861 as an alternative solution for this

PR #19009, opened 2020-05-27
Support Babel's envName option in React Refresh plugin

kevinweber:
## Summary

[`babel.getEnv()`](https://github.com/babel/babel/blob/f6c7bf36cec81baaba8c37e572985bb59ca334b1/packages/babel-core/src/config/helpers/environment.js) ignores Babel's [`envName`](https://babeljs.io/docs/en/options#envname) option that can be used, for example, when calling `transformFromAst` from `@babel/core` directly. This can throw an error if `skipEnvCheck == false` and `process.env.BABEL_ENV !== 'development'`.

Using [`babel.env()`](https://babeljs.io/docs/en/config-files#apienv) prioritizes `envName` over `process.env.<x>` values and resolves the issue.

## Test Plan

Added a test. If I change `babel.env()` back to `babel.getEnv()`, the test fails.

### DIFF
--- a/packages/react-refresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-refresh/src/ReactFreshBabelPlugin.js
@@ -8,7 +8,7 @@
 'use strict';
 
 export default function(babel, opts = {}) {
-  if (typeof babel.getEnv === 'function') {
+  if (typeof babel.env === 'function') {
     // Only available in Babel 7.
     const env = babel.env();
     if (env !== 'development' && !opts.skipEnvCheck) {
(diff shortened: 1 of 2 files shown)


packages/react-refresh/src/ReactFreshBabelPlugin.js
```diff
@@ -10,7 +10,7 @@
 export default function(babel, opts = {}) {
   if (typeof babel.getEnv === 'function') {
```
gaearon, line 11: Should we change this check too?

kevinweber: @gaearon Good callout, thanks! It's fixed now.

kevinweber:
@gaearon Any update on getting this merged? It's not urgent; I'm just guessing that this will never get merged unless I keep pinging you :D

kevinweber:
@acdlite Mind taking a look at this? Or close this PR if nobody is going to merge this fix 🙏

gaearon:
Seems good to me

PR #19017, opened 2020-05-27
Don't build unstable entry point for STABLE

gaearon:
Let's see what breaks?

This just excludes bundles from the build. The entry points would still be there. Maybe that's a reasonable protection. I don't know if any tools try to crawl the entry points preemptively and if shipping intentionally missing ones could cause any issues. As a layer above, we could also exclude the entry points themselves but that adds more moving parts and not sure if worth the complexity.

### DIFF
--- a/scripts/rollup/build.js
+++ b/scripts/rollup/build.js
@@ -516,6 +516,10 @@ function shouldSkipBundle(bundle, bundleType) {
       return true;
     }
   }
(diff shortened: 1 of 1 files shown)


sebmarkbage:
I moved the branching on EXPERIMENTAL to the bundles file in this PR: https://github.com/facebook/react/pull/19025/files#diff-6c6554c881731a0c8ac33c8ddba3876eR357

We can configure this in the bundles file instead.

PR #19021, opened 2020-05-27
Clean up passive effects experimental flags

### LINKED ISSUE #17945: Passive effect destroy and create functions are interleaved

bvaughn:
Cleanup this experiment. It was a success and we'll roll it out in the upcoming major.

Resolves #17945.


bvaughn:
Seems like we need to decide whether this is going to go out in v17 or v18 before I land it. @sebmarkbage seemed unsure about v17.

sebmarkbage:
Meh. Let's just not use it as precedence to add more.

PR #19022, opened 2020-05-27
Add ES2015 Node Build Config without ESM nor PROD/DEV

sebmarkbage:
I've had this pending in a branch for a while so I figured I'd just get it in.

The webpack plugin doesn't really need a separate prod and dev build. It also needs to be ES2015 otherwise we can't extend native classes.

Notably, unlike other build variants we'll add for ESM, this compiles to CommonJS. This is basically the current idiomatic output for Node.js script builds.

We could probably use this for other Node scripts as well like eslint-plugin-react-hooks.

### DIFF
--- a/packages/react-transport-dom-webpack/npm/plugin.js
+++ b/packages/react-transport-dom-webpack/npm/plugin.js
@@ -1,7 +1,3 @@
 'use strict';
 
-if (process.env.NODE_ENV === 'production') {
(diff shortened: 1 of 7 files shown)


sebmarkbage:
I need a special rule for linting these. I could use a special directory but I think the lint should use the same setup as the build so that we can use the same set of configurations.

sebmarkbage:
This is now stacked on top of, and blocked on, https://github.com/facebook/react/pull/19025

PR #19028, opened 2020-05-28
[DevTools] Support FunctionComponent.contextTypes

### LINKED ISSUE #15039: react-debug-tools doesn't support legacy context (Component.contextType)
Original report: https://github.com/facebook/react-devtools/issues/1304

Repro case: https://64yojj2wyk.codesandbox.io/

bl00mber:
resolve #15039
Since stateless components (FunctionComponent) have no state, legacyContext cannot be extracted from them. The proposed solution is to traverse fiber tree upwards until stateful component is met and extract legacyContext from it.


sebmarkbage:
This argument is about to be removed from React and replace with other things so not sure it's worth supporting at this point. DevTools would need to fork to try to figure out if it should include it or not.

bl00mber:
Maybe add version check and disable on renderers > v17 then?

sebmarkbage:
Another case is that forwardRef passes a second arg. That kind of doesn't work already but that's probably fine since it needs to be resilient to missing refs anyway. However, it shouldn't pass context to a forwardRef component even if it has contextTypes.

(We also have an experimental feature, Blocks, that provides a second arg but I'll remove that one.)

bl00mber:
Disabled for all types except FunctionComponent.

gaearon:
Seems like it doesn't merge cleanly?

bl00mber:
Updated

bvaughn:
Why are there changes to the `yarn.lock` file in this branch?

bl00mber:
There was a bug which prevented from building devtools plugins. Seems it has been [fixed](#19647) since, so I'm reverting the commit intended to fix it.

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -2250,6 +2266,25 @@ export function attach(
...
+          if (
+            current.stateNode &&
+            current.stateNode.__reactInternalMemoizedMergedChildContext
+          ) {
+            childContextFound = true;
+            context =
+              current.stateNode.__reactInternalMemoizedMergedChildContext;
+          }
+          current = current.return;
+        }
```
bvaughn, line 2287: Rather than check based off of a feature flag, let's use a check like React uses to see if `contextTypes` was declared?

```suggestion
      } else if (tag === FunctionComponent) {
        const hasLegacyContext = !!fiber.elementType.contextTypes;
        if (hasLegacyContext) {
          let current = fiber.return;
          let childContextFound = false;
          while (current !== null && childContextFound === false) {
            const instance = current.stateNode;
            if (
              instance &&
              instance.__reactInternalMemoizedMergedChildContext
            ) {
              childContextFound = true;
              context = instance.__reactInternalMemoizedMergedChildContext;
            }
            current = current.return;
          }
        }
```

This could potentially give us a false positive I guess, if someone had a React component with the legacy context property on it with a newer version of React, but the behavior in the case would just be an unnecessary tree traversal.

The behavior in the current case (going off the flag check) is that we're potentially passing a context in when it's not asked for.

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -594,21 +599,42 @@ export function inspectHooks<Props>(
   return buildTree(rootStack, readHookLog);
 }
 
-function setupContexts(contextMap: Map<ReactContext<any>, any>, fiber: Fiber) {
+function setupContexts(
```
bvaughn, line 602: The overloading of this function (sometimes a return type, sometimes not) and the newly-added `enableLegacyContext` flag, seems a little confusing to me. Let's split them up?

```js
function getLegacyContext(fiber: Fiber): Object | null {
  const hasLegacyContext = !!fiber.elementType.contextTypes;
  if (hasLegacyContext) {
    let current = fiber;
    while (current !== null) {
      const instance = current.stateNode;
      if (instance && instance.__reactInternalMemoizedMergedChildContext) {
        return instance.__reactInternalMemoizedMergedChildContext;
      }
      current = current.return;
    }
  }
  return null;
}

function setupContext(contextMap: Map<ReactContext<any>, any>, fiber: Fiber) {
  let current = fiber;
  while (current) {
    if (current.tag === ContextProvider) {
      const providerType: ReactProviderType<any> = current.type;
      const context: ReactContext<any> = providerType._context;
      if (!contextMap.has(context)) {
        // Store the current value that we're going to restore later.
        contextMap.set(context, context._currentValue);
        // Set the inner most provider value on the context.
        context._currentValue = current.memoizedProps.value;
      }
    }
    current = current.return;
  }
}
```

packages/react-debug-tools/src/ReactDebugHooks.js
```diff
@@ -686,16 +713,25 @@ export function inspectHooksOfFiber(
...
   try {
-    setupContexts(contextMap, fiber);
-    if (fiber.tag === ForwardRef) {
+    if (fiber.tag === FunctionComponent) {
+      const legacyContext = setupContexts(
+        contextMap,
+        fiber,
+        enableLegacyContext,
+      );
+      return inspectHooks(type, props, currentDispatcher, legacyContext);
```
bvaughn, line 722: Let's separate the code paths a little more explicitly?
```suggestion
    if (fiber.tag === FunctionComponent) {
      const hasLegacyContext = !!fiber.elementType.contextTypes;
      if (hasLegacyContext) {
        const legacyContext = getLegacyContext(fiber);
        return inspectHooks(type, props, currentDispatcher, legacyContext);
      } else {
        setupContext(contextMap, fiber);
        return inspectHooks(type, props, currentDispatcher);
      }
```

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -111,7 +111,12 @@ type ReactTypeOfSideEffectType = {|
   Placement: number,
 |};
 
+type VersionFlags = {|
+  enableLegacyContext: boolean,
+|};
```
bvaughn, line 116: We don't need this new type of flag. (See below.)

bvaughn (changes requested):
I'm concerned by the coupling between the `react-debug-tools` package and a legacy React feature that only DevTools should support going forward. (This PR does not introduce that coupling. It just highlights it.)

A change like this should probably have several new unit tests as well (for both `react-debug-tools` and React DevTools). You down to include those with this PR?

bvaughn:
> A change like this should probably have several new unit tests as well (for both `react-debug-tools` and React DevTools). You down to include those with this PR?

I'm going to add test coverage for the DevTools side of this:
https://github.com/facebook/react/pull/19774

For now, I'll have to disable the function + legacy context tests though.

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -125,6 +126,7 @@ export function getInternalReactConstants(
   ReactPriorityLevels: ReactPriorityLevelsType,
   ReactTypeOfSideEffect: ReactTypeOfSideEffectType,
   ReactTypeOfWork: WorkTagMap,
+  versionFlags: VersionFlags,
```
bvaughn, line 129: Accidentally left over?

bl00mber:
Added a test which fails when PR contents are disabled

packages/react-debug-tools/src/__tests__/ReactHooksInspectionIntegration-test.js
```diff
@@ -335,6 +337,46 @@ describe('ReactHooksInspectionIntegration', () => {
...
+      <LegacyContextProvider>
+        <FunctionalLegacyContextConsumer />
+      </LegacyContextProvider>,
+    );
+
+    const childFiber = renderer.root
+      .findByType(FunctionalLegacyContextConsumer)
+      ._currentFiber();
+    const hasLegacyContext = !!childFiber.elementType.contextTypes;
+    if (hasLegacyContext) {
```
bvaughn, line 373: Why is the test doing this check? Seems kind of like lower-level implementation details, and could also lead to a false positive (e.g. replace with `!!childFiber.elementType.contextTypess` above and the test would still pass because the boolean check would just fail silently).

Since we *know* the legacy context is being used here, why do we need a conditional at all?

bl00mber: That was my probably inappropriate attempt to prevent test from running on versions where context should be deprecated.
Removed a conditional & added value verification.

Ishaan28malik (approved):

bvaughn (commented):
I think the changes on the DevTools side look fine, but I'm still a little concerned about the added complexity from this changes to `react-debug-tools` (and the fact that it adds an optional parameter to the exported `inspectHooks` method that `react-debug-tools` will need to continue supporting in future releases– unless we break with a major bump).

What are your thoughts about this, @sebmarkbage?

bvaughn (approved):

PR #19030, opened 2020-05-28
Delete entrypoints without exports from package.json and build output

sebmarkbage:
_This is an alternative approach instead of #19029_

If a file contains a .stable.js or .experimental.js fork, and that fork doesn't have any exports, then we remove the file from the build directory and from the package.json.

This allows us to remove bundles from being built in the stable release channel.


sebmarkbage:
There's really only one more file here because the fb ones aren't affected.

This approach has a flaw though because it still builds the bundle. So it actually ends up in the /cjs/ folder.

sebmarkbage:
I'll go with the other approach since it avoids including it into the /cjs/ folder.

PR #19033, opened 2020-05-28
Add Lazy Elements Behind a Flag

sebmarkbage:
We really needed this for Flight before as well but we got away with it because Blocks were lazy but with the removal of Blocks, we'll need this to ensure that we can lazily stream in part of the content.

Luckily LazyComponent isn't really just a Component. It's just a generic type that can resolve into anything kind of like a Promise.

So we can use that to resolve elements just like we can components. This allows keys and props to become lazy as well.

To accomplish this, we suspend during reconciliation. This causes us to not be able to render siblings because we don't know if the keys will reconcile. For initial render we could probably special case this and just render a lazy component fiber like we do with other lazy stuff. This is easier in blocking/concurrent mode only since the semantics for the LazyComponent tag is just to indefinitely suspend when rendered.

Throwing in reconciliation didn't work correctly with direct nested siblings of a Suspense boundary before but it does now so it depends on new reconciler.

### DIFF
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -33,7 +33,11 @@ import {
   Block,
 } from './ReactWorkTags';
 import invariant from 'shared/invariant';
-import {warnAboutStringRefs, enableBlocksAPI} from 'shared/ReactFeatureFlags';
+import {
+  warnAboutStringRefs,
+  enableBlocksAPI,
+  enableLazyElements,
+} from 'shared/ReactFeatureFlags';
 
 import {
   createWorkInProgress,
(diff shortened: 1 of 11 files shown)


gaearon:
>Throwing in reconciliation didn't work correctly with direct nested siblings of a Suspense boundary before but it does now so it depends on new reconciler.

Is that difference in behavior accidental or intentional?

sebmarkbage:
> Is that difference in behavior accidental or intentional?

Accidental. I mean we intentionally didn't put things that threw in reconciliation due to this quirk. So it wasn't observable before.

The original source has to do with errors throwing. If an error throws inside an error boundary for whatever reason, we don't ever want to let that error boundary try to handle it. However, reconciling children isn't really part of the error boundary, it's just accidentally how it works.

Even if we wanted to keep this for errors, it doesn't have to apply to suspense boundaries. It just accidentally did because we treated them same as errors. We could do the traversal one level up for throwing promises.

However, since the new reconciler changes how suspense boundaries work, this issue accidentally got solved anyway.

PR #19035, opened 2020-05-28
Exploration: find why we leak fibers in new reconciler

trueadm:
Note: this is more of an exploration PR for finding why we leak fibers internally with the new reconciler when we didn't with the older one. I don't expect us to merge it.

After speaking to Andrew recently, he mentioned that there might be a place that we don't properly set the return pointer on a fragment fiber. I checked between the old and new reconcilers for all return pointers that we mutate and this was the only difference between the two. I'm not 100% sure if it's still applicable here, but I'll put this up for now anyway.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -859,6 +859,8 @@ function hideOrUnhideAllChildren(finishedWork, isHidden) {
       ) {
         // Found a nested Offscreen component that is hidden. Don't search
         // any deeper. This tree should remain hidden.
+        const fallbackChildFragment: Fiber = (node.child: any).sibling;
+        fallbackChildFragment.return = node;
       } else if (node.child !== null) {
         node.child.return = node;
         node = node.child;


trueadm:
I guess this isn't right. Another thing I've maybe found that differs is how we mark fragments for Deletion between reconcilers, specifically:

We did this before in the old:

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberCompleteWork.old.js#L926

We don't seem to do it anymore in the new one:

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberCompleteWork.new.js#L911-L912

acdlite:
> Another thing I've maybe found that differs is how we mark fragments for Deletion between reconcilers

I think this one is fine. The deletion moved to the begin phase (which is where deletions are normally marked, during child reconciliation): https://github.com/facebook/react/blob/4985bb0a80f5cbeaa61d21a7daf7da5ecff2d892/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L2056-L2061

Also if we weren't deleting them correctly, then the nodes would still be visible in the DOM.

trueadm:
@acdlite Gotcha. Then I'm still stumped. I'll take another look again tomorrow.

Issue #19002, opened 2020-05-26
Bug: DevTools tabs don't show up sometimes in latest release


bengry:
# Note this issue is due to [CR bug 1085215](https://bugs.chromium.org/p/chromium/issues/detail?id=1085215)
**It is impacting several other popular extensions as well- including React, Redux, Relay, and Vue devtools.**

-----

React version: 16.13.1
DevTools version: 4.7.0 (5/18/2020)
macOS version: 10.15.4 (19E287)

## Steps To Reproduce

1. Create a new profile in Google Chrome
2. Install the [React Developer Tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi)
3. Go to https://reactjs.org
4. Open the browser Dev Tools

## Screenshots
Not sure this helps, but this is basically what I get after following the aforementioned steps:
![image](https://user-images.githubusercontent.com/260431/82867560-c70eb600-9f33-11ea-9e95-d59ce4e3e548.png)

When using Opera the tabs do show up:
![image](https://user-images.githubusercontent.com/260431/82867684-f4f3fa80-9f33-11ea-9b1d-2a54f8a9101e.png)

## The current behavior
Sometimes the Components and Profiler (i.e. React devtools tabs) don't show up.

## The expected behavior
These tabs should show for any site using React.

## More details

I've tried a few browsers, including:
1. Google Chrome (83.0.4103.61)
2. Microsoft Edge (83.0.478.37)
3. Opera (68.0.3618.125)

All of them should be working, since they're all Chromium based.

The most reliable was Opera, which always shows the dev tools tabs. Chrome and Edge behaved the same way, sometimes showing them and other times not.

I've also taken a look at the background pages of the React Dev Tools extension and they don't show any errors, only these performance metrics in the `devtools_app.html` page:
![image](https://user-images.githubusercontent.com/260431/82867862-413f3a80-9f34-11ea-9b7c-ccd0c8225b01.png)

The main background page shows nothing in the console at all times.

Roman-Popov:
Hi there. I have the same problem. Maybe this info will be useful:
1) When I open some of "pages" (let it be `P1`) of my SPA, then open DevTools I can't see "Components" and "Profiler" tabs. 
2) I close DevTools
3) Then I switch to the other "page" of SPA (let it be `P2`)
4) Open DevTools while I'm on `P2`
5) Now I see "Components" and "Profiler" tabs
[...]

PR #19036, opened 2020-05-28
Revert recent Webpack changes [DRAFT]

bvaughn:
In case they're related to the recent increase in DevTools panels not appearing (#19002)

Reverts #18860

[ReactDevTools.zip](https://github.com/facebook/react/files/4697348/ReactDevTools.zip)


bvaughn:
This has been determined to not be the cause.

PR #19039, opened 2020-05-29
Lint autofix wants me to fix this

sebmarkbage:
I'm not sure why this isn't just failing the lint on CI though. It's supposed to fail if const is not used.


sebmarkbage:
I'll just merge this to unblock reenabling the lint.

PR #19040, opened 2020-05-29
Revert autofix lint

sebmarkbage:
I accidentally committed this since I had it on locally so I didn't have to manually convert things to const.

However, this causes things to always pass lint since CI also runs this which fixes the lint violations in CI without committing it.

We should probably add this as an option to the command line tool. Ideally all the eslint options should be passable.


sebmarkbage:
Ya'll probably don't see it locally because it runs against only changed files. The way that's implemented doesn't work with how I use git so it runs against everything for me.

sebmarkbage:
This reveals this violation. https://github.com/facebook/react/pull/19039

PR #19044, opened 2020-05-29
Remove unnecessary throw catch

sebmarkbage:
This was originally added so you could use "break on caught exceptions" but that feature is pretty useless these days since it's used for feature detection and Suspense.

The better pattern is to use the stack trace, jump to source and set a break point here.

Since DevTools injects its own console.error, we could inject a "debugger" statement in there. Conditionally. E.g. React DevTools could have a flag to toggle "break on warnings".


bvaughn:
Nice!

PR #19048, opened 2020-05-29
DevTools: Add break-on-warn feature

### LINKED ISSUE #19045: [DevTools Feature Request] Break on Warnings
It'd be nice to have a toggle to pause the debugger when warnings fire so you can inspect the stack as it's happening.

```
console.error = function() {
  ...
  if (isBreakOn) {
    debugger;
  }
}
```

See https://github.com/facebook/react/pull/19044

bvaughn:
This commit adds a new tab to the Settings modal: Debugging

This new tab has the "append component stacks" feature and a new one: "break on warn". This new feature adds a `debugger` statement into the console override.

Note that the proposed feature was to break *only* on `console.error` but the proposed wording "break on warn" made this confusing in my opinion, so this implementation breaks on both `console.error` and `console.warn`. It would be trivial to change this though if there's a concern.

Resolves #19045

## Demo
breakonwarn

### DIFF
--- a/packages/react-devtools-shared/src/backend/console.js
+++ b/packages/react-devtools-shared/src/backend/console.js
@@ -80,9 +80,25 @@ export function registerRenderer(renderer: ReactRenderer): void {
   }
 }
 
+const consoleSettingsRef = {
+  appendComponentStack: false,
+  breakOnConsoleErrors: false,
+};
+
 // Patches whitelisted console methods to append component stack for the current fiber.
 // Call unpatch() to remove the injected behavior.
-export function patch(): void {
+export function patch({
+  appendComponentStack,
+  breakOnConsoleErrors,
+}: {
+  appendComponentStack: boolean,
+  breakOnConsoleErrors: boolean,
+}): void {
+  // Settings may change after we've patched the console.
+  // Using a shared ref allows the patch function to read the latest values.
+  consoleSettingsRef.appendComponentStack = appendComponentStack;
+  consoleSettingsRef.breakOnConsoleErrors = breakOnConsoleErrors;
+
   if (unpatchFn !== null) {
     // Don't patch twice.
     return;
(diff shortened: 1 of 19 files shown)


packages/react-devtools-shared/src/backend/console.js
```diff
@@ -105,40 +121,56 @@ export function patch(): void {
...
           }
-        } catch (error) {
-          // Don't let a DevTools or React internal error interfere with logging.
+        }
+
+        if (latestBreakOnConsoleErrors) {
+          // --- Welcome to debugging with React DevTools ---
+          // This debugger statement means that you've enabled the "break on warnings" feature.
+          // Use the browser's Call Stack panel to step out of this override function-
+          // to where the original warning or error was logged.
```
bvaughn, line 171: Unfortunately this comment will be stripped during the build process. I think I need to add `output:{comments: true}` to Terser config too...

adir1661:
this feature had forced me to break many times without marking the checkbox of "Break on warning"

gaearon:
@adir1661: If you want to report a problem please file a new issue and provide as much information as you can. We don’t track issue reports in merged PR comments.

adir1661:
ok will consider it, thanks!

packages/react-devtools-shared/src/utils.js
```diff
@@ -248,6 +249,25 @@ export function setAppendComponentStack(value: boolean): void {
...
+export function getBreakOnConsoleErrors(): boolean {
+  try {
+    const raw = localStorageGetItem(
+      LOCAL_STORAGE_SHOULD_BREAK_ON_CONSOLE_ERRORS,
+    );
+    if (raw != null) {
+      return JSON.parse(raw);
+    }
+  } catch (error) {}
+  return true;
```
gaearon, line 261: This doesn't seem intentional?

gaearon:
It's already tracked in https://github.com/facebook/react/issues/19308.

packages/react-devtools-shared/src/utils.js
bvaughn: Nope. This was an oversight.

PR #19053, opened 2020-05-30
Bugfix: Infinite loop in beforeblur event

acdlite:
If the focused node is hidden by a Suspense boundary, we fire the beforeblur event. Our check for whether a tree is being hidden isn't specific enough. It should only fire when the tree is initially hidden, but it's being fired for updates, too.

To optimize this, we should use a dedicated effect tag and mark it in the render phase. I've left this for a follow-up, though. Maybe can revisit after the planned refactor of the commit phase.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberTreeReflection.js
+++ b/packages/react-reconciler/src/ReactFiberTreeReflection.js
@@ -333,13 +333,21 @@ export function findCurrentHostFiberWithNoPortals(parent: Fiber): Fiber | null {
   return null;
 }
 
+// This function detects when a Suspense boundary goes from visible to hidden.
+// It returns false if the boundary is already hidden.
+// TODO: Use an effect tag. And move this to ReactFiberCommitWork.
 export function isFiberSuspenseAndTimedOut(fiber: Fiber): boolean {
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactFiberTreeReflection.js
```diff
@@ -333,13 +333,21 @@ export function findCurrentHostFiberWithNoPortals(parent: Fiber): Fiber | null {
   return null;
 }
 
+// This function detects when a Suspense boundary goes from visible to hidden.
+// It returns false if the boundary is already hidden.
+// TODO: Use an effect tag. And move this to ReactFiberCommitWork.
 export function isFiberSuspenseAndTimedOut(fiber: Fiber): boolean {
```
trueadm, line 339: This function is also used by Scopes for traversal, will this change affect that too? See:

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberScope.new.js#L51

acdlite: My mistake, didn't search for other uses.

I moved the logic into the commit phase.

acdlite:
Thanks for the quick review!

Issue #19020, opened 2020-05-27
Feature request: have Context.Provider throw error if missing `value` prop


heath-freenome:
Just about every time I set up a new `Context.Provider`, I end up accidentally specifying a `values` prop rather than `value`.  While it's a minor error, generally I build the container in which the provider resides and commit it to the code base before I ever use it.  It's only later when I go to use it that I realize I did it again.  Since the `Context.Provider` seems pretty much useless without a `value` prop specified, I'd love it if there was a prop error if it is missing... especially if another prop is defined on the `Context.Provider` instead.

gaearon:
I think it does make sense to warn if `!('value' in props)`. Want to send a PR?

bvaughn:
This issue is all yours, @brunogonzales 

I've added the **"good first issue (taken)"** label so that others will know not to start work on the issue. If you change your mind about the issue, no worries! Just let me know so that I can remove the label and free it up for someone else to claim.

Cheers!

heath-freenome:
@brunogonzales go for it! I wouldn't have time to get to it for a month at least.  Please tag me on the PR as I would love to see the solution.

charlie1404:
@brunogonzales Have you started working on it ?? 😅 
I have made changes locally to add warning and (added + updated) the tests but not really sure if that is the correct way to do that.
I am not sure if I am supposed to raise a pr if an issue is taken 😅 @bvaughn 
but tried to do this, even when taken, because this seemed a bit easy to do with the limited knowledge I have 😓 and could get me to familiarize myself with this code.
If you have not started work on it or not planning to pick soon, shall I create PR with what I have done?
Else if you have started, please feel free to ignore, but please tag me on the PR, I would love to see the solution.

brunogonzales:
@charlie1404 you can totally take this off my hands, I haven't started working on it yet. Please tag me on your PR as I would like to know the solution too.

charlie1404:
@bvaughn How to get reviewers on pr ?
Is that automated/queued or, need to comment/tag someone on PR ??

hemakshis:
Hi! Is this issue still open and available? I would be happy to take it up if no one is working on it ATM.
Thanks

eps1lon:
> Hi! Is this issue still open and available? I would be happy to take it up if no one is working on it ATM.
> Thanks

[...]

PR #19054, opened 2020-05-30
Added warning to <Context.Provider> in case no value prop is provided

charlie1404:
## Summary
This PR aims to add a fix for #19020 

1. There is a duplication in what I have done in files
    - `packages/react-reconciler/src/ReactFiberBeginWork.new.js`
    - `packages/react-reconciler/src/ReactFiberBeginWork.old.js`

In feature flag saw `const enableNewReconciler = false` but still cannot jump to any solid conclusions hence added in both.

Test Added
`packages/react-reconciler/src/__tests__/ReactNewContext-test.js` test to detect warning.

Tests Updated Reason
As now `<Context.Provider>` now throws a warning if value prop is not given, and hence updated the ones which were not having value prop to have `null` now.

Points to be discussed
1. What should be the warning message, (not really good with this) 😅 
2. Should this be included in both old and new, or I have added in the totally wrong place.
    - if correct, out of curiosity why there are 2 files, seems almost identical, and only these but a whole lot files are in the same format. (please ignore if very long explanation and out of the scope of this PR)
3. Should any more test cases to be added in this?

cc @brunogonzales @heath-freenome

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -2814,7 +2814,9 @@ function updateContextProvider(
     if (!('value' in newProps)) {
       if (!hasWarnedAboutUsingNoValuePropOnContextProvider) {
         hasWarnedAboutUsingNoValuePropOnContextProvider = true;
-        console.error('<Context.Provider> must have a value prop');
+        console.error(
+          'The prop `value` is required in `Context.Provider`, have you misspelled it',
+        );
       }
     }
     const providerPropTypes = workInProgress.type.propTypes;
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -2795,6 +2795,8 @@ function updatePortalComponent(
   return workInProgress.child;
 }
 
+let hasWarnedAboutUsingNoValuePropOnContextProvider = false;
+
 function updateContextProvider(
   current: Fiber | null,
   workInProgress: Fiber,
@@ -2809,6 +2811,12 @@ function updateContextProvider(
   const newValue = newProps.value;
 
   if (__DEV__) {
+    if (!('value' in newProps)) {
+      if (!hasWarnedAboutUsingNoValuePropOnContextProvider) {
+        hasWarnedAboutUsingNoValuePropOnContextProvider = true;
+        console.error('<Context.Provider> must have a value prop');
+      }
+    }
     const providerPropTypes = workInProgress.type.propTypes;
 
     if (providerPropTypes) {
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -2815,7 +2815,7 @@ function updateContextProvider(
       if (!hasWarnedAboutUsingNoValuePropOnContextProvider) {
         hasWarnedAboutUsingNoValuePropOnContextProvider = true;
         console.error(
-          'The prop `value` is required in `Context.Provider`, have you misspelled it',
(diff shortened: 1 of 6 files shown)


charlie1404:
did not like the name of the code sandbox 🤣.

heath-freenome:
Question, would making the `value` prop `.isRequired` also accomplish this?

charlie1404:
> Question, would making the `value` prop `.isRequired` also accomplish this?

@heath-freenome 
Initially gave a thought on that idea, but did not proceed for 2 reasons: 

1. If we use prop types there is a possibility that value prop can actually be `null` or `undefined`, say if the value is being set from some key in `state`, which is an object, but the key used is `null` or `undefined` for some reason, in that case, it will start adding warnings to console, not sure if we should do that, **worth discussing**. The `'value' in props` only gives a warning if there is no value prop supplied to `<Context.Provider>`.

2. Now if we use prop types I had to do `Object.assign` with `workInProgress.type.propTypes`, but as this is an internal component and had no prop types defined hence was undefined, not sure if I should add a completely new object, ignoring old one or still do Object.assign maybe I was looking at the wrong place.

But pointing to prop types gave me a better message, thanks for that, updating it now.

eps1lon:
React just recently removed all runtime dependencies to `prop-types`. It's unlikely they re-introduce it. Especially given `prop-types` cannot distinguish between a missing prop and an undefined prop which @charlie1404 already pointed out.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2809,6 +2811,14 @@ function updateContextProvider(
   const newValue = newProps.value;
 
   if (__DEV__) {
+    if (!('value' in newProps)) {
+      if (!hasWarnedAboutUsingNoValuePropOnContextProvider) {
+        hasWarnedAboutUsingNoValuePropOnContextProvider = true;
+        console.error(
+          'The prop `value` is required in `Context.Provider`, have you misspelled it',
```
gaearon, line 2818: Let's reword this a bit:

```
The `value` prop is required for the `<Context.Provider>`. Did you misspell it or forget to pass it?
```

packages/react-reconciler/src/ReactFiberBeginWork.old.js
```diff
@@ -2806,6 +2808,14 @@ function updateContextProvider(
   const newValue = newProps.value;
 
   if (__DEV__) {
+    if (!('value' in newProps)) {
+      if (!hasWarnedAboutUsingNoValuePropOnContextProvider) {
+        hasWarnedAboutUsingNoValuePropOnContextProvider = true;
+        console.error(
+          'The prop `value` is required in `Context.Provider`, have you misspelled it',
```
gaearon, line 2815: Same

packages/react/src/__tests__/ReactContextValidator-test.js
```diff
@@ -263,13 +263,13 @@ describe('ReactContextValidator', () => {
 
     class Component extends React.Component {
       render() {
-        return <TestContext.Provider />;
+        return <TestContext.Provider value={null} />;
```
gaearon, line 266: Make this `value={undefined}` instead and you won't need to change the message.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
charlie1404: Thanks for the message 😅 .

charlie1404: updated

packages/react/src/__tests__/ReactContextValidator-test.js
charlie1404: now using undefined and reverted message

gaearon:
Coolio, thanks.

Issue #162, opened 2013-07-05
Rename onChange to onSomethingNotInTheSpecAlready


zpao:
We talked about not breaking expectations for an event that already exists. I think of our options, `onValueChange` was the best.

PR #19055, opened 2020-05-31
Preventing a possible property of undefined error

MetinArslanturk:
It can throw "reaching nodeType of null" or "reaching nodeType of undefined" error. Added falsy check to prevent it.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMEventHandle.js
+++ b/packages/react-dom/src/client/ReactDOMEventHandle.js
@@ -163,7 +163,7 @@ export function createEventHandle(
         callback: null | ((SyntheticEvent<EventTarget>) => void),
       ): void {
         // Check if the target is a DOM element.
-        if ((target: any) && target.nodeType === ELEMENT_NODE) {
+        if (target && (target: any).nodeType === ELEMENT_NODE) {
           const targetElement = ((target: any): Element);
           // Check if the DOM element is managed by React.
           const targetFiber = getClosestInstanceFromNode(targetElement);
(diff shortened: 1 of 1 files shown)


packages/react-dom/src/client/ReactDOMEventHandle.js
```diff
@@ -163,7 +163,7 @@ export function createEventHandle(
         callback: null | ((SyntheticEvent<EventTarget>) => void),
       ): void {
         // Check if the target is a DOM element.
-        if ((target: any).nodeType === ELEMENT_NODE) {
+        if (target && (target: any).nodeType === ELEMENT_NODE) {
```
ashr81, line 166: As per the type mentioned at line [#162](https://github.com/facebook/react/pull/19055/files#diff-fc22adb8dcb54b24931084f898446dd2R162). It should never be `undefined`. I am just curious to know in which case you got `target` has undefined?

ArslanturkMetin: I meant null or undefined. As i saw it does not have null check, and null can pass type check?

ashr81: This will silence the real error. This function always expects as `target` argument. But in case it not found. it silently fails without any warning. Silencing an error at the library level is not a good idea. Let's see what React Core Team prefers.

MetinArslanturk: No, you are wrong. This is a conditional check operation which i added extra check. The condition check operation "should not fail". Just should not prove condition and continue the method body (So in here if it is a null or an undefined then okay it not matches condition just that. Move on. This is just another scenario on "not matched condition scenarios". This (condition check operations) should not give null pointer error in any core-library code.

gaearon:
Thanks, but you have not explained the case in which it would happen. We generally don't add null guards "just in case" — if there is null there, it is a bug, and that bug needs to be fixed at its source instead.

PR #19059, opened 2020-06-01
Remove synamic modern event system flag for FB

trueadm:
This makes the modern event system flag enabled statically for FB bundles.

### DIFF
--- a/scripts/rollup/modules.js
+++ b/scripts/rollup/modules.js
@@ -16,6 +16,7 @@ const importSideEffects = Object.freeze({
   'react-dom/server': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
   'react/jsx-dev-runtime': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
   'react-fetch/node': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
+  'react-dom': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
 });
 
 // Bundles exporting globals that other modules rely on.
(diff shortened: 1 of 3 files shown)


scripts/rollup/modules.js
```diff
@@ -16,6 +16,7 @@ const importSideEffects = Object.freeze({
   'react-dom/server': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
   'react/jsx-dev-runtime': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
   'react-fetch/node': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
+  'react-dom': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
 });
```
aap88-afk, line 20: Duplicate_string}delete%

scripts/rollup/modules.js
```diff
@@ -16,6 +16,7 @@ const importSideEffects = Object.freeze({
   'react-dom/server': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
   'react/jsx-dev-runtime': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
   'react-fetch/node': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
+  'react-dom': HAS_NO_SIDE_EFFECTS_ON_IMPORT,
```
sebmarkbage, line 19: Why is this new?

trueadm: We have a block of code in ReactTestUtils that is wrapped in the feature flag. Inside this block we reference an import from `react-dom`. When we remove the block, we end up with having a dangling import to `react-dom` at the top of the module without it being used. In turn, this causes Rollup to fail the build.

This will organically go away when we remove the flags entirely from OSS.

Issue #19061, opened 2020-06-01
Bug: react-hooks/exhaustive-deps not recognizing dependencies when accessed via optional chaining


fredvollmer:
This is, I believe, an extension of the issue reported in https://github.com/facebook/react/issues/18985.

While the PR that resolved that issue fixed the lint rule for the `useEffect` hook, it seems to remain an issue for `useCallback` and `useMemo`.

When the optional chaining operator (`?.`) is used to access a path inside a `useMemo` or `useCallback` hook, and a prefix of that path is listed in the dependencies array, eslint-plugin-react-hooks reports an unexpected "missing dependency" warning. Replacing `?.` with `. ` produces the expected behaviour: the warning is no longer reported.

In many cases, adding the complete path, rather than a prefix, to the dependencies array is the correct answer. However, in some cases, the complete path might point to a prototype method of an object, in which case only a prefix of the path should be listed. (Hopefully the example below makes this more clear.)

React version: 16.13.1
eslint-plugin-react-hooks version: 4.0.4

## Steps To Reproduce

```javascript
 import React, {useEffect} from 'react';

 export function MyComponent(props) {
   useCallback(() => {
     console.log(props.foo?.toString());
   }, [props.foo]);
    
    return null;
  }
```
- It doesn't matter what `foo`, represents, other than it has a prototype method, `toString()`.
- We don't want to reference the prototype method of `foo` (`props.foo?.toString`) in the dependencies array because our function depends on `props.foo`, not the prototype of `props.foo`.

## The current behavior
The following warning is reported:
```
React Hook useCallback has an unnecessary dependency: 'props.foo'. Either exclude it or remove the dependency array  react-hooks/exhaustive-deps
```
- Note that if I list `props.foo?.toString` instead of `props.foo` in the dependencies, no warning is reported. However, this is not acceptable as the callback depends on `props.foo`, not the prototype of `props.foo`.
- If I remove the optional chaining operator from the above code (so that the body of the callback reads `console.log(props.foo?.toString());`) then no warning is reported, as expected.

## The expected behavior
No warning should be reported.

darkbasic:
It used to work so I think it's a regression.

gaearon:
Fixed by https://github.com/facebook/react/pull/19062

PR #19062, opened 2020-06-01
fix(eslint-plugin-react-hooks): Support optional chaining when accessing prototype method inside useCallback and useMemo #19061

fredvollmer:
Relates to #19061
## Summary

This fix ensures that optional chaining can be used to access a prototype method of an object, inside a `useCallback` or `useEffect` hook. 

For example, assume we use optional chaining to call the `toString()` method of an object:
```
foo?.toString();
```

In such a case, `foo`, **not** `foo?.toString`, should be added to the dependencies array of the `useCallback` or `useMemo` hooks. However, doing so currently throws a false "unnecessary dependencies" warning.

## Test Plan

Additional tests were added to cover using the optional chaining operator in the `useCallback` and `useMemo` hooks, when only a prefix of the object path is provided (as it would be if a prototype method is being referenced in the callback.)

### DIFF
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -1287,12 +1287,13 @@ function collectRecommendations({
   function scanTreeRecursively(node, missingPaths, satisfyingPaths, keyToPath) {
     node.children.forEach((child, key) => {
       const path = keyToPath(key);
+      // For analyzing dependencies, we want the "normalized" path, without any optional chaining ("?.") operator
+      // foo?.bar -> foo.bar
+      const normalizedPath = path.replace(/\?$/, '');
       if (child.isSatisfiedRecursively) {
         if (child.hasRequiredNodesBelow) {
           // Remember this dep actually satisfied something.
-          // Here we only want to compare the "normalized" path, without any optional chaining ("?.") operator
-          // foo?.bar -> foo.bar
-          satisfyingPaths.add(path.replace(/\?$/, ''));
+          satisfyingPaths.add(normalizedPath);
         }
         // It doesn't matter if there's something deeper.
         // It would be transitively satisfied since we assume immutability.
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -1290,7 +1290,9 @@ function collectRecommendations({
       if (child.isSatisfiedRecursively) {
         if (child.hasRequiredNodesBelow) {
           // Remember this dep actually satisfied something.
-          satisfyingPaths.add(path);
+          // Here we only want to compare the "normalized" path, without any optional chaining ("?.") operator
+          // foo?.bar -> foo.bar
+          satisfyingPaths.add(path.replace(/\?$/, ''));
```
coolkev, line 1295: The same change should be made below when adding path to missingPaths (line 1304). Version 4.04 is now getting an eslint error:
ESLint stack trace:
[Error - 4:20:28 PM] TypeError: Cannot read property 'references' of undefined

Your same change to missingPaths fixes that error.
```
missingPaths.add(path.replace(/\?$/, ''));
```

fredvollmer: Great catch! Just made this change.

nathggns:
Do we know when this might get merged?

mdestagnol:
Hi! Looks like the change is approved. Is there a reason why it's not merged yet?

Thomazella:
Running into this error too.

coolkev:
@gaearon, any chance we can get this merged/deployed? It is breaking eslint completely for me when it occurs. The issue was introduced in PR #19008

gaearon:
Out in eslint-plugin-react-hooks@4.0.5

krailler:
On eslint-plugin-react-hooks@4.0.5 with ESLint 7.3.1 is not working for me... 

The error "TypeError: Cannot read property 'references' of undefined" appears.

Any thoughts?

gaearon:
File a new issue with code causing it.

gaearon:
Locking so we don't have to track discussion in a merged PR. If there is a new problem, please file a new issue.

gaearon:
Optional chaining problems should be fixed in `eslint-plugin-react-hooks@4.0.6`. If you experience some problem with optional chaining after 4.0.6, please file a new issue. Thanks.

PR #19069, opened 2020-06-03
Add console group methods to dev disabled groups

gffuma:
## Summary

Add `console.group` `console.groupEnd` and `console. groupCollapsed` to disable console group because have `console.log` disable while `console.group` is enabled lead to strange console output.


gaearon:
Would you mind rebasing? Thanks.

PR #19071, opened 2020-06-04
remove duplicate invoke

oft:
remove UNSAFE_[ lifecycle ] invoke when [ lifecycle ] is defined

### DIFF
--- a/packages/react-reconciler/src/ReactFiberClassComponent.new.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.new.js
@@ -711,7 +711,7 @@ function callComponentWillMount(workInProgress, instance) {
 
   if (typeof instance.componentWillMount === 'function') {
     instance.componentWillMount();
-  }else if (typeof instance.UNSAFE_componentWillMount === 'function') {
+  } else if (typeof instance.UNSAFE_componentWillMount === 'function') {
     instance.UNSAFE_componentWillMount();
   }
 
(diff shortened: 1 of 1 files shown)


packages/react-reconciler/src/ReactFiberClassComponent.new.js
```diff
@@ -711,8 +711,7 @@ function callComponentWillMount(workInProgress, instance) {
 
   if (typeof instance.componentWillMount === 'function') {
     instance.componentWillMount();
-  }
-  if (typeof instance.UNSAFE_componentWillMount === 'function') {
+  }else if (typeof instance.UNSAFE_componentWillMount === 'function') {
```
razh, line 714:
```suggestion
  } else if (typeof instance.UNSAFE_componentWillMount === 'function') {
```

To match the existing code style:

https://github.com/facebook/react/blob/4c7036e807fa18a3e21a5182983c7c0f05c5936e/packages/react-reconciler/src/ReactFiberClassComponent.new.js#L559-L560

---

This can potentially be applied to [`ReactFiberClassComponent.old.js`](https://github.com/facebook/react/blob/4c7036e807fa18a3e21a5182983c7c0f05c5936e/packages/react-reconciler/src/ReactFiberClassComponent.old.js#L764-L769) as well.

gaearon:
Thanks for the PR, but as mentioned in the above comment, this behavior is intentional. Additionally, it would be a breaking change.

Issue #14099, opened 2018-11-05
useCallback() invalidates too often in practice


gaearon:
This is related to https://github.com/facebook/react/issues/14092, https://github.com/facebook/react/issues/14066, and some other issues.

The problem is that we often want to avoid invalidating a callback (e.g. to preserve shallow equality below or to avoid re-subscriptions in the effects). But if it depends on props or state, it's likely it'll invalidate too often. See https://github.com/facebook/react/issues/14092#issuecomment-435907249 for current workarounds.

`useReducer` doesn't suffer from this because the reducer is evaluated directly in the render phase. @sebmarkbage had an idea about giving `useCallback` similar semantics but it'll likely require complex implementation work. Seems like we'd have to do _something_ like this though.

I'm filing this just to acknowledge the issue exists, and to track further work on this.

gokalina:
@gaearon thank you for your answer in reactjs/rfcs#83. I've look at sources of `useReducer`. But I can't understand how it is related to `useCallback`. What issues has "mutation of ref during rendering"? Can you explain me in brief?

gaearon:
> I've look at sources of useReducer. But I can't understand how it is related to useCallback

`useCallback` lets you memoize the callback to avoid a different function being passed down every time. But you have to specify everything it depends on in the second array argument. If it's something from props or state, your callback might get invalidated too often.

`useReducer` doesn't suffer from this issue. The `dispatch` function it gives you will stay the same between re-renders even if the reducer itself closes over props and state. This works because the reducer runs during the next render (and thus has natural ability to read props and state). It would be nice if `useCallback` could also do something like this but it's not clear how.

>What issues has "mutation of ref during rendering"? Can you explain me in brief?

In concurrent mode (not yet released), it would "remember" the last rendered version, which isn't great if we render different work-in-progress priorities. So it's not "async safe".

strayiker:
Would be nice if second argument of `useCallback` was injected as dependencies to callback function.

```javascript
  function useCallback(cb, deps) => {
    lastDeps = deps; // save current deps deep in somewhere

    if (!cached) {
      cached = (...args) => cb(...lastDeps)(...args); // memoize that forevere
    }

    return cached; // never invalidates
  }

  const myCallback = useCallback(
    (state, props) => (arg1, arg2) => {},
    [state, props]
  );

```

gaearon:
[...]

PR #19095, opened 2020-06-07
Add ESLint Rule for Safe Ref Usage

louisscruz:
## Summary

Mutating a given ref's `current` property as the result of a render or hook call is not safe in concurrent mode. The reason for this is described [here](https://github.com/facebook/react/issues/14099#issuecomment-440794435).

As an example, this is not safe:

```js
const MyComponent = React.memo({ someProp }) => {
  const myRef = useRef();

  myRef.current = someProp;

  // ...
});
```

This also wouldn't be safe:

```js
const useMyCustomHook = () => {
  const [someState, setSomeState] = useState('');
  const myRef = useRef();

  myRef.current = someState;

  // ...
};
```

There currently isn't a guard rail in place to prevent applications not written for concurrent mode first to easily identify this problem before the application is opted-in to concurrent mode. Without a guard rail, applications that move from non-concurrent mode to concurrent mode are at some point are at a serious risk of experiencing cumbersome regressions as a result of the move.

So, I'm proposing an ESLint rule that fails in the case that any top-level ref mutations are found and suggests that they be moved into either `useEffect` or `useLayoutEffect` hook calls so that way the mutation only occurs in the case that a commit to render has happened.

At the time that this PR is opened, the rule is called `safe-ref-mutations`. It is initially not marked as recommended because I'd like to get some core React contributor guidance first.

This rule didn't seem to fit in with either `rules-of-hooks` or `exhaustive-deps`, so I wrote it as its own rule. If there's a preference to move it into one or the other, I'm happy to do so. But in general, I wonder if there are more rules that React should have in place for safe concurrency mode usage. Maybe some kind of `safe-concurrency` rule would be better?

As part of this PR, I needed some functions that accomplished the same as those in `RulesOfHooks.js`, so I've extracted them into a `utils.js` for re-use.

If there's a preference that I open an issue for this, I'm happy to do so.

## Test Plan

The ESLint rule will be heavily tested if this is the route that is chosen.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintSafeRefMutations-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintSafeRefMutations-test.js
@@ -233,8 +233,7 @@ const tests = {
       }
     `,
     // When ref mutations occur in callbacks
-    {
-      code: normalizeIndent`
+    normalizeIndent`
       const ComponentWithRef = () => {
         const someRef = useRef();
 
@@ -245,10 +244,19 @@ const tests = {
         return <div onClick={someCallback}>some child</div>;
       };
     `,
-    },
-    // TODO: Scenarios where React is no in play at all
+    // Scenarios where React is not in play at all
     normalizeIndent`
       myObject.current = 'some value';
+    `,
+    normalizeIndent`
+      function someNonComponentFunction() {
+        someObject.current = 'some value';
+      }
+    `,
+    normalizeIndent`
+      const someNonComponentFunction = () => {
+        someObject.current = 'some value';
+      }
     `,
   ],
   invalid: [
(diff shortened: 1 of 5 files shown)


EECOLOR:
@louisscruz You point at a comment talking about a very specific use case where a ref is used inside a `callback` that is returned from a hook. This allows that code to be used anywhere in the render tree.

With the examples you provide, you simply state: 'this is not safe'. Could you explain how these constructs could cause problems?

louisscruz:
@EECOLOR https://github.com/facebook/react/issues/14099#issuecomment-440794435

If one branch’s rendering is interrupted, but its rendering has the side effect of mutating a ref, that ref value will still have been mutated and retain the value intended for that branch rendering in other branch renderings.

EECOLOR:
@louisscruz Thank you for responding. It however does not help me understanding the problematic nature. How can we achieve the same instance of a component to end up in two branches?

Issue #16474, opened 2019-08-19
DevTools: Named hooks


bvaughn:
**Note** this issue is outdated. The current thinking is that the alternative, "*load source code (with source maps) and parse for name*", is probably the best course of action.

---

## The problem

One common piece of feedback about DevTools hooks integration is that hooks have no name and can be confusing. Consider the following example:
```js
function useSomeCustomHook() {
  const [foo, setFoo] = useState(true);
  const [bar, setBar] = useState(false);

  // ...
}

function Example() {
  const baz = useSomeCustomHook();

  // ...
}
```

Currently in DevTools the above component would be displayed as follows:
```
SomeCustomHook:
  State: true
  State: false
```

This information isn't as rich as we would prefer. ☹️

The next question is often: "can you use the name of the variable the hook return value is assigned to?" but this is tricky because DevTools doesn't actually have any way to access that variable. (Even if DevTools has a handle on the `Example` function above, how would it access the `useSomeCustomHook` function?)

## The proposal

The solution to this would be some form of user-defined metadata (preferably generated by a code transform). Building on the precedent of the `useDebugValue` hook (https://github.com/facebook/react/pull/14559), we might introduce a new no-op hook e.g. `useDebugName`.

The above example could make use of this hook like so:
```js
function useSomeCustomHook() {
  const [foo, setFoo] = useState(true);
  useDebugName("foo"); // injected by Babel transform
  const [bar, setBar] = useState(false);
  useDebugName("bar"); // injected by Babel transform

  // ...
}

function Example() {
  const baz = useSomeCustomHook();

  // ...
}
```

DevTools could then display something like:
```
SomeCustomHook:
  State (foo): true
  State (bar): true
```

## Implementation details

The new `useDebugName` hook might be a noop hook provided by React (similar to `useDebugValue`) _or_ it could even be an export from the (soon to be released [`react-debug-hooks` package](https://github.com/facebook/react/blob/master/packages/react-debug-tools/)). The key concerns would be that:
1. It has no effect (and adds no overhead) when DevTools is not present.
2. Not calling it at all (or only calling it for _some_ hooks) should not break or corrupt anything.

DevTools could override the no-op `useDebugName` implementation before inspecting a component and automatically associate the provided name with the most recently called native hook.

For example, the following code should only result in one named hook (the second `useState` call).
```js
const [foo, setFoo] = useState(true);
const [bar, setBar] = useState(false);
useDebugName("bar"); // injected by Babel transform
[...]

PR #19097, opened 2020-06-08
[DevTools] Named hooks

bl00mber:
## #16474
This is the implementation of main proposed solution based on babel DEV-only AST transform, supporting built-in hooks.
If name is the default hook name (i.e. state for setState, reducer for useReducer), the hook is not injected by transform plugin and ignored by devtools.
Three methods of hook name extraction are used: first array item, identifier name, first argument name.

Supports:
* useState - first array item
* useReducer - first array item
* useRef - identifier name
* useCallback - identifier name
* useMemo - identifier name
* useContext - first argument name

## Before
named-hooks-before

## After
named-hooks-after


gaearon:
Sorry we haven't gotten around to reviewing yet. This will be important to get back to but we haven't prioritized it yet.

eps1lon:
Bump

bl00mber:
@bvaughn

bvaughn:
I think this is probably more of the direction we'll be heading with this feature, once we have the bandwidth to land the change and test it and roll it out:
https://github.com/MLH-Fellowship/react/pull/115

bl00mber:
@bvaughn thanks for the review.

bvaughn:
Thank you for this PR!

We ended up going with another solution for this: #21641

PR #19102, opened 2020-06-09
use allowlist instead of whitelist

shengxinjing:
## Test Plan

use allowlist instead of whitelist for some reason


pingfengafei:
**I REALLY DISAGREE WITH THIS PR**

I worry about the racial discrimination have happened in US from long time ago to recent as a human being, while I also insist to regard `blacklist, whitelist, blackbox testing, whitebox testing, master, slave, etc` to be neutral nouns in `IT area` as a software engineer. They just info to opposite or relevant methods in computer science without any emotion. The more one tries to hide, the more one is exposed.

My voice may seem to be sharp on this rename PR. Indeed, that shows my attitude about racial discrimination and mercy about people who suffer from brutality all around the world.   As a Chinese, we know the eval history of `Negro Slaves` and experienced `Literary Inquisition` about 400 years ago. We should show really action instead of showing shows.

shengxinjing:
it's for fun  beacuse google use replace blacklist with blocklist in chrome and [go](https://github.com/golang/go/blob/6bf2eea62a3425c57f3d908ec32047a9ae41c025/src/cmd/compile/internal/gc/plive.go#L905)

vsTianhao:
Agree @pingfengafei .
Don't let word game anesthetize us, If everyone pays attention, discrimination will slowly disappear, but if we pretend that it has disappeared now, it may exist forever, The more one tries to hide, the more one is exposed.

acdlite:
Thanks for the PR! Landed in 655affa302437208e6f03c9ca6d170ea1707ace3

PR #19108, opened 2020-06-09
Land Lanes implementation in old fork

acdlite:
Replaces the old reconciler fork with the contents of the new reconciler fork.

There are still separate "old" and "new" forks, but they are now identical. We will continue to use the fork infra to land risky changes and refactors.

Most of the changes were automated using a new command, `yarn replace-fork`. I put the manual changes (removing dead branches, cleaning up tests, etc) into a separate commit.


scripts/eslint-rules/no-cross-fork-imports.js
```diff
@@ -17,38 +17,67 @@ function isNewFork(filename) {
...
+module.exports = {
+  meta: {
+    type: 'problem',
+
+    // docs: {
+    //     description: "disallow unnecessary semicolons",
+    //     category: "Possible Errors",
+    //     recommended: true,
+    //     url: "https://eslint.org/docs/rules/no-extra-semi"
+    // },
```
acdlite, line 29: Oops, copypasta. Will delete.

packages/react-devtools-shared/src/backend/renderer.js
```diff
@@ -2160,7 +2160,7 @@ export function attach(
     const dependencies =
       (fiber: any).dependencies ||
       (fiber: any).dependencies_old ||
-      (fiber: any).dependencies_new;
+      (fiber: any).dependencies;
```
sophiebits, line 2163: this statement doesn't look quite right

acdlite:
@bvaughn Some of the DevTools tests were broken. Only noticed now because DevTools tests only run against the old fork. The failures were related to the layout of the Suspense fiber, since the primary tree is now always wrapped in an extra fiber. I was able to fix most of them with a small change.

There is one remaining test failure that I couldn't figure out. It's related to the same thing, but since I'm not familiar with that part of the DevTools code, I think if I tried to change anything else I'd end up breaking it in some other way. Since it's already been running in www for weeks, I figure it's OK to land as-is and fix it forward, once you're back from PTO and can give me some pointers.

bvaughn:
Ooh, we should get DevTools tests running against both reconcilers I guess. I wonder how hard that would be. Are we building the new reconciler into any artifacts (e.g. experimental?

bvaughn:
Is this the error you were seeing?
> Error: Uncaught [Error: Expected to find a host parent. This error is likely caused by a bug in React. Please file an issue.]

bvaughn:
I'm still a little baffled by this error. Through a bit of bisecting, I was able to suppress the bizarre `console.error` call by disabling the `_adjustParentTreeWeight` call inside of the `TREE_OPERATION_ADD` case in `onBridgeOperations` but I'm not yet sure why this would have anything to do with anything.

bvaughn:
Okay. Coming back at this today with a fresh perspective.

The test has a component called `Never`:
```js
const Never = () => {
  throw new Promise(() => {});
};
```

I think what's happening is this:
1. The `Never` component is rendered, but never completes (because it throws a Promise that never resolves).
1. The `store-test` tells DevTools to force `Never`'s parent `Suspense` boundary to *un-suspend*- so React tags it with a `DidCapture` effect.
1. React later tries to commit a deletion for the `Never` fiber. However it was already deleted, as part of the previous render, so the `return` pointer is `null` and React throws.

Step 3 is different behavior between the old and new reconciler implementations. (The old one never tried to commit a deletion for the `Never` fiber after the forced suspend step.)

So why does the new one? Stepping through the code, the force suspend doesn't actually tag `Never` (or anything else) with a `Deletion` effect. It looks like this effect is actually *left over* from the previous commit.

The failing test case can be reduced to these three steps:
```js
act(() =>
  ReactDOM.render(
    <Wrapper
      suspendParent={false}
      suspendFirst={true}
      suspendSecond={true}
    />,
    container
  )
);

// This update deletes the Never component rendered by the previous commit:
act(() =>
  ReactDOM.render(
    <Wrapper
      suspendParent={false}
      suspendFirst={false}
      suspendSecond={false}
    />,
    container
  )
);

// This update incorrectly tries to commit a second deletion for Never:
const rendererID = getRendererID();
act(() =>
  agent.overrideSuspense({
    id: store.getElementIDAtIndex(4),
    rendererID,
    forceFallback: true,
  })
);
```

bvaughn:
While debugging this, I also noticed that running tests with the new `backend/renderer` (after your changes in this PR) against an older React build fails- which concerns me. DevTools should support older versions of React. Although *maybe* in this case the version it fails for was never actually released and so it would not be a real problem...

packages/react-reconciler/src/ReactFiberBeginWork.old.js
```diff
@@ -1920,290 +1840,382 @@ function updateSuspenseComponent(
...
       }
     } else {
-      // The current tree has not already timed out. That means the primary
-      // children are not wrapped in a fragment fiber.
-      const currentPrimaryChild = current.child;
-      if (nextDidTimeout) {
-        // Timed out. Wrap the children in a fragment fiber to keep them
-        // separate from the fallback children.
+      // The current tree is not already showing a fallback.
+      if (showFallback) {
```
bvaughn, line 1899: @acdlite This branch looks weird. The larger structure is:

```js
if (showFallback) {
  // ...
} else {
  if (showFallback) {
    // ...
  } else {
    // ...
  }
}
```

acdlite:
> Although maybe in this case the version it fails for was never actually released and so it would not be a real problem... we need to run the test suite against 16.13.1 at some point and make sure it doesn't break for it.

Ooh I did remove the check for `dependencies_new` and `dependencies_old`. Can't recall if that was ever released.

acdlite:
This ended up being the crux of the issue:

> I think the problem is that the call to updateSuspenseFallbackChildren is putting Never back into the effect list after step 3. In the old reconciler, I think the equivalent section of code cloned the children (which reset their effects).

See #19164 for the fix. Thanks for narrowing that down!

acdlite:
> Ooh, we should get DevTools tests running against both reconcilers I guess. I wonder how hard that would be. Are we building the new reconciler into any artifacts (e.g. experimental)?

The `ReactDOMForked-*` www artifacts use the new reconciler.

PR #19115, opened 2020-06-10
feat(ReactDOMServer): add renderToBrowserStream

OverGlass:
And renderToStaticBrowserStream. Using ReadableStream browser API

## Summary

Hello :)
This PR add possibility to use `ReactDOM.renderToNodeStream()` inside the **browser** or in **Deno.land** with `ReactDOM.renderToBrowserStream()` & `ReactDOM.renderToStaticBrowserStream()`method.

It use browser API [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)

If you don't want this feature, can we have access to ReactPartialRenderer to be able to do that ?
It's my first PR. Sorry if I waste your time here !

Thank you !

## Test Plan
I tried to do some tests, but I don't know how to run browser api test in node Environment.
I'm motivate to do that but, I need a little help here :)
**EDIT** : I figured out how handle this issue
```javascript
  describe('renderToStaticBrowserStream', () => {
    it('should generate simple markup', async () => {
      const SuccessfulElement = React.createElement(() => <img />);
      const stream = ReactDOMServerBrowser.renderToStaticBrowserStream(
        SuccessfulElement,
      );
      const reader = stream.getReader();
      const string = (await reader.read()).value;
      expect(string).toMatch(new RegExp('<img' + '/>'));
    });

    it('should handle errors correctly', () => {
      const FailingElement = React.createElement(() => {
        throw new Error('An Error');
      });
      const stream = ReactDOMServerBrowser.renderToStaticBrowserStream(
        FailingElement,
      );
      const response = stream.getReader();
      let handleError = false;
      response
        .read()
        .catch(e => {
          handleError = true;
        })
        .finally(() => {
          expect(handleError).toBeTruthy();
        });
    });
  });
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "through2": "^3.0.1",
     "tmp": "^0.1.0",
     "typescript": "^3.7.5",
+    "web-streams-polyfill": "^2.1.1",
     "webpack": "^4.41.2"
   },
   "devEngines": {
--- a/packages/react-dom/server.browser.js
+++ b/packages/react-dom/server.browser.js
@@ -12,5 +12,7 @@ export {
   renderToStaticMarkup,
   renderToNodeStream,
   renderToStaticNodeStream,
+  renderToBrowserStream,
+  renderToStaticBrowserStream,
   version,
 } from './src/server/ReactDOMServerBrowser';
--- a/packages/react-dom/server.node.js
+++ b/packages/react-dom/server.node.js
@@ -13,5 +13,7 @@ export {
   renderToStaticMarkup,
   renderToNodeStream,
   renderToStaticNodeStream,
+  renderToBrowserStream,
+  renderToStaticBrowserStream,
   version,
 } from './src/server/ReactDOMServerNode';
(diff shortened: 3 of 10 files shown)


OverGlass:
Hello :)
I add testing, I figured out how to handle ReadableStream with Jest.
The testings are similar to [ReactServerRendering-test](https://github.com/facebook/react/blob/4c7036e807fa18a3e21a5182983c7c0f05c5936e/packages/react-dom/src/__tests__/ReactServerRendering-test.js#L579)

gaearon:
This is something we're supporting in the new server renderer infra (not yet actively worked on) but I'm not sure it's valuable to bring it over to the old one.

OverGlass:
I wasn’t aware of this new server renderer infra..! If this is something you're supporting in infra, I don't think it's valuable to bring it over to the old one indeed. 
Thank you for your time 🙏

gaearon:
To be clear, the new one is way off. It's hard to say when it replaces the old one. But it also doesn't seem like this particular feature is widely used anyway so it seems like it could wait.

OverGlass:
To be honest I just want to be able doing Progressive Server Side Rendering with Deno. I don't know if there is lot of people that want to do that today. So we can just wait for infra 😉

PR #19121, opened 2020-06-12
Decouple update priority tracking from Scheduler package

rickhanlonii:
## Summary

We'd like to remove a dependency on the Scheduler priority from React internals by replacing it with a currentLanePriority which will be tracked inside React. This PR starts currentLanePriority implementation by setting it everywhere that the scheduler priority is set, and then logging mismatches as they occur.

The next step is to investigate mismatches and handle them appropriately.


packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1498,25 +1503,43 @@ function rerenderDeferredValue<T>(
...
-      } finally {
-        ReactCurrentBatchConfig.suspense = previousConfig;
-      }
-    },
-  );
+  const previousLanePriority = getCurrentLanePriority();
+  try {
+    setCurrentLanePriority(
+      priorityLevel < UserBlockingPriority
+        ? InputContinuousLanePriority
```
rickhanlonii, line 1510: Not sure what lane priority this should be.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -417,6 +421,12 @@ export function requestUpdateLane(
       currentEventWipLanes,
       currentEventPendingLanes,
     );
+  } else if (
+    enableCurrentLanePriority &&
+    getCurrentLanePriority() !== NoLanePriority
+  ) {
```
rickhanlonii, line 427: I'll add a comment here before finishing the draft.

sebmarkbage:
For perspective the way I look at setState is that in an optimal reactive system it's essentially this:

```
setState = (reducer) => {
  inst.nextState = reducer(inst.nextState);
  if (!inst.isDirty) {
    inst.isDirty = true;
    dirtyInstances.push(inst);
  }
}
```

The goal is to get close to this but we're _very_ far away from this for a simple state change atm. I think we might need to do something more radical.

acdlite:
> Do we have a plan for what we want to do with 1) execution context, 2) runWithPriority and 3) how we can avoid special cases like priorityLevel < UserBlockingPriority all over the code base?

This PR is meant to address fix 2 and 3. It doesn't in the current state but that's one of the goals.

acdlite:
I think we agree on the final end state. The goals for this first step are a bit confusing and contradictory so I'll list them here:

- The main goal is to decouple the Scheduler priority from React update priority. It should be possible for us to control the priority of updates without going through Scheduler. This helps not only with coupling, but also means (in the future) we can avoid the overhead of `runWithPriority`. All of our APIs that change the priority already have a try/finally block; we can mutate the current update priority directly without a wrapper function.
- We can't remove `runWithPriority` entirely without a semantic change. So in this first step, we'll keep all the existing `runWithPriority` calls. We'll start removing them behind the feature flag Rick has added (could maybe move that to a separate PR).
- Similarly, we need to keep the `getCurrentPriorityLevel` check at the end of the `requestUpdateLane` algorithm so that outside `runWithPriorityLevel` calls still work.

Once the decoupling is complete, then we can move on to actually removing the extra wrappers. This will require some semantic changes and we'll need to audit www for existing uses.

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1498,25 +1503,43 @@ function rerenderDeferredValue<T>(
...
+        try {
+          setPending(false);
+          callback();
+        } finally {
+          ReactCurrentBatchConfig.suspense = previousConfig;
+        }
+      },
+    );
+  } finally {
+    setCurrentLanePriority(previousLanePriority);
```
acdlite, line 1541: We can merge this with the `finally` block above. `callback` is the only thing that is expected to potentially throw.

packages/react-reconciler/src/ReactFiberHooks.old.js
```diff
@@ -1498,25 +1503,43 @@ function rerenderDeferredValue<T>(
...
+        : priorityLevel,
+      () => {
+        setPending(true);
+      },
+    );
+
+    setCurrentLanePriority(
+      priorityLevel > NormalPriority
+        ? DefaultLanePriority
+        : schedulerPriorityToLanePriority(priorityLevel),
```
acdlite, line 1525: This should be one of the transition priorities. Short or long, depending on the timeout. Here's where that logic currently exists: https://github.com/facebook/react/blob/8f05f2bd6d131a39835d468622e248b231ccbf8e/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L402-L406

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -120,6 +120,16 @@ export const OffscreenLane: Lane = /*                   */ 0b1000000000000000000
 
 export const NoTimestamp = -1;
 
+let currentLanePriority: LanePriority = NoLanePriority;
```
acdlite, line 123: Let's call this `currentUpdatePriority`  or `currentUpdateLanePriority`

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1777,10 +1802,16 @@ function resetChildLanes(completedWork: Fiber) {
 
 function commitRoot(root) {
   const renderPriorityLevel = getCurrentPriorityLevel();
-  runWithPriority(
-    ImmediateSchedulerPriority,
-    commitRootImpl.bind(null, root, renderPriorityLevel),
-  );
+  const previousLanePriority = getCurrentLanePriority();
+  try {
+    setCurrentLanePriority(SyncLanePriority);
```
acdlite, line 1807: Move this into `commitRoot`. There's only one place that throws, at the very end, so the try/finally isn't needed.

acdlite:
This is a really great first React PR!

acdlite:
Sorry, I guess I mean first reconciler PR. I know it's not your first React PR ever :D

packages/shared/ReactFeatureFlags.js
```diff
@@ -123,3 +123,6 @@ export const enableLegacyFBSupport = false;
 // interleaved event. Remove this flag once we have migrated to the
 // new behavior.
 export const deferRenderPhaseUpdateToNextBatch = true;
+
+// Replacement for runWithPriority in React internals.
+export const enableCurrentLanePriority = false;
```
acdlite, line 128: You can put this ReactFeatureFlags.www-dynamic so that tests run against both values:

https://github.com/facebook/react/blob/8f05f2bd6d131a39835d468622e248b231ccbf8e/packages/shared/forks/ReactFeatureFlags.www-dynamic.js#L10-L20

packages/react-reconciler/src/ReactFiberHooks.old.js
rickhanlonii: Should the priority on line 1509 also use the short/long lane priorities or are those fine?

acdlite: That one should always be InputContinuousLanePriority.

rickhanlonii: Are you sure about this? The transition lanes don't have an update lane: 

https://github.com/rickhanlonii/react/blob/0993d6dad6303e13098a3bf4304332ad2f67b546/packages/react-reconciler/src/ReactFiberLane.js#L538

rickhanlonii: Changing to the transition lane fails a couple tests for: 

Screen Shot 2020-06-12 at 3 34 47 PM

packages/shared/ReactFeatureFlags.js
```diff
@@ -123,3 +123,6 @@ export const enableLegacyFBSupport = false;
 // interleaved event. Remove this flag once we have migrated to the
 // new behavior.
 export const deferRenderPhaseUpdateToNextBatch = true;
+
+// Replacement for runWithPriority in React internals.
+export const enableCurrentUpdateLanePriority = true;
```
rickhanlonii, line 128: TODO: disable this before landing, enabling it to see the test failures in CI

packages/react-reconciler/src/ReactFiberHooks.old.js
acdlite: See the comment on line 538. Since this is a transition, there should be a SuspenseConfig object, and so `requestUpdateLane` should call `findTransitionLane` instead of `findUpdateLane`:

https://github.com/facebook/react/blob/103ed08c46198d01119ef35c37d78c6bc89705db/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L398-L419

There might be a better way to factor that, but since we currently still need the SuspenseConfig for timeouts, anyway, this should work for now.

packages/shared/ReactFeatureFlags.js
acdlite: You can run `yarn test-www-variant` to run the tests with the flag on.

acdlite: (We have too many test commands and combinations, sorry it's so confusing. I want to replace it with a single command with multiple arguments, like `yarn test --variant --release_channel=FB_WWW`.)

packages/react-reconciler/src/ReactFiberHooks.old.js
rickhanlonii: The tests are failing because there's no SuspenseConfig object. How would you expect the SuspenseConfig to be set (e.g. for [this test](https://github.com/facebook/react/blob/fe7163e73dadceda2655736d97cdd745d7abc8ea/packages/react-reconciler/src/__tests__/ReactTransition-test.js#L62))?

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1498,12 +1503,22 @@ function rerenderDeferredValue<T>(
 
 function startTransition(setPending, config, callback) {
   const priorityLevel = getCurrentPriorityLevel();
+  const previousLanePriority = getCurrentUpdateLanePriority();
+  setCurrentUpdateLanePriority(InputContinuousLanePriority);
```
acdlite, line 1507: Sorry, I misspoke. If the previous priority is higher than InputContinuousLanePriority, then we shouldn't lower it.

packages/react-reconciler/src/__tests__/ReactIncrementalUpdates-test.js
```diff
@@ -474,11 +481,17 @@ describe('ReactIncrementalUpdates', () => {
 
     await ReactNoop.act(async () => {
       pushToLog('A');
-      Scheduler.unstable_runWithPriority(
+
+      // TODO: Double wrapping is temporary while we remove Scheduler runWithPriority.
+      ReactDOM.unstable_runWithPriority(
```
rickhanlonii, line 486: @acdlite I don't think I can use ReactDOM here because the tests fail, but I also don't see how I can add it to the Noop. Any tips to fixing this last test?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -412,28 +415,76 @@ export function requestUpdateLane(
...
-    if (
-      // TODO: Temporary. We're removing the concept of discrete updates.
-      (executionContext & DiscreteEventContext) !== NoContext &&
-      schedulerPriority === UserBlockingSchedulerPriority
-    ) {
-      lane = findUpdateLane(InputDiscreteLanePriority, currentEventWipLanes);
-    } else {
-      const lanePriority = schedulerPriorityToLanePriority(schedulerPriority);
-      lane = findUpdateLane(lanePriority, currentEventWipLanes);
+  // TODO: Remove this dependency on the Scheduler priority.
```
rickhanlonii, line 425: I refactored this method again so that the feature flag only opts into the new behavior if the scheduler priority matches the update lane priority. If it doesn't match, we log the mismatch and fall back to the old behavior.

This will allow us to fix all of the warnings incrementally, while also turning on the feature flag to test the experiment, instead of needing to fix all the issues before testing.

Let me know what you think 👍

packages/react-dom/src/events/ReactDOMEventReplaying.js
```diff
@@ -540,16 +558,35 @@ function attemptExplicitHydrationTarget(
...
+    const updateLanePriority = higherLanePriority(
+      getCurrentUpdateLanePriority(),
+      DefaultLanePriority,
+    );
     const queuedTarget: QueuedHydrationTarget = {
       blockedOn: null,
       target: target,
-      priority: priority,
+      priority: schedulerPriority,
+      lanePriority: updateLanePriority,
```
rickhanlonii, line 570: @sebmarkbage can you review the changes to hydration I made in this file? I'm adding the update lane priority with the goal of removing the scheduler priority in a future diff after we confirm update lane priority works as expected.

packages/react-reconciler/src/__tests__/ReactIncrementalUpdates-test.js
rickhanlonii: There are two pieces of context I'm missing:

- What is the rebasing behavior that this test is testing for?
- Is it ok for the Noop renderer to import ReactFiberLanes to implement a runWithPriority like ReactDOM has? I.e. what are the boundaries we want to enforce for the Noops renderer?

packages/react-dom/src/client/ReactDOM.js
```diff
@@ -71,6 +71,12 @@ import {
   enqueueStateRestore,
   restoreStateIfNeeded,
 } from '../events/ReactDOMControlledComponent';
+import type {ReactPriorityLevel} from 'react-reconciler/src/ReactInternalTypes';
+import {
+  getCurrentUpdateLanePriority,
+  schedulerPriorityToLanePriority,
+  setCurrentUpdateLanePriority,
```
acdlite, line 78: These should be imported from `ReactFiberReconciler` instead. Typically, renderers should never import methods from deep inside the `react-reconciler` package. Only from the entry point. The main exceptions are types or constants, but even those we should try not to leak to the renderers.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -412,28 +415,76 @@ export function requestUpdateLane(
...
+  // To do that, we're replacing it with an update lane priority.
+  const schedulerPriority = getCurrentPriorityLevel();
+
+  if (enableCurrentUpdateLanePriority) {
```
acdlite, line 429: This flag changes the semantics. I don't think we should do that yet. I think the only thing it should do is fire the warning. Then, once we fix all the warnings, we can enable the semantic change and be confident it won't change any behavior.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
acdlite: I think the only thing it should do is fire the warning. The goal of this first step is that no behavior should change. We're just shuffling around an implementation detail.

acdlite: If the flag changes the semantics, then we can't turn it on safely in www, which goes against the purpose of using the warning to find the places that need to change.

packages/react-dom/src/events/ReactDOMEventReplaying.js
```diff
@@ -540,16 +558,35 @@ function attemptExplicitHydrationTarget(
...
     };
     let i = 0;
-    for (; i < queuedExplicitHydrationTargets.length; i++) {
-      if (priority <= queuedExplicitHydrationTargets[i].priority) {
-        break;
+    if (enableCurrentUpdateLanePriority) {
```
acdlite, line 573: Same note as here: https://github.com/facebook/react/pull/19121/files/c533cb3335f4728f798d1b7fc2b7ad0847dfd02d#r445670773

The flag should only enable the warning. Shouldn't affect any behavior.

packages/shared/ReactFeatureFlags.js
```diff
@@ -123,3 +123,6 @@ export const enableLegacyFBSupport = false;
 // interleaved event. Remove this flag once we have migrated to the
 // new behavior.
 export const deferRenderPhaseUpdateToNextBatch = true;
+
+// Replacement for runWithPriority in React internals.
+export const enableCurrentUpdateLanePriority = false;
```
acdlite, line 128: Let's name this something higher level, like `decoupleUpdatePriorityFromScheduler`.

packages/react-reconciler/src/__tests__/ReactIncrementalUpdates-test.js
acdlite: - "Rebasing" occurs when a high pri update interrupts a lower priority update. React will commit the high pri update first. Then, when it renders the lower priority update, it needs to maintain the original order. So it replays the high pri update on top of the low pri update. That way the final order in which the updates are processed is the same as order that they were dispatched.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
rickhanlonii: Sounds good - should the warning be behind the feature flag?

rickhanlonii: I got the answer below, thanks 👍

rickhanlonii:
Published for review, this should be good to go now 👍

packages/react-dom/src/events/ReactDOMEventReplaying.js
```diff
@@ -119,6 +119,13 @@ import {IS_REPLAYED, PLUGIN_EVENT_SYSTEM} from './EventSystemFlags';
 import {legacyListenToTopLevelEvent} from './DOMLegacyEventPluginSystem';
 import {listenToTopLevelEvent} from './DOMModernPluginEventSystem';
 import {addResponderEventSystemEvent} from './DeprecatedDOMEventResponderSystem';
+import {
+  DefaultLanePriority,
+  getCurrentUpdateLanePriority,
+  higherLanePriority,
+  setCurrentUpdateLanePriority,
+} from 'react-reconciler/src/ReactFiberLane';
```
acdlite, line 127: Shouldn't import stuff from `src/`

packages/react-dom/src/events/ReactDOMEventReplaying.js
```diff
@@ -540,15 +557,20 @@ function attemptExplicitHydrationTarget(
 
 export function queueExplicitHydrationTarget(target: Node): void {
   if (enableSelectiveHydration) {
-    const priority = getCurrentPriorityLevel();
+    const schedulerPriority = getCurrentPriorityLevel();
+    const updateLanePriority = higherLanePriority(
```
acdlite, line 561: What's this part for?

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1498,12 +1504,24 @@ function rerenderDeferredValue<T>(
...
+    higherLanePriority(previousLanePriority, InputContinuousLanePriority),
+  );
   runWithPriority(
     priorityLevel < UserBlockingPriority ? UserBlockingPriority : priorityLevel,
     () => {
       setPending(true);
     },
   );
+
+  setCurrentUpdateLanePriority(
```
acdlite, line 1518: This doesn't actually do anything, right? Since in `requestUpdateLane`, it will hit the SuspenseConfig branch and early return.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -412,28 +415,57 @@ export function requestUpdateLane(
...
+  // To do that, we're replacing it with an update lane priority.
+  const schedulerPriority = getCurrentPriorityLevel();
+
+  if (decoupleUpdatePriorityFromScheduler) {
+    // In the new strategy, we will track the current update lane priority
+    // inside React and use that priority to select a lane for this update.
+    // For now, we're just logging when they're different so we can assess.
+    const updateLaneSchedulerPriority = lanePriorityToSchedulerPriority(
+      getCurrentUpdateLanePriority(),
+    );
```
acdlite, line 435: I think we're still on different pages regarding what this warning is supposed to do.

- `findUpdateLane` returns an update lane.
- In the existing algorithm, it reads the current Scheduler priority and converts it to a LanePriority.
- In the future algorithm, it will instead read different priority field that is tracked by React. This field is already a LanePriority, so no conversion is needed.
- The warning should fire if the two algorithms produce different results — i.e., if they result in different lane priorities.

So the warning should compare the LanePriority for the update, not the Scheduler priority.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -412,28 +415,57 @@ export function requestUpdateLane(
...
+  // which is probably not what we want.
+  let lane;
+  if (
+    // TODO: Temporary. We're removing the concept of discrete updates.
+    (executionContext & DiscreteEventContext) !== NoContext &&
+    schedulerPriority === UserBlockingSchedulerPriority
+  ) {
+    lane = findUpdateLane(InputDiscreteLanePriority, currentEventWipLanes);
+  } else {
+    const lanePriority = schedulerPriorityToLanePriority(schedulerPriority);
```
acdlite, line 465: This is the block where the warning should go.

packages/react-dom/src/events/ReactDOMEventReplaying.js
rickhanlonii: This is done a few times in this file already, since it's temporary is it fine for now?

packages/react-dom/src/events/ReactDOMEventReplaying.js
rickhanlonii: The scheduler priority defaults to `NormalPriority` but the `currentUpdateLanePriority` defaults to `NoLanePriority` (so that we can tell whether we've set the priority or not).

We can remove this for now since we're not using the lane priority in this function anymore, but when it was changing the queue order we needed to use the default priority to match the same order as the scheduler.

packages/react-dom/src/events/ReactDOMEventReplaying.js
rickhanlonii: It looks like we import `react-reconciler/src/ReactFiberReconciler` everywhere and almost never just import from `react-reconciler` except in the NoopRenderer, so I moved this to import from the `ReactFiberReconciler`.

Let me know if that still wrong.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
rickhanlonii: That was my understanding as well, I think we just differed on the implementing details and what gets logged.

packages/react-dom/src/events/ReactDOMEventReplaying.js
rickhanlonii: Update: switching to `react-reconciler` causes failures due to a dependency cycle:

```
ReactDOMEventReplaying
  => react-reconciler
  => ReactFiberReconciler
  => ReactFiberReconciler.new.js
  => ReactFiberHostConfig
  => ReactDOMHostConfig
  => ReactDOMEventReplaying
```

This causes the host config to only be partially evaluated before it's used in places like [here](https://github.com/facebook/react/blob/a30a1c6ef3a621a57e9372c8249439dbfd9a7375/packages/react-reconciler/src/ReactFiberCompleteWork.new.js#L161), breaking all of the tests.

I think since we're already importing from `react-reconciler/src` here, for now it's safer to not introduce the cycle and import from `react-reconciler/src` again.

acdlite: > I think since we're already importing from react-reconciler/src here, for now it's safer to import from react-reconciler/src again and not introduce the cycle.

This rationale is exactly why I don't want to add more deep imports, because the next person will say "oh, well Rick already did it so it must be ok" 😆

As a concrete example for why deep imports aren't great, imagine if we decide to fork ReactFiberLane.js into separate new and old modules. Then we'd have to also fork the entirety of the React DOM renderer.

The dependency cycle is unfortunate. Longer term, we need to stop performing side effects during module initialization. In ReactFiberCompleteWork, the forking between mutation mode and persistent mode should happen inside of each export implementation, instead of at the module level.

Trying to think of a temporary solution...

packages/react-reconciler/src/ReactFiberHooks.new.js
acdlite: Assuming I'm correct here, can you go ahead and delete this please? Along with a note explaining why, since we'll likely refactor `useTransition` soon.

packages/react-dom/src/events/ReactDOMEventReplaying.js
acdlite: I would check for `NoLanePriority` explicitly

packages/react-reconciler/src/ReactFiberHooks.new.js
rickhanlonii: It does when there's no suspense config, e.g. [this test](https://github.com/facebook/react/blob/fe7163e73dadceda2655736d97cdd745d7abc8ea/packages/react-reconciler/src/__tests__/ReactTransition-test.js#L63).

rickhanlonii: I think for this we should go back to using the NormalPriority/DefaultLanePriority like the scheduler uses below, otherwise we'll need to special case the warning for this case since, as you can see, they'll mismatch.

rickhanlonii: The behavior then is "if you use startTransition without a suspense config then we use the default lane priority". We can update that later after we move off the scheduler priority if we want.

packages/react-dom/src/events/ReactDOMEventReplaying.js
acdlite: You could use the same injection strategy as here. Then the workarounds are consistent. https://github.com/facebook/react/blob/f4097c1aef173ea0cb873bc18b47d1ef12bab4b3/packages/react-dom/src/events/ReactDOMEventReplaying.js#L60-L66

But now that I'm looking at that, couldn't all the priority stuff move into `attemptHydrationAtCurrentPriority`? That's a reconciler function, and it's already being injected.

packages/react-reconciler/src/ReactFiberHooks.new.js
acdlite: Oh right

acdlite: I think in the future `startTransition` should set a default Suspense config object when one isn't provided explicitly. But let's punt on that for now.

packages/react-dom/src/events/ReactDOMEventReplaying.js
rickhanlonii: Agreed that it's problematic and that I'd be making the problem worse. Would you be up for me filing a task to fix the cycle after this lands? Then I can work on that while we're collecting the warnings that are logged.

If we move it into `attemptHydrationAtCurrentPriority`, how would we get the priority at the time that the hydration was queued?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -412,26 +415,56 @@ export function requestUpdateLane(
...
+      // inside React and use that priority to select a lane for this update.
+      // For now, we're just logging when they're different so we can assess.
+      const currentUpdateLanePriority = getCurrentUpdateLanePriority();
+
+      if (
+        schedulerLanePriority !== currentUpdateLanePriority &&
+        currentUpdateLanePriority !== NoLanePriority
+      ) {
+        if (__DEV__) {
+          console.error(
```
rickhanlonii, line 458: @acdlite is this the correct way to log the warning? There's no error code created with this, but we don't want to use invariant because it will throw. Will this still work with the warning filter and logging infra you expect?

packages/react-reconciler/src/ReactFiberHooks.new.js
rickhanlonii: SGTM!

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
acdlite: This is right. Warnings don't have codes. Only runtime errors. The filter works by matching the text of the warning.

packages/react-dom/src/events/ReactDOMEventReplaying.js
acdlite: Yeah I guess you'd still need a getter. How about:

- `getCurrentUpdatePriority()`
- `attemptHydrationAtPriority(fn, priority)`

And use injection, like we do for the other reconciler methods in this file

acdlite: If you fix this last thing then I'm comfortable stamping the PR.

The reason I'm being a stickler is because figuring out the layering of where the state (`currentUpdatePriority`) lives is crucial for deciding whether the overall approach is correct / worth it.

rickhanlonii: Sounds good, `attemptHydrationAtPriority` is the same as `ReactFiberReconciler.runWithPriority` anyway.

rickhanlonii:
I feel pretty good about this now, ship it?

packages/react-dom/src/events/ReactDOMEventReplaying.js
rickhanlonii: Done, note that `runWithPriority`, `getCurrentUpdatePriority` and `attemptHydrrationAtPriortity` all return/accept the `ReactPriorityLevel` instead of a `LanePriority` with the thinking that `ReactPriorityLevel` is already exposed and we probably don't want to expose Lane implementations until we know for sure that callers will need to know about them.

If you feel strongly that we should expose the LanePriory instead lmk and I can switch them before landing. The internal callsites still all use `setCurrentUpdateLanePriorty` and `getCurrentUpdateLanePriority` which _do_ operate on the LanePriority.

PR #19133, opened 2020-06-14
refactor(fixtures/art): Move Hardcoded colors from renderGraphic() funtion

ramsunvtech:
This simple PR moved the hard coded color code to outside class as variable in Fixtures.


gaearon:
Thanks for the effort, but it doesn't seem to meaningfully improve the code in my opinion. In general, we don't take purely stylistic changes.

ramsunvtech:
got it @gaearon. will come up with solid contribution next time 👍

PR #19136, opened 2020-06-15
[DRAFT] Devtools remove inspected window.eval

bvaughn:
Improves #19002 and general initialization/detection logic.

* Removes most `devtools.inspetedWindow.eval` usage (all that's in the critical path).
* Updates the test script to make it easier to test other URLs.

This PR also makes a *controversial* change to always show both "⚛️ Components" and "⚛️ Profiler" tabs. This lets us remove the `setInterval` polling logic that may have contributed to some cases of us failing to show the tabs, and also sometimes lead to confusion (semi frequent questions of "why don't I see the tabs?"). I can revert this change if it's frowned on though.

Here's what the extension looks like if you're on e.g. www.google.com

### Light mode
Screen Shot 2020-06-05 at 12 04 14 PM

### Dark mode
Screen Shot 2020-06-05 at 12 04 05 PM


packages/react-devtools-extensions/src/injectGlobalHook.js
```diff
@@ -16,28 +16,43 @@ function injectCode(code) {
...
-    };
-    chrome.runtime.sendMessage(lastDetectionResult);
-  } else if (evt.data.source === 'react-devtools-inject-backend') {
-    const script = document.createElement('script');
-    script.src = chrome.runtime.getURL('build/react_devtools_backend.js');
-    document.documentElement.appendChild(script);
-    script.parentNode.removeChild(script);
+
+  switch (data.source) {
+    case 'react-deventools-detector':
```
tjallingt, line 33:
```suggestion
    case 'react-devtools-detector':
```

typo i think?

bvaughn:
This does not fix things. Maybe we decide to always show the tab anyway, at some point but I'm going to close this for now as a failed investigation.

PR #19138, opened 2020-06-15
Fix spelling errors and typos

rickhanlonii:
## Overview

I committed a typo so to avoid embarrassment I hid the fix somewhere in this diff.


packages/react-devtools-inline/README.md
```diff
@@ -107,7 +107,7 @@ Sandboxed `iframe`s are also supported but require more complex initialization.
 ```js
 import { activate, initialize } from "react-devtools-inline/backend";
 
-// The DevTooks hook needs to be installed before React is even required!
```
bvaughn, line 110: I liked DevTooks 🙃

PR #19139, opened 2020-06-15
react-interactions: define dom events for contextmenu right click event and appropriate test

engprodigy:
simple event abstraction for mouse right click that brings up a context menu. This has not been exclusively defined before. This will serve as a basis for further abstraction of the right click mouse event that brings up a context menu @necolas and @trueadm 

`export function contextmenurightclick(payload) {
  return createMouseEvent('contextmenu', {
    button: buttonType.secondary,
    ...payload,
  });
}`


necolas:
The `contextmenu` event sequence already accepts a `pointerType` option that dispatches the event sequence expected for either mouse or touch pointers

engprodigy:
> The `contextmenu` event sequence already accepts a `pointerType` option that dispatches the event sequence expected for either mouse or touch point
@necolas Please can you demonstrate a left click event that brings up a context menu. Thanks.

engprodigy:
@necolas What can you suggest I work on in react interaction package? Thanks.

necolas:
We're not developing the package anymore

engprodigy:
> We're not developing the package anymore

Oh okay. Thanks for the reply. I've been away for some time though. Since I'm now back I will pick up something else then.

PR #19142, opened 2020-06-16
Re-enabled DebugTracing feature for old reconciler fork

bvaughn:
It was temporarily removed by @sebmarkbage via PR #18697. Newly re-added tracing is simplified, since the lane(s) data type does not require the (lossy) conversion between priority and expiration time values.

@sebmarkbage mentioned that he removed this because it might get in the way of his planned discrete/sync refactor. I'm not sure if that concern still applies, but just in case- I have only re-added it to the *old* reconciler fork for now.

For example, this code:
```js
class Example extends React.Component {
  state = {didMount: false};
  componentDidMount() {
    this.setState({didMount: true});
  }
  render() {
    return null;
  }
}

const root = createRoot(container);
root.render(
  <React.unstable_DebugTracingMode>
    <Example />
  </React.unstable_DebugTracingMode>,
);
```

Would produce this debug trace:
```
▼ ⚛️ commit (0b000000000000000000001000000000)
| ▼ ⚛️ layout effects (0b000000000000000000001000000000)
| | ⚛️ Example updated state (0b000000000000000000000000000001)
````

### DIFF
--- a/packages/react-reconciler/src/DebugTracing.js
+++ b/packages/react-reconciler/src/DebugTracing.js
@@ -7,6 +7,7 @@
  * @flow
  */
 
+import type {Lane, Lanes} from './ReactFiberLane';
 import type {Wakeable} from 'shared/ReactTypes';
 
 import {enableDebugTracing} from 'shared/ReactFeatureFlags';
@@ -16,6 +17,10 @@ let nativeConsoleLog: null | Function = null;
 
 const pendingGroupArgs: Array<any> = [];
 let printedGroupIndex: number = -1;
+
+function formatLanes(laneOrLanes: Lane | Lanes): string {
+  return '0b' + (laneOrLanes: any).toString(2).padStart(31, '0');
+}
 
 function group(...groupArgs): void {
   pendingGroupArgs.push(groupArgs);
@@ -57,11 +62,11 @@ function log(...logArgs): void {
 const REACT_LOGO_STYLE =
   'background-color: #20232a; color: #61dafb; padding: 0 2px;';
 
-export function logCommitStarted(priorityLabel: string): void {
+export function logCommitStarted(lanes: Lanes): void {
   if (__DEV__) {
     if (enableDebugTracing) {
       group(
-        `%c⚛️%c commit%c (priority: ${priorityLabel})`,
+        `%c⚛️%c commit%c (${formatLanes(lanes)})`,
         REACT_LOGO_STYLE,
         '',
         'font-weight: normal;',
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by bvaughn** [comment]
@sebmarkbage if you're no longer concerned about this, given the simplified lanes implementation, let me know and I'll re-add to the *new* fork as well.

File: packages/react-reconciler/src/DebugTracing.js [hunk 1]
```diff
@@ -17,6 +18,10 @@ let nativeConsoleLog: null | Function = null;
 const pendingGroupArgs: Array<any> = [];
 let printedGroupIndex: number = -1;
 
+function formatLanes(laneOrLanes: Lane | Lanes): string {
+  return '0b' + (laneOrLanes: any).toString(2).padStart(30, '0');
```
bvaughn, line 22: We could do fancier formatting here (including showing the lane "names" based on which bit positions were flipped on, but that would be more work to keep in sync and may change frequently with the implementation details- so just printing the number is probably sufficient?)

acdlite: I think we shouldn't leak that the Lanes type is a bitmask to the DevTools profiler. We should treat it more like a thread ID. So if you started rendering with lanes `0b10011`, the profiler would understand that as "Started rendering lanes: 0, 1, and 5." The actual transport format could use a bitmask for performance, but the profiler wouldn't deal with bitmasks anywhere else.

But I think it's good to use a bitmask for the debug tracing feature, since it's only used by us and will eventually  be replaced by the full profiler. Just making a note for the next step.

packages/shared/forks/ReactFeatureFlags.www.js
```diff
@@ -79,7 +79,7 @@ export const enableModernEventSystem = true;
 export const enableNewReconciler = __VARIANT__;
 
 // TODO: This does not currently exist in the Lanes implementation.
-export const enableDebugTracing = false;
+export const enableDebugTracing = !__VARIANT__;
```
acdlite, line 82: Can remove the TODO

packages/react-reconciler/src/DebugTracing.js
acdlite: I think it should be `padStart(31, '0')` since 31 is the max number of lanes.

acdlite: > the profiler wouldn't deal with bitmasks anywhere else.

Or it could, for performance reasons, but it would be its own type that doesn't assume anything about the bitmask type used by React.

The idea is that React should be able to change the layout of the Lanes type without breaking DevTools. For example, React might reserve one of the bits for some non-Lane purpose. Not sure how likely that is, but it's not inconceivable.

packages/shared/forks/ReactFeatureFlags.www.js
bvaughn: I think I just need to re-word it since it still doesn't exist in the .new implementation. (Until @sebmarkbage tells me it's okay to add there as well.)

packages/react-reconciler/src/DebugTracing.js
bvaughn: I think we'd still probably use a bitmask/number for serialization/transport, but we can use a separate opaque Flow type to avoid leaking the `Lane`/`Lanes` type directly.

> but it would be its own type that doesn't assume anything about the bitmask type used by React.

Right, other than the fact that it is a bitmask.

That being said, I think we may end up wanting to convey *some* information to the Profiler somehow (metadata or Flow types) if we want to be able to e.g. treat the sync lane specially in some way.

Have you seen my mockup (in the internal Core group) btw?

PR #19143, opened 2020-06-16
Remove React Flare discrete timeStamp flushing heuristic

trueadm:
This PR removes a heuristic we introduced with React Flare. Given we're in the process of gradually removing React Flare, we should also remove this heuristic as it causes a lack of flushing when dealing with discrete events that do not come from Flare (i.e. using the new `createEventHandle` primitive).

The reason for this is because Flare controlled the flushing itself, whilst the normal React event workflow is to flush when encountering a native event that is discrete. Flare introduced the timeStamp logic to prevent too many flushes, but now we're seeing that it actually causes issues with normal React (under-flushing). This is most evident when dealing with a11y use-cases, like `focus` and `blur`.


trueadm:
I'm not sure we need this anymore. Will do some more testing...

PR #19154, opened 2020-06-18
Update config.yml

ghost:
## Message
`yarn --frozen-lokfile` will be deleted in a future release of Yarn.
`yarn --immutable --immutable-cache --check-cache` slightly improves security.

From https://yarnpkg.com/cli/install:

> If the --immutable option is set, Yarn will abort with an error exit code if anything in the install artifacts (yarn.lock, .pnp.js, ...) was to be modified. For backward compatibility we offer an alias under the name of --frozen-lockfile, but it will be removed in a later release.
> 
> If the --immutable-cache option is set, Yarn will abort with an error exit code if the cache folder was to be modified (either because files would be added, or because they'd be removed).
> 
> If the --check-cache option is set, Yarn will always refetch the packages and will ensure that their checksum matches what's 1/ described in the lockfile 2/ inside the existing cache files (if present). **This is recommended as part of your CI workflow if you're both following the Zero-Installs model and accepting PRs from third-parties, as they'd otherwise have the ability to alter the checked-in packages before submitting them.**

## Summary
Update Yarn + slightly improved security

## Test Plan
- Invalidate CircleCI's cache
- Run CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ aliases:
   - &run_yarn
     run:
       name: Install Packages
-      command: yarn --frozen-lockfile
+      command: yarn --immutable --immutable-cache --check-cache
 
   - &attach_workspace
     at: build


prins-p:
Signed CLA.

rmdias (approved):

gaearon:
Is this a Yarn 1 or Yarn 2 option?

prins-p:
@gaearon good catch! I checked the source of V1 and this is not supported. Closing PR.

PR #19161, opened 2020-06-19
Re-enabled DebugTracing feature for old reconciler fork

bvaughn:
Debug tracing was reverted in #19159 because of a regression in `commitRootImpl`. This PR re-adds it without that regression.

---

#### Original commit message

It was temporarily removed by @sebmarkbage via PR #18697. Newly re-added tracing is simplified, since the lane(s) data type does not require the (lossy) conversion between priority and expiration time values.

@sebmarkbage mentioned that he removed this because it might get in the way of his planned discrete/sync refactor. I'm not sure if that concern still applies, but just in case- I have only re-added it to the old reconciler fork for now.

### DIFF
--- a/packages/react-reconciler/src/DebugTracing.js
+++ b/packages/react-reconciler/src/DebugTracing.js
@@ -7,6 +7,7 @@
  * @flow
  */
 
+import type {Lane, Lanes} from './ReactFiberLane';
 import type {Wakeable} from 'shared/ReactTypes';
 
 import {enableDebugTracing} from 'shared/ReactFeatureFlags';
@@ -16,6 +17,10 @@ let nativeConsoleLog: null | Function = null;
 
 const pendingGroupArgs: Array<any> = [];
 let printedGroupIndex: number = -1;
+
+function formatLanes(laneOrLanes: Lane | Lanes): string {
+  return '0b' + (laneOrLanes: any).toString(2).padStart(31, '0');
+}
 
 function group(...groupArgs): void {
   pendingGroupArgs.push(groupArgs);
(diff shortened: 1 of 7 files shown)


bvaughn:
CI failure from Code Sandbox is unrelated 😦
> Error: EACCES: permission denied, mkdir '/tmp/output/react'

bvaughn:
I had hoped to add a test that caught the previous regression, but I haven't been able to trigger it. As a sanity check, I also added an intentional `throw` statement if the lanes differed before and after, and run all of our tests and none of them triggered it either.

Open to suggestions!

acdlite:
@lunaruan is working on getting a repro so we can write a test that mimics the actual scenario that came up

bvaughn:
@acdlite Good to hear. I'll be happy to help her dig into this on Monday if she's interested :smile:

Given that this PR is purely additive (some new logging, nothing moved around or otherwise changed) do you think the PR needs to block on that test? Seems pretty safe IMO.

acdlite:
The only reason I would want to block the PR is to make sure we follow through on adding a test for the bug that manifested in www. The fact that we don't fully understand what happened implies to me that it's likely an important regression.

I had asked @lunaruan not to land a fix without a test because 1) I don't think it's blocking anything, since the profiler work can use this PR as a reference, and 2) I'm guessing it won't take more than a day or two to narrow down the bug.

But if you decide to merge without a test, I trust y'all to follow up with a test after.

bvaughn:
That makes sense. I'll check in with @lunaruan once she's in this morning about the test. I'll decide after talking to her whether I wait to reland this or not.

bvaughn:
Luna and I have been unable to repro this so far. We have a pairing session tomorrow to look into it with the person who reported the regression internally.

I am going to reland this now. In my mind, the test is now decoupled from this PR anyway since this PR no longer moves anything around.

Issue #19155, opened 2020-06-18
Bug: ESLint react-hooks/rules-of-hooks false negative when using default export with high-order functional component


tabuchi0919:
React version: 16.13.1
eslint-plugin-react-hooks version: 4.0.4

## Steps To Reproduce

1. Enable react-hooks/rules-of-hooks rule.
2. Lint following code.

```js
import React from 'react';
export const Foo = () => () => {
  React.useState(undefined);
  return <></>;
};
export default () => () => {
  React.useState(undefined);
  return <></>;
};
```

## The current behavior

ESLint does not show react-hooks/rules-of-hooks error when using export default.

## The expected behavior

ESLint shows following react-hooks/rules-of-hooks error even when using export default.

>React Hook "React.useState" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function.eslint(react-hooks/rules-of-hooks)

yuichkun:
I started working on a fix on this.

As I've read through the code and did some searching, I'm starting to think this is a bit complicated situation to fix regarding the structure of the AST and how `export default` works in combination with an anonymous function.

With that being said: 
I'll shortly send a PR, and hope to get feedback since I'm no expert in code analysis sort of thing :)

gaearon:
>ESLint shows following react-hooks/rules-of-hooks error even when using export default.

This is correct behavior. There is nothing for the lint rule to know that your unnamed function is actually a React component. In general, not giving exports names is a bad practice because you won't get meaningful stack traces.

```js
export default function generateMyComponent() {
  return function MyComponent() {
    React.useState(undefined);
    return <></>;
  }
};
```

This should work.

tabuchi0919:
Thank you for your comment!
I will use named export from now on.

PR #19169, opened 2020-06-20
[eslint-plugin-react-hooks]: enable linting default exported function

yuichkun:
## Summary
Relates to #19155 

So, basically I came across a situation where `eslint-plugin-react-hooks` does not throw any error on a default exported function.

As I've read through the code and found out that the root cause for this was around the `getFunctionName` function:

https://github.com/facebook/react/blob/6ba25b96df5d4179bf8aba3c3fe1ace3dce28234/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L339-L345

where it gets the function name from the given AST and (currently)  returns `undefined` for a default exported anonymous function.

So, this PR aims to handle that specific case, by

1. **not** letting `getFunctionName` return `undefined`
1. letting `isInsideComponentOrHook` return `true`
1. letting `isSomewhereInsideComponentOrHook` to be `true`
1. letting print the generic error message

#### Named Export
This throws a `genericError` as it is supposed to. (a testcase to make sure it works for named exports as well)
```javascript
export const Foo = () => {
  const cb = useCallback(() => {
    useHook()
  }, [])
  return <button onClick={cb}></button>
}
```

#### Default Export
This **now** throws a `genericError` and a `functionError`.
```javascript
export default () => {
  const cb = useCallback(() => {
    useHook()
  }, [])
  return <button onClick={cb}></button>
}
```

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
@@ -380,7 +380,7 @@ const tests = {
     },
     {
       code: `
-        // This is a false positive (it's valid) that unfortunately 
+        // This is a false positive (it's valid) that unfortunately
         // we cannot avoid. Prefer to rename it to not start with "use"
         class Foo extends Component {
           render() {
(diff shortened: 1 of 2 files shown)


gaearon:
I don't understand the PR description. `export default () => { ... }` should not be considered as a component by the rule. But you're saying it should be?

gaearon:
https://github.com/facebook/react/issues/19155#issuecomment-651875221

PR #19170, opened 2020-06-21
should check array first

oft:
reconcileChildFibers should check array first,  because typeof `typeof [] === "object"`,  check array first will reduce a judge for array component


gaearon:
Yes but this would force each non-array to get through `isArray || getIteratorFn` which is probably more expensive than we're willing to pay.

PR #19174, opened 2020-06-22
Refactor createEventHandle signature

trueadm:
This PR refactors the `ReactDOM.createEventHandle` API so that the signature is simpler and more barebones. Previously, `createEventHandle` returned an object with the methods `setListener` and `clear`, which mimiced what `useEvent` does.

After utilizing this API internally, primarily behind `useEvent`, it makes little sense for `createEventHandle` to have this signature. It still makses sense for `useEvent` to have this signtuare, but it's not important that the core API works the same. All we really want this core API to do is add an event and remove and event – sort of like a better native `addEventListener`.

This change means we no longer return an object from `createEventHandle`, but rather a function that sets the listener. Upon using the setter, you get back an unsubscribe/clear listener function.

```jsx
const setClick = ReactDOM.createEventHandle('click');
const clear = setClick(domNode, callback);
clear();
```

To show a more concrete example, previously you might use the API in this way:

```jsx
const clickHandle = ReactDOM.createEventHandle('click');

function Button({onClick}) {
  const ref = useRef(null);
  useEffect(() => {
    clickHandle.setListener(ref.current, onClick);
  });

  return <button ref={ref} />
}
```

With the changes in this PR, the above would look like this instead:

```jsx
const setClickListener = ReactDOM.createEventHandle('click');

function Button({onClick}) {
  const ref = useRef(null);
  useEffect(() => {
    return setClickListener(ref.current, onClick);
  }, [onClick]);

  return <button ref={ref} />
}
```

This change offers some benefits:
- reduces the amount of code in ReactDOM
- improves the runtime performance of using `createEventHandle` by not having as much overhead
- avoids accidently leaking nodes that React doesn't manage, by having to explicitly clear down correctly
- we don't need to garbage collect event listeners when unmount DOM host components

### DIFF
--- a/packages/react-dom/src/client/ReactDOMEventHandle.js
+++ b/packages/react-dom/src/client/ReactDOMEventHandle.js
@@ -47,6 +47,8 @@ type EventHandleOptions = {|
   priority?: EventPriority,
 |};
 
+const PossiblyWeakSet = typeof WeakSet === 'function' ? WeakSet : Set;
+
 function getNearestRootOrPortalContainer(node: Fiber): null | Element {
   while (node !== null) {
     const tag = node.tag;
@@ -72,12 +74,10 @@ function createEventHandleListener(
   type: DOMTopLevelEventType,
   capture: boolean,
   callback: (SyntheticEvent<EventTarget>) => void,
-  destroy: (target: EventTarget | ReactScopeInstance) => void,
 ): ReactDOMEventHandleListener {
   return {
     callback,
     capture,
-    destroy,
     type,
   };
 }
@@ -111,6 +111,65 @@ function registerEventOnNearestTargetContainer(
   );
 }
 
+function registerReactDOMEvent(
+  target: EventTarget | ReactScopeInstance,
+  topLevelType: DOMTopLevelEventType,
+  passive: boolean | void,
+  capture: boolean,
+  priority: EventPriority | void,
+): void {
+  // Check if the target is a DOM element.
+  if ((target: any).nodeType === ELEMENT_NODE) {
+    const targetElement = ((target: any): Element);
+    // Check if the DOM element is managed by React.
+    const targetFiber = getClosestInstanceFromNode(targetElement);
+    if (targetFiber === null) {
+      invariant(
+        false,
+        'ReactDOM.createEventHandle: setListener called on an element ' +
+          'target that is not managed by React. Ensure React rendered the DOM element.',
+      );
+    }
+    registerEventOnNearestTargetContainer(
+      targetFiber,
+      topLevelType,
+      passive,
+      priority,
+    );
+  } else if (enableScopeAPI && isReactScope(target)) {
+    const scopeTarget = ((target: any): ReactScopeInstance);
+    const targetFiber = getFiberFromScopeInstance(scopeTarget);
+    if (targetFiber === null) {
+      // Scope is unmounted, do not proceed.
+      return;
+    }
+    registerEventOnNearestTargetContainer(
+      targetFiber,
+      topLevelType,
+      passive,
+      priority,
+    );
+  } else if (isValidEventTarget(target)) {
+    const eventTarget = ((target: any): EventTarget);
+    const listenerMap = getEventListenerMap(eventTarget);
+    listenToTopLevelEvent(
+      topLevelType,
+      eventTarget,
+      listenerMap,
+      PLUGIN_EVENT_SYSTEM | IS_TARGET_PHASE_ONLY,
+      capture,
+      passive,
+      priority,
+    );
+  } else {
+    invariant(
+      false,
+      'ReactDOM.createEventHandle: setter called on an invalid ' +
+        'target. Provide a valid EventTarget or an element managed by React.',
+    );
+  }
+}
+
 export function createEventHandle(
   type: string,
   options?: EventHandleOptions,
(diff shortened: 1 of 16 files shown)


scripts/error-codes/codes.json
```diff
@@ -271,7 +269,6 @@
   "269": "Profiler must specify an \"id\" string and \"onRender\" function as props",
   "270": "The current renderer does not support persistence. This error is likely caused by a bug in React. Please file an issue.",
   "271": "Failed to replay rendering after an error. This is likely caused by a bug in React. Please file an issue with a reproducing case to help us find it.",
-  "272": "The current renderer does not support hydration. This error is likely caused by a bug in React. Please file an issue.",
```
bvaughn, line 274: Will deleting these duplicate error messages cause older versions of React to throw error codes that can't be looked up in our error decoder? Seems like it would?

trueadm: I just ran `extract-errors` and this was autogenerated. Should I re-add the old codes in?

bvaughn: Hm 🤔  I don't know!

bvaughn: I think it depends on if we ever made releases of React that logged the error code numbers you've deleted.

The [website error decoder](https://github.com/reactjs/reactjs.org/blob/master/src/components/ErrorDecoder/ErrorDecoder.js) uses a [custom Gatsby plug-in](gatsby-source-react-error-codes) we wrote which embeds [whatever the latest error codes file is](https://github.com/reactjs/reactjs.org/blob/master/plugins/gatsby-source-react-error-codes/gatsby-node.js#L3-L4) at the time we are updating the website (which we do with each release).

So I think this means that if an older version of React logged e.g. error code 272 rather than 305, then the website would not have any idea how to decode that error.

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
```diff
@@ -10,7 +10,7 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 
-const {useEffect, useRef} = React;
+const {useLayoutEffect, useRef} = React;
```
bvaughn, line 13: Was the replacement of `useEffect` with `useLayoutEffect` made to keep timing consistent with what we used to do in `commitUnmount`? At first I thought maybe it was related to the recent passive effects timing changes.

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
```diff
@@ -30,20 +30,48 @@ export default function useEvent(
...
+      ): void {
+        let clear = clears.get(target);
+        if (clear !== undefined) {
+          clear();
+        }
+        if (callback === null) {
+          clears.delete(target);
+          return;
+        }
+        clear = setListener(target, callback);
```
bvaughn, line 53: Hmm, so you can only attach a single listener to a given target for a specific event type? I wonder if this constraint might cause any composability issues with custom hooks. It's a little confusing since non-`useEvent`-event handler uniqueness is the combination of target + type + function, where as this... adding a listener to a target would silently remove a previously added listener. Why?

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
trueadm: Actually, we have a `useEvent` on www and found that we need to mostly use `useLayoutEffect` to add events, otherwise you run into issues with `focus` and `blur` not getting captured because they happen before `useEffect` fires. I just made the change to the React repo for this too.

bvaughn: Gotcha. That makes sense.

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
trueadm: This is how `useEvent` worked before, calling `setListener` would mean updating the listener to the new callback you pass, removing the old one. Passing `null` would remove the listener altogther. The difference weith this change is that we first unmount the old listener, before mounting the new one. It's now expected that the user would put the callback for the event listener in their dependencies array (for their `useLayoutEffect` or `useEffect`, so this should only occur if the callback actually changes).

bvaughn: The case I was concerned about was ... some part of code listening to an e.g. "click" handler and then another bit of code also listening and unwittingly removing the first handler. I guess you'll just need to be careful to avoid sharing `setListener` between code.

scripts/error-codes/codes.json
trueadm: I can re-add those entries and then we can have a discussion in our next sync about this.

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
trueadm: Yeah, that's another reason why I wanted to change `createEventHandle` – so at least it follows the mental model better. You subscribe to event X and get back an unsubscribe function from it.

In terms of `useEvent`, I agree the mental model changes. It's less about `addEventListener` here, more about "write-less-code" when hooking up events, and have the lifecycles clean up for you.

bvaughn: > Yeah, that's another reason why I wanted to change `createEventHandle` – so at least it follows the mental model better. You subscribe to event X and get back an unsubscribe function from it.

This make sense.

> In terms of `useEvent`, I agree the mental model changes. It's less about `addEventListener` here, more about "write-less-code" when hooking up events, and have the lifecycles clean up for you.

Maybe we could help here- avoid potential confusion- by requiring an explicit `clear` before `setListener` is called with a new function (and warning if you ever call `setListener` twice without clearing in between)? Seems like that would map nicely with the effects create+destroy pattern anyway.

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
necolas: > you run into issues with `focus` and `blur` not getting captured because they happen before useEffect fires

What are the specific issues?

bvaughn: I assumed the issue was that e.g. a newly added element would be focusable before the "focus" event listener was actually attached (so the event would be missed)

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
trueadm: I think we should tackle changing `useEvent` in a seperate PR, especially as it involves changing lots of internals on www, which blocks the syncs unless the changes are merged together with the infra changes needed to the `useEvent` on www.

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
trueadm: The were three cases that caused hundreds of internal tests to fail:
- when using the `autoFocus` prop on elements
- where we have user-land `focus()`, which is common in modals and rich text fields like DraftJS.
- there were some e2e test failures which seemed to be related to iframes

Switching to `useLayoutEffect` fixed all the above cases, and I don't imagine there to be any perf loss here either, so it seemed like a sensible change given the problems encountered.

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
bvaughn: Okay~ so long as we follow up on it? :smile:

Calling `setListener` twice without `clear` in between seems like an anti-pattern. It indicates you aren't using the effect cleanup callback correctly, or that you've written code that interferes with other code. I think we should have a warning for it.

trueadm: @bvaughn It depends what you're calling `setListener` on right?

You might be using the same `target` and wanting to update the listener to a new callback, in which case, that's fine.

Calling `clear` tears down all targets. Which means if you've used different targets then they also get affected. Example:

```jsx
function ConfirmationButtons({onAccept, onDecline}) {
  const clickHandle = useEvent('click');
  const acceptRef = useRef(null);
  const declineRef = useRef(null);

  useEffect(() => {
    clickHandle.setListener(acceptRef.current, onAccept);
    clickHandle.setListener(declineRef.current, onDecline);
  }, [onAccept, onDecline]);

  return (
    <>
      <button ref={declineRef}>Decline</button>
      <button ref={acceptRef}>Accept</button>
    </>
  )
}
```

The above pattern is meant to be promoted and you shouldn't need a `clear` anywhere here. `useEvent` automatically clears down its events when `ConfirmationButtons` gets unmounted.

bvaughn: > The above pattern is meant to be promoted and you shouldn't need a clear anywhere here. useEvent automatically clears down its events when ConfirmationButtons gets unmounted.

I think I overlooked the fact that you were auto clearing. I think I read it as you just clearing the Map. (`clears.clear()` is used in a lot of places.)

packages/react-dom/src/client/ReactDOMEventHandle.js
```diff
@@ -140,110 +199,39 @@ export function createEventHandle(
...
-
-    const destroy = (target: EventTarget | ReactScopeInstance): void => {
-      const listener = listeners.get(target);
-      if (listener !== undefined) {
-        listeners.delete(target);
-        const targetListeners = getEventHandlerListeners(target);
-        if (targetListeners !== null) {
-          targetListeners.delete(listener);
-        }
+    const registeredReactDOMEvents = new PossiblyWeakSet();
```
gaearon, line 202: Does this mean we leak memory on IE11? It looks like we always add targets to this Set, but it's never cleared (since it's assumed to be a WeakSet in modern browsers).

trueadm: Fixed in https://github.com/facebook/react/pull/19686.

PR #19184, opened 2020-06-24
Add new test cli

rickhanlonii:
## Summary

This PR adds a new test cli that will allow us to remove some of the test scripts in favor of command line options.

## Details
There are four base test scripts:
- `test oss`
- `test-www`
- `test-build-devtools`

But we also include these scripts:
- `debug-test`
- `test-www-variant`
- `test-prod-www`
- `test-prod-www-variant`
- `debug-test-persistent`
- `debug-test-prod`
- `test-build-prod`
- `debug-test-build-devtools`

These scripts are just permutations of the base scripts with these variations: 
- `NODE_ENV=production`
- `VARIANT=true`
- `debug=true`
- `release-channel=stable`
- `persistent=true`
- `build=true`

The issue is that these script aliases are hard to remember and are incomplete (e.g. `debug-test-prod-www-variant` is missing). To solve this, I've moved the three variations to cli arguments, so that we can run commands like:

```bash
$ yarn test --project=www --prod --debug --variant
```

All of the base commands can be ran with these flags, and we'll build the correct Jest config for the given options while passing additional options through to Jest, as in:

```bash
$ yarn test --prod --watch MyTest.js
```

## Deprecations

For now I'm keeping all of the existing test scripts, but I recommend using one of the base scripts so we can remove the aliases in the future.

## Debugger improvements

I've also:
- Added messages for the debug option with instructions to to open Chrome.
- Created a new entrypoint for the debugger's first breakpoint and provided a better message.

## Screens

Screen Shot 2020-06-24 at 5 05 02 PM

Screen Shot 2020-06-24 at 5 06 02 PM

Screen Shot 2020-06-24 at 5 07 38 PM

Screen Shot 2020-06-24 at 5 08 24 PM

Screen Shot 2020-06-23 at 11 27 30 PM


scripts/jest/jest-cli.js
```diff
@@ -0,0 +1,130 @@
...
+      default: false,
+    },
+    variant: {
+      alias: 'v',
+      describe: 'Run with www variant set to true.',
+      requiresArg: false,
+      type: 'boolean',
+      default: false,
+    },
+    config: {
```
acdlite, line 44: What's the use case for this? I'd prefer to think of the config files as an implementation detail. If there are config files aren't covered by the other options, might as well create an ad hoc test command.

scripts/jest/jest-cli.js
```diff
@@ -0,0 +1,130 @@
...
+  })
+  .options({
+    debug: {
+      alias: 'd',
+      describe: 'Run with node debugger attached.',
+      requiresArg: false,
+      type: 'boolean',
+      default: false,
+    },
+    prod: {
```
acdlite, line 30: There are three different build types: `prod`, `dev`, and `profiling`. We don't currently run tests against the profiling builds, but we might in the future. So we should implement this as an enum.

You can keep `--prod` as an alias, though.

acdlite:
Instead of `--config` let's try to choose higher level concepts.

I think the two most important testing dimensions are the Release Channel and the Build Type.

- **--release-channel**
  - `oss_stable` -> `RELEASE_CHANNEL=stable yarn test`
  - `oss_experimental` -> `RELEASE_CHANNEL=experimental yarn test`
  - `www_classic` -> `RELEASE_CHANNEL=stable yarn test-www`
  - `www_modern` -> `RELEASE_CHANNEL=experimental yarn test-www`
- **--build-type**
  - `dev` -> `yarn test`
  - `prod` -> `yarn test-prod`
  - `profiling` -> We don't currently test this build type

Running the tests against source versus the compiled artifacts is another dimension. Idk what to call that option. Maybe `--target=[source|build]`?

`--persistent` can also be its own flag.

acdlite:
A good test of your new API would be to translate all the combinations in the CircleCI config: https://github.com/facebook/react/blob/master/.circleci/config.yml

scripts/jest/jest-cli.js
```diff
@@ -0,0 +1,130 @@
...
+
+function getCommandArgs() {
+  // Add the correct Jest config.
+  const args = ['./scripts/jest/jest.js', '--config'];
+  args.push(configMap[argv.config]);
+
+  // Set the debug options, if necessary.
+  if (argv.debug) {
+    args.unshift('--inspect-brk');
+    args.push('--runInBand');
```
bvaughn, line 80: Nice 👍

scripts/jest/jest-cli.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  // Push the remaining args onto the command.
+  // This will send args like `--watch` to Jest.
+  args.push(argv._);
```
bvaughn, line 85: Won't this also pass through args like e.g. `--prod` and `--variant` (which don't mean anything to Jest)? It's probably NBD but might be confusing if one of our vanity args ever does conflict with Jest somehow. Maybe we should filter the known ones out before passing the `...rest`

rickhanlonii: It will not, yargs doesn't include the configured args in `argv._`. It actually doesn't include any named args at all unless you opt-in to the `unknown-options-as-args` as I've done here.

bvaughn: Oh! I didn't know that about `yargs`. Cool beans.

scripts/jest/jest-cli.js
rickhanlonii: Short term my goal was to make it easy to match the existing test scripts which are basically narrowed to these configs as:

- test
- test-www
- test-persistent
- test-build
- test-build-devtools

I'd agree that the configs are implementation details, but I also don't have a lot of context on what the other options should be and how they relate to which config should be run.

scripts/jest/jest-cli.js
bvaughn: IMO we can iterate on this. We weren't already running anything against profiling builds, so the new script doesn't really need to support this out of the box.

That being said, I would assume `buildType=prod|dev|profiling` with maybe `--prod`, `--dev`, and `--profiling` shortcut flags as optional syntactic sugar if you really felt like it.

rickhanlonii:
Thanks for the feedback @acdlite.

I'm not wild about the ergonomics of the `--release-channel` option because it's hard to type and remember, and only allows for one default instead of two. 

What if we do this instead (I'd like a better name for "config" here):

- `--release-channel (-r)` -> `stable` (default) or `experimental`
- `--config (-c)`-> `oss` (default) or `www`
- `--build-type (-b) -> `dev` (default) or `prod`
- `--target (-t)` -> `source` (default) or `build`
- `--persistent (-p)` -> bool
- `--variant (-v)` -> bool

This would allow you to do something like this:

```bash
yarn test -c www 
yarn test -b prod
yarn test -r experimental
yarn test -c www -r experimental -b prod -t build -v -p
```

For `target` and `persistent`, we have undefined behavior for options like:

```
yarn test --config www --persistent
yarn test --config www --target=build
yarn test --config devtools --target=source
```

I think for those we'll validate and error until we come back to update the Jest configs for those options.

scripts/jest/jest-cli.js
rickhanlonii: Looks like there are three "configs" (we need a better name, maybe "projects"):

- oss
- www
- devtools

rickhanlonii:
Here are all the tests we run in CI and the missing permutations for oss and www.

## oss
| name | build | stable | prod  | persistent |
| ----- | ----- | ----- | ----- |  ----- | 
| yarn_test | 🟥 | 🟥 | 🟥 | 🟥 | 
|  | 🟥 | 🟥 | 🟥 | 🟩 |
| yarn_test_prod | 🟥 | 🟥 | 🟩 | 🟥 |
|  | 🟥 | 🟥 | 🟩 | 🟩 | 
| stable_yarn_test | 🟥 | 🟩 | 🟥 | 🟥 |
| stable_yarn_test_persistent | 🟥 | 🟩 | 🟥 | 🟩 | 
| stable_yarn_test_prod | 🟥 | 🟩 | 🟩 | 🟥 | 
|  | 🟩 | 🟩 | 🟥 | 🟩 | 
| yarn_test_build | 🟩 | 🟥 | 🟥 | 🟥 |
|  | 🟩 | 🟥 | 🟥 | 🟩 | 
| yarn_test_build_prod  | 🟩 | 🟥 | 🟩 | 🟥 |
|  | 🟩 | 🟥 | 🟩 | 🟩 | 
| stable_yarn_test_build | 🟩 | 🟩 | 🟥 | 🟥 |
|  | 🟩 | 🟩 | 🟥 | 🟩 | 
| stable_yarn_test_build_prod | 🟩 | 🟩 | 🟩 | 🟥 |
|  | 🟩 | 🟩 | 🟩 | 🟩 | 

## www

| name | build | stable | prod | variant |
| ----- | ----- | ----- | ----- | ----- |
| yarn_test_www | 🟥 | 🟥 | 🟥 | 🟥 | 
| yarn_test_www_variant | 🟥 | 🟥 | 🟥 | 🟩 | 
| yarn_test_prod_www | 🟥 | 🟥 | 🟩 | 🟥 | 
| yarn_test_prod_www_variant | 🟥 | 🟥 | 🟩 | 🟩 | 
| stable_yarn_test_www | 🟥 | 🟩 | 🟥 | 🟥 | 
| stable_yarn_test_www_variant | 🟥 | 🟩 | 🟥 | 🟩 | 
| stable_yarn_test_prod_www | 🟥 | 🟩 | 🟩 | 🟥 |  
| stable_yarn_test_prod_www_variant | 🟥 | 🟩 | 🟩 | 🟩 | 
|  | 🟩 | 🟥 | 🟥 | 🟥 | 
|  | 🟩 | 🟥 | 🟥 | 🟩 | 
|   | 🟩 | 🟥 | 🟩 | 🟥 |
|  | 🟩 | 🟥 | 🟩 | 🟩 | 
|  | 🟩 | 🟩 | 🟥 | 🟥 |
|  | 🟩 | 🟩 | 🟥 | 🟩 | 
|  | 🟩 | 🟩 | 🟩 | 🟥 |
|  | 🟩 | 🟩 | 🟩 | 🟩 |

rickhanlonii:
- Updated the flags to match the strategy above
- Added TODOs and validation warnings for invalid options
- Updated circle.yaml to use the new commands

.circleci/config.yml
```diff
@@ -99,8 +99,7 @@ jobs:
       - *run_yarn
       - run:
           environment:
-            RELEASE_CHANNEL: stable
-          command: yarn test --maxWorkers=2
+          command: yarn test --release-channel=stable --ci
```
bvaughn, line 102: nit: I think `environment` can be removed now that it's empty, and `command` can be collapsed as well so this is just
```yaml
run: yarn test --release-channel=stable --ci
```

.circleci/config.yml
```diff
@@ -110,9 +109,7 @@ jobs:
       - *restore_yarn_cache
       - *run_yarn
       - run:
-          environment:
-            RELEASE_CHANNEL: experimental
-          command: yarn test --maxWorkers=2
+          command: yarn test --ci
```
bvaughn, line 112: same nit

.circleci/config.yml
```diff
@@ -122,9 +119,7 @@ jobs:
       - *restore_yarn_cache
       - *run_yarn
       - run:
-          environment:
-            RELEASE_CHANNEL: stable
-          command: yarn test-www --maxWorkers=2
+          command: yarn test --project=www --release-channel=stable --ci
```
bvaughn, line 122: same nit

Edit: I'm just going to stop saying this, but the same nit for all of the
```yaml
run:
  command: ...
```
blocks below :D

package.json
```diff
@@ -106,21 +107,21 @@
...
-    "test-build": "cross-env NODE_ENV=development jest --config ./scripts/jest/config.build.js",
-    "test-build-prod": "cross-env NODE_ENV=production jest --config ./scripts/jest/config.build.js",
-    "test-build-devtools": "cross-env NODE_ENV=development jest --config ./scripts/jest/config.build-devtools.js",
-    "debug-test-build-devtools": "cross-env NODE_ENV=development node --inspect-brk node_modules/jest/bin/jest.js --config ./scripts/jest/config.build-devtools.js",
+    "debug-test": "yarn test --debug",
+    "test": "node ./scripts/jest/jest-cli.js",
+    "test-www": "yarn test --project www",
+    "test-www-variant": "yarn test --project www --variant",
+    "test-prod-www": "yarn test --prod --project www ",
+    "test-prod-www-variant": "yarn test --prod --project www --variant",
```
bvaughn, line 115: Super trivial nit that I expect you to ignore *but*

Especially when we're mixing flags and enums, I think explicit "=" for enums make the command easier to visually parse.

scripts/jest/jest-cli.js
```diff
@@ -0,0 +1,268 @@
...
+      alias: 'd',
+      describe: 'Run with node debugger attached.',
+      requiresArg: false,
+      type: 'boolean',
+      default: false,
+    },
+    project: {
+      alias: 'p',
+      describe: 'Run the given project.',
+      requiresArg: false,
```
bvaughn, line 37: Shouldn't this be true?

scripts/jest/jest-cli.js
```diff
@@ -0,0 +1,268 @@
...
+      type: 'string',
+      default: 'oss',
+      choices: Object.keys(configMap),
+    },
+    releaseChannel: {
+      alias: 'r',
+      describe: 'Run with the given release channel.',
+      requiresArg: false,
```
bvaughn, line 45: Shouldn't this be true?

scripts/jest/jest-cli.js
```diff
@@ -0,0 +1,268 @@
...
+      type: 'string',
+      default: 'experimental',
+      choices: ['experimental', 'stable'],
+    },
+    env: {
+      alias: 'e',
+      describe: 'Run with the given node environment.',
+      requiresArg: false,
```
bvaughn, line 53: Shouldn't this be true?

scripts/jest/jest-cli.js
```diff
@@ -0,0 +1,268 @@
...
+    );
+    success = false;
+  }
+
+  if (argv.prod && argv.project === 'devtools') {
+    logError('DevTools do not support --prod. Remove this option to continue.');
+    success = false;
+  }
+
+  if (argv.env === 'production' && argv.project === 'devtools') {
```
bvaughn, line 115: DevTools doesn't support any of the following flags: `--dev`, `--prod`, `--env`, `--variant`, `--persistent `

Really it only supports `--debug` (and I guess `--build`) for the time being.

nit: Might be a little easier to read and reason about this if you grouped all of the "if devtools" conditionals within a single block, e.g.

```js
if (project === 'devtools') {
  // Other checks here...
}
```

scripts/jest/jest-cli.js
```diff
@@ -0,0 +1,268 @@
...
+  if (argv.build) {
+    args.push(buildConfig);
+  } else if (argv.persistent) {
+    args.push(persistentConfig);
+  } else {
+    args.push(configMap[argv.project]);
+  }
+
+  if (argv.build) {
+    // TODO: We could build this if it hasn't been built yet.
```
bvaughn, line 189: minor nit: We could check the contents of the `build` folder and give you a better warning if they're empty (or maybe even if they're older than some recent timestamp?)

acdlite:
That looks good but I would prefer to treat www’s Classic and Modern as release channels instead of adding a “project” dimension. That’s how we’re eventually planning to model it in the build script, too.

package.json
rickhanlonii: Thanks for the feedback, I went back and forth on this a few times, will update!

scripts/jest/jest-cli.js
rickhanlonii: We default to the OSS project so that `yarn test` runs the OSS config like it does today. Same for release channel, where we default to `experimental` like the configs do.

acdlite:
Using `—project` for DevTools makes sense, though

scripts/jest/jest-cli.js
bvaughn: Right, a default value makes sense- but passing the `--project` arg without any value (which is what `requiresArg` suggests) doens't.

rickhanlonii:
@andrew so you're saying:

```
--project=[core|devtools]
--release-channel=[stable, experimental, www-classic, www-modern]
```

scripts/jest/jest-cli.js
rickhanlonii: I'll update - can the devtools run without `--build`?

scripts/jest/jest-cli.js
rickhanlonii: Oh I misunderstood what `requiresArg` was, thanks for catching it!

scripts/jest/jest-cli.js
bvaughn: No. It's kind of maybe unfortunate, but DevTools (including its tests) can only run against a pre-built version of React for the time being.

bvaughn: The reason I said "and I guess `--build`" is because it isn't really a configurable option. Only one value is allowed 😅

acdlite:
@rickhanlonii Yeah that looks good

scripts/jest/jest-cli.js
rickhanlonii: Oh I see, so users should run `yarn build` before running the devtools tests? So we'll want to include the build directory check (I was excluding it). Will update now 👍

scripts/jest/jest-cli.js
rickhanlonii: Done!

.circleci/config.yml
```diff
@@ -121,10 +115,7 @@ jobs:
       - checkout
       - *restore_yarn_cache
       - *run_yarn
-      - run:
-          environment:
-            RELEASE_CHANNEL: stable
-          command: yarn test-www --maxWorkers=2
+      - run: yarn test --release-channel=www-classic --ci
 
   RELEASE_CHANNEL_stable_yarn_test_www_variant:
```
rickhanlonii, line 120: @acdlite should we rename this tests to `yarn_test_www_classic_variant` and similar for others?

acdlite: Yeah I was also hoping we could express the in terms of the new matrix jobs feature: https://circleci.com/blog/circleci-matrix-jobs/

scripts/jest/jest-cli.js
bvaughn: In most cases they can just run `./scripts/release/download-experimental-build.js`.

The only reason to run `yarn build-for-devtools` is to test a local change.

.circleci/config.yml
acdlite: That can be in a follow-up though. Doesn’t need to block.

PR #19186, opened 2020-06-24
Modern Event System: use focusin/focusout for onFocus/onBlur

trueadm:
Note: this PR requires [changes in JSDOM](https://github.com/jsdom/jsdom/pull/2996), otherwise Jest tests fail. Everything passes when I make the JSDOM changes locally. This PR also includes https://github.com/facebook/react/pull/19221.

Before the changes in this PR, we mapped `onFocus` and `onBlur` to `focus` and `blur`, but we did so in the capture phase. Doing so in the capture phase enables us to listen to all events as if they bubbled, which works great when you listen on the `document`.

With the modern event system, we no longer listen on the `document` for events, but rather with listen to the React root (or portal) container element. This means that if we listen to `focus` and `blur` in the capture phase, they actually occur in reverse order, which can lead to inconsistent UIs if `onFocus` or `onBlur` are used like `focusin` and `focusout` work (which we have recommended users to do in the past).

This PR alters how the modern event system handles `onFocus` and `onBlur` so that they now use `focusin` and `focusout` rather than `focus` and `blur`. This means we no longer need to listen to these events in the capture phase, ensuring consistentcy when using `onFocus` and `onBlur` with the expectation that bubbling order should correctly traverse through containers as expected.

In terms of breaking changes, this change will mean that React will not support Firefox versions earlier than 52. Earlier versions do not support `focusin` and `focusout`.

### DIFF
--- a/packages/react-dom/src/events/DOMEventProperties.js
+++ b/packages/react-dom/src/events/DOMEventProperties.js
@@ -41,7 +41,6 @@ const eventPriorities = new Map();
 
 // prettier-ignore
 const discreteEventPairsForSimpleEventPlugin = [
-  DOMTopLevelEventTypes.TOP_BLUR, 'blur',
   DOMTopLevelEventTypes.TOP_CANCEL, 'cancel',
   DOMTopLevelEventTypes.TOP_CLICK, 'click',
   DOMTopLevelEventTypes.TOP_CLOSE, 'close',
@@ -53,7 +52,8 @@ const discreteEventPairsForSimpleEventPlugin = [
   DOMTopLevelEventTypes.TOP_DRAG_END, 'dragEnd',
   DOMTopLevelEventTypes.TOP_DRAG_START, 'dragStart',
   DOMTopLevelEventTypes.TOP_DROP, 'drop',
-  DOMTopLevelEventTypes.TOP_FOCUS, 'focus',
+  DOMTopLevelEventTypes.TOP_FOCUS_IN, 'focus',
+  DOMTopLevelEventTypes.TOP_FOCUS_OUT, 'blur',
   DOMTopLevelEventTypes.TOP_INPUT, 'input',
   DOMTopLevelEventTypes.TOP_INVALID, 'invalid',
   DOMTopLevelEventTypes.TOP_KEY_DOWN, 'keyDown',
--- a/packages/react-dom/src/events/plugins/ModernChangeEventPlugin.js
+++ b/packages/react-dom/src/events/plugins/ModernChangeEventPlugin.js
@@ -17,10 +17,10 @@ import isTextInputElement from '../isTextInputElement';
 import {canUseDOM} from 'shared/ExecutionEnvironment';
 
 import {
-  TOP_BLUR,
+  TOP_FOCUS_OUT,
   TOP_CHANGE,
   TOP_CLICK,
-  TOP_FOCUS,
+  TOP_FOCUS_IN,
   TOP_INPUT,
   TOP_KEY_DOWN,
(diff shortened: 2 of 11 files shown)


packages/react-dom/src/events/DOMEventProperties.js
```diff
@@ -52,104 +140,119 @@ const eventPriorities = new Map();
...
-    DOMTopLevelEventTypes.TOP_BEFORE_BLUR,
-    DOMTopLevelEventTypes.TOP_AFTER_BLUR,
+if (enableModernEventSystem) {
+  // Map "onFocus" to "onFocusIn" and "onBlur" to "onFocusOut"
+  discreteEventPairsForSimpleEventPlugin.push(
+    TOP_FOCUS_IN,
+    'focus',
+    TOP_FOCUS_OUT,
+    'blur',
+  );
```
bvaughn, line 193: Should we also map `onFocusIn` -> "focusin" and `onFocusOut` -> "focusout" ?

packages/react-dom/src/events/plugins/ModernChangeEventPlugin.js
```diff
@@ -40,10 +40,10 @@ const eventTypes = {
       captured: 'onChangeCapture',
     },
     dependencies: [
-      TOP_BLUR,
+      TOP_FOCUS_IN,
       TOP_CHANGE,
       TOP_CLICK,
-      TOP_FOCUS,
+      TOP_FOCUS_OUT,
```
bvaughn, line 46: This is backwards.

packages/react-dom/src/events/plugins/ModernChangeEventPlugin.js
```diff
@@ -40,10 +40,10 @@ const eventTypes = {
       captured: 'onChangeCapture',
     },
     dependencies: [
-      TOP_BLUR,
+      TOP_FOCUS_IN,
```
bvaughn, line 43: This is backwards.

packages/react-dom/src/events/DOMEventProperties.js
trueadm: They are invalid events, you get a warning for using them.

bvaughn: Right, but why? Seems a little odd that we *disallow* you to listen for "focusin" (via `onFocusIn`) when that's what we're really using anyway.

trueadm: It's the same for why we don't allow `onInput` but use it anyway behind the scenes right?

bvaughn: I dunno. I mean, that also seems odd to me. 😄  I was just asking because it stood out. Your call on whether we should change it.

packages/react-dom/src/events/DOMEventProperties.js
```diff
@@ -79,6 +79,10 @@ const discreteEventPairsForSimpleEventPlugin = [
 
 const otherDiscreteEvents = [
   DOMTopLevelEventTypes.TOP_CHANGE,
+  // Ensure we still treat "focus" and "blur" as
+  // discrete events.
```
gaearon, line 83: Why can't we delete them altogether? What is their use now?

trueadm: Good point. I guess early on in this PR, we didn't know if we could remove focus and blur entirely. We believe we can now, so it's all good. :)

craigkovatch:
Thanks for this one, @trueadm !

PR #19188, opened 2020-06-24
Adds React.unstable_useSnapshotBeforeCommit

trueadm:
## Overview

This PR adds a new hook cakked `React.unstable_useSnapshotBeforeCommit`.

This new hook triggers a callback before we commit a fiber tree:

```jsx
// The order of console.logs should be:
// useSnapshotBeforeCommit
// useLayoutEffect

function MyComponent() {
  unstable_useSnapshotBeforeCommit(() -> {
    console.log('useSnapshotBeforeCommit fired');
  });
  useLayoutEffect(() => {
  console.log('useLayoutEffect fired');
  });

  return null;
}
```

## More details:
- `React.unstable_useSnapshotBeforeCommit` does __not__ allow for a cleardown function to be returned. (@sebmarkbage was this was you expect too?)
- It does allow taking a second argument to take dependencies as an array, like other hooks.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -127,6 +127,7 @@ import {
   HasEffect as HookHasEffect,
   Layout as HookLayout,
   Passive as HookPassive,
+  Snapshot as HookSnapshot,
 } from './ReactHookEffectTags';
 import {didWarnAboutReassigningProps} from './ReactFiberBeginWork.new';
 import {
@@ -231,6 +232,7 @@ function commitBeforeMutationLifeCycles(
     case ForwardRef:
     case SimpleMemoComponent:
     case Block: {
+      commitSnapshotHookEffectList(finishedWork);
       return;
     }
     case ClassComponent: {
@@ -375,6 +377,33 @@ function commitHookEffectListMount(tag: number, finishedWork: Fiber) {
               'An effect function must not return anything besides a function, ' +
                 'which is used for clean-up.%s',
               addendum,
+            );
+          }
+        }
+      }
+      effect = effect.next;
+    } while (effect !== firstEffect);
+  }
+}
+
+function commitSnapshotHookEffectList(finishedWork: Fiber) {
+  const updateQueue: FunctionComponentUpdateQueue | null = (finishedWork.updateQueue: any);
+  const lastEffect = updateQueue !== null ? updateQueue.lastEffect : null;
+  if (lastEffect !== null) {
+    const firstEffect = lastEffect.next;
+    let effect = firstEffect;
+    do {
+      if ((effect.tag & HookSnapshot) === HookSnapshot) {
+        // Run snapshot effect
+        const create = effect.create;
+        // There shouldn't be a return
+        const possibleBadValue = create();
+
+        if (__DEV__) {
+          if (typeof possibleBadValue === 'function') {
+            console.error(
+              "useSnapshotBeforeCommit's function must not return anything. Use an effect " +
+                'for when you want to return a function that handles clean-up.',
             );
           }
         }
--- a/packages/react-reconciler/src/ReactFiberCommitWork.old.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.old.js
@@ -127,6 +127,7 @@ import {
   HasEffect as HookHasEffect,
   Layout as HookLayout,
   Passive as HookPassive,
+  Snapshot as HookSnapshot,
 } from './ReactHookEffectTags';
 import {didWarnAboutReassigningProps} from './ReactFiberBeginWork.old';
 import {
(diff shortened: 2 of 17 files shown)


trueadm:
I've added tests now, so the PR is ready to review.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -384,6 +386,33 @@ function commitHookEffectListMount(tag: number, finishedWork: Fiber) {
...
+        // Run snapshot effect
+        const create = effect.create;
+        // There shouldn't be a return
+        const possibleBadValue = create();
+
+        if (__DEV__) {
+          if (typeof possibleBadValue === 'function') {
+            console.error(
+              "useSnapshotBeforeCommit's function must not return anything. Use an effect " +
+                'for when you want to return a function that handles clean-up.',
```
bvaughn, line 406: This error messaging seems weird.

> Use effect for when you want to return a function that handles clean-up.

That wouldn't be *the reason* you wanted to use an effect hook. They're for different use cases.

I think this method should be rewritten to be clear that we don't allow any side effects requiring cleanup during this phase.

packages/react-reconciler/src/ReactFiberCommitWork.old.js
```diff
@@ -384,6 +386,33 @@ function commitHookEffectListMount(tag: number, finishedWork: Fiber) {
...
+        // Run snapshot effect
+        const create = effect.create;
+        // There shouldn't be a return
+        const possibleBadValue = create();
+
+        if (__DEV__) {
+          if (typeof possibleBadValue === 'function') {
+            console.error(
+              "useSnapshotBeforeCommit's function must not return anything. Use an effect " +
+                'for when you want to return a function that handles clean-up.',
```
bvaughn, line 406: Same comment as on the .new file 😁

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1298,6 +1301,20 @@ function updateLayoutEffect(
   return updateEffectImpl(UpdateEffect, HookLayout, create, deps);
 }
 
+function mountSnapshotBeforeCommit(
+  create: () => void,
+  deps: Array<mixed> | void | null,
+): void {
+  return mountEffectImpl(SnapshotEffect, HookSnapshot, create, deps);
```
bvaughn, line 1308: Seems a little weird to reuse the effect hook infra for this. This isn't really an effect, so the type (which has an optional `destroy` function) seems wrong to reuse.

packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.js
```diff
@@ -2318,6 +2320,51 @@ describe('ReactHooksWithNoopRenderer', () => {
...
+          Scheduler.unstable_yieldValue(`Layout: ${getCommittedText()}`);
+        });
+        return <Text text={props.count} />;
+      }
+
+      ReactNoop.render(<Counter count={0} />, () =>
+        Scheduler.unstable_yieldValue('Sync effect'),
+      );
+      expect(Scheduler).toFlushAndYieldThrough([
+        [0, 'Snapshot: 0'],
```
bvaughn, line 2350: I'm confused. Why is this an array? Looks like you're yielding a string above.

InventingWithMonster:
Will this allow us to snapshot a component before it's unmounted?

bvaughn:
> Will this allow us to snapshot a component before it's unmounted?

No, at least not in the way it's currently implemented (and not if it follows the design of `getSnapshotBeforeUpdate` which I believe it will). When we mark a component for deletion, we clear other effects.

sebmarkbage:
I don’t think we’ll ship this particular implementation in a stable. There are other considerations for a full RFC.

packages/react-reconciler/src/ReactFiberHooks.new.js
trueadm: Good point.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
trueadm: I'm not sure what to word this to. Stating that this cannot have side-effects also kind of feels related to the name of this hook too, as if it can't have side-effects, should it refer to `render` in its name?

packages/react-reconciler/src/ReactFiberHooks.new.js
trueadm: Looking into this more, it seems that I'd be re-writing much of the logic for how effects are dealt with, which means most of the code feels duplicated. It also seems this is an effect, specifically a Snapshot effect, so it needs to be inserted into the effect list. I guess we handle the updateEffectImpl differently here, as we don't need to deal with adding a passive effect for the destroy.

trueadm:
Closing this as we'll add an internal work-around. In terms of this API, we can propose it first in an RFC.

PR #19204, opened 2020-06-27
Updated grammar typos in README markdown content

sasidharank:


sophiebits:
These are correct as written.

PR #19207, opened 2020-06-29
refactor • lines [ 1-10 ] : Change runTest function place.

MenaiAla:
Move `runTest` function to down for developer readability.


gaearon:
Thanks but this doesn't seem necessary.

Issue #19211, opened 2020-06-29
Bug: React hook state not cleared when rendering using ReactDOMServer if component errors


pmaccart:
In the react-dom package, React maintains some global internal state for hooks when server rendering a component (https://github.com/facebook/react/blob/v16.13.1/packages/react-dom/src/server/ReactPartialRendererHooks.js#L44-L54). The state is reset after [each function component finishes rendering](https://github.com/facebook/react/blob/v16.13.1/packages/react-dom/src/server/ReactPartialRenderer.js#L535) via [the `finishHooks` function](https://github.com/facebook/react/blob/v16.13.1/packages/react-dom/src/server/ReactPartialRendererHooks.js#L197-L201). 

Since the state of the hooks are only reset after a component finishes rendering, we can observe incorrect hooks state if a component that was using hooks raised an error while rendering, thus causing the finishHooks call to never execute. The next function component to render within the same process would then use the hooks state from the Component that previously failed to render, potentially causing mismatches.

We can work around this bug by rendering a no-op function component at the top of our react tree (which will cause `finishHooks` to properly run), but it seems like a more ideal fix would be to reset the hooks state as part of [the `prepareHooks` call](https://github.com/facebook/react/blob/v16.13.1/packages/react-dom/src/server/ReactPartialRendererHooks.js#L161-L173).  The comments actually have the code already present -- maybe there is a good reason for the state not to be reset there?

React version: 16.13.1

## Steps To Reproduce
Refer to the steps in the README of the example repo.

Link to code example:
https://github.com/pmaccart/react-hooks-ssr-state-leak

## The current behavior
The hooks state of a component is not cleared between renders

## The expected behavior
The hooks state of a component is cleared between renders

PR #19212, opened 2020-06-30
Fix state leaking when a function component throws on server render

pmaccart:
## Summary
Reset the internal hooks state when preparing a component for hooks usage, rather than after a component finishes rendering.  This fixes https://github.com/facebook/react/issues/19211, where the internal hooks state is not cleared if a function component throws an error while rendering.  

## Test Plan
- ` yarn test-prod`: validated no test regressions were introduced
- Updated the repro case from https://github.com/pmaccart/react-hooks-ssr-state-leak to use the new build; validated that the example scenario successfully rendered after set

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.js
@@ -889,20 +889,16 @@ describe('ReactDOMServerHooks', () => {
       return <div>{count}</div>;
     }
 
-    const container = document.createElement('div');
-
     // First, render a component that will throw an error during a re-render triggered
     // by a dispatch call.
-    try {
-      container.innerHTML = ReactDOMServer.renderToString(
-        <ThrowingComponent />,
-      );
-    } catch (e) {}
-    expect(container.children.length).toEqual(0);
+    expect(() => ReactDOMServer.renderToString(<ThrowingComponent />)).toThrow(
+      'Error from ThrowingComponent',
+    );
 
     // Next, assert that we can render a function component using hooks immediately
     // after an error occurred, which indictates the internal hooks state has been
     // reset.
+    const container = document.createElement('div');
     container.innerHTML = ReactDOMServer.renderToString(
       <NonThrowingComponent />,
     );
--- a/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.js
@@ -867,6 +867,48 @@ describe('ReactDOMServerHooks', () => {
     });
   });
 
+  it('renders successfully after a component using hooks throws an error', () => {
+    function ThrowingComponent() {
+      const [value, dispatch] = useReducer((state, action) => {
+        return state + 1;
+      }, 0);
+
+      // throw an error if the count gets too high during the re-render phase
+      if (value >= 3) {
+        throw new Error('Error from ThrowingComponent');
+      } else {
+        // dispatch to trigger a re-render of the component
+        dispatch();
+      }
+
+      return <div>{value}</div>;
+    }
+
+    function NonThrowingComponent() {
+      const [count] = useState(0);
+      return <div>{count}</div>;
+    }
+
+    const container = document.createElement('div');
+
+    // First, render a component that will throw an error during a re-render triggered
+    // by a dispatch call.
+    try {
+      container.innerHTML = ReactDOMServer.renderToString(
+        <ThrowingComponent />,
+      );
+    } catch (e) {}
+    expect(container.children.length).toEqual(0);
+
+    // Next, assert that we can render a function component using hooks immediately
+    // after an error occurred, which indictates the internal hooks state has been
+    // reset.
+    container.innerHTML = ReactDOMServer.renderToString(
+      <NonThrowingComponent />,
+    );
+    expect(container.children[0].textContent).toEqual('0');
+  });
+
   if (__EXPERIMENTAL__) {
     describe('useOpaqueIdentifier', () => {
       it('generates unique ids for server string render', async () => {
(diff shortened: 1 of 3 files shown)


gaearon:
Can you add a test that would have been failing before? That would make it easier to check the changes.

gaearon:
Can we instead add `resetHooksAfterThrow`, mirroring what we do in `ReactFiberHooks`?

FiberesimaJoseph:
Joseph Fiberesima

const Signup = () => {
    const [username, setUsername] = useState('')
    const [email, setEmail] = useState('')
    const [password, setPassword] = useState('')
    const [passwordConfirmation, setPasswordConfirmation] = useState('')

    const clearState = () => {
        setUsername('')
        setEmail('')
        setPassword('')
        setPasswordConfirmation('')
    }

    const handleSubmit = signupUser => e => {
        e.preventDefault()
        signupUser().then(data => {
            console.log(data)
            clearState() // <-----------
        })
    }

    return <JSX />
}

export default Signup
Each piece of state is used for a controlled input for the form.
.

Your code looks good, but if you want to decrease the functions needed you can put your entire form state in a single state variable object and reset to the initial object.

Example

const { useState } = React;

function signupUser() {
  return new Promise(resolve => {
    setTimeout(resolve, 1000);
  });
}

const initialState = {
  username: "",
  email: "",
  password: "",
  passwordConfirmation: ""
};

const Signup = () => {
  const [
    { username, email, password, passwordConfirmation },
    setState
  ] = useState(initialState);

  const clearState = () => {
    setState({ ...initialState });
  };

  const onChange = e => {
    const { name, value } = e.target;
    setState(prevState => ({ ...prevState, [name]: value }));
  };

  const handleSubmit = e => {
    e.preventDefault();
    signupUser().then(clearState);
  };

  return (
    <form onSubmit={handleSubmit}>
      <div>
        <label>
          Username:
          <input value={username} name="username" onChange={onChange} />
        </label>
      </div>
      <div>
        <label>
          Email:
          <input value={email} name="email" onChange={onChange} />
        </label>
      </div>
      <div>
        <label>
          Password:
          <input
            value={password}
            name="password"
            type="password"
            onChange={onChange}
          />
        </label>
      </div>
      <div>
        <label>
          Confirm Password:
          <input
            value={passwordConfirmation}
            name="passwordConfirmation"
            type="password"
            onChange={onChange}
          />
        </label>
      </div>
      <button>Submit</button>
    </form>
  );
};

ReactDOM.render(<Signup />, document.getElementById("root"));
<script src="https://unpkg.com/react@16/umd/react.development.js"></script>
<script src="https://unpkg.com/react-dom@16/umd/react-dom.development.js"></script>

<div id="root"></div>

pmaccart:
@gaearon Thanks for the feedback, and sorry for the slow response.

I made a couple updates to the PR:
1.  Added a unit test that fails prior to the fix ([64f35a1](https://github.com/facebook/react/pull/19212/commits/64f35a1f1966e1d850b76508c55b48ef0761ddab))
2.  Added a commit that implements a similar approach to what is taken in `ReactFiberHooks` -- eager to get your feedback on whether this approach is preferable, or whether we want to stick with resetting the state in the `prepareToUseHooks` function.  Happy to go either way on this.

packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.js
```diff
@@ -867,6 +867,48 @@ describe('ReactDOMServerHooks', () => {
...
+    function NonThrowingComponent() {
+      const [count] = useState(0);
+      return <div>{count}</div>;
+    }
+
+    const container = document.createElement('div');
+
+    // First, render a component that will throw an error during a re-render triggered
+    // by a dispatch call.
+    try {
```
gaearon, line 896: Can you wrap this in `expect(...).toThrow()` instead so we can assert the throw happened and the message matched what we expected?

pmaccart: Ah yes -- done.

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -202,16 +202,12 @@ export function finishHooks(
...
-  numberOfReRenders = 0;
-  renderPhaseUpdates = null;
-  workInProgressHook = null;
+  resetHooksState();
   if (__DEV__) {
     isInHookUserCodeInDev = false;
   }
 
-  // These were reset above
+  // These were reset via the resetHooksState() call
```
gaearon, line 210: Let's just remove this comment?

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -202,16 +202,12 @@ export function finishHooks(
 
     children = Component(props, refOrContext);
   }
-  currentlyRenderingComponent = null;
-  firstWorkInProgressHook = null;
-  numberOfReRenders = 0;
-  renderPhaseUpdates = null;
-  workInProgressHook = null;
+  resetHooksState();
   if (__DEV__) {
```
gaearon, line 206: Move this DEV-only block to reset too?

packages/react-dom/src/server/ReactPartialRendererHooks.js
pmaccart: Done.

packages/react-dom/src/server/ReactPartialRendererHooks.js
pmaccart: Done.

packages/react-dom/src/server/ReactPartialRendererHooks.js
```diff
@@ -202,24 +202,21 @@ export function finishHooks(
 
     children = Component(props, refOrContext);
   }
-  currentlyRenderingComponent = null;
```
gaearon, line 205: Did we lose currentlyRenderingComponent?

pmaccart: Oops, yes -- added that in as well.

gaearon:
Looks good. Thx

Issue #18844, opened 2020-05-06
Bug: Render after suspense is thrown away leaving DOM unmatched with state


joeljeske:
React version: 16.12.0

## Steps To Reproduce

 A component that suspends as a result of a context update is never rendered to DOM after suspension is finished.

Link to code example:

https://codesandbox.io/s/react-suspense-context-bug-rklls?file=/package.json:163-170

## The current behavior

If a component suspends as a result of a context update, the component's render is called after suspension is finished but the result of the render is not reflected in the DOM (and effects aren't called). 

## The expected behavior

If a component suspends as a result of a context update, the component's render should be called and flushed to the DOM and effects called. 

## Related Issues: 

https://github.com/facebook/react/issues/17356 - seems related to memoization but this does not use any memoization

chriserickson-cb:
We ran into this same problem, the codesandbox seems like a pretty good reproduction case.

We're on React 16.9 / React Native 0.61.5.

gaearon:
I think this is a duplicate of https://github.com/facebook/react/issues/17356.

>seems related to memoization but this does not use any memoization

It does use memoization in the sense that `children` is always going to be the same element since it's passed from above.

PR #19216, opened 2020-06-30
Don't bailout after Suspending in Legacy Mode

gaearon:
Fixes https://github.com/facebook/react/issues/17356.
Fixes https://github.com/facebook/react/issues/18844.

Has regression tests + verified sandboxes work.

Builds on the test case in https://github.com/facebook/react/pull/18572 and approach from https://github.com/facebook/react/pull/18572#issuecomment-612269971.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -63,6 +63,7 @@ import {
   Update,
   Ref,
   Deletion,
+  ForceUpdateForLegacySuspense,
 } from './ReactSideEffectTags';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import {
@@ -548,6 +549,13 @@ function updateSimpleMemoComponent(
           workInProgress,
           renderLanes,
         );
+      } else if (
+        (current.effectTag & ForceUpdateForLegacySuspense) !==
+        NoEffect
+      ) {
+        // This is a special case that only exists for legacy mode.
+        // See https://github.com/facebook/react/pull/19216.
+        didReceiveUpdate = true;
       }
     }
   }
@@ -3253,11 +3261,17 @@ function beginWork(
       }
       return bailoutOnAlreadyFinishedWork(current, workInProgress, renderLanes);
     } else {
-      // An update was scheduled on this fiber, but there are no new props
-      // nor legacy context. Set this to false. If an update queue or context
-      // consumer produces a changed value, it will set this to true. Otherwise,
-      // the component will assume the children have not changed and bail out.
-      didReceiveUpdate = false;
+      if ((current.effectTag & ForceUpdateForLegacySuspense) !== NoEffect) {
+        // This is a special case that only exists for legacy mode.
+        // See https://github.com/facebook/react/pull/19216.
+        didReceiveUpdate = true;
+      } else {
+        // An update was scheduled on this fiber, but there are no new props
+        // nor legacy context. Set this to false. If an update queue or context
+        // consumer produces a changed value, it will set this to true. Otherwise,
+        // the component will assume the children have not changed and bail out.
+        didReceiveUpdate = false;
+      }
     }
   } else {
     didReceiveUpdate = false;
--- a/packages/react-reconciler/src/ReactFiberBeginWork.old.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.old.js
@@ -63,6 +63,7 @@ import {
   Update,
   Ref,
   Deletion,
+  ForceUpdateForLegacySuspense,
 } from './ReactSideEffectTags';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import {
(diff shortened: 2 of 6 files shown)


packages/react-reconciler/src/ReactFiberThrow.new.js
```diff
@@ -260,6 +263,20 @@ function throwException(
...
+            if (currentSourceFiber !== null) {
+              // When we try rendering again, we should not reuse the current fiber,
+              // since it's known to be in an inconsistent state. Use a force update to
+              // prevent a bail out.
+              const update = createUpdate(NoTimestamp, SyncLane, null);
+              update.tag = ForceUpdate;
+              if (sourceFiber.updateQueue === null) {
+                sourceFiber.updateQueue = createFunctionComponentUpdateQueue();
+              }
+              ((sourceFiber.updateQueue: any): FunctionComponentUpdateQueue).hasForceUpdate = true;
```
gaearon, line 277: Should this ever get reset to false somewhere? I assume not because the update queue itself is supposed to get cleared up.

gaearon:
Seems like I'll need a broader fix with the effect tag so that it works for other Fiber types too.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -3069,6 +3071,7 @@ function beginWork(
     if (
       oldProps !== newProps ||
       hasLegacyContextChanged() ||
+      (current.effectTag & ForceUpdateForLegacySuspense) !== NoEffect ||
```
gaearon, line 3074: Kinda sucks to have this in the hot path. Not sure what else to do.

acdlite: If you mark an update on `fiber.lanes`, that will force it into the `else`  block.

gaearon: Don't we already do it? https://github.com/facebook/react/blob/f4097c1aef173ea0cb873bc18b47d1ef12bab4b3/packages/react-reconciler/src/ReactFiberThrow.old.js#L277

Doesn't seem to help.

gaearon: Ohh I see.

gaearon:
Upd.

packages/react-reconciler/src/ReactFiberBeginWork.old.js
```diff
@@ -3253,11 +3255,17 @@ function beginWork(
       }
       return bailoutOnAlreadyFinishedWork(current, workInProgress, renderLanes);
     } else {
-      // An update was scheduled on this fiber, but there are no new props
-      // nor legacy context. Set this to false. If an update queue or context
-      // consumer produces a changed value, it will set this to true. Otherwise,
-      // the component will assume the children have not changed and bail out.
-      didReceiveUpdate = false;
+      if ((current.effectTag & ForceUpdateForLegacySuspense) !== NoEffect) {
```
acdlite, line 3258: I wouldn't expect this check to be necessary, since whenever you hit this block, you'll always fall into the `updateSimpleMemoComponent` path, so you can do everything in there.

packages/react-reconciler/src/ReactFiberBeginWork.old.js
```diff
@@ -524,6 +525,7 @@ function updateSimpleMemoComponent(
     if (
       shallowEqual(prevProps, nextProps) &&
       current.ref === workInProgress.ref &&
+      (current.effectTag & ForceUpdateForLegacySuspense) === NoEffect &&
```
acdlite, line 528: I would move this check into the block below, and it it's true, set `didReceiveUpdate` to true to prevent a later bailout.

packages/react-reconciler/src/ReactFiberBeginWork.old.js
acdlite: Related: https://github.com/facebook/react/pull/19216/files#r447959102

packages/react-reconciler/src/ReactSideEffectTags.js
```diff
@@ -10,29 +10,29 @@
...
+export const Deletion = /*                     */ 0b000000000001000;
+export const ContentReset = /*                 */ 0b000000000010000;
+export const Callback = /*                     */ 0b000000000100000;
+export const DidCapture = /*                   */ 0b000000001000000;
+export const Ref = /*                          */ 0b000000010000000;
+export const Snapshot = /*                     */ 0b000000100000000;
+export const Passive = /*                      */ 0b000001000000000;
+export const Hydrating = /*                    */ 0b000010000000000;
+export const HydratingAndUpdate = /*           */ 0b000010000000100;
+export const Incomplete = /*                   */ 0b000100000000000;
```
acdlite, line 29: The reason these were in a separate section is that they aren't really effects, i.e. they don't have anything to  do with the commit phase. We just use this field to avoid adding another one.

Same with the new one you've added.

packages/react-reconciler/src/ReactFiberBeginWork.old.js
gaearon: This isn't specific to SimpleMemo though. This block is to fix the issue for all other cases (e.g. normal memo, or a function component).

gaearon:
Upd2. I think the generic check is still needed because this is a fix for more than SimpleMemo path. https://github.com/facebook/react/issues/18844 was about functions in general, with the referential equality bailout as the problem.

packages/react-reconciler/src/ReactFiberBeginWork.old.js
acdlite: Ah. God Legacy Mode is annoying. OK that makes sense then.

Issue #16734, opened 2019-09-10
window.print() crashes if a 'print' event listener causes a rerender (Chrome, DEV-mode only)


mwszekely:
**Do you want to request a *feature* or report a *bug*?**
Report a bug.

**What is the current behavior?**
Programmatically calling window.print() can cause React to report strange errors before crashing under certain circumstances.  The trigger *seems to be* a call to print() that results in a React state change somewhere (which, in Chrome, seems to happen because the print preview it shows can cause media query events, which can be hooked up to calls to a setState function).  This does not always happen, however--as shown in the example if the code is not called from in a setTimeout it doesn't seem to crash.  In addition, if the user initiates printing instead (e.g. via CTRL+P), React never crashes.

The two errors I've seen happen as a result of this are `Failed to execute 'handleEvent' on 'EventListener': The provided callback is no longer runnable` (needs "Pause on caught exceptions" in Chrome's DevTools to catch), and after a few of those they're followed by `Maximum update depth exceeded` (even though the setState function is only called once).  Once this happens, there's a good chance the tab become completely unresponsive after the print dialog is closed.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

A minimal example can be found here: https://codesandbox.io/s/immutable-snowflake-06cj2 . There are no external dependencies. The code is commented with instructions to reproduce the behavior.

I've also included a copy of the code here for completeness' sake:

```jsx
import React from "react";
import ReactDOM from "react-dom";

// Custom hook to match media queries.
// The setMatch call in this function is at the bottom of the error stack.
function useMediaQuery(query) {

  const [match, setMatch] = React.useState(() => window.matchMedia(query).matches);

  React.useEffect(() => {
    const queryList = window.matchMedia(query);

    /*************************************************************/
    /* If this function is not called, the crash does not occur. */
    /* When window.print is called, this effect is run as a      */
    /* consequence. However React handles it causes errors and   */
    /* other strange behavior. Because Firefox handles printing  */
    /* differently, it does not crash.                           */
    /* While this is an example of using setState in useEffect,  */
    /* it is only run once before the error is thrown anyway.    */
    /*************************************************************/
    const handleMatch = () => setMatch(queryList.matches);
    handleMatch();

    // Bookkeeping, not relevant to the crash.
    queryList.addListener(handleMatch);
    return () => queryList.removeListener(handleMatch);
  }, [query]);

  return match;
}


export function App() {
[...]

Issue #16585, opened 2019-08-27
useState's setState hangs when called in closed window


aaronjensen:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

If useState's setState is called in a closed window, it hangs the browser (at least Chrome). I agree that this seems like a weird issue, but we do this all of the time in our plugin we've built for an application that has iframe based UI.

This reproduces it. Just click "Do it". Note that your codesandbox service worker will spin up to 100% cpu and I can't figure out a way to stop it other than to kill that task in the Chrome task manager

https://codesandbox.io/s/react-setstate-on-hidden-window-hangs-repro-cixqn

**What is the expected behavior?**

No hang, it should just ignore the setState call, ideally.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

16.9. I'm not sure, but hooks are fairly new. This reproduces in Chrome 76, but not Firefox as far as I can tell.

zachary-nguyen:
I can confirm that the service worker indeed hits 100% cpu and hangs the page. Will take a look into this :)

aaronjensen:
FYI this still repros on 16.12.0.

gaearon:
https://github.com/facebook/react/pull/19220

aaronjensen:
Thank you, @gaearon :heart:

PR #19220, opened 2020-06-30
Fix development mode hang when iframe is removed

gaearon:
Verified it fixes https://github.com/facebook/react/issues/16585.
Verified it fixes https://github.com/facebook/react/issues/16734.

Follows @sophiebits's suggestion in https://github.com/facebook/react/issues/16734#issuecomment-609923768.

This is a workaround for a [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=956832) that also exists in Safari and doesn't seem to be gaining momentum. The actual cases in which this happens may be niche but I have a hunch I've run into other manifestations of this issue so I think it's worth fixing.

The basic idea is that if our shim doesn't work, we fall back to prod behavior with a simple try/catch.

### DIFF
--- a/packages/shared/invokeGuardedCallbackImpl.js
+++ b/packages/shared/invokeGuardedCallbackImpl.js
@@ -116,12 +116,7 @@ if (__DEV__) {
         'event',
       );
 
-      // Create an event handler for our fake event. We will synchronously
-      // dispatch our fake event using `dispatchEvent`. Inside the handler, we
-      // call the user-provided callback.
-      const funcArgs = Array.prototype.slice.call(arguments, 3);
-      function callCallback() {
-        didCall = true;
+      function restoreAfterDispatch() {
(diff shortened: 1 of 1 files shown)


packages/shared/invokeGuardedCallbackImpl.js
```diff
@@ -138,7 +133,15 @@ if (__DEV__) {
         ) {
           window.event = windowEvent;
```
sophiebits, line 134: I wonder why this doesn't do

```js
if (windowEventDescriptor) {
    Object.defineProperty(window, 'event', windowEventDescriptor);
}
```

as below.

gaearon: I'm scared to touch it. https://github.com/facebook/react/pull/11696#discussion_r153686685

PR #19221, opened 2020-06-30
Modern Event System: make on*Capture events use capture phase

trueadm:
Note: this PR is rebased on https://github.com/facebook/react/pull/19244.

This PR alters the modern event system so that events that are of the format `on*Capture` (capture events) are actually put into the browser's native capture event phase when React creates the event listener. This is a deparature of the previous logic, which "virtualized" the capture and bubble phases within the native bubble event listener.

This is an important change to the modern event system, as the modern event system delegates to React roots (or portals) rather than the `document`. This means events that are setup to be capture listener, i.e. `onClickCapture` now correctly behave has expected. This also fixes some long-standing issues with how React and the browser differ in regards to how events are handled in co-ordination with native event listeners.

Lastly, we still (for now) enforce `blur` and `focus` to be capture phase listeners (so they emulate two phase propagation in the native capture phase, as we did before). We also emulate two phase propagation in the native bubble phase, as we did before, for the `ChangeEventPlugin`, `BeforeInputEventPlugin` and `SelectEventPlugin` event plugins.

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -29,8 +29,8 @@ import {
   PLUGIN_EVENT_SYSTEM,
   LEGACY_FB_SUPPORT,
   IS_REPLAYED,
-  IS_TARGET_PHASE_ONLY,
   IS_CAPTURE_PHASE,
+  IS_TARGET_PHASE_ONLY,
 } from './EventSystemFlags';
 
 import {
@@ -131,6 +131,12 @@ function extractEvents(
   eventSystemFlags: EventSystemFlags,
   targetContainer: null | EventTarget,
 ) {
+  // TODO: we should remove the concept of a "SimpleEventPlugin".
+  // This is the basic functionality of the event system. All
+  // the other plugins are essentially polyfills. So the plugin
+  // should probably be inlined somewhere and have its logic
+  // be core the to event system. This would potentially allow
+  // us to ship builds of React without the polyfilled plugins below.
   ModernSimpleEventPlugin.extractEvents(
     dispatchQueue,
     topLevelType,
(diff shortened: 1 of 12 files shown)


trueadm:
@gaearon What Map access were you referencing specifically? I was also of the understanding that the difference was tiny, especially when the inline cache is utilized. I remember this being a perf issue a few years back with V8, but Sets and Maps are really fast now.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -229,38 +257,42 @@ function executeDispatch(
   event.currentTarget = null;
 }
 
-function executeDispatchesInOrder(
+function processDispatchQueueItemsInOrder(
   event: ReactSyntheticEvent,
-  capture: DispatchQueueItemPhase,
-  bubble: DispatchQueueItemPhase,
+  phase: DispatchQueueItemPhase,
```
gaearon, line 262: I find the `phase` naming confusing both in this argument and in the property. When I read `phase` I think of `'capture' | 'bubble'`. But it is actually more like an array of handlers. Can we rename this to `listeners`?

Same with the type name, I thought it refers to the string constants.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -362,11 +399,19 @@ export function listenToReactPropEvent(
   // this React prop event again.
   listenerMap.set(reactPropEvent, null);
   const dependencies = registrationNameDependencies[reactPropEvent];
+  const dependenciesLength = dependencies.length;
+  // If the dependencies length is 1, that means we're not using a polyfill
```
gaearon, line 403: This heuristic feels fragile. Although I guess it makes sense.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -362,11 +399,19 @@ export function listenToReactPropEvent(
...
+  const registrationCapturePhase =
+    isCaptureRegistrationName(reactPropEvent) && dependenciesLength === 1;
 
-  for (let i = 0; i < dependencies.length; i++) {
+  for (let i = 0; i < dependenciesLength; i++) {
     const dependency = dependencies[i];
-    const capture = capturePhaseEvents.has(dependency);
-
+    const capture =
+      capturePhaseEvents.has(dependency) || registrationCapturePhase;
```
gaearon, line 414: We already calculated `registrationCapturePhase` so let's reorder so it goes first and we don't always have to check the Set?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -348,6 +380,11 @@ export function listenToTopLevelEvent(
   }
 }
 
+function isCaptureRegistrationName(registrationName: string): boolean {
+  const len = registrationName.length;
+  return registrationName.substr(len - 7) === 'Capture';
```
gaearon, line 385: Nit: you can just write `registrationName.substr(-7)`

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -362,11 +399,19 @@ export function listenToReactPropEvent(
   // this React prop event again.
   listenerMap.set(reactPropEvent, null);
   const dependencies = registrationNameDependencies[reactPropEvent];
+  const dependenciesLength = dependencies.length;
+  // If the dependencies length is 1, that means we're not using a polyfill
+  // plugin like ChangeEventPlugin, BeforeInputPlugin, EnterLeavePlugin and
+  // SelectEventPlugin. SimpleEventPlugin always only has a single dependency.
+  // Given this, we know that we never need to apply capture phase event
+  // listeners to anything other than the SimpleEventPlugin.
+  const registrationCapturePhase =
```
gaearon, line 408: Nit: can we rename this to better signal what it represents? Like `isSomething`. I wouldn't nitpick but I genuinely struggle to piece together what exactly this Boolean is telling us

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -667,32 +716,35 @@ export function accumulateTwoPhaseListeners(
             const listener = listenersArr[i];
             const {callback, capture, type} = listener;
             if (type === targetType) {
-              if (capture === true) {
-                capturePhase.push(
+              if (capture && inCapturePhase) {
```
gaearon, line 719: I'm really struggling to follow what's happening here due to very similar naming. E.g. `capture`, `captured`, and `inCapturePhase` probably represent something different but I'm getting lost which is which because they're all used in the same function.

trueadm:
I'll merge this PR and address the concerns in a follow up, as much of what I've done there already fixes the comments from Dan.

PR #19223, opened 2020-07-01
Add user timing marks for scheduling profiler tool

taneliang:
## Summary

This PR adds User Timing marks at key points in the reconciler's execution.

The marks are used by this new [concurrent mode profiler tool](https://github.com/MLH-Fellowship/scheduling-profiler-prototype) that @kartik918 and I are building under @bvaughn and @jevakallio's guidance.

image

High level breakdown of this PR:

* Add a `enableSchedulingProfiling` feature flag.
* Add functions that call User Timing APIs to a new SchedulingProfiling.js file. The file follows `DebugTracing`'s structure.
* Add user timing marks to places where `DebugTracing` logs.
* Add user timing marks to most other places where @bvaughn's [original draft `DebugTracing` branch](https://github.com/facebook/react/compare/master...bvaughn:root-event-marks) marks.
* Tests added

More context (and discussions with @bvaughn) available at our internal PR https://github.com/MLH-Fellowship/react/pull/11 and issue https://github.com/MLH-Fellowship/react/issues/5.

Similar to `DebugTracing`, we've only added scheduling profiling calls to the old reconciler fork.

## Test Plan

- `yarn test`
- `yarn test-prod`
- `yarn lint`
- `yarn flow dom`
- This [test app](https://github.com/taneliang/concurrent-demo) that I created has a custom build of React with `enableSchedulingProfiling` and `enableDebugTracing` set to true. When profiled with Firefox Profiler, the marks are visible in the [output](https://profiler.firefox.com/public/cfcebc03b7ccd591136a7a9f8a5ff0be589b15fa/marker-chart/?globalTrackOrder=0-1-2&localTrackOrderByPid=20696-0-1~26532-0~&thread=5&v=4).
	image


packages/react-reconciler/src/ReactFiberReconciler.old.js
```diff
@@ -265,6 +267,15 @@ export function updateContainer(
   const suspenseConfig = requestCurrentSuspenseConfig();
   const lane = requestUpdateLane(current, suspenseConfig);
 
+  if (enableSchedulingProfiling) {
+    let componentName = 'Unknown';
+    if (ReactCurrentFiberCurrent !== null) {
+      componentName =
+        getComponentName(ReactCurrentFiberCurrent.type) || componentName;
```
bvaughn, line 274: Since `ReactCurrentFiberCurrent` is DEV-only, I think we should just remove the concept of component name from render-scheduled marks. It seems undesirable that "Unknown" could mean:
1. A component without a name is rendering.
2. No component is rendering.
3. A component is rendering, but we're in production mode so we don't know its name.

It's also odd that we're passing in the *name* of the current Fiber along with the root Fiber. Those always match up in other places.

packages/react-reconciler/src/SchedulingProfiling.js
```diff
@@ -0,0 +1,189 @@
...
+function getWakeableID(wakeable: Wakeable): number {
+  if (!wakeableIDs.has(wakeable)) {
+    wakeableIDs.set(wakeable, wakeableID++);
+  }
+  return ((wakeableIDs.get(wakeable): any): number);
+}
+
+export function markComponentSuspended(
+  componentName: string,
+  fiber: Fiber,
```
bvaughn, line 56: nit: Since we're passing the `Fiber` in question to these mark methods, we don't really need to pass the name as well. We could always call `getComponentName` in the mark methods.

packages/react-reconciler/src/SchedulingProfiling.js
```diff
@@ -0,0 +1,189 @@
...
+  wakeable: Wakeable,
+): void {
+  if (enableSchedulingProfiling) {
+    if (supportsUserTiming) {
+      const id = getWakeableID(wakeable);
+      const componentStack = getStackByFiberInDevAndProd(fiber) || '';
```
gaearon, line 62: Any concern that this is a bit expensive? I guess we only do that for suspended components.

gaearon: I'm wondering if we should cache stacks for Fibers we've already seen.

packages/react-reconciler/src/SchedulingProfiling.js
```diff
@@ -0,0 +1,189 @@
...
+      // TODO (brian) Generate and store temporary ID so DevTools can match up a component stack later.
```
bvaughn, line 63: I like Dan's suggestion of caching component stacks (using a weak map) for fibers we've seen.

Let's move these TODO comments above the calls to `getStackByFiberInDevAndProd` since it relates to those calls.

The idea behind the "TODO" was that- once this functionality is more tightly integrated into DevTools- try to store a (weak) map of id-to-Fiber here, and then let DevTools compute the component stack asynchronously later, using the Fiber. I'm not sure if that is feasible though, especially with future plans to replace the `alternate` model.

packages/react-reconciler/src/SchedulingProfiling.js
```diff
@@ -0,0 +1,189 @@
...
+      performance.mark(
+        `--suspense-suspend-${componentName}-${id}-${componentStack}`,
+      );
+      wakeable.then(
+        () =>
+          performance.mark(
+            `--suspense-resolved-${componentName}-${id}-${componentStack}`,
```
bvaughn, line 70: Do we have plans to use these ids in the profiler/visualizations? To be honest, I haven't thought about trying to dedup promises in the profiler UI. Maybe we don't need to maintain this map for now? At least not until we have a concrete idea for how we'd use the data?

packages/react-reconciler/src/SchedulingProfiling.js
```diff
@@ -0,0 +1,189 @@
...
+}
+
+export function markRenderScheduled(
+  componentName: string,
+  fiber: Fiber,
+  lane: Lane,
+): void {
+  if (enableSchedulingProfiling) {
+    if (supportsUserTiming) {
+      const componentStack = getStackByFiberInDevAndProd(fiber) || '';
```
bvaughn, line 144: I don't think we need a component stack here, because the `fiber` in this case is the root most Fiber. Let's just show its name.

packages/react-reconciler/src/SchedulingProfiling.js
```diff
@@ -0,0 +1,189 @@
...
+      // TODO (brian) Generate and store temporary ID so DevTools can match up a component stack later.
+      performance.mark(
+        `--schedule-forced-update-${componentName}-${formatLanes(
```
bvaughn, line 165: nit: Maybe `markForceUpdateScheduled` and `markStateUpdateScheduled` should align with the other mark methods and log the lane info first (followed by component name and stack)? Seems like that consistency would be nice.

packages/react-reconciler/src/SchedulingProfiling.js
gaearon: Btw we have something called `fiber._debugID`. Maybe we can repurpose this?

bvaughn: That sounds like it could work.

I'm not sure what the implications of the alternate -> previous change would be for this idea. It was kind of a half thought out idea anyway. Just something I wanted to put some thought into later because of the cost of getting the component stack.

Component stack calculation has gotten even more expensive recently with the "native" component stacks so... I don't know how I feel about this at the moment. It's definitely useful info to have, but maybe it's no longer worth the cost?

@sebmarkbage may have an opinion.

packages/shared/ReactFeatureFlags.js
```diff
@@ -15,6 +15,10 @@ export const enableFilterEmptyStringAttributesDOM = false;
 // Intended to enable React core members to more easily debug scheduling issues in DEV builds.
 export const enableDebugTracing = false;
 
+// Adds user timing marks for e.g. state updates, suspense, and work loop stuff,
+// for an experimental scheduling profiler tool.
+export const enableSchedulingProfiling = false;
```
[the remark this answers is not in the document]
taneliang: Sure, I'll rename this. Would we also want to rename the files from `SchedulingProfiling.js` to `SchedulingProfiler.js`? I'm just slightly concerned that it could mislead people into thinking that the file contains an actual profiler (like the DevTools Profiler).

packages/react-reconciler/src/SchedulingProfiling.js
taneliang: I haven't thought about it either, but I think it could be useful if we highlighted all suspense events related to the wakeable when one of them is hovered over? It's currently not possible to tell when a suspense event resolves/rejects, especially if there are many instances of the same component suspending in quick succession, which I'd imagine will be a very common case.

image

packages/react-reconciler/src/SchedulingProfiling.js
taneliang: I've removed both the component stack and the name because of the earlier comment https://github.com/facebook/react/pull/19223#discussion_r448383821. I don't have a clear understanding of what the `current` fiber that's passed to `markRenderScheduled` is so I can't really reason about what's happening, but I'm definitely only seeing an "Unknown" component in all my render scheduled calls. Let me know if I should add the component name back :)

packages/shared/ReactFeatureFlags.js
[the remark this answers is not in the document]
bvaughn: It would probably be nicer for the file and the feature flag to align, yes.

I don't think there's too much risk of confusion based on the name, since the package names are pretty clear :smile:

packages/react-reconciler/src/SchedulingProfiling.js
bvaughn: Okay. If we plan to use the data (and the example you mention seems reasonable) then 👍

packages/react-reconciler/src/SchedulingProfiling.js
bvaughn: Nope, this is fine 👍

packages/shared/ReactFeatureFlags.js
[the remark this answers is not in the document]
taneliang: Done!

bvaughn:
@kartik918 @taneliang When you're done pushing changes and this is ready for another review, let me know.

taneliang:
@bvaughn I've pushed up my last changes, this PR is ready

kartik918:
Same @bvaughn, you can review it again.

packages/shared/forks/ReactFeatureFlags.www-dynamic.js
```diff
@@ -17,7 +17,9 @@ export const warnAboutSpreadingKeyToJSX = __VARIANT__;
 export const disableInputAttributeSyncing = __VARIANT__;
 export const enableFilterEmptyStringAttributesDOM = __VARIANT__;
 export const enableLegacyFBSupport = __VARIANT__;
-export const enableDebugTracing = !__VARIANT__;
+// TODO: These features do not currently exist in the new reconciler fork.
+export const enableDebugTracing = !__VARIANT__ && __PROFILE__;
+export const enableSchedulingProfiler = !__VARIANT__;
```
bvaughn, line 22: Whoops, looks like I misspoke on my previous comment. My apologies. I meant to suggest:
```suggestion
// TODO: These features do not currently exist in the new reconciler fork.
export const enableDebugTracing = !__VARIANT__;
export const enableSchedulingProfiler = !__VARIANT__ && __PROFILE__;
```

packages/react-reconciler/src/SchedulingProfiler.js
```diff
@@ -0,0 +1,180 @@
...
+
+// $FlowFixMe: Flow cannot handle polymorphic WeakMaps
+const cachedFiberStacks: WeakMap<Fiber, string> = new PossiblyWeakMap();
+function cacheFirstGetComponentStackByFiber(workInProgress: Fiber): string {
+  if (!cachedFiberStacks.has(workInProgress)) {
+    const componentStack = getStackByFiberInDevAndProd(workInProgress) || '';
+    cachedFiberStacks.set(workInProgress, componentStack);
+  }
+  return ((cachedFiberStacks.get(workInProgress): any): string);
+}
```
bvaughn, line 65: Worth checking to see if we cached the alternate too! (Double check my code below, didn't test it.)
```suggestion
function cacheFirstGetComponentStackByFiber(fiber: Fiber): string {
  if (cachedFiberStacks.has(fiber)) {
    return ((cachedFiberStacks.get(fiber): any): string);
  } else {
    const alternate = fiber.alternate;
    if (alternate !== null && cachedFiberStacks.has(alternate)) {
      return ((cachedFiberStacks.get(alternate): any): string);
    }
  }
  const componentStack = getStackByFiberInDevAndProd(fiber) || "";
  cachedFiberStacks.set(fiber, componentStack);
  return componentStack;
}
```

packages/react-reconciler/src/SchedulingProfiler.js
```diff
@@ -0,0 +1,180 @@
...
+
+export function markComponentSuspended(fiber: Fiber, wakeable: Wakeable): void {
+  if (enableSchedulingProfiler) {
+    if (supportsUserTiming) {
+      const componentName = getComponentName(fiber.type) || 'Unknown';
+      const id = getWakeableID(wakeable);
+      const componentStack = cacheFirstGetComponentStackByFiber(fiber) || '';
```
bvaughn, line 72: nit: the `|| ''` here is redundant.

packages/react-reconciler/src/SchedulingProfiler.js
```diff
@@ -0,0 +1,180 @@
...
+        `--suspense-suspend-${componentName}-${id}-${componentStack}`,
+      );
+      wakeable.then(
+        () =>
+          performance.mark(
+            `--suspense-resolved-${componentName}-${id}-${componentStack}`,
+          ),
+        () =>
+          performance.mark(
+            `--suspense-rejected-${componentName}-${id}-${componentStack}`,
```
bvaughn, line 83: tiny nit: component name and stacks go together (at the end) for the other marks. Might be worth making this mark consistent too?

One small downside is potential confusion that the number we're logging here is an id, rather than a "lane" like the others, but I think that's fine :smile:

packages/react-reconciler/src/SchedulingProfiler.js
```diff
@@ -0,0 +1,180 @@
...
+      performance.mark(`--schedule-render-${formatLanes(lane)}`);
+    }
+  }
+}
+
+export function markForceUpdateScheduled(fiber: Fiber, lane: Lane): void {
+  if (enableSchedulingProfiler) {
+    if (supportsUserTiming) {
+      const componentName = getComponentName(fiber.type) || 'Unknown';
+      const componentStack = cacheFirstGetComponentStackByFiber(fiber) || '';
```
bvaughn, line 158: nit: the `|| ''` here is redundant.

packages/react-reconciler/src/SchedulingProfiler.js
```diff
@@ -0,0 +1,180 @@
...
+      );
+    }
+  }
+}
+
+export function markStateUpdateScheduled(fiber: Fiber, lane: Lane): void {
+  if (enableSchedulingProfiler) {
+    if (supportsUserTiming) {
+      const componentName = getComponentName(fiber.type) || 'Unknown';
+      const componentStack = cacheFirstGetComponentStackByFiber(fiber) || '';
```
bvaughn, line 172: nit: the `|| ''` here is redundant.

taneliang:
@bvaughn all done!

packages/react-reconciler/src/ReactFiberComponentStack.js
```diff
@@ -59,6 +59,7 @@ function describeFiber(fiber: Fiber): string {
 }
 
 export function getStackByFiberInDevAndProd(workInProgress: Fiber): string {
+  // TODO (brian) Generate and store temporary ID so DevTools can match up a component stack later.
```
bvaughn, line 62: Oops I missed this. This is the wrong place for this comment. The comment belongs in the user timing mark code, as a note that maybe we don't need to pre-generate stacks in the future.

taneliang: Oh sorry, I misread your earlier comment. I'll move this comment to above the calls to `getStackByFiberInDevAndProd`

bvaughn: No worries.

packages/react-reconciler/src/ReactFiberReconciler.old.js
```diff
@@ -265,6 +267,10 @@ export function updateContainer(
   const suspenseConfig = requestCurrentSuspenseConfig();
   const lane = requestUpdateLane(current, suspenseConfig);
 
+  if (enableSchedulingProfiler) {
+    markRenderScheduled(current, lane);
```
bvaughn, line 271: Sorry I missed this earlier, but doing a final review before merging and noticed the `current` param is still being passed (but not used). Can we remove it?

bvaughn:
Looks like we also need to rebase on `master` one more time and resolve conflicts in a few of the feature flag files. Then I'll merge!

bvaughn:
Not sure why off the top of my head, but the "merge master" commit seems to be causing tests to fail.

Maybe you could undo that commit, and do a rebase instead? That would make it easier to read since d2c090e has a lot of random changes in it.

bvaughn:
Let me kick CI to re-run just in case.

taneliang:
This is really strange, because I can repro this locally with `yarn test --release-channel=stable SchedulingProfiler`, and I can also repro this with the earlier commits in this PR even though I'm pretty sure CI was passing at that time.

bvaughn:
Ah, good to note.

bvaughn:
Try changing all of your `@gate` expressions to remove the `experimental` condition.

bvaughn:
Nice! Great work.

taneliang:
Thanks, that did it. Interesting though, why was CI passing in the previous commits, since the CI scripts should have also run the tests in non-experimental release channels 🤔

bvaughn:
@taneliang Unfortunately we've been having some flaky CI troubles the past few days. It's impacted more than just this PR ☹️ One of us is going to have to dig in and figure it out but we haven't yet.

gaearon:
>One of us is going to have to dig in and figure it out but we haven't yet.

Didn't we fix it? https://github.com/facebook/react/pull/19265

The failures were silent *before* that fix which is why you didn't see them until you updated this branch.

bvaughn:
Nice! I didn't know Dominic had already fixed that. Thanks for the pointer.

taneliang:
Ah, that explains it, looks like it was pretty subtle. Thanks for the explanation!

PR #19224, opened 2020-07-01
Update README

yangteresa:


bvaughn:
Not sure what's going on here, but it looks like this PR was opened in mistake?

PR #19225, opened 2020-07-01
Run ReactBrowserEventEmitter test on bundles

gaearon:
This test is bizarre. I wanted to delete it but meh, it's still a bit useful. So I just removed internal usage.


packages/react-dom/src/__tests__/ReactBrowserEventEmitter-test.js
```diff
@@ -159,47 +126,6 @@ describe('ReactBrowserEventEmitter', () => {
     container = null;
   });
 
-  it('should store a listener correctly', () => {
```
gaearon, line 162: Meh on these.

PR #19228, opened 2020-07-01
Fork legacy-events folder into react-dom and react-native

gaearon:
This removes top-level `legacy-events/*` and instead copies all files into two folders:

- `react-dom/src/legacy-events/*`
- `react-native-renderer/src/legacy-events/*`

I updated all the import references.

This gives us a clear separation and lets us remove most of that from the DOM folder when we flip the flag.


packages/react-native-renderer/src/legacy-events/EventBatching.js
```diff
@@ -0,0 +1,66 @@
+/**
```
gaearon, line 1: Everything below is just folder copy paste. I made no changes.

PR #19230, opened 2020-07-01
Enable modern event system and delete dead code

gaearon:
In https://github.com/facebook/react/pull/19228, I forked legacy event system folder for DOM and Native.
In this PR, I am removing the `enableModernEventSystem` flag altogether:

- In DOM fork, I am removing all its falsy branches and consider `enableModernEventSystem = true`.
- In Native, we remove all its truthy branches and consider `enableModernEventSystem = false`.

This enables us to remove a bunch of files and codepaths in DOM. There's more to do there but this is already fairly invasive so I will send the rest as follow-ups.

You should review without whitespace. https://github.com/facebook/react/pull/19230/files?diff=split&w=1


packages/react-dom/src/events/plugins/__tests__/ModernSimpleEventPlugin-test.js
```diff
@@ -231,15 +231,14 @@ describe('SimpleEventPlugin', function() {
...
-
       ReactDOM = require('react-dom');
       Scheduler = require('scheduler');
     });
 
     // @gate experimental
     it('flushes pending interactive work before extracting event handler', () => {
       container = document.createElement('div');
-      const root = ReactDOM.unstable_createRoot(container);
+      const root = ReactDOM.createRoot(container);
```
gaearon, line 241: I didn't change this one. Git just thinks I renamed it whereas I actually deleted old test.

Kind of confusing why it uses unprefixed name. I thought that doesn't work anymore.

packages/react-native-renderer/src/legacy-events/SyntheticEvent.js
```diff
@@ -77,10 +76,8 @@ function SyntheticEvent(
   this.dispatchConfig = dispatchConfig;
   this._targetInst = targetInst;
   this.nativeEvent = nativeEvent;
-  if (!enableModernEventSystem) {
-    this._dispatchListeners = null;
-    this._dispatchInstances = null;
-  }
+  this._dispatchListeners = null;
```
trueadm, line 79: We don't need these. I think you got the condition the wrong way around here.

packages/react-native-renderer/src/legacy-events/SyntheticEvent.js
```diff
@@ -157,79 +154,71 @@ Object.assign(SyntheticEvent.prototype, {
    * won't be added back into the pool.
    */
   persist: function() {
-    // Modern event system doesn't use pooling.
-    if (!enableModernEventSystem) {
-      this.isPersistent = functionThatReturnsTrue;
-    }
+    this.isPersistent = functionThatReturnsTrue;
```
trueadm, line 157: Same here, we don't need this line.

packages/react-native-renderer/src/legacy-events/SyntheticEvent.js
```diff
@@ -157,79 +154,71 @@ Object.assign(SyntheticEvent.prototype, {
...
-        }
-      }
-      this.dispatchConfig = null;
-      this._targetInst = null;
-      this.nativeEvent = null;
-      this.isDefaultPrevented = functionThatReturnsFalse;
-      this.isPropagationStopped = functionThatReturnsFalse;
-      this._dispatchListeners = null;
-      this._dispatchInstances = null;
+    const Interface = this.constructor.Interface;
```
trueadm, line 171: Same here

packages/react-native-renderer/src/legacy-events/SyntheticEvent.js
```diff
@@ -339,28 +325,21 @@ function createOrGetPooledEvent(
...
 
 function addEventPoolingTo(EventConstructor) {
   EventConstructor.getPooled = createOrGetPooledEvent;
-
-  // Modern event system doesn't use pooling.
-  if (!enableModernEventSystem) {
-    EventConstructor.eventPool = [];
-    EventConstructor.release = releasePooledEvent;
-  }
+  EventConstructor.eventPool = [];
```
trueadm, line 341: And here

packages/react-native-renderer/src/legacy-events/SyntheticEvent.js
gaearon: These are in the native renderer. The native renderer uses the old event system, so we need these.

gaearon:
Hmm. I'm not sure I follow. The native renderer uses the _legacy_ system, right? We don't yet remove pooling on RN either.

PR #19234, opened 2020-07-01
Make event plugin injection statically resolvable

gaearon:
There are only two parts of an event "plugin". Extracting function and a list of event types.

This turns plugins into ESM that use named exports for these.
Then it changes the "plugin registry" to resolve each of these two things statically.

This makes inlining possible. Doesn't seem like it influences the size but I think this is a step towards reducing the indirection in code. The next bastion are the configs themselves.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -8,7 +8,7 @@
  */
 
 import {
-  registrationNameModules,
+  registrationNames,
   possibleRegistrationNames,
 } from '../events/EventPluginRegistry';
 import {canUseDOM} from 'shared/ExecutionEnvironment';
@@ -133,7 +133,7 @@ if (__DEV__) {
     validateARIAProperties(type, props);
     validateInputProperties(type, props);
     validateUnknownProperties(type, props, {
-      registrationNameModules,
+      registrationNames,
       possibleRegistrationNames,
     });
   };
@@ -356,7 +356,7 @@ function setInitialDOMProperties(
       // We could have excluded it in the property list instead of
       // adding a special case here, but then it wouldn't be emitted
       // on server rendering (but we *do* want to emit it in SSR).
-    } else if (registrationNameModules.hasOwnProperty(propKey)) {
+    } else if (registrationNames.hasOwnProperty(propKey)) {
       if (nextProp != null) {
         if (__DEV__ && typeof nextProp !== 'function') {
           warnForInvalidEventListener(propKey, nextProp);
@@ -694,7 +694,7 @@ export function diffProperties(
       // Noop
     } else if (propKey === AUTOFOCUS) {
       // Noop. It doesn't work on updates anyway.
-    } else if (registrationNameModules.hasOwnProperty(propKey)) {
+    } else if (registrationNames.hasOwnProperty(propKey)) {
       // This is a special case. If any listener updates we need to ensure
       // that the "current" fiber pointer gets updated so we need a commit
       // to update this element.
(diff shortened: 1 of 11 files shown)


packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -8,7 +8,7 @@
  */
 
 import {
-  registrationNameModules,
+  registrationNames,
```
gaearon, line 11: This used to be an object hash of name -> plugin. But the plugin object wasn't actually used. I changed it to a hash of name -> true. This lets us get rid of object "plugin" representations.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -218,22 +219,17 @@ function dispatchEventsForPlugins(
...
-      dispatchQueue,
-      topLevelType,
-      targetInst,
-      nativeEvent,
-      nativeEventTarget,
-      eventSystemFlags,
-      targetContainer,
-    );
-  }
+  extractEvents(
```
gaearon, line 224: This calls into a function that will statically enumerate each of them.

packages/react-dom/src/events/EventPluginRegistry.js
```diff
@@ -135,17 +103,71 @@ export const registrationNameDependencies = {};
...
-  const publishedEvents = pluginModule.eventTypes;
-  for (const eventName in publishedEvents) {
-    publishEventForPlugin(publishedEvents[eventName], pluginModule, eventName);
+function injectEventPlugin(eventTypes: EventTypes): void {
+  for (const eventName in eventTypes) {
+    publishEventForPlugin(eventTypes, eventName);
   }
 }
 
+export function extractEvents(
```
gaearon, line 112: This is the new static thing.

packages/react-dom/src/events/PluginModuleType.js
```diff
@@ -50,16 +36,3 @@ export type DispatchQueueItem = {|
 |};
 
 export type DispatchQueue = Array<DispatchQueueItem>;
-
-export type ModernPluginModule<NativeEvent> = {
```
gaearon, line 54: This concept sort of goes away.

packages/react-dom/src/events/plugins/ModernChangeEventPlugin.js
```diff
@@ -5,6 +5,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import type {EventTypes} from '../PluginModuleType';
```
gaearon, line 8: This isn't covered by Flow yet but this is preparing for your other PR.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -114,6 +114,74 @@ import {
 } from './EventListener';
 import {removeTrappedEventListener} from './DeprecatedDOMEventResponderSystem';
 import {topLevelEventsToDispatchConfig} from './DOMEventProperties';
+import * as ModernBeforeInputEventPlugin from './plugins/ModernBeforeInputEventPlugin';
```
trueadm, line 117: Love this :D

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -114,6 +114,74 @@ import {
...
+function extractEvents(
+  dispatchQueue: DispatchQueue,
+  topLevelType: TopLevelType,
+  targetInst: null | Fiber,
+  nativeEvent: AnyNativeEvent,
+  nativeEventTarget: null | EventTarget,
+  eventSystemFlags: EventSystemFlags,
+  targetContainer: null | EventTarget,
+) {
+  ModernSimpleEventPlugin.extractEvents(
```
trueadm, line 139: Do these inline fully still? Nice if they do!

gaearon: Yeah, these inline fully.

PR #19235, opened 2020-07-01
Remove TestUtils dependency on event registry

gaearon:
This removes the last TestUtils link to the event system internals.

The dispatch config map was used to populate the Simulate helpers. I just took a snapshot of the list we have today, and inlined it in TestUtils. We also used the config to tell whether an event is two-phased or not. However there are only four events that don't have a capture phase so I have explicitly listed them instead.

In the future we should be able to remove more of these intermediate objects in the event system itself.

### DIFF
--- a/packages/react-dom/src/events/EventPluginRegistry.js
+++ b/packages/react-dom/src/events/EventPluginRegistry.js
@@ -13,11 +13,6 @@ import type {EventTypes} from './PluginModuleType';
 import invariant from 'shared/invariant';
 
 /**
- * Mapping from event name to dispatch config
- */
-export const eventNameDispatchConfigs = {};
-
-/**
  * Mapping from registration name to plugin module
  */
 export const registrationNames = {};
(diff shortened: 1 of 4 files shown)


packages/react-dom/src/events/EventPluginRegistry.js
```diff
@@ -40,15 +35,7 @@ function publishEventForPlugin(
   eventTypes: EventTypes,
   eventName: string,
 ): boolean {
-  invariant(
```
gaearon, line 43: While this invariant is useful to us in theory we have a similar one that would catch the same issue below anyway. We should also probably make that one a DEV-only warning.

packages/react-dom/src/test-utils/ReactTestUtils.js
```diff
@@ -571,7 +577,19 @@ function makeSimulator(eventType) {
         'a component instance. Pass the DOM node you wish to simulate the event on instead.',
     );
 
-    const dispatchConfig = eventNameDispatchConfigs[eventType];
+    // Reconstruct more or less what the original event system produced.
+    // We could remove this indirection here but we also don't plan to invest in Simulate anyway.
+    const dispatchConfig = {};
```
gaearon, line 582: While a real dispatch config would have more fields, they're not used by TestUtils.

packages/react-dom/src/test-utils/ReactTestUtils.js
```diff
@@ -549,6 +548,13 @@ function accumulateTwoPhaseDispatchesSingle(event) {
 
 const Simulate = {};
 
+const directDispatchEventTypes = new Set([
```
gaearon, line 551: Not sure this actually makes sense since those should have a special enter-leave traversal? But this is what the code is already doing.

gaearon:
Ignore the size alert, it's talking about TestUtils.

PR #19236, opened 2020-07-01
Make event config an implementation detail of each plugin

gaearon:
Each synthetic event used to have a `dispatchConfig` that contains the React names for this event. This makes the notion of event configs difficult to remove.

However, on the web there are always just two cases. Either our `dispatchConfig.registrationName` is the React event name. Or we have bubble event with that name and capture event with `Capture` prefix.

Taking advantage of this, I replaced `event.dispatchConfig` with `event._reactName` which is just a string. This will unlock removing the config metaprogramming because nothing else depends on the config format except the initial plugin registration.

Most of the work is in the second commit. The first commit just merges two variables that have the same purpose.

### DIFF
--- a/packages/react-dom/src/events/DOMModernPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMModernPluginEventSystem.js
@@ -16,10 +16,7 @@ import type {
   DispatchQueueItemPhase,
   DispatchQueueItemPhaseEntry,
 } from './PluginModuleType';
-import type {
-  ReactSyntheticEvent,
-  CustomDispatchConfig,
-} from './ReactSyntheticEventType';
+import type {ReactSyntheticEvent} from './ReactSyntheticEventType';
 import type {
   ElementListenerMap,
   ElementListenerMapEntry,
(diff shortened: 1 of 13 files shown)


packages/react-dom/src/events/plugins/ModernSimpleEventPlugin.js
```diff
@@ -46,41 +46,6 @@ import getEventCharCode from '../getEventCharCode';
 
 import {enableCreateEventHandleAPI} from 'shared/ReactFeatureFlags';
 
-// Only used in DEV for exhaustiveness validation.
```
gaearon, line 49: I got rid of the exhaustiveness check. Meh.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -639,11 +631,11 @@ export function accumulateTwoPhaseListeners(
   event: ReactSyntheticEvent,
   accumulateEventHandleListeners?: boolean,
 ): void {
-  const phasedRegistrationNames = event.dispatchConfig.phasedRegistrationNames;
+  const bubbled = event._reactName;
+  const captured = bubbled !== null ? bubbled + 'Capture' : null;
```
gaearon, line 635: We reconstruct the captured name here now.

packages/react-dom/src/events/plugins/ModernEnterLeaveEventPlugin.js
```diff
@@ -126,16 +126,16 @@ function extractEvents(
 
   if (topLevelType === TOP_MOUSE_OUT || topLevelType === TOP_MOUSE_OVER) {
     eventInterface = SyntheticMouseEvent;
```
trueadm, line 128: Could we maybe hoist the strings out to variables? It might help code size non gzip.

gaearon:
I can take over that one and update it.

packages/react-dom/src/events/plugins/ModernEnterLeaveEventPlugin.js
gaearon: Maybe! I want to do a few more simplifying passes before optimizing.

PR #19239, opened 2020-07-02
Add some type detection

sunnibfgi:


bvaughn:
cc @trueadm 

Not sure if `@flow` annotations were omitted from these files intentionally? Looks like some others are also missing the pragma (e.g. `SelectEventPlugin`)

trueadm:
Thanks for this!

PR #19244, opened 2020-07-02
Refine event registration + event signatures

trueadm:
Note: This PR was rebased on top of https://github.com/facebook/react/pull/19244.

In an attempt to break up the amount of changes in https://github.com/facebook/react/pull/19221, I decied to pull out the logic that was not strictly related to moving to the `capture` phase for events. This was mostly logic around handling event registration and the some of the clean up around the event workflow. This should hopefully make the other PR smaller and simpler to understand once it gets rebased on top of this PR.


packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -904,17 +918,16 @@ export function accumulateEventTargetListeners(
   if (eventListeners !== null) {
     const listenersArr = Array.from(eventListeners);
     const targetType = ((event.type: any): DOMTopLevelEventType);
-    const isCapturePhase = (event: any).eventPhase === 1;
```
gaearon, line 907: This seems like a significant change. Instead of reading this from the event, we now pass it from outside based on whether it's in a list. Can you please explain why this is a refactoring rather than a semantic change? And what does it accomplish, exactly?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -896,6 +909,7 @@ export function accumulateEventTargetListeners(
   dispatchQueue: DispatchQueue,
```
gaearon, line 909: So this is only used by the new EventHandle code? Can we rename this to `accumulateEventHandleListeners` or something like it? It took me a few reads to notice this wasn't actually used by the regular event system. The "event target" name looked generic to me.

packages/react-dom/src/events/plugins/ModernSelectEventPlugin.js
```diff
@@ -197,19 +163,17 @@ function extractEvents(
...
-    !isListeningToEvent(TOP_SELECTION_CHANGE, doc) ||
     // If we are handling TOP_SELECTION_CHANGE, then we don't need to
     // check for the other dependencies, as TOP_SELECTION_CHANGE is only
     // event attached from the onChange plugin and we don't expose an
     // onSelectionChange event from React.
-    (topLevelType !== TOP_SELECTION_CHANGE &&
-      !isListeningToEvents(rootTargetDependencies, targetContainer))
+    topLevelType !== TOP_SELECTION_CHANGE &&
+    !eventListenerMap.has('onSelect') &&
+    !eventListenerMap.has('onSelectCapture')
```
gaearon, line 176: Can you explain why this check is equivalent? It's not obvious to me.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -619,7 +632,7 @@ function createDispatchQueueItem(
   };
 }
 
-export function accumulateTwoPhaseListeners(
+export function accumulatePhaseListeners(
```
gaearon, line 635: Any particular reason for rename? Is it not always two phases?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: I guess there are two things here to note:

- This code is only used with `createEventHandle` logic, so this isn't that much of a significant change given we have no product code anywhere that uses this code path right now.
- It avoids us having to rely on `eventPhase` which can change if you add a listener to a node and the listener triggers on it (which would mean `eventPhase` is `2`, even if it were a capture phase). When we setup the listener we can add this information to the event flags as a detail, completely avoiding this check.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: Renamed to `accumulateEventHandleTargetListeners`.

packages/react-dom/src/events/plugins/ModernSelectEventPlugin.js
trueadm: Rather than going through all the dependencies one by one, we can instead check if we've registered `onSelect` or `onSelectCapture`. The dependencies are both the same, as we listen to the same events for both `onSelect` and `onSelectCapture`.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: Reverted this change.

PR #19251, opened 2020-07-03
Improve readability of isValidElementType

behnammodi:
`isValidElementType` is hard to understand and maintenance, so I tried to improve it

### DIFF
--- a/packages/shared/isValidElementType.js
+++ b/packages/shared/isValidElementType.js
@@ -28,28 +28,39 @@ import {
 } from 'shared/ReactSymbols';
 
 export default function isValidElementType(type: mixed) {
-  return (
-    typeof type === 'string' ||
-    typeof type === 'function' ||
-    // Note: its typeof might be other than 'symbol' or 'number' if it's a polyfill.
-    type === REACT_FRAGMENT_TYPE ||
-    type === REACT_PROFILER_TYPE ||
-    type === REACT_DEBUG_TRACING_MODE_TYPE ||
-    type === REACT_STRICT_MODE_TYPE ||
-    type === REACT_SUSPENSE_TYPE ||
-    type === REACT_SUSPENSE_LIST_TYPE ||
-    type === REACT_LEGACY_HIDDEN_TYPE ||
-    (typeof type === 'object' &&
-      type !== null &&
-      (type.$$typeof === REACT_LAZY_TYPE ||
-        type.$$typeof === REACT_MEMO_TYPE ||
-        type.$$typeof === REACT_PROVIDER_TYPE ||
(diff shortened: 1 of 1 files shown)


packages/shared/isValidElementType.js
```diff
@@ -28,28 +28,39 @@ import {
...
+  if (
+    [
+      REACT_FRAGMENT_TYPE,
+      REACT_PROFILER_TYPE,
+      REACT_DEBUG_TRACING_MODE_TYPE,
+      REACT_STRICT_MODE_TYPE,
+      REACT_SUSPENSE_TYPE,
+      REACT_SUSPENSE_LIST_TYPE,
+      REACT_LEGACY_HIDDEN_TYPE,
+    ].indexOf(type) !== -1
```
bvaughn, line 42: A switch/case would probably read just as nicely without the overhead of creating an new array and calling `indexOf` each time.

behnammodi: Thank you for review, yes I agree with you, I will update it by `switch/case`

packages/shared/isValidElementType.js
```diff
@@ -28,28 +28,35 @@ import {
 } from 'shared/ReactSymbols';
 
 export default function isValidElementType(type: mixed) {
-  return (
-    typeof type === 'string' ||
-    typeof type === 'function' ||
-    // Note: its typeof might be other than 'symbol' or 'number' if it's a polyfill.
```
rickhanlonii, line 34: Should we keep this comment?

behnammodi:
@rickhanlonii Nice, I will update it

behnammodi:
@rickhanlonii I updated it.

packages/shared/isValidElementType.js
```diff
@@ -28,28 +28,37 @@ import {
...
+  if (typeof type === 'string' || typeof type === 'function') return true;
```
gaearon, line 31: Nit: please use multiline `if` statements because otherwise it's very easy to miss the early `return`.

behnammodi: @gaearon Thanks for code review, I updated it.

behnammodi:
@gaearon Thanks, sure

packages/shared/isValidElementType.js
bvaughn: I think Dan was suggesting:

```js
if (typeof type === 'string' || typeof type === 'function') {
  return true;
}
```

Rather than:
```js
if (typeof type === 'string') return true;
if (typeof type === 'function') return true;
```

behnammodi: @bvaughn Thanks, I fixed it

PR #19253, opened 2020-07-03
removed duplicate code

behnammodi:
use `isElement` from `react-is` inside `isValidElement`


gaearon:
Thanks, but we don't want a dependency on this package from `react`, and we also don't want to accidentally inline some code from it during bundling. The duplication is intentional.

PR #19260, opened 2020-07-06
Fix ExhaustiveDeps ESLint rule throwing with optional chaining

### LINKED ISSUE #19243: Bug: Cannot read property 'references' of undefined in eslint-plugin-react-hooks v4.0.5
Certain code patterns using optional chaining syntax causes eslint-plugin-react-hooks to throw an error.

React version: 16.10.2

## Steps To Reproduce

1. Install eslint-plugin-react-hooks v4.0.5
2. Put this code in a file:

```tsx
import React, { useEffect } from 'react';

export const Repro = (props) => {
  const foo = {};

  const bar = () => ({
    pizza: foo.pizza,
    pasta: foo?.pasta,
  });

  useEffect(bar, []);

  return <div />;
};

```

## The current behavior

ESLint throws the following error:

```
TypeError: Cannot read property 'references' of undefined
Occurred while linting /path/to/repo/file.ts:102
    at /path/to/repo/node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js:1681:23
    at Set.forEach (<anonymous>)
    at visitFunctionWithDependencies (/path/to/repo/node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js:1672:29)
    at visitCallExpression (/path/to/repo/node_modules/eslint-plugin-react-hooks/cjs/eslint-plugin-react-hooks.development.js:886:19)
    at /path/to/repo/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (/path/to/repo/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/path/to/repo/node_modules/eslint/lib/linter/node-event-generator.js:254:26)
    at NodeEventGenerator.applySelectors (/path/to/repo/node_modules/eslint/lib/linter/node-event-generator.js:283:22)
    at NodeEventGenerator.enterNode (/path/to/repo/node_modules/eslint/lib/linter/node-event-generator.js:297:14)
```

## The expected behavior

ESLint does not throw an error.

This looks related to https://github.com/facebook/react/issues/19043 and https://github.com/facebook/react/pull/19062.

I've noticed that it does not throw if a number of slight variations are made to the code. The following do not throw:

```tsx
import React, { useEffect } from 'react';

export const Repro = (props) => {
  const foo = {};

  const bar = {
    pizza: foo.pizza,
    pasta: foo?.pasta,
  };

  useEffect(bar, []);

  return <div />;
};
```

```tsx
import React, { useEffect } from 'react';

export const Repro = (props) => {
  const foo = {};

  const bar = () => ({
    pizza: foo?.pizza,
    pasta: foo?.pasta,
  });

  useEffect(bar, []);

  return <div />;
};
```

```tsx
import React, { useEffect } from 'react';

export const Repro = (props) => {
  const foo = {};

  const bar = () => ({
    pizza: something.pizza,
    pasta: foo?.pasta,
  });

  useEffect(bar, []);

  return <div />;
};
```

cc @krailler

lencioni:
## Summary

Certain code patterns using optional chaining syntax causes
eslint-plugin-react-hooks to throw an error.

We can avoid the throw by adding some guards. I didn't read through the
code to understand how it works, I just added a guard to every place
where it threw, so maybe there is a better fix closer to the root cause
than what I have here.

In my test case, I noticed that the optional chaining that was used in
the code was not included in the suggestions description or output,
but it seems like it should be. This might make a nice future
improvement on top of this fix, so I left a TODO comment to that effect.

Fixes #19243

## Test Plan

```sh
yarn test --watch eslint-plugin
```

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -6628,6 +6628,44 @@ const testsTypescript = {
         },
       ],
     },
+    {
+      // https://github.com/facebook/react/issues/19243
+      code: normalizeIndent`
+        function MyComponent() {
+          const pizza = {};
+
+          useEffect(() => ({
+            crust: pizza.crust,
+            toppings: pizza?.toppings,
+          }), []);
+        }
+      `,
+      errors: [
+        {
+          message:
+            "React Hook useEffect has missing dependencies: 'pizza.crust' and 'pizza.toppings'. " +
+            'Either include them or remove the dependency array.',
+          suggestions: [
+            {
+              // TODO the description and suggestions should probably also
+              // preserve the optional chaining.
+              desc:
+                'Update the dependencies array to be: [pizza.crust, pizza.toppings]',
+              output: normalizeIndent`
+                function MyComponent() {
+                  const pizza = {};
+
+                  useEffect(() => ({
+                    crust: pizza.crust,
+                    toppings: pizza?.toppings,
+                  }), [pizza.crust, pizza.toppings]);
+                }
+              `,
+            },
+          ],
+        },
+      ],
+    },
   ],
 };
 
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -1016,11 +1016,11 @@ export default {
           // Is this a variable from top scope?
           const topScopeRef = componentScope.set.get(missingDep);
           const usedDep = dependencies.get(missingDep);
-          if (usedDep.references[0].resolved !== topScopeRef) {
+          if (usedDep && usedDep.references[0].resolved !== topScopeRef) {
             return;
           }
           // Is this a destructured prop?
-          const def = topScopeRef.defs[0];
+          const def = topScopeRef && topScopeRef.defs[0];
           if (def == null || def.name == null || def.type !== 'Parameter') {
             return;
           }
@@ -1062,7 +1062,7 @@ export default {
             return;
           }
           const usedDep = dependencies.get(missingDep);
-          const references = usedDep.references;
+          const references = usedDep ? usedDep.references : [];
           let id;
           let maybeCall;
           for (let i = 0; i < references.length; i++) {


gaearon:
>I didn't read through the code to understand how it works, I just added a guard to every place where it threw, so maybe there is a better fix closer to the root cause than what I have here.

I think at the very least we should look at *why* those items weren't resent in those Maps. Perhaps the key is wrong? Then we should fix the key.

lencioni:
It looks like there is a mismatch between the keys in the dependencies Map and the keys in missingDependencies. The Map includes the optional chaining in its keys (e.g. `pizza?.toppings`) while the missingDependencies array does not (e.g. `pizza.toppings`).

gaearon:
Can we think of a fix that would address the root of the problem?

lencioni:
Yep, looking into that now. In b0533fe33c it looks like the missingDependencies is normalized to remove the `?` bits, so I suppose we could do the same for the dependencies Map keys, which I think would just mean using the normalizedPath instead of path in `scanTreeRecursively`.

lencioni:
I've poked around a bit and I'm not sure what the best solution is yet. One way to fix this would be to normalize the paths in the dependencies Map like this maybe:

```diff
--- a/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
+++ b/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
@@ -561,7 +561,7 @@ export default {
             const isStatic =
               memoizedIsStaticKnownHookValue(resolved) ||
               memoizedIsFunctionWithoutCapturedValues(resolved);
-            dependencies.set(dependency, {
+            dependencies.set(dependency.replace(/\?/g, ''), {
               isStatic,
               references: [reference],
             });
```

but the downside is the error messages and suggestions will drop the optional chaining, which is not ideal.

I started to investigate fixing it on the other end (so that the missingDependencies Set includes the optional chaining) but that train ended up getting deeper than I have time to investigate at the moment (missingDependencies is built from `depTree` which comes from `createDepTree` which ends up normalizing the paths somehow--I think possibly by `getChildByKey`).

I can modify this PR to normalize out the optional chaining, or leave it as is. Anything more involved I'll have to set this aside for a while. Or perhaps @yanneves or @fredvollmer can offer some better insights, since it seems they've also been in this part of the codebase recently?

gaearon:
@bvaughn Any particular reason you merged this fix? It doesn't address the root of the problem and I'm worried we'll just get more confusing reports later on.

bvaughn:
@gaearon Yes? I merged it because it fixed a seemingly legitimate use case (test case attached) that broke without the fix.

If you disagree with landing the PR, feel free to revert...

bvaughn:
Wow. I feel like a clown. I did not see *any* of the above running discussion between the two of you. I'm not sure how I overlooked it but I somehow did.

I was looking through recent issues and PRs (because of oncall)- saw this, tested it, thought it looked reasonable, so I merged it.

I am sorry!

gaearon:
https://github.com/facebook/react/pull/19273

PR #19261, opened 2020-07-06
Effects list refactor

bvaughn:
This branch is phase one of a planned effects list refactor. It includes the following changes:
* Replace effects list usage with traversing the Fiber tree for before-mutation, mutation, and layout effects phases.
* Track pending deletions in a new `deletions` Fiber field.
* Track which effects exist below a Fiber using a new `subtreeTag` mask field to optimize traversal during commit phase.

All unit tests pass in this branch, although one or two required some slight tweaking.

This PR specifically does not include the following refactor pieces:
* Removing the effects list attributes (`firstEffect`, `lastEffect`, and `nextEffect`).
* Traversing the Fiber tree for passive effects.
* Any of the new hidden/offscreen features.
* Optimizing the new `subtreeTag` field.

Additional follow up work:
* We may be over-traversing during the commit phase now in the case where only some children are cloned. (In this case, we may end up continually traversing bailed out trees.) Maybe a better route would be to set a flag on a Fiber that explicitly marks when we bailed out (during the begin phase). Then we could just check this attribute to *know* without having to use the hacky `wasFiberCloned` heuristic.

---

Co-authored-by: Luna Ruan <luna@fb.com>

### DIFF
--- a/packages/react-native-renderer/src/__tests__/createReactNativeComponentClass-test.internal.js
+++ b/packages/react-native-renderer/src/__tests__/createReactNativeComponentClass-test.internal.js
@@ -45,12 +45,8 @@ describe('createReactNativeComponentClass', () => {
 
     expect(Text).not.toBe(View);
 
-    ReactNative.render(
-      <View>
-        <Text />
-      </View>,
-      1,
-    );
+    ReactNative.render(<Text />, 1);
+    ReactNative.render(<View />, 1);
   });
 
   it('should not allow viewConfigs with duplicate uiViewClassNames to be registered', () => {
--- a/packages/react-native-renderer/src/__tests__/createReactNativeComponentClass-test.internal.js
+++ b/packages/react-native-renderer/src/__tests__/createReactNativeComponentClass-test.internal.js
@@ -45,8 +45,12 @@ describe('createReactNativeComponentClass', () => {
 
     expect(Text).not.toBe(View);
 
-    ReactNative.render(<Text />, 1);
-    ReactNative.render(<View />, 1);
+    ReactNative.render(
+      <View>
+        <Text />
+      </View>,
+      1,
+    );
   });
 
   it('should not allow viewConfigs with duplicate uiViewClassNames to be registered', () => {
--- a/packages/react-reconciler/src/ReactFiber.new.js
+++ b/packages/react-reconciler/src/ReactFiber.new.js
@@ -144,6 +144,8 @@ function FiberNode(
 
   // Effects
   this.effectTag = NoEffect;
+  this.subtreeTag = NoEffect;
+  this.deletions = [];
   this.nextEffect = null;
 
   this.firstEffect = null;
@@ -287,6 +289,8 @@ export function createWorkInProgress(current: Fiber, pendingProps: any): Fiber {
     // We already have an alternate.
     // Reset the effect tag.
     workInProgress.effectTag = NoEffect;
+    workInProgress.subtreeTag = NoEffect;
+    workInProgress.deletions = [];
 
     // The effect list is no longer valid.
     workInProgress.nextEffect = null;
@@ -826,6 +830,8 @@ export function assignFiberPropertiesInDEV(
   target.dependencies = source.dependencies;
   target.mode = source.mode;
   target.effectTag = source.effectTag;
+  target.subtreeTag = source.subtreeTag;
+  target.deletions = source.deletions;
   target.nextEffect = source.nextEffect;
   target.firstEffect = source.firstEffect;
   target.lastEffect = source.lastEffect;
--- a/packages/react-reconciler/src/ReactFiber.new.js
+++ b/packages/react-reconciler/src/ReactFiber.new.js
@@ -145,7 +145,7 @@ function FiberNode(
   // Effects
   this.effectTag = NoEffect;
   this.subtreeTag = NoEffect;
-  this.deletions = [];
+  this.deletions = null;
   this.nextEffect = null;
 
   this.firstEffect = null;
@@ -290,7 +290,7 @@ export function createWorkInProgress(current: Fiber, pendingProps: any): Fiber {
     // Reset the effect tag.
     workInProgress.effectTag = NoEffect;
     workInProgress.subtreeTag = NoEffect;
-    workInProgress.deletions = [];
+    workInProgress.deletions = null;
 
     // The effect list is no longer valid.
     workInProgress.nextEffect = null;
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.new.js
@@ -1066,9 +1066,7 @@ function completeWork(
                   // TODO (effects) This probably isn't the best approach. Discuss with Brian
                   let child = workInProgress.child;
                   while (child !== null) {
-                    if (child.deletions.length > 0) {
-                      child.deletions = [];
-                    }
+                    child.deletions = null;
                     child = child.sibling;
                   }
                 }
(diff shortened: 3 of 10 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2101,145 +2070,305 @@ function commitRootImpl(root, renderPriorityLevel) {
...
-    if (effectTag & Ref) {
-      const current = nextEffect.alternate;
-      if (current !== null) {
-        commitDetachRef(current);
-      }
+function commitMutationEffects(
+  fiber: Fiber,
+  root: FiberRoot,
+  renderPriorityLevel,
+) {
```
sebmarkbage, line 2168: What if we called `commitMutationEffectsDeletions(fiber.deletions, ...)` here instead of in `commitMutationEffectsImpl`?

lunaruan: We talked about this! I think the reason why we didn't proceed was because refactoring it was a bit annoying and Brian wanted to get your response to the approach before we committed to doing this?

bvaughn: Having a "duh" moment now for not coming to this conclusion earlier.

Good suggestion. Thanks!

packages/react-native-renderer/src/__tests__/createReactNativeComponentClass-test.internal.js
```diff
@@ -45,8 +45,12 @@ describe('createReactNativeComponentClass', () => {
 
     expect(Text).not.toBe(View);
 
-    ReactNative.render(<Text />, 1);
-    ReactNative.render(<View />, 1);
+    ReactNative.render(
```
gaearon, line 48: Why?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1772,24 +1780,31 @@ function resetChildLanes(completedWork: Fiber) {
...
       newChildLanes = mergeLanes(
         newChildLanes,
         mergeLanes(child.lanes, child.childLanes),
       );
-      if (shouldBubbleActualDurations) {
+
+      subtreeTag |= child.subtreeTag;
+      subtreeTag |= child.effectTag & HostEffectMask;
+      if (child.deletions.length > 0) {
+        subtreeTag |= Deletion;
```
sebmarkbage, line 1793: Can we add this tag where we do `child.deletions.push(...)` so we only do this work when we hit a deletion in reconciliation instead of checking this for every Fiber in a render?

packages/react-reconciler/src/ReactFiber.new.js
```diff
@@ -144,6 +144,8 @@ function FiberNode(
 
   // Effects
   this.effectTag = NoEffect;
+  this.subtreeTag = NoEffect;
+  this.deletions = [];
```
sebmarkbage, line 148: Can we make this nullable instead of allocating an array?

packages/react-native-renderer/src/__tests__/createReactNativeComponentClass-test.internal.js
bvaughn: Ah- because of the deletions ordering thing. This was a use case (similar to document -> body) that failed. But I think...now it does not fail anymore :) so we can revert this change.

packages/react-reconciler/src/ReactFiber.new.js
bvaughn: Done.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
bvaughn: Maybe. Naively doing this breaks ~150 tests, but it seems like it should be doable. We'll look into it and get back to you.

bvaughn: Okay, yup. This is doable too. I'll leave TODO comments to rename the effect tag to something more appropriate once we've merged new->old.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1764,6 +1765,12 @@ function resetChildLanes(completedWork: Fiber) {
   }
 
   let newChildLanes = NoLanes;
+  let subtreeTag = NoEffect;
+  let childrenDidNotComplete = false;
+
+  const wasFiberCloned =
+    completedWork.alternate === null ||
+    completedWork.child !== completedWork.alternate.child;
```
bvaughn, line 1773: I believe I see one potential over-traversal problem. Although I'm not sure if it affects this branch, it does affect the follow up branch (passive effects): If a component gets a state update, but its siblings don't, we still end up traversing the sibling path when running passive effects which leads us to over-count commit times. Fortunately the Profiler tests are pretty good and caught this.

Maybe using an explicit flag to mark bailed-out-fibers would (like we talked about on Tuesday) could help with that too.

packages/react-reconciler/src/ReactFiberCompleteWork.new.js
```diff
@@ -1062,6 +1062,13 @@ function completeWork(
                 // Reset the effect list before doing the second pass since that's now invalid.
                 if (renderState.lastEffect === null) {
                   workInProgress.firstEffect = null;
+                  workInProgress.subtreeTag = NoEffect;
+                  // TODO (effects) This probably isn't the best approach. Discuss with Brian
```
bvaughn, line 1066: @lunaruan You left a TODO to talk to me here. Still relevant?

lunaruan: Nope we're good :D

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
acdlite: It should be fine that the siblings are traversed in that case as long as they are also cloned, because if they are cloned then we should be able to clear the effect tags in `createWorkInProgressFiber`.

I do think the explicit flag approach is worth approaching, though.

bvaughn: Gotcha. I'm seeing a few failures that are being caused by over-traversal during passive effects phase, so something doesn't seem to be getting reset correctly. Will keep digging in.

Tried the explicit flag approach initially and it failed more things but I'm probably missing some of the bailout spots.

bvaughn: Explicit bailout flag approach has been implemented in 2nd part PR, #19322

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2172,160 +2122,310 @@ function commitRootImpl(root, renderPriorityLevel) {
...
+    try {
+      commitBeforeMutationEffectsImpl(fiber);
+    } catch (error) {
+      captureCommitPhaseError(fiber, error);
+    }
+  }
 
-      commitBeforeMutationEffectOnFiber(current, nextEffect);
+  if (fiber.sibling !== null) {
+    commitBeforeMutationEffects(fiber.sibling);
```
sebmarkbage, line 2155: It's unusually to have very deep trees but it's not that unusual to have a few hundred of items in a list. This could cause a stack overflow in that case. We should probably keep the while loop for siblings but not for the recursive effects.

bvaughn:
Already merged via #19322

PR #19263, opened 2020-07-06
Fork ReactFiberLane (new and old)

acdlite:
This wasn't previously forked because it used to only exist in the new one.


packages/react-reconciler/src/ReactFiberOffscreenComponent.js
```diff
@@ -8,7 +8,7 @@
  */
 
 import type {ReactNodeList} from 'shared/ReactTypes';
-import type {Lanes} from './ReactFiberLane';
+import type {Lanes} from './ReactFiberLane.old';
```
rickhanlonii, line 11: Should this be using the old?

acdlite:
Never mind, can't both fork the module and keep `Lanes` opaque without either forking the names of the Fiber fields or doing something else clever. Probably not worth it. Closing for now.

PR #19264, opened 2020-07-06
Move Scope API ref resolution to mutation phase

trueadm:
To unblock internal progression on replacing React Flare, let's instead move resolution of Scope API refs to the mutation phase, so we can attach event listeners to scopes before the layout phase begins.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -2219,8 +2219,12 @@ function commitMutationEffects(root: FiberRoot, renderPriorityLevel) {
       if (current !== null) {
         commitDetachRef(current);
       }
-      if (enableScopeAPI && nextEffect.tag === ScopeComponent) {
-        commitAttachRef(nextEffect);
+      if (enableScopeAPI) {
+        // This is a temporary solution that allows us to transition away
+        // from React Flare on www.
+        if (nextEffect.tag === ScopeComponent) {
+          commitAttachRef(nextEffect);
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2163,6 +2165,9 @@ function commitMutationEffects(root: FiberRoot, renderPriorityLevel) {
       if (current !== null) {
         commitDetachRef(current);
       }
+      if (enableScopeAPI && nextEffect.tag === ScopeComponent) {
+        commitAttachRef(nextEffect);
+      }
```
bvaughn, line 2170: Can we separate the feature flag check here?

```suggestion
      if (enableScopeAPI) {
        if (nextEffect.tag === ScopeComponent) {
          commitAttachRef(nextEffect);
        }
      }
```

I find this to be more readable for feature flags.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2163,6 +2165,9 @@ function commitMutationEffects(root: FiberRoot, renderPriorityLevel) {
       if (current !== null) {
         commitDetachRef(current);
       }
+      if (enableScopeAPI && nextEffect.tag === ScopeComponent) {
```
bvaughn, line 2168: Could we add some TODO comments above these changes since they are only intended as as temporary approach until we have a more robust API design?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2233,7 +2238,10 @@ function commitLayoutEffects(root: FiberRoot, committedLanes: Lanes) {
       commitLayoutEffectOnFiber(root, current, nextEffect, committedLanes);
     }
 
-    if (effectTag & Ref) {
+    if (
+      effectTag & Ref &&
+      (!enableScopeAPI || nextEffect.tag !== ScopeComponent)
+    ) {
```
bvaughn, line 2244: IMO I think this would be a bit more readable if the feature flag check was separated, e.g.
```suggestion
if (enableScopeAPI) {
  if ((effectTag & Ref) !== NoEffect && nextEffect.tag !== ScopeComponent) {
    commitAttachRef(nextEffect);
  }
} else {
  if ((effectTag & Ref) !== NoEffect) {
    commitAttachRef(nextEffect);
  }
}
```

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2217,6 +2219,13 @@ function commitMutationEffects(root: FiberRoot, renderPriorityLevel) {
       if (current !== null) {
         commitDetachRef(current);
       }
+      if (enableScopeAPI) {
+        // This is a temporary solution that allows us to transition away
```
bvaughn, line 2223: super nit: Could we prefix these with "TODO" so they stand out more than a regular comment?

PR #19265, opened 2020-07-06
Handle test-cli failure case for CI

trueadm:
We should handle the failure case for the test CLI for when a process exits with a code that is `1`.

### DIFF
--- a/scripts/jest/jest-cli.js
+++ b/scripts/jest/jest-cli.js
@@ -317,7 +317,17 @@ function main() {
   }
 
   // Run Jest.
-  spawn('node', args, {stdio: 'inherit', env: {...envars, ...process.env}});
+  const jest = spawn('node', args, {
+    stdio: 'inherit',
+    env: {...envars, ...process.env},
+  });
+  // Ensure we close our process when we get a failure case.
+  jest.on('close', code => {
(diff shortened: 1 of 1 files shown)


scripts/jest/jest-cli.js
```diff
@@ -317,7 +317,17 @@ function main() {
   }
 
   // Run Jest.
-  spawn('node', args, {stdio: 'inherit', env: {...envars, ...process.env}});
+  const process = spawn('node', args, {
```
rickhanlonii, line 320: Will this override the node `process` global?

scripts/jest/jest-cli.js
```diff
@@ -317,7 +317,17 @@ function main() {
...
+    stdio: 'inherit',
+    env: {...envars, ...process.env},
+  });
+  // Ensure we close our process when we get a failure case.
+  process.on('close', code => {
+    // If we get a failure case then stop the main process.
```
rickhanlonii, line 326: This reads as though it will stop on the first failure, but it doesn't stop until Jest completes:

```suggestion
    // Forward the exit code from the Jest process.
```

bvaughn:
Thanks for this PR!

PR #19266, opened 2020-07-06
Fix test suite in CI

rickhanlonii:
## Summary

Forward the child process exit code so that failures show up in Circle.

## Test Plan

Pushed a bad commit, Circle should fail and we'll revert that commit before landing.


rickhanlonii:
Dupe of https://github.com/facebook/react/pull/19265

PR #19271, opened 2020-07-07
Update useEvent implementation

trueadm:
This updates the `useEvent` implementation within this repo to match that of our internal version, for consistency.

### DIFF
--- a/packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
+++ b/packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
@@ -11,7 +11,7 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 
 const {useLayoutEffect, useRef} = React;
-const {unstable_createEventHandle: createEventHandle} = ReactDOM;
+const {unstable_createEventHandle} = ReactDOM;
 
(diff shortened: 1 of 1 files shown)


bvaughn:
> This updates the useEvent implementation within this repo to match that of our internal version, for consistency.

I'm confused. This looks less like the internal version than before.

packages/react-interactions/events/src/dom/create-event-handle/useEvent.js
```diff
@@ -29,14 +29,12 @@ export default function useEvent(
     priority?: 0 | 1 | 2,
   |},
 ): UseEventHandle {
-  const handleRef = useRef(null);
-  let setListener;
-  let clears;
-  let useEventHandle;
+  const handleRef = useRef<UseEventHandle | null>(null);
```
bvaughn, line 32: Any reason we use generics for the event and target types internally, but not in this repo?

trueadm: We use Scopes internally, which means that the target needs be generic to cover them too. On the web, we can stick with EventTarget, so I could simplify it on githib. Good question though!

PR #19273, opened 2020-07-07
[ESLint] Consistently treat optional chaining as regular chaining

gaearon:
We've merged a bunch of PRs which added support for optional chaining to the ESLint rule about dependencies (https://github.com/facebook/react/pull/18820, https://github.com/facebook/react/pull/19008, https://github.com/facebook/react/pull/19062, https://github.com/facebook/react/pull/19260). However there was no single cohesive strategy behind them, and they led to crashes and edge cases because they attacked the symptoms rather than the root cause (the algorithm wasn’t designed to deal with optional chaining). There was no attempt to rethink the algorithm so the interim solutions have gradually made it broken and inconsistent. However, we did get new tests from them so they were still useful PRs. We just need to rethink our strategy.

As a first step I want to simplify this by getting back to where we were and removing the code that breaks algorithm’s assumptions. In particular, I am treating all `foo?.bar` as `foo.bar` from the dependency tracking perspective. The downside is that our error messages and fix suggestions will suggest regular chaining (but they will also be okay with optional chaining — so you can always fix them manually). The upside is that we can reason about what’s happening again because the inconsistencies are removed, and the algorithm only deals with regular property paths, just like before.

This fixes a bunch of bugs, for which I’ve added regression tests.

As a follow-up to this, I think we should be able to figure out a simple solution that will preserve optional chaining where it makes sense in the messages and suggestions. However, I think this first step is still useful to get in as a base.

**Best reviewed as individual commits**

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -6628,6 +6628,44 @@ const testsTypescript = {
         },
       ],
     },
+    {
+      // https://github.com/facebook/react/issues/19243
+      code: normalizeIndent`
+        function MyComponent() {
+          const pizza = {};
+
+          useEffect(() => ({
+            crust: pizza.crust,
+            toppings: pizza?.toppings,
+          }), []);
+        }
+      `,
+      errors: [
+        {
+          message:
+            "React Hook useEffect has missing dependencies: 'pizza.crust' and 'pizza.toppings'. " +
+            'Either include them or remove the dependency array.',
+          suggestions: [
+            {
+              // TODO the description and suggestions should probably also
+              // preserve the optional chaining.
+              desc:
+                'Update the dependencies array to be: [pizza.crust, pizza.toppings]',
+              output: normalizeIndent`
+                function MyComponent() {
+                  const pizza = {};
+
+                  useEffect(() => ({
+                    crust: pizza.crust,
+                    toppings: pizza?.toppings,
+                  }), [pizza.crust, pizza.toppings]);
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -1264,17 +1302,16 @@ const tests = {
       errors: [
         {
           message:
-            "React Hook useCallback has a missing dependency: 'props.foo?.toString'. " +
+            "React Hook useCallback has a missing dependency: 'props.foo.toString'. " +
```
gaearon, line 1305: This is the regression. But it's acceptable IMO because it only affects the message and the autofix. All other regression tests that have been added are passing.

gaearon: In fact this test is also kind of wrong because it suggests `props.foo?.toString` should be added — whereas in practice it should have suggested `props.foo`. So this is something we need to fix too.

gaearon: Pushed a fix to this test.

bvaughn:
Nice cleanup!

radko93:
Hey! It would be great to have a new version published with this, right now npm v4.0.5 breaks linters in projects.

gaearon:
@radko93 Thanks, we're aware from the reports. :-) I am working on a proper fix.

gaearon:
https://github.com/facebook/react/pull/19275

gaearon:
`eslint-plugin-react-hooks@4.0.6` is out with fixes.

PR #19275, opened 2020-07-07
[ESLint] Handle optional member chains

gaearon:
Follow-up to https://github.com/facebook/react/pull/19273.

Our algorithm treats `a.b.c.d` as unique paths in a many places and those parts should work the same way regardless of whether these are optional or required members. So in https://github.com/facebook/react/pull/19273, I removed `?.` from the internal representation.

However, ideally we still want to print `?.` where needed **if all usages of that member in the user code used optional chaining**. For example, if we use `pizza?.crust` and `pizza?.crust.density`, then we should suggest `pizza?.crust` in deps, but if one of them didn’t have `?`, then we should suggest `pizza.crust` instead. Because then, at least one of them is required anyway. None of this should have any effects on the dependency logic — both variations, if typed by user, should satisfy the constraint.

The implementation is split in two stages.

It intentionally doesn’t interfere with the main algorithm that checks the dependencies. Instead, when we initially generate member paths, we also fill a Map. In that Map, we mark for each deep path whether it was a required or an optional one. By the time we have collected all dependencies, the Map will contain a mapping from the key path to a Boolean: `true` if all usages were optional, and `false` if some of them were required.

Finally, when we're ready to print our suggestions and/or fix the code, we will use that Map to pretty-print our paths. We will consult it to see if each segment should be concatenated using `?.` or `.`. If at least one usage was required, or if that path was not found in the code, then we will default to `.` as a default path separator. So if you don't use `?.` you won’t see it.

This adds a few tests and also fixes a bunch of existing tests to use better output.

**This is split in two commits.** The first one is just a bunch of renames because "optional" clashed with another concept we previously called the same way. Review commits individually.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -1384,17 +1384,16 @@ const tests = {
       errors: [
         {
           message:
-            // TODO: Ideally this would suggest props.foo?.bar.baz instead.
-            "React Hook useCallback has a missing dependency: 'props.foo.bar.baz'. " +
+            "React Hook useCallback has a missing dependency: 'props.foo?.bar.baz'. " +
             'Either include it or remove the dependency array.',
           suggestions: [
             {
-              desc: 'Update the dependencies array to be: [props.foo.bar.baz]',
+              desc: 'Update the dependencies array to be: [props.foo?.bar.baz]',
               output: normalizeIndent`
                 function MyComponent(props) {
                   useCallback(() => {
                     console.log(props.foo?.bar.baz);
-                  }, [props.foo.bar.baz]);
+                  }, [props.foo?.bar.baz]);
                 }
               `,
             },
@@ -1413,17 +1412,17 @@ const tests = {
       errors: [
         {
           message:
-            // TODO: Ideally this would suggest props.foo?.bar?.baz instead.
-            "React Hook useCallback has a missing dependency: 'props.foo.bar.baz'. " +
+            "React Hook useCallback has a missing dependency: 'props.foo?.bar?.baz'. " +
             'Either include it or remove the dependency array.',
           suggestions: [
             {
-              desc: 'Update the dependencies array to be: [props.foo.bar.baz]',
+              desc:
+                'Update the dependencies array to be: [props.foo?.bar?.baz]',
               output: normalizeIndent`
                 function MyComponent(props) {
                   useCallback(() => {
                     console.log(props.foo?.bar?.baz);
-                  }, [props.foo.bar.baz]);
+                  }, [props.foo?.bar?.baz]);
                 }
               `,
             },
@@ -1442,17 +1441,16 @@ const tests = {
       errors: [
         {
           message:
-            // TODO: Ideally this would suggest props.foo?.bar instead.
-            "React Hook useCallback has a missing dependency: 'props.foo.bar'. " +
+            "React Hook useCallback has a missing dependency: 'props.foo?.bar'. " +
             'Either include it or remove the dependency array.',
           suggestions: [
             {
-              desc: 'Update the dependencies array to be: [props.foo.bar]',
+              desc: 'Update the dependencies array to be: [props.foo?.bar]',
               output: normalizeIndent`
                 function MyComponent(props) {
                   useCallback(() => {
                     console.log(props.foo?.bar.toString());
-                  }, [props.foo.bar]);
+                  }, [props.foo?.bar]);
                 }
               `,
             },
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -7000,22 +6997,122 @@ const testsTypescript = {
...
-            "React Hook useEffect has missing dependencies: 'pizza.crust' and 'pizza.toppings'. " +
+            "React Hook useEffect has missing dependencies: 'pizza.crust' and 'pizza?.toppings'. " +
             'Either include them or remove the dependency array.',
           suggestions: [
             {
-              // TODO the description and suggestions should probably also
-              // preserve the optional chaining.
               desc:
-                'Update the dependencies array to be: [pizza.crust, pizza.toppings]',
+                'Update the dependencies array to be: [pizza.crust, pizza?.toppings]',
```
gaearon, line 7005: Note: technically this could be further improved to not add `?` at all because another dep didn't use it. However, I don't think it matters because at least this doesn't cause a crash. Additionally, this points your attention to unnecessary `?` inside the function — since presumably if data is immutable, you shouldn't need it anyway.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -1466,25 +1489,47 @@ function getDependency(node) {
 /**
  * Assuming () means the passed node.
  * (foo) -> 'foo'
- * foo.(bar) -> 'foo.bar'
- * foo.bar.(baz) -> 'foo.bar.baz'
+ * foo(.)bar -> 'foo.bar'
+ * foo.bar(.)baz -> 'foo.bar.baz'
```
gaearon, line 1493: This comment didn't make sense so I fixed it.

ljharb:
>  Because then, at least one of them is required anyway. 

That's not how optional chaining works, if I'm reading it right. `a?.b.c` will short-circuit, and not evaluate `.c`, if `a` is nullish.

gaearon:
>That's not how optional chaining works, if I'm reading the OP right. a?.b.c will short-circuit, and not evaluate .c, if a is nullish (and will throw if a is non-nullish and a.b is nullish)

Hmm. Maybe I wasn't clear or maybe I'm missing something.

I'm referring to this scenario:

```js
useEffect(() => {
  console.log(a?.b.x)
  console.log(a?.b.x)
})
```

If we were to auto-fill dependencies, we'd want to suggest `a?.b.x`. Because suggesting `a.b.x` would lead to a crash.

However, with this scenario:

```js
useEffect(() => {
  console.log(a?.b.x)
  console.log(a.b.x)
})
```

In this case, suggesting `a?.b.x` as a dep may be a bit misleading because we *do* use `a.b.x` inside. So it is better to call user’s attention to that by suggesting `a.b.x` as a dep. They can still edit it if they want to.

packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js
```diff
@@ -1466,25 +1489,47 @@ function getDependency(node) {
...
+    const property = analyzePropertyChain(node.property, null);
+    const result = `${object}.${property}`;
+    if (optionalChains) {
+      // Note: OptionalMemberExpression doesn't necessarily mean this node is optional.
+      // It just means there is an optional member somewhere inside.
+      // This particular node might still represent a required member, so check .optional field.
+      if (node.optional) {
+        // We only want to consider it optional if *all* usages were optional.
+        if (!optionalChains.has(result)) {
+          // Mark as optional.
```
rickhanlonii, line 1524:
```suggestion
          // Mark as (maybe) optional. If there's a required usage, this will be overridden.
```

ljharb:
@gaearon as long as there's no getters in the chain that might change the nullishness of properties on later evaluation, and no assignments or mutations in between the two statements, you're right :-)

gaearon:
Yeah. This rule in general assumes that render-scope objects you read inside Hook callbacks are immutable POJOs or treated as such. Otherwise dependencies don't work as a concept in general.

packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -7000,22 +6997,122 @@ const testsTypescript = {
...
+          useEffect(() => ({
+            crust: pizza?.crust,
+            density: pizza.crust.density,
+          }), []);
+        }
+      `,
+      errors: [
+        {
+          message:
+            "React Hook useEffect has a missing dependency: 'pizza.crust'. " +
```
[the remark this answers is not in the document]
gaearon: Meh. I don't think this is useful enough to clutter the message.

PR #19276, opened 2020-07-08
Remove event constants

gaearon:
I changed these to a string enum and tightened the types where needed to make them checked.
Also renamed a few things.

This shouldn't affect the output because GCC was always smart enough.


packages/react-dom/src/events/DOMEventProperties.js
```diff
@@ -41,104 +122,108 @@ const eventPriorities = new Map();
...
-  DOMTopLevelEventTypes.TOP_PLAYING, 'playing',
-  DOMTopLevelEventTypes.TOP_PROGRESS, 'progress',
-  DOMTopLevelEventTypes.TOP_SEEKING, 'seeking',
-  DOMTopLevelEventTypes.TOP_STALLED, 'stalled',
-  DOMTopLevelEventTypes.TOP_SUSPEND, 'suspend',
-  DOMTopLevelEventTypes.TOP_TIME_UPDATE, 'timeUpdate',
-  DOMTopLevelEventTypes.TOP_TRANSITION_END, 'transitionEnd',
-  DOMTopLevelEventTypes.TOP_WAITING, 'waiting',
+  'abort', 'abort',
+  ANIMATION_END, 'animationEnd',
```
trueadm, line 203: Are these intentionally not changed over?

packages/react-dom/src/events/DOMEventProperties.js
```diff
@@ -41,104 +122,108 @@ const eventPriorities = new Map();
...
+  'loadedmetadata', 'loadedMetadata',
+  'loadstart', 'loadStart',
+  'lostpointercapture', 'lostPointerCapture',
+  'playing', 'playing',
+  'progress', 'progress',
+  'seeking', 'seeking',
+  'stalled', 'stalled',
+  'suspend', 'suspend',
+  'timeupdate', 'timeUpdate',
+  TRANSITION_END, 'transitionEnd',
```
trueadm, line 225: Same for this one.

packages/react-dom/src/events/DOMEventProperties.js
gaearon: It’s calculated dynamically like before. Vendor prefixed. (I do wonder if we can drop that though.)

PR #19278, opened 2020-07-08
Remove capturePhaseEvents and separate events by bubbling

trueadm:
This PR does two things. Primarily it removes `capturePhaseEvents` entirely and now we properly attach events to the phase specified by the user in all case. For events like `onSubmit`, media events, etc, where we previously used the capture phase – we move these back to being attached to their target. This means we don't use event delegation for these events, but that's actually more compatible than the current design.

I also took the opportunity to simplify much of how the event registering flow works, now that we don't have as many forks in behavior to deal with. That also gave me an opportunity to tackle some of Dan's recent feedback about the confusing naming, which has also been confusing to me in some cases too!

Note: we also remove `close` and `cancel` from event replaying.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -7,7 +7,6 @@
  * @flow
  */
 
-import type {DOMTopLevelEventType} from '../events/TopLevelEventTypes';
 import type {ElementListenerMapEntry} from '../client/ReactDOMComponentTree';
 
 import {
@@ -88,8 +87,8 @@ import {
 } from 'shared/ReactFeatureFlags';
 import {
   listenToReactEvent,
-  listenToNativeEvent,
   mediaEventTypes,
+  listenToNonDelegatedEvent,
 } from '../events/DOMModernPluginEventSystem';
 import {getEventListenerMap} from './ReactDOMComponentTree';
 import {
@@ -272,13 +271,6 @@ if (__DEV__) {
     testElement.innerHTML = html;
     return testElement.innerHTML;
   };
-}
-
-function listenToNonDelegatedEvent(
-  topLevelType: DOMTopLevelEventType,
-  targetElement: Element,
-): void {
-  listenToNativeEvent(topLevelType, false, targetElement, targetElement);
 }
 
 export function ensureListeningTo(
@@ -577,9 +569,6 @@ export function setInitialProperties(
       // listeners still fire for error and load events.
       listenToNonDelegatedEvent(TOP_ERROR, domElement);
       listenToNonDelegatedEvent(TOP_LOAD, domElement);
-      props = rawProps;
-      break;
-    case 'form':
       props = rawProps;
       break;
     case 'details':
--- a/packages/react-dom/src/client/ReactDOMComponent.js
+++ b/packages/react-dom/src/client/ReactDOMComponent.js
@@ -89,6 +89,7 @@ import {
 import {
   listenToReactEvent,
   listenToNativeEvent,
+  mediaEventTypes,
 } from '../events/DOMModernPluginEventSystem';
 import {getEventListenerMap} from './ReactDOMComponentTree';
 import {
@@ -549,17 +550,22 @@ export function setInitialProperties(
     case 'iframe':
     case 'object':
     case 'embed':
-      // We listen this event in case to ensure emulated bubble
+      // We listen to this event in case to ensure emulated bubble
       // listeners still fire for the load event.
       listenToNonDelegatedEvent(TOP_LOAD, domElement);
       props = rawProps;
       break;
     case 'video':
     case 'audio':
+      // We listen to these events in case to ensure emulated bubble
+      // listeners still fire for all the media events.
+      for (let i = 0; i < mediaEventTypes.length; i++) {
+        listenToNonDelegatedEvent(mediaEventTypes[i], domElement);
+      }
       props = rawProps;
       break;
     case 'source':
-      // We listen this event in case to ensure emulated bubble
+      // We listen to this event in case to ensure emulated bubble
       // listeners still fire for the error event.
       listenToNonDelegatedEvent(TOP_ERROR, domElement);
       props = rawProps;
@@ -577,7 +583,7 @@ export function setInitialProperties(
       props = rawProps;
       break;
     case 'details':
-      // We listen this event in case to ensure emulated bubble
+      // We listen to this event in case to ensure emulated bubble
       // listeners still fire for the toggle event.
       listenToNonDelegatedEvent(TOP_TOGGLE, domElement);
       props = rawProps;
@@ -585,7 +591,7 @@ export function setInitialProperties(
     case 'input':
       ReactDOMInputInitWrapperState(domElement, rawProps);
       props = ReactDOMInputGetHostProps(domElement, rawProps);
-      // We listen this event in case to ensure emulated bubble
+      // We listen to this event in case to ensure emulated bubble
       // listeners still fire for the invalid event.
       listenToNonDelegatedEvent(TOP_INVALID, domElement);
       // For controlled components we always need to ensure we're listening
@@ -599,7 +605,7 @@ export function setInitialProperties(
     case 'select':
       ReactDOMSelectInitWrapperState(domElement, rawProps);
       props = ReactDOMSelectGetHostProps(domElement, rawProps);
-      // We listen this event in case to ensure emulated bubble
+      // We listen to this event in case to ensure emulated bubble
       // listeners still fire for the invalid event.
       listenToNonDelegatedEvent(TOP_INVALID, domElement);
       // For controlled components we always need to ensure we're listening
@@ -609,7 +615,7 @@ export function setInitialProperties(
     case 'textarea':
       ReactDOMTextareaInitWrapperState(domElement, rawProps);
       props = ReactDOMTextareaGetHostProps(domElement, rawProps);
-      // We listen this event in case to ensure emulated bubble
+      // We listen to this event in case to ensure emulated bubble
       // listeners still fire for the invalid event.
       listenToNonDelegatedEvent(TOP_INVALID, domElement);
       // For controlled components we always need to ensure we're listening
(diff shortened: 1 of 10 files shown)


trueadm:
This PR is currently blocked by the events for replaying. We need to decide on what approach to take, either:

https://github.com/facebook/react/pull/19300 or https://github.com/facebook/react/pull/19285

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -266,6 +266,7 @@ if (__DEV__) {
 export function ensureListeningTo(
   rootContainerInstance: Element | Node,
   reactPropEvent: string,
+  targetElement?: Element,
```
gaearon, line 269: Optional arguments can be pretty confusing to follow. Can we make it nullable but always pass it?

packages/react-dom/src/client/ReactDOMEventHandle.js
```diff
@@ -155,15 +154,16 @@ function registerReactDOMEvent(
...
+    // These are valid event targets, but they are also
+    // non-managed React nodes.
     listenToNativeEvent(
       topLevelType,
-      eventTarget,
-      listenerMap,
-      PLUGIN_EVENT_SYSTEM | IS_EVENT_HANDLE_NON_MANAGED_NODE,
       isCapturePhaseListener,
+      eventTarget,
+      undefined,
```
gaearon, line 163: nit: we usually use null for absence so it's clear it's intentional

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -155,8 +152,11 @@ function extractEvents(
     targetContainer,
   );
   const shouldProcessPolyfillPlugins =
-    (eventSystemFlags & IS_CAPTURE_PHASE) === 0 ||
-    capturePhaseEvents.has(topLevelType);
+    (eventSystemFlags & IS_EVENT_HANDLE_NON_MANAGED_NODE) === 0 &&
+    (eventSystemFlags & IS_NON_DELEGATED) === 0 &&
+    ((eventSystemFlags & IS_CAPTURE_PHASE) === 0 ||
+      topLevelType === TOP_FOCUS ||
```
gaearon, line 158: This needs to be rebased on https://github.com/facebook/react/pull/19186 and changed?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -155,8 +152,11 @@ function extractEvents(
     targetContainer,
   );
   const shouldProcessPolyfillPlugins =
-    (eventSystemFlags & IS_CAPTURE_PHASE) === 0 ||
-    capturePhaseEvents.has(topLevelType);
+    (eventSystemFlags & IS_EVENT_HANDLE_NON_MANAGED_NODE) === 0 &&
+    (eventSystemFlags & IS_NON_DELEGATED) === 0 &&
+    ((eventSystemFlags & IS_CAPTURE_PHASE) === 0 ||
```
gaearon, line 157: You could unify these flags first and then do one `&` operation.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -214,23 +214,22 @@ function extractEvents(
   }
 }
 
-export const capturePhaseEvents: Set<DOMTopLevelEventType> = new Set([
-  TOP_FOCUS,
-  TOP_BLUR,
+// We should not delegate these events to the container, but rather
+// set them on the actual target element itself.
```
gaearon, line 218: Maybe add that it's because these events don't consistently bubble in the DOM.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -214,23 +214,22 @@ function extractEvents(
...
+export const nonDelegatedEvents: Set<DOMTopLevelEventType> = new Set([
   TOP_SCROLL,
   TOP_LOAD,
   TOP_ABORT,
   TOP_CANCEL,
   TOP_CLOSE,
-  TOP_INVALID,
-  TOP_ABORT,
```
gaearon, line 226: This confused me for a moment. But they're actually both still in the list. INVALID moved down and ABORT was listed twice.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -249,12 +248,8 @@ export const capturePhaseEvents: Set<DOMTopLevelEventType> = new Set([
 ]);
 
 if (!enableFormEventDelegation) {
-  capturePhaseEvents.add(TOP_SUBMIT);
-  capturePhaseEvents.add(TOP_RESET);
-}
-
-if (enableCreateEventHandleAPI) {
-  capturePhaseEvents.add(TOP_AFTER_BLUR);
```
gaearon, line 257: Why does it not need special handling anymore?

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -341,20 +336,31 @@ function shouldUpgradeListener(
 
 export function listenToNativeEvent(
   topLevelType: DOMTopLevelEventType,
-  target: EventTarget,
-  listenerMap: ElementListenerMap,
-  eventSystemFlags: EventSystemFlags,
   isCapturePhaseListener: boolean,
+  rootContainerElement: EventTarget,
+  targetElement?: Element,
```
gaearon, line 341: Let's not make many parameters optional? This makes mistakes harder to catch.

If you want to express an nullable value, you can write `targetElement: ?Element,` instead of `targetElement?: Element`. Then it would accept `null` and `undefined`, but not let you skip the arguments completely.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -341,20 +336,31 @@ function shouldUpgradeListener(
...
   // otherwise it won't capture incoming events that are only
   // triggered on the document directly.
   if (topLevelType === TOP_SELECTION_CHANGE) {
-    target = (target: any).ownerDocument || target;
-    listenerMap = getEventListenerMap(target);
+    target = (rootContainerElement: any).ownerDocument;
+  }
+  // If the event can be delegated, we can register it to the root container.
+  // Otherwise, we should register the event to the target element.
+  if (targetElement !== undefined && nonDelegatedEvents.has(topLevelType)) {
```
gaearon, line 355: Seems fragile that `targetElement` accepts either `null` or `undefined`, but we check specifically for `undefined`. Let's strengthen the typing a bit.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -341,20 +336,31 @@ function shouldUpgradeListener(
...
+    eventSystemFlags |= IS_NON_DELEGATED;
+    target = targetElement;
+  }
+  // This is temporary, we will remove this when introducing focusin/focusout
```
gaearon, line 359: Let's stack this PR on https://github.com/facebook/react/pull/19186 and remove special cases.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -387,50 +393,51 @@ export function listenToNativeEvent(
...
-function isCaptureRegistrationName(registrationName: string): boolean {
-  const len = registrationName.length;
-  return registrationName.substr(len - 7) === 'Capture';
-}
-
 export function listenToReactEvent(
-  reactPropEvent: string,
+  reactEvent: string,
   rootContainerElement: Element,
+  targetElement?: Element,
```
gaearon, line 405: Let's make it required? Or take null.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
gaearon: (This not actionable, just note for future readers)

packages/react-dom/src/events/DOMModernPluginEventSystem.js
trueadm: It wasn't ever needed in the capture phase after it was refactored. I must have missed this line.

gaearon:
If you rebase on https://github.com/facebook/react/pull/19395, do tests pass?

trueadm:
@gaearon They do! Woop.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -208,23 +208,18 @@ function extractEvents(
...
 
-// We should not delegate these events to the container, but rather
-// set them on the actual target element itself. This is primarily
-// because these events do not consistently bubble in the DOM.
-export const nonDelegatedEvents: Set<DOMTopLevelEventType> = new Set([
-  TOP_SCROLL,
-  TOP_LOAD,
+// List of events that need to be individually attached to media elements.
+// Note that events in this list will *not* be listened to at the top level
+// unless they're explicitly listed in `ReactBrowserEventEmitter.listenTo`.
```
gaearon, line 213: Nit: there's no `ReactBrowserEventEmitter.listenTo` in the codebase. Outdated?

packages/react-dom/src/client/ReactDOMComponent.js
```diff
@@ -532,29 +549,45 @@ export function setInitialProperties(
...
     case 'img':
     case 'image':
     case 'link':
+      // We listen to these events in case to ensure emulated bubble
+      // listeners still fire for error and load events.
+      listenToNonDelegatedEvent(TOP_ERROR, domElement);
+      listenToNonDelegatedEvent(TOP_LOAD, domElement);
       props = rawProps;
       break;
     case 'form':
```
gaearon, line 582: You can remove form case, it doesn't do anything. Default case is the same.

packages/react-dom/src/events/DOMModernPluginEventSystem.js
```diff
@@ -329,20 +337,27 @@ function shouldUpgradeListener(
 
 export function listenToNativeEvent(
   topLevelType: DOMTopLevelEventType,
-  target: EventTarget,
-  listenerMap: ElementListenerMap,
-  eventSystemFlags: EventSystemFlags,
   isCapturePhaseListener: boolean,
+  rootContainerElement: EventTarget,
+  targetElement: Element | null,
```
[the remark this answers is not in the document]
trueadm: > For a direct subscription, is most of the code here actually necessary? I mean getting/updating the listenerMap, the "upgrading" branch, the nonDelegatedEvents check.

We don't need all the checks, but we still need to handle the listenerMap logic, so we don't add multiple event listeners. I can do this.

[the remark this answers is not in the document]
trueadm: I've made the changes, please let me know if they make sense.

PR #19280, opened 2020-07-08
Updates Electron version for react-devtools

### LINKED ISSUE #19279: Security: 4 Electron (react-devtools dep) security advisories
React version: `16.8.6`

There were 4 security issues filed against `electron`, which `react-devtools` has as a dep. The lowest version that fixes all 4 is `7.2.4` but the version requirement of `electron` for `react-devtools` is `^5.0.0`.

I freely admit that a good solution is to install `react-devtools` as a dev dependency, but for "reasons" that does not work for us. There are likely others out there in similar situations.

* [Issue 1](https://github.com/advisories/GHSA-6vrv-94jv-crrg): low
* [Issue 2](https://github.com/advisories/GHSA-h9jc-284h-533g): high
* [Issue 3](https://github.com/advisories/GHSA-f9mq-jph6-9mhm): moderate
* [Issue 4](https://github.com/advisories/GHSA-m93v-9qjc-3g79): high

These were buried deep in the releases so I am including the links here:

[Electron Changelog from 5 -> 6](https://github.com/electron/electron/releases?after=v7.0.0-nightly.20190730)
[Electron Changelog from 6 -> 7](https://github.com/electron/electron/releases?after=v6.1.1)

Thank you so much for any advice that you may be able to provide. Also thank you for all the work that you do. React, it's community, and it's ecosystem are awesome! 😎

gsimone:
## Summary

Closes #19279 

Updates electron to latest ( 9.1 ) to pull in security fixes, as detailed in #19279 

## Test Plan

Tested the devtools on MacOS 10.14.6, debugging a web app running React 16.2 and an expo react-native app.

CC @bvaughn

### DIFF
--- a/packages/react-devtools/package.json
+++ b/packages/react-devtools/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "cross-spawn": "^5.0.1",
-    "electron": "^5.0.0",
+    "electron": "^9.1.0",
     "ip": "^1.1.4",
     "minimist": "^1.2.3",
     "react-devtools-core": "4.7.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1119,6 +1119,22 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
+"@electron/get@^1.0.1":
+  version "1.12.2"
+  resolved "https://registry.yarnpkg.com/@electron/get/-/get-1.12.2.tgz#6442066afb99be08cefb9a281e4b4692b33764f3"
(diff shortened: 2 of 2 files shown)


rickhanlonii:
Is it surprising that we upgraded 4 major versions with no breaking changes?

gsimone:
It’s probably because the last few majors of electron have been focused on moving away from the remote module, https://www.electronjs.org/docs/breaking-changes

bvaughn:
> Is it surprising that we upgraded 4 major versions with no breaking changes?

I was also surprised, but looking at the breaking changes for [v6](https://github.com/electron/electron/releases/tag/v6.0.0), [v7](https://github.com/electron/electron/releases/tag/v7.0.0), [v8](https://github.com/electron/electron/releases/tag/v8.0.0), and [v9](https://github.com/electron/electron/releases/tag/v9.0.0) it looks like there were few, and none of them are features we rely on.

PR #19281, opened 2020-07-08
feat(devtools): add HOC name in Profiler chart

jyash97:
## Summary

When doing profiling HOC name is not displayed in Chart. We show HOC name in the `Components` tab.

## Test Plan

Ran -> 

From root: `yarn build-devtools` -> For building devtools specific bundles
From `react-devtools-inline`: `yarn start`  -> For starting up app to test with inline package
From `react-devtools-shell`: `yarn start`

Test App -> 

Updated the `react-devtools-inline` app:

Created a HOC component and wrapped `ListItem` twice. Testing video: https://www.loom.com/share/df805218e51d414abf68ea8e3e585cb9

### DIFF
--- a/packages/react-devtools-shared/src/devtools/views/Components/Badge.js
+++ b/packages/react-devtools-shared/src/devtools/views/Components/Badge.js
@@ -15,6 +15,7 @@ import type {ElementType} from 'react-devtools-shared/src/types';
 
 type Props = {|
   className?: string,
+  extraLabelClassName?: string,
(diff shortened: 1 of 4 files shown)


bvaughn:
Thanks a bunch for this PR! 🙇 

Looks like another one was opened around the same time, #19283. It has a visual treatment that I'm slightly more a fan of, and passes CI, so I'm going to move forward with it.

But thank you again.

jyash97:
Thanks for the review. Really Appreciate :)

PR #19282, opened 2020-07-08
Add regression test for #19269

gaearon:
Verified it breaks if #19269 is reverted.


packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
```diff
@@ -2897,6 +2897,89 @@ describe('ReactDOMServerPartialHydration', () => {
...
+    expect(submits).toBe(0);
+
+    // Resolving the promise so that rendering can complete.
+    suspend = false;
+    resolve();
+    await promise;
+
+    Scheduler.unstable_flushAll();
+    jest.runAllTimers();
+    expect(submits).toBe(1);
```
gaearon, line 2978: This is the part that fails because before the fix, it would double-fire.

trueadm:
Nice one :)

PR #19283, opened 2020-07-08
[DevTools] Show HOC names in profiler

### LINKED ISSUE #19259: Suggestion: show HOC names in profiler
(Deleted template as this is a suggestion, not a bug.)

The dev tools helpfully extracts HOC names and shows them in the components tree. [Example](https://react-devtools-tutorial.now.sh/higher-order-components):

image

However, it doesn't give the same treatment to components in the profiler:

image

In large trees, it is very confusing to see two components with the same name, so it would be useful to show the HOC name here as well.

As a workaround for now, users can click through to the "components" tab from the profiler, when a component is selected in the profiler flamegraph, to see this extra information.

bl00mber:
close #19259
related #18802

Screenshot from 2020-07-09

### DIFF
--- a/packages/react-devtools-shared/src/devtools/ProfilerStore.js
+++ b/packages/react-devtools-shared/src/devtools/ProfilerStore.js
@@ -211,6 +211,7 @@ export default class ProfilerStore extends EventEmitter<{|
         id: elementID,
         children: element.children.slice(0),
         displayName: element.displayName,
+        hocDisplayNames: element.hocDisplayNames,
         key: element.key,
         type: element.type,
       };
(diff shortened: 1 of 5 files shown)


jyash97:
Hey @bl00mber 

I have also created the PR for the same. Let me know if I need to close the PR. Here's the link: #19281

bl00mber:
@jyash97 my method of passing the value may be better, however your idea with badges seems interesting. The color of badges on your video is indistinguishable from the blocks on which they are placed: maybe they need to be rendered white. If it will be considered appropriate by react developers I may integrate badges to this PR with your credit.

jyash97:
@bl00mber I agree the way you are passing the values is way better than I did as I wasnt aware of the full codebase so I did what I could figure out. 

We can wait for brian or other react team members for the next steps here. 

Thanks for looking into the PR. 😁

bvaughn:
I will review both in the morning. I believe I have an initial preference for the UI shown on this PR but we'll see when I dig in.

PR #13462, opened 2018-08-22
Revert "Rely on bubbling for submit and reset events (#13358)"

gaearon:
This reverts commit 725e499cfb4da60da46aa5b44c4bad29cad3d08f.

It broke registration on Workplace (thankfully end-to-end tests caught this before the rollout 😛).

I'm not sure why yet. We'll need to dig in deeper. The symptom is that the form that's normally controlled from JavaScript was always being submitted with default action. I guess one way this could happen is the React handler does `e.preventDefault()`, but there's a native DOM handler in the middle that does `e.stopPropagation()`. This would have no effect before this change, but with this change it would actually prevent the React handler from reaching the document. That's my hypothesis anyway. I don't know if this is what happened.

This means we probably can't make more changes to how native/React event mapping works without putting them behind feature flags. There is too much legacy code that can break unintentionally.

PR #19285, opened 2020-07-08
Make form events bubble again

trueadm:
This PR changes form events, specifically `submit` and `reset` so that behave as native bubbling events. Today, with the modern event system, that means making them not use the capture phase.

There are a few issues though. We want to move towards removing the notion of have a special capture phase for certain events. One effort is to make them listen only to their target element, as we do in this PR: https://github.com/facebook/react/pull/19278. That works for all tests, apart from the recent regression test that @gaearon added, where we expect `submit` to be event replayed. We can't replay events that are only attached to their target, as that means we'll double-fire if we try to listen to them on the root too.

Moving `submit` and `reset` to bubble phase sound sensible then right? They natively bubble anyway according to DOM spec, and this includes all major browsers we support. The issue is that we have a Jest test and a reference PR that shows where this went wrong in the past:

https://github.com/facebook/react/pull/13462

Can we maybe go back and fix this problem internally? Then it would unblock this particular set of events and make them behave as they should do. It would also mean that we don't need to treat it as a target only event in https://github.com/facebook/react/pull/19278, and thus the test that @gaearon added should pass.


trueadm:
Alternatively, we could change the Jest test for testing `submit` for event replaying so that it only tests this logic without event replaying. @sebmarkbage would this be a suitable compromise? I'd love your thoughts.

bvaughn:
I don't feel like I have enough context to approve or reject this PR. Seems like the answer hinges on this question:
> Can we maybe go back and fix this problem internally?

So then my question is: Who should make that decision? (That's the thing I really don't have context on- how much effort would it be to fix Workplace.)

trueadm:
@bvaughn 

> So then my question is: Who should make that decision? (That's the thing I really don't have context on- how much effort would it be to fix Workplace.)

I'm looking into this now with @philipp-spiess. If we can't make this change though, the only other thing would be to not do this PR right now, and keep `capture` and `reset` events on the target. That would also mean removing them event replaying, which @sebmarkbage would have to be happy with, until he can implement his ideas around changing replaying to the capture phase.

trueadm:
I've put up an alternative approach: https://github.com/facebook/react/pull/19300

gaearon:
>which @sebmarkbage would have to be happy with

😄

Issue #13567, opened 2018-09-05
Consider stopping auto-adding `px` to number style values (except for a small whitelist)


mgol:
**Do you want to request a *feature* or report a *bug*?**

A removal of a feature, in a sense.

**What is the current behavior?**

React automatically adds the `px` suffix for numerical values passed to the `style` prop. As some CSS properties accept unitless values, React maintains a blacklist of properties that shouldn't get `px` auto-appended.

The problem is that this solution doesn't scale. It requires us to add more & more properties to the list as CSS specs expand and recently the list grows faster; Flexbox & Grid added quite a few of them. What's more confusing, some of those props would work both with & without the `px` suffix and that changes the meaning (`lineHeight` is suffering from that).

Although I'm a React newbie I'm quite familiar with this issue due to being a member of the jQuery Core team. jQuery has the same logic as React here and we keep having to add to the list. We've actually exposed the list at [jQuery.cssNumber](https://api.jquery.com/jQuery.cssNumber/) so that people don't always have to wait for us to add support for a property and do a release.

That's why we decided that in jQuery 4 we'll drop the auto-prefixing blacklist and turn to a whitelist that lists only a few most common properties to which we want to auto-append `px` (mostly because they're extremely common and we don't want to break the world too much); we plan to _not_ expand that list unless we missed something really common. You can see the current plan in my PR: https://github.com/jquery/jquery/pull/4055. In particular, see the proposed whitelist in a (visualized) regexp in:
https://github.com/jquery/jquery/blob/03e9dba3882868e1ee79f1fb0504326da925644f/src/css/isAutoPx.js.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem. Your bug will get fixed much faster if we can run your code and it doesn't have dependencies other than React. Paste the link to your JSFiddle (https://jsfiddle.net/Luktwrdm/) or CodeSandbox (https://codesandbox.io/s/new) example below:**

**What is the expected behavior?**

I propose that React could do the same thing jQuery is planning to and switch the ever-expanding blacklist of CSS props that shouldn't have the `px` suffix applied to a small whitelist that should have the suffix applied.

This topic has been initially described in #13550.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

All browses & OSs. I don't know how old this logic is in React.

philipp-spiess:
This is an awesome proposal. Thanks for writing this up!

[...]

PR #19286, opened 2020-07-08
Add missing unitless CSS properties

kripod:
## Summary

Similarly to [a proposal for Preact](https://github.com/preactjs/preact/issues/2607), I acquired the list of standard and experimental CSS properties which accept unitless values using [MDN data](https://github.com/mdn/data).

## Test Plan

These changes only _expand_ the list of unitless CSS properties, avoiding some issues with unwanted 'px' postfixes.


packages/react-dom/src/shared/CSSProperty.js
```diff
@@ -9,7 +9,9 @@
  * CSS properties which accept numbers but are not in units of "px".
  */
 export const isUnitlessNumber = {
+  animation: true,
```
bvaughn, line 12: This one looks wrong to me. I assume it refers to the "duration" or "delay" parts of the shorthand? Looking at the [`animation-duration` docs](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-duration), it says:
> The time that an animation takes to complete one cycle. This may be specified in either seconds (s) or milliseconds (ms). The value must be positive or zero **and the unit is required**. A value of 0s, which is the default value, indicates that no animation should occur.

The [`animation-delay` docs](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-delay) say the same:
> The time offset, from the moment at which the animation is applied to the element, at which the animation should begin. This may be specified in either seconds (s) or milliseconds (ms). **The unit is required.**

packages/react-dom/src/shared/CSSProperty.js
```diff
@@ -34,11 +37,19 @@ export const isUnitlessNumber = {
...
   gridColumnStart: true,
   fontWeight: true,
+  initialLetter: true,
   lineClamp: true,
   lineHeight: true,
+  maskBorder: true,
+  maskBorderOutset: true,
+  maskBorderSlice: true,
+  maskBorderWidth: true,
+  maxLines: true,
```
bvaughn, line 47: This property isn't standardize and *I think* has been superseded by `line-clamp`? So maybe we shouldn't add it here?

bvaughn:
Since this is a change to a part of ReactDOM that I'm not very familiar with, would you mind doing some testing to ensure the before/after behavior for each of these new properties is *correct*?

packages/react-dom/src/shared/CSSProperty.js
eps1lon: It's a bit hidden under the [grammar for `animation`](https://drafts.csswg.org/css-animations/#animation): `<time> || <easing-function> || <time> || <single-animation-iteration-count> || <single-animation-direction> || <single-animation-fill-mode> || <single-animation-play-state> || [ none | <keyframes-name> ]` where `<single-animation-iteration-count>` is valid syntax which can be a single number.

Reproducible in chrome: 
animation-shorthand-unitless

--https://codesandbox.io/s/animation-unitless-mg06t

bvaughn: I see. I think changing `animation` still seems like a not-great idea since the majority of the numeric units this shorthand property covers do require units.

gaearon:
This adds some bytes to the bundle size but so far (?) no one has asked for these specific properties. Is it worth doing?

If the fix is just to remove the inconsistency, maybe it's not worth doing now. Instead we could flip the list and use a simpler heuristic in the future (https://github.com/facebook/react/issues/13567). Which would be a breaking change, but would do away with any inconsistencies.

trueadm:
I'm not sure there's any real gain from making this change. It adds more size to ReactDOM and there's no real demand for these properties (other than this PR). Furthermore, adding `px` is definitely a historial feature that isn't as relevant these days. Folks should be opting to use relative units rather than static units (like `px`), otherwise they break their app/website to users who use zooming tools or accessibility tooling – invaliding [Web Content Accessibility Guidelines](https://www.w3.org/TR/WCAG21/).

Ideally, we should move React to use a small whitelist like mentioned above, with the ultimate goal of removing this feature entirely in a future version of React.

packages/react-dom/src/shared/CSSProperty.js
sophiebits: Since `animation: 1px` is meaningless/invalid, I don’t think adding `animation` can break anything, at least (except if perhaps someone has a unitless number that gets px and thus ignored today).

bvaughn: Yeah, I'm not familiar with this part of the code, so I don't know what the implications of adding this rule for a shorthand property that can contain multiple numeric values. I agree `animation: 1px` is meaningless though 😆

packages/react-dom/src/shared/CSSProperty.js
kripod: That’s right, I just wanted to add any property with a status of standard or experimental.

kripod:
Thank you, I appreciate all those feedbacks and thoughts mentioned above. I haven’t been aware of alternative approaches before submitting this PR, and for now, I decided to close this PR until a more sustainable solution emerges.

PR #19295, opened 2020-07-09
Fixed test script for DevTools build config

### LINKED ISSUE #19293: DevTools test script runs the wrong tests
```sh
yarn test --build --project devtools
```

This doesn't actually run any DevTools tests. It runs the main React tests.

Console output shows that it's choosing the wrong config:
```sh
$ yarn test --build --project devtools
$ node ./scripts/jest/jest-cli.js --build --project devtools
$ NODE_ENV=development RELEASE_CHANNEL=experimental node ./scripts/jest/jest.js --config ./scripts/jest/config.build.js
```

I wonder if this is also happening for other targets?

bvaughn:
Fixes #19293

### DIFF
--- a/scripts/jest/jest-cli.js
+++ b/scripts/jest/jest-cli.js
@@ -229,12 +229,12 @@ function validateOptions() {
 function getCommandArgs() {
   // Add the correct Jest config.
   const args = ['./scripts/jest/jest.js', '--config'];
-  if (argv.build) {
+  if (argv.project === 'devtools') {
+    args.push(devToolsConfig);
+  } else if (argv.build) {
     args.push(buildConfig);
   } else if (argv.persistent) {
     args.push(persistentConfig);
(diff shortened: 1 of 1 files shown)


scripts/jest/jest-cli.js
```diff
@@ -229,12 +229,12 @@ function validateOptions() {
 function getCommandArgs() {
   // Add the correct Jest config.
   const args = ['./scripts/jest/jest.js', '--config'];
-  if (argv.build) {
+  if (argv.project === 'devtools') {
```
bvaughn, line 232: DevTools *always* uses the `build` config, but not vice versa, so it needs to come first.

Having one if/else condition flow for two flags seems a bit fragile, but I didn't want to make a larger change here.

rickhanlonii: Yeah that makes sense.

PR #19307, opened 2020-07-10
Use Lanes to check if a render is a Suspense retry

acdlite:
Now that Suspense retries have their own dedicated set of lanes (#19287), we can determine if a render includes only retries by checking if its lanes are a subset of the retry lanes.

Previously we inferred this by checking `workInProgressRootLatestProcessedEventTime`. If it's not set, that implies that no updates were processed in the current render, which implies it must be a Suspense retry. The eventual plan is to get rid of `workInProgressRootLatestProcessedEventTime` and instead track event times on the root; this change is one the steps toward that goal.

The relevant tests were originally added in #15769.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberLane.js
+++ b/packages/react-reconciler/src/ReactFiberLane.js
@@ -451,8 +451,11 @@ export function getLanesToRetrySynchronouslyOnError(root: FiberRoot): Lanes {
 export function returnNextLanesPriority() {
   return return_highestLanePriority;
 }
-export function hasUpdatePriority(lanes: Lanes) {
+export function includesNonIdleWork(lanes: Lanes) {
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -451,9 +451,12 @@ export function getLanesToRetrySynchronouslyOnError(root: FiberRoot): Lanes {
 export function returnNextLanesPriority() {
   return return_highestLanePriority;
 }
-export function hasUpdatePriority(lanes: Lanes) {
+export function includesNonIdleWork(lanes: Lanes) {
```
acdlite, line 454: This rename is unrelated. Just felt slightly better than the old one.

Issue #19308, opened 2020-07-10
Bug: Unexpected debugger statement in DevTools (solved)


eiamei:
Looks like you forgot about the debugger inside the react_devtools_backend.js. 

React dev tools version: 4.8.0

## Steps To Reproduce

1. Update (or install) the latest version of devtools
2. Open devtools
3. The code will be stopped on the debugger inside react_devtools_backend.js

![IMAGE 2020-07-10 13:43:52](https://user-images.githubusercontent.com/20472264/87146286-6523c900-c2b3-11ea-892f-92c01d0a2412.jpg)

## The current behavior
If I open devtools with react dev tools extension it will stop every time on debugger inside react_devtools_backend.js

## The expected behavior
It doesn't seem like it should be happening 😃

JakubRimal:
Same here, I've had to disable the extension because of this... :smile:

martinhorsky:
Same for me.

gaearon:
Looks bad.

antislash21:
Same here

Lukortech:
https://github.com/facebook/react/blob/14084be286d09df49d5410c751bc28f0180a54b2/packages/react-devtools-shared/src/backend/console.js#L173

That's the guy you're looking for :D

gaearon:
Yes, but see the comment above it. :-) It is intentional, but this option should have been off by default.

przemuh:
As a temporary solution you can "blackbox" script in Chrome Dev Tools:

![image](https://user-images.githubusercontent.com/3865478/87158097-7e327700-c2bf-11ea-86ea-68013efcbd5c.png)

gasparsigma:
Temporary workaround if you don't want to disable the extension is to directly edit the src of the extension:
C:\Users\[Your User]\AppData\Local\Google\Chrome\User Data\Default\Extensions\fmkadmapgofadopljbjfkapdkoienihi\4.8.0_0\build\react_devtools_backend.js
You can just remove the debugger for that line

gaearon:
You should also be able to go to Components tab, press Settings gear, add checkbox to Break on Warning, then do this again, and remove the checkbox as a temporary workaround.

gaearon:
I'll keep it open until we cut a release

eiamei:
Thanks 🙌

gaearon:
[...]

PR #19309, opened 2020-07-10
DevTools: Make break-on-warn off by default

gaearon:
Fixes https://github.com/facebook/react/issues/19308.

Changelog said https://github.com/facebook/react/pull/19048 should be off by default but it's on by default.


packages/react-devtools-shared/src/utils.js
```diff
@@ -258,7 +258,7 @@ export function getBreakOnConsoleErrors(): boolean {
       return JSON.parse(raw);
     }
   } catch (error) {}
-  return true;
+  return false;
```
bvaughn, line 261: Oh, damn. My bad.

PR #19314, opened 2020-07-10
Removed debugger

prayas26:
Solves the issue that forces a debugger when installed extension in chrome

## Summary
Solves the issue that forces a debugger on extension


gaearon:
This is not right the fix. If you read right above the `debugger`, it explains why this line exists.

The correct fix has already been merged, but we're waiting for Chrome to publish the update.

In the meantime, you can find a workaround here: https://github.com/facebook/react/issues/19308#issuecomment-656682924.

ycjcl868:
+1

image

PR #19315, opened 2020-07-10
Extract some of the tidy up changes from 19278

trueadm:
In order to make https://github.com/facebook/react/pull/19278 a bit smaller in terms of changes, this extracts some of the tidy up changes from that PR. https://github.com/facebook/react/pull/19278 should now only refer to the changes reflecting to moving away from capture phase events.

### DIFF
--- a/packages/react-dom/src/client/ReactDOMEventHandle.js
+++ b/packages/react-dom/src/client/ReactDOMEventHandle.js
@@ -24,14 +24,14 @@ import {
 } from './ReactDOMComponentTree';
 import {ELEMENT_NODE} from '../shared/HTMLNodeType';
 import {
-  listenToTopLevelEvent,
+  listenToNativeEvent,
   addEventTypeToDispatchConfig,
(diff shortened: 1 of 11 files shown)


packages/react-dom/src/client/ReactDOMEventHandle.js
```diff
@@ -146,21 +146,21 @@ function registerReactDOMEvent(
...
+      isPassiveListener,
+      listenerPriority,
+      isCapturePhaseListener,
     );
   } else if (isValidEventTarget(target)) {
     const eventTarget = ((target: any): EventTarget);
     const listenerMap = getEventListenerMap(eventTarget);
-    listenToTopLevelEvent(
+    listenToNativeEvent(
       topLevelType,
```
rickhanlonii, line 157: Should this change to nativeEventType then?

trueadm: Dan's further change to remove the concept `topLevelType` will address this.

PR #19316, opened 2020-07-10
reference not add to dependencies if only its type is used as typeof arg

### LINKED ISSUE #18828: Bug [ESLint Hooks Plugin]: When using a `typeof` type guard it requires the value as a dependency
**Version:** eslint-plugin-react-hooks@1.7.0

## Steps To Reproduce

1. Create a component with local state
2. Create a useEffect with a variable inside, that points to the `typeof` of the state variable

Link to code example:

https://codesandbox.io/s/nervous-williamson-2i99l?file=/src/App.tsx:210-261

## The current behavior
`react-exhaustive-deps` complains that the state variable should be included in the dependency list

## The expected behavior
`react-exhaustive-deps` shouldn't complain that the state variable should be included in the dependency list because it's value is never used.

delca85:
## Summary

Fixes #18828 .

`react-exhaustive-deps` does not complain anymore about a missing dependency if its value is never used but its type is used as `typeof` argument.

Checking if a reference parent type has type `TSTypeQuery`, a reference is not added to dependencies.

## Test Plan

Added test to check the expected behavior and even that the dependency is required by `react-exhaustive-deps` if its type is used and its value too.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -6983,6 +6983,18 @@ const testsTypescript = {
         }
       `,
     },
+    {
+      code: normalizeIndent`
+        function MyComponent() {
+          const [state, setState] = React.useState<number>(0);
+
+          useEffect(() => {
+            const someNumber: typeof state = 2;
+            setState(prevState => prevState + someNumber);
+          }, [])
+        }
+      `,
+    },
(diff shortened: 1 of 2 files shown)


rickhanlonii:
@delca85 thanks for this!

Will this also fix https://github.com/facebook/react/issues/19327? Could we add a test for that as well?

delca85:
Hey @rickhanlonii, I am going to check it as soon as possible!

gaearon:
I'll rebase and get this in. Confirmed it fixes https://github.com/facebook/react/issues/19327.

PR #19317, opened 2020-07-11
Fixed some minor Typos

Ashikpaul:
## Summary

Found some typos in the error codes

## Test Plan

This PR doesn't require a test plan

Thank you


Ashikpaul:
@facebook-github-bot I have signed it

scripts/error-codes/codes.json
```diff
@@ -149,7 +149,7 @@
   "147": "Missing owner for string ref %s. This error is likely caused by a bug in React. Please file an issue.",
   "148": "Expected ref to be a function or a string.",
   "149": "Element ref was specified as a string (%s) but no owner was set. You may have multiple copies of React loaded. (details: https://fb.me/react-refs-must-have-owner).",
-  "150": "An object is not an iterable. This error is likely caused by a bug in React. Please file an issue.",
+  "150": "An object is not iterable. This error is likely caused by a bug in React. Please file an issue.",
```
rickhanlonii, line 152: This one is correct, "iterable" here is a noun not a verb (see the Interators section [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators)).

Ashikpaul: Thank you @rickhanlonii 😍.

I am not sure if I am doing it right. I checked ```accumulate.js``` , ```accumulateInto.js``` and some files in that path and didn't find anything that requires an update. Maybe I am not looking at the right file. Could you please check.

rickhanlonii: The update is for the other error. That error is used [here](https://github.com/facebook/react/blob/e3f4eb7272d4ca0ee49f27577156b57eeb07cf73/packages/react-native-renderer/src/legacy-events/accumulate.js#L25) and [here](https://github.com/facebook/react/blob/e3f4eb7272d4ca0ee49f27577156b57eeb07cf73/packages/react-native-renderer/src/legacy-events/accumulateInto.js#L31) so the error message string needs to match the text in this file. 

When you run `yarn extract-errors` there should be no changes.

bvaughn:
Looks like this one is stale and conflicted. Closing.

PR #19319, opened 2020-07-11
fix(act): Don't warn if an effect was not queued

### LINKED ISSUE #19318: Bug: act warning misleading if an effect wasn't run
React version: 16.13.1 and `next`

## Steps To Reproduce

1. Re-render a component where an effect wasn't run because dependencies didn't change
2. Receive missing act warning

Link to code example: https://codesandbox.io/s/missing-act-on-every-effect-ldy3b?file=/src/index.test.js

## The current behavior

React creates warnings for each call of React.useEffect regardless of whether an effect was actually scheduled i.e. the effect was run.

## The expected behavior

If it is impossible for an effect to be run I shouldn't have to wrap it in act. Changing the behavior would help identifying what effects were unexpectedly scheduled. Right now we receive warnings for effects that are unrelated to a specific update.

I consider the current warning misleading and unnecessary because the [documentation](https://reactjs.org/docs/testing-recipes.html#act) talks about "all updates related to these “units”" which is not the case here in my opinion. The act warning says "When testing, code that causes React state updates should be wrapped into act(...):" which also doesn't apply here: The code in question could not cause a state update from an effect.

eps1lon:
## Summary

Closes #19318 

## Test Plan

- [x] build green
- [x] codesandbox in #19318 passes: https://codesandbox.io/s/missing-act-on-every-effect-forked-1iwp2?file=/src/index.test.js
- [x] use codesandbox build for an existing codebase where act warnings were enabled by setting `globalThis.jest = null`

### DIFF
--- a/packages/react-reconciler/src/ReactFiberHooks.new.js
+++ b/packages/react-reconciler/src/ReactFiberHooks.new.js
@@ -1432,12 +1432,11 @@ function updateEffectImpl(fiberFlags, hookFlags, create, deps): void {
   }
 
   if (__DEV__) {
-    if (
-      (hookFlags & HookPassive) !== HookNoFlags &&
-      // $FlowExpectedError - jest isn't a global, and isn't recognized outside of tests
-      'undefined' !== typeof jest
-    ) {
-      warnIfNotCurrentlyActingEffectsInDEV(currentlyRenderingFiber);
+    // $FlowExpectedError - jest isn't a global, and isn't recognized outside of tests
+    if ('undefined' !== typeof jest) {
+      if ((hookFlags & HookPassive) !== HookNoFlags) {
+        warnIfNotCurrentlyActingEffectsInDEV(currentlyRenderingFiber);
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1275,12 +1282,6 @@ function updateEffect(
   create: () => (() => void) | void,
   deps: Array<mixed> | void | null,
 ): void {
-  if (__DEV__) {
```
eps1lon, line 1278: Should we move this in `mountEffect` as well for symmetry?

eps1lon:
Bump

53jk1 (approved):

eps1lon:
Rebased and fixed some misconceptions about tags vs flags. I treated the hook flags as a single number when they're actually a bitset.

eps1lon:
Will no longer be relevant for React 18 where even a render needs to be wrapped in `act`. See https://github.com/reactwg/react-18/discussions/23#discussioncomment-1088385

PR #19322, opened 2020-07-11
Effects list refactor continued: did-bailout flag

bvaughn:
Builds on top of #19261

* Adds an explicit "did bailout" flag rather than checking whether or not children were cloned.
* Updates `subtreeTag` to use its own meta values rather than copying `effectTag` values.

[**View only the delta.**](https://github.com/bvaughn/react/compare/effects_list_refactor...effects_list_refactor_bailout_flag)

### DIFF
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -293,6 +293,8 @@ function ChildReconciler(shouldTrackSideEffects) {
     } else {
       returnFiber.deletions.push(childToDelete);
     }
+    // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
+    returnFiber.effectTag |= Deletion;
     childToDelete.nextEffect = null;
     childToDelete.effectTag = Deletion;
   }
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -297,7 +297,6 @@ function ChildReconciler(shouldTrackSideEffects) {
     // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
     returnFiber.effectTag |= Deletion;
     childToDelete.nextEffect = null;
-    childToDelete.effectTag = Deletion;
   }
 
   function deleteRemainingChildren(
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -288,7 +288,11 @@ function ChildReconciler(shouldTrackSideEffects) {
     } else {
       returnFiber.firstEffect = returnFiber.lastEffect = childToDelete;
     }
-    returnFiber.deletions.push(childToDelete);
+    if (returnFiber.deletions === null) {
+      returnFiber.deletions = [childToDelete];
+    } else {
+      returnFiber.deletions.push(childToDelete);
+    }
     childToDelete.nextEffect = null;
     childToDelete.effectTag = Deletion;
   }
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -280,6 +280,7 @@ function ChildReconciler(shouldTrackSideEffects) {
     // deletions, so we can just append the deletion to the list. The remaining
     // effects aren't added until the complete phase. Once we implement
     // resuming, this may not be true.
+    // TODO (effects) Get rid of effects list update here.
     const last = returnFiber.lastEffect;
     if (last !== null) {
       last.nextEffect = childToDelete;
@@ -287,6 +288,7 @@ function ChildReconciler(shouldTrackSideEffects) {
     } else {
       returnFiber.firstEffect = returnFiber.lastEffect = childToDelete;
     }
+    returnFiber.deletions.push(childToDelete);
     childToDelete.nextEffect = null;
     childToDelete.effectTag = Deletion;
   }
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -291,11 +291,11 @@ function ChildReconciler(shouldTrackSideEffects) {
     const deletions = returnFiber.deletions;
     if (deletions === null) {
       returnFiber.deletions = [childToDelete];
+      // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
+      returnFiber.effectTag |= Deletion;
     } else {
       deletions.push(childToDelete);
     }
-    // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
-    returnFiber.effectTag |= Deletion;
     childToDelete.nextEffect = null;
   }
 
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -288,10 +288,11 @@ function ChildReconciler(shouldTrackSideEffects) {
     } else {
       returnFiber.firstEffect = returnFiber.lastEffect = childToDelete;
     }
-    if (returnFiber.deletions === null) {
+    const deletions = returnFiber.deletions;
+    if (deletions === null) {
       returnFiber.deletions = [childToDelete];
     } else {
-      returnFiber.deletions.push(childToDelete);
+      deletions.push(childToDelete);
     }
     // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
     returnFiber.effectTag |= Deletion;
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -2066,6 +2066,7 @@ function updateSuspensePrimaryChildren(
     currentFallbackChildFragment.nextEffect = null;
     currentFallbackChildFragment.effectTag = Deletion;
     workInProgress.firstEffect = workInProgress.lastEffect = currentFallbackChildFragment;
+    workInProgress.deletions.push(currentFallbackChildFragment);
   }
 
   workInProgress.child = primaryChildFragment;
@@ -2131,9 +2132,11 @@ function updateSuspenseFallbackChildren(
       workInProgress.firstEffect = primaryChildFragment.firstEffect;
       workInProgress.lastEffect = progressedLastEffect;
       progressedLastEffect.nextEffect = null;
+      workInProgress.deletions = [];
     } else {
       // TODO: Reset this somewhere else? Lol legacy mode is so weird.
       workInProgress.firstEffect = workInProgress.lastEffect = null;
+      workInProgress.deletions = [];
     }
   } else {
     primaryChildFragment = createWorkInProgressOffscreenFiber(
@@ -3040,6 +3043,7 @@ function remountFiber(
     } else {
       returnFiber.firstEffect = returnFiber.lastEffect = current;
     }
+    returnFiber.deletions.push(current);
     current.nextEffect = null;
     current.effectTag = Deletion;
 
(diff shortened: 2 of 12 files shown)


packages/react-reconciler/src/ReactChildFiber.new.js
```diff
@@ -280,13 +280,21 @@ function ChildReconciler(shouldTrackSideEffects) {
...
     // resuming, this may not be true.
+    // TODO (effects) Get rid of effects list update here.
     const last = returnFiber.lastEffect;
     if (last !== null) {
       last.nextEffect = childToDelete;
       returnFiber.lastEffect = childToDelete;
     } else {
       returnFiber.firstEffect = returnFiber.lastEffect = childToDelete;
     }
+    if (returnFiber.deletions === null) {
```
trueadm, line 291: Maybe hoist `returnFiber.deletions` out so it can be re-used below?

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2066,6 +2067,13 @@ function updateSuspensePrimaryChildren(
     currentFallbackChildFragment.nextEffect = null;
     currentFallbackChildFragment.effectTag = Deletion;
     workInProgress.firstEffect = workInProgress.lastEffect = currentFallbackChildFragment;
+    if (workInProgress.deletions === null) {
```
trueadm, line 2070: Same here.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -3040,6 +3056,13 @@ function remountFiber(
     } else {
       returnFiber.firstEffect = returnFiber.lastEffect = current;
     }
+    if (returnFiber.deletions === null) {
```
trueadm, line 3059: Same here.

packages/react-reconciler/src/ReactFiberHydrationContext.new.js
```diff
@@ -125,6 +125,13 @@ function deleteHydratableInstance(
   childToDelete.stateNode = instance;
   childToDelete.return = returnFiber;
   childToDelete.effectTag = Deletion;
+  if (returnFiber.deletions === null) {
```
trueadm, line 128: Same here.

acdlite:
Overall the implementation looks right, exactly what I had expected. 👍 

There are a few other bailout cases in begin phase where we need to set the `DidBailout` flag. Like:

https://github.com/facebook/react/blob/9ea0f6752da28a91bdb56382367e2a07390cb733/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L1842-L1848

Basically anywhere there's a `return null` inside `ReactFiberBeginWork`. Maybe we should replace all of them with a function call, with a name like `reuseCurrentChild`.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1789,24 +1803,40 @@ function resetChildLanes(completedWork: Fiber) {
...
     let child = completedWork.child;
     while (child !== null) {
       newChildLanes = mergeLanes(
         newChildLanes,
         mergeLanes(child.lanes, child.childLanes),
       );
-      if (shouldBubbleActualDurations) {
+
+      if (!didBailout) {
+        subtreeTag |= child.subtreeTag & ~DidBailout;
```
acdlite, line 1814: Don't need to remove the `DidBailout` bit since `!didBailout` implies that it's already 0.

acdlite: Oh never mind, different fiber

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1833,11 +1863,35 @@ function resetChildLanes(completedWork: Fiber) {
...
+        childrenDidNotComplete = true;
+      }
+
       child = child.sibling;
     }
   }
 
   completedWork.childLanes = newChildLanes;
+
+  if (!childrenDidNotComplete) {
```
acdlite, line 1892: I wonder why this check is necessary. `resetChildLanes` shouldn't be called if something in the subtree didn't complete. So I wouldn't expect `childrenDidNotComplete` to  ever be true.

Do you remember if any tests failed without this guard?

packages/react-reconciler/src/ReactSubtreeTags.js
```diff
@@ -0,0 +1,16 @@
...
+ * @flow
+ */
+
+export type SubtreeTag = number;
+
+export const NoEffect = /*        */ 0b0000;
+export const DidBailout = /*      */ 0b0001;
+export const BeforeMutation = /*  */ 0b0010;
+export const Mutation = /*        */ 0b0100;
+export const Layout = /*          */ 0b1000;
```
acdlite, line 16: As a follow up, we should move the `Incomplete`, `ShouldCapture`, and `DidCapture` flags to this new field.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
lunaruan: I don't remember exactly why this was here (I think left over from a previous implementation when tests were failing to match with what the effect list was doing [here](https://github.com/facebook/react/blob/7171494026b256a5e0778414e3d0bf1d4c0e9dee/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L1681). All the tests pass without this though so I'll remove.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2375,7 +2386,7 @@ function updateDehydratedSuspenseComponent(
       retry = Schedule_tracing_wrap(retry);
     }
     registerSuspenseInstanceRetry(suspenseInstance, retry);
-    return null;
+    return reuseCurrentChild(workInProgress);
```
bvaughn, line 2389: nit: Seems odd to call this and have it return null for us. Let's just call this and return `null` on the next line?

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2980,10 +2996,15 @@ function bailoutOnAlreadyFinishedWork(
 
   // Check if the children have any pending work.
   if (!includesSomeLane(renderLanes, workInProgress.childLanes)) {
+    // TODO (effects)  The case we are trying to handle is an error thrown during commit,
+    // and we replay the work, even though we skip render phase (by bailing out)
+    // we don't want to bailout on commit effects too.
+    // However, this check seems overly aggressive in some cases.
```
bvaughn, line 3002: Is this comment still relevant, @lunaruan?

lunaruan: Oops nope.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2141,35 +2141,40 @@ function commitRootImpl(root, renderPriorityLevel) {
 }
 
 function commitBeforeMutationEffects(fiber: Fiber) {
```
sebmarkbage, line 2143: As a convention, I've used `firstChild` as the name of the parameter when it's conceptually passed a linked list of children rather than a specific fiber.

https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberBeginWork.new.js#L2453

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1789,24 +1803,36 @@ function resetChildLanes(completedWork: Fiber) {
...
-
     let child = completedWork.child;
     while (child !== null) {
       newChildLanes = mergeLanes(
         newChildLanes,
         mergeLanes(child.lanes, child.childLanes),
       );
-      if (shouldBubbleActualDurations) {
+
+      if (!didBailout) {
```
sebmarkbage, line 1813: Nit: It might be better to fork this loop into two and check the boolean before entering each loop instead of checking it for each iteration.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1833,11 +1859,29 @@ function resetChildLanes(completedWork: Fiber) {
         newChildLanes,
         mergeLanes(child.lanes, child.childLanes),
       );
+
+      if (!didBailout) {
```
sebmarkbage, line 1863: Nit: It might be better to fork this loop into two and check the boolean before entering each loop instead of checking it for each iteration.

In fact, we probably could just copy the childLanes field from "current" since it must be the same then.

lunaruan: It seems like just copying the childLanes doesn't seem to work. Just doing the former for now. @bvaughn can maybe look into it more tomorrow?

acdlite: @lunaruan You can only copy from `childLanes` when `didBailout` is true. Is that what you tried? If that fails some tests I'd be curious to know which ones.

acdlite: In fact, you shouldn't even have to copy it in the `didBailout` path because it should already equal the one on `current`. Might be some edge case where this doesn't hold true, but I can't think of one and if any exists it's probably wrong.

packages/react-reconciler/src/ReactChildFiber.new.js
```diff
@@ -280,13 +280,22 @@ function ChildReconciler(shouldTrackSideEffects) {
...
       returnFiber.firstEffect = returnFiber.lastEffect = childToDelete;
     }
+    const deletions = returnFiber.deletions;
+    if (deletions === null) {
+      returnFiber.deletions = [childToDelete];
+    } else {
+      deletions.push(childToDelete);
+    }
+    // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
+    returnFiber.effectTag |= Deletion;
```
acdlite, line 298: You shouldn't need to set this on `childToDelete` anymore, right?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
lunaruan: Yeah when `didBailout` is `true` I tried setting `completedWork.childLanes` to `completedWork.alternate.childLanes`. I didn't investigate much yet but a ton of tests failed.

packages/react-reconciler/src/ReactChildFiber.new.js
acdlite: Which is nice, since `childToDelete` is a current fiber, so that wouldn't work in a world where the commit phase is concurrent with subsequent renders

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -3040,6 +3050,14 @@ function remountFiber(
...
       returnFiber.firstEffect = returnFiber.lastEffect = current;
     }
+    const deletions = returnFiber.deletions;
+    if (deletions === null) {
+      returnFiber.deletions = [current];
+    } else {
+      deletions.push(current);
+    }
+    // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
+    returnFiber.effectTag |= Deletion;
```
acdlite, line 3060: Same here. Can remove the flag from the child.

packages/react-reconciler/src/ReactChildFiber.new.js
acdlite: Oh, I guess you might still need it for passive effects phase, which still uses the effect list. Can remove it after that, though.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
bvaughn: Whoops, yup.

packages/react-reconciler/src/ReactChildFiber.new.js
bvaughn: Oops yeah.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
bvaughn: I find these multi-line forks get a little hard to trace through, but you're right we avoid re-checking `didBailout` a bunch this way so... :smile: fair enough!

bvaughn:
Thanks for the reviews folks

PR #19336, opened 2020-07-13
fix mistake in React devtools

b-ponomarenko:
## Summary

Fixed problem when someone is trying to connect to React devtools via SSL


gaearon:
Can you expand on what this fix is doing?

Issue #19338, opened 2020-07-13
Bug: react-devtools error when KEY and CERT are provided, SSL not usable


digicide:
SSL is not usable because key and cert options are added to an uninitialized options object.

Line 149 of react-devtools/app.html reads 'let options;'
Line 154 and 155 begin with 'options.cert = …'

In every case, this throws an error and enters the catch block.

React version: 16.5.2
react-devtools version: 4.8.1

## Steps To Reproduce

1. on command line, execute
KEY=react-devtools-ssl-key.pem CERT=react-devtools-ssl-cert.pem ./node_modules/react-devtools/bin.js

Regardless of whether the key and cert are valid, the error "Cannot set property 'key' of undefined" is thrown.

digicide:
I was able to fix this by changing line 149 in my local copy to:
let options = {};

bvaughn:
cc @ittaibaratz Interested in picking this one up?

OR @digicide, want to submit a fix? :)

ittaibaratz:
@digicide's fix sounds good to me - I'll submit a PR shortly.

bvaughn:
Looks like a PR already got submitted #19336

ittaibaratz:
> cc @ittaibaratz Interested in picking this one up?
> 
> OR @digicide, want to submit a fix? :)

Actually, there's already a PR that fixes a slightly different way - https://github.com/facebook/react/pull/19336/files

bvaughn:
Sorry for the inconvenience here. Will probably publish a release in the morning.

bvaughn:
Version 4.8.2 has just been published to NPM with a fix for this issue.

PR #19339, opened 2020-07-13
Fix devtools uninitialized options for SSL

digicide:
## Summary

Fix bug described in #19338.


bvaughn:
Thanks for the fix! But we just merged another one, #19336

PR #19342, opened 2020-07-14
Set current update lane priority for user blocking events

rickhanlonii:
## Summary

We missed these callsites in https://github.com/facebook/react/pull/19121 and one is the most common mismatch right now.

### DIFF
--- a/packages/react-dom/index.experimental.js
+++ b/packages/react-dom/index.experimental.js
@@ -25,7 +25,6 @@ export {
   unstable_scheduleHydration,
   // DO NOT USE: Temporarily exposing these to migrate off of Scheduler.runWithPriority.
   unstable_runWithPriority,
-  unstable_IdleLanePriority,
   // Disabled behind disableUnstableRenderSubtreeIntoContainer
   unstable_renderSubtreeIntoContainer,
   // Disabled behind disableUnstableCreatePortal
--- a/packages/react-dom/index.experimental.js
+++ b/packages/react-dom/index.experimental.js
@@ -23,8 +23,9 @@ export {
   createBlockingRoot as unstable_createBlockingRoot,
   unstable_flushControlled,
   unstable_scheduleHydration,
-  // DO NOT USE: Temporarily exposing this to migrate off of Scheduler.runWithPriority.
+  // DO NOT USE: Temporarily exposing these to migrate off of Scheduler.runWithPriority.
   unstable_runWithPriority,
+  unstable_IdleLanePriority,
   // Disabled behind disableUnstableRenderSubtreeIntoContainer
   unstable_renderSubtreeIntoContainer,
   // Disabled behind disableUnstableCreatePortal
--- a/packages/react-dom/index.experimental.js
+++ b/packages/react-dom/index.experimental.js
@@ -23,7 +23,7 @@ export {
   createBlockingRoot as unstable_createBlockingRoot,
   unstable_flushControlled,
   unstable_scheduleHydration,
-  // DO NOT USE: Temporarily exposing these to migrate off of Scheduler.runWithPriority.
+  // DO NOT USE: Temporarily exposing this to migrate off of Scheduler.runWithPriority.
   unstable_runWithPriority,
   // Disabled behind disableUnstableRenderSubtreeIntoContainer
   unstable_renderSubtreeIntoContainer,
--- a/packages/react-dom/src/events/DeprecatedDOMEventResponderSystem.js
+++ b/packages/react-dom/src/events/DeprecatedDOMEventResponderSystem.js
@@ -56,6 +56,8 @@ import {
 // Intentionally not named imports because Rollup would use dynamic dispatch for
 // CommonJS interop named imports.
 import * as Scheduler from 'scheduler';
+import {unstable_runWithPriority} from '../client/ReactDOM';
+
 const {
   unstable_UserBlockingPriority: UserBlockingPriority,
   unstable_runWithPriority: runWithPriority,
@@ -101,9 +103,12 @@ const eventResponderContext: ReactDOMResponderContext = {
         break;
       }
       case UserBlockingEvent: {
-        runWithPriority(UserBlockingPriority, () =>
-          executeUserEventHandler(eventListener, eventValue),
-        );
+        // TODO: Double wrapping is necessary while we decouple Scheduler priority.
+        unstable_runWithPriority(UserBlockingPriority, () => {
+          runWithPriority(UserBlockingPriority, () =>
+            executeUserEventHandler(eventListener, eventValue),
+          );
+        });
         break;
       }
       case ContinuousEvent: {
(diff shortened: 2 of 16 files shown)


acdlite:
The feature flag should be static since it only exists to prevent it from leaking into open source. Can be static `true` in FB dev builds since we use the warning filter to gate.

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -148,16 +149,19 @@ function dispatchUserBlockingUpdate(
...
-    dispatchEvent.bind(
-      null,
-      topLevelType,
-      eventSystemFlags,
-      container,
-      nativeEvent,
-    ),
-  );
+  // TODO: Double wrapping is necessary while we decouple Scheduler priority.
+  unstable_runWithPriority(UserBlockingPriority, () => {
```
acdlite, line 153: Since this is part of the renderer, we should be able to mutate the value directly without a wrapper function.

Not gonna block the PR over this, but does need to be fixed before we can land the experiment..

acdlite: Wait, also, isn't UserBlockingPriority the wrong type? Should be a lane priority.

rickhanlonii: > isn't UserBlockingPriority the wrong type? Should be a lane priority.

It must have been buried, but I mentioned this in the comment [here](https://github.com/facebook/react/pull/19121#discussion_r449238705) - ReactDOM.runWithPriority accepts a `ReactPriorityLevel` instead of a `LanePriority` so that we don't have to leak the lane implementation while we do the decoupling. Internally in React, everything operates on the Lane priority.

Were you expecting that we would export all of the LanePriorities, and the `schedulerPriorityToLanePriority` conversion function from ReactFiberReconciler so we can pass LanePriorities to ReactDOM.runWithPriority?

> Since this is part of the renderer, we should be able to mutate the value directly without a wrapper function.

Do you mean that we wouldn't need the try/catch? That would require that we export the current update lane getter and setter from ReactFiberReconciler as well as all the LanePriorities like mentioned above. Right now we only export `runWithPriority (ReactPriorityLevel)` and `getCurrentUpdatePriority (ReactPriorityLevel)`.

acdlite: I think it's fine to leak the lane priorities to the renderer because it's part of the same build artifact. It doesn't cross any package boundary.

Leaking to via an API `ReactDOM.runWithPriority` is different because that's called by external code.

> ReactDOM.runWithPriority accepts a ReactPriorityLevel instead of a LanePriority so that we don't have to leak the lane implementation while we do the decoupling. Inside the reconciler, everything operates on the Lane priority.

I missed that comment originally but I don't buy that rationale. `ReactDOM.runWithPriority` is an unstable API that only exists internally if/when we need it to temporarily migrate external callers. We're going to delete it ASAP, so "leaking" update lane priorities is not a huge issue. But more importantly, the caller in this PR isn't external, so that's not relevant in this case.

rickhanlonii:
@acdlite I replaced runWithPriority with direct updates, and updated the exposed methods to use LanePriority. I can follow up with the feature flags in the next diff so this is easier to review if you want to re-review this now.

rickhanlonii:
The next branch is here: https://github.com/rickhanlonii/react/compare/rh-add-rwp-to-events...rickhanlonii:rh-ff-update-lane?expand=1

packages/react-dom/src/events/ReactDOMEventListener.js
```diff
@@ -148,6 +153,9 @@ function dispatchUserBlockingUpdate(
   container,
   nativeEvent,
 ) {
+  // TODO: Double wrapping is necessary while we decouple Scheduler priority.
+  const previousPriority = getCurrentUpdateLanePriority();
+  setCurrentUpdateLanePriority(InputContinuousLanePriority);
```
acdlite, line 158: Needs a try-catch, since runWithPriority could throw

packages/react-dom/src/events/DeprecatedDOMEventResponderSystem.js
```diff
@@ -101,9 +108,12 @@ const eventResponderContext: ReactDOMResponderContext = {
         break;
       }
       case UserBlockingEvent: {
+        const previousPriority = getCurrentUpdateLanePriority();
+        setCurrentUpdateLanePriority(InputContinuousLanePriority);
         runWithPriority(UserBlockingPriority, () =>
           executeUserEventHandler(eventListener, eventValue),
         );
+        setCurrentUpdateLanePriority(previousPriority);
```
acdlite, line 116: Needs a try-catch

packages/react-dom/index.experimental.js
```diff
@@ -23,8 +23,9 @@ export {
   createBlockingRoot as unstable_createBlockingRoot,
   unstable_flushControlled,
   unstable_scheduleHydration,
-  // DO NOT USE: Temporarily exposing this to migrate off of Scheduler.runWithPriority.
+  // DO NOT USE: Temporarily exposing these to migrate off of Scheduler.runWithPriority.
   unstable_runWithPriority,
+  unstable_IdleLanePriority,
```
acdlite, line 28: Just hardcode the number 2 instead of importing.

packages/react-noop-renderer/src/ReactNoop.js
```diff
@@ -47,9 +47,9 @@ export const {
   act,
   dumpTree,
   getRoot,
-  // TODO: Remove this once callers migrate to alternatives.
-  // This should only be used by React internals.
+  // TODO: Remove these after callers migrate to alternatives.
   unstable_runWithPriority,
+  unstable_InputContinuousLanePriority,
```
acdlite, line 52: Same comment as before. These are only used internally. Hardcode the number wherever it’s used, with a comment and a name that’s easily searchable.

packages/react-dom/src/events/ReactDOMEventListener.js
rickhanlonii: I see, I misunderstood "we should be able to mutate the value directly without a wrapper function" as not needing a try/catch.

packages/react-noop-renderer/src/ReactNoop.js
gaearon, line 52: You can remove this now?

packages/react-noop-renderer/src/createReactNoop.js
```diff
@@ -949,6 +949,8 @@ function createReactNoop(reconciler: Function, useMutation: boolean) {
     },
 
     unstable_runWithPriority: NoopRenderer.runWithPriority,
+    unstable_InputContinuousLanePriority:
```
gaearon, line 952: Same.

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -66,7 +66,7 @@ const RetryLanePriority: LanePriority = 5;
 const SelectiveHydrationLanePriority: LanePriority = 4;
 
 const IdleHydrationLanePriority: LanePriority = 3;
-const IdleLanePriority: LanePriority = 2;
+export const IdleLanePriority: LanePriority = 2;
```
gaearon, line 69: This change isn't needed now?

Issue #18945, opened 2020-05-18
DevTools: Improve browser extension iframe support


dmail:
When react is inside an iframe, chrome extension for react devtools fails to detect react.

This is because the extension sets `__REACT_DEVTOOLS_GLOBAL_HOOK__` only on the top level window. Apparently it's possible to have `__REACT_DEVTOOLS_GLOBAL_HOOK__` on iframes too by adding `all_frames: true` in extension [manifest.json](https://github.com/facebook/react/blob/c390ab3643612dc08ca4bebadc5b0377e9e7eb79/packages/react-devtools-extensions/chrome/manifest.json#L49). It was done by redux devtools extension in https://github.com/zalmoxisus/redux-devtools-extension/pull/56.

Have you considered adding `all_frames: true` to chrome extension ?

## Steps To Reproduce

1. Create a file `react-main.html`
```html
<!DOCTYPE html>
<html>
  <head>
    <script crossorigin src="https://unpkg.com/react@16/umd/react.development.js"></script>
    <script crossorigin src="https://unpkg.com/react-dom@16/umd/react-dom.development.js"></script>
  </head>
  <body></body>
</html>

```
2. Create a file `react-iframe.html`
```html
<!DOCTYPE html>
<html>
  <head></head>
  <body>
    <iframe src="./react-main.html" />
  </body>
</html>
```
3. Open `react-iframe.html` in chrome

## The current behavior

![react-devtools-not-detected](https://user-images.githubusercontent.com/443639/82196553-21879100-98fa-11ea-932b-b9f49a4079c7.png)

## The expected behavior

![react-devtools-detected](https://user-images.githubusercontent.com/443639/82196588-2cdabc80-98fa-11ea-925d-186e8c999f44.png)

Pull request: https://github.com/facebook/react/pull/18952

bvaughn:
The frame limitation is mentioned in the docs, FWIW:
https://github.com/facebook/react/tree/master/packages/react-devtools

It's not a bug, just a known limitation. We currently suggest using the standalone version (as linked above) or (if you also control the iframe) passing the hook through from the parent:
https://github.com/facebook/react/blob/d897c35ecfb7471b55e41933fd4d993b98dbf600/fixtures/devtools/regression/14.9.html#L9-L11

I don't think that adding the Manifest setting you mentioned would be enough on its own to make the React DevTools extension detect and support versions of React running inside of iframes.

[...]

PR #19345, opened 2020-07-14
Support iframe for React-Devtools browser extension.

Reflex-Gravity:
## Summary

This pull request adds a workaround for the React-DevTools iframe support issue ( #18945).
It contains the modified [regression test](https://github.com/facebook/react/blob/d897c35ecfb7471b55e41933fd4d993b98dbf600/fixtures/devtools/regression/server.js).

For this to work `"all_frames": true` should be set in the browser extension, which @dmail had mentioned.
By adding the below code inside an iframe, fixes this issue:

```javascript
for (const key in __REACT_DEVTOOLS_GLOBAL_HOOK__) {
      __REACT_DEVTOOLS_GLOBAL_HOOK__[key] = parent.__REACT_DEVTOOLS_GLOBAL_HOOK__[key]
}
```
I don't clearly understand the reason why the below code doesn't work, So, I need @bvaughn help to better understand the reason behind this.

```javascript
__REACT_DEVTOOLS_GLOBAL_HOOK__ = parent.__REACT_DEVTOOLS_GLOBAL_HOOK__; 
```

Before:
react_devtools_iframe_

After adding the workaround:
react_dev_tools_issue

## Test Plan

I have modified the regression test, So, that you can test this workaround.
You can run the regression test with the below commands:

1. `node ./fixtures/devtools/regression/server.js`
2. `cd packages/react-devtools-extensions/`
3. `yarn build:chrome`
4. `yarn test:chrome`
5. Open `localhost:3000` in the browser.


bvaughn:
Before, iframes wouldn't work unless you added:
```js
 __REACT_DEVTOOLS_GLOBAL_HOOK__ = parent.__REACT_DEVTOOLS_GLOBAL_HOOK__; 
```

Now they won't work unless you add:
```js
for (const key in __REACT_DEVTOOLS_GLOBAL_HOOK__) {
  __REACT_DEVTOOLS_GLOBAL_HOOK__[key] = parent.__REACT_DEVTOOLS_GLOBAL_HOOK__[key]
}
```

#18945 says "Improve browser extension iframe support" but this feels like it's actually worse.

Reflex-Gravity:
@bvaughn Do you have any clue, what might be causing this? Or any info which will help me.

bvaughn:
No, sorry.

PR #19352, opened 2020-07-14
Tidy up React Scope API

trueadm:
This PR changes the React Scopes (experimental) API, so it's simpler. Previously, you used Scopes like this:

```jsx
const MyScope = React.unstable_createScope()

function MyComponent() {
  return <MyScope>...</MyScope>
}
```

With this PR:

```jsx
const Scope = React.unstable_Scope;

function MyComponent() {
  return <Scope>...</Scope>
}
```

So they work like `Fragment` or other common React exports, and can be used directly as the element type.

The reason we have `unstable_createScope` was because in the older design of the Scopes API, `createScope` used to take an argument. That specific feature was removed ages ago, so it actually makes no sense to have to do `createScope` each time.


gaearon:
I'm not quite following. Wasn't the idea that you declare separate Scope per piece of functionality, similar to Context? How can you be sure of the type of the exposed scope's refs after this change? Presumably any traversal will now "see" unrelated Scopes, unless I'm missing something.

trueadm:
@gaearon 

> Wasn't the idea that you declare separate Scope per piece of functionality, similar to Context? How can you be sure of the type of the exposed scope's refs after this change? Presumably any traversal will now "see" unrelated Scopes, unless I'm missing something.

That's how scopes _used_ to work. They changed a while ago and lost their traversal methods entirely, instead opting to use React Context to traverse up and a method to find child React Context in a sub-tree. This was one of the stems of work we had last year when Sebastian was in London, around making Scopes API "safter" and less super-powered.

packages/react-reconciler/src/ReactFiber.new.js
```diff
@@ -496,6 +496,11 @@ export function createFiberFromTypeAndProps(
         return createFiberFromOffscreen(pendingProps, mode, lanes, key);
       case REACT_LEGACY_HIDDEN_TYPE:
         return createFiberFromLegacyHidden(pendingProps, mode, lanes, key);
+      case REACT_SCOPE_TYPE:
+        if (enableScopeAPI) {
+          return createFiberFromScope(type, pendingProps, mode, lanes, key);
+        }
+        break;
```
gaearon, line 503: This `break` doesn't look right to me. If the feature flag is off, I think we want it to full through to the default case where it would throw with a nice message. Is that not right?

packages/react-reconciler/src/ReactFiber.old.js
```diff
@@ -496,6 +496,11 @@ export function createFiberFromTypeAndProps(
         return createFiberFromOffscreen(pendingProps, mode, lanes, key);
       case REACT_LEGACY_HIDDEN_TYPE:
         return createFiberFromLegacyHidden(pendingProps, mode, lanes, key);
+      case REACT_SCOPE_TYPE:
+        if (enableScopeAPI) {
+          return createFiberFromScope(type, pendingProps, mode, lanes, key);
+        }
+        break;
```
gaearon, line 503: Same, I'm not sure `break` is right here.

packages/react-reconciler/src/ReactFiber.new.js
trueadm: It would still fall through the the error case, which doesn't happen in `default` but rather by the fact we aren't doing a `return`.

trueadm: Ah yes, that's right.

PR #19356, opened 2020-07-14
Update peerDependencies in package.json

MichaelDeBoey:
https://twitter.com/SemanticRelease/status/1249795728253353986

### DIFF
--- a/packages/create-subscription/package.json
+++ b/packages/create-subscription/package.json
@@ -15,7 +15,7 @@
     "cjs/"
   ],
   "peerDependencies": {
-    "react": "^16.3.0"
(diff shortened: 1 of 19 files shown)


rickhanlonii:
@MichaelDeBoey is this something your scripting (I see the same to other projects) or am I missing something?

MichaelDeBoey:
@rickhanlonii It takes less time to do it manually I think, so I didn't script it
I just wanted to update this

gaearon:
This isn't right. The whole purpose of the peerDeps is to show we're not necessarily compatible with future majors.

PR #19364, opened 2020-07-15
Devtools: improve getID guard

trueadm:
Today I ran into an issue where I saw `invalid value used as weak map key` when debugging an internal issue. Unfortunately, I wasn't able to reproduce it but looking through the code. However, I did notice there was a guard in the devtools code that could be strengthened, and thus helping narrow down a codepath that leads to us setting an invalid WeakMap key.

### DIFF
--- a/packages/react-devtools-shared/src/backend/legacy/renderer.js
+++ b/packages/react-devtools-shared/src/backend/legacy/renderer.js
@@ -165,7 +165,7 @@ export function attach(
   }
 
   function getID(internalInstance: InternalInstance): number {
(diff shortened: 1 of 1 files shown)


bvaughn:
Just to clarify, the place this check was strengthened only impacts v15 React, which FB would not be running.

Still seems fine to do though :)

trueadm:
@bvaughn Ah my bad. I'll take another look later at the other codepaths too :)

bvaughn:
Just FYI `backend/renderer.js` is used for v16 and `backend/legacy/renderer.js` is used for v15.

PR #19365, opened 2020-07-15
Revert "Revert #19348"

trueadm:
Reverts facebook/react#19354


trueadm:
Opps, reverted the wrong PR number.

PR #19368, opened 2020-07-15
Fix Suspense-wrapping heuristic

### LINKED ISSUE #18831: Error: "Commit tree does not contain fiber 256. This is a bug in React DevTools."
Describe what you were doing when the bug occurred:
1. Profiling a slow component
In a component that rendered 5000 pre tags with single lines of text in them, that has an unrelated controlled text box is the same component that was typed into while profiling. App hung a while and, when it rendered again the error was in the profiler.

---------------------------------------------
Please do not remove the text below this line
---------------------------------------------

DevTools version: 4.6.0-6cceaeb67

Call stack: at j (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:40:162825)
    at N (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:40:161628)
    at e.getCommitTree (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:40:164582)
    at ec (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:40:339280)
    at ci (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:59620)
    at Ll (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:109960)
    at qc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:102381)
    at Hc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:102306)
    at Vc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:102171)
    at Tc (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/main.js:32:98781)

Component stack: in ec
    in div
    in div
    in div
    in So
    in Unknown
    in n
    in Unknown
    in div
    in div
    in rl
    in Ze
    in fn
    in Ga
    in _s

bvaughn:
Fixes #18831 

**This PR also includes a package version bump** (`16.13.1` -> `17.0.0-alpha.0`) so that DevTools tests will pass CI.

PR #19108 broke DevTools `Suspense` integration.

DevTools uses the renderer's version number to figure out what the tag numbers are:
https://github.com/facebook/react/blob/45eef8b6b5df27c661b3e33e579abc11b563c65b/packages/react-devtools-shared/src/backend/renderer.js#L156-L240

As of #19108, DevTools also decides how to traverse Suspense fibers using the tag values as a form of feature detection:
https://github.com/facebook/react/blob/45eef8b6b5df27c661b3e33e579abc11b563c65b/packages/react-devtools-shared/src/backend/renderer.js#L1241-L1247

The reason this PR also bumps version numbers is that we previously had multiple `Suspense` implementations published under 16.13.1 (both the NPM releases and our CI-built releases that tests run against), so the DevTools heuristic failed.

To prevent this type of regression from happening again, a larger effort needs to be made in the form of follow up PRs to get CI running DevTools tests against a range of React releases (not just what's in master).


packages/react-devtools-shared/src/__tests__/profilingCache-test.js
```diff
@@ -716,4 +716,93 @@ describe('ProfilingCache', () => {
       TestRenderer.create(<Validator commitIndex={0} rootID={rootID} />);
     });
   });
+
+  // See https://github.com/facebook/react/issues/18831
+  it('should not crash during route transitions with Suspense', () => {
```
bvaughn, line 721: This newly added test caught the regression between 16.13.1 (NPM release) and previous DevTools master.

bvaughn:
To be clear, I think we should still move forward with this fix and release an updated version of DevTools since this fixes the issue all open source users (who use Suspense) are currently facing. This can be confirmed by building and running the extension against this repro case I created: https://cx4rd.csb.app/

bvaughn:
CI is failing because of the shallow renderer, which no longer lives in this repo. We can't update to point at v17 of it because that doesn't exist, and the latest version the does exist references react-is v16.

bvaughn:
I'm going to try removing all references to the shallow renderer to see if this fixes our CI failures. We can talk more before committing to this (or we could move forward and revert that specific part of the commit before the actual v17 release) but it would be nice to buy ourselves time without having to commit to v17 of shallow renderer.

bvaughn:
This PR seems to have gotten screwed up by a Circle CI issue, so I'm just going to close it and move to #19373

PR #19372, opened 2020-07-15
Don't build shallow renderer UMD bundles

gaearon:
Fixes the blocker for https://github.com/facebook/react/pull/19368.

We've stopped building CJS bundles for shallow, but we're still building UMDs. They're only available in the UMD folder. I think it's reasonable to drop UMDs for the shallow renderer in our package because they can be found in https://unpkg.com/browse/react-shallow-renderer/umd/ anyway. And since master is already 17-only we can do it now.

Note this doesn't delete shallow renderer integration tests or the entry point. That keeps working.

### DIFF
--- a/scripts/rollup/bundles.js
+++ b/scripts/rollup/bundles.js
@@ -428,19 +428,6 @@ const bundles = [
         ]),
       }),
   },
-  {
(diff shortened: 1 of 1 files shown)


bvaughn:
This change was manually cherry picked into master, so I'm going to close this PR

PR #19374, opened 2020-07-15
Effects list refactor continued: passive effects traversal

bvaughn:
* Adds new `Passive` subtree tag value.
* Bubbles passive flag to ancestors in the case of an unmount.
* Adds recursive traversal for passive effects (mounts and unmounts).
* Removes `pendingPassiveHookEffectsMount` and `pendingPassiveHookEffectsUnmount` arrays from work loop.
* Re-adds sibling and child pointer detaching (temporarily removed in previous PR).
* Addresses some minor TODO comments left over from previous PRs.

---

Co-authored-by: Luna Ruan <luna@fb.com>

### DIFF
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -15,7 +15,16 @@ import type {Fiber} from './ReactInternalTypes';
 import type {Lanes} from './ReactFiberLane';
 
 import getComponentName from 'shared/getComponentName';
-import {Placement, Deletion} from './ReactSideEffectTags';
+import {
+  Deletion,
+  NoEffect,
+  PassiveMask,
+  Placement,
+} from './ReactSideEffectTags';
+import {
+  NoEffect as NoSubtreeTag,
+  Passive as PassiveSubtreeTag,
+} from './ReactSubtreeTags';
 import {
   getIteratorFn,
   REACT_ELEMENT_TYPE,
@@ -293,6 +302,14 @@ function ChildReconciler(shouldTrackSideEffects) {
       returnFiber.deletions = [childToDelete];
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       returnFiber.effectTag |= Deletion;
+
+      // If we are deleting a subtree that contains a passive effect,
+      // mark the parent so that we're sure to traverse after commit and run any unmount functions.
+      const primaryEffectTag = childToDelete.effectTag & PassiveMask;
+      const primarySubtreeTag = childToDelete.subtreeTag & PassiveSubtreeTag;
+      if (primaryEffectTag !== NoEffect || primarySubtreeTag !== NoSubtreeTag) {
+        returnFiber.subtreeTag |= PassiveSubtreeTag;
+      }
     } else {
       deletions.push(childToDelete);
     }
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -15,16 +15,8 @@ import type {Fiber} from './ReactInternalTypes';
 import type {Lanes} from './ReactFiberLane';
 
 import getComponentName from 'shared/getComponentName';
-import {
-  Deletion,
-  NoEffect,
-  PassiveMask,
-  Placement,
-} from './ReactSideEffectTags';
-import {
-  NoEffect as NoSubtreeTag,
-  Passive as PassiveSubtreeTag,
-} from './ReactSubtreeTags';
+import {Deletion, Placement} from './ReactSideEffectTags';
+import {Passive as PassiveSubtreeTag} from './ReactSubtreeTags';
 import {
   getIteratorFn,
   REACT_ELEMENT_TYPE,
@@ -303,13 +295,11 @@ function ChildReconciler(shouldTrackSideEffects) {
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       returnFiber.effectTag |= Deletion;
 
-      // If we are deleting a subtree that contains a passive effect,
-      // mark the parent so that we're sure to traverse after commit and run any unmount functions.
-      const primaryEffectTag = childToDelete.effectTag & PassiveMask;
-      const primarySubtreeTag = childToDelete.subtreeTag & PassiveSubtreeTag;
-      if (primaryEffectTag !== NoEffect || primarySubtreeTag !== NoSubtreeTag) {
-        returnFiber.subtreeTag |= PassiveSubtreeTag;
-      }
+      // We are deleting a subtree that may contain a passive effect.
+      // Mark the parent so we traverse this path after commit and run any unmount functions.
+      // This may cause us to traverse unnecessarily in some cases, but effects are common,
+      // and the cost of over traversing is small (just the path to the deleted node).
+      returnFiber.subtreeTag |= PassiveSubtreeTag;
     } else {
       deletions.push(childToDelete);
     }
--- a/packages/react-reconciler/src/ReactChildFiber.new.js
+++ b/packages/react-reconciler/src/ReactChildFiber.new.js
@@ -16,7 +16,6 @@ import type {Lanes} from './ReactFiberLane';
 
 import getComponentName from 'shared/getComponentName';
 import {Deletion, Placement} from './ReactSideEffectTags';
-import {Passive as PassiveSubtreeTag} from './ReactSubtreeTags';
 import {
   getIteratorFn,
   REACT_ELEMENT_TYPE,
@@ -294,12 +293,6 @@ function ChildReconciler(shouldTrackSideEffects) {
       returnFiber.deletions = [childToDelete];
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       returnFiber.effectTag |= Deletion;
-
-      // We are deleting a subtree that may contain a passive effect.
-      // Mark the parent so we traverse this path after commit and run any unmount functions.
-      // This may cause us to traverse unnecessarily in some cases, but effects are common,
-      // and the cost of over traversing is small (just the path to the deleted node).
-      returnFiber.subtreeTag |= PassiveSubtreeTag;
     } else {
       deletions.push(childToDelete);
     }
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -64,13 +64,9 @@ import {
   Ref,
   Deletion,
   ForceUpdateForLegacySuspense,
-  PassiveMask,
   StaticMask,
 } from './ReactSideEffectTags';
-import {
-  NoEffect as NoSubtreeTag,
-  Passive as PassiveSubtreeTag,
-} from './ReactSubtreeTags';
+import {Passive as PassiveSubtreeTag} from './ReactSubtreeTags';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import {
   debugRenderPhaseSideEffectsForStrictMode,
@@ -2079,15 +2075,11 @@ function updateSuspensePrimaryChildren(
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       workInProgress.effectTag |= Deletion;
 
-      // If we are deleting a subtree that contains a passive effect,
-      // mark the parent so that we're sure to traverse after commit and run any unmount functions.
-      const primaryEffectTag =
-        currentFallbackChildFragment.effectTag & PassiveMask;
-      const primarySubtreeTag =
-        currentFallbackChildFragment.subtreeTag & PassiveSubtreeTag;
-      if (primaryEffectTag !== NoEffect || primarySubtreeTag !== NoSubtreeTag) {
-        workInProgress.subtreeTag |= PassiveSubtreeTag;
-      }
+      // We are deleting a subtree that may contain a passive effect.
+      // Mark the parent so we traverse this path after commit and run any unmount functions.
+      // This may cause us to traverse unnecessarily in some cases, but effects are common,
+      // and the cost of over traversing is small (just the path to the deleted node).
+      workInProgress.subtreeTag |= PassiveSubtreeTag;
     } else {
       deletions.push(currentFallbackChildFragment);
     }
@@ -3073,13 +3065,11 @@ function remountFiber(
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       returnFiber.effectTag |= Deletion;
 
-      // If we are deleting a subtree that contains a passive effect,
-      // mark the parent so that we're sure to traverse after commit and run any unmount functions.
-      const primaryEffectTag = current.effectTag & PassiveMask;
-      const primarySubtreeTag = current.subtreeTag & PassiveSubtreeTag;
-      if (primaryEffectTag !== NoEffect || primarySubtreeTag !== NoSubtreeTag) {
-        returnFiber.subtreeTag |= PassiveSubtreeTag;
-      }
+      // We are deleting a subtree that may contain a passive effect.
+      // Mark the parent so we traverse this path after commit and run any unmount functions.
+      // This may cause us to traverse unnecessarily in some cases, but effects are common,
+      // and the cost of over traversing is small (just the path to the deleted node).
+      returnFiber.subtreeTag |= PassiveSubtreeTag;
     } else {
       deletions.push(current);
     }
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -64,7 +64,13 @@ import {
   Ref,
   Deletion,
   ForceUpdateForLegacySuspense,
+  PassiveMask,
+  StaticMask,
 } from './ReactSideEffectTags';
+import {
+  NoEffect as NoSubtreeTag,
+  Passive as PassiveSubtreeTag,
+} from './ReactSubtreeTags';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import {
   debugRenderPhaseSideEffectsForStrictMode,
@@ -2064,13 +2070,24 @@ function updateSuspensePrimaryChildren(
   if (currentFallbackChildFragment !== null) {
     // Delete the fallback child fragment
     currentFallbackChildFragment.nextEffect = null;
-    currentFallbackChildFragment.effectTag = Deletion;
+    currentFallbackChildFragment.effectTag =
+      (currentFallbackChildFragment.effectTag & StaticMask) | Deletion;
     workInProgress.firstEffect = workInProgress.lastEffect = currentFallbackChildFragment;
     const deletions = workInProgress.deletions;
     if (deletions === null) {
       workInProgress.deletions = [currentFallbackChildFragment];
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       workInProgress.effectTag |= Deletion;
+
+      // If we are deleting a subtree that contains a passive effect,
+      // mark the parent so that we're sure to traverse after commit and run any unmount functions.
+      const primaryEffectTag =
+        currentFallbackChildFragment.effectTag & PassiveMask;
+      const primarySubtreeTag =
+        currentFallbackChildFragment.subtreeTag & PassiveSubtreeTag;
+      if (primaryEffectTag !== NoEffect || primarySubtreeTag !== NoSubtreeTag) {
+        workInProgress.subtreeTag |= PassiveSubtreeTag;
+      }
     } else {
       deletions.push(currentFallbackChildFragment);
     }
@@ -3055,6 +3072,14 @@ function remountFiber(
       returnFiber.deletions = [current];
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       returnFiber.effectTag |= Deletion;
+
+      // If we are deleting a subtree that contains a passive effect,
+      // mark the parent so that we're sure to traverse after commit and run any unmount functions.
+      const primaryEffectTag = current.effectTag & PassiveMask;
+      const primarySubtreeTag = current.subtreeTag & PassiveSubtreeTag;
+      if (primaryEffectTag !== NoEffect || primarySubtreeTag !== NoSubtreeTag) {
+        returnFiber.subtreeTag |= PassiveSubtreeTag;
+      }
     } else {
       deletions.push(current);
     }
--- a/packages/react-reconciler/src/ReactFiberBeginWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberBeginWork.new.js
@@ -64,9 +64,7 @@ import {
   Ref,
   Deletion,
   ForceUpdateForLegacySuspense,
-  StaticMask,
 } from './ReactSideEffectTags';
-import {Passive as PassiveSubtreeTag} from './ReactSubtreeTags';
 import ReactSharedInternals from 'shared/ReactSharedInternals';
 import {
   debugRenderPhaseSideEffectsForStrictMode,
@@ -2066,20 +2064,12 @@ function updateSuspensePrimaryChildren(
   if (currentFallbackChildFragment !== null) {
     // Delete the fallback child fragment
     currentFallbackChildFragment.nextEffect = null;
-    currentFallbackChildFragment.effectTag =
-      (currentFallbackChildFragment.effectTag & StaticMask) | Deletion;
     workInProgress.firstEffect = workInProgress.lastEffect = currentFallbackChildFragment;
     const deletions = workInProgress.deletions;
     if (deletions === null) {
       workInProgress.deletions = [currentFallbackChildFragment];
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       workInProgress.effectTag |= Deletion;
-
-      // We are deleting a subtree that may contain a passive effect.
-      // Mark the parent so we traverse this path after commit and run any unmount functions.
-      // This may cause us to traverse unnecessarily in some cases, but effects are common,
-      // and the cost of over traversing is small (just the path to the deleted node).
-      workInProgress.subtreeTag |= PassiveSubtreeTag;
     } else {
       deletions.push(currentFallbackChildFragment);
     }
@@ -3064,12 +3054,6 @@ function remountFiber(
       returnFiber.deletions = [current];
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       returnFiber.effectTag |= Deletion;
-
-      // We are deleting a subtree that may contain a passive effect.
-      // Mark the parent so we traverse this path after commit and run any unmount functions.
-      // This may cause us to traverse unnecessarily in some cases, but effects are common,
-      // and the cost of over traversing is small (just the path to the deleted node).
-      returnFiber.subtreeTag |= PassiveSubtreeTag;
     } else {
       deletions.push(current);
     }
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -69,7 +69,6 @@ import {
   Snapshot,
   Update,
   Passive,
-  PassiveUnmountPendingDev,
 } from './ReactSideEffectTags';
 import getComponentName from 'shared/getComponentName';
 import invariant from 'shared/invariant';
@@ -882,17 +881,6 @@ function commitUnmount(
                 effect.tag |= HookHasEffect;
 
                 current.effectTag |= Passive;
-
-                if (__DEV__) {
-                  // This flag is used to avoid warning about an update to an unmounted component
-                  // if the component has a passive unmount scheduled.
-                  // Presumably the listener would be cleaned up by that unmount.
-                  current.effectTag |= PassiveUnmountPendingDev;
-                  const alternate = current.alternate;
-                  if (alternate !== null) {
-                    alternate.effectTag |= PassiveUnmountPendingDev;
-                  }
-                }
 
                 schedulePassiveEffectCallback();
               } else {
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -881,19 +881,6 @@ function commitUnmount(
               if ((tag & HookPassive) !== NoHookEffect) {
                 effect.tag |= HookHasEffect;
 
-                // subtreeTags bubble in resetChildLanes which doens't get called for unmounted subtrees.
-                // So in the case of unmounts, we need to bubble passive effects explicitly.
-                let ancestor = current.return;
-                while (ancestor !== null) {
-                  ancestor.subtreeTag |= PassiveSubtreeTag;
-                  const alternate = ancestor.alternate;
-                  if (alternate !== null) {
-                    alternate.subtreeTag |= PassiveSubtreeTag;
-                  }
-
-                  ancestor = ancestor.return;
-                }
-
                 current.effectTag |= Passive;
 
                 if (__DEV__) {
--- a/packages/react-reconciler/src/ReactFiberCommitWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCommitWork.new.js
@@ -68,6 +68,8 @@ import {
   Placement,
   Snapshot,
   Update,
+  Passive,
+  PassiveUnmountPendingDev,
 } from './ReactSideEffectTags';
 import getComponentName from 'shared/getComponentName';
 import invariant from 'shared/invariant';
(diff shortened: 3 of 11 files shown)


gaearon:
Is the delta link out of date? I see no changes.

bvaughn:
@gaearon Whoops, the other two PRs have landed now :smile: You can just view it normally. I'll delete that line.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -892,7 +879,35 @@ function commitUnmount(
             const {destroy, tag} = effect;
             if (destroy !== undefined) {
               if ((tag & HookPassive) !== NoHookEffect) {
-                enqueuePendingPassiveHookEffectUnmount(current, effect);
+                effect.tag |= HookHasEffect;
+
+                // subtreeTags bubble in resetChildLanes which doens't get called for unmounted subtrees.
+                // So in the case of unmounts, we need to bubble passive effects explicitly.
```
acdlite, line 885: I was thinking we could move all this logic to the passive commit phase. We would have a separate subtree tag that represents whether there are mounted passive effects anywhere in the tree. It would always be set regardless of whether the dependencies changed, or whether there's an unmount callback.

That means we might visit some nodes that have effects but don't have destroy effects, but that's not a big deal.

We'd use this same flag in the Offscreen implementation to find effects that need to be mounted/unmounted.

acdlite: > I was thinking we could move all this logic to the passive commit phase.

Sorry, that was a confusing way to phrase my suggestion. I meant to say, we can avoid bubbling the `subtreeTag` here  by adding a new bit that is always set for any parent of a mounted effect.

bvaughn: Just to make sure I'm understanding your suggestion correctly- the problem this loop is addressing is that effects deep inside of an unmounted tree aren't reflected by the `subtreeTag` value unless we recurse and find them. Setting a special bit for mounted effects wouldn't help– since we always reset `subtreeTag` to `NoSubtreeEffect` during cloning– unless we were to just never reset that special bit.

So the relevant line in `createWorkInProgress` would become:
```js
workInProgress.subtreeTag &= SpecialNewTagWeAreTalkingAbout;
```

Is this what you had in mind?

acdlite: Yeah that's right. We're planning a similar type of tag to track which trees contain portals. So it could be `&= StaticTags` or something like that, where `StaticTags` is the union of all these types of bits.

bvaughn: Okay. This makes perfect sense then. I'll make the change.

acdlite: Although you need to reset the bit when all the children that have mounted effects are deleted. So I think this is more like the tree duration flags. You need a bit that represents "self", and a bit that represents the children.

acdlite: I guess the "self" tags go on `effectTag` field. Does what I'm saying make sense?

bvaughn: I think 157aa81d7 is the change we discussed.

bvaughn: Huh, your last two comments did not show up until after I pushed [157aa81](https://github.com/facebook/react/pull/19374/commits/157aa81d7b7fda4dd7225c474b20ef0214eb1d96) and commented.

Yes, `effectTag` is the equivalent of `selfBaseDuration` in this case.

> Although you need to reset the bit when all the children that have mounted effects are deleted. So I think this is more like the tree duration flags. You need a bit that represents "self", and a bit that represents the children.

I don't know how to achieve this without deeply traversing the entire subtree (not just the unmounted bit). How else would we know there were no more passive effects? After all, we aren't counting them.

bvaughn: I believe that as of [this change](https://github.com/facebook/react/pull/19374/commits/157aa81d7b7fda4dd7225c474b20ef0214eb1d96) we have two potential over-traversal cases:
* Bailout, although we can avoid that by checking for it in `flushPassiveMountEffects`.
* After the last passive effect has been unmounted in a subtree, we might still traverse part of the subtree (since we never reset the `Static` effects).

Is this second over-traversal safe? I think it's what your previous comment is pointing out, but I don't see how your suggestion for addressing it would work.

lunaruan: One potential issue without the bailout case is that Offscreen children also satisfy the condition for bailing out. I could check whether a fiber is in an Offscreen subtree but it seems slightly messy, so I wonder if there's a better way to determine whether or not children not being cloned is because of bailout or because of Offscreen (perhaps a tag on the fiber indicating that we should stop traversing?)

bvaughn: We talked out of band and a summary is:
* We should add this new static bit to `effectTag` rather than `subtreeTag`.
* The static bit will then bubble to `subtreeTag` during complete phase for things that are still mounted.

bvaughn:
@acdlite @lunaruan I think 15288ed is the change we discussed earlier.

packages/react-reconciler/src/ReactChildFiber.new.js
```diff
@@ -293,6 +302,14 @@ function ChildReconciler(shouldTrackSideEffects) {
       returnFiber.deletions = [childToDelete];
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       returnFiber.effectTag |= Deletion;
+
+      // If we are deleting a subtree that contains a passive effect,
+      // mark the parent so that we're sure to traverse after commit and run any unmount functions.
+      const primaryEffectTag = childToDelete.effectTag & PassiveMask;
+      const primarySubtreeTag = childToDelete.subtreeTag & PassiveSubtreeTag;
+      if (primaryEffectTag !== NoEffect || primarySubtreeTag !== NoSubtreeTag) {
```
acdlite, line 310: Maybe this optimization isn't worth it and we  should always set the PassiveSubtreeTag on the parent when a child is deleted. i.e. add Deletion to PassiveMask.

Effects are very common — any large tree is almost certain to have at least one. So the optimization really only kicks in when you delete a leaf node. Not sure it's worth the extra code.

I'm almost tempted to say that PassiveStatic isn't worth it for the same reason; but since we're planning to have other static tags for things like portals, PassiveStatic doesn't add any net additional overhead, so might as well keep it.

What do you think?

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -1013,8 +1015,11 @@ function commitNestedUnmounts(
 }
 
 function detachFiberMutation(fiber: Fiber) {
-  // Cut off the return pointers to disconnect it from the tree. Ideally, we
-  // should clear the child pointer of the parent alternate to let this
+  // Cut off the return pointers to disconnect it from the tree.
+  // Note that we can't clear child or sibling pointers yet,
+  // because they may be required for passive effects.
+  // These pointers will be cleared in a separate pass.
+  // Ideally, we should clear the child pointer of the parent alternate to let this
```
acdlite, line 1022: If a Deletion always gets a passive effect scheduled on it, as suggested above, we could move this whole function to the passive phase. (Except maybe `return`  and `alternate`, since those affect semantics, like if you call `setState` on an unmounted component.)

bvaughn: If we don't clear `return` during layout, it would break a few things:
* The unmounted state warning
* How events bubble for the new API Dominic has been working on

That being said, it would *fix* a limitation of the Profiler API when it comes to measuring time spent in passive effect cleanup. That time currently doesn't get reported during an unmount (either a regular unmount or because of an error boundary) because the `return` pointer has been detached previously and so it can't bubble up to the nearest Profiler node. If we deferred detaching it until after passive, that time would be reported which is a plus.

Not sure if there would be a viable workaround for the event case, but for the unmounted state warning- we *could* walk the `return` path and look for Fibers with `Deletion` effects scheduled, except that this would have false positives for deletions that were scheduled and not committed so that's a bummer.

bvaughn: That being said, I believe we can safely defer everything *except* the `return` pointer cleanup to the passive effects phase which would be *slightly* better. I'll do that...

packages/react-reconciler/src/ReactChildFiber.new.js
bvaughn: > Maybe this optimization isn't worth it and we should always set the PassiveSubtreeTag on the parent when a child is deleted.

Given that the concept of static flags seems like a positive– and so keeping `PassiveStatic` around doesn't add any net overhead– the only question seems to be whether the few extra bytes required for us to check `effectTag` and `subtreeTag` are worth it, yeah?

With the current check, we might over traverse in the event that a child was deleted that had a passive effect, but no cleanup function. (Seems unlikely.) If we removed this check, we might over traverse any time a leaf node is deleted, period. Still, I guess that wouldn't be *that bad* since we'd only traverse the path to where the leaf was deleted.

TBH I think the cost is pretty small on either side (number or bytes or potential over traversing). Happy to remove the check for now.

acdlite: > If we removed this check, we might over traverse any time a leaf node is deleted, period. Still, I guess that wouldn't be that bad since we'd only traverse the path to where the leaf was deleted.

Yeah I think that's fine. It's also related to the `detachFiber` discussion. Since we're moving most of that work to the passive phase, then we need to visit every deletion during the passive phase anyway.

packages/react-reconciler/src/ReactChildFiber.new.js
```diff
@@ -293,6 +294,12 @@ function ChildReconciler(shouldTrackSideEffects) {
       returnFiber.deletions = [childToDelete];
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       returnFiber.effectTag |= Deletion;
+
+      // We are deleting a subtree that may contain a passive effect.
+      // Mark the parent so we traverse this path after commit and run any unmount functions.
+      // This may cause us to traverse unnecessarily in some cases, but effects are common,
+      // and the cost of over traversing is small (just the path to the deleted node).
+      returnFiber.subtreeTag |= PassiveSubtreeTag;
```
acdlite, line 302: Could you instead add the Deletion bit to PassiveMask? Then you don't have to do anything extra in this function.

packages/react-reconciler/src/ReactChildFiber.new.js
bvaughn: That's a good point.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2064,13 +2066,20 @@ function updateSuspensePrimaryChildren(
   if (currentFallbackChildFragment !== null) {
     // Delete the fallback child fragment
     currentFallbackChildFragment.nextEffect = null;
-    currentFallbackChildFragment.effectTag = Deletion;
+    currentFallbackChildFragment.effectTag =
+      (currentFallbackChildFragment.effectTag & StaticMask) | Deletion;
```
acdlite, line 2070: Don't need the Deletion tag here because it goes on the parent  (`workInProgress` below).

acdlite: So I think this entire line can be removed.

packages/react-reconciler/src/ReactChildFiber.new.js
bvaughn: That's a good point I guess. I hadn't considered that aspect.

I consider the "detach" methods as kind of an optional optimization.

packages/react-reconciler/src/ReactFiberBeginWork.new.js
```diff
@@ -2064,13 +2066,20 @@ function updateSuspensePrimaryChildren(
...
     if (deletions === null) {
       workInProgress.deletions = [currentFallbackChildFragment];
       // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
       workInProgress.effectTag |= Deletion;
+
+      // We are deleting a subtree that may contain a passive effect.
+      // Mark the parent so we traverse this path after commit and run any unmount functions.
+      // This may cause us to traverse unnecessarily in some cases, but effects are common,
+      // and the cost of over traversing is small (just the path to the deleted node).
+      workInProgress.subtreeTag |= PassiveSubtreeTag;
```
acdlite, line 2082: Same comment as https://github.com/facebook/react/pull/19374/files#r462472081

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -1013,29 +1015,19 @@ function commitNestedUnmounts(
...
-  fiber.lastEffect = null;
-  fiber.memoizedProps = null;
-  fiber.memoizedState = null;
-  fiber.pendingProps = null;
+  // one so we'll settle for GC:ing the subtree of this child.
+  // This child itself will be GC:ed when the parent updates the next time.
+  //
+  // Note that we can't clear child or sibling pointers yet.
+  // They're needed for passive effects and for findDOMNode.
+  // We defer those fields, and all other cleanup, to the passive phase (see detachFiberAfterEffects).
```
acdlite, line 1029: Let's detach `alternate` and `alternate.return`, too. (Because in the `setState` algorithm, we traverse up the return path of each fiber in the pair.)

packages/react-reconciler/src/ReactFiberBeginWork.new.js
bvaughn: Right. I assumed that comment applied to all 4 places we were bubbling :)

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1270,7 +1271,7 @@ function mountEffect(
     }
   }
   return mountEffectImpl(
-    UpdateEffect | PassiveEffect,
+    UpdateEffect | PassiveEffect | PassiveStaticEffect,
```
acdlite, line 1274: Nit: Let's move this to `mountEffectImpl`. Makes it less likely we'll forget about it.

packages/react-reconciler/src/ReactFiberHooks.new.js
```diff
@@ -1288,7 +1289,7 @@ function updateEffect(
     }
   }
   return updateEffectImpl(
-    UpdateEffect | PassiveEffect,
+    UpdateEffect | PassiveEffect | PassiveStaticEffect,
```
acdlite, line 1292: Don't think you need it here, only on mount. Since it's not removed during clone.

packages/react-reconciler/src/ReactFiberHydrationContext.new.js
```diff
@@ -124,12 +125,18 @@ function deleteHydratableInstance(
...
   if (deletions === null) {
     returnFiber.deletions = [childToDelete];
     // TODO (effects) Rename this to better reflect its new usage (e.g. ChildDeletions)
     returnFiber.effectTag |= Deletion;
+
+    // We are deleting a subtree that may contain a passive effect.
+    // Mark the parent so we traverse this path after commit and run any unmount functions.
+    // This may cause us to traverse unnecessarily in some cases, but effects are common,
+    // and the cost of over traversing is small (just the path to the deleted node).
+    returnFiber.subtreeTag |= PassiveSubtreeTag;
```
acdlite, line 139: Same comment as https://github.com/facebook/react/pull/19374/files#r462472081

packages/react-reconciler/src/ReactChildFiber.new.js
bvaughn: Hm, yes, I think that would work. Sorry if I misinterpreted your original suggestion. I realize now that you were probably suggesting that instead of this.

packages/react-reconciler/src/ReactFiberHooks.new.js
bvaughn: `mountEffectImpl` is used for layout effects also. This static flag only applies to passive effects.

packages/react-reconciler/src/ReactFiberHooks.new.js
bvaughn: Yeah I guess that's fair.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2538,7 +2560,23 @@ function commitMutationEffectsDeletions(
         captureCommitPhaseError(childToDelete, error);
       }
     }
-    // Don't clear the Deletion effect yet; we also use it to know when we need to detach refs later.
+
+    // If there are no pending passive effects, it's safe to detach remaining pointers now.
+    const primarySubtreeTag = childToDelete.subtreeTag & PassiveSubtreeTag;
+    const primaryEffectTag = childToDelete.effectTag & PassiveMask;
+    if (primarySubtreeTag === NoSubtreeTag && primaryEffectTag === NoEffect) {
+      detachFiberAfterEffects(childToDelete);
```
acdlite, line 2568: Assuming we always schedule a passive effect for deletions, don't need this block now.

bvaughn: Yes. I know. Adding `Deletion` to the static mask means we can remove 4 of these places. ;)

acdlite:
Neat! This is an exciting PR

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
bvaughn: Ah, yeah fair enough. It is safe to remove now.

I'm a tiny bit concerned that it's not intuitive how re-adding that optimization would impact a few of these places, but maybe that won't actually be a concern.

bvaughn:
Thanks for the feedback @acdlite! 😄

Made the suggested tweaks:
* Added `Deletion` effect to `PassiveMask`.
  * Removed the no longer needed explicit bubbling of `Passive` `subtreeTag` during deletion.
* Removed unnecessary deletion `effectTag` from fallback child.
* Detached `alternate.return` as well during layout.
* Removed `PassiveStaticEffect` from hook update.
* Removed unnecessary call to `detachFiberAfterEffects`.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2396,19 +2417,20 @@ function commitMutationEffects(
...
-        root,
-        renderPriorityLevel,
-      );
+    const deletions = fiber.deletions;
+    if (deletions !== null) {
+      commitMutationEffectsDeletions(deletions, root, renderPriorityLevel);
 
-      // TODO (effects) Don't clear this yet; we may need to cleanup passive effects
-      fiber.deletions = null;
+      // If there are no pending passive effects, clear the deletions Array.
```
acdlite, line 2424: Can move this to passive phase, too

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2645,44 +2676,239 @@ export function enqueuePendingPassiveProfilerEffect(fiber: Fiber): void {
...
-  if (__DEV__) {
-    fiber.effectTag |= PassiveUnmountPendingDev;
-    const alternate = fiber.alternate;
-    if (alternate !== null) {
-      alternate.effectTag |= PassiveUnmountPendingDev;
+function flushPassiveMountEffects(firstChild: Fiber): void {
+  let fiber = firstChild;
+  while (fiber !== null) {
+    const didBailout =
+      fiber.alternate !== null && fiber.alternate.child === fiber.child;
```
acdlite, line 2688: The primary subtree tag check should be sufficient, right? That's what we do in the other phases.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -892,7 +879,22 @@ function commitUnmount(
             const {destroy, tag} = effect;
             if (destroy !== undefined) {
               if ((tag & HookPassive) !== NoHookEffect) {
-                enqueuePendingPassiveHookEffectUnmount(current, effect);
+                effect.tag |= HookHasEffect;
```
acdlite, line 882: I wonder if we can get rid of the rest of this block. I can't remember all the details for how HookHasEffect is used.

(This can be pushed to a follow up. Maybe drop a TODO with context.)

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -892,7 +879,22 @@ function commitUnmount(
...
+
+                current.effectTag |= Passive;
```
acdlite, line 884: Trying to move away from mutating the current tree.

I think we can move these checks to passive phase after paint?

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -892,7 +879,22 @@ function commitUnmount(
...
+
+                if (__DEV__) {
+                  // This flag is used to avoid warning about an update to an unmounted component
+                  // if the component has a passive unmount scheduled.
+                  // Presumably the listener would be cleaned up by that unmount.
+                  current.effectTag |= PassiveUnmountPendingDev;
```
acdlite, line 890: I think this could go on the deleted fiber only. In the warning function, can search up return path.

(Non-blocking)

packages/react-reconciler/src/ReactFiberCommitWork.new.js
```diff
@@ -892,7 +879,22 @@ function commitUnmount(
...
+                  const alternate = current.alternate;
+                  if (alternate !== null) {
+                    alternate.effectTag |= PassiveUnmountPendingDev;
+                  }
+                }
+
+                schedulePassiveEffectCallback();
```
acdlite, line 897: Don't think we need this now that we check and schedule at the root, too.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2170,6 +2182,17 @@ function commitRootImpl(root, renderPriorityLevel) {
       markLayoutEffectsStopped();
     }
 
+    // If there are pending passive effects, schedule a callback to process them.
+    if ((root.current.subtreeTag & PassiveSubtreeTag) !== NoSubtreeTag) {
```
acdlite, line 2186: Nit: `root.current` is `finishedWork`

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
bvaughn: Hm, true. Same concern as previously expressed about the number of subtle changes (many non-observable) hinge on the fact that we removed ab optimization though.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
bvaughn: Oh, duh.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
bvaughn: What do you mean by, "the rest of this block"?

We still need `PassiveUnmountPendingDev` in DEV in order to avoid over-warning for state updates triggered by passive effect subscriptions.

I'm not sure what you want the TODO to say.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
bvaughn: Hm, maybe. We'd need to track when we're inside of an unmounted tree, so we would know to always run passive effects (regardless of `HookHasEffect`).

packages/react-reconciler/src/ReactFiberCommitWork.new.js
acdlite: > What do you mean by, "the rest of this block"?

Everything inside `if ((tag & HookPassive) !== NoHookEffect) { ... }`

The reason I'm asking is because eventually we want to stop traversing the whole deleted subtree during the  synchronous commit phase. Instead we'd only traverse the paths that truly need to do synchronous clean-up, like ref detaching, `useLayoutEffect` clean up, and `componentWillUnmount`. (Longer term, we want to replace `useLayoutEffect` with a render phase equivalent, and `componentWillUnmount` is class-only. So it's really only ref detaching that we'll need to do.)

> I'm not sure what you want the TODO to say.

Can just be "TODO: Consider if we can move this block out of the synchronous commit phase"

bvaughn: Sure. I can add such a comment so we can follow up and consider the change later 👍

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2645,44 +2676,239 @@ export function enqueuePendingPassiveProfilerEffect(fiber: Fiber): void {
...
+      for (let i = 0; i < deletions.length; i++) {
+        const fiberToDelete = deletions[i];
+        // If this fiber (or anything below it) has passive effects then traverse the subtree.
+        const primaryEffectTag = fiberToDelete.effectTag & PassiveMask;
+        const primarySubtreeTag = fiberToDelete.subtreeTag & PassiveSubtreeTag;
+        if (
+          primarySubtreeTag !== NoSubtreeTag ||
+          primaryEffectTag !== NoEffect
+        ) {
+          flushPassiveUnmountEffects(fiberToDelete);
```
acdlite, line 2799: Ah I didn't notice this was a recursive call.

I think this should be a different function, analagous to `commitNestedUnmounts`. They do slightly different things. This new function gets passed a fiber from the old tree, not the new tree. So for example, it doesn't have to check for nested deletions, because you're already inside a deleted tree.

(This might be related to the `didBailout` thing?)

bvaughn: They don't really do different things. They traverse a tree and execute any passive effects that are scheduled.

> This might be related to the `didBailout` thing?

Which thing?

packages/react-reconciler/src/ReactFiberCommitWork.new.js
bvaughn: It's still necessary.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
acdlite: The key difference is that the top-most `flushPassiveUnmountEffects` gets passed a "previous" fiber, from the old tree,  whereas the new function I'm proposing gets passed a "next" fiber, from the just-finished tree.

It goes back to the `didBailout` conversation from the last PR. The `effectTag` and `subtreeTag` on the "previous" tree represents a render that already happened. So we don't want to rely on those being correct for the current render, because that requires us to reset them in the commit phase. (However, it's fine to read the static part `effectTag`, and their corresponding accumulation in `subtreeTag`, since aren't specific to a particular render.)

Same with the `deletions` field. We shouldn't read `previous.deletions` because those deletions aren't part of the current commit phase; we already processed those. In other words, we shouldn't have to null out `finishedWork.deletions` once we're done processing them; I think it's only necessary right now because of the issue I'm describing.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
bvaughn: Hm, maybe...

If we weren't careful, we might miss legit warnings that way. We wouldn't want to incorrectly suppress a warning for a state update for a component that's within a tree that happened to have a subtree tag, even if the component in question doesn't have a passive effect scheduled. I guess we could walk the return path to see if an ancestor has `PassiveUnmountPendingDev`, then walk the Fiber receiving the update to see if its effect list contains a passive effect.

But as I write that, I realize that if we are willing to walk the effect list, we don't really need to look at an ancestor. We could just check for a `Passive` effect on the fiber an update is being scheduled for, then walk its effects list to confirm there's actually a destroy function. So I think we could maybe remove the `PassiveUnmountPendingDev` bit entirely?

Let me take a pass at that.

bvaughn: So essentially this:
```js
// If there are pending passive effects unmounts for this Fiber,
// we can assume that they would have prevented this update.
if ((fiber.effectTag & PassiveUnmountPendingDev) !== NoEffect) {
  return;
}
```

Would become this:
```js
const alternate = fiber.alternate;
if ((fiber.effectTag & PassiveStatic) !== NoEffect) {
  const updateQueue: FunctionComponentUpdateQueue | null = (fiber.updateQueue: any);
  if (updateQueue !== null) {
    const lastEffect = updateQueue.lastEffect;
    if (lastEffect !== null) {
      const firstEffect = lastEffect.next;

      let effect = firstEffect;
      do {
        if (effect.destroy !== undefined) {
          if ((effect.tag & HookPassive) !== NoHookEffect) {
            return;
          }
        }
        effect = effect.next;
      } while (effect !== firstEffect);
    }
  }
}
```

The latter is clearly worse, but it's lazy *and* it's DEV only code, so it seems like an improvement.

bvaughn: I don't know. I'll make this change but I don't know that it's worth doing. :)

packages/react-reconciler/src/ReactFiberCommitWork.new.js
acdlite: Could you explain why? I believe you, just curious.

The reason I guess this would work is because a Deletion tag results in a PassiveSubtree tag on all the ancestors up to the root, which we would then detect with this check: https://github.com/facebook/react/pull/19374/files#diff-71727dd0c0ae3972b0f7725a87294f4cR2186

(Not a blocker regardless, btw)

packages/react-reconciler/src/ReactFiberCommitWork.new.js
bvaughn: I'm going to leave this ^ to the TODO as well.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -2645,44 +2669,225 @@ export function enqueuePendingPassiveProfilerEffect(fiber: Fiber): void {
...
+        detachFiberAfterEffects(fiberToDelete);
+      }
+
+      // Clear deletions now that passive effects have  been procssed.
+      fiber.deletions = null;
+    }
+
+    const didBailout =
+      fiber.alternate !== null && fiber.alternate.child === fiber.child;
+    if (!didBailout) {
```
acdlite, line 2799: Nit: This one, too

packages/react-reconciler/src/ReactFiberCommitWork.new.js
bvaughn: Yeah. My bad, I should have left a more thorough response initially.

A `Deletion` effect would get bubbled into a `PassiveSubtree` tag, but that bubbling happens during the render phase, whereas this function is in the commit phase. (There's no bubbling after this point.)

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
bvaughn: > So for example, it doesn't have to check for nested deletions, because you're already inside a deleted tree.

This is a reasonable point. I can split the two functions apart, if you think the added bytes is better than the unnecessary null check.

> In other words, we shouldn't have to null out `finishedWork.deletions` once we're done processing them

Clearing the `deletions` array was unnecessary. I'll remove that line.

bvaughn: @acdlite Let me know if ae37c7c addresses your suggestion here?

bvaughn:
There's been a lot of back and forth. I think I've made all of the suggested changes (that I can make). Let me know if I missed something!

packages/react-reconciler/src/ReactFiberCommitWork.new.js
acdlite: There's a Deletion effect on the ancestor that got deleted, though, which does get bubbled during render phase

packages/react-reconciler/src/ReactFiberCommitWork.new.js
acdlite: Cool, yeah this isn't super important either way. I was mostly thinking of what we'd have to do in the future to avoid visiting passive effect fibers during the layout phase.

packages/react-reconciler/src/ReactFiberCommitWork.new.js
bvaughn: The `return` pointer has been cleared already though

bvaughn:
No problem. Thanks for the fast feedback!

packages/react-reconciler/src/ReactFiberCommitWork.new.js
bvaughn: Sorry, please disregard my other comment. I thought I was responding to a different thread when I got the email notification. :smile:

PR #19334, opened 2020-07-13
Scheduling profiler updates

bvaughn:
* Enable `enableSchedulingProfiler` statically for experimental+profiling builds and www+profiling builds.
* Synced debug tracing and scheduling profiler code from old -> new fork.
* Update test `@gate` conditions.

To verify my manual syncing over of these features, I used diff:

### ReactFiberClassComponent
```diff
12c12
< import type {UpdateQueue} from './ReactUpdateQueue.old';
---
> import type {UpdateQueue} from './ReactUpdateQueue.new';
23c23
< import ReactStrictModeWarnings from './ReactStrictModeWarnings.old';
---
> import ReactStrictModeWarnings from './ReactStrictModeWarnings.new';
31c31
< import {resolveDefaultProps} from './ReactFiberLazyComponent.old';
---
> import {resolveDefaultProps} from './ReactFiberLazyComponent.new';
44c44
< } from './ReactUpdateQueue.old';
---
> } from './ReactUpdateQueue.new';
52,53c52,53
< } from './ReactFiberContext.old';
< import {readContext} from './ReactFiberNewContext.old';
---
> } from './ReactFiberContext.new';
> import {readContext} from './ReactFiberNewContext.new';
58c58
< } from './ReactFiberWorkLoop.old';
---
> } from './ReactFiberWorkLoop.new';
```

### ReactFiberHooks
```diff
48,49c48,49
< import {readContext} from './ReactFiberNewContext.old';
< import {createDeprecatedResponderListener} from './ReactFiberDeprecatedEvents.old';
---
> import {readContext} from './ReactFiberNewContext.new';
> import {createDeprecatedResponderListener} from './ReactFiberDeprecatedEvents.new';
69c69
< } from './ReactFiberWorkLoop.old';
---
> } from './ReactFiberWorkLoop.new';
74c74
< import {markWorkInProgressReceivedUpdate} from './ReactFiberBeginWork.old';
---
> import {markWorkInProgressReceivedUpdate} from './ReactFiberBeginWork.new';
81,82c81,82
< } from './SchedulerWithReactIntegration.old';
< import {getIsHydrating} from './ReactFiberHydrationContext.old';
---
> } from './SchedulerWithReactIntegration.new';
> import {getIsHydrating} from './ReactFiberHydrationContext.new';
93c93
< } from './ReactMutableSource.old';
---
> } from './ReactMutableSource.new';
771c771
<         // its priority which is a derived and not reversible value.
---
>         // its priority which is a derived and not reverseable value.
[...]

PR #19375, opened 2020-07-15
Make debug tracing and scheduling profiler flags static

bvaughn:
#19334 was reverted in #19366 because it caused test failures during a www sync.

This PR pulls in only part of that change, so we can see if it was what caused the failure:
* Make `enableSchedulingProfiler` static for experimental+profiling builds and www+profiling builds to remove flag checks from hot paths.


bvaughn:
Merged via #19396

PR #19376, opened 2020-07-15
Sync debug tracing and scheduler profiling to new fork

bvaughn:
#19334 was reverted in #19366 because it caused test failures during a www sync.

This PR builds on top of PR #19375 and is comprised of a single commit, [79637ec](https://github.com/facebook/react/pull/19376/commits/79637ec670db7eb3eff3966b51e179415c5aac62?w=1).

* Syncs debug tracing and scheduling profiler code from old -> new fork.
  * (This was pretty manual given the ongoing effects list refactor.)
* Enables scheduling profiler for experimental OSS builds.
* Updates test `@gate` conditions.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.old.js
@@ -1361,7 +1361,7 @@ function handleError(root, thrownValue): void {
         // sibling, or the parent if there are no siblings. But since the root
         // has no siblings nor a parent, we set it to null. Usually this is
         // handled by `completeUnitOfWork` or `unwindWork`, but since we're
-        // interntionally not calling those, we need set it here.
+        // intentionally not calling those, we need set it here.
         // TODO: Consider calling `unwindWork` to pop the contexts.
         workInProgress = null;
         return;
--- a/packages/shared/forks/ReactFeatureFlags.www.js
+++ b/packages/shared/forks/ReactFeatureFlags.www.js
@@ -26,7 +26,6 @@ export const {
   deferRenderPhaseUpdateToNextBatch,
   decoupleUpdatePriorityFromScheduler,
   enableDebugTracing,
-  enableSchedulingProfiler,
 } = dynamicFeatureFlags;
 
 // On WWW, __EXPERIMENTAL__ is used for a new modern build.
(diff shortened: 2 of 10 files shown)


packages/shared/forks/ReactFeatureFlags.www.js
```diff
@@ -36,6 +34,10 @@ export const {
 export const enableProfilerTimer = __PROFILE__;
 export const enableProfilerCommitHooks = __PROFILE__;
 
+// TODO: These features do not yet exist in the new reconciler fork.
+export const enableDebugTracing = !__VARIANT__;
```
acdlite, line 38: What's the reason for this change? Is `enableDebugTracing` no longer a dynamic flag in www?

bvaughn: I guess it's okay for it to be, since it's wrapped in `__DEV__ `checks. I'll move it back only.

packages/shared/forks/ReactFeatureFlags.www.js
```diff
@@ -36,6 +34,10 @@ export const {
...
+export const enableSchedulingProfiler = !__VARIANT__ && __PROFILE__;
```
acdlite, line 39: I think this should just be `__PROFILE__`.

packages/shared/forks/ReactFeatureFlags.www.js
acdlite: I think it's fine either way because it only triggers if the mode is set. But if it's meant to be enabled, this can just be `true`.

bvaughn: It's not meant to be enabled always, because it's pretty noisy in the console. It's meant to be something we flip on manually when we're debugging.

PR #19375 is what makes the flag changes though; I've added some additional commenting about when/why you'd want to use the flag.

packages/shared/forks/ReactFeatureFlags.www.js
bvaughn: Yeah, you're right. I hadn't updated that in this branch yet. My bad.

packages/react-reconciler/src/ReactFiberWorkLoop.old.js
```diff
@@ -1361,7 +1361,7 @@ function handleError(root, thrownValue): void {
         // sibling, or the parent if there are no siblings. But since the root
         // has no siblings nor a parent, we set it to null. Usually this is
         // handled by `completeUnitOfWork` or `unwindWork`, but since we're
-        // interntionally not calling those, we need set it here.
+        // intentionally not calling those, we need set it here.
```
bvaughn, line 1364: er mah gerd

bvaughn:
This is ready for review, even though it may not be ready to sync to www yet (I don't think master is either for that matter). Going to run a few test syncs today to see what's up.

bvaughn:
Merged via #19396

PR #19377, opened 2020-07-15
Run DevTools tests against npm releases

acdlite:
WIP

Adds a CircleCI job that runs the DevTools tests against a given version of React.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 
 aliases:
   - &docker
@@ -371,6 +371,23 @@ jobs:
       - *run_yarn
       - run: yarn test --project=devtools --build --ci
 
+  yarn_test_release_devtools:
+    docker: *docker
+    environment: *environment
+    parameters:
+      version:
+        type: string
+        default: next
+    steps:
+      - checkout
+      - attach_workspace: *attach_workspace
+      - *restore_yarn_cache
+      - *run_yarn
+      - run:
+          name: Download a specific version of React + renderers
+          command: node scripts/jest/install-devtools-release-test-deps.js << parameters.version >>
+      - run: yarn test --project=devtools --build --ci
+
   RELEASE_CHANNEL_stable_yarn_test_dom_fixtures:
     docker: *docker
     environment: *environment
@@ -515,6 +532,9 @@ workflows:
           requires:
             - yarn_build
       - yarn_test_build_devtools:
+          requires:
+            - yarn_build
+      - yarn_test_release_devtools:
           requires:
             - yarn_build
       # FIXME: Temporarily disabled to unblock master.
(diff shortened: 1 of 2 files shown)


acdlite:
@bvaughn This works but the tests don't actually pass. Some of them are implementation details, though, and I'm not sure how important they are.

Maybe before landing this, I can set up the `@gate` pragma and we can skip the ones that we don't care about? Could also gate using version number.

acdlite:
Or I could land this now and make it so it's not push blocking. I was thinking we'd want to do that anyway so we don't slow down unrelated PRs. We can run it on a cron schedule instead.

bvaughn:
> Could also gate using version number.

Yeah, gating by version number seems really helpful for this.

> Or I could land this now and make it so it's not push blocking. I was thinking we'd want to do that anyway so we don't slow down unrelated PRs. We can run it on a cron schedule instead.

This also sounds fine.

scripts/jest/install-devtools-release-test-deps.js
```diff
@@ -0,0 +1,39 @@
...
+async function main() {
+  await exec('mkdir build-tmp');
+  await exec('npm install --prefix ./build-tmp react-dom@' + version);
+  const modules = fs.readdirSync('build-tmp/node_modules');
+  for (const dep of modules) {
+    if (
+      dep.startsWith('.') ||
+      // Don't replace React isomorphic package, since the backend may use
+      // component types/features that don't exist in an older version
+      dep === 'react'
```
bvaughn, line 28: Will we hit any snags with earlier 16.x react-dom versions and later 16.x react versions?

scripts/jest/install-devtools-release-test-deps.js
```diff
@@ -0,0 +1,39 @@
...
+// by updating React DOM to an older release, it will affect the "backend" (the
+// code that runs in the browser) but not the "frontend" (the code that runs
+// the DevTools React app itself). If we were to use React DOM in the DevTools
+// tests, we'd need to figure out a way to load separate versions for the
+// backend and the front end.
+
+const version = process.argv[2];
+
+async function main() {
+  await exec('mkdir build-tmp');
```
bvaughn, line 20: We should probably have some cleanup for if this script fails midway (it does for me at the moment, and it leaves a build-tmp dir which causes it to fail next run)

.circleci/config.yml
```diff
@@ -371,6 +371,23 @@ jobs:
...
+        default: next
+    steps:
+      - checkout
+      - attach_workspace: *attach_workspace
+      - *restore_yarn_cache
+      - *run_yarn
+      - run:
+          name: Download a specific version of React + renderers
+          command: node scripts/jest/install-devtools-release-test-deps.js << parameters.version >>
+      - run: yarn test --project=devtools --build --ci
```
domiverse, line 389: yarn_test_release_devtools:
    docker: *docker
    environment: *environment
    parameters:
      version:
        type: string
        default: next
    steps:
      - checkout
      - attach_workspace: *attach_workspace
      - *restore_yarn_cache
      - *run_yarn
      - run:
          name: Download a specific version of React + renderers
          command: node scripts/jest/install-devtools-release-test-deps.js << parameters.version >>
      - run: yarn test --project=devtools --build --ci

PR #19381, opened 2020-07-16
Remove firstEffect from Persistent Mode Optimization

sebmarkbage:
__Builds on top of #19322 with one commit e829f133da79294ae766c905d1884e4f419b6992__

Persistent mode needs to clone a parent and add its children if a child has changed.

We have an optimization in persistent mode where we don't do that if no child could've changed. If there are no effects scheduled for any child then there couldn't have been changes.

Instead of checking for this on firstEffect, we now check this on the children's effectTag and subtreeTags.

This is quite unfortunate because if we could just do this check a little bit later we would've already gotten it transferred to the completed work's subtreeTag. Now we have to loop over all the children and if any of them changed, we have to loop over them again. Doing at least two loops per parent.

One possible fix to this would be to move the resetChildLanes to be an explicit call within each completeWork branch before they return. That way we could call it before returning in this case.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.new.js
@@ -65,7 +65,9 @@ import {
   NoEffect,
   DidCapture,
   Snapshot,
+  MutationMask,
 } from './ReactSideEffectTags';
+import {NoEffect as NoSubtreeTag, Mutation} from './ReactSubtreeTags';
 import invariant from 'shared/invariant';
 
 import {
@@ -154,6 +156,25 @@ function markRef(workInProgress: Fiber) {
   workInProgress.effectTag |= Ref;
 }
 
+function hadNoMutationsEffects(current: null | Fiber, completedWork: Fiber) {
+  const didBailout = current !== null && current.child === completedWork.child;
+  if (didBailout) {
+    return true;
+  }
+
+  let child = completedWork.child;
+  while (child !== null) {
+    if ((child.effectTag & MutationMask) !== NoEffect) {
+      return false;
+    }
+    if ((child.subtreeTag & Mutation) !== NoSubtreeTag) {
+      return false;
+    }
+    child = child.sibling;
+  }
+  return true;
+}
+
 let appendAllChildren;
 let updateHostContainer;
 let updateHostComponent;
@@ -198,7 +219,7 @@ if (supportsMutation) {
     }
   };
 
-  updateHostContainer = function(workInProgress: Fiber) {
+  updateHostContainer = function(current: null | Fiber, workInProgress: Fiber) {
     // Noop
   };
   updateHostComponent = function(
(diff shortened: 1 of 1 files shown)


acdlite:
I'm trying to remember the reason why `resetChildLanes` exists. An alternative would be to reset `lanes` (and `subtreeTag`, and similar fields) to 0 in the begin phase, and accumulate the child values in each ones complete phase. Instead of doing another traversal.

We already do something similar for the `Incomplete` effect tag, which essentially has the same properties as a `subtreeTag` since it is set on all parents of an incomplete fiber (something that threw):

https://github.com/facebook/react/blob/392277f0abc2251ea93b7debad4ae86cd0f1bdff/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L1746

I'm thinking maybe the original reason has something to do with resuming? Maybe it's not relevant anymore?

If so, maybe we could get rid of `resetChildLanes` completely and move everything to `completeWork`.

acdlite:
I was hoping to do something like this anyway to get rid of this Offscreen check that's currently sitting smack dab in the middle of the hot path:

https://github.com/facebook/react/blob/fed4ae0247883050aab6d77d9a83f9b6399d9a45/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L1769-L1777

acdlite:
Lol next time I'll read the description:

> One possible fix to this would be to move the resetChildLanes to be an explicit call within each completeWork branch before they return. That way we could call it before returning in this case.

Yeah this makes sense to me. Let's do it.

bvaughn:
> One possible fix to this would be to move the resetChildLanes to be an explicit call within each completeWork branch before they return. That way we could call it before returning in this case.

> If so, maybe we could get rid of resetChildLanes completely and move everything to completeWork.

👍  This sounds like a good change. Who owns it? Want me to pick it up?

sebmarkbage:
Accumulating the effectTag as we go is effectively what the effect list did. However that has some downsides too. It makes rendering something only to throw it out difficult. It also makes things order dependent.

Interestingly I think the new refactor will “fix” the suspense list issue by making effects fire in first-to-last child order instead of the reveal order as was before.

appendAllChildren is another one of those that would be nice to move to happen inside the children and append to the parent but it suffers from the same problem.

Moving it to the complete phase as a second pass should work though. It also allows passes to ignore it if it knows it doesn’t have to like if something is Offscreen. However it also requires us to remember to call it.

acdlite:
> Interestingly I think the new refactor will “fix” the suspense list issue by making effects fire in first-to-last child order instead of the reveal order as was before.

Yeah the key difference between the effect list and `lanes`, `subtreeTag`, et al is that they aren't order dependent. And now that effect list is gone and we operate on the tree, effects aren't order dependent either.

lunaruan:
> Interestingly I think the new refactor will “fix” the suspense list issue by making effects fire in first-to-last child order instead of the reveal order as was before.

What is the suspense list issue you all are referring to?

sebmarkbage:
> What is the suspense list issue you all are referring to?

@lunaruan SuspenseList with `revealOrder="backwards"` caused the effects of siblings to fire backwards. I.e. mount effects in a list of A, B, C would fire mount for C first, then B, then A. That's fine because we don't really guarantee it but that was the only case that happened.

PR #19387, opened 2020-07-16
Track event times per lane on the root

acdlite:
Previous strategy was to store the event time on the update object and accumulate the most recent one during the render phase.

Among other advantages, by tracking them on the root, we can read the event time before the render phase has finished.

I haven't removed the `eventTime` field from the update object yet, because it's still used to compute the timeout. Tracking the timeout on the root is my next step.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberLane.js
+++ b/packages/react-reconciler/src/ReactFiberLane.js
@@ -374,6 +374,25 @@ export function getNextLanes(root: FiberRoot, wipLanes: Lanes): Lanes {
   return nextLanes;
 }
 
+export function getMostRecentEventTime(root: FiberRoot, lanes: Lanes): number {
+  const eventTimes = root.eventTimes;
+
+  let mostRecentEventTime = NoTimestamp;
+  while (lanes > 0) {
+    const index = pickArbitraryLaneIndex(lanes);
+    const lane = 1 << index;
+
+    const eventTime = eventTimes[index];
+    if (eventTime > mostRecentEventTime) {
+      mostRecentEventTime = eventTime;
+    }
+
+    lanes &= ~lane;
+  }
+
+  return mostRecentEventTime;
+}
+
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -723,13 +756,18 @@ export function markRootFinished(root: FiberRoot, remainingLanes: Lanes) {
...
   const expirationTimes = root.expirationTimes;
+
+  // Clear the lanes that no longer have pending work
   let lanes = noLongerPendingLanes;
   while (lanes > 0) {
     const index = pickArbitraryLaneIndex(lanes);
     const lane = 1 << index;
 
-    // Clear the expiration time
+    entanglements[index] = NoTimestamp;
```
acdlite, line 769: This line was an oversight from when I added the entanglement feature. It's theoretically observable but if I write a test it'll be super convoluted, since we currently only use entanglement for `useMutableSource`, and only in a particular de-opt case.

acdlite:
Old fork tests are failing because I forgot to sync changes to ReactFiberRoot.old. Will fix in a bit.

packages/react-reconciler/src/ReactFiberLane.js
```diff
@@ -723,13 +756,18 @@ export function markRootFinished(root: FiberRoot, remainingLanes: Lanes) {
...
   const expirationTimes = root.expirationTimes;
+
+  // Clear the lanes that no longer have pending work
   let lanes = noLongerPendingLanes;
   while (lanes > 0) {
     const index = pickArbitraryLaneIndex(lanes);
     const lane = 1 << index;
 
-    // Clear the expiration time
+    entanglements[index] = NoLanes;
```
acdlite, line 769: This line was an oversight from when I added the entanglement feature. It's theoretically observable but if I write a test it'll be super convoluted, since we currently only use entanglement for useMutableSource, and only in a particular de-opt case.

PR #19396, opened 2020-07-17
Conditionally enable profiler component stacks

bvaughn:
This PR stacks on top of #19376 and includes only one commit, [e6e3d8b](https://github.com/facebook/react/commit/e6e3d8b7a4321ee9a0b98d14173ce89a7b727e91?w=1)

The new component stack approach introduced in #18495 seems to cause bugs (e.g. [1](https://reactjs.org/docs/error-decoder.html?invariant=310), [2](https://reactjs.org/docs/error-decoder.html/?invariant=300)) when used by the new profiler marks (#19376), so this PR moves that part of the profiler behind a feature flag (off by default).

This will enable us to sync the new profiler marks to the new reconciler and let people start using them within Facebook safely, while still enable me to further investigate why the errors are happening. (I've done a www sync of this PR and verified that it fixes the 300 and 310 errors I've been seeing on Facebook.com)

Note that this is a dynamic feature flag, but it's only used once during initialization so it should be okay.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberClassComponent.new.js
+++ b/packages/react-reconciler/src/ReactFiberClassComponent.new.js
@@ -16,6 +16,8 @@ import {Update, Snapshot} from './ReactSideEffectTags';
 import {
   debugRenderPhaseSideEffectsForStrictMode,
   disableLegacyContext,
+  enableDebugTracing,
+  enableSchedulingProfiler,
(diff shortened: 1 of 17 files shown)


bvaughn:
This PR fixes the 300 and 310 errors I've been seeing on Facebook.com (tested with an on-demand just now).

PR #19401, opened 2020-07-18
Add feature flag for setting update lane priority

rickhanlonii:
## Overview

Wrap calls to set update lane priority in a feature flag so it doesn't impact the stable release channel.


rickhanlonii:
Bump

packages/shared/ReactFeatureFlags.js
```diff
@@ -128,3 +128,4 @@ export const deferRenderPhaseUpdateToNextBatch = true;
 
 // Replacement for runWithPriority in React internals.
 export const decoupleUpdatePriorityFromScheduler = false;
+export const enableSetUpdateLanePriority = false;
```
acdlite, line 131: I don't think this needs to be a separate flag. Can use `decoupleUpdatePriorityFromScheduler`.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1250,16 +1260,21 @@ export function flushSync<A, R>(fn: A => R, a: A): R {
...
+      setCurrentUpdateLanePriority(SyncLanePriority);
+    }
     if (fn) {
       return runWithPriority(ImmediateSchedulerPriority, fn.bind(null, a));
     } else {
       return (undefined: $FlowFixMe);
     }
   } finally {
-    setCurrentUpdateLanePriority(previousLanePriority);
+    if (enableSetUpdateLanePriority && previousLanePriority != null) {
```
acdlite, line 1275: Factoring nit: Move the flag check outside the try-catch. Then you don't have to do the weird `previousLanePriority` check.

Something like:

```js
if (fn) {
  if (enableSetUpdateLanePriority) {
    return runWithPriority(ImmediateSchedulerPriority, fn.bind(null, a));
  } else {
    const previousLanePriority = getCurrentUpdateLanePriority();
    setCurrentUpdateLanePriority(SyncLanePriority);
    try {
      return runWithPriority(ImmediateSchedulerPriority, fn.bind(null, a));
    } finally {
      setCurrentUpdateLanePriority(previousLanePriority);
    }
  }
}
```

I find this easier to see the difference in behavior.

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -1271,12 +1286,17 @@ export function flushSync<A, R>(fn: A => R, a: A): R {
 export function flushControlled(fn: () => mixed): void {
   const prevExecutionContext = executionContext;
   executionContext |= BatchedContext;
-  const previousLanePriority = getCurrentUpdateLanePriority();
+  let previousLanePriority;
   try {
-    setCurrentUpdateLanePriority(SyncLanePriority);
+    if (enableSetUpdateLanePriority) {
+      previousLanePriority = getCurrentUpdateLanePriority();
```
acdlite, line 1292: Same nit in all these other places

packages/shared/ReactFeatureFlags.js
rickhanlonii: Agreed, I thought you said we should use a separate flag so that this one could be hard coded to either true or false while the other one is dynamic but maybe I misunderstood?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
rickhanlonii: +1

packages/shared/ReactFeatureFlags.js
acdlite: Either I was wrong or there was a miscommunication :D

One flag is fine

Issue #19402, opened 2020-07-18
Bug: React Hook "useContext" is called in function "login" which is neither a React function component or a custom React Hook function  react-hooks/rules-of-hooks


mfhassan22:
i was creating new react app and was integrating with mobx while coding the following error arise which is wrong with respect to hook rules stated below

![Screen Shot 2020-07-18 at 10 17 08 AM](https://user-images.githubusercontent.com/1241338/87842851-4be9d080-c8e2-11ea-925e-9b86ae595af6.png)

React version:"react": "^16.13.1"

where as the code is correct with respect to given rules
![Screen Shot 2020-07-18 at 10 35 08 AM](https://user-images.githubusercontent.com/1241338/87842869-6fad1680-c8e2-11ea-9c69-1db23e335f7d.png)

[hook rules](https://reactjs.org/warnings/invalid-hook-call-warning.html#breaking-the-rules-of-hooks)
## The current behavior

React Hook "useContext" is called in function "login" which is neither a React function component or a custom React Hook function  react-hooks/rules-of-hooks

## The expected behavior
No error

milesj:
@mfhassan22 Components must be capitalized. It should be `Login` not `login`.

mfhassan22:
thank you and really sorry after four years i first time i realize this

mfhassan22:
my mistake

rickhanlonii:
This still seems like a bug. The error message you got says that "login is not a React function component", but it _is_ a function component, it just isn't valid because of the casing. Let's update this check so that it shows a similar message to the runtime error which is:

> The tag <login> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.

Something like:

> React Hook "useContext" is called in function "login" which is not a valid React component because it does not start its name with an uppercase letter.

rickhanlonii:
[Here's the tests to fix](https://github.com/facebook/react/blob/ddcc69c83b59ef0f895aa5020196e2ae9de36133/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js#L643-L671) and [here's the code to change](https://github.com/facebook/react/blob/30b47103d4354d9187dc0f1fb804855a5208ca9f/packages/eslint-plugin-react-hooks/src/RulesOfHooks.js#L478-L487).

sladyn98:
Would love to submit a pr if no one is working on it

DanielJDupont:
I would like to work on it as my first issue, if that is possible.

[...]

PR #19407, opened 2020-07-19
Fixed error message "is not a React function component" to "not valid because of casing"

DanielJDupont:
## Summary
Fixed an error message that incorrectly stated "login is not a React function component" for something that is a function component. The error message now states that the component isn't valid because of casing.

From issue: https://github.com/facebook/react/issues/19402

## Test Plan
Corrected the relevant test in: packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
@@ -913,12 +913,21 @@ function loopError(hook) {
 }
 
 function functionError(hook, fn) {
-  return {
-    message:
-      `React Hook "${hook}" is called in function "${fn}" that is neither ` +
-      'a React function component nor a custom React Hook function.' +
-      ' React component names must start with an uppercase letter.',
-  };
+  if (hook[0] === hook[0].toLowerCase()) {
+    return {
+      message:
+        `React Hook "${hook}" is called in function "${fn}" which is not ` +
+        'a valid React component because it does not start its name with ' +
+        'an upper case letter.',
+    };
+  } else {
+    return {
+      message:
+        `React Hook "${hook}" is called in function "${fn}" that is neither ` +
+        'a React function component nor a custom React Hook function.' +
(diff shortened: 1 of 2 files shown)


DanielJDupont:
Whoops, I will make the corrections.

packages/eslint-plugin-react-hooks/__tests__/ESLintRulesOfHooks-test.js
```diff
@@ -913,12 +913,21 @@ function loopError(hook) {
 }
 
 function functionError(hook, fn) {
-  return {
-    message:
-      `React Hook "${hook}" is called in function "${fn}" that is neither ` +
-      'a React function component nor a custom React Hook function.' +
-      ' React component names must start with an uppercase letter.',
-  };
+  if (hook[0] === hook[0].toLowerCase()) {
```
rickhanlonii, line 916: Since this is a helper function, we should split it out into a different function so we have both:

- `functionError`
- `functionInvalidError`

That way we can assert exactly which one we expect.

DanielJDupont: Ah, thank you, I have just been experimenting with getting a build of react with my changes working, I will change that.

gaearon:
It seems like https://github.com/facebook/react/issues/19402 is a misunderstanding. The suffix explaining the reason for the error has already been added in https://github.com/facebook/react/pull/18881 but it doesn't appear yet because CRA is using an old version of the plugin.

We'd be happy to take further improvements to the message but I'm struggling to follow what this PR is doing. What are the "other cases" that the check is supposed to catch? I'm not sure how an algorithm can possibly distinguish something like `button` (invalid because it should've been `Button`) from `formatText` (invalid because it's not a component). If we could somehow distinguish what is a component, we wouldn't need the capital name heuristic in the first place.

I might be missing something!

DanielJDupont:
Thank you for your time and response, I believe that clarifies everything for myself. I feel happy with the error message that will be in the next release and I don't feel any need for this pull request.

PR #19408, opened 2020-07-20
Fix for react-tester-renderer swallowing error boundary logs

rickhanlonii:
## Overview

Previously, when using react-test-renderer, errors caught by error boundaries would not log the original error message, it would _only_ print the addendum "The above error occurred in the <ErrorThrowingComponent> component".

This PR fixes that behavior so that the test render logs errors even when error boundaries catch them, just like ReactDOM.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberErrorLogger.js
+++ b/packages/react-reconciler/src/ReactFiberErrorLogger.js
@@ -32,6 +32,16 @@ export function logCapturedError(
       const source = errorInfo.source;
       const stack = errorInfo.stack;
       const componentStack = stack !== null ? stack : '';
+
+      if (error != null && boundary.type != null && !error._loggedByBrowser) {
+        // If an error boundary caught this error then we need to log it.
+        // However, we can't just always log it because the browser's
+        // implementation of invokeGuardedCallback will log the error and we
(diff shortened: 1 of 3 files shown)


rickhanlonii:
@gaearon @acdlite could you take a look at the proposed fix here?

It breaks 100 tests (legitimately, I think). For example, [this test](https://github.com/facebook/react/blob/30b47103d4354d9187dc0f1fb804855a5208ca9f/packages/react-reconciler/src/__tests__/ReactIncrementalErrorHandling-test.internal.js#L1509-L1512) asserts that only the "the above error" addendum is logged (not the error itself) and [this test](https://github.com/facebook/react/blob/8f05f2bd6d131a39835d468622e248b231ccbf8e/packages/react-reconciler/src/__tests__/ReactLazy-test.internal.js#L92-L120) makes no log assertions even though it intentionally throws in render.

Want to confirm the approach before I fix all 94 tests 🥵

rickhanlonii:
@acdlite @gaearon if you want a good place to start, check out the test that I changed.

You can see that in dev, console.log is only called once. It does not include the actual error that was thrown, only the "the above error" addendum. After the fix, we log both the error and the addendum.

Issue #19362, opened 2020-07-15
Scheduler callback is only checked for null value and therefore throws when callback is undefined.


RobiFerentz:
https://github.com/facebook/react/blob/30b47103d4354d9187dc0f1fb804855a5208ca9f/packages/scheduler/src/Scheduler.js#L180-L185

JohnMHoyt24:
I noticed in line 179, of Scheduler.js, "const callback = currentTask.callback;

Should the if statement be testing if (callback !== currentTask.callback)?

const callback = currentTask.callback;
    **if (callback !== currentTask.callback)** {
      currentTask.callback = null;
      currentPriorityLevel = currentTask.priorityLevel;
      const didUserCallbackTimeout = currentTask.expirationTime <= currentTime;
      markTaskRun(currentTask, currentTime);
      const continuationCallback = callback(didUserCallbackTimeout);

aweary:
@RobiFerentz is there a specific issue or bug you're reporting here? I believe the strict comparison is intentional as `callback` will only ever be `null` or a function.

RobiFerentz:
Well, the I use the payment-modals package. It looks like it's using react just fine. So I went deeper and found the console error i get originates from those lines.

Whatever the reason is, it would be a better idea to check for function type, no?

rickhanlonii:
@RobiFerentz confirmed, we'll throw if you do `scheduleCallback(ImmediatePriority, undefined)`, nice catch;

The test we'll want to add to `Scheduler-test.js` is:

```
  it('gracefully handles scheduled tasks that are null or undefined', () => {
    scheduleCallback(ImmediatePriority, null);

    expect(Scheduler).toFlushWithoutYielding();

    scheduleCallback(ImmediatePriority, undefined);

    expect(Scheduler).toFlushWithoutYielding();
  });
```

We should probably throw if you try to schedule a nullish task but that's a breaking change.

rickhanlonii:
@RobiFerentz do you want to submit a PR to fix?

nsantos16:
If @RobiFerentz don't have a PR to fix, I'll like to take this issue 😄

RobiFerentz:
Sure, I'll create a PR

rickhanlonii:
PR has been merged, thanks @RobiFerentz!

PR #19412, opened 2020-07-20
fix: issue #19362

RobiFerentz:
## Summary

Should fix issue #19362

### DIFF
--- a/packages/scheduler/src/Scheduler.js
+++ b/packages/scheduler/src/Scheduler.js
@@ -177,7 +177,7 @@ function workLoop(hasTimeRemaining, initialTime) {
       break;
     }
     const callback = currentTask.callback;
-    if (callback !== null) {
+    if (typeof callback === 'function') {
       currentTask.callback = null;
       currentPriorityLevel = currentTask.priorityLevel;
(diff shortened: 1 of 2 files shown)


packages/scheduler/src/Scheduler.js
```diff
@@ -177,7 +177,7 @@ function workLoop(hasTimeRemaining, initialTime) {
       break;
     }
     const callback = currentTask.callback;
-    if (callback !== null) {
+    if (typeof callback === 'function') {
```
rickhanlonii, line 180: I think for now we should change this to:

```suggestion
    if (callback != null) {
```

More complex checks should be done when the task is scheduled, as @gaearon said.

gaearon: The `typeof` check might counterintuitively be faster because it doesn't need to check for the weird `document.all` case that `!=` takes into account. In the past `!=` was known to deopt things. Not sure it's still the case but I think the `typeof` check is fine.

rickhanlonii: Oh, nice!

rickhanlonii:
Thanks @RobiFerentz for your first React PR!

PR #19420, opened 2020-07-20
isFBBundle should be true if it's FB for www or FB for RN

sebmarkbage:
This ensures that the .fb.js override files kick in for RN. Otherwise we won't have FB specific exports in the isomorphic modules.

### DIFF
--- a/scripts/rollup/build.js
+++ b/scripts/rollup/build.js
@@ -346,7 +346,7 @@ function getPlugins(
     bundleType === UMD_DEV ||
     bundleType === UMD_PROD ||
     bundleType === UMD_PROFILING;
-  const isFBBundle =
+  const isFBWWWBundle =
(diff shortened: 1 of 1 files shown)


scripts/rollup/build.js
```diff
@@ -357,7 +357,7 @@ function getPlugins(
     bundleType === RN_FB_DEV ||
     bundleType === RN_FB_PROD ||
     bundleType === RN_FB_PROFILING;
-  const shouldStayReadable = isFBBundle || isRNBundle || forcePrettyOutput;
+  const shouldStayReadable = isFBWWWBundle || isRNBundle || forcePrettyOutput;
```
sebmarkbage, line 360: Arguably `shouldStayReadable` should be false for non-FB since we don't trust the minifier.

PR #19425, opened 2020-07-21
Added class component example

ZachyDev:
I have used a class component that achieves the same result as the one given ( functonal component).

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,11 +48,23 @@ ReactDOM.render(
   <HelloMessage name="Taylor" />,
   document.getElementById('container')
 );
+class HelloMessage extends React.Component {  
+  const {name} = this.props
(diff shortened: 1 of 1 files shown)


ZachyDev:
I have added the class component example ,to show how both components work

bvaughn:
Hey thanks for the README suggestion! 🙇 

I think this change makes the first introductory example a little less easy to follow though, because it's not clear why we're showing two components and two render calls until a few paragraphs later. I also think we'll be de-emphasizing class components in our documentation more going forward, given that function components provide ergonomic and performance benefits.

Dan and Rachel are currently working on a docs rewrite with a function+hooks first focus. Maybe we can revisit this README in a couple of months once that project is done and reconsider!

PR #19428, opened 2020-07-21
Regression test for media event bubbling

gaearon:
This test passes on trunk and in https://github.com/facebook/react/pull/19278, but fails before https://github.com/facebook/react/pull/19278 included the last two commits. This reproduces the internal regression caught in the failing E2E tests.


trueadm:
Nice one :)

PR #19433, opened 2020-07-22
Move event things around

gaearon:
No substantive changes. This just moves stuff:

* I put all SyntheticEvents in one file. So you don't have to jump around that much and all the metaprogramming happens in one place. This makes it easier to undo the metaprogramming in the future.
* I inlined a few helpers that were previously in separate files. Now they don't have to be.
* I removed `Modern` from all filenames since there's only one event system now.

The file line reduction is because there's fewer imports and license blocks (cause fewer files).


gaearon:
The "size increase" is from TestUtils, not actual ReactDOM.

PR #19435, opened 2020-07-22
Remove SyntheticEvent metaprogramming

gaearon:
This probably increases the size but worth a try. If it increases I can probably code-golf it a bit without bringing back the `extends` stuff.


gaearon:
Haha size changes misleading again. But yes it goes up a little bit.

gaearon:
Alternatively maybe instead we should remove the constructors and "embrace the config". So there's just one constructor.

gaearon:
I prefer https://github.com/facebook/react/pull/19436.

PR #19436, opened 2020-07-22
Remove SyntheticEvent subtypes

gaearon:
Alternative to https://github.com/facebook/react/pull/19435. These are configs, subtyping wasn't really useful.


gaearon:
As usual, misleading sizes but DOM did decrease a tiny bit.

PR #19439, opened 2020-07-23
Add invalid/close events to <dialog> element

trueadm:
We can support propagation for dialog events by registering them to all `<dialog>` elements we encounter (like we do with other events). Added regression test too. This leave `onScroll` being the only event we don't handle this way.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMEventListener-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMEventListener-test.js
@@ -502,6 +502,39 @@ describe('ReactDOMEventListener', () => {
     }
   });
 
+  it('should delegate dialog events even without a direct listener', () => {
+    const container = document.createElement('div');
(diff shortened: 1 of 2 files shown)


tjallingt:
This isn't exactly the right place to ask this question but I've been wondering since you where implementing the event capture system: why add this special behavior for certain events? Is the non-bubbling behavior considered a bug?

I understand the desire to not have needless breaking changes but if that is the only reason to keep this behavior around are you considering deprecating it in the future?

trueadm:
@tjallingt This is a good question. Ideally, we want to try and make things non-breaking changes where possible, otherwise it can affect the amount of work required to upgrade between versions of React. However, there are cases where we might feel this is valuable, or is of more benefit to change something vs the required work needed to upgrade. Given that these events only require a few extra lines to provide backwards support, then it probably makes sense to keep them for now.

PR #19440, opened 2020-07-23
Effects list refactor: Move resetChildLanes into complete work

bvaughn:
Builds on top of #19374

- [x] Move `resetChildLanes` out of work loop and into complete work (named `bubbleProperties`)
- [x] Move tag-type checks in `bubbleProperties` into their corresponding switch/case statements.
- [ ] Change new `bubbleProperties` method to bubble up to the parent as each Fiber completes, rather than iterating over its children. (This will enable us to remove a potentially expensive while loop.)

### DIFF
--- a/packages/react-reconciler/src/ReactFiberCompleteWork.new.js
+++ b/packages/react-reconciler/src/ReactFiberCompleteWork.new.js
@@ -71,6 +71,7 @@ import {
   DidCapture,
   Snapshot,
   MutationMask,
+  PerformedWork,
   StaticMask,
 } from './ReactFiberFlags';
 import invariant from 'shared/invariant';
@@ -155,7 +156,10 @@ import {
 } from './ReactFiberLane';
 import {resetChildFibers} from './ReactChildFiber.new';
 import {createScopeInstance} from './ReactFiberScope.new';
-import {transferActualDuration} from './ReactProfilerTimer.new';
+import {
+  stopProfilerTimerIfRunning,
+  stopProfilerTimerIfRunningAndRecordDelta,
+} from './ReactProfilerTimer.new';
 
 function markUpdate(workInProgress: Fiber) {
   // Tag the fiber with an update effect. This turns a Placement into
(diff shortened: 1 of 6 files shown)


packages/react-reconciler/src/ReactFiberCompleteWork.new.js
```diff
@@ -649,6 +671,124 @@ function cutOffTailIfNeeded(
   }
 }
 
+function bubbleProperties(completedWork: Fiber): void {
```
bvaughn, line 674: This isn't the greatest name, but neither was `resetChildLanes`. Open to suggestions.

bvaughn:
Based on @acdlite's comment on PR #19381...
> I'm trying to remember the reason why resetChildLanes exists. An alternative would be to reset the childLanes (and subtreeTag, and similar fields) to 0 in the begin phase, and accumulate the child values in each child fiber's complete phase. Instead of doing another traversal.

...it seems there may have been an additional refactoring goal here to bubble properties to a completed Fiber's parent, so we could remove the child loop that currently exists. I think this is doable, but I think it would also require us to reset `treeBaseDuration` on a Fiber during the begin phase (when we clone) in order for the children to be able to bubble their times up correctly. Maybe that would work and maybe it would cause a problem. I'm actually not sure, but I can take a shot at it and see.

We also still loop over children to bubble `actualDurations` in the work loop, in the event of an error, but I assume that could stay as-is.
https://github.com/facebook/react/blob/909b612e27d1a59aa601e2b8461a44661faaffa8/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L1795-L1803

bvaughn:
Similar to the base duration issue mentioned above, properly bubbling `childLanes` from the child to the parent seems trickier than the other way around. (More specifically knowing when to *reset* the value is more difficult.)

acdlite:
>  Regarding the treeBaseDuration issue mentioned above, I think one plausible approach might be to reset treeBaseDuration to 0 during clone, but then copy it back in the event of a bailout.

That makes sense to me.

> However I'm running into another more pressing snag- properly bubbling childLanes from the child to the parent seems trickier than the other way around. (More specifically knowing when to reset the value is more difficult.)

My guess is that the issue you ran into is when a fiber bails out and you reuse the current child (i.e. returning null from `beginWork` instead of continuing on the child). The children won't be rendered, so they won't have a complete phase, so the values from the children won't be accumulated. So in that case, you still need to call something like `resetChildLanes`/`bubbleProperties`.

acdlite:
> So in that case, you still need to call something like resetChildLanes/bubbleProperties.

Actually, I forgot that in the case of a complete bailout (workInProgress.child ===  current.child) the bubbled values are the same as current. So you can read off the current fiber, no need to traverse children.

bvaughn:
Right. This is what I was thinking too, and what I tried, but it didn't work. Probably missed something though. I'll try again after though, I've made the static change we're discussing on the other PR.

bvaughn:
Superseded by #19801

PR #19441, opened 2020-07-23
generate react components from command line

Mridul11:


rickhanlonii:
Hey @Mridul11, thanks for the PR. Not sure what the intent is, but a generator doesn't seem like something we're going to add to core.

PR #19458, opened 2020-07-26
Remove callbackId field from FiberRoot

acdlite:
The old expiration times implementation used this field to infer when the priority of a task had changed at a more granular level than a Scheduler priority level.

Now that we have the LanePriority type, which is React-specific, we no longer need the `callbackId` field.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -677,10 +677,10 @@ function markUpdateLaneFromFiberToRoot(
 }
 
 // Use this function to schedule a task for a root. There's only one task per
-// root; if a task was already scheduled, we'll check to make sure the
-// expiration time of the existing task is the same as the expiration time of
-// the next level that the root has work on. This function is called on every
-// update, and right before exiting a task.
+// root; if a task was already scheduled, we'll check to make sure the priority
+// of the existing task is the same as the priority of the next level that the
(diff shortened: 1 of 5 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -689,37 +689,32 @@ function ensureRootIsScheduled(root: FiberRoot, currentTime: number) {
   markStarvedLanesAsExpired(root, currentTime);
 
   // Determine the next lanes to work on, and their priority.
-  const newCallbackId = getNextLanes(
+  const nextLanes = getNextLanes(
```
rickhanlonii, line 692: Looks like the comment on line 681 (which references the expiration time) is out of date, should we update that as well?

packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -741,7 +736,6 @@ function ensureRootIsScheduled(root: FiberRoot, currentTime: number) {
     );
   }
 
-  root.callbackId = newCallbackId;
```
rickhanlonii, line 744: We used to update the root `callbackId` with the next lane, why don't we need to do anything with the next lane in the new version?

acdlite: It was only used by this function. Since we don't need it for the logic above anymore, we don't have to store it.

Issue #18051, opened 2020-02-17
Bug: [eslint-plugin-react-hooks] exhaustive-deps false positive on "unnecessary" dependency if its a React component


zeorin:
## Steps to reproduce

1. create a memoized value using `useMemo`
2. a React component is used in the creation of this value, in a JSX expression
3. specify the React component in the dependency array

Link to code example: https://github.com/zeorin/eslint-plugin-react-hooks-repro

## The current behavior

```
React Hook useMemo has an unnecessary dependency: 'Component'. Either exclude it or remove the dependency array react-hooks/exhaustive-deps
```

## The expected behavior

No lint errors.

## More details

A simple repro (taken from the link above) is:

```javascript.jsx
function Foo({ component: Component }) {
	const memoized = useMemo(() => ({
		render: () => <Component />
	}), [Component]);

	return memoized.render();
}
```

## Workarounds

If one changes the component to lowercase, the lint error goes away. It does also mean that we need to change the way we render the component:

```javascript.jsx
function Foo({ component }) {
	const memoized = useMemo(() => ({
		render: component
	}), [component]);

	return memoized.render();
}
```

Alternatively we can decide not to use JSX, in which case the lint rule functions correctly, too:

```javascript.jsx
function Foo({ component: Component }) {
	const memoized = useMemo(() => ({
		render: () => React.createElement(Component)
	}), [Component]);

	return memoized.render();
}
```

## Impact

Currently it is hard to use props that are components in a JSX expression if one is using the `exhaustive-deps` rule.

This is also compounded by the fact that this rule has a ESLint fix that removes the dependency, thus changing the behaviour of the code and leading to bugs. See https://github.com/facebook/react/issues/16313 for that bug report.

gaearon:
This looks like a legit bug to me. Thank you for reporting!
If you're interested in looking into a fix, let me know and I can give a few pointers.

zeorin:
I am interested in looking into a fix, I had a look but I think there's something implicit that I'm missing. I'd appreciate a few pointers.

simbathesailor:
I tried finding out the location where , it can be solved. So basically 

https://github.com/facebook/react/blob/be76966f6b23e158f3c163061dcc266a294fddc1/packages/eslint-plugin-react-hooks/src/ExhaustiveDeps.js#L360
 
gatherDependenciesRecursively right now is now not able to gather JSXElement as the dependency.

[...]

PR #19462, opened 2020-07-27
react-hooks-eslint-rule: add component props to pureScopes (#18051), …

delca85:
…update yarn.lock

## Summary

Following #18051 [comment](https://github.com/facebook/react/issues/18051#issuecomment-600856619) by an unexpected behavior is found related code like this:
```js
function Example({ sortBy }) {
  const foo = () => useMemo(() => console.log(sortBy), [sortBy])
 }
```
An error message is thrown since `sortBy` is evaluated as an unnecessary dependency while it is a necessary dependency.

According to me this was happening because `pureScopes` array did not contain the scope where `sortBy` is defined. I think this is a bug because component props (as well as argument received by a custom hook) should be contained by `pureScopes` array.
I changed the code in order to make these scopes always contained by `pureScopes` array.

## Test Plan

I have tested the desired behavior is obtained and even that  Component `props` are defined as necessary only if they are used by the hook.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -1376,6 +1376,27 @@ const tests = {
         }
       `,
     },
+    {
+      code: normalizeIndent`
+        function Example({ sortBy }) {
+          const foo = () => useMemo(() => console.log(sortBy), [sortBy])
+        }
+      `,
+    },
+    {
+      code: normalizeIndent`
+        function Example(props) {
+          const usefoo = () => useMemo(() => console.log(props.sortBy), [props.sortBy])
+        }
+      `,
+    },
+    {
+      code: normalizeIndent`
+        function useMyThing(myRef) {
+          const foo = () => useCallback(() => console.log(myRef), [myRef]);
+        }
+      `,
+    },
   ],
   invalid: [
     {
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
```diff
@@ -1376,6 +1376,27 @@ const tests = {
         }
       `,
     },
+    {
+      code: normalizeIndent`
+        function Example({ sortBy }) {
+          const foo = () => useMemo(() => console.log(sortBy), [sortBy])
```
gaearon, line 1382: The thing that's odd about these examples is that they break the Rules of Hooks. Concretely, here we have a Hook call inside a callback. But Hook calls should only happen at the top level.

delca85: Yes, it's totally right. So there is no bug but only a wrong hook usage.

delca85: Given that, I think I could close this PR because it eventually fixes a situation caused by a wrong Hook usage.
Since Hook calls should only happen at the top level, component props will be alway in `pureScopes` array.

delca85: @gaearon sorry for coming back to this argument, I have just one question.
Is the use I made in [this](https://codesandbox.io/s/nervous-haze-0pwlf?file=/src/App.jsx) sandbox breaking the hook rule? Because I don't receive any message from `rules-of-hooks` and I am thinking it could be a legit use (even in a really contrived example) because the hook is called twice a render.

Sorry again if I am missing something.

delca85: Totally misunderstood.
These examples are breaking the Rules of Hooks because the hook is called inside a function that is not a custom hook and neither in a JSX component.
The reason why the example in CodeSandbox is working is that the function within the hook is called begins with `use`. So...double error.

Sorry for the waste of time.

PR #19463, opened 2020-07-27
Delgated all capture events

trueadm:
When we encounter an event that should be applied in the capture phase, we should always delegate the event, even if it doesn't get delegated in the bubble phase.

### DIFF
--- a/packages/react-dom/src/events/plugins/SimpleEventPlugin.js
+++ b/packages/react-dom/src/events/plugins/SimpleEventPlugin.js
@@ -40,10 +40,9 @@ import {
 import {IS_EVENT_HANDLE_NON_MANAGED_NODE} from '../EventSystemFlags';
 
 import getEventCharCode from '../getEventCharCode';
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/events/plugins/SimpleEventPlugin.js
```diff
@@ -166,22 +165,11 @@ function extractEvents(
...
-    // If we are not handling accumulateTargetOnly, then we should traverse
-    // through all React fiber tree, finding all relevant useEvent and
-    // on* prop events as we traverse the tree. Otherwise, we should
-    // only handle the target fiber and stop traversal straight after.
-    if (accumulateTargetOnly) {
-      targetInst = getClosestInstanceFromNode(((targetContainer: any): Node));
-    }
+    // In which case we will want to make this flag boolean and ensure
+    // we change the targetInst to be of the container instance. Like:
+    const accumulateTargetOnly = false;
```
gaearon, line 172: Let's remove this until we use it?

gaearon: I guess I'm fine keeping it too.

Issue #19156, opened 2020-06-18
Bug: Strange onScroll behaviour when child editing


vladislav-andreevich:
React version: 16.13.1
OS: Mac OS
Browser: Chrome 83.0.4103.97

## Steps To Reproduce

1. Create a div and add onScroll event to it
2. Create input inside the created div
3. Start typing in the input box.

## The current behavior

Event onScroll is fired on the parent.

## The expected behavior

Scroll event should not fire.

## Link to code example:

https://codepen.io/vladislav-andreevich/pen/zYroQjN

AhmaudArbery:
I haven't worked on React yet, I am still looking, but I did some digging. It seems like this is the intended behavior of React: https://github.com/facebook/react/issues/5622. I'd wait for someone else to verify.

guy490:
I have worked on React, but I think its a bug because "onScroll" should happen when we scroll.
 in this situation, we are not scrolling something, we only enter text.
so I think its an issue.

JoaoVitoGomes:
> It is correct that the event should be caught on the parent, but in this issue, there is no event to catch because there is no scrolling happen.
The input element fires a scroll event.

guy490:
> 
> The input element fires a scroll event. I don't think this is a bug.

Yes, but as I see in Vanilla JS code it does not fire a scroll event, and this "feature" throws :)

illuminist:
It's weird that one of my machine still run on older version of chromium(80.0.3987.163) and scroll event doesn't fire on react component. But another machine with latest chromium(83.0.4103.61) fires the event.

gaearon:
React 17 doesn't bubble scroll events so this would be fixed.
With React 16 or lower, you can add `if (e.target === e.currentTarget)` condition to your handler.

Issue #15723, opened 2019-05-23
Strange onScroll behaviour


O4epegb:
**Do you want to request a *feature* or report a *bug*?**

Possible bug

**What is the current behavior?**

onScroll callback on parent element fires when children element is scrolled.
Native listener working as expected, thought.

Example with reproduction
https://codesandbox.io/s/kk3th

Just try to scroll little box with items.

**What is the expected behavior?**

I am not sure if this behaviour is correct, but it was unexpected for me, so it might be a bug.
I was not expecting onScroll to fire at all.

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**

"react": "16.8.6"
macOs Mojave 10.14.5

Did not tried any other versions

kunukn:
The syntax for adding event listener is this

`something.addEventListener(event, function, useCapture)`

The native event don't fire for the code you provide

```
this.wrapperNode.current.addEventListener('scroll', () =>
      console.log('native scroll')
    );
```

But it will for this

```
this.wrapperNode.current.addEventListener('scroll', () =>
      console.log('native scroll'), true
    );
```

I suspect behind the scenes, the React onScroll event does the addEventListener with useCapture set to true

O4epegb:
Sorry, you probably misunderstood me!

> The native event don't fire for the code you provide

This is actually intended, this is correct behaviour I am expecting. It should not fire because there is no capturing.

> React onScroll event does the addEventListener with useCapture set to true

So, if this is true then it is very unexpected! Why would it do this?

kunukn:
Regarding my comment that I suspected the useCapture was set to true by React, I was speculating. 

The React scroll event was not added to the wrapper or child element.

From the docs, it says we can read the nativeEvent https://reactjs.org/docs/events.html
I applied `console.log(e.nativeEvent.currentTarget);` I got `#document`

When I inspect what event listeners there are attached to the `document` from Chrome devtools in the console `getEventListeners(document)`

I got this

```
{scroll: Array(1)}
scroll: Array(1)
0: {listener: ƒ, useCapture: true, passive: false, once: false, type: "scroll"}
length: 1
```


Maybe somebody else could give some more insight of how the `scroll` events works in a React?
I probably need to look more into the source code before I can provide more accurate answers.

O4epegb:
So you were right it actually use capturing!

But for me it seems like a bug, or at least very unintuitive behavior. Isn't it?

kunukn:
[...]

PR #19464, opened 2020-07-27
Don't emulate bubbling of the scroll event

gaearon:
Currently, scroll event is somewhat broken on master.

This will fire both handlers if we scroll the child:

```js
<div onScroll={onParentScroll}>
  <div onScroll={onChildScroll} />
</div>
```

But this **will not** fire the parent handler if we scroll the child:

```js
<div onScroll={onParentScroll}>
  <div />
</div>
```

In other words, bubbling the scroll event currently doesn't work if the event target itself doesn't have a React listener.

This is a regression in the modern system.

## Possible Fixes

This is based on @trueadm's writeup.

* **Option 1:** Add native `scroll` handler to every `div` so that we can't "miss" the event — whether it has a React handler or not. This is a similar strategy we use for other events that don't bubble in the browser (e.g. `toggle`, `invalid`, `play`, etc). However, for those cases, we know *which* elements need it (e.g. `audio` and `video`). Adding a `scroll` listener to every single `div` seems untenable. So this is ruled out for performance reasons.

* **Option 2 (Breaking Change):** Stop emulating bubbling for `onScroll`. It it confusing anyway (https://github.com/facebook/react/issues/19156, https://github.com/facebook/react/issues/15723). **This is what this PR does.** The downside is that now this event is special and different from all others.

* **Option 3 (Even Bigger Breaking Change):** Stop emulating bubbling for all events that don't bubble in the browser. This includes `invalid`, `load`, `toggle`, `close`, `cancel`, and media events. We could consider this but it's a bigger breaking change. Arguably, some of them, like `onInvalid`, are actually useful and not too confusing to bubble. So I think this needs to wait until 18.

* **Option 4 (Status Quo):** We could go back to React 16 behavior, which attaches `onScroll` in the capture phase. That would solve the current inconsistency, but wouldn't be in line with our strategy for how nested Reacts should work. It also wouldn't solve the problem that bubbling `onScroll` is inherently confusing.

### DIFF
--- a/packages/react-dom/src/events/plugins/SimpleEventPlugin.js
+++ b/packages/react-dom/src/events/plugins/SimpleEventPlugin.js
@@ -165,13 +165,17 @@ function extractEvents(
       inCapturePhase,
     );
   } else {
-    // TODO: We may also want to re-use the accumulateTargetOnly flag to
-    // special case bubbling for onScroll/media events at a later point.
-    // In which case we will want to make this flag boolean and ensure
-    // we change the targetInst to be of the container instance. Like:
-    const accumulateTargetOnly = false;
+    // Some events don't bubble in the browser.
+    // In the past, React has always bubbled them, but this can be surprising.
+    // We're going to try aligning closer to the browser behavior by not bubbling
+    // them in React either. We'll start by not bubbling onScroll, and then expand.
+    const accumulateTargetOnly =
+      !inCapturePhase &&
+      // TODO: ideally, we'd eventually add all events from
+      // nonDelegatedEvents list in DOMPluginEventSystem.
+      // Then we can remove this special list.
+      topLevelType === DOMTopLevelEventTypes.TOP_SCROLL;
 
-    // We traverse only capture or bubble phase listeners
     accumulateSinglePhaseListeners(
       targetInst,
       dispatchQueue,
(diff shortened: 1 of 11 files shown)


sebmarkbage:
What does onScrollCapture do?

gaearon:
Not sure what you mean. It does regular capturing? That mirrors the browser behavior AFAIK.

gaearon:
Meaning, it would currently capture at the root and call React capture listeners.

packages/react-dom/src/events/plugins/SimpleEventPlugin.js
```diff
@@ -165,13 +165,17 @@ function extractEvents(
...
-    // we change the targetInst to be of the container instance. Like:
-    const accumulateTargetOnly = false;
+    // Some events don't bubble in the browser.
+    // In the past, React has always bubbled them, but this can be surprising.
+    // We're going to try aligning closer to the browser behavior by not bubbling
+    // them in React either. We'll start by not bubbling onScroll, and then expand.
+    const accumulateTargetOnly =
+      !inCapturePhase &&
+      // TODO: ideally, we'd eventually add all events from
+      // nonDelegatedEvents list in ReactDOMEventListener.
```
trueadm, line 175: The list is in `DOMPluginEventSystem`, not `ReactDOMEventListener`.

gaearon:
I'm thinking this is worth putting behind a flag. Kind of risky.

gaearon:
I'm gonna land this behind a flag so we can quickly roll back if something's broken.

PR #19466, opened 2020-07-27
Bail-out of attaching non-delegated listeners

trueadm:
This PR applys a bail-out case for non-delegated event listeners (other than `scroll`), so that we avoid attaching event listeners to elements that have the event, but won't actually ever recieve the native event from the browser. This is how React 16 dealt with events added to target elements only.


gaearon:
Can we add a test?

trueadm:
@gaearon Done :)

PR #19469, opened 2020-07-27
Don't rely on Scheduler.didTimeout to tell us that SyncBatched is synchronous

acdlite:
Tasks with SyncBatchedPriority — used by Blocking Mode — should always be rendered by the `performSyncWorkOnRoot` path, not `performConcurrentWorkOnRoot`.

Currently, they go through the `performConcurrentWorkOnRoot` callback. Then, we check `didTimeout` to see if the task expired. Since SyncBatchedPriority translates to ImmediatePriority in the Scheduler, `didTimeout` is always `true`, so we mark it as expired. Then it exits and re-enters in the `performSyncWorkOnRoot` path.

Aside from being overly convoluted, we shouldn't rely on Scheduler to tell us that SyncBatchedPriority work is synchronous. We should handle that ourselves.

This will allow us to remove the `didTimeout` check. And it further decouples us from the Scheduler priority, so we can eventually remove that, too.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -144,6 +144,7 @@ import {
 import {
   NoLanePriority,
   SyncLanePriority,
+  SyncBatchedLanePriority,
   InputDiscreteLanePriority,
   TransitionShortLanePriority,
   TransitionLongLanePriority,
@@ -726,6 +727,11 @@ function ensureRootIsScheduled(root: FiberRoot, currentTime: number) {
     newCallbackNode = scheduleSyncCallback(
       performSyncWorkOnRoot.bind(null, root),
     );
+  } else if (newCallbackPriority === SyncBatchedLanePriority) {
+    newCallbackNode = scheduleCallback(
+      ImmediateSchedulerPriority,
+      performSyncWorkOnRoot.bind(null, root),
+    );
   } else {
     const schedulerPriorityLevel = lanePriorityToSchedulerPriority(
       newCallbackPriority,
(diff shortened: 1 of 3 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -726,6 +727,11 @@ function ensureRootIsScheduled(root: FiberRoot, currentTime: number) {
     newCallbackNode = scheduleSyncCallback(
       performSyncWorkOnRoot.bind(null, root),
     );
+  } else if (newCallbackPriority === SyncBatchedLanePriority) {
+    newCallbackNode = scheduleCallback(
```
rickhanlonii, line 731: Are we intentionally not using `scheduleSyncCallback` like in the block above?

What if other sync work is already scheduled, do we want this to always come after?

acdlite: The naming is confusing. `scheduleSyncCallback` adds stuff to a special queue that can be flushed synchronously without going through Scheduler. It's used for SyncLane and expired updates.

"sync" in `performSyncWorkOnRoot` means that it doesn't yield to the main thread and it doesn't check Suspense timeouts (though that part might change), so it has fewer checks than concurrent updates.

SyncBatched is kind of a mix  between sync mode and concurrent mode. It doesn't yield to the main thread, but it does go through Scheduler (i.e. is scheduled by postMessage), hence the weird naming conflicts.

rickhanlonii: I see, thanks for explaining 👍

PR #19470, opened 2020-07-27
Add SchedulerHostConfig fork for post task

rickhanlonii:
## Overview

This PR adds a new fork for the Scheduler so we can add the postTask implementation.

### DIFF
--- a/packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
@@ -1,0 +1,250 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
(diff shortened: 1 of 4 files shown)


packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
```diff
@@ -0,0 +1,250 @@
+/**
```
rickhanlonii, line 1: Just copied this from the default for the initial fork

scripts/rollup/bundles.js
```diff
@@ -634,6 +634,14 @@ const bundles = [
     externals: [],
   },
 
+  {
+    bundleTypes: [FB_WWW_DEV, FB_WWW_PROD, FB_WWW_PROFILING],
```
rickhanlonii, line 638: We only need these in the www builds.

PR #19471, opened 2020-07-28
Fix flushing suspense fallbacks at the end of Act when the scheduler is mocked

rickhanlonii:
## Overview

WIP - I think this treats the symptom and not the cause, but shows the issue, what works, and what tests break.

### DIFF
--- a/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
+++ b/packages/react-reconciler/src/ReactFiberWorkLoop.new.js
@@ -3717,7 +3717,7 @@ function finishPendingInteractions(root, committedLanes) {
 // access to the same internals that we do here. Some trade offs in the
 // implementation no longer make sense.
 
-let isFlushingAct = false;
+let isFlushingAct = null;
 let isInsideThisAct = false;
 
 // TODO: Yes, this is confusing. See above comment. We'll refactor it.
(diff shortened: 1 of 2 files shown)


packages/react-reconciler/src/ReactFiberWorkLoop.new.js
```diff
@@ -3726,6 +3726,13 @@ function shouldForceFlushFallbacksInDEV() {
...
   }
+
+  // `isFlushingAct` is used by ReactTestRenderer version of `act`.
+  if (isFlushingAct != null) {
+    // Flush callbacks at the end.
+    return isFlushingAct;
+  }
+
   // `IsThisRendererActing.current` is used by ReactTestUtils version of `act`.
   if (IsThisRendererActing.current) {
```
rickhanlonii, line 3737: The core issue for the bug I'm fixing is that `IsThisRendererActing.current` is not reset to false before we check `shouldForceFlushFallbacksInDEV`, meaning that we miss the `isFlushingAct = true` value, and we schedule the fallback [here](https://github.com/facebook/react/blob/a27a0d704712a3d3386e8229f438eb94bf8e7992/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L976) instead of committing it.

rickhanlonii: @acdlite what was the expected behavior here? Should we be checking for isFlushingAct first?

Is the test that failed correct or is it actually wrong?

acdlite:
I think we need to start over on the act stuff. It's too convoluted. I'm loathe to add another hacky workaround.

To unblock the RN sync, can you mitigate by disabling the test or something? This doesn't seem important enough to block the whole sync.

rickhanlonii:
Yeah I can do that 👍

PR #19472, opened 2020-07-28
Add "unstable_" prefix to experimental mutable source APIs

bvaughn:
This change affects only the NPM `@experimental` release channel. The following two API names will be changed in the next release:
* `createMutableSource` -> `unstable_createMutableSource`
* `useMutableSource` -> `unstableUseMutableSource`

This should have been done originally. Sorry for the oversight.

### Why is this important?

Since experimental APIs do not appear in any stable release, we should be free to change or remove them later without breaking any downstream projects.

However downstream projects may add feature detection, e.g.:
```js
const useSomeAPI = React.useSomeAPI || fallbackUseSomeAPI;
```

The pattern shown above assumes that the version of `useSomeAPI` that exists in the Experimental channel today has the same API contract as the final `useSomeAPI` API that we\'ll eventually ship to stable.

To discourage feature detection, we add an "unstable_" prefix to all unstable/experimental APIs.'


bvaughn:
Adding the "unstable_" prefix to `useMutableSource` revealed an interesting test failure.

In a development build, the code is:
```js
function useMutableSource(source, getSnapshot, subscribe) {
  var dispatcher = resolveDispatcher();
  return dispatcher.useMutableSource(source, getSnapshot, subscribe);
}
// ...
exports.unstable_useMutableSource = useMutableSource;
```

But in a production build, the code is:
```js
exports.unstable_useMutableSource=function(a,b,c){return U().useMutableSource(a,b,c)};
```

This results in a different runtime stack, which is observable due to how we're using `error-stack-parser`.

In development, the parsed stack looks like:
```js
{
  columnNumber: 21,
  lineNumber: 197,
  fileName: '.../packages/react/src/ReactHooks.js',
  functionName: 'Object.useMutableSource',
  source: '    at Object.useMutableSource (.../packages/react/src/ReactHooks.js:197:21)'
},
```

But in a production build the stack looks like:
```js
{
  columnNumber: 21,
  lineNumber: 1623,
  fileName: '.../build/node_modules/react/cjs/react.development.js',
  functionName: 'Object.useMutableSource [as unstable_useMutableSource]',
  source: '    at Object.useMutableSource [as unstable_useMutableSource] (.../build/node_modules/react/cjs/react.development.js:1623:21)'
},
```

The `functionName` difference causes the export to be mistaken for a *custom hook* and so it gets added into the hooks tree twice.

I "fixed" this in 8f7ea9d (made both builds behavior similarly) by removing the initialization call to `useMutableSource` inside of `getPrimitiveStackCache()` - which is indeed what we seem to be doing for the other composite hooks (`useTransition`, `useDeferredValue`, and `useOpaqueIdentifier`). I'm not sure if this is the right fix (if the other hooks are intentional or an oversight) but it seems to produce the best (consistent) behavior so for now, I'll go with that.

markerikson:
Out of curiosity, any ETA on `useMutableSource`actually getting shipped?

PR #19479, opened 2020-07-29
Add postTask browser scheduler implementation 

rickhanlonii:
## Overview

This diff adds a Scheduler implementation that uses the experimental postTask API provided by Chrome.

## Details

The `postTask` API is an experimental first step towards adding main thread scheduling to browsers.

It's based off of the [main-thread-scheduling](https://github.com/WICG/main-thread-scheduling) proposal and help solves the issue of coordinating work at various priorities. In React's case, it allows a browser-native strategy for breaking up long tasks (such as rendering) into smaller chucks that yield to other, possibly more important, work in between.

React already has a mechanism for breaking rendering into smaller chucks that yield to other work, and we use it in concurrent mode. The way we do that today is to use the MessageChannel API to "schedule" work:

```js
let scheduledWork;

function doWork() {
  scheduledWork();
}

// Create a MessageChannel that we will post work to.
// the browser will call onmessage when it's our turn to work.
const channel = new MessageChannel();
const port = channel.port2;
channel.port1.onmessage = doWork;

// Entry point to scheduling work
export function scheduleWork(callback) {
  scheduledWork = callback;
  port.postMessage(null);
};
```

The change in this diff replaces the above strategy with [postTask](https://github.com/WICG/main-thread-scheduling/blob/master/PrioritizedPostTask.md):

```js
export function scheduleWork(callback) {
  scheduler.postTask(callback);
};
```

## Why

There are a few disadvantages of the MessageChannel implementation.

First, all work is scheduled with the same priority. This initial change doesn't solve this because we're still posting all tasks with the same priority. In the future, the scheduler API [will allow us to](https://github.com/WICG/main-thread-scheduling/blob/master/PrioritizedPostTask.md#priorities) post tasks with different priorities levels and the browser will do the scheduling.

Second, the MessageChannel implementation requires us to do work in 5ms increments and yield to the browser to see if there's more browser work to do. In the future, this can be improved with additional scheduler APIs which will allow us to check if there [is pending input](https://github.com/WICG/is-input-pending). This will allow React to continue rendering until the browser has more important work today, and fill more gaps in under utilized CPU time.

Finally, the MessageChannel implementation does not have a way to continue work that's interrupted. This means, when rendering is paused to yield to the browser, we cannot continue until the event loops comes back around again and we will pay the overhead of letting other (lower-priority) task run in between. What we want to do is ask the browser to yield to higher priority work, but continue with our work before anything else that's the same priority or lower is allowed to complete. This will be possible with [`scheduler.yield`](https://github.com/WICG/main-thread-scheduling/blob/master/YieldAndContinuation.md).

For the first step though, we'll continue to manage priority in user land and yield to entire event loop every 5s.

### DIFF
--- a/packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
@@ -11,6 +11,16 @@ const perf = window.performance;
 const setTimeout = window.setTimeout;
 const clearTimeout = window.clearTimeout;
 
+// This check should be done upstream but do it again for a clear failure message.
+if (global.scheduler === undefined || global.scheduler.postTask === undefined) {
+  throw new Error('Cannot use postTask Scheduler without global.scheduler');
+}
+
+function postTask(callback) {
+  // Use experimental Chrome Scheduler postTask API.
+  global.scheduler.postTask(callback);
+}
+
 let getNow;
 if (typeof perf === 'object' && typeof perf.now === 'function') {
   getNow = () => perf.now();
@@ -72,12 +82,12 @@ const performWorkUntilDeadline = () => {
       } else {
         // If there's more work, schedule the next message event at the end
         // of the preceding one.
-        port.postMessage(null);
+        postTask(performWorkUntilDeadline);
       }
     } catch (error) {
       // If a scheduler task throws, exit the current browser task so the
       // error can be observed.
-      port.postMessage(null);
+      postTask(performWorkUntilDeadline);
       throw error;
     }
   } else {
@@ -85,15 +95,11 @@ const performWorkUntilDeadline = () => {
   }
 };
 
-const channel = new MessageChannel();
-const port = channel.port2;
-channel.port1.onmessage = performWorkUntilDeadline;
-
 export const requestHostCallback = function(callback) {
   scheduledHostCallback = callback;
   if (!isMessageLoopRunning) {
     isMessageLoopRunning = true;
-    port.postMessage(null);
+    postTask(performWorkUntilDeadline);
   }
 };
 
--- a/packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
+++ b/packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
@@ -6,30 +6,20 @@
  */
 
 // Capture local references to native APIs, in case a polyfill overrides them.
-const date = window.Date;
 const perf = window.performance;
 const setTimeout = window.setTimeout;
 const clearTimeout = window.clearTimeout;
-
-// This check should be done upstream but do it again for a clear failure message.
-if (global.scheduler === undefined || global.scheduler.postTask === undefined) {
-  throw new Error('Cannot use postTask Scheduler without global.scheduler');
-}
 
 function postTask(callback) {
   // Use experimental Chrome Scheduler postTask API.
   global.scheduler.postTask(callback);
 }
 
-let getNow;
-if (typeof perf === 'object' && typeof perf.now === 'function') {
-  getNow = () => perf.now();
-} else {
-  const initialTime = date.now();
-  getNow = () => date.now() - initialTime;
+function getNow() {
+  return perf.now();
 }
 
-let isMessageLoopRunning = false;
+let isTaskLoopRunning = false;
 let scheduledHostCallback = null;
 let taskTimeoutID = -1;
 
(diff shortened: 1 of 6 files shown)


packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
```diff
@@ -5,246 +5,117 @@
...
+
+// Since we yield every frame regardless, `requestPaint` has no effect.
+export const requestPaint = function() {};
+
+export const forceFrameRate = function(fps) {
+  if (fps < 0 || fps > 125) {
+    // Using console['error'] to evade Babel and ESLint
+    console['error'](
+      'forceFrameRate takes a positive int between 0 and 125, ' +
+        'forcing frame rates higher than 125 fps is not unsupported',
```
aweary, line 57: > not unsupported

😀

packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
```diff
@@ -5,246 +5,117 @@
...
+// `isInputPending` is not available. Since we have no way of knowing if
+// there's pending input, always yield at the end of the frame.
+export const shouldYieldToHost = function() {
+  return getNow() >= deadline;
+};
+
+// Since we yield every frame regardless, `requestPaint` has no effect.
+export const requestPaint = function() {};
+
+export const forceFrameRate = function(fps) {
```
acdlite, line 52: This was added a few years ago for lower end devices but I'm not sure we even need it anymore.

Let's just leave this as a noop for now.

packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
```diff
@@ -5,246 +5,117 @@
...
-import {enableIsInputPending} from '../SchedulerFeatureFlags';
+// Capture local references to native APIs, in case a polyfill overrides them.
+const date = window.Date;
+const perf = window.performance;
+const setTimeout = window.setTimeout;
+const clearTimeout = window.clearTimeout;
+
+// This check should be done upstream but do it again for a clear failure message.
+if (global.scheduler === undefined || global.scheduler.postTask === undefined) {
+  throw new Error('Cannot use postTask Scheduler without global.scheduler');
```
acdlite, line 16: Let's assume these exist. We'll check in the wrapper module in www.

packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
```diff
@@ -5,246 +5,117 @@
...
-  cancelHostTimeout = function() {
-    clearTimeout(_timeoutID);
-  };
-  shouldYieldToHost = function() {
-    return false;
-  };
-  requestPaint = forceFrameRate = function() {};
+let getNow;
+if (typeof perf === 'object' && typeof perf.now === 'function') {
+  getNow = () => perf.now();
```
acdlite, line 26: Same here. Any browser that supports `postTask`  will also support `performance.now`.

packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
```diff
@@ -5,246 +5,117 @@
...
-    }
-    if (typeof cancelAnimationFrame !== 'function') {
-      // Using console['error'] to evade Babel and ESLint
-      console['error'](
-        "This browser doesn't support cancelAnimationFrame. " +
-          'Make sure that you load a ' +
-          'polyfill in older browsers. https://fb.me/react-polyfills',
-      );
-    }
+let isMessageLoopRunning = false;
```
acdlite, line 32: `isTaskLoopRunning`?

packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
```diff
@@ -5,246 +5,117 @@
...
-    }
-    // Yielding to the browser will give it a chance to paint, so we can
-    // reset this.
-    needsPaint = false;
-  };
-
-  const channel = new MessageChannel();
-  const port = channel.port2;
-  channel.port1.onmessage = performWorkUntilDeadline;
+export const requestHostCallback = function(callback) {
```
acdlite, line 98: Nit: Switch to function declarations, i.e. `export function requestHostCallback`

I don't think it makes much of a difference and Closure and Rollup will optimize either form just as easily.

Mostly I just prefer to follow the convention we use in (most of) the other modules.

packages/scheduler/src/forks/SchedulerHostConfig.post-task.js
rickhanlonii: Fixed in `SchedulerHostConfig.default.js`

PR #19483, opened 2020-07-29
Event propagation test suite

gaearon:
We've been making a series of changes to the event system with the goal of improving interop in the scenarios where a React app is nested into another React app — or even a non-React app. In particular, because React used event delegation to the document, `stopPropagation()` did not correctly work between two Reacts, as events would not bubble between them.

@trueadm has landed most of the work related to this, but we don't have a very good regression suite for events. We also don't have a place where all the behaviors are documented. So I'm adding a test suite that verifies the expected propagation scenarios, both within *and* across isolated React roots.

This is a suite of jsdom tests so that we notice regressions early on. Initially, I wrote it as a browser fixture (and it's not hard to convert back), but I found that it wasn't particularly useful since I was emulating the events anyway, and we know well which ones bubble (or don't bubble) in the browsers. So I just hardcoded that in the test.

The suite doesn't include *all* events, but it includes <s>all *kinds* of events</s> (edit: there's a few more I need to add) for which we have different behavior. We can add more concrete events to it but I don't think this is a blocker for merging the suite.

Note that the suite would *not* pass in React 16. It documents the current behavior, with tests verifying that cross-root propagation works as well as it can. React 16 did not do a very good job at that.

I will add some other comments inline.

### DIFF
--- a/packages/react-dom/src/__tests__/ReactDOMEventPropagation-test.js
+++ b/packages/react-dom/src/__tests__/ReactDOMEventPropagation-test.js
@@ -1,0 +1,1327 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+'use strict';
+
+describe('ReactDOMEventListener', () => {
+  let React;
+  let OuterReactDOM;
+  let InnerReactDOM;
+  let container;
+
+  beforeEach(() => {
+    jest.resetModules();
+    React = require('react');
+    jest.isolateModules(() => {
+      OuterReactDOM = require('react-dom');
+    });
+    jest.isolateModules(() => {
+      InnerReactDOM = require('react-dom');
+    });
+    expect(OuterReactDOM).not.toBe(InnerReactDOM);
+  });
+
+  afterEach(() => {
+    cleanup();
+  });
+
+  function cleanup() {
+    if (container) {
+      OuterReactDOM.unmountComponentAtNode(container);
+      document.body.removeChild(container);
+      container = null;
+    }
+  }
+
+  function render(tree) {
+    cleanup();
+    container = document.createElement('div');
+    document.body.appendChild(container);
+    OuterReactDOM.render(tree, container);
+  }
+
+  describe('bubbling events', () => {
+    // This test will fail in legacy mode (only used in WWW)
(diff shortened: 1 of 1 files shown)


packages/react-dom/src/__tests__/ReactDOMEventPropagation-test.js
```diff
@@ -0,0 +1,1327 @@
...
+
+  beforeEach(() => {
+    jest.resetModules();
+    React = require('react');
+    jest.isolateModules(() => {
+      OuterReactDOM = require('react-dom');
+    });
+    jest.isolateModules(() => {
+      InnerReactDOM = require('react-dom');
+    });
```
gaearon, line 24: This is to ensure we're actually testing the interop scenario.

packages/react-dom/src/__tests__/ReactDOMEventPropagation-test.js
```diff
@@ -0,0 +1,1327 @@
...
+        type: 'div',
+        reactEvent: 'onClick',
+        nativeEvent: 'click',
+        dispatch(node) {
+          node.click();
+        },
+      });
+    });
+
+    it('onKeyPress', () => {
```
gaearon, line 63: Each event has a separate test case so that it's easy to "focus" on it.

packages/react-dom/src/__tests__/ReactDOMEventPropagation-test.js
```diff
@@ -0,0 +1,1327 @@
...
+      testNativeBubblingEvent({
```
gaearon, line 64: However, the actual tests are a sequence of several in each case. I didn't want to repeat because we want consistent behavior across events in the same group. So we call a helper function that runs all tests. If you want to run one case in particular you can always comment out the calls inside of that function.

packages/react-dom/src/__tests__/ReactDOMEventPropagation-test.js
```diff
@@ -0,0 +1,1327 @@
...
+          node.dispatchEvent(e);
+        },
+      });
+    });
+
+    it('onFocus', () => {
+      testNativeBubblingEvent({
+        type: 'input',
+        reactEvent: 'onFocus',
+        nativeEvent: 'focusin',
```
gaearon, line 162: This one is interesting. It's in the bubbling category because it's effectively an alias for the browser `onfocusin` which *does* bubble. So it's React event name that's a misnomer.

PR #19486, opened 2020-07-29
Use feature flags for React Native in the test renderer

rickhanlonii:
## Overview

The React Native renderer has `enableComponentStackLocations` disabled, but the react-test-renderer for React Native has `enableComponentStackLocations ` enabled. This causes issues syncing React to React Native.

### DIFF
--- a/packages/shared/forks/ReactFeatureFlags.test-renderer.native.js
+++ b/packages/shared/forks/ReactFeatureFlags.test-renderer.native.js
@@ -1,0 +1,56 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import typeof * as FeatureFlagsType from 'shared/ReactFeatureFlags';
+import typeof * as ExportsType from './ReactFeatureFlags.test-renderer';
+
+export const debugRenderPhaseSideEffectsForStrictMode = false;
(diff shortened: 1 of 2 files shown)


packages/shared/forks/ReactFeatureFlags.test-renderer.native.js
```diff
@@ -0,0 +1,56 @@
...
+export const warnAboutDefaultPropsOnFunctionComponents = false;
+export const warnAboutStringRefs = false;
+export const disableLegacyContext = false;
+export const disableSchedulerTimeoutBasedOnReactExpirationTime = false;
+export const enableTrustedTypesIntegration = false;
+export const disableTextareaChildren = false;
+export const disableModulePatternComponents = false;
+export const warnUnstableRenderSubtreeIntoContainer = false;
+export const warnAboutSpreadingKeyToJSX = false;
+export const enableComponentStackLocations = false;
```
rickhanlonii, line 44: This file is copied from `ReactFeatureFlags.test-renderer.js` except for this flag.

gaearon:
How will this work for RN GitHub? Will it also use the fork?

rickhanlonii:
Good point.

The issue I'm solving is that when we sync React to RN from master and fix the tests internally, the OSS tests fail. Disabling this setting now allows us the react-test-renderer to match the way open source currently works, but that will break when we update RN to the next OSS React release.

To fix this fully, we'll need to either turn this flag off for all react-test-renderer, or completely upgrade React Native to use the new component stack formatting.

PR #19487, opened 2020-07-29
Fix onGot/LostPointerCapture events

gaearon:
Open to better ideas but this will work.

### DIFF
--- a/packages/react-dom/src/events/DOMPluginEventSystem.js
+++ b/packages/react-dom/src/events/DOMPluginEventSystem.js
@@ -470,8 +470,15 @@ export function listenToReactEvent(
       }
     }
   } else {
-    // Check if the react event ends in "Capture"
-    const isCapturePhaseListener = reactEvent.substr(-7) === 'Capture';
+    const isCapturePhaseListener =
+      reactEvent.substr(-7) === 'Capture' &&
+      // Edge case: onGotPointerCapture and onLostPointerCapture
+      // end with "Capture" but that's part of their event names.
+      // The Capture versions would end with CaptureCapture.
+      // So we have to check against that.
+      // This check works because none of the events we support
+      // end with "Pointer".
+      reactEvent.substr(-14, 7) !== 'Pointer';
     listenToNativeEvent(
       dependencies[0],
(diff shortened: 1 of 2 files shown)


packages/react-dom/src/events/DOMPluginEventSystem.js
```diff
@@ -470,8 +470,15 @@ export function listenToReactEvent(
...
-    // Check if the react event ends in "Capture"
-    const isCapturePhaseListener = reactEvent.substr(-7) === 'Capture';
+    const isCapturePhaseListener =
+      reactEvent.substr(-7) === 'Capture' &&
+      // Edge case: onGotPointerCapture and onLostPointerCapture
+      // end with "Capture" but that's part of their event names.
+      // The Capture versions would end with CaptureCapture.
+      // So we have to check against that.
+      // This check works because none of the events we support
+      // end with "Pointer".
```
rickhanlonii, line 480: Why not check the exact event names?

gaearon: Wanted to avoid extra bytes.

trueadm: We already reference both strings though right? It might actually be less bytes to do it that way

gaearon: We only have the browser event versions, right? I don't think those would help since the *actual capture* versions of these events would also have the same strings as `dependencies[0]`. It's only the React event name that tells us the story.

trueadm: I was thinking that we could maybe dynamically create a Set of all capture React names rather than do anything substr lookups. I.e around here https://github.com/facebook/react/blob/master/packages/react-dom/src/events/DOMEventProperties.js#L171 where we populate all the events with a single dependency.

trueadm: Probably overkill though.

gaearon: Yeah that might be nicer. Not sure what's faster though.

PR #19489, opened 2020-07-29
Check PassiveStatic instead of Passive

acdlite:
Saves us from having to set a flag on `current` during the layout phase.

Could result in some redundant traversal, since PassiveStatic includes effects that don't need clean-up. But it's worth it to remove the work from the layout phase.

While I was editing this, I also re-arranged it so that we check the `effectTag` check before we check the `tag`, since the `effectTag` check is the one that's more likely to fail.


acdlite:
I usually make a Quip and then forget about it :D

PR #19493, opened 2020-07-30
Use unprefixed animation/transition events

gaearon:
Let's drop this polyfill.

The prefixed version of `transitionend` only existed in old iOS, Opera, and old (pre 4.4.4) Androids.

All of those have very low market share, about 0.3% in total.

Screenshot 2020-07-30 at 16 57 17

Screenshot 2020-07-30 at 16 57 47

Screenshot 2020-07-30 at 17 13 45
Screenshot 2020-07-30 at 17 13 56


gaearon:
Looks like animation prefixing is still needed for IE 10. Bummer.
https://www.sitepoint.com/css3-animation-javascript-event-handlers/

gaearon:
Nah, breaking IE10 seems bad. Will have to wait for 18.

PR #19500, opened 2020-07-30
Move passive logic out of layout phase

acdlite:
### tl;dr

Some minor refactoring so that I can delete `effect.tag |= HookHasEffect` from the synchronous deletion path.

---

We don't need to check HookHasEffect during a deletion; all effects are unmounted.

So we also don't have to set HookHasEffect during a deletion, either.

This allows us to remove the last remaining passive effect logic from the synchronous layout phase. (For deletions; there's still  some passive logic in the layout phase for mounts/updates that we've yet to remove.)


acdlite:
It's really only the third commit that's interesting. The first two are minor factoring changes that I noticed and didn't feel like opening a separate PR for.

Issue #18937, opened 2020-05-16
Bug: eslint exhaustive-deps doesn't recognize JSXIdentifier as used


sophiebits:
```
function Component({reportType}) {
  let Report = Reports[reportType];
  let child = useMemo(() => {
    return <Report />;
  }, [Report]);
```

says that `Report` is an unnecessary dependency. It's not.

This is because eslint scope `.references` does not include JSX. I _suspect_ if you pull in

https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-uses-vars.js

to your app's config then this won't happen. But maybe we should do something in the Hooks plugin itself too (since the others aren't officially recommended by the React team anyway).

muskeinsingh:
Can I pick this up?

yanneves:
@sophiebits would what you're doing not be better solved by `React.memo()` instead?

```jsx
const Child = React.memo(() => <Report />)
```

https://reactjs.org/docs/react-api.html#reactmemo

sophiebits:
@yanneves Not quite. See the second code example in https://reactjs.org/docs/hooks-faq.html#how-to-memoize-calculations for more detail on the pattern I'm using.

eps1lon:
Duplicate of #18051

PR #19506, opened 2020-07-31
add failing test (#18051)

delca85:
## Summary

Trying to solve #18051  and #18937 .
`exhaustive-deps` eslint rule does not recognize a `JSXIdentifier` as needed dependency.

This is because eslint scope `.references` does not include JSX.
I have pushed only a failing test with the aim to keep on working on it.

## Test Plan

Failing test where an hook uses a `JSXIdentifier` that is properly added to deps array while the eslint rules marks it as unnecessary.

### DIFF
--- a/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
+++ b/packages/eslint-plugin-react-hooks/__tests__/ESLintRuleExhaustiveDeps-test.js
@@ -1376,6 +1376,16 @@ const tests = {
         }
       `,
     },
+    {
+      code: normalizeIndent`
+        function Component({ reportType }) {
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks, this is a nice test case.

delca85:
I have tried the [code](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-uses-vars.js) suggested by Sophie Bits in #18937.
Through that code, eslint is able to mark as used (when they are) the variables related to JSX Component.
I am not able to insert it in the flow used by the `exhaustive-deps` rule. I don't know if there is a way to populate the `scope.references` array using the suggested code and adding JSX Component variables to that array.

PR #19507, opened 2020-07-31
Support untagged releases

gaearon:
This makes a few changes to the release script.

Previously, it allowed to pass multiple `--tags`. But in practice there's only a few scenarios we want to support:

* `--tags next`
* `--tags experimental`
* `--tags latest next` (because they should be in sync)

Additionally, we'd like to support a way to publish without tagging at all. (I tried this in #19505 but it led to a bug where we publish under the `undefined` tag.)

**In this PR, I replace `--tags` with a singular `--tag` argument:**

* `--tag next` publishes `@next`
* `--tag experimental` publishes `@experimental`
* `--tag latest` publishes `@latest` **and** `@next`
* No `--tag` doesn't tag the release at all
  - (Technically, it tags it with a temporary `@untagged` tag since npm doesn't allow tag-less publish, and then deletes that tag right away.)

---

I'll do some testing before merge.


scripts/release/README.md
```diff
@@ -170,7 +176,9 @@ Upon completion, this script provides instructions for tagging the Git commit th
 **Specify a `--dry` flag when running this script if you want to skip the NPM-publish step.** In this event, the script will print the NPM commands but it will not actually run them.
 
 #### Example usage
-To publish a release to NPM as both `next` and `latest`:
+To publish a release to NPM as `latest`:
 ```sh
-scripts/release/publish.js --tags latest
```
gaearon, line 175: This was wrong before. In particular, to publish both `next` and `latest`, you *had to* pass them both. This seems to have gotten lost in an edit (9e158c). Now it actually works.

scripts/release/publish-commands/validate-tag.js
```diff
@@ -0,0 +1,57 @@
...
+    cwd,
+    'build',
+    'node_modules',
+    arbitraryPackageName,
+    'package.json'
+  );
+  const {version} = await readJson(packageJSONPath);
+  if (version.indexOf('0.0.0') === 0) {
+    if (tag === 'latest') {
+      if (version.indexOf('experimental') !== -1) {
```
bvaughn, line 23: Looks like you meant to use `===` here? (Or you meant to reverse the log messages below?)

Given the number of `indexOf` checks in this section, it might be clearer if you assigned that to a variable e.g. `isExperimentalVersion`.

acdlite:
> `--tags latest next (because they should be in sync)`

I slightly disagree with this, even though it's usually true in practice.

`latest` could lag behind `next`, because it can take us several days or weeks to test a particular revision before we actually release it. In the meantime, we may have published additional `next` releases. If `next` were published using a cron job (which is the eventual goal), then this is pretty likely.

Another case is patches, which are cut from a branch on top of the `latest` release. Those have no relationship to the `next` channel.

acdlite:
Another way of looking at it is that every `latest` release is promoted from a release that was already part of the `next` channel. So either it's the tip of the channel, in which case re-publishing doesn't make much sense. Or the main branch has advanced in the meantime.

gaearon:
I see. Should I revert?

bvaughn:
> Another case is patches, which are cut from a branch on top of the latest release. Those have no relationship to the next channel.

This seems like a reasonably compelling argument.

PR #19508, opened 2020-07-31
Revert "Support untagged releases"

gaearon:
Reverts facebook/react#19507

See https://github.com/facebook/react/pull/19507#issuecomment-667296183.

I didn't think of the use case where `latest` and `next` intentionally disagree. Since sometimes they do, and sometimes they don't, being able to pass multiple tags seems useful. Which goes against what facebook/react#19507 was trying to do.

I can do a more limited version of the change instead which just makes `untagged` a default if `tags` are not specified.


acdlite:
Cool with reverting or fixing forward, not a big deal either way.

> I can do a more limited version of the change instead which just makes untagged a default if tags are not specified.

Yeah that part is cool! I've done that manually a few times.

PR #19514, opened 2020-08-02
Register more node types that are used later as JSXIdentifiers

Andarist:
## Summary

I got a report in the Emotion's monorepo that when our Babel plugin gets used Fast Refresh stops working correctly:
https://github.com/emotion-js/emotion/issues/1953#issuecomment-667673237

I've debugged it down to this:
```js
const Comp = styled('div')({ color: 'hotpink' })
const App = () => <Comp/>
```
not being treated as a component.

Styled API allows this - and it's not that uncommon, because it allows object styles to be used (which has better types support out of the box). We also precompute some concatenated styles, source map, labels and hash identifiers within our Babel plugin at build time and we pass the result to our `styled` calls so it can be consumed faster and with ease at runtime - that's why with Emotion this issue is even more significant.

## Test Plan

`yarn test --watch`

### DIFF
--- a/packages/react-refresh/src/ReactFreshBabelPlugin.js
+++ b/packages/react-refresh/src/ReactFreshBabelPlugin.js
@@ -140,9 +140,6 @@ export default function(babel, opts = {}) {
             } else if (calleeType === 'MemberExpression') {
               // Could be something like React.forwardRef(...)
               // Pass through.
-            } else {
-              // More complicated call.
-              return false;
             }
             break;
           }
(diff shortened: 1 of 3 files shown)


packages/react-refresh/src/ReactFreshBabelPlugin.js
```diff
@@ -140,9 +140,6 @@ export default function(babel, opts = {}) {
             } else if (calleeType === 'MemberExpression') {
               // Could be something like React.forwardRef(...)
               // Pass through.
-            } else {
```
Andarist, line 143: from what I understand this was just an optimization as the declared binding is being tested later more to check if it should be registered anyway

gaearon:
Have you checked that with this change, it works as expected end to end?

Andarist:
@gaearon yes, I was testing this locally by tweaking my node_modules. I've rechecked it now and pushed out a branch for verification (patch-package ❤️ ): https://github.com/Andarist/emotion-react-refresh-demo/commit/378397813e1fbcfa43162bf9440555abddf25b6f . Works like a charm.

gaearon:
OK, let's give this a try,

PR #19520, opened 2020-08-03
Remove the deprecated React Flare event system

trueadm:
Note: we should not merge this PR until we have fully switched React Flare off internally.

This PR removes the React Flare event system from the React codebase, along with any event responders and their associated tests. The React Flare event system was experimental, and we deprecated it earlier this year to make way for alternative APIs that can achieve similar, if not better, results (based off our internal research).

As part of this PR, I also converted some SSR hyrdration tests from Flare to use `createEventHandle`, as not lose coverage in important places. I also removed some the event system flags that are no longer required anymore (including the base plugin system type, as there's now only a single event system in play).


packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
```diff
@@ -2216,17 +2217,18 @@ describe('ReactDOMServerPartialHydration', () => {
...
 
     function Button() {
-      const listener = React.DEPRECATED_useResponder(TestResponder, {});
-      return <a DEPRECATED_flareListeners={listener}>Click me</a>;
+      const ref = React.useRef(null);
+      const effect = () => {
+        return setClick(ref.current, onEvent);
+      };
+      if (isServerRendering) {
+        React.useEffect(effect);
```
gaearon, line 2227: What is this for? You might as well omit it since server doesn't call effects.

packages/react-dom/src/__tests__/ReactDOMServerSelectiveHydration-test.internal.js
```diff
@@ -398,11 +407,7 @@ describe('ReactDOMServerSelectiveHydration', () => {
 
     // This should synchronously hydrate the root App and the second suspense
     // boundary.
-    const preventDefault = jest.fn();
-    target.virtualclick({preventDefault});
-
-    // The event should have been canceled because we called preventDefault.
-    expect(preventDefault).toHaveBeenCalled();
```
gaearon, line 405: What is this change about?

packages/react-dom/src/events/ReactDOMUpdateBatching.js
```diff
@@ -102,27 +87,8 @@ export function discreteUpdates(fn, a, b, c, d) {
...
-  // we are safe from the resolution differences. If the time stamp is 0
-  // we bail-out of preventing the flush, which can affect semantics,
-  // such as if an earlier flush removes or adds event listeners that
-  // are fired in the subsequent flush. However, this is the same
-  // behaviour as we had before this change, so the risks are low.
-  if (
-    !isInsideEventHandler &&
-    (!enableDeprecatedFlareAPI ||
-      timeStamp === 0 ||
-      lastFlushedEventTimeStamp !== timeStamp)
```
gaearon, line 123: Can you remind me why we no longer care about time stamps? Why was this Flare-specific?

trueadm: It was Flare specific, but we could re-include it. Some of the SimpleEventPlugin tests might fail again though.

packages/react-dom/src/__tests__/ReactDOMServerSelectiveHydration-test.internal.js
trueadm: It was to do with how Flare and its usePress worked. I changed this to an alternative version

packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
trueadm: It’s still valid as this is what we do internally. We can’t use useEffect though as the event listeners won’t be added at the right time.

gaearon: I'm still not following. I'm saying that

```js
if (isServerRendering) {
  React.useEffect(effect);
} else {
  React.useLayoutEffect(effect);
}
```

should be completely equivalent to writing

```js
if (!isServerRendering) {
  React.useLayoutEffect(effect);
}
```

because on the server, effects are ignored.

packages/react-dom/src/events/ReactDOMUpdateBatching.js
gaearon: I'm not arguing to change this, but more just curious why we only enabled it for Flare listeners when batching itself is generic.

packages/react-dom/src/__tests__/ReactDOMServerPartialHydration-test.internal.js
trueadm: Gotcha!

packages/react-dom/src/events/ReactDOMUpdateBatching.js
trueadm: With Flare, we dispatched events again if they were discrete, meaning in some cases we'd over-flush and it causes UI issues internally. Given we've moved back to only handling the flushing on incoming discrete events, then we don't necessarily need this behavior anymore. I can see how we might want to re-introduce it in the future though, maybe one for @sebmarkbage to think about when he revists some of the flushing related work.

PR #19521, opened 2020-08-03
Turn off new component stacks for React Native

rickhanlonii:
# Overview

This internal React build for React Native running off the main branch is not ready for the new component stack (e.g it breaks LogBox). To unblock the sync to React Native and test the React main branch we're turning off the new component stacks by using the React Native feature flags for React.

I've confirmed that the only major feature flag changes between the default and RN flags is the component stack flag.

### DIFF
--- a/scripts/rollup/forks.js
+++ b/scripts/rollup/forks.js
@@ -137,6 +137,19 @@ const forks = Object.freeze({
             return 'shared/forks/ReactFeatureFlags.testing.www.js';
         }
         return 'shared/forks/ReactFeatureFlags.testing.js';
+      case 'react':
+        switch (bundleType) {
+          case FB_WWW_DEV:
+          case FB_WWW_PROD:
+          case FB_WWW_PROFILING:
+            return 'shared/forks/ReactFeatureFlags.www.js';
+          case RN_FB_DEV:
+          case RN_FB_PROD:
+          case RN_FB_PROFILING:
+            return 'shared/forks/ReactFeatureFlags.native-fb.js';
+          default:
+            return 'shared/ReactFeatureFlags.js';
+        }
       default:
         switch (bundleType) {
           case FB_WWW_DEV:


scripts/rollup/forks.js
```diff
@@ -137,6 +137,19 @@ const forks = Object.freeze({
...
+          case FB_WWW_DEV:
+          case FB_WWW_PROD:
+          case FB_WWW_PROFILING:
+            return 'shared/forks/ReactFeatureFlags.www.js';
+          case RN_FB_DEV:
+          case RN_FB_PROD:
+          case RN_FB_PROFILING:
+            return 'shared/forks/ReactFeatureFlags.native-fb.js';
+          default:
+            return 'shared/ReactFeatureFlags.js';
```
gaearon, line 151: This should just be missing or null? See the case at the bottom.

rickhanlonii: That fails eslint `no-fallthrough`, and I think for good reason - it's nice to explicitly know which flags are used in the default case. For example, I didn't really know or believe that we used the non-scoped `ReactFeatureFlags.js` file for all of React. If this fell through (especially if there's a case added after this before the default case later), it would be more difficult to see which flag is used.

rickhanlonii:
Tested this by running a git diff before and after `yarn build` to confirm that the only change is to the way we gather component stack frames in React-dev.js, as expected.

scripts/rollup/forks.js
gaearon: I'm just a bit confused why this particular case is different from other switch'es above which don't have a default case.

rickhanlonii: I see, yes this could just return.

rickhanlonii: This was changed in https://github.com/facebook/react/pull/19522 anyway 👍

